Issue #758, opened 2020-10-17
error saving notification acknowledgement


chappjc:
I'm getting the following (after a swap and clicking on the bell menu, I think:

```
2020-10-16 21:47:44.379 [ERR] CORE: error saving notification acknowledgement for 1a4500cf8f610128: notification not found
2020-10-16 21:47:44.379 [ERR] CORE: error saving notification acknowledgement for 9caae748dd033b7c: notification not found
2020-10-16 21:47:44.379 [ERR] CORE: error saving notification acknowledgement for 75fb2fd05174feac: notification not found
2020-10-16 21:47:44.379 [ERR] CORE: error saving notification acknowledgement for 92e807fdac7de5f7: notification not found
```

_Originally posted by @chappjc in https://github.com/decred/dcrdex/pull/757#issuecomment-710733137_

PR #776, opened 2020-10-22
app: set new notes by wiping old ones first

chappjc:
Resolves https://github.com/decred/dcrdex/issues/758

### DIFF
--- a/client/db/bolt/db.go
+++ b/client/db/bolt/db.go
@@ -12,6 +12,7 @@ import (
 	"sort"
 	"time"
 
+	"decred.org/dcrdex/client/db"
 	dexdb "decred.org/dcrdex/client/db"
 	"decred.org/dcrdex/dex"
 	"decred.org/dcrdex/dex/encode"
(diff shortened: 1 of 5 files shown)


client/db/bolt/db.go
```diff
@@ -12,7 +12,6 @@ import (
 	"sort"
 	"time"
 
-	"decred.org/dcrdex/client/db"
 	dexdb "decred.org/dcrdex/client/db"
```
buck54321, line 15: Didn't even know this would compile like this.

buck54321 (commented):
I think this is a javascript bug you could correct real quick. 

Replace the commented line at https://github.com/decred/dcrdex/blob/3a26fcbf5b154ff87f20335e9ff5485a7e213ac3/client/webserver/site/src/js/app.js#L445

with

```javascript
this.notes = []
```

I did a few quick tests and it seems to correct the issue.

chappjc:
> ```js
> this.notes = []
> ```

Ah, so this wipes out old notes from local storage from previous runs (rather than adding to them)?

buck54321:
Just from memory here, but outdated the browser storage will be overwritten in `prependNoteElement` too.

chappjc:
Hmm.  I've noticed that when I nuke dex.db and start fresh, I sometimes end up with ancient notes from before, and these were causing the harmless errors in https://github.com/decred/dcrdex/issues/758.  Not always though.  I think not acking them before nuking is the key to repro.

buck54321:
That's right. Before, the ancient notes were loaded into `this.notes` on startup. When `app.setNotes` is called from `RegisterPage` or `LoginPage`, new notes were appended to the ancient notes in `this.notes`. Only the new ones would be displayed in the bell. Then, when you clicked the bell, we were trying to ack all of the unacked notes in `this.notes`, which still contained the ancients.

client/db/bolt/db.go
chappjc: removed all the go changes

chappjc:
This is now included https://github.com/decred/dcrdex/pull/781, but merging anyway.

PR #777, opened 2020-10-23
client: startup improvements and minor fixes

chappjc:
Startup now prints a couple helpful new messages:

```
2020-10-22 19:19:54.820 [INF] WEB: Web server listening on 127.0.0.1:6758

	****  OPEN IN YOUR BROWSER TO TRADE  --->  http://127.0.0.1:6758  ****

2020-10-22 19:19:55.118 [DBG] CORE: notify: |POKE| (conn) DEX connected - DEX at 127.0.0.1:7232 has connected
2020-10-22 19:19:55.120 [DBG] CORE: Broadcast timeout = 1m0s, ticking every 20s
2020-10-22 19:19:55.120 [INF] CORE: Connected to DEX server at 127.0.0.1:7232 and listening for messages.
2020-10-22 19:19:55.120 [DBG] CORE: connectDEX for 127.0.0.1:7232 completed, checking account...
2020-10-22 19:19:55.120 [DBG] CORE: dex connection to 127.0.0.1:7232 ready
2020-10-22 19:19:55.120 [INF] CORE: Successfully connected to 1 out of 1 DEX servers
2020-10-22 19:19:55.120 [INF] CORE: Loaded 1 active orders.
2020-10-22 19:19:55.120 [INF] CORE: Loaded 0 active match orders
2020-10-22 19:19:55.120 [WRN] CORE: 

	 ****  IMPORTANT: You have 1 active orders on 127.0.0.1:7232. LOGIN immediately!  **** 
```

The second "IMPORTANT" message is only displayed if there are active orders in the client DB.  This involves loading up the active orders prior to DEX auth, but they are discarded after being counted, to be loaded again on connect when the trackedTrades are created.  This is slightly inefficient, but it's simple.

The second commit fixes an issue where post-swap resumed trades had coinsLocked=true but no stored coins.  Normally coinsLocked is set to false in swapMatches, but on restart it creates a trackedTrade with it set to true always.  This was resulting in a harmless but confusing message when a trade got retired (`[WRN] CORE: Unable to return btc funding coins: cannot return zero coins`).

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -2845,7 +2845,7 @@ func (c *Core) initialize() {
 	c.connMtx.RLock()
 	c.log.Infof("Successfully connected to %d out of %d DEX servers", len(c.conns), len(accts))
 	for dexName, dc := range c.conns {
-		activeOrders, _ := c.dbOrders(dc)
+		activeOrders, _ := c.dbOrders(dc) // non-nil error will load 0 orders, and any subsequent db error will cause a shutdown on dex auth or sooner
 		if n := len(activeOrders); n > 0 {
 			c.log.Warnf("\n\n\t ****  IMPORTANT: You have %d active orders on %s. LOGIN immediately!  **** \n", n, dexName)
 		}
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -3218,6 +3218,9 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
 				notifyErr("Order coin error", "Source coins retrieval error for %s %s: %v", unbip(wallets.fromAsset.ID), tracker.token(), err)
 				continue
 			}
+			// NOTE: change and changeLocked are not set even if the funding
+			// coins were loaded from the DB's ChangeCoin.
+			tracker.coinsLocked = true
 			tracker.coins = mapifyCoins(coins)
 		}
 
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -2784,7 +2784,7 @@ func (c *Core) AssetBalance(assetID uint32) (*WalletBalance, error) {
 func (c *Core) initialize() {
 	accts, err := c.db.Accounts()
 	if err != nil {
-		c.log.Errorf("Error retrieve accounts from database: %v", err)
+		c.log.Errorf("Error retrieving accounts from database: %v", err) // panic?
 	}
 	var wg sync.WaitGroup
 	for _, acct := range accts {
@@ -2793,7 +2793,7 @@ func (c *Core) initialize() {
 			defer wg.Done()
 			host, err := addrHost(acct.Host)
 			if err != nil {
-				c.log.Errorf("skipping loading of %s due to address parse error: %w", acct.Host, err)
+				c.log.Errorf("skipping loading of %s due to address parse error: %v", acct.Host, err)
 				return
 			}
 			dc, err := c.connectDEX(acct)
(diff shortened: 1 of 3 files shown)


client/core/core.go
```diff
@@ -2793,7 +2793,7 @@ func (c *Core) initialize() {
 			defer wg.Done()
 			host, err := addrHost(acct.Host)
 			if err != nil {
-				c.log.Errorf("skipping loading of %s due to address parse error: %w", acct.Host, err)
+				c.log.Errorf("skipping loading of %s due to address parse error: %v", acct.Host, err)
```
chappjc, line 2796: %w really only applies to wrapping `error` values, and in fact some libraries (maybe our logger) can panic on unrecognized printf specifiers.

client/core/core.go
```diff
@@ -2844,6 +2844,12 @@ func (c *Core) initialize() {
 	wg.Wait()
 	c.connMtx.RLock()
 	c.log.Infof("Successfully connected to %d out of %d DEX servers", len(c.conns), len(accts))
+	for dexName, dc := range c.conns {
+		activeOrders, _ := c.dbOrders(dc)
```
chappjc, line 2848: non-nil error will load 0 orders, and any subsequent db error will cause a shutdown on dex auth or sooner.

client/core/core.go
```diff
@@ -3202,6 +3218,9 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
 				notifyErr("Order coin error", "Source coins retrieval error for %s %s: %v", unbip(wallets.fromAsset.ID), tracker.token(), err)
 				continue
 			}
+			// NOTE: change and changeLocked are not set even if the funding
+			// coins were loaded from the DB's ChangeCoin.
```
buck54321, line 3222: Good point. We should look into doing this a little more carefully, if not just for accounting purposes. Right now we're not displaying the current or previous change coins anywhere, so it's of little consequence, but still a little sloppy.

client/core/trade.go
```diff
@@ -186,7 +186,7 @@ func (t *trackedTrade) coreOrderInternal() *Order {
 // BUY order.
 // lockedAmount should be called with the mtx >= RLocked.
 func (t *trackedTrade) lockedAmount() (locked uint64) {
-	if t.coinsLocked { // implies no swap has been sent
+	if t.coinsLocked { // implies no swap has been sent or the trade has been resumed on restart after swap
```
buck54321, line 189: Isn't this not true anymore because of your work here? Only pre-swap startup loads will have coins locked, right?

buck54321 (approved):
If you could just set me straight on the one comment. 

I'll look into writing a simpler **client/db** method to check for > 0 active orders later.

client/core/core.go
buck54321: Leave a comment?

client/core/trade.go
chappjc: `coinsLocked==true` means one of two things currently:
- swap has NOT yet been sent, and this is normal negotiation with no restart
- swap has been sent, but there was a restart that put change coin in `coins` and set `coinsLocked=true` (instead of `change` where it would go in the first bullet)

chappjc: I revised this comment.  Let me know if it makes more sense now.
BTW, I'd like to test a resumed trade for a partially filled order that is still booked.

buck54321: I see now. We're almost certainly handling this wrong at startup, but let's look at that part separately.

chappjc: Right I am hesitant to change it to start assigning to change/changeLocked, but that would be closer to restoring things properly.

PR #781, opened 2020-10-25
ui: bugs and usability fixes

buck54321:
1. Fix bug with depth chart unbinding its resize handler
2. Links to transaction outputs open in new tab
3. Harmonize "pre-split" and "pre-size" at the UI level
4. Use more familiar "immediate or cancel" for "match for one epoch only". Add a tooltip with more explanation
5. Change order submission button text and color to indicate buy/sell

6. Improves the setNotes -> prependNoteElement loop to skip intermediate saves
7. Add a warning in order the order confirmation dialog about keeping the client and blockchains daemons running during settlement.

### DIFF
--- a/client/webserver/site/src/html/markets.tmpl
+++ b/client/webserver/site/src/html/markets.tmpl
@@ -186,7 +186,7 @@
                       <input id="tifNow" class="form-check-input" type="checkbox" value="">
                       <label class="form-check-label" for="tifNow">
                         Immediate or cancel
-                        <span class="ico-info fs12" data-tooltip="If the order doesn't fully match during the next match cycle, any unmatched quantity will be canceled."></span>
+                        <span class="ico-info fs12" data-tooltip="If the order doesn't fully match during the next match cycle, any unmatched quantity will not be booked or matched again. Taker-only order."></span>
                       </label>
                     </div>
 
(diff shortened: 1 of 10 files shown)


chappjc:
Grr, I tweaked the cache buster too.

buck54321:
That's a conflict I welcome any time. It'd be great to automate the cache buster away somehow. Or maybe catch js or css changes that don't include a cache buster change in testing.

chappjc:
Conflict with cache buster again after your last one merge.

client/webserver/site/src/html/markets.tmpl
```diff
@@ -182,16 +182,17 @@
                     <div class="mt-2 fs14 text-right" id="orderPreview"></div>
 
                     {{- /* TIME-IN-FORCE CHECK BOX */ -}}
-                    <div class="mt-2 text-left pl-4" id="tifBox">
+                    <div class="my-1 text-left pl-4" id="tifBox">
                       <input id="tifNow" class="form-check-input" type="checkbox" value="">
                       <label class="form-check-label" for="tifNow">
-                        Match for one epoch only
+                        Immediate or cancel
+                        <span class="ico-info fs12" data-tooltip="If the order doesn't fully match during the next match cycle, any unmatched quantity will be canceled."></span>
```
chappjc, line 189: Slightly hesitant about the "cancel" wording here because the order status will be executed if there's no match or it just doesn't go on the books.

client/webserver/site/src/html/markets.tmpl
```diff
@@ -391,6 +394,13 @@
           <button id="vSubmit" type="button" class="w-75 mt-1 justify-content-center fs15 bg2 selected">Submit</button>
         </div>
       </div>
+      <hr class="dashed my-4">
+      <div class="disclaimer mt-3 fs14">
+        <span class="red">IMPORTANT</span>: Trades take time to settle, and you cannot turn off the DEX client software, or
+        the <span data-unit="quote"></span> or <span data-unit="base"></span> blockchain daemons, until settlement is complete.
+        Settlement can complete as quickly as a few minutes or as long as a few hours.
```
chappjc, line 401: The average user probably won't know what "daemons" is talking about.  I'd be fine calling it "wallet software (e.g. dcrwallet and bitcoind)" so there is no confusion.

In a second one-sentence paragraph, this might be a good place to remind users to ensure their blockchains are done syncing, although we'll have to check that discussed.

client/webserver/site/src/html/markets.tmpl
chappjc: Could we work with a variation of "Do not allow to become booked" or "Taker-only"?

client/webserver/site/src/html/markets.tmpl
buck54321: You mind if I leave this in anticipation of #785?

chappjc (approved):

Issue #501, opened 2020-06-18
client,server: verify BTC fee estimation is enabled before making transactions


chappjc:
We need to start making BTC transactions with the fee rate obtained from `estimatesmartfee 1` rather than the static `FeeRate` (https://github.com/decred/dcrdex/issues/502), and when we do that it will be necessary to ensure fee estimation is functional in the running bitcoin node.

Even on a fully-syncd bitcoind node, time is needed for fee estimates to be available.

This can be demonstrated without wiping the node's entire data directory by clearing the fee_estimates.dat file (maybe also mempool.dat):

```
$  # bitcoind datadir
$  mv fee_estimates.dat fee_estimates.dat.bak
$  mv mempool.dat mempool.dat.bak
```

Start bitcoind (mainnet):

```
$   ./bitcoind -server -txindex -debug=rpc
```

Fully-syncd on mainnet:

```
$   ./bitcoin-cli getblockchaininfo | jq '.chain, .blocks, .headers, .initialblockdownload'
"main"
635320
635320
false
```

Fee estimation is not available despite being at the network tip:

```
$   ./bitcoin-cli estimatesmartfee 1
{
  "errors": [
    "Insufficient data or no feerate found"
  ],
  "blocks": 0
}
```

Even after running for several minutes and observing 2 blocks, with thousands of txns in mempool:

```
$   ./bitcoin-cli getblockchaininfo | jq '.chain, .blocks, .headers, .initialblockdownload'
"main"
635322
635322
false

$   ./bitcoin-cli getmempoolinfo
{
  "loaded": true,
  "size": 2496,
  "bytes": 873194,
  "usage": 4014208,
  "maxmempool": 300000000,
  "mempoolminfee": 0.00001000,
  "minrelaytxfee": 0.00001000
}

$   ./bitcoin-cli estimatesmartfee 1
{
  "errors": [
    "Insufficient data or no feerate found"
  ],
  "blocks": 1
}
```

This demonstrates that without recent fee data and/or a fee_estimates.dat bitcoind/bitcoin-qt will not make fee estimation available.  As such, we should be checking `estimatesmartfee` to see if bitcoin is ready.  It started working for me after 5 blocks were mined:

```
 $   ./bitcoin-cli -rpcuser=btc -rpcpassword=btc123 -rpcwallet= estimatesmartfee 1
{
  "feerate": 0.00038189,
  "blocks": 2
}
$   ./bitcoin-cli getblockchaininfo | jq '.blocks'
635325
[...]

PR #785, opened 2020-10-26
multi: no trades while blockchain syncing

buck54321:
Client waits for blockchain sync to complete. Server too, but server also waits for fee estimation to be primed. 

I have a bug to chase down in the tests, but ready for a peek.


server/market/market.go
```diff
@@ -599,21 +602,35 @@ func (m *Market) Cancelable(oid order.OrderID) bool {
 // means: (1) an order in the book or epoch queue, (2) type limit with
 // time-in-force standing (implied for book orders), and (3) AccountID field
 // matching the provided account ID.
-func (m *Market) CancelableBy(oid order.OrderID, aid account.AccountID) bool {
+func (m *Market) CancelableBy(oid order.OrderID, aid account.AccountID) (bool, error) {
```
buck54321, line 605: I made this change to troubleshoot, but I think it might be useful. Technically, could just return the `error` without the `bool`, but I'm not crazy about the semantics there.

go.mod
```diff
@@ -26,13 +26,14 @@ require (
 	github.com/decred/dcrwallet/wallet/v3 v3.2.1
 	github.com/decred/go-socks v1.1.0
 	github.com/decred/slog v1.1.0
+	github.com/gcash/bchd v0.17.0
```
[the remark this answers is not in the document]
buck54321: I don't even know what gcash or bchd are. But I cleaned up the screwed up go.mods and go.sums now, so I guess you were commenting on an older commit.

server/market/market.go
```diff
@@ -599,21 +602,35 @@ func (m *Market) Cancelable(oid order.OrderID) bool {
...
+	if ord == nil {
+		return false, ErrTargetNotActive
 	}
-	return false
+
+	lo, ok := ord.(*order.LimitOrder)
+	if !ok {
+		return false, ErrTargetNotCancelable
+	}
+	if lo.Force == order.StandingTiF {
```
chappjc, line 627: This appears backwards.  Standing is cancellable, immediate is not.

server/market/market.go
```diff
@@ -599,21 +602,35 @@ func (m *Market) Cancelable(oid order.OrderID) bool {
...
+		return false, ErrTargetNotCancelable
+	}
+	if lo.AccountID == aid {
+		return false, ErrCancelNotPermitted
```
chappjc, line 631: This also seems backwards.

client/asset/dcr/dcr.go
```diff
@@ -1847,6 +1851,22 @@ func (dcr *ExchangeWallet) shutdown() {
...
 }
 
+// SyncStatus is information about the blockchain sync status.
+func (dcr *ExchangeWallet) SyncStatus() (bool, float32, error) {
+	chainInfo, err := dcr.node.GetBlockChainInfo()
+	if err != nil {
+		return false, 0, fmt.Errorf("getblockchaininfo error: %w", err)
+	}
+	toGo := chainInfo.Headers - chainInfo.Blocks
+	if toGo > 1 {
```
chappjc, line 1861: basically thinking `if toGo > 1 || chainInfo.InitialBlockDownload {`

client/webserver/site/src/html/markets.tmpl
```diff
@@ -2,6 +2,7 @@
 <span data-state="sleeping" class="ico-sleeping grey d-hide"></span>
 <span data-state="locked" class="ico-locked grey d-hide"></span>
 <span data-state="unlocked" class="ico-unlocked d-hide"></span>
+<span data-state="syncing" class="ico-sync fs12 d-hide" data-tooltip="placeholder text"></span>
```
chappjc, line 5: tooltip still placeholder

chappjc (changes requested):

server/asset/dcr/dcr.go
```diff
@@ -251,6 +254,26 @@ func (dcr *Backend) ValidateSecret(secret, contract []byte) bool {
...
+		return false, fmt.Errorf("GetBlockChainInfo error: %w", err)
+	}
+	if chainInfo.Headers-chainInfo.Blocks > 1 {
+		return false, nil
+	}
+	if dcr.net == dex.Mainnet {
+		_, err = dcr.FeeRate()
+		if err != nil {
+			dcr.log.Debugf("Synced = false because of FeeRate error = %v", err)
+		}
```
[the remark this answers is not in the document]
buck54321: I actually thought that this was suggested in #501. I'm removing the `false` return value, but we should consider other solutions for a default fee on the server. Especially with bitcoin max fees now >= 300 sats/vbyte, during periods of low activity, max fees might be 5x or 10x the fees we'd get from `estimatesmartfee` on a primed node.

client/webserver/site/src/html/markets.tmpl
buck54321: I replace the text with javascript before it's ever seen. I'm just avoiding an empty string in case I try to evaluate its truthiness somewhere.

client/core/core.go
```diff
@@ -1007,13 +1007,52 @@ func (c *Core) connectedWallet(assetID uint32) (*xcWallet, error) {
...
+				w.mtx.Unlock()
+				if synced {
+					return
+				}
+				c.notify(newWalletStateNote(w.state()))
+
+			case <-innerCtx.Done():
+				return
+			}
+		}()
```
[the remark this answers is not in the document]
buck54321: > The `w.connector.Wait` that cancels the context is to stop the initial sync check if the `xcWallet` shuts down before sync?

Yes. I've added a comment to that effect.

client/core/core.go
```diff
@@ -1008,13 +1011,57 @@ func (c *Core) connectedWallet(assetID uint32) (*xcWallet, error) {
...
+		// loop.
+		innerCtx, cancel := context.WithCancel(c.ctx)
+		go func() {
+			w.connector.Wait()
+			cancel()
+		}()
+		// The syncer loop.
+		go func() {
+			for {
+				ticker := time.NewTicker(syncTickerPeriod)
```
chappjc, line 1032: I think you want the NewTicker outside the loop otherwise you get a new one each iteratoin.  Also a defer ticker.Stop to prevent a ticker leak.

buck54321: Third try's the charm.

client/core/core.go
```diff
@@ -1028,8 +1028,8 @@ func (c *Core) connectWallet(w *xcWallet) error {
 		}()
 		// The syncer loop.
 		go func() {
+			ticker := time.NewTicker(syncTickerPeriod)
```
chappjc, line 1031: Still gonna leak the ticker though. need a `defer ticker.Stop()`.

server/asset/dcr/dcr.go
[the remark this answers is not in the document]
chappjc: Not sure about 5x or 10x even, but a certain percent over the `estimatesmartfee` result seems like a great idea given how easy it is to get buried by other txns where blocks are taking too long.  I was thinking more like 150% though, so it would indicate 450 for 300 e.g. and that would have been more than sufficient throughout this latest congestion.

client/core/core.go
```diff
@@ -1008,13 +1011,58 @@ func (c *Core) connectedWallet(assetID uint32) (*xcWallet, error) {
...
+
 // Connect to the wallet if not already connected. Unlock the wallet if not
 // already unlocked.
 func (c *Core) connectAndUnlock(crypter encrypt.Crypter, wallet *xcWallet) error {
 	if !wallet.connected() {
-		err := wallet.Connect(c.ctx)
+		err := c.connectWallet(wallet)
 		if err != nil {
-			return newError(walletErr, "error connecting %s wallet: %v", unbip(wallet.AssetID), err)
+			return err
```
chappjc, line 1065: Say the client does something like starts a swap and their computer restarts or something happens where they restart and a bitcoind is till syncing.  Even if that's not enough to take action we'd still want to unlock the wallet I think, but this won't unlock the wallet if it's still syncing.

How would a use handle this state?  What would the problem be to unlock the wallet regardless of sync status?

chappjc: Oh, I think I misread the flow.  `connectWallet` does NOT till returns a non-nil error for not syncd status, so it would still proceed to unlock, right?

server/asset/btc/btc.go
```diff
@@ -363,6 +379,51 @@ func (btc *Backend) blockInfo(verboseTx *btcjson.TxRawResult) (blockHeight uint3
...
+	Blocks               int64  `json:"blocks"`
+	Headers              int64  `json:"headers"`
+	BestBlockHash        string `json:"bestblockhash"`
+	InitialBlockDownload bool   `json:"initialblockdownload"`
+}
+
+// getBlockchainInfo sends the getblockchaininfo request and returns the result.
+func (btc *Backend) getBlockchainInfo() (*getBlockchainInfoResult, error) {
+	var chainInfo *getBlockchainInfoResult
+	err := btc.call(methodGetBlockchainInfo, nil, &chainInfo)
```
chappjc, line 420: Here and in client asset, you're using two levels of indirection (`**getBlockchainInfoResult`) for the final input arg to `call`.  That might work but we're not doing that in the other calls.

server/asset/dcr/dcr.go
```diff
@@ -75,13 +75,16 @@ type dcrNode interface {
...
+	GetBlockChainInfo() (*chainjson.GetBlockChainInfoResult, error)
 }
 
 // Backend is an asset backend for Decred. It has methods for fetching output
 // information and subscribing to block updates. It maintains a cache of block
 // data for quick lookups. Backend implements asset.Backend, so provides
 // exported methods for DEX-related blockchain info.
 type Backend struct {
+	// net is the dex.Network.
+	net dex.Network
```
chappjc, line 87: I might be missing the usage, but it seems `net` is an extraneous change here and in btc.

server/market/market.go
```diff
@@ -46,7 +46,9 @@ const (
 	ErrQuantityTooHigh        = Error("order quantity exceeds user limit")
 	ErrDuplicateCancelOrder   = Error("equivalent cancel order already in epoch")
 	ErrTooManyCancelOrders    = Error("too many cancel orders in current epoch")
-	ErrInvalidCancelOrder     = Error("cancel order account does not match targeted order account")
+	ErrCancelNotPermitted     = Error("cancel order account does not match targeted order account")
+	ErrTargetNotActive        = Error("target order not active on this market")
+	ErrTargetNotCancelable    = Error("targeted order is not a limit order with immediate time-in-force")
```
chappjc, line 51: error string says immediate instead of standing

client/core/core.go
buck54321: That's right. The sync loop is run in a goroutine. `connectWallet` is just a convenient place to start it.

server/asset/dcr/dcr.go
buck54321: It was from when I was doing `Synced -> false` for `estimatesmartfee` errors on mainnet. Removing.

chappjc:
Looking good in my tests so far, but I'm realizing that with dcrwallet, client doesn't know that wallet might still be syncing even if dcrd is sync'd.  Not sure if that is of high concern though since wallet syncs quite quickly because it's usually local with dcrd, but something we might want to address at somepoint.

chappjc (approved):

PR #786, opened 2020-10-26
client/core: handle cancel order for trades with reconciled status

chappjc:
If a trade and a cancel order targeting that order were in epoch status, and order status reconciliation happened, the cancel order would stay in epoch status forever.

This updates `(*dexConnection).reconcileTrades` so that when a trade order gets its status updated (EDIT: to a status other than canceled), any associated cancel order is unlinked and put into revoked status (retired) to it is not loaded forever on startup.

This also updates `(c *Core).dbTrackers` to retire cancel orders loaded from the DB that have no associated active trade. This is a "should not happen" case, but which I hit when testing unrelated server changes.

This also moves the final log statement in dexc's main.go to *before* the log rotator closes.


client/core/core.go
```diff
@@ -608,6 +608,19 @@ func (dc *dexConnection) reconcileTrades(srvOrderStatuses []*msgjson.OrderStatus
 		previousStatus := trade.metaData.Status
 		newStatus := order.OrderStatus(srvOrderStatus.Status)
 		trade.metaData.Status = newStatus
+		// If there is an associated cancel order, and we are revising the
+		// status of the targeted order, we can infer the cancel order is done.
+		if trade.cancel != nil {
+			cid := trade.cancel.ID()
+			err := trade.db.UpdateOrderStatus(cid, order.OrderStatusRevoked)
```
itswisdomagain, line 615: This bit, for updating the deleted cancel order status, is not done in `trackedTrade.deleteStaleCancelOrder()`, perhaps adding that there will prevent such cancel orders from being loaded from db on restart.

client/core/core.go
```diff
@@ -608,6 +608,19 @@ func (dc *dexConnection) reconcileTrades(srvOrderStatuses []*msgjson.OrderStatus
...
+			if err != nil {
+				dc.log.Errorf("Error updating status in db for cancel order %v to executed", cid)
+			}
+			// Unlink the cancel order from the trade.
+			trade.cancel = nil
+			trade.metaData.LinkedOrder = order.OrderID{}
+		}
```
itswisdomagain, line 622: What do you think of refactoring this whole section into a `trackedTrade.deleteCancelOrder()` method, which can then be called from here AND from the existing `trackedTrade.deleteStaleCancelOrder()` method?

Could also call this new `trackedTrade.deleteCancelOrder()` method when a match/nomatch request sets a trade's status to Executed (any associated cancel order would have failed to match and updating the status accordingly should prevent the order from being reloaded from db on restart).

client/core/core.go
```diff
@@ -608,6 +608,19 @@ func (dc *dexConnection) reconcileTrades(srvOrderStatuses []*msgjson.OrderStatus
 		previousStatus := trade.metaData.Status
 		newStatus := order.OrderStatus(srvOrderStatus.Status)
 		trade.metaData.Status = newStatus
+		// If there is an associated cancel order, and we are revising the
+		// status of the targeted order, we can infer the cancel order is done.
```
itswisdomagain, line 612: Only orders that were previously active (Epoch / Booked) can get here, either from a direct reconciliation using the status returned from the `connect` response or using the `order_status` route. Regardless of how this update is triggered, **stale** cancel orders for such knownActive trades are deleted, before the update takes place.

It'd be more accurate to note here _(but don't know if it's really important or better to leave it to reader's inference)_ that any associated cancel order would have been deleted if it had stayed at Epoch status longer than 2 epochs, and this specifically targets cancel orders whose epoch may not have passed.

itswisdomagain (changes requested):

client/core/core.go
chappjc: That's not it though.  I got here testing another PR that made the server revoke the epoch status trade without telling client, so that the associated cancel would never have a reference to it.  That's also where the other block in https://github.com/decred/dcrdex/pull/786/files#diff-13a4a6b12c1ae844ef90151638f194427cfdc545fceb58d52c0f6f33b81d859bR3121 helps.

chappjc: Noting these lines: https://github.com/decred/dcrdex/pull/786/files#diff-13a4a6b12c1ae844ef90151638f194427cfdc545fceb58d52c0f6f33b81d859bR650-R657
It couldn't even get to `deleteStaleCancelOrder` in the case I was testing because the trade wasn't in `srvActiveOrderStatuses`.

itswisdomagain: Yeah, retiring obsolete cancels on load from db is vital, some cancels will just not get retired naturally, as you've seen and pointed out.

I'm not able to visit the code section referenced in the link in your second comment, but looking at the code again, you're right. Trades not in `srvActiveOrderStatuses` don't get to `deleteStaleCancelOrder`.

client/core/core.go
chappjc: Good suggestion to reuse this as `deleteCancelOrder`.  I have it called now from `deleteStaleCancelOrder` and here in the `updateOrder` closure.

Thinking about `nomatch(tradeID)`, I'm hesitant to delete any cancel order from it because if an associated cancel order failed to match there would be a `nomatch(cancelID)`, which does the right thing for the cancel order's status.  So it should be redundant, but might also be undesirable in some way.

client/core/core.go
```diff
@@ -608,6 +608,11 @@ func (dc *dexConnection) reconcileTrades(srvOrderStatuses []*msgjson.OrderStatus
...
+		// Update the status of the cancel order and unlink it from the trade.
+		trade.deleteCancelOrder()
```
buck54321, line 614: What if we're revising the status to `OrderStatusCanceled`? Wouldn't we be unlinking a correctly linked order?

This would be the case if we crashed after preimage reveal, but before match notifications.

client/core/core.go
```diff
@@ -3090,6 +3098,7 @@ func (c *Core) dbTrackers(dc *dexConnection) (map[order.OrderID]*trackedTrade, e
 		if err != nil {
 			c.log.Errorf("Error setting cancel order info %s: %v", co.ID(), err)
 		}
+		delete(cancels, cancelID)
```
buck54321, line 3101: It appears that you're adding to `cancels` and deleting from `cancels` in the same loop, but there's no guarantee that we'll see the cancel order before it's trade, and if the trade comes first, the cancel will be added and never deleted, and you'll end up setting valid cancels as revoked.

buck54321 (changes requested):

client/core/core.go
chappjc: Ah, good catch.  Didn't notice that.

chappjc:
All set.  Good catches @buck54321

client/core/core.go
```diff
@@ -3072,20 +3088,32 @@ func (c *Core) dbTrackers(dc *dexConnection) (map[order.OrderID]*trackedTrade, e
 		return nil, err
 	}
 
+	// Index all of the cancel orders so we can account for them when loading
+	// the trade orders. Whatever remains is orphaned.
+	unknownCanels := make(map[order.OrderID]struct{})
```
buck54321, line 3093: unknownCan**c**els

buck54321 (approved):
Just one typo.

Issue #474, opened 2020-06-13
cmd/dcrdex: Unable to stop app with active orders on the book


itswisdomagain:
Stalls at `DEX: BookRouter shutdown`. Only happens when there are unmatched orders at shutdown time (only noticed when shutting down almost immediately after placing the orders anyway).

```
2020-06-13 13:25:42.598 [TRC] MKT: Received order 75b367a80686147738aca21cc221e8bce0ffd8d78ad3ca709e88280985155358 at 2020-06-13 13:25:42.598 +0000 UTC
2020-06-13 13:25:42.669 [TRC] MKT: Received order 769f3abd95a485ecba4e93cc59781f0d7b22edf588b964ade7f9a2967a623538 at 2020-06-13 13:25:42.669 +0000 UTC
2020-06-13 13:25:42.679 [TRC] COMM: Websocket output handler done for peer 127.0.0.1
2020-06-13 13:25:42.679 [INF] COMM: Shutting down link for 127.0.0.1 with 0 queued messages.
2020-06-13 13:25:42.679 [TRC] COMM: Disconnected websocket client 127.0.0.1
2020-06-13 13:25:42.679 [TRC] COMM: Websocket output handler done for peer 127.0.0.1
2020-06-13 13:25:42.679 [INF] COMM: Shutting down link for 127.0.0.1 with 0 queued messages.
2020-06-13 13:25:42.679 [TRC] COMM: Disconnected websocket client 127.0.0.1
2020-06-13 13:25:48.000 [DBG] AUTH: error sending request: peer disconnected
^C2020-06-13 13:32:41.813 [INF] MAIN: Received signal (interrupt). Shutting down...
2020-06-13 13:32:41.813 [INF] MAIN: Stopping DEX...
2020-06-13 13:32:41.813 [INF] DEX: Stopping subsystems...
2020-06-13 13:32:41.813 [INF] COMM: RPC server shutting down...
2020-06-13 13:32:41.814 [DBG] COMM: RPC listener done for 127.0.0.1:17273
2020-06-13 13:32:41.814 [INF] COMM: RPC server shutdown complete
2020-06-13 13:32:41.814 [INF] DEX: Comms Server shutdown.
2020-06-13 13:32:41.814 [INF] MKT: Book router terminating for market "dcr_btc"
2020-06-13 13:32:41.814 [INF] DEX: BookRouter shutdown.
^C2020-06-13 13:32:46.273 [INF] MAIN: Shutdown signaled. Already shutting down...
[...]

PR #787, opened 2020-10-26
server/{dex,market}: cleaner shutdown

chappjc:
**server/dex**

This reorders the subsystem stack so that DEX manager shutdown is ordered properly now.  Namely:

- <s>comms server shuts down after the subsystems that use it (all but assets)</s> comms didn't move because we want it _started_ after the subsystems that use it are up for the sake of _incoming_ message handling, even though stopping it before those dependent subsystems stop can result in dropped _outgoing_ messages
- book router shuts down after the markets that use it

**server/market**

When a market is stopped by context cancellation (not a market suspend admin command), it does not wait for the current epoch to close, so any orders in that truncated epoch were just being dropped without changing their status from epoch to executed (how a no-match/fail is normally handled when epochs are processed by the matcher). This updates the primary defer function in `(*Market).Run` so that if there are orders left in the `epochOrders` map after the closed epoch pipeline is drained, their statuses are change in the DB so that they aren't reported to clients as active orders for eternity.

Resolves https://github.com/decred/dcrdex/issues/474, although it didn't hang any more.

### DIFF
--- a/server/dex/dex.go
+++ b/server/dex/dex.go
@@ -195,11 +195,12 @@ func (cr *configResponse) remarshal() {
 // Stop shuts down the DEX. Stop returns only after all components have
 // completed their shutdown.
 func (dm *DEX) Stop() {
-	log.Infof("Stopping subsystems...")
+	log.Infof("Stopping all DEX subsystems.")
 	for _, ssw := range dm.stopWaiters {
+		log.Infof("Stopping %s...", ssw.name)
 		ssw.Stop()
 		ssw.WaitForShutdown()
-		log.Infof("%s shutdown.", ssw.name)
+		log.Infof("%s is now shut down.", ssw.name)
 	}
 	if err := dm.storage.Close(); err != nil {
 		log.Errorf("DEXArchivist.Close: %v", err)
@@ -250,7 +251,7 @@ func NewDEX(cfg *DexConf) (*DEX, error) {
 	startSubSys := func(name string, r dex.Runner) {
 		ssw := dex.NewStartStopWaiter(r)
 		ssw.Start(ctx)
-		stopWaiters = append([]subsystem{{ssw, name}}, stopWaiters...)
+		stopWaiters = append([]subsystem{{ssw, name}}, stopWaiters...) // top of stack
 	}
 
 	abort := func() {
@@ -474,6 +475,14 @@ func NewDEX(cfg *DexConf) (*DEX, error) {
 		}
 	}
 
+	// Client comms RPC server.
+	server, err := comms.NewServer(cfg.CommsCfg)
+	if err != nil {
+		abort()
+		return nil, fmt.Errorf("NewServer failed: %v", err)
+	}
+	startSubSys("Comms Server", server)
+
 	// Having enumerated all users with booked orders, configure the AuthManager
 	// to expect them to connect in a certain time period.
 	authMgr.ExpectUsers(usersWithOrders, cfg.BroadcastTimeout)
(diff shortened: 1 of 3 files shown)


server/dex/dex.go
```diff
@@ -474,6 +475,14 @@ func NewDEX(cfg *DexConf) (*DEX, error) {
...
 	}
 
+	// Client comms RPC server.
+	server, err := comms.NewServer(cfg.CommsCfg)
+	if err != nil {
+		abort()
+		return nil, fmt.Errorf("NewServer failed: %v", err)
+	}
+	startSubSys("Comms Server", server)
+
```
buck54321, line 485: I see how this would achieve the desired effect of shutting down the server last, but is it ok to start the server before the lower-level systems that it relies on? I realize of course that sub-system `Run` methods are run in goroutines with no guarantee of sequence anyway, but we still want ot start the server last, no?

buck54321 (commented):
> comms server shuts down after the subsystems that use it (all but assets)

Is this to allow the sub-systems to continue sending messages. Seems like we'd want to immediately end incoming messages before the shutting down the lower-level systems.

server/dex/dex.go
chappjc: That's a good point about the comms order.  The issue I noticed was that messages at shutdown are dropped if comms goes does first.  But your point about start-up is taken.

server/market/market.go
```diff
@@ -790,6 +790,21 @@ func (m *Market) Run(ctx context.Context) {
...
+			if co, ok := ord.(*order.CancelOrder); ok {
+				if err := m.storage.FailCancelOrder(co); err != nil {
+					log.Errorf("Failed set orphaned epoch cancel order %v as executed: %v", oid, err)
+				}
+				continue
+			}
+			if err := m.storage.ExecuteOrder(ord); err != nil {
+				log.Errorf("Failed set orphaned epoch trade order %v as executed: %v", oid, err)
+			}
+		}
```
chappjc, line 807: The suspend command should be used to complete the epoch, but ctrl+c should at least not ditch orders in epoch status even though they do not go through matching.

server/dex/dex.go
chappjc: The comms subsys is moved back to its original location, but I'll let you resolve this comment in case you have thoughts about the outgoing messages on shutdown.

server/market/market.go
buck54321: Just noticing the difference between the executed status here, and the revoked status when we do a book purge. Seems like similar situations.

chappjc: Yeah, maybe revoked is better for the trades too.  Will switch it.  Really just avoiding eternal epoch status orders though.

chappjc: OK, so we could use RevokeOrder{Uncounted} for the trade order, but that creates a cancel order to go with it.  The general idea of order revocation is that it's a forced unbooking, a server-generated cancellation.  So while executed may not fix exactly here going from epoch->executed is more inline with how this is handled elsewhere, such as epoch orders not matching becoming executed, not revoked.

buck54321: Would an operator have any way to know by looking at the order row in the DB whether this order was executed via matching vs. dirty shutdown?

chappjc: No preimage.

chappjc: Hmm, need to double check that these won't count as a preimage miss, but had be pretty darn uncommon to do a dirty shutdown without suspending the markets first.

chappjc: OK, so by making it executed we also escape having it counted as a preimage miss because the preimage miss query looks for **revoked** status orders.
Still thinking through other pitfalls with this though.

buck54321 (approved):

PR #788, opened 2020-10-27
client: add support for dcr spv wallets

itswisdomagain:
The following is a summary of changes made to support dcr spv wallets:
- Validate and (re-)broadcast txData when auditing contracts.
- Replace getblockchaininfo with syncstatus to determine sync status.
- In spv mode, use block cfilters to find tx confirmations if `gettxout` and `gettransaction` can't find a tx. Both rpcs only returns results for wallet unspent outputs.
- Re-audit contracts after they receive the required confirmations but before sending counter-swap (if taker) and before sending redeem (if maker).

The dcr harness is modified to use SPV for the second trading wallet to facilitate testing.


itswisdomagain:
The commit message on https://github.com/decred/dcrdex/pull/788/commits/9b51633a09c1169e2c40e8c4959b3a3367e5b23f provides my reasoning for making clients re-audit contracts before acting.

Copied below for posterity:
```
There is a very slight possibility that contract outputs that pass
audit while in mempool or in some cases before they are broadcasted
(rawtx audit) may differ from the output later observed in a block
for the same coin.

This risk is higher for spv wallets that will mostly only perform
audits on rawtxs before broadcasting the txs, without a guarantee
that the tx is accepted to the mempool. A malicious actor could
broadcast a different tx with same hash (theoretically possible)
but with a different output at the expected vout index.

There is risk of funds if clients only later check that the hash for
the earlier-audited tx is found in a block and proceed to send their
counter swap or expose their contract secret via a redemption. This
commit aims to mitigate that risk by repeating contract audits after
the initial tx hash is observed on the blockchain, ensuring that the
tx now observed on the blockchain is as desired.
```

The risk of fund loss I describe above feels less real than when I first thought about it. I've made that change a separate commit so I can easily remove it if the risk identified is not realistic.

chappjc:
> The commit message on [9b51633](https://github.com/decred/dcrdex/commit/9b51633a09c1169e2c40e8c4959b3a3367e5b23f) provides my reasoning for making clients re-audit contracts before acting.
> 
> Copied below for posterity:
> 
> ```
> There is a very slight possibility that contract outputs that pass
> audit while in mempool or in some cases before they are broadcasted
> (rawtx audit) may differ from the output later observed in a block
> for the same coin.
> 
> This risk is higher for spv wallets that will mostly only perform
> audits on rawtxs before broadcasting the txs, without a guarantee
> that the tx is accepted to the mempool. A malicious actor could
> broadcast a different tx with same hash (theoretically possible)
> but with a different output at the expected vout index.
> 
> There is risk of funds if clients only later check that the hash for
> the earlier-audited tx is found in a block and proceed to send their
> counter swap or expose their contract secret via a redemption. This
> commit aims to mitigate that risk by repeating contract audits after
> the initial tx hash is observed on the blockchain, ensuring that the
> tx now observed on the blockchain is as desired.
> ```
> 
> The risk of fund loss I describe above feels less real than when I first thought about it. I've made that change a separate commit so I can easily remove it if the risk identified is not realistic.

Yeah barring hash collisions, I don't think the txns are malleable in any way that is significant for the counterparty, but checking again sounds just fine to me.

itswisdomagain:
I haven't really added any new tests but I will. Code's set for review in the meantime.

itswisdomagain (commented):
Oops.

chappjc (changes requested):
Partial review.  I'm still going through externaltx.go, but I wanted to point out the block header thing.

dex/testing/dcr/harness.sh
```diff
@@ -215,26 +218,30 @@ sleep 3
 ################################################################################
 
 echo "Creating simnet alpha wallet"
+USE_SPV="0"
```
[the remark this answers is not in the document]
itswisdomagain: On the alpha wallet or beta? Can spv wallets vote on tickets? Can't remember.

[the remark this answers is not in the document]
chappjc: Hmm, good point about ticket voting.  I don't believe SPV can vote on tickets.  I tend to use beta when testing anyway.  Let me get confirmation about the voting thing.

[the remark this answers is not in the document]
chappjc: Alright, that SPV an option only for beta then.  It doesn't seem likely that voting will be allowed on SPV wallets because a sensible block validation process isn't gonna be possible short of pulling the blocks.

client/asset/dcr/externaltx.go
```diff
@@ -0,0 +1,444 @@
...
+		blk, err := dcr.getBlock(blockHash, true)
+		if err != nil {
+			return false, err
+		}
+		for i := range blk.RawTx {
+			blkTx := &blk.RawTx[i]
+			if txSpendsOutput(blkTx, output) {
+				dcr.log.Debugf("Found spender for output %s in block %d (%s), spender tx hash %s.",
+					output, blockHeight, blockHash, blkTx.Txid)
+				output.spender.relevantBlock = &block{hash: blockHash, height: blockHeight}
```
chappjc, line 373: It's a little curious to me that `spender` does not actually record the spending txn, just the block that it's in.  Unlike the redemption search, I suppose we only care to know **if it is spent** for (1) audit contract, and (2) `(*ExchangeWallet).Confirmations` method (for `shouldBeginFindRedemption` and `counterPartyConfirms`).  Is that right?

client/asset/dcr/externaltx.go
```diff
@@ -0,0 +1,444 @@
...
+		currentTip.height, currentTip.hash, lastScannedBlock.height, lastScannedBlock.hash)
+	return false, nil
+}
+
+// findExternalTxOutputSpender returns true if a block is found that contains
+// a tx that spends the provided output AND the block is a mainchain block.
+// If no block has been found to contain a tx that spends the provided output
+// or the block that was found to contain the spender is re-orged out of the
+// mainchain, this method matches the output's pkScript against multiple block
+// cfilters (starting from the block containing the output up till the current
```
chappjc, line 308: To determine the purpose of `txBlockHash` at a glance, perhaps a minor doc revision: "...starting from the block containing the output, **txBlockHash**, up to the current..."

And then perhaps something about using `lastScannedBlock` if a previous scan was already performed, up to current...

client/asset/dcr/externaltx.go
```diff
@@ -0,0 +1,444 @@
...
+		for i := range blk.RawSTx {
+			blkTx := &blk.RawSTx[i]
+			if txSpendsOutput(blkTx, output) {
+				dcr.log.Debugf("Found spender for output %s in block %d (%s), spender tx hash %s.",
+					output, blockHeight, blockHash, blkTx.Txid)
+				output.spender.relevantBlock = &block{hash: blockHash, height: blockHeight}
+				return true, nil
+			}
+		}
+		output.spender.lastScannedBlock = blockHash
```
chappjc, line 386: Getting here should be fairly rare according to the GCS parameters, so I think this is worth a log.  If we see this hit and logged often, something is probably wrong.

chappjc (changes requested):
Code mostly makes sense, but there are somethings I think you can help me understand, plus I need to actually fire up dcrwallet with `--spv` to test.

client/asset/dcr/externaltx.go
```diff
@@ -0,0 +1,444 @@
...
+	}
+	mainchainBlockHash, err := nodeGetBlockHashFn(scanner.relevantBlock.height)
+	if err != nil {
+		return nil, fmt.Errorf("cannot get hash for block %d: %v", scanner.relevantBlock.height, err)
+	}
+	if mainchainBlockHash.IsEqual(scanner.relevantBlock.hash) {
+		return scanner.relevantBlock.hash, nil
+	}
+	scanner.relevantBlock = nil // clear so we don't keep checking if it is a mainchain block
+	return nil, nil
```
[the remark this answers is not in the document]
itswisdomagain: Yes, if a blockhash was previously saved but it's no longer a part of the mainchain, then it must have been reorged out. The assumption is that callers of this method were not previously aware of the now-invalid blockhash or if they were, they are rechecking, so they'd know that there is no longer a known blockhash for the tx.

But as with normal node rpc, if the tx was found in a blockhash that is later reorged out of the mainchain, previous callers who were aware of the now-invalid blockhash do not know of the reorg unless they repeat the rpc request. Nothing we can do about "old" callers.

client/asset/dcr/dcr.go
```diff
@@ -1614,67 +1625,254 @@ func (dcr *ExchangeWallet) SignMessage(coin asset.Coin, msg dex.Bytes) (pubkeys,
...
+		SecretHash: secretHash,
+		Recipient:  receiver.String(),
+		Expiration: time.Unix(int64(stamp), 0).UTC(),
+	}, nil
+}
+
+// findContractInBlockchain attempts to locate the specified contract output in
+// the blockchain. If found and unspent, the output is returned along with its
+// pkScript. Returns an error if the output is spent. It is not an error if the
+// tx cannot be located, a nil response and nil error is returned.
```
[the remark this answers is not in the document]
itswisdomagain: Will come back to this.

client/asset/dcr/externaltx.go
```diff
@@ -0,0 +1,444 @@
...
+		stopBlockHash, stopBlock, err := dcr.mainChainAncestor(tx.lastScannedBlock)
+		if err != nil {
+			return false, fmt.Errorf("error looking up mainchain ancestor for block %s", err)
+		}
+		lastScannedBlock.height = stopBlock.Height
+		lastScannedBlock.hash = stopBlockHash
+		tx.lastScannedBlock = stopBlockHash
+	} else {
+		// TODO: Determine a stopHeight to use based on when this tx was first seen
+		// or some constant min block height value.
```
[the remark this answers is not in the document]
itswisdomagain: With #1089 in, we have match time which can be used to determine lowest block height to check.

client/asset/dcr/externaltx.go
itswisdomagain: Right. I'd thought to record the actual spending tx but it doesn't really have any use.

client/asset/dcr/externaltx.go
```diff
@@ -0,0 +1,444 @@
...
+					return false, fmt.Errorf("invalid hex for tx %s: %v", tx.hash, err)
+				}
+
+				tx.relevantBlock = &block{hash: blockHash, height: blockHeight}
+				tx.tree = determineTxTree(msgTx)
+
+				// Store the pkScripts for all of this tx's outputs so they
+				// can be used to later to determine if an output is spent.
+				tx.outputs = make(map[uint32]*externalTxOutput)
+				for i := range blkTx.Vout {
```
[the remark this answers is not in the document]
itswisdomagain: That was my first inclination as well. While those errors are unlikely, it does mean we can't use those outputs to check spending information. A new thought I have now is to simply store the tx.Vout slice and when a caller wants something to do with a particular vout, we can return these errors if they occur.

client/asset/dcr/externaltx.go
```diff
@@ -0,0 +1,444 @@
...
+	if txBlockHash == nil {
+		tx.scanMtx.RUnlock()
+		return nil, 0, false, asset.CoinNotFoundError
+	}
+	output := tx.outputs[vout]
+	txTree := tx.tree
+	tx.scanMtx.RUnlock()
+
+	if output == nil {
+		return nil, 0, false, fmt.Errorf("tx %s has no output at index %d", hash, vout)
```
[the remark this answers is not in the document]
itswisdomagain: As per my previous comment, will use a slice instead and keep all vouts gotten from the block tx. So this error message can remain valid like this. Any other error with amt or pkScript will be returned directly to avoid confusing callers and to aid troubleshooting where necessary.

client/asset/dcr/dcr.go
```diff
@@ -627,19 +634,25 @@ func (dcr *ExchangeWallet) Connect(ctx context.Context) (*sync.WaitGroup, error)
...
 	ver, exists = versions["dcrdjsonrpcapi"]
 	if !exists {
-		return nil, fmt.Errorf("dcrwallet.Version response missing 'dcrdjsonrpcapi'")
-	}
-	nodeSemver := dex.NewSemver(ver.Major, ver.Minor, ver.Patch)
-	if !dex.SemverCompatible(requiredNodeVersion, nodeSemver) {
-		return nil, fmt.Errorf("dcrd has an incompatible JSON-RPC version: got %s, expected %s",
-			nodeSemver, requiredNodeVersion)
-	}
+		dcr.spvMode = true
```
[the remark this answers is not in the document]
chappjc: wallet RPC version bumped to 8.7.0 now, so we can update our require and use that bool

client/asset/dcr/externaltx.go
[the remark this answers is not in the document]
chappjc: OK. I'm fine either way, map or unprocessed slice.  No need to change now unless you see a significant improvement.

client/asset/dcr/dcr.go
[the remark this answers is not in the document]
itswisdomagain: I'm reworking most of these bits with the changes introduced in #1089, but I agree that `findTxOut` is indeed a better method name. Will comment on the flow after making changes to utilize the new features introduced by #1089.

chappjc:
Nothing to be done about this, but SPV harness wallet startup requires a very long (40-60 second delay) before issuing `walletlock` to the beta wallet.  The reason seems to be wallet disconnecting for unknown reasons and not reestablishing a p2p node connection for some time.  This was discussed in matrix. If the wallet is locked before wallet does account discovery, it will fail account discovery because it needs the wallet keys to derive the account keys used during discovery.

The only resolution I see is to move the `setaccountpassphrase` and `walletlock` RPCs for beta down to below the final `./mine-alpha 2` with a `sleep 40` or so preceding it.

We can also remove the comment about `a bug in dcrwallet that breaks validateaddress if the wallet is unlocked` because as we noted recently, this is no longer a bug.

Going through and testing latest changes that use the contract and match time.  I'm optimistic this will be merge-ready in the next few days.

client/asset/dcr/dcr.go
```diff
@@ -158,7 +159,6 @@ var (
 // rpcClient is an rpcclient.Client, or a stub for testing.
 type rpcClient interface {
 	EstimateSmartFee(ctx context.Context, confirmations int64, mode chainjson.EstimateSmartFeeMode) (*chainjson.EstimateSmartFeeResult, error)
```
chappjc, line 161: Fee estimation does not work in SPV mode, so I see `[WRN] CORE[dcr]: Unable to get optimal fee rate, using fallback of 20: -9: RPC passthrough requires dcrd RPC synchronization` when making the reg fee transaction.

```
[WRN] CORE[dcr]: Unable to get optimal fee rate, using fallback of 20: -9: RPC passthrough requires dcrd RPC synchronization
[DBG] CORE[dcr]: Change output size = 36, addr = SsoSsZf84s7Zoy2CVJUDVE2V1o6bJ3A94fy
[DBG] CORE[dcr]: 3 signature cycles to converge on fees for tx eb801ec2e893b8bfb1023a3f3055defc6a7995bc357d075fb95b119fa3542f57: min rate = 20, actual fee rate = 20 (5040 for 252 bytes), change = true
```

In `PayFee`, there's no `feeSuggestion` provided to `feeRateWithFallback`, so it has to use the fallback, not ideal.
One solution would be to use `dc.fetchFeeRate` in `Register` and pass that to a modified `PayFee`.

i.e.

```diff
diff --git a/client/core/core.go b/client/core/core.go
index 5d3c61f9..4f020210 100644
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -2542,7 +2542,8 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
        c.log.Infof("Attempting registration fee payment to %s, account ID %v, of %d units of %s. "+
                "Do NOT manually send funds to this address even if this fails.",
                incompleteReg.regRes.Address, acctID, incompleteReg.regRes.Fee, regFeeAssetSymbol)
-       coin, err := wallet.PayFee(incompleteReg.regRes.Address, incompleteReg.regRes.Fee)
+       feeRateSuggestion := dc.fetchFeeRate(feeAsset.ID)
+       coin, err := wallet.PayFee(incompleteReg.regRes.Address, incompleteReg.regRes.Fee, feeRateSuggestion)
        if err != nil {
                return nil, newError(feeSendErr, "error paying registration fee: %v", err)
        }
```

```diff
diff --git a/client/asset/dcr/dcr.go b/client/asset/dcr/dcr.go
index e11fd7b1..d4acca7d 100644
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -750,10 +750,16 @@ func (dcr *ExchangeWallet) feeRate(confTarget uint64) (uint64, error) {
 // feeRateWithFallback attempts to get the optimal fee rate in atoms / byte via
 // FeeRate. If that fails, it will return the configured fallback fee rate.
 func (dcr *ExchangeWallet) feeRateWithFallback(confTarget, feeSuggestion uint64) uint64 {
-       feeRate, err := dcr.feeRate(confTarget)
-       if err == nil {
-               dcr.log.Tracef("Obtained local estimate for %d-conf fee rate, %d", confTarget, feeRate)
-               return feeRate
+       var err error
+       if !dcr.spvMode {
+               var feeRate uint64
+               feeRate, err = dcr.feeRate(confTarget)
+               if err == nil {
+                       dcr.log.Tracef("Obtained local estimate for %d-conf fee rate, %d", confTarget, feeRate)
+                       return feeRate
+               }
+       } else {
+               err = errors.New("SPV does not support estimatesmartfee")
        }
        if feeSuggestion > 0 && feeSuggestion < dcr.fallbackFeeRate && feeSuggestion < dcr.feeRateLimit {
```

With the above, I then see this:

```
[TRC] CORE[dcr]: feeRateWithFallback using caller's suggestion for 1-conf fee rate, 10. Local estimate unavailable ("SPV does not support estimatesmartfee")
[DBG] CORE[dcr]: Change output size = 36, addr = SsUzFsEy1kG75JhPfmgkGwdouw9Nh4VArVr
[DBG] CORE[dcr]: 2 signature cycles to converge on fees for tx fc465eecb5699ed9d93fd69f377c3736f0c29e48f207ea461d72b295bea9a40d: min rate = 10, actual fee rate = 10 (2520 for 252 bytes), change = true
```

The RegisterResult could be modified to include a suggested rate for the asset.  Also, the regFees part of the config response could also be modified to include a current suggested fee rate, but I don't like that idea for several reasons.  I think fetching the fee rate when needed is preferable.

chappjc: @buck54321 Related to the above comment, I noticed that `(*Core).feeSuggestion` has an unused `isBase bool` input arg.  Was this a leftover, or part of future work plans?

itswisdomagain:
> We can also remove the comment about `a bug in dcrwallet that breaks validateaddress if the wallet is unlocked` because as we noted recently, this is no longer a bug.

Will do so in my next commit.

client/asset/interface.go
```diff
@@ -152,14 +153,24 @@ type Wallet interface {
...
+	// NOTE: For SPV wallets, a successful audit response is no gaurantee that
+	// the txData provided was actually broadcasted to the blockchain. An error
+	// may have occurred while trying to broadcast the txData or even if there
+	// was no broadcast error, the tx might still not enter mempool or get mined
+	// e.g. if the tx references invalid or already spent inputs.
+	//
+	// Granted, clients wait for the contract tx to be included in a block before
+	// taking further actions on a match; but it is generally safer to repeat this
+	// audit after the contract tx is mined to ensure that the tx observed on the
+	// blockchain is as expected.
```
buck54321, line 173: This is a red flag for me that says we should change the interface to separate audit of server-supplied transaction data from the confirmation of the tx on the blockchain. Both wallets can work fine with that model, and we wouldn't need notes like that basically just say that return values may mean different things depending on implementation-specific details.

client/core/trade.go
```diff
@@ -2136,6 +2158,70 @@ func (t *trackedTrade) auditContract(match *matchTracker, coinID, contract, txDa
...
+	if match.Side == order.Maker && match.Status != order.TakerSwapCast {
+		return fmt.Errorf("Invalid repeat audit request for maker at status %s.", match.Status)
+	}
+	if match.Side == order.Taker && match.Status != order.MakerSwapCast {
+		return fmt.Errorf("Invalid repeat audit request for taker at status %s.", match.Status)
+	}
+
+	proof := &match.MetaData.Proof
+	coinID, contract := match.counterSwap.Coin.ID(), proof.CounterContract
+	auditInfo, err := t.wallets.toWallet.AuditContract(coinID, contract, nil, time.Time{})
```
buck54321, line 2207: Does `matchTime` not matter here because we know the `externalTx` is already cached? Shouldn't we just pass the actual `match.matchTime()` here anyway?

client/core/trade.go
```diff
@@ -893,7 +893,18 @@ func (t *trackedTrade) isSwappable(ctx context.Context, match *matchTracker) boo
...
+				return false
+			}
+			// Re-audit the contract now that we know it is mined
+			// and has the required confirmations.
+			if err := t.reAuditContract(match); err != nil {
+				t.dc.log.Errorf("Match %s not swappable: repeat audit of maker's contract failed: %v",
+					match, err)
+				match.swapErr = fmt.Errorf("Counter-party contract repeat audit error: %v", err)
+				return false
+			}
```
buck54321, line 906: I really don't think we need to do this. I understand the compulsion, but unless we're willing to say that the attacker can spoof a transaction hash, this doesn't really do anything for us. And the way I understand it, spoofing a transaction hash is realistically impossible. 

The fact that we're using `AuditContract` to also find the transaction on-chain causes a lot of ambiguity around the roles these methods perform, and results in extra work. We could have a cleaner, more user-friendly interface if we just define `AuditContract` to audit the server-supplied tx data, and `SwapConfirmation` to deal with the blockchain.

buck54321 (commented):
First pass. I think there's some simplification to be done around `externalTx`, but I don't have specific suggestions for you at this time. Most of these comments deal with the `AuditContract` and `SwapConfirmations` sequence. I really hope we can agree to simplify the role of `AuditContract`.

client/asset/dcr/dcr.go
```diff
@@ -612,10 +617,16 @@ func (dcr *ExchangeWallet) feeRate(confTarget uint64) (uint64, error) {
...
+	var err error
+	if !dcr.wallet.SpvMode() {
+		var feeRate uint64
+		feeRate, err = dcr.feeRate(confTarget)
+		if err == nil {
+			dcr.log.Tracef("Obtained local estimate for %d-conf fee rate, %d", confTarget, feeRate)
+			return feeRate
+		}
+	} else {
+		err = errors.New("SPV does not support estimatesmartfee")
```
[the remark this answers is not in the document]
itswisdomagain: Right. I suspect @chappjc wanted a more direct error message for logging purposes, but can always use the error returned by the wallet type.

client/asset/dcr/dcr.go
```diff
@@ -1483,86 +1496,382 @@ func (dcr *ExchangeWallet) SignMessage(coin asset.Coin, msg dex.Bytes) (pubkeys,
...
+// tx output. Returns asset.CoinNotFoundError if the output is not found.
+// NOTE: SPV wallets are only able to lookup outputs for transactions that are
+// tracked by the wallet.
+func (dcr *ExchangeWallet) checkWalletForTxOutput(ctx context.Context, op outPoint, mempool bool) (*txOutput, outputSpendStatus, error) {
+	errorOut := func(err error) (*txOutput, outputSpendStatus, error) {
+		return nil, outputSpendStatusUnknown, err
+	}
+
+	// First use wallet.GetTxOut to look up the output.
+	unspentTxOut, txTree, err := dcr.wallet.GetTxOut(ctx, &op.txHash, op.vout, wire.TxTreeUnknown, mempool)
```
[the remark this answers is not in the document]
itswisdomagain: Thought I renamed `GetTxOut` to `UnspentOutput` since it really only returns tx outputs that are unspent (hence the references to UnspentOutput you noted in some docs). I will complete that renaming to make things less confusing, since `gettxout` is quite really an rpc-only concept for fetching unspent outputs.

`checkWalletForTxOutput` does check wallet transactions but first calls the `wallet.GetTxOut` method to attempt finding the (unspent) output _using the wallet_. `GetTxOut` is a method the wallet provides for looking up unspent outputs. The difference is that full node wallets are able to consult the connected full node to find unspent outputs that may not be related to the wallet, while spv wallets only check unspent wallet outputs.

client/asset/dcr/dcr.go
```diff
@@ -2845,6 +3137,63 @@ func (dcr *ExchangeWallet) getBestBlock(ctx context.Context) (*block, error) {
...
+			return nil, 0, fmt.Errorf("error decoding previous hash %s for block %s: %w",
+				checkBlock.PreviousHash, checkHash.String(), err)
+		}
+	}
+}
+
+func (dcr *ExchangeWallet) isMainchainBlock(ctx context.Context, block *block) (bool, error) {
+	if block == nil {
+		return false, nil
+	}
```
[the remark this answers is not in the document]
itswisdomagain: I wasn't checking at the call site initially, will remove.

client/asset/dcr/dcr.go
```diff
@@ -1483,86 +1496,382 @@ func (dcr *ExchangeWallet) SignMessage(coin asset.Coin, msg dex.Bytes) (pubkeys,
...
+			// TODO: It'd be unnecessary to continue trying to find this
+			// contract or to broadcast the rawtx if the tx was already
+			// broadcasted, mined AND spent.
+			// Consider modifying dcr.externalTxOut to use block filters to
+			// check if the tx exists on the blockchain if gettxout returns
+			// a nil response. If the tx does exist, return CoinIsSpentError
+			// to the caller.
+
+			return nil, asset.CoinNotFoundError
 		}
```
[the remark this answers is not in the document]
itswisdomagain: Broadcast errors in full node wallets typically mean the tx already exist. If that is the case, and gettxout did not find it earlier in this method, doesn't make much sense to return a successful audit response, unless as you suggest, we make this method only concern itself with auditing the txData. Then we can assume that a broadcast error in full node mode means the tx already exists on the blockchain.

client/asset/interface.go
itswisdomagain: That can work, it's really just the same question of whether we want to audit the tx seen on the blockchain or just the txData alone. Confirming the tx on the blockchain doesn't audit it and we already have a separate method for that, so this could just simplify to auditing and broadcasting the txData. If we do want to audit the contract found on the blockchain, we could have `AuditContractData` and `AuditBlockchainContract` to simplify this method. Or do we just do away with repeat contract audits?

client/core/trade.go
itswisdomagain: 🤔

client/core/trade.go
itswisdomagain: Works for me. @chappjc?

itswisdomagain:
> So an SPV wallet calling `lookupTxOutput` from `AuditContract` will **always** get `CoinNotFoundError` back, correct?

Not really. If `AuditContract` is called after `SwapConfirmations` finds the contract onchain, the output data cached from the blockchain will be returned.

itswisdomagain:
I've got a refactor coming for the `lookupTxOutput` and `checkWalletForTxOutput` methods, primarily to make the flow simpler for `AuditContract` and `refundTx` callers that do not perform filters check for the output.

client/core/trade.go
chappjc: Yes, agree.

client/asset/dcr/dcr.go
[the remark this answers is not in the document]
chappjc: I had the same though as buck here.  It's fine to remove the condition if it is effectively handled the same way.

client/asset/dcr/dcr.go
```diff
@@ -1484,65 +1490,53 @@ func (dcr *ExchangeWallet) SignMessage(coin asset.Coin, msg dex.Bytes) (pubkeys,
...
 	if err != nil {
 		return nil, fmt.Errorf("error extracting swap addresses: %w", err)
 	}
-	// Get the contracts P2SH address from the tx output's pubkey script.
-	txOutRes, txTree, err := dcr.getTxOut(txHash, vout, true)
-	if err != nil {
-		return nil, fmt.Errorf("error finding unspent contract: %w", err)
+
+	contractTx := wire.NewMsgTx()
+	if err := contractTx.Deserialize(bytes.NewReader(txData)); err != nil {
```
chappjc, line 1511: nbd, but we have a `msgTxFromBytes` that looks just like these lines

client/asset/dcr/dcr.go
```diff
@@ -2208,36 +2287,65 @@ func (dcr *ExchangeWallet) ValidateSecret(secret, secretHash []byte) bool {
...
+	_, p2shScript := scriptAddr.PaymentScript()
+
+	// Find the contract and it's spend status using block filters.
+	dcr.log.Debugf("Contract output %s:%d NOT yet found, will attempt finding it with block filters.", txHash, vout)
+	confs, spent, err = dcr.lookupTxOutWithBlockFilters(ctx, newOutPoint(txHash, vout), p2shScript, matchTime)
 	if err != nil {
 		return 0, false, err
 	}
-	return uint32(tx.Confirmations), true, nil
+	return confs, spent, err
```
chappjc, line 2325: `return dcr.lookupTxOutWithBlockFilters(...` is good given how you've written that method.

client/asset/dcr/dcr.go
```diff
@@ -2208,36 +2287,65 @@ func (dcr *ExchangeWallet) ValidateSecret(secret, secretHash []byte) bool {
...
 }
 
-// SwapConfirmations gets the number of confirmations for the specified coin ID
-// by first checking for a unspent output, and if not found, searching indexed
-// wallet transactions.
-func (dcr *ExchangeWallet) SwapConfirmations(ctx context.Context, coinID dex.Bytes, _ dex.Bytes, _ time.Time) (confs uint32, spent bool, err error) {
-	return dcr.coinConfirmations(ctx, coinID)
+func (dcr *ExchangeWallet) confirms(blockHeight int64) uint32 {
```
chappjc, line 2328: `confirms` is only used by `lookupTxOutWithBlockFilters`, which is also in a different file (externaltx.go).  It's also a bit specialized in that it converts to uint32 and falls back to the cached best block.

So that suggests to me this might be better off in `lookupTxOutWithBlockFilters` where it can prevent a negative `int64` being cast into a huge `uint32`.  The ctx arg can be used instead that way too.

client/asset/dcr/dcr.go
```diff
@@ -2846,6 +2942,60 @@ func (dcr *ExchangeWallet) getBestBlock(ctx context.Context) (*block, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("error retrieving block %s: %w", blockHash, translateRPCCancelErr(err))
+	}
+	return blockVerbose, nil
+}
+
+func (dcr *ExchangeWallet) getBlockHash(blockHeight int64) (*chainhash.Hash, error) {
+	blockHash, err := dcr.wallet.GetBlockHash(dcr.ctx, blockHeight)
+	if err != nil {
+		return nil, translateRPCCancelErr(err)
```
chappjc, line 2956: The `dcr.Wallet` interface implementations can be responsible for translating into an `asset.ErrRequestTimeout`.  Looks like the `rpcWallet` methods are doing it already.

As a bonus, this will reduce some of these to oneliners.

client/asset/dcr/dcr.go
```diff
@@ -2846,6 +2942,60 @@ func (dcr *ExchangeWallet) getBestBlock(ctx context.Context) (*block, error) {
 	return &block{hash: hash, height: height}, nil
 }
 
+func (dcr *ExchangeWallet) getBlock(ctx context.Context, blockHash *chainhash.Hash, verboseTx bool) (*chainjson.GetBlockVerboseResult, error) {
+	blockVerbose, err := dcr.wallet.GetBlockVerbose(ctx, blockHash, verboseTx)
+	if err != nil {
+		return nil, fmt.Errorf("error retrieving block %s: %w", blockHash, translateRPCCancelErr(err))
```
chappjc, line 2948: The caller knows the block hash they requested, and the non-nil error tells them all they need to know.  Together with my comment about `translateRPCCancelErr` being done in the `dcr.Wallet` implementer, this will become a oneliner.

client/asset/dcr/externaltx.go
```diff
@@ -0,0 +1,475 @@
...
+
+// txSpendsOutput returns true if the passed tx has an input that spends the
+// specified output.
+func txSpendsOutput(tx *chainjson.TxRawResult, op outPoint) bool {
+	if tx.Txid == op.txHash.String() {
+		return false // no need to check inputs if this tx is the same tx that pays to the specified op
+	}
+	for i := range tx.Vin {
+		input := &tx.Vin[i]
+		if input.Vout == op.vout && input.Txid == op.txHash.String() {
```
chappjc, line 431: `String` will be called multiple times, starting with the initial `tx.Txid` check, so might as well make a local var on the first line of the func.

client/asset/dcr/externaltx.go
```diff
@@ -0,0 +1,475 @@
...
+	}
+
+	// Scan block filters to find the tx block if it is yet unknown.
+	if txBlock == nil {
+		txBlock, err = dcr.scanFiltersForTxBlock(ctx, tx, [][]byte{pkScript}, earliestTxTime)
+		if err != nil {
+			return nil, nil, fmt.Errorf("error checking if tx %s is mined: %v", tx.hash, err)
+		}
+	}
+	if txBlock == nil {
```
chappjc, line 101: Would prefer this in the previous condition:

```go
		return nil, nil, fmt.Errorf("error checking if tx %s is mined: %v", tx.hash, err)
	}
	if txBlock == nil {
		return nil, nil, asset.CoinNotFoundError
	}
}
```

client/asset/dcr/dcr.go
```diff
@@ -2846,6 +2942,60 @@ func (dcr *ExchangeWallet) getBestBlock(ctx context.Context) (*block, error) {
...
+		}
+	}
+}
+
+func (dcr *ExchangeWallet) isMainchainBlock(ctx context.Context, block *block) (bool, error) {
+	blockHeader, err := dcr.wallet.GetBlockHeaderVerbose(ctx, block.hash)
+	if err != nil {
+		return false, fmt.Errorf("getblockheader error for block %s: %w", block.hash, translateRPCCancelErr(err))
+	}
+	return blockHeader.Confirmations > -1 && int64(blockHeader.Height) == block.height, nil
```
chappjc, line 2990: Just noting that now that we don't just ask dcrd via dcrwallet for everything, we're also taking on the burden of accounting for stake invalidation too.  "invalidated" in callers so far is just being moved to a side chain.  Lets move ahead like you have, but we would really need to look at the next block's votebits to see if this one has had its regular tx tree invalidated.  :(

chappjc (changes requested):
Things have become much more easily reasoned about in these last few commits.  Liking it!

Some thoughts in this partial review.

client/asset/dcr/dcr.go
```diff
@@ -2361,14 +2456,15 @@ func (dcr *ExchangeWallet) convertCoin(coin asset.Coin) (*output, error) {
...
+	// TODO: We just need the tree.
+	txOut, err := dcr.wallet.UnspentOutput(dcr.ctx, txHash, vout, wire.TxTreeUnknown)
 	if err != nil {
 		return nil, fmt.Errorf("error finding unspent output %s:%d: %w", txHash, vout, err)
 	}
 	if txOut == nil {
 		return nil, asset.CoinNotFoundError // maybe spent
 	}
-	return newOutput(txHash, vout, coin.Value(), tree), nil
+	return newOutput(txHash, vout, coin.Value(), txOut.Tree), nil
```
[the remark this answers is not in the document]
itswisdomagain: Yeah, I think `convertCoin` can do without rpc, so callers always know the tree is unknown and should deal with that if they need to.

client/asset/dcr/dcr.go
```diff
@@ -1554,16 +1548,123 @@ func (dcr *ExchangeWallet) AuditContract(coinID, contract, txData dex.Bytes, _ t
...
+	return wire.TxTreeRegular
+}
+
+// lookupTxOutput attempts to find and return details for the specified output,
+// first checking for an unspent output and if not found, checking wallet txs.
+// Returns asset.CoinNotFoundError if the output is not found.
+// NOTE: This method is only guaranteed to return results for outputs belonging
+// to transactions that are tracked by the wallet, although full node wallets
+// are able to look up non-wallet outputs that are unspent.
+func (dcr *ExchangeWallet) lookupTxOutput(ctx context.Context, txHash *chainhash.Hash, vout uint32) (*wire.TxOut, uint32, int8, bool, error) {
```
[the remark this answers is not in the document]
itswisdomagain: Ah, crap. Was used at one point, ah yes, when `AuditContract` called this method. Will remove.

client/asset/dcr/dcr.go
itswisdomagain: SMH. Should have been obvious, `lookupTxOutWithBlockFilters` had a different return signature once upon a time I suspect.

client/asset/dcr/dcr.go
itswisdomagain: Hmm. Should be a quick fix. Will add.

client/asset/btc/btc.go
```diff
@@ -1824,9 +1824,10 @@ func (btc *ExchangeWallet) SignMessage(coin asset.Coin, msg dex.Bytes) (pubkeys,
 	return
 }
 
-// AuditContract retrieves information about a swap contract on the blockchain.
-// AuditContract would be used to audit the counter-party's contract during a
-// swap.
+// AuditContract retrieves information about a swap contract from the provided
+// txData. The extracted information would be used to audit the counter-party's
+// contract during a swap.
+// TODO: Broadcast the txData.
```
chappjc, line 1830: How about we do this now with the following right before `return`:

```go
hashSent, err := btc.node.sendRawTransaction(tx)
if err != nil {
      btc.log.Debugf("Rebroadcasting counterparty contract %v (THIS MAY BE NORMAL): %v", txHash, err)
} else if !hashSent.IsEqual(txHash) {
       btc.log.Errorf("Counterparty contract %v was rebroadcast as %v!", txHash, hashSent)
}
```

client/asset/dcr/externaltx.go
```diff
@@ -0,0 +1,489 @@
...
+	}
+
+	// Cache relevant spender info if the spender is found.
+	spent := spenderTx != nil
+	if spent {
+		spenderBlockHash, err := chainhash.NewHashFromStr(spenderTx.BlockHash)
+		if err != nil {
+			// dcr.log.Errorf("Invalid hash (%s) for tx that spends output %s: %v", spenderTx.BlockHash, output.op, err)
+			return false, fmt.Errorf("invalid hash (%s) for tx that spends output %s: %v", spenderTx.BlockHash, output.op, err)
+		} else {
```
chappjc, line 372: If you need to do another commit, please remove the else block and commented log.

itswisdomagain: That's the intent, yes. Your version is clearer, will just adopt it.

buck54321 (approved):
Thanks for the follow up. This seems solid now.

I'll be following up quickly to propose some small changes to the `Wallet` interface.

chappjc (approved):

PR #789, opened 2020-10-27
server/book: Clear uses correct heap orientation

chappjc:
The book's Clear function used when purging the book during a suspend operation works by allocating new priority queues.  This fixes the heap orientation of the sell side, and reduced the default order book capacity to a more sensible level.

### DIFF
--- a/server/book/book.go
+++ b/server/book/book.go
@@ -13,9 +13,9 @@ import (
 
 const (
 	// DefaultBookHalfCapacity is the default capacity of one side (buy or sell)
-	// of the order book. It is set to 2^21 orders (2 mebiorders = 2,097,152
+	// of the order book. It is set to 2^20 orders (1 mebiorders = 1,048,576
 	// orders) per book side.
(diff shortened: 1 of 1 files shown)


server/book/book.go
```diff
@@ -50,8 +50,9 @@ func New(lotSize uint64, halfCapacity ...uint32) *Book {
 // Clear reset the order book with configured capacity.
 func (b *Book) Clear() {
 	b.mtx.Lock()
+	b.buys, b.sells = nil, nil
 	b.buys = NewMaxOrderPQ(b.halfCap)
-	b.sells = NewMaxOrderPQ(b.halfCap)
+	b.sells = NewMinOrderPQ(b.halfCap)
```
chappjc, line 55: This function only happens on book purge with market suspend, so if the op resumed without restarting the whole thing, it would be backwards on the sell side.

The reallocation was seen to press system memory since the PQ is the bulk of the allocated memory for dcrdex, so I cut the default book capacity in half and nil'd both queue variables before allocating new ones in case that helps Go GC through it.

buck54321 (approved):

PR #796, opened 2020-10-29
multi: add market data API endpoints

buck54321:
Exposes a rate-limited HTTP API. Data endpoints are accessible via HTTP or WebSockets. The result for the WebSocket request is identical to the REST API response.

/spots is the spot price and booked volume of all markets.

/candles is candlestick data, available in bin sizes of 24h, 1h, 15m, and per-epoch sticks. An example URL is /candles/dcr/btc/15m.

/orderbook is already a WebSocket route, but is now also accessible by HTTP. An example URL is /orderbook/dcr/btc

/config is another WebSocket route that is also now available over HTTP too.

The data API implements a data cache, but does not cache pre-encoded responses for /candles or /orderbook (yet).

**server/db**

Market history is stored as a row per epoch. I've created a new table for this, though we could also combine with the existing epoch table that holds match proof info if preferred. 

I've made no attempt to integrate pre-existing data. Data will be collected from the time of upgrade only. The upgrade is silent, since we don't have upgrade infrastructure in place yet for the server. 

**server/comms**

The config route is now registered as an http handler, but also called from the websocket message handler.

HTTP requests are rate-limited using `"golang.org/x/time/rate"`. There is both a global rate limiter and a per-IP rate limiter. The rate limits apply to un-authenticated routes only. The /config request is has an exception as a "critical' route.

___

Try these with the simnet harness running...

`wget -qO- --no-check-certificate  'https://127.0.0.1:17232/api/candles/dcr/btc/5m'`
`wget -qO- --no-check-certificate  'https://127.0.0.1:17232/api/spots'`
`wget -qO- --no-check-certificate  'https://127.0.0.1:17232/api/config'`
`wget -qO- --no-check-certificate  'https://127.0.0.1:17232/api/orderbook/dcr/btc'`


chappjc:
If there's connectivity trouble with the server, and all clients have to reconnect, each client will hit /config on reconnect.  We'll have to ensure that doesn't dos ourselves.

I've been very nervous about the implications of this addition, but I'd feel a lot better if there was a way to switch off the data api with an admin command and a conf setting.  Does that seem like it can be done easily?

server/comms/server.go
```diff
@@ -452,3 +536,40 @@ func parseListeners(addrs []string) ([]string, []string, bool, error) {
...
+	}
+	resp, err := handler(thing)
+	if err != nil {
+		writeJSONWithStatus(w, map[string]string{"error": err.Error()}, http.StatusBadRequest)
+		return
+	}
+	writeJSONWithStatus(w, resp, http.StatusOK)
+}
+
+// routeHandler creeates a HandlerFunc for a route. Middleware should have
```
JoeGruffins, line 557: creeates

server/comms/server.go
```diff
@@ -452,3 +536,40 @@ func parseListeners(addrs []string) ([]string, []string, bool, error) {
...
+	}
+}
+
+// writeJSONWithStatus writes the JSON response with the specified HTTP response
+// code.
+func writeJSONWithStatus(w http.ResponseWriter, thing interface{}, code int) {
+	w.Header().Set("Content-Type", "application/json; charset=utf-8")
+	w.WriteHeader(code)
+	encoder := json.NewEncoder(w)
+	// encoder.SetIndent("", indent)
```
JoeGruffins, line 571: Comment can be removed?

JoeGruffins (approved):
Srry no substantial comments. Code looks great to me. Am able to hit all the endpoints with curl, no problem.

server/comms/middleware.go
```diff
@@ -0,0 +1,121 @@
...
+// These are the keys for different types of values stored in a request context.
+const (
+	ctxThing contextKey = iota
+)
+
+// limitRate is rate-limiting middleware that checks whether a request can be
+// fulfilled.
+func (s *Server) limitRate(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		code, err := s.meterIP(r.RemoteAddr)
```
chappjc, line 30: When I had played with per IP limits in https://github.com/chappjc/dcrdex/commit/9607307e806275be77ddc313ae956cad439f823a#diff-e7126c606b49bf8ca255171ae2f485e8d92f442af17e953700f49bcea7215507R27, I realized that dealing with IPv6 was a bit trickier because it's really the prefix that you need to consider since the device address is trivially adjusted by the client. https://www.ibm.com/support/knowledgecenter/en/STCMML8/com.ibm.storage.ts3500.doc/opg_3584_IPv4_IPv6_prefix_subnet_mask.

The code I landed on to take an IP address string like RemoteAddr and give a key that masks out the the non-prefix bits of IPv6:

```go
type IP [net.IPv6len]byte

func parseIP(addr string) IP {
	ip := net.ParseIP(addr)
	if ip.To4() == nil {
		ip = ip.Mask(net.CIDRMask(64, 128))
	}
	var ipKey IP
	copy(ipKey[:], ip)
	return ipKey
}
```

server/comms/server.go
```diff
@@ -452,3 +553,39 @@ func parseListeners(addrs []string) ([]string, []string, bool, error) {
...
+	}
+}
+
+// writeJSONWithStatus writes the JSON response with the specified HTTP response
+// code.
+func writeJSONWithStatus(w http.ResponseWriter, thing interface{}, code int) {
+	w.Header().Set("Content-Type", "application/json; charset=utf-8")
+	w.WriteHeader(code)
+	encoder := json.NewEncoder(w)
+	if err := encoder.Encode(thing); err != nil {
```
chappjc, line 588: An issue I ran into with this that if their is a json marshalling error, it still sends junk to the client.  https://github.com/decred/dcrdex/commit/ae3b4ff45adcd9614aa175caf0bb1acc825af672
If we are positive `thing` will always marshal without error, it's OK though.  Plus we're following the NewEncoder pattern in all the other http handlers and packages, so maybe we stay consistent with them for now.

server/comms/server.go
```diff
@@ -452,3 +553,39 @@ func parseListeners(addrs []string) ([]string, []string, bool, error) {
...
+		return
+	}
+	writeJSONWithStatus(w, resp, http.StatusOK)
+}
+
+// routeHandler creates a HandlerFunc for a route. Middleware should have
+// already processed the request and added the request struct to the Context.
+func routeHandler(route string) func(w http.ResponseWriter, r *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		handleHTTP(w, r, route, r.Context().Value(ctxThing))
```
chappjc, line 578: It's unfortunate that each request will do a lookup in the `httpRoutes` map.  I was wondering if this could be:

```go
// routeHandler creates a HandlerFunc for a route. Middleware should have
// already processed the request and added the request struct to the Context.
func routeHandler(route string) func(w http.ResponseWriter, r *http.Request) {
	handler := httpRoutes[route]
	if handler == nil {
		panic("no known handler")
	}
	return func(w http.ResponseWriter, r *http.Request) {
		thing := r.Context().Value(ctxThing)
		resp, err := handler(thing)
		if err != nil {
			writeJSONWithStatus(w, map[string]string{"error": err.Error()}, http.StatusBadRequest)
			return
		}
		writeJSONWithStatus(w, resp, http.StatusOK)
	}
}
```

Or might all the http routes not be registered when comms.Server starts?

server/apidata/apidata.go
```diff
@@ -0,0 +1,238 @@
...
+	"decred.org/dcrdex/server/matcher"
+)
+
+const (
+	// DefaultCandleRequest is the number of candles to return if the request
+	// does not specify otherwise.
+	DefaultCandleRequest = 50
+	// CacheSize is the default cache size. Also represents the maximum number
+	// of candles that can be requested at once.
+	CacheSize = 1000
```
chappjc, line 28: Can unexport?

server/apidata/apidata.go
```diff
@@ -0,0 +1,238 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Add the candlestick.
+	s.cacheMtx.Lock()
+	mktCaches := s.marketCaches[mktName]
+	if mktCaches == nil {
+		s.cacheMtx.Unlock()
+		return fmt.Errorf("unkown market %q", mktName)
```
chappjc, line 132: unkown

chappjc (changes requested):
I'm still thinking about how the  `apidata` package integrates with  Market and BookRouter, but here are some minor comments.

server/apidata/apidata.go
buck54321: I thought it might come in handy for apps like dcrdata to keep a cache from the websocket feed on the client side.

server/comms/server.go
buck54321: They actually weren't, but that was breaking my own rules from the docs of `(*Server).Run`, which said

> `Run should be called only after all routes are registered`

 so I've corrected the sequence in **server/dex** and used your suggestion here.

server/apidata/apidata.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+// SetBookSource should be called before the first call to handleBook.
+func (s *DataAPI) SetBookSource(bs BookSource) {
+	s.bookSource = bs
+}
+
+// ReportEpoch should be called by every Market after every match cycle to
+// report their epoch stats.
+func (s *DataAPI) ReportEpoch(base, quote uint32, epochIdx uint64, stats *matcher.MatchCycleStats) (*msgjson.Spot, error) {
```
buck54321, line 129: I've added the `Spot` return value here because I'd like clients to be able to subscribe to a feed of `Spots`. This would allow us to display little green arrows and percent changes next to the market names in the UI.

server/db/driver/pg/internal/epochs.go
```diff
@@ -14,6 +14,35 @@ const (
...
+
+	// CreateEpochReportTable creates an epoch_reports table that holds
+	// epoch-end reports that can be used to contruct market history data sets.
+	CreateEpochReportTable = `CREATE TABLE IF NOT EXISTS %s (
+		epoch_end INT8 PRIMARY KEY, -- using timestamp instead of index to facilitate sorting and filtering with less math
+		epoch_dur INT4,             -- epoch duration in milliseconds
+		match_volume INT8,          -- total matched during epoch's match cycle
+		quote_volume INT8,          -- total matched volume in terms of quote asset
+		book_volume INT8,           -- book volume after matching
+		order_volume INT8,          -- epoch order volume
```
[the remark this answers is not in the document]
buck54321: Dropped `order_volume` in favor of a few data points that make up a rough depth profile.

server/db/driver/pg/internal/epochs.go
```diff
@@ -14,6 +14,35 @@ const (
...
 		VALUES ($1, $2, $3, $4, $5, $6, $7);`
+
+	// CreateEpochReportTable creates an epoch_reports table that holds
+	// epoch-end reports that can be used to contruct market history data sets.
+	CreateEpochReportTable = `CREATE TABLE IF NOT EXISTS %s (
+		epoch_end INT8 PRIMARY KEY, -- using timestamp instead of index to facilitate sorting and filtering with less math
+		epoch_dur INT4,             -- epoch duration in milliseconds
+		match_volume INT8,          -- total matched during epoch's match cycle
+		quote_volume INT8,          -- total matched volume in terms of quote asset
+		book_volume INT8,           -- book volume after matching
```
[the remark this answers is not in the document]
buck54321: Dropped `book_volume` in favor of a few data points that make up a rough depth profile.

server/db/driver/pg/internal/epochs.go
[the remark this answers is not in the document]
buck54321: Dropped `order_volume`. Not capturing unmatched order quantities at all now.

chappjc (changes requested):

client/orderbook/orderbook.go
```diff
@@ -344,6 +344,12 @@ func (ob *OrderBook) UpdateRemaining(note *msgjson.UpdateRemainingNote) error {
 	return ob.updateRemaining(note, false)
 }
 
+// LogEpochReport just checks the notification sequence.
+func (ob *OrderBook) LogEpochReport(note *msgjson.EpochReportNote) error {
+	ob.setSeq(note.Seq)
```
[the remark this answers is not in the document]
chappjc: Oh, please disregard this seq comment.  I was conflating the epoch order notes note with this epoch close timing.

server/admin/api.go
```diff
@@ -38,11 +38,16 @@ func writeJSON(w http.ResponseWriter, thing interface{}) {
...
-	encoder.SetIndent("", "    ")
-	if err := encoder.Encode(thing); err != nil {
+	b, err := json.MarshalIndent(thing, "", "    ")
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
 		log.Errorf("JSON encode error: %v", err)
+		return
+	}
+	w.WriteHeader(code)
+	_, err = w.Write(append(b, byte('\n')))
```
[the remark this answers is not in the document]
buck54321: Hmmm. I thought it was functional, but it doesn't seem necessary. I must have got myself twisted up trying to understand the [behavior for `(*json.Encoder).Encode` vs `json.Marshal`](https://github.com/golang/go/issues/37083).

chappjc (approved):

Issue #780, opened 2020-10-24
taker redemptions can't leverage grouping


buck54321:
The maker redeems based on block confirmations, so matches from the same epoch have a good chance of being grouped into a single transaction because they will often receive their swapconf'th confirmation with the same block. 

The taker, on the other hand, typically redeems based on a notification from the counter-party, so will never be able to group redemptions. But if `handleRedemptionRoute` did not call `tick` immediately, and instead just scheduled a tick or something, then we could group the redemption and save some coin. I'm not certain if there are situations where a small delay would be unsafe.

PR #797, opened 2020-10-29
client/core: batch trade ticks for audit and redemption reqs

chappjc:
Resolves https://github.com/decred/dcrdex/issues/780

This delays a trade's tick that is usually launched by the redemption (and audit) requests to allow for multiple redemptions to be batched more effectively. Specifically, the handlers now process the message and update the trade and match fields so that when tick happens after a delay, there can be multiple redeems provided to `redeemMatches` instead of one at a time several times.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -821,6 +821,9 @@ type Core struct {
 	waiterMtx    sync.Mutex
 	blockWaiters map[uint64]*blockWaiter
 
+	tickSchedMtx sync.Mutex
+	tickSched    map[order.OrderID]*time.Timer
+
 	userMtx sync.RWMutex
 	user    *User
 
@@ -857,6 +860,7 @@ func New(cfg *Config) (*Core, error) {
 		lockTimeMaker: dex.LockTimeMaker(cfg.Net),
 		blockWaiters:  make(map[uint64]*blockWaiter),
 		piSyncers:     make(map[order.OrderID]chan struct{}),
+		tickSched:     make(map[order.OrderID]*time.Timer),
 		// Allowing to change the constructor makes testing a lot easier.
 		wsConstructor: comms.NewWsConn,
 		newCrypter:    encrypt.NewCrypter,
@@ -4056,6 +4060,45 @@ func handleNoMatchRoute(c *Core, dc *dexConnection, msg *msgjson.Message) error 
 	return err
 }
 
+func (c *Core) schedTradeTick(tracker *trackedTrade) {
+	oid := tracker.ID()
+	c.tickSchedMtx.Lock()
+	defer c.tickSchedMtx.Unlock()
+	if _, found := c.tickSched[oid]; found {
+		return // already going to tick this trade
+	}
+
+	tick := func() {
+		assets, err := c.tick(tracker)
+		if len(assets) > 0 {
+			tracker.dc.refreshMarkets()
+			c.updateBalances(assets)
+		}
+		if err != nil {
+			c.log.Errorf("tick error for order %v: %v", oid, err)
+		}
+	}
+
+	numMatches := len(tracker.activeMatches())
+	if numMatches == 1 {
+		go tick()
+		return
+	}
+
+	// Schedule a tick for this trade.
+	delay := time.Duration(numMatches) * time.Second / 10 // 1 sec delay for every 10 active matches
+	if delay > 5*time.Second {
+		delay = 5 * time.Second
+	}
+	c.log.Debugf("Waiting %v to tick trade %v with %d active matches", delay, oid, numMatches)
+	c.tickSched[oid] = time.AfterFunc(delay, func() {
+		c.tickSchedMtx.Lock()
+		defer c.tickSchedMtx.Unlock()
+		defer delete(c.tickSched, oid)
+		tick()
+	})
+}
+
 // handleAuditRoute handles the DEX-originating audit request, which is sent
 // when a match counter-party reports their initiation transaction.
 func handleAuditRoute(c *Core, dc *dexConnection, msg *msgjson.Message) error {
@@ -4075,12 +4118,9 @@ func handleAuditRoute(c *Core, dc *dexConnection, msg *msgjson.Message) error {
 	if err != nil {
 		return err
 	}
-	assets, err := c.tick(tracker)
-	if len(assets) > 0 {
-		dc.refreshMarkets()
-		c.updateBalances(assets)
-	}
-	return err
+	// Are we ticking this trade after an audit in case it's mined already?
+	c.schedTradeTick(tracker)
+	return nil
 }
 
 // handleRedemptionRoute handles the DEX-originating redemption request, which
@@ -4102,12 +4142,8 @@ func handleRedemptionRoute(c *Core, dc *dexConnection, msg *msgjson.Message) err
 	if err != nil {
 		return err
 	}
-	assets, err := c.tick(tracker)
-	if len(assets) > 0 {
-		dc.refreshMarkets()
-		c.updateBalances(assets)
-	}
-	return err
+	c.schedTradeTick(tracker)
+	return nil
 }
 
 // removeWaiter removes a blockWaiter from the map.
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -4086,7 +4086,7 @@ func (c *Core) schedTradeTick(tracker *trackedTrade) {
 	}
 
 	// Schedule a tick for this trade.
-	delay := time.Duration(numMatches) * time.Second / 10 // 1 sec delay for every 10 active matches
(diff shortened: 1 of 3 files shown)


chappjc:
This successfully creates a single larger redeem by the taker.  Some of the timing is arbitrary, so I'd like to review some of the logs we've seen from the wild recently to get an a better idea about appropriate delay/rate.

client/core/core.go
```diff
@@ -4056,12 +4099,9 @@ func handleAuditRoute(c *Core, dc *dexConnection, msg *msgjson.Message) error {
...
 	}
-	assets, err := c.tick(tracker)
-	if len(assets) > 0 {
-		dc.refreshMarkets()
-		c.updateBalances(assets)
-	}
-	return err
+	// Are we ticking this trade after an audit in case it's mined already?
+	c.schedTradeTick(tracker)
+	return nil
```
chappjc, line 4104: Not sure we need to tick after audit.

client/core/core.go
```diff
@@ -4037,6 +4041,45 @@ func handleNoMatchRoute(c *Core, dc *dexConnection, msg *msgjson.Message) error
...
+	if numMatches == 1 {
+		go tick()
+		return
+	}
+
+	// Schedule a tick for this trade.
+	delay := time.Duration(numMatches) * time.Second / 10 // 1 sec delay for every 10 active matches
+	if delay > 5*time.Second {
+		delay = 5 * time.Second
+	}
```
chappjc, line 4073: Thinking we can infer an appropriate delay based on the number of active matches, not that all the active matches are in the same state, but it's at least an indicator of the upper limit on the reasonable delay.

buck54321: What we're actually up against is the various latencies of all counterparties. I like the idea of increasing the delay a little bit when there are a lot of matches, but I'm thinking 200 ms is too low for 2 matches. How about a constant term like

`delay := 2 * time.Second + time.Duration(numMatches) * time.Second / 10`

chappjc: That's what I realized reviewing logs, there's other party latency here, especially when there are multiple maker accounts for the different matched makers.  Gonna try a constant lower limit for that as you suggested.

client/core/core_test.go
```diff
@@ -2763,18 +2787,40 @@ func TestTradeTracking(t *testing.T) {
...
+		t.Fatalf("wrong match status wanted %v, got %v", order.MatchComplete, status)
+	}
 	if !bytes.Equal(proof.MakerRedeem, redemptionCoin) {
 		t.Fatalf("redemption coin ID not logged")
 	}
 	if len(proof.TakerRedeem) == 0 {
 		t.Fatalf("taker redemption not sent")
 	}
+	rig.db.updateMatchChan = nil
+	tBtcWallet.redeemErrChan = nil
```
buck54321, line 2823: I don't think you need the `defer func` above then.

client/core/core_test.go
```diff
@@ -2638,9 +2658,12 @@ func TestTradeTracking(t *testing.T) {
 	if auth.AuditStamp != audit.Time {
 		t.Fatalf("audit time not set")
 	}
+	time.Sleep(250 * time.Millisecond) // let async tick do nothing
```
buck54321, line 2661: Isn't this too short now? Can make that `2*time.Second` factor a package var for testing, or just drop this `Sleep`, since you're calling `tickAsset` right away anyways.

buck54321 (changes requested):
Not sure why, but I'm still seeing two redemptions for the taker.

1. place 2 lot buy order. wait for it to be booked
2. place 2 1-lot sell orders in the same epoch
3. mine btc, then dcr

```
2020-11-02 13:28:06.708 [INF] CORE: Notified of maker's redemption (dcr: 9fd304c99224e4662421768ae37f9ca07a1f741fd891ca029e25616c606c4e80:0) and validated secret for order e754e62606c889e0f87c436ba74d0a7c42971eec962e658ae72639e468dac002...
2020-11-02 13:28:06.722 [DBG] CORE: Redeemable match 467a9c130b9d6eb8745947f3e2f9092014f3a5f19cf4ca11e48ff889bc93f720 for order e754e62606c889e0f87c436ba74d0a7c42971eec962e658ae72639e468dac002 (Taker)
2020-11-02 13:28:06.724 [INF] CORE: Notified of maker's redemption (dcr: 9fd304c99224e4662421768ae37f9ca07a1f741fd891ca029e25616c606c4e80:1) and validated secret for order 10de46d1df98898c1cc5af0acdfb61675ad09fd202b7a4f4cd9df50b1d068aea...
2020-11-02 13:28:06.726 [DBG] CORE: notify: |POKE| (order) Match complete - Redeemed 2.00000000 dcr on order 72fde0d0 - Order: 72fde0d0367f0c86e7830b40394f3744aef518c5ff4a9a902d7b5cb93107f534
2020-11-02 13:28:06.748 [DBG] CORE: Redeemable match 537a4ea65bee55c4a3bd801576a154b20a4a4ace286dbf8d874379c5e020c9df for order 10de46d1df98898c1cc5af0acdfb61675ad09fd202b7a4f4cd9df50b1d068aea (Taker)
2020-11-02 13:28:06.797 [INF] CORE: Broadcasted redeem transaction spending 1 contracts for order e754e62606c889e0f87c436ba74d0a7c42971eec962e658ae72639e468dac002, paying to 04106f15e8245b4e4d9bca11f66e0057242749e34217041467f48f5482889c44:0 (btc)
2020-11-02 13:28:06.800 [INF] CORE: Broadcasted redeem transaction spending 1 contracts for order 10de46d1df98898c1cc5af0acdfb61675ad09fd202b7a4f4cd9df50b1d068aea, paying to d460a1b395ec1a1c310abe1a892e332b5370048f0419b02222b75d06ac189c1e:0 (btc)
```

chappjc:
> Not sure why, but I'm still seeing two redemptions for the taker.
> 
>     1. place 2 lot buy order. wait for it to be booked
> 
>     2. place 2 1-lot sell orders in the same epoch
> 
>     3. mine btc, then dcr
> 
> 
> ```
> 2020-11-02 13:28:06.708 [INF] CORE: Notified of maker's redemption (dcr: 9fd304c99224e4662421768ae37f9ca07a1f741fd891ca029e25616c606c4e80:0) and validated secret for order e754e62606c889e0f87c436ba74d0a7c42971eec962e658ae72639e468dac002...
> 2020-11-02 13:28:06.722 [DBG] CORE: Redeemable match 467a9c130b9d6eb8745947f3e2f9092014f3a5f19cf4ca11e48ff889bc93f720 for order e754e62606c889e0f87c436ba74d0a7c42971eec962e658ae72639e468dac002 (Taker)
> 2020-11-02 13:28:06.724 [INF] CORE: Notified of maker's redemption (dcr: 9fd304c99224e4662421768ae37f9ca07a1f741fd891ca029e25616c606c4e80:1) and validated secret for order 10de46d1df98898c1cc5af0acdfb61675ad09fd202b7a4f4cd9df50b1d068aea...
> 2020-11-02 13:28:06.726 [DBG] CORE: notify: |POKE| (order) Match complete - Redeemed 2.00000000 dcr on order 72fde0d0 - Order: 72fde0d0367f0c86e7830b40394f3744aef518c5ff4a9a902d7b5cb93107f534
> 2020-11-02 13:28:06.748 [DBG] CORE: Redeemable match 537a4ea65bee55c4a3bd801576a154b20a4a4ace286dbf8d874379c5e020c9df for order 10de46d1df98898c1cc5af0acdfb61675ad09fd202b7a4f4cd9df50b1d068aea (Taker)
> 2020-11-02 13:28:06.797 [INF] CORE: Broadcasted redeem transaction spending 1 contracts for order e754e62606c889e0f87c436ba74d0a7c42971eec962e658ae72639e468dac002, paying to 04106f15e8245b4e4d9bca11f66e0057242749e34217041467f48f5482889c44:0 (btc)
> 2020-11-02 13:28:06.800 [INF] CORE: Broadcasted redeem transaction spending 1 contracts for order 10de46d1df98898c1cc5af0acdfb61675ad09fd202b7a4f4cd9df50b1d068aea, paying to d460a1b395ec1a1c310abe1a892e332b5370048f0419b02222b75d06ac189c1e:0 (btc)
> ```

This groups matches for a given taker trade.  Sounds like you're making two taker orders (different trackedTrades)

client/core/core_test.go
chappjc: The defer was in case this test fails above before getting here, so that subsequent test funcs don't have these chans set.

buck54321: Unlike some of the server packages, the "rig" isn't reused in **core** tests.

client/core/core_test.go
chappjc: There's the exception for `numMatches == 1`, and this sleep is my lazy way of letting the tick complete before setting the SwapConf a couple lines down.  Let me add some proper sync instead.

buck54321 (approved):

Issue #752, opened 2020-10-16
ui: need a visible notification when an order is revoked


buck54321:
We currently send out a `Data` notification to force a status update in the "my orders" table. That is sent from `(*trackedTrade).revoke`. That can't be a `ErrorLevel` or `WarningLevel` notification, because `revoke` is used from two places with different severity. So the caller will need to send the notification instead. the `"revoke"`-subject notification itself can be removed as long as we are consistent about the caller notifying.

chappjc:
@kevinstl is taking this

kevinstl:
> That can't be a `ErrorLevel` or `WarningLevel` notification, because `revoke` is used from two places with different severity.

I see two places where revoke is initiated, handleRevokeOrderMsg() and handleTradeSuspensionMsg().

I'm assuming:

handleRevokeOrderMsg() is WarningLevel
handleTradeSuspensionMsg() is ErrorLevel

Is that correct?

buck54321:
I was actually thinking the opposite. Maybe someone else can weigh in.

JoeGruffins:
I think trade suspension error level would depend on whether or not the user was informed beforehand. If expected, warning. An unexpected suspension would be an error in my opinion.

PR #798, opened 2020-10-29
ui: need a visible notification when an order is revoked

kevinstl:
closes https://github.com/decred/dcrdex/issues/752

I could not quickly figure out how to reproduce triggering of the revoke notifications, but the change seemed simple enough that I decided to push this version to get the ball rolling.


client/core/trade.go
```diff
@@ -1085,7 +1085,7 @@ func (t *trackedTrade) revoke() {
 		t.dc.log.Errorf("unable to update order: %v", err)
 	}
 
-	t.notify(newOrderNote("revoke", "", db.Data, t.coreOrderInternal()))
+	t.notify(newOrderNote("revoke", "", severity, t.coreOrderInternal()))
```
buck54321, line 1088: This was a `Data`-`Severity`notification, which is why there is no message and the subject is not capitalized. [`Data` notifications are not meant for display](https://github.com/decred/dcrdex/blob/dfbb3b334d740dd353d6d23237aad03fde9f9904/client/db/types.go#L32). If you set the severity to `WarningLevel`, `ErrorLevel`, or `Success`, the user will get a UI notification (the bell) with no details and just the work "revoke" for a subject.

The caller needs to send the notification themselves, with custom messaging and severity. Once **all** of the callers are sending notifications, we can actually remove this line. This was just here to force an update of the browsers cached order data, but any `OrderNote` will do.

client/core/core_test.go
```diff
@@ -76,6 +76,7 @@ var (
 	tSwapFeesPaid       uint64 = 500
 	tRedemptionFeesPaid uint64 = 350
 	tLogger                    = dex.StdOutLogger("TCORE", dex.LevelTrace)
+	tNotified                  = false
```
buck54321, line 79: This shouldn't be a package-level var. Use a `NotificationFeed` in the test function, and look for only the notification you want.

server/cmd/dexcoin/go.sum
```diff
@@ -33,6 +33,7 @@ github.com/dchest/siphash v1.2.1/go.mod h1:q+IRvb2gOSrUnYoPqHiyHXS0FOBBOdl6tONBl
 github.com/decred/base58 v1.0.0/go.mod h1:LLY1p5e3g91byL/UO1eiZaYd+uRoVRarybgcoymu9Ks=
```
buck54321, line 33: These changes snuck in on you somehow. Please undo.

buck54321 (changes requested):

client/core/trade.go
kevinstl: Callers are sending notifications now.

client/core/core_test.go
kevinstl: I'm using NotificationFeed now.

server/cmd/dexcoin/go.sum
kevinstl: Done.

client/core/bookie.go
```diff
@@ -511,6 +511,8 @@ func handleTradeSuspensionMsg(c *Core, dc *dexConnection, msg *msgjson.Message)
 			tracker.metaData.Host == dc.acct.host && tracker.metaData.Status == order.OrderStatusBooked {
 			// Locally revoke the purged book order.
 			tracker.revoke()
+			details := fmt.Sprintf("Order %v revoked.", tracker.Order.ID())
+			c.notify(newOrderNote("Order Revoked", details, db.ErrorLevel, tracker.coreOrderInternal()))
```
chappjc, line 515: Not sure we want to have this one be an error since it's a scheduled event. For `handleRevokeOrderMsg`, I'd agree error is best there.  Maybe warning? @buck54321?

kevinstl: Buck indicated it should be warning, it should have been that way already. The test missed it because it was not properly checking. It's fixed now.

client/core/bookie.go
```diff
@@ -511,6 +511,8 @@ func handleTradeSuspensionMsg(c *Core, dc *dexConnection, msg *msgjson.Message)
 			tracker.metaData.Host == dc.acct.host && tracker.metaData.Status == order.OrderStatusBooked {
 			// Locally revoke the purged book order.
 			tracker.revoke()
+			details := fmt.Sprintf("Order %v revoked.", tracker.Order.ID())
```
buck54321, line 514: Something like... 

`fmt.Sprintf("Order %s on market %s at %s revoked due to market suspension", tracker.token(), sp.MarketID, dc.acct.host)`

client/core/core.go
```diff
@@ -3671,6 +3671,9 @@ func handleRevokeOrderMsg(c *Core, dc *dexConnection, msg *msgjson.Message) erro
 
 	tracker.revoke()
 
+	details := fmt.Sprintf("Order %v revoked.", oid)
```
buck54321, line 3674: Something like...

```
fmt.Sprintf("Order %s on market %s at %s has been revoked by the server", tracker.token, tracker.mktID, dc.acct.host)
```

client/core/core_test.go
```diff
@@ -4287,6 +4299,56 @@ func TestHandleTradeSuspensionMsg(t *testing.T) {
...
+			case *OrderNote:
+				orderNote = note
+			}
+		}
+		return orderNote
+	}
+	return getOrderNote
+}
+
+func verifyRevokeNotification(getOrderNote func() *OrderNote, expectedOrderNote *OrderNote, t *testing.T) {
```
buck54321, line 4332: Notification subjects should be unique, so that's the only thing needed for comparison.

client/core/core_test.go
```diff
@@ -4287,6 +4299,56 @@ func TestHandleTradeSuspensionMsg(t *testing.T) {
...
+		for orderNote == nil {
+			ntfn := getNotification("feepayment or balance")
+			switch note := ntfn.(type) {
+			case *OrderNote:
+				orderNote = note
+			}
+		}
+		return orderNote
+	}
+	return getOrderNote
```
buck54321, line 4329: `return func() *OrderNote {...`

client/core/core_test.go
```diff
@@ -4287,6 +4299,56 @@ func TestHandleTradeSuspensionMsg(t *testing.T) {
 	}
 }
 
+func setupOrderNotificationVerification(tCore *Core, t *testing.T) func() *OrderNote {
```
buck54321, line 4302: I'd suggest a couple of changes here. 1) The name doesn't seem descriptive of the function, 2) How about a goroutine and a channel instead. Otherwise, your feed is just filling with notes while you're not running the `getOrderNote` function.

Something like ....

```
func orderNoteFeed(tCore *Core, t *testing.T) chan *OrderNote {
  ch := tCore.NotificationFeed()
  outChan := make(chan *OrderNote)
  go func(){
      for {
          select {
          case n := <-ch:
              ordNote, ok := n.(*OrderNote)
              if ok {
                  outChan <- ordNote
              }
          case <-tCtx.Done():
              return
          }
      }
  }()
  return outChan
}
```

client/core/bookie.go
```diff
@@ -511,6 +511,8 @@ func handleTradeSuspensionMsg(c *Core, dc *dexConnection, msg *msgjson.Message)
...
+			c.notify(newOrderNote("Order Revoked", details, db.WarningLevel, tracker.coreOrderInternal()))
```
buck54321, line 515: Unique subjects when possible. How about `Order Auto-revoked` here, and leave `Order Revoked` below.

buck54321 (changes requested):

client/core/core.go
chappjc: For future work we'll need to add details and reaons.  This will require passing more data around in a number of places e.g. swapper->unbookHook

client/core/core_test.go
```diff
@@ -2402,12 +2402,18 @@ func TestHandleRevokeOrderMsg(t *testing.T) {
...
+
 	// Success
 	err = handleRevokeOrderMsg(rig.core, rig.dc, req)
 	if err != nil {
 		t.Fatalf("handleRevokeOrderMsg error: %v", err)
 	}
 
+	details := fmt.Sprintf("Order %s on market %s at %s has been revoked by the server", tracker.token(), tracker.mktID, dc.acct.host)
+	expectedOrderNote := newOrderNote("Order Revoked", details, db.ErrorLevel, tracker.coreOrderInternal())
+	verifyRevokeNotification(ch, expectedOrderNote, t)
```
buck54321, line 2415: You don't need to construct or compare an entire notification here. Subject comparison is enough.

client/core/core_test.go
```diff
@@ -4361,6 +4374,54 @@ func TestHandleTradeSuspensionMsg(t *testing.T) {
...
+						expectedOrderNote.SubjectText)
+				}
+				return
+			case <-tCtx.Done():
+				return
+			case <-time.NewTimer(time.Second).C:
+				t.Fatal("timed out waiting for OrderNote notification")
+				return
+			}
+		}
```
buck54321, line 4420: If I'm not mistaken, the `WaitGroup`, the goroutine, and the `for` loop are all unnecessary and identical to just doing the select block alone.

buck54321 (changes requested):

client/core/core_test.go
kevinstl: The problem I encountered without a WaitGroup was that it seemed the test could complete before timing out in the case that the notification was not sent. I ensured that omitting the notification cause a test failure from timing out. Without the wait the test would intermittently pass or fail.

client/core/core_test.go
kevinstl: Passing subject only now.

client/core/core_test.go
chappjc: I don't follow either.  How could there be more than one iteration of the `for` loop?  Each case has a return and the timeout will fire if the other cases do not?

Also, Fatal/Fatalf are prohibited in goroutines that are not the main test goroutine.

```go
func verifyRevokeNotification(ch chan *OrderNote, expectedSubject string, t *testing.T) {
	select {
	case actualOrderNote := <-ch:
		if expectedSubject != actualOrderNote.SubjectText {
			t.Fatalf("SubjectText mismatch. %s != %s", actualOrderNote.SubjectText,
				expectedSubject)
		}
		return
	case <-tCtx.Done():
		return
	case <-time.After(time.Second):
		t.Fatalf("timed out waiting for OrderNote notification")
		return
	}
}
```

Isn't the above correct?

kevinstl: That does work. I thought I had tried that as I went through many different iterations, but I should've tested buck's assertion.

client/core/core_test.go
```diff
@@ -4373,6 +4382,46 @@ func TestHandleTradeSuspensionMsg(t *testing.T) {
...
+			case <-tCtx.Done():
+				return
+			case <-time.NewTimer(time.Second).C:
+				t.Fatal("timed out waiting for notifications")
+				return
+			}
+		}
+	}()
+	return outChan
+}
```
chappjc, line 4407: The goroutine started by `orderNoteFeed` has a similar issue with `Fatal`, which is not allowed to be called from goroutines other than the one running the test function.  See https://golang.org/pkg/testing/#B.FailNow (called by `Fatal`/`Fatalf`)

Also, there's nothing in the tests that makes the goroutine return before the test returns.  I suggest changing the function to:

```go
func orderNoteFeed(tCore *Core) (orderNotes chan *OrderNote, done func()) {
	orderNotes = make(chan *OrderNote, 1)

	ntfnFeed := tCore.NotificationFeed()
	feedDone := make(chan struct{})
	var wg sync.WaitGroup
	wg.Add(1)
	go func() {
		defer wg.Done()
		for {
			select {
			case n := <-ntfnFeed:
				ordNote, ok := n.(*OrderNote)
				if ok {
					orderNotes <- ordNote
					return // just one OrderNote and done
				}
			case <-tCtx.Done():
				return
			case <-feedDone:
				return
			}
		}
	}()

	done = func() {
		close(feedDone) // close first on return
		wg.Wait()
	}
	return orderNotes, done
}
```

And calling it like:

```go
orderNotes, feedDone := orderNoteFeed(tCore)
defer feedDone()
```

client/core/core.go
```diff
@@ -3790,6 +3790,9 @@ func handleRevokeOrderMsg(c *Core, dc *dexConnection, msg *msgjson.Message) erro
 
 	tracker.revoke()
 
+	details := fmt.Sprintf("Order %s on market %s at %s has been revoked by the server", tracker.token(), tracker.mktID, dc.acct.host)
+	c.notify(newOrderNote("Order Revoked", details, db.ErrorLevel, tracker.coreOrderInternal()))
```
chappjc, line 3794: This needs to be `coreOrder()` instead of `coreOrderInternal()` since the trade's mutex is not locked at this point.

client/core/bookie.go
```diff
@@ -511,6 +511,8 @@ func handleTradeSuspensionMsg(c *Core, dc *dexConnection, msg *msgjson.Message)
 			tracker.metaData.Host == dc.acct.host && tracker.metaData.Status == order.OrderStatusBooked {
 			// Locally revoke the purged book order.
 			tracker.revoke()
+			details := fmt.Sprintf("Order %s on market %s at %s revoked due to market suspension", tracker.token(), sp.MarketID, dc.acct.host)
+			c.notify(newOrderNote("Order Auto-revoked", details, db.WarningLevel, tracker.coreOrderInternal()))
```
chappjc, line 515: Same here with `coreOrder()`

client/core/core_test.go
kevinstl: That makes sense. Changes made.

client/core/bookie.go
chappjc, line 515: @buck54321 since these are no longer `"revoked"` (now `SubjectRevoke`), is the frontend going to have trouble updating the order status or is the subject ignored and just updates status?  I think that's what you said, but I want to be sure.

@kevinstl Depending on what @buck54321 says, I believe you can change `SubjectRevoke` to `"Order Revoked"` and use that in handleRevokeOrderMsg, and add a new subject var for `"Order Auto-revoked"` in `handleTradeSuspensionMsg`.

Otherwise, good to go here.

buck54321: I do switch on a couple of subjects on the front end, but not these. Good to go.

chappjc: Just to clarify, good to make these subject constants or just literals?

chappjc: @kevinstl This is pretty much ready to go.  Conflict resolution here would be easiest to squash first, then rebase.  One way:

```
git checkout master
git pull
git checkout revoke-order-notification
git rebase -i $(git merge-base master HEAD)
```

Then "fix" all commits but the first unless you want to include any of the messages with "squash":

```
pick e195c29e Allow for revoke notifications at variable levels
f bb613c6e Send revoke notification from callers
f 323b1f50 Revert changes to go.sum
f 1ee91dd6 Set suspension revoke to warning and make test effective
f 1939b56e Add token, market id and host to order suspension revoke notification details
f 4012cd3d Add token, market id and host to revoke order notification details
f beb19f77 Set trade suspension order revoke note subject to unique value
f f33cf135 Follow NotificationFeed pattern to listen specifically for OrderNote
f ee6aee65 Pass subject string to revoke notification verification for comparison
f 8a44a0fe Use select statement instead of for loop within goroutine with wait
f deec90c1 Ensure orderNoteFeed goroutine returns prior to test returning
f 1613f887 Use coreOrder function that does not require locked mutex
f 475dc3bd Revert changes to go.sum
```

That will have squashed all the commits down to one without rebasing on `master` yet.

Then rebase on `master` and resolve conflicts:

```
git rebase master
git mergetool # or whatever you want to do to resolve
# when all conflicts resolved and changes are staged:
git rebase --continue
```

(just remove the `t.notify` from `(*trackedTrade).revoke` afaict)

buck54321: > Just to clarify, good to make these subject constants or just literals?

Please make constants, `SubjectOrderRevoked` and `SubjectOrderAutoRevoked`.

chappjc (approved):

buck54321 (approved):

PR #799, opened 2020-10-30
client/core: get trade_simnet_test working again

chappjc:
This gets the trade_simnet_test working again.

Since the BTC gamma and delta simnet wallets are no longer encrypted (requires no pass), the previous trick used by `TestNoMaker` to prevent a swap txn stopped working for this BTC wallet.  Instead, this now sets `xcWallet.encPW` and `xcWallet.pw` to force `locallyUnlocked` to block swap actions.

`TestMakerGhostingAfterTakerRedeem` was not working before, and apparently it was because `swapErr` was not `nil`ed to allow resuming the swap process after the maker disconnects from the DEX server.  This is fixed now.

There is also a lot of cleanup to trade_simnet_test.go.

The lot size of the dcrdex simnet harness was increased recently, so the tests are updated with appropriate order quantities, and the asset harnesses modified to fund the secondary (non-alpha) wallets much more generously.

I've also noticed that the `signrawtransaction` and `dumpprivkey` RPCs return **very** confusing error messages that actually just mean the wallet is locked, so I've supplemented these errors and log messages.

### DIFF
--- a/client/core/trade_simnet_test.go
+++ b/client/core/trade_simnet_test.go
@@ -52,7 +52,7 @@ var (
 		appPass: []byte("client1"),
 		wallets: map[uint32]*tWallet{
 			dcr.BipID: dcrWallet("trading1"),
-			btc.BipID: btcWallet("beta", "delta"),
+			btc.BipID: btcWallet("beta", ""), // beta default ("") is encrypted, delta is not
 		},
 	}
 	client2 = &tClient{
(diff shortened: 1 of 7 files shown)


chappjc:
Also, `TestMakerGhostingAfterTakerRedeem` doesn't work on master anymore.  The Maker does not seem to do the redeem that is required for the test.

client/core/trade_simnet_test.go
```diff
@@ -605,10 +617,10 @@ func TestOrderStatusReconciliation(t *testing.T) {
 	}
 	totalLockedByTrades := c2Balance.Locked - preTradeLockedBalance
 	preDisconnectLockedBalance := c2Balance.Locked   // should reduce after funds are returned
-	preDisconnectAvialableBal := c2Balance.Available // should increase after funds are returned
+	preDisconnectAvailableBal := c2Balance.Available // should increase after funds are returned
 
-	// Disconnect the DEX and allow sometime for DEX to udpate order statuses.
-	client2.log("disconnecting DEX")
+	// Disconnect the DEX and allow sometime for DEX to update order statuses.
```
JoeGruffins, line 622: some time

JoeGruffins (approved):
```
PASS
ok  	decred.org/dcrdex/client/core	463.970s
```

PR #802, opened 2020-10-30
fix extras loop in authDEX

chappjc:


chappjc:
This is going to be combined with match resolution fixed in a coming PR.  Closing.

Issue #379, opened 2020-05-18
core: need targeted updates to (*Core).user object


buck54321:
The `Core` keeps a cached `User` struct, which is basically an application state that is populated whenever the user is logged in. The `Core.user` is only updated when `refreshUser` is called, and the `dexConnection`'s `marketMap` must also be updated when changes are made using `refreshMarkets`. This is a pretty unsophisticated way of handling updates to the cached `User`, since it recreates every `Match`, `Order`, `Market`, and `Exchange`.

This is not actually a bug, but the pattern is prone to errors because it relies on the programmer to call the required function(s) every time an update is made to basically anything. A couple of possible corrective actions.

1. Catch the `OrderNote`, `FeePaymentNote`, etc. in `(*Core).notify` and only update the appropriate `Exchange.Markets` entry.
2. A separate method for wallet updates, which do not affect the `User.Exchanges` map.

These are just suggestions, and there may be better ways to handle this.

chappjc:
I had just made a duplicate issue (#904) because I forgot about this one.  I'll just reiterate the issue from a different angle here:

```
In `client/core.Core` there a numerous calls to `refreshUser`, and most of these calls are required only to update one small piece of the information that `refreshUser` updates.  e.g.  If `WalletState` for a certain `xcWallet` changes, there's no reason to update all of the other assets, orders, exchanges, etc.  This is particularly problematic because there are many mutexes locked by `refreshUser` and it's callers, which can easily lead to deadlock and/or contorted patterns to avoid deadlock.

In cases where `refreshUser` really is the best option, not a more targeted mutator method, each `refreshUser` call site should be documented with the reasoning.  If `refreshUser` is used for convenience regardless, the call site should document why exactly it is being called (what just changed that requires updating?).
```

buck54321:
[See this comment too](https://github.com/decred/dcrdex/pull/805#discussion_r519236529) for a possible solution. Since every update to the `User` should involve a notification, we could hook into the notification loop to apply these targeted updates, as is done with `updateCoreMatch`.

chappjc:
Ah, yes, call the targeted update method from the notify loop. That might be a good solution, although I've not felt it out carefully.  @amassarwi  Maybe see if this approach make sense for calling the updateUserWalletState method as you've done in https://github.com/decred/dcrdex/pull/902/commits/7b6232acfdea9d7fb3d2059664bf65b96c101b70?  I'm not sure if it would simplify it or make it less error prone, but worth trying it it out.

[...]

PR #805, opened 2020-11-02
core,ui: add confirmation for currently waiting swap

buck54321:
Add confirmation count to match cards on the order details page, for matches in `MakerSwapCast` or `TakerSwapCast`. It's a bigger job than it sounds like, because 1) we weren't tracking confirmation count on our own swaps yet, and 2) I didn't think we should add the count unless it updated live, so the count, match status, and various coin ID links will update live now. 

There are just a few more pieces needed to make the whole page update live. Will save the rest for another PR.

### DIFF
--- a/client/core/notification.go
+++ b/client/core/notification.go
@@ -192,10 +192,9 @@ type MatchNote struct {
 }
 
 func newMatchNote(subject, details string, severity db.Severity, t *trackedTrade, match *matchTracker) *MatchNote {
-	oid := t.ID()
 	return &MatchNote{
 		Notification: db.NewNotification(NoteTypeMatch, subject, details, severity),
-		OrderID:      oid[:],
+		OrderID:      t.ID().Bytes(),
 		Match: matchFromMetaMatchWithConfs(t.Order, &match.MetaMatch, match.swapConfirms,
 			int64(t.wallets.fromAsset.SwapConf), match.counterConfirms, int64(t.wallets.toAsset.SwapConf)),
 		Host:     t.dc.acct.host,
--- a/client/core/notification.go
+++ b/client/core/notification.go
@@ -4,6 +4,7 @@
 package core
 
 import (
+	"bytes"
 	"fmt"
 
 	"decred.org/dcrdex/client/db"
@@ -15,6 +16,7 @@ const (
 	NoteTypeFeePayment   = "feepayment"
 	NoteTypeWithdraw     = "withdraw"
 	NoteTypeOrder        = "order"
+	NoteTypeMatch        = "match"
 	NoteTypeEpoch        = "epoch"
 	NoteTypeConnEvent    = "conn"
 	NoteTypeBalance      = "balance"
@@ -45,6 +47,11 @@ func (c *Core) notify(n Notification) {
 	}
 	logFun("notify: %v", n)
 
+	switch nt := n.(type) {
+	case *MatchNote:
+		c.updateCoreMatch(nt)
+	}
+
 	c.noteMtx.RLock()
 	for _, ch := range c.noteChans {
 		select {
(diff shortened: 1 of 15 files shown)


client/core/notification.go
```diff
@@ -155,6 +182,27 @@ func newOrderNote(subject, details string, severity db.Severity, corder *Order)
...
+	Match    *Match    `json:"match"`
+	Host     string    `json:"host"`
+	MarketID string    `json:"marketID"`
+}
+
+func newMatchNote(subject, details string, severity db.Severity, t *trackedTrade, match *matchTracker) *MatchNote {
+	oid := t.ID()
+	return &MatchNote{
+		Notification: db.NewNotification(NoteTypeMatch, subject, details, severity),
+		OrderID:      oid[:],
```
chappjc, line 198: This is fine, but you could do `OrderID:     t.ID().Bytes(),`

client/core/types.go
```diff
@@ -126,41 +126,126 @@ type Match struct {
...
+}
+
+// matchFromMetaMatch constructs a *Match from an Order and a *MetaMatch.
+// This function is intended for use with inactive matches. For active matches,
+// use matchFromMetaMatchWithConfs.
+func matchFromMetaMatch(ord order.Order, metaMatch *db.MetaMatch) *Match {
+	return matchFromMetaMatchWithConfs(ord, metaMatch, 0, 0, 0, 0)
+}
+
+// matchFromMmatchFromMetaMatchWithConfs constructs a *Match from a *MetaMatch,
```
chappjc, line 183: stray "m" in the function name in docs

client/core/types.go
```diff
@@ -126,41 +126,126 @@ type Match struct {
...
+	return &Coin{
+		ID:       coinID,
+		StringID: coinIDString(assetID, coinID),
+		AssetID:  assetID,
+		Symbol:   unbip(assetID),
 	}
+}
+
+// Confirmations is the confirmation count and confirmation requirement of a
+// SwapCoin.
```
chappjc, line 162: SwapCoin -> "swap Coin"?

client/core/types.go
```diff
@@ -126,41 +126,126 @@ type Match struct {
...
+// use matchFromMetaMatchWithConfs.
+func matchFromMetaMatch(ord order.Order, metaMatch *db.MetaMatch) *Match {
+	return matchFromMetaMatchWithConfs(ord, metaMatch, 0, 0, 0, 0)
+}
+
+// matchFromMmatchFromMetaMatchWithConfs constructs a *Match from a *MetaMatch,
+// and sets the confirmations for swaps-in-waiting.
+func matchFromMetaMatchWithConfs(ord order.Order, metaMatch *db.MetaMatch, swapConfs, swapReq, counterSwapConfs, counterReq int64) *Match {
+	side := metaMatch.Match.Side
+	sell := ord.Trade().Sell
```
chappjc, line 187: `ord.Trade()` is nil for cancel orders, and thus panic, and while this shouldn't be called for a cancel order match (I think) I'd feel better if this function had logic to deal with that possibility.

client/core/types.go
```diff
@@ -188,7 +273,7 @@ type Order struct {
 	Cancelling   bool              `json:"cancelling"`
 	Canceled     bool              `json:"canceled"`
 	FeesPaid     *FeeBreakdown     `json:"feesPaid"`
-	FundingCoins []dex.Bytes       `json:"fundingCoins"`
+	FundingCoins []*Coin           `json:"fundingCoins"`
```
chappjc, line 276: This and the the Match changes rule out 0.1.x unfortunately, which is no big deal as we'll move on to 0.2.0 right away.

client/core/trade.go
```diff
@@ -741,15 +746,29 @@ func (t *trackedTrade) isSwappable(match *matchTracker) bool {
...
+			// If the maker is the counterparty, we can determine swappability
+			// based on the confirmations.
+			confs, req, changed := t.counterPartyConfirms(match)
+			ready := confs >= req
+			if changed && !ready {
+				t.dc.log.Debugf("Match %v not yet swappable: current confs = %d, required confs = %d",
+					match.id, confs, req)
+			}
+			return ready
+		} else {
```
chappjc, line 761: No `else` needed since previous block ends with a `return`.  Same in `isRedeemable`.

chappjc (changes requested):

chappjc:
Bump, but no rush.  Just wanted to make sure this PR doesn't get lost.

client/core/notification.go
```diff
@@ -56,6 +63,26 @@ func (c *Core) notify(n Notification) {
 	c.noteMtx.RUnlock()
 }
 
+// updateCoreMatch performs a targeted update of a match in the User struct.
+func (c *Core) updateCoreMatch(n *MatchNote) {
```
buck54321, line 67: This function and the way it's called from the notification feed might be a model for a solution to https://github.com/decred/dcrdex/issues/379.

chappjc (approved):
Looks good.  Working flawlessly.

PR #813, opened 2020-11-04
align 64-bit fields for atomic

buck54321:
See https://golang.org/pkg/sync/atomic/#pkg-note-BUG. Have seen a panic in the wild.

### DIFF
--- a/client/comms/wsconn.go
+++ b/client/comms/wsconn.go
@@ -99,10 +99,12 @@ type WsCfg struct {
 
 // wsConn represents a client websocket connection.
 type wsConn struct {
+	// 64-bit atomic variables first. See
+	// https://golang.org/pkg/sync/atomic/#pkg-note-BUG.
+	rID    uint64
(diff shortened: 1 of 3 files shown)


chappjc (approved):
Had the same PR brewing.  Those are the only three 64 bit atomic accesses I found too.

client/comms/wsconn.go
```diff
@@ -99,10 +99,12 @@ type WsCfg struct {
 
 // wsConn represents a client websocket connection.
 type wsConn struct {
+	// 64-bit atomic variables first. See
+	// https://golang.org/pkg/sync/atomic/#pkg-note-BUG.
+	rID    uint64
 	cancel context.CancelFunc
 	wg     sync.WaitGroup
```
chappjc, line 106: of course WaitGroup is 12 bytes

chappjc:
Will you please dup this pr on `release-0.1` for our tester?

PR #814, opened 2020-11-04
release-0.1: align 64-bit fields for atomic

buck54321:
Release branch version of #813.


chappjc (commented):
Tested well on ARM 32 bit

PR #817, opened 2020-11-05
core: avoid wallet locked errors

buck54321:
This PR introduces wallet password caching and some additional logic to ensure wallets are unlocked when needed. Another important change is that the `asset.Wallet` interface now has a `Locked() bool` method that needs to be implemented. Already done for **dcr** and **btc** obviously.

### DIFF
--- a/client/asset/btc/wallettypes.go
+++ b/client/asset/btc/wallettypes.go
@@ -109,6 +109,10 @@ type GetWalletInfoResult struct {
 	KeyPoolSizeHDInternal uint32  `json:"keypoolsize_hd_internal"`
 	PayTxFee              float64 `json:"paytxfee"`
 	HdSeedID              string  `json:"hdseedid"`
+	// UnlockedUntil is a pointer because for encrypted locked wallets, it will
+	// be zero, but for unencrypted wallets the field won't be present in the
+	// response.
+	UnlockedUntil *int64 `json:"unlocked_until"`
 	// HDMasterKeyID is dropped in Bitcoin Core 0.18
 	HdMasterKeyID     string `json:"hdmasterkeyid"`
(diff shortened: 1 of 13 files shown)


client/asset/btc/wallettypes.go
```diff
@@ -109,6 +109,10 @@ type GetWalletInfoResult struct {
 	KeyPoolSizeHDInternal uint32  `json:"keypoolsize_hd_internal"`
 	PayTxFee              float64 `json:"paytxfee"`
 	HdSeedID              string  `json:"hdseedid"`
+	// UnlockedUntil is a pointer because for encrypted locked wallets, it will
+	// be zero, but for unencrypted wallets the field won't be present in the
+	// response.
+	UnlockedUntil *int64 `json:"unlocked_until"`
```
chappjc, line 115: Did you want omitempty for the unencrypted wallet case?

client/core/core.go
```diff
@@ -1297,7 +1297,7 @@ func (c *Core) CreateWallet(appPW, walletPW []byte, form *WalletForm) error {
 		return fmt.Errorf(s, a...)
 	}
 
-	err = wallet.Unlock(crypter, aYear)
```
chappjc, line 1300: `aYear` is unused now

chappjc (changes requested):
Looks good, just the one clean-up item with `aYear`.

Also noting that the trade_simnet_test.go tests are gonna be even more broken now because wallet locking was actually used to stop actions in those tests.  https://github.com/decred/dcrdex/pull/799/files#diff-a759d94cd368816442988c181142442065b32de45d225ac4a4da1ab9fe2ac44dR689

chappjc (approved):

PR #717, opened 2020-10-03
testing: add a load tester

buck54321:
The loadbot package provides "programs" for load testing the simnet harness. Non-ideal network conditions can be simulated with the use of [a proxy server](https://github.com/Shopify/toxiproxy). The bot will quit as soon as an error is encountered while trying to place an order.

#### Run a program

There are three primary programs. You run one program at a time. 
```
./loadbot -p pingpong
./loadbot -p sidestacker
./loadbot -p compound
```
Debug logging can be enabled with the `-debug` flag. Trace logging can be enabled with the `-trace` flag.

See [the loadbot README](https://github.com/buck54321/dcrdex/blob/loadbot/dex/testing/loadbot/README.md) for network condition simulation options.

PR #818, opened 2020-11-06
core: constant subjects for notifications

buck54321:
Broken off from #717. Includes `MatchNote` changes that aren't yet used, and a correction for the details of an `OrderNote`.

### DIFF
--- a/client/core/notification.go
+++ b/client/core/notification.go
@@ -9,6 +9,7 @@ import (
 
 	"decred.org/dcrdex/client/db"
 	"decred.org/dcrdex/dex"
+	"decred.org/dcrdex/dex/order"
 )
 
 // Notifications should use the following note type strings.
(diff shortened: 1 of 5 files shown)


chappjc:
@dajohi Please have a look at the constants and how they are used when you have a chance https://github.com/decred/dcrdex/pull/818/files#diff-c5fccd8f7f5c11f3e53e2cf6cd8e633e9899ca29b7a20a30df549739dfcb5964R169

client/core/notification.go
```diff
@@ -148,6 +166,38 @@ type OrderNote struct {
...
+	MatchesRefundedSubject      = "Matches Refunded"
+	MatchRevokedSubject         = "Match revoked"
+	RevokeSubject               = "revoke"
+	MatchRecoveredSubject       = "Match recovered"
+	CancellingOrderSubject      = "Cancelling order"
+	OrderStatusUpdateSubject    = "Order status update"
+	MatchResolutionErrorSubject = "Match resolution error"
+	FailedCancelSubject         = "Failed cancel"
+)
+
```
dajohi, line 200: I'd prefer "Subject" be the prefix instead.

dajohi (dismissed):

client/core/notification.go
chappjc: Second that.

buck54321: Done

client/core/notification.go
```diff
@@ -163,6 +174,11 @@ type WithdrawNote struct {
 	db.Notification
 }
 
+const (
+	WithdrawErrorSubject = "Withdraw error"
+	WithdrawSendSubject  = "Withdraw sent"
```
chappjc, line 179: Missed the subject prefix on these two.

chappjc (changes requested):
Looks good, just two subject variable names need adjusting.

chappjc (commented):
Actually, I'm getting the following for new matches:

image

chappjc (approved):

PR #819, opened 2020-11-06
processAudit runs txn search async, redo match res. locking

chappjc:
This prevents contract auditing from blocking all incoming messages (and pretty much any other activities for the trade in question).

Instead of expiring the coin waiter after broadcast timeout, it just goes until it finds the contract or the match is revoked.

The trackedTrade locking is also changed a great deal in match status resolution, no longer locking a trackedTrade unless and until needed.

### DIFF
--- a/client/core/core_test.go
+++ b/client/core/core_test.go
@@ -5442,7 +5442,7 @@ func TestMatchStatusResolution(t *testing.T) {
 	rig.db.updateMatchChan = make(chan order.MatchStatus, 4)
 	tCore.authDEX(dc)
 	for i := 0; i < 4; i++ {
-		fmt.Println(<-rig.db.updateMatchChan)
+		<-rig.db.updateMatchChan
 	}
 	trade.mtx.Lock()
 	newStatus1 := match.MetaData.Status
--- a/client/core/core_test.go
+++ b/client/core/core_test.go
@@ -1710,9 +1710,19 @@ func TestLogin(t *testing.T) {
 	tCore = rig.core
 	rig.acct.markFeePaid()
 	rig.queueConnect(nil, []*msgjson.Match{knownMsgMatch /* missing missingMatch! */, extraMsgMatch}, nil)
-	_, err = tCore.Login(tPW)
+	// Login>authDEX will do 4 match DB updates for these two matches:
+	// missing -> revoke -> update match
+	// extra -> negotiate -> newTrackers -> update match
+	// matchConflicts (from extras) -> resolveMatchConflicts -> resolveConflictWithServerData
+	// 	-> update match after spawning auditContract
+	// 	-> update match in auditContract (second because of lock) ** the ASYNC one we have to wait for **
+	rig.db.updateMatchChan = make(chan order.MatchStatus, 4)
+	_, err = tCore.Login(tPW) // authDEX -> async contract audit for the extra match
 	if err != nil || !rig.acct.authed() {
 		t.Fatalf("final Login error: %v", err)
+	}
+	for i := 0; i < 4; i++ {
+		<-rig.db.updateMatchChan
 	}
 
 	if !tracker.matches[missingID].MetaData.Proof.SelfRevoked {
@@ -1730,8 +1740,8 @@ func TestLogin(t *testing.T) {
 		t.Errorf("Extra trade not accepted into matches")
 	}
 	tracker.mtx.Lock()
+	defer tracker.mtx.Unlock()
 	match = tracker.matches[extraID]
-	tracker.mtx.Unlock()
 	if !bytes.Equal(match.MetaData.Proof.CounterScript, missedContract) {
 		t.Errorf("Missed maker contract not retrieved, %s, %s", match.id, hex.EncodeToString(match.MetaData.Proof.CounterScript))
 	}
@@ -2626,7 +2636,7 @@ func TestTradeTracking(t *testing.T) {
 
 	// Check audit errors.
 	tBtcWallet.auditErr = tErr
-	err = tracker.auditContract(match, audit.CoinID, audit.Contract, time.Second)
+	err = tracker.auditContract(match, audit.CoinID, audit.Contract)
 	if err == nil {
 		t.Fatalf("no maker error for AuditContract error")
 	}
@@ -2634,7 +2644,7 @@ func TestTradeTracking(t *testing.T) {
 	// Check expiration error.
 	match.MetaData.Proof.SelfRevoked = true // keeps trying unless revoked
 	tBtcWallet.auditErr = asset.CoinNotFoundError
-	err = tracker.auditContract(match, audit.CoinID, audit.Contract, time.Second)
+	err = tracker.auditContract(match, audit.CoinID, audit.Contract)
 	if err == nil {
 		t.Fatalf("no maker error for AuditContract expiration")
 	}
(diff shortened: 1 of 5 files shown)


client/core/trade.go
```diff
@@ -1739,79 +1758,112 @@ func (t *trackedTrade) refundMatches(matches []*matchTracker) (uint64, error) {
...
-			errChan <- nil
+			if errors.Is(err, asset.CoinNotFoundError) {
+				// Didn't find it that time.
+				if t.MatchIsRevoked(match) {
+					errChan <- ExpirationErr(fmt.Sprintf("match revoked while waiting to find counterparty contract coin %v (%s). "+
+						"Check your internet and wallet connections!", contractID, contractSymb))
+					return wait.DontTryAgain
+				}
+				if tries > 2 {
+					t.dc.log.Infof("Still searching for contract coin %v (%s) for match %v. "+
```
chappjc, line 1853: Still considering if it should still try at least a couple times even if it is revoked, or is being revoked a good indicator that the audit now irrelevant...

chappjc:
Working as intended, although I'm uncertain about the audit timeout during resolution.

client/core/trade.go
```diff
@@ -1470,6 +1483,17 @@ func (c *Core) finalizeSwapAction(t *trackedTrade, match *matchTracker, coinID,
 		errs.add("error storing swap details in database for match %s, coin %s: %v",
 			match.id, coinIDString(t.wallets.fromAsset.ID, coinID), err)
 	}
+
+	// With updated swap data, attempt match status resolution.
+	if needsResolution {
```
chappjc, line 1488: Now making sure match status and proof data updates happen before firing off the resolution goroutine.

client/core/trade.go
chappjc: Audit is only useful for the normal swap negotiation path, so I've concluded no, the audit is not needed for a revoked match, which will go to refund.  However, I believe there is an opportunity for an improvement that would require more work: https://github.com/decred/dcrdex/issues/829

client/core/trade.go
```diff
@@ -1796,79 +1828,112 @@ func (t *trackedTrade) refundMatches(matches []*matchTracker) (uint64, error) {
...
-	}
 	return nil
 }
 
 // auditContract audits the contract for the match and relevant MatchProof
-// fields are set. The changes are not saved to the database. auditContract must
-// be called with the trackedTrade.mtx locked.
-func (t *trackedTrade) auditContract(match *matchTracker, coinID []byte, contract []byte) error {
+// fields are set. The changes are not saved to the database. This may block for
+// a long period, and should be run in a goroutine. The trackedTrade mtx must
```
buck54321, line 1890: > This may block for a long period

Well, it'll block for as long as `timeout`, which is not necessarily long.

buck54321 (changes requested):
Tests well. 

I do feel like there is some new added complexity around lock handling in `status.go` that could have been avoided. For example, the new (exported?) getter methods in `trackedTrade` are only needed because we're not locking for all of `resolveConflictWithServerData` anymore. But there were only two paths (I think) that had the potential for long delays from mutex locking, and you've already stated that the correct solution is to run `auditContract` in a `goroutine` and unlock. New pattern seems racier.

client/core/trade.go
```diff
@@ -1796,79 +1815,112 @@ func (t *trackedTrade) refundMatches(matches []*matchTracker) (uint64, error) {
...
-	if err != nil {
-		return errs.addErr(err)
-	}
+	// Start searching for and audit the contract. This can take some time
+	// depending on node connectivity, so this is run in a goroutine. If the
+	// contract and coin (amount) are successfully validated, the matchTracker
+	// data are updated.
+	go func() {
+		// Search until it's known to be revoked.
+		err := t.auditContract(match, audit.CoinID, audit.Contract, 48*time.Hour)
```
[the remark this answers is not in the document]
chappjc: it's broadcast timeout *after swapconf*, which is harder to discern.  This time might as well be 1 year as this just breaks on revoke.

client/core/trade.go
```diff
@@ -677,6 +677,24 @@ func (t *trackedTrade) isActive() bool {
...
+	t.mtx.RLock()
+	defer t.mtx.RUnlock()
+	return match.MetaData.Status
+}
+
+func (t *trackedTrade) SelfRevokeMatch(match *matchTracker) {
+	t.mtx.Lock()
+	defer t.mtx.Unlock()
+	match.MetaData.Proof.SelfRevoked = true
+}
```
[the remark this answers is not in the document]
chappjc: Just because of the locking.  Will unexport.

But I don't see a way to run auditContract in a goroutine in the resolvers.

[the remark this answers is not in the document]
buck54321: In another goroutine?

chappjc:
I also snuck in a notification with the lats commit yesterday.  It's not pretty because the IDs are too long, but I think it needs at least the coin ID (utxo)

image

chappjc:
I discovered a very similar issue to the one addressed by this PR, which is the clients 'init' and 'redeem' requests are also blocking as they finalizeSwapAction and finalizeRedeemAction wait until the server responds.  A server response requires the server to search for the txn, so we have a similar failure mode as we saw with the audits.  We should address this without delay.

client/core/trade.go
```diff
@@ -1796,79 +1813,115 @@ func (t *trackedTrade) refundMatches(matches []*matchTracker) (uint64, error) {
...
+		Expiration: time.Now().Add(24 * time.Hour), // effectively forever
 		TryFunc: func() bool {
 			var err error
 			auditInfo, err = t.wallets.toWallet.AuditContract(coinID, contract)
-			if err != nil {
-				if errors.Is(err, asset.CoinNotFoundError) {
-					return wait.TryAgain
-				}
-				errChan <- err
+			if err == nil {
```
chappjc, line 1897: @buck54321 The hack I was using to simulate slow wallet AuditContract discovery of the coin is the following right after `AuditContract`:

```go
if tries < 13 {
	err = asset.CoinNotFoundError
}
```

client/core/core_test.go
```diff
@@ -5389,7 +5436,12 @@ func TestMatchStatusResolution(t *testing.T) {
 		f(resp)
 		return nil
 	})
+	// 2 matches resolved via contract audit: 2 synchronous updates, 2 async
+	rig.db.updateMatchChan = make(chan order.MatchStatus, 4)
 	tCore.authDEX(dc)
+	for i := 0; i < 4; i++ {
+		fmt.Println(<-rig.db.updateMatchChan)
```
buck54321, line 5445: Did you mean to leave the `Println`?

client/core/trade.go
```diff
@@ -1796,79 +1813,115 @@ func (t *trackedTrade) refundMatches(matches []*matchTracker) (uint64, error) {
...
-			errChan <- nil
+			if errors.Is(err, asset.CoinNotFoundError) {
+				// Didn't find it that time.
+				t.dc.log.Tracef("Still searching for counterparty's contract coin %v (%s) for match %v. ", contractID, contractSymb, match.id)
+				if t.matchIsRevoked(match) {
+					errChan <- ExpirationErr(fmt.Sprintf("match revoked while waiting to find counterparty contract coin %v (%s). "+
+						"Check your internet and wallet connections!", contractID, contractSymb))
+					return wait.DontTryAgain
+				}
+				if tries > 0 && tries%12 == 0 {
```
buck54321, line 1919: Love the notification metering.

chappjc:
With that rebase, I added one commit (https://github.com/decred/dcrdex/pull/819/commits/703767f2f2882b8e9f3e5c751336434da770af83) that adapts the recheck interval to the recently raised default broadcast timeout so that periodic ticks are every 90 seconds instead of every 4 minutes (12 minutes / 3).
The commit also tweaks the `checkTrades` flow so that the `tradeMtx` is never locked while a `trackedTrade` is locked. We don't want the trade map held unnecessarily, especially given the possibility for a trade to remain locked for a significant duration (e.g. `finalizeSwapAction`).

buck54321 (approved):

PR #821, opened 2020-11-07
core: allow filename for tls cert

buck54321:
Previously, caller had to load the file themselves. This allows a either the file name or the file contents. `Core` will load the file if a file name (`string`) is provided. `[]byte` will be assumed to be the certificate file contents and used unmodified.

This was originally part of #717.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"math"
 	"net"
 	"net/url"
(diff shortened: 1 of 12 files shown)


client/core/core.go
```diff
@@ -1708,20 +1709,19 @@ func (c *Core) isRegistered(host string) bool {
 // GetFee creates a connection to the specified DEX Server and fetches the
 // registration fee. The connection is closed after the fee is retrieved.
 // Returns an error if user is already registered to the DEX.
-func (c *Core) GetFee(dexAddr, certStr string) (uint64, error) {
+func (c *Core) GetFee(dexAddr string, certI interface{}) (uint64, error) {
```
chappjc, line 1712: GetFee docs should describe `certI`

client/core/core.go
```diff
@@ -1707,21 +1708,21 @@ func (c *Core) isRegistered(host string) bool {
 
 // GetFee creates a connection to the specified DEX Server and fetches the
 // registration fee. The connection is closed after the fee is retrieved.
-// Returns an error if user is already registered to the DEX.
-func (c *Core) GetFee(dexAddr, certStr string) (uint64, error) {
+// Returns an error if user is already registered to the DEX. A TLS certificate,
+// certI, can be provided as either a string filename, or []byte file contents.
+func (c *Core) GetFee(dexAddr string, certI interface{}) (uint64, error) {
```
chappjc, line 1713: @dajohi Please note this change for 0.2 if you are using `GetFee`

chappjc (approved):

Issue #779, opened 2020-10-23
app: show account ID for each DEX server somewhere


chappjc:
This will be the first step of account issue resolution when communicating with server operators.

chappjc:
@kevinstl Is taking this on.

amassarwi:
Done in #825

PR #825, opened 2020-11-09
ui: show account ID for each DEX server somewhere

kevinstl:
This pull request addresses: https://github.com/decred/dcrdex/issues/779

The account id is being displayed below the dex address on the wallets page.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -984,8 +984,8 @@ func (c *Core) Exchanges() map[string]*Exchange {
 		requiredConfs := uint32(dc.cfg.RegFeeConfirms)
 		infos[host] = &Exchange{
 			Host:          host,
-			AcctID:        dc.acct.ID().String(),
 			IsAuthed:      dc.acct.authed(),
+			AcctID:        dc.acct.ID().String(),
 			Markets:       dc.markets(),
 			Assets:        dc.assets,
 			FeePending:    dc.acct.feePending(),
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -32,6 +32,7 @@ import (
 	"decred.org/dcrdex/dex/msgjson"
 	"decred.org/dcrdex/dex/order"
 	"decred.org/dcrdex/dex/wait"
+	"decred.org/dcrdex/server/account"
 	"github.com/decred/dcrd/dcrec/secp256k1/v3"
 	"github.com/decred/dcrd/dcrec/secp256k1/v3/ecdsa"
 	"github.com/decred/go-socks/socks"
@@ -1827,8 +1828,9 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
 
 	// Prepare and sign the registration payload.
 	// The account ID is generated from the public key.
+	pkBytes := privKey.PubKey().SerializeCompressed()
 	dexReg := &msgjson.Register{
-		PubKey: privKey.PubKey().SerializeCompressed(),
+		PubKey: pkBytes,
 		Time:   encode.UnixMilliU(time.Now()),
 	}
 	regRes := new(msgjson.RegisterResult)
@@ -1860,8 +1862,10 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
 	}
 
 	// Pay the registration fee.
-	c.log.Infof("Attempting registration fee payment for %s of %d units of %s", regRes.Address,
-		regRes.Fee, regAsset.Symbol)
+	acctID := account.NewID(pkBytes)
+	c.log.Infof("Attempting registration fee payment to %s, account ID %v, of %d units of %s. "+
+		"Do NOT manually send funds to this address even if this fails.",
+		regRes.Address, acctID, regRes.Fee, regAsset.Symbol)
 	coin, err := wallet.PayFee(regRes.Address, regRes.Fee)
 	if err != nil {
 		return nil, newError(feeSendErr, "error paying registration fee: %v", err)
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -983,9 +983,15 @@ func (c *Core) Exchanges() map[string]*Exchange {
 		dc.cfgMtx.RLock()
 		dc.assetsMtx.RLock()
 		requiredConfs := uint32(dc.cfg.RegFeeConfirms)
+		// Set AcctID to empty string if not initialized.
+		acctID := dc.acct.ID().String()
+		var emptyAcctID account.AccountID
+		if dc.acct.ID() == emptyAcctID {
+			acctID = ""
+		}
 		infos[host] = &Exchange{
 			Host:          host,
-			AcctID:        dc.acct.ID().String(),
+			AcctID:        acctID,
 			Markets:       dc.markets(),
 			Assets:        dc.assets,
 			FeePending:    dc.acct.feePending(),
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -983,7 +983,8 @@ func (c *Core) Exchanges() map[string]*Exchange {
 		dc.assetsMtx.RLock()
 		requiredConfs := uint32(dc.cfg.RegFeeConfirms)
 		infos[host] = &Exchange{
-			Host:          host,
+			Host: host,
+			//AcctID:        dc.acct.ID().String(),
 			Markets:       dc.markets(),
 			Assets:        dc.assets,
 			FeePending:    dc.acct.feePending(),
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -985,7 +985,6 @@ func (c *Core) Exchanges() map[string]*Exchange {
 		requiredConfs := uint32(dc.cfg.RegFeeConfirms)
 		infos[host] = &Exchange{
 			Host:          host,
-			IsAuthed:      dc.acct.authed(),
 			AcctID:        dc.acct.ID().String(),
 			Markets:       dc.markets(),
 			Assets:        dc.assets,
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -985,7 +985,7 @@ func (c *Core) Exchanges() map[string]*Exchange {
 		infos[host] = &Exchange{
 			Host:          host,
 			AcctID:        dc.acct.ID().String(),
-			IsAuthed:      dc.acct.isAuthed,
+			IsAuthed:      dc.acct.authed(),
 			Markets:       dc.markets(),
 			Assets:        dc.assets,
 			FeePending:    dc.acct.feePending(),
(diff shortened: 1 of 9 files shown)


JoeGruffins (commented):
I think it would be good to add a label. It's not immediately obvious what the hex is. The address could maybe use one too in that case...

acctid

Could you also add the entry to the dexcctl help? 
Maybe above this line https://github.com/decred/dcrdex/blob/96ed03b323f2e4c67dbd3ea24159994e5f4b538f/client/rpcserver/handlers.go#L718
Same depth as the below "market" and should look something like
```
        "acctID" (string):  The client's account ID associated with this DEX.
```

buck54321:
This is kind of an odd place to display this information. Could you instead list them out on the settings page right here?

image

client/core/core.go
```diff
@@ -975,7 +975,7 @@ func (c *Core) Exchanges() map[string]*Exchange {
 		infos[host] = &Exchange{
 			Host:          host,
 			AcctID:        dc.acct.ID().String(),
-			IsAuthed:      dc.acct.isAuthed,
+			IsAuthed:      dc.acct.authed(),
```
chappjc, line 977: Did the race detector find this?

chappjc:
Would you please sneak in the following diff (just manually edit since line numbers will be off):

```diff
diff --git a/client/core/core.go b/client/core/core.go
index 874d2d4c..1839bea5 100644
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -32,6 +32,7 @@ import (
        "decred.org/dcrdex/dex/msgjson"
        "decred.org/dcrdex/dex/order"
        "decred.org/dcrdex/dex/wait"
+       "decred.org/dcrdex/server/account"
        "github.com/decred/dcrd/dcrec/secp256k1/v3"
        "github.com/decred/dcrd/dcrec/secp256k1/v3/ecdsa"
        "github.com/decred/go-socks/socks"
@@ -1824,8 +1825,9 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
 
        // Prepare and sign the registration payload.
        // The account ID is generated from the public key.
+       pkBytes := privKey.PubKey().SerializeCompressed()
        dexReg := &msgjson.Register{
-               PubKey: privKey.PubKey().SerializeCompressed(),
+               PubKey: pkBytes,
                Time:   encode.UnixMilliU(time.Now()),
        }
        regRes := new(msgjson.RegisterResult)
@@ -1857,8 +1859,10 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
        }
 
        // Pay the registration fee.
-       c.log.Infof("Attempting registration fee payment for %s of %d units of %s", regRes.Address,
-               regRes.Fee, regAsset.Symbol)
+       acctID := account.NewID(pkBytes)
+       c.log.Infof("Attempting registration fee payment to %s, account ID %v, of %d units of %s. "+
+               "Do NOT manually send funds to this address even if this fails.",
+               regRes.Address, acctID, regRes.Fee, regAsset.Symbol)
        coin, err := wallet.PayFee(regRes.Address, regRes.Fee)
        if err != nil {
                return nil, newError(feeSendErr, "error paying registration fee: %v", err)

```

client/core/core.go
kevinstl: Yes it did.

client/webserver/site/src/html/settings.tmpl
```diff
@@ -17,6 +17,13 @@
         Show pop-up notifications
       </label>
     </div>
+    <div id="dexBox">
+        <div class="bg2 px-2 py-1" id="dexInfo">
+            Dex Address: <span class="dexInfoHost"></span><br>
```
buck54321, line 22: CSS class names don't use camelCase.

client/webserver/site/src/js/settings.js
```diff
@@ -61,6 +63,27 @@ export default class SettingsPage extends BasePage {
     })
   }
 
+  async showDexInfo (page) {
+    const dexBox = page.dexBox
+    const dexInfo = page.dexInfo.cloneNode(true)
+    const dexInfoAcct = page.dexInfo.querySelector('.dexInfoAcct').cloneNode(true)
+
```
buck54321, line 70: If you have multiple dexes, you will be creating multiple elements with the same `id`, which is not allowed. Check out e.g. how I'm using templates at https://github.com/decred/dcrdex/blob/cd855bef19c85f010fbc0f5385ca483b4eeb6490/client/webserver/site/src/js/markets.js#L97

client/webserver/site/src/js/settings.js
```diff
@@ -61,6 +63,27 @@ export default class SettingsPage extends BasePage {
...
+    Doc.hide(dexBox)
+    page.dexInfo.remove()
+    dexInfo.removeAttribute('id')
+    dexInfo.querySelector('.dexInfoAcct').remove()
+
+    for (const [host, xc] of Object.entries(app.user.exchanges)) {
+      dexInfo.querySelector('.dexInfoHost').textContent = host
+      if (xc.isAuthed) {
+        dexInfoAcct.querySelector('.dexInfoAcctID').textContent = xc.acctID
```
buck54321, line 79: Instead of using `querySelector` and CSS classes, please use `data-tmpl` attributes and `Doc.tmplElement`. You can find examples all over.

client/webserver/site/src/js/settings.js
```diff
@@ -61,6 +63,27 @@ export default class SettingsPage extends BasePage {
     })
   }
 
+  async showDexInfo (page) {
```
buck54321, line 66: I've left some comments on the implementation of this function, but really, I don't think you should be using javascript for this at all. Do it all in the template.

client/webserver/site/src/html/settings.tmpl
buck54321, line 22: You should just iterate through the `.UserInfo.Exchanges` and add the elements here.

buck54321 (changes requested):
image

1. No special background color.
2. No additional indentation for the text.
3. No horizontal rule between dex list and add a dex button.
4. Please generate a screenshot for multiple servers listed. My guess is that this scheme is not going to work very well when there is more than one server.

client/webserver/site/src/js/settings.js
kevinstl: Everything is being done in template now.

kevinstl:
> 4. Please generate a screenshot for multiple servers listed. My guess is that this scheme is not going to work very well when there is more than one server.

image

I think it will all work and be much simpler now that it's being done in the template.

client/rpcserver/handlers_test.go
```diff
@@ -476,6 +476,8 @@ func TestHandleRegister(t *testing.T) {
 const exchangeIn = `{
   "https://127.0.0.1:7232": {
     "host": "https://127.0.0.1:7232",
+    "isAuthed": true,
+	"acctID": "edc7620e02",
```
JoeGruffins, line 480: not super important but indent is off

JoeGruffins (commented):
I tried this out on the mainnet dex and clicking on settings isn't doing anything. I see in console that it is throwing "500 Internal Server Error." Can you confirm this behavior?

kevinstl:
> I tried this out on the mainnet dex and clicking on settings isn't doing anything. I see in console that it is throwing "500 Internal Server Error." Can you confirm this behavior?

I've only ran simnet so far. I'll get mainnet setup so I can test it.

JoeGruffins (approved):
Doh! It looks like the problem was that I forgot to rebuild the dexc binary. It's working mainnet. Sorry about that!

chappjc:
@buck54321 Will have to comment on refreshing the list with js, but I suspect that is expanding the scope of this PR too much.

client/webserver/site/src/html/settings.tmpl
```diff
@@ -17,12 +17,30 @@
...
       </label>
     </div>
-    <div{{if not .UserInfo.Authed}} class="d-hide"{{end}}>
-      <p>
+    <div>
+      <div {{if eq (len .UserInfo.Exchanges) 0}} class="d-hide"{{end}}>
+        <div>
+          {{range $host, $xc := .UserInfo.Exchanges}}
+            <div>
+              Dex Address: {{$host}}
```
chappjc, line 25: Dex -> DEX

client/webserver/site/src/html/settings.tmpl
```diff
@@ -17,12 +17,30 @@
...
+              {{if $xc.IsAuthed}}
```
chappjc, line 26: I'm not sure we want to add IsAuthed as a field to the Exchanges struct, at least not for this display logic.  Why not display AcctID regardless?

chappjc (changes requested):
I'm fine with the display although I expect @buck54321 has some refinements in mind.

But I'd like to not add extra fields to the exported types like `Exchanges` unless really necessary, so `IsAuthed` should probably go.  There will be a non-zero account ID when the user decrypts the account private key on login to Core, while "auth" relates to the success of a `'connect'` request.

kevinstl:
Here is how the latest version looks:

image

kevinstl:
Here is how the latest version looks:

image

chappjc:
That's how it looks before login, while after login the account IDs become non-zeros?  The `AccountID` type can be given an `IsZero` method that the template can call to conditionally show the account ID or "<login to show Account ID>" or something similar.

kevinstl:
> That's how it looks before login, while after login the account IDs become non-zeros? The `AccountID` type can be given an `IsZero` method that the template can call to conditionally show the account ID or `"Account ID: <login to show>"` or something similar.

Yes:

image

kevinstl:
> That's how it looks before login, while after login the account IDs become non-zeros? The `AccountID` type can be given an `IsZero` method that the template can call to conditionally show the account ID or `"Account ID: <login to show>"` or something similar.

I'll push something up with a dynamic display of Account ID....

chappjc:
Gotcha, so to clarify the suggestion:

```diff
diff --git a/server/account/account.go b/server/account/account.go
index a6dbe215..6e60b393 100644
--- a/server/account/account.go
+++ b/server/account/account.go
@@ -18,6 +18,7 @@ import (
 var (
        HashFunc = blake256.Sum256
        century  = time.Hour * 24 * 365 * 100
+       zeroAcct AccountID
 )
 
 const (
@@ -45,6 +46,10 @@ func (aid AccountID) MarshalJSON() ([]byte, error) {
        return json.Marshal(aid.String())
 }
 
+func (aid AccountID) IsZero() bool {
+       return aid == zeroAcct
+}
+
```

and you could then do something like:

```
Account ID: <span class="fs12">{{if $xc.AcctID.IsZero}}[login to show]{{else}}{{$xc.AcctID}}{{end}}</span>
```

kevinstl:
Latest appearance when not logged in:

image

chappjc (approved):

chappjc (approved):
Working as expected and the settings page looks good to me.

Issue #824, opened 2020-11-08
match proof issues


chappjc:
There seems to be a bug in what the client includes in epoch orders when computing and comparing match proofs.

```
2020-11-08 17:06:00.095 [TRC] CORE: notify: |DATA| (epoch) - Index: 160485516
2020-11-08 17:06:00.103 [DBG] CORE[dcr_btc][book]: Validating match proof note with 8 preimages and 0 misses.
2020-11-08 17:06:00.103 [ERR] CORE: Route 'match_proof' notification handler error (DEX dex.decred.org:7232): match proof note references 8 orders, but local epoch queue has 9
2020-11-08 17:06:10.226 [TRC] CORE: notify: |DATA| (epoch) - Index: 160485517
2020-11-08 17:06:10.226 [DBG] CORE[dcr_btc][book]: Validating match proof note with 10 preimages and 0 misses.
2020-11-08 17:06:10.226 [ERR] CORE: Route 'match_proof' notification handler error (DEX dex.decred.org:7232): match proof note references 10 orders, but local epoch queue has 9
```

It appears to be a sequence issue as epoch A above includes 9 orders in client's queue but server had 8, while epoch B had 9 orders in the client's queue but server had 9.

Fortunately this does not break anything, but there is a bug.

chappjc:
I've seen the following depth chart anomaly twice, and I suspect it is a related issue, but it could be something completely related?

Leaving the markets page open for a long while, returning to this:

![image](https://user-images.githubusercontent.com/9373513/98888365-7ea17100-245d-11eb-9235-93f143f6bfff.png)

F5:

![image](https://user-images.githubusercontent.com/9373513/98888376-83febb80-245d-11eb-942d-d8d3d457d848.png)


Both the depth chart and the orders lists have indicate something strange with the order feed.

**v0.1.1 release**

chappjc:
I've seen the following depth chart anomaly twice, and I suspect it is a related issue, but it could be something completely related?

Leaving the markets page open for a long while, returning to this:

![image](https://user-images.githubusercontent.com/9373513/98888603-e788e900-245d-11eb-9dc0-b57bf07aab5d.png)

F5:

![image](https://user-images.githubusercontent.com/9373513/98888612-ebb50680-245d-11eb-8e97-04b4894e3977.png)

Both the depth chart and the orders lists have indicate something strange with the order feed.

**v0.1.1 release**

chappjc:
[...]

Issue #628, opened 2020-08-24
client/core: don't fuss about failed match proof when starting mid-epoch


chappjc:
[ERR] CORE: route 'match_proof' notification handler error: match proof note references 2 orders, but local epoch queue has 0

The client (dexc) started up and then subscribed to that market's order book in the middle of an epoch that already had orders.  When the epoch closed and server sent a match proof, it referenced the orders that the client did not know about.  This error can simply be ignored.

chappjc:
Resolved by https://github.com/decred/dcrdex/pull/834

PR #834, opened 2020-11-13
client/core: epoch tracking right

chappjc:
match_proof notifications are not in sequence with other book messages and should not be used to indicate a transition from one epoch to another.
Since match_proof lags epoch close by up to preimage request timeout and some additional latency, the `OrderBook` needs to track several epochs for the purposes of evaluating these lagging match proofs.

Resolves https://github.com/decred/dcrdex/issues/824 and https://github.com/decred/dcrdex/issues/628

### DIFF
--- a/client/core/bookie.go
+++ b/client/core/bookie.go
@@ -698,7 +698,7 @@ func handleEpochOrderMsg(c *Core, dc *dexConnection, msg *msgjson.Message) error
 	}
 
 	if dc.setEpoch(note.MarketID, note.Epoch) {
-		c.refreshUser()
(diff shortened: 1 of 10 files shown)


chappjc:
This definitely resolved https://github.com/decred/dcrdex/issues/824, but not the chart and orders list, so those must be unrelated issues.  New issue for the zero-quantity orders getting left in the chart and tables: https://github.com/decred/dcrdex/issues/835

chappjc:
Been running for several days on a mainnet client with no trouble.

buck54321 (approved):

PR #837, opened 2020-11-16
ui: more chart interactivity

buck54321:
1. Indicators on the depth chart for the user's orders. When the mouse hovers near the indicator, the order is highlighted in the "Your Orders" table. Conversely, when the mouse hovers over a row in the "Your Orders" table, the indicator is highlighted on the chart.
2. Move the legend and hover info to the top. 
3. When a rate is entered in the order form for a limit order, display a line indicator at that rate on the depth chart.
4. When the user hovers over an order in the buy/sell tables, display an indicator at that rate on the depth chart.
5. Last zoom level is saved across markets and reloads.
6. Display the total buy/sell volume.

Can test with the [demo server](https://github.com/decred/dcrdex/wiki/Test-App-Server).


chappjc:
This is great!

What's the "X" button supposed to do at the top right here:

image


Clicking it has no effect.

buck54321:
Oh. I meant to remove that. I just wanted to see how a button might work there, for when there are more options and chart types.

JoeGruffins (approved):
LGTM

chappjc (approved):

PR #838, opened 2020-11-16
update rpcclient, dcrutil, blockchain/stake, hdkeychain

chappjc:
This updates rpcclient from v5 to v6.

This also completes the update of dcrutil/v2 to v3, removing the conversion helpers.

Also updated are blockchain/stake and hdkeychain.  This is the last of the older dcrd modules.


client/asset/dcr/dcr.go
```diff
@@ -131,26 +131,26 @@ var (
...
-	GetRawTransactionVerbose(txHash *chainhash.Hash) (*chainjson.TxRawResult, error)
-	LockUnspent(unlock bool, ops []*wire.OutPoint) error
-	GetRawChangeAddress(account string, net dcrutil.AddressParams) (dcrutil.Address, error)
-	GetNewAddressGapPolicy(string, rpcclient.GapPolicy, dcrutil.AddressParams) (dcrutil.Address, error)
-	DumpPrivKey(address dcrutil.Address, net [2]byte) (*dcrutil.WIF, error)
-	GetTransaction(txHash *chainhash.Hash) (*walletjson.GetTransactionResult, error)
-	WalletLock() error
-	WalletPassphrase(passphrase string, timeoutSecs int64) error
+	EstimateSmartFee(ctx context.Context, confirmations int64, mode chainjson.EstimateSmartFeeMode) (float64, error)
+	GetBlockChainInfo(context.Context) (*chainjson.GetBlockChainInfoResult, error)
```
davecgh, line 135: Almost all the others have the variable named, but this one doesn't.

client/asset/dcr/dcr.go
```diff
@@ -131,26 +131,26 @@ var (
...
+	GetNewAddressGapPolicy(context.Context, string, dcrwallet.GapPolicy) (dcrutil.Address, error)
+	DumpPrivKey(ctx context.Context, address dcrutil.Address) (*dcrutil.WIF, error)
+	GetTransaction(ctx context.Context, txHash *chainhash.Hash) (*walletjson.GetTransactionResult, error)
+	WalletLock(context.Context) error
+	WalletPassphrase(ctx context.Context, passphrase string, timeoutSecs int64) error
 	Disconnected() bool
-	RawRequest(method string, params []json.RawMessage) (json.RawMessage, error)
-	WalletInfo() (*walletjson.WalletInfoResult, error)
+	RawRequest(ctx context.Context, method string, params []json.RawMessage) (json.RawMessage, error)
+	WalletInfo(context.Context) (*walletjson.WalletInfoResult, error)
```
davecgh, line 153: Same here.  Inconsisten.

server/asset/btc/btc.go
```diff
@@ -239,15 +240,38 @@ func (btc *Backend) Redemption(redemptionID, contractID []byte) (asset.Coin, err
...
+	}()
+
+	var utxo *UTXO
+	select {
+	case <-ctx.Done():
+		err = asset.ErrRequestTimeout
+	case r := <-res:
+		utxo, err = r.UTXO, r.error
+	}
+
```
chappjc, line 270: This may be pure madness, since the rpc calls aren't actually canceled, just left to rot.  It's probably better not to do this and instead ignore the input arg, naming it `_ context.Context` to make it clear there's no actual timeout or cancellation possible with the BTC backend

server/asset/btc/btc.go
```diff
@@ -81,6 +81,7 @@ type btcNode interface {
 // cache of block data for quick lookups. Backend implements asset.Backend, so
 // provides exported methods for DEX-related blockchain info.
 type Backend struct {
+	ready chan struct{}
```
chappjc, line 84: This is a downside of using the Backend subsystem's context to also cancel externally-initiated RPCs on subsystem shutdown.  I didn't want to plumb every method with a context but maybe...

server/asset/dcr/dcr.go
```diff
@@ -69,20 +69,31 @@ const (
...
+	return err
 }
 
 // Backend is an asset backend for Decred. It has methods for fetching output
 // information and subscribing to block updates. It maintains a cache of block
 // data for quick lookups. Backend implements asset.Backend, so provides
 // exported methods for DEX-related blockchain info.
 type Backend struct {
+	ctx   context.Context
+	ready chan struct{}
```
chappjc, line 96: This is a downside of using the Backend subsystem's context to also cancel externally-initiated RPCs on subsystem shutdown.  I didn't want to plumb every method with a context but maybe...

chappjc:
I'm plannign to follow this up with changes to SyncStatus, checking wallet connectivity on order placement, and other things related to connectivity that are going undetected presently.

server/asset/btc/btc.go
chappjc: I've removed these acrobatics.  We can plan to use *decred's* rpcclient for BTC with all requests going through `RawRequest`.

JoeGruffins (approved):
Tested quite a bit and everything fine.

server/asset/dcr/dcr.go
```diff
@@ -127,7 +138,7 @@ func NewBackend(configPath string, logger dex.Logger, network dex.Network) (*Bac
 
 	// Ensure the network of the connected node is correct for the expected
 	// dex.Network.
-	net, err := dcr.client.GetCurrentNet()
+	net, err := dcr.client.GetCurrentNet(context.TODO())
```
buck54321, line 141: How about a `context.WithTimeout(context.Background(), 5 * time.Second)` to limit how long the constructor can hang.

server/asset/dcr/dcr.go
buck54321: Can solve this by converting `Backend` from a `dex.Runner` to a `dex.Connector`.

server/asset/dcr/dcr.go
```diff
@@ -926,53 +914,53 @@ func (dcr *Backend) getUnspentTxOut(txHash *chainhash.Hash, vout uint32) (*chain
...
-		return nil, nil, nil, fmt.Errorf("GetRawTransactionVerbose for txid %s: %w", txHash, err)
+		return nil, nil, nil, fmt.Errorf("GetRawTransactionVerbose for txid %s: %w", txHash, translateRPCCancelErr(err))
 	}
 	return txOut, verboseTx, pkScript, nil
 }
 
 // Get the block information, checking the cache first. Same as
 // getDcrBlock, but takes a string argument.
-func (dcr *Backend) getBlockInfo(blockid string) (*dcrBlock, error) {
+func (dcr *Backend) getBlockInfo(ctx context.Context, blockid string) (*dcrBlock, error) {
```
buck54321, line 931: Why not use `dcr.ctx` instead of passing the context as an argument for these unexported functions throughout?

server/asset/dcr/dcr_test.go
```diff
@@ -765,7 +765,7 @@ func TestUTXOs(t *testing.T) {
 	// Set the value of this one.
 	txout.Value = 2.0
 	// There is no block info to add, since this is a mempool transaction
-	utxo, err := dcr.utxo(txHash, msg.vout, nil)
+	utxo, err := dcr.utxo(context.Background(), txHash, msg.vout, nil)
```
buck54321, line 768: I think [in other tests](https://github.com/decred/dcrdex/blob/fb6f3eaa7dae7d043aae8ceb5771638554dd5d1c/client/asset/dcr/dcr_test.go#L494-L503), we've created a `context.WithCancel` in `TestMain` for these purposes.

buck54321 (changes requested):

server/asset/dcr/dcr.go
chappjc: This traces back to the Confirmations and FundingCoin methods where the caller wants to specify timeout.  Same reason for the getMainchainDcrBlock method.

server/asset/dcr/dcr.go
chappjc: Had crossed me mind, since that's now it's done on client side, but I ran away screaming because it doesn't fit with the `server/dex`'s subsystem stack.  I'll see how to work that out.

server/asset/dcr/dcr.go
chappjc: With the change from Runner => Connector just pushed, no more RPCs are in the constructor.

server/asset/dcr/dcr_test.go
chappjc: `testBackend()` pretty  much does that for the tests individually.  But since this isn't a "live" test, these context go nowhere anyway.

PR #842, opened 2020-11-18
client/multi: add max order estimation

buck54321:
Get maximum order estimations for a given market and side based on the wallet balance and, in the case of buy orders, the rate in the rate input field. The data is sourced directly from the wallet, because the limits change based on available funding coins.

The data is shown in the UI as a small message above the rate field. When you click on the label, the quantity fields are pre-populated with the max order.

image

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -741,8 +741,8 @@ func (btc *ExchangeWallet) RedemptionFees() (uint64, error) {
 // dex.Bytes should be appended to the redeem scripts collection for coins with
 // no redeem script.
 func (btc *ExchangeWallet) FundOrder(ord *asset.Order) (asset.Coins, []dex.Bytes, error) {
-	btc.log.Debugf("Attempting to fund order for %d %s, maxFeeRate = %d, max swaps = %d",
-		ord.Value, btc.walletInfo.Units, ord.DEXConfig.MaxFeeRate, ord.MaxSwapCount)
+	btc.log.Debugf("Attempting to fund order for %.8f %s, maxFeeRate = %d, max swaps = %d",
+		toBTC(ord.Value), btc.walletInfo.Units, ord.DEXConfig.MaxFeeRate, ord.MaxSwapCount)
 
 	if ord.Value == 0 {
 		return nil, nil, fmt.Errorf("cannot fund value = 0")
@@ -761,7 +761,7 @@ func (btc *ExchangeWallet) FundOrder(ord *asset.Order) (asset.Coins, []dex.Bytes
 	}
 	if avail < ord.Value {
 		return nil, nil, fmt.Errorf("insufficient funds. %.8f requested, %.8f available",
-			btcutil.Amount(ord.Value).ToBTC(), btcutil.Amount(avail).ToBTC())
+			toBTC(ord.Value), toBTC(avail))
 	}
 
 	sum, size, coins, fundingCoins, redeemScripts, spents, err := btc.fund(ord.Value, ord.MaxSwapCount, utxos, ord.DEXConfig)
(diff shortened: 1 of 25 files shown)


client/asset/btc/btc.go
```diff
@@ -660,6 +660,79 @@ func (btc *ExchangeWallet) feeRateWithFallback(confTarget uint64) uint64 {
...
+			if avail >= reqFunds+extraFees {
+				return &asset.OrderEstimate{
+					Lots:          lots,
+					Value:         val,
+					MaxFees:       maxFees + extraFees,
+					EstimatedFees: estFees + extraEstFees,
+					Locked:        val + maxFees + extraFees,
+				}, nil
+			}
+		}
```
chappjc, line 708: What if `avail < reqFunds+extraFees`?  It should return `&asset.OrderEstimate{}` right?  Instead it seems to go with the no split txn estimate.

client/asset/btc/btc.go
```diff
@@ -690,18 +763,48 @@ func (btc *ExchangeWallet) FundOrder(ord *asset.Order) (asset.Coins, []dex.Bytes
...
+			return nil, nil, err
+		} else if split {
+			return splitCoins, []dex.Bytes{nil}, nil // no redeem script required for split tx output
+		} else {
+			return splitCoins, redeemScripts, nil // splitCoins == coins
+		}
+	}
+
+	btc.log.Infof("Funding %d %s order with coins %v worth %d",
+		ord.Value, btc.walletInfo.Units, coins, sum)
```
chappjc, line 784: How about we use ToBTC on the Value for the log since we're in the backend?

client/asset/dcr/dcr.go
```diff
@@ -609,6 +609,85 @@ func (dcr *ExchangeWallet) feeRateWithFallback(confTarget uint64) uint64 {
...
+	for _, utxo := range utxos {
+		avail += toAtoms(utxo.rpc.Amount)
+	}
+
+	// Start by attempting max lots with no fees.
+	lots := avail / lotSize
+	for lots > 0 {
+		val := lots * lotSize
+		sum, size, _, _, _, err := dcr.tryFund(utxos, orderEnough(val, lots, nfo))
+		// The only failure mode of btc.fund is when there is not enough funds,
```
chappjc, line 636: **btc**.fund

chappjc:
Need a better way to handle no fee estimate available:

image

client/asset/btc/btc.go
buck54321: The estimate is like a dry run. Falling back to non-split is what will happen when they order.

chappjc: OK, I honestly did not recall that it falls back to non-split if needed.  That actually sounds undesirable as it violates the user's settings.

buck54321: It only falls back when the excess is not enough to cover the fees associated with the split. In that case, we are doing them a favor by not using a split transaction. Overall cost is lower. Split only makes sense when the excess is substantial.

chappjc: Ah, I was aware of that.  Dunno how I got that mixed up.  All good.

chappjc:
Can you adapt this with an extra Core method(s) to get order estimates (with redeem fees) for an arbitrarily sized order too?

As an aside, I'm wondering if the host is really necessary as an input for these methods.

buck54321:
> Can you adapt this with an extra Core method(s) to get order estimates (with redeem fees) for a specified order quantity too?

Yep. The intention is to also support estimates at order time in the future.

> As an aside, I'm wondering if the host is really necessary as an input for these methods.

Assuming you mean the `core` methods, then yes. We need the lot size.

chappjc (approved):

PR #849, opened 2020-11-23
.dexcctl folder

chappjc:
Look for the dexcctl.conf file in ~/.dexcctl (or appdata(dexc, false)) instead of ~/.dexc.

dexcctl will also look for rpc.cert in .dexcctl if not specified on the command line, but if it is not found there, it will check in ~/.dexc.

### DIFF
--- a/client/cmd/dexcctl/config.go
+++ b/client/cmd/dexcctl/config.go
@@ -115,6 +115,9 @@ func configure() (*config, []string, bool, error) {
 		if !fileExists(cfg.RPCCert) { // Then in ~/.dexc
 			cfg.RPCCert = cleanAndExpandPath(filepath.Join(dexcAppDir, defaultRPCCertFile))
 		}
+	} else {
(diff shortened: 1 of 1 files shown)


buck54321 (approved):

client/cmd/dexcctl/config.go
```diff
@@ -110,16 +110,17 @@ func configure() (*config, []string, bool, error) {
...
+			cfg.RPCCert = cleanAndExpandPath(filepath.Join(dexcAppDir, defaultRPCCertFile))
+		}
 	}
 
 	if cfg.RPCAddr == "" {
 		cfg.RPCAddr = defaultRPCAddr
 	}
 
-	// Handle environment variable expansion in the RPC certificate path.
-	cfg.RPCCert = cleanAndExpandPath(cfg.RPCCert)
```
chappjc, line 121: I just realized this is still needed in case the user *does* specify something and the branch above isn't followed.

PR #850, opened 2020-11-23
[0.1.3 backport] client/core: epoch tracking right

chappjc:
Backport of https://github.com/decred/dcrdex/pull/834 for 0.1.3

match_proof notifications are not in sequence with other book messages
and should not be used to indicate a transition from one epoch to
another. Since match_proof lags epoch close by up to preimage request
timeout and some additional latency, the OrderBook needs to track
several epochs for the purposes of evaluating these lagging match
proofs.

unexport some internal orderbook types and functions

PR #851, opened 2020-11-23
core: fix recursive wallet rlock

buck54321:


client/core/wallet.go
```diff
@@ -105,7 +105,7 @@ func (w *xcWallet) state() *WalletState {
 	return &WalletState{
 		Symbol:       unbip(w.AssetID),
 		AssetID:      w.AssetID,
-		Open:         w.locallyUnlocked(),
+		Open:         len(w.encPW) == 0 || len(w.pw) > 0,
```
chappjc, line 108: The recursive RLock shouldn't be a problem though, right?  I haven't made any effort to diagnose https://github.com/decred/dcrdex/pull/850/checks?check_run_id=1443539618#step:5:406 however.

buck54321: I think it can, yes. The other place the mutex is hanging is in a call to `setBalance`, which is a write-lock. If the order is rlock1 -> lock -> rlock2, it'll hang. You'd have to sneak that lock in there just right, but an audit of the places the `xcWallet.mtx` is locked indicated that this was the only possible place something could happen.

chappjc: Ah, that's right if the lock is right between the two rlocks.  Good catch.

chappjc (approved):

PR #856, opened 2020-11-25
swapper resume from db

chappjc:
This removes the swap state gob files, loading necessary from DB on Swapper startup.

**server/db**

Add `SwapArchiver.ActiveSwaps` (impl. by `(*Archiver).ActiveSwaps`) to retrieve full details of all active swaps across all markets.  `Swapper`'s constructor uses this to resume active swaps.

Remove all the swap state hash get/set functions.

The `meta` table and it's `state_hash` column still exist, but I think I'm going to entirely drop this table and follow up with a PR that re-adds the meta table with a versioning scheme.

**server/swap**

Rip out:

- All of the swap state/gob stuff.
- The `orderSwapTracker`, which was used to infer order "completion" for the cancellation rate computations.  The market now has these order fill/settlement/success responsibilities.  A `swapDone` callback is provided to the `Swapper` from the consumer (DEX manager) to do this.
- The `offBook` argument of `Negotiate`, which was used with the `orderSwapTracker` junk.
- The the live waiter tracking.  Clients retry init/redeem requests now, so jumping through hoops to recreate live coin waiters on swapper startup is not necessary.

The Swapper constructor now loads all active matches from the DB and reconstructs all of the `matchTrackers`.  This uses the new `Backend.CoinConfTime` method for the `SwapConfirmTime`, which is not stored in the DB.  The match request time field of `matchTracker` is also not recorded, so this is guessed on resume as the epoch close time plus a minute, which at worst gives the maker a little extra breathing room when resuming in `NewlyMatched` status.

**server/market**

Assume the responsibility of deciding when an order is successfully "completed" as per the cancellation rate definition.  This is facilitated by the `SwapDone` method that the DEX manager bridges with the `Swapper`.

The `SwapDone` method also performs the same unbooking that the Swapper previous requested via the `unbookHook`, now removed.

**server/dex**

Remove the unbook hook, replacing it with a `swapDone` dispatcher.

**server/auth**

`Sign` no longer returns an error since the secp256k1 changes recently eliminated the error return.


JoeGruffins (approved):
Made a lot of orders and swaps, suspending and just closing the server at different steps, then comparing the orderbooks and matches returned by the admin console, all returns with expected values. Am able to continue trading. Matches completed while the server is down are updated when it comes back online. Can't find anything wrong. And wow is the webserver ui easy to use now. Holy crap. DEX is legit.

chappjc:
Just a comment that the second commit (https://github.com/decred/dcrdex/pull/856/commits/0e5c64b9971497b987ce1bfe0da1aa2d3306ba20) is only _touching_ on a mostly-unrelated can of worms.  That is, the init and redeem requests sent by the client are synchronous, blocking in `finalize{Swap,Redeem}Action` while they wait for a response from server.  That needs to change.  Also, this commit addresses an issue with pending requests and reconnects: I realized that when these requests are sent and the server starts a coinwaiter or whatever other processing, and then the connection is lost before getting a response, the `signAndRequest` call will never get a response, just waiting until timeout, which is unfortunately very long (broadcast timeout).  This commit forcibly expires any request handlers on reconnect causing `signAndRequest` to return.  @buck54321 We've waffled on the appropriate timeouts for these requests and I've steered toward just as long as allowable, but given (1) the blocking issue and (2) the fact that core will retry `finalize{Swap,Redeem}Action`, I think the timeouts for `signAndRequest` can be shorter, but I still don't know what the right timeout is.  If we make these things asynchronous, it might be a moot issue though.
Finally, in `finalize{Swap,Redeem}Action`, I think the `match.SetStatus` and update of `proof.{T,M}akerStatus` should be done on broadcast (e.g. in `swapMatchGroup`), esp. since on the first call after bcast you could be sitting at `signAndRequest` without having stored the coin you just broadcasted, plus it's redundant for repeated calls.  The only thing we need to update the match with after `signAndRequest` is `auth.{InitSig,InitStamp}` if we got a response.

server/market/market.go
```diff
@@ -263,10 +276,29 @@ ordersLoop:
...
+		settling[match.Taker] += match.Quantity
+		settling[match.Maker] += match.Quantity
+		// Note: we actually don't want to bother with matches for orders that
+		// were canceled or had at-fault match failures, since including them
+		// give that user another shot to get a successfully "completed" order
+		// if they complete these remaining matches, but it's OK. We'd have to
+		// query these order statuses, and look for at-fault match failures
+		// involving them, so just give the user the benefit of the doubt.
+	}
+	log.Infof("Track for %d orders with %d active matches.", len(settling), len(activeMatches))
```
buck54321, line 295: ~~Track for~~ -> Tracking ?

server/swap/swap.go
```diff
@@ -554,38 +358,66 @@ func (s *Swapper) ChainsSynced(base, quote uint32) (bool, error) {
...
 			ss.swap = swap
 			ss.swapTime = encode.UnixTimeMilli(ssd.SwapTime)
-		}
 
-		if ssd.SwapConfirmTime != 0 {
-			ss.swapConfirmed = encode.UnixTimeMilli(ssd.SwapConfirmTime)
+			ss.swapConfirmed, err = swapAsset.Backend.CoinConfTime(swapCoin, int64(swapAsset.SwapConf))
+			if err != nil {
+				log.Warnf("No swap confirmed time for %v: %v", swap, err)
+			}
```
buck54321, line 420: Since this probably will not work because of the policies regarding block times, I think we could just give the client the benefit here too and use the current time. Market restarts are a messy business, so the more leniency we can give them, the fewer problems we'll encounter is my guess.

server/swap/swap.go
```diff
@@ -601,172 +433,125 @@ func (s *Swapper) restoreState(state *State, allowPartial bool) error {
...
-			},
-			takerStatus: &swapStatus{
-				swapAsset:   makerRedeemAsset, // mtd.TakerStatus.SwapAsset
-				redeemAsset: makerSwapAsset,   // mtd.TakerStatus.RedeemAsset
-			},
+		// Load the maker's order.LimitOrder and taker's order.Order.
+		taker, _, err := s.storage.Order(sd.MatchData.Taker, sd.Base, sd.Quote)
+		if err != nil {
+			log.Errorf("Failed to load taker order: %v", err)
+			continue
```
buck54321, line 451: Noting that this is a departure from normal because the `Order` we'll be tracking in `Swapper` is not backed by the same underlying being tracked in `Market`. Is that okay?  I think the only thing we'd need to worry about is if we were somehow relying on the `Filled/Remaining` quantities to remain accurate, and a quick scan seems to say that we are not.

server/swap/swap_test.go
```diff
@@ -1934,27 +1898,14 @@ func TestCancel(t *testing.T) {
 	}
 }
 
-func TestState(t *testing.T) {
+/*
```
buck54321, line 1901: Are you keeping these tests around to adapt them?

buck54321 (changes requested):
> `+803 −1,237`

Nice. So much cleaner now and roles are much clearer.

server/asset/btc/btc.go
```diff
@@ -337,6 +337,52 @@ func (btc *Backend) BlockChannel(size int) <-chan *asset.BlockUpdate {
...
+		if errors.As(err, &rpcErr) && rpcErr.Code == btcjson.ErrRPCOutOfRange {
+			return time.Time{}, nil // not that many confs yet
+		}
+		return time.Time{}, fmt.Errorf("getblockhash failed: %w", err)
+	}
+	gbv, err = btc.node.GetBlockVerbose(blockHash)
+	if err != nil {
+		return time.Time{}, fmt.Errorf("getblock failed: %w", err)
+	}
+	return time.Unix(gbv.Time, 0).UTC(), nil
```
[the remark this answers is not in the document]
chappjc: That had crossed my mind but I forgot to look into it.  As you mentioned in the other comment where ConfCoinTime is called, we should just use current time (if the coin does have swapConf).

server/swap/swap_test.go
chappjc: Nope.  They're trash too.

server/swap/swap.go
chappjc: Right, good catch and it's part of why it's better that market is back in charge of order status fully again.  I will certainly make a note about the caveat you identified and have a scan myself.

chappjc: Turned out that `(*Market).SwapDone` was calling `Remaining()`.  Fortunately since it's in Market, we don't have to infer based on type/force/fill because we can just to `Book.HaveOrder()`, so that was an easy fix.

Probably safest would be not to have and `order.Order` as an argument, and instead pass a whole bunch of separate data (ID, type, force, base, quote, user), but we'd need to hack together an Order anyway to use the coin unlocking and storage functions and I think that's even more error prone.  We're returning to the idea of a `FilledOrder` that adds the annoying Filled/Remaining stuff so that the order.Order type can be immutable..  Maybe in the future.  It's a big change to do right.

chappjc:
I reverted the commit "client/comms: expire response handlers on reconnect" because it was only appropriate for reconnects where the server had restarted, not for when a reconnect was just a connection issue.  This is because authenticated requests handlers (or their child goroutines) will respond via `AuthManager.Send`, which looks up the current comms.Link on which the response should be sent, which means client can still get responses to requests made on a replaced link.

So instead of expiring response handlers client-side, I just changed the request timeout used for the `signAndRequest` calls in `finalize{Init,Redeem}Action` to a timeout that is a fraction of broadcast timeout.  Related issue: https://github.com/decred/dcrdex/issues/863

buck54321 (approved):

Issue #552, opened 2020-07-17
duplicate fee payment notifications sent to server


chappjc:
```
2020-07-17 14:59:34.725 [TRC] CORE: processing tip change for dcr
2020-07-17 14:59:34.726 [DBG] CORE: Registration fee txn 0deadd383776cb445b8872a6b40f1a3c06788f7ea5b73da036485818d29a2f90:0 now has 1 confirmations.
2020-07-17 14:59:34.726 [INF] CORE: Notifying dex dex-test.ssgen.io:7232 of fee payment.
2020-07-17 14:59:34.726 [DBG] CORE: Registration fee txn 0deadd383776cb445b8872a6b40f1a3c06788f7ea5b73da036485818d29a2f90:0 now has 1 confirmations.
2020-07-17 14:59:34.726 [INF] CORE: Notifying dex dex-test.ssgen.io:7232 of fee payment.
2020-07-17 14:59:34.760 [ERR] CORE: notify: |ERROR| (feepayment) Fee payment error - Error encountered while paying fees to dex-test.ssgen.io:7232: notify fee result error: rpc error: 40: 'notifyfee' send for paid account
2020-07-17 14:59:34.770 [DBG] CORE: authenticated connection to dex-test.ssgen.io:7232, 0 active matches
2020-07-17 14:59:34.771 [INF] CORE: notify: |SUCCESS| (feepayment) Account registered - You may now trade at dex-test.ssgen.io:7232
```

chappjc:
Seems like two different things are triggering this when there shouldn't be:

```
2020-07-30 16:26:05.050 [TRC] CORE: processing tip change for dcr
2020-07-30 16:26:05.051 [DBG] CORE: Registration fee txn 11097c9734592799185baa075525abc2ddb8874d6cc387e23ac81c91889eb4dc:0 now has 1 confirmations.
2020-07-30 16:26:05.051 [INF] CORE: Notifying dex 127.0.0.1:7232 of fee payment.
2020-07-30 16:26:05.051 [DBG] CORE: Registration fee txn 11097c9734592799185baa075525abc2ddb8874d6cc387e23ac81c91889eb4dc:0 now has 1 confirmations.
2020-07-30 16:26:05.051 [INF] CORE: Notifying dex 127.0.0.1:7232 of fee payment.
[...]

PR #860, opened 2020-12-01
client/core: duplicate fee payment notifications sent to server

kevinstl:
closes https://github.com/decred/dcrdex/issues/552

Using an index number as the key in blockWaiters allows duplicate waiters to exist in blockWaiters. Duplicate waiters will cause duplicate notifications to be sent to the server.

If we use AssetID as the key to blockWaiters we will no longer send duplicate notifications as only one waiter will exist per registration.

I think this may cause issues if the user registers more than one dex, but I believe that was a problem prior to using AssetID as the key to blockWaiters.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -839,7 +839,7 @@ type Core struct {
 	wallets   map[uint32]*xcWallet
 
 	waiterMtx    sync.Mutex
-	blockWaiters map[uint64]*blockWaiter
+	blockWaiters map[uint32]*blockWaiter
 
 	tickSchedMtx sync.Mutex
 	tickSched    map[order.OrderID]*time.Timer
@@ -879,7 +879,7 @@ func New(cfg *Config) (*Core, error) {
 		net:           cfg.Net,
 		lockTimeTaker: dex.LockTimeTaker(cfg.Net),
 		lockTimeMaker: dex.LockTimeMaker(cfg.Net),
-		blockWaiters:  make(map[uint64]*blockWaiter),
+		blockWaiters:  make(map[uint32]*blockWaiter),
 		piSyncers:     make(map[order.OrderID]chan struct{}),
 		tickSched:     make(map[order.OrderID]*time.Timer),
 		// Allowing to change the constructor makes testing a lot easier.
@@ -2337,7 +2337,7 @@ var waiterID uint64
 func (c *Core) wait(assetID uint32, trigger func() (bool, error), action func(error)) {
 	c.waiterMtx.Lock()
 	defer c.waiterMtx.Unlock()
-	c.blockWaiters[atomic.AddUint64(&waiterID, 1)] = &blockWaiter{
+	c.blockWaiters[assetID] = &blockWaiter{
 		assetID: assetID,
 		trigger: trigger,
 		action:  action,
@@ -4283,7 +4283,7 @@ func handleRedemptionRoute(c *Core, dc *dexConnection, msg *msgjson.Message) err
 }
 
 // removeWaiter removes a blockWaiter from the map.
-func (c *Core) removeWaiter(id uint64) {
+func (c *Core) removeWaiter(id uint32) {
 	c.waiterMtx.Lock()
 	delete(c.blockWaiters, id)
 	c.waiterMtx.Unlock()
@@ -4303,7 +4303,7 @@ func (c *Core) tipChange(assetID uint32, nodeErr error) {
 		if waiter.assetID != assetID {
 			continue
 		}
-		go func(id uint64, waiter *blockWaiter) {
+		go func(id uint32, waiter *blockWaiter) {
 			ok, err := waiter.trigger()
 			if err != nil {
 				waiter.action(err)
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -3082,6 +3082,13 @@ func (c *Core) checkUnpaidFees(dcrWallet *xcWallet) {
 // called if the client was shutdown after a fee was paid, but before it had the
 // requisite confirmations for the 'notifyfee' message to be sent to the server.
 func (c *Core) reFee(dcrWallet *xcWallet, dc *dexConnection) {
+
+	// Return if the coin is already in blockWaiters.
+	regFeeAssetID, _ := dex.BipSymbolID(regFeeAssetSymbol)
+	if _, ok := c.blockWaiters[coinIDString(regFeeAssetID, dc.acct.feeCoin)]; ok {
+		return
+	}
+
 	// Get the database account info.
 	acctInfo, err := c.db.Account(dc.acct.host)
 	if err != nil {
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -839,7 +839,7 @@ type Core struct {
 	wallets   map[uint32]*xcWallet
 
 	waiterMtx    sync.Mutex
-	blockWaiters map[uint32]*blockWaiter
+	blockWaiters map[string]*blockWaiter
 
 	tickSchedMtx sync.Mutex
 	tickSched    map[order.OrderID]*time.Timer
@@ -879,7 +879,7 @@ func New(cfg *Config) (*Core, error) {
 		net:           cfg.Net,
 		lockTimeTaker: dex.LockTimeTaker(cfg.Net),
 		lockTimeMaker: dex.LockTimeMaker(cfg.Net),
-		blockWaiters:  make(map[uint32]*blockWaiter),
+		blockWaiters:  make(map[string]*blockWaiter),
 		piSyncers:     make(map[order.OrderID]chan struct{}),
 		tickSched:     make(map[order.OrderID]*time.Timer),
 		// Allowing to change the constructor makes testing a lot easier.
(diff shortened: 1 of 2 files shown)


JoeGruffins (commented):
I would rather the waiter not be added at all if we don't need it. I can reproduce with the following steps on master simnet:
1. Register an account and pay.
2. Lock and unlock the dcrwallet.
3. Move the chain forward.

```
2020-12-01 17:10:45.138 [DBG] CORE[dcr]: tip change: 647 (1db324499b4302311d36dc1083a6e2b82156bd27f7c91f831811875938152db2) => 648 (40a8da7a4ef5cad6b9c5c39c82bf8e42da74fcc61574d72b50a4eaf6c26a9700)
2020-12-01 17:10:45.138 [TRC] CORE: processing tip change for dcr
2020-12-01 17:10:45.139 [DBG] CORE: Registration fee txn 43e499cba6a7c974fa7fc25cbd4f9413ab13ffb84c97da417b2c3f7d1ddb4c43:0 now has 1 confirmations.
2020-12-01 17:10:45.139 [INF] CORE: Notifying dex 127.0.0.1:17273 of fee payment.
2020-12-01 17:10:45.139 [DBG] CORE: Registration fee txn 43e499cba6a7c974fa7fc25cbd4f9413ab13ffb84c97da417b2c3f7d1ddb4c43:0 now has 1 confirmations.
2020-12-01 17:10:45.139 [INF] CORE: Notifying dex 127.0.0.1:17273 of fee payment.
2020-12-01 17:10:45.139 [DBG] CORE: Registration fee txn 43e499cba6a7c974fa7fc25cbd4f9413ab13ffb84c97da417b2c3f7d1ddb4c43:0 now has 1 confirmations.
2020-12-01 17:10:45.139 [INF] CORE: Notifying dex 127.0.0.1:17273 of fee payment.
2020-12-01 17:10:45.139 [DBG] CORE: Registration fee txn 43e499cba6a7c974fa7fc25cbd4f9413ab13ffb84c97da417b2c3f7d1ddb4c43:0 now has 1 confirmations.
2020-12-01 17:10:45.139 [INF] CORE: Notifying dex 127.0.0.1:17273 of fee payment.
2020-12-01 17:10:45.139 [DBG] CORE: Registration fee txn 43e499cba6a7c974fa7fc25cbd4f9413ab13ffb84c97da417b2c3f7d1ddb4c43:0 now has 1 confirmations.
2020-12-01 17:10:45.139 [INF] CORE: Notifying dex 127.0.0.1:17273 of fee payment.
2020-12-01 17:10:45.149 [ERR] CORE: notify: |ERROR| (feepayment) Fee payment error - Error encountered while paying fees to 127.0.0.1:17273: notify fee result error: rpc error: error code 41: 'notifyfee' send for paid account
2020-12-01 17:10:45.150 [ERR] CORE: notify: |ERROR| (feepayment) Fee payment error - Error encountered while paying fees to 127.0.0.1:17273: notify fee result error: rpc error: error code 41: 'notifyfee' send for paid account
2020-12-01 17:10:45.151 [ERR] CORE: notify: |ERROR| (feepayment) Fee payment error - Error encountered while paying fees to 127.0.0.1:17273: notify fee result error: rpc error: error code 41: 'notifyfee' send for paid account
2020-12-01 17:10:45.153 [ERR] CORE: notify: |ERROR| (feepayment) Fee payment error - Error encountered while paying fees to 127.0.0.1:17273: notify fee result error: rpc error: error code 41: 'notifyfee' send for paid account
2020-12-01 17:10:45.155 [TRC] CORE: notify: |DATA| (balance) balance updated
2020-12-01 17:10:45.160 [DBG] CORE: Authenticated connection to 127.0.0.1:17273, 0 active orders, 0 active matches, score 0
2020-12-01 17:10:45.161 [INF] CORE: notify: |SUCCESS| (feepayment) Account registered - You may now trade at 127.0.0.1:17273
```
I believe this is because `checkUnpaidFees` is called over and over, which in turn calls `reFee`. `reFee` is also called by `initializeDEXConnections`. What if you made the CoinID the map key, then you check if that key exists first before `reFee` is called. You could also stop in `reFee`, and just don't add a waiter for the same coin.

Also, `waiterID` is no longer used.

JoeGruffins (approved):
Looks great to me!

client/core/core.go
```diff
@@ -3084,6 +3082,13 @@ func (c *Core) checkUnpaidFees(dcrWallet *xcWallet) {
 // called if the client was shutdown after a fee was paid, but before it had the
 // requisite confirmations for the 'notifyfee' message to be sent to the server.
 func (c *Core) reFee(dcrWallet *xcWallet, dc *dexConnection) {
+
+	// Return if the coin is already in blockWaiters.
+	regFeeAssetID, _ := dex.BipSymbolID(regFeeAssetSymbol)
+	if _, ok := c.blockWaiters[coinIDString(regFeeAssetID, dc.acct.feeCoin)]; ok {
```
chappjc, line 3088: Concurrent map access.  Perhaps an `isWaiting` method next to the `removeWaiter` method that wraps the access in `c.waiterMtx.RLock()`?

client/core/core.go
```diff
@@ -3084,6 +3082,13 @@ func (c *Core) checkUnpaidFees(dcrWallet *xcWallet) {
 // called if the client was shutdown after a fee was paid, but before it had the
```
chappjc, line 3082: In `checkUnpaidFees` above, the `feeLock` and it's atomic accesses are now obsolete.  I believe it was not serving its intended purpose because `checkUnpaidFees` was not the only path to `verifyRegistrationFee`.

client/core/core_test.go
```diff
@@ -785,7 +785,7 @@ func newTestRig() *testRig {
 			lockTimeTaker: dex.LockTimeTaker(dex.Testnet),
 			lockTimeMaker: dex.LockTimeMaker(dex.Testnet),
 			wallets:       make(map[uint32]*xcWallet),
-			blockWaiters:  make(map[uint64]*blockWaiter),
+			blockWaiters:  make(map[string]*blockWaiter),
```
chappjc, line 788: Since `blockWaiters` is only used for registration fee txns, I believe `account.AccountID` would be a more appropriate key for the map given you can't register twice with the same account ID, at present anyway.  Or even the DEX's pubkey, which would be a `[secp256k1.PubKeyBytesLenCompressed]byte`.
But it's OK to leave as is.

chappjc (changes requested):
The issue was indeed that there were multiple waiters created for the same coin.  The different paths:

1. `OpenWallet` > go `checkUnpaidFees` > `reFee` > `verifyRegistrationFee`
2. `Login` > `initializeDEXConnections` > `reFee` > `verifyRegistrationFee`
3. `Register` > `verifyRegistrationFee`

@JoeGruffins Described a repro that used Register (#3) followed by OpenWallet (#1)

The reason I often hit it is because my workflow is to do the init, setup, and register via dexcctl (#3), and *only then* opening the browser page to login (#2) *while the fee is still pending*, thus creating two waiters for the same coin.

I'm fine with the current solution that re-keys the `blockWaiters` map with a string for the coin being waited on, but there are a couple remaining cleanup items.

client/core/core.go
chappjc: Actually, maybe the `feeLock` is still required.  OK to leave it.

client/core/core_test.go
kevinstl: What would you think about using dex host as key?

chappjc: Maybe.  I kinda think that some broader refactoring is warranted for future work though.  `blockWaiters` makes it seem like a generic waiter, and it is, but really it's just used for watching fee txns for confirms.  So if this is made explicitly into a `regFeeWaiters` map then it would make more sense to key it on the DEX server's pubkey or host.

I think this work fixes issue https://github.com/decred/dcrdex/issues/552 so it can stop there for now.

chappjc: But the reason I suggested the pubkey is because in general (1) strings make less efficient map keys than say arrays or structs even though any comparable type can be used, and (2) the semantics of the string as a key are not obvious at a glance while when you see `map[account.AccountID]whatever` it's perfectly clear.

chappjc (approved):
I can no longer reproduce the issue as described in https://github.com/decred/dcrdex/pull/860#pullrequestreview-543213418

Future work may investigate changing the blockWaiters map, but I don't think it's a high priority.

client/core/core.go
```diff
@@ -3084,6 +3082,13 @@ func (c *Core) checkUnpaidFees(dcrWallet *xcWallet) {
...
 // requisite confirmations for the 'notifyfee' message to be sent to the server.
 func (c *Core) reFee(dcrWallet *xcWallet, dc *dexConnection) {
+
+	// Return if the coin is already in blockWaiters.
+	regFeeAssetID, _ := dex.BipSymbolID(regFeeAssetSymbol)
+	if c.existsWaiter(coinIDString(regFeeAssetID, dc.acct.feeCoin)) {
```
chappjc, line 3088: Looking back I realized this isn't perfect since different assets can have the same coinid but that's super unlikely to happen.  Also, it really doesn't need to be converted with `coinIDString` to have a map key as we could have just done hex.EncodeToString on the feeCoin.
Regardless, I'm flagging this PR as part of the [0.1.4 milestone](https://github.com/decred/dcrdex/milestone/10).

PR #861, opened 2020-12-01
dexcctl: Add rpcpass to simnet-setup script.

JoeGruffins:


client/cmd/dexcctl/simnet-setup.sh
```diff
@@ -2,18 +2,22 @@
...
+${AUTH} -p abc newwallet 42 ~/dextest/dcr/alpha/alpha.conf '{"account":"default"}'
 echo configuring Bitcoin wallet
-./dexcctl -p abc -p "" newwallet 0 ~/dextest/btc/alpha/alpha.conf '{"walletname":"gamma"}'
+${AUTH} -p "" newwallet 0 ~/dextest/btc/alpha/alpha.conf '{"walletname":"gamma"}'
 echo configuring Litecoin wallet
-./dexcctl -p abc -p "" newwallet 2 ~/dextest/ltc/alpha/alpha.conf '{"walletname":"gamma"}'
+${AUTH} -p "" newwallet 2 ~/dextest/ltc/alpha/alpha.conf '{"walletname":"gamma"}'
 echo registering with DEX
-./dexcctl -p abc register 127.0.0.1:17273 100000000 ~/dextest/dcrdex/rpc.cert
+${AUTH} register 127.0.0.1:17273 100000000 ~/dextest/dcrdex/rpc.cert
```
JoeGruffins, line 19: I had forgotten the brackets here, but it looks like my bash didn't need them...

client/cmd/dexcctl/simnet-setup.sh
```diff
@@ -2,18 +2,22 @@
 # Set up DCR and BTC wallets and register with the DEX.
 # dcrdex, dexc, and the wallet simnet harnesses should all be running before
 # calling this script.
+
+RPC_PASS="abc"
```
chappjc, line 6: Is the motivation for these changes so you can more easily edit the rpc pass?  If so, perhaps we should do `RPC_PASS="${1:-abc}"` or similar.

JoeGruffins: I don't think the rpcserver works anymore if you don't set a password. Does it? So the pass must be supplied for the commands. But your suggestion is better and I will do that.

chappjc: Oh, I bet you were affected by the move of dexcctl.conf from ~/.dexc to ~/.dexcctl recently.

JoeGruffins: ah, that's what changed? hmm. I guess this is not needed then, I should just make a conf file there.

JoeGruffins: I think this is still useful enough. And I already made the pr... Feel free to close.

chappjc: Yeah this is still nice.  Will test drive it first though

chappjc (approved):

PR #864, opened 2020-12-01
server/{admin,asset}: allow scaling fee rate

chappjc:
This adds a configurable per-asset scale factor that *server* asset backends will apply to their optimal fee rate estimates.

The admin server is updated with api/asset/{assetSymbol} and /api/asset/{assetSymbol}/setfeescale/{scale} routes.  I initially had the set done via POST on the same endpoint, but it's simpler this way even if not very RESTish.

The GET response for `/api/asset/dcr`:

```json
{
    "id": 42,
    "symbol": "dcr",
    "lotSize": 200000000,
    "rateStep": 5000,
    "maxFeeRate": 16,
    "swapSize": 251,
    "swapSizeBase": 85,
    "swapConf": 1,
    "currentFeeRate": 10,
    "scaledFeeRate": 12,
    "synced": true
}

```

<s>POST via `curl --cacert /home/user/.dcrdex/rpc.cert --basic -u u:pass --header "Content-Type: text/plain" --data-binary '{"feeRateScale":1.2}' https://127.0.0.1:6542/api/asset/btc`

Setting the fee rate scale is done via POST with a JSON payload instead of a specific URL pattern with GET because there may be several asset settings changeable in a single request at some point.</s>

Setting the fee rate scale is done with a GET to `/api/asset/{symbol}/setfeescale/{scale}`.

### DIFF
--- a/client/core/trade.go
+++ b/client/core/trade.go
@@ -503,6 +503,13 @@ func (t *trackedTrade) makeMetaMatch(msgMatch *msgjson.Match) *db.MetaMatch {
 		feeRateSwap = msgMatch.FeeRateBase
 	}
 
+	// Consider: bump fee rate here based on a user setting in dexConnection.
+	// feeRateSwap = feeRateSwap * 11 / 10
+	// maxFeeRate := t.dc.assets[swapAssetID].MaxFeeRate // swapAssetID according to t.Trade().Sell and t.Base()/Quote()
+	// if feeRateSwap > maxFeeRate {
+	// 	feeRateSwap = maxFeeRate
+	// }
+
 	var oid order.OrderID
 	copy(oid[:], msgMatch.OrderID)
 	var mid order.MatchID
--- a/server/admin/api.go
+++ b/server/admin/api.go
@@ -52,7 +52,7 @@ func writeJSONWithStatus(w http.ResponseWriter, thing interface{}, code int) {
 }
 
 // apiPing is the handler for the '/ping' API request.
-func (*Server) apiPing(w http.ResponseWriter, _ *http.Request) {
+func apiPing(w http.ResponseWriter, _ *http.Request) {
 	writeJSON(w, pongStr)
 }
 
@@ -75,11 +75,13 @@ func (s *Server) apiAsset(w http.ResponseWriter, r *http.Request) {
 	}
 
 	var errs []string
-	dexAsset := asset.Asset
 	backend := asset.Backend
+	var scaledFeeRate uint64
 	currentFeeRate, err := backend.FeeRate()
 	if err != nil {
 		errs = append(errs, fmt.Sprintf("unable to get current fee rate: %v", err))
+	} else {
+		scaledFeeRate = s.core.ScaleFeeRate(assetID, currentFeeRate)
 	}
 
 	synced, err := backend.Synced()
@@ -88,8 +90,9 @@ func (s *Server) apiAsset(w http.ResponseWriter, r *http.Request) {
 	}
 
 	res := &AssetInfo{
-		Asset:          dexAsset,
+		Asset:          asset.Asset,
 		CurrentFeeRate: currentFeeRate,
+		ScaledFeeRate:  scaledFeeRate,
 		Synced:         synced,
 		Errors:         errs,
 	}
@@ -103,7 +106,7 @@ func (s *Server) apiAssetPOST(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, fmt.Sprintf("unknown asset %q", assetSymbol), http.StatusBadRequest)
 		return
 	}
-	asset, err := s.core.Asset(assetID)
+	_, err := s.core.Asset(assetID) // asset return may be used if other asset settings are modified
 	if err != nil {
 		http.Error(w, fmt.Sprintf("unsupported asset %q / %d", assetSymbol, assetID), http.StatusBadRequest)
 		return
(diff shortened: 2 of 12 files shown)


chappjc:
Just added a second commit with a basic client side bump (basic because it could be wired with a user accessible option).

Converting to draft since this PR is for discussion.

client/core/trade.go
```diff
@@ -496,9 +496,17 @@ func (t *trackedTrade) makeMetaMatch(msgMatch *msgjson.Match) *db.MetaMatch {
...
+	base, quote := t.Base(), t.Quote()
+	t.dc.assetsMtx.RLock()
+	feeRateSwap, maxFeeRate := msgMatch.FeeRateQuote, t.dc.assets[quote].MaxFeeRate
 	if t.Trade().Sell {
-		feeRateSwap = msgMatch.FeeRateBase
+		feeRateSwap, maxFeeRate = msgMatch.FeeRateBase, t.dc.assets[base].MaxFeeRate
+	}
+	t.dc.assetsMtx.RUnlock()
+
+	feeRateSwap = feeRateSwap * 11 / 10 // This could be based on a normal/fast user setting somewhere in dexConnection
```
chappjc, line 507: maybe `feeRateSwap += feeRateSwap / 10` instead

chappjc:
Still in draft as I want to add an admin command to adjust this at runtime.  For example,

`POST /api/asset/btc` with data `{"swapFeeAdj": 110}` or a `GET /api/asset/btc/swapfeepct/110` to make it simpler for curling.

server/admin/api.go
```diff
@@ -47,7 +47,7 @@ func writeJSONWithStatus(w http.ResponseWriter, thing interface{}, code int) {
 }
 
 // apiPing is the handler for the '/ping' API request.
-func (_ *Server) apiPing(w http.ResponseWriter, _ *http.Request) {
+func (*Server) apiPing(w http.ResponseWriter, _ *http.Request) {
```
buck54321, line 50: I'm wondering why this is a `Server` method to begin with.

server/admin/api.go
```diff
@@ -56,6 +56,86 @@ func (s *Server) apiConfig(w http.ResponseWriter, _ *http.Request) {
...
+		http.Error(w, fmt.Sprintf("invalid POST data: %v", err), http.StatusBadRequest)
+		return
+	}
+
+	var settingsFound bool
+
+	if data.FeeRateScale != nil {
+		settingsFound = true
+		log.Infof("Setting %s (%d) fee rate scale factor to %f", strings.ToUpper(assetSymbol), assetID, *data.FeeRateScale)
+		asset.Backend.SetFeeRateScale(*data.FeeRateScale)
```
buck54321, line 129: Seems like your just using the backend as a `float64` field here. The `Backend` doesn't really need to know anything about this scale factor, does it?

buck54321, line 129: It should be noted that allowing the operator to apply these scale factors means that the operator now shares the blame for stuck swaps with the asset developer. With great power comes great responsibility and great potential for chaos.

buck54321 (changes requested):

server/admin/api.go
chappjc: I don't follow.  The backend applies the scale to the result of estimatesmartfee internally:  https://github.com/decred/dcrdex/pull/864/files/479ef31c6924215cb9fa09134553f4f878bfb6b9#diff-51dc1aa092977ea75e0a2e2e2bc5263254edd21ca3f653b0a27035c5c667c829R378-R386

chappjc: We could make 1.0 the min

buck54321: But every `Backend` is expected to apply the factor in exactly the same way, which to me indicates that it's not really a `Backend` role. The fee adjustment is just not the concern of the `Backend`. This factor can just as easily be applied by the caller without saddling asset developers with boilerplate.

buck54321: Aren't there times when the fees are outrageously high too?

chappjc: For sure.  Just saying it's something we could do out of caution because of the stuck swap risk.  I anticipated setting <1 for cooling down periods like what we've seen recently, but there's certainly a risk if is left <1 without close monitoring.  I think setting <1 could just be met with a warning to the operator.

chappjc: Right, gotcha.  I did it this way as a lazy way of making it asset-specific (and perhaps applicable to other rate uses even though there are none), but that is indeed achievable with a scale map in Swapper where it gets applied in `(*Swapper).Negotiate`

chappjc:
Adding this to the [0.1.4 milestone](https://github.com/decred/dcrdex/pull/860/files#r539434267) although it's not really patch material since we don't want to wait for 0.2 to have this capability.

JoeGruffins (approved):

server/admin/server.go
```diff
@@ -150,6 +154,10 @@ func NewServer(cfg *SrvConfig) (*Server, error) {
 			rm.Get("/forgive_match/{"+matchIDKey+"}", s.apiForgiveMatchFail)
 			rm.Post("/notify", s.apiNotify)
 		})
+		r.Route("/asset/{"+assetSymKey+"}", func(rm chi.Router) {
+			rm.Get("/", s.apiAsset)
+			rm.Post("/", s.apiAssetPOST)
```
buck54321, line 159: How about `/api/setfeescale`? We're only setting one field of the struct returned in the GET response.

buck54321 (approved):

server/admin/server.go
chappjc: Might as well.  I was anticipating a need to change other asset parameters such as maxFeeRate, lotsize, ratestep, etc. but on reflection, I believe those hot changes would be done via editing markets.json and signalling to reload/parse it.

PR #865, opened 2020-12-02
client/asset/btc: favor confirmed utxos in FundOrder

chappjc:
The changes the client's BTC backend so that the FundOrder implementation will favor confirmed utxos.

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -922,11 +922,9 @@ func (btc *ExchangeWallet) split(value uint64, lots uint64, outputs []*output, i
 	// from server, and have server check fee rate on unconf funding coins.
 	feeRate, err := btc.feeRate(1)
 	if err != nil {
-		// fallback fee rate is NO good here.
-		// feeRate = nfo.MaxFeeRate // maybe
-		return nil, false, fmt.Errorf("Unable to get optimal fee rate for pre-split transaction: %v", err)
+		// Fallback fee rate is NO GOOD here.
+		return nil, false, fmt.Errorf("unable to get optimal fee rate for pre-split transaction: %w", err)
 	}
-	feeRate += feeRate / 10 // cost of chaining unconf txn, don't gamble.
 	if feeRate > nfo.MaxFeeRate {
 		feeRate = nfo.MaxFeeRate
 	}
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -175,7 +175,7 @@ func newOutPoint(txHash *chainhash.Hash, vout uint32) outPoint {
 }
 
 // String is a string representation of the outPoint.
-func (pt *outPoint) String() string {
+func (pt outPoint) String() string {
 	return pt.txHash.String() + ":" + strconv.Itoa(int(pt.vout))
 }
 
@@ -628,6 +628,12 @@ func (btc *ExchangeWallet) feeRate(confTarget uint64) (uint64, error) {
 	return 1 + uint64(satPerKB)/1000, nil
 }
 
+type sat uint64
+
+func (s sat) String() string {
+	return strconv.FormatFloat(btcutil.Amount(s).ToBTC(), 'f', -1, 64) // dec, but no trailing zeros
+}
+
 // feeRateWithFallback attempts to get the optimal fee rate in sat / byte via
 // FeeRate. If that fails, it will return the configured fallback fee rate.
 func (btc *ExchangeWallet) feeRateWithFallback(confTarget uint64) uint64 {
@@ -721,8 +727,8 @@ func (btc *ExchangeWallet) RedemptionFees() (uint64, error) {
 // dex.Bytes should be appended to the redeem scripts collection for coins with
 // no redeem script.
 func (btc *ExchangeWallet) FundOrder(ord *asset.Order) (asset.Coins, []dex.Bytes, error) {
-	btc.log.Debugf("Attempting to fund order for %.8f %s, maxFeeRate = %d, max swaps = %d",
-		toBTC(ord.Value), btc.walletInfo.Units, ord.DEXConfig.MaxFeeRate, ord.MaxSwapCount)
+	btc.log.Debugf("Attempting to fund order for %s %s, maxFeeRate = %d, max swaps = %d",
+		sat(ord.Value), btc.symbol, ord.DEXConfig.MaxFeeRate, ord.MaxSwapCount)
 
 	if ord.Value == 0 {
 		return nil, nil, fmt.Errorf("cannot fund value = 0")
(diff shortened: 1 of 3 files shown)


client/asset/btc/btc.go
```diff
@@ -846,7 +865,15 @@ func (btc *ExchangeWallet) split(value uint64, lots uint64, outputs []*output, i
...
-	feeRate := btc.feeRateWithFallback(1) // these must fund swaps, so don't under-pay (is this an issue with no fundconf requirement?)
+	// This must fund swaps, so don't under-pay. TODO: get and use a fee rate
+	// from server, and have server check fee rate on unconf funding coins.
+	feeRate, err := btc.feeRate(1)
+	if err != nil {
+		// fallback fee rate is NO good here.
+		// feeRate = nfo.MaxFeeRate // maybe
+		return nil, false, fmt.Errorf("Unable to get optimal fee rate for pre-split transaction: %v", err)
+	}
+	feeRate += feeRate / 10 // cost of chaining unconf txn, don't gamble.
```
buck54321, line 876: If I'm not mistaken, we've been running with zero buffer and have yet to hit an issue. If a buffer gives us some comfort, so be it, but a 10% bump is a lot more than the couple of sats we should need.

client/asset/btc/btc.go
```diff
@@ -820,9 +839,9 @@ func (btc *ExchangeWallet) split(value uint64, lots uint64, outputs []*output, i
 	if baggage > excess {
 		btc.log.Debugf("Skipping split transaction because cost is greater than potential over-lock. "+
 			"%d > %d", baggage, excess)
-		btc.log.Infof("Funding %d %s order with coins %v worth %d",
-			value, btc.walletInfo.Units, coins, coinSum)
-		return coins, false, nil
+		btc.log.Infof("Funding %s %s order with coins %v worth %s",
+			sat(value), btc.symbol, coins, sat(coinSum))
+		return coins, false, nil // why fundingCoins is an input
```
buck54321, line 844: > // why fundingCoins is an input

Nuke it

buck54321 (changes requested):

client/asset/btc/btc.go
chappjc: This comment was supposed to be declaratory actually.  On return with `err==nil`, the deferred function processes the `fundingCoins`.  The `fundingCoins` arg is super confusing though in this regard.  It was removable in dcr, but not btc because of this, and I really didn't want to refactor too much stuff.  I'll clarify this comment.

client/asset/btc/btc.go
chappjc: Well, I'm not so sure actually we haven't been bit, primarily because the estimate is the client's rather than server's.  This is a bad band-aid though.  I'll remove this little bump and we can focus on solving this properly.

client/asset/btc/btc.go
```diff
@@ -660,6 +660,12 @@ func (btc *ExchangeWallet) feeRateWithFallback(confTarget uint64) uint64 {
 	return feeRate
 }
 
+type sat uint64
+
+func (s sat) String() string {
+	return strconv.FormatFloat(btcutil.Amount(s).ToBTC(), 'f', -1, 64) // dec, but no trailing zeros
+}
+
```
chappjc, line 668: @buck54321 BTW, I noticed in one of your PRs you made a bunch of changes to the logging in the same vein, although my approach with FormatFloat was so that the value is only displayed with as many digits as required.  That is 1.21 DCR would display as `1.21000000` with `%.8f`, but `1.21` with `FormatFloat` with the `-1` precision arg.  How do you feel about the display and removing the trailing zeros?  I feel like most of the time it hurts readability to keep the zeros.

client/asset/btc/btc.go
```diff
@@ -2191,7 +2219,7 @@ func (btc *ExchangeWallet) spendableUTXOs(confs uint32) ([]*compositeUTXO, map[o
 			// unlocked something or even restarted the wallet software.
 			pt := newOutPoint(txHash, txout.Vout)
 			if btc.fundingCoins[pt] != nil {
-				btc.log.Warnf("Known order-funding coin %v returned by listunspent!", pt)
+				btc.log.Warnf("Known order-funding coin %s returned by listunspent!", pt)
```
chappjc, line 2222: This line is why I changed the `outPoint`'s Stringer to value receiver. https://github.com/decred/dcrdex/pull/865/files/812da1ebbf8eb03471bb3d80328dd177f8eac16c#diff-1b07e494f4bdd99b2a617ac110fbadc081e5b6ae1abf23f64032bbee5ebebef1R177
It was just showing hex or something definitely not decoded.

client/asset/btc/btc.go
buck54321: That's good. I'll adapt mine.

chappjc:
That was a hairy rebase with https://github.com/decred/dcrdex/pull/842 now in master, so I'll want to retest manually again.

JoeGruffins (approved):

client/asset/btc/btc.go
```diff
@@ -648,6 +648,12 @@ func (btc *ExchangeWallet) feeRate(confTarget uint64) (uint64, error) {
 	return 1 + uint64(satPerKB)/1000, nil
 }
 
+type sat uint64
+
+func (s sat) String() string {
+	return strconv.FormatFloat(btcutil.Amount(s).ToBTC(), 'f', -1, 64) // dec, but no trailing zeros
+}
```
buck54321, line 655: I'm all for a custom type for this purpose, but I was a little confused since the name is `sat`, but it's primary purpose is for a string in conventional units. Confusion isn't helped by the way it is used, `sat(v)`, which could appear to be a function call to convert a conventional unit value to Satoshis. 

Since it's handling values in two representations, how about the more ambiguous `type amount uint64`?

client/asset/btc/btc.go
```diff
@@ -920,7 +938,13 @@ func (btc *ExchangeWallet) split(value uint64, lots uint64, outputs []*output, i
...
 	}
 
-	feeRate := btc.feeRateWithFallback(1) // these must fund swaps, so don't under-pay (is this an issue with no fundconf requirement?)
+	// This must fund swaps, so don't under-pay. TODO: get and use a fee rate
+	// from server, and have server check fee rate on unconf funding coins.
+	feeRate, err := btc.feeRate(1)
+	if err != nil {
+		// Fallback fee rate is NO GOOD here.
+		return nil, false, fmt.Errorf("unable to get optimal fee rate for pre-split transaction: %w", err)
+	}
```
buck54321, line 947: People are probably going to hit this, but we need this and a full solution is outside of the scope of this PR. I would just suggest that we get a more robust solution in before 0.2.

buck54321 (changes requested):

client/asset/btc/btc.go
chappjc: I think it just needs to be viewed as a compatible type conversion.  That is, you assert that the value you have is Satoshis via `sat(v)`.  This is similar to doing `dcrutil.Amount(someInt64)`, which gives you an `Amount` with the same underlying type, just new methods.  Admittedly it just has one method, so the value of Stringerization vs a function call is limited.  I will go back and toy with a function, but I think making it more ambiguous with `amount` vs. `sat` is actually more confusing.

Part of the consideration for this approach (stringer vs function) is that in a log call, it will only call the stringer when the debug level requires evaluating the stringer, but that's admittedly not a strong reason.

client/asset/btc/btc.go
chappjc: Yeah, we do need a proper solution before next release.  I'm pretty sure people do hit this case now (and use the too-low fallback), but it's not clear how often it is BTC and for taker orders.

chappjc: Oh, but I think my reasoning for letting this ride is that they're aren't entirely blocked from placing orders, just placing orders with a pre-split txn. :)

PR #870, opened 2020-12-04
client/{asset,core}: sendtoaddress methods

chappjc:


client/core/core.go
```diff
@@ -2435,6 +2438,19 @@ func (c *Core) Withdraw(pw []byte, assetID uint32, value uint64, address string)
 	return coin, nil
 }
 
+// SendToAddress sends the an amount to an address. Unlike Withdraw, transaction
+// fees are in addition to the registration fee, and the fee rate is taken from
+// the DEX configuration.
```
chappjc, line 2443: <s>registration fee</s> sent amount

chappjc:
I will likely close this since reportedly Withdraw is adequate for the application this was indended for.  However, this PR may be useful if the browser's Withdraw form could be given a check box to toggle subtraction of the txn fee from output.

e.g.

image

PR #871, opened 2020-12-05
client: Network specific loopback IPs

### LINKED ISSUE #557: dexc: rpc and web ports should change by network
dexc binds the rpc and web servers on 5757 and 5758 by default.  It would be nice if the defaults changed depending on main/test/simnet.

Note that this would require dexcctl to have `--testnet` and `--simnet` switches too, which it presently does not have.

### LINKED ISSUE #845: client: unintentional dexauth cookie invalidation
Because cookies are not port specific, if you have two instances of dexc running (e.g. mainnet and simnet) and open browser and login to both, one of those will lose the valid cookie because the dex auth cookie applies to any page at the given domain **regardless of port**.  To prevent this, we might consider putting another identifier in the cookie name, such as "dexauth_{portnumber}" or "dexauth_simnet" but that complicates the auth middleware.  Perhaps a better option is to bind to different loopback Ip addresses as suggested in https://github.com/decred/dcrdex/issues/845#issuecomment-730571250.

A related issue seems to be that calling (*Core).Login twice as would be done if having to login again can break the internal wallet lock state.

amassarwi:
The diff resolves a cookies conflict users experienced when running two different networks together, as both networks had
the same loopback IP (127.0.0.1), this commit adds two different loopback IPs for `testent` & `simnet` as described in https://github.com/decred/dcrdex/issues/845#issuecomment-730571250:

-    mainnet: 127.0.0.1
-   testnet: 127.0.0.2
-    simnet: 127.0.0.3

Users would still be able to specify custom addresses be providing `WebAddr`/`RPCAddr` configs.

In addition this diff adds `--testnet` & `--simnet` flags to `dexcctl` side as well, in order to hit the correct network specific address. 

Closes #845 
Closes #557


buck54321 (commented):
This could resolve the cookie issue, but this doesn't actually resolve #557, because two dexc instances for different networks can still not run simultaneously (without custom config). We need network-custom ports too. 

As a solution to the cookie issue, this works, but once we have custom ports, this can become a bit cumbersome. Now I'll have to remember both the custom port and the custom IP for each network. Have you considered just having network-specific cookies instead?

buck54321 (changes requested):
Ignore my previous comment. This works well for its intended purpose, but this does break the `dexcctl` script, which will now need `--testnet` and `--simnet` flags.

amassarwi:
I see, will take care of `dcrctl` as well.

chappjc:
I initially considered the network specific cookie names in https://github.com/decred/dcrdex/issues/845, but it seemed more complex of a solution than this.  In any case, I take this different loopback approach manually already, so I think it's a good solution to the cookies issue.

client/cmd/dexc/config.go
```diff
@@ -54,18 +57,37 @@ func setNet(applicationDirectory, net string) string {
 	return filepath.Join(netDirectory, "dexc.db")
 }
 
+// defaultHostByNetwork accepts configured host with configured network
+// flags and returns the network specific default host if the provided host is
+// empty
+func defaultHostByNetwork(host string, testnet, simnet bool) string {
```
chappjc, line 63: How about `net dex.Network` instead of `testnet, simnet bool`?

client/cmd/dexc/config.go
```diff
@@ -149,6 +171,28 @@ func configure() (*Config, error) {
 	if cfg.Simnet && cfg.Testnet {
 		return nil, fmt.Errorf("simnet and testnet cannot both be specified")
 	}
+
+	// Calc RPC & Web addresses, if any of the hosts isn't available fallback
+	// to default network specific host
+	RPCHost := defaultHostByNetwork(cfg.RPCHost, cfg.Testnet, cfg.Simnet)
+	WebHost := defaultHostByNetwork(cfg.WebHost, cfg.Testnet, cfg.Simnet)
```
chappjc, line 178: We'll have cfg.Net set in the next switch below, so we can use that as an argument to `defaultHostByNetwork` if this is moved below that switch.

chappjc (changes requested):

client/cmd/dexc/config.go
amassarwi: True, nice observation

client/cmd/dexc/config.go
amassarwi: Yep

amassarwi:
@chappjc updated

buck54321 (changes requested):
I hit this 

`unable to send request: Post "https://127.0.0.3:5757": x509: certificate is valid for 127.0.0.1, ::1, 192.168.0.2,`

So it looks like we'll have to add the new hosts to the certificate and everyone with an existing certificate will have to figure out that they need to delete it. 

To add new hosts, you'll need to add the IP addresses as strings in the last argument to `certgen.NewTLSCertPair`

https://github.com/decred/dcrdex/blob/08afde3f7bb5d9df5579621ed0e7ae9850424744/client/rpcserver/rpcserver.go#L93-L94

Also, please add the `--simnet` flag to *dexcctl/simnet-setup.sh*.

amassarwi:
@buck54321 done

chappjc:
I brought it up in matrix, but do we need to go with separate host/port arguments?  I hate breaking things for users.  If we do have to do it separately like this, we'll at least need some kind of backward compatibility and a deprecation warning for `webaddr`

amassarwi:
@chappjc Good point, seems like we still could achieve the same without splinting the address argument,
will push an update in a bit

amassarwi:
@chappjc updated

JoeGruffins (commented):
I'm able to talk to dexc with the rpcserver after deleting my rpc.cert/key, ~but for some reason I don't see anything when visiting `http://127.0.0.3:5758` with my browser on simnet. Is there another step needed?~ edit: It's working.

JoeGruffins (approved):
Also tried it out on mainnet and looks fine. Didn't need to do anything to existing files.

Saw one wierd message in the server, probably not related to these changes. `[TRC] COMM[WS]: Sent 29 and dropped 0 messages to [127 0 0 1 0 0 0 0] before shutdown.`

buck54321 (approved):

chappjc (approved):

chappjc:
@JoeGruffins There were recent changes to wslink's IP and address data, and there were a few missed logs it seems.  I'll tweak that in a follow up PR.

PR #874, opened 2020-12-07
client: Fee rate limit setting

### LINKED ISSUE #858: client: fee rate limit setting
The maxFeeRate config item from the server describes the maximum possible fee rate that could be required at match time for swap txns, however the client software should have a user-configurable limit on what max fee rates are acceptable here.

amassarwi:
This diff adds new fee rate threshold wallet setting which will be checked against server `maxFeeRate` when placing and order.
if `maxFeeRate` is greater than the threshold, client throws an error.

Part of #858

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -127,9 +127,9 @@ var (
 			DisplayName: "Pre-size funding inputs",
 			Description: "When placing an order, create a \"split\" transaction to " +
 				"fund the order without locking more of the wallet balance than " +
-				"necessary. Otherwise, excess funds may be reserved to fund the order" +
+				"necessary. Otherwise, excess funds may be reserved to fund the order " +
 				"until the first swap contract is broadcast during match settlement, " +
-				"or the order is canceled. This an extra transaction for which network" +
+				"or the order is canceled. This an extra transaction for which network " +
 				"mining fees are paid. Used only for standing-type orders, e.g. limit " +
 				"orders without immediate time-in-force.",
 			IsBoolean: true,
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -53,6 +53,7 @@ const (
 	// target in blocks used by estimatesmartfee to get the optimal fee for a
 	// redeem transaction.
 	defaultRedeemConfTarget = 2
+	smallestUnit            = 1e8
 
 	minNetworkVersion  = 190000
 	minProtocolVersion = 70015
@@ -103,7 +104,7 @@ var (
 			DisplayName: "Fallback fee rate",
 			Description: "The fee rate to use for fee payment and withdrawals when" +
 				" estimatesmartfee is not available. Units: BTC/kB",
-			DefaultValue: defaultFee * 1000 / 1e8,
+			DefaultValue: defaultFee * 1000 / smallestUnit,
 		},
 		{
 			Key:         "feeratelimit",
@@ -112,7 +113,7 @@ var (
 				"pay on swap transactions. If feeratelimit is lower than a market's " +
 				"maxfeerate, you will not be able to trade on that market with this " +
 				"wallet.  Units: BTC/kB",
-			DefaultValue: defaultFeeRateLimit * 1000 / 1e8,
+			DefaultValue: defaultFeeRateLimit * 1000 / smallestUnit,
 		},
 		{
 			Key:         "redeemconftarget",
@@ -435,7 +436,8 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network) 
 // with minimal coding.
 func BTCCloneWallet(cfg *BTCCloneCFG) (*ExchangeWallet, error) {
 	// Read the configuration parameters
-	btcCfg, err := dexbtc.LoadConfigFromSettings(cfg.WalletCFG.Settings, cfg.Symbol, cfg.Network, cfg.Ports)
+	btcCfg, err := dexbtc.LoadConfigFromSettings(cfg.WalletCFG.Settings,
+		cfg.Symbol, cfg.Network, cfg.Ports)
 	if err != nil {
 		return nil, err
 	}
@@ -451,14 +453,20 @@ func BTCCloneWallet(cfg *BTCCloneCFG) (*ExchangeWallet, error) {
 		Pass:         btcCfg.RPCPass,
 	}, nil)
 	if err != nil {
-		return nil, fmt.Errorf("error creating BTC RPC client: %w", err)
+		return nil, fmt.Errorf("error creating BTC RPC client: %v", err)
 	}
 
-	return newWallet(cfg, btcCfg, client), nil
+	btc, err := newWallet(cfg, btcCfg, client)
+	if err != nil {
+		return nil, fmt.Errorf("error creating BTC ExchangeWallet: %v", err)
+	}
+	btc.client = client
+
+	return btc, nil
 }
 
 // newWallet creates the ExchangeWallet and starts the block monitor.
-func newWallet(cfg *BTCCloneCFG, btcCfg *dexbtc.Config, node rpcClient) *ExchangeWallet {
+func newWallet(cfg *BTCCloneCFG, btcCfg *dexbtc.Config, node rpcClient) (*ExchangeWallet, error) {
 	// If set in the user config, the fallback fee will be in conventional units
 	// per kB, e.g. BTC/kB. Translate that to sats/byte.
 	fallbackFeesPerByte := toSatoshi(btcCfg.FallbackFeeRate / 1000)
@@ -469,7 +477,13 @@ func newWallet(cfg *BTCCloneCFG, btcCfg *dexbtc.Config, node rpcClient) *Exchang
 		fallbackFeesPerByte, cfg.WalletInfo.Units)
 
 	// If set in the user config, the fee rate limit will be in units of BTC/KB.
-	// Convert to sats/byte.
+	// Ensure value isn't smaller than smallest unit & convert to sats/byte.
+	if btcCfg.FeeRateLimit != 0 &&
+		// Reject rate < 1e-5
+		btcCfg.FeeRateLimit < 1/(smallestUnit/1000) {
+		return nil, fmt.Errorf("Fee rate limit is smaller than smallest unit: %v",
+			btcCfg.FeeRateLimit)
+	}
 	feesLimitPerByte := toSatoshi(btcCfg.FeeRateLimit / 1000)
 	if feesLimitPerByte == 0 {
 		feesLimitPerByte = cfg.DefaultFeeRateLimit
@@ -499,7 +513,7 @@ func newWallet(cfg *BTCCloneCFG, btcCfg *dexbtc.Config, node rpcClient) *Exchang
 		useLegacyBalance:    cfg.LegacyBalance,
 		segwit:              cfg.Segwit,
 		walletInfo:          cfg.WalletInfo,
-	}
+	}, nil
 }
 
 var _ asset.Wallet = (*ExchangeWallet)(nil)
@@ -2421,7 +2435,7 @@ func (btc *ExchangeWallet) wireBytes(tx *wire.MsgTx) []byte {
 
 // Convert the BTC value to satoshi.
 func toSatoshi(v float64) uint64 {
-	return uint64(math.Round(v * 1e8))
+	return uint64(math.Round(v * smallestUnit))
 }
 
 // blockHeader is a partial btcjson.GetBlockHeaderVerboseResult with mediantime
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -53,7 +53,8 @@ const (
 	// target in blocks used by estimatesmartfee to get the optimal fee for a
 	// redeem transaction.
 	defaultRedeemConfTarget = 2
-	smallestUnit            = 1e8
+	// Smallest unit is 1 satoshi = 1e-8 BTC
+	smallestUnit = 1e8
 
 	minNetworkVersion  = 190000
 	minProtocolVersion = 70015
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -47,8 +47,6 @@ const (
 	// The default fee is passed to the user as part of the asset.WalletInfo
 	// structure.
 	defaultFee = 100
-	// defaultFeeLimit is the default value for feelimit.
-	//defaultFeeLimit = 600
 	// defaultRedeemConfTarget is the default redeem transaction confirmation
 	// target in blocks used by estimatesmartfee to get the optimal fee for a
 	// redeem transaction.
@@ -107,7 +105,7 @@ var (
 		{
 			Key:         "feelimit",
 			DisplayName: "Highest acceptable fee rate",
-			Description: "This is the highest network fee rate you are willing to pay on swap transactions. If you set this too low, you may not be able to place orders with servers that allow for higher swap rates as network conditions demand. Units: BTC/kB",
+			Description: "This is the highest network fee rate you are willing to pay on swap transactions. If you set this too low, you may not be able to place orders with servers that allow for higher swap rates as network conditions demand. Units: sats/kB",
 		},
 		{
 			Key:          "redeemconftarget",
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -111,14 +111,14 @@ var (
 			Description: "This is the highest network fee rate you are willing to " +
 				"pay on swap transactions. If feeratelimit is lower than a market's " +
 				"maxfeerate, you will not be able to trade on that market with this " +
-				"wallet.  Units: sats/byte",
+				"wallet.  Units: BTC/kB",
 			DefaultValue: defaultFeeRateLimit * 1000 / 1e8,
 		},
 		{
 			Key:         "redeemconftarget",
 			DisplayName: "Redeem confirmation target",
 			Description: "The target number of blocks for the redeem transaction " +
-				"to get a confirmation. Used to set the transaction's fee rate. " +
+				"to be mined. Used to set the transaction's fee rate. " +
 				"(default: 2 blocks)",
 			DefaultValue: defaultRedeemConfTarget,
 		},
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -53,8 +53,6 @@ const (
 	// target in blocks used by estimatesmartfee to get the optimal fee for a
 	// redeem transaction.
 	defaultRedeemConfTarget = 2
-	// Smallest unit is 1 satoshi = 1e-8 BTC
-	smallestUnit = 1e8
 
 	minNetworkVersion  = 190000
 	minProtocolVersion = 70015
@@ -105,7 +103,7 @@ var (
 			DisplayName: "Fallback fee rate",
 			Description: "The fee rate to use for fee payment and withdrawals when" +
 				" estimatesmartfee is not available. Units: BTC/kB",
-			DefaultValue: defaultFee * 1000 / smallestUnit,
+			DefaultValue: defaultFee * 1000 / 1e8,
 		},
 		{
 			Key:         "feeratelimit",
@@ -114,7 +112,7 @@ var (
 				"pay on swap transactions. If feeratelimit is lower than a market's " +
 				"maxfeerate, you will not be able to trade on that market with this " +
 				"wallet.  Units: BTC/kB",
-			DefaultValue: defaultFeeRateLimit * 1000 / smallestUnit,
+			DefaultValue: defaultFeeRateLimit * 1000 / 1e8,
 		},
 		{
 			Key:         "redeemconftarget",
@@ -477,16 +475,14 @@ func newWallet(cfg *BTCCloneCFG, btcCfg *dexbtc.Config, node rpcClient) (*Exchan
 		fallbackFeesPerByte, cfg.WalletInfo.Units)
 
 	// If set in the user config, the fee rate limit will be in units of BTC/KB.
-	// Ensure value isn't smaller than smallest unit & convert to sats/byte.
-	if btcCfg.FeeRateLimit != 0 &&
-		// Reject rate < 1e-5
-		btcCfg.FeeRateLimit < 1/(smallestUnit/1000) {
-		return nil, fmt.Errorf("Fee rate limit is smaller than smallest unit: %v",
-			btcCfg.FeeRateLimit)
-	}
-	feesLimitPerByte := toSatoshi(btcCfg.FeeRateLimit / 1000)
-	if feesLimitPerByte == 0 {
-		feesLimitPerByte = cfg.DefaultFeeRateLimit
+	// Convert to sats/byte & error if value is smaller than smallest unit.
+	feesLimitPerByte := uint64(defaultFeeRateLimit)
+	if btcCfg.FeeRateLimit > 0 {
+		feesLimitPerByte = toSatoshi(btcCfg.FeeRateLimit / 1000)
+		if feesLimitPerByte == 0 {
+			return nil, fmt.Errorf("Fee rate limit is smaller than smallest unit: %v",
+				btcCfg.FeeRateLimit)
+		}
 	}
 	cfg.Logger.Tracef("Fees rate limit set at %d sats/byte", feesLimitPerByte)
 
@@ -2435,7 +2431,7 @@ func (btc *ExchangeWallet) wireBytes(tx *wire.MsgTx) []byte {
 
 // Convert the BTC value to satoshi.
 func toSatoshi(v float64) uint64 {
-	return uint64(math.Round(v * smallestUnit))
+	return uint64(math.Round(v * 1e8))
 }
 
 // blockHeader is a partial btcjson.GetBlockHeaderVerboseResult with mediantime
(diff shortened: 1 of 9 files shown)


client/core/core.go
```diff
@@ -2565,12 +2586,34 @@ func (c *Core) Trade(pw []byte, form *TradeForm) (*Order, error) {
 	return corder, nil
 }
 
+// validateFeeRateLimit checks if user provided fee rate limit value, if yes
+// it ensures that the server'r MaxFeeRate is smaller than the wallet's
+// configured feeRateLimit.
+func validateFeeRateLimit(w *xcWallet, a dex.Asset) error {
```
chappjc, line 2592: For clarity in the calling code, it's best to pass what's needed and not more.  That is, 
`func validateFeeRateLimit(feeRateLimit, maxFeeRate uint64) error {` But the function is so trivial I'd be inclined to just do the logic in the caller.  The only thing that warrants the function is the `feeRateLimit != 0` check and it seems that's not possible since the default is non-zero, right?

client/core/wallet.go
```diff
@@ -27,13 +27,14 @@ type xcWallet struct {
 	dbID         []byte
 	synced       bool
 	syncProgress float32
+	feeRateLimit uint32
```
chappjc, line 30: All other fees and fee rates are uint64, so let's be consistent.

client/asset/dcr/dcr.go
```diff
@@ -104,6 +104,11 @@ var (
 			Description:  "The fee rate to use for fee payment and withdrawals when estimatesmartfee is not available. Units: DCR/kB",
 			DefaultValue: defaultFee * 1000 / 1e8,
 		},
+		{
+			Key:         "feelimit",
```
chappjc, line 108: `"feelimit"` => `"feeratelimit"` for clarity

client/core/core.go
```diff
@@ -1385,6 +1388,24 @@ func (c *Core) loadWallet(dbWallet *db.Wallet) (*xcWallet, error) {
...
+	feeLimit := walletCfg.Settings["feelimit"]
+	if feeLimit != "" {
+		floatLimit, err := strconv.ParseUint(feeLimit, 10, 32)
+		if err != nil {
+			return nil, err
+		}
+		wallet.feeRateLimit = uint32(floatLimit)
+	} else {
+		// If user didn't provide limit use default limit
+		wallet.feeRateLimit = defaultFeeRateLimit
```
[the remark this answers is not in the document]
amassarwi: Currently `ConfigOpts` works just fine with empty default values, but you are right, using `ConfigOpts` defaults would make it easier to have asset specific defaults, actually that's was my first approach but then users will see defaults in the form instead of empty optional fields so I changed to empty by default and set value in core.

Anyway, will revert that to have asset specific `ConfigOpts` default without optional fields

client/core/core.go
amassarwi: true, that check was there before adding the default will get red of it

amassarwi:
@chappjc Updated.

Moved default value from Core to ConfigOpts to define asset specific values.
However currently user need to save the wallet settings at least one time for
the default values to be used in xcWallet. 
Will try figure out how to force defaults without saving the wallet settings.

client/core/core.go
```diff
@@ -1385,6 +1385,21 @@ func (c *Core) loadWallet(dbWallet *db.Wallet) (*xcWallet, error) {
...
+	// ExchangeWallet instance (see above asset.Setup call) as it required on
+	// order placement time
+	feeLimit := walletCfg.Settings["feeratelimit"]
+	if feeLimit != "" {
+		uintLimit, err := strconv.ParseUint(feeLimit, 10, 64)
+		if err != nil {
+			return nil, err
+		}
+		wallet.feeRateLimit = uintLimit
+	}
```
buck54321, line 1401: Roles are mixed here. These are wallet settings. `Core` doesn't deal with wallet settings. This needs to be handled in the **btc** and **dcr** packages.

Have the `ExchangeWallet` reject these in `FundOrder` based on the `asset.Order.DexConfig.MaxFeeRate`

buck54321 (changes requested):
> However currently user need to save the wallet settings at least one time for the default values to be used in xcWallet.

This would require some kind of `Core`-level upgrade, which isn't implemented yet. You can instead interpret zero as default in the `ExchangeWallet`.

client/core/core.go
```diff
@@ -49,6 +49,9 @@ const (
 	// tickCheckDivisions is how many times to tick trades per broadcast timeout
 	// interval. e.g. 12 min btimeout / 8 divisions = 90 sec between checks.
 	tickCheckDivisions = 8
+	// defaultFeeRateLimit is used to as default when no fee rate limit value
+	// specified by user. unit/kB
+	defaultFeeRateLimit = 10000
```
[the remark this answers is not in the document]
buck54321: For `fallbackfee` we used atoms/kB to match dcrwallet.

client/core/core.go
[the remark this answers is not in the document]
buck54321: I agree. This whole thing is best done at the **core** level, but it's a lot more work. Can switch later. Wallet settings are easy.

client/core/core.go
amassarwi: @chappjc preferred it to be on `xcWallet` instead of each asset's `ExchangeWallet` which was the first approach I considered.  The reason was that we need them on `Core`-level when placing the order.

I see your point regarding mixing roles here tho, as we have all other fields logic on ExchangeWallet level.

chappjc: Only Core deals with this setting though.  I don't see the point of giving this to the ExchangeWallet.

amassarwi: Looks like moving them to `ExchangeWallet` would solve the blocking issue i'm currently having, we would be able
to store the defaults on `ExchangeWallet` without the user saving the settings, having them Settings only and parsing them on Core level would make forcing defaults a messy task.

buck54321: It would be nice to implement this in **core**, but we chose to use wallet settings instead since it's easier, so we should implement this in the wallet.

buck54321: Just as a general rule, `Core` should never try to interpret the `WalletConfig.Settings`.

amassarwi: yep, using defaults already exists for other fields( e.g feeFallback: https://github.com/decred/dcrdex/blob/master/client/asset/dcr/dcr.go#L447-L451)

chappjc: If you want, but ExchangeWallet does not use this setting so it is confusing to have to define it in each backend's `Config` struct and then make accessors so Core can get it.  Why can't Core just pull it out of the map?

chappjc: If this is really a hang up for you guys, we shouldn't use ConfigOpts at all then.  This is completely unrelated to the ExchangeWallet.

amassarwi: Hrmm, that's what it currently does, Settings doesn't seem to store the defaults, so in the above code:
`feeLimit := walletCfg.Settings["feeratelimit"] \ == ""`

chappjc: Sure but you can get the defaults from WalletInfo

amassarwi: I agree that it would be confusing to store it on `ExchangeWallet` and use it only in Core tho,
Let me try to find out if I could somehow pull also the defaults in Core-level

amassarwi: @chappjc good point, sounds reasonable to me!

chappjc: Please follow @buck54321's advice though.  I completely defer to him on decisions of wallet config.  I am just concerned that if we make ExchangeWallet handle the setting in it's own Config, then we have to make getter methods on the ExchangeWallets so that Core can get the value, even though ExchangeWallet doesn't use it internally.

buck54321: If we want to implement this in `Core`, we should, but I'm not looking to use a hybrid approach that breaks important conventions just for convenience. It's not really a wallet setting, but we've chose to make it a wallet setting for now, so let's use it like a wallet setting. It's also easier, I think. The wallet knows it has a value to use. `Core` would have to check every wallet's map for existence. And in principle, we don't control what the asset developer uses that settings key for. We would also be breaking conventions about who defines and interprets settings keys. Before, the wallet settings were for wallet specific settings that the asset developer needed. Now there are wallet settings whose definition and use are dictated by the **core** implementation.

If we choose to implement this in **core** later, we can switch it up. It's no harder to implement this as wallet settings, and it apparently simplifies the task. It appears that the hybrid solutions is a source of complication rather than a convenience anyway.

buck54321: > I am just concerned that if we make ExchangeWallet handle the setting in it's own Config, then we have to make getter methods on the ExchangeWallets so that Core can get the value, even though ExchangeWallet doesn't use it internally.

Core doesn't need the value, does it?

chappjc: It's not a wallet setting, but it is an asset-specific setting.  And because this *has to be implemented in Core*, I guess we need to do this the hard way, without having it in the ConfigOpts.

client/asset/dcr/dcr.go
```diff
@@ -104,6 +106,12 @@ var (
 			Description:  "The fee rate to use for fee payment and withdrawals when estimatesmartfee is not available. Units: DCR/kB",
 			DefaultValue: defaultFee * 1000 / 1e8,
 		},
+		{
+			Key:          "feeratelimit",
+			DisplayName:  "Highest acceptable fee rate",
+			Description:  "This is the highest network fee rate you are willing to pay on swap transactions. If you set this too low, you may not be able to place orders with servers that allow for higher swap rates as network conditions demand.  Units: atoms/byte",
```
JoeGruffins, line 112: Here also.

JoeGruffins (approved):
Working for me.

amassarwi:
@JoeGruffins Updated all strings to follow 80 chars limit.

client/asset/dcr/dcr.go
```diff
@@ -99,24 +101,39 @@ var (
...
 			Key:         "txsplit",
 			DisplayName: "Pre-size funding inputs",
-			Description: "When placing an order, create a \"split\" transaction to fund the order without locking more of the wallet balance than " +
-				"necessary. Otherwise, excess funds may be reserved to fund the order until the first swap contract is broadcast " +
-				"during match settlement, or the order is canceled. This an extra transaction for which network mining fees are paid. " +
-				"Used only for standing-type orders, e.g. limit orders without immediate time-in-force.",
+			Description: "When placing an order, create a \"split\" transaction to " +
+				"fund the order without locking more of the wallet balance than " +
+				"necessary. Otherwise, excess funds may be reserved to fund the order " +
+				"until the first swap contract is broadcast during match settlement, or" +
```
buck54321, line 133: Need a space at the end.

client/asset/btc/btc.go
```diff
@@ -97,24 +99,39 @@ var (
...
 		{
-			Key:          "redeemconftarget",
-			DisplayName:  "Redeem confirmation target",
-			Description:  "The target number of blocks for the redeem transaction to get a confirmation. Used to set the transaction's fee rate. (default: 2 blocks)",
+			Key:         "feeratelimit",
+			DisplayName: "Highest acceptable fee rate",
+			Description: "This is the highest network fee rate you are willing to " +
+				"pay on swap transactions. If you set this too low, you may not be " +
+				"able to place orders with servers that allow for higher swap rates as" +
+				" network conditions demand. Units: sats/byte",
```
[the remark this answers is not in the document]
buck54321: [My previous comment](https://github.com/decred/dcrdex/pull/874#discussion_r545429003) was resolved and maybe a little unclear on my position, but there is now a discrepancy between the `fallbackfee` units and the `feeratelimit` units. For consistency, I do think we should use the same units for both.

buck54321 (changes requested):

client/asset/btc/btc.go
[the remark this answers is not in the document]
chappjc: I'm aware of the consistency consideration, but I feel strongly that feeratelimit should be integer units like sat/bytes.  The consideration is how these settings are used.  fallbackfee matches the node and RPC settings there it takes e.g. btc/kb, while the limit corresponds to e.g. sat/byte as per server setting.  I would must rather have these settings changable without mental gymnastics than have them be consistent.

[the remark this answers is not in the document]
chappjc: Hmm, I guess I have to correct myself.  `fallbackFeeRate` is converted to integer too.  Ugh, well I *hate* the decimal units and all the prefixed zeros, but changing fallbackfee units is too disruptive, so that's not changing now.  I especially dislike btc/kb for this limit since it is not applied in that format anywhere, but if we insist on these two settings have consistent units, Ok.  However, I do not thing consistency between these two settings is a more powerful consideration that clarity and easy of use.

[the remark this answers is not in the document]
amassarwi: Updated the message, @buck54321 are you convinced regarding the integer units ?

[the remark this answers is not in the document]
chappjc: Something to consider: UI could eventually have a units dropdown similar to the Bitcoin Core Qt interface:

image

So it was made that way so user could specify units anyway they like, the question then becomes "what is the natural/native unit to use on the backend?"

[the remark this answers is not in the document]
buck54321: > I especially dislike btc/kb for feeratelimit since it is not applied in that format anywhere and the server config (maxfeerate) is in unit/byte

Those seem like implementation details. As something that hits the UI, the unit chosen should be the one that most users expect or are comfortable with. I also personally think that atoms/byte is the best unit to use for fees, but atoms/kB was the UI unit chosen by bitcoind and dcrwallet, and even in that screenshot from Bitcoin Qt, the **per kB** part is not selectable. I think we're losing this battle.

> However, I do not think consistency between these two settings is a more powerful consideration than clarity and easy of use.

Consistency is not mutually exclusive with clarity and ease of use, and it's easy to argue that different units for the same thing on the same form is a formula for bad UX. What's clear and easy is typically what's expected or conventional, no?


More than anything, I just don't care to fret about this any more. If you're adamant about atoms/byte, go for it.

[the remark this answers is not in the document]
chappjc: Well I was actually trying to talk myself out of this hangup with the possiblity for GUI providing a units selector say some point.  If that seems reasonable to implement at some point I no longer have an issue with starting this with with consistent units on the GUI, if somewhat difficult to reconcile with the backend and corresponding units.
See any reason that couldn't be achieved at some point in GUI @buck54321 ?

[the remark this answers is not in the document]
buck54321: I can see that. Maybe as part of a larger solution to solve our hard-coded `1e8` problem, which will need to be fixed for eth. The backends can define conventional and alternative units, each with their own divisor.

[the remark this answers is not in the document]
amassarwi: Like the sound of that comprehensive solution 😻

amassarwi:
@buck54321 @chappjc I've changed the fee rate limit field type to float and adjusted the code, it's consistent with
fallback fee now.

client/asset/btc/btc.go
```diff
@@ -97,24 +99,39 @@ var (
...
 		{
-			Key:          "redeemconftarget",
-			DisplayName:  "Redeem confirmation target",
-			Description:  "The target number of blocks for the redeem transaction to get a confirmation. Used to set the transaction's fee rate. (default: 2 blocks)",
+			Key:         "feeratelimit",
+			DisplayName: "Highest acceptable fee rate",
+			Description: "This is the highest network fee rate you are willing to " +
+				"pay on swap transactions. If feeratelimit is lower than a market's " +
+				"maxfeerate, you will not be able to trade on that market with this " +
+				"wallet.  Units: sats/byte",
```
buck54321, line 114: Maybe my understanding of our conclusion was wrong, but I though we were going to go to BTC/kB and commit to improving our units handling to allow for user selection.

client/asset/btc/btc.go
```diff
@@ -97,24 +99,39 @@ var (
...
+			DefaultValue: defaultFeeRateLimit * 1000 / 1e8,
+		},
+		{
+			Key:         "redeemconftarget",
+			DisplayName: "Redeem confirmation target",
+			Description: "The target number of blocks for the redeem transaction " +
+				"to get a confirmation. Used to set the transaction's fee rate. " +
```
buck54321, line 121: Not certain here, but would ~~`to get a confirmation`~~ -> `to be mined` be an improvement?

client/asset/btc/btc.go
```diff
@@ -460,12 +480,21 @@ func BTCCloneWallet(cfg *BTCCloneCFG) (*ExchangeWallet, error) {
...
 	if fallbackFeesPerByte == 0 {
 		fallbackFeesPerByte = cfg.DefaultFallbackFee
 	}
-	cfg.Logger.Tracef("Fallback fees set at %d %s/vbyte", fallbackFeesPerByte, cfg.WalletInfo.Units)
+	cfg.Logger.Tracef("Fallback fees set at %d %s/vbyte",
+		fallbackFeesPerByte, cfg.WalletInfo.Units)
+
+	// If set in the user config, the fee rate limit will be in units of BTC/KB.
+	// Convert to sats/byte.
+	feesLimitPerByte := toSatoshi(btcCfg.FeeRateLimit / 1000)
```
buck54321, line 493: Ah. It is expected to be in BTC/kB. Just need to update the `configOpts` description then.

client/asset/btc/btc_test.go
```diff
@@ -391,13 +391,14 @@ func tNewWallet(segwit bool) (*ExchangeWallet, *tRPCClient, func()) {
...
-		WalletInfo:         WalletInfo,
-		DefaultFallbackFee: defaultFee,
-		Segwit:             segwit,
+		WalletCFG:           walletCfg,
+		Symbol:              "btc",
+		Logger:              tLogger,
+		ChainParams:         &chaincfg.MainNetParams,
+		WalletInfo:          WalletInfo,
+		DefaultFallbackFee:  defaultFee,
+		DefaultFeeRateLimit: defaultFeeRateLimit,
```
buck54321, line 400: You'll need to do this in **client/asset/ltc/ltc.go** too. Might be worth investigating why no tests caught that.

client/asset/dcr/dcr.go
```diff
@@ -99,24 +101,39 @@ var (
...
 		{
-			Key:          "redeemconftarget",
-			DisplayName:  "Redeem confirmation target",
-			Description:  "The target number of blocks for the redeem transaction to get a confirmation. Used to set the transaction's fee rate. (default: 1 block)",
+			Key:         "feeratelimit",
+			DisplayName: "Highest acceptable fee rate",
+			Description: "This is the highest network fee rate you are willing to " +
+				"pay on swap transactions. If feeratelimit is lower than a market's " +
+				"maxfeerate, you will not be able to trade on that market with this " +
+				"wallet.  Units: atoms/byte",
```
buck54321, line 116: DCR/kB

client/asset/btc/btc.go
```diff
@@ -460,12 +480,21 @@ func BTCCloneWallet(cfg *BTCCloneCFG) (*ExchangeWallet, error) {
...
+	if feesLimitPerByte == 0 {
+		feesLimitPerByte = cfg.DefaultFeeRateLimit
+	}
```
buck54321, line 496: Is it worth doing this in the constructor and storing the value to a field? Fine the way you have it too though.

buck54321 (changes requested):

client/asset/btc/btc_test.go
amassarwi: Added

amassarwi:
@buck54321 updated mate

buck54321 (approved):

client/asset/btc/btc.go
buck54321, line 496: You might consider rejecting fee rates `1e-8 <= r < 1e-5 `, since the user is clearly trying to set something, and is probably trying to set a low value, but these will truncate to zero, resulting in the default value being used.

chappjc: Good point about rounding to zero.  @amassarwi You'll have to rebase to resolve some conflicts before merge.

amassarwi: Done

amassarwi:
@buck54321 @chappjc updated

client/asset/btc/btc.go
chappjc: Sorry to unresolve this since the PR was so close to done, but I think it just got a little hairier than necessary with the resolution to this that was pushed.

There are two things that I'm not content with now: (1) The new `dcrCfg.FeeRateLimit != 0 && dcrCfg.FeeRateLimit < 1/(smallestUnit/1000)` check could instead be _inferred_ after computing `feesLimitPerByte`, and (2) the error return seems avoidable with the reasonable resolution of rounding up to 1 sat/byte for fee rates <1e-5 DCR/kB.

For dcr, consider:

```go
	feesLimitPerByte := uint64(defaultFeeRateLimit)
	if dcrCfg.FeeRateLimit > 0 {
		feesLimitPerByte = toAtoms(dcrCfg.FeeRateLimit / 1000)
		if feesLimitPerByte == 0 {
			feesLimitPerByte = 1 // round up very low specified rates
		}
	}
```

IMO, we just want to avoid using the default when the specified rate was truncated/rounded to zero when converting units.  I don't have a concern about rounding up (to 1 atom/byte) as a resolution when the specified rate is below 1e-5 DCR/kB.  But I would like to avoid duplicate math with the `smallestUnit` const and the added error return.

Personally, I don't think any user will have a problem with rounding up to the nearest atom/byte, but we're bloating lots of code to add this error return, and adding some error prone and redundant computations.

client/asset/btc/btc.go
```diff
@@ -47,10 +47,14 @@ const (
...
 	// structure.
 	defaultFee = 100
+	// defaultFeeRateLimit is the default value for the feeratelimit.
+	defaultFeeRateLimit = 600
 	// defaultRedeemConfTarget is the default redeem transaction confirmation
 	// target in blocks used by estimatesmartfee to get the optimal fee for a
 	// redeem transaction.
 	defaultRedeemConfTarget = 2
+	// Smallest unit is 1 satoshi = 1e-8 BTC
+	smallestUnit = 1e8
```
buck54321, line 57: This is a little confusing. I see you're trying to use this in two different places, so I assume the reason this is `1e8` instead of `1e-8` like it says in the doc is to make that easier? Regardless, `1e8` is not `small`. It would be more clear to just hard code the `1e8` and `1e-5` in the two places you're using this new constant, I think.

client/asset/btc/btc.go
chappjc: @buck54321 round up to 1 {atom,sat}/byte instead of lots more error return handling sit ok with you?

chappjc: OK, we just spoke offline and decided that actually it's safest to meet `if feesLimitPerByte == 0 {` with an error instead of rounding up to 1.  But let's clean up the math and the variable name as suggested in the other comment.

buck54321: So we did chat a little on matrix. The question on the table is two-fold. 1. Should values lower than `1e-5` be rounded up to `1e-5`, and 2) should it be an error for values < `1e-8`?

I think the safest bet is to assume nothing and error for invalid values < `1e-5`, which are arguably invalid. On the other hand, we could error only for values < `1e-8`, and round values `1e-8 <= r < 1e-5` up to the minimum valid value, `1e-5`, but we are making an assumption about the user's intention in that case, so my preference is to simply error for values `v != 0 && v < 1e-5`.

client/asset/btc/btc.go
```diff
@@ -431,21 +454,41 @@ func BTCCloneWallet(cfg *BTCCloneCFG) (*ExchangeWallet, error) {
...
+		return nil, fmt.Errorf("error creating BTC ExchangeWallet: %v", err)
 	}
 
-	return newWallet(cfg, btcCfg, client), nil
+	return btc, nil
 }
 
 // newWallet creates the ExchangeWallet and starts the block monitor.
-func newWallet(cfg *BTCCloneCFG, btcCfg *dexbtc.Config, node rpcClient) *ExchangeWallet {
+func newWallet(cfg *BTCCloneCFG, btcCfg *dexbtc.Config, node rpcClient) (*ExchangeWallet, error) {
```
buck54321, line 469: Instead of adding an `error` return value here, can you just check the config in `BTCCloneWallet`, which already has the `error`?

client/asset/btc/btc.go
chappjc: We agree about the handling to error for `v != 0 && v < 1e-5`, although it's best to keep the check bulletproof by simply checking as in my suggested code, but instead of defaulting to 1, return an error.  Small final change.

client/asset/btc/btc.go
chappjc: That's a good idea too. Although doing it in the caller, you'd have to keep the arguably redundant and error prone math of doing the check before the `toSatoshi( ... /1000)` that would definitively tell you if you actually hit the error condition.  I won't nit pick much more though.  The PR is very close to done and we can follow up.  Just a little cleanup as feels right for now is good with me.

client/asset/btc/btc.go
amassarwi: 👍 

will do

client/asset/btc/btc.go
amassarwi: done, actually no need to use 1e-5 hard coded anywhere with last commit changes.

chappjc (approved):

buck54321 (approved):

Issue #812, opened 2020-11-04
app: tx rejection when withdrawing dcr - payment of 0 is dust


karamble:
just saw this the first time. when trying to withdraw/empty out the trading account trough the app dcrd gives an error:
`withdraw error: sendrawtx error: -4: wallet.PublishTransaction(...)
is not standard: transaction output 1: payment of 0 is dust, raw tx:`

chappjc:
Withdraw does seem to have a logic issue with dust sometimes.  Apparently affects dcrd, maybe btc too.

chappjc:
Apparently just DCR since BTC's sendtoaddress has a subtract option we use that dcrwallet lacks, necessitating a custom txn.  I will look into this tonight.

PR #875, opened 2020-12-08
fix change handling in sendWithReturn when subtracting from existing out

chappjc:
Resolves https://github.com/decred/dcrdex/issues/812

### DIFF
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -1175,16 +1175,10 @@ func (dcr *ExchangeWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin
 		return nil, nil, 0, fmt.Errorf("fewer outputs than swaps. %d < %d", len(baseTx.TxOut), swapCount)
 	}
 
-	// Grab a change address.
-	changeAddr, err := dcr.node.GetRawChangeAddress(dcr.ctx, dcr.acct, chainParams)
-	if err != nil {
-		return nil, nil, 0, fmt.Errorf("error creating change address: %w", translateRPCCancelErr(err))
-	}
-
 	// Add change, sign, and send the transaction.
 	dcr.fundingMtx.Lock()         // before generating change output
 	defer dcr.fundingMtx.Unlock() // hold until after returnCoins and lockFundingCoins(change)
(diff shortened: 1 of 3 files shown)


chappjc:
This is going to need careful review and thorough testing.  Still, I'm considering 0.1.4 for this given the [DCR withdraw bug](https://github.com/decred/dcrdex/issues/812) is solved by this.

chappjc:
TODO: if the withdraw amount is the entire balance, the dcr wallet backend could use the `sweepaccount` RPC for robustness.

client/asset/dcr/dcr.go
```diff
@@ -2195,64 +2179,85 @@ func (dcr *ExchangeWallet) signTx(baseTx *wire.MsgTx) (*wire.MsgTx, error) {
...
+			if err != nil {
+				return nil, nil, "", 0, err
+			}
+			dcr.log.Debugf("Change output size = %d, addr = %s", changeOutput.SerializeSize(), changeAddress.String())
+			changeAdded = true
+			baseTx.AddTxOut(changeOutput) // unsigned txn
+			remaining -= changeValue
 		} else {
-			changeOutput.Value -= int64(minFee + changeFees)
+			fmt.Println("DUSTING IT", changeValue, remaining, minFeeWithChange)
```
chappjc, line 2248: will remove

client/asset/dcr/dcr.go
```diff
@@ -2195,64 +2179,85 @@ func (dcr *ExchangeWallet) signTx(baseTx *wire.MsgTx) (*wire.MsgTx, error) {
...
-		size += changeSize
-		lastFee = feeRate * uint64(size)
-		subtractee.Value = int64(reservoir - lastFee)
+	// If change added or subtracting from an existing output, iterate on fees.
+	if changeAdded || subtractFrom >= 0 {
+		subtractee := changeOutput
+		if subtractFrom >= 0 {
+			subtractee = baseTx.TxOut[subtractFrom]
+		}
+		reservoir := remaining + uint64(subtractee.Value)
```
chappjc, line 2260: will comment on this in the code, but this reasoning is that `reservoir` is what is available for fees, `remaining` is what is presently already free and thus would be fees, and `subtractee.Value` is additional amount that can be add to fees or absorb excess fees from `remaining`.

buck54321 (approved):

Issue #382, opened 2020-05-18
core: read out trade history in a portable format


buck54321:
The **db** package allows reading out orders and matches. It would be useful to format the info in CSV for importing into another program, e.g. spreadsheet. For the GUI, a history view could be provided for filtering and viewing the data, and a download button for exporting to CSV. Probably want an export method on the RPC server as well.

chappjc:
Assigned to @skynet0590

martonp:
I'll pick this up.

PR #877, opened 2020-12-09
implement export orders button on orders page

skynet0590:
Resolve issue #382

### DIFF
--- a/client/webserver/http.go
+++ b/client/webserver/http.go
@@ -231,6 +231,8 @@ func (s *WebServer) handleExportOrders(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
 		return
 	}
+	w.Header().Set("Content-Disposition", "attachment; filename=orders.csv")
+	w.Header().Set("Content-Type", "text/csv")
 	csvWriter := csv.NewWriter(w)
 	err = csvWriter.Write([]string{
 		"Trade",
@@ -277,8 +279,6 @@ func (s *WebServer) handleExportOrders(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 	}
-	w.Header().Set("Content-Disposition", "attachment; filename=orders.csv")
-	w.Header().Set("Content-Type", "text/csv")
 	csvWriter.Flush()
 }
 
--- a/client/webserver/http.go
+++ b/client/webserver/http.go
@@ -4,6 +4,7 @@
 package webserver
 
 import (
+	"bytes"
 	"encoding/csv"
 	"fmt"
 	"io"
@@ -231,9 +232,10 @@ func (s *WebServer) handleExportOrders(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
 		return
 	}
-	w.Header().Set("Content-Disposition", "attachment; filename=orders.csv")
-	w.Header().Set("Content-Type", "text/csv")
-	csvWriter := csv.NewWriter(w)
+
+	var buffer bytes.Buffer
+	csvWriter := csv.NewWriter(&buffer)
+
 	err = csvWriter.Write([]string{
 		"Trade",
 		"Market",
@@ -242,6 +244,7 @@ func (s *WebServer) handleExportOrders(w http.ResponseWriter, r *http.Request) {
 		"Rate",
 		"Filled (%)",
 		"Settled (%)",
+		"Quantity",
 		"Time",
 	})
 	if err != nil {
@@ -251,27 +254,20 @@ func (s *WebServer) handleExportOrders(w http.ResponseWriter, r *http.Request) {
 	for _, ord := range ords {
 		ordReader := orderReader{ord}
 
-		rate := float64(ord.Rate) / 1e8
-		tradeStr := fmt.Sprintf("%.8f %s → %.8f %s",
-			ordReader.FromQty(), ordReader.FromSymbol(),
-			ordReader.ToQty(), ordReader.ToSymbol())
-
-		typeStr := ord.Type.String() + " "
-		if ord.Sell {
-			typeStr += "sell"
-		} else {
-			typeStr += "buy"
-		}
+		tradeStr := fmt.Sprintf("%s %s → %s %s",
+			ordReader.OfferString(), ordReader.FromSymbol(),
+			ordReader.AskString(), ordReader.ToSymbol())
 
 		srvTime := encode.UnixTimeMilli(int64(ord.Stamp))
 		err = csvWriter.Write([]string{
 			tradeStr, // Trade
 			fmt.Sprintf("%s-%s @ %s", ord.BaseSymbol, ord.QuoteSymbol, ord.Host), // Market
-			typeStr,                                   // Type
-			ord.Status.String(),                       // Status
-			fmt.Sprintf("%.8f", rate),                 // Rate
+			ordReader.TypeString(),                    // Type
+			ordReader.StatusString(),                  // Status
+			ordReader.RateString(),                    // Rate
 			ordReader.FilledPercent(),                 // Filled
 			ordReader.SettledPercent(),                // Settled
+			ordReader.OfferString(),                   // Quantity
 			srvTime.Format("02/01/2006, 03:04:05 PM"), // Time
 		})
 		if err != nil {
@@ -280,6 +276,15 @@ func (s *WebServer) handleExportOrders(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 	csvWriter.Flush()
+	data := buffer.Bytes()
+	w.Header().Set("Content-Disposition", "attachment; filename=orders.csv")
+	w.Header().Set("Content-Type", "text/csv")
+	w.Header().Set("Content-Length", fmt.Sprintf("%d", len(data)))
+	_, err = w.Write(data)
+	if err != nil {
+		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+		return
+	}
 }
 
 type orderTmplData struct {
--- a/client/webserver/http.go
+++ b/client/webserver/http.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"sort"
 	"strconv"
+	"strings"
 
 	"decred.org/dcrdex/client/core"
 	"decred.org/dcrdex/dex"
@@ -201,6 +202,7 @@ func (s *WebServer) handleExportOrders(w http.ResponseWriter, r *http.Request) {
 	filter := new(core.OrderFilter)
 	err := r.ParseForm()
 	if err != nil {
+		log.Errorf("error parsing form for export order: %v", err)
 		http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
 		return
 	}
@@ -211,6 +213,7 @@ func (s *WebServer) handleExportOrders(w http.ResponseWriter, r *http.Request) {
 	for k, assetStrId := range assets {
 		assetNumId, err := strconv.Atoi(assetStrId)
 		if err != nil {
+			log.Errorf("error parsing asset id: %v", err)
 			http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
 			return
 		}
@@ -222,6 +225,7 @@ func (s *WebServer) handleExportOrders(w http.ResponseWriter, r *http.Request) {
 		statusNumId, err := strconv.Atoi(statusStrId)
 		if err != nil {
 			http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
+			log.Errorf("error parsing status id: %v", err)
 			return
 		}
 		filter.Statuses[k] = order.OrderStatus(statusNumId)
@@ -229,16 +233,22 @@ func (s *WebServer) handleExportOrders(w http.ResponseWriter, r *http.Request) {
 
 	ords, err := s.core.Orders(filter)
 	if err != nil {
+		log.Errorf("error retrieving order: %v", err)
 		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
 		return
 	}
 
 	var buffer bytes.Buffer
 	csvWriter := csv.NewWriter(&buffer)
+	// UseCRLF for Window users
+	csvWriter.UseCRLF = strings.Contains(r.UserAgent(), "Windows")
 
 	err = csvWriter.Write([]string{
-		"Trade",
-		"Market",
+		"Host",
+		"Base",
+		"Quote",
+		"Ask",
+		"Offer",
 		"Type",
 		"Status",
 		"Rate",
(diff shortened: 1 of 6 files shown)


client/webserver/http.go
```diff
@@ -190,6 +195,96 @@ func (s *WebServer) handleOrders(w http.ResponseWriter, r *http.Request) {
 	})
 }
 
+// handleOrders is the handler for the /orders/export page request.
```
chappjc, line 198: handleOrders -> handleExportOrders

client/webserver/http.go
```diff
@@ -190,6 +195,96 @@ func (s *WebServer) handleOrders(w http.ResponseWriter, r *http.Request) {
...
+					settled += float64(match.Qty)
+				}
+			}
+		}
+		srvTime := encode.UnixTimeMilli(int64(ord.Stamp))
+		err = csvWriter.Write([]string{
+			tradeStr, // Trade
+			fmt.Sprintf("%s-%s @ %s", ord.BaseSymbol, ord.QuoteSymbol, ord.Host), // Market
+			typeStr,                       // Type
+			ord.Status.String(),           // Status
```
chappjc, line 271: An unrecognized order status value will panic in String, although that shouldn't happen unless there's DB corruption. Still, I think it's time to change `(OrderStatus) String()` in dex/order/status.go so it does not panic and instead returns "invalid".

chappjc (changes requested):
I think the `orderReader` and `matchReader` types in client/webserver/types.go should be used for most of the logic and conversions you're doing manually.  e.g. `(*orderReader).FilledPercent()` does a lot of the error prone hard work your'e doing here.  @buck54321 actually added those types for the order details web page and I think it's directly applicable here.

client/webserver/http.go
```diff
@@ -190,6 +195,96 @@ func (s *WebServer) handleOrders(w http.ResponseWriter, r *http.Request) {
...
+	assets := r.Form["assets"]
+	filter.Assets = make([]uint32, len(assets))
+	for k, assetId := range assets {
+		filter.Assets[k] = uint32(parseStrToInt(assetId))
+	}
+	statuses := r.Form["statuses"]
+	filter.Statuses = make([]order.OrderStatus, len(statuses))
+	for k, statusId := range statuses {
+		filter.Statuses[k] = order.OrderStatus(parseStrToInt(statusId))
+	}
```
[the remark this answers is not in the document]
skynet0590: I know it is easier way to deal with parsing data and available solution. But as my knowledge of web. The browser only support for Get method when open a new browser to download a file. But I will do some research to give better solution if it is possible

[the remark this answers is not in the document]
chappjc: Not that it must be post, but the example of apiOrders and the frontend post to /api/orders should be clear.

I think using the orderReader and matchReader methods are more important to focus on for preparing the strings reliably.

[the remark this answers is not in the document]
skynet0590: Thank you. I will apply them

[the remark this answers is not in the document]
skynet0590: the frontend post to /api/oders. So we will generate the file from the frontend? To be honest it is not a normally way I have seen. I usually see that the file is generated on the server

client/webserver/http.go
```diff
@@ -190,6 +195,96 @@ func (s *WebServer) handleOrders(w http.ResponseWriter, r *http.Request) {
...
+		})
+		if err != nil {
+			s.writeAPIError(w, "export error: %v", err)
+			return
+		}
+	}
+	timeName := "orders"
+	w.Header().Set("Content-Disposition", "attachment; filename="+timeName+".csv")
+	w.Header().Set("Content-Type", "text/csv")
+	csvWriter.Flush()
```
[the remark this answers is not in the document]
skynet0590: If put the `csvWriter.Flush()` before the headers are set. The content will be not downloaded and simply is text display on the browser. I tested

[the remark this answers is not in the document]
chappjc: Flush is just whatever is still buffered.  The headers should be before all the writes (each `csvWriter.Write`), not just the flush.

client/webserver/http.go
```diff
@@ -190,6 +195,96 @@ func (s *WebServer) handleOrders(w http.ResponseWriter, r *http.Request) {
...
+	}
+	statuses := r.Form["statuses"]
+	filter.Statuses = make([]order.OrderStatus, len(statuses))
+	for k, statusId := range statuses {
+		filter.Statuses[k] = order.OrderStatus(parseStrToInt(statusId))
+	}
+
+	ords, err := s.core.Orders(filter)
+	if err != nil {
+		s.writeAPIError(w, "Orders error: %v", err)
```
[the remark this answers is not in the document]
skynet0590: Yes I think it should be a web view.

client/webserver/http.go
skynet0590: I change it to returns "invalid" but it is not matched the test. So I revert it to the original. May be need someone else change the test first

client/webserver/http.go
```diff
@@ -190,6 +196,97 @@ func (s *WebServer) handleOrders(w http.ResponseWriter, r *http.Request) {
...
+// handleExportOrders is the handler for the /orders/export page request.
+func (s *WebServer) handleExportOrders(w http.ResponseWriter, r *http.Request) {
+	filter := new(core.OrderFilter)
+	err := r.ParseForm()
+	if err != nil {
+		http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
+		return
+	}
+
+	filter.Hosts = r.Form["hosts"]
```
buck54321, line 208: If we need to send data in the query string, let's use `chi.URLParam` to retrieve the data as we do elsewhere. This would also enable us to reuse the encoding code from `submitFilter` in `exportOrders`. Just gotta do a `strings.Split` instead.

client/webserver/http.go
```diff
@@ -190,6 +196,97 @@ func (s *WebServer) handleOrders(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	ords, err := s.core.Orders(filter)
+	if err != nil {
+		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+		return
+	}
+
+	var buffer bytes.Buffer
+	csvWriter := csv.NewWriter(&buffer)
```
buck54321, line 237: In dcrdata, starting [here](https://github.com/decred/dcrdata/blob/967628e8f638c7a673ac3f2b24195b33035d5855/explorer/explorerroutes.go#L1373) and ending [here](https://github.com/decred/dcrdata/blob/967628e8f638c7a673ac3f2b24195b33035d5855/api/apiroutes.go#L304), we are setting a special line feed for Windows clients. Should we be doing the same here?

buck54321 (changes requested):
The export button is dropped a little unceremoniously at the bottom of the filter list. How about making it a button?

image

The way that browsers handle file downloads seemingly necessitates the use of either query parameters or form data. I do think there are better ways to do this, but none quite as straightforward. This is a decent first approach.

client/webserver/site/src/js/orders.js
```diff
@@ -179,6 +183,40 @@ export default class OrdersPage extends BasePage {
...
   }
 
+  /* exportOrders open a new page to export orders based on the current filter */
+  exportOrders () {
+    const page = this.page
+    this.offset = ''
+    const filterState = this.filterState
+    filterState.hosts = parseSubFilter(page.hostFilter)
+    filterState.assets = parseSubFilter(page.assetFilter)
+    filterState.statuses = parseSubFilter(page.statusFilter)
```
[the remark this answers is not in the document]
skynet0590: Yes. Thank you

client/webserver/http.go
skynet0590: Somewhere are using `chi.URLParam` because in that place there are some `must have` parameter. While in here all the parameters are optional. So I think it is harder if we use that way.

client/webserver/http.go
skynet0590: okay

client/webserver/site/src/js/orders.js
```diff
@@ -179,6 +183,40 @@ export default class OrdersPage extends BasePage {
...
+      const subFilter = filterState[k]
+      search.delete(k)
+      if (subFilter.length !== 0) {
+        if (Array.isArray(subFilter)) {
+          subFilter.forEach(e => {
+            search.append(k, e)
+          })
+        } else {
+          search.set(k, subFilter)
+        }
```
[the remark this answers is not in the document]
skynet0590: Okay. I will fix it

client/webserver/http.go
```diff
@@ -190,6 +196,97 @@ func (s *WebServer) handleOrders(w http.ResponseWriter, r *http.Request) {
...
+		srvTime := encode.UnixTimeMilli(int64(ord.Stamp))
+		err = csvWriter.Write([]string{
+			tradeStr, // Trade
+			fmt.Sprintf("%s-%s @ %s", ord.BaseSymbol, ord.QuoteSymbol, ord.Host), // Market
+			ordReader.TypeString(),                    // Type
+			ordReader.StatusString(),                  // Status
+			ordReader.RateString(),                    // Rate
+			ordReader.FilledPercent(),                 // Filled
+			ordReader.SettledPercent(),                // Settled
+			ordReader.OfferString(),                   // Quantity
```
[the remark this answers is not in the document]
skynet0590: Thank you

JoeGruffins:
This pr must be rebased onto master. It is currently impossible to review changes. https://git-scm.com/docs/git-rebase

chappjc:
It's not clear what happened to make all those extra commits, but you'll want to squash them all away.  With a current `master` branch:

```
git checkout 382_trade_history_in_a_portable_format
git rebase -i master
```

Then you'll want to squash down all the strange commits like:

```
pick c35fd3b3 implement export orders button on orders page
pick 80d4cb88 fix follow test script
pick c0c50173 export orders: fix display error with web header inst
pick 4f1f9c93 revert order.Status return panic with invalid value s
pick 8d13c31e move set header to before csv Write
pick fbd6b70f export orders: write the content to a buffer to avoid
pick 755caa30 gofmt
pick 70e2415d export order: add log error, add UseCRLF into csvWrit
fix 25c73392 fix conflict
fix 1483fed3 implement export orders button on orders page
fix c1563f68 fix follow test script
fix a175996c export orders: fix display error with web header inste
fix 1cfb70eb revert order.Status return panic with invalid value si
fix 143ca6a2 move set header to before csv Write
fix ebd732cd export orders: write the content to a buffer to avoid 
fix 342456ad implement export orders button on orders page
fix 8ea19cdb fix follow test script
fix 8373e881 export orders: fix display error with web header inste
fix d5014579 revert order.Status return panic with invalid value si
fix 266be02b export orders: write the content to a buffer to avoid 
fix 2a020def implement export orders button on orders page
fix f18cdb9e fix follow test script
fix 0cf2d981 export orders: fix display error with web header inste
fix a41279d1 revert order.Status return panic with invalid value si
fix ee125a94 move set header to before csv Write
fix 2a500509 export orders: write the content to a buffer to avoid 
```

client/webserver/http.go
```diff
@@ -190,6 +197,109 @@ func (s *WebServer) handleOrders(w http.ResponseWriter, r *http.Request) {
...
+	csvWriter.Flush()
+	data := buffer.Bytes()
+	w.Header().Set("Content-Disposition", "attachment; filename=orders.csv")
+	w.Header().Set("Content-Type", "text/csv")
+	w.Header().Set("Content-Length", fmt.Sprintf("%d", len(data)))
+
+	_, err = w.Write(data)
+	if err != nil {
+		log.Errorf("error writing to client body: %v", err)
+		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
```
chappjc, line 298: At this point if there is an error from `http.ResponseWriter.Write`, it's too late to call `http.Error` (already written the header).  Just log an let the function return.

client/webserver/http.go
```diff
@@ -190,6 +197,109 @@ func (s *WebServer) handleOrders(w http.ResponseWriter, r *http.Request) {
...
+			ordReader.OfferString(),                   // Quantity
+			srvTime.Format("02/01/2006, 03:04:05 PM"), // Time
+		})
+		if err != nil {
+			log.Errorf("error writing CSV: %v", err)
+			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+			return
+		}
+	}
+	csvWriter.Flush()
```
chappjc, line 289: Poking in the [`encoding/csv.Writer docs`](https://golang.org/pkg/encoding/csv/#Writer.Error), it looks like we're supposed to call `csvWriter.Error()` after `Flush`:

> After all data has been written, the client should call the Flush method to guarantee all data has been forwarded to the underlying io.Writer. Any errors that occurred should be checked by calling the Error method.

> Flush writes any buffered data to the underlying io.Writer. To check if an error occurred during the Flush, call Error.

client/webserver/site/src/js/orders.js
```diff
@@ -179,6 +183,40 @@ export default class OrdersPage extends BasePage {
...
+    const page = this.page
+    this.offset = ''
+    const filterState = this.filterState
+    filterState.hosts = parseSubFilter(page.hostFilter)
+    filterState.assets = parseSubFilter(page.assetFilter)
+    filterState.statuses = parseSubFilter(page.statusFilter)
+
+    const url = new URL(window.location)
+    const search = new URLSearchParams(url.search)
+    search.delete('offset')
```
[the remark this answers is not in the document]
chappjc: Is the idea because the _export_ http handler only recognizes `'hosts'`, `'assets'`, and `'statuses'`?  If so, that means it doesn't use `'offset'` **or** `'n'`.  It does seem to be used in `exportOrders` via `window.open(url.toString())`.  Anyway, that's not the case in `submitFilter`, where it's doing a POST with `currentFilter` to get the request body... @buck54321 maybe you had originally written `submitFilter` using url query parameters instead of a POST?

chappjc (changes requested):
Looking good, some minor nits, but mostly working.

image

image

Market buy filled % is wrong.  Here's the details of that market buy:

image

Also, as @buck54321 noted, perhaps the rate data could be purely numeric.  The units of the rate are known from the base and quote columns.

client/webserver/http.go
buck54321: Not sure that I follow. `chi.URLParam` returns an empty string if the parameter is not set.

client/webserver/http.go
```diff
@@ -190,6 +197,109 @@ func (s *WebServer) handleOrders(w http.ResponseWriter, r *http.Request) {
...
+			ord.QuoteSymbol,                           // Quote
+			ordReader.AskString(),                     // Ask
+			ordReader.OfferString(),                   // Offer
+			ordReader.TypeString(),                    // Type
+			ordReader.StatusString(),                  // Status
+			ordReader.RateString(),                    // Rate
+			ordReader.FilledPercent(),                 // Filled
+			ordReader.SettledPercent(),                // Settled
+			ordReader.OfferString(),                   // Quantity
+			srvTime.Format("02/01/2006, 03:04:05 PM"), // Time
```
buck54321, line 281: We want this to be both parseable and human-readable if possible. Let's use a standardized format. Consider using `MarshalText` to get an RFC 3339 format.

client/webserver/site/src/html/orders.tmpl
```diff
@@ -35,6 +35,9 @@
         {{end}}
         <div class="apply-bttn d-hide"><button class="bg2 demi">apply</button></div>
       </div>
+      <div id="exportOrders" class="filter-display bg1">
+        <a class="ico-open fs14 pointer plainlink"></a> Export
+      </div>
```
buck54321, line 40: image

You've styled it as if it's a header. Please style it as if it's a button similar to other buttons in the client.

buck54321 (changes requested):

client/webserver/http.go
chappjc: I think we might be miscommunicating here.  It does seem like path components make for a tricky interface and implementation here.  What path format are you envisioning, @buck54321?
URLParam with path parts would have the request be something like:
`/orders/export/hosts/dex.decred.org,127.0.0.1/assets/dcr,btc/statuses/`
but it feels more natural to use url query parmeters like `/orders/export/?hosts=dex.decred.org,127.0.0.1&assets=dcr,btc`
(well, I'm not sure if you'd use commas or repeat parameters actually)

buck54321: You're absolutely right. I was confusing query parameters with url parameters. I guess there is `http.Request.URL.Query()`, but your way is fine, @skynet0590. Sorry about that.

client/webserver/http.go
skynet0590: Thank you

chappjc:
Let us know if you are stuck @skynet0590.

buck54321:
This feature has been requested, so people want this. If anyone else wants to pick this up, we can close this PR for inactivity.

chappjc:
Was just thinking of this.  It's up for grabs guys.  Just grab skynet's commits from https://github.com/skynet0590/dcrdex/commits/382_trade_history_in_a_portable_format, squash, and continue.

PR #882, opened 2020-12-11
fix logs showing ipkey instead of address string

chappjc:
This fixes some log statements in dex/ws that were showing the illegible ip key.

NOTE: I'm still considering changing `dex.IPKey` to a full `net.IPv6len` array and using a CIDR mask in `NewIPKey` instead of a partial copy so that the `IPKey` can be converted to an IP string, right now it is an irreversible conversion from a `net.IP` to a `dex.IPKey` (can no longer distinguish ipv4 from ipv6).  Making this change would allow passing only the `IPKey` (or a `net.IP`) to the `WSLink` constructor, and the address string could be extracted on construction instead of passed as another input.

### DIFF
--- a/dex/ip.go
+++ b/dex/ip.go
@@ -5,6 +5,7 @@ package dex
 
 import (
 	"net"
+	"strings"
 )
 
 // IPKey is a IP address byte array.
@@ -16,18 +17,21 @@ func NewIPKey(addr string) IPKey {
 	host, _, err := net.SplitHostPort(addr)
 	if err == nil && host != "" {
 		addr = host
+	} else {
+		// If SplitHostPort failed, IPv6 addresses may still have brackets.
+		addr = strings.Trim(addr, "[]")
 	}
 
 	ip := net.ParseIP(addr)
 	if ip == nil {
-		return IPKey{}
+		return IPKey{} // i.e. net.IPv6unspecified
 	}
 	// IPv4 is encoded in a net.IP of length IPv6len as
 	// 00:00:00:00:00:00:00:00:00:00:ff:ff:xx:xx:xx:xx
 	// Thus, must copy all 16 bytes for IPv4.
 	N := net.IPv6len
-	if ip.To4() == nil {
-		// Drop the last 64 bits (interface) of IPv6 addresses.
+	if ip.To4() == nil && !ip.Equal(net.IPv6loopback) {
+		// Drop the last 64 bits (interface) of non-loopback IPv6 addresses.
 		N = net.IPv6len / 2 // i.e. ip = ip.Mask(net.CIDRMask(64, 128))
 	}
 	var ipKey IPKey
--- a/dex/ip.go
+++ b/dex/ip.go
@@ -7,22 +7,38 @@ import (
 	"net"
 )
 
-// IPKey is a IP address byte array. For an IPV6 address, the IPKey drops the
-// interface identifier, which is the second half of the address.
-type IPKey [net.IPv6len / 2]byte
+// IPKey is a IP address byte array.
+type IPKey [net.IPv6len]byte
 
-// NewIPKey parses an IP address string into an IPKey.
+// NewIPKey parses an IP address string into an IPKey. For an IPV6 address, this
+// drops the interface identifier, which is the second half of the address
 func NewIPKey(addr string) IPKey {
 	host, _, err := net.SplitHostPort(addr)
 	if err == nil && host != "" {
 		addr = host
 	}
-	netIP := net.ParseIP(addr)
-	ip := netIP.To4()
+
+	ip := net.ParseIP(addr)
 	if ip == nil {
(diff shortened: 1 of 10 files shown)


buck54321 (approved):

dex/ws/wslink.go
```diff
@@ -354,7 +354,7 @@ func (c *WSLink) outHandler(ctx context.Context) {
 		// need to drain it, just the outQueue so SendNow never hangs.
 
 		c.log.Tracef("Sent %d and dropped %d messages to %v before shutdown.",
-			writeCount, lostCount, c.ip)
+			writeCount, lostCount, c.addr)
 	}()
```
JoeGruffins, line 358: I'm not seeing anything locally now. Doesn't output with loopback?
```
2020-12-18 16:19:42.443 [DBG] AUTH: Link down: id=2, ip=.
2020-12-18 16:19:42.443 [TRC] COMM[WS]: Sent 5 and dropped 0 messages to  before shutdown.
```

chappjc: Bug.  Going back in...

dex/ws/wslink.go
```diff
@@ -88,9 +86,9 @@ type sendData struct {
 }
 
 // NewWSLink is a constructor for a new WSLink.
-func NewWSLink(ip dex.IPKey, addr string, conn Connection, pingPeriod time.Duration, handler func(*msgjson.Message) *msgjson.Error, logger dex.Logger) *WSLink {
+func NewWSLink(addr string, conn Connection, pingPeriod time.Duration, handler func(*msgjson.Message) *msgjson.Error, logger dex.Logger) *WSLink {
 	return &WSLink{
-		ip:         ip,
+		addr:       addr,
```
chappjc, line 91: `addr` was dropped.  However, only `server/comms.Server` needs the `dex.IPKey` for rate limiting and banishing.

dex/ws/wslink_test.go
```diff
@@ -115,7 +115,8 @@ func TestWSLink_handleMessageRecover(t *testing.T) {
 		inMsg: make(chan []byte, 1),
 		inErr: make(chan error, 1),
 	}
-	wsLink := NewWSLink(dex.IPKey{}, "addr", conn, time.Second, inMsgHandler, tLogger)
+	ipk := dex.IPKey{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255, 255 /* ipv4 */, 127, 0, 0, 1}
+	wsLink := NewWSLink(ipk.String(), conn, time.Second, inMsgHandler, tLogger)
```
chappjc, line 119: As I mentioned, dex.IPKey is only needed in server/comms, but this is here for an example in the test.

dex/ws/wslink.go
chappjc: Sorry, guys, I decided to rework this a little more thoroughly.

Logging is now as before:

dcrdex (client accessing)

```
[DBG] COMM: Starting websocket handler for 127.0.0.1:37068
[TRC] COMM: New websocket client 127.0.0.1
[TRC] COMM[WS]: Starting websocket messaging with peer 127.0.0.1
...
[TRC] COMM[WS]: Sent 11 and dropped 0 messages to 127.0.0.1 before shutdown.
[DBG] AUTH: Link down: id=0, ip=127.0.0.1.
[TRC] COMM: Disconnected websocket client 127.0.0.1
```

dexc (browser accessing)

```
[DBG] WEB[WS]: New websocket client 127.0.0.1:55064
[TRC] WEB[WS][127.0.0.1:55064]: Starting websocket messaging with peer 127.0.0.1:55064
...
[TRC] WEB[WS][127.0.0.1:55064]: Sent 4 and dropped 0 messages to 127.0.0.1:55064 before shutdown.
[TRC] WEB[WS]: Disconnected websocket client 127.0.0.1:55064
```

I changed `dex.IPKey` so the address is recoverable, and I limited its use to `server/comms.Server` where it is actually used for rate limiting and banishment.

dex/ip.go
```diff
@@ -16,18 +17,21 @@ func NewIPKey(addr string) IPKey {
...
-		return IPKey{}
+		return IPKey{} // i.e. net.IPv6unspecified
 	}
 	// IPv4 is encoded in a net.IP of length IPv6len as
 	// 00:00:00:00:00:00:00:00:00:00:ff:ff:xx:xx:xx:xx
 	// Thus, must copy all 16 bytes for IPv4.
 	N := net.IPv6len
-	if ip.To4() == nil {
-		// Drop the last 64 bits (interface) of IPv6 addresses.
+	if ip.To4() == nil && !ip.Equal(net.IPv6loopback) {
```
chappjc, line 33: So the interface bits aren't masked out if the address is IPv6 loopback.

JoeGruffins (approved):

buck54321 (approved):

Issue #722, opened 2020-10-06
client: account private key import/export


chappjc:
Backing up dexc.db is the only way presently to backup a DEX account.  We should allow a user to export and import the private key for a DEX account.

Related, a seed mnemonic could be used, but that is a bigger change, and breaking: https://github.com/decred/dcrdex/issues/434

kevinstl:
For the export we can provide a new endpoint in webserver.go:

```
r.Get("/keys", s.keys)
```

That endpoint will respond with a json file for download with the keys included:

```
[
   {
      "host": "dex.decred.org",
      "accountID": "hexstringhereadsfadf",
      "privKey": "hexstringhereadsfadf",
      "pubKey": "eifoiwfjoskldfjlfjfvlkldjksfkjsfdjknhheyrnn"
   }
]
```

kevinstl:
Per @chappjc the /keys endpoint will both be behind the auth middleware and require the app pass.

PR #885, opened 2020-12-12
client: account private key import/export

kevinstl:
This pull request addresses: https://github.com/decred/dcrdex/issues/722

The user will be able to export and import account info from the settings page.


kevinstl:
Here is the current state of the settings page with export:

image

The account file will download when the user clicks "Export Keys".

Let me know if there any issues with the export capability so far.

client/core/core.go
```diff
@@ -2565,6 +2565,25 @@ func (c *Core) Trade(pw []byte, form *TradeForm) (*Order, error) {
 	return corder, nil
 }
 
+// Trade is used to place a market or limit order.
+func (c *Core) AccountKeys(pw []byte, host string) (*AccountKeysResponse, error) {
```
chappjc, line 2569: `pw` is not used.  I was thinking a password check like in https://github.com/decred/dcrdex/pull/885/files#diff-13a4a6b12c1ae844ef90151638f194427cfdc545fceb58d52c0f6f33b81d859bR2528, but throwing out the returned crypter since it's not used to sign anything.  But perhaps AccountKeys does not need a password as long as the api methods ensure the user is authenticated to call it.  I just figure since this is a sensitive operation it should be password protected.

client/core/types.go
```diff
@@ -695,6 +695,12 @@ type OrderFilter struct {
 	Statuses []order.OrderStatus `json:"statuses"`
 }
 
+// AccountKeysResponse holds data returned from AccountKeys.
+type AccountKeysResponse struct {
+	Host    string `json:"host"`
+	PrivKey []byte `json:"privKey"`
```
chappjc, line 701: Should be a hex string or something that marshals to a hex string.

client/webserver/webserver.go
```diff
@@ -268,6 +269,7 @@ func New(core clientCore, addr string, logger dex.Logger, reloadHTML bool) (*Web
 			apiAuth.Post("/withdraw", s.apiWithdraw)
 			apiAuth.Post("/maxbuy", s.apiMaxBuy)
 			apiAuth.Post("/maxsell", s.apiMaxSell)
+			apiAuth.Post("/accountKeys", s.apiAccountKeys)
```
chappjc, line 272: Seeing as it is under the `rejectUnauthed`, I suppose we can eliminate the password argument to the Core method.

chappjc (changes requested):

client/core/core.go
kevinstl: It's checking the password prior to export now.

client/core/core.go
```diff
@@ -2565,6 +2566,29 @@ func (c *Core) Trade(pw []byte, form *TradeForm) (*Order, error) {
...
```
JoeGruffins, line 2570: Comment is from a different method. Trade -> AccountKeys

client/core/core.go
```diff
@@ -2565,6 +2566,29 @@ func (c *Core) Trade(pw []byte, form *TradeForm) (*Order, error) {
...
+
+	// Get the dexConnection and the dex.Asset for each asset.
+	c.connMtx.RLock()
+	dc, found := c.conns[host]
+	c.connMtx.RUnlock()
+	if !found {
+		return nil, fmt.Errorf("unknown DEX %s", host)
+	}
+
+	accountKeysResponse := &AccountKeysResponse{Host: host, PrivKey: hex.EncodeToString(dc.acct.privKey.Serialize()), OK: true}
```
JoeGruffins, line 2588: Since OK is never false, I don't think it's needed.

JoeGruffins (commented):
So, if someone got this key's plaintext, what's the worst they could do? I think they could impersonate you and get you banned. I don't think they could use it in any way to get at your funds though. Still, I think it might be best to encrypt the key with either the appPass, or with a new password upon exporting and require that upon importing. It just seems like good practice to me...

I'm not sure of the best way to encrypt though. I don't think our present system will let you decrypt something unless you have the key stored in the database as well, so it can't be used for importing into a new database.

edit: Oops, this was still in draft. Srry if it wasn't ready to be looked at.

kevinstl:
The import account is functioning.

It still needs to:

1. Verify that if  user is paid or not rather than taking the value straight from the import file.
2. Refresh dexConnections with newly imported account. (Currently you have to restart the client.)

kevinstl:
Account is now verified prior to inserting into db.

The user has to log out and log back in to see their account id displayed.

Still need unit test account import.

client/webserver/api.go
```diff
@@ -189,6 +189,39 @@ func (s *WebServer) apiTrade(w http.ResponseWriter, r *http.Request) {
...
+	form := new(accountExportAuthForm)
+	if !readPost(w, r, form) {
+		return
+	}
+	r.Close = true
+	accountResponse, err := s.core.AccountExport(form.Pass, form.Host)
+	if err != nil {
+		s.writeAPIError(w, "error retrieving keys: %v", err)
+		return
+	}
```
buck54321, line 203: A couple of things need tweaking here. 

1. The `AccountExport` method returns a `core.Account`, not `core.AccountResponse`, also meaning that `core.AccountResponse` is unused. 
2. **But**, you do want to have separate types account and response. `core.AccountResponse` should be a **webserver** type though, like all of the other responses.
3. After you resolve **2**, remove the `OK` field from the `core.Account`. That field is only for the API response.

client/webserver/site/src/js/settings.js
```diff
@@ -46,6 +52,23 @@ export default class SettingsPage extends BasePage {
     Doc.bind(page.addCert, 'click', () => this.page.certFile.click())
     forms.bind(page.dexAddrForm, page.submitDEXAddr, () => { this.verifyDEX() })
     forms.bind(page.confirmRegForm, page.submitConfirm, () => { this.registerDEX() })
+    forms.bind(page.authorizeAccountExportForm, page.authorizeExportAccountConfirm, () => { this.exportAccount() })
+
+    const exchangesDiv = page.exchanges
+    var exportAccountKeyButton
+    // eslint-disable-next-line no-unused-vars
+    for (const [host, xc] of Object.entries(app.user.exchanges)) {
```
buck54321, line 60: `for (const host of Object.keys(app.user.exchanges)) {`

Don't disable the linter, please.

buck54321 (commented):
@kevinstl this is looking pretty good. Anything else before removing draft status?

client/webserver/site/src/js/settings.js
```diff
@@ -46,6 +53,22 @@ export default class SettingsPage extends BasePage {
     Doc.bind(page.addCert, 'click', () => this.page.certFile.click())
     forms.bind(page.dexAddrForm, page.submitDEXAddr, () => { this.verifyDEX() })
     forms.bind(page.confirmRegForm, page.submitConfirm, () => { this.registerDEX() })
+    forms.bind(page.authorizeAccountExportForm, page.authorizeExportAccountConfirm, () => { this.exportAccount() })
+
+    const exchangesDiv = page.exchanges
+    var exportAccountButton
+    for (const host of Object.keys(app.user.exchanges)) {
+      exportAccountButton = Doc.tmplElement(exchangesDiv, 'exportAccount-' + host)
```
chappjc, line 61: Seems like the declaration outside the loop is not necessary.  Can be `const exportAccountButton = ...` inside the loop?

client/webserver/site/src/js/settings.js
```diff
@@ -61,11 +84,106 @@ export default class SettingsPage extends BasePage {
...
+    const page = this.page
+    const pw = page.exportAccountAppPass.value
+    const host = page.exportAccountHost.textContent
+    page.exportAccountAppPass.value = ''
+    const req = {
+      pw: pw,
+      host: host
+    }
+    const loaded = app.loading(this.body)
+    var res = await postJSON('/api/exportaccount', req)
```
chappjc, line 105: Should be fine with `const` here.

JoeGruffins (commented):
Exporting working great.
Am unable to import. clicking "Authorize Import" doesn't seem to do anything.
I think this add button should look like when you load a wallet config.
button

It's not very obvious that one can go to settings to do this when creating a new account. I guess the timing would be after connecting to a dex but before paying the fee, or on the fee screen, a hint that they can restore from `.pepk` as well.

As a side note, this will probably also be needed for the rpcserver, but that does not need to be done in this pr.

kevinstl:
> Am unable to import. clicking "Authorize Import" doesn't seem to do anything.


I've tested import on chrome and firefox & it is working for me so I'm not sure where to go from here.

client/core/core.go
```diff
@@ -3215,6 +3273,34 @@ func (c *Core) initialize() {
...
+	}
+	dc, err := c.connectDEX(acct)
+	if err != nil {
+		c.log.Errorf("error connecting to DEX %s: %v", acct.Host, err)
+		return false
+	}
+	c.log.Debugf("connectDEX for %s completed, checking account...", acct.Host)
+	if !acct.Paid {
+		if len(acct.FeeCoin) == 0 {
+			// Register should have set this when creating the account
```
chappjc, line 3290: If not paid, but there is a fee coin, we'd need to do something like `checkUnpaidFees` or `reFee` (not sure), otherwise dexc would need a restart to finish this up.

client/core/core.go
```diff
@@ -3152,30 +3231,9 @@ func (c *Core) initialize() {
...
 			defer wg.Done()
-			host, err := addrHost(acct.Host)
-			if err != nil {
-				c.log.Errorf("skipping loading of %s due to address parse error: %v", acct.Host, err)
-				return
-			}
-			dc, err := c.connectDEX(acct)
-			if err != nil {
-				c.log.Errorf("error connecting to DEX %s: %v", acct.Host, err)
+			if !c.verifyAccount(acct) {
```
chappjc, line 3234: I like the refactoring here, but it does raid the question of what we should do upon importing an account with active orders.  Before authorizing with the imported dex, we'd want to run all the other code in initialize like loading wallets and db trades to that when authorizing with dex, the orders are all loaded up.

client/core/types.go
```diff
@@ -695,6 +695,18 @@ type OrderFilter struct {
 	Statuses []order.OrderStatus `json:"statuses"`
 }
 
+// Account holds data returned from AccountExport.
+type Account struct {
+	Host      string `json:"host"`
+	AccountID string `json:"accountID"`
+	PrivKey   string `json:"privKey"`
+	PubKey    string `json:"pubKey"`
+	EncKey    string `json:"encKey"`
```
chappjc, line 704: I'm not certain we want to include `EncKey`.  For a given app password, `EncKey` is derived from `PrivKey`.

client/core/core.go
```diff
@@ -2632,6 +2633,84 @@ func (c *Core) Trade(pw []byte, form *TradeForm) (*Order, error) {
...
+	}
+	accountInfo.DEXPubKey, err = secp256k1.ParsePubKey(pubKey)
+	if err != nil {
+		return codedError(parseKeyErr, err)
+	}
+	encKey, err := hex.DecodeString(account.EncKey)
+	if err != nil {
+		return codedError(decodeErr, err)
+	}
+	accountInfo.EncKey = encKey
```
chappjc, line 2691: I'm not certain we want to include `EncKey`.  For a given app password and encryption key parameters stored in the DB, `EncKey` is **derived from `PrivKey`**.  Basically on import we'd be doing similar to `setupEncryption`, but with the crypter from `(*Core).encryptionKey` we'd compute an `encKey` from the imported `privKey` rather than a newly generated private key.

chappjc (changes requested):
I don't think we want to import/export `EncKey` since that depends on the current app password and encryption key parameters, and the `PrivKey` is what's really needed.

Also, I think we can have part 2 of this work be dealing with the startup of imported accounts.  Right now a restart of dexc is needed to get it moving, both for active orders and for unpaid accounts.

client/core/core.go
chappjc: Ah, I see you've named that as next step in your comment. https://github.com/decred/dcrdex/pull/885#issuecomment-749339222 Sounds good.

client/core/core.go
buck54321: Yes. The `EncKey` does encode the key parameters, but it is derived from the `PrivKey` and including both is pointless. Drop `EncKey`. If users want the backup to be encrypted, they'll have to do it themselves.

chappjc: For my own understanding, take the following in client/core/types.go, where the encKey is decrypted to get privKey:

```go
func (a *dexAccount) unlock(crypter encrypt.Crypter) error {
	keyB, err := crypter.Decrypt(a.encKey)
	...
	privKey := secp256k1.PrivKeyFromBytes(keyB)
```

and in client/core/core.go where the app password is used to create the (de)crypter when combined with keyParams:

```go
func (c *Core) encryptionKey(pw []byte) (encrypt.Crypter, error) {
	keyParams, err := c.db.Get(keyParamsKey)
	if err != nil {
		return nil, fmt.Errorf("key retrieval error: %w", err)
	}
	crypter, err := c.reCrypter(pw, keyParams)
```

The `keyParams`, which includes the salt and other encryption parameters, is combined with the app password to create the `crypter`.  The `crypter` can decrypt `encKey`.  So to decrypt an import `encKey`, you'd need to make a crypter with pw (from input) + keyParams (usually from DB).

I maybe misunderstanding, but I don't believe `EncKey` includes the key parameters, although they are generally stored in the same DB file as `EncKey` so it appears to usually be a non-issue.

buck54321: Oh right. The `EncKey` doesn't encode the parameters, the encrypted `Crypter` does.

client/core/core.go
```diff
@@ -2632,6 +2633,84 @@ func (c *Core) Trade(pw []byte, form *TradeForm) (*Order, error) {
...
+	accountInfo.Cert = cert
+	feeCoin, err := hex.DecodeString(account.FeeCoin)
+	if err != nil {
+		return codedError(decodeErr, err)
+	}
+	accountInfo.FeeCoin = feeCoin
+	if !c.verifyAccount(&accountInfo) {
+		return newError(accountVerificationErr, "Account not verified for host: "+account.Host, err)
+	}
+	accountInfo.Paid = c.conns[account.Host].acct.isPaid
```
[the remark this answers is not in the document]
kevinstl: The c.verifyAccount() method populates c.conns.

I would think that you want IsPaid to come from the value known from the dex. Otherwise the user could just change the value in the file. I'm not sure of the implications of them faking that value into the db.

Is your point that the user would have no motivation to fake data since the dex isPaid value is always verified? Or in other words we don't have to trust any client data, it exists purely for convenience?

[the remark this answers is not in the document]
chappjc: Ah, no I just neglected verifyAccount when I made the comment.

client/core/core.go
chappjc: Also good for a part 2

client/core/core.go
kevinstl: Now that I removed encKey I'm getting "zero-length EncKey not allowed' from: 

```go
func (db *BoltDB) CreateAccount(ai *dexdb.AccountInfo) error {
	if ai.Host == "" {
		return fmt.Errorf("empty host not allowed")
	}
	if ai.DEXPubKey == nil {
		return fmt.Errorf("nil DEXPubKey not allowed")
	}
	if len(ai.EncKey) == 0 {
		return fmt.Errorf("zero-length EncKey not allowed")
	}
```

Should I follow the steps above to put encKey in or remove the validation from CreateAccount?

chappjc: Take the crypter that's currently discarded from `_, err := c.encryptionKey(pw)` and compute `encPW` from `pw` similar to how it's done in `setupEncryption`.

client/core/types.go
```diff
@@ -695,6 +695,18 @@ type OrderFilter struct {
 	Statuses []order.OrderStatus `json:"statuses"`
 }
 
+// Account holds data returned from AccountExport.
+type Account struct {
+	Host      string `json:"host"`
+	AccountID string `json:"accountID"`
+	PrivKey   string `json:"privKey"`
+	PubKey    string `json:"pubKey"`
```
chappjc, line 703: Let's rename `PubKey` to `DEXPubKey`, otherwise it looks like this is the pubkey that goes with the user's `PrivKey`, but it's the DEX server's public key.

client/core/core.go
chappjc: I'm noticing that both `AccountID` and `PrivKey` aren't used.  You're going to use `PrivKey` as I've described above, but the `AccountID` seems to be purely informational in the json file.  That's OK, but it's worth noting in the `AccountImport` that the `AccountID` is not validated.  To do so would involving doing similar to what `(*dexAccount).unlock` does, namely `account.AccountID == account.NewID(privKey.PubKey().SerializeCompressed())`.  Hmm, saying that, I think the `account` variable name declared in `AccountImport(pw []byte, account Account)` would be better as `acct` to avoid a collision with the `account` package.

client/core/core.go
```diff
@@ -2675,6 +2676,100 @@ func (c *Core) Trade(pw []byte, form *TradeForm) (*Order, error) {
...
+	if err != nil {
+		return codedError(decodeErr, err)
+	}
+
+	accountInfo.FeeCoin, err = hex.DecodeString(acct.FeeCoin)
+	if err != nil {
+		return codedError(decodeErr, err)
+	}
+
+	accountInfo.Paid = acct.IsPaid
```
chappjc, line 2750: Well, I had not thought about this, but the `Paid` part is not encoded in the same DB bucket with the `AccountInfo`
See `(db *BoltDB).AccountPaid` and `feeProofKey`.  I think we're going to need to backup the `AccountProof` too.

```go
type AccountProof struct {
	Host  string
	Stamp uint64
	Sig   []byte
}
```

The new account struct already has a `Host string`, so maybe the exported struct with json tags can have an `FeeProofSig` and `FeeProofStamp`.  This means on import using the `AccountPaid` db method to store the loaded fee proof, and for export creating a new BoltDB method to load the AccountProof since we don't have one.

chappjc: Maybe something like the following to load the `AccountProof` for export:

```go
func (db *BoltDB) AccountProof(url string) (*dexdb.AccountProof, error) {
	var acctProof *dexdb.AccountProof
	acctKey := []byte(url)
	return acctProof, db.acctsView(func(accts *bbolt.Bucket) error {
		acct := accts.Bucket(acctKey)
		if acct == nil {
			return fmt.Errorf("account not found for %s", url)
		}
		var err error
		acctProof, err = dexdb.DecodeAccountProof(getCopy(acct, feeProofKey))
		if err != nil {
			return err
		}
		return nil
	})
}
```

For import there's the existing `(db *BoltDB) AccountPaid(proof *dexdb.AccountProof)`

client/core/account.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		return codedError(decodeErr, err)
+	}
+	accountInfo.EncKey, err = crypter.Encrypt(privKey)
+	if err != nil {
+		return codedError(encryptionErr, err)
+	}
+	if !c.verifyAccount(&accountInfo) {
+		return newError(accountVerificationErr, "Account not verified for host: %s err: %v", acct.Host, err)
+	}
```
chappjc, line 107: Please make a comment that `verifyAccount` makes a connection to the DEX to verify it (I'm actually not sure that's desirable though), but a restart is required to use the DEX.  I am actually thinking we might want to comment out the call to `verifyAccount` since it doesn't really do anything, and adding the `dexConnection` to the `conns` map might actually be problematic since we're not authenticating the connection.

In part 2, we'll have to factor out of `resolveActiveTrades` the `loadDBTrades` and `resumeTrades` bits and call it in this function to make it work without a restart.

client/core/account.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		return codedError(decodeErr, err)
+	}
+
+	accountInfo.FeeCoin, err = hex.DecodeString(acct.FeeCoin)
+	if err != nil {
+		return codedError(decodeErr, err)
+	}
+
+	accountInfo.Paid = acct.IsPaid
```
chappjc, line 95: Let's remove the `IsPaid` field of the new `Account` struct and here do `accountInfo.Paid = acct.FeeProofSig != "" && acct.FeeProofStamp != 0`.

client/core/account.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	err = c.db.CreateAccount(&accountInfo)
+	if err != nil {
+		return codedError(dbErr, err)
+	}
+
+	var accountProof db.AccountProof
+	if acct.FeeProofSig != "" && acct.FeeProofStamp != 0 {
+		accountProof.Sig, err = hex.DecodeString(acct.FeeProofSig)
```
chappjc, line 116:
```go
	if accountInfo.Paid {
		var accountProof db.AccountProof
		accountProof.Sig, err = hex.DecodeString(acct.FeeProofSig)
```

chappjc (changes requested):
Just about there. Thanks for all the thorough tests! The `IsPaid` field should be removed though since it's inferred by the `FeeProofStamp` and `FeeProofSig` fields.

client/core/account.go
```diff
@@ -0,0 +1,132 @@
...
+	c.connMtx.RLock()
+	dc, found := c.conns[host]
+	c.connMtx.RUnlock()
+	if !found {
+		return nil, newError(unknownDEXErr, "DEX: %s", host)
+	}
+
+	dc.acct.keyMtx.RLock()
+	accountId := dc.acct.id.String()
+	privKey := hex.EncodeToString(dc.acct.privKey.Serialize())
```
chappjc, line 31: This panics if account is not unlocked.  Either return with an error if the account is locked or unlock it for them

```go
	if err = dc.acct.unlock(crypter); err != nil {
		return nil, codedError(acctKeyErr, err)
	}
```

JoeGruffins:
I am unable to load a second account.
```
2021-01-20 16:13:36.093 [ERR] WEB: error importing account: failed to create account bucket
```
I am able to load as a new account, but there are still no visual cues as to this being an option. It is however working well when the account bucket does not yet exist, and I know where to look.

client/webserver/site/src/js/settings.js
```diff
@@ -68,11 +91,106 @@ export default class SettingsPage extends BasePage {
...
+  // importAccount imports the account
+  async importAccount () {
+    const page = this.page
+    const pw = page.importAccountAppPass.value
+    page.importAccountAppPass.value = ''
+    let accountString = ''
+    if (page.accountFile.value) {
+      accountString = await page.accountFile.files[0].text()
+    }
+    const account = JSON.parse(accountString)
```
chappjc, line 164: For followup, there needs to be handling for a parse error. cc @buck54321

client/webserver/api.go
```diff
@@ -189,6 +189,45 @@ func (s *WebServer) apiTrade(w http.ResponseWriter, r *http.Request) {
...
+	w.Header().Set("Connection", "close")
+	res := &struct {
+		OK      bool          `json:"ok"`
+		Account *core.Account `json:"account"`
+	}{
+		OK:      true,
+		Account: account,
+	}
+	writeJSON(w, res, s.indent)
```
chappjc, line 212: Also for followup, we should probably always indent this as it's a download that the user should be able to visually parse.

chappjc (approved):

chappjc:
> I am unable to load a second account.
> 
> ```
> 2021-01-20 16:13:36.093 [ERR] WEB: error importing account: failed to create account bucket
> ```
> 
> I am able to load as a new account, but there are still no visual cues as to this being an option. It is however working well when the account bucket does not yet exist, and I know where to look.

Can you clarify what you mean by "I am able to load as a new account, but there are still no visual cues as to this being an option."  I can only import if the host does not already exist as an account.

client/webserver/api.go
kevinstl: It looks like the indent argument is ultimately ignored:

```
// writeJSON writes marshals the provided interface and writes the bytes to the
// ResponseWriter with the specified response code.
func writeJSONWithStatus(w http.ResponseWriter, thing interface{}, code int, indent bool) {
	w.Header().Set("Content-Type", "application/json; charset=utf-8")
	b, err := json.Marshal(thing)
	if err != nil {
		w.WriteHeader(http.StatusInternalServerError)
		log.Errorf("JSON encode error: %v", err)
		return
	}
	w.WriteHeader(code)
	_, err = w.Write(append(b, byte('\n')))
	if err != nil {
		log.Errorf("Write error: %v", err)
	}
}
```

I was able to indent successfully from a javascript change:

```
a.setAttribute('href', 'data:text/json,' + JSON.stringify(accountForExport, null, 2))
```

JoeGruffins:
@chappjc 

> Can you clarify what you mean by "I am able to load as a new account, but there are still no visual cues as to this being an option."

I mean, I am able to load before creating an account.  I didn't touch the json.

Sorry I have not been following this issue closely, but I would think the greatest use for exporting/importing is restoring your account if you lost your db for some reason. So, the user should be able to see the option before paying the fee and making a new account if they have one to import.

Is there a one account per dex per dexc rule? If so then it is right to error, but the error could be better. "i.e. you already have an account with this dex"

chappjc:
> Is there a one account per dex per dexc rule? If so then it is right to error, but the error could be better. "i.e. you already have an account with this dex"

Ah, yeah, that's the rule.  We'll revise the error message displayed in a follow-up.

JoeGruffins (approved):
Thanks for testing the negative paths in the tests. The method you use is a bit unorthodox however. Just for example take this: https://github.com/decred/dcrdex/blob/ca9882d36463c94375c200ca09cf1a44740f6d1a/client/rpcserver/rpcserver_test.go#L349-L435

There is a test setup and tear down. The tests themselves run off of an array of test structs where you specify what changes and what the output should be. This allows testing vectors to be easily added or removed, and is also fairly easy to reason about and is standard, I believe.

buck54321 (approved):

kevinstl:
> tests

If the consensus is that unit tests should be written the way you've specified I will adhere to it.

I have a few reasons for structuring them as I have:

1. It is very easy to see the reason for the failure & go directly to the test that is failing.
2. It is also easier to run the individual test through a debugger with break points, rather that having to iterate through each case until you get to the one that you care about.
3. There is no chance of tests depending on or interfering with each other as they are completely standalone functions.

kevinstl:
> there are still no visual cues as to this being an option

Perhaps this can be addressed with a link to import along the register path?

chappjc:
With closely related tests where it's going through the different inputs to a function like AccountExport, @JoeGruffins is generally right that a slice of test cases in a single function is better than a test function for each case. It also makes maintaining the test easier.  But I am not totally hung up on the tests.

kevinstl:
> With closely related tests where it's going through the different inputs to a function like AccountExport, @JoeGruffins is generally right that a slice of test cases in a single function is better than a test function for each case. It also makes maintaining the test easier. But I am not totally hung up on the tests.

@JoeGruffins 

I'll stick with the pattern in rpcserver_test.go going forward. Thanks for the feedback.

Issue #884, opened 2020-12-11
cannot change wallet settings with active swaps


chappjc:
I've always suggested users not change wallet config with active swaps, but I've just tested it on simnet and the dcr wallet can't get counterparty swap confs

```
2020-12-11 14:50:13.375 [ERR] CORE: Failed to get confirmations of the counter-party's swap cf37027a1d903645f44c958c4dde072603ad1598865279c29845e04dbe06d9d4:0 (dcr) for match 30c87404f5aa47aca004a1781879b9c8468d137e9af74f76781a803f8d5d7955, order be1dabef8d48842fd557437b86ce3dd7a558ce774b8858f82777b661a47d7bcd
```

I believe it's the block cache getting toasted or the Coin losing it's backend/node when a new ExchangeWallet is created by ReconfigureWallet, causing the Confirmations method on the coin to fail. Restarting dexc is the only way out of it.

EDIT: I believe the `matchTracker.counterSwap` field (an `asset.AuditInfo`) becomes invalid as it contains a pointer to the old `ExchangeWallet.node` (the old rpc client), which is no longer connected.

We should get this fixed ASAP for 0.1.4.

repro:

```
create match allowing it to get to makerswap cast and for the audit to finish
2020-12-11 14:47:30.021 [DBG] CORE: Registering match 30c87404f5aa47aca004a1781879b9c8468d137e9af74f76781a803f8d5d7955 for order 2cb1d6a03d1e65b896cb30c17f9d5e19f0a62c3228e2bf9f42cd7f86605d2222 (Maker) in status NewlyMatched
2020-12-11 14:47:30.028 [DBG] CORE: Registering match 30c87404f5aa47aca004a1781879b9c8468d137e9af74f76781a803f8d5d7955 for order be1dabef8d48842fd557437b86ce3dd7a558ce774b8858f82777b661a47d7bcd (Taker) in status NewlyMatched
2020-12-11 14:47:30.030 [INF] CORE: Starting negotiation for match 30c87404f5aa47aca004a1781879b9c8468d137e9af74f76781a803f8d5d7955 for order be1dabef8d48842fd557437b86ce3dd7a558ce774b8858f82777b661a47d7bcd with swap fee rate = 1, quantity = 200000000
[...]

Issue #883, opened 2020-12-11
canceling a partially filled order causes revokes


raedah:
Canceling a partially filled order removes the remaining order from the order book and as well causes any in-progress swaps to go into revoke. Expected behavior when canceling a partially filled order is to remove the remaining order form the order book and to complete the outstanding swaps that have already matched. Forcing a revoke on swap would be a separate action that could be accessed on the order details if needed. Also note that order details doesn't currently display the status of any ongoing revoke. Match revoke status is only shown on the notifications.

chappjc:
Cancelling an order with active swaps does not trigger the match to be revoked.  I'll need more information about what you observed, but matches cannot be interrupted just by cancelling the order.  I suspect there was an unrelated match revoke.

There is no way for a dex user to revoke a swap aside from disconnecting something.  We don't intend to support a way to back out of a swap.

raedah:
perhaps this

[ERR] CORE: cannot redeem order ..., match .., because dcr wallet  is not unlocked

and perhaps the wallet locking is due to not having open orders

chappjc:
That'll do it.  Wallet won't get locked when orders are canceled though.  I do believe it was locked unintentionally somehow.  Any chance you did something with dcrctl?

I would advise getting on the latest dexc release though because [0.1.2 included an automatic wallet (re)unlocking](https://github.com/decred/dcrdex/commit/de409134c37270145dc7094e89d6ef9d8e2d1f74) to avoid revokes that would otherwise occur in such unexpectedly locked wallet situations.  However, you will still need to be logged into dexc for it to be able to unlock wallets automatically like this.

raedah:
Already on 0.1.3. Strange part about it is dcrctl walletinfo is showing the wallet status as unlocked true. dexc output is still showing  not unlocked error occurring.

chappjc:
Hmm, it sounds like it's too late, but you can restart dexc and log back in asap, and it will resume.

chappjc:
@buck54321 This is the locallyUnlocked state error that @raedah cited, while the wallet itself appears unlocked according to dcrctl walletinfo.  What circumstances do you think this can get out of agreement?  I feel like we've seen some similar disagreement as a result of killing dexc, but it's not clear exactly how that could happen.

raedah:
restarting dexc results in these msgs

[DBG] CORE: AuditContract error for match ... status TakerSwapCast, refunded = false, revoked = true: coin not found
[...]

PR #887, opened 2020-12-13
remove Confirmations method from client/asset.Coin

buck54321:
Since the `Confirmations` method on `asset.Coin` required the `{btc,dcr}.output` to retain a reference to the `ExchangeWallet`, after `Reconfiguring` the wallet in `Core`, anyh old `Coin`s would error. This would happen for Decred even if no settings were changed (error "the client has been shutdown"). 

This PR removes that method from output, and adds a `gettxout` check to the `(*ExchangeWallet).Confirmations` method. 

Resolves https://github.com/decred/dcrdex/issues/884


client/asset/interface.go
```diff
@@ -156,8 +156,8 @@ type Wallet interface {
 	// the registration fee, and the fee rate is taken from the DEX configuration.
 	PayFee(address string, feeAmt uint64) (Coin, error)
 	// Confirmations gets the number of confirmations for the specified coin ID.
-	// The ID need not represent an unspent coin, but coin IDs unknown to this
-	// wallet may return an error.
+	// If the coin is not unspent, and is not owned by this wallet,
+	// Confirmations may return an error.
 	Confirmations(ctx context.Context, id dex.Bytes) (uint32, error)
```
buck54321, line 161: I considered adding an `unspent bool` return value to the method too, but after examining the way the method is used in `Core`, I don't think it's necessary.

chappjc: Probably OK, but the recovery paths should be tested.   I'm also wondering if the isSwappable/isRedeemable checks might behave differently under certain circumstances if spent contracts are no longer a coin not found error.

Unfortunately the harness is still busted in a myriad of ways: https://github.com/decred/dcrdex/pull/799  We could at least change that test to use an encrypted btc wallet so the hack to block actions would work again.

chappjc: Also, while I there was a match recovery issue we need to look into in https://github.com/decred/dcrdex/issues/883#issuecomment-743417815, whatever happened there would have changed too because the audit would not have said "coin not found" (it was auditing a spent contract for some reason).  That issue still needs investigation despite being closed, but I'm noting it here because such an audit might have passed despite the contract being spent.

chappjc: Thoughts about use of `gettransaction` as fallback after `gettxout` in `asset.Wallet.Confirmations`: `getransaction` must be wallet transactions, which counterparty swaps shouldn't be, but asset backends might behave differently, or the contract could be made by counterparty to include an extra output that wallet recognizes thus changing the result.  The uncertainty of how this `Confirmations` method will behave for spent counterparty contracts is a concern.

---

`isSwappable` in `MakerSwapCast`
- As maker, will now notify confirmations after redeem and refund, until retire.  Still returns false though.
- As taker, could return `ready=true` even after the coin (maker's contract) is spent *depending on wallet capability* (getransaction), which in `MakerSwapCast` indicates maker refunded already.  This probably means revoked, which should make it return false before the confs check, but it's not guaranteed to be marked as revoked (e.g. server down) although not being revoked after maker contract expire is unlikely. Aside: taker swapping this late in the game would require taker making a contract that is already expired -- I think we've actually seen that happen, and we should be checking `lockTime` is not in the past in `swapMatchGroup`.

`isRedeemable` in `TakerSwapCast`
- As taker, will notify confirmations after redeem and refund, until retire.  Still returns false though.
- As maker, could say ready=true even after spent *depending on wallet capability* (getransaction), which in `TakerSwapCast` indicates taker refunded, which implies revoked, but `isRedeemable` **does not check `IsRevoked`** for sake of auto-redeem. (**Not good**) Thus, maker might try to redeem the refunded taker contract instead of going to refund their own by switching on `isRefundable`.  Unable to refund scenario.

`shouldBeginFindRedemption` as taker already used `wallet.Confirmations` to check own contract confirms. It will now hit `gettxout` first and will fall back to `gettransaction` when spent, so logging when spent might actually be useful since that's when you expect a redeem will be found right away.

The `t.notify` in `counterPartyConfirms` should be in the `if match.counterConfirms != int64(have)` condition.  Or is it necessary to (re)send this value to frontend on ticks and not just changes to the value?

buck54321:
> isSwappable in MakerSwapCast: As taker, could return ready=true even after the coin (maker's contract) is spent

> isRedeemable in TakerSwapCast: As maker, could say ready=true even after spent

Don't both of these scenarios imply violation of the contract lock time, i.e. early refund? If the contracts are spent before the lock time, we redeemed them ourselves, and the match status will no longer be in `MakerSwapCast`/`TakerSwapCast`. 

Regardless, while I don't believe that the unspent status is critical to secure and effective DEX client operation, I acknowledge that the lack of unspent status presents issues in troubleshooting and is just generally hard to swallow. I'll add the return value.

chappjc:
They could just be refunded after the lock time expired, and before they reached swapConf.  Which is why I mentioned that they _should_ be revoked in these cases, but we're not handling that well on either client or server side.

We have definitely seen taker contracts broadcast with already expired contracts, which is a separate issue.

chappjc:
All simnet trade tests passed with the fixed harness from https://github.com/decred/dcrdex/pull/799.  Will give this a final pass, but looking good.

client/core/trade.go
```diff
@@ -574,19 +574,17 @@ func (t *trackedTrade) processCancelMatch(msgMatch *msgjson.Match) error {
...
 	needed = t.wallets.toAsset.SwapConf
 
 	// Check the confirmations on the counter-party's swap.
 	coin := match.counterSwap.Coin()
-	var err error
-	ctx, cancel := context.WithTimeout(context.Background(), confCheckTimeout)
-	defer cancel()
-	have, err = coin.Confirmations(ctx)
+
+	have, spent, err := t.wallets.toWallet.Confirmations(ctx, coin.ID())
```
chappjc, line 584: Just a note, no need to change.  I'm really not wild about mixing declaration and assignment with the short variable declaration operator (`:=`), although since `have` and `spent` exist they are assignments, not new variables shadowing the named outputs of `counterPartyConfirms`.  I often go out of my way to declare stuff like `var err error` to avoid the need to analyze the surrounding code to understand what's going on with declaration vs assignment.  Maybe that's just me though.

client/asset/btc/btc.go
```diff
@@ -1945,21 +1928,28 @@ func (btc *ExchangeWallet) send(address string, val uint64, feeRate uint64, subt
...
+	}
+	// Check wallet transactions.
 	tx, err := btc.wallet.GetTransaction(txHash.String())
 	if err != nil {
 		if isTxNotFoundErr(err) {
-			return 0, asset.CoinNotFoundError
+			return 0, true, asset.CoinNotFoundError
 		}
-		return 0, err
+		return 0, true, err
```
chappjc, line 1950: If gettransaction can't find it, I don't think we should return `spent=true`

client/core/trade.go
```diff
@@ -574,19 +574,17 @@ func (t *trackedTrade) processCancelMatch(msgMatch *msgjson.Match) error {
...
 	if err != nil {
-		t.dc.log.Errorf("Failed to get confirmations of the counter-party's swap %s (%s) for match %v, order %v",
-			coin, t.wallets.toAsset.Symbol, match.id, t.UID())
+		t.dc.log.Errorf("Failed to get confirmations of the counter-party's swap %s (%s) for match %v, order %v: %v",
+			coin, t.wallets.toAsset.Symbol, match.id, t.UID(), err)
 		have = 0 // should already be
```
chappjc, line 588: Considering that `counterPartyConfirms` returns `spent=true` when `err != nil`, this is going to tell the caller that the txn (contract) is spent when really it just wasn't found.  Could be the case due to network latency or whatever else.

client/core/wallet.go
```diff
@@ -164,3 +164,9 @@ func (w *xcWallet) Disconnect() {
 	w.hookedUp = false
 	w.mtx.Unlock()
 }
+
+func (w *xcWallet) Confirmations(ctx context.Context, coinID []byte) (uint32, bool, error) {
```
chappjc, line 168: Might as well doc if you're pushing another commit.

client/asset/btc/btc.go
buck54321: With an non-nil error, it the caller should probably be ignoring the value anyway, so I'll defer to the zero value in that case. We can't say that it's a spent output anyway, though we can say that it's not an unspent output.

JoeGruffins (approved):
Tested with the new harness and manually and looks fine.

chappjc:
Gonna merge this today.  I just need to go through a wallet reconfig at each stage.

chappjc (approved):

Issue #879, opened 2020-12-09
BTC deposit address not in dex wallet


peterzen:
I sent a BTC deposit to the address shown in the GUI, but it didn't end up in the configured `dex` wallet.

```
# grep Deposit mainnet/logs/dexc.log
./mainnet/logs/dexc.log:2020-12-09 16:02:06.376 [INF] CORE: Loaded btc wallet configuration. Deposit address = bc1deadbeef....
```

Verifying whether the deposit address from `dexc` belongs to the `dex` wallet:
```
# bitcoin-cli -conf=/home/bitcoin/.bitcoin/bitcoin.conf getaddressinfo bc1deadbeef....
{
  "address": "bc1deadbeef....",
  "ismine": false,
  "solvable": false,
  "iswatchonly": false,
  "isscript": false,
  "iswitness": true,
  "ischange": false,
}
```

Confirming that it's the `dex` wallet that's open:

```
# bitcoin-cli -conf=/home/bitcoin/.bitcoin/bitcoin.conf getwalletinfo
{
  "walletname": "dex",
  "walletversion": 169900,
  "balance": 0.00000000,
  "unconfirmed_balance": 0.00000000,
  "immature_balance": 0.00000000,
  "txcount": 12,
  "keypoololdest": 1606075542,
  "keypoolsize": 10000,
  "hdseedid": "xxxxxxx",
  "keypoolsize_hd_internal": 10000,
  "unlocked_until": 0,
  "paytxfee": 0.00000000,
  "private_keys_enabled": true,
  "avoid_reuse": false,
  "scanning": false
}
```

Any clues what might be the issue here?  The wallet was working correctly, a BTC withdrawal worked correctly yesterday. `bitcoind` is caught up to the tip as evidenced in `dexc.log`.

chappjc:
Are there other walletnames in `listwallets` and `listwalletdir`, and are you certain that the `"dex"` walletname is configured in the dexc GUI?

peterzen:
I verified the above and reconstructed the timeline:

1. New bitcoind instance & default wallet created, initial sync completed. 
`wallet=dex` not present in `bitcoin.conf` at this point.
2. Added BTC wallet in `dcrdex`, configured rpc, verified the wallet connection
3. `bitcoin-cli createwallet dex`, `bitcoin-cli -rpcwallet=dex encryptwallet`
4. Changed BTC wallet in `dcrdex` to `dex` wallet, added `wallet=dex` to `bitcoin.conf`
[...]

PR #893, opened 2020-12-17
prevent wallet Lock from blocking shutdown

chappjc:
The first commit, https://github.com/decred/dcrdex/pull/893/commits/3910136e938918698764bf219bd1d8b6c07b6e04, prevents hanging on shutdown with a disconnected wallet.  Shutdown would hang because the subsystem context was not canceled by `PromptShutdown` until *after* wallets were locked. This modifies both dcr and btc `(*ExchangeWallet).Lock` methods to check node disconnected status, and modifies the dcr method with a 5 second timeout since it has context-enabled RPC methods unlike btc.  This commit also adds a basic dcr `OnClientConnected` notification handler to log when the client (re)connects.

The second commit, https://github.com/decred/dcrdex/pull/893/commits/905b1be23e460150b2b838fb0effa7c1124883f3, consistently uses the `(*Core).connectWallet` method instead of calling `(*ExchangeWallet).Connect` directly.  This ensures that the sync status loop is always started regardless of the path used to connect the wallet.  Previously only `(*Core).connectAndUnlock` was calling `connectWallet`, although `(*ExchangeWallet).Connect` at least checked `SyncStatus` once.

Commit https://github.com/decred/dcrdex/pull/893/commits/0441ddaed67c46d792b68daff07994c194fde0d9 has the `connectWallet` loop on SyncStatus continue on errors.  It will still break out if the wallet is disconnected.  This helps because the underlying rpcs (e.g. getblockchaininfo) can return unexpected errors even though it is connected.  We may also consider modifying `(*ExchangeWallet).Connect` so that it does not error out on `SyncStatus` as long as it connects successfully because in this case there is no connect retry loop.  Even more generally, we should change `Core`'s wallet connection logic so that it continually tries to make a connection to the wallet when it fails, since while a wallet does try to reconnect once initially connected, we do not have a way to continually try to establish a new connection (e.g. on login).

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -138,6 +138,7 @@ type rpcClient interface {
 	GetRawMempool() ([]*chainhash.Hash, error)
 	GetRawTransactionVerbose(txHash *chainhash.Hash) (*btcjson.TxRawResult, error)
 	RawRequest(method string, params []json.RawMessage) (json.RawMessage, error)
+	Disconnected() bool
 }
 
 // BTCCloneCFG holds clone specific parameters.
@@ -345,7 +346,6 @@ type ExchangeWallet struct {
 	// 64-bit atomic variables first. See
 	// https://golang.org/pkg/sync/atomic/#pkg-note-BUG
 	tipAtConnect      int64
-	client            *rpcclient.Client
 	node              rpcClient
 	wallet            *walletClient
 	walletInfo        *asset.WalletInfo
@@ -451,10 +451,7 @@ func BTCCloneWallet(cfg *BTCCloneCFG) (*ExchangeWallet, error) {
 		return nil, fmt.Errorf("error creating BTC RPC client: %w", err)
 	}
 
-	btc := newWallet(cfg, btcCfg, client)
-	btc.client = client
-
-	return btc, nil
+	return newWallet(cfg, btcCfg, client), nil
 }
 
 // newWallet creates the ExchangeWallet and starts the block monitor.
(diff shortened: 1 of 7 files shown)


JoeGruffins (approved):
It would be cool if the wallet status changed from "ready", but this does solve the issue for me.

client/core/core.go
```diff
@@ -1032,20 +1032,21 @@ func (c *Core) encryptionKey(pw []byte) (encrypt.Crypter, error) {
...
+		return nil, newError(missingWalletErr, "no configured wallet found %s (%d)", unbip(assetID), assetID)
 	}
 	if !wallet.connected() {
-		c.log.Infof("connecting wallet for %s", unbip(assetID))
-		err := wallet.Connect(c.ctx)
+		c.log.Infof("Connecting wallet for %s", unbip(assetID))
+		err := c.connectWallet(wallet)
 		if err != nil {
-			return nil, fmt.Errorf("Connect error: %w", err)
+			return nil, err // core.Error with code connectWalletErr
```
chappjc, line 1041: The work @amassarwi is planning to resolve https://github.com/decred/dcrdex/issues/879 is going to be working around wallet connection, so we should get this merged to minimize conflict.  @buck54321 any comments on this PR?

chappjc: I suspect an appropriate place to verify the `ismine` status of the deposit address would be in `connectWallet` around here: https://github.com/decred/dcrdex/blob/0441ddaed67c46d792b68daff07994c194fde0d9/client/core/core.go#L1055-L1061  But do investigate and determine if that's right.

amassarwi: Sweet, will take from here

amassarwi (commented):
👍

client/asset/btc/btc.go
```diff
@@ -1083,6 +1080,9 @@ func (btc *ExchangeWallet) Unlock(pw string) error {
 
 // Lock locks the ExchangeWallet and the underlying bitcoind wallet.
 func (btc *ExchangeWallet) Lock() error {
+	if btc.node.Disconnected() {
+		return asset.ErrConnectionDown
+	}
```
buck54321, line 1085: From what I can tell, since we're using `HTTPPostMode`, `Disconnected` is always `false`, (ultimately because `rpcclient.Client.disconnected` will never  be `true`).

PR #898, opened 2020-12-18
client: validate max fee rate at match time

buck54321:
This is part of a two-part solution to plug a hole where the server could maliciously assign an exorbitant fee rate. 

This solution is for the match validation side, but there is still a potential path for sneaky fee rate increases because of the gap between the time when the max fee rate is observed and the time the order is placed, during which a malicious server could sneak in a higher max fee rate in a config change.


amassarwi:
👍

client/db/bolt/upgrades.go
```diff
@@ -143,3 +146,36 @@ func v1Upgrade(dbtx *bbolt.Tx) error {
...
+	}
+
+	// For each order, set a maxfeerate of max uint64.
+	maxFeeB := uint64Bytes(^uint64(0))
+
+	master := dbtx.Bucket(ordersBucket)
+	if master == nil {
+		return fmt.Errorf("failed to open orders bucket")
+	}
+
```
chappjc, line 173: Need to update the stored version to 2, no?  `setDBVersion(dbtx, newVersion)` and a check in `verifyV2Upgrade`

chappjc (changes requested):

JoeGruffins (approved):

client/db/bolt/upgrades.go
```diff
@@ -85,8 +88,17 @@ func (db *BoltDB) upgradeDB() error {
...
+			// err := upgrade(tx)
+			// if err != nil {
+			// 	return fmt.Errorf("error upgrading DB: %v", err)
+			// }
+			// // Persist the database version.
+			// err = setDBVersion(tx, uint32(i+1))
+			// if err != nil {
+			// 	return fmt.Errorf("error setting DB version: %v", err)
+			// }
+			err := doUpgrade(tx, upgrade, uint32(i+1))
```
chappjc, line 101: Thinking about the value of `i` in this loop for different values of `version`... unless I'm mistaken, `uint32(i+1)` here is appropriate only for when the starting `version` is 0.

Say we have versions all the way up to 10, and we start up with v5 db file, then in `for i, upgrade := range upgrades[version:]` the first iteration of the loop will be `i=0`, telling `doUpgrade` that `newVersion` is 1, but it needs to be 6.
Even for a v1 db, it seems it will set the version to 1 after the v2 upgrade.  Perhaps a quick test of the `upgradeDB` function?

JoeGruffins: Obvious now that you point it out. When I look again, I see `[INF] CORE[DB]: Upgrading database from version 1 to 2` every time I start the app, meaning that after the upgrade the version is being set to `1` again.

chappjc (approved):

PR #899, opened 2020-12-18
allow apps to listen on IPv6 addresses

chappjc:
This is based on PR #882, so the commit of interest is https://github.com/decred/dcrdex/commit/bfb407e2ccd4005c3dccaf783a1b6ddfb5148cfa.

The issue is that `host + ":" port` is not always correct for IPv6 because the host needs brackets around it when joined with a port.  This updates all the listen address parsing and construction code to work properly with IPv6.

For dcrdex with `--rpclisten=[::1]:17232`:

**before**

`NewServer failed: address ::1:17232: too many colons in address`

**after**

`[INF] COMM: RPC server listening on [::1]:17232`

A similar fix is applied to client.

### DIFF
--- a/client/cmd/dexc/config.go
+++ b/client/cmd/dexc/config.go
@@ -5,6 +5,7 @@ package main
 
 import (
 	"fmt"
+	"net"
 	"os"
 	"os/user"
 	"path/filepath"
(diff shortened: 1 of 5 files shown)


buck54321:
This ready after rebase? Seems good to me.

JoeGruffins (approved):
Working well with localhost. I know nothing about ipv6 though. Should logs also have brackets?
```
2020-12-28 15:43:43.434 [TRC] COMM[WS]: Sent 37 and dropped 0 messages to ::1 before shutdown.
2020-12-28 15:43:43.434 [DBG] AUTH: Link down: id=1, ip=::1.
2020-12-28 15:43:43.434 [TRC] COMM: Disconnected websocket client ::1
```

chappjc:
@JoeGruffins perhaps brackets would look nicer there, but the net.IP stringer doesn't put them in and I'm not sure it's worth the extra work to wrap the string if it's ipv6.  The brackets would be totally necessary for readability if there were a port though.

PR #900, opened 2020-12-22
change dcrdex harness port

chappjc:
dcrdex listens on 7232 by default.  For some reason our harnesses use 172**73**, which deviates from the usual tweak of adding or modifying a single digit.  I suspect this was a typo from months ago that just stuck.  However, every so often I try to interact manually with the dcrdex harness and it takes me a while to recall this oddity.  This changes the port to 17232 (i.e. 7232 + 10000)

### DIFF
--- a/client/cmd/dexcctl/simnet-setup.sh
+++ b/client/cmd/dexcctl/simnet-setup.sh
@@ -12,7 +12,7 @@ echo configuring Bitcoin wallet
 echo configuring Litecoin wallet
 ./dexcctl -p abc -p "" --simnet newwallet 2 ~/dextest/ltc/alpha/alpha.conf '{"walletname":"gamma"}'
 echo registering with DEX
-./dexcctl -p abc --simnet register 127.0.0.1:17273 100000000 ~/dextest/dcrdex/rpc.cert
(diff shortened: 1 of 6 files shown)


buck54321:
Fine, but I think the convention is really to add 10k for **testnet**, and then add one to that for **simnet**. If you suspect that we'll ever want a separate default port for testnet, now might be the time to do it.

chappjc:
Hmm good point @buck54321.  I skipped a network here.

Although checking with dcrd, I don't know what the heck the simnet convention is:

```
$ dcrd --simnet
...
[INF] RPCS: RPC server listening on [::]:19556
[INF] CMGR: Server listening on [::]:18555                          // vs 9108 (mainnet) and 19108 (testnet)
[INF] RPCS: RPC server listening on 127.0.0.1:19556
[INF] CMGR: Server listening on 0.0.0.0:18555
```

bitcoind seems to have its own bizarre convention.

JoeGruffins (approved):

chappjc:
Although I find 17273 (not 17232) very confusing, this is not inappropriate for simnet since testnet would be 17273.  Closing.  Maybe we change all these dcrdex simnet ports to something like 17255 in the future, but it's not worth messing with this now.

PR #901, opened 2020-12-23
client: Update wallet's settings & password simultaneously

### LINKED ISSUE #855: Update btc wallet from unencrypted to encrypted
There's currently no way to update the btc wallet config from an unencrypted wallet to an encrypted wallet.  

When you try to update the wallet name first, you get the following error:
`reconfig error: wallet successfully connected, but errored unlocking. reconfiguration not saved: unlockWallet unlock error: Wallet reporting as locked, but no password has been set`

When you try to update the password first, you get the following error:
`password change error: Error unlocking wallet. Is the new password correct?: rawrequest error: -15: Error: running with an unencrypted wallet, but walletpassphrase was called.`

The gui does not allow you to update both the wallet name and password concurrently.

amassarwi:
This adds the possibility to update wallet's settings/name & password simultaneously.
With this users which used an unencrypted wallet could switch to a different encrypted wallet.

Closes #855

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1579,18 +1579,10 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg 
 			assetID, unbip(assetID), err)
 	}
 
-	newPasswordSet := newWalletPW != nil // INcludes empty but non-nil
+	newPasswordSet := newWalletPW != nil // Includes empty but non-nil
 	// If newWalletPW is non-nil, update the wallet's password.
 	if newPasswordSet {
-		// Encrypt password if it's not an empty string
-		encPW := newWalletPW
-		if string(newWalletPW) != "" {
-			encPW, err = crypter.Encrypt(newWalletPW)
-			if err != nil {
-				return newError(encryptionErr, "encryption error: %v", err)
-			}
-		}
-		err = c.setWalletPassword(wallet, newWalletPW, encPW)
+		err = c.setWalletPassword(wallet, newWalletPW, crypter)
 		if err != nil {
 			return err
 		}
@@ -1703,33 +1695,10 @@ func (c *Core) SetWalletPassword(appPW []byte, assetID uint32, newPW []byte) err
 		return newError(missingWalletErr, "wallet for %s (%d) is not known", unbip(assetID), assetID)
 	}
 
-	// Connect if necessary.
-	wasConnected := wallet.connected()
-	if !wasConnected {
-		if err = c.connectAndUpdateWallet(wallet); err != nil {
-			return newError(connectionErr, "SetWalletPassword connection error: %v", err)
-		}
-	}
-
-	wasUnlocked := wallet.unlocked()
-	var encPW []byte
 	if newPasswordSet {
-		encPW, err = crypter.Encrypt(newPW)
-		if err != nil {
-			return newError(encryptionErr, "encryption error: %v", err)
-		}
-		err = c.setWalletPassword(wallet, newPW, encPW)
+		err = c.setWalletPassword(wallet, newPW, crypter)
 		if err != nil {
 			return err
-		}
-	}
-	// TODO: else wallet.encPW = nil?
-
-	if !wasConnected {
-		wallet.Disconnect()
-	} else if !wasUnlocked {
-		if err = wallet.Lock(); err != nil {
-			c.log.Warnf("Unable to relock %s wallet: %v", unbip(assetID), err)
 		}
 	}
 
@@ -1737,7 +1706,16 @@ func (c *Core) SetWalletPassword(appPW []byte, assetID uint32, newPW []byte) err
 }
 
 // setWalletPassword updates the (encrypted) password for the wallet.
-func (c *Core) setWalletPassword(wallet *xcWallet, newPW, encNewPW []byte) error {
+func (c *Core) setWalletPassword(wallet *xcWallet, newPW []byte, crypter encrypt.Crypter) error {
+	// Connect if necessary.
+	wasConnected := wallet.connected()
+	if !wasConnected {
+		if err := c.connectAndUpdateWallet(wallet); err != nil {
+			return newError(connectionErr, "SetWalletPassword connection error: %v", err)
+		}
+	}
+
+	wasUnlocked := wallet.unlocked()
 	newPasswordSet := len(newPW) > 0 // excludes empty but non-nil
 
 	// Check that the new password works. If the new password is empty, skip
@@ -1746,7 +1724,12 @@ func (c *Core) setWalletPassword(wallet *xcWallet, newPW, encNewPW []byte) error
 	// otherwise does not require a password. Perhaps an
 	// asset.Wallet.RequiresPassword wallet method?
 	if newPasswordSet {
-		err := wallet.Wallet.Unlock(string(newPW))
+		// Encrypt password if it's not an empty string
+		encNewPW, err := crypter.Encrypt(newPW)
+		if err != nil {
+			return newError(encryptionErr, "encryption error: %v", err)
+		}
+		err = wallet.Wallet.Unlock(string(newPW))
 		if err != nil {
 			return newError(authErr,
 				"setWalletPassword unlocking wallet error, is the new password correct?: %v", err)
@@ -1759,6 +1742,14 @@ func (c *Core) setWalletPassword(wallet *xcWallet, newPW, encNewPW []byte) error
 	err := c.db.SetWalletPassword(wallet.dbID, wallet.encPW)
 	if err != nil {
 		return codedError(dbErr, err)
+	}
+
+	if !wasConnected {
+		wallet.Disconnect()
+	} else if !wasUnlocked {
+		if err = wallet.Lock(); err != nil {
+			c.log.Warnf("Unable to relock %s wallet: %v", unbip(wallet.AssetID), err)
+		}
 	}
 
 	details := fmt.Sprintf("Password for %s wallet has been updated.",
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1551,8 +1551,9 @@ func (c *Core) WalletSettings(assetID uint32) (map[string]string, error) {
 	return dbWallet.Settings, nil
 }
 
-// ReconfigureWallet updates the wallet configuration settings.
-func (c *Core) ReconfigureWallet(appPW []byte, assetID uint32, cfg map[string]string) error {
+// ReconfigureWallet updates the wallet configuration settings, it also updates
+// the password if newWalletPW is non-nil.
+func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg map[string]string) error {
 	crypter, err := c.encryptionKey(appPW)
 	if err != nil {
 		return newError(authErr, "ReconfigureWallet password error: %v", err)
@@ -1561,7 +1562,8 @@ func (c *Core) ReconfigureWallet(appPW []byte, assetID uint32, cfg map[string]st
 	defer c.walletMtx.Unlock()
 	oldWallet, found := c.wallets[assetID]
 	if !found {
-		return newError(missingWalletErr, "%d -> %s wallet not found", assetID, unbip(assetID))
+		return newError(missingWalletErr, "%d -> %s wallet not found",
+			assetID, unbip(assetID))
 	}
 	dbWallet := &db.Wallet{
 		AssetID:     oldWallet.AssetID,
@@ -1573,7 +1575,21 @@ func (c *Core) ReconfigureWallet(appPW []byte, assetID uint32, cfg map[string]st
 	// Reload the wallet with the new settings.
 	wallet, err := c.loadWallet(dbWallet)
 	if err != nil {
-		return newError(walletErr, "error loading wallet for %d -> %s: %v", assetID, unbip(assetID), err)
+		return newError(walletErr, "error loading wallet for %d -> %s: %v",
+			assetID, unbip(assetID), err)
+	}
+
+	newPasswordSet := newWalletPW != nil // INcludes empty but non-nil
+	// If newWalletPW is non-nil, update the wallet's password.
+	if newPasswordSet {
+		encPW, err := crypter.Encrypt(newWalletPW) // TODO: not correct if string(newWalletPW) == ""
+		if err != nil {
+			return newError(encryptionErr, "encryption error: %v", err)
+		}
+		err = c.setWalletPassword(wallet, newWalletPW, encPW)
+		if err != nil {
+			return err
+		}
 	}
 
 	// Must connect to ensure settings are good.
@@ -1582,10 +1598,11 @@ func (c *Core) ReconfigureWallet(appPW []byte, assetID uint32, cfg map[string]st
 		return err
 	}
 
-	// Carry over any cached password regardless of backend lock state.
-	// loadWallet already copied encPW, so this will decrypt pw rather than
-	// actually copying it, and it will ensure the backend is also unlocked.
-	if oldWallet.locallyUnlocked() {
+	// If the password was not changed, carry over any cached password
+	// regardless of backend lock state. loadWallet already copied encPW, so
+	// this will decrypt pw rather than actually copying it, and it will
+	// ensure the backend is also unlocked.
+	if !newPasswordSet && oldWallet.locallyUnlocked() {
 		err := wallet.Unlock(crypter)
 		if err != nil {
 			wallet.Disconnect()
@@ -1632,8 +1649,10 @@ func (c *Core) ReconfigureWallet(appPW []byte, assetID uint32, cfg map[string]st
 	c.wallets[assetID] = wallet
 
 	c.notify(newBalanceNote(assetID, balances)) // redundant with wallet config note?
-	details := fmt.Sprintf("Configuration for %s wallet has been updated. Deposit address = %s", unbip(assetID), wallet.address)
-	c.notify(newWalletConfigNote(SubjectWalletConfigurationUpdated, details, db.Success, wallet.state()))
+	details := fmt.Sprintf("Configuration for %s wallet has been updated. Deposit address = %s",
+		unbip(assetID), wallet.address)
+	c.notify(newWalletConfigNote(SubjectWalletConfigurationUpdated,
+		details, db.Success, wallet.state()))
 
 	// Clear any existing tickGovernors for suspect matches.
 	c.connMtx.RLock()
@@ -1670,7 +1689,7 @@ func (c *Core) SetWalletPassword(appPW []byte, assetID uint32, newPW []byte) err
 		return newError(authErr, "SetWalletPassword password error: %v", err)
 	}
 
-	newPasswordSet := len(newPW) > 0
+	newPasswordSet := len(newPW) > 0 // excludes empty but non-nil
 
 	// Check that the specified wallet exists.
 	c.walletMtx.Lock()
@@ -1688,15 +1707,19 @@ func (c *Core) SetWalletPassword(appPW []byte, assetID uint32, newPW []byte) err
 		}
 	}
 
-	// Check that the new password works. If the new password is empty, skip
-	// this step, since an empty password signifies an unencrypted wallet.
 	wasUnlocked := wallet.unlocked()
+	var encPW []byte
 	if newPasswordSet {
-		err = wallet.Wallet.Unlock(string(newPW))
+		encPW, err = crypter.Encrypt(newPW)
 		if err != nil {
-			return newError(authErr, "Error unlocking wallet. Is the new password correct?: %v", err)
+			return newError(encryptionErr, "encryption error: %v", err)
+		}
+		err = c.setWalletPassword(wallet, newPW, encPW)
+		if err != nil {
+			return err
 		}
 	}
+	// TODO: else wallet.encPW = nil?
 
 	if !wasConnected {
 		wallet.Disconnect()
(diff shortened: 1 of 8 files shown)


buck54321 (changes requested):
UI looks good. One thing I think you should consider is adding a way to hide the new password subform. Sort of a "nevermind, I don't want to change the password" button.  See the "show additional options" / "hide additional options" function on the new wallet form for an example.

client/webserver/site/src/js/wallets.js
```diff
@@ -37,11 +35,16 @@ export default class WalletsPage extends BasePage {
       'withdrawUnit', 'withdrawPW', 'withdrawErr'
     ])
 
+    // Hide update password section by default
+    this.changeWalletPW = false
+    Doc.hide(page.changePW)
+    Doc.show(page.showChangePW)
```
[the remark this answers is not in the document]
amassarwi: True, will keep the one in `showReconfig` as it needed to handle the case when user clicks on other asset's settings
gear while editing the other asset's settings - it should default hidden in that case.

amassarwi:
@buck54321 updated mate, with the last commit Core does not export `SetWalletPassword`, instead it's been called
from `ReconfigureWallet` to make it possible to update settings & password simultaneously.

client/core/core.go
```diff
@@ -1515,8 +1517,20 @@ func (c *Core) ReconfigureWallet(appPW []byte, assetID uint32, cfg map[string]st
...
+
+	// If provided SetNewPW flag is true update wallet's password
+	if setNewPW {
+		err = c.setWalletPassword(appPW, wallet, newPW)
+		if err != nil {
+			return newError(walletErr,
+				"error setting wallet password for %d -> %s: %v",
+				assetID, unbip(assetID), err)
+		}
+	}
```
JoeGruffins, line 1532: You *could* just do this based on whether `newPW` is `!= nil`. I don't think you need the boolean, but I guess it doesn't hurt.

client/core/core_test.go
```diff
@@ -4806,70 +4816,6 @@ func TestReconfigureWallet(t *testing.T) {
 	}
 }
 
-func TestSetWalletPassword(t *testing.T) {
```
JoeGruffins, line 4809: Why remove a test? It would be much better to adjust it so that it works with the changes.

JoeGruffins (commented):
I tried this out and had confusing messaged from the ui and cli logs. I switched to the simnet trading1 wallet. The cli says:
```
2020-12-28 16:40:25.813 [INF] CORE[dcr]: Setting up new DCR wallet at 127.0.0.1:19581 with TLS certificate "/home/joe/dextest/dcr/trading1/rpc.cert".
2020-12-28 16:40:25.845 [INF] CORE[dcr]: Connected to Decred wallet at 127.0.0.1:19581
2020-12-28 16:40:25.846 [INF] CORE[dcr]: Connected to dcrwallet (JSON-RPC API v8.4.0) proxying dcrd (JSON-RPC API v6.2.0) on SimNet
2020-12-28 16:40:25.850 [INF] CORE: notify: |SUCCESS| (walletconfig) Wallet Password Updated - Password for dcr wallet has been updated.
2020-12-28 16:40:25.850 [ERR] WEB: reconfig error: Decred Wallet connect error: websocket client has already connected
```
The web ui said successful in the top right as well. But the bottom of the form shows the websocket failure as well. I am able to use that wallet and trade as normal after unlocking it.

client/core/core.go
amassarwi: That's was my first assumption but then @buck54321 mentioned that it's possible to set empty password.

Think of the case when have encrypted wallet & want to change to unencrypted wallet, so change name & set epty password which considered as unencrypted.

client/core/core_test.go
amassarwi: Hrmmm, `SetWalletPassword` isn't exported any more from core and the logic was moved to `RecofnigureWallet`
so I added one test case to the ReconfigureWallet tests to test settings new password using `ReconfigureWallet`

amassarwi:
@JoeGruffins 
I'm Investigating the switch failure you reported above.

amassarwi:
@JoeGruffins Should be fixed now, when updating both settings and password, and changed the order of the
execution, first connectWallet and then set new password logic.

client/core/core.go
```diff
@@ -1595,29 +1614,21 @@ func (c *Core) ReconfigureWallet(appPW []byte, assetID uint32, cfg map[string]st
...
-
-	// Connect if necessary.
 	wasConnected := wallet.connected()
 	if !wasConnected {
 		if err = c.connectWallet(wallet); err != nil {
-			return newError(connectionErr, "SetWalletPassword connection error: %v", err)
+			return newError(connectionErr, "setWalletPassword connection error: %v",
+				err)
 		}
 	}
```
chappjc, line 1634: I noticed we have an existing issue with the password check below.  That is, it only checks the password if one is set, but it might be incorrect to not have one set. 

In the `// Check that the new password works...` comment below, maybe create a TODO:

```go
// TODO: find a way to verify that the wallet actually is unencrypted or otherwise does
// not require a password. Perhaps an asset.Wallet.RequiresPassword wallet method?
```

I'm thinking of a method added to client/asset.Wallet like:

```go
// RequiresPassword indicates if the wallet requires a password for certain
// operations such as sending funds.
// RequiresPassword(bool, error) // TODO
```

client/core/core.go
```diff
@@ -1492,8 +1492,9 @@ func (c *Core) WalletSettings(assetID uint32) (map[string]string, error) {
 	return dbWallet.Settings, nil
 }
 
-// ReconfigureWallet updates the wallet configuration settings.
-func (c *Core) ReconfigureWallet(appPW []byte, assetID uint32, cfg map[string]string) error {
+// ReconfigureWallet updates the wallet configuration settings, it also updates
+// the password when setNewPW flag is true.
+func (c *Core) ReconfigureWallet(appPW, newPW []byte, assetID uint32, cfg map[string]string, setNewPW bool) error {
```
chappjc, line 1497:
```suggestion
// ReconfigureWallet updates the wallet configuration settings, it also updates
// the password if newWalletPW is non-nil.
func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg map[string]string) error {
```

client/webserver/api.go
```diff
@@ -427,19 +407,26 @@ func (s *WebServer) apiOrder(w http.ResponseWriter, r *http.Request) {
 // apiReconfig sets new configuration details for the wallet.
 func (s *WebServer) apiReconfig(w http.ResponseWriter, r *http.Request) {
 	form := &struct {
-		AssetID uint32            `json:"assetID"`
-		Config  map[string]string `json:"config"`
-		AppPW   encode.PassBytes  `json:"pw"`
+		AssetID  uint32            `json:"assetID"`
+		Config   map[string]string `json:"config"`
+		SetNewPW bool              `json:"setNewPW"`
```
chappjc, line 412: Can eliminate this `SetNewPW` field using the `NewPW != nil` check in `ReconfigureWallet`

chappjc (changes requested):

client/webserver/site/src/js/wallets.js
```diff
@@ -150,6 +147,21 @@ export default class WalletsPage extends BasePage {
...
+  setPWSettingViz (visible) {
+    if (visible) {
+      Doc.hide(this.page.showIcon)
+      Doc.show(this.page.hideIcon, this.page.changePW)
+      this.page.switchPWMsg.textContent = 'exclude password field'
+      return
+    }
+    Doc.hide(this.page.hideIcon, this.page.changePW)
+    Doc.show(this.page.showIcon)
+    this.page.switchPWMsg.textContent = 'include password field'
```
chappjc, line 162: Forgot to mention that I find "include password field" and "exclude password field" very confusing as they hint at implementation and not meaning to the user.

How about "+ set a new wallet password" and "- do not set a new wallet password"?

amassarwi:
@chappjc Updated 🔔

client/core/core.go
JoeGruffins: I see.

amassarwi: We could use != nil here as we use []byte and not a string which has a default value.

See my last commit: 5710d0e

JoeGruffins:
When trying to update settings, I get `reconfig error: ReconfigureWallet password error: encryption key deserialization error: incorrect password` but I'm pretty sure the password is right. Do you see this?

chappjc:
I got that a bunch of times too, @JoeGruffins , but it's seemed to go back to normal after I rebuilt everything.  But it was also with a bunch of my proposed changes that it stopped that, so maybe it's related.

JoeGruffins (approved):
Oh yeah. After revisiting and rebuilding it's working. Guess I had missed building something. I still feel pretty strongly about not removing a test, but otherwise works for me.

chappjc:
Yeah, I agree the test can stay, just reworked a little bit for the unexpired method.

amassarwi:
@JoeGruffins @chappjc I've added the adjusted `TestSetWalletPassword`.

chappjc:
> Yeah, you need to verify that `ReconfigureWallet` does indeed _call_ `setWalletPassword`, giving coverage to the few lines in `ReconfigureWallet` that deal with this case. All the password edge cases can be left to `TestSetWalletPassword`, as you have done.

Oh dang, I just looked closer at `TestSetWalletPassword`... I wasn't expecting it would be testing `setWalletPassword` **via`ReconfigureWallet`**.  I assumed you'd be adding in tests that called `setWalletPassword` directly.  That's fine though.  Sorry for the misunderstanding.

chappjc:
So, I'm recommending restoring an exported `SetWalletPassword` (which wraps the unexported one) on account of it being impossible to change the wallet password without also providing the wallet settings map, which is a burden for other Go consumers as it requires retrieving the current wallet settings only to pass it back into `ReconfigureWallet`.  The original `SetWalletPassword` tests can be restored almost verbatim as a consequence.  This will also provide better coverage for `setWalletPassword` since `ReconfigureWallet` was intercepting certain error cases that `setWalletPassword` would have handled. However, we should ensure the error codes are consistent as I noticed one returns `connectWalletErr` and the other `connectionErr`.

amassarwi:
Updated to follow @chappjc's suggestions.

amassarwi:
Needs a rebase after #902 merged.

amassarwi:
Done

buck54321 (commented):
Panic when trying to change my password.

```
Panic: runtime error: invalid memory address or nil pointer dereference
goroutine 150 [running]:
runtime/debug.Stack(0x48, 0x0, 0x0)
	/usr/local/go/src/runtime/debug/stack.go:24 +0x9f
runtime/debug.PrintStack()
	/usr/local/go/src/runtime/debug/stack.go:16 +0x25
github.com/go-chi/chi/middleware.Recoverer.func1.1(0xc005030400, 0xe962c0, 0xc0001f0000)
	/home/buck/go/pkg/mod/github.com/go-chi/chi@v4.0.2+incompatible/middleware/recoverer.go:28 +0x1bf
panic(0xb85520, 0x1268a70)
	/usr/local/go/src/runtime/panic.go:969 +0x175
github.com/decred/dcrd/rpcclient/v6.receiveFuture(0x0, 0x0, 0xc000a8c6c0, 0xc66da3, 0x10, 0xc000939920, 0x2, 0x2)
	/home/buck/go/pkg/mod/github.com/decred/dcrd/rpcclient/v6@v6.0.2/infrastructure.go:873 +0x40
github.com/decred/dcrd/rpcclient/v6.(*FutureRawResult).Receive(...)
	/home/buck/go/pkg/mod/github.com/decred/dcrd/rpcclient/v6@v6.0.2/rawrequest.go:23
github.com/decred/dcrd/rpcclient/v6.(*Client).RawRequest(0xc0001eb0e0, 0x0, 0x0, 0xc66da3, 0x10, 0xc000939920, 0x2, 0x2, 0x7f1fe6cd1c58, 0x0, ...)
	/home/buck/go/pkg/mod/github.com/decred/dcrd/rpcclient/v6@v6.0.2/rawrequest.go:79 +0xa5
decred.org/dcrwallet/rpc/client/dcrwallet.(*rawRequester).Call(0xc00083e3c0, 0x0, 0x0, 0xc66da3, 0x10, 0x0, 0x0, 0xc000a0e1e0, 0x2, 0x2, ...)
	/home/buck/go/pkg/mod/decred.org/dcrwallet@v1.6.0-rc4/rpc/client/dcrwallet/client.go:44 +0x234
decred.org/dcrwallet/rpc/client/dcrwallet.(*Client).WalletPassphrase(0xc000a0e180, 0x0, 0x0, 0xc0005d445c, 0x3, 0x225c17d04, 0x0, 0x0)
	/home/buck/go/pkg/mod/decred.org/dcrwallet@v1.6.0-rc4/rpc/client/dcrwallet/methods.go:562 +0x119
decred.org/dcrdex/client/asset/dcr.(*ExchangeWallet).Unlock(0xc0050084e0, 0xc0005d445c, 0x3, 0xc0005d445c, 0x3)
	/home/buck/programs/dcrdex/client/asset/dcr/dcr.go:1866 +0x70
decred.org/dcrdex/client/core.(*Core).setWalletPassword(0xc000368000, 0xc0001ea140, 0xc000025370, 0x3, 0x3, 0xc000c40300, 0x2e, 0x40, 0x0, 0x0)
	/home/buck/programs/dcrdex/client/core/core.go:1729 +0x435
decred.org/dcrdex/client/core.(*Core).ReconfigureWallet(0xc000368000, 0xc00002533d, 0x3, 0x3, 0xc000025370, 0x3, 0x3, 0x2a, 0xc0007139e0, 0x0, ...)
	/home/buck/programs/dcrdex/client/core/core.go:1576 +0x1106
```

client/webserver/api.go
```diff
@@ -339,26 +339,6 @@ func (s *WebServer) apiWalletSettings(w http.ResponseWriter, r *http.Request) {
...
-	if !readPost(w, r, form) {
-		return
-	}
-	err := s.core.SetWalletPassword(form.AppPW, form.AssetID, form.NewPW)
-	if err != nil {
-		s.writeAPIError(w, "password change error: %v", err)
-		return
-	}
-	writeJSON(w, simpleAck(), s.indent)
-}
```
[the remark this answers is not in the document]
chappjc: No this has too much potential to cause problems for 0.1

client/webserver/api.go
```diff
@@ -427,19 +407,25 @@ func (s *WebServer) apiOrder(w http.ResponseWriter, r *http.Request) {
 // apiReconfig sets new configuration details for the wallet.
 func (s *WebServer) apiReconfig(w http.ResponseWriter, r *http.Request) {
 	form := &struct {
-		AssetID uint32            `json:"assetID"`
-		Config  map[string]string `json:"config"`
-		AppPW   encode.PassBytes  `json:"pw"`
+		AssetID     uint32            `json:"assetID"`
+		Config      map[string]string `json:"config"`
+		NewWalletPW encode.PassBytes  `json:"newWalletPW"`
```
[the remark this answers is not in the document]
chappjc: The plan was it for caller to _omit_ the field, not make pass `null`.

amassarwi:
@buck54321 can't repro the panic, tried to set a nonempty password & then empty password again and it worked as expected.

client/webserver/api.go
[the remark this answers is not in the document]
amassarwi: Comment added

amassarwi:
@buck54321  Other comment addressed.

amassarwi:
Currently when I switch from encrypted wallet to unencrypted wallet successfully I'm not able to lock/unlock the wallet.
Having a look ..

amassarwi:
> Currently when I switch from encrypted wallet to unencrypted wallet successfully I'm not able to lock/unlock the wallet.
> Having a look ..

This seems to be related to:
https://github.com/decred/dcrdex/pull/901/files#diff-13a4a6b12c1ae844ef90151638f194427cfdc545fceb58d52c0f6f33b81d859bR1720-R1731

When running an encrypted wallet then switching to an unencrypted wallet currently there is so way to unlock the
unencrypted wallet, the next unlock try would use an empty string and then get the following error from the backend wallet:
```
Error: running with an unencrypted wallet, but walletpassphrase was called.
```

It sounds like we need a way to mark an unencrypted wallet as unlocked once the user switches to an unencrypted wallet
from encrypted wallet, thoughts about that ?

cc: @chappjc @buck54321

client/core/core.go
```diff
@@ -1680,24 +1710,37 @@ func (c *Core) SetWalletPassword(appPW []byte, assetID uint32, newPW []byte) err
...
+
+// setWalletPassword updates the (encrypted) password for the wallet.
+func (c *Core) setWalletPassword(wallet *xcWallet, newPW, encNewPW []byte) error {
+	newPasswordSet := len(newPW) > 0
+
+	// Check that the new password works. If the new password is empty, skip
+	// this step, since an empty password signifies an unencrypted wallet.
+	// TODO: find a way to verify that the wallet actually is unencrypted or
+	// otherwise does not require a password. Perhaps an
+	// asset.Wallet.RequiresPassword wallet method?
```
chappjc, line 1724: I think we can work around this but I'll have to give you a suggested change in the a.m.  Just about out of steam for the day...

amassarwi: Sounds great

chappjc: There's some interplay here with the existing unlock bug fixed in https://github.com/decred/dcrdex/pull/910.  I'm afraid it's going to be easier to get that in first.

I recognize the grief that causes you so I'm more than happy to rebase this (and force push to your branch if that's OK) and make a suggested edit after that.  But part of the issue that is preventing reconfigure from an encrypted->unencrypted wallet is that with the added `newWalletPW` arg, we no longer want to create the `db.Wallet` in `ReconfigureWallet` with `EncryptedPW: oldWallet.encPW` unless `newWalletPW == nil`.

amassarwi: I see, makes total sense.

Please feel free to rebaae this PR and I can then add the suggested solution.

chappjc: Just did the rebase and will make some suggestions shortly. I suggest to git fetch and then reset hard your local branch to the new remote that was force pushed.  I did however add a few comments where I though changes were needed to resolve the issue.  Will follow up right away.

client/core/core.go
```diff
@@ -1573,7 +1575,21 @@ func (c *Core) ReconfigureWallet(appPW []byte, assetID uint32, cfg map[string]st
...
 	if err != nil {
-		return newError(walletErr, "error loading wallet for %d -> %s: %v", assetID, unbip(assetID), err)
+		return newError(walletErr, "error loading wallet for %d -> %s: %v",
+			assetID, unbip(assetID), err)
+	}
+
+	newPasswordSet := newWalletPW != nil // INcludes empty but non-nil
+	// If newWalletPW is non-nil, update the wallet's password.
+	if newPasswordSet {
+		encPW, err := crypter.Encrypt(newWalletPW) // TODO: not correct if string(newWalletPW) == ""
```
chappjc, line 1585: This appears to the the root of the issue, but I'm still poking around.

client/core/core.go
amassarwi: @chappjc 
Sweet, got the updated branch locally 👍

client/core/core.go
```diff
@@ -1573,7 +1575,25 @@ func (c *Core) ReconfigureWallet(appPW []byte, assetID uint32, cfg map[string]st
...
+	if newPasswordSet {
+		// Encrypt password if it's not an empty string
+		encPW := newWalletPW
+		if string(newWalletPW) != "" {
+			encPW, err = crypter.Encrypt(newWalletPW)
+			if err != nil {
+				return newError(encryptionErr, "encryption error: %v", err)
+			}
+		}
+		err = c.setWalletPassword(wallet, newWalletPW, encPW)
```
buck54321, line 1593: Doesn't `setWalletPassword` unlock the wallet? It should definitely check the password, but this is a change in behavior for a user setting a password on a locked wallet, since the wallet will be unlocked afterwards. There was some previous effort made to ensure the lock state was the same after as it was before. 

In `SetWalletPassword`, this is handled with the `wasUnlocked` variable. You might consider moving that logic into the unexported `setWalletPassword` method.

client/core/core.go
```diff
@@ -1688,15 +1711,19 @@ func (c *Core) SetWalletPassword(appPW []byte, assetID uint32, newPW []byte) err
...
+		if err != nil {
+			return newError(encryptionErr, "encryption error: %v", err)
+		}
+		err = c.setWalletPassword(wallet, newPW, encPW)
 		if err != nil {
-			return newError(authErr, "Error unlocking wallet. Is the new password correct?: %v", err)
+			return err
 		}
 	}
+	// TODO: else wallet.encPW = nil?
```
buck54321, line 1726: I believe you can move most of the `if newPasswordSet` blocks in `SetWalletPassword` and `ReconfigureWallet` into the unexported `setWalletPassword`, you'll just pass the unencrypted PW instead of the encrytped.

You're handling the `encPW = nil` part correctly in `setWalletPassword`.

client/core/core.go
```diff
@@ -1573,7 +1575,25 @@ func (c *Core) ReconfigureWallet(appPW []byte, assetID uint32, cfg map[string]st
 	// Reload the wallet with the new settings.
 	wallet, err := c.loadWallet(dbWallet)
 	if err != nil {
-		return newError(walletErr, "error loading wallet for %d -> %s: %v", assetID, unbip(assetID), err)
+		return newError(walletErr, "error loading wallet for %d -> %s: %v",
+			assetID, unbip(assetID), err)
+	}
+
+	newPasswordSet := newWalletPW != nil // INcludes empty but non-nil
```
buck54321, line 1582: ~~INcludes~~ -> Includes

client/core/core.go
```diff
@@ -1670,7 +1693,7 @@ func (c *Core) SetWalletPassword(appPW []byte, assetID uint32, newPW []byte) err
 		return newError(authErr, "SetWalletPassword password error: %v", err)
 	}
 
-	newPasswordSet := len(newPW) > 0
+	newPasswordSet := len(newPW) > 0 // excludes empty but non-nil
```
buck54321, line 1696: One thing that's a little confusing is that you're using the same name for this variable in both `SetWalletPassword` and `ReconfigureWallet`, but they have different meanings. Maybe try to capture their function more clearly in some new names, e.g. `isSettingNewPW` and `usingNonNilPW` or whatever.

buck54321 (changes requested):
> @buck54321 can't repro the panic, tried to set a nonempty password & then empty password again and it worked as expected.

Not seeing it now either. Looking almost ready. A couple of last suggestions.

client/core/core.go
chappjc: When I did the rebase for @amassarwi, i added these two comments to point out just that distinction.  I agree changing the names would be best.

amassarwi: Agreed, will do

client/core/core.go
amassarwi: good point 👍

amassarwi: I would need to pass the crypter also, as setWalletPassword is taking care
of encrypting the password if it's not empty.

client/core/core.go
amassarwi: moved

client/core/core.go
amassarwi: Actually looks like `SetWalletPassword` shouldn't exclude empty strings, instead it should check if newPW
isn't nil as we do in `ReconfigureWallet`.

We check for non empty string inside `setWalletPassword`.

amassarwi:
@buck54321  @chappjc  updated

client/core/core.go
```diff
@@ -1695,7 +1693,8 @@ func (c *Core) SetWalletPassword(appPW []byte, assetID uint32, newPW []byte) err
 		return newError(missingWalletErr, "wallet for %s (%d) is not known", unbip(assetID), assetID)
 	}
 
-	if newPasswordSet {
+	isSettingNewPW := newPW != nil // includes empty but non-nil
+	if isSettingNewPW {
```
chappjc, line 1697: Makes me wonder what it means to do `core.SetWalletPassword(nil, ...)` and if that should be an error and documented as invalid.  What do you think @amassarwi?

amassarwi: Agreed 💯

amassarwi: will add a test for that as well

amassarwi: done

chappjc:
> > @buck54321 can't repro the panic, tried to set a nonempty password & then empty password again and it worked as expected.
> 
> Not seeing it now either. Looking almost ready. A couple of last suggestions.

Looking into the panic stack dump, it seems like an rpcclient method (`WalletPassphrase`) was called with a nil `Context` via `setWalletPassphrase`>`Unlock` **with an unconnected wallet**, meaning the `client/asset.Wallet` implementation wasn't Connected yet.

I see that a recent commit has since moved the connected check into `setWalletPassphrase` at @buck54321's suggestion, and this solved the issue, but this is something to be mindful of lest we panic elsewhere.  Maybe we should either consider a recover wrapper internal to the ExchangeWallet calls and/or rethink the subsystem context stored there to make it impossible to make such requests.  Nothing specific to change, just wanted to bring it up.

chappjc (approved):
LGTM.  Will let @buck54321 follow up on his suggestions before merging.

buck54321 (approved):

PR #902, opened 2020-12-25
client: Validate known deposit address after wallet connection

### LINKED ISSUE #879: BTC deposit address not in dex wallet

amassarwi:
This fixes #879 by validating deposit address for each wallet after successful connection.

I wiped a running wallet dir and created fresh one as #879 describes and the client
validated the deposit address and generated new one.

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -565,6 +565,15 @@ func (btc *ExchangeWallet) SyncStatus() (bool, float32, error) {
 	return true, 1, nil
 }
 
+// IsAddressMine returns wether given address belongs to wallet or not
+func (btc *ExchangeWallet) IsAddressMine(address string) (bool, error) {
+	ai, err := btc.wallet.GetAddressInfo(address)
+	if err != nil {
+		return false, err
+	}
+	return ai.IsMine, nil
+}
+
 // Balance returns the total available funds in the wallet. Part of the
 // asset.Wallet interface.
 func (btc *ExchangeWallet) Balance() (*asset.Balance, error) {
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -565,8 +565,8 @@ func (btc *ExchangeWallet) SyncStatus() (bool, float32, error) {
 	return true, 1, nil
 }
 
-// IsAddressMine returns wether given address belongs to wallet or not
-func (btc *ExchangeWallet) IsAddressMine(address string) (bool, error) {
+// ValidateAddress returns wether given address belongs to wallet or not
+func (btc *ExchangeWallet) ValidateAddress(address string) (bool, error) {
 	ai, err := btc.wallet.GetAddressInfo(address)
 	if err != nil {
 		return false, err
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -565,7 +565,7 @@ func (btc *ExchangeWallet) SyncStatus() (bool, float32, error) {
 	return true, 1, nil
 }
 
-// ValidateAddress returns wether given address belongs to wallet or not
+// ValidateAddress indicates if an address belongs to the wallet.
 func (btc *ExchangeWallet) ValidateAddress(address string) (bool, error) {
 	ai, err := btc.wallet.GetAddressInfo(address)
 	if err != nil {
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -565,8 +565,8 @@ func (btc *ExchangeWallet) SyncStatus() (bool, float32, error) {
 	return true, 1, nil
 }
 
-// ValidateAddress indicates if an address belongs to the wallet.
-func (btc *ExchangeWallet) ValidateAddress(address string) (bool, error) {
+// OwnsAddress indicates if an address belongs to the wallet.
+func (btc *ExchangeWallet) OwnsAddress(address string) (bool, error) {
 	ai, err := btc.wallet.GetAddressInfo(address)
 	if err != nil {
 		return false, err
--- a/client/asset/btc/walletclient.go
+++ b/client/asset/btc/walletclient.go
@@ -32,6 +32,7 @@ const (
 	methodSendToAddress     = "sendtoaddress"
 	methodSetTxFee          = "settxfee"
 	methodGetWalletInfo     = "getwalletinfo"
+	methodGetAddressInfo    = "getaddressinfo"
 )
 
 // walletClient is a bitcoind wallet RPC client that uses rpcclient.Client's
@@ -225,7 +226,14 @@ func (wc *walletClient) GetWalletInfo() (*GetWalletInfoResult, error) {
 	return wi, wc.call(methodGetWalletInfo, nil, wi)
 }
 
-// call is used internally to  marshal parmeters and send requests to  the RPC
+// GetAddressInfo gets information about given address by calling
+// getaddressinfo RPC command.
+func (wc *walletClient) GetAddressInfo(address string) (*GetAddressInfoResult, error) {
+	ai := new(GetAddressInfoResult)
+	return ai, wc.call(methodGetAddressInfo, anylist{address}, ai)
+}
+
+// call is used internally to marshal parmeters and send requests to the RPC
 // server via (*rpcclient.Client).RawRequest. If `thing` is non-nil, the result
 // will be marshaled into `thing`.
 func (wc *walletClient) call(method string, args anylist, thing interface{}) error {
--- a/client/asset/btc/walletclient.go
+++ b/client/asset/btc/walletclient.go
@@ -226,7 +226,7 @@ func (wc *walletClient) GetWalletInfo() (*GetWalletInfoResult, error) {
 	return wi, wc.call(methodGetWalletInfo, nil, wi)
 }
 
-// GetAddressInfo gets information about given address by calling
+// GetAddressInfo gets information about the given address by calling
 // getaddressinfo RPC command.
 func (wc *walletClient) GetAddressInfo(address string) (*GetAddressInfoResult, error) {
 	ai := new(GetAddressInfoResult)
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -394,7 +394,7 @@ type combinedClient struct {
 	*walletClient
 }
 
-// ValidateAddress disambiguates the node and wallet methods
+// ValidateAddress disambiguates the node and wallet methods.
 func (cc *combinedClient) ValidateAddress(ctx context.Context, address dcrutil.Address) (*walletjson.ValidateAddressWalletResult, error) {
 	return cc.walletClient.ValidateAddress(ctx, address)
 }
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -554,8 +554,8 @@ func (dcr *ExchangeWallet) Connect(ctx context.Context) (*sync.WaitGroup, error)
 	return &wg, nil
 }
 
-// IsAddressMine returns wether given address belongs to wallet or not
-func (dcr *ExchangeWallet) IsAddressMine(address string) (bool, error) {
+// ValidateAddress returns wether given address belongs to wallet or not
+func (dcr *ExchangeWallet) ValidateAddress(address string) (bool, error) {
 	a, err := dcrutil.DecodeAddress(address, chainParams)
 	va, err := dcr.node.ValidateAddress(dcr.ctx, a)
 	if err != nil {
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -394,6 +394,10 @@ type combinedClient struct {
 	*walletClient
 }
 
+func (cc *combinedClient) ValidateAddress(ctx context.Context, address dcrutil.Address) (*walletjson.ValidateAddressWalletResult, error) {
+	return cc.walletClient.ValidateAddress(ctx, address)
+}
+
 // Check that ExchangeWallet satisfies the Wallet interface.
 var _ asset.Wallet = (*ExchangeWallet)(nil)
 
@@ -418,7 +422,7 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network) 
 		return nil, fmt.Errorf("DCR ExchangeWallet.Run error: %w", err)
 	}
 	// Beyond this point, only node
-	dcr.node = combinedClient{dcr.client, dcrwallet.NewClient(dcrwallet.RawRequestCaller(dcr.client), chainParams)}
+	dcr.node = &combinedClient{dcr.client, dcrwallet.NewClient(dcrwallet.RawRequestCaller(dcr.client), chainParams)}
 
 	return dcr, nil
 }
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -151,6 +151,7 @@ type rpcClient interface {
 	Disconnected() bool
 	RawRequest(ctx context.Context, method string, params []json.RawMessage) (json.RawMessage, error)
 	WalletInfo(ctx context.Context) (*walletjson.WalletInfoResult, error)
+	ValidateAddress(ctx context.Context, address dcrutil.Address) (*walletjson.ValidateAddressWalletResult, error)
 }
 
 // The rpcclient package functions will return a rpcclient.ErrRequestCanceled
@@ -547,6 +548,16 @@ func (dcr *ExchangeWallet) Connect(ctx context.Context) (*sync.WaitGroup, error)
 		dcr.shutdown()
 	}()
 	return &wg, nil
+}
+
+// IsAddressMine returns wether given address belongs to wallet or not
+func (dcr *ExchangeWallet) IsAddressMine(address string) (bool, error) {
+	a, err := dcrutil.DecodeAddress(address, chainParams)
+	va, err := dcr.node.ValidateAddress(dcr.ctx, a)
+	if err != nil {
+		return false, err
+	}
+	return va.IsMine, nil
 }
 
 // Balance should return the total available funds in the wallet. Note that
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -394,6 +394,7 @@ type combinedClient struct {
 	*walletClient
 }
 
+// ValidateAddress disambiguates the node and wallet methods
 func (cc *combinedClient) ValidateAddress(ctx context.Context, address dcrutil.Address) (*walletjson.ValidateAddressWalletResult, error) {
 	return cc.walletClient.ValidateAddress(ctx, address)
 }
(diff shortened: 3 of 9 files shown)


client/asset/btc/btc.go
```diff
@@ -565,6 +565,15 @@ func (btc *ExchangeWallet) SyncStatus() (bool, float32, error) {
 	return true, 1, nil
 }
 
+// ValidateAddress returns wether given address belongs to wallet or not
```
chappjc, line 568: wether -> whether.  But maybe rephrase like `ValidateAddress indicates if an address belongs to the wallet.`

client/asset/dcr/dcr.go
```diff
@@ -393,6 +394,10 @@ type combinedClient struct {
 	*walletClient
 }
 
+func (cc *combinedClient) ValidateAddress(ctx context.Context, address dcrutil.Address) (*walletjson.ValidateAddressWalletResult, error) {
```
chappjc, line 397: Maybe a quick doc string explaining why this is necessary (disambiguate the node and wallet methods).

client/core/core.go
```diff
@@ -1052,11 +1052,28 @@ func (c *Core) connectedWallet(assetID uint32) (*xcWallet, error) {
...
 		return codedError(connectWalletErr, err)
 	}
+	// Check if known address belongs to connected wallet
+	mine, err := w.ValidateAddress(w.address)
+	if err != nil {
+		return err
+	}
+	if !mine {
+		c.log.Infof("[%v]: Deposit address %v does not belong to connected wallet"+
+			", generating new address", unbip(w.AssetID), w.address)
```
chappjc, line 1071: This should be uncommon, so I would not object to `Warnf`.  Also, since you're logging the old deposit address, you should log the new one too.

client/core/core.go
```diff
@@ -1052,11 +1052,28 @@ func (c *Core) connectedWallet(assetID uint32) (*xcWallet, error) {
...
+		_, err := c.NewDepositAddress(w.AssetID)
```
chappjc, line 1072: No need to shadow `err` with a new variable here.

client/core/core.go
```diff
@@ -1530,7 +1542,13 @@ func (c *Core) ReconfigureWallet(appPW []byte, assetID uint32, cfg map[string]st
 		return newError(addrErr, "error getting wallet address: %v", err)
 	}
 	dbWallet.Address = addr
-	wallet.address = addr
+	wallet.setAddress(addr)
+
+	// Must connect to ensure settings are good.
+	err = c.connectWallet(wallet)
```
chappjc, line 1548: Without having tested these changes, I'm confused about the `ReconfigureWallet` edit because it would seem to be requesting a new deposit address from the wallet backend *before* connecting to the new wallet.

But I'm thinking that with the changes to `connectWallet` that ensures the current deposit address are owned by the wallet, that `ReconfigureWallet` no longer even needs to unconditionally get a new deposit address.

client/core/core.go
```diff
@@ -1677,7 +1695,7 @@ func (c *Core) NewDepositAddress(assetID uint32) (string, error) {
 	}
 
 	c.walletMtx.Lock()
-	w.address = addr
+	w.setAddress(addr)
 	c.walletMtx.Unlock()
```
chappjc, line 1699: I think this locking `Core.walletMtx` was a bug. `walletMtx` is meant to guard the `wallets` map.  The `xcWallet` has it's own mutex, which `setAddress` uses.

Review the use of `walletMtx` and the declaration of the mutex and the `wallets` field in the `Core` struct to see what I mean.  @buck54321 please confirm that there was not another consideration here, but it appears to be a misuse of `Core`'s `walletMtx` instead of the `xcWallet` mutex.

client/core/core.go
```diff
@@ -2039,14 +2056,17 @@ func (c *Core) Login(pw []byte) (*LoginResult, error) {
 					return
 				}
 			}
+			// We wait for connectWalleet to finish before holding the lock as it
+			// may update the deposit address
+			c.walletMtx.Lock()
```
chappjc, line 2061: The walletMtx was locked outside of the loop over `c.wallets` because it guards access to the `wallets` map.  It needs to go back that way for the loop.

But not the the loop starts goroutines, so the goroutines will fire off, the loop will complete, and `c.walletMtx.Unlock()` will allow each of the goroutines that need to lock it to lock it.

client/core/core.go
chappjc, line 1072: I'm also noting what appears to be an awkward use pattern that is making issues (or at least confusion with the walletMtx). `(*Core).NewDepositAddress(uint32)` was previously only consumed by the webserver api for generating new addresses, so it included a lookup of an `xcWallet` by id in the `wallets` map, thus necessitating locking `walletsMtx` for the map access.  However, here in `connectWallet` you already have the target `wcWallet`.  I would factor out the guts of `NewDepositAddress` into an unexported `newDepositAddress(w *xcWallet`)`, which you would use here.  The exported version would remain, doing a lookup by id and then calling the unexported one.

chappjc (changes requested):

client/core/core.go
amassarwi: Sounds good, will do, thanks 👍

client/core/core.go
amassarwi: Oh, that's make sense, will update it, this might explain the deadlock I had when generating new address

client/core/core.go
amassarwi: Yep, that's true, but calling the `NewDepositAddress` inside the connectWallet needed the lock which caused 
a deadlock when generating new address, but as you described above, I would have already the xcWallet and can call the unexported method.

amassarwi:
@chappjc Updated & tested, looks good & clean now, thanks for the patient review

amassarwi:
@chappjc updated again 🔔

client/core/core.go
```diff
@@ -1687,12 +1705,28 @@ func (c *Core) NewDepositAddress(assetID uint32) (string, error) {
...
+	if !exists {
+		return "", newError(missingWalletErr, "no wallet found for %s", unbip(assetID))
+	}
+
+	addr, err := c.newDepositAddress(w)
+	if err != nil {
+		return "", err
+	}
 
 	return addr, nil
```
chappjc, line 1731: Since we've defined `newDepositAddress` above to return an empty string for a non-nil error, it would be acceptable in this case to just do `return c.newDepositAddress(w)`, although generally it's wise to ensure return values work this way when dealing with exported methods from other packages.

Anyway, I've not actually tested any of this PR, so feel free to hold off for further comments.

amassarwi: Right, can do

client/asset/interface.go
```diff
@@ -147,6 +147,8 @@ type Wallet interface {
 	Refund(coinID, contract dex.Bytes) (dex.Bytes, error)
 	// Address returns an address for the exchange wallet.
 	Address() (string, error)
+	// ValidateAddress indicates if an address belongs to the wallet.
+	ValidateAddress(address string) (bool, error)
```
buck54321, line 151: An address can be valid without being "owned" by the wallet. Maybe `OwnsAddress` instead?

amassarwi: Agreed, in a sec.

client/asset/dcr/dcr.go
```diff
@@ -549,6 +555,19 @@ func (dcr *ExchangeWallet) Connect(ctx context.Context) (*sync.WaitGroup, error)
 	return &wg, nil
 }
 
+// OwnsAddress returns wether given address belongs to wallet or not
```
buck54321, line 558: w**h**ether

client/asset/btc/walletclient.go
```diff
@@ -225,7 +226,14 @@ func (wc *walletClient) GetWalletInfo() (*GetWalletInfoResult, error) {
 	return wi, wc.call(methodGetWalletInfo, nil, wi)
 }
 
-// call is used internally to  marshal parmeters and send requests to  the RPC
+// GetAddressInfo gets information about given address by calling
```
buck54321, line 229: about **the** given address,

client/core/core.go
```diff
@@ -1052,11 +1052,29 @@ func (c *Core) connectedWallet(assetID uint32) (*xcWallet, error) {
 	return wallet, nil
 }
 
+// connectWallet connects to wallet and validate the known deposit address
+// after successful connection, if deposit address does not belong to
```
buck54321, line 1056: ~~after successful connection**. If the** deposit address...

buck54321 (changes requested):
Running the ***simnet-setup.sh*** script, I'm getting an error

`error creating dcr wallet: decoded address is of unknown format: invalid format: version and/or checksum bytes missing`

amassarwi:
@buck54321 updated mate, in CreateWallet the first address generation comes after the first connection, so I needed to skip the address validate in `connectWallet` if no address found for wallet.

client/core/core.go
```diff
@@ -1052,11 +1052,35 @@ func (c *Core) connectedWallet(assetID uint32) (*xcWallet, error) {
...
+	// If found address doesn't belong to wallet generate new one.
+	addr := w.address
+	var mine bool
+	if addr != "" {
+		mine, err = w.OwnsAddress(addr)
+		if err != nil {
+			return err
+		}
+	}
+	if !mine && addr != "" {
```
chappjc, line 1075: The second `if` should be nested in the first.

```go
if addr != "" {
	mine, err := w.OwnsAddress(addr)
	if err != nil {
		return err
	}
	if !mine {
		nAddr, err := c.newDepositAddress(w)
```

client/core/core.go
```diff
@@ -1052,11 +1052,35 @@ func (c *Core) connectedWallet(assetID uint32) (*xcWallet, error) {
...
+// therefore it might hold the wallet lock
 func (c *Core) connectWallet(w *xcWallet) error {
 	err := w.Connect(c.ctx)
 	if err != nil {
 		return codedError(connectWalletErr, err)
 	}
+	// If wallet has deposit address check if it's belongs to connected
+	// wallet.
+	// If found address doesn't belong to wallet generate new one.
+	addr := w.address
```
chappjc, line 1067: `w.address` being accessed here without RLock

client/core/core.go
amassarwi: opps

amassarwi:
@chappjc Updated

JoeGruffins (approved):
Working for me.

JoeGruffins:
I noticed while reviewing this that while the address is updated, if you change to another wallet with the same password, the balance is not updated until you restart the app. Also, changing a wallet mid trade could get you in trouble maybe? `[ERR] CORE: 127.0.0.1:17273 tick: {redeemMatches order e9f674a0e1419e53ffc958d6882985b263b64fcc33190087998e615d751ad152 - {error sending redeem transaction: -4: wallet.DumpWIFPrivateKey: item does not exist:: wallet.LoadPrivateKey: no address with id 121201201d26555503138db12cf615d2c26534ec (is wallet locked?)}}`

amassarwi:
@JoeGruffins Thanks for the review mate, I noticed the balance issue as well and I think it's the same on master
and I was planning to create a separate issue/PR for that, does it make sense to fix this in a different PR ?

Regarding changing wallet in mid of trade no sure what would be the best approach here ?
We could maybe block block settings change while having a active matches ?
 
cc: @chappjc @buck54321

chappjc:
@JoeGruffins What do you mean by "changing a wallet mid trade"?  Do you mean just a ReconfigureWallet or actually swapping out the actual wallet for one with entirely different keys and addresses?

We definitely should allow updating wallet settings at any time.  But if they actually change the underlying wallet, that's a no-no on the user's part.

amassarwi:
@JoeGruffins @chappjc 
Created separated issue to update balance on wallet switch.
Will work on it after merging this one.

client/core/core.go
```diff
@@ -1498,16 +1538,15 @@ func (c *Core) ReconfigureWallet(appPW []byte, assetID uint32, cfg map[string]st
 	if err != nil {
 		return newError(authErr, "ReconfigureWallet password error: %v", err)
 	}
-	c.walletMtx.Lock()
-	defer c.walletMtx.Unlock()
-	oldWallet, found := c.wallets[assetID]
```
chappjc, line 1503: I'm not so sure this mutex change is desirable.  While there is the `(*Core).wallet` method that handles the locking, I think the reasoning for locking it up hear and keeping it locked until the new wallet is written into the map is to make the process atomic.  That is, we are guaranteed to be replacing oldWallet with the new one the old way, but now another goroutine could modify the map between this access and the write below.

chappjc (changes requested):
Just the one remaining comment about the mutex in ReconfigureWallet, but also please review all the docs and ensure they read well and are properly punctuated.

changing BTC from gamma to default wallet:

[WRN] CORE: [btc]: Deposit address bcrt1q0ctryhtspw4zq3d6dgxsycqcq5ty66z405q9cm does not belong to connected wallet, generated new address: bcrt1q322tg0y2hzyp9zztr7d2twdclhqg88anvzxwwr

Changing DCR from beta to alpha account:

[WRN] CORE: [dcr]: Deposit address SsnbEmxCVXskgTHXvf3rEa17NA39qQuGHwQ does not belong to connected wallet, generated new address: SsoiAXujKA2aj3zsVVTNztfEPugwscyUrVz

reflected in web's deposit address:

image

client/core/core.go
amassarwi: Reverted

amassarwi:
Updated 🔔

JoeGruffins:
> What do you mean by "changing a wallet mid trade"? Do you mean just a ReconfigureWallet with the same wallet backend or actually swapping out the actual wallet for one with entirely different keys and addresses?

The second one. Isn't that what the user with the issue did, and the reason we need a different deposit address?

chappjc:
@JoeGruffins yah, for sure the reason.  Just wanted to make sure that error wasn't possible with just a reconfigure during swaps.
This PR is only intended to facilitate wallet switches when there are no active swaps or orders. Perhaps we should put a warning to that effect on the wallet settings page.

JoeGruffins:
At lease a warning would be prudent imo. You cannot change your backend wallet with active trades. Your funds locked in the order will not make it back to you.

chappjc (approved):
Looks good.  For a follow-up, I'll consider appropriate wording for a warning regarding changing the wallet to a different account/wallet with active swaps.

amassarwi:
@chappjc 

I've pushed a placeholder:
image

Let me know when you a have final version in mind.

JoeGruffins:
I think `might cause funds to be lost.` is better.

JoeGruffins:
I notice that I see the warning: `[WRN] CORE: [dcr]: Deposit address SsdiLuQuNnjhpb8ezzgqfKmyRFPW5hQknsP does not belong to connected wallet, generated new address: SsfzvqzSByoRwb1htP7VDK15F5XSK2zb6np` once more upon the first restart and login with a different generated address. I don't see it on subsequent logins. So, I'm guessing it isn't being saved to the database when you first change it, but is upon login. Not a huge issue though...

amassarwi:
@JoeGruffins Shouldn't regenerate on login, will double check

amassarwi:
@JoeGruffins Updated in https://github.com/decred/dcrdex/pull/902/commits/67a53f7e3450bcf3ac9a41332f3f2f6c3bcdd915:

Needed to write `dbWallet` to db before calling `connectWallet`.   Before my last commit `connectWallet` would generate new address and write it to db but when it returns to `ReconfigureWallet` the `dbWallet` was saved to the db again overriding the new address with old address which caused another address generation on next login.

chappjc (changes requested):

client/core/core.go
```diff
@@ -1517,32 +1557,24 @@ func (c *Core) ReconfigureWallet(appPW []byte, assetID uint32, cfg map[string]st
...
 		return newError(walletErr, "error loading wallet for %d -> %s: %v", assetID, unbip(assetID), err)
 	}
+	err = c.db.UpdateWallet(dbWallet)
+	if err != nil {
+		wallet.Disconnect()
+		return newError(dbErr, "error saving wallet configuration: %v", err)
+	}
+
 	// Must connect to ensure settings are good.
 	err = c.connectWallet(wallet)
```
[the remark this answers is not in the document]
amassarwi: Valid concerns

chappjc:
In fixing a wallet lock issue, I did some balance update work that adds the balance update in `ReconfigureWallet`. https://github.com/decred/dcrdex/pull/910/files#diff-13a4a6b12c1ae844ef90151638f194427cfdc545fceb58d52c0f6f33b81d859bR1556
However, I'll try to have this PR go in first to avoid making a conflict resolution job for you.

amassarwi:
@chappjc great, will about better approach regarding https://github.com/decred/dcrdex/pull/902#discussion_r550345812

client/core/core.go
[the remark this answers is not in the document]
amassarwi: Updated to follow the solution I mentioned on matrix:

> chappjc: trying to figure out proper way to resolve https://github.com/decred/dcrdex/pull/902#pullrequestreview-560219940 - one way would be returning a bool from `connectWallet` something like: `generatedNewAddress` true/false.
then i could check that flag in `ReconfigureWallet` after calling connectWallet - if true I can store xcWallet's (new) address as dbWallet.Address before updating the dbwallet in `ReconfigureWallet`

amassarwi:
@chappjc updated according to https://github.com/decred/dcrdex/pull/902#discussion_r550827492

PR #903, opened 2020-12-26
server: Add order id to cancel route error message

### LINKED ISSUE #810: Add order ID to failed cancel error message
`2020-11-04 03:13:14.076 [DBG] AUTH: Handling of 'cancel' request for user <hex> failed: error code 33: target order not known`

please add the target order id to this message.

amassarwi:
This adds order id to returned error message when the target is not known.

Closes #810

### DIFF
--- a/server/market/orderrouter.go
+++ b/server/market/orderrouter.go
@@ -627,7 +627,7 @@ func (r *OrderRouter) handleCancel(user account.AccountID, msg *msgjson.Message)
 	copy(targetID[:], cancel.TargetID)
 
 	if !tunnel.Cancelable(targetID) {
-		return msgjson.NewError(msgjson.OrderParameterError, "target order not known")
+		return msgjson.NewError(msgjson.OrderParameterError, "target order %v not known", targetID)
 	}
 
 	// Check that OrderType is set correctly
--- a/server/market/orderrouter.go
+++ b/server/market/orderrouter.go
@@ -617,7 +617,7 @@ func (r *OrderRouter) handleCancel(user account.AccountID, msg *msgjson.Message)
 
 	if !tunnel.Running() {
 		mktName, _ := dex.MarketName(cancel.Base, cancel.Quote)
-		return msgjson.NewError(msgjson.MarketNotRunningError, "market %s closed to new orders", mktName)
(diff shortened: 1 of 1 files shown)


server/market/orderrouter.go
```diff
@@ -627,7 +627,7 @@ func (r *OrderRouter) handleCancel(user account.AccountID, msg *msgjson.Message)
 	copy(targetID[:], cancel.TargetID)
 
 	if !tunnel.Cancelable(targetID) {
-		return msgjson.NewError(msgjson.OrderParameterError, "target order not known")
```
chappjc, line 630: For debugging and grepping, I can say from experience that it's often frustrating to have an identifier in the middle.  While I can use a regexp instead, not all users seeking support will be able or willing to do that easily (they'd have grep for "target order" or "not known", which could each match more than just this message). That is, I would tend to write this like `"target order not known: %v"` so that you can do a simple glob grep for `"target order not known"`.  That said, we don't want to mangle strings esp. sentences this way where it hurts readability, but in this case it's no less clear with the identifier at the end.

chappjc: But more fundamental to this change is that you are handling it servers side rather than client side.  The reason the target order ID is not in the server's error message already is because the client knows what order they were targeting.  Thus, when the client is logging the error, it should supplement the server's message with the target order ID instead, unless there is good reason why that can't be done.

chappjc (changes requested):

server/market/orderrouter.go
amassarwi: @chappjc 
Hrmm I thought it's server error message as the original message mentioned in #810 found here:
https://github.com/decred/dcrdex/blob/master/server/auth/auth.go#L500 and as I don't have the order id on that level I  adjusted the message from `orderrouter.go` side.

chappjc: Ahhh, the issue pertains to the server message. Nevermind then.

amassarwi: All good, will update the message structure.

JoeGruffins (approved):

chappjc (approved):

PR #906, opened 2020-12-28
client/asset/dcr: do not include fee address in sendRegFee

chappjc:
The Confirm Registration dialog error was showing the address it was trying to send to even though that address is of no use at best, and resulted in users manually sending funds there (e.g. dcrctl sendtoaddress) when that would not complete the registration.

image

This PR simply removes the "to address %s" part of the error message to avoid this pitfall.


chappjc:
Will sneak this in 0.1.4 if it looks good.

JoeGruffins (approved):

PR #543, opened 2020-07-13
testing/dcrdex: allow setting --pgpass.

dnldd:
I needed to set the password for the user to get things working for PostgreSQL12.2. Would get `db.Open: pq: password authentication failed for user "dcrdex"` if the user had no password set. This updates the harness to allow setting the pg password through dcrdex's config.

PR #907, opened 2020-12-30
dex/testing: harness improvements

itswisdomagain:
Modify dcrdex harness to receive custom epoch duration value from env variable rather than cli arg to retain support for relaying all cli args to dcrdex.

Update dcrdex harness readme to mention the supported `PG_PASS` and `EPOCH` env variables and the behaviour of passing cli args verbatim to dcrdex.

### DIFF
--- a/dex/testing/dcr/harness.sh
+++ b/dex/testing/dcr/harness.sh
@@ -277,6 +277,11 @@ if [ "$MINE" = "1" ]; then
   done
 fi
 
+# Hack to enable further mining on alpha or beta
+tmux send-keys -t $SESSION:0 "./mine-beta 1${WAIT}" C-m\; wait-for donedcr
+tmux send-keys -t $SESSION:0 "./beta addnode 127.0.0.1:${ALPHA_NODE_PORT} remove${WAIT}" C-m\; wait-for donedcr
+tmux send-keys -t $SESSION:0 "./beta addnode 127.0.0.1:${ALPHA_NODE_PORT} add${WAIT}" C-m\; wait-for donedcr
+
 # Have alpha share a little more wealth, esp. for trade_simnet_test.go
 RECIPIENTS="{\"${TRADING_WALLET1_ADDRESS}\":24,\"${TRADING_WALLET2_ADDRESS}\":24,\"${BETA_MINING_ADDR}\":24}"
 for i in {1..60}; do
--- a/dex/testing/dcr/harness.sh
+++ b/dex/testing/dcr/harness.sh
@@ -277,11 +277,6 @@ if [ "$MINE" = "1" ]; then
   done
 fi
 
-# Hack to enable further mining on alpha or beta
-tmux send-keys -t $SESSION:0 "./mine-beta 1${WAIT}" C-m\; wait-for donedcr
-tmux send-keys -t $SESSION:0 "./beta addnode 127.0.0.1:${ALPHA_NODE_PORT} remove${WAIT}" C-m\; wait-for donedcr
-tmux send-keys -t $SESSION:0 "./beta addnode 127.0.0.1:${ALPHA_NODE_PORT} add${WAIT}" C-m\; wait-for donedcr
-
 # Have alpha share a little more wealth, esp. for trade_simnet_test.go
 RECIPIENTS="{\"${TRADING_WALLET1_ADDRESS}\":24,\"${TRADING_WALLET2_ADDRESS}\":24,\"${BETA_MINING_ADDR}\":24}"
 for i in {1..60}; do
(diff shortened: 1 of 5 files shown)


dex/testing/dcr/harness.sh
```diff
@@ -277,6 +277,11 @@ if [ "$MINE" = "1" ]; then
   done
 fi
 
+# Hack to enable further mining on alpha or beta
+tmux send-keys -t $SESSION:0 "./mine-beta 1${WAIT}" C-m\; wait-for donedcr
+tmux send-keys -t $SESSION:0 "./beta addnode 127.0.0.1:${ALPHA_NODE_PORT} remove${WAIT}" C-m\; wait-for donedcr
+tmux send-keys -t $SESSION:0 "./beta addnode 127.0.0.1:${ALPHA_NODE_PORT} add${WAIT}" C-m\; wait-for donedcr
```
chappjc, line 283: This workaround should not be needed with https://github.com/decred/dcrd/pull/2513 merged in dcrd.  Can you verify that you still require this workaround?  If so, we should investigate the issue and see if there's a new bug.

itswisdomagain: Right. I was gonna highlight this after submitting the PR to confirm if it's a known issue with a known fix or workaround but it skipped my mind last minute.

I've updated dcrd to master (commit `8c289356a7e3ef0106627818601cab8acea04e49`) and there seem to be a new issue. Occurs after generating 2 addresses for each wallet; when trying to send 24 DCR to the wallets:
https://github.com/decred/dcrdex/blob/14f6b927d7c162333c84ef5ca588eaebf2f564bc/dex/testing/dcr/harness.sh#L281-L284

#### Screenshot
Screenshot 2020-12-30 at 18 50 10

Will investigate further later but dropping this here in case anyone else had encountered this before or knows off the bat what the issue is.

itswisdomagain: Works fine with commit a17427395603bc58cb461a7555a125d65a2e3334 (i.e. https://github.com/decred/dcrd/pull/2513). The issue I noted above is with latest dcrd/master which has its rpc version bumped to 7.0.0. while dcrwallet/master requires 6.0.0.

This workaround is not needed. Removing.

dex/testing/dcrdex/README.md
```diff
@@ -22,6 +22,13 @@ arguments when executing the harness script e.g.:
 ./harness.sh --pgpass=dexpass
```
buck54321, line 22: I'm having trouble seeing how this incantation actually does anything. 
`./harness.sh --pgpass=dexpass`
Where are these arguments being parsed into `PG_PASS` or otherwise appended to dcrdex.conf?

itswisdomagain: https://github.com/decred/dcrdex/blob/f2f78d32a833516f8ae0a79d3f31ad25df93e2a0/dex/testing/dcrdex/harness.sh#L162

See also https://github.com/decred/dcrdex/pull/543

chappjc: I think this is a problem then.  The command that gets executed is nonsensical:

```sh
dcrdex --appdata=/home/jon/dextest/dcrdex 2000 --pgpass=dexpass; tmux wait-for -S donedex
```

jessevdk/go-flags silently ignore the "2000" there, but I think we should address this.

buck54321: I certainly carry some responsibility for this confusion. When I added the custom bcast timeout, I had forgotten you could even pass other arguments. 

@itswisdomagain, if you'd like to just remove the custom bcast timeout, feel free. I can revisit it again next time I need it.

itswisdomagain: Replaced cli arg for custom epoch duration with `EPOCH` env variable.

dex/testing/dcrdex/harness.sh
```diff
@@ -22,12 +22,12 @@ sudo -u postgres -H psql -c "DROP DATABASE IF EXISTS ${TEST_DB}" \
 -c "CREATE DATABASE ${TEST_DB} OWNER dcrdex"
 
 EPOCH_DURATION=15000
-if [ "$#" -eq 1 ]; then
-    if [ "$1" -lt 1000 ]; then
+if [ -n "${EPOCH}" ]; then
```
chappjc, line 25: With the new env variable approach, I think it would be ok to use shell parameter expansion instead of a test.

`EPOCH_DURATION=${EPOCH:-1500}` or a variation with just one variable using `${EPOCH_DURATION:=val}`

itswisdomagain: oh, nice,
`EPOCH_DURATION` seem too long to type in terminal, will go with the former

buck54321 (approved):

JoeGruffins (approved):

Issue #417, opened 2020-05-27
dexc,core: shutdown refreshUser error


chappjc:
On shutdown:

> 
> [INF] WEB: Web server off
> [DBG] COMMS: Sending close 1000 (normal) message.
> [TRC] WEB: Disconnected websocket client 127.0.0.1
> [INF] CORE: DEX client core off
> [WRN] CORE: notify: |WARNING| (conn) DEX disconnected - DEX at 127.0.0.1:7232 has disconnected
> **[ERR] CORE: refreshUser: error checking if app is initialized: database not open**
> [INF] RPC: RPC server off

PR #913, opened 2021-01-05
start db first and stop it last

chappjc:
This resolves the following shutdown error message that we've been ignoring with remarkable conviction since forever:

```
[INF] CORE: DEX client core off
[ERR] CORE: refreshUser: error checking if app is initialized: database not open
[INF] DEXC: Exiting dexc main.
```

The error comes from a `refreshUser` called from `(*Core).handleConnectEvent` via a `dexConnection's` `ConnectEventFunc` (see `(*Core).connectDEX`), which fires on disconnect too, such as on shutdown.

Resolves https://github.com/decred/dcrdex/issues/417

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -905,17 +905,27 @@ func (c *Core) Run(ctx context.Context) {
 	c.ctx = ctx
 	c.initialize()
 	close(c.ready)
-	c.wg.Add(1)
+
+	// The DB starts first and stops last.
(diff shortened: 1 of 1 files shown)


buck54321 (approved):
I haven't been seeing this error myself, but this looks good anyway.

JoeGruffins (approved):
I actually never see that ERR but it must be a race and I'm just lucky.

PR #914, opened 2021-01-06
update badges

chappjc:
"high priority" was humor, but thanks for the prompt review. :)


buck54321 (approved):
Seems good to me

image

PR #915, opened 2021-01-07
[0.1.4] backport wallet lock state fix

chappjc:
This is a backport to 0.1.4 of the wallet lock commits in https://github.com/decred/dcrdex/pull/910 that resolve issues with the internal wallet lock state that could occur when starting dexc with both an unlocked wallet and active orders or swaps.

This is the last planned change for 0.1.4.


chappjc:
Given these are in master and this is the last planned 0.1.4 commit, merging.  We should do more testing of `release-0.1` asap so we can tag and be ready for next 1.6 rc.

PR #753, opened 2020-10-16
client/webserver: /api paths didn't enforce auth/login appropriately

chappjc:

PR #918, opened 2021-01-12
app: require auth for user and defaultwalletcfg API endpoints

buck54321:
re: https://github.com/decred/dcrdex/pull/753#discussion_r507247831

Moves the **/user** and **/defaultwalletcfg** endpoints behind the user auth middleware.

### DIFF
--- a/client/webserver/site/src/js/app.js
+++ b/client/webserver/site/src/js/app.js
@@ -32,11 +32,17 @@ const constructors = {
   order: OrderPage
 }
 
+// unathedPages are pages that don't require authorization to load.
+// These are endpoints outside of the requireLogin block in webserver.New.
+const unauthedPages = ['register', 'login', 'settings']
+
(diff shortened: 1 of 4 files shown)


client/webserver/site/src/js/app.js
```diff
@@ -105,6 +107,8 @@ export default class Application {
     // associated with  one of the available constructors.
     this.main = idel(document, 'main')
     const handler = this.main.dataset.handler
+    // Don't fetch the user until we know what page we're on.
+    await this.fetchUser()
```
chappjc, line 111: All seems to work well, although I want to confirm what's expected to happen here.  Before login or init, it is expected that a GET to /user will be a 401 unauthorized, correct?

chappjc (approved):

client/webserver/site/src/js/app.js
buck54321: That is my intention.

PR #919, opened 2021-01-12
client/core: cancel matches should not be loaded into matches map

chappjc:
When active orders are loaded on login, their matches are also loaded and put into the `trackedTrade.matches` map. However, cancel order matches were incorrectly being put into the matches map, which is only for negotiating matches (matches with other trades).  This catches these cancel order matches. They are still loaded into the `trackedTrade.cancel` field via `LinkedOrder` => `cancelTrade`.

The bug can be reproduced by creating a 2 lot standing limit order, partially filling it but NOT fully settling the swap, then cancelling the partially filled order, and once the cancel order is matched on epoch close and the order is actually cancelled (not just cancelling) restart dexc.  The shutdown will require an override at the prompt because of the actively settling match.  On startup and login, the cancel match was being added to the matches, so you will see a missing matches message during resolution followed by self-revoking the cancel match.

The second commit addresses a related pitfall of calling `Login` more than once. We should reevaluate the function of `Login` so that it does not repeat so much if `Core` has already successfully logged in from say another browser or dexcctl, but adding a short path right now is a bit risky as there are possible wallet related actions that happen on Login that users may need to resolve certain issues without restarting dexc.  So in this PR we just ensure not to overwrite existing `trackedTrade` entries in a `dexConnection.trades` map.


buck54321 (approved):

chappjc:
Rebased and squashed in a log statement, otherwise unchanged after review.

PR #921, opened 2021-01-14
client/core: avoid potential wallet reconfigure deadlock

chappjc:
This prevents a deadlock on wallet reconfigure that would happen if the client asset backend's tip change handler (a Core method) attempted to acquire the Core.walletMtx because it was already locked in ReconfigureWallet.  The deadlock was on `oldWallet.Disconnect` inside `ReconfigureWallet` because it would be waiting for the block monitor goroutine to return that was likely to itself be waiting on `Core`'s `walletMtx` (e.g. monitorBlocks>checkForNewBlocks>tipChange).

I hit the deadlock in testing release-0.1 today.

The `tipChange` goroutine that tried to re-lock `walletMtx` inside `refreshUser` via `updateBalances`:

```
goroutine 15952 [semacquire, 5 minutes]:
sync.runtime_SemacquireMutex(0xc0002b1ef4, 0x0, 0x0)
	/home/jon/go115/src/runtime/sema.go:71 +0x47
sync.(*RWMutex).RLock(0xc0002b1ee8)
	/home/jon/go115/src/sync/rwmutex.go:50 +0xa6
decred.org/dcrdex/client/core.(*Core).SupportedAssets(0xc0002b1e40, 0x0)
	/home/jon/github/decred/dcrdex/client/core/core.go:1269 +0xc8
decred.org/dcrdex/client/core.(*Core).refreshUser(0xc0002b1e40)
	/home/jon/github/decred/dcrdex/client/core/core.go:1302 +0xbc
decred.org/dcrdex/client/core.(*Core).updateBalances(0xc0002b1e40, 0xc000fbda68)
	/home/jon/github/decred/dcrdex/client/core/core.go:1244 +0x3ce
decred.org/dcrdex/client/core.(*Core).tipChange(0xc0002b1e40, 0xc00000002a, 0x0, 0x0)
	/home/jon/github/decred/dcrdex/client/core/core.go:4399 +0x545
decred.org/dcrdex/client/core.(*Core).loadWallet.func1(0x0, 0x0)
	/home/jon/github/decred/dcrdex/client/core/core.go:1433 +0x5d
decred.org/dcrdex/client/asset/dcr.(*ExchangeWallet).checkForNewBlocks(0xc004c9a240)
	/home/jon/github/decred/dcrdex/client/asset/dcr/dcr.go:2341 +0x476
decred.org/dcrdex/client/asset/dcr.(*ExchangeWallet).monitorBlocks(0xc004c9a240, 0x1357da0, 0xc0012482c0)
	/home/jon/github/decred/dcrdex/client/asset/dcr/dcr.go:2309 +0xbb
decred.org/dcrdex/client/asset/dcr.(*ExchangeWallet).Connect.func1(0xc000500550, 0xc004c9a240, 0x1357da0, 0xc0012482c0)
	/home/jon/github/decred/dcrdex/client/asset/dcr/dcr.go:529 +0x8a
created by decred.org/dcrdex/client/asset/dcr.(*ExchangeWallet).Connect
	/home/jon/github/decred/dcrdex/client/asset/dcr/dcr.go:527 +0xdfa
```

The `ReconfigureWallet` goroutine that was holding `walletMtx` and then called `oldWallet.Disconnect`:

```
goroutine 16881 [semacquire, 5 minutes]:
sync.runtime_Semacquire(0xc000500558)
	/home/jon/go115/src/runtime/sema.go:56 +0x45
sync.(*WaitGroup).Wait(0xc000500550)
	/home/jon/go115/src/sync/waitgroup.go:130 +0xe5
decred.org/dcrdex/dex.(*ConnectionMaster).Disconnect(0xc0002422d0)
	/home/jon/github/decred/dcrdex/dex/runner.go:113 +0x89
decred.org/dcrdex/client/core.(*xcWallet).Disconnect(0xc0002743c0)
	/home/jon/github/decred/dcrdex/client/core/wallet.go:211 +0x57
decred.org/dcrdex/client/core.(*Core).ReconfigureWallet(0xc0002b1e40, 0xc00024957d, 0x1, 0x1, 0xc00000002a, 0xc004ed13b0, 0x0, 0x0)
	/home/jon/github/decred/dcrdex/client/core/core.go:1629 +0x1707
decred.org/dcrdex/client/webserver.(*WebServer).apiReconfig(0xc000900000, 0x7faee57ef560, 0xc001248180, 0xc0008b5f00)
	/home/jon/github/decred/dcrdex/client/webserver/api.go:437 +0x242
```

`Disconnect` was deadlocked because `ExchangeWallet.Connect` was waiting forever for its `monitorBlocks` goroutine.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1452,7 +1452,7 @@ func (c *Core) loadWallet(dbWallet *db.Wallet) (*xcWallet, error) {
 	walletCfg := &asset.WalletConfig{
 		Settings: dbWallet.Settings,
 		TipChange: func(err error) {
-			go c.tipChange(assetID, err)
+			c.tipChange(assetID, err)
 		},
 	}
 	logger := c.log.SubLogger(unbip(assetID))
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1452,7 +1452,11 @@ func (c *Core) loadWallet(dbWallet *db.Wallet) (*xcWallet, error) {
 	walletCfg := &asset.WalletConfig{
 		Settings: dbWallet.Settings,
 		TipChange: func(err error) {
-			c.tipChange(assetID, err)
+			// asset.Wallet implementations should not need wait for the
+			// callback, as they don't know what it is, and will likely launch
+			// TipChange as a goroutine. However, guard against the possibility
+			// of deadlocking a Core method that calls Wallet.Disconnect.
+			go c.tipChange(assetID, err)
 		},
 	}
 	logger := c.log.SubLogger(unbip(assetID))
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1452,7 +1452,7 @@ func (c *Core) loadWallet(dbWallet *db.Wallet) (*xcWallet, error) {
 	walletCfg := &asset.WalletConfig{
 		Settings: dbWallet.Settings,
 		TipChange: func(err error) {
-			c.tipChange(assetID, err)
+			go c.tipChange(assetID, err)
 		},
 	}
 	logger := c.log.SubLogger(unbip(assetID))
@@ -1573,15 +1573,14 @@ func (c *Core) WalletSettings(assetID uint32) (map[string]string, error) {
 }
 
 // ReconfigureWallet updates the wallet configuration settings, it also updates
-// the password if newWalletPW is non-nil.
+// the password if newWalletPW is non-nil. Do not make concurrent calls to
+// ReconfigureWallet for the same asset.
 func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg map[string]string) error {
 	crypter, err := c.encryptionKey(appPW)
 	if err != nil {
 		return newError(authErr, "ReconfigureWallet password error: %v", err)
 	}
-	c.walletMtx.Lock()
-	defer c.walletMtx.Unlock()
-	oldWallet, found := c.wallets[assetID]
+	oldWallet, found := c.wallet(assetID)
 	if !found {
 		return newError(missingWalletErr, "%d -> %s wallet not found",
 			assetID, unbip(assetID))
@@ -1660,10 +1659,11 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg 
 	}
 	c.connMtx.RUnlock()
 
-	if oldWallet.connected() {
-		oldWallet.Disconnect()
-	}
+	c.walletMtx.Lock()
 	c.wallets[assetID] = wallet
+	c.walletMtx.Unlock()
+
+	go oldWallet.Disconnect()
 
 	c.notify(newBalanceNote(assetID, balances)) // redundant with wallet config note?
 	details := fmt.Sprintf("Configuration for %s wallet has been updated. Deposit address = %s",
@@ -4558,16 +4558,22 @@ func (c *Core) tipChange(assetID uint32, nodeErr error) {
 		}(id, waiter)
 	}
 	c.waiterMtx.Unlock()
+
 	c.connMtx.RLock()
-	assets := make(assetMap)
+	conns := make([]*dexConnection, 0, len(c.conns))
 	for _, dc := range c.conns {
+		conns = append(conns, dc)
+	}
+	c.connMtx.RUnlock()
+	assets := make(assetMap)
+	for _, dc := range conns {
 		newUpdates := c.tickAsset(dc, assetID)
 		if len(newUpdates) > 0 {
 			dc.refreshMarkets()
 			assets.merge(newUpdates)
 		}
 	}
-	c.connMtx.RUnlock()
+
 	// Ensure we always at least update this asset's balance regardless of trade
 	// status changes.
 	assets.count(assetID)
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1590,7 +1590,7 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg 
 		Settings:    cfg,
 		Balance:     &db.Balance{}, // in case retrieving new balance after connect fails
 		EncryptedPW: oldWallet.encPW,
-		Address:     oldWallet.address,
+		Address:     oldWallet.currentDepositAddress(),
 	}
 	// Reload the wallet with the new settings.
 	wallet, err := c.loadWallet(dbWallet)
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -492,18 +492,13 @@ func newWallet(cfg *BTCCloneCFG, btcCfg *dexbtc.Config, node rpcClient) (*Exchan
 	}
 	cfg.Logger.Tracef("Redeem conf target set to %d blocks", redeemConfTarget)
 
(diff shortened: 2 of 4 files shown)


client/core/core.go
```diff
@@ -1573,15 +1573,14 @@ func (c *Core) WalletSettings(assetID uint32) (map[string]string, error) {
...
+// ReconfigureWallet for the same asset.
 func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg map[string]string) error {
 	crypter, err := c.encryptionKey(appPW)
 	if err != nil {
 		return newError(authErr, "ReconfigureWallet password error: %v", err)
 	}
-	c.walletMtx.Lock()
-	defer c.walletMtx.Unlock()
-	oldWallet, found := c.wallets[assetID]
+	oldWallet, found := c.wallet(assetID)
```
chappjc, line 1583: I did not see a solid reason for the function to guard the wallets map from access to update, other than to ensure proper results if ReconfigureWallet were called for the same asset several times concurrently.  Resolved with docs.

client/core/core.go
```diff
@@ -1452,7 +1452,7 @@ func (c *Core) loadWallet(dbWallet *db.Wallet) (*xcWallet, error) {
 	walletCfg := &asset.WalletConfig{
 		Settings: dbWallet.Settings,
 		TipChange: func(err error) {
-			c.tipChange(assetID, err)
+			go c.tipChange(assetID, err)
```
chappjc, line 1455: The asset backend should not need this callback to be anything in particular let alone synchronous.

client/core/core.go
```diff
@@ -1660,10 +1659,11 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg
...
 	c.connMtx.RUnlock()
 
-	if oldWallet.connected() {
-		oldWallet.Disconnect()
-	}
+	c.walletMtx.Lock()
 	c.wallets[assetID] = wallet
+	c.walletMtx.Unlock()
+
+	go oldWallet.Disconnect()
```
chappjc, line 1666: Even with the deadlock resolved by not holding the walletMtx at this point, this can be done lazily.

client/core/core.go
```diff
@@ -4558,16 +4558,22 @@ func (c *Core) tipChange(assetID uint32, nodeErr error) {
...
+
 	c.connMtx.RLock()
-	assets := make(assetMap)
+	conns := make([]*dexConnection, 0, len(c.conns))
 	for _, dc := range c.conns {
+		conns = append(conns, dc)
+	}
+	c.connMtx.RUnlock()
+	assets := make(assetMap)
+	for _, dc := range conns {
```
chappjc, line 4569: Another case where I don't believe there is good reason why the connMtx needed to be held during tickAsset.  The map access just needs to be safe protected from concurrent writes, but I don't think it needs to be guarded during the trade ticks.

client/core/core.go
buck54321: > other than to ensure proper results if ReconfigureWallet were called for the same asset several times concurrently

Yes, that was the reasononing.

client/core/core.go
buck54321: I see that in **runner.go**, we have

```
func (c *ConnectionMaster) Disconnect() {
	c.cancel()
	c.mtx.RLock()
	c.wg.Wait()
	c.mtx.RUnlock()
}
```

Should that be 

```
func (c *ConnectionMaster) Disconnect() {
	c.cancel()
	c.mtx.RLock()
	wg := c.wg
	c.mtx.RUnlock()
	wg.Wait()
}
```

so that we wouldn't have to do this?

chappjc: It would still be stuck at `wg.Wait()`.  The ConnectionMaster's mutex isn't under contention.

client/core/core.go
itswisdomagain: Should we let the asset backend decide? Call `go {dcr/btc}.tipChange` so it's more immediately obvious that the asset backend is not relying on the configured callback to finish processing before proceeding.

client/core/core.go
itswisdomagain: How about if an attempt is made to access this wallet while this reconfiguration is going on? I do understand not preventing access to all wallets in order to reconfigure one but I do think the wallet being reconfigured should be locked for access.

chappjc: Problem is there are already numerous places that just grab the xcWallet with the `wallet()` method and move on to using it, so if one of those started and then ReconfigureWallet happened, even keeping this locked for the duration won't prevent the other goroutine from getting their wallet disconnected unexpectedly.

client/core/core.go
chappjc: Good points.  I had gone back and forth on this actually.  You make a good case, but my concern is that we then rely on the asset author to avoid deadlocking Core.  I'm not decided yet.

client/core/core.go
itswisdomagain: oh, noted

client/core/core.go
itswisdomagain: Avoiding a deadlock in Core wouldn't be the main goal though, just a nice side effect. The main goal I see is making `monitorBlocks/checkForNewBlocks` faster, not waiting on the opaque execution of a callback, which could take however long _and_ which we really don't need to wait for.

chappjc: I agree with you.  There are also a million other ways the backend can make Disconnect hang indefinitely, so it's up to the consumer (Core) to deal.

client/core/core.go
chappjc: I've been starting to suspect recently that we need two types of reconfigure: (1) one that affects actually wallet connection like RPC params, and (2) other random options that can just be applied to the existing wallet backend with target mutator methods.

client/core/core.go
chappjc: I've resolved this by wrapping the provided callback in the constructor rather than putting `go` in front over every place `tipChange` is called.  My reasoning is that there is no case where it would need to be synchronous and the programmer might forget. Although it does make it less obvious at the call site that it is asynchronous, but it is completely opaque so I don't think that says much to the reader anyway.

client/core/core.go
buck54321: Only the asset developer really knows which settings require reconnect, so I'd recommend we add a `Reconfigure(cfg map[string]string)` method to `asset.Wallet` instead.

chappjc: Right, I didn't mean two different Core methods necessarily.

Although the ReconfigureWallet interface is unfortunately cumbersome for many settings changes so I'd not rule out convenience methods that avoid the need to get the current settings map only to feed it back with maybe one tweak to it.  A different idea entirely though.

JoeGruffins (approved):

itswisdomagain (approved):

buck54321 (approved):
g2g, but I would still prefer to do the `go tipChange` in **core**. It's a requirement and every wallet implementation would need to do it anyway.

chappjc:
> g2g, but I would still prefer to do the `go tipChange` in **core**. It's a requirement and every wallet implementation would need to do it anyway.

Funny you should mention it because I'm swinging back in that direction too.  Did it in draft https://github.com/decred/dcrdex/pull/928 as I was having to call Disconnect there too.

PR #928, opened 2021-01-15
client: don't lock in PromptShutdown, do it on Core shutdown

chappjc:
Previously we gave `(*Core).PromptShutdown` the expanded responsibility of not just checking and prompting if a shutdown was OK but also locking the wallets.  This was not the right place to do that for several reasons:

1. `PromptShutdown` is just a check and prompt function, not a "partial shutdown" function.  Shutdown via context cancellation is the correct approach.
2. `Core` consumers could and were not using `PromptShutdown`, and instead just cancelling `Core`'s context thus leaving the wallets unlocked after shutdown, thus leading to the now-fixed "wallet not unlocked" bug (https://github.com/decred/dcrdex/pull/915) that was requiring such consumers to also restart their wallets when restarting.
3. Locking the wallets and the dex accounts before shutting down `(*Core).Run` goroutine was incorrect. After the DB and `dexConnection`s stop, the wallets may be locked and dex account private keys cleared.

<s>A secondary bug resolved here is that while the prompt was shown (`"You have active orders. Shutting down now may result in..."`) the `connMtx` was held, and blocked just about all critical `Core` functions.</s> EDIT: independently resolved in https://github.com/decred/dcrdex/commit/c75c5797a3b62a1e9197361791a590dcb4931034

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1670,9 +1670,7 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg 
 	c.wallets[assetID] = wallet
 	c.walletMtx.Unlock()
 
-	if oldWallet.connected() {
-		go oldWallet.Disconnect()
-	}
+	go oldWallet.Disconnect()
 
 	c.notify(newBalanceNote(assetID, balances)) // redundant with wallet config note?
 	details := fmt.Sprintf("Configuration for %s wallet has been updated. Deposit address = %s",
@@ -4563,8 +4561,6 @@ func (c *Core) tipChange(assetID uint32, nodeErr error) {
 // or if the user has confirmed they want to shutdown with active orders.
 func (c *Core) PromptShutdown() bool {
 	conns := c.dexConnections()
-
-
 	var haveActive bool
 	for _, dc := range conns {
 		if dc.hasActiveOrders() {
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -891,11 +891,11 @@ func New(cfg *Config) (*Core, error) {
 
 // Run runs the core. Satisfies the runner.Runner interface.
 func (c *Core) Run(ctx context.Context) {
-	c.log.Infof("started DEX client core")
+	c.log.Infof("Starting DEX client core")
 	// Store the context as a field, since we will need to spawn new DEX threads
 	// when new accounts are registered.
 	c.ctx = ctx
-	c.initialize()
+	c.initialize() // connectDEX gets ctx for the wsConn
 	close(c.ready)
 
 	// The DB starts first and stops last.
@@ -913,11 +913,35 @@ func (c *Core) Run(ctx context.Context) {
 		c.latencyQ.Run(ctx)
 	}()
 
-	c.wg.Wait()
+	c.wg.Wait() // block here until all goroutines except DB complete
 
 	// Stop the DB after dexConnections and other goroutines are done.
 	stopDB()
 	dbWG.Wait()
+
+	// Lock and disconnect the wallets.
+	c.walletMtx.Lock()
+	for assetID, wallet := range c.wallets {
+		symb := strings.ToUpper(unbip(assetID))
+		c.log.Infof("Locking %s wallet", symb)
+		if err := wallet.Lock(); err != nil {
+			c.log.Errorf("Failed to lock %v wallet: %v", symb, err)
+		}
+		wallet.Disconnect()
+		delete(c.wallets, assetID)
+	}
+	c.walletMtx.Unlock()
+
+	// Clear account private keys and wait for the DEX ws connections that began
+	// shutting down on context cancellation.
+	c.connMtx.Lock()
+	defer c.connMtx.Unlock()
+	for _, dc := range c.conns {
+		// context should be canceled allowing just a Wait(), but just in case
+		// use Disconnect otherwise this could hang forever.
+		dc.connMaster.Disconnect()
+		dc.acct.lock()
+	}
 
 	c.log.Infof("DEX client core off")
 }
@@ -1110,7 +1134,7 @@ func (c *Core) connectedWallet(assetID uint32) (*xcWallet, error) {
 // synching, this also starts a goroutine to monitor sync status, emitting
 // WalletStateNotes on each progress update.
 func (c *Core) connectWallet(w *xcWallet) (depositAddr string, err error) {
-	err = w.Connect(c.ctx) // ensures valid deposit address
+	err = w.Connect() // ensures valid deposit address
 	if err != nil {
 		return "", codedError(connectWalletErr, err)
 	}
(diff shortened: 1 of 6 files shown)


chappjc:
Oh I remember why the wallet locking was done before context cancellation - that shuts down the wallet connections too.  I now recall thinking the right place to lock the wallets is in the asset backend impl. when the context is canceled.

client/core/core.go
```diff
@@ -1646,9 +1670,7 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg
 	c.wallets[assetID] = wallet
 	c.walletMtx.Unlock()
 
-	if oldWallet.connected() {
-		go oldWallet.Disconnect()
-	}
+	go oldWallet.Disconnect()
```
chappjc, line 1673: `(*ConnectionMaster).Disconnect` is now safe to call in any state.  See the change in dex/runner.go

client/core/core.go
```diff
@@ -1110,7 +1134,7 @@ func (c *Core) connectedWallet(assetID uint32) (*xcWallet, error) {
 // synching, this also starts a goroutine to monitor sync status, emitting
 // WalletStateNotes on each progress update.
 func (c *Core) connectWallet(w *xcWallet) (depositAddr string, err error) {
-	err = w.Connect(c.ctx) // ensures valid deposit address
+	err = w.Connect() // ensures valid deposit address
```
chappjc, line 1137: `xcWallets` are now stopped via their `Disconnect` method, as done now in `(*Core).Run`.

dex/runner.go
```diff
@@ -108,15 +108,26 @@ func (c *ConnectionMaster) Connect(ctx context.Context) (err error) {
...
 }
 
-// Disconnect closes the connection and waits for shutdown.
+// Disconnect closes the connection and waits for shutdown. This is safe to use
+// on an unconnected ConnectionMaster as long as it was constructed with
+// NewConnectionMaster.
 func (c *ConnectionMaster) Disconnect() {
-	c.cancel()
 	c.mtx.RLock()
+	defer c.mtx.RUnlock()
```
chappjc, line 116: AFAICT, there's no contention on the `ConnectionMaster`'s (`contextManager`'s) mutex.  In fact, I'm not even sure it is needed in a proper use pattern, but there may be different uncoordinated goroutines that can each do `Connect` and `Disconnect` so I'm not just nuking it.

dex/testing/loadbot/mantle.go
```diff
@@ -110,7 +110,8 @@ out:
 		}
 	}
 
-	m.ForceShutdown()
+	// Let Core shutdown and lock up.
+	m.waiter.WaitForShutdown()
```
chappjc, line 114: I haven't retested loadbot with this, but `ForceShutdown` was the wallet and account locking part of `PromptPass` that is now done by `Core` shutdown.

itswisdomagain (changes requested):
First pass. Mostly questions.

dex/runner.go
chappjc: I think calling `Connect` twice is veering into the territory of improper use of the type's methods, but eliminating the mutex would be lovely.  Just like doing `Run(ctx)` more than once (at least with the first call still running) for just about all `Runners` is an implicitly prohibited pattern that applies here too I think.  I suppose we need to consider the design goals of `ConnectionMaster`, but I can see it going either way - single use or allow repeated connect/disconnect calls.

Ideally all of `ConnectionMaster` would be not have any synchronization mechanisms and the consumer would just be aware of its state... I'm starting to recall why I kept this in draft now.  Hmm, I might revert most of the last two commits and go back to the `if oldWallet.connected()` check since I'm clearly touching on a design issue with `ConnectionMaster` (and how `Core` manages them) that I don't know I want to solve in this bugfix PR.

dex/runner.go
```diff
@@ -108,15 +108,26 @@ func (c *ConnectionMaster) Connect(ctx context.Context) (err error) {
...
+		// The Connector failed on Connect, assigning a nil *WaitGroup.
+		return
+	}
 	c.wg.Wait()
-	c.mtx.RUnlock()
 }
 
-// Wait waits for the the WaitGroup returned by Connect.
+// Wait waits for the the WaitGroup returned by Connect. This is NOT safe to use
+// unless Connect has already been called without error.
```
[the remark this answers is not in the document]
chappjc: As per my other comment, I thought about that but was still up in the air about the type in general, so just documented current semantics.  I think I've talked myself into reverting the `ConnectionManager` changes and saving it for another day.  Will switch back everything in the last two commits except the fix to `(*xcWallet).Connect` for not leaving the wallet connected if `hookedUp` doesn't get set on account of a `SyncStatus` or other initial rpc failure.

client/core/core.go
```diff
@@ -913,12 +913,36 @@ func (c *Core) Run(ctx context.Context) {
...
 
+	// Lock and disconnect the wallets.
+	c.walletMtx.Lock()
+	for assetID, wallet := range c.wallets {
+		symb := strings.ToUpper(unbip(assetID))
+		c.log.Infof("Locking %s wallet", symb)
+		if err := wallet.Lock(); err != nil {
+			c.log.Errorf("Failed to lock %v wallet: %v", symb, err)
+		}
+		wallet.Disconnect()
```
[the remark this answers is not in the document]
chappjc: Have to consider cases like ReconfigureWallet where we don't want to lock before disconnect.  The sequence there is connect to new wallet, then disconnect from old, but the old one is very often the exact same wallet, just some different settings that are unrelated to RPC connection such as redeem conf target. So in that case you'd be locking the same wallet shared by both of the xcWallets.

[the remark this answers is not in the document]
chappjc: This was a good suggestion, but I decided to manually lock the wallet before disconnect in `CreateWallet` and `setWalletPassword` instead of making it part of the `Disconnect` method as it wasn't universally correct to lock before disconnect.  See https://github.com/decred/dcrdex/pull/928/commits/448be802856310c95ab3822e3e6c83353f2cbd45 just pushed.

client/core/core.go
```diff
@@ -1392,13 +1419,15 @@ func (c *Core) CreateWallet(appPW, walletPW []byte, form *WalletForm) error {
 	}
 
 	initErr := func(s string, a ...interface{}) error {
+		_ = wallet.Lock() // just try, but don't confuse the user with an error
```
JoeGruffins, line 1422: Debug log?

client/core/core.go
```diff
@@ -1647,6 +1676,8 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg
 	c.walletMtx.Unlock()
 
 	if oldWallet.connected() {
+		// NOTE: Cannot lock the wallet backend because it may be the same as
+		// the one just connected.
```
JoeGruffins, line 1680: Should really check if this is the same wallet and user has active trades. If both are yes error, can't use a different wallet with active trades.

We have the old wallet's deposit address, so maybe a simple `isMine` method on wallets would be good enough?

Also, after checking we will know if we should disconnect here or not.

chappjc: That's a great point about blocking the user from switching to a wallet that will break active orders or swaps! We should do that, but let's make an issue and another PR for it.  Scope is creeping on this PR.

Shutdown locking in the wrong place was causing consumers to hit a different bug in previous releases (the "not unlocked" one), and has encouraged the unfortunate habit of restarting wallets to avoid that now-fixed bug, which in turn caused another bug to manifest, which we also [just fixed](https://github.com/decred/dcrdex/pull/965).  😌 Phew.  Anyway, I want to close the book on this PromptShutdown design flaw as concisely as possible so it can backport into 0.1.5 with minimal trouble.

JoeGruffins (approved):

client/core/core.go
buck54321: Any reason to use the `dex.Connector` for `asset.Wallet` at all now?

buck54321 (approved):

client/core/core.go
chappjc: I suppose just for On and Wait.  Could revisit this.

PR #930, opened 2021-01-18
core: remove cached User struct

### LINKED ISSUE #379: core: need targeted updates to (*Core).user object

buck54321:
Resolves #379

I experimented with some different schemes for updating the cached `*User` struct and then creating copies on calls to `(*Core).User`, but ultimately decided that we don't really need to cache a `*User` at all, and the cleanest solution is to just get rid of it.

- remove `user` and `userMtx` fields from `Core`. `(*Core).User` now generates a new `*core.User` on demand.
- remove the `marketMap` field from `Core`. Anything that used the `core.Market` from the `marketMap` now uses the `msgjson.Market` from `cfg` instead.
- `dexConnections` and `xcWallets` methods to relieve some lock contention
- `*OrderLocked` and `*ContractLocked` fields changed to methods (breaking change)

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -120,10 +120,11 @@ func (dc *dexConnection) marketConfig(mktID string) *msgjson.Market {
 // [base]_[quote].
 func (dc *dexConnection) marketMap() map[string]*Market {
 	dc.cfgMtx.RLock()
-	defer dc.cfgMtx.RUnlock()
+	mktConfigs := dc.cfg.Markets
+	dc.cfgMtx.RUnlock()
 
-	marketMap := make(map[string]*Market, len(dc.cfg.Markets))
-	for _, mkt := range dc.cfg.Markets {
+	marketMap := make(map[string]*Market, len(mktConfigs))
+	for _, mkt := range mktConfigs {
 		// The presence of the asset for every market was already verified when the
 		// dexConnection was created in connectDEX.
 		dc.assetsMtx.RLock()
@@ -980,34 +981,49 @@ func (c *Core) Exchanges() map[string]*Exchange {
 	return c.exchangeMap()
 }
 
+// dexConnections creates a slice of the *dexConnection in c.conns.
+func (c *Core) dexConnections() []*dexConnection {
+	c.connMtx.RLock()
+	defer c.connMtx.RUnlock()
+	conns := make([]*dexConnection, 0, len(c.conns))
+	for _, conn := range c.conns {
+		conns = append(conns, conn)
+	}
+	return conns
+}
+
 // exchangeMap creates a map of *Exchange keyed by host, including markets and
 // orders.
 func (c *Core) exchangeMap() map[string]*Exchange {
-	c.connMtx.RLock()
-	defer c.connMtx.RUnlock()
 	infos := make(map[string]*Exchange, len(c.conns))
-	for host, dc := range c.conns {
+	for _, dc := range c.dexConnections() {
 		dc.cfgMtx.RLock()
-		dc.assetsMtx.RLock()
 		requiredConfs := uint32(dc.cfg.RegFeeConfirms)
+		dc.cfgMtx.RUnlock()
+
+		dc.assetsMtx.RLock()
+		assets := make(map[uint32]*dex.Asset, len(dc.assets))
+		for assetID, dexAsset := range dc.assets {
+			assets[assetID] = dexAsset
+		}
+		dc.assetsMtx.RUnlock()
+
 		// Set AcctID to empty string if not initialized.
 		acctID := dc.acct.ID().String()
 		var emptyAcctID account.AccountID
 		if dc.acct.ID() == emptyAcctID {
 			acctID = ""
 		}
-		infos[host] = &Exchange{
-			Host:          host,
+		infos[dc.acct.host] = &Exchange{
+			Host:          dc.acct.host,
 			AcctID:        acctID,
 			Markets:       dc.marketMap(),
-			Assets:        dc.assets,
+			Assets:        assets,
 			FeePending:    dc.acct.feePending(),
 			Connected:     dc.connected,
 			ConfsRequired: requiredConfs,
 			RegConfirms:   dc.getRegConfirms(),
 		}
-		dc.cfgMtx.RUnlock()
-		dc.assetsMtx.RUnlock()
 	}
 	return infos
 }
@@ -1210,9 +1226,7 @@ func (c *Core) updateWalletBalance(wallet *xcWallet) (*WalletBalance, error) {
 // swaps (orderLocked). Only applies to trades where the specified assetID is
 // the fromAssetID.
 func (c *Core) lockedAmounts(assetID uint32) (contractLocked, orderLocked uint64) {
-	c.connMtx.RLock()
-	defer c.connMtx.RUnlock()
-	for _, dc := range c.conns {
+	for _, dc := range c.dexConnections() {
 		dc.tradeMtx.RLock()
 		for _, tracker := range dc.trades {
 			tracker.mtx.RLock()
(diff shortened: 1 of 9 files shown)


client/core/core.go
```diff
@@ -112,10 +109,17 @@ func (dc *dexConnection) running(mkt string) bool {
...
+	defer dc.cfgMtx.RUnlock()
+	return dc.marketConfig(mktID)
+}
+
 // refreshMarkets rebuilds the market map. The updated markets map may be safely
-// accessed with the markets method. refreshMarkets is used when a change to the
-// status of a market or the user's orders on a market has changed.
-func (dc *dexConnection) refreshMarkets() {
+// accessed with the markets method. refreshMarkets should be called when the
+// market configuration changes.
```
buck54321, line 121: Whoops. Wrong docs.

client/core/types.go
```diff
@@ -360,19 +362,71 @@ func coreOrderFromTrade(ord order.Order, metaData *db.OrderMetaData) *Order {
...
-	QuoteID             uint32   `json:"quoteid"`
-	QuoteSymbol         string   `json:"quotesymbol"`
-	EpochLen            uint64   `json:"epochlen"`
-	StartEpoch          uint64   `json:"startepoch"`
-	MarketBuyBuffer     float64  `json:"buybuffer"`
-	Orders              []*Order `json:"orders"`
-	BaseOrderLocked     uint64   `json:"baseorderlocked"`
-	QuoteOrderLocked    uint64   `json:"quoteorderlocked"`
-	BaseContractLocked  uint64   `json:"basecontractlocked"`
-	QuoteContractLocked uint64   `json:"quotecontractlocked"`
```
buck54321, line 375: These fields have been replaced by methods. Breaking change, obviously.

client/core/core.go
```diff
@@ -995,34 +975,52 @@ func (c *Core) Network() dex.Network {
...
+
+// dexConnections creates a slice of the *dexConnection in c.conns.
+func (c *Core) dexConnections() []*dexConnection {
 	c.connMtx.RLock()
 	defer c.connMtx.RUnlock()
+	conns := make([]*dexConnection, 0, len(c.conns))
+	for _, conn := range c.conns {
+		conns = append(conns, conn)
+	}
+	return conns
```
buck54321, line 992: This was a pattern that @chappjc used in #921, and I wanted to use as well, and then realized that it really is better just about everywhere we were doing `range c.conns`. The same reasoning also applied to `c.wallets`, so I made the `xcWallets` method as well. Sorry for the scope expansion.

client/core/core.go
```diff
@@ -136,42 +141,28 @@ func (dc *dexConnection) refreshMarkets() {
...
-// markets returns the current market map.
-func (dc *dexConnection) markets() map[string]*Market {
-	dc.marketMtx.RLock()
-	defer dc.marketMtx.RUnlock()
-	return dc.marketMap
+// marketTrades is a slice of trades in the trades map.
+func (dc *dexConnection) marketTrades(mktID string) (trades []*trackedTrade) {
+	dc.tradeMtx.RLock()
+	defer dc.tradeMtx.RUnlock()
+	for _, trade := range dc.trades {
```
chappjc, line 160: Micro-optimization, but want to allocate `trades` here?

client/core/core.go
```diff
@@ -1309,39 +1320,15 @@ func (c *Core) SupportedAssets() map[uint32]*SupportedAsset {
...
 // User is a thread-safe getter for the User.
 func (c *Core) User() *User {
-	c.userMtx.RLock()
-	defer c.userMtx.RUnlock()
-	return c.user
-}
-
-// refreshUser is a thread-safe way to update the current User. This method
-// should be called after adding wallets and DEXes.
-func (c *Core) refreshUser() {
```
chappjc, line 1319: :heart: :heart: :heart: :heart:

client/core/trade.go
```diff
@@ -211,8 +211,14 @@ func (t *trackedTrade) coreOrder() *Order {
...
 	corder := coreOrderFromTrade(t.Order, t.metaData)
+
 	corder.Epoch = t.dc.marketEpoch(t.mktID, t.Prefix().ServerTime)
-	corder.LockedAmt = t.lockedAmount()
+
+	if t.coinsLocked {
+		corder.LockedAmt = sumCoinMap(t.coins)
+	} else if t.changeLocked {
+		corder.LockedAmt = t.change.Value()
+	}
```
chappjc, line 221: Since we're keeping `(*trackedTrade).lockedAmount` because it's used by `(*Core).lockedAmounts` why not keep using it here?

client/core/types.go
chappjc: Just thinking about the rpc server and other places where we marshal this before sending it, as opposed to Go consumers that can call the method... is their a net benefit from excluding the fields?

I'm looking at `sideContractLocked`, and because the computed value never changes once a `*Market` is generated, why not stick with fields? AFAICT, the `core.Order`s generated for this `*Market` are immutable, assuming it's only coming from the `(*dexConnection).marketMap` method.

client/core/types.go
```diff
@@ -360,19 +361,71 @@ func coreOrderFromTrade(ord order.Order, metaData *db.OrderMetaData) *Order {
...
+func (m *Market) QuoteContractLocked() uint64 {
+	return m.sideContractLocked(false)
+}
+
+func (m *Market) sideContractLocked(sell bool) (amt uint64) {
+	for _, ord := range m.Orders {
+		if ord.Sell != sell {
+			continue
+		}
+		for _, m := range ord.Matches {
```
chappjc, line 392: Could you change the loop variable so it does not shadow the method receiver?

chappjc (changes requested):
Huge fan of this PR.  A handful of minor suggestions and questions.

JoeGruffins (approved):

client/core/types.go
buck54321: I guess the counter-argument is that the data is already there and this sort of "summary field" is just a maintenance item. And while you're correct that we do not mutate the Go types returned through the `Core` frontend, in our javascript app that's exactly what we do to the decoded structs, so these fields are outdated with the first order or match update.

Based on that, I'm guessing a Go app which made ample use of this API might employ the same technique to keep everything synchronized as order/match updates come in. This is in fact the exact thing we were doing internally (until this PR) to avoid regenerating the entire `User` struct too much. But that optimization became unnecessary, mostly because the client was doing the struct updates anyway, so wasn't requesting the `User` as often. 

I guess my point is that even though we don't mutate these structs in Go, the caller may want to, in which case these fields could be confusing. These are the only fields we have whose accuracy relies on the immutability of the associated matches, so my preference is to remove them. 

That said, I do recognize that this leaves in the cold anyone who was consuming these via JSON conversion to another language (e.g. us). They would have to reimplement the methods themselves. If you think that's a strong enough reason to keep them, I can go back to fields.

chappjc: That all makes sense.  I'm not aware of the details of how Go consumers actually use the Users struct (do they really hold the whole `User` struct modify it in response to order notes or do they just extract `Orders`?), but my feeling is that the consumer should treat the struct as a snapshot and thus the locked amounts that derive from current order-match statuses would change as the matches change.  Yet I completely get your point of view that derived fields might as well be replaced with the functions that derive them.

But my initial comment was mainly considering someone doing `dexcctl exchanges` (or GET the analogous http api endpoint, with no interest in websocket notifications) to just get a snapshot of how much they have locked in these states, which would now require going through all the orders returned.

chappjc: Pinging some users real quick...

chappjc: OK, so the wallets RPC will still provide this info, along with `BalanceNote`s.  I'm fine with this change.

chappjc (approved):

PR #931, opened 2021-01-19
update go-chi to latest with compatible module

chappjc:
For reference: https://github.com/go-chi/chi/blob/master/CHANGELOG.md#v150-2020-11-12---now-with-gomod-support
In a nutshell, go-chi maintainers decided to convert the v4+incompatible module to a v1 module do avoid the /vX part of the import that it would necessitate if they made it a compatible v4 module.


chappjc:
Forcing this in to master now as it's just a module versioning thing with no actual actual API changes.  Plus I can't get dcrdata to use v1.5.1 unless dcrdex is using it because ofc v4.0.2 > v1.5.1.

PR #934, opened 2021-01-22
client: show correct error message on Add a Dex register form

### LINKED ISSUE #933: missing dex address validation on register page
I'm not fully familiar with dex structure, and I guess this is minor, but while setting it up, I tried to add a dex address and no visual feedbacks were displayed when I entered an invalid address.

IMO, we should at least display an error message.

add-a-dex-without-feedback

victorgcramos:
Fixes #933

Preview:
Screen Shot 2021-01-22 at 12 26 00 PM


buck54321 (approved):

chappjc (approved):
I'm 100% certain we'll have another js change before 0.2 is near release, so we'll just have to remember to bump the query in bodybuilder.tmpl down the line.

PR #633, opened 2020-08-24
client/asset/btc: sendrawtransaction syntax change

chappjc:
From node version 0.19, the `sendrawtransaction` RPC changed the
`allowHighFees` boolean argument to a `maxFeeRate` numeric argument.
btcd's rpclient detects which to use depending on the version obtained
from `getnetworkinfo`, however this detection is broken for nodes that
report a subversion that is not of the form "/Satoshi:0.20.1/". For
example, rpcclient fails to recognize "/LitecoinCore:0.20.1/" as
requiring the boolean argument and tries to send a numberic fee rate.

To fix this, implement `sendrawtransaction` with `rawrequest`.
The `(*ExchangeWallet).SendRawTransaction` method will use the correct
argument type according to the integer network version (e.g. 200100)
detected on `ExchangeWallet` construction.

Tested with:
- litecoind v0.18.1
- litecoind v0.20.0.0-649aea5fc from thrasher- branch
- bitcoind v0.20.1.0

PR #938, opened 2021-01-23
use decred's rpcclient for BTC

### LINKED ISSUE #841: use decred's rpcclient for BTC with all requests going through RawRequest
btcsuite's rpcclient does not support request cancellation, so we might consider using decred's rpcclient/v6 and have all requests use RawRequest.

_Originally posted by @chappjc in https://github.com/decred/dcrdex/pull/838#discussion_r526242776_

amassarwi:
This uses decred's `rpcclient/v6` for BTC wallet that all RPC calls are sent through `RawRequest`.

Closes #841

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -641,7 +641,7 @@ func (btc *ExchangeWallet) Balance() (*asset.Balance, error) {
 	if btc.useLegacyBalance {
 		return btc.legacyBalance()
 	}
-	balances, err := btc.wallet.Balances(btc.ctx)
+	balances, err := btc.wallet.Balances()
 	if err != nil {
 		return nil, err
 	}
@@ -2353,7 +2353,6 @@ type utxo struct {
 	vout    uint32
 	address string
 	amount  uint64
-	tree    int8
 }
 
 // Combines utxo info with the spending input information.
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -465,7 +465,9 @@ func BTCCloneWallet(cfg *BTCCloneCFG) (*ExchangeWallet, error) {
 
 	requester, err := newRequester(endpoint, btcCfg.RPCUser, btcCfg.RPCPass,
 		cfg.Symbol, cfg.Logger)
-
+	if err != nil {
+		return nil, err
+	}
 	btc, err := newWallet(requester, cfg, btcCfg)
 	if err != nil {
 		return nil, fmt.Errorf("error creating %s ExchangeWallet: %v", cfg.Symbol,
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -344,7 +344,6 @@ type ExchangeWallet struct {
 	// 64-bit atomic variables first. See
 	// https://golang.org/pkg/sync/atomic/#pkg-note-BUG
 	tipAtConnect      int64
-	node              *walletClient
 	wallet            *walletClient
 	walletInfo        *asset.WalletInfo
 	chainParams       *chaincfg.Params
@@ -505,11 +504,8 @@ func newWallet(requester RawRequester, cfg *BTCCloneCFG, btcCfg *dexbtc.Config) 
 	}
 	cfg.Logger.Tracef("Redeem conf target set to %d blocks", redeemConfTarget)
 
-	walletClient := newWalletClient(requester, cfg.Segwit, cfg.ChainParams)
-
 	return &ExchangeWallet{
-		node:                walletClient,
-		wallet:              walletClient,
+		wallet:              newWalletClient(requester, cfg.Segwit, cfg.ChainParams),
 		symbol:              cfg.Symbol,
 		chainParams:         cfg.ChainParams,
 		log:                 cfg.Logger,
@@ -550,7 +546,7 @@ func (btc *ExchangeWallet) Connect(ctx context.Context) (*sync.WaitGroup, error)
 		return nil, fmt.Errorf("node software out of date. version %d is less than minimum %d", codeVer, minProtocolVersion)
 	}
 	// Initialize the best block.
-	h, err := btc.node.GetBestBlockHash()
+	h, err := btc.wallet.GetBestBlockHash()
 	if err != nil {
 		return nil, fmt.Errorf("error initializing best block for %s: %w", btc.symbol, err)
 	}
@@ -558,7 +554,7 @@ func (btc *ExchangeWallet) Connect(ctx context.Context) (*sync.WaitGroup, error)
 	btc.currentTip, err = btc.blockFromHash(h.String())
 	btc.tipMtx.Unlock()
 	if err != nil {
-		return nil, fmt.Errorf("error initializing best block for %s: %w", btc.symbol, err)
+		return nil, fmt.Errorf("error parsing best block for %s: %w", btc.symbol, err)
 	}
 	atomic.StoreInt64(&btc.tipAtConnect, btc.currentTip.height)
 	var wg sync.WaitGroup
@@ -674,7 +670,7 @@ func (btc *ExchangeWallet) legacyBalance() (*asset.Balance, error) {
 
 // FeeRate returns the current optimal fee rate in sat / byte.
 func (btc *ExchangeWallet) feeRate(confTarget uint64) (uint64, error) {
-	feeResult, err := btc.node.EstimateSmartFee(int64(confTarget),
+	feeResult, err := btc.wallet.EstimateSmartFee(int64(confTarget),
 		&btcjson.EstimateModeConservative)
 	if err != nil {
 		return 0, err
@@ -1107,7 +1103,7 @@ func (btc *ExchangeWallet) FundingCoins(ids []dex.Bytes) (asset.Coins, error) {
 		if !notFound[pt] {
 			continue
 		}
-		txOut, err := btc.node.GetTxOut(txHash, rpcOP.Vout, true)
+		txOut, err := btc.wallet.GetTxOut(txHash, rpcOP.Vout, true)
 		if err != nil {
 			return nil, fmt.Errorf("gettxout error for locked outpoint %v: %w", pt.String(), err)
 		}
@@ -1411,7 +1407,7 @@ func (btc *ExchangeWallet) Redeem(redemptions []*asset.Redemption) ([]dex.Bytes,
 
 	// Send the transaction.
 	checkHash := msgTx.TxHash()
-	txHash, err := btc.node.SendRawTransaction(msgTx, false)
+	txHash, err := btc.wallet.SendRawTransaction(msgTx, false)
 	if err != nil {
 		return nil, nil, 0, err
 	}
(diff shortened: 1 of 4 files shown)


client/asset/btc/walletclient.go
```diff
@@ -193,7 +255,7 @@ func (wc *walletClient) Unlock(pass string) error {
 	return wc.call(methodUnlock, anylist{pass, 100000000}, nil)
 }
 
-// Lock locks the wallet.
+// Lock locks the wallet
```
amassarwi, line 258: revert

client/asset/btc/btc_test.go
```diff
@@ -225,9 +225,7 @@ func (c *tRPCClient) GetRawTransactionVerbose(txHash *chainhash.Hash) (*btcjson.
 	return c.mpVerboseTxs[txHash.String()], c.rawVerboseErr
 }
 
-func (c *tRPCClient) Disconnected() bool { return false }
-
-func (c *tRPCClient) RawRequest(method string, params []json.RawMessage) (json.RawMessage, error) {
+func (c *tRPCClient) RawRequest(_ context.Context, method string, params []json.RawMessage) (json.RawMessage, error) {
```
chappjc, line 228: Note that all the other `tRPCClient` need to go away and instead have `RawRequest` handle them now.

amassarwi: 💯

client/asset/btc/walletclient.go
```diff
@@ -56,6 +70,54 @@ func newWalletClient(node rpcClient, segwit bool, chainParams *chaincfg.Params)
...
+	return res, wc.call(methodEstimateSmartFee, anylist{confTarget, mode}, res)
+}
+
+func (wc *walletClient) SendRawTransaction(tx *wire.MsgTx, allowHighFees bool) (*chainhash.Hash, error) {
+	txBytes, err := serializeMsgTx(tx)
+	if err != nil {
+		return nil, err
+	}
+	return wc.callHashGetter(methodSendRawTransaction, anylist{
+		hex.EncodeToString(txBytes), allowHighFees})
```
chappjc, line 84: The trouble you're going to run into here is with how the sendrawtransaction RPC's syntax has changed.  See https://github.com/decred/dcrdex/pull/633

The `allowHighFees` boolean argument changed to a `maxFeeRate` numeric argument.  btcd's rpclient detects which to use depending on the version obtained from `getnetworkinfo`.  We'll need a switch to say which syntax to use.  e.g. https://github.com/decred/dcrdex/pull/633/files#diff-1b07e494f4bdd99b2a617ac110fbadc081e5b6ae1abf23f64032bbee5ebebef1R654

You are welcome to cherry-pick the two commits from that now-closed PR, or just make the same changes.  The btc_test.go changes might also be of help.

amassarwi: 🔥

client/asset/btc/btc_test.go
```diff
@@ -203,32 +172,65 @@ func (c *tRPCClient) GetBestBlockHeight() int64 {
...
+				bestBlkHeight = height
+				bestHash = hash
+			}
+		}
+		return json.Marshal(bestHash)
+	case methodGetRawMempool:
+		if c.mpErr != nil {
+			return nil, c.mpErr
+		}
+		return json.Marshal(c.mempoolTxs)
```
chappjc, line 226: I can give you an illustrative diff in the morning, but one of the things that's giving you trouble is (un)marshalling into/from types that do not support it.  Note how most of the dcrjson types have hashes as strings, and the rpcclient methods that do return types like `[]*chainhash.Hash` will first have to unmarshal the json.RawMessage from the remote server into a string, then decode the hex.  Unfortunately that's the burden of mocking out a fake server.  We have to deal with that.

client/asset/btc/btc.go
```diff
@@ -444,27 +463,22 @@ func BTCCloneWallet(cfg *BTCCloneCFG) (*ExchangeWallet, error) {
...
 
-	client, err := rpcclient.New(&rpcclient.ConnConfig{
-		HTTPPostMode: true,
-		DisableTLS:   true,
-		Host:         endpoint,
-		User:         btcCfg.RPCUser,
-		Pass:         btcCfg.RPCPass,
-	}, nil)
+	requester, err := newRequester(endpoint, btcCfg.RPCUser, btcCfg.RPCPass,
+		cfg.Symbol, cfg.Logger)
```
buck54321, line 467: I'd rather revert this for a couple of reasons. 1) the new function, `newRequester`, is only used once, and its name is not descriptive, since it actually returns an `rpcclient.Client`, not just a `RawRequester`. 2) It's a net increase in loc with what I feel is no improvement in readability.

client/asset/btc/btc.go
```diff
@@ -534,15 +549,15 @@ func (btc *ExchangeWallet) Connect(ctx context.Context) (*sync.WaitGroup, error)
 		return nil, fmt.Errorf("node software out of date. version %d is less than minimum %d", codeVer, minProtocolVersion)
 	}
 	// Initialize the best block.
-	h, err := btc.node.GetBestBlockHash()
+	h, err := btc.wallet.GetBestBlockHash()
```
buck54321, line 552: It's no longer just a wallet. Let's rename the `wallet` field to be more descriptive. Maybe just bring back `node`?

client/asset/btc/walletclient.go
```diff
@@ -5,48 +5,62 @@ package btc
...
+	methodGetRawTransaction  = "getrawtransaction"
 )
 
+type RawRequester interface {
+	RawRequest(context.Context, string, []json.RawMessage) (json.RawMessage, error)
+}
+
 // walletClient is a bitcoind wallet RPC client that uses rpcclient.Client's
 // RawRequest for wallet-related calls.
 type walletClient struct {
```
buck54321, line 53: Should rename this now, since it's far more than just the wallet functions. Now that we don't need the interface, maybe recycle `rpcClient`?

buck54321 (changes requested):
Testing well. Just a little cleanup needed.

amassarwi:
@buck54321 Updated 🤙

client/asset/btc/btc.go
```diff
@@ -2013,6 +2001,28 @@ func (btc *ExchangeWallet) send(address string, val uint64, feeRate uint64, subt
...
+	err = btc.node.call(methodSendRawTx,
+		anylist{hex.EncodeToString(b), float64(btc.feeRateLimit) * 1e-5}, &txid)
+	if err != nil {
+		return nil, err
+	}
+	if err != nil {
+		btc.log.Errorf("Error unmarshaling hash returned from sendrawtransaction."+
+			" Expected hash %s: %v", tx.TxHash(), err)
+		txid = tx.TxHash().String()
+	}
```
JoeGruffins, line 2022: Already returned on error so can't make it to the second condition statement.

client/asset/btc/btc.go
```diff
@@ -1143,20 +1134,17 @@ func (btc *ExchangeWallet) FundingCoins(ids []dex.Bytes) (asset.Coins, error) {
...
+	return btc.node.Unlock(pw)
 }
 
 // Lock locks the ExchangeWallet and the underlying bitcoind wallet.
 func (btc *ExchangeWallet) Lock() error {
-	if btc.node.Disconnected() {
-		return asset.ErrConnectionDown
-	}
-	return btc.wallet.Lock()
+	return btc.node.Lock()
```
JoeGruffins, line 1142: Now that this is `node` this reads awkwardly. What are we locking? I would argue that prefixing some of these methods with `Wallet` would be good. i.e. `btc.node.WalletLock()` There may also be an `AccountLock` at some point in time, so prepares for that as well. Just my opinion though.

client/asset/btc/walletclient.go
```diff
@@ -56,14 +70,77 @@ func newWalletClient(node rpcClient, segwit bool, chainParams *chaincfg.Params)
...
+	if err != nil {
+		return nil, err
+	}
+	return wc.callHashGetter(methodSendRawTransaction, anylist{
+		hex.EncodeToString(txBytes), allowHighFees})
+}
+
+func (wc *rpcClient) GetTxOut(txHash *chainhash.Hash, index uint32, mempool bool) (*btcjson.GetTxOutResult, error) {
+	// Note that we pass to call pointer to a pointer (&res) so that
+	//json.Unmarshal can  nil the pointer if the method returns the JSON null.
```
JoeGruffins, line 89: add a space before `json.Unmarshal` and remove a space after `can`

client/asset/btc/walletclient.go
```diff
@@ -56,14 +70,77 @@ func newWalletClient(node rpcClient, segwit bool, chainParams *chaincfg.Params)
 // sent via RawRequest.
 type anylist []interface{}
 
+func (wc *rpcClient) EstimateSmartFee(confTarget int64, mode *btcjson.EstimateSmartFeeMode) (*btcjson.EstimateSmartFeeResult, error) {
```
JoeGruffins, line 73: Can add comments for all new exported functions in this file?

client/asset/btc/walletclient.go
```diff
@@ -188,19 +265,19 @@ func (wc *walletClient) GetTransaction(txid string) (*GetTransactionResult, erro
 }
 
 // Unlock unlocks the wallet.
-func (wc *walletClient) Unlock(pass string) error {
+func (wc *rpcClient) Unlock(pass string) error {
```
JoeGruffins, line 268: `Unlock` -> `WalletUnlock` ?

client/asset/btc/walletclient.go
```diff
@@ -188,19 +265,19 @@ func (wc *walletClient) GetTransaction(txid string) (*GetTransactionResult, erro
...
 	// 100000000 comes from bitcoin-cli help walletpassphrase
 	return wc.call(methodUnlock, anylist{pass, 100000000}, nil)
 }
 
 // Lock locks the wallet.
-func (wc *walletClient) Lock() error {
+func (wc *rpcClient) Lock() error {
```
JoeGruffins, line 274: `Lock` -> `WalletLock` ?

JoeGruffins (commented):
Tests well!

client/asset/btc/btc.go
amassarwi: pffff

JoeGruffins (approved):

chappjc (approved):
Most of my involvement with this PR has been in matrix DM, but I've reviewed it.  All is working as expected with BTC.

Only thing I noticed, which must have existed as a bug prior, is that I was able to switch from an unencrypted wallet (`"gamma"`) to an encrypted wallet (alpha's `""`) and it made the reconfigure fine even though the wallet was incompletely configured.  I knew gamma required a password so I went back and set it (abc) but there was no indication of a problem.

client/asset/btc/btc.go
```diff
@@ -2029,6 +2017,23 @@ func (btc *ExchangeWallet) send(address string, val uint64, feeRate uint64, subt
...
+	if btc.legacyRawFeeLimit {
+		return btc.node.SendRawTransaction(tx, false)
+	}
+	b, err := serializeMsgTx(tx)
+	if err != nil {
+		return nil, err
+	}
+	var txid string
+	err = btc.node.call(methodSendRawTx,
+		anylist{hex.EncodeToString(b), float64(btc.feeRateLimit) * 1e-5}, &txid)
```
chappjc, line 2030: I just noticed `feeRateLimit` here.  Wallet should not be using `feeRateLimit` since it was only added
to help Core decide if it will work with a given server.  For sendrawtransaction, this should just be something large by default (or omitted).

buck54321: I agree. This is an optional argument, and we just shouldn't specify it here, defaulting to the node's default e.g. `0.1` for bitcoind. `feeratelimit` check is for `FundOrder`.

PR #944, opened 2021-01-27
multi: add Bitcoin Cash

buck54321:
Add server and client support for Bitcoin Cash.

The primary challenge is that Bitcoin Cash uses a special address encoding called Cash Address that encodes a pubkey hash using Bech32, but does not signify segwit. Another challenge is that the Bitcoin Cash uses a different signature hash algorithm.

Bitcoin Cash also doesn't implement `estimatesmartfee`. The closest thing they have is `estimatefee`, and their version takes no confirmation target argument. I'll look more into how the algo works.

Similar to Litecoin, Bitcoin has not yet implemented the new `sendrawtransaction` argument signature, and still uses the older `allowHighFees bool` argument.

To solve these challenges, I've made liberal use of the **github.com/gcash/bchd** and **github.com/gcash/bchutil** packages, which appears to be under active maintenance. Regardless, the functionality we use is likely to be battle-tested at this point, so I think we're good to go.

**IMPORTANT:** Bitcoin Cash uses the same data directory, `~/.bitcoin`, and the same executable names, `bitcoind` and `bitcoin-cli`, as Bitcoin Core. This makes life very difficult for both developers and users.

1. DO NOT start Bitcoin Cash without specifying a custom `-datadir`.  IF YOU FAIL to specify a custom data directory, your Bitcoin Core data will be destroyed and YOU'LL HAVE TO RESYNC BITCOIN.

**Avoid overwriting your Bitcoin block data** by adding `networkactive=1` to your **.bitcoin/bitcoin.conf**. `1` is the default value, but Bitcoin Cash doesn't support the setting, so will error if you forget to specify a custom `datadir`. 

2. Many of our simnet scripts rely on our executables being in `PATH`. **For dcrdex developers**, it is expected that the executables (or symlinks) are named `bitcoincashd` and `bitcoincash-cli`. The live tests on the server backend also expect your mainnet directory to be at **~/.bch**.


client/asset/interface.go
```diff
@@ -215,17 +215,17 @@ type Receipt interface {
...
 	// Coin is the coin that contains the contract.
-	Coin() Coin
+	Coin Coin
 	// Contract is the contract script.
-	Contract() dex.Bytes
+	Contract dex.Bytes
 	// SecretHash is the contract's secret hash.
-	SecretHash() dex.Bytes
+	SecretHash dex.Bytes
 }
```
buck54321, line 229: I needed to modify the `Recipient` outside of **client/asset/btc**, so making this a field was a solution, but I didn't see any reason that any of these needed to be methods.

dex/networks/btc/script.go
```diff
@@ -301,32 +301,24 @@ func ParseScriptType(pkScript, redeemScript []byte) BTCScriptType {
...
-func MakeContract(recipient, sender string, secretHash []byte, lockTime int64, segwit bool, chainParams *chaincfg.Params) ([]byte, error) {
-	rAddr, err := btcutil.DecodeAddress(recipient, chainParams)
-	if err != nil {
-		return nil, fmt.Errorf("error decoding recipient address %s: %w", recipient, err)
-	}
-	sAddr, err := btcutil.DecodeAddress(sender, chainParams)
-	if err != nil {
-		return nil, fmt.Errorf("error decoding sender address %s: %w", sender, err)
-	}
+func MakeContract(rAddr, sAddr btcutil.Address, secretHash []byte, lockTime int64, segwit bool, chainParams *chaincfg.Params) ([]byte, error) {
```
buck54321, line 304: Changing these arguments from `string` to `btcutil.Address` was needed to make things work with the new customizable `AddressDecoder`. It also appears to be a simplification when the changes are taken as a whole.

server/asset/dcr/dcr_test.go
```diff
@@ -1090,18 +1090,13 @@ func TestUTXOs(t *testing.T) {
...
-	err = contract.auditContract() // sets refund and swap addresses
+	contract, err := auditContract(utxo.Output) // sets refund and swap addresses
 	if err != nil {
 		t.Fatalf("case 13 - unexpected error auditing contract: %v", err)
 	}
-	if contract.SwapAddress() != swap.recipient.String() {
-		t.Fatalf("case 13 - wrong recipient. wanted '%s' got '%s'", contract.SwapAddress(), swap.recipient.String())
-	}
-	if contract.RefundAddress() != swap.refund.String() {
-		t.Fatalf("case 13 - wrong recipient. wanted '%s' got '%s'", contract.RefundAddress(), swap.refund.String())
```
buck54321, line 1104: The `RefundAddress` method remained unused, so I didn't replicate it as a field. Can always add it later.

server/asset/btc/btc.go
```diff
@@ -225,13 +268,8 @@ func (btc *Backend) Contract(coinID []byte, redeemScript []byte) (asset.Contract
...
 		return nil, err
 	}
-	contract := &Contract{Output: output}
 	// Verify contract and set refundAddress and swapAddress.
-	err = btc.auditContract(contract)
-	if err != nil {
-		return nil, err
-	}
-	return contract, nil
+	return btc.auditContract(output)
```
chappjc, line 272: Big thumbs up for the much clearer semantics here.

chappjc:
Just flagged this work for the 0.3 milestone, but I don't see why it can't be merged with the 0.2 release as an experimental and/or in-progress feature, perhaps just without full support for production mainnet use.

client/asset/btc/btc.go
```diff
@@ -361,6 +382,10 @@ type ExchangeWallet struct {
 	useSplitTx        bool
 	useLegacyBalance  bool
 	segwit            bool
+	legacyRawFeeLimit bool
+	signNonSegwit     TxInSigner
+	estimateFee       func(RPCClient, uint64) (uint64, error)
```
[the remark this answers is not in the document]
buck54321: The bch rpc for `estimatefee` is not the same as the legacy bitcoind `estimatefee` rpc. The bitcoin RPC at least took a confirmation target argument. Regardless, since it's not a difference between legacy/non-legacy, I figured a custom function is better.

JoeGruffins (commented):
Unable to run tests or set up a wallet with this error:
```
error creating bch wallet: error getting wallet balance for bch: rawrequest error: -32601: Method not found
```
```
livetest.go:160: error getting available: rawrequest error: -32601: Method not found
```

Could you add the info about what to name executables and being careful not to erase bitcoin's datadir to a README.md in the dex/testing/bch folder?

buck54321:
Should be fixed now @JoeGruffins . Lost something in a rebase.

JoeGruffins (approved):
Working well on simnet.

dex/networks/bch/params.go
```diff
@@ -0,0 +1,51 @@
...
+		// registration, so our only requirement is that it is unique. This one
+		// was just generated with a prng.
+		Net: 0xe8f3e1e3,
+	})
+	// TestNet3Params are the clone parameters for testnet.
+	TestNet3Params = btc.ReadCloneParams(&btc.CloneParams{
+		PubKeyHashAddrID: bchchaincfg.TestNet3Params.LegacyPubKeyHashAddrID,
+		ScriptHashAddrID: bchchaincfg.TestNet3Params.LegacyScriptHashAddrID,
+		Bech32HRPSegwit:  "bchtest",
+		CoinbaseMaturity: 100,
```
chappjc, line 31: These 4 appear same on testnet4, https://gitlab.com/bitcoin-cash-node/bitcoin-cash-node/-/blob/a5e1d83f308f0428b401816b113e9cf3fd1ecb03/src/chainparams.cpp#L499-500 (6f and c4)

client/asset/btc/btc.go
```diff
@@ -2542,7 +2600,28 @@ func (btc *ExchangeWallet) scriptHashAddress(contract []byte) (btcutil.Address,
 		return btcutil.NewAddressWitnessScriptHash(btc.hashContract(contract), btc.chainParams)
 	}
 	return btcutil.NewAddressScriptHash(contract, btc.chainParams)
+}
 
+// feeRate returns the current optimal fee rate in sat / byte using the
+// estimatesmartfee RPC.
+func (btc *ExchangeWallet) feeRate(ctx context.Context, _ RawRequester, confTarget uint64) (uint64, error) {
```
chappjc, line 2607: Can I ask that this stays where is was near `feeRateWithFallback`?

go.mod
```diff
@@ -24,14 +24,16 @@ require (
...
+	github.com/gcash/bchutil v0.0.0-20210113190856-6ea28dff4000
 	github.com/go-chi/chi v1.5.1
 	github.com/gorilla/websocket v1.4.2
 	github.com/jessevdk/go-flags v1.4.1-0.20200711081900-c17162fe8fd7
 	github.com/jrick/logrotate v1.0.0
 	github.com/lib/pq v1.2.0
 	github.com/smartystreets/goconvey v1.6.4 // indirect
 	go.etcd.io/bbolt v1.3.5
-	golang.org/x/crypto v0.0.0-20200820211705-5c72a883971a
+	golang.org/x/crypto v0.0.0-20201016220609-9e8e0b390897
```
chappjc, line 36: minor conflict on this now that we updated it further

server/asset/bch/live_test.go
```diff
@@ -0,0 +1,97 @@
...
+		dexAsset, err := NewBackend(configPath, logger, dex.Mainnet)
+		if err != nil {
+			fmt.Printf("NewBackend error: %v\n", err)
+			return 1
+		}
+
+		var ok bool
+		bch, ok = dexAsset.(*BCHBackend)
+		if !ok {
+			fmt.Printf("Could not cast asset.Backend to *Backend")
```
chappjc, line 66: to *BCHBackend

chappjc (changes requested):
Reads great, but still testing.  Just a few nits for now.

I'm happy to see the deinterfacing of both AuditInfo and Contract as it simplifies things and makes code dealing with those much more transparent.

server/asset/bch/live_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+func TestMain(m *testing.M) {
+	// Wrap everything for defers.
+	doIt := func() int {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithCancel(context.Background())
+		wg := new(sync.WaitGroup)
+		defer func() {
+			cancel()
+			wg.Wait()
```
chappjc, line 47: I noticed this panics if `Connect` does not succeed and thus wg.Add is never called.  We probably don't want to defer this closure unless Connect returns without error.

client/asset/btc/btc.go
```diff
@@ -502,8 +530,17 @@ func newWallet(requester RawRequester, cfg *BTCCloneCFG, btcCfg *dexbtc.Config)
...
 		legacyRawFeeLimit:   cfg.LegacyRawFeeLimit,
+		signNonSegwit:       nonSegwitSigner,
+		estimateFee:         cfg.FeeEstimator,
+		decodeAddr:          addrDecoder,
 		walletInfo:          cfg.WalletInfo,
-	}, nil
+	}
+
+	if w.estimateFee == nil {
+		w.estimateFee = w.feeRate
```
chappjc, line 540: The first time we call this, maybe even here in construction, if response is "method unknown" (or whatever the code is), I think it should be a hard error.  I say this because I was trying to run the BCH live_test.go and it actually passed the test with the BTC bitcoind except for an error about method unknown for the fee estimate.  Also makes me wonder if we should inspect some other RPC result to be sure we're even connecting to the correct asset.

chappjc:
New conflicts with fee-estimate merged.

dex/networks/bch/params.go
```diff
@@ -0,0 +1,51 @@
...
+		Net:              0xf4f3e5f4,
+	})
```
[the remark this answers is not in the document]
buck54321: So I started switching everything to testnet 4, but quickly ran into some hurdles because we're using the bchd `chaincfg.Params` in a few places, and no such `Params` exist for testnet4. The clone params are easy and I know exactly which fields are being used there, but I can't say the same for the bchd paths, so we can't use the partial params like we do with clone params. 

Can still do it, but punting for now. We should consider opening a PR at bchd with the new parameters too.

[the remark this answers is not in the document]
chappjc: Sounds good.  Was really just taking notes on testnet4 here.

client/asset/btc/btc.go
```diff
@@ -576,6 +577,13 @@ func (btc *ExchangeWallet) Connect(ctx context.Context) (*sync.WaitGroup, error)
 	if err != nil {
 		return nil, fmt.Errorf("error initializing best block for %s: %w", btc.symbol, err)
 	}
+	// Check for method unkown error for feeRate method.
+	_, err = btc.estimateFee(ctx, btc.node.requester, 1)
+	var rpcErr *dcrjson.RPCError
+	if errors.As(err, &rpcErr) && rpcErr.Code == dcrjson.ErrRPCMethodNotFound.Code {
```
chappjc, line 583: Sorry to circle back on this, but I believe we should be using the constants in btcjson, although clearly we'd need to assert with `dcrjson.RPCError`.  Either that or we just make a package level const with the expected value of `ErrRPCMethodNotFound.Code` to spare us importing btcjson just for one number.  Just that the error codes between dcr and btc's rpc servers aren't necessarily the same, and I've noted commits with bitcoin where they've actually changed the codes and I'm sure we didn't change them too.

server/asset/bch/live_test.go
```diff
@@ -63,15 +58,16 @@ func TestMain(m *testing.M) {
...
 			return 1
 		}
 
-		wg, err = dexAsset.Connect(ctx)
+		wg, err := dexAsset.Connect(ctx)
 		if err != nil {
 			fmt.Printf("Connect failed: %v", err)
 			return 1
 		}
+		defer wg.Wait()
```
chappjc, line 70: Look like it's going to start waiting before cancelling now when `m.Run()` completes because this defer is on the top of the defer stack.  When looking at the previous revision where I had requested a change, I had envisioned moving the `ctx, cancel = context.WithCancel(context.Background())` down to right before `dexAsset.Connect(ctx)` because `ctx` isn't used before that, thus allowing the `cancel(); wg.Wait()` to go in the same closure again, just down here.

chappjc (approved):

PR #954, opened 2021-01-30
client/asset/dcr: refactor findRedemptionsInBlockRange to use cfilters

itswisdomagain:
Also
- modify `TestFindRedemption` to test finding redeem txs in mempool
- fix re-org depth calculation in {dcr,btc}.checkNewBlocks.

### DIFF
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -2882,51 +2882,67 @@ func (dcr *ExchangeWallet) checkForNewBlocks() {
 		return
 	}
 
-	// Use the previous tip hash to determine the starting point for
-	// the redemption search. If there was a re-org, the starting point
-	// would be the common ancestor of the previous tip and the new tip.
-	// Otherwise, the starting point would be the block at previous tip
-	// height + 1.
-	var startPoint *block
-	var startPointErr error
-	prevTipBlock, err := dcr.node.GetBlockVerbose(dcr.ctx, prevTip.hash, false)
-	switch {
-	case err != nil:
-		startPointErr = fmt.Errorf("getBlockHeader error for prev tip hash %s: %w", prevTip.hash, translateRPCCancelErr(err))
-	case prevTipBlock.Confirmations < 0:
-		// There's been a re-org, common ancestor will be height
-		// plus negative confirmation e.g. 155 + (-3) = 152.
-		reorgHeight := prevTipBlock.Height + prevTipBlock.Confirmations
-		dcr.log.Debugf("reorg detected from height %d to %d", reorgHeight, newTip.height)
-		reorgHash, err := dcr.node.GetBlockHash(dcr.ctx, reorgHeight)
-		if err != nil {
-			startPointErr = fmt.Errorf("getBlockHash error for reorg height %d: %w", reorgHeight, translateRPCCancelErr(err))
-		} else {
-			startPoint = &block{hash: reorgHash, height: reorgHeight}
-		}
-	case newTip.height-prevTipBlock.Height > 1:
-		// 2 or more blocks mined since last tip, start at prevTip height + 1.
-		afterPrivTip := prevTipBlock.Height + 1
-		hashAfterPrevTip, err := dcr.node.GetBlockHash(dcr.ctx, afterPrivTip)
-		if err != nil {
-			startPointErr = fmt.Errorf("getBlockHash error for height %d: %w", afterPrivTip, translateRPCCancelErr(err))
-		} else {
-			startPoint = &block{hash: hashAfterPrevTip, height: afterPrivTip}
-		}
-	default:
-		// Just 1 new block since last tip report, search the lone block.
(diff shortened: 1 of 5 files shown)


client/asset/dcr/dcr.go
```diff
@@ -1882,67 +1885,98 @@ func (dcr *ExchangeWallet) findRedemptionsInMempool(contractOutpoints []outPoint
...
+// redemption attempts for any of these contracts.
+func (dcr *ExchangeWallet) findRedemptionsInBlockRange(startBlockHeight, endBlockHeight int64, contractOutpoints []outPoint) {
+	totalContracts := len(contractOutpoints)
 	dcr.log.Debugf("finding redemptions for %d contracts in blocks %d - %d",
-		contractsCount, startBlock.height, endBlock.height)
+		totalContracts, startBlockHeight, endBlockHeight)
 
-	nextBlockHash := startBlock.hash
-	var lastScannedBlockHeight int64
+	var lastScannedBlockHeight int64 = -1 // we've not scanned any block yet
```
chappjc, line 1914: `lastScannedBlockHeight` is only used for the `dcr.log.Debugf` after the loop, so there's little benefit from using a non-zero/default init afaict.  Having it -1 suggest to me I'm missing something or there's a logic error that would lead to the -1 value playing a role, but it just seems to be overwritten except on an early error.

client/asset/dcr/dcr.go
```diff
@@ -1882,67 +1885,98 @@ func (dcr *ExchangeWallet) findRedemptionsInMempool(contractOutpoints []outPoint
...
+			dcr.fatalFindRedemptionsError(err, contractOutpoints)
+			return
+		}
+		if !hit {
+			lastScannedBlockHeight = blockHeight
+			continue // block does not reference any of these contracts, continue to next block
+		}
+
+		// Pull the block info to confirm if any of its inputs spends a contract of interest.
+		blk, err := dcr.node.GetBlockVerbose(dcr.ctx, blockHash, true)
```
chappjc, line 1952: This is fine, but just a note: we don't really need getblock with either verbose=true or verboseTx=true since with a block's bytes we could deserialize and work on the `MsgTx`s.  No changes requests though.  It's going to be a headache to change, esp. on account of the helper methods needing `TxRawResult`s.  Besides this is not a hot path in dex client code.

chappjc (changes requested):
Good changes, esp. fixing `checkForNewBlocks`.  Mostly some nits, but also requesting to change to getblock**header** in `checkForNewBlocks`.

BTW, `TestMakerGhostingAfterTakerRedeem` is passing as expected.

client/asset/dcr/dcr.go
itswisdomagain: Will add a todo to that effect.

client/asset/dcr/dcr.go
```diff
@@ -2350,6 +2384,33 @@ func (dcr *ExchangeWallet) SyncStatus() (bool, float32, error) {
...
+	filterB, err := hex.DecodeString(blockCf2.Filter)
+	if err != nil {
+		return false, err
+	}
+	keyB, err := hex.DecodeString(blockCf2.Key)
+	if err != nil {
+		return false, err
+	}
+
+	filter, err := gcs.FromBytesV2(blockcf2.B, blockcf2.M, filterB)
```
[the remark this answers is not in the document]
itswisdomagain: Interesting crap. How did my eye survive that?

itswisdomagain:
dcr, btc simnet tests and TestMakerGhostingAfterTakerRedeem passing

client/asset/dcr/dcr.go
```diff
@@ -2048,6 +2084,33 @@ func (dcr *ExchangeWallet) fatalFindRedemptionsError(err error, contractOutpoint
 	dcr.findRedemptionMtx.Unlock()
 }
 
+// blockMaybeContainsScripts uses the cfilters of the specified block to
+// determine if the block likely includes any of the passed scripts.
+func (dcr *ExchangeWallet) blockMaybeContainsScripts(blockHash string, scripts [][]byte) (bool, error) {
```
chappjc, line 2089: Let's remember to refactor this method in terms of the `blockFilter` struct and the `getBlockFilterV2` methods slated to be added in https://github.com/decred/dcrdex/pull/788

itswisdomagain: Yeah, this will likely go in favor of that. Wanted a simple version in first.

chappjc (approved):
Looks and test well.  Just the mutex change in the test.

I'll leave this up for a day for others to check out.  I'm 50% through https://github.com/decred/dcrdex/pull/788 at the moment, and it's looking pretty good too.

JoeGruffins (approved):
Looks great and works. We should fix that reorg script...

itswisdomagain:
> We should fix that reorg script...

Yeah. I want to revisit that today.

PR #956, opened 2021-01-31
testing: Add Ethereum simnet harness.

### LINKED ISSUE #940: testing: Ethereum simnet harness.
Add an harness that sets up an eth [development network](https://ethereum.org/en/developers/docs/development-networks/). If possible I would like to avoid large dependencies, but am not sure what is needed yet. I will update this issue. I think this should happen first as far as eth support so we can test.

Edit: It looks like everything we need is contained in the go-ethereum repo, so there shouldn't be any other dependencies.

JoeGruffins:
This adds a basic ethereum private network testing harness.

This network relies on proof of authority. More about this can be found here https://eips.ethereum.org/EIPS/eip-225

It creates two nodes, either of which can add new blocks to the network. Each node starts with funds that can be sent to addresses that we later create in tests.

The harness does not start an rpc client. I think it is best that we use ipc (Inter-process communication) as this seems to be the most secure. Other options are an http or ws server, but neither is permissioned. They can be restricted to the local network, but at that point it doesn't seem much different than using ipc.

closes #940

### DIFF
--- a/dex/testing/eth/create-node.sh
+++ b/dex/testing/eth/create-node.sh
@@ -1,0 +1,82 @@
+#!/bin/sh
+# Script for creating eth nodes.
+set -e
+
+# The following are required script arguments
+TMUX_WIN_ID=$1
+NAME=$2
+NODE_PORT=$3
+CHAIN_ADDRESS=$4
+CHAIN_PASSWORD=$5
+CHAIN_ADDRESS_JSON=$6
+CHAIN_ADDRESS_JSON_FILE_NAME=$7
+ADDRESS=$8
+ADDRESS_PASSWORD=$9
+ADDRESS_JSON=${10}
+ADDRESS_JSON_FILE_NAME=${11}
+NODE_KEY=${12}
+
+WALLET_DIR="${NODES_ROOT}/${NAME}"
+mkdir -p "${WALLET_DIR}"
+
+# Write wallet ctl script.
+cat > "${NODES_ROOT}/harness-ctl/${NAME}" <<EOF
+#!/bin/sh
+geth --datadir="${WALLET_DIR}" \$*
+EOF
+chmod +x "${NODES_ROOT}/harness-ctl/${NAME}"
+
+# Write mine script.
+cat > "${NODES_ROOT}/harness-ctl/mine-${NAME}" <<EOF
+#!/bin/sh
+  case \$1 in
+      ''|*[!0-9]*)  ;;
+      *) NUM=\$1 ;;
+  esac
+  for i in \$(seq \$NUM) ; do
+    ./${NAME} attach --preload "${MINE_JS}" --exec 'mine()'
+  done
+EOF
+chmod +x "${NODES_ROOT}/harness-ctl/mine-${NAME}"
+
+
+# Write password file to unlock accounts later.
+cat > "${WALLET_DIR}/password" <<EOF
+$CHAIN_PASSWORD
+$ADDRESS_PASSWORD
+EOF
+
+# Create a tmux window.
+tmux new-window -t "$TMUX_WIN_ID" -n "${NAME}"
+tmux send-keys -t "$TMUX_WIN_ID" "set +o history" C-m
+tmux send-keys -t "$TMUX_WIN_ID" "cd ${WALLET_DIR}" C-m
+
+# Create and wait for a wallet initiated with a predefined genesis json.
+echo "Creating simnet ${NAME} wallet"
+tmux send-keys -t "$TMUX_WIN_ID" "${NODES_ROOT}/harness-ctl/${NAME} init "\
+	"$GENESIS_JSON_FILE_LOCATION; tmux wait-for -S ${NAME}" C-m
+tmux wait-for "${NAME}"
+
+# Create two accounts. The first is used to mine blocks. The second contains
+# funds.
+echo "Creating account"
+cat > "${WALLET_DIR}/keystore/$CHAIN_ADDRESS_JSON_FILE_NAME" <<EOF
+$CHAIN_ADDRESS_JSON
+EOF
+cat > "${WALLET_DIR}/keystore/$ADDRESS_JSON_FILE_NAME" <<EOF
+$ADDRESS_JSON
+EOF
+
+# The node key lets us control the enode address value.
+echo "Setting node key"
+cat > "${WALLET_DIR}/geth/nodekey" <<EOF
+$NODE_KEY
+EOF
+
+# Start the eth node with both accounts unlocked, listening restricted to
+# localhost, and syncmode set to full.
+echo "Starting simnet ${NAME} wallet"
+tmux send-keys -t "$TMUX_WIN_ID" "${NODES_ROOT}/harness-ctl/${NAME} --port " \
+	"${NODE_PORT} --nodiscover --unlock ${CHAIN_ADDRESS},${ADDRESS} " \
+	"--password ${WALLET_DIR}/password --miner.etherbase ${CHAIN_ADDRESS} " \
+	"--syncmode full --netrestrict 127.0.0.1/32" C-m
--- a/dex/testing/eth/create-node.sh
+++ b/dex/testing/eth/create-node.sh
@@ -16,13 +16,14 @@ ADDRESS_JSON=${10}
 ADDRESS_JSON_FILE_NAME=${11}
 NODE_KEY=${12}
 
-WALLET_DIR="${NODES_ROOT}/${NAME}"
-mkdir -p "${WALLET_DIR}"
+GROUP_DIR="${NODES_ROOT}/${NAME}"
+NODE_DIR="${GROUP_DIR}/node"
+mkdir -p "${NODE_DIR}"
 
 # Write wallet ctl script.
 cat > "${NODES_ROOT}/harness-ctl/${NAME}" <<EOF
 #!/bin/sh
-geth --datadir="${WALLET_DIR}" \$*
+geth --datadir="${NODE_DIR}" \$*
 EOF
 chmod +x "${NODES_ROOT}/harness-ctl/${NAME}"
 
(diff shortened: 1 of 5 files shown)


dex/testing/eth/harness.sh
```diff
@@ -0,0 +1,139 @@
...
+# Write mining javascript.
+# NOTE: This sometimes mines more than one block. It is a race.
+cat > "${MINE_JS}" <<EOF
+function mine() {
+  miner.start();
+  miner.stop();
+  admin.sleep(1.1);
+  return true;
+}
+EOF
```
JoeGruffins, line 88: This could definitely improved, but I currently can't see how. I don't think there's a way to specify the number of blocks you want to mine. calling start and stop in succession usually mines one, but sometimes, rarley, it mines two. Also, mining will only happen every second. So, if you were to call start stop start stop without waiting, it would only mine one block (or maybe two...) Sleeping for 1.1 secounds seems to ensure that you mine at least the number of blocks you wanted, but could be more.

dex/testing/eth/harness.sh
```diff
@@ -0,0 +1,139 @@
...
+
+echo "Starting simnet beta node"
+"${HARNESS_DIR}/create-node.sh" "$SESSION:2" "beta" "$BETA_NODE_PORT " \
+	"$CHAIN_ADDRESS" "$PASSWORD" "$CHAIN_ADDRESS_JSON " \
+	"$CHAIN_ADDRESS_JSON_FILE_NAME" "$BETA_ADDRESS" "$PASSWORD" \
+	"$BETA_ADDRESS_JSON " "$BETA_ADDRESS_JSON_FILE_NAME" "$BETA_NODE_KEY"
+
+# NOTE: This will cause beta to connect automatically to alpha.
+echo "Connecting nodes"
+"${NODES_ROOT}/harness-ctl/alpha" "attach --exec admin.addPeer('enode://${BETA_ENODE}@127.0.0.1:$BETA_NODE_PORT')"
```
JoeGruffins, line 134: Maybe this needs a sleep in-between starting and connecting, but this is working fine for me.

dex/testing/eth/harness.sh
```diff
@@ -0,0 +1,139 @@
+#!/bin/sh
+# Tmux script that sets up a simnet harness.
+set -ex
+
+SESSION="eth-harness"
+
+CHAIN_ADDRESS_JSON_FILE_NAME="UTC--2021-01-27T08-20-38.123221057Z--9ebba10a6136607688ca4f27fab70e23938cd027"
+CHAIN_ADDRESS="9ebba10a6136607688ca4f27fab70e23938cd027"
+CHAIN_ADDRESS_JSON='{"address":"9ebba10a6136607688ca4f27fab70e23938cd027","crypto":{"cipher":"aes-128-ctr","ciphertext":"dcfbe17de6f315c732855111b782496d76b2d703169afddaaa69e1bc9e02ec51","cipherparams":{"iv":"907e5e050649d1c5c0be782ec7db5cf1"},"kdf":"scrypt","kdfparams":{"dklen":32,"n":262144,"p":1,"r":8,"salt":"060f4e16d601069a6bccae0693a15cd72090baf1ab20e408c89883117d4f7c51"},"mac":"b9ca7dad75a04b77dc7751a814c051f32752603334e4bb4046caf927196a5579"},"id":"74805e39-6a2f-46eb-8125-70c41d12c6d9","version":3}'
```
JoeGruffins, line 9: This is basically what an eth wallet looks like. The wiki says that users should use this: https://geth.ethereum.org/docs/clef/tutorial
afaict, clef can provide deterministic keys. These are then saved as json here in the datadir/keystore directory. As I picture it now, we will need to require a user's clef master key and clef password and we will have to do all of the address management ourselves. I am of course an eth noob so not sure at all yet.

dex/testing/eth/harness.sh
```diff
@@ -0,0 +1,139 @@
...
+mkdir -p "${NODES_ROOT}/harness-ctl"
+
+echo "Writing ctl scripts"
+################################################################################
+# Control Scripts
+################################################################################
+
+# Write genesis json. ".*Block" fields represent block height where certain
+# protocols take effect. "clique" is our proof of authority scheme. One block
+# can be mined per second with the a signature belonging to the address in
```
JoeGruffins, line 47: ~the~

JoeGruffins:
Added a "group" directory so that we can have `alpha/{node, clef, other...}` and returning the probable number of blocks mined with the mine script. If you try something like `./mine-alpha 30` you are sure so see at lease one `2`

dex/testing/eth/harness.sh
JoeGruffins: Adding clef "wallets" in another pr. Perhaps they cannot be used while mining? I was able to cause some panics... Will add separate trading node/wallets that make use of clef.

JoeGruffins: I've decided against using clef. It only seems to complicate things. We can get what we need straight from geth. I don't think geth's account management should be something we should have to worry about so much. I have the basics for using clef going here, in case it turns out we should use it: https://github.com/JoeGruffins/dcrdex/commit/2240d499c5c938f121ff9351f6d261a936ecacbc

From the clef [readme](https://github.com/ethereum/go-ethereum/tree/master/cmd/clef) it looks like this is mostly a plus for remote signers. Since we will be requiring the eth node be local, I don't think it benefits us. If we decided to go with explorers, we would use it. That has not been our design pattern so far however.

JoeGruffins:
I had to add more funds to send a simple transaction. This still may not be enough.

dex/testing/eth/harness.sh
```diff
@@ -13,14 +13,14 @@ ALPHA_ADDRESS_JSON_FILE_NAME="UTC--2021-01-28T08-47-02.993754951Z--18d65fb8d60c1
...
-ALPHA_NODE_PORT="30302"
+ALPHA_NODE_PORT="30304"
 
 BETA_ADDRESS="4f8ef3892b65ed7fc356ff473a2ef2ae5ec27a06"
 BETA_ADDRESS_JSON_FILE_NAME="UTC--2021-01-27T08-20-58.179642501Z--4f8ef3892b65ed7fc356ff473a2ef2ae5ec27a06"
 BETA_ADDRESS_JSON='{"address":"4f8ef3892b65ed7fc356ff473a2ef2ae5ec27a06","crypto":{"cipher":"aes-128-ctr","ciphertext":"c5672bb829df9e209ca8ce18dbdd1fed69c603d639e06ab09127b672a609c121","cipherparams":{"iv":"24460eb2934c8b61cee3ad0aa7b843c0"},"kdf":"scrypt","kdfparams":{"dklen":32,"n":262144,"p":1,"r":8,"salt":"1f85da881994ca7b4a23f0698da70500a4b79f97a4450b83b129ebf3b4c28f50"},"mac":"1ecea707f1bffa1f6f944cb47e83118d8179e8a5005b83c88610b7e8692a1197"},"id":"56633762-6fb1-4cbf-8396-3a2e4661f7d4","version":3}'
 BETA_NODE_KEY="0f3f23a0f14202da009bd59a96457098acea901986629e54d5be1eea32fc404a"
 BETA_ENODE="b1d3e358ee5c9b268e911f2cab47bc12d0e65c80a6d2b453fece34facc9ac3caed14aa3bc7578166bb08c5bc9719e5a2267ae14e0b42da393f4d86f6d5829061"
-BETA_NODE_PORT="30303"
+BETA_NODE_PORT="30305"
```
JoeGruffins: 30303 was mainnet port, whoops

dex/testing/eth/create-node.sh
```diff
@@ -0,0 +1,118 @@
+#!/bin/sh
```
chappjc, line 1: Please update shebang and the new-window + new-session commands as per https://github.com/decred/dcrdex/pull/971

dex/testing/eth/create-node.sh
```diff
@@ -0,0 +1,118 @@
...
+
+[Node]
+DataDir = "${NODE_DIR}"
+
+[Node.P2P]
+NoDiscovery = true
+BootstrapNodes = []
+BootstrapNodesV5 = []
+ListenAddr = ":${NODE_PORT}"
+NetRestrict = [ "127.0.0.1/32" ]
```
chappjc, line 83: Can we add `::1` to this list? And maybe change the ipv4 loopback network to 127.0.0.0/8

chappjc (approved):
Awesome work, @JoeGruffins!  I think we're going to to learn a lot about how to do stuff like reliably mining a single block, and managing multiple accounts (wallets), as we proceed, and this is a wonderful start.

Just one change if it's possible.

dex/testing/eth/harness.sh
chappjc: How did you generate this json for `CHAIN_ADDRESS_JSON`?

JoeGruffins: I just copy pasted straight from an "account" created with geth. You can find them as files in `~/.ethereum/keystore/long-name-with-date`.

dex/testing/eth/harness.sh
JoeGruffins: 0 also possible
```
[harness-ctl]$ ./mine-alpha 6
1
1
1
0
2
1
```

dex/testing/eth/create-node.sh
JoeGruffins: Ok.

dex/testing/eth/create-node.sh
JoeGruffins: Looks ok?

JoeGruffins:
Odd test failure. Something known or should look into it?

chappjc:
> Odd test failure. Something known or should look into it?

That's a new one.  Will have to review what merged today to see if anything is to blame.  For now, just retriggering the tests.

dex/testing/eth/harness.sh
chappjc: Dang.  There's gotta be a js while loop solution to spin until a block is mined.

dex/testing/eth/create-node.sh
chappjc: Need it in the heredocs too. e.g. https://github.com/decred/dcrdex/pull/956/commits/d76d92deab2fba32cfe6883b30ff8438f7f4490e#diff-6fd0fb5ffa4a8526221349744d156b536c50c9959606a486034aa7b807960c61R33

JoeGruffins: ah

PR #957, opened 2021-02-01
client/{core,db}: introduce match.Retired field to track match activeness

itswisdomagain:
Set trackedMatch.Retired to true when a match becomes inactive, i.e. any of the
following occurs:

- `redeem` request to DEX succeeds after redeeming counter-party swap.
 Final step that is REQUIRED for any party to a match (taker or maker) is notifying
 the DEX of their redemption of the counter-party swap.
- counter-party swap redemption is broadcast after the match has been revoked.
 Regardless of the reason for the revocation, a client cannot notify a DEX of their
 redemption for a revoked match. After a revoked match party successfully broadcasts
 their redemption of the counter-party swap, the party has no need to take further
 actions on the revoked match and the match can be safely retired.
- own swap is refunded, for whatever reason.
- match is revoked, either prior to the client sending a swap or client has refunded
 own swap or (auto-)redeemed the counter-party's swap.

Use the trackedMatch.Retired field to determine match activeness instead of the
trackedTrade.matchIsActive method.


JoeGruffins (approved):

client/db/types.go
```diff
@@ -212,10 +212,51 @@ func (m *MetaMatch) ID() []byte {
...
 }
 
+// Revoke marks this match as revoked and retires the match if no further
+// action is required. Recovery actions may be required for a revoked match
+// if the client has broadcasted a swap contract and has not redeemed the
+// counter-party's contract or refunded own swap.
+// Returns whether or not the match is retired by this revocation action.
+func (m *MetaMatch) Revoke(fromServer bool) bool {
+	m.MetaData.Proof.serverRevoked = fromServer
+	m.MetaData.Proof.selfRevoked = !fromServer
```
chappjc, line 222: I think this is problematic as many of such matches will start out as server revoked and get switched by the refund or auto-redeem functions to self-revoked.

client/db/types.go
```diff
@@ -212,10 +212,51 @@ func (m *MetaMatch) ID() []byte {
...
+	// is necessary. This should not be applied to cancel order matches.
+	wasRetired := m.MetaData.Retired
+	status, side := m.Match.Status, m.Match.Side
+	m.MetaData.Retired = m.Match.Address == "" || // cancel order match
+		status == order.NewlyMatched || // no swap sent yet, no recovery required
+		(status == order.MakerSwapCast && side == order.Taker) || // we're taker, haven't sent swap, no recovery required
+		(side == order.Maker && status == order.MakerRedeemed) || // we're maker and we've redeemed, no recovery required
+		status == order.MatchComplete // we've redeemed, no further action is necessary
+
+	return !wasRetired && m.MetaData.Retired
```
chappjc, line 234: Docs for this method should be clearer that it could return `false` if the match was **already** retired.

itswisdomagain: > to avoid repeated revoke notifications

Yeah... DB updates may still need to be performed regardless.

client/db/types.go
itswisdomagain: I wonder why that should happen though. If it was already server revoked, no need to self-revoke. Although if a server revoke comes after a self-revoke, it'd be necessary to update that (e.g. contract locktime elapses and client auto-refunds before server revokes the match).

In any case, it'll be confusing to have both server- and self-revoke set to true for a match.

itswisdomagain:
I just remembered I have to add a commit for the db migration.

client/db/types.go
chappjc: The server would almost certainly revoke the match before client tries a refund, then the refund code in this PR sets self-revoked, unless I'm reading it wrong.  https://github.com/decred/dcrdex/pull/957/files#diff-8962b4c526755f0f98ed9f46ae05827e8edbc43ed5671ea4a2624bdf1617496cR1854

client/core/trade.go
```diff
@@ -1873,7 +1851,7 @@ func (t *trackedTrade) refundMatches(matches []*matchTracker) (uint64, error) {
 			refundedQty += calc.BaseToQuote(match.Match.Rate, match.Match.Quantity)
 		}
 		proof.RefundCoin = []byte(refundCoin)
-		proof.SelfRevoked = true // Set match as revoked.
+		t.revokeMatch(match, false) // self-revoke this match
 		err = t.db.UpdateMatch(&match.MetaMatch)
```
chappjc, line 1855: This is where I figure the match is already revoked by the server.  Maybe there should be a condition to check first if it's already revoked.

client/db/types.go
itswisdomagain: It does and I'm thinking, if the match is already server revoked, self-revocation should be prevented. This `Revoke` method might be the best place to prevent that.

If the match was already self-revoked, then perhaps we can allow _also_ setting server-revoked but not the reverse.

Or perhaps, could retain current behaviour of setting a second revoke if there was already an initial revoke.

JoeGruffins:
> I just remembered I'm yet to add a commit for the db migration.

I was thinking that too, but I tried master -> this in the middle of a match and it didn't seem to matter.

client/db/bolt/upgrades_test.go
```diff
@@ -150,6 +128,39 @@ func verifyV2Upgrade(t *testing.T, db *bbolt.DB) {
...
 
+func verifyV3Upgrade(t *testing.T, db *bbolt.DB) {
+	verifyMatches := map[string]bool{
+		"090f54bc6b13652d4e4c23ec94b4f70751c61db043ae732c014d0f5cf4ccccb7": false, // status < MatchComplete, not revoked, not refunded
+		"5ee45ebb29edf84ab2a7f35aef619b20c3e2cd26017a5e281e7cc1bfde0ed435": true,  // status < MatchComplete, revoked at NewlyMatched
+		"84455dae423b09f149396f09b6063380eeb312e1b67051a695e78cd5c985cb49": false, // status == MakerSwapCast, side Maker, revoked, requires refund
+		"59c054d537fbd98f51621671a84d25b55350d0e33a6b3c2c4abd874054f8a438": true,  // status == TakerSwapCast, side Maker, revoked, refunded
+		"f153a3b8d5824c3e237911cd1b108dfb2e04555f6c5899681d182216caa328fc": false, // status == MatchComplete, no RedeemSig
+		"3996ffea2a3db99262b954cf6dcf883f562baf273d767b7cf91b6aacddd4206f": true,  // status == MatchComplete, RedeemSig set
+	}
```
itswisdomagain, line 139: I wrote [a tool](https://github.com/itswisdomagain/dcrdex/blob/00d492e9c8a5dd98b6720e102d497a4684596800/client/db/bolt/testdata/v2.db.gen.go) to generate these matches using the previous db version and log the ids and expected retirement status for each generated match.

Can verify by running this test with the `v2.db.gz` file generated by that tool and pasting the log output here.

chappjc: I'd be good with checking that in if you give it a build tag like `// +build ignore` so it doesn't break anything.

client/db/types.go
buck54321: There is no required relation between `ServerRevoked` and `SelfRevoked`. We should set them as needed, regardless of what the other field is set to.

client/db/bolt/upgrades_test.go
itswisdomagain: Looks like that file will have to go in before this PR because it cannot be executed if it goes in with this PR. If it goes in a preceding commit, it can be executed by checking out that commit. Separate PR or nah?

client/core/trade.go
```diff
@@ -735,51 +737,15 @@ func (t *trackedTrade) isActive() bool {
...
-	return true
+	return match.IsRevoked()
 }
 
 func (t *trackedTrade) activeMatches() []*matchTracker {
 	var actives []*matchTracker
 	t.mtx.RLock()
 	defer t.mtx.RUnlock()
 	for _, match := range t.matches {
-		if t.matchIsActive(match) {
```
chappjc, line 782: Although I think we'll end up accepting these changes, we need to reflect on the risk-reward of this diff because my view is that the changes are considerably more extensive and risky than necessary.

The problem to solve is fundamentally a DB issue:
- only taker
- match with final redeem broadcast
- not yet completed the `'redeem'` request
- restart of core does not reload the match **because it does not consider RedeemSig** in the DB

The severity of the issue is extremely low:
- very hard to hit (only taker, must shutdown after redeem bcast but before redeem request)
- very low impact (all txns bcast, no loss of funds, only the very lowest penalty)

Given the above, the way I consider the acceptable resolution is that we have very little tolerance for extensive changes or the possibility of new bugs, and the fix should be absolutely minimal.  I've commented on this line

`if t.matchIsActive(match) {` ==> `if !match.MetaData.Retired {`

because it is gets to what I think is too broad with the diff.  We are making changes to how `core` deals with order handling, both in "activeness" and revoke handling.  But afaict, nothing really needed to change in these places.  Now we have a new concern: keeping `Retired` consistent with all the conditions that were previously defined in `(*trackedTrade).matchIsActive`.

Considering this: you've added `Retired` to `dexdb.MatchMetaData`, so that solves the loading issue.  To support that we only needed to compute the correct `Retired` flag **when storing** Match data (a DB job), not necessarily change anything in core's operation.

I'll also note that yet another solution that was discussed that is even less involved, with no DB upgrade, is simply decoding match proof to check the redeem sig for taker matches with MatchSomplete status.  That solution is very very low risk, with only a minor (likely imperceptible) performance penalty when loading matches.  I ack'd that a `Retired` flag would probably be a better solution partly because of the inelegance of loading the match proof for all taker matches, hence this PR, but these were my concerns about it.

I won't ask to revert lots of code, but I would ask for comments regarding my conclusions above.  Were the changes in core strictly needed?  Are other problems solved beyond what I described, or are these more significant improvements to core than I have concluded?

client/core/trade.go
```diff
@@ -1739,11 +1717,11 @@ func (c *Core) sendRedeemAsync(t *trackedTrade, match *matchTracker, coinID, sec
 		auth := &match.MetaData.Proof.Auth
 		auth.RedeemSig = ack.Sig
 		auth.RedeemStamp = encode.UnixMilliU(time.Now())
-		if match.Match.Side == order.Maker {
-			// As maker, this is the end. However, this diverges from server,
-			// which still needs taker's redeem.
-			match.SetStatus(order.MatchComplete)
```
chappjc, line 1745: This `SetStatus` was already set as taker, hence the condition.  But it's OK to have a redundant SetStatus to be safe.  Was there another reason to remove the condition?  The implications of setting a status are that the status is changing, so if debugging this it's going to look like a bug for taker, at least without a comment.

client/core/trade.go
```diff
@@ -1659,18 +1629,26 @@ func (c *Core) redeemMatchGroup(t *trackedTrade, matches []*matchTracker, errs *
...
+			// match.MetaData.Retired = true // Don't retire until `redeem` request to DEX completes.
 			proof.TakerRedeem = coinID
 		} else {
 			match.SetStatus(order.MakerRedeemed)
 			proof.MakerRedeem = coinID
 		}
+		revoked := match.IsRevoked()
+		if revoked {
+			match.MetaData.Retired = true // Redemption or refund is the last step for revoked matches.
+		}
```
chappjc, line 1642: Pls comment that this would be expected in the auto-redeem situation where counterparty broadcasted their redeem too late and without server coordination.

But this an example of where we've distributed the responsibility of deciding "activeness" from one place to all over core, and I think that's more error prone than doing/checking with a single piece of code.

client/core/status.go
```diff
@@ -69,9 +69,7 @@ func (c *Core) resolveMatchConflicts(dc *dexConnection, statusConflicts map[orde
 				// but it was revoked in the short period since then.
 				c.log.Errorf("Server did not report a status for match during resolution. %s",
 					statusResolutionID(dc, trade, match))
-				// revokeMatch only returns an error for a missing match ID, and
-				// we already checked in compareServerMatches.
-				_ = trade.revokeMatch(match.id, false)
```
chappjc, line 74: This is why the individual resolvers don't have to call `revokeMatch` to get the notifications and return of coins, instead justting set `match.MetaData.Proof.{Self,Server}Revoked`.  Aren't we now doing those actions twice (in the resolver and here)?

client/core/trade.go
itswisdomagain: I do see that setting the `Retired` flag on db store only will resolve the issue at hand; didn't really think in that direction. One key reason for making the change wider was to have a more consistent definition of match activeness.

I'd earlier commented (much earlier) that having separate codes on db and core perform match activeness checks is prone to errors, as we've now seen and are trying to fix. While having the `Retired` flag work only with the db will solve the issue at hand, we'll still have 2 separate codes that determine match activeness, which could cause another issue of this sort to re-occur.

That said, I do also see that with the current approach taken in this PR, setting `match.Retired` is scattered all over the place. I had thought of making a single function that toggles that value, a function called when a match status progresses and when a match is retired; or any other scenario that changes a match activeness status. It could even still be part of the db match update method but that'd mean that simply changing a match status or retiring a match without an immediate successive call to db update will not set the Retired field appropriately. I'd initially gone the route of having a single code perform match activeness checks and setting the `Retired` field in one place but it requires the changes in #962 to go in first.

In summary, (apologies for that pretty long and hopefully not-confusing explanation), I agree with making this fix much simpler for the reasons you've mentioned above. I do still think we should have just one function that updates the `Retired` field appropriately and consistently and then both core and db can use that field to check for match activeness rather than having both pkgs maintain different codes for determining match activeness.

client/core/status.go
itswisdomagain: I'll take another look here, but I'm sure it can be modified to ensure it's not done twice.

itswisdomagain:
I've simplified the changes here to focus on db match update effects only.

chappjc:
Pretty certain this became obsolete when @JoeGruffins added the archived orders bucket.  Thoughts on this guys?

JoeGruffins:
That was orders, and the purpose was better performance, not having to sift through a lot of finished orders to find the one's that are active.

What was the end goal of this I don't quite remember? If performance maybe a separate bucket for active and retired would be best here too?

chappjc:
Ah right, that was an orders bucket, good point @JoeGruffins.

This PR was addressing a fairly hard to hit edge case described here: https://github.com/decred/dcrdex/pull/957#discussion_r589494078

chappjc:
Hmm, gosh I need to re-review now that the core changes are reverted as per my request. :blush:

JoeGruffins:
I wish I had a strong opinion but I don't. What @itswisdomagain says about having a single source for truth about whether a match is retired is definitely true, if that is not the case right now, I'm not sure.

chappjc:
`(*trackedTrade).matchIsActive` is the source of truth, _if you have a `trackedTrade`_, which of course `core` does, but `db` just has buckets of bytes.  So, we're just dancing around the issue of not wanting to load and decode the match proof in the DB for each and every match to determine if it's indeed "retired".  I think wisdom's latest changes that restricted a `Retired` flag to the db package are probably what will seal the deal for me.  I'm afraid I missed that push around May 25.  Will look again soon

itswisdomagain:
> restricted a `Retired` flag to the db package

Yeah, that greatly simplifies things. The `Retired` flag is set and stored when a match's data is about to be persisted to db. Makes loading only active matches very straightforward subsequently - db code just checks the `retired` field of the match's db data.



> Maybe code can be shared between `(*trackedTrade).matchIsActive` and `db.isRetiredMatch` to address all concerns.

Just looked at `db.isRetiredMatch` again and you're right. I'll look into that.

client/db/bolt/upgrades_test.go
chappjc: I had meant to respond to this with "preceding commit" for the gen tool.

Although I keep going back to @JoeGruffins' active orders bucket upgrade (`v4Upgrade` and the `moveActiveOrders` helper) as a model solution for the matches too.  I don't think we considered this when this work was initiated, but it not only solves the problem of knowing which matches in the DB are active or retired, but it provides a major performance boost by not having to iterate over ancient matches.

Only thing I'm not sure about is if there's an active order that has some retired matches, would it become a burden to have to check two match buckets for such an order's matches (we display all matches for active orders including the inactive ones).

I will poke around to see.  This Retired flag does seem to work, but if it would be obsoleted with a separate matches bucket, I'd hate to go ahead with the DB upgrade, esp. because we've seen OOM issues with these upgrades on large DBs and updating every entry of the matches bucket is likely to be the worst case scenario.

I will test some things with the current PR and look at these ideas.  No need to do anything yet.  Let's just prioritize #1210 for now.

chappjc:
This was replaced with https://github.com/decred/dcrdex/pull/1263, which included this PRs commits.  These commits were merged into master, so this effort is on master @itswisdomagain

PR #958, opened 2021-02-02
multi: order fee estimation

buck54321:
An order fee estimate consists of 3 values for the swap and 2 for the redemption. The swap has a max possible fees (`MaxFeeRate`), and both high and low estimates corresponding to the best and worst case settlement sequences, but at the prevailing network fee rate.

Similarly, the redemption estimate has high and low estimates evaluated for the worst and best settlement sequences (1 tx per lot
vs 1 tx for all).


client/asset/interface.go
```diff
@@ -290,20 +297,81 @@ type Order struct {
...
+	// Estimate is the OrderEstimate for an order with the specified
+	// option values.
+	Estimate *SwapEstimate
+}
+
+// PreSwap is a SwapEstimate returned from Wallet.PreSwap. The struct will be
+// expanded in in-progress work to accomondate order-time options.
+type PreSwap struct {
+	Estimate *SwapEstimate `json:"estimate"`
+}
```
buck54321, line 361: This is one of a few seemingly unnecessary wrapper structs that will soon be used to accommodate the order-time options. I've made some temporary comments in this file on the to explain their purpose.

Since order-time options are entirely optional, setting up the structs now means that when the options become available, we don't need to break the API.

client/asset/btc/btc_test.go
```diff
@@ -2281,3 +2311,116 @@ func TestSyncStatus(t *testing.T) {
...
+		t.Fatalf("no PreSwap error for not enough funds: %v", err)
+	}
+	setFunds(minReq)
+
+	node.estFeeErr = tErr
+	_, err = wallet.PreSwap(form)
+	if err == nil {
+		t.Fatalf("no PreSwap error for not enough funds: %v", err)
+	}
+	node.estFeeErr = nil
```
chappjc, line 2385: Minor conflict with the `if c.estFeeErr != nil { return nil, c.estFeeErr }` needing to be in the `RawRequest` stub now.  Also looks like the `Fatalf` message here is copy-pasta'd from the previous test case.

client/core/core.go
```diff
@@ -2333,18 +2371,33 @@ func (c *Core) MaxSell(host string, base, quote uint32) (*OrderEstimate, error)
...
-	return &OrderEstimate{
-		Lots:           maxSell.Lots,
-		Value:          maxSell.Value,
-		MaxFees:        maxSell.MaxFees,
-		EstimatedFees:  maxSell.EstimatedFees,
-		Locked:         maxSell.Locked,
-		RedemptionFees: maxSell.Lots * sellRedemptionPer,
+	// Estimate a quote-converted lot size.
+	lotSize := quoteAsset.LotSize
+	book, found := dc.books[marketName(base, quote)]
```
chappjc, line 2382: The `books` map can change out from under us in `handleReconnect` and a few spots in bookie.go.

client/core/core.go
```diff
@@ -2333,18 +2371,33 @@ func (c *Core) MaxSell(host string, base, quote uint32) (*OrderEstimate, error)
...
+	if found {
+		midGap, err := book.MidGap()
+		if err == nil {
+			lotSize = calc.BaseToQuote(midGap, baseAsset.LotSize)
+		}
+	}
```
chappjc, line 2388: Would you comment on the meaning of `!found`?  Looks like if there's no book for the base-quote pair, then the quote asset's lot size is used instead of a "quote-converted" lots size based on current exchange rates.  I'm wondering about the utility of this case, or if it should be an error considering the `MaxSell` caller is giving a base-quote pair they intend to trade on.

client/core/core.go
```diff
@@ -2333,18 +2371,33 @@ func (c *Core) MaxSell(host string, base, quote uint32) (*OrderEstimate, error)
...
+
+	preRedeem, err := quoteWallet.PreRedeem(&asset.PreRedeemForm{
+		LotSize: lotSize,
+		Lots:    1,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("%s RedemptionFees error: %v", unbip(base), err)
```
chappjc, line 2395: Old error message?  Also, `unbip(quote)` instead of base?

client/asset/interface.go
```diff
@@ -88,10 +88,11 @@ type Wallet interface {
...
 	// will be an estimate based on current market conditions.
-	MaxOrder(lotSize uint64, nfo *dex.Asset) (*OrderEstimate, error)
-	// RedemptionFees is an estimate of the redemption fees for a 1-swap
-	// redemption.
-	RedemptionFees() (uint64, error)
+	MaxOrder(lotSize uint64, nfo *dex.Asset) (*SwapEstimate, error)
+	// PreSwap gets an OrderEstimate for the specified order size(s).
+	PreSwap(*PreSwapForm) (*PreSwap, error)
+	// PreRedeem gets an OrderEstimate for the specified order size(s).
+	PreRedeem(*PreRedeemForm) (*PreRedeem, error)
```
chappjc, line 95: Both are documented as returning an `OrderEstimate`.

client/core/core.go
```diff
@@ -2333,18 +2371,33 @@ func (c *Core) MaxSell(host string, base, quote uint32) (*OrderEstimate, error)
...
+	if found {
+		midGap, err := book.MidGap()
+		if err == nil {
+			lotSize = calc.BaseToQuote(midGap, baseAsset.LotSize)
+		}
+	}
+
+	preRedeem, err := quoteWallet.PreRedeem(&asset.PreRedeemForm{
+		LotSize: lotSize,
+		Lots:    1,
```
chappjc, line 2392: This probably relates to your future plans for these methods, but `Lots` is 1 in both `MaxSell` and `MaxBuy`.  Also, `LotSize` is not used in either `PreRedeem` implementation.
So taken together, that means that `preRedeem` is constant regardless of lot size for either asset, the book, or maxSell.

client/orderbook/orderbook.go
```diff
@@ -558,3 +558,18 @@ func (ob *OrderBook) MidGap() (uint64, error) {
 	}
 	return (s[0].Rate + b[0].Rate) / 2, nil
 }
+
+// BestFill is the best (rate, quantity) fill for an order of the type and
+// quantity specified.
```
chappjc, line 563: Maybe mention that `BestFillMarketBuy` should be used for market buys, but this is good for limits and market sells, where quantity is in multiples of the base asset's lot size.

chappjc (changes requested):
Looks pretty good.  Just some minor nits and a little confusion on my part about how `PreRedeem` is called from `MaxBuy` and `MaxSell`, also if `PreRedeem` is going to use lotsize at some point because it's only using lots now.

client/asset/btc/btc.go
```diff
@@ -705,72 +705,149 @@ func (btc *ExchangeWallet) feeRateWithFallback(confTarget uint64) uint64 {
...
+	return &asset.PreSwap{
+		Estimate: est,
+	}, nil
 }
 
-// RedemptionFees is an estimate of the redemption fees for a 1-swap redemption.
-func (btc *ExchangeWallet) RedemptionFees() (uint64, error) {
+// estimateSwap prepares an *asset.SwapEstimate.
+func (btc *ExchangeWallet) estimateSwap(lots, lotSize, networkFeeRate uint64, utxos []*compositeUTXO,
+	nfo *dex.Asset, trySplit bool) (*asset.SwapEstimate, bool /*split used*/, error) {
```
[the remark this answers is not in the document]
buck54321: > It will be in the future?

Yes.

client/asset/btc/btc_test.go
```diff
@@ -927,10 +942,10 @@ func TestFundEdges(t *testing.T) {
 	// swap_size_base: 76 bytes (225 - 149 p2pkh input) (InitTxSizeBase)
 	// lot_size: 1e6
 	// swap_value: 1e7
-	// lots = swap_size / lot_size = 10
-	//   total_bytes = first_swap_size + chained_swap_sizes
+	// lots = swap_value / lot_size = 10
 	//   chained_swap_sizes = (lots - 1) * swap_size
 	//   first_swap_size = swap_size_base + backing_bytes
+	//   total_bytes = first_swap_size + chained_swap_sizes
 	//   total_bytes  = swap_size_base + backing_bytes + (lots - 1) * swap_size
```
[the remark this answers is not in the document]
buck54321: These are actually just two different formula for the same thing. I probably just shouldn't have moved this. It made more sense where it was, I think.

client/asset/btc/btc.go
```diff
@@ -705,72 +705,149 @@ func (btc *ExchangeWallet) feeRateWithFallback(confTarget uint64) uint64 {
...
+		outputSize = dexbtc.P2PKHOutputSize
 	}
-	return size * feeRate, nil
+	best += inputSize*req.Lots + outputSize
+	worst += (inputSize + outputSize) * req.Lots
+
+	return &asset.PreRedeem{
+		Estimate: &asset.RedeemEstimate{
+			RealisticWorstCase: worst * feeRate,
+			RealisticBestCase:  best * feeRate,
```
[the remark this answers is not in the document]
buck54321: All we can really say before matching is that the fees will (realistically) fall in a range between the best and worst case. Best case provides the lower limit. Worst cast provides the upper limit.

client/asset/dcr/dcr.go
```diff
@@ -687,68 +687,143 @@ func (a amount) String() string {
...
+	if err != nil {
+		return nil, err
 	}
-	return &asset.OrderEstimate{}, nil
+	if maxLots < req.Lots {
+		return nil, fmt.Errorf("%d lots available for %d-lot order", maxLots, req.Lots)
+	}
+
+	// Get the estimate at using the current configuration
+	est, _, err := dcr.estimateSwap(req.Lots, req.LotSize, feeRate, utxos, req.AssetConfig, dcr.useSplitTx)
```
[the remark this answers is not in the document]
buck54321: You're right that it is only the number of lots right now, so this message makes no sense. I updated the message. This message is from the future.

client/core/core.go
buck54321: You've convinced me. Error is better.

client/core/core.go
buck54321: Hmm. I think I should be using the number of lots returned in the swap estimate.

client/asset/btc/btc.go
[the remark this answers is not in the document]
buck54321: Just to expand on this a little, the worst case is probably the one I would pay attention to the most, but for users who place large orders, they may want to collect statistics over time to better estimate where their risk falls between these two points. 

From a UI perspective, this will allow us to offer some additional clarity on fees instead of just showing the worst-case value.

chappjc (approved):

PR #959, opened 2021-02-02
pg: add schema_version and upgrade for epoch_reports

chappjc:
This adds DB scheme versioning to the postgresql backend (server/db/driver/pg) and implements two upgrades:
1. v1: On the `meta` table, remove the `state_hash` column and add `schema_version`.  The `state_hash` column was used prior to https://github.com/decred/dcrdex/pull/856, and was left for this upgrade.
2. v2: Create the `epoch_reports` table (if not exists) and import match volume data for each epoch in the `epochs` table by reading match data from the `matches` table.

Notes about the v2 upgrade (`epoch_reports` data import):
- A row in `epoch_reports` is created for each row of the `epochs` table, even if there were no matches.  This happens during normal operation.
- This only accurately sets values for `match_volume`, `quote_volume`, `low_rate`, and `high_rate`.
- The `start_rate` and `end_rate` fields are not based on book data as during normal operation where the book's midGap before/after matching is used.  Instead `end_rate`, is set to the average of an epoch's high and low rates, while `start_rate` is taken from the previous epoch *with trade matches*.  As such, these are ballpark rates, not midGap book rates.
- The `book_buys*` and `book_sells*` fields (booked volume on either side within certain distance from midGap/market rate) are set to zero in the absence of a book snapshot to compute them.  See `InsertPartialEpochReport`.
- Requires a temporary index on `matches (epochidx, epochdur)` during the upgrade.

Tested on a DB with 1,030,671 epochs containing 16,557 matches, the v2 upgrade completed in 76 seconds (~13.5k epochs/sec, 217 matches/sec):

```
2021-02-02 03:17:14.116 [INF] DB: PostgreSQL 13.1 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 10.2.0, 64-bit
2021-02-02 03:17:14.123 [INF] DB: Upgrading DB scheme from 0 to 2
2021-02-02 03:17:14.123 [DBG] DB: Upgrading DB scheme to 1...
2021-02-02 03:17:14.123 [DBG] DB: Upgrading DB scheme to 2...
2021-02-02 03:17:14.123 [INF] DB: Populating epochs with volume data for market "dcr_btc" matches...
2021-02-02 03:17:14.126 [DBG] DB: Creating the "dcr_btc.epoch_reports" table.
2021-02-02 03:17:14.240 [INF] DB: Processing all 1030671 of the 10000 ms "dcr_btc" epochs from idx 160191348 to 161222018...
2021-02-02 03:17:14.872 [INF] DB:  - Processing epochs [160200000, 160250000)...
2021-02-02 03:17:18.509 [INF] DB:  - Processing epochs [160250000, 160300000)...
2021-02-02 03:17:22.174 [INF] DB:  - Processing epochs [160300000, 160350000)...
2021-02-02 03:17:25.845 [INF] DB:  - Processing epochs [160350000, 160400000)...
2021-02-02 03:17:29.505 [INF] DB:  - Processing epochs [160400000, 160450000)...
2021-02-02 03:17:33.144 [INF] DB:  - Processing epochs [160450000, 160500000)...
2021-02-02 03:17:36.809 [INF] DB:  - Processing epochs [160500000, 160550000)...
2021-02-02 03:17:40.479 [INF] DB:  - Processing epochs [160550000, 160600000)...
2021-02-02 03:17:44.125 [INF] DB:  - Processing epochs [160600000, 160650000)...
2021-02-02 03:17:47.832 [INF] DB:  - Processing epochs [160650000, 160700000)...
2021-02-02 03:17:51.571 [INF] DB:  - Processing epochs [160700000, 160750000)...
2021-02-02 03:17:55.256 [INF] DB:  - Processing epochs [160750000, 160800000)...
2021-02-02 03:17:58.942 [INF] DB:  - Processing epochs [160800000, 160850000)...
2021-02-02 03:18:02.642 [INF] DB:  - Processing epochs [160850000, 160900000)...
2021-02-02 03:18:06.300 [INF] DB:  - Processing epochs [160900000, 160950000)...
2021-02-02 03:18:09.970 [INF] DB:  - Processing epochs [160950000, 161000000)...
2021-02-02 03:18:13.883 [INF] DB:  - Processing epochs [161000000, 161050000)...
2021-02-02 03:18:17.599 [INF] DB:  - Processing epochs [161050000, 161100000)...
2021-02-02 03:18:21.324 [INF] DB:  - Processing epochs [161100000, 161150000)...
2021-02-02 03:18:25.116 [INF] DB:  - Processing epochs [161150000, 161200000)...
2021-02-02 03:18:28.902 [INF] DB:  - Processing epochs [161200000, 161222019)...
2021-02-02 03:18:30.528 [DBG] DB: Processed 16557 matches doing 1233360 in dcr volume (1667.51332 in btc volume)
2021-02-02 03:18:30.539 [INF] DB: Upgrades complete. DB is at version 2
2021-02-02 03:18:30.539 [INF] DB: Configuring 2 markets tables: [dcr_btc dcr_ltc]
```

Draft while I decide about `pgonline` tests.


server/db/driver/pg/internal/meta.go
```diff
@@ -3,11 +3,13 @@ package internal
 const (
 	// CreateMetaTable creates a table to hold DEX metadata.
 	CreateMetaTable = `CREATE TABLE IF NOT EXISTS %s (
```
itswisdomagain, line 5: The table name is hardcoded below in `CreateMetaRow, SelectDBVersion, SetDBVersion`. Do we want to do the same here? I see from usage that the table name is always set to the constant `metaTableName     = "meta"`.

server/db/driver/pg/tables.go
```diff
@@ -117,45 +117,57 @@ func CreateTable(db *sql.DB, schema, tableName string) (bool, error) {
...
-		log.Trace("Creating new markets table.")
+		// This is a fresh install. Setup the meta table, and create the markets
+		// and accounts tables.
+		_, err = db.Exec(internal.CreateMetaRow)
+		if err != nil {
+			return fmt.Errorf("failed to create row for meta table: %w", err)
+		}
+		err = setDBVersion(db, dbVersion) // skip upgrades
+		if err != nil {
+			return fmt.Errorf("failed to create row for meta table: %w", err)
```
itswisdomagain, line 136:
```suggestion
			return fmt.Errorf("failed to set db version in meta table: %w", err)
```

server/db/driver/pg/upgrades.go
```diff
@@ -0,0 +1,302 @@
...
+	"decred.org/dcrdex/dex/calc"
+	"decred.org/dcrdex/server/db/driver/pg/internal"
+)
+
+const dbVersion = 2
+
+// The number of upgrades defined MUST be equal to dbVersion.
+var upgrades = []func(db *sql.DB) error{
+	v1Upgrade,
+	v2Upgrade,
```
itswisdomagain, line 20: inline comments for what these upgrades do?

server/db/driver/pg/upgrades.go
```diff
@@ -0,0 +1,302 @@
...
+		targetVer := current + uint32(i) + 1
+		log.Debugf("Upgrading DB scheme to %d...", targetVer)
+		if err = up(db); err != nil {
+			return fmt.Errorf("failed to upgrade to db version %d: %w", targetVer, err)
+		}
+
+		if err = setDBVersion(db, targetVer); err != nil {
+			return fmt.Errorf("failed to set new DB version %d: %w", targetVer, err)
+		}
+	}
```
itswisdomagain, line 294: If setting the db version fails, we'll have an upgraded database with an older version number. Can avoid this by using a transaction for each upgrade, pass the transaction to each upgrade function, after the upgrade function returns without error, set the db version before committing. Any error should rollback all changes.

itswisdomagain (changes requested):

server/db/driver/pg/internal/meta.go
chappjc: Good call!  Hopefully no upgrades push what can be done in a single transaction.

server/db/driver/pg/upgrades.go
chappjc: Good call!  Hopefully no upgrades push what can be done in a single transaction.

server/db/driver/pg/internal/meta.go
chappjc: Maaaybe, but the only reason it's a placeholder in `CreateMetaTable` is to shoehorn it into the `tableMap` used by `CreateTable`, which expects each query to have this format with one specifier for the name of the table.  With meta in the public schema (default) and a name that isn't going to change, I'm inclined to keep the callign code for these other queries simple with no sprintf business.

server/db/driver/pg/upgrades.go
chappjc: I didn't encode it all in a transaction yet, but otherwise ready for review.  I'm actually thinking that `setDBVersion` failing is next to impossible though.

JoeGruffins (approved):
Code looks ok to me and was able to upgrade without problems.

server/db/driver/pg/pg.go
```diff
@@ -194,6 +194,10 @@ func NewArchiver(ctx context.Context, cfg *Config) (*Archiver, error) {
 		return nil, fmt.Errorf("unknown network %d", cfg.Net)
 	}
 
+	// TODO: remove feeKeyBranch and feeKeyBranch fields, and the CreateKeyEntry
+	// call. These are deprecated with the introduction of the new registration
+	// protocol, where the fee addresses are managed by the AuthManager.
+
```
chappjc, line 200: This snuck in by accident.  Removing

server/db/driver/pg/upgrades.go
```diff
@@ -0,0 +1,324 @@
...
+	// exists, we should not create it fresh with the current scheme since one
+	// or more subsequent upgrades may alter the meta scheme.
+	metaV0Stmt := `CREATE TABLE IF NOT EXISTS %s (state_hash BYTEA)`
+	metaCreated, err := createTable(db, metaV0Stmt, publicSchema, metaTableName)
+	if err != nil {
+		return fmt.Errorf("failed to create meta table: %w", err)
+	}
+	if metaCreated {
+		log.Infof("Created new %q table", metaTableName)    // from 0.2+pre master
+		_, err = db.Exec(`INSERT INTO meta DEFAULT VALUES`) // might be CreateMetaRow, but ping to v0 stmt
```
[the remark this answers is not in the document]
chappjc: Right, I made a comment explaining why I didn't though `// might be CreateMetaRow, but ping to v0 stmt`.  Typo there (ping -> pin), but the idea is that I don't know that `CreateMetaRow` might be different in higher versions, but for this v0-v1 upgrade, I want the current effect guaranteed and not worry about this upgrade breaking in the future.

server/db/driver/pg/upgrades.go
```diff
@@ -0,0 +1,324 @@
...
+}
+
+// v2Upgrade populates the epoch_reports table with historical data from the
+// matches table.
+func v2Upgrade(db *sql.DB) error {
+	mkts, err := loadMarkets(db, marketsTableName)
+	if err != nil {
+		return fmt.Errorf("failed to read markets table: %w", err)
+	}
+	tx, err := db.Begin()
```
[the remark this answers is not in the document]
chappjc: Hmm, that could work for cancellation since the Rollback would catch it.  This upgrade is expected to take several minutes on busy servers, so I don't think a timeout would be easy to pick.  Let me wire up ctx so the operator can bail out.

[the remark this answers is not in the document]
chappjc: That was a minor rabbit hole, but I addressed this and making the upgrade + version bump atomic with a sql.Tx in https://github.com/decred/dcrdex/pull/959/commits/6b4d7c73fb65966372f31e5b7344e47e2cd37aa4.  It's very hard to write a test for it, but it did verify the functionality of a rollback I believe.  I also re-tested the massive upgrade of the DB referenced in the PR description.

server/db/driver/pg/upgrades.go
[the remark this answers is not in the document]
buck54321: Oh. You mean that comment right there at the end of the same line that I completely ignored.

server/db/driver/pg/upgrades.go
```diff
@@ -0,0 +1,348 @@
...
+			return err
+		}
+	}
+
+	current, err = DBVersion(db)
+	if err != nil {
+		return fmt.Errorf("failed to get DB version: %w", err)
+	}
+	log.Infof("Upgrades complete. DB is at version %d", current)
+	return nil
```
buck54321, line 347: Going back to wisdom's comment

> If setting the db version fails, we'll have an upgraded database with an older version number. Can avoid this by using a transaction for each upgrade, pass the transaction to each upgrade function, after the upgrade function returns without error, set the db version before committing. Any error should rollback all changes.

I seems to me like the idea was to use a single transaction for all upgrades and the db version, and not commit the transaction until the db version is recorded successfully. But the wording "using a transaction for each upgrade" seems to conflict with that assumption too, so I guess I don't know. 

But wisdom's initial concern about an upgrade being performed but the version not being recorded has not been addressed, I don't think.

chappjc: I believe it is addressed here: https://github.com/decred/dcrdex/commit/6b4d7c73fb65966372f31e5b7344e47e2cd37aa4#diff-d5886fb14518c5d25332c7350f812301f9c6628f337111aa73258edac37c7472R309-R338

If anything with the actual upgrade or version setting fails, both rollback and you are at a consistent database level.  That it is perhaps at an upgraded version but not the best version isn't a problem.

chappjc: I suppose we could technically wrap the loop of all the upgrades in the same tx like you're thinking, but I'm getting a bit anxious about the limits of postgresql here. I wonder if it's gonna be ok to have say 10 versions all modifying the tables heavily.

chappjc: Ugh, dang I have to re-rig NewDEX for cancellation.  It doesn't work presently to cancel.

chappjc:
Upgrade on dex-test.ssgen.io with current commit went smoothly:

```
2021-03-26 14:57:28.896 [INF] DB:  - Processing epochs [161650000, 161677002)...                                                                                                                                                                                                                                
2021-03-26 14:57:35.740 [DBG] DB: Processed 578 matches doing 65460 in DCR volume (108.16776 in BTC volume)                                                                                                                                                                                                     
2021-03-26 14:57:35.743 [INF] DB: Upgrades complete. DB is at version 2  
```

Very few actual trade matches on testnet, but tons of epochs.

server/db/driver/pg/upgrades.go
buck54321: > I believe it is addressed here: 6b4d7c7#diff-d5886fb14518c5d25332c7350f812301f9c6628f337111aa73258edac37c7472R309-R338

Ah. I see what I did wrong. Carry on.

buck54321 (approved):

PR #962, opened 2021-02-03
client/{core,db}: refactor db.MetaMatch struct, remove duplicate fields

itswisdomagain:
Remove following duplicate fields:
- ~`matchTracker.id`~ => `matchTracker.Match.MatchID` (from `order.UserMatch` embedded via `db.MetaMatch`)
- ~`db.MetaMatch.MetaData.Status`~ => `db.MetaMatch.Status` (from embedded `order.UserMatch`); this change also reflects in ~`matchTracker.MetaData.Status`~ => `matchTracker.Status`, `matchTracker` embeds `db.MetaMatch`.

Having duplicate fields in different paths of the same object means both fields have to be simultaneously updated otherwise an error could result. Primary motivation for this is making it tougher to shoot ourselves in the foot in this regard.

The issue with the duplicate id fields came to the fore when I needed to use both `matchTracker.id` and `db.MetaMatch.Match.MatchID` fields in a recent modification to the core_tests (67e2c6c1d33e7d64302e572034a4c62cdf157714) and noticed I had to manually set both to the same value in different places.

Haven't ran into an issue with the duplicate status fields yet, but having a single updateable status field will come in handy in #957, where status updates may require updating the `match.Retired` field as well; updating the "wrong" status field could prevent a match from being retired. There exists a `db.MetaMatch.SetStatus()` method that should prevent potential issues by updating both status fields but since neither status fields are private, they can easily be modified without using the `SetStatus` method and we can proactively prevent this unwanted behaviour by keeping just a single status field.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -3213,7 +3213,7 @@ func (c *Core) authDEX(dc *dexConnection) error {
 		// Flag each of the missing matches as revoked.
 		for _, match := range missing {
 			c.log.Warnf("DEX %s did not report active match %s on order %s - assuming revoked.",
-				dc.acct.host, match.MatchID, oid)
+				dc.acct.host, match, oid)
 			// Must have been revoked while we were gone. Flag to allow recovery
 			// and subsequent retirement of the match and parent trade.
 			match.MetaData.Proof.SelfRevoked = true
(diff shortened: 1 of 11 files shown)


itswisdomagain:
I just remembered I have to add a commit for the db migration.

chappjc:
Some pretty extensive conflicts now with some of your work and others' merged.  Sorry.

Also, please consider if `tdb.updateMatchMtx.Unlock` and the new `TDB` methods are really required.  Please see the changes I made on the `release-0.1` branch do deal with the extra db match updates: https://github.com/decred/dcrdex/pull/973#issuecomment-778525810

itswisdomagain:
> Also, please consider if `tdb.updateMatchMtx.Unlock` and the new `TDB` methods are really required. Please see the changes I made on the `release-0.1` branch do deal with the extra db match updates: [#973 (comment)](https://github.com/decred/dcrdex/pull/973#issuecomment-778525810)

Apparently only needed the extra channel reads in 2 places. Increasing the channel buffer size is not even necessary as match updates are performed from separate goroutines which will only block until the channel is read in the test.

chappjc (approved):
All looks good. Very positive changes.  Just one suggestion, but it it doesn't feel right, you can pass on it.

chappjc:
Please also rebase and resolve the conflicts, but looks good to merge.  The rebase was not totally clean.  Here's what I ended up with on my end, but please rebase as you see fit: https://github.com/chappjc/dcrdex/commits/refactor-match-types

buck54321 (approved):

PR #965, opened 2021-02-04
client/core: resumeTrades auditInfo search must retry

chappjc:
This fixes a bug where resuming trades with required audit info can fail too readily.

I'm considered using `tickGovernor` instead of `swapErr` while searching in the background for the counterparty contract and audit info, but this turned out cleaner, and it's possible for wallet reconfigure to clear the `tickGovernor` unexpectedly.

This also avoids a bug where an order could be retired without saving the match revoked status to DB, making the order potentially load forever if there was a subsequent error such as funding coins not found.  This was caused by the self-revoke in `resumeTrades` not actually being stored to DB, just memory.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -3556,8 +3556,8 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
 				if dbMatch.Status < order.MakerSwapCast {
 					needsCoins = true
 				}
-				if dbMatch.Status < order.MakerRedeemed && dbMatch.Status >= order.TakerSwapCast {
-					needsAuditInfo = true
+				if dbMatch.Status == order.TakerSwapCast {
+					needsAuditInfo = true // maker needs AuditInfo for takers contract
 					counterSwap = metaData.Proof.TakerSwap
 				}
 			} else { // Taker
@@ -3565,7 +3565,7 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
 					needsCoins = true
 				}
 				if dbMatch.Status < order.MatchComplete && dbMatch.Status >= order.MakerSwapCast {
-					needsAuditInfo = true
+					needsAuditInfo = true // taker needs AuditInfo for maker's contract
 					counterSwap = metaData.Proof.MakerSwap
 				}
 			}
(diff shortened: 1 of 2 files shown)


client/core/trade.go
```diff
@@ -1950,6 +1954,13 @@ func (t *trackedTrade) auditContract(match *matchTracker, coinID []byte, contrac
 				tries++
 				return wait.TryAgain
 			}
+			// Even retry for unrecognized errors, at least for a little while.
+			// With a default recheckInterval of 5 seconds, this is 2 minutes.
+			if tries < 24 {
```
chappjc, line 1959: @buck54321 I made a change to have it go for a flat 2 minutes, but (a) that makes the unit tests go for 2 minutes where it expects audit to fail in certain cases and (b) unit tests are using a crazy short recheck interval so it's retries like 50 million times.  I just revised the comment instead.

client/core/core.go
```diff
@@ -3582,15 +3582,47 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
...
+				// mempool, or the wallet is otherwise not ready. Note that this
+				// does not actually audit the contract's value, recipient,
+				// expiration, or secret hash (if maker), as that was already
+				// done when it was initially stored as CounterScript.
 				auditInfo, err := wallets.toWallet.AuditContract(counterSwap, counterContract)
 				if err != nil {
-					c.log.Debugf("AuditContract error for match %v status %v, refunded = %v, revoked = %v: %v",
-						match.id, match.MetaData.Status, len(match.MetaData.Proof.RefundCoin) > 0, match.MetaData.Proof.IsRevoked(), err)
+					contractStr := coinIDString(wallets.toAsset.ID, counterSwap)
+					c.log.Warnf("Starting search for counterparty contract %v (%s)", contractStr, unbip(wallets.toAsset.ID))
```
chappjc, line 3593: I tested this for both btc and dcr contracts (mined and unmined) by restarting the wallets and nodes, also by forcing the goroutine search by changing the condition to `err == nil`.  I'm not sure if restarting the wallets and nodes forced it though because I added these two log statements after that but before the hacked code test.

chappjc: While the search is ongoing, trace logs say:

> 2021-02-04 15:02:20.396 [TRC] CORE: Match 9deab745f96c0cd7398ef1a09fab4e68c6627df2ccec402a352a8664e42fa714 not swappable: swapErr = audit in progress, please wait, revoked = false, metered = false
> 2021-02-04 15:02:20.396 [TRC] CORE: Match 9deab745f96c0cd7398ef1a09fab4e68c6627df2ccec402a352a8664e42fa714 not redeemable: swapErr = audit in progress, please wait, RefundCoin = , metered = false

tested that with 

```diff
diff --git a/client/core/core.go b/client/core/core.go
index a3d2cf61..7b149d07 100644
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -3588,7 +3588,7 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
                                // expiration, or secret hash (if maker), as that was already
                                // done when it was initially stored as CounterScript.
                                auditInfo, err := wallets.toWallet.AuditContract(counterSwap, counterContract)
-                               if err != nil {
+                               if err == nil {
                                        contractStr := coinIDString(wallets.toAsset.ID, counterSwap)
                                        c.log.Warnf("Starting search for counterparty contract %v (%s)", contractStr, unbip(wallets.toAsset.ID))
                                        // Start the audit retry waiter. Set swapErr to block tick
@@ -3597,6 +3597,7 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
                                        match.swapErr = fmt.Errorf("audit in progress, please wait") // don't frighten the users
                                        go func(tracker *trackedTrade, match *matchTracker) {
                                                auditInfo, err := tracker.searchAuditInfo(match, counterSwap, counterContract)
+                                               time.Sleep(time.Minute)
                                                tracker.mtx.Lock()
                                                defer tracker.mtx.Unlock()
                                                if err != nil {
```

client/core/core.go
```diff
@@ -3582,15 +3582,48 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
...
+				// mempool, or the wallet is otherwise not ready. Note that this
+				// does not actually audit the contract's value, recipient,
+				// expiration, or secret hash (if maker), as that was already
+				// done when it was initially stored as CounterScript.
 				auditInfo, err := wallets.toWallet.AuditContract(counterSwap, counterContract)
 				if err != nil {
-					c.log.Debugf("AuditContract error for match %v status %v, refunded = %v, revoked = %v: %v",
-						match.id, match.MetaData.Status, len(match.MetaData.Proof.RefundCoin) > 0, match.MetaData.Proof.IsRevoked(), err)
-					match.swapErr = fmt.Errorf("audit error, order %s, match %s: %w", tracker.ID(), match.id, err)
-					match.MetaData.Proof.SelfRevoked = true
```
chappjc, line 3590: This was the source of the secondary bug addressed by this PR.  Namely, by revoking it in memory, the order would be declared inactive and retired, but on restart this match would reload as active.  However, since the order was already retired and any change coins were returned, it was likely that the order would fail to load further down in `resumeTrades` when trying to find funding coins.  The result is seeing the funding coin error forever each time dexc is started.

client/core/core.go
```diff
@@ -3582,15 +3582,48 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
...
+							// If not already revoked, set it as self-revoked.
+							if match.MetaData.Proof.IsRevoked() {
+								return
+							}
+							match.MetaData.Proof.SelfRevoked = true
+							err = tracker.db.UpdateMatch(&match.MetaMatch)
+							if err != nil {
+								tracker.dc.log.Errorf("Error updating database for match %v: %v", match.id, err)
+							}
+							return
```
chappjc, line 3618: I'm considering if it is best NOT to self-revoke it at all.  With swapErr still set, there won't be any attempted swaps or redeems, but the user could restart dexc if they fixed something with their rig and it would only retry if we did not revoke.  The server will eventually send a revoke_match or the client will discover it revoked later on 'connect'...

client/core/core.go
```diff
@@ -3582,15 +3582,48 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
...
+						match.counterSwap = auditInfo
+						match.swapErr = nil // unblock tick actions
+						c.log.Infof("Successfully located and re-validated counterparty contract %v (%s)",
+							contractStr, unbip(wallets.toAsset.ID))
+					}(tracker, match)
+
+					continue // leave auditInfo nil
 				}
 				match.counterSwap = auditInfo
 				continue
```
buck54321, line 3629: A taker in `MakerSwapCast` will need both coins and audit info, but the `AuditInfo` error recovery path still `continue`s, so the coins will never be grabbed, and a match that is able to recover will error when trying to broadcast their swap, I think.

buck54321 (approved):

client/core/core.go
```diff
@@ -3582,15 +3582,40 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
...
+							// The match may become revoked by server.
+							return
+						}
+						match.counterSwap = auditInfo
+						match.swapErr = nil // unblock tick actions
+						c.log.Infof("Successfully located and re-validated counterparty contract %v (%s)",
+							contractStr, unbip(wallets.toAsset.ID))
+					}(tracker, match)
+
+					continue // leave auditInfo nil
```
chappjc, line 3618: Noticed that in my email a comment from you @buck54321:

> A taker in MakerSwapCast will need both coins and audit info, but the AuditInfo error recovery path still continues, so the coins will never be grabbed, and a match that is able to recover will error when trying to broadcast their swap, I think.

Did you delete the comment?  I suspect you noticed that this `continue` is in the matches loop and it still gets to the funding coins step, but I want to be sure.

buck54321: Yeah. I figured that out as soon as I hit send.

PR #967, opened 2021-02-05
client/{core,cmd/dexc}: load unfunded orders for recovery

chappjc:
This is part 2 of a `resumeTrades` issue resolution following https://github.com/decred/dcrdex/pull/965

This addresses the issue of orders that lose their funding coins one way or another (e.g. spend outside of dexc) can never be loaded.  Prior to this PR, that meant two things:

1. On login, the user will be forever spammed by an "Order coin error" /
  "Source coins retrieval error".
2. Matches for such orders can never be recovered, even if they do not
  require funding coins. This made recovery a manual task if there were
  matches in progress or requiring refund, etc.

Unfunded orders that require funding (epoch/booked or with matches needing to send swap txns) are explicitly blocked from negotiating new matches, and any offending matches that require funding coins are blocked with `swapErr`. However, notably, the `trackedTrade` is added to the `dc.trades` map, unlike before this change.  In addition to allowing unaffected matches to proceed, match status resolution can be performed for the order and its matches, potentially revoking it, and revoke_order/revoke_matches requests from the server will be recognized.

This allows recovery or completion of other unaffected matches belonging to the trade.  The order and the unfunded matches stay active (but halted via `swapErr`) until they are either revoked or the user possibly resolves a wallet issue that made the funding coins inaccessible and restarts dexc to try loading the funding coins again.  If the coins are not spent, they probably are using the wrong wallet that does not control the referenced coins.

Unfunded standing limit orders that are either epoch or booked are also canceled to prevent further matches that will be likely to fail.

> [INF] CORE: Loaded 1 active orders.
> [INF] CORE: Loaded 0 active match orders
> [ERR] CORE: notify: |ERROR| (order) Order coin error - Source coins retrieval error for order f9b17e7e (btc): funding coin not found: 9504587e6f92acb8515d0c12b8d8160367e78d8e5e8b947abdd81c532399c1e9:0 - Order: f9b17e7eb7b7185bb315bc1886a9309aac9821f928d475db2275d5762bc37c28
> 309aac9821f928d475db2275d5762bc37c28
> [WRN] CORE: Check the status of your BTC wallet and the coins logged above! Resolve the wallet issue if possible and restart the DEX client.
> [WRN] CORE: Unfunded order f9b17e7eb7b7185bb315bc1886a9309aac9821f928d475db2275d5762bc37c28 will be canceled on connect, but 0 active matches need funding coins!",
> [TRC] CORE: notify: |DATA| (order) Order loaded - Order: f9b17e7eb7b7185bb315bc1886a9309aac9821f928d475db2275d5762bc37c28
> [TRC] CORE: notify: |DATA| (balance) balance updated
> [INF] CORE: loaded 1 incomplete orders
> [DBG] CORE: Authenticated connection to 127.0.0.1:17273, acct e6e066dc1e41430b6dffd456914dbe58e8fc522743da03c6657e3831c7d0424c, 1 active orders, 0 active matches, score 0
> [TRC] CORE: Status reconciliation not required for order f9b17e7eb7b7185bb315bc1886a9309aac9821f928d475db2275d5762bc37c28, status "booked", server-reported status "booked"
> [WRN] CORE: Canceling unfunded standing limit order f9b17e7eb7b7185bb315bc1886a9309aac9821f928d475db2275d5762bc37c28
> [INF] CORE: Cancel order 7daaefb750e1ea48db4d3dea5b76158a8da4cea11804bf3c9d651239cd8e6d95 targeting order f9b17e7eb7b7185bb315bc1886a9309aac9821f928d475db2275d5762bc37c28 at 127.0.0.1:17273 has been placed
> [DBG] CORE: notify: |POKE| (order) Cancelling order - A cancel order has been submitted for order f9b17e7e - Order: f9b17e7eb7b7185bb315bc1886a9309aac9821f928d475db2275d5762bc37c28

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -237,16 +237,23 @@ func (dc *dexConnection) findOrder(oid order.OrderID) (tracker *trackedTrade, pr
 func (c *Core) tryCancel(dc *dexConnection, oid order.OrderID) (found bool, err error) {
 	tracker, _, _ := dc.findOrder(oid)
 	if tracker == nil {
-		return
+		return // false, nil
 	}
-	found = true
+	return true, c.tryCancelTrade(dc, tracker)
+}
+
+// tryCancelTrade attempts to cancel the order.
+func (c *Core) tryCancelTrade(dc *dexConnection, tracker *trackedTrade) error {
+	oid := tracker.ID()
+	if lo, ok := tracker.Order.(*order.LimitOrder); !ok || lo.Force != order.StandingTiF {
+		return fmt.Errorf("cannot cancel %s order %s that is not a standing limit order", tracker.Type(), oid)
+	}
 
 	tracker.mtx.Lock()
 	defer tracker.mtx.Unlock()
 
-	if lo, ok := tracker.Order.(*order.LimitOrder); !ok || lo.Force != order.StandingTiF {
-		err = fmt.Errorf("cannot cancel %s order %s that is not a standing limit order", tracker.Type(), oid)
-		return
+	if status := tracker.metaData.Status; status != order.OrderStatusEpoch && status != order.OrderStatusBooked {
+		return fmt.Errorf("order %v not cancellable in status %v", oid, status)
 	}
 
 	if tracker.cancel != nil {
@@ -255,8 +262,8 @@ func (c *Core) tryCancel(dc *dexConnection, oid order.OrderID) (found bool, err 
 		tracker.deleteStaleCancelOrder()
 
 		if tracker.cancel != nil {
-			err = fmt.Errorf("order %s - only one cancel order can be submitted per order per epoch. still waiting on cancel order %s to match", oid, tracker.cancel.ID())
-			return
+			return fmt.Errorf("order %s - only one cancel order can be submitted per order per epoch. "+
+				"still waiting on cancel order %s to match", oid, tracker.cancel.ID())
 		}
 	}
 
@@ -274,9 +281,9 @@ func (c *Core) tryCancel(dc *dexConnection, oid order.OrderID) (found bool, err 
 		},
 		TargetOrderID: oid,
 	}
-	err = order.ValidateOrder(co, order.OrderStatusEpoch, 0)
+	err := order.ValidateOrder(co, order.OrderStatusEpoch, 0)
 	if err != nil {
-		return
+		return err
 	}
 
 	// Create and send the order message. Check the response before using it.
@@ -284,20 +291,18 @@ func (c *Core) tryCancel(dc *dexConnection, oid order.OrderID) (found bool, err 
 	var result = new(msgjson.OrderResult)
 	err = dc.signAndRequest(msgOrder, route, result, DefaultResponseTimeout)
 	if err != nil {
-		return
+		return err
 	}
 	err = validateOrderResponse(dc, result, co, msgOrder)
 	if err != nil {
-		err = fmt.Errorf("Abandoning order. preimage: %x, server time: %d: %w",
+		return fmt.Errorf("Abandoning order. preimage: %x, server time: %d: %w",
 			preImg[:], result.ServerTime, err)
-		return
 	}
 
 	// Store the cancel order with the tracker.
 	err = tracker.cancelTrade(co, preImg)
 	if err != nil {
-		err = fmt.Errorf("error storing cancel order info %s: %w", co.ID(), err)
-		return
+		return fmt.Errorf("error storing cancel order info %s: %w", co.ID(), err)
 	}
 
 	// Now that the trackedTrade is updated, sync with the preimage request.
@@ -317,8 +322,7 @@ func (c *Core) tryCancel(dc *dexConnection, oid order.OrderID) (found bool, err 
 		Order: co,
 	})
 	if err != nil {
-		err = fmt.Errorf("failed to store order in database: %w", err)
-		return
+		return fmt.Errorf("failed to store order in database: %w", err)
 	}
 
 	c.log.Infof("Cancel order %s targeting order %s at %s has been placed",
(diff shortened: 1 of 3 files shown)


chappjc:
Note that there are two better changes that can be made eventually/alternatively:

1. allow orders to change their funding coins, but there still needs to be handling for the case when insufficient funding coins are available
2. client can query the server earlier in resumption (in resumeTrades, before they are even added to the trades map) so it can automatically revoke these.  Presently order status and match status resolution don't happen until the orders are put in the trades map

chappjc:
Taking this back to draft because it is evident that unfunded order still need to be added to the trades map.  That is needed for any active matches that do not need funding coins anymore (post-swap-sent) as well as to allow status resolutions to run for this order and its matches.

chappjc:
Major change to this PR.  The dexc/Core option is removed, and unfunded orders are now loaded into the trades map so that unaffected matches can be processed as usual, and the order and affected matches can be revoked via status resolution or server requests.

client/core/core.go
```diff
@@ -3178,10 +3209,10 @@ func (c *Core) initialize() {
 	// for authentication when Login is triggered.
 	wg.Wait()
 	c.log.Infof("Successfully connected to %d out of %d DEX servers", len(c.conns), len(accts))
-	for dexName, dc := range c.dexConnections() {
+	for _, dc := range c.dexConnections() {
```
chappjc, line 3212: Minor bug I spotted - it is now a slice range so the first variable was no longer the dex name string but an integer

client/core/core.go
```diff
@@ -3622,39 +3682,47 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
...
-				byteIDs = append(byteIDs, []byte(cid))
-			}
-			if len(byteIDs) == 0 {
-				notifyErr(SubjectOrderCoinError, "No coins for loaded order %s %s: %v", unbip(wallets.fromAsset.ID), tracker.token(), err)
-				continue
-			}
-			coins, err := wallets.fromWallet.FundingCoins(byteIDs)
-			if err != nil {
-				notifyErr(SubjectOrderCoinError, "Source coins retrieval error for %s %s: %v", unbip(wallets.fromAsset.ID), tracker.token(), err)
-				continue
```
chappjc, line 3647: Removal of these `continue`s are the crux of this change, allowing this `trackedTrade` to enter the trades map.

client/core/core.go
```diff
@@ -3695,7 +3763,7 @@ func generateDEXMaps(host string, cfg *msgjson.ConfigResult) (map[uint32]*dex.As
 }
 
 // runMatches runs the sorted matches returned from parseMatches.
-func (c *Core) runMatches(dc *dexConnection, tradeMatches map[order.OrderID]*serverMatches) (assetMap, error) {
+func (c *Core) runMatches(tradeMatches map[order.OrderID]*serverMatches) (assetMap, error) {
```
chappjc, line 3766: The `dexConnection` input became unused when `refreshUser` was removed.

client/core/trade.go
```diff
@@ -231,6 +231,11 @@ func (t *trackedTrade) coreOrderInternal() *Order {
 	return corder
 }
 
+// hasFundingCoins indicates if either funding or change coins are locked.
+func (t *trackedTrade) hasFundingCoins() bool {
```
buck54321, line 235: Should doc that `hasFundingCoins` must be called with the mutex locked.

client/core/core.go
```diff
@@ -3581,14 +3637,15 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
 		// If matches haven't redeemed, but the counter-swap has been received,
 		// reload the audit info.
 		isActive := tracker.metaData.Status == order.OrderStatusBooked || tracker.metaData.Status == order.OrderStatusEpoch
-		var needsCoins bool
+		var matchesNeedingCoins []*matchTracker
 		for _, match := range tracker.matches {
 			dbMatch, metaData := match.Match, match.MetaData
-			var needsAuditInfo bool
+			var needsCoins, needsAuditInfo bool
```
buck54321, line 3643: `needsCoins` is only used for logging now, it appears. Could just use `len(matchesNeedingCoins) > 0`.

client/core/core.go
```diff
@@ -3154,6 +3160,31 @@ func (c *Core) authDEX(dc *dexConnection) error {
...
+			brokenTrades = append(brokenTrades, trade)
+		}
+		trade.mtx.RUnlock()
+	}
+	dc.tradeMtx.RUnlock()
+	for _, trade := range brokenTrades {
+		c.log.Warnf("Canceling unfunded standing limit order %v", trade.ID())
+		if err = c.tryCancelTrade(dc, trade); err != nil {
+			c.log.Warnf("Unable to cancel unfunded trade %v: %v", trade.ID(), err)
+		}
```
buck54321, line 3185: Would be good to get some test coverage on this.

buck54321 (changes requested):

client/core/core.go
chappjc: Good call. Coverage added in `TestLogin`.

client/core/core.go
chappjc: Resolved separately in https://github.com/decred/dcrdex/pull/977 by @JoeGruffins

JoeGruffins (approved):

chappjc:
Thanks, @JoeGruffins, I had forgotten about this PR since the backport was already merged.  I'm still catching up on review.

client/core/core.go
```diff
@@ -237,16 +237,23 @@ func (dc *dexConnection) findOrder(oid order.OrderID) (tracker *trackedTrade, pr
...
-	found = true
 
 	tracker.mtx.Lock()
 	defer tracker.mtx.Unlock()
 
-	if lo, ok := tracker.Order.(*order.LimitOrder); !ok || lo.Force != order.StandingTiF {
-		err = fmt.Errorf("cannot cancel %s order %s that is not a standing limit order", tracker.Type(), oid)
-		return
+	if status := tracker.metaData.Status; status != order.OrderStatusEpoch && status != order.OrderStatusBooked {
+		return fmt.Errorf("order %v not cancellable in status %v", oid, status)
```
chappjc, line 256: We removed this from the `release-0.1` backport (already merged), but it's still on this master PR.  What do you say to leaving @buck54321?

PR #971, opened 2021-02-08
scripts: force bash with shebang.

### LINKED ISSUE #951: harnesses: force bash with shebang or make posix compatible
The decision is between:

 1. Make all the scripts specify bash in their shebang (`#!/usr/bin/env bash`) and freely use bashisms.
 2. Stick with `#!/bin/sh` and ensure the scripts are always POSIX-compatible.

Door number 1 makes people use bash of course, but we don't have to consider other shells in making changes.
Door number 2 restricts us a bit more and perhaps creates a testing burden.  Do we test every harness change on multiple shells or with a POSIX only mode (`bash --posix`)?

Vote or comment.

amassarwi:
This makes all the scripts specify bash in their shebang - using `#!/usr/bin/env bash`.
I still needed to change the scripts to use `nolog` option instead of `history` to make run on macOS.

Closes #951.

### DIFF
--- a/dex/testing/btc/base-harness.sh
+++ b/dex/testing/btc/base-harness.sh
@@ -26,6 +26,8 @@ WAIT="wait-for ${SYMBOL}"
 
 SESSION="${SYMBOL}-harness"
 
+export SHELL=$(which bash)
+
 ################################################################################
 # Load prepared wallet if the files exist.
 ################################################################################
@@ -39,7 +41,7 @@ else
   mkdir -p "${BETA_DIR}"
 fi
 
-cd ${NODES_ROOT} && tmux new-session -d -s $SESSION
+cd ${NODES_ROOT} && tmux new-session -d -s $SESSION $SHELL
 
 ################################################################################
 # Write config files.
@@ -86,7 +88,7 @@ sleep 3
 # Setup the beta node.
 ################################################################################
 
-tmux new-window -t $SESSION:1 -n 'beta'
+tmux new-window -t $SESSION:1 -n 'beta' $SHELL
 tmux send-keys -t $SESSION:1 "set +o history" C-m
 tmux send-keys -t $SESSION:1 "cd ${BETA_DIR}" C-m
 
@@ -101,7 +103,7 @@ sleep 3
 # Setup the harness-ctl directory
 ################################################################################
 
-tmux new-window -t $SESSION:2 -n 'harness-ctl'
+tmux new-window -t $SESSION:2 -n 'harness-ctl' $SHELL
 tmux send-keys -t $SESSION:2 "set +o history" C-m
 tmux send-keys -t $SESSION:2 "cd ${HARNESS_DIR}" C-m
 sleep 1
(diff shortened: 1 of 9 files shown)


chappjc:
Oh I thought the nolog change was because it was actually using zsh on your machine.  Still doesn't like `history` even with the shebang changes?

amassarwi:
Yep thought the same, but now I think that I had still the nolog change when I tried to run with the shebang change.

JoeGruffins (approved):

dex/testing/btc/base-harness.sh
```diff
@@ -73,7 +73,7 @@ EOF
 ################################################################################
 
 tmux rename-window -t $SESSION:0 'alpha'
-tmux send-keys -t $SESSION:0 "set +o history" C-m
+tmux send-keys -t $SESSION:0 "set +o nolog" C-m
```
chappjc, line 76: Two things:
1. `nolog` has the opposite semantics from `history`, so to stop recording history it would use a `-o nolog` instead of `+o`.
2. `nolog` is apparently ignored by bash, so I don't understand why `history` isn't working for you.  Can you double check this and the bash version you have?

Snip from my bash man page:

```
                      noglob  Same as -f.
                      nolog   Currently ignored.
                      notify  Same as -b.
                      nounset Same as -u.
```

And from https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html:

```
noglob

    Same as -f.
nolog

    Currently ignored.
notify

    Same as -b.
```

buck54321 (changes requested):
Something is busted with the `for` loops, because seeing about 60 of these 

```
./alpha sendmany default '{"SsjTp2QaT8qkPGWKKSeFi4DtaZsgkoPbgXt":24,"SsYW5LPmGCvvHuWok8U9FQu1kotv8LpvoEt":24,"Ssge52jCzbixgFC736RSTrwAnvH3a4hcPRX":24}'; tmux wait-for -S donedcr
9e1c7703dfd5e5275d2e489e3be2b4895c7bcf5ecb1af1c8a2556e1fba1547ef
```
in the wallet harnesses.

chappjc:
> Something is busted with the `for` loops, because I'm seeing about 60 of these
> 
> ```
> ./alpha sendmany default '{"SsjTp2QaT8qkPGWKKSeFi4DtaZsgkoPbgXt":24,"SsYW5LPmGCvvHuWok8U9FQu1kotv8LpvoEt":24,"Ssge52jCzbixgFC736RSTrwAnvH3a4hcPRX":24}'; tmux wait-for -S donedcr
> 9e1c7703dfd5e5275d2e489e3be2b4895c7bcf5ecb1af1c8a2556e1fba1547ef
> ```
> 
> in the wallet harnesses.

I think that's right.  I added these sends in a loop some time ago to prep a ton of utxos.

dex/testing/btc/base-harness.sh
```diff
@@ -39,7 +41,7 @@ else
   mkdir -p "${BETA_DIR}"
 fi
 
-cd ${NODES_ROOT} && tmux new-session -d -s $SESSION
+cd ${NODES_ROOT} && tmux new-session -d -s $SESSION $SHELL
```
chappjc, line 44: Having discussed this quite a bit in matrix chat with @amassarwi, I just want to note that this is the `shell-command` argument of tmux's `new-session` and `new-window` commands.  Without this, it seems to use the user's *login* shell instead of the currently running shell.
And while tmux docs claim it will use the shell in the `SHELL` environment variable for `shell-command`, @amassarwi found that this was not accurate.

buck54321 (approved):

chappjc (approved):

PR #911, opened 2021-01-01
client/core: async init/redeem requests

itswisdomagain:
Resolves #863.

PR #973, opened 2021-02-12
[0.1.5] client/core: store swap/redeem immediately after bcast

chappjc:
A minimal version of https://github.com/decred/dcrdex/pull/911.

This changes `Core` so that swap and redeem data is recorded in the DB immediately after the swap or redeem is broadcast, rather than after the `'init'` or `'redeem'` requests completes (or times out), which can take minutes.  When `Core` was killed ungracefully in this state, there is no recovery for certain matches.

The requests are still synchronous so users may experience long delays and perceived hangs when trying to quit (may have to wait a ~90 sec or so for a response timeout), but if they do a hard kill of the process, at least the DB will be consistent with the transactions they have broadcasted.

### DIFF
--- a/client/core/core_test.go
+++ b/client/core/core_test.go
@@ -2728,7 +2728,7 @@ func TestTradeTracking(t *testing.T) {
 	auditInfo.expiration = matchTime.Add(tracker.lockTimeTaker)
 
 	// success, full handleAuditRoute>processAuditMsg>auditContract
-	rig.db.updateMatchChan = make(chan order.MatchStatus, 1)
+	rig.db.updateMatchChan = make(chan order.MatchStatus, 2)
 	err = handleAuditRoute(tCore, rig.dc, msg)
 	if err != nil {
 		t.Fatalf("audit error: %v", err)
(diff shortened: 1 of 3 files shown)


chappjc:
@itswisdomagain I did this on `release-0.1` branch, and I think I realized why you had trouble with the match update channel in core_test.go.  Three things I had to address: (1) it needed a larger buffer in two spots to accommodate an additional send otherwise the tests can deadlock, and (2) there would be a race at the end of some test when `nil`ing the channel if there were additional sends not accounted for, and (3) the test cases were spilling over data structures because there was no clean up of the test rigs between the tests.  I believe it's addressed here now, although this isn't going on master.

client/core/core_test.go
```diff
@@ -765,15 +766,28 @@ func newTestRig() *testRig {
...
+	}()
+
+	dc, conn, acct := testDexConnection(ctx)
 
-	dc, conn, acct := testDexConnection()
+	shutdown := func() {
+		cancel()
+		wg.Wait()
+		dc.connMaster.Wait()
+	}
```
chappjc, line 783: Similar changes in PR https://github.com/decred/dcrdex/pull/967.  On that work I had noticed dozens of latency queues ticking away in the later test cases.

client/core/core_test.go
```diff
@@ -2680,7 +2714,7 @@ func TestTradeTracking(t *testing.T) {
 	auditInfo.expiration = matchTime.Add(tracker.lockTimeTaker)
 
 	// success, full handleAuditRoute>processAuditMsg>auditContract
-	rig.db.updateMatchChan = make(chan order.MatchStatus, 1)
+	rig.db.updateMatchChan = make(chan order.MatchStatus, 2)
```
chappjc, line 2717: One of a few larger test buffers to capture the extra synchronous send without deadlock.

client/core/core_test.go
```diff
@@ -2825,6 +2864,7 @@ func TestTradeTracking(t *testing.T) {
 	if len(proof.TakerSwap) == 0 {
 		t.Fatalf("swap not broadcast with confirmations")
 	}
+	<-rig.db.updateMatchChan // initsig update after init response
```
chappjc, line 2867: One of a few new receives for the auth.InitSig and auth.Redeem sig that come after the requests.

chappjc:
I have NOT tested with trade_simnet_test.go or interactively with dexc.

chappjc:
I'm open to instead grabbing @itswisdomagain's async init/redeem commit instead, but I have some reservations about going that far for a 0.1 patch.

I've used this PR a fair bit on testnet.

buck54321 (approved):

chappjc:
Please have a look @itswisdomagain as soon as convenient for you.  Would like to merge in next 24 hrs.

PR #974, opened 2021-02-13
client/core: fix concurrent conns map write

chappjc:


client/core/core.go
```diff
@@ -2404,13 +2404,15 @@ func (c *Core) initializeDEXConnections(crypter encrypt.Crypter) []*DEXBrief {
 			disabledAccountHosts = append(disabledAccountHosts, disabledAccountHost)
 		}
 		if len(disabledAccountHosts) > 0 {
+			c.connMtx.Lock()
```
chappjc, line 2407: This was removed when the range below started using `c.dexConnections()`, but this was in different goroutine.  Repeated Logins with disabled accounts hit this.

JoeGruffins (approved):

PR #975, opened 2021-02-14
server: use decred's rpcclient for BTC

### LINKED ISSUE #841: use decred's rpcclient for BTC with all requests going through RawRequest

amassarwi:
That's the second part of and closes #841, https://github.com/decred/dcrdex/pull/938 was the first part and it
migrated the client to use decred's v6 context aware rpcclient.


server/asset/btc/btc_test.go
```diff
@@ -303,73 +303,77 @@ func txOutID(txHash *chainhash.Hash, index uint32) string {
...
+		blockHash, err := chainhash.NewHashFromStr(blkHash)
+		if err != nil {
+			return nil, fmt.Errorf("rawrequest: %v method param unmarshal error: %v",
+				method, err)
+		}
+		block, found := testChain.blocks[*blockHash]
+		if !found {
+			return nil, fmt.Errorf("test block not found")
+		}
+		return json.Marshal(block)
```
chappjc, line 331: There are places where we have hacks to modify the `GetBlockVerboseResult` in the `testchain.blocks` map, however a workaround is needed because with this stub, it marshals to json, then the new `(*rpcClient).GetBlockVerbose` uses `call` to unmarshal into a *new* `*GetBlockVerboseResult`, not the same one that's in the cache.

So for example in `TestReorg` > `setSidechainConfs` we have:

```go
			blk, err := btc.node.GetBlockVerbose(hash)
			if err != nil {
				t.Fatalf("error retrieving sidechain block to set confirmations: %v", err)
			}
			// Set Confirmations
			btc.blockCache.mtx.Lock() // read from (*blockCache).add in cache.go
			testChainMtx.Lock()       // field of testChain
			blk.Confirmations = -1
			testChainMtx.Unlock()
			btc.blockCache.mtx.Unlock()
```

Instead, we need to be setting the `Confirmations` field of the blocks in the `testchain` map.  I think this can be done directly (always could have), but let me have a look...

server/asset/btc/btc_test.go
```diff
@@ -303,73 +303,77 @@ func txOutID(txHash *chainhash.Hash, index uint32) string {
...
+	case methodGetBestBlockHash:
+		testChainMtx.RLock()
+		defer testChainMtx.RUnlock()
+		bbHash := testBestBlock.hash
+		return json.Marshal(bbHash.String())
+	case methodGetBlock:
+		testChainMtx.RLock()
+		defer testChainMtx.RUnlock()
+		var blkHash string
+		_ = json.Unmarshal(params[0], &blkHash)
```
chappjc, line 321: I suggest either checking this error, or if we never expect an error (probably this), replace with something like `mustUnmarshal(params[0], &blkHash)` where you use a helper:

```go

func mustUnmarshal(data []byte, thing interface{}) {
	if err := json.Unmarshal(data, thing); err != nil {
		panic(err.Error())
	}
}
```

amassarwi: 💯

server/asset/btc/btc_test.go
amassarwi: Riiiight, good catch, let me try set that directly on the testchain map.

JoeGruffins (commented):
I had the simnet tests fail a few times while testing this, but not sure if related to these changes.

server/asset/btc/btc.go
```diff
@@ -180,12 +160,13 @@ func (btc *Backend) Connect(ctx context.Context) (*sync.WaitGroup, error) {
 		return nil, fmt.Errorf("error creating %q RPC client: %w", btc.name, err)
 	}
 
-	// Setting the client field will enable shutdown
-	btc.client = client
-	btc.node = client
+	btc.node = &rpcClient{
+		ctx:       ctx,
+		requester: client,
+	}
```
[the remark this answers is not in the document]
JoeGruffins: I wish I has said `newClient` now, but a comment is also fine for the limited use this will have.

amassarwi:
@chappjc had to adjust: https://github.com/decred/dcrdex/pull/975/commits/8d431649ec106fb7c7a74e2f5ba9ab90dc3b42a3 after rebasing onto master.

server/asset/bch/bch.go
```diff
@@ -107,8 +107,8 @@ func (bch *BCHBackend) Contract(coinID []byte, redeemScript []byte) (*asset.Cont
 
 // estimateFee estimates the network transaction fee rate using the estimatefee
 // RPC.
-func estimateFee(node btc.BTCNode) (uint64, error) {
-	resp, err := node.RawRequest("estimatefee", nil)
+func estimateFee(node btc.RPCClient) (uint64, error) {
+	resp, err := node.Requester.RawRequest(nil, "estimatefee", nil)
```
JoeGruffins: I'm not %100 sure, but it looks like you could just change estimateFee to take the Requester and you would then not have to export btc.RPCClient as well.

JoeGruffins (commented):
Does cloneCfg.FeeEstimator/estimateFee's signature work as `estimateFee func(context.Context, RawRequester) (uint64, error)`? I'm not sure but maybe this would avoid a lot of exporting?

amassarwi:
@JoeGruffins Thanks for the suggestion, was sure it's not idiomatic.

chappjc:
You could also define a wrapper for callers that are not context-enabled so that it just uses the btc.Backend's subsystem shutdown context

```go
func (rc *RPCClient) RawRequest(string, []json.RawMessage) (json.RawMessage, error){
    return rc.Requester.RawRequest(rc.ctx, method, params)
}
```

amassarwi:
@JoeGruffins @chappjc updated

JoeGruffins (approved):
Works for me.

server/asset/btc/btc.go
```diff
@@ -219,12 +197,14 @@ func (btc *Backend) Connect(ctx context.Context) (*sync.WaitGroup, error) {
 		return nil, fmt.Errorf("error creating %q RPC client: %w", btc.name, err)
 	}
 
-	// Setting the client field will enable shutdown
-	btc.client = client
-	btc.node = client
+	// Requester must be set otherwise there can be a panic on shutdown.
```
buck54321, line 200: > there can be a panic on shutdown

If you're talking about `(*Backend).shutdown`, then I don't think that's true, since you check `btc.node != nil`

server/asset/btc/btc.go
```diff
@@ -107,7 +86,7 @@ type Backend struct {
 	// use the provided logger.
 	log         dex.Logger
 	decodeAddr  dexbtc.AddressDecoder
-	estimateFee func(BTCNode) (uint64, error)
+	estimateFee func(RPCClient) (uint64, error)
```
buck54321, line 89: ~~`RPCClient`~~ -> `*RPCClient`

server/asset/btc/rpcclient.go
```diff
@@ -0,0 +1,142 @@
...
+	methodGetBestBlockHash  = "getbestblockhash"
+	methodGetBlockchainInfo = "getblockchaininfo"
+	methodEstimateSmartFee  = "estimatesmartfee"
+	methodGetTxOut          = "gettxout"
+	methodGetRawTransaction = "getrawtransaction"
+	methodGetBlock          = "getblock"
+)
+
+// RawRequester defines dcrd's rpcclient RawRequest func where all RPC
+// requests sent through. For testing, it can be satisfied by a stub.
```
buck54321, line 26: > RawRequester defines dcrd's rpcclient RawRequest func

The interface doesn't **define** rpcclient's `RawRequest`, and `RawRequest` is not a function. 

How about "RawRequester is for sending context-aware RPC requests, and has methods for shutting down the underlying connection."

server/asset/btc/rpcclient.go
```diff
@@ -0,0 +1,142 @@
...
+	chainInfo := new(GetBlockchainInfoResult)
+	err := rc.call(methodGetBlockchainInfo, nil, chainInfo)
+	if err != nil {
+		return nil, err
+	}
+	return chainInfo, nil
+}
+
+// EstimateSmartFee requests the server to estimate a fee level based on the
+// given parameters.
```
buck54321, line 76: The general statement `based on the given parameters` could be appended to the end of almost any docstring. It doesn't add anything, really. If parameters are important enough to mention, you should be specific about the parameters you think are worth mentioning, e.g. `based on a target confirmation level`. If there are none worth mentioning specifically, then you don't need to mention parameters at all. It's okay to let a function signature with well-named parameters speak for itself.

server/asset/btc/rpcclient.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// GetRawTransaction retrieves tx's information.
+func (rc *RPCClient) GetRawTransaction(txHash *chainhash.Hash) (*btcutil.Tx, error) {
+	res := new(btcutil.Tx)
+	return res, rc.call(methodGetRawTransaction, anylist{txHash.String()}, res)
+}
+
+// GetRawTransactionVerbose retrieves tx's information with verbose flag set
+// to true.
```
buck54321, line 99: How about "GetRawTransactionVerbose retrieves the verbose tx information" ?

server/asset/btc/rpcclient.go
```diff
@@ -0,0 +1,142 @@
...
+func (rc *RPCClient) GetRawTransactionVerbose(txHash *chainhash.Hash) (*btcjson.TxRawResult, error) {
+	res := new(btcjson.TxRawResult)
+	return res, rc.call(methodGetRawTransaction, anylist{txHash.String(),
+		true}, res)
+}
+
+// GetBlockVerbose returns a data structure from the server with information
+// about a block given its hash.
```
buck54321, line 107: How about "GetBlockVerbose fetches verbose block data for the specified hash." ?

buck54321 (changes requested):

server/asset/btc/btc.go
amassarwi: ⚡

server/asset/btc/rpcclient.go
amassarwi: Agreed.

server/asset/btc/rpcclient.go
amassarwi: Yep, reads much better

chappjc (approved):

PR #977, opened 2021-02-15
core: Fix active orders log.

JoeGruffins:
It looks like this was changed from a map to slice at some point and the log here was overlooked.


chappjc (approved):
Thanks for this.  I picked up on it as well in https://github.com/decred/dcrdex/pull/967#discussion_r574175540, but it's worth a separate fix.

PR #978, opened 2021-02-16
dexc: change application password feature.

### LINKED ISSUE #678: app: reset password and change password features
- Forgot password -> Reset password feature. Reset password requests should probably be authorized somehow, perhaps by providing a recovery phrase that the user can back up on dexc setup. Should likely require user to re-provide wallet passwords.

- Change password feature for logged in users. Should not require re-supplying wallet passwords.

amassarwi:
Closes #678

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1568,7 +1568,7 @@ func (c *Core) loadWallet(dbWallet *db.Wallet) (*xcWallet, error) {
 			OrderLocked:    orderLockedAmt,
 			ContractLocked: contractLockedAmt,
 		},
-		encPW:   dbWallet.EncryptedPW,
+		encPass: dbWallet.EncryptedPW,
 		address: dbWallet.Address,
 		dbID:    dbWallet.ID(),
 	}, nil
@@ -1664,6 +1664,96 @@ func (c *Core) WalletSettings(assetID uint32) (map[string]string, error) {
 	return dbWallet.Settings, nil
 }
 
+func (c *Core) setAppPass(pw []byte) (encrypt.Crypter, error) {
+	if len(pw) == 0 {
+		return nil, fmt.Errorf("empty password not allowed")
+	}
+
+	crypter := c.newCrypter(pw)
+	err := c.db.Store(keyParamsKey, crypter.Serialize())
+	if err != nil {
+		return nil, fmt.Errorf("error storing key parameters: %w", err)
+	}
+
+	return crypter, nil
+}
+
+func (c *Core) updateWalletsEncPW(oldCrypter, newCrypter encrypt.Crypter) error {
+	updateWalletPW := func(w *xcWallet) error {
+		// If encrypted password set, decrypt using old app pw then encrypt again
+		// using new app pw.
+		oldEncPW := w.encPW()
+		if len(oldEncPW) == 0 {
+			return nil
+		}
+
+		// Decrypt with old app pw.
+		pwB, err := oldCrypter.Decrypt(oldEncPW)
+		if err != nil {
+			return err
+		}
+		// Re-encrypt with new app pw.
+		encPW, err := newCrypter.Encrypt(pwB)
+		if err != nil {
+			return err
+		}
+		// Store encrypted pw.
+		w.setEncPW(encPW)
+		err = c.db.SetWalletPassword(w.dbID, encPW)
+		if err != nil {
+			return codedError(dbErr, err)
+		}
+		return nil
+	}
+
+	for _, w := range c.xcWallets() {
+		err := updateWalletPW(w)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// ChangeAppPass updates the application password to the provided new password
+// after validating the current password.
+func (c *Core) ChangeAppPass(appPW, newAppPW []byte) error {
+	// Validate current password.
+	oldCrypter, err := c.encryptionKey(appPW)
+	if err != nil {
+		return newError(authErr, "ChangeAppPass password error: %v", err)
+	}
+	// Set new password.
+	newCrypter, err := c.setAppPass(newAppPW)
+	if err != nil {
+		return err
+	}
+	// Update xcWallets' encrypted passwords.
+	err = c.updateWalletsEncPW(oldCrypter, newCrypter)
+	if err != nil {
+		return fmt.Errorf("failed to update encrypted wallet password: %w", err)
+	}
+	// Update dexConnections' dexAccount key encryption.
+	c.connMtx.Lock()
+	defer c.connMtx.Unlock()
+	for _, dc := range c.conns {
+		// Re-encrypt private keys.
+		err = dc.acct.updateKeysEncryption(oldCrypter, newCrypter)
+		if err != nil {
+			return fmt.Errorf("ChangeAppPass error updating account's keys"+
+				" encryption: %v", err)
+		}
+		// Update account's db entry.
+		err = c.db.UpdateAccount(dc.acct.dbInfo())
+		if err != nil {
+			return codedError(dbErr, err)
+		}
+	}
+
+	return nil
+}
+
 // ReconfigureWallet updates the wallet configuration settings, it also updates
 // the password if newWalletPW is non-nil. Do not make concurrent calls to
 // ReconfigureWallet for the same asset.
@@ -1681,7 +1771,7 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg 
 		AssetID:     oldWallet.AssetID,
 		Settings:    cfg,
 		Balance:     &db.Balance{}, // in case retrieving new balance after connect fails
-		EncryptedPW: oldWallet.encPW,
+		EncryptedPW: oldWallet.encPW(),
 		Address:     oldWallet.currentDepositAddress(),
 	}
 	// Reload the wallet with the new settings.
@@ -1840,7 +1930,7 @@ func (c *Core) setWalletPassword(wallet *xcWallet, newPW []byte, crypter encrypt
 	// otherwise does not require a password. Perhaps an
 	// asset.Wallet.RequiresPassword wallet method?
 	if newPasswordSet {
-		// Encrypt password if it's not an empty string
+		// Encrypt password if it's not an empty string.
 		encNewPW, err := crypter.Encrypt(newPW)
 		if err != nil {
 			return newError(encryptionErr, "encryption error: %v", err)
@@ -1850,12 +1940,12 @@ func (c *Core) setWalletPassword(wallet *xcWallet, newPW []byte, crypter encrypt
 			return newError(authErr,
 				"setWalletPassword unlocking wallet error, is the new password correct?: %v", err)
 		}
-		wallet.encPW = encNewPW
+		wallet.setEncPW(encNewPW)
 	} else {
-		wallet.encPW = nil
+		wallet.setEncPW(nil)
 	}
 
-	err := c.db.SetWalletPassword(wallet.dbID, wallet.encPW)
+	err := c.db.SetWalletPassword(wallet.dbID, wallet.encPW())
 	if err != nil {
 		return codedError(dbErr, err)
 	}
@@ -2192,16 +2282,9 @@ func (c *Core) InitializeClient(pw []byte) error {
 		return fmt.Errorf("already initialized, login instead")
 	}
 
-	if len(pw) == 0 {
-		return fmt.Errorf("empty password not allowed")
-	}
-
-	crypter := c.newCrypter(pw)
-	err := c.db.Store(keyParamsKey, crypter.Serialize())
-	if err != nil {
-		return fmt.Errorf("error storing key parameters: %w", err)
-	}
-	return nil
+	// Set app password.
+	_, err := c.setAppPass(pw)
+	return err
 }
 
 // Login logs the user in, decrypting the account keys for all known DEXes.
--- a/client/core/core_test.go
+++ b/client/core/core_test.go
@@ -5,6 +5,7 @@ package core
 import (
 	"bytes"
 	"context"
+	"crypto/rand"
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
@@ -282,6 +283,7 @@ type TDB struct {
 	accountProofErr       error
 	verifyAccountPaid     bool
 	verifyCreateAccount   bool
+	verifyUpdateAccount   bool
 	accountInfoPersisted  *db.AccountInfo
 	accountProofPersisted *db.AccountProof
 	disabledAcct          *db.AccountInfo
@@ -312,6 +314,13 @@ func (tdb *TDB) DisableAccount(url string) error {
 	acct, _ := tdb.Account(url)
 	tdb.disabledAcct = acct
 	return tdb.disableAccountErr
+}
+
+func (tdb *TDB) UpdateAccount(ai *db.AccountInfo) error {
+	tdb.verifyUpdateAccount = true
+	tdb.acct = ai
+	tdb.accountInfoPersisted = ai
+	return nil
 }
 
 func (tdb *TDB) UpdateOrder(m *db.MetaOrder) error {
@@ -523,17 +532,19 @@ func newTWallet(assetID uint32) (*xcWallet, *TXCWallet) {
 		confs:      make(map[string]uint32),
 		confsErr:   make(map[string]error),
 	}
-	return &xcWallet{
+	xcWallet := &xcWallet{
 		Wallet:       w,
 		connector:    dex.NewConnectionMaster(w),
 		AssetID:      assetID,
 		hookedUp:     true,
 		dbID:         encode.Uint32Bytes(assetID),
-		encPW:        []byte{0x01},
+		encPass:      []byte{0x01},
 		synced:       true,
 		syncProgress: 1,
 		pw:           string(tPW),
-	}, w
+	}
+
+	return xcWallet, w
 }
 
 func (w *TXCWallet) Info() *asset.WalletInfo {
@@ -712,15 +723,42 @@ func (w *TXCWallet) Confirmations(ctx context.Context, coinID dex.Bytes) (uint32
 	return w.confs[id], false, w.confsErr[id]
 }
 
+type tCrypterSmart struct {
+	encryptErr error
+	decryptErr error
+	recryptErr error
+}
+
+// Encrypt appends 8 random bytes to given []byte to mock.
+func (c *tCrypterSmart) Encrypt(b []byte) ([]byte, error) {
+	randSuffix := make([]byte, 8)
+	rand.Read(randSuffix)
+	b = append(b, randSuffix...)
+	return b, c.encryptErr
+}
+
+// Decrypt deletes the last 8 bytes from given []byte.
+func (c *tCrypterSmart) Decrypt(b []byte) ([]byte, error) {
+	return b[:len(b)-8], c.decryptErr
+}
+
+func (c *tCrypterSmart) Serialize() []byte { return nil }
+
+func (c *tCrypterSmart) Close() {}
+
 type tCrypter struct {
 	encryptErr error
 	decryptErr error
 	recryptErr error
 }
 
-func (c *tCrypter) Encrypt(b []byte) ([]byte, error) { return b, c.encryptErr }
+func (c *tCrypter) Encrypt(b []byte) ([]byte, error) {
+	return b, c.encryptErr
+}
 
-func (c *tCrypter) Decrypt(b []byte) ([]byte, error) { return b, c.decryptErr }
+func (c *tCrypter) Decrypt(b []byte) ([]byte, error) {
+	return b, c.decryptErr
+}
 
 func (c *tCrypter) Serialize() []byte { return nil }
 
@@ -748,7 +786,7 @@ type testRig struct {
 	ws       *TWebsocket
 	dc       *dexConnection
 	acct     *dexAccount
-	crypter  *tCrypter
+	crypter  encrypt.Crypter
 }
 
 func newTestRig() *testRig {
@@ -1232,9 +1270,9 @@ func TestCreateWallet(t *testing.T) {
 	rig.db.encKeyErr = nil
 
 	// Crypter error.
-	rig.crypter.encryptErr = tErr
+	rig.crypter.(*tCrypter).encryptErr = tErr
 	ensureErr("Encrypt")
-	rig.crypter.encryptErr = nil
+	rig.crypter.(*tCrypter).encryptErr = nil
 
 	// Try an unknown wallet (not yet asset.Register'ed).
 	ensureErr("unregistered asset")
@@ -1442,12 +1480,12 @@ func TestRegister(t *testing.T) {
 	}
 
 	// password error
-	rig.crypter.recryptErr = tErr
+	rig.crypter.(*tCrypter).recryptErr = tErr
 	_, err = tCore.Register(form)
 	if !errorHasCode(err, passwordErr) {
 		t.Fatalf("wrong password error: %v", err)
 	}
-	rig.crypter.recryptErr = nil
+	rig.crypter.(*tCrypter).recryptErr = nil
 
 	// no host error
 	form.Addr = ""
@@ -1506,13 +1544,13 @@ func TestRegister(t *testing.T) {
 	dc.cfg.Markets = mkts
 
 	// error creating signing key
-	rig.crypter.encryptErr = tErr
+	rig.crypter.(*tCrypter).encryptErr = tErr
 	rig.queueConfig()
 	run()
 	if !errorHasCode(err, acctKeyErr) {
 		t.Fatalf("wrong account key error: %v", err)
 	}
-	rig.crypter.encryptErr = nil
+	rig.crypter.(*tCrypter).encryptErr = nil
 
 	// register request error
 	rig.queueConfig()
@@ -4922,6 +4960,57 @@ func TestWalletSettings(t *testing.T) {
 	}
 }
 
+func TestChangeAppPass(t *testing.T) {
+	rig := newTestRig()
+	defer rig.shutdown()
+	// Use the smarter crypter.
+	smartCrypter := &tCrypterSmart{}
+	rig.crypter = smartCrypter
+	rig.core.newCrypter = func([]byte) encrypt.Crypter { return smartCrypter }
+	rig.core.reCrypter = func([]byte, []byte) (encrypt.Crypter, error) { return smartCrypter, smartCrypter.recryptErr }
+
+	tCore := rig.core
+	var assetID uint32 = 54322
+	newTPW := []byte("apppass")
+
+	// App Password error
+	rig.crypter.(*tCrypterSmart).recryptErr = tErr
+	err := tCore.ChangeAppPass(tPW, newTPW)
+	if !errorHasCode(err, authErr) {
+		t.Fatalf("wrong error for password error: %v", err)
+	}
+	rig.crypter.(*tCrypterSmart).recryptErr = nil
+
+	wallet, _ := newTWallet(assetID)
+	encPW, _ := rig.crypter.Encrypt([]byte{0x01})
+	wallet.setEncPW(encPW)
+	tCore.wallets[assetID] = wallet
+
+	acctEncKey := make([]byte, len(tCore.conns[tDexHost].acct.encKey))
+	copy(acctEncKey, tCore.conns[tDexHost].acct.encKey)
+	origPass := make([]byte, len(wallet.encPW()))
+	copy(origPass, wallet.encPW())
+	err = tCore.ChangeAppPass(tPW, newTPW)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Ensure db's UpdateAccount func was called.
+	if !rig.db.verifyUpdateAccount {
+		t.Fatal("expected execution of db.UpdateAccount")
+	}
+	// Ensure db's account encrypted key changed.
+	if bytes.Equal(rig.db.accountInfoPersisted.EncKey, acctEncKey) {
+		t.Fatalf("expected db account EncKey to change, old: %x new: %x",
+			acctEncKey, rig.db.accountInfoPersisted.EncKey)
+	}
+	// Ensure xcWallet encrypted password updated.
+	if bytes.Equal(origPass, wallet.encPW()) {
+		t.Fatalf("expected xcwallet encPW to change: old %x new %x",
+			origPass, wallet.encPW())
+	}
+}
+
 func TestReconfigureWallet(t *testing.T) {
 	rig := newTestRig()
 	defer rig.shutdown()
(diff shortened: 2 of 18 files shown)


client/core/core.go
```diff
@@ -1586,6 +1586,32 @@ func (c *Core) WalletSettings(assetID uint32) (map[string]string, error) {
...
+// ChangeAppPass updates the application password to the provided new password
+// after validating the current password.
+func (c *Core) ChangeAppPass(appPW, newAppPW []byte) error {
+	// Validate current password.
+	_, err := c.encryptionKey(appPW)
+	if err != nil {
+		return newError(authErr, "ChangePassApp password error: %v", err)
+	}
+	// Set new password.
+	return c.setAppPass(newAppPW)
```
chappjc, line 1612: Things that are encrypted with the old pw:
- each `xcWallet`'s pass (for protected wallets) in `encPW`/`pw` (see `(*xcWallet).Unlock`), in memory and DB
- each `dexAccount`'s private key in `encKey`/`privKey` (see `(*dexAccount).setupEncryption`), in memory and DB

These "enc" fields need to be updated and the DB entries patched.

I believe your suggestion in matrix that a `db.UpdateAccount` or perhaps a more purposeful `UpdateAccountKey` method would be needed for this.  Fortunately there's already a `db.SetWalletPassword` that re-saves the `encPW`.

amassarwi: With https://github.com/decred/dcrdex/pull/978/commits/2a725e20a2ed3f0b03bd54b83ec962fb8354ba73
I'm able to change app password lock & unlock wallets but when trying to place an order I get:
```
2021-03-05 02:06:17.245 [ERR] WEB: error placing order: new order request with DEX server 127.0.0.1:17273 market dcr_btc failed: rpc error: error code 33: account ID mismatch

```

server's side:
```
2021-03-05 02:05:15.294 [INF] AUTH: New user registered: acct cc934689ce3a489a9f6cb19551d44c4351b852ad6f47cd5fddedf2c6fb9d45fd from 127.0.0.1 paid 100000000 to SsbeD1czwoAwqxV8fiq1sbhT7JsypYN5LPK
2021-03-05 02:05:15.394 [TRC] AUTH: User cc934689ce3a489a9f6cb19551d44c4351b852ad6f47cd5fddedf2c6fb9d45fd cancellation rate is now NaN% (0 cancels : 0 successes). Violation = false
2021-03-05 02:05:15.439 [DBG] AUTH: User cc934689ce3a489a9f6cb19551d44c4351b852ad6f47cd5fddedf2c6fb9d45fd score = 0: 0 (violations) + 0 (0 preimage misses) - 0 (0 successes)
2021-03-05 02:05:15.456 [INF] AUTH: Authenticated account cc934689ce3a489a9f6cb19551d44c4351b852ad6f47cd5fddedf2c6fb9d45fd from 127.0.0.1 with 0 active orders, 0 active matches
2021-03-05 02:06:17.245 [DBG] AUTH: Handling of 'limit' request for user cc934689ce3a489a9f6cb19551d44c4351b852ad6f47cd5fddedf2c6fb9d45fd failed: error code 33: account ID mismatch


```

chappjc: Oh gosh, seeing this diff you pushed I think I may have misguided you.  I might have been seeing way more work than necessary.  The password encrypts the _encryption keys themselves_ in the DB... and I actually don't think there is any change to the encryption parameters themselves (the serialized Crypter), and thus all the encPW and encKey fields might even stay the same.

:grimacing:

I gotta open up the branch and look at it.

Referring to this:

```go
func (c *Core) encryptionKey(pw []byte) (encrypt.Crypter, error) {
	keyParams, err := c.db.Get(keyParamsKey)
	if err != nil {
		return nil, fmt.Errorf("key retrieval error: %v", err)
	}
	crypter, err := c.reCrypter(pw, keyParams)
	if err != nil {
		return nil, fmt.Errorf("encryption key deserialization error: %v", err)
	}
	return crypter, nil
}
```

The app password decreypts (via reCrypter) the keyParams from the DB to get the crypter.  And (crap) I think that the serialized crypter should be the same after a successful password change, with only the encrypted key data (keyParams) being changed, which would mean all the enc fields aren't any different either.  

I'm going to look at this in the next couple hours and help straighten it out, sorry.

chappjc: Nah, false alarm.  Changing the password changes he encryption key as it's derived from the password (not encrypted by the password).
I'll look into what's causing that account id issue.

client/core/core.go
```diff
@@ -1629,6 +1629,80 @@ func (c *Core) WalletSettings(assetID uint32) (map[string]string, error) {
...
+		return err
+	}
+	// Update xcWallets' encrypted passwords.
+	c.updateWalletsEncPW(oldCrypter, newCrypter)
+	// Update dexConnections' dexAccount private & encrypted key.
+	c.connMtx.Lock()
+	defer c.connMtx.Unlock()
+	for _, dc := range c.conns {
+		// Generate new dexAccount private key and encrypt it.
+		dc.acct.generatePrivateAndEncKeys(newCrypter)
```
chappjc, line 1695: We don't want to generate new private keys, because the existing public-private key pair is the account identity;instead we need to re-encrypt the privatekey.

amassarwi: Sweet, will do.
Thanks @chappjc appreciate your insights and input! 🥂

amassarwi: @chappjc Updated and ready for a serious review.

client/core/core.go
```diff
@@ -1629,6 +1629,80 @@ func (c *Core) WalletSettings(assetID uint32) (map[string]string, error) {
...
+		return err
+	}
+	// Update xcWallets' encrypted passwords.
+	c.updateWalletsEncPW(oldCrypter, newCrypter)
+	// Update dexConnections' dexAccount private & encrypted key.
+	c.connMtx.Lock()
+	defer c.connMtx.Unlock()
+	for _, dc := range c.conns {
+		// Re-encrypt  private keys.
+		dc.acct.updateKeysEncryption(oldCrypter, newCrypter)
```
chappjc, line 1695: Should check error return before updating it in DB.

client/core/core.go
```diff
@@ -1629,6 +1629,80 @@ func (c *Core) WalletSettings(assetID uint32) (map[string]string, error) {
...
+		// Update account's db entry.
+		err = c.db.UpdateAccountKey(dc.acct.dbInfo())
```
chappjc, line 1697: Initially I though `UpdateAccountKey` would be good if it were just targeting the key piece of the account data, but since it updates the entire account info, let's go with `UpdateAccount` with a doc on the interface that the only piece of account info that is likely to change is the encrypted representation of the private keys since changing the key pair itself effectively creates a new account.

client/core/core.go
```diff
@@ -1629,6 +1629,80 @@ func (c *Core) WalletSettings(assetID uint32) (map[string]string, error) {
...
+	if err != nil {
+		return newError(authErr, "ChangePassApp password error: %v", err)
+	}
+	// Set new password.
+	newCrypter, err := c.setAppPass(newAppPW)
+	if err != nil {
+		return err
+	}
+	// Update xcWallets' encrypted passwords.
+	c.updateWalletsEncPW(oldCrypter, newCrypter)
```
chappjc, line 1689: Unchecked error

client/core/core_test.go
```diff
@@ -310,6 +310,10 @@ func (tdb *TDB) DisableAccount(ai *db.AccountInfo) error {
 	return nil
 }
 
+func (tdb *TDB) UpdateAccountKey(ai *db.AccountInfo) error {
```
chappjc, line 313: Could use a `TestChangeAppPass`

client/core/types.go
```diff
@@ -552,8 +552,27 @@ func (a *dexAccount) ID() account.AccountID {
 	return a.id
 }
 
-// setupEncryption creates and returns a new privkey for the account.
-// Returns existing privkey if previously set up.
+// updateKeysEncryption decrypts account's EncKey with given oldCrypter, then
+// it re
```
chappjc, line 556: Partial comment.

client/db/bolt/db.go
```diff
@@ -308,6 +308,13 @@ func (db *BoltDB) CreateAccount(ai *dexdb.AccountInfo) error {
 	})
 }
 
+// UpdateAccountKey Updates account's EncKey.
+func (db *BoltDB) UpdateAccountKey(ai *dexdb.AccountInfo) error {
+	return db.acctsUpdate(func(accts *bbolt.Bucket) error {
+		return accts.Put(ai.EncKey, ai.Encode())
```
chappjc, line 314: Non-disabled accounts are keyed on `ai.Host` I believe (See `(*BoltDB).{CreateAccount,AccountPaid}`).  I think the implementation of `UpdateAccount` should look similarly to `AccountPaid`, except instead of `acct.Put(feeProofKey, proof.Encode())`, we'd be updating the account info like `acct.Put(accountKey, ai.Encode())`.

I suppose the valid sub-buckets and keys inside each bucket could be documented better though since this isn't totally clear without studying the code.

chappjc (changes requested):

client/core/core.go
amassarwi: 💯

client/core/core.go
amassarwi: grrrrr

amassarwi:
@chappjc Updated

client/core/types.go
```diff
@@ -552,8 +552,28 @@ func (a *dexAccount) ID() account.AccountID {
 	return a.id
 }
 
-// setupEncryption creates and returns a new privkey for the account.
-// Returns existing privkey if previously set up.
```
JoeGruffins, line 556: Can move this instead of deleting?

client/core/core.go
```diff
@@ -1629,6 +1629,90 @@ func (c *Core) WalletSettings(assetID uint32) (map[string]string, error) {
...
+
+	return crypter, nil
+}
+
+func (c *Core) updateWalletsEncPW(oldCrypter, newCrypter encrypt.Crypter) error {
+	c.walletMtx.Lock()
+	defer c.walletMtx.Unlock()
+	for _, w := range c.wallets {
+		w.mtx.Lock()
+		defer w.mtx.Unlock()
```
JoeGruffins, line 1651: This defer will happen at the end of the function, and not the end of the loop. I think it is fine here, but could lead to a "gotcha" if this method is ever expanded downwards.

JoeGruffins (commented):
comment message of 9d326c8773c maybe should change mutuable -> mutable

The behavior when you enter something wrong differs for these fields:
notgoaway

If you enter the wrong current app pass, the current and new are cleared, but the confirm new is not. If new and confirm are different, it errors but none are cleared. Maybe all should be cleared for any mistake or success?

Unrelated to this pr:
I had the simnet test fail once:
```
--- FAIL: TestResendPendingRequests (39.07s)
    trade_simnet_test.go:823: [client 2] dcr balance change not in expected range 19.00000000 - 20.00000000, got 9.99993000
FAIL
exit status 1
FAIL    decred.org/dcrdex/client/core   500.617s
```
And I see this as an error in client logs, but should maybe be a warning?
```
[ERR] WEB: max order estimation error: dcr wallet MaxOrder error: error getting network fee estimate: -32603: no bucket with the minimum required success percentage found
```

client/webserver/site/src/js/register.js
```diff
@@ -34,19 +34,17 @@ export default class RegistrationPage extends BasePage {
...
-    bindForm(page.appPWForm, page.appPWSubmit, () => { this.setAppPass() })
+    bindForm(page.appPWForm, page.appPWSubmit, () => this.setAppPass())
 
     // NEW DCR WALLET
     // This form is only shown if there is no DCR wallet yet.
-    this.walletForm = new NewWalletForm(app, page.newWalletForm, () => {
-      this.changeForm(page.newWalletForm, page.dexAddrForm)
-    })
+    this.walletForm = new NewWalletForm(app, page.newWalletForm, () =>
+      this.changeForm(page.newWalletForm, page.dexAddrForm))
```
buck54321, line 42: Please don't change this. I really prefer not to use an un-bracketed two-liner syntax. For consistency, if a block is more than one line, it's at least three.

client/webserver/site/src/js/register.js
```diff
@@ -34,19 +34,17 @@ export default class RegistrationPage extends BasePage {
...
 
     // OPEN DCR WALLET
     // This form is only shown if there is a wallet, but it's not open.
-    bindOpenWallet(app, page.unlockWalletForm, () => {
-      this.changeForm(page.unlockWalletForm, page.dexAddrForm)
-    })
+    bindOpenWallet(app, page.unlockWalletForm, () =>
+      this.changeForm(page.unlockWalletForm, page.dexAddrForm))
```
buck54321, line 47: Please don't change this.

client/webserver/site/src/js/settings.js
```diff
@@ -86,9 +92,8 @@ export default class SettingsPage extends BasePage {
     }
     Doc.bind(document, 'keyup', this.keyup)
 
-    page.forms.querySelectorAll('.form-closer').forEach(el => {
-      Doc.bind(el, 'click', () => { closePopups() })
-    })
+    page.forms.querySelectorAll('.form-closer').forEach(el =>
+      Doc.bind(el, 'click', () => closePopups()))
```
buck54321, line 96: Here too.

client/webserver/site/src/js/settings.js
```diff
@@ -299,6 +301,39 @@ export default class SettingsPage extends BasePage {
...
+    }
+    const res = await postJSON('/api/changeapppass', req)
+    page.appPW.value = ''
+    page.newAppPW.value = ''
+    if (!app.checkResponse(res, true)) {
+      page.changePWErrMsg.textContent = res.msg
+      Doc.show(page.changePWErrMsg)
+      return
+    }
+    loaded()
```
buck54321, line 333: You are returning before hiding the loader on error, no?

client/webserver/site/src/html/forms.tmpl
```diff
@@ -203,6 +203,31 @@
...
+    <label for="newAppPW" class="pl-1 mb-1">New Password</label>
+    <input type="password" class="form-control select" id="newAppPW" autocomplete="new-password">
+  </div>
+  <div class="mt-2">
+    <label for="confirmNewPW" class="pl-1 mb-1">Confirm New Password</label>
+    <input type="password" class="form-control select" id="confirmNewPW" autocomplete="new-password">
+  </div>
+  <div class="d-flex justify-content-end mt-4">
+    <button id="submitNewPW" type="submit" class="col-8 justify-content-center fs15 bg2 selected">Submit</button>
+  </div>
```
buck54321, line 226: Would you add some margin between the buttons while you're here?
image

buck54321 (changes requested):

client/db/interface.go
```diff
@@ -30,6 +30,12 @@ type DB interface {
 	CreateAccount(ai *AccountInfo) error
 	// DisableAccount sets the AccountInfo disabled status to true.
 	DisableAccount(ai *AccountInfo) error
+	// UpdateAccount updates account's info blob.
+	//
+	// The only piece of account info that is likely to change is the encrypted
+	// representation of the private key since changing the key pair itself
+	// effectively creates a new account.
```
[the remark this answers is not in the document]
chappjc: I suggested some kind of comment here because the updater is overly broad, updating all the fields, but it actually is _incorrect_ if the key pair itself changes because that means it's a new account by definition.  I think some kind of comment to that effect is warranted until/unless a targeted updater is created.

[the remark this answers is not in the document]
buck54321: Gotcha. Ignore me, @amassarwi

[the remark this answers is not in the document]
chappjc: Could reword it so it does not say "only piece of account info that is likely to change", because now that I think of it I suppose host or cert etc. could change too.  Just wanted to doc that it's almost certainly incorrect to change any of the keys.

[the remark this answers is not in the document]
buck54321: Perfect

client/core/core.go
amassarwi: Thanks guys will do.
@buck54321 Nice pattern to resolve such scenarios.

client/core/core.go
```diff
@@ -1629,6 +1629,90 @@ func (c *Core) WalletSettings(assetID uint32) (map[string]string, error) {
...
+	if err != nil {
+		return newError(authErr, "ChangeAppPass password error: %v", err)
+	}
+	// Set new password.
+	newCrypter, err := c.setAppPass(newAppPW)
+	if err != nil {
+		return err
+	}
+	// Update xcWallets' encrypted passwords.
+	err = c.updateWalletsEncPW(oldCrypter, newCrypter)
```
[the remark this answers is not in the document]
amassarwi: 👍

client/core/core.go
```diff
@@ -1804,21 +1888,25 @@ func (c *Core) setWalletPassword(wallet *xcWallet, newPW []byte, crypter encrypt
...
 		if err != nil {
+			wallet.mtx.Unlock()
 			return newError(authErr,
 				"setWalletPassword unlocking wallet error, is the new password correct?: %v", err)
 		}
 		wallet.encPW = encNewPW
 	} else {
 		wallet.encPW = nil
 	}
+	wallet.mtx.Unlock()
```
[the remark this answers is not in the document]
amassarwi: 👍

client/core/wallet.go
```diff
@@ -19,9 +19,9 @@ type xcWallet struct {
 	connector *dex.ConnectionMaster
 	AssetID   uint32
 	dbID      []byte
-	encPW     []byte // empty means wallet not password protected
 
 	mtx          sync.RWMutex
+	encPW        []byte // empty means wallet not password protected
```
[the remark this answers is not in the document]
amassarwi: @buck54321 Thanks for this nice suggestion. 
will keep it mind for future mutex protected fields.

client/db/interface.go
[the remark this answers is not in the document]
amassarwi: What about ?

```
// It almost certainly incorrect to change any of the keys since changing the key pair itself
// effectively creates a new account.
```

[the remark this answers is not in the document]
buck54321: That reads well.

client/core/core_test.go
```diff
@@ -4937,6 +4945,55 @@ func TestWalletSettings(t *testing.T) {
...
+	}
+	if !bytes.Equal(rig.db.accountInfoPersisted.EncKey, acctEncKey) {
+		t.Fatalf("expected db account EncKey to not change, want: %v got: %v",
+			string(acctEncKey), string(rig.db.accountInfoPersisted.EncKey))
+	}
+	// Ensure xcWallet encrypted password didn't change.
+	nWallet := tCore.wallets[assetID]
+	if !bytes.Equal(wallet.encPW, nWallet.encPW) {
+		t.Fatal("expected xcwallet to not change")
+	}
```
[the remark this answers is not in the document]
amassarwi: It doesn't seem to update, with the following changes:

```
diff --git a/client/core/core_test.go b/client/core/core_test.go
index bd1ad1b8..cd7ee82d 100644
--- a/client/core/core_test.go
+++ b/client/core/core_test.go
@@ -4973,6 +4973,7 @@ func TestChangeAppPass(t *testing.T) {
 	}
 	defer wallet.Disconnect()
 
+	origPass := wallet.encPW
 	err = tCore.ChangeAppPass(tPW, newTPW)
 	if err != nil {
 		t.Fatal(err)
@@ -4987,10 +4988,10 @@ func TestChangeAppPass(t *testing.T) {
 		t.Fatalf("expected db account EncKey to not change, want: %v got: %v",
 			string(acctEncKey), string(rig.db.accountInfoPersisted.EncKey))
 	}
-	// Ensure xcWallet encrypted password didn't change.
+	// Ensure xcWallet encrypted password updated.
 	nWallet := tCore.wallets[assetID]
-	if !bytes.Equal(wallet.encPW, nWallet.encPW) {
-		t.Fatal("expected xcwallet to not change")
+	if bytes.Equal(origPass, nWallet.encPW) {
+		t.Fatalf("expected xcwallet to change: old %x new %x", origPass, nWallet.encPW)
 	}
 }
 
```

I had the test failing with:
`core_test.go:4994: expected xcwallet to change: old 01 new 01`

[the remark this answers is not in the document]
amassarwi: Could https://github.com/decred/dcrdex/blob/master/client/core/core_test.go#L740-L752 be the reason why ?

Stub `Encrypt` & `Decrypt` return the `[]byte` as is.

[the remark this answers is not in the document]
buck54321: That explains it.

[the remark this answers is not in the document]
amassarwi: What would you suggest to improve that ?
Adding an extra random char to the []byte on encrypt and deleting it on decrypt ?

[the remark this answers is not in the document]
chappjc: That's a thought as long as that char doesn't turn out the same for the two `encPW`s, so maybe 8 random bytes.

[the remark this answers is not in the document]
amassarwi: Sweet will do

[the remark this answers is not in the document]
amassarwi: done

JoeGruffins (approved):

client/core/wallet.go
```diff
@@ -106,8 +122,7 @@ func (w *xcWallet) unlocked() bool {
 // this is true only if the decrypted password is cached. Use this to determine
 // if the wallet may be unlocked without user interaction (via refreshUnlock).
 func (w *xcWallet) locallyUnlocked() bool {
-	// return len(w.encPW) == 0 || len(w.pw) > 0
-	if len(w.encPW) == 0 {
+	if len(w.encPW()) == 0 {
```
chappjc, line 125: In both returns, the `w.mtx` gets locked one way or the other, so this should be:

```go
w.mtx.RLock()
defer w.mtx.RUnlock()
if len(w.encPass) == 0 {
    return true // unencrypted wallet
}
return len(w.pw) > 0 // cached password for encrypted wallet
```

client/core/trade_simnet_test.go
```diff
@@ -1492,9 +1492,8 @@ func (client *tClient) disableWallets() {
 	// NOTE: this is not reversible, but could be made so with the undo data:
 	// walletPasses[cid] = make(map[uint32]passes, len(client.core.wallets))
 	for _, wallet := range client.core.wallets {
+		wallet.setEncPW([]byte{0})
```
chappjc, line 1495: This change wasn't necessary since the mutex is locked right below to update `pw`

client/core/wallet.go
```diff
@@ -70,21 +84,23 @@ func (w *xcWallet) refreshUnlock() (unlockAttempted bool, err error) {
...
+			unbip(w.AssetID))
 	}
 	return true, w.Wallet.Unlock(w.pw)
 }
 
 // Lock the wallet. For encrypted wallets (encPW set), this clears the cached
 // decrypted password and attempts to lock the wallet backend.
 func (w *xcWallet) Lock() error {
-	if len(w.encPW) == 0 {
+	if len(w.encPW()) == 0 {
```
chappjc, line 103: See comment with the same pattern as in `locallyUnlocked`

client/core/wallet.go
```diff
@@ -70,21 +84,23 @@ func (w *xcWallet) refreshUnlock() (unlockAttempted bool, err error) {
...
 	// Locked backend requires both encrypted and decrypted passwords.
-	if len(w.encPW) == 0 {
-		return false, fmt.Errorf("%s wallet reporting as locked but no password has been set", unbip(w.AssetID))
+	if len(w.encPW()) == 0 {
+		return false, fmt.Errorf("%s wallet reporting as locked but no password"+
+			" has been set", unbip(w.AssetID))
 	}
 	w.mtx.RLock()
 	defer w.mtx.RUnlock()
 	if len(w.pw) == 0 {
```
chappjc, line 93: See comment with the same pattern as in `locallyUnlocked`

chappjc (changes requested):

client/core/trade_simnet_test.go
amassarwi: true!

client/core/wallet.go
```diff
@@ -117,18 +132,19 @@ func (w *xcWallet) locallyUnlocked() bool {
...
 		Symbol:       unbip(w.AssetID),
 		AssetID:      w.AssetID,
-		Open:         len(w.encPW) == 0 || len(w.pw) > 0,
+		Open:         len(encPW) == 0 || len(w.pw) > 0,
 		Running:      w.connector.On(),
 		Balance:      w.balance,
 		Address:      w.address,
 		Units:        winfo.Units,
-		Encrypted:    len(w.encPW) > 0,
+		Encrypted:    len(encPW) > 0,
```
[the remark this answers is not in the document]
amassarwi: 👍

client/core/core_test.go
```diff
@@ -743,9 +755,18 @@ type tCrypter struct {
 	recryptErr error
 }
 
-func (c *tCrypter) Encrypt(b []byte) ([]byte, error) { return b, c.encryptErr }
+// Encrypt appends 8 random bytes to given []byte to mock.
+func (c *tCrypter) Encrypt(b []byte) ([]byte, error) {
+	randSuffix := make([]byte, 8)
+	rand.Read(randSuffix)
+	b = append(b, randSuffix...)
```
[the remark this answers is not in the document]
amassarwi: Thanks for the insights! 🍷

amassarwi:
@chappjc updated

client/core/core_test.go
```diff
@@ -541,24 +550,27 @@ type TXCWallet struct {
 	preRedeem         *asset.PreRedeem
 }
 
-func newTWallet(assetID uint32) (*xcWallet, *TXCWallet) {
+func newTWallet(assetID uint32, crypter *tCrypter) (*xcWallet, *TXCWallet) {
```
buck54321, line 553: Nah. Instead of modifying the existing `tCrypter` and forcing all tests to use it, create a new `Crypter`-implementing stub type, and only do the `encPW` stuff in the test that needs it, not in this general purpose constructor.

client/core/types.go
```diff
@@ -552,6 +552,28 @@ func (a *dexAccount) ID() account.AccountID {
...
 }
 
+// updateKeysEncryption decrypts account's EncKey with given oldCrypter,
+// re-encrypts it with the given newCrypter, then stores the new EncKey on
+// the dexaccount struct.
+func (a *dexAccount) updateKeysEncryption(oldCrypter, newCrypter encrypt.Crypter) error {
+	a.keyMtx.Lock()
+	defer a.keyMtx.Unlock()
+	// Decrypt the encrpyted key with old crypter.
+	encKey, err := oldCrypter.Decrypt(a.encKey)
```
buck54321, line 562: `encKey` -> `key`. It's not encrypted at this point.

client/webserver/site/src/js/settings.js
```diff
@@ -299,6 +302,47 @@ export default class SettingsPage extends BasePage {
...
+      return
+    }
+    const loaded = app.loading(page.changeAppPW)
+    const req = {
+      appPW: page.appPW.value,
+      newAppPW: page.newAppPW.value
+    }
+    clearValues()
+    const res = await postJSON('/api/changeapppass', req)
+    if (!app.checkResponse(res, true)) {
```
buck54321, line 336: You can put `loaded()` in between these two lines so you'll only have to do it once.

buck54321 (changes requested):
Tests well.

client/core/core.go
```diff
@@ -1804,23 +1892,30 @@ func (c *Core) setWalletPassword(wallet *xcWallet, newPW []byte, crypter encrypt
...
-		if err != nil {
-			return newError(encryptionErr, "encryption error: %v", err)
-		}
-		err = wallet.Wallet.Unlock(string(newPW))
-		if err != nil {
-			return newError(authErr,
-				"setWalletPassword unlocking wallet error, is the new password correct?: %v", err)
+	updateWalletPW := func(w *xcWallet) error {
+		if newPasswordSet {
+			// Encrypt password if it's not an empty string.
```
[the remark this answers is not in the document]
amassarwi: Hrmm you are right regarding no utility, i do find a bit more readable that way, instead of have the logic inside the for
loop.

Another option would be to define it as unexported `Core` func, what do you think ?

client/core/core.go
```diff
@@ -2154,14 +2249,9 @@ func (c *Core) InitializeClient(pw []byte) error {
...
-
-	crypter := c.newCrypter(pw)
-	err := c.db.Store(keyParamsKey, crypter.Serialize())
-	if err != nil {
-		return fmt.Errorf("error storing key parameters: %w", err)
+	// Set app password.
+	if _, err := c.setAppPass(pw); err != nil {
+		return err
 	}
 	return nil
```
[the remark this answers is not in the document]
amassarwi: `c.setAppPass(pw)` returns two values, so i'm not sure if that would work.

client/core/core_test.go
amassarwi: Good point 👍

client/core/core.go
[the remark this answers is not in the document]
buck54321: Derp. My bad.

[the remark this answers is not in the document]
chappjc: Could still be:

```go
_, err := c.setAppPass(pw)
return err
```

amassarwi:
@buck54321 updated mate

client/core/core.go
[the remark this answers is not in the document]
amassarwi: true! 😱

client/core/core.go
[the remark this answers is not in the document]
buck54321: > inside the for loop

Did you mean the `if` block, or were you thinking of the `for` loop in `updateWalletsEncPW`? If you meant the `if` block, it's still there, just inside of a function now. I don't really see how that makes it more readable.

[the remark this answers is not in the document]
amassarwi: I was thinking of the `for` in `updateWalletsEncPW` but it's not that bad having the logic there, moving it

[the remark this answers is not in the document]
amassarwi: Updating this one, both inline funcs are `updateWalletPW` this why I got confused 🙈.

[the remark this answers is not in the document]
amassarwi: Done

client/webserver/site/src/html/settings.tmpl
```diff
@@ -43,7 +43,8 @@
         servers.
         </p>
         <button id="addADex" class="bg2 selected">Add a DEX</button>
-        <button id="importAccount" class="bg2 selected">Import Account</button>
+        <button id="importAccount" class="bg2 selected ml-2">Import Account</button>
+        <button id="changeAppPW" class="bg2 selected ml-2">Change App Password</button>
```
buck54321, line 47: Let's make a new section for this button. This section is for DEX accounts.

buck54321 (approved):
One last thing, but looks great.

client/webserver/site/src/html/settings.tmpl
amassarwi: image

amassarwi:
Needs a rebase after #985.

chappjc:
@amassarwi I just rebased it.  Sorry didn't see your comment

amassarwi:
@chappjc all good, thank you

Issue #939, opened 2021-01-25
backends: Add Ethereum.


JoeGruffins:
### Ethereum support is coming to the dex.

This issue is intended to get the ball rolling and should be closed by adding eth backends for the server and client. This issue does not include atomic swaps, only basic communication abilities in an attempt to keep pr's smaller.

### Client

I propose using [geth](https://github.com/ethereum/go-ethereum) because it has active development and is written in golang.

### Networks

A simnet setup script will have a separate issue.

There are four [testnets](https://ethereum.org/en/developers/docs/networks/).  After looking into them, I propose using [Gorli](https://goerli.net/) for three reasons. 
 1. It is supported by geth with just the `--goerli` flag so setup is easy.
 2. It seems to be stable. 
 3. The faucet, while requiring captcha, works.

chappjc:
@JoeGruffins Thanks for feeling this out.  I was also settled on geth, but I hadn't gotten to the bottom of the four different testnets.  However, I had already made a note of Ropsten being the best target given its description as "A proof-of-work testnet. This means it's the best like-for-like representation of Ethereum." vs the other PoA networks.  I'm actually kinda surprised you can choose the consensus algorithm on testnet and development networks, so perhaps this is considered a minor detail in ethereum?

JoeGruffins:
I read a bit of this https://gitter.im/ethereum/ropsten and decided it looks like a headache. You need to beg for eth because faucets don't work:
```
 bold
@AboldUSER
Jan 15 08:28
Is there something going on with the Ropsten network today? I have tried to queue or receive ETH from 4 different faucets but non seem to be working today for me
```
and apparently the chain gets actively attacked at times or simply has no miners and stops moving: 
```
0xlegos
@0xlegos
Oct 30 2020 21:30
Since yesterday, the ropsten network has been paralyzed. Has anyone attacked the test network? Can they get any benefits?
```

JoeGruffins:
geth has these flags, so robstein is still easy to setup:
```
  --goerli                            Görli network: pre-configured proof-of-authority test network
  --rinkeby                           Rinkeby network: pre-configured proof-of-authority test network
  --yolov2                            YOLOv2 network: pre-configured proof-of-authority shortlived test network.
  --ropsten                           Ropsten network: pre-configured proof-of-work test network
```
I will start it up and see what it's like.

chappjc:
[...]

PR #979, opened 2021-02-18
multi: Add Ethereum server asset.

JoeGruffins:
Work towards #939
Depends upon #956 and #944

Adds a skeleton eth server client.


dex/testing/eth/create-node.sh
```diff
@@ -56,19 +56,17 @@ function mine() {
 }
 EOF
 
-# Write password file to unlock accounts later.
-cat > "${GROUP_DIR}/password" <<EOF
+  # Write password file to unlock accounts later.
+  cat > "${GROUP_DIR}/password" <<EOF
 $CHAIN_PASSWORD
-$ADDRESS_PASSWORD
```
JoeGruffins, line 62: It does not appear that this was unlocking the account. We will pass the password when sending funds.

server/asset/eth/eth.go
```diff
@@ -0,0 +1,370 @@
...
+// decodeCoinID decodes the coin ID into a contract address and secret hash.
+func decodeCoinID(coinID []byte) (common.Address, []byte, error) {
+	if len(coinID) != coinIDSize {
+		return common.Address{}, nil, fmt.Errorf("coin ID wrong length. expected %d, got %d",
+			coinIDSize, len(coinID))
+	}
+	secretHash := make([]byte, 32)
+	copy(secretHash, coinID[20:])
+	return common.BytesToAddress(coinID[:20]), secretHash, nil
+}
```
JoeGruffins, line 354: I believe a "coin" is best identified by the address of the contract it belongs to and the secret hash it is keyed by. This is tentative and up for debate. See #1001

buck54321: Wouldn't the eth contract address be a constant?

JoeGruffins: yes. If we update the contract, though, I assume it would change.

dex/testing/eth/harness.sh
```diff
@@ -107,19 +130,64 @@ tmux send-keys -t $SESSION:0 "cd ${NODES_ROOT}/harness-ctl" C-m
...
+
+SEND_AMT=5000000000000000000000
+echo "Sending 5000 eth to beta and gamma."
+"${NODES_ROOT}/harness-ctl/alpha" "attach --preload ${NODES_ROOT}/harness-ctl/send.js --exec send(\"${ALPHA_ADDRESS}\",\"${GAMMA_ADDRESS}\",${SEND_AMT})"
+"${NODES_ROOT}/harness-ctl/alpha" "attach --preload ${NODES_ROOT}/harness-ctl/send.js --exec send(\"${ALPHA_ADDRESS}\",\"${DELTA_ADDRESS}\",${SEND_AMT})"
+
+# TODO: After we start mining blocks on alpha, it seems that we can never mine
+# blocks on beta. Doing so sends alpha into some sort of error mode. Find out
+# why that is and correct it if a settings miss.
+#
```
JoeGruffins, line 175: It looks like this is no longer the case on geth `v1.10.1` so was probably due to a bug. Will remove this todo.

dex/testing/eth/create-node.sh
```diff
@@ -2,19 +2,18 @@
...
-ADDRESS_JSON=${10}
-ADDRESS_JSON_FILE_NAME=${11}
-NODE_KEY=${12}
+ADDRESS_JSON=$8
+ADDRESS_JSON_FILE_NAME=$9
+NODE_KEY=${10}
+SYNC_MODE=${11}
 
 GROUP_DIR="${NODES_ROOT}/${NAME}"
 MINE_JS="${GROUP_DIR}/mine.js"
```
buck54321, line 19: I had to add these two lines a few lines below here.

```
mkdir -p "${NODE_DIR}/keystore"
mkdir -p "${NODE_DIR}/geth"
```

server/asset/eth/client_harness_test.go
```diff
@@ -0,0 +1,61 @@
+// +build harness
```
buck54321, line 1: Having `client` in the filename is a little confusing, since this is the server backend.

buck54321 (changes requested):
This is a great start to the server-side work. Having this skeleton in place will help us to visualize and compartmentalize the work for efficient development.

After the `create-node.sh` fix, this tests well. This will all be iterated over as we get deeper into implementation, so I think we can get this in soon. 

@chappjc, any interest in maintaining a separate eth integration branch?

chappjc:
> @chappjc, any interest in maintaining a separate eth integration branch?

We can if it seems like eth work has the potential to break other things, but I don't mind having incomplete features on master if everything else is functional.

server/asset/eth/eth.go
JoeGruffins: I'm not sure I follow that last part of that. There would not be any need to change a coin ID. Client would initiate the swap with a contract. Server will have a map of contract addresses and reject older contracts if necessary with a penalty. It makes perfect sense that all servers will be using the same contract. I don't think that is a problem.

I believe contracts will need to be updated for security reasons, and if eth consensus rules change.

buck54321: > There would not be any need to change a coin ID

Sorry. Not an individual coin ID. The coin ID encoding protocol. It's hard to change it later.

dex/testing/eth/create-node.sh
JoeGruffins: Added.

server/asset/eth/client_harness_test.go
JoeGruffins: Renamed most `client` to `rpcclient` for clarity. Better that way?

dex/testing/eth/create-node.sh
```diff
@@ -111,9 +122,16 @@ cat > "${NODE_DIR}/geth/nodekey" <<EOF
...
-	"--password ${GROUP_DIR}/password" C-m
+if [ "${SYNC_MODE}" = "fast" ]; then
+  # Start the eth node with both accounts unlocked, listening restricted to
+  # localhost, and syncmode set to fast.
+  tmux send-keys -t "$TMUX_WIN_ID" "${NODES_ROOT}/harness-ctl/${NAME} --nodiscover " \
+	  "--config ${NODE_DIR}/eth.conf --unlock ${CHAIN_ADDRESS} " \
+	  "--password ${GROUP_DIR}/password --light.serve 25" C-m
+
+else
+  # Start the eth node listening restricted to localhost, and syncmode set to light.
```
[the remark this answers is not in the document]
JoeGruffins: Changed.

server/asset/eth/eth.go
JoeGruffins: I think I will add a byte for flags. At least we might want unique ids for maker and taker, otherwise they would be the same. If there's somewhere using coin ids as a map key, expecting them to be unique... A one bit flag in the byte field could resolve this, and store a limited amount of arbitrary data in the future. maybe make it two bytes?

JoeGruffins: Added flags anyway. The scheme may be different in the end, as long as we decide on something before allowing mainnet. It may seem like this doesn't allow direct payments for a fee, but, a contract may be also be the best way to handle a fee payment. We could also have a flag that designates this is a transaction hash, then the "secret hash" could instead be the txid in that case.

chappjc: I was wondering about normal transactions too, such as withdraw/deposit and how those would be encoded.  I suspect we will need some kind of dynamic encoding with the flags to switch it to tx hash as you suggested.  I'm just catching up on your eth work though so take that with a grain of salt.

chappjc: Thinking through the coin ID issue: https://github.com/decred/dcrdex/issues/1021

chappjc (approved):
Good to go with a go.sum resolve please.

I'm pretty confident we will have to evolve the coin ID encoding for ethereum, sometimes holding transactions and sometimes accounts and other data like secret hash as you have here, but we can definitely go forward with this.

PR #982, opened 2021-02-21
Bump x/crypto for blake2 .s stack corruption fixes

dajohi:


chappjc (approved):
OK.

We should consider this a reason to switch to a hash function in Go's core library (no need to update our dependencies, just rebuild), such as `crypto/sha256`.

PR #984, opened 2021-02-22
Fix missing arg on simnet mine-alpha script

vctt94:
Without it the mine-alpha script will fail


chappjc (approved):
I think this is good, although I've always done `./mine-alpha 1`, so never noticed that this is in mine-beta but not mine-alpha.

JoeGruffins (approved):

PR #985, opened 2021-02-23
dexc: disable account feature.

### LINKED ISSUE #961: need UI for manually disabling account
The client db has a DisableAccount method, but it is currently only used if a dex server reports account not found.  There should be a way to manually disable an account with Core method and a GUI button.  In the absence of these capabilities, to re-register a suspended account at a given DEX required deleting the dexc.db

amassarwi:
This adds a new button to dexc's GUI to disable account manually, this is useful when a user wants to
re-register a suspended account which required deleting the dexc.db before this commit.

Closes #961.

### DIFF
--- a/client/core/account.go
+++ b/client/core/account.go
@@ -17,18 +17,10 @@ func (c *Core) AccountDisable(pw []byte, addr string) error {
 		return codedError(passwordErr, err)
 	}
 
-	// Parse address.
-	host, err := addrHost(addr)
+	// Get dex connection by host.
+	dc, _, err := c.dex(addr)
 	if err != nil {
-		return newError(addressParseErr, "error parsing address: %v", err)
-	}
-
-	// Get the dexConnection.
-	c.connMtx.RLock()
-	defer c.connMtx.RUnlock()
-	dc, found := c.conns[host]
-	if !found {
-		return newError(unknownDEXErr, "DEX: %s", host)
+		return newError(unknownDEXErr, "error retrieving dex conn: %v", err)
 	}
 
 	// Check active orders.
@@ -36,9 +28,9 @@ func (c *Core) AccountDisable(pw []byte, addr string) error {
 		return fmt.Errorf("cannot disable account with active orders")
 	}
 
-	err = c.db.DisableAccount(host)
+	err = c.db.DisableAccount(dc.acct.host)
 	if err != nil {
-		return newError(accountDisableErr, "Error disabling account: %v", err)
+		return newError(accountDisableErr, "error disabling account: %v", err)
 	}
 	// Stop dexConnection books.
 	dc.cfgMtx.RLock()
@@ -63,7 +55,7 @@ func (c *Core) AccountDisable(pw []byte, addr string) error {
 	dc.cfgMtx.RUnlock()
 	// Disconnect and delete connection from map.
 	dc.connMaster.Disconnect()
-	delete(c.conns, host)
+	delete(c.conns, dc.acct.host)
 
 	return nil
 }
--- a/client/core/account.go
+++ b/client/core/account.go
@@ -2,10 +2,58 @@ package core
 
 import (
 	"encoding/hex"
+	"fmt"
 
 	"decred.org/dcrdex/client/db"
 	"github.com/decred/dcrd/dcrec/secp256k1/v3"
 )
+
+// AccountDisable is used to disable an account by given host and application
+// password.
+func (c *Core) AccountDisable(pw []byte, host string) error {
+	// Validate password.
+	_, err := c.encryptionKey(pw)
+	if err != nil {
+		return codedError(passwordErr, err)
+	}
+
+	// Parse address.
+	addr, err := addrHost(host)
+	if err != nil {
+		return newError(addressParseErr, "error parsing address: %v", err)
+	}
+
+	// Get the dexConnection.
+	c.connMtx.RLock()
+	defer c.connMtx.RUnlock()
+	dc, found := c.conns[addr]
+	if !found {
+		return newError(unknownDEXErr, "DEX: %s", addr)
+	}
+
+	// Check active orders.
+	if dc.hasActiveOrders() {
+		return fmt.Errorf("cannot disable account with active orders")
+	}
+
+	acctInfo, err := c.db.Account(dc.acct.host)
+	if err != nil {
+		return newError(accountRetrieveErr, "Error retrieving account: %v", err)
+	}
+	err = c.db.DisableAccount(acctInfo)
+	if err != nil {
+		return newError(accountDisableErr, "Error disabling account: %v", err)
+	}
+	// Stop dexConnection books.
+	for _, market := range dc.marketMap() {
+		dc.stopBook(market.BaseID, market.QuoteID)
+	}
+	// Disconnect and delete connection from map.
+	dc.connMaster.Disconnect()
+	delete(c.conns, addr)
+
+	return nil
+}
 
 // AccountExport is used to retrieve account by host for export.
 func (c *Core) AccountExport(pw []byte, host string) (*Account, error) {
--- a/client/core/account.go
+++ b/client/core/account.go
@@ -10,7 +10,7 @@ import (
 
 // AccountDisable is used to disable an account by given host and application
 // password.
-func (c *Core) AccountDisable(pw []byte, host string) error {
+func (c *Core) AccountDisable(pw []byte, addr string) error {
 	// Validate password.
 	_, err := c.encryptionKey(pw)
 	if err != nil {
@@ -18,7 +18,7 @@ func (c *Core) AccountDisable(pw []byte, host string) error {
 	}
 
 	// Parse address.
-	addr, err := addrHost(host)
+	host, err := addrHost(addr)
 	if err != nil {
 		return newError(addressParseErr, "error parsing address: %v", err)
 	}
@@ -26,9 +26,9 @@ func (c *Core) AccountDisable(pw []byte, host string) error {
 	// Get the dexConnection.
 	c.connMtx.RLock()
 	defer c.connMtx.RUnlock()
-	dc, found := c.conns[addr]
+	dc, found := c.conns[host]
 	if !found {
-		return newError(unknownDEXErr, "DEX: %s", addr)
+		return newError(unknownDEXErr, "DEX: %s", host)
 	}
 
 	// Check active orders.
@@ -36,22 +36,23 @@ func (c *Core) AccountDisable(pw []byte, host string) error {
 		return fmt.Errorf("cannot disable account with active orders")
 	}
 
-	acctInfo, err := c.db.Account(dc.acct.host)
-	if err != nil {
-		return newError(accountRetrieveErr, "Error retrieving account: %v", err)
-	}
-	err = c.db.DisableAccount(acctInfo)
+	err = c.db.DisableAccount(host)
 	if err != nil {
 		return newError(accountDisableErr, "Error disabling account: %v", err)
 	}
 	// Stop dexConnection books.
 	dc.cfgMtx.RLock()
 	for _, m := range dc.cfg.Markets {
-		// Emptry bookie's feeds map & stop close timers.
+		// Empty bookie's feeds map, close feeds' channels & stop close timers.
 		dc.booksMtx.Lock()
 		if b, found := dc.books[m.Name]; found {
+			b.mtx.Lock()
+			for _, f := range b.feeds {
+				close(f.C)
+			}
 			b.feeds = make(map[uint32]*BookFeed, 1)
 			b.closeTimer.Stop()
+			b.mtx.Unlock()
 		}
 		dc.booksMtx.Unlock()
 
--- a/client/core/account.go
+++ b/client/core/account.go
@@ -44,10 +44,20 @@ func (c *Core) AccountDisable(pw []byte, host string) error {
 	if err != nil {
 		return newError(accountDisableErr, "Error disabling account: %v", err)
 	}
-	// Close dexConnection books.
-	for _, book := range dc.books {
-		book.close()
+	// Stop dexConnection books.
+	dc.cfgMtx.RLock()
+	for _, m := range dc.cfg.Markets {
+		// Emptry bookie's feeds map & stop close timers.
+		dc.booksMtx.Lock()
+		if b, found := dc.books[m.Name]; found {
+			b.feeds = make(map[uint32]*BookFeed, 1)
+			b.closeTimer.Stop()
+		}
+		dc.booksMtx.Unlock()
+
+		dc.stopBook(m.Base, m.Quote)
 	}
+	dc.cfgMtx.RUnlock()
 	// Disconnect and delete connection from map.
 	dc.connMaster.Disconnect()
 	delete(c.conns, addr)
--- a/client/core/account.go
+++ b/client/core/account.go
@@ -51,7 +51,9 @@ func (c *Core) AccountDisable(pw []byte, addr string) error {
 				close(f.C)
 			}
 			b.feeds = make(map[uint32]*BookFeed, 1)
-			b.closeTimer.Stop()
+			if b.closeTimer != nil {
+				b.closeTimer.Stop()
+			}
 			b.mtx.Unlock()
 		}
 		dc.booksMtx.Unlock()
@@ -61,7 +63,7 @@ func (c *Core) AccountDisable(pw []byte, addr string) error {
 	dc.cfgMtx.RUnlock()
 	// Disconnect and delete connection from map.
 	dc.connMaster.Disconnect()
-	delete(c.conns, addr)
+	delete(c.conns, host)
 
 	return nil
 }
--- a/client/core/account.go
+++ b/client/core/account.go
@@ -44,9 +44,9 @@ func (c *Core) AccountDisable(pw []byte, host string) error {
 	if err != nil {
 		return newError(accountDisableErr, "Error disabling account: %v", err)
 	}
-	// Stop dexConnection books.
-	for _, market := range dc.marketMap() {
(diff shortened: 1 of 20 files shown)


chappjc:
Also noting that it's required to eliminate an account for a server that no longer exists, or if the user's dexc.db has multiple DEX server accounts and they only want to delete one, making deleting the whole dexc.db file not an option.

client/core/account.go
```diff
@@ -7,6 +7,39 @@ import (
...
+	acctInfo, err := c.db.Account(dc.acct.host)
+	if err != nil {
+		return newError(accountRetrieveErr, "Error retrieving account: %v", err)
+	}
+	err = c.db.DisableAccount(acctInfo)
+	if err != nil {
+		return newError(accountDisableErr, "Error disabling account: %v", err)
+	}
+	c.conns[host].connMaster.Disconnect()
+	delete(c.conns, host)
```
chappjc, line 38: I know this is in draft, but a few observations: (1) you never unlock the connMtx, (2) you have `dc`, so no need to do `c.conns[host].connMaster.Disconnect()`, and (3) I don't think we want to do any of this while the dexConnection has active trades, so we should probably check first.

client/core/account.go
```diff
@@ -2,11 +2,54 @@ package core
...
 )
 
+// AccountDisable is used to disable an account by given host and application
+// password.
+func (c *Core) AccountDisable(pw []byte, host string) error {
+	// Validate password.
+	_, err := c.encryptionKey(pw)
+	if err != nil {
+		return codedError(passwordErr, err)
+	}
```
JoeGruffins, line 18: Just want to point out that validating the password is kinda superficial. One could comment out this code, compile, and get past that.

client/core/account.go
```diff
@@ -2,11 +2,54 @@ package core
...
+
+	// Parse address.
+	_, err = addrHost(host)
+	if err != nil {
+		return newError(addressParseErr, "error parsing address: %v", err)
+	}
```
JoeGruffins, line 24: I believe you should use the returned, parsed host here, as that is what is used as the key in the first place. addrHost is tested to always produce the same string on a second parsing as well.

client/core/account.go
```diff
@@ -2,11 +2,54 @@ package core
...
 	"encoding/hex"
+	"fmt"
 
 	"decred.org/dcrdex/client/db"
 	"github.com/decred/dcrd/dcrec/secp256k1/v3"
 )
 
+// AccountDisable is used to disable an account by given host and application
+// password.
+func (c *Core) AccountDisable(pw []byte, host string) error {
```
JoeGruffins, line 13: Could you make a `TestAccountDisable` in `account_test.go`?

client/core/account.go
amassarwi: Thought the same, any easy way to improve that ?

chappjc: I think the alternative is to not require the password.  Which should be OK if the auth middleware is working correctly, and cookies aren't stolen, and on the RPC server side they've auth'd properly there.
But we also request the app password for certain actions that are potentially irreversible, such as this one.  I'm ok either way for this function.

client/core/account.go
amassarwi: surething

client/core/account.go
JoeGruffins: Actually, this would stop a remote attacker, as they may not be able to swap your executable, so please leave it.

amassarwi: Yep. 
Weak protection better than no protection.

client/core/account.go
amassarwi: donez

JoeGruffins: Would like to suggest some changez.

https://github.com/amassarwi/dcrdex/pull/1/commits/e6fc9ca832bc0e45c6119004349023e1a2d5ae08

amassarwi: Great improvement, thanks @JoeGruffins.
I will open another PR to adjust the other account tests to follow same approach.

JoeGruffins:
After disabling:
```
[ERR] CORE[wss://127.0.0.1:17273/ws]: (*wsConn).Request(route 'unsub_orderbook') Send error (cannot send on a broken connection), unregistering msg ID 6 handler
[ERR] CORE: request error unsubscribing from dcr_btc orderbook: cannot send on a broken connection
```
Maybe needs to do a little more tear-down before deleting? Unsubscribe I guess?

amassarwi:
@JoeGruffins Addressed mate.

JoeGruffins:
I'm still seeing the error if I disable the account and wait for a while.

amassarwi:
@JoeGruffins 
hrmmm, lemme see

amassarwi:
@JoeGruffins It should be solved for good after https://github.com/decred/dcrdex/pull/985/commits/fb916b134648b0faacdd3a38a28b3573462c09cc

Needed to add a check in bookie's close func to skip re-closing the book connection if was already disconnected.

client/core/account.go
chappjc: I just want to point out that it's the job of say the RPC server (rpcuser/pass) and webserver (auth cookie) to deal with permissions, not the Core Go API, although requiring a password to make this harder is an existing pattern so I don't mind it staying here.

client/core/account.go
```diff
@@ -2,11 +2,59 @@ package core
...
+		return newError(accountRetrieveErr, "Error retrieving account: %v", err)
+	}
+	err = c.db.DisableAccount(acctInfo)
+	if err != nil {
+		return newError(accountDisableErr, "Error disabling account: %v", err)
+	}
+	// Close dexConnection books.
+	for _, book := range dc.books {
+		book.close()
+	}
```
chappjc, line 50: Need to rethink this because the range on the `dc.books` map allows for a concurrent map write in another goroutine since this doesn't lock it's mutex.  However since `book.close` is actually `stopBook`, which locks the mutex too, we can't simply lock it here (this an example of why I don't love hiding the true implementations as opaque functions like `bookie.close`).

Maybe `dc.stopBook` for each market in `dc.cfg.Markets` (holding `dc.cfgMtx`), similar to in `handleReconnect`.  So, similar to what you had before https://github.com/decred/dcrdex/commit/fb916b134648b0faacdd3a38a28b3573462c09cc, but not ranging on `dc.marketMap()` but `dc.cfg.Markets` instead.

client/core/bookie.go
```diff
@@ -194,7 +194,11 @@ func (dc *dexConnection) syncBook(base, quote uint32) (*BookFeed, error) {
 			return nil, err
 		}
 
-		booky = newBookie(dc.log.SubLogger(mktID), func() { dc.stopBook(base, quote) })
+		booky = newBookie(dc.log.SubLogger(mktID), func() {
+			if _, found := dc.books[mktID]; found {
+				dc.stopBook(base, quote)
+			}
+		})
```
chappjc, line 201: `stopBook` already does the check, and importantly it locks `dc.booksMtx`, which isn't happening with this new map access here.  So I believe this change isn't needed.

chappjc (changes requested):

chappjc:
Guys, I'm actually wondering if you were seeing `"Aborting booky %p unsubscribe for market %s with active feeds"` logged.  Because the reason a minute or later you saw `unsub_orderbook` error is because of the duplicate `StopBook` call from the `time.AfterFunc` started by `(*bookie).closeFeed` when `(*marketSyncer).Run` returns (because of `wsUnmarket` stopping it I think).

I would expect the  `"Aborting booky %p unsubscribe for market %s with active feeds"` warning if `stopBook` were called with active feeds, like the websocket from the browser, and that does seem to be the case.  I'm still looking into this... bookie.go is kinda hard to reason about.

chappjc:
Ah, so I think leaving the /markets page for the /settings page stopped the `marketSyncer` and that deleted that feed and set the `AfterFunc` to unsubscribe after a minute delay.  Then in /settings it stopped and unsubscribed, but then when the `AfterFunc` fired, it tried to unsubscribe again (via the close function there).  So it does seem like the code in disable account needs to both delete exitsing feeds form the feeds map AND stop any of these Timers.

amassarwi:
@chappjc  Updated https://github.com/decred/dcrdex/pull/985/commits/76bab86577d6dacb9c1de8cb7d57bb2cc1ed79eb, let me know if that's how you imagined it.

chappjc:
Pretty much @amassarwi, although the one newBookie call can go back to one line like it was at the start.  Will give you a proper review soon though so no need to push a diff yet.

client/core/account.go
```diff
@@ -2,11 +2,69 @@ package core
...
+	if err != nil {
+		return newError(accountDisableErr, "Error disabling account: %v", err)
+	}
+	// Stop dexConnection books.
+	dc.cfgMtx.RLock()
+	for _, m := range dc.cfg.Markets {
+		// Emptry bookie's feeds map & stop close timers.
+		dc.booksMtx.Lock()
+		if b, found := dc.books[m.Name]; found {
+			b.feeds = make(map[uint32]*BookFeed, 1)
```
buck54321, line 53: Shouldn't just discard the old feeds. I think you can iterate through `b.feeds` and `close` the `BookFeed.C` channels. Looks like we're handling that in websocket, but not in `loadbot`, which is fine for now since we won't be changing the app password there.

client/core/account.go
```diff
@@ -2,11 +2,69 @@ package core
...
+			b.closeTimer.Stop()
+		}
+		dc.booksMtx.Unlock()
```
buck54321, line 56: `booksMtx` is for `dc.books`. `b.mtx` is for the `bookie` fields.

client/core/account_test.go
```diff
@@ -88,6 +90,93 @@ func setupRigAccountProof(host string, rig *testRig) {
...
+		host:           tDexHost,
+		disableAcctErr: errors.New(""),
+		wantErr:        true,
+		wantErrCode:    true,
+		errCode:        accountDisableErr,
+	}}
+
+	for _, test := range tests {
+		rig := newTestRig()
+		tCore := rig.core
```
buck54321, line 149: It's good practice to call `rig.shutdown` at some point. Ideally, that would be at the end of the subtest, but at the end of `TestAccountDisable` is better than nothing.

client/core/core_test.go
```diff
@@ -307,7 +308,7 @@ func (tdb *TDB) CreateAccount(ai *db.AccountInfo) error {
 
 func (tdb *TDB) DisableAccount(ai *db.AccountInfo) error {
 	tdb.accts = nil
```
buck54321, line 310: From what I can tell, the `TDB.accts` field is completely unused. What would be better here is something like

```
tdb.disabledAcct = ai
```

and then you can check that the request got through in the tests.

buck54321 (changes requested):

client/core/account.go
```diff
@@ -2,11 +2,69 @@ package core
...
+	}
+
+	acctInfo, err := c.db.Account(dc.acct.host)
+	if err != nil {
+		return newError(accountRetrieveErr, "Error retrieving account: %v", err)
+	}
+	err = c.db.DisableAccount(acctInfo)
+	if err != nil {
+		return newError(accountDisableErr, "Error disabling account: %v", err)
+	}
```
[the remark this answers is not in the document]
amassarwi: yep

client/core/account.go
amassarwi: I'm iterating on dc.books here no ?

client/core/account.go
chappjc: We might to live with the websocket's `marketSyncer` erroring trying to unsubscribe because of the `defer m.feed.Close()` that is really a (very obfuscated with wrappers of wrappers) callback to `(*dexConnection).stopBook`  in `(*marketSyncer).Run`.  I was referring to this in my two comments [here](https://github.com/decred/dcrdex/pull/985#issuecomment-789832980).  One solution I can see is eliminating that defer and only having `(*marketSyncer).Run` do `m.feed.Close` on context cancellation, not a closed feed channel.  But @buck54321's suggestion is what I was saying with:

>  For consumers receiving on the BookFeed.C channel, it will also be necessary to close these channels after the removing the BookFeed from the feeds map.

Stopping the timer will prevent redundant unsubscribes from `marketSyncer`s that stopped before doing this disable function, but for active `marketSyncer`s, the channel close is needed so they can return.  I think making the `m.feed.Close` call in `(*marketSyncer).Run` not a defer could solve the issue of `stopBook` getting called when the channel is closed.

client/core/account.go
buck54321: You're reading and writing to `bookie` fields too

amassarwi: Oh I see

client/core/account.go
chappjc: To restate that, [`Run`](https://github.com/decred/dcrdex/blob/bea35a9e57536b4c3d3bd8215992710e89fae0eb/client/websocket/websocket.go#L239) returns in two ways:

1. `wsUnmarket` does `feedLoop.Stop`, which cancels the context
2. `update, ok := <-m.feed.C` where `!ok` (closed channel)

When you navigate away from the /markets page, such as when going to /settings, it does `wsUnmarket`.

But there can be other tabs or websocket consumers with active `marketSyncer`s (`Run`) going when you do disable account.  So to get them to return, gotta close that `feed.C` channel.

And because of how `m.feed.Close()` cascades all the way to `(*bookie).closeFeed`, it creates a new `closeTimer`, which in return can call `dexConnection.stopBook` and attempt to unsubscribe from long after loosing the connection to the dex server.

amassarwi:
@buck54321 @chappjc Updated

JoeGruffins (commented):
Something has made the depth graph stop working correctly and I think it's in 361efe5a95256c57aa1b15eb1374ba68aba3006c 
Try making a few orders but the graph will not show the changes unless you refresh the page.

amassarwi:
@JoeGruffins Just found that https://github.com/decred/dcrdex/commit/361efe5a95256c57aa1b15eb1374ba68aba3006c#diff-7f37696124d4f63a3aa051d69c0cd58a5b81ad862d9be5f58746d56445fdb431R239 causing the websocket bug,  where not only adding orders also canceling orders won't reflect on depth graph unless you refresh the page.

I guess the Close is happening too early with my latest changes.
@chappjc Do I need to make additional changes for the Close func to work properly without the defer ?

chappjc:
@amassarwi responded on the code (https://github.com/decred/dcrdex/commit/361efe5a95256c57aa1b15eb1374ba68aba3006c#r48311533) regarding the Close

amassarwi:
@JoeGruffins @chappjc Updated

client/websocket/websocket.go
```diff
@@ -257,9 +256,11 @@ out:
 				break out
 			}
 		case <-ctx.Done():
-			break out
+			// We are skipping m.feed.Close if the feed were closed (external sig).
+			return
 		}
 	}
+	m.feed.Close()
```
chappjc, line 263: Following the convo at https://github.com/decred/dcrdex/commit/361efe5a95256c57aa1b15eb1374ba68aba3006c#r48311533, I called out the wrong case for return with no `m.feed.Close()`.  We skip it in `case update, ok := <-m.feed.C` where `!ok`.  But also be sure that only the one case (external signal to close) does return and the others do `break out`.  Presently there's another return

client/core/account.go
chappjc: With the closed channel handling in `(*marketSyncer).Run` getting sorted out, don't forget to actually close the channel so `(*marketSyncer).Run` does hit the closed channel `case`.  This was @buck54321's suggestion.  I just wanted to be clear about what closing the channel does in the receiver.

JoeGruffins (approved):
Works for me.

I almost think it is too easy to disable though. If a user does this mistakenly, how to recover?

chappjc:
Noticed a double dialog when clicking "export".  Hard to explain so here's a screen cap vid:

https://user-images.githubusercontent.com/9373513/111687854-c6177f80-87f8-11eb-98dd-1478a73bcafa.mp4

amassarwi:
Double modals issue fixed.

amassarwi:
Added a red note:

image

chappjc (approved):
Works for me .

client/db/interface.go
```diff
@@ -29,7 +29,7 @@ type DB interface {
 	// CreateAccount saves the AccountInfo.
 	CreateAccount(ai *AccountInfo) error
 	// DisableAccount sets the AccountInfo disabled status to true.
-	DisableAccount(ai *AccountInfo) error
+	DisableAccount(url string) error
```
buck54321, line 32: ~~`url`~~ -> `host`

client/webserver/site/src/js/settings.js
```diff
@@ -124,6 +139,28 @@ export default class SettingsPage extends BasePage {
...
+      host
+    }
+    const loaded = app.loading(this.body)
+    const res = await postJSON('/api/disableaccount', req)
+    loaded()
+    if (!app.checkResponse(res)) {
+      page.disableAccountErr.textContet = res.msg
+      Doc.show(page.disableAccountErr)
+    }
+    Doc.hide(page.forms)
```
buck54321, line 159: A few issues here.

1. If you're going to display the error on the form, should should pass `true` for the second argument of `checkResponse`. That argument is called `skipNote`, and it prevents the error from going to the notification bell.
2. You are not actually populating the text of `disableAccountErr`, since you have a typo in the property name. `textContet` should be `textContent`
3. Even if you were populating `disableAccountErr` correctly, you are immediately hiding the form anyway, so nobody will ever see it. You'll need to `return` after `Doc.show(page.disableAccountErr)`.

buck54321 (changes requested):

amassarwi:
@buck54321 updated 👍

buck54321 (approved):

PR #986, opened 2021-02-24
README: update requirements

chappjc:
We need to specify that the 1.6 versions of dcrd and dcrwallet should be used.

This also makes Go 1.15 and Node.js 14 the minimum versions.

I was also considering bumping "At least 2 GB of available system memory" to 4 GB, but left it.  I feel like 2 GB for dcrd, dcrwallet, bitcoind, and dexc is pretty slim.


amassarwi (approved):
👍

Issue #952, opened 2021-01-28
core: need a refund path that doesn't rely on dex or wallet connections


buck54321:
This issue deals with the ability to properly refund a swap-gone-wrong. There are two parts to this issue.

1. **bug:** We don't load orders from the DB until we have already established a connection with the order's DEX. What happens when a DEX goes offline suddenly, causing swaps to fail, and doesn't come back? If a client is restarted before refund, they won't be able to refund until the DEX comes back online, potentially leaving the client's failed swap un-refunded beyond the lock time. This is a bug and a security hole that needs a solution.
2. Even if the DEX is connected, if the wallet is not unlocked, we can't refund. But if we generated the refund at the same time as the swap output, there would be no reason to have an unlocked wallet to refund. We'd just be broadcasting. In theory, we don't even need a wallet connection to refund, as we could devise a path to broadcast the refund to e.g. dcrdata without a wallet backend at all. But having a refund path on a locked wallet is a start, as there would be a conceivable system configuration in which a refund could happen without user input in the case of an unexpected restart.

chappjc:
A very dumbed-down stopgap solution for refund would be to generate/sign and log the hex of the raw refund txn at the time the swap is broadcast.  That way, in a pinch where even their dexc.db is toast, they can do the refund easily (without scraping for the script and using the atomicswap repo tools) like you said with dcrdata or other means of broadcast.

But that does not cover the case where maker covertly redeems, requiring taker to redeem manually too by searching out the secret key.  Issue 1 with server not being reachable would break that.

JoeGruffins:
I can work on this if noone else is atm.

chappjc:
I had started on this one actually.  I have a branch with some changes to order loading that facilitate this, and I'd like to see how it goes.  Will try to prioritize this.

chappjc:
A related thing I think we should be doing is creating and logging the raw refund txn when the swap transaction is created.  That way its dead simple to refund as a last resort even if dexc is dead. This would of course require dexc detecting when the user did the refund (spent contract), but that can be addressed.

JoeGruffins:
bump

chappjc:
> I had to update my psql databases and borked it. Not sure how to refund some swaps as I also lost the logs. On testnet so not a big deal, but seems important to get this done in case a dex went down. Would be panic mode for all users.

So you lost the psql *server* databases, but which logs did you also lose?

[...]

PR #989, opened 2021-02-26
Show exchange disconnected if connection fails

### LINKED ISSUE #943: ui: down DEX server leads to register dialog sequence
If you have dexc configured with a registered DEX account at some server, that is the only configured DEX account, and then startup cannot connect for any reason, the UI enters into the add-a-dex path.

image

image

But really it's because the already-configured dex couldn't connect. e.g.:

```
[ERR] CORE: error connecting to DEX 127.0.0.1:17273: dial tcp 127.0.0.1:17273: connect: connection refused
[INF] CORE: Successfully connected to 0 out of 1 DEX servers
```

Preferably the UI should say there was a connection failure and not prompt you to re-register (which wouldn't work even if you wanted to pay again because the DEX is actually unreachable in this situation).

vctt94:
closes #943

Right now I am appending the exchange host on `Core.conns[host]` in case of connection failure, instead of only appending it on succcess. This way it is possible to give a feedback in the client that the exchange is not connected.

~~Right now it stills a work in progress, because there is a loader which I am not being able to remove.~~

This is how it looks if a dex server connection fails:

image

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -122,8 +122,12 @@ func (dc *dexConnection) marketConfig(mktID string) *msgjson.Market {
 // [base]_[quote].
 func (dc *dexConnection) marketMap() map[string]*Market {
 	dc.cfgMtx.RLock()
-	mktConfigs := dc.cfg.Markets
+	cfg := dc.cfg
 	dc.cfgMtx.RUnlock()
+	if cfg == nil {
+		return nil
+	}
+	mktConfigs := cfg.Markets
 
 	marketMap := make(map[string]*Market, len(mktConfigs))
 	for _, mkt := range mktConfigs {
@@ -1069,9 +1073,14 @@ func (c *Core) dexConnections() []*dexConnection {
 func (c *Core) exchangeMap() map[string]*Exchange {
 	infos := make(map[string]*Exchange, len(c.conns))
 	for _, dc := range c.dexConnections() {
+		var requiredConfs uint32
 		dc.cfgMtx.RLock()
-		requiredConfs := uint32(dc.cfg.RegFeeConfirms)
+		cfg := dc.cfg
 		dc.cfgMtx.RUnlock()
+
+		if cfg != nil {
+			requiredConfs = uint32(cfg.RegFeeConfirms)
+		}
 
 		dc.assetsMtx.RLock()
 		assets := make(map[uint32]*dex.Asset, len(dc.assets))
@@ -3405,6 +3414,11 @@ func (c *Core) verifyAccount(acct *db.AccountInfo) bool {
 	dc, err := c.connectDEX(acct)
 	if err != nil {
 		c.log.Errorf("error connecting to DEX %s: %v", acct.Host, err)
+		// update failed connection, so we can give feedback.
+		c.connMtx.Lock()
+		c.conns[host] = dc
+		c.connMtx.Unlock()
+		c.log.Debugf("dex connection to %s failed", acct.Host)
 		return false
 	}
 	c.log.Debugf("connectDEX for %s completed, checking account...", acct.Host)
@@ -3458,6 +3472,14 @@ func (c *Core) checkUnpaidFees(dcrWallet *xcWallet) {
 // requisite confirmations for the 'notifyfee' message to be sent to the server.
 func (c *Core) reFee(dcrWallet *xcWallet, dc *dexConnection) {
 
+	// cfg can be null if connection to dex server fails.
+	dc.cfgMtx.RLock()
+	cfg := dc.cfg
+	dc.cfgMtx.RUnlock()
+	if cfg == nil {
+		return
+	}
+
 	// Return if the coin is already in blockWaiters.
 	regFeeAssetID, _ := dex.BipSymbolID(regFeeAssetSymbol)
 	if c.existsWaiter(coinIDString(regFeeAssetID, dc.acct.feeCoin)) {
@@ -3485,11 +3507,8 @@ func (c *Core) reFee(dcrWallet *xcWallet, dc *dexConnection) {
 		c.log.Errorf("reFee %s - error getting coin confirmations: %v", dc.acct.host, err)
 		return
 	}
-	dc.cfgMtx.RLock()
-	requiredConfs := uint32(dc.cfg.RegFeeConfirms)
-	dc.cfgMtx.RUnlock()
 
-	if confs >= requiredConfs {
+	if confs >= uint32(cfg.RegFeeConfirms) {
 		err := c.notifyFee(dc, acctInfo.FeeCoin)
 		if err != nil {
 			c.log.Errorf("reFee %s - notifyfee error: %v", dc.acct.host, err)
@@ -4055,12 +4074,24 @@ func (c *Core) connectDEX(acctInfo *db.AccountInfo) (*dexConnection, error) {
 
 	connMaster := dex.NewConnectionMaster(conn)
 	err = connMaster.Connect(c.ctx)
+	dc := &dexConnection{
+		WsConn:     conn,
+		log:        c.log,
+		connMaster: connMaster,
+		books:      make(map[string]*bookie),
+		acct:       newDEXAccount(acctInfo),
+		trades:     make(map[order.OrderID]*trackedTrade),
+		notify:     c.notify,
+	}
 	// If the initial connection returned an error, shut it down to kill the
 	// auto-reconnect cycle.
 	if err != nil {
+		dc.connected = 0
 		connMaster.Disconnect()
-		return nil, err
+		return dc, err
 	}
+	// if no error happened update dc.connected value.
+	dc.connected = 1
 
 	// Request the market configuration. Disconnect from the DEX server if the
 	// configuration cannot be retrieved.
@@ -4070,28 +4101,18 @@ func (c *Core) connectDEX(acctInfo *db.AccountInfo) (*dexConnection, error) {
 		connMaster.Disconnect()
 		return nil, fmt.Errorf("Error fetching DEX server config: %w", err)
 	}
+	dc.cfg = dexCfg
 
 	assets, epochMap, err := generateDEXMaps(host, dexCfg)
 	if err != nil {
 		return nil, err
 	}
+	dc.epoch = epochMap
+	dc.assets = assets
 
 	// Create the dexConnection and listen for incoming messages.
 	bTimeout := time.Millisecond * time.Duration(dexCfg.BroadcastTimeout)
-	dc := &dexConnection{
-		WsConn:       conn,
-		log:          c.log,
-		connMaster:   connMaster,
-		assets:       assets,
-		cfg:          dexCfg,
-		tickInterval: bTimeout / tickCheckDivisions,
-		books:        make(map[string]*bookie),
-		acct:         newDEXAccount(acctInfo),
-		trades:       make(map[order.OrderID]*trackedTrade),
-		notify:       c.notify,
-		epoch:        epochMap,
-		connected:    1,
-	}
+	dc.tickInterval = bTimeout / tickCheckDivisions
 
 	c.log.Debugf("Broadcast timeout = %v, ticking every %v", bTimeout, dc.tickInterval)
 
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -47,9 +47,13 @@ const (
 	// regConfirmationsPaid is used to indicate completed registration to
 	// (*Core).setRegConfirms.
 	regConfirmationsPaid uint32 = math.MaxUint32
+
 	// tickCheckDivisions is how many times to tick trades per broadcast timeout
 	// interval. e.g. 12 min btimeout / 8 divisions = 90 sec between checks.
 	tickCheckDivisions = 8
+	// defaultTickInterval is the tick interval used before the broadcast
+	// timeout is known (e.g. startup with down server).
+	defaultTickInterval = 30 * time.Second
 )
 
 var (
@@ -61,14 +65,35 @@ var (
 	syncTickerPeriod = 10 * time.Second
 )
 
+type dexTicker struct {
+	dur int64 // atomic
+	*time.Ticker
+}
+
+func newDexTicker(dur time.Duration) *dexTicker {
+	return &dexTicker{
+		dur:    int64(dur),
+		Ticker: time.NewTicker(dur),
+	}
+}
+
+func (dt *dexTicker) Reset(dur time.Duration) {
+	atomic.StoreInt64(&dt.dur, int64(dur))
+	dt.Ticker.Reset(dur)
+}
+
+func (dt *dexTicker) Dur() time.Duration {
+	return time.Duration(atomic.LoadInt64(&dt.dur))
+}
+
 // dexConnection is the websocket connection and the DEX configuration.
 type dexConnection struct {
 	comms.WsConn
-	connMaster   *dex.ConnectionMaster
-	log          dex.Logger
-	tickInterval time.Duration
-	acct         *dexAccount
-	notify       func(Notification)
+	connMaster *dex.ConnectionMaster
+	log        dex.Logger
+	acct       *dexAccount
+	notify     func(Notification)
+	ticker     *dexTicker
 
 	assetsMtx sync.RWMutex
 	assets    map[uint32]*dex.Asset
@@ -1922,10 +1947,13 @@ func (c *Core) GetFee(dexAddr string, certI interface{}) (uint64, error) {
 		Host: host,
 		Cert: cert,
 	})
+	if dc != nil {
+		// Stop (re)connect loop, which may be running even if err != nil.
+		defer dc.connMaster.Disconnect()
+	}
 	if err != nil {
 		return 0, codedError(connectionErr, err)
 	}
-	defer dc.connMaster.Disconnect()
 
 	dc.cfgMtx.RLock()
 	defer dc.cfgMtx.RUnlock()
@@ -1990,6 +2018,10 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
 		Cert: cert,
 	})
 	if err != nil {
+		if dc != nil {
+			// Stop (re)connect loop, which may be running even if err != nil.
+			dc.connMaster.Disconnect()
+		}
 		return nil, codedError(connectionErr, err)
 	}
 
@@ -2034,7 +2066,7 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
 	}
 
 	dc.cfgMtx.RLock()
-	fee := dc.cfg.Fee
+	fee, requiredConfs := dc.cfg.Fee, dc.cfg.RegFeeConfirms
 	dc.cfgMtx.RUnlock()
 
 	// Check that the fee is non-zero.
@@ -2074,10 +2106,6 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
 	}
 
 	c.updateAssetBalance(regFeeAssetID)
-
-	dc.cfgMtx.RLock()
-	requiredConfs := dc.cfg.RegFeeConfirms
-	dc.cfgMtx.RUnlock()
 
 	details := fmt.Sprintf("Waiting for %d confirmations before trading at %s", requiredConfs, dc.acct.host)
 	c.notify(newFeePaymentNote(SubjectFeePaymentInProgress, details, db.Success, dc.acct.host))
@@ -3355,16 +3383,20 @@ func (c *Core) initialize() {
 	if err != nil {
 		c.log.Errorf("Error retrieving accounts from database: %v", err) // panic?
 	}
+
+	// Start connecting to DEX servers.
+	var liveConns uint32
 	var wg sync.WaitGroup
 	for _, acct := range accts {
 		wg.Add(1)
 		go func(acct *db.AccountInfo) {
 			defer wg.Done()
-			if !c.verifyAccount(acct) {
-				return
+			if _, connected := c.connectAccount(acct); connected {
+				atomic.AddUint32(&liveConns, 1)
 			}
 		}(acct)
 	}
+
 	dbWallets, err := c.db.Wallets()
 	if err != nil {
 		c.log.Errorf("error loading wallets from database: %v", err)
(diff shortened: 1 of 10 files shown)


client/core/core.go
```diff
@@ -120,6 +120,9 @@ func (dc *dexConnection) marketConfig(mktID string) *msgjson.Market {
 // marketMap creates a map of this DEX's *Market keyed by name/ID,
 // [base]_[quote].
 func (dc *dexConnection) marketMap() map[string]*Market {
+	if !dc.connected {
```
chappjc, line 123: On master `connected` is a uint32 atomic, so you'll have to do similarly to other places like `atomic.LoadUint32(&dc.connected) == 1`

vctt94: Thanks for the review. Now the CI passes!

client/core/core.go
```diff
@@ -121,6 +121,9 @@ func (dc *dexConnection) marketConfig(mktID string) *msgjson.Market {
 // marketMap creates a map of this DEX's *Market keyed by name/ID,
 // [base]_[quote].
 func (dc *dexConnection) marketMap() map[string]*Market {
+	if atomic.LoadUint32(&dc.connected) == 0 {
+		return nil
```
chappjc, line 125: What's the motivation for this change?  Was there an existing bug?  It's not clear to me this is going to be a benign change in all cases.

vctt94: Right now when calling the `User` method on core, it calls the method  `exchangeMap`, which calls the method `marketMap` from the connection. 

As I am returning the connection now, when it fails to connect, I need this check, otherwise, `dc.cfg` will be nil, and break things.

JoeGruffins: If what you are really after is `dc.cfg` being nil or not, I would prefer checking that.

client/core/core.go
```diff
@@ -1069,9 +1072,12 @@ func (c *Core) dexConnections() []*dexConnection {
...
 	for _, dc := range c.dexConnections() {
-		dc.cfgMtx.RLock()
-		requiredConfs := uint32(dc.cfg.RegFeeConfirms)
-		dc.cfgMtx.RUnlock()
+		var requiredConfs uint32
+		if atomic.LoadUint32(&dc.connected) == 1 {
+			dc.cfgMtx.RLock()
+			requiredConfs = uint32(dc.cfg.RegFeeConfirms)
+			dc.cfgMtx.RUnlock()
+		}
```
JoeGruffins, line 1080: Here too could check that "dc.cfg" is not nil.

JoeGruffins (commented):
This looks to solve the issue to me. Maybe a little more info for the user would be useful? What should they do? I guess that host is now fubar for the life of this dexc? They should close the app and try again later?

vctt94:
Thanks for the review @JoeGruffins . I agree with you, and I think it is a good idea to have more information. The problem is this is a server problem, so there is not how to be sure how long will it take.

Maybe add a message for retry later? What do you guys think

client/core/core.go
```diff
@@ -121,6 +121,9 @@ func (dc *dexConnection) marketConfig(mktID string) *msgjson.Market {
 // marketMap creates a map of this DEX's *Market keyed by name/ID,
 // [base]_[quote].
 func (dc *dexConnection) marketMap() map[string]*Market {
+	if dc.cfg == nil {
+		return nil
+	}
 	dc.cfgMtx.RLock()
 	mktConfigs := dc.cfg.Markets
 	dc.cfgMtx.RUnlock()
```
buck54321, line 129: The `cfgMtx` protects the `dc.cfg` field. What you want here is something like

```
dc.cfgMtx.RLock()
cfg := dc.cfg
dc.cfgMtx.RUnlock()
if cfg = nil { ... }
mktConfigs := cfg.Markets
```

buck54321, line 129: The `cfgMtx` protects the `dc.cfg` field. What you want here is something like

```
dc.cfgMtx.RLock()
cfg := dc.cfg
dc.cfgMtx.RUnlock()
if cfg = nil { ... }
mktConfigs := cfg.Markets
```

client/core/core.go
```diff
@@ -1069,9 +1072,12 @@ func (c *Core) dexConnections() []*dexConnection {
...
 	for _, dc := range c.dexConnections() {
-		dc.cfgMtx.RLock()
-		requiredConfs := uint32(dc.cfg.RegFeeConfirms)
-		dc.cfgMtx.RUnlock()
+		var requiredConfs uint32
+		if dc.cfg != nil {
+			dc.cfgMtx.RLock()
+			requiredConfs = uint32(dc.cfg.RegFeeConfirms)
+			dc.cfgMtx.RUnlock()
+		}
```
buck54321, line 1080: Here again, can't check `if dc.cfg != nil` without the `cfgMtx` locked. The `cfgMtx` doesn't protect the fields of the stored `*msgjson.ConfigResult`, it protects the `dexConnection.cfg` field itself.

client/core/core.go
```diff
@@ -3405,6 +3411,11 @@ func (c *Core) verifyAccount(acct *db.AccountInfo) bool {
 	dc, err := c.connectDEX(acct)
 	if err != nil {
 		c.log.Errorf("error connecting to DEX %s: %v", acct.Host, err)
+		// update failed connection, so we can give feedback.
+		c.connMtx.Lock()
+		c.conns[host] = dc
+		c.connMtx.Unlock()
+		c.log.Debugf("dex connection to %s ready", acct.Host)
```
buck54321, line 3418: message seems misleading

client/core/core.go
```diff
@@ -3405,6 +3411,11 @@ func (c *Core) verifyAccount(acct *db.AccountInfo) bool {
 	dc, err := c.connectDEX(acct)
 	if err != nil {
 		c.log.Errorf("error connecting to DEX %s: %v", acct.Host, err)
+		// update failed connection, so we can give feedback.
+		c.connMtx.Lock()
+		c.conns[host] = dc
+		c.connMtx.Unlock()
```
buck54321, line 3417: In the case of an error, `dc` will be `nil`. Might as well just `c.conns[host] = nil` here so there's no question. Though if we're going to use a nil-valued entry like this, the intention needs to be very clearly documented somewhere.

client/core/core.go
```diff
@@ -4058,8 +4069,18 @@ func (c *Core) connectDEX(acctInfo *db.AccountInfo) (*dexConnection, error) {
...
+		dc := &dexConnection{
+			WsConn:     conn,
+			log:        c.log,
+			connMaster: connMaster,
+			books:      make(map[string]*bookie),
+			acct:       newDEXAccount(acctInfo),
+			trades:     make(map[order.OrderID]*trackedTrade),
+			notify:     c.notify,
+			connected:  0,
+		}
```
buck54321, line 4081: Rather than repeat this, might as well do this once up above, and just assign to the other fields directly in the non-error case.

client/webserver/site/src/html/markets.tmpl
```diff
@@ -42,6 +42,7 @@
     {{- /* CHART */ -}}
 
     <div id="marketChart" class="w-100">
+      <div id="errMsg" class="fs15 pt-3 text-center d-hide errcolor"></div>
```
buck54321, line 45:
```suggestion
      <div id="chartErrMsg" class="fs15 pt-3 text-center d-hide errcolor"></div>
```

client/webserver/site/src/js/app.js
```diff
@@ -498,8 +498,8 @@ export default class Application {
       case 'walletstate':
       case 'walletconfig': {
         const wallet = note.wallet
-        this.assets[wallet.assetID].wallet = wallet
-        this.walletMap[wallet.assetID] = wallet
+        if (this.assets !== undefined) this.assets[wallet.assetID].wallet = wallet
+        if (this.walletMap !== undefined) this.walletMap[wallet.assetID] = wallet
```
buck54321, line 502: Not sure about this. What are you fixing?

client/webserver/site/src/js/markets.js
```diff
@@ -476,6 +478,12 @@ export default class MarketsPage extends BasePage {
     const { confs, confsrequired } = dex
     const feePending = this.hasFeePending()
 
+    // if dex is not connected to server, is not possible to know fee
```
buck54321, line 481:
```suggestion
    // If dex is not connected to server, is not possible to know fee
```

buck54321, line 481:
```suggestion
    // If dex is not connected to server, is not possible to know fee
```

client/webserver/site/src/js/markets.js
```diff
@@ -476,6 +478,12 @@ export default class MarketsPage extends BasePage {
...
+    // registration status.
+    if (!dex.connected) {
+      return
+    }
```
buck54321, line 485:
```suggestion
    if (!dex.connected) return
```

client/webserver/site/src/js/markets.js
```diff
@@ -504,6 +512,19 @@ export default class MarketsPage extends BasePage {
   /* setMarket sets the currently displayed market. */
   async setMarket (host, base, quote) {
     const dex = app.user.exchanges[host]
+    if (!dex.connected) {
+      this.market = {
+        dex: dex
+      }
```
buck54321, line 518:
```suggestion
      this.market = { dex: dex }
```

client/webserver/site/src/js/markets.js
```diff
@@ -1612,6 +1633,10 @@ class MarketList {
   first () {
     const firstXC = this.sortedSections()[0]
     const firstMkt = firstXC.first()
+    // can not find markets if server connection failed.
+    if (!firstMkt) {
+      return makeMarket(firstXC.host)
+    }
```
buck54321, line 1639:
```suggestion
    if (!firstMkt)  return makeMarket(firstXC.host)
```

client/webserver/site/src/js/markets.js
```diff
@@ -1612,6 +1633,10 @@ class MarketList {
   first () {
     const firstXC = this.sortedSections()[0]
     const firstMkt = firstXC.first()
+    // can not find markets if server connection failed.
```
buck54321, line 1636:
```suggestion
    // Cannot find markets if server connection failed.
```

client/webserver/site/src/js/markets.js
```diff
@@ -1657,12 +1682,14 @@ class ExchangeSection {
     this.rows = Doc.tmplElement(box, 'mkts')
     const rowTmpl = Doc.tmplElement(this.rows, 'mktrow')
     this.rows.removeChild(rowTmpl)
-    for (const mkt of Object.values(dex.markets)) {
-      this.marketRows.push(new MarketRow(rowTmpl, mkt))
-    }
-
-    for (const market of this.sortedMarkets()) {
-      this.rows.appendChild(market.row)
+    // if disconnected is not possible to get the markets from the server.
```
buck54321, line 1685:
```suggestion
    // If disconnected is not possible to get the markets from the server.
```

client/webserver/site/src/js/markets.js
```diff
@@ -1657,12 +1682,14 @@ class ExchangeSection {
...
+    if (dex.markets) {
```
buck54321, line 1686: Prefer less indentation, so

```suggestion
    if (!dex.markets) return
```

buck54321 (changes requested):

client/core/core.go
vctt94: dc will not be nil in this case, because now I am returning dc's value if it is disconnect, as well: https://github.com/decred/dcrdex/pull/989/files#diff-13a4a6b12c1ae844ef90151638f194427cfdc545fceb58d52c0f6f33b81d859bR4072

buck54321: My bad. I meant to delete this comment before submitting my review.

buck54321: My bad. I meant to delete this comment before submitting my review.

vctt94:
Thanks for your reveiw buck. Updated your comments

client/webserver/site/src/js/app.js
vctt94: You are right. Removed it

JoeGruffins (commented):
It will show the message `Connection to dex server failed. You can close dexc and try again later.` even if this was a previously established connection that will reconnect automatically at some time.

imo it would be best if the connection were started in the reconnect mode, although still lacking a lot of initiation, that could be done whenever the connection does come up. There may be something making that more difficult than it sounds though, I am not sure.

I would think you could add a field to the dexConnection, a bool like `initiated` and pull out the last part of `connectDEX` into a separate function. If connect gets to the end successfully, `initiated` is true, if not keep trying to reconnect. reconnect has a switch for if `initiated` and if not go through the init procedure, which is the bottom half of `connectDEX`. However, if the init part fails still need to make this host fubar and stop trying to connect.

vctt94:
I am not sure how to reproduce this @JoeGruffins, can you say the steps?

If connected, and stopping the server, for example, it just shows as disconnected

chappjc:
@vctt94 I'm building on your branch presently w.r.t. the start in reconnect mode and some other tasks.  https://github.com/decred/dcrdex/commit/bec320c4
I'll have a number of other suggested changes for your branch shortly.

chappjc:
2 commits building on your branch: https://github.com/chappjc/dcrdex/commits/wsconn-reconnect-loop-start-err
In addition to the `client/comms` change, there were some other gotchas like the `(*Core).listen(*dexConnection)` goroutine needing to be running for which such a connection eventually does connect.  It's a bunch of tangential work though so I suggest we merge your PR and I follow up with these changes.

But about your changes, the browser console is very fussy about `markets` being `null` and `assets` being `{}`.  I know without a connection we just don't have this info, but perhaps some if checks in the js to prevent all the errors.

client/core/core.go
```diff
@@ -59,16 +56,17 @@ var (
 	// When waiting for a wallet to sync, a SyncStatus check will be performed
 	// ever syncTickerPeriod. var instead of const for testing purposes.
 	syncTickerPeriod = 10 * time.Second
+	// tickInterval is how often to check trades.
+	tickInterval = 60 * time.Second
```
chappjc, line 61: @buck54321 @itswisdomagain Reason for this is because tickInterval had an arbitrary formula based on the connected market's broadcast timeout.  Needing to have the `listen` goroutine, which also does the ticks, running from the start and thus before broadcast timeout is known made a problem.  The whole `tickCheckDivisions` thing was a bit questionable anyway.  I don't think I ever had a solid justification for a variable tick interval for markets with longer broadcast timeouts.

client/core/core.go
```diff
@@ -4208,22 +4206,18 @@ func (c *Core) handleReconnect(host string) {
...
-	c.connMtx.Lock()
-	if dc, found := c.conns[host]; found {
-		var v uint32
-		if connected {
-			v = 1
-		}
-		atomic.StoreUint32(&dc.connected, v)
+func (dc *dexConnection) handleConnectEvent(connected bool) {
+	var v uint32
+	if connected {
```
chappjc, line 4212: Just realized the two `if` statements can be merged.

chappjc:
Do you see any way to make this cleaner or hidden when in the disconnected state:

image

That is, when the depth chart is showing "Connection to dex server failed. You can close dexc and try again later."

Other comment is that that message now could be amended to say "You can close dexc and try again later **or wait for it to reconnect.**"  Although the `NoteTypeConnEvent` that comes when it does reconnect does not reload the page since usually a page reload isn't needed.  Not sure how frontend can change to deal with that.

client/webserver/site/src/js/app.js
```diff
@@ -462,6 +462,7 @@ export default class Application {
   notify (note) {
     // Handle type-specific updates.
     this.log('notes', 'notify', note)
+    console.log(note)
```
chappjc, line 465: leftovers

client/webserver/site/src/js/markets.js
```diff
@@ -1243,11 +1246,20 @@ export default class MarketsPage extends BasePage {
     }
   }
 
+  setBalanceVisibility () {
+    if (this.market.dex.connected) {
+      Doc.show(this.page.marketBalance)
+    } else {
+      Doc.hide(this.page.marketBalance)
+    }
+  }
```
chappjc, line 1255: This seems alright to me but in other places we show/hide the whole order form that includes the balance table (`Doc.{show,hide}(page.orderForm)`), such as when the account is not yet paid.  That actually happens in `setRegistrationStatusVisibility`, which is called right above the `setRegistrationStatusVisibility` call you added. @fernandoabolafio or buck could advise of that's more appropriate or not.

vctt94: I am also hiding the form now. 

I will wait to see if it is ok to check if to handleBalanceNote or setting connection status.

I am fine either way

JoeGruffins (approved):
Works great.

client/core/core.go
```diff
@@ -3395,33 +3410,40 @@ func (c *Core) initialize() {
...
-		return false
+		c.log.Errorf("skipping loading of %s due to address parse error: %v", host, err)
+		return
 	}
-	dc, err := c.connectDEX(acct)
-	if err != nil {
-		c.log.Errorf("error connecting to DEX %s: %v", acct.Host, err)
-		return false
+	dc, err = c.connectDEX(acct)
+	if dc == nil {
```
chappjc, line 3432: @buck54321 had a good point in another thread about using dc.acct.host instead of doing the superfluous `addrHost` call a few lines up since `connectDEX` does it too.

chappjc (changes requested):
One more minor request from me, and please follow up on any outstanding requests from Joe and myself.  I see two small ones.

client/core/core.go
buck54321: It wasn't arbitrary. The broadcast timeout is an upper limit on what our tick interval should be. Basing the tick interval on the broadcast timeout is a reasonable approach. But using a 60s constant **is** arbitrary. What if the broadcast timeout is 30s and we never even get a second `tick`? Clearly, the operator that set the timeout is an ass, but that's not really the point.

client/core/core.go
vctt94: aparently the rebase changed it

client/core/core.go
chappjc: I mean the number of ticks per timeout was arbitrary, and the fact that tick rate was variable was awkward imo.
A broadcast timeout on the order of seconds is crazy as you've pointed out.
Alternatives to this?  I don't have a solution.

chappjc: Ya know, the `(*Ticker).Reset(time.Duration)` method that was added in 1.15 is now fair game for us.... let me see about that.  There are still likely some challenges though as the ticker's going to have to be in a `dexConnection` field to be able to reset it from the reconnect handler.

client/webserver/site/src/js/app.js
```diff
@@ -496,6 +496,8 @@ export default class Application {
         break
       case 'walletstate':
       case 'walletconfig': {
+        // assets can be null if failed to connect to dex server.
+        if (!this.assets) return
```
buck54321, line 500: The `Assets` field of the `User` object should be populated regardless of any particular dex connection status. Are you sure this comment is true?

client/webserver/site/src/js/app.js
```diff
@@ -508,6 +510,9 @@ export default class Application {
         if (ord) updateMatch(ord, note.match)
         break
       }
+      case 'conn': {
+        this.reconnected()
+      }
```
buck54321, line 515: No need for braces is you're not creating variables. 

```
case 'conn':
  this.reconnected()
```

client/webserver/site/src/js/markets.js
```diff
@@ -66,7 +68,9 @@ export default class MarketsPage extends BasePage {
       // Max order section
       'maxOrd', 'maxLbl', 'maxFromLots', 'maxFromAmt', 'maxFromTicker',
       'maxToAmt', 'maxToTicker', 'maxAboveZero', 'maxLotBox', 'maxFromLotsLbl',
-      'maxBox'
+      'maxBox',
+      // errors
+      'chartErrMsg'
```
buck54321, line 73: Please put this in the `// Chart and legend` section.

client/webserver/site/src/js/markets.js
```diff
@@ -47,6 +47,8 @@ export default class MarketsPage extends BasePage {
       'baseWalletIcons', 'quoteWalletIcons', 'lotSize', 'rateStep', 'lotField',
       'tifNow', 'mktBuyBox', 'mktBuyLots', 'mktBuyField', 'minMktBuy', 'qtyBox',
       'loaderMsg', 'balanceTable', 'orderPreview',
+      // Balance
+      'marketBalance',
```
buck54321, line 51: No new section here. It's the same section as the `balanceTable`, which is the section labeled `// Order form`.

buck54321 (approved):
Just couple nits.

client/core/core.go
buck54321: Oh, nice. I didn't even know about `(*Ticker).Reset`.

client/webserver/site/src/js/app.js
vctt94: If the server is off, it is not possible to know which assets they have, or is it?

buck54321: `this.assets` is just `this.user.assets`, which comes from the Go `core.User` type. The `Assets` field is a `map[uint32]*SupportedAsset`, which are assets that the client supports, not `dex.Asset`, which would be associated with a DEX. The big clue is that it is a field of `User`, not of `Exchange`. Look at `(*Core).assetMap` to see how the `User.Assets` map is made.

chappjc:
Generally working quite well.  The one thing left for a follow-up PR that I'm preparing already is to allow a logged in dexc to start working immediately and without restart.  Right now it can't load orders into trackedTrades in the following case: dexc user has active orders, dcrdex server is down, dexc shuts down, dexc starts back up, dexc user does login (unlocking wallet and account keys), dcrdex server comes back online.  In this case it does reconnect as expected, but the trackedTrades are not loaded fully and because information from the server was required to load them, so dexc logs unknown orders reported by the server on auth.  A dexc restart resolves it, but this is next work for me.

Note that the solution to the above is tantamount to solving https://github.com/decred/dcrdex/issues/952.  I'm getting pretty close but it's not terribly pretty yet.

PR #991, opened 2021-03-01
client/core: put the reg fee payment note back

chappjc:
This resolves an intermittent failure in `TestRegister` caused by `verifyRegistrationFee` racing with a subsequent notification.


client/core/core.go
```diff
@@ -2070,14 +2070,14 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
 	requiredConfs := dc.cfg.RegFeeConfirms
 	dc.cfgMtx.RUnlock()
 
+	details := fmt.Sprintf("Waiting for %d confirmations before trading at %s", requiredConfs, dc.acct.host)
+	c.notify(newFeePaymentNote(SubjectFeePaymentInProgress, details, db.Success, dc.acct.host))
+
```
chappjc, line 2075: Since `verifyRegistrationFee` is async, having this `newFeePaymentNote` after it creates a race with the notification send by `verifyRegistrationFee`.

client/core/core_test.go
```diff
@@ -6080,6 +6080,7 @@ func TestParseCert(t *testing.T) {
 
 func TestPreOrder(t *testing.T) {
 	rig := newTestRig()
+	defer rig.shutdown()
```
chappjc, line 6083: This is pretty inconsequential as the last test, but added anyway.  Although I wonder if there's an issue with this test still given the logs:

```
[TRC] TCORE: notify: |DATA| (walletstate)
[DBG] TCORE[book]: Processing 0 cached order notes
[ERR] TCORE[book]: notification received out of sync. 1 != 0 - 1
[ERR] TCORE[book]: notification received out of sync. 1 != 0 - 1
[INF] TCORE: Connecting wallet for dcr
[ERR] TCORE: Error connecting to dcr wallet: test error
[INF] TCORE: Connecting wallet for btc
[ERR] TCORE: Error connecting to btc wallet: test error

```

JoeGruffins (approved):

client/core/core.go
buck54321: I can't recall, but this seems fine.

buck54321 (approved):

buck54321:
Actually, I meant to note before approving that I had to `go mod tidy` to get **loadbot** working to run.

PR #995, opened 2021-03-02
Use configPath rpc.cert if set

vctt94:
Use rpc.cert from config path, if set

### DIFF
--- a/server/asset/dcr/config.go
+++ b/server/asset/dcr/config.go
@@ -65,6 +65,8 @@ func loadConfig(configPath string, network dex.Network) (*config, error) {
 	// If no path provided, use default dcrd path.
 	if configPath == "" {
 		configPath = defaultConfigPath
+	} else {
+		defaultRPCCert = filepath.Join(filepath.Dir(configPath), "rpc.cert")
 	}
 
 	if _, err := os.Stat(configPath); os.IsNotExist(err) {


JoeGruffins (commented):
imo, not needed because one can set `cfg.RPCCert` separately. If something like this is needed, maybe an `appdata` option would be more standard considering other decred executables?

chappjc:
@JoeGruffins The appdata part of that is something I suggested as well.  One scenario is: user's config file specifies a non-default `appdata`, but **no** `rpccert`, so their dcrd is going to be using a cert in `appdata`/rpc.cert, not [configPath]/rpc.cert unless configPath is equal to `appdata`, and that's not guaranteed.  While the suggestion that they should just set `rpccert` in their dcrd.config file, thus setting `cfg.RPCCert`, solves it, it is arguably incorrect for dex to simply ignore the custom `appdata`.  (But this PR doesn't address this.)

The different scenario @vctt94 is trying to address is: user's config file does **not** change the `appdata`, and does **not** specify `rpccert`.  
So, @vctt94 in this second scenario that you hit, where was `appdata` (you said not default) and how did you set it with dcrd?  Seems like the same solution to the first scenario (look in `appdata`/rpc.cert if `cfg.RPCCert` is not set) would work for this too.

server/asset/dcr/config.go
```diff
@@ -65,6 +65,8 @@ func loadConfig(configPath string, network dex.Network) (*config, error) {
 	// If no path provided, use default dcrd path.
 	if configPath == "" {
 		configPath = defaultConfigPath
+	} else {
+		defaultRPCCert = filepath.Join(filepath.Dir(configPath), "rpc.cert")
```
chappjc, line 69: If the config file doesn't specify `rpccert`, what should dcrdex do? Options are to look in either: (a) config file folder, or (b) any `appdata` set by the config file.  I think only (b) is correct in this case, although (a) seemed reasonable.

If the config file does not specify an `appdata`, that must mean either dcrd is using the default `appdata` OR the user started dcrd with `--appdata`.  If they started with `--appdata`, I don't think we can work around that. They will just have to edit the config file they provide to dcrdex.

So, I think the only "right" change to make now is: if no rpccert set, see if there is a custom `appdata` set (regardless of where `configPath` comes from), and reset `defaultRPCCert` to look there.  We could argue that if neither `rpccert` or `appdata` are set, they it should adjust as with this diff in the `configPath`, but that's not necessarily correct, and the resolution is simply to set one of those options in the config file.

PR #996, opened 2021-03-02
webserver: add isinitialized web api endpoint

buck54321:
Since the `/user` route is behind the auth cookie now, web API users have no way to know whether the client is already initialized. 

We already had the exported `Core` method, we just needed to expose it.

### DIFF
--- a/client/webserver/api.go
+++ b/client/webserver/api.go
@@ -275,6 +275,22 @@ func (s *WebServer) apiInit(w http.ResponseWriter, r *http.Request) {
 	s.actuallyLogin(w, r, login)
 }
 
+// apiIsInitialized is the handler for the '/isinitialized' request.
+func (s *WebServer) apiIsInitialized(w http.ResponseWriter, r *http.Request) {
+	inited, err := s.core.IsInitialized()
+	if err != nil {
+		s.writeAPIError(w, "isinitialized error: %v", err)
+		return
+	}
+	writeJSON(w, &struct {
+		OK          bool `json:"ok"`
+		Initialized bool `json:"initialized"`
+	}{
+		OK:          true,
+		Initialized: inited,
+	}, s.indent)
+}
+
 // apiLogin handles the 'login' API request.
 func (s *WebServer) apiLogin(w http.ResponseWriter, r *http.Request) {
 	login := new(loginForm)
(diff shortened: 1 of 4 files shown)


chappjc:
There's certainly no intuitive or clean way to check this anyway, so this route makes sense, but here's what I came up with:

If I start fresh and do /api/init:

```
curl -v -X POST -d '{"pass":"asdf"}' -H 'content-type: application/json' http://127.0.0.3:5758/api/init
...
< HTTP/1.1 200 OK
...
< Content-Length: 23
< 
{"ok":true,"notes":[]}
```

And if I do it again, still responds with `HTTP/1.1 200 OK`, but the payload is now
`{"ok":false,"msg":"initialization error: already initialized, login instead"}`

Now I can do _login_ and it keeps returning `{"ok":true,"notes":[]}`.

As an asside, /api/login gives the cookie. e.g.
`< Set-Cookie: dexauth=c174add2ea2c939393ba7f190d4d047fe9e1043a4f4062ae04e933159b9954f7; Path=/; SameSite=Strict`
then you can hit the endpoints that are behind the rejectUnauthed middleware with the cookie like:
`curl -v -b "dexauth=c174add2ea2c939393ba7f190d4d047fe9e1043a4f4062ae04e933159b9954f7" http://127.0.0.3:5758/api/user`

So I believe there are two exiting, but awkward options, both of which have a request body that requires an expected or intended app password:
1. hit /api/login and if it says 428, the app is not initialized
2. hit /api/init and if it responds with a payload {"ok":false,"msg":"initialization error: already initialized, login instead"}, then app is initialized
I suppose the latter should also return a 400 level status code too, but it's 200 presently

chappjc (approved):
Looks good.  Holding off for a little while we chat with @alexlyp to get it straight.

chappjc:
```
$    curl -v http://127.0.0.3:5758/api/isinitialized
...
< HTTP/1.1 200 OK
...
{"ok":true,"initialized":false}

$    curl -v -X POST -d '{"pass":"asdf"}' -H 'content-type: application/json' http://127.0.0.3:5758/api/init
...
< HTTP/1.1 200 OK
...
{"ok":true,"notes":[]}

$    curl -v http://127.0.0.3:5758/api/isinitialized
...
< HTTP/1.1 200 OK
...
{"ok":true,"initialized":true}
```

PR #999, opened 2021-03-03
experiment: demo electron integration

buck54321:
This PR is to demonstrate a deep Electron app integration. Key points are

1. `Core` is compiled and imported as a node module (see, for example, [this tutorial](https://morioh.com/p/cfe8050589db)), precluding the need to run a separate executable and allowing us to use `Core` directly in the Electron main process. 
2. You can work with `Core` before even starting an HTTP server, e.g. initialization, wallet creation, dex registration. Once all of that is done, you can start the server and point your window at it. With some minor changes, we could probably even get around HTTP altogether by implementing a custom `net.Listener` and `net.Conn` for the `Server` to use.

**The demo is for Linux**, but should be portable without too much effort. To run, go to `client/cmd/electron-demo` and run `make build run`

On some Linux, you may see a message about a file called `chrome-sandbox`. You'll need to tighten permissions.

```
chmod 4755 node_modules/electron/dist/chrome-sandbox
sudo chown root:root node_modules/electron/dist/chrome-sandbox
```

**If you have the dcrdex simnet harnesses running** for `dcrdex`, `dcr`, and `btc`, you can also use the full demo where the Electron app prepares the wallets and registers with the DEX. Just `make build run-simnet`.

This is just for demo for now, but I can clean this up for merging if desired. 

I'm using **node version v15.8.0**

### DIFF
--- a/client/cmd/electron-demo/index.js
+++ b/client/cmd/electron-demo/index.js
@@ -1,0 +1,162 @@
+const { ipcRenderer } = require('electron')
+const path = require('path')
+
+Mainnet = 0
+Testnet = 1
+Simnet = 2
+
+NetName = {
+    [Mainnet]: Mainnet,
+    [Testnet]: Testnet,
+    [Simnet]: Simnet
+}
+
+LogLevelTrace = 0
+LogLevelDebug = 1
+LogLevelInfo = 2
+LogLevelWarn = 3
+LogLevelError = 4
+LogLevelCritical = 5
+LogLevelOff = 6
+
+DCR_ID = 42
+BTC_ID = 0
+
+function appGetPath (name) {
+    return fetchIPC('getPath', name)
+}
+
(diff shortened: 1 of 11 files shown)


client/cmd/electron-demo/index.js
```diff
@@ -0,0 +1,162 @@
...
+        const serverCertPath = path.join(dextestDir, 'dcrdex', 'rpc.cert')
+        DEX.register(simnetDexAddr, pw, defaultRegFee, serverCertPath)
+    }
+}
+
+async function run () {
+    let net = Mainnet
+    const args = fetchIPC('cmdArgs', '')
+    if (args.indexOf('--simnet') > -1) net = Simnet
+    else if (args.indexOf('--testnet') > -1) net = Simnet
```
chappjc, line 145: `Testnet`

Issue #835, opened 2020-11-14
zero qty orders in chart and orders


chappjc:
Leaving the markets page open for a long while, returning to this:

![image](https://user-images.githubusercontent.com/9373513/98888603-e788e900-245d-11eb-9dc0-b57bf07aab5d.png)

F5:

![image](https://user-images.githubusercontent.com/9373513/98888612-ebb50680-245d-11eb-8e97-04b4894e3977.png)

Both the depth chart and the orders lists have indicate something strange with the order feed.

**v0.1.1 release**

_Originally posted by @chappjc in https://github.com/decred/dcrdex/issues/824#issuecomment-725791277_

chappjc:
The symptom seems to be zero qty sell orders building up in the list

![image](https://user-images.githubusercontent.com/9373513/99543435-c9008180-2978-11eb-8216-313fd3918b42.png)

and the depth chart getting a zero quantity level on the buy side.

amassarwi:
will be working on that after my current task.

chappjc:
Got it while testing LTC-DCR

![image](https://user-images.githubusercontent.com/9373513/172652852-b822ae1d-e4f4-4dc2-92d0-346cbf298f03.png)

But only on the depth chart, not the tables.

chappjc:
Sort of resolved by https://github.com/decred/dcrdex/pull/1543

PR #1000, opened 2021-03-04
multi: fee rate estimate feed

buck54321:
Add the most recent fee rate estimate to 1) The `orderbook` response, 2) the `epoch_report` notification, and 3) a custom `fee_rate` endpoint. There are two major reason for this.

1. SPV clients will have no way to estimate fees. They can use the suggested fee rate with some basic sanity checks.
2. This finally closes a small vulnerability regarding fee rates on zero-conf funding coins. Previously, we handled this only client side by ensuring a reasonably high fee rate. But since the server's latest fee rate is available at the client now, we can more justifiably reject low-fee zero-conf coin-funded orders. Right now, the filter is set to >= 90% of the most recently procured rate estimate.

Order estimates should also be more accurate now.

### DIFF
--- a/client/asset/interface.go
+++ b/client/asset/interface.go
@@ -90,7 +90,7 @@ type Wallet interface {
 	// fees are an estimate based on current network conditions, and will be <=
 	// the fees associated with the Asset.MaxFeeRate. For quote assets, lotSize
 	// will be an estimate based on current market conditions.
-	MaxOrder(lotSize uint64, nfo *dex.Asset) (*SwapEstimate, error)
+	MaxOrder(lotSize, feeSuggestion uint64, nfo *dex.Asset) (*SwapEstimate, error)
 	// PreSwap gets a pre-swap estimate for the specified order size.
 	PreSwap(*PreSwapForm) (*PreSwap, error)
 	// PreRedeem gets a pre-redeem estimate for the specified order size.
@@ -273,6 +273,13 @@ type Redemption struct {
 // expanded in in-progress work to accommodate order-time options.
 type RedeemForm struct {
 	Redemptions []*Redemption
+	// FeeSuggestion is a suggested fee from the server. For redemptions, the
+	// suggestion is provided as a convenience for wallets without full
+	// chain backing which cannot get an estimate otherwise. Since this is the
+	// redemption, there is no obligation on the client to use the fee
+	// suggestion in any way, but obviously fees that are too low may result in
+	// the redemption getting stuck in mempool.
+	FeeSuggestion uint64
 }
 
 // Order is order details needed for FundOrder.
@@ -297,6 +304,10 @@ type Order struct {
 	// standing order, likely a market order or a limit order with immediate
 	// time-in-force.
 	Immediate bool
+	// FeeSuggestion is a suggested fee from the server. If a split transaction
+	// is used, the fee rate used should be at least the suggested fee, else
+	// zero-conf coins might be rejected.
+	FeeSuggestion uint64
 }
 
 // SwapEstimate is an estimate of the fees and locked amounts associated with
@@ -346,6 +357,10 @@ type PreSwapForm struct {
 	// standing order, likely a market order or a limit order with immediate
 	// time-in-force.
 	Immediate bool
+	// FeeSuggestion is a suggested fee from the server. If a split transaction
+	// is used, the fee rate used should be at least the suggested fee, else
+	// zero-conf coins might be rejected.
+	FeeSuggestion uint64
 }
 
 // SwapOption is an OrderEstimate and it's associated ConfigOption.
@@ -370,6 +385,8 @@ type PreRedeemForm struct {
 	LotSize uint64
 	// Lots is the number of lots in the order.
 	Lots uint64
+	// FeeSuggestion is a suggested fee from the server.
+	FeeSuggestion uint64
 }
 
 // PreRedeem is an estimate of the fees for redemption. The struct will be
--- a/client/core/trade.go
+++ b/client/core/trade.go
@@ -118,6 +118,14 @@ type trackedCancel struct {
 // The trackedTrade has methods for handling requests from the DEX to progress
 // match negotiation.
 type trackedTrade struct {
+	// redeemFeeSuggestion is cached fee suggestion for redemption. We can't
+	// request a fee suggestion at redeem time because it would require making
+	// the full redemption routine async (TODO?). This fee suggestion is
+	// intentionally not stored as part of the db.OrderMetaData, and should
+	// be repopulated if the client is restarted.
+	// Used with atomic. Leave redeemFeeSuggestion as the first field in
+	// trackedTrade.
+	redeemFeeSuggestion uint64
 	order.Order
 	// mtx protects all read-write fields of the trackedTrade and the
 	// matchTrackers in the matches map.
@@ -1037,6 +1045,9 @@ func (c *Core) tick(t *trackedTrade) (assetMap, error) {
 	// Check all matches for and resend pending requests as necessary.
 	c.resendPendingRequests(t)
 
+	// Make sure we have a redemption fee suggestion cached.
+	c.cacheRedemptionFeeSuggestion(t)
+
 	// Check all matches and send swap, redeem or refund as necessary.
 	var sent, quoteSent, received, quoteReceived uint64
 	for _, match := range t.matches {
@@ -1594,10 +1605,18 @@ func (c *Core) redeemMatchGroup(t *trackedTrade, matches []*matchTracker, errs *
 		})
 	}
 
+	// Don't use (*Core).feeSuggestion here, since can incur an RPC request.
+	// If we don't have a synced book, use t.redemption
+	feeSuggestion := t.dc.bestBookFeeSuggestion(t.wallets.toAsset.ID)
+	if feeSuggestion == 0 {
+		feeSuggestion = atomic.LoadUint64(&t.redeemFeeSuggestion)
+	}
+
 	// Send the transaction.
 	redeemWallet, redeemAsset := t.wallets.toWallet, t.wallets.toAsset // this is our redeem
 	coinIDs, outCoin, fees, err := redeemWallet.Redeem(&asset.RedeemForm{
-		Redemptions: redemptions,
+		Redemptions:   redemptions,
+		FeeSuggestion: feeSuggestion,
 	})
 	// If an error was encountered, fail all of the matches. A failed match will
 	// not run again on during ticks.
--- a/client/orderbook/orderbook.go
+++ b/client/orderbook/orderbook.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"fmt"
 	"sync"
+	"sync/atomic"
 
 	"decred.org/dcrdex/dex"
 	"decred.org/dcrdex/dex/msgjson"
@@ -22,6 +23,15 @@ type Order struct {
 	Time     uint64
 	// Epoch is only used in the epoch queue, otherwise it is ignored.
 	Epoch uint64
+}
+
+// copy creates a copy. Note that the OrderID is not a deep copy.
+func (o *Order) copy() *Order {
+	return &(*o)
+}
+
+func (o *Order) sell() bool {
+	return o.Side == msgjson.SellOrderNum
 }
 
 // RemoteOrderBook defines the functions a client tracked order book
@@ -44,23 +54,34 @@ type cachedOrderNote struct {
 
 // OrderBook represents a client tracked order book.
 type OrderBook struct {
-	log          dex.Logger
-	seqMtx       sync.Mutex
-	seq          uint64
-	marketID     string
-	noteQueue    []*cachedOrderNote
+	log      dex.Logger
+	seqMtx   sync.Mutex
+	seq      uint64
+	marketID string
+
 	noteQueueMtx sync.Mutex
-	orders       map[order.OrderID]*Order
-	ordersMtx    sync.Mutex
-	buys         *bookSide
-	sells        *bookSide
-	syncedMtx    sync.Mutex
-	synced       bool
+	noteQueue    []*cachedOrderNote
+
+	ordersMtx sync.Mutex
+	orders    map[order.OrderID]*Order
+
+	buys  *bookSide
+	sells *bookSide
+
+	syncedMtx sync.Mutex
+	synced    bool
 
 	epochMtx     sync.Mutex
 	currentEpoch uint64
 	proofedEpoch uint64
 	epochQueues  map[uint64]*EpochQueue
+
+	// feeRates is a separate struct to account for atomic field alignment in
+	// 32-bit systems. See also https://golang.org/pkg/sync/atomic/#pkg-note-BUG
+	feeRates struct {
+		base  uint64
+		quote uint64
+	}
 }
 
 // NewOrderBook creates a new order book.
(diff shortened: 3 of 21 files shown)


server/market/market.go
```diff
@@ -2113,3 +2177,31 @@ func (m *Market) orderResponse(oRecord *orderRecord) (*msgjson.Message, error) {
 	// Encode the order response as a message for the client.
 	return msgjson.NewResponse(oRecord.msgID, res, nil)
 }
+
+// SetFeeRateScale sets a swap fee scale factor for the given asset.
+// SetFeeRateScale should be called regardless of whether the Market is
+// suspended.
+func (m *Market) SetFeeRateScale(assetID uint32, scale float64) {
+	m.feeScalesMtx.Lock()
+	m.feeScales[assetID] = scale
```
chappjc, line 2186: I think I might have `feeScales [2]float64` or even `baseFeeScale, quoteFeeScale uint32` instead of a map, but it's not unpalatable as a map.

client/orderbook/orderbook.go
```diff
@@ -44,23 +54,30 @@ type cachedOrderNote struct {
...
+	syncedMtx sync.Mutex
+	synced    bool
 
 	epochMtx     sync.Mutex
 	currentEpoch uint64
 	proofedEpoch uint64
 	epochQueues  map[uint64]*EpochQueue
+
+	baseFeeRate  uint64
+	quoteFeeRate uint64
```
chappjc, line 80: Will need these as the first OrderBook fields because of the sync/atomic bug on 32-bit platforms.

server/market/orderrouter.go
```diff
@@ -593,6 +618,29 @@ func (r *OrderRouter) handleMarket(user account.AccountID, msg *msgjson.Message)
...
 
+// Check the FundingCoin confirmations, and if zero, ensure the tx fee rate
+// is sufficient, > 90% of our last recorded estimate for the asset.
+func (r *OrderRouter) checkZeroConfs(dexCoin asset.FundingCoin, fundingAsset *asset.BackedAsset) *msgjson.Error {
+	// Verify that zero-conf coins are within 10% of the last known fee
+	// rate.
+	confs, err := coinConfirmations(dexCoin)
+	if err != nil {
+		log.Debugf("Confirmations error for %s coin %s: %v", fundingAsset.Symbol, dexCoin, err)
+		return msgjson.NewError(msgjson.FundingError, fmt.Sprintf("failed to verify coin %v", dexCoin))
```
chappjc, line 629: I believe the `checkZeroConfs` calls in `handleLimit` and `handleMarket` will need to be moved into the TryFunc given to the `latencyQ` waiter's `TryFunc`.  Otherwise it'll reject txsplit orders pretty reliably, esp. with realistic network conditions.

chappjc: OMG, misread, sorry!

client/core/trade.go
```diff
@@ -1599,7 +1599,8 @@ func (c *Core) redeemMatchGroup(t *trackedTrade, matches []*matchTracker, errs *
 	// Send the transaction.
 	redeemWallet, redeemAsset := t.wallets.toWallet, t.wallets.toAsset // this is our redeem
 	coinIDs, outCoin, fees, err := redeemWallet.Redeem(&asset.RedeemForm{
-		Redemptions: redemptions,
+		Redemptions:   redemptions,
+		FeeSuggestion: c.feeSuggestion(t.dc, t.dc.bookie(t.mktID), redeemWallet.AssetID, !t.Trade().Sell),
```
chappjc, line 1603: The `feeSuggestion` call can make a request to the server.  That's a problem in `redeemMatchGroup`.  We split out the actual init and redeem requests to make them asynchronous (`sendInitAsync` and `sendRedeemAsync`) because of the trouble this can cause, but ideally we'd have no requests at all for this. All of the automated (tick or block based) actions need to be snappy and not have the potential to block.  Is there any way to avoid this here, either changing `feeSuggestion` or ensuring there is already a value ready somewhere for this?

server/market/orderrouter.go
```diff
@@ -593,6 +618,30 @@ func (r *OrderRouter) handleMarket(user account.AccountID, msg *msgjson.Message)
...
+	}
+	if confs > 0 {
+		return nil
+	}
+	lastKnownFeeRate := r.feeSource.LastFeeRate(fundingAsset.ID)
+	feeMinimum := uint64(math.Round(float64(lastKnownFeeRate) * 0.9))
```
chappjc, line 635: 0.9 seems pretty reasonable.  Let's make this a named constant and describe.

Also what do you think about a minimum difference allowed?  e.g. say the last was 15, then `feeMinimum` is 14 (round(15*0.9 = 13.5)), and it makes me wonder if that's too tight tolerance toward the smaller end.
Obviously we don't want things slowed down, but blocking incoming orders would suck?

Or if all client eventually update would there theoretically be no difference if they all use the fee suggestion from one of the various sources?

server/market/bookrouter.go
```diff
@@ -608,6 +636,34 @@ func (r *BookRouter) handleUnsubOrderBook(conn comms.Link, msg *msgjson.Message)
...
+			Message: "unknown asset",
+		}
+	}
+	resp, err := msgjson.NewResponse(msg.ID, atomic.LoadUint64(rate), nil)
+	if err != nil {
+		log.Errorf("failed to encode fee_rate response: %v", err)
+	}
+	err = conn.Send(resp)
+	if err != nil {
+		log.Debugf("error sending unsub_orderbook response: %v", err)
```
JoeGruffins, line 662: fee_rate response

JoeGruffins (approved):

client/core/trade.go
```diff
@@ -146,6 +146,12 @@ type trackedTrade struct {
 	notify        func(Notification)
 	epochLen      uint64
 	fromAssetID   uint32
+	// redeemFeeSuggestion is cached fee suggestion for redemption. We can't
+	// request a fee suggestion at redeem time because it would require making
+	// the full redemption routine async (TODO?). This fee suggestion is
+	// intentionally not stored as part of the db.OrderMetaData, and should
+	// be repopulated if the client is restarted.
+	redeemFeeSuggestion uint64
```
chappjc, line 154: 64 bit atomic load/store bug issue here.  Could probably be 32 bit, but perhaps best to put it as the first field in `trackedTrade`.

chappjc:
When you rebase, you are welcome to squash too.  A couple notes: context was removed from data api constructor (was unused), and `handleFeeRate` can switch to using `msg.Unmarshal` as per the changes in https://github.com/decred/dcrdex/pull/1014.  The one conflict with the stopBook closure can go back to one line too if you like.

client/orderbook/orderbook.go
```diff
@@ -24,6 +25,15 @@ type Order struct {
 	Epoch uint64
 }
 
+// copy creates a copy. Note that the OrderID is not a deep copy.
+func (o *Order) copy() *Order {
+	return &(*o)
```
chappjc, line 30: Since order.OrderID is an array, it would be a deep copy too.

client/orderbook/orderbook.go
```diff
@@ -196,48 +227,55 @@ func (ob *OrderBook) Reset(snapshot *msgjson.OrderBook) error {
 	ob.seq = snapshot.Seq
 	ob.seqMtx.Unlock()
 
-	ob.marketID = snapshot.MarketID
+	atomic.StoreUint64(&ob.feeRates.base, snapshot.BaseFeeRate)
+	atomic.StoreUint64(&ob.feeRates.quote, snapshot.QuoteFeeRate)
 
-	// Clear all orders, if any.
-	ob.orders = make(map[order.OrderID]*Order)
```
chappjc, line 202: Funny, we must have been (unintentionally?) avoiding a race with a higher level synchronization mechanism in the caller.

client/orderbook/orderbook.go
```diff
@@ -326,17 +364,24 @@ func (ob *OrderBook) updateRemaining(note *msgjson.UpdateRemainingNote, cached b
...
+	if found {
+		newOrder := ord.copy()
+		newOrder.Quantity = note.Remaining
+		ob.orders[oid] = newOrder
 	}
-
-	ord = ob.buys.UpdateRemaining(oid, note.Remaining)
-	if ord != nil {
-		return nil
+	ob.ordersMtx.Unlock()
```
chappjc, line 374: I'm a little unsure about the need to replace the order in the map with a fresh Order instance, but also the need to do anything with the `orders` map at all since the two booksides refer to the map entries (pointers).
Did you observe a bug or did this just seem like a needed change?

client/orderbook/orderbook.go
```diff
@@ -374,7 +421,9 @@ func (ob *OrderBook) unbook(note *msgjson.UnbookOrderNote, cached bool) error {
 	var oid order.OrderID
 	copy(oid[:], note.OrderID)
 
+	ob.ordersMtx.Lock()
 	order, ok := ob.orders[oid]
+	ob.ordersMtx.Unlock()
```
chappjc, line 426: Hmm, another surprising one.  OrderBook use patterns must have been preventing concurrent access to the map.

server/market/bookrouter.go
```diff
@@ -296,6 +313,11 @@ func (r *BookRouter) Run(ctx context.Context) {
 	wg.Wait()
 }
 
+// LastFeeRate is the last fee rate registered for an asset.
+func (r *BookRouter) LastFeeRate(assetID uint32) uint64 {
+	return atomic.LoadUint64(r.feeRateCache[assetID])
+}
```
chappjc, line 319: I had some thoughts about the roles of `BookRouter`, `Market`, and `OrderRouter` with respect to the fee rate fetching, caching, and retrieval that I shared in matrix.  The following is my high level understanding of how the pieces currently fit and what is responsible for what:

The assets implement `market.FeeFetcher`, which is a `FeeRate() (uint64, error)` method. These are provided to `Market`, who uses the fetcher at match time to get fee rates per asset for all the match sets.
These base/quote fee rates make their way the outgoing `BookRouter` via an `epoch_report` signal. The `BookRouter` stores these rates when it gets them.
`BookRouter` also implements `LastFeeRate(assetID uint32) uint64` (this commented line), and `OrderRouter` uses `BookRouter` as a `FeeSource interface` for this method. The `OrderRouter` uses this interface to ensure zero conf funded orders have high enough fee rate on funding txn.
So `BookRouter` with it's fee rate cache is the fee source for both itself and `OrderRouter`, while `Market` uses the assets directly for fee rates.

One question I had was what fee rates are known/cached before the first epoch_report? Also, how can we populate the book router's fee rate cache earlier, ideally construction time?

A general suggestion I had was to reduce and clarify the roles of these three `server/market` subsystems with respect to fee rate fetching/caching.  Presently the flow of fee rates is from RPC fetch in `Market` on epoch close -> fee rate cache in `BookRouter` via epoch_report -> retrieve from cache in `OrderRouter` via `BookRouter`'s `LastFeeRate` method.  This design makes `BookRouter` implement the cache, while `Market` actually does retrieval.  So I'm contemplating if a `FeeSource` can be some other type implemented outside of the market package, like something in server/dex that can do the RPC fetch and cache management without troubling the `server/market` types. Also, `server/dex.DEX` has the backends so it can hit `FeeRate()` to prepopulate the fee rate caches before even firing up the market subsystems.

Part of the appeal of a type that handles fee fetching and caching is consolidating these responsibilities that are now shared between the three market subsystems, and a fetch can automatically update the cache directly.  Both `OrderRouter` and `BookRouter` can get the same `FeeSource`, and calls from `Market` to this type's `FetchFeeRate(assetID uint32)` method (indirectly via wrapper for the relevant assetID) could update the cache.

client/orderbook/orderbook.go
buck54321: Oh, yeah. Nice.

client/orderbook/orderbook.go
chappjc: Curious that the race detector hasn't hit it before.  Maybe because when that code is called in bookie.go we've always been locking `booky.mtx`?

client/asset/interface.go
```diff
@@ -295,6 +302,10 @@ type Order struct {
 	// standing order, likely a market order or a limit order with immediate
 	// time-in-force.
 	Immediate bool
+	// FeeSuggestion is a suggested fee from the server. If a split transaction
+	// is used, the fee rate used should be at least the suggested fee, else
+	// zero-conf coins might be rejected.
+	FeeSuggestion uint64
```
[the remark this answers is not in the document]
buck54321: All of the fields refer to the same thing, which is just "the last known fee rate estimate". At a given time, the fee suggestion will be identical for all mentioned uses, so I think having a uniform name is reasonable. I can doc the intention of the individual fields, but there's also a case to be made for removing any specific recommendation from the `asset.Wallet` docs and just letting the concrete types use the suggestion however they need. ETH won't need split transactions, for instance, but may still use the fee suggestion in one of the `asset.Order` methods.

client/orderbook/orderbook.go
buck54321: I do believe that's the case, but not behavior I wanted to rely upon.

server/swap/swap.go
```diff
@@ -2028,8 +2024,8 @@ func readMatches(matchSets []*order.MatchSet, feeRates map[uint32]uint64) []*mat
 				takerSwapAsset = base
 			}
 
-			match.FeeRateBase = feeRates[base]
-			match.FeeRateQuote = feeRates[quote]
+			match.FeeRateBase = matchSet.FeeRateBase
+			match.FeeRateQuote = matchSet.FeeRateQuote
```
chappjc, line 2028: I think `matchSet.Matches()` -> `dex/order.newMatch` has this job now, so redundant I believe.

client/core/trade.go
```diff
@@ -1594,10 +1605,18 @@ func (c *Core) redeemMatchGroup(t *trackedTrade, matches []*matchTracker, errs *
 		})
 	}
 
+	// Don't use (*Core).feeSuggestion here, since can incur an RPC request.
+	// If we don't have a synced book, use t.redemption
+	feeSuggestion := t.dc.bestBookFeeSuggestion(t.wallets.toAsset.ID)
+	if feeSuggestion == 0 {
+		feeSuggestion = t.redeemFeeSuggestion
```
chappjc, line 1612: looks like this should be an atomic load

server/dex/feemgr.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"strconv"
+	"sync/atomic"
+
+	"decred.org/dcrdex/server/asset"
+	"decred.org/dcrdex/server/market"
+)
+
+// FeeManager manages fee fetchers and a fee cache.
+type FeeManager struct {
```
chappjc, line 12: If you're readying another commit, I'd suggest the following assertions for both documentation and anti-fragility, otherwise not critical:

```go
var _ market.FeeSource = (*FeeManager)(nil)
var _ market.FeeFetcher = (*feeFetcher)(nil)
```

client/orderbook/orderbook.go
chappjc: Comment still says `not a deep copy`

chappjc (approved):
Looks good and tests well.  I had a handful of other questions and suggestions, but I'm happy to merge this regardless.

Getting client to hit the fee_rate route was a bit tricky too, needing to use `dexcctl trade` with the browser closed and unsubscribed from the book.  Does that seem right?

I'm also still a bit confused by the `OrderBook` stuff and I'm wondering if this is gonna resolve https://github.com/decred/dcrdex/issues/835

server/dex/feemgr.go
```diff
@@ -0,0 +1,81 @@
...
+		BackedAsset: asset,
+		lastRate:    lastRate,
+	}
+}
+
+// FeeRate fetches a new fee rate and updates the cache.
+func (f *feeFetcher) FeeRate() uint64 {
+	r, err := f.Backend.FeeRate()
+	if err != nil {
+		log.Errorf("Error retrieving fee rate for %s: %v", f.Symbol, err)
```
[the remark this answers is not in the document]
buck54321: There was a little chat in matrix about this. If the backend returns an error, we would want to pass that on to the client as a zero, indicating no estimate is available. Returning any other value would require the server to reason about either the validity of outdated estimates or the meaning of particular errors, which is undesireable, imo.

client/orderbook/orderbook.go
buck54321: crap

chappjc: That's alright.  I realized this wasn't making a copy of anything actually.  Go was simplifying it to just returning the same pointer: https://github.com/decred/dcrdex/commit/a71b9d6724af4a33c266beba7a5f6810fcf14618
Working on some tweaks for this.

PR #1002, opened 2021-03-05
update linter, golang/term repo, and auth test fixes

chappjc:
Locally updating the golangci-lint to 1.38 revealed some previously unnoticed issues:

- In server/auth/auth_test.go we were assigning a `*msgjson.Error` (concrete type that happened to implement the `error` interface) to an existing `error` (an interface), so even if the pointer were `nil`, the error interface was a non-nil interface wrapping a nil `*msgjson.Error`, and the condition was impossible. See https://staticcheck.io/docs/checks#SA4023.  The tests were also incorrect because the modified payloads were not being remarshalled into a request.
- Several of the server/auth/auth_test.go tests were relying on side effects from earlier tests, namely a signature set for the user, causing a panic if these tests were run on their own.
- The `golang.org/x/crypto/ssh/terminal` package was deprecated and is now in `github.com/golang/term`.
- "disabled" dex accounts appear to lose their account proof (fee payment records) when they are moved to the disabled accounts bucket. NOT FIXED, just commented with a TODO.

This also updates golangci-lint to 1.38, which helped find the above issues.

This includes some other random linting, and cleanup to a lot of slop in the `AccountImport` and `AccountExport` methods of `client.Core`.

I've separated the term, auth, and linting commits, but they all stemmed from the linter update so they are in the same PR.

Side note: while the set of enabled linters is not change, I enable the following *additional* linters locally with golangci-lint: unused, golint, ifshort, revive, predeclared, makezero, thelper, misspell. I often ignore them but usually they are good to follow.


dex/order/order.go
```diff
@@ -552,12 +552,6 @@ func (o *MarketOrder) String() string {
 
 // serializeSize returns the length of the serialized MarketOrder.
 func (o *MarketOrder) serializeSize() int {
-	// Compute the size of the serialized Coin IDs.
-	var coinSz int
-	for _, coinID := range o.Coins {
-		coinSz += len(coinID)
-		// TODO: ensure all Coin IDs have the same size, indicating the same asset?
-	}
```
chappjc, line 560: Copy pasta from `(*Trade).serializeSize()`

server/admin/prompt.go
```diff
@@ -10,7 +10,7 @@ import (
 	"os"
 
 	"decred.org/dcrdex/dex/encode"
-	"golang.org/x/crypto/ssh/terminal"
```
chappjc, line 13: deprecated in https://github.com/golang/crypto/commit/4be66e5b658251a93e17d931a68d9c0ecba9f83a

server/auth/auth_test.go
```diff
@@ -931,6 +931,7 @@ func TestAccountErrors(t *testing.T) {
 
 func TestRoute(t *testing.T) {
 	user := tNewUser(t)
+	rig.signer.sig = user.randomSignature()
 	connectUser(t, user)
```
chappjc, line 935: All these panicked in `connectUser` when given a nil sig.  Each one was relying on it being set in a prior test.

server/auth/auth_test.go
```diff
@@ -1560,15 +1567,17 @@ func TestMatchStatus(t *testing.T) {
 	}
 
 	reqPayload[0].MatchID = []byte{}
-	err = rig.mgr.handleMatchStatus(user.conn, req)
-	if err == nil {
```
chappjc, line 1564: This was essentially `var err error = *msgjson.Error`, so `err` could never be nil regardless of the pointer value.

chappjc: Oh, I see you resolved this independently in the txData pr @buck54321.  I'm still changing the return variable to `msgErr` to highlight that the return type is a concrete type (`*msgjson.Error`) and not just an `error`.

PR #1004, opened 2021-03-06
multi: add blockcache to client/asset/dcr

itswisdomagain:
The blockcache currently only caches basic block information but will ultimately include transactions and block filters.

### DIFF
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -412,7 +412,7 @@ type ExchangeWallet struct {
 
 	// The blockCache stores just enough info about blocks to prevent repeated
 	// calls to GetBlockVerbose for the same block.
-	blockCache *blockCache
+	blockCache *dexdcr.BlockCache
 
 	// Coins returned by Fund are cached for quick reference.
 	fundingMtx   sync.RWMutex
@@ -531,7 +531,7 @@ func unconnectedWallet(cfg *asset.WalletConfig, dcrCfg *Config, chainParams *cha
 		feeRateLimit:        feesLimitPerByte,
 		redeemConfTarget:    redeemConfTarget,
 		useSplitTx:          dcrCfg.UseSplitTx,
-		blockCache:          newBlockCache(logger),
+		blockCache:          dexdcr.NewBlockCache(logger),
 	}, nil
 }
 
@@ -1744,7 +1744,7 @@ func (dcr *ExchangeWallet) FindRedemption(ctx context.Context, coinID dex.Bytes)
 	if contractBlock == nil {
 		dcr.findRedemptionsInMempool([]outPoint{contractOutpoint})
 	} else {
-		bestHash, bestHeight := dcr.blockCache.tip()
+		bestHash, bestHeight := dcr.blockCache.Tip()
 		bestBlock := &block{bestHeight, bestHash}
 		dcr.findRedemptionsInBlockRange(contractBlock, bestBlock, []outPoint{contractOutpoint})
 	}
@@ -2776,7 +2776,7 @@ func (dcr *ExchangeWallet) checkForNewBlocks() {
 		return
 	}
 
-	cachedBestHash, cachedBestHeight := dcr.blockCache.tip()
+	cachedBestHash, cachedBestHeight := dcr.blockCache.Tip()
 	if newTip.hash.IsEqual(cachedBestHash) {
 		return
 	}
@@ -2825,7 +2825,7 @@ func (dcr *ExchangeWallet) checkForNewBlocks() {
 				return fmt.Errorf("getblockhash error for reorg height %d: %w", reorgHeight, translateRPCCancelErr(err))
 			}
 			dcr.log.Debugf("Tip change (reorg) from %s (%d) => %d (%s)", reorgHash, reorgHeight, newTip.height, newTip.hash)
-			dcr.blockCache.purgeMainchainBlocks(reorgHeight)
+			dcr.blockCache.PurgeMainchainBlocks(reorgHeight)
 			redemptionSearchStartBlock = &block{reorgHeight, reorgHash}
 			return nil
 		}
@@ -2849,7 +2849,7 @@ func (dcr *ExchangeWallet) checkForNewBlocks() {
 	}
 
 	// Add the new best block to the cache. Also updates the cached tip.
-	_, err = dcr.blockCache.add(newBestBlock)
+	_, err = dcr.blockCache.Add(newBestBlock)
 	if err != nil {
 		dcr.log.Errorf("error adding new best block to cache: %v", err)
 	}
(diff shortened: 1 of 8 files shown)


itswisdomagain:
The first commit adds a blockcache to `client/asset/dcr`.
Second commit combines the server and asset blockcaches into one at `dex/networks/dcr/BlockCache`. The merge looks good to me but I can take off the second commit if it makes things less clear.

The client asset blockcache isn't really needed yet but will become handy when using cfilters to check confirmations in #788. Can also come in handy for #954 which uses cfilters for contract redemption search. I'm okay with adding this as a commit to either of those PRs but thought to put it up separately for easier review.

chappjc:
I need to review, but please have a look and consider how a flaw in the current blockCache usage can cause an infinite recursion and panic; https://github.com/decred/dcrdex/pull/1016/checks?check_run_id=2108248511#step:5:137
I'd really like to engineer out such recursive calls if possible.  And in general this is one of the reasons I am wary of building cache unless they are needed - they are difficult to implement and use correctly.

chappjc:
Will hit this as soon as https://github.com/decred/dcrdex/pull/953 goes in

client/asset/dcr/dcr.go
```diff
@@ -1640,9 +1666,8 @@ func (dcr *ExchangeWallet) FindRedemption(ctx context.Context, coinID dex.Bytes)
 	if contractBlock == nil {
 		dcr.findRedemptionsInMempool([]outPoint{contractOutpoint})
 	} else {
-		dcr.tipMtx.RLock()
-		bestBlock := dcr.currentTip
-		dcr.tipMtx.RUnlock()
+		bestHash, bestHeight := dcr.blockCache.Tip()
```
chappjc, line 1669: Looking to get this PR in now that the 0.3 cycle is well underway.  When it's rebased, the new method `calcPastMedianTime()` needs to use `bestHash, _ := dcr.blockCache.Tip()` (and a change to the test) but other than that, it looks good.
Thanks @itswisdomagain!

server/asset/dcr/utxo.go
```diff
@@ -65,25 +65,25 @@ func (txio *TXIO) confirmations(ctx context.Context, checkApproval bool) (int64,
...
 		}
 		if mainchainBlock != nil && checkApproval {
 			nextBlock, err := txio.dcr.getMainchainDcrBlock(ctx, txio.height+1)
 			if err != nil {
 				return -1, fmt.Errorf("error retrieving approving block tx %s: %w", txio.tx.hash, err)
 			}
-			if nextBlock != nil && !nextBlock.vote {
-				return -1, fmt.Errorf("transaction %s block %s has been voted as invalid", txio.tx.hash, nextBlock.hash)
+			if nextBlock != nil && !nextBlock.Vote {
+				return -1, fmt.Errorf("transaction %s block %s has been voted as invalid", txio.tx.hash, nextBlock.Hash)
```
chappjc, line 86: Hadn't considered this recently, but what is the right thing to do if a txn was in a stake-invalidated block?  I think it would be to reset `txio.height` to 0 and try again (making `confirmations` do getrawtransaction) to see if it is now back in mempool or simply in a different block with fewer confirmations.
Right now the various `Confirmations` methods only do the above when the error is `ErrReorgDetected`, but it seems like the proper resolution for stake invalidation too.  Well, actually they makes a new `TXIO` instead of resetting the `height` field, but you know what I mean.
BTW, did you gain any insight about how the infinite recursion happened in that one issue?

dex/networks/dcr/blockcache.go
```diff
@@ -0,0 +1,146 @@
...
+// rpcclient.GetblockVerbose.
+type BlockCache struct {
+	mtx       sync.RWMutex
+	blocks    map[chainhash.Hash]*Block
+	mainchain map[int64]*chainhash.Hash
+	bestBlock struct {
+		hash   *chainhash.Hash
+		height int64
+	}
+	log dex.Logger
```
chappjc, line 33: `log` appears unused. Remove?

dex/networks/dcr/blockcache.go
```diff
@@ -0,0 +1,146 @@
...
+	"decred.org/dcrdex/dex"
+	"github.com/decred/dcrd/chaincfg/chainhash"
+	chainjson "github.com/decred/dcrd/rpc/jsonrpc/types/v2"
+)
+
+// Block defines basic information about a block.
+type Block struct {
+	Hash         *chainhash.Hash
+	Height       int64
+	PreviousHash string
```
chappjc, line 19: `PreviousHash` appears unused.  Plans for it for your SPV work?

dex/networks/dcr/blockcache.go
```diff
@@ -0,0 +1,146 @@
...
+
+	blk := &Block{
+		Hash:         hash,
+		Height:       block.Height,
+		PreviousHash: block.PreviousHash,
+		Vote:         block.VoteBits&1 != 0,
+	}
+	cache.blocks[*hash] = blk
+
+	// Orphaned blocks will have -1 confirmations. Don't add them to mainchain.
```
chappjc, line 64: Will have **-1** confirmations, or negative of the distance from mainchain?  The logic in `checkForNewBlocks` requires distance from mainchain. https://github.com/decred/dcrdex/pull/1004/files#diff-cd95defa7320b76287f90152b40ee5ec5edec59ca9af2eb9579ecda539ece20aR2688

dex/networks/dcr/blockcache.go
```diff
@@ -0,0 +1,146 @@
...
+		return nil, fmt.Errorf("error decoding block hash %s: %w", block.Hash, err)
+	}
+
+	blk := &Block{
+		Hash:         hash,
+		Height:       block.Height,
+		PreviousHash: block.PreviousHash,
+		Vote:         block.VoteBits&1 != 0,
+	}
+	cache.blocks[*hash] = blk
```
chappjc, line 62: Given there's no obvious way to know when a certain block should be evicted, even for the caller, we may need to consider a capacity and some kind of eviction priority based on height.  But I don't think we have to do this right away though since these are small structures. A TODO comment would be fine.

chappjc (changes requested):

client/asset/dcr/dcr.go
```diff
@@ -2744,97 +2768,139 @@ func (dcr *ExchangeWallet) monitorBlocks(ctx context.Context) {
...
+	// successor of the currently cached best block. This MUST be done before
+	// adding the new best block to the cache and notifying consumers of the tip
+	// change.
+	var redemptionSearchStartBlock *block
+	err = func() error {
+		// If the new best block builds on the cached best block or the cache is empty,
+		// it's good to add.
+		if cachedBestHash == nil {
+			return nil // cache is empty
+		}
```
JoeGruffins, line 2798: In this case, `dcr.findRedemptionsInBlockRange` will start from the genesis block?

JoeGruffins (commented):
Does the dcr `reorg` script work? Would be cool to use to test this...

chappjc:
You mentioned in matrix you wanted to hold off on the block cache for now?  If that's correct, can we close this and reopen later if/when needed?

itswisdomagain:
Yes, yes. Closing for now.

PR #1007, opened 2021-03-08
UI: update wallet lock state on order placement.

### LINKED ISSUE #969: UI: wallet lock state not updated on automatic unlock
Lately I have noticed that the wallet lock icon does not update to reflect unlocked status when placing an order that automatically unlocks a locked wallet.  A refresh is needed to update the lock icons now.

amassarwi:
This re-adds `await app.fetchUser()` to fetch updated wallet lock state on placing an order.
This call has been deleted in https://github.com/decred/dcrdex/commit/8cc1502b0bb511513c0cf70117cbc29b0952f8db#diff-177f3e40cc5694b2f1822d102b13d0e71cef86831ce3522ab379c68784fea9d0L1187 as we switched to update the orders list in https://github.com/decred/dcrdex/commit/8cc1502b0bb511513c0cf70117cbc29b0952f8db#diff-b2831f969918f14b3aac94dc85fb036c6ca85613696187f7caaf83133ff4ec4bR481-R486.

Fixes #969.


buck54321:
I'm trying to get rid of all of the `await app.fetchUser()` calls in favor of the notification system. The question that needs to be answered is -- why is the wallet being unlocked without emitting a `WalletStateNote`, or if `Core` is emitting a `WalletStateNote`, why isn't the UI being updated?

chappjc:
Please make sure you can reproduce https://github.com/decred/dcrdex/issues/969 as well.  It could be stale.

amassarwi:
@buck54321 good question
@chappjc yup was able to reproduce

amassarwi:
https://github.com/decred/dcrdex/blob/master/client/core/core.go#L1605
https://github.com/decred/dcrdex/blob/master/client/core/core.go#L1583
Looks like `Core` is indeed emitting the notifications.
So I guess these aren't handled in gui side yet?

buck54321:
The `MarketsPage` is only subscribed to 5 notifications

```
this.notifiers = {
  order: note => { this.handleOrderNote(note) },
  epoch: note => { this.handleEpochNote(note) },
  conn: note => { this.marketList.setConnectionStatus(note) },
  balance: note => { this.handleBalanceNote(note) },
  feepayment: note => { this.handleFeePayment(note) }
}
```

Looks like you'll need to subscribe to the `walletstate` and `walletconfig` notifications, and add a handler that calls `MarketsPage.balanceWgt.updateAsset`.

amassarwi:
@buck54321 Thanks for the hint, adding a `walletstate` notification handler to MarketsPage 
fixed the issue for good.

JoeGruffins (approved):
dcr lock icon is now updating for me.

client/core/core.go
```diff
@@ -3153,10 +3153,13 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
 	if err != nil {
 		return nil, 0, err
 	}
+	c.notify(newWalletStateNote(fromWallet.state()))
+
 	err = prepareWallet(toWallet)
 	if err != nil {
 		return nil, 0, err
 	}
+	c.notify(newWalletStateNote(toWallet.state()))
```
chappjc, line 3162: These address the issue, but what if we did this notify from `connectAndUnlock`, and then only if there is a wallet state change (i.e. it does either connect or unlock)? We could forgo wallet state notes when they convey no new info.

amassarwi: Sounds good, will give it a try thanks.

amassarwi: Done 👍🏽

chappjc (approved):

PR #1008, opened 2021-03-09
[0.1] prevent preimage request from blocking if order submission fails

chappjc:
If an order request is sent but the connection drops before the response is received, this results in an order on the server, but the client receives no response with the order ID.  In this state, the order must be abandoned, but later a preimage request would still be sent by the server to the client on epoch close.  In this situation, the client will not recognize the order ID, and `handlePreimageRequest` will block for up to a minute waiting for a signal (from `syncOrderPlaced`) that will never come, resulting in the error:

```
timed out syncing preimage request for %s, order %s"
```

The problem is not the error per se, but the fact that it blocks all incoming request handling for up to a minute because of this.  All request handlers should not have the potential to block for significant periods of time. If there are N orders with the same failure, they will all backup for N minutes.  This is why we made init/redeem requests asynchronous.

This PR handles this relatively rare case of when a preimage request arrives before the order submission response is received (or if it will never be received) by going asynchronous and waiting in a goroutine so that other incoming requests and notifications may be processed.

PR https://github.com/decred/dcrdex/pull/1009 based on master expands on this by adding the order commitment to the preimage request.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -97,7 +97,7 @@ type dexConnection struct {
 // successfully sent.
 const (
 	DefaultResponseTimeout = comms.DefaultResponseTimeout
-	fundingTxWait          = 2 * time.Minute
+	fundingTxWait          = time.Minute // TODO: share var with server/market or put in config
 )
 
 // running returns the status of the provided market.
@@ -4289,6 +4289,7 @@ func handlePreimageRequest(c *Core, dc *dexConnection, msg *msgjson.Message) err
 		// and we can go ahead.
 		delete(c.piSyncers, oid)
 		c.piSyncMtx.Unlock()
+		dc.log.Debugf("Received preimage request for known order %v", oid)
 		return processPreimageRequest(c, dc, msg.ID, oid)
 	}
 
@@ -4302,14 +4303,14 @@ func handlePreimageRequest(c *Core, dc *dexConnection, msg *msgjson.Message) err
 	// could be a bogus preimage request, so we do not want to block the caller,
 	// (*Core).listen if this hangs. Preimage requests are ok to handle
 	// asynchronously since there can be no matches until we respond to this.
-	c.log.Warnf("Got preimage request for %v before we got an order submission response!", oid)
+	c.log.Warnf("Received preimage request for %v before we got an order submission response!", oid)
 	go func() {
 		select {
 		case <-syncChan:
 			if err := processPreimageRequest(c, dc, msg.ID, oid); err != nil {
-				c.log.Errorf("processPreimageRequest for %v failed: %v", oid, err)
+				c.log.Errorf("async processPreimageRequest for %v failed: %v", oid, err)
 			} else {
-				c.log.Debugf("processPreimageRequest for %v succeeded", oid)
+				c.log.Debugf("async processPreimageRequest for %v succeeded", oid)
 			}
 			// The channel is deleted from the piSyncers map by syncOrderPlaced.
 			return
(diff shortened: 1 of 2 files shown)


client/core/core.go
```diff
@@ -4260,64 +4266,71 @@ out:
...
+	// could be a bogus preimage request, so we do not want to block the caller,
+	// (*Core).listen if this hangs. Preimage requests are ok to handle
+	// asynchronously since there can be no matches until we respond to this.
+	c.log.Warnf("Got preimage request for %v before we got an order submission response!", oid)
+	go func() {
 		select {
 		case <-syncChan:
-		case <-time.After(time.Minute):
-			deletePiSyncer()
-			return fmt.Errorf("timed out syncing preimage request for %s, order %s", dc.acct.host, oid)
```
chappjc, line 4300: This was the trouble.  `prepareTrackedTrade` and `tryCancel` could return before getting to `syncOrderPlaced`, but with server having actually received the order and thus sent this preimage request that client cannot handle.

client/core/core.go
```diff
@@ -4260,64 +4266,71 @@ out:
...
-		// Hack around the race in prepareTrackedTrade from submitting the
-		// order, receiving the order result, and registering the tracked trade.
-		// It's possible that the dc.trades map won't have the oid yet after
-		// submitting the order, so if (1) we didn't find this oid, AND (2)
-		// we're in that window, that order result is about to be stored, so
-		// wait for it and check again.
-		tracker, preImg, isCancel = dc.findOrder(oid)
-		if tracker == nil {
-			return fmt.Errorf("no active order found for preimage request for %s", oid)
-		}
```
chappjc, line 4318: Let over code from before `piSyncers` was added.

client/core/core.go
```diff
@@ -2751,23 +2756,23 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
...
-	// TODO: Notify the client somehow.
 	logAbandon := func(err interface{}) {
-		c.log.Errorf("Abandoning order. "+
-			"preimage: %x, server time: %d: %v",
+		c.log.Errorf("Abandoning order. preimage: %x, server time: %d: %v",
 			preImg[:], result.ServerTime, err)
 	}
 
-	err = validateOrderResponse(dc, result, ord, msgOrder)
+	err = validateOrderResponse(dc, result, ord, msgOrder) // stamps the order, giving it a valid ID
```
chappjc, line 2772: You can manually test the async handling by sleeping for a few seconds here, allowing server to send a preimage request if the epoch closes while sleeping but before we put it in our trades map.  Works as expected. e.g.:

> (order placed and sleeping)
2021-03-09 08:47:45.001 [WRN] CORE: Received preimage request for d20502da753eb5912ea984efa69a2c3b5f8ec9d8e7f89491eac772a24900684b before we got an order submission response!
(still sleeping with nothing blocking, then awake and signalling on syncChan)
> 2021-03-09 08:47:46.911 [DBG] CORE: notify: |POKE| (order) Order placed - selling 10.00000000 dcr, rate = 0.10000000 (d20502da) - Order: d20502da753eb5912ea984efa69a2c3b5f8ec9d8e7f89491eac772a24900684b
> 2021-03-09 08:47:46.911 [TRC] CORE: notify: |DATA| (order) preimage sent - Order: d20502da753eb5912ea984efa69a2c3b5f8ec9d8e7f89491eac772a24900684b
> 2021-03-09 08:47:46.911 [DBG] CORE: async processPreimageRequest for d20502da753eb5912ea984efa69a2c3b5f8ec9d8e7f89491eac772a24900684b succeeded
> 2021-03-09 08:47:46.912 [INF] CORE: Standing order d20502da did not match and is now booked.

JoeGruffins (approved):

chappjc:
Closing since this is included in https://github.com/decred/dcrdex/pull/1009 for 0.2, and it no longer appears urgent for 0.2, which BTW is also frozen.

PR #1009, opened 2021-03-09
include order commitment in preimage request

chappjc:
This builds on https://github.com/decred/dcrdex/pull/1008 by including the `Commitment` in the `PreimageRequest`, which solves multiple issues with the current race between order submission response processing and incoming preimage requests.  Namely, we can record the commitment *before* submitting the order/cancel request, thus avoiding a race necessitating the `piSyncers` map hoop jumping.  More importantly however, we can reject outright preimage requests containing unrecognized commitments.  Without this, we had to blindly wait for an arbitrary timeout (was 1 min) to see if we got an order submission response back with the order ID of interest.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -322,7 +322,11 @@ func (c *Core) tryCancelTrade(dc *dexConnection, tracker *trackedTrade) error {
 	var result = new(msgjson.OrderResult)
 	err = dc.signAndRequest(msgOrder, route, result, DefaultResponseTimeout)
 	if err != nil {
-		return err
+		// At this point there is a possibility that the server got the request
+		// and created the cancel order, but we lost the connection before
+		// receiving the response with the cancel's order ID. Any preimage
+		// request will be unrecognized. This order is ABANDONED.
+		return fmt.Errorf("failed to submit cancel order targeting trade %v: %w", oid, err)
 	}
 	err = validateOrderResponse(dc, result, co, msgOrder)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


client/core/core.go
```diff
@@ -2993,6 +2997,12 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
 		return nil, 0, fmt.Errorf("wallet %v failed to sign coins: %w", wallets.fromAsset.ID, err)
 	}
 
+	commitSig := make(chan struct{})
+	defer close(commitSig)
```
chappjc, line 3001: Whether `signAndRequest` succeeds or fails, which was the trouble before with `piSyncers`, which was keyed on order ID, a value that was only *available* after a *successful* request/response, two problematic conditions.

dex/msgjson/types.go
```diff
@@ -841,6 +841,7 @@ type TradeResumption struct {
 // PreimageRequest is the server-originating preimage request payload.
 type PreimageRequest struct {
 	OrderID        Bytes `json:"orderid"`
+	Commitment     Bytes `json:"commit"`
 	CommitChecksum Bytes `json:"csum"`
```
chappjc, line 845: Not entirely sure why CommitChecksum is in this payload.

chappjc: Oh derp it's because server should not be asking for preimages until the epoch is closed.  So client should be remembering this `CommitChecksum` and verifying that this orders doesn't end up in an epoch's match proof with a different `csum`, which would indicate server is cheating (e.g. we handed over our preimage so the server could game the shuffle).

chappjc:
@JoeGruffins I closed https://github.com/decred/dcrdex/pull/1008, which you reviewed, but that's the first commit (https://github.com/decred/dcrdex/pull/1009/commits/461bde1644a4d1174192c4be04d15ef5c90a4cae) of this PR, so no time wasted reviewing it.
This PR takes a bigger conceptual leap to solve this preimage request-by oid issue

JoeGruffins (approved):

buck54321 (approved):
Can't wait til the 1.0 purge

PR #1010, opened 2021-03-09
client/assets/eth: in-process geth demo

buck54321:
Intended as a complement to #1005, hoping to be cannibalized. This snippet demonstrates running geth natively without HTTP RPCs.


client/asset/eth/eth.go
```diff
@@ -0,0 +1,291 @@
...
+func NewETH(appDir string, goerli bool) (*ETH, error) {
+	stackConf := &node.Config{DataDir: appDir}
+
+	stackConf.P2P.MaxPeers = 10
+	var key *ecdsa.PrivateKey
+	var err error
+	if key, err = crypto.GenerateKey(); err != nil {
+		return nil, err
+	}
+	stackConf.P2P.PrivateKey = key
```
buck54321, line 45: May or may not have to do this. Needs more testing.

client/asset/eth/eth.go
```diff
@@ -0,0 +1,306 @@
...
+	ethCfg := ethconfig.Defaults
+	if goerli {
+		ethCfg.Genesis = core.DefaultGoerliGenesisBlock()
+		ethCfg.NetworkId = params.GoerliChainConfig.ChainID.Uint64()
+	}
+
+	// ethCfg.Genesis = core.DefaultGenesisBlock() // eth will do this anyway when Genesis == nil is encountered, and the messaging it better that way.
+	// ethCfg.SyncMode = downloader.SnapSync // Supposed to be enabled with Berlin on April 14th 2021
+	// ethCfg.SyncMode = downloader.FastSync  // The default
+	ethCfg.SyncMode = downloader.LightSync
```
buck54321, line 92: Updated here after @JoeGruffins reminded me that fast != light. Much faster now. I'm leaning towards think only enabling light mode to start, and then add snap mode option after Berlin.

chappjc: I got the impression that snap sync was going to be the *default* after Berlin, but it could actually work now if other nodes are making snapshots.  But that's aside because light is the way to go for client.

buck54321:
Incorporated into #1005. Closing.

PR #1011, opened 2021-03-09
server/market: suspend with purge needs to clear BookRouter

chappjc:
Although market suspend with persist=false was purging the book both in
Market's in-memory book and in the DB, the outgoing BookRouter's msgBook
was not resetting its orders map.  You can observe the bug without this change
by doing a suspend?persist=false, notice the book and chart clearing, but then
refreshing or restarting dexc brings the book back even though the market
really has no booked orders.

This change reinitializes the msgBook.orders map on purge.

This also moves a bit of repetative code out of the main runBook loop.

### DIFF
--- a/server/market/bookrouter.go
+++ b/server/market/bookrouter.go
@@ -311,11 +311,12 @@ func (r *BookRouter) runBook(ctx context.Context, book *msgBook) {
 		log.Infof("Book router terminating for market %q", book.name)
 	}()
 
+	book.mtx.Lock()
+	book.running = true
+	book.mtx.Unlock()
+
 out:
 	for {
-		book.mtx.Lock()
-		book.running = true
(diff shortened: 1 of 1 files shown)


chappjc:
To reproduce and test:

1. start up server and a client
2. book orders
3. dexadm suspend&persist=false (e.g. `curl --cacert ~/dextest/dcrdex/rpc.cert --basic -u user:adminpass https://127.0.0.1:16542/api/market/dcr_btc/suspend?persist=false`), then wait for it to suspend and purge at the end of the epoch
4. orders will disappear from the markets browser page
5. F5 the browser page, and it will show the old booked orders because the book router still has them even though they are gone from the actual Market's book

buck54321 (approved):

PR #1012, opened 2021-03-10
client/webserver: allow multiple authorized sessions

chappjc:
This allow multiple authenticated client http sessions by tracking multiple cookies in the WebServer. You can now login in different browsers without one login invalidating the other's auth cookie.

Logout ends all authenticated sessions however.  This is done because the logout route also performs `(*Core).Logout`, crippling other authenticated http sessions.  Also, if we just removed the auth token for the one session performing the logout, the backend would really have no way to keep the map clean and it could conceivable keep filling up if requests kept reauthing and not logging out.

### DIFF
--- a/client/webserver/api.go
+++ b/client/webserver/api.go
@@ -328,6 +328,10 @@ func (s *WebServer) apiLogout(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


JoeGruffins (approved):
Works well.

buck54321 (approved):

PR #1014, opened 2021-03-12
multi: graceful support for json unmarshal of null

chappjc:
A number of `json.Unmarshal` calls now properly handle a `Payload`
of the JSON `null` value, which depending on the caller's use of
`json.Unmarshal` can set a pointer of the type to `nil`.

The docs for `(*Message).Unmarshal` now call out the `null` `Payload`
handling explicitly. A `null` `Payload` is not an error however, since
the caller may be happy to accept (or even expect) a `null` `Payload`.
Further, for the caller to even allow their input to be set to `nil`, they must
call `Unmarshal` with sufficient indirection. For example,

```go
t := new(T) // *T
err := json.Unmarshal(msg.Payload, &t) // passed **T, can set t to nil
```

Only with the above approach can `json.Unmarshal` or `(*Message).Unmarshal`
set `t` to `nil`. Callers can either pass a `*T` instead and accept a
zero-value `T` in the case of a `null` `Payload`, or pass a `**T` and check
`if err != nil || t == nil` when they need to identify this case.

For example, the client's RPC server's `handleRequest` method permits,
and in the case of the version RPC expects, a `null` `Payload` without
error and as such unmarshals with:

```go
params := new(RawParams)
err := req.Unmarshal(params) // NOT &params - cannot be set to nil for []byte("null") Payload
```

For the client RPC server, this is the safest approach, and a `nil`
error is _required_. Similarly, `dex/ws.(*WSLink).inhandler` expects no
error for a `null` `Payload` and recognizes it with a specific check:

```go
msg := new(msgjson.Message)
err = json.Unmarshal(msgBytes, msg) // *Message
...
if msg.ID == 0 { // also covers msgBytes []byte("null")
    c.SendError(1, msgjson.NewError(msgjson.RPCParseError, "request id cannot be zero"))
```

The `(*Message).Response` method unmarshals with a `**T`, and is updated
with this `nil` check, and returns an error:

```go
fmt.Errorf("null response payload").
```

The docs for `msgjson.DecodeMessage` are updated to reflect that the
both returned `*Message` and error can be `nil` in this `null` case.
`DecodeMessage` is only used in tests presently.

Each of the following route handlers require detecting a `null` payload,
and are thus updated with post `Unmarshal` checks: `handleConnect`,
`handleRegister`, `handleNotifyFee`, `handleInit`, and `handleRedeem`.
All routes were called from a helper with a deferred recover.

This also adds the method `client/comms.(*wsConn).SendRaw([]byte)` for
testing purposes. It is not part of the `WsConn interface` however.

NOTE: This is [well-documented behavior of `json.Unmarshal`](https://golang.org/pkg/encoding/json/#Unmarshal), but here's an illustration: https://play.golang.org/p/YbLpU9IQUtn

### DIFF
--- a/server/auth/auth.go
+++ b/server/auth/auth.go
@@ -6,7 +6,6 @@ package auth
 import (
 	"context"
 	"encoding/hex"
-	"encoding/json"
 	"fmt"
 	"math"
 	"sync"
@@ -1033,8 +1032,8 @@ func (auth *AuthManager) loadUserScore(user account.AccountID) (int32, error) {
 // a response is issued, and a clientInfo is created or updated.
 func (auth *AuthManager) handleConnect(conn comms.Link, msg *msgjson.Message) *msgjson.Error {
 	connect := new(msgjson.Connect)
-	err := json.Unmarshal(msg.Payload, &connect)
-	if err != nil {
+	err := msg.Unmarshal(&connect)
+	if err != nil || connect == nil {
 		return &msgjson.Error{
 			Code:    msgjson.RPCParseError,
 			Message: "error parsing connect request",
@@ -1343,10 +1342,12 @@ func (auth *AuthManager) handleMatchStatus(conn comms.Link, msg *msgjson.Message
 			"cannot use route 'match_status' on an unauthorized connection")
 	}
 	var matchReqs []msgjson.MatchRequest
-	err := json.Unmarshal(msg.Payload, &matchReqs)
-	if err != nil {
+	err := msg.Unmarshal(&matchReqs)
+	if err != nil || matchReqs == nil /* null Payload */ {
 		return msgjson.NewError(msgjson.RPCParseError, "error parsing match_status request")
 	}
+	// NOTE: If len(matchReqs)==0 but not nil, Payload was `[]`, demanding a
+	// positive response with `[]` in ResponsePayload.Result.
 
 	mkts := make(map[string]*marketMatches)
 	var count int
@@ -1374,7 +1375,7 @@ func (auth *AuthManager) handleMatchStatus(conn comms.Link, msg *msgjson.Message
 		}
 	}
 
-	results := make([]*msgjson.MatchStatusResult, 0, count)
+	results := make([]*msgjson.MatchStatusResult, 0, count) // should be non-nil even for count==0
 	for _, mm := range mkts {
 		statuses, err := auth.storage.MatchStatuses(client.acct.ID, mm.base, mm.quote, mm.idList())
 		// no results is not an error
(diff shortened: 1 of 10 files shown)


server/auth/auth.go
```diff
@@ -1039,7 +1039,7 @@ func (auth *AuthManager) loadUserScore(user account.AccountID) (int32, error) {
 func (auth *AuthManager) handleConnect(conn comms.Link, msg *msgjson.Message) *msgjson.Error {
 	connect := new(msgjson.Connect)
 	err := json.Unmarshal(msg.Payload, &connect)
```
buck54321, line 1041: Can you switch these to `msg.Unmarshal` while you're here? These were written before the method existed, I think.

buck54321: That'll also point them towards the new documentation.

chappjc: Done. https://github.com/decred/dcrdex/pull/1014/commits/bd69d11f2661ca1845a47602ac7c99f69813e540

Also added unit tests for these edge cases with for the `(*Message).{Unmarshal,Response}` methods

Further, I decided to explicitly reject null payloads with the orderbook and unsub_orderbook unauthenticated routes, as well as in the the limit, market, and cancel routes handled by the OrderRouter since the zero-values are not sensible in any of these cases and we might as well be defensive.

server/auth/auth.go
```diff
@@ -1348,7 +1347,7 @@ func (auth *AuthManager) handleMatchStatus(conn comms.Link, msg *msgjson.Message
 			"cannot use route 'match_status' on an unauthorized connection")
 	}
 	var matchReqs []msgjson.MatchRequest
-	err := json.Unmarshal(msg.Payload, &matchReqs)
+	err := msg.Unmarshal(&matchReqs)
```
chappjc, line 1350: `nil` slice is no trouble for `handleMatchStatus` and `handleOrderStatus` afaict

server/market/bookrouter.go
```diff
@@ -542,8 +541,8 @@ func (r *BookRouter) msgOrderBook(book *msgBook) *msgjson.OrderBook {
 // notifications.
 func (r *BookRouter) handleOrderBook(conn comms.Link, msg *msgjson.Message) *msgjson.Error {
 	sub := new(msgjson.OrderBookSubscription)
-	err := json.Unmarshal(msg.Payload, sub)
-	if err != nil {
+	err := msg.Unmarshal(&sub)
+	if err != nil || sub == nil {
```
chappjc, line 545: The zero values for `OrderBookSubscription` will be sure to be rejected anyway, so best to bail here.  Same as in `handleUnsubOrderBook` and the handlers in `OrderRouter`.

server/swap/swap.go
```diff
@@ -998,8 +997,8 @@ func (s *Swapper) respondError(id uint64, user account.AccountID, code int, errM
 		return // this should not be possible, but don't pass nil msg to Send
 	}
 	if err := s.authMgr.Send(user, msg); err != nil {
-		log.Infof("Unable to send %s error response (code = %d, msg = %s) to disconnected user %v: %q",
-			msg.Route, code, errMsg, user, err)
+		log.Infof("Unable to send error response (code = %d, msg = %s) to disconnected user %v: %q",
+			code, errMsg, user, err)
```
chappjc, line 1001: `Response`s have no `Route` set, and indeed `msgjson.NewResponse` does not set one.

server/auth/auth.go
buck54321: But it's not sensical, so should be an error, no?

chappjc: Pushed https://github.com/decred/dcrdex/pull/1014/commits/bdf63b8ea94ad0c9a794fc5e9722db60d3009d2a.
`handleOrderStatus` caught it already with a len check, and I documented the intent there.
But `handleMatchStatus` currently is happy to issue a positive response of `[]` for a payload of `[]`, so I didn't want to change that behavior.  I did however make `null` payload an error now, and documented the current behavior.

JoeGruffins (approved):

itswisdomagain (approved):
LGTM

Issue #433, opened 2020-06-01
client: Persistent DEX ID


JoeGruffins:
Currently an account isn't linked to a DEX, but to a DEX's address, which may change. I think each DEX needs a more unique, permanent identifier. This should be a value decided by the DEX and communicated to the client upon `getfee` and `registration`. It should be the same value for all clients. It should also include a timestamped signature so that one DEX cannot impersonate another DEX. This way, a client can associate their account with a DEX, regardless of the DEX's location in cyberspace.

JoeGruffins:
@chappjc has pointed out that this value should be the DEXPubKey that is already stored in each account.

@behindtext has pointed out that in the future, servers will run in a server-mesh. That should also be taken into account.

chappjc:
I believe this is resolved by https://github.com/decred/dcrdex/pull/1015, in that it uses bytes of the dex pub key in the HD derivation path.  Is there more to this issue @buck54321 that was not addressed there?

ukane-philemon:
Closed?

JoeGruffins:
The address/host is still part of the account and saved in the db. I suppose this issue is solved, but there needs to be a way to change the address a dex points too, and the certificate, I think? Maybe this is better thought out in the context of "mesh" and bonds, as one account would have several addresses then? Closing anyhow. Will open a new one related to this when/if appropriate.

Issue #414, opened 2020-05-27
web site: wallet RPC pass is allowed to be saved by browser


chappjc:
![image](https://user-images.githubusercontent.com/9373513/83035603-43220000-9fff-11ea-86f7-4827de743ca5.png)

https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#The_autocomplete_attribute_and_login_fields

https://stackoverflow.com/questions/32369/disable-browser-save-password-functionality/37292424#37292424

PR #1015, opened 2021-03-12
core: app seed and HD keys

buck54321:
Overhauls the client credentials system and introduces deterministic account key generation.

### registration
**Old way**: Random private key generated for each new DEX account. Clients can export the key through the settings screen.

**New way**: An app seed is generated during client initialization. When a new dex account is created, the server's pubkey is used to deterministically generate a private key for the account. Users will save a copy of their app seed. In a simple world, that's all they need. During subsequent registrations, the same key will be generated, the server will report the key as already paid, and the registration fee payment will be skipped, the account ready for immediate use. The option to export individual account keys is still there though, as users may have a need to move accounts a la carte to e.g. other machines. Account keys can be imported as well. Upgraded (legacy) account keys and imported keys are treated identically, but differently than hierarchical keys.

### client/db
**Old way**: A single set of `Crypter` parameters linked to the app password was stored using the `Store` method. All DEX account private keys and wallet passwords were encrypted with that `Crypter`.

**New way**: `Store` is gone. There is now an **inner key**, which acts as a single static password for the client. The inner key is generated once and only once during initialization. The inner key is itself encrypted by the **outer key** that changes with the user password. This means that when a user updates their password, no account keys or wallet passwords need to be re-encrypted. The keys, encrypted app seed, and the relevant `Crypter` parameters are embodied by the `PrimaryCredentials struct`, and the `DB` interface has new methods for upgrading to, initializing, and updating the `PrimaryCredentials`.

### UI
The app password initialization form now has an option to specify the app seed. 

Since the password can now be cached for registration, and the /register view may start the user on any number of forms, I've added a password field to the one form that didn't already have one, the DEX address form. The field is displayed conditionally based on whether we already have the password cached. To make this work, I combined the two forms from /register and /settings into a single `template` for use on both pages. The form has also been given a corresponding javascript class `DEXAddressForm`.

When the user first registers, if not supplying a seed, they will be notified to go to the settings page and save their app seed asap. This is in lieu of a more proactive solution such as requiring the user to immediately echo back some portion of the seed. While I do recognize the value of such a flow, I would want to do that substantial work in a separate effort.

### client/core
- Creates the `PrimaryCredentials` during initialization, or during `Login` if upgrading. Decrypts the inner key or seed as necessary.
- Updates the `PrimaryCredentials` during app password changes.
- Deterministically generates DEX private keys (and soon wallet seeds for built-ins). When registering, recognizes new **msgjson** errors for account exists or account suspended, and act accordingly. For suspended account, the solution now is to just generate a random key, treat it as imported, and repay the fee. The implications are that if a client wants to get their account unsuspended via operator diplomacy, they need to do that before restoring the account.
- `PreRegister` method combines fetching of the DEX configuration and sending the /register request. This allows the client to know as soon as they've submitted the DEX URL whether we were able to restore the account with the HD key, solving the bad UX mentioned [here](https://github.com/decred/dcrdex/pull/1015#discussion_r626045335) and [here](https://github.com/decred/dcrdex/pull/1015#pullrequestreview-652943777).

### server
**db**: Return special errors during registration when the client's key already exists or is suspended. Return the fee coin as a courtesy.
**auth**: Translate db errors during registration to **msgjson** errors.


client/core/core.go
```diff
@@ -50,6 +51,8 @@ const (
 	// tickCheckDivisions is how many times to tick trades per broadcast timeout
 	// interval. e.g. 12 min btimeout / 8 divisions = 90 sec between checks.
 	tickCheckDivisions = 8
+	// Version is the core version.
+	Version = 1
 )
```
buck54321, line 56: The upgrade required us to have the old `Crypter` in-hand, so couldn't be done through a database upgrade. This PR introduces core versions and "authenticated upgrades" that are performed in `Login`.

client/core/core.go
```diff
@@ -1999,34 +2033,105 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
...
-		return nil, codedError(registerErr, err)
+		var msgErr *msgjson.Error
+		if !errors.As(err, &msgErr) || msgErr.Code != msgjson.AccountExistsError {
+			return nil, codedError(registerErr, err)
+		}
+
+		// It's an AccountExistsError. This is now account recovery, which is
+		// great news since we don't have to pay the fee.
+		// DRAFT TODO: How to handle suspended accounts?
+
```
buck54321, line 2083: Just want to draw attention to this remaining issue. What do we do if the recovered key is associated with a suspended account?

1. Can we allow repaying a fee on the same account ID to get unsuspended?
2. Should we just generate a random key and treat it as an imported account, requiring separate backup?

client/core/core.go
```diff
@@ -4078,6 +4338,15 @@ func (c *Core) connectDEX(acctInfo *db.AccountInfo) (*dexConnection, error) {
...
 	}
 
+	var dexPubKey *secp256k1.PublicKey
+	if len(dexCfg.DEXPubKey) > 0 {
+		var err error
+		dexPubKey, err = secp256k1.ParsePubKey(dexCfg.DEXPubKey)
+		if err != nil {
+			return nil, fmt.Errorf("error decoding public key: %w", err)
+		}
+	}
```
buck54321, line 4348: I've chosen to use the server's pubkey as the source of derivation indices rather than the host name. See also #433. This means that we needed the pubkey before the `'register'` request, but we were returning the pubkey in the `'register'` response. I've added the pubkey to the `'config'` response as well now, and we just handle old servers that don't have this change yet as legacy/imported accounts.

client/core/types.go
```diff
@@ -527,54 +531,42 @@ type dexAccount struct {
...
-		defer a.keyMtx.RUnlock()
-		// no need to generate a new privKey, return existing privKey AFTER unlocking.
-		err := a.unlock(crypter)
-		return a.privKey, err
-	}
+	encKey := a.encKey
 	a.keyMtx.RUnlock()
+	if encKey != nil {
+		return nil, nil, fmt.Errorf("key already set")
+	}
```
buck54321, line 559: This is a change in behavior, but I couldn't figure out why this shouldn't be an error. The only place we are using this function is in `Register`, where the presence of a encrypted key would signal an unexpected state.

client/webserver/site/src/html/register.tmpl
```diff
@@ -71,6 +77,9 @@
           Enter your app password to confirm DEX registration.
           When you submit this form, <span id="feeDisplay"></span> DCR will be spent from your Decred wallet to pay registration fees.
         </div>
+        <div id="restorationMsg" class="d-hide">
+          If the DEX server indicates during registration that it already knows our HD account number, no fee will be necessary.
+        </div>
```
buck54321, line 82: I'm not crazy about this UX. We should really just check whether the server knows our seed-derived key as soon as we know their pubkey after `/getdexinfo`, and skip this form altogether. Unfortunately, I didn't see a way to do that without opening a can of worms.

client/core/core.go
JoeGruffins: I don't think one is good, because what if the server decides to forgive us? Two is a shame when you already got this great deterministic scheme set up.

Can add one more `kid` in `setupCryptoV2` and if told here we're suspended, increment that index. Increment until we find an account ID that is not suspended/banned.

JoeGruffins (commented):
When restoring from seeds, the wallet config modules still say "Your Decred wallet is required to pay registration fees." Could be more ambiguous with "may be required"

After that I get the "Confirm Registration" which looks like I need to pay.

Then I need to restart dexc in order to get an authenticated connection and trade, and I don't think there's any indication to the user that we are good to go after that. Can of course be addressed in another pr.

------
I am unable to scroll on the markets page with this diff.

-----
Da memory test looks good.
```
(pprof) top
Showing nodes accounting for 1589.62MB, 94.38% of 1684.34MB total
Dropped 27 nodes (cum <= 8.42MB)
Showing top 10 nodes out of 42
      flat  flat%   sum%        cum   cum%
  317.72MB 18.86% 18.86%   317.72MB 18.86%  go.etcd.io/bbolt.Open.func1
  280.16MB 16.63% 35.50%   335.16MB 19.90%  go.etcd.io/bbolt.(*Bucket).openBucket
  228.37MB 13.56% 49.05%   228.37MB 13.56%  go.etcd.io/bbolt.(*node).put
  166.09MB  9.86% 58.92%   166.09MB  9.86%  decred.org/dcrdex/dex/encode.ExtractPushes
  134.51MB  7.99% 66.90%   134.51MB  7.99%  go.etcd.io/bbolt.(*Cursor).search
  121.10MB  7.19% 74.09%   121.10MB  7.19%  go.etcd.io/bbolt.(*node).read
  104.58MB  6.21% 80.30%   104.58MB  6.21%  go.etcd.io/bbolt.(*Bucket).write
  102.54MB  6.09% 86.39%   102.54MB  6.09%  decred.org/dcrdex/client/db.decodeMatchProof_v2
   79.46MB  4.72% 91.11%   506.62MB 30.08%  go.etcd.io/bbolt.(*Bucket).Bucket
   55.09MB  3.27% 94.38%   176.20MB 10.46%  go.etcd.io/bbolt.(*Bucket).node
```

client/db/interface.go
```diff
@@ -5,19 +5,31 @@ package db
...
+	// PrimaryCredentials fetches the *PrimaryCredentials.
+	PrimaryCredentials() (*PrimaryCredentials, error)
+	// LegacyKeyParams get the app key parameters stored using the deprecated
+	// Store method.
+	LegacyKeyParams() ([]byte, error)
+	// UpgradeLegacyCredentials updates the stored credentials to the new
+	// *PrimaryCredential format and updates the stored core version.
+	UpgradeLegacyCredentials(creds *PrimaryCredentials, oldCrypter, newCrypter encrypt.Crypter,
+		coreVersion uint8) (walletUpdates map[uint32][]byte, acctUpdates map[string][]byte, err error)
+	// CoreVersion is the core version set with SetCoreVersion. If no core
```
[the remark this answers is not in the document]
buck54321: Good catch. As of right now, I'm allowing the core version to be passed in with the UpdateLegacyCredentials function since that allows the entire update to occur in a single transaction. We may need a `SetCoreVersion` method someday, but maybe I'll just update the docs for now.

client/webserver/site/src/html/register.tmpl
buck54321: I opened the can. Might want to hold off on reviews for a minute.

buck54321:
To resolve the UX issues, I had to rework the registration sequence. Highlights are...

1. **app**: The `RegisterPage` keeps a cache of the user's password. The user will only ever have to enter the password once during registration now.
2. **server**: The server returns more detail from the `/register` route. Clients now know if their account exists, or exists but is suspended. The only remaining wedge case would be if the user nuked their db before the fee coin had confirmations. No resolution is available for this yet, and as of right now, the client will just re-pay the fee in this case.
3. **core**: Brought back the `PreRegister` method, which combines a fee retrieval with the call to `/register`. 

Revisiting logging and docs now.

buck54321:
fbd1ef8 is mostly just documentation updates, but sorry if you're mid-review, @chappjc.

client/core/core.go
```diff
@@ -490,6 +493,84 @@ func (dc *dexConnection) ack(msgID uint64, matchID order.MatchID, signable msgjs
...
 
+// register submits a new 'register' request to the server. Depending on the
+// server version, either encKey or encKeyLegacy will be set, but not both.
+func (c *Core) register(dc *dexConnection, crypter encrypt.Crypter) (regRes *msgjson.RegisterResult,
+	acctPubB, encKey, encKeyLegacy []byte, paid bool, err error) {
+
+	creds := c.creds()
+
+	// register is called before the *dexConnection is in the Core.conns map,
+	// so no risk of
```
buck54321, line 504: I have this comment fixed locally. Will push after review.

client/core/core.go
```diff
@@ -2257,68 +2421,94 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
...
-	msg := regRes.Serialize()
-	dexPubKey, err := checkSigS256(msg, regRes.DEXPubKey, regRes.Sig)
-	if err != nil {
-		return nil, newError(signatureErr, "DEX signature validation error: %v", err)
+	c.preRegisterMtx.RLock()
+	reg = c.preRegistration
+	c.preRegisterMtx.RUnlock()
+
+	if reg == nil || reg.host != host || time.Since(reg.stamp) >= time.Minute*5 {
+		var paid bool
```
chappjc, line 2443: <s>var paid bool</s>

client/core/core.go
```diff
@@ -2161,23 +2236,112 @@ func (c *Core) GetDEXConfig(dexAddr string, certI interface{}) (*Exchange, error
...
+		return nil, err
 	}
 
 	// Since connectDEX succeeded, we have the server config. exchangeInfo is
-	// guaranteed to return and *Exchange with full asset and market info.
+	// guaranteed to return an *Exchange with full asset and market info.
 	return dc.exchangeInfo(), nil
 }
 
+func (c *Core) PreRegister(dexAddr string, appPW []byte, certI interface{}) (*Exchange, bool, error) {
```
chappjc, line 2266: Needs some docs to explain the returns (paid mainly), but also when and how PreRegister should be used in conjunction with Register, and when Register can be used on its own.

client/core/core.go
```diff
@@ -2397,20 +2629,107 @@ func (c *Core) verifyRegistrationFee(assetID uint32, dc *dexConnection, coinID [
...
+	encInnerKey, err := outerCrypter.Encrypt(innerKey)
+	if err != nil {
+		return nil, nil, fmt.Errorf("inner key encryption error: %w", err)
+	}
+
+	// Generate a seed to use as the root for all future key generation.
+	seed := restorationSeed
+	if len(seed) == 0 {
+		seed = encode.RandomBytes(64)
+	} else if len(seed) != 64 {
```
chappjc, line 2716: The 512-bit/64-byte characteristic should probable be set with a constant since DB also requires this

client/core/core.go
```diff
@@ -2397,20 +2629,107 @@ func (c *Core) verifyRegistrationFee(assetID uint32, dc *dexConnection, coinID [
...
+
+	// Generate the outer key.
+	outerCrypter := c.newCrypter(pw)
+	encInnerKey, err := outerCrypter.Encrypt(innerKey)
+	if err != nil {
+		return nil, nil, fmt.Errorf("inner key encryption error: %w", err)
+	}
+
+	// Generate a seed to use as the root for all future key generation.
+	seed := restorationSeed
```
chappjc, line 2713: Looks like the input arg can just be `seed`.

client/core/core.go
```diff
@@ -2478,6 +2814,57 @@ func (c *Core) Login(pw []byte) (*LoginResult, error) {
...
+	}
+	oldCrypter, err := c.reCrypter(pw, oldKeyParams)
+	if err != nil {
+		return fmt.Errorf("legacy encryption key deserialization error: %w", err)
+	}
+
+	newCrypter, creds, err := c.generateCredentials(pw, nil)
+	if err != nil {
+		return err
+	}
```
chappjc, line 2835: Maybe it's just me, but I'd prefer my app to print the seed to stdout right at this point.  Users should absolutely grab the seed and back it up, and making them call an RPC or go to the /settings page, or (for Go consumers) call the `ExportSeed` method seems like enough of a hurdle that upgrading users will be unlikely to backup their new seed.

chappjc (changes requested):
First pass, looks good.  ACK of all conceptual changes.

JoeGruffins (commented):
I'm still unable to scroll on the Markets page.

---------------

When restoring an account, and adding the initial wallet, the language is :
```
Your Decred wallet is required to pay registration fees.
```
Perhaps that could be softened to `may be required`, as we would not need to pay the fee in the off chance we are restoring and have a good account there.

---------------

The app seed is stored by my browser and suggested the next time I go to that field. Is there a way to stop it from being stored, like a password?

---------------

If I had active orders, I am guessing they will not be refunded later unless I do it manually?
```
[ERR] CORE: parseMatches errors: order 06deff445e300b4a6ff10a5815ffd77d21648ac8a8d4cecf53ff7696f8a09db6 not found, order b30f64a57bff5e4530571a395d7c2451fd9140ab863aee31591a1871b16ed301 not found`
...
[ERR] CORE: Route 'revoke_match' notification handler error (DEX 127.0.0.1:17273): no order found with id b30f64a57bff5e4530571a395d7c2451fd9140ab863aee31591a1871b16ed301 (not an error if you've completed your side of the swap)
```

---------------

When attempting to restore seed and register with a dex with a suspended account I am unable to proceed:
````
[ERR] WEB: registration error: failed to unlock account: DecodeBlob: zero length blob not allowed
````

File: client/core/types.go [hunk 13]
```diff
@@ -537,85 +541,145 @@ type dexAccount struct {
...
+
+	// Start with the root key
+	extKey := root
+
+	for _, i := range kids {
+		extKey, err = genChild(extKey, i)
+		if err != nil {
+			return nil, nil, fmt.Errorf("genChild error: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
chappjc: Oh, yes I read that wrong.

client/core/types.go
```diff
@@ -537,85 +541,145 @@ type dexAccount struct {
...
+	// key, and there's nothing particularly special about doing it this way,
+	// but it works.
+
+	genChild := func(parent *hdkeychain.ExtendedKey, i uint32) (*hdkeychain.ExtendedKey, error) {
+		err := hdkeychain.ErrUnusableSeed
+		for err == hdkeychain.ErrUnusableSeed {
+			var kid *hdkeychain.ExtendedKey
+			// Hardened child by modding i first, technically doubling the
+			// collision rate, but OK.
+			kid, err = parent.Child(i%hdkeychain.HardenedKeyStart + hdkeychain.HardenedKeyStart)
```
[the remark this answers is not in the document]
chappjc: Oh, yes I read that wrong.  I saw it as index.

chappjc:
> The app seed is stored by my browser and suggested the next time I go to that field. Is there a way to stop it from being stored, like a password?

This reminds me of this issue: https://github.com/decred/dcrdex/issues/414


> If I had active orders, I am guessing they will not be refunded later unless I do it manually?

Yeah, I think all active orders cannot be handled.  Should not be disastrous if we log the raw refund txn when bcasting the contract as per https://github.com/decred/dcrdex/issues/1087

client/core/types.go
```diff
@@ -537,85 +541,145 @@ type dexAccount struct {
...
+		return nil, err
+	}
+
+	// Prepare the chain of child indices.
+	kids := make([]uint32, 0, 10) // 1 x purpose, 1 x version (incl. oddness), 8 x 4-byte uint32s.
+	// Hardened "purpose" key.
+	kids = append(kids, HDKeyPurpose)
+	// Second child is the the format/oddness byte.
+	kids = append(kids, uint32(dexPkB[0]))
+	byteSeq := dexPkB[1:]
```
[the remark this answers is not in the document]
buck54321: > Perhaps a mesh would be identified by a magic key

Oh yeah. In my rough vision, the mesh would be identified by a hash of a serialization of all critical network parameters.

client/core/core.go
```diff
@@ -490,6 +493,84 @@ func (dc *dexConnection) ack(msgID uint64, matchID order.MatchID, signable msgjs
...
+			}
+		}
+		dexReg := &msgjson.Register{
+			PubKey: acctPubB,
+			Time:   encode.UnixMilliU(time.Now()),
+		}
+		regRes = new(msgjson.RegisterResult)
+		err = dc.signAndRequest(dexReg, msgjson.RegisterRoute, regRes, DefaultResponseTimeout)
+		if err != nil {
+			var msgErr *msgjson.Error
```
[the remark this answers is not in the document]
buck54321: I actually did know that, but I thought since I was 4 levels of indentation deeper than the `for {`, this might improve readability.

client/core/core.go
```diff
@@ -2161,23 +2236,112 @@ func (c *Core) GetDEXConfig(dexAddr string, certI interface{}) (*Exchange, error
...
+		if err != nil {
+			return nil, false, err
+		}
+
+		c.log.Infof("%s is reporting that this account already exists. Skipping fee payment.", dc.acct.host)
+		return dc.exchangeInfo(), true, nil
+	}
+
+	c.preRegisterMtx.Lock()
+	c.preRegistration = &inProcessRegistration{
```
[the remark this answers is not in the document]
buck54321: Not sure about this. There is virtually no consequence for not following through with a pre-registration immediately, and I'd hate to lock up the `Register` route if the caller is registering multiple DEXes and goes out of order for some reason.

client/core/core.go
buck54321: Constant is good, but DB never actually handles the decrypted seed.

client/core/core.go
buck54321: I was trying to avoid dumping the seed in the terminal history, especially since users are likely not using a CLI and wouldn't be considering the risks involved with that.

JoeGruffins:
Looks to be working well! Just a few things I noticed:

Entering a bad seed has no indication on the UI. It does show up in a message in the logs.

----

It looks like this server log includes a pointer which is not helpful.
```
[DBG] AUTH: CreateAccount(&{48dcaa4a88e01fed985e2ae0258385c4fce859f143a1dd2b9d1e5717a087a341 0xc0007afc70}) failed: account already exists: b49a1a6265771b7df0be462b9acbf3de695bbcdacc4e881497db7859c76fa6e100000000
```

----

I am still unable to scroll on the markets page and cannot view my orders.

client/core/core.go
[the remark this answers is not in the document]
chappjc: I see, so the user could preregister for host A, say nevermind, preregister for host B and then go on to register host B.

This makes me think then that either: (1) `Pregegister` should return and `InProcessRegistration` that they pass to `Register` (would be confusing to consumers though), or (2) `c.preRegistration` should be a map keyed by host (would never clean out unused preregister results though).

So having talked through that I see why you have it this way and I think it's OK.

client/core/core.go
```diff
@@ -491,6 +494,87 @@ func (dc *dexConnection) ack(msgID uint64, matchID order.MatchID, signable msgjs
...
+			if err != nil {
+				return nil, nil, nil, nil, false, codedError(acctKeyErr, err)
+			}
+		}
+		dexReg := &msgjson.Register{
+			PubKey: acctPubB,
+			Time:   encode.UnixMilliU(time.Now()),
+		}
+		regRes = new(msgjson.RegisterResult)
+		err = dc.signAndRequest(dexReg, msgjson.RegisterRoute, regRes, DefaultResponseTimeout)
```
chappjc, line 530: The PR pretty well establishes using the `'register'` route to determine `AccountExistsError` or not, but this could be achieved with `'connect'` as well I believe, and that is something I am going to have to change in my `postbond` branch, which entirely deprecates the `'register'` route.  While I may not use `'connect'`, I will have to figure out account existence some other way.

client/core/core.go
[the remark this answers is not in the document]
chappjc: Sounds good, it's just part of the reason why in that snipped I put `if err == nil { break }`, to reduce the indent and bring that break closer to the loop start.  Not important.

client/core/bookie.go
```diff
@@ -651,6 +652,18 @@ func (dc *dexConnection) refreshServerConfig() error {
...
 	dc.assetsMtx.Unlock()
 
+	// If we're fetching config and the server sends the pubkey in config, set
+	// the dexPubKey now. We also know that we're fetching the config for the
+	// first time (via connectDEX), and the dexConnection has not been assigned
+	// to dc.conns yet, so we can still update the acct.dexPubKey field without
+	// a data race.
+	if dc.acct.dexPubKey == nil && len(cfg.DEXPubKey) > 0 {
+		dc.acct.dexPubKey, err = secp256k1.ParsePubKey(cfg.DEXPubKey)
+		if err != nil {
```
chappjc, line 662: I started making a similar change in the postbond branch because I needed DEXPubKey in the config response too, but `refreshServerConfig` is called on each reconnect.  So I had `connectDEX` do this with a `cfg` returned from `refreshServerConfig`: https://github.com/decred/dcrdex/pull/1120/commits/7c98429574c2f7826dd9b95af8777ba9113c639f#diff-13a4a6b12c1ae844ef90151638f194427cfdc545fceb58d52c0f6f33b81d859bR4940-R4953
And `handleReconnect` (the other caller of `refreshServerConfig`) doesn't touch `cfg.DEXPubKey`.
Is the idea here that `dc.acct.dexPubKey == nil` signifies whet it is actually the first connect? (*if* we're fetching the config for the first time)? I suppose that does make this safe with `cfgMtx` held...
I can do the same thing if that's the intent.

client/core/core.go
chappjc: Then since the user is getting an automatic upgrade when they login, I think there needs to be a notification with warning level severity when the upgrade is complete that they should go to /settings to export/backup their seed.

client/core/core.go
```diff
@@ -2500,6 +2824,23 @@ func (c *Core) Login(pw []byte) (*LoginResult, error) {
...
 
+	// Check for necessary authenticated updates.
+	coreVersion, err := c.db.CoreVersion()
+	if err != nil {
+		return nil, fmt.Errorf("CoreVersion error: %v", err)
+	}
+
+	switch coreVersion {
+	case 0:
+		err := c.upgradeCoreToV1(pw)
```
chappjc, line 2835: Similar to the unauthenticated upgrades, I'd like if we could make a backup db file first.

client/core/core.go
```diff
@@ -1777,93 +1907,45 @@ func (c *Core) WalletSettings(assetID uint32) (map[string]string, error) {
...
-	oldCrypter, err := c.encryptionKey(appPW)
+	if len(newAppPW) == 0 {
+		return fmt.Errorf("application password cannot be empty")
+	}
+	creds := c.creds()
+	if creds == nil {
+		return fmt.Errorf("no primary credentials. Is the client initialized?")
+	}
+
+	outerCrypter, err := c.reCrypter(appPW, creds.OuterKeyParams)
```
chappjc, line 1922: With the "authenticated" upgrades, it would seem to require `Login` before `ChangeAppPass`, which I don't think was required previously.

client/db/bolt/upgrades.go
```diff
@@ -28,6 +28,10 @@ var upgrades = [...]upgradefunc{
 	v3Upgrade,
 	// v3 => v4 splits orders into active and archived.
 	v4Upgrade,
+	// v4 => v5 adds PrimaryCredentials with determinstic client seed, but the
+	// only thing we need to do during the DB upgrade is to update the
+	// db.AccountInfo to differentiate legacy vs. new-style key.
+	v5Upgrade,
```
chappjc, line 34: I'm not _totally_ hung up on the idea of "authenticate upgrades", but I do have concerns I'd like to sort out, hopefully with just docs.
The main thing I'm wondering is what if there is a `v6Upgrade` unauthenticated upgrade here administered by client/db automatically on DB open, and it wants to do something with the structure of the account bucket and its sub-buckets and values, but now what if the first authenticated core version **have not been run yet**.  The two concerns are: (1) would v6 upgrade work if the core version upgrade associated with v5 hadn't been run yet, and (2) if v6 did work but login the first authenticated core versoin upgrade is going to be assuming v5, not v6 where it really is.

Start stacking up more unauthenticated and authenticate upgrades, and I can see this getting fragile.

I'm wondering more if there's a conceivable way to have the client/db-administered upgrades have a crypter argument that to all upgrades, even the authenticated ones, are in a single upgrade chain. If no crypter, break the upgrade chain and resume it on login.

buck54321: Good point. I bet we can just move all DB upgrades to `Login` and pass the `Crypter`?

client/core/core.go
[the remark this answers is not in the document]
buck54321: > I see, so the user could preregister for host A, say nevermind, preregister for host B and then go on to register host B.

Or preregister for A, preregister for B, register for A, register for B. That scenario would be improved by a `map[string]*inProcessRegistration` (+ thread safety + nanny for cleanup) , but I don't think it's worth the complexity considering everything.

client/db/bolt/upgrades.go
chappjc: Seems like a good bet. Seem like it's gonna work out?

chappjc: There are some nasty consequences of doing DB upgrades to login.  It means there can be no DB access prior to login, which means stuff like checking for active orders and advising the user to login immediately can't happen.  Maybe other things too.

buck54321: Yeah, upgrade in `Login` gets messy real quick. Hmm.

buck54321: I think I'm going to try to make it work.

martonp (commented):
On the show seed popup, if I press 'enter' instead of pressing the button, it will reload the page instead of show the seed.

buck54321:
I was going to refactor everything so that all db upgrades were performed during `(*Core).Login`, but it was a can of worms. I'm still running some manual tests, but would appreciate a concept ack, @chappjc.

```
Instead of creating a full-fledged system for authenticated upgrades,
perform an upgrade of the database structure via a standard DB upgrade,
but with a sparsely populated skeleton for the stored PrimaryCredentials.
Core recognizes this condition during Login and updates the DB.
We are relying on an assumption that even if future DB upgrades are
expecting a particular structure, they won't necessarily require that
the PrimaryCredentials fields are populated. This is probably safe,
since the PrimaryCredentials themselves are pretty worthless without
the user's password in hand.
```

chappjc:
> We are relying on an assumption that even if future DB upgrades are
> expecting a particular structure, they won't necessarily require that
> the PrimaryCredentials fields are populated. This is probably safe,
> since the PrimaryCredentials themselves are pretty worthless without
> the user's password in hand

Sounds good to me.  I was pretty anxious about making all upgrades happen after login anyway given it precludes the ability to interact with the DB at all prior to login (e.g. checking for active orders, etc.).

I'll get reviewing 7e5cc87 now.  Thanks!

client/core/core.go
buck54321: This is true only for the first `Login` after the db upgrade, I think.

chappjc:
I'd love for the server harness commit (https://github.com/decred/dcrdex/pull/1015/commits/c562ff54794e7e93952b127eba4d5367d49a4ec0) to go in straight away.  I do those actions manually frequently and it's so annoying.  Plus it's mostly unrelated to this PR.  Would you put that one in it's own PR and we can get it merged.

But what's the idea of:

```sh
HARNESS_DIR=$(
  cd $(dirname "$0")
  pwd
)
```

compared to 

```sh
HARNESS_DIR=$(dirname "$0")
```

?

chappjc (changes requested):
Working great for me.  I've done an upgrade, init with no existing seed, and init with provided seed successfully.

Just two comments in the review, plus I think an exportseed rpc would be valuable since headless clients cannot retrieve their automatically generated seed.

buck54321:
> But what's the idea of:
> 
> ```shell
> HARNESS_DIR=$(
>   cd $(dirname "$0")
>   pwd
> )
> ```

For a script invoked like `./harness`, `$(dirname "$0")` yields `.`.

chappjc:
Thanks.  You can drop the harness commit in a rebase now.

This PR works perfectly well for me, but would be good if someone else had a quick review of the latest changes too.

chappjc:
> On the show seed popup, if I press 'enter' instead of pressing the button, it will reload the page instead of show the seed.

@martonp  Do you know if this is still an issue?  I didn't test for this behavior.

@JoeGruffins Any more issues on the market page? Or UX issues with bad seed entry?

buck54321:
Oh, I see that I have a couple important @JoeGruffins comments to address still. UI stuff.

client/webserver/site/src/css/main.scss
```diff
@@ -77,7 +77,7 @@ a.logo {
 div.main {
   display: flex;
   flex-grow: 1;
-  min-height: 0;
+  min-height: min-content;
```
buck54321, line 80: I can't seem to figure out why I did this. Hmmm. But this seems to be the source of the unscrollable markets view.

client/core/bookie.go
buck54321: > Is the idea here that dc.acct.dexPubKey == nil signifies when it is actually the first connect? 

That is the idea, but please to improve things in postbond if you can. It's a little roundabout.

client/webserver/site/src/js/settings.js
```diff
@@ -84,9 +101,14 @@ export default class SettingsPage extends BasePage {
     Doc.bind(page.removeAccount, 'click', () => this.clearAccountFile())
     Doc.bind(page.addAccount, 'click', () => this.page.accountFile.click())
 
+    Doc.bind(page.exportSeed, 'click', () => this.showForm(this.page.exportSeedAuth))
+    Doc.bind(page.exportSeedSubmit, 'click', () => this.submitExportSeedReq())
```
martonp, line 105:
```suggestion
    forms.bind(page.exportSeedAuth, page.exportSeedSubmit, () => this.submitExportSeedReq())
```

JoeGruffins (approved):
My issues have been resolved.

JoeGruffins:
I was able to get the server into a wierd state when shutting down. It keeps repeating:
```
021-08-03 17:59:33.318 [TRC] SWAP: checkInactionEventBased: match d3796cff7bc483ebac3f0631eec4ffa750682497c9daf49e246c18e64cb50530 (TakerSwapCast)
2021-08-03 17:59:48.317 [TRC] SWAP: checkInactionEventBased: match d3796cff7bc483ebac3f0631eec4ffa750682497c9daf49e246c18e64cb50530 (TakerSwapCast)
2021-08-03 18:00:03.317 [TRC] SWAP: checkInactionEventBased: match d3796cff7bc483ebac3f0631eec4ffa750682497c9daf49e246c18e64cb50530 (TakerSwapCast)
2021-08-03 18:00:18.317 [TRC] SWAP: checkInactionEventBased: match d3796cff7bc483ebac3f0631eec4ffa750682497c9daf49e246c18e64cb50530 (TakerSwapCast)
```
Not sure if related to these changes or what. The harness stopped mining blocks, you know how it does, so that could also be the issue. Will attempt to reproduce.

JoeGruffins:
Sometimes it's like I can get the server to hang on shutdown but I'm unable to reproduce and it may be my imagination. May be nothing but I'll keep trying.

martonp (approved):

client/core/core.go
```diff
@@ -4506,12 +4885,10 @@ func (c *Core) connectDEX(acctInfo *db.AccountInfo, temporary ...bool) (*dexConn
...
 
-	if listen {
-		wsCfg.ReconnectSync = func() {
-			go c.handleReconnect(host)
-		}
-		wsCfg.ConnectEventFunc = dc.handleConnectEvent
+	wsCfg.ReconnectSync = func() {
+		go c.handleReconnect(host)
 	}
+	wsCfg.ConnectEventFunc = dc.handleConnectEvent
```
itswisdomagain, line 4891: This method's comment says `If temporary is provided and true, no connect and reconnect handlers are registered`. May need to take off that bit and add a comment here with the rationale for allowing reconnections on temp dcs.

client/core/core.go
```diff
@@ -4646,6 +5023,10 @@ func (c *Core) handleReconnect(host string) {
 	}
 }
 
+func (dc *dexConnection) broadcastingConnect() bool {
+	return atomic.LoadUint32(&dc.reportingConnects) == 1
+}
```
itswisdomagain, line 5028: Don't know if this is just me, but it wasn't immediately clear to me what purpose the `dc.reportingConnects` field served. I had to scan through the usages in code to realise it is basically used to differentiate between temporary and non-temporary dcs. I want to recommend renaming to `temporary` and this method to `isTemporary()`.

itswisdomagain (changes requested):
Looks like `GetFee` and `GetDEXConfig` are no longer used. Mark as deprecated or remove altogether?

client/core/account.go
```diff
@@ -101,8 +100,11 @@ func (c *Core) AccountExport(pw []byte, host string) (*Account, error) {
 
 	// Account ID is exported for informational purposes only, it is not used during import.
 	acct := &Account{
-		Host:          host,
-		AccountID:     accountID,
+		Host:      host,
+		AccountID: dc.acct.id.String(),
+		// PrivKey: Note that we don't differentiate between legacy and
+		// hierarchical private keys here. On import, all keys are treated as
+		// legacy keys.
```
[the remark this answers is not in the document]
buck54321: Good point. I did consider this, but just never pulled the trigger. Feel free to pick it up or open an issue.

client/core/core.go
```diff
@@ -1780,93 +1912,45 @@ func (c *Core) WalletSettings(assetID uint32) (map[string]string, error) {
...
 	if err != nil {
-		return newError(authErr, "ChangeAppPass password error: %v", err)
+		return newError(authErr, "old password error: %v", err)
 	}
-	// Set new password.
-	newCrypter, err := c.setAppPass(newAppPW)
+	innerKey, err := outerCrypter.Decrypt(creds.EncInnerKey)
 	if err != nil {
-		return err
+		return fmt.Errorf("inner key decryption error: %w", err)
```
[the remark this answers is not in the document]
chappjc: Just completing the upgrade with the app pw was what I had in mind too (didn't suggest).  Seems straightforward to just do it at this point if needed.

client/core/core.go
```diff
@@ -2336,68 +2501,97 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
...
+			// It's an AccountExistsError. This is now account recovery, which is
+			// great news since we don't have to pay the fee.
+			err = c.authDEX(dc)
+			if err != nil {
+				return nil, fmt.Errorf("error authorizing pre-paid account: %v", err)
+			}
+			c.log.Infof("%s is reporting that this account already exists. Skipping fee payment.", dc.acct.host)
+			res := &RegisterResult{FeeID: hex.EncodeToString(dc.acct.feeCoin), ReqConfirms: 0}
+			return res, c.markAccountPrePaid(dc, encKey, encKeyLegacy)
+		}
```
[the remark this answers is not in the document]
chappjc: That sounds like a good idea, although I request a function with a more meaningful name like `registerHD` instead of `registerV2`

chappjc:
> Looks like `GetFee` and `GetDEXConfig` are no longer used. Mark as deprecated or remove altogether?

Unfortunately we'll have to consider libdexc. `GetDEXConfig` is still used there even though frontend isn't using `/api/getdexinfo` anymore.  https://github.com/decred/decrediton/blob/3a6f2adb132c8deb614983447fe3a1e7946598c2/modules/dex/libdexc/adapter.go#L257

buck54321:
`GetFee` is still used by `RPCServer`.

client/core/core.go
```diff
@@ -494,6 +497,86 @@ func (dc *dexConnection) ack(msgID uint64, matchID order.MatchID, signable msgjs
...
+					paid = true
+				}
+				return
+			case msgjson.AccountSuspendedError:
+				// Retry with a legacy-style key.
+				c.log.Infof("HD account key for %s was reported as suspended. Generating a legacy-style key instead.", dc.acct.host)
+				accountSuspended = true
+				dc.acct.resetKeys()
+				continue
+			}
```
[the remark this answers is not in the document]
chappjc: I'm leaning strongly toward not supporting legacy servers.  I think it's undesirable to continue creating local client accounts that are not recoverable with the seed.

Regarding the suspended path, I'm a little uncertain.  If the user can't resolve the issue, there's probably good reason for the client to have to re-initialize to get back on that server.  But even if that doesn't feel right, the derivation path can be extended for that server, and it can be iterated through until a "not suspended" or new account is found. This would be preferable to defaulting to legacy keys.

With the bond work, I'm dropping the old register protocol support entirely anyway because it has to go away server side for a few reasons.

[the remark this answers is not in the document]
chappjc: Oh, I also was going to suggest we push 0.2.2 with DEXPubKey in the config response

client/core/core.go
```diff
@@ -494,6 +497,86 @@ func (dc *dexConnection) ack(msgID uint64, matchID order.MatchID, signable msgjs
...
+	// Check the DEX server's signature.
+	msg := regRes.Serialize()
+
+	// If this was a legacy registration, we'll be setting the pubkey here. For
+	// v2 registrations, we're overwriting the existing key, but whatever. We
+	// already checked that they're identical a few lines up.
+	dc.acct.dexPubKey, err = checkSigS256(msg, regRes.DEXPubKey, regRes.Sig)
+	if err != nil {
+		return nil, nil, nil, nil, false, newError(signatureErr, "%s pubkey error: %v", dc.acct.host, err)
+	}
```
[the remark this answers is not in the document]
chappjc: Relating to my comment about putting DEXPubKey in a 0.2.2 patch, this could go away anyway if we do that.  A patch release is not asking a lot for operators.  It can be applied well in advance of 0.3 too.

PR #1016, opened 2021-03-14
responsive GUI

### LINKED ISSUE #997: dexc ui: resize / responsive quirks to solve for Decrediton windowing
Here's a capture of some of the quirks of resizing the UI that are only recoverable with page refresh

https://user-images.githubusercontent.com/9373513/109718687-3c678100-7b6d-11eb-96de-7124c2cd3017.mp4

amassarwi:
Final Result:
https://user-images.githubusercontent.com/10324528/111234218-be888880-85f6-11eb-9057-a71a471bf40f.mp4

Closes #997.

### DIFF
--- a/client/webserver/site/src/css/market.scss
+++ b/client/webserver/site/src/css/market.scss
@@ -1,5 +1,4 @@
 div.marketlist {
-  width: 175px;
   border-right: 1px solid #626262;
   z-index: 99;
   background-color: #dbdbdb;
@@ -288,6 +287,15 @@ table.balance-table button:hover {
   }
 }
 
+#rightSide {
+  height: 85vh;
+  overflow-y: auto;
+
+  @media (min-width: map-get($grid-breakpoints, md)) {
+    height: 100vh;
+  }
+}
+
 #marketLoader {
   z-index: 50;
   background-color: rgba(0, 0, 0, 0.5);
@@ -346,6 +354,18 @@ table.balance-table button:hover {
   .ico-unlocked {
     color: #1e7d11;
   }
+
+  .userorders {
+    min-height: 200px;
+  }
+}
+
+.orders-left {
+  min-height: 350px;
+}
+
+.order-panel {
+  min-height: 500px;
 }
 
 #loaderMsg {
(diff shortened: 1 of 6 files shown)


amassarwi:
Current State:

https://user-images.githubusercontent.com/10324528/111085911-35a01d00-8522-11eb-8516-f1a760b61784.mp4

chappjc:
Looks nice.  Please tweak the url query at `<link href="/css/style.css?v=1rOIoK" rel="stylesheet">` to defeat browser cache

amassarwi:
@chappjc Thanks & will do, pushing even nicer final version in a bit.

client/webserver/site/src/css/market.scss
```diff
@@ -1,5 +1,4 @@
 div.marketlist {
-  width: 175px;
```
buck54321, line 2: I'd much prefer this div to stay a fixed width. The width was chosen because it should accommodate most ticker combos without wrapping. For instance, in the test app server, which has some 4-letter tickers

image

You can see some of the tickers are wrapping, and that the long name DEX is displaying fewer letters.

Wrapping could be solved with `white-space: nowrap`, but that will just result in the ticker text being cut off on the right side, I think. 

This calls for a fixed-width div. If you needed to get it smaller, you could specify a smaller width, logo size, and font size below a certain breakpoint.

buck54321 (changes requested):

chappjc (approved):

client/webserver/site/src/css/market.scss
```diff
@@ -288,6 +288,16 @@ table.balance-table button:hover {
   }
 }
 
+#rightSide {
+  width: calc(100% - 175px);
```
buck54321, line 292: It seems like `flex-grow: 1` would be better, but that doesn't seem to work. Hmm.

Good like this too.

buck54321 (changes requested):
At smaller widths, the chart is very short in the y, and there is a gap underneath the right column. Starts at the `md` breakpoint `768px`.

buck54321:
@amassarwi see https://github.com/amassarwi/dcrdex/compare/responsiveness...buck54321:amass?expand=1 for suggested fixes

amassarwi:
@buck54321 updated

buck54321 (approved):

Issue #1018, opened 2021-03-15
infinite recursion in Confirmations during reorg


chappjc:
We have to rethink the following pattern in https://github.com/decred/dcrdex/blob/bea35a9e57536b4c3d3bd8215992710e89fae0eb/server/asset/dcr/utxo.go#L135-L146

This is in both the Input and Output implementations of Confirmations.

The use of the blockCache in `(*TXIO).confirmations` is also difficult to follow and seems to lead to runaway recursion. e.g. https://github.com/decred/dcrdex/pull/1016/checks?check_run_id=2108248511#step:5:137

My opinion on the blockCache for `confirmations` is that it is not justified in its complexity although if it can be simplified that's fine.

PR #1019, opened 2021-03-19
multi: Add ETHSwapV0.

JoeGruffins:
Have the eth harness start with a contract. Add contract bindings and
    basic calls to clients.


JoeGruffins:
Is this test failure #1018 ?

server/asset/eth/contracts/ETHSwapV0.sol
```diff
@@ -0,0 +1,84 @@
...
+        _;
+    }
+
+    function swap(bytes32 secretHash)
+        public view returns(Swap memory)
+    {
+        return swaps[secretHash];
+    }
+
+    function initiate(uint refundTimestamp, bytes32 secretHash, address participant)
```
martonp, line 53: It might save some gas fees if there’s a batch version of this function that can be used in case there are multiple matches in the same epoch.

JoeGruffins: I think that's true. @chappjc want to add that functionality to the contract? Can also do it in V1 if we want to keep things as simple as possible for now.

chappjc: I didn't bring it up in this context, but it may be worth investigating.  It's not clear to me that gas use is reduced by batching in the same way that tx size and thus fees in utxo assets is reduced, but worth considering.

chappjc: Maybe we defer this idea though unless it's really simple and we clearly characterize the gas consumption changes for various cases.

JoeGruffins: Actually, would prefer to keep things lean and simple as possible for the first interation. Just needs to work and be secure.

chappjc:
You may rebase with PR #1078 in so we can give it a review.

client/asset/eth/rpcclient_harness_test.go
```diff
@@ -6,7 +6,10 @@
 // These tests are expected to be run in descending as some depend on the tests before. They cannot
 // be run in parallel.
 //
-// NOTE: Occationally tests will fail with "timed out". Please try again...
+// TODO: Occasionally tests will fail with "timed out" because transactions are
+// not being mined. If one tx is not mined, following txs from the same account
+// with a higher nonce also cannot be mined, so it causes all tests after to
+// fail as well. Find out why the first tx fails to be broadcast.
```
JoeGruffins, line 12: This seems to be unrelated to  #1130

JoeGruffins:
> You may rebase with PR #1078 in so we can give it a review.

Srry, stuck another one in front.

client/asset/eth/rpcclient_harness_test.go
JoeGruffins: Possibly realated...

https://github.com/ethereum/go-ethereum/issues/21385

I've spent a bit of time on this, but unable to stop this from happening so far.

dex/networks/eth/contracts/ETHSwapV0.sol
```diff
@@ -0,0 +1,84 @@
...
+        payable(msg.sender).transfer(swaps[secretHash].value);
+        swaps[secretHash].state = State.Redeemed;
+        swaps[secretHash].secret = secret;
+    }
+
+    function refund(bytes32 secretHash)
+        public
+        isRefundable(secretHash, msg.sender)
+    {
+        payable(msg.sender).transfer(swaps[secretHash].value);
```
chappjc, line 81: Big :+1: for `msg.sender` and not `tx.origin`.  But I wonder if we should outright prevent indirect calls by a proxy contract, say with something like `require(tx.origin == msg.sender)` to restrict the caller to being a regular "codeless" address.

Ref: https://github.com/ethereumbook/ethereumbook/blob/develop/09smart-contracts-security.asciidoc#txorigin-authentication

JoeGruffins: Lets do that.

The purpose of this pr is mostly to get more hooked up. I would not be suprised at all if there are vulns in the contract. It is mostly a stripped down version of the original at https://github.com/decred/atomicswap/blob/2f6679b9150b2ef364cd7bc75a49c21a7d124a91/cmd/ethatomicswap/contract/src/contracts/AtomicSwap.sol

I make no claim that is safe at this point.

chappjc: Indeed, we're working toward PoC.  The work proposed to the atomicswap repo is absolutely the right starting point, and done by ETH oriented developers.  I think we should all agree that none of this will be default-enabled on mainnet until much farther along in the process, and maybe not at all without outside consultation and extended testing... such is the nature of ETH.  We definitely need to be clear about that.

But I have been meaning to discuss other contract related concerns I've been pondering, ranging from specific lessons learned in recent weeks, to higher level decisions like using a single contract for EVERYONE.   I personally think this is one of the mistakes made by not just thor but eth and defi in general -- putting all the eggs in one basket with a single fallible contract.  It was a hard lesson for The DAO years ago, and again with the Thor Router.  (don't just hack one person, hack all involved)

In UTXO-land we make a unique contract for each swap, and while I get that a single reusalble contract published by a third party can have utility in many ways including saving on tx fees, it does make me uneasy... from a security point of view as well as bringing a third party (devs / contract publishers) into the mix.

Just something I'd like to consider in the future.  Nothing to hold up this work though since as you've said we're building the ETH **framework for DCRDEX**.  The contract we make now is meant to be the perfect solution for product.

JoeGruffins: Interesting.

I hope you don't mine me quoting this for the contract issue.

chappjc: No, I meant to be explicit about this here.   This is a process and PR 1019 isn't the end-all.

client/asset/eth/rpcclient_harness_test.go
JoeGruffins: Headway on this. After adding better logging, the server was telling us: 
`TRACE[07-28|11:37:15.770] Discarding invalid transaction           hash=5a1f01..40f83e err="nonce too low"`

Which suggests that the light client sometimes has trouble discerning the correct nonce. Probably because we destroy it's persistent data every test suite.

JoeGruffins:
So far unable to hit several errors that I was seeing in tests when using the same internal node directory. Please let me know of any random failures.

dex/networks/eth/contracts/ETHSwapV0.sol
```diff
@@ -0,0 +1,92 @@
...
+        swaps[secretHash].value = msg.value;
+        swaps[secretHash].state = State.Filled;
+    }
+
+    function redeem(bytes32 secret, bytes32 secretHash)
+        public
+        senderIsOrigin()
+        isRedeemable(secretHash, secret, msg.sender)
+    {
+        payable(msg.sender).transfer(swaps[secretHash].value);
```
chappjc, line 79: Maybe it doesn't apply to us, but I was looking into send/transfer/call, and their is a recommendation to stop using `call` or `transfer`, and instead use `call`.  https://consensys.net/diligence/blog/2019/09/stop-using-soliditys-transfer-now, https://consensys.github.io/smart-contract-best-practices/recommendations/#dont-use-transfer-or-send, and https://ethereum.stackexchange.com/a/78136

`send`/`transfer` being introduced after The DAO hack to mitigate reentrancy exploits seems to be seen as a half-measure at guarding reentrancy, plus the possibility of operations changing gas costs (see istanbul/SLOAD) makes `transfer` not recommended.

Although we're not making state updates to a "user balance" map, I'd be concerned that the `transfer` could use up all the gas before storing the `secret`, thus screwing over the counterparty!  I think this depends entirely on the receiving account and their `payable` function, if they have one?  I think prevailing advice for mitigating reentrancy is the checks-effects-interactions pattern, whereby you put **external calls at the after all state changes** (i.e. so that `isRedeemable` will prevent another send if state has already been updated before the external call that could call back to this function).  This pattern seems to be preferred to a mutex these days (https://consensys.github.io/smart-contract-best-practices/known_attacks/#reentrancy).

dex/networks/eth/contracts/ETHSwapV0.sol
```diff
@@ -0,0 +1,92 @@
...
+
+    modifier isRedeemable(bytes32 secretHash, bytes32 secret, address redeemer) {
+        require(swaps[secretHash].state == State.Filled);
+        require(swaps[secretHash].participant == redeemer);
+        require(sha256(abi.encodePacked(secret)) == secretHash);
+        _;
+    }
+
+    modifier isNotInitiated(bytes32 secretHash) {
+        require(swaps[secretHash].state == State.Empty);
```
chappjc, line 37: curious why checking like this when I feel `swaps[secretHash]` should just be verified to be null

chappjc (changes requested):

dex/networks/eth/contracts/ETHSwapV0.sol
JoeGruffins: It looks like we need to update the state, i.e. `swaps[secretHash].state = State.Redeemed;` before sending funds and that should prevent any replay attacks here, so any amount of gas would be fine. ofc no expert tho.

JoeGruffins: Ah but wait, if sending then failed, the funds would be locked forever... Ok, still studying.

JoeGruffins: > Although we're not making state updates to a "user balance" map

I think we pretty much are, and in it's current state it looks like a redeemer or even refunder could indeed drain the contract using a replay attack with enough gas, which transfer, and it looks like send too, limits. The difference between the two being that transfer throws an exception on error while send just return false.

chappjc: > Ah but wait, if sending then failed, the funds would be locked forever... Or would failure sending revert the state?

Wouldn't the txn fail, and the external call be reverted?

JoeGruffins: I'm on it with a solution and a test.

chappjc: Nice.  But I'm far less concerned about `transfer`->`call` advice than about preventing reentrancy attacks, which I think we'd accomplish with updating state before external calls.  In it's current form an attacker could indeed drain the entire contract given enough gas and their own contract with a `fallback` function that called back to out contract's `redeem`.  This is just the kind of thing I meant about defi liking to put all the eggs in one basket, but I don't have a better suggestion at the moment.

vctt94 (commented):
Hey @JoeGruffins, So when I try to add the alpha ethereum wallet from harness, I got the following:

image

Not sure if I am doing something wrong here, have you seen this?

dex/networks/eth/contracts/ETHSwapV0.sol
JoeGruffins: I took a deep delve into the contract and am feeling pretty confident now.

dex/networks/eth/contracts/ETHSwapV0.sol
JoeGruffins: I took a deep delve into the contract and am feeling pretty confident now.

JoeGruffins:
Have not looked into vctt's issue yet but will **soon**. It should not be pointed to any active nodes though, it should create a node at an empty location and reuse that.

chappjc (commented):
Awesome work doing a PoC for the reentrant contract vulnerability.  Just some conceptual questions about testing the vuln contract, and the possibility for public read-only debugging methods.

dex/networks/eth/contracts/ETHSwapV0.sol
```diff
@@ -14,47 +42,75 @@ contract ETHSwap {
...
 
+    // hasNoNilValues ensures that value and locktime of the swap are not zero.
+    // Zero values would likely indicate a mistake on the part of the sender.
     modifier hasNoNilValues(uint refundTime) {
         require(msg.value > 0);
         require(refundTime > 0);
         _;
     }
 
+    // senderIsOrigin ensures that this contract cannot be used by other
```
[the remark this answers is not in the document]
JoeGruffins: At present, I don't think contracts can actively do anything. They need a normal account to perform methods. The original account is always a human/normal account **afaict**. I really tried to find a way to spoof origin in the reentry attack but could not. If it is possible it is not common knowledge.

dex/networks/eth/contracts/ETHSwapV0.sol
```diff
@@ -14,47 +42,75 @@ contract ETHSwap {
...
+    // constructor is empty. This contract has no connection to the original
+    // sender after deployed. It can only be interacted with by users
+    // initiating, redeeming, and refunding swaps.
     constructor() {}
 
-    modifier isRefundable(bytes32 secretHash, address refunder) {
+    // isRefundable checks that a swap can be refunded. The requirements are
+    // the initiator is msg.sender, the state is Filled, and the block
+    // timestamp be after the swap's stored refundBlockTimestamp.
+    modifier isRefundable(bytes32 secretHash) {
```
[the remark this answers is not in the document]
JoeGruffins: The reasoning for removing was simply because it didn't seem necessary as we are just sending `msg.sender` anyway. So, for simplicity.

I have not used etherscan.io's read contract feature... I will go check it out. However, msg.send should be public as part of the initial transaction?

client/asset/eth/rpcclient_harness_test.go
```diff
@@ -623,78 +627,215 @@ func TestRefund(t *testing.T) {
...
+	wantBal := big.NewInt(0).Sub(originalAcctBal, txFee)
+
+	bal, err := ethClient.balance(ctx, simnetAddr)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// If the exploit worked, the test will fail here, with all funds
+	// drained from the contract.
+	if bal.Cmp(wantBal) != 0 {
```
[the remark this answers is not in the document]
JoeGruffins: You would then need to replace the contract bin hex in the test harness with the created `contract.go`, minus the leading `0x`, and restart the harness.

Currently we use the contract deployed in the harness to test. We could also deploy a new contract per test.

I could add a test that also deploys this contract separately to test that the vulnerability contract does work on a vulnerable contract. I didn't know if that was going too far though, as a use of time I mean.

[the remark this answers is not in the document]
JoeGruffins: Also, `--out` should point to the contract in `dex/networks/eth`. I actually meant to write a small Readme, so I think I'll do that.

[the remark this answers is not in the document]
chappjc: No need to test the vuln contact forever.  Just wondering

dex/networks/eth/contracts/ETHSwapV0.sol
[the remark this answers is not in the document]
chappjc: e.g.  some random contract https://etherscan.io/address/0xb1e9157c2fdcc5a856c8da8b2d89b6c32b3c1229#readContract

You can do stuff like balanceOf(account) in the explorer without messing with transactions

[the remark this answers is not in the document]
JoeGruffins: It looks like you can submit code to get some of that debugging. You can see that this contract is not set up to do that: https://etherscan.io/address/0x7f3a4085dcdccd5d52249dbf604ba9236b433040#code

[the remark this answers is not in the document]
JoeGruffins: I think we could make some separate public view functions that inform about states, and so would be visible and useable by an explorer. I read somewhere that public functions shouldn't call other public functions. Probably because of vulnerabilities like reentry attack creeping in undetected. Will look into it some more, but separate public functions for debugging like you say should be doable and fine, if you think it would be good.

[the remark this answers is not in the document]
chappjc: I'm sure we'll want such methods, and the contract source will be published and probably submitted too, but it's not something to hold up this PR anyway.
I expect these separate public functions won't add to gas cost for regular swaps, just the tx that publishes the actual contract, so I figure it's alright to add any such functions to the contract interface, right?

JoeGruffins:
@vctt94 This should work now. I wasn't able to observe your issue specifically, but there was some missing logic for creating an account. Added a very temporary function to add a testing account that has funds. I have not tried on testnet, but anyway you should point the node to an empty directory, default should be fine, which is `$HOME/.dexethclient`

I notice a display issue that I won't try to correct here, but Balance is in gwei which is 1e9th of an eth. This should say 5,000 eth.
notatoms

It looks like the UI assumes atoms in a few places.

vctt94 (commented):
Hey joe, yeah, it does create a new account now, on an empty dir, but yeah not many UI things working right now.

Will give it another round of tests later. Good job on this PR so far!

chappjc (approved):
LGTM.  This PR was super productive IMO.  I think we've learned a ton about ETH in general from you @JoeGruffins and about contract security.  Pretty cool we were actually able to drain the proposed contract from https://github.com/decred/atomicswap/pull/76 because it also did not set state prior to making external calls (https://github.com/decred/atomicswap/pull/76/files#diff-bbbe323063803d4430a2f007b2e39e9cdefae18521d8d28664686728278db38aR158-R161)

chappjc:
Also thanks for your testing @vctt94.  Very helpful, please keep it up.

JoeGruffins:
> OK, but out of curiosity, is there no rpc to create a new account and return the address and private key?

There is! But for testing it's convenient to have funds, and for mainnet I think we will use an account based on the app seed so it can be recreated.

dex/networks/eth/contracts/ETHSwapV0.sol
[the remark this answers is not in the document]
JoeGruffins: Yes, I think that the gas cost for the initial contract would be higher, probably not so much higher, but it shouldn't affect other methods at all.

xaur:
> we were actually able to drain the proposed (but modified) contract from [decred/atomicswap#76](https://github.com/decred/atomicswap/pull/76) because it also did not set state prior to making external calls

If there is a vuln in that PR it might be worth posting a summary in it. The PR looks dead but just in case.

chappjc:
Even though it didn't follow best practice, theoretically permitting reentrant calls, the send instruction should limit gas in most cases.
Not as bad as it could have been.

PR #1023, opened 2021-03-23
client/asset/dcr: support account-based locking

chappjc:
This supports Decred account-based encryption/passwords.

The `(*ExchangeWallet).Lock/Unlock` methods use the new `accountunlocked` wallet RPC being added in https://github.com/decred/dcrwallet/pull/2020 to reliably determine how (un)locking should be performed.  This also uses `accountunlocked` in the `Locked` method. I briefly considered a flag to indicate account vs. wallet locking that get's figured out on walletreconfigure/creation/startup/etc, but this is far simpler, and it won't get confused if the user sets the account pass while the ExchangeWallet is running.

Also note that unlike `walletlock`, `lockaccount` throws an error if already locked, so we catch that too.

NOTE: to test, you must use dcrwallet built from github.com/chappjc/dcrwallet v1.6.3-0.20210324151833-873c564976fd

### DIFF
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -65,7 +65,7 @@ const (
 )
 
 var (
-	requiredWalletVersion = dex.Semver{Major: 8, Minor: 4, Patch: 0}
+	requiredWalletVersion = dex.Semver{Major: 8, Minor: 5, Patch: 0}
 	requiredNodeVersion   = dex.Semver{Major: 6, Minor: 1, Patch: 2}
 )
 
@@ -166,6 +166,9 @@ type rpcClient interface {
 	GetTransaction(ctx context.Context, txHash *chainhash.Hash) (*walletjson.GetTransactionResult, error)
 	WalletLock(ctx context.Context) error
 	WalletPassphrase(ctx context.Context, passphrase string, timeoutSecs int64) error
+	AccountUnlocked(ctx context.Context, account string) (*walletjson.AccountUnlockedResult, error)
+	LockAccount(ctx context.Context, account string) error
+	UnlockAccount(ctx context.Context, account, passphrase string) error
 	Disconnected() bool
 	RawRequest(ctx context.Context, method string, params []json.RawMessage) (json.RawMessage, error)
 	WalletInfo(ctx context.Context) (*walletjson.WalletInfoResult, error)
@@ -630,6 +633,11 @@ func (dcr *ExchangeWallet) OwnsAddress(address string) (bool, error) {
 	}
 	va, err := dcr.node.ValidateAddress(dcr.ctx, a)
 	if err != nil {
+		// Work around a bug with dcrwallet that prevents validateaddress for
+		// locked accounts.
+		if isAccountNotEncryptedErr(err) {
+			return false, nil
+		}
 		return false, err
 	}
 	return va.IsMine && dcr.acct == va.Account, nil
@@ -881,7 +889,7 @@ func (dcr *ExchangeWallet) FundOrder(ord *asset.Order) (asset.Coins, []dex.Bytes
 	// Check wallet's fee rate limit against server's max fee rate
 	if dcr.feeRateLimit < ord.DEXConfig.MaxFeeRate {
 		return nil, nil, fmt.Errorf(
-			"%v: server's max fee rate %v higher than configued fee rate limit %v",
+			"%v: server's max fee rate %v higher than configured fee rate limit %v",
 			ord.DEXConfig.Symbol,
 			ord.DEXConfig.MaxFeeRate,
 			dcr.feeRateLimit)
@@ -2031,9 +2039,35 @@ func (dcr *ExchangeWallet) Address() (string, error) {
 	return addr.String(), nil
 }
 
+func (dcr *ExchangeWallet) accountUnlocked(ctx context.Context, acct string) (encrypted, unlocked bool, err error) {
+	var res *walletjson.AccountUnlockedResult
+	res, err = dcr.node.AccountUnlocked(ctx, acct)
+	if err != nil {
+		err = translateRPCCancelErr(err)
+		return
+	}
+	encrypted = res.Encrypted
+	if res.Unlocked != nil { // should only be when encrypted
+		unlocked = *res.Unlocked
+	}
+	return
+}
+
 // Unlock unlocks the exchange wallet.
 func (dcr *ExchangeWallet) Unlock(pw string) error {
-	return translateRPCCancelErr(dcr.node.WalletPassphrase(dcr.ctx, pw, int64(time.Duration(math.MaxInt64)/time.Second)))
+	encryptedAcct, unlocked, err := dcr.accountUnlocked(dcr.ctx, dcr.acct)
+	if err != nil {
+		return err
+	}
+	if !encryptedAcct {
+		return translateRPCCancelErr(dcr.node.WalletPassphrase(dcr.ctx, pw, 0))
+
+	}
+	if unlocked {
+		return nil
+	}
+
+	return translateRPCCancelErr(dcr.node.UnlockAccount(dcr.ctx, dcr.acct, pw))
 }
 
 // Lock locks the exchange wallet.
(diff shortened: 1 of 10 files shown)


go.mod
```diff
@@ -2,8 +2,10 @@ module decred.org/dcrdex
 
 go 1.15
 
+replace decred.org/dcrwallet => github.com/chappjc/dcrwallet v1.6.3-0.20210324151833-873c564976fd
```
chappjc, line 5: Temporary, of course.  The dcrwallet require will be updated when/if https://github.com/decred/dcrwallet/pull/2020 is merged and backported to the dcrwallet release-v1.6 branch.

client/asset/dcr/dcr.go
```diff
@@ -2031,27 +2039,82 @@ func (dcr *ExchangeWallet) Address() (string, error) {
...
 
 // Locked will be true if the wallet is currently locked.
+// Q: why are we ignoring RPC errors in this?
 func (dcr *ExchangeWallet) Locked() bool {
+	// First return locked status of the account, falling back to walletinfo if
+	// the account is not individually password protected.
+	encrypted, unlocked, err := dcr.accountUnlocked(dcr.ctx, dcr.acct)
+	if err != nil {
+		dcr.log.Errorf("accountunlocked error: %v", err)
+		// return false // or try walletinfo???
```
chappjc, line 2112: I'm not sure why we are eating any errors in this method and returning false...

client/asset/dcr/dcr.go
```diff
@@ -65,7 +65,7 @@ const (
 )
 
 var (
-	requiredWalletVersion = dex.Semver{Major: 8, Minor: 4, Patch: 0}
+	requiredWalletVersion = dex.Semver{Major: 8, Minor: 5, Patch: 0}
```
chappjc, line 68: Requires unreleased dcrwallet (unmerged work actually).  We technically already have a bug if the user decided to use individually-encrypted accounts, and fixing it properly would seem to require determining if the account is encrypted, otherwise we might be failing (un)lock appropriately.

go.mod
chappjc: The replace is now removed, and the dcrwallet require updated.

client/asset/dcr/dcr.go
buck54321: I think I just really didn't want an error return value. Would `true` be better?

buck54321 (changes requested):
I can't seem to get through registration. When submitting the form with the simnet harness DEX address and rpc certificate, the registration page reloads, the same form shows, and there are no errors. The `w-beta` tmux window shows

```
2021-03-29 11:14:31.745 [INF] RPCS: RPC method "accountunlocked" invoked by 127.0.0.1:43884
2021-03-29 11:14:31.746 [INF] RPCS: RPC method "walletlock" invoked by 127.0.0.1:43884
2021-03-29 11:14:31.746 [WRN] RPCS: Websocket receive failed from client 127.0.0.1:43884: websocket: close 1006 (abnormal closure): unexpected EOF                                                                                           
2021-03-29 11:14:31.746 [INF] RPCS: Disconnected websocket client 127.0.0.1:43884
```

Verified that I don't see this behavior on master. Did `npm ci` and rebuild, and a hard refresh in the browser. If you can't reproduce, I can continue digging.

chappjc:
Thanks for testing.  Will put it through the paces again to see what's up with registration.  Interesting that your w-beta log shows `walletlock` since I think I made beta an encrypted account.

client/asset/dcr/dcr.go
```diff
@@ -2358,7 +2421,7 @@ func (dcr *ExchangeWallet) signTx(baseTx *wire.MsgTx) (*wire.MsgTx, error) {
 		return nil, fmt.Errorf("failed to encode MsgTx: %w", err)
 	}
 	var res walletjson.SignRawTransactionResult
-	err = dcr.nodeRawRequest(methodSignRawTransaction, anylist{txHex}, &res)
+	err = dcr.nodeRawRequest(dcr.ctx, methodSignRawTransaction, anylist{txHex}, &res)
```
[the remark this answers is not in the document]
chappjc: Good catch.  This was from an earlier version of the PR where I used rawrequest for accountunlocked, but now it's in rpcclient.

chappjc:
I see the same register failure too.  Register works fine with dexcctl but with the browser interface it goes right back to "Add a DEX" and console says:

image

I wonder if I broke something in https://github.com/decred/dcrdex/pull/989/commits/d0e9ca73c87883516c8730ca120a2a9639d8ee4c.... will see what is happening on master too.

chappjc:
Ok I get the same error on `master`.  Pretty sure I broke something in that commit.

chappjc:
Interesting, it's the change:

```diff
-(c *Core) handleConnectEvent(host string, connected bool)
+(dc *dexConnection) handleConnectEvent
```

The Core method looked up the current `dexConnection` by `host` in the `conns` map, and the second already has the relevant `dexConnection`.  And we were relying on some funny interactions with getfee and register, where getfee's connection is temporary and with the **async disconnect event**, that only sent a conn ntfn if it found it in the map, the event wasn't being sent.  Now it is and it kills the frontend's handling of the register process.

I'm testing a resolution.

chappjc:
Ah, the actual issue is that now there's a "DEX connected" ntfn for the initial connection, while previously the first `handleConnectEvent` would not find the connection yet in the `c.conns` map and thus would not send a ntfn.  The conn event ntfn sent when the Register call first establishes the connection seems to be breaking the frontend.

I've disabled the event handlers this for the GetFee call since it's a temporary connection in https://github.com/decred/dcrdex/pull/1023/commits/145880a0323e1ff603119392e58d9befc2d795d7, but we need to figure out how to address the new DEX connected notification coming on initial connect that didn't happen before.  Thoughts about this @buck54321?  I think it was unintentional that the initial connect didn't send a connect event ntfn when it couldn't find it in the map yet.  A resolution probably involves changing how the connect ntfns are handled by the frontend, at least while in this register process.

chappjc:
Tracked it down to a page reload triggered by a connect event: https://github.com/decred/dcrdex/pull/989/commits/3623119fde172ef3efae9e0a7519242b402dc4b2#diff-b2831f969918f14b3aac94dc85fb036c6ca85613696187f7caaf83133ff4ec4bR513-R514
This seems to have been added in response to the last part of my comment here: https://github.com/decred/dcrdex/pull/989#issuecomment-802020164
Note sure how to resolve this.  Clearly we do not want to page to reload in the middle of the registration request.

client/webserver/site/src/js/app.js
```diff
@@ -511,7 +511,12 @@ export default class Application {
         break
       }
       case 'conn':
-        this.reconnected()
+        // TODO: if the page is presently loaded to markets showing "Connection
+        // to dex server failed" etc. indicating the initial DEX connection was
+        // never established and it is now connected for the first time, this
+        // should reload the markets page, but not for the initial connection.
+        // But we do not want to reload the page for every connect e.g.
+        // connections created by getfee and register.
```
buck54321, line 519: `App.reconnected` appears to be unused now.

chappjc: It's still bound to the messagesocket's connect handler `ws.connect(getSocketURI(), this.reconnected)` right?

client/asset/dcr/dcr.go
chappjc: :man_shrugging: I'm inclined to keep it lest something that expects this behavior break.

buck54321: Can you mention v1.0 in the comment and maybe we'll pick it up later?

client/webserver/site/src/js/app.js
chappjc: To summarize some chat, `ws.connect(getSocketURI(), this.reconnected)` is for dealing with browser->dexc connection changes where a fairly brute reload is not unwarranted, while a `'conn'` notification from dexc indicates the dexc->dcrdex server connection changed.  So however we end up handling these things,
- We don't want to reload just any page when a dex server connection changes. It's not inappropriate on the markets page, but it does break the register page flow, so this `'conn'` case should be removed.
- For the case where a configured dex server's connection has not yet been established (showing the message on the markets page, the solution from PR #989), to get the markets page loaded, we'd need either a page reload or a lot more info in the `conn` ntfn because we don't have any previous server asset or market config.
- For the browser->dexc reconnect wired by `ws.connect(...)`, that's also a bit excessive to reload the window, but not really a problem.

chappjc:
Everything is working again now AFAICT, including registration and the down-DEX on dexc startup that was addressed in PR #989.    @buck54321 This includes your commit and one after it to fix it up for down DEXes.  Diff for both combined: https://github.com/decred/dcrdex/compare/5d5550d9244435a76fd51f288fa9466410100a76..8924e8f9ab4e2366111b7ed25087d00a09671008
`handleConnectEvent` had to be modified to get user/exchange data that was completely missing in the case when dexc started while the DEX was inaccessible, but it's also sensible for ensuring the DEX config is up to date as the server may have restarted if we're seeing a conn event.

buck54321 (approved):

PR #1024, opened 2021-03-24
update chi for fixes to SIV snafu and a data race

chappjc:
This updates to chi/v5, which is the same as v1.5.4, just with proper
semantic import versioning (SIV). The project conceded the 1.5.x
approach did not work out since latest was previously v4.

See https://github.com/go-chi/chi/issues/561 for an epic discussion of this saga.

This also pulls in a chi data race fix in v1.5.2.


buck54321 (approved):

buck54321:
Oh. Just realized this is still in draft. Any reason to hold it up?

chappjc:
Haven't tested it myself is all.

PR #1025, opened 2021-03-25
client/{asset,core}: do not disconnect wallets in setWalletPassword

chappjc:
When an `asset.Wallet` is disconnected, it may not be possible to
reconnect it.  Document this with the `asset.Wallet` interface.
This is the case with Decred's rpcclient-powered `ExchangeWallet`
where calls to `rpcclient.(*Client).Connect` after having `Shutdown`
a previous connection will be met with the error:

`websocket client has already connected`

Always leave wallets connected in `setWalletPassword`. Callers like
`SetWalletPassword` have the `xcWallet` in the wallets map, so it should
stay live and usable.

In `ReconfigureWallet`, call `connectWallet` before `setWalletPassword`
anyway because the connection approach used by `setWalletPassword`
was too involved for the fresh `xcWallet` created by `ReconfigureWallet`
that is not yet in Core's wallets map.

### DIFF
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -561,9 +561,19 @@ func (dcr *ExchangeWallet) Info() *asset.WalletInfo {
 	return WalletInfo
 }
 
+// var logup uint32
+
+// func rpclog(log dex.Logger) {
+// 	if atomic.CompareAndSwapUint32(&logup, 0, 1) {
+// 		rpcclient.UseLogger(log)
+// 	}
+// }
+
 // Connect connects the wallet to the RPC server. Satisfies the dex.Connector
-// interface.
+// interface. WARNING: Once stopped, it cannot reconnect, requiring NewWallet to
+// construct a new ExchangeWallet to Connect again.
 func (dcr *ExchangeWallet) Connect(ctx context.Context) (*sync.WaitGroup, error) {
+	// rpclog(dcr.log)
 	dcr.ctx = ctx
 	err := dcr.client.Connect(ctx, false)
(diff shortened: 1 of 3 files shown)


client/asset/dcr/dcr.go
```diff
@@ -561,9 +561,19 @@ func (dcr *ExchangeWallet) Info() *asset.WalletInfo {
...
+// func rpclog(log dex.Logger) {
+// 	if atomic.CompareAndSwapUint32(&logup, 0, 1) {
+// 		rpcclient.UseLogger(log)
+// 	}
+// }
+
 // Connect connects the wallet to the RPC server. Satisfies the dex.Connector
-// interface.
+// interface. WARNING: Once stopped, it cannot reconnect, requiring NewWallet to
+// construct a new ExchangeWallet to Connect again.
```
chappjc, line 574: We could consider allowing this, but it would require creating a new rpcclient.Client on the fly, which would require storing the config and detecting if the client had been previously shutdown.  I'm inclined to document proper use here instead.

buck54321 (commented):
When attempting to update the wallet password through the /wallets view, I get the following error. 



This appears to be a bug that already exists on master.

chappjc:
Hmm, that's the bug this PR was meant to fix.  LOL.  I know I tested, but something must have changed.

chappjc:
I did just retest with this branch and that bug, which this PR is supposed to fix, does not happen show up.  When you get a chance can you fetch/reset/rebuild and all that to retest?  Thanks.

buck54321 (approved):

PR #1027, opened 2021-03-29
multi: no Seq in msgjson.EpochReportNote

chappjc:
Since there's no book or order updates with `epoch_report`, it should not increment `Seq`.  Having seq incremented in a message route that 0.1.x client's don't understand also breaks backward compatibility with 0.2 servers.

The book router messages that don't reflect book changes don't increment or include seq. For example, `MatchProof`, `TradeResumption`, and `TradeSuspension` when `Persist` is true.

This also updates the docs for `seq` in the `server/market.subscribers` struct to reflect the purpose of the [`Seq` value as per the spec](https://github.com/decred/dcrdex/blob/master/spec/orders.mediawiki/#order-book-subscriptions):

> The order book and all updates include a sequence ID, which increments by +1 whenever the DEX accepts, removes, or modifies an order. The client is responsible for tracking the sequence ID to ensure all order updates are received. If an update appears to be missing, the client should re-subscribe to the market to synchronize the order book from scratch.

### DIFF
--- a/client/core/bookie.go
+++ b/client/core/bookie.go
@@ -602,7 +602,9 @@ func (dc *dexConnection) refreshServerConfig() error {
 	bTimeout := time.Millisecond * time.Duration(cfg.BroadcastTimeout)
 	tickInterval := bTimeout / tickCheckDivisions
 	dc.log.Debugf("Server %v broadcast timeout %v. Tick interval %v", dc.acct.host, bTimeout, tickInterval)
-	dc.ticker.Reset(tickInterval)
+	if dc.ticker.Dur() != tickInterval {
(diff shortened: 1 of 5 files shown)


chappjc:
Tested with the following client-server pairs:
- 0.2 client and 0.2 server: 0.2 client handles the epoch_report notes as expected, sequence not lost
- 0.1.5 client and 0.2 server: 0.1.5 client logs unrecognized `epoch_report` notes, but sequence is not lost

Not tested: 0.1.5 server with 0.2 client, but in this case server will not send `epoch_report` notes.

buck54321 (approved):

PR #1030, opened 2021-03-31
client/core: inactive order handling

chappjc:
The `User` and `Exchanges` methods of `Core` include orders slices, which have
thus far included all orders including those that are inactive but about to
be retired on the next tick. Commit 461f3fc82c3c3de702e5f580d686612db5f36672 changes `marketTrades` to only return
**active** trades. Now `Logout`'s handling of active orders is consistent with the
orders slice returned by `User` or `Exchanges`, giving consumers such as
Decrediton or direct go consumers a better indication of when it is actually
safe to logout.

Commit 9fd6dbee5739ab6b14ff772c6f219dfd24bed791 also addresses some possibly missing cleanup actions on
`(*Core).Run` shutdown:  
The ticker has always retired inactive orders, which also returns any coins that
may still be locked in the wallets for such orders. This supplements the
shutdown of `(*Core).Run` with logging of active orders, and unlocks any
coins for such inactive orders _that the ticker has not yet retired_.

Note that if any coins are still locked for inactive orders, that
indicates a bug elsewhere since we should have returned these coins when
the order reached an inactive status. This should help debugging such cases.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1046,8 +1046,42 @@ func (c *Core) Run(ctx context.Context) {
 	stopDB()
 	dbWG.Wait()
 
+	// At this point, it should be safe to access the data structures without
+	// mutex protection. Goroutines have returned, and consumers should not call
+	// Core methods after shutdown. We'll play it safe anyway.
+
+	// Clear account private keys and wait for the DEX ws connections that began
+	// shutting down on context cancellation (the listen goroutines have already
+	// returned however). Warn about specific active orders, and unlock any
+	// locked coins for inactive orders that are not yet retired.
+	for _, dc := range c.dexConnections() {
+		// context is already canceled, allowing just a Wait(), but just in case
+		// use Disconnect otherwise it could hang forever.
+		dc.connMaster.Disconnect()
+		dc.acct.lock()
+
+		// Note active orders, and unlock any coins locked by inactive orders.
+		dc.tradeMtx.Lock()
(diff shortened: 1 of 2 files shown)


client/core/core.go
```diff
@@ -1046,8 +1047,42 @@ func (c *Core) Run(ctx context.Context) {
 	stopDB()
 	dbWG.Wait()
 
+	// At this point, it should be safe to access the data structures without
+	// mutex protection. Goroutines have returned, and consumers should not call
+	// Core methods after shutdown. We'll play it safe anyway.
```
buck54321, line 1052: > Goroutines have returned

I'm also seeing some routes like `Login`, `initializeDEXConnections`,  `checkUnpaidFees` that either are or start unmonitored goroutines  that don't use the `Core`'s `WaitGroup`. Not sure how problematic they are for this assertion though.

buck54321 (approved):

client/core/core.go
chappjc: True, `checkUnpaidFees` is a wild one, and I'd like to see it go away soon anyway, with the new registration protocol.  I've kept the `conns` and `trades` map accesses safe, since a concurrent map write would cause a panic, unlike with a regular variable.

Consumers really should not cancel until all calls to exported Core methods are returned, but that may not be possible if something is hung, so let's consider this comment a design goal.  I think the redeem search goroutine might also be wild.

PR #1031, opened 2021-04-01
webserver: add custom webserver site directory argument

buck54321:


client/webserver/webserver.go
```diff
@@ -136,26 +138,27 @@ func New(core clientCore, addr string, logger dex.Logger, reloadHTML bool) (*Web
...
-					"Place the 'site' folder in the executable's directory %q or the working directory, "+
-					"or run dexc from within the client/cmd/dexc source workspace folder.", execPath)
-			}
+	var siteDir string
+	absDir, _ := filepath.Abs("site")
+	for _, dir := range []string{
+		customSiteDir,
+		filepath.Join(execPath, "site"),
+		absDir,
+		filepath.Clean(filepath.Join(execPath, "../../webserver/site")),
```
buck54321, line 147: A couple function calls and some slice allocation that otherwise probably wouldn't happen, but a small price to pay to get out of the nesting situation.

client/cmd/dexc/config.go
```diff
@@ -75,6 +75,7 @@ func defaultHostByNetwork(network dex.Network) string {
 type Config struct {
 	AppData      string `long:"appdata" description:"Path to application directory."`
 	Config       string `long:"config" description:"Path to an INI configuration file."`
+	SiteDir      string `long:"sitedir" description:"Path to the 'site' directory with packaged web files. Unspecifed = default is good in most cases."`
```
chappjc, line 78: Would you add something to the sample config https://github.com/decred/dcrdex/blob/master/client/cmd/dexc/sample-dexc.conf?

chappjc (changes requested):
Looks good.  If you could update the sample dexc conf, this is g2g.

chappjc (approved):

PR #1032, opened 2021-04-02
client/orderbook: fix csum computation with missed preimages

chappjc:
client/orderbook: fix csum computation with missed preimages

The commitment checksum is computed from **all** orders in the epoch queue.
The shuffle seed is computed from the received preimages, but the csum
includes all the orders received in an epoch that were send to orderbook
subscribers in the `epoch_order` notes.

This fixes `(*EpochQueue).GenerateMatchProof` to use all epoch queue
commitments in csum computation, and it prints a message when
there are epoch queue orders that were not accounted for in either the
preimages or misses slices. An error is not thrown in this case since
one of the shuffle seed or commitment checksum will mismatch.

Now server and client compute the same csums when there are missed
preimages.  Their shuffle seeds already matched with misses.

Common test vectors are now used between all the client and server
packages that work with the csum.

### DIFF
--- a/spec/fundamentals.mediawiki
+++ b/spec/fundamentals.mediawiki
@@ -274,11 +274,13 @@ The process continues with the next order in the list and iterates until all
 orders have been processed.
 
 The preimages and seed are published at the start of the matching process. In
-addition, a '''''checksum''''', which is the Blake-256 hash of the
-lexicographically sorted order commitments, is sent to the client. Because
-the client already has the commitments from their order book subscription,
-the checksum allows the client to check that the orders underlying the preimages
-are indeed the same set received in their order notification feed.
+addition, a '''''checksum''''' defined as the Blake-256 hash of the
+concatenation of the lexicographically sorted commitments of all orders in the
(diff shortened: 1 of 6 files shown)


spec/fundamentals.mediawiki
```diff
@@ -274,11 +274,13 @@ The process continues with the next order in the list and iterates until all
...
-the client already has the commitments from their order book subscription,
-the checksum allows the client to check that the orders underlying the preimages
-are indeed the same set received in their order notification feed.
+addition, a '''''checksum''''' defined as the Blake-256 hash of the
+concatenation of the lexicographically sorted commitments of all orders in the
+epoch queue, is sent to the client. Because the client already has the
+commitments from their order book subscription, the checksum allows the client
+to check that the orders underlying the preimages are indeed the same set
+received in their order notification feed, and that no received orders were
+omitted from either preimage collection or the shuffling.
```
buck54321, line 283: ~~the~~

buck54321 (approved):

PR #1033, opened 2021-04-02
client/orderbook: fix copy Order operation in updateRemaining

chappjc:
This fixes a copy operation not being performed on account of a go compiler
simplification.  See staticcheck SA4001 / https://play.golang.org/p/1X44V19HxFj

Instead of making multiple copies on each add/book, go back to storing
pointers, but now `bookSide` is responsible for modifying `Quantity`, and
making copies when returning an `Order` from it's methods.

This also resolves an issue with the `OrderBook.{buys,sells}` fields being
overwritten unsafely in the `Reset` method.  The resolution is a new
`(*bookSide).reset` method to clear the internal structures. This is also
less error prone as the initial `orderPreference` cannot be changed.

### DIFF
--- a/client/orderbook/bookside.go
+++ b/client/orderbook/bookside.go
@@ -85,18 +85,18 @@ func (d *bookSide) Add(order *Order) {
 	}
 }
 
-// Remove deletes an order from its associated bin.
-func (d *bookSide) Remove(order *Order) error {
+// Remove deletes an order from the given rate bin.
+func (d *bookSide) Remove(oid order.OrderID, rateBin uint64) error {
 	d.mtx.Lock()
 	defer d.mtx.Unlock()
 
-	bin, exists := d.bins[order.Rate]
+	bin, exists := d.bins[rateBin]
 	if !exists {
-		return fmt.Errorf("no bin found for rate %d", order.Rate)
+		return fmt.Errorf("no bin found for rate %d", rateBin)
 	}
 
 	for i := range bin {
-		if order.OrderID == bin[i].OrderID {
+		if oid == bin[i].OrderID {
 			// Remove the entry and preserve the sort order.
 			if i < len(bin)-1 {
 				copy(bin[i:], bin[i+1:])
@@ -106,29 +106,31 @@ func (d *bookSide) Remove(order *Order) error {
 
 			// Delete the bin if there are no orders left in it.
 			if len(bin) == 0 {
-				delete(d.bins, order.Rate)
-				return d.rateIndex.Remove(order.Rate)
+				delete(d.bins, rateBin)
+				return d.rateIndex.Remove(rateBin)
 			}
 
-			d.bins[order.Rate] = bin
+			d.bins[rateBin] = bin
 			return nil
 		}
 	}
 
-	return fmt.Errorf("order %s not found", order.OrderID)
+	return fmt.Errorf("order %s not found with rate %d", oid, rateBin)
 }
 
 // UpdateRemaining updates the remaining quantity for an order.
-func (d *bookSide) UpdateRemaining(oid order.OrderID, remaining uint64) {
+func (d *bookSide) UpdateRemaining(oid order.OrderID, rateBin, remaining uint64) {
 	d.mtx.Lock()
 	defer d.mtx.Unlock()
 
-	for _, bin := range d.bins {
-		for _, ord := range bin {
-			if ord.OrderID == oid {
-				ord.Quantity = remaining
-				return
-			}
+	bin := d.bins[rateBin]
+	for i, ord := range bin {
+		if ord.OrderID == oid {
+			// Make a new Order so other copies aren't modified.
+			newOrder := *ord // deep copy
+			newOrder.Quantity = remaining
+			bin[i] = &newOrder
+			return
 		}
 	}
 }
@@ -164,12 +166,13 @@ func (d *bookSide) BestNOrders(n int) ([]*Order, bool) {
 
 // BestFill returns the best fill for the provided quantity.
 func (d *bookSide) BestFill(qty uint64) ([]*Fill, bool) {
-	d.mtx.RLock()
-	defer d.mtx.RUnlock()
 	return d.bestFill(qty, false, 0)
 }
 
 func (d *bookSide) bestFill(quantity uint64, convert bool, lotSize uint64) ([]*Fill, bool) {
+	d.mtx.RLock()
+	defer d.mtx.RUnlock()
+
 	remainingQty := quantity
 	best := make([]*Fill, 0)
 
@@ -236,7 +239,7 @@ func (d *bookSide) iterateOrders(check func(*Order) bool) {
 		bin := d.bins[d.rateIndex.Rates[calcIdx(ir)]]
 		for _, ord := range bin {
 			if !check(ord) {
-				break // really next rate bin? or break outer / return?
+				return
 			}
 		}
 	}
--- a/client/orderbook/bookside.go
+++ b/client/orderbook/bookside.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 
 	"decred.org/dcrdex/dex/calc"
+	"decred.org/dcrdex/dex/order"
 )
 
 // orderPreference represents ordering preference for a sort.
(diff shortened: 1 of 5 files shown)


client/orderbook/orderbook.go
```diff
@@ -237,8 +231,8 @@ func (ob *OrderBook) Reset(snapshot *msgjson.OrderBook) error {
 		defer ob.ordersMtx.Unlock()
 
 		ob.orders = make(map[order.OrderID]*Order)
-		ob.buys = newBookSide(descending)
-		ob.sells = newBookSide(ascending)
```
chappjc, line 241: These fields were note safe to change like this.

client/orderbook/orderbook.go
```diff
@@ -25,11 +25,6 @@ type Order struct {
 	Epoch uint64
 }
 
-// copy creates a copy. Note that the OrderID is not a deep copy.
-func (o *Order) copy() *Order {
-	return &(*o)
```
chappjc, line 30: The staticcheck SA4001 gotcha https://play.golang.org/p/1X44V19HxFj

client/orderbook/orderbook.go
```diff
@@ -461,6 +433,7 @@ func (ob *OrderBook) Unbook(note *msgjson.UnbookOrderNote) error {
 }
 
 // BestNOrders returns the best n orders from the provided side.
+// NOTE: This is UNUSED, and test coverage is a near dup of bookside_test.go.
```
chappjc, line 436: There's a huge chunk of test code in orderbook_test.go for this unused method, and it's is covered just as well by a similar test in bookside_test.go.  Maybe we eliminate it unless we're providing this `OrderBook` instance to Go consumers?

client/orderbook/orderbook.go
```diff
@@ -366,21 +361,15 @@ func (ob *OrderBook) updateRemaining(note *msgjson.UpdateRemainingNote, cached b
 
 	ob.ordersMtx.Lock()
 	ord, found := ob.orders[oid]
-	var newOrder *Order
-	if found {
-		newOrder = ord.copy()
-		newOrder.Quantity = note.Remaining
-		ob.orders[oid] = newOrder
-	}
```
chappjc, line 374: Updating the Order instance in the bookSide that contains it takes care of this.  `bookSide` is now in charge of `Quantity` access exclusively.

client/orderbook/bookside.go
```diff
@@ -142,7 +152,9 @@ func (d *bookSide) BestNOrders(n int) ([]*Order, bool) {
 		if count == 0 {
 			return false
 		}
-		best = append(best, ord)
+		// Return copies for thread-safe access to the Quantity field.
+		ordCopy := *ord
+		best = append(best, &ordCopy)
```
chappjc, line 157: I felt this was preferable to making multiple copies of each Order, and doing it those copies on each add/book.

client/orderbook/bookside.go
```diff
@@ -216,14 +230,13 @@ func (d *bookSide) idxCalculator() func(i int) int {
...
+	calcIdx := d.idxCalculator() // ascending or descending Rates index
 
-		for idx := 0; idx < len(bin); idx++ {
-			if !check(bin[idx]) {
-				break
+	for ir := range d.rateIndex.Rates {
+		bin := d.bins[d.rateIndex.Rates[calcIdx(ir)]]
+		for _, ord := range bin {
+			if !check(ord) {
+				break // really next rate bin? or break outer / return?
```
chappjc, line 239: Oh, I wasn't sure if this was a bug.  Seems like if `check` returns `false`, the iterations should stop altogether.  This `break` just breaks the *inner* for loop.

chappjc:
I don't have anything else for this, so marking r4r. It's not tested in a demanding simulation, but it didn't give any trouble in a quick interactive test I ran.  @buck54321 I hope this solution isn't too drastic of a change, but I decided it would be more efficient to copy when pulling out of OrderBook/bookSide than when putting in. https://github.com/decred/dcrdex/pull/1033/files#r606455877

client/orderbook/bookside.go
buck54321: We were only creating copies when the `Remaining` field was changed, which isn't going to happen for every order. Or maybe I'm not following what you mean by "making multiple copies... and doing those copies on each add/book".

The proposed change guarantees that every order is copied every time it is requested, which could be repeatedly if the user e.g. is switching back and forth between a couple of markets.

chappjc: Mmm, you're right, sorry about that bad assertion... it's the update that creates copies, not add/book.  What happens more frequently though, updating orders or requesting the fresh book?  Subscriptions and the associated orderbooks stick around for a bit before being deleted I believe.

Regardless, I'm not married to copying one place or the other, but part of what bothered me is that there was not previously a need to update the orders map at all on updateRemaining since the the purpose of the map appears to be: index of all orders in the book sides, and remember each one's rate for bookSide.Remove and side... while the bookSide deals with remaining exclusively. But copy-on-update requires updating the orders map unless we let those entries just have invalid remaining amounts.  Also I felt like UpdateRemaining vs ReplaceOrder was clearer about the purpose of the action.

I'll revisit this again, going back to `ReplaceOrder`, just with an actual copy (and the other bug fixes), but this `orders` map fundamentally bothers me for some reason.

buck54321: > this `orders` map fundamentally bothers me for some reason.

Probably because its purpose is to avoid having to iterate the `bookSide.bins`. And that would be great if we weren't passing the `*Order` out directly. So either we create/copy the `*Order`s on every call to `(*OrderBook).{Orders, BestFill, ...}`, or we iterate `bookSide.bins` to replace the order during `update_remining`. I do think the second option is more efficient, but I'm not digging my heels in on this one, so if you're hardcore to copy, go for it. 

We could almost get away with just indexing the order id -> rate rather than the whole `*Order`, but we are using the `Side` field in [one other place](https://github.com/chappjc/dcrdex/blob/be02613c3ce6bdc09a81a2b0b78e4334f290b8b6/client/orderbook/orderbook.go#L415-L426). Could probably be worked around though.

chappjc: > We could almost get away with just indexing the order id -> rate rather than the whole `*Order`, but we are using the `Side` field in [one other place](https://github.com/chappjc/dcrdex/blob/be02613c3ce6bdc09a81a2b0b78e4334f290b8b6/client/orderbook/orderbook.go#L415-L426). Could probably be worked around though

This had occurred to me too.  That is making `orders map[OrderID]struct{rate uint64; sell bool}`.  I'm going to get to this PR today.

client/orderbook/bookside.go
buck54321: Looks like it should just `return`

client/orderbook/orderbook.go
```diff
@@ -624,5 +597,7 @@ func (ob *OrderBook) BestFill(sell bool, qty uint64) ([]*Fill, bool) {
 // BestFillMarketBuy is the best (rate, quantity) fill for a market buy order.
 // The qty given will be in units of quote asset.
 func (ob *OrderBook) BestFillMarketBuy(qty, lotSize uint64) ([]*Fill, bool) {
+	ob.sells.mtx.RLock()
+	defer ob.sells.mtx.RUnlock()
 	return ob.sells.bestFill(qty, true, lotSize)
```
buck54321, line 602: Shouldn't we lock the mutex in `(*bookSide).bestFill`?

client/orderbook/orderbook.go
chappjc: I left this for now.

client/orderbook/bookside.go
chappjc: Back to making a copy on `update_remaining`, but it's done inside `bookSide` now. The `OrderBook.orders` map no longer stores the `*Order`, which was my main hangup with the map.  It indexes the orders and keeps the info needed to locate the order in a `bookSide` (rate and side), nothing more.

client/orderbook/bookside.go
chappjc: Changed to `return`.  Pretty certain the idea is that there's no reason to check other bins if check says false for an order in current bin?

buck54321 (approved):
Much better like this.

client/orderbook/bookside.go
chappjc: Oh silly me I forgot to remove this copy.

PR #1035, opened 2021-04-03
client/{dexc,webserver}: cpu and http profilers

chappjc:


client/cmd/dexc/main.go
```diff
@@ -24,20 +25,42 @@ import (
...
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
+	}
+	os.Exit(0)
+}
+
+func mainCore() error {
 	appCtx, cancel := context.WithCancel(context.Background())
+	defer cancel() // don't leak on the earliest returns
```
buck54321, line 40: So much better.

buck54321 (approved):

PR #1040, opened 2021-04-07
round-up of testing fixes

chappjc:
This is a few minor changes I made in testing for 0.2:

1. Enable dcrwallet's http profiler in the alpha and beta harnesses (ports 19577 and 19578)
2. Use a sane minimum response timeout for init/redeem requests, which is really only an issue with super short broadcast timeouts such as the 1 min timeout used by the dcrdex harness.
3. Fix a loadbot `Ticker` leak that was creating tickers that would never be garbage collected and would tick (goroutines) forever.  These were created in `orderMetered`, so quite frequently.  This commit also changes `blockEvery2` so that the smallest possible delay for a block is 2 seconds instead of 1. This is because dcrwallet will eventually miss too many ticket purchases with enough fast blocks and dcrd will suffer from ticket exhaustion.
4. Fix `TestRegister'`s `queueTipChange` being able to `Errorf` after `TestRegister` returned.  This occurred in github CI, but because this caused it's own testing error from go, it wasn't clear what if anything in the test failed.  Now the goroutine from `queueTipChange` is allowed to return before `TestRegister` returns.


dex/testing/loadbot/mantle.go
```diff
@@ -211,6 +211,7 @@ func (m *Mantle) orderMetered(ords []*orderReq, dur time.Duration) {
 	}
 	go func() {
 		ticker := time.NewTicker(dur / time.Duration(len(ords)))
+		defer ticker.Stop()
 		for {
 			select {
```
buck54321, line 216: New `Timer` every loop might be better than a `Ticker` too, but that's unrelated to your fix.

buck54321 (approved):
Looks good.

related blabbering: `loadbot` could use some iteration once the ticket buyer issue is fixed. My typical match rate on the `compound` program is just under one per second. It'd be good to see rates of at least a few per second, but achieving that isn't as simple as just turning the knob, if I remember correctly.

chappjc:
> It'd be good to see rates of at least a few per second, but achieving that isn't as simple as just turning the knob, if I remember correctly

Much of this PR came from working with the `heavy` program.  One hurdle with that is ticket exhaustion if dcrwalet is too slow / blocks too fast, and I think I've mitigated that a little with the 2 second min block time, but I suspect we'll need to bump up `ticketbuyer.limit` to 7-10 from it's current of 6.

Another thing I inevitably hit with `heavy` is insufficient balance after maybe ~1hr of running. Not sure why since I see it replenishing from alpha.  I think alpha might run out but I haven't checked to see who ended up with all the funds in this case.

`heavy` is also the only program to use the beta wallet in addition to alpha, and this revealed a few interesting things as well.  One is init/redeem request timeouts getting computed as 60sec / 5 = 15 sec ended up causing the request to timeout before too long into the test.  Server would actually respond, just a bit later.  This happened when things got really busy.  Initially I changed the dcrdex harness' bcasttimeout to 2 min, but reverted that.  I suspect it is probably smarter to just not use the dcrdex harness with loadbot.  Having 10k ticker goroutines running was driving up CPU too I believe, which made the init/redeem timeouts more likely after a while of running.

A lot of the dcrwallet issues I encountered came from `heavy` too, including a [concurrent map read/write](https://github.com/decred/dcrwallet/pull/2024) and a deadlock that is still unsolved and I can't reproduce again.  But with the http profilers enabled for wallet now, we can grab the goroutine stack when it hangs again.

Issue #1041, opened 2021-04-07
should show lot size when registering


chappjc:
The "Confirm Registration" dialog should show some of the server's config so the user can make an informed decision before sending their fee.  The lot sizes for the different markets are the most important config to show.

![image](https://user-images.githubusercontent.com/9373513/113890843-2495a500-978a-11eb-86b3-e4058eb18abe.png)

PR #1042, opened 2021-04-07
client/{core,webserver}: getdexinfo instead of getfee

chappjc:
image

Resolves https://github.com/decred/dcrdex/issues/1041

The idea is the user should have all the information about the DEX markets and assets before they shell out to register, and lot size is arguably the most important (what if they wanted to trade less than lot size?).

I'm aware that if the DEX does not have DCR supported this will be lacking a value, but presently DEX servers must support it.  Also, this message will have to change to something more informative about all the markets at some point, as will the fee amount shown when multiple reg fee assets becomes supported.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -229,6 +229,51 @@ func (dc *dexConnection) setRegConfirms(confs uint32) {
 	dc.regConfirms = &confs
 }
 
+func (dc *dexConnection) exchangeInfo() *Exchange {
+	// Set AcctID to empty string if not registered.
+	acctID := dc.acct.ID().String()
+	var emptyAcctID account.AccountID
+	if dc.acct.ID() == emptyAcctID {
+		acctID = ""
+	}
+
+	dc.cfgMtx.RLock()
+	cfg := dc.cfg
+	dc.cfgMtx.RUnlock()
+	if cfg == nil { // no config, assets, or markets data
+		return &Exchange{
+			Host:       dc.acct.host,
+			AcctID:     acctID,
+			FeePending: dc.acct.feePending(),
+			Connected:  atomic.LoadUint32(&dc.connected) == 1,
+			// RegConfirms:   dc.getRegConfirms(), // could show, but we don't know required
+		}
+	}
+
+	dc.assetsMtx.RLock()
+	assets := make(map[uint32]*dex.Asset, len(dc.assets))
+	for assetID, dexAsset := range dc.assets {
+		assets[assetID] = dexAsset
+	}
+	dc.assetsMtx.RUnlock()
+
+	return &Exchange{
+		Host:          dc.acct.host,
+		AcctID:        acctID,
+		Markets:       dc.marketMap(),
+		Assets:        assets,
+		FeePending:    dc.acct.feePending(),
+		Connected:     atomic.LoadUint32(&dc.connected) == 1,
+		ConfsRequired: uint32(cfg.RegFeeConfirms),
+		RegConfirms:   dc.getRegConfirms(),
+		Fee: &FeeAsset{ // v0 is only DCR, but there will be a RegFees map[string]*FeeAsset field
+			ID:    42,
+			Amt:   cfg.Fee,
+			Confs: uint32(cfg.RegFeeConfirms),
+		},
+	}
+}
+
 // hasActiveAssetOrders checks whether there are any active orders or negotiating
 // matches for the specified asset.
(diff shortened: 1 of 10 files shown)


client/core/core.go
```diff
@@ -218,6 +218,51 @@ func (dc *dexConnection) setRegConfirms(confs uint32) {
...
+		Assets:        assets,
+		FeePending:    dc.acct.feePending(),
+		Connected:     atomic.LoadUint32(&dc.connected) == 1,
+		ConfsRequired: uint32(cfg.RegFeeConfirms),
+		RegConfirms:   dc.getRegConfirms(),
+		Fee: &FeeAsset{ // v0 is only DCR, but there will be a RegFees map[string]*FeeAsset field
+			ID:    42,
+			Amt:   cfg.Fee,
+			Confs: uint32(cfg.RegFeeConfirms),
+		},
```
chappjc, line 262: In anticipation of multiple registration fee assets, which is started in https://github.com/decred/dcrdex/pull/1017/files#diff-6fe96257ee07551d0c31a6e6d14c4d9557b48de4262b5fede53cc307eaee5a91R1100

client/core/core.go
```diff
@@ -2061,26 +2075,46 @@ func (c *Core) AutoWalletConfig(assetID uint32) (map[string]string, error) {
...
+// GetFee creates a temporary connection to the specified DEX Server and fetches
+// the registration fee. The connection is closed after the fee is retrieved.
 // Returns an error if user is already registered to the DEX. A TLS certificate,
 // certI, can be provided as either a string filename, or []byte file contents.
 func (c *Core) GetFee(dexAddr string, certI interface{}) (uint64, error) {
+	exch, err := c.GetDEXConfig(dexAddr, certI)
+	if err != nil {
+		return 0, err
+	}
+	return exch.Fee.Amt, nil
```
chappjc, line 2098: `GetFee` must change or be removed when multiple fee assets become supported.

client/core/types.go
```diff
@@ -448,6 +455,8 @@ type Exchange struct {
 	Connected     bool                  `json:"connected"`
 	ConfsRequired uint32                `json:"confsrequired"`
 	RegConfirms   *uint32               `json:"confs,omitempty"`
+	Fee           *FeeAsset             `json:"feeAsset"`
+	// RegFees       map[string]*FeeAsset  `json:"regfees"`
```
chappjc, line 459: `Fee` could be `FeeQty` at present given the hard-coded assumption of DCR as the fee asset, but I want to float the use of `FeeAsset` so frontend can get use to having it be fully specified like this.

client/webserver/site/src/html/register.tmpl
```diff
@@ -71,6 +71,12 @@
           Enter your app password to confirm DEX registration.
           When you submit this form, <span id="feeDisplay"></span> DCR will be spent from your Decred wallet to pay registration fees.
         </div>
+        <div class="fs16 mt-4">
+          The DCR lot size for this DEX is <span id="dexDCRLotSize"></span> DCR.
+          All trades are in multiples of this lot size.
+          This is the minimum possible trade amount in DCR.
+          <!-- this will change when lot size is a market setting, not an asset setting -->
+        </div>
```
chappjc, line 79: This is bare minimum to resolve https://github.com/decred/dcrdex/issues/1041, but with the getdexinfo response, there can be a table of the markets and all the parameters, and other reg fee assets when we get there.

client/webserver/site/src/js/register.js
```diff
@@ -162,9 +162,11 @@ export default class RegistrationPage extends BasePage {
 
       return
     }
-    this.fee = res.fee
+    this.fee = res.exch.feeAsset.amount
 
-    page.feeDisplay.textContent = Doc.formatCoinValue(res.fee / 1e8)
+    page.feeDisplay.textContent = Doc.formatCoinValue(this.fee / 1e8)
+    const dcrAsset = res.exch.assets['42'] // res.exch.markets['dcr_btc']
```
buck54321, line 168: Can the comment go?

Could create a reverse mapping of `BipIDs` in `doc.js` for so we could do something like `const dcrAsset = res.exch.assets[BipIDSymbols['dcr']]` for readability, but this is going to change soon anyway, so this is fine too.

buck54321 (approved):
No stoppers. A couple small suggestions.

chappjc:
Did a quick re-test after the polish commit. Review suggestions applied.

PR #1046, opened 2021-04-12
spec: Add apiver to server config response.

JoeGruffins:
Work towards #530

PR #1047, opened 2021-04-12
multi: Version server API.

### LINKED ISSUE #530: multi: Enable client/server API version checking.
I think this is not currently enabled because of the current development status, but at some point the client and server must check for ability to communicate before attempting to communicate. The places where this will happen are in the `connect` route.

For the client that is currently here where the client version is set:
https://github.com/decred/dcrdex/blob/4d807f2876734b91ac090fa8a76bf4be2778f90b/client/core/core.go#L1834-L1841

And where the server should check at some point:
https://github.com/decred/dcrdex/blob/4d807f2876734b91ac090fa8a76bf4be2778f90b/server/auth/auth.go#L697-L705

JoeGruffins:
Add a version to the server that is communicated in the Config response.
Client checks and saves the version. The burden of conforming to an
API version is placed completely on the client.

depends on #1046 
closes #530

### DIFF
--- a/client/core/bookie.go
+++ b/client/core/bookie.go
@@ -4,6 +4,7 @@
 package core
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 	"sync/atomic"
@@ -597,6 +598,38 @@ func (dc *dexConnection) refreshServerConfig() error {
 	err := sendRequest(dc.WsConn, msgjson.ConfigRoute, nil, cfg, DefaultResponseTimeout)
 	if err != nil {
 		return fmt.Errorf("unable to fetch server config: %w", err)
+	}
+
+	// Check that we are able to communicate with this DEX.
+	apiVer := atomic.LoadInt32(&dc.apiVer)
+	cfgAPIVer := int32(cfg.APIVersion)
+
+	if apiVer != cfgAPIVer {
+		// If not initiation, do not allow api changes
+		// between connections.
+		if apiVer != -1 {
+			return fmt.Errorf("server API version unexpectedly changed from %v to %v",
+				apiVer, cfgAPIVer)
+		}
+		if found := func() bool {
+			for _, version := range serverAPIVers {
+				ver := int32(version)
+				if cfgAPIVer == ver {
+					dc.log.Debugf("Setting server api version to %v.", ver)
+					atomic.StoreInt32(&dc.apiVer, ver)
+					return true
+				}
+			}
+			return false
+		}(); !found {
+			errMsg := fmt.Sprintf("unknown server API version: %v", cfgAPIVer)
+			if cfgAPIVer > apiVer {
+				// TODO: Send this to the Web UI.
+				errMsg = fmt.Sprintf("%s\nYou may need to update your client to trade at %v.",
+					errMsg, dc.acct.host)
+			}
+			return errors.New(errMsg)
+		}
 	}
 
 	bTimeout := time.Millisecond * time.Duration(cfg.BroadcastTimeout)
--- a/client/core/bookie.go
+++ b/client/core/bookie.go
@@ -595,8 +595,8 @@ func handleTradeResumptionMsg(c *Core, dc *dexConnection, msg *msgjson.Message) 
 }
 
 // refreshServerConfig fetches and replaces server configuration data. It also
-// initially checks that a server's API version is one of apiVers.
-func (dc *dexConnection) refreshServerConfig(apiVers []int) error {
+// initially checks that a server's API version is one of serverAPIVers.
+func (dc *dexConnection) refreshServerConfig() error {
 	// Fetch the updated DEX configuration.
 	cfg := new(msgjson.ConfigResult)
 	err := sendRequest(dc.WsConn, msgjson.ConfigRoute, nil, cfg, DefaultResponseTimeout)
@@ -609,14 +609,8 @@ func (dc *dexConnection) refreshServerConfig(apiVers []int) error {
 	cfgAPIVer := int32(cfg.APIVersion)
 
 	if apiVer != cfgAPIVer {
-		// If not initiation, do not allow api changes
-		// between connections.
-		if apiVer != -1 {
-			return fmt.Errorf("server API version unexpectedly changed from %v to %v",
-				apiVer, cfgAPIVer)
-		}
 		if found := func() bool {
-			for _, version := range apiVers {
+			for _, version := range serverAPIVers {
 				ver := int32(version)
 				if cfgAPIVer == ver {
 					dc.log.Debugf("Setting server api version to %v.", ver)
--- a/client/core/bookie.go
+++ b/client/core/bookie.go
@@ -21,6 +21,8 @@ import (
 var (
 	feederID        uint32
 	bookFeedTimeout = time.Minute
+
+	outdatedClientErr = errors.New("outdated client")
 )
 
 // BookFeed manages a channel for receiving order book updates. The only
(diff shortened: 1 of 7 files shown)


JoeGruffins:
Will add some more tests if the idea looks ok.

chappjc:
Thanks, @JoeGruffins.

The plan we decided on was to introduce the API version and have it communicated asap, but keeping it at v0 until dcrdex 1.0 or SPV (probably same release).  v0 because it's not stable, but still communicated and checked so when a major bump to v1 happens the client sees that they're going to need to update to a client that knows the v1 API.

Of course we may have to bump 0->1 sooner than SPV if there are changes we can't solve with clever backward compatibility, but I think it's best if we keep it v0 to indicate that there are no promises yet.  In my mind the necessity with this PR is having the version checking machinery in place well in advance of such major breaking version bumps.

client/core/core.go
```diff
@@ -64,6 +65,9 @@ var (
 	// When waiting for a wallet to sync, a SyncStatus check will be performed
 	// ever syncTickerPeriod. var instead of const for testing purposes.
 	syncTickerPeriod = 10 * time.Second
+	// serverAPIVers are the DEX server API versions this client is capable
+	// of communicating with.
+	serverAPIVers = []int{serverdex.InitialAPIVersion}
```
chappjc, line 70: This makes sense and I'm OK with using a `server/dex` import for this now, but I suspect we're going to want to break the tight coupling of client -> server packages at some point.

Probably we're just discovering things (like server/account) that would be better in the "common"  `decred.org/dcrdex/dex` packages at some point.

server/dex/dex.go
```diff
@@ -31,6 +31,14 @@ import (
 	"github.com/decred/dcrd/dcrec/secp256k1/v3/ecdsa"
 )
 
+const (
+	// InitialAPIVersion is the initial dex server API version.
+	InitialAPIVersion = iota
+
+	// APIVersion is the current API version.
+	APIVersion = InitialAPIVersion
```
chappjc, line 39: Good to start at v0 as you have now.

chappjc (commented):
Looks like a good start.  Having clients looking at the version starting with 0.2 is important.

The other more fine grained versioning we'll need is for each asset.  This PR is the version for the comms protocol and maybe some other agreed upon things, but I believe we'll need to version each asset for when things with an asset are changed that don't affect the dex API (e.g. kind of addresses, transaction structure, coin ID encoding, contract address, lock times, encoding of txdata payload, etc.)... things that the concern a particular asset backend rather than the higher level core packages or comms.

JoeGruffins:
> The other more fine grained versioning we'll need is for each asset.

Maybe something similar to this but a map of asset id to api versions in the config response?

client/core/core.go
```diff
@@ -4406,8 +4414,11 @@ func (c *Core) connectDEX(acctInfo *db.AccountInfo, temporary ...bool) (*dexConn
...
 
 	// Request the market configuration.
-	err = dc.refreshServerConfig() // handleReconnect must too
+	err = dc.refreshServerConfig(serverAPIVers) // handleReconnect must too
 	if err != nil {
+		if errors.Is(err, outdatedClientErr) {
+			sendOutdatedClientNotification(c, dc)
+		}
 		return dc, err
 	}
```
JoeGruffins, line 4423: Might be less awkward if just do the api check in a separate function, although it makes sense to me that the API version is the first thing checked in the response and a quick fail is best.

client/core/core_test.go
```diff
@@ -6455,3 +6455,74 @@ func TestPreOrder(t *testing.T) {
 		t.Fatalf("PreOrder error after fixing everything: %v", err)
 	}
 }
+
+func TestRefreshServerConfig(t *testing.T) {
```
JoeGruffins, line 6459: go test -run="TestRefreshServerConfig" -coverprofile=cov.out && go tool cover -func=cov.out  | grep -E "refreshServerConfig"
PASS
coverage: 1.5% of statements
ok      decred.org/dcrdex/client/core   0.081s
decred.org/dcrdex/client/core/bookie.go:599:            refreshServerConfig                     100.0%

client/core/bookie.go
```diff
@@ -591,14 +594,46 @@ func handleTradeResumptionMsg(c *Core, dc *dexConnection, msg *msgjson.Message)
...
 
+	// Check that we are able to communicate with this DEX.
+	apiVer := atomic.LoadInt32(&dc.apiVer)
+	cfgAPIVer := int32(cfg.APIVersion)
+
+	if apiVer != cfgAPIVer {
+		// If not initiation, do not allow api changes
+		// between connections.
+		if apiVer != -1 {
+			return fmt.Errorf("server API version unexpectedly changed from %v to %v",
```
chappjc, line 615: I've been thinking about this, and this seems to be inevitable, so I don't think we have a choice but to go along with the new version if it's recognized.  A server update that changes API version _will_ happen and I don't think we should always torpedo connected clients of they recognize the new api version.

client/core/bookie.go
```diff
@@ -591,14 +594,46 @@ func handleTradeResumptionMsg(c *Core, dc *dexConnection, msg *msgjson.Message)
 	return nil
 }
 
-func (dc *dexConnection) refreshServerConfig() error {
+// refreshServerConfig fetches and replaces server configuration data. It also
+// initially checks that a server's API version is one of apiVers.
+func (dc *dexConnection) refreshServerConfig(apiVers []int) error {
```
chappjc, line 599: Why not use the package-level var `serverAPIVers` instead of an arg?

chappjc:
> > The other more fine grained versioning we'll need is for each asset.
> 
> Maybe something similar to this but a map of asset id to api versions in the config response?

Yeah, that would be appropriate.  I think we could put a version in the `dex/msgjson.Asset` struct.

client/core/bookie.go
JoeGruffins: I agree. I just think it opens some angles possibly, that are hard to notice. Will remove, and we'll just be careful when updating.

client/core/bookie.go
JoeGruffins: For testing...
If we use the package level variable, it makes testing in parallel a bit difficult, although doable. Will see what it looks like, I guess will need to add a lock to the test harness.

chappjc: Ah, I see.  I don't think there will be any concurrency issues though. I think you can just change the slice and and call `refreshServerConfig` for each test case.  But if for some reason something is a problem with that, I don't think you even need a running tCore.  That is, you could manually put together a `dexConnection` either with `testDexConnection` or manually with just what's needed to test this method.

client/core/bookie.go
chappjc: K. A warning would probably be good though.

JoeGruffins: Added.

client/core/bookie.go
JoeGruffins: Just using the top numbers of a uint16 which are probably not going to ever be actual versions.

chappjc:
> > > The other more fine grained versioning we'll need is for each asset.
> > 
> > 
> > Maybe something similar to this but a map of asset id to api versions in the config response?
> 
> Yeah, that would be appropriate. I think we could put a version in the `dex/msgjson.Asset` struct.

@buck54321 any objections to this?

In short these versions would be for things with an asset are changed that don't affect the dex API (e.g. kind of addresses, transaction structure, coin ID encoding, contract address, lock times, encoding of txdata payload, etc.)... things that concern a particular asset backend rather than the higher level core packages or comms.

Checking the asset version at order placement time for both base and quote seems right to me for that.

buck54321:
> @buck54321 any objections to this?
> 
> In short these versions would be for things with an asset are changed that don't affect the dex API (e.g. kind of addresses, transaction structure, coin ID encoding, contract address, lock times, encoding of txdata payload, etc.)... things that concern a particular asset backend rather than the higher level core packages or comms.
> 
> Checking the asset version at order placement time for both base and quote seems right to me for that.

Sounds about right, though it's not readily apparent to me how an upgrade would work. Also not quite clear what to do at the UI level when we know that a server's asset version is incompatible with the client backend. I guess these details will get worked out in time. I think the important bit is to convey the info now so we're not trying to coordinate this in a patch release or something dumb like that.

chappjc:
Yeah, I think at a minimum it's a seatbelt that says "to trade with this asset at X you gotta update to version Y", to prevent making swaps or whatever that will be rejected (or worse with an old version that possibly had a security issue).

chappjc:
I'm doing the asset versioning @JoeGruffins... shoulda mentioned.

Issue #1036, opened 2021-04-05
Updated password not persisted


buck54321:
from @alexlyp on Element:

> so something i noticed during a recent test run i was working on. I have my dcrwallet fully registered and working with dexc. i then update the private passphrase of the wallet. close everything down then restart. dexc appears to connect with wallet as expected after login, and then any attempt to buy/sell/unlock it correctly errors due to not the correct passphrase. then within the dexc/site i update the wallet private pass and i'm able to do everything from there. but when i restart everything, it's like dexc doesn't save that updated wallet passphrase and complains about incorrect pass
guessing it should be easily reproducible via CLI

amassarwi:
Will look into it.

chappjc:
Thanks for looking into it @amassarwi, but before that, pls hit up https://github.com/decred/dcrdex/pull/1016

amassarwi:
Sure thing on it now

amassarwi:
from @alexlyp on Element:
> ah turns out i was just doing something incorrect in my code

I guess we can close this one ?

chappjc:
Hmm, possibly.  I thought that comment was with regard to something else.  Can we first verify that the wallet password can be changed, dexc shutdown, and it uses the new password?

alexlyp:
That was in reference to the "empty vs nil passphrase" discussion we were having yesterday.  This issue should definitely be reproducible.

amassarwi:
oh i see, looking into it

alexlyp:
It should be easily reproducible by doing the following.  Start dexc, login and connect to existing dcrwallet.  Change the dcrwallet private passphrase then attempt to place an order.  It should error due to incorrect passphrase.  Then go to the wallets page and click the gears to update the passphrase.  Then you should be able to unlock the wallet.  Then shut everything down then restart dexc, login and attempt to place an order.  Should error for bad pass. 

That is the steps I have taken, only difference being it's logging in via decrediton (but it's still using the same internal passphrase update in the site/ )

PR #1048, opened 2021-04-12
client: fix db store of enc pass in ReconfigureWallet

chappjc:
When `ReconfigureWallet` is used to change the wallet's passphrase, there
are two DB updates, but the second one was using the original
`EncryptedPW`.  The resolution is to simply update the `EncryptedPW` field
of the client/db.Wallet instance passed to `db.UpdateWallet` in
`ReconfigureWallet` after updating the wallet balance. While two DB writes
is inefficient, this still uses the `(*Core).setWalletPassword` method as is
since it also checks the password and emits a notification.

This also revises `(*BoltDB).SetWalletPassword` so that it does not bother
to decode the current wallet's `db.Balance` since only the wallet encoding
bucket (`walletKey`) is updated, not the balance bucket (`balanceKey`).
A deep copy of the new encrypted PW is also done out of caution.

Resolves https://github.com/decred/dcrdex/issues/1036

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1896,6 +1896,9 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg 
 			wallet.Disconnect()
 			return err
 		}
+		// Update dbWallet so db.UpdateWallet below reflects the new password.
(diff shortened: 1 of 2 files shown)


JoeGruffins (approved):
Can confirm that this fixes the issue.

PR #1051, opened 2021-04-15
client/{core,webserver}: start internationalization

buck54321:
Begins the work for internationalization. See also #1127.

- HTML templates are now pre-compiled into language-specific versions in their own directories. This replaces any static strings in the template with localized variants. Build the **client/webserver/site/template-builder** program and run from **client/webserver/site** to build the localized templates. 
- Core notifications are formatted using a `message.Printer` from **golang.org/x/text/message**. Translators are tasked with registering variants using `registerTranslations` via a function called in `init`.
- `Doc.formatCoinValue` uses a localization-friendly `Intl` formatter now. Some numbers may display differently than before, but in normal use, nobody should notice. 

The localized html needs to be built and checked in when changes are made. Run `go generate` from **client/webserver/site** to rebuild.


client/core/locale_en.go
```diff
@@ -0,0 +1,143 @@
...
+// TemplateKeys. Other translators will define each entry in a struct literal.
+var EnLocale = map[string]*Translation{}
+
+var Subjects map[string]map[string]string
+
+func registerTranslations(lang language.Tag, translations map[string]*Translation) {
+	for subject, t := range translations {
+		tmplKey := TemplateKeys[subject]
+		message.SetString(lang, tmplKey, t.Template)
+		message.SetString(lang, subject, t.Subject)
```
buck54321, line 127: We can use `message.Set` and vary templates based on plurality too.

client/webserver/site/src/js/doc.js
```diff
@@ -14,13 +14,12 @@ const BipIDs = {
...
 const BipSymbols = Object.values(BipIDs)
 
-// Parameters for printing asset values.
-const fullPrecisionSpecs = {
-  minimumSignificantDigits: 4,
-  maximumSignificantDigits: 8,
-  minimumFractionDigits: 8,
+const intFormatter = new Intl.NumberFormat(navigator.languages)
+
+const decimalFormatter = new Intl.NumberFormat(navigator.languages, {
```
buck54321, line 19: `navigator.languages` should maybe come from a cookie based on the dexc config setting instead.

client/webserver/site/template-builder/main.go
```diff
@@ -0,0 +1,96 @@
...
+	"os"
+	"path/filepath"
+	"regexp"
+
+	"decred.org/dcrdex/client/webserver/locales"
+)
+
+var (
+	workingDirectory, _ = os.Getwd()
+	keyRegExp           = regexp.MustCompile(`\[\[\[([a-zA-Z0-9 _.-]+)\]\]\]`)
```
buck54321, line 17: Hmm. Maybe `\[\[\[([^\[\]]+)\]\]\]` instead.

buck54321:
I thought maybe people would have comments on implementation, but I'll just assume that everyone loves it and keep moving.

chappjc:
It's da bomb.

client/webserver/locales/en-us.go
```diff
@@ -0,0 +1,157 @@
+package locales
+
+var EnUS = map[string]string{
```
JoeGruffins, line 3: In my opinion these need to be ordered. Take for example https://github.com/decred/decrediton/blob/master/app/i18n/translations/original.json which is ordered by identifier or even  https://github.com/planetdecred/dcrandroid/blob/master/app/src/main/res/values/strings.xml which is ordered by section.

Also, keys are all over the place. I like lower case underscored keys... But anything uniform would be good.

JoeGruffins (commented):
Looks good to me. Was unable to run it.
<details closed>
<summary>panic</summary>

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0xad55cd]

goroutine 1 [running]:
decred.org/dcrdex/client/webserver.(*WebServer).buildTemplates.func1(0xc00002e0c0, 0x3d, 0x002e0c0)
        /home/joe/git/dcrdex/client/webserver/webserver.go:327 +0x6d
path/filepath.WalkDir(0xc00002e0c0, 0x3d, 0xc00054f998, 0xc00002e0c0, 0x3d)
        /home/joe/.local/go/src/path/filepath/path.go:463 +0x8b
decred.org/dcrdex/client/webserver.(*WebServer).buildTemplates(0xc000890000, 0xc6985b, 0x5, 
        /home/joe/git/dcrdex/client/webserver/webserver.go:326 +0x1e5
decred.org/dcrdex/client/webserver.New(0xc00054fe48, 0xc6810b, 0x3, 0xeb3028)
        /home/joe/git/dcrdex/client/webserver/webserver.go:204 +0x85b
main.mainCore(0x0, 0x0)
        /home/joe/git/dcrdex/client/cmd/dexc/main.go:153 +0x80b
main.main()
        /home/joe/git/dcrdex/client/cmd/dexc/main.go:29 +0x26
```
</details>

client/webserver/locales/en-us.go
buck54321: I can organize by template.

Using the English text itself as the key has the benefit that the templates can still be read without a lookup. I've only shortened the key when the text is really long.

buck54321: I'm not opposed to going to all custom keys if there's a good reason, but I just wanted to point out the benefit of doing it this way.

.gitignore
```diff
@@ -24,3 +24,4 @@ docs/examples/rpcclient/rpcclient
 wiki
 dex/testing/loadbot/loadbot
 bin/
+localized_html/
```
chappjc, line 27: Why not check these in?  Would be nice to have one less thing in our workflow.

client/webserver/webserver.go
```diff
@@ -308,6 +316,56 @@ func New(core clientCore, addr, customSiteDir string, logger dex.Logger, reloadH
...
+			continue
+		}
+		if fi.Name() == lang {
+			match = fi.Name()
+			break
+		}
+
+		tag, err := language.Parse(fi.Name())
+		if err != nil {
+			log.Debugf("error parsing language tag %q: %v", fi.Name(), err)
```
chappjc, line 342: Warnf? or does this happen routinely with random files in that folder?

client/webserver/webserver.go
```diff
@@ -308,6 +316,56 @@ func New(core clientCore, addr, customSiteDir string, logger dex.Logger, reloadH
...
+			continue
+		}
+		langs = append(langs, tag)
+		dirs = append(dirs, fi.Name())
+	}
+
+	if match == "" {
+		matcher := language.NewMatcher(langs)
+		_, idx := language.MatchStrings(matcher, lang)
```
chappjc, line 351: `MatchStrings` is a bit confusing but it says if no match then "default language is returned if no other language matched.", which is odd (how do we know we didn't pick the default?).

Maybe we should have this be an error, and print the `langs` slice in the event of no match.  This would require cannibalizing `language.MatchStrings` like:

```go
if match == "" {
	// Try to identify candate languages.
	acceptLangs, _, err := language.ParseAcceptLanguage(lang)
	if err != nil {
		return fmt.Errorf("unable to parse requested language: %v", err)
	}
	// Match against template languages.
	matcher := language.NewMatcher(langs)
	tag, idx, conf := matcher.Match(acceptLangs...)
	switch conf {
	case language.Exact:
	case language.High, language.Low:
		log.Infof("Using language %v", tag)
	case language.No:
		return fmt.Errorf("no match for %q in recognized languages %v", lang, langs)
	}
	match = dirs[idx]
}
``` 

Which is just as well because we're only testing for one lang.

run_tests.sh
```diff
@@ -8,6 +8,12 @@ modules="."
 GV=$(go version | sed "s/^.*go\([0-9.]*\).*/\1/")
 echo "Go version: $GV"
 
+cd client/webserver/site/template-builder
+go build
+cd ../
+template-builder/template-builder
```
chappjc, line 14:
```sh
cd client/webserver/site
go run ./template-builder/main.go  # or go generate
```

go.mod
```diff
@@ -37,6 +37,7 @@ require (
 	golang.org/x/crypto v0.0.0-20210220033148-5ea612d1eb83
 	golang.org/x/sync v0.0.0-20200625203802-6e8e738ad208
 	golang.org/x/term v0.0.0-20210220032956-6a3ed077a48d
+	golang.org/x/text v0.3.4
```
chappjc, line 40: v0.3.6 or even at 3115f89c4b99a620c7f1a4395a2b4405e95b82b6

client/core/trade_simnet_test.go
```diff
@@ -758,9 +758,10 @@ func TestResendPendingRequests(t *testing.T) {
 		for !foundSwapErrorNote {
 			select {
 			case note := <-notes:
-				foundSwapErrorNote = note.Severity() == db.ErrorLevel && note.Subject() == SubjectSwapError
+				foundSwapErrorNote = note.Severity() == db.ErrorLevel && (note.Subject() == SubjectSwapSendError ||
+					note.Subject() == SubjectInitError || note.Subject() == SubjectReportRedeemError)
 			case <-time.After(4 * time.Second):
-				return fmt.Errorf("client %d: no init/redeem error note after 4 seconds", client.id)
+				return fmt.Errorf("client %d: no init/redeem error note after 1 second", client.id)
```
chappjc, line 764: It does have a 4 second timeout

chappjc (changes requested):
Looking pretty nice

client/core/trade_simnet_test.go
buck54321: Whoops. Rebase conflict resolution mistake.

.gitignore
buck54321: Done. Now I'm thinking about how we can check these in automated testing.

client/webserver/site/template-builder/main.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+func main() {
```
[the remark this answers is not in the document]
buck54321: I'm going back and forth on this. I also maybe think that we should just move it to **client/cmd**.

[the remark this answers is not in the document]
chappjc: I think it's correct under client/webserver/site somewhere because there's no need to actually build a CLI binary executable, and it would only apply in this folder.  This makes the most sense to me as a build operation (not a cmd app) executed wither via `go run ./template-builder/main.go` or `go generate`

chappjc:
Looks like all review items are resolved, but did you have a local `client/webserver/site/gen.go` file to check in to enable the `go generate command` as described in https://github.com/decred/dcrdex/pull/1051#discussion_r683575368?

Can pls resolve conflicts for final review.  Looks like conflicts with the app password cache PR merged, so... more strings, sorry. :(

client/webserver/webserver.go
```diff
@@ -164,7 +177,7 @@ func New(core clientCore, addr, customSiteDir string, logger dex.Logger, reloadH
 	var siteDir string
 	absDir, _ := filepath.Abs("site")
 	for _, dir := range []string{
-		customSiteDir,
+		cfg.CustomSiteDir,
```
chappjc, line 180: Another possible follow-up: in the loop body, do a `continue` if `dir == ""`, as is usually the case with the default `CustomSiteDir`.  Otherwise we see something confusing like this:

```
[DBG] WEB: Looking for site in 
[DBG] WEB: Looking for site in /home/user/github/decred/dcrdex/client/cmd/dexc/site
[DBG] WEB: Looking for site in /home/user/github/decred/dcrdex/client/cmd/dexc/site
[DBG] WEB: Looking for site in /home/user/github/decred/dcrdex/client/webserver/site
[INF] WEB: Located "site" folder at /home/user/github/decred/dcrdex/client/webserver/site
```

chappjc (approved):

Issue #1049, opened 2021-04-14
client/core: Register should verify there is no account bucket for the DEX


chappjc:
Although the register dialog bug that showed up when the connection to the dex was down on client startup was resolved in https://github.com/decred/dcrdex/pull/989 and the `Core.conns` map should now include all configure DEXs even if they are not reachable, we should add some additional sanity checks to `(*Core).Register`.  Namely, before attempting a `wallet.PayFee`, the client DB should be consulted for presence of an account for the given DEX.

ukane-philemon:
@chappjc was this addressed in #1015 ?

PR #1052, opened 2021-04-15
client/core: Register should check DB for an account

chappjc:
Resolves https://github.com/decred/dcrdex/issues/1049


chappjc:
Closing for now.  This can be absorbed in upcoming client registration work in either #1015 or #1120

PR #1054, opened 2021-04-16
asset versioning

chappjc:


client/asset/dcr/dcr.go
```diff
@@ -892,6 +895,10 @@ func orderEnough(val, lots uint64, nfo *dex.Asset) func(sum uint64, size uint32,
 // dex.Bytes should be appended to the redeem scripts collection for coins with
 // no redeem script.
 func (dcr *ExchangeWallet) FundOrder(ord *asset.Order) (asset.Coins, []dex.Bytes, error) {
+	if ord.DEXConfig.Version != dcr.Info().Version {
```
buck54321, line 898: Server provides version to client, so client is tasked with version enforcement. So if **server upgrades first**, the existing client backends are immediately useless and client will have to upgrade **dexc** to trade that asset again. On the other hand, if we engineer **client upgrades before server** to mitigate this problem, then clients need to be backwards compatible to remain viable until the server upgrade.

chappjc: Depending on the nature of the change that bumps an asset version, I don't think backward compatibility can be ensured at all.  Something fundamental about coin id encoding or contract structure may change that will cause one party to fail to interpret or validate data from the other party.  I think an asset version update should really be reserved for incompatible changes, as a signal to when it's really a bad idea to try to place an order at all without updating (or allowing server to update).

chappjc: I also commented this bit in `FundOrder` because it was more of a note.  The main point of checking is in `(*Core).walletSet` which intercepts not only new orders in `prepareTrackedTrade` and `(*Core).PreOrder` but order resumption in `(*Core).resumeTrades`.  I'm not totally certain if this is best TBH, but with the interpretation of version mismatch as a serious (breaking) incompatibility at some level, resuming and thus making new transactions that are likely to fail validation would not be a good idea.

chappjc: Been mulling over clients supporting multiple versions (client upgrades before server), and I can see an approach using versioned asset packages and the client starts a backend for each one:

```go
import (
    dcrv1 "decred.org/dcrdex/client/asset/dcr"
    dcrv2 "decred.org/dcrdex/client/asset/dcr/v2"
)
```

```
    // in constructor
    c.wallets[42][1] = dcrv1.Setup(...) // v1 DCR
    c.wallets[42][2] = dcrv2.Setup(...) // v2 DCR
```

```
    backend := c.wallets[42][dexDCRVer]
    ord := backend.FundOrder(...)
    t := trackedTrade{ord, dexDCRVer}
```

This doesn't need to be done right away, but it does suggest we should start the asset versions at 1 right now instead of 0.

buck54321: Does each versioned wallet create a separate connection? If yes, won't they need separate accounts (ultimately, separate configurations) in order to handle locking, unlocking, lockunspent, etc. without interfering with each other? If, on the other hand, they can share an underlying connection and account (my preference), then this pattern isn't a great match, and I'll offer the following alternative for consideration.

We can change the interface to separate the idea of a `Wallet` into something like a `MasterWallet`, which will handle the configuration, connection, internal caches, etc., and a `TradingWallet` which is bound to a `*dex.Asset` and ultimately the asset version. How that is handled internally may end up being similar to your suggested directory pattern, but we'd have more flexibility to handle upgrades appropriately. This pattern could even be adapted to handle the first scenario above, where there are separate RPC connections.

To expand, `Core` would control the `MasterWallet` = connection and lock state, and the `*dexConnection` would control the `TradingWallet`, retrieved though a method something like `(MasterWallet).DEXWallet(*dex.Asset) TradingWallet`. The only thing `Core` would need to do then is check that the version is supported so that we can adjust the UI appropriately and signal the (now DEX-specific) `SupportedAsset`s. Version support can be signaled in the `*asset.WalletInfo`. I think this would be largely compatible with the existing client database structure, but an upgrade to store the version used with each order would probably be needed regardless of the approach used.

chappjc: I think that makes sense for when we need two versions.  There would indeed be potential for racing to lock coins between two instances sharing the same actual wallet.

Just planning ahead right?  Are we good to just start sending a version (0) and checking it client-side for now?

chappjc:
Tested a bit.

When a server does go to v1 for an asset, clients built with this checking will stop working with that server.

The plan for client to support multiple asset versions as described in https://github.com/decred/dcrdex/pull/1054#discussion_r616266089 can be implemented when we have a v1 to use.

client/asset/dcr/dcr.go
buck54321: > Just planning ahead right? Are we good to just start sending a version (0) and checking it client-side for now?

Yes. My vote is to just do version 0 for now.

PR #1055, opened 2021-04-19
core: Check can trade with new wallet.

### LINKED ISSUE #968: ensure new wallet can process existing orders when connecting
When doing ReconfigureWallet, we should verify that the new wallet can settle any existing swaps, and since we don't support changing funding coins yet, any active orders.

> Should really check if this is the same wallet and user has active trades. If both are yes error, can't use a different wallet with active trades.
> 
> We have the old wallet's deposit address, so maybe a simple `isMine` method on wallets would be good enough?
> 
> Also, after checking we will know if we should disconnect here or not.

_Originally posted by @JoeGruffins in https://github.com/decred/dcrdex/pull/928#discussion_r571744313_

JoeGruffins:
If a user has active trades, they must be settled by a wallet that can
complete those trades. Disallow changing to a new wallet that cannot do
this.

closes #968

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -1687,6 +1687,15 @@ func (btc *ExchangeWallet) AuditContract(coinID, contract, txData dex.Bytes) (*a
 	}, nil
 }
 
+// RefundAddress extracts and returns the refund address from a contract.
+func (btc *ExchangeWallet) RefundAddress(contract dex.Bytes) (string, error) {
+	sender, _, _, _, err := dexbtc.ExtractSwapDetails(contract, btc.segwit, btc.chainParams)
+	if err != nil {
+		return "", fmt.Errorf("error extracting refund address: %w", err)
+	}
+	return sender.String(), nil
+}
+
 // LocktimeExpired returns true if the specified contract's locktime has
 // expired, making it possible to issue a Refund.
 func (btc *ExchangeWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time, error) {
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1899,6 +1899,33 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg 
 		return err
 	}
 
+	// If there are active trades, make sure they can be settled by the
+	// keys held within the new wallet.
+	sameWallet := func() error {
+		for _, dc := range c.dexConnections() {
+			for _, trade := range dc.trackedTrades() {
+				if trade.isActive() {
+					addr := trade.Trade().Address
+					owns, err := wallet.OwnsAddress(addr)
+					if err != nil {
+						return err
+					}
+					if !owns {
+						return fmt.Errorf("new wallet does not own address found in active trades: %v", addr)
+					}
+					// Assume all trades are owned by the
+					// new wallet if one is.
+					return nil
+				}
+			}
+		}
+		return nil
+	}
+	if err := sameWallet(); err != nil {
+		wallet.Disconnect()
+		return newError(walletErr, "new wallet cannot be used with current active trades: %v", err)
+	}
+
 	// If newWalletPW is non-nil, update the wallet's password.
 	if newWalletPW != nil { // includes empty non-nil slice
 		err = c.setWalletPassword(wallet, newWalletPW, crypter)
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1913,8 +1913,8 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg 
 					if !owns {
 						return fmt.Errorf("new wallet does not own address found in active trades: %v", addr)
 					}
-					// Assume all trades are owned by the
-					// new wallet if one is.
+					// Assume all trade addresses are owned
+					// by the new wallet if one is.
 					return nil
 				}
 			}
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1899,58 +1899,69 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg 
 		return err
 	}
 
-	ownsAddr := func(addr string) error {
-		owns, err := wallet.OwnsAddress(addr)
-		if err != nil {
-			return err
-		}
-		if !owns {
-			return fmt.Errorf("new wallet does not own address found in active trades: %v", addr)
-		}
-		return nil
-	}
-
 	// If there are active trades, make sure they can be settled by the
 	// keys held within the new wallet.
 	sameWallet := func() error {
+		ownsAddr := func(addr string) error {
+			owns, err := wallet.OwnsAddress(addr)
+			if err != nil {
+				return err
+			}
+			if !owns {
+				return fmt.Errorf("new wallet does not own address found in active trades: %v", addr)
+			}
+			return nil
+		}
 		for _, dc := range c.dexConnections() {
+			maybeDifferentWallet := false
 			for _, trade := range dc.trackedTrades() {
-				if trade.isActive() {
-					t := trade.Trade()
-					waID := wallet.AssetID
-					// If the to asset, check if we own an
-					// contract.Address.
-					if (t.Sell && trade.Quote() == waID) ||
-						(!t.Sell && trade.Base() == waID) {
-						if err := ownsAddr(t.Address); err != nil {
+				if !trade.isActive() {
+					continue
+				}
+				waID := wallet.AssetID
+				// If the to asset, check if we own an
+				// contract.Address.
+				if trade.wallets.toAsset.ID == waID {
+					if err := ownsAddr(trade.Trade().SwapAddress()); err != nil {
+						return err
+					}
+					// Assume all trade addresses are
+					// owned by the new wallet if one is.
+					return nil
+				}
+				// If the from asset, check if we own a
+				// refund address for a match if any exist.
+				if trade.wallets.fromAsset.ID == waID {
+					for _, match := range trade.matches {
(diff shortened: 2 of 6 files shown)


client/core/core_test.go
```diff
@@ -5216,7 +5227,25 @@ func TestReconfigureWallet(t *testing.T) {
...
+	if !errorHasCode(err, walletErr) {
+		t.Fatalf("wrong error when expecting auth error: %v", err)
+	}
+	tXyzWallet.ownsAddressErr = nil
+
+	// Wallet doesn't own address.
+	tXyzWallet.ownsAddress = false
+	err = tCore.ReconfigureWallet(tPW, nil, assetID, newSettings)
+	if !errorHasCode(err, walletErr) {
+		t.Fatalf("wrong error when expecting auth error: %v", err)
```
JoeGruffins, line 5246: auth error -> address owns

client/core/core.go
```diff
@@ -1899,6 +1899,33 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg
...
+					addr := trade.Trade().Address
+					owns, err := wallet.OwnsAddress(addr)
+					if err != nil {
+						return err
+					}
+					if !owns {
+						return fmt.Errorf("new wallet does not own address found in active trades: %v", addr)
+					}
+					// Assume all trades are owned by the
+					// new wallet if one is.
```
JoeGruffins, line 1917: trade **addresses** are owned

client/core/core.go
```diff
@@ -1899,6 +1899,33 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg
...
 	}
 
+	// If there are active trades, make sure they can be settled by the
+	// keys held within the new wallet.
+	sameWallet := func() error {
+		for _, dc := range c.dexConnections() {
+			for _, trade := range dc.trackedTrades() {
+				if trade.isActive() {
+					addr := trade.Trade().Address
+					owns, err := wallet.OwnsAddress(addr)
```
buck54321, line 1909: This is only the right address for the "from asset" (quote asset + buy order, or base asset + sell order). For the other asset, you could in theory check the refund address, but `Core` doesn't store that and it's not part of the `*asset.Contract` right now. Hmm. Not sure what to do.

I have to take blame for this situation, since I declined to store the refund address in the `OrderMetadata` and even removed access to the refund address when `asset.Contract` was changed from an interface to a concrete type.

buck54321 (changes requested):

client/core/core.go
JoeGruffins: I think it's now checking refund addresses should it find any. Do I have my from/to asset terminology backwards?

JoeGruffins:
Had this strange test error once but don't think it's related to this pr, has this been seen before?
```
--- FAIL: TestReorg (1.01s)
    btc_test.go:1253: chain A block still mainchain for test 3:3 after reorg
2021-05-03 15:26:54.379 [ERR] TEST: CheckAddress error for btc : decoded address is of unknown format
2021-05-03 15:26:54.381 [ERR] TEST: CheckAddress error for btc 28Zpft83eov56iESWuPpV8XFLJ1b8gMZy7: checksum mismatch
2021-05-03 15:26:54.381 [ERR] TEST: CheckAddress error for btc 3GD2fSQxhkXDAW66i6JBwCqhLFSvhMNrtO: decoded address is of unknown format
2021-05-03 15:26:54.381 [ERR] TEST: CheckAddress error for btc 3GD2fSQx: checksum mismatch
FAIL
FAIL    decred.org/dcrdex/server/asset/btc      2.970s
```

client/core/core.go
```diff
@@ -1899,6 +1899,67 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg
...
+		return nil
+	}
+
+	// If there are active trades, make sure they can be settled by the
+	// keys held within the new wallet.
+	sameWallet := func() error {
+		for _, dc := range c.dexConnections() {
+			for _, trade := range dc.trackedTrades() {
+				if trade.isActive() {
+					t := trade.Trade()
```
buck54321, line 1919:
```Go
if !trade.isActive() {
    continue
}
```

client/asset/btc/btc.go
```diff
@@ -1687,6 +1687,15 @@ func (btc *ExchangeWallet) AuditContract(coinID, contract, txData dex.Bytes) (*a
 	}, nil
 }
 
+// RefundAddress extracts and returns the refund address from a contract.
+func (btc *ExchangeWallet) RefundAddress(contract dex.Bytes) (string, error) {
```
buck54321, line 1691: For `*BCHWallet` you'll need to convert with `RecodeCashAddress`. See e.g. `(*BCHWallet).Address()`.

client/core/core.go
```diff
@@ -1899,6 +1899,67 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg
...
+					waID := wallet.AssetID
+					// If the to asset, check if we own an
+					// contract.Address.
+					if (t.Sell && trade.Quote() == waID) ||
+						(!t.Sell && trade.Base() == waID) {
```
buck54321, line 1924:
```Go
if t.wallets.toAsset.ID == waID
```

client/core/core.go
```diff
@@ -1899,6 +1899,67 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg
...
+						return nil
+					}
+					// If the from asset, check if we own a
+					// refund address for a match if any exist.
+					if trade.fromAssetID == waID {
+						for _, match := range trade.matches {
+							script := match.MetaData.Proof.Script
+							if len(script) == 0 {
+								continue
+							}
```
buck54321, line 1939: What if the new wallet doesn't have the balance to cover the trades? What about locked outpoints?

This has the potential to get really messy, I think. I'm not sure that we can reject changes under any circumstances without creating the potential for wedged orders. There are some changes that simply cannot be made with active orders like changing the account, and some that absolutely must be allowed if we are to completely avoid wedged orders, like changing the wallet password.

buck54321 (changes requested):

client/core/core.go
JoeGruffins: Maybe once wallets are internalized, this problem goes away? Will that bee real soon? Can consider this a bandaid?

buck54321: Even when we have built-ins, remote wallets will still be an option. 

We can probably get away with a simpler filter for now and punt on a comprehensive solution. But maybe just reject any changes when the wallet is the `fromWallet` on an active match. Don't even check the refund address, just error.

JoeGruffins: > What if the new wallet doesn't have the balance to cover the trades? What about locked outpoints?

The purpose is to not allow changing to a different wallet with active trades. I think there is a change of the "toAsset" falling through if not yet matched because we don't have a refund address yet. Will the trade error out later if not enough funds?

>  changing the account

I see, we really need to check that an address belongs to an account, and not a wallet, to ensure the account hasn't changed?

> changing the wallet password.

There shouldn't be a scenario where this blocks that.

Should I work on something more complex, or can we is alright to specify this as temporary? Still guessing not an issue once wallets are internal by default.

JoeGruffins: Actually, ownsAddress does check if it's this account that owns it.

JoeGruffins: > when the wallet is the fromWallet on an active match.

The from or the to wallet? It looks like the refund address comes from the to wallet?

Is it alright to check, then error if we didn't find a refund address?

buck54321: We can maybe roll with the refund address check as you have it. Better to be over-persmissive, I guess. But note that the following situation is problematic. 

Our order matches as taker. Counterparty has broadcast. We have received `audit` request, but are waiting for confirmation before broadcasting our swap. In this case, if the user tries to reconfigure the `fromWallet`, even changing the account or wallet altogether, they will encounter no resistance, but will definitely error at swap time.

I wonder if we can reject changes when matches are specifically in `MakerSwapCast`?

JoeGruffins: Still confused with to/from, but we are talking about having no refund address yet here?

buck54321: You have it right in the code.You only check the refund address `if trade.wallets.fromAsset.ID == waID`. The refund address always comes from the `fromWallet`. 

If my thinking is right, then answering this question should help. In what situation would we have no refund address = `len(match.MetaData.Proof.Script) == 0`?

JoeGruffins:
Sorry for that rebase, commit message was wrong.

client/core/core.go
JoeGruffins: I think `no refund address = len(match.MetaData.Proof.Script) == 0` itself it always correct? But it's true from order status epoch up until match status `MakerSwapCast` for the maker and `TakerSwapCast` for the taker?

buck54321: Looks like `maybeDifferentWallet` resolves my concerns. Will retest and approve.

buck54321 (approved):

chappjc (approved):

client/core/core.go
```diff
@@ -1899,6 +1899,78 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg
...
+						// are owned by the new
+						// wallet if one is.
+						return nil
+					}
+					// If we did not find a refund address,
+					// we cannot be sure that this is the
+					// same wallet.
+					//
+					// TODO: Implement a way to check that
+					// these accounts are the same or not.
```
chappjc, line 1957: Epic indent here! :smile: I think I might have made a package level function or dexConnection method taking an xcWallet instead of a closure, but this is fine.

PR #1057, opened 2021-04-19
dcr: do not use estimatesmartfee

chappjc:
estimatesmartfee with Decred is presently (1) pointless given block
utilization, and (2) based on the bitcoin core algorithm that tends to
provide excessively high estimates for several reason including the lack
of any block fullness metric or any special handling for a backlog of
tickets in mempool that is unrelated to the fee market on account of the
20 ticket max per block.

This changes the client and server dcr asset backend to return a
static fee rate of 10 atoms/B, which is the current default relay fee.

The motivation for this change is that we have observed the smart fee
algorithm recently responding is unpredictable ways to txns that sit
too long in mempool, and inconsistently between different nodes.
Using these fee rates will create a very poor experience for users.

### DIFF
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -689,7 +689,18 @@ func (dcr *ExchangeWallet) Balance() (*asset.Balance, error) {
 	return &balance, err
 }
 
-// FeeRate returns the current optimal fee rate in atoms / byte.
+// feeRate returns the current optimal fee rate in atoms / byte.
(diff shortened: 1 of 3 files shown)


chappjc:
I believe @matheusd said that dcrlnd and dcrwallet do not use estimatesmartfee and instead the default relay fee (unless otherwise set by user).  Is it unwise to have a constant rate like this?

I'll not that server-side we can manually apply a fee scale in the unlikely event of a large attack on the Decred network with txns paying just above the min relay fee.

PR #1058, opened 2021-04-20
dcr: Use median time in LocktimeExpired.

### LINKED ISSUE #897: client/asset/dcr: contracts try to refund too soon
The contract expiry logic indicates a contract is refundable too soon, resulting in ~50 minutes of refund attempts with "not finalized" transaction errors until eventually network consensus permits the refund.

JoeGruffins:
Transactions utilizing locktimes should watch the mediantime and not the
actual time to know when they can be finalized.

closes #897

### DIFF
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -157,6 +157,7 @@ type rpcClient interface {
 	GetBestBlock(ctx context.Context) (*chainhash.Hash, int64, error)
 	GetBlockHash(ctx context.Context, blockHeight int64) (*chainhash.Hash, error)
 	GetBlockVerbose(ctx context.Context, blockHash *chainhash.Hash, verboseTx bool) (*chainjson.GetBlockVerboseResult, error)
+	GetBlockHeaderVerbose(ctx context.Context, blockHash *chainhash.Hash) (*chainjson.GetBlockHeaderVerboseResult, error)
 	GetRawMempool(ctx context.Context, txType chainjson.GetRawMempoolTxTypeCmd) ([]*chainhash.Hash, error)
 	GetRawTransactionVerbose(ctx context.Context, txHash *chainhash.Hash) (*chainjson.TxRawResult, error)
 	LockUnspent(ctx context.Context, unlock bool, ops []*wire.OutPoint) error
@@ -1629,7 +1630,43 @@ func (dcr *ExchangeWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time,
 		return false, time.Time{}, fmt.Errorf("error extracting contract locktime: %w", err)
 	}
 	contractExpiry := time.Unix(int64(locktime), 0).UTC()
-	return time.Now().UTC().After(contractExpiry), contractExpiry, nil
+	medianTime, err := dcr.calcPastMedianTime()
+	if err != nil {
+		return false, time.Time{}, fmt.Errorf("error calculating median time: %w", err)
+	}
+	return medianTime.After(contractExpiry), contractExpiry, nil
+}
+
+// calcPastMedianTime calculates the median time of the previous few blocks
+// prior to, and including, the best block.
+func (dcr *ExchangeWallet) calcPastMedianTime() (time.Time, error) {
+	dcr.tipMtx.RLock()
+	hash := dcr.currentTip.hash
+	dcr.tipMtx.RUnlock()
+
+	// Look at the last 11 blocks, which is consistent with dcrd.
+	timestamps := make([]int64, 0, 11)
+	for i := 0; i < cap(timestamps); i++ {
+		blockHeader, err := dcr.node.GetBlockHeaderVerbose(dcr.ctx, hash)
+		if err != nil {
+			return time.Time{}, err
+		}
+		timestamps = append(timestamps, blockHeader.Time)
+		if blockHeader.Height == 0 {
+			break
+		}
+		hash, err = chainhash.NewHashFromStr(blockHeader.PreviousHash)
+		if err != nil {
+			return time.Time{}, fmt.Errorf("error decoding previous hash: %w", err)
+		}
+	}
+
+	sort.Slice(timestamps, func(i, j int) bool {
+		return timestamps[i] < timestamps[j]
+	})
+
+	medianTimestamp := timestamps[len(timestamps)/2]
+	return time.Unix(medianTimestamp, 0), nil
 }
 
 // FindRedemption watches for the input that spends the specified contract
--- a/client/asset/dcr/dcr_test.go
+++ b/client/asset/dcr/dcr_test.go
@@ -8,6 +8,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"math/rand"
 	"os"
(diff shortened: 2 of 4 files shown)


chappjc:
Nice, @JoeGruffins.  I had it in my head we were doing DCR the same way as BTC but obviously not!  I'm glad this turned out to be a straightforward resolution.  We'll flag this for 0.2 if the trade_simnet_tests. are working.

client/asset/dcr/dcr.go
```diff
@@ -1629,7 +1629,43 @@ func (dcr *ExchangeWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time,
...
+}
+
+// calcPastMedianTime calculates the median time of the previous few blocks
+// prior to, and including, the best block.
+func (dcr *ExchangeWallet) calcPastMedianTime() (time.Time, error) {
+	dcr.tipMtx.RLock()
+	hash := dcr.currentTip.hash
+	dcr.tipMtx.RUnlock()
+
+	// Look at the last 11 blocks, which is consistant with dcrd.
```
chappjc, line 1646: consistant

chappjc (changes requested):
Code looks great, just the one typo.

But it does look like we need to adjust some of the trade_simnet_test.go tests.  Checking them is a little irritating because which asset is maker or taker is random because both orders are placed in the same epoch and we don't have them with deterministic preimages to fix the shuffle, but it's possible to run `TestNoTakerSwap` repeatedly until the maker (who does the refund) is the DCR contract.  Here's what I see:

```
2021-04-20 11:47:00.031 [INF] CORE:1: Broadcasted transaction with 1 swap contracts for order 0cee81eda3a45720721ffe1351d1b7bf89442b617afba7a32e678754bc711d37. Fee rate = 10. Receipts (dcr): [c8d00f1e32d9af7997b308f5a6e9fa8dbd7599506446dc327da4dcb8f2b2ae03:0]
2021-04-20 11:47:00.031 [INF] CORE:1: Contract coin c8d00f1e32d9af7997b308f5a6e9fa8dbd7599506446dc327da4dcb8f2b2ae03:0 (dcr), value = 3000000000, refundable at 2021-04-20 16:48:00 +0000 UTC (script = ...)
...
2021-04-20 11:47:01.352 [INF] CORE:1: Contract for match f82c561205cd4be2fb8653a7658f1d1b0fefcb8e44efa3584c3e17a7b63ac3de with swap coin c8d00f1e32d9af7997b308f5a6e9fa8dbd7599506446dc327da4dcb8f2b2ae03:0 (dcr) has an expiry time of 2021-04-20 16:48:00 +0000 UTC (59s), not yet expired.
...
2021-04-20 11:47:01.529 [INF] TEST: [client 1] checking if refunds are necessary
2021-04-20 11:47:01.529 [INF] TEST: [client 2] checking if refunds are necessary
2021-04-20 11:47:01.529 [INF] TEST: [client 1] Found refundable swaps worth 30.00000000 dcr and 0.00000000 btc
2021-04-20 11:47:01.529 [INF] TEST: [client 2] No refunds necessary
2021-04-20 11:47:01.529 [INF] TEST: [client 1] Waiting 58.470818677s before checking wallet balances for expected refunds
...
2021-04-20 11:48:01.392 [WRN] CORE:1: notify: |WARNING| (order) Match revoked - Match f82c5612 has been revoked - Order: 0cee81eda3a45720721ffe1351d1b7bf89442b617afba7a32e678754bc711d37
...
<a few more ticks and Checking matches, then:>
    trade_simnet_test.go:890: client 1 reported 1 unrefunded swaps after 30s
```

The logs giving an expiry time are a bit misleading since that's not wallet time where refund can happen for DCR now and never was for BTC, but the test itself needs adjusting to allow the refund to trigger.  Might need @itswisdomagain feedback.  Please see if it's a clear tweak though @JoeGruffins

client/asset/dcr/dcr.go
```diff
@@ -1629,7 +1629,43 @@ func (dcr *ExchangeWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time,
...
+	timestamps := make([]int64, 0, 11)
+	for i := 0; i < cap(timestamps); i++ {
+		block, err := dcr.node.GetBlockVerbose(dcr.ctx, hash, true)
```
buck54321, line 1649: `GetBlockVerbose` is overkill. Everything you need should be in the header, no?

chappjc: Crossed my mind too.  Could add a getblockheader to the client interface I suppose.

chappjc:
It just occurred to me that it's strange that with the old `LocktimeExpired` that the simnet test dcrd was accepting the refund tx.  Does that mean that the new one is saying `false` incorrectly or that the test was simply getting lucky with how long it was waiting or mining extra blocks?

Another funny thing about `checkAndWaitForRefunds` in trade_simnet_test.go is that I don't think it's really necessary to mine 11 blocks just to get the median of the last 11 to be after expiry.

client/asset/dcr/dcr_test.go
```diff
@@ -2221,3 +2226,110 @@ func TestPreRedeem(t *testing.T) {
...
+	defer shutdown()
+
+	var prefix string
+	for i := 0; i < 31; i++ {
+		prefix += "00"
+	}
+
+	hash, err := chainhash.NewHashFromStr(fmt.Sprintf("%v00", prefix))
+	if err != nil {
+		t.Fatalf("error creating chaing hash: %v", err)
```
JoeGruffins, line 2241: chaing

client/asset/dcr/dcr.go
```diff
@@ -1629,7 +1629,43 @@ func (dcr *ExchangeWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time,
...
+		if err != nil {
+			return time.Time{}, err
+		}
+		timestamps = append(timestamps, block.Time)
+		if block.Height == 0 {
+			break
+		}
+		hash, err = chainhash.NewHashFromStr(block.PreviousHash)
+		if err != nil {
+			return time.Time{}, fmt.Errorf("error decoding txid: %w", err)
```
JoeGruffins, line 1659: not txid

client/asset/dcr/dcr.go
JoeGruffins: consistent? Misspelling this for 20 years I guess.

JoeGruffins:
Looking at the core simnet tests.

JoeGruffins:
I have this simnet test failure a lot, but I think it is pre-existing:
```
--- FAIL: TestResendPendingRequests (69.42s)
    trade_simnet_test.go:823: [client 2] dcr balance change not in expected range 19.00000000 - 20.00000000, got 9.99996150
```
```
--- FAIL: TestTradeSuccess (71.90s)
    trade_simnet_test.go:875: [client 2] dcr balance change not in expected range 39.00000000 - 40.00000000, got 19.99996150
```

chappjc:
Thanks for revising the refund tests.  I haven't seen that balance error recently, but the tests to often see the test hang seemingly before a pass for about a minute, which is irritating.  Does the balance error occur on master for you?

chappjc:
Ugh, `TestResendPendingRequests` isn't one I usually run but there are data races for that one. Guess I'll look into those.

JoeGruffins:
> Does the balance error occur on master for you?

Yes. I was able to observer the balance failure first try just now on master.

chappjc:
OK, dang I can't reproduce any of the balance errors.  I'm sure you've got dcrd/dcrwallet on latest release-v1.6 branch with a working harness, so I'm not sure what could be going on there.  Occasionally dcrwallet has been hanging, but it doesn't sound like the problem for you... still maybe confirm it happens even with a fresh dcrd harness pls.

But I at least see why `TestTradeSuccess` intermittently hangs and have a fix.  I also see the reason for the race in `TestResendPendingRequests` but I don't have a fix for that yet.

chappjc:
@JoeGruffins Since the balance errors are reproducible for you, can you try the following diff:

```diff
@@ -983,6 +986,12 @@ func monitorTrackedTrade(ctx context.Context, client *tClient, tracker *trackedT
                for _, match := range tracker.matches {
                        side, status := match.Side, match.Status
                        if status >= finalStatus {
+                               // With async redeem request, wait for redeem sig before
+                               // declaring the match complete.
+                               if finalStatus == order.MatchComplete && len(match.MetaData.Proof.Auth.RedeemSig) == 0 {
+                                       client.log("Completed match waiting for async redeem response...")
+                                       continue
+                               }
```

I need this for `TestTradeSucces` success to not hang frequently on shutdown and I'm wondering if the root cause is the same.

JoeGruffins:
That does stop hangs! But I still get the error. I also sometimes fail with:
```
trade_simnet_test.go:811: client 2: no init/redeem error note after 1 second
```

chappjc:
I got that one too and had to increase the wait time.  Dunno what's changed with what, but it's irritating.

Another thing I noticed is that when that log in that diff is hit at all in a test, it's hit repeatedly for several seconds, and it corresponds with the balance errors.  This indicates it's waiting for server to accept it's redeem txn.  I can see in the logs the it takes a couple seconds to find it instead of immediately.  Feels like something strange with wallet or node.

There are plenty of other issues with these tests, but there's something odd with txn broadcast and locating it in mempool.

chappjc:
Regarding the race in `TestResendPendingRequests`, it kinds sickens me to suggest a change to the product code to deal, but it turns out that the following is better anyway:

```diff
@@ -1976,7 +1976,7 @@ func (t *trackedTrade) processAuditMsg(msgID uint64, audit *msgjson.Audit) error
                }
 
                // Respond to DEX, but this is not consequential.
-               err = t.dc.ack(msgID, match.MatchID, audit)
+               err = t.dc.ack(msgID, mid, audit)
```

Although I think we should reconsider `TestResendPendingRequests` and how it mutates `MatchID` to get an init request to fail.  I'm not too clear on that's being tested with this except to make sure the logic `resendPendingRequests` is OK, and I think that can be done in simpler ways.

chappjc (approved):
`TestNoMakerRedeem` and `TestNoTakerSwap` are working reliably now for DCR and BTC.

I may append a diff to this PR to address the random 2-3 sec test shutdown delay caused when a redeem request to the server takes several seconds instead of the usual couple milliseconds.   Plus a tweak for the race in `TestResendPendingRequests`.

client/asset/dcr/dcr.go
```diff
@@ -157,6 +157,7 @@ type rpcClient interface {
 	GetBestBlock(ctx context.Context) (*chainhash.Hash, int64, error)
 	GetBlockHash(ctx context.Context, blockHeight int64) (*chainhash.Hash, error)
 	GetBlockVerbose(ctx context.Context, blockHash *chainhash.Hash, verboseTx bool) (*chainjson.GetBlockVerboseResult, error)
+	GetBlockHeaderVerbose(ctx context.Context, blockHash *chainhash.Hash) (*chainjson.GetBlockHeaderVerboseResult, error)
```
buck54321, line 160: Did you need the **verbose** block header? `wire.BlockHeader` has a `TimeStamp` and a `PrevBlock`.

JoeGruffins: Hmm, did i not? In the future the median time will be included in the verbose results. I think.

chappjc: Good point @JoeGruffins.  Link to your dcrd PR so we can find it later: https://github.com/decred/dcrd/pull/2638#pullrequestreview-642063821

PR #1060, opened 2021-04-20
server: ensure all fee rates are capped by config

chappjc:
Both `BoookRouter` and `OrderRouter` use `FeeManager.LastRate` to obtain
fee rates, and these need to have the configured `MaxFeeRate` applied.
This could be done in the caller but the `FeeManager` embeds the `dex.Asset`
and thus should obey the known limits itself.

The rate from `FeeFetcher.FeeRate` is also now capped, but consumers scale
this and apply the limit after scaling anyway. `(*Market).processReadyEpoch`
already applied the `MaxFeeRate` after scale, but `(*Server).apiAsset` was
neglecting this cap and it is now applied.  e.g. the limit being applied at match-time
to a bad rate from a bad DCR `estimatesmartfee` result:

```
[DBG] SWAP: Optimal fee rate for "dcr": 630
[WRN] SWAP: Optimal fee rate 630 > max fee rate 10, using max fee rate.
```

The above check and limit was only being done for rates provided with matches.
This PR also limits to the rates returned by the `'orderbook'` and `'fee_rate'`
responses from the `BookRouter`.  Huge DCR rates were making it to clients
through these routes without the limit applied.

This also applies to the zero-conf funding coin check performed by the
`OrderRouter` in `checkZeroConf`. Previously zero-conf coins were
held to a different and potentially higher fee rate than swaps that
could exceed `MaxFeeRate`.  This would cause pre-split backed orders to
be rejected quite easily, especially given the large discrepancies we have
seen between nodes w.r.t. their `estimatesmartfee` results.

This replaces PR https://github.com/decred/dcrdex/pull/1057.

### DIFF
--- a/server/market/bookrouter.go
+++ b/server/market/bookrouter.go
@@ -558,7 +558,7 @@ func (r *BookRouter) msgOrderBook(book *msgBook) *msgjson.OrderBook {
 		MarketID:     book.name,
 		Epoch:        uint64(epochIdx),
 		Orders:       ords,
-		BaseFeeRate:  r.feeSource.LastRate(book.baseID),
+		BaseFeeRate:  r.feeSource.LastRate(book.baseID), // MaxFeeRate applied inside feeSource
 		QuoteFeeRate: r.feeSource.LastRate(book.quoteID),
 	}
 }
@@ -646,6 +646,7 @@ func (r *BookRouter) handleFeeRate(conn comms.Link, msg *msgjson.Message) *msgjs
 		}
 	}
 
+	// Note that MaxFeeRate is applied inside feeSource.
 	resp, err := msgjson.NewResponse(msg.ID, r.feeSource.LastRate(assetID), nil)
 	if err != nil {
 		log.Errorf("failed to encode fee_rate response: %v", err)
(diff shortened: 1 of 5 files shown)


server/market/bookrouter.go
```diff
@@ -558,7 +558,7 @@ func (r *BookRouter) msgOrderBook(book *msgBook) *msgjson.OrderBook {
 		MarketID:     book.name,
 		Epoch:        uint64(epochIdx),
 		Orders:       ords,
-		BaseFeeRate:  r.feeSource.LastRate(book.baseID),
+		BaseFeeRate:  r.feeSource.LastRate(book.baseID), // MaxFeeRate applied inside feeSource
 		QuoteFeeRate: r.feeSource.LastRate(book.quoteID),
```
chappjc, line 562: I couldn't find a justification for grabbing the `MaxFeeRate` and applying it in the caller (here) instead of in `LastRate`.  This is just easier.

I've documented the `(*feeFetcher).MaxFeeRate` to explain why it still exists despite being applied internally.

chappjc:
With this in, we'll tag `release-v0.2-beta` and see what shakes out before tagging the final 0.2 release.

chappjc:
I've had this PR stacked with https://github.com/decred/dcrdex/pull/1065 and https://github.com/decred/dcrdex/pull/1026 running on dex-test (testnet) for about 10 hrs now.

server/market/bookrouter.go
buck54321: Definitely cleaner this way.

buck54321 (approved):

PR #1061, opened 2021-04-20
app: abandon pre-order results if a new request is queued

### LINKED ISSUE #998: UI: schedulePreOrder maxLoaded is not a function
image

Hit randomly.  Usually don't see this, as evidenced by the successful /api/maxbuy calls prior to this error.

It seems easy to hit it by clicking around on the depth chart.

It's also quite surprising that the maxbuy endpoint takes `1300ms`.

buck54321:
Resolves #998


chappjc (approved):
Confirmed resolution for https://github.com/decred/dcrdex/issues/998

PR #1062, opened 2021-04-20
final npm updates for bugfixes and webpack 5

chappjc:
One last update of site node packages for 0.2. The updates are mostly bugfix releases, but there are a few stragglers now officially supporting webpack 5 and dropping ancient deps (hence the smaller package-lock.json).

The third commit removes `webpack-dev-server`. AFAIK, we use `npm run watch`, but there is also a (broken) `npm run start` that uses `webpack-dev-server` to launch an http server on localhost:54321.  How would this even be used?  I think `webpack --watch` started with `npm run watch` that keeps the webpack outputs updated is all we use.

Comments on the major version bumps:
- `clean-webpack-plugin` does the same thing as before (deletes `/dist/`), just [webpack 5 stuff](https://github.com/johnagan/clean-webpack-plugin/pull/195).
- `css-minimizer-webpack-plugin` also has no differences to us, just [getting on the webpack 5 bandwagon](https://github.com/webpack-contrib/css-minimizer-webpack-plugin/pull/69) and updating its cssnano dep.


chappjc:
We don't really need this for 0.2 and we definitely can't use the distraction. Closing.

PR #1066, opened 2021-04-21
build dcrdex harness bin to dextest folder

chappjc:
The dex/test/dcrdex/harness.sh script builds a dcrdex binary with short lock times for testing, but it also installs the binary to `${GOPATH}/bin` which is a little dangerous and also inconvenient. The harness would also start the binary without an absolute path, assuming that `${GOPATH}/bin` would be on the top of the user's `PATH`.  This modifies the harness to generate a dcrdex build into the `~/dextest/dcrdex`, which is also the `--appdata` folder for the harness.

### DIFF
--- a/dex/testing/dcrdex/harness.sh
+++ b/dex/testing/dcrdex/harness.sh
@@ -3,17 +3,19 @@
 
 set -e
 
-# Rebuild dcrdex with the required simnet locktime settings.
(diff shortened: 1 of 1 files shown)


JoeGruffins (approved):

buck54321 (approved):
Nice. It's always bugged me that the harness installed to GOPATH.

PR #1069, opened 2021-04-23
client/webserver: apiShutdown and /api/shutdown

### LINKED ISSUE #993: client/webserver: apiShutdown and /api/shutdown
Shutdown of dexc will currently involves ctlr+c/sigint and potentially a console-based prompt if there are active orders, however for Decrediton or another process to govern dexc (without exposing the console for this purpose), it is necessary to have a way to command the shutdown and to override the active orders check.  I propose an `/api/shutdown` endpoint and an `apiShutdown` function.

Response:

```{
    "ok": bool,
    "msg": string (optional)
}
```

With a basic/empty request payload, the handler would check for active orders, and then begin Core shutdown via context cancellation.  In this case, the response might be `{"ok": true}`.

If there were active orders, the response might be `{"ok": false, "msg": "6 active orders"}`

Then the user could deal with those orders, or force a shutdown:

Request:

```json
{
    "force": bool (optional)
}
```

amassarwi:
Closes #993.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1004,7 +1004,6 @@ type Config struct {
 // database access, match negotiation and more.
 type Core struct {
 	ctx           context.Context
-	ctxCancel     context.CancelFunc
 	wg            sync.WaitGroup
 	ready         chan struct{}
 	cfg           *Config
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1083,16 +1083,6 @@ func New(cfg *Config) (*Core, error) {
 	return core, nil
 }
 
-// RunWithCancel stores the given context cancel func, then calls Run to run
-// the core.
-func (c *Core) RunWithCancel(ctx context.Context, ctxCancel context.CancelFunc) {
-	// Store the context CancelFunc as a field, since we will need to be able
-	// to shutdown Core via context cancellation.
-	c.ctxCancel = ctxCancel
-	// Run the core.
-	c.Run(ctx)
-}
-
 // Run runs the core. Satisfies the runner.Runner interface.
 func (c *Core) Run(ctx context.Context) {
 	c.log.Infof("Starting DEX client core")
@@ -2406,23 +2396,6 @@ func (c *Core) verifyRegistrationFee(assetID uint32, dc *dexConnection, coinID [
 
 }
 
-// Shutdown shuts down the client.
-func (c *Core) Shutdown(force bool) error {
-	// We skip the active orders check if the force flag is true.
-	if !force {
-		// Check active orders
-		conns := c.dexConnections()
-		for _, dc := range conns {
-			if dc.hasActiveOrders() {
-				return fmt.Errorf("cannot shutdown with active orders")
-			}
-		}
-	}
-	// Shutdown core via context cancellation
-	c.ctxCancel()
-	return nil
-}
-
 // IsInitialized checks if the app is already initialized.
 func (c *Core) IsInitialized() (bool, error) {
 	return c.db.ValueExists(keyParamsKey)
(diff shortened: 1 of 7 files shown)


amassarwi:
`TCore` need to implement the new interface func.

client/core/core.go
```diff
@@ -1082,6 +1083,16 @@ func New(cfg *Config) (*Core, error) {
 	return core, nil
 }
 
+// RunWithCancel stores the given context cancel func, then calls Run to run
+// the core.
+func (c *Core) RunWithCancel(ctx context.Context, ctxCancel context.CancelFunc) {
```
buck54321, line 1088: If you're expecting the caller to be in charge of the calling the `CancelFunc`, then you don't need to pass it in. If you're expecting `Core` to be in charge of the `CancelFunc`, then the caller shouldn't be creating it, `Core` should create it, and again, you don't need to pass it in.

I'm wondering if we need any new methods to accomplish this. The API already has everything needed, no? The caller should just check for active orders, and cancel the `Context` they provided to `Run`. A helper method to quickly check for active orders/matches might be handy, but even that's questionable since @chappjc just ensured with #1030 that the `User` and `Exchanges` endpoints only return active orders. 

That said, we do have a `PromptShutdown` method with very similar logic as `Shutdown`, but that method has been a pain point for us in the past because it confuses the role of shutdowner between the Core and the caller, and I personally think that we should get rid of it and have the caller handle all of the shutdown logic, with no training wheels by default. I think @chappjc has expressed a strong opinion about `PromptShutdown` in the past too. Maybe he can add something here.

chappjc: I think we miscommunicated @amassarwi.  It's the webserver and rpcserver that need the CancelFunc, not Core.

amassarwi: Makes total sense, updated.

client/webserver/api.go
```diff
@@ -316,6 +316,29 @@ func (s *WebServer) apiInit(w http.ResponseWriter, r *http.Request) {
...
+		if err != nil {
+			s.writeAPIError(w, "shutdown error: %v", err)
+			return
+		}
+	}
+
+	// Cancel Core's context.
+	s.coreCancel()
+
+	writeJSON(w, simpleAck(), s.indent)
```
JoeGruffins, line 339: Not sure, but I guess we are waiting for all handlers to finish after shutdown? If not would need to acknowledge first, but probably ok.

JoeGruffins (approved):
Code looks sane to me.

client/webserver/webserver.go
```diff
@@ -261,6 +263,7 @@ func New(core clientCore, addr, customSiteDir string, logger dex.Logger, reloadH
 	mux.Route("/api", func(r chi.Router) {
 		r.Use(middleware.AllowContentType("application/json"))
 		r.Post("/init", s.apiInit)
+		r.Post("/shutdown", s.apiShutdown)
```
buck54321, line 266: Hmm. Seems like anyone can shutdown our client from anywhere at any time. 

I seriously question whether this should exist as an API endpoint at all. For the specific case of Decrediton integration, would it be better to add a function to `libdexc` `preInitMethods` to handle context cancellation? Or is there another case for a general purpose API endpoint to shut down the client?

chappjc: This certainly should be behind the auth middleware.

This is also not applicable to Decrediton integration anymore, but I don't see a fundamental problem with controlling the full lifetime of dexc programmatically.  Signals don't work because we catch them with a prompt.  A different signal could skip the prompt, but that's getting very OS specific.

chappjc: We could consider eliminating the terminal prompt as a resolution, but that makes me pretty uncomfortable too knowing how people use software.

chappjc:
Well, this was good work @amassarwi but given the security implications of potential http routing/auth errors, plus the lack of an absolute requirement given the architectural changes to Decrediton's dex interface (i.e. libdex) make this merge less likely now.  Still, I maintain that there could be a legit use for using just the http api to control the lifecycle of dexc.  For now, let's close this and consider reopening in the future.

PR #1071, opened 2021-04-27
client/db: Add archivedOrdersBucket.

### LINKED ISSUE #1039: client/db/bolt: create archived orders bucket
With large client databases, we have been seeing long startup times while it iterates through the orders bucket to find active orders.  Profiles indicate that the worst hotspots in these slow operations are all descending from a `bbolt.(*Bucket).ForEach` with most of the execution time on either `bbolt.(*Bucket).Bucket` (just looking up a sub-bucket), or `bbolt.(*Bucket).Get` (retrieving an entry from a bucket), both ultimately heating up `bbolt.(*Cursor).search`.  The `Get` is a much smaller and less frequent hot spot than `Bucket` though, so this indicates looking up a given order's *sub-bucket* is the low hanging fruit for improvement here.

Occasionally, a given machine will take extra long with a separate hot path on `bbolt.(*Cursor).Next` ... `bbolt.(*elemRef).isLeaf`.  It's not clear why this path only gets hot occasionally.

I believe we should move known completed orders into an archived orders bucket so there's no need to iterate over them on startup for eternity.  The order export function can read this bucket if needed.

JoeGruffins:
In order to make searching for active orders in the database faster,
split the order bucket into active and archived orders.

closes #1039


client/db/bolt/db.go
```diff
@@ -719,12 +768,49 @@ func decodeOrderBucket(oid []byte, oBkt *bbolt.Bucket) (*dexdb.MetaOrder, error)
...
+// be in active orders and returns the bucket.
+func updateOrderBucket(ob, aob *bbolt.Bucket, oid order.OrderID, status order.OrderStatus) (*bbolt.Bucket, error) {
+	executed := status > order.OrderStatusBooked
+	if executed {
+		if bkt := ob.Bucket(oid[:]); bkt != nil {
+			oBkt, err := aob.CreateBucket(oid[:])
+			if err != nil {
+				return nil, err
+			}
+			// NOTE: Assumes no buckets saved in bkt.
```
JoeGruffins, line 784: Safe to assume?

client/db/bolt/db.go
```diff
@@ -509,17 +530,16 @@ func (db *BoltDB) MarketOrders(dex string, base, quote uint32, n int, since uint
 func (db *BoltDB) ActiveDEXOrders(dex string) ([]*dexdb.MetaOrder, error) {
 	dexB := []byte(dex)
 	return db.filteredOrders(func(oBkt *bbolt.Bucket) bool {
-		status := oBkt.Get(statusKey)
-		return bEqual(dexB, oBkt.Get(dexKey)) && (bEqual(status, byteEpoch) || bEqual(status, byteBooked))
-	})
+		return bEqual(dexB, oBkt.Get(dexKey))
+	}, false)
```
JoeGruffins, line 534: No longer excluding unknown status. When would an order become status unknown? I assume is it unexpected?

client/db/bolt/db.go
```diff
@@ -719,12 +768,49 @@ func decodeOrderBucket(oid []byte, oBkt *bbolt.Bucket) (*dexdb.MetaOrder, error)
...
+		}
+		oBkt := aob.Bucket(oid[:])
+		if oBkt == nil {
+			return nil, fmt.Errorf("order %s not found", oid)
+		}
+		return oBkt, nil
+	}
+	if aob.Bucket(oid[:]) != nil {
+		return nil, fmt.Errorf("archived order %s set to active", oid)
+	}
```
JoeGruffins, line 799: Maybe this check is not necessary?

client/db/bolt/db_test.go
```diff
@@ -706,15 +709,15 @@ func TestOrderFilters(t *testing.T) {
 			filter: &db.OrderFilter{
 				N: orderCount,
 			},
-			expected: []int{4, 5, 3, 2, 1, 0},
+			expected: []int{4, 3, 5, 2, 1, 0},
```
JoeGruffins, line 712: active orders now come before archived orders with not filters specified. Is this ever a problem?

client/db/bolt/upgrades.go
```diff
@@ -213,6 +216,54 @@ func reloadMatchProofs(tx *bbolt.Tx) error {
...
+	return ob.ForEach(func(k, _ []byte) error {
+		oBkt := ob.Bucket(k)
+		if oBkt == nil {
+			return fmt.Errorf("order %x bucket is not a bucket", k)
+		}
+		status := order.OrderStatus(intCoder.Uint16(oBkt.Get(statusKey)))
+		if status == order.OrderStatusUnknown {
+			println(fmt.Sprintf("Encountered order with unknown status: %x", k))
+			return nil
+		}
```
JoeGruffins, line 250: Not sure of the best way to handle this.

JoeGruffins:
@chappjc 
```
(pprof) top
Showing nodes accounting for 4422.77MB, 92.01% of 4806.87MB total
Dropped 20 nodes (cum <= 24.03MB)
Showing top 10 nodes out of 51
      flat  flat%   sum%        cum   cum%
  971.35MB 20.21% 20.21%   971.35MB 20.21%  go.etcd.io/bbolt.(*node).put
  900.48MB 18.73% 38.94%   900.48MB 18.73%  go.etcd.io/bbolt.Open.func1
  707.32MB 14.71% 53.66%   707.32MB 14.71%  go.etcd.io/bbolt.(*node).dereference
  464.27MB  9.66% 63.31%   560.77MB 11.67%  go.etcd.io/bbolt.(*Bucket).openBucket
  457.02MB  9.51% 72.82%   457.02MB  9.51%  go.etcd.io/bbolt.(*Cursor).search
  349.25MB  7.27% 80.09%   349.25MB  7.27%  go.etcd.io/bbolt.(*Bucket).write
  158.64MB  3.30% 83.39%   158.64MB  3.30%  go.etcd.io/bbolt.(*node).read
  157.09MB  3.27% 86.66%   157.09MB  3.27%  decred.org/dcrdex/dex/encode.ExtractPushes
  129.37MB  2.69% 89.35%   288.01MB  5.99%  go.etcd.io/bbolt.(*Bucket).node
  127.99MB  2.66% 92.01%   856.77MB 17.82%  go.etcd.io/bbolt.(*Bucket).Bucket
```

Should I try to find ways to make the update do less? Could make an "active orders" bucket and just move the active orders, which should be a short operation. Or I could try setting up batches in another pr and do this in segments.

chappjc:
That is looking high on the utilization.

Without having looked at this code yet, I suspect they most efficient approach would be to leave things where they are and either rename that bucket (not sure if that's even a thing with bolt though), or do like you said and only move active orders.  I think I'm ok with that.

JoeGruffins:
```
(pprof) top
Showing nodes accounting for 1914.05MB, 94.99% of 2014.95MB total
Dropped 26 nodes (cum <= 10.07MB)
Showing top 10 nodes out of 44
      flat  flat%   sum%        cum   cum%
  433.25MB 21.50% 21.50%   513.76MB 25.50%  go.etcd.io/bbolt.(*Bucket).openBucket
  319.74MB 15.87% 37.37%   319.74MB 15.87%  go.etcd.io/bbolt.Open.func1
  226.88MB 11.26% 48.63%   226.88MB 11.26%  go.etcd.io/bbolt.(*node).put
  212.01MB 10.52% 59.15%   212.01MB 10.52%  go.etcd.io/bbolt.(*Cursor).search
  163.09MB  8.09% 67.25%   163.09MB  8.09%  decred.org/dcrdex/dex/encode.ExtractPushes
  130.61MB  6.48% 73.73%   130.61MB  6.48%  go.etcd.io/bbolt.(*node).read
  117.59MB  5.84% 79.56%   117.59MB  5.84%  go.etcd.io/bbolt.(*Bucket).write
  115.32MB  5.72% 85.29%   783.09MB 38.86%  go.etcd.io/bbolt.(*Bucket).Bucket
  115.05MB  5.71% 91.00%   115.05MB  5.71%  decred.org/dcrdex/client/db.decodeMatchProof_v2
   80.50MB  4.00% 94.99%    80.50MB  4.00%  go.etcd.io/bbolt.newBucket (inline)
```

client/db/bolt/db.go
```diff
@@ -437,12 +438,33 @@ func (db *BoltDB) UpdateOrder(m *dexdb.MetaOrder) error {
...
+
+		// Create or move an order bucket based on order status. Orders
+		// that have status less than or equal to booked go in the
+		// ordersBucket. Finished orders go in the archivedOrdersBucket.
+		var oBkt *bbolt.Bucket
+		var err error
+		executed := md.Status > order.OrderStatusBooked
+		if executed {
+			if ob.Bucket(oid[:]) != nil {
+				ob.DeleteBucket(oid[:])
```
chappjc, line 452: Let's at least log the error from `DeleteBucket`.  But can you comment on the meaning if `ob.Bucket(oid[:])` _does_ return `nil`?  This would mean the order was either (1) never added to the active orders bucket, or (2) it is already in the archived orders bucket?

client/db/bolt/db.go
```diff
@@ -437,12 +438,33 @@ func (db *BoltDB) UpdateOrder(m *dexdb.MetaOrder) error {
...
+			if err != nil {
+				return fmt.Errorf("order bucket error: %w", err)
+			}
+		} else {
+			if aob.Bucket(oid[:]) != nil {
+				return fmt.Errorf("archived order %s set to active", oid)
+			}
+			oBkt, err = ob.CreateBucketIfNotExists(oid[:])
+			if err != nil {
+				return fmt.Errorf("order bucket error: %w", err)
```
chappjc, line 464: Let's supplement these `CreateBucketIfNotExists` error strings with "archived" and "active"

client/db/bolt/db.go
```diff
@@ -437,12 +438,33 @@ func (db *BoltDB) UpdateOrder(m *dexdb.MetaOrder) error {
...
+		if executed {
+			if ob.Bucket(oid[:]) != nil {
+				ob.DeleteBucket(oid[:])
+			}
+			oBkt, err = aob.CreateBucketIfNotExists(oid[:])
+			if err != nil {
+				return fmt.Errorf("order bucket error: %w", err)
+			}
+		} else {
+			if aob.Bucket(oid[:]) != nil {
```
chappjc, line 459: This is expensive, and perhaps unnecessary.  If updating an active order requires an `aob.Bucket` call, we're really loosing a lot of the optimization we wanted from segregating archived orders.  Thoughts on failure modes and if we can live without this?

client/db/bolt/db.go
chappjc: I think that's ok. Also, the insert/update order functions could be modified to prevent it from ever being a possibility.

client/db/bolt/db.go
```diff
@@ -532,24 +552,30 @@ func (db *BoltDB) marketOrdersSince(dexB, baseB, quoteB []byte, n int, since uin
...
-	})
+	}, true)
 }
 
 // newestOrders returns the n newest orders, filtered with a supplied filter
 // function. Each order's bucket is provided to the filter, and a boolean true
 // return value indicates the order should is eligible to be decoded and
 // returned.
-func (db *BoltDB) newestOrders(n int, filter func([]byte, *bbolt.Bucket) bool) ([]*dexdb.MetaOrder, error) {
+func (db *BoltDB) newestOrders(n int, filter func([]byte, *bbolt.Bucket) bool, allOrders bool) ([]*dexdb.MetaOrder, error) {
```
chappjc, line 562: Just a nit, but maybe call it `includeArchived` for clarity and tweak the docs.

client/db/bolt/db.go
chappjc: Same thought in the other function.  It's also one of the expensive things we are trying to avoid.

chappjc (commented):
Partial review.  Will give complete one this afternoon.

client/db/bolt/db.go
```diff
@@ -532,24 +552,30 @@ func (db *BoltDB) marketOrdersSince(dexB, baseB, quoteB []byte, n int, since uin
...
+		}
+		for _, master := range buckets {
+			pairs := newestBuckets(master, n, updateTimeKey, filter)
+			for _, pair := range pairs {
+				oBkt := master.Bucket(pair.k)
+				o, err := decodeOrderBucket(pair.k, oBkt)
+				if err != nil {
+					return err
+				}
+				orders = append(orders, o)
```
chappjc, line 577: At a glance, it looks like it would return more than `n` orders total, just `n` in each active/archived category.  That'll be unexpected for callers of `MarketOrders`, `AccountOrders`, and `Orders` where an `n` is specified.

If that assessment is correct, how about in `newestOrders` we sort the slice from `db.ordersView` and then truncate to `n`.

client/db/bolt/upgrades.go
```diff
@@ -247,6 +250,70 @@ func reloadMatchProofs(tx *bbolt.Tx, skipCancels bool) error {
...
+			return err
+		}
+		// NOTE: Assumes no buckets saved in oBktV1.
+		err = oBktV1.ForEach(func(k, v []byte) error {
+			return oBktV2.Put(k, v)
+		})
+		if err != nil {
+			return err
+		}
+		ordersBktV1.DeleteBucket(k)
```
chappjc, line 312: Either eat the error and comment why, or check and log non-nil error.

client/db/bolt/upgrades.go
```diff
@@ -247,6 +250,70 @@ func reloadMatchProofs(tx *bbolt.Tx, skipCancels bool) error {
...
+	if err != nil {
+		return err
+	}
+
+	ordersBktV1 := tx.Bucket(ordersBucketV1)
+	ordersBktV2 := tx.Bucket(ordersBucketV2)
+	var (
+		oBktV1, oBktV2 *bbolt.Bucket
+		status         order.OrderStatus
+	)
```
chappjc, line 287: How come declared outside the `ForEach` closure?  I tend to prefer limiting scope.

client/db/bolt/upgrades.go
chappjc: Hmm, I'm thinking Printf an error message and <s>continue to next order</s> dump it into the archived orders bucket even though unknown<booked.  As mentioned in my other comment, I think we should rig all the insert/update functions to refuse to write `OrderStatusUnknown` to a key.

client/db/bolt/upgrades.go
```diff
@@ -247,6 +250,70 @@ func reloadMatchProofs(tx *bbolt.Tx, skipCancels bool) error {
...
+			return nil
+		}
+		if status > order.OrderStatusBooked {
+			return nil
+		}
+		oBktV2, err = ordersBktV2.CreateBucket(k)
+		if err != nil {
+			return err
+		}
+		// NOTE: Assumes no buckets saved in oBktV1.
```
chappjc, line 305: I feel like this and similar comments would be clearer like "Assume the order bucket contains only values, no sub-buckets".

chappjc (changes requested):

client/db/bolt/db.go
JoeGruffins: Yes, either it wasn't added to the active orders, or was already deleted, so probably adding a second time. Going by the comment on `UpdateOrder` I thought that was probably expected. It just replaces whatever. Really delete/insert more than update.

client/db/bolt/db.go
JoeGruffins: I think the optimization was targeting reads and not writes. Even with it shouldn't affect getting active orders on startup. I can't imagine a situation where an order should go from archive to active, maybe a status unknown order?, but if one does this will create a duplicate order, which is different behavior than before. Since it probably won't happen, removing the check. Lets just not reverse order statuses...

client/db/bolt/db.go
JoeGruffins: I hadn't noticed this. Added a convoluted answer...

client/db/bolt/db.go
```diff
@@ -558,31 +584,44 @@ func (db *BoltDB) newestOrders(n int, filter func([]byte, *bbolt.Bucket) bool) (
...
 
 // Order fetches a MetaOrder by order ID.
 func (db *BoltDB) Order(oid order.OrderID) (mord *dexdb.MetaOrder, err error) {
 	oidB := oid[:]
-	err = db.ordersView(func(master *bbolt.Bucket) error {
-		oBkt := master.Bucket(oidB)
+	err = db.ordersView(func(ob, aob *bbolt.Bucket) error {
+		oBkt := ob.Bucket(oidB)
+		if oBkt == nil {
+			oBkt = aob.Bucket(oidB)
```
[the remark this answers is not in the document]
JoeGruffins: Commented and renamed to archivedOB.

client/db/bolt/db.go
```diff
@@ -729,12 +778,48 @@ func decodeOrderBucket(oid []byte, oBkt *bbolt.Bucket) (*dexdb.MetaOrder, error)
...
+// updateOrderBucket expects an oid to exist in either the orders or archived
+// order buckets. If status is greater than booked, it first checks active
+// orders. If found it moves the order to the archived bucket and returns the
+// order bucket. If status is less than or equal to booked, it expects the order
+// to already be in active orders and returns the bucket.
+func updateOrderBucket(ob, aob *bbolt.Bucket, oid order.OrderID, status order.OrderStatus) (*bbolt.Bucket, error) {
+	executed := status > order.OrderStatusBooked
+	if executed {
+		if bkt := ob.Bucket(oid[:]); bkt != nil {
+			oBkt, err := aob.CreateBucket(oid[:])
```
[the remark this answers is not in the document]
JoeGruffins: Ok.

client/db/bolt/db.go
```diff
@@ -729,12 +778,48 @@ func decodeOrderBucket(oid []byte, oBkt *bbolt.Bucket) (*dexdb.MetaOrder, error)
...
+			if err != nil {
+				return nil, err
+			}
+			// NOTE: Assumes no buckets saved in bkt.
+			bkt.ForEach(func(k, v []byte) error {
+				return oBkt.Put(k, v)
+			})
+			ob.DeleteBucket(oid[:])
```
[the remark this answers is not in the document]
JoeGruffins: Accidental ignore.

client/db/bolt/db.go
```diff
@@ -729,12 +778,48 @@ func decodeOrderBucket(oid []byte, oBkt *bbolt.Bucket) (*dexdb.MetaOrder, error)
...
+			return oBkt, nil
+		}
+		oBkt := aob.Bucket(oid[:])
```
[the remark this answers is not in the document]
JoeGruffins: Ok.

client/db/bolt/db.go
```diff
@@ -729,12 +778,48 @@ func decodeOrderBucket(oid []byte, oBkt *bbolt.Bucket) (*dexdb.MetaOrder, error)
...
+		if oBkt == nil {
+			return nil, fmt.Errorf("order %s not found", oid)
+		}
+		return oBkt, nil
+	}
+	if aob.Bucket(oid[:]) != nil {
```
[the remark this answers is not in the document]
JoeGruffins: Removing this check.

client/db/bolt/upgrades.go
```diff
@@ -247,6 +250,70 @@ func reloadMatchProofs(tx *bbolt.Tx, skipCancels bool) error {
...
+	if err := ensureVersion(dbtx, oldVersion); err != nil {
+		return err
+	}
+
+	// Move any executed orders to the new archivedOrdersBucket.
+	return moveActiveOrders(dbtx)
+}
+
+// moveActiveOrders searches the v1 ordersBucket for orders that are not yet
+// booked booked status, adds those to the v2 ordersBucket, and deletes them
```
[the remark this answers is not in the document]
JoeGruffins: Ok. Also make status.IsActive a thing.

client/db/bolt/upgrades.go
JoeGruffins: Unintended ignore.

client/db/bolt/upgrades.go
JoeGruffins: Limited.

client/db/bolt/upgrades.go
JoeGruffins: Ok.

client/db/bolt/db.go
JoeGruffins: I believe all the update functions now error on unknown status.

dex/order/status.go
```diff
@@ -74,3 +74,8 @@ func (s OrderStatus) String() string {
 	}
 	return name
 }
+
+// IsActive returns wether the status is considered active.
```
chappjc, line 78: whether

client/db/bolt/db.go
```diff
@@ -532,24 +559,53 @@ func (db *BoltDB) marketOrdersSince(dexB, baseB, quoteB []byte, n int, since uin
...
+		}
+		idx := newTimeIndexNewest(n)
+		for i, master := range buckets {
+			err := master.ForEach(func(k, _ []byte) error {
+				bkt := master.Bucket(k)
+				stamp := intCoder.Uint64(bkt.Get(updateTimeKey))
+				if filter == nil || filter(k, bkt) {
+					// Append the bucket position to the key
+					// so that we know which bucket it is later.
+					k = append(k, byte(i))
```
chappjc, line 584: I see what you're doing in this commit, but modifying the key's `[]byte` that `ForEach` provides makes me super uncomfortable.  The `bbolt.Cursor` docs indicate the key and value byte slices are only valid for the life of the transaction, which is not the problem per se, except that this indicates these are _bolt-internal buffers_ that should only be read or copied.

Also TBH I really liked the simplicity of the previous `newestOrders`, which could stay simple with just a sort at the end.  The sort would involve few swaps if the active orders bucket is first.  Open to other solutions though.

JoeGruffins: > The bbolt.Cursor docs indicate the key and value byte slices are only valid for the life of the transaction, which is not the problem per se, except that this indicates these are bolt-internal buffers that should only be read or copied.

I believe `newestBuckets` was already doing this.

I really just adapted that method, because it looks like `timeIndexNewest` is good for handling our set of `n` orders.

Will push two commits, one that fixes the copy issue, and another that just uses `newestBuckets` and sorts and you can decide which is better.

chappjc: I'm referring to `append(k, byte(i))` changing `k` from `ForEach`.

JoeGruffins:
Actually, this is probably better, because the order is actually newest to oldest, as we see in the test...

client/db/bolt/db_test.go
JoeGruffins: Fixed with 831a745

client/db/bolt/db.go
JoeGruffins: Yeah, it is just reading. I see now.

client/db/bolt/db.go
```diff
@@ -52,7 +52,8 @@ var (
 	appBucket              = []byte("appBucket")
 	accountsBucket         = []byte("accounts")
 	disabledAccountsBucket = []byte("disabledAccounts")
-	ordersBucket           = []byte("orders")
+	ordersBucket           = []byte("activeOrders")
```
buck54321, line 55: What to you think about renaming the `const` to `activeOrdersBucket`?

buck54321 (changes requested):
Looking good, testing well. Just a couple little things.

client/db/bolt/db.go
chappjc: > I think the optimization was targeting reads

Right I'm was referring to `aob.Bucket(oid[:]) != nil` being expensive as it has to search through the archived orders bucket.

client/db/bolt/db.go
```diff
@@ -558,31 +605,48 @@ func (db *BoltDB) newestOrders(n int, filter func([]byte, *bbolt.Bucket) bool) (
...
+	return orders, db.ordersView(func(ob, archivedOB *bbolt.Bucket) error {
+		buckets := []*bbolt.Bucket{ob}
+		if includeArchived {
+			buckets = append(buckets, archivedOB)
+		}
+		for _, master := range buckets {
+			err := master.ForEach(func(oid, _ []byte) error {
+				oBkt := master.Bucket(oid)
+				// If the order is not in the active bucket,
+				// check the archived orders bucket.
```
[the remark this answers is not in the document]
JoeGruffins: No.

client/db/bolt/upgrades.go
```diff
@@ -247,6 +250,64 @@ func reloadMatchProofs(tx *bbolt.Tx, skipCancels bool) error {
...
+	// Move any inactive orders to the new archivedOrdersBucket.
+	return moveActiveOrders(dbtx)
+}
+
+// moveActiveOrders searches the v1 ordersBucket for orders that are inactive,
+// adds those to the v2 ordersBucket, and deletes them from the v1 ordersBucket,
+// which becomes the archived orders bucket.
+func moveActiveOrders(tx *bbolt.Tx) error {
+	ordersBucketV1 := []byte("orders")
+	ordersBucketV2 := []byte("activeOrders")
```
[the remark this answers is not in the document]
JoeGruffins: I thought it would be better to keep these local to the upgrade functions, as the package variables may change and unknowingly break the upgrade.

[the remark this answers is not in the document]
JoeGruffins: Will work on the naming.

client/db/bolt/db.go
```diff
@@ -431,18 +432,46 @@ func (db *BoltDB) disabledAcctsUpdate(f bucketFunc) error {
...
+			oBkt, err = archivedOB.CreateBucketIfNotExists(oid[:])
+			if err != nil {
+				return fmt.Errorf("archive order bucket error: %w", err)
+			}
+		} else {
+			oBkt, err = ob.CreateBucketIfNotExists(oid[:])
+			if err != nil {
+				return fmt.Errorf("active order bucket error: %w", err)
+			}
 		}
```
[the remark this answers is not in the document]
JoeGruffins: Ok.

client/db/bolt/db.go
JoeGruffins: Works for me!

client/db/bolt/upgrades_test.go
```diff
@@ -128,14 +131,15 @@ func verifyV1Upgrade(t *testing.T, db *bbolt.DB) {
 func verifyV2Upgrade(t *testing.T, db *bbolt.DB) {
 	t.Helper()
 	maxFeeB := uint64Bytes(^uint64(0))
+	ordersBucketV1 := []byte("orders")
```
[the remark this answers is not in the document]
JoeGruffins: Not wanting to shadow the package lvl name, going with `oldOrdersBucket` and `newActiveOrdersBucket`

chappjc (approved):
Wow, putting the bucket in the `timeIndexNewest` solved multiple problems.  Great suggestion @buck54321

chappjc:
I should note that depending on how testing goes with this, we may want to follow up https://github.com/decred/dcrdex/issues/1039 with an active *matches* bucket similar to how we segregated the orders.

client/db/bolt/upgrades.go
```diff
@@ -247,6 +250,64 @@ func reloadMatchProofs(tx *bbolt.Tx, skipCancels bool) error {
...
+	archivedOrdersBkt := tx.Bucket(oldOrdersBucket)
+	activeOrdersBkt := tx.Bucket(newActiveOrdersBucket)
+	return archivedOrdersBkt.ForEach(func(k, _ []byte) error {
+		archivedOBkt := archivedOrdersBkt.Bucket(k)
+		if archivedOBkt == nil {
+			return fmt.Errorf("order %x bucket is not a bucket", k)
+		}
+		status := order.OrderStatus(intCoder.Uint16(archivedOBkt.Get(statusKey)))
+		if status == order.OrderStatusUnknown {
+			fmt.Printf("Encountered order with unknown status: %x\n", k)
```
buck54321, line 291: Is it worth adding a `dex.Logger` argument to all the upgrade funcs?

buck54321 (approved):

client/db/bolt/upgrades.go
chappjc: Let's table that. But maybe.

PR #1072, opened 2021-04-27
chart: stop the wheel event propagation.

### LINKED ISSUE #1063: mouse wheel scrolls page AND zooms depth chart
Just updated to Firefox 89 and on the depth chart the mousewheel doesn't just zoom, it now scrolls the whole window AND zooms.  This seems like a Firefox bug, but I'm not sure.

EDIT: It's just the long orders tables, which I think need to be limited in height and with their own scrolls.  Open to other solutions though.

amassarwi:
This stops the wheel event propagation to prevent the page scroll bug when having long orders list.

Closes #1063.

### DIFF
--- a/client/webserver/site/src/js/charts.js
+++ b/client/webserver/site/src/js/charts.js
@@ -116,6 +116,7 @@ export class DepthChart {
   // wheel is a mousewheel event handler.
   wheel (e) {
     this.zoom(e.deltaY < 0)
+    e.preventDefault()
   }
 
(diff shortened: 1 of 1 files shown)


chappjc (approved):
Confirmed this stops page scroll when over the depth chart.  I think we should get this in 0.2 because Decrediton's dex window is likely to be a bit smaller by default, although it will be resizable.

buck54321 (approved):

PR #1073, opened 2021-04-28
Handle all the errors when creating DEX.

LasTshaMAN:
I've been looking at `DEX` codebase lately, and one of the things I've noticed (addressing it via this PR) is unhandled errors when creating `DEX` instance.
It seems to me, `err` should be propagated further from `NewDEX` if any of the `subsystems` failed to `Connect` (cause I doubt there are attempts to re-connect a subsystem if it fails to connect on the very first try, are there ?).

Or, if that's not the case, then maybe we should ignore `error`s from `startSubSys` explicitely (to show our intentions) ? Leaving a comment for why we ignore error would help as well.

Disclamer: this is my first contribution to `DEX`.

### DIFF
--- a/server/dex/dex.go
+++ b/server/dex/dex.go
@@ -560,7 +560,7 @@ func NewDEX(ctx context.Context, cfg *DexConf) (*DEX, error) {
 	// Start the AuthManager and Swapper subsystems after populating the markets
 	// map used by the unbook callbacks, and setting the AuthManager's unbook
 	// timers for the users with currently booked orders.
-	//
+
 	// This subsystem is a dex.Runner, it doesn't return errors.
 	_ = startSubSys("Auth manager", authMgr)
 
--- a/server/dex/dex.go
+++ b/server/dex/dex.go
@@ -345,10 +345,11 @@ func NewDEX(ctx context.Context, cfg *DexConf) (*DEX, error) {
 			}
 		}
 
-		err = startSubSys(fmt.Sprintf("Asset[%s]", symbol), be)
+		assetName := fmt.Sprintf("Asset[%s]", symbol)
+		err = startSubSys(assetName, be)
 		if err != nil {
 			abort()
-			return nil, fmt.Errorf("failed to start asset %q: %w", symbol, err)
+			return nil, fmt.Errorf("start subsystems for %s: %w", assetName, err)
 		}
 
 		initTxSize := uint64(be.InitTxSize())
@@ -559,8 +560,18 @@ func NewDEX(ctx context.Context, cfg *DexConf) (*DEX, error) {
 	// Start the AuthManager and Swapper subsystems after populating the markets
 	// map used by the unbook callbacks, and setting the AuthManager's unbook
 	// timers for the users with currently booked orders.
-	startSubSys("Auth manager", authMgr)
-	startSubSys("Swapper", swapper)
+	const authManagerName = "Auth manager"
+	err = startSubSys(authManagerName, authMgr)
+	if err != nil {
+		abort()
+		return nil, fmt.Errorf("start subsystems for %s: %w", authManagerName, err)
+	}
+	const swapperName = "Swapper"
+	err = startSubSys(swapperName, swapper)
+	if err != nil {
+		abort()
+		return nil, fmt.Errorf("start subsystems for %s: %w", swapperName, err)
+	}
 
 	// Set start epoch index for each market. Also create BookSources for the
 	// BookRouter, and MarketTunnels for the OrderRouter.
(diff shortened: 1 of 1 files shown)


JoeGruffins (approved):
Hello! Welcome!

This does look more correct to me. Thank you.

chappjc:
Initially the `subsystem` type was created for `dex.Runner`s, which don't have an error when launching, only errors from the constructors.  The type was changed to accommodate a `dex.Connector`, which does have an error.

If you'd like to start these like `_ = startSubSys(..) // dex.Runner has no error`, I don't object.  That is good practice and it shuts up `errcheck`.

LasTshaMAN:
Frankly, I would not combine `dex.Runner` & `dex.Connector` under the same `subsystem` object in this way, looks like the only reason to do it to have a single construction to `stop` these sub-systems:
```
for _, ss := range subsystems {
        ss.stop()
}
```

Also, there might appear some additional differences in the future, which render `subsystem` concept even more problematic.

chappjc:
The single stack is the main motivation, yes. But it's not just because a single range is tighter, but because we can have the subsystem stop in the reverse order that we started them.  There's also a market subsystem lookup helper, but there are other solutions to that such as a map to these market subsystems.

I respect you point of view, and agree that it's a bit janky that the Connector has been shoe-horned into subsystem with connector, but churning on this doesn't strike me as productive.  (What problem are we solving by changing this now?)

LasTshaMAN:
> we can have the subsystems stop in the reverse order

That's a good point, `NewDEX` just seem overall a bit messy (read complex), so that at some point in the future this might need to be addressed (or not, who knows ...)

> but churning on this doesn't strike me as productive. (What problem are we solving by changing this now?)

I didn't check whether every unhandled `err` is coming from `dex.Runner`, not from `dex.Connector`:
1) if it does, this PR probably doesn't "solve anything" (except introduce me to the project ;)
2) if not, then we possibly have some unhandled errors (from `dex.Connector`), and given the complexity of the components involved I'm not sure whether it can lead to any bugs or not

so, if you believe it to be 1) I guess we can close this PR ?

chappjc:
Again, it's worth verifying that we don't have any unhandled errors, and documenting the ones that we mean to ignore. It's trivial to determine which subsystems are `dex.Runner`s.   I just don't want to refactor things that do not benefit from changes in any meaningful way.

But IMO `NewDEX` is extremely straightforward. It has a list of subsystems to bridge and start, and that's all it does.  It could be broken up into several functions, but for little benefit and, I believe, at the cost of clarity.  I welcome constructive feedback, but please be specific about what issues you see in the future because I can only guess what you're thinking.  Bear in mind that this `server/dex.DEX` type has a very limited role, mostly comprising code that runs just once for the lifetime of a DEX server, with immediate and obvious failure modes.

LasTshaMAN:
Understood. 

Okay, I've modified this change a bit. To summarize, instead of checking error for every `subsystem`, I've followed through each one of them (used in `NewDEX`) and
- for every `dex.Runner` I've added explicit error handling with `_ = ...`
- for every `dex.Connector` I've added the error handling (as was done previously in this PR)

> But IMO `NewDEX` is extremely straightforward.

Hope you don't hate me for sharing an opinion, but I want to :)
Overall it is pretty straightforward, but, I believe, there very often are potentially hidden issues in very long `func`s (not to mention, as I've pointed out above, `Backend` subsystem being `dex.Connector`s returns  unhandled error), for example,

it looks "buggy" to me to (missing `abort` call) have these 2 blocks of code one after another, first block calling both  `abort` & `return` while in the next block there is no call to `abort`, just `return` (I assume there is no difference between `fmt.Errorf("no DCR backend configured")` and `fmt.Errorf("invalid regfeexpub: %w", err)` errors).
```go
	// Ensure their is a DCR asset backend.
	if dcrBackend == nil {
		abort()
		return nil, fmt.Errorf("no DCR backend configured")
	}
	// Validate the registration fee extended public key.
	if err := dcrBackend.ValidateXPub(cfg.RegFeeXPub); err != nil {
		return nil, fmt.Errorf("invalid regfeexpub: %w", err)
	}
```

And over time, changing/adding stuff to `NewDEX` might result in even more bugs (although these bugs might not be severe from user experience point of view, they are bugs all the same), so these things need to be refactored from time to time.

As a possible further improvement (not gonna do it, just pointing out), I would suggest removing `subsystems` abstraction, and just use a separate `defer` for each `subsystem` to ensure reverse termination order.

buck54321:
@LasTshaMAN You've spent thousands of words pontificating and repeatedly wasted @chappjc's time, and have contributed zero lines of consequential code. We've asked you both in public and private to find something productive to do, yet here you are again, trying to waste our time.

I'm closing this PR to stop the time bleed you insist on imposing on us. You've been assigned an issue to work on now. Don't bother us with stuff like this again until you've shown some aptitude.

chappjc:
>  I would suggest removing `subsystems` abstraction, and just use a separate `defer` for each `subsystem` to ensure reverse termination order.

That's a thought to replace abort with a defer that checks a reused err, but obviously dex lifetime doesn't end at NewDEX return, so the exported Stop method with the subsystem stack is still required.  Also, accidentally shadowing err is a pretty common bug so I'm not inclined to make changes to abort right now, but it's a thought.

Let's move on to other issues.  We have a long way to go before 1.0 and we unfortunately don't have the luxury of churning on working code.  You'll continue to find aesthetically unpleasing code, but we have to pick our battles.  Thanks for understanding the realities of running this project.

LasTshaMAN:
Thanks for the proper reply, @chappjc.

PR #1074, opened 2021-05-02
Add information about fees to readme. 

### LINKED ISSUE #822: Add information about transaction fees to readme
DCRDEX users pay on chain transaction fees to execute their trades, and a number of transactions are required to set up and execute an atomic swap. At present, the readme does not explain how many transactions are required to make an atomic swap trade through DCRDEX, or how much the user might expect to pay in fees. A description of the number of transactions required to open and execute an order (with pre-sizing and without) would manage users' expectations about the fees they will pay to trade and allow them to more easily make their own estimates.

martonp:
Closes #822.

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,7 +213,7 @@ to appeal.
 
 ## Fees
 
-The dex does not charge trading fees, but users pay on-chain transaction fees.
+DEX does not charge trading fees, but users pay on-chain transaction fees.
 Transaction fees vary based on how orders are matched.
 
 To ensure that on-chain transaction fees do not eat a significant portion of the
--- a/README.md
+++ b/README.md
@@ -213,20 +213,22 @@ to appeal.
 
 ## Fees
 
-The dex does not charge fees, but users pay on-chain transaction fees.
+The dex does not charge trading fees, but users pay on-chain transaction fees.
+Transaction fees vary based on how orders are matched.
 
 To ensure that on-chain transaction fees do not eat a significant portion of the
 order quantity, orders must be specified in increments of a minimum lot size.
-For example, if on-chain transaction fees worked out to $5, and a user was able
+To illustrate, if on-chain transaction fees worked out to $5, and a user was able
 to place an order to trade $10, they would lose half of their trade to
-transaction fees. So for high-fee assets, the operator must set a large lot
-size.
+transaction fees. For chains with single-match fees of $5, if the operator wanted
+to limit possible fees to under 1% of the trade, the minimum lot size would need
+to be set to about $500.
 
-The best case scenario is for an entire order be consumed by a single match.
-If this happens, the user pays 2 transaction fees: one swap and one redemption.
-The worst case is for the order to be consumed by single lot matches per epoch,
-in which case the user pays for a separate swap and redemption transaction per
-lot in their order.
+The scenario with the lowest fees is for an entire order to be consumed by a
+single match. If this happens, the user pays 2 transaction fees: one swap and
+one redemption. The worst case is for the order to be consumed by one single-lot
+match per epoch, in which case the user pays for a separate swap and redemption
+transaction per lot in their order.
 Check the
 [dex specification](https://github.com/decred/dcrdex/blob/master/spec/atomic.mediawiki)
 for more details about how atomic swaps work.
--- a/README.md
+++ b/README.md
@@ -227,7 +227,7 @@ sending and recieving transaction for each match. Check the
 [dex specification](https://github.com/decred/dcrdex/blob/master/spec/atomic.mediawiki)
 for more details about how atomic swaps work.
 
-The client contains on option to pre-split UTXOs in order to not lock up
+The client contains an option to pre-split UTXOs in order to not lock up
 more funds than necessary for the duration of an atomic swap. This option will
 potentially introduce an additional transaction (for each match) on the chain
 you are sending funds from, but it could also speed up the entire process.
--- a/README.md
+++ b/README.md
@@ -213,9 +213,9 @@ to appeal.
 
 ## Fees
 
-The dex does not charge fees, but users have to pay on-chain transaction fees.
-Unlike on centralized exchanges where users can trade any arbitrary amount of
-funds, orders on the dex must be made in increments of a specified lot size.
+The dex does not charge trading fees, but users have to pay on-chain transaction
+fees. Unlike on centralized exchanges where users can trade any arbitrary amount
+of funds, orders on the dex must be made in increments of a specified lot size.
 For example, if the lot size for the DCR/BTC market is 40 DCR, users can
 make orders to buy or sell DCR in increments of 40.
 
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ privileges and forfeiture of registration fee.
 - [Client Quick Start Installation](#client-quick-start-installation)
 - [Client Configuration](#client-configuration)
 - [Important Stuff to Know](#important-stuff-to-know)
+- [Fees](#fees)
 - [Advanced Client Installation](#advanced-client-installation)
 - [DEX Specification](#dex-specification)
 - [Client Applications and the Core Package](#client-applications-and-the-core-package)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -210,6 +211,27 @@ For dex.decred.org, reach out
 [on Element](https://matrix.to/#/!mlRZqBtfWHrcmgdTWB:decred.org?via=decred.org&via=matrix.org&via=planetdecred.org)
 to appeal.
 
+## Fees
+
+The dex does not charge fees, but users have to pay on-chain transaction fees.
```
JoeGruffins, line 216: Currently, there is a sign up fee. So I would rather this be "The dex does not charge **trading** fees."

JoeGruffins (approved):
Looks good to me! Just one nit.

README.md
```diff
@@ -210,6 +211,27 @@ For dex.decred.org, reach out
...
+You might be lucky and have your entire order matched by one other user. In
+this case, you will have to pay 2 transaction fees: one for sending funds on
+one chain, and one for receiving funds on the other. In the worst case, each
+lot of your order is matched by a different user, and you will pay a separate
+sending and recieving transaction for each match. Check the 
+[dex specification](https://github.com/decred/dcrdex/blob/master/spec/atomic.mediawiki)
+for more details about how atomic swaps work.
+
+The client contains an option to pre-split UTXOs in order to not lock up
+more funds than necessary for the duration of an atomic swap. This option will
```
buck54321, line 231: ~~The client contains an option to pre-split UTXOs in order to not lock up more funds than necessary for the duration of an atomic swap.~~

Wallets for UTXO-based assets provide an option to use an additional transaction per order to pre-size outputs and prevent over-locking of funds that might otherwise occur until the first swap is broadcast for the order.

---

But really, I think this is more detail than is appropriate for the README. We have a wiki for this kind of stuff.

README.md
```diff
@@ -210,6 +211,27 @@ For dex.decred.org, reach out
...
+## Fees
+
+The dex does not charge trading fees, but users have to pay on-chain transaction
+fees. Unlike on centralized exchanges where users can trade any arbitrary amount
+of funds, orders on the dex must be made in increments of a specified lot size.
+For example, if the lot size for the DCR/BTC market is 40 DCR, users can
+make orders to buy or sell DCR in increments of 40.
+
+You might be lucky and have your entire order matched by one other user. In
+this case, you will have to pay 2 transaction fees: one for sending funds on
```
buck54321, line 223: The order can match multiple users and still only require 2 transactions. Matches can be grouped. You are correct in your description of the conditions necessary for the best case scenario, but the benefit you describe (2 transactions) is not unique to that particular mode. In fact, at the API level, there is almost no concept of a transaction. All we should really say is that **the best case scenario is when the entire order is consumed by a single match**, in which case only one swap and one redemption will be created, and the worst-case scenario is **one single-lot match per epoch**.

README.md
```diff
@@ -210,6 +211,27 @@ For dex.decred.org, reach out
 [on Element](https://matrix.to/#/!mlRZqBtfWHrcmgdTWB:decred.org?via=decred.org&via=matrix.org&via=planetdecred.org)
 to appeal.
 
+## Fees
+
+The dex does not charge trading fees, but users have to pay on-chain transaction
+fees. Unlike on centralized exchanges where users can trade any arbitrary amount
+of funds, orders on the dex must be made in increments of a specified lot size.
```
buck54321, line 218: No reason to compare and contrast with centralized exchanges here. Just give them our reasoning.

> To ensure that on-chain transaction fees do not eat a significant portion of the order quantity, orders must be specified in increments of a minimum lot size. For example, if on-chain transaction fees worked out to $5, and a user was able to place an order to trade $10, they would lose half of their trade to transaction fees. So for high-fee assets, the operator must set a large lot size.

buck54321 (changes requested):

martonp:
@buck54321 Thanks for your review.

README.md
martonp: If you match with two separate users, won't you have to pay for two separate redemptions?

README.md
```diff
@@ -213,25 +213,24 @@ to appeal.
...
-of funds, orders on the dex must be made in increments of a specified lot size.
-For example, if the lot size for the DCR/BTC market is 40 DCR, users can
-make orders to buy or sell DCR in increments of 40.
-
-You might be lucky and have your entire order matched by one other user. In
-this case, you will have to pay 2 transaction fees: one for sending funds on
-one chain, and one for receiving funds on the other. In the worst case, each
-lot of your order is matched by a different user, and you will pay a separate
-sending and recieving transaction for each match. Check the 
+The dex does not charge fees, but users pay on-chain transaction fees.
```
buck54321, line 216: Actually, "**trading** fees" is better here, since there is still a registration fee.

README.md
```diff
@@ -213,25 +213,24 @@ to appeal.
...
+To ensure that on-chain transaction fees do not eat a significant portion of the
+order quantity, orders must be specified in increments of a minimum lot size.
+For example, if on-chain transaction fees worked out to $5, and a user was able
+to place an order to trade $10, they would lose half of their trade to
+transaction fees. So for high-fee assets, the operator must set a large lot
+size.
+
+The best case scenario is for an entire order be consumed by a single match.
+If this happens, the user pays 2 transaction fees: one swap and one redemption.
+The worst case is for the order to be consumed by single lot matches per epoch,
```
buck54321, line 227: single lot matches per epoch -> one single-lot match per epoch

Just to explain, you can have more than one single-lot match in an epoch if you match with different orders. In that case, matches can be grouped, so it's no longer worst-case.

README.md
```diff
@@ -210,6 +211,26 @@ For dex.decred.org, reach out
 [on Element](https://matrix.to/#/!mlRZqBtfWHrcmgdTWB:decred.org?via=decred.org&via=matrix.org&via=planetdecred.org)
 to appeal.
 
+## Fees
+
+The dex does not charge fees, but users pay on-chain transaction fees.
+
+To ensure that on-chain transaction fees do not eat a significant portion of the
+order quantity, orders must be specified in increments of a minimum lot size.
+For example, if on-chain transaction fees worked out to $5, and a user was able
```
buck54321, line 220: For example -> To illustrate

README.md
```diff
@@ -210,6 +211,26 @@ For dex.decred.org, reach out
...
+to place an order to trade $10, they would lose half of their trade to
+transaction fees. So for high-fee assets, the operator must set a large lot
+size.
```
buck54321, line 223: So -> This is why

But...

Consider sticking to example mode here. Something like

> For chains with single-match fees of $5, if the operator wanted to limit possible fees to under 1% of the trade, the minimum lot size would need to be set to about $500.

buck54321 (changes requested):

README.md
martonp: Got it.

README.md
martonp: Definitely nicer.

README.md
```diff
@@ -210,6 +211,28 @@ For dex.decred.org, reach out
...
+order quantity, orders must be specified in increments of a minimum lot size.
+To illustrate, if on-chain transaction fees worked out to $5, and a user was able
+to place an order to trade $10, they would lose half of their trade to
+transaction fees. For chains with single-match fees of $5, if the operator wanted
+to limit possible fees to under 1% of the trade, the minimum lot size would need
+to be set to about $500.
+
+The scenario with the lowest fees is for an entire order to be consumed by a
+single match. If this happens, the user pays 2 transaction fees: one swap and
+one redemption. The worst case is for the order to be consumed by one single-lot
```
chappjc, line 229: > If this happens, the user pays 2 transaction fees: one swap and one redemption.

In the above, I would state that these are on different chains.  That's important because if one asset is low fee like DCR, and the other is high like BTC, you really only care about the tx on the high fee chain.

chappjc (changes requested):

README.md
```diff
@@ -210,6 +211,28 @@ For dex.decred.org, reach out
...
+match per epoch, in which case the user pays for a separate swap and redemption
```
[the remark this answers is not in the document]
martonp: If an order is filled by multiple matches in the same epoch, wouldn't the client create a single transaction with multiple outputs that satisfy each match?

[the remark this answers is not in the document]
chappjc: Not always.

Say the user submits an order that is taker on matches with 3 different booked (maker) orders.  Those makers will each have their own swap contracts.  The taker may or may not be able to redeem them in the same transaction depending on how they are broadcast and mined, but even if that does happen just right, they will be separate inputs to a redeem txn.

Further consider if the taker order in the above is a 4 lot order, where one lot gets booked, and filled as maker in the same epoch.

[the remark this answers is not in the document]
chappjc: Conversely, you can have matches in different epochs that end up getting batched in the same redeem txn if the swaps happen to reach req. confirmations at the same height, which happens all the time.

README.md
chappjc: Correct it's no longer the worst-case, but not best either given the txn size is bigger.  Referring to discussion in https://github.com/decred/dcrdex/pull/1074#discussion_r645634303

The per-epoch thing is also somewhat irrelevant for taker orders, as I've illustrated there.

README.md
chappjc: Agree with best case scenario 100%.  Worst-case I think is more nuanced than that, esp for taker where it's still potentially worst case with multiple matches in the same epoch.

README.md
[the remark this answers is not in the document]
martonp: Makes sense, thanks.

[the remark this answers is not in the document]
chappjc: I've edited my suggestion to include the word "potentially".  I think focusing on "per epoch" is a bit misleading unless you're the maker, so I'd prefer to describe the worst case in terms of when there is a separate match for each lot of the order.  How these get batched depends on maker/taker, and other factors including counterparty timing and the miners.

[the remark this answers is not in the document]
martonp: Basically it's between two and `2* #lots` transactions, half on one chain and half on the other, but there can many scenarios in between. I've updated it.. I think it's clear now.

chappjc (approved):

README.md
```diff
@@ -210,6 +211,29 @@ For dex.decred.org, reach out
 [on Element](https://matrix.to/#/!mlRZqBtfWHrcmgdTWB:decred.org?via=decred.org&via=matrix.org&via=planetdecred.org)
 to appeal.
 
+## Fees
+
+The dex does not charge trading fees, but users pay on-chain transaction fees.
```
buck54321, line 216: ~~The dex~~ -> DEX or Decred DEX

buck54321 (changes requested):
One last thing and it's good to go.

README.md
martonp: Done.

PR #1075, opened 2021-05-03
client/db: Make client DB interface more specific

### LINKED ISSUE #935: client/db: rename Get/Store to be specific to their use (enc params)
https://github.com/decred/dcrdex/blob/792602bdab53f2b933f51142df79432fada378e8/client/db/interface.go#L13-L18

The `client/db.DB` interface is pretty clean and dex specific, but the two Get and Store methods are a little too general purpose and k/v oriented.  As far as I can tell we're only using this to access the encryption key parameters on app init.  As such, these should be renamed to something like EncryptionParams and SetEncryptionParams.

LasTshaMAN:
Simple renaming, as described in https://github.com/decred/dcrdex/issues/935, clarifying minor details in the related comments.

Seems like an appropriate issue to tackle for a beginner :)

closes #935

### DIFF
--- a/client/db/bolt/db.go
+++ b/client/db/bolt/db.go
@@ -159,8 +159,9 @@ func (db *BoltDB) Run(ctx context.Context) {
 	db.Close()
 }
 
-// SetEncryptionParams stores encryption parameters params (associated with key k)
-// overwriting previous value if any already exists for key k.
+// SetEncryptionParams stores encryption parameters
+// params (associated with key k) overwriting previous value
+// if any already exists for key k.
 func (db *BoltDB) SetEncryptionParams(k string, v []byte) error {
 	if len(k) == 0 {
 		return fmt.Errorf("cannot store with empty key")
@@ -175,7 +176,8 @@ func (db *BoltDB) SetEncryptionParams(k string, v []byte) error {
 	})
 }
 
-// IsEncryptionParamsSet checks whether encryption parameters associated with key k were previously stored in DB
+// IsEncryptionParamsSet checks whether encryption parameters
+// associated with key k were previously stored in DB
 // via a call to SetEncryptionParams.
 func (db *BoltDB) IsEncryptionParamsSet(k string) (bool, error) {
 	var exists bool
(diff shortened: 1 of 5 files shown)


buck54321:
I agree that this is the right solution, but note that this work is superceded by #1015. I have no problem merging this work first so that you keep the contribution, but come chat with us [on Element](https://chat.decred.org/#/room/#dexdev:decred.org) before you get too far into another issue.

LasTshaMAN:
Since this is pretty small change it was easier for me to drop a PR right away, instead of discussing it first.

So, let us not worry about my contribution, lets do what's best for this project. Feel free to throw it away (and maybe reference https://github.com/decred/dcrdex/issues/935 issue in  https://github.com/decred/dcrdex/pull/1015).

For anything bigger I'll definitely ask first, thanks for invitation ;)

client/db/bolt/db.go
```diff
@@ -174,9 +175,9 @@ func (db *BoltDB) Store(k string, v []byte) error {
 	})
 }
 
-// ValueExists checks if a value was previously stored in the general-use
-// bucket at the specified key.
-func (db *BoltDB) ValueExists(k string) (bool, error) {
+// IsEncryptionParamsSet checks whether encryption parameters associated with key k were previously stored in DB
+// via a call to SetEncryptionParams.
```
JoeGruffins, line 179: There is a effort to keep comments under 80 columns in width from the left. It's not set in stone, but this comment  and others go too far to the right.

JoeGruffins (commented):
If this is meant to close/fix the issue, put something like `closes #935` in the OP to make it clear and automatic.

client/db/bolt/db.go
LasTshaMAN: Fixed.

Although, I would point out:
- usually, linters work better to enforce such conventions

chappjc: There are valid exceptions to this convention, so it's not great for a linter.  Please just wrap such comments.  Setup something in your editor and you'll never have to think about it.

LasTshaMAN: Okay, np.

client/db/bolt/db.go
```diff
@@ -159,8 +159,10 @@ func (db *BoltDB) Run(ctx context.Context) {
 	db.Close()
 }
 
-// Store stores a value at the specified key in the general-use bucket.
-func (db *BoltDB) Store(k string, v []byte) error {
+// SetEncryptionParams stores encryption parameters
+// params (associated with key k) overwriting previous value
+// if any already exists for key k.
+func (db *BoltDB) SetEncryptionParams(k string, v []byte) error {
```
chappjc, line 165: Almost the solution. Making these fully specific to the encryption parameters would be to eliminate the key `k` input argument since that becomes a DB implementation detail.  The `k`-`v` only make sense for generic Get/Store, but that's what we wanted to eliminate.

Anyway, this is a moot point with https://github.com/decred/dcrdex/pull/1015

Thank you for your contribution here though!

LasTshaMAN: Yeah, I thought about that too. 

I think at this point it is no longer worth to develop this change any further, so I'll close in favor of https://github.com/decred/dcrdex/pull/1015.

LasTshaMAN:
Closing in favor of https://github.com/decred/dcrdex/pull/1015.

Issue #1020, opened 2021-03-19
client should store and check csum in preimage request


chappjc:
Server should not be asking for preimages until the epoch is closed.  So client should be remembering the `CommitChecksum` field of `PreimageRequest` and verify that the order doesn't end up in an epoch's match proof with a different `csum`, which would indicate server is cheating (e.g. we handed over our preimage so the server could game the shuffle).

_Originally posted by @chappjc in https://github.com/decred/dcrdex/pull/1009#discussion_r597358893_

buck54321:
Relevant spec section is https://github.com/decred/dcrdex/blob/master/spec/orders.mediawiki#Preimage_Reveal, as well as the `preimage` and `match_proof` requests.

1. Store the checksum in `handlePreimageRequest`, probably as a field in the `trackedTrade`.
2. After `ValidateMatchProof` in `handleMatchProofMsg`, check the stored checksum from part 1 against the `note.CSum`. If mismatch, then ...

Under normal circumstances, it would be almost impossible for a client to store the checksum, shut down, restart, and log in before the match_proof comes, so I don't see a compelling case for storing the checksum to the database and the upgrade it would involve.

chappjc:
@LasTshaMAN is assigned.

PR #1077, opened 2021-05-05
Client should properly validate match commit checksum upon match_proof message.

LasTshaMAN:
I didn't try to deploy DEX locally and test this change manually (yet), do we have some basic automatic smoke tests that would uncover issues (if any) with changes similar to these ?

Closes https://github.com/decred/dcrdex/issues/1020.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -117,7 +117,8 @@ type dexConnection struct {
 
 	// tradeMtx is used to synchronize access to the trades map.
 	tradeMtx sync.RWMutex
-	trades   map[order.OrderID]*trackedTrade
+	// trades tracks outstanding orders issued by this client.
+	trades map[order.OrderID]*trackedTrade
 
 	epochMtx sync.RWMutex
 	epoch    map[string]uint64
@@ -4648,7 +4649,30 @@ func handleMatchProofMsg(c *Core, dc *dexConnection, msg *msgjson.Message) error
 			note.MarketID)
 	}
 
-	return book.ValidateMatchProof(note)
+	err = book.ValidateMatchProof(note)
+	if err != nil {
+		return fmt.Errorf("match proof validation failed: %w", err)
+	}
+
+	// Validate server match checksum for client orders (trades) in this epoch.
+	for _, trade := range dc.trackedTrades() {
+		// TODO - think about len(trade.csum) > 0 fits with the rest of these conditions, and at least document the reason for having it.
+		if trade.metaData.Status == order.OrderStatusEpoch && len(trade.csum) > 0 && !bytes.Equal(note.CSum, trade.csum) {
+			// Validation can fail either due to server trying to cheat
+			// (by seeing preimage before committing to csum for a match), or
+			// client loosing (e.g. due to restarting, since we don't
+			// persistently store these at the moment) his csums for the trades
+			// involved.
+			//
+			// Just warning the user for now, later on we might wanna revoke the order if this happens.
+			c.log.Warnf("validating match checksum for epoch %d, "+
+				"match commit checksum %s is not of the same value as "+
+				"client's order commit checksum %s, order ID: %s",
+				note.Epoch, note.CSum, trade.csum, trade.ID())
+		}
+	}
+
+	return nil
 }
 
 // handleRevokeOrderMsg is called when a revoke_order message is received.
@@ -4987,7 +5011,7 @@ func handlePreimageRequest(c *Core, dc *dexConnection, msg *msgjson.Message) err
 		go func() {
 			// Order request success OR fail closes the channel.
 			<-commitSig
-			if err := processPreimageRequest(c, dc, msg.ID, oid); err != nil {
+			if err := processPreimageRequest(c, dc, msg.ID, oid, req.CommitChecksum); err != nil {
 				c.log.Errorf("async processPreimageRequest for %v failed: %v", oid, err)
 			} else {
 				c.log.Debugf("async processPreimageRequest for %v succeeded", oid)
@@ -5013,7 +5037,7 @@ func handlePreimageRequest(c *Core, dc *dexConnection, msg *msgjson.Message) err
 		delete(c.piSyncers, oid)
 		c.piSyncMtx.Unlock()
 		dc.log.Debugf("Received preimage request for known order %v", oid)
-		return processPreimageRequest(c, dc, msg.ID, oid)
+		return processPreimageRequest(c, dc, msg.ID, oid, req.CommitChecksum)
 	}
 
 	// If we didn't find an entry, Trade or Cancel is still running. Add a chan
@@ -5030,7 +5054,7 @@ func handlePreimageRequest(c *Core, dc *dexConnection, msg *msgjson.Message) err
 	go func() {
 		select {
 		case <-syncChan:
-			if err := processPreimageRequest(c, dc, msg.ID, oid); err != nil {
+			if err := processPreimageRequest(c, dc, msg.ID, oid, req.CommitChecksum); err != nil {
 				c.log.Errorf("async processPreimageRequest for %v failed: %v", oid, err)
 			} else {
 				c.log.Debugf("async processPreimageRequest for %v succeeded", oid)
@@ -5049,11 +5073,14 @@ func handlePreimageRequest(c *Core, dc *dexConnection, msg *msgjson.Message) err
 	return nil
 }
 
-func processPreimageRequest(c *Core, dc *dexConnection, reqID uint64, oid order.OrderID) error {
+func processPreimageRequest(c *Core, dc *dexConnection, reqID uint64, oid order.OrderID, commitChecksum dex.Bytes) error {
 	tracker, preImg, isCancel := dc.findOrder(oid)
 	if tracker == nil {
 		return fmt.Errorf("no active order found for preimage request for %s", oid)
 	}
+
+	tracker.csum = commitChecksum
+
 	// Clean up the sentCommits now that we loaded the commitment. This can be
 	// removed when the old piSyncers method is removed.
 	defer func() {
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -4654,21 +4654,22 @@ func handleMatchProofMsg(c *Core, dc *dexConnection, msg *msgjson.Message) error
 		return fmt.Errorf("match proof validation failed: %w", err)
 	}
 
-	// Validate server match checksum for client orders (trades) in this epoch.
+	// Validate match_proof commitment checksum for client orders in this epoch.
 	for _, trade := range dc.trackedTrades() {
-		// TODO - think about len(trade.csum) > 0 fits with the rest of these conditions, and at least document the reason for having it.
-		if trade.metaData.Status == order.OrderStatusEpoch && len(trade.csum) > 0 && !bytes.Equal(note.CSum, trade.csum) {
-			// Validation can fail either due to server trying to cheat
-			// (by seeing preimage before committing to csum for a match), or
-			// client loosing (e.g. due to restarting, since we don't
-			// persistently store these at the moment) his csums for the trades
-			// involved.
-			//
-			// Just warning the user for now, later on we might wanna revoke the order if this happens.
-			c.log.Warnf("validating match checksum for epoch %d, "+
-				"match commit checksum %s is not of the same value as "+
-				"client's order commit checksum %s, order ID: %s",
-				note.Epoch, note.CSum, trade.csum, trade.ID())
+		if note.MarketID != trade.mktID {
+			continue
+		}
+
+		// Validation can fail either due to server trying to cheat (by
+		// requesting a preimage before closing the epoch to more orders), or
+		// client losing trades' epoch csums (e.g. due to restarting, since we
+		// don't persistently store these at the moment).
+		//
+		// Just warning the user for now, later on we might wanna revoke the
+		// order if this happens.
+		if err = trade.verifyCSum(note.CSum, note.Epoch); err != nil {
+			c.log.Warnf("Failed to validate commitment checksum for %s epoch %d at %s: %v"+
+				note.MarketID, note.Epoch, dc.acct.host, err)
 		}
 	}
 
@@ -5079,7 +5080,17 @@ func processPreimageRequest(c *Core, dc *dexConnection, reqID uint64, oid order.
 		return fmt.Errorf("no active order found for preimage request for %s", oid)
 	}
 
-	tracker.csum = commitChecksum
+	// Record the commitment checksum so we can verify that the subsequent
+	// match_proof with this order has the same checksum. If it does not, the
+	// server may have used the knowledge of this preimage we are sending them
+	// now to alter the epoch shuffle.
+	tracker.mtx.Lock()
+	if isCancel {
+		tracker.cancel.csum = commitChecksum
+	} else {
+		tracker.csum = commitChecksum
+	}
+	tracker.mtx.Unlock()
 
 	// Clean up the sentCommits now that we loaded the commitment. This can be
 	// removed when the old piSyncers method is removed.
--- a/client/core/core_test.go
+++ b/client/core/core_test.go
@@ -2535,6 +2535,10 @@ func TestHandlePreimageRequest(t *testing.T) {
 		)
 	}
 
+	// TODO
+	// We shouldn't allow for a duplicate request to contain different csum (test against that),
+	// see https://github.com/decred/dcrdex/pull/1077#discussion_r641593845 for details.
+
 	// negative paths
 	ensureErr := func(tag string, req *msgjson.Message, errPrefix string) {
 		t.Helper()
--- a/client/core/core_test.go
+++ b/client/core/core_test.go
@@ -2498,8 +2498,8 @@ func TestHandlePreimageRequest(t *testing.T) {
 		commit := preImg.Commit()
 		commitSig := readyCommitment(commit)
 		payload = &msgjson.PreimageRequest{
-			OrderID:        oid[:],
-			Commitment:     commit[:],
+			OrderID:    oid[:],
+			Commitment: commit[:],
 		}
 		reqCommit, _ := msgjson.NewRequest(rig.dc.NextID(), msgjson.PreimageRoute, payload)
 
--- a/client/core/core_test.go
+++ b/client/core/core_test.go
@@ -2495,10 +2495,12 @@ func TestHandlePreimageRequest(t *testing.T) {
 	}
 
 	commit := preImg.Commit()
+	commitCSum := dex.Bytes{2, 3, 5, 7, 11, 13}
 	commitSig := readyCommitment(commit)
 	payload = &msgjson.PreimageRequest{
-		OrderID:    oid[:],
-		Commitment: commit[:],
+		OrderID:        oid[:],
+		Commitment:     commit[:],
+		CommitChecksum: commitCSum,
 	}
 	reqCommit, _ := msgjson.NewRequest(rig.dc.NextID(), msgjson.PreimageRoute, payload)
 
@@ -2509,6 +2511,7 @@ func TestHandlePreimageRequest(t *testing.T) {
 	if err != nil {
 		t.Fatalf("handlePreimageRequest error: %v", err)
 	}
+
 	// It has gone async now, waiting for commitSig.
 	// i.e. "Received preimage request for %v with no corresponding order submission response! Waiting..."
 	close(commitSig) // pretend like the order submission just finished
@@ -2520,6 +2523,16 @@ func TestHandlePreimageRequest(t *testing.T) {
 		}
 	case <-time.After(time.Second):
 		t.Fatal("no order note from preimage request handling")
+	}
+
+	// tracker might be modified in different go-routines from the one running here,
+	// due to the usage of channel above its state should already be synchronized here.
+	if !bytes.Equal(commitCSum, tracker.csum) {
+		t.Fatalf(
+			"handlePreimageRequest must initialize tracker csum, exp: %s, got: %s",
+			commitCSum,
+			tracker.csum,
+		)
 	}
 
 	// negative paths
@@ -4334,8 +4347,7 @@ func TestHandleMatchProofMsg(t *testing.T) {
 
 	req, _ := msgjson.NewRequest(rig.dc.NextID(), msgjson.MatchProofRoute, payload)
 
-	// Ensure match proof validation generates an error for a non-existent
-	// orderbook generates an error.
+	// Ensure match proof validation generates an error for a non-existent orderbook.
 	err = handleMatchProofMsg(rig.core, rig.dc, req)
 	if err == nil {
 		t.Fatal("[handleMatchProofMsg] expected a non-existent orderbook error")
@@ -4488,7 +4500,7 @@ func TestSetEpoch(t *testing.T) {
 		t.Fatalf("error advancing epoch: %v", err)
 	}
 	if mktEpoch() != 2 {
-		t.Fatalf("expected epoch 1, got %d", mktEpoch())
+		t.Fatalf("expected epoch 2, got %d", mktEpoch())
 	}
 
 	payload.Epoch = 0
@@ -4498,7 +4510,7 @@ func TestSetEpoch(t *testing.T) {
 		t.Fatalf("error handling match proof: %v", err)
 	}
 	if mktEpoch() != 2 {
-		t.Fatalf("epoch decremented")
+		t.Fatalf("epoch changed, expected epoch 2, got %d", mktEpoch())
 	}
 }
 
--- a/client/core/core_test.go
+++ b/client/core/core_test.go
@@ -2451,128 +2451,267 @@ func TestCancel(t *testing.T) {
 }
 
 func TestHandlePreimageRequest(t *testing.T) {
-	rig := newTestRig()
-	defer rig.shutdown()
-	ord := &order.LimitOrder{P: order.Prefix{ServerTime: time.Now()}}
-	oid := ord.ID()
-	preImg := newPreimage()
-	payload := &msgjson.PreimageRequest{
-		OrderID: oid[:],
-		// No commitment in this request.
-	}
-	reqNoCommit, _ := msgjson.NewRequest(rig.dc.NextID(), msgjson.PreimageRoute, payload)
-
-	tracker := &trackedTrade{
-		Order:    ord,
-		preImg:   preImg,
-		mktID:    tDcrBtcMktName,
-		db:       rig.db,
-		dc:       rig.dc,
-		metaData: &db.OrderMetaData{},
-	}
-
-	// Simulate an order submission request having completed.
-	loadSyncer := func() {
-		rig.core.piSyncMtx.Lock()
-		rig.core.piSyncers[oid] = nil // set nil to ensure it's just a map entry
-		rig.core.piSyncMtx.Unlock()
-	}
-
-	rig.dc.trades[oid] = tracker
-	loadSyncer()
-	err := handlePreimageRequest(rig.core, rig.dc, reqNoCommit)
-	if err != nil {
-		t.Fatalf("handlePreimageRequest error: %v", err)
-	}
-
-	// Test the new path with rig.core.sentCommits.
-	readyCommitment := func(commit order.Commitment) chan struct{} {
-		commitSig := make(chan struct{}) // close after fake order submission is "done"
-		rig.core.sentCommitsMtx.Lock()
-		rig.core.sentCommits[commit] = commitSig
-		rig.core.sentCommitsMtx.Unlock()
-		return commitSig
-	}
-
-	commit := preImg.Commit()
-	commitCSum := dex.Bytes{2, 3, 5, 7, 11, 13}
-	commitSig := readyCommitment(commit)
-	payload = &msgjson.PreimageRequest{
-		OrderID:        oid[:],
-		Commitment:     commit[:],
-		CommitChecksum: commitCSum,
-	}
-	reqCommit, _ := msgjson.NewRequest(rig.dc.NextID(), msgjson.PreimageRoute, payload)
-
-	notes := rig.core.NotificationFeed()
-
-	rig.dc.trades[oid] = tracker
-	err = handlePreimageRequest(rig.core, rig.dc, reqCommit)
-	if err != nil {
-		t.Fatalf("handlePreimageRequest error: %v", err)
-	}
-
-	// It has gone async now, waiting for commitSig.
-	// i.e. "Received preimage request for %v with no corresponding order submission response! Waiting..."
-	close(commitSig) // pretend like the order submission just finished
-
-	select {
-	case note := <-notes:
-		if note.Subject() != SubjectPreimageSent {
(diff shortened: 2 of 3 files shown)


client/core/trade.go
```diff
@@ -182,14 +184,6 @@ func newTrackedTrade(dbOrder *db.MetaOrder, preImg order.Preimage, dc *dexConnec
 	return t
 }
 
-// rate returns the order's rate, or zero if a market or cancel order.
-func (t *trackedTrade) rate() uint64 {
```
LasTshaMAN, line 186: this is dead code, so, removing it

dex/msgjson/types.go
```diff
@@ -828,7 +829,7 @@ type OrderBook struct {
 type MatchProofNote struct {
 	MarketID  string  `json:"marketid"`
 	Epoch     uint64  `json:"epoch"`
-	Preimages []Bytes `json:"preimages"`
+	Preimages []order.Preimage `json:"preimages"`
```
LasTshaMAN, line 832: Is it okay to change this type ? Changing this type seems only natural to me, to avoid unnecessary type-conversions when finding relevant tracked orders in `validateMatchCsum` func.

chappjc: Depends on how it marshals/unmarshals to JSON.  I don't believe it implements either json.Marshaller or Unmarshaller, so I suspect it's going to be an array of integers as a result instead of hex.

We lack a TestMatchProof in msg_test, but you could make one to see if it marshals and unmarshals as expected.

```go
func TestMatchProof(t *testing.T) {
	pimgB := randomBytes(32)
	var pimg order.Preimage
	copy(pimg[:], pimgB)
	match := &MatchProofNote{
		Preimages: []order.Preimage{pimg},
	}

	matchB, err := json.Marshal(match)
	if err != nil {
		t.Fatalf("marshal error: %v", err)
	}

	t.Logf("%v", string(matchB))

    // check things here

	var matchBack Match
	err = json.Unmarshal(matchB, &matchBack)
	if err != nil {
		t.Fatalf("unmarshal error: %v", err)
	}

    // check more things here
}
```

> === RUN   TestMatchProof
>     /home/jon/github/decred/dcrdex/dex/msgjson/msg_test.go:83: {"marketid":"","epoch":0,"preimages":[[82,253,252,7,33,130,101,79,22,63,95,15,154,98,29,114,149,102,199,77,16,3,124,77,123,187,4,7,209,226,198,73]],"misses":null,"csum":"","seed":""}
> 

With `[]Bytes`:

> === RUN   TestMatchProof
>     /home/jon/github/decred/dcrdex/dex/msgjson/msg_test.go:80: {"marketid":"","epoch":0,"preimages":["52fdfc072182654f163f5f0f9a621d729566c74d10037c4d7bbb0407d1e2c649"],"misses":null,"csum":"","seed":""}

client/core/trade.go
chappjc: My linters have caught that too.  Thanks.

client/core/core_test.go
```diff
@@ -2518,15 +2518,17 @@ func TestHandlePreimageRequest(t *testing.T) {
...
+	// we need to properly synchronize any read for its fields.
+	tracker.mtx.RLock()
 	if !bytes.Equal(commitCSum, tracker.csum) {
 		t.Fatalf(
 			"handlePreimageRequest must initialize tracker csum, exp: %s, got: %s",
 			commitCSum,
 			tracker.csum,
 		)
 	}
+	tracker.mtx.RUnlock()
```
chappjc, line 2531: Deadlock if it does hit t.Fatal.

client/core/core.go
```diff
@@ -4973,11 +5023,16 @@ func handlePreimageRequest(c *Core, dc *dexConnection, msg *msgjson.Message) err
...
-func processPreimageRequest(c *Core, dc *dexConnection, reqID uint64, oid order.OrderID) error {
+func processPreimageRequest(c *Core, dc *dexConnection, reqID uint64, oid order.OrderID, commitChecksum dex.Bytes) error {
 	tracker, preImg, isCancel := dc.findOrder(oid)
 	if tracker == nil {
 		return fmt.Errorf("no active order found for preimage request for %s", oid)
 	}
+
+	tracker.mtx.Lock()
+	tracker.csum = commitChecksum
+	tracker.mtx.Unlock()
```
chappjc, line 5034: Changes to product code just to facilitate test, particularly blocking calls, is to be avoided like the plague.

If the test gets the `SubjectPreimageSent` note emitted below, we can be sure that this code has already executed and there's no risk of a data race.  Was a race detected?

client/core/core.go
```diff
@@ -900,10 +900,12 @@ func (c *Core) connectedDEX(addr string) (*dexConnection, error) {
...
 
-// setEpoch sets the epoch. If the passed epoch is greater than the highest
-// previously passed epoch, an epoch notification is sent to all subscribers and
-// true is returned.
-func (dc *dexConnection) setEpoch(mktID string, epochIdx uint64) bool {
+// tryUpdateEpoch updates the epoch for market with mktID to epochIdx,
+// sends an epoch notification to all subscribers
+// and returns true only
+// if the passed epoch is greater than the highest previously observed epoch.
+// Otherwise tryUpdateEpoch returns false.
```
chappjc, line 907: I know this gripe seems silly, but it's really not an burden to just setup your editor with a key combo to automatically wrap selected text.  It takes virtually no thought, it makes docs look consistent, and we all do it unless the content of the comment demands an exception.  This looks super janky right now.

JoeGruffins: I just use `:set colorcolumn=80` with vim and do it manually though that's a little "neckbeard"ish.

client/core/core.go
LasTshaMAN: > Changes to product code just to facilitate tests

I agree, but this has nothing to do with tests, 

`tracker.csum` gets written to in one request (go-routine 1), while we a reading them in another request (go-routine 2),

maybe due to the time difference between these 2 requests it is safe to assume that `tracker.csum` will be synchronized, but its not 100% the case, so I'm just playing it safe here,
if this causes any performance issues, then we can probably omit locking here,

> Was a race detected?

nope, there is syncronization on channel that prevents race, given its current implementation

client/core/core.go
```diff
@@ -900,10 +900,12 @@ func (c *Core) connectedDEX(addr string) (*dexConnection, error) {
...
+func (dc *dexConnection) tryUpdateEpoch(mktID string, epochIdx uint64) bool {
```
buck54321, line 908: This name change adds nothing and is completely unrelated to the task at hand. In general, do not change anything except what's necessary. We will sometimes accept an un-related or stylistic change as part of PR, but not until you've established that you can keep your work properly scoped.

client/core/core.go
```diff
@@ -4563,16 +4565,64 @@ func handleMatchProofMsg(c *Core, dc *dexConnection, msg *msgjson.Message) error
...
+// commited checksums on corresponding orders (trades)
+// against csum for this match, making sure they are of the same value.
+//
+// This is necessary to ensure that for an order O
+// "server preimage request" and "O being matched" happen in the same epoch,
+// otherwise server could game match shuffle, if it sees O's preimage
+// before committing to csum for a match.
+func validateMatchCSum(preimages []order.Preimage, matchCSum dex.Bytes, trades []*trackedTrade) error {
+	for _, preimage := range preimages {
+		err := validateMatchCSumForPreimage(preimage, matchCSum, trades)
```
buck54321, line 4605: This is going to error virtually every time. The server is sending every preimage of every order from the epoch, not just our own.

client/core/core.go
```diff
@@ -4563,16 +4565,64 @@ func handleMatchProofMsg(c *Core, dc *dexConnection, msg *msgjson.Message) error
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func validateMatchCSumForPreimage(preimage order.Preimage, matchCSum dex.Bytes, trades []*trackedTrade) error {
+	for _, trade := range trades {
+		if preimage == trade.preImg {
```
buck54321, line 4616: Don't check every trade, only those in `OrderStatusEpoch` for which we've received a preimage request.

client/core/core.go
```diff
@@ -4563,16 +4565,64 @@ func handleMatchProofMsg(c *Core, dc *dexConnection, msg *msgjson.Message) error
...
 			note.MarketID)
 	}
 
-	return book.ValidateMatchProof(note)
+	err = book.ValidateMatchProof(note)
+	if err != nil {
+		return fmt.Errorf("match proof validation failed: %w", err)
+	}
+
+	err = validateMatchCSum(note.Preimages, note.CSum, dc.trackedTrades())
```
buck54321, line 4582: The use of the two helper methods is overkill. Just iterate here.

```
for _, trade := range dc.trackedTrades() {
    if trade.metaData.Status == order.OrderStatusEpoch && len(trade.csum) > 0 && !bytes.Equal(note.CSum, trade.csum)
        // Probably self-revoke the order so we don't acknowledge match requests. Also log the error.
    }
}
```

buck54321 (changes requested):

client/core/core.go
LasTshaMAN: Changed formatting for this section.

I'm formatting everything manually at the moment the way that seems to be the most readable to me, but looks like it's not working as expected :)

Long-term, I should probably setup something to help me. Other comments look OK ?

client/core/core_test.go
LasTshaMAN: true, although I believe `TestHandlePreimageRequest` to be a pretty simple isolated test case with nothing vital here being locked forever,

when we hit `t.Fatal` here, all the related stuff around `tracker.mtx` becomes useless,

but maybe it is better to just copy value, re-wrote this part

client/core/core.go
LasTshaMAN: hmm, good point, haven't realized that `trackedTrade` represents an order issued by "this client only" and not just any order with additional meta,

shall I modify a description for `trackedTrade` ?

```
// trackedTrade is an order, its matches, and its cancel order, if applicable.
// The trackedTrade has methods for handling requests from the DEX to progress
// match negotiation.
```
to this
```
// trackedTrade is an order (issued by this client), its matches, and its cancel
// order, if applicable. The trackedTrade has methods for handling requests
// from the DEX to progress match negotiation.
```

as well as add a comment for `dexConnection.trades` to clarify that ?

client/core/core.go
LasTshaMAN: Right, should I remove this change ?

client/core/core.go
LasTshaMAN: What's the purpose for this check `len(trade.csum) > 0` ? 

My understanding is - if `len(trade.csum) == 0` this validation must fail as well.

client/core/core.go
LasTshaMAN: fixed.

client/core/core.go
chappjc: We shouldn't roll the dice with data races based on general ideas about expected timing, no.

I'm commenting about the fact that in the path in `handleMatchProofMsg` you aren't locking the mutex for the `csum` read, only in the test and here in `processPreimageRequest`.  Hence it looks like you've set up the locking for the test.

But here's the sequence:
- when an epoch closes, server computes the csum from all order commitments in that epoch, and it sends preimage requests to clients for all those orders
- client gets a preimage request, starts `processPreimageRequest`, which records the csum that came with the request and **then** responds to the server with the preimage. So server only has the preimage if the csum is recorded.
- when the server gets all the epoch preimages (or times out on any of them), it performs matching with those it has preimages for
- server sends a match_proof note, which contains the preimages **that it received (obviously)** and the csum it computed from all the epoch order commitments
- `handleMatchProofMsg` looks (or should) for orders with the corresponding preimages.  If the a known preimage is found, that's because `processPreimageRequest` has completed already.

So if the preimage for the order is in the match_proof note, that means the server has received it from `processPreimageRequest` and the tracker.csum has already been saved.  No race.

client/core/core.go
```diff
@@ -4563,16 +4565,40 @@ func handleMatchProofMsg(c *Core, dc *dexConnection, msg *msgjson.Message) error
...
+			// client loosing (e.g. due to restarting, since we don't
+			// persistently store these at the moment) his csums for the trades
+			// involved.
+			c.log.Warnf("validating match checksum for epoch %d, "+
+				"match commit checksum %s is not of the same value as "+
+				"client's order commit checksum %s, order ID: %s",
+				note.Epoch, note.CSum, trade.csum, trade.ID())
+
+			// Client doesn't want to proceed with the match for this order.
+			trade.revoke()
```
chappjc, line 4597: To start I only want to warn.  We've done nothing so far, so we should at least use this in production with just a warning until we're sure we're not going to be bailing out on orders incorrectly, which will cause swaps to fail and users to be penalized (account suspended)

client/core/core.go
chappjc: Yeah, pls leave `setEpoch` and it's docs as-is. Thanks.

client/core/core.go
```diff
@@ -4563,16 +4565,40 @@ func handleMatchProofMsg(c *Core, dc *dexConnection, msg *msgjson.Message) error
...
-	return book.ValidateMatchProof(note)
+	err = book.ValidateMatchProof(note)
+	if err != nil {
+		return fmt.Errorf("match proof validation failed: %w", err)
+	}
+
+	// Validate server match checksum for client orders (trades) in this epoch.
+	for _, trade := range dc.trackedTrades() {
+		// TODO - len chkeck, why ?
+		if trade.metaData.Status == order.OrderStatusEpoch && len(trade.csum) > 0 && !bytes.Equal(note.CSum, trade.csum) {
```
chappjc, line 4585: Since epochs are continuous and there is time between epoch close and this match_proof msg (up to 20 sec for the preimage request timeout + network latency, although the spec incorrectly says 5 sec), there can be trades in `OrderStatusEpoch` from more than one epoch (different csums).

I'm afraid the `trackedTrade` doesn't know what epoch index it belongs to (sadly, server's `OrderResponse` doesn't include that), so finding the order with the corresponding preimage may be needed after all. However, that facilitates the lock-free solution to the `csum` data race issue.  I may be overlooking something...  @buck54321 I'm thinking cycling over preimages is gonna be needed after all, although I agree it doesn't quite warrant 2 helper functions.

As an aside, I'm really not sure if `dc.epoch` (and `setEpoch`) serves a purpose anymore, but let's not go down that rabbit how right now.  There are relics here and there we should note, but not change too readily.

client/core/core_test.go
chappjc: I get what you're saying, but it could hang at 

```
defer rig.shutdown()
````

`Core` might be unable to stop depending on what it's doing at the time.  I've had tests hang on t.Fatals before and it's very confusing so best to play it safe.  Thanks for changing.

File: client/core/core.go (hunk 8 above)
Comment (chappjc) [reply client/core/core.go]: I commented in https://github.com/decred/dcrdex/pull/1077#discussion_r627812050, but I actually think we have to look at all of our trades' preimages because OrderStatusEpoch orders can be from multiple epochs.  If trackedTrade knew which epoch index it was in, that would work too, alas it does not know.

File: client/core/core.go (hunk 9 above)
Comment (chappjc) [reply client/core/core.go]: That would imply dexc has restarted and csum was not restored (a rare event as restarting with epoch orders is a super dumb idea), or the trade is from a different epoch and a preimage request has not been received for that trade.  Related to my other comments regarding orders in OrderStatusEpoch potentially spanning multiple epochs.

File: client/core/core.go [hunk 12]
```diff
@@ -117,7 +117,8 @@ type dexConnection struct {
 
 	// tradeMtx is used to synchronize access to the trades map.
 	tradeMtx sync.RWMutex
-	trades   map[order.OrderID]*trackedTrade
+	// trades tracks outstanding orders issued by this client
```
JoeGruffins, line 120: Srry for nitpick, but period at the end. I treat `trades` here as a proper noun although not capitalized, so grammatically correct to call it a sentence, with that one exception. I think.

client/core/core.go
JoeGruffins: Either way is fine with me... but another pr, even for just this one change is probably better.

client/core/core.go
```diff
@@ -4563,16 +4565,40 @@ func handleMatchProofMsg(c *Core, dc *dexConnection, msg *msgjson.Message) error
 		return fmt.Errorf("match proof note unmarshal error: %w", err)
 	}
 
-	// Expire the epoch
-	dc.setEpoch(note.MarketID, note.Epoch+1)
+	// Expire previous epoch.
```
JoeGruffins, line 4568: I do like the period here!

JoeGruffins (commented):
Running the simnet tests:
```
2021-05-07 09:54:15.016 [INF] MKT: Book router terminating for market "dcr_btc"
panic: runtime error: index out of range [0] with length 0

goroutine 135 [running]:
decred.org/dcrdex/server/market.(*BookRouter).runBook(0xc0005af050, 0xd60df8, 0xc000576d00, 0xc000347500)
        /home/joe/git/dcrdex/server/market/bookrouter.go:455 +0x7d5
decred.org/dcrdex/server/market.(*BookRouter).Run.func1(0xc0005af050, 0xd60df8, 0xc000576d00, 0xc0000256d0, 0xc000347500)
        /home/joe/git/dcrdex/server/market/bookrouter.go:312 +0x49
created by decred.org/dcrdex/server/market.(*BookRouter).Run
        /home/joe/git/dcrdex/server/market/bookrouter.go:311 +0x10e
```

client/core/core.go
LasTshaMAN: Okay, I see ... Turns out `preimage` and `match_proof` requests are both websocket requests that get through the same channel (thus it makes any inmemory state between them synchronized).

So, there is no need to synchronize explicitely.

Leaving here my flawed (due to lack of context) reasoning just for completeness.

<details>
<summary>my flawed reasoning</summary>
<p>

I understand the sequence you've described, but not sure what you mean by "saved" (`trackedTrade`s reside in process memory between `preimage` and `match_proof` requests) in "and the tracker.csum has already been saved. No race.".

> I'm commenting about the fact that in the path in handleMatchProofMsg you aren't locking the mutex for the csum read, only in the test and here in processPreimageRequest.

That's a very good point, and I didn't look at it from this perspective, let me clarify.

The way it is implemented right now, it looks indeed as if locking on `tracker.mtx` here is done purely for testing purposes. But it not the intent. 
The intent is - for any other go-rotine G that is interested in reading up-to-date `tracker.csum` value (or modifying it), the developer implementing G will only have to worry about aquiring/releasing this lock in G properly, and not follow through the code checking whether `tracker.csum` was written to with this lock being used or not.
This is the contract that, I assumed, everyone should follow when working with `trackedTrade` fields, protected by this lock (based on its description):
```go
	// mtx protects all read-write fields of the trackedTrade and the
	// matchTrackers in the matches map.
	mtx      sync.RWMutex
```
To summarize, we aquire this lock here in order to not shoot ourselves in the foot later on when implementing features that depend on `tracker.csum` value. 
Whether this opens a whole lot of different issues to worry about, such as deadlocks or bad performance, is another matter also worthy to think through.

Now, back to `handleMatchProofMsg`, why didn't I lock on `tracker.mtx` here ? This is my bad indeed, looks like we need to lock it there as well ...
</p>
</details>

dex/msgjson/types.go
LasTshaMAN: Rolled back type change.

client/core/core.go
LasTshaMAN: Will move this to a separate PR.

client/core/core.go
LasTshaMAN: Okay, warning it is for now.

LasTshaMAN:
> Running the simnet tests:
> 
> ```
> 2021-05-07 09:54:15.016 [INF] MKT: Book router terminating for market "dcr_btc"
> panic: runtime error: index out of range [0] with length 0
> 
> goroutine 135 [running]:
> decred.org/dcrdex/server/market.(*BookRouter).runBook(0xc0005af050, 0xd60df8, 0xc000576d00, 0xc000347500)
>         /home/joe/git/dcrdex/server/market/bookrouter.go:455 +0x7d5
> decred.org/dcrdex/server/market.(*BookRouter).Run.func1(0xc0005af050, 0xd60df8, 0xc000576d00, 0xc0000256d0, 0xc000347500)
>         /home/joe/git/dcrdex/server/market/bookrouter.go:312 +0x49
> created by decred.org/dcrdex/server/market.(*BookRouter).Run
>         /home/joe/git/dcrdex/server/market/bookrouter.go:311 +0x10e
> ```

This is due to a bug (with slice creation) I've introduced when changing `Bytes` -> `order.Preimage` type, as disscussed here - https://github.com/decred/dcrdex/pull/1077#discussion_r626638772

image

rolled back this type change, hopefully its fixed now.

LasTshaMAN:
I've re-worked this PR trying to address all the comments, and closed off all discussions that became irrelevant due to all the latest changes.

Waiting for @buck54321's response to @chappjc's concerns about `server epoch` vs `order.OrderStatusEpoch` mismatch before modifying anything further.

client/core/core.go
buck54321: No need. It doesn't improve clarity or readability, and even if it did, it's not something to spend time on right now.

client/core/core.go
chappjc: I'm honestly pleased that you're attuned to the need to synchronize with concurrent writes, but no need to go into it much - we're all on the same page about the need for that.

Yes, there is implicit synchronization that prevents concurrent writes, but it's not because "both websocket requests that get through the same channel" if I understand your meaning.  It's because a match_proof request can not include the preimage if `processPreimageRequest` has not already "saved" the csum (`tracker.csum = commitChecksum`). The other goroutine that reads `tracker.csum` will only attempt that read if the server knew about that preimage to include it with the match_proof note, which means it has already been stored in `tracker.csum`.  This all assumes the first iteration of these changes where preimages were checked to identify relevant trades.


Now a comment about the following:

> developer implementing G will only have to worry about aquiring/releasing this lock in G properly, and not follow through the code checking whether tracker.csum was written to with this lock being used or not.

You see this as being safe, but in reality not understanding access patterns and instead falling back to just locking mutexes liberally is seriously problematic.  When possible, it is generally wiser to design access patterns that do not involve the possibility for concurrent writes.  You'll have to believe me when I say it can become very messy with functions that maybe lock and maybe don't, called from other methods that do lock... so is it gonna deadlock or is it gonna be a race or is it gonna be correct?  This is particularly easy to get into with `trackedTrade.RWMutex` given how it's used so extensively in Core.

In short, the `csum` field should not be under the `mtx` in the struct definition because: it will only ever be accessed AFTER it is written, if the access patterns design for that.

chappjc: Although I've neglected the possibility of a malfunctioning server sending multiple preimage requests for the same order/commitment... :sigh:  I'm considerably less confident that we'd not end up with two concurrent writes (two `processPreimageRequest`).  We'd also want logic to reject such requests if `csum` is already stored.  Let's first see how this plays out after buck weighs in on my assertion that there can be epoch status orders for multiple epochs.

We might end up needing a sync mechanism for the `csum` field after all.  If that is the case we can consider it's own mutex or something more low-level like an atomic uint32 CAS that just says if it has been stored already so that we don't have to worry about an locking at the read sites.

client/core/core.go
buck54321: > If trackedTrade knew which epoch index it was in

As long as we have the `order.Prefix.ServerTime` and the `msgjson.Market.EpochLen`, we have the epoch index, and we can match against `msgjson.MatchProofNote.Epoch`.

buck54321: Yep, which can simplify the filter too.

```
epochIdx := encode.UnixMilliU(trade.Prefix().ServerTime) / trade.epochLen
if note.Epoch == epochIdx && note.MarketID == trade.mktID {
    // do stuff
}
```

Note the addition of the market ID comparison, which I had failed to notice before.

client/core/core.go
LasTshaMAN: Thanks to the detailed explanations above, I think now we are on the same page regarding:
- related concurrency considerations
- we must assume the server is maliscious

Below are some conclusions I've arrived at, please let me know wheter any of these are incorrect.

1. Checking `len(trade.cSum) > 0` in `handleMatchProofMsg` is probably not sufficient, because maliscious server might just provide empty value for `CommitChecksum` in `handlePreimageRequest` resulting in client omitting `cSum` validation altogether.

2. I believe we shouldn't check for duplicate `handlePreimageRequest` requests from the server, instead we should apply every request (overwriting the previous `cSum` value) and just verifying the value matches should be enough to spot any maliscious behavior by the server. This looks like a simple enough solution, given (3).

3. 
> but that might warrant resetting it to 0 if the websocket Send just fails due to connectivity or whatever so a subsequent request can be fielded.

Not sure how to better handle that ... because it might be difficult to differentiate between "server playing games" & "connectivity issues".

JoeGruffins (approved):
Looks good to me.

client/core/core.go
```diff
@@ -4572,7 +4573,30 @@ func handleMatchProofMsg(c *Core, dc *dexConnection, msg *msgjson.Message) error
...
 
-	return book.ValidateMatchProof(note)
+	err = book.ValidateMatchProof(note)
+	if err != nil {
+		return fmt.Errorf("match proof validation failed: %w", err)
+	}
+
+	// Validate server match checksum for client orders (trades) in this epoch.
+	for _, trade := range dc.trackedTrades() {
+		// TODO - think about len(trade.csum) > 0 fits with the rest of these conditions, and at least document the reason for having it.
```
JoeGruffins, line 4583: Is it possible for `processPreimageRequest` to have not happened yet?

I don't see a check for `nil` in `processPreimageRequest`, unless I am missing a check somewhere else. Maybe server can send nil in the `PreimageRequest` request?

client/core/core.go
```diff
@@ -4911,7 +4935,7 @@ func handlePreimageRequest(c *Core, dc *dexConnection, msg *msgjson.Message) err
 		go func() {
 			// Order request success OR fail closes the channel.
 			<-commitSig
-			if err := processPreimageRequest(c, dc, msg.ID, oid); err != nil {
+			if err := processPreimageRequest(c, dc, msg.ID, oid, req.CommitChecksum); err != nil {
```
JoeGruffins, line 4938: I just tried irl and it seems passing a nil `req.CommitChecksum` is fine atm. Maybe we want to error (or just warn for now) before this if the check sum is nil or wrong length? I'm guess up around the check for `req.OrderID` length.

client/core/core.go
LasTshaMAN: Yes, I've mentioned that in my comment above (section 3), we should probably check against that,

https://github.com/decred/dcrdex/pull/1077#discussion_r628744407

JoeGruffins: Ah, somehow I missed that long conversation. Well, I don't know the best answer, but thank you for your time on this. imo coming along nicely!

client/core/core.go
LasTshaMAN: @chappjc @buck54321 any comments on my summary above ?

client/core/core.go
```diff
@@ -4973,11 +4997,14 @@ func handlePreimageRequest(c *Core, dc *dexConnection, msg *msgjson.Message) err
...
 
-func processPreimageRequest(c *Core, dc *dexConnection, reqID uint64, oid order.OrderID) error {
+func processPreimageRequest(c *Core, dc *dexConnection, reqID uint64, oid order.OrderID, commitChecksum dex.Bytes) error {
 	tracker, preImg, isCancel := dc.findOrder(oid)
 	if tracker == nil {
 		return fmt.Errorf("no active order found for preimage request for %s", oid)
 	}
+
+	tracker.csum = commitChecksum
+
```
chappjc, line 5007: We've forgotten about cancel orders, which are not `trackedTrades` themselves, but the `trackedTrade.cancel` field that links it with the targeted trade.  Please see the following suggested diff: https://github.com/decred/dcrdex/commit/cbf49a8512913b4683e611e2d1c272a068b9e64f

I'd also prefer the `trackedTrade` formatted like in that diff too if you don't mind applying that too.

LasTshaMAN: Okay, I've applied your diff, thanks! Looks like it addresses all the concerns I've outlined above - https://github.com/decred/dcrdex/pull/1077#discussion_r628744407

client/core/core.go
chappjc: > I believe we shouldn't check for duplicate handlePreimageRequest requests from the server, instead we should process every request (overwriting the previous csum value) and just verifying the value matches should be enough to spot any maliscious behavior by the server.

On the contrary, allowing duplicate requests to essentially update the commitment checksum is facilitating the server cheating.  Server would just have to send a final preimage request just before the match_proof ntfn with a csum that matches, thus defeating the check in handleMatchProof.

If the trade/cancel's `csum` is already set and not equal to what the repeated request is providing, that's definitely server cheating and should be handled as such.

chappjc: Also, if you could place a TODO comment in the test for handleMatchProof, we can get to that when we are ready to change these warnings into errors that prevent client from negotiating matches for such orders.

client/core/core.go
chappjc: I should just note as a follow-up to the other discussion in this PR w.r.t. mutex/atomic that I felt using the `trackedTrade.mtx` was alright in the end because we had to safely access the `cancel` field as well.

client/core/core.go
LasTshaMAN: Added a `TODO` comment about handling duplicates (just a comment, no code to handle that for now).

buck54321 (changes requested):
This is really looking good now. A little test coverage to round out, but otherwise seems ready to me. Thanks for sticking it out.

client/core/core_test.go
```diff
@@ -2522,6 +2525,20 @@ func TestHandlePreimageRequest(t *testing.T) {
...
+	// tracker might be modified in different go-routines from the one running here,
+	// due to the usage of channel above its state should already be synchronized here.
+	if !bytes.Equal(commitCSum, tracker.csum) {
+		t.Fatalf(
+			"handlePreimageRequest must initialize tracker csum, exp: %s, got: %s",
+			commitCSum,
+			tracker.csum,
+		)
+	}
+
```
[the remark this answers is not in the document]
LasTshaMAN: will do)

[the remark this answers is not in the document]
LasTshaMAN: Moved `csum` related checks into separate tests, since trying to tests everything in a single huge test is pretty tricky and unreadable.

chappjc (approved):
OK.

client/core/core.go
chappjc: I've approved the PR, but if @buck54321 has anything else that requires a commit, I'd really like to have `processPreimageRequest` return an error if the `tracker.{cancel.}csum` is non-nil.  Otherwise these changes offer no real protection yet, just a start in that direction.

buck54321 (approved):

client/core/core.go
LasTshaMAN: @chappjc I can create a follow-up PR to check for: `tracker.{cancel.}csum` is non-nil case.

PR #1078, opened 2021-05-07
client/asset/eth: Use gwei in Balance. 

### LINKED ISSUE #1076: eth: uint64 too small to store balances
Client assets use this to handle balances:
https://github.com/decred/dcrdex/blob/91d5a376592317ca60b76f488a194dca34c7ca2a/client/asset/interface.go#L185-L195

The max value for a uint64 (2^64) is ~1.8e19

The smallest unit of measure for eth is the [wei](https://ethdocs.org/en/latest/ether.html#denominations) and it takes 1e18 wei to make one eth. This means that using a uint64, the most a wallet can track is 18 eth.

We could do something complicated like bit shifting and using two uint64 which should give values up to 3.4e38 which I am guessing is safe.
We could have all balances as `*big.Int`.
We could round all eth values and store as 1e10 units?

Amounts are also stored in the db. I'm guessing we will need to make encoding asset specific. Maybe Balance itself can be made an interface and current coins can continue using `uint64` while eth uses `*big.Int`.

JoeGruffins:
closes #1076

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -439,7 +439,10 @@ func (eth *ExchangeWallet) SyncStatus() (bool, float32, error) {
 		return false, 0, err
 	}
 	if sp != nil {
-		ratio := float32(sp.CurrentBlock) / float32(sp.HighestBlock)
+		var ratio float32
+		if sp.HighestBlock != 0 {
+			ratio = float32(sp.CurrentBlock) / float32(sp.HighestBlock)
+		}
 		return false, ratio, nil
 	}
 	bh, err := eth.node.bestHeader(eth.ctx)
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -33,8 +33,8 @@ func init() {
 const (
 	// BipID is the BIP-0044 asset ID.
 	BipID              = 60
-	defaultGasFee      = 82_000_000_000
-	defaultGasFeeLimit = 200_000_000_000
+	defaultGasFee      = 82  // gwei
+	defaultGasFeeLimit = 200 // gwei
 )
 
 var (
@@ -49,7 +49,7 @@ var (
 			Description: "This is the highest network fee rate you are willing to " +
 				"pay on swap transactions. If gasfeelimit is lower than a market's " +
 				"maxfeerate, you will not be able to trade on that market with this " +
-				"wallet.  Units: wei",
+				"wallet.  Units: gwei / gas",
 			DefaultValue: defaultGasFeeLimit,
 		},
 		{
@@ -62,7 +62,7 @@ var (
 	// WalletInfo defines some general information about a Ethereum wallet.
 	WalletInfo = &asset.WalletInfo{
 		Name:              "Ethereum",
-		Units:             "wei",
+		Units:             "gwei",
 		DefaultConfigPath: defaultAppDir, // Incorrect if changed by user?
 		ConfigOpts:        configOpts,
 	}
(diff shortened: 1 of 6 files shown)


dex/testing/eth/harness.sh
```diff
@@ -78,6 +78,9 @@ cat > "${NODES_ROOT}/genesis.json" <<EOF
     "byzantiumBlock": 0,
     "constantinopleBlock": 0,
     "petersburgBlock": 0,
+    "istanbulBlock": 0,
+    "muirGlacierBlock": 0,
+    "berlinBlock": 0,
```
JoeGruffins, line 83: Should put us at the same.. fork as mainnet.

dex/testing/eth/harness.sh
```diff
@@ -183,9 +186,7 @@ echo "Sending 5000 eth to delta and gamma."
 
 # Our transactions will PROBABLY appear in one of these blocks.
 echo "Mining some blocks"
-"${NODES_ROOT}/harness-ctl/mine-beta" "1"
-"${NODES_ROOT}/harness-ctl/mine-alpha" "1"
-"${NODES_ROOT}/harness-ctl/mine-beta" "5"
+"${NODES_ROOT}/harness-ctl/mine-beta" "15"
```
JoeGruffins, line 189: I had a few times where it looked like mining on alpha made transaction inclusion speed worse. Will attempt to solve the problem in a separate pr.

JoeGruffins:
It is 18 billion eth possible correct?

client/asset/eth/eth.go
```diff
@@ -439,7 +439,10 @@ func (eth *ExchangeWallet) SyncStatus() (bool, float32, error) {
 		return false, 0, err
 	}
 	if sp != nil {
-		ratio := float32(sp.CurrentBlock) / float32(sp.HighestBlock)
+		var ratio float32
+		if sp.HighestBlock != 0 {
+			ratio = float32(sp.CurrentBlock) / float32(sp.HighestBlock)
+		}
```
JoeGruffins, line 445: Just in case.

chappjc:
> It is 18 billion eth possible correct?

You mean the ultimate supply?  I thought it was uncapped, but with 18 million ETH _per year_, so with about 117 million currently existing, I'd say even 1 billion is a ton

dex/testing/eth/harness.sh
chappjc: Dang what a drag. Linking back to the [original discussion](https://github.com/decred/dcrdex/pull/1005#discussion_r637244211) and karalabe's [comment in a different thread.](https://github.com/ethereum/go-ethereum/issues/2769#issuecomment-230249342)

chappjc: Another related idea was to have the non-miners (clients) connected to each other or at least another non-miner (or not the primary miner): https://github.com/ethereum/go-ethereum/issues/2822#issuecomment-240998150

> ivooz: The problem goes away when all the non-mining nodes are connected to each other, previously they only knew about miner nodes.

> karalabe: ... your first topology were that all nodes were connected only to the miners. This can be problematic as the miners are heavily loaded churning out blocks, so if everyone is bugging them to distribute the data too, I can imagine that they are simply overwhelmed (you essentially DDOS the miners).
> 
> I'd recommend enabling discovery (with a custom bootnode, maybe one of the miners, should be fine) so nodes can properly connect to each other and form a decent overlay network to propagate the data.

chappjc: Oh and about the first idea of starting 2 miners, one guy said he "updated the file "static-nodes.json" accordingly".  Know what we should do there?

JoeGruffins:
Oh no, I meant the amount of eth we can save in a uint64 as gwei. There is no eth max supply, and even if there were, they don't have a great track record of doing what that say. But 18 billion is still a far way off.

dex/testing/eth/harness.sh
JoeGruffins: Will look into this directly after I get #1019 up to speed.

chappjc:
Hah, ok!  Yup `math.MaxUint64 / 1e9` is 18,446,744,073.

dex/testing/eth/harness.sh
JoeGruffins: Yeah, still scratching my head. These convos are from 2016, so maybe something different. Also, putting sleeps in the wrong places in the harness can just make everything fail.

JoeGruffins: Pretty sure I found and fixed the problem in #1130 Calling `miner.start()` `miner.stop()` in quick succession seems to mess with the ability of txs to make it into blocks.

chappjc (approved):
Looks good. Didn't like having `gweiFactorBig` as a package-level var?

JoeGruffins:
> Didn't like having gweiFactorBig as a package-level var?

Because it's mutatable, didn't think it was a great idea to export. Maybe a bad idea to have it as a package-level var in the first place? 

The big math isn't the easiest to reason about imo. It all takes a pointer, then it returns the value? Is it returning the same value, did it mutate the original value? Need to check the documentation every time to refresh my memory. Just seems safer to do that math locally every time, with a new `*big.Int`, so possible future mistakes don't mess up everything else.

What do you think?

chappjc:
Sounds good to me.  Just curious.  I agree about the math/big API.

PR #448, opened 2020-06-05
multi: add client db upgrade infrastructure.

dnldd:
**This depends on #460**

This updates the client database to persist its current version and adds the needed infrastructure for upgrading the database as well as verifying the upgrade via tests.

Resolves #437.

PR #1079, opened 2021-05-07
bolt/upgrades: Remove redundant fetchDBVersion.

JoeGruffins:
Using an old database on master, I was getting:

```
./dexc
2021-05-07 05:11:03.456 [INF] DEXC: dexc version 0.2.0-pre+dev (Go version go1.16)
2021-05-07 05:11:03.456 [INF] DEXC: Logging with UTC time stamps. Current local time is 14:11:03 JST
2021-05-07 05:11:03.459 [INF] CORE[DB]: Upgrading database from version 0 to 3
2021-05-07 05:11:03.459 [DBG] CORE[DB]: Upgrading to version 1...
error creating client core: database initialization error: error upgrading DB: error fetching database version: database version not found
```
Although `getVersionTx` returns version 0 for no bucket, `fetchDBVersion` would error. I can't see the purpose is having both of these, so removing one.

This probably means that our gzipped test database has a version with value 0.

### DIFF
--- a/client/db/bolt/upgrades.go
+++ b/client/db/bolt/upgrades.go
@@ -101,8 +101,7 @@ func getVersionTx(tx *bbolt.Tx) (uint32, error) {
 	}
 	versionB := bucket.Get(versionKey)
 	if versionB == nil {
-		// A nil version indicates a version 0 database.
-		return 0, nil
+		return 0, fmt.Errorf("database version not found")
 	}
 	return intCoder.Uint32(versionB), nil
 }
--- a/client/db/bolt/upgrades.go
+++ b/client/db/bolt/upgrades.go
@@ -33,20 +33,6 @@ var upgrades = [...]upgradefunc{
 // upgrades prevent reverting to older software).
 const DBVersion = uint32(len(upgrades))
 
-func fetchDBVersion(tx *bbolt.Tx) (uint32, error) {
-	bucket := tx.Bucket(appBucket)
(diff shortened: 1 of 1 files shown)


chappjc:
Thinking about it.  Will need to review the history of this, but I wonder if this error is purposeful when there's no version stored because an update can't be guaranteed to work in this case.

JoeGruffins:
Better to fail? Either way, I think having both is redundant. When to fail and when to call it zero?

chappjc (approved):
I do think it should always be an error if there's no version stored because that's super old (pre-stable) and as per my previous comment we can't guarantee a correct upgrade.

Can you please update the comment in the `var upgrades` definition for `v1Upgrade` since that upgrade no longer adds a version key, only updating MatchProofs.

client/db/bolt/upgrades.go
```diff
@@ -115,8 +101,7 @@ func getVersionTx(tx *bbolt.Tx) (uint32, error) {
 	}
 	versionB := bucket.Get(versionKey)
 	if versionB == nil {
-		// A nil version indicates a version 0 database.
-		return 0, nil
+		return 0, fmt.Errorf("database version not found")
```
buck54321, line 104: We set the DB to version 1 with the merging of #448 on July 6th, 2020, long before we tagged release-v0.1.0 on Oct 20, 2020, so this is fine.

buck54321 (approved):

PR #1080, opened 2021-05-07
spec: Add eth server details.

JoeGruffins:
Adds some more details to per asset variables, removes some bad white spaces, and adds an eth example to the example markets.json

### DIFF
--- a/spec/admin.mediawiki
+++ b/spec/admin.mediawiki
@@ -32,19 +32,13 @@ possible.
 {|
 ! variable !! units !! description
 |-
-| lotSize  || int ||  the minimum order quantity and the order quantity increment when an asset is the base asset
+| lot&nbsp;size  || atoms ||  the minimum order quantity and the order quantity increment when an asset is the base asset
 |-
-| rateStep || int || the minimum price rate and the price rate increment when an asset is the quote asset. [[comm.mediawiki/#Rate_Encoding|message-rate encoding]]
+| rate&nbsp;step || atoms || the minimum price rate and the price rate increment when an asset is the quote asset. [[orders.mediawiki/#Rate_Encoding|message-rate encoding]]
 |-
-| swapConf || int || the Minimum confirmations before acting on a swap transaction
+| fee&nbsp;rate  || atoms/byte || the minimum fee rate for swap transactions
 |-
-| maxFeeRate  || int || the maximum fee rate for swap transactions
-|-
-| network || string || the network the coin daemon is running one (i.e. "mainnet")
-|-
-| bip44symbol || string || the symbol of the [https://github.com/satoshilabs/slips/blob/master/slip-0044.md registered coin type]
-|-
-| configPath || string || the path to the coin daemon's config file or ipc file in the case of Ethereum
+| swap&nbsp;confirmations|| count || the Minimum confirmations before acting on a swap transaction
 |}
 
 See also [[fundamentals.mediawiki/#Exchange_Variables|Exchange Variables]].
--- a/spec/admin.mediawiki
+++ b/spec/admin.mediawiki
@@ -32,13 +32,21 @@ possible.
 {|
 ! variable !! units !! description
 |-
-| lot&nbsp;size  || atoms ||  the minimum order quantity and the order quantity increment when an asset is the base asset
+| lot&nbsp;size  || atoms or gwei ||  the minimum order quantity and the order quantity increment when an asset is the base asset
 |-
-| rate&nbsp;step || atoms || the minimum price rate and the price rate increment when an asset is the quote asset. [[orders.mediawiki/#Rate_Encoding|message-rate encoding]]
+| rate&nbsp;step || atoms or gwei || the minimum price rate and the price rate increment when an asset is the quote asset. [[comm.mediawiki/#Rate_Encoding|message-rate encoding]]
 |-
-| fee&nbsp;rate  || atoms/byte || the minimum fee rate for swap transactions
+| fee&nbsp;rate  || atoms/byte or gwei/gas || the minimum fee rate for swap transactions
 |-
 | swap&nbsp;confirmations|| count || the Minimum confirmations before acting on a swap transaction
+|-
+| max fee rate  || atoms/byte or gwei/gas || the maximum fee rate for swap transactions
+|-
+| network || string || the network the coin daemon is running one (i.e. "mainnet")
+|-
+| bip44symbol || int || the symbol of the [https://github.com/satoshilabs/slips/blob/master/slip-0044.md registered coin type]
+|-
+| config || string || the path to the coin daemon's config file or ipc file in the case of Ethereum
 |}
 
 See also [[fundamentals.mediawiki/#Exchange_Variables|Exchange Variables]].
--- a/spec/admin.mediawiki
+++ b/spec/admin.mediawiki
@@ -32,21 +32,19 @@ possible.
 {|
 ! variable !! units !! description
 |-
-| lot&nbsp;size  || atoms or gwei ||  the minimum order quantity and the order quantity increment when an asset is the base asset
+| lotSize  || int ||  the minimum order quantity and the order quantity increment when an asset is the base asset
 |-
-| rate&nbsp;step || atoms or gwei || the minimum price rate and the price rate increment when an asset is the quote asset. [[comm.mediawiki/#Rate_Encoding|message-rate encoding]]
+| rateStep || int || the minimum price rate and the price rate increment when an asset is the quote asset. [[comm.mediawiki/#Rate_Encoding|message-rate encoding]]
 |-
-| fee&nbsp;rate  || atoms/byte or gwei/gas || the minimum fee rate for swap transactions
+| swapConf || int || the Minimum confirmations before acting on a swap transaction
 |-
-| swap&nbsp;confirmations|| count || the Minimum confirmations before acting on a swap transaction
-|-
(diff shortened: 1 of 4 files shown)


JoeGruffins:
Can be seen at the current comment here: https://github.com/JoeGruffins/dcrdex/tree/02c7e769b467ff5a61054bb6491d24b045b362c0

spec/admin.mediawiki
```diff
@@ -32,13 +32,19 @@ possible.
...
 ! variable !! units !! description
 |-
-| lot&nbsp;size  || atoms ||  the minimum order quantity and the order quantity increment when an asset is the base asset
+| lot&nbsp;size  || atoms or gwei ||  the minimum order quantity and the order quantity increment when an asset is the base asset
 |-
-| rate&nbsp;step || atoms || the minimum price rate and the price rate increment when an asset is the quote asset. [[orders.mediawiki/#Rate_Encoding|message-rate encoding]]
+| rate&nbsp;step || atoms or gwei || the minimum price rate and the price rate increment when an asset is the quote asset. [[comm.mediawiki/#Rate_Encoding|message-rate encoding]]
 |-
-| fee&nbsp;rate  || atoms/byte || the minimum fee rate for swap transactions
+| max fee rate  || atoms/byte or gwei/gas || the minimum fee rate for swap transactions
```
JoeGruffins, line 39: I can't remember the reason spaces are being encoded. It looks ok, but there must be something...

spec/admin.mediawiki
```diff
@@ -32,13 +32,21 @@ possible.
...
 ! variable !! units !! description
 |-
-| lot&nbsp;size  || atoms ||  the minimum order quantity and the order quantity increment when an asset is the base asset
+| lot&nbsp;size  || atoms or gwei ||  the minimum order quantity and the order quantity increment when an asset is the base asset
 |-
-| rate&nbsp;step || atoms || the minimum price rate and the price rate increment when an asset is the quote asset. [[orders.mediawiki/#Rate_Encoding|message-rate encoding]]
+| rate&nbsp;step || atoms or gwei || the minimum price rate and the price rate increment when an asset is the quote asset. [[comm.mediawiki/#Rate_Encoding|message-rate encoding]]
 |-
-| fee&nbsp;rate  || atoms/byte || the minimum fee rate for swap transactions
+| fee&nbsp;rate  || atoms/byte or gwei/gas || the minimum fee rate for swap transactions
```
JoeGruffins, line 39: Am I interpreting this section wrong? Should these variables mirror this?
```
// Asset is the configurable asset variables.
type Asset struct {
	ID           uint32 `json:"id"`
	Symbol       string `json:"symbol"`
	Version      uint32 `json:"version"`
	LotSize      uint64 `json:"lotSize"`
	RateStep     uint64 `json:"rateStep"`
	MaxFeeRate   uint64 `json:"maxFeeRate"`
	SwapSize     uint64 `json:"swapSize"`
	SwapSizeBase uint64 `json:"swapSizeBase"`
	SwapConf     uint32 `json:"swapConf"`
}
```

chappjc: Hmm, yeah this section of the spec does seem to describe the config items in markets.json (for Per-asset Variables) and dcrdex.conf (for Exchange Variables).

I'm noticing an error with `bip44symbol` being called an `int`.

Regarding this "fee rate", this is another out of date item because it's really *max* fee rate, not minimum.  I do recall we had a different idea about how fees would be set and enforced before we realized it needed to be dynamic based on network conditions with just a max.

Also, I'm thinking we should not set the units of fee rate explicitly since we can't keep adding units to the spec as we add assets.  It'll just have to be described as an integer, with units value per some cost such as bytes or gas.

JoeGruffins: Took out `fee rate`. Made variable names mirror whats in the `markets.json` file. Use in instead of unit names. Fix `bip44symbol` type.

spec/admin.mediawiki
```diff
@@ -32,13 +32,19 @@ possible.
...
 |-
-| swap&nbsp;confirmations|| count || the Minimum confirmations before acting on a swap transaction
+| maxFeeRate  || int || the maximum fee rate for swap transactions
+|-
+| network || string || the network the coin daemon is running one (i.e. "mainnet")
+|-
+| bip44symbol || string || the symbol of the [https://github.com/satoshilabs/slips/blob/master/slip-0044.md registered coin type]
+|-
+| configPath || string || the path to the coin daemon's config file or ipc file in the case of Ethereum
 |}
```
buck54321, line 48: This section was in the initial draft of the spec, before any code had been written. It's wasn't meant to describe a particular struct, but the motivation aligns most closely with `dex.Asset` rather than `AssetConf` because the specification describes only the API that servers implement, not implementation details. For example, we would never describe a variable called `configPath` in the spec, because it has nothing to do with how the server and client communicate.

The `AssetConf` info could be outlined in a wiki page or something, but not the spec.

chappjc: This is this the "Exchange Administration" section though, not client-server comms.

So while your comment about `AssetConf` being in wiki or somewhere else makes sense, this page actually does pertain to server config and administration details.

buck54321: > This is this the "Exchange Administration" section though, not client-server comms.

I just meant that this section isn't intended to describe implementation, just the relevant parameters for which the operator has knobs. The entire specification can be roughly defined as a definition of protocols for client-server comms, and that's the sense I was trying to use. The parameters listed in this section should eventually be communicated to the client, otherwise they don't really need to be listed.

chappjc: Gotcha, `configPath` is definitely not spec material, even in this section.

But spec/admin.mediawiki really does deviate from the proper scope of the spec as we understand it.  We have the asset params that are communicated defined in https://github.com/decred/dcrdex/blob/master/spec/fundamentals.mediawiki/#configuration-data-request.  So this Exchange Administration section really is somewhat tangential, and arguably out of place in the spec as a whole.

I just wanted to point out that the dex.Asset stuff is formalized in fundamentals.mediawiki, while this odd-ball section is actually aligned with AssetConf (and it does get uncomfortably close to the implementation).  I don't care too much what happens in this section as long as there's nothing grossly inaccurate.

JoeGruffins: Uhh. Should I revert changes here and add it all to the wiki?

chappjc: I guess that's a buck question, but I'm fine keeping this admin stuff here without the configpath.

spec/admin.mediawiki
```diff
@@ -32,13 +32,19 @@ possible.
...
+| swapConf || int || the Minimum confirmations before acting on a swap transaction
 |-
-| swap&nbsp;confirmations|| count || the Minimum confirmations before acting on a swap transaction
+| maxFeeRate  || int || the maximum fee rate for swap transactions
+|-
+| network || string || the network the coin daemon is running one (i.e. "mainnet")
+|-
+| bip44symbol || string || the symbol of the [https://github.com/satoshilabs/slips/blob/master/slip-0044.md registered coin type]
+|-
+| configPath || string || the path to the coin daemon's config file or ipc file in the case of Ethereum
```
buck54321, line 47: I'd say these three are not necessary. The network is not a **per-asset variable**. The symbol is not really a variable either.

buck54321 (changes requested):

spec/admin.mediawiki
chappjc: Then lets get rid of these tables entirely.  This page isn't describing the asset and market variables.  Those are covered elsewhere in the spec.

JoeGruffins:
Reverted the spec changes. New wiki page can be seen on my [master](https://github.com/JoeGruffins/dcrdex/wiki)
here's the wiki diff at present
<details closed>
<summary>diff</summary>

```
diff --git a/Server-Admin.md b/Server-Admin.md
new file mode 100644
index 0000000..dd04398
--- /dev/null
+++ b/Server-Admin.md
@@ -0,0 +1,32 @@
+- [exchange settings](#exchange-settings)
+  - [markets json](#markets-json)
+
+## Exchange Settings
+
+### Markets JSON
+
+```
+{
+    "markets" (array): Array of market objects.
+    [
+        {
+            "base" (string): The coin ticker shorthand followed by network of the base asset. i.e. DCR_testnet
+            "quote" (string): The coin ticker shorthand followed by network of the quote asset. i.e. BTC_testnet
+            "epochDuration" (int): The length of one epoch in milliseconds
+            "marketBuyBuffer" (float): A coefficient that when multiplied by the market's lot size specifies the minimum required amount for a market buy order
+        },...
+    ],
+    "assets" (object): Map of coin ticker shorthand followed by network of the base asset to an asset object.
+    {
+        "[TICKER_network]": {
+            "bip44symbol" (string): The coin ticker. i.e. dcr
+            "network" (string): The network the coin daemon is running on. i.e. testnet
+            "lotSize" (int): The amount of basic units of a coin in one lot
+            "rateStep" (int): The price rate increment in basic units of this coin
+            "maxFeeRate" (int): The maximum fee rate for swap transactions
+            "swapConf" (int): The minimum confirmations before acting on a swap transaction
+            "configPath" (string): The path to the coin daemon's config file or ipc file in the case of Ethereum
+        },...
+    }
+}
+```
```
</details>

chappjc:
Sounds good to me @JoeGruffins.  Thanks.  I'll have to remember to update the wiki too when the lot size and rate step move to the market from asset.
Instead of futzing with the wiki git repo, which doesn't have pull requests, I'll just apply that diff really quick.

chappjc (approved):

PR #1082, opened 2021-05-10
server/asset/eth: Implement Synced and FeeRate.

JoeGruffins:
Because SyncProgress as returned by geth will show the same result
before syncing has started and after it has finished, use the block
count to decide when syncing has started and use SyncProgress after
that.

FeeRate uses suggestGasPrice to fetch the suggested fee in gwei per unit
gas.


vctt94 (commented):
Looks like server is starting and accepting orders with this PR:

```
2021-05-11 01:29:26.031 [INF] COMM: RPC server listening on 127.0.0.1:7232
2021-05-11 01:29:30.001 [DBG] MKT: Delaying start of dcr_eth market because chains aren't synced
2021-05-11 01:29:30.002 [INF] MKT: Market dcr_btc now accepting orders, epoch 162069657:10000
2021-05-11 01:29:32.863 [DBG] ASSET[eth]: New block cf01cc4354f9bd89a6185fb16f5ce34f11d360bd24e34cbd771f84b8d6b0da92 (4770049)
2021-05-11 01:29:36.002 [INF] MKT: Market dcr_eth now accepting orders, epoch 270116096:6000

File: server/asset/eth/eth.go [hunk 1]
```diff
@@ -189,7 +209,34 @@ func (eth *Backend) ValidateSecret(secret, contract []byte) bool {
...
+	// node.SyncProgress for the lifetime of the eth backend.
+	syncingStarted := atomic.LoadUint32(&eth.syncingStarted)
+	if syncingStarted != 1 {
+		ibn := atomic.LoadUint64(&eth.initBlockNum)
+		bn, err := eth.node.blockNumber(eth.ctx)
+		if err != nil {
+			return false, err
+		}
+		if ibn == 0 {
+			atomic.StoreUint64(&eth.initBlockNum, bn)
```
buck54321, line 224: Should we set the `initBlockNum` in `(*Backend).Connect`?

server/asset/eth/rpcclient.go
```diff
@@ -71,3 +75,31 @@ func (c *rpcclient) block(ctx context.Context, hash common.Hash) (*types.Block,
 	}
 	return block, nil
 }
+
+// suggestGasPrice retrieves the currently suggested gas price to allow a timely
+// execution of a transaction.
+func (c *rpcclient) suggestGasPrice(ctx context.Context) (sgp *big.Int, err error) {
+	// NOTE: geth will panic if this is called too soon after creating the
+	// node.
```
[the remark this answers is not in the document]
JoeGruffins: I apologize, this was my fault. I need to set the ctx for the rpcclient to use before letting Connect return.

server/asset/eth/rpcclient.go
```diff
@@ -71,3 +75,31 @@ func (c *rpcclient) block(ctx context.Context, hash common.Hash) (*types.Block,
...
+	defer func() {
+		if r := recover(); r != nil {
+			sgp = nil
+			err = fmt.Errorf("geth panic: %v", r)
+		}
+	}()
```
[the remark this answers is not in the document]
JoeGruffins: I apologize, this was my fault. I need to set the ctx for the rpcclient to use before letting Connect return.

server/asset/eth/eth.go
```diff
@@ -59,17 +62,26 @@ func (d *Driver) DecodeCoinID(coinID []byte) (string, error) {
...
 
 // Backend is an asset backend for Ethereum. It has methods for fetching output
 // information and subscribing to block updates. It maintains a cache of block
 // data for quick lookups. Backend implements asset.Backend, so provides
 // exported methods for DEX-related blockchain info.
 type Backend struct {
+	// syncingStarted and initBlockNum are atomics and are used to determine
+	// whether the node has started syncing.
+	syncingStarted uint32
+	initBlockNum   uint64
```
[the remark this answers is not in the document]
JoeGruffins: Ok.

server/asset/eth/eth.go
```diff
@@ -189,7 +209,34 @@ func (eth *Backend) ValidateSecret(secret, contract []byte) bool {
 
 // Synced is true if the blockchain is ready for action.
 func (eth *Backend) Synced() (bool, error) {
-	return false, notImplementedErr
+	// node.SyncProgress will return nil both before syncing has begun and
+	// after it has finished. In order to discern when syncing has begun,
+	// wait for at least one change in block count, then defer to
+	// node.SyncProgress for the lifetime of the eth backend.
```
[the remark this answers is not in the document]
JoeGruffins: > What is geth doing before sync?

On testnet at least, looking for peers for a long time. And I believe there is a window between that and sync starting, checking the node or whatever.

> There must be a better indicator.

Searching...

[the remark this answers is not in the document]
chappjc: I ask because perhaps we can key off of `peers` length as well.  If it's zero and syncprogress returns nil, assume not in sync.  If it's non-zero peer count and syncprogress is nil, assume in sync. 
There may be similar rpcs that indicate early startup stages as well.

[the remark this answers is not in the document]
chappjc: Could also ask in the go-ethereum repo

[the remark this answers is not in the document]
JoeGruffins: Is there a way to ask questions besides making an issue? I asked in their discord, and someone suggested checking the block. We could use an external api for that...

But I think we can also use the txpool, or mempool. It seems there will be nothing in there before syncing is finished. So, if something is in there, we can fail fast, falling back to blocks for simnet or on the off chance eth has nothing in its mempool, which seems unlikely on main or testnets.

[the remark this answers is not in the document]
JoeGruffins: I think I will just use peers.

JoeGruffins:
Upon shutting down the server, getting a wall of:
```
2021-06-07 20:43:44.482 [ERR] ASSET[eth]: reorg block not found on mainchain at height 4927285 for a reorg from 4392720 to 4927298
2021-06-07 20:43:44.482 [ERR] ASSET[eth]: reorg block not found on mainchain at height 4927286 for a reorg from 4392720 to 4927298
2021-06-07 20:43:44.482 [ERR] ASSET[eth]: reorg block not found on mainchain at height 4927287 for a reorg from 4392720 to 4927298
2021-06-07 20:43:44.483 [ERR] ASSET[eth]: reorg block not found on mainchain at height 4927288 for a reorg from 4392720 to 4927298
2021-06-07 20:43:44.483 [ERR] ASSET[eth]: reorg block not found on mainchain at height 4927289 for a reorg from 4392720 to 4927298
2021-06-07 20:43:44.483 [ERR] ASSET[eth]: reorg block not found on mainchain at height 4927290 for a reorg from 4392720 to 4927298
2021-06-07 20:43:44.483 [ERR] ASSET[eth]: reorg block not found on mainchain at height 4927291 for a reorg from 4392720 to 4927298
2021-06-07 20:43:44.483 [ERR] ASSET[eth]: reorg block not found on mainchain at height 4927292 for a reorg from 4392720 to 4927298
2021-06-07 20:43:44.483 [ERR] ASSET[eth]: reorg block not found on mainchain at height 4927293 for a reorg from 4392720 to 4927298
2021-06-07 20:43:44.483 [ERR] ASSET[eth]: reorg block not found on mainchain at height 4927294 for a reorg from 4392720 to 4927298
2021-06-07 20:43:44.483 [ERR] ASSET[eth]: reorg block not found on mainchain at height 4927295 for a reorg from 4392720 to 4927298
2021-06-07 20:43:44.483 [ERR] ASSET[eth]: reorg block not found on mainchain at height 4927296 for a reorg from 4392720 to 4927298
2021-06-07 20:43:44.483 [ERR] ASSET[eth]: reorg block not found on mainchain at height 4927297 for a reorg from 4392720 to 4927298
```
Will look into.

server/asset/eth/eth.go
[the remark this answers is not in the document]
chappjc: > I asked in their discord, and someone suggested checking the block.

What does this mean, "checking the block"?  Which block, and check for what?

> I think we can also use the txpool

That's something to think about.  Let's see how peers works out first.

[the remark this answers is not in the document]
JoeGruffins: > What does this mean, "checking the block"? Which block, and check for what?

The only reply I got:
```
before syncing has begun = before it has started to pull any blocks at all?
if so, i guess you could try to check the block number as well or something, idk
```
idk if any of the devs are watching the discord, there may be a better place to ask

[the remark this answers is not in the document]
chappjc: Oh their response sounds like looking for initial sync rather than just startup at some unknown height.  Oh well.

server/asset/eth/eth.go
JoeGruffins: Looks like took a different direction, and no longer watching the block number.

server/asset/eth/rpcclient.go
```diff
@@ -8,28 +8,37 @@ import (
...
 	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/ethereum/go-ethereum/p2p"
 	"github.com/ethereum/go-ethereum/rpc"
 )
 
 // Check that rpcclient satisfies the ethFetcher interface.
 var _ ethFetcher = (*rpcclient)(nil)
 
 type rpcclient struct {
+	// ec wraps the client with some useful calls.
```
buck54321, line 23: Are you describing `ec`? Seems like you're describing the `ethclient.Client` type, which wraps an `*rpc.Client`.

server/asset/eth/rpcclient.go
```diff
@@ -8,28 +8,37 @@ import (
...
 	ec *ethclient.Client
+	// c is a direct client for raw calls.
+	c *rpc.Client
 }
 
+const gweiFactor = 1e9
+
+var gweiFactorBig = big.NewInt(gweiFactor)
```
buck54321, line 31: These aren't really specific to the RPC client. Can you move these to **eth.go**. 

My guess would be that the conversion logic you have in `FeeRate` will end up as a function something like

```Go
func toGwei(wei *big.Int) (uint64, error) {
	bigGP.Div(wei, gweiFactorBig)
	if !bigGP.IsUint64() {
		return 0, fmt.Errorf("suggest gas price %v gwei is too big for a uint64", bigGP)
	}
	return bigGP.Uint64(), nil
}
```

which will be used mostly in **eth.go**.

server/asset/eth/rpcclient.go
```diff
@@ -71,3 +80,33 @@ func (c *rpcclient) block(ctx context.Context, hash common.Hash) (*types.Block,
...
 }
+
+// suggestGasPrice retrieves the currently suggested gas price to allow a timely
+// execution of a transaction.
+func (c *rpcclient) suggestGasPrice(ctx context.Context) (sgp *big.Int, err error) {
+	sgp, err = c.ec.SuggestGasPrice(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return sgp, nil
```
buck54321, line 91:
```
return c.ec.SuggestGasPrice(ctx)
```

server/asset/eth/eth.go
```diff
@@ -30,6 +33,7 @@ const (
 	// The blockPollInterval is the delay between calls to bestBlockHash to
 	// check for new blocks.
 	blockPollInterval = time.Second
+	requiredNPeers    = 2
```
buck54321, line 36: Any reason to have 2 instead of 1? I'm wondering if this'll still work when we expand the harness tests.

server/asset/eth/eth.go
```diff
@@ -189,7 +208,23 @@ func (eth *Backend) ValidateSecret(secret, contract []byte) bool {
...
+	// ensure we are connected to at least requiredNPeers, assume the node
+	// has started syncing from those peers were they ahead, and then defer
+	// to syncProgress.
+	peers, err := eth.node.peers(eth.ctx)
+	if err != nil {
+		return false, err
+	}
+	if len(peers) < requiredNPeers {
+		return false, nil
+	}
```
buck54321, line 222: It's not clear to me that just having the peers is enough. In fact, I'm almost certain it's not. Looking through go-ethereum, I believe that `SyncProgress` maps to `(*PublicEthereumAPI).Syncing`. If `progress.CurrentBlock >= progress.HighestBlock`, you'll get the nil `*ethereum.SyncProgress`. `progress.HighestBlock` is set according to `Downloader.syncStatsChainHeight`, which is set in `(*Downloader).syncWithPeer`, which is ultimately invoked via a goroutine in `(*chainSyncer) startSync`. In short, we could have peers, but still have not started syncing, and will have a `nil` `*ethereum.SyncProgress`.

buck54321 (changes requested):

server/asset/eth/eth.go
chappjc: Same thought.  It's common in production to have a single peer where that peer is one you control, like a proxy node.

server/asset/eth/eth.go
chappjc: @JoeGruffins Since you aren't getting responses in their Discord, and it seems like the `eth_syncing` method is fundamentally flawed, I don't think it's out of line to create an issue in their github repo describing the issue and how workarounds are also incomplete.

server/asset/eth/eth.go
```diff
@@ -136,12 +145,14 @@ func (eth *Backend) Connect(ctx context.Context) (*sync.WaitGroup, error) {
 		eth.log.Errorf("error adding new best block to cache: %v", err)
 	}
 
+	running := make(chan struct{})
```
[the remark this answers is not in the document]
JoeGruffins: I thought about that, but `run(ctx)` seemed more correct. Will change however.

server/asset/eth/eth.go
JoeGruffins: Ok.

server/asset/eth/rpcclient.go
JoeGruffins: ec is a ethclient.Client. I'll reword this.

server/asset/eth/rpcclient.go
JoeGruffins: moving

server/asset/eth/eth.go
JoeGruffins: For better chances of reporting isSyncing correctly. Yeah, the harness would require a third node with this I suppose...

server/asset/eth/eth.go
JoeGruffins: https://github.com/ethereum/go-ethereum/issues/23099

JoeGruffins: They are in no hurry to comment it seems. Will try again to get a goodish temp solution in the meantime...

JoeGruffins:
Here is some code to see the highest recent block times.
<details closed>
<summary>something.js</summary>

```
function blockTimes(numBlocks) {
  last = Math.floor(Date.now() / 1000)
  bn = eth.blockNumber
  console.log("Starting block: ", bn)
  timesMap = {}
  for (i=0; i < numBlocks; i++) {
    ts = eth.getHeaderByNumber(bn-i).timestamp
    n = parseInt(ts)
    diff = last - n
    addSecs(timesMap, diff)
    last = n
  }
  printSorted(timesMap)
  return "done"
}

function addSecs(m, s) {
  m[s] = m[s] ? m[s] + 1 : 1
  var keys = Object.keys(m)
  if (keys.length > 20) {
    var lowest = 0
    keys.forEach(function cb(k) {
      var asInt = parseInt(k)
      if (lowest == 0 || asInt < lowest) {
        lowest = asInt
      }
    })
    delete(m[lowest])
  }
}

function printSorted(m) {
  keys = Object.keys(m).sort(function(a, b) {
    return parseInt(b) - parseInt(a)
  });
  sortedMap = {}
  keys.forEach(function cb(k) {
    var replacements = {"%SEC%":k,"%NUM%":m[k]},
    s = "%SEC% sec blocks happened %NUM% times"
    s = s.replace(/%\w+%/g, function(all) {
      return replacements[all] || all
    })
    console.log(s)
  })
}
```
</details>

And you can run it like:
```
$ geth --datadir=$HOME/.ethereum attach --preload "./something.js" --exec "blockTimes(10000)"
Starting block:  12765331
157 sec blocks happened 1 times
118 sec blocks happened 2 times
112 sec blocks happened 1 times
109 sec blocks happened 1 times
95 sec blocks happened 1 times
92 sec blocks happened 1 times
90 sec blocks happened 1 times
89 sec blocks happened 1 times
88 sec blocks happened 1 times
87 sec blocks happened 2 times
86 sec blocks happened 3 times
85 sec blocks happened 1 times
81 sec blocks happened 1 times
80 sec blocks happened 2 times
79 sec blocks happened 1 times
77 sec blocks happened 3 times
76 sec blocks happened 2 times
75 sec blocks happened 2 times
74 sec blocks happened 3 times
73 sec blocks happened 4 times
"done"
```
To get an idea of what kind of blocktimes to expect. Unable to go back much further with a light node for some reason, so syncing up a snap node again, although I was sure I had one synced on mainnet, I don't know where it went.

server/asset/eth/eth_test.go
```diff
@@ -332,40 +339,38 @@ func TestFeeRate(t *testing.T) {
...
-		peerInfo: make([]*p2p.PeerInfo, 2),
 		syncProg: new(ethereum.SyncProgress),
 	}, {
-		name:     "peers error",
-		peersErr: errors.New(""),
-		wantErr:  true,
+		name:    "ok header too old",
+		subSecs: maxBlockInterval,
+	}, {
+		name:       "best header",
```
JoeGruffins, line 357: best header error

server/asset/eth/eth.go
chappjc: Let's not let perfect get in the way of this.  Recall that we didn't even have a sync status at one point and it wasn't the end of the world.  If eth returns a non-nil sync status OR peers==0, we can say it is not synchronized.  Otherwise (nil sync status and peers>0) is a fairly decent indicator that it's alright to proceed.  Despite being possible to have no peers and no a nil syncstatus object, my understanding is that's not a long-lived state.

chappjc:
I'd much rather falsely indicate synced than hold up server for a long time waiting for another eth block.  It's super common to have 1min block times.  You don't even have to watch etherscan.io very long to observe that.  I think this is a minor issue though TBH.  Our Synced method should be reasonable, not necessarily perfect, and the server operator should be competent enough to get geth in synch before starting anyway.

JoeGruffins:
Was the peers method preferred? It looks like blocks don't often go over three minutes, so this method shouldn't hold anyone up, maybe randomly in a blue moon, or after an ice age starts, which one won't.

chappjc:
> Was the peers method preferred? It looks like blocks don't often go over three minutes, so this method shouldn't hold anyone up, maybe randomly in a blue moon, or after an ice age starts, which one won't.

Server won't start a market taking orders until the asset is reported as synced, so anything over 10 seconds is a no-go IMO

server/asset/eth/eth.go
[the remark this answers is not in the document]
chappjc: The primary purpose of `Context` provided by the consumer to `Connect` is to shutdown the `Backend` (in `run`, trigger `eth.shutdown()` and break the block polling loop).  So I agree it's right to have `run(ctx)` syntax for *that purpose*...

However, the `Context` we put in `eth.ctx` is to cancel RPCs.  Even if each method had a Context input, we'd still want a _connection-scoped_ context to cancel the RPCs (unless `eth.shutdown` is good enough for that?).

So although it's awkward to both have a `eth.rpcCtx` and do `eth.run(ctx)`, it seems appropriate unless we know `eth.shutdown` is going to have the same effect in which case we could just give the RPCs the background context.

Perhaps it would be more palatable if the `Backend` constructor created a `Context` and `CancelFunc` just for the RPCs, which is then called when the user cancels the context that was given to `Connect`.

i.e.

```diff
diff --git a/server/asset/eth/eth.go b/server/asset/eth/eth.go
index 917f9415..8604ecdc 100644
--- a/server/asset/eth/eth.go
+++ b/server/asset/eth/eth.go
@@ -81,7 +81,11 @@ type ethFetcher interface {
 // data for quick lookups. Backend implements asset.Backend, so provides
 // exported methods for DEX-related blockchain info.
 type Backend struct {
-       ctx  context.Context
+       // A connection-scoped Context is used to used to cancel active RPCs on
+       // subsystem shutdown.
+       rpcCtx     context.Context
+       cancelRPCs context.CancelFunc
+
        cfg  *config
        node ethFetcher
        // The backend provides block notification channels through the BlockChannel
@@ -102,7 +106,10 @@ var _ asset.Backend = (*Backend)(nil)
 // unconnectedETH returns a Backend without a node. The node should be set
 // before use.
 func unconnectedETH(logger dex.Logger, cfg *config) *Backend {
+       ctx, cancel := context.WithCancel(context.Background())
        return &Backend{
+               rpcCtx:     ctx,
+               cancelRPCs: cancel,
                cfg:        cfg,
                blockCache: newBlockCache(logger),
                log:        logger,
@@ -148,14 +155,12 @@ func (eth *Backend) Connect(ctx context.Context) (*sync.WaitGroup, error) {
                eth.log.Errorf("error adding new best block to cache: %v", err)
        }
 
-       running := make(chan struct{})
        var wg sync.WaitGroup
        wg.Add(1)
        go func() {
-               eth.run(ctx, running)
+               eth.run(ctx)
                wg.Done()
        }()
-       <-running
        return &wg, nil
 }
 
@@ -177,7 +182,7 @@ func (eth *Backend) InitTxSizeBase() uint32 {
 
 // FeeRate returns the current optimal fee rate in gwei / gas.
 func (eth *Backend) FeeRate() (uint64, error) {
-       bigGP, err := eth.node.suggestGasPrice(eth.ctx)
+       bigGP, err := eth.node.suggestGasPrice(eth.rpcCtx)
        if err != nil {
                return 0, err
        }
@@ -214,14 +219,14 @@ func (eth *Backend) Synced() (bool, error) {
        // node.SyncProgress will return nil both before syncing has begun and
        // after it has finished. In order to discern when syncing has begun,
        // check that the best header came in under maxBlockInterval.
-       sp, err := eth.node.syncProgress(eth.ctx)
+       sp, err := eth.node.syncProgress(eth.rpcCtx)
        if err != nil {
                return false, err
        }
        if sp != nil {
                return false, nil
        }
-       bh, err := eth.node.bestHeader(eth.ctx)
+       bh, err := eth.node.bestHeader(eth.rpcCtx)
        if err != nil {
                return false, err
        }
@@ -265,17 +270,13 @@ func (eth *Backend) VerifyUnspentCoin(ctx context.Context, coinID []byte) error
 // run processes the queue and monitors the application context. The supplied
 // running channel will be closed upon setting the context which is used by the
 // rpcclient.
-func (eth *Backend) run(ctx context.Context, running chan struct{}) {
-       eth.ctx = ctx
-
-       // Indicate the backend is ready to accept commands.
-       close(running)
-
+func (eth *Backend) run(ctx context.Context) {
        var wg sync.WaitGroup
        wg.Add(1)
        // Shut down the RPC client on ctx.Done().
        go func() {
                <-ctx.Done()
+               eth.cancelRPCs()
                eth.shutdown()
                wg.Done()
        }()
```

Although I acknowledge this has no practical difference from using the same Context for both run and the RPCs.

And the whole thing is unnecessary if `eth.shutdown` interrupts active RPCs.  I just don't know.

server/asset/eth/eth.go
```diff
@@ -30,6 +33,9 @@ const (
 	// The blockPollInterval is the delay between calls to bestBlockHash to
 	// check for new blocks.
 	blockPollInterval = time.Second
+	// maxBlockInterval is the number of seconds since the last header came
+	// in over which we consider the chain to be out of sync.
+	maxBlockInterval = 180
```
chappjc, line 38: :+1:

chappjc (changes requested):
I'm happy with this.  Sorry it took so long. I just had a comment about the context business, but no need to make any changes there if you're good with it as-is.  One minor change to error wrapping, and buck's requests, and we're good.

server/swap/swap.go
```diff
@@ -355,7 +355,7 @@ func (s *Swapper) ChainsSynced(base, quote uint32) (bool, error) {
 	}
 	baseSynced, err := b.Backend.Synced()
 	if err != nil {
-		return false, fmt.Errorf("Error checking sync status for %d", base)
+		return false, fmt.Errorf("Error checking sync status for %d: %v", base, err)
```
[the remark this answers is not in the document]
JoeGruffins: ok

server/asset/eth/rpcclient.go
JoeGruffins: simplified a few places

server/asset/eth/eth.go
JoeGruffins: no longer looking at peers

server/asset/eth/eth.go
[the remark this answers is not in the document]
JoeGruffins: included these changes

chappjc (approved):

PR #1083, opened 2021-05-10
server/auth: scoring for offline users must load history from DB

chappjc:
This resolves an elusive and long-standing issue where users would seemingly not have their score offset by match successes in their history. The cause is that offline users do not have entries in the `AuthManager`'s `matchOutcomes` and `preimageOutcomes` maps, but `registerMatchOutcome` and `registerPreimageOutcome`, which are often called when a user is offline, handle the lack of map entries by creating fresh histories that just included the outcome being registered at the time.

This PR resolves the issue by loading the user's match and preimage history from DB when an outcome is registered while they are offline and these outcome tracking map entries are non-existent.  This involves some minor refactoring to the `loadUserScore` method to separate the DB loading component (`loadUserOutcomes`) from score computation (`integrateOutcomes`) and insertion of outcome map entries.  I considered just deferring account scoring until the next login, but it is important for unbooking orders of suspended users.

This also modifies `handleConnect` so that an account will automatically be reinstated if computed user score indicates their account should not be closed but it is.  This is not just for bug recovery, but to allow operator changing the ban score threshold without having to manually apply the change to suspended accounts.

This also modifies the `add` methods of both latest outcome tracking structures so that it is not possible to add duplicate matches or orders.  This is done to ensure concurrent login/connect and match or swap outcome registration cannot create duplicate entries.  With small outcome slices (~100 elements) this dumb search is inexpensive.

### DIFF
--- a/server/auth/auth.go
+++ b/server/auth/auth.go
@@ -1191,8 +1191,8 @@ func (auth *AuthManager) handleConnect(conn comms.Link, msg *msgjson.Message) *m
 		auth.storage.CloseAccount(user, account.FailureToAct)
 		log.Debugf("Suspended account %v (score = %d) connected.", acctInfo.ID, score)
 	} else if score < int32(auth.banScore) && !open {
-		// banScore is a configurable threshold that may have change. This also
-		// assists recover in the event of an online accounting bug.
+		// banScore is a configurable threshold that may have changed. This also
+		// assists account recovery in the event of an online accounting bug.
 		if err = auth.Unban(user); err == nil {
 			log.Warnf("Restoring suspended account %v (score = %d).", acctInfo.ID, score)
 			client.suspended = false


server/auth/auth.go
```diff
@@ -657,44 +657,73 @@ func (auth *AuthManager) UserSettlingLimit(user account.AccountID, mkt *dex.Mark
...
+func (auth *AuthManager) registerMatchOutcome(user account.AccountID, misstep NoActionStep, mmid db.MarketMatchID, value uint64, refTime time.Time) (score int32) {
 	violation := misstep.Violation()
 
 	auth.violationMtx.Lock()
 	defer auth.violationMtx.Unlock()
-	outcomes, found := auth.matchOutcomes[user]
-	if !found {
-		outcomes = newLatestMatchOutcomes(scoringMatchLimit)
-		auth.matchOutcomes[user] = outcomes
-	}
```
chappjc, line 683: This is the gist of the bug.  I suspect it was written with the thinking "no outcomes, so this must be a new user" but the absence of an entry really means offline user.

server/auth/auth.go
```diff
@@ -657,44 +657,73 @@ func (auth *AuthManager) UserSettlingLimit(user account.AccountID, mkt *dex.Mark
...
+		log.Errorf("Failed to load swap and preimage outcomes for user %v: %v", user, err)
+		return 0
+	}
+
+	// Make outcome entries for the user to optimize subsequent outcomes calls
+	// while they are disconnected? This could lead to adding duplicate outcomes
+	// with a concurrent connect/login or subsequent outcomes while offline.
+	//
+	// auth.matchOutcomes[user] = matchOutcomes
+	// auth.preimgOutcomes[user] = piOutcomes
```
chappjc, line 720: This would be safe with the new duplicate checks in the outcome structs' `add` methods, but it's a little messy and these map entries are assumed not to be present for a disconnected user.

Even for the largest matches table so far, the `loadUserOutcomes` query is ~200ms on my (admittedly fast) machine. Even the occasional string of a dozen back-to-back outcomes for offline users (match failures or preimage misses) should be tolerable. We can experiment with this optimization or even separate outcome maps if it becomes slow to hit the DB for each outcome for an offline user.

server/auth/auth.go
```diff
@@ -1131,6 +1190,13 @@ func (auth *AuthManager) handleConnect(conn comms.Link, msg *msgjson.Message) *m
 		client.suspended = true
 		auth.storage.CloseAccount(user, account.FailureToAct)
 		log.Debugf("Suspended account %v (score = %d) connected.", acctInfo.ID, score)
+	} else if score < int32(auth.banScore) && !open {
+		if err = auth.Unban(user); err == nil {
+			log.Warnf("Restoring suspended account %v (score = %d).", acctInfo.ID, score)
+		} else {
+			log.Errorf("Failed to restore suspended account %v (score = %d): %v.",
+				acctInfo.ID, score, err)
+		}
```
chappjc, line 1199: `banScore` is an operator setting, but even so this branch would have automated account restoration on reconnect in the event that this bug manifested.

server/market/market.go
```diff
@@ -1719,6 +1717,8 @@ func (m *Market) prepEpoch(orders []order.Order, epochEnd time.Time) (cSum []byt
 			log.Errorf("Failed to revoke order %v with a new cancel order: %v",
 				ord.UID(), err)
 		}
+		// Register the preimage miss violation, adjusting the user's score.
+		m.auth.MissedPreimage(ord.User(), epochEnd, ord.ID())
```
chappjc, line 1721: Moved below the DB storage of the preimage miss.

server/swap/swap.go
```diff
@@ -1497,6 +1493,11 @@ func (s *Swapper) processRedeem(msg *msgjson.Message, params *msgjson.Redeem, st
 		// Neither party's fault. Continue.
 	}
 
+	// Credit the user for completing the swap, adjusting the user's score.
+	if actor.user != counterParty.user || newStatus == order.MatchComplete { // if user is both sides, only credit on MatchComplete (taker done too)
+		s.authMgr.SwapSuccess(actor.user, db.MatchID(match.Match), match.Quantity, redeemTime) // maybe call this in swapDone callback
+	}
```
chappjc, line 1499: Fairly inconsequential for a success, but moved below the DB store for consistency.

chappjc:
Also verified fix on simnet.  And deployed on dex-test (testnet).  Will merge shortly to include in release-v0.2

chappjc:
Tests updated for the auto-unban change, with minor fix applied to avoid need to reconnect in this case.

PR #1086, opened 2021-05-12
multi: Fix maxOrders panic.

### LINKED ISSUE #1085: rate below the rate step causes divide-by-zero panic
The recoverer middleware catches it.

image

image

JoeGruffins:
Ignore prices below the rate step in markets.js.

    Error on zero lots in MaxOrder.

    Use an estimation for number of starting lots that includes a basic fee.
    This lets estimations for very large orders start closer to the final
    fee.

closes #1085

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -767,6 +767,10 @@ func (btc *ExchangeWallet) MaxOrder(lotSize, feeSuggestion uint64, nfo *dex.Asse
 // []*compositeUTXO to be used for further order estimation without additional
 // calls to listunspent.
 func (btc *ExchangeWallet) maxOrder(lotSize, feeSuggestion uint64, nfo *dex.Asset) (utxos []*compositeUTXO, est *asset.SwapEstimate, err error) {
+	if lotSize == 0 {
+		return nil, nil, errors.New("cannot divide by lotSize zero")
+	}
+
 	btc.fundingMtx.RLock()
 	utxos, _, avail, err := btc.spendableUTXOs(0)
 	btc.fundingMtx.RUnlock()
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -766,14 +766,18 @@ func (btc *ExchangeWallet) MaxOrder(lotSize, feeSuggestion uint64, nfo *dex.Asse
 // []*compositeUTXO to be used for further order estimation without additional
 // calls to listunspent.
 func (btc *ExchangeWallet) maxOrder(lotSize, feeSuggestion uint64, nfo *dex.Asset) (utxos []*compositeUTXO, est *asset.SwapEstimate, err error) {
+	if lotSize == 0 {
+		return nil, nil, errors.New("cannot divide by lotSize zero")
+	}
 	btc.fundingMtx.RLock()
 	utxos, _, avail, err := btc.spendableUTXOs(0)
 	btc.fundingMtx.RUnlock()
 	if err != nil {
 		return nil, nil, fmt.Errorf("error parsing unspent outputs: %w", err)
 	}
-	// Start by attempting max lots with no fees.
-	lots := avail / lotSize
+	// Start by attempting max lots with a basic fee.
+	basicFee := nfo.SwapSize * nfo.MaxFeeRate
+	lots := avail / (lotSize + basicFee)
 	for lots > 0 {
 		est, _, err := btc.estimateSwap(lots, lotSize, feeSuggestion, utxos, nfo, btc.useSplitTx)
 		// The only failure mode of estimateSwap -> btc.fund is when there is
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -757,6 +757,7 @@ func (btc *ExchangeWallet) feeRateWithFallback(confTarget, feeSuggestion uint64)
 // estimate based on current network conditions, and will be <= the fees
 // associated with nfo.MaxFeeRate. For quote assets, the caller will have to
 // calculate lotSize based on a rate conversion from the base asset's lot size.
+// lotSize must not be zero and will cause a panic if so.
 func (btc *ExchangeWallet) MaxOrder(lotSize, feeSuggestion uint64, nfo *dex.Asset) (*asset.SwapEstimate, error) {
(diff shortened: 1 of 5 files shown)


JoeGruffins:
Extremely large orders would start with an initial lots estimate tens of millions of lots off of what we end up with. You can test this out with the simnet harness. I believe this code puts the initial lots closer to what we end up with.

buck54321:
Seems righteous to prevent division by zero in the asset packages, but should we reject invalid parameters in **core** before it gets that far too?

JoeGruffins:
Do mean specifically in core's `MaxBuy` and `MaxSell`?

buck54321:
> Do you mean specifically in core's `MaxBuy` and `MaxSell`?

That sounds about right.

client/asset/btc/btc.go
```diff
@@ -772,8 +773,9 @@ func (btc *ExchangeWallet) maxOrder(lotSize, feeSuggestion uint64, nfo *dex.Asse
 	if err != nil {
 		return nil, nil, fmt.Errorf("error parsing unspent outputs: %w", err)
 	}
-	// Start by attempting max lots with no fees.
-	lots := avail / lotSize
+	// Start by attempting max lots with a basic fee.
+	basicFee := nfo.SwapSize * nfo.MaxFeeRate
+	lots := avail / (lotSize + basicFee)
```
buck54321, line 778: What is this `basicFee` adjustment for? This seems wrong. `estimateSwap` will use the feeSuggestion internally.

JoeGruffins: I think it is returning before using the feeSuggestion at `tryFund`. 

The problem I am seeing is that the below estimateSwap will run an extreme number of times for large numbers of swaps if not including a fee. Please try on master with a small lot size and large amount of asset (dcr) and I believe your client will hang. Does it not?

Enter values similar to these on simnet.

image

chappjc: Yes, confirmed.  It's a problem.  Even for more reasonable values, the loop (`for lots > 0 {`) iterates excessively and we don't always realize it.

When the starting `lots` is huge, as Joe described, it has to iterate a lot more to hit on the appropriate fees.  Note than the btc ExchangeWallet.maxOrder method, `avail` and `lotSize` are sats, so when you type 0.0001 as the rate in a buy order in a DCR-BTC market with 10 DCR lot size, it starts the method with `lotSize` of 100000.  With the default simnet BTC wallet, 44000000000 sats available, and that's 440000 starting lots.  It iterates down to 381611 lots, one lot less at a time.

It is correct that `estimateSwap` returns because fund/tryfund has an error with the provided number of lots, so the loop continues with one less lot, but it's often needing to iterate thousands of times and it effectively hangs.

I think we need to start with some fees as Joe is suggesting or rethink the maxOrder algo.  `basicFee := nfo.SwapSize * nfo.MaxFeeRate` does seem correct here.

chappjc: With the proposed change:

```
avail: 44000000000  lotSize: 100000
start btc lots: 381613
final btc lots:  381611
```

So it only has to iterate twice to get the highest possible lots.

client/core/core.go
```diff
@@ -2717,6 +2717,11 @@ func (c *Core) MaxBuy(host string, base, quote uint32, rate uint64) (*MaxOrderEs
 	}
 
 	quoteLotEst := calc.BaseToQuote(rate, baseAsset.LotSize)
+
+	if quoteLotEst == 0 {
+		return nil, errors.New("cannot divide by lot size zero")
+	}
+
```
chappjc, line 2724: Can go before `c.dex` and `c.feeSuggestion`, right after getting `baseAsset` from `marketWallets`

client/core/core.go
```diff
@@ -2767,6 +2772,10 @@ func (c *Core) MaxSell(host string, base, quote uint32) (*MaxOrderEstimate, erro
 		return nil, fmt.Errorf("failed to get redeem fee suggestion for %s at %s", unbip(quote), host)
 	}
 
+	if baseAsset.LotSize == 0 {
+		return nil, errors.New("cannot divide by lot size zero")
+	}
```
chappjc, line 2777: Same move.

client/asset/interface.go
```diff
@@ -93,7 +93,8 @@ type Wallet interface {
 	// associated fees that the wallet can support for the specified DEX. The
 	// fees are an estimate based on current network conditions, and will be <=
 	// the fees associated with the Asset.MaxFeeRate. For quote assets, lotSize
-	// will be an estimate based on current market conditions.
+	// will be an estimate based on current market conditions. lotSize must
+	// not be zero and will cause a panic if so.
```
chappjc, line 97: "will cause a panic if so" wording is awkward in an interface definition because it seems to instruct backend developers to have the backend panic instead of return error, but I think this comment was noting the observed behavior of the current backends, and the comment was meant as a caution to consumers.  I don't see a problem with an asset backend developer checking for zero lot size and returning an error rather than panic.

chappjc (approved):
This PR looks and works well.

For the squashed commit, is it accurate to summarize like the following (but I can do the actual squash)?

- client/asset: maxOrder fix for starting number of lots
- client/core: reject lot sizes of zero 
- js: Ignore prices below the rate step entered on Markets page

client/core/core.go
JoeGruffins: Ok.

client/core/core.go
JoeGruffins: Ok.

client/asset/interface.go
JoeGruffins: Ok.

JoeGruffins: Hopefully did that right.

chappjc: I might have miscommunicated a bit.  "don't see a problem with an asset backend developer checking for zero lot size and returning an error rather than panic". So I just meant the interface comment saying it will panic was out of place.  I have absolutely no objection to the `if lotSize == 0 {` in maxOrder.  That was good.  Just that interface comments are fundamentally different from implementation comments in their purpose.

chappjc: > Hopefully did that right.

Yah, last commit looks great

PR #1089, opened 2021-05-13
client: prepare interfaces for bitcoin spv

buck54321:
This is a refactor to prepare for the Bitcoin SPV PR, submitted separately here so that the main PR is of a sane size. 

All normal tests are passing. Manual test and loadbot are passing. Still need to check the live tests though and probably adjust some docs.


buck54321:
Removing from draft. Noticing some errors during live testing (**livetest.go**), but also seeing issues on master. Still investigating.

client/asset/btc/btc.go
```diff
@@ -1747,336 +1733,206 @@ func (btc *ExchangeWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time,
...
-		for t := range blk.Tx {
-			tx := &blk.Tx[t]
-			found, canceled := btc.findRedemptionsInTx(scanPoint, tx, contractOutpoints)
-			totalFound += found
-			totalCanceled += canceled
-			if totalFound+totalCanceled == contractsCount {
-				break rangeBlocks
-			}
+		// If the startBlock height is lower than the lowest swap block height,
+		// ignore the startBlock.
```
itswisdomagain, line 1876: Don't see a reason why we can't just use startBlock directly without checking for the lowest swap block.

- When called from `btc.FindRedemption`, there's just one request and the lowest swap block is the block in which the single contract is mined. Can just pass that block in as the startBlock. If unmined, that block is nil and blocks search should be skipped. Currently, the genesis block will be searched if the single request passed to this method is for an unmined contract.

- When called from the block monitoring loop, a start block hash is provided, which is the hash of the latest block. The only time this start block will not be used is if it is lower than the lowest swap block height, as commented here. I expect the best block height would never be lower than the block height of any swap. Or is there a way this could happen?

client/asset/btc/btc.go
```diff
@@ -1747,336 +1733,206 @@ func (btc *ExchangeWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time,
...
-		if err != nil {
-			return nil, nil, err
-		}
-		var witness [][]byte
-		for _, hexB := range tx.Vin[vin].Witness {
-			var b []byte
-			b, err = hex.DecodeString(hexB)
-			if err != nil {
-				break
+		discovered := btc.node.searchBlockForRedemptions(validReqs, *iHash)
```
itswisdomagain, line 1891: The first time this loop runs, there may be no `undiscovered` contracts, only mempool. `len(validReqs)` would be 0 and there'd be no point to fetch this block data.

client/asset/btc/btc.go
```diff
@@ -1747,336 +1733,206 @@ func (btc *ExchangeWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time,
...
-		for i := range tx.Vin {
-			input := &tx.Vin[i]
-			if input.Vout != contractOutpoint.vout || input.Txid != contractOutpoint.txHash.String() {
-				continue // input doesn't redeem this contract, check next input
+		if len(discovered) > 0 {
+			// If we found some, we may be able to skip some blocks
+			lowHash, lowHeight := lowestBlock()
+			if lowHeight >= iHeight+1 {
+				iHash, iHeight = lowHash, lowHeight
+				continue
```
itswisdomagain, line 1911: Related to a previous comment. I can't think up a scenario where any contract's block is higher than the start block height. This code block won't be hit if there's just one request to process and the redemption has been found for that request. If there are multiple requests, then the start block is most likely the best block at the time the search was started. The contracts being searched for then could not have a block height higher than the best block height.

client/asset/btc/btc.go
```diff
@@ -1747,336 +1733,206 @@ func (btc *ExchangeWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time,
...
-		pendingContracts := make([]outPoint, 0, pendingContractsCount)
-		for _, contractOutpoint := range contractOutpoints {
-			if _, pending := btc.findRedemptionQueue[contractOutpoint]; pending {
-				pendingContracts = append(pendingContracts, contractOutpoint)
+	for iHeight <= tipHeight {
+		validReqs := make(map[outPoint]*findRedemptionReq, len(undiscovered))
+		for outPt, req := range undiscovered {
+			if iHeight >= req.blockHeight {
+				validReqs[outPt] = req
 			}
```
itswisdomagain, line 1888: I want to recommend checking the findRedemptionQueue at this point to confirm that the undiscovered contract has NOT been discovered by some other search op. E.g. the following sequence of events happen:
- the initial search op for a contract is ongoing 
- a new block is mined and another redemption search is triggered and the redemption is found in the new block
- the initial search keeps trying to find the redemption in blocks between the contract block and the newly mined block

itswisdomagain (changes requested):
Initial review. Mostly touching on the btc find redemption changes.

client/asset/btc/btc.go
```diff
@@ -1747,336 +1733,206 @@ func (btc *ExchangeWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time,
...
 	if err != nil {
-		if isTxNotFoundErr(err) {
-			return nil, nil, asset.CoinNotFoundError
-		}
-		return nil, nil, fmt.Errorf("error finding transaction %s in wallet: %w", txHash, err)
+		return exitError("scriptHashAddress error: %w", err)
 	}
-	msgTx := wire.NewMsgTx(wire.TxVersion)
-	err = msgTx.Deserialize(bytes.NewBuffer(tx.Hex))
+	pkScript, err := txscript.PayToAddrScript(contractAddr)
```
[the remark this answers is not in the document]
buck54321: This is a little more complicated than that, since we don't actually have the transaction output here. We just have a `GetTransactionResult`, which won't contain any `WalletTxDetails` about the output, since it's to an unimported p2sh address, and also wouldn't contain the pubkey script. As far as I can tell, with a Bitcoin SPV wallet, there's no way to access full transaction data until it's mined, so we can only check the contract once we find the block hash via `getTransaction`. I've added that check in the commit I'm working on now.

client/asset/btc/btc.go
```diff
@@ -1747,336 +1733,206 @@ func (btc *ExchangeWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time,
...
-		resultChan:   resultChan,
+	// If this contract is still in the findRedemptionQueue, remove from the queue
+	// to prevent further redemption search attempts for this contract.
+	btc.findRedemptionMtx.Lock()
+	delete(btc.findRedemptionQueue, outPt)
+	btc.findRedemptionMtx.Unlock()
+
+	// result would be nil if ctx is canceled or the result channel is closed
+	// without data, which would happen if the redemption search is aborted when
+	// this ExchangeWallet is shut down.
```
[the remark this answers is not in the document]
buck54321: I think it is.
https://github.com/buck54321/dcrdex/blob/1a4a24dfa6d2a2a602b0d2526afb85c1c5899b3f/client/asset/btc/btc.go#L615

client/asset/btc/btc.go
```diff
@@ -1747,336 +1733,206 @@ func (btc *ExchangeWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time,
...
-		return nil, nil, fmt.Errorf("vout index %d out of range for transaction %s", vout, txHash)
+
+	req := &findRedemptionReq{
+		outPt:        outPt,
+		contract:     contract,
+		blockHash:    blockHash,
+		resultChan:   make(chan *findRedemptionResult, 1),
+		blockHeight:  blockHeight,
+		pkScript:     pkScript,
+		contractHash: contractHash,
```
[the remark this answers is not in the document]
buck54321: I don't see a problem with storing the hash.

client/asset/btc/btc.go
```diff
@@ -1422,7 +1413,7 @@ func (btc *ExchangeWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin
 	txHash := msgTx.TxHash()
 	for i, contract := range swaps.Contracts {
 		output := newOutput(&txHash, uint32(i), contract.Value)
-		signedRefundTx, err := btc.refundTx(output.ID(), contracts[i], contract.Value, refundAddrs[i])
+		signedRefundTx, err := btc.refundTx(output.ID(), contracts[i], contract.Value, refundAddrs[i], time.Now().Add(-time.Hour))
 		if err != nil {
 			return nil, nil, 0, fmt.Errorf("error creating refund tx: %w", err)
```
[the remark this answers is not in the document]
buck54321: Issue #1163 created.

client/asset/btc/btc.go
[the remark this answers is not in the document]
chappjc: I don't see a problem either, just that it's that `contractHash == pkScript[2:34]` so one provides the other. No objection from me to either approach.

buck54321:
**regnet_test.go** isn't passing. Chasing down the issue.

chappjc:
> **regnet_test.go** isn't passing. Chasing down the issue.

The ports just changed with the reorg harness PR merged.  Maye you need to restart the  harnesses to get the updated config files for `tBackend`?

chappjc:
Err, nope the change only affected DCR and it's broken with BTC, even on master.  Dang.

client/asset/btc/btc.go
```diff
@@ -1747,336 +1741,204 @@ func (btc *ExchangeWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time,
...
 
-		for i := range tx.Vin {
-			input := &tx.Vin[i]
-			if input.Vout != contractOutpoint.vout || input.Txid != contractOutpoint.txHash.String() {
-				continue // input doesn't redeem this contract, check next input
+		iHeight += 1
+		if iHash, err = btc.node.getBlockHash(int64(iHeight)); err != nil {
+			if iHeight < tipHeight {
+				btc.log.Errorf("error getting block hash for height %d: %v", iHeight, err)
 			}
```
JoeGruffins, line 1919: Could also check here that iHash was equal to this new block's previous hash so we know we aren't in the middle of a reorg and can just stop. Or is `checkForNewBlocks()` canceling this remotely if there's a reorg, and leaving it to that? This function would continue however.

client/asset/btc/btc.go
```diff
@@ -1747,336 +1741,204 @@ func (btc *ExchangeWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time,
...
-// registered result chan.
-// Also checks mempool for potential redemptions if spending info is not found
-// for any of these contracts in the specified block range.
-func (btc *ExchangeWallet) findRedemptionsInBlockRange(startBlock, endBlock *block, contractOutpoints []outPoint) {
-	contractsCount := len(contractOutpoints)
-	btc.log.Debugf("finding redemptions for %d contracts in blocks %d - %d",
-		contractsCount, startBlock.height, endBlock.height)
+// tryRedemptionRequests searches all mainchain blocks with height >= startBlock
+// for redemptions.
+func (btc *ExchangeWallet) tryRedemptionRequests(startBlock *chainhash.Hash, reqs []*findRedemptionReq) {
```
JoeGruffins, line 1829: On any of the returns in the function after error, wouldn't it be best to call `req.fail("epic fail")` in case someone is waiting and blocking for a value on the channel?

client/asset/btc/btc.go
```diff
@@ -1747,336 +1741,204 @@ func (btc *ExchangeWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time,
...
-	btc.findRedemptionMtx.Lock()
-	btc.log.Debugf("stopping redemption search for %d contracts in queue: %v", len(contractOutpoints), err)
-	for _, contractOutpoint := range contractOutpoints {
-		req, exists := btc.findRedemptionQueue[contractOutpoint]
-		if !exists {
+	if len(mempoolReqs) == 0 {
+		return
+	}
+
+	// Do we really want to do this? Mempool could be huge.
```
JoeGruffins, line 1933: Add a long timeout and loud error log if it happens?

client/asset/btc/btc.go
```diff
@@ -2304,11 +2151,41 @@ func (btc *ExchangeWallet) run(ctx context.Context) {
 	}
 }
 
+// prepareRedemptionRequestsForBlockCheck prepares a copy of the
+// findRedemptionQueue, checking for missing block data along the way.
+func (btc *ExchangeWallet) prepareRedemptionRequestsForBlockCheck() []*findRedemptionReq {
+	// Search for contract redemption in new blocks if there
+	// are contracts pending redemption.
+	btc.findRedemptionMtx.RLock()
```
JoeGruffins, line 2159: It looks like `btc.trySetRedemptionRequestBlock(req)` can insert values and so the lock should be held for writes.

client/asset/btc/btc.go
```diff
@@ -2304,11 +2151,41 @@ func (btc *ExchangeWallet) run(ctx context.Context) {
...
+	reqs := btc.prepareRedemptionRequestsForBlockCheck()
+	// Redemption search would be compromised if the starting point cannot
+	// be determined, as searching just the new tip might result in blocks
+	// being omitted from the search operation. If that happens, cancel all
+	// find redemption requests in queue.
+	notifyFatalFindRedemptionError := func(s string, a ...interface{}) {
+		for _, req := range reqs {
+			req.fail("tipChange handler - "+s, a...)
+		}
+	}
```
JoeGruffins, line 2186: I think it would be good if this also canceled the searches. It looks like it can only notify the callers.

JoeGruffins (commented):
This is a partial review.

Running the simnet tests, they are failing half the time at  `TestMakerGhostingAfterTakerRedeem (135.12s)` with a failure like:
```
[ERR] CORE:10: Error finding redemption of taker's btc contract c2fa8deedb55e70087e26b3cebca53b1f0cc962ee43c07f7e7f44a492ded51a1:0 for order ba1fa57287f696f85eaf7a2dc0cbb85d909d717d654f5b4ff3fa431365a83c8c, match 7a7934dd44a6f142d3ed60ba37fc664270c77c0e595ce00fba823df49bb37dc2: shutdown.
```

client/asset/btc/btc.go
```diff
@@ -400,20 +398,33 @@ type block struct {
 // which is added to the findRedemptionQueue with the contract outpoint as
 // key.
 type findRedemptionReq struct {
-	ctx          context.Context
-	contractHash []byte
+	outPt        outPoint
+	contract     []byte
+	blockHash    *chainhash.Hash
```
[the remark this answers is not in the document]
chappjc: It's a practice established in https://github.com/decred/dcrd/commit/20a8ccc27bef9812d7e548f25a37011fe1cb542f that also makes sense to me, but I'm not going to be too dogmatic about it.  I'd probably put it as a value, but this is rarely executed coded so it's not important to me either way.

client/asset/btc/btc.go
```diff
@@ -1747,336 +1741,204 @@ func (btc *ExchangeWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time,
...
-			return nil, nil, err
-		}
-		if tx.Vin[vin].ScriptSig == nil {
-			return nil, nil, fmt.Errorf("no sigScript")
-		}
-		sigScript, err := hex.DecodeString(tx.Vin[vin].ScriptSig.Hex)
-		if err != nil {
-			return nil, nil, err
+		if len(validReqs) == 0 {
+			continue
```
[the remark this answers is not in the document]
buck54321: Whoops. Should be fixed and tested now.

client/asset/btc/btc.go
```diff
@@ -1846,6 +1846,13 @@ func (btc *ExchangeWallet) tryRedemptionRequests(startBlock *chainhash.Hash, req
 		undiscovered[req.outPt] = req
 	}
 
+	epicFail := func(s string, a ...interface{}) {
+		errMsg := fmt.Sprintf(s, a...)
+		for _, req := range reqs {
+			req.fail(errMsg)
+		}
+	}
+
```
JoeGruffins: Oh, nice name here.

client/asset/btc/btc.go
```diff
@@ -2416,13 +2297,81 @@ func (btc *ExchangeWallet) checkForNewBlocks() {
...
+		resultChan:   req.resultChan,
+		pkScript:     req.pkScript,
+		contractHash: req.contractHash,
+	}
+	btc.findRedemptionQueue[req.outPt] = req
+}
+
+// checkRedemptionBlockDetails looks retrieves the block at hashStr and checks
+// that the provided pkScript matches the specified outpoint
+func (btc *ExchangeWallet) checkRedemptionBlockDetails(outPt outPoint, blockStr string, pkScript []byte) (*chainhash.Hash, int32, error) {
```
JoeGruffins, line 2338: ~looks~ and hashStr -> blockStr as per the variable name

JoeGruffins (approved):
Looks to be working well. Not seeing any more simnet core test failures and code seems all right to me.

client/asset/btc/btc.go
```diff
@@ -2256,44 +2135,33 @@ func (btc *ExchangeWallet) send(address string, val uint64, feeRate uint64, subt
...
 	if err != nil {
 		if isTxNotFoundErr(err) {
-			return 0, false, asset.CoinNotFoundError
+			return 0, asset.CoinNotFoundError
 		}
-		return 0, false, err
+		return 0, err
 	}
-	return uint32(tx.Confirmations), true, nil
+	return uint32(tx.Confirmations), nil
```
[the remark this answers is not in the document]
buck54321: Any error returns before we get to here, regardless of the error code, but ltc will return the same code as btc, I think.

client/asset/btc/btc.go
buck54321: It's not impossible that this request is also being run in another thread. We're now checking the condition with the helper function [here](https://github.com/buck54321/dcrdex/blob/f6ce73bc4908b41a2fd8487863212ae13dc308ef/client/asset/btc/btc.go#L1895). It's also of little consequence if another thread returns a result, since [subsequent results are simply ignored](https://github.com/buck54321/dcrdex/blob/f6ce73bc4908b41a2fd8487863212ae13dc308ef/client/asset/btc/btc.go#L419).

chappjc:
I think we'll branch `release-v0.3` this week so we can get this and @itswisdomagain's https://github.com/decred/dcrdex/pull/788 merged without much more delay.

chappjc:
@itswisdomagain Are you happy with @buck54321's revisions in response to your review?

We will begin merging this and then https://github.com/decred/dcrdex/pull/788 as soon as I branch release-v0.3, which should happen this week since we can branch any time, but ideally after the 0.3 PRs are master.

client/asset/interface.go
```diff
@@ -148,14 +155,14 @@ type Wallet interface {
...
-	FindRedemption(ctx context.Context, coinID dex.Bytes) (redemptionCoin, secret dex.Bytes, err error)
+	FindRedemption(ctx context.Context, coinID, contract dex.Bytes) (redemptionCoin, secret dex.Bytes, err error)
 	// Refund refunds a contract. This can only be used after the time lock has
 	// expired AND if the contract has not been redeemed/refunded.
 	// NOTE: The contract cannot be retrieved from the unspent coin info as the
 	// wallet does not store it, even though it was known when the init transaction
 	// was created. The client should store this information for persistence across
 	// sessions.
-	Refund(coinID, contract dex.Bytes) (dex.Bytes, error)
+	Refund(coinID, contract dex.Bytes, startTime time.Time) (dex.Bytes, error)
```
chappjc, line 165: Would you mind just reviewing the comments for the modified `Wallet` interface functions and add a little bit about the time and contract args please, especially where it's not obvious why they are required? e.g. it's not clear why `Refund` needs the time, but `FindRedemption` does not, or if a `Wallet` implementation may or may not require the values depending on if they are spv or full-node backed.

Because an `interface` defines an API but also the _intent_ of a concrete implementation, I feel it would be helpful to describe the args in terms like *may*, *should*, and *must* so that the person implementing understands when and if an arg is not relevant or how it should be considered.

client/asset/dcr/dcr.go
```diff
@@ -1719,7 +1719,7 @@ func (dcr *ExchangeWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time,
 //
 // This method blocks until the redemption is found, an error occurs or the
 // provided context is canceled.
-func (dcr *ExchangeWallet) FindRedemption(ctx context.Context, coinID dex.Bytes) (redemptionCoin, secret dex.Bytes, err error) {
+func (dcr *ExchangeWallet) FindRedemption(ctx context.Context, coinID, contract dex.Bytes) (redemptionCoin, secret dex.Bytes, err error) {
```
chappjc, line 1722: Maybe replace `contract` with `_` for consistency with the other placeholders.

client/asset/btc/rpcclient.go
```diff
@@ -39,78 +42,146 @@ const (
...
-// RawRequester is for sending context-aware RPC requests, and has methods for
-// shutting down the underlying connection.  For testing, it can be satisfied
-// by a stub. The returned error should be of type dcrjson.RPCError if non-nil.
+// RawRequester defines dcred's rpcclient RawRequest func where all RPC
+// requests sent through. For testing, it can be satisfied by a stub.
 type RawRequester interface {
+	RawRequest(string, []json.RawMessage) (json.RawMessage, error)
+}
+
+// RawRequesterWithContext defines dcred's rpcclient RawRequest func where all
```
chappjc, line 58: dcred's

client/asset/btc/btc.go
```diff
@@ -1241,27 +1241,36 @@ func (btc *ExchangeWallet) FundingCoins(ids []dex.Bytes) (asset.Coins, error) {
 		if !notFound[pt] {
 			continue
 		}
-		txOut, err := btc.node.GetTxOut(txHash, rpcOP.Vout, true)
+
+		tx, err := btc.node.getTransaction(txHash)
```
chappjc, line 1245: Just curious about the motivation for this change.  Same change is in `lockedSats`, but I don't recall why gettxout was an issue for this in spv, just not supported with btcwallet in spv?  But `swapConfirmations` is using gettxout, so I'm not clear on this.

client/asset/btc/btc.go
```diff
@@ -362,7 +360,7 @@ type ExchangeWallet struct {
 	// 64-bit atomic variables first. See
 	// https://golang.org/pkg/sync/atomic/#pkg-note-BUG
 	tipAtConnect      int64
-	node              *rpcClient
+	node              Wallet
```
chappjc, line 363: Without seeing how the spv btwallet integration works, I'm not entirely getting the motivation for this `Wallet` interface, which is a bunch of unexported methods, so it would have to be coming from this package anyway.  The interface makes code navigation just that much more difficult cause it won't just jump to an actual implementation.

I assume you're planning a different concrete type like `spvRPCClient` that embeds an `*rpcClient` and overrides certain methods?

I can live with this, but from a clarity point of view I'm just wondering if having a single concrete implementation that has a flag for spv mode to switch between method implementations would make for easier navigation... but you've gone further down this path so I defer to you.

chappjc (changes requested):
All looks very reasonable to me, but I'm not picking apart the find redemptions code because you and @itswisdomagain are iterating on that.

client/asset/btc/btc.go
buck54321: What you've described might make sense for the initial Decred SPV implementation that still uses an RPC client underneath. For Bitcoin SPV, we are working directly with the `*btcwallet.Wallet`. There is no RPC. The implementations have little in common.

chappjc: Makes sense. Thanks.

client/asset/btc/btc.go
```diff
@@ -1799,13 +1799,8 @@ func (btc *ExchangeWallet) FindRedemption(ctx context.Context, coinID dex.Bytes)
...
-	btc.findRedemptionMtx.Lock()
-	oldRedemption := btc.findRedemptionQueue[outPt]
-	btc.findRedemptionQueue[outPt] = req
-	btc.findRedemptionMtx.Unlock()
-
-	if oldRedemption != nil {
-		oldRedemption.fail("Duplicate FindRedemption request received. Aborting the old request.")
+	if err := btc.queueFindRedemptionRequest(req); err != nil {
+		return exitError("queueFindRedemptionRequest error for redemption %s: %v", outPt, err)
 	}
```
buck54321: This change is in response to https://github.com/decred/dcrdex/pull/1089#discussion_r684511117, but I'm not sure if it's better this way. The question is who should see the error, the previously waiting goroutine, or the new one.

client/asset/dcr/simnet_test.go
```diff
@@ -268,10 +270,12 @@ func runTest(t *testing.T, splitTx bool) {
...
 
 	confCoin := receipts[0].Coin()
+	confContract := receipts[0].Contract()
 	checkConfs := func(n uint32, expSpent bool) {
 		t.Helper()
-		confs, spent, err := rig.beta().Confirmations(tCtx, confCoin.ID())
-		if err != nil {
+		confs, err := rig.beta().SwapConfirmations(tCtx, confCoin.ID(), confContract, tStart)
+		spent := errors.Is(err, asset.ErrSpentSwap)
+		if err != nil || !spent {
```
chappjc, line 278: In this test I'm getting `simnet_test.go:292: error getting 0 confs: <nil>`.  Looks like the spent test is below where it compares to `expSpent`?

client/asset/dcr/dcr.go
```diff
@@ -2341,10 +2341,10 @@ func (dcr *ExchangeWallet) ValidateSecret(secret, secretHash []byte) bool {
...
 }
 
-// Confirmations gets the number of confirmations for the specified coin ID by
-// first checking for a unspent output, and if not found, searching indexed
-// wallet transactions.
-func (dcr *ExchangeWallet) Confirmations(ctx context.Context, id dex.Bytes) (confs uint32, spent bool, err error) {
+func (dcr *ExchangeWallet) coinConfirmations(ctx context.Context, id dex.Bytes) (confs uint32, spent bool, err error) {
+	// Could check with gettransaction first, figure out the tree, and look for a
+	// redeem script with listscripts, but the listunspent entry has all the
+	// necessary fields already.
```
chappjc, line 2347: The comment about listunspent seems out of place here.

client/asset/dcr/simnet_test.go
chappjc: Why not go back to a `spent bool` return from `SwapConfirmations` if it's less error prone and easier to reason about?

client/core/trade.go
```diff
@@ -950,7 +953,8 @@ func (t *trackedTrade) isRedeemable(ctx context.Context, match *matchTracker) bo
 			return ready
 		}
 		// If we're the taker, check the confirmations anyway so we can notify.
-		confs, _, err := t.wallets.fromWallet.Confirmations(ctx, match.MetaData.Proof.TakerSwap)
+		confs, err := t.wallets.fromWallet.SwapConfirmations(ctx, match.MetaData.Proof.TakerSwap,
+			match.MetaData.Proof.Script, match.MetaData.Stamp)
```
chappjc, line 957: Although the tests in trade_simnet_test.go pass, I now see the error below logged like

`[ERR] CORE:1: error getting confirmation for our own swap transaction: swap has been spent`

While the logic here or in `isSwappable` does not change the outcome, we now see this log and a `MatchNote` will have 0 for confirmations while previously there was the actual confirmations and no error.

client/asset/dcr/simnet_test.go
buck54321: I went back.

client/core/trade.go
```diff
@@ -953,11 +952,14 @@ func (t *trackedTrade) isRedeemable(ctx context.Context, match *matchTracker) bo
...
 		}
 		// If we're the taker, check the confirmations anyway so we can notify.
-		confs, err := t.wallets.fromWallet.SwapConfirmations(ctx, match.MetaData.Proof.TakerSwap,
+		confs, spent, err := t.wallets.fromWallet.SwapConfirmations(ctx, match.MetaData.Proof.TakerSwap,
 			match.MetaData.Proof.Script, match.MetaData.Stamp)
 		if err != nil {
 			t.dc.log.Errorf("error getting confirmation for our own swap transaction: %v", err)
 		}
+		if spent {
+			t.dc.log.Errorf("our (taker) swap for match %s is being reported as spent", match)
```
chappjc, line 961: Only thing with these ERR logs is that they will happen routinely in the time between when the counterparty spends your contract and when they notify server and server audits it and server notifies us.  So it doesn't reflect a problem, at least not immediately.

chappjc (approved):

PR #1090, opened 2021-05-14
ui: Bin order tables by rate.

### LINKED ISSUE #945: ui: markets view order book tables should bin by rate
It has been noticed that different orders with the same rate have separate rows in the "Buy Orders" and "Sell Orders" tables. 
image
While that's technically correct because we're listing "orders", it's also bad UX. Instead, the table rows should bin orders by rate, displaying the combined quantity. 

This is pretty standard practice in market views, and many data API "order books" actually provide pseudo-orders that represent the rate bin. I don't recommend we do that, though. We can do the binning right where we create the row elements in markets.js.

martonp:
This PR updates the order tables on the market UI. Instead of displaying each individual order's quantity and rate on a separate row, the total quantity available at each rate will be displayed. Also, as discussed in matrix, epoch orders will be binned along with all other orders, and the check mark will no longer be displayed.

Closes #945.


JoeGruffins (approved):
Looks perfect to me.

chappjc:
Appears to work as indented.

However, we should be aware that while this is a UX improvement to some, it actually hides some important information that I hope we can strive to reveal again, perhaps in a future PR.

Take the following simnet buy side entirely comprising my own orders:

image

Another party just sees the Buy Orders table and has no idea if the 20 DCR best bid is one order or two.  This may be an important factor in the user deciding if the transaction will be maximally efficient if they decide to sell.  Even if the stars align and those maker orders do get batched by the taker/seller's redeem txn, that redeem is still larger than it would have been if there were just one match instead of two.

Then consider the 30 DCR at 0.003945.  If a user's sell fills 20 DCR at that price level, what would they match against? They would need to see the unbinned orders sorted by time stamp (as before) to know what to expect (assuming no other takers got to those bids first).

martonp:
@chappjc There could be a little pop up that shows up when you hover over each row that contains this information. I may be wrong, but I don’t think people would be so concerned with this if the lot size is set properly and the transaction fees are negligible compared to the order. It definitely can’t hurt to have it there though.

client/webserver/site/src/js/markets.js
```diff
@@ -1464,23 +1482,28 @@ export default class MarketsPage extends BasePage {
...
+      if (order.rate === row.getRate()) {
+        row.insertOrder(order)
+        return
+      } else if ((order.rate < row.getRate()) === order.sell) {
+        const tr = this.orderTableRow([order], cssClass)
         tbody.insertBefore(tr, row)
         return
       }
       row = row.nextSibling
     }
```
buck54321, line 1505: You'll need to figure out a way to separate epoch queue orders and booked orders and create a row for each.

client/webserver/site/src/js/markets.js
```diff
@@ -1545,33 +1567,68 @@ export default class MarketsPage extends BasePage {
...
+    }
+    tr.removeOrder = (token) => {
+      const index = tr.orderBin.findIndex(order => order.token === token)
+      if (index < 0) return false
+      tr.orderBin.splice(index, 1)
+      if (!tr.orderBin.length) {
+        tr.remove()
+      } else {
+        tr.updateQty()
+      }
```
buck54321, line 1617: Can be two lines

client/webserver/site/src/js/markets.js
```diff
@@ -1545,33 +1567,68 @@ export default class MarketsPage extends BasePage {
...
+    }
+    tr.removeEpochOrders = (newEpoch) => {
+      tr.orderBin = tr.orderBin.filter((order) => {
+        return !(order.epoch && order.epoch !== newEpoch)
+      })
+      if (!tr.orderBin.length) {
+        tr.remove()
+      } else {
+        tr.updateQty()
+      }
```
buck54321, line 1628: Can be two lines.

buck54321 (changes requested):
Looking good for a first pass. I do want to separate epoch orders from booked orders.

As far as the UI, I think a good first approach would just be to label the quantity of orders in the bin. e.g.

image

You could do that in a separate PR too.

martonp:
@buck54321 I've added a new column that shows the number of orders, let me know what you think. Also the epoch and non-epoch orders are now in separate rows.

client/webserver/site/package.json
```diff
@@ -23,7 +23,7 @@
     "bootstrap": "^4.6.0",
     "clean-webpack-plugin": "^3.0.0",
     "css-loader": "^5.1.2",
-    "css-minimizer-webpack-plugin": "^1.2.0",
+    "css-minimizer-webpack-plugin": "^3.0.1",
```
martonp, line 26: The CI check failed due to a vulnerability found in this package. It is now passing again after updating this.

chappjc: I don't think CI failed because of a vuln in this. https://github.com/decred/dcrdex/runs/2869810692#step:4:6
Appears to be some transient npm error.

Rather than have a  +7,140 −10,057 diff, pls revers this last diff and we'll address it in another PR.
I have a branch with a batch of npm dep updates.

chappjc:
> @buck54321 I've added a new column that shows the number of orders, let me know what you think. Also the epoch and non-epoch orders are now in separate rows.

I think it uses a bit too much space.  Just a number and no bulky column header like @buck54321 showed would be fine IMO.  Here's what I see now:

image

But it gets a little glitchy at various window widths, putting "# Orders" on two rows, and often creating a horizontal scroll bar:

image

Otherwise, it works as intended.  There's a separate row for epoch orders, and when booked, it merges that with the booked orders row and updates the # Orders count.  No issues in browser JS console.  Overall nice work, but I don't think the column header adds much, possibly even the separate column is unnecessary if the number is just after the quantity in the same column with appropriate styling.  Like buck suggested, but perhaps with a tweaked styling:

image

chappjc (changes requested):
> @buck54321 I've added a new column that shows the number of orders, let me know what you think. Also the epoch and non-epoch orders are now in separate rows.

I think it uses a bit too much space.  Just a number and no bulky column header like @buck54321 showed would be fine IMO.  Here's what I see now:

image

But it gets a little glitchy at various window widths, putting "# Orders" on two rows, and often creating a horizontal scroll bar:

image

Otherwise, it works as intended.  There's a separate row for epoch orders, and when booked, it merges that with the booked orders row and updates the # Orders count.  No issues in browser JS console.  Overall nice work, but I don't think the column header adds much, possibly even the separate column is unnecessary if the number is just after the quantity in the same column with appropriate styling.  Like buck suggested, but perhaps with a tweaked styling:

image

martonp:
@chappjc I've updated it. This definitely looks better. I thought it would be confusing to just have a number there without a column header, but I added a tooltip which explains what the number means.

chappjc (approved):
I like it.  Nice work!

image

@buck54321 LGTM, some frontend changes since your last review though.

client/webserver/site/src/js/markets.js
```diff
@@ -1525,53 +1557,103 @@ export default class MarketsPage extends BasePage {
...
     bind(tr, 'click', () => {
       this.reportClick(rate)
     })
     let qtyTD
+    const updateQtyTD = () => {
+      const mainDiv = qtyTD.children[0]
+      const [qtyDiv, ordersDiv] = mainDiv.children
+      const qty = tr.orderBin.reduce((total, curr) => total + curr.qty, 0)
+      qtyDiv.innerText = qty.toFixed(8)
+      ordersDiv.innerText = tr.orderBin.length
```
buck54321, line 1582: Let's only show the `qtyDiv` when there is more than one order.

client/webserver/site/src/css/market.scss
```diff
@@ -427,6 +427,12 @@ table.balance-table button:hover {
   }
 }
 
+div.numorders {
+  color: white;
```
buck54321, line 431: No white text, please. Inherited `color` will work fine once you get the background colors adjusted.

client/webserver/site/src/css/market_dark.scss
```diff
@@ -24,6 +24,11 @@ body.dark {
     }
   }
 
+  div.numorders {
+    background-color: rgb(25, 25, 180);
```
buck54321, line 28: Let's darken this up quite a bit. Maybe like a `#141488`.

client/webserver/site/src/css/market.scss
```diff
@@ -427,6 +427,12 @@ table.balance-table button:hover {
...
+  background-color: rgb(94, 94, 130);
+  border-radius: 3px;
+}
```
buck54321, line 434: The coloring is a little bright. It should blend in and match existing styles more closely.

image

To do that, I added `line-height: 1; padding: 1px;` here, and then removed the `px-1` class and added the `align-items-center` class to the wrapper `div`. You might want to play with `font-size: 0.9em` too, but the font-size is already pretty small here so I'm not sure we want to push it.

client/webserver/site/src/css/market.scss
```diff
@@ -427,6 +427,12 @@ table.balance-table button:hover {
   }
 }
 
+div.numorders {
+  color: white;
+  background-color: rgb(94, 94, 130);
```
buck54321, line 432: Let's lighten this up quite a bit. Maybe like a `#c5c5e6`

client/webserver/site/src/js/markets.js
```diff
@@ -1525,53 +1557,103 @@ export default class MarketsPage extends BasePage {
...
+      if (!tr.orderBin.length) tr.remove()
+      else updateQtyTD()
+    }
+    tr.getRate = () => {
+      return tr.orderBin[0].rate
+    }
+    tr.getEpoch = () => {
+      return tr.orderBin[0].epoch
+    }
+    tr.compare = (order) => {
```
buck54321, line 1649: This is good stuff, and it makes the semantics of updating these rows really clean, but once we're adding seven new functions as custom properties to an `td` element, that's a big clue to me that we should just write a custom class to manage the element and make these class methods.

client/webserver/site/src/js/markets.js
```diff
@@ -1525,53 +1557,103 @@ export default class MarketsPage extends BasePage {
...
+    const { rate, sell } = orderBin[0]
+    const isEpoch = !!orderBin[0].epoch
+    tr.orderBin = orderBin
     bind(tr, 'click', () => {
       this.reportClick(rate)
     })
     let qtyTD
+    const updateQtyTD = () => {
+      const mainDiv = qtyTD.children[0]
+      const [qtyDiv, ordersDiv] = mainDiv.children
```
buck54321, line 1579: Try using the `data-tmpl`/`Doc.tmplElement` system used throughout, and assigning these elements as properties of the custom class mentioned in my other comment.

buck54321 (changes requested):
Good stuff. I've made a few more suggestions to round things out.

client/webserver/site/src/css/market_dark.scss
chappjc: Maybe we could test out one of our primary colors: https://decred.org/brand/

Keep the hue / rgb ratio and tweak the brightness?  If that clashes with the rest of the DEX scheme, no biggie.

martonp: Screen Shot 2021-07-08 at 12 28 08 PM
Screen Shot 2021-07-08 at 12 28 21 PM

The first one is a lighter version of the dark blue primary color and the second is @buck54321 's suggestion. Very similar, but I think @buck54321 's suggestion is a closer match to the dex logo, which doesn't seem to be using any of the Decred colors.

chappjc: k, no sweat.  Just wanted to try to keep us on-brand if possible, but best to stay internally consistent with other dex styles unless we rework the whole scheme.

client/webserver/site/src/js/markets.js
martonp: I've added a manager class that has a reference to the `td` element the `td` element has a reference to the manager class. It's definitely a lot cleaner.

chappjc (approved):
Re-approving latest changes.

client/webserver/site/src/js/markets.js
```diff
@@ -1986,3 +1995,102 @@ function cleanTemplates (...tmpls) {
...
+
+  getEpoch () {
+    return this.epoch
+  }
+
+  isSell () {
+    return this.sell
+  }
+
+  compare (order) {
```
buck54321, line 2087: I try to doc all methods, but this method in particular could use a docstring.

client/webserver/site/src/js/markets.js
```diff
@@ -1446,14 +1446,43 @@ export default class MarketsPage extends BasePage {
...
+    bins.push(currNonEpochBin)
+    bins.push(currEpochBin)
+    return bins.filter(bin => bin.length > 0)
+  }
+
   /* loadTables loads the order book side into its table. */
   loadTableSide (sell) {
     const bookSide = sell ? this.book.sells : this.book.buys
     const tbody = sell ? this.page.sellRows : this.page.buyRows
     const cssClass = sell ? 'sellcolor' : 'buycolor'
```
buck54321, line 1481: `cssClass` is unused outside of `OrderTableRowManager` (via `orderTableRow`). I'd recommend moving this `cssClass` stuff into the `OrderTableRowManager` constructor, and just passing the `sell` boolean as an argument instead.

buck54321 (changes requested):
If I place two identical orders in quick succession, the epoch orders are not binned.

image

Then, when the epoch expires, and the orders are booked, they are properly grouped, but the epoch check mark is still displayed.

image

client/webserver/site/src/js/markets.js
martonp: Good call.

martonp:
@buck54321 The bug is fixed now.

buck54321 (approved):

PR #1092, opened 2021-05-17
Add form when failing to register with low balance

### LINKED ISSUE #768: ui: tell user when Decred balance can't cover registration
During registration, after the user connects a Decred wallet and after `getdexinfo` (the `(*Core).GetDEXConfig` method), we should tell the user if their wallet doesn't have enough balance to cover registration fees, and display a deposit address. 

We could also periodically poll the balance and watch for it to change to auto progress the reg process.

vctt94:
Resolves #768

### DIFF
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -2444,7 +2444,7 @@ func (dcr *ExchangeWallet) sendRegFee(addr dcrutil.Address, regFee, netFeeRate u
 	}
 	coins, _, _, _, err := dcr.fund(enough)
 	if err != nil {
-		return nil, 0, fmt.Errorf("unable to pay registration fee of %s DCR with fee rate of %d atoms/byte: %w",
+		return nil, 0, fmt.Errorf("Unable to pay registration fee of %s DCR with fee rate of %d atoms/byte: %w",
 			amount(regFee), netFeeRate, err)
 	}
 	return dcr.sendCoins(addr, coins, regFee, netFeeRate, false)
--- a/client/webserver/api.go
+++ b/client/webserver/api.go
@@ -69,7 +69,7 @@ func (s *WebServer) apiRegister(w http.ResponseWriter, r *http.Request) {
 	dcrID, _ := dex.BipSymbolID("dcr")
 	wallet := s.core.WalletState(dcrID)
 	if wallet == nil {
-		s.writeAPIError(w, core.NewError("No Decred wallet", nil))
+		s.writeAPIError(w, core.NewError("No Decred wallet"))
 		return
 	}
 
@@ -110,7 +110,7 @@ func (s *WebServer) apiNewWallet(w http.ResponseWriter, r *http.Request) {
 		Config:  form.Config,
 	})
 	if err != nil {
-		s.writeAPIError(w, err)
+		s.writeAPIError(w, core.NewError("error creating %s wallet: %v", unbip(form.AssetID), err))
 		return
 	}
 
--- a/client/webserver/api.go
+++ b/client/webserver/api.go
@@ -5,6 +5,7 @@ package webserver
 
 import (
 	"errors"
+	"fmt"
 	"net/http"
 	"time"
 
@@ -69,7 +70,7 @@ func (s *WebServer) apiRegister(w http.ResponseWriter, r *http.Request) {
 	dcrID, _ := dex.BipSymbolID("dcr")
 	wallet := s.core.WalletState(dcrID)
 	if wallet == nil {
-		s.writeAPIError(w, core.NewError("No Decred wallet"))
+		s.writeAPIError(w, errors.New("No Decred wallet"))
 		return
 	}
 
@@ -101,7 +102,8 @@ func (s *WebServer) apiNewWallet(w http.ResponseWriter, r *http.Request) {
 	}
 	has := s.core.WalletState(form.AssetID) != nil
 	if has {
-		s.writeAPIError(w, core.NewError("already have a wallet for %s", unbip(form.AssetID)))
+
+		s.writeAPIError(w, fmt.Errorf("already have a wallet for %s", unbip(form.AssetID)))
 		return
 	}
 	// Wallet does not exist yet. Try to create it.
@@ -110,7 +112,7 @@ func (s *WebServer) apiNewWallet(w http.ResponseWriter, r *http.Request) {
 		Config:  form.Config,
 	})
 	if err != nil {
-		s.writeAPIError(w, core.NewError("error creating %s wallet: %v", unbip(form.AssetID), err))
+		s.writeAPIError(w, fmt.Errorf("error creating %s wallet: %w", unbip(form.AssetID), err))
 		return
 	}
 
@@ -127,12 +129,12 @@ func (s *WebServer) apiOpenWallet(w http.ResponseWriter, r *http.Request) {
 	}
 	status := s.core.WalletState(form.AssetID)
 	if status == nil {
-		s.writeAPIError(w, core.NewError("No wallet for %d -> %s", form.AssetID, unbip(form.AssetID)))
+		s.writeAPIError(w, fmt.Errorf("No wallet for %d -> %s", form.AssetID, unbip(form.AssetID)))
 		return
 	}
 	err := s.core.OpenWallet(form.AssetID, form.Pass)
 	if err != nil {
-		s.writeAPIError(w, core.NewError("error unlocking %s wallet: %v", unbip(form.AssetID), err))
+		s.writeAPIError(w, fmt.Errorf("error unlocking %s wallet: %w", unbip(form.AssetID), err))
 		return
 	}
 
@@ -148,14 +150,14 @@ func (s *WebServer) apiNewDepositAddress(w http.ResponseWriter, r *http.Request)
 		return
 	}
 	if form.AssetID == nil {
-		s.writeAPIError(w, core.NewError("missing asset ID"))
+		s.writeAPIError(w, errors.New("missing asset ID"))
 		return
 	}
 	assetID := *form.AssetID
 
 	addr, err := s.core.NewDepositAddress(assetID)
 	if err != nil {
-		s.writeAPIError(w, core.NewError("error connecting to %s wallet: %v", unbip(assetID), err))
+		s.writeAPIError(w, fmt.Errorf("error connecting to %s wallet: %w", unbip(assetID), err))
 		return
 	}
 
@@ -179,7 +181,7 @@ func (s *WebServer) apiConnectWallet(w http.ResponseWriter, r *http.Request) {
 	}
 	err := s.core.ConnectWallet(form.AssetID)
 	if err != nil {
-		s.writeAPIError(w, core.NewError("error connecting to %s wallet: %v", unbip(form.AssetID), err))
+		s.writeAPIError(w, fmt.Errorf("error connecting to %s wallet: %v", unbip(form.AssetID), err))
 		return
 	}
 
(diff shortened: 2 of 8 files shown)


client/webserver/site/src/html/register.tmpl
```diff
@@ -85,7 +85,19 @@
...
         </div>
+      </div>
+    </form>
+
+    {{- /* registration failed low balance. */ -}}
+    <form class="card mx-auto my-5 bg1 d-hide" id="failedRegForm">
+      <div class="bg2 px-2 py-1 text-center fs18">Insuficient funds</div>
+      <div class="p-4">
+        <div class="fs16">
+          Looks like you do not have enough funds for paing the fee in the
```
JoeGruffins, line 96: paing -> paying

JoeGruffins (commented):
I get `Looks like you do not have enough funds` even if the failure is because I put in the wrong app password. The popup looks good to me, but it would really be better if it didn't have the possibility of showing misleading information.

vctt94:
Thanks for the review joe, fixed it.

JoeGruffins (commented):
It seems I only get the new form if I have exactly 0 in the account. If I have, say, .01 and the dex want's 1, I do not see the form, although I do have insufficient funds.

vctt94:
Are you trying to register to `dex-test.ssgen.io:7232`? Because if that is the case, the fee is 0.1 dcr (10000000 atoms), so you have sufficient funds, but no sufficient funds for the tx fee.

I added an `insuficient funds` into the message, so the form also shows in this case.

client/asset/dcr/dcr.go
```diff
@@ -2444,7 +2444,7 @@ func (dcr *ExchangeWallet) sendRegFee(addr dcrutil.Address, regFee, netFeeRate u
 	}
 	coins, _, _, _, err := dcr.fund(enough)
 	if err != nil {
-		return nil, 0, fmt.Errorf("unable to pay registration fee of %s DCR with fee rate of %d atoms/byte: %w",
+		return nil, 0, fmt.Errorf("Insufficient funds for fee rate. Unable to pay registration fee of %s DCR with fee rate of %d atoms/byte: %w",
```
JoeGruffins, line 2447: I think must be lowercase `i` in order to match the string.

But really, would be good to have an error code or something, as matching strings is very fragile.

Also, one day fees may be payable by different coins. So, other assets must also pay attention to their error messages. Also, this may cause problems with internationalization.

client/webserver/site/src/js/register.js
```diff
@@ -189,6 +191,14 @@ export default class RegistrationPage extends BasePage {
     const res = await postJSON('/api/register', registration)
     loaded()
     if (!app.checkResponse(res)) {
+      // show different form with no passphrase input in case of no funds.
+      if (res.msg.includes('insufficient funds')) {
```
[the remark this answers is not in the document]
vctt94: I added two new methods as a POC, `newCoreErr`and `codedCoreError`, which are the same of `newError` and `codedError`, but instead of returning `error`, they are returning `core.Error`. 

This way it is possible to have access to the code error, and we can use it for treating in the frontend. 

The problem with that approach is that we would need to change the error handling in all webserver api call, but it would give more independence, for treating errors.

What do you think?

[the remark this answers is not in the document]
chappjc: Yeah, so I'm pretty sure almost none of those changes were really needed, namely the `(*Core).Register` changes and the introduction of `newCoreErr`.  Note that `core.Error` is an `error`, hence why the existing `newError` and `codedError` returned a `*core.Error` as an `error.

Downstream when inspecting the regular `error` you'd just do:

```go
var cErr core.Error
if errors.As(err, &cErr){
    // grab cErr.Code
}
```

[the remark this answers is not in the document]
chappjc: What I had in mind was a mod to `writeAPIError` to accept an `err error` instead of `format string` because then you can do the errors cast and code checking, plus you get the error message from `Error()`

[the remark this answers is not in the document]
vctt94: Okay, pushed the requested changes, sending an error instead of `format string`.

For now I added a new method `writeAPIErrorPOC`, and if approved I can call it in other api errors handling

client/webserver/site/src/js/register.js
```diff
@@ -189,6 +194,14 @@ export default class RegistrationPage extends BasePage {
     const res = await postJSON('/api/register', registration)
     loaded()
     if (!app.checkResponse(res)) {
+      // show different form with no passphrase input in case of no funds.
+      if (res.code === sendFeeErr) {
```
chappjc, line 198: If `Code` is omitted (either the backend used the old `writeAPIError`, or used the new `writeAPIErrorPOC` but with `Code=0` which `omitempty` also includes), wont this have a js error with res.code being undefined?

client/webserver/site/src/js/register.js
```diff
@@ -6,6 +6,9 @@ import { NewWalletForm, bindOpenWallet, bind as bindForm } from './forms'
 const DCR_ID = 42
 const animationLength = 300
 
+// XXX: move code errors to a specific file?
+const sendFeeErr = 8;
```
chappjc, line 10: CI is mad here

chappjc (changes requested):
Yup, this approach is what I had in mind.  Just a little clean-up and docs, and maybe consider something for js not to fail if `code` is undefined.

client/webserver/site/src/html/register.tmpl
```diff
@@ -89,6 +89,19 @@
...
+    <form class="card mx-auto my-5 bg1 d-hide" id="failedRegForm">
+      <div class="bg2 px-2 py-1 text-center fs18">Insuficient funds</div>
+      <div class="p-4">
+        <div class="fs16">
+          Looks like you do not have enough funds for paying the fee in the
+          selected account. Please fund the account and try again.
+        </div>
+        <div class="fs15 pt-3 text-center d-hide errcolor" id="regFundsErr"></div>
+        <hr class="dashed mt-4">
+      </div>
```
chappjc, line 102: This is still just a guess that it's insufficient funds, but really it's any error from `wallet.PayFee`, which returns a `feeSendErr`-coded error.

It could be many things including a bad fee address, or an error from `sendWithReturn` such as failure to sign the tx or failure to broadcast with sendrawtransaction.

Thus, I think it's warranted to soften the language in this dialog to indicate an error, but suggest the most likely cause of insufficient funds, and ofc show the whole error message as you are doing.

client/webserver/site/src/js/register.js
vctt94: Yeah, it will be undefined, which will be !== from `sendFeeErr`, so we are good. It would crash everything if res is undefined, which I believe it never is, is that right? 

Lookin into `checkResponse`, it does not check for res to be defined.

vctt94:
This PR is ready for another review.

chappjc (changes requested):

client/asset/dcr/dcr.go
```diff
@@ -2444,7 +2444,7 @@ func (dcr *ExchangeWallet) sendRegFee(addr dcrutil.Address, regFee, netFeeRate u
 	}
 	coins, _, _, _, err := dcr.fund(enough)
 	if err != nil {
-		return nil, 0, fmt.Errorf("unable to pay registration fee of %s DCR with fee rate of %d atoms/byte: %w",
+		return nil, 0, fmt.Errorf("Unable to pay registration fee of %s DCR with fee rate of %d atoms/byte: %w",
```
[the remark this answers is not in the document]
vctt94: Interisting. Why do you object it?

vctt94:
Thanks for the review chapp. Addressed your comments

client/asset/dcr/dcr.go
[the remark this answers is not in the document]
chappjc: It's a common complaint from linters, including staticcheck ([ST1005)](https://staticcheck.io/docs/checks#ST1005) and Go's own [golint](https://github.com/golang/lint/blob/738671d3881b9731cc63024d5d88cf28db875626/lint.go#L1150-L1151), with the reasoning explained in the  [Go Code Review Comments wiki page](https://github.com/golang/go/wiki/CodeReviewComments#error-strings).
The reasoning, which is generally true, is that the error strings follow context like `"Failure with X: %w", err"`. Same reason why trailing punctuation is to be avoided in error strings.

client/webserver/api.go
```diff
@@ -278,7 +280,7 @@ func (s *WebServer) apiCancel(w http.ResponseWriter, r *http.Request) {
 	}
 	err := s.core.Cancel(form.Pass, form.OrderID)
 	if err != nil {
-		s.writeAPIError(w, core.NewError("error cancelling order %s: %v", form.OrderID, err))
+		s.writeAPIError(w, errors.New(fmt.Sprintf("error cancelling order %s: %v", form.OrderID, err)))
```
chappjc, line 283: These `errors.New(fmt.Sprintf` should be replaced with `fmt.Errorf`.

Also, when there is an `err` as one of the printf args, the specifier should be `%w` so that errors.Is/As can unwrap the core.Error (or whatever type it is)

client/webserver/api.go
```diff
@@ -681,12 +683,17 @@ func (s *WebServer) apiUser(w http.ResponseWriter, r *http.Request) {
...
+	var code int
+	if errors.As(err, &cErr) {
+		code = cErr.Code()
+	}
 	resp := &standardResponse{
-		OK:  false,
-		Msg: errMsg,
+		OK:   false,
+		Msg:  err.Error(),
+		Code: &code,
```
chappjc, line 695: Um so I made an incomplete suggestion.  When the error is not detected as a core.Error above, it needs to be a nil pointer to omit the code.

```
var code *int
if errors.As(err, &cErr) {
    // set code
....
Code: code,
```

client/webserver/api.go
vctt94: Got it, my bad.

client/webserver/api.go
vctt94: should be good now

chappjc:
I've revised the PR description so it doesn't close https://github.com/decred/dcrdex/issues/768 when merged.

This PR will recognize an error from PayFee and then show the following:

image

However, because the above comes after confirming the following dialog, the issue isn't quite resolved.

image

In https://github.com/decred/dcrdex/issues/768, @buck54321 was pointing out that instead of showing them this dialog when the backend already knows the balance is insufficient, it should show a different dialog that prevents the registration and fee payment attempt entirely.

Detecting a certain fee payment error code and displaying an appropriate dialog, as done in this PR, makes complete sense, but it does not quite resolve https://github.com/decred/dcrdex/issues/768

vctt94:
okay, now it goes directly to the registration failed form, if there are not enough funds.

If available funds are equal the reg fee, than the process occurs and only fails in the end, as we do not check for the network fee.

JoeGruffins (approved):
Working well for me now.

client/webserver/api.go
```diff
@@ -681,12 +683,17 @@ func (s *WebServer) apiUser(w http.ResponseWriter, r *http.Request) {
...
+	var code *int
+	if errors.As(err, &cErr) {
+		code = cErr.Code()
+	}
 	resp := &standardResponse{
-		OK:  false,
-		Msg: errMsg,
+		OK:   false,
+		Msg:  err.Error(),
+		Code: code,
```
chappjc, line 695: Good. Code is omitted when there is no error (just OK=true).

chappjc (approved):
Working well.

UI after entering dex address:

image

(doesn't have you try to confirm payment first -- good)

dexcctl register:

> error paying registration fee: Unable to pay registration fee of 1 DCR with fee rate of 11 atoms/byte: insufficient funds. 0 DCR available to spend in "broke" account

PR #1093, opened 2021-05-19
ui: Display order errors on the form.

### LINKED ISSUE #1045: ui: order errors should display on the form
As a general rule, any error encountered when submitting a form should be displayed on the form itself, not the notification bell.

amass01:
Closes #1045.

> As a general rule, any error encountered when submitting a form should be displayed on the form itself, not the notification bell. 

We will need to skip notification in checkResponse function and just return false if there is an error, and then display `res.msg` in related form's error section for all other forms.

### DIFF
--- a/client/webserver/site/src/js/markets.js
+++ b/client/webserver/site/src/js/markets.js
@@ -1326,7 +1326,6 @@ export default class MarketsPage extends BasePage {
    */
   async submitOrder () {
     const page = this.page
-    const market = this.market
     Doc.hide(page.forms)
     const order = this.parseOrder()
     const pw = page.vPass.value
@@ -1337,14 +1336,10 @@ export default class MarketsPage extends BasePage {
     }
     if (!this.validateOrder(order)) return
     const res = await postJSON('/api/trade', req)
-    if (!app.checkResponse(res)) return
-    // If the wallets are not open locally, they must have been opened during
-    // ordering. Grab updated info.
-    const baseWallet = app.walletMap[market.base.id]
-    const quoteWallet = app.walletMap[market.quote.id]
-    if (!baseWallet.open || !quoteWallet.open) {
-      this.balanceWgt.updateAsset(market.base.id)
-      this.balanceWgt.updateAsset(market.quote.id)
+    if (!app.checkResponse(res, true)) {
+      page.orderErr.textContent = res.msg
+      Doc.show(page.orderErr)
+      return
     }
     this.refreshActiveOrders()
     this.chart.draw()
--- a/client/webserver/site/src/js/markets.js
+++ b/client/webserver/site/src/js/markets.js
@@ -51,7 +51,8 @@ export default class MarketsPage extends BasePage {
       'forms', 'openForm', 'uwAppPass',
       // Order submission is verified with the user's password.
       'verifyForm', 'vHeader', 'vSideHeader', 'vSide', 'vQty', 'vBase', 'vRate',
-      'vTotal', 'vQuote', 'vPass', 'vSideSubmit', 'vBaseSubmit', 'vSubmit', 'verifyLimit', 'verifyMarket',
+      'vTotal', 'vQuote', 'vPass', 'vSideSubmit', 'vBaseSubmit', 'vSubmit',
+      'vLoader', 'verifyLimit', 'verifyMarket',
       'vmTotal', 'vmAsset', 'vmLots', 'mktBuyScore', 'vErr',
       // Create wallet form
       'walletForm',
@@ -1326,7 +1327,6 @@ export default class MarketsPage extends BasePage {
    */
   async submitOrder () {
     const page = this.page
-    page.vSubmit.disabled = true
     Doc.hide(page.orderErr, page.vErr)
     const order = this.parseOrder()
     const pw = page.vPass.value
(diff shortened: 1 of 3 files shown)


client/webserver/site/src/js/markets.js
```diff
@@ -1313,14 +1312,10 @@ export default class MarketsPage extends BasePage {
...
     if (!this.validateOrder(order)) return
     const res = await postJSON('/api/trade', req)
-    if (!app.checkResponse(res)) return
-    // If the wallets are not open locally, they must have been opened during
-    // ordering. Grab updated info.
-    const baseWallet = app.walletMap[market.base.id]
-    const quoteWallet = app.walletMap[market.quote.id]
-    if (!baseWallet.open || !quoteWallet.open) {
-      this.balanceWgt.updateAsset(market.base.id)
-      this.balanceWgt.updateAsset(market.quote.id)
```
amassarwi, line 1323: This isn't relevant after #1007 which handles this for both wallets using notification.

JoeGruffins (approved):
Works for me.

chappjc:
This works for me too.

But we have a more fundamental UI flaw with orders in that you can submit multiple orders before getting any response from the /api/trade endpoint request.  This is most obvious with the following: (1) testnet and (2) tx split enabled.  A single /api/trade can take ~10sec for BTC, and a few for DCR.  It's very easy to fire off several before getting a response from the first.

Perhaps a solution to the above issue is to simply disable/gray-out the submit (Place order to buy/sell...) button until a response (success or fail) is received from the /api/trade POST.

amassarwi:
@chappjc Hrmm good point, I can add that.

client/webserver/site/src/js/markets.js
```diff
@@ -1302,7 +1302,6 @@ export default class MarketsPage extends BasePage {
    */
   async submitOrder () {
     const page = this.page
-    const market = this.market
     Doc.hide(page.forms)
```
buck54321, line 1305: `Doc.hide(page.forms, page.orderErr)`

buck54321 (changes requested):

chappjc:
To clarify my comment, I believe the following dialog should be modal and the target for errors:

image

Presently it disappears when you submit, and any amount of time can pass before server accepts the order and client gets the response and frontend is informed.

When it goes away immediately you are left hanging with no sign of an order anywhere until server processes the request and gets back to you and informs frontend.  Its usually a "fast" submission so we deal with it, but it's still bad UX.  And it's terrible when the delay is longer than a second as with txsplit in use on testnet or mainnet.  But any server-side slowdown can do this, or connectoin drop, etc.

Hence I think the right solution is to keep this dialog opened after clicking until i response is received.  During that time, the Buy/Sell DCR button should be disabled.

amassarwi:
> This works for me too.
> 
> But we have a more fundamental UI flaw with orders in that you can submit multiple orders before getting any response from the /api/trade endpoint request. This is most obvious with the following: (1) testnet and (2) tx split enabled. A single /api/trade can take ~10sec for BTC, and a few for DCR. It's very easy to fire off several before getting a response from the first.
> 
> Perhaps a solution to the above issue is to simply disable/gray-out the submit (Place order to buy/sell...) button until a response (success or fail) is received from the /api/trade POST. EDIT: This needs some more thought though. I wonder if there needs to be some kind of indicator of a pending order submission POST request...

As discussed on matrix, @chappjc suggests we change a bit the flow, and keep the confirmation modal visible with 
grayed out submit button, and in case of error display it on the confirmation modal where the user have the order's fixed numbers displayed.

amassarwi:
cc: @buck54321 thoughts regarding the above ?

amassarwi:
@chappjc updated 👍🏽

buck54321:
> cc: @buck54321 thoughts regarding the above ?

@chappjc's recommendation is what I would have suggested too. Sorry for the late response.

chappjc:
Good, the dialog does not close now while the /api/trade request is processing.

But when I first entered my pass and clicked "Buy DCR" the password field cleared out, but nothing else happened until the order was accepted. Although the button is actually disabled, there really needs to be more indication that the order submission is processing:

- Graying out the red/green button would be a bare minimum
- Some kind of "working" indicator like a spinning / pulsing thing.  We have a spinning thing for when the markets page is loading, and when max order is fetching, that could be repurposed next to or in place of the Buy/Sell button once it's clicked.  See `spinner` and `ico-spinner` css classes, and the `loader`, `marketLoader`, `baseSpinner`, and `quoteSpinner` element ids.

amassarwi:
@chappjc adding a spinner in place of the button.

client/webserver/site/src/js/markets.js
```diff
@@ -1313,15 +1313,16 @@ export default class MarketsPage extends BasePage {
...
+    // If errors display error on confirmation modal.
+    if (!app.checkResponse(res, true)) {
+      page.vErr.textContent = res.msg
+      Doc.show(page.vErr)
+      page.vSubmit.disabled = false
+      return
     }
+    // Hide confirmation modal only on success.
+    Doc.hide(page.forms)
+    page.vSubmit.disabled = false
```
buck54321, line 1325: Can move `page.vSubmit.disabled = false` up before `if (!app.checkResponse(res, true)) {` and only do it once.

buck54321 (changes requested):

client/webserver/site/src/js/markets.js
buck54321: Bump. Good to merge after this last thing.

chappjc: The button needs to go gray too if not a spinner.

amassarwi: on it now

amassarwi: Done

chappjc (approved):
Spinner changes working well. Examples below.

Error with insufficient balance (spinner only flashes very briefly in this test):

https://user-images.githubusercontent.com/9373513/127744030-e6c3dfa7-e533-45f7-8dfd-2c0d77c4826e.mp4

Error from the `asset.CoinNotFoundError` bug fixed in https://github.com/decred/dcrdex/pull/1128/files#r675698564, where sell DCR fails with txplit:

https://user-images.githubusercontent.com/9373513/127744321-dadb2879-7bb0-43ce-9363-9009686cfbe1.mp4

With a `time.Sleep(2 * time.Second)` in `apiTrade` to simulate latency, where you can see the spinner:


https://user-images.githubusercontent.com/9373513/127744338-462d10ee-7430-4949-86f8-e3fac85b9394.mp4

PR #1094, opened 2021-05-24
frontend: order status for unfilled non-standing orders

### LINKED ISSUE #1084: ui: settling booked order only says booked
Partially filled booked orders that are settling used to say something like "booked/settling", but it now just says booked.  Observed the following on testnet:

image

The last order 50% filled is the one matched with the first order in the list and still settling:

image

martonp:
This PR updates the status string for executed orders with 0% filled to "expired" (mentioned in #945), and for orders that are partially booked and partially currently settling to "booked/settling".

Closes #1084.

### DIFF
--- a/client/webserver/types.go
+++ b/client/webserver/types.go
@@ -310,7 +310,7 @@ func (ord *orderReader) StatusString() string {
 			return "settling"
 		}
 		if ord.Filled == 0 {
-			return "expired"
+			return "no match"
 		}
 		return "executed"
 	case order.OrderStatusCanceled:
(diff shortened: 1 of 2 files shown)


client/webserver/types.go
```diff
@@ -301,11 +301,17 @@ func (ord *orderReader) StatusString() string {
...
+		if isLive {
+			return "booked/settling"
+		}
 		return "booked"
 	case order.OrderStatusExecuted:
 		if isLive {
 			return "settling"
 		}
+		if ord.Filled == 0 {
+			return "expired"
```
chappjc, line 313: Good changes in this PR, but I think expired is a bit confusing as it hints at some time limit or deadline that passed, but it was just not booked and had no matches to settle.  Some more accurate ideas: "missed", "no match", "unfilled", and combinations with a "executed / ..." prefix

martonp: I agree that's more clear for sure. It's updated.

chappjc (approved):
Tests well.

PR #1098, opened 2021-05-28
geth/harness: Use snap mode for server nodes.

JoeGruffins:
As mentioned [here](https://gist.github.com/chappjc/4feefe593f870cd008180c00f72f85ca) use snap nodes for the server as they greatly reduce the resources needed by the server after caught up. Also incorporate `datadir.ancient` which further helps with the cost of running an eth full node by saving some data on cheaper hdd's. Afaik other data must still be stored on an ssd due to the large number of reads and writes. Requires geth 1.10+

### DIFF
--- a/dex/testing/eth/create-node.sh
+++ b/dex/testing/eth/create-node.sh
@@ -125,12 +125,13 @@ $NODE_KEY
 EOF
 
 echo "Starting simnet ${NAME} node"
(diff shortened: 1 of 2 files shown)


chappjc (approved):
LGTM.  Obv not important for harness chains, but I think it's important to be testing with a config that is intended for production, so ACK.

PR #1099, opened 2021-05-28
asset/dcr: Remove calcPastMedianTime.

JoeGruffins:
The value has been added to the verbose block header. Retrieve it from
there.


chappjc (approved):
Thanks for following up.

Did the trade_simnet_test.go tests pass?

JoeGruffins:
> Did the trade_simnet_test.go tests pass?

They do for me.

PR #1101, opened 2021-05-31
begin 0.3

chappjc:
Start the v0.3 cycle. Also update the linter, and fix an unbuffered `os.Signal` channel.

### DIFF
--- a/client/cmd/dexc/main.go
+++ b/client/cmd/dexc/main.go
@@ -90,7 +90,7 @@ func mainCore() error {
 
 	// Catch interrupt signal (e.g. ctrl+c), prompting to shutdown if the user
 	// is logged in, and there are active orders or matches.
-	killChan := make(chan os.Signal)
+	killChan := make(chan os.Signal, 1)
 	signal.Notify(killChan, os.Interrupt)
 	go func() {
 		for range killChan {
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Check out source
         uses: actions/checkout@v2
       - name: Install Linters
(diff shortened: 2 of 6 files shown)


client/cmd/dexc/main.go
```diff
@@ -90,7 +90,7 @@ func mainCore() error {
 
 	// Catch interrupt signal (e.g. ctrl+c), prompting to shutdown if the user
 	// is logged in, and there are active orders or matches.
-	killChan := make(chan os.Signal)
+	killChan := make(chan os.Signal, 1)
```
buck54321, line 93: This is in line with what the **signal** docs recommend, but it's not clear that this was problematic to begin with. The docs also clearly state that

> Package signal will not block sending to c

So the only real difference I see with an unbuffered `chan` is if the signal is sent before we are listening, e.g. the user hits ctrl+C after `signal.Notify`, but before the `go func` on the next line launches.

Anyway, this is fine.

buck54321 (approved):

client/cmd/dexc/main.go
chappjc: Right, that's the only way it's a problem, and why govet now complains about it.  There's a possibility that the signal will be lost/ignored if the other goroutine isn't (yet) receiving on the channel yet.  It's a legit concern since the go runtime isn't required to prioritize the other goroutine's execution over this one's.

The fact that package signal doesn't block is why govet does not complain about `signal.Notify(make(chan os.Signal), os.Interrupt)` (although that is basically like using `signal.Ignore`) but it does complain about this.

JoeGruffins (approved):

client/cmd/dexc/main.go
chappjc: Just for ref, this is the govet error that shows up with the current govet:

`::error file=client/cmd/dexc/main.go,line=94,col=2::sigchanyzer: misuse of unbuffered os.Signal channel as argument to signal.Notify (govet)`

Issue #960, opened 2021-02-02
make lot size a market parameter, not an asset parameter


chappjc:
It was probably a mistake to make lot size an asset parameter instead of a market (asset *pair*) parameter.
It precludes setting up a low lot value market like dcr_**ltc**.  Also I, don't think the per-asset lot size facilitates going between markets as I think we originally intended because of tx fees, but it's pretty much a non-issue anyway given how orders are funded with whatever utxos and even with the pre-split option.

This is going to be a major change with sweeping consequences, so this should be for 0.3/1.0.

chappjc:
Forgot to assign this to myself.  Almost done so I hope no one else was working on it.

PR #1102, opened 2021-06-04
multi: lot size and rate step are market params

chappjc:
Resolves https://github.com/decred/dcrdex/issues/960

This makes lot size and rate step *market* parameters instead of asset parameters.

`msgjson.Market` now includes `LotSize` and `RateStep`.

`msgjson.Asset` still has `LotSize` and `RateStep`, but they are tagged with `omitempty`, and will be zero/omitted when market configs require multiple different lot sizes or rate steps for an asset.

`dex.Asset` no longer has `LotSize` or `RateStep` fields.

The dcrdex server harness uses a configuration with two markets with DCR as the base asset and *different* lot sizes.

The dcrdex server app now errors if markets.json has lot size or rate step set for any asset, and if any market lacks a non-zero lot size or rate step.

`server/dex.AssetConf` renamed `LotSize` and `RateStep` to `LotSizeOLD` and `RateStepOLD`, and added `omitempty` to both.

`server/dex.NewDEX` generates a `'config'` response in different ways depending on if a given asset has consistent lot sizes and rate steps across all markets.  If consistent, it will set the `LotSize`/`RateStep` in the `msgjson.Asset`.  If not consistent, the values will be zero/omitted and a warning will be logged that old clients will not work with the asset.

While here in `NewDEX`, abort in a defer with a `ready bool`.


chappjc:
Still need to try an older client to ensure there's a reasonably graceful failure with zero lot size, but the PR seems to be working as expected, so opening to review.

client/webserver/site/src/js/register.js
```diff
@@ -165,8 +165,8 @@ export default class RegistrationPage extends BasePage {
     this.fee = res.xc.feeAsset.amount
 
     page.feeDisplay.textContent = Doc.formatCoinValue(this.fee / 1e8)
-    const dcrAsset = res.xc.assets['42']
-    if (dcrAsset) page.dexDCRLotSize.textContent = Doc.formatCoinValue(dcrAsset.lotSize / 1e8)
+    // const dcrAsset = res.xc.assets['42']
+    page.dexDCRLotSize.textContent = '<depends on market>' // Doc.formatCoinValue(dcrAsset.lotSize / 1e8)
```
chappjc, line 169: Open for suggestions on how to handle this. Perhaps changing the text to something like "The server has markets with DCR as the base asset with the following lot sizes: ..."  Not sure how to cleanly implement or if there's a better way to just summarize *all* of the server's markets in a table for the user to decide.

JoeGruffins (approved):
Working well.

client/webserver/site/src/js/register.js
buck54321: A small table does sound pretty good. It's not going to scale that well to a DEX with lots of DCR (or other reg asset) markets, but we can worry about that when we have more than 4 assets.

@chappjc, if you don't want to do the html, feel free to use a simpler solution of your choosing in the interim and we can open an issue.

server/dex/dex.go
```diff
@@ -305,6 +309,32 @@ func NewDEX(ctx context.Context, cfg *DexConf) (*DEX, error) {
 		assetIDs[i] = ID
 	}
 
+	// Check each market's base lot size to see if any asset has different base
+	// lots sizes, which will break older clients. In this case, set
+	// msgjson.Asset.LotSize 0 so older clients cannot use the market. Do the
+	// same for rate step.
```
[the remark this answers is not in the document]
chappjc: I'd like to do that, but that would require bumping client's expected api version to 1 as well.  However, we're not formally at v1, so we might end up with the next release at v2 or higher (never having released with v1).  I guess that's OK but I really wanted to only switch to v1 for release.

client/webserver/site/src/js/register.js
chappjc: Went with this lot size for whatever DCR base market, preferring DCR_BTC, and a basic display:

image

buck54321 (changes requested):
Need to tweak the web asset IDs, but otherwise good.

I confirmed that the markets do not work on a mutli-market simnet server with old client. The end result is that you cannot place an order because it always errors for zero quantity. Would it be better to hide the order form and display a message? Can follow up too.

client/webserver/site/src/js/markets.js
```diff
@@ -535,17 +535,20 @@ export default class MarketsPage extends BasePage {
...
     this.market = {
       dex: dex,
-      sid: marketID(baseCfg.symbol, quoteCfg.symbol), // A string market identifier used by the DEX.
+      sid: mktId, // A string market identifier used by the DEX.
       // app.assets is a map of core.SupportedAsset type, which can be found at
       // client/core/types.go.
       base: app.assets[base],
       quote: app.assets[quote],
+      lotSize: dex.markets[mktId].lotsize,
+      rateStep: dex.markets[mktId].ratestep,
```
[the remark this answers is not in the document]
chappjc: Done, although I make the field `cfg` because `market.mktCfg` stuttered.  Anyway, the market config properties might be of use so the whole cfg is good to have.

buck54321:
Gonna tweak the cache busters?

buck54321 (approved):

PR #1103, opened 2021-06-08
client/front-end: Export orders as CSV.

### LINKED ISSUE #382: core: read out trade history in a portable format

martonp:
This PR adds a button on the orders page that allows the user to download their orders in CSV format.

Closes #382

### DIFF
--- a/client/webserver/site/src/js/orders.js
+++ b/client/webserver/site/src/js/orders.js
@@ -167,7 +167,10 @@ export default class OrdersPage extends BasePage {
   /* exportOrders opens a new page that downloads a csv of orders based on the
    * current filters */
   async exportOrders () {
-    console.log('wtf')
+    const currentFilter = { ...this.currentFilter() }
+    // return all orders instead of only the ones on the current page
+    currentFilter.n = 0
+    currentFilter.offset = ''
     const res = await postJSON('/api/orderscsv', this.currentFilter())
     if (res.ok) {
       // eslint-disable-next-line no-undef
@@ -180,6 +183,7 @@ export default class OrdersPage extends BasePage {
       a.download = 'dcrdex-orders.csv'
       a.click()
       window.URL.revokeObjectURL(url)
+      document.body.removeChild(a)
     }
   }
 
--- a/client/webserver/site/src/js/orders.js
+++ b/client/webserver/site/src/js/orders.js
@@ -18,7 +18,7 @@ export default class OrdersPage extends BasePage {
     this.loading = false
     const page = this.page = Doc.parsePage(main, [
       'rowTmpl', 'tableBody', 'hostFilter', 'assetFilter', 'statusFilter',
-      'orderLoader', 'ordersTable'
+      'orderLoader', 'ordersTable', 'exportOrders'
     ])
     this.orderTmpl = page.rowTmpl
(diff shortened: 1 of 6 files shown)


client/webserver/site/src/js/orders.js
```diff
@@ -149,25 +153,6 @@ export default class OrdersPage extends BasePage {
     filterState.assets = parseSubFilter(page.assetFilter)
     filterState.statuses = parseSubFilter(page.statusFilter)
 
-    const url = new URL(window.location)
```
martonp, line 152: I removed this since the updated query string is never used.

client/webserver/site/src/js/orders.js
```diff
@@ -179,6 +164,29 @@ export default class OrdersPage extends BasePage {
...
+   * current filters */
+  async exportOrders () {
+    const currentFilter = { ...this.currentFilter() }
+    // return all orders instead of only the ones on the current page
+    currentFilter.n = 0
+    currentFilter.offset = ''
+    const res = await postJSON('/api/orderscsv', this.currentFilter())
+    if (res.ok) {
+      // eslint-disable-next-line no-undef
+      const csvFile = new Blob([res.csv], { type: 'text/csv' })
```
martonp, line 177: A POST request was needed to be able to send a JSON object which could be deserialized as an OrderFilter. To trigger a download after a POST request, a fake link has to be created and clicked programatically. It's a bit hacky, but it seems to be the way most people do this.

chappjc (changes requested):
Thanks for picking this up.

The download would be best as a .csv file download served from a handler on the web router (e.g. /orders/export) rather than CSV-in-JSON from an api endpoint.

In the previous incomplete work on this (https://github.com/decred/dcrdex/pull/877), URL query parameters with a GET were used instead of the fetch api with a POST. That previous way allowed the download link (an `a` tag, but I expect it could be a button too) to be clicked and the user provided with a download (i.e. `"Content-Disposition", "attachment; filename=orders.csv"` and `"Content-Type", "text/csv"`).

One thing that was unresolved with the previous shot at this, and which this PR would also need to change, is to convert the handler to do a streaming response instead of creating the entire csv in memory and then writing it.  Please have a look at how dcrdata presently does this with `csv.NewWriter(w http.ResponseWriter)`: https://github.com/decred/dcrdata/blob/ec6b0e6a4cf4f7bcc9b803ce41787197bfffa94e/cmd/dcrdata/api/apiroutes.go#L1619-L1622

We may end up having to go further than this by actually rigging the DB methods with streaming versions that feed each matching entry into a closure that encodes and writes, rather than returning all matching entries and then encoding and writing in the handler.  But let's move forward with just the streaming writes and the existing DB methods for now.

The draft in https://github.com/decred/dcrdex/pull/877 was not actually far off. Please review that solution and the reviews there from @buck54321 and myself.

martonp:
@chappjc Yeah definitely better to stream it down than to have it all in memory. I will close this PR and open a new one.

PR #1106, opened 2021-06-11
Client must allow only one preimage request per order

LasTshaMAN:
This PR builds upon https://github.com/decred/dcrdex/pull/1077 to provide basic protection against malicious server behavior described in https://github.com/decred/dcrdex/issues/1020.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -5084,10 +5084,39 @@ func processPreimageRequest(c *Core, dc *dexConnection, reqID uint64, oid order.
 	// match_proof with this order has the same checksum. If it does not, the
 	// server may have used the knowledge of this preimage we are sending them
 	// now to alter the epoch shuffle.
+	//
+	// Allow to initialize csum just once per order to prevent possible
+	// malicious behavior.
 	tracker.mtx.Lock()
 	if isCancel {
+		if tracker.cancel.csum != nil {
+			csumErr := errors.New("csum is already initialized")
+			resp, err := msgjson.NewResponse(reqID, nil,
+				msgjson.NewError(msgjson.InvalidRequestError, csumErr.Error()))
+			if err != nil {
+				return fmt.Errorf("preimage response encoding error: %w", err)
+			}
+			err = dc.Send(resp)
+			if err != nil {
+				return fmt.Errorf("preimage send error: %w", err)
+			}
+			return csumErr
+		}
 		tracker.cancel.csum = commitChecksum
 	} else {
+		if tracker.csum != nil {
+			csumErr := errors.New("csum is already initialized")
+			resp, err := msgjson.NewResponse(reqID, nil,
+				msgjson.NewError(msgjson.InvalidRequestError, csumErr.Error()))
+			if err != nil {
+				return fmt.Errorf("preimage response encoding error: %w", err)
+			}
+			err = dc.Send(resp)
+			if err != nil {
+				return fmt.Errorf("preimage send error: %w", err)
+			}
+			return csumErr
+		}
 		tracker.csum = commitChecksum
 	}
 	tracker.mtx.Unlock()
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -5085,7 +5085,7 @@ func processPreimageRequest(c *Core, dc *dexConnection, reqID uint64, oid order.
 	// server may have used the knowledge of this preimage we are sending them
 	// now to alter the epoch shuffle.
 	//
-	// Allow to initialize csum just once per order to prevent possible
+	// Allow to initialize csum only once per order to prevent possible
 	// malicious behavior.
 	tracker.mtx.Lock()
 	if isCancel {
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -5080,9 +5080,21 @@ func processPreimageRequest(c *Core, dc *dexConnection, reqID uint64, oid order.
 		return fmt.Errorf("no active order found for preimage request for %s", oid)
 	}
 
-	err := acceptCsum(dc, reqID, tracker, isCancel, commitChecksum)
-	if err != nil {
-		return fmt.Errorf("accept csum: %w", err)
+	// Record the csum if this preimage request is novel, and deny it if this is
+	// a duplicate request with an altered csum.
+	if !acceptCsum(tracker, isCancel, commitChecksum) {
+		csumErr := errors.New("invalid csum in duplicate preimage request")
+		resp, err := msgjson.NewResponse(reqID, nil,
+			msgjson.NewError(msgjson.InvalidRequestError, csumErr.Error()))
+		if err != nil {
+			c.log.Errorf("Failed to encode response to denied preimage request: %v", err)
+			return csumErr
+		}
+		err = dc.Send(resp)
+		if err != nil {
+			c.log.Errorf("Failed to send response to denied preimage request: %v", err)
+		}
+		return csumErr
 	}
 
 	// Clean up the sentCommits now that we loaded the commitment. This can be
@@ -5114,48 +5126,28 @@ func processPreimageRequest(c *Core, dc *dexConnection, reqID uint64, oid order.
 // acceptCsum will record the commitment checksum so we can verify that the
 // subsequent match_proof with this order has the same checksum. If it does not,
 // the server may have used the knowledge of this preimage we are sending them
-// now to alter the epoch shuffle.
-func acceptCsum(dc *dexConnection, reqID uint64, tracker *trackedTrade, isCancel bool, commitChecksum dex.Bytes) error {
-	var respondWithErr bool
-
+// now to alter the epoch shuffle. The return value is false if a previous
+// checksum has been recorded that differs from the provided one.
+func acceptCsum(tracker *trackedTrade, isCancel bool, commitChecksum dex.Bytes) bool {
 	// Do not allow csum to be changed once it has been committed to
 	// (initialized to something other than `nil`) because it is probably a
 	// malicious behavior by the server.
 	tracker.mtx.Lock()
+	defer tracker.mtx.Unlock()
+
 	if isCancel {
-		if tracker.cancel.csum != nil {
-			if !bytes.Equal(commitChecksum, tracker.cancel.csum) {
-				respondWithErr = true
-			}
-		} else {
+		if tracker.cancel.csum == nil {
 			tracker.cancel.csum = commitChecksum
+			return true
 		}
-	} else {
-		if tracker.csum != nil {
-			if !bytes.Equal(commitChecksum, tracker.csum) {
-				respondWithErr = true
-			}
-		} else {
-			tracker.csum = commitChecksum
-		}
+		return bytes.Equal(commitChecksum, tracker.cancel.csum)
 	}
-	tracker.mtx.Unlock()
-
-	if respondWithErr {
-		csumErr := errors.New("csum is already initialized")
-		resp, err := msgjson.NewResponse(reqID, nil,
-			msgjson.NewError(msgjson.InvalidRequestError, csumErr.Error()))
-		if err != nil {
-			return fmt.Errorf("preimage response encoding error: %w", err)
-		}
-		err = dc.Send(resp)
-		if err != nil {
-			return fmt.Errorf("preimage send error: %w", err)
-		}
-		return csumErr
+	if tracker.csum == nil {
+		tracker.csum = commitChecksum
+		return true
 	}
 
-	return nil
+	return bytes.Equal(commitChecksum, tracker.csum)
 }
 
 // handleMatchRoute processes the DEX-originating match route request,
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -5080,43 +5080,10 @@ func processPreimageRequest(c *Core, dc *dexConnection, reqID uint64, oid order.
 		return fmt.Errorf("no active order found for preimage request for %s", oid)
 	}
 
-	checkCsum := func(csum dex.Bytes) error {
-		if csum != nil {
-			csumErr := errors.New("csum is already initialized")
-			resp, err := msgjson.NewResponse(reqID, nil,
-				msgjson.NewError(msgjson.InvalidRequestError, csumErr.Error()))
-			if err != nil {
-				return fmt.Errorf("preimage response encoding error: %w", err)
-			}
-			err = dc.Send(resp)
-			if err != nil {
-				return fmt.Errorf("preimage send error: %w", err)
-			}
-			return csumErr
-		}
-		return nil
+	err := acceptCsum(dc, reqID, tracker, isCancel, commitChecksum)
+	if err != nil {
+		return fmt.Errorf("accept csum: %w", err)
 	}
-
-	// Record the commitment checksum so we can verify that the subsequent
-	// match_proof with this order has the same checksum. If it does not, the
-	// server may have used the knowledge of this preimage we are sending them
-	// now to alter the epoch shuffle.
-	//
-	// Allow to initialize csum only once per order to prevent possible
-	// malicious behavior.
-	tracker.mtx.Lock()
-	if isCancel {
-		if err := checkCsum(tracker.cancel.csum); err != nil {
-			return err
-		}
-		tracker.cancel.csum = commitChecksum
-	} else {
-		if err := checkCsum(tracker.csum); err != nil {
-			return err
-		}
-		tracker.csum = commitChecksum
-	}
-	tracker.mtx.Unlock()
 
 	// Clean up the sentCommits now that we loaded the commitment. This can be
 	// removed when the old piSyncers method is removed.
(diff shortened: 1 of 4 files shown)


JoeGruffins (approved):
Looks ok to me. Just one suggestion.

client/core/core.go
```diff
@@ -5084,10 +5084,39 @@ func processPreimageRequest(c *Core, dc *dexConnection, reqID uint64, oid order.
...
+				msgjson.NewError(msgjson.InvalidRequestError, csumErr.Error()))
+			if err != nil {
+				return fmt.Errorf("preimage response encoding error: %w", err)
+			}
+			err = dc.Send(resp)
+			if err != nil {
+				return fmt.Errorf("preimage send error: %w", err)
+			}
+			return csumErr
+		}
```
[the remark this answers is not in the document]
LasTshaMAN: yeah, thought about it too, 

refactored.

client/core/core.go
```diff
@@ -5080,14 +5080,40 @@ func processPreimageRequest(c *Core, dc *dexConnection, reqID uint64, oid order.
...
 	// match_proof with this order has the same checksum. If it does not, the
 	// server may have used the knowledge of this preimage we are sending them
 	// now to alter the epoch shuffle.
+	//
+	// Allow to initialize csum only once per order to prevent possible
+	// malicious behavior.
 	tracker.mtx.Lock()
 	if isCancel {
+		if err := checkCsum(tracker.cancel.csum); err != nil {
+			return err
```
chappjc, line 5110: The tracker.mtx is still locked if you return here.

chappjc (changes requested):

client/core/core.go
LasTshaMAN: Good catch ;) Maybe there is already enough `csum`-relted logic to move it to a separate `func` ? 

Fixed by doing just that.

client/core/core.go
```diff
@@ -5118,6 +5111,49 @@ func processPreimageRequest(c *Core, dc *dexConnection, reqID uint64, oid order.
...
+			if err != nil {
+				return fmt.Errorf("preimage send error: %w", err)
+			}
+			return csumErr
+		}
+		return nil
+	}
+
+	tracker.mtx.Lock()
+	defer tracker.mtx.Unlock()
```
LasTshaMAN, line 5133: Just want to point out that `err = dc.Send(resp)` (which currently has a timeout of `writeWait = time.Second * 3`) might happen when `mtx` is locked, it is usually a bad practice to hold `mtx` for too long (doing network calls and such), 

so this might be an attack vector for a `server` to cause a DOS to `client`, any thoughts on that ?

LasTshaMAN: I rewrote the implementation a bit to solve this issue.

chappjc: Thanks, I would have commented on that.  Def want to avoid high latency ops with this mutex held.  Will review your latest solution shortly.

client/core/core.go
```diff
@@ -5118,6 +5111,48 @@ func processPreimageRequest(c *Core, dc *dexConnection, reqID uint64, oid order.
...
+	// malicious behavior.
+	tracker.mtx.Lock()
+	if isCancel {
+		if tracker.cancel.csum != nil {
+			respondWithErr = true
+		} else {
+			tracker.cancel.csum = commitChecksum
+		}
+	} else {
+		if tracker.csum != nil {
```
chappjc, line 5131: Checking `!= nil` is one possibility I suggested in the previous PR discussion.  Another is non-nil *and* `!bytes.Equal`, which would allow client to respond with the order's preimage as long as the server didn't change up the csum from the initial request.

While non-nil is a good indicator that server has already asked and something is fishy, I can see the latter being a legit case if server has to retry the request because of perhaps a ill-timed restart or even a timeout on a previously request because of a broken connection (note that the csum is stored by the client in tracker *before* attempting to send the `PreimageResponse` and that `Send` could fail - server is not explicitly prohibited from asking again).

Essentially, if the csum field is already set, we clearly don't need to rewrite it, but we don't have to reject/ignore the requests unless the csum would change value.

chappjc (commented):
Two things to discuss:

1. Distinguishing duplicate preimage requests from preimage requests with altered commitment checksums.  Only the latter is really a problem, but the former is unusual and still deserves a warning.
2. Reasoning for client responding at all in the above problem case.  Arguably a normal response is alright for dup requests as long as csum wasn't altered.  And if csum was altered, what is the purpose of a response with an `InvalidRequestError` error in the payload?  `handlePreimageResp` doesn't understand it presently (no `msg.UnmarshalResult` or check of `resp.Error` there), but it will at least finish the preimage request instead of causing it to timeout.

client/core/core.go
```diff
@@ -5118,6 +5111,48 @@ func processPreimageRequest(c *Core, dc *dexConnection, reqID uint64, oid order.
...
+			msgjson.NewError(msgjson.InvalidRequestError, csumErr.Error()))
+		if err != nil {
+			return fmt.Errorf("preimage response encoding error: %w", err)
+		}
+		err = dc.Send(resp)
+		if err != nil {
+			return fmt.Errorf("preimage send error: %w", err)
+		}
+		return csumErr
+	}
```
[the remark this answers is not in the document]
LasTshaMAN: > What's the reasoning for any response at all?

The reasoning is - having some response instead of no response or unknown response would make testing/debugging any related issues easier for ourselves.

> It would seem to be a courtesy to a server that is misbehaving.

Since this code is open source I see no value in security through obscurity here.

> That's the present approach.

Although I wasn't able to implement it properly on my first try, I'll prefer the approach 4) "returning a meaningful error", 
if not then returning no response and just timing out seems like the simplest to implement (and cleanest to understand) option among those you've listed above, so I would implement it instead.

client/core/core.go
LasTshaMAN: Makes sense, will implement this soon.

client/core/core.go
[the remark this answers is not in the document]
chappjc: > Since this code is open source I see no value in security through obscurity here

I'm not implying it's a security benefit to not respond, but responding does allow the server issuing the request to move on with the epoch matching sooner instead of waiting for a timeout, which is something of an undeserved gift to a server that is misbehaving.  I'm OK with a response at this point because it does assist debugging and minimize disruption due to just bugs, but I had envisioned just not responding...  Just wanted to point out that if we get to a point where it's clear that if this does happen that it is genuinely malicious, not a bug, then client probably should give server the cold shoulder.

> Although I wasn't able to implement it properly on my first try, I'll prefer the approach 4) "returning a meaningful error",
if not then returning no response and just timing out seems like the simplest to implement (and cleanest to understand) option among those you've listed above, so I would go for it instead

You're pretty close to this.  As I suggested in the main comment for this review, the `handlePreimageResp` method can get modified to check `if resp.Error != nil {` before `json.Unmarshal`, which would probably be the minimal code change.  This is similar to `msg.UnmarshalResult`, but I don't think we always want to `m.repsondError` (see below).

Note that in `handlePreimageResp`, all the `m.respondError` calls are kinda dumb because client won't find a response handler since this isn't a client-originating request, but it will at least log something at the client.  But if it detects a non-nil `resp.Error` field between `msg.Response` and `json.Unmarshal`, that condition should omit the `m.respondError` because the client needs no further response.

i.e.

```diff
diff --git a/server/market/market.go b/server/market/market.go
index 31118188..74d62ae5 100644
--- a/server/market/market.go
+++ b/server/market/market.go
@@ -1533,6 +1533,11 @@ func (m *Market) handlePreimageResp(msg *msgjson.Message, reqData *piData) {
                        "error parsing preimage notification response")
                return
        }
+       if resp.Error != nil {
+               log.Warnf("Client failed to handle preimage request: %v", resp.Error)
+               sendPI(nil) // thanks for the response, client, we can move on now
+               return
+       }
        err = json.Unmarshal(resp.Result, &piResp)
        if err != nil {
                sendPI(nil)

```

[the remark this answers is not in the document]
LasTshaMAN: > if we get to a point where it's clear that if this does happen that it is genuinely malicious, not a bug, then client probably should give server the cold shoulder.

Good point, although I don't have any ideas at the moment on how to reliably differentiate between a bug and malicious behavior.

Otherwise very thorough description for what is going on, thanks for that! I've pushed these changes (added a test for these as well).

client/core/core.go
LasTshaMAN: Done.

client/core/core_test.go
```diff
@@ -2624,10 +2635,142 @@ func TestHandlePreimageRequest(t *testing.T) {
...
+		// i.e. "Received preimage request for %v with no corresponding order submission response! Waiting..."
+		close(commitSig) // pretend like the order submission just finished
+
+		select {
+		case note := <-notes:
+			t.Fatalf("got unexpected note: %v", note)
+		case <-time.After(time.Millisecond):
+			// We've waited enough for a note to come, hopefully it won't after
+			// this point in time.
+		}
```
LasTshaMAN, line 2695: Reading through this code once again I've realized the following tests:
- `more than one preimage request for order (different csums)`
- `more than one preimage request for cancel order (different csums)`

are flaky as is currently implemented, 
I found some usage of `time.Sleep` in the tests in `dcrdex`, so I asume it is an accepted approach - "to allow for flakiness when it is hard/time-consuming to implement otherwise" ?

Should I try to address this, or we better move on ?

chappjc: Indeed sleeps in test are generally to be avoided, and were often a thorn in my side early on.  These "different csums" tests are tricky of course because you are verifying the _absence_ of an event, and the tested code is launched by `handlePreimageRequest` in a (async) goroutine.

You could hack a channel send into `(conn *TWebsocket) Send` perhaps, and receive on that like the other tests receive on the `notes` chan.  A 250ms wait here in this select isn't terrible either, but 1 ms is too short.  My only hesitation with the channel hacks, despite being an effective sync mechanism, is that they make the tests difficult to change in the future.

Another suggestion I have is to keep the NewResponse/Send in processPreimage, which will both reduce the input args to `acceptCsum` and allow a defer to unlock the tracker mutex.

The gist of the ideas:

```diff
diff --git a/client/core/core.go b/client/core/core.go
index e73c82b7..b2f9ce9b 100644
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -5080,11 +5080,6 @@ func processPreimageRequest(c *Core, dc *dexConnection, reqID uint64, oid order.
                return fmt.Errorf("no active order found for preimage request for %s", oid)
        }
 
-       err := acceptCsum(dc, reqID, tracker, isCancel, commitChecksum)
-       if err != nil {
-               return fmt.Errorf("accept csum: %w", err)
-       }
-
        // Clean up the sentCommits now that we loaded the commitment. This can be
        // removed when the old piSyncers method is removed.
        defer func() {
@@ -5093,6 +5088,24 @@ func processPreimageRequest(c *Core, dc *dexConnection, reqID uint64, oid order.
                delete(c.sentCommits, preImg.Commit()) // redundant if the commitment was in request
                c.sentCommitsMtx.Unlock()
        }()
+
+       // Record the csum if this preimage request is novel, and deny it if this is
+       // a duplicate request with an altered csum.
+       if !acceptCsum(tracker, isCancel, commitChecksum) {
+               csumErr := errors.New("invalid csum in duplicate preimage request")
+               resp, err := msgjson.NewResponse(reqID, nil,
+                       msgjson.NewError(msgjson.InvalidRequestError, csumErr.Error()))
+               if err != nil {
+                       c.log.Errorf("Failed to encode response to denied preimage request: %v", err)
+                       return csumErr
+               }
+               err = dc.Send(resp)
+               if err != nil {
+                       c.log.Errorf("Failed to send response to denied preimage request: %v", err)
+               }
+               return csumErr
+       }
+
        resp, err := msgjson.NewResponse(reqID, &msgjson.PreimageResponse{
                Preimage: preImg[:],
        }, nil)
@@ -5114,48 +5127,27 @@ func processPreimageRequest(c *Core, dc *dexConnection, reqID uint64, oid order.
 // acceptCsum will record the commitment checksum so we can verify that the
 // subsequent match_proof with this order has the same checksum. If it does not,
 // the server may have used the knowledge of this preimage we are sending them
-// now to alter the epoch shuffle.
-func acceptCsum(dc *dexConnection, reqID uint64, tracker *trackedTrade, isCancel bool, commitChecksum dex.Bytes) error {
-       var respondWithErr bool
-
+// now to alter the epoch shuffle. The return value is false if a previous
+// checksum has been recorded that differs from the provided one.
+func acceptCsum(tracker *trackedTrade, isCancel bool, commitChecksum dex.Bytes) bool {
        // Do not allow csum to be changed once it has been committed to
        // (initialized to something other than `nil`) because it is probably a
        // malicious behavior by the server.
        tracker.mtx.Lock()
+       defer tracker.mtx.Unlock()
        if isCancel {
-               if tracker.cancel.csum != nil {
-                       if !bytes.Equal(commitChecksum, tracker.cancel.csum) {
-                               respondWithErr = true
-                       }
-               } else {
+               if tracker.cancel.csum == nil {
                        tracker.cancel.csum = commitChecksum
+                       return true
                }
-       } else {
-               if tracker.csum != nil {
-                       if !bytes.Equal(commitChecksum, tracker.csum) {
-                               respondWithErr = true
-                       }
-               } else {
-                       tracker.csum = commitChecksum
-               }
+               return bytes.Equal(commitChecksum, tracker.cancel.csum)
        }
-       tracker.mtx.Unlock()
 
-       if respondWithErr {
-               csumErr := errors.New("csum is already initialized")
-               resp, err := msgjson.NewResponse(reqID, nil,
-                       msgjson.NewError(msgjson.InvalidRequestError, csumErr.Error()))
-               if err != nil {
-                       return fmt.Errorf("preimage response encoding error: %w", err)
-               }
-               err = dc.Send(resp)
-               if err != nil {
-                       return fmt.Errorf("preimage send error: %w", err)
-               }
-               return csumErr
+       if tracker.csum == nil {
+               tracker.csum = commitChecksum
+               return true
        }
-
-       return nil
+       return bytes.Equal(commitChecksum, tracker.csum)
 }
```

```diff
diff --git a/client/core/core_test.go b/client/core/core_test.go
index 4eb1ae22..14982369 100644
--- a/client/core/core_test.go
+++ b/client/core/core_test.go
@@ -136,12 +136,13 @@ var (
 )
 
 type TWebsocket struct {
-       mtx        sync.RWMutex
-       id         uint64
-       sendErr    error
-       reqErr     error
-       connectErr error
-       msgs       <-chan *msgjson.Message
+       mtx            sync.RWMutex
+       id             uint64
+       sendErr        error
+       sendMsgErrChan chan *msgjson.Error
+       reqErr         error
+       connectErr     error
+       msgs           <-chan *msgjson.Message
        // handlers simulates a peer (server) response for request, and handles the
        // response with the msgFunc.
        handlers map[string][]func(*msgjson.Message, msgFunc) error
@@ -229,7 +230,20 @@ func (conn *TWebsocket) NextID() uint64 {
        conn.id++
        return conn.id
 }
-func (conn *TWebsocket) Send(msg *msgjson.Message) error { return conn.sendErr }
+func (conn *TWebsocket) Send(msg *msgjson.Message) error {
+       if conn.sendMsgErrChan != nil {
+               resp, err := msg.Response()
+               if err != nil {
+                       return err
+               }
+               if resp.Error != nil {
+                       conn.sendMsgErrChan <- resp.Error
+                       return nil // the response was sent successfully
+               }
+       }
+
+       return conn.sendErr
+}
 func (conn *TWebsocket) Request(msg *msgjson.Message, f msgFunc) error {
        return conn.RequestWithTimeout(msg, f, 0, func() {})
 }
@@ -2674,7 +2688,10 @@ func TestHandlePreimageRequest(t *testing.T) {
                }
                reqCommit, _ := msgjson.NewRequest(rig.dc.NextID(), msgjson.PreimageRoute, payload)
 
-               notes := rig.core.NotificationFeed()
+               // Prepare to have processPreimageRequest respond with a payload with
+               // the Error field set.
+               rig.ws.sendMsgErrChan = make(chan *msgjson.Error, 1)
+               defer func() { rig.ws.sendMsgErrChan = nil }()
 
                rig.dc.trades[oid] = tracker
                err := handlePreimageRequest(rig.core, rig.dc, reqCommit)
@@ -2687,11 +2704,12 @@ func TestHandlePreimageRequest(t *testing.T) {
                close(commitSig) // pretend like the order submission just finished
 
                select {
-               case note := <-notes:
-                       t.Fatalf("got unexpected note: %v", note)
-               case <-time.After(time.Millisecond):
-                       // We've waited enough for a note to come, hopefully it won't after
-                       // this point in time.
+               case msgErr := <-rig.ws.sendMsgErrChan:
+                       if msgErr.Code != msgjson.InvalidRequestError {
+                               t.Fatalf("expected error code %d got %d", msgjson.InvalidRequestError, msgErr.Code)
+                       }
+               case <-time.After(time.Second):
+                       t.Fatal("no msgjson.Error sent from preimage request handling")
                }
 
                tracker.mtx.RLock()
```

LasTshaMAN: > My only hesitation with the channel hacks, despite being an effective sync mechanism, is that they make the tests difficult to change in the future.

Noted on that. Since you've already sketched out the implementation, I'd rather use it here :)

> Another suggestion I have is to keep the NewResponse/Send in processPreimage, which will both reduce the input args to acceptCsum and allow a defer to unlock the tracker mutex.

Yeah, looks cleaner indeed, applied.

chappjc (approved):
OK.  `handlePreimageRequest` is working as expected if it has already a csum recorded.

Just a note however that because of  how `handlePreimageRequest` currently cleans up the `c.sentCommits` map after receiving a request, the client is not currently going to handle duplicate requests in practice, rather erroring like `Received preimage request for order %v with known commitment %v"`.  This is good for now as server does not make duplicate requests, but in terms of logic, `acceptCsum` is performing as it should be.  The `sentCommits` mechanism could change and we need an `acceptCsum` kind of check in place to ensure we can't let the server cheat.  For example the `delete(c.sentCommits, commit)` could easily be deferred for `handleMatchProofMsg` to allow `handlePreimageRequest` to be idempotent.  No need to pursue this now however.

Thanks, @LasTshaMAN .

PR #1108, opened 2021-06-11
[0.2] assets: use script versions meticulously

chappjc:
`gettxout` version is the transaction version, not script, and there is no `GetTxOutResult.ScriptPubKey.Version` in dcrd 1.6.  So, for audit with dcrd 1.6 we have no choice but to pull blocks to get the raw transaction if the transaction is confirmed.  If it is still unconfirmed we can use `getrawtransaction`.

harness beta node now does not enable txindex

### DIFF
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -1575,6 +1575,56 @@ func (dcr *ExchangeWallet) SignMessage(coin asset.Coin, msg dex.Bytes) (pubkeys,
 	return pubkeys, sigs, nil
 }
 
+// verboseTxTheHardWay gets a TxRawResult for the given tx with a known number
+// of confirmations as of a certain best block. This is required for
+// dcrd/dcrwallet 1.6 without txindex enabled.
+func (dcr *ExchangeWallet) verboseTxTheHardWay(txHash *chainhash.Hash, confs int64, bestBlock string) (*chainjson.TxRawResult, error) {
+	if confs > 0 {
+		bestBlockHash, err := chainhash.NewHashFromStr(bestBlock)
+		if err != nil {
+			return nil, fmt.Errorf("error decoding best block hash %s: %w",
+				bestBlock, err)
+		}
+		block, err := dcr.node.GetBlockVerbose(dcr.ctx, bestBlockHash, true)
+		if err != nil {
+			return nil, fmt.Errorf("error getting block header: %w", translateRPCCancelErr(err))
+		}
+		if confs > 1 { // 1 means it is in the best block above
+			txHeight := block.Height - confs + 1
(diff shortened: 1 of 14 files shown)


client/asset/dcr/dcr.go
```diff
@@ -1596,13 +1596,76 @@ func (dcr *ExchangeWallet) AuditContract(coinID, contract, txData dex.Bytes) (*a
 	if txOut == nil {
 		return nil, asset.CoinNotFoundError
 	}
+
+	// Since txOut.Version with dcrd 1.6 is the transaction version NOT the
+	// output script version, we need to pull the entire transaction to get it.
+	// Unfortunately, for confirmed transactions we cannot do that with
+	// getrawtransaction without txindex. If confirmed, use
+	// GetTxOutResult.Confirmations and BestBlock to pull the block that is
+	// expected to include this tx.
```
chappjc, line 1605: So unfortunate, but there's no other way to get the output script's version, and we really need it.

buck54321:
Why can't we just specify that DEX uses version 0 scripts?

chappjc:
While we will be making txns with v0 output scripts, the audits of third-party txns need to consider the script version since they can be included in blocks with other versions, which should make them fail audit.

client/asset/dcr/dcr.go
```diff
@@ -1596,13 +1596,76 @@ func (dcr *ExchangeWallet) AuditContract(coinID, contract, txData dex.Bytes) (*a
...
+		bestBlockHash, err := chainhash.NewHashFromStr(txOut.BestBlock)
+		if err != nil {
+			return nil, fmt.Errorf("error decoding best block hash %s: %w",
+				txOut.BestBlock, err)
+		}
+		block, err := dcr.node.GetBlockVerbose(dcr.ctx, bestBlockHash, true)
+		if err != nil {
+			return nil, fmt.Errorf("error getting block header: %w", translateRPCCancelErr(err))
+		}
+		if txOut.Confirmations > 0 { // 1 means it is in the best block above
```
buck54321, line 1623:
```Go
if txOut.Confirmations > 1 {
```
?

buck54321 (changes requested):
Test well on latest 1.6 release tags of **dcrwallet** and **dcrd**. Just the one comment.

client/asset/dcr/dcr.go
chappjc: Yes!  I hacked it like this temporarily to test this, but didn't change back

JoeGruffins (approved):
Except for the confirmations thing.

chappjc:
Simnet tests for client now verify each path of the `verboseTxTheHardWay` code. The confirmations thing is also fixed.

PR #1109, opened 2021-06-14
client/front-end: Export orders in csv file

### LINKED ISSUE #382: core: read out trade history in a portable format

martonp:
This PR adds a button on the orders page that allows the user to download their orders in CSV format.

Closes #382

### DIFF
--- a/client/webserver/http.go
+++ b/client/webserver/http.go
@@ -10,6 +10,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	"decred.org/dcrdex/client/core"
 	"decred.org/dcrdex/dex"
@@ -253,16 +254,17 @@ func (s *WebServer) handleExportOrders(w http.ResponseWriter, r *http.Request) {
 		"Host",
 		"Base",
 		"Quote",
-		"Ask",
-		"Offer",
+		"Base Quantity",
+		"Order Rate",
+		"Actual Rate",
+		"Base Fees",
+		"Quote Fees",
 		"Type",
 		"Side",
 		"Time in Force",
 		"Status",
-		"Rate",
 		"Filled (%)",
 		"Settled (%)",
-		"Quantity",
 		"Time",
 	})
 	if err != nil {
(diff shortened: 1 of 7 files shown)


client/webserver/http.go
```diff
@@ -190,6 +195,125 @@ func (s *WebServer) handleOrders(w http.ResponseWriter, r *http.Request) {
...
+		log.Errorf("error parsing form for export order: %v", err)
+		http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
+		return
+	}
+
+	filter.Hosts = r.Form["hosts"]
+	assets := r.Form["assets"]
+	filter.Assets = make([]uint32, len(assets))
+	for k, assetStrID := range assets {
+		assetNumID, err := strconv.Atoi(assetStrID)
```
JoeGruffins, line 218: Maybe https://golang.org/pkg/strconv/#ParseUint is slightly better in this case and for the status id, since they shouldn't be negative.

JoeGruffins (commented):
Working great for me.

I noticed that the time is local when viewing orders on the browser, but it UTC in the csv. I would prefer csv be the same time zone as the orders screen.

martonp:
Local time sounds good to me.

client/webserver/http.go
JoeGruffins: `strconv.ParseUint(assetStrID, 10, 32)` Is probably better for the uint32, and order status is a uint16, so `strconv.ParseUint(assetStrID, 10, 16)`. I actually didn't know 0 will do base 10, but looks ok trying it out in the playground.

martonp: 0 will do base 10 unless the prefix in the string matches another base (0x for base 16). These should always be 10 though, so I've updated it to 10.

JoeGruffins: Good to know! 0 was probably fine then, but probably don't expect hex for these smallish numbers.

JoeGruffins (approved):

client/webserver/http.go
```diff
@@ -190,6 +195,126 @@ func (s *WebServer) handleOrders(w http.ResponseWriter, r *http.Request) {
...
+	err = csvWriter.Error()
+	if err != nil {
+		log.Errorf("error writing CSV: %v", err)
+		return
+	}
+	wf.Flush()
+
+	for _, ord := range ords {
+		ordReader := orderReader{ord}
+		timestamp, err := encode.UnixTimeMilliLocal(int64(ord.Stamp)).MarshalText()
```
chappjc, line 282: `MarshalText` just applies `RFC3339Nano` formatting, so i'd suggest just doing `.Format(time.RFC3339Nano)`.  RFC3339 is not the most human friendly format, but it's standardized and good for parsing, so I think for CSV it's good.

Also, since we know these stamps are only millisecond precision, we could specify the format accordingly like `2006-01-02T15:04:05.999Z07:00`.

dex/encode/encode.go
```diff
@@ -95,6 +95,14 @@ func UnixTimeMilli(msEpoch int64) time.Time {
 	return time.Unix(sec, msec*1e6).UTC()
 }
 
+// UnixTimeMilliLocal returns a Time for an elapsed time in milliseconds since
+// the Unix Epoch. The time will have Location set to the local machine time.
+func UnixTimeMilliLocal(msEpoch int64) time.Time {
+	sec := msEpoch / 1000
+	msec := msEpoch % 1000
+	return time.Unix(sec, msec*1e6).Local()
```
chappjc, line 103: I think we can do without `UnixTimeMilliLocal` and just call like `UnixTimeMilli(stamp).Local()`.  Might as well keep the `dex/encode` package API smaller since this is mostly dup code. and `time.Time`s can change locations back and forth with no trouble.

chappjc (changes requested):
Looking pretty good. Two quick comments before I give it a test.

client/webserver/http.go
```diff
@@ -190,6 +195,126 @@ func (s *WebServer) handleOrders(w http.ResponseWriter, r *http.Request) {
...
+			ordReader.AskString(),        // Ask
+			ordReader.OfferString(),      // Offer
+			ordReader.Type.String(),      // Type
+			ordReader.SideString(),       // Side
+			ord.TimeInForce.String(),     // Time in Force
+			ordReader.StatusString(),     // Status
+			ordReader.SimpleRateString(), // Rate
+			ordReader.FilledPercent(),    // Filled
+			ordReader.SettledPercent(),   // Settled
+			ordReader.OfferString(),      // Quantity
```
chappjc, line 301: "Offer" and "Quantity" are both set to `ordReader.OfferString()`. Was "Quantity" supposed to be effectively `precision8(ord.Qty)`?

martonp: I just copied this from the previous PR, but I've made some changes to the columns now. Let me know what you think.

chappjc: Thanks, I don't know what was going on with the previous columns.  I like this average rate you added, which makes sense when there are multiple matches that aren't necessarily same rate.  Really nice.

chappjc (approved):

chappjc:
We'll probably iterate on the csv format going forward, but this is working great and I have no complaints.

PR #1111, opened 2021-06-21
site: dep updates for bugfixes and webpack 5 

chappjc:
The audit failures came up recently, and the [**postcss** vuln fix is available with **stylelint** now](https://github.com/stylelint/stylelint/issues/4942) and the css-minimizer-webpack-plugin>svgo>css-select-css-what [fix is available](https://github.com/svg/svgo/pull/1485#event-4942679039). 

The package-lock.json update is committed as a dummy user in a separate commit, as I have done in the past.

A notable change from the Dart **sass** update is that it now complains about **bootstrap** using the `/` operator instead of `math.div`.  This is [fixed now in bootstrap 5](https://github.com/twbs/bootstrap/pull/34245), but it's just a warning. See https://github.com/twbs/bootstrap/issues/34051 for discussion on this fairly contentious sass change. Still, we should look into bootstrap 5 at some point.

This also removes **webpack-dev-server** as that was never used with dex (or dcrdata where this was taken from).


chappjc:
@martonp I should have put this up sooner but I was waiting for some upstream fixes.  Still not everything I was hoping for though.  In any case, CI on your PR ended up working so it looks like it was a transient issue after all.

martonp:
@chappjc Why is it misleading?

chappjc:
In terms of actual work done on a repository.

chappjc:
svgo [just got patched](https://github.com/svg/svgo/pull/1485#event-4942679039) so now npm audit passes.

JoeGruffins:
> In terms of actual work done on a repository.

imo super not important, besides, there would be not repository w/o you
imo don't waste you time on fake users

JoeGruffins (approved):
Builds and runs well. Also see the deprecated warnings about slashes.

amassarwi:
> > In terms of actual work done on a repository.
> 
> imo super not important, besides, there would be not repository w/o you
> imo don't waste you time on fake users

💯

PR #1112, opened 2021-06-22
mulit: Move shutdown prompt to app.

### LINKED ISSUE #1104: Remove (*Core).PromptShutdown
> ... what would be proper is for a PromptShutdown to be an app function (not a Core method) that uses Logout and then cancels the context that the app gave to Core, but that's for another day.

from @chappjc https://github.com/decred/dcrdex/pull/1069#discussion_r619368899

JoeGruffins:
closes #1104

### DIFF
--- a/client/cmd/dexc/main.go
+++ b/client/cmd/dexc/main.go
@@ -189,7 +189,7 @@ func promptShutdown(clientCore *core.Core) bool {
 	}
 	if !errors.Is(err, core.ActiveOrdersLogoutErr) {
 		log.Errorf("unable to logout: %v", err)
-		return false
+		return true
 	}
 
 	fmt.Print("You have active orders. Shutting down now may result in failed swaps and account penalization.\n" +
--- a/client/cmd/dexc/main.go
+++ b/client/cmd/dexc/main.go
@@ -4,12 +4,15 @@
 package main
 
 import (
+	"bufio"
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/signal"
 	"runtime"
 	"runtime/pprof"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


chappjc:
Just what I had in mind. Nice clean change.  Thanks, @JoeGruffins.

client/cmd/dexc/main.go
```diff
@@ -175,3 +178,36 @@ func mainCore() error {
...
+// shutdown if there are. The return value indicates if it is safe to stop Core
+// or if the user has confirmed they want to shutdown with active orders.
+func promptShutdown(clientCore *core.Core) bool {
+	err := clientCore.Logout()
+	if err == nil {
+		return true
+	}
+	if !errors.Is(err, core.ActiveOrdersLogoutErr) {
+		log.Errorf("unable to logout: %v", err)
+		return false
```
JoeGruffins, line 192: true?

client/cmd/dexc/main.go
```diff
@@ -175,3 +178,36 @@ func mainCore() error {
...
+	}
+
+	fmt.Print("You have active orders. Shutting down now may result in failed swaps and account penalization.\n" +
```
JoeGruffins, line 195: Change fmt prints to log?

client/cmd/dexc/main.go
chappjc: Yah, if it's not `hasActiveOrders`, just some error locking a wallet, OK to log the issue and say true (shutdown is safe).

client/cmd/dexc/main.go
chappjc: I kinda like this just an unstamped stdout message to go with the stdin scanner.  I don't see a lot of value in this prompt being logged. The shutdown messages that follow are logged so that seems good to me.

chappjc (changes requested):
Looks good. I think a true return for a wallet lock error or anything other than active orders is right though, as you hinted.

PR #1113, opened 2021-06-25
Fix precision out of range when redrawing chart

### LINKED ISSUE #1056: depth chart resize js error
When changing the window width:

image

```js
// makeLabels attempts to create the appropriate labels for the specified
// screen size, context, and label spacing.
function makeLabels (ctx, screenW, min, max, spacingGuess, step, unit) {
  const n = screenW / spacingGuess
  const diff = max - min
  const tickGuess = diff / n
  // make the tick spacing a multiple of the step
  const tick = tickGuess + step - (tickGuess % step)
  let x = min + tick - (min % tick)
  const absMax = Math.max(Math.abs(max), Math.abs(min))
  // The Math.round part is the minimum precision required to see the change in the numbers.
  // The 2 accounts for the precision of the tick.
  const sigFigs = Math.round(Math.log10(absMax / tick)) + 2
  const pts = []
  let widest = 0
  while (x < max) {
    x = Number(x.toPrecision(sigFigs))  <<====here
```

vctt94:
closes #1056

### DIFF
--- a/client/webserver/site/src/js/charts.js
+++ b/client/webserver/site/src/js/charts.js
@@ -83,8 +83,8 @@ export class DepthChart {
     }
 
     bind(this.canvas, 'wheel', e => { this.wheel(e) })
-    this.resize = h => { this.resize_(h) }
-    bind(window, 'resize', () => { this.resize(parent.clientHeight) })
+    this.resize = () => this.resize_(parent.clientHeight)
+    bind(window, 'resize', this.resize)
     bind(this.canvas, 'click', e => { this.click(e) })
     this.resize(parent.clientHeight)
   }


client/webserver/site/src/js/charts.js
```diff
@@ -721,11 +721,19 @@ function makeLabels (ctx, screenW, min, max, spacingGuess, step, unit) {
...
+      lbls: pts
+    }
+  }
   // The Math.round part is the minimum precision required to see the change in the numbers.
   // The 2 accounts for the precision of the tick.
   const sigFigs = Math.round(Math.log10(absMax / tick)) + 2
-  const pts = []
-  let widest = 0
   while (x < max) {
     x = Number(x.toPrecision(sigFigs))
```
buck54321, line 738: This is where the error is seen, I think, when `sigFigs == 0`.

client/webserver/site/src/js/charts.js
```diff
@@ -721,11 +721,19 @@ function makeLabels (ctx, screenW, min, max, spacingGuess, step, unit) {
...
+  // precision for recalculating it.
+  if (absMax < 1) {
+    return {
+      widest: widest,
+      lbls: pts
+    }
+  }
   // The Math.round part is the minimum precision required to see the change in the numbers.
   // The 2 accounts for the precision of the tick.
   const sigFigs = Math.round(Math.log10(absMax / tick)) + 2
```
buck54321, line 736: This is almost certainly where the bug is. The intention is to make sure to show enough sig figs that two consecutive ticks marks never show the same value because of rounding. Not sure how this would ever give `sigFigs` equal to zero, since presumably `absMax > 0` and `tick < absMax`.

buck54321 (changes requested):

buck54321:
Is anybody able to reliably reproduce the error?

chappjc:
Resize the window in Firefox. Happens reliably.  Either resize the whole window or the console width when it's on the right.  I'll make a video.

client/webserver/site/src/js/charts.js
chappjc: Yup, it's here

chappjc:
The key to reproducing seems to be navigating away from markets and back to it, but I also see it sometimes when opening a new tab directly to /markets I think.

Here's a vid where at first I couldn't repro, and then it started after visiting the wallets page:

https://user-images.githubusercontent.com/9373513/125165554-137b7c00-e15d-11eb-964f-cb21fea0319d.mp4

The FF debugger reallly sucks, but in Chromium:

image

The unit is btc/dcr.  screenW is negative, and min and max are tiny.  absMax is tiny, and sigFigs gets to be NaN.

chappjc:
In the caller:
image

chappjc:
When I refresh the page to make it start working correctly:

image

vctt94:
Yeah, you are right and that approach was only hiding the error.

This error happens because of the `screenW` param of `makeLabels (ctx, screenW, min, max, spacingGuess, step, unit)` being negative. 

That was happen because of the min diffence being bigger than the max when resizing. 

Now I am making this difference an absolute value, and with that the error is gone.

chappjc:
> Now I am making this difference an absolute value, and with that the error is gone

I'm afraid that's also not correct yet.  The problem is apparently because when you navigate to the markets page from say wallets by clicking on the Markets menu item, the elements aren't rendered when the `DepthChart` constructor is called, resulting in zeros for `clientWidth` and `clientHeight`.  See the screenshot in https://github.com/decred/dcrdex/pull/1113#issuecomment-877644244.  That has set `min` in our code, but max is zero because it's not rendered.

vctt94:
Yeah, you are correct.

The root of the problem (Now I am pretty sure this is correct!) is: It had an arrow function into the bind function, which generates a new reference, different from `this.resize`. 

So it was never found when unbinding. 

 The bug can still happen though, when resizing for a really small width, which can be fixed if having an absolute value, between max and min, on the `Extents` class. 

If you guys want I can add that, as well. 

image

client/webserver/site/src/js/charts.js
```diff
@@ -83,8 +83,8 @@ export class DepthChart {
     }
 
     bind(this.canvas, 'wheel', e => { this.wheel(e) })
-    this.resize = h => { this.resize_(h) }
-    bind(window, 'resize', () => { this.resize(parent.clientHeight) })
```
chappjc, line 87: I had a feeling it had something to do with these and `parent` changing, but these lines are making my head spin.

chappjc:
I have both this PR and https://github.com/decred/dcrdex/pull/1126 flagged for 0.2.1 cherry-pick, so please have a look at these when you get a chance @buck54321

buck54321 (approved):
That makes more sense. Well, at least you had the pleasure of learning some of the chart code.

client/webserver/site/src/js/charts.js
```diff
@@ -83,8 +83,8 @@ export class DepthChart {
...
+    this.resize = () => this.resize_(parent.clientHeight)
+    bind(window, 'resize', this.resize)
```
buck54321, line 87: Whoops. This actually broke the markets page a little. We can't hard-code the `resize_` argument, since `resize` is called other places. https://github.com/decred/dcrdex/blob/cfa95ebaa34d954f078fef27289c5232ecb0fbae/client/webserver/site/src/js/markets.js#L302

buck54321: I snuck in a fix into #1051

chappjc: Damn I already put it in 0.2.1.  Where's the diff for that fix?

PR #1115, opened 2021-06-28
order form: hide by default and show iff wallets are set.

### LINKED ISSUE #1114: Buy order misses after little while
When starting dexc into browser, it correctly shows the buy/sell form:

image

But after a little while, it hides itself (no action needed)

image

I think it might be happening because I do not have a btc wallet on it

amassarwi:
Closes #1114.

### DIFF
--- a/client/webserver/site/src/js/markets.js
+++ b/client/webserver/site/src/js/markets.js
@@ -514,23 +514,24 @@ export default class MarketsPage extends BasePage {
   /* setMarket sets the currently displayed market. */
   async setMarket (host, base, quote) {
     const dex = app.user.exchanges[host]
+    const { page } = this
     // If we have not yet connected, there is no dex.assets or any other
     // exchange data, so just put up a message and wait for the connection to be
     // established, at which time handleConnNote will refresh and reload.
     if (!dex.connected) {
       this.market = { dex: dex }
-      this.page.chartErrMsg.textContent = 'Connection to dex server failed. ' +
+      page.chartErrMsg.textContent = 'Connection to dex server failed. ' +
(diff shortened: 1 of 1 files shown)


JoeGruffins (approved):
Works great for me. I do not see the form until fee is confirmed and I have both wallets set up.

There are two typos in the original commit message.

amassarwi:
@JoeGruffins thanks for the review, updated the commit message.

btw `iff` was intended - used in some code bases as `if and only if`. `be default` was indeed a typo.

vctt94 (approved):
Tested and works for me

client/webserver/site/src/js/markets.js
```diff
@@ -514,23 +514,24 @@ export default class MarketsPage extends BasePage {
   /* setMarket sets the currently displayed market. */
   async setMarket (host, base, quote) {
     const dex = app.user.exchanges[host]
+    const { page } = this
```
buck54321, line 517: `const page = this.page` is preferred for grabbing a single property.

client/webserver/site/src/js/markets.js
```diff
@@ -1265,10 +1266,8 @@ export default class MarketsPage extends BasePage {
   setBalanceVisibility () {
     if (this.market.dex.connected) {
       Doc.show(this.page.balanceTable)
-      Doc.show(this.page.orderForm)
     } else {
       Doc.hide(this.page.balanceTable)
-      Doc.hide(this.page.orderForm)
     }
```
buck54321, line 1271: Ah. This is the bug, huh.

buck54321 (approved):
One tiny nit. Nice fix.

PR #1117, opened 2021-06-29
add price rate to browser title

### LINKED ISSUE #1053: display price in browser window title
Just a nice to have feature request.  It would be cool to show the price of whatever pair is selected in the browser title like Bittrex and some others do, instead of the current title.  This way I can see it without clicking the tab.
image

vctt94:
closes #1053

### DIFF
--- a/client/webserver/site/src/js/markets.js
+++ b/client/webserver/site/src/js/markets.js
@@ -965,15 +965,14 @@ export default class MarketsPage extends BasePage {
     // gets first price value from buy or from sell, so we can show it on
     // title.
     const midGapValue = this.midGap()
+    if (!midGapValue) return
 
     const market = this.market
     const [b, q] = [market.baseCfg, market.quoteCfg]
     const baseSymb = b.symbol.toUpperCase()
     const quoteSymb = q.symbol.toUpperCase()
-    if (midGapValue) {
-      // more than 6 numbers it gets too big for the title.
-      document.title = `${midGapValue.toFixed(6)} | ${baseSymb}${quoteSymb} | ${this.ogTitle}`
-    }
+    // more than 6 numbers it gets too big for the title.
+    document.title = `${midGapValue.toFixed(6)} | ${baseSymb}${quoteSymb} | ${this.ogTitle}`
   }
 
   /* handleBookRoute is the handler for the 'book' notification, which is sent
--- a/client/webserver/site/src/js/markets.js
+++ b/client/webserver/site/src/js/markets.js
@@ -963,6 +963,16 @@ export default class MarketsPage extends BasePage {
   handleBookRoute (note) {
     app.log('book', 'handleBookRoute:', note)
     const mktBook = note.payload
+    const { book } = mktBook
+    // gets first price value from buy or from sell, so we can show it on
+    // title.
+    const firstKnownValue = book.buys[0] ? book.buys[0] : book.sells[0]
+    if (firstKnownValue) {
+      // remove rate from title if already have it.
+      document.title = (document.title).replace(/\d+([.])?/g, '')
+      // more than 6 numbers it gets too big for the title.
+      document.title = `${firstKnownValue.rate.toFixed(6)} ${document.title}`
+    }
     const market = this.market
     const page = this.page
     const host = market.dex.host
--- a/client/webserver/site/src/js/markets.js
+++ b/client/webserver/site/src/js/markets.js
@@ -91,6 +91,8 @@ export default class MarketsPage extends BasePage {
     this.ordersSortKey = 'stamp'
     // 1 if sorting ascendingly, -1 if sorting descendingly.
     this.ordersSortDirection = 1
+    // store original title so we can re-append it when updating market value.
+    this.ogTitle = document.title
 
     const reporters = {
       click: p => { this.reportClick(p) },
@@ -966,11 +968,11 @@ export default class MarketsPage extends BasePage {
 
     const market = this.market
     const [b, q] = [market.baseCfg, market.quoteCfg]
-    const firstSymbol = b.symbol.toUpperCase()
-    const secondSybol = q.symbol.toUpperCase()
+    const baseSymb = b.symbol.toUpperCase()
+    const quoteSymb = q.symbol.toUpperCase()
     if (midGapValue) {
       // more than 6 numbers it gets too big for the title.
-      document.title = `${midGapValue.toFixed(5)} ${firstSymbol}/${secondSybol}`
+      document.title = `${midGapValue.toFixed(6)} | ${baseSymb}${quoteSymb} | ${this.ogTitle}`
     }
   }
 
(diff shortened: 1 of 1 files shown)


JoeGruffins (commented):
Looks good but doesn't seem to update unless I refresh the page, so if you are viewing a different page it doesn't change.

client/webserver/site/src/js/markets.js
```diff
@@ -959,6 +959,16 @@ export default class MarketsPage extends BasePage {
   handleBookRoute (note) {
     app.log('book', 'handleBookRoute:', note)
     const mktBook = note.payload
+    const { book } = mktBook
+    // gets first price value from buy or from sell, so we can show it on
+    // title.
+    const firstKnownValue = book.buys[0] ? book.buys[0] : book.sells[0]
```
buck54321, line 965: With an empty book

image

client/webserver/site/src/js/markets.js
```diff
@@ -959,6 +959,16 @@ export default class MarketsPage extends BasePage {
   handleBookRoute (note) {
     app.log('book', 'handleBookRoute:', note)
     const mktBook = note.payload
+    const { book } = mktBook
```
buck54321, line 962: Use `const book = mktBook.book` for grabbing just one property.

client/webserver/site/src/js/markets.js
buck54321, line 965: You can update the title after the call to `this.handleBook` below and just use `this.midGap()` for the value.

client/webserver/site/src/js/markets.js
```diff
@@ -959,6 +959,16 @@ export default class MarketsPage extends BasePage {
...
+    // gets first price value from buy or from sell, so we can show it on
+    // title.
+    const firstKnownValue = book.buys[0] ? book.buys[0] : book.sells[0]
+    if (firstKnownValue) {
+      // remove rate from title if already have it.
+      document.title = (document.title).replace(/\d+([.])?/g, '')
+      // more than 6 numbers it gets too big for the title.
+      document.title = `${firstKnownValue.rate.toFixed(6)} ${document.title}`
```
buck54321, line 970: I'd like to see the market name. Something like `0.00512 DCR/BTC`.

I don't know that we need to save the original title, which said `Markets | Decred DEX`, but if you do want to append the original text, grab a copy in the `MarketPage` constructor, `this.ogTitle = document.title` and just append it rather than using `replace`.

client/webserver/site/src/js/markets.js
```diff
@@ -959,6 +959,16 @@ export default class MarketsPage extends BasePage {
   handleBookRoute (note) {
```
buck54321, line 959: Doing this in `handleBookRoute` will set the price when the book is originally received, but will never update the value. I'm guessing we want the value to update when the book changes.

buck54321 (changes requested):

client/webserver/site/src/js/markets.js
vctt94: I removed the replace and now I am adding the selected assets into the title. 

Added a new method for updating it, and now I call it on `handleBookRoute`, `handleUnbookOrderRoute`, and `handleBookOrderRoute`.

This way if we adding or removing new bids, it will update the title based on it.

JoeGruffins (approved):
Looks pretty great to me.

chappjc (changes requested):

client/webserver/site/src/js/markets.js
```diff
@@ -952,6 +952,24 @@ export default class MarketsPage extends BasePage {
...
+    // title.
+    const midGapValue = this.midGap()
+
+    const market = this.market
+    const [b, q] = [market.baseCfg, market.quoteCfg]
+    const firstSymbol = b.symbol.toUpperCase()
+    const secondSybol = q.symbol.toUpperCase()
+    if (midGapValue) {
+      // more than 6 numbers it gets too big for the title.
+      document.title = `${midGapValue.toFixed(5)} ${firstSymbol}/${secondSybol}`
```
chappjc, line 969: This is gonna say something like 0.004 DCR/BTC.  The rate is inversed.

client/webserver/site/src/js/markets.js
chappjc: > I'd like to see the market name. Something like `0.00512 DCR/BTC`.

Despite calling it the market name, is giving units to that rate, so it'll need to be BTC/DCR

client/webserver/site/src/js/markets.js
buck54321: Maybe use a dash instead of a slash. When naming a market, the base symbol goes first by convention.

chappjc: I will note that Bittrex bucks this trend

image

probably because of how confusing it is to have it written the other way.

Still, it's true that the base asset is usually first.

client/webserver/site/src/js/markets.js
chappjc: Discussion in this thread: https://github.com/decred/dcrdex/pull/1117#discussion_r671288493
Looks like use "DCR-BTC" and don't have the number with units.

client/webserver/site/src/js/markets.js
buck54321: Somewhere in the Bittrex API docs, I remember reading an acknowledgement of this mistake.

chappjc: OK fair enough.  I see the url as https://bittrex.com/Market/Index?MarketName=BTC-DCR, but I don't really strive to emulate Bittrex either way.

This PR is working well for me.  However, here's what I see presently:

image
image


And here's what I'd like to see:

image
image

OR even without a dash, but it's not a big deal:

image

vctt94: I changed it and now I am showing as your last option, without dash.

client/webserver/site/src/js/markets.js
```diff
@@ -956,6 +958,24 @@ export default class MarketsPage extends BasePage {
...
   }
 
+  /* updateTitle update the browser title based on the midgap value and the
+   * selected assets.
+   */
+  updateTitle () {
+    // gets first price value from buy or from sell, so we can show it on
+    // title.
+    const midGapValue = this.midGap()
+
```
buck54321, line 968: `if (!midGapValue) return`

client/webserver/site/src/js/markets.js
```diff
@@ -956,6 +958,24 @@ export default class MarketsPage extends BasePage {
...
+    const market = this.market
+    const [b, q] = [market.baseCfg, market.quoteCfg]
+    const baseSymb = b.symbol.toUpperCase()
+    const quoteSymb = q.symbol.toUpperCase()
+    if (midGapValue) {
+      // more than 6 numbers it gets too big for the title.
+      document.title = `${midGapValue.toFixed(6)} | ${baseSymb}${quoteSymb} | ${this.ogTitle}`
+    }
```
buck54321, line 976: Cleaner to `return` above.

buck54321 (approved):
Can clean up a tiny bit, but looks great.

chappjc (approved):

PR #1119, opened 2021-07-02
multi: add ability to persist password on UI login 

### LINKED ISSUE #946: multi: allow persistent encryption key
When the user logs in, they could be presented with a checkbox similar to a "remember me" option that will cause the app password to be effectively cached for the duration of their session. That said, the app password is all powerful, so I don't think we should cache the actual password or even the decoded `Crypter`. Instead, I think we should generate some kind of "session key" that enables generating the password for the duration of the user's session, but does not itself encode the encryption key.

```go
type SessionAuth interface {
	PW() ([]byte, error)
}

func (c *Core) Session(pw []byte) (SessionAuth, error)
```

The  `SessionAuth` itself would encrypt the password with some kind of temporary key and only reveal it when prompted. This approach has the benefit that `Core` is still in control of destroying the session key since `Core` controls the implementation of the concrete type.

When the user selects the "remember my password" checkbox, the `WebServer` knows to generate and cache a `core.SessionAuth`. Then, whenever `WebServer` has a password, they can do something like 

```go
if len(form.PW) == 0 && s.sessionAuth != nil {
	form.PW, err := s.sesssionAuth.PW()
	// handle error
}
```

The front end would need to know somehow whether there is an active "session" so we can hide the password fields. An empty or `null` password also takes on a special significance with this scheme, signaling from to the **webserver** to check for a `SessionAuth`.

On logout, the `WebServer` **should** forget the cached session, but `Core` should also disable the `SessionAuth` somehow so it doesn't really matter.

Anyway, this is just one suggestion, and there are probably better ways to do it. I've been revisiting this for over and over though, and haven't come up with anything super clean.

martonp:
This PR adds a checkbox to the login page that gives the user the option to persist their password for the session. When selected, all the password fields, other than disable account, will no longer be displayed.

When the checkbox is selected, the password is encrypted with a random key, and the encrypted password is stored in a cache in core. The key to decrypt the password is put into a cookie. Whenever a request that requires a password is made with an empty password field, the webserver uses the key from the cookie to retrieve the password from core.

Closes #946.


client/webserver/api.go
```diff
@@ -242,6 +275,14 @@ func (s *WebServer) apiAccountImport(w http.ResponseWriter, r *http.Request) {
...
 	}
 	r.Close = true
+	if len(form.Pass) == 0 {
+		cachedPass, err := s.getCachedPassword(r)
+		if err != nil {
+			log.Errorf("unable to get cached pw: %v", err)
+		} else {
+			form.Pass = cachedPass
+		}
+	}
```
buck54321, line 285: This block is repeated enough to justify a helper method.

```Go
func (s *WebServer) resolvePass(appPW []byte) ([]byte, error) {
	if len(appPW) > 0 {
		return appPW, nil
	}
	cachedPass, err := s.getCachedPassword(r)
	if err != nil {
		if errors.Is(err, ErrNoCachedPW) {
			return nil, fmt.Errorf("app pass cannot be empty")
		}
		return nil, fmt.Errorf("error retrieving cached pw: %v", err)
	}
	return cachedPass, nil
}
```

where `ErrNoCachedPW` will be some constant error that identifies that there is no password cached.

client/webserver/api.go
```diff
@@ -242,6 +275,14 @@ func (s *WebServer) apiAccountImport(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	r.Close = true
+	if len(form.Pass) == 0 {
+		cachedPass, err := s.getCachedPassword(r)
+		if err != nil {
+			log.Errorf("unable to get cached pw: %v", err)
```
buck54321, line 281: Error message would be nonsensical if no password has been cached and user is not utilizing the feature.

client/webserver/api.go
```diff
@@ -362,6 +411,14 @@ func (s *WebServer) apiLogout(w http.ResponseWriter, r *http.Request) {
 		SameSite: http.SameSiteStrictMode,
 	})
 
+	http.SetCookie(w, &http.Cookie{
+		Name:     pwKeyCK,
+		Path:     "/",
+		Value:    "",
+		Expires:  time.Unix(0, 0),
+		SameSite: http.SameSiteStrictMode,
+	})
```
buck54321, line 420: This is repeated enough to warrant a helper function, `func clearPassCacheCookie(w http.ResponseWriter)`.

client/webserver/api.go
```diff
@@ -516,13 +573,50 @@ func (s *WebServer) apiChangeAppPass(w http.ResponseWriter, r *http.Request) {
...
+				SameSite: http.SameSiteStrictMode,
+				Expires:  time.Unix(0, 0),
+			})
+		} else {
+			http.SetCookie(w, &http.Cookie{
+				Name:     pwKeyCK,
+				Value:    hex.EncodeToString(key),
+				Path:     "/",
+				SameSite: http.SameSiteStrictMode,
+			})
```
buck54321, line 616: How about a helper function `func setPassCacheCookie(w http.ResponseWriter, v []byte)`.

client/webserver/site/src/html/forms.tmpl
```diff
@@ -62,14 +62,15 @@
...
+  {{$passwordIsCached := .UserInfo.PasswordIsCached}}
+  <div {{if $passwordIsCached}}class="d-hide"{{end}}>
     {{if .}}
     <label for="nwAppPass" class="pt-3 pl-1 mb-0">App Password</label>
     <div class="fs14 px-1 mb-1">Your app password is always required when performing sensitive wallet operations.</div>
     {{else}}
     <label for="nwAppPass" class="pt-3 pl-1 mb-1">App Password</label>
     {{end}}
-    <input type="password" class="form-control select" id="nwAppPass" autocomplete="off">
+    <input type="password" class="form-control select" id="nwAppPass" autocomplete="off" {{if $passwordIsCached}}hidden{{end}}>
```
buck54321, line 73: Is this necessary? You're hiding the containing `<div>`.

client/webserver/site/src/html/markets.tmpl
```diff
@@ -423,14 +423,15 @@
...
-          <div class="col-12 p-0 text-right">
-            <div>&nbsp;</div>
+        {{$passwordIsCached := .UserInfo.PasswordIsCached}}
+        <div class="fs16 text-center {{if $passwordIsCached}}d-hide{{end}}">Authorize this order with your app password.</div>
+        <div class="d-flex mt-3 {{if $passwordIsCached}}justify-content-center{{end}}" >
+            <div class="col-12 p-0 text-left {{if $passwordIsCached}}d-hide{{end}}">
+              <label for="vPass" class="pl-1 mb-1">Password</label>
+              <input type="password" class="form-control select" id="vPass">
+            </div>
+          <div class="col-12 p-0 {{if not $passwordIsCached}}text-right{{end}}">
```
buck54321, line 433: Indentation is not consistent. Might need to change some settings in your IDE or something so this doesn't happen.

client/webserver/site/src/html/forms.tmpl
```diff
@@ -272,12 +275,12 @@
   The remaining amount may change before the cancel order is matched.
 </div>
 <hr class="dashed mt-2">
-<div class="d-flex flex-row align-items-end pb-4 px-3">
-  <div class="col-12 p-0">
+<div class="d-flex flex-row {{if $passwordIsCached}}justify-content-center{{else}}align-items-end{{end}} pb-4 px-3">
+  <div class="col-12 p-0 {{if $passwordIsCached}}d-hide{{end}}">
```
buck54321, line 279: Please don't center align the submit button. Keep it on the right side. This would be consistent with other forms as well.

Use the `visibility` css attribute via bootstrap's `invisible` class so you don't need to set the justification on the parent element.

`<div class="col-12 p-0 {{if $passwordIsCached}}invisible{{end}}">`

client/webserver/site/src/html/wallets.tmpl
```diff
@@ -88,6 +89,8 @@
   {{- /* RIGHT SIDE */ -}}
   <div id="rightBox" class="col-10 px-2 py-5 d-flex justify-content-center align-items-start overflow-auto">
 
+    <div class="fs14 pt-3 text-center d-hide errcolor" id="rightBoxErr"></div>
```
buck54321, line 92: The problem with having a single error message box outside of the forms is that this will need to be cleared for any new action the user does anywhere on the wallets page, otherwise the error message will persist forever. My read is that you need to do this because unlocking a wallet with a cached password no longer requires display of any forms. I'm thinking the best solution then is to just display the unlock form but hide the app password field, showing only the error message.

client/core/core.go
```diff
@@ -1255,6 +1260,58 @@ func (c *Core) wallet(assetID uint32) (*xcWallet, bool) {
...
+		return nil, fmt.Errorf("error decrypting password: %v", err)
+	}
+
+	return pw, nil
+}
+
+// CacheAppPassword encrypts the app password with a random key, caches it,
+// and returns the key. The mapping string argument is used to lookup the
+// cached password using the GetCachedPassword function.
+func (c *Core) CacheAppPassword(appPW []byte, mapping string) ([]byte, error) {
```
buck54321, line 1296: Just a couple of notes on this approach. 

1. Even though I had originally suggested that `Core` be involved with the generation of these keys, based on our discussion since, I'm now thinking that we can just handle the entire thing in **webserver**. Is there any reason `Core` needs to be involved?
1. If `mapping` is a unique string associated with the generated key, then `mapping` should be generated by `Core`, perhaps stored in the key blob.

client/webserver/site/src/js/settings.js
```diff
@@ -109,7 +109,11 @@ export default class SettingsPage extends BasePage {
     const page = this.page
     page.exportAccountHost.textContent = host
     page.exportAccountErr.textContent = ''
-    this.showForm(authorizeAccountExportForm)
+    if (authorizeAccountExportForm.hidden) {
```
buck54321, line 112: If this is the reason you're setting the `hidden` attribute on hidden form elements, you can use `Doc.isHidden(authorizeAccountExportForm.parentElement)` instead. 

Or even better, write a function/method to tell whether there is a cached password. Perhaps a method on `State` that uses `State.getCookie` under the hood.

client/webserver/webserver.go
```diff
@@ -397,11 +400,58 @@ func (s *WebServer) isAuthed(r *http.Request) bool {
...
+		return nil, err
+	}
+}
+
+// isAuthed checks if the incoming request is from an authorized user/device.
+// Requires the auth token cookie to be set in the request and for the token
+// to match `WebServer.validAuthToken`.
+func (s *WebServer) isAuthed(r *http.Request) bool {
+	authToken := getAuthToken(r)
 	s.mtx.RLock()
```
buck54321, line 427:
```Go
if authToken == "" {
    return false
}
```

client/core/core.go
```diff
@@ -1255,6 +1260,58 @@ func (c *Core) wallet(assetID uint32) (*xcWallet, bool) {
...
 
+// GetCachedPassword takes the mapping string passed to CacheAppPassword and
+// the key returned from it to lookup and decrypt the app password.
+func (c *Core) GetCachedPassword(mapping string, key []byte) ([]byte, error) {
+	_, pushes, err := encode.DecodeBlob(key)
+	if err != nil {
+		return nil, fmt.Errorf("error decoding blob: %v", err)
+	}
+
+	key, deserializedCrypter := pushes[0], pushes[1]
```
buck54321, line 1271: The name `key` is used to represent two different things here. Maybe the method argument can be named `keyBlob` or something.

buck54321 (changes requested):
First pass. Looks good conceptually, though I've suggested some architectural changes.

client/webserver/site/src/js/settings.js
martonp: The reason I like adding the hidden attribute is that there is not only a check to make sure that the cookie contains the key, but there is also a check (in `authMiddleware`) to make sure that the key can unlock a password.

client/core/core.go
martonp: `mapping` is  the `authToken` that's also in the cookies. I think you're right, this can just go in `webserver`.

client/webserver/site/src/html/forms.tmpl
martonp: Screen Shot 2021-07-15 at 9 38 55 AM
Screen Shot 2021-07-15 at 9 39 33 AM

`invisible` gives the large gap you see in the second picture. With `d-hide` and `justify-content-end` it looks like the first one.

client/webserver/site/src/html/register.tmpl
```diff
@@ -26,12 +26,12 @@
 
     {{- /* Set up the Decred wallet. Only shown if user has not connected their Decred wallet. */ -}}
     <form class="card mx-auto my-5 bg1{{if not .WalletStep}} d-hide{{end}}" id="newWalletForm">
-      {{template "newWalletForm" "Your Decred wallet is required to pay registration fees."}}
+      {{template "newWalletForm" .}}
```
martonp, line 29: I didn't notice this before, but rendering these forms when called from the register page caused an error because they could not find `.UserInfo.IsPasswordCached`.`

client/webserver/webserver_test.go
```diff
@@ -396,7 +403,9 @@ func TestAPIWithdraw(t *testing.T) {
 		return resp.OK
 	}
 
-	body = &withdrawForm{}
+	body = &withdrawForm{
```
martonp, line 406: This is needed because `resolvePass` will return an error if this is an empty string.

client/webserver/site/src/js/settings.js
martonp: I ended up going with a new method on `State`. This is cleaner, and there shouldn't be an invalid key in the cookies unless there is a bug elsewhere.

martonp:
@buck54321 This is ready for another review.

client/webserver/site/src/html/forms.tmpl
```diff
@@ -79,6 +81,8 @@
 {{end}}
 
 {{define "unlockWalletForm"}}
+{{$isRegisterPage := eq .Title "Register | Decred DEX"}}
+{{$passwordIsCached := .UserInfo.PasswordIsCached}}
```
buck54321, line 85: We really want to avoid string matching here. How about a two-argument `templateFuncs` function?

#### Or even better...

The wallet unlock form is one of the last forms in **forms.tmpl** that doesn't have a javascript class. It is still bound with a function (which is how they all were at one point) called `bindOpenWallet`. Maybe you can create the class (see others in **forms.js**) and create a method to hide the PW input and submit button, and then just hide them in code. We can do that, because this form is never displayed by default on a page, so the user will never see the elements flash on screen before being hidden. A major benefit of doing it that way is that if the user caches their password using a different form on the same page, we'll have the machinery in place to hide these fields without a page reload.

client/webserver/site/src/html/login.tmpl
```diff
@@ -8,6 +8,8 @@
         <div class="pb-2">
           <label for="pw" class="pl-1 mb-1">Password</label>
           <input type="password" class="form-control select" id="pw" autocomplete="current-password">
+          <label for="rememberPass" class="pl-1 mb-1">Remember my password</label>
+          <input type="checkbox" id="rememberPass">
```
buck54321, line 12: We'll probably want something similar on the register page too, no? Otherwise the user would have to log out and log back in to use the feature?

client/webserver/webserver.go
```diff
@@ -398,11 +410,105 @@ func (s *WebServer) isAuthed(r *http.Request) bool {
...
 	return s.authTokens[authToken]
 }
 
+// getCachedPassword takes authToken passed to and the keyBlob returned from
+// cacheAppPassword to retrieve and decrypt the app password.
+func (s *WebServer) getCachedPassword(keyBlob []byte, authToken string) ([]byte, error) {
+	_, pushes, err := encode.DecodeBlob(keyBlob)
+	if err != nil {
+		return nil, fmt.Errorf("error decoding blob: %v", err)
+	}
```
buck54321, line 451: `if len(pushes) != 2 { ...` for good measure.

client/webserver/webserver_test.go
```diff
@@ -610,3 +620,63 @@ func TestGetOrderIDCtx(t *testing.T) {
...
+		Pass: encode.PassBytes(""),
+	}
+	want := `{"ok":false,"msg":"password error: app pass cannot be empty"}`
+	tCore.notHas = true
+	ensureResponse(t, s.apiNewWallet, want, reader, writer, body, nil)
+
+	want = `{"ok":true}`
+	ensureResponse(t, s.apiNewWallet, want, reader, writer, body, map[string]string{
+		authCK:  authToken1,
+		pwKeyCK: hex.EncodeToString(key1),
```
buck54321, line 674: We should store the serialized `Crypter` locally and only send the key in the cookie. How about we update the `authTokens map[string]bool` field to be something like

```Go
auths map[string]*struct{
    CachedCrypter encrypt.Crypter
}
```

then if they provide a key we can regenerate, but nobody ever has all the pieces needed in memory.

buck54321 (changes requested):
Partial review. Lemme know what you think about these notes. I promise to review faster now.

client/webserver/webserver_test.go
martonp: That makes sense, then the cookie can be smaller and we don't have to regenerate the crypter each time.
How about having all the password related things in a cachedPasswords map like this:
```go
cachedPasswords map[string]*struct{
    EncryptedPass []byte
    Crypter encrypt.Crypter 
}

File: client/webserver/webserver.go (hunk 17 above)
Comment (martonp) [reply client/webserver/webserver.go:451]: Can't hurt.

File: client/webserver/site/src/html/forms.tmpl [hunk 19]
```diff
@@ -33,6 +33,7 @@
 {{end}}
 
 {{define "newWalletForm"}}
+{{$isRegisterPage := eq .Title "Register | Decred DEX"}}
```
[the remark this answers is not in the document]
martonp: I'm still using `.UserInfo.PasswordIsCached` below. I should have moved it to the top.

client/webserver/site/src/html/forms.tmpl
martonp: Yeah that's true.. it's better not to have to reload the screen. I'll just try and do the whole thing without using the template arguments. Then, if the the client ever gets a key from the UI that cannot decode a password, the cookie will be cleared.

client/webserver/webserver_test.go
martonp: If you have the `Crypter` then you don't need the key anymore. I ended up caching the serialized crypter instead.

martonp:
@buck54321 It's ready for another look.

In the registration workflow (where the page is not reloaded between forms) `State.passwordIsCached` is being used. I kept using the template arguments on the other pages though since it makes the code much more simple. There is always a page reload when going from the login/registration pages to every other page.

chappjc:
@martonp Please rebase and resolve when you get a chance.

chappjc:
@martonp Please try to avoid merging from master to get the branch up-to-date.  Instead, rebase on master.

It is tricky to get rid of the merge commit when you synchronized with master @ 16da72f.  I'm squashing this branch down to a single commit with 16da72f as it's base like:

```
git checkout rememberPW
git reset --hard marton/rememberPW
git merge --squash HEAD@{1}
git commit -m "Add ability to remember password at login."
```

Then it can be rebased on master a bit more easily since it's just one commit now (`git rebase master`).

Although I see that the conflicts are gonna be tricky to resolve, particularly in forms.js, so thank you.

client/webserver/site/src/html/login.tmpl
buck54321, line 12: To be consistent with other forms, put the input before the label, and add `class="form-check-input"` to the checkbox. I think you'll need to wrap everything in a `class="pl-4"` div too, but not certain.

Also looks like you need some padding between the password field and the remember-password input.

image

buck54321 (commented):
Looking really good. I want to give it one more look, but it's testing well. Just a couple of little things for now.

client/webserver/api.go
```diff
@@ -567,6 +602,24 @@ func (s *WebServer) apiChangeAppPass(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Force other sessions to login again. Without this, any sessions that
+	// had a cached password will no longer work. However, we assign a new auth
+	// token and cache the new password (if it was previously cached) for this
+	// session.
+	passwordIsCached := s.isPasswordCached(r)
+	s.deauth()
```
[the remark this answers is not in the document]
martonp: If I do that then I won't know if the password was cached before I called deauth. I can put the comment between the two lines.

client/webserver/site/src/html/settings.tmpl
```diff
@@ -38,7 +39,7 @@
         {{end}}
       </div>
       <br>
-      <div{{if not .UserInfo.Authed}} class="d-hide"{{end}}>
+      <div {{if not .UserInfo.Authed}}class="d-hide"{{end}}>
```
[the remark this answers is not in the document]
martonp: This is a mistake.. it was red in my IDE the previous way, and I thought they were equivalent.

client/webserver/site/src/html/login.tmpl
martonp: Looks way better now.

client/webserver/webserver.go
```diff
@@ -402,11 +422,102 @@ func (s *WebServer) isAuthed(r *http.Request) bool {
...
+	cachedPassword, ok := s.cachedPasswords[authToken]
+	s.mtx.Unlock()
+	if !ok {
+		return nil, fmt.Errorf("cached encrypted password not found for"+
+			" auth token: %v", authToken)
+	}
+
+	crypter, err := encrypt.Deserialize(key, cachedPassword.SerializedCrypter)
+	if err != nil {
+		return nil, fmt.Errorf("error deserializing crypter: %v", err)
```
buck54321, line 470: Might as well `%w`.

client/webserver/webserver.go
```diff
@@ -402,11 +422,102 @@ func (s *WebServer) isAuthed(r *http.Request) bool {
...
+	}
+
+	pw, err := crypter.Decrypt(cachedPassword.EncryptedPass)
+	if err != nil {
+		return nil, fmt.Errorf("error decrypting password: %v", err)
```
buck54321, line 475: Might as well `%w`.

client/webserver/webserver.go
```diff
@@ -102,19 +112,28 @@ type clientCore interface {
...
-	authTokens map[string]bool
+	wsServer        *websocket.Server
+	mux             *chi.Mux
+	core            clientCore
+	addr            string
+	csp             string
+	srv             *http.Server
+	html            *templates
+	indent          bool
+	mtx             sync.RWMutex
```
buck54321, line 134: We should really rename this to `authMtx` or something, shouldn't we?

client/webserver/api.go
```diff
@@ -610,7 +667,12 @@ func (s *WebServer) apiWithdraw(w http.ResponseWriter, r *http.Request) {
 		s.writeAPIError(w, fmt.Errorf("no wallet found for %s", unbip(form.AssetID)))
 		return
 	}
-	coin, err := s.core.Withdraw(form.Pass, form.AssetID, form.Value, form.Address)
+	pass, err := s.resolvePass(form.Pass, r)
+	if err != nil {
+		s.writeAPIError(w, fmt.Errorf("password error: %v", err))
+		return
+	}
+	coin, err := s.core.Withdraw(pass, form.AssetID, form.Value, form.Address)
```
buck54321, line 675: We should always require the password to be entered manually for withdraws.

client/webserver/site/src/html/wallets.tmpl
```diff
@@ -152,14 +153,14 @@
         </div>
       </div>
       <hr class="dashed my-4 mx-4">
-      <div class="fs16 px-4 text-center">Authorize the withdraw with your app password.</div>
-      <div class="d-flex px-4 mt-3">
-        <div class="col-12 p-0">
+      <div class="fs16 px-4 text-center {{if $passwordIsCached}}d-hide{{end}}">Authorize the withdraw with your app password.</div>
+      <div class="d-flex px-4 mt-3 {{if $passwordIsCached}}justify-content-end{{end}}">
+        <div class="col-12 p-0 {{if $passwordIsCached}}d-hide{{end}}">
```
buck54321, line 158: Should always require password for withdraws.

buck54321 (changes requested):

martonp:
@buck54321 I agree with all the comments, and I've updated them all.

buck54321 (approved):
Please increment the cache busters in **bodybuilder.js**, and this is good with me.

client/webserver/webserver.go
```diff
@@ -102,19 +112,28 @@ type clientCore interface {
...
+	mux             *chi.Mux
+	core            clientCore
+	addr            string
+	csp             string
+	srv             *http.Server
+	html            *templates
+	indent          bool
+	authMtx         sync.RWMutex
+	authTokens      map[string]bool
+	cachedPasswords map[string]*cachedPassword
```
chappjc, line 136: The keys for these maps are the same, right (both auth tokens)?  Don't worry about refactoring at this stage, but it suggests a single map would be adequate if the logic around access to the map were updated to reflect new semantics (`found` would indicate known auth token regardless of value stored including nil, while non-nil value would indicate a cached password).  Rather than change anything, I just request a simple trailing comment on this delc line to doco the map key

```go
cachedPasswords map[string]*cachedPassword // cached passwords keyed by auth token
```

chappjc (changes requested):

client/webserver/webserver.go
martonp: Yeah that makes sense, I'll do another PR for this.

chappjc (approved):

PR #1120, opened 2021-07-02
fidelity bonds (proof of concept)

chappjc:
### Summary

**Status**: Working. Tests need fixing.  Some frontend issues to address.

See [TODOs](#todos).

Replaces PR https://github.com/decred/dcrdex/pull/1017, tossing out 2,306 loc for a better solution.

This PR replaces the registration fee, which was paid to an address provided by the server operator, with a **time-locked fidelity bond, which is redeemable by the user who posted the bond after a certain time**. This creates a time cost to use DCRDEX instead of a monetary cost.

### Bond Transaction Structure

There must be at least two outputs: the bond output (P2SH) and an account commitment (nulldata).

See:
- [`dex/networks/dcr.ExtractBondDetails`](https://github.com/decred/dcrdex/blob/6358bea6e101f356a7c4fd38f88639b6a7a743f3/dex/networks/dcr/script.go#L439)
- [`client/asset/dcr.(*ExchangeWallet).{MakeBondTx,RefundBond}`](https://github.com/decred/dcrdex/blob/6358bea6e101f356a7c4fd38f88639b6a7a743f3/client/asset/dcr/dcr.go#L2212)
- [`server/assset/dcr.ParseBondTx`](https://github.com/decred/dcrdex/blob/6358bea6e101f356a7c4fd38f88639b6a7a743f3/server/asset/dcr/dcr.go#L101)

#### Time-locked Bond (output 0)

Output 0 of the fidelity bond transaction must be a P2SH output paying to the bond script.

The redeem script looks similar to the refund path of an atomic swap script:

`<locktime[4]> OP_CHECKLOCKTIMEVERIFY OP_DROP <pubkey[33]> OP_CHECKSIG`

The above uses P2PK, and we can do P2PKH instead, but since part of validation requires the pubkey we probably don't gain anything from this:

`<locktime[4]> OP_CHECKLOCKTIMEVERIFY OP_DROP OP_DUP OP_HASH160 <pubkeyhash[20]> OP_EQUALVERIFY OP_CHECKSIG`

I've tested both formats, and I will probably switch to the non-hashed version because validation becomes even more asset-specific given hash160 is not a standard (e.g. ripemd160 of either sha256 for Bitcoin or blake256 for Decred).

Server validates that this P2SH output actually pays to the provided redeem script.

The pubkey referenced by the script need not be controlled by the user's wallet, but in this implementation it comes from their wallet. The client could even use their account pubkey in the script, but for security reasons it is best not to reuse pubkeys in this bond outputs, which are unspent on the blockchain for long periods.

The `lockTime` must be after a bond expiry time, the `lockTimeThreshold`. A bond becomes expired for a DEX account when duration until `lockTime` is less than a `bondExpiry` duration, which is on the order of a month. This, a user should create a bond with a `lockTime` at least `bondExpiry` in the future, but a practical `lockTime` would be more like 2x `bondExpiry` in the future so the account is active on the DEX for `bondExpiry`.

User must demonstrate ownership of the locked funds: sign something with private key corresponding to the pubkey in the TL redeemscript.

#### DEX Account commitment (output 1)

The account commitment OP_RETURN output should reference the account ID that corresponds to the account pubkey in the `postbond` request.

Having it in the raw like this allows the txn alone to identify the account, and without requiring the bond output pay to the account's private keys.

### Protocol

```go
type PostBond struct {
	Signature

}
```

Server must use provided account pubkey to verify signed `postbond` request, otherwise the user could be spamming or putting up a bond for a bogus account or someone else's account.

The response payload:

```go
type PostBondResult struct {
	Amt       uint64 `json:"amt"`
}
```

  - If it is not found, a `PostBondResult` is sent immediately with `Confs=-1` set.  The bond is NOT stored in the DB. This was a courtesy pre-validation. The user should then broadcast the transaction, wait for the required number of confirmations, and send `postbond` again.
  - Otherwise, a response is sent with the observed confirmation count, and a coin waiter is started to watch for the txn to reach the required number of confirmations. The bond is stored in DB as pending. When the transaction reaches the required confirmations, it is flagged active in the DB and a `bondconfirmed` notification is sent.

### TODOs

Auto add bond in `client.Core`.  There are comments in the code regarding an option to automatically postbond when previous bonds are about to expire on DEX.  There are questions pertaining to timing, UI, and UX.  Presently this PR requires using dexcctl to call the `addbond` RPC (or a direct http API call to the endpoint of the same name, which is created for said UI plans).

Bond amount.  It's pretty clear the bond amount has to be at least an order of magnitude more than the previous registration fee since you get it back, and it's only an opportunity cost having it locked up.  I'm thinking like 5 DCR per tier (the bond increment).

### Open Questions

What problems would be solved and/or create by using `OP_CHECKSEQUENCEVERIFY` (relative time lock) instead of `OP_CHECKLOCKTIMEVERIFY`?

When and if accounts go away, would output 1 commit to a magic DEX network key instead of account ID?

Is this transaction structure and account commitment output sufficient to prevent txns from being dual purposed for other services with similar TL output structure?

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -1721,24 +1721,35 @@ func (btc *ExchangeWallet) RefundAddress(contract dex.Bytes) (string, error) {
 	return sender.String(), nil
 }
 
-// LocktimeExpired returns true if the specified contract's locktime has
+// LockTimeExpired returns true if the specified locktime has expired, making it
+// possible to redeem the locked coins.
+func (btc *ExchangeWallet) LockTimeExpired(lockTime time.Time) (bool, error) {
+	// lockTime := time.Unix(int64(lockTimeUnux), 0).UTC()
+	bestBlockHash, err := btc.node.GetBestBlockHash()
+	if err != nil {
+		return false, fmt.Errorf("get best block hash error: %w", err)
+	}
+	bestBlockHeader, err := btc.getBlockHeader(bestBlockHash.String())
+	if err != nil {
+		return false, fmt.Errorf("get best block header error: %w", err)
+	}
+	bestBlockMedianTime := time.Unix(bestBlockHeader.MedianTime, 0).UTC()
+	return bestBlockMedianTime.After(lockTime), nil
+}
+
+// ContractLockTimeExpired returns true if the specified contract's locktime has
 // expired, making it possible to issue a Refund.
-func (btc *ExchangeWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time, error) {
+func (btc *ExchangeWallet) ContractLockTimeExpired(contract dex.Bytes) (bool, time.Time, error) {
 	_, _, locktime, _, err := dexbtc.ExtractSwapDetails(contract, btc.segwit, btc.chainParams)
 	if err != nil {
 		return false, time.Time{}, fmt.Errorf("error extracting contract locktime: %w", err)
 	}
 	contractExpiry := time.Unix(int64(locktime), 0).UTC()
-	bestBlockHash, err := btc.node.GetBestBlockHash()
+	expired, err := btc.LockTimeExpired(contractExpiry)
 	if err != nil {
-		return false, time.Time{}, fmt.Errorf("get best block hash error: %w", err)
+		return false, time.Time{}, err
 	}
-	bestBlockHeader, err := btc.getBlockHeader(bestBlockHash.String())
-	if err != nil {
-		return false, time.Time{}, fmt.Errorf("get best block header error: %w", err)
-	}
-	bestBlockMedianTime := time.Unix(bestBlockHeader.MedianTime, 0).UTC()
-	return bestBlockMedianTime.After(contractExpiry), contractExpiry, nil
+	return expired, contractExpiry, nil
 }
 
 // FindRedemption watches for the input that spends the specified contract
@@ -2207,6 +2218,26 @@ func (btc *ExchangeWallet) PayFee(address string, regFee uint64) (asset.Coin, er
 		return nil, err
 	}
 	return newOutput(txHash, vout, sent), nil
+}
+
+func (btc *ExchangeWallet) MakeBondTx(amt uint64, lockTime time.Time, acctID []byte) (*asset.Bond, error) {
+	return nil, fmt.Errorf("not implemented")
+}
+
+func (btc *ExchangeWallet) RefundBond(coinID, script []byte, privKey []byte) ([]byte, error) {
+	return nil, fmt.Errorf("not implemented")
+}
+
+func (btc *ExchangeWallet) SendTransaction(rawTx []byte) ([]byte, error) {
+	msgTx, err := msgTxFromBytes(rawTx)
+	if err != nil {
+		return nil, err
+	}
+	txHash, err := btc.node.SendRawTransaction(msgTx, false)
+	if err != nil {
+		return nil, err
+	}
+	return toCoinID(txHash, 0), nil
 }
 
 // Withdraw withdraws funds to the specified address. Fees are subtracted from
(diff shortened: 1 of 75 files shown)


davecgh:
> The above uses P2PK, and we can do P2PKH instead, but since part of validation requires the pubkey we probably don't gain anything from this:

There are two primary benefits of using P2PKH over P2PK for outputs:

- The public key is not revealed until the output is redeemed
- The utxoset is smaller when the size of the hash + the additional 2 opcodes is smaller than the size of a public key

Given that, I would argue that it is generally strictly worse to put a P2PKH in a P2SH because:
- The redeem script is generally not revealed until the output is redeemed, which means the public key isn't either
- The output size is fixed by the P2SH output, so what it's wrapping has no bearing on the size of the utxoset
- The redemption of the wrapped P2PKH ultimately takes up more space on the chain, given it needs the additional hash and opcodes

One case where wrapped P2PKH can make sense is if the redeem script must be revealed publicly prior to redemption, but the public key(s) it refers to are not needed until redemption time.  However, in this particular case, as you noted, the public key is required for the DEX instance(s) to validate the bond.  That implies both the redeem script and the public key must be revealed to those DEX instance(s) via an out-of-band mechanism, and therefore wrapped P2PKH would provide no additional security in that regard.  The redeem script and public keys will only known to the originator and the DEX instance(s) that are validating them, so as long as neither reveals them publicly, the public keys will also only known to them.

chappjc:
That's for providing that input.  It confirms what I had concluded that there is actually nothing to be gained from wrapping p2pkh in p2sh (assuming server can't be trusted to keep the pubkey secret).  I had started with p2pk instead of p2pkh and will revert it for the reasons you cited, but also because it simplifies part of the validation (direct pubkey comparision after RecoverCompact, rather than having to hash160 (Decred's way!) the pubkey).

>  The redeem script and public keys will only be known to the originator and the DEX instance(s) that are validating them, so as long as neither reveals them publicly, the public keys will also only known to them

I think this is what even made me consider wrapping p2pkh, but at the end of the day, I don't think "trusting" the server to not reveal the pubkey is smart.  Plus in a mesh config, it's likely going to be necessary to have it propagated between DEX nodes.

davecgh:
>  think this is what even made me consider wrapping p2pkh, but at the end of the day, I don't think "trusting" the server to not reveal the pubkey is smart. Plus in a mesh config, it's likely going to be necessary to have it propagated between DEX nodes.

Right.  I didn't explicitly state it, but what I also meant to imply by that final statement is that the server needs the public key to validate things anyway, so if they are going to reveal the public keys by way of the P2PK redeem script, they can just reveal the public keys directly too.  It's the same trust model regardless.

chappjc:
Just did a challenging rebase after HD keys and stdaddr PRs were merged.  I've not retested with this however, so this is a more "drafty" than before.  It's probably broken at the moment.

The `(*Core).register` method is removed since it's primary functions were: (1) switch between legacy and hd account creation, and (2) deal with suspended accounts by defaulting to legacy keys.  Both of these are slated for removal in the release containing this PR.  This is because the server's `'config'` response will contain the required pubkey, and the tier system removes the notion of account suspension entirely.

`(*Core).PreRegister` remains (below `AddBond`), although it was not in the initial draft of this PR.  I kept it because the UX issue it solved with the introduction of HD keys still exists regardless of the change from fee => bond.  However, there is no longer an `inprocessRegistration` since the new protocol does not use the legacy `'register'` route and there is no longer a fee address. Existing account discovery is done by attempting `'connect'` via `authDEX`.  `PreRegister` uses `setupCryptoV2` exclusively when attempting to connect to a new host, where "new" means not in the `c.conns` map and thus not in the DB as this map is populated on `Core` startup.

Also, `PreRegister` no longer returns a `paid bool`, as this notion has changed with bond. Consumers are intended to consult the returned `*core.Exchange` structure that includes `Tier int64` and `BondsPending bool` fields.  This is documented.

chappjc:
Superseded by https://github.com/decred/dcrdex/pull/1480

PR #1126, opened 2021-07-13
Fix maxOrder when not enough balance

### LINKED ISSUE #1097: ui: buy side order size dialog issues
The Buy mode of the Limit Order dialog seems to be broken with respect to (a) the preview of total order size in quote units and (b) Max Size.

Refresh the /markets page and do not click on Sell because that changes the behavior.

image

The Max Buy isn't filled in, but the main issue I believe is the missing text that would read "Total: 0.1446 BTC" just below Quantity.  See the Sell dialog:

image

The browser console has the following error with the Buy dialog:

image

vctt94:
This PR closes #1097.

It also adds a `Not enough funds` message when not having balance for the purchase.

image

### DIFF
--- a/client/webserver/site/src/js/markets.js
+++ b/client/webserver/site/src/js/markets.js
@@ -65,7 +65,7 @@ export default class MarketsPage extends BasePage {
       'hoverVolume', 'chartLegend', 'chartErrMsg',
       // Max order section
       'maxOrd', 'maxLbl', 'maxFromLots', 'maxFromAmt', 'maxFromTicker',
-      'maxToAmt', 'maxToTicker', 'maxAboveZero', 'maxNotEnough', 'maxLotBox', 'maxFromLotsLbl',
+      'maxToAmt', 'maxToTicker', 'maxAboveZero', 'maxLotBox', 'maxFromLotsLbl',
       'maxBox'
     ])
     this.main = main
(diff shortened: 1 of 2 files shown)


chappjc:
Trivial conflict resolution for you.

JoeGruffins:
I was able to reproduce this when I learned that you need to have a high rate.

It is possible to have the message show if you so something like rate 1 -> 99999 -> 1
When you go back to one it keeps the message from 99999

hasnotenuf

chappjc (changes requested):
OK it works well now when I set my simnet's BTC balance to almost zero and try to get a max buy estimate.  No more `toString` error.

image

But the "Not enough funds..." message runs into the "0 lots" test.  And with 0 lots now clearly shown, I'm not sure if the `maxNotEnough` span serves a purpose anymore.

Also I've seen the  "Not enough funds..." message hang around even after it's not applicable.  Let's remove it.

vctt94:
Good catch @JoeGruffins haven't noticed it.

I removed the span.

chappjc (approved):
LGTM.  @JoeGruffins can reevaluate at his leisure now that the span is removed.

n2s: This needs to be cherry-picked for v0.2.1.

client/webserver/site/src/js/markets.js
```diff
@@ -675,6 +675,8 @@ export default class MarketsPage extends BasePage {
     const quoteAsset = app.assets[order.quote]
     const total = Doc.formatCoinValue(order.rate / 1e8 * order.qty / 1e8)
     page.orderPreview.textContent = `Total: ${total} ${quoteAsset.symbol.toUpperCase()}`
+    const isSell = this.isSell()
+    isSell ? this.preSell() : this.preBuy()
```
buck54321, line 679:
```javascript
if (this.isSell()) this.preSell()
else this.preBuy()
```
Same LOC, better semantics.

buck54321 (changes requested):

PR #1127, opened 2021-07-14
js intl

vctt94:
This PR is a work torward the internationalization. 

It is a collective work with #1051 and I am focusing on the js internationalization here.


chappjc:
PR #1051 is now merged.

buck54321:
Can we split off the **pt-BR** work some time before merging?

chappjc:
Also pls squash it down before hitting ready for review.  I've been testing a little here and there and it's mostly working, except the things we've already chatted about in matrix: a number of notifications in core.go still need to go through `formatString`, and teh map in client/core/locale_htfn.go has to be a `map[string]*Translation` to get the subject translations too.  Thanks, @vctt94

client/webserver/site/src/js/locales.js
```diff
@@ -0,0 +1,121 @@
...
+
+const localesMap = {
+  'en-us': templateKeys,
+  'pt-br': templateKeysPT
+}
+
+export default class Locales {
+  constructor (locale) {
+    console.log((document.documentElement.lang).toLowerCase())
+    // lang is set programatically by backend
```
chappjc, line 103: Since it's now part of the localized html built at compile-time, it would be a little clearer and less ambiguous just to say that lang is set as a `lang` attribute in the `<html>` document/root element.

client/webserver/site/src/js/locales.js
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  // formatDetails will format the message to its locale.
+  // need to add one for plurals
+  formatDetails (stringKey, args = undefined) {
+    return this.stringTemplateParser(localesMap[this.locale][stringKey], args)
+  }
+
+  stringTemplateParser (expression, valueObj) {
+    const templateMatcher = /{{\s?([^{}\s]*)\s?}}/g
```
chappjc, line 114: Let's break this regex literal down with some docs: something in brackets like  `{{ text }}`, where there is a space before and after `text`, and `text` is is anything but brackets `{}` or whitespace.  And it is a global replace (`g`)

client/webserver/site/src/js/locales.js
```diff
@@ -0,0 +1,121 @@
...
+}
+
+const localesMap = {
+  'en-us': templateKeys,
+  'pt-br': templateKeysPT
+}
+
+export default class Locales {
+  constructor (locale) {
+    console.log((document.documentElement.lang).toLowerCase())
```
chappjc, line 102: Can remove.

client/webserver/site/src/js/locales.js
```diff
@@ -0,0 +1,121 @@
...
+    return this.stringTemplateParser(localesMap[this.locale][stringKey], args)
+  }
+
+  stringTemplateParser (expression, valueObj) {
+    const templateMatcher = /{{\s?([^{}\s]*)\s?}}/g
+    const text = expression.replace(templateMatcher, (substring, value, index) => {
+      value = valueObj[value]
+      return value
+    })
+    return text
```
chappjc, line 119: `return expression.replace(...`?

client/webserver/site/src/js/locales.js
```diff
@@ -0,0 +1,121 @@
...
+  // need to add one for plurals
+  formatDetails (stringKey, args = undefined) {
+    return this.stringTemplateParser(localesMap[this.locale][stringKey], args)
+  }
+
+  stringTemplateParser (expression, valueObj) {
+    const templateMatcher = /{{\s?([^{}\s]*)\s?}}/g
+    const text = expression.replace(templateMatcher, (substring, value, index) => {
+      value = valueObj[value]
+      return value
```
chappjc, line 117: `return valueObj[value]`?

client/webserver/site/src/js/locales.js
```diff
@@ -0,0 +1,121 @@
...
+    this.locale = (document.documentElement.lang).toLowerCase()
+  }
+
+  // formatDetails will format the message to its locale.
+  // need to add one for plurals
+  formatDetails (stringKey, args = undefined) {
+    return this.stringTemplateParser(localesMap[this.locale][stringKey], args)
+  }
+
+  stringTemplateParser (expression, valueObj) {
```
chappjc, line 113: I think it's worth documenting usage of the function, where `expression` is a template string with placeholders of a certain format (e.g. `buy {{ asset }}`) , and `valueObj` is an object with keys equal to the placeholder keys (e.g. `{"asset": "dcr"}`)

client/webserver/site/src/js/locales.js
```diff
@@ -0,0 +1,121 @@
...
+    const templateMatcher = /{{\s?([^{}\s]*)\s?}}/g
+    const text = expression.replace(templateMatcher, (substring, value, index) => {
```
chappjc, line 115: We're intentionally not using the `index` arg, so can we just omit that? `(_, value) => {`?  Or something like that?  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_function_as_a_parameter

client/webserver/site/src/js/forms.js
```diff
@@ -223,12 +229,12 @@ export class WalletConfigForm {
...
       Doc.hide(this.showIcon)
       Doc.show(this.hideIcon, this.otherSettings)
-      this.showHideMsg.textContent = 'hide additional settings'
+      this.showHideMsg.textContent = window.locales.formatDetails(ID_SHOW_ADDIIONAL_SETTINGS)
       return
     }
     Doc.hide(this.hideIcon, this.otherSettings)
     Doc.show(this.showIcon)
-    this.showHideMsg.textContent = 'show additional settings'
+    this.showHideMsg.textContent = window.locales.formatDetails(ID_HIDE_ADDIIONAL_SETTINGS)
```
chappjc, line 237: These two show/hide texts appear switched

client/webserver/site/src/js/forms.js
```diff
@@ -27,7 +33,7 @@ export class NewWalletForm {
     bind(form, fields.submitAdd, async () => {
       const pw = fields.nwAppPass.value || (this.pwCache ? this.pwCache.pw : '')
       if (!pw && !State.passwordIsCached()) {
-        fields.newWalletErr.textContent = 'app password cannot be empty'
+        fields.newWalletErr.textContent = window.locales.formatDetails(ID_NO_PASS_ERROR_MSG)
```
chappjc, line 36: Why is this one different now?  No longer says "app" password, but that's what this is referring to.

client/webserver/site/src/js/locales.js
```diff
@@ -0,0 +1,121 @@
+export const ID_NO_PASS_ERROR_MSG = 'ID_NO_PASS_ERROR_MSG'
+export const ID_NO_APP_PASS_ERROR_MSG = 'ID_NO_APP_PASS_ERROR_MSG'
+export const ID_SET_BUTTON_BUY = 'ID_SET_BUTTON_BUY'
+export const ID_SET_BUTTON_SELL = 'ID_SET_BUTTON_SELL'
+export const ID_OFF = 'ID_OFF'
```
chappjc, line 5: This all seems peculiar to me.  Seems like these consts should just be an enumeration or something with integers.  Is this really what JavaScript forces us to do?

client/webserver/site/src/js/markets.js
```diff
@@ -784,6 +800,7 @@ export default class MarketsPage extends BasePage {
     Doc.show(page.maxOrd, page.maxLotBox, page.maxAboveZero)
     const sell = this.isSell()
     page.maxFromLots.textContent = maxOrder.lots.toString()
+    // XXX add plural into format details, so we don't need this
```
chappjc, line 803: How about just `ID_LOT` and `ID_LOTS`?

client/webserver/site/src/js/locales.js
```diff
@@ -0,0 +1,121 @@
...
+  [ID_ORDER_PREVIEW]: 'Total: {{ total }} {{ asset }}',
+  [ID_CALCULATING]: 'calculating...',
+  [ID_ESTIMATE_UNAVAILABLE]: 'estimate unavailable',
+  [ID_NO_ZERO_RATE]: 'zero rate not allowed',
+  [ID_NO_ZERO_QUANTITY]: 'zero quantity not allowed',
+  [ID_TRADE]: 'trade',
+  [ID_NO_ASSET_WALLET]: 'No {{ asset }} wallet',
+  [ID_EXECUTED]: 'executed',
+  [ID_BOOKED]: 'booked',
+  [ID_CANCELING]: 'cancelling',
```
chappjc, line 57: Seeing as this is en-**US**, let's change to American english (canceling) instead of British english (cancelling) here.

chappjc (changes requested):

client/core/core.go
```diff
@@ -1171,6 +1171,7 @@ func New(cfg *Config) (*Core, error) {
 			return nil, fmt.Errorf("Error parsing language tag %q: %w", cfg.Language, err)
```
chappjc, line 1171: I'm noticing that `Core's language matching is much more strict that `Webserver`'s.  I think we'll need to mirror the logic used in `Webserver`'s constructor here.  Something along the lines of this:

```go
		acceptLangs, _, err := language.ParseAcceptLanguage(cfg.Language)
		if err != nil {
			return nil, fmt.Errorf("unable to parse requested language: %w", err)
		}
		// fmt.Println("parsed input:", acceptLangs)
		var langs []language.Tag
		for locale := range locales {
			tag, err := language.Parse(locale)
			if err != nil {
				return nil, fmt.Errorf("bad %v: %w", locale, err)
			}
			langs = append(langs, tag)
		}
		// fmt.Println("defined langs:", langs)
		matcher := language.NewMatcher(langs)
		tag, _, conf := matcher.Match(acceptLangs...)
		switch conf {
		case language.Exact:
		case language.High, language.Low:
			cfg.Logger.Infof("Using language %v", tag)
		case language.No:
			return nil, fmt.Errorf("no match for %q in recognized languages %v", cfg.Language, langs)
		}
		lang = tag
```

client/webserver/site/src/js/locales.js
chappjc: Can leave this if it's Decrediton's approach, unless you or someone else has a better suggestion.

client/webserver/locales/en-us.go
```diff
@@ -165,4 +165,6 @@ var EnUS = map[string]string{
 	"Restoration Seed":       "Restoration Seed",
 	"Restore from seed":      "Restore from seed",
 	"Import Account":         "Import Account",
+	"no_wallet":              "no wallet",
+	"create_a_x_wallet":      "Create a {{.Info.Name}} Wallet", // .Info.Name will be an asset symbol like DCR
```
chappjc, line 169: To document the reasoning for this, in portuguese we use that different from english, so we say "criar uma carteira decred" like it was: "create a wallet decred".  So doing ` <button data-action="create">[[[Create a]]] {{.Info.Name}} [[[Wallet]]]</button>` created a challenge.
My suggestion was simply to use go html template xpressions in the locale strings and let them find their way in to the localized_html files, which they do.  `<button data-action="create">Criar uma carteira {{.Info.Name}}</button>`

chappjc: Also we can now get rid of the `"Create a"` map entry, correct?

chappjc (approved):
Working well.  We can discuss tweaks to the translation machinery in https://github.com/decred/dcrdex/pull/1185, which will follow this.

PR #1128, opened 2021-07-15
btc: with dcrd's rpcclient, errors are dcrjson.RPCError

chappjc:
With both client and server btc asset backends using dcrd's rpcclient,
the error type from all methods including `RawRequest` would be
`dcrjson.RPCError` not `btcjson.RPCError`.

The numeric codes will still pertain the btcjson constants, even though
the `dcrjson.RPCError.Code` field is of type `dcrjson.RPCErrorCode`.

Thus, error matching first interprets the error as a `dcrjson.RPCError`,
and then compares the `int` representation of `Code` to the expected
`int` code form btcjson.

### DIFF
--- a/server/asset/dcr/dcr.go
+++ b/server/asset/dcr/dcr.go
@@ -345,6 +345,9 @@ func (dcr *Backend) FundingCoin(ctx context.Context, coinID []byte, redeemScript
 	}
 	utxo, err := dcr.utxo(ctx, txHash, vout, redeemScript)
 	if err != nil {
+		if isTxNotFoundErr(err) {
+			return nil, asset.CoinNotFoundError
+		}
 		return nil, err
 	}
 	if utxo.nonStandardScript {
@@ -405,7 +408,7 @@ func (dcr *Backend) TxData(coinID []byte) ([]byte, error) {
 		if isTxNotFoundErr(err) {
 			return nil, asset.CoinNotFoundError
 		}
-		return nil, fmt.Errorf("GetRawTransactionVerbose for txid %s: %w", txHash, err)
+		return nil, fmt.Errorf("TxData: GetRawTransactionVerbose for txid %s: %w", txHash, err)
 	}
 	return stdaddrTx.MsgTx().Bytes()
 }
@@ -484,7 +487,7 @@ func (dcr *Backend) OutputSummary(txHash *chainhash.Hash, vout uint32) (txOut *T
 	}
 
 	if int(vout) > len(verboseTx.Vout)-1 {
-		err = asset.CoinNotFoundError
+		err = asset.CoinNotFoundError // should be something fatal?
 		return
 	}
 
@@ -836,7 +839,7 @@ func (dcr *Backend) newTXIO(txHash *chainhash.Hash) (*TXIO, int64, error) {
 		if isTxNotFoundErr(err) {
 			return nil, 0, asset.CoinNotFoundError
 		}
-		return nil, 0, fmt.Errorf("GetRawTransactionVerbose for txid %s: %w", txHash, translateRPCCancelErr(err))
+		return nil, 0, fmt.Errorf("newTXIO: GetRawTransactionVerbose for txid %s: %w", txHash, translateRPCCancelErr(err))
 	}
 	tx, err := dcr.transaction(txHash, verboseTx)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


chappjc:
All simnet trade disruption tests in trade_simnet_test.go pass when tested on top of https://github.com/decred/dcrdex/pull/1096#pullrequestreview-707784868 (now merged).

JoeGruffins (approved):
Looks right to me. I see the changes are already in client.

chappjc:
> Looks right to me. I see the changes are already in client.

This PR makes the changes to client too.  Which changes are already in client?

JoeGruffins:
It looks like I was mistaken. This does add to both. Not sure why I got confused there.

chappjc:
Hmm, actually thinking about the client side changes, I might want to make a client-only backport for 0.2.1 because I'm pretty certain that the 0.2 branch had the dcrd rpcclient being used for btc, but only on client, not server.  And I think there is some improper error handling as a result.  My recollection is that it's for finding redemptions and contract audits, both of which would retry anyway, but probably with some extra delay and a scary error message.

chappjc:
Wow I don't know when this broke, but server rejects orders **funded by DCR** when the coin isn't found instead of searching (like it should with txsplit enable).

On simnet with txsplit enabled, and server and client on different nodes (alpha vs beta) , this is a problem.  e.g.

```
[TRC] MKT: Searching for dcr coins [79c41b1e065602867fe68bfe162edfa3d7f99cb07ce30dd9fae282c4425edbc8:0] for new limit order
[ERR] MKT: Error retreiving limit order funding coin ID c8db5e42c482e2fad90de37cb09cf9d7a3df2e16fe8be67f860256061e1bc47900000000. user = 23bf67966195c53767d27613802693b55317499d05a69dc354675036537f4c54: GetRawTransactionVerbose for txid 79c41b1e065602867fe68bfe162edfa3d7f99cb07ce30dd9fae282c4425edbc8: -5: No information available about transaction 79c41b1e065602867fe68bfe162edfa3d7f99cb07ce30dd9fae282c4425edbc8
[DBG] MKT: Error going to user 23bf67966195c53767d27613802693b55317499d05a69dc354675036537f4c54: error code 36: error retrieving coin ID c8db5e42c482e2fad90de37cb09cf9d7a3df2e16fe8be67f860256061e1bc47900000000
```

It is supposed to retry and look like this:

```
[TRC] MKT: Searching for btc coins [525efba12f407f6a98785f3c0753f6de73ea81c6a7afbfb75dbc1ac85f439ae0:0] for new limit order
[TRC] MKT: Found and validated btc coins [525efba12f407f6a98785f3c0753f6de73ea81c6a7afbfb75dbc1ac85f439ae0:0] for new limit order
[TRC] MKT: Received order bf12d02b41658154439faf870002d1af28cd424ddaec745ae9903f33f603e366 at 2021-07-23 15:51:58.903 +0000 UTC
```

I have it fixed in a pending commit, but I'm baffled about why it didn't do that before.

server/asset/dcr/dcr.go
```diff
@@ -963,7 +966,10 @@ func (dcr *Backend) getUnspentTxOut(ctx context.Context, txHash *chainhash.Hash,
 func (dcr *Backend) getTxOutInfo(ctx context.Context, txHash *chainhash.Hash, vout uint32) (*chainjson.GetTxOutResult, *chainjson.TxRawResult, []byte, error) {
 	verboseTx, err := dcr.node.GetRawTransactionVerbose(ctx, txHash)
 	if err != nil {
-		return nil, nil, nil, fmt.Errorf("GetRawTransactionVerbose for txid %s: %w", txHash, translateRPCCancelErr(err))
+		if isTxNotFoundErr(err) { // since we're calling gettxout after this, check now
+			return nil, nil, nil, asset.CoinNotFoundError
+		}
```
chappjc, line 971: I'm at a loss to understand how we did without this.  Today I was suddenly getting errors from server when placing orders with txsplit, such that the server is expected not to see the coin initially and keep searching for a short period.  This definitely used to work, but it's clear that `getTxOutInfo` requires this check because unlike on the btc backend, we are calling getrawtransaction *before* gettxout.  The order router was rejecting orders because the error `FundingCoin` was not `asset.CoinNotFound`.

I'd really like to know when this broke, but it was very recent because I quite often use txsplit in simnet and testnet.

chappjc: Tracked down the change to https://github.com/decred/dcrdex/commit/bc4e19932d4ba242b55035e10284483eab84aada#diff-70ea3126faa1f8216d381a32b3c2bc13238608b0045a5316d83949e091d2dff7L953-L961
The gettxout call was moved after the sendrawtransaction call.  I've confirmed that this is not an issue on `release-v0.2` so no need to backport this server/asset/dcr change, although I do intend to add the client/asset/btc/btc.go to `release-v0.2`.

chappjc:
@JoeGruffins I've tacked on https://github.com/decred/dcrdex/pull/1128/commits/6a0d7ec6317059e6f8ed3dc4d323425b22e17d32 after you approved.  Even though it does not pertain to the `RPCError` type issue this PR is primarily addressing, it's very closely related and it's tested on this branch, so including here.
Please have a look at that commit and my description in https://github.com/decred/dcrdex/pull/1128#discussion_r675683513 of the issue.

PR #1129, opened 2021-07-16
Check LotSize and RateStep for older servers compatibility

vctt94:
This diff was made by @chappjc in order for newer clients to still have older servers compatibility.

This still open for discussion if needed.


chappjc (approved):
This worked for both @vctt94 and @amassarwi to work with the 0.2 testnet server with a 0.3+pre client.

PR #1130, opened 2021-07-19
harness/eth: Fix mining script.

JoeGruffins:
Immediatly stopping the miner after starting to mine was causing
    transactions to not be included in blocks. Change to start once and
    stop a fraction of a second before N blocks should be mined.

    Update internal geth to slightly after 1.10.5 and set harness to start
    at the london fork.

### DIFF
--- a/dex/testing/eth/create-node.sh
+++ b/dex/testing/eth/create-node.sh
@@ -31,6 +31,7 @@ chmod +x "${NODES_ROOT}/harness-ctl/${NAME}"
 
 # Write mine script if CHAIN_ADDRESS is present.
 if [ "${CHAIN_ADDRESS}" != "_" ]; then
+  # The mining script may end up mining more or less blocks than specified.
   cat > "${NODES_ROOT}/harness-ctl/mine-${NAME}" <<EOF
 #!/usr/bin/env bash
   NUM=1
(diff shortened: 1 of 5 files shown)


dex/testing/eth/create-node.sh
```diff
@@ -31,33 +31,25 @@ chmod +x "${NODES_ROOT}/harness-ctl/${NAME}"
...
       ''|*[!0-9]*)  ;;
       *) NUM=\$1 ;;
   esac
-  for i in \$(seq \$NUM) ; do
-    "${NODES_ROOT}/harness-ctl/${NAME}" attach --preload "${MINE_JS}" --exec 'mine()'
-  done
+  echo "Mining..."
+  BEFORE=\$("${NODES_ROOT}/harness-ctl/${NAME}" attach --exec 'eth.blockNumber')
+  "${NODES_ROOT}/harness-ctl/${NAME}" attach --exec 'miner.start()' > /dev/null
+  sleep \$(echo "scale=4;\$NUM-0.9" | bc)
```
JoeGruffins, line 45: I don't think scale=4 is needed or even used in subtraction.

dex/testing/eth/harness.sh
```diff
@@ -202,8 +203,6 @@ do
   "${NODES_ROOT}/harness-ctl/mine-alpha" "5"
 done
 
-# Transactions can take an eternity to be mined...
-# TODO: Determine why this is.
```
chappjc, line 206: In a nutshell, why was this?

chappjc (approved):

dex/testing/eth/harness.sh
JoeGruffins: I think it comes down to a race here https://github.com/ethereum/go-ethereum/blob/f05419f0fb8c5328dca92ea9fb184d082300344a/miner/worker.go#L387-L409

If you stop before this switch can make it down to including more transactions at `<-timer.C`, it just doesn't include them.

dex/testing/eth/create-node.sh
```diff
@@ -31,33 +31,25 @@ chmod +x "${NODES_ROOT}/harness-ctl/${NAME}"
...
       ''|*[!0-9]*)  ;;
       *) NUM=\$1 ;;
   esac
-  for i in \$(seq \$NUM) ; do
-    "${NODES_ROOT}/harness-ctl/${NAME}" attach --preload "${MINE_JS}" --exec 'mine()'
-  done
+  echo "Mining..."
+  BEFORE=\$("${NODES_ROOT}/harness-ctl/${NAME}" attach --exec 'eth.blockNumber')
+  "${NODES_ROOT}/harness-ctl/${NAME}" attach --exec 'miner.start()' > /dev/null
+  sleep \$(echo "\$NUM-0.9" | bc)
```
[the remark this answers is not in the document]
JoeGruffins: Can change if need be. That is super hard to read.

JoeGruffins:
Sorry for the small change.

dex/testing/eth/harness.sh
chappjc: Hmm, looking at the git blames, I see they made the timer duration adjustable: https://github.com/ethereum/go-ethereum/pull/17444

JoeGruffins: Hadn't noticed the `RecommitInterval` setting until now.

JoeGruffins: I tried some logs, and it does seem that it's the case that txs aren't included when we stop before the timer can check there, I am not sure where the txs are sometimes included even when we don't make it there, however.

dex/testing/eth/harness.sh
```diff
@@ -184,9 +185,9 @@ echo "Sending 5000 eth to delta and gamma."
 "${NODES_ROOT}/harness-ctl/alpha" "attach --preload ${NODES_ROOT}/harness-ctl/send.js --exec send(\"${ALPHA_ADDRESS}\",\"${GAMMA_ADDRESS}\",${SEND_AMT})"
 "${NODES_ROOT}/harness-ctl/alpha" "attach --preload ${NODES_ROOT}/harness-ctl/send.js --exec send(\"${ALPHA_ADDRESS}\",\"${DELTA_ADDRESS}\",${SEND_AMT})"
 
-# Our transactions will PROBABLY appear in one of these blocks.
 echo "Mining some blocks"
-"${NODES_ROOT}/harness-ctl/mine-beta" "15"
+"${NODES_ROOT}/harness-ctl/mine-alpha" "2"
+"${NODES_ROOT}/harness-ctl/mine-beta" "2"
```
chappjc, line 190: So this ended up working better?  It wasn't in the previous revision.

JoeGruffins: Yes. It appears that alternating is also important. Although, it was alternating at 15 and 15 blocks, it seems that is overkill, and it also sometimes cause the second node mined to fail to accept blocks from the first node, apparently because there were too many blocks from the first node initially.

PR #1132, opened 2021-07-19
[0.2.1] backports

chappjc:
This bumps the app versions to 0.2.1 and backports the following:

- client/asset/trade: incorrect asset symbol logged
- client/asset/dcr: shutdown ws client on Connect error
- max order size computations could easily cause hang while iterating on max lots given no initial fees and large number of lots

The `OrderBook` race fix is more nuanced so that is separate: https://github.com/decred/dcrdex/pull/1131

### DIFF
--- a/server/cmd/dcrdex/version.go
+++ b/server/cmd/dcrdex/version.go
@@ -25,7 +25,7 @@ const (
 	AppName  string = "dcrdex"
 	AppMajor uint   = 0
 	AppMinor uint   = 2
-	AppPatch uint   = 0
+	AppPatch uint   = 1
 )
 
 // go build -v -ldflags "-X main.appPreRelease= -X main.appBuild=$(git rev-parse --short HEAD)"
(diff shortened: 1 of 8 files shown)


chappjc:
@JoeGruffins We have enough important bug fixes for 0.2 that we need to release p1.  Pls have a look at these cherry-picked commits for the `release-v0.2` branch.  Thanks.

server/cmd/dcrdex/version.go
```diff
@@ -25,7 +25,7 @@ const (
 	AppName  string = "dcrdex"
 	AppMajor uint   = 0
 	AppMinor uint   = 2
-	AppPatch uint   = 0
+	AppPatch uint   = 1
```
JoeGruffins, line 28: I don't think anything changed with the server. Is it important to keep the versions the same? But I guess release version is the same as this, which makes sense to keep the same.

JoeGruffins (approved):

server/cmd/dcrdex/version.go
chappjc: I'd probably bump them together regardless, but there was one commit for the [0.2.1 milestone](https://github.com/decred/dcrdex/milestone/13?closed=1) that touched server, the script version handling.

PR #1133, opened 2021-07-19
server: Allow cancel orders while market is suspended.

### LINKED ISSUE #792: server/market: allow cancel orders while market is suspended
This will require an alternate order processing path because a market is suspended by draining the epoch processing pipeline and actually letting Run return. (Resume starts Run again). The simplest way to allow cancel orders is simply to process them directly with no epoch shuffling and matching.  Whether or not preimages should still be requested in this context is not clear, but they would serve no purpose except perhaps consistent book keeping.

martonp:
This PR adds the ability for clients to submit cancel orders while a market is suspended.

The database entries after a cancel order during a market suspension will be identical to one that gets matched while the market is running, except the preimage will not be recorded. Also there will be no entry in the epochs table for the epoch in which this order was "matched". 

Closes #792

### DIFF
--- a/dex/order/order.go
+++ b/dex/order/order.go
@@ -752,6 +752,16 @@ func ValidateOrder(ord Order, status OrderStatus, lotSize uint64) error {
 	return nil
 }
 
+// ExtractAddress extracts the address from the order. If the order is a cancel
+// order, an empty string is returned.
+func ExtractAddress(ord Order) string {
+	trade := ord.Trade()
+	if trade == nil {
+		return ""
+	}
+	return trade.Address
+}
+
 // Some commonly used time transformations.
 var unixMilli = encode.UnixMilli
 var unixMilliU = encode.UnixMilliU
--- a/server/market/market.go
+++ b/server/market/market.go
@@ -681,7 +681,6 @@ func (m *Market) processCancelOrderWhileSuspended(rec *orderRecord, errChan chan
 	}
 
 	errChan <- sendResponseErr
-	return
 }
 
 // SubmitOrderAsync submits a new order for inclusion into the current epoch.
--- a/server/market/market.go
+++ b/server/market/market.go
@@ -596,9 +596,9 @@ func (m *Market) SubmitOrder(rec *orderRecord) error {
 // 1. Removes the target order from the book.
 // 2. Unlocks the order coins.
 // 3. Updates the storage with the new cancel order and cancels the existing limit order.
-// 3. Responds to the client that the order was received.
-// 4. Sends the unbooked order to the order feeds.
-// 5. Creates a match object, stores it, and notifies the client of the match.
+// 4. Responds to the client that the order was received.
+// 5. Sends the unbooked order to the order feeds.
+// 6. Creates a match object, stores it, and notifies the client of the match.
 func (m *Market) processCancelOrderWhileSuspended(rec *orderRecord, errChan chan<- error) {
 	co, ok := rec.order.(*order.CancelOrder)
 	if !ok {
@@ -629,14 +629,14 @@ func (m *Market) processCancelOrderWhileSuspended(rec *orderRecord, errChan chan
 	// committed to the storage.
 	respMsg, err := m.orderResponse(rec)
 	if err != nil {
-		errChan <- fmt.Errorf("Failed to create order response: %v", err)
+		errChan <- fmt.Errorf("failed to create order response: %w", err)
 		return
 	}
 
 	dur := int64(m.EpochDuration())
 	now := encode.UnixMilli(time.Now())
 	epochIdx := now / dur
-	if err := m.storage.NewArchivedCancel(co, epochIdx, int64(m.EpochDuration())); err != nil {
+	if err := m.storage.NewArchivedCancel(co, epochIdx, dur); err != nil {
 		errChan <- err
 		return
 	}
@@ -645,12 +645,8 @@ func (m *Market) processCancelOrderWhileSuspended(rec *orderRecord, errChan chan
 		return
 	}
 
-	// If we fail to send the response, then we send this error at the end of
-	// the function. If we send the error here and return the function, then
-	// the match will not be recorded and the server DB will be in an inconsistent
-	// state.
-	sendResponseErr := m.auth.Send(rec.order.User(), respMsg)
-	if sendResponseErr != nil {
+	err = m.auth.Send(rec.order.User(), respMsg)
+	if err != nil {
 		log.Errorf("Failed to send cancel order response: %v", err)
 	}
 
@@ -675,24 +671,28 @@ func (m *Market) processCancelOrderWhileSuspended(rec *orderRecord, errChan chan
 		FeeRateBase:  m.getFeeRate(m.Base(), m.baseFeeFetcher),
 		FeeRateQuote: m.getFeeRate(m.Quote(), m.quoteFeeFetcher),
 	}
-	m.storage.InsertMatch(&match)
+	// insertMatchErr is sent on errChan at the end of the function. We
+	// want to send the match request to the client even if this insertion
+	// fails.
+	insertMatchErr := m.storage.InsertMatch(&match)
+
 	makerMsg, takerMsg := matchNotifications(&match)
 	m.auth.Sign(makerMsg)
 	m.auth.Sign(takerMsg)
 	msgs := []msgjson.Signable{makerMsg, takerMsg}
 	req, err := msgjson.NewRequest(comms.NextID(), msgjson.MatchRoute, msgs)
 	if err != nil {
-		log.Errorf("Failed to create match reqeust: %v", err)
+		log.Errorf("Failed to create match request: %v", err)
 	} else {
 		err = m.auth.Request(rec.order.User(), req, func(_ comms.Link, resp *msgjson.Message) {
 			m.processMatchAcksForCancel(rec.order.User(), resp)
 		})
 		if err != nil {
-			log.Errorf("Failed to send match reqeust: %v", err)
+			log.Errorf("Failed to send match request: %v", err)
 		}
 	}
 
-	errChan <- sendResponseErr
+	errChan <- insertMatchErr
 }
 
 // matchNotifications creates a pair of msgjson.Match from a match.
@@ -742,7 +742,7 @@ func (m *Market) processMatchAcksForCancel(user account.AccountID, msg *msgjson.
 			fmt.Sprintf("expected %d acknowledgements, got %d", len(acks), expectedNumAcks))
 		return
 	}
-	log.Debugf("processMatchAcksForCancel: 'match' ack recieived from %v", user)
+	log.Debugf("processMatchAcksForCancel: 'match' ack received from %v", user)
 }
 
 // SubmitOrderAsync submits a new order for inclusion into the current epoch.
--- a/server/market/market.go
+++ b/server/market/market.go
@@ -60,6 +60,7 @@ type Swapper interface {
 	CheckUnspent(ctx context.Context, asset uint32, coinID []byte) error
 	UserSwappingAmt(user account.AccountID, base, quote uint32) (amt, count uint64)
 	ChainsSynced(base, quote uint32) (bool, error)
+	ProcessCancelMatchDuringSuspendedMarket(*order.Match) error
 }
 
 type DataCollector interface {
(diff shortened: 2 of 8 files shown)


server/market/market.go
```diff
@@ -614,6 +688,10 @@ func (m *Market) SubmitOrderAsync(rec *orderRecord) <-chan error {
 	select {
 	case <-m.running:
 	default:
+		if rec.order.Type() == order.CancelOrderType {
+			err := m.processCancelOrderWhileSuspended(rec)
```
JoeGruffins, line 692: I'm not sure if this is ever a problem, but the function this is in is `SubmitOrderAsync` and this has the potential to block for a bit with two db interactions.

JoeGruffins (commented):
Working very well!

Just a few nits inline. chapp and buck are the authorities here though. I have not spent much time on the server.

Also, please don't attempt to solve here, but canceling orders while a market is suspended still gives bad rep which doesn't seem necessary. Depending on when the order was placed, the user might not have know it was coming. They should not be penalized imo. Maybe others feel different?

server/market/market.go
```diff
@@ -587,6 +588,79 @@ func (m *Market) SubmitOrder(rec *orderRecord) error {
...
+		return err
+	}
+
+	respMsg, err := m.orderResponse(rec)
+	if err != nil {
+		return fmt.Errorf("error creating order response: %v", err)
+	}
+	if err := m.auth.Send(rec.order.User(), respMsg); err != nil {
+		return fmt.Errorf("error sending response: %v", err)
+	}
```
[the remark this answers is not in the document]
martonp: I think you're, better not to put the server DB into an inconsistent state.

server/market/market.go
martonp: It's not a problem now since `SubmitOrderAsync` is only called from `SubmitOrder` which blocks, but it's definitely not the intention for this function to block, so I'll update it.

martonp:
@JoeGruffins Thanks for the reivew. I fixed the nits that you mentioned. Regarding the issue of bad rep, I'm not calling `auth.RecordCancel` (the function that records the users cancel orders compared to completed orders) so the users will not get banned for cancelling during a suspended market.

chappjc:
I'll get into this soon, but don't worry much about cancel rate penalties.  They are very hard to hit, and frankly I think the cancel rate concept is going to either be removed or reimagined completely.

There are a few things I'll need to check, particularly in the DB.  One is there may be unforeseen consequences of orders that don't require preimages, quite a big exception.  These might be detected as preimage misses when a user connects and queries recompute score from the user's history in the DB.  I *think* only "revoked" orders will trip this, but if the preimage column is still null it might have some other side-effect.  I'll review it soon, but I recall that server-generated cancel orders that are created with revocation (either user behavior or just whole-market purge) also use a null preimage for the synthetic cancel order, so it's probably all good.

server/market/market.go
```diff
@@ -587,6 +588,101 @@ func (m *Market) SubmitOrder(rec *orderRecord) error {
...
+		return
+	}
+
+	// If we fail to send the response, then we send this error at the end of
+	// the function. If we send the error here and return the function, then
+	// the match will not be recorded and the server DB will be in an inconsistent
+	// state.
+	sendResponseErr := m.auth.Send(rec.order.User(), respMsg)
+	if sendResponseErr != nil {
+		log.Errorf("error sending response: %v", err)
```
chappjc, line 654: Yeah, a down connection doesn't affect our actions at this point.  Also can you change the log like:

`log.Errorf("Failed to send cancel order response: %v", err)`

server/market/market.go
```diff
@@ -587,6 +588,79 @@ func (m *Market) SubmitOrder(rec *orderRecord) error {
...
+// 3. Responds to the client that the order was recieved.
+// 4. Sends the unbooked order to the order feeds.
+// 5. Creates a match object and sends it to the swapper to notify the client of the match.
+func (m *Market) processCancelOrderWhileSuspended(rec *orderRecord) error {
+	co, ok := rec.order.(*order.CancelOrder)
+	if !ok {
+		return ErrInvalidOrder
+	}
+
+	lo, ok := m.book.Remove(co.TargetOrderID)
```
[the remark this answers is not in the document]
chappjc: Yeah our general convention for mutexes is to have the fields of a struct grouped by what mutex is intended to guard the fields.  In this case @JoeGruffins is right that the `settling` map does need the mutex locked.  I'm less sure about calling `book.Remove` at the moment because it has it's own internal synchronization mechanisms, but holding `bookMtx` for both ops is fine.

I'm pretty sure in most places when we lock `bookMtx` around accesses to the `book` field, we are really trying preform other operations with a consistent book view, rather than avoid a data race, but I do see some discrepancies.  I'll look closer.

server/market/market.go
```diff
@@ -587,6 +588,101 @@ func (m *Market) SubmitOrder(rec *orderRecord) error {
...
 
+// processCancelOrderWhileSuspended is called when cancelling an order while
+// the market is suspended and Run is not runnning. The error sent on errChan
+// is returned to the client.
+//
+// This function:
+// 1. Removes the target order from the book.
+// 2. Unlocks the order coins.
+// 3. Updates the storage with the new cancel order and cancels the existing limit order.
+// 3. Responds to the client that the order was recieved.
```
chappjc, line 599: Spelling: "runnning" and "recieved".

server/market/market.go
```diff
@@ -587,6 +588,101 @@ func (m *Market) SubmitOrder(rec *orderRecord) error {
...
+	if cancelable, err := m.CancelableBy(co.TargetOrderID, co.AccountID); !cancelable {
+		errChan <- err
+		return
+	}
+
+	m.bookMtx.Lock()
+	lo, ok := m.book.Remove(co.TargetOrderID)
+	if !ok {
+		errChan <- ErrTargetNotCancelable
+		return
```
chappjc, line 618: If you return here, `m.bookMtx` is still locked and the server will deadlock thereafter.

server/market/market.go
```diff
@@ -587,6 +588,101 @@ func (m *Market) SubmitOrder(rec *orderRecord) error {
...
+	delete(m.settling, co.TargetOrderID)
+	m.bookMtx.Unlock()
+
+	m.unlockOrderCoins(lo)
+
+	sTime := time.Now().Truncate(time.Millisecond).UTC()
+	co.SetTime(sTime)
+
+	// Create the client response here, but don't send it until the order has been
+	// commited to the storage.
```
chappjc, line 629: s/commited/committed

server/market/market.go
```diff
@@ -587,6 +588,101 @@ func (m *Market) SubmitOrder(rec *orderRecord) error {
...
+		Quantity: lo.Remaining(),
+		Rate:     lo.Rate,
+		Epoch: order.EpochID{
+			Idx: uint64(epochIdx),
+			Dur: m.EpochDuration(),
+		},
+		FeeRateBase:  m.getFeeRate(m.Base(), m.baseFeeFetcher),
+		FeeRateQuote: m.getFeeRate(m.Quote(), m.quoteFeeFetcher),
+	}
+	if err := m.swapper.ProcessCancelMatchDuringSuspendedMarket(&match); err != nil {
```
chappjc, line 678: This is my only real issue with `processCancelOrderWhileSuspended`.  It's unfortunate that Swapper is presently handling all `match` requests even if they are not negotiating matches (see `toMonitor` in `(*Swapper).Negotiate`).  I get that the reason is that the handler for the response is a `Swapper` method (`processMatchAcks`), but there's no conceptual reason why the `Swapper` should be involved for this or any cancel orders AFAIK.

I don't want to recommend large refactoring for all cancels, but perhaps just for `processCancelOrderWhileSuspended` we can look into implementing a match ack handler in this package.  Note that the acks are basically worthless unless they signal the start of a swap negotiation, and not strictly necessary even in that case for a successful swap.  That wasn't always the case, and the reason it's still organized this way presently.  Although the match ack signature is stored, I don't think it's actually required anymore, even for swap matches.

Hence I think a barebones handler just needs to unmarshal a `msgjson.Acknowledgement` and log it for debugging, since there won't be any swap matches batched with the cancel match in this non-epoch match.

chappjc (changes requested):
Great job on `processCancelOrderWhileSuspended`.

server/market/market.go
martonp: I did think it was a bit strange that I was calling `Swapper` when there is no swap happening :).
I've added a new handler, let me know what you think.

martonp:
@chappjc If other places are expecting to have a consistent book view because they held the mutex then it's better to hold it here too. Although I don't think anything will be happening in the market other than cancellations when the market is suspended.

martonp:
I checked, after reconnecting to the server, these cancels are not counted as a preimage miss.

chappjc:
> I checked, after reconnecting to the server, these cancels are not counted as a preimage miss.

Thanks.  I did check the query and it does require revoked status, so that would exclude canceled.  Thank you for verifying.

server/market/market.go
```diff
@@ -587,6 +588,163 @@ func (m *Market) SubmitOrder(rec *orderRecord) error {
...
 
+// processCancelOrderWhileSuspended is called when cancelling an order while
+// the market is suspended and Run is not running. The error sent on errChan
+// is returned to the client.
+//
+// This function:
+// 1. Removes the target order from the book.
+// 2. Unlocks the order coins.
+// 3. Updates the storage with the new cancel order and cancels the existing limit order.
+// 3. Responds to the client that the order was received.
```
chappjc, line 599: 3. 3.

server/market/market.go
```diff
@@ -587,6 +588,163 @@ func (m *Market) SubmitOrder(rec *orderRecord) error {
...
+		Quantity: lo.Remaining(),
+		Rate:     lo.Rate,
+		Epoch: order.EpochID{
+			Idx: uint64(epochIdx),
+			Dur: m.EpochDuration(),
+		},
+		FeeRateBase:  m.getFeeRate(m.Base(), m.baseFeeFetcher),
+		FeeRateQuote: m.getFeeRate(m.Quote(), m.quoteFeeFetcher),
+	}
+	m.storage.InsertMatch(&match)
```
chappjc, line 678: At least log any error.  If `InsertMatch` does fail, that sucks but it it's not a disaster as the order is already cancelled and a cancel order archived.  EDIT: see the last comment in the review as I think any error from `InsertMatch` is more important to send than a comms error.

server/market/market.go
```diff
@@ -587,6 +588,163 @@ func (m *Market) SubmitOrder(rec *orderRecord) error {
...
+		return
+	}
+	// The acknowledgment for both the taker and maker should come from the same user
+	expectedNumAcks := 2
+	if len(acks) != expectedNumAcks {
+		m.respondError(msg.ID, user, msgjson.AckCountError,
+			fmt.Sprintf("expected %d acknowledgements, got %d", len(acks), expectedNumAcks))
+		return
+	}
+	log.Debugf("processMatchAcksForCancel: 'match' ack recieived from %v", user)
```
chappjc, line 745: recieived

server/market/market.go
```diff
@@ -587,6 +588,163 @@ func (m *Market) SubmitOrder(rec *orderRecord) error {
...
+	respMsg, err := m.orderResponse(rec)
+	if err != nil {
+		errChan <- fmt.Errorf("Failed to create order response: %v", err)
+		return
+	}
+
+	dur := int64(m.EpochDuration())
+	now := encode.UnixMilli(time.Now())
+	epochIdx := now / dur
+	if err := m.storage.NewArchivedCancel(co, epochIdx, int64(m.EpochDuration())); err != nil {
```
chappjc, line 639: Use `dur` in the input arg.

server/market/market.go
```diff
@@ -587,6 +588,163 @@ func (m *Market) SubmitOrder(rec *orderRecord) error {
...
+	m.unlockOrderCoins(lo)
+
+	sTime := time.Now().Truncate(time.Millisecond).UTC()
+	co.SetTime(sTime)
+
+	// Create the client response here, but don't send it until the order has been
+	// committed to the storage.
+	respMsg, err := m.orderResponse(rec)
+	if err != nil {
+		errChan <- fmt.Errorf("Failed to create order response: %v", err)
```
chappjc, line 632: %w and lower case `error` message: `fmt.Errorf("failed to create order response: %w", err)`

server/market/market.go
```diff
@@ -587,6 +588,163 @@ func (m *Market) SubmitOrder(rec *orderRecord) error {
...
+	msgs := []msgjson.Signable{makerMsg, takerMsg}
+	req, err := msgjson.NewRequest(comms.NextID(), msgjson.MatchRoute, msgs)
+	if err != nil {
+		log.Errorf("Failed to create match reqeust: %v", err)
+	} else {
+		err = m.auth.Request(rec.order.User(), req, func(_ comms.Link, resp *msgjson.Message) {
+			m.processMatchAcksForCancel(rec.order.User(), resp)
+		})
+		if err != nil {
+			log.Errorf("Failed to send match reqeust: %v", err)
```
chappjc, line 691: "reqeust" in both spots

server/market/market.go
```diff
@@ -587,6 +588,163 @@ func (m *Market) SubmitOrder(rec *orderRecord) error {
...
+		}
+	}
+
+	errChan <- sendResponseErr
```
chappjc, line 695: I think that sending `sendResponseErr` back is not super valuable, at least not any more than errors from `InsertMatch` or `auth.Request`.  I do agree with your comment way above that even if a send fails, we still need to do `InsertMatch` though.
Any errors will find their way back to `handleCancel` and then up to the comms message handling, which generally just logs since that's all there is to do.
In short, I think it's fine to log send/request errors in this function, but be sure to return errors from "state" related issues like an error from `InsertMatch`.  That might be different if `processMatchAcksForCancel` actually had some effect, but the only error of real consequence is the one from `InsertMatch`.

chappjc (changes requested):
Looks like `(*Market).processMatchAcksForCancel` works out nicely instead of calling a `Swapper` method.

Just some minor stuff to address and we can merge.

server/market/market.go
martonp: This word is my nemesis.

JoeGruffins: Just remember "i before e except after c"

I have to repeat this to myself about 10 times a day.

server/market/market.go
martonp: Returning that error makes much more sense. I've updated it.

server/market/market.go
martonp: Nice, I didn't know about these wrapping errors.

server/market/market.go
martonp: I just put the i on both sides because I couldn't decide haha.
I'll keep repeating it though and it probably won't ever happen again :)

dex/order/order.go
```diff
@@ -752,6 +752,16 @@ func ValidateOrder(ord Order, status OrderStatus, lotSize uint64) error {
...
+// ExtractAddress extracts the address from the order. If the order is a cancel
+// order, an empty string is returned.
+func ExtractAddress(ord Order) string {
+	trade := ord.Trade()
+	if trade == nil {
+		return ""
+	}
+	return trade.Address
+}
+
```
JoeGruffins, line 764: It makes more sense to me for this (and ValidateOrder actually) to be part of the interface. I guess it's fine though.

server/market/market.go
```diff
@@ -587,6 +588,163 @@ func (m *Market) SubmitOrder(rec *orderRecord) error {
...
+		if err != nil {
+			log.Errorf("Failed to send match request: %v", err)
+		}
+	}
+
+	errChan <- insertMatchErr
+}
+
+// matchNotifications creates a pair of msgjson.Match from a match.
+func matchNotifications(match *order.Match) (makerMsg *msgjson.Match, takerMsg *msgjson.Match) {
```
JoeGruffins, line 699: With returns or args of the same type, there's a shorthand where you don't need the type twice. This simplifies to:
```
func matchNotifications(match *order.Match) (makerMsg, takerMsg *msgjson.Match) {
```

JoeGruffins (approved):
Works well for me.

chappjc (approved):

server/market/market.go
xaur: Make up any mnemonic that is easy to remember e.g. "a weighty sovereign reindeer receives forfeit caffeine".

PR #1135, opened 2021-07-22
Add table showing available markets when registering for dex.

### LINKED ISSUE #1125: client,ui: register dialog should summarize the DEX markets
Discussed here https://github.com/decred/dcrdex/pull/1102#discussion_r648656593

This stems from the fact that lot size is now a market setting, but in general a user registering on a server would like to know all of its capabilities and relevant settings.

martonp:
This PR adds a table on the confirm register form to display what markets the DEX provides.
Screen Shot 2021-07-21 at 3 29 28 PM

Also there is a bit of refactoring since the form was used in multiple places, but the code was completely replicated.

Closes #1125.


chappjc:
Looks very nice, @martonp!

chappjc (approved):
Looks good to me, see below simnet harness result:

image


Refactoring of forms and other frontend stuff looks good to me too, but @buck54321 will have to review all that since it's more extensive than I anticipated. The refactoring does look positive though with narrower concerns (e.g. getCertFile, getDexAddr, registerDEXFundsFail, etc), but it's not my strong suit.

martonp:
@chappjc I updated the bind function to be a class. All the logic stayed the same though.

chappjc:
A good number of conflicts created for you @martonp with the app seed PR merged, although this is just about ready to go pending a look from buck.

martonp:
I think it's better to wait for #1119 to be merged first, since that will cause some more conflicts here.

chappjc:
This is now unblocked with https://github.com/decred/dcrdex/pull/1119 merged.

martonp:
This is merged now.

client/webserver/site/src/html/forms.tmpl
```diff
@@ -140,25 +140,48 @@
...
+<div class="bg2 px-2 py-1 text-center fs18">
   Confirm Registration
   <div class="form-closer hoverbg"><span class="ico-cross"></span></div>
 </div>
 <div class="p-4">
   <div class="fs16">
-    <span {{if $passwordIsCached}}class="d-hide"{{end}}>Enter your app password to confirm DEX registration.</span>
-    When you submit this form, <span id="feeDisplay"></span> DCR will be spent from your Decred wallet to pay
-    registration fees.
+    Enter your app password to confirm DEX registration.
```
chappjc, line 149: Seems like this line needs to be hidden conditionally too.

Also, apologies in advance because we're likely going to merge https://github.com/decred/dcrdex/pull/1051 first since we keep making @buck54321 update it.  When that goes in, you'll have to resolve the differences with the `confirmRegistrationForm` template (and any other conflicts), define any new strings in client/webserver/locales/en-us.go, then do the `go generate` command in `client/webserver/site` to rebuild the localized_html files.  This is going to be part of the site workflow going forward, so we might as well trial it with you after #1051 goes in.

martonp: All good, no problem.

chappjc:
With PR #1051 merged, you'll need to update to with strings in client/webserver/locales/en-us.go and the new `[[[ adf ]]]` template notation, then rebuilid the localized_html with a `go generate` in client/webserver/site.  The conflicts in `confirmRegistrationForm` are relatively minor though.  Thanks, @martonp!

martonp:
I'm currently hiding the password field if the user selected the `Remember my password` checkbox, but not using the password cache in the register workflow. I think it's better this way because the submit button on this form causes them to spend money. Let me know if you guys agree though.

chappjc:
Makes sense to me to always require pass confirmation for register for the reasons you stated, plus it's a relatively rare event.

chappjc:
Hmm, not working anymore:

image

(rebased on master, no conflicts)

chappjc:
Geeze, nevermind that, it's alright just my tempates out of date.

Looks good on my simnet harness:

image

client/webserver/site/src/js/register.js
```diff
@@ -26,10 +24,10 @@ export default class RegistrationPage extends BasePage {
       // Form 3: Unlock Decred wallet
       'unlockWalletForm',
       // Form 4: Configure DEX server
-      'dexAddrForm',
+      'dexAddrForm', 'dexAddr',
       // Form 5: Confirm DEX registration and pay fee
       'confirmRegForm', 'feeDisplay', 'dcrBaseMarketName', 'dexDCRLotSize', 'appPass', 'submitConfirm', 'regErr',
```
chappjc, line 29: Can clean up some of these.  `dexDCRLotSize` is gone from markets.tmpl now, maybe others.

PR #1138, opened 2021-07-26
client/asset/eth: Use dex.Logger in internal node.

JoeGruffins:
Use dex's logger in order to more easily manage log messages from the
    internal geth node.

### DIFF
--- a/client/asset/eth/node.go
+++ b/client/asset/eth/node.go
@@ -46,50 +46,80 @@ type ethLogger struct {
 func (el *ethLogger) New(ctx ...interface{}) log.Logger {
 	s := ""
 	for _, v := range ctx {
-		s = fmt.Sprintf("%s:%s", s, v)
+		if s == "" {
+			s = fmt.Sprintf("%s", v)
+			continue
+		}
+		s = fmt.Sprintf("%s: %s", s, v)
 	}
 	l := el.dl.SubLogger(s)
 	return &ethLogger{dl: l}
 }
 
+// dummyHandler is used to return from the logger's GetHandler method in order
+// to avoid null pointer errors in case geth ever uses that function.
+type dummyHandler struct{}
+
+// Log does nothing and return nil.
+func (dummyHandler) Log(r *log.Record) error {
+	return nil
+}
+
+// Check that *dummyHandler satisfies the log.Handler interface.
+var _ log.Handler = (*dummyHandler)(nil)
+
 // GetHandler gets the handler associated with the logger. Unused in geth. Does
 // nothing and returns a nil interface here.
 func (el *ethLogger) GetHandler() log.Handler {
-	return nil
+	return new(dummyHandler)
 }
 
 // SetHandler updates the logger to write records to the specified handler.
 // Used during setup in geth when a logger is not supplied. Does nothing here.
 func (el *ethLogger) SetHandler(h log.Handler) {}
 
+func formatEthLog(msg string, ctx ...interface{}) []interface{} {
+	msgs := []interface{}{msg, "         "}
+	alternator := 0
+	for _, v := range ctx {
+		deliminator := "="
+		if alternator == 0 {
+			deliminator = " "
+		}
+		alternator = 1 - alternator
+		msgs = append(msgs, deliminator, v)
+	}
+	return msgs
+}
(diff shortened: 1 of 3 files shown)


client/asset/eth/node.go
```diff
@@ -32,8 +33,68 @@ var simnetGenesis string
...
+	return nil
+}
+
+// SetHandler updates the logger to write records to the specified handler.
+// Used during setup in geth when a logger is not supplied. Does nothing here.
+func (el *ethLogger) SetHandler(h log.Handler) {}
+
+// Trace logs at Trace level.
+func (el *ethLogger) Trace(msg string, ctx ...interface{}) {
+	el.dl.Trace(append([]interface{}{msg}, ctx...))
```
chappjc, line 67: Why not `el.dl.Tracef(msg, ctx...)`?

client/asset/eth/node.go
```diff
@@ -32,8 +33,68 @@ var simnetGenesis string
...
+		s = fmt.Sprintf("%s:%s", s, v)
+	}
+	l := el.dl.SubLogger(s)
+	return &ethLogger{dl: l}
+}
+
+// GetHandler gets the handler associated with the logger. Unused in geth. Does
+// nothing and returns a nil interface here.
+func (el *ethLogger) GetHandler() log.Handler {
+	return nil
```
chappjc, line 58: geth is always cool with `nil`?  Don't need a no-op function like `func(*Record) error { return nil }`?

client/asset/eth/node.go
chappjc: Oh, I just looked at the go-ethereum code and I see that `ctx` is supposed to be key/value pairs, with the entirety of the formatted message in `mgs`.  e.g. `log.Debug("Failed to decode keystore key", "path", path, "err", err)`  How should that display?

client/asset/eth/node.go
JoeGruffins: That would probably bet better. It is totally unused in the code at present, but that could change.

client/asset/eth/node.go
JoeGruffins: How would you like it to display? lol. I can't really see any different using this and the default logger. will get a comparison here in a sec...

chappjc: Yeah, it's totally up to us.  I'm just wondering how geth (the app) would print that out, so I can understand what's usually in these key-value pairs in `ctx`.  I grabbed that line randomly from the go-ethereum code base.  Seems like the intention is to show something like:

```
[DBG] GETH: Failed to decode keystore key. Details: path = /mnt/blah/key.db, err = "oh darn"
```

JoeGruffins: them
```
DEBUG[07-27|15:20:52.909] Removing p2p peer                        peercount=2 id=bad85b887d6606b2 duration="355.116µs" req=false err="Key missing from list - recentTxLookup"
```
us
```
2021-07-26 17:53:35.039 [TRC] ETHTEST[NODE]2021-07-26 17:53:35.039 [DBG] ETHTEST[NODE]: [Removing p2p peer peercount 2 id bad85b887d6606b246bb493caa638499f2f0576a7c2ed84fc72706b3de7ddceb duration 446.036µs req false err Key missing from list - recentTxLookup]
```
I'll make it more like theirs.

JoeGruffins: at latest commit:
```
2021-07-27 16:33:56.327 [DBG] ETHTEST[NODE]: Removing p2p peer             peercount = 3   id = a150cfb3940507cc8b340fb6b647855bf674ee82afec58f0c8426a5f354d2107   duration = 568.345µs   req = false   err = Key missing from list - recentTxLookup
```

client/asset/eth/node.go
JoeGruffins: Added the noop type. Was I able to do this with an anonymous function?

chappjc: dummyHandler looks nice

client/asset/eth/node.go
chappjc: Works for me.  A little space-heavy, but OK.

chappjc (approved):

PR #1141, opened 2021-07-30
client: remove obsolete auto account disable

### LINKED ISSUE #1116: client: remove obsolete auto account disable
There is a piece of legacy code in `client/Core` that disables the account in the local DB if the account was not found server-side, but this is now both obsolete and problematic. At one point I had my account toasted because of a postgresql connectivity glitch on the server that would have resolved itself had client not auto-disabled the account permanently in its db.  The resolution was to manually restore a dexc.db (client db) from backup, which I was fortunate to have.

https://github.com/decred/dcrdex/blob/fcd3926fe989d987d18e873ff604f4a754af4ce4/client/core/core.go#L2903-L2905

This logic isn't needed anymore because accounts can now be disabled manually by the client on the /settings page.  It was also a poor approach since it was not easily reversible by the client except by restoring a backup db file.

amass01:
Closes #1116.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -3431,31 +3431,6 @@ func (c *Core) MaxSell(host string, base, quote uint32) (*MaxOrderEstimate, erro
 // account is disabled and the corresponding entry in c.conns is removed
 // which will result in the user being prompted to register again.
 func (c *Core) initializeDEXConnections(crypter encrypt.Crypter) []*DEXBrief {
-	// Connections will be attempted in parallel, so we'll need to protect the
-	// errorSet.
-	var reconcileConnectionsWg sync.WaitGroup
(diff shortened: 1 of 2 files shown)


chappjc (changes requested):
I can confirm that this works as expected.

`[ERR] CORE: notify: |ERROR| (dex_auth) DEX auth error - 127.0.0.1:17273: 'connect' error: rpc error: error code 50: no account found for account ID: c4e45c5072d06e5ce4399f4ba85352158dcd384563fcc4b1ba6c43210f543177`

The DEX Account shows in the settings page, where I was able to manually disable it.

This PR just needs to have `TestInitializeDEXConnectionsAccountNotFoundError` removed and it is good to go.

chappjc (approved):

PR #1142, opened 2021-07-31
ui: restore vscroll to buy/sell orders tables

### LINKED ISSUE #1091: Create a vscroll for Buy/Sell Orders tables
As mentioned in issue #945 :

In addition to the bins, the Buy/Sell Orders tables need to be limited in height with vscroll of their own. Otherwise the Your Orders tables that goes below them in narrower windows leaves the visible part of the screen. This only happens when one of the Buy/Sell Orders tables is so full that they extend down off the screen.

martonp:
Closes #1091

### DIFF
--- a/client/webserver/site/src/css/market.scss
+++ b/client/webserver/site/src/css/market.scss
@@ -56,6 +56,34 @@ div.popupforms {
   z-index: 100;
 }
 
+div.market-orders-table-container {
+  @include media-breakpoint-down(md) {
+    max-height: 200px;
+  }
+
+  @include media-breakpoint-up(lg) {
+    position: absolute;
+    left: 0;
+    top: 0;
+    bottom: 0;
+    right: 0;
+  }
+}
+
+div.user-orders-table-container {
+  @include media-breakpoint-down(xl) {
+    max-height: 200px;
+  }
+
+  @include media-breakpoint-up(xxl) {
+    position: absolute;
+    left: 0;
+    top: 0;
+    bottom: 0;
+    right: 0;
+  }
+}
+
 table.ordertable {
   border-collapse: collapse;
   border: none;
@@ -345,10 +373,6 @@ table.balance-table button:hover {
 #orders {
   .ico-unlocked {
     color: #1e7d11;
-  }
-
-  .userorders {
-    min-height: 200px;
   }
 }
 
--- a/client/webserver/site/.stylelintrc
+++ b/client/webserver/site/.stylelintrc
@@ -1,3 +1,6 @@
 {
-    "extends": "stylelint-config-standard"
+    "extends": "stylelint-config-standard",
+    "rules": {
+        "at-rule-no-unknown": null
+    }
 }
--- a/client/webserver/site/src/css/main.scss
+++ b/client/webserver/site/src/css/main.scss
@@ -52,6 +52,10 @@ input:invalid {
   overflow-y: auto;
 }
 
+.overflow-x-hidden {
+  overflow-x: hidden;
+}
+
 header.maintop {
   width: 100%;
   height: 60px;
--- a/client/webserver/site/src/html/markets.tmpl
+++ b/client/webserver/site/src/html/markets.tmpl
@@ -37,7 +37,7 @@
   </div>
 
   {{- /* RIGHT COLUMN */ -}}
-  <div id="rightSide" class="d-flex flex-column flex-grow-1 h-100 overflow-y-auto">
+  <div id="rightSide" class="d-flex flex-column flex-grow-1 h-100 overflow-y-auto overflow-x-hidden">
 
     {{- /* CHART */ -}}
 
@@ -73,33 +73,35 @@
         <div class="row m-0 h-100">
 
           {{- /* BUY ORDER LIST */ -}}
-          <div class="marketorders order-lg-1 order-2 col-lg-7 p-0 bg1">
+          <div class="marketorders order-lg-1 order-2 col-lg-7 p-0 bg1 d-flex flex-column">
             <div class="market-header text-center fs14 py-1">[[[Buy Orders]]]</div>
-            <div class="max-h-100 stylish-overflow">
-              <table class="ordertable">
-                {{block "thead" .}}
-                <thead>
-                  <tr>
-                    <th class="text-left pl-2">[[[Quantity]]]</th>
-                    <th class="text-right pr-2">[[[Rate]]]</th>
-                    <th class="text-right pr-2">[[[Epoch]]]</th>
-                  </tr>
-                </thead>
-                {{end}}
-                <tbody id="buyRows">
-                  {{- /* This row is used by the app as a template. */ -}}
-                  <tr id="rowTemplate">
-                    <td class="pl-2">
-                      <div class="d-flex justify-content-start align-items-center">
-                        <div data-tmpl="qty">-</div>
-                        <div class="numorders text-center ml-2" data-tmpl="numorders">-</div>
-                      </div>
-                    </td>
-                    <td class="text-right pr-2" data-tmpl="rate">-</td>
-                    <td class="text-right pr-3" data-tmpl="epoch"></td>
-                  </tr>
-                </tbody>
-              </table>
+            <div class="brdrleft flex-grow-1 bg1 position-relative">
+              <div class="market-orders-table-container stylish-overflow">
+                <table class="ordertable">
+                  {{block "thead" .}}
+                  <thead>
+                    <tr>
+                      <th class="text-left pl-2">[[[Quantity]]]</th>
+                      <th class="text-right pr-2">[[[Rate]]]</th>
+                      <th class="text-right pr-2">[[[Epoch]]]</th>
+                    </tr>
+                  </thead>
+                  {{end}}
+                  <tbody id="buyRows">
+                    {{- /* This row is used by the app as a template. */ -}}
+                    <tr id="rowTemplate">
+                      <td class="pl-2">
+                        <div class="d-flex justify-content-start align-items-center">
+                          <div data-tmpl="qty">-</div>
+                          <div class="numorders text-center ml-2" data-tmpl="numorders">-</div>
+                        </div>
+                      </td>
+                      <td class="text-right pr-2" data-tmpl="rate">-</td>
+                      <td class="text-right pr-3" data-tmpl="epoch"></td>
+                    </tr>
+                  </tbody>
+                </table>
+              </div>
             </div>
           </div>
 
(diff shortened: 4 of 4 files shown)


chappjc:
I'm seeing some side-effects:

image

The `ordertable`s and the "Your Orders" have both light gray horizontal and vertical bars that weren't there before.

I can't seem to Inspect them either. The above is Firefox.  It's even crazier in Chromium:

image


BTW, there are other pre-existing scroll quirks if you start playing around.  One is the main page's horizontal scroll that is there on load but will disappear if you change the window height.

martonp:
@chappjc Wow.. I don't see any of those issues when I try on either firefox or brave. Can you try to clear your cache or run in a private tab? Also, are you using windows? Maybe  it displays differently on the windows versions of the browsers.

chappjc:
I hard-refreshed both, but will try again. Linux, Firefox developer edition (91b9) and Chromium 92.  Will try with 
Epiphany too.

BTW, one thing we like to do with frontend change is to bump the url queries on style.csc and entry.js in bodybuilder.tmpl.  In a perfect world we'd only do that on release, but it tends to solve cache issues during dev.

chappjc:
I don't see the problem on Epiphany (essentially Safari), but it's definitely there on FF and Chromium

chappjc:
Looks like we should use `auto` for `overflow-y`
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow

chappjc:
Actually, I noticed that the style of the scrollbar is quite inconsistent between browsers too, so think we should use our `.stylish-overflow` class.

client/webserver/site/src/css/market.scss
```diff
@@ -63,24 +63,33 @@ table.ordertable {
...
+  tr {
+    display: table;
+    width: 100%;
+    table-layout: fixed;
+  }
+
+  tbody {
+    display: block;
+    width: 100%;
+    overflow: scroll;
```
chappjc, line 89: Although I had suggested `overflow: auto`, I think we should just not set overflow here at all since the parent elements already are `stylish-overflow` class, which makes a scroll bar with consistent appearance across browsers.

client/webserver/site/src/css/market.scss
```diff
@@ -63,24 +63,33 @@ table.ordertable {
...
+    table-layout: fixed;
```
chappjc, line 90: Can you also review the changes in the various responsive views, particularly the most narrow where all the tables and the order form are on top of each other?  In that view it seems like the tables are reserving more vertical space than before, but I could be mistaken.

martonp: Before they just took up as much space as the rows in the table took, but I'll make them take up the same amount of space as the "Your Orders" table.

chappjc: "Before" was ok right?  Now when I'm in the narrow view they take a ton of space for nothing.  Not for you?

martonp:
@chappjc I've enabled scrollbars on mac and everything looks pretty good now. Let me know what you see.

client/webserver/site/src/css/market.scss
martonp: Check it out now, see how you like it. It's the same smaller space in every case.

client/webserver/site/src/js/markets.js
```diff
@@ -246,6 +246,31 @@ export default class MarketsPage extends BasePage {
...
+        // We subtract 10 pixels because if the markets tables becomes taller than the order
+        // panel, then the order panel will resize based on the market tables, and they will
+        // all keep getting taller.
+        marketTableHeight =
+          page.orderPanel.offsetHeight -
+          page.buysTableHeader.offsetHeight -
+          page.buysHeader.offsetHeight -
+          10
+      }
+      page.sellRows.style.height = page.buyRows.style.height = `${marketTableHeight}px`
```
chappjc, line 269: I'd expect the table heights would be something for the css and bootstrap responsive design classes to handle automatically, rather than computing height and ending up with something like `<tbody id="buyRows" class="stylish-overflow" style="height: 519px;">`, but I don't really have the solution.  Perhaps @buck54321 or another ui dev with bootstrap knowledge can make a suggestion.

The `largeScreenCutoffWidth = 992` matches a responsive cutoff (`$grid-breakpoints` in application.scss), so it feels like that should come out of the css somehow or just be used by the responsive and flex classes.  

The computed height still seems to be too high in a lot of cases.  e.g. a simnet layout with this PR:

image

The two order tables are making it higher than necessary above.

A similar layout on testnet without these changes:

image

Note the order form not being taller than necessary.

client/webserver/site/src/js/markets.js
```diff
@@ -246,6 +246,31 @@ export default class MarketsPage extends BasePage {
...
 
+    // Resize elements on screen when the page is resized
+    this.resize = () => {
+      let marketTableHeight
+      const largeScreenCutoffWidth = 992
+      if (window.innerWidth < largeScreenCutoffWidth) {
+        // In smaller screens, the order panel and the market tables are stacked
+        // on top of each other, and we want the market tables to have the same height
+        // as the live orders table.
+        marketTableHeight = page.liveTableContainer.offsetHeight - page.liveTableHead.offsetHeight
```
chappjc, line 257: The `orders-left col-24` div seems alright when narrow, but it does seem like the computation over allocates height a bit.  See the sell orders table below:

image

Compared with a testnet setup (no vscroll, but also no more height than needed):

image

chappjc (commented):
I'm pretty uncertain about the use of JS to compute heights here.  It's also appearing to set heights sub-optimally.  Perhaps other devs can make a suggestion regarding responsive design and flex to achieve this more simply with just bootstrap classes.

martonp:
@chappjc I found a much nicer way to do this without JS. I'll still review this a bit later, but I wanted to check it in so no one else wastes time on it.

martonp:
This is ready for another look.

buck54321:
The thing is, we already have `stylish-overflow` applied to those `div`s, and it used to work.

https://github.com/decred/dcrdex/blob/1640b7021f70efae83c18795d89ad3700c04cc68/client/webserver/site/src/html/markets.tmpl#L78

We need to figure out why it stopped working. It was almost certainly a change from #1016.

chappjc:
That's what I realized as well.  https://github.com/decred/dcrdex/pull/1142#discussion_r681298558
I think the responsive design changes lacked a max-height, but not sure.

martonp:
What I currently have addresses the issues that were created in the responsive design. The two order tables and the order panel are all in a flex row (`orders-left`), and the height of the row depends on the largest column in the row. As more rows are added to buy/sell order tables, the entire row expands. Instead, what we want is the buy/sell order tables to depend on the height of the order panel. This is the same relationship the `userorders` div has to `orders-left`. 

The `userorders` div has the `fill-abs` class and a minimum height, so on large screens its height depends on the height of `orders-left` and on smaller screens it has the minimum height. What I have now for the buy/sell tables is the same, except on large screens it has the same properties as `fill-abs` and on smaller screens it just has a max-height. The benefit of this is that it avoids additional space when there are very few rows (this shouldn't matter in practice).

I will update `userorders` to use a max height instead of a min height as well.

JoeGruffins (approved):
Tried all the sizes and configurations I could think of on firefox and looks fine.

chappjc:
> What I currently have addresses the issues that were created in the responsive design. The two order tables and the order panel are all in a flex row (`orders-left`), and the height of the row depends on the largest column in the row. As more rows are added to buy/sell order tables, the entire row expands. Instead, what we want is the buy/sell order tables to depend on the height of the order panel. This is the same relationship the `userorders` div has to `orders-left`.
> 
> The `userorders` div has the `fill-abs` class and a minimum height, so on large screens its height depends on the height of `orders-left` and on smaller screens it has the minimum height. What I have now for the buy/sell tables is the same, except on large screens it has the same properties as `fill-abs` and on smaller screens it just has a max-height. The benefit of this is that it avoids additional space when there are very few rows (this shouldn't matter in practice).

^^ Thanks for this.    Going forward, this kind of description is desirable for the git commit messages, for the purposes of assisting review by describing the root cause of the issue and the approach used to fix it, and to express the intent of the code changes for posterity.

martonp:
@chappjc What do you mean? Should I copy this to the description of the PR?

chappjc:
I'm referring to the git commit messages.  So instead of a one sentence like "Add vscroll to buy/sell order tables." or "Update.", you'd describe the change like you did well in that github comment.  No need to update anything now though, just going forward.  When I merge the PR I will squash down and describe the change in terms of what was done to restore the vscroll, why it was broken, and any other improvements.

client/webserver/site/src/css/market.scss
```diff
@@ -56,6 +56,34 @@ div.popupforms {
...
+div.market-orders-table-container {
+  @media (min-width: 992px) {
+    position: absolute;
+    left: 0;
+    top: 0;
+    bottom: 0;
+    right: 0;
+  }
+
+  @media (max-width: 991px) {
```
chappjc, line 68: I feel like there should be a way to use the break points defined in application.scss, no?  Like `media-breakpoint-down(md)` and `media-breakpoint-up(lg)`

https://github.com/decred/dcrdex/blob/90b1d9289eb55e74d59840f3a856d44a313559df/client/webserver/site/src/css/application.scss#L17-L21

If doing so makes things more complex, then this is fine, but it seems a bit fragile because we risk changing one but not the other.

martonp: With `@media-breakpoint-down` and `@media-breakpoint-up` that are different from what we are using, but what I have now has the same effect.

client/webserver/site/src/css/market.scss
```diff
@@ -57,31 +59,35 @@ div.popupforms {
 }
 
 div.market-orders-table-container {
-  @media (min-width: 992px) {
+  $lg: map.get($grid-breakpoints, "lg");
+
+  @media (max-width: calc(#{$lg} - 1px)) {
```
chappjc, line 64: Bummer the mixing wasn't quite right.  I'm still uncertain why it wouldn't fit the bill here though.  Basically the "docs" at https://mdbootstrap.com/docs/standard/layout/breakpoints/ and the code in node_modules/bootstrap/scss/mixins/_breakpoints.scss indicate it's doing basically what you're doing except with a `- 0.02` instead of `- 1` for the `media-breakpoint-down` impl, which does still sound appropriate.

martonp: The `$grid-breakpoints` variables that we are using in `application.scss` is different from the one in bootstrap. The `xxl` size we have is 1750px instead of 1400px. The `lg` size is the same but I think it's better to be consistent.

chappjc: Ok, I see your concern now.  But the map we define in application.scss seems to override the one in bootstrap's _variables.scss.  This seems to be on account of the use of the SASS `!default` flag (https://sass-lang.com/documentation/variables#default-values)

With the mixin:

image

image

chappjc: Er, bootstrap link: https://getbootstrap.com/docs/4.3/getting-started/theming/#variable-defaults

martonp: The issue was that I was using the same size in both the up and down case, not that it was using the incorrect variable. It's fixed now.

chappjc (approved):

client/webserver/site/src/css/market.scss
chappjc: Just making a note that when we update to Bootstrap 5, there's a breaking change we might not catch:

> Media query mixins parameters have changed for a more logical approach.
> 
>   -  media-breakpoint-down() uses the breakpoint itself instead of the next breakpoint (e.g., media-breakpoint-down(lg) instead of media-breakpoint-down(md) targets viewports smaller than lg).
>   - Similarly, the second parameter in media-breakpoint-between() also uses the breakpoint itself instead of the next breakpoint (e.g., media-between(sm, lg) instead of media-breakpoint-between(sm, md) targets viewports between sm and lg).
>

PR #1149, opened 2021-08-04
rpc: Add appseed endpoint.

### LINKED ISSUE #1147: rpcserver: need endpoint for dumping app seed
As mentioned [here](https://github.com/decred/dcrdex/pull/1015#pullrequestreview-719262759), we need an client RPC endpoint for dumping the application seed.

JoeGruffins:
appseed is able to return the app's seed as hex if supplied the correct
application password.

While here, fix error checking of type tests to not ignore expected
errors when no error happens.

Bump the rpcserver minor version to four because dexcctl was already at
three for some reason, and this adds a feature that is backwards
compatible because it is just a new endpoint.

closes #1147

### DIFF
--- a/client/cmd/dexcctl/main.go
+++ b/client/cmd/dexcctl/main.go
@@ -26,7 +26,11 @@ const (
 	listCmdMessage  = "Specify -l to list available commands"
 )
 
-var version = semver{major: 0, minor: 3, patch: 0}
+// version is the dexcctl version and should correspond to the rpcclient's
+// version. rpcclients with a higher minor are newer with newer features while
+// a higher major indicates breaking changes and dexcctl should be updated
+// before attempting to communicate.
+var version = semver{major: 0, minor: 4, patch: 0}
 
 // semver holds dexcctl's semver values.
 type semver struct {
@@ -60,6 +64,7 @@ var promptPasswords = map[string][]string{
 	"register":   {"App password:"},
 	"trade":      {"App password:"},
 	"withdraw":   {"App password:"},
+	"appseed":    {"App password:"},
 }
 
 // optionalTextFiles is a map of routes to arg index for routes that should read
--- a/client/cmd/dexcctl/main.go
+++ b/client/cmd/dexcctl/main.go
@@ -26,11 +26,8 @@ const (
 	listCmdMessage  = "Specify -l to list available commands"
 )
 
-// version is the dexcctl version and should correspond to the rpcclient's
-// version. rpcclients with a higher minor are newer with newer features while
-// a higher major indicates breaking changes and dexcctl should be updated
-// before attempting to communicate.
-var version = semver{major: 0, minor: 4, patch: 0}
+// version is the dex server's release version.
+var version = semver{major: 0, minor: 3, patch: 0}
 
 // semver holds dexcctl's semver values.
 type semver struct {
--- a/client/rpcserver/rpcserver.go
+++ b/client/rpcserver/rpcserver.go
@@ -36,9 +36,12 @@ const (
 	// is closed.
 	rpcTimeoutSeconds = 10
 
-	// RPC version
+	// RPC version. Move major up one for breaking changes. Move minor for
+	// backwards compatible features. Move patch for bug fixes. It is
+	// intended that this always match dexcctl's version, so also update
+	// there accordingly.
 	rpcSemverMajor = 0
-	rpcSemverMinor = 0
+	rpcSemverMinor = 4
 	rpcSemverPatch = 0
 )
 
@@ -69,6 +72,7 @@ type clientCore interface {
 	Wallets() (walletsStates []*core.WalletState)
 	WalletState(assetID uint32) *core.WalletState
 	Withdraw(appPass []byte, assetID uint32, value uint64, addr string) (asset.Coin, error)
+	ExportSeed(pw []byte) ([]byte, error)
 }
 
 // RPCServer is a single-client http and websocket server enabling a JSON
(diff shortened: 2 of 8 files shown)


client/rpcserver/rpcserver.go
```diff
@@ -36,9 +36,12 @@ const (
...
-	// RPC version
+	// RPC version. Move major up one for breaking changes. Move minor for
+	// backwards compatible features. Move patch for bug fixes. It is
+	// intended that this always match dexcctl's version, so also update
+	// there accordingly.
 	rpcSemverMajor = 0
-	rpcSemverMinor = 0
+	rpcSemverMinor = 4
 	rpcSemverPatch = 0
 )
```
JoeGruffins, line 46: Am I using these wrong? I guess it should be 4 if dexcctl was at 3.

chappjc: dexcctl was for the product like v0.3.  I think these `rpcSemver*` vars represent the RPC protocol version.  I think we should bump these RPC versions independently of product, like the dcrd/dcrwallet RPC versions do.  So, we can bump this to `rpcSemverMinor = 1` I think.

client/cmd/dexcctl/main.go
```diff
@@ -26,7 +26,11 @@ const (
 	listCmdMessage  = "Specify -l to list available commands"
 )
 
-var version = semver{major: 0, minor: 3, patch: 0}
+// version is the dexcctl version and should correspond to the rpcclient's
+// version. rpcclients with a higher minor are newer with newer features while
+// a higher major indicates breaking changes and dexcctl should be updated
+// before attempting to communicate.
+var version = semver{major: 0, minor: 4, patch: 0}
```
chappjc, line 33: I'm not sure about this.  I was following the is the app product version (dexcctl v0.3 for the upcoming release), while the RPC protocol version is the rpcSemverMinor stuff.

client/rpcserver/rpcserver.go
chappjc: What I mean is that we probably should be at `rpcSemverMinor = 20` by now, but we haven't bothered.  (Is even there any communication of the version to RPC clients to make this useful yet?)

client/cmd/dexcctl/main.go
JoeGruffins: That makes sense.

But how will one know this dexcctl will work with the dexc rpcserver? I guess it's reasonable to just require the same release version, although there isn't a way to check from the rpcserver...

client/rpcserver/rpcserver.go
JoeGruffins: I will set it to minor 1.

client/cmd/dexcctl/main.go
JoeGruffins: Maybe I'll try to find a solution in another pr, if it is a problem.

chappjc: It's the same with dcrctl.  It doesn't check protocol version, it just tries cause it's a single shot http post.

> I guess it's reasonable to just require the same release version, although there isn't a way to check from dexcctl...

I don't follow the above.  What can't be checked?

JoeGruffins: The dexc version. dexcctl could be build from a different commit than dexc. So you don't know what version dexc is from rpc commands. There is the rpc version, but there's nothing atm for the user to know from that. So I have version 0.1.0, but I don't know what version dexcctl supports.

chappjc: Gotcha.  That's basically the situation when someone uses say rpcclient to connect to either dcrd or dcrwallet - they have t code what versions are acceptable.

But with dcrctl and dexcctl, being a single shot http post, I'm not sure how to use that info internally (request version prior to every user request?) or if we'd even want to restrict when things might work even out of some predefined range.

JoeGruffins: Since an old dexcctl may work with a new dexc, I would like to put like, a supported dexc rpc version return to dexcctl -V. And this would return 0.1.0 atm because that`s what the rpcserver version is at. And a user would know it supports up to rpc version "0.##.##" Assuming it has users.

chappjc: Makes sense.  That could happen by either introducing supported proto version consts in the dexcttl main package or exporting the ones from client/rpcserver?  And then beyond that it's the user's responsibility to use an rpc server with compatible version, as whatever they deem to be compatible.

JoeGruffins: There have been many times I've used dcrctl and been met with error that left me confused until I realized I needed to update dcrctl.

JoeGruffins: > But with dcrctl and dexcctl, being a single shot http post, I'm not sure how to use that info internally (request version prior to every user request?)

I think I tried to sneak this in when I fist made the thing, but I would be happy with just a way to check manually, like with an extra -V version.

chappjc: > The onus is on the user but they do have a way to check before getting an error.

Agree with this.  I'd rather not cripple dexcctl by being over-restrictive.  A major version bump might actually only affect one RPC, and the tool would otherwise work.

JoeGruffins: #1151

chappjc (approved):
Looks good, works great.

Simnet:

```
$ ./dexcctl --simnet -p a appseed
5a1cac6269b7ac88a75699c669be59509dd45eeac03593bd74348efedada7302adf247f734f912db3ab5a6dc61011f9a4a2b2bac3b03fb0137943193694b7d62

$ ./dexcctl -V
dexcctl version 0.3.0 (Go version go1.16.6 linux/amd64)

$ ./dexcctl --simnet version
0.1.0

```

image

PR #1150, opened 2021-08-04
server: address RPC hangs with disconnected RPC server

chappjc:
When using rpcclient with dcrd with a WebSocket connection and its auto-reconnect feature, RPCs will hang until the client reconnects or the `Context` provided to them is cancelled.  `asset.Backend` methods sometimes accept a `Context` from consumers, and sometimes we use a `Backend`'s built-in "RPC context" to cancel RPCs when the `Backend` subsystem stops.  Unfortunately, because subsystems that consume `Backends` (e.g. `Market`) are generally shutdown *before* the `Backend`s, their shutdown can hang on RPCs.  This suggests that we really do have to have every `Backend` method that performs an RPC accept a `Context`.  (Shutting down the `Backend`s first is not desirable.)  This also may suggest we should not use auto-reconnect.

I'm not prepared to take that leap to contextifying everything yet, so this PR trials two approaches with the heavily used `FeeRate` and `Synced` methods:
- Give `FeeRate` a context arg, where consumer decided timeouts are reasonable. Unfortunately, a `Backend` can't cancel RPCs on shutdown now, but stopping the `rpcclient.Client` _should_ have the same effect if it  happens to be shutdown before consumers.
- Try a different solution for the `Synced` method: add an internal 2 second timeout. This minimizes API changes and continues to allow RPC cancellation by the `Backend` subsystem shutdown, which using an external `Context` precludes with out serious acrobatics and extra goroutine overhead.

Together with dcrd PR https://github.com/decred/dcrd/pull/2696, the dcrdex process no longer hangs when trying to shutdown with an rpcclient in it's reconnect loop.  In draft until that is merged and the rpcclient/v7 require is updated.

Modify `server/dex.feeFetcher` with the context arg, and add a `LastRate` method for consumers to use as a fallback if `FeeRate` fails. Use the `LastRate` fallback in `market.(*Market).getFeeRate`.

Also fix a `feeFetcher` issue where a zero rate would be cached on `FeeRate` error.

Example of changes when rpcclient's reconnect loop starts (note the new logging subsystem, FeeRate timeouts, and LastRate fallback):

```
2021-08-04 15:24:08.911 [ERR] ASSET[dcr][RPC]: Websocket receive error from 127.0.0.1:19570: websocket: close 1006 (abnormal closure): unexpected EOF
2021-08-04 15:24:08.911 [INF] ASSET[dcr][RPC]: Failed to connect to 127.0.0.1:19570: dial tcp 127.0.0.1:19570: connect: connection refused           
2021-08-04 15:24:08.911 [INF] ASSET[dcr][RPC]: Retrying connection to 127.0.0.1:19570 in 5s                                                          
2021-08-04 15:24:13.912 [INF] ASSET[dcr][RPC]: Failed to connect to 127.0.0.1:19570: dial tcp 127.0.0.1:19570: connect: connection refused           
2021-08-04 15:24:13.912 [INF] ASSET[dcr][RPC]: Retrying connection to 127.0.0.1:19570 in 10s                                                         
2021-08-04 15:24:16.001 [ERR] DEX: Error retrieving fee rate for dcr: request timeout                                                                
2021-08-04 15:24:16.001 [WRN] MKT: Failed to get latest fee rate for dcr. Using last known rate 10.                                                  
2021-08-04 15:24:16.001 [ERR] DEX: Error retrieving fee rate for dcr: request timeout                                                                
2021-08-04 15:24:16.001 [WRN] MKT: Failed to get latest fee rate for dcr. Using last known rate 10.                                                  
```

Previously it would hang *in the epoch processing* pipeline forever.

### DIFF
--- a/server/asset/btc/btc.go
+++ b/server/asset/btc/btc.go
@@ -381,7 +381,7 @@ func (btc *Backend) InitTxSizeBase() uint32 {
 }
 
 // FeeRate returns the current optimal fee rate in sat / byte.
-func (btc *Backend) FeeRate() (uint64, error) {
+func (btc *Backend) FeeRate(_ context.Context) (uint64, error) {
 	return btc.estimateFee(btc.node)
 }
 
--- a/server/asset/dcr/dcr.go
+++ b/server/asset/dcr/dcr.go
@@ -36,6 +36,13 @@ type Driver struct{}
 
 // Setup creates the DCR backend. Start the backend with its Run method.
 func (d *Driver) Setup(configPath string, logger dex.Logger, network dex.Network) (asset.Backend, error) {
+	// With a websocket RPC client with auto-reconnect, setup a logging
+	// subsystem for the rpcclient.
+	logger = logger.SubLogger("RPC")
+	if logger.Level() == dex.LevelTrace {
+		logger.SetLevel(dex.LevelDebug)
+	}
+	rpcclient.UseLogger(logger)
 	return NewBackend(configPath, logger, network)
 }
 
--- a/server/asset/dcr/dcr.go
+++ b/server/asset/dcr/dcr.go
@@ -222,7 +222,7 @@ func (dcr *Backend) Connect(ctx context.Context) (*sync.WaitGroup, error) {
 		}
 	}
 
-	if _, err = dcr.FeeRate(); err != nil {
+	if _, err = dcr.FeeRate(ctx); err != nil {
 		dcr.log.Warnf("Decred backend started without fee estimation available: %v", err)
 	}
 
@@ -247,9 +247,9 @@ func (dcr *Backend) InitTxSizeBase() uint32 {
 }
 
 // FeeRate returns the current optimal fee rate in atoms / byte.
-func (dcr *Backend) FeeRate() (uint64, error) {
+func (dcr *Backend) FeeRate(ctx context.Context) (uint64, error) {
 	// estimatesmartfee 1 returns extremely high rates on DCR.
-	estimateFeeResult, err := dcr.node.EstimateSmartFee(dcr.ctx, 2, chainjson.EstimateSmartFeeConservative)
+	estimateFeeResult, err := dcr.node.EstimateSmartFee(ctx, 2, chainjson.EstimateSmartFeeConservative)
 	if err != nil {
 		return 0, translateRPCCancelErr(err)
 	}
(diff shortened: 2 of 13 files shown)


server/asset/btc/btc.go
```diff
@@ -381,7 +381,7 @@ func (btc *Backend) InitTxSizeBase() uint32 {
 }
 
 // FeeRate returns the current optimal fee rate in sat / byte.
-func (btc *Backend) FeeRate() (uint64, error) {
+func (btc *Backend) FeeRate(_ context.Context) (uint64, error) {
```
chappjc, line 384: Tons of refactoring needed down to the `call` method to use the context with BTC. Fortunately with http posts (not websocket with autoreconnect), requests fail immediately when connection cannot be established.

server/asset/eth/eth.go
```diff
@@ -173,8 +173,8 @@ func (eth *Backend) InitTxSizeBase() uint32 {
 }
 
 // FeeRate returns the current optimal fee rate in gwei / gas.
-func (eth *Backend) FeeRate() (uint64, error) {
-	bigGP, err := eth.node.suggestGasPrice(eth.rpcCtx)
+func (eth *Backend) FeeRate(ctx context.Context) (uint64, error) {
+	bigGP, err := eth.node.suggestGasPrice(ctx)
```
chappjc, line 177: With ETH's IPC RPCs, I'm less sure about this, but if the rule is to shutdown consumers (and cancel these passed contexts) before shutting down the `Backend`, we may not need the `rpcCtx` if we give more methods a `Context` arg.

chappjc:
The dcrd/rpcclient/v7 fix is merged (https://github.com/decred/dcrd/pull/2696), so this PR is r4r.

Open to thoughts for handling request cancellation and Backend shutdown.  I've resisted contextifying every single method, but it's starting to look like it's the only way forward.

The idea of disabling rpcclient's auto-reconnect feature with dcrd/dcrwallet websockets is also on the table.  We'd just have to setup our own reconnection handling.

JoeGruffins (approved):
This looks good and works for me.

I think it's probably best to contextify all the methods. It makes sense for the life of the method to belong to the caller, while the life of the rpcclient belongs to the server.

I can't remember why atm but I wrote a re-connecting interface a year ago or so: https://github.com/decred/dcrstakepool/blob/master/backend/stakepoold/stakepool/client.go

server/asset/dcr/dcr.go
```diff
@@ -310,7 +317,12 @@ func (dcr *Backend) ValidateSecret(secret, contract []byte) bool {
 
 // Synced is true if the blockchain is ready for action.
 func (dcr *Backend) Synced() (bool, error) {
-	chainInfo, err := dcr.node.GetBlockChainInfo(dcr.ctx)
+	// With ws autoreconnect enabled, requests hang when backend is
+	// disconnected. TODO: Plumb everything method with a Context, complicating
```
JoeGruffins, line 321: Forget to comment on this, but super not important. everything -> every

chappjc:
> This looks good and works for me.
> 
> I think it's probably best to contextify all the methods. It makes sense for the life of the method to belong to the caller, while the life of the rpcclient belongs to the server.
> 
> I can't remember why atm but I wrote a re-connecting interface a year ago or so: https://github.com/decred/dcrstakepool/blob/master/backend/stakepoold/stakepool/client.go

Ah, I recalled you did some work in this a while ago.  Thanks.

I've been very reluctant to contextify all be the things, but I'm seeing more and more the issues solved.  I still don't love that the BTC backend would be at the mercy of it's consumers to finish or cancel their requests, rather than being able to cancel them, but it's not even how we're designing the subsystem hierarchy, so... I guess that's alright

buck54321 (approved):
https://github.com/golang/go/issues/36503 would maybe be handy. Oh well. Confirmed this works.

PR #1152, opened 2021-08-05
app: new solution for the chart resize bug

buck54321:


chappjc:
The bug wasn't FF specific. The clearest repro was actually with Chromium https://github.com/decred/dcrdex/pull/1113#issuecomment-877644244.

I'll try this out on both browsers though.

The repro was just to click to wallets, then back to markets.

client/webserver/site/src/js/charts.js
```diff
@@ -83,8 +83,8 @@ export class DepthChart {
     }
 
     bind(this.canvas, 'wheel', e => { this.wheel(e) })
-    this.resize = () => this.resize_(parent.clientHeight)
-    bind(window, 'resize', this.resize)
+    this.resize = h => { this.resize_(h) }
+    bind(window, 'resize', () => { this.resize(parent.clientHeight) })
```
chappjc, line 87: I'm confused by the bug.  I think this puts it back.  `parent.clientHeight` is undefined now so until you refresh the `resize_` function isn't going to work (in the scenario where you go from markets -> wallets -> markets by clicking the menu).

chappjc: Oh wait, this is another unbinding issue.  So the console error we were seeing was for an orphaned resize arrow function that we were unable to rebind, and there was another one that was actually working?

chappjc: Hmm, that actually seemed to work:

```diff
diff --git a/client/webserver/site/src/js/charts.js b/client/webserver/site/src/js/charts.js
index 715974f5..9cd0e7e5 100644
--- a/client/webserver/site/src/js/charts.js
+++ b/client/webserver/site/src/js/charts.js
@@ -84,19 +84,19 @@ export class DepthChart {
 
     bind(this.canvas, 'wheel', e => { this.wheel(e) })
     this.resize = h => { this.resize_(h) }
-    bind(window, 'resize', () => { this.resize(parent.clientHeight) })
+    this.resizeSelf = () => { this.resize(parent.clientHeight) }
+    bind(window, 'resize', this.resizeSelf)
     bind(this.canvas, 'click', e => { this.click(e) })
     this.resize(parent.clientHeight)
   }
 
   // The market handler will call unattach when the markets page is unloaded.
   unattach () {
-    unbind(window, 'resize', this.resize)
+    unbind(window, 'resize', this.resizeSelf)
   }
 
   // resize_ is a 'resize' event handler.
   resize_ (parentHeight) {
-    if (!parentHeight) return // Fix for firefox bug. See #1113 for discussion.
     this.canvas.width = this.parent.clientWidth
     this.canvas.height = parentHeight - 20 // magic number derived from a soup of css values.
     const xLblHeight = 30
```

buck54321:
Fixed now with input from #1152.

vctt94:
this PR actually broke the resize for me

image

chappjc:
We're converging on something here, but yeah, the bound handler doesn't run this way.

buck54321:
Yeah. I'm trying to avoid wrapping our wrapper, but it's looking more like a solution.

chappjc:
Can we just rename `resize_` to be a "not private" function?

buck54321:
> Can we just rename `resize_` to be a "not private" function?

LOL. Beat ya to it. I'm updating the docs one more time though.

chappjc:
Hah, yeah that looks good! :smile:

chappjc (approved):

vctt94 (approved):
Looking good now!

chappjc:
Will merge soonish, but gonna leave up for any more eyes or if you wanna do anything differently @buck54321

PR #1153, opened 2021-08-05
ui: fix depth chart ratio change improperly resizing chart

chappjc:
Relating to https://github.com/decred/dcrdex/pull/1152/files#r683601740, this fixes the chart ratio change handler to call the `resize` function with no parent height argument because `resize` gets it from its parent (`marketChart`).  This could also call like `resize_(h)`.

### DIFF
--- a/client/webserver/site/src/js/markets.js
+++ b/client/webserver/site/src/js/markets.js
@@ -299,7 +299,7 @@ export default class MarketsPage extends BasePage {
 
       page.marketChart.style.height = `${h}px`
 
-      this.chart.resize(h)
+      this.chart.resize() // resize gets height from parent (marketChart)
     }
     const chartDivRatio = State.fetch(chartRatioKey)
     if (chartDivRatio) {


client/webserver/site/src/js/markets.js
```diff
@@ -299,7 +299,7 @@ export default class MarketsPage extends BasePage {
 
       page.marketChart.style.height = `${h}px`
 
-      this.chart.resize(h)
+      this.chart.resize() // resize gets height from parent (marketChart)
```
buck54321, line 302:
```javascript
console.log("style.height:", page.marketChart.style.height, ", clientHeight:", page.marketChart.clientHeight)
```

When you set the `style.height`, the 'clientHeight' is not affected until the next render or something. This is especially a problem on reload after the user has resized the chart. Their setting is now effectively forgotten, and the chart is the default size again.

Just two steps to repro:
1. Resize chart
2. Refresh

chappjc: Ahhh.  This also worked with `resize_(h)`.  That a solution?

buck54321: That would work, but the trailing `_` is kinda like a "private" method.

Is that better than #1152?

chappjc: #1152 leaves the orphaned event handlers bound because the unbind doesn't match the bind anymore.  If you click back and forth between pages, you end up with many bound handlers and only the latest one bound actually has a valid parent.  If you put a console.log above the `if (!parentHeight)` you'll see the count of logged zeros grows.

buck54321: Ah. Fixed that now.

chappjc:
https://github.com/decred/dcrdex/pull/1152 now unbinds the resize event handler, so closing this.

Issue #1021, opened 2021-03-22
eth: coin ID uses and encodings (discussion)


chappjc:
This is closely related to the issue "[eth: Solidity atomic swap contract discussion](https://github.com/decred/dcrdex/issues/1001)", but is intended to shed light on the more broad issue of coin IDs in eth. Also the PR https://github.com/decred/dcrdex/pull/979 where the initial coin ID encoding is being defined.

### Coin IDs in DCRDEX

In DCRDEX (client and server), the notion of a "coin ID" is used for several purposes. Currently it is always tied to a txn, for which a confirmation count can be obtained. It can be used to identify unspent value, locate a swap contract and check confirmations of the payment to that contract, and generally refer to transactions needing confirms (no particular utxo).

Question: In ETH, how are account balances handled w.r.t. the confirmations of the transactions that credit/debit the accounts? Specify block count or is it standard?

### Uses of Coin IDs

**funding coins**: used to account for spendable value

 In eth, an regular (non-contract) account with a balance. Sufficient amount must be order qty PLUS an explicitly specified tx fee amount for executing the swap txns.
 
 For server to validate, check balance of the encoded eth account. More specifically, for ALL of the user's orders using this account as a funding source, sum the required values, and compare to that account balance.

Note: presumably all orders would use the same eth account.

**swap contract coin:** both value and tx needing confirms

 In eth, a tx executing The One Swap Contract.  Would encode: tx for confirmations, input data to the txn such as secret hash (and nonce, amount, dest acct, etc?). What data would be part of the swap coin ID, and which would just be part of the messages (e.g. init/audit)?  Presently the contract itself is in message data, and that decoded provides secret hash and counterparty address/pkh, while value comes from the amount of that txn's output.
 
 Or if the answer to the confirmations question indicates we don't need to monitor the transaction executing the contract, do we just check The One contract account's state data for the secret hash (the contract's map key), etc?

Would we encode the txn executing The One contract, and an audit would mean checking that it indeed executes the expected contract, and provides the expected "input data" for the terms of the trade?

**change coin**: see funding coins

 Does NOT exist in ETH. On swap, balance of funding account is reduced by precise amount, while balance in The One contract is increased.

**redeem coins**: track completion of trade on confirms, can become funding coins

 In eth, a tx for user to track confs for receipt of the swap value.
 
 No need for client to track it as funding account balance will just go up. (Similar for UTXOs where client would just eventually list those utxos as confirmed.)

**withdrawls**: txn for user to check

[...]

PR #1156, opened 2021-08-12
eth: Add CoinID flags.

JoeGruffins:
Add TxIDFlag and SwapFlag in order to specify that a coin ID represents
    either a txid or a swap. IsTxIDCoinID and IsSwapCoinID package methods
    are included so that callers can easily discern if they have the right
    type of coin ID for the operation they are performing.

related to #1021

### DIFF
--- a/server/asset/eth/common.go
+++ b/server/asset/eth/common.go
@@ -101,8 +101,7 @@ func CoinIDToString(coinID []byte) (string, error) {
 // ToCoinID converts the address and secret hash, or txid to a coin ID.
 func ToCoinID(flags CoinIDFlag, addr *common.Address, hash []byte) []byte {
 	b := make([]byte, coinIDSize)
-	b[0] = byte(flags >> 8)
-	b[1] = byte(flags)
+	binary.BigEndian.PutUint16(b[:2], uint16(flags))
 	if IsSwapCoinID(flags) {
 		copy(b[2:], addr[:])
 	}
--- a/server/asset/eth/common.go
+++ b/server/asset/eth/common.go
@@ -37,17 +37,17 @@ const (
 	GweiFactor = 1e9
 )
 
-// TxIDFlag and SwapFlag are used in CoinIDs to signify a coinID
+// CIDTxID and CIDSwap are used in CoinIDs to signify a coinID
 // as either a transaction ID or a combination of a swap contract
 // address and secret hash. One or the other must be set.
 const (
-	// TxID indicates that this coin ID's hash is a txid. The address
+	// CIDTxID indicates that this coin ID's hash is a txid. The address
 	// portion is zeros and unused.
-	TxIDFlag CoinIDFlag = 1 << iota
-	// SwapFlag indicates that this coin ID represents a swap with a
+	CIDTxID CoinIDFlag = 1 << iota
+	// CIDSwap indicates that this coin ID represents a swap with a
 	// contract address and secret hash used to fetch data about a swap
 	// from the live contract.
-	SwapFlag
+	CIDSwap
 )
 
 // ToGwei converts a *big.Int in wei (1e18 unit) to gwei (1e9 unit) as a uint64.
@@ -76,19 +76,23 @@ func (ss SwapState) String() string {
 	return "unknown"
 }
 
-// DecodeCoinID decodes the coin ID into flags, a contract address which may be
-// zeroed, and hash that represents either a secret or txid depending on flags.
+// DecodeCoinID decodes the coin ID into flags, a contract address, and hash
+// that represents either a secret or txid depending on flags. The CIDTxID flag
+// indicates that this coinID represents a simple txid. The address return is
+// unused and bytes are a 32 byte txid. Bytes are in the same order as the txid.
+// The CIDSwap flag indicates that this coinID represents a swap contract. The
+// address is the swap contract's address and the bytes return is the 32 byte
+// secret hash of a swap. Errors if the passed coinID is not the expected length.
 func DecodeCoinID(coinID []byte) (CoinIDFlag, *common.Address, []byte, error) {
-	var addr common.Address
 	if len(coinID) != coinIDSize {
 		return 0, nil, nil, fmt.Errorf("coin ID wrong length. expected %d, got %d",
 			coinIDSize, len(coinID))
 	}
-	hash := make([]byte, 32)
+	hash, addr := make([]byte, 32), new(common.Address)
 	copy(hash, coinID[22:])
 	copy(addr[:], coinID[2:22])
 	return CoinIDFlag(binary.BigEndian.Uint16(coinID[:2])),
-		&addr, hash, nil
+		addr, hash, nil
 }
 
 // CoinIDToString converts coinID into a human readable string.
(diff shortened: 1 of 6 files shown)


server/asset/eth/common.go
```diff
@@ -57,32 +76,48 @@ func (ss SwapState) String() string {
...
-func ToCoinID(flags uint16, addr *common.Address, secretHash []byte) []byte {
+// ToCoinID converts the address and secret hash, or txid to a coin ID.
+func ToCoinID(flags CoinIDFlag, addr *common.Address, hash []byte) []byte {
 	b := make([]byte, coinIDSize)
-	b[0] = byte(flags)
-	b[1] = byte(flags >> 8)
-	copy(b[2:], addr[:])
-	copy(b[22:], secretHash[:])
+	b[0] = byte(flags >> 8)
+	b[1] = byte(flags)
```
JoeGruffins, line 105: Good think I added a test.

JoeGruffins:
Using pointers to `common.Address` where possible. https://golang.org/doc/effective_go#arrays But more that that it bothered me to sorta use one or the other. Some of the `ethclient` methods from go-ethereum still take the value anyway.

chappjc:
> Using pointers to `common.Address` where possible. https://golang.org/doc/effective_go#arrays But more than that it bothered me to sorta use one or the other. Some of the `ethclient` methods from go-ethereum still take the value anyway.

Probably go-ethereum works with value args in places for the sweeter syntax (less `&` or `new` usage) and because pointer vs value often conveys intent to modify in the function, but +1 to passing pointers to avoid needless allocations.

If we have a `common.Address` as a struct field however, we should probably stick to values for sake of data locality, but that's a separate concern.

server/asset/eth/common.go
```diff
@@ -78,15 +78,17 @@ func (ss SwapState) String() string {
 
 // DecodeCoinID decodes the coin ID into flags, a contract address which may be
 // zeroed, and hash that represents either a secret or txid depending on flags.
-func DecodeCoinID(coinID []byte) (CoinIDFlag, common.Address, []byte, error) {
+func DecodeCoinID(coinID []byte) (CoinIDFlag, *common.Address, []byte, error) {
+	var addr common.Address
 	if len(coinID) != coinIDSize {
-		return 0, common.Address{}, nil, fmt.Errorf("coin ID wrong length. expected %d, got %d",
+		return 0, &addr, nil, fmt.Errorf("coin ID wrong length. expected %d, got %d",
```
chappjc, line 84: Looks like we can/should be returning nil for the `*common.Address` in the event of non-nil error return, but I haven't really gotten into this yet.

server/asset/eth/eth_test.go
```diff
@@ -180,7 +180,7 @@ func TestDecodeCoinID(t *testing.T) {
 			t.Fatalf("want flags value of %v but got %v for test %v",
 				test.wantFlags, flags, test.name)
 		}
-		if addr != test.wantAddr {
+		if !bytes.Equal(addr[:], test.wantAddr[:]) {
```
chappjc, line 183: comparing slices with bytes.Equal is fine, but I'd probably go with `if *addr != *test.wantAddr` a la `(*chainhash.Hash).IsEqual`

https://github.com/decred/dcrd/blob/48627ee18c3d43319d9b27d51b70afe0e5eb6eb2/chaincfg/chainhash/hash.go#L63-L71

JoeGruffins: I forgot this worked.

server/asset/eth/common.go
JoeGruffins: Returning nil.

JoeGruffins:
> If we have a common.Address as a struct field however, we should probably stick to values for sake of data locality, but that's a separate concern.

Do you have some reading material about this? We often store pointers in structs, and in some cases it seems to just be better to use pointers to avoid unexpected bugs like #1124

chappjc:
> > If we have a common.Address as a struct field however, we should probably stick to values for sake of data locality, but that's a separate concern.
> 
> Do you have some reading material about this? We often store pointers in structs, and in some cases it seems to just be better to use pointers to avoid unexpected bugs like #1124

Embedding of a struct by value had a unique set of gotchas, especially when it contained a mutex.  Although I'm really thinking about fields that are simple types rather than structs.  For a field that is a simple type like an array, it makes sense to have the memory for all of the struct's data, including the array, be contiguous, from a single allocation.  I suppose it depends on the specific case though.

https://github.com/decred/dcrd/commit/20a8ccc27bef9812d7e548f25a37011fe1cb542f

JoeGruffins:
ty! Garbage collection is the main reason?

chappjc:
> ty! Garbage collection is the main reason?

GC, but also memory access efficiency.  Accessing adjacent memory addresses will often benefit from low-level cache lines, thus speeding up successive access to multiple fields. Go's 64-bit alignment of structs memory and elimination of excess padding for small (<8 byte) fields makes this even more likely.  Optimizing for this in this application is surely not critical, but it's a reason.

Also if we have a pointer to an array as a struct field, that's a separate heap allocated object.  Might as well have the array just be a segment of the struct's memory.  Of course there can be reasons why this isn't desirable, but in general I feel it's cleanest to just avoid the extra level of indirection, esp for small arrays like Hash or this common.Address type.  However if we fine ourselves doing lots of unnecessary copies or obtaining pointers to these fields, we might reconsider this approach.

server/asset/eth/common.go
```diff
@@ -57,32 +76,49 @@ func (ss SwapState) String() string {
 	return "unknown"
 }
 
-// DecodeCoinID decodes the coin ID into flags, a contract address, and secret hash.
-func DecodeCoinID(coinID []byte) (uint16, common.Address, []byte, error) {
+// DecodeCoinID decodes the coin ID into flags, a contract address which may be
+// zeroed, and hash that represents either a secret or txid depending on flags.
```
chappjc, line 80: If the meaning of "contract address which may be zeroed" is that a `CoinIDFlag` of `TxIDFlag` will return a pointer to a a zero-value `common.Address`, I think it would be most helpful to have that spelled out in the `DecodeCoinID` docs instead of just the `TxIDFlag` docs.

server/asset/eth/common.go
```diff
@@ -57,32 +76,49 @@ func (ss SwapState) String() string {
...
 	if err != nil {
 		return "", err
 	}
-	return fmt.Sprintf("%x:%x:%x", flags, addr, secretHash), nil
+	return fmt.Sprintf("%x:%x:%x", flags, addr, hash), nil
 }
 
-// ToCoinID converts the address and secret hash to a coin ID.
-func ToCoinID(flags uint16, addr *common.Address, secretHash []byte) []byte {
+// ToCoinID converts the address and secret hash, or txid to a coin ID.
```
chappjc, line 103: Let's add that the `hash []byte` should be 32 bytes, and longer slices are truncated.  Either that or we panic or something, but a doc would be fine imo.

server/asset/eth/common.go
```diff
@@ -31,6 +37,19 @@ const (
...
+// as either a transaction ID or a combination of a swap contract
+// address and secret hash. One or the other must be set.
+const (
+	// TxID indicates that this coin ID's hash is a txid. The address
+	// portion is zeros and unused.
+	TxIDFlag CoinIDFlag = 1 << iota
+	// SwapFlag indicates that this coin ID represents a swap with a
+	// contract address and secret hash used to fetch data about a swap
+	// from the live contract.
+	SwapFlag
```
chappjc, line 50: With enum value names I really like having a common prefix that assists with discovering usable values (e.g. IDE completion).  e.g. `SigHashAll`, `SigHashNone`, `SigHashSingle`, etc. being various `SigHashType`s. e.g. `STEcdsaSecp256k1`, `STEd25519`, and `STSchnorrSecp256k1` being the various `SignatureType`s. e.g. all the `txscript.ScriptFlags`. Also same convention that linters like to suggest for error type names.

As in the dcrd examples above, the prefix can deviate from the type names for brevity.  Ideas

`CoinFlagTxID`
`CoinFlagSwap`

`CIDFlagTxID`
`CIDFlagSwap`

`CIDTxID`
`CIDFlagSwap`

Feelings about the general approach of prefixing or specific prefixes?

chappjc (commented):
Looks good, just a few thoughts.

server/asset/eth/common.go
JoeGruffins: Sounds Good. Will go with one of those.

JoeGruffins: Will also change the SwapStates naming I guess.

server/asset/eth/common.go
JoeGruffins: Added a doc.

server/asset/eth/common.go
JoeGruffins: Tried to be more informative.

chappjc (approved):

Issue #1154, opened 2021-08-06
eth: Add backend methods.


JoeGruffins:
This issue is meant to track adding functionality to eth swaps. Currently there is a bare bones implementation. Anyone who would like to work on adding a method, please comment here first so work is not duplicated.

As the dex was originally designed with UTXO based coins in mind, some methods **may** require far reaching changes.

server
- [x] TxData(coinID []byte) ([]byte, error)
- [x] InitTxSize() uint32
- [x] InitTxSizeBase() uint32
- [x] Contract(coinID []byte, redeemScript []byte) (*asset.Contract, error)
- [x] ValidateSecret(secret, contract []byte) bool
- [x] Redemption(redemptionID, contractID []byte) (asset.Coin, error)

~- [ ] FundingCoin(ctx context.Context, coinID []byte, redeemScript []byte) (asset.FundingCoin, error)~
- [x] ValidateCoinID(coinID []byte) (string, error)
- [x] ValidateContract(contract []byte) error

~- [ ] VerifyUnspentCoin(ctx context.Context, coinID []byte)~
- [x] run(ctx context.Context) // fix TODOs and refactor
- [ ] ValidateSignature(addr string, pubkey, msg, sig []byte) error

client
- [x] Balance() (*asset.Balance, error) // fix TODO
- [x] MaxOrder(lotSize uint64, feeSuggestion uint64, nfo *dex.Asset) (*asset.SwapEstimate, error)
- [x] PreSwap(req *asset.PreSwapForm) (*asset.PreSwap, error)
- [x] PreRedeem(req *asset.PreRedeemForm) (*asset.PreRedeem, error)
- [x] FundOrder(ord *asset.Order) (asset.Coins, []dex.Bytes, error)
- [x] ReturnCoins(unspents asset.Coins) error
- [x] FundingCoins(ids []dex.Bytes) (asset.Coins, error)
- [x] Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin, uint64, error)
- [x] Redeem(form *asset.RedeemForm) ([]dex.Bytes, asset.Coin, uint64, error)
- [x] SignMessage(coin asset.Coin, msg dex.Bytes) (pubkeys, sigs []dex.Bytes, err error)
- [x] AuditContract(coinID, contract, txData dex.Bytes) (*asset.AuditInfo, error)
- [x] LocktimeExpired(contract dex.Bytes) (bool, time.Time, error)
[...]

PR #1157, opened 2021-08-13
server/eth: Implement InitTxSize.

JoeGruffins:
Also implement InitTxSizeBase.
    
    The size of a tx does not actually tell us how much a transaction paying
    to a contract with extra data will cost to send. The evm will determine
    how much gas will be used. The exact cost can probably be determined but
    will take a lot of digging, and will change if the contract changes. For
    now, use an approximation of used gas based off of tests.
    
    Our current implementation also does not allow for batch transactions.
    As such, return zero for InitTxSizeBase in order to ensure fee
    calculations equate to what we expect, which is a one for one
    initilizaion and gas fee.

Part of #1154

### DIFF
--- a/server/asset/eth/eth_test.go
+++ b/server/asset/eth/eth_test.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"decred.org/dcrdex/dex"
+	"decred.org/dcrdex/dex/calc"
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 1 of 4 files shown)


server/asset/eth/eth_test.go
```diff
@@ -498,3 +499,28 @@ func TestToCoinID(t *testing.T) {
...
+func TestRequiredOrderFunds(t *testing.T) {
+	eth := new(Backend)
+	swapVal := uint64(1000000000)                // gwei
+	numSwaps := uint64(17)                       // swaps
+	initSizeBase := uint64(eth.InitTxSizeBase()) // 0 gas
+	initSize := uint64(eth.InitTxSize())         // init value gas
+	feeRate := uint64(30)                        // gwei
+
+	// We want the fee calculation to simply be the gas used for each swap
+	// plus the initial value.
```
chappjc, line 514: Clear enough, but would it be more accurate to say "**the cost** of the gas used for each swap", because it's gas*gas_price = gas * (gwei / gas) = gwei.

In that vein, I feel like it could provide more clarity with `feeRate` line comment as `// gwei / gas`, even though just `gwei` seems to be most common.

chappjc (approved):
Question about gas cost units, but otherwise g2g.

server/asset/eth/eth_test.go
JoeGruffins: Wait, what was the question. lol. Changed the wording in the test. Was there another place?

It looks like I did make that distinction in the `backend.FeeRate` method doc. 

Gas is a fixed unit in the evm and certain actions cost a certain amount of it. It doesn't care about supply/demand. the gasPrice is how much one gas costs, and is wei or gwei per gas. It is variable depending on market conditions. Most recently 100 gwei per gas.

I'm sure you understood already, just reinterating your reinteration. gasPrice is the cost of a gas unit.

chappjc: > the gasPrice is how much one gas costs, and is wei or gwei per gas

Just this.  You say "fee calculation to simply be the gas used for each swap plus" but it's gas * feeRate plus...

Yeah, the change in https://github.com/decred/dcrdex/pull/1157/commits/c60a4e29db37f70e474f810e459b4c9a29cc2c07 is exactly what I was getting at, that gas cost is (gwei / gas) regardless of what websites say, and that why the units work out when multiplying by gas.

PR #1159, opened 2021-08-17
server/eth: Add Contract method. 

JoeGruffins:
The eth contract requires the initial transaction used to initiate a
    swap to extrapolate swap data. Confirm data upon Confirmations being
    called.

Part of #1154

### DIFF
--- a/server/asset/eth/coiner.go
+++ b/server/asset/eth/coiner.go
@@ -1,0 +1,284 @@
+// This code is available on the terms of the project LICENSE.md file,
+// also available online at https://blueoakcouncil.org/license/1.0.0.
+
+//go:build lgpl
+// +build lgpl
+
+package eth
+
+import (
+	"context"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"math/big"
+
+	dexeth "decred.org/dcrdex/dex/networks/eth"
+	"decred.org/dcrdex/server/asset"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+type swapCoinType uint8
+
+const (
+	sctInit swapCoinType = iota
+	sctRedeem
+)
+
+var _ asset.Coin = (*swapCoin)(nil)
+
+type swapCoin struct {
+	backend                    *Backend
+	contractAddr, counterParty common.Address
+	secret, secretHash         [32]byte
+	value                      uint64
+	gasPrice                   uint64
+	txid                       string
+	locktime                   int64
+	sct                        swapCoinType
+}
+
+// newSwapCoin creats a new swapCoin that stores and retrieves info about a
+// swap. It requires a coinID that is a txid type of the initial transaction
+// initializing or redeeming the swap. A txid type and not a swap type is
+// required because the contract will give us no useful information about the
+// swap before it is mined. Having the initial transaction allows us to track
+// it in the mempool. It also tells us all the data we need to confirm a tx
+// will do what we expect if mined and satisfies contract contraints. These
+// fields are verified the first time the Confirmations method is called, and
+// an error is returned then if something is different than expected. As such,
+// the swapCoin expects Confirmations to be called with confirmations
+// available at least once before the swap be trusted for swap initializations.
+func newSwapCoin(backend *Backend, coinID []byte, sct swapCoinType) (*swapCoin, error) {
+	switch sct {
+	case sctInit, sctRedeem:
+	default:
+		return nil, fmt.Errorf("unknown swapCoin type: %d", sct)
+	}
+
+	cID, err := DecodeCoinID(coinID)
+	if err != nil {
+		return nil, err
+	}
+	txCoinID, ok := cID.(*TxCoinID)
+	if !ok {
+		return nil, errors.New("coin ID not a txid")
+	}
+	txid := txCoinID.TxID
+	tx, _, err := backend.node.transaction(backend.rpcCtx, txid)
+	if err != nil {
+		return nil, fmt.Errorf("unable to fetch transaction: %v", err)
+	}
+
+	txdata := tx.Data()
+	// Transactions that call contract functions must have extra data to do
+	// so.
+	if len(txdata) == 0 {
+		return nil, errors.New("tx calling contract functions should have extra data")
+	}
+
+	var (
+		counterParty       = new(common.Address)
+		secret, secretHash [32]byte
+		locktime           int64
+	)
+
+	switch sct {
+	case sctInit:
+		counterParty, secretHash, locktime, err = dexeth.ParseInitiateData(txdata)
+	case sctRedeem:
+		secret, secretHash, err = dexeth.ParseRedeemData(txdata)
+	}
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse call data: %v", err)
+	}
+
+	contractAddr := tx.To()
+	if *contractAddr != backend.contractAddr {
+		return nil, errors.New("contract address is not supported")
+	}
+
+	// Gas price is not stored in the swap, and is used to determine if the
+	// initialization transaction could take a long time to be mined. A
+	// transaction with a very low gas price may need to be resent with a
+	// higher price.
+	gasPrice, err := ToGwei(tx.GasPrice())
+	if err != nil {
+		return nil, fmt.Errorf("unable to convert gas price: %v", err)
+	}
+
+	// Value is stored in the swap but it is also this value. Because we
+	// can't yet be sure this transaction represents the passed secret
+	// hash, value should be checked by using the contract once it has
+	// confirmations.
+	value, err := ToGwei(tx.Value())
+	if err != nil {
+		return nil, fmt.Errorf("unable to convert value: %v", err)
+	}
+
+	// For redemptions, the transaction should move no value.
+	if sct == sctRedeem && value != 0 {
+		return nil, fmt.Errorf("expected swapCoin value of zero for redeem but got: %d", value)
+	}
+
+	return &swapCoin{
+		backend:      backend,
+		contractAddr: *contractAddr,
+		secret:       secret,
+		secretHash:   secretHash,
+		value:        value,
+		gasPrice:     gasPrice,
+		txid:         hex.EncodeToString(txid[:]),
+		counterParty: *counterParty,
+		locktime:     locktime,
+		sct:          sct,
+	}, nil
+}
+
+// Confirmations returns the number of confirmations for a Coin's
+// transaction. Because a Coin can become invalid after once being
+// considered valid, this condition should be checked for during
+// confirmation counting and an error returned if this Coin is no longer
+// ready to spend. An unmined transaction should have zero confirmations. A
+// transaction in the current best block should have one confirmation. A
+// negative number can be returned if error is not nil.
+//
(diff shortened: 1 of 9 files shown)


JoeGruffins:
A difference here with other coins is that we can't verify some parts of the swap until the tx is mined. We can't decode the data in the initial transaction (w/o hacking?) and the contract doesn't return data on the swap until mined.

server/asset/eth/coiner.go
```diff
@@ -0,0 +1,248 @@
...
+type swapCoinerType uint8
+
+const (
+	sctInit swapCoinerType = iota
+	sctRedeem
+)
+
+var _ asset.Coin = (*swapCoiner)(nil)
+
+type swapCoiner struct {
```
JoeGruffins, line 26: Naming on this correct? It should be the interface with `-er` and not the implementer?

server/asset/eth/coiner.go
```diff
@@ -0,0 +1,248 @@
...
+
+// TxID is the original init transaction txid.
+func (c *swapCoiner) TxID() string {
+	return c.txid
+}
+
+// String is a human readable representation of the swap coin.
+func (c *swapCoiner) String() string {
+	// Not currently possible to error here, so error is ignored.
+	s, _ := CoinIDToString(ToCoinID(CIDSwap, &c.contractAddr, c.secretHash[:]))
```
JoeGruffins, line 234: Don't ignore.

JoeGruffins:
> Because we are unable to get some data from the purchasing tx

I though this was true, but not so sure now. Maybe can decode? This looks like a hackish decoding https://ethereum.stackexchange.com/questions/29809/how-to-decode-input-data-with-abi-using-golang/100247

buck54321:
I see something similar in geth. https://github.com/ethereum/go-ethereum/blob/84ff152de56b9d65f6888daef9dc00c97a6aafd3/signer/fourbyte/abi.go#L118, if that means anything.

JoeGruffins:
@buck54321 thank you so much for pointing that out. At first glance it looks like we can use it ver-batim  #1215

server/asset/eth/coiner.go
JoeGruffins: I think the interface should have the `-er` so mass renamed.

server/asset/eth/coiner.go
JoeGruffins: Not ignoring any more.

JoeGruffins:
The second parameter for Contract, should we do something with that? What exactly is the purpose for other coins? Is it that the script would be hashed and we can't get some values any longer from an on-chain tx output? Here we can get everything on chain I believe.

chappjc:
> The second parameter for Contract, should we do something with that? What exactly is the purpose for other coins? Is it that the script would be hashed and we can't get some values any longer from an on-chain tx output? Here we can get everything on chain I believe.

Presently there's no purpose for that with ethereum afaict.  When server's `Swapper` gets a `init` request, it gives it to a `(Backend).Contract` to validate that the contract script for UTXO coins actually pertains to the UTXO.  It validates that, and also picks out the following into an `asset.Contract` struct:

```go
type Contract struct {
	Coin
	// SwapAddress is the receiving address of the swap contract.
	SwapAddress string
	// RedeemScript is the contract redeem script.
	RedeemScript []byte
	// LockTime is the refund locktime.
	LockTime time.Time
	// TxData is raw transaction data. This data is provided for some assets
	// to aid in SPV compatibility.
	TxData []byte
}
```

In `(*Swapper).processInit`, there are a bunch of checks on these things like mainly the counterparty's swap address being correct.  So in ETH, the `Contract` method can get this data for itself without a "redeemScript" providing it, I believe.

If we start working with different contract addresses,  I can see this argument, which comes from the `msgjson.Init.Contract` field, being the contract address being used in the swap, but there's clearly no script as you point out.

Ideally we won't have to modify `Swapper` too much, but I suspect some changes will be needed as we get further along.

server/asset/eth/coiner.go
```diff
@@ -0,0 +1,281 @@
...
+		return -1, fmt.Errorf("redemption in failed state with swap at %s state", ss)
+
+	case sctInit:
+		// Uninitiated state is zero confs. It could still be in mempool.
+		// It is important to only trust confirmations according to the
+		// swap contract. Until there are confirmations in the contract,
+		// we cannot be sure that initiation happened successfuly. If
+		// the swap has zero confirmations but the init transaction has
+		// confirmations, something is wrong or initialization failed
+		// and this contract should be dropped.
```
JoeGruffins, line 206: contract -> swap

server/asset/eth/coiner.go
```diff
@@ -0,0 +1,281 @@
...
+			if err := assertMempool(); err != nil {
+				return -1, err
+			}
+			// Assume the tx still has a chance of being mined.
+			return 0, nil
+		}
+		// Any other swap state is ok. We are sure that initialization
+		// happened.
+
+		// The contract has some number of confirmations, and we are
```
JoeGruffins, line 217: contract -> swap

chappjc: Mmm, yeah, the "contract" language was too utxo specific here.  "swap" as you suggest is probably best, but I wouldn't object to "swap initiation transaction" in places that you don't mind being more explicit, but probably that would be needlessly wordy to say everywhere.

JoeGruffins: Ahh, I don't know. Transactions still gotta happen... We are requiring the transaction tx, though we could not, it's probably good to get things started in mempool like other coins. Also we don't save the "txfee" in the contract, we could, but having the initial tx works too.

server/asset/eth/eth.go
```diff
@@ -202,8 +218,17 @@ func (eth *Backend) BlockChannel(size int) <-chan *asset.BlockUpdate {
 }
 
 // Contract is part of the asset.Backend interface.
-func (eth *Backend) Contract(coinID []byte, redeemScript []byte) (*asset.Contract, error) {
-	return nil, notImplementedErr
+func (eth *Backend) Contract(coinID, _ []byte) (*asset.Contract, error) {
+	sc, err := newSwapCoin(eth, coinID, sctInit)
+	if err != nil {
+		return nil, fmt.Errorf("unable to create coiner: %v", err)
+	}
```
JoeGruffins, line 225: Maybe should check Confirmations once for errors here, in case it is already mined.

server/asset/eth/coiner.go
JoeGruffins: Changed here.

server/asset/eth/eth.go
```diff
@@ -106,13 +120,15 @@ var _ asset.Backend = (*Backend)(nil)
 // before use.
 func unconnectedETH(logger dex.Logger, cfg *config) *Backend {
 	ctx, cancel := context.WithCancel(context.Background())
+	mainnetContractAddr := common.HexToAddress(mainnetContractAddr)
```
chappjc, line 123: For a first go, we'd at least want this switching the contract address based on the network, no?

server/asset/eth/rpcclient.go
```diff
@@ -27,16 +29,22 @@ type rpcclient struct {
...
+func (c *rpcclient) connect(ctx context.Context, IPC string, contractAddr *common.Address) error {
 	client, err := rpc.DialIPC(ctx, IPC)
 	if err != nil {
 		return fmt.Errorf("unable to dial rpc: %v", err)
 	}
 	c.ec = ethclient.NewClient(client)
+	c.es, err = swap.NewETHSwap(*contractAddr, c.ec)
+	if err != nil {
+		return fmt.Errorf("unable to find swap contract: %v", err)
+	}
```
chappjc, line 47: Feels a bit awkward and confusing that `contractAddr` is an input arg to `connect`.  Can `(*rpcclient).swap` have the contract as an input instead?  I think we'd need it as an arg to `swap` eventually when multiple contract addresses are possible, as would be needed for even contract upgrades.

Maybe I'm just fussing because there's no real `rpcclient` constructor, presently just `(*Backend) Connect` manually making the struct, and instead `connect` is being dual purposed.  These are all internals so I guess it's fine, but an API I would get would have a constructor that takes a slice/map of contracts to expect/support, then `connect` would setup an `ETHSwap` for each one using the provided IPC client, then `swap` would pick the one for the provided contract address.

server/asset/eth/coiner.go
```diff
@@ -0,0 +1,284 @@
...
+// initializing or redeeming the swap. A txid type and not a swap type is
+// required because the contract will give us no useful information about the
+// swap before it is mined. Having the initial transaction allows us to track
+// it in the mempool. It also tells us all the data we need to confirm a tx
+// will do what we expect if mined and satisfies contract contraints. These
+// fields are verified the first time the Confirmations method is called, and
+// an error is returned then if something is different than expected. As such,
+// the swapCoin expects Confirmations to be called with confirmations
+// available at least once before the swap be trusted for swap initializations.
+func newSwapCoin(backend *Backend, coinID []byte, sct swapCoinType) (*swapCoin, error) {
```
chappjc, line 52: Seems clear and reasonable.  Just a couple docs typos: "creats" and "contraints"

server/asset/eth/coiner.go
```diff
@@ -0,0 +1,284 @@
...
+		if ss == SSRedeemed {
+			// While not completely accurate, we know that if the
+			// swap is redeemed the redemption has at least one
+			// confirmation.
+			return 1, nil
+		}
+		// If swap is in the Initiated state, the transaction must be
+		// unmined.
+		if ss == SSInitiated {
+			if err := assertMempool(); err != nil {
```
chappjc, line 187: Can you clarify this for me?  Why must it be in mempool?
If this is a redeem coin/txn, and the contract claims that the state is in initiated still, I think that makes sense.  However, this looks like a race to me -- get swap state, if "initiated" get redeem transaction mempool status.  If in initiated state and this is a redeem coin confirmations check, would it hurt to just report 0 without asserting mempool?

server/asset/eth/coiner.go
```diff
@@ -0,0 +1,284 @@
...
+		locktime:     locktime,
+		sct:          sct,
+	}, nil
+}
+
+// Confirmations returns the number of confirmations for a Coin's
+// transaction. Because a Coin can become invalid after once being
+// considered valid, this condition should be checked for during
+// confirmation counting and an error returned if this Coin is no longer
+// ready to spend. An unmined transaction should have zero confirmations. A
```
chappjc, line 142: Does "this condition should be checked" refer to something the caller should do, or something that should be done internal to `Confirmation`?  This sounds like you're describing something `Confirmations` does, and which the caller should recognize.  If so, which error path is it?  The one from `c.backend.node.transaction` (`unable to fetch transaction`)?  Should this be coded?

And is this generally referring to how all eth txns are replacable by making a txn with the same sender nonce, which you are describing in the last paragraph of this method doc, or is this part alluding to something else?

server/asset/eth/coiner.go
JoeGruffins: Just assuming the tx is still in mempool if we don't see the contract changing seems like the cleanest solution for now. Removed assertion.

chappjc (approved):
All looks fine.  We will obviously need to evolve the contract address business quite a bit as this matures, but it's good for now.

PR #1160, opened 2021-08-18
build: go [1.16,1.17], nodejs [14,16], golangci-lint 1.42

chappjc:


dajohi (approved):

go.mod
```diff
@@ -1,6 +1,6 @@
 module decred.org/dcrdex
 
-go 1.15
+go 1.16
 
 replace decred.org/dcrdex/internal/eth/reentryattack => ./internal/eth/reentryattack
```
JoeGruffins, line 5: I guess you can take this out now if you want. I think we just needed it before the new package was merged to master.

JoeGruffins (approved):

go.mod
chappjc: Yeah, removed.  Strange though because "decred.org/dcrdex/internal/eth/reentryattack" isn't even a module.

PR #1162, opened 2021-08-20
eth/config: Add contract address to configs.

JoeGruffins:
server/eth/config: Add contract address.

    Add ability to parse a config file rather than just the ipc location.

    client/eth/settings: Add contract address.

    Rather than use a hardcoded address, require the address be set upon
    initializaion of the wallet.

Was part of #1159 but decided to make a separate pr. I believe this is the correct direction to go for which contract to use. The contract must still have the same byte code as the hard-coded one, but we don't currently have a way to verify that. I wonder if a different contract will cause initial binding to fail? Need to look into that.

### DIFF
--- a/client/asset/eth/config.go
+++ b/client/asset/eth/config.go
@@ -4,10 +4,12 @@
 package eth
 
 import (
+	"errors"
 	"fmt"
 
 	"decred.org/dcrdex/dex"
 	"decred.org/dcrdex/dex/config"
+	"github.com/ethereum/go-ethereum/common"
 )
 
 // Config holds the parameters needed to initialize an ETH wallet.
@@ -15,6 +17,7 @@ type Config struct {
 	AppDir         string  `ini:"appdir"`
 	NodeListenAddr string  `ini:"nodelistenaddr"`
 	GasFee         float64 `ini:"gasfee"`
+	ContractAddr   string  `ini:"contractaddr"`
 }
 
 // loadConfig loads the Config from a setting map and checks the network.
@@ -33,5 +36,13 @@ func loadConfig(settings map[string]string, network dex.Network) (*Config, error
 	default:
 		return nil, fmt.Errorf("unknown network ID: %d", uint8(network))
 	}
+
+	if cfg.ContractAddr == "" {
+		return nil, errors.New("no swap contract address specified in config file")
+	}
+	if !common.IsHexAddress(cfg.ContractAddr) {
+		return nil, errors.New("contract address is structually invalid")
+	}
+
 	return cfg, nil
 }
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -57,10 +57,19 @@ var (
 			DefaultValue: defaultGasFeeLimit,
 		},
 		{
-			Key:          "appdir",
-			DisplayName:  "DCR Dex Ethereum directory location.",
-			Description:  "Location of the ethereum client data. This SHOULD NOT be a directory used by other ethereum applications. The default is recommended.",
+			Key:         "appdir",
+			DisplayName: "DCR Dex Ethereum directory location.",
+			Description: "Location of the ethereum client data. This SHOULD NOT be a " +
+				"directory used by other ethereum applications. The default is recommended.",
 			DefaultValue: defaultAppDir,
+		},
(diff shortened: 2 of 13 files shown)


buck54321:
What happens when the server and client have different contract addresses?

JoeGruffins:
Currently, the server would error on "Contract" method being called. There must be some validation somewhere that server and client are on the same contract. Where to put?

JoeGruffins:
I don't think the wallet versioning can be used for this if we allow operators to choose the contract. If the changes in this diff seem reasonable, we would need to advertise the supported contract address(es) in `Exchanges` I guess.

martonp:
Why do you say that wallet versioning won't work if the operator decides which contract to use?

For me it makes most sense for the server to decide which contract to use, and the client must support at least the version of the contract that the server decides. For example, if there is a new version of the contract that supports redemptions by a third party, and one client wants to create the swap on a contract that doesn't support this feature yet, they cannot be matched with clients that require this feature. If the server is deciding which contract to use, then all clients of the dex will be on the same page, and this won't be an issue.

JoeGruffins:
> Why do you say that wallet versioning won't work if the operator decides which contract to use?

Because if different operators could set different contracts, there would be no way to ensure version 3 is this contract for every server. I am specifically referring to the number versioning we already have in place for assets.

After re-reading, maybe you are referring to the contract version. I do think the asset versions can handle that well, but not the addresses of those contracts.

> For me it makes most sense for the server to decide which contract to use, and the client must support at least the version of the contract that the server decides.

I think that's optimal as well. The version of the contract and the address are different however. The server should decide which contracts it supports. Currently we only support one, which is a problem for upgrades and will need to be fixed.

For me the debate is, hard-code the address, which would allow easier versioning, or allow choice. However, I feel like hard-coding this is not best for decentralization. There should be nothing wrong with a server using a different contract, although it should contain the same byte code the client expects. Plus it won't currently work if the client were to assign a different address. Also, perhaps having any connection to a specific contract is a liability that dcrdex doesn't need?

martonp:
I don’t the address should be hardcoded either. The server should just select the contract they want to use, and the client can validate that the bytecode is the same as expected. When there’s a new version of the contract, the server would have to let clients know they need to update their software before switching.

If this is how it’s done, then you wouldn’t need to input the contract address on the clients config file. I don’t think dex users should have to figure out which contract to use anyways.

JoeGruffins:
> If this is how it’s done, then you wouldn’t need to input the contract address on the clients config file. I don’t think dex users should have to figure out which contract to use anyways.

Currently we need the contract address in Connect, so I think there will need to be some more changes to allow this, but it does sound like a good idea. As long as the byte code is as expected, it only matters as far as coordinating the swap, which is the server's job.

martonp (approved):
Looks good.

martonp:
For sure, this is good for now. I'll add an issue to figure out how to compare bytecodes of contracts.

chappjc:
> For sure, this is good for now. I'll add an issue to figure out how to compare bytecodes of contracts.

Thanks.

> The server should just select the contract they want to use, and the client can validate that the bytecode is the same as expected. ... I don’t think dex users should have to figure out which contract to use anyways.

I agree with the above.  I don't know how a dex user can be expected to pick a contract address.

Requiring it in Connect also clear needs to change to do this right.  Don't forget that a user can be connected to multiple servers at once.

I would expect (assuming a client can in fact verify contract acceptability via a bytecode comparison) that a server would convey it's ethereum swap contract address on connect in the `config` response.

chappjc:
Relating to @buck54321's comment here: https://github.com/decred/dcrdex/issues/1198#issuecomment-916485681

I'm also quite uneasy about having the contract be either user-configurable or even communicated by the server to clients at this point, although the latter sits a little better with me despite being ripe for exploits.

I'd really like the first iteration of ETH and ERC-20 to use hard coded contract addresses.  We get that working and we can get sophisticated an allow the contract addresses to be variable.

JoeGruffins:
> Would this PR still be required even if dexc or even Core make the user use a contract we have audited here in our repo?

No. This pr was meant to take it in the opposite direction.

Closing. Can always re-open though.

Will put the test updates in another pr.

PR #1164, opened 2021-08-20
rpc/client: Minimum fee rate for swap transactions

martonp:
This PR adds the ability to specify a custom fee rate for swap transactions. The ability to specify it on the UI will be added in another PR. 

If the custom transaction fee is greater than the server's max transaction fee, then it will be the fee used in the transaction. A transaction fee less than the server's max transaction fee may also be specified, but in this case it acts as a minimum transaction fee. If in the match request sent from the server there is a higher fee rate than the one specified by the user, the higher fee will be used.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -609,12 +609,12 @@ func (dc *dexConnection) parseMatches(msgMatches []*msgjson.Match, checkSigs boo
 
 		if !isCancel &&
 			swapRate > tracker.metaData.MaxFeeRate &&
-			(tracker.metaData.CustomSwapFeeRate == nil || swapRate > *tracker.metaData.CustomSwapFeeRate) {
+			(tracker.metaData.MinSwapFeeRate == nil || swapRate > *tracker.metaData.MinSwapFeeRate) {
 
-			if tracker.metaData.CustomSwapFeeRate != nil {
+			if tracker.metaData.MinSwapFeeRate != nil {
 				errs = append(errs,
-					fmt.Sprintf("rejecting match %s for order %s because assigned rate (%d) is > MaxFeeRate (%d) and customSwapFeeRate (%d)",
-						msgMatch.MatchID, msgMatch.OrderID, swapRate, tracker.metaData.MaxFeeRate, *tracker.metaData.CustomSwapFeeRate))
+					fmt.Sprintf("rejecting match %s for order %s because assigned rate (%d) is > MaxFeeRate (%d) and minSwapFeeRate (%d)",
+						msgMatch.MatchID, msgMatch.OrderID, swapRate, tracker.metaData.MaxFeeRate, *tracker.metaData.MinSwapFeeRate))
 			} else {
 				errs = append(errs,
 					fmt.Sprintf("rejecting match %s for order %s because assigned rate (%d) is > MaxFeeRate (%d)",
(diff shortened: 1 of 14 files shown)


client/core/core.go
```diff
@@ -606,9 +606,20 @@ func (dc *dexConnection) parseMatches(msgMatches []*msgjson.Match, checkSigs boo
...
-				msgMatch.MatchID, msgMatch.OrderID, swapRate, tracker.metaData.MaxFeeRate))
+
+		if !isCancel &&
+			swapRate > tracker.metaData.MaxFeeRate &&
+			(tracker.metaData.CustomSwapFeeRate == nil || swapRate > *tracker.metaData.CustomSwapFeeRate) {
+
+			if tracker.metaData.CustomSwapFeeRate != nil {
+				errs = append(errs,
+					fmt.Sprintf("rejecting match %s for order %s because assigned rate (%d) is > MaxFeeRate (%d) and customSwapFeeRate (%d)",
+						msgMatch.MatchID, msgMatch.OrderID, swapRate, tracker.metaData.MaxFeeRate, *tracker.metaData.CustomSwapFeeRate))
```
chappjc, line 617: Hmm, the server is free to prescribe any rate up to `MaxFeeRate`, so I don't think rejecting the match on account of `CustomSwapFeeRate` is correct.

client/rpcserver/handlers.go
```diff
@@ -862,8 +862,10 @@ Registration is complete after the fee transaction has been confirmed.`,
     quote (int): The BIP-44 coin index for the market's quote asset.
     qty (int): The number of units to buy/sell. Must be a multiple of the lot size.
     rate (int): The atoms quote asset to pay/accept per unit base asset. e.g.
-      156000 satoshi/DCR for the DCR(base)_BTC(quote).
-    immediate (bool): Require immediate match. Do not book the order.`,
+	56000 satoshi/DCR for the DCR(base)_BTC(quote).
+    immediate (bool): Require immediate match. Do not book the order.
+    customSwapFeeRate (int): Optional. Custom fee rate to use for swap transactions. The fee
+	may be higher if server's max fee rate is higher, but will never be lower.`,
```
chappjc, line 868: There's a little ambiguity here.  The custom rate should be used when `server's match rate < customSwapFeeRate < maxFeeRate`.  This doc line would be clearer if "The fee" were "The actual fee rate used in the swap transaction".  The doc could even say that the setting is used to use a higher fee rate than what might be prescribed by the server.

chappjc: This would also need to specify what asset the rate applies to (it depends on `sell`).

client/rpcserver/handlers.go
```diff
@@ -862,8 +862,13 @@ Registration is complete after the fee transaction has been confirmed.`,
     quote (int): The BIP-44 coin index for the market's quote asset.
     qty (int): The number of units to buy/sell. Must be a multiple of the lot size.
     rate (int): The atoms quote asset to pay/accept per unit base asset. e.g.
-      156000 satoshi/DCR for the DCR(base)_BTC(quote).
-    immediate (bool): Require immediate match. Do not book the order.`,
+      56000 satoshi/DCR for the DCR(base)_BTC(quote).
```
chappjc, line 865: Just noticing the unintentional change from 156000 => 56000.   When this was written (May 28, 2020), it was describing an exchange rate of 0.00156 BTC/DCR, so I'd not object to a new optimistic example rate of 543210 (0.0054321) instead, but definitely not 0.00056 :)

buck54321:
I've resurrected some old order-options work (origin of #958 and cryptically alluded to [here](https://github.com/decred/dcrdex/blob/f3f2f160b1a46a83c9436d8bfec39ca8506190eb/client/asset/interface.go#L283-L284)). I think there's a larger conversation to be had about providing a system for handling these lower level options. I'll try to get something up today so you can see what I mean.

chappjc:
Recommend closing in favor of #1170.  I never like wasted work, but I think this started a productive dialog around order options.

PR #1166, opened 2021-08-24
client/asset/{dcr,btc}: prevent swap broadcast on refund tx errors

### LINKED ISSUE #1163: client/asset/{dcr,btc}: shouldn't error on failure to generate refund tx
As [pointed out by @itswisdomagain](https://github.com/decred/dcrdex/pull/1089#discussion_r684492015), if we encounter an error [generating the signed refund tx](https://github.com/decred/dcrdex/blob/75c9b7a765767f0555e7642fc70e8df3160b65a1/client/asset/btc/btc.go#L1427), we shouldn't return an error, since the swap has already been broadcast. I see two possible solutions. 

1) Print error, don't populate the `signedRefund` field, handle empty field in **core**.
2) Separate generating/signing from broadcasting. Generate and sign. Get tx hash and generate refund. Return any errors encountered, else broadcast.

itswisdomagain:
Do not broadcast swap(s) if there's any trouble generating or signing
the refund txn(s).

Resolves #1163.

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -1411,8 +1411,9 @@ func (btc *ExchangeWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin
 		return nil, nil, 0, fmt.Errorf("error creating change address: %w", err)
 	}
 
-	// Sign, add change, and send the transaction.
-	msgTx, change, fees, err := btc.sendWithReturn(baseTx, changeAddr, totalIn, totalOut, swaps.FeeRate)
+	// Sign, add change, but don't send the transaction yet until
+	// the individual swap refund txs are prepared and signed.
+	msgTx, change, fees, err := btc.signTxAndAddChange(baseTx, changeAddr, totalIn, totalOut, swaps.FeeRate)
 	if err != nil {
 		return nil, nil, 0, err
 	}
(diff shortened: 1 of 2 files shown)


client/asset/btc/btc.go
```diff
@@ -2439,12 +2446,25 @@ func (btc *ExchangeWallet) convertCoin(coin asset.Coin) (*output, error) {
 // dust) for the change.
 func (btc *ExchangeWallet) sendWithReturn(baseTx *wire.MsgTx, addr btcutil.Address,
 	totalIn, totalOut, feeRate uint64) (*wire.MsgTx, *output, uint64, error) {
```
chappjc, line 2448: Only the MsgTx and error outputs are used now.  Fine to remove the others I think

client/asset/btc/btc.go
```diff
@@ -2439,12 +2446,25 @@ func (btc *ExchangeWallet) convertCoin(coin asset.Coin) (*output, error) {
...
-	// Sign the transaction to get an initial size estimate and calculate whether
-	// a change output would be dust.
+
+	signedTx, change, fee, err := btc.signTxAndAddChange(baseTx, addr, totalIn, totalOut, feeRate)
+	if err == nil {
```
chappjc, line 2451: Why not `if err != nil { return nil, nil, 0, err }`?  That's a bit more idiomatic and less confusing IMO

chappjc (changes requested):
Nice minimal diff. Thanks for a quick fix.  Just a couple minor things.  I also need to test.

client/asset/btc/btc.go
itswisdomagain: Yeah... Was what I wrote initially but looked unnecessarily repetitive. It's surely clearer. Will do.

chappjc (approved):
Looks good. Works for me.  @JoeGruffins a look from you would be good if you have a little time for it.

PR #1167, opened 2021-08-24
CI: check for localized_html modifications

chappjc:


chappjc:
Added the updated localized_html/en-US/markets.tmpl and CI passed, so that was a decent test of the CI change, I think.

Issue #1176, opened 2021-08-28
[0.3] eth: remove all eth asset init code for release


chappjc:
Since 0.3 does not support eth yet, but we have been merging ethereum code as we go, we will need to disable all of the eth asset `init` code when we branch `release-0.3`

e.g. lots of go-ethereum goroutines are running in dexc even when the asset is otherwise not used

```
32 @ 0x440016 0x40ddcc 0x40d838 0xbab95b 0x470a41
#	0xbab95a	github.com/ethereum/go-ethereum/core.(*txSenderCacher).cache+0x3a	/home/jon/go/pkg/mod/github.com/ethereum/go-ethereum@v1.10.6/core/tx_cacher.go:63
...
1 @ 0x440016 0x40ddcc 0x40d838 0x917b85 0x470a41
#	0x917b84	github.com/ethereum/go-ethereum/metrics.(*meterArbiter).tick+0x64	/home/jon/go/pkg/mod/github.com/ethereum/go-ethereum@v1.10.6/metrics/meter.go:290
...
1 @ 0x440016 0x44fcb2 0xb4faa5 0x470a41
#	0xb4faa4	github.com/ethereum/go-ethereum/consensus/ethash.(*remoteSealer).loop+0x204	/home/jon/go/pkg/mod/github.com/ethereum/go-ethereum@v1.10.6/consensus/ethash/sealer.go:278
```

This is just dexc sitting uninitialized.  I recently noticed many many more when debugging an separate issue.

PR #1177, opened 2021-08-28
[0.3] dexc: comment eth import for release

chappjc:
Resolves https://github.com/decred/dcrdex/issues/1176

The base branch to which this PR applies will need to be changed when we branch `release-0.3`.  I'm putting this up now so I do not forget.


JoeGruffins (approved):

buck54321 (approved):

chappjc:
@JoeGruffins @martonp Can this actually just go on `master` and stay there until some future point, or would you want this import already for dexc?  I think the work you guys are doing doesn't require this at all (yet), but I'm not certain.

martonp:
I've only been running unit tests and the harness test, so I'm fine with this.

PR #1178, opened 2021-08-28
Fix max order UI.

### LINKED ISSUE #1169: site: o.maxLoaded is not a function
Started testing master after updating dex-test to 0.3 and my client hit the following JS console error:

image

I had placed a buy order at 0.004 for 1 lot by typing in 0.004, then for a sell order I clicked on the depth chart to get 0.004022 as the sell price and submitted the order.  I'm not certain if the buy was still epoch.

There may have been other clicks. I can't be sure.  Should be reproducible or at least clear from the code how this state can be reached.

martonp:
There were multiple issues with the max order UI. There would be a console
error saying the the maxLoaded function was null. This would happen if there
were two concurrent requests where the one that started first finished last.
Also, if a cached value was input while a another value was being fetched,
the cached result would be overridden by the value being fetched from the
client causing a mismatch between the rate in the input box and the result
on the screen.

Closes #1169.

### DIFF
--- a/client/webserver/site/src/js/markets.js
+++ b/client/webserver/site/src/js/markets.js
@@ -764,6 +764,10 @@ export default class MarketsPage extends BasePage {
       if (!app.checkResponse(res, true)) {
         console.warn('max order estimate not available:', res)
         page.maxFromLots.textContent = 'estimate unavailable'
+        if (this.maxLoaded) {
+          this.maxLoaded()
+          this.maxLoaded = null
+        }
         return
       }
       success(res)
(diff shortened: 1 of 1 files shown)


buck54321 (changes requested):
Righteous changes, but there's a bug for estimating the **max sell with an empty book** where the loading spinner is not removed.

Screenshot from 2021-08-28 12-19-33

Granted, this won't be encountered often in the wild.

martonp:
@buck54321 Should have been hiding the spinner after errors as well. It should be good now.

chappjc (approved):
Issue is resolved.

JoeGruffins (approved):

buck54321 (approved):

PR #1183, opened 2021-08-31
client/core: app seed optimisations

### LINKED ISSUE #1172: client: error when restoring from seed
When restoring from seed, seeing error message: **Connection to dex server failed. You can close dexc and try again later or wait for it to reconnect.**

itswisdomagain:
Primarily resolves #1172.

Also splits `core.register` into `core.registerHD` and `core.registerLegacy`. `core.PreRegister` now only attempts HD account registration to determine if an account exists and can be restored.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -2656,8 +2656,7 @@ func (c *Core) register(dc *dexConnection, encKey, encKeyLegacy, acctPubB []byte
 				return nil, true, fmt.Errorf("error authorizing pre-paid account: %v", err)
 			}
 
-			if !dc.broadcastingConnect() { // dc might already be listening for messages if called from Register which creates a non-temporary dc.
-				atomic.StoreUint32(&dc.reportingConnects, 1)
+			if atomic.CompareAndSwapUint32(&dc.reportingConnects, 0, 1) { // dc might already be listening for messages if called from Register which creates a non-temporary dc.
 				c.wg.Add(1)
 				go c.listen(dc)
 			}
@@ -4852,11 +4851,10 @@ func sendOutdatedClientNotification(c *Core, dc *dexConnection) {
 
 // connectDEX establishes a ws connection to a DEX server using the provided
 // account info, but does not authenticate the connection through the 'connect'
-// route. If temporary is provided and true, no reconnect handler is registered
-// and the c.listen(dc) goroutine is not started so that associated trades are
-// not processed and no incoming requests are notifications are handled. A
-// temporary dexConnection may be used to inspect the config response or check
-// if a (paid) HD account exists with a DEX.
+// route. If temporary is provided and true, the c.listen(dc) goroutine is not
+// started so that associated trades are not processed and no incoming requests
+// and notifications are handled. A temporary dexConnection may be used to
+// inspect the config response or check if a (paid) HD account exists with a DEX.
 func (c *Core) connectDEX(acctInfo *db.AccountInfo, temporary ...bool) (*dexConnection, error) {
 	// Get the host from the DEX URL.
 	host, err := addrHost(acctInfo.Host)
@@ -4905,10 +4903,8 @@ func (c *Core) connectDEX(acctInfo *db.AccountInfo, temporary ...bool) (*dexConn
 	wsCfg.ConnectEventFunc = func(connected bool) {
 		c.handleConnectEvent(dc, connected)
 	}
-	if listen {
-		wsCfg.ReconnectSync = func() {
-			go c.handleReconnect(host)
-		}
+	wsCfg.ReconnectSync = func() {
+		go c.handleReconnect(host)
 	}
 
 	// Create a websocket connection to the server.
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -6,7 +6,6 @@ package core
 import (
 	"bytes"
 	"context"
-	"encoding/binary"
 	"encoding/hex"
 	"errors"
 	"fmt"
@@ -36,7 +35,6 @@ import (
 	"decred.org/dcrdex/dex/wait"
 	"decred.org/dcrdex/server/account"
 	serverdex "decred.org/dcrdex/server/dex"
-	"github.com/decred/dcrd/chaincfg/chainhash"
 	"github.com/decred/dcrd/dcrec/secp256k1/v3"
 	"github.com/decred/dcrd/dcrec/secp256k1/v3/ecdsa"
 	"github.com/decred/go-socks/socks"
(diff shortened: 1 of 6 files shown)


client/core/core.go
```diff
@@ -2618,42 +2523,154 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
...
+
+			dc.acct.markFeePaid()
+
+			err = c.authDEX(dc)
+			if err != nil {
+				return nil, true, fmt.Errorf("error authorizing pre-paid account: %v", err)
+			}
+
+			if !dc.broadcastingConnect() { // dc might already be listening for messages if called from Register which creates a non-temporary dc.
+				atomic.StoreUint32(&dc.reportingConnects, 1)
```
chappjc, line 2660: A compare and swap here would be appropriate instead of using the `broadcastingConnect` helper

chappjc:
Thanks for enabling the reconnect handler too, as discussed in https://github.com/decred/dcrdex/issues/1172#issuecomment-909660161.  I should have commented on the PR instead of the issue.

The PreRegister change does make its function consistent with its docs.  Will review that asap.  Tagged for the 0.3 mileston.

chappjc (changes requested):
This looks very good to me.  I think the refactoring of the registration methods adds clarity and robustness to the process.

@buck54321 For Decrediton/libdexc, it's not clear if `PreRegister` will become part of the workflow.  I think it depends on if Decrediton is updated with a "restore from seed?" prompt prior to calling `InitializeClient` or if it just passes a `nil` seed for now.  Ultimately, DEX account recovery from Decrediton should work the same way as with our update browser dialogs, but I don't know if there is time to change Decrediton views for 1.7.0 at this point.

CC @alexlyp and @matheusd regarding the possibility to restore dex accounts from a seed, which would require new prompts and changes to libdexc.  At a minimum a required change is updating the `InitializeClient` call with the added seed arg (nil means no restore).  An optional change would be to modify Decrediton dialogs and DEX workflow to first ask for a restoration seed. At this point, it can then call `PreRegister` method, which will perform account discovery at a specified server, but will not actually make any txns.  The consumer _can_ just skip `PreRegister` and go straight to `Register`, but the UX with that will still be suggesting that a fee will be paid if the account is not found, when it's probably desirable to know first.  Does that make sense to you guys?

See my suggested revisions to the `PreRegister` docs in this review for more info on that.

buck54321 (changes requested):
Nice work. Looks and tests well. Just a couple of tiny things for now.

chappjc:
Bump.  This PR is pretty solid and a needed fix for 0.3.  Can we get an iteration on the current reviews in the next day or two?  Thanks!

client/core/core.go
```diff
@@ -2625,42 +2523,153 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
...
+// The result of this registration attempt will be returned to enable follow up
+// action if this is a fresh registration with the server or if the account
+// already exists but is suspended.
+// The registration result is nil if the account exists and the fee is paid. If
+// the account exists and fee is paid, the account is restored, the dc is auth'ed
+// and added to the conns map and a goroutine is started to listen for server
+// messages.
+func (c *Core) register(dc *dexConnection, encKey, encKeyLegacy, acctPubB []byte) (inProcessReg *inProcessRegistration, paid bool, err error) {
+	if len(encKey) > 0 && len(encKeyLegacy) > 0 {
+		return nil, false, fmt.Errorf("cannot register with BOTH HD acct key and legacy acct key.")
```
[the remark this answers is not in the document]
itswisdomagain: Yeah, I know about this, must have missed it. Thanks!

client/core/core.go
```diff
@@ -2625,42 +2523,153 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
...
+			}
+
+			dc.acct.markFeePaid()
+
+			err = c.authDEX(dc)
+			if err != nil {
+				return nil, true, fmt.Errorf("error authorizing pre-paid account: %v", err)
+			}
+
+			if atomic.CompareAndSwapUint32(&dc.reportingConnects, 0, 1) { // dc might already be listening for messages if called from Register which creates a non-temporary dc.
```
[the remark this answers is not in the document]
itswisdomagain: I think the order was previously adding dc to c.conns before authDEX but I reverted the order here to ensure dc is not promoted and does not remain in c.conns if authDEX errors. Will revert back to the original order to prevent unwittingly breaking something.

client/core/core.go
```diff
@@ -2625,42 +2523,153 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
...
+	regRes := new(msgjson.RegisterResult)
+	err = dc.signAndRequest(dexReg, msgjson.RegisterRoute, regRes, DefaultResponseTimeout)
+	inProcessReg = &inProcessRegistration{
+		host:         dc.acct.host,
+		regRes:       regRes,
+		acctPub:      acctPubB,
+		encKey:       encKey,
+		encKeyLegacy: encKeyLegacy,
+		stamp:        time.Now(),
+	}
```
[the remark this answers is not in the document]
itswisdomagain: This value is returned from inside the `if err == nil` block, and also outside it.

itswisdomagain:
So I renamed PreRegister to DiscoverAccount at once. Hope that's fine.

client/core/core.go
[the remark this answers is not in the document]
chappjc: Understood that this PR isn't changing the order.  Just observing that it's a deviation from how things are set up with a regular registration whereby `authDEX` comes after the connection is listening for messages and added to the `conns` map. I'll take a closer look to see if anything will fail, but I don't believe there's an issue if it remains in `c.conns` if `authDEX` errors.  In fact, it's how `initializeDEXConnections` operates presently, and it might even be necessary to prevent a repeated `Register`/`DiscoverAccount`>`CreateAccount` that would fail.

client/core/core.go
```diff
@@ -2621,10 +2642,19 @@ func (c *Core) register(dc *dexConnection, encKey, encKeyLegacy, acctPubB []byte
 			}
 
 			dcrID, _ := dex.BipSymbolID(regFeeAssetSymbol)
-			cid, _ := asset.DecodeCoinID(dcrID, dc.acct.feeCoin)
+			cid, err := asset.DecodeCoinID(dcrID, dc.acct.feeCoin)
+			if err != nil {
+				// feeCoin decoded from the server's error message above is invalid,
+				// use a dummy dcr fee coin.
+				var zeroHash chainhash.Hash
+				dc.acct.feeCoin = make([]byte, chainhash.HashSize+4)
```
chappjc, line 2650: Quite unfortunate to have to do this.  Alternatives are fail registration entirely even though server recognizes us (not good), or do `dc.acct.feeCoin = []byte{1}` so the len > 0 checks that are all over the place pass.  The latter only would work if we ensure there's no other call to `asset.DecodeCoinID(dcrID, dc.acct.feeCoin)` where an error would cause a problem.

Thoughts @buck54321?  We have to check this error as it's server provided data.

buck54321: My read is that we can set it to a constant, even if invalid, as long as we don't error at `c.db.AccountPaid`. The `(*BoltDB).Account` methods use `acctInfo.Paid = len(acct.Get(feeProofKey)) > 0`, and we really don't use the address again, e.g. we'll never call `reFee` with this coin.

```Go
dc.acct.feeCoin, err = hex.DecodeString(msgErr.Message)
if err != nil {
	c.log.Errorf("Failed to decode fee coin from pre-paid account info message = %q, err = %w", msgErr.Message, err)
	dc.acct.feeCoin = []byte("DUMMY COIN")
} else {
	dcrID, _ := dex.BipSymbolID(regFeeAssetSymbol)
	cid, err := asset.DecodeCoinID(dcrID, dc.acct.feeCoin)
	if err != nil {
		c.log.Errorf("Failed to decode Decred coin ID for pre-paid account from feeCoin = %x, err = %w", dc.acct.feeCoin, err)
	} else {
		c.log.Infof("Recovered paid account for %s. Fee previously paid with %s", dc.acct.host, cid)
	}
}
```

as long as we can be certain that `(*BoltDB).AccountPaid` passes?

chappjc: > we'll never call `reFee` with this coin

Yeah, true.  That's where I saw that the content mattered but it's not used after this.  Let's go with the code you proposed @buck54321. **Except** we still need to catch `msgErr.Message == ""` 

https://play.golang.org/p/uB6PV_9wpc8

itswisdomagain: Is there a reason to think `(*BoltDB).AccountPaid` might fail? Just looked through the implementation and doesn't seem to concern itself with the fee coin. Looks safe to me.

chappjc (approved):

PR #1184, opened 2021-08-31
eth/client: Implement MaxOrder

martonp:
This diff completes the MaxOrder function for ETH and also adds
functions that calculate the gas needed for initializing and
redeeming a swap on ETH. The gas calculation is function is used
in MaxOrder, but in case of a calculation failure, the hardcoded
gas value that is returned from the server is used.

Part of #1154

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -307,9 +307,7 @@ func (eth *ExchangeWallet) MaxOrder(lotSize uint64, feeSuggestion uint64, nfo *d
 	now := time.Now().Unix()
 	var secretHash [32]byte
 	copy(secretHash[:], encode.RandomBytes(32))
-	// TODO: replace with configured contract address
-	contractAddress := common.HexToAddress("2f68e723b8989ba1c6a9f03e42f33cb7dc9d606f")
-	initGas, err := eth.node.initiateGas(eth.ctx, now, secretHash, &eth.acct.Address, &contractAddress)
+	initGas, err := eth.node.initiateGas(eth.ctx, now, secretHash, &eth.acct.Address, &mainnetContractAddr)
 	if err != nil {
 		eth.log.Warnf("error getting init gas, falling back to server's value: %v", err)
 		initGas = nfo.SwapSize
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -40,6 +40,8 @@ const (
 	BipID              = 60
 	defaultGasFee      = 82  // gwei
 	defaultGasFeeLimit = 200 // gwei
+
+	RedeemGas = 63000 // gas
 )
 
 var (
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -121,7 +121,7 @@ type ethFetcher interface {
 	importAccount(pw string, privKeyB []byte) (*accounts.Account, error)
 	listWallets(ctx context.Context) ([]rawWallet, error)
 	initiate(opts *bind.TransactOpts, netID int64, refundTimestamp int64, secretHash [32]byte, participant *common.Address) (*types.Transaction, error)
-	initiateGas(ctx context.Context, msg ethereum.CallMsg) (uint64, error)
+	estimateGas(ctx context.Context, msg ethereum.CallMsg) (uint64, error)
 	lock(ctx context.Context, acct *accounts.Account) error
 	nodeInfo(ctx context.Context) (*p2p.NodeInfo, error)
 	pendingTransactions(ctx context.Context) ([]*types.Transaction, error)
@@ -316,7 +316,7 @@ func (eth *ExchangeWallet) getInitGas() (uint64, error) {
 		Gas:   0,
 		Data:  data,
 	}
-	return eth.node.initiateGas(eth.ctx, msg)
+	return eth.node.estimateGas(eth.ctx, msg)
 }
 
 // MaxOrder generates information about the maximum order size and associated
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -299,22 +299,22 @@ func (eth *ExchangeWallet) Balance() (*asset.Balance, error) {
 
 // getInitGas gets an estimate for the gas required for initiating a swap.
 func (eth *ExchangeWallet) getInitGas() (uint64, error) {
-	now := time.Now().Unix()
 	var secretHash [32]byte
 	copy(secretHash[:], encode.RandomBytes(32))
 	parsedAbi, err := abi.JSON(strings.NewReader(swap.ETHSwapABI))
 	if err != nil {
 		return 0, err
 	}
-	data, err := parsedAbi.Pack("initiate", big.NewInt(now), secretHash, &eth.acct.Address)
+	data, err := parsedAbi.Pack("initiate", big.NewInt(1), secretHash, &eth.acct.Address)
 	if err != nil {
 		return 0, err
 	}
 	msg := ethereum.CallMsg{
-		From: eth.acct.Address,
-		To:   &mainnetContractAddr,
-		Gas:  0,
-		Data: data,
+		From:  eth.acct.Address,
+		To:    &mainnetContractAddr,
+		Value: big.NewInt(1),
+		Gas:   0,
+		Data:  data,
 	}
 	return eth.node.initiateGas(eth.ctx, msg)
 }
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -24,6 +24,7 @@ import (
 	"github.com/decred/dcrd/dcrutil/v4"
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts"
+	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
@@ -120,7 +121,7 @@ type ethFetcher interface {
 	importAccount(pw string, privKeyB []byte) (*accounts.Account, error)
 	listWallets(ctx context.Context) ([]rawWallet, error)
 	initiate(opts *bind.TransactOpts, netID int64, refundTimestamp int64, secretHash [32]byte, participant *common.Address) (*types.Transaction, error)
-	initiateGas(ctx context.Context, refundTimestamp int64, secretHash [32]byte, participant, contractAddress *common.Address) (uint64, error)
+	initiateGas(ctx context.Context, msg ethereum.CallMsg) (uint64, error)
 	lock(ctx context.Context, acct *accounts.Account) error
 	nodeInfo(ctx context.Context) (*p2p.NodeInfo, error)
 	pendingTransactions(ctx context.Context) ([]*types.Transaction, error)
@@ -296,6 +297,28 @@ func (eth *ExchangeWallet) Balance() (*asset.Balance, error) {
 	return bal, nil
 }
 
+// getInitGas gets an estimate for the gas required for initiating a swap.
+func (eth *ExchangeWallet) getInitGas() (uint64, error) {
+	now := time.Now().Unix()
+	var secretHash [32]byte
+	copy(secretHash[:], encode.RandomBytes(32))
+	parsedAbi, err := abi.JSON(strings.NewReader(swap.ETHSwapABI))
+	if err != nil {
+		return 0, err
+	}
+	data, err := parsedAbi.Pack("initiate", big.NewInt(now), secretHash, &eth.acct.Address)
+	if err != nil {
+		return 0, err
+	}
+	msg := ethereum.CallMsg{
+		From: eth.acct.Address,
+		To:   &mainnetContractAddr,
+		Gas:  0,
+		Data: data,
+	}
+	return eth.node.initiateGas(eth.ctx, msg)
+}
+
 // MaxOrder generates information about the maximum order size and associated
 // fees that the wallet can support for the given DEX configuration. The fees are an
 // estimate based on current network conditions, and will be <= the fees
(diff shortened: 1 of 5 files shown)


martonp:
I actually implemented this using the `EstimateGas` function at first, but changed it use the hardcoded value. I think it could be nice to use it though, as the amount of gas needed will always be accurate.

chappjc:
If you had an implementation of `EstimateGas` or some other helpful code, even if to demonstrate in a test that the hardcoded value is correct, I'd not object.  But I agree calling it all the time for the same result doesn't sound necessary.

martonp:
I put that code into another PR: #1186

JoeGruffins (approved):
Looks correct to me.

I just want to note that nfo.SwapSize will be gas and not size, but I think the calculations are still the same.

client/asset/eth/eth_test.go
```diff
@@ -349,3 +350,124 @@ func TestBalance(t *testing.T) {
...
+	}
+
+	ethToWei := func(eth int64) *big.Int {
+		return big.NewInt(0).Mul(big.NewInt(eth*dexeth.GweiFactor), big.NewInt(dexeth.GweiFactor))
+	}
+
+	tests := []struct {
+		name          string
+		bal           *big.Int
+		balErr        error
```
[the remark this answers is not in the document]
martonp: Thanks, I'll use that from now on.

martonp:
It's slightly off topic, but another thing to note is that now with EIP-1559, there is the base gas fee and also the gas tip. I'm not 100% sure how to handle this yet, it seems like the max fee rate coming from the server should be a combination of both, since that will be used to determine the amount of funds that will be locked up. For the `feeSuggestion` in the `MaxOrder` function, the combined fee will be fine, but when the server sends the fee suggestion after a match, then it will need to send a suggestion for just the gas tip.

JoeGruffins:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md

"By burning the base fee, we can no longer guarantee a fixed Ether supply." Jumped out at me. unrelated...

I don't quite understand yet but will reread a few times.

JoeGruffins:
Do the miners run completely on "tips" now? Do we need to tip to get a tx in a block?

chappjc:
Just need base fee, and the tip is like a bribe to get ahead of the queue.

I'm not certain actually what is meant by "but when the server sends the fee suggestion after a match, then it will need to send a suggestion for just the gas tip".  I would expect it would send the combined fee (base+tip).  Part of the reason we have server sending fee rates for the other assets is because client's cannot be relied upon to determine anything accurately, e.g. their node is not synchronized and is reporting lower base fee or something.

JoeGruffins:
After reading a few times, I am still not clear on the importance of the tip. If that means miners can just include based on tips, it's not really a change at all is it? They'll include txs with a high tip. I'll find more to read later.

chappjc:
Supposedly the tip is a lot less important now because of the algorithm that modulates base fee. A key to it working is supposedly having a lot more block space possible, but the algorithm targeting 50% full in terms of a 30 million gas use cap or something.  So supposedly it would be unlikely to be delayed by more than a handful of blocks, whereas before your txn could get sunk for like 18 hrs when fees swung from single digits to triple digits.

buck54321:
Tip can be an order option, à la #1170.

JoeGruffins:
> it would be unlikely to be delayed by more than a handful of blocks

We would still include something reasonable I think... I guess can have a non-zero default and let the user decide if they want something else.

martonp:
> Do the miners run completely on "tips" now? Do we need to tip to get a tx in a block?

They still get the new inflation, and the tip (priority fee) is a bonus.

> I'm not certain actually what is meant by "but when the server sends the fee suggestion after a match, then it will need to send a suggestion for just the gas tip". I would expect it would send the combined fee (base+tip). 

The base fee is not specified in the transaction, it's automatically determined by the protocol in each block. The only things specified in the transaction are the tip and the "max fee rate" which is the max fee (base + tip) that the user is willing to pay. Therefore, the server just has to tell the client what tip to specify. The max fee rate we used to determine how much funds to lock can be the max fee rate specified in the transaction. 

I did a quick look through the ETH block explorer, and it seems like everyone puts between 1-2 gwei as the tip. Maybe miners will just leave out 0 tip transactions as a matter of principle. If there is a congested period when the blocks are completely full and your tip is lower than everyone else's, you might end up paying a much higher base fee, because by the time your transaction gets in, the base fee will have risen significantly. There is a `suggestTipCap` function in geth, I think we can return that from the server, and the user can specify a minimum fee or a `feeBump` as an option.

chappjc:
> If there is a congested period when the blocks are completely full and your tip is lower than everyone else's, you might end up paying a much higher base fee, because by the time your transaction gets in, the base fee will have risen significantly

This is a bit I'm still trying to grasp, the max fee thing.  There's a base fee we can't change.  There's a tip to specify.  And there's a max fee, which I guess is how high the txn is allowed to pay in (base+tip)?

martonp:
The max fee specifies the most you're willing to pay. So if you submit your transaction and `base_fee + tip > max_fee`, the transaction will just stay in the mempool until the `base_fee + tip` drops below `max_fee`.

chappjc:
OK, thanks @martonp.  This is finally becoming clear.  I checkout out some post-london wallets and I think I get what you describing now.

Since you can't set or in any way control `base_fee` as it is set by consensus with an algo and changes _from block to block_, all you can do is specify `tip` that goes on top of whatever the `base_fee` happens to be **at any given block**.  That part is what I didn't at first.  You do not send your tx into mempool/txpool without an absolute constant fee rate, just a specific tip that makes your fee relative to whatever `base_fee` happens to be when it does get mined.

So I suspect we'll be doing the following:
- `max_fee` for the tx will have to be something fairly high, like 350 gwei/gas, and this would actually be taken from the `dex.Asset.MaxFeeRate` field from the server's `'config'` response.
- suggested fee rate sent from the server will be just like you said @martonp, and that will be something like 2 gwei/gas

So in this PR, the `feeSuggestion` in the `MaxOrder` args would be something else, perhaps 110% of current `base_fee`, because `base_fee` can be changing rapidly and the estimate should have some padding to remain be relevant.  It seems that taking the _current_ `base_fee + tip` as the input `feeSuggestion` could end up being too low quite soon after (but also too high).  e.g. see how much base fee can change from one block to the next: https://etherchain.org/burn

martonp:
@chappjc I agree with all this. Maybe taking the average of the past 10 blocks and bumping that by a certain amount would work as well, but that will be part of a separate PR.

I've incorporated the gas estimation functions into this PR. `initiateGas` is used in `MaxOrder`, but the hardcoded value from the server will be used as a backup. We cannot always do an estimation, especially for redemption, since estimating the gas for redemption requires a swap to already exist, but in this way we can do a best effort. We will also have the simnet tests that will make sure the hardcoded values are accurate, and we can update the hardcoded values if there are any ETH upgrades. Also, a high enough max fee rate will give us a nice barrier.

Also, I'm marking this as a draft because this PR now depends on #1162 .

martonp:
This is ready to go now that we are using a hardcoded contract address.

client/asset/eth/eth.go
```diff
@@ -297,8 +301,39 @@ func (eth *ExchangeWallet) Balance() (*asset.Balance, error) {
...
+		return nil, err
+	}
+	now := time.Now().Unix()
+	var secretHash [32]byte
+	copy(secretHash[:], encode.RandomBytes(32))
+	initGas, err := eth.node.initiateGas(eth.ctx, now, secretHash, &eth.acct.Address, &mainnetContractAddr)
+	if err != nil {
+		eth.log.Warnf("error getting init gas, falling back to server's value: %v", err)
+		initGas = nfo.SwapSize
+	}
```
buck54321, line 316: How about warning `if initGas != nfo.SwapSize`.

buck54321 (changes requested):

client/asset/eth/eth.go
```diff
@@ -297,8 +301,39 @@ func (eth *ExchangeWallet) Balance() (*asset.Balance, error) {
...
+	availableBalance := balance.Available
```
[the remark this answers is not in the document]
martonp: Yes I'm working on that but I don't think anything should change over here. The available amount returned from `Balance()` will already have the locked funds subtracted from it.

client/asset/eth/eth.go
martonp: It's not going to be equal. We are making the hardcoded value a bit higher to be safe.

client/asset/eth/rpcclient.go
```diff
@@ -256,6 +258,56 @@ func (c *rpcclient) initiate(txOpts *bind.TransactOpts, netID int64, refundTimes
 	return c.es.Initiate(txOpts, big.NewInt(refundTimestamp), secretHash, *participant)
 }
 
+// initiateGas checks the amount of gas that is used for a call to the initiate function.
+func (c *rpcclient) initiateGas(ctx context.Context, refundTimestamp int64, secretHash [32]byte, participant *common.Address, contractAddress *common.Address) (uint64, error) {
+	parsedAbi, err := abi.JSON(strings.NewReader(swap.ETHSwapABI))
+	if err != nil {
+		return 0, err
+	}
+	data, err := parsedAbi.Pack("initiate", big.NewInt(refundTimestamp), secretHash, participant)
```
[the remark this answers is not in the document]
martonp: I was thinking to do that also at first too, but then I thought it would be nicer if this function was more flexible and took the arguments that the initiate function on the contract takes and returned the amount of gas it would use with those arguments. If someone initiated a swap where the `secretHash` was 1, then this estimation would constantly break. If you feel strongly about it I'm fine with changing it though, as that probably would never happen.

JoeGruffins (approved):
Still looks good to me. Some small nits but can ignore.

client/asset/eth/rpcclient.go
```diff
@@ -256,6 +258,56 @@ func (c *rpcclient) initiate(txOpts *bind.TransactOpts, netID int64, refundTimes
 	return c.es.Initiate(txOpts, big.NewInt(refundTimestamp), secretHash, *participant)
 }
 
+// initiateGas checks the amount of gas that is used for a call to the initiate function.
+func (c *rpcclient) initiateGas(ctx context.Context, refundTimestamp int64, secretHash [32]byte, participant *common.Address, contractAddress *common.Address) (uint64, error) {
```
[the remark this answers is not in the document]
martonp: I've updated to take the call message.

client/asset/eth/rpcclient.go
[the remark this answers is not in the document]
chappjc: I might be misunderstanding this conversation, but `[32]byte{0x01}` is still a 32-byte array, it's just a syntax for setting one byte of that array so as to make it not the zero value for the secret hash.

client/asset/eth/rpcclient.go
```diff
@@ -256,6 +258,56 @@ func (c *rpcclient) initiate(txOpts *bind.TransactOpts, netID int64, refundTimes
...
+	parsedAbi, err := abi.JSON(strings.NewReader(swap.ETHSwapABI))
```
chappjc: We might consider doing this in the rpcclient constructor and storing the parsedAbi in a field.  It never changes, right?

client/asset/eth/eth.go
```diff
@@ -296,6 +297,28 @@ func (eth *ExchangeWallet) Balance() (*asset.Balance, error) {
...
+	if err != nil {
+		return 0, err
+	}
+	msg := ethereum.CallMsg{
+		From: eth.acct.Address,
+		To:   &mainnetContractAddr,
+		Gas:  0,
+		Data: data,
+	}
+	return eth.node.initiateGas(eth.ctx, msg)
```
chappjc, line 319: This looks much better to me now.

client/asset/eth/eth.go
```diff
@@ -296,6 +297,28 @@ func (eth *ExchangeWallet) Balance() (*asset.Balance, error) {
...
+// getInitGas gets an estimate for the gas required for initiating a swap.
+func (eth *ExchangeWallet) getInitGas() (uint64, error) {
+	now := time.Now().Unix()
+	var secretHash [32]byte
+	copy(secretHash[:], encode.RandomBytes(32))
+	parsedAbi, err := abi.JSON(strings.NewReader(swap.ETHSwapABI))
+	if err != nil {
+		return 0, err
+	}
+	data, err := parsedAbi.Pack("initiate", big.NewInt(now), secretHash, &eth.acct.Address)
```
chappjc, line 309: Is there reason to believe we need to use anything but `math.NewInt(1)` and `[32]byte{1}`?  Although we won't be calling `getInitGas` on a hot path, the system clock and crypto/rand are relatively expensive and it's not clear we need to use them at all.

chappjc (approved):
LGTM. @JoeGruffins pls circle back on your review now that `ethereum.CallMsg` is being passed to be sure it's what you had in mind.

client/asset/eth/eth.go
martonp: I guess the only real reason is if someone initiates a swap with secret key 1, then it will break gas estimation for everyone.

chappjc: Ahhh, so the idea is this secret hash must be unused by the contract on-chain so that it follows the expected execution path and thus gas estimate?

martonp: Yes. I'll can definitely change the use of system clock though.

client/asset/eth/rpcclient.go
```diff
@@ -259,53 +257,13 @@ func (c *rpcclient) initiate(txOpts *bind.TransactOpts, netID int64, refundTimes
...
-	}
-	gas, err := c.ec.EstimateGas(ctx, msg)
-	if err != nil {
-		return 0, err
-	}
-
-	return gas, nil
+func (c *rpcclient) redeemGas(ctx context.Context, msg ethereum.CallMsg) (uint64, error) {
+	return c.ec.EstimateGas(ctx, msg)
 }
```
JoeGruffins: Now that both of these functions are the same, I think you should combine them to a `*rpcclient.estimateGas` function.

chappjc: Yeah, good point. @martonp We'll merge this today if you merge these guys

martonp: Done. This is nicer, good call.

PR #1186, opened 2021-09-01
client/eth: Add functions for estimating gas usage

martonp:
Adds functions to the rpcclient that estimate how much gas would be used to initiate/redeem a swap. They are currently only used in the rpcclient_harness_test to make sure that the hardcoded values for gas are accurate.

### DIFF
--- a/client/asset/eth/rpcclient.go
+++ b/client/asset/eth/rpcclient.go
@@ -293,6 +293,32 @@ func (c *rpcclient) initiateGas(ctx context.Context, contractAddress *common.Add
 	return gas, nil
 }
 
+// initiateGas checks the amount of gas that is used for a call to the redeem function.
+func (c *rpcclient) redeemGas(ctx context.Context, secret, secretHash [32]byte, participant *common.Address, contractAddress *common.Address) (uint64, error) {
+	parsedAbi, err := abi.JSON(strings.NewReader(swap.ETHSwapABI))
+	if err != nil {
+		return 0, err
+	}
+
+	data, err := parsedAbi.Pack("redeem", secret, secretHash)
+	if err != nil {
+		return 0, nil
+	}
+
+	msg := ethereum.CallMsg{
+		From: *participant,
+		To:   contractAddress,
+		Gas:  0,
+		Data: data,
+	}
+	gas, err := c.ec.EstimateGas(ctx, msg)
+	if err != nil {
+		return 0, err
+	}
+
+	return gas, nil
+}
+
 // redeem redeems a swap contract. The redeemer will be the account at txOpts.From.
 // Any on-chain failure, such as this secret not matching the hash, will not cause
 // this to error.
--- a/client/asset/eth/rpcclient.go
+++ b/client/asset/eth/rpcclient.go
@@ -13,6 +13,7 @@ import (
 
 	"decred.org/dcrdex/dex/encode"
 	swap "decred.org/dcrdex/dex/networks/eth"
+	dexeth "decred.org/dcrdex/server/asset/eth"
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts"
 	"github.com/ethereum/go-ethereum/accounts/abi"
@@ -282,7 +283,7 @@ func (c *rpcclient) initiateGas(ctx context.Context, contractAddress *common.Add
 		From:  participant,
 		To:    contractAddress,
 		Gas:   0,
-		Value: big.NewInt(100000000),
+		Value: big.NewInt(dexeth.GweiFactor),
 		Data:  data,
 	}
 	gas, err := c.ec.EstimateGas(ctx, msg)
--- a/client/asset/eth/rpcclient.go
+++ b/client/asset/eth/rpcclient.go
@@ -5,12 +5,17 @@ package eth
 
 import (
 	"context"
+	"crypto/sha256"
 	"fmt"
 	"math/big"
+	"strings"
+	"time"
 
+	"decred.org/dcrdex/dex/encode"
 	swap "decred.org/dcrdex/dex/networks/eth"
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts"
+	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/accounts/keystore"
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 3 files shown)


client/asset/eth/rpcclient.go
```diff
@@ -256,6 +261,64 @@ func (c *rpcclient) initiate(txOpts *bind.TransactOpts, netID int64, refundTimes
...
+		Gas:   0,
+		Value: big.NewInt(100000000),
+		Data:  data,
+	}
+	gas, err := c.ec.EstimateGas(ctx, msg)
+	if err != nil {
+		return 0, err
+	}
+
+	return gas, nil
```
JoeGruffins, line 293: I would prefer if the rpcclient methods had as little logic as possible. So, the caller do all the set up and the rpcclient just calls this last bit. This is because we can test better, as we cannot make a fake *rpcclient struct, but we can make a fake interface that implements the node interface and test the code above this block there. Does that make sense? The client can just accept the msg as an arg. If this sounds silly let me know, it's not set in stone.

server/asset/eth/common.go
```diff
@@ -74,6 +74,8 @@ const (
 	// TODO: When the contract is solidified, break down evm functions
 	// called and the gas used for each. (◍•﹏•)
 	InitGas = 180000 // gas
+
+	RedeemGas = 63000
```
JoeGruffins, line 78: `// gas` at the end for clarity I guess. Lots of different units going around so best to be precise imo.

client/asset/eth/rpcclient.go
```diff
@@ -256,6 +261,64 @@ func (c *rpcclient) initiate(txOpts *bind.TransactOpts, netID int64, refundTimes
...
+	data, err := parsedAbi.Pack("initiate", big.NewInt(refundTimestamp), secretHash, participant)
+	if err != nil {
+		return 0, nil
+	}
+
+	msg := ethereum.CallMsg{
+		From:  participant,
+		To:    contractAddress,
+		Gas:   0,
+		Value: big.NewInt(100000000),
```
JoeGruffins, line 285: I think this is one less power of 10 wei than one gwei, so less than a value we can actually send, as we are rounding everything to gwei. If so could you add a 0?

client/asset/eth/rpcclient_harness_test.go
```diff
@@ -368,6 +368,20 @@ func TestPeers(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("unexpected error from initiateGas: %v", err)
+	}
+	if gas > eth.InitGas {
+		t.Fatalf("actual gas %v is greater than eth.InitGas %v", gas, eth.InitGas)
+	}
+	if gas+10000 < eth.InitGas {
+		t.Fatalf("actual gas %v is much less than eth.InitGas %v", gas, eth.InitGas)
+	}
+	fmt.Printf("Gas used for initiate: %v", gas)
```
JoeGruffins, line 382: Could you add a `\n` for the new line like you have for redeem.

client/asset/eth/rpcclient_harness_test.go
```diff
@@ -454,6 +468,33 @@ func TestInitiate(t *testing.T) {
...
+	}
+	gas, err := ethClient.redeemGas(ctx, secret, secretHash, &participantAddr, &contractAddr)
+	if err != nil {
+		t.Fatalf("Error getting gas for redeem function: %v", err)
+	}
+	if gas > eth.RedeemGas {
+		t.Fatalf("actual gas %v is greater than eth.RedeemGas %v", gas, eth.RedeemGas)
+	}
+	if gas+3000 < eth.RedeemGas {
+		t.Fatalf("actual gas %v is much less than eth.InitGas %v", gas, eth.InitGas)
```
JoeGruffins, line 493: InitGas -> RedeemGas

JoeGruffins:
Also, feel free to ignore, but the commit subject message is rather long. I try to go by this: https://chris.beams.io/posts/git-commit/

client/asset/eth/rpcclient.go
martonp: I don't think having this logic here really restricts testing. If we end up using these two functions outside of the test code, we could easily just return predetermined values for how much gas is used by the two functions.

I was thinking that this functionality that is missing from the autogenerated ETHSwap code. It would be nice if the go code generated from the solidity allowed you not only to call the functions, but also to check how much gas they would use before you call them. SInce `*swap.ETHSwap` is part of the `rpcclient` struct, I think this is a good place to add that extended functionality.

martonp:
I haven't been able to find a case where the gas consumption for an operation has been updated in the history of Ethereum, but we could ask in a geth issue. In some more complicated solidity code, the code path might change based on the state of the contract, but this never happens in our case.

chappjc:
I was just going to point that out @martonp.  It's one of the reasons the `send` and `transfer` functions [aren't advised anymore](https://consensys.net/diligence/blog/2019/09/stop-using-soliditys-transfer-now).  https://chainsecurity.com/istanbul-hardfork-eips-increasing-gas-costs-and-more/

    SLOAD: gas cost changes from 200 to 800 gas;
    BALANCE: gas cost changes from 400 to 700 gas;
    EXTCODEHASH: gas cost changes from 400 to 700 gas.

So hard coded gas limits can go wrong.

client/asset/eth/rpcclient.go
JoeGruffins: If you want to add it in this particular file, you could also make it a separate function... so it's testable. You don't use `c` until the very end at `c.ec.EstimateGas`

JoeGruffins:
Ok. let's set up some max functions to get what are now constants. Could you make an issue @martonp ? You can just go crazy with this pr then, or make a new one. Will server will need something similar? Not sure.

client/asset/eth/rpcclient.go
```diff
@@ -256,6 +262,64 @@ func (c *rpcclient) initiate(txOpts *bind.TransactOpts, netID int64, refundTimes
...
+	}
+	refundTimestamp := time.Now().Unix()
+	var secret [32]byte
+	copy(secret[:], encode.RandomBytes(32))
+	secretHash := sha256.Sum256(secret[:])
+	// just an arbitrary address
+	participant := common.HexToAddress("345853e21b1d475582E71cC269124eD5e2dD3422")
+	data, err := parsedAbi.Pack("initiate", big.NewInt(refundTimestamp), secretHash, participant)
+	if err != nil {
+		return 0, nil
```
buck54321, line 279:
```Go
return 0, err
```

client/asset/eth/rpcclient.go
```diff
@@ -256,6 +262,64 @@ func (c *rpcclient) initiate(txOpts *bind.TransactOpts, netID int64, refundTimes
...
+// initiateGas checks the amount of gas that is used for a call to the redeem function.
+func (c *rpcclient) redeemGas(ctx context.Context, secret, secretHash [32]byte, participant *common.Address, contractAddress *common.Address) (uint64, error) {
+	parsedAbi, err := abi.JSON(strings.NewReader(swap.ETHSwapABI))
+	if err != nil {
+		return 0, err
+	}
+
+	data, err := parsedAbi.Pack("redeem", secret, secretHash)
+	if err != nil {
+		return 0, nil
```
buck54321, line 306:
```Go
return 0, err
```

buck54321 (changes requested):
A couple of things for now.

We should really route `c.ec *ethclient.Client` through an interface so that we can stub it out and do unit tests.

client/asset/eth/rpcclient.go
```diff
@@ -256,6 +262,64 @@ func (c *rpcclient) initiate(txOpts *bind.TransactOpts, netID int64, refundTimes
...
+	if err != nil {
+		return 0, err
+	}
+	refundTimestamp := time.Now().Unix()
+	var secret [32]byte
+	copy(secret[:], encode.RandomBytes(32))
+	secretHash := sha256.Sum256(secret[:])
+	// just an arbitrary address
+	participant := common.HexToAddress("345853e21b1d475582E71cC269124eD5e2dD3422")
+	data, err := parsedAbi.Pack("initiate", big.NewInt(refundTimestamp), secretHash, participant)
```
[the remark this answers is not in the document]
JoeGruffins: > It's a preloaded simnet address. 

I don't think addresses have any kind of encoding or checksum, I think they're just 20 bytes as well, the hash of the pubkey.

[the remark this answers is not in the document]
JoeGruffins: Maybe that gas error is because we are using with the harness and that address in the test actually does have funds. That would be a problem with making these functions usable for every start up to get estimates, if you actually need funds to make a swap to get the estimate.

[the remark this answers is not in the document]
martonp: There needs to be enough gas to pay for the value of the swap, so an account with an empty balance will not work. The gas price is set to 0, so that doesn't matter.

martonp:
I'll close this pull request and add this functionality to #1184.

PR #1187, opened 2021-09-01
client/db: export CandleCache candles

buck54321:
`CandleCache` is designed to be used by clients, who may want access to the cached candles directly. Currently the only way to access the data is with `WireCandles`, which is not always efficient. Both dcrdata's `ExchangeBot` and dexc will benefit from having the candles exported.


buck54321:
> LGTM. Any use for binSize exported too? The others are more internals though.

Might as well.

PR #1190, opened 2021-09-02
server/eth: Implement ValidateSecret.

JoeGruffins:
Part of #1154

### DIFF
--- a/server/asset/eth/eth_test.go
+++ b/server/asset/eth/eth_test.go
@@ -8,6 +8,7 @@ package eth
 import (
 	"bytes"
 	"context"
+	"crypto/sha256"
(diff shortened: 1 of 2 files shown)


server/asset/eth/eth_test.go
```diff
@@ -527,3 +528,30 @@ func TestContract(t *testing.T) {
 		}
 	}
 }
+
+func TestValidateSecret(t *testing.T) {
+	secret, blankHash := make([]byte, 32), make([]byte, 32)
+	copy(secret[:], encode.RandomBytes(32))
```
chappjc, line 534: I don't care because it's just a test, but no need to reslice `secret` in this copy.

PR #1192, opened 2021-09-02
server/comms: more fine grained websocket route limiter

chappjc:
Several of the websocket routes are assigned both sustained and burst
rate limits of their own.

Despite challenges of DDoS, address spoofing, and large IPv6 subnets,
limiters are created by remote IP, with uniqueness determined by
`dex.IPKey`, which masks interface bits of IPv6 addresses, and provides
a method to retrieve the 48-bit prefix (public topology).

An IP's limiter is removed from memory a minute after last conn
is disconnected.  This is done to prevent churning connections
to reset limits.

`rpcMaxConnsPerIP` is the maximum number of active websocket connections
allowed per IP, loopback excluded.

This adds code to help devise possible prefix aggregation according to
some heuristic based on the number of different subnets in use from
the same prefix. Just logging of such connections at present.

Use `NewIPKey` in `TestWSRateLimiter`.

### DIFF
--- a/server/comms/server.go
+++ b/server/comms/server.go
@@ -37,12 +37,27 @@ const (
 	// allowed.
 	rpcMaxClients = 10000
 
+	// rpcMaxConnsPerIP is the maximum number of active websocket connections
+	// allowed per IP, loopback excluded.
+	rpcMaxConnsPerIP = 8
+
 	// banishTime is the default duration of a client quarantine.
 	banishTime = time.Hour
 
-	// Per-ip rate limits for HTTP routes.
+	// Per-ip rate limits for market data API routes.
 	ipMaxRatePerSec = 1
 	ipMaxBurstSize  = 5
+
+	// Per-websocket-connection request limits. Rate should be a reasonable
+	// sustained rate, while burst should consider bulk reconnect operations.
+	wsRateStatus, wsBurstStatus     = 10, 500     // order_status and match_status (combined)
+	wsRateOrder, wsBurstOrder       = 5, 100      // market, limit, and cancel (combined)
+	wsRateInfo, wsBurstInfo         = 10, 200     // low-cost route limiter for: config, fee_rate  (combined)
+	wsRateBook, wsBurstBook         = 1, 100      // orderbook, assuming 100 markets subscribed max
+	wsRateRegister, wsBurstRegister = 1 / 60.0, 1 // register, rate.Every(time.Minute) but const
+	// The cumulative rates below would need to be less than sum of above to
+	// actually trip unless it is also applied to unspecified routes.
+	wsRateTotal, wsBurstTotal = 40, 1000
 )
 
 var (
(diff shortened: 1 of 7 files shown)


server/comms/server.go
```diff
@@ -158,30 +177,100 @@ type RPCConfig struct {
...
+
 	// The quarantine map maps IP addresses to a time in which the quarantine will
 	// be lifted.
-	banMtx      sync.RWMutex
-	quarantine  map[dex.IPKey]time.Time
-	dataEnabled uint32
+	banMtx     sync.RWMutex
+	quarantine map[dex.IPKey]time.Time
+
+	dataEnabled uint32 // atomic
```
JoeGruffins, line 266: Best to have atomics at the top of the struct? Something about 32 bit systems.

server/comms/server.go
```diff
@@ -380,6 +478,108 @@ func (s *Server) banish(ip dex.IPKey) {
...
+			// Try a prefix-aggregated limiter.
+			if prefix == nil {
+				return
+			}
+			wsLimiter = s.wsLimiters[*prefix]
+			if wsLimiter == nil {
+				return
+			}
+			ip = *prefix
+		*/
```
JoeGruffins, line 566: Intended to leave in?

Would an untracked limiter be expected? Are these the loopback addresses?

JoeGruffins (commented):
Would the best way to test these irl be to set up a server with an actual different ip or is there an easy way to fake a non-loopback address.

server/comms/server.go
chappjc: That particular atomic bug affects 64-bit data on 32-bit systems.  But it still may be good practice to put them first.  Will think about that.

server/comms/server.go
chappjc: This code would pair with the commented line in `wsLimiter` that switches the connection to a "prefix-aggregated" limiter.  https://github.com/decred/dcrdex/pull/1192/files#diff-5e901c2a3618ece1ff4d4750b55e2d616dded4726f9258f4043c994c6f41b390R503

I have this all commented because it's an incomplete approach, since when it does detect the need to aggregate, it leaves existing connections from the same prefix with their individual limiters.  It would probably need to replace their limiters with the aggregated one to be most effective.  The topic of IPv6 rate limiting is very much unsolved, but the best suggestion I've seen so far is probably https://serverfault.com/a/919324/190378

It's just experimental, so I'll remove this block.  Presently the only purpose of that `v6Prefixes` map is the `log.Debugf` line so we can learn in the field.

chappjc:
> Would the best way to test these irl be to set up a server with an actual different ip or is there an easy way to fake a non-loopback address.

Very very good question.  Aside from the unit tests where I crafted non-loopback addresses for the request's `RemoteAddr`, I simply commented the `IsLoopback` for interactive testing on my own machine.  I don't think it's really worth using two separate machines unless that's already in your workflow.

server/comms/server.go
chappjc: > Would an untracked limiter be expected? Are these the loopback addresses?

Right.

server/comms/server.go
JoeGruffins: Ok, 64 bit atomics on old 32 bit processors. I will remember this from now.

JoeGruffins (approved):

buck54321 (approved):

PR #1193, opened 2021-09-03
client/core: Check deposit addr in reconfigure.

### LINKED ISSUE #1136: cannot reconfigure wallet even though it's the same wallet
I've gotten the new error even when it's the same wallet:

2021-07-23 10:25:20.548 [ERR] WEB: reconfig error: new wallet cannot be used with current active trades: unable to change wallets **with active trades**. Trades with this wallet must match or be canceled in order to change settings

image

This one was DCR, but the issue could be Core logic.

JoeGruffins:
Rather than search for addresses in contracts, use the deposit address
    to determine if the new wallet is the same as the old one.

closes #1136

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1962,14 +1962,9 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg 
 	sameWallet := func() error {
 		hasActiveTrade := false
 		for _, dc := range c.dexConnections() {
-			for _, trade := range dc.trackedTrades() {
-				if !trade.isActive() {
-					continue
-				}
-				if dc.hasActiveAssetOrders(wallet.AssetID) {
-					hasActiveTrade = true
-					break
-				}
+			if dc.hasActiveAssetOrders(wallet.AssetID) {
+				hasActiveTrade = true
+				break
 			}
 		}
 		if hasActiveTrade {
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1980,7 +1980,6 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg 
 			}
 			if !owns {
 				return errors.New("new wallet does not own old deposit address")
-
 			}
 		}
 		return nil
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1966,8 +1966,7 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg 
 				if !trade.isActive() {
 					continue
 				}
-				waID := wallet.AssetID
-				if trade.wallets.toAsset.ID == waID || trade.wallets.fromAsset.ID == waID {
+				if dc.hasActiveAssetOrders(wallet.AssetID) {
 					hasActiveTrade = true
 					break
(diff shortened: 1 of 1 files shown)


client/core/core.go
```diff
@@ -2008,66 +2009,27 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg
...
-					"order to change settings")
+		}
+		if hasActiveTrade {
+			owns, err := wallet.OwnsAddress(oldDepositAddr)
+			if err != nil {
+				return err
+			}
+			if !owns {
+				return errors.New("new wallet does not own old deposit address")
+
```
JoeGruffins, line 2032: Extra new line.

martonp (approved):
Looks good.

client/core/core.go
```diff
@@ -2016,66 +2017,26 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg
...
-					// If we did not find a refund address,
-					// we cannot be sure that this is the
-					// same wallet.
-					//
-					// TODO: Implement a way to check that
-					// these accounts are the same or not.
-					maybeDifferentWallet = true
+				if trade.wallets.toAsset.ID == waID || trade.wallets.fromAsset.ID == waID {
+					hasActiveTrade = true
+					break
```
chappjc, line 2029: Without the `RefundAddress` work going on, we can reduce this with `hasActiveAssetOrders`:

```go
			if dc.hasActiveAssetOrders(wallet.AssetID) {
				hasActiveTrade = true
				break
			}
```

client/core/core.go
```diff
@@ -2016,66 +2017,26 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg
...
-					"trades with this wallet must match or be canceled in " +
-					"order to change settings")
+		}
+		if hasActiveTrade {
+			owns, err := wallet.OwnsAddress(oldDepositAddr)
+			if err != nil {
+				return err
+			}
+			if !owns {
+				return errors.New("new wallet does not own old deposit address")
```
chappjc, line 2039: Best to mention that this is a problem because there are active trades.

chappjc (changes requested):

client/core/core.go
JoeGruffins: Using.

client/core/core.go
```diff
@@ -1959,66 +1960,25 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg
...
-			}
-			return nil
-		}
+		hasActiveTrade := false
 		for _, dc := range c.dexConnections() {
-			maybeDifferentWallet := false
 			for _, trade := range dc.trackedTrades() {
 				if !trade.isActive() {
 					continue
 				}
```
chappjc, line 1968: I think you wanted to get rid of the `dc.trackedTrades` loop entirely, right?

JoeGruffins: Is it correct now?

chappjc: Yup pretty sure.  hasActiveAssetOrders loops through for us

chappjc (approved):

PR #1194, opened 2021-09-03
multi: Remove all io/ioutil usage.

### LINKED ISSUE #1165: migrate from ioutil to io and os packages
With Go 1.16 being the min supported Go version, we can stop using the ioutil package that was deprecated in that version.

https://golang.org/doc/go1.16#ioutil

JoeGruffins:
io/ioutil is deprecated in go 1.16. Replace methods with os and io
    methods.

closes #1165

### DIFF
--- a/client/db/types.go
+++ b/client/db/types.go
@@ -6,7 +6,7 @@ package db
 import (
 	"bytes"
 	"fmt"
-	"io/ioutil"
+	"os"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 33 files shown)


client/db/types.go
```diff
@@ -662,12 +662,12 @@ func decodeAccountBackup(b []byte) (*AccountBackup, error) {
 // Save persists an account backup to file.
 func (ab *AccountBackup) Save(path string) error {
 	backup := ab.Serialize()
-	return ioutil.WriteFile(path, backup, 0o600)
+	return os.WriteFile(path, backup, 0o600)
```
JoeGruffins, line 665: o for octal? I don't think I had ever seen this before.

buck54321: Yep. And the `o` is optional, so `0600` is also octal.

chappjc (approved):

PR #1201, opened 2021-09-09
client/core: discover account via connect instead of register

chappjc:
This is a prerequisite for multi-asset registration fee support.

Instead of using the `'register'` route to discover an existing account when restoring from an seed, use `'connect'` (`authDEX`).  Initially suggested https://github.com/decred/dcrdex/pull/1015#discussion_r666440449

This allows account discovery without having to request a registration address.  More importantly, when registration may be done with more than one asset, this avoids having to commit to a certain registration asset in order to discover an existing account.  This is because the `'register'` route will require an asset ID in a multi-asset registration scheme.

This eliminates the "in-process"/"incomplete" registration field and simplifies some logic in `Register`.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1067,9 +1067,6 @@ type Core struct {
 
 	sentCommitsMtx sync.Mutex
 	sentCommits    map[order.Commitment]chan struct{}
-
-	incompleteRegMtx sync.RWMutex
-	incompleteReg    *incompleteRegistration
 }
 
 // New is the constructor for a new Core.
@@ -2278,7 +2275,7 @@ func (c *Core) GetDEXConfig(dexAddr string, certI interface{}) (*Exchange, error
 // to register on a DEX, and Register may be called directly, although it requires
 // the expected fee amount as an additional input and it will pay the fee if the
 // account is not discovered and paid.
-func (c *Core) DiscoverAccount(dexAddr string, appPW []byte, certI interface{}) (xc *Exchange, paid bool, err error) {
+func (c *Core) DiscoverAccount(dexAddr string, appPW []byte, certI interface{}) (*Exchange, bool, error) {
 	if !c.IsInitialized() {
 		return nil, false, fmt.Errorf("cannot register DEX because app has not been initialized")
 	}
@@ -2293,13 +2290,23 @@ func (c *Core) DiscoverAccount(dexAddr string, appPW []byte, certI interface{}) 
 		return nil, false, codedError(passwordErr, err)
 	}
 
+	var ready bool
 	dc, err := c.tempDexConnection(host, certI)
-	if dc != nil {
-		// Stop (re)connect loop, which may be running even if err != nil.
+	if dc != nil { // (re)connect loop may be running even if err != nil
 		defer func() {
-			if !paid {
+			// Either disconnect or promote this connection.
+			if !ready {
 				dc.connMaster.Disconnect()
+				return
 			}
+			// Promote this dexConnection and start listening for messages.
+			atomic.StoreUint32(&dc.reportingConnects, 1)
+			c.wg.Add(1)
+			go c.listen(dc)
+
+			c.connMtx.Lock()
+			c.conns[dc.acct.host] = dc
+			c.connMtx.Unlock()
 		}()
 	}
 	if err != nil {
@@ -2311,16 +2318,59 @@ func (c *Core) DiscoverAccount(dexAddr string, appPW []byte, certI interface{}) 
 		return dc.exchangeInfo(), false, nil
 	}
 
-	incompleteReg, paid, err := c.registerHD(dc, crypter)
+	// Setup our account keys and attempt to authorize with the DEX.
+	creds := c.creds()
+	encKey, _, err := dc.acct.setupCryptoV2(creds, crypter)
 	if err != nil {
-		return nil, false, err
+		return dc.exchangeInfo(), false, newError(acctKeyErr, "setupCryptoV2 error: %v", err)
 	}
-	if !paid {
-		c.incompleteRegMtx.Lock()
-		c.incompleteReg = incompleteReg
-		c.incompleteRegMtx.Unlock()
+
+	// Instead of calling register, attempt authDEX.
+	err = c.authDEX(dc)
+	if err != nil {
+		var mErr *msgjson.Error
+		if errors.As(err, &mErr) && mErr.Code == msgjson.AccountNotFoundError {
+			return dc.exchangeInfo(), false, nil
+		}
+		return nil, false, newError(authErr, "unexpected authDEX error: %v", err)
 	}
-	return dc.exchangeInfo(), paid, nil
+	if dc.acct.isSuspended {
+		c.log.Infof("HD account key for %s was reported as suspended. Register to make a new one.", dc.acct.host)
+		return dc.exchangeInfo(), false, nil // just paid=false, no error
+	}
+	// NOTE: An older server may not report account suspension status, just
+	// score. If the user discovers they cannot place orders (and their score is
+	// bad), they should disable the account and call Register directly.
+
+	// Actual fee asset ID and coin are unknown, but paid.
+	dc.acct.isPaid = true
+	dc.acct.feeCoin = []byte("DUMMY COIN")
+
+	err = c.db.CreateAccount(&db.AccountInfo{
+		Host:         dc.acct.host,
+		Cert:         dc.acct.cert,
+		DEXPubKey:    dc.acct.dexPubKey,
+		EncKeyV2:     encKey,
+		LegacyEncKey: nil,
+		FeeCoin:      dc.acct.feeCoin,
+		// Paid set with AccountPaid below.
+	})
+	if err != nil {
+		return nil, false, fmt.Errorf("error saving restored account: %w", err)
+	}
+
+	err = c.db.AccountPaid(&db.AccountProof{
+		Host:  dc.acct.host,
+		Stamp: 54321,
+		Sig:   []byte("RECOVERY SIGNATURE"),
+	})
+	if err != nil {
+		return nil, false, fmt.Errorf("error marking recovered account as paid: %w", err)
+	}
+
+	ready = true // do not disconnect
+
+	return dc.exchangeInfo(), true, nil
 }
 
 // Register registers an account with a new DEX. If an error occurs while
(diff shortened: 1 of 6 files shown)


client/core/core.go
```diff
@@ -2356,16 +2363,52 @@ func (c *Core) DiscoverAccount(dexAddr string, appPW []byte, certI interface{})
...
+	// Setup our account keys and attempt to authorize with the DEX.
+	creds := c.creds()
+	encKey, _, err := dc.acct.setupCryptoV2(creds, crypter)
 	if err != nil {
-		return nil, false, err
+		return dc.exchangeInfo(), false, newError(acctKeyErr, "setupCryptoV2 error: %v", err)
+	}
+
+	// Instead of calling register, attempt authDEX.
+	err = c.authDEX(dc)
```
chappjc, line 2374: One difference with this approach is that `err == nil` when the account is suspended (primarily for purposes of negotiating active swaps).
Connect response has `score`, but client does not know ban score, which is the threshold to determine suspension.
Can the user manually disable the account in the /settings page?  I haven't tried this for accounts derived from the hd keys, only legacy.

chappjc: Just pushed a commit to attempt to address this https://github.com/decred/dcrdex/pull/1201/commits/44d208740bca3139040fcfbe3f8d9e8400e91d36

```
[INF] CORE: HD account key for 127.0.0.1:17273 was reported as suspended. Register to make a new one.
"POST http://127.0.0.3:5758/api/discoveracct HTTP/1.1" from 127.0.0.1:40232 - 200 968B in 1.247533373s
```

takes to Confirm screen with the markets table, then click confirm for it to call Register, which discovers the HD key won't work and makes a legacy account...

```
[INF] CORE: HD account key for 127.0.0.1:17273 was reported as suspended.
[INF] CORE: Generating a legacy-style key to register with 127.0.0.1:17273.
```

While this is the current behavior on master, it does make me a little concerned that the user might not realize it went the "legacy" account path here, meaning the account cannot be recovered by the seed, only export/import.  I think this just means we probably do want to pursue multiple child keys at the derivation path for a given DEX host.

client/core/core.go
```diff
@@ -2356,16 +2363,59 @@ func (c *Core) DiscoverAccount(dexAddr string, appPW []byte, certI interface{})
...
+		}
+		return nil, false, newError(authErr, "unexpected authDEX error: %v", err)
+	}
+	if dc.acct.isSuspended {
+		c.log.Infof("HD account key for %s was reported as suspended. Register to make a new one.", dc.acct.host)
+		return dc.exchangeInfo(), false, nil // just paid=false, no error
+	}
+	// NOTE: An older server may not report account suspension status, just
+	// score. If the user discovers they cannot place orders (and their score is
+	// bad), they should disable the account and call Register directly.
```
itswisdomagain, line 2388: Is it possible to detect from the place order error message that the account is suspended?

itswisdomagain (commented):
Code LGTM. Haven't tested though.

client/core/core.go
chappjc: Programatically... not reliably.  The server returns an error coded somewhat vaguely with `MarketNotRunningError` for that.

https://github.com/decred/dcrdex/blob/5e2ab91533d3fdf9ce774d104fbff844347baa1f/server/market/orderrouter.go#L215-L216

But the user does see a clear message in red *on the order confirmation form*, so yes.

chappjc: Regarding this note you're commenting on ("An older server may not report account suspension status"), to be fair, and older server won't be reporting the helpful error codes and data we expect from the register route either, since the initial HD keys PR modified the register request handler. https://github.com/decred/dcrdex/commit/b66a35f24215456763a16cee9c32875f61f9814b#diff-e7126c606b49bf8ca255171ae2f485e8d92f442af17e953700f49bcea7215507R62-R77
So it would seem to be no less reliable.

itswisdomagain: This means #1204 will also have to deal with (skip) HD accounts that are manually disabled if the user wants to re-register with the DEX.

itswisdomagain: Re #1204, do you want this to go in first before a fix for #1204?

chappjc: I'm not even sure about #1204 for 0.3 or not.
This PR certainly doesn't depend on it, at least there are no regressions with this PR compared to master.  So I think this should go in regardless.

There certainly are a lot of edge cases. I'm wondering if manually disabling an HD account should simply be disallowed.  If it is suspended, the software can disable it.  That way suspended <==> disabled for HD accounts.

itswisdomagain: If an HD account is suspended by a legacy server, can the software determine the suspension and thus auto-disable the account?

#1024 doesn't seem to me to be major work but important - prevent users from creating new DEX accounts with unrecoverable keys once they've started using a seed. Might be a hard to hit edge, but it wouldn't make sense to have and use a seed if a DEX account still cannot be recovered in certain situations.

chappjc: > If an HD account is suspended by a legacy server, can the software determine the suspension and thus auto-disable the account?

Seems like the answer is no, not really.

However, although a 0.2 patch release added the dex pub key to the config response, thus enabling HD accounts with that, it's not actually deployed anywhere, and doesn't need to be if we are now realizing that's creating a pickle for us.

It's also not unreasonable to require the server to be upgraded for 0.3 clients if we want to go that route.

itswisdomagain: Just remembered that legacy keys will still be used to support older servers. Did we agree to NOT do that; and compel clients to only register with servers that support HD account keys going forward?

chappjc: No, I'm just saying that's on the table if it untangles this mess most efficiently.  I'm not sure yet.

chappjc: Should we move on to address these questions in https://github.com/decred/dcrdex/pull/1210 then?

buck54321 (changes requested):

client/core/core.go
```diff
@@ -2356,16 +2363,59 @@ func (c *Core) DiscoverAccount(dexAddr string, appPW []byte, certI interface{})
...
+	// Instead of calling register, attempt authDEX.
+	err = c.authDEX(dc)
+	if err != nil {
+		var mErr *msgjson.Error
+		if errors.As(err, &mErr) && mErr.Code == msgjson.AccountNotFoundError {
+			return dc.exchangeInfo(), false, nil
+		}
+		return nil, false, newError(authErr, "unexpected authDEX error: %v", err)
+	}
+	if dc.acct.isSuspended {
```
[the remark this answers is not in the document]
chappjc: Hmm, tbh I'm not sure if the `account/{aid}/ban` admin endpoint works as indented anymore.  For a while now I think the server reevaluates the user's ban score and will automatically unban based on the score if needed.

I tested by getting tons of preimage misses.

dexc:

```
[INF] CORE: Connected to DEX server at 127.0.0.1:17273 but NOT listening for messages.
[INF] CORE: HD account key for 127.0.0.1:17273 was reported as suspended.
```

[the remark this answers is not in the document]
chappjc: Hmm, although that log I'm seeing there is from `register` not `DiscoverAccount`... looking at this closer now.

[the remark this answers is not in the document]
chappjc: Ah, I rebuild dexc and I'm getting the expected log from DiscoverAccount now

```
2021-09-15 12:24:22.444 [DBG] CORE: Authenticated connection to 127.0.0.1:17273, acct 51e231ab80dd963ac9456d508f07ea68593711ac1737bb08118b2aedcebe4be9, 0 active orders, 0 active matches, score 22 (suspended = true)
2021-09-15 12:24:22.444 [INF] CORE: HD account key for 127.0.0.1:17273 was reported as suspended. Register to make a new one.
```

You must have been encountering the automatic account restoration that server does based on ban score.

client/core/core.go
```diff
@@ -2356,16 +2363,59 @@ func (c *Core) DiscoverAccount(dexAddr string, appPW []byte, certI interface{})
...
+		c.log.Infof("HD account key for %s was reported as suspended. Register to make a new one.", dc.acct.host)
```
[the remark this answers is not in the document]
chappjc: Well, noting the frontend workflow, the paid=false return is handled programatically by the /register page by proceeding to the Confirm Registration dialog and then /api/register.  I'd actually consider making this log a Debugf.

Looking at how the previous revision of core.go called registerHD here instead, that has equivalent handling (log and return as appropriate for caller):

https://github.com/decred/dcrdex/blob/ad4321a869156c5340fb698fcdd828b751349117/client/core/core.go#L2614-L2615

If we consider next steps with https://github.com/decred/dcrdex/pull/1210, we'll then be looping through keys until we find either an unsuspended or unregistered account, so a notification for each suspended account discovered would not really be appropriate there.  The return args direct the caller.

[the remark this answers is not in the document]
buck54321: The difference is that in the old way and the future way, the user is not left holding a suspended account that they can't trade on. In this work they are. That requires a notification at the UI, imo.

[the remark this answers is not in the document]
chappjc: This isn't an asynchronous process though.  They get this information as the return value from the /api/discoveraccount call.  The result is that they do not have a paid account, and frontend takes them to register.  There are no notifications even shown at this point in the frontend workflow.

client/core/core.go
buck54321: > Just remembered that legacy keys will still be used to support older servers.

And imported accounts.

client/core/core.go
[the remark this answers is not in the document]
buck54321: Ah. I got confused. They're only left holding a suspended account on older servers, and we'd have no way of knowing that here.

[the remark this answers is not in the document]
chappjc: I've added a note (with no translation) in 098de27b, but I think it's unnecessary at best and probably confusing considering DiscoverAccount would only find a suspended account if they are reseeding, and they probably know from the previous dexc instance about the suspension.

Where I do agree a notification is warranted is the call to `authDEX` on login (initializeDEXConnections), so we can make an issue to add a proper notification with a translation there.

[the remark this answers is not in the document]
buck54321: If you want to remove the ntfn, please do. I think you're right.

buck54321 (approved):

JoeGruffins (approved):

PR #1202, opened 2021-09-09
multi: register with different assets, add BTC

chappjc:
This is built on https://github.com/decred/dcrdex/pull/1201 (first commit in this branch).

This allows registration fees to be paid in different assets, and implements BTC as an accepted asset with the server.

Tested and working (when I initially put this PR up).

**Next step for frontend is an asset selection dropdown on the confirmation dialog.**  Takers?

**dex/msgjson:** update `ConfigResult`, `Register`, `RegisterResult`.

`RegFees map[string]*FeeAsset json:"regFees"` is added to `ConfigResult`.

 **server/db:** external key derivation, add fee_asset

Add `FeeAsset uint32` to the `db.Account` type and accounts table.

HD key derivation and address generation is no longer the role of the database driver.

Add the `db.FeeKeyIndexer interface`, which retrieves and records an integer index for a `[]byte`, a serialized pubkey corresponding to a master pubkey's child and it's index. `DEXArchivist` is now required to be a `FeeKeyIndexer`.

`CreateAccount` now requires asset ID and reg address provided as inputs rather than it generating and returning a reg address.
`AccountRegAddr` now also returns asset ID with the address.

 **server/asset**: `Addresser` and `HDKeyIndexer` interfaces

Add fee address generators to DCR and BTC.

Implement `FeeCoin` (fee checker) in BTC.

**server/auth**: use multi-asset fee addressers and checkers

**server/cmd/dcrdex**: per-asset fee config

 **server/dex**: stitch together multi-asset fee addressers and checkers between DB, auth manager, and assets

`FeeCoiner` and `AddresserFactor` provide the helpers for the auth manager.

Set the `RegFees` field of the config response.

Start archivist before asset backends because the fee `Addressers` require it for key index storage and retrieval.

**client/db**:

Add `db.AccountInfo.FeeAssetID uint32`, and updates the DB encoding.

v2 `db.AccountInfo` encoding is updated, and fields reordered.
When earlier (v0/v1) blobs are loaded, `FeeAssetID` is loaded as 42.

**client/core**:

Add asset ID to `RegisterForm`.

Add `Exchange.RegFees map[string]*FeeAsset` field, and a `PendingFeeState` struct field that describes the asset and confs of the pending fee, removing the `RegConfirms` (`'confs'`) field.  Also remove the `ConfsRequired` (`'confsrequired'`) and `FeePending` (`'feePending'`) fields.

Add `feeAssetID uint32` to `dexAccount`.

In `Core`, use the `Exchange.RegFees` map, and deal with older servers providing the old config response.

Change `(*Core).GetFee` to `(*Core).GetRegFees`, which is unused.

 **client/rpcserver and dexcctl:** Kill `getfee`. Implement `getdexconfig` instead.

**client/webserver**: update, but use DCR only

Update `apiRegister` and `registrationForm` with asset ID.

Kill `apiGetFee`.

Modify js for `pendingFee` in `FeePaymentNote`, and removed `confs` and `confsrequired` from `Exchange` struct (`this.market.dex` in frontend).

Frontend still defaults to DCR/42, with no selection yet.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -5067,7 +5067,7 @@ func (c *Core) handleReconnect(host string) {
 		return
 	}
 
-	if !dc.acct.locked() && len(dc.acct.feeCoin) > 0 {
+	if !dc.acct.locked() && dc.acct.feePaid() {
 		err = c.authDEX(dc)
 		if err != nil {
 			c.log.Errorf("handleReconnect: Unable to authorize DEX at %s: %v", host, err)
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -9,7 +9,6 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"math"
 	"net"
 	"net/url"
 	"os"
@@ -43,13 +42,8 @@ import (
 )
 
 const (
-	keyParamsKey      = "keyParams"
-	conversionFactor  = 1e8
-	regFeeAssetSymbol = "dcr" // Hard-coded to Decred for registration fees, for now.
-
-	// regConfirmationsPaid is used to indicate completed registration to
-	// (*Core).setRegConfirms.
-	regConfirmationsPaid uint32 = math.MaxUint32
+	keyParamsKey     = "keyParams"
+	conversionFactor = 1e8
 
 	// tickCheckDivisions is how many times to tick trades per broadcast timeout
 	// interval. e.g. 12 min btimeout / 8 divisions = 90 sec between checks.
@@ -99,6 +93,11 @@ func (dt *dexTicker) Dur() time.Duration {
 	return time.Duration(atomic.LoadInt64(&dt.dur))
 }
 
+type pendingFeeState struct {
+	confs uint32
+	asset uint32
+}
+
 // dexConnection is the websocket connection and the DEX configuration.
 type dexConnection struct {
 	comms.WsConn
@@ -129,8 +128,8 @@ type dexConnection struct {
 	// connected is a best guess on the ws connection status.
 	connected uint32
 
-	regConfMtx  sync.RWMutex
-	regConfirms *uint32 // nil regConfirms means no pending registration.
+	pendingFeeMtx sync.RWMutex
+	pendingFee    *pendingFeeState
 
 	reportingConnects uint32
 }
@@ -151,6 +150,16 @@ func (dc *dexConnection) running(mkt string) bool {
 		return false // not found means not running
 	}
 	return mktCfg.Running()
+}
+
+func (dc *dexConnection) feeAsset(assetID uint32) *msgjson.FeeAsset {
+	dc.cfgMtx.RLock()
+	defer dc.cfgMtx.RUnlock()
+	if dc.cfg == nil {
+		return nil
+	}
+	symb := unbip(assetID)
+	return dc.cfg.RegFees[symb]
 }
 
 // marketConfig is the market's configuration, as returned by the server in the
@@ -227,25 +236,32 @@ func (dc *dexConnection) marketTrades(mktID string) []*trackedTrade {
 	return trades
 }
 
-// getRegConfirms returns the number of confirmations received for the
-// dex registration or nil if the registration is completed
-func (dc *dexConnection) getRegConfirms() *uint32 {
-	dc.regConfMtx.RLock()
-	defer dc.regConfMtx.RUnlock()
-	return dc.regConfirms
+func (dc *dexConnection) setPendingFee(asset, confs uint32) {
+	dc.pendingFeeMtx.Lock()
+	dc.pendingFee = &pendingFeeState{asset: asset, confs: confs}
+	dc.pendingFeeMtx.Unlock()
 }
 
-// setRegConfirms sets the number of confirmations received
-// for the dex registration
-func (dc *dexConnection) setRegConfirms(confs uint32) {
-	dc.regConfMtx.Lock()
-	defer dc.regConfMtx.Unlock()
-	if confs == regConfirmationsPaid {
-		// A nil regConfirms indicates that there is no pending registration.
-		dc.regConfirms = nil
-		return
+func (dc *dexConnection) clearPendingFee() {
+	dc.pendingFeeMtx.Lock()
+	dc.pendingFee = nil
+	dc.pendingFeeMtx.Unlock()
+}
+
+// getPendingFee returns the PendingFeeState for the dex registration or nil if
+// no registration fee is pending.
+func (dc *dexConnection) getPendingFee() *PendingFeeState {
+	dc.pendingFeeMtx.RLock()
+	defer dc.pendingFeeMtx.RUnlock()
+	pf := dc.pendingFee
+	if pf == nil {
+		return nil
 	}
-	dc.regConfirms = &confs
+	return &PendingFeeState{
+		AssetID: pf.asset,
+		Symbol:  unbip(pf.asset),
+		Confs:   pf.confs,
+	}
 }
 
 func (dc *dexConnection) exchangeInfo() *Exchange {
@@ -263,9 +279,8 @@ func (dc *dexConnection) exchangeInfo() *Exchange {
 		return &Exchange{
 			Host:       dc.acct.host,
 			AcctID:     acctID,
-			FeePending: dc.acct.feePending(),
 			Connected:  atomic.LoadUint32(&dc.connected) == 1,
-			// RegConfirms:   dc.getRegConfirms(), // could show, but we don't know required
+			PendingFee: dc.getPendingFee(),
 		}
 	}
 
@@ -276,20 +291,33 @@ func (dc *dexConnection) exchangeInfo() *Exchange {
 	}
 	dc.assetsMtx.RUnlock()
 
-	return &Exchange{
-		Host:          dc.acct.host,
-		AcctID:        acctID,
-		Markets:       dc.marketMap(),
-		Assets:        assets,
-		FeePending:    dc.acct.feePending(),
-		Connected:     atomic.LoadUint32(&dc.connected) == 1,
-		ConfsRequired: uint32(cfg.RegFeeConfirms),
-		RegConfirms:   dc.getRegConfirms(),
-		Fee: &FeeAsset{ // v0 is only DCR, but there will be a RegFees map[string]*FeeAsset field
+	feeAssets := make(map[string]*FeeAsset, len(cfg.RegFees))
+	for symb, asset := range cfg.RegFees {
+		feeAssets[symb] = &FeeAsset{
+			ID:    asset.ID,
+			Confs: asset.Confs,
+			Amt:   asset.Amt,
+		}
+	}
+	dcrAsset := feeAssets["dcr"]
+	if dcrAsset == nil { // should have happened in refreshServerConfig
+		dcrAsset = &FeeAsset{
 			ID:    42,
 			Amt:   cfg.Fee,
 			Confs: uint32(cfg.RegFeeConfirms),
-		},
+		}
+		feeAssets["dcr"] = dcrAsset
+	}
+
+	return &Exchange{
+		Host:       dc.acct.host,
+		AcctID:     acctID,
+		Markets:    dc.marketMap(),
+		Assets:     assets,
+		Connected:  atomic.LoadUint32(&dc.connected) == 1,
+		Fee:        dcrAsset,
+		RegFees:    feeAssets,
+		PendingFee: dc.getPendingFee(),
 	}
 }
 
@@ -1413,7 +1441,8 @@ func (c *Core) connectAndUpdateWallet(w *xcWallet) error {
 func (c *Core) connectedWallet(assetID uint32) (*xcWallet, error) {
 	wallet, exists := c.wallet(assetID)
 	if !exists {
-		return nil, newError(missingWalletErr, "no configured wallet found %s (%d)", unbip(assetID), assetID)
+		return nil, newError(missingWalletErr, "no configured wallet found for %s (%d)",
+			strings.ToUpper(unbip(assetID)), assetID)
 	}
 	if !wallet.connected() {
 		err := c.connectAndUpdateWallet(wallet)
@@ -1822,9 +1851,7 @@ func (c *Core) OpenWallet(assetID uint32, appPW []byte) error {
 		state.Symbol, balances.Available, balances.Locked, balances.ContractLocked,
 		state.Address)
 
-	if dcrID, _ := dex.BipSymbolID(regFeeAssetSymbol); assetID == dcrID {
-		go c.checkUnpaidFees(wallet)
-	}
+	go c.checkUnpaidFees(wallet)
 
 	c.notify(newWalletStateNote(state))
 
@@ -2208,18 +2235,6 @@ func (c *Core) isRegistered(host string) bool {
 	return len(dc.acct.encKey) > 0 // should be for all in conns map, but check
 }
 
-// GetFee creates a temporary connection to the specified DEX Server and fetches
-// the registration fee. The connection is closed after the fee is retrieved.
-// Returns an error if user is already registered to the DEX. A TLS certificate,
-// certI, can be provided as either a string filename, or []byte file contents.
-func (c *Core) GetFee(dexAddr string, certI interface{}) (uint64, error) {
-	xc, err := c.GetDEXConfig(dexAddr, certI)
-	if err != nil {
-		return 0, err
-	}
-	return xc.Fee.Amt, nil
-}
-
 // tempDexConnection creates an unauthenticated dexConnection. The caller must
 // dc.connMaster.Disconnect when done with the connection.
 func (c *Core) tempDexConnection(dexAddr string, certI interface{}) (*dexConnection, error) {
@@ -2345,6 +2360,7 @@ func (c *Core) DiscoverAccount(dexAddr string, appPW []byte, certI interface{}) 
 	// Actual fee asset ID and coin are unknown, but paid.
 	dc.acct.isPaid = true
 	dc.acct.feeCoin = []byte("DUMMY COIN")
+	dc.acct.feeAssetID = 42
 
 	err = c.db.CreateAccount(&db.AccountInfo{
 		Host:         dc.acct.host,
@@ -2352,6 +2368,7 @@ func (c *Core) DiscoverAccount(dexAddr string, appPW []byte, certI interface{}) 
 		DEXPubKey:    dc.acct.dexPubKey,
 		EncKeyV2:     encKey,
 		LegacyEncKey: nil,
+		FeeAssetID:   dc.acct.feeAssetID,
 		FeeCoin:      dc.acct.feeCoin,
 		// Paid set with AccountPaid below.
 	})
@@ -2403,7 +2420,13 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
 		return nil, newError(dupeDEXErr, "already registered at %s", form.Addr)
 	}
 
-	regFeeAssetID, _ := dex.BipSymbolID(regFeeAssetSymbol)
+	// Default to using DCR unless specified.
+	regFeeAssetID := uint32(42)
+	if form.Asset != nil {
+		regFeeAssetID = *form.Asset
+	}
+	regFeeAssetSymbol := dex.BipIDSymbol(regFeeAssetID)
+
 	wallet, err := c.connectedWallet(regFeeAssetID)
 	if err != nil {
 		// Wrap the error from connectedWallet, a core.Error coded as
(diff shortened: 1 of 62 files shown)


client/core/core.go
```diff
@@ -2480,6 +2503,20 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
...
+	// Ensure this DEX supports this asset for registration fees, and get the
+	// required confirmations and fee amount.
+	dc.cfgMtx.RLock()
+	feeAsset, supported := dc.cfg.RegFees[regFeeAssetSymbol]
+	dc.cfgMtx.RUnlock()
+	if !supported || feeAsset == nil {
+		return nil, newError(assetSupportErr, "dex server does not accept registration fees in asset %q", regFeeAssetSymbol)
+	}
+	reqConfs := feeAsset.Confs
+	if reqConfs > 1000 { // basic sanity check on required confirms
```
JoeGruffins, line 2515: Could be less, and per asset.

client/core/core.go
```diff
@@ -43,13 +42,8 @@ import (
...
 const (
-	keyParamsKey      = "keyParams"
-	conversionFactor  = 1e8
-	regFeeAssetSymbol = "dcr" // Hard-coded to Decred for registration fees, for now.
-
-	// regConfirmationsPaid is used to indicate completed registration to
-	// (*Core).setRegConfirms.
-	regConfirmationsPaid uint32 = math.MaxUint32
+	keyParamsKey     = "keyParams"
+	conversionFactor = 1e8
```
JoeGruffins, line 46: Unrelated, but I believe this is a small problem for eth, and possible other coins. We will do calculations in gwei, which is 1e9 wei, and then one eth is 1e9 of that, so a different conversion factor in order to speak in terms of one eth. I think just used for notifications though.

There are probably more instances of this in the webserver js.

JoeGruffins (commented):
That was more changes than I was expecting. Good job.

server/asset/btc/addresser.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, fmt.Errorf("key is for the wrong network: %s", chainParams.Name)
+	}
+	if key.IsPrivate() {
+		return nil, errors.New("private key provided")
+	}
+	external, err := key.Child(0)
+	if err != nil {
+		return nil, err
+	}
+	pubkey, _ := external.ECPubKey()
```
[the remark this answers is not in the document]
JoeGruffins: Was removed with last commit.

chappjc:
> That was more changes than I was expecting

It's unfortunate, I know.  :(  The server DB functionality prior to this was convenient but really a problem for anything but DCR addresses.  It also wasn't really a concern of the DB to generate addresses anyway.

client/core/core.go
chappjc: I might want to eliminate this new check.  It is asset-specific as you said, and it's kind out of scope for this PR, at least to come up with a per-asset limit.

client/webserver/site/src/js/app.js
```diff
@@ -417,17 +417,18 @@ export default class Application {
...
       // setting the null value in the 'confs' field indicates that the fee
       // payment was completed
-      dex.confs = null
+      dex.pendingFee = null
       return
     }
 
-    dex.confs = confs
+    const symbol = this.assets[asset].symbol
+    dex.pendingFee = { confs, asset, symbol }
```
chappjc, line 431: NOTE: these JS changes with `pendingFee` in particular are not super robust yet.  You can get a JS error if you register via BTC with `dexcctl` while the page is opened.  But the current interactive registration workflow with the browser only is smooth and without error.

server/auth/registrar.go
```diff
@@ -211,30 +293,49 @@ func (auth *AuthManager) handleNotifyFee(conn comms.Link, msg *msgjson.Message)
...
-				log.Warnf("error sending notifyfee result to link: %v", err)
-			}
-		}
-	}()
-	if val < auth.regFee {
+	if confs < int64(feeAsset.confs) {
+		return wait.TryAgain
+	}
+
+	if val < feeAsset.amt {
```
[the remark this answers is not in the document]
chappjc: I had a similar though in this function, but `val` comes out of `checkFee` too.

server/asset/btc/addresser.go
```diff
@@ -0,0 +1,85 @@
...
+			continue
+		case err == nil:
+			return child, i
+		default:
+			panic(err.Error()) // bad xkey or something
+		}
+	}
+}
+
+// NextAddress retrieves the pkh address for the next pubkey.
```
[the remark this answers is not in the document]
chappjc: Good call.  I'm gonna make them thread-safe rather than require the caller to manage it.

server/asset/btc/btc.go
```diff
@@ -352,6 +360,82 @@ func (btc *Backend) VerifyUnspentCoin(_ context.Context, coinID []byte) error {
 	return nil
 }
 
+// FeeCoin gets the recipient address, value, and confirmations of a transaction
+// output encoded by the given coinID. A non-nil error is returned if the
+// output's pubkey script is not a non-stake P2PKH requiring a single
```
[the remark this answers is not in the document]
chappjc: Good catch.  Also the docs should say P2**W**PKH

client/core/core.go
chappjc: I was sure we had an issue for this but I cannot find it.  Whenever I see this `conversionFactor = 1e8` I just pretend I didn't see it and run the other way.

server/asset/dcr/addresser.go
```diff
@@ -0,0 +1,90 @@
...
+// NextAddress retrieves the pkh address for the next pubkey.
+func (ap *AddressDeriver) NextAddress() string {
+	child, i := getChild(ap.xpub, ap.next)
+	hash := dcrutil.Hash160(child.SerializedPubKey())
+	ap.storeIdx(i) // not necessarily ap.next++
+	ap.next = i + 1
+
+	addr, err := stdaddr.NewAddressPubKeyHashEcdsaSecp256k1V0(hash, ap.addrParams)
+	if err != nil {
+		panic(err.Error()) // only errors if hash is not 20 bytes (ripemd160.Size), but we guarantee it
```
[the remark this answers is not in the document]
chappjc: I think I changed my stance on panics half way through this PR.  I've made things more consistent in the commit just pushed for your review.

chappjc:
Not re-tested with last commit.  The btcutil update looks clean, but I can't be sure if it affected any of the btcd packages.  See https://github.com/decred/dcrdex/issues/1209

dex/networks/btc/script.go
```diff
@@ -372,6 +372,31 @@ func IsDust(txOut *wire.TxOut, minRelayTxFee uint64) bool {
 	return txOut.Value/(3*int64(totalSize)) < int64(minRelayTxFee)
 }
 
+// ExtractScriptData extracts script type, addresses, and required signature
+// count from a pkScript. Non-standard scripts are not necessarily an error;
+// non-nil errors are only returned if the script cannot be parsed. See also
+// InputInfo for additional signature script size data
+func ExtractScriptData(script []byte, chainParams *chaincfg.Params) (BTCScriptType, []string, int, error) {
```
chappjc, line 379: stolen from dex/networks/dcr

server/asset/btc/addresser.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	if key.IsPrivate() {
+		return nil, errors.New("private key provided")
+	}
+	external, err := key.Child(0)
+	if err != nil {
+		return nil, err
+	}
+	pubkey, _ := external.ECPubKey()
+	pkb := pubkey.SerializeCompressed()
```
chappjc: To anyone reviewing just this commit rather than the entire PR, this was changed in the "FIX" commit to go back to indexing on the xpub string.

server/asset/btc/addresser.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// NextAddress retrieves the pkh address for the next pubkey.
+func (ap *AddressDeriver) NextAddress() string {
+	child, i := getChild(ap.xpub, ap.next)
+	addr, _ := child.Address(ap.params) // cannot error
+	ap.storeIdx(i)                      // not necessarily ap.next++
+	ap.next = i + 1
+	hash := addr.Hash160()
+	addrP2WPKH, _ := btcutil.NewAddressWitnessPubKeyHash(hash[:], ap.params) // cannot error when hash is 20 bytes
```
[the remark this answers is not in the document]
chappjc: Last thing I'm still thinking on is if we really want to just silently derive p2**w**pkh addresses regardless of the extended pubkey encoding.  That is, xpub/tpub actually signify a derivation path for p2pkh, while zpub/vpub signify p2wpkh.
https://github.com/satoshilabs/slips/blob/master/slip-0132.md#registered-hd-version-bytes
The owner of the key will own the private keys regardless of the address format, but if they provide an xpub/tpub, they may not be set up with a segwit wallet that will recognize payments to the p2wpkh addresses.

A related consideration is that it's possible there's another coin that uses identical version bytes, but that would be one messed up coin, plus the only "problem" there would be the likelihood of creating gaps in the used addresses, which is already a thing to handle.

vctt94 (commented):
Tested this PR with dexcctl and it I was able to fully register, just fine. 

For making it possible to accept multi fee payments in the frontend, we will need to change the currently auth flow, which is:

set app password -> register dcr wallet.

Chatting with @chappjc we think it might be a good idea to change that to:

set app password -> choose wallet for paying the fee -> going to setting wallet.

What are your thoughts on that @buck54321 ?

chappjc:
To expand on the /register flow change, it could go like this:
- set app pass dialog
- add a dex dialog
- (it fetches the specified dex config and does discoveracccount)
- show the confirm registration dialog, **but** with a table or something showing the supported fee assets

image

So the above dialog would also show the supported reg asset and their amounts. And next to each one of the reg assets, there would be a button to "Add a {asset symbol} wallet" (or just a checkmark if the wallet is already configured).

chappjc:
Something that looks like this perhaps:

image

server/db/driver/pg/accounts.go
```diff
@@ -158,37 +144,62 @@ func (a *Archiver) PayAccount(aid account.AccountID, coinID []byte) error {
...
 	}
+	return child, nil
+}
 
-	return addr.String(), nil
+// SetFeeKeyIndex records the child index for an xpub. An error is returned
+// unless exactly 1 row is updated.
+func (a *Archiver) SetFeeKeyIndex(idx uint32, xpub string) error {
+	keyHash := dcrutil.Hash160([]byte(xpub))
+	log.Debugf("Recording new index %d for xpub %s (%x)", idx, xpub, keyHash)
```
JoeGruffins, line 189: How do you feel about truncating and just showing the first 20ish letters of the pubkey base58. For one it is very long and I doubt all needed to compare two pubkeys
```
[DBG] DB: Recording new index 1 for xpub tpubD6NzVbkrYhZ4XgiXtGrdW5XDAPFCL9h7we1vwNCpn8tGbBcgfVYjXyhWo4E1xkh56hjod1RhGjxbaTLV3X4FyWuejifB9jusQ46QzG87VKp (450907c35b4fa2c399bde54b1511aa48f5e4c3da)
```
And another, maybe it's not ideal to have it all over the logs. I would say it is sensitive info. Not super sensitive, but could be used in some sort of 300IQ attack.

JoeGruffins (approved):
Working great on simnet. Could you add a 42 in the simnet-setup.sh:
```
echo registering with DEX
./dexcctl -p abc --simnet register 127.0.0.1:17273 100000000 42 ~/dextest/dcrdex/rpc.cert
```

Or even `20000000 0`

server/db/driver/pg/accounts.go
chappjc: Got a point there.  I had figured if you have the control over the db you have control over the logger, but risk leaking data on yet another channel.  It's only public key, but still...

chappjc:
I squashed the fix commits into the individual commits where they belong, assuming reviewers will go through the commits individually.

server/dex/dex.go
```diff
@@ -139,19 +138,26 @@ type configResponse struct {
...
 		BinSizes:         apidata.BinSizes,
 		DEXPubKey:        cfg.DEXPrivKey.PubKey().SerializeCompressed(),
+		RegFees:          regAssets,
 	}
 
+	// Set the deprecated DCR fee fields.
+	dcrAsset := regAssets["dcr"]
+	if dcrAsset == nil {
+		return nil, fmt.Errorf("DCR is required as a fee asset for backward compatibility")
+	}
```
buck54321, line 157: Do this before `configMsg := &msgjson.ConfigResult{`, and you can just set the two fields in the struct literal.

buck54321 (changes requested):
First pass. Just some minor stuff. Tested up and down and saw no issues.

server/auth/auth.go
```diff
@@ -180,19 +181,25 @@ func (client *clientInfo) respHandler(id uint64) *respHandler {
...
 	storage        Storage
 	signer         Signer
-	regFee         uint64
 	checkFee       FeeChecker
-	feeConfs       int64
 	miaUserTimeout time.Duration
 	unbookFun      func(account.AccountID)
 
+	feeAddress func(assetID uint32) string
+	feeAssets  map[uint32]*feeAsset
```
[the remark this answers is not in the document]
chappjc: Good question.  I went back and forth on that too.  My reasoning isn't terribly solid, so I think I'll just switch to the msgjson type.

server/asset/btc/addresser.go
```diff
@@ -0,0 +1,109 @@
...
+	xpub     *hdkeychain.ExtendedKey
+
+	mtx  sync.Mutex
+	next uint32 // next child index to generate
+}
+
+// NewAddressDeriver creates a new AddressDeriver for the provided extended
+// public key, HDKeyIndexer, and network parameters. Note that if the source
+// wallet has accounts, the extended key should be for an account.
+func NewAddressDeriver(xpub string, keyIndexer asset.HDKeyIndexer, chainParams *chaincfg.Params) (*AddressDeriver, error) {
```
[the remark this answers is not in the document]
chappjc: I was actually making a point _not_ to have the `AddressDeriver` be a `Backend` method because there's no need to have a connection to a node to derive addresses.  Setting up a backend if you just want an `AddressDeriver` would be an unnecessary burden.

In fact, I was considering going ever further and separating it from the `Driver`, but it would be more code to create them by asset symbol.  Hence this convenience method: https://github.com/decred/dcrdex/pull/1202/commits/528b73953157edbd828937e3d190d8cccf1cf898#diff-51dc1aa092977ea75e0a2e2e2bc5263254edd21ca3f653b0a27035c5c667c829R62

If I were writing some other package or app that needed to derive Bitcoin addresses specifically, I'd definitely just import the `server/asset/btc` package and use `NewAddressDeriver`.

I completely acknowledge that the `fmt.Printf`s are sucky though.  They were mostly there for debugging, so I wouldn't miss them if they were just removed.

Does that makes sense?  I would find it really unfortunate to make every type and function in the server/asset/{symb} packages require a configured Backend to be useable.

client/core/core_test.go
```diff
@@ -1734,6 +1704,7 @@ func testRegister(t *testing.T, legacyKeys bool) {
 	if feeNote.Severity() != db.Success {
 		t.Fatalf("fee payment error notification: %s: %s", feeNote.Subject(), feeNote.Details())
 	}
+	getFeeAndBalanceNote() // account registered
```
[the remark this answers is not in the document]
chappjc: Right, I couldn't get subsequent parts of the test to work without cleaning up after these.

client/core/core.go
```diff
@@ -5029,7 +5112,7 @@ func (c *Core) handleReconnect(host string) {
 		return
 	}
 
-	if !dc.acct.locked() {
+	if !dc.acct.locked() && len(dc.acct.feeCoin) > 0 {
```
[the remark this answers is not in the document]
chappjc: Hmm, just got sloppy I guess.  `isPaid` is the right one though I think.

server/asset/btc/btc.go
```diff
@@ -352,6 +360,82 @@ func (btc *Backend) VerifyUnspentCoin(_ context.Context, coinID []byte) error {
...
+	Addresses    []string
+	SigsRequired int
+	ScriptType   dexbtc.BTCScriptType
+}
+
+// OutputSummary gets transaction output data, including recipient addresses,
+// value, script type, and number of required signatures, plus the current
+// confirmations of a transaction output. If the output does not exist, an error
+// will be returned. Non-standard scripts are not an error.
+func (btc *Backend) OutputSummary(txHash *chainhash.Hash, vout uint32) (txOut *TxOutData, confs int64, err error) {
```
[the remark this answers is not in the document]
chappjc: You and @JoeGruffins seem to agree about this (https://github.com/decred/dcrdex/pull/1202#discussion_r707060104), so I'll go along and unexport here and in dcr, but I did like having the method available to other packages and didn't feel like it detracted from the API significantly.

server/asset/btc/addresser.go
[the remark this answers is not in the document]
chappjc: I think you'll find commit https://github.com/decred/dcrdex/pull/1202/commits/e8f9e57382b7555a22b071a50894ae02f013ee91 more palatable.

Instead of being lazy and type-asserting the `Backend` to see if it is also `NewAddresser`, I've updated the `Driver` implementations so the addressers can be generated by asset symbol just like the `Backend`s.  This way no `Backend` is required to use a `NewAddresser` (it doesn't need a node), and one can be created either via (1) `asset.NewAddresser(symbol, ...` **or** (2) directly if you have the specific asset's package in mind like `server/asset/dcr.NewAddresser`

server/dex/dex.go
```diff
@@ -664,6 +704,33 @@ func NewDEX(ctx context.Context, cfg *DexConf) (*DEX, error) {
...
+	idx, err := ki.FeeKeyIndex(xpub)
+	if err == nil {
+		log.Infof("Resuming fee keys at index %d for extended pubkey %.40s...", idx, xpub)
+		return idx, nil
+	}
+	if !db.IsErrUnknownFeeKey(err) {
+		return 0, err
+	}
+	log.Infof("Creating new fee key entry for extended pubkey %.40s...", xpub)
+	return ki.CreateFeeKeyEntryFromPubKey(xpub)
```
buck54321, line 724: Why not handle this right in `(*Archiver).FeeKeyIndex`? You could seemingly preclude the need for `hdKeyIndexer` altogether if you fiddled with the method names for the interfaces.

server/cmd/dcrdex/sample-markets.json
```diff
@@ -23,7 +23,10 @@
             "network": "mainnet",
             "maxFeeRate": 10,
             "swapConf": 4,
-            "configPath": "/home/dcrd/.dcrd/dcrd.conf"
+            "configPath": "/home/dcrd/.dcrd/dcrd.conf",
+            "regConfs": 2,
+            "regFee": 0.1,
```
buck54321, line 28: `regFee` should be in atoms. Type is `uint64`.

server/dex/dex.go
```diff
@@ -228,6 +233,12 @@ func (dm *DEX) handleDEXConfig(interface{}) (interface{}, error) {
 	return dm.configResp.configEnc, nil
 }
 
+// FeeCoiner describes a type that can check a transaction output, namely a fee
+// payment, for a particular asset.
+type FeeCoiner interface {
+	FeeCoin(coinID []byte) (addr string, val uint64, confs int64, err error)
```
buck54321, line 239: Not really returning a `Coin`, but I don't care that much.

buck54321 (changes requested):

chappjc:
Thank you for those two very meticulous reviews.  Excellent comments.  I really appreciate it.

server/db/driver/pg/accounts.go
```diff
@@ -11,9 +11,7 @@ import (
 	"decred.org/dcrdex/server/account"
 	"decred.org/dcrdex/server/db"
 	"decred.org/dcrdex/server/db/driver/pg/internal"
-	"github.com/decred/dcrd/chaincfg/chainhash"
-	"github.com/decred/dcrd/hdkeychain/v3"
-	"github.com/decred/dcrd/txscript/v4/stdaddr"
+	"github.com/decred/dcrd/dcrutil/v4" // deprecated usage, move to "crypto/sha256"
```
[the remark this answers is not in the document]
chappjc: Meant to remove this comment, but when removing all the decred-specific stuff from the db packages I realized that the hash function in use for the tables is Decred's hash160, when something more ubiquitous and part of the standard libraries like crypto/sha256 would be better and easier to maintain.

server/dex/dex.go
chappjc: I didn't like the name either when doing this but I stuck to the common [interface naming guidelines](https://golang.org/doc/effective_go#interface-names) here for some reason.

Rightly, we often veer off that guidance when it makes a better interface name, and I think this is such a case.  Maybe `FeeChecker` instead?  Not sure if I want to rename the actual `FeeCoin` methods too, but those could become `CheckFee` or since the methods are actually performing generic checks of a coin, `CheckCoin` methods for a `CoinChecker` interface?

buck54321: I do think that `CheckCoin` is more descriptive, but I really really don't care that much.

server/auth/auth.go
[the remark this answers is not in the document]
buck54321: Ah. I see the reasoning now. It's unfortunate to have the string map indices in e.g. `msgjson.ConfigResult.RegFees`, but it makes sense because JSON, and any alternative is uglier. It ultimately won't be an issue, since the symbol is unique enough here.

buck54321 (approved):

PR #1205, opened 2021-09-10
eth/testing: Ensure node shutdown happens.

JoeGruffins:
os.Exit does not respect defer statements. Run tests in a separate
    function for intended node teardown to happen.

was part of #1162

### DIFF
--- a/server/asset/eth/rpcclient_harness_test.go
+++ b/server/asset/eth/rpcclient_harness_test.go
@@ -23,17 +23,24 @@ var (
 )
 
 func TestMain(m *testing.M) {
-	var cancel context.CancelFunc
-	ctx, cancel = context.WithCancel(context.Background())
-	defer func() {
-		cancel()
-		ethClient.shutdown()
-	}()
-	if err := ethClient.connect(ctx, ipc); err != nil {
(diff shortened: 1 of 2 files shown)


server/asset/eth/rpcclient_harness_test.go
```diff
@@ -23,17 +23,24 @@ var (
...
+			return 1, fmt.Errorf("Connect error: %v\n", err)
+		}
+		return m.Run(), nil
 	}
-	os.Exit(m.Run())
+	exitCode, err := run()
+	if err != nil {
+		fmt.Println(err)
+	}
+	os.Exit(exitCode)
```
chappjc, line 43: The Go developers I think realized this awkward issue and recently made it so `TestMain` no longer needs to call `os.Exit` in some cases.  https://golang.org/doc/go1.15#testing

> A TestMain function is no longer required to call os.Exit. If a TestMain function returns, the test binary will call os.Exit with the value returned by m.Run. 

Sadly, they didn't create a `(*testing.M).Fail` method to call before an early return during setup (prior to `m.Run()`), which would have solved this issue without a closure.

chappjc (approved):

PR #1206, opened 2021-09-10
dex/networks/btc: recognize rpcconnect

chappjc:
The **bitcoin-cli** setting, `rpcconnect=<IP>`, should override any host from
`rpcbind` or the default.

`checkConfig` is already overwriting `RPCBind` with the selected host and
port from a combination of defaults and `RPCPort`, and now `RPCConnect`,
so also patch `RPCPort` if the used port is taken from `RPCBind`.
This consistent approach is now documented on the `Config` struct.


dex/networks/btc/config.go
```diff
@@ -105,10 +114,23 @@ func checkConfig(cfg *Config, name string, network dex.Network, ports NetPorts)
 			}
 			if p != "" {
 				port = p
+				// Patch cfg.RPCPort for consistency with RPCBind.
+				if rpcPort, err := strconv.Atoi(port); err != nil {
+					// port was already parsed as an int by SplitHostPort, but
+					// be cautious.
+					cfg.RPCPort = rpcPort
+				}
```
chappjc, line 122: This is the only questionable/unnecessary part of this I think, but it does seem good to generate a consistent struct.

JoeGruffins (approved):

buck54321 (approved):

PR #1208, opened 2021-09-13
app: add historical market data chart

buck54321:
Adds a candlestick chart to the markets view.

server/db: Candle types moved from **/server/db** to **/dex/candles**.

core: Synchronize candlesticks and serve via `BookFeed`. `BookFeed` is changed from a concrete type to an interface, and changes are made to improve the design of the underlying core type. bookie now handles a map of `*candles.Cache`.

app: `DepthChart` now inherits a new parent class, `Chart`, as does the new `CandleChart`.

webserver: New `wsLoadCandles` method handles the `'loadcandles'` websocket request by finding (or creating new) the `BookFeed` and calling it's `Candles` method. This causes candlestick data to be sent on the `BookFeeds` update channel, and subsequently updates for that bin width.

Screenshot from 2021-09-13 08-24-08


client/webserver/site/src/js/markets.js
```diff
@@ -98,13 +105,24 @@ export default class MarketsPage extends BasePage {
...
+    const depthReporters = {
+      click: p => { this.reportDepthClick(p) },
+      volume: d => { this.reportDepthVolume(d) },
+      mouse: d => { this.reportDepthMouse(d) },
+      zoom: z => { this.reportDepthZoom(z) }
+    }
+    this.depthChart = new DepthChart(page.marketChart, depthReporters, State.fetch(depthZoomKey))
+
+    const candleReporters = {
+      mouse: c => { this.reportMouseCandle(c) }
```
chappjc, line 118: When I go from depth -> candles -> depth, the mouse position seems to be reported incorrectly.  The vertical line as well as the clicked locations are shifted to the right.

https://user-images.githubusercontent.com/9373513/133164351-fba24021-24cc-40c1-9654-7c3b46cb41eb.mp4

chappjc (changes requested):

chappjc:
Oh, we could put this in 0.3 if the issues are resolvable and it's smooth for other testers.  But the other PRs tagged for 0.3 take precedence, so if we can't get this until 0.4 that's fine because we are gonna make that happen super quick after 0.3 anyway.

client/core/bookie.go
```diff
@@ -26,29 +27,78 @@ var (
...
+	// feedMtx.
+	c      chan *BookUpdate
+	bookie *bookie
+	id     uint32
+}
+
+// Next returns the channel for receiving updates.
+func (f *bookFeed) Next() <-chan *BookUpdate {
+	return f.c
 }
```
JoeGruffins, line 50: Is it expected that the bookies feedMtx be held when calling this?

JoeGruffins (approved):
Working well.

There are a few things that could be better in the UI imo, but can always be added later.

For a new market, there is just one big candle.
image
And as it progresses.
image

Would be better to put some constant number of divisions and then these would just fill up from the left.

Also, when going to a rate that is higher than what the graph currently shows, it takes some time to show the top. Would be nice if it was more reactive there. And maybe a zoom?

Also, for rates under 0, like 0.1 to 0.00001 I don't see any change. I would think it would drop further. For example, here I bought at 10 btc/dcr, then I dropped it down to .000001 bt/dcr. Shouldn't it go down further, or is my brain too flat?

image

client/core/bookie.go
chappjc: Just for the sends and close, I believe.

chappjc: On this topic, the safest pattern with channels imo is to have the same goroutine that sends on the channel also be responsible for closing it.  Doing that means having a dedicated send loop and when the loop breaks it closes the channel to signal to receivers.

Not that I have a clear picture of how the bookie stuff would ideally look, but these pieces have always been pretty difficult to reason about.

chappjc (approved):
g2g pending any updates for Joe's review

client/websocket/websocket.go
```diff
@@ -264,35 +277,71 @@ out:
...
+func wsLoadCandles(s *Server, cl *wsClient, msg *msgjson.Message) *msgjson.Error {
+	req := new(candlesLoad)
+	err := json.Unmarshal(msg.Payload, req)
+	if err != nil {
+		errMsg := fmt.Sprintf("error unmarshalling candlesLoad payload: %v", err)
+		s.log.Errorf(errMsg)
+		return msgjson.NewError(msgjson.RPCInternal, errMsg)
+	}
+	cl.feedMtx.RLock()
+	feed := cl.feed
```
chappjc, line 332: After taking another look, I'm wondering if there's anyway a `"loadcandles"` could happen before `"loadmarket"`, which makes the feed.  Should `wsLoadCandles` either error if not already subscribed (with a non-nil feed) or even do a `loadMarket` with the embedded `candlesLoad.marketLoad`?  Just error maybe?

chappjc:
Merged, but have a look at [this comment](https://github.com/decred/dcrdex/pull/1208#discussion_r712631686) @buck54321 and anything from Joe's review that you'd like to sneak in somewhere else.

PR #1210, opened 2021-09-14
client/core: support multiple HD accounts for same DEX

### LINKED ISSUE #1204: core: extend deriation path for dex to address suspended accounts
In https://github.com/decred/dcrdex/pull/1201, I was dealing with the issue of suspended accounts and registration with a legacy key fallback.  I think this should be changed, as has been suggested previously, to derive new keys that branch at the DEX host's path.

Demonstrated on that PR, but the behavior on master is the same:

```
[INF] CORE: HD account key for 127.0.0.1:17273 was reported as suspended. Register to make a new one.
"POST http://127.0.0.3:5758/api/discoveracct HTTP/1.1" from 127.0.0.1:40232 - 200 968B in 1.247533373s
```

takes to Confirm screen with the markets table, then click confirm for it to call Register, which discovers the HD key won't work and makes a legacy account...

```
[INF] CORE: HD account key for 127.0.0.1:17273 was reported as suspended.
[INF] CORE: Generating a legacy-style key to register with 127.0.0.1:17273.
```

While this is the current behavior on master, it does make me a little concerned that the user might not realize it went the "legacy" account path here, meaning the account cannot be recovered by the seed, only export/import.  I think this just means we probably do want to pursue multiple child keys at the derivation path for a given DEX host.

_Originally posted by @chappjc in https://github.com/decred/dcrdex/pull/1201#r705735386_

itswisdomagain:
Resolves #1204. Depends on #1201.

Add ability to derive other HD account keys for a DEX if the
initial/primary account is no longer usable (e.g. is suspended).


client/core/core.go
```diff
@@ -2380,6 +2379,12 @@ func (c *Core) DiscoverAccount(dexAddr string, appPW []byte, certI interface{})
...
 	}
 	if dc.acct.isSuspended {
+		// TODO: Recover this account if it has active matches. If there are no
+		// active matches, should we check if the next usable account for this
+		// DEX is also suspended and maybe has active matches?
+		// If we're definite we want to register with a new/different key, clear
+		// the current account key using dc.acct.resetKeys() so subsequent calls
+		// to dc.acct.setupCryptoV2 does not return this suspended key.
 		c.log.Infof("HD account key for %s was reported as suspended. Register to make a new one.", dc.acct.host)
 		return dc.exchangeInfo(), false, nil // just paid=false, no error
```
itswisdomagain: Looks like an edge case we hadn't previously considered. Should suspended accounts that have active matches be recovered to enable the client complete those match negotiations?

client/core/core.go
```diff
@@ -2614,12 +2607,27 @@ func (c *Core) registerHD(dc *dexConnection, crypter encrypt.Crypter) (*incomple
...
+
+		incompleteReg, paid, suspended, err := c.register(dc, encKey, nil, acctPubB)
+		if err == nil && suspended {
+			// Reset the account keys and try again. A new key pair will be
+			// derived. If the available key pool for this DEX is exhausted,
+			// dc.acct.setupCryptoV2 will produce an error.
+			// TODO: If the user did not attempt account discovery, this
+			// suspended account may have active matches that the user will
+			// not be aware of.
+			c.log.Infof("HD account key for %s was reported as suspended. Deriving another key to use.", dc.acct.host)
```
itswisdomagain, line 2624: Similar consideration for direct registration.

client/core/core.go
```diff
@@ -2632,7 +2640,16 @@ func (c *Core) registerLegacy(dc *dexConnection, crypter encrypt.Crypter) (*inco
...
-	return c.register(dc, nil, encKeyLegacy, acctPubB)
+	incompleteReg, paid, suspended, err := c.register(dc, nil, encKeyLegacy, acctPubB)
+	if suspended {
+		// TODO: This shouldn't happen since dc.acct.setupCryptoLegacy generates
+		// random keys, but perhaps reset the keys and re-generate new ones if
+		// this happens?
+		// Note that currently, the caller will think this account does not exist
+		// and attempt to pay the fee.
+		c.log.Infof("Legacy-style account key for %s was reported as suspended.", dc.acct.host)
+	}
```
itswisdomagain, line 2651: I'm leaning more and more towards not supporting DEXes that do not support HD accounts. It beats the purpose of using a seed if DEX operators do not support HD accounts.

client/core/types.go
```diff
@@ -533,10 +533,11 @@ type dexAccount struct {
...
 
-	keyMtx  sync.RWMutex
-	encKey  []byte
-	privKey *secp256k1.PrivateKey
-	id      account.AccountID
+	keyMtx     sync.RWMutex
+	encKey     []byte
+	privKey    *secp256k1.PrivateKey
+	id         account.AccountID
+	hdKeyIndex uint32 // child index of account key, starts as 0 and increments for alternative keys for same DEX
```
itswisdomagain, line 540: `index of last child of account keypair hd path`? 🤔

client/core/types.go
```diff
@@ -599,9 +600,18 @@ func (a *dexAccount) setupCryptoLegacy(crypter encrypt.Crypter) (encPriv, pubB [
 }
 
 // setupCryptoV2 generates a hierarchical deterministic key for the account.
+// Returns the same key pair if previously set up, unless the previosuly derived
```
itswisdomagain, line 603: `previously`

client/core/types.go
```diff
@@ -685,6 +699,11 @@ func (a *dexAccount) setupCryptoV2(creds *db.PrimaryCredentials, crypter encrypt
 
 func (a *dexAccount) resetKeys() {
 	a.keyMtx.Lock()
+	if len(a.encKey) > 0 {
+		// TODO: Should only increment if the current keys were derived
+		// from a seed.
```
itswisdomagain, line 704: This would be unnecessary if we decide to only generate hd accounts going forward.

itswisdomagain (commented):
Some todo comments for discussion.

client/core/core.go
chappjc: Thought about that, and if they are recovering that means their db lacks all the match and order data that would be needed to complete those swaps, so I don't see a reason to use it in this case.
The only way to complete those swaps would be to use an existing dexc.db

chappjc: Saying that, `DiscoverAccount` should probably be doing a `c.isRegistered` like `Register` does too.

client/core/core.go
chappjc: It's not unreasonable IMO either.  A server op could at least install the 0.2 patch that sends the dexpubkey if they don't want to update to 0.3.
Let's see what @buck54321 thinks about this.

client/core/types.go
```diff
@@ -640,7 +650,7 @@ func (a *dexAccount) setupCryptoV2(creds *db.PrimaryCredentials, crypter encrypt
 	}
 
 	// Prepare the chain of child indices.
-	kids := make([]uint32, 0, 10) // 1 x purpose, 1 x version (incl. oddness), 8 x 4-byte uint32s.
+	kids := make([]uint32, 0, 10) // 1 x purpose, 1 x version (incl. oddness), 7 x 4-byte uint32s, 1 x acct key index.
```
chappjc, line 653: Isn't it still 8 r-byte uint32s?

client/core/types.go
```diff
@@ -650,6 +660,10 @@ func (a *dexAccount) setupCryptoV2(creds *db.PrimaryCredentials, crypter encrypt
 	for i := 0; i < 8; i++ {
 		kids = append(kids, binary.LittleEndian.Uint32(byteSeq[i*4:i*4+4]))
 	}
+	// Last child is the account key index, which is 0 for the primary/default
+	// account and 0 > k < 2^31 for supplementary accounts where previously
+	// derived account keys have been cleared/reset.
+	kids = append(kids, keyIndex)
```
chappjc, line 666: I know we haven't released an HD keys dex client yet, but anyone that was using HD accounts prior to this commit will then loose the ability to restore their previous account because we now starting one branch deeper.  Is that correct?

itswisdomagain: It is. Can be resolved by not appending last child if `keyIndex == 0` but don't know if that's necessary since there's no live client with this feature yet.

client/core/types.go
itswisdomagain: Hmm. It is. Wonder how I arrived at 7 before! Will undo.

client/core/types.go
chappjc: Just testnet accounts is all.  Devs can re-register on dex-test if they did use it with the prior derivation path.

client/core/core.go
```diff
@@ -2614,12 +2607,27 @@ func (c *Core) registerHD(dc *dexConnection, crypter encrypt.Crypter) (*incomple
...
+		if err == nil && suspended {
+			// Reset the account keys and try again. A new key pair will be
+			// derived. If the available key pool for this DEX is exhausted,
+			// dc.acct.setupCryptoV2 will produce an error.
+			// TODO: If the user did not attempt account discovery, this
+			// suspended account may have active matches that the user will
+			// not be aware of.
+			c.log.Infof("HD account key for %s was reported as suspended. Deriving another key to use.", dc.acct.host)
+			dc.acct.resetKeys()
+			continue
```
chappjc, line 2626: Let's add a something like 200 ms sleep here so we don't get our requests denied on account of request rate.

client/core/core.go
chappjc: But regarding these lines of code, if `registerLegacy` finds the account is suspended, that should probably return and error to the caller for the reasons you cited (it's a fresh random identity).

client/core/types.go
chappjc: `hdKeyIndex` will be back to zero each time Core is started.

Since I'm not inclined to add this to the database, I think it can just be a variable that is local to `registerHD`.  Specifically start with `var hdKeyIndex uint32` at the beginning of `registerHD`, pass `hdKeyIndex` to `setupCryptoV2` (removing the mutex locking and `HardenedKeyStart` check), if account is suspended, `hdKeyIndex++` right there instead of in `resetKeys`.  This also makes the [`len(a.encKey) > 0` check question](https://github.com/decred/dcrdex/pull/1210/commits/4322ac8029c5ce818c005244399a23b567381798#r708500390) a non-issue.

client/core/core.go
chappjc: Umm... I'm realizing that I haven't tested recently getting my account suspended.  We want to re-register in that case **without** re-seeding.  Was/is that handled somehow.  The `c.isRegistered` is certainly blocking that from `Register` at least, regardless of the active swaps issue.

chappjc:
The open questions in my mind, with the new ability to derive child HD accounts for a given dex:

- Should it be permitted to export an hd account?  I think no.  We don't want such an account imported, only restored by seed.
- Should it be permitted to manually disable an hd account that is *not* suspended?  I think no, because how would re-seed know if an account was previously disabled.  Only disabling a suspended HD account makes sense to me, although I think such an action should be part of a meta action also involves registering the next account, can be done in either this PR or a follow-up.
- Should we even permit a user to register on a server that's does not facilitate hd keys?  I'm also leaning no on that now too.
- The question this PR settles so far is, should a user be permitted to register a _legacy_ account on a server that supports HD keys.  The answer is now no with child key derivation.

client/core/core.go
JoeGruffins: imo make everyone update, and don't support older dexes. I believe that is in line with the decred ethos, no soft forks.

JoeGruffins:
in my opinion

> Should it be permitted to export an hd account?

No, it would serve no purpose. You can do this with a decred wallet (but not btc apparently) because you can watch addresses for funds, but there would be nothing to benefit here. We want the user to keep the seed and restore from seed.

> Should it be permitted to manually disable an hd account that is not suspended? 

No. They can start with a new seed for that.

> Should we even permit a user to register on a server that's does not facilitate hd keys?

No. Have operator(s?) update. If they cbf to do that they don't need to run a server.

> The question this PR settles so far is, should a user be permitted to register a legacy account on a server that supports HD keys.

Good, let them update too.

client/core/types.go
```diff
@@ -645,6 +660,10 @@ func (a *dexAccount) setupCryptoV2(creds *db.PrimaryCredentials, crypter encrypt
...
```
JoeGruffins, line 666: Actually `0' + keyIndex` in the end or `2^31 < k < (2^32)-1` because `genChild` hardens it.

chappjc:
https://github.com/decred/dcrdex/pull/1201 is merged, so you can rebase away that commit

chappjc:
> > Should it be permitted to manually disable an hd account that is not suspended?
> 
> No. They can start with a new seed for that.

That's what I said initially too, but the use case is a user that has several dex accounts, and is just "done" with one of them for whatever reason.  I'm tempted to say whatever just let them ignore that dex, but on login Core connects to all configured DEXs and the user may want to stop all comms with a particular dex.

client/core/core.go
chappjc: Here's the gist of what I propose for `DiscoverAccount`, a loop similar to what `registerHD` does.

```go
// Setup our account keys and attempt to authorize with the DEX.
creds := c.creds()

// Start at key index 0 and attempt to authorized accounts until either (1)
// the server indicates the account is not found, and we return paid=false
// to signal a new account should be registered, or (2) an account is found
// that is not suspended, and we return with paid=true after storing the
// discovered account and promoting it to a persistent connection. In this
// process, we will increment key index and try again whenever the connect
// response indicates a suspended account is found. This instance of Core
// lacks and order or match history for this dex to complete any active
// swaps that might exist for a suspended account, so the user had better
// have another instance with this data if they hope to recover those swaps.
var keyIndex uint32
var encKey []byte
for {
	encKey, _, err = dc.acct.setupCryptoV2(creds, crypter, keyIndex)
	if err != nil {
		return dc.exchangeInfo(), false, newError(acctKeyErr, "setupCryptoV2 error: %v", err)
	}

	// Discover the account by attempting a 'connect' (authorize) request.
	err = c.authDEX(dc)
	if err != nil {
		var mErr *msgjson.Error
		if errors.As(err, &mErr) && mErr.Code == msgjson.AccountNotFoundError {
			return dc.exchangeInfo(), false, nil // all good, just go register now
		}
		return nil, false, newError(authErr, "unexpected authDEX error: %v", err)
	}
	if dc.acct.isSuspended {
		c.log.Infof("HD account key for %s was reported as suspended. Deriving another account key.", dc.acct.host)
		dc.acct.resetKeys()
		keyIndex++
		continue
	}

	break // great, the account at this key index is paid and ready
}
```

On thing to consider however is helping out any subsequent `Register` that will have to do the same iteration all over if there are suspended accounts, possibly caching `keyIndex` so it can be used to speed up the `Register` call.

client/core/types.go
```diff
@@ -680,6 +699,11 @@ func (a *dexAccount) setupCryptoV2(creds *db.PrimaryCredentials, crypter encrypt
 
 func (a *dexAccount) resetKeys() {
 	a.keyMtx.Lock()
+	if len(a.encKey) > 0 {
+		// TODO: Should only increment if the current keys were derived
+		// from a seed.
+		a.hdKeyIndex++ // increment this counter to ensure next derived hd key differs from this
```
chappjc, line 705: One gotcha that's not so obvious is that `genChild` in `setupCryptoV2` may end up using a higher key index than the one provided in its input argument since it increments if `Child` errors.  This is supposedly extremely rare but possible.  The result is that when we loop through keys in both `DiscoverAccount` and `registerHD` and increment the key, we may be incrementing to the one that was actually just tried by `genChild` if it has incremented it internally.  I do not thing this is a problem however as it will just check the same key twice and then go on to the next, but it's worth noting.

client/core/types.go
chappjc: I actually kinda think this comment is getting to far into the weeds.  I'd be good with just `// Last child is based on the account key index.` for this `append`.  The `genChild` docs can explain that hardened children are generated.

client/core/core.go
chappjc: Cool, let's nuke `registerLegacy` and simplify `(*Core).Register` logic.

We'll still have to maintain some logic for imported accounts, but at least the registration bits won't have to worry about it.

client/core/core.go
```diff
@@ -2614,12 +2607,27 @@ func (c *Core) registerHD(dc *dexConnection, crypter encrypt.Crypter) (*incomple
...
-	if err != nil {
-		return nil, false, newError(acctKeyErr, "setupCryptoV2 error: %v", err)
-	}
+	for {
+		encKey, acctPubB, err := dc.acct.setupCryptoV2(creds, crypter)
+		if err != nil {
+			return nil, false, newError(acctKeyErr, "setupCryptoV2 error: %v", err)
+		}
+
+		incompleteReg, paid, suspended, err := c.register(dc, encKey, nil, acctPubB)
```
chappjc, line 2616: Oh dang I was testing the loop in DiscoverAccount, and it worked well, but they going to register immediately hits the `'register'` request rate limit of once per minute.

`rpc error: error code 59: too many requests to register`

I'm considering how we want to deal with this now.  I'm thinking Register can call DiscoverAccount (have it do the loop) and use the discovered `keyIndex`...

chappjc: Gosh I really should have handled this in my last PR when I started using the `connect` route in `DiscoverAccount`.  I wanted to wrap my mind around the edges so I hacked up the idea above.

https://github.com/decred/dcrdex/commit/656507796eee694da6578a94a07b4d6192680844

It does seem to work well, and only ever calls the `register` route once.

I was able to both discover and register for a seed with two suspended accounts.

I was also able to disable the account, and then by doing the registration dialog again it discovered it without a fuss.

client/core/core.go
buck54321: As long as users can still import accounts, OK. Really should be waiting for a major version change, but...

chappjc: True, and when we are at 1.0, I think the rules will change that way.  Although, right now 0.x means no promises.  Not that we should make a habit of making life difficult for users and operators, just that on the way to 1.0 we can be more forward-looking than we might otherwise be, otherwise we risk burning too much time and creating spaghetti to support obsolete features.
Definitely ack that POV though @buck54321.

buck54321:
> > Should it be permitted to export an hd account?
>
> No, it would serve no purpose. You can do this with a decred wallet (but not btc apparently) because you can watch addresses for funds, but there would be nothing to benefit here. We want the user to keep the seed and restore from seed.

If I was running dexc on two different machines, and wanted to move an account, I should be able to.

chappjc:
> If I was running dexc on two different machines, and wanted to move an account, I should be able to.

Oh, different seeds on those machines?  Interesting use case.

JoeGruffins:
> If I was running dexc on two different machines, and wanted to move an account, I should be able to.

With built-in wallets, wouldn't you still need the same app seed to get the same wallets, if wallet's are also based off the seed for purpose of registration. So to use the same dex account you would still need the same wallet accounts. Or maybe I don't understand the flow those will take.

JoeGruffins:
Using the same wallets on separate machines at the same time may cause a lot of problems, how to lock coins? I imagine trying to trade with the same account on two machines at the same time would also cause something to break. I'm sure you don't mean that, just saying. It's not something I had thought of until just now.

buck54321:
> So to use the same dex account you would still need the same wallet accounts.

Not at all. You obviously can't do dumb stuff like start an order with one wallet and finish it with another.

> Wouldn't we need to keep trade data in the db synced to have the same account in two places?

This isn't a concern specific to HD accounts, but yes. If you want to run DEX on two different machines, you can't do it the same time.

chappjc:
Just pushed one more cleanup commit.  Going to test again with a string of two suspended accounts to iterate through, as I did in https://github.com/decred/dcrdex/pull/1210#discussion_r710554573, and they I'm planning to squash and merge this today.

I will keep https://github.com/decred/dcrdex/pull/1210/commits/ff8bb500fc45301717c02510f368a63e1a1ce015 separate however, because that isn't actually related to this PR directly.  The others will become one commit.

chappjc:
Retested the entire thing.  Looks solid and will merge first thing Monday (CDT) unless there are concerns.

Register a new/fresh account.  No existing account found server-side; just register:

> [DBG] CORE: Authenticated connection to 127.0.0.1:17273, acct e1a90d7ea3701f5ec63e389543fc62088738c5794e947c427cb957d517eb9180, 0 active orders, 0 active matches, score 0 (suspended = false)
> [INF] CORE: notify: |SUCCESS| (feepayment) Account registered - You may now trade at 127.0.0.1:17273

Get suspended (with preimage misses spread over two epochs).
Disable the suspended account. 
Add a DEX.  DiscoverAccount logs:

> [INF] CORE: Connected to DEX server at 127.0.0.1:17273 but NOT listening for messages.
> [DBG] CORE: Authenticated connection to 127.0.0.1:17273, acct e1a90d7ea3701f5ec63e389543fc62088738c5794e947c427cb957d517eb9180, 0 active orders, 0 active matches, score 20 (suspended = true)
> [INF] CORE: HD account key for 127.0.0.1:17273 was reported as suspended. Deriving another account key.
> [DBG] CORE[wss://127.0.0.1:17273/ws]: Sending close 1000 (normal) message.
> 2021/10/03 13:02:52 "POST http://127.0.0.3:5758/api/discoveracct HTTP/1.1" from 127.0.0.1:59850 - 200 1062B in 493.628624ms

Confirm Registration comes up (discover said not paid), confirm:

> [DBG] CORE: Authenticated connection to 127.0.0.1:17273, acct e1a90d7ea3701f5ec63e389543fc62088738c5794e947c427cb957d517eb9180, 0 active orders, 0 active matches, score 20 (suspended = true)
> [INF] CORE: HD account key for 127.0.0.1:17273 was reported as suspended. Deriving another account key.
> [INF] CORE: Attempting registration fee payment to SsbeD1czwoAwqxV8fiq1sbhT7JsypYN5LPK, account ID 40eb5bad3a76569b24d60a72a0c2380369571de4655a4b16927de8aba3740274, of 100000000 units of dcr. Do NOT manually send funds to this address even if this fails.
> ...
> 2021/10/03 13:03:06 "POST http://127.0.0.3:5758/api/register HTTP/1.1" from 127.0.0.1:59856 - 200 12B in 529.536538ms
> ...
> [DBG] CORE: Authenticated connection to 127.0.0.1:17273, acct 40eb5bad3a76569b24d60a72a0c2380369571de4655a4b16927de8aba3740274, 0 active orders, 0 active matches, score 0 (suspended = false)

Get suspended again.  This time instead of disabling the account, shutdown dexc and DELETE dexc.db. 
Restore from seed. 
Add a DEX.  DiscoverAccount logs:

> [DBG] CORE: Authenticated connection to 127.0.0.1:17273, acct e1a90d7ea3701f5ec63e389543fc62088738c5794e947c427cb957d517eb9180, 0 active orders, 0 active matches, score 20 (suspended = true)
> [INF] CORE: HD account key for 127.0.0.1:17273 was reported as suspended. Deriving another account key.
> [DBG] CORE: Authenticated connection to 127.0.0.1:17273, acct 40eb5bad3a76569b24d60a72a0c2380369571de4655a4b16927de8aba3740274, 0 active orders, 0 active matches, score 20 (suspended = true)
> [INF] CORE: HD account key for 127.0.0.1:17273 was reported as suspended. Deriving another account key.

Two suspended accounts found, no paid account found.
Confirm Register:

> [DBG] CORE: Authenticated connection to 127.0.0.1:17273, acct e1a90d7ea3701f5ec63e389543fc62088738c5794e947c427cb957d517eb9180, 0 active orders, 0 active matches, score 20 (suspended = true)
> [INF] CORE: HD account key for 127.0.0.1:17273 was reported as suspended. Deriving another account key.
> [DBG] CORE: Authenticated connection to 127.0.0.1:17273, acct 40eb5bad3a76569b24d60a72a0c2380369571de4655a4b16927de8aba3740274, 0 active orders, 0 active matches, score 20 (suspended = true)
> [INF] CORE: HD account key for 127.0.0.1:17273 was reported as suspended. Deriving another account key.
> [INF] CORE: Attempting registration fee payment to Ssm8Vk5C35s8dNezzCdZjrtHdkzPafCMQ27, account ID a903901bfb2e5abf9cfe3d1938036ec3c55ea5ab69da402b635ec85024c6f0f5, of 100000000 units of dcr. Do NOT manually send funds to this address even if this fails.
> ...
> 2021/10/03 13:05:40 "POST http://127.0.0.3:5758/api/register HTTP/1.1" from 127.0.0.1:59956 - 200 12B in 747.262867ms
> ...
> [INF] CORE: Notifying dex 127.0.0.1:17273 of fee payment.
> [DBG] CORE: Authenticated connection to 127.0.0.1:17273, acct a903901bfb2e5abf9cfe3d1938036ec3c55ea5ab69da402b635ec85024c6f0f5, 0 active orders, 0 active matches, score 0 (suspended = false)
> [INF] CORE: notify: |SUCCESS| (feepayment) Account registered - You may now trade at 127.0.0.1:17273

Great.
Now, shutdown and DELETE dex.db. 
Restore from seed. 
Add a DEX.  DiscoverAccount logs:

> [INF] CORE: Connected to DEX server at 127.0.0.1:17273 but NOT listening for messages.
> [DBG] CORE: Authenticated connection to 127.0.0.1:17273, acct e1a90d7ea3701f5ec63e389543fc62088738c5794e947c427cb957d517eb9180, 0 active orders, 0 active matches, score 20 (suspended = true)
> [INF] CORE: HD account key for 127.0.0.1:17273 was reported as suspended. Deriving another account key.
> [DBG] CORE: Authenticated connection to 127.0.0.1:17273, acct 40eb5bad3a76569b24d60a72a0c2380369571de4655a4b16927de8aba3740274, 0 active orders, 0 active matches, score 20 (suspended = true)
> [INF] CORE: HD account key for 127.0.0.1:17273 was reported as suspended. Deriving another account key.
> [DBG] CORE: Authenticated connection to 127.0.0.1:17273, acct a903901bfb2e5abf9cfe3d1938036ec3c55ea5ab69da402b635ec85024c6f0f5, 0 active orders, 0 active matches, score 0 (suspended = false)
> 2021/10/03 13:06:27 "POST http://127.0.0.3:5758/api/discoveracct HTTP/1.1" from 127.0.0.1:59978 - 200 1061B in 1.199696599s

chappjc:
For completeness, using dexcctl with the above account (no discover account step, just straight to register):

> [DBG] CORE: Authenticated connection to 127.0.0.1:17273, acct e1a90d7ea3701f5ec63e389543fc62088738c5794e947c427cb957d517eb9180, 0 active orders, 0 active matches, score 20 (suspended = true)
> [INF] CORE: HD account key for 127.0.0.1:17273 was reported as suspended. Deriving another account key.
> [DBG] CORE: Authenticated connection to 127.0.0.1:17273, acct 40eb5bad3a76569b24d60a72a0c2380369571de4655a4b16927de8aba3740274, 0 active orders, 0 active matches, score 20 (suspended = true)
> [INF] CORE: HD account key for 127.0.0.1:17273 was reported as suspended. Deriving another account key.
> [DBG] CORE: Authenticated connection to 127.0.0.1:17273, acct a903901bfb2e5abf9cfe3d1938036ec3c55ea5ab69da402b635ec85024c6f0f5, 0 active orders, 0 active matches, score 0 (suspended = false)

buck54321 (changes requested):
Looks good. Just a couple of things.

---

When I nuke my db before there's enough confirmations to complete registration, but then try to restore from seed, I get this on the DEX address form.



Which comes from the `connect` response. It appears the only way around this (from the web UI, at least) would be to start with a new seed, which of course loses any other DEX accounts or built-in wallets. I think might need to look for the `UnpaidAccountError` and handle it somehow.

---

Not from this PR, but when registering fresh, I'm seeing the poke notification `TopicDEXDisconnected`

```
notify: |POKE| (conn) Server disconnect - 127.0.0.1:17273 is disconnected
```

Which it looks like is actually supposed to `TopicDEXConnected`, but they are set wrong in `handleConnectEvent`.

client/core/core_test.go
```diff
@@ -1514,12 +1531,18 @@ func testRegister(t *testing.T, legacyKeys bool) {
...
+		rig.queueConnect(accountNotFoundError, nil, nil) // for discoverAccount
+	}
+
+	queueResponses := func() {
+		queueConfigAndConnectUnknownAcct()
 		rig.queueRegister(regRes)
 		queueTipChange()
-		rig.queueNotifyFee()
-		rig.queueConnect(nil, nil, nil)
+		rig.queueNotifyFee() // when request received, does rig.queueConnect(nil, nil, nil) in request handler (before response)
```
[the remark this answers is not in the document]
chappjc: Ah, how silly.  This had solved a problem for me initially, but I cannot see the purpose anymore.

chappjc:
> When I nuke my db before there's enough confirmations to complete registration, but then try to restore from seed, I get this on the DEX address form.
> ... I think might need to look for the `UnpaidAccountError` and handle it somehow.

Ooof good catch.  Will update so the seed is at least salvageable.  If the client had already made the txn and was just waiting for confs, salvaging the seed by making another txn is likely the only option if the user screwed up like this because they don't have the fee coin info to watch and report with `notify_fee`.

Whether or not they had made a txn yet, they'd need to get the assigned fee address and asset ID to make one to continue using the seed as you've pointed out.  The register response to an account in that state includes that info, so the resolution in this case is as you suggested -- if the connect response is a `UnpaidAccountError` it can treat that as the same as `AccountNotFoundError` and use the register route to get the assigned fee address.  That actually looks very straightforward in `discoverAccount`, but probably worth a special log.

This would at least allow the seed to be salvageable, but I see no easy way to recover a txn that was broadcast but simply not yet fully confirmed.

chappjc:
I should clarify that @itswisdomagain and I are collaborating on this PR, and I believe he's focusing on https://github.com/decred/dcrdex/pull/1227 and some dcrlibwallet work.  I don't mean to hijack this or anything.  It seems close to done.

JoeGruffins (approved):

PR #1213, opened 2021-09-15
introduce server/cmd/usermatches

chappjc:
This creates a new CLI app called **usermatches** for directly retrieving match data from the DB.

This also modifies server/db/driver/pg with a new constructor, `NewArchiverForRead`, for accessing the `Archiver` in a read-only mode.

### DIFF
--- a/server/cmd/usermatches/main.go
+++ b/server/cmd/usermatches/main.go
@@ -1,0 +1,256 @@
+// This code is available on the terms of the project LICENSE.md file,
+// also available online at https://blueoakcouncil.org/license/1.0.0.
+
+package main
+
+import (
+	"context"
+	"encoding/csv"
+	"encoding/hex"
+	"flag"
+	"fmt"
+	"os"
+	"os/signal"
+	"strconv"
+	"strings"
+
+	"decred.org/dcrdex/dex"
+	"decred.org/dcrdex/server/account"
+	"decred.org/dcrdex/server/asset"
+	"decred.org/dcrdex/server/asset/btc"
+	"decred.org/dcrdex/server/asset/dcr"
+	"decred.org/dcrdex/server/asset/ltc"
+	"decred.org/dcrdex/server/db"
+	"decred.org/dcrdex/server/db/driver/pg"
+)
+
+// We do not need a Backend Setup, just the Drivers to call DecodeCoinID. While
+// we can do this with asset.DecodeCoinID(assetID, coinID), doing the following
+// drastically reduces the locking/unlocking (asset.driversMtx) that would be
+// required to decode coin IDs, and we will likely be doing many.
+var assets = map[uint32]asset.Driver{
+	0:  &btc.Driver{},
+	2:  &ltc.Driver{},
+	42: &dcr.Driver{},
+}
+
+var dbhost = flag.String("host", "/run/postgresql", "pg host") // default to unix socket, but 127.0.0.1 would be common too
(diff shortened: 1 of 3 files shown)


martonp (commented):
Works well for me.

The program overwrites previous versions of the file, so it would be nice to add a parentheses and a version if there is already a file with the same name.. e.g. acct_matches_123(1).csv. This wouldn't usually matter with a production server since the db would not get cleared, but if someone's testing on simnet, creates a file, clears the db, and creates another file, it would be annoying to lose the original file.

Searching without an acct could create a shorter filename than all those 0s, but that's nbd.

chappjc:
> Searching without an acct could create a shorter filename than all those 0s, but that's nbd.

This is a good point.  Should be a special file name when there's no account filter.

About the output file, I think a number appended like "(1)" would actually be a little error prone as you'd run the program and copy the csv that you expect it to be named.  I'd be more inclined to error out if the targeted file name exists.  Will consider that.

martonp:
Maybe append a timestamp?

chappjc:
That's a thought too, or even the highest epoch index.  However, I'm really inclined to have a predictable file name for scripting purposes (run app, copy output file, etc.).  One might argue I should go the extra mile and have it go to stdout by default so the caller can just redirect to a file of their choice, or even add and --output switch, but this is sorta a one-off tool that I didn't want to spend much time on unless necessary.

server/cmd/usermatches/main.go
```diff
@@ -0,0 +1,257 @@
...
+			err = fmt.Errorf("txinout (%s) not formatted as a txin/out", txinout)
+			return
+		}
+		_, err = strconv.ParseUint(txsplit[1], 10, 32)
+		if err != nil {
+			err = fmt.Errorf("strconv.ParseUint(%s): %w", txsplit[1], err)
+			return
+		}
+		return txsplit[0], txsplit[1], nil
+	}
```
buck54321, line 213: Ethereum is going to break this.

server/cmd/usermatches/main.go
```diff
@@ -0,0 +1,257 @@
...
+)
+
+// We do not need a Backend Setup, just the Drivers to call DecodeCoinID. While
+// we can do this with asset.DecodeCoinID(assetID, coinID), doing the following
+// drastically reduces the locking/unlocking (asset.driversMtx) that would be
+// required to decode coin IDs, and we will likely be doing many.
+var assets = map[uint32]asset.Driver{
+	0:  &btc.Driver{},
+	2:  &ltc.Driver{},
+	42: &dcr.Driver{},
```
buck54321, line 34: I know this will work, but if you wanted to add an exported getter in **server/asset/driver.go** for the package-registered `asset.Driver`, that'd be fine too, even if we're not using it in production.

buck54321 (approved):

server/cmd/usermatches/main.go
chappjc: I'm ok with this I think, but a `asset.CoinDecoder(assetID uint32) func([]byte) (string, error)` that just returns the named asset's `(*Driver).DecodeCoinID` function might be good too.

server/cmd/usermatches/main.go
chappjc: Mmmm, yeah.  I guess it remains to be seen what pieces of the MatchData for eth are what.  Can cross that bridge if we want eth as a supported asset for usermatches.

PR #1215, opened 2021-09-16
networks/eth: Add methods to decode tx data.

JoeGruffins:
In order to check data in a mempool transaction is formatted as we
    expect and in order to read its arguments, add package methods that are
    able to parse and validate Initiate and Redeem transaction data.

    Some of this code is taken directly from the go-ethereum repository
    where it is unexported there. abi.go and abi_test.go should be updated
    if they change over there.

### DIFF
--- a/client/asset/eth/doc.go
+++ b/client/asset/eth/doc.go
@@ -1,13 +1,14 @@
-//go:build gpl
-// +build gpl
-//
 // This code is available on the terms of the project LICENSE.md file,
 // also available online at https://blueoakcouncil.org/license/1.0.0.
 //
 // This package also imports go-ethereum code and so carries the burden of
 // go-ethereum's GNU Lesser General Public License.
 
+//go:build lgpl
+// +build lgpl
+
 // Package eth implements methods to work with ethereum swap contracts and
-// transactions. The code is unique as it requires a more restrictive liscense
-// and connot be used in proprietary software.
+// transactions. The LGPL is a more restrictive license that may be more of a
+// burden for closed source software. As such, the "lpgl" build tag is required
+// for this package.
 package eth
--- a/client/asset/eth/doc.go
+++ b/client/asset/eth/doc.go
@@ -1,0 +1,13 @@
+//go:build gpl
+// +build gpl
+//
+// This code is available on the terms of the project LICENSE.md file,
+// also available online at https://blueoakcouncil.org/license/1.0.0.
+//
+// This package also imports go-ethereum code and so carries the burden of
+// go-ethereum's GNU Lesser General Public License.
+
+// Package eth implements methods to work with ethereum swap contracts and
+// transactions. The code is unique as it requires a more restrictive liscense
+// and connot be used in proprietary software.
+package eth
--- a/client/cmd/dexc/gplimport.go
+++ b/client/cmd/dexc/gplimport.go
@@ -1,0 +1,16 @@
+//go:build gpl
+// +build gpl
+//
+// This code is available on the terms of the project LICENSE.md file,
+// also available online at https://blueoakcouncil.org/license/1.0.0.
+//
+// When using eth, this app also imports go-ethereum code and so carries the
+// burden of go-ethereum's GNU Lesser General Public License.
+
+package main
+
+// import (
+// 	ETH is commented to prevent go-ethereum goroutines from launching via
+// 	it's init(). Uncomment this when ETH is ready:
+// 	_ "decred.org/dcrdex/client/asset/eth" // register eth asset
+// )
--- a/dex/networks/eth/abi.go
+++ b/dex/networks/eth/abi.go
@@ -1,3 +1,6 @@
+//go:build gpl
+// +build gpl
+//
 // Copyright 2019 The go-ethereum Authors
 // This file is part of the go-ethereum library.
 //
--- a/dex/networks/eth/abi.go
+++ b/dex/networks/eth/abi.go
@@ -1,0 +1,166 @@
+// Copyright 2019 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+// This file lifted ver-batim from go-ethereum/signer/fourbyte at v1.10.6 commit
+// 576681f29b895dd39e559b7ba17fcd89b42e4833.
+package eth
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// decodedCallData is an internal type to represent a method call parsed according
+// to an ABI method signature.
+type decodedCallData struct {
+	signature string
+	name      string
+	inputs    []decodedArgument
+}
+
+// decodedArgument is an internal type to represent an argument parsed according
+// to an ABI method signature.
+type decodedArgument struct {
+	soltype abi.Argument
+	value   interface{}
+}
+
+// String implements stringer interface, tries to use the underlying value-type
+func (arg decodedArgument) String() string {
+	var value string
+	switch val := arg.value.(type) {
+	case fmt.Stringer:
+		value = val.String()
+	default:
+		value = fmt.Sprintf("%v", val)
+	}
+	return fmt.Sprintf("%v: %v", arg.soltype.Type.String(), value)
+}
+
+// String implements stringer interface for decodedCallData
+func (cd decodedCallData) String() string {
+	args := make([]string, len(cd.inputs))
+	for i, arg := range cd.inputs {
+		args[i] = arg.String()
+	}
+	return fmt.Sprintf("%s(%s)", cd.name, strings.Join(args, ","))
+}
+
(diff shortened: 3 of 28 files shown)


martonp:
What do you want to use this for? Just wondering.

JoeGruffins:
For the server, when initially receiving an initialize or redeem request, it is able to do more verification while the tx is in mempool. Otherwise we can only take the client's word that they sent a swap with x secret hash and conterparty y. We can look at the mempool tx to decided if it looks ok before mined, and can fail faster in the case it is a bum transaction. We will still check the contract once there are confirmations.

dex/networks/eth/abi.go
```diff
@@ -0,0 +1,166 @@
+// Copyright 2019 The go-ethereum Authors
+// This file is part of the go-ethereum library.
+//
+// The go-ethereum library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
```
chappjc, line 5: I _think_ it's OK to mix **L**GPL.  Fortunately they only use gpl for the actual geth executable app.

dex/networks/eth/abi.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return nil, err
+	}
+	values, err := method.Inputs.UnpackValues(argdata)
+	if err != nil {
+		return nil, fmt.Errorf("signature %q matches, but arguments mismatch: %v", method.String(), err)
+	}
+	// Everything valid, assemble the call infos for the signer
+	decoded := decodedCallData{signature: method.Sig, name: method.RawName}
+	for i := 0; i < len(method.Inputs); i++ {
```
chappjc, line 146: We should probably just leave the code verbatim so we can more easily apply fixes/updates from upstream, but it bugs me they don't just `range`... or allocate `decoded.inputs`.  Oh well

chappjc (approved):

dex/networks/eth/abi.go
chappjc: OK, so feedback from @jrick ... the benefit of lgpl over gpl only applies when _dynamically linking_.  While we are an open source project and it doesn't affect us, any proprietary software building on dcrdex would be affected unless they rip out these bits.  If we revisit the plugins idea, which is is dynamically linked libraries if I understand that correctly, it would be alright, but plugins sucked for a number of reasons including cgo.
Thoughts?

chappjc: As long as we only import go-ethereum from obviously eth-specific packages (we do), a third consumer can fairly easily avoid accidentally importing (and statically linking) LGPL code.

The design we have of only the actual applications like dexc and dcrdex importing the asset drivers is fortunate since all other importable packages in our repository, other than the `.../eth` ones, will be free of the LGPL code.

To make sure this doesn't change, I suggest we comment or make a readme in each package that either imports go-ethereum code or has such lifted code, describing clearly that by importing the package the consumer is linking LGPL code.

jrick: gate all eth usage behind a build tag.

JoeGruffins:
I assume this is the right direction.

client/cmd/dexc/gplimport.go
```diff
@@ -0,0 +1,16 @@
+//go:build gpl
```
chappjc, line 1: Was thinking about the actual tag. `eth` would be good to express the purpose, while `lgpl` would convey the consequences.  `ethlgpl` would be a little long.  I'm inclined to say `lgpl` since it's most accurate and important.

dex/networks/eth/doc.go
```diff
@@ -0,0 +1,13 @@
...
+//
+// This code is available on the terms of the project LICENSE.md file,
+// also available online at https://blueoakcouncil.org/license/1.0.0.
+//
+// This package also imports go-ethereum code and so carries the burden of
+// go-ethereum's GNU Lesser General Public License.
+
+// Package eth implements methods to work with ethereum swap contracts and
+// transactions. The code is unique as it requires a more restrictive liscense
+// and connot be used in proprietary software.
```
chappjc, line 12: I'm not totally sure of the wording here.  I know GPL code is used all the time in closed source systems, and it's not quite as simple as this.  I'm pretty sure some the requirements are distributing the license, possibly the gpl code or a link to the gpl code, maybe the modifications, it all depends.  Whatever the specifics, there are a lot more strings attached.  https://opensource.stackexchange.com/questions/2338/can-i-use-gpl-libraries-in-a-closed-source-project-if-only-the-output-is-distrib  

The developers of the proprietary software can sort out the details for their product. So, I think we should change the wording from:

> The code is unique as it requires a more restrictive liscense and connot be used in proprietary software

to

> The LGPL is a more restrictive license that may be more of a burden for closed source software. As such, the "lgpl" build tag is required for this package.

dex/networks/eth/doc.go
```diff
@@ -0,0 +1,13 @@
+//go:build gpl
```
chappjc, line 1: The way I see the golang project structuring their docs is with the license at the top, the `go:build` and `+build` lines separated from other comments by white space, and then the `package` docs.  So three chunks of comments.  https://github.com/golang/text/commit/8f690f22cf1c026c950adddf3d45258bfd0912f0#diff-e03ec920b0cdec012a5f833c15db00c9d4ce59332ac08e9a90bd5e32e9ad16fb

When I run my Go 1.17 `go fmt` on these packages, it even separates the `go:build` parts from the rest of the comments for me.  Essentially, the file should look like this:

```go
// This code is available on the terms of the project LICENSE.md file,
// also available online at https://blueoakcouncil.org/license/1.0.0.
//
// This package also imports go-ethereum code and so carries the burden of
// go-ethereum's GNU Lesser General Public License.

//go:build gpl
// +build gpl

// Package eth implements methods to work with ethereum swap contracts and
// transactions. The code is unique as it requires a more restrictive liscense
// and connot be used in proprietary software.
package eth
```

client/asset/eth/doc.go
```diff
@@ -0,0 +1,13 @@
...
+// This code is available on the terms of the project LICENSE.md file,
+// also available online at https://blueoakcouncil.org/license/1.0.0.
+//
+// This package also imports go-ethereum code and so carries the burden of
+// go-ethereum's GNU Lesser General Public License.
+
+// Package eth implements methods to work with ethereum swap contracts and
+// transactions. The code is unique as it requires a more restrictive liscense
+// and connot be used in proprietary software.
+package eth
```
chappjc, line 13: As with all the files, `gofmt` is forcing a newline between the tags and other docs.  Let's follow: license, tags, package docs in separate blocks in that order:

```go
// This code is available on the terms of the project LICENSE.md file,
// also available online at https://blueoakcouncil.org/license/1.0.0.
//
// This package also imports go-ethereum code and so carries the burden of
// go-ethereum's GNU Lesser General Public License.

//go:build gpl
// +build gpl

// Package eth implements ...
package eth
```

chappjc (changes requested):
Looks about right, although I request a little less specific wording about the consequences of using LGPL in closed source software because it's a lot more nuanced and really the concern of those developers.  We just want to make it harder to screw up an use lgpl code assuming everything in this repo is permissive.

Also, my `gofmt` is segregating the build tag comments.  You must be using Go 1.16, because the formatter changed in Go 1.17.  So I'd like to separate and reorder file comments: license, build tags, and then `package` docs (the ones that should start `// Package x ... `.  examples https://github.com/golang/text/commit/8f690f22cf1c026c950adddf3d45258bfd0912f0#diff-e03ec920b0cdec012a5f833c15db00c9d4ce59332ac08e9a90bd5e32e9ad16fb

dex/networks/eth/abi.go
```diff
@@ -0,0 +1,169 @@
...
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+// This file lifted ver-batim from go-ethereum/signer/fourbyte at v1.10.6 commit
```
chappjc, line 20: "ver-batim" is an unusual way to write this.  Just "verbatim"?

client/cmd/dexc/gplimport.go
JoeGruffins: I thought we were using lgpl and eth was gpl. I must be confused though. Can change to eth anyhow, but I thought this might be a thing for other coins as well. Coins we haven't considered adding yet.

dex/networks/eth/doc.go
JoeGruffins: I have been relying on my editor (gvim...) plugin to format, which is probably behind the times. Will format correctly.

chappjc: `go fmt` behaviour actually change from go 1.16 -> go 1.17, which is why CI's version of golangci-lint isn't catching this.  Locally it changes though.

client/cmd/dexc/gplimport.go
chappjc: Ah, we're using Blue Oak, which is close enough to MIT and ISC and the simpler BSD licenses for government work.

The go-ethereum repository is a mix of gpl and lgpl, with lgpl used for their importable packages, and gpl just for the actual geth cmd app.

So, the build tag should express that by setting the flag, it's going to include lgpl code.

dex/networks/eth/doc.go
JoeGruffins: editor plugin (go-vim) was using gopls by default, changed to gofmt

dex/networks/eth/abi.go
JoeGruffins: Changed.

client/asset/eth/doc.go
JoeGruffins: I think it's like this now.

dex/networks/eth/doc.go
JoeGruffins: Hopefully fixed.

dex/networks/eth/doc.go
JoeGruffins: Changed to the suggested wording.

client/cmd/dexc/gplimport.go
JoeGruffins: Changed to "lgpl".

chappjc (approved):

PR #1216, opened 2021-09-16
Draft: order options

martonp:
Rough draft uploaded for discussion.

### DIFF
--- a/client/asset/interface.go
+++ b/client/asset/interface.go
@@ -184,6 +184,12 @@ type Wallet interface {
 	SyncStatus() (synced bool, progress float32, err error)
 	// RefundAddress extracts and returns the refund address from a contract.
 	RefundAddress(contract dex.Bytes) (string, error)
+	// SupportedSwapOptions returns a SwapOptions with the non-nil options being the ones
+	// that the wallet supports.
+	SupportedSwapOptions() SwapOptions
+	// SupportedRedeemOptions returns a RedeemOptions with the non-nil options being the ones
+	// that the wallet supports.
+	SupportedRedeemOptions() RedeemOptions
 }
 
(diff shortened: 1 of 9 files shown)


client/asset/interface.go
```diff
@@ -184,6 +184,12 @@ type Wallet interface {
 	SyncStatus() (synced bool, progress float32, err error)
 	// RefundAddress extracts and returns the refund address from a contract.
 	RefundAddress(contract dex.Bytes) (string, error)
+	// SupportedSwapOptions returns a SwapOptions with the non-nil options being the ones
+	// that the wallet supports.
+	SupportedSwapOptions() SwapOptions
+	// SupportedRedeemOptions returns a RedeemOptions with the non-nil options being the ones
+	// that the wallet supports.
+	SupportedRedeemOptions() RedeemOptions
```
buck54321, line 192: What do you plan to use these for? Where does the information end up being displayed?

martonp: They are used in `SupportedOrderOptions` and `ParseSupportedOrderOptions` which are both in `orderoptions.go`. The plan is to call `ParseSupportedOrderOptions` on a `map[string]string`, and only the options which the wallets support will be populated in the resulting `asset.OrderOptions`. `SupportedOrderOptions` will return a slice of options which are supported by a combination of swap/redeem wallets, and this can be used when creating the UI.

martonp: @buck54321 Do you think this approach makes sense? If you agree, I'll implement the options that you have implemented in #1170

buck54321: I've been hacking at #1170 and I'll push an update today. I'd still like to leave **core** out of the picture as much as possible.

PR #1217, opened 2021-09-17
multi: nuke hard-coded 1e8

buck54321:
Get rid of all hard-coded 1e8. When it's a rate encoding factor, use an appropriately named constant. Otherwise, use the asset's `UnitInfo` as reported by the client asset drivers in the `WalletInfo`.


client/core/core.go
```diff
@@ -3904,16 +3903,18 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
...
+		ui := wallets.quoteWallet.Info().UnitInfo
 		subject, details := c.formatDetails(TopicYoloPlaced,
-			float64(corder.Qty)/conversionFactor, unbip(form.Quote), tracker.token())
+			ui.ConventionalString(corder.Qty), ui.Conventional.Unit, tracker.token())
 		c.notify(newOrderNote(TopicYoloPlaced, subject, details, db.Poke, corder))
 	} else {
 		rateString := "market"
 		if form.IsLimit {
-			rateString = strconv.FormatFloat(float64(corder.Rate)/conversionFactor, 'f', 8, 64)
+			rateString = strconv.FormatFloat(float64(corder.Rate)/calc.RateConversionFactor, 'f', 8, 64)
```
JoeGruffins, line 3913: For the purposes of showing the rate to users, will this need to also be divided by `ui.Conventional.Unit`? If I look at the `dcr_btc` market for our default siment I will see `Rate Step: 0.000001 BTC`. If ethereum was the quote asset, it would need one more zero to be correct, as one unit will be 1e9 of one eth.

buck54321: Rates are technically in units of quote/base, e.g. `BTC/DCR` but in units of sats and atoms. Some places we simply list it as the quote asset's symbol like you said, which I feel like is common but technically wrong.

But you did just make me realize that there is a difference between 1) converting the rate from the fractional to integer form for float-free comms, and 2) displaying a conventional rate or parsing one from input. For example, if someone was trading 1 ETH for 1 DCR on the DCR-ETH market, the exchange rate they type into the form would just be `1`. But for official purposes, in atomic units, it would be `1e9/1e8 = 10`, (which would be encoded as `10 * 1e8 = 1e9`). So I guess when displaying and parsing rates, we need to be scaling by the ratio of the conventional conversion factors. Crap.

dex/testing/loadbot/mantle.go
```diff
@@ -361,7 +361,7 @@ func mustJSON(thing interface{}) string {
 }
 
 // Convert the conventional units value to atoms/sats.
-func toAtoms(v float64) uint64 {
+func encodeRate(v float64) uint64 {
 	return uint64(math.Round(v * 1e8))
```
chappjc, line 365: got a 1e8 here still :)
Also about 40 lines down.  If loadbot is an exception for now though, that's fine.

dex/asset.go
```diff
@@ -119,3 +121,30 @@ type Asset struct {
...
+	Alternatives []Denomination `json:"denominations"`
+}
+
+// ConventionalString converts the quantity to conventional units, and returns
+// the formatted float string.
+func (ui *UnitInfo) ConventionalString(v uint64) string {
+	c := ui.Conventional.ConversionFactor
+	prec := int(math.Round(math.Log10(float64(c)))) // Assumes integer powers of 10
+	return strconv.FormatFloat(float64(v)/float64(c), 'f', prec, 64)
+}
```
chappjc, line 150: I like `UnitInfo` a lot.  Good solution.
I think a simple unit test for the `ConventionalString` method would be helpful to explore edges and test formatting.  Mainly I was wondering about the computation of `prec` and how that improves on different formatting like `prec=-1` or `%g`.  I'm not saying it doesn't, just that some tests to ensure certain outputs would be nice.

client/core/core.go
```diff
@@ -3934,6 +3934,13 @@ type walletSet struct {
 	quoteWallet *xcWallet
 }
 
+func (w *walletSet) conventionalRate(atomicRate uint64) float64 {
+	atomicRatio := float64(atomicRate) / calc.RateEncodingFactor
+	baseFactor := float64(w.baseWallet.Info().UnitInfo.Conventional.ConversionFactor)
+	quoteFactor := float64(w.quoteWallet.Info().UnitInfo.Conventional.ConversionFactor)
+	return atomicRatio * (baseFactor / quoteFactor)
```
chappjc, line 3941: Whoah tricky.  Looks good to me though

client/core/core.go
buck54321: Fixed in e6931e9

There are actually three rates encodings of interest. One is the atomic ratio, which is `quoteQty / baseQty` Quantities for this rate are in atomic units. Then there is the message-rate encoding, which is roughly `uint64(atomicRatio * 1e8)`. Then there is the conventional ratio, which is similar to the atomic ratio, but in conventional units. The relationship between the atomic ratio and the conventional ratio is

`conventionalRatio = atomicRatio * (baseConversionFactor / quoteConversionFactor)`

where the conversion factors are the factors for the conventional units.

Users will want to work with the `conventionalRatio`, but we'll always prefer to encode as message-rate on the backend. `SimpleOrder` is updated to use the message-rate encoding too.

client/core/types.go
```diff
@@ -480,27 +480,29 @@ func newDisplayIDFromSymbols(base, quote string) string {
...
-	Qty   float64 `json:"qty"`
-	Rate  float64 `json:"rate"`
-	Epoch uint64  `json:"epoch,omitempty"`
-	Sell  bool    `json:"sell"`
-	Token string  `json:"token"`
+// SimpleOrder is minimal information about an order in a market's order book.
+// Replaced MiniOrder, which had a float Qty in conventional units.
+type SimpleOrder struct {
+	Qty   uint64 `json:"qty"`
+	Rate  uint64 `json:"rate"`
```
chappjc, line 487: So if I were a consumer of the orderbook rpc, which returns `SimpleOrder`s, which functions could I use to get from this `Rate` (message rate or is it atomic rate?) to conventional rate?

If it's too much of a burden for the consumer to get all the conversion factors from the assets' unitinfos, perhaps we have more than one rate field here like `Rate float64` for the conventional rate and `RateAtomic uint64` for the other?

JoeGruffins (commented):
In the file `client/webserver/types.go` there's a helper `precision8(v uint64) string` I see used for rates, which I suppose is good, but also for quantity and values. Is this still correct?

chappjc:
> In the file `client/webserver/types.go` there's a helper `precision8(v uint64) string` I see used for rates, which I suppose is good, but also for quantity and values. Is this still correct?

I don't mind a little code duplication to have a bit less coupling between packages, but your comment does make me wonder if we should do the same trimming of the `rateString` in `prepareTrackedTrade`'s notification.  I don't mind it will full precision that much, but it's something to consider for a more concise "order placed" message.

client/core/bookie.go
```diff
@@ -58,25 +60,32 @@ func (f *BookFeed) Close() {
...
 }
 
 // newBookie is a constructor for a bookie. The caller should provide a callback
 // function to be called when there are no subscribers and the close timer has
 // expired.
 func newBookie(base, quote uint32, logger dex.Logger, close func()) *bookie {
+	// Ignore errors. There's just no way we've gotten this far without
+	// WalletInfo available.
+	baseInfo, _ := asset.Info(base)
+	quoteInfo, _ := asset.Info(quote)
```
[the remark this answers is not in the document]
buck54321: Of course. And I was so cocky about it too.

JoeGruffins:
I thought that I would try to change decred's hard coded ConversionFactor and maybe it would work, but I was unable to pay the registration fee. Is this expected?

client/core/locale_ntfn.go
```diff
@@ -147,27 +147,27 @@ var enUS = map[Topic]*translation{
...
 	TopicRefundFailure: {
 		subject:  "Refund Failure",
-		template: "Refunded %.8f %s on order %s, with some errors",
+		template: "Refunded %s %s on order %s, with some errors",
 	},
 	// [qty, ticker, token]
 	TopicMatchesRefunded: {
 		subject:  "Matches Refunded",
-		template: "Refunded %.8f %s on order %s",
+		template: "Refunded %s %s on order %s",
```
chappjc, line 170: Will have to update ptBR and zhCN in this file as well.

client/core/core_test.go
```diff
@@ -83,8 +83,15 @@ var (
 	tUnparseableHost           = string([]byte{0x7f})
 	tSwapFeesPaid       uint64 = 500
 	tRedemptionFeesPaid uint64 = 350
-	tLogger                    = dex.StdOutLogger("TCORE", dex.LevelTrace)
+	tLogger                    = dex.StdOutLogger("TCORE", dex.LevelCritical)
```
chappjc, line 86: Wow really quieting down Core?  All the logs are oppressive, but when I bother to look at them I often see issues we were sorta ignoring.

chappjc:
> In the file `client/webserver/types.go` there's a helper `precision8(v uint64) string` I see used for rates, which I suppose is good, but also for quantity and values. Is this still correct?

Oh, I think I just digested what you were getting at here.  client/webserver/types.go is still working with 1e8 for a number of different conversions, including rates and amounts.  Follow-up maybe @buck54321?

chappjc:
Testing my grasp on the new UnitInfo, would an update for the usermatches cmd app look like this:

```diff
diff --git a/server/cmd/usermatches/main.go b/server/cmd/usermatches/main.go
index 45be450a..19c8a572 100644
--- a/server/cmd/usermatches/main.go
+++ b/server/cmd/usermatches/main.go
@@ -211,6 +211,17 @@ func mainCore() error {
                return txsplit[0], txsplit[1], nil
        }
 
+       baseAsset := assets[base]
+       if baseAsset == nil {
+               return fmt.Errorf("asset %d not found", base)
+       }
+       baseConversionFactor := baseAsset.UnitInfo().Conventional.ConversionFactor
+       quoteAsset := assets[quote]
+       if quoteAsset == nil {
+               return fmt.Errorf("asset %d not found", quote)
+       }
+       quoteConversionFactor := float64(quoteAsset.UnitInfo().Conventional.ConversionFactor)
+
        _, err = MarketMatchesStreaming(archiver, base, quote, true, -1, func(md *MatchData) error {
                if err := ctx.Err(); err != nil {
                        return err
@@ -240,8 +251,8 @@ func mainCore() error {
                        strconv.FormatUint(md.Epoch.Idx*md.Epoch.Dur/1000, 10),
                        md.MakerAcct.String(),
                        md.TakerAcct.String(),
-                       strconv.FormatInt(int64(md.Quantity)/1e8, 10),
-                       strconv.FormatFloat(float64(md.Rate)/1e8, 'f', -1, 64),
+                       strconv.FormatInt(int64(md.Quantity/baseConversionFactor), 10),
+                       strconv.FormatFloat(float64(md.Rate)/quoteConversionFactor, 'f', -1, 64),
                        strconv.FormatBool(md.TakerSell),
```

I mean it compiles, but would that be your intended usage?

client/core/core_test.go
buck54321: Whoops. I meant to undo this.

buck54321:
> Testing my grasp on the new UnitInfo, would an update for the usermatches cmd app look like this:
> 
> ```diff
> diff --git a/server/cmd/usermatches/main.go b/server/cmd/usermatches/main.go
> index 45be450a..19c8a572 100644
> --- a/server/cmd/usermatches/main.go
> +++ b/server/cmd/usermatches/main.go
> @@ -211,6 +211,17 @@ func mainCore() error {
>                 return txsplit[0], txsplit[1], nil
>         }
>  
> +       baseAsset := assets[base]
> +       if baseAsset == nil {
> +               return fmt.Errorf("asset %d not found", base)
> +       }
> +       baseConversionFactor := baseAsset.UnitInfo().Conventional.ConversionFactor
> +       quoteAsset := assets[quote]
> +       if quoteAsset == nil {
> +               return fmt.Errorf("asset %d not found", quote)
> +       }
> +       quoteConversionFactor := float64(quoteAsset.UnitInfo().Conventional.ConversionFactor)
> +
>         _, err = MarketMatchesStreaming(archiver, base, quote, true, -1, func(md *MatchData) error {
>                 if err := ctx.Err(); err != nil {
>                         return err
> @@ -240,8 +251,8 @@ func mainCore() error {
>                         strconv.FormatUint(md.Epoch.Idx*md.Epoch.Dur/1000, 10),
>                         md.MakerAcct.String(),
>                         md.TakerAcct.String(),
> -                       strconv.FormatInt(int64(md.Quantity)/1e8, 10),
> -                       strconv.FormatFloat(float64(md.Rate)/1e8, 'f', -1, 64),
> +                       strconv.FormatInt(int64(md.Quantity/baseConversionFactor), 10),
> +                       strconv.FormatFloat(float64(md.Rate)/quoteConversionFactor, 'f', -1, 64),
>                         strconv.FormatBool(md.TakerSell),
> ```
> 
> I mean it compiles, but would that be your intended usage?

Yes, but that `FormatInt(int64(md.Quantity)/1e8, 10)` seems wrong to begin with. Should be `FormatFloat`?

PR #1218, opened 2021-09-17
eth/client: PreSwap

martonp:
Part of #1154

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -161,6 +161,9 @@ type ExchangeWallet struct {
 	currentTip *types.Block
 
 	acct *accounts.Account
+
+	cachedInitGas    uint64
+	cachedInitGasMtx sync.Mutex
 }
 
 // Info returns basic information about the wallet and asset.
@@ -300,8 +303,18 @@ func (eth *ExchangeWallet) Balance() (*asset.Balance, error) {
 	return bal, nil
 }
 
-// getInitGas gets an estimate for the gas required for initiating a swap.
+// getInitGas gets an estimate for the gas required for initiating a swap. The
+// result is cached and reused in subsequent requests.
+//
+// This function will return an error if the wallet has a non zero balance.
 func (eth *ExchangeWallet) getInitGas() (uint64, error) {
+	eth.cachedInitGasMtx.Lock()
+	defer eth.cachedInitGasMtx.Unlock()
+
+	if eth.cachedInitGas > 0 {
+		return eth.cachedInitGas, nil
+	}
+
 	var secretHash [32]byte
 	copy(secretHash[:], encode.RandomBytes(32))
 	parsedAbi, err := abi.JSON(strings.NewReader(swap.ETHSwapABI))
@@ -319,7 +332,12 @@ func (eth *ExchangeWallet) getInitGas() (uint64, error) {
 		Gas:   0,
 		Data:  data,
 	}
-	return eth.node.estimateGas(eth.ctx, msg)
+	gas, err := eth.node.estimateGas(eth.ctx, msg)
+	if err != nil {
+		return 0, err
+	}
+	eth.cachedInitGas = gas
+	return gas, nil
 }
 
 // MaxOrder generates information about the maximum order size and associated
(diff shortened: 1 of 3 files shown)


client/asset/eth/eth.go
```diff
@@ -304,21 +307,48 @@ func (eth *ExchangeWallet) MaxOrder(lotSize uint64, feeSuggestion uint64, nfo *d
...
+	return est, nil
+}
+
+// PreSwap gets order estimates based on the available funds and the wallet
+// configuration.
+func (eth *ExchangeWallet) PreSwap(req *asset.PreSwapForm) (*asset.PreSwap, error) {
+	maxEst, err := eth.MaxOrder(req.LotSize, req.FeeSuggestion, req.AssetConfig)
+	if err != nil {
+		return nil, err
+	}
```
JoeGruffins, line 331: `PreSwap` only uses `maxEst.Lots`, which is fine I guess, but does do extra calculations. It looks like you could stick the lots check in estimateSwap pretty easily. It gets the InitGas anyway.

client/asset/eth/eth.go
```diff
@@ -331,13 +361,32 @@ func (eth *ExchangeWallet) MaxOrder(lotSize uint64, feeSuggestion uint64, nfo *d
...
+func (eth *ExchangeWallet) getInitGas() (uint64, error) {
+	eth.cachedInitGasMtx.RLock()
+	cachedInitGas := eth.cachedInitGas
+	eth.cachedInitGasMtx.RUnlock()
+	if cachedInitGas != nil {
+		return *cachedInitGas, nil
+	}
+
+	eth.cachedInitGasMtx.Lock()
+	defer eth.cachedInitGasMtx.Unlock()
```
JoeGruffins, line 378: I believe it would be better to hold the write mtx the whole time. I say so because the read would be super quick anyway, and the small gap here allows a space for multiple callers to go for the expensive node calls needlessly, which defeats the benifit of using a mtx over an atomic. imo

client/asset/eth/eth.go
```diff
@@ -331,13 +361,32 @@ func (eth *ExchangeWallet) MaxOrder(lotSize uint64, feeSuggestion uint64, nfo *d
...
+
+	now := time.Now().Unix()
+	var secretHash [32]byte
+	copy(secretHash[:], encode.RandomBytes(32))
+	initGas, err := eth.node.initiateGas(eth.ctx, now, secretHash, &eth.acct.Address, &mainnetContractAddr)
+	if err != nil {
+		return 0, err
+	}
```
JoeGruffins, line 386: Was it confirmed that a wallet without a certain balance would always fail here? I may be misremembering, but if so can check balance and fail faster. Although it looks like the places where this is called already have some balance check? If so could comment on that assumption.

JoeGruffins (commented):
Looks good, just some nits.

client/asset/eth/eth.go
```diff
@@ -155,6 +155,9 @@ type ExchangeWallet struct {
 	currentTip *types.Block
 
 	acct *accounts.Account
+
+	cachedInitGas    *uint64
+	cachedInitGasMtx sync.RWMutex
```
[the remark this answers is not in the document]
martonp: The reason I used a pointer was that a nil pointer could represent that the cached value had not yet been populated and the init gas still needed to be looked up, but 0 could also represent that.

client/asset/eth/eth.go
martonp: I'll add a comment.

client/asset/eth/eth.go
```diff
@@ -300,8 +303,18 @@ func (eth *ExchangeWallet) Balance() (*asset.Balance, error) {
 	return bal, nil
 }
 
-// getInitGas gets an estimate for the gas required for initiating a swap.
+// getInitGas gets an estimate for the gas required for initiating a swap. The
+// result is cached and reused in subsequent requests.
+//
+// This function will return an error if the wallet has a non zero balance.
```
JoeGruffins, line 309: Will it error if wallet has a zero balance or non-zero? Does it need at least the init gas fee?

JoeGruffins (approved):
Looks reasonable to me.

client/asset/eth/eth.go
martonp: The reason it errors is because initializing a swap requires you to send a non zero value. It doesn't error due to a gas fee. I set the gas price to 0 in `estimateGas`.

chappjc (approved):

PR #1223, opened 2021-09-28
[wip] re-work register flow to accept new assets fee

vctt94:
This PR is the frontend part of (#1202)


client/webserver/site/src/html/register.tmpl
```diff
@@ -39,11 +39,6 @@
       {{template "newWalletForm" "[[[reg_dcr_required]]]"}}
     </form>
```
chappjc, line 40: Seems like the `WalletStep` form is always hidden now too and can be removed as well.

client/webserver/locales/en-us.go
```diff
@@ -174,4 +174,9 @@ var EnUS = map[string]string{
 	"ready":                  "ready",
```
chappjc, line 174: The `reg_dcr_required` and `reg_dcr_unlocked` entries look to be obsolete now.

client/webserver/http.go
```diff
@@ -96,18 +95,14 @@ func (s *WebServer) handleRegister(w http.ResponseWriter, r *http.Request) {
...
+	// feeWalletExists := feeWalletStatus != nil
+	// feeWalletOpen := feeWalletExists && feeWalletStatus.Open
 
 	switch {
 	case !cArgs.UserInfo.Initialized:
 		data.InitStep = true
-	case !feeWalletExists:
-		data.WalletStep = true
-	case !feeWalletOpen:
-		data.OpenStep = true
```
chappjc, line 110: Let's clean up the `registerTmplData` struct, and audit all of frontend for places where these are/were used, and any other obsolete strings.

client/webserver/http.go
```diff
@@ -96,18 +95,14 @@ func (s *WebServer) handleRegister(w http.ResponseWriter, r *http.Request) {
...
 	default:
 		data.DEXStep = true
```
chappjc, line 107: We can simplify things a lot now that wallet setup and unlock are not an automated part of the workflow.  Now: (a) wallet *setup* is being initiated by user interaction on the fee assets table, and (b) wallet *unlock* happens automatically (in `(*Core).Register`) when the user submits the registration form with their app pass.

Thus, consider:

```go

type registerTmplData struct {
	CommonArguments
	Initialized bool
}

func (s *WebServer) handleRegister(w http.ResponseWriter, r *http.Request) {
...

	data := &registerTmplData{
		CommonArguments: *cArgs,
		Initialized:     cArgs.UserInfo.Initialized,
	}

	s.sendTemplate(w, "register", data)
}
```

Then modify register.tmpl: Only the `Initialized` field is needed now (if initialized, it's implied that it's on the DEX entry step because wallet interactions come after that as I've indicated at the top of this comment).  So opposite semantics to the previous `InitStep` field and no need for any of the other booleans afaict.

client/webserver/locales/en-us.go
```diff
@@ -27,7 +27,7 @@ var EnUS = map[string]string{
 	"Submit":                         "Submit",
 	"Confirm Registration":           "Confirm Registration",
 	"app_pw_reg":                     "Enter your app password to confirm DEX registration.",
-	"reg_confirm_submit":             `When you submit this form, <span id="feeDisplay"></span> DCR will be spent from your Decred wallet to pay registration fees.`, // TODO for multi-asset reg
+	"reg_confirm_submit":             `When you submit this form, <span id="feeDisplay"></span> will be spent from your Decred wallet to pay registration fees.`, // TODO for multi-asset reg
```
chappjc, line 30: can remove TODO comment

client/webserver/site/src/html/forms.tmpl
```diff
@@ -145,6 +145,43 @@
...
   </div>
+  <div class="fs16 mt-4">
+    [[[provided_markets]]]
+    <table class="marketstable">
+      <thead>
+        <tr>
+          <th class="text-center">[[[asset_header]]]</th>
+          <th class="text-center">[[[fee_price_header]]]</th>
+          <th class="text-center">[[[fee_confs_header]]]</th>
+          <th class="text-center">[[[wallet_header]]]</th>
```
chappjc, line 156: Since `wallet_header` is an empty string, can we just skip the translation business and just have no text for this column header?

client/webserver/site/src/html/forms.tmpl
```diff
@@ -145,6 +145,43 @@
...
+              <div data-tmpl="fee"></div>
+            </div>
+          </td>
+          <td class="text-center">
+            <div class="d-flex justify-content-center align-items-center">
+              <div data-tmpl="confs"></div>
+            </div>
+          </td>
+          <td class="text-center">
+            <button id="setupWallet" type="button" class="justify-content-center p-1 selected">[[[SetupWallet]]]</button>
```
chappjc, line 178: For discussion:

image

Ideally we would have the rows be shorter (less vertical space), and I think we could do that with buttons that just said "Choose" and "Setup", or perhaps just by making the buttons wider.  There is plenty of horizontal room because the values in the Fee and Confs columns don't take much space.

client/webserver/site/src/js/forms.js
```diff
@@ -298,31 +299,70 @@ export class WalletConfigForm {
...
+      Doc.tmplElement(tr, 'asseticon').src = Doc.logoPath(asset)
+      Doc.tmplElement(tr, 'asset').innerText = asset.toUpperCase()
+      Doc.tmplElement(tr, 'confs').innerText = fee.confs
+      Doc.tmplElement(tr, 'fee').innerText = fee.amount / 1e8
+
+      const haveWallet = app.user.assets[fee.id] ? app.user.assets[fee.id].wallet : null
+      const setupWallet = tr.querySelector('#setupWallet')
+      if (haveWallet) {
+        setupWallet.innerText = intl.prep(intl.ID_CHOOSE_WALLET)
+        Doc.bind(setupWallet, 'click', () => this.setWalletFee(fee.id))
```
chappjc, line 353: I think this action should also update a something to indicate the selection was made and that clicking "Register" will in fact use the intended asset and amount.

For this I had envisioned either or both of: (1) modify the table row's style or just the clicked this button to show it is selected (sorry if this is already done, I haven't tested), which would obviously require undoing this styling change if the user clicks another row, or (2) Update a new sentence just above the Password field to say something like "This will pay a fee of [X amount] [Y asset symbol]."

Also, the styling of the button would be different (perhaps color) between Choose and Setup

chappjc (changes requested):
Thanks for taking this on @vctt94!

Here's a quick first pass with some change I think can be made, but I've not actually tested it yet.
Overall, looking nice.

vctt94:
Thanks for your review chapp.

I changed the table row now for being clickable instead of the button, and made them with low oppacity, so when the are selected they fade in.

image

image

client/webserver/site/src/html/register.tmpl
vctt94: Actually this form stills in use, when setting the wallet.

client/webserver/locales/en-us.go
```diff
@@ -27,7 +27,7 @@ var EnUS = map[string]string{
 	"Submit":                         "Submit",
 	"Confirm Registration":           "Confirm Registration",
 	"app_pw_reg":                     "Enter your app password to confirm DEX registration.",
-	"reg_confirm_submit":             `When you submit this form, <span id="feeDisplay"></span> DCR will be spent from your Decred wallet to pay registration fees.`, // TODO for multi-asset reg
+	"reg_confirm_submit":             `When you submit this form, it will spent from your chosen wallet to pay the registration fee.`,
```
chappjc, line 30:
```suggestion
	"reg_confirm_submit":             `When you submit this form, it will spend from your chosen wallet to pay the registration fee.`,
	"accepted_fee_assets":            "This DEX accepts fee payment with the following assets. Select one:",
```

client/webserver/locales/en-us.go
```diff
@@ -131,8 +131,7 @@ var EnUS = map[string]string{
 	"Set App Password":          "Set App Password",
 	"reg_set_app_pw_msg":        "Set your app password. This password will protect your DEX account keys and connected wallets.",
 	"Password Again":            "Password Again",
-	"reg_dcr_required":          "Your Decred wallet is required to pay registration fees.", // TODO
-	"reg_dcr_unlocked":          "Unlock your Decred wallet to pay registration fees.",      // TODO
+	"reg_dcr_required":          "A wallet is required to pay registration fees. You can set one with this form.",
```
chappjc, line 134: The key name and text I think should change.

```suggestion
	"reg_wallet_required":       "A wallet is required to pay registration fees.",
```

Although I'm thinking this text is not even required on the form at all since the user just got here from the register form by clicking the Setup button, so they know why they are on this form.

client/webserver/site/src/css/forms.scss
```diff
@@ -19,4 +19,17 @@ table.marketstable {
...
+    opacity: 0.5;
+  }
+
+  tr.clickable:hover {
+    opacity: 1;
+  }
+
+  tr.clickable.selected {
+    opacity: 1;
+  }
```
chappjc, line 34: I got the button padding and the selected row border with the following, but you can do it any way you like:

```suggestion
  button {
    padding: 4px 12px;
  }

  tr.clickable {
    cursor: pointer;
    opacity: 0.5;
  }

  tr.clickable:hover {
    opacity: 1;
  }

  tr.clickable.selected {
    opacity: 1;
    border: 1px solid #2d528c;
    border-radius: 1px;
  }
```

client/webserver/site/src/html/forms.tmpl
```diff
@@ -145,6 +145,44 @@
     <span id="appPassSpan">[[[app_pw_reg]]]</span>
     [[[reg_confirm_submit]]]
   </div>
+  <div class="fs16 mt-4">
+    [[[provided_markets]]]
```
chappjc, line 149:
```suggestion
    [[[accepted_fee_assets]]]
```

client/webserver/site/src/html/forms.tmpl
```diff
@@ -145,6 +145,44 @@
...
+            </div>
+          </td>
+          <td class="text-center">
+            <button id="setupWallet" type="button" class="justify-content-center w-100 selected"></button>
+            <span id="walletReady"></span>
+          </td>
+        </tr>
+      </tbody>
+    </table>
+    [[[lot_size_headsup]]]
```
chappjc, line 184: Remove this one.
```suggestion
```

client/webserver/site/src/html/register.tmpl
```diff
@@ -35,17 +35,12 @@
...
 
     {{- /* Set up the Decred wallet. Only shown if user has not connected their Decred wallet. */ -}}
-    <form class="card mx-auto my-5 bg1{{if not .WalletStep}} d-hide{{end}}" id="newWalletForm">
+    <form class="card mx-auto my-5 bg1 d-hide" id="newWalletForm">
```
chappjc, line 38: The comment and the translation tag:

```
    {{- /* Form to set up a wallet. Shown on demand when a user clicks a setupWallet button. */ -}}
    <form class="card mx-auto my-5 bg1 d-hide" id="newWalletForm">
      {{template "newWalletForm" "[[[reg_wallet_required]]]"}}
    </form>
```

client/webserver/site/src/js/forms.js
```diff
@@ -298,31 +299,82 @@ export class WalletConfigForm {
...
+
+  setWalletFee (assetId) {
+    this.feeWallet = assetId
   }
 
   /*
    * setExchange populates the form with the details of an exchange.
    */
   setExchange (xc) {
+    // store xc in case we need to refresh the data, like after setupping
```
chappjc, line 327: setting up

client/webserver/site/src/html/register.tmpl
```diff
@@ -35,17 +35,12 @@
...
       {{template "newWalletForm" "[[[reg_dcr_required]]]"}}
```
chappjc, line 39: A UX issue with this form is that when you get here from the Confirm Registration form by clicking Setup, there is no way back unless you complete the wallet setup.  Ideally you'd be able to "X" the form and it would take you back.

chappjc (changes requested):
I like the direction this is going.  I think it could look more like this:

image

Selected:

image

client/webserver/site/src/css/forms.scss
vctt94: I like this idea of the border, but I don't really like this blue. 

Tried something more gray, with #686767. what do you think?

Also removed border-radius, as it is not really doing anything

chappjc: Sure, gray sounds good.  I'll check it out in both dark and light modes to be sure, but I don't have a strong pref.

I noticed the radius wasn't working for me either but didn't look into it.  A little softer boundary would be nice, but it's fine.

client/webserver/locales/en-us.go
chappjc: Since we went with getting rid of this as an argument to the `"newWalletForm"` template, it's the right time to clean up that template definition because it no longer gets an input from any callers and the template only has one possible rendering now.

That is, when called like `{{template "newWalletForm" "reg reg reg"}}`, we see:

image

When called like `{{template "newWalletForm"}}`, as it is now:

image

The two differences are obviously this text at the top, but also the text about the wallet password goes into a tooltip:

image

So in forms.html where we have `{{define "newWalletForm"}}`, there are a bunch of `if`s using `.` that will never be true anymore since only this register form was ever executing this template with an arg.

Regarding the wallet password texts, we could just go with the existing `w_password_tooltip`, or combine it with the text from `w_password_helper`.  i.e.  `"w_password_tooltip":             "This is the password you have configured with your wallet software. Leave the password empty if there is no password required for the wallet.",`

image

client/webserver/site/src/js/register.js
```diff
@@ -43,33 +42,11 @@ export default class RegistrationPage extends BasePage {
     // NEW DCR WALLET
     // This form is only shown if there is no DCR wallet yet.
```
chappjc, line 43: Pls remove these two comments now that they are inaccurate.

client/webserver/site/src/js/register.js
```diff
@@ -161,7 +140,7 @@ export default class RegistrationPage extends BasePage {
     this.auth()
```
chappjc, line 140: A couple lines up there is `if (seed) this.walletForm.setRegMsg('Your Decred wallet is required')`.
I'm struggling to recall why restoring from seed means the Decred wallet is required.
@buck54321

client/webserver/site/src/js/forms.js
chappjc: I'm pretty happy now with how solution (1) is implemented.  With a boundary around the selected asset's row, it's quite clear what is going to happen when you submit the form.

Maybe another reviewer will think the text proposed in (2) is also valuable, but I can live without it.

client/webserver/site/src/js/register.js
JoeGruffins: It would make sense if wallet was required when not restoring from seed, because pay fee.

chappjc: Well in any case, seems like this conditional message is no longer correct.

buck54321 (changes requested):
First pass. Looking pretty good.

---

When I register from scratch, and choose to "remember password", the DEX address form shows the app password input. 



That input used to be shown only when landing on the registration page with the app already initialized and a Decred wallet already connected, and no password cached. 

---

If I select an asset with no wallet, and click the "Register" button, I just get an error that says "no wallet". 



Maybe in this case we should just show the wallet setup form for the selected asset, as if they'd clicked the "Setup" button.

chappjc:
> Maybe in this case we should just show the wallet setup form for the selected asset, as if they'd clicked the "Setup" button

Or even gray out the Register button if a row is highlighted that is not a set up wallet?

client/webserver/site/src/js/forms.js
```diff
@@ -350,12 +409,22 @@ export class ConfirmRegistrationForm {
   /*
    * submitForm is called when the form is submitted.
    */
-  async submitForm () {
+  async submitForm (assetId) {
     const fields = this.fields
+    const symbol = Doc.assetById(assetId)
+    if (!assetId || !symbol) {
```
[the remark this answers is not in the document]
vctt94: `!==`, right?

vctt94:
I went with graying the button cause I think it was a simpler solution

client/webserver/site/src/html/forms.tmpl
```diff
@@ -38,10 +38,6 @@
   <div class="form-closer hoverbg"><span class="ico-cross"></span></div>
 </div>
 <div class="p-4">
-  {{if .}}
```
chappjc, line 41: There are a bunch more `if`s below looking at `.`

vctt94:
whoops. mb

client/webserver/site/src/js/forms.js
```diff
@@ -298,31 +302,87 @@ export class WalletConfigForm {
...
-    const dcr = this.fees.dcr // TODO: display all options and give choice
-    fields.feeDisplay.textContent = `${Doc.formatCoinValue(dcr.amount / 1e8)} ${app.assets[dcr.id].symbol.toUpperCase()}`
     while (fields.marketsTableRows.firstChild) {
       fields.marketsTableRows.removeChild(fields.marketsTableRows.firstChild)
     }
+    while (fields.feeTableRows.firstChild) {
+      fields.feeTableRows.removeChild(fields.feeTableRows.firstChild)
+    }
+    for (const [symbol, fee] of Object.entries(xc.regFees)) {
+      const haveWallet = app.user.assets[fee.id].wallet
```
chappjc, line 342: This is not easy to test, but if the user's dexc does not support an asset that the server does and does for fees, this will be an error of the sort `"can't access property "wallet", ...user.assets[l.id] is undefined"`. 

If `app.user.assets[fee.id]` is `undefined`, this loop could either: (1) continue to next entry in `regFee` and not show the row, or (2) show the row somehow with N/A and unsupported in appropriate places.  For this iteration, let's go with (1).

client/webserver/site/src/js/forms.js
```diff
@@ -298,31 +302,91 @@ export class WalletConfigForm {
...
+    }
+    for (const [symbol, fee] of Object.entries(xc.regFees)) {
+      // if asset fee is not supported by the client we can skip it.
+      if (app.user.assets[fee.id] === undefined) {
+        continue
+      }
+      const haveWallet = app.user.assets[fee.id].wallet
+      const tr = fields.feeRowTemplate.cloneNode(true)
+      Doc.bind(tr, 'click', () => {
+        this.setWalletFee(fee.id)
```
buck54321, line 349: See other comment about `setFeeWallet`, but here you can just do `this.feeWallet = assetID`, or `this.feeAssetID = assetID` if you go with my naming suggestion.

client/webserver/site/src/js/forms.js
```diff
@@ -350,12 +414,22 @@ export class ConfirmRegistrationForm {
   /*
    * submitForm is called when the form is submitted.
    */
-  async submitForm () {
+  async submitForm (assetID) {
```
buck54321, line 417: Don't need to pass in the `assetID`. Get it from `this.feeWallet`.

client/webserver/site/src/js/forms.js
```diff
@@ -298,31 +302,91 @@ export class WalletConfigForm {
...
     this.form = form
-    bind(form, this.fields.submitConfirm, () => this.submitForm())
+    this.setupWalletFn = setupWalletFn
+    this.feeWallet = null
+    bind(form, this.fields.submitConfirm, () => this.submitForm(this.feeWallet))
+  }
+
+  setWalletFee (assetID) {
+    this.feeWallet = assetID
   }
```
buck54321, line 324: Lots going on with the naming here. Is it a fee or a wallet or an asset ID? How about renaming `feeWallet` to `feeAssetID` or similar, and getting rid of `setWalletFee` and just set the field directly.

client/webserver/site/src/js/forms.js
```diff
@@ -298,31 +302,91 @@ export class WalletConfigForm {
...
     const fields = this.fields
     this.fees = xc.regFees
-    const dcr = this.fees.dcr // TODO: display all options and give choice
-    fields.feeDisplay.textContent = `${Doc.formatCoinValue(dcr.amount / 1e8)} ${app.assets[dcr.id].symbol.toUpperCase()}`
     while (fields.marketsTableRows.firstChild) {
       fields.marketsTableRows.removeChild(fields.marketsTableRows.firstChild)
     }
+    while (fields.feeTableRows.firstChild) {
+      fields.feeTableRows.removeChild(fields.feeTableRows.firstChild)
+    }
```
buck54321, line 340: Can you change both of these to use `Doc.empty`?

client/webserver/site/src/js/forms.js
```diff
@@ -298,31 +302,91 @@ export class WalletConfigForm {
...
+    for (const [symbol, fee] of Object.entries(xc.regFees)) {
+      // if asset fee is not supported by the client we can skip it.
+      if (app.user.assets[fee.id] === undefined) {
+        continue
+      }
```
buck54321, line 345: One line is good 

```Javascript
if (app.user.assets[fee.id] === undefined) continue
```

client/webserver/site/src/js/forms.js
```diff
@@ -298,31 +302,91 @@ export class WalletConfigForm {
...
+      const haveWallet = app.user.assets[fee.id].wallet
+      const tr = fields.feeRowTemplate.cloneNode(true)
+      Doc.bind(tr, 'click', () => {
+        this.setWalletFee(fee.id)
+        // remove selected class from all others row.
+        const rows = Array.from(fields.feeTableRows.querySelectorAll('tr.selected'))
+        rows.forEach(row => {
+          row.classList.remove('selected')
+        })
```
buck54321, line 354: One line is good

```Javascript
rows.forEach(row => { row.classList.remove('selected') })
```

client/webserver/site/src/js/register.js
```diff
@@ -156,12 +133,11 @@ export default class RegistrationPage extends BasePage {
       Doc.show(page.appPWErrMsg)
       return
     }
-    if (seed) this.walletForm.setRegMsg('Your Decred wallet is required')
     this.pwCache.pw = pw
     this.auth()
     app.updateMenuItemsDisplay()
     this.walletForm.refresh()
-    await this.changeForm(page.appPWForm, page.newWalletForm)
+    await this.changeForm(page.appPWForm, page.dexAddrForm)
```
buck54321, line 140: I think you need a `this.dexAddrForm.refresh()` here too. If I choose the "remember password" option, the password field is still shown on the dex address form.

buck54321 (changes requested):
When trying to create a wallet from the markets view, I'm getting some errors

image

Same thing from the wallets view.

chappjc:
I was able to add a DCR wallet from the /markets page after registering with BTC

PR #1225, opened 2021-09-29
eth/client/core: Derive account from app seed

martonp:
`asset.Driver` is now updated with a `Create` function, which is used to initialize a seeded wallet, and `(Driver).Setup` has been renamed to `(Driver).Open`. The seed for seeded wallets is derived from the app seed using `blake256(appSeed|asssetID)`. The private key for the eth wallet is derived from this seed using the following BIP32 derivation path: `m/44'/60'/0'/0/0`.

### DIFF
--- a/client/asset/driver.go
+++ b/client/asset/driver.go
@@ -17,7 +17,7 @@ var (
 
 // CreateWalletParams are the parameters for internal wallet creation. The
 // Settings provided should be the same wallet configuration settings passed to
-// Create.
+// Open.
 type CreateWalletParams struct {
 	Seed     []byte
 	Pass     []byte
--- a/client/asset/driver.go
+++ b/client/asset/driver.go
@@ -62,7 +62,7 @@ func Register(assetID uint32, driver Driver) {
 }
 
 // CreateWallet creates a new wallet. This method should only be used once to create a
-// seeded wallet, after which OpenWallet should be used to load and access the wallet.
+// built-in wallet, after which OpenWallet should be used to load and access the wallet.
 func CreateWallet(assetID uint32, seedParams *CreateWalletParams) error {
 	return withDriver(assetID, func(drv Driver) error {
 		return drv.Create(seedParams)
--- a/client/asset/driver.go
+++ b/client/asset/driver.go
@@ -15,11 +15,33 @@ var (
 	drivers    = make(map[uint32]Driver)
 )
 
+// CreateWalletParams are the parameters for internal wallet creation. The
+// Settings provided should be the same wallet configuration settings passed to
+// Create.
+type CreateWalletParams struct {
+	Seed     []byte
+	Pass     []byte
+	Settings map[string]string
+	DataDir  string
+	Net      dex.Network
+}
+
 // Driver is the interface required of all exchange wallets.
 type Driver interface {
-	Setup(*WalletConfig, dex.Logger, dex.Network) (Wallet, error)
+	Create(*CreateWalletParams) error
+	Open(*WalletConfig, dex.Logger, dex.Network) (Wallet, error)
 	DecodeCoinID(coinID []byte) (string, error)
 	Info() *WalletInfo
+}
+
+func withDriver(assetID uint32, f func(Driver) error) error {
+	driversMtx.Lock()
+	defer driversMtx.Unlock()
+	drv, ok := drivers[assetID]
+	if !ok {
+		return fmt.Errorf("asset: unknown driver asset %d", assetID)
+	}
+	return f(drv)
 }
 
 // Register should be called by the init function of an asset's package.
@@ -39,27 +61,28 @@ func Register(assetID uint32, driver Driver) {
 	drivers[assetID] = driver
 }
 
-// Setup sets up the asset, returning the exchange wallet.
-func Setup(assetID uint32, cfg *WalletConfig, logger dex.Logger, network dex.Network) (Wallet, error) {
-	driversMtx.Lock()
-	drv, ok := drivers[assetID]
-	driversMtx.Unlock()
-	if !ok {
-		return nil, fmt.Errorf("asset: unknown asset driver %d", assetID)
-	}
-	return drv.Setup(cfg, logger, network)
+// CreateWallet creates a new wallet. Only use Create for seeded wallet types.
+func CreateWallet(assetID uint32, seedParams *CreateWalletParams) error {
+	return withDriver(assetID, func(drv Driver) error {
+		return drv.Create(seedParams)
+	})
+}
+
+// OpenWallet sets up the asset, returning the exchange wallet.
+func OpenWallet(assetID uint32, cfg *WalletConfig, logger dex.Logger, net dex.Network) (w Wallet, err error) {
+	return w, withDriver(assetID, func(drv Driver) error {
+		w, err = drv.Open(cfg, logger, net)
+		return err
+	})
 }
 
 // DecodeCoinID creates a human-readable representation of a coin ID for a named
 // asset with a corresponding driver registered with this package.
-func DecodeCoinID(assetID uint32, coinID []byte) (string, error) {
-	driversMtx.Lock()
-	drv, ok := drivers[assetID]
-	driversMtx.Unlock()
-	if !ok {
-		return "", fmt.Errorf("asset: unknown asset driver %d", assetID)
-	}
-	return drv.DecodeCoinID(coinID)
+func DecodeCoinID(assetID uint32, coinID []byte) (cid string, err error) {
+	return cid, withDriver(assetID, func(drv Driver) error {
+		cid, err = drv.DecodeCoinID(coinID)
+		return err
+	})
 }
 
 // A registered asset is information about a supported asset.
--- a/client/asset/driver.go
+++ b/client/asset/driver.go
@@ -62,7 +62,7 @@ func Register(assetID uint32, driver Driver) {
 }
 
 // CreateWallet creates a new wallet. This method should only be used once to create a
-// built-in wallet, after which OpenWallet should be used to load and access the wallet.
+// seeded wallet, after which OpenWallet should be used to load and access the wallet.
 func CreateWallet(assetID uint32, seedParams *CreateWalletParams) error {
 	return withDriver(assetID, func(drv Driver) error {
 		return drv.Create(seedParams)
--- a/client/asset/driver.go
+++ b/client/asset/driver.go
@@ -61,7 +61,8 @@ func Register(assetID uint32, driver Driver) {
 	drivers[assetID] = driver
 }
 
-// CreateWallet creates a new wallet. Only use Create for seeded wallet types.
+// CreateWallet creates a new wallet. This method should only be used once to create a
+// seeded wallet, after which OpenWallet should be used to load and access the wallet.
 func CreateWallet(assetID uint32, seedParams *CreateWalletParams) error {
 	return withDriver(assetID, func(drv Driver) error {
 		return drv.Create(seedParams)
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -14,6 +14,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"path/filepath"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -22,9 +23,11 @@ import (
 	"decred.org/dcrdex/client/asset"
 	"decred.org/dcrdex/dex"
 	"decred.org/dcrdex/dex/encode"
+	"decred.org/dcrdex/dex/keygen"
 	swap "decred.org/dcrdex/dex/networks/eth"
 	dexeth "decred.org/dcrdex/server/asset/eth"
 	"github.com/decred/dcrd/dcrutil/v4"
+	"github.com/decred/dcrd/hdkeychain/v3"
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts"
 	"github.com/ethereum/go-ethereum/accounts/abi"
@@ -78,6 +81,7 @@ var (
 		DefaultConfigPath: defaultAppDir, // Incorrect if changed by user?
 		ConfigOpts:        configOpts,
 		UnitInfo:          swap.UnitInfo,
+		Seeded:            true,
 	}
 	mainnetContractAddr = common.HexToAddress("")
 )
@@ -88,8 +92,8 @@ var _ asset.Driver = (*Driver)(nil)
 // Driver implements asset.Driver.
 type Driver struct{}
 
-// Setup creates the ETH exchange wallet. Start the wallet with its Run method.
-func (d *Driver) Setup(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network) (asset.Wallet, error) {
+// Open opens the ETH exchange wallet. Start the wallet with its Run method.
+func (d *Driver) Open(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network) (asset.Wallet, error) {
 	return NewWallet(cfg, logger, network)
 }
 
@@ -105,6 +109,10 @@ func (d *Driver) DecodeCoinID(coinID []byte) (string, error) {
 // Info returns basic information about the wallet and asset.
 func (d *Driver) Info() *asset.WalletInfo {
 	return WalletInfo
+}
+
+func (d *Driver) Create(params *asset.CreateWalletParams) error {
+	return CreateWallet(params)
 }
 
 // rawWallet is an unexported return type from the eth client. Watch for changes at
@@ -181,28 +189,83 @@ func (*ExchangeWallet) Info() *asset.WalletInfo {
 	return WalletInfo
 }
 
+// CreateWallet creates a new internal ETH wallet and stores the private key
+// derived from the wallet seed.
+func CreateWallet(createWalletParams *asset.CreateWalletParams) error {
+	walletDir := getWalletDir(createWalletParams.DataDir, createWalletParams.Net)
+	nodeCFG := &nodeConfig{
+		net:    createWalletParams.Net,
+		appDir: walletDir,
+	}
+	node, err := prepareNode(nodeCFG)
+	if err != nil {
+		return err
+	}
+
+	// m/44'/60'/0'/0/0
+	path := []uint32{hdkeychain.HardenedKeyStart + 44,
+		hdkeychain.HardenedKeyStart + 60,
+		hdkeychain.HardenedKeyStart,
+		0,
+		0}
+	extKey, err := keygen.GenDeepChild(createWalletParams.Seed, path)
+	defer extKey.Zero()
+	if err != nil {
+		return err
+	}
+
+	privateKey, err := extKey.SerializedPrivKey()
+	defer keygen.ZeroBytes(privateKey)
+	if err != nil {
+		return err
+	}
+
+	importAccountToNode(node, privateKey, createWalletParams.Pass)
+	if err != nil {
+		return err
+	}
+
+	return node.Close()
+}
+
 // NewWallet is the exported constructor by which the DEX will import the
 // exchange wallet. It starts an internal light node.
 func NewWallet(assetCFG *asset.WalletConfig, logger dex.Logger, network dex.Network) (*ExchangeWallet, error) {
-	cfg, err := loadConfig(assetCFG.Settings, network)
-	if err != nil {
-		return nil, err
-	}
+	walletDir := getWalletDir(assetCFG.DataDir, network)
 	nodeCFG := &nodeConfig{
 		net:    network,
-		appDir: cfg.AppDir,
+		appDir: walletDir,
 		logger: logger.SubLogger("NODE"),
 	}
-	node, err := runNode(nodeCFG)
+
+	node, err := prepareNode(nodeCFG)
 	if err != nil {
 		return nil, err
 	}
+
+	err = startNode(node, network)
+	if err != nil {
+		return nil, err
+	}
+
+	accounts := exportAccountsFromNode(node)
+	if len(accounts) != 1 {
+		return nil,
+			fmt.Errorf("NewWallet: eth node keystore should only contain 1 account, but contains %v",
+				len(accounts))
+	}
+
 	return &ExchangeWallet{
 		log:          logger,
 		tipChange:    assetCFG.TipChange,
 		internalNode: node,
 		lockedFunds:  make(map[string]uint64),
+		acct:         &accounts[0],
 	}, nil
+}
+
+func getWalletDir(dataDir string, network dex.Network) string {
+	return filepath.Join(dataDir, network.String())
 }
 
 func (eth *ExchangeWallet) shutdown() {
@@ -220,11 +283,6 @@ func (eth *ExchangeWallet) Connect(ctx context.Context) (*sync.WaitGroup, error)
 	}
 	eth.node = &c
 	eth.ctx = ctx
-
-	eth.acct, err = eth.initAccount()
-	if err != nil {
-		return nil, err
-	}
 
 	// Initialize the best block.
 	bestHash, err := eth.node.bestBlockHash(ctx)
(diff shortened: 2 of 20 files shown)


client/core/core.go
```diff
@@ -1726,6 +1729,16 @@ func (c *Core) CreateWallet(appPW, walletPW []byte, form *WalletForm) error {
 		// Balance and Address are set after connect.
 	}
 
+	if walletInfo.IsInternalWallet {
+		encXPrivKey, xPubKey, err := c.deriveWalletSeed(assetID, appPW)
+		if err != nil {
+			return fmt.Errorf("unable to derive wallet seed: %w", err)
```
chappjc, line 1735: Seems very reasonable to me, but you'll want to coordinate with @buck54321 on this approach since he has done the same thing on his btc spv branch, in his own way.  It's very good to get two approaches and agree on the best.

martonp: Ah I'll check it out.

chappjc: https://github.com/buck54321/dcrdex/commit/2165bc06754a40cf66d8e485995e3d60e5168d41#diff-13a4a6b12c1ae844ef90151638f194427cfdc545fceb58d52c0f6f33b81d859bR1637, but this is months behind master now and he probably has a different idea about it given how hd keys have settled and other issues.

buck54321: @martonp has it right.

dex/keygen/keygen.go
```diff
@@ -0,0 +1,129 @@
...
+		return nil, nil, err
+	}
+
+	encXPrivKey, err = EncryptExtendedKey(extKey, pw)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	xPubKey = extKey.Neuter()
+
```
JoeGruffins, line 127: Can Zero extKey here.

dex/keygen/keygen.go
```diff
@@ -0,0 +1,129 @@
...
+		err = fmt.Errorf("expected seed length 64 but got %v", len(seed))
+		return nil, nil, err
+	}
+
+	root, err := hdkeychain.NewMaster(seed, &RootKeyParams{})
+	if err != nil {
+		err = fmt.Errorf("NewMaster error: %w", err)
+		return nil, nil, err
+	}
+
```
JoeGruffins, line 110: Can defer a `root.Zero()`?

JoeGruffins (commented):
A few comments about zeroing sensitive info. For most cases it's probably silly, but it's a preventive measure against "memory scraping". I believe it can be an issue if the client is run on a shared server, like a vps, or some sort of cloud computing, where used memory could be read by other users. imo it doesn't hurt to be safe.

client/asset/interface.go
```diff
@@ -41,6 +41,9 @@ type WalletInfo struct {
 	// description for each option. This can be used to request config info from
 	// users e.g. via dynamically generated GUI forms.
 	ConfigOpts []*ConfigOption `json:"configopts"`
+	// IsInternalWallet is true for a wallet that requires keys to be derived
+	// from the app seed.
+	IsInternalWallet bool
```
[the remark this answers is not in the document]
buck54321: It's usually advantageous to apply the zero value to pre-existing behavior, like is done here. That way, existing consumers don't need to change anything to continue working correctly. If it was `IsExternalWallet`, every existing wallet implementation would need to update their `WalletInfo` to work correctly. I know there are only like 4 of those, but it's good practice anyway. 

I think "internal" is just used in the sense of "built-in" here, like we're doing with bitcoin spv and ethereum. I'm using the term "seeded" e.g. `WalletInfo.IsSeeded bool` in my bitcoin spv work, but it's the same thing.

client/asset/interface.go
```diff
@@ -71,6 +74,15 @@ type WalletConfig struct {
...
 	DataDir string
+	// EncXPrivKey is the extended private key for account 0 of the asset the
+	// wallet is providing. (m/44'/{assetID}'/0') It is encrypted using the wallet
+	// password. This config only needs to be used for internal wallets.
+	EncXPrivKey []byte
+	// XPubKey is the extended public key for account 0 of the asset the
+	// wallet is providing. (m/44'/{assetID}'/0') This is needed so that a locked
+	// wallet can give information about addresses/accounts. This config only
+	// needs to be used for internal wallets.
+	XPubKey string
```
buck54321, line 85: What we're actually interested in is determinism. Some wallets such as btc are deterministic based on supplying the same random seed as an input to wallet creation, where it appears that for geth, you just import the private key. So what we actually need to pass in is a deterministic seed. I agree with the way you derive the private key from the seed, but it should be done inside of **eth**. 

Just fyi, I will be introducing two new methods on `asset.Driver`, `Create` and `Exists`, that will be used to initialize "seeded" wallets, e.g. those with `IsInternalWallet = true`. So you won't pass the seed to `asset.Setup` as part of the `WalletConfig`. You'll just pass it once to the `(*Driver).Create` method.

martonp: I was considering the extended private key that is derived from the app seed using the path `m/44'/{assetID}'/0'` to be the seed for each of the wallets. From this wallet seed, I derive the actual private key used for signing transactions inside of eth. Is the naming I have confusing, or do you think we should pass the app seed into the wallet backends?

client/asset/eth/eth.go
```diff
@@ -450,8 +468,42 @@ func (eth *ExchangeWallet) Address() (string, error) {
...
 func (eth *ExchangeWallet) Unlock(pw string) error {
+	eth.importAccountMtx.Lock()
+	defer eth.importAccountMtx.Unlock()
+
+	if !eth.importedAccount {
+		err := eth.importAccount(pw)
+		if err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
martonp: I put this into unlock because that's the first time the wallet backend has access to the wallet password, which is needed to decrypt the private key that was passed during `Setup`. Although might not need to be here now that @buck54321 is adding some new methods on `asset.Driver`.

client/asset/interface.go
buck54321: The is the right derivation for the key, but when I initialize a btc or dcr wallet, I give the seed and the derivation is done internaly to **btcwallet** or **dcrwallet**. See, for example in dcrwallet, `wallet.Create` → `udb.Initialize` → `createAddressManager` → `deriveCoinTypeKey`. And since we have built-in Bitcoin (and likely Decred) wallets coming any day now, I know we need the seed itself there. 

**eth** will need to handle the key derivation internally.

dex/keygen/keygen.go
```diff
@@ -0,0 +1,129 @@
...
+	decryptedSeed := base58.Encode(decrypted[0:extendedKeyLength])
+
+	return hdkeychain.NewKeyFromString(decryptedSeed, &RootKeyParams{})
+}
+
+// GenDeepChild derives the leaf of a path of children from a root extended key.
+func GenDeepChild(root *hdkeychain.ExtendedKey, kids []uint32) (*hdkeychain.ExtendedKey, error) {
+	genChild := func(parent *hdkeychain.ExtendedKey, childIdx uint32) (*hdkeychain.ExtendedKey, error) {
+		err := hdkeychain.ErrUnusableSeed
+		for err == hdkeychain.ErrUnusableSeed {
```
chappjc, line 73: These are the wrong error codes.  I fixed this recently in https://github.com/decred/dcrdex/pull/1210/files#diff-a0c3b3a03517588f82d5bb3553e3e46ef5377c8bbcb4f8813ebb6d3cb0a6ee86R629-R630

client/asset/interface.go
chappjc: Thanks for clarifying all that @buck54321.  Only thing I would add is that some time in the future I can two different asset-dependent seeding approaches - one for backends like dcrwallet and btcwallet that are good with a seed that is not an extended key at all, and one for assets where it's more convenient to just get an extended key or maybe a vanilla private key.  For now a unified approach is best where each asset takes the seed (not a key) and uses it's own approach to generating keys from that.

On the topic of eth accounts, one thing I'd like to see it the ability to get an account private key out of the eth package that is suitable for import into common Ethereum wallets such as MetaMask.  The internal {0,0} derivation path is probably irrelevant here unless we do intend to generate multiple accounts.  I mention this functionality because I would also like the ability to get the dcr and btc seeds so I can restore those wallets externally, but with eth the seed is meaningless without replicating whatever the eth package does internally to get the private key from a seed.

buck54321: Actually, @jrick recommends BLAKE256 (or BLAKE2 or SHA3, but we already import `github.com/decred/dcrd/crypto/blake256`).

JoeGruffins: BLAKE256(appseed|assetid) makes sense in the context that not all wallets will be able to run with an account key. It's also simple and easy to reproduce.

martonp: A agree also that BLAKE256(appseed|assetid) makes sense.

It looks like there's a good bit of overlap with your PR @buck54321 , so I'm thinking of maybe putting those new `asset.Driver` methods into this PR. This PR would assume that there is only a single wallet type for each asset, and it may be either seeded or not. Seeded wallets have the extra step of being created by calling `asset.Create`. Then your PR would add the ability to have multiple types of wallets for the same asset. Do you think this makes sense, or should I hold off on finishing this until you complete your PR?

Regarding giving the ability to export a seed and importing into an external wallet, the metamask default derivation path is `m/44'/0'/0`, but metamask also supports the ledger live default path which is `m/44'/0'/0/0`. I think either of these would be fine.

buck54321: > Seeded wallets have the extra step of being created by calling asset.Create. Then your PR would add the ability to have multiple types of wallets for the same asset. Do you think this makes sense, or should I hold off on finishing this until you complete your PR?

That sounds about right.

I'll be opening Bitcoin SPV by Monday(ish), and the `Driver` bits are implemented there. You're welcome to do it here so we can get it in earlier though.

client/asset/eth/node.go
```diff
@@ -219,22 +228,41 @@ func runNode(cfg *nodeConfig) (*node.Node, error) {
...
+
+// importAccountToNode imports an account into the ethereum wallet by private key
+// that can be unlocked with password.
+func importAccountToNode(node *node.Node, privateKey, password []byte) error {
+	ecdsaPrivateKey, err := crypto.ToECDSA(privateKey)
+	if err != nil {
+		return err
+	}
+	ks := node.AccountManager().Backends(keystore.KeyStoreType)[0].(*keystore.KeyStore)
+	_, err = ks.ImportECDSA(ecdsaPrivateKey, hex.EncodeToString(password))
```
chappjc, line 257: Since `NewWallet` errors if there is more than one account, seems like `importAccountToNode` should refuse to import if there is already one?
Besides dex not using more than one account, was there a consideration for `NewWallet` erroring for `len != 1` instead of just `len <  1` and using the first one in the slice regardless of its length?

client/core/core.go
```diff
@@ -1765,6 +1779,48 @@ func (c *Core) CreateWallet(appPW, walletPW []byte, form *WalletForm) error {
...
+		return nil, fmt.Errorf("seed decryption error: %w", err)
+	}
+
+	c.log.Infof("Initializing a built-in %s wallet", unbip(assetID))
+
+	b := make([]byte, len(root)+4)
+	copy(b, root)
+	assetB := make([]byte, 4)
+	binary.BigEndian.PutUint32(assetB, assetID)
+	copy(b[len(root):], assetB)
```
chappjc, line 1802: This is updated in the other PR, but this can just be:

```go
b := make([]byte, len(root)+4)
copy(b, root)
binary.BigEndian.PutUint32(b[len(root):], assetID
```

https://github.com/decred/dcrdex/pull/1230#discussion_r728421901

client/core/core.go
```diff
@@ -1765,6 +1779,48 @@ func (c *Core) CreateWallet(appPW, walletPW []byte, form *WalletForm) error {
...
+
+	seed := blake256.Sum256(b)
+	pw := encode.RandomBytes(32)
+	if err = asset.CreateWallet(assetID, &asset.CreateWalletParams{
+		Seed:     seed[:],
+		Pass:     pw[:],
```
chappjc, line 1808: also changed: `Pass: pw,` (no need to re-slice)

client/asset/eth/rpcclient_harness_test.go
```diff
@@ -151,7 +154,24 @@ func TestMain(m *testing.M) {
...
+			DataDir:  testDir,
+			Net:      dex.Simnet,
+		}
+		err = CreateWallet(&createWalletParams)
+		if err != nil {
+			return 1, fmt.Errorf("error creating node: %v\n", err)
+		}
+		wallet, err := NewWallet(&walletConfig, tLogger, dex.Simnet)
 		if err != nil {
 			return 1, fmt.Errorf("error starting node: %v\n", err)
```
chappjc, line 176: Please also remove the `\n` from these error strings

client/asset/eth/node.go
martonp: Yeah `importAccountToNode` could definitely fail to import. Maybe it could even check if the one that's already imported is the same as the one it's attempting to import, and if not, replace it.

I did consider that, but if there are more than one accounts stored, it means something went wrong somewhere. In `NewWallet` we don't have access to the wallet seed, so there's no way make sure that the account that is being loaded is the correct one. This situation should never happen, but maybe in this case we should give the user the option to recreate the wallet?

client/core/core.go
```diff
@@ -1765,6 +1779,48 @@ func (c *Core) CreateWallet(appPW, walletPW []byte, form *WalletForm) error {
...
+	if creds == nil {
+		return nil, fmt.Errorf("no v2 credentials stored")
+	}
+
+	root, err := crypter.Decrypt(creds.EncSeed)
+	if err != nil {
+		return nil, fmt.Errorf("seed decryption error: %w", err)
+	}
+
+	c.log.Infof("Initializing a built-in %s wallet", unbip(assetID))
```
itswisdomagain, line 1796: Can we use this grammar (built-in wallet) generally instead of seeded wallet? The latter sounds confusing to me and I expect it also won't make immediate sense to a regular user, but no hard preferences. I do prefer builtin because all wallets are seeded really, aren't they?

Regardless of which terminology we stick with, I'll recommend consistent messaging in logs and errors, e.g. in `core.CreateWallet`, I saw `external password incompatible with seeded wallet`.

client/core/core.go
```diff
@@ -1765,6 +1779,48 @@ func (c *Core) CreateWallet(appPW, walletPW []byte, form *WalletForm) error {
...
+// password is returned for encrypting and storing.
+func (c *Core) createSeededWallet(assetID uint32, crypter encrypt.Crypter, form *WalletForm) ([]byte, error) {
+	creds := c.creds()
+	if creds == nil {
+		return nil, fmt.Errorf("no v2 credentials stored")
+	}
+
+	root, err := crypter.Decrypt(creds.EncSeed)
+	if err != nil {
+		return nil, fmt.Errorf("seed decryption error: %w", err)
```
itswisdomagain, line 1793: app seed decryption error?

client/core/core.go
```diff
@@ -1765,6 +1779,48 @@ func (c *Core) CreateWallet(appPW, walletPW []byte, form *WalletForm) error {
...
+// createSeededWallet initializes a seeded wallet with an asset-specific seed
+// derived deterministically from the app seed and a random password. The
+// password is returned for encrypting and storing.
+func (c *Core) createSeededWallet(assetID uint32, crypter encrypt.Crypter, form *WalletForm) ([]byte, error) {
+	creds := c.creds()
+	if creds == nil {
+		return nil, fmt.Errorf("no v2 credentials stored")
+	}
+
+	root, err := crypter.Decrypt(creds.EncSeed)
```
itswisdomagain, line 1791: appSeed instead of root.

client/core/core.go
```diff
@@ -1765,6 +1779,48 @@ func (c *Core) CreateWallet(appPW, walletPW []byte, form *WalletForm) error {
 	return nil
 }
 
+// createSeededWallet initializes a seeded wallet with an asset-specific seed
+// derived deterministically from the app seed and a random password. The
```
itswisdomagain, line 1783: Outdated comment? Don't see a random password being used in deriving the asset-specific seed. Also makes me wonder how deterministic it is if the derivation uses a randomly generated input.

client/core/core.go
```diff
@@ -1954,6 +2011,11 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, assetID uint32, cfg
 		EncryptedPW: oldWallet.encPW(),
 		Address:     oldDepositAddr,
 	}
+	seeded := oldWallet.Info().Seeded
+	if seeded && len(newWalletPW) > 0 {
+		return newError(passwordErr, "cannot set a password on a seeded wallet")
+	}
```
itswisdomagain, line 2017: Can move this to before the `oldDepositAddr := oldWallet.currentDepositAddress()` line. Prefer to perform checks like these early on before creating/assigning variables that will not be used if the check fails.

client/asset/eth/eth.go
```diff
@@ -107,6 +111,10 @@ func (d *Driver) Info() *asset.WalletInfo {
 	return WalletInfo
 }
 
+func (d *Driver) Create(params *asset.CreateWalletParams) error {
```
itswisdomagain, line 114: Possible to move this to before `Open`? Also in other asset files please, to give visual representation of the order in which the methods are/should be called.

itswisdomagain (changes requested):
I did not get into the eth-specific changes and a couple nitpicks here, please forgive.

client/core/core.go
chappjc: This refers to the `pw := encode.RandomBytes(32)` fed into `CreateWallet` with the deterministic seed as well.  The `pw` is arbitrary, for whatever purposes the wallet backend needs it for, be it encryption or whatetever, while the seed is the deterministic part for key/address generation.

client/core/core.go
```diff
@@ -1780,10 +1836,11 @@ func (c *Core) loadWallet(dbWallet *db.Wallet) (*xcWallet, error) {
 			go c.tipChange(assetID, err)
 		},
 	}
+
 	logger := c.log.SubLogger(unbip(assetID))
-	w, err := asset.Setup(assetID, walletCfg, logger, c.net)
+	w, err := asset.OpenWallet(assetID, walletCfg, logger, c.net)
 	if err != nil {
-		return nil, fmt.Errorf("error creating wallet: %w", err)
+		return nil, fmt.Errorf("error opening wallet: %w", err)
```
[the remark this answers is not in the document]
chappjc: Very true that the naming should reflect the purpose.  In that regard, I believe the previous `Setup` was too vague, as it was before built-in/dex-managed wallets, when the wallet must have already existed externally and we just wanted to connect/open it.

So here's the change as I understand it:
- `Setup` -> `Open` -- applies to both external and internal wallets
- [new] `Create` -- only applies to internal wallets, which must necessarily be created by the dex process.

The btc.go example makes sense to me I think: https://github.com/decred/dcrdex/pull/1225/files#diff-1b07e494f4bdd99b2a617ac110fbadc081e5b6ae1abf23f64032bbee5ebebef1R332-R340

client/core/core.go
martonp: The phrasing is just confusing.

chappjc: TBH, the bit about the password is not the concern of the caller of this function, and thus not totally necessary to be in the docs for the function.  The random password can be explained in the code if needed.

client/core/core.go
martonp: I think built-in makes more sense too. All wallets are seeded.

dex/keygen/keygen.go
```diff
@@ -0,0 +1,60 @@
...
+
+	return extKey, nil
+}
+
+// ZeroBytes assigns 0 to each element in a byte slice.
+func ZeroBytes(b []byte) {
+	for i := 0; i < len(b); i++ {
+		b[i] = byte(0)
+	}
+}
```
buck54321, line 60: We already have `encode.ClearBytes`.

client/core/core.go
```diff
@@ -1708,6 +1704,24 @@ func (c *Core) CreateWallet(appPW, walletPW []byte, form *WalletForm) error {
 		}
 	}
 
+	if walletInfo.BuiltIn {
+		if len(walletPW) > 0 {
+			return fmt.Errorf("external password incompatible with built in wallet")
```
buck54321, line 1709: **built-in** should be hyphenated.

buck54321 (changes requested):

client/asset/interface.go
```diff
@@ -41,6 +41,9 @@ type WalletInfo struct {
 	// description for each option. This can be used to request config info from
 	// users e.g. via dynamically generated GUI forms.
 	ConfigOpts []*ConfigOption `json:"configopts"`
+	// BuiltIn is true for built-in wallets. Built-in wallets must be created
+	// with a deterministic seed using (Driver).Create before it can be opened.
+	BuiltIn bool
```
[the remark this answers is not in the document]
chappjc: We're having trouble ironing this out because we're mixing up the purpose in _code_ with what the _user_ should know and see.  This flag is used by Core/consumers to know when to use `CreateWallet` (in Core's case by making a seed and dummy password, etc for `CreateWalletParams`).  In this regard both "created" and "seeded" would be fine.

But TBH I'm pretty darn apathetic about the variable name at this point.  I only want to convey to the user in errors and logs that this is talking about a wallet created and managed by DEX, not that it is "seeded".  Since most external wallets are seeded too, such messages will be super confusing.

So, we can change the variable back as `Seeded` that's fine, let's just decide and move on.  We can change the variable names any time to match changing semantics, but presently this actually creates a wallet stored at the provided DataDir with the provided Seed.

I don't care so please clarify @buck54321

client/core/core.go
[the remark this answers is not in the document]
itswisdomagain: Yeah, it's also my understanding that `Setup` -> `Open` applies to both external and internal wallets while the new `Create` only applies to internal wallets.

While `Create` actually creates a wallet, does `Open` actually open it? If so, in what sense? Looking at the implementations of this method by most of the assets, I don't see a wallet being "opened", the implementations mostly just create and return an unconnected instance of an `ExchangeWallet`.

I renamed `Setup` to `LoadWallet` in https://github.com/decred/dcrdex/commit/4a920aef2cbcbf73080e0a1b0c3d442d14087cd2, and that seems to agree with how Core uses the `asset.Setup/OpenWallet` function; the function is called from `core.loadWallet`. Also looking at btc's implementation of the method for an inbuilt wallet, I see `loadSPVWallet` which tells me that this method really just creates and loads an `ExchangeWallet` into memory, which Core can then run/start using the wallet's `Connect` method.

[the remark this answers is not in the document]
itswisdomagain: Works for me, will get to reviewing those more thoroughly now.

client/asset/driver.go
```diff
@@ -15,13 +15,35 @@ var (
 	drivers    = make(map[uint32]Driver)
 )
 
+// CreateWalletParams are the parameters for internal wallet creation. The
+// Settings provided should be the same wallet configuration settings passed to
+// Create.
```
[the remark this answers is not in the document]
itswisdomagain: Still confused. Doesn't `The settings provided` refer to the `Settings` field of this `CreateWalletParams` struct? Isn't that also automatically the exact same settings passed to `Create`, by virtue of being a part of this struct that is passed to `Create`?

client/asset/eth/eth.go
```diff
@@ -81,7 +81,7 @@ var (
 		Units:             "gwei",
 		DefaultConfigPath: defaultAppDir, // Incorrect if changed by user?
 		ConfigOpts:        configOpts,
-		Seeded:            true,
+		BuiltIn:           true,
```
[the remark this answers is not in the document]
itswisdomagain: Maybe I'm misunderstanding this, but what I think of when I see builtin, is created or seeded by us, regardless of whether the actual creation happened via RPC or not. External to me, means created outside of the DEX, we're only just connecting to and using it. Seeded to me, means this wallet was created from a seed, not necessarily that the seed was provided by the DEX. It also doesn't mean to me that the wallet _should be_ seeded, just that it is seeded.

In any case, I agree the priority is consistency in user-facing messages. But we shouldn't also neglect wordings used internally in comments and variable names.

client/asset/interface.go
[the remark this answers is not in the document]
itswisdomagain: > This flag is used by Core/consumers to know when to use CreateWallet.

Not to push it and certainly not a reason to hold back any of the PRs because of this variable name (as long as the user-facing messages are clear and user-friendly), but I do want to add that if this is the purpose of the flag, then `Seeded` doesn't do it for me. As I said in a previous comment, what comes to mind when I see Seeded is, this wallet was created with a seed, not that it even needs to be created with a seed. Maybe that's just me though.

client/asset/eth/eth.go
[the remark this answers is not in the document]
martonp: Maybe `SeededByAppSeed` would clear up all confusion?

buck54321:
Thanks @martonp. I really didn't intend this to be such a thing 😂. After #1230 is in, we can talk about refactoring things if names are still unclear. We've just got a lot of moving pieces right now, so let's not try to muddy things up too much.

PR #1226, opened 2021-09-29
server/db/driver/pg: log postgres notices

chappjc:
This updates the postgresql driver, lib/pq, from 1.2.0 to 1.10.3. This contains no breaking changes.  It is primarily some bug fixes, also a few context enabled methods, new array types, and support for handling postgres notices.

Notices from postgres are now handled by logging them according to their severity and error code class using the new `pq.ConnectorWithNoticeHandler` `Connector` factory function. There is some ambiguity in severity levels, but this will at least log anything important that was previously ignored.

'pgonline' tests are still passing.

In addition to the pq update and notice logging, remove an old pseudo-upgrade.

This pseudo-upgrade was added prior to the first release and prior to schema versioning that would create the `matches.forgiven` column if it does not exist.  It has existed in the `CREATE TABLE` statement since first release so this `ALTER TABLE` is always a no-op that just results in a logged `NOTICE`:

`[DBG] DB: pq: NOTICE (42701) - duplicate_column: column "forgiven" of relation "matches" already exists, skipping`

The comments around this indicate the same, that this should have been removed when actual upgrades were implemented. Schema v1 does indeed include this `forgiven` column, and this change can thus remove the `ALTER TABLE` statement that tries to re-add it on every startup.

The `forgiven` column was on v0.1.0: https://github.com/decred/dcrdex/commit/73d04b4caa6565dd4ac939f242704aa923e13ef4


JoeGruffins (approved):

buck54321 (approved):
Looks good. Please run a `go mod tidy` on loadbot.

PR #1227, opened 2021-10-02
client/asset/dcr: refactor to support non-rpc wallets

itswisdomagain:
Introduce a `Wallet` interface for actual wallet functionality to
be implemented by consumers as desired.

`rpcWallet` type is added to implement the new `Wallet` interface using
json-rpc commands over an rpc connection. This maintains the existing
state of the dcr ExchangeWallet while allowing external consumers
implement other types that do not require an rpc connection.


chappjc:
@itswisdomagain and I chatted about this and I think it's a great idea.  It opens the door to other Decred wallets like https://github.com/planetdecred/godcr, maybe something mobile long-term, maybe a built-in, etc.  It takes an approach to a `Wallet interface` similar to what was done in #1089 , and it generally seems like a clean change that doesn't affect the existing RPC wallets.

martonp (approved):
Looks good to me.

client/asset/dcr/rpcwallet.go
```diff
@@ -0,0 +1,537 @@
...
+	}
+
+	// The websocket client is connected now, so if any of the following checks
+	// fails and we return with a non-nil error, we must shutdown the rpc client
+	// or subsequent reconnect attempts will be met with "websocket client has
+	// already connected".
+	var success bool
+	defer func() {
+		if !success {
+			w.client.Shutdown()
```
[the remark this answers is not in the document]
itswisdomagain: Yeah, thought about it but wanted to maintain as much as possible what was in the codebase.

client/asset/dcr/interface.go
```diff
@@ -0,0 +1,89 @@
+package dcr
```
JoeGruffins, line 1: File needs the license info I guess.

client/asset/dcr/rpcwallet.go
```diff
@@ -0,0 +1,526 @@
...
+	GetTransaction(ctx context.Context, txHash *chainhash.Hash) (*walletjson.GetTransactionResult, error) // Should return asset.CoinNotFoundError if tx is not found.
+	WalletLock(ctx context.Context) error
+	WalletPassphrase(ctx context.Context, passphrase string, timeoutSecs int64) error
+	AccountUnlocked(ctx context.Context, account string) (*walletjson.AccountUnlockedResult, error)
+	LockAccount(ctx context.Context, account string) error
+	UnlockAccount(ctx context.Context, account, passphrase string) error
+	Disconnected() bool
+	RawRequest(ctx context.Context, method string, params []json.RawMessage) (json.RawMessage, error)
+	WalletInfo(ctx context.Context) (*walletjson.WalletInfoResult, error)
+	ValidateAddress(ctx context.Context, address stdaddr.Address) (*walletjson.ValidateAddressWalletResult, error)
```
JoeGruffins, line 67: There is another ValidateAddress in `dcr.go` still. Should it be removed?

client/asset/btc/btc.go
```diff
@@ -349,7 +349,7 @@ func (d *Driver) Info() *asset.WalletInfo {
 	return WalletInfo
 }
 
-func init() {
+func RegisterDriver() {
 	asset.Register(BipID, &Driver{})
 }
```
JoeGruffins, line 354: I'm not sure if because of this change, but the simnet tests stopped working.

client/cmd/dexc/main.go
```diff
@@ -28,6 +28,14 @@ import (
 	"decred.org/dcrdex/dex"
 )
 
+// Register asset drivers.
+func init() {
+	bch.RegisterDriver()
+	btc.RegisterDriver()
+	dcr.RegisterDriver()
+	ltc.RegisterDriver()
+}
```
JoeGruffins, line 37: Missing eth, which is registered on condition.

JoeGruffins (commented):
Manual testing looks like everything is working.

client/asset/dcr/rpcwallet.go
```diff
@@ -0,0 +1,526 @@
...
+	if err != nil {
+		return false, err
+	}
+	va, err := w.node.ValidateAddress(ctx, a)
+	if err != nil {
+		// Work around a bug with dcrwallet that prevents validateaddress for
+		// locked accounts.
+		if isAccountNotEncryptedErr(err) {
+			return false, nil
+		}
```
JoeGruffins, line 225: I can't seem to make this happen. Is this still an issue?

client/asset/dcr/rpcwallet.go
itswisdomagain: Ah, moved. Irrc, the `ValidateAddress` in dcr.go is the one to be used, it overrides the one here. But I'll take another look to be sure nothing has gone wrong there. Thanks for pointing this out.

client/cmd/dexc/main.go
itswisdomagain: Don't think eth is currently imported on master. Think it was removed sometime ago in preparation for the 0.3 release.

Having the asset drivers auto-registered whenever their respective packages are imported will cause issues for consumers that want to register a different driver for an asset, that's what motivated this change, making it the responsibility of consumers to call the `RegisterDriver` function as desired.

This PR opens the possibility for consumers to register a slightly different driver for dcr wallets.
For example , the default dcr driver that is registered above using `dcr.RegisterDriver()` has the following `Setup` instruction:
```
func (d *Driver) Setup(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network) (asset.Wallet, error) {
	return NewWallet(cfg, logger, network, &rpcWallet{})
}
```

With this PR, a different consumer might register a driver that does this instead:
```
func (d *Driver) Setup(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network) (asset.Wallet, error) {
	return NewWallet(cfg, logger, network, &spvWallet{})
}
```

If the default driver is auto-registered whenever the `client/asset/dcr` pkg is imported, consumers would be unable to register a different driver.

client/asset/btc/btc.go
itswisdomagain: Likely. What's the error you get? I'll run the simnet tests later today anyways, I expected there might be a need to make some changes to get them working.

client/asset/dcr/rpcwallet.go
itswisdomagain: @chappjc can shed more light, I guess.

chappjc: Was fixed in wallet with https://github.com/decred/dcrwallet/pull/2023.

Can remove this case I think.

client/cmd/dexc/main.go
JoeGruffins: I see! Thank you for the explanation. imo don't worry about eth, it's still in development.

client/asset/btc/btc.go
JoeGruffins: <details closed>
<summary>error</summary>

```
=== RUN   TestTradeSuccess
2021-10-05 09:18:09.784 [INF] TEST: === SETUP
2021-10-05 09:18:09.784 [INF] CORE:1[DB]: creating new version 5 database
2021-10-05 09:18:09.784 [DBG] CORE:1: Using locale printer for "en-US"
2021-10-05 09:18:09.784 [DBG] CORE:1: new client core created
2021-10-05 09:18:09.784 [INF] TEST: [client 1] Core created
2021-10-05 09:18:09.784 [INF] CORE:1: Starting DEX client core
2021-10-05 09:18:09.784 [INF] CORE:1: Successfully connected to 0 out of 0 DEX servers
2021-10-05 09:18:09.817 [INF] CORE:1: notify: |SUCCESS| (security) Don't forget to back up your application seed - A new application seed has been created. Make a back up now in the settings view.
2021-10-05 09:18:09.817 [INF] TEST: [client 1] Core initialized
    trade_simnet_test.go:200: error starting clients: asset with BIP ID 42 is unknown. Did you _ import your asset packages?
--- FAIL: TestTradeSuccess (0.06s)
```
</details>

chappjc:
OK I think I'm looking at this too rigidly, thinking that any possible `dcr.Wallet` interface implementation would need to be defined _in the client/asset/dcr package_.

But I think you're looking at an external package that implements `dcr.Wallet` and passes that to the `asset.Driver` methods as an `interface{}`, and when the dcr package gets it, it would ensure it satisfies `dcr.Wallet`, and if so, use that when doing `Open` (presumably if the `WalletConfig.Type` agrees with something from the provided Wallet)?

So in this scenario, would it not be better to leave `Create` alone and add a `UseWallet(interface{})` for this different approach?

I'm now aware that @buck54321 has thoughts about doing something similar, so I'll leave this to you guys to sort out.

chappjc:
> The assets that support wallet creation can then convert the params provided to `*CreateWalletParams` and do their thing, while other assets like dcr can convert the params to `dcr.Wallet` and do its thing.

This is why I think these are two separate functions.  One needs `CreateWalletParams` and one needs a `Wallet`/`WalletConstructor` (whichever is yet to be decided).  Both are followed by Open/Load (whichever is yet to be decided).  But just taking an `interface{}` is confusing.

The answer to constructor vs. wallet interface provide to RegisterCustomWallet/UseCustomWallet is whichever most effectively separates the caller from the concerns of the underlying custom wallet implementation.  With a constructor as the arg, the caller does not need to make the actual wallet, only inform the asset package what constructor is to be used to make a wallet for a particular `WalletDefinition`, then the they can use Open/Load once they've registered that constructor.  It keeps the setup workflow more standard.

PR #1230, opened 2021-10-11
client: bitcoin spv

buck54321:
Add a built-in Bitcoin SPV wallet. Uses neutrino and btcwallet.

#### client/asset:

Add `Create` and `Exists` methods to the `Driver`. `Create` is used to
create a "seeded" (native, built-in) wallet, and should
not be used for non-seeded wallets. `Exists` checks the existence of
a wallet. `Exists` is written so far to check even RPC wallet
existence, though it's only used for seeded wallets right now.

Adds the `WalletDefinition` type. `WalletInfo` has a new field,
`AvailableWallets []*WalletDefinition`.

`WalletConfig` type has a new field, `Type string`, which should match
the `Type` field of one of the `AvailableWallets`. Existing backends
must handle `Type == ""` for pre-existing wallets.

#### client/asset/btc:

Quite a few changes to configuration parsing to more clearly
separate RPC config from wallet settings. New `spvWallet` type
manages the `*btcwallet.Wallet` and `*neutrino.ChainService`. The
`Wallet` interface was already defined, so the work was just
implementing the methods using direct calls to wallet methods.

Everything is run through an interface for testing. Existing tests
required relatively few modifications to work with spv as well.
There are just a couple of new unit tests specific to the more
nuanced spv methods. The regnet_test.go/livetest.go test is adapted
to run an all-SPV trio for one test.

**client/core**: Work with new wallet definitions and give special handling
to seeded wallets.

#### front-end:

`NewWalletForm` and `WalletConfigForm` updated to work with
wallet definitions. Wallet type is selected via a tab at the top of
the `NewWalletForm`. Seeded wallets take no user-supplied password
(enforced in core). Special handling of headers and options so that
creating a seeded wallet can be very close to a 1-button, 1-click
experience.

A new form is added after the new wallet form conditionally if the
wallet balance is lower than required for the fee. This will always
be the case for the first native spv wallet. The new form shows a
deposit address and the current balance, and directs the user to
fund the address for registration fees. The balance is checked in
a loop and progresses back to the fee confirmation form when the
balance is sufficient. When working on simnet, just fund it from
the alpha node in harness-ctl.

#### client/db: 

Store the new `Type` field as part of the wallet. I wrote
an upgrade function to store an empty string for the new fields 
position in the database, but we're not populating it anyway, 
and existing wallets handle that, so I don't see a good reason 
to "upgrade".


client/asset/btc/spv.go
```diff
@@ -0,0 +1,1600 @@
...
+		if checkPt.blockHash != nil && checkPt.spend != nil {
+			// We already have the output and the spending input, and
+			// checkpointBlock already verified it's still mainchain.
+			return checkPt, nil
+		}
+		height, err := w.getBlockHeight(checkBlock)
+		if err != nil {
+			return nil, fmt.Errorf("getBlockHeight error: %w", err)
+		}
+		limitHeight = height + 1
```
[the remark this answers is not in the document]
buck54321: Yep.

client/asset/driver.go
```diff
@@ -15,13 +15,37 @@ var (
 	drivers    = make(map[uint32]Driver)
 )
 
+// CreateWalletParams are the parameters for internal wallet creation. The
+// Settings provided should be the same wallet configuration settings passed to
+// Create.
+type CreateWalletParams struct {
+	Type     string
+	Seed     []byte
+	Pass     []byte
```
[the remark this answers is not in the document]
buck54321: Hmm. Good point. I'm tempted to go the other way though, and change the `(Wallet).Unlock` method to accept a `[]byte`.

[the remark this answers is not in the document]
martonp: Won't that require a bunch of changes for non seeded wallets? The geth node takes a string to lock/unlock the private key as well.

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1601 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	txHash := tx.TxHash()
+
+	return &txHash, nil
+}
+
+// swapConfirmations attempts to get the numbe of confirmations and the spend
```
chappjc, line 635: numbe

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1601 @@
...
+		return nil, 0
+	}
+	if !w.blockIsMainchain(blockHash, blockHeight) {
+		return nil, 0
+	}
+	return blockHash, blockHeight
+}
+
+// findBlockForTime locates a good start block so that a search beginning at the
+// returned block has a very low likelyhood of missing any blocks that have time
```
chappjc, line 910: likelyhood -> likelihood

also, `MAX_FUTURE_BLOCK_TIME` -> `maxBlockTimeOffset`?

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1601 @@
...
+		}
+		return iTime.After(offsetTime)
+	})
+	if err != nil {
+		return nil, 0, fmt.Errorf("binary search error finding best block for time %q: %w", matchTime, err)
+	}
+
+	// We're actually breaking an assumption of sort.Search here because block
+	// times aren't always monotonically increasing. This won't matter though
+	// as long as there are not > blockTimeSearchBuffer blocks with inverted
```
chappjc, line 950: Mean to say medianTimeBlocks here instead of blockTimeSearchBuffer?

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1601 @@
...
+	// time order.
+	var count int
+	var iHash *chainhash.Hash
+	var iTime time.Time
+	for iHeight > 0 {
+		iHash, iTime, err = getBlockTimeForHeight(int32(iHeight))
```
chappjc, line 956: I think different binary searches are likely to call `getBlockTimeForHeight` for the same heights.  I might instrument these closures with a bunch of prints to get a bit more intuition on the process, but we may benefit from caching block hashes and timestamps just within `findBlockForTime` to we don't have repeated RPCs, not that they will be costly.

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1601 @@
...
+		var err error
+		blockHeight, err = w.cl.GetBlockHeight(blockHash)
+		if err != nil {
+			w.log.Errorf("Error getting block height for hash %s", blockHash)
+			return false
+		}
+	}
+	checkHash, err := w.cl.GetBlockHash(int64(blockHeight))
+	if err != nil {
+		w.log.Errorf("Error retriving block hash for height %d", blockHeight)
```
chappjc, line 884: retriving

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1601 @@
...
+		return false
+	}
+
+	return *checkHash == *blockHash
+}
+
+// mainchainBlockForStoredTx gets the block hash and height for the transaction
+// IFF an entry has been stored in the blockTxs index.
```
chappjc, line 892: Refers to the `blockTxs` variable that was renamed.

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1601 @@
...
+
+// checkpointBlock returns a filterScanResult and the checkpoint block hash. If
+// a result is found with an orphaned checkpoint block hash, it is cleared from
+// the cache and not returned.
+func (w *spvWallet) checkpointBlock(txHash *chainhash.Hash, vout uint32) (*filterScanResult, *chainhash.Hash) {
+	res := w.checkpoint(txHash, vout)
+	if res == nil {
+		return nil, nil
+	}
+	blockHash := &res.checkpoint
```
chappjc, line 343: It's not clear why you there is the second Hash output since the `filterScanResult` contains the hash.
There's only one caller of `checkpointBlock`, and it requires the whole `filterScanResult`.

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1601 @@
...
+		var err error
+		limitHeight, err = w.getBlockHeight(blockHash)
+		if err != nil {
+			return nil, fmt.Errorf("error getting height for supplied block hash %s", blockHash)
+		}
+	}
+
+	// We only care about the limitHeight now, but we can tell if it's been set
+	// by whether blockHash is nil, since that means it wasn't passed in, and
+	// it wasn't found in the database.
```
chappjc, line 1011: ...found in the database **or in a checkpoint**

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1601 @@
...
+
+// filterScanFromHeight scans BIP158 filters beginning at the specified block
+// height until the tip, or until a spending transaction is found.
+func (w *spvWallet) filterScanFromHeight(txHash chainhash.Hash, vout uint32, pkScript []byte, startBlockHeight int32, checkPt *filterScanResult) (*filterScanResult, error) {
+	tip, err := w.getBestBlockHeight()
+	if err != nil {
+		return nil, err
+	}
+
+	res := checkPt
```
chappjc, line 1103: I'm thinking we want to make a copy of the struct itself because we modify fields below, and this `*filterScanResult` is still in the checkpoints map and thus accessible by other goroutines.

Although I don't yet have a sense for if it is possible for simultaneous `scanFilters` calls for the same output.

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1601 @@
...
+	SpentStatusUnknown        = dex.ErrorKind("spend status not known")
+
+	// see btcd/blockchain/validate.go
+	maxBlockTimeOffset = 2 * time.Hour
+	neutrinoDBName     = "neutrino.db"
+)
+
+var (
+	SimnetSeed    = []byte("simnet-seed-of-considerable-length")
+	SimnetAddress = "bcrt1qq8t6vmptznycut4f3vxtguxgknmcnmpqlvd8wf"
```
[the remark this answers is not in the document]
buck54321: Leftovers from early dev. I'll get rid of em.

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1601 @@
...
+}
+
+// spvWallet is an in-process btcwallet.Wallet + neutrino light-filter-based
+// Bitcoin wallet. spvWallet controls an instance of btcwallet.Wallet directly
+// and does not run or connect to the RPC server.
+type spvWallet struct {
+	ctx         context.Context
+	chainParams *chaincfg.Params
+	wallet      btcWallet
+	cl          neutrinoService
```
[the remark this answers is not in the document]
buck54321: Code navigation challenge. Gotta have a strong search game.

client/asset/btc/spv_test.go
```diff
@@ -0,0 +1,607 @@
+//go:build !spvlive
+// +build !spvlive
```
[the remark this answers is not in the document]
buck54321: No. It was just some initial connection testing that is more than superceded by the regnet_test stuff now.

client/asset/btc/spv.go
buck54321: Oh. I was originally returning the nearest mainchain ancestor block hash, but then I realized that I would need to clear `filterScanResult` fields if there were orphaned blocks containing outputs or spends, and that looked like it was going to be messy, so I decided just to invalidate the entire checkpoint instead.

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1601 @@
...
+// 	BlockStamp() (*waddrmgr.BlockStamp, error)
+// 	SendRawTransaction(*wire.MsgTx, bool) (*chainhash.Hash, error)
+// 	Rescan(*chainhash.Hash, []btcutil.Address, map[wire.OutPoint]btcutil.Address) error
+// 	NotifyReceived([]btcutil.Address) error
+// 	NotifyBlocks() error
+// 	Notifications() <-chan interface{}
+// 	BackEnd() string
+// }
+
+// func (w *Wallet) MakeMultiSigScript(addrs []btcutil.Address, nRequired int) ([]byte, error)
```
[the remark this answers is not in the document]
buck54321: I think I'll get rid of it. It is nice to have, but probably more of an early dev tool.

client/asset/btc/btc.go
```diff
@@ -326,11 +353,106 @@ func (r *swapReceipt) SignedRefund() dex.Bytes {
...
 type Driver struct{}
 
-// Setup creates the BTC exchange wallet. Start the wallet with its Run method.
-func (d *Driver) Setup(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network) (asset.Wallet, error) {
+// Exists checks the existence of the wallet. For the RPC wallet, this attempts
+// to connect and request getnetworkinfo to verify existence. For the SPV wallet,
+// we ultimately just look for wallet files.
+func (d *Driver) Exists(walletType, dataDir string, settings map[string]string, net dex.Network) (bool, error) {
+	switch walletType {
+	case "", WalletTypeRPC:
```
chappjc, line 410: `caes WalletTypeLegacy, WalletTypeRPC:`

client/asset/btc/btc.go
```diff
@@ -326,11 +353,106 @@ func (r *swapReceipt) SignedRefund() dex.Bytes {
...
+		netDir := filepath.Join(dataDir, chainParams.Name)
+		// timeout and recoverWindow arguments borrowed from btcwallet directly.
+		loader := wallet.NewLoader(chainParams, netDir, true, 60*time.Second, 250)
+		return loader.WalletExists()
+	}
+
+	return false, fmt.Errorf("no Bitcoin wallet of type %q available", walletType)
+}
+
+// Create cretes a new SPV wallet.
```
chappjc, line 434: cretes

chappjc:
Different race, but also related to logging:

<details>
```
==================
WARNING: DATA RACE
Read at 0x00c000633688 by goroutine 53:
  github.com/jrick/logrotate/rotator.(*Rotator).Write()
      /home/jon/go/pkg/mod/github.com/jrick/logrotate@v1.0.0/rotator/rotator.go:136 +0xba
  decred.org/dcrdex/client/asset/btc.logWriter.Write()
      /home/jon/github/decred/dcrdex/client/asset/btc/spv.go:198 +0x4a
  github.com/btcsuite/btclog.(*Backend).printf()
      /home/jon/go/pkg/mod/github.com/btcsuite/btclog@v0.0.0-20170628155309-84c8d2346e9f/log.go:309 +0x3ee
  github.com/btcsuite/btclog.(*slog).Infof()
      /home/jon/go/pkg/mod/github.com/btcsuite/btclog@v0.0.0-20170628155309-84c8d2346e9f/log.go:391 +0xd7
  github.com/lightninglabs/neutrino.(*blockManager).Stop()
      /home/jon/go/pkg/mod/github.com/lightninglabs/neutrino@v0.12.1/blockmanager.go:346 +0x133
  github.com/lightninglabs/neutrino.(*ChainService).Stop()
      /home/jon/go/pkg/mod/github.com/lightninglabs/neutrino@v0.12.1/neutrino.go:1571 +0x218
  decred.org/dcrdex/client/asset/btc.(*spvWallet).stop()
      /home/jon/github/decred/dcrdex/client/asset/btc/spv.go:902 +0x1cc
  decred.org/dcrdex/client/asset/btc.(*spvWallet).connect.func1()
      /home/jon/github/decred/dcrdex/client/asset/btc/spv.go:804 +0x264

Previous write at 0x00c000633688 by goroutine 43:
  github.com/jrick/logrotate/rotator.New()
      /home/jon/go/pkg/mod/github.com/jrick/logrotate@v1.0.0/rotator/rotator.go:73 +0xa4
  decred.org/dcrdex/client/asset/btc.(*spvWallet).connect()
      /home/jon/github/decred/dcrdex/client/asset/btc/spv.go:757 +0xfd
  decred.org/dcrdex/client/asset/btc.(*ExchangeWallet).Connect()
      /home/jon/github/decred/dcrdex/client/asset/btc/btc.go:748 +0x93
  decred.org/dcrdex/dex.(*ConnectionMaster).Connect()
      /home/jon/github/decred/dcrdex/dex/runner.go:106 +0x9b
  decred.org/dcrdex/client/asset/btc/livetest.tBackend()
      /home/jon/github/decred/dcrdex/client/asset/btc/livetest/livetest.go:69 +0x8b9
  decred.org/dcrdex/client/asset/btc/livetest.Run()
      /home/jon/github/decred/dcrdex/client/asset/btc/livetest/livetest.go:159 +0x3c4
  decred.org/dcrdex/client/asset/btc/livetest.TestWallet()
      /home/jon/github/decred/dcrdex/client/asset/btc/livetest/regnet_test.go:182 +0x5b0
  testing.tRunner()
      /home/jon/go117/src/testing/testing.go:1259 +0x22f
  testing.(*T).Run·dwrap·21()
      /home/jon/go117/src/testing/testing.go:1306 +0x47

Goroutine 53 (running) created at:
  decred.org/dcrdex/client/asset/btc.(*spvWallet).connect()
      /home/jon/github/decred/dcrdex/client/asset/btc/spv.go:778 +0x964
  decred.org/dcrdex/client/asset/btc.(*ExchangeWallet).Connect()
      /home/jon/github/decred/dcrdex/client/asset/btc/btc.go:748 +0x93
  decred.org/dcrdex/dex.(*ConnectionMaster).Connect()
      /home/jon/github/decred/dcrdex/dex/runner.go:106 +0x9b
  decred.org/dcrdex/client/asset/btc/livetest.TestWallet.func1()
      /home/jon/github/decred/dcrdex/client/asset/btc/livetest/regnet_test.go:117 +0x4e8
  decred.org/dcrdex/client/asset/btc/livetest.TestWallet.func2()
      /home/jon/github/decred/dcrdex/client/asset/btc/livetest/regnet_test.go:168 +0x371
  decred.org/dcrdex/client/asset/btc/livetest.tBackend()
      /home/jon/github/decred/dcrdex/client/asset/btc/livetest/livetest.go:63 +0x724
  decred.org/dcrdex/client/asset/btc/livetest.Run()
      /home/jon/github/decred/dcrdex/client/asset/btc/livetest/livetest.go:159 +0x3c4
  decred.org/dcrdex/client/asset/btc/livetest.TestWallet()
      /home/jon/github/decred/dcrdex/client/asset/btc/livetest/regnet_test.go:182 +0x5b0
  testing.tRunner()
      /home/jon/go117/src/testing/testing.go:1259 +0x22f
  testing.(*T).Run·dwrap·21()
      /home/jon/go117/src/testing/testing.go:1306 +0x47

Goroutine 43 (running) created at:
  testing.(*T).Run()
      /home/jon/go117/src/testing/testing.go:1306 +0x726
  testing.runTests.func1()
      /home/jon/go117/src/testing/testing.go:1598 +0x99
  testing.tRunner()
      /home/jon/go117/src/testing/testing.go:1259 +0x22f
  testing.runTests()
      /home/jon/go117/src/testing/testing.go:1596 +0x7ca
  testing.(*M).Run()
      /home/jon/go117/src/testing/testing.go:1504 +0x9d1
  main.main()
      _testmain.go:43 +0x22b
```
</details>

dex/networks/btc/config.go
```diff
@@ -32,56 +31,83 @@ const (
...
+// 	return checkConfig(cfg, name, network, ports)
+// }
+
+// // LoadConfigFromSettings loads the configuration settings from a settings map.
+// func LoadConfigFromSettings(settings map[string]string, name string, network dex.Network, ports NetPorts, cfg interface{}) (error) {
+// 	if err := config.Unmapify(settings, cfg); err != nil {
+// 		return nil, fmt.Errorf("error parsing connection settings: %w", err)
+// 	}
+// 	return checkConfig(cfg, name, network, ports)
+// }
```
chappjc, line 103: These are commented and duplicated above.  Keep one or get rid of both?

go.mod
```diff
@@ -34,6 +41,7 @@ require (
 	github.com/jessevdk/go-flags v1.4.1-0.20200711081900-c17162fe8fd7
 	github.com/jrick/logrotate v1.0.0
 	github.com/lib/pq v1.2.0
+	github.com/lightninglabs/neutrino v0.12.1
```
chappjc, line 44: Patch release v0.12.3 tagged a couple weeks ago. https://github.com/lightninglabs/neutrino/compare/v0.12.1...v0.12.3
Just one fix to rebroadcasting code it appears.

chappjc:
> Here's a weird one that's not related to logging at all, all inside neutrino and btcd code..

This does look like a API design issue that makes the `btcd/addrmgr.(*AddrManager).GetAddress` method essentially unusable by consumers.  Considering a PR: https://github.com/btcsuite/btcd/compare/master...chappjc:safe-addr-accessors?expand=1

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1631 @@
...
+							break search
+						}
+						// The output could still be in this block, just not
+						// in this transaction.
+						continue nextTx
+					}
+				}
+			}
+			// Only check for the output if this is the right transaction.
+			if tx.TxHash() != txHash {
```
JoeGruffins, line 1211: Not sure but maybe `if res.blockHash != nil ||  tx.TxHash() != txHash {` is faster because you can skip `tx.TxHash()` sometimes. Maybe doesn't matter.

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1631 @@
...
+		for outPt, res := range newlyDiscovered {
+			discovered[outPt] = res
+		}
+	}
+	return
+}
+
+// findRedemptionsInMempool is unsupported for SPV.
+func (w *spvWallet) findRedemptionsInMempool(ctx context.Context, reqs map[outPoint]*findRedemptionReq) (discovered map[outPoint]*findRedemptionResult) {
+	return
```
JoeGruffins, line 1296: Will return nil rather than an empty map (I think) if that`s fine.

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1631 @@
...
+	}
+
+	ret := &GetTransactionResult{
+		TxID:         txHash.String(),
+		Hex:          txBuf.Bytes(), // 'Hex' field name is a lie, kinda
+		Time:         uint64(details.Received.Unix()),
+		TimeReceived: uint64(details.Received.Unix()),
+	}
+
+	if details.Block.Height != -1 {
```
JoeGruffins, line 1368: For sure `details.Block` is never nil? Also `details.MsgTx` above. Assuming they are pointers.

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1631 @@
...
+
+	// Option #2
+	// This is what the JSON-RPC does (and has since at least May 2018).
+	details, err := w.wallet.walletTransaction(txHash)
+	if err != nil {
+		return nil, err
+	}
+	if details == nil {
+		return nil, WalletTransactionNotFound
+	}
```
JoeGruffins, line 1348: It looks like `w.wallet.walletTransaction` is already checking for this, but fine to check again, as wallet is an interface I guess.

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1631 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("BestBlock error: %v", err)
+	}
+
+	blockHeight, err := w.cl.GetBlockHeight(blockHash)
+	if err != nil {
+		return nil, err
+	}
+
+	confs := uint32(tip.Height - blockHeight)
```
JoeGruffins, line 709: Does this need a +1? you do have a `confrims` helper method as well.

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1631 @@
...
+	txDetails, err := w.wallet.walletTransaction(txHash)
+	if err == nil {
+		tip, err := w.cl.BestBlock()
+		if err != nil {
+			return nil, 0, fmt.Errorf("BestBlock error: %v", err)
+		}
+
+		var confs uint32
+		if txDetails.Block.Height > 0 {
+			confs = uint32(tip.Height - txDetails.Block.Height + 1)
```
JoeGruffins, line 1107: Could use the `confirms` helper.

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1631 @@
...
+	if utxo.blockHash == nil {
+		return nil, 0, fmt.Errorf("output %s:%v not found", txHash, vout)
+	}
+
+	tip, err := w.cl.BestBlock()
+	if err != nil {
+		return nil, 0, fmt.Errorf("BestBlock error: %v", err)
+	}
+
+	confs := uint32(tip.Height) - utxo.blockHeight + 1
```
JoeGruffins, line 1143: Could use the `confirms` helper function.

client/asset/driver.go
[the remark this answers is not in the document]
JoeGruffins: I thought we had a `dex.Bytes` type for passwords.

client/asset/dcr/dcr.go
```diff
@@ -377,8 +381,47 @@ type fundingCoin struct {
...
+		return false, fmt.Errorf("no Decred wallet type %q available", walletType)
+	}
+	cfg, _, err := loadConfig(settings, net)
+	if err != nil {
+		return false, err
+	}
+	cl, err := rpcclient.New(&rpcclient.ConnConfig{
+		HTTPPostMode: true,
+		DisableTLS:   true,
+		Host:         cfg.RPCListen + "/wallet/" + cfg.Account,
```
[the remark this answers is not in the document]
buck54321: You're right about the host. It's not actually used anywhere, since we only check existence of seeded wallets. I should probably just define the method that way and get rid of this.

client/asset/dcr/dcr.go
```diff
@@ -534,7 +577,7 @@ func unconnectedWallet(cfg *asset.WalletConfig, dcrCfg *Config, chainParams *cha
 	return &ExchangeWallet{
 		log:                 logger,
 		chainParams:         chainParams,
-		acct:                cfg.Settings["account"],
```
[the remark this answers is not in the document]
buck54321: Just seems like it should have been there the whole time. I was trying to figure out why it wouldn't be. I think at one time I thought account would be a `WalletConfig` field, but I don't know if that's relevant. Anyway, sorry for the random change in an already large PR.

client/core/core.go
```diff
@@ -1936,24 +1998,50 @@ func (c *Core) ChangeAppPass(appPW, newAppPW []byte) error {
...
+		if err != nil {
+			return newError(existenceCheckErr, "error checking wallet pre-existence: %v", err)
+		}
+
+		if !exists {
+			newWalletPW, err = c.createSeededWallet(assetID, crypter, form)
+			if err != nil {
+				return newError(createWalletErr, "error creating new %q-type %s wallet: %v", form.Type, unbip(assetID), err)
+			}
+			// return newError(existenceCheckErr, "cannot reconfigure wallet that doesn't exist")
```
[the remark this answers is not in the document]
chappjc: Ah I hadn't thought about switching between wallet types.  That's a good point.

client/asset/dcr/dcr.go
```diff
@@ -1676,8 +1735,11 @@ func (dcr *ExchangeWallet) AuditContract(coinID, contract, txData dex.Bytes, _ t
 		return nil, fmt.Errorf("contract hash doesn't match script address. %x != %x",
 			contractHash, addrScript)
 	}
+	if coinNotFound {
+		return nil, asset.CoinNotFoundError
+	}
```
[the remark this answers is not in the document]
chappjc: > and be more permissive of `CoinNotFound` error there

I guess this is the twist.

> I don't personally see a lot of value in double checking that the on-chain contract validates the same as the txData

Felt prudent to me too, but I'm not totally set on that.

client/core/core.go
```diff
@@ -1936,24 +1998,50 @@ func (c *Core) ChangeAppPass(appPW, newAppPW []byte) error {
...
 		EncryptedPW: oldWallet.encPW(),
 		Address:     oldDepositAddr,
 	}
+
+	if walletDef.Seeded {
+		if len(newWalletPW) > 0 {
+			return newError(passwordErr, "cannot set a password on a seeded wallet")
+		}
+
+		exists, err := asset.WalletExists(assetID, form.Type, c.assetDataDirectory(assetID), form.Config, c.net)
```
itswisdomagain, line 2031: The btc driver implementation of this method (haven't checked others) expects that this method may be called for RPC wallets as well. Is that the case or is that planned?

itswisdomagain (changes requested):
Haven't really gone through all the changes yet, was just going through the changes to the asset.Driver interface and have a question. Will do a more thorough review later.

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1631 @@
...
+
+// Lock locks the wallet.
+func (w *spvWallet) Lock() error {
+	w.wallet.Lock()
+	return nil
+}
+
+// sendToAddress sends the amount to the address. feeRate is in units of
+// sats/byte.
+func (w *spvWallet) sendToAddress(address string, value, feeRate uint64, subtract bool) (*chainhash.Hash, error) {
```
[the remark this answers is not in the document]
buck54321: Woah. Totally missed that. Derp.

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1631 @@
...
+
+	tip, err := w.cl.BestBlock()
+	if err != nil {
+		return nil, fmt.Errorf("BestBlock error: %v", err)
+	}
+
+	blockHeight, err := w.cl.GetBlockHeight(blockHash)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
buck54321: Not in a `*wire.BlockHeader`.

client/asset/btc/spv.go
JoeGruffins: Not pointers I see. Ty.

client/asset/btc/btc.go
```diff
@@ -429,78 +572,100 @@ type findRedemptionResult struct {
...
+	case dex.Regtest:
+		return &chaincfg.RegressionNetParams, nil
+	}
+	return nil, fmt.Errorf("unknown network ID %v", net)
+}
+
 // NewWallet is the exported constructor by which the DEX will import the
 // exchange wallet. The wallet will shut down when the provided context is
 // canceled. The configPath can be an empty string, in which case the standard
 // system location of the bitcoind config file is assumed.
```
itswisdomagain, line 590:
```suggestion
// NewWallet is the exported constructor by which the DEX will import the
// exchange wallet. The wallet will shut down when the provided context is
// canceled.
```

client/asset/btc/btc.go
```diff
@@ -429,78 +572,100 @@ type findRedemptionResult struct {
...
 		DefaultFallbackFee:  defaultFee,
 		DefaultFeeRateLimit: defaultFeeRateLimit,
 		Segwit:              true,
 	}
 
-	return BTCCloneWallet(cloneCFG)
+	switch cfg.Type {
+	case WalletTypeSPV:
+		if atomic.LoadUint32(&initialized) == 0 {
+			return nil, fmt.Errorf("must call asset.Initialize if you want to create an SPV wallet")
```
itswisdomagain, line 614: Haven't encountered any doc highlighting this requirement. Since this is called from `asset.OpenWallet`, that would be a good place to indicate this requirement. Also, as part of this function's doc and the Initializer interface methods and implementations.

Is it possible to call `Initialize` here if it hasn't been called before?

Re the error message, this isn't an attempt to create an SPV wallet, more like load to me or as per the call stack leading here, open.

client/asset/btc/btc.go
```diff
@@ -429,78 +572,100 @@ type findRedemptionResult struct {
...
+	if err != nil {
+		return nil, fmt.Errorf("error creating %s ExchangeWallet: %v", cfg.Symbol,
+			err)
+	}
+
+	return btc, nil
+}
 
-	client, err := rpcclient.New(&rpcclient.ConnConfig{
+func newRPCWalletConnection(cfg *RPCWalletConfig) (*rpcclient.Client, error) {
```
itswisdomagain, line 643: This function looks simple and is only called from `ParseRPCWalletConfig`, can probably inline it there instead.

client/asset/btc/btc.go
```diff
@@ -567,6 +728,29 @@ func newWallet(requester RawRequesterWithContext, cfg *BTCCloneCFG, btcCfg *dexb
 	return w, nil
 }
 
+// openSPVWallet opens the pre-existing native SPV wallet.
```
itswisdomagain, line 731: pre-existing -> previously created?

client/asset/driver.go
```diff
@@ -4,24 +4,64 @@
...
 )
 
 var (
 	driversMtx sync.RWMutex
 	drivers    = make(map[uint32]Driver)
 )
 
+// CreateWalletParams are the parameters for internal wallet creation. The
+// Settings provided should be the same wallet configuration settings passed to
+// Create.
```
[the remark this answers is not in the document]
chappjc: I'm assuming this is saying it's the same settings map in WalletConfig, passed to OpenWallet, as you say.

client/asset/btc/btc.go
```diff
@@ -326,11 +355,100 @@ func (r *swapReceipt) SignedRefund() dex.Bytes {
...
-func (d *Driver) Setup(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network) (asset.Wallet, error) {
+// Check that Driver implements Driver, Creator, and Initializer.s
+var _ asset.Driver = (*Driver)(nil)
+var _ asset.Creator = (*Driver)(nil)
+var _ asset.Initializer = (*Driver)(nil)
+
+// Exists checks the existence of the wallet. For the RPC wallet, this attempts
+// to connect and request getnetworkinfo to verify existence. For the SPV wallet,
+// we ultimately just look for wallet files.
+func (d *Driver) Exists(walletType, dataDir string, settings map[string]string, net dex.Network) (bool, error) {
```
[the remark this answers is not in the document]
chappjc: Agree, this was suggested by buck too after I pointed out it's not called for RPC wallets anyway. Should change

client/asset/btc/btc.go
chappjc: Having reviewed the latest commits, some thoughts:  Please push back if you have solid reasons to do it now, @buck54321, but actually I kindly request de-scoping `Initialize` for this PR unless it's a ton more work.  Considerations (please retort if I'm seeing things wrong):

* the current function of `Initialize`, cleaning up the `logRotator` used by neutrino, does not even apply except with `WalletTypeSPV`, making it's closing more of a responsibility for a goroutine started by probably `logNeutrino` or `loadSPVWallet`/`createSPVWallet`
* the `lang` arg isn't used yet
* the purpose of the `logger` arg is unclear.  In fact it won't be used unless closing the rotator fails, so it's presence is somewhat dubious.
* it's necessary (under certain conditions and what are they?) to do asset.Initialize in the caller (Core).  In general more docs are needed to explain how the Driver is supposed to be used correctly.

ref: `(d *Driver) Initialize(ctx context.Context, wg *sync.WaitGroup, logger dex.Logger, lang language.Tag)`

client/asset/driver.go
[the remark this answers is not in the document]
buck54321: Yes. Same typo here as there. Sorry about that.

client/asset/btc/spv.go
```diff
@@ -889,9 +917,12 @@ func (w *spvWallet) connect(ctx context.Context, wg *sync.WaitGroup) error {
 	// Nanny for the spendingTxs cache. We'll keep the cache entries for 2 hours
 	// past their last access time.
 	wg.Add(1)
-	teardownWaitGroup.Add(1)
 	go func() {
-		defer teardownWaitGroup.Done()
+		if err := logNeutrino(w.netDir); err != nil {
+			w.log.Warnf("error initializing btcwallet+neutrino logging: %v", err)
+		} else {
+			defer unlogNeutrino()
```
chappjc, line 924: At the top of the defer stack, `wg.Done` is going to happen before this.  Think we should put `defer wg.Done` as the first line in this closure.

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1712 @@
...
+		return nil, err
+	}
+
+	txHash := tx.TxHash()
+
+	return &txHash, nil
+}
+
+func (w *spvWallet) sendWithSubtract(pkScript []byte, value, feeRate uint64) (*chainhash.Hash, error) {
+	const unfundedTxSize = dexbtc.MinimumTxOverhead + 2*dexbtc.P2WPKHOutputSize
```
chappjc, line 680: The assumption here is that the pkScript is a P2WPKH.  So I'd suggest doing something like:

```go
const unfundedTxSize = dexbtc.MinimumTxOverhead + dexbtc.P2WPKHOutputSize + // change
	TxOutOverhead + len(pkScript) // send-to address
```

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1712 @@
...
+		return nil, fmt.Errorf("error listing unspent outputs: %w", err)
+	}
+
+	utxos, _, _, err := convertUnspent(0, unspents, w.chainParams)
+	if err != nil {
+		return nil, fmt.Errorf("error converting unspent outputs: %w", err)
+	}
+
+	enough := func(inputsSize, inputsVal uint64) bool {
+		return inputsVal >= value
```
chappjc, line 693: Is `inputSize` not needed because this is a subtract-from-amount send (withdraw), where the output amount per se is not the target (fees may eat all that's needed as long as the right amount is consumed by the inputs)?  If that reasoning is correct, let's comment and/or `_` out the first arg to `enough`.

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1712 @@
...
+// logNeutrino initializes logging in the neutrino + wallet packages. Logging
+// only has to be initialized once, and the rotating file logger must be Close'd
+// at some point, so a package level counter is used to synchronize feeders.
+// The unlogNeutrino function should be invoked by callers of logNeutrino when
+// they no longer need logging.
+func logNeutrino(netDir string) error {
+	rotatorMtx.Lock()
+	defer rotatorMtx.Unlock()
+	if loggerCount > 0 {
+		return nil
```
chappjc, line 196: I think you still want to increment here.

```go
if loggerCount > 0 {
	loggerCount++
	return nil
}
```

seems like the only place not to increment is with a non-nil return setting up the logger.

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1712 @@
...
+	// Actually, should use btcwallet.Wallet.NtfnServer ?
+
+	// Nanny for the caches checkpoints and txBlocks caches.
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		defer unlogNeutrino()
+		defer w.stop()
+
+		ticker := time.NewTicker(time.Minute * 20)
```
chappjc, line 928: `defer ticket.Stop()`

chappjc:
With my testnet client setup, I went to the Bitcoin wallet settings to switch to SPV.  It displayed the settings for SPV wallets however:

image

I was able to click "+ change wallet type" and select SPV from the dropdown and submit, which created the built-in SPV wallet, then started syncing it.  But it looks like it may not be considering the existing wallet type correctly when showing the current settings.

client/asset/btc/spv.go
```diff
@@ -0,0 +1,1712 @@
...
+	if err != nil {
+		return nil, 0, err
+	}
+
+	if utxo == nil || utxo.spend != nil {
+		return nil, 0, nil
+	}
+
+	if utxo.blockHash == nil {
+		return nil, 0, fmt.Errorf("output %s:%v not found", txHash, vout)
```
chappjc, line 1298: Instead of the usual `[TRC] CORE: Still searching for counterparty's contract coin ...`,
I'm getting `[ERR] CORE: Unexpected audit contract f89e23bc0...:0 (btc) error (will try again): error finding unspent contract: f89e23bc0...:0 : output f89e23bc0...:0 not found` for a couple dozen retries and then a hard fail:
`[ERR] CORE: Failed to audit contract coin %v (%s) for match %s: %v f89e23bc0...:0 btc ee22a089b7edd7280e885f14ff7688207cb3058fd49b5b0db331250009154a3b error finding unspent contract: f89e23bc0...:0 : output f89e23bc0....:0 not found`

I think this needs to return a `CoinNotFoundError` otherwise bad things happen in the caller.

client/asset/btc/spv.go
chappjc: I confirmed that the remaining log races go away with this, but then realized that in `createSPVWallet` (the other place where `logNeutrino` is called), we need a `defer unlogNeutrino()` too otherwise it never decrements to zero to close the log rotator.

Sadly, this reveals a problem: https://github.com/lightninglabs/neutrino/blob/51686db787e01a3709b1583af3e20e916811d585/neutrino.go#L1136-L1137

The time drain on this logger is unfortunate, and it's looking like only ever calling UseLogger _once_ (somehow), even if we are synchronizing the best we can, is the only actual solution.  I'm not even sure a dumb Sleep would do it, but perhaps it's worth doing that just to save ourselves further wasted time.

client/asset/btc/spv.go
buck54321: > Also, I'm surprised that since I'm trading with myself that walletTransaction didn't find the "counterparty" contract.... another proper testing challenge really here since a different btc wallet would be appropriate for a realistic test. Still, it must have fallen back to scanFilters, but I don't get why, even if that would be how trading with someone else would go.

We know the transaction, but we're not tracking the output's spend status, so we still have to scan.

> It also seemed to fail to find the tx with a cfilters scan in the block that contained the tx, and then start scanning the next (wrong) height.

Are you sure you're not seeing calls to `(*ExchangeWallet).SwapConfirmations` → `(*spvWallet).swapConfirmations` → `(*spvWallet).scanFilters`? Based on your logs, the swap was mined in 838 and Core found the swap on the first scan. `SwapConfirmations` would be called during `tick`s with the scan starting at 839`, which isn't mined yet, so the scan is a no-op because `limitHeight` > `tip`. And since there was a 2 conf requirement for btc, the swap couldn't proceed until 839 was actually mined. 

I'm not sure about the `It was a dcr tip change that triggered it`. That's the only part that doesn't make sense. 

> I think this needs to return a CoinNotFoundError

Oh. If the output isn't located, `getTxOut` should actually return a `nil` error, which has the same effect as `CoinNotFoundError` in `AuditContract`.

buck54321: Oh, is it only a 1-conf requirement for testnet btc?

buck54321: Ah. I see what was happening now. You're absolutely right. The `CoinNotFoundError` issue prevented us from advancing the match status because we hit the 24 `!errors.Is(err, asset.CoinNotFoundError)` limit, ending our `AuditContract` search early.

client/core/core.go
```diff
@@ -2056,8 +2077,27 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, form *WalletForm) er
...
+				return newError(authErr, "error retrieving wallet password: %v", err)
+			}
+		}
+
+		if oldWallet.connected() {
+			oldDef, err := walletDefinition(assetID, oldWallet.walletType)
+			// Error can be normal if the wallet was created before wallet types
+			// were a thing. Just assume this is an old wallet and therefore not
+			// seeded.
+			if err == nil && oldDef.Seeded {
```
chappjc, line 2092: You already have `oldDef` from 30 lines up before `if walletDef.Seeded`.

chappjc: Should be good with:

```go
if oldDef.Seeded && oldWallet.connected() {
	oldWallet.Disconnect()
	restartOnFail = true
}
```

chappjc: NVM, I have some updates that will follow.  Going to merge this shortly.

PR #1232, opened 2021-10-15
multi: market overview and price feed

### LINKED ISSUE #941: api: need a spot price subscription
The `Market` generates a `msgjson.Spot` at the end of every epoch, but just discards it right now. For the client to maintain a candlestick cache, those spots need to be sent to order book subscribers. But even if not subscribed to the order book, the ability to subscribe to just the spots for all markets enables some nice UI components. 

For example, here's what you see on the homepages of popular CEXs.

 

and here's what you'll see in the market selection panel in their markets views

  

They all provide general info about a market without the need to actually load the market's page. This only comes with the ability to get a general overview of all markets.

Additionally, I think we need a subscription for the latest fee rate used by the server during matching. This would help mitigate the issue of clients getting bad estimates because their nodes aren't warmed up. Those with active websocket connections to the server should just use the server's latest fee rate for order estimation. This also helps resolve an outstanding issue with calculating fees for zero-conf funding inputs as would be common for clients using split transaction for pre-sized funding.

This fee rate subscription doesn't necessarily need to be the same subscription as the spots, but I will note that the timing is very convenient, since the spot is sent at the end of the match cycle, exactly when a fee rate is calculated. 

Issue is not unrelated to #142, but definitely a separate issue.

buck54321:
Resolves #941

Add new `price_feed` subscription. The response to the subscription request contains the spot prices for all known markets. Subscribers will then receive an update to the spot price for every market at the end of every epoch.

**server/market**

`BookRouter` handles the subscriptions. The spot price is passed by `Market` as part of the `sigDataEpochReport`. It was already being generated by `(*DataAPI).ReportEpoch`, but we were discarding it. 

**dex/candles**

`(*Cache).Delta` method updated to ignore leading and trailing zeros. Zeros occur when the book is empty.

**client/core** 

Subscribe to the price feed right after fetching configuration, including after reconnect. Populate the new `Market.Spot` field. Send updates as notifications.

**front-end**

Display 24-hour percent change and current rate in the list on the left side of the markets view.

### DIFF
--- a/client/webserver/site/src/css/market.scss
+++ b/client/webserver/site/src/css/market.scss
@@ -36,7 +36,7 @@ div.marketrow {
   flex-direction: column;
   align-items: stretch;
   margin: 3px;
-  background-color: #34383e;
+  background-color: #c2c9d4;
   border-radius: 3px;
   // border: 1px solid #7775;
 
@@ -51,7 +51,7 @@ div.marketrow {
   &:hover,
   &.selected {
     padding: 2px 3px;
-    border: 1px solid #7a97;
+    border: 1px solid #1b966d;
   }
 
   span.upgreen {
--- a/client/webserver/site/src/css/market_dark.scss
+++ b/client/webserver/site/src/css/market_dark.scss
@@ -37,6 +37,14 @@ body.dark {
     }
   }
 
+  .marketrow {
+    background-color: #34383e;
+
+    &.selected {
+      border: 1px solid #7a97;
+    }
+  }
+
   label.market-search {
     background-color: black;
   }


client/core/bookie.go
```diff
@@ -876,6 +876,46 @@ func (dc *dexConnection) refreshServerConfig() error {
...
+		var msgErr *msgjson.Error
+		// Ignore old servers' errors.
+		if !errors.As(err, &msgErr) || msgErr.Code != msgjson.UnknownMessageType {
+			dc.log.Errorf("unable to fetch market overview: %w", err)
+		}
+		return
+	}
+
+	dc.spotsMtx.Lock()
+	dc.spots = spots
```
chappjc, line 894: Thinking we should not assign if `spots` is still `nil` after unmarshalling the response.  Server's response could be valid and result in `spots` staying nil if response payload is json `null` I think.
Would want to avoid panic assigning to nil map in `handlePriceUpdateNote`.

chappjc (changes requested):
Looks good and tests well.

Will leave it up for a a day for others.

client/webserver/site/src/js/doc.js
```diff
@@ -169,6 +169,16 @@ export default class Doc {
     return ancestor.querySelector(`[data-tmpl="${s}"]`)
   }
 
+  /*
+  * parseTemplate returns an object of data-tmpl elements, keyed by their
+  * data-tmpl values.
+  */
+  static parseTemplate (ancestor) {
+    const d = {}
+    for (const el of ancestor.querySelectorAll('[data-tmpl')) d[el.dataset.tmpl] = el
```
buck54321, line 178: Whoops. Missing a closing bracket. Weird that it still worked.

buck54321:
If anyone else looking at this on the simnet harness, I'll note that testing is a pain because we interpret our 24 hour change from the 5-minute candles, so even if you put some action on the books immediately, it can take 5 minutes before you see a non-zero change.

PR #1234, opened 2021-10-18
app: improved registration sequence and form design

buck54321:
Add an asset selection form to the registration sequence. Re-style forms and add new animations. Enable login through registration page so that the `pwCache` can be leveraged. Form styling includes new light-weight font style.


buck54321:
> Showing markets involving each fee asset makes sense I think, but in that screenshot, BCH fee asset lists a LTC-DCR market. Is that a live test glitch or a bug?

It's almost certainly a mis-assigned asset ID in **client/webserver/live_test.go**. Thanks for pointing that out.

chappjc:
Getting this on release-v0.3 might be tough,, but I understand that's your intent.

Ways to deal with that:
- completely catch up release-v0.3 with master, which includes quite a lot more than we wanted, including: the UnitInfo business, asset interface changes, price feed.  Depends on 1.7 and thus 0.3 timeline.
- catch up release-v0.3 with just the frontend changes, no UnitInfo/denominations or SPV work

Here's where we are: 
image

buck54321:
> Where would markets be listed that do not involve a fee asset on either side?

I've added a listing of all markets.

> Similarly, would anything break if there is a fee asset that is not in any market pair?

Nothing would break. The table would just be empty.

PR #1236, opened 2021-10-19
app: import app via registry getter

buck54321:
Instead of sneaking the `Application` into modules via class constructors, register the application with a "registry" that exports a getter.

### DIFF
--- a/client/webserver/live_test.go
+++ b/client/webserver/live_test.go
@@ -295,10 +295,17 @@ var tExchanges = map[string]*core.Exchange{
 			mkid(3, 22): mkMrkt("doge", "mona"),
 		},
 		Connected: true,
-		Fee: &core.FeeAsset{
-			ID:    42,
-			Confs: 1,
-			Amt:   1e8,
+		RegFees: map[string]*core.FeeAsset{
+			"dcr": &core.FeeAsset{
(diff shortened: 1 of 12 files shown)


client/webserver/live_test.go
```diff
@@ -283,10 +283,17 @@ var tExchanges = map[string]*core.Exchange{
...
+			"dcr": &core.FeeAsset{
+				ID:    42,
+				Confs: 1,
+				Amt:   1e8,
+			},
+			"btc": &core.FeeAsset{
+				ID:    0,
+				Confs: 1,
+				Amt:   1e6,
+			},
```
buck54321, line 296: Only had to update these because the running against the test server in the legacy mode was resulting in an error at 

https://github.com/decred/dcrdex/blob/9e1e0cb8787ca6ec48463ac006084c3a1e0ee461/client/webserver/site/src/js/forms.js#L334

Are we already done supporting older servers?

chappjc:
Alright that's fine.  Seems to me though that the main benefit is that you don't have to pass app the constructors, not necessarily elimination of the module-level variables.  I'm naive about most things JS, but all the function calls smell like overhead to me compared to a more tightly scoped variable.  It's no problem though.

PR #1237, opened 2021-10-19
app: drop ids arg to parsePage

buck54321:
Maintaining ID lists was tedious. Instead, create an object with all descendants that have id attributes set. The `page` object will be slightly larger in some places, since we'll pick up some form elements we're not using directly, but I'm looking to move those to `data-tmpl` attributes anyway.

### DIFF
--- a/client/webserver/site/src/js/app.js
+++ b/client/webserver/site/src/js/app.js
@@ -240,63 +240,57 @@ export default class Application {
     this.popupTmpl = Doc.tmplElement(this.popupNotes, 'note')
     this.popupTmpl.remove()
     this.tooltip = idel(document.body, 'tooltip')
-    const pg = this.page = Doc.parsePage(this.header, [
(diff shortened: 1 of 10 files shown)


client/webserver/site/src/js/app.js
```diff
@@ -240,63 +240,57 @@ export default class Application {
     this.popupTmpl = Doc.tmplElement(this.popupNotes, 'note')
     this.popupTmpl.remove()
     this.tooltip = idel(document.body, 'tooltip')
-    const pg = this.page = Doc.parsePage(this.header, [
-      'noteIndicator', 'noteList', 'noteTmpl', 'marketsMenuEntry',
-      'walletsMenuEntry', 'noteMenuEntry', 'loader', 'profileMenuEntry',
-      'profileIndicator', 'profileSignout', 'innerNoteIcon', 'innerProfileIcon',
```
chappjc, line 246: Nice.  This was error prone.  We've forgotten to add and remove them many times.  e.g. `profileIndicator` doesn't exist

PR #1239, opened 2021-10-19
server/asset/btc: deserialize grt hex into btcutil.Tx

chappjc:
This fixes `server/asset/btc.(*RPCClient).GetRawTransaction` so that it unmarshals the raw tx hex into a string and then deserializes into a `btcutil.Tx`.

This fixes match_status request handling for requests that require retrieving "tx data" for a BTC swap transaction.


server/asset/btc/live_test.go
```diff
@@ -45,13 +45,28 @@ import (
...
+	tx, err := btc.node.GetRawTransaction(hash)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if got := tx.MsgTx().TxHash(); got != *hash {
+		t.Errorf("Got tx id %v, expected %v", got, hash)
+	}
+}
+
 func TestMain(m *testing.M) {
```
JoeGruffins, line 70: Unrelated, but this panics if connect fails, which is probably not intended. I guess `wg` is set to nil.

PR #1243, opened 2021-10-21
server/eth: Implement FundingCoin.

JoeGruffins:
A funding coin represents value on the blockchain. For utxo this is a
    concrete value, but for account based coins it is nebulous and changing.
    For ethereum, the funding coin is represented by an address or account
    and a value that account must have.

    Authorization of the use of those funds is very similar to other coins,
    but the user only need prove they own the keys for one account or
    address.

part of #1154

### DIFF
--- a/server/asset/eth/coiner.go
+++ b/server/asset/eth/coiner.go
@@ -290,10 +290,19 @@ func (c *swapCoin) FeeRate() uint64 {
 var _ asset.FundingCoin = (*amountCoin)(nil)
 
 type amountCoin struct {
-	backend *Backend
-	cID     *AmountCoinID
+	backend  *Backend
+	cID      *AmountCoinID
+	gasPrice uint64
 }
 
+// newAmountCoin creates a asset.Funding coin for ethereum. It is comprised of
+// an account and the value we expect that account to have ready. This differes
+// from utxo coins because the value held may change at any time. The value is
+// checked to have at least as much as we expect upon creation and once every
+// time Confirmations is called. Another difference is that the tx fee, or gas
+// price, is not one number. Here we check once upon initiation if their are
+// pending transactions and just use the lowest from that time if the confirmed
+// balance doesn't reach our needed threshold.
 func (backend *Backend) newAmountCoin(coinID []byte) (*amountCoin, error) {
 	cID, err := DecodeCoinID(coinID)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


server/asset/eth/eth.go
```diff
@@ -317,9 +319,10 @@ func (eth *Backend) Redemption(redeemCoinID, contractCoinID []byte) (asset.Coin,
...
 }
 
-// FundingCoin is an unspent output.
-func (eth *Backend) FundingCoin(ctx context.Context, coinID []byte, redeemScript []byte) (asset.FundingCoin, error) {
-	return nil, notImplementedErr
+// FundingCoin is an unspent amount on the blockchain. It points to an account
+// that should maintain at least a certain value.
+func (eth *Backend) FundingCoin(ctx context.Context, coinID []byte, _ []byte) (asset.FundingCoin, error) {
+	return newAmountCoin(eth, coinID)
 }
```
JoeGruffins, line 326: I don't think we need to worry about `redeemScript`, but, It could be used to determine the spend size. I wonder if the SwapCoinID should include data on whether that coin represents an initiation or a redemption. That data, included here, would allow us to decide the spend size.

server/asset/eth/coiner.go
```diff
@@ -284,3 +286,151 @@ func (c *swapCoin) Value() uint64 {
...
+	if !ok {
+		return nil, errors.New("coin ID not an amount")
+	}
+	if _, err := accountHasBalance(backend, &aCoinID.Address, aCoinID.Amount); err != nil {
+		return nil, err
+	}
+	return &amountCoin{
+		backend: backend,
+		cID:     aCoinID,
+	}, nil
```
JoeGruffins, line 312: This uses the same nonce we get from the client. Is it important to ensure it is unique for the server?

JoeGruffins: Makes sense to me.

server/asset/eth/coiner.go
```diff
@@ -284,3 +286,143 @@ func (c *swapCoin) Value() uint64 {
...
+}
+
+// Value is the amount held in this amountCoin.
+func (c *amountCoin) Value() uint64 {
+	return c.cID.Amount
+}
+
+// FeeRate is not applicable to value held in an account.
+func (c *amountCoin) FeeRate() uint64 {
+	return 0
```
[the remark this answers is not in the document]
JoeGruffins: Added a pretty dumb method to do this basically. It just looks at the lowest pending gas upon creation and returns that.

JoeGruffins:
Test failure seems unrelated to these changes.

martonp (approved):
Looks good to me. The tests run fine on my machine.

JoeGruffins:
FundingCoin removed by #1284

PR #1244, opened 2021-10-21
run_tests: Add lgpl build tags to linter.

JoeGruffins:


client/asset/eth/eth_test.go
```diff
@@ -555,7 +555,7 @@ func TestFundOrderReturnCoinsFundingCoins(t *testing.T) {
 	node.bal.Mul(node.bal, big.NewInt(2))
 
 	// Test funding two coins with the same id
-	coins, err = eth2.FundingCoins([]dex.Bytes{coins2[0].ID(), coins2[0].ID()})
+	_, err = eth2.FundingCoins([]dex.Bytes{coins2[0].ID(), coins2[0].ID()})
```
chappjc, line 558: in the following checkBalance, should it be referencing `coins2` instead of `coins1`

chappjc (approved):
Looks good.  I didn't know about golangci-lint's `--built-tags`, but that's just the ticket.

Just one question about the test case you're touching on.

PR #1248, opened 2021-10-23
eth/client: Swap

martonp:
This diff implements the Swap function in the eth client wallet. In addition, the eth client wallet no longer does gas estimations using the eth client and uses hardcoded values instead. Gas estimations do not vary unless the eth client is updated, and these changed will be caught by the harness tests, so it is safe to just use the hardcoded values.

Part of #1154

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -727,22 +727,22 @@ func (r *swapReceipt) Expiration() time.Time {
 func (r *swapReceipt) Coin() asset.Coin {
 	return &coin{
 		value: r.value,
-		id:    dex.Bytes(r.txHash[:]),
+		id:    r.txHash[:],
 	}
 }
 
 // Contract returns the swap's secret hash.
 func (r *swapReceipt) Contract() dex.Bytes {
-	return dex.Bytes(r.secretHash[:])
+	return r.secretHash[:]
 }
 
 // String returns a string representation of the swapReceipt.
 func (r *swapReceipt) String() string {
-	return fmt.Sprintf("tx hash: %x, secret hash: %x", r.txHash, r.secretHash)
+	return fmt.Sprintf("{ tx hash: %x, secret hash: %x }", r.txHash, r.secretHash)
 }
 
 // SignedRefund returns an empty byte array. ETH does not support a pre-signed
-// redeem script becuase the nonce needed in the transaction cannot be previously
+// redeem script because the nonce needed in the transaction cannot be previously
 // determined.
 func (*swapReceipt) SignedRefund() dex.Bytes {
 	return dex.Bytes{}
@@ -861,10 +861,8 @@ func (eth *ExchangeWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin
 	eth.unlockFunds(swaps.Inputs)
 	var change asset.Coin
 	changeAmount := totalInputValue - totalUsedValue
-	if changeAmount > 0 {
+	if changeAmount > 0 && swaps.LockChange {
 		change = eth.createAmountCoin(changeAmount)
-	}
-	if swaps.LockChange && change != nil {
 		eth.lockFunds(asset.Coins{change})
 	}
 
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -47,6 +47,7 @@ const (
 	BipID              = 60
 	defaultGasFee      = 82  // gwei
 	defaultGasFeeLimit = 200 // gwei
+	MinGasTipCap       = 2   // gwei
 )
 
 var (
@@ -151,7 +152,6 @@ type ethFetcher interface {
 	connect(ctx context.Context, node *node.Node, contractAddr *common.Address) error
 	listWallets(ctx context.Context) ([]rawWallet, error)
 	initiate(txOpts *bind.TransactOpts, netID int64, initiations []dexeth.ETHSwapInitiation) (*types.Transaction, error)
-	estimateGas(ctx context.Context, msg ethereum.CallMsg) (uint64, error)
 	lock(ctx context.Context, acct *accounts.Account) error
 	nodeInfo(ctx context.Context) (*p2p.NodeInfo, error)
 	pendingTransactions(ctx context.Context) ([]*types.Transaction, error)
@@ -165,6 +165,7 @@ type ethFetcher interface {
 	swap(ctx context.Context, from *accounts.Account, secretHash [32]byte) (*dexeth.ETHSwapSwap, error)
 	unlock(ctx context.Context, pw string, acct *accounts.Account) error
 	signData(addr common.Address, data []byte) ([]byte, error)
+	suggestGasTipCap(ctx context.Context) (*big.Int, error)
 }
 
 // Check that ExchangeWallet satisfies the asset.Wallet interface.
@@ -183,15 +184,14 @@ type ExchangeWallet struct {
 	log       dex.Logger
 	tipChange func(error)
 
+	networkID int64
+
 	internalNode *node.Node
 
 	tipMtx     sync.RWMutex
 	currentTip *types.Block
 
 	acct *accounts.Account
-
-	initGasCache    map[int]uint64
-	initGasCacheMtx sync.Mutex
 
 	lockedFunds    map[string]uint64 // gwei
 	lockedFundsMtx sync.RWMutex
@@ -271,13 +271,18 @@ func NewWallet(assetCFG *asset.WalletConfig, logger dex.Logger, network dex.Netw
 				len(accounts))
 	}
 
+	networkID, err := getNetworkID(network)
+	if err != nil {
+		return nil, fmt.Errorf("NewWallet: unable to get network ID: %w", err)
+	}
+
 	return &ExchangeWallet{
 		log:          logger,
 		tipChange:    assetCFG.TipChange,
 		internalNode: node,
 		lockedFunds:  make(map[string]uint64),
 		acct:         &accounts[0],
-		initGasCache: make(map[int]uint64),
+		networkID:    networkID,
 	}, nil
 }
 
@@ -465,48 +470,6 @@ func (eth *ExchangeWallet) balanceImpl() (*asset.Balance, error) {
 	return bal, nil
 }
 
-// getInitGas gets an estimate for the gas required for initiating a swap. The
-// result is cached and reused in subsequent requests.
-//
-// This function will return an error if the wallet has a non zero balance.
-func (eth *ExchangeWallet) getInitGas(numSwaps int) (uint64, error) {
-	eth.initGasCacheMtx.Lock()
-	defer eth.initGasCacheMtx.Unlock()
-
-	if cachedGas, ok := eth.initGasCache[numSwaps]; ok {
-		return cachedGas, nil
-	}
-
-	var secretHash [32]byte
-	copy(secretHash[:], encode.RandomBytes(32))
-	initiations := make([]dexeth.ETHSwapInitiation, 0, numSwaps)
-	for i := 0; i < numSwaps; i++ {
-		initiations = append(initiations, dexeth.ETHSwapInitiation{
-			RefundTimestamp: big.NewInt(1),
-			SecretHash:      secretHash,
-			Participant:     eth.acct.Address,
-			Value:           big.NewInt(1),
-		})
-	}
-	data, err := dexeth.PackInitiateData(initiations)
-	if err != nil {
-		return 0, err
-	}
-	msg := ethereum.CallMsg{
-		From:  eth.acct.Address,
-		To:    &mainnetContractAddr,
-		Value: big.NewInt(int64(numSwaps)),
-		Gas:   0,
-		Data:  data,
-	}
-	gas, err := eth.node.estimateGas(eth.ctx, msg)
-	if err != nil {
-		return 0, err
-	}
-	eth.initGasCache[numSwaps] = gas
-	return gas, nil
-}
-
 // MaxOrder generates information about the maximum order size and associated
 // fees that the wallet can support for the given DEX configuration. The fees are an
 // estimate based on current network conditions, and will be <= the fees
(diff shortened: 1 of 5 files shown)


client/asset/eth/eth.go
```diff
@@ -670,12 +691,155 @@ func (eth *ExchangeWallet) unlockFunds(coins asset.Coins) error {
...
+	}
+
+	opts := bind.TransactOpts{
+		From:      eth.acct.Address,
+		Value:     big.NewInt(0).SetUint64(totalContractValue),
+		GasFeeCap: big.NewInt(0).SetUint64(swaps.FeeRate),
+		GasTipCap: big.NewInt(GasTipCap),
+		Context:   eth.ctx}
+
+	tx, err := eth.node.initiate(&opts, eth.networkID, initiations)
```
chappjc, line 828: Do we have a bigger problem with EIP-1559.  If we lock value (the "change" amount) based on `getInitGas` and `swaps.FeeRate` (the `GasFeeCap`), then `tx` gets mined using some unknown lower gas cost, the actual amount available is something else.
I suppose this just means we'll have more locked than is necessary until the order is fully executed and the change is returned?

client/asset/eth/eth.go
```diff
@@ -670,12 +691,155 @@ func (eth *ExchangeWallet) unlockFunds(coins asset.Coins) error {
...
+	var change asset.Coin
+	changeAmount := totalInputValue - totalUsedValue
+	if changeAmount > 0 {
+		change = eth.createAmountCoin(changeAmount)
+	}
+	if swaps.LockChange && change != nil {
+		eth.lockFunds(asset.Coins{change})
+	}
+
+	feesUsed := swaps.FeeRate * initGas
```
chappjc, line 861: We have `initGas * swaps.FeeRate` above when calculating `totalUsedValue`, so how about we assign it to a variable like `maxPossibleFee` to convey it's meaning better in both places.

client/asset/eth/eth.go
martonp: That's right. We could just use a legacy transaction and all the funds could be used up and this issue wouldn't exist, but I think we might as well let the user save a little bit.

chappjc: OK, we'll use an eip 1559 txn, but let's doc the unknown actual cost issue somewhere in this function.

client/asset/eth/eth.go
```diff
@@ -668,12 +619,166 @@ func (eth *ExchangeWallet) unlockFunds(coins asset.Coins) error {
...
+
+// String returns a string representation of the swapReceipt.
+func (r *swapReceipt) String() string {
+	return fmt.Sprintf("tx hash: %x, secret hash: %x", r.txHash, r.secretHash)
+}
+
+// SignedRefund returns an empty byte array. ETH does not support a pre-signed
+// redeem script becuase the nonce needed in the transaction can not be previously
+// determined. We will need to come up with a different way to support manual
+// refunds in ETH.
```
JoeGruffins, line 659: Could you add the `TODO:` tag to this somewhere? That makes it easier to find all the todo's later.

client/asset/eth/eth.go
```diff
@@ -668,12 +619,166 @@ func (eth *ExchangeWallet) unlockFunds(coins asset.Coins) error {
...
+func (*swapReceipt) SignedRefund() dex.Bytes {
+	return dex.Bytes{}
+}
+
+var _ asset.Receipt = (*swapReceipt)(nil)
+
+// Swap sends the swaps in a single transaction. The Receipts returned can
+// be used to refund a failed transaction. The fees used returned are the max
```
JoeGruffins, line 667: Maybe the Receipts comment is wrong for eth. They cannot be used to refund, because the the signature will need to include the next nonce for the account to be accepted by the network. Right? And those must be in ascending order incrementing by one for each mined transaction.

client/asset/eth/eth.go
```diff
@@ -668,12 +619,166 @@ func (eth *ExchangeWallet) unlockFunds(coins asset.Coins) error {
...
+			return fail(fmt.Errorf("Swap: address in contract is not valid %v", contract.Address))
+		}
+		address := common.HexToAddress(contract.Address)
+
+		initiations = append(initiations,
+			dexeth.ETHSwapInitiation{
+				RefundTimestamp: big.NewInt(0).SetUint64(contract.LockTime),
+				SecretHash:      secretHash,
+				Participant:     address,
+				Value:           big.NewInt(0).SetUint64(contract.Value),
```
JoeGruffins, line 717: I think this need to be converted to wei from gwei here correct? I believe all uint64 values the server deals with are gwei, and the actual contract wants the basic unit.

client/asset/eth/eth.go
```diff
@@ -668,12 +619,166 @@ func (eth *ExchangeWallet) unlockFunds(coins asset.Coins) error {
...
+	if suggestedGasTipCapGwei > MinGasTipCap {
+		gasTipCap = suggestedGasTipCap
+	} else {
+		gasTipCap = big.NewInt(MinGasTipCap * srveth.GweiFactor)
+	}
+
+	opts := bind.TransactOpts{
+		From:      eth.acct.Address,
+		Value:     big.NewInt(0).SetUint64(totalContractValue),
+		GasFeeCap: big.NewInt(0).SetUint64(swaps.FeeRate),
```
JoeGruffins, line 750: Don't these also need to be converted from gwei to wei? I could be wrong, just let me know.

client/asset/eth/eth_test.go
```diff
@@ -108,16 +119,34 @@ func (n *testNode) syncProgress(ctx context.Context) (*ethereum.SyncProgress, er
 func (n *testNode) pendingTransactions(ctx context.Context) ([]*types.Transaction, error) {
 	return nil, nil
 }
-func (n *testNode) initiate(opts *bind.TransactOpts, netID int64, initiations []swap.ETHSwapInitiation) (*types.Transaction, error) {
-	return nil, nil
+func (n *testNode) initiate(opts *bind.TransactOpts, netID int64, initiations []dexeth.ETHSwapInitiation) (*types.Transaction, error) {
```
JoeGruffins, line 122: Maybe worth noting that this is not concurrent safe because it sets a few values, although the real initiate probably is. It's probably fine that way.

JoeGruffins (commented):
Looks good. Could you make the commit messages longer? Maybe comment on the need to remove getInitGas and some other stuff this does.

client/asset/eth/eth.go
martonp: I just made in issue instead.

client/asset/eth/eth.go
martonp: Yes this was wrong.

client/asset/eth/eth.go
martonp: Yes

client/asset/eth/eth_test.go
martonp: Yeah in the real one only the locked funds are updated, and those are protected by a mutex. The tests aren't expected to be concurrent safe are they?

client/asset/eth/eth.go
martonp: Yes that's right, the comment is wrong.

client/asset/eth/eth.go
```diff
@@ -506,14 +453,18 @@ func (*ExchangeWallet) PreRedeem(req *asset.PreRedeemForm) (*asset.PreRedeem, er
 
 // coin implements the asset.Coin interface for ETH
 type coin struct {
-	id srveth.AmountCoinID
+	id dex.Bytes
```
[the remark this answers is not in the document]
martonp: I'll leave this part for now, the previous agreement was to just send the transaction hash, but I'm thinking it would be better to send the entire `TxCoinID` including the index. I commented over here:
https://github.com/decred/dcrdex/issues/1154#issuecomment-979251509

client/asset/eth/eth_test.go
chappjc: That depends if they are used concurrently.  :p

To ensure the product code does not have data races, which have security implications in addition to the potential to cause incorrect behavior and even to panic the app (e.g. concurrent map writes), we run tests with the race detector.  If there is no concurrent use of the method, it doesn't need a sync mechanism, but otherwise it would.

As an aside, having the race detector on with tests helps us be acutely aware of the access patterns of the product code using these stubs.  In many cases we're better of rethinking design than slapping a mutex on everything.  In particular, I'd think these methods don't need synchronization mechanisms (under what circumstances would they be called concurrently?).  A common anti-pattern we face is each test case not cleaning up after itself, and subsequent tests causing data races.  Meddling in internals too much is another one.  It's easy to ignore these issue without the race detector.  I know you weren't suggesting to turn it off though.

I'm always in favor of designing code, both product and test, in such a way that it is lock-free.  That's obv not always possible, just saying.

martonp: Makes sense. As long as we're not reusing a `testNode` between tests, we'll be good.

client/asset/eth/eth_test.go
```diff
@@ -883,6 +824,396 @@ func TestPreSwap(t *testing.T) {
...
+		},
+	}
+	inputs = refreshWalletAndFundCoins(5, []uint64{ethToGwei(2)})
+	swaps = asset.Swaps{
+		Inputs:     inputs,
+		Contracts:  contracts,
+		FeeRate:    200,
+		LockChange: false,
+	}
+	testSwap("one contract, don't lock change", swaps, false)
```
buck54321, line 1124: Probably OK, but we really shouldn't need a change coin if `LockChange` is `false`. **core** shouldn't require it. It does look like we are returning the change coin for other assets. I don't think that's on purpose.

server/asset/eth/common.go
```diff
@@ -226,6 +227,16 @@ func decodeAmountCoinID(coinID []byte) (*AmountCoinID, error) {
...
+func CreateAmountCoinID(address common.Address, amount uint64) *AmountCoinID {
+	var nonce [8]byte
+	copy(nonce[:], encode.RandomBytes(8))
+	return &AmountCoinID{
+		Address: address,
+		Amount:  amount,
+		Nonce:   nonce,
+	}
+}
+
```
buck54321, line 239: As per convo elsewhere, we will probably only use this client-side. Can move and unexport here or separately. Up to you.

client/asset/eth/eth.go
```diff
@@ -668,12 +619,168 @@ func (eth *ExchangeWallet) unlockFunds(coins asset.Coins) error {
...
+	return r.expiration
+}
+
+// Coin returns the coin used to fund the swap.
+func (r *swapReceipt) Coin() asset.Coin {
+	return &coin{
+		value: r.value,
+		id:    dex.Bytes(r.txHash[:]),
+	}
+}
```
buck54321, line 644: Probably just the `txHash` here.

client/asset/eth/eth.go
```diff
@@ -668,12 +619,168 @@ func (eth *ExchangeWallet) unlockFunds(coins asset.Coins) error {
...
+
+// Contract returns the swap's secret hash.
+func (r *swapReceipt) Contract() dex.Bytes {
+	return dex.Bytes(r.secretHash[:])
```
buck54321, line 648: `[]byte` is about the only thing in Go that you can cast implicitly. Just `return r.secretHash[:]`.

client/asset/eth/eth.go
```diff
@@ -668,12 +619,168 @@ func (eth *ExchangeWallet) unlockFunds(coins asset.Coins) error {
...
+}
+
+// String returns a string representation of the swapReceipt.
+func (r *swapReceipt) String() string {
+	return fmt.Sprintf("tx hash: %x, secret hash: %x", r.txHash, r.secretHash)
```
buck54321, line 653: How do you feel about wrapping the output in a delimiter?

```
return fmt.Sprintf("{ tx hash: %x, secret hash: %x }", r.txHash, r.secretHash)
```

client/asset/eth/eth.go
```diff
@@ -668,12 +619,168 @@ func (eth *ExchangeWallet) unlockFunds(coins asset.Coins) error {
...
+}
+
+// SignedRefund returns an empty byte array. ETH does not support a pre-signed
+// redeem script becuase the nonce needed in the transaction cannot be previously
```
buck54321, line 657: because

client/asset/eth/eth.go
```diff
@@ -668,12 +619,168 @@ func (eth *ExchangeWallet) unlockFunds(coins asset.Coins) error {
...
+// determined.
+func (*swapReceipt) SignedRefund() dex.Bytes {
+	return dex.Bytes{}
```
buck54321, line 660: We'll have to do something, but we can punt for now.

client/asset/eth/eth_test.go
```diff
@@ -883,6 +824,396 @@ func TestPreSwap(t *testing.T) {
...
+					testName, time.Unix(expiration, 0), receipts[0].Expiration())
+			}
+			if receipt.Coin().Value() != contract.Value {
+				t.Fatalf("%v: receipt coin value: %v != expected: %v",
+					testName, receipt.Coin().Value(), contract.Value)
+			}
+			if len(receipt.Contract()) != srveth.SecretHashSize {
+				t.Fatalf("%v: expected length of contract to be %v but got %v",
+					testName, srveth.SecretHashSize, len(receipt.Contract()))
+			}
```
buck54321, line 922: Length check is implicit in `bytes.Equal`.

buck54321 (changes requested):
Obviously a lot going to conflict with #1301, but this was created first, so I can handle the rebase if this is ready before.

client/asset/eth/eth.go
chappjc: Issue: https://github.com/decred/dcrdex/issues/1311

client/asset/eth/eth.go
```diff
@@ -668,12 +619,168 @@ func (eth *ExchangeWallet) unlockFunds(coins asset.Coins) error {
...
+
+	totalUsedValue := totalContractValue + maxPossibleFee
+	if totalInputValue < totalUsedValue {
+		return fail(fmt.Errorf("Swap: coin inputs value %d < required %d", totalInputValue, totalUsedValue))
+	}
+
+	suggestedGasTipCap, err := eth.node.suggestGasTipCap(eth.ctx)
+	if err != nil {
+		return fail(fmt.Errorf("Swap: failed to get suggested gas tip cap %w", err))
+	}
```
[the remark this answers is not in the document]
martonp: `GasFeeCap` will always be the maximum fee used. The actual gas used will be `BaseFee + Tip`. The tip will only be less than the the `GasTipCap` if for example `BaseFee=99`,  `GasFeeCap=100`, and `GasTipCap=2`. In this case,  only 1 gwei will be spent for the tip.

My plan here was to have 2 be the default `GasTipCap` and the server can validate that clients put at least 2. Generally all wallets just use 1 for the `GasTipCap` though, 2 is for more safety and expediency. If the node recommends that the `GasTipCap` should be higher, then a higher one is used, but the `dex.Asset.MaxFeeRate` will still never be exceeded.

client/asset/eth/eth.go
martonp: Sounds good to me.

server/asset/eth/common.go
martonp: I'll do it separately.

PR #1262, opened 2021-10-31
eth: Remove original initiate function

### LINKED ISSUE #1260: eth: Remove initiate function from Swap contract
Since initiating a single swap can be done with the `initiateBatch` function and the gas savings of using the original `initiate` function are negligible, in the interest of having less solidity code and simpler ETH client/server code, the original `initiate` function should be removed.

martonp:
This diff updates the eth swap contract by removing the original
initiate function which only allows initiating one swap at a
time and renaming initiateBatch to initiate. The gas savings
from initiating a single swap using the old initiate function
are negligible and removing it greatly simplifies the codebase.

Closes #1260


chappjc:
I'm not totally sure we want to remove it.  If there is an issue with initiateBatch, we'd want the regular initiate function available so we don't have to deploy a new contract and deal with migrating to the new one

JoeGruffins:
I'm for this change for the reduction of code surface of the contract. initiateBatch and initiate are pretty similar, so I can't image why one breaks but not the other, but I can understand the reservation. Will hold of on reviewing I guess unless @chappjc comes around.

martonp:
I agree with @JoeGruffins.. they are very similar. Also, in order to be able to conveniently switch from using one function to the other, just leaving both functions in the solidity code wouldn't be enough. We would need to make both the client and server support both, and then have an option in the client to pick one or the other. Unless we have this, there would have to be new versions of both the client and server before people could switch.

As a side note, it might be worth considering putting an emergency shutdown switch that can be switched on by certain admin addresses and blocks the initiate function. This would just stop people from initiating swaps, but not affect any other functionality. In the case of a security issue, no new money would go into the smart contract.

chappjc:
Updating the software to use one method or another is simpler if there is no need to deploy a contract and switch the contract address.  The server checks shouldn't even care which method you used to insert the swap, nor should the counterparty.

I'm not totally opposed to this PR, but I don't see the downside to keeping the initiate method.  Things in ethereum contracts look simple until something subtle wrecks it.

I am opposed to an admin address with shutoff permissions though.

JoeGruffins:
This removes code from the contract, leaving less things to break, is my take.

chappjc:
Alright then.

dex/networks/eth/txdata.go
```diff
@@ -43,22 +43,22 @@ func ParseInitiateData(calldata []byte) (*common.Address, [32]byte, int64, error
 	if len(args) != numInputArgs {
 		return fail(fmt.Errorf("expected %v input args but got %v", numInputArgs, len(args)))
 	}
-	locktime, ok := args[0].value.(*big.Int)
+	initiations, ok := args[0].value.([]struct {
+		RefundTimestamp *big.Int       `json:"refundTimestamp"`
+		SecretHash      [32]byte       `json:"secretHash"`
+		Participant     common.Address `json:"participant"`
+		Value           *big.Int       `json:"value"`
+	})
```
JoeGruffins, line 51: Looks like we need the json tags to convert here? I guess it's fine because this won't build if you can't cast to `ETHSwapInitiation` below. I would have thought structs have default tags like this but I guess not.

dex/networks/eth/txdata_test.go
```diff
@@ -18,16 +22,60 @@ func mustParseHex(s string) []byte {
 	return b
 }
 
+func initiationsAreEqual(i1, i2 ETHSwapInitiation) bool {
+	return i1.RefundTimestamp.Cmp(i2.RefundTimestamp) == 0 &&
+		bytes.Equal(i1.SecretHash[:], i2.SecretHash[:]) &&
+		bytes.Equal(i1.Participant.Bytes(), i2.Participant.Bytes()) &&
```
JoeGruffins, line 28: You can also just do:
```
...
i1.SecretHash == i2.SecretHash &&
i1.Participant == i2.Participant &&
...
```
Because they are arrays and not slices.

Also not important, but using numerals after variable names is kind of confusing to me. I would prefer maybe just a, b, c, ... but it's probably just a style thing. `1` in particular can be confusing as a variable name because it kind of looks like an upper case I or a lower case l or the numeral 1 or even or |.

dex/networks/eth/txdata_test.go
```diff
@@ -18,16 +22,60 @@ func mustParseHex(s string) []byte {
...
+		i1.Value.Cmp(i2.Value) == 0
+}
+
 func TestParseInitiateData(t *testing.T) {
-	participantAddr := common.HexToAddress("8d83B207674bfd53B418a6E47DA148F5bFeCc652")
-	secretHashSlice := mustParseHex("4aec4dc47fc6bd1fd5091c1aa4067c7fbd6bbcdb476209756354ec784d6082dc")
+	participantAddr := common.HexToAddress("345853e21b1d475582E71cC269124eD5e2dD3422")
 	var secretHash [32]byte
-	copy(secretHash[:], secretHashSlice)
+	var secretHash2 [32]byte
```
JoeGruffins, line 35: Again sorry for the nit pik but would prefer secretHashA and secretHashB, or something like that rather than a number. Feel free to ignore.

JoeGruffins (approved):
Changes look good to me. Hope you can add back in a live `initiate` test though.

client/asset/eth/rpcclient_harness_test.go
```diff
@@ -721,100 +686,24 @@ func TestInitiateBatch(t *testing.T) {
 	}
 }
 
-func TestInitiate(t *testing.T) {
```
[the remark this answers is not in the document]
martonp: `TestInitiateBatch` is renamed to `TestInitiate` and the test for the old initiate function is removed.

dex/networks/eth/txdata.go
```diff
@@ -15,18 +15,18 @@ import (
...
-// the secret hash, and locktime.
-func ParseInitiateData(calldata []byte) (*common.Address, [32]byte, int64, error) {
-	fail := func(err error) (*common.Address, [32]byte, int64, error) {
-		return nil, [32]byte{}, 0, err
+// initiate with expected argument types. It returns the array of initiations
+// with which initiate was called.
+func ParseInitiateData(calldata []byte) ([]ETHSwapInitiation, error) {
+	fail := func(err error) ([]ETHSwapInitiation, error) {
+		return nil, err
 	}
```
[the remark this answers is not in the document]
martonp: True

dex/networks/eth/txdata.go
martonp: Yeah the fakeArg in the code you copy pasted from geth adds the tags and you need the correct tags to do the type conversion.
Someone was trying to update go to ignore tags at one point, but I guess it didn't get implemented:
https://go.googlesource.com/proposal/+/master/design/16085-conversions-ignore-tags.md

dex/networks/eth/txdata_test.go
martonp: It's all good, go ahead and nit pik :). I'll update it. I think it's nicer too.

client/asset/eth/rpcclient_harness_test.go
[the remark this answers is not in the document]
JoeGruffins: Oh, ok.

JoeGruffins (approved):

server/asset/eth/common.go
```diff
@@ -81,19 +81,21 @@ const (
 // to the smart contract. This type of ID is useful to identify coins that
 // were sent in transactions that have not yet been mined.
 type TxCoinID struct {
-	TxID [32]byte
+	TxID  [32]byte
+	Index uint32
```
chappjc, line 85: Will you please document Index in the TxCoinID docs when you rebase.

martonp: Done

chappjc: Yeesh I'm blind, it was already described.  Thanks

martonp: No it wasn't haha, I did one update as a part of the rebase, then I did another small update.

client/asset/eth/eth.go
```diff
@@ -384,12 +385,12 @@ func (eth *ExchangeWallet) balanceImpl() (*asset.Balance, error) {
...
-	defer eth.cachedInitGasMtx.Unlock()
+func (eth *ExchangeWallet) getInitGas(numSwaps int) (uint64, error) {
+	eth.initGasCacheMtx.Lock()
+	defer eth.initGasCacheMtx.Unlock()
 
-	if eth.cachedInitGas > 0 {
-		return eth.cachedInitGas, nil
+	if cachedGas, ok := eth.initGasCache[numSwaps]; ok {
+		return cachedGas, nil
 	}
```
buck54321, line 394: I'm not crazy about this solution. Have we decided against using constants derived from your previous results?

```
Gas used for batch initiating 1 swaps: 180037. 180037 more than previous 
Gas used for batch initiating 2 swaps: 338287. 158250 more than previous 
Gas used for batch initiating 3 swaps: 496537. 158250 more than previous 
Gas used for batch initiating 4 swaps: 654775. 158238 more than previous 
Gas used for batch initiating 5 swaps: 813025. 158250 more than previous 
Gas used for batch initiating 6 swaps: 971275. 158250 more than previous 
Gas used for batch initiating 7 swaps: 1129513. 158238 more than previous 
Gas used for batch initiating 8 swaps: 1287763. 158250 more than previous 
Gas used for batch initiating 9 swaps: 1446025. 158262 more than previous
```

martonp: Yeah I think we might as well do that. It'll make everything more simple.

martonp: How about I do a separate PR for this.

buck54321: Not a priority. I'm already using `InitGas = 181000` in some in-progress work.

martonp: It's actually updated to `InitGas = 157000` now. I also have `AdditionalInitGas` in there now, but I don't think you should worry about. Generally you should just assume the worst case scenario.

martonp:
The tests are failing due to some race condition in the BTC tests, unrelated to these changes.

client/asset/eth/eth.go
buck54321: I don't follow. What is `AdditionalInitGas` and why would `InitGas` it be 157000? A single lot swap is `180037`, so we do `lots * 180037` to get the required fee reserves. So `SwapSize = SwapSizeBase = InitGas = 181000` (with `inputsSize = 0` in e.g. `calc.RequiredOrderFunds`) makes the math work out everywhere.

buck54321: Ah. Values are outdated. 

```
Gas used for batch initiating 1 swaps: 156485. 156485 more than previous 
Gas used for batch initiating 2 swaps: 291127. 134642 more than previous 
Gas used for batch initiating 3 swaps: 425793. 134666 more than previous
```

PR #1263, opened 2021-11-02
client/db: active and archived match buckets

chappjc:
This builds on https://github.com/decred/dcrdex/pull/957.  @itswisdomagain's commits will be merged without squashing into this.  I've added his v5 test db generator app as well, just with an added cancel order match.

This redefines the v6 client db upgrade to introduce a new active matches bucket instead of creating a `Retired` flag on each match.

The db and upgrade code is modeled after the active/archived orders buckets changes in https://github.com/decred/dcrdex/pull/1071, whereby existing matches remain in what is to be the archived bucket, and only active ones are moved into the new active matches bucket.

`client/db.MatchIsActive` is updated to detect cancel matches as inactive. This is detected as `MatchComplete` and `len(proof.Auth.InitSig) == 0` since cancel matches begin as complete and have no init/redeem data set in the `Auth` field.
The `Address` field is only empty for the maker/trade match with a cancel, but for the taker/cancel's match with the trade it has to this point contained the canceled maker order's address.

This also updates client/core to clear the `Address` field for both sides of a cancel match because it is pointless to store. Removing this useless data from a `MetaMatch` will save a bit of storage space, but previous cancel matches are not updated by the v6 upgrade.  Similarly, the server's swapper is updated to not send the maker address with a cancel `msgjson.Match`.

Shut down the db tests properly.

**Maybe 0.3**


client/db/types.go
```diff
@@ -275,13 +275,18 @@ func (m *MetaMatch) MatchOrderUniqueID() []byte {
 }
 
 // MatchIsActive returns false (i.e. the match is inactive) if: (1) status is
-// complete AND proof.RedeemSig is set, (2) the match is refunded, or (3) it is
+// complete AND (proof.RedeemSig is set OR no match address signaling a cancel
```
chappjc, line 278: needs update to refer to InitSig instead of Address

client/db/bolt/upgrades.go
```diff
@@ -69,6 +71,14 @@ func (db *BoltDB) upgradeDB() error {
...
 	db.log.Infof("Upgrading database from version %d to %d", version, DBVersion)
 
+	// matches, err := db.filteredMatches(func(*bbolt.Bucket) bool { return true }, false, true)
+	// if err != nil {
+	// 	return err
+	// }
+	// for i := range matches {
+	// 	db.log.Info(matches[i])
+	// }
+
```
JoeGruffins, line 81: leftover log?

client/db/bolt/upgrades.go
```diff
@@ -269,6 +281,85 @@ func v5Upgrade(dbtx *bbolt.Tx) error {
...
+		// address. Detect this and patch the address field to reflect that it
+		// is a cancel order match with no counterparty/trade.
+		if match.Status == order.MatchComplete && len(proof.Auth.InitSig) == 0 {
+			// Set match.Address to empty so it is considered inactive in the
+			// future by MatchIsActive.
+			fmt.Println("patching complete cancel match", match.MatchID)
+			match.Address = ""
+			return archivedMBkt.Put(matchKey, order.EncodeMatch(match))
+		}
+		*/
```
JoeGruffins, line 347: Leaving this comment in on purpose?

JoeGruffins (approved):
Works well. Some of the code looks really good to.

I didn't realized the UI was updated until just now. Looks good.

client/core/core.go
```diff
@@ -4673,7 +4673,7 @@ func (c *Core) dbTrackers(dc *dexConnection) (map[order.OrderID]*trackedTrade, e
 		for _, dbMatch := range dbMatches {
 			// Only trade matches are added to the matches map. Detect and skip
 			// cancel order matches, which have an empty Address field.
-			if dbMatch.Address == "" {
+			if dbMatch.Address == "" { // only correct for maker's cancel match
```
chappjc, line 4676: This is a sorta redundant check anyway since `MatchesForOrder` a few lines up should screen them out, but if it fails to do so, then this check is effective for screening out this trade order's cancel match.

client/db/bolt/db.go
```diff
@@ -91,9 +92,6 @@ var (
 	outerKeyParamsKey      = []byte("outerKeyParams")
 	legacyKeyParamsKey     = []byte("keyParams")
 	byteTrue               = encode.ByteTrue
-	byteFalse              = encode.ByteFalse
-	byteEpoch              = uint16Bytes(uint16(order.OrderStatusEpoch))
-	byteBooked             = uint16Bytes(uint16(order.OrderStatusBooked))
```
chappjc, line 96: Were used before the orders bucket was split.

client/db/bolt/upgrades.go
```diff
@@ -15,6 +15,13 @@ import (
 
 type upgradefunc func(tx *bbolt.Tx) error
 
+// TODO: consider changing upgradefunc to accept a bbolt.DB so it may create its
+// own transactions. The individual upgrades have to happen in separate
+// transactions because they get too big for a single transaction. Plus we
+// should consider switching certain upgrades from ForEach to a more cumbersome
+// Cursor()-based iteration that will facilitate partitioning updates into
+// smaller batches of buckets.
```
buck54321, line 23: 👍

client/db/bolt/upgrades.go
```diff
@@ -118,7 +127,8 @@ func v1Upgrade(dbtx *bbolt.Tx) error {
 		return fmt.Errorf("appBucket not found")
 	}
 	skipCancels := true // cancel matches don't get revoked, only cancel orders
-	return reloadMatchProofs(dbtx, skipCancels)
+	matchesBucket := []byte("matches")
+	return reloadMatchProofs(dbtx, skipCancels, matchesBucket)
```
buck54321, line 131: `archivedMatchesBucket`? Or would that be confusing?

client/db/bolt/upgrades.go
```diff
@@ -269,6 +280,76 @@ func v5Upgrade(dbtx *bbolt.Tx) error {
...
+// potentially huge set of matches and bolt eats too much memory with ForEach.
+func v6Upgrade(dbtx *bbolt.Tx) error {
+	const oldVersion = 5
+
+	if err := ensureVersion(dbtx, oldVersion); err != nil {
+		return err
+	}
+
+	oldMatchesBucket := []byte("matches")
+	newActiveMatchesBucket := []byte("activeMatches")
```
buck54321, line 303: Not using `activeMatchesBucket` and `archivedMatchesBucket` because this is better semantics?

buck54321 (approved):

client/db/bolt/upgrades.go
chappjc: Just trying to make these older upgrades more robust to future changes to package level variables by defining the bucket/key names as they are known to be at that particular version.

client/db/bolt/upgrades.go
chappjc: Missed this comment, sorry.  Same consideration as other comment - a v7+ upgrade might change the package-level variables, but this v6 upgrade will always need to work with a specific key as it was in v5

PR #1264, opened 2021-11-02
client/core: only subscribe to price feed for persistent conns

chappjc:
Temporary DEX connections should not subscribe to the price feed.
This launches the `subPriceFeed` goroutine only when the listen goroutine
is started with a persistent dex connection.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1420,12 +1420,12 @@ func (c *Core) connectAndUpdateWallet(w *xcWallet) error {
 	addr := w.currentDepositAddress()
 	newAddr, err := c.connectWallet(w)
 	if err != nil {
-		return err // core.Error with code connectWalletErr
+		return fmt.Errorf("connectWallet: %w", err) // core.Error with code connectWalletErr
 	}
 	if newAddr != addr {
 		c.log.Infof("New deposit address for %v wallet: %v", unbip(assetID), newAddr)
 		if err = c.storeDepositAddress(w.dbID, newAddr); err != nil {
-			return err
+			return fmt.Errorf("storeDepositAddress: %w", err)
 		}
 	}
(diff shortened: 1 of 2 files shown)


client/core/core.go
```diff
@@ -1420,12 +1420,12 @@ func (c *Core) connectAndUpdateWallet(w *xcWallet) error {
...
 	newAddr, err := c.connectWallet(w)
 	if err != nil {
-		return err // core.Error with code connectWalletErr
+		return fmt.Errorf("connectWallet: %w", err) // core.Error with code connectWalletErr
 	}
 	if newAddr != addr {
 		c.log.Infof("New deposit address for %v wallet: %v", unbip(assetID), newAddr)
 		if err = c.storeDepositAddress(w.dbID, newAddr); err != nil {
-			return err
+			return fmt.Errorf("storeDepositAddress: %w", err)
```
chappjc, line 1428: trying to debug wallet connect timeouts a bit with these.  I'm pretty sure it's just trying to open another wallet instance using the same btcwallet db though

client/core/core.go
```diff
@@ -2561,6 +2561,7 @@ func (c *Core) DiscoverAccount(dexAddr string, appPW []byte, certI interface{})
 			atomic.StoreUint32(&dc.reportingConnects, 1)
 			c.wg.Add(1)
 			go c.listen(dc)
+			go dc.subPriceFeed()
```
chappjc, line 2564: On second though, I wonder if `listen` should just do `dc.subPriceFeed`.

JoeGruffins: It looks like `dc.subPriceFeed` is always called after `listen` so that makes sense.

JoeGruffins (approved):

buck54321 (approved):

Issue #1246, opened 2021-10-22
client/core/trade_simnet_test: Various test failures.


JoeGruffins:
Random core simnet test failures at commit d202f6d4d0de

```
2021-10-22 11:45:34.025 [TRC] CORE:9: notify: |DATA| (balance) balance updated
2021-10-22 11:45:34.025 [INF] TEST: [client 1] placed order selling 10.00000000 dcr at 0.00025000 btc/dcr (59b7c18be3dd2a04)
2021-10-22 11:45:34.133 [DBG] CORE:10[btc]: Attempting to fund order for 0.0025 btc, maxFeeRate = 100, max swaps = 1
    trade_simnet_test.go:341: client2 place buy order error: FundOrder error for btc, funding quantity 250000 (1 lots): insufficient funds. 0.0025 requested, 0 available
2021-10-22 11:45:34.133 [ERR] CORE:9[DB]: unable to backup database: unable to create backup directory: mkdir /tmp/1272056831/backup: file exists
2021-10-22 11:45:34.133 [DBG] CORE:9: notify: |POKE| (conn) Server disconnect - 127.0.0.1:17273 is disconnected
...
2021-10-22 11:45:34.235 [INF] CORE:10: DEX client core off
2021-10-22 11:45:34.235 [INF] TEST: [client 2] Client 2 done
--- FAIL: TestMakerGhostingAfterTakerRedeem (27.38s)
```
-----
```
=== RUN   TestOrderStatusReconciliation
...
2021-10-22 11:50:20.501 [ERR] CORE:12: Connecting to wallet btc failed: couldn't load wallet: timeout
2021-10-22 11:50:20.502 [ERR] CORE:12: notify: |ERROR| (order) Order load failure - Some orders failed to load from the database: 127.0.0.1:17273: {could not complete order 36fa35bd because the wallet for btc cannot be used, could not complete order 6cc54944 because the wallet for btc cannot be used, could not complete order cb3f4b2e because the wallet for btc cannot be used}
2021-10-22 11:50:20.507 [DBG] CORE:12: Authenticated connection to 127.0.0.1:17273, acct 3d85e236253b0908598681ba570993c718630787bdf993bd4125cc8f802d69c8, 0 active orders, 0 active matches, score 6 (suspended = false)
    trade_simnet_test.go:714: client 2 order 36fa35bd5b258aba7481b72c194dc1cb8a4493cce8fef66706e553e440209eae not found after re-initializing core
[...]

PR #1265, opened 2021-11-02
trade_simnet_test fixes

chappjc:
Work toward fixing https://github.com/decred/dcrdex/issues/1246, just the trade tests though, not the order status or pending requests tests.

Allow spv wallet to sync before funding it.  Wait a bit after funding it for it to filter the new blocks and update balance.

Add a sleepFactor that multiplies various test sleeps when the race detector is enabled.

Add a sleep in places before mining blocks, where it follows an SPV wallet doing a broadcast.  The send transaction commands return from wallet before the txn actually reaches miners.  Always was a race we were winning until SPV.

### DIFF
--- a/client/asset/btc/spv.go
+++ b/client/asset/btc/spv.go
@@ -780,7 +780,10 @@ func (w *spvWallet) swapConfirmations(txHash *chainhash.Hash, vout uint32, pkScr
 	var assumedMempool bool
 	switch err {
 	case WalletTransactionNotFound:
+		w.log.Tracef("swapConfirmations - WalletTransactionNotFound: %v:%d", txHash, vout)
 	case SpentStatusUnknown:
+		w.log.Tracef("swapConfirmations - SpentStatusUnknown: %v:%d (block %v, confs %d)",
+			txHash, vout, blockHash, confs)
 		if blockHash == nil {
 			// We generated this swap, but it probably hasn't been mined yet.
 			// It's SpentStatusUnknown because the wallet doesn't track the
@@ -804,6 +807,8 @@ func (w *spvWallet) swapConfirmations(txHash *chainhash.Hash, vout uint32, pkScr
 	}
 
 	// Our last option is neutrino.
+	w.log.Tracef("swapConfirmations - scanFilters: %v:%d (block %v, start time %v)",
+		txHash, vout, blockHash, startTime)
 	utxo, err := w.scanFilters(txHash, vout, pkScript, startTime, blockHash)
 	if err != nil {
 		return 0, false, err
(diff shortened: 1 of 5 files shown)


client/core/trade_simnet_test.go
```diff
@@ -172,9 +172,15 @@ func startClients(ctx context.Context) error {
 			if err != nil {
 				return err
 			}
-			c.log("Connected %s wallet", unbip(assetID))
+			c.log("Connected %s wallet (spv = %v)", unbip(assetID), wallet.spv)
```
itswisdomagain, line 175: Want to just rename `wallet.spv` to `wallet.created`? Upcoming dcr spv wallets via json-rpc won't have this field set to true.

client/asset/btc/spv.go
```diff
@@ -1380,6 +1389,7 @@ search:
 			continue search
 		}
 		// Pull the block.
+		w.log.Tracef("Block %v matched pkScript %v. Pulling the block...", blockHash, pkScript)
```
itswisdomagain, line 1392: Can you also correct `res.spend.blockHash -> res.spend.txHash` a few lines up
```
			w.log.Warnf("A spending input (%s) was found during the scan but the output (%s) "+
				"itself wasn't found. Was the startBlockHeight early enough?",
				newOutPoint(&res.spend.blockHash, res.spend.vin),
				newOutPoint(&txHash, vout),
			)
```

client/core/trade_simnet_test.go
```diff
@@ -383,6 +384,11 @@ func TestMakerGhostingAfterTakerRedeem(t *testing.T) {
 			client.log("tick failure: %v", err)
 		}
 
+		// Propagation to miners can take some time after the send RPC
+		// completes, especially with SPV wallets, so wait a bit before mining
+		// blocks in monitorTrackedTrade.
+		time.Sleep(sleepFactor * time.Second)
```
itswisdomagain, line 390: Does this only affect this test? I think I noticed it somewhere else and added this sleep to before the `tmuxRun` in `mineBlocks`.

chappjc: Ah, I added it in `monitorTrackedTrade` just before `mineBlocks` as well.
 https://github.com/decred/dcrdex/pull/1265/files#diff-a759d94cd368816442988c181142442065b32de45d225ac4a4da1ab9fe2ac44dR1102

So redundant.  Can also go in `mineBlocks` too suppose.  I had that at some point in the past too.

Let me know what you want to do with this PR and #788, but TBH I'd really like if 788 had anything useful from this PR.  Seems like you've addressed a number of issues with #788, and even made the maker/taker randomness a moot point with AuditContract not even searching the chain.

itswisdomagain: I can cherry-pick this commit into #788, if that works for you.

chappjc: OK.  I wasn't planning to fix the resend pending requests or order status reconciliation test in this one though, just the trade tests so that we know that btc spv is fine before we merge dcr spv.

So I'll just merge this PR and you can reshape it as you like for 788.

client/core/trade_simnet_test.go
chappjc: You have this in #788, so passing this to you.

chappjc: Oh, regarding the synced loop below, I think that applies to `created` really instead of `spv`, but it won't hurt for spv either.

Also, the sync status fix in https://github.com/decred/dcrdex/pull/1271 may also solve some issue and possibly eliminate the need for the sleep about  20 lines down from here.  We'll see when that goes in I guess.

PR #1267, opened 2021-11-03
client/asset/dcr: NewWallet must recognize the legacy type

chappjc:
The legacy wallet Type of `""` must be recognized as an RPC wallet type.


JoeGruffins (approved):
Could also do `walletTypeDcrwRPC = ""`?

chappjc:
The requirement for `WalletDefinition`s going forward is to specify a descriptive type, but we have to handle wallet config stored in the DB for wallets setup prior to this.  As in btc:
https://github.com/decred/dcrdex/blob/6c95f4fdbdd6274465fc2d2b2f2595323bd09142/client/asset/btc/btc.go#L597-L601

PR #1271, opened 2021-11-05
client/asset/btc: syncStatus considers wallet/addrmgr sync

chappjc:
There is actually a two-stage sync with btcwallet-neutrino sync.  When the chain service reaches its target height, we have to start watching the address manager's sync height.

We could just track the address manager's sync height and ignore the chain service height, but then it could appear to be stuck at zero percent for a while.  That actually might be fine as the chain service seems to sync headers consistently quickly while the address manager with cfilters fetching can do it quite slowly depending on the peer.


client/asset/btc/spv.go
```diff
@@ -268,8 +268,8 @@ func (w logWriter) Write(p []byte) (n int, err error) {
 // and does not run or connect to the RPC server.
 type spvWallet struct {
 	chainParams  *chaincfg.Params
-	wallet       btcWallet
-	cl           neutrinoService
+	wallet       *walletExtender
+	cl           *neutrino.ChainService
```
chappjc, line 272: Not to worry - I'm not proposing this as a change.  I just dev this way.

client/asset/btc/spv.go
```diff
@@ -1036,13 +1070,13 @@ func (w *spvWallet) startWallet() error {
...
 	}
-
 	chainService, err := neutrino.NewChainService(neutrino.Config{
-		DataDir:          w.netDir,
-		Database:         w.neutrinoDB,
-		ChainParams:      *w.chainParams,
+		DataDir:     w.netDir,
+		Database:    w.neutrinoDB,
+		ChainParams: *w.chainParams,
+		// AddPeers:         []string{"localhost"}, // worth a shot
```
chappjc, line 1079: Almost 4 minutes exactly with a remote mainnet peer.  This can be painful if you get a poor peer, and neutrino doesn't do multipeer, at least not for getcfilters, so I put up a node at 45.76.69.19 and got the 4 minute sync result.  I think we can hard code this as an addpeer to make it less likely a user will get a terrible sync nodes for their compact filters.

chappjc:
The sync progress indicator now reflects wallet sync instead of just the chain service, and livetest/regnet tests passing along with the other unit tests, so making r4r.

client/asset/btc/spv.go
```diff
@@ -465,28 +468,59 @@ func (w *spvWallet) syncHeight() int32 {
...
+		// With chain service synced, now track the address manager's sync.
+		walletBlock := w.wallet.syncedTo()
+		blk = &block{
+			height: int64(walletBlock.Height),
+			hash:   walletBlock.Hash,
+		}
+		currentHeight = walletBlock.Height
+		synced = currentHeight >= target // maybe && w.wallet.ChainSynced()
+		w.log.Debugf("chain = %d, wallet = %d, target = %d (synced = %v)",
+			chainBlk.Height, walletBlock.Height, target, synced)
```
chappjc, line 507: will remove

client/asset/btc/spv.go
```diff
@@ -465,28 +468,59 @@ func (w *spvWallet) syncHeight() int32 {
...
+		return &syncStatus{
+			Height:  currentHeight,
+			Syncing: true,
+		}, nil
+	}
+
+	var synced bool
+	var blk *block
+	// Wallet address manager sync height.
+	if currentHeight >= target {
```
chappjc, line 497: Note that this condition intentionally prevents `SyncStatus` from reporting `synced = true` before the wallet address manager says so.  So even if progress *may* gets close to 1 in the few seconds while the chain service syncs before consulting the address manager, it will never tell the consumer that it's done.

The ultimate solution might be to get the wallet birthday _block_ from the Wallet instance somehow, and when `w.cl.BestBlock()` is at the birthday block, switch to the address manager's status at that point.  But all I want from this fix is to only report 1.0 and synced=true when the address manager says it has reached the target.

chappjc: The db approach to pulling the birthday _block_ failed because it's not set until the wallet address manager hits the birthday, which makes perfect sense.

All we have in terms of birthday is a `time.Time`.  We should be able to compare the `BlockStamp.Timestamp` with the birthday time to know when to switch over, but the chain service's `BestBlock` does not set `Timestamp`.  This may be resolved in https://github.com/lightninglabs/neutrino/pull/235.  We should circle back on `SyncStatus` when/if that happens.

client/asset/btc/spv.go
```diff
@@ -1138,17 +1138,27 @@ func (w *spvWallet) startWallet() error {
...
-		w.connectPeers = []string{"localhost:20575"}
+	// Depending on the network, we add some addpeers or a connect peer. On
+	// regtest, if the peers haven't been explicitly set, add the simnet harness
+	// alpha node as an additional peer so we don't have to type it in. On
+	// mainet and testnet3, add a known reliable persistent peer to be used in
+	// addition to normal DNS seed-based peer discovery.
+	var addPeers []string
+	switch w.chainParams.Net {
+	case wire.MainNet:
+		addPeers = []string{"45.76.69.19"}
```
davecgh, line 1149: Would  probably be better to use a domain name here if possible in case the IP changes.

client/asset/btc/rpcclient.go
```diff
@@ -464,7 +464,7 @@ func (wc *rpcClient) syncStatus() (*syncStatus, error) {
 	return &syncStatus{
 		Target:  int32(chainInfo.Headers),
 		Height:  int32(chainInfo.Blocks),
-		Syncing: chainInfo.InitialBlockDownload || chainInfo.Headers-chainInfo.Blocks > 1,
+		Syncing: chainInfo.InitialBlockDownload || chainInfo.Headers-chainInfo.Blocks > 1, // sure not > 0?
```
davecgh, line 467: Looks like a dev comment left over.

client/asset/btc/spv.go
chappjc: Good point, plus there could be a AAAA record for ipv6.
Set to cfilters.ssgen.io

davecgh (approved):

client/asset/btc/spv.go
```diff
@@ -1039,17 +1138,27 @@ func (w *spvWallet) startWallet() error {
...
-		w.connectPeers = append(w.connectPeers, "localhost:20575")
+	// Depending on the network, we add some addpeers or a connect peer. On
+	// regtest, if the peers haven't been explicitly set, add the simnet harness
+	// alpha node as an additional peer so we don't have to type it in. On
+	// mainet and testnet3, add a known reliable persistent peer to be used in
+	// addition to normal DNS seed-based peer discovery.
+	var addPeers []string
+	switch w.chainParams.Net {
+	case wire.MainNet:
+		addPeers = []string{"cfilters.ssgen.io"}
```
buck54321, line 1149: I wonder whether this could result in bottlenecks. What are the chances that this node is picked for the filter source? Is it random, or since this is in addpeer, does it result in this node being chosen more often or even always? Hard-coding a mainnet node also feels pretty icky, though I admit I've considered it as a solution for various problems. I would hope that if we do have "sanctioned" nodes, we would make them opt-out at the very least.

client/asset/btc/spv.go
```diff
@@ -433,14 +433,17 @@ func (w *spvWallet) getBlockHeight(h *chainhash.Hash) (int32, error) {
...
-		return nil, err
-	}
-	return &blk.Hash, err
+	blk := w.wallet.syncedTo()
+	return &blk.Hash, nil
 }
 
 func (w *spvWallet) getBestBlockHeight() (int32, error) {
+	return w.wallet.syncedTo().Height, nil
+}
```
buck54321, line 442: Please doc `getBestBlockHeight`, esp. in contrast with `getChainHeight`.

buck54321 (changes requested):

client/asset/btc/spv.go
chappjc: Good questions.  The "addpeer" nodes are only suggestions, so they get mixed into the other nodes from DNS seeders.  I had the same icky feeling at first, but when you consider we are told what nodes to use in part from the Bitcoin Core devs, some of whom were opposed to compact filters, it's quite fair that we at least throw one in the mix.

https://github.com/btcsuite/btcd/blob/7b6c2b342379f3cba880d9f3cc14c77fc7500969/chaincfg/params.go#L261-L266

I'm not sure opting out provides much benefit considering that we cannot opt out of the DNS seed nodes either.  This could be prevented however if we expose the `connect` peers, which are exclusive and prevent discovery.

Regarding if the addpeer takes some priority or gets chosen more readily, I cannot tell for sure, but in my observations it does seem to take some precedence.

buck54321 (approved):

chappjc:
For posterity, here's a command I use to tail the neutrino+wallet log for the most informative sync info:

`tail -f assetdb/btc/mainnet/logs/neutrino.log | grep --line-buffered -iE "Processed|Scanning|Recover|starting recovery|syncing|finished|rescanned|catching up|subscri|ERR|Catching up"`

PR #1272, opened 2021-11-05
admin: Remove ban and unban.

### LINKED ISSUE #1214: auth/admin: banning by admin API ineffective
The admin API **/account/{accountID}/ban?rule=RULE** = `(*Server).apiBan` doesn't really work, because the `(*AuthManager).handleConnect` method will just [unban the user based on score](https://github.com/decred/dcrdex/blob/ad4321a869156c5340fb698fcdd828b751349117/server/auth/auth.go#L1193-L1196). Perhaps we need a new "violation" with a penalty of `math.MaxInt32` to re-enable administrative banning of users.

JoeGruffins:
The api ban and unban endpoints do not change the user's ban score. The
    server then bans or unbans upon connect based on the user's ban score.
    This makes the api enpoints ineffective. Remove those api's until and if
    an alternative can be discussed and implemented.

closes #1214


chappjc:
Looks fine, but will test.  Probably best to leave the auth manager code alone as you have.

JoeGruffins:
>  Probably best to leave the auth manager code alone as you have.

If I understand correctly, yeah. the Penalize and Unban methods still work as intended as long as they happen as a result of calculating the ban score. 

Thinking a bit more. Penalize will tell the user what happened, so it has a place. But I don't know how useful the account rule is? We could just go off ban score, but I guess that's a different issue.

chappjc (approved):

Issue #1235, opened 2021-10-19
eth: Move server/eth/common.go to dex/networks/eth.


JoeGruffins:
Because the client also uses these values, it makes more sense for them to be in dex/networks.

There is a file in dex/networks called contract.go that is automatically generated from the `.solc` file in the `contracts` folder there. All of that should point to a new folder/package, probably named `swap` to avoid naming conflicts that may arise when auto-generating. The `contracts` folder can probably just be moved to the new `swap` folder without changing anything there. There may be random tests looking for the `contracts` folder that need to be tweaked as well, although I can't find any at the moment, so maybe not. A lot of imports will need to be adjusted. internal/eth/reentryattack/README.md should be edited to include the new location.

JoeGruffins:
I'll do this.

PR #1273, opened 2021-11-05
server/eth: Implement ValidateContract.

JoeGruffins:
server/eth: Implement ValidateContract.

    Require the the Contract field of msgjson.Init be filled with the txdata
    used to create the swap. Validate that txdata to be of the correct
    format sending the expected types of arguments.

work towards #1154 depends on #1235

### DIFF
--- a/server/asset/eth/common.go
+++ b/server/asset/eth/common.go
@@ -59,6 +59,9 @@ const (
 	// which holds the ETH, an amount of ETH in gwei, and a random nonce to
 	// avoid duplicate coin IDs.
 	CIDAmount
+	// SecretHashSize is the byte-length of the hash of the secret key used
+	// in swaps.
+	SecretHashSize = 32
 )
 
 // CoinID is an interface that objects which represent different types of ETH
--- a/server/asset/eth/eth.go
+++ b/server/asset/eth/eth.go
@@ -18,7 +18,7 @@ import (
 
 	"decred.org/dcrdex/dex"
 	"decred.org/dcrdex/dex/encode"
-	swap "decred.org/dcrdex/dex/networks/eth"
+	dexeth "decred.org/dcrdex/dex/networks/eth"
 	"decred.org/dcrdex/server/asset"
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/common"
@@ -74,7 +74,7 @@ func (d *Driver) DecodeCoinID(coinID []byte) (string, error) {
 
 // UnitInfo returns the dex.UnitInfo for the asset.
 func (d *Driver) UnitInfo() dex.UnitInfo {
-	return swap.UnitInfo
+	return dexeth.UnitInfo
 }
 
 // ethFetcher represents a blockchain information fetcher. In practice, it is
@@ -94,7 +94,7 @@ type ethFetcher interface {
 	syncProgress(ctx context.Context) (*ethereum.SyncProgress, error)
 	blockNumber(ctx context.Context) (uint64, error)
 	peers(ctx context.Context) ([]*p2p.PeerInfo, error)
-	swap(ctx context.Context, secretHash [32]byte) (*swap.ETHSwapSwap, error)
+	swap(ctx context.Context, secretHash [32]byte) (*dexeth.ETHSwapSwap, error)
 	transaction(ctx context.Context, hash common.Hash) (tx *types.Transaction, isMempool bool, err error)
 }
 
@@ -346,9 +346,13 @@ func (eth *Backend) ValidateCoinID(coinID []byte) (string, error) {
 }
 
 // ValidateContract ensures that the swap contract is constructed properly, and
-// contains valid sender and receiver addresses.
-func (eth *Backend) ValidateContract(contract []byte) error {
-	return notImplementedErr
+// contains valid counterparty, secret hash, and locktime.
+func (eth *Backend) ValidateContract(txdata []byte) error {
+	_, err := dexeth.ParseInitiateData(txdata)
+	if err != nil {
+		return fmt.Errorf("unable to parse contract txdata: %w", err)
+	}
+	return nil
 }
 
 // CheckAddress checks that the given address is parseable.
--- a/server/asset/eth/eth.go
+++ b/server/asset/eth/eth.go
@@ -345,12 +345,10 @@ func (eth *Backend) ValidateCoinID(coinID []byte) (string, error) {
 	return coinId.String(), nil
 }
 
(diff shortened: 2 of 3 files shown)


server/asset/eth/coiner.go
```diff
@@ -284,3 +286,203 @@ func (c *swapCoin) Value() uint64 {
...
+// FeeRate returns the lowest gas fee among pending transactions to the
+// account. It is not set if there was not pending balance upon initiation of
+// the funding coin.
+func (c *amountCoin) FeeRate() uint64 {
+	return c.gasPrice
+}
+
+// Auth ensures that a user really has the keys to spend a funding coin by
+// verifying a message signed by the funding coin address.
+func (c *amountCoin) Auth(pubkeys, sigs [][]byte, msg []byte) error {
```
buck54321, line 450: I guess we'll still have to figure out when and how we confirm ownership of the account, but it won't be via coin like this anymore.

server/asset/eth/eth.go
```diff
@@ -332,9 +336,13 @@ func (eth *Backend) ValidateCoinID(coinID []byte) (string, error) {
...
-func (eth *Backend) ValidateContract(contract []byte) error {
-	return notImplementedErr
+// contains valid counterparty, secret hash, and locktime.
+func (eth *Backend) ValidateContract(txdata []byte) error {
+	_, _, _, err := dexeth.ParseInitiateData(txdata)
+	if err != nil {
+		return fmt.Errorf("unable to parse contract txdata: %w", err)
+	}
+	return nil
 }
```
buck54321, line 346: I think this is the essential part of this PR after the changes re #1154 / #1243.

server/asset/eth/coiner.go
JoeGruffins: Yeah, this pr is only the last commit, so I'll rebase and remove that.

martonp:
Why did you choose the contract to be the txdata? Now that we are doing batch initiations, the contract will represent multiple swaps.

chappjc:
> Why did you choose the contract to be the txdata? Now that we are doing batch initiations, the contract will represent multiple swaps.

I'd like to get on the same page here, so to provide context for discussion:

- server gets init message with a coinID []byte and a contract []byte
- performs basic validation of the coinID and contract.  For the contract, using DCR as an example, this means decoding the script to ensure it's the right form of contract to perform a swap, not even checking the amounts, address, locktime, etc.  For ETH, I believe we would just need to ensure that a known deployed swap contract is being used.  Depending on what the ETH coinID encodes now, this may be a no-op.

```go
func (dcr *Backend) ValidateContract(contract []byte) error {
	_, _, _, _, err := dexdcr.ExtractSwapDetails(contract, chainParams)
	return err
}
```

- after basic validation, server starts searching for the transaction on the network to actually check the swap details because `ValidateContract` does not do much here, using the `Contract(coinID, contract)` method
- the result of finding it is the swap address, lockTime, and amount of the particular swap (which keyed by secret hash, wherever we are encoding that with ETH now).  For ETH, I'm not yet certain if this data should come out of the transaction call data, or wait until the contract account's **state** reflects the new init.  In any case, if `txData` does have a role server-side (besides relaying that to the counter-party, I do not think it would come at the `ValidateContract` stage, but `Contract`. To me it looks like `ValidateContract` is just to check that the intended contract address is one of the ones known to the server.

chappjc:
So, please bring me up to speed: what is the current coinID encoding and contract bytes **in the `init` request**?  All that needs to be conveyed I think is: 1) the swap's unique secret hash, and 2) the intended swap contract address.
The `txData` comes from the returned `contract` object from `chain.Contract(params.CoinID, params.Contract)`, so it's definitely not going to be around for the preliminary `ValidateContract` call.

buck54321:
@chappjc see also https://github.com/decred/dcrdex/issues/1154#issuecomment-969330835

I recommend that for `Init` we send only secret hash = contract and txid = coin ID. Don't need an index, since we have the secret hash, I think. I made a comment [here](https://github.com/decred/dcrdex/pull/1307#discussion_r752623545) too, but I don't think we need any special encoding for "coin ID" = txid.

With that scheme, the only thing `ValidateContract` needs to do for ETH is to check the length of the secret hash.

chappjc:
In that comment, @buck54321:

> 2 I'm going to propose something different for contract address validation, but if we do want to validate the contract address, it'll need to be done earlier than this anyway, so let's not plan to send it here.

What do you mean by "earlier than this"?  Init is the start of the swap, the contract address that's targeted for the swap would need to come here.  Do you intend to pick it out of the transaction data after retrieving it via the txid?  This would come in processInit from the `Contract` method.

buck54321:
> What do you mean by "earlier than this"? Init is the start of the swap, so the contract address that's targeted for the swap would need to come here.

We shouldn't even place an order on a server using an unsupported contract address. #1301 links the contract address to the asset version.

chappjc:
Yes, of course. But this is server-side validation.  In these server checks, the contract address has to be checked.  I think the right place to get that is from the tx data retrieved by the txid, but I want to understand your thinking on this.

buck54321:
> Presumably the Contract method that tries to get all the swap data like counterparty address, swap amount, and locktime, will have to get that data from the transaction it gets via txid, but also validate the effect on the contract state at some point.

Right. And the contract address doesn't need be decoded from the txdata. It should be the **to** address of the transaction.

chappjc:
Great, we are 100% on the same page now.  The contract address is validated when `Contract` verifies the txn in fact interacts with the right contract (the to address as you point out).
I also agree that `ValidateAddress` can do no more than check the length of the secret hash.  Anything else has to come out of the transaction and its call data located on-network, or the effects of the transaction on the contract state.

server/asset/eth/common.go
```diff
@@ -59,6 +59,9 @@ const (
 	// which holds the ETH, an amount of ETH in gwei, and a random nonce to
 	// avoid duplicate coin IDs.
 	CIDAmount
+	// SecretHashSize is the byte-length of the hash of the secret key used
+	// in swaps.
+	SecretHashSize = 32
```
JoeGruffins, line 64: Will be moved to dex/networks

JoeGruffins:
> The txData comes from the returned contract object from chain.Contract(params.CoinID, params.Contract), so it's definitely not going to be around for the preliminary ValidateContract call.

Whoops.

> Now that we are doing batch initiations, the contract will represent multiple swaps.

Also whoops, but I guess we don't have that data yet anyway.

martonp (approved):
Looks good.

chappjc (approved):

PR #1274, opened 2021-11-07
eth: Batchify solidity redeem function

martonp:
This PR updates the redeem function in swap contract to be able to process multiple redemptions in a single call.


JoeGruffins (approved):
Was it the server want's all the redeems at once was the reason for batching?

Otherwise I wouldn't think there's a great reason to batchify the redeems, because I wouldn't expect them to happen in the same trade.tick() like maybe init would. (Is this where they get batched?) Also it's much more important that redeem not fail, so forcing one at a time maybe be more prudent, but I'm not sure.

But other than that, the changes look correct for what it's doing and mirror the init changes so approving.

martonp:
I did it because the current design of the client expects the wallets to be able to batch redeem.

chappjc:
@JoeGruffins' comment:

> Was it the server want's all the redeems at once was the reason for batching?

No, in fact the server requires N `'redeem'` requests for each swap redeemed.

@martonp's comment:

> I did it because the current design of the client expects the wallets to be able to batch redeem.

While the `client/asset.Wallet.Redeem` method does receive a slice of `client/asset.Redemption`s, the only reason the implementation must do them in a single transaction presently is because of the return `out Coin`, which is not a slice.  I do agree it's easiest to keep this and the calling code the same, and instead implement a batch redeem method of the contract, but it could be done with multiple transactions instead or the ETH implementation could return an error for more than one redemption, which presently causes `Core` to "ungroup" the redeems.  Not pretty, but just pointing out it's not set in stone that the ETH contract must permit this.

Back to @JoeGruffins' comment:

> Otherwise I wouldn't think there's a great reason to batchify the redeems, because I wouldn't expect them to happen in the same trade.tick() like maybe init would. (Is this where they get batched?) Also it's much more important that redeem not fail, so forcing one at a time maybe be more prudent, but I'm not sure.

It is true that batched redeem transactions are extremely rare in practice.  I'm not saying we should reject this change that makes batch `redeem` contract method, but in almost all cases it will be a waste of 2k gas.

Maker init's are the only reliable time when batching of any swap actions happens because the `'match'` route payload is a `[]msgjson.Match`, a slice, which allows the inits for a single trade to be batched via `(*trackedTrade).negotiate([]*msgjson.Match)`.  This does _not_ bundle matches from different orders however.

Redeems happen:
- For **m**aker, when the taker's init txn reaches the target confirmations (a.k.a. block-based).  Even if those counterparty txns were broadcast at the exact same time, which they are often not, there is no guarantee they get mined in the same block.  Even when they belong to the same order, and matches made in the same epoch, they often do not reach this point at the same block.
- For **t**aker, when the server notifies them of maker's redeem (a.k.a. event-based).  See `handleRedemptionRoute`.  These events are even less synchronized because there are different messages and likely different counterparties, all timed independently.  We make some effort to delay the redeem action in response to these messages to allow more to be batched (see `schedTradeTick`), but that has only occasionally been successful at facilitating batch redeems as taker.

This is all just for background and is not an argument against batchifying the contract's redeem function.

dex/networks/eth/contracts/ETHSwapV0.sol
```diff
@@ -151,15 +145,26 @@ contract ETHSwap {
...
-        swaps[secretHash].state = State.Redeemed;
-        (bool ok, ) = payable(msg.sender).call{value: swaps[secretHash].value}("");
+        uint amountToRedeem = 0;
+        for (uint i = 0; i < redemptions.length; i++) {
+            Redemption calldata redemption = redemptions[i];
+            Swap storage swapToRedeem = swaps[redemption.secretHash];
+
+            require(swapToRedeem.state == State.Filled);
+            require(swapToRedeem.participant == msg.sender);
+            require(sha256(abi.encodePacked(redemption.secret)) == redemption.secretHash);
```
chappjc, line 159: What would be the overhead if any of calling a public function isRedeemable instead?

dex/networks/eth/txdata.go
```diff
@@ -62,32 +62,33 @@ func ParseInitiateData(calldata []byte) ([]ETHSwapInitiation, error) {
...
 	// TODO: If any of the checks prove redundant, remove them.
 	if len(args) != numRedeemArgs {
-		return fail(fmt.Errorf("expected %v redeem args but got %v", numRedeemArgs, len(args)))
+		return nil, fmt.Errorf("expected %v redeem args but got %v", numRedeemArgs, len(args))
 	}
-	secret, ok := args[0].value.([32]byte)
+	redemptions, ok := args[0].value.([]struct {
+		Secret     [32]byte `json:"secret"`
+		SecretHash [32]byte `json:"secretHash"`
+	})
```
chappjc, line 84: It super sucks that `ETHSwapRedemption` (generated code) does not tag the struct fields.
I wonder if we should make an issue with geth.  Maybe they'd tell us to do something differently.

chappjc (changes requested):
Looks good @martonp, thanks.

I'd just like to hear your thoughts about keeping `isRedeemable` around as a read only public function.  It would only affect deployment gas cost I believe, and that's not much of an issue.  But I do see it being important for clients to use before attempting to redeem and revealing their secret key in the call data.  Plus if it adds no gas to use from the loop in `redeem` that would be a consistency bonus.

dex/networks/eth/contracts/ETHSwapV0.sol
JoeGruffins: I think it is advised not to call functions from functions, except in this case where functions have the `modifier` tag/type, because that may lead to unexpected exploits.

dex/networks/eth/txdata.go
JoeGruffins: The code we are using to examine the txdata is not even exported. I doubt they want us to do this. We are supposed to just use the contract I believe.

dex/networks/eth/contracts/ETHSwapV0.sol
```diff
@@ -60,17 +60,6 @@ contract ETHSwap {
         require(block.timestamp >= refundBlockTimestamp);
         _;
     }
-
-    // isRedeemable checks that a swap can be redeemed. The requirements are
-    // the participant is msg.sender, the state is Filled, and the passed secret
-    // hashes to secretHash.
-    modifier isRedeemable(bytes32 secretHash, bytes32 secret) {
```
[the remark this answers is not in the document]
JoeGruffins: The swaps map is public, so I think it is possible to determine redeemability, but an exported public function is still fine to add imo. Maybe better because it will use solidity's logic, which may be different somehow than what we expect.

dex/networks/eth/contracts/ETHSwapV0.sol
JoeGruffins: So, can't find where I read that and don't see it here https://consensys.github.io/smart-contract-best-practices/recommendations/ so maybe fine.

martonp: There's about a 200 gas overhead to calling a separate function. 

@JoeGruffins Maybe what you read was not to call other public non-view functions. I don't think a function that doesn't update state would cause a vulnerability.

dex/networks/eth/contracts/ETHSwapV0.sol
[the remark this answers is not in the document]
martonp: I agree with @JoeGruffins, we could just look into the swaps map. Adding the the isRedeemable function would cost about an extra $25 when deploying the contract if gas price is 100 gwei, so I guess it's not too much for the added convenience.

martonp:
I'm surprised that batched redemptions don't happen more frequently, especially with DCR and BTC which have longer block times. If two transactions with the same tx fee are submitted almost simultaneously, the chances of them being included in the same block are really that small? With ETH it will probably be more rare due to the shorter block time.

chappjc:
> I'm surprised that batched redemptions don't happen more frequently, especially with DCR and BTC which have longer block times. If two transactions with the same tx fee are submitted almost simultaneously, the chances of them being included in the same block are really that small? With ETH it will probably be more rare due to the shorter block time.

It's extremely rare in practice. It also doesn't help that any batched redeems currently must belong to the same order too.

dex/networks/eth/contracts/ETHSwapV0.sol
chappjc: Skip the overhead then.  Not worth it.

dex/networks/eth/txdata.go
chappjc: I think they would say that too.  My note about a real audit involving isRedeemable before attempting the redeem is a required step we will need to take. https://github.com/decred/dcrdex/pull/1274#discussion_r754401945

chappjc (approved):

PR #1277, opened 2021-11-09
client/asset: Use max sequence only for refunds.

JoeGruffins:
Both bitcoin and decred do not need the sequence to be set to one less
    than max in order to redeem, only to refund. Remove the unneeded
    assignments and move the doc. Point decred's doc at decred material.

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -1688,9 +1688,6 @@ func (btc *ExchangeWallet) Redeem(form *asset.RedeemForm) ([]dex.Bytes, asset.Co
 		addresses = append(addresses, receiver)
 		contracts = append(contracts, contract)
 		txIn := wire.NewTxIn(cinfo.output.wireOutPoint(), nil, nil)
-		// Enable locktime
-		// https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki#Spending_wallet_policy
-		txIn.Sequence = wire.MaxTxInSequenceNum - 1
 		msgTx.AddTxIn(txIn)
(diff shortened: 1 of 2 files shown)


JoeGruffins:
I apologize that this doesn't really do much, but I think this is correct. I also cannot find a real doc to point decred at, could keep it btc I guess.

chappjc:
No problem.  Can we corroborate with the old atomicswap repo?
We should also finish smoothing out the trade_simnet_tests first now that both dcr and btc spv work is merged, but this may be right to merge

JoeGruffins:
> Can we corroborate with the old atomicswap repo?

Where was that again?

chappjc:
This stuff: https://github.com/decred/atomicswap/blob/master/cmd/dcratomicswap/main.go

JoeGruffins:
It doesn't look like it sets this for redeem either. I think that tx creation starts here https://github.com/decred/atomicswap/blob/7c6928f1d1a8cd6ebe4bb6e0a0c907d423f75e13/cmd/dcratomicswap/main.go#L818

For refund it is set to 0. So, I guess that's fine? Still a little confused about the sequence, because if zero it is not disabled. Is it only considered if used in conjunction with the OP_CHECKSEQUENCEVERIFY opcode?

JoeGruffins:
I see what I was missing now. We are exclusively using version 1 transactions, so the sequence locks don't apply to us, for decred at least.

buck54321 (approved):

chappjc (approved):
LGTM.  Refund code is untouched, only redeem, so there's no hiding edge cases here.

PR #1278, opened 2021-11-09
client/asset: clean up Wallet interfaces

buck54321:
**btc**: Instead of passing around a `RawRequester`, just return the
`rpcclient.Client` from the (renamed) `RPCCloneWallet` constructor.

Allow registration of external wallets using the same method as dcr,
which is to register a constructor alongside a `WalletDefinition`. Some
internal reorganization to accommodate. Adds an exported helper method
for constructing an SPV based wallet.

**dcr**: Change `Wallet` methods to use lower-level, common types instead
of json types, e.g. `wire.BlockHeader` instead of `chainjson.GetBlockHeaderVerboseResult`
and `secp256k1.PrivateKey` instead of `btcutil.WIF`.

Remove account argument from all methods and 'Account' from all method
names. The account is already passed into the constructor. In the same
vein, some rpc implementation-specific error handling is move to the
appropriate `rpcWallet` methods, such as `signrawtransaction` errors.

### DIFF
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -23,14 +23,12 @@ import (
 	"decred.org/dcrdex/dex"
 	"decred.org/dcrdex/dex/calc"
 	dexdcr "decred.org/dcrdex/dex/networks/dcr"
-	"decred.org/dcrwallet/v2/rpc/client/dcrwallet"
 	walletjson "decred.org/dcrwallet/v2/rpc/jsonrpc/types"
 	"github.com/btcsuite/btcwallet/wallet"
 	"github.com/decred/dcrd/blockchain/stake/v4"
 	"github.com/decred/dcrd/blockchain/v4"
 	"github.com/decred/dcrd/chaincfg/chainhash"
 	"github.com/decred/dcrd/chaincfg/v3"
-	"github.com/decred/dcrd/dcrec/secp256k1/v4"
 	"github.com/decred/dcrd/dcrec/secp256k1/v4/ecdsa"
 	"github.com/decred/dcrd/dcrutil/v4"
 	chainjson "github.com/decred/dcrd/rpc/jsonrpc/types/v3"
@@ -367,7 +365,6 @@ type ExchangeWallet struct {
 	wallet           Wallet
 	chainParams      *chaincfg.Params
 	log              dex.Logger
-	acct             string
 	tipChange        func(error)
 	fallbackFeeRate  uint64
 	feeRateLimit     uint64
@@ -482,7 +479,6 @@ func unconnectedWallet(cfg *asset.WalletConfig, dcrCfg *Config, chainParams *cha
 	return &ExchangeWallet{
 		log:                 logger,
 		chainParams:         chainParams,
-		acct:                dcrCfg.Account,
 		tipChange:           cfg.TipChange,
 		fundingCoins:        make(map[outPoint]*fundingCoin),
 		findRedemptionQueue: make(map[outPoint]*findRedemptionReq),
@@ -533,14 +529,6 @@ func (dcr *ExchangeWallet) Connect(ctx context.Context) (*sync.WaitGroup, error)
 		}
 	}()
 
-	curnet, err := dcr.wallet.Network(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("unable to fetch wallet network: %w", err)
-	}
-	if curnet != dcr.chainParams.Net {
-		return nil, fmt.Errorf("unexpected wallet network %s, expected %s", curnet, dcr.chainParams.Net)
-	}
-
 	// Initialize the best block.
 	dcr.tipMtx.Lock()
 	dcr.currentTip, err = dcr.getBestBlock(ctx)
@@ -571,7 +559,11 @@ func (dcr *ExchangeWallet) Connect(ctx context.Context) (*sync.WaitGroup, error)
 
 // OwnsAddress indicates if an address belongs to the wallet.
 func (dcr *ExchangeWallet) OwnsAddress(address string) (bool, error) {
-	return dcr.wallet.AccountOwnsAddress(dcr.ctx, dcr.acct, address)
+	addr, err := stdaddr.DecodeAddress(address, dcr.chainParams)
+	if err != nil {
+		return false, err
+	}
+	return dcr.wallet.OwnsAddress(dcr.ctx, addr)
 }
 
 // Balance should return the total available funds in the wallet. Note that
@@ -584,7 +576,7 @@ func (dcr *ExchangeWallet) Balance() (*asset.Balance, error) {
 	if err != nil {
 		return nil, err
 	}
-	ab, err := dcr.wallet.AccountBalance(dcr.ctx, dcr.acct, 0)
+	ab, err := dcr.wallet.Balance(dcr.ctx, 0)
 	if err != nil {
 		return nil, err
 	}
@@ -598,11 +590,15 @@ func (dcr *ExchangeWallet) Balance() (*asset.Balance, error) {
 
 // FeeRate returns the current optimal fee rate in atoms / byte.
 func (dcr *ExchangeWallet) feeRate(confTarget uint64) (uint64, error) {
+	estimator, is := dcr.wallet.(FeeRateEstimator)
+	if !is {
+		return 0, fmt.Errorf("fee rate estimation unavailable")
+	}
 	// estimatesmartfee 1 returns extremely high rates on DCR.
 	if confTarget < 2 {
 		confTarget = 2
 	}
-	estimatedFeeRate, err := dcr.wallet.EstimateSmartFeeRate(dcr.ctx, int64(confTarget), chainjson.EstimateSmartFeeConservative)
+	estimatedFeeRate, err := estimator.EstimateSmartFeeRate(dcr.ctx, int64(confTarget), chainjson.EstimateSmartFeeConservative)
 	if err != nil {
 		return 0, err
 	}
@@ -881,12 +877,12 @@ func (dcr *ExchangeWallet) fund(enough func(sum uint64, size uint32, unspent *co
 
 // spendableUTXOs generates a slice of spendable *compositeUTXO.
 func (dcr *ExchangeWallet) spendableUTXOs() ([]*compositeUTXO, error) {
-	unspents, err := dcr.wallet.Unspents(dcr.ctx, dcr.acct)
+	unspents, err := dcr.wallet.Unspents(dcr.ctx)
 	if err != nil {
 		return nil, err
 	}
 	if len(unspents) == 0 {
-		return nil, fmt.Errorf("insufficient funds. 0 DCR available to spend in %q account", dcr.acct)
+		return nil, fmt.Errorf("insufficient funds. 0 DCR available to spend")
 	}
 
 	// Parse utxos to include script size for spending input.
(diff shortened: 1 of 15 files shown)


client/asset/dcr/wallet.go
```diff
@@ -69,19 +62,14 @@ type Wallet interface {
...
-	// LockedOutputs fetches locked outputs for the specified account.
-	LockedOutputs(ctx context.Context, account string) ([]chainjson.TransactionInput, error)
-	// EstimateSmartFeeRate returns a smart feerate estimate.
-	EstimateSmartFeeRate(ctx context.Context, confTarget int64, mode chainjson.EstimateSmartFeeMode) (float64, error)
-	// Unspents fetches unspent outputs for the specified account.
-	Unspents(ctx context.Context, account string) ([]walletjson.ListUnspentResult, error)
-	// GetChangeAddress returns a change address from the specified account.
-	GetChangeAddress(ctx context.Context, account string) (stdaddr.Address, error)
+	// OwnsAddress checks if the provided address belongs to the Wallet.
+	OwnsAddress(ctx context.Context, address string) (bool, error)
```
buck54321, line 66: Also changing this argument to an `stdaddr.Address`.

client/asset/interface.go
```diff
@@ -42,6 +42,9 @@ type WalletDefinition struct {
 	// description for each option. This can be used to request config info from
 	// users e.g. via dynamically generated GUI forms.
 	ConfigOpts []*ConfigOption `json:"configopts"`
+	// Managed can be set to true to disable wallet re-configuration through the
+	// standard UI.
+	Managed bool `json:"managed"`
```
buck54321, line 47: Any objections?

chappjc: Can you expand on the comment and the function of this bool more please?  I don't see how you intend to use this based on the diff in this PR.

chappjc:
My current thinking on this is to branch 0.4 _before_ merging this. Considerations: we have working btc and dcr client SPV-enable packages on master, we urgently need to branch and finalize 0.4, and I haven't identified a powerful reason to include this refactor for 0.4 that outweighs the risks of bugs.

client/asset/interface.go
buck54321: I'll probably remove this until I get the details worked out, but the idea is that if we're providing an external wallet, we already have a user interface and a way to configure the wallet, and having a separate interface to configure the wallet would be somewhat confusing and difficult to program around, for example if I needed to know when the configuration is updated for some external accounting purposes.

client/asset/dcr/dcr.go
```diff
@@ -2157,85 +2180,40 @@ func (dcr *ExchangeWallet) refundTx(coinID, contract dex.Bytes, val uint64, refu
...
+	return dcr.wallet.Lock(dcr.ctx)
 }
 
 // Locked will be true if the wallet is currently locked.
 // Q: why are we ignoring RPC errors in this?
 func (dcr *ExchangeWallet) Locked() bool {
-	// First return locked status of the account, falling back to walletinfo if
-	// the account is not individually password protected.
-	encrypted, unlocked, err := dcr.accountUnlocked(dcr.ctx, dcr.acct)
+	unlocked, err := dcr.wallet.Unlocked(dcr.ctx)
```
itswisdomagain, line 2211: This makes it the responsibility of the `Wallet` impl to check if the account is encrypted and perform an account unlocking rather than a wallet unlocking but the method says `wallet.Unlock`. At the very least, the doc for `wallet.Unlock` should mention this. But I think it's better to keep the account and wallet lock/unlock/locked methods separate. The `Wallet` interface does not represent an `ExchangeWallet` but a real dcr wallet, the semantics should resemble that of a real dcr wallet while the `ExchangeWallet` makes use of the actual wallet methods as needed, in this case, checking account lock status before trying to perform wallet unlocking.

client/asset/dcr/dcr.go
```diff
@@ -2598,7 +2543,7 @@ func (dcr *ExchangeWallet) signTxAndAddChange(baseTx *wire.MsgTx, feeRate uint64
 	// Sign the transaction to get an initial size estimate and calculate
 	// whether a change output would be dust.
 	sigCycles := 1
-	msgTx, err := dcr.signTx(baseTx)
+	msgTx, err := dcr.wallet.SignRawTransaction(dcr.ctx, baseTx)
```
itswisdomagain, line 2546: Tried this before, the `baseTx` ends up getting mutated and affects subsequent calls to srt from the loop below. Since the rpc only needs an hex string, can pass that around and have other implementations do with that as needed.

client/asset/dcr/dcr.go
```diff
@@ -2912,6 +2833,19 @@ func (dcr *ExchangeWallet) handleTipChange(newTipHash *chainhash.Hash, newTipHei
 	go dcr.findRedemptionsInBlockRange(startHeight, newTipHeight, contractOutpoints)
 }
 
+func (dcr *ExchangeWallet) getMainchainHeader(blockHash *chainhash.Hash) (*wire.BlockHeader, bool, error) {
```
itswisdomagain, line 2836: The header returned isn't necessarily a mainchain header. Method name might be confusing.

client/asset/dcr/dcr.go
```diff
@@ -2925,50 +2859,19 @@ func (dcr *ExchangeWallet) getBestBlock(ctx context.Context) (*block, error) {
...
-	if err != nil {
-		return false, fmt.Errorf("block %s has invalid nexthash value %s: %v",
-			block.hash, blockHeader.NextHash, err)
-	}
-	nextBlockHeader, err := dcr.wallet.GetBlockHeaderVerbose(ctx, nextBlockHash)
-	if err != nil {
-		return false, fmt.Errorf("getblockheader error for block %s: %w", nextBlockHash, err)
+		checkHash = &checkBlock.PrevBlock
 	}
-	validated := nextBlockHeader.VoteBits&1 != 0
```
itswisdomagain, line 2970: To reduce the possibility of alternative `Wallet` implementations bombing themselves, we should include in the `IsMainchain` doc the need to check the next header vote bits.

itswisdomagain (changes requested):
Preliminary review. Only looked at the dcr changes for now.

client/asset/dcr/dcr.go
```diff
@@ -584,7 +576,7 @@ func (dcr *ExchangeWallet) Balance() (*asset.Balance, error) {
 	if err != nil {
 		return nil, err
 	}
-	ab, err := dcr.wallet.AccountBalance(dcr.ctx, dcr.acct, 0)
+	ab, err := dcr.wallet.Balance(dcr.ctx, 0)
```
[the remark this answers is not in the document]
buck54321: I'm struggling to imagine a scenario when the `ExchangeWallet` would need to know the account number because the custom `Wallet` doesn't. As an example, if we ever implement a privacy `Wallet`, will we need to re-write `ExchangeWallet` to be aware of mixed and unmixed account? These are implementation details. Let's leave `ExchangeWallet` out of it. 

> Decred wallets differentiate wallet balance from account balance, so wallet.Balance reads as though it's fetching the wallet balance

That's an implementation detail that the `ExchangeWallet` doesn't need to care about.

> Also, I don't other wallet implementations having to figure out when to use the configured account,

I don't quite follow. What other account would the wallet implementation use besides the configured one, and when would the `ExchangeWallet` know it, but not the `Wallet`?

> Also covers cases where the ExchangeWallet is reconfigured and the account changes

New `ExchangeWallet` is created on reconfiguration anyway.

client/asset/dcr/dcr.go
```diff
@@ -1253,7 +1249,7 @@ func (dcr *ExchangeWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin
 		totalOut += contract.Value
 		// revokeAddrV2 is the address belonging to the key that will be
 		// used to sign and refund a swap past its encoded refund locktime.
-		revokeAddrV2, err := dcr.wallet.GetNewAddressGapPolicy(dcr.ctx, dcr.acct, dcrwallet.GapPolicyIgnore)
+		revokeAddrV2, err := dcr.wallet.ExternalAddress(dcr.ctx)
```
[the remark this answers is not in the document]
buck54321: If I'm implementing an external custom `Wallet`, I probably already have a gap policy.

client/asset/dcr/dcr.go
```diff
@@ -1937,9 +1964,7 @@ rangeBlocks:
 
 		lastScannedBlockHeight = blockHeight
 		scanPoint := fmt.Sprintf("block %d", blockHeight)
-		blkTxs := append(blk.RawTx, blk.RawSTx...)
-		for t := range blkTxs {
-			tx := &blkTxs[t]
+		for _, tx := range blk.Transactions {
```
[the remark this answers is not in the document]
buck54321: Whoops. Forgot what asset I'm working on.

client/asset/dcr/dcr.go
```diff
@@ -1674,14 +1670,41 @@ func (dcr *ExchangeWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time,
...
-		return false, time.Time{}, fmt.Errorf("unable to retrieve block header: %w", err)
+		return false, time.Time{}, fmt.Errorf("unable to calculate median time: %w", err)
 	}
-	medianTime := time.Unix(blockHeader.MedianTime, 0)
-	return medianTime.After(contractExpiry), contractExpiry, nil
+	return time.Unix(medianTimeStamp, 0).After(contractExpiry), contractExpiry, nil
+}
+
+// medianTime calculates the median time of the previous few blocks
+// prior to, and including, the block node.
```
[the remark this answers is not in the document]
buck54321: >  but we can wrap *wire.BlockHeader in a new struct that also provides the median time and anything else we may need

Will do.

client/asset/dcr/dcr.go
buck54321: Account vs wallet unlocking is implemented by dcrwallet and is a new feature. I definitely wouldn't assume that all wallets will differentiate. Regardless, this is the most general form, so should be suitable for any implementation. What happens under the hood is up to the `Wallet`.

client/asset/dcr/dcr.go
```diff
@@ -1477,12 +1473,12 @@ func (dcr *ExchangeWallet) SignMessage(coin asset.Coin, msg dex.Bytes) (pubkeys,
 	if err != nil {
 		return nil, nil, fmt.Errorf("error decoding address: %w", err)
 	}
-	priv, pub, err := dcr.getKeys(address)
+	priv, err := dcr.wallet.AddressPrivKey(dcr.ctx, address)
```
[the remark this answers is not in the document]
buck54321: I don't think so. Same information, just encoded differently.

client/asset/dcr/dcr.go
buck54321: It's a good point though, and I'll mention something in the docs about account vs. wallet unlocking.

client/asset/dcr/dcr.go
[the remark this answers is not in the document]
buck54321: The "account" config option itself is part of the `WalletDefinition`. If RPC wallet wasn't built-in, we wouldn't even be decoding the setting. `ExchangeWallet` shouldn't be relying on it's presence.

client/asset/dcr/dcr.go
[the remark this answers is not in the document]
buck54321: Still need to do this one, but I think I've hit the other review comments. Will get this done today.

chappjc:
I would like to split this into two different PRs, one for DCR and one for BTC.  The changes are unrelated AFAICT.

I'm also having some trouble with the btc changes, and it would be best to address it separately.  Namely, there are tons of new complications with `spvWallet.runningWallet` and `SPVCore` that don't relate to the improvement to `estimateFee` or the addition of custom wallet constructors.  I know it's unwelcome work, but I think the spv.go refactoring to introduce `LoadExternalSPVWallet` should even be separate from the other btc changes.  The added abstraction with the `runningWallet func() (*SPVCore, error)` field is unnecessary at present.  Can we cross the `LoadExternalSPVWallet` bridge separately?

buck54321:
I'll come back with this stuff separately.

PR #1280, opened 2021-11-10
client/core: fix conns data races

chappjc:
In testing https://github.com/decred/dcrdex/pull/1141 with manual account disable I hit some races with the `c.conns` maps.

The one in `(*Core).exchangeMap()` is a little surprising since that method is used all the time.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1366,8 +1366,9 @@ func (c *Core) dexConnections() []*dexConnection {
 // exchangeMap creates a map of *Exchange keyed by host, including markets and
 // orders.
 func (c *Core) exchangeMap() map[string]*Exchange {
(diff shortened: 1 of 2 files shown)


client/core/core.go
```diff
@@ -1366,8 +1366,9 @@ func (c *Core) dexConnections() []*dexConnection {
 // exchangeMap creates a map of *Exchange keyed by host, including markets and
 // orders.
 func (c *Core) exchangeMap() map[string]*Exchange {
-	infos := make(map[string]*Exchange, len(c.conns))
```
chappjc, line 1369: It was the `len(c.conns)`.  I was just gonna remove the pre-alloc, but it's just as easy to grab the slice first and use it in the `range`

JoeGruffins (approved):

PR #1283, opened 2021-11-10
client/asset/{btc,dcr}: broadcast async

chappjc:
Broadcasting the counterparty swap transaction in `AuditContract` is not required to pass the audit, yet for SPV wallets it can be quite slow.  It can even timeout if there is p2p trouble.
As such, just broadcast in a goroutine so the `AuditInfo` return can be recorded as soon as possible.

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -1885,12 +1885,15 @@ func (btc *ExchangeWallet) AuditContract(coinID, contract, txData dex.Bytes, sin
 			contractHash, addr.ScriptAddress())
 	}
 
-	// Broadcast the transaction.
-	if hashSent, err := btc.node.sendRawTransaction(tx); err != nil {
-		btc.log.Debugf("Rebroadcasting counterparty contract %v (THIS MAY BE NORMAL): %v", txHash, err)
-	} else if !hashSent.IsEqual(txHash) {
-		btc.log.Errorf("Counterparty contract %v was rebroadcast as %v!", txHash, hashSent)
-	}
+	// Broadcast the transaction, but do not block because this is not required
+	// and does not affect the audit result.
+	go func() {
+		if hashSent, err := btc.node.sendRawTransaction(tx); err != nil {
+			btc.log.Debugf("Rebroadcasting counterparty contract %v (THIS MAY BE NORMAL): %v", txHash, err)
+		} else if !hashSent.IsEqual(txHash) {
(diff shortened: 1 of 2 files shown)


client/asset/btc/btc.go
```diff
@@ -1885,12 +1885,15 @@ func (btc *ExchangeWallet) AuditContract(coinID, contract, txData dex.Bytes, sin
...
-	}
+	// Broadcast the transaction, but do not block because this is not required
+	// and does not affect the audit result.
+	go func() {
+		if hashSent, err := btc.node.sendRawTransaction(tx); err != nil {
+			btc.log.Debugf("Rebroadcasting counterparty contract %v (THIS MAY BE NORMAL): %v", txHash, err)
+		} else if !hashSent.IsEqual(txHash) {
+			btc.log.Errorf("Counterparty contract %v was rebroadcast as %v!", txHash, hashSent)
+		}
+	}()
```
JoeGruffins, line 1896: imo it would be best to also add a timeout and give up and log if it's something absurd.

client/asset/dcr/dcr.go
```diff
@@ -1551,14 +1551,16 @@ func (dcr *ExchangeWallet) AuditContract(coinID, contract, txData dex.Bytes, _ t
...
-	}
+	// The counter-party should have broadcasted the contract tx but rebroadcast
+	// just in case to ensure that the tx is sent to the network. Do not block
+	// because this is not required and does not affect the audit result.
+	go func() {
+		if hashSent, err := dcr.wallet.SendRawTransaction(dcr.ctx, contractTx, true); err != nil {
+			dcr.log.Debugf("Rebroadcasting counterparty contract %v (THIS MAY BE NORMAL): %v", txHash, err)
+		} else if !hashSent.IsEqual(txHash) {
+			dcr.log.Errorf("Counterparty contract %v was rebroadcast as %v!", txHash, hashSent)
+		}
```
JoeGruffins, line 1562: could create a new context with a timeout?

JoeGruffins (approved):

client/asset/btc/btc.go
buck54321: I agree with @JoeGruffins in spirit, but it isn't straightforward to get a custom `context.Context` into `call`, so you'd have to write a custom call handler or something. Since it's really just for messaging, it's hard to justify the effort.

buck54321 (approved):

PR #1284, opened 2021-11-11
server,eth: add method for account balance retrieval

buck54321:
The interfaces are going to need to diverge when it comes to tracking funding ability. This PR is a first thrust at a reasonable interface scheme so that we can begin work in `Swapper`, `Market`, and `CoinLocker`.

### DIFF
--- a/server/asset/eth/rpcclient.go
+++ b/server/asset/eth/rpcclient.go
@@ -120,3 +120,13 @@ func (c *rpcclient) swap(ctx context.Context, secretHash [32]byte) (*swap.ETHSwa
 func (c *rpcclient) transaction(ctx context.Context, hash common.Hash) (tx *types.Transaction, isMempool bool, err error) {
 	return c.ec.TransactionByHash(ctx, hash)
 }
+
+// accountBalance gets the account balance, including the effects of known
+// unmined transactions.
+func (c *rpcclient) accountBalance(ctx context.Context, addr common.Address) (*big.Int, error) {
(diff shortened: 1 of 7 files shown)


JoeGruffins (approved):
Looks reasonable to me. Please add meaningful commit messages.

chappjc:
Needs resolve after txData PR merged.  Might as well squash down with a final commit too.

server/asset/eth/rpcclient.go
```diff
@@ -120,3 +120,13 @@ func (c *rpcclient) swap(ctx context.Context, secretHash [32]byte) (*swap.ETHSwa
 func (c *rpcclient) transaction(ctx context.Context, hash common.Hash) (tx *types.Transaction, isMempool bool, err error) {
 	return c.ec.TransactionByHash(ctx, hash)
 }
+
+// accountBalance gets the account balance, including the effects of known
+// unmined transactions.
+func (c *rpcclient) accountBalance(ctx context.Context, addr common.Address) (*big.Int, error) {
+	bigBal, err := c.ec.PendingBalanceAt(ctx, addr)
```
chappjc, line 127: I wonder how this works with multiple txns with the same nonce, e.g. replacements commonly used to either cancel a txn or speed up a txn

chappjc: Exactly, a 0 ETH transaction to yourself.  At least that's how I've seen it on etherscan.

chappjc (approved):

PR #1285, opened 2021-11-11
client/asset/btc: log the acct xpub on start

chappjc:
The btcwallet account extended public key for the p2wpkh keys are useful
for a variety of reasons, so add a trace log in startWallet to show it.

### DIFF
--- a/server/asset/btc/addresser_test.go
+++ b/server/asset/btc/addresser_test.go
@@ -51,6 +51,9 @@ func TestAddressDeriver(t *testing.T) {
 	// params := &chaincfg.TestNet3Params
 	// zpub := "vpub5UpDTWU6Nuj9uuU5TZnxjS97SAjKwY4UiLvvnYNRC6vmvbDa5toZRG1BkqCnLSRmNYuLuqLDPtdDq6YvELoMUSPjVSFTsX1H42kabkJDwWD"
 	// xpubWant := "tpubD8rNb5dcPYmZtJ3AaUMJMXfV7E7spAYY6CZfhtDLJ2SRn5WjfRymopzJ3HCjrkgqyjfs2N3CrKANHrGf4bUcxMDEHiPcisJVBSUH8wNhdoA"
+	// params := &chaincfg.RegressionNetParams
+	// zpub := "vpub5YqS5eyUTyzmeuQiyW96jgBcRiCNSEMapKkUcF6L8kWSrma29AXi9Zr4e5rxCjk4eZnP8sTpeXNJDdFr6kMGjh8ZqraXpYyyq1XfaaPwt32"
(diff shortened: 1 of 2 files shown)


server/asset/btc/addresser_test.go
```diff
@@ -51,6 +51,9 @@ func TestAddressDeriver(t *testing.T) {
 	// params := &chaincfg.TestNet3Params
 	// zpub := "vpub5UpDTWU6Nuj9uuU5TZnxjS97SAjKwY4UiLvvnYNRC6vmvbDa5toZRG1BkqCnLSRmNYuLuqLDPtdDq6YvELoMUSPjVSFTsX1H42kabkJDwWD"
 	// xpubWant := "tpubD8rNb5dcPYmZtJ3AaUMJMXfV7E7spAYY6CZfhtDLJ2SRn5WjfRymopzJ3HCjrkgqyjfs2N3CrKANHrGf4bUcxMDEHiPcisJVBSUH8wNhdoA"
+	// params := &chaincfg.RegressionNetParams
+	// zpub := "vpub5YqS5eyUTyzmeuQiyW96jgBcRiCNSEMapKkUcF6L8kWSrma29AXi9Zr4e5rxCjk4eZnP8sTpeXNJDdFr6kMGjh8ZqraXpYyyq1XfaaPwt32"
```
chappjc, line 55: I verified the addresses spewed in this test match the deposit addresses shown on the /wallets page.

buck54321:
Isn't this a privacy issue? Can't someone with this key identify every one of your transactions? If so, I don't want to dump it to my unencrypted log file every startup. We need to look at exporting keys in a way that is asset-specific and can include extended public keys.

chappjc:
Yes, it definitely is a privacy concern.  Which is why I've both used TRC level, but also kept this in draft.

An export key function would be ideal, but also a feature that almost no one would use and code to maintain.  :(

Not sure if this PR or that export function are something to do at all.  This diff is something I would use to get the xpub for the btcwallet I can seed with https://github.com/chappjc/dcrdex/commit/b6ac848bc6571dd12bdc6600ea511db8a1f3440d

chappjc:
Ah, BTW, @buck54321 thoughts about the `assetseed` cli in https://github.com/chappjc/dcrdex/commit/b6ac848bc6571dd12bdc6600ea511db8a1f3440d?
I haven't PR'd that yet because I want to test out seeds for both dcrwallet and btcwallet first, but it's something I think that should be possible.

buck54321:
> Ah, BTW, @buck54321 thoughts about the `assetseed` cli in [chappjc@b6ac848](https://github.com/chappjc/dcrdex/commit/b6ac848bc6571dd12bdc6600ea511db8a1f3440d)?

No objections from me.

PR #1288, opened 2021-11-12
server/comms: add new routes to limiter

chappjc:
Add the newish price feed, spots, and candles to the `routeLimiter`.
The "info" limiter includes the spots and candles routes.
A new shared limiter for subscription routes is created for the `order_book` and the new `price_feed` routes.
Add the connect route with its own rates. Burst is still 100. Clients should self limit during account discover, as we do.

### DIFF
--- a/server/comms/server.go
+++ b/server/comms/server.go
@@ -53,9 +53,10 @@ const (
 	// operations. Consider which routes are authenticated when setting these.
 	wsRateStatus, wsBurstStatus     = 10, 500     // order_status and match_status (combined)
 	wsRateOrder, wsBurstOrder       = 5, 100      // market, limit, and cancel (combined)
-	wsRateInfo, wsBurstInfo         = 10, 200     // low-cost route limiter for: config, fee_rate  (combined)
-	wsRateBook, wsBurstBook         = 1, 100      // orderbook, assuming 100 markets subscribed max
+	wsRateInfo, wsBurstInfo         = 10, 200     // low-cost route limiter for: config, fee_rate, spots, candles (combined)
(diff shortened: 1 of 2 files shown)


server/comms/server.go
```diff
@@ -53,9 +53,10 @@ const (
 	// operations. Consider which routes are authenticated when setting these.
 	wsRateStatus, wsBurstStatus     = 10, 500     // order_status and match_status (combined)
 	wsRateOrder, wsBurstOrder       = 5, 100      // market, limit, and cancel (combined)
-	wsRateInfo, wsBurstInfo         = 10, 200     // low-cost route limiter for: config, fee_rate  (combined)
-	wsRateBook, wsBurstBook         = 1, 100      // orderbook, assuming 100 markets subscribed max
+	wsRateInfo, wsBurstInfo         = 10, 200     // low-cost route limiter for: config, fee_rate, spots, candles (combined)
+	wsRateSubs, wsBurstSubs         = 2, 100      // subscriptions: orderbook and price feed (combined)
 	wsRateRegister, wsBurstRegister = 1 / 60.0, 1 // register, rate.Every(time.Minute) but const
+	wsRateConnect, wsBurstConnect   = 4, 100      // connect, account discovery requires bursts - (*Core).discoverAccount
```
buck54321, line 59: Wouldn't the extra connections just use some of the burst tokens? A constant rate of 4 connections per second seems really high. What would be the harm in going lower, maybe `1 / 10.0` ?

In general, these rates seem pretty permissive.

chappjc: That's a good point re the connect route.  I agree.
I've tried to be cautious about the limits, but this one in particular really has no good case for **sustained** 4 per second.  Will update.

PR #1289, opened 2021-11-12
[for discussion] eth: locked amount accounting updates

chappjc:


client/asset/eth/eth.go
```diff
@@ -535,165 +540,151 @@ func decodeCoinID(coinID []byte) (*coin, error) {
...
+// amount minus the actual amount *spent* by this swap txn.
 func (*ExchangeWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin, uint64, error) {
+	// TODO: remember to unlockFunds when ETH is sent to a swap contract.
 	return nil, nil, 0, asset.ErrNotImplemented
 }
 
 // Redeem sends the redemption transaction, which may contain more than one
 // redemption.
 func (*ExchangeWallet) Redeem(form *asset.RedeemForm) ([]dex.Bytes, asset.Coin, uint64, error) {
+	// TODO: remember to unlockFunds for the reserved redeem gas amount.
```
martonp, line 669: When would those funds get locked in the first place?

client/asset/eth/eth.go
```diff
@@ -501,26 +511,21 @@ func (*ExchangeWallet) PreRedeem(req *asset.PreRedeemForm) (*asset.PreRedeem, er
 	}, nil
 }
 
-// coin implements the asset.Coin interface for ETH
+// coin implements the asset.Coin interface for ETH.
 type coin struct {
-	id srveth.AmountCoinID
+	srveth.AmountCoinID
```
martonp, line 516: I'm updating this struct in #1248, so there's no point to do this unless you disagree with those changes.

client/asset/eth/eth.go
```diff
@@ -192,13 +202,10 @@ type ExchangeWallet struct {
 	tipMtx     sync.RWMutex
 	currentTip *types.Block
 
-	acct *accounts.Account
+	acct *lockedAcct
 
 	cachedInitGas    uint64
 	cachedInitGasMtx sync.Mutex
-
-	lockedFunds    map[string]uint64 // gwei
```
martonp, line 200: We discussed having the map vs just an integer over here:
https://github.com/decred/dcrdex/pull/1221#discussion_r722888631

I don't think there's any benefit other than serving as a safety check, but I don't think there's any downside either.

Having the `lockedAcct` struct is definitely nicer though in case we decide to support multiple addresses.

client/asset/eth/eth.go
chappjc: This is an open question, but a as a thought I made a comment https://github.com/decred/dcrdex/pull/1289/files#diff-671406d18db2f8e1a281545be05fcdefd789b63614398d678d38e2b2814cf788R549 `FundBuyOrder`

client/asset/eth/eth.go
chappjc: That's fine.  I'm not going to ram this PR through, just thinking through the fund locking and nonce issue.

client/asset/eth/eth.go
```diff
@@ -535,165 +540,151 @@ func decodeCoinID(coinID []byte) (*coin, error) {
...
 }
 
-// FundOrder selects coins for use in an order. The coins will be locked, and
-// will not be returned in subsequent calls to FundOrder or calculated in calls
-// to Available, unless they are unlocked with ReturnCoins.
-// In UTXO based coins, the returned []dex.Bytes contains the redeem scripts for the
-// selected coins, but since there are no redeem scripts in Ethereum, nil is returned.
-// Equal number of coins and redeem scripts must be returned.
+// TODO: "fund" a buy order by locking funds for the redeem transaction gas.
+// func (eth *ExchangeWallet) FundBuyOrder(ord *asset.Order) { do not use ord.Value, just MaxSwapCount }
```
chappjc, line 549: This doesn't need to be part of every asset's interface, just a method to assert for at runtime to see if it's necessary.

client/asset/eth/eth.go
```diff
@@ -535,165 +540,151 @@ func decodeCoinID(coinID []byte) (*coin, error) {
...
 
 // Swap sends the swaps in a single transaction. The Receipts returned can be
 // used to refund a failed transaction. The Input coins are manually unlocked
 // because they're not auto-unlocked by the wallet and therefore inaccurately
 // included as part of the locked balance despite being spent.
+//
+// NOTE: we must also emulate "change", which becomes the new amount locked by
+// parent order that the consumer will unlock with ReturnCoins when the order is
+// retired (no more swaps and no longer booked). The change amount is the input
+// amount minus the actual amount *spent* by this swap txn.
```
chappjc, line 660: @martonp When diving into this, I was considering your Swap PR.  Namely https://github.com/decred/dcrdex/pull/1248/commits/ee847d09c8e13970abcf2024396a37278051affe#diff-671406d18db2f8e1a281545be05fcdefd789b63614398d678d38e2b2814cf788R831-R838.

I believe this is actually a way it's good to just have amounts and not unique coin ids keying in a map of values.  Instead of `unlockFunds(inputs)` followed  by `lockFunds(new fake coin for change)`, it would just be `unlockFunds(totalUsedValue)`

client/asset/eth/eth.go
chappjc: I recall that discussion, and the earlier one about adding the nonce in the first place.  https://github.com/decred/dcrdex/pull/1221#discussion_r722737273
I'm still not convinced the more complicated coin id encoding helps much, and I'm trying to get to the bottom of it in the PR, which is just for discussion at this point.

client/asset/eth/eth.go
martonp: I agree with doing this.

client/asset/eth/eth.go
martonp: Yeah I'm not opposed to removing this map either tbh, I think it might be more trouble than it's worth.

client/asset/eth/eth.go
martonp: I think this will work.

client/asset/eth/eth.go
martonp: Unless anyone is very adamant about keeping the map, might as well front run that PR and update this.

chappjc: Whatever we do, I did mean to imply that https://github.com/decred/dcrdex/pull/1248 should go first.  I'm not trying to jump the line.  :)

martonp: @buck54321 @JoeGruffins What's your opinion on this?

buck54321: I anxious to see where chappjc takes this. I do think the idea of discrete "coins" is of little value server-side, but could provide a convenient way for account client-side. Maybe just a quick thought experiment.

We'll call the counter method **A** and the coin method **B**.

Let's say a multi-lot order requires 100 units max. We'll lock up that value by either 

A) incrementing the counter by 100 units. new value 100 units.

B) creating a coin with an associated value of 100 units and adding it to the map

OK. Time to swap.  

```Go
Swap(*Swaps) (receipts []Receipt, changeCoin Coin, feesPaid uint64, err error)
```

A) I guess we just don't pass a "coin" in the `Swaps.Inputs`. If we're not going to put a nonce in, and the value isn't necessarily unique, what's the purpose of a coin? We decrement the counter by the amount sent. Not sure how we track the remainder. Maybe `Core` needs special handling. And what if this is the last swap. We will always have a remainder from original locked amount. How is that remainder tracked and returned after the order completes?

B) Decrement the amount decoded from the Swap.Inputs, create a new coin with the decremented value and return it as the change coin. If this is the last swap in the order, Core indicates indicates that with the `Swaps.LockChange` field, and we decrement the entire input coin.

@chappjc's obviously got something working, but to me, it looks like "coins" make a good accounting tool client-side and fit very nicely in the existing **core** system with virtually no changes.

chappjc: > A) I guess we just don't pass a "coin" in the `Swaps.Inputs`. If we're not going to put a nonce in, and the value isn't necessarily unique, what's the purpose of a coin? We decrement the counter by the amount sent. Not sure how we track the remainder. Maybe `Core` needs special handling. And what if this is the last swap. We will always have a remainder from original locked amount. How is that remainder tracked and returned after the order completes?

Here it would still use a changeCoin here to convey the remainder amount that needs to be unlocked later.  That won't need to change.  Everything is in place with `trackedTrade.{cancel,change}` for this in Core, and we rely on it presently, so I don't believe there's really much if any change on the Core machinery.
In ETH on the client, the only purpose of a Coin is the value, at least until/if we have multiple account support.

buck54321: So we still have discrete "coins", but we just use them to know how much to change the counter rather than storing them in a map? That should work fine. 

And since we don't need a map key, I guess the coins don't need a nonce. We can just trust that core is handling them properly.

chappjc: Yes, that's the idea.  We don't need to get away from the `trackedTrade.{coin,change}` workings at all, we just don't need a nonce.  I convinced myself this was fine in reviewing the Core bits around the `trackedTrade` and returncoins.

Anyway, I think I'm just going to let this PR sit until a number of other ETH PRs are in first, but I wanted to discuss this asap because I had a terrible itch regarding the nonce business.

martonp: The original reason for the nonce was to make the signed messages that are sent to the server unique.

chappjc: > The original reason for the nonce was to make the signed messages that are sent to the server unique.

I'm glad you're pointing this out @martonp, because that was never really an issue to solve I believe.  You get reused UTXOs commonly too  https://github.com/decred/dcrdex/pull/1221#discussion_r721714244
And now with @buck54321's account-balance-based solution in https://github.com/decred/dcrdex/pull/1293, the server has no need for them to be unique.

JoeGruffins:
There were a few things I thought the nonce helped with client side.

One is persisting the locked coins in the db. But I guess there's no reason to do this, as the lock coins are recalculated during resumeTrades.

The other is to make sure you aren't locking or unlocking the same coins for the same trades. I would think it's easier to do accounting with a way to know what values are locked for what trade, but I guess it's comparable to eth's own accounting, where you forget what txs led up to an account's state. Although, they do have an incremented nonce. Maybe the trade and match trackers will ensure that nothing is double locked or unlocked.

But I do think troubleshooting would be easier, if a strange amount was locked, and the amount had something that connected it to an order or match, or why it's locked. Unique coin id's were helpful when trying to figure out why more coins were being locked for dcr in the simnet tests recently. I was able to pinpoint that it was happening with the taker's coins partly because of unique coin ids.

chappjc:
I'm still baffled by the nonce, but we'll figure it out

PR #1290, opened 2021-11-14
eth/client: Replace gas estimates with hardcoded values

martonp:
There is no variation in the execution path for initiating swaps, so there's no point in doing gas estimations. The gas estimation happens without any external calls, and on any updates to the geth code which would invalidate the hardcoded values, the harness tests will let us know.

This will come on top of #1262


martonp:
I incorporated this change into #1248

PR #1291, opened 2021-11-14
eth/client: Lock funds for possible refund

martonp:
Prior to this, `FundOrder` did not lock funds to use for the gas fees in the case of a refund.


chappjc (approved):
LGTM.  Will merge in a few days unless there are other approvals sooner.

JoeGruffins (approved):

buck54321 (approved):
A couple constants moved in #1301, but this should go in first.

PR #1293, opened 2021-11-15
server/{swap,market}: add funding checks for account-based assets

buck54321:
Adds funding validation for account-based assets i.e. eth.

**dex**: Add `RedeemSize` field to `Asset`.

**msgjson**: Add `RedeemSig` field to `Trade`. `RedeemSig` should be
populated when the redemption is to an account-based asset.
This PR also indirectly proposes a protocol for specifying
the funding account. The account address should be utf-8 encoded
and passed as the coin ID of the only coin, along with the pubkey
and signature. The data signed will be the serialized `msgjson.Limit` or
`msgjson.Market`, with server stamp set to 0. Same signature input
for the `RedeemSig`.

**dex/order*: The `Trade` type has methods for accessing the account address.

**server/asset**: `AccountBalancer` gets a `ValidateSignature` method.

**general scheme**: Orders and matches involving account-based assets are
indexed by `Swapper` and `OrderPQ` in the same way they are indexed for
users (`account.AccountID`). When an order comes into the order router,
stats for existing orders, including outstanding redemptions, is
fetched from all markets and the swapper. These stats are used with
the new order info to validate the users balance. The user must have
sufficient balance to cover all outstanding orders and redemptions.

**server/book**: All `OrderPQ` are given an `AccountTracker` set up for
any account-based assets, or none. The `AccountTracker` is essentially
the same as the `userOrders` index, but indexed by account address
instead. This is important because more than one user could be
using the same account.

`AccountTracker` has methods for iterating booked orders, exposed via
`Book` and `Market` methods used by `Market` as part of the `MarketTunnel`
interface method, `PendingAccount`, described below.

**server/swap**: `Swapper` gets an account index to mirror its user index.
The new `AccountStats` method returns information about in-process matches
that aren't yet swapped/redeemed for a particular asset.

**server/market**: `OrderRouter` is refactored for improved re-use between
`handleLimit` and `handleMarket`. For account based assets, `OrderRouter`
checks the `MarketTunnels` and `MatchNegotiator` for outstanding
order and match info, and then queries the backend (via `DEXBalancer`)
to see if the account has sufficient balance.

`NewMarket` handles balance checks on startup. This is accomplished
by additional tracking through the orders loop, and then a balance
check for account-based assets immediately before adding to the
order book. utxo-based asset handling is unchanged except that the
lot size compliance check it moved earlier in the loop.

### DIFF
--- a/dex/calc/fees.go
+++ b/dex/calc/fees.go
@@ -14,12 +14,16 @@ import (
 // number of lots in the order. For the quote asset, maxSwaps is not swapVal /
 // lotSize, so it must be a separate parameter. The chained swap txns will be
 // the standard size as they will spend a previous swap's change output.
+// For account-based assets, inputsSize will be zero, and nfo.SwapSize =
+// nfo.SwapSizeBase.
 func RequiredOrderFunds(swapVal, inputsSize, maxSwaps uint64, nfo *dex.Asset) uint64 {
 	return RequiredOrderFundsAlt(swapVal, inputsSize, maxSwaps, nfo.SwapSizeBase, nfo.SwapSize, nfo.MaxFeeRate)
 }
 
 // RequiredOrderFundsAlt is the same as RequiredOrderFunds, but built-in type
 // parameters.
+// For account-based assets, inputsSize will be zero, and swapSize =
+// swapSizeBase.
 func RequiredOrderFundsAlt(swapVal, inputsSize, maxSwaps, swapSizeBase, swapSize, feeRate uint64) uint64 {
 	baseBytes := maxSwaps * swapSize
 	// SwapSize already includes one input, replace the size of the first swap
--- a/dex/order/order.go
+++ b/dex/order/order.go
@@ -473,6 +473,42 @@ func (t *Trade) SwapAddress() string {
 	return t.Address
 }
 
+// FromAccount is the account that the order originates from. Only useful for
+// account-based assets. Use of this method assumes that account coin has
+// already been added.
+func (t *Trade) FromAccount() string {
+	if len(t.Coins) == 0 {
+		return "no coins?"
+	}
+	return string(t.Coins[0])
+}
+
+// ToAccount is the account that the order pays to. Only useful for
+// account-based assets.
+func (t *Trade) ToAccount() string {
+	return t.Address
+}
+
+// BaseAccount is the account address associated with the base asset for the
+// order. Only useful for account-based assets. Use of this method assumes that
+// account coin has already been added (when sell = true).
+func (t *Trade) BaseAccount() string {
+	if t.Sell {
+		return t.FromAccount()
+	}
+	return t.ToAccount()
+}
+
+// QuoteAccount is the account address associated with the quote asset for the
+// order. Only useful for account-based assets. Use of this method assumes that
+// account coin has already been added (when sell = false).
+func (t *Trade) QuoteAccount() string {
+	if t.Sell {
+		return t.ToAccount()
+	}
+	return t.FromAccount()
+}
+
 // serializeSize returns the length of the serialized Trade.
 func (t *Trade) serializeSize() int {
 	// Compute the size of the serialized Coin IDs.
--- a/dex/order/order.go
+++ b/dex/order/order.go
@@ -480,6 +480,8 @@ func (t *Trade) FromAccount() string {
 	if len(t.Coins) == 0 {
 		return "no coins?"
 	}
+	// The coin ID should be the UTF-8 encoded address string, not the address
+	// byte-array. t.Coins[0] = []byte(addrStr).
 	return string(t.Coins[0])
 }
 
--- a/server/asset/common.go
+++ b/server/asset/common.go
@@ -62,6 +62,9 @@ type Backend interface {
 	CheckAddress(string) bool
 	// ValidateCoinID checks the coinID to ensure it can be decoded, returning a
 	// human-readable string if it is valid.
+	// Note: ValidateCoinID is NOT used for funding coin IDs for account-based
+	// assets. This rule is only enforced by code patterns right now, but we may
+	// consider adding separate methods in the future.
 	ValidateCoinID(coinID []byte) (string, error)
 	// ValidateContract ensures that the swap contract is constructed properly
 	// for the asset.
--- a/server/asset/common.go
+++ b/server/asset/common.go
@@ -92,6 +92,9 @@ type OutputTracker interface {
 type AccountBalancer interface {
 	// AccountBalance retrieves the current account balance.
 	AccountBalance(addr string) (uint64, error)
+	// ValidateSignature checks that the pubkey is correct for the address and
+	// that the signature shows ownership of the associated private key.
+	ValidateSignature(addr string, pubkey, msg, sig []byte) error
 }
 
 // Coin represents a transaction input or output.
--- a/server/asset/common.go
+++ b/server/asset/common.go
@@ -97,6 +97,10 @@ type AccountBalancer interface {
 	AccountBalance(addr string) (uint64, error)
 	// ValidateSignature checks that the pubkey is correct for the address and
 	// that the signature shows ownership of the associated private key.
+	// IMPORTANT: As part of signature validation, the asset backend should
+	// validate the address against a STRICT standard. Case, prefixes, suffixes,
+	// etc. must be exactly the same order-to-order, since the address string
+	// is used as a key for various accounting operations throughout DEX.
 	ValidateSignature(addr string, pubkey, msg, sig []byte) error
 }
 
--- a/server/asset/eth/rpcclient.go
+++ b/server/asset/eth/rpcclient.go
@@ -138,6 +138,14 @@ func (c *rpcclient) accountBalance(ctx context.Context, addr common.Address) (*b
 		return nil, err
 	}
 
+	// We need to subtract and pending outgoing value, but ignore any pending
+	// incoming value since that can't be spent until mined. So we can't using
+	// PendingBalanceAt or BalanceAt by themselves.
+	// We'll iterate tx pool transactions and subtract any value and fees being
+	// sent from this account. The rpc.Client doesn't expose the
+	// txpool_contentFrom => (*TxPool).ContentFrom RPC method, for whatever
+	// reason, so we'll have to use CallContext and copy the mimic the
+	// internal RPCTransaction type.
 	var txs map[string]map[string]*RPCTransaction
 	err = c.c.CallContext(ctx, &txs, "txpool_contentFrom", addr)
 	if err != nil {
@@ -153,6 +161,8 @@ func (c *rpcclient) accountBalance(ctx context.Context, addr common.Address) (*b
 				outgoing.Add(outgoing, new(big.Int).Mul(gas, tx.GasPrice.ToInt()))
 			} else if tx.GasFeeCap != nil {
 				outgoing.Add(outgoing, new(big.Int).Mul(gas, tx.GasFeeCap.ToInt()))
+			} else {
+				return nil, fmt.Errorf("cannot find fees for tx %s", tx.Hash)
 			}
 		}
 	}
@@ -165,11 +175,12 @@ type RPCTransaction struct {
 	Gas       hexutil.Uint64 `json:"gas"`
 	GasPrice  *hexutil.Big   `json:"gasPrice"`
 	GasFeeCap *hexutil.Big   `json:"maxFeePerGas,omitempty"`
+	Hash      common.Hash    `json:"hash"`
 	// BlockHash        *common.Hash      `json:"blockHash"`
 	// BlockNumber      *hexutil.Big      `json:"blockNumber"`
 	// From             common.Address    `json:"from"`
 	// GasTipCap        *hexutil.Big      `json:"maxPriorityFeePerGas,omitempty"`
-	// Hash             common.Hash       `json:"hash"`
+
 	// Input            hexutil.Bytes     `json:"input"`
 	// Nonce            hexutil.Uint64    `json:"nonce"`
 	// To               *common.Address   `json:"to"`
--- a/server/asset/eth/rpcclient.go
+++ b/server/asset/eth/rpcclient.go
@@ -15,6 +15,7 @@ import (
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/ethereum/go-ethereum/p2p"
@@ -22,7 +23,11 @@ import (
 )
 
 // Check that rpcclient satisfies the ethFetcher interface.
-var _ ethFetcher = (*rpcclient)(nil)
+var (
+	_ ethFetcher = (*rpcclient)(nil)
+
+	bigZero = new(big.Int)
+)
 
 type rpcclient struct {
 	// ec wraps a *rpc.Client with some useful calls.
@@ -124,9 +129,55 @@ func (c *rpcclient) transaction(ctx context.Context, hash common.Hash) (tx *type
 // accountBalance gets the account balance, including the effects of known
 // unmined transactions.
 func (c *rpcclient) accountBalance(ctx context.Context, addr common.Address) (*big.Int, error) {
-	bigBal, err := c.ec.PendingBalanceAt(ctx, addr)
+	tip, err := c.blockNumber(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("blockNumber error: %v", err)
+	}
+	currentBal, err := c.ec.BalanceAt(ctx, addr, big.NewInt(int64(tip)))
 	if err != nil {
 		return nil, err
 	}
-	return bigBal, nil
+
+	var txs map[string]map[string]*RPCTransaction
+	err = c.c.CallContext(ctx, &txs, "txpool_contentFrom", addr)
+	if err != nil {
+		return nil, fmt.Errorf("contentFrom error: %w", err)
+	}
+
+	outgoing := new(big.Int)
+	for _, group := range txs { // 2 groups, pending and queued
+		for _, tx := range group {
+			outgoing.Add(outgoing, tx.Value.ToInt())
+			gas := new(big.Int).SetUint64(uint64(tx.Gas))
+			if tx.GasPrice != nil && tx.GasPrice.ToInt().Cmp(bigZero) > 0 {
+				outgoing.Add(outgoing, new(big.Int).Mul(gas, tx.GasPrice.ToInt()))
+			} else if tx.GasFeeCap != nil {
+				outgoing.Add(outgoing, new(big.Int).Mul(gas, tx.GasFeeCap.ToInt()))
+			}
+		}
+	}
+
+	return currentBal.Sub(currentBal, outgoing), nil
+}
+
+type RPCTransaction struct {
+	Value     *hexutil.Big   `json:"value"`
+	Gas       hexutil.Uint64 `json:"gas"`
+	GasPrice  *hexutil.Big   `json:"gasPrice"`
+	GasFeeCap *hexutil.Big   `json:"maxFeePerGas,omitempty"`
+	// BlockHash        *common.Hash      `json:"blockHash"`
+	// BlockNumber      *hexutil.Big      `json:"blockNumber"`
+	// From             common.Address    `json:"from"`
+	// GasTipCap        *hexutil.Big      `json:"maxPriorityFeePerGas,omitempty"`
+	// Hash             common.Hash       `json:"hash"`
+	// Input            hexutil.Bytes     `json:"input"`
+	// Nonce            hexutil.Uint64    `json:"nonce"`
+	// To               *common.Address   `json:"to"`
+	// TransactionIndex *hexutil.Uint64   `json:"transactionIndex"`
+	// Type             hexutil.Uint64    `json:"type"`
+	// Accesses         *types.AccessList `json:"accessList,omitempty"`
+	// ChainID          *hexutil.Big      `json:"chainId,omitempty"`
+	// V                *hexutil.Big      `json:"v"`
+	// R                *hexutil.Big      `json:"r"`
+	// S                *hexutil.Big      `json:"s"`
 }
--- a/server/dex/dex.go
+++ b/server/dex/dex.go
@@ -604,7 +604,8 @@ func NewDEX(ctx context.Context, cfg *DexConf) (*DEX, error) {
 	// they are all instantiated, so we are synchronous in our use of the
 	// marketTunnels map.
 	marketTunnels := make(map[string]market.MarketTunnel, len(cfg.Markets))
-	dexBalancer := market.NewDEXBalancer(marketTunnels, backedAssets, swapper)
+	pendingAccounters := make(map[string]market.PendingAccounter, len(cfg.Markets))
+	dexBalancer := market.NewDEXBalancer(pendingAccounters, backedAssets, swapper)
 
 	// Markets
 	usersWithOrders := make(map[account.AccountID]struct{})
@@ -635,6 +636,7 @@ func NewDEX(ctx context.Context, cfg *DexConf) (*DEX, error) {
 		}
 		markets[mktInf.Name] = mkt
 		marketTunnels[mktInf.Name] = mkt
+		pendingAccounters[mktInf.Name] = mkt
 		log.Infof("Preparing historical market data API for market %v...", mktInf.Name)
 		err = dataAPI.AddMarketSource(mkt)
 		if err != nil {
--- a/server/dex/dex.go
+++ b/server/dex/dex.go
@@ -118,7 +118,7 @@ func (ss *subsystem) stop() {
 type DEX struct {
 	network     dex.Network
 	markets     map[string]*market.Market
-	assets      map[uint32]*swap.LockableAsset
+	assets      map[uint32]*swap.SwapperAsset
 	storage     db.DEXArchivist
 	authMgr     *auth.AuthManager
 	swapper     *swap.Swapper
@@ -390,7 +390,7 @@ func NewDEX(ctx context.Context, cfg *DexConf) (*DEX, error) {
 	xpubs := make(map[string]string) // to enforce uniqueness
 
 	// Start asset backends.
-	lockableAssets := make(map[uint32]*swap.LockableAsset, len(cfg.Assets))
+	lockableAssets := make(map[uint32]*swap.SwapperAsset, len(cfg.Assets))
 	backedAssets := make(map[uint32]*asset.BackedAsset, len(cfg.Assets))
 	cfgAssets := make([]*msgjson.Asset, 0, len(cfg.Assets))
 	assetLogger := cfg.LogBackend.Logger("ASSET")
@@ -467,9 +467,9 @@ func NewDEX(ctx context.Context, cfg *DexConf) (*DEX, error) {
 		}
 
 		backedAssets[assetID] = ba
-		lockableAssets[assetID] = &swap.LockableAsset{
+		lockableAssets[assetID] = &swap.SwapperAsset{
 			BackedAsset: ba,
-			CoinLocker:  dexCoinLocker.AssetLocker(assetID).Swap(),
+			Locker:      dexCoinLocker.AssetLocker(assetID).Swap(),
 		}
 		feeMgr.AddFetcher(ba)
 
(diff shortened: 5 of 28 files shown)


chappjc:
All looks sensible based on a quick scan through, although I didn't anticipate a diff this large for the task.  :)

Not really complaints so much as a few things to note that jumped out at me:

- Either the `MatchNegotiator` interface or its one `AccountStats` method could use some reconsideration of their name(s).
- In general I am slightly off-put by the overt account awareness in all the non-asset server packages, esp. `book`, but I don't see a clearer solution or one that does not become very contrived to fit into the existing market machinery.  I'd probably have gone the same way.
- I thought we decided not to check the redeeming account balance and ownership server side, but I see the solution you proposed is not too complex.  I was a bit taken aback by the specialized `RedeemSig` in the `Trade` message, but I think it's palatable - if we require proof of funding coins, it's not too wild to require proof that the receiving account is owned and funded as well.  However, I do maintain that this is an unnecessary check.  It's easy to pass the check temporarily to place an order, and causing a trade to fail after both swaps are broadcast means you've locked your funds too (already discourages this).

Full review Mon/Tues.  Let's try to get https://github.com/decred/dcrdex/pull/1262 in asap.

buck54321:
> I didn't anticipate a diff this large for the task. :)

More than 700 lines of the the net LOC gain is tests, if that makes it any better.

> Either the `MatchNegotiator` interface or its one `AccountStats` method could use some reconsideration of their name(s).

I see that now. Will do.

> In general I am slightly off-put by the overt account awareness in all the non-asset server packages, esp. book, but I don't see a clearer solution or one that does not become very contrived to fit into the existing market machinery. I'd probably have gone the same way.

I feel the similarly. I got here because this is where the solution naturally evolved, and I think it's the right solution. It does touch some low-level, critical areas of code, but in fairly uncomplicated and auditable ways. For example, `OrderPQ` and `Book` now have methods for traversing the data, but all logic is kept in the new `AccountPending` method, and the `accountTracker` has no error return values or possible panics.

> I thought we decided not to check the redeeming account balance and ownership server side

I didn't remember any specific conclusions on the matter, so sorry if this work somehow doesn't reflect a previous discussion. I do think it's in-line with the validation that the server already performs for utxo-based assets. We do our best to make sure that the user can complete their order when it's placed. An order redeeming to an account-based asset cannot be completed without funds to pay the redemption tx fees, and the taker would be left hanging if the maker can't redeem. IMO, the reason we check any of this is ensure a smooth experience for the other party, and that still applies here.

> However, I do maintain that this is an unnecessary check.

The same could be said for funding coins for utxo-based assets. We could totally make it a free-for-all without any funding checks, and nobody's funds would be at risk.

server/market/market.go
```diff
@@ -249,37 +303,92 @@ ordersLoop:
...
+			var addr string
+			var qty, lots uint64
+			var redeems int
+			if lo.Sell {
+				// address is zeroth coin
+				if len(lo.Coins) != 1 {
+					log.Errorf("rejecting account-based-base-asset order %s that has no coins ¯\\_(ツ)_/¯", lo.ID())
+					continue ordersLoop
+				}
+				addr = string(lo.Coins[0])
```
martonp, line 316: This would be an encoded `AmountCoinID`, not just an address.

martonp (commented):
This method looks good to me, but did you consider just keeping an integer record of how much funds are currently in use by each account, and then incrementing/decrementing as needed? The code would definitely be much simpler. I guess the downside is if the record somehow gets out of sync, it's not really possible to recover.

martonp:
What we discussed was that the server just needs to check things that if incorrectly done by one user would negatively affect another user. In this case, not having enough funds to redeem a swap would only negatively affect the user making the mistake, so this check could just be handled by the client code.

buck54321:
>  In this case, not having enough funds to redeem a swap would only negatively affect the user making the mistake, so this check could just be handled by the client code.

If the maker fails to redeem, the taker has stuck funds.

buck54321:
> > In this case, not having enough funds to redeem a swap would only negatively affect the user making the mistake, so this check could just be handled by the client code.
> 
> If the maker fails to redeem, the taker has stuck funds.

Just some additional shower thoughts on this...

So technically, we don't need to check it if it's a market order or limit order with time-in-force = immediate, since the taker cannot inconvenience the maker and can only shoot themselves in the foot. This means that nobody can really point at DEX when things go wrong, 

**BUT**

they will anyway. And while I agree it's not the server's job to write the client software, that won't stop some noob using a crappy or misconfigured client from running around screaming "DEX locked my money in a contract and wants me to pay to get it out". Screw that guy and his crappy client, obviously, but my point is that we maintain an interest in providing a smooth user experience and loss of funds is not the only bad thing that can happen to slow DEX adoption.

dex/calc/fees.go
```diff
@@ -14,12 +14,16 @@ import (
 // number of lots in the order. For the quote asset, maxSwaps is not swapVal /
 // lotSize, so it must be a separate parameter. The chained swap txns will be
 // the standard size as they will spend a previous swap's change output.
+// For account-based assets, inputsSize will be zero, and nfo.SwapSize =
```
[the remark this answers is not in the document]
buck54321: I'm fine with having custom functions for dealing with account-based asset required funds. Outside of the asset backends, `OrderRouter` is the only user of `RequiredOrderFunds` and `RequiredOrderFundsAlt`, and `OrderRouter` is already account-aware (in this work) so the functions are only used in a context where we know already know what kind of asset we're dealing with. So we should be free to diverge these `calc` functions with little ripple-effect. I'm not 100% certain what you mean by "new backend method". Do you mean to calculate the input size? If so, there's really nothing to do. Since `OrderRouter` already knows it's account-based, it never collects the `asset.FundingCoin` (only returned from `asset.OutputTracker` anyway), which is where the input size comes from.

server/market/market.go
buck54321: As we wiggle everything into place, one of the things we need to do is to "minimize" the API by eliminating any extraneous information. The server doesn't need the value encoded in the coin ID. It's knows how much to lock for the order.

Just looking forward, I would expect a number of changes to the existing server-side coin ID system. I'll comment in more detail on #1154 today so we can plan it out together. I sort of feel bad for not being more involved with the details of the coin IDs earlier, but it was so early in the process and I hadn't quite wrapped my head around everything. I'm getting there now.

server/asset/eth/rpcclient.go
```diff
@@ -120,3 +120,13 @@ func (c *rpcclient) swap(ctx context.Context, secretHash [32]byte) (*swap.ETHSwa
 func (c *rpcclient) transaction(ctx context.Context, hash common.Hash) (tx *types.Transaction, isMempool bool, err error) {
 	return c.ec.TransactionByHash(ctx, hash)
 }
+
+// accountBalance gets the account balance, including the effects of known
+// unmined transactions.
+func (c *rpcclient) accountBalance(ctx context.Context, addr common.Address) (*big.Int, error) {
+	bigBal, err := c.ec.PendingBalanceAt(ctx, addr)
```
martonp, line 127: Just return the result of `PendingBalanceAt`

server/market/orderrouter.go
```diff
@@ -474,21 +366,10 @@ func (r *OrderRouter) handleMarket(user account.AccountID, msg *msgjson.Message)
 	var commit order.Commitment
 	copy(commit[:], market.Commit)
 
-	fundingAsset := assets.funding
+	// fundingAsset := assets.funding
```
JoeGruffins, line 369: Remove?

server/market/orderrouter.go
```diff
@@ -518,10 +399,112 @@ func (r *OrderRouter) handleMarket(user account.AccountID, msg *msgjson.Message)
...
+	user := oRecord.order.User()
+	trade := oRecord.order.Trade()
+
+	// If the receiving asset is account-based, we need to check that they can
+	// cover fees for the redemption, since they can't be subtracted from the
+	// received amount.
+	receivingBalancer, isToAccount := assets.receiving.Backend.(asset.AccountBalancer)
+	if isToAccount {
+		if redeemSig == nil {
+			log.Errorf("user %s did not include a RedeemSig for received asset %s", user, assets.receiving.Symbol)
```
JoeGruffins, line 419: Are all of the new logs here really error level? I suggest warn, or info, as they are not really a server problem.

JoeGruffins (approved):
Tests fine for me for UTXO coins. No real complaints.

server/asset/eth/rpcclient.go
```diff
@@ -124,9 +129,55 @@ func (c *rpcclient) transaction(ctx context.Context, hash common.Hash) (tx *type
...
+			gas := new(big.Int).SetUint64(uint64(tx.Gas))
+			if tx.GasPrice != nil && tx.GasPrice.ToInt().Cmp(bigZero) > 0 {
+				outgoing.Add(outgoing, new(big.Int).Mul(gas, tx.GasPrice.ToInt()))
+			} else if tx.GasFeeCap != nil {
+				outgoing.Add(outgoing, new(big.Int).Mul(gas, tx.GasFeeCap.ToInt()))
+			}
+		}
+	}
+
+	return currentBal.Sub(currentBal, outgoing), nil
```
chappjc, line 160: "smarter" in the sense that it excludes pending incoming?

buck54321: I guess. I think we have to, no?

- Lot size 1 eth
- User has current balance 1.5 eth
- Pending balance also 1.5, but with 1 incoming transfer of 1 eth and one outgoing transfer of 1 eth. 
- Place order.
- The outgoing gets mined but not the incoming. current balance = 0.5, pending = 1.5. 
- Match as maker.
- Can't swap.

buck54321: My bad. I didn't mean to drop this out of nowhere. I thought this was a revelation from the Matrix convo last week, but maybe I misunderstood. Is my read right, though?

chappjc: Sure sure, I'm not totally surprised to see the change given what we've been talking about, but there's no code comments about the intent of all this code.

One might naively just use `PendingBalanceAt` or `BalanceAt` of course so I think there needs to be explanation.  When I see `"txpool_contentFrom"` and a us cloning some json-tagged type `RPCTransaction`, I raise an eyebrow.  It's surprising to me that we're the first ones to need to do this kind of accounting or that there are no straightforward methods and types ready for this.

I say fine, lets run with it, but it deserves some docs and serious consideration because it just smells very low level.

server/market/market.go
```diff
@@ -178,20 +183,39 @@ func NewMarket(cfg *Config) (*Market, error) {
...
-		bookOrdersByID[ord.ID()] = ord
+	for _, lo := range bookOrders {
+		// Limit order amount requirements are simple unlike market buys.
+		if lo.Quantity%mktInfo.LotSize != 0 || lo.FillAmt%mktInfo.LotSize != 0 {
+			// To change market configuration, the operator should suspended the
+			// market with persist=false, but that may not have happened, or
+			// maybe a revoke failed.
+			log.Errorf("Not rebooking order %v with amount (%v/%v) incompatible with current lot size (%v)",
+				lo.ID(), lo.FillAmt, lo.Quantity, mktInfo.LotSize)
+			// Revoke the order, but do not count this against the user.
```
chappjc, line 211: Thank you for adding this.  It needs to be pulled out for a 0.4 commit as well, and ideally would be a separate commit in this PR too.

server/market/balancer.go
```diff
@@ -0,0 +1,96 @@
...
+// CheckBalance checks if there is sufficient balance to support the specified
+// new funding and redemptions, given the existing orders throughout DEX that
+// fund from or redeem to the specified account address for the account-based
+// asset. It is an internally logged error to call CheckBalance for a
+// non-account-based asset or an asset that is was not provided to the
+// constructor.
+func (b *DEXBalancer) CheckBalance(acctAddr string, assetID uint32, qty, lots uint64, redeems int) bool {
+	backedAsset, found := b.assets[assetID]
+	if !found {
+		log.Errorf("asset ID %d not found in accountBalancer assets map", assetID)
```
chappjc, line 53: There's a `backedBalancer` type instead of `accountBalancer` now, but since this isn't an error return, how about making it easier to grep back to this code location with `"(*DEXBalancer).CheckBalance: asset ID %d is not a configured AccountBalancer"` or similar

server/market/balancer.go
```diff
@@ -0,0 +1,96 @@
...
+// for the balance required to fulfill new + existing orders and outstanding
+// redemptions.
+type DEXBalancer struct {
+	tunnels         map[string]MarketTunnel
+	assets          map[uint32]*backedBalancer
+	matchNegotiator MatchNegotiator
+}
+
+// NewDEXBalancer is a constructor for a DEXBalancer. Provided assets will
+// be filtered for those that are account-based. The matchNegotitator is
```
chappjc, line 22: matchNegotiator

server/market/orderrouter.go
```diff
@@ -97,6 +97,7 @@ type MarketTunnel interface {
 	// cancellation marked against the user, coins unlocked, and orderbook
 	// subscribers notified). See Unbook for details.
 	CheckUnfilled(assetID uint32, user account.AccountID) (unbooked []*order.LimitOrder)
+	AccountPending(acctAddr string, assetID uint32) (qty, lots uint64, redeems int)
```
chappjc, line 100: doc

server/market/market.go
```diff
@@ -1037,6 +1155,94 @@ func (m *Market) CheckUnfilled(assetID uint32, user account.AccountID) (unbooked
 	return m.checkUnfilledOrders(assetID, unfilled)
 }
 
+// AccountPending sums the orders quantities that pay to or from the specified
+// account address.
```
chappjc, line 1159: Could stand to describe `redeems int` as well.

server/dex/dex.go
```diff
@@ -587,11 +587,37 @@ func NewDEX(ctx context.Context, cfg *DexConf) (*DEX, error) {
...
+	//
+	// The current behavior is to reject all orders for the market if the
+	// account balance is too low to support them all, though an algorithm could
+	// be developed to do reject only some orders, based on available funding.
+	//
+	// This pattern is only safe because the markets are not Run until after
+	// they are all instantiated, so we are synchronous in our use of the
+	// marketTunnels map.
+	marketTunnels := make(map[string]market.MarketTunnel, len(cfg.Markets))
+	dexBalancer := market.NewDEXBalancer(marketTunnels, backedAssets, swapper)
```
chappjc, line 607: Observations:
- DEXBalancer only needs an "PendingAccounter" that implements `AccountPending`, not a full `MarketTunnel` 
- Conversely, `AccountPending` does not need to be a `MarketTunnel` interface method as other `MarketTunnel` consumers don't use it.
- Alternative to this approach of providing an empty map to `NewDEXBalancer` because `NewMarket` needs the `DEXBalancer` and then filling it in, we could have `NewMarket` register with `DEXBalancer`, which would only need to be constructed with the `backedAssets` and `swapper`.

Just some thoughts.

server/market/market.go
buck54321: It was actually just moved from below.

server/market/orderrouter.go
```diff
@@ -135,14 +136,19 @@ type FeeSource interface {
 	LastRate(assetID uint32) (feeRate uint64)
 }
 
+type MatchNegotiator interface {
+	AccountStats(acctAddr string, assetID uint32) (qty, swaps uint64, redeems int)
+}
```
chappjc, line 141: Maybe define this in balancer.go instead.  None of the other market types care about this.

I think balancer.go could define both this and a `PendingAccounter` that I proposed in another comment.

chappjc (changes requested):

server/market/market.go
chappjc: Ayyyy, it's been there since launch. Derp. https://github.com/decred/dcrdex/commit/6d80ad18914ff6803e2d3669edbd586ca6ad6dae

server/market/orderrouter.go
```diff
@@ -518,10 +398,112 @@ func (r *OrderRouter) handleMarket(user account.AccountID, msg *msgjson.Message)
...
+	fundingBalancer, isAccountFunded := assets.funding.Backend.(asset.AccountBalancer)
+	if isAccountFunded {
+		// Validate that the coins are correct for an account-based-asset-funded
+		// order. There should be 1 coin, 1 sig, 1 pubkey, and no redeem script.
+		if len(coins) != 1 {
+			log.Info("user %s submitted an %s-funded order with %d coin IDs", user, assets.funding.Symbol, len(coins))
+			return msgjson.NewError(msgjson.OrderParameterError, "account-type asset funding requires exactly one coin ID")
+		}
+		acctProof := coins[0]
+		if len(acctProof.PubKeys) != 1 || len(acctProof.Sigs) != 1 || len(acctProof.Redeem) > 0 {
```
chappjc, line 445: Working through things here:

- The "coin ID" in both `acctProof.ID` and `trade.Coins[0]` are equal, as evident in the callers, `handleMarket` and `handleLimit` who populate the `order.Trade.Coins` from each `msgjson.{Market,Limit}Order.Trade.Coins[i].ID`
- These coin ID encodings deviate from the new transaction id/hash convention for coin IDs being established in https://github.com/decred/dcrdex/pull/1307.  Hence `ValidateCoinID` cannot be used on them.
- They are also the same as `order.(*Trade).FromAccount` before hex string encoding.
- `order.(*Trade).ToAccount` is the same as `order.Trade.Address` (`order.(*Trade).SwapAddress`)

Mostly I want to be sure this is all the intended design and I'm not missing anything.

dex/order/order.go
```diff
@@ -473,6 +473,42 @@ func (t *Trade) SwapAddress() string {
...
 }
 
+// FromAccount is the account that the order originates from. Only useful for
+// account-based assets. Use of this method assumes that account coin has
+// already been added.
+func (t *Trade) FromAccount() string {
+	if len(t.Coins) == 0 {
+		return "no coins?"
+	}
+	return string(t.Coins[0])
```
chappjc, line 483: Shouldn't this be `t.Coins[0].String()`?

buck54321: > The account address should be utf-8 encoded

I've chosen to use the utf-8 encoding of the address hex string rather than the the address bytes so that we can do `string(coinID)` anywhere, regardless of whether we have a `msgjson.CoinID` or a `[]byte`. The alternative would require modifying field types in a number of places, if I remember correctly.

server/market/orderrouter.go
buck54321: That mostly seems right. The term "coin ID" is so general now as to be confusing. The definition of coin ID need to be evaluated separately for each message and asset. 

`ValidateCoinID` is in a weird place right now, I guess. It's used for funding/swap/redeem coins for utxo-based assets, but only for swap/redeem coins for acct-based. I wouldn't mind breaking that method into as many as 3 separate methods, `ValidateSwapCoin`, `ValidateRedeemCoin`, and `ValidateFundingCoin` (`OutputTracker`-only). We could also combine the swap and redeem methods into something like `ValidateTradeCoin` for now. Or we could just leave it. It's all terrible.

dex/order/order.go
chappjc: OK, I see. So the client when setting this will basically do `.Coins[i] = []byte(addressStr)`?

server/market/orderrouter.go
chappjc: Let's solve with docs for now rather than end up with a ton of new methods and wrappers and stuff

dex/order/order.go
chappjc: Actually, can you specify `account address should be utf-8 encoded`?  Is the address prior to casting to `[]byte` obtained from `common.Address.String()` or `common.Address.Hex()` or something else?

buck54321: Uh huh. That's what I was going for. I can doc this here.

buck54321: I think our messages crossed there, but here's what I'm looking at

```
// The coin ID should be the UTF-8 encoded address string, not the address
// byte-array. t.Coins[0] = []byte(addrStr).
```

chappjc: K, that's what I figured, but my question now is where does `addrStr` come from. https://github.com/decred/dcrdex/pull/1293#discussion_r760601794
The canonical encoding, hex with a 0x prefix, or just bare hex, lower case, upper, etc?

buck54321: I see. Yeah, we'll have to standardize that. Either of your suggestions would be fine. We could define a `dex/networks/eth` function, `func FundingCoinID(string) []byte`, if not just to enforce a standardization and provide documentation.

chappjc: The bulletproof approach I think would be to use the `common.Address[:]` bytes, and have `FromAccount` return bytes and the `AccountBalancer` take bytes too for the account to check.  I sense you thought through that but I'm not clear on the complications with that.

server/asset/common.go
```diff
@@ -92,6 +95,9 @@ type OutputTracker interface {
 type AccountBalancer interface {
 	// AccountBalance retrieves the current account balance.
 	AccountBalance(addr string) (uint64, error)
+	// ValidateSignature checks that the pubkey is correct for the address and
+	// that the signature shows ownership of the associated private key.
+	ValidateSignature(addr string, pubkey, msg, sig []byte) error
```
buck54321, line 100: Thinking of just updating the docs here to direct enforcement of an address standard.

```
// ValidateSignature checks that the pubkey is correct for the address and
// that the signature shows ownership of the associated private key. The
// address string will be validated against a strict standard format. For
// example, ethereum address should have the "0x" prefix and be lower-case.
// Any other format should return an error.
```

chappjc: `ValidateSignature`, being an ETH backend method, shouldn't have trouble with any of the normal string encodings.

`common.HexToAddress` handles all valid variants regardless of case or 0x prefix.
https://pkg.go.dev/github.com/ethereum/go-ethereum/common#HexToAddress

buck54321: But this string encoding is used as a unique ID throughout dcrdex.

chappjc: Oh, it certainly needs to be standardized for all the other uses, yes, I just don't think `ValidateSignature` is the place where it matters and hence not the best place to doc the standard we pick.

buck54321: `ValidateSignature` is run on all incoming orders. It seems like a great place to validate the signature, no?

chappjc: Yes, to validate the signature, just not to document the expected address format IMO.

buck54321: I'm not sure we're on the same page. I'm trying to instruct asset backend developers to validate the address string against a strict stadard. I'm just using ethereum as an example.

buck54321: This is under the assumption that we use the scheme defined in this PR, which is that the address is either the `msgjson.Trade.Address` or `string(msgjson.Trade.Coins[0])`.

chappjc: Alright, well since order router and ValidateSignature are the initial validation points for the address string, I suppose have it be strict about the encoding, more strict than it itself would require?

buck54321: That's what I'm trying to get at. Can improve the language, it seems.

chappjc: Another approach would be for `ValidateSignature` to return a canonicalized address string after processing whatever string came in via the `msgjson.Trade` fields.

The fact that `msgjson.Trade.Address` has always been a string and that's hard to change now kind means we have to have stick with a string for the other address, but @davecgh did have a nice suggestion if we had bytes that could be put in a type like `type Address interface { fmt.Stringer; Bytes() []byte }`, but alas the issue is upstream of that with the communications type committing us to a string.  So I think we have to work with it as-is.

chappjc (changes requested):

server/market/orderrouter.go
```diff
@@ -567,25 +552,29 @@ func (r *OrderRouter) handleMarket(user account.AccountID, msg *msgjson.Message)
...
+					midGap = tunnel.RateStep()
+				}
+				buyBuffer := tunnel.MarketBuyBuffer()
+				lotWithBuffer := uint64(float64(lotSize) * buyBuffer)
+				minReq := matcher.BaseToQuote(midGap, lotWithBuffer)
+				if trade.Quantity < minReq {
+					errStr := fmt.Sprintf("order quantity does not satisfy market buy buffer. %d < %d. midGap = %d", trade.Quantity, minReq, midGap)
+					return false, msgjson.NewError(msgjson.FundingError, errStr)
+				}
+				reqVal = calc.RequiredOrderFunds(minReq, uint64(spendSize), 1, &assets.quote.Asset)
```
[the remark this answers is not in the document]
buck54321: Yeah. Going back to this convo.

image

server/swap/swap.go
```diff
@@ -347,26 +403,88 @@ func (s *Swapper) UserSwappingAmt(user account.AccountID, base, quote uint32) (a
...
+		} else if mt.Status < redeemedStatus {
+			p.redeems++
+		}
+	}
+}
+
+// AccountStats is part of the MatchNegotiator interface to report in-process
+// match information for a asset account address.
+func (s *Swapper) AccountStats(acctAddr string, assetID uint32) (qty, swaps uint64, redeems int) {
+	stats := newPendingAccountStats(acctAddr, assetID)
```
chappjc, line 455: So these three stats values returned describe remaining transactions?  As in `redeems` is a count of redeem transactions that are expected to still take place for active matches, and similar for `swaps` (could that be `inits` to clarify?).
This part is reasonable clear from the `AccountStats interface` docs, but `qty` could be clearer there imo.  `qty` is the amount that will leave this asset-account when said inits/swaps are sent.

server/market/balancer.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	// Add in-process swaps.
+	newQty, newLots, newRedeems := b.matchNegotiator.AccountStats(acctAddr, assetID)
+	lots += newLots
+	qty += newQty
+	redeems += newRedeems
+
+	assetInfo := backedAsset.assetInfo
+	redeemCosts := uint64(redeems) * assetInfo.RedeemSize * assetInfo.MaxFeeRate
```
chappjc, line 98: Since `MaxFeeRate` isn't actually required for redeems, I think this deserves a comment that this is a conservatively high estimate of possible redeem costs.

server/market/market.go
```diff
@@ -74,6 +74,10 @@ type FeeFetcher interface {
 	MaxFeeRate() uint64
 }
 
+type Balancer interface {
```
chappjc, line 77: Doc or maybe even unexport `Balancer`?  It's internal to the `server/market` package anyway, only an interface because a concrete `DEXBalancer` is a headache to create, right?

chappjc (changes requested):
Almost certainly my last review.  Mainly nits at this point.

server/market/orderrouter.go
chappjc: I'm even thinking DBG for these, but we can keep INF for now I guess.  Anything that a user can do at will with little cost shouldn't be too chatty in the logs.

server/market/market.go
chappjc: Oh, no it's used in the NewMarket constructor from server/dex.... OK, doc I guess

buck54321:
@chappjc I started moving `accountTracker` to `Book`, but it fell apart because the `Remove` method only takes and order ID and the `RemoveUserOrders` method only takes a DEX `AccountID`, so `accountTracker` would need to add removal methods and possibly indices for both, which would largely mimic the logic already present in `OrderPQ`'s methods. Leaving it seems like a better option to me.

chappjc:
> @chappjc I started moving `accountTracker` to `Book`, but it fell apart because the `Remove` method only takes and order ID and the `RemoveUserOrders` method only takes a DEX `AccountID`, so `accountTracker` would need to add removal methods and possibly indices for both, which would largely mimic the logic already present in `OrderPQ`'s methods. Leaving it seems like a better option to me.

Thanks for looking into it.  What led me to suggest the change was because it looks like the returns from the `OrderPQ` methods `Remove` and `RemoveUserOrders` both return any `*order.LimitOrder`s that were removed, while `Insert` says if it was added.

Sorry if I'm still missing a detail, but roughly here's what I figure would work to keep the `OrderPQ` out of it:

```go
func (b *Book) Insert(o *order.LimitOrder) bool {
	if o.Quantity%b.lotSize != 0 {
		log.Warnf("(*Book).Insert: Refusing to insert an order with a quantity that is not a multiple of lot size.")
		return false
	}
	b.mtx.Lock()
	defer b.mtx.Unlock()
	if o.Sell {
		if b.sells.Insert(o) {
			b.acctTracker.add(o)
			return true
		}
		return false
	}
	if b.buys.Insert(o) {
		b.acctTracker.add(o)
		return true
	}
	return false
}

func (b *Book) Remove(oid order.OrderID) (*order.LimitOrder, bool) {
	b.mtx.Lock()
	defer b.mtx.Unlock()
	if removed, ok := b.sells.RemoveOrderID(oid); ok {
		b.acctTracker.remove(removed)
		return removed, true
	}
	if removed, ok := b.buys.RemoveOrderID(oid); ok {
		b.acctTracker.remove(removed)
		return removed, true
	}
	return nil, false
}

func (b *Book) RemoveUserOrders(user account.AccountID) (removedBuys, removedSells []*order.LimitOrder) {
	b.mtx.Lock()
	defer b.mtx.Unlock()
	removedBuys = b.buys.RemoveUserOrders(user)
	for _, lo := range removedBuys {
		b.acctTracker.remove(lo)
	}
	removedSells = b.sells.RemoveUserOrders(user)
	for _, lo := range removedSells {
		b.acctTracker.remove(lo)
	}
	return
}

func (b *Book) IterateBaseAccount(acctAddr string, f func(lo *order.LimitOrder)) {
	b.mtx.RLock()
	defer b.mtx.RUnlock()
	b.acctTracker.iterateBaseAccount(acctAddr, f)
}

func (b *Book) IterateQuoteAccount(acctAddr string, f func(lo *order.LimitOrder)) {
	b.mtx.RLock()
	defer b.mtx.RUnlock()
	b.acctTracker.iterateQuoteAccount(acctAddr, f)
}
```

Perhaps I'm incorrectly lumping the two accountTracker's together?

chappjc:
I made a mess of the tests, but it seems to work: https://github.com/decred/dcrdex/commit/4264adf396b54e0e6df7bc4b6f6a1d78ef6d94b8
Is it wrong to have a single `accountTracker`?

chappjc:
I said I was gonna let that sit a while longer, but we've been over it plenty and we have a backlog of ETH PRs stacking up.

PR #1296, opened 2021-11-15
client/asset/btc: fix test race

buck54321:
Protect `getBestBlockHashErr` with the `blockchainMtx` since it's accessed asynchronously from the block monitor loop via `getBlockHeader` -> `BestBlock`.

### DIFF
--- a/client/asset/btc/btc_test.go
+++ b/client/asset/btc/btc_test.go
@@ -133,27 +133,27 @@ type testData struct {
 	signFunc      func(*wire.MsgTx)
 	signMsgFunc   func([]json.RawMessage) (json.RawMessage, error)
 
-	blockchainMtx     sync.RWMutex
-	verboseBlocks     map[string]*msgBlockWithHeight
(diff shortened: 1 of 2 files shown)


chappjc:
Resolving https://github.com/decred/dcrdex/issues/1292 looks like

client/asset/btc/btc_test.go
```diff
@@ -2573,14 +2576,18 @@ func testSyncStatus(t *testing.T, segwit bool, walletType string) {
 	}
 
 	node.getBlockchainInfoErr = tErr // rpc
-	node.getBestBlockHashErr = tErr  // spv BestBlock()
-	delete(node.mainchain, 100)      // force spv to BestBlock() with no wallet block
+	node.blockchainMtx.Lock()
+	node.getBestBlockHashErr = tErr // spv BestBlock()
```
chappjc, line 2580: I thought about just not testing this with SPV, but oh well.

chappjc (approved):

PR #1301, opened 2021-11-16
client/eth: refactor eth client

buck54321:
The client/asset/eth package was using go-ethereum like an RPC client. That's
not crazy, considering we have still an `*rpc.Client`, but in reality,
there is no RPC server, and the client is just a shortcut to RPC handlers
that call functions that we already have access to through e.g.
`*les.LightEthereum` or `*keystore.KeyStore`. This PR removes all
`CallContext` calls in favor of calling the underlying methods directly.
There are probably more `*rpc.Client` calls we could bypass as well.

Moves some common types to dex/networks/eth. As a general rule, if
the client/asset/ imports from server/asset, we should just move it
to dex/networks.

Creates the `Contractor` interface, which serves as a clear boundary
between dex-space and abigen-space. Implements contract versions.

Switches to EIP1559 transactions and removes hard-coded base fee rate
from harness. Due to this change, some live tests are still busted because
balance changes need to be fixes. Leaving in draft until that's done.

I realize this a big change in an area where we have a lot of hands right now,
but I'd rather do it earlier than later.


client/asset/eth/nodeclient.go
```diff
@@ -0,0 +1,373 @@
...
+}
+
+// sendTransaction uses a raw request to send tx.
+func (n *nodeClient) sendTransaction(ctx context.Context, txOpts *bind.TransactOpts,
+	to common.Address, data []byte) (*types.Transaction, error) {
+
+	nonce, err := n.ec.NonceAt(ctx, n.creds.addr, nil)
+	if err != nil {
+		return nil, fmt.Errorf("error getting nonce: %v", err)
+	}
```
buck54321, line 207: It's not impossible that this is insufficient to guarantee a unique nonce. I haven't verified that this is incremented before the transaction is mined.

client/asset/eth/nodeclient.go
```diff
@@ -313,38 +332,36 @@ func (n *nodeClient) refund(txOpts *bind.TransactOpts, secretHash [32]byte, serv
...
 
-	fmt.Println("--base, tip", base, tip)
-
 	return b, weiToGwei(tip), nil
 }
 
 // getCodeAt retrieves the runtime bytecode at a certain address.
-func (n *nodeClient) getCodeAt(ctx context.Context, contractAddr *common.Address) ([]byte, error) {
-	bn, err := n.ec.BlockNumber(ctx)
+func (n *nodeClient) getCodeAt(ctx context.Context, contractAddr common.Address) ([]byte, error) {
```
buck54321, line 355: This one's busted too and I don't know why yet. It's busted if I use the existing method too, but that's not ruling out some other mistake I made.

client/asset/eth/eth.go
```diff
@@ -916,3 +902,28 @@ func (eth *ExchangeWallet) checkForNewBlocks() {
 		prevTip.Hash(), newTip.NumberU64(), newTip.Hash())
 	go eth.tipChange(nil)
 }
+
+func ethToGwei(v uint64) uint64 {
```
martonp, line 906: These conversion functions could go in the common code.

client/asset/eth/nodeclient.go
```diff
@@ -0,0 +1,390 @@
...
+	return n.leth.ApiBackend.SyncProgress()
+}
+
+// peers returns connected peers.
+func (n *nodeClient) peers(ctx context.Context) []*p2p.Peer {
+	return n.p2pSrv.Peers()
+}
+
+// swap gets a swap keyed by secretHash in the contract.
+func (n *nodeClient) swap(ctx context.Context, secretHash [32]byte, serverVer uint32) (swap *dexeth.SwapState, err error) {
```
martonp, line 257: `contractVer` would be more accurate then `serverVer`.

client/asset/eth/nodeclient.go
```diff
@@ -0,0 +1,390 @@
...
+var _ ethFetcher = (*nodeClient)(nil)
+
+// nodeClient satisfies the ethFetcher interface. Do not use until Connect is
+// called.
+type nodeClient struct {
+	net    dex.Network
+	log    dex.Logger
+	creds  *accountCredentials
+	p2pSrv *p2p.Server
+	// c is a direct client for raw calls.
```
martonp, line 44: c -> cl

martonp (commented):
This looks great.
I say let's get the #1262 and #1274 in first since that would finalize the contract, and then put this one before any other ones.

client/asset/eth/nodeclient.go
martonp: Works fine for me.. maybe try restarting the eth harness?

dex/networks/eth/params.go
```diff
@@ -5,10 +5,31 @@ package eth
...
+	// contract methods are called it is difficult to descern where gas is
+	// used. The formal declaration of what costs how much can be found at
+	// https://ethereum.github.io/yellowpaper/paper.pdf in Appendix G.
+	// The current value here is an appoximation based on tests.
+	//
+	// TODO: When the contract is solidified, break down evm functions
+	// called and the gas used for each. (◍•﹏•)
+	InitGas = 158000 // gas
+
+	RedeemGas = 63000 // gas
```
martonp, line 24: Since the contracts are versioned now, these gas estimates will also need to be versioned.

client/asset/eth/nodeclient.go
buck54321: Hmm. I'm still failing. What version of geth are you running?

buck54321: Nevermind. I found the problem.

martonp: What was wrong?

client/asset/eth/eth.go
```diff
@@ -789,25 +782,20 @@ func (*ExchangeWallet) SwapConfirmations(ctx context.Context, id dex.Bytes, cont
...
 //
 // TODO: Return the asset.Coin.
 // TODO: Subtract fees from the value.
-func (eth *ExchangeWallet) Withdraw(addr string, value, feeSuggestion uint64) (asset.Coin, error) {
-	bigVal := big.NewInt(0).SetUint64(value)
-	gweiFactorBig := big.NewInt(srveth.GweiFactor)
-	_, err := eth.sendToAddr(common.HexToAddress(addr),
-		bigVal.Mul(bigVal, gweiFactorBig), big.NewInt(0).SetUint64(defaultGasFee))
+func (eth *ExchangeWallet) Withdraw(addr string, value, _ uint64) (asset.Coin, error) {
+	_, err := eth.sendToAddr(common.HexToAddress(addr), value)
```
buck54321, line 798: We should revisit this at some point. If a user tries to withdraw their entire balance, it will fail and they'll lose the tx fees, I believe.

chappjc: Yikes, how does one sweep an ETH wallet completely, leaving nothing left if gas doesn't end up whatever was allocated to gas?

client/asset/eth/nodeclient_harness_test.go
```diff
@@ -694,127 +687,56 @@ func TestInitiateBatch(t *testing.T) {
...
 			}
-			state := srveth.SwapState(swap.State)
-			if test.success && state != srveth.SSInitiated {
-				t.Fatalf("unexpected swap state for test %v: want %s got %s", test.name, srveth.SSInitiated, state)
+
+			fmt.Printf("--done checking swap %s \n\n", hex.EncodeToString(testSwap.SecretHash))
+
+			state := dexeth.SwapStep(swap.State)
+			if test.success && state != dexeth.SSInitiated {
+				t.Fatalf("unexpected swap state for test %v: want %s got %s", test.name, dexeth.SSInitiated, state)
```
buck54321, line 716: I cannot figure out why this isn't passing. I'm verifying the secret hashes match thoughout. There are no errors. I can see the transaction with `eth.getTransaction` afterwards, and I can see the secret hash in the tx data, but the `ETHSwapSwap` is completely zeroed.

client/asset/eth/eth.go
buck54321: Maybe the wallet won't allow the transaction to be sent the way I've described. I'm not certain.

chappjc: for ref: https://www.reddit.com/r/ethdev/comments/pits6s/sweep_all_eth_using_dynamic_fees_eip1559_and/
EIP-1559 does make this a headache if not impossible to do without leaving dust

buck54321:
Rebased and un-renamed the rpcclient*.go files back to the original names in an attempt to provide a usable diff for review. Will rename again later.

I'm still banging my head against the wall on the swap state issue, but I'm gonna step away for a minute. @martonp, if you have a sec to take a look, I'd really appreciate it.

client/asset/eth/eth.go
martonp: Haha, basically the way to sweep it is to give the dust to the miners.

For the original issue of the failing transaction, I would think that the node would have a check that refuses to submit a transaction where the `gasLimit*gasPrice + value > balance`, but I haven't tried it out yet.

client/asset/eth/nodeclient_harness_test.go
martonp: I've been trying to figure it out what's wrong.. haven't had any luck yet. 

I also saw that you removed the `addSignerToOpts` calls from the functions that submit transactions (`initiate`, `redeem`, etc.) and you're calling them from the test code, but in some cases you forgot to call it. What's wrong with this going into `intiate` and `redeem`?

buck54321: Here's a clue. 

https://github.com/decred/dcrdex/compare/master...buck54321:fail-init-real-balance?expand=1

That fails too.

buck54321: It's failing for any value > 1 gwei

buck54321: Update: I've got everything passing except for `TestReplayAttack`.

martonp: Nice, what was wrong? I was getting a headache from that.

dex/testing/eth/create-node.sh
```diff
@@ -82,7 +82,6 @@ if [ "${CHAIN_ADDRESS}" != "_" ]; then
 
 [Eth.Miner]
 Etherbase = "0x${CHAIN_ADDRESS}"
-GasPrice = 81000000000
```
JoeGruffins, line 85: Yeah, sorry I couldn't answer you somewhere where you asked about this, I don't know. I'm not up to speed on the base fee, thing. The miners can set the gas price over time is all I think I know. Also, I think geth complained without setting this, maybe it doesn't now.

client/asset/eth/nodeclient_harness_test.go
buck54321: I think I was handling a `*big.Int` wrong. Not making a copy when I should have or something.

JoeGruffins:
So, the reentry test, it doesn't look like it's communicating with the contract. The balances aren't the simnet's contract's balance, and the initiations aren't being sent to the contract. Or, did you just fix it?

client/asset/eth/rpcclient.go
```diff
@@ -8,213 +8,268 @@ package eth
...
+
 // balance gets the current balance of an address.
-func (c *rpcclient) balance(ctx context.Context, addr *common.Address) (*big.Int, error) {
-	return c.ec.BalanceAt(ctx, *addr, nil)
+func (n *nodeClient) balance(ctx context.Context, addr *common.Address) (*big.Int, error) {
+	state, _, err := n.pendingState(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return state.GetBalance(n.creds.addr), nil
```
JoeGruffins, line 161: This should just be addr?

buck54321:
> So, the reentry test, it doesn't look like it's communicating with the contract. The balances aren't the simnet's contract's balance, and the initiations aren't being sent to the contract. Or, did you just fix it?

Thanks, @JoeGruffins. The contract address must've been part of it. It's partly passing now, but still seeing an unexpected change in the contract address. Still missing something.

JoeGruffins:
Passing for me. Please see the rpcclient.Balance comment I made. `*addr`

client/asset/eth/nodeclient.go
buck54321: I think I've forgotten now. Sorry I missed this question.

client/asset/eth/rpcclient.go
buck54321: Yep. Fixed now. Thanks.

client/asset/eth/rpcclient.go
```diff
@@ -235,67 +342,140 @@ func (c *rpcclient) signData(addr common.Address, data []byte) ([]byte, error) {
...
 }
 
-// redeem redeems a list of swap contracts. The redeemer will be the account at
-// txOpts.From. Any on-chain failure, such as this secret not matching the hash,
-// will not cause this to error.
-func (c *rpcclient) redeem(txOpts *bind.TransactOpts, netID int64, redemptions []swap.ETHSwapRedemption) (*types.Transaction, error) {
-	err := c.addSignerToOpts(txOpts, netID)
-	if err != nil {
+// estimateGas checks the amount of gas that is used for a function call.
+func (n *nodeClient) estimateRedeemGas(ctx context.Context, secrets [][32]byte, contractVer uint32) (gas uint64, err error) {
```
JoeGruffins, line 382: doc to `estimateRedeemGas`

dex/networks/eth/contractor.go
```diff
@@ -0,0 +1,192 @@
...
+		return nil, err
+	}
+
+	return &SwapState{
+		BlockHeight: state.InitBlockNumber.Uint64(),
+		LockTime:    time.Unix(state.RefundBlockTimestamp.Int64(), 0),
+		Secret:      state.Secret,
+		Initiator:   state.Initiator,
+		Participant: state.Participant,
+		Value:       state.Value.Uint64(),
```
JoeGruffins, line 134: I think should convert to gwei here.

dex/networks/eth/contractor.go
```diff
@@ -0,0 +1,192 @@
...
+func (c *contractorV0) IsRedeemable(secretHash, secret [32]byte) (bool, error) {
+	return c.ETHSwap.IsRedeemable(&bind.CallOpts{From: c.acctAddr}, secretHash, secret)
+}
+
+func (c *contractorV0) EstimateRedeemGas(ctx context.Context, secrets [][32]byte) (uint64, error) {
+	redemps := make([]ETHSwapRedemption, 0, len(secrets))
+	for _, secret := range secrets {
+		redemps = append(redemps, ETHSwapRedemption{
+			Secret:     secret,
+			SecretHash: sha256.Sum256(secret[:]),
```
JoeGruffins, line 148: These couldn't be fake? I think there was some discussion about this but I forget. EstimateInit is using fake secret hashes.

client/asset/eth/eth.go
```diff
@@ -919,3 +812,19 @@ func (eth *ExchangeWallet) checkForNewBlocks() {
 		prevTip.Hash(), newTip.NumberU64(), newTip.Hash())
 	go eth.tipChange(nil)
 }
+
+func ethToGwei(v uint64) uint64 {
```
JoeGruffins, line 816: It looks like there were added so this section could be removed.

JoeGruffins (commented):
The Contractor interface looks good.

I'm not completely sold on the rpcclient changes. I think it was less work for us before. But it's fine I guess.

client/asset/eth/eth.go
```diff
@@ -131,42 +131,29 @@ func (d *Driver) Create(params *asset.CreateWalletParams) error {
...
-	initiate(txOpts *bind.TransactOpts, netID int64, initiations []dexeth.ETHSwapInitiation) (*types.Transaction, error)
-	estimateGas(ctx context.Context, msg ethereum.CallMsg) (uint64, error)
-	lock(ctx context.Context, acct *accounts.Account) error
-	nodeInfo(ctx context.Context) (*p2p.NodeInfo, error)
-	pendingTransactions(ctx context.Context) ([]*types.Transaction, error)
-	peers(ctx context.Context) ([]*p2p.PeerInfo, error)
-	transactionReceipt(ctx context.Context, txHash common.Hash) (*types.Receipt, error)
-	sendTransaction(ctx context.Context, tx map[string]string) (common.Hash, error)
+	connect(ctx context.Context) error
+	initiate(ctx context.Context, contracts []*asset.Contract, maxFeeRate uint64, contractVer uint32) (*types.Transaction, error)
```
buck54321, line 144: The `contractVer` doesn't need to map precisely to the `dex.Asset.Version`, but as long as we make a rule that a change in contract always accompanies a bump in `dex.Asset.Version`, then we can find a scheme to get the contract version from the asset version and we don't need any new data.

client/asset/eth/rpcclient.go
```diff
@@ -8,213 +8,320 @@ package eth
...
 // called.
-type rpcclient struct {
-	// c is a direct client for raw calls.
-	c *rpc.Client
-	// ec wraps the client with some useful calls.
-	ec *ethclient.Client
-	n  *node.Node
-	// es is a wrapper for contract calls.
-	es *swap.ETHSwap
+type nodeClient struct {
```
chappjc, line 47: Wanna `git mv` to rename the file too?

client/asset/eth/rpcclient.go
```diff
@@ -8,213 +8,320 @@ package eth
...
+	net         dex.Network
+	log         dex.Logger
+	creds       *accountCredentials
+	p2pSrv      *p2p.Server
+	ec          *ethclient.Client
+	node        *node.Node
+	leth        *les.LightEthereum
+	chainID     *big.Int
+	contractors map[uint32]dexeth.Contractor
+	es          *dexeth.ETHSwap
```
chappjc, line 57: relic?

client/asset/eth/rpcclient.go
```diff
@@ -8,213 +8,320 @@ package eth
...
-	c.es, err = swap.NewETHSwap(*contractAddr, c.ec)
-	if err != nil {
-		return fmt.Errorf("unable to find swap contract: %v", err)
+	n.ec = ethclient.NewClient(client)
+	n.p2pSrv = n.node.Server()
+	if n.p2pSrv == nil {
+		return fmt.Errorf("no *p2p.Server")
 	}
+
+	n.es, _ = dexeth.NewETHSwap(dexeth.ContractAddresses[0][dex.Simnet], n.ec)
```
chappjc, line 108: Relic? (`NewETHSwap` is currently specific to v0 contract API and I think this is all hidden in `Contractor` now, plus Simnet hard coded)

client/asset/eth/rpcclient.go
```diff
@@ -8,213 +8,320 @@ package eth
...
-	bn, err := c.ec.BlockNumber(ctx)
-	if err != nil {
-		return nil, err
-	}
-	header, err := c.ec.HeaderByNumber(ctx, big.NewInt(int64(bn)))
-	if err != nil {
-		return nil, err
-	}
-	return header, nil
+func (n *nodeClient) bestHeader(ctx context.Context) (*types.Header, error) {
```
chappjc, line 153: Plans for `Header` return, or remove?

client/asset/eth/rpcclient.go
```diff
@@ -8,213 +8,320 @@ package eth
...
-// Check that rpcclient satisfies the ethFetcher interface.
-var _ ethFetcher = (*rpcclient)(nil)
+// Check that nodeClient satisfies the ethFetcher interface.
+var (
+	// https://github.com/ethereum/go-ethereum/blob/16341e05636fd088aa04a27fca6dc5cda5dbab8f/eth/backend.go#L110-L113
+	// ultimately results in a minimum fee rate by the filter applied at
+	// https://github.com/ethereum/go-ethereum/blob/4ebeca19d739a243dc0549bcaf014946cde95c4f/core/tx_pool.go#L626
+	minGasPrice = ethconfig.Defaults.Miner.GasPrice
+
+	_ ethFetcher = (*nodeClient)(nil)
```
chappjc, line 42: `// Check that nodeClient satisfies the ethFetcher interface.` comment here instead.

dex/networks/eth/contractor.go
```diff
@@ -0,0 +1,192 @@
...
+	0: {
+		dex.Mainnet: common.Address{},
+		dex.Simnet:  common.Address{},
+		dex.Testnet: common.Address{},
+	},
+}
+
+type ContractorConstructor func(net dex.Network, addr common.Address, ec *ethclient.Client) (Contractor, error)
+
+// contractorv0 is the Contractor for contract version 0.
```
chappjc, line 51: contractorv0 -> contractorV0

dex/networks/eth/contractor.go
```diff
@@ -0,0 +1,192 @@
...
+	return c.ec.EstimateGas(ctx, ethereum.CallMsg{
+		From:  c.acctAddr,
+		To:    &c.contractAddr,
+		Value: big.NewInt(int64(n)),
+		Gas:   0,
+		Data:  data,
+	})
+}
+
+var ContractorContstructors = map[uint32]ContractorConstructor{
```
chappjc, line 190: `ContractorConstructors` (spelling)

chappjc (changes requested):
A few quick comments, will have a proper review later today.

dex/networks/eth/contractor.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/ethclient"
+)
+
+// Contractor is a translation layer between the abigen bindings and the DEX app.
+// The intention is that if a new contract is implemented, the Contractor
+// interface itself will not require any updates.
+type Contractor interface {
```
chappjc, line 31: In general use for Contractor server-side?  why not in client/asset/eth?

dex/networks/eth/contractor.go
```diff
@@ -0,0 +1,192 @@
...
+		dex.Testnet: common.Address{},
+	},
+}
+
+type ContractorConstructor func(net dex.Network, addr common.Address, ec *ethclient.Client) (Contractor, error)
+
+// contractorv0 is the Contractor for contract version 0.
+// Redeem and Refund methods of ETHSwap already have suitable return types.
+type contractorV0 struct {
+	*ETHSwap
```
chappjc, line 54: We should save the following suggestion for another day, but since `ETHSwap` is the v0 contract API compiled from it's solidity, we'll probably want to move contract.go -> contracts/v0.go or contracts/apis/v0.go

client/asset/eth/rpcclient.go
chappjc: Hmm, I think I meant to comment on `pendingState`.  One of these had their callers always discarding the header return, but clearly it's not `bestHeader`

chappjc (changes requested):
A bit more.  Sorry I'm reviewing this in 20 minute stints.

client/asset/eth/rpcclient.go
```diff
@@ -8,213 +8,320 @@ package eth
...
 }
 
 // sendTransaction uses a raw request to send tx.
-func (c *rpcclient) sendTransaction(ctx context.Context, tx map[string]string) (common.Hash, error) {
-	res := common.Hash{}
-	err := c.c.CallContext(ctx, &res, "eth_sendTransaction", tx)
+func (n *nodeClient) sendTransaction(ctx context.Context, txOpts *bind.TransactOpts,
+	to common.Address, data []byte) (*types.Transaction, error) {
+
+	nonce, err := n.leth.ApiBackend.GetPoolNonce(ctx, n.creds.addr)
```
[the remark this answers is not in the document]
buck54321: This appears to be what would have happened with the old method

https://github.com/ethereum/go-ethereum/blob/c10a0a62c3537fbcf899358e3c3d6be9507fa18c/internal/ethapi/transaction_args.go#L132-L138

Probably worth noting that internally, both `PrivateAccountAPI` and `PublicTransactionPoolAPI` do it the same way, but they use a mutex ('nonceLock') to sequence sign-send operations when a nonce is being generated with `setDefaults`. It might be worth adding some mutex protection to methods that generate transaction for the same reasons.

client/asset/eth/rpcclient.go
```diff
@@ -8,213 +8,320 @@ package eth
...
+		Nonce:     nonce,
+		Gas:       txOpts.GasLimit,
+		GasFeeCap: txOpts.GasFeeCap,
+		GasTipCap: txOpts.GasTipCap,
+		Value:     txOpts.Value,
+		Data:      data,
+	}), n.chainID)
+
+	if err != nil {
+		return nil, fmt.Errorf("signing error: %v", err)
```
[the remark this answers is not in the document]
buck54321: See other comment about nonce too, but go-ethereum seems to operate with the assumption that `GetPoolNonce` is accurate as soon as `SendTx` returns.

dex/networks/eth/contractor.go
buck54321: Apparently not.

dex/networks/eth/contractor.go
buck54321: Good question. The only abigen method we're using server-side is `Swap`, and the server will only support one version at a time. I guess I should move it to **client/asset/eth** then.

dex/networks/eth/params.go
```diff
@@ -3,12 +3,131 @@
...
+// SSRedeemed indicates a successful swap where the participant was able to
+// redeem with the secret hash. SSRefunded indicates a failed swap, where the
+// initiating party refunded their coins after the locktime passed. A swap no
+// longer changes states after reaching SSRedeemed or SSRefunded.
+const (
+	// SSNone indicates that the swap is not initiated. This is the default
+	// state of a swap.
+	SSNone SwapStep = iota
+	// SSInitiated indicates that the swap has been initiated.
+	SSInitiated
```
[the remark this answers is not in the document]
buck54321: The `SwapState` is already a mirror of the `ETHSwapState`, which is the contract-specific struct. The `Contractor` is translating. `SwapStep` and `SwapState` are meant to defined for the needs of the DEX protocol, so in theory, shouldn't need modification with future contract changes. I wouldn't rule out that we do need to modify this stuff someday though, esp. `SwapState`, but it's a struct so we can always make it work, I think.

client/asset/eth/eth.go
```diff
@@ -791,26 +700,12 @@ func (*ExchangeWallet) SwapConfirmations(ctx context.Context, id dex.Bytes, cont
...
 //
 // TODO: Return the asset.Coin.
 // TODO: Subtract fees from the value.
-func (eth *ExchangeWallet) Withdraw(addr string, value, feeSuggestion uint64) (asset.Coin, error) {
-	bigVal := big.NewInt(0).SetUint64(value)
-	gweiFactorBig := big.NewInt(srveth.GweiFactor)
-	_, err := eth.sendToAddr(common.HexToAddress(addr),
-		bigVal.Mul(bigVal, gweiFactorBig), big.NewInt(0).SetUint64(defaultGasFee))
+func (eth *ExchangeWallet) Withdraw(addr string, value, _ uint64) (asset.Coin, error) {
+	_, err := eth.node.sendToAddr(eth.ctx, common.HexToAddress(addr), value)
```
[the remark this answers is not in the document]
buck54321: Continuation of [previous conversation](https://github.com/decred/dcrdex/pull/1301#discussion_r751337288), I think, but yes. We need to address this.

dex/networks/eth/contractor.go
chappjc: It was wrong with random bytes?  I would expect only the length to be important, not the content.

dex/networks/eth/params.go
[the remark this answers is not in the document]
chappjc: Understood that it's supposed to be a version-independent translation of `ETHSwapState`.  I'm just crossing my fingers that it  will always be sufficient.

chappjc (approved):
Looks good.  Holding out for https://github.com/decred/dcrdex/pull/1248, and for the outstanding reviewers to circle back.

dex/networks/eth/params.go
```diff
@@ -3,12 +3,131 @@
...
+	"math/big"
+	"time"
+
+	"decred.org/dcrdex/dex"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+const (
+	// GweiFactor is the amount of wei in one gwei. Eth balances are floored
+	// as gwei, or 1e9 wei. This is used in factoring.
```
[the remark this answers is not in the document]
JoeGruffins: > just mean it's used to convert units

yes

dex/networks/eth/params.go
```diff
@@ -3,12 +3,149 @@
...
+
+// GweiToWei converts uint64 Gwei to *big.Int Wei.
+func GweiToWei(v uint64) *big.Int {
+	return new(big.Int).Mul(big.NewInt(int64(v)), BigGweiFactor)
+}
+
+// GweiToWei converts *big.Int Wei to uint64 Gwei.
+func WeiToGwei(v *big.Int) uint64 {
+	return new(big.Int).Div(v, BigGweiFactor).Uint64()
+}
```
JoeGruffins, line 100: The original value could potentially be negative or overflow the returned uint64. There is also already a ToGwei in cointypes.go

JoeGruffins (approved):

martonp (approved):
LGTM

client/asset/eth/eth_test.go
```diff
@@ -1085,58 +941,9 @@ func TestSwap(t *testing.T) {
 					testName, receipt.Contract(), contract.SecretHash)
 			}
 
-			if !bytes.Equal(node.lastInitiation.hash.Bytes(), receipt.Coin().ID()) {
```
martonp, line 1088: Why did you remove these? Shouldn't we be checking that `initiate` is called with correct arguments?

chappjc: Good call. You probably see the PR, but I'll link for all: https://github.com/decred/dcrdex/pull/1318

PR #1302, opened 2021-11-17
eth/client: Redeem

martonp:
Part of #1154

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -145,7 +145,7 @@ type ethFetcher interface {
 	initiate(ctx context.Context, contracts []*asset.Contract, maxFeeRate uint64, contractVer uint32) (*types.Transaction, error)
 	shutdown()
 	syncProgress() ethereum.SyncProgress
-	redeem(txOpts *bind.TransactOpts, redemptions []*asset.Redemption, contractVer uint32) (*types.Transaction, error)
+	redeem(ctx context.Context, redemptions []*asset.Redemption, maxFeeRate uint64, contractVer uint32) (*types.Transaction, error)
 	refund(txOpts *bind.TransactOpts, secretHash [32]byte, contractVer uint32) (*types.Transaction, error)
 	swap(ctx context.Context, secretHash [32]byte, contractVer uint32) (*dexeth.SwapState, error)
 	lock() error
@@ -672,8 +672,38 @@ func (eth *ExchangeWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin
 
 // Redeem sends the redemption transaction, which may contain more than one
 // redemption.
-func (*ExchangeWallet) Redeem(form *asset.RedeemForm) ([]dex.Bytes, asset.Coin, uint64, error) {
-	return nil, nil, 0, asset.ErrNotImplemented
+func (eth *ExchangeWallet) Redeem(form *asset.RedeemForm) ([]dex.Bytes, asset.Coin, uint64, error) {
+	fail := func(err error) ([]dex.Bytes, asset.Coin, uint64, error) {
+		return nil, nil, 0, err
+	}
+
+	if len(form.Redemptions) == 0 {
+		return fail(errors.New("Redeem: must be called with at least 1 redemption"))
+	}
+
+	inputs := make([]dex.Bytes, 0, len(form.Redemptions))
+	var redeemedValue uint64
+	for _, redemption := range form.Redemptions {
+		var secretHash [32]byte
+		copy(secretHash[:], redemption.Spends.SecretHash)
+		swapData, err := eth.node.swap(eth.ctx, secretHash, form.AssetVersion)
+		if err != nil {
+			return nil, nil, 0, fmt.Errorf("Redeem: error finding swap state: %w", err)
+		}
+		redeemedValue += swapData.Value
+		inputs = append(inputs, redemption.Spends.Coin.ID())
+	}
+	outputCoin := eth.createAmountCoin(redeemedValue)
+	fundsRequired := dexeth.RedeemGas(len(form.Redemptions), form.AssetVersion) * form.FeeSuggestion
+
+	// TODO: make sure the amount we locked for redemption is enough to cover the gas
+	// fees.
+	_, err := eth.node.redeem(eth.ctx, form.Redemptions, form.FeeSuggestion, form.AssetVersion)
+	if err != nil {
+		return fail(fmt.Errorf("Redeem: redeem error: %w", err))
+	}
+
+	return inputs, outputCoin, fundsRequired, nil
 }
 
 // SignMessage signs the message with the private key associated with the
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -686,6 +686,9 @@ func (eth *ExchangeWallet) Redeem(form *asset.RedeemForm) ([]dex.Bytes, asset.Co
 	for _, redemption := range form.Redemptions {
 		var secretHash [32]byte
 		copy(secretHash[:], redemption.Spends.SecretHash)
+		if secretHash != sha256.Sum256(redemption.Secret) {
+			return fail(fmt.Errorf("Redeem: secretHash %x != sha256(%x)", secretHash, redemption.Secret))
+		}
 		swapData, err := eth.node.swap(eth.ctx, secretHash, form.AssetVersion)
 		if err != nil {
 			return nil, nil, 0, fmt.Errorf("Redeem: error finding swap state: %w", err)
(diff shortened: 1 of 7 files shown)


client/asset/eth/eth_test.go
```diff
@@ -134,10 +135,32 @@ func (n *testNode) initiate(ctx context.Context, contracts []*asset.Contract, ma
...
 	}), nil
 }
-func (n *testNode) redeem(opts *bind.TransactOpts, redemptions []*asset.Redemption, contractVer uint32) (*types.Transaction, error) {
-	return nil, nil
+
+func (n *testNode) redeem(ctx context.Context, redemptions []*asset.Redemption, maxFeeRate uint64, contractVer uint32) (*types.Transaction, error) {
+	if n.redeemErr != nil {
+		return nil, n.redeemErr
+	}
+	/*baseTx := &types.DynamicFeeTx{
```
buck54321, line 143: Did you want to leave these comments?

client/asset/eth/nodeclient.go
```diff
@@ -417,12 +417,13 @@ func (n *nodeClient) estimateRefundGas(ctx context.Context, secretHash [32]byte,
 	})
 }
 
-// redeem redeems a swap contract. The redeemer will be the account at txOpts.From.
-// Any on-chain failure, such as this secret not matching the hash, will not cause
-// this to error.
-func (n *nodeClient) redeem(txOpts *bind.TransactOpts, redemptions []*asset.Redemption, contractVer uint32) (tx *types.Transaction, err error) {
+// redeem redeems a swap contract. Any on-chain failure, such as this secret not
+// matching the hash, will not cause this to error.
```
buck54321, line 421: > such as this secret not matching the hash

We should check this in `Redeem` though.

client/asset/interface.go
```diff
@@ -335,6 +335,9 @@ type RedeemForm struct {
 	// suggestion in any way, but obviously fees that are too low may result in
 	// the redemption getting stuck in mempool.
 	FeeSuggestion uint64
+	// AssetVersion is the swap protocol version, which may indicate a specific
+	// contract or form of contract.
+	AssetVersion uint32
```
buck54321, line 340: This is good for now. Can rethink in or after #1320. 

Side note: It may be possible we don't need the `FromVersion` and `ToVersion` added to the `OrderMetaData` in #1301 after all.

client/asset/eth/eth.go
```diff
@@ -672,8 +672,38 @@ func (eth *ExchangeWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin
...
+
+	if len(form.Redemptions) == 0 {
+		return fail(errors.New("Redeem: must be called with at least 1 redemption"))
+	}
+
+	inputs := make([]dex.Bytes, 0, len(form.Redemptions))
+	var redeemedValue uint64
+	for _, redemption := range form.Redemptions {
+		var secretHash [32]byte
+		copy(secretHash[:], redemption.Spends.SecretHash)
```
buck54321, line 688: We should check the validity of the secret here.

buck54321 (changes requested):
Looks great. 

The live test `TestContract/testRedeem` is not passing.

```
--- FAIL: TestRedeem (9.37s)
    nodeclient_harness_test.go:953: ok before locktime: unexpected balance change: want 10999999818286 got 10999999757732, diff = 60554
```

martonp:
I got that too but it went away after another run.. I think it has to do with the issue with a fresh harness.

client/asset/eth/eth_test.go
martonp: No.. I'll take them out of initiate also.

buck54321 (approved):

client/asset/eth/eth.go
```diff
@@ -672,8 +672,41 @@ func (eth *ExchangeWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin
...
+		swapData, err := eth.node.swap(eth.ctx, secretHash, form.AssetVersion)
+		if err != nil {
+			return nil, nil, 0, fmt.Errorf("Redeem: error finding swap state: %w", err)
+		}
+		redeemedValue += swapData.Value
+		inputs = append(inputs, redemption.Spends.Coin.ID())
+	}
+	outputCoin := eth.createAmountCoin(redeemedValue)
+	fundsRequired := dexeth.RedeemGas(len(form.Redemptions), form.AssetVersion) * form.FeeSuggestion
+
```
chappjc, line 701: This is where we should be calling `isRedeemable`, I believe.  The next step is to broadcast any number of secrets, so the txn better not fail.

client/asset/eth/eth.go
```diff
@@ -672,8 +672,41 @@ func (eth *ExchangeWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin
...
+	// TODO: make sure the amount we locked for redemption is enough to cover the gas
+	// fees.
+	_, err := eth.node.redeem(eth.ctx, form.Redemptions, form.FeeSuggestion, form.AssetVersion)
+	if err != nil {
+		return fail(fmt.Errorf("Redeem: redeem error: %w", err))
+	}
```
JoeGruffins, line 707: Shoud a failure here unlock the coins, or does core take care of that?

JoeGruffins (commented):
It does seem like the first round of simnet tests always fails recently...

client/asset/eth/eth.go
martonp: Yes this was the whole point haha, I've updated it

client/asset/eth/eth.go
martonp: A success should unlock the coins as well. It's actually not so easy to handle the failure, since failure will actually consume some gas as well.. and if it keeps failing then we may use more gas than we had locked in the first place. Anyways, I have a TODO in there regarding this and there will be another PR regarding locking/unlocking funds for redemption, so this can be handled there.

JoeGruffins: If the error is a communication error I don't think the fee would be lost. I'm not sure if their could be another reason for a failure here. If the swap fails I don't think there will be an error.

A TODO is fine for now imo. Will be easier to diagnose when we start actually trading with eth.

martonp: Ah you're right, on chain error doesn't fail there. In that case we don't need to unlock funds for failure since it will be retried anyways.

JoeGruffins: It could be a communication error though with the internal geth node and the tx is never sent over the wire. I think that's the only type of error possible, but not sure. In that case the redeem gas stays locked. Doesn't it? I could be wrong.

JoeGruffins: Actually not even using IPC now, so not sure how this might error.

martonp: Yeah it shouldn't be unlocked.. those funds will need to be used on the retry.

JoeGruffins: Or, I guess the funds are not locked yet at this point?

martonp: They will be locked when creating the order.

JoeGruffins: Are these locked? I see `fundsRequired` saved to `t.metaData.RedemptionFeesPaid` in trade.go, but I don't see any locking or unlocking.

chappjc: https://github.com/decred/dcrdex/pull/1289/commits/990db478bba8b774accdfd71df7b33655114a0e8#r748809759

chappjc: https://github.com/decred/dcrdex/pull/1314

JoeGruffins: Oh, ok. Thank you.

JoeGruffins: We can figure it out there then.

JoeGruffins (approved):

client/asset/interface.go
chappjc: Pretty sure I'll clean it out in 1320, merging with this solution now

client/asset/eth/eth.go
```diff
@@ -145,6 +145,7 @@ type ethFetcher interface {
 	initiate(ctx context.Context, contracts []*asset.Contract, maxFeeRate uint64, contractVer uint32) (*types.Transaction, error)
 	shutdown()
 	syncProgress() ethereum.SyncProgress
+	isRedeemable(secretHash, secret [32]byte, contractVer uint32) (bool, error)
```
chappjc, line 148: This is good for me, but I have a suspicious that `redeem` can call the contractor method of `isRedeemable` internally.

PR #1303, opened 2021-11-17
eth/client: Pending balance

martonp:
This PR updates the `Balance` method of the ETH client wallet by taking into account the effects of pending transactions. Pending incoming funds need to be represented by the `Immature` field in the `asset.Balance` struct, and pending outgoing funds must be subtracted from the `Available` field. When retrieving the pending balance from the eth client, the effects of incoming and outgoing funds in pending transactions are aggregated, so the pending transactions are all checked for their effects individually.

Part of #1154

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -146,6 +146,7 @@ type ethFetcher interface {
 	accounts() []*accounts.Account
 	addPeer(ctx context.Context, peer string) error
 	balance(ctx context.Context, addr *common.Address) (*big.Int, error)
+	pendingBalance(ctx context.Context, addr *common.Address) (*big.Int, error)
 	bestBlockHash(ctx context.Context) (common.Hash, error)
 	bestHeader(ctx context.Context) (*types.Header, error)
 	block(ctx context.Context, hash common.Hash) (*types.Block, error)
@@ -344,8 +345,6 @@ func (eth *ExchangeWallet) OwnsAddress(address string) (bool, error) {
 
 // Balance returns the total available funds in the account. The eth node
 // returns balances in wei. Those are flored and stored as gwei, or 1e9 wei.
-//
-// TODO: Return Immature and Locked values.
 func (eth *ExchangeWallet) Balance() (*asset.Balance, error) {
 	eth.lockedFundsMtx.Lock()
 	defer eth.lockedFundsMtx.Unlock()
@@ -368,6 +367,15 @@ func (eth *ExchangeWallet) balanceImpl() (*asset.Balance, error) {
 		return nil, err
 	}
 
+	pendingBal, err := eth.node.pendingBalance(eth.ctx, &eth.acct.Address)
+	if err != nil {
+		return nil, err
+	}
+	gweiPendingBal, err := srveth.ToGwei(pendingBal)
+	if err != nil {
+		return nil, err
+	}
+
 	var amountLocked uint64
 	for _, value := range eth.lockedFunds {
 		amountLocked += value
@@ -377,10 +385,15 @@ func (eth *ExchangeWallet) balanceImpl() (*asset.Balance, error) {
 			fmt.Errorf("amount locked: %v > available: %v", amountLocked, gweiBal)
 	}
 
+	immature := gweiPendingBal - gweiBal
+	if gweiBal > gweiPendingBal {
+		immature = 0
+	}
+
 	bal := &asset.Balance{
 		Available: gweiBal - amountLocked,
 		Locked:    amountLocked,
-		// Immature: , How to know?
+		Immature:  immature,
 	}
 	return bal, nil
 }
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -15,7 +15,6 @@ import (
 	"fmt"
 	"math/big"
 	"path/filepath"
-	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -30,7 +29,6 @@ import (
 	"github.com/decred/dcrd/hdkeychain/v3"
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts"
-	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
@@ -146,7 +144,6 @@ type ethFetcher interface {
 	accounts() []*accounts.Account
 	addPeer(ctx context.Context, peer string) error
 	balance(ctx context.Context, addr *common.Address) (*big.Int, error)
-	pendingBalance(ctx context.Context, addr *common.Address) (*big.Int, error)
 	bestBlockHash(ctx context.Context) (common.Hash, error)
 	bestHeader(ctx context.Context) (*types.Header, error)
 	block(ctx context.Context, hash common.Hash) (*types.Block, error)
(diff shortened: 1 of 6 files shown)


JoeGruffins (approved):
I wasn't sure what immature meant for eth, but this looks reasonable.

client/asset/eth/eth.go
```diff
@@ -372,10 +380,15 @@ func (eth *ExchangeWallet) balanceImpl() (*asset.Balance, error) {
 			fmt.Errorf("amount locked: %v > available: %v", amountLocked, gweiBal)
 	}
 
+	immature := gweiPendingBal - gweiBal
+	if gweiBal > gweiPendingBal {
+		immature = 0
```
chappjc, line 385: I'd tend to reverse the logic and use the default zero value immature, but meh, it's fine.  `var immature uint64; if gweiPendingBal < gweiBal { ...`

chappjc: I'm gathering that if you _send from_ the wallet, that debit would only be reflected in `pendingBalance` but not `balance`?

If so, that's different from dcr/btc where sending immediately reduces the spendable/available balance even if the sending tx is unconfirmed.  With dcr/btc, receiving funds at least reflects "pending" in the unconfirmed category, but if it's normal to see `gweiBal > gweiPendingBal` that must be different here.

What are your observations on `balance` vs `pendingBalance` when 1) sending from the wallet, and 2) receiving to the wallet.

I'm wondering if we need to reduce Available if immature would be negative because of a pending debit.

buck54321:
I don't think eth has an analogue for "immature". Is there a maturity period for block rewards? If not, then immature should always be zero for eth, and "pending balance" is available balance for our purposes.

chappjc:
That's right @buck54321 we do include unconfirmed in Available for both dcr and btc.  I do wonder though if ETH should work differently given things are far more replaceable with ETH.  The second consideration with eth and eip-1559 in particular is that you can't know the amount the txn consumes until it is mined because the base fee fluctuates and isn't decided until the txn is mined.

buck54321:
It depends on whether "pending balance" considers the current base fee or the `GasFeeCap` in it's calculations. I'll need to confirm.

chappjc:
Either way it will very likely be something different when it does get mined.  It will be less if it uses GasFeeCap, and it will almost certainly be something other than "current" base fee unless it gets mined in the very next block.  I think we need to consider pending eth txn separate from available because of this.

chappjc:
Oh, there's also "gas limit" (set in txn submitted to txpool) vs "gas used" (only known after mining).

https://etherscan.io/txsPending

e.g.

image

So probably it's more fundamental than the EIP-1559 issue.
Throw replacement issues on top of that and I don't know if it's prudent to consider pending as available.  Curious what @martonp thinks about these issues too.

buck54321:
> Either way it will very likely be something different when it does get mined. **It will be less if it uses GasFeeCap**

If that's the case, then we don't need special handling because realized will always be greater than pending.

chappjc:
I'm mainly pointing out that the number will change just by txns being mined, and I feel like if the Available number shouldn't be variable with just time.

martonp:
I definitely didn't think this through as much as I should have.

We should decrease the Available funds when there's a pending outgoing transaction. Consider the case where a swap is submitted and the amount of locked funds decreases. If we don't reduce the balance as a result of this transaction, the Available amount would increase until the transaction is mined, which makes no sense. 

Regarding pending incoming transactions, I assumed that these would be represented by the amount in "Immature". When I see immature on the DEX UI, I just assume those are funds being received in transactions which have not yet been mined. I don't think many people are mining/staking with their dex accounts. @chappjc, you mentioned that in BTC/DCR unconfirmed are a part of Available, but in this code:
https://github.com/decred/dcrdex/blob/0f7d561a76e57b95174bcdde88a61c51d291f226/client/asset/btc/btc.go#L853
looks like only the "Trusted" amount is part of Avalable, and "Untrusted" is part of Immature.

Since we treat incoming and outgoing pending transactions differently, checking the pending balance isn't helpful because it aggregates the effects of both. I think the way to go is to just grab the pending transactions for the account and determine everything ourselves. It'll be fairly simple, and we'll have to do the same thing for ERC-20 anyways.

chappjc:
There's lots of [discussion on Matrix](https://matrix.to/#/!SFRQQFIHUUNXARfvew:decred.org/$vgVj6Yp5E_1BaVTz8JBAix7Jn1EHnLUIbGjJx1fZui0?via=decred.org&via=matrix.org&via=planetdecred.org) now, but I wanna link to a relevant github discussion: https://github.com/decred/dcrdex/pull/1243#discussion_r735231141 (regular balance does not even reflect you **sending from your account** until those txns are mined) There seem to be a number of qualities about regular balance that make it decidedly "not available" in a variety of ways.  Pretty irritating.

client/asset/eth/eth.go
```diff
@@ -338,15 +337,92 @@ func (eth *ExchangeWallet) OwnsAddress(address string) (bool, error) {
...
+	incoming, outgoing = new(big.Int), new(big.Int)
+
+	// Max possible gas fee
+	feeCap := tx.GasFeeCap()
+	gas := big.NewInt(int64(tx.Gas()))
+	outgoing.Mul(feeCap, gas)
+
+	// If the tx contains value, it is either an initiation or withdrawal.
+	// In either case we don't need to parse the tx data, so we can return
+	// here.
```
buck54321, line 361: Won't any transaction that transfers value from the sender have a non-zero `tx.Value()`? Will this catch all outgoing, regardless of type?

buck54321 (approved):
If you want to merge, @chappjc, I'll rebase #1301.

client/asset/eth/eth.go
chappjc: I think that's the interpretation of "withdrawal" here.

What I'm struggling to understand is how do we know we are the sender in this `types.Transaction`.  Its not yet clear to me that `eth_pendingTransactions` only pertains to this account's relevant transactions.  Do we need a `Signer` to determine the sender?

buck54321: Ultimately, we'll be using `(*light.TxPool).GetPoolTransactions`. The docs for `light.TxPool` say 

```
// TxPool implements the transaction pool for light clients, which keeps track
// of the status of locally created transactions...
```

Maybe we should check the `From` address, though.

chappjc: What about deposits then?  Those aren't locally-created transactions.

buck54321: That's true. Deposits will need a confirmation to be seen by the wallet, I guess.

chappjc: Apparently so: https://github.com/ethereum/go-ethereum/issues/16216#issuecomment-369191599

chappjc: BTW, in `startNode`, why don't we need to track and stop the light client:

```go
light, err := les.New(node, &ethCfg)
	if err != nil {
		return err
	}
	// on shutdown, light.BlockChain().Stop
```

chappjc: Nevermind that q about the light client.  I see it is registered with `Node`

client/asset/eth/eth.go
```diff
@@ -338,15 +337,92 @@ func (eth *ExchangeWallet) OwnsAddress(address string) (bool, error) {
...
+	if txValue.Cmp(new(big.Int)) > 0 {
+		outgoing.Add(outgoing, txValue)
+		return incoming, outgoing, nil
+	}
+
+	if redemptions, err := dexeth.ParseRedeemData(tx.Data()); err == nil {
+		for _, redemption := range redemptions {
+			swap, err := eth.node.swap(eth.ctx, eth.acct, redemption.SecretHash)
+			if err != nil {
+				return nil, nil, err
```
chappjc, line 372: Wondering if log and continue would be better here so the balance isn't wrecked.  We can just see how this goes though.

client/asset/eth/eth.go
```diff
@@ -338,15 +337,92 @@ func (eth *ExchangeWallet) OwnsAddress(address string) (bool, error) {
 
 // Balance returns the total available funds in the account. The eth node
 // returns balances in wei. Those are flored and stored as gwei, or 1e9 wei.
```
chappjc, line 339: We should doc that unconfirmed incoming value other than our own redeem and refund txns is not reflected.

dex/networks/eth/txdata.go
```diff
@@ -92,3 +96,62 @@ func ParseRedeemData(calldata []byte) ([]ETHSwapRedemption, error) {
...
+	parsedAbi, err := abi.JSON(strings.NewReader(ETHSwapABI))
+	if err != nil {
+		return nil, err
+	}
+	return parsedAbi.Pack(redeemFuncName, redemptions)
+}
+
+// PackRedeemData converts a secret hash to call data for the refund function.
+func PackRefundData(secretHash [32]byte) ([]byte, error) {
+	parsedAbi, err := abi.JSON(strings.NewReader(ETHSwapABI))
```
chappjc, line 152: This suggestion can wait, but I think we should have an `init` function prepare the `parsedABI`

buck54321: I'll resolve on rebase.

chappjc:
I'm pushing a conflict resolution on your branch because we're just about to merge it anyway. Sorry, @martonp.

chappjc (approved):

PR #1306, opened 2021-11-17
eth: compare parsed address in OwnsAddress

chappjc:
The string address can have different formattings, so parse the input string into a `common.Address` and compare that to the account address.  For example, it can be formatted with mixed case and with or without "0x", among other things.

Also use direct array comparison for address equality instead of `bytes.Equal`.

Also address some missing error checks.

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -237,7 +237,7 @@ func CreateWallet(createWalletParams *asset.CreateWalletParams) error {
 		return err
 	}
 
-	importKeyToNode(node, privateKey, createWalletParams.Pass)
+	err = importKeyToNode(node, privateKey, createWalletParams.Pass)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


client/asset/eth/eth.go
```diff
@@ -331,11 +331,17 @@ func (eth *ExchangeWallet) Connect(ctx context.Context) (*sync.WaitGroup, error)
...
+// not be a EIP55-compliant formatted address. It may or may not have a 0x
+// prefix, and case is not important.
 //
 // In Ethereum, an address is an account.
 func (eth *ExchangeWallet) OwnsAddress(address string) (bool, error) {
-	return strings.ToLower(eth.acct.Address.String()) == strings.ToLower(address), nil
+	if !common.IsHexAddress(address) {
+		return false, errors.New("invalid address")
+	}
+	addr := common.HexToAddress(address)
```
chappjc, line 343: `common.NewMixedCaseAddress` would work fine for this too, but it has a little extra overhead we don't need

JoeGruffins (approved):

buck54321 (approved):

chappjc:
Thank you.  I saw the same SignMessage receiver change in #1301 , which I have been reviewing today

PR #1307, opened 2021-11-18
server/asset/eth: update server coin ID parsing

buck54321:
Older notions of coin ID have evolved. Except in the case of a funding coin, ETH coin ID will simply be the tx hash bytes.

### DIFF
--- a/server/asset/eth/eth.go
+++ b/server/asset/eth/eth.go
@@ -65,11 +65,11 @@ func (d *Driver) Setup(configPath string, logger dex.Logger, network dex.Network
 
 // DecodeCoinID creates a human-readable representation of a coin ID for Ethereum.
 func (d *Driver) DecodeCoinID(coinID []byte) (string, error) {
-	coinId, err := dexeth.DecodeCoinID(coinID)
+	txHash, err := decodeCoinID(coinID)
 	if err != nil {
 		return "", err
 	}
-	return coinId.String(), nil
+	return txHash.String(), nil
 }
 
 // UnitInfo returns the dex.UnitInfo for the asset.
@@ -209,18 +209,17 @@ func (eth *Backend) Connect(ctx context.Context) (*sync.WaitGroup, error) {
 
 // TxData fetches the raw transaction data.
 func (eth *Backend) TxData(coinID []byte) ([]byte, error) {
-	cnr, err := dexeth.DecodeCoinID(coinID)
+	txHash, err := decodeCoinID(coinID)
 	if err != nil {
 		return nil, fmt.Errorf("coin ID decoding error: %v", err)
 	}
-	c, is := cnr.(*dexeth.TxCoinID)
-	if !is {
-		return nil, fmt.Errorf("wrong type of coin ID, %v", cnr)
-	}
 
-	tx, _, err := eth.node.transaction(eth.rpcCtx, c.TxID)
+	tx, _, err := eth.node.transaction(eth.rpcCtx, txHash)
 	if err != nil {
 		return nil, fmt.Errorf("error retrieving transaction: %w", err)
+	}
+	if tx == nil { // Possible?
+		return nil, fmt.Errorf("no transaction %s", txHash)
 	}
 
 	return tx.MarshalBinary()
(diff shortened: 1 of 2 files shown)


server/asset/eth/eth_test.go
```diff
@@ -172,113 +171,157 @@ func TestLoad(t *testing.T) {
...
+		if tt.wantErr {
+			t.Fatalf("%s: no error", tt.name)
+		}
+		if res != tt.expRes {
+			t.Fatalf("%s: wrong result. wanted %s, got %s", tt.name, tt.expRes, res)
+		}
 	}
 }
 
+// func TestCoinIDs(t *testing.T) {
```
buck54321, line 218: Moved in #1301.

server/asset/eth/eth_test.go
```diff
@@ -646,7 +689,7 @@ func TestTxData(t *testing.T) {
 	addr := randomAddress()
 	data := encode.RandomBytes(5)
 	tx := tTx(gasPrice, value, addr, data)
-	goodCoinID := (&TxCoinID{TxID: tx.Hash()}).Encode()
```
buck54321, line 649: I don't know, maybe we still want to have a versioned coin ID. But from where I'm at this just seems like avoidable overhead. Communications for ETH appear to be simpler than utxo-based assets.

server/asset/eth/eth.go
```diff
@@ -207,19 +207,18 @@ func (eth *Backend) Connect(ctx context.Context) (*sync.WaitGroup, error) {
...
-	}
 
-	tx, _, err := eth.node.transaction(eth.rpcCtx, c.TxID)
+	tx, _, err := eth.node.transaction(eth.rpcCtx, txHash)
 	if err != nil {
 		return nil, fmt.Errorf("error retrieving transaction: %w", err)
 	}
+	if tx == nil { // Possible?
+		return nil, fmt.Errorf("no transaction %s", txHash)
+	}
```
JoeGruffins, line 221: In my limited experience no tx is an `ethereum.NotFound` err, but this is probably safe.

server/asset/eth/eth.go
```diff
@@ -65,11 +65,11 @@ func (d *Driver) Setup(configPath string, logger dex.Logger, network dex.Network
 
 // DecodeCoinID creates a human-readable representation of a coin ID for Ethereum.
 func (d *Driver) DecodeCoinID(coinID []byte) (string, error) {
-	coinId, err := DecodeCoinID(coinID)
```
chappjc, line 68: What's your vision for removing the exported `DecodeCoinID` that switches on a `CoinIDFlag`?

buck54321: `DecodeCoinID` is moved in #1301. To **dex/networks/eth/cointypes.go ** for now, but may end up in **client/asset/eth** since I don't think we'll end up using them at all on the server.

server/asset/eth/eth_test.go
```diff
@@ -172,113 +171,157 @@ func TestLoad(t *testing.T) {
...
-	if !ok {
-		t.Fatalf("expected coin to be a TxCoin")
-	}
-	if !bytes.Equal(originalTxCoin.TxID[:], decodedTxCoin.TxID[:]) {
-		t.Fatalf("expected txIds to be equal before and after decoding")
-	}
+func TestDecodeCoinID(t *testing.T) {
+	drv := &Driver{}
+	txid := "0x1b86600b740d58ecc06eda8eba1c941c7ba3d285c78be89b56678da146ed53d1"
+	txHashB, _ := hex.DecodeString("1b86600b740d58ecc06eda8eba1c941c7ba3d285c78be89b56678da146ed53d1")
```
JoeGruffins, line 177: Good to use a panicking helper to decode imo, just in case.

JoeGruffins (approved):

chappjc (approved):

server/asset/eth/eth.go
```diff
@@ -335,11 +334,11 @@ func (eth *Backend) Redemption(redeemCoinID, contractCoinID []byte) (asset.Coin,
 
 // ValidateCoinID attempts to decode the coinID.
 func (eth *Backend) ValidateCoinID(coinID []byte) (string, error) {
-	coinId, err := dexeth.DecodeCoinID(coinID)
+	txHash, err := decodeCoinID(coinID)
```
chappjc, line 337: This is going to OK because with https://github.com/decred/dcrdex/pull/1293, `ValidateCoinID` is not used for order funding with account based assets, only UTXO assets?  I think we need to document this here.

PR #1308, opened 2021-11-19
dcr: stdscript migration

chappjc:
<s>Requires https://github.com/decred/dcrwallet/pull/2108 for a clean go.sum, but otherwise it's fine to review.</s>

This migrates to the new `txscript/v4/stdscript` package with the newly tagged txscript v4.0.0 module.

This also updates to tagged versions of the following dcrd modules:
- chaincfg/v3 v3.1.0
- edwards/v2 v2.0.2
- secp256k1/v4 v4.0.1
- dcrutil/v4 v4.0.0
- wire v1.5.0
- gcs/v3.0.0
- hdkeychain/v3.0.1

The new stdscript package lends itself to quite a bit of simplification and removal of duplicate code, which as it turns out was a bit more fragile than necessary w.r.t. script versions, new stake opcodes, etc.


client/asset/dcr/dcr.go
```diff
@@ -2411,6 +2408,11 @@ func (dcr *ExchangeWallet) parseUTXOs(unspents []walletjson.ListUnspentResult) (
 			}
 			return nil, fmt.Errorf("error reading asset info: %w", err)
 		}
+		if nfo.ScriptType == dexdcr.ScriptUnsupported {
+			// InputInfo does this for P2SH with non-standard redeem scripts.
+			// Don't return these since they cannot fund arbitrary txns.
+			continue
+		}
```
chappjc, line 2415: There's no change elsewhere that requires this, I believe we were just overlooking this.

dex/networks/dcr/script.go
```diff
@@ -701,85 +508,64 @@ func DataPrefixSize(data []byte) uint8 {
...
+	scriptType, scriptAddrs := ExtractScriptAddrs(version, pkScript, chainParams)
+	switch {
+	case scriptType == ScriptUnsupported:
 		return nil, dex.UnsupportedScriptError
+	case scriptType.IsP2PKH():
+		return &SpendInfo{
+			SigScriptSize: P2PKHSigScriptSize,
+			ScriptAddrs:   scriptAddrs,
+			ScriptType:    scriptType,
+		}, nil
```
chappjc, line 524: I didn't *need* to rework `InputInfo` this much, but nothing of substance is changed and it is far more readable IMO.

davecgh (commented):
Looks good.  Some comments inline.

client/asset/dcr/dcr.go
```diff
@@ -1521,19 +1522,13 @@ func (dcr *ExchangeWallet) AuditContract(coinID, contract, txData dex.Bytes, reb
...
-	}
-	if scriptClass != txscript.ScriptHashTy {
+	scriptClass, addrs := stdscript.ExtractAddrs(contractTxOut.Version, contractTxOut.PkScript, dcr.chainParams)
+	if scriptClass != stdscript.STScriptHash {
 		return nil, fmt.Errorf("unexpected script class %d", scriptClass)
 	}
 	if len(addrs) != 1 {
 		return nil, fmt.Errorf("unexpected number of addresses for P2SH script: %d", len(addrs))
 	}
-	if sigsReq != 1 {
```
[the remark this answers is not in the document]
chappjc: Thanks for calling this out.  I believe this was a "why not" kind of check since in this method (`AuditContract`), we ensure the script class is `STScriptHash` with just a single p2sh address successfully decoded from the pkScript (noting that as per our [recent discussion](https://github.com/decred/dcrd/pull/2807#discussion_r752492737) `ExtractAddrs` may return a standard script class even if address decoding failed).

In version 0 scripts, stdscript claims 1 sig for `STScriptHash`, but how meaningful that is is not clear (the number of signatures actually required to spend depends on the SH script).  The important script check other than the pkScript class and address checks is the contract sh script done separately.

However, I don't mind putting back the `sigsReq` check with `stdscript.DetermineRequiredSigs` if my reasoning is off.

dex/networks/dcr/script.go
```diff
@@ -556,90 +402,51 @@ func IsDustVal(sz, amt, minRelayTxFee uint64) bool {
...
-
-	// Could switch on class from ExtractPkScriptAddrs, but use ParseScriptType
-	// for internal consistency.
-	scriptType := ParseScriptType(version, script, nil)
-
+// count from a pkScript. Non-standard scripts are not an error; non-nil errors
+// are only returned if the script cannot be parsed. See also InputInfo for
+// additional signature script size data
+func ExtractScriptData(version uint16, script []byte, chainParams *chaincfg.Params) (ScriptType, []string, int) {
+	class, addrs := stdscript.ExtractAddrs(version, script, chainParams)
```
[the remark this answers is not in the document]
chappjc: Ah, `scriptType` is used below after converting to this package's own `ScriptType` type with the bit masks.  This var could be `stdScriptType` perhaps to denote it's from `stdscript`.  I figure I'll keep it though so stay in line with the code prior to this diff however.

client/asset/dcr/dcr.go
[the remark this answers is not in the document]
davecgh: Personally, I think it's more than fine to remove the check, because, for the most part, the number of required signatures is mostly legacy baggage brought along for the ride and, realistically, has dubious meaning in various circumstances.

As you pointed out, once such example is for `STScriptHash`, which is just an assumption that there will be at least one signature required, but you really need the redeem script to determine anything more.  However, it's even more ambiguous when you start talking about threshold signatures which just appear on-chain to require a single signatures, but in reality, it might actually need hundreds.

I mainly just wanted to call out that it was being removed so there was no chance of it looking like an accident.

server/asset/dcr/utxo.go
```diff
@@ -381,7 +379,10 @@ func auditContract(op *Output) (*asset.Contract, error) {
 		return nil, fmt.Errorf("invalid script version %d", op.scriptVersion)
 	}
 	output := tx.outs[int(op.vout)]
-	scriptHash := dexdcr.ExtractScriptHash(output.pkScript)
+	if output.version != 0 {
+		return nil, fmt.Errorf("unsupported script version %d", output.version)
+	}
+	scriptHash := stdscript.ExtractScriptHashV0(output.pkScript)
```
JoeGruffins, line 385: I see 
```
if op.scriptVersion != 0 {
		return nil, fmt.Errorf("invalid script version %d", op.scriptVersion)
}
```
right above this, is it the same check?

Also, something I've been wondering is if we should also verify the transaction version is 1? 

I believe we are already using one exclusively. There are three possible at the moment, 1, the initial or default we use, 2 enables sequence locks, which I'm not sure but could be used to make a transaction that we think is fine when examining but cannot be mined because of the sequence? and three enables treasury opcodes, I believe. So maybe forcing one also prevents funding with treasury gens?

If there was a new version in the future that allows something that breaks the dex, we would want to prevent that here. I think.

chappjc: Good spot.  It's actually quite peculiar that this function is using `op.tx.outs` at all.  Both the version and the pkScript itself are fields of the `Output`...  The double check of version in the two different places is alright, but weird.  Not sure if I wanna change it yet.

Good question about the tx version.  Perhaps we should make an issue to get that sorted with outside input.

dex/networks/dcr/script.go
```diff
@@ -701,85 +508,64 @@ func DataPrefixSize(data []byte) uint8 {
...
 		SigScriptSize: uint32(sigScriptSize),
 		ScriptAddrs:   scriptAddrs,
 		ScriptType:    scriptType,
 	}, nil
 }
 
-// ExtractContractHash extracts the contract P2SH address from a pkScript. A
-// non-nil error is returned if the hexadecimal encoding of the pkScript is
-// invalid, or if the pkScript is not a P2SH script.
-func ExtractContractHash(scriptHex string) ([]byte, error) {
```
buck54321, line 770: This wasn't used?

dex/networks/dcr/script_test.go
```diff
@@ -268,44 +286,49 @@ func Test_nonstandardScript(t *testing.T) {
...
 	}
 
 	// Double check that the pkScript's script hash matches the hash of the
 	// redeem (contract) script.
-	scriptHash, err := ExtractScriptHashByType(scriptType, pkScript)
-	if err != nil {
-		t.Fatalf(fmt.Sprintf("ExtractScriptHashByType error: %v", err))
+	scriptHash := ExtractScriptHash(scriptVersion, pkScript)
+	if scriptHash == nil {
+		t.Fatalf("ExtractScriptHashV0 failed")
```
buck54321, line 305: Super nit, but should probably just say `"ExtractScriptHash failed"`

buck54321 (changes requested):
Tried to register with the simnet-setup script and saw 

```
error paying registration fee: sendrawtx error: -4: wallet.PublishTransaction(e17b08e754740831b134a016b8091db1ef3ade1754604745107562948d18bbf8):: dcrd.PublishTransactions: rejected transaction e17b08e754740831b134a016b8091db1ef3ade1754604745107562948d18bbf8: Tx e17b08e754740831b134a016b8091db1ef3ade1754604745107562948d18bbf8 attempted to spend an OP_SSTX tagged output, however it was not an SSGen or SSRtx tx; SSGen err: SSgen tx has an invalid number of inputs, SSRtx err: SSRtx output at output index 0 was not an OP_SSRTX tagged output, raw tx: 0100000001dd75751cd7e5db1eb8f7cc44b684e8f2cea8bba28e01c468ceeb7ac94706084c0000000001ffffffff0200e1f5050000000000001976a9142ed2fd8155ca933e3f8c392a83268fd1675ce26288ac136085000000000000001976a9143c636c24aa6f8c13529ddf7782504db936a5219c88ac000000000000000001dc4b7b060000000000000000ffffffff6a473044022060ff25463a49fd3467704d591e6e26e501d90c36334d3551ff98aeabcb550d68022030ec3ee587e9df1d11fd6e7d312fe6744986f2c9323a6e518ba84623ca77592e012103509677f91f91150fe444ebef77a6aa27ae07715bb862a37d0d0564f86c1edc94
```

dex/networks/dcr/script.go
chappjc: Found no reference outside of tests.  It might have been at one point.

dex/networks/dcr/script.go
```diff
@@ -120,216 +120,62 @@ const (
...
-}
-
-// Check if the opcode is one of a small number of acceptable stake opcodes that
-// are prepended to P2PKH scripts. Return the opcode if it is, else OP_0.
-func stakeOpcode(script []byte) byte {
-	if len(script) == 0 {
-		return 0
-	}
-	opcode := script[0]
-	if opcode == txscript.OP_SSGEN || opcode == txscript.OP_SSRTX {
```
chappjc, line 295: Ahh, the `IsStakePubkeyHashScript` and `IsStakeScriptHashScript` were not detecting ticket outputs as stake and lumping them as nonstandard and thus not spendable.
We need to allow `InputInfo` to only return a non-nil `*SpendInfo` for the actually spendable stake output types.

chappjc: It previously reported vote outputs, so it is strange that we didn't have trouble because those have maturity rules that were never checked.

dex/networks/dcr/script.go
```diff
@@ -701,85 +508,64 @@ func DataPrefixSize(data []byte) uint8 {
...
-			// Should have been caught by ParseScriptType.
-			return nil, fmt.Errorf("bare multisig is unsupported")
-		}
-		// P2PK also should have been caught by ParseScriptType.
-		return nil, fmt.Errorf("unsupported pkScript: %d", scriptType)
+	// NOTE: scriptAddrs are now from the redeemScript, and we add the multisig
+	// bit to scriptType.
+	if redeemScriptType.IsMultiSig() {
+		scriptType |= ScriptMultiSig
+	}
```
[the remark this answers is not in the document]
chappjc: I'll have another go at this, but you end up having to call `stdscript.ExtractAddrs` an extra time for no reason.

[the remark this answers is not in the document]
buck54321: Only if it's easy.

dex/networks/dcr/script.go
chappjc: Well, I'm now checking the `ListUnspentResult.Spendable` field in the funding and spendable utxo parsing methods in https://github.com/decred/dcrdex/pull/1308/commits/4bc82f77293d79da2126fc0cd3cc8c0b619fb7d5
We do the same in BTC.
I'm actually really surprised that immature and thus unspendable `stakegen` outputs from votes never threw these errors.

dex/networks/dcr/script.go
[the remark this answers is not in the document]
chappjc: Easy, yes, but from what I recall, tricky if we want to avoid a repeated script class detection.  I'll look again.

dex/networks/dcr/script.go
chappjc: > We need to allow InputInfo to only return a non-nil *SpendInfo for the actually spendable stake output types.

The above is still on the table as a complement to the `ListUnspentResult.Spendable` in the calling code, but I've not done this.  It would mean calling various stake output types "unsupported" script classes even though we can identify them as stake, so it feels icky going back to that behavior.  Might yet though...

dex/networks/dcr/script.go
```diff
@@ -120,216 +120,62 @@ const (
...
+		scriptType = ScriptP2SH
+	case stdscript.STStakeChangeScriptHash, stdscript.STStakeGenScriptHash,
+		stdscript.STStakeRevocationScriptHash, stdscript.STStakeSubmissionScriptHash,
+		stdscript.STTreasuryGenScriptHash:
+		scriptType = ScriptP2SH | ScriptStake
+	// P2MS
+	case stdscript.STMultiSig:
+		scriptType = ScriptMultiSig
+	default: // STNonStandard, STNullData, STTreasuryAdd, etc
+		return ScriptUnsupported
```
chappjc, line 176: The previous behavior was effectively to include in this `default` the stake outputs except for stakegen and stakerevocation.  We could go back to that, but I think we were still overlooking maturity requirements for both of these types, so my feeling now is that it's best to detect the script type as accurately as possible (within the bounds of our `ScriptType` enum) and determine spendability from the signer, dcrwallet, in the listunspent result.
Not totally set on that, but it feels right.

JoeGruffins (approved):

client/asset/dcr/dcr.go
```diff
@@ -938,7 +938,7 @@ func (dcr *ExchangeWallet) tryFund(utxos []*compositeUTXO, enough func(sum uint6
 
 		okUTXOs := make([]*compositeUTXO, 0, len(utxos)) // over-allocate
 		for _, cu := range utxos {
-			if cu.confs >= minconf {
+			if cu.confs >= minconf && cu.rpc.Spendable {
```
buck54321, line 941: This is just extra cautious though, right? Should never get this far with a non-Spendable?

buck54321 (approved):

client/asset/dcr/dcr.go
chappjc: Yes, extra.  unspendable shouldn't make it up here.

chappjc:
Played around some more after those last module updates to dcrwallet (both the dcrwallet cli itself and our require).

Issue #1252, opened 2021-10-25
update geth to at least 1.10.11


chappjc:
I noticed that abigen from go-ethereum 1.10.10 is now generating a contract.go with breaking changes.  The data has been reorganized, but it generates code for a version of the `abi/bind` package that apparently has breaking changes of its own.

A quick update of the require revealed other breaking changes.  The main one that broke the build is that a `node.Node` no longer has account manager backends by default, requiring them to be registered.  https://github.com/ethereum/go-ethereum/pull/23019
This seems to require changes like:

```diff
- ks := node.AccountManager().Backends(keystore.KeyStoreType)[0].(*keystore.KeyStore)
+ ks := keystore.NewKeyStore(node.KeyStoreDir(), keystore.LightScryptN, keystore.LightScryptP)
+ node.AccountManager().AddBackend(ks)
```

JoeGruffins:
@martonp no problems! I'll always make an issue first. Didn't notice I needed to update though, will do now.

PR #1309, opened 2021-11-23
networks/eth: Move generated files to swap.

### LINKED ISSUE #1235: eth: Move server/eth/common.go to dex/networks/eth.

JoeGruffins:
networks/eth: Move generated files to swap.
    
    In order to avoid possible naming conflicts with generated files, move
    them to their own package.

closes #1235


JoeGruffins:
I made some changes to `dex/networks/eth/updatecontract.sh` so pls make sure it still functions as expected. notably creating a temp folder to delete at the end, and removing the temp file from `sed`.

Alos, there are unexpected changes to `contract.go`. Is it maybe my version of abigen? I'm on `1.10.6-stable` atm.

chappjc:
> Alos, there are unexpected changes to contract.go. Is it maybe my version of abigen? I'm on 1.10.6-stable atm

Yes, it turns out that abigen produces breaking code with just patch releases. :/ 
https://github.com/decred/dcrdex/issues/1252
We had to regnerate code for 1.10.10 recently.  I recommend updating your geth toolchain.

chappjc:
Oh hey fun, another hardfork by decree: https://eips.ethereum.org/EIPS/eip-4345

JoeGruffins:
Keeping up with eth changes is a full time job...

JoeGruffins:
"Difficulty Bomb Delay" this isn't the place, but seriously, who are they kidding.

Or maybe it is the place? What do we do if eth ends? Can just swap everything with eth2?

JoeGruffins:
```
$ abigen --version
abigen version 1.10.11-stable
```
Most of the changes didn't happen, but the bin value is changing. Want to know why. Is it maybe a timestamp in there?

dex/networks/eth/swap/BinRuntimeV0.go
```diff
@@ -1,6 +1,6 @@
...
...fffffffffffff808211156107b357600080fd5b818501915085601f8301126107c757600080fd5b8135818111156107d657600080fd5b8660208260061b850101111561077657600080fd5b6000602082840312156107fd57600080fd5b5035919050565b60006020828403121561081657600080fd5b5051919050565b6000806040838503121561083057600080fd5b50508035926020909101359150565b6004811061085d57634e487b7160e01b600052602160045260246000fd5b9052565b6000825160005b818110156108825760208186018101518583015201610868565b81811115610891576000828501525b509190910192915050565b8781526020810187905260408101869052606081018590526001600160a01b038481166080830152831660a082015260e081016108dc60c083018461083f565b98975050505050505050565b600060e08201905082518252602083015160208301526040830151604083015260608301516060830152608083015160018060a01b0380821660808501528060a08601511660a0850152505060c083015161094660c084018261083f565b5092915050565b6000821982111561096057610960610980565b500190565b600060001982141561097957610979610980565b5060010190565b634e487b7160e01b600052601160045260246000fd5b634e487b7160e01b600052602160045260246000fd5b634e487b7160e01b600052603260045260246000fdfea2646970667358221220521ba5e5a8d228ddbd60c84eb643d79f7b061442c3245f63fc317513a5b3c12464736f6c63430008060033"
```
JoeGruffins, line 6: At the end here
from:
aabe46bd3b24f22b9df02893c3ee5bf05defb2833d18f954ba6014f5dbb9d2ee
to:
521ba5e5a8d228ddbd60c84eb643d79f7b061442c3245f63fc317513a5b3c124

JoeGruffins: It seems this is directly related to where the `.sol` files are when using `abigen`.
`abigen --sol ETHSwapV0.sol --pkg swap --out ./contract.go` will pruduce the value we had while
`abigen --sol contracts/ETHSwapV0.sol --pkg swap --out ./contract.go` will produce this new value.

So, I guess fine. Hashing file names for some reason.

It does have implications for checking that a contract is created from a certain `.sol` file. As you can see here the transaction receipt will also reflect this data, if it is deployed using byte code (I don't know at how deploying with go would look, but I assume it would be the same):
```
eth.getTransaction("0x3dd24c93f5734273d350b4ef720849ed72bec6bb4326c9bfeb051580587eeaf5")
{
  blockHash: "0x1c5c634ac73a3187f9b4515da743674a786883798a19a4bc09cb96f41f40a269",
  blockNumber: 4,
  from: "0x18d65fb8d60c1199bb1ad381be47aa692b482605",
  gas: 603931,
  gasPrice: 82000000000,
  hash: "0x3dd24c93f5734273d350b4ef720849ed72bec6bb4326c9bfeb051580587eeaf5",
  input: "0x608060405234801561001057600080fd5b506109f8806100206000396000f3fe6080604052600436106100555760003560e01c80637249fbb61461005a57806376467cbd1461007c578063a8793f94146100b2578063bfd2fd97146100c5578063eb84e7f2146100f5578063f4fd17f914610171575b600080fd5b34801561006657600080fd5b5061007a6100753660046107eb565b610191565b005b34801561008857600080fd5b5061009c6100973660046107eb565b6102a6565b6040516100a991906108e8565b60405180910390f35b61007a6100c0366004610713565b610381565b3480156100d157600080fd5b506100e56100e036600461081d565b6104ad565b60405190151581526020016100a9565b34801561010157600080fd5b5061015e6101103660046107eb565b60006020819052908152604090208054600182015460028301546003840154600485015460059095015493949293919290916001600160a01b0391821691811690600160a01b900460ff1687565b6040516100a9979695949392919061089c565b34801561017d57600080fd5b5061007a61018c366004610788565b610584565b32331461019d57600080fd5b806001600082815260208190526040902060050154600160a01b900460ff1660038111156101cd576101cd610996565b146101d757600080fd5b6000818152602081905260409020600401546001600160a01b031633146101fd57600080fd5b6000818152602081905260409020600301544281111561021c57600080fd5b60008381526020819052604080822060058101805460ff60a01b1916600360a01b1790556001015490513391905b60006040518083038185875af1925050503d8060008114610287576040519150601f19603f3d011682016040523d82523d6000602084013e61028c565b606091505b50909150506001811515146102a057600080fd5b50505050565b6102e36040805160e081018252600080825260208201819052918101829052606081018290526080810182905260a081018290529060c082015290565b60008281526020818152604091829020825160e08101845281548152600182015492810192909252600281015492820192909252600380830154606083015260048301546001600160a01b039081166080840152600584015490811660a084015291929160c0840191600160a01b90910460ff169081111561036757610367610996565b600381111561037857610378610996565b90525092915050565b32331461038d57600080fd5b6000805b8281101561049b57368484838181106103ac576103ac6109ac565b9050608002019050600080600083602001358152602001908152602001600020905060008260600135116103df57600080fd5b81356103ea57600080fd5b60006005820154600160a01b900460ff16600381111561040c5761040c610996565b1461041657600080fd5b436002820155813560038201556004810180546001600160a01b0319163317905561044760608301604084016106ea565b6005820180546060850135600185018190556001600160a01b03939093166001600160a81b031990911617600160a01b179055610484908561094d565b93505050808061049390610965565b915050610391565b503481146104a857600080fd5b505050565b60006001600084815260208190526040902060050154600160a01b900460ff1660038111156104de576104de610996565b14801561050457506000838152602081905260409020600501546001600160a01b031633145b801561057d57508260028360405160200161052191815260200190565b60408051601f198184030181529082905261053b91610861565b602060405180830381855afa158015610558573d6000803e3d6000fd5b5050506040513d601f19601f8201168201806040525081019061057b9190610804565b145b9392505050565b32331461059057600080fd5b6000805b828110156106d457368484838181106105af576105af6109ac565b6020604091820293909301838101356000908152938490529220919250600190506005820154600160a01b900460ff1660038111156105f0576105f0610996565b146105fa57600080fd5b60058101546001600160a01b0316331461061357600080fd5b81602001356002836000013560405160200161063191815260200190565b60408051601f198184030181529082905261064b91610861565b602060405180830381855afa158015610668573d6000803e3d6000fd5b5050506040513d601f19601f8201168201806040525081019061068b9190610804565b1461069557600080fd5b60058101805460ff60a01b1916600160a11b1790558135815560018101546106bd908561094d565b9350505080806106cc90610965565b915050610594565b506000336001600160a01b03168260405161024a565b6000602082840312156106fc57600080fd5b81356001600160a01b038116811461057d57600080fd5b6000806020838503121561072657600080fd5b823567ffffffffffffffff8082111561073e57600080fd5b818501915085601f83011261075257600080fd5b81358181111561076157600080fd5b8660208260071b850101111561077657600080fd5b60209290920196919550909350505050565b6000806020838503121561079b57600080fd5b823567ffffffffffffffff808211156107b357600080fd5b818501915085601f8301126107c757600080fd5b8135818111156107d657600080fd5b8660208260061b850101111561077657600080fd5b6000602082840312156107fd57600080fd5b5035919050565b60006020828403121561081657600080fd5b5051919050565b6000806040838503121561083057600080fd5b50508035926020909101359150565b6004811061085d57634e487b7160e01b600052602160045260246000fd5b9052565b6000825160005b818110156108825760208186018101518583015201610868565b81811115610891576000828501525b509190910192915050565b8781526020810187905260408101869052606081018590526001600160a01b038481166080830152831660a082015260e081016108dc60c083018461083f565b98975050505050505050565b600060e08201905082518252602083015160208301526040830151604083015260608301516060830152608083015160018060a01b0380821660808501528060a08601511660a0850152505060c083015161094660c084018261083f565b5092915050565b6000821982111561096057610960610980565b500190565b600060001982141561097957610979610980565b5060010190565b634e487b7160e01b600052601160045260246000fd5b634e487b7160e01b600052602160045260246000fd5b634e487b7160e01b600052603260045260246000fdfea2646970667358221220aabe46bd3b24f22b9df02893c3ee5bf05defb2833d18f954ba6014f5dbb9d2ee64736f6c63430008060033",
  nonce: 3,
  r: "0x907bc6819b7062abe4c77d88f6f8a7709f34701786c9208cfcc09585d64c63e2",
  s: "0x497ac7065ed84aaaab19a0e0a75fce230ae57090cf8d0e44002413150f9a6b2e",
  to: null,
  transactionIndex: 3,
  type: "0x0",
  v: "0x77",
  value: 0
}
```

Also, it doesn't look like there will be an easy way to find the transaction that creates a contract, so the txid should be recorded whenever one is deployed at the time of deploying and made public beside the address I think. I wonder if we could record the deploying txid to a variable in the contract?

martonp:
updateContract.sh no longer works on mac with your change. The temp file is required.

dex/networks/eth/swap/BinRuntimeV0.go
martonp: https://docs.soliditylang.org/en/develop/metadata.html
"Since the bytecode of the resulting contract contains the metadata hash by default, any change to the metadata might result in a change of the bytecode. This includes changes to a filename or path, and since the metadata includes a hash of all the sources used, a single whitespace change results in different metadata, and different bytecode."

JoeGruffins:
guess waiting for #1301 to rebase

Will put the temp file back in sed for mac.

JoeGruffins:
The second commit moved the file, but #1301 moved it so just popped that commit off.

server/asset/eth/eth_test.go
```diff
@@ -19,7 +19,7 @@ import (
 	"decred.org/dcrdex/dex/calc"
 	"decred.org/dcrdex/dex/encode"
 	dexeth "decred.org/dcrdex/dex/networks/eth"
-	swap "decred.org/dcrdex/dex/networks/eth"
```
chappjc, line 22: Toasting these duplicate imports, good.

dex/networks/eth/swap/contract.go
```diff
@@ -1,7 +1,7 @@
 // Code generated - DO NOT EDIT.
 // This file is a generated binding and any manual changes will be lost.
 
-package eth
+package swap
```
chappjc, line 4: No objection to the move.  In fact this was recently discussed and it will eventually be needed to have these files versioned to avoid collisions.

https://github.com/decred/dcrdex/pull/1301#discussion_r757660939

Although it'll actually need to be versioned **folders** at that point to have versioned packages.

client/asset/eth/contractor_test.go
```diff
@@ -7,7 +7,7 @@ import (
 
 	"decred.org/dcrdex/client/asset"
 	"decred.org/dcrdex/dex/encode"
-	dexeth "decred.org/dcrdex/dex/networks/eth"
+	"decred.org/dcrdex/dex/networks/eth/swap"
```
chappjc, line 10: Thinking about multiple versions, I think we'll end up like:

```
swapv0 "decred.org/dcrdex/dex/networks/eth/swap/v0"
swapv1 "decred.org/dcrdex/dex/networks/eth/swap/v1"
```

dex/networks/eth/swap/contract.go
JoeGruffins: You mean v0 v1 v2 folders?

client/asset/eth/contractor_test.go
JoeGruffins: That makes sense.

JoeGruffins: Moved it to a v0 package.

JoeGruffins:
I think I had the eth simnet tests randomly fail a couple times, but not sure now. I may have messed up a doc or two with replacements, but I think it's all alright.

buck54321 (approved):

dex/networks/eth/txdata.go
```diff
@@ -28,8 +29,8 @@ const (
 // contract with extra data. It will error if the call data does not call
 // initiate with expected argument types. It returns the array of initiations
 // with which initiate was called.
-func ParseInitiateData(calldata []byte) ([]ETHSwapInitiation, error) {
-	decoded, err := parseCallData(calldata, ETHSwapABI)
+func ParseInitiateData(calldata []byte) ([]swapv0.ETHSwapInitiation, error) {
+	decoded, err := parseCallData(calldata, swapv0.ETHSwapABI)
```
chappjc, line 33: I'm likely to work on making ParseInitiateData and other dex/networks/eth helpers version-aware following a chain of other PRs: https://github.com/decred/dcrdex/pull/1307, https://github.com/decred/dcrdex/pull/1313, https://github.com/decred/dcrdex/pull/1320

@buck54321 pls note.  We've chatted about these needing versioning.  While I don't insist on doing it personally, I just wanna coordinate.

PR #1313, opened 2021-11-28
eth: Update TxCoinID and Contract

martonp:
TxCoinID contained an Index field which is now removed, and in many places the code assumed that a contract is a SwapCoinID when in fact it should be just a secret hash.

### DIFF
--- a/dex/networks/eth/cointypes.go
+++ b/dex/networks/eth/cointypes.go
@@ -48,8 +48,8 @@ type CoinID interface {
 }
 
 const (
-	// coin type id (2) + tx id (32) + index (4) = 38
-	txCoinIDSize = 38
+	// coin type id (2) + tx id (32) = 34
+	txCoinIDSize = 34
 	// coin type id (2) + address (20) + secret has (32) = 54
 	swapCoinIDSize = 54
 	// coin type id (2) + address (20) + amount (8) + nonce (8) = 38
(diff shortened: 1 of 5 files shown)


martonp:
Seems to me like we don't need `SwapCoinID`. Since we are just using a single contract for now, we don't need to keep track of the contract address, and we can pass the secret hash back and forth between server and client as the contract. We only need `AmountCoinID` and `TxCoinID` for coin id.

martonp:
I hadn't looked at https://github.com/decred/dcrdex/pull/1307 yet. Most of this is still useful though.

chappjc:
> I hadn't looked at #1307 yet. Most of this is still useful though.

Yes, most of this is good.

To feel out how the coin ID saga may play out if we commit to a single txid-based encoding for comms and server-side code, I rebased this PR on master and stacked a few commits on top.  https://github.com/decred/dcrdex/compare/master...chappjc:eth-coins?expand=1
I was talking it through on matrix, but it's starting to look doable this way.  Please let us know how this sits with you.  I'm on the fence on the whole thing, having suggested the type flag encodings at one point too, but I appreciate the desire to have one consistent encoding that is a tx id for communications.  It creates a few more hoops to jump through, but it doesn't seem much worse than polymorphic coin IDs.

chappjc:
I have https://github.com/decred/dcrdex/pull/1326 requiring this PR, so I'd like to have this vetted and merged as early in the week as possible.  I rebased this PRs commit in https://github.com/decred/dcrdex/pull/1326, but you might go another way on the details and that's fine, I will adapt mine once this goes in.

martonp:
I think this PR is fine to go in.. or even might as well get rid of it and just go with your PR since this is just an intermediate solution. One issue with this is that `newSwapCoin` assumes that contract is just a secret hash, not a versioned string. This is still true throughout the server code, so that can be updated separately.

chappjc:
> One issue with this is that `newSwapCoin` assumes that contract is just a secret hash, not a versioned string. This is still true throughout the server code, so that can be updated separately

I plan to update this in a number of places with https://github.com/decred/dcrdex/pull/1320

PR #1314, opened 2021-11-29
core,eth: handle redemption funding

buck54321:
```   
    client/asset:
    Add AccounterRedeemer interface with methods for locking and unlocking
    funds for redemption. ReserveN is used to reserve funds for an order.
    Only the amount reserved is returned. All accounting from there is
    handled by Core, who will return funds proportionally according to
    match quantities and order size. The amount reserved is persisted in
    the OrderMetaData so that accounting can be picked up exactly as it was on restarts.
    This pattern was chosen over trying to figure out how much to lock
    on restart with calculations based on order info because you would need
    to consider changes in asset configuration and therefore persist more
    data and ... it gets messy.
    
    client/core:
    Add RedeemSig to orders redeeming to and AccountRedeemer.
    Track and return reserved redemption funds. Return funds manually
    for refunds and revocations. Handle dust that can arise from rounding
    error. Log inconsistencies.
    
    client/asset/eth:
    Implement asset.AccountRedeemer. Track redemption reserves separately
    than swap reserves, but add them when calculating locked balance.
```

### DIFF
--- a/client/asset/eth/nodeclient.go
+++ b/client/asset/eth/nodeclient.go
@@ -23,12 +23,12 @@ import (
 	dexeth "decred.org/dcrdex/dex/networks/eth"
 	ethv0 "decred.org/dcrdex/dex/networks/eth/contracts/v0"
 	"github.com/ethereum/go-ethereum"
-	"github.com/ethereum/go-ethereum/accounts"
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/state"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/eth/ethconfig"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/ethereum/go-ethereum/les"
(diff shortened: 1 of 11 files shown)


chappjc:
I was imagining a `FundBuyOrder` to mirror the `FundOrder` method, which is implicitly for a sale of the asset.  Then just use the usual `ReturnCoins` if cancelling.  Does splitting the redemption reserves away from other locked coins solve a major challenge with that approach?

buck54321:
> I was imagining a `FundBuyOrder` to mirror the `FundOrder` method, which is implicitly for a sale of the asset. Then just use the usual `ReturnCoins` if cancelling or executing unfilled. Does splitting the redemption reserves away from other locked coins solve a major challenge with that approach?

Not 100% sure I follow, but you should try it out. This is maybe the third different direction I started down, and the others got messy. 

I think `FundBuyOrder` might be confusing though, since it's presumably still used for sell orders.

chappjc:
I will consider it more.  I do appreciate some of the issues you've addressed by tracking amounts instead of coins or lots independently.
Perhaps `FundBuyOrder` would be better as `FundReceivingOrder`, but it's probably a moot point.

client/asset/eth/nodeclient.go
```diff
@@ -17,11 +17,12 @@ import (
 	"decred.org/dcrdex/dex"
 	dexeth "decred.org/dcrdex/dex/networks/eth"
 	"github.com/ethereum/go-ethereum"
-	"github.com/ethereum/go-ethereum/accounts"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/state"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/crypto/secp256k1"
```
chappjc, line 25: Uh oh this is a cgo package.

chappjc: We should be able to recover the pubkey with our own secp256k1 packages, otherwise surely we can pull the pubkey from the keystore.

chappjc: Resolved in https://github.com/decred/dcrdex/pull/1358

buck54321:
Still need match revocation handling, but if nobody objects, I'll work on getting this out of draft.

chappjc:
OK.

buck54321:
**Order Completion**

| Result | Action | 
|---------|-----------|
| Revoked before matching | return all [[1]](https://github.com/buck54321/dcrdex/blob/28add9fae545f11fb6ef8289b54e22c91625a5ab/client/core/trade.go#L1323-L1331) |
| Revoked from booked | return remaining [[1]](https://github.com/buck54321/dcrdex/blob/28add9fae545f11fb6ef8289b54e22c91625a5ab/client/core/trade.go#L1323-L1331) | 
| Canceled | return remaining [[2]](https://github.com/buck54321/dcrdex/blob/28add9fae545f11fb6ef8289b54e22c91625a5ab/client/core/trade.go#L565-L573) | 
| nomatch - market order | return all [[3]](https://github.com/buck54321/dcrdex/blob/28add9fae545f11fb6ef8289b54e22c91625a5ab/client/core/trade.go#L388-L390)  |
| Executed - market buy | do nothing. will return when redeemed |
| Executed - market sell | return remaining [[2]](https://github.com/buck54321/dcrdex/blob/28add9fae545f11fb6ef8289b54e22c91625a5ab/client/core/trade.go#L565-L573) |
| Executed - limit | nothing |

**Match Revoked** 

| Step | Maker | Taker |
|-------|----------|--------|
| NewlyMatched | return all [[4]](https://github.com/buck54321/dcrdex/blob/28add9fae545f11fb6ef8289b54e22c91625a5ab/client/core/trade.go#L1369-L1379) | return all  [[4]](https://github.com/buck54321/dcrdex/blob/28add9fae545f11fb6ef8289b54e22c91625a5ab/client/core/trade.go#L1369-L1379) |
| MakerSwapCast | nothing (will refund) [[5]](https://github.com/buck54321/dcrdex/blob/28add9fae545f11fb6ef8289b54e22c91625a5ab/client/core/trade.go#L2049-L2055) | return all [[4]](https://github.com/buck54321/dcrdex/blob/28add9fae545f11fb6ef8289b54e22c91625a5ab/client/core/trade.go#L1369-L1379) |
| TakerSwapCast | nothing (will redeem) [[6]](https://github.com/buck54321/dcrdex/blob/28add9fae545f11fb6ef8289b54e22c91625a5ab/client/core/trade.go#L1740) | nothing (will redeem via findRedemption, or refund) [[5]](https://github.com/buck54321/dcrdex/blob/28add9fae545f11fb6ef8289b54e22c91625a5ab/client/core/trade.go#L2049-L2055) |
| MakerRedeemed | nothing | nothing (will redeem) [[6]](https://github.com/buck54321/dcrdex/blob/28add9fae545f11fb6ef8289b54e22c91625a5ab/client/core/trade.go#L1740) |

JoeGruffins (approved):

martonp (approved):
LGTM, just minor comments

client/asset/eth/nodeclient.go
```diff
@@ -362,23 +362,38 @@ func (n *nodeClient) swap(ctx context.Context, secretHash [32]byte, contractVer
...
+		return nil, nil, err
+	}
+	if len(sig) != 65 {
+		return nil, nil, fmt.Errorf("unexpected signature length %d", len(sig))
+	}
+
+	pubKey, err = recoverPubkey(crypto.Keccak256(data), sig)
+	if err != nil {
+		return nil, nil, fmt.Errorf("SignMessage: error recovering pubkey %w", err)
+	}
```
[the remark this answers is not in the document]
buck54321: I actually had it that way at one point.

client/asset/interface.go
```diff
@@ -227,6 +227,20 @@ type Wallet interface {
 	RegFeeConfirmations(ctx context.Context, coinID dex.Bytes) (confs uint32, err error)
 }
 
+type AccountRedeemer interface {
```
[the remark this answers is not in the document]
buck54321: I assume you mean the problems you're addressing in #1392? I don't think we'll need any new `Wallet` methods for swap funding.

chappjc (approved):

PR #1315, opened 2021-11-30
client/asset/eth: SwapConfirmations

buck54321:
Add an asset version argument to `(Wallet).SwapConfirmations` and implement the method for eth.

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -751,8 +751,32 @@ func (*ExchangeWallet) PayFee(address string, regFee, feeRateSuggestion uint64) 
 	return nil, asset.ErrNotImplemented
 }
 
-func (*ExchangeWallet) SwapConfirmations(ctx context.Context, id dex.Bytes, contract dex.Bytes, startTime time.Time) (confs uint32, spent bool, err error) {
-	return 0, false, asset.ErrNotImplemented
+// SwapConfirmations gets the number of confirmations and the spend status
+// for the specified swap.
+func (eth *ExchangeWallet) SwapConfirmations(ctx context.Context, _ dex.Bytes, secretHashB dex.Bytes,
+	_ time.Time, assetVersion uint32) (confs uint32, spent bool, err error) {
+
+	hdr, err := eth.node.bestHeader(ctx)
+	if err != nil {
+		return 0, false, fmt.Errorf("error fetching best header: %w", err)
+	}
+
+	var secretHash [32]byte
+	copy(secretHash[:], secretHashB)
+
+	swapData, err := eth.node.swap(ctx, secretHash, assetVersion)
+	if err != nil {
+		return 0, false, fmt.Errorf("error finding swap state: %w", err)
+	}
+
+	if swapData.State == dexeth.SSNone {
+		return 0, false, asset.CoinNotFoundError
+	}
+
+	spent = swapData.State >= dexeth.SSRedeemed
+	confs = uint32(hdr.Number.Uint64() - swapData.BlockHeight + 1)
+
+	return
 }
 
 // Withdraw withdraws funds to the specified address. Value is gwei.
--- a/client/asset/eth/eth_test.go
+++ b/client/asset/eth/eth_test.go
@@ -12,6 +12,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"errors"
+	"fmt"
 	"math/big"
 	"math/rand"
 	"strings"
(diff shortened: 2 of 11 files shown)


client/core/trade.go
```diff
@@ -1064,7 +1064,7 @@ func (t *trackedTrade) shouldBeginFindRedemption(ctx context.Context, match *mat
 		return false
 	}
 
-	confs, spent, err := t.wallets.fromWallet.SwapConfirmations(ctx, swapCoinID, proof.Script, match.MetaData.Stamp)
+	confs, spent, err := t.wallets.fromWallet.SwapConfirmations(ctx, swapCoinID, proof.Script, match.MetaData.Stamp, t.metaData.FromVersion)
```
chappjc, line 1067: Presently eth's `(*swapReceipt).Contract() dex.Bytes` returns a slice that contains the secret hash.

What if it returned a slice that was a concatenation of contract version and secret hash, so that `client.asset.Receipt.Contract()`, which is assigned to `proof.Script`, has this info.  It's already not a script at all as implied by the `MatchProof.Script` field name, and it's intended to be an opaque blob for multi-asset support, so why not have it encode the version too?

That is, we're already freely reinterpreting the `Receipt.Contract` method, so I'd just assume we use it to convey all the info needed about this contract/swap rather than add an arg to `SwapConfirmations`, which without close examination is kind of a head-scratcher.

client/asset/eth/eth.go
```diff
@@ -751,8 +751,32 @@ func (*ExchangeWallet) PayFee(address string, regFee, feeRateSuggestion uint64)
...
+	if err != nil {
+		return 0, false, fmt.Errorf("error finding swap state: %w", err)
+	}
+
+	if swapData.State == dexeth.SSNone {
+		return 0, false, asset.CoinNotFoundError
+	}
+
+	spent = swapData.State >= dexeth.SSRedeemed
+	confs = uint32(hdr.Number.Uint64() - swapData.BlockHeight + 1)
```
chappjc, line 777: Looking good in this method, and it does seem to be alright if the `contract dex.Bytes` arg encoded both the asset version and the secret hash.   Open to that?

chappjc: i.e.

```go
func (eth *ExchangeWallet) SwapConfirmations(ctx context.Context, _ dex.Bytes, contractData dex.Bytes,
	_ time.Time) (confs uint32, spent bool, err error) {

	hdr, err := eth.node.bestHeader(ctx)
	if err != nil {
		return 0, false, fmt.Errorf("error fetching best header: %w", err)
	}

	if len(contractData) != 36 {
		return 0, false, fmt.Errorf("invalid contract data")
	}

	// contractData fully specifies this swap in terms of which contract version
	// is used and the secret hash that keys the unique swap.
	var secretHash [32]byte
	copy(secretHash[:], contractData[:32])
	assetVersion := binary.BigEndian.Uint32(contractData[33:])

	swapData, err := eth.node.swap(ctx, secretHash, assetVersion)
	if err != nil {
		return 0, false, fmt.Errorf("error finding swap state: %w", err)
	}

	if swapData.State == dexeth.SSNone {
		return 0, false, asset.CoinNotFoundError
	}

	spent = swapData.State >= dexeth.SSRedeemed
	confs = uint32(hdr.Number.Uint64() - swapData.BlockHeight + 1)

	return
}
```

client/asset/eth/eth_test.go
```diff
@@ -937,7 +937,7 @@ func TestSwap(t *testing.T) {
 				t.Fatalf("%v: receipt coin value: %v != expected: %v",
 					testName, receipt.Coin().Value(), contract.Value)
 			}
-			if !bytes.Equal(receipt.Contract(), contract.SecretHash[:]) {
+			if !bytes.Equal(receipt.Contract()[4:], contract.SecretHash[:]) {
```
chappjc, line 940: Would you please quick update with the version check to ensure `Contract` is encoding it there as in https://github.com/decred/dcrdex/commit/bbc9d81b3697e328ca732c176913ff94f991b98a#diff-59ea012074143e9b443cb736910be9904ee1f5101bd081f609fd88b22e128491R942

```
			contractData := receipt.Contract()
			if swaps.AssetVersion != binary.BigEndian.Uint32(contractData[:4]) {
				t.Fatal("wrong contract version")
			}
			if !bytes.Equal(contractData[4:], contract.SecretHash[:]) 
```

before calling:

```
contractData := make([]byte, 36)
binary.BigEndian.PutUint32(contractData[:4], 12)
copy(contractData[4:], secretHash[:])
```

PR #1317, opened 2021-11-30
client/asset/eth: RegFeeConfirmations

buck54321:
Implement `RegFeeConfirmations`. Add `transactionConfirmations` method
to `nodeClient`. Check local tx pool first, since non-zero confs will
always request from peers, and the tx we're looking for is ours.

### DIFF
--- a/client/asset/eth/nodeclient.go
+++ b/client/asset/eth/nodeclient.go
@@ -500,6 +500,32 @@ func (n *nodeClient) isRedeemable(secretHash [32]byte, secret [32]byte, contract
 	})
 }
 
+// transactionConfirmations gets the number of confirmations for the specified
+// transaction.
+func (n *nodeClient) transactionConfirmations(ctx context.Context, txHash common.Hash) (uint32, error) {
+	// We'll check the local tx pool first, since from what I can tell, a light
+	// client always requests tx data from the network for anything else.
+	if tx := n.leth.ApiBackend.GetPoolTransaction(txHash); tx != nil {
+		return 0, nil
+	}
+	hdr, err := n.bestHeader(ctx)
+	if err != nil {
+		return 0, err
+	}
+	tx, _, blockHeight, _, err := n.leth.ApiBackend.GetTransaction(ctx, txHash)
+	if err != nil {
+		return 0, err
+	}
+	if tx != nil {
+		return uint32(hdr.Number.Uint64() - blockHeight + 1), nil
+	}
+	// TODO: There may be a race between when the tx is removed from our local
+	// tx pool, and when our peers are ready to supply the info. I saw a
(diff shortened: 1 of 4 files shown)


martonp (approved):
LGTM

client/asset/eth/nodeclient.go
```diff
@@ -500,6 +500,32 @@ func (n *nodeClient) isRedeemable(secretHash [32]byte, secret [32]byte, contract
...
+	// We'll check the local tx pool first, since from what I can tell, a light
+	// client always requests tx data from the network for anything else.
+	if tx := n.leth.ApiBackend.GetPoolTransaction(txHash); tx != nil {
+		return 0, nil
+	}
+	hdr, err := n.bestHeader(ctx)
+	if err != nil {
+		return 0, err
+	}
+	tx, _, blockHeight, _, err := n.leth.ApiBackend.GetTransaction(ctx, txHash)
```
chappjc, line 515: OK, but it's curious that they do this first internally: https://github.com/ethereum/go-ethereum/blob/40cdcf8c47ff094775aca08fd5d94051f9cf1dbb/internal/ethapi/api.go#L1158-L1168
Could it be possible that there would be a result from `GetPoolTransaction` when `GetTransaction` would find it?
You got a CoinNotFoundError once...

client/asset/eth/nodeclient.go
```diff
@@ -500,6 +500,32 @@ func (n *nodeClient) isRedeemable(secretHash [32]byte, secret [32]byte, contract
 	})
 }
 
+// transactionConfirmations gets the number of confirmations for the specified
+// transaction.
+func (n *nodeClient) transactionConfirmations(ctx context.Context, txHash common.Hash) (uint32, error) {
+	// We'll check the local tx pool first, since from what I can tell, a light
+	// client always requests tx data from the network for anything else.
```
chappjc, line 507: Curious what you discovered regarding network requests.

I started digging and it looks like it at least has a local db that it checks before sending off a request, like a cache.
From les/api_backend.go, `(*LesApiBackend).GetTransaction` > `light.GetTransaction`, then in light/odr_util.go:
`GetTransaction` > `GetBody` > `GetBodyRLP` > `rawdb.ReadBodyRLP` otherwise `odr.Retrieve` for the data.

client/asset/eth/nodeclient.go
JoeGruffins: It makes more chronological sense to check the pool first, otherwise you might miss it checking "finalized" transactions first and it gets "finalized" before you check the pool where it was just removed.

JoeGruffins (approved):

chappjc (approved):
Ok.

I still find it curious that `(*PublicTransactionPoolAPI) GetTransactionByHash` does the checks in the opposite order.

PR #1318, opened 2021-11-30
client/asset/eth: re-implement translation tests for abigen types

buck54321:
These were ripped out with the changes in #1301 because `ExchangeWallet` no longer deals with the abigen types. Now re-implemented through contractor tests.

### DIFF
--- a/client/asset/eth/contractor.go
+++ b/client/asset/eth/contractor.go
@@ -157,8 +157,6 @@ func (c *contractorV0) swap(ctx context.Context, secretHash [32]byte) (*dexeth.S
 		return nil, err
 	}
 
-	fmt.Println("--swap", state.Value, dexeth.WeiToGwei(state.Value))
-
 	return &dexeth.SwapState{
 		BlockHeight: state.InitBlockNumber.Uint64(),
 		LockTime:    time.Unix(state.RefundBlockTimestamp.Int64(), 0),
--- a/client/asset/eth/contractor.go
+++ b/client/asset/eth/contractor.go
@@ -118,11 +118,13 @@ func (c *contractorV0) initiate(txOpts *bind.TransactOpts, contracts []*asset.Co
 			Value:           new(big.Int).Mul(bigVal, dexeth.BigGweiFactor),
 		})
 	}
+
 	return c.contractV0.Initiate(txOpts, inits)
 }
 
 func (c *contractorV0) redeem(txOpts *bind.TransactOpts, redemptions []*asset.Redemption) (*types.Transaction, error) {
 	redemps := make([]swapv0.ETHSwapRedemption, 0, len(redemptions))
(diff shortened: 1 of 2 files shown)


chappjc:
> Looks good, but I think `tContractV0` should be used inside the `testNode` in eth_test.go, and then, for example, we can make sure that the `Swap` function is calling `Initiate` with the proper arguments.

eth_test.go needs updating for the contract version args, but not sure if tContractV0 should be in those tests.
I pushed a small tweak on the [just merged commit](https://github.com/decred/dcrdex/commit/70df246081af876be00e51c2b8463d1ad6f0b76d#diff-59ea012074143e9b443cb736910be9904ee1f5101bd081f609fd88b22e128491R151). See `swapVers` to illustrate what I think is missing from the other tests.  Sorry to force this @buck54321.  You can kill it completely in this PR or another if you like.

JoeGruffins (approved):

martonp:
The reason I said to include tContractV0 in the tests is because it already keeps track of the last initiations and last redeems, which are useful for some of the tests in eth_test.go.

buck54321:
> The reason I said to include tContractV0 in the tests is because it already keeps track of the last initiations and last redeems, which are useful for some of the tests in eth_test.go.

This isn't so clear anymore, since the `[]*asset.Contract` args is passed through `ExchangeWallet` and `nodeClient` unmodified now.

buck54321:
> eth_test.go needs updating for the contract version args, but not sure if tContractV0 should be in those tests.
I pushed a small tweak on the just merged commit. See `swapVers` to illustrate what I think is missing from the other tests. Sorry to force this @buck54321. You can kill it completely in this PR or another if you like.

I don't think I follow. What is the ask here?

chappjc:
Weird I accidentally edited your comment instead of adding one.  Sorry.  Anyway...

> I don't think I follow. What is the ask here?

See the swapVers bits in the test. I put those there, but feel free to remove them. I believe part of what @martonp is saying was correct that the tests in eth_test.go were a little lacking on basic input verification.

client/asset/eth/contractor.go
```diff
@@ -145,6 +152,8 @@ func (c *contractorV0) swap(ctx context.Context, secretHash [32]byte) (*dexeth.S
 		return nil, err
 	}
 
+	fmt.Println("--swap", state.Value, dexeth.WeiToGwei(state.Value))
```
chappjc, line 155: Can remove

client/asset/eth/contractor_test.go
```diff
@@ -1,54 +1,67 @@
+//go:build lgpl
+// +build lgpl
```
chappjc, line 2: Thanks, I noticed this too.

chappjc (approved):
Looks good if you remove the Println from debugging

PR #1319, opened 2021-11-30
eth/client: AuditContract

martonp:
Part of #1154

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -160,6 +160,7 @@ type ethFetcher interface {
 	signData(addr common.Address, data []byte) ([]byte, error)
 	sendToAddr(ctx context.Context, addr common.Address, val uint64) (*types.Transaction, error)
 	transactionConfirmations(context.Context, common.Hash) (uint32, error)
+	sendSignedTransaction(ctx context.Context, tx *types.Transaction) error
 }
 
 // Check that ExchangeWallet satisfies the asset.Wallet interface.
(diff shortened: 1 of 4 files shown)


client/asset/eth/nodeclient.go
```diff
@@ -500,6 +500,11 @@ func (n *nodeClient) isRedeemable(secretHash [32]byte, secret [32]byte, contract
 	})
 }
 
+// sendSignedTransaction injects a signed transaction into the pending pool for execution.
+func (n *nodeClient) sendSignedTransaction(ctx context.Context, tx *types.Transaction) error {
+	return n.ec.SendTransaction(ctx, tx)
```
chappjc, line 505: This can now be `n.leth.ApiBackend.SendTx(ctx, tx)`, but it looks like the client method gives some extra logging.  Still looking into it.

buck54321: Please don't use the `ethclient.Client`. It's an unnecessary encode-decode for both the transaction and the result. It's going to `n.leth.ApiBackend.SendTx(ctx, tx)` in the end anyway.

client/asset/eth/eth.go
```diff
@@ -697,8 +698,63 @@ func (eth *ExchangeWallet) SignMessage(coin asset.Coin, msg dex.Bytes) (pubkeys,
...
+	if err != nil {
+		return nil, fmt.Errorf("AuditContract: failed to unmarshal transaction: %w", err)
+	}
+
+	txHash := tx.Hash()
+	if !bytes.Equal(coinID, txHash[:]) {
+		return nil, fmt.Errorf("AuditContract: coin id != txHash - coin id: %x, txHash: %x", coinID, tx.Hash())
+	}
+
+	initiations, err := dexeth.ParseInitiateData(tx.Data())
```
chappjc, line 713: OK, so with the push toward versioned contracts, we'd have to do the same encoding scheme with (ver|secretHash) for the `contract` input, decode the version, and then use a version-aware `dexeth.ParseInitiateData`, returning a new `SwapInitiation` type that is translated internally from each version's `ETHSwapInitiation`.

This would mean that either `msgjson.Audit.Contract` from the server is gonna need that encoding, or when we call `AuditContract`, we prepend the version we're expecting based on the server config.  I'm not yet clear on which way we want to go with this.

chappjc: Oh, sorry it would just be in `OrderMetaData.{FromVersion,ToVersion}`.  Just getting used to the new scheme from 1301...

buck54321: My vote is to prepend the version.

buck54321: I think we can just use `*asset.Contract` as the common type.

buck54321: Sorry to spam the comments here, but I just realized that the version is already prepended, since the counter-party's `swapReceipt` encodes it that way.

chappjc: I'm working on some server updates for contract versioning, but yeah, @buck54321 is right that the `msgjson.Audit.Contract` passed to `AuditContract` already has it.

martonp: Things are changing quickly lately!

chappjc: They certainly are.  Last night I started on updates for server to handle the version in what I'm calling contract data from the swap receipt and thus in the `msgjson.Init.Contract` and everywhere downstream:  https://github.com/decred/dcrdex/pull/1320
I will have that follow https://github.com/decred/dcrdex/pull/1313, which will follow https://github.com/decred/dcrdex/pull/1307

client/asset/eth/eth.go
```diff
@@ -700,9 +701,71 @@ func (eth *ExchangeWallet) SignMessage(coin asset.Coin, msg dex.Bytes) (pubkeys,
...
+
+	secretHash, _, err := decodeVersionedBytes(contract)
+	if err != nil {
+		return nil, fmt.Errorf("AuditContract: failed to decode versioned bytes: %w", err)
+	}
+
+	var initiation *dexeth.ETHSwapInitiation
+	for _, init := range initiations {
+		if bytes.Equal(init.SecretHash[:], secretHash) {
+			initiation = &init
```
JoeGruffins, line 732: Can `break` after finding.

JoeGruffins (approved):
I'm embarrassingly not up to date with all the communication changes, but this all looks good to me.

buck54321 (commented):
I was holding off on reviewing this until the `ParseInitiateData` changes come through, but if it's important to anyone to get it earlier, we can.

chappjc:
> I was holding off on reviewing this until the `ParseInitiateData` changes come through, but if it's important to anyone to get it earlier, we can.

I dunno if order matters too much.  Whoever does the ParseInitiateData etc changes can apply it to AuditContract too.

martonp:
> > I was holding off on reviewing this until the `ParseInitiateData` changes come through, but if it's important to anyone to get it earlier, we can.
> 
> I dunno if order matters too much. Whoever does the ParseInitiateData etc changes can apply it to AuditContract too.

I'm working on them right now, but yeah this can be merged first.

buck54321 (approved):
This looks good for now. Obviously a lot will be changing here with versioned `ParseInitiateData` and #1320.

client/asset/eth/eth.go
```diff
@@ -700,9 +702,72 @@ func (eth *ExchangeWallet) SignMessage(coin asset.Coin, msg dex.Bytes) (pubkeys,
...
+	if !bytes.Equal(coinID, txHash[:]) {
+		return nil, fmt.Errorf("AuditContract: coin id != txHash - coin id: %x, txHash: %x", coinID, tx.Hash())
+	}
+
+	initiations, err := dexeth.ParseInitiateData(tx.Data())
+	if err != nil {
+		return nil, fmt.Errorf("AuditContract: failed to parse initiate data: %w", err)
+	}
+
+	secretHash, _, err := decodeVersionedBytes(contract)
```
chappjc, line 725: Just a note here that a number of things come together because this `contract` input, as well as the  `asset.AuditInfo.Contract` where it is stored and pass back to the caller, contain the secret hash and contract version.
This is important because in #1320 the `asset.RedeemForm` input to `Redeem` gets this via `asset.Redemption.Spends` (this `*Auditinfo`).
As @buck54321 suggested, this does indeed appear to make `ToVersion` obsolete.  Will push a commit shortly, but since this PR actually implementing `AuditContract` isn't merged, it might not be obvious how it works when I push.

chappjc: With #1320 merged, should now use `dexeth.DecodeContractData` and remove the added helper

chappjc (approved):

PR #1320, opened 2021-12-01
dex/networks,server/eth: decode swap data message blob

chappjc:
The `msgjson.Init.Contract` and `server/asset.Contract.RedeemScript` fields
contain an encoding of the ETH contract version concatenated with the
swap's secret hash that is the unique key for the swap.

This updates server/asset/eth's `ValidateContract` and `Contract` methods
to decode and check this data.

This also adds the `dexeth.EncodeContractData` and `DecodeContractData` functions.

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -10,7 +10,6 @@ import (
 	"bytes"
 	"context"
 	"crypto/sha256"
-	"encoding/binary"
 	"encoding/hex"
 	"errors"
 	"fmt"
@@ -131,6 +130,12 @@ func (d *Driver) Exists(walletType, dataDir string, settings map[string]string, 
 
 func (d *Driver) Create(params *asset.CreateWalletParams) error {
 	return CreateWallet(params)
+}
+
+// Balance is the current balance, including information about the pending
+// balance.
+type Balance struct {
+	Current, PendingIn, PendingOut *big.Int
 }
 
 // ethFetcher represents a blockchain information fetcher. In practice, it is
@@ -564,7 +569,7 @@ func (eth *ExchangeWallet) unlockFunds(coins asset.Coins) error {
 // swapReceipt implements the asset.Receipt interface for ETH.
 type swapReceipt struct {
 	txHash     common.Hash
-	secretHash []byte
+	secretHash [dexeth.SecretHashSize]byte
 	// expiration and value can be determined with a blockchain
 	// lookup, but we cache these values to avoid this.
 	expiration time.Time
@@ -586,9 +591,10 @@ func (r *swapReceipt) Coin() asset.Coin {
 	}
 }
 
-// Contract returns the swap's secret hash.
+// Contract returns the swap's identifying data, which the concatenation of the
+// contract version and the secret hash.
 func (r *swapReceipt) Contract() dex.Bytes {
-	return versionedBytes(r.ver, r.secretHash[:])
+	return dexeth.EncodeContractData(r.ver, r.secretHash)
 }
 
 // String returns a string representation of the swapReceipt.
@@ -645,12 +651,14 @@ func (eth *ExchangeWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin
 
 	txHash := tx.Hash()
 	for _, swap := range swaps.Contracts {
+		var secretHash [dexeth.SecretHashSize]byte
+		copy(secretHash[:], swap.SecretHash)
 		receipts = append(receipts,
 			&swapReceipt{
 				expiration: encode.UnixTimeMilli(int64(swap.LockTime)),
 				value:      swap.Value,
 				txHash:     txHash,
-				secretHash: swap.SecretHash,
+				secretHash: secretHash,
 				ver:        swaps.AssetVersion,
 			})
 	}
@@ -667,7 +675,10 @@ func (eth *ExchangeWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin
 }
 
 // Redeem sends the redemption transaction, which may contain more than one
-// redemption.
+// redemption. All redemptions must be for the same contract version because the
+// current API requires a single transaction reported (asset.Coin output), but
+// conceptually a batch of redeems could be processed for any number of
+// different contract addresses with multiple transactions.
 func (eth *ExchangeWallet) Redeem(form *asset.RedeemForm) ([]dex.Bytes, asset.Coin, uint64, error) {
 	fail := func(err error) ([]dex.Bytes, asset.Coin, uint64, error) {
 		return nil, nil, 0, err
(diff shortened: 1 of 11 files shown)


dex/networks/eth/params.go
```diff
@@ -65,6 +70,26 @@ var v0Gases = &Gases{
...
+// slice for communicating a swap's identity.
+func EncodeSwapData(contractVersion uint32, swapKey [SecretHashSize]byte) []byte {
+	b := make([]byte, SecretHashSize+4)
+	binary.BigEndian.PutUint32(b[:4], contractVersion)
+	copy(b[4:], swapKey[:])
+	return b
+}
+
+// DecodeSwapData unpacks the contract version and secret hash.
+func DecodeSwapData(data []byte) (contractVersion uint32, swapKey [SecretHashSize]byte, err error) {
```
buck54321, line 83: How about `DecodeVersionedHash` or something? `DecodeSwapData` seems like it could do the same thing as `ParseInitiateData` to me.

chappjc: That's alright with me, but the "versioned" wording is slightly confusing to me because I see this as a piece of data that identifies a swap by `(contract version, secret hash)` rather than a hash that is versioned.  I guess that's pretty subtle.
I was initially wanting this to be worded generically to reflect that it's global swap identifying data, rather than describing that's in the data, but this is all internal to the eth packages, so I suppose there's no need to be too generic, plus we can always rename if there's more data we find necessary to include in this blob.

chappjc:
Not quite ready, but rebased and slightly updated.

live_test.go needs the lgpl build flag now.

dex/networks/eth/params.go
```diff
@@ -65,6 +70,26 @@ var v0Gases = &Gases{
 	RefundGas: 43000,
 }
 
+// EncodeContractData packs the contract version and the secret hash into a byte
+// slice for communicating a swap's identity.
+func EncodeContractData(contractVersion uint32, swapKey [SecretHashSize]byte) []byte {
```
chappjc, line 75: I won't object to a rename here, but this feels right to me.

client/asset/eth/eth.go
```diff
@@ -692,19 +722,20 @@ func (eth *ExchangeWallet) Redeem(form *asset.RedeemForm) ([]dex.Bytes, asset.Co
...
 
-		swapData, err := eth.node.swap(eth.ctx, secretHash, form.AssetVersion)
+		swapData, err := eth.node.swap(eth.ctx, secretHash, ver)
 		if err != nil {
 			return nil, nil, 0, fmt.Errorf("Redeem: error finding swap state: %w", err)
 		}
 		redeemedValue += swapData.Value
 		inputs = append(inputs, redemption.Spends.Coin.ID())
 	}
-	outputCoin := eth.createAmountCoin(redeemedValue)
```
chappjc, line 702: Just a build error on master,  otherwise unrelated.

client/asset/eth/eth.go
```diff
@@ -677,13 +688,32 @@ func (eth *ExchangeWallet) Redeem(form *asset.RedeemForm) ([]dex.Bytes, asset.Co
...
 	var redeemedValue uint64
-	for _, redemption := range form.Redemptions {
-		var secretHash, secret [32]byte
-		copy(secretHash[:], redemption.Spends.SecretHash)
+	for i, redemption := range form.Redemptions {
+		// NOTE: redemption.Spends.SecretHash is a dup of the hash extracted
+		// from redemption.Spends.Contract. Even for scriptable UTXO assets, the
+		// redeem script in this Contract field is redundant with the SecretHash
+		// field as ExtractSwapDetails can be applied to extract the hash.
+		ver, secretHash, err := dexeth.DecodeContractData(redemption.Spends.Contract)
```
chappjc, line 699: Note that `AuditContract` is still in PR, but see also my comment: https://github.com/decred/dcrdex/pull/1319/files#r764101981

client/asset/eth/eth_test.go
```diff
@@ -944,12 +943,16 @@ func TestSwap(t *testing.T) {
...
-			if swaps.AssetVersion != binary.BigEndian.Uint32(contractData[:4]) {
+			ver, secretHash, err := dexeth.DecodeContractData(contractData)
+			if err != nil {
+				t.Fatalf("failed to decode contract data: %v", err)
+			}
+			if swaps.AssetVersion != ver {
 				t.Fatal("wrong contract version")
 			}
-			if !bytes.Equal(contractData[4:], contract.SecretHash[:]) {
+			if !bytes.Equal(contractData[4:], secretHash[:]) {
```
chappjc, line 953: hah, this check is silly now.

server/asset/eth/coiner.go
```diff
@@ -117,7 +117,7 @@ func (backend *Backend) newSwapCoin(coinID []byte, contract []byte, sct swapCoin
 			return nil, fmt.Errorf("tx %v does not contain initiation with secret hash %x", txHash, secretHash)
 		}
 		counterParty = &initiation.Participant
-		secretHash = initiation.SecretHash
+		secretHash = initiation.SecretHash // no-op
```
chappjc, line 120: Actually, I think it would be right to compare the secretHash from the contract data against the one from the call data here.  That's why I removed `validateRedeem`.

client/asset/eth/eth.go
```diff
@@ -586,9 +591,12 @@ func (r *swapReceipt) Coin() asset.Coin {
 	}
 }
 
-// Contract returns the swap's secret hash.
+// Contract returns the swap's identifying data, which the concatenation of the
+// contract version and the secret hash.
 func (r *swapReceipt) Contract() dex.Bytes {
-	return versionedBytes(r.ver, r.secretHash[:])
+	var secretHash [32]byte
+	copy(secretHash[:], r.secretHash)
```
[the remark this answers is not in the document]
buck54321: How about making `swapReceipt.secretHash` a `[32]byte`?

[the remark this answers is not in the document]
chappjc: Hmm, yah that actually worked out with minimal fallout.

client/core/trade.go
```diff
@@ -1529,7 +1529,11 @@ func (c *Core) swapMatchGroup(t *trackedTrade, matches []*matchTracker, errs *er
 		// to notify the server of the swap.
 		proof := &match.MetaData.Proof
 		contract, coinID := receipt.Contract(), []byte(coin.ID())
-		proof.Script = contract
+		// NOTE: receipt.Contract() uniquely identifies this swap. Only the
+		// asset backend can decode this information, which may be a redeem
+		// script with UTXO assets, or a secret hash + contract version for
+		// contracts on account-based assets.
+		proof.Script = contract // expected conflict: buck is renaming this
```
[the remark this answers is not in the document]
buck54321: Renamed `ContractData` in #1334.

client/asset/interface.go
```diff
@@ -335,9 +343,6 @@ type RedeemForm struct {
 	// suggestion in any way, but obviously fees that are too low may result in
 	// the redemption getting stuck in mempool.
 	FeeSuggestion uint64
-	// AssetVersion is the swap protocol version, which may indicate a specific
-	// contract or form of contract.
-	AssetVersion uint32
```
buck54321, line 340: Nice. I was hoping.

dex/networks/eth/params.go
buck54321: It's good. The use of `contractData` is becoming consistent.

buck54321 (approved):

PR #1321, opened 2021-12-01
client/asset/eth: fix client balance calcs

buck54321:
The client apparently doesn't maintain a pending state, so the
way I had calculated balance client-side didn't work. Go back
to @martonp's initial implementation which parses redeem
and refund data from transactions.

Remove the `Pending` field from `Balance`. It wasn't used.

### DIFF
--- a/client/asset/eth/nodeclient.go
+++ b/client/asset/eth/nodeclient.go
@@ -196,7 +196,7 @@ func (n *nodeClient) balance(ctx context.Context) (*Balance, error) {
 		return nil, fmt.Errorf("pending balance error: %w", err)
 	}
 
-	pendingTxs, err := n.pendingTransactions(ctx)
+	pendingTxs, err := n.pendingTransactions()
 	if err != nil {
 		return nil, fmt.Errorf("error getting pending txs: %w", err)
 	}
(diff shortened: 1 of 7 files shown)


client/asset/eth/nodeclient.go
```diff
@@ -264,7 +271,11 @@ func (n *nodeClient) transactionReceipt(ctx context.Context, txHash common.Hash)
 	if len(receipts) <= int(index) {
 		return nil, fmt.Errorf("no receipt at index %d in block %s for tx %s", index, blockHash, txHash)
 	}
-	return receipts[index], nil
+	receipt := receipts[index]
+	if receipt == nil {
+		return nil, fmt.Errorf("nil receipt at index %d in block %s for tx %s", index, blockHash, txHash)
+	}
+	return receipt, nil
```
buck54321, line 278: I saw a `nil, nil` here a while back. Been meaning to do something about it.

client/asset/eth/nodeclient.go
```diff
@@ -207,20 +202,32 @@ func (n *nodeClient) balance(ctx context.Context) (*Balance, error) {
...
-		outgoing.Add(outgoing, tx.Value())
+		gas := new(big.Int).SetUint64(tx.Gas())
+		fees := new(big.Int).Mul(gas, tx.GasFeeCap())
+		outgoing.Add(outgoing, fees)
+		v := tx.Value()
+		if v.Cmp(zero) == 0 {
+			for ver, c := range n.contractors {
+				in, err := c.incomingValue(ctx, tx)
+				if err != nil {
+					n.log.Errorf("version %d contractor incomingValue error: %v", ver, err)
```
chappjc, line 218: and `continue`?

Comments on each side of the `if v.Cmp(zero) == 0 {` conditional would be nice too I think (e.g. if sending non-zero value add to outgoing; if zero value, attempt to detect contract interaction that pays to the account).

client/asset/eth/nodeclient.go
```diff
@@ -207,20 +202,32 @@ func (n *nodeClient) balance(ctx context.Context) (*Balance, error) {
 	}
 
 	outgoing := new(big.Int)
+	incoming := new(big.Int)
+	zero := new(big.Int)
 	for _, tx := range pendingTxs {
 		// TODO: Check that the from address matches addr. types.Transaction
 		// does not expose the from address directly.
```
chappjc, line 209: Think it would be this:

```go
ethSigner := types.LatestSigner(n.leth.ApiBackend.ChainConfig()) // "latest" good for pending
...
for _, tx := range pendingTxs {
	from, _ := ethSigner.Sender(tx) // zero Address on error
	if from != n.address() {
		continue
	}
```

chappjc:
Mostly unrelated to this PR, but there are a lot of unused Context args to `ethFetcher` methods, including `pendingTransactions`.  Is there a conceivable use for these with another ethFetcher impl?

client/asset/eth/nodeclient.go
```diff
@@ -207,20 +202,32 @@ func (n *nodeClient) balance(ctx context.Context) (*Balance, error) {
...
-		outgoing.Add(outgoing, tx.Value())
+		gas := new(big.Int).SetUint64(tx.Gas())
+		fees := new(big.Int).Mul(gas, tx.GasFeeCap())
```
martonp, line 211: Should we also handle legacy transactions (`GasPrice` is populated and not `GasFeeCap`)? This would only be an issue if someone imports their seed into another wallet.

JoeGruffins (approved):

chappjc (approved):

PR #1322, opened 2021-12-02
server/asset/eth: Fix run.

JoeGruffins:
Only cache the best hash and use it to determine if we see a reorg. Move
    polling and notification logic to hashCache.

part of #1154


server/asset/eth/eth_test.go
```diff
@@ -225,31 +212,27 @@ func TestDecodeCoinID(t *testing.T) {
...
+	node := &testNode{
+		bestHdr: &types.Header{Number: big.NewInt(0)},
+	}
 	backend.node = node
+	backend.hashCache.prime(ctx, node)
+	*node = testNode{
+		bestHdr: &types.Header{
+			Number: big.NewInt(1),
+		},
+	}
```
buck54321, line 227: Why not `node.bestHdr = &types.Header{Number: big.NewInt(1)}`?

server/asset/eth/eth_test.go
```diff
@@ -225,31 +212,27 @@ func TestDecodeCoinID(t *testing.T) {
...
-	backend.blockCache.mtx.Lock()
-	best := backend.blockCache.best
-	backend.blockCache.mtx.Unlock()
-	if best.hash != blockHash1 {
-		t.Fatalf("want header hash %x but got %x", blockHash1, best.hash)
-	}
-	cancel()
+	// This timing out indicates failure. run should call a block update
+	// after the pollInterval and we have the listening function cancel the
+	// context above.
```
buck54321, line 235: Hmm. The `context.Context` doesn't have a timeout on it and I don't see how you differentiate a timeout from a success. Do you mean the default test timeout of 10 minutes which results in a panic?

JoeGruffins: yes, a panic. Can add a shorter timeout and real failure.

server/asset/eth/eth.go
```diff
@@ -154,9 +141,8 @@ func unconnectedETH(logger dex.Logger, cfg *config) *Backend {
 		rpcCtx:       ctx,
 		cancelRPCs:   cancel,
 		cfg:          cfg,
-		blockCache:   newBlockCache(logger),
+		hashCache:    newHashCache(logger),
```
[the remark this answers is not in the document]
JoeGruffins: Sure.

server/asset/eth/eth_test.go
JoeGruffins: yeah, that probably makes more sense

JoeGruffins: Changed.

server/asset/eth/eth_test.go
JoeGruffins: Shortened.

server/asset/eth/eth.go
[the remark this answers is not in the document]
JoeGruffins: Also removed the constructor...

server/asset/eth/eth_test.go
```diff
@@ -214,25 +214,32 @@ func TestDecodeCoinID(t *testing.T) {
...
-	// after the pollInterval and we have the listening function cancel the
-	// context above.
+	// Ok if ctx was canceled above. Linters complain about calling t.Fatal
+	// in the goroutine above.
+	select {
+	case <-ctx.Done():
+		return
+	default:
+		t.Fatal("test timeout")
+	}
```
JoeGruffins: If I put t.Fatal in the goroutine govet would tell me `Error: testinggoroutine: call to (*T).Fatal from a non-test goroutine (govet)` and the wording suggests there's a non-test goroutine? Is there a way to fail in the goroutine?

chappjc: I haven't looked at this pr at all yet, but generally you have to signal back to the main testing goroutine.

JoeGruffins: I see, so the `TestSomething(t *testing.T)` itself is the goroutine it is referring to.

server/asset/eth/cache.go
```diff
@@ -7,130 +7,107 @@
...
+			str += " Detected reorg."
+		case fastBlocks:
+			str += " Fast blocks."
 		}
+		hc.log.Info(str)
+		hc.best.hash = hash
+		hc.best.height = height
+	}
+	if bhdr.ParentHash == hc.best.hash {
+		// Sequencial hash, report a block update.
```
buck54321, line 87: Sequential. Here and one place in tests.

buck54321 (changes requested):
Couple tiny things.

We should consider some kind of block update metering, maybe in the consumer though. I imagine ethereum blocks come in pretty tight clusters sometimes.

server/asset/eth/eth.go
```diff
@@ -185,21 +175,22 @@ func (eth *Backend) Connect(ctx context.Context) (*sync.WaitGroup, error) {
...
+		log:        eth.log.SubLogger("hashcache"),
+		node:       &c,
+		signalMtx:  &eth.signalMtx,
+		blockChans: eth.blockChans,
+		best: hashN{
+			height: hdr.Number.Uint64(),
+			hash:   hdr.Hash(),
+		},
 	}
+	eth.hashCache = hc
```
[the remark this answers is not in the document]
JoeGruffins: Changed.

server/asset/eth/cache.go
JoeGruffins: Fixed.

JoeGruffins:
What, tests are passing for me locally.

chappjc:
The bug on master is fixed here: https://github.com/decred/dcrdex/pull/1320#discussion_r764194657

JoeGruffins:
rebased to correct a commit typo

buck54321 (approved):

JoeGruffins:
Rebased to fix the test error.

chappjc (changes requested):

server/asset/eth/eth.go
```diff
@@ -184,20 +176,20 @@ func (eth *Backend) Connect(ctx context.Context) (*sync.WaitGroup, error) {
...
-		return nil, fmt.Errorf("error getting best block from geth: %w", err)
-	}
-	_, err = eth.blockCache.add(block)
-	if err != nil {
-		eth.log.Errorf("error adding new best block to cache: %v", err)
+	eth.hashCache = &hashCache{
+		log:        eth.log.SubLogger("hashcache"),
+		node:       &c,
+		signalMtx:  &eth.signalMtx,
+		blockChans: eth.blockChans,
```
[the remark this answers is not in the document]
JoeGruffins: We don't still need to hold a read lock for ranging over the `hc.blockChans` map?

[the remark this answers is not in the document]
chappjc: `hashCache` has it's own mutex, and the use pattern in `poll` as written is that it will always be locked when both `blockChans` and `best` are accessed.  Do you need two mutexes?

server/asset/eth/eth.go
```diff
@@ -114,9 +108,8 @@ type Backend struct {
 	// method. signalMtx locks the blockChans array.
 	signalMtx  sync.RWMutex
 	blockChans map[chan *asset.BlockUpdate]struct{}
-	// The block cache stores just enough info about the blocks to prevent future
-	// calls to Block.
-	blockCache *blockCache
+	// The hash cache stores just enough info to detect reorgs.
```
chappjc, line 111: `... it also handles block polling and new block notifications to BlockChannel consumers.`

server/asset/eth/eth.go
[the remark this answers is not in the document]
JoeGruffins: Yeah, you are right.

So originally this was created with the unconnected wallet, so a lot of the wierdness is to make the exported `BlockChannel` not panic before hashCache is created in `Connect`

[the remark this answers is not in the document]
JoeGruffins: I think I'll just remove the type since it doesn't really do that much anymore.

server/asset/eth/eth.go
JoeGruffins: Removed the blockCache in the last commit. Let me know how it looks.

chappjc: Looks like an improved design to me.  In `run` there is a polling loop, and if that's the only caller of `poll` and it's a synchronous call, then the fields of `best` don't need a mutex any more as it appears you've concluded as well.  And only needs to lock `blockChansMtx` if something actually changes, not each call to `poll`.

:+1: 

That the idea?

JoeGruffins: Yes.

chappjc (approved):

PR #1323, opened 2021-12-03
[backport/release-v0.4] server/market: fix wrong asset for RequiredOrderFunds on market-buy

buck54321:
Twas an old bug.

### DIFF
--- a/server/market/orderrouter.go
+++ b/server/market/orderrouter.go
@@ -576,7 +576,7 @@ func (r *OrderRouter) handleMarket(user account.AccountID, msg *msgjson.Message)
 			buyBuffer := tunnel.MarketBuyBuffer()
 			lotWithBuffer := uint64(float64(lotSize) * buyBuffer)
 			minReq := matcher.BaseToQuote(midGap, lotWithBuffer)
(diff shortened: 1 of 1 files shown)


server/market/orderrouter.go
```diff
@@ -576,7 +576,7 @@ func (r *OrderRouter) handleMarket(user account.AccountID, msg *msgjson.Message)
 			buyBuffer := tunnel.MarketBuyBuffer()
 			lotWithBuffer := uint64(float64(lotSize) * buyBuffer)
 			minReq := matcher.BaseToQuote(midGap, lotWithBuffer)
-			reqVal = calc.RequiredOrderFunds(minReq, uint64(spendSize), 1, &assets.base.Asset)
+			reqVal = calc.RequiredOrderFunds(minReq, uint64(spendSize), 1, &assets.quote.Asset)
```
chappjc, line 579: assetSet also has `funding` and `receiving` aliases, as used in the `sell` case above.  I think we have those to reduce mental gymnastics in cases like this so might as well use them.
This is the "funding" asset of interest here, correct?  If so, we should use `assets.funding.Asset`.

chappjc (changes requested):

chappjc (approved):

PR #1325, opened 2021-12-05
eth/client: Refund

martonp:
client/asset/eth:
- implement the refund function
- update coin locking to just keep track of a uint64 instead of
   a map of unique coin ids

dex/networks/eth/swap0:
- create `isRedeemable` view function

Part of #1154


client/asset/eth/eth_test.go
```diff
@@ -1262,6 +1274,87 @@ func TestMaxOrder(t *testing.T) {
...
+		_, err := eth.Refund(nil, test.contract, test.maxFeeRate)
+		if test.wantErr {
+			if err == nil {
+				t.Fatalf(`%v: expected error but did not get: %v`, test.name, err)
+			}
+			continue
+		}
+		if err != nil {
+			t.Fatalf(`%v: unexpected error: %v`, test.name, err)
+		}
```
martonp, line 1353: @buck54321 I would want to check that the hash of the tx used to do the refund is the same as the coin id returned from `Refund`. I'm wondering what you think about what we were discussing in this PR:
https://github.com/decred/dcrdex/pull/1318

Personally I would just add a `lastRefundTxHash` to the `testNode` and use that to compare, but if you're planning to do something else, I'll hold off on this.

buck54321: Please do @martonp. I got the things I though were most important in #1318, but if you see a place to improve, you should.

dex/networks/eth/contracts/ETHSwapV0.sol
```diff
@@ -199,10 +198,11 @@ contract ETHSwap {
     function refund(bytes32 secretHash)
         public
         senderIsOrigin()
-        isRefundable(secretHash)
     {
-        swaps[secretHash].state = State.Refunded;
-        (bool ok, ) = payable(msg.sender).call{value: swaps[secretHash].value}("");
+        require(isRefundable(secretHash));
+        Swap storage swapToRefund = swaps[secretHash];
```
chappjc, line 203: I don't expect it's significant, but I'm curious how refund gas costs change with this.  Refund is also not the usual swap outcome, so it's not critical, just wondering.

martonp: This change actually saves a little gas, 42822 -> 42740.
The function call increased the gas, but using the local storage variable saved some.

JoeGruffins (commented):
I don't think this needs the `maxfeerate` argument. I think it should be using the user's set max fee.

In other coins, the user set max fee is tested against the server's value in `FundOrder`:
https://github.com/decred/dcrdex/blob/6c947e08bee15ea51254c2cc9c1a4fce9986b16a/client/asset/dcr/dcr.go#L819-L826

The user set max fee should be higher than the server's, so this isn't super bad to do, but imo we should just use the default constant `defaultGasFeeLimit` here and change it when fixing #1336

dex/networks/eth/contracts/ETHSwapV0.sol
martonp: It doesn't help on isRedeemable because there are only 2 lookups there instead of the 3 here.

martonp:
#1344 should go before this.

buck54321 (changes requested):

client/asset/eth/eth.go
```diff
@@ -169,6 +168,17 @@ type ethFetcher interface {
 // Check that ExchangeWallet satisfies the asset.Wallet interface.
 var _ asset.Wallet = (*ExchangeWallet)(nil)
 
+// lockedAccount pairs an ethereum address with a locked amount.
+type lockedAddr struct {
+	address   common.Address
+	locked    uint64
+	lockedMtx sync.RWMutex
+}
```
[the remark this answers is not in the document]
martonp: I was kind of replicating @chappjc old PR, this would allow us to support multiple addresses in the wallet, but we don't really need it now though.

buck54321 (approved):

chappjc:
Ok, all set for this.  Please pull master though before rebase because there are conflicts.

chappjc:
Also, you may squash to the merge-base _before_ rebasing, other wise you'll have two batches of conflicts to resolve from the two commits.

client/asset/eth/eth.go
```diff
@@ -1000,8 +966,29 @@ func (eth *ExchangeWallet) findSecret(ctx context.Context, secretHash [32]byte,
 
 // Refund refunds a contract. This can only be used after the time lock has
 // expired.
-func (*ExchangeWallet) Refund(coinID, contract dex.Bytes, feeSuggestion uint64) (dex.Bytes, error) {
-	return nil, asset.ErrNotImplemented
+func (eth *ExchangeWallet) Refund(_, contract dex.Bytes, feeSuggestion uint64) (dex.Bytes, error) {
+	version, secretHash, err := dexeth.DecodeContractData(contract)
+	if err != nil {
+		return nil, fmt.Errorf("Refund: failed to decode contract: %w", err)
```
JoeGruffins, line 972: Just a silly suggestion, so feel free to ignore. If you would like to put the method name in every error message, you could define it at the beginning like `const op = "Refund"` then do `fmt.Errorf("%s: failed: %v", op, err)`

Really just a thing I have seen ppl do if you like it. Probably prevents some typos and repitition? If the method name were to change can get all the messages at once. No need to change at all though, just a suggestion.

dex/networks/eth/contracts/v0/BinRuntimeV0.go
```diff
@@ -3,4 +3,4 @@
...
...fffffffffffff8082111561082157600080fd5b818501915085601f83011261083557600080fd5b81358181111561084457600080fd5b8660208260061b85010111156107e457600080fd5b60006020828403121561086b57600080fd5b5035919050565b60006020828403121561088457600080fd5b5051919050565b6000806040838503121561089e57600080fd5b50508035926020909101359150565b600481106108cb57634e487b7160e01b600052602160045260246000fd5b9052565b6000825160005b818110156108f057602081860181015185830152016108d6565b818111156108ff576000828501525b509190910192915050565b8781526020810187905260408101869052606081018590526001600160a01b038481166080830152831660a082015260e0810161094a60c08301846108ad565b98975050505050505050565b600060e08201905082518252602083015160208301526040830151604083015260608301516060830152608083015160018060a01b0380821660808501528060a08601511660a0850152505060c08301516109b460c08401826108ad565b5092915050565b600082198211156109ce576109ce6109ee565b500190565b60006000198214156109e7576109e76109ee565b5060010190565b634e487b7160e01b600052601160045260246000fd5b634e487b7160e01b600052602160045260246000fd5b634e487b7160e01b600052603260045260246000fdfea2646970667358221220a36f1109c6419ae7a881c27974458b5cf52dbb692ee3f12b0e0f81ca0e4dde6d64736f6c63430008060033"
```
JoeGruffins, line 6: This is not about the pr.

I thought that changing this value would change the tx that makes the contract and thereby change the contract address, but it does not. I wonder what decides contract addresses.

client/asset/eth/eth.go
```diff
@@ -1000,8 +966,29 @@ func (eth *ExchangeWallet) findSecret(ctx context.Context, secretHash [32]byte,
...
+		return nil, fmt.Errorf("Refund: failed to check isRefundable: %w", err)
+	}
+	if !refundable {
+		return nil, fmt.Errorf("Refund: swap with secret hash %x is not refundable", secretHash)
+	}
+
+	tx, err := eth.node.refund(eth.ctx, secretHash, feeSuggestion, version)
+	if err != nil {
+		return nil, fmt.Errorf("Refund: failed to call refund: %w", err)
+	}
```
JoeGruffins, line 986: This uses the server's feeSuggestion as the maxFee.

So... With the new dynamic txs, should the suggestion be the TipCap? We cannot choose the base fee, and I personally think the overall max should be the user's chosen max in all places.

Correct me if my understand is wrong. You have the base fee which is out of our control, must be paid. You have the tip or priority fee which we should put at 2 gwei basically, but this is the fee as we usually know it. Then we can choose the max that base plus tip cannot go over.

chappjc: For refunds, it would be fine to ignore server's feeSuggestion for `GasFeeCap` (max base fee) and instead use the user's specified own.  But this arg to `(*nodeClient).refund` ends up in `bind.TransactionOpts.GasFeeCap`, and the tip is requested on the fly by `txOpts` > `netFeeState`.

Regarding the units of the server's provided fee rates, the server's `(*Backend).FeeRate` returns a suggestion that is (*current* base + suggested fee tip cap).  See the call to `suggestGasPrice` that is handled by `(*PublicEthereumAPI).GasPrice` in internal/ethapi/api.go.    This is documented as "for legacy transactions" for obvious reasons -- it sums the two rates.  This does raise some questions.

First is that we've talked about the server-prescribed and suggested rates, which find their way to the `GasFeeCap` for dynamic client txns, being based on some **multiple of the current base fee rate** (e.g. 125%) rather than just adding the current suggested tip to the current base fee, like the `PublicEthereumAPI` method does as described above.  We only have to look at how base fee swings from one block to the next

Second is that for the purposes of the init txns, the prescribed fee rate for matches must currently refer to the max gas fee cap because if the user's unconfirmed init txn specifies any less, then it should rejected.  See how `server/asset/eth.(*Backend).newSwapCoin` uses `tx.GasPrice()`, which for dynamic txns returns `GasFeeCap`.  A related shower thought is that the server's `(*Backend).FeeRate` method could return a second fee value so the tip cap can also be enforced, but I have a feeling just the gas cap is fine.

chappjc: Sorry, that was almost completely off-topic.  This is the Refund PR, and yes the `(*nodeClient).refund` method can be changed to use `feeSuggestion` only as a fallback or not at all, if it determines a reasonable base fee cap locally.

PR #1326, opened 2021-12-05
finalize ETH coin IDs

chappjc:
This completes the change of definition of ETH's "coin ID" to be a transaction hash almost universally. This means:
- For client-server communications and everywhere server-side, and ETH coin ID can only be decoded as a transaction hash.  In the new dex/networks/eth/common.go, there is a new `DecodeCoinID` helper for this definition of an ETH coin ID.
- Client-side, there is also a `FundingCoinID`, previously known as `AmountCoinID` in the removed cointypes.go, for internal tracking of funding amounts.  The polymorphic `CoinID` with a byte as a type flag (`CoinIDFlag`) is now removed however.

For this encoding change to work, `server/asset/btc.Backend.Redemption` is given a new input argument for "contract data", which in the case of ETH is a blob encoding contract version and the secret hash.  This has been referred to elsewhere as "versionedBytes" or "versionedHash", and which will be formalized in https://github.com/decred/dcrdex/pull/1320.

Related to the server's `Redemption` change, the `server/asset/Contract.RedeemScript []byte` field is renamed to `ContractData` and the corresponding inputs to the other methods are renamed accordingly, namely `server/asset/btc.Backend.{Contract,ValidateSecret,Redemption}`.  This consistent naming should make it clear that this is the same piece of data.


client/asset/eth/fundingcoin.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+const fundingCoinIDSize = 36 // address (20) + amount (8) + nonce (8) = 36
+
+// FundingCoinID is an identifier for a coin which has not yet been sent to the
+// swap contract.
+type FundingCoinID struct {
+	Address common.Address
+	Amount  uint64
+	Nonce   [8]byte
```
buck54321, line 24: I actually thought we were going to get rid of the `Nonce`. Do you see value in it?

chappjc: I think we can eliminate it, but with the present client-side coin locking with these as map keys, it has to stay for now.
I'd like another pr to revisit the client's funding coin locking, but I don't think it's critical.  I had closed the one I had previously started for discussion.
This pr at least isolates this funding coin notion to client packages.

chappjc: Sorry, I should have pointed out in the PR description that this type is essentially the old `AmountCoinID` from the now removed cointypes.go, just with the coin type flag removed.

buck54321 (approved):
This looks good. I'll be watching for #1320 to come out of draft.

client/asset/eth/fundingcoin.go
martonp: I can remove the nonce in #1325. The Refund function will not receive the funding coin ids used to reserve funds to use for gas in case of a refund, so the locked funds map will need to be removed and replaced by a uint64 as @chappjc had done in #1289.

martonp (approved):
LGTM

dex/networks/eth/common.go
```diff
@@ -0,0 +1,49 @@
...
+
+// SecretHashSize is the byte-length of the hash of the secret key used in
+// swaps.
+const SecretHashSize = 32
+
+// ToGwei converts a *big.Int in wei (1e18 unit) to gwei (1e9 unit) as a uint64.
+// Errors if the amount of gwei is too big to fit fully into a uint64.
+func ToGwei(wei *big.Int) (uint64, error) {
+	gweiFactorBig := big.NewInt(GweiFactor)
+	gwei := big.NewInt(0).Div(wei, gweiFactorBig)
```
JoeGruffins, line 37: Maybe not the best time to ask, but can you slip in a check that wei is not negative here? I believe negative will still fail below but with the wrong error message.

server/asset/eth/coiner.go
```diff
@@ -239,7 +246,7 @@ func (c *swapCoin) Confirmations(_ context.Context) (int64, error) {
 
 		// The swap initiation transaction has some number of
 		// confirmations, and we are sure the secret hash belongs to
-		// this swap. Assert that the value, reciever, and locktime are
+		// this swap. Assert that the value, receiver, and locktime are
```
JoeGruffins, line 249: God my spelling is awful. Srry.

JoeGruffins (approved):

PR #1332, opened 2021-12-07
client/eth: implement Exists

buck54321:
`Exists` works by trying to load the keystore and checking for > 1
wallets. Remove unused `"appdir"` setting from custom config. We already
have `CreateWalletParams.DataDir`.

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -28,6 +28,7 @@ import (
 	"github.com/decred/dcrd/hdkeychain/v3"
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/accounts/keystore"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto"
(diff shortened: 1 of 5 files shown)


client/asset/eth/eth.go
```diff
@@ -126,7 +123,12 @@ func (d *Driver) Info() *asset.WalletInfo {
 
 // Exists checks the existence of the wallet.
 func (d *Driver) Exists(walletType, dataDir string, settings map[string]string, net dex.Network) (bool, error) {
-	return false, errors.New("unimplemented")
+	if walletType != walletTypeGeth {
+		return false, fmt.Errorf("wallet type %q unrecognized", walletType)
+	}
+	keyStoreDir := filepath.Join(getWalletDir(dataDir, net), "keystore")
+	ks := keystore.NewKeyStore(keyStoreDir, keystore.LightScryptN, keystore.LightScryptP)
```
chappjc, line 130: There are three different magic values, including the new "keystore".  Can we make these package-level?

chappjc (approved):

chappjc:
Will allow more reviews since this isn't required for other urgent work.

JoeGruffins (approved):

JoeGruffins:
Have a test failure in #1335 that may or may not be related to these changes.

PR #1335, opened 2021-12-08
client/eth: Use bytes as string in key import.

JoeGruffins:
client/eth: Use bytes as string in key import.

    This change is needed because unlocking will use the bytes as a string.

    Also set the simnet contract address.

    client/eth: Remove unused config.


JoeGruffins:
Could also change Unlock to use the bytes encoded as hex, but other wallets look to just use string().

Also, I believe the simnet contract address is deterministic as long as it's sent from the same account with the same nonce and the same txdata.

Not sure about this test failure. Am not seeing locally.

Also, need to remove the comments in `dexc/importlgpl.go` to use in dexc.

JoeGruffins:
I ammended the commit and it's not failing now. Here was the test though https://github.com/decred/dcrdex/runs/4453620451?check_suite_focus=true

chappjc:
Can rebase with https://github.com/decred/dcrdex/pull/1332 merged.
The error you reported is of course https://github.com/decred/dcrdex/issues/1337, resolved by https://github.com/decred/dcrdex/pull/1341, as we know now.  Thank you for calling that out!

JoeGruffins:
> The error you reported

I was lucky to hit the 1/256

buck54321 (approved):

chappjc (approved):
In client/asset/driver.go, the definition of `CreateWalletParams` should probably have a comment to reduce ambiguity like `Pass     []byte  // utf-8`, but this PR is good.

PR #1339, opened 2021-12-09
eth: Versionize tx data handling

martonp:
The functions used to parse and pack call data for the eth swap contract only supported version 0. This diff makes those functions version agnostic.

### DIFF
--- a/dex/networks/eth/txdata.go
+++ b/dex/networks/eth/txdata.go
@@ -10,40 +10,128 @@ import (
 	"fmt"
 	"math/big"
 	"strings"
+	"time"
 
 	swapv0 "decred.org/dcrdex/dex/networks/eth/contracts/v0"
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
 )
 
-const (
-	initiateFuncName = "initiate"
-	numInputArgs     = 1
-	redeemFuncName   = "redeem"
-	numRedeemArgs    = 1
-	refundFuncName   = "refund"
-	numRefundArgs    = 1
-)
+// ParseRefundData parses the calldata used to call the initiate function of a
+// specific version of the swap contract. It returns the the list of initiations
+// done in the call and errors if the call data does not call initiate with expected
+// argument types.
+func ParseInitiateData(calldata []byte, contractVersion uint32) ([]Initiation, error) {
+	txDataHandler, ok := txDataHandlers[contractVersion]
+	if !ok {
+		return nil, fmt.Errorf("contract version %v does not exist", contractVersion)
+	}
 
-// ParseInitiateData accepts call data from a transaction that pays to a
-// contract with extra data. It will error if the call data does not call
-// initiate with expected argument types. It returns the array of initiations
-// with which initiate was called.
-func ParseInitiateData(calldata []byte) ([]swapv0.ETHSwapInitiation, error) {
+	return txDataHandler.parseInitiateData(calldata)
+}
+
+// ParseRefundData parses the calldata used to call the redem function of a
+// specific version of the swap contract. It returns the the list of redemptions
+// done in the call and errors if the call data does not call redeem with expected
+// argument types.
+func ParseRedeemData(calldata []byte, contractVersion uint32) ([]Redemption, error) {
+	txDataHandler, ok := txDataHandlers[contractVersion]
+	if !ok {
+		return nil, fmt.Errorf("contract version %v does not exist", contractVersion)
+	}
+
+	return txDataHandler.parseRedeemData(calldata)
+}
+
+// ParseRefundData parses the calldata used to call the refund function of a
+// specific version of the swap contract. It returns the secret hash and errors
+// if the call data does not call refund with expected argument types.
+func ParseRefundData(calldata []byte, contractVersion uint32) ([32]byte, error) {
+	txDataHandler, ok := txDataHandlers[contractVersion]
+	if !ok {
+		return [32]byte{}, fmt.Errorf("contract version %v does not exist", contractVersion)
+	}
+
+	return txDataHandler.parseRefundData(calldata)
+}
+
+// PackInitiateData converts a list of Initiation to the call data for the
+// initiate function for the contract specified by contractVersion.
+func PackInitiateData(initiations []Initiation, contractVersion uint32) ([]byte, error) {
+	txDataHandler, ok := txDataHandlers[contractVersion]
+	if !ok {
+		return nil, fmt.Errorf("contract version %v does not exist", contractVersion)
+	}
+
+	return txDataHandler.packInitiateData(initiations)
+}
+
+// PackRedeemData converts a list of Redemption to the call data for the
+// redeem function for the contract specified by contractVersion.
+func PackRedeemData(redemptions []Redemption, contractVersion uint32) ([]byte, error) {
+	txDataHandler, ok := txDataHandlers[contractVersion]
+	if !ok {
+		return nil, fmt.Errorf("contract version %v does not exist", contractVersion)
+	}
+
+	return txDataHandler.packRedeemData(redemptions)
+}
+
+// PackInitiateData converts a secret hash to the call data for the
+// refund function for the contract specified by contractVersion.
+func PackRefundData(secretHash [32]byte, contractVersion uint32) ([]byte, error) {
+	txDataHandler, ok := txDataHandlers[contractVersion]
+	if !ok {
+		return nil, fmt.Errorf("contract version %v does not exist", contractVersion)
+	}
+
+	return txDataHandler.packRefundData(secretHash)
+}
+
+var txDataHandlers = map[uint32]txDataHandler{
+	0: &txDataHandlerV0{
+		initiateFuncName: "initiate",
+		redeemFuncName:   "redeem",
+		refundFuncName:   "refund",
+		numInputArgs:     1,
+		numRedeemArgs:    1,
+		numRefundArgs:    1,
+	},
+}
+
+type txDataHandler interface {
+	parseInitiateData([]byte) ([]Initiation, error)
+	parseRedeemData([]byte) ([]Redemption, error)
+	parseRefundData([]byte) ([32]byte, error)
+	packInitiateData([]Initiation) ([]byte, error)
+	packRedeemData([]Redemption) ([]byte, error)
+	packRefundData([32]byte) ([]byte, error)
+}
+
+type txDataHandlerV0 struct {
+	initiateFuncName string
+	redeemFuncName   string
+	refundFuncName   string
+	numInputArgs     int
+	numRedeemArgs    int
+	numRefundArgs    int
+}
+
+func (t *txDataHandlerV0) parseInitiateData(calldata []byte) ([]Initiation, error) {
 	decoded, err := parseCallData(calldata, swapv0.ETHSwapABI)
 	if err != nil {
 		return nil, fmt.Errorf("unable to parse call data: %v", err)
 	}
-	if decoded.name != initiateFuncName {
-		return nil, fmt.Errorf("expected %v function but got %v", initiateFuncName, decoded.name)
+	if decoded.name != t.initiateFuncName {
+		return nil, fmt.Errorf("expected %v function but got %v", t.initiateFuncName, decoded.name)
 	}
 	args := decoded.inputs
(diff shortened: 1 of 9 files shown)


martonp:
Just as a note, `PackInitiateData(ParseInitiateData(data))` will not be equal to data anymore due to the loss of precision when converting to the value to uint64. I can't think of a case where this is an issue, but I'm wondering if anyone disagrees.

chappjc:
https://github.com/decred/dcrdex/pull/1319 is merged

dex/networks/eth/txdata.go
```diff
@@ -128,31 +234,44 @@ func ParseRefundData(calldata []byte) ([32]byte, error) {
 	return secretHash, nil
 }
 
-// PackInitiateData converts a list of swapv0.ETHSwapInitiation to call data for the
-// initiate function.
-func PackInitiateData(initiations []swapv0.ETHSwapInitiation) ([]byte, error) {
+// packInitiateData converts a list of Initiate to call data for the initiate function.
+func (t *txDataHandlerV0) packInitiateData(initiations []Initiation) ([]byte, error) {
 	parsedAbi, err := abi.JSON(strings.NewReader(swapv0.ETHSwapABI))
```
chappjc, line 239: Thoughts about doing this `parsedAbi, err := abi.JSON(strings.NewReader(swapv0.ETHSwapABI))` just once in either an `init()` function or a `txDataHandlerV0` constructor?
It always spits out the same `parsedAbi`, and it's in three different methods.

chappjc (changes requested):
Looks pretty good overall.  Just some copy pasta to fix and some minor design questions.

dex/networks/eth/txdata.go
```diff
@@ -10,40 +10,128 @@ import (
...
+func ParseInitiateData(calldata []byte, contractVersion uint32) ([]Initiation, error) {
+	txDataHandler, ok := txDataHandlers[contractVersion]
+	if !ok {
+		return nil, fmt.Errorf("contract version %v does not exist", contractVersion)
+	}
+
+	return txDataHandler.parseInitiateData(calldata)
+}
+
+// ParseRefundData parses the calldata used to call the redem function of a
```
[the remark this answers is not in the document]
chappjc: Also `redem` spelling

dex/networks/eth/txdata.go
```diff
@@ -10,40 +10,128 @@ import (
...
+	packRefundData([32]byte) ([]byte, error)
+}
+
+type txDataHandlerV0 struct {
+	initiateFuncName string
+	redeemFuncName   string
+	refundFuncName   string
+	numInputArgs     int
+	numRedeemArgs    int
+	numRefundArgs    int
```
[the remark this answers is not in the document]
martonp: There was no real reason. Previously they were defined as constants outside of the functions as well, and I just kind of imitated the previous implementation, but I agree, better to have them in the functions.

dex/networks/eth/txdata.go
```diff
@@ -55,33 +143,45 @@ func ParseInitiateData(calldata []byte) ([]swapv0.ETHSwapInitiation, error) {
 		return nil, fmt.Errorf("expected first arg of type []swapv0.ETHSwapInitiation but got %T", args[0].value)
 	}
 
-	toReturn := make([]swapv0.ETHSwapInitiation, 0, len(initiations))
+	// This is done for the compiler to ensure that the type defined above and swapv0.ETHSwapInitiation
+	// are the same, other than the tags.
+	if len(initiations) > 0 {
+		_ = swapv0.ETHSwapInitiation(initiations[0])
```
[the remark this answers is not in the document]
martonp: When doing the type assertion, it must be the exact same type, including the name of the type. If it's anonymous struct, asserting with a named struct, even if it contains the same fields/tags, will not work:
https://go.dev/play/p/hJO53Ptj0Af

> But also doesn't the assertion `args[0].value.([]struct...` do the same thing effectively?

What do you mean? The reason for this is so the compiler will tell us to change this if we change the solidity code. This line won't do that.

dex/networks/eth/txdata.go
```diff
@@ -55,33 +143,45 @@ func ParseInitiateData(calldata []byte) ([]swapv0.ETHSwapInitiation, error) {
 		return nil, fmt.Errorf("expected first arg of type []swapv0.ETHSwapInitiation but got %T", args[0].value)
 	}
 
-	toReturn := make([]swapv0.ETHSwapInitiation, 0, len(initiations))
+	// This is done for the compiler to ensure that the type defined above and swapv0.ETHSwapInitiation
+	// are the same, other than the tags.
```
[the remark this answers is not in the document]
martonp: By rewrap, you mean fix the length right? <=80?

dex/networks/eth/txdata.go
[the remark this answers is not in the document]
martonp: I'll leave the function names though since they are reused.

dex/networks/eth/txdata.go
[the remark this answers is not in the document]
chappjc: Right.

dex/networks/eth/txdata.go
```diff
@@ -10,40 +10,128 @@ import (
...
+		numRefundArgs:    1,
+	},
+}
+
+type txDataHandler interface {
+	parseInitiateData([]byte) ([]Initiation, error)
+	parseRedeemData([]byte) ([]Redemption, error)
+	parseRefundData([]byte) ([32]byte, error)
+	packInitiateData([]Initiation) ([]byte, error)
+	packRedeemData([]Redemption) ([]byte, error)
```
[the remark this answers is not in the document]
martonp: No you're right, there's no reason to force copies.

dex/networks/eth/txdata.go
martonp: Yes that makes sense.

dex/networks/eth/txdata.go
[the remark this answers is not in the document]
chappjc: > including the name of the type

Ughhh, yes according to the spec "A **defined** type is always different from any other type. Otherwise, two types are identical if ... [the rest of the rules about fields and tags]"  https://go.dev/ref/spec#Type_identity

This all feels like such a hack using an anonymous type.  I think I know the answer, but again, I would like to ask the go-ethererum developers why there is no exported interface for inspecting the call data or why the exported type generated by abigen is not also json-tagged.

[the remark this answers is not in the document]
martonp: > I think I know the answer, but again, I would like to ask the go-ethererum developers why there is no exported interface for inspecting the call data or why the exported type generated by abigen is not also json-tagged.

What do you mean? You don't thing they'll do it?

dex/networks/eth/txdata.go
```diff
@@ -10,40 +10,134 @@ import (
...
-	redeemFuncName   = "redeem"
-	numRedeemArgs    = 1
-	refundFuncName   = "refund"
-	numRefundArgs    = 1
-)
+// ParseInitiateData parses the calldata used to call the initiate function of a
+// specific version of the swap contract. It returns the the list of initiations
+// done in the call and errors if the call data does not call initiate initiate
+// with expected argument types.
+func ParseInitiateData(calldata []byte, contractVersion uint32) ([]*Initiation, error) {
```
buck54321, line 24: How about a `map[[SecretHashSize]byte]*Initiation`?

dex/networks/eth/txdata.go
```diff
@@ -10,40 +10,134 @@ import (
...
+	}
+
+	return txDataHandler.parseInitiateData(calldata)
+}
+
+// ParseRedeemData parses the calldata used to call the redeem function of a
+// specific version of the swap contract. It returns the the list of redemptions
+// done in the call and errors if the call data does not call redeem with expected
+// argument types.
+func ParseRedeemData(calldata []byte, contractVersion uint32) ([]*Redemption, error) {
```
buck54321, line 37: `map[[SecretHashSize]byte]*Redemption`?

dex/networks/eth/txdata.go
```diff
@@ -10,40 +10,134 @@ import (
...
+	return txDataHandler.packRefundData(secretHash)
+}
+
+type txDataHandler interface {
+	parseInitiateData([]byte) ([]*Initiation, error)
+	parseRedeemData([]byte) ([]*Redemption, error)
+	parseRefundData([]byte) ([32]byte, error)
+	packInitiateData([]*Initiation) ([]byte, error)
+	packRedeemData([]*Redemption) ([]byte, error)
+	packRefundData([32]byte) ([]byte, error)
```
buck54321, line 97: Usually we try to leave code that is exclusive to a test in the test file. Is there a good reason to define these here?

dex/networks/eth/txdata.go
```diff
@@ -10,40 +10,134 @@ import (
...
+	return &txDataHandlerV0{
+		initiateFuncName: "initiate",
+		redeemFuncName:   "redeem",
+		refundFuncName:   "refund",
+		parsedAbi:        parsedAbi,
+	}
+}
+
+func (t *txDataHandlerV0) parseInitiateData(calldata []byte) ([]*Initiation, error) {
 	decoded, err := parseCallData(calldata, swapv0.ETHSwapABI)
```
buck54321, line 126: This is actually the part that's gotta be refactored. The ABI need to be initialized in a package-level `init` function, and then accessed via map lookup throughout, including the e.g. `newV0contractor`.

dex/networks/eth/txdata.go
```diff
@@ -55,33 +149,46 @@ func ParseInitiateData(calldata []byte) ([]swapv0.ETHSwapInitiation, error) {
...
 	args := decoded.inputs
 	// Any difference in number of args and types than what we expect
 	// should be caught by parseCallData, but checking again anyway.
 	//
 	// TODO: If any of the checks prove redundant, remove them.
-	if len(args) != numRedeemArgs {
-		return nil, fmt.Errorf("expected %v redeem args but got %v", numRedeemArgs, len(args))
+	numArgs := 1
+	if len(args) != numArgs {
+		return nil, fmt.Errorf("expected %v redeem args but got %v", numArgs, len(args))
```
buck54321, line 191: I don't know that this is an improvement, but if you're going to do this, do a `const numArgs = 1`

buck54321 (changes requested):

dex/networks/eth/txdata.go
chappjc: It was my opinion that having these integers as fields of these types made no sense given they can only ever be one specific value because the type is for a certain contract version.  `const` would be better though.

dex/networks/eth/txdata.go
chappjc: Ah, I neglected this part too.  `parseCallData` can take an `*abi.ABI` instead of a string and `&t.parsedAbi` can be provided to it.  I was ambivalent about package level vs txDataHandlerVX constructor though.

Looks like making this change will also lead to `verifySelector`, but that appear unused.

dex/networks/eth/txdata.go
buck54321: Oh. I guess I was thinking that it might as well just be `if len(args) != 1`, but I wasn't being very clear. With `const`s, it's ok.

dex/networks/eth/txdata.go
martonp: good point.. `packInitiateData` is used in a test in the client/eth package, but I could just duplicate that in the other test.

dex/networks/eth/txdata.go
martonp: I wanted to not touch that file since it was copied verbatim from the geth repo, but I guess there's no harm.

dex/networks/eth/txdata.go
```diff
@@ -155,26 +126,26 @@ func (t *txDataHandlerV0) parseInitiateData(calldata []byte) ([]*Initiation, err
 		_ = swapv0.ETHSwapInitiation(initiations[0])
 	}
 
-	toReturn := make([]*Initiation, 0, len(initiations))
+	toReturn := make(map[[SecretHashSize]byte]*Initiation)
```
chappjc, line 129: Should still allocate `len(initiations)`, but it's not gonna hold up my approval.

buck54321 (approved):

chappjc (approved):

Issue #1337, opened 2021-12-08
hdkeychain private key padding/length decision


chappjc:
Decred's `hdkeychain.(*ExtendedKey).Child` method is [known](https://github.com/decred/dcrd/blob/08528c28b7ef2d3ce59ea9a79579a01ac5d5cded/hdkeychain/extendedkey.go#L320-L324) to generate private keys [with zero padding removed and thus <32 bytes in length](https://github.com/btcsuite/btcutil/issues/172), at a rate of 1/256 hardened key derivations.

The btcsuite/btcutil version of `ExtendedKey` [was modified](https://github.com/btcsuite/btcutil/pull/182) by making a new `Derive` method and renaming `Child` to `DeriveNonStandard`.

We have been using Decred's hdkeychain, which should be fine as long as we are consistent, but another consideration is that in some cases we are using Decred's hdkeychain _internal to asset backends to derive keys for specific paths_.  For example, in `client/asset/eth`:

https://github.com/decred/dcrdex/blob/6c947e08bee15ea51254c2cc9c1a4fce9986b16a/client/asset/eth/eth.go#L207-L225

And the serialized private key, which may be less than 32 bytes, is used by `go-ethereum/crypto.ToECDSA` to initialize the ETH account's keystore.

https://github.com/decred/dcrdex/blob/6c947e08bee15ea51254c2cc9c1a4fce9986b16a/client/asset/eth/node.go#L257-L259

Because the eth functions have a [strict length requirement by default](https://github.com/ethereum/go-ethereum/blob/d8ff53dfb8a516f47db37dbc7fd7ad18a1e8a125/crypto/crypto.go#L132-L139), this fails occasionally as [we have seen on CI](https://github.com/decred/dcrdex/pull/1335#issuecomment-988542726) recently since this missing error check was [added less than 2 weeks ago](https://github.com/decred/dcrdex/commit/3991a0a17822b4daf5269ff783562ddc748eb2bc#diff-671406d18db2f8e1a281545be05fcdefd789b63614398d678d38e2b2814cf788R238).

While in this instance we might be able to use `ToECDSAUnsafe` to skip this check, **I think we should consider BIP-32 compliant key derivation, at least when used from inside the specific asset packages where we are trying to be compliant.**

On the other hand, I would like to stick with the dcrd/hdkeychain package if possible, so another idea might be to introduce a similar `DeriveBIP32Standard` method that simply omits the stripping of zero padding:

[...]

PR #1341, opened 2021-12-09
dex/keygen: use ChildBIP32Std  for extended key derivation

chappjc:
<s>Requires https://github.com/decred/dcrd/pull/2845</s>
Updates the `github.com/decred/dcrd/hdkeychain/v3` require to v3.1.0.

**WARNING:** Keys derived from the app seed are likely to be different with the updated strict BIP32 HD key derivation.  This includes
- DEX accounts
- Any keys derived by `Wallet`s such as in `client/asset/eth` where it prepares a private key for the BIP44 path m/44'/60'/0'/0/0
As such, you should not rely on any of these accounts or keys being recoverable.

This updates the `GenDeepChild` helper, which is used in both client/core to setup DEX account keys from the app seed, and in client/asset/eth to derive the account private key from the wallet seed.

This is especially in the eth key derivation, where the intent is to follow BIP44 with the path m/44'/60'/0'/0/0, which requires generating child keys as per BIP32 strictly.

For generating DEX account keys from the app seed, it is also good from a standardness point of view to use strict BIP32 derivation.  The legacy derivation is only required for Decred wallets for compatibility.

Resolves https://github.com/decred/dcrdex/issues/1337
This eliminates the occasional error from `client/asset/eth.CreateWallet`

### DIFF
--- a/dex/keygen/keygen.go
+++ b/dex/keygen/keygen.go
@@ -28,7 +28,7 @@ func GenDeepChild(seed []byte, kids []uint32) (*hdkeychain.ExtendedKey, error) {
 		err := hdkeychain.ErrInvalidChild
 		for err == hdkeychain.ErrInvalidChild {
 			var kid *hdkeychain.ExtendedKey
-			kid, err = parent.Child(childIdx)
+			kid, err = parent.ChildBIP32Std(childIdx)
 			if err == nil {
 				return kid, nil
 			}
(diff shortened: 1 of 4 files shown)


JoeGruffins (approved):

buck54321 (approved):
Good stuff, @chappjc!

Issue #1327, opened 2021-12-06
client/eth/nodeclient_harness_test: First try with new harness fails.


JoeGruffins:
The first run of tests always seem to fail in a few spots. Probably just needs some more time to sync? Subsequent tests use the same node directory which is probably why this happens on the first run, when the internal node is created.

Some error messages:
```
...
nodeclient_harness_test.go:438: no receipt at index 0 in block 0x0000000000000000000000000000000000000000000000000000000000000000 for tx 0x4abfbfd098973ea86b65c53ba2e89240129548b2b7b3fba7916444ab12789948
...
--- FAIL: TestAccount (2.52s)
    --- PASS: TestAccount/testAccounts (0.00s)
    --- PASS: TestAccount/testBalance (0.00s)
    --- PASS: TestAccount/testUnlock (0.07s)
    --- PASS: TestAccount/testLock (0.00s)
    --- PASS: TestAccount/testListWallets (0.00s)
    --- PASS: TestAccount/testSendTransaction (1.19s)
    --- FAIL: TestAccount/testTransactionReceipt (1.1
```
```
nodeclient_harness_test.go:632: 1 swap ok: receipt error: no receipt at index 0 in block 0x0000000000000000000000000000000000000000000000000000000000000000 for tx 0xe3eff8eb097ef2ab570630a48011e3e5ae063f223019efa1309067321eeb4803
...
--- FAIL: TestContract (127.41s)
    --- PASS: TestContract/testSwap (0.00s)
    --- FAIL: TestContract/testInitiate (1.17s)
    --- PASS: TestContract/testRedeem (63.64s)
    --- PASS: TestContract/testRefund (62.59s)
```
This last one is another testInitiate failure, but is kind of different than the others and may be something else.
```
nodeclient_harness_test.go:656: 1 swap ok: unexpected balance change: want 10999999781551 got 10999999646531, diff = 135019
```

JoeGruffins:
I'll look into this then.

chappjc:
Thanks! Much appreciated

JoeGruffins:
oh wtf
```
[joe@hyrule harness-ctl]$ ./gamma attach --exec 'eth.getTransaction("0x0dec40b2d89c9f64d456bfe26d64851788ee1a2ef9ccdb1f032a3e0f97a8ce79")'
null
[...]

Issue #1268, opened 2021-11-03
neutrino test data race


chappjc:
```
==================
WARNING: DATA RACE
Read at 0x00c000290088 by goroutine 55:
  decred.org/dcrdex/client/asset/btc.(*tNeutrinoClient).BestBlock()
      /home/runner/work/dcrdex/dcrdex/client/asset/btc/spv_test.go:293 +0x5d
  decred.org/dcrdex/client/asset/btc.(*spvWallet).getBlockHeader()
      /home/runner/work/dcrdex/dcrdex/client/asset/btc/spv.go:868 +0x12e
  decred.org/dcrdex/client/asset/btc.(*ExchangeWallet).reportNewTip()
      /home/runner/work/dcrdex/dcrdex/client/asset/btc/btc.go:2539 +0x46d
  decred.org/dcrdex/client/asset/btc.(*ExchangeWallet).watchBlocks()
      /home/runner/work/dcrdex/dcrdex/client/asset/btc/btc.go:2484 +0x2bb
  decred.org/dcrdex/client/asset/btc.tNewWallet.func2()
      /home/runner/work/dcrdex/dcrdex/client/asset/btc/btc_test.go:639 +0x89

Previous write at 0x00c000290088 by goroutine 56:
  decred.org/dcrdex/client/asset/btc.testSyncStatus()
      /home/runner/work/dcrdex/dcrdex/client/asset/btc/btc_test.go:2597 +0x42e
  decred.org/dcrdex/client/asset/btc.runRubric.func3()
      /home/runner/work/dcrdex/dcrdex/client/asset/btc/btc_test.go:682 +0x61
  testing.tRunner()
      /opt/hostedtoolcache/go/1.16.9/x64/src/testing/testing.go:1193 +0x202

Goroutine 55 (running) created at:
  decred.org/dcrdex/client/asset/btc.tNewWallet()
      /home/runner/work/dcrdex/dcrdex/client/asset/btc/btc_test.go:637 +0x53b
  decred.org/dcrdex/client/asset/btc.testSyncStatus()
      /home/runner/work/dcrdex/dcrdex/client/asset/btc/btc_test.go:2575 +0x6e
  decred.org/dcrdex/client/asset/btc.runRubric.func3()
      /home/runner/work/dcrdex/dcrdex/client/asset/btc/btc_test.go:682 +0x61
  testing.tRunner()
      /opt/hostedtoolcache/go/1.16.9/x64/src/testing/testing.go:1193 +0x202

[...]

PR #1346, opened 2021-12-13
client/eth: Fix test gas calculations.

### LINKED ISSUE #1342: eth/client/simnet_tests: testSendSignedTransaction failing.
Fails with `nodeclient_harness_test.go:446: pending in not showing`.

JoeGruffins:
Use the base gas fee for the block before a tx was mined to caculate
    exact gas used and test using that expected amount.

closes #1342

### DIFF
--- a/client/asset/eth/nodeclient.go
+++ b/client/asset/eth/nodeclient.go
@@ -226,6 +226,8 @@ func (n *nodeClient) balance(ctx context.Context) (*Balance, error) {
 
 	for _, tx := range pendingTxs {
 		from, _ := ethSigner.Sender(tx) // zero Address on error
+		// Intentionally ignoring receives from addresses that do not
+		// belong to us.
 		if from != n.creds.addr {
 			continue
 		}
@@ -285,9 +287,12 @@ func (n *nodeClient) sendToAddr(ctx context.Context, addr common.Address, amt ui
 
 // transactionReceipt retrieves the transaction's receipt.
 func (n *nodeClient) transactionReceipt(ctx context.Context, txHash common.Hash) (*types.Receipt, error) {
-	_, blockHash, _, index, err := n.leth.ApiBackend.GetTransaction(ctx, txHash)
+	tx, blockHash, _, index, err := n.leth.ApiBackend.GetTransaction(ctx, txHash)
 	if err != nil {
-		return nil, nil
+		return nil, err
+	}
+	if tx == nil {
+		return nil, fmt.Errorf("transaction %v not found", txHash)
 	}
 	receipts, err := n.leth.ApiBackend.GetReceipts(ctx, blockHash)
 	if err != nil {
@@ -540,7 +545,7 @@ func (n *nodeClient) transactionConfirmations(ctx context.Context, txHash common
 	// tx pool, and when our peers are ready to supply the info. I saw a
 	// CoinNotFoundError in TestAccount/testSendTransaction, but haven't
 	// reproduced.
-	return 0, asset.CoinNotFoundError
+	return 0, fmt.Errorf("cannot find %v: %w", txHash, asset.CoinNotFoundError)
 }
 
 // sendSignedTransaction injects a signed transaction into the pending pool for execution.
--- a/client/asset/eth/nodeclient.go
+++ b/client/asset/eth/nodeclient.go
@@ -543,7 +543,8 @@ func (n *nodeClient) transactionConfirmations(ctx context.Context, txHash common
 	// tx pool, and when our peers are ready to supply the info. I saw a
 	// CoinNotFoundError in TestAccount/testSendTransaction, but haven't
 	// reproduced.
-	return 0, fmt.Errorf("cannot find %v: %w", txHash, asset.CoinNotFoundError)
+	n.log.Warnf("transactionConfirmations: cannot find %v", txHash)
+	return 0, asset.CoinNotFoundError
 }
 
 // sendSignedTransaction injects a signed transaction into the pending pool for execution.
--- a/client/asset/eth/nodeclient.go
+++ b/client/asset/eth/nodeclient.go
@@ -226,8 +226,6 @@ func (n *nodeClient) balance(ctx context.Context) (*Balance, error) {
 
 	for _, tx := range pendingTxs {
 		from, _ := ethSigner.Sender(tx) // zero Address on error
-		// Intentionally ignoring receives from addresses that do not
-		// belong to us.
 		if from != n.creds.addr {
 			continue
 		}
--- a/client/asset/eth/nodeclient_harness_test.go
+++ b/client/asset/eth/nodeclient_harness_test.go
@@ -49,10 +49,12 @@ import (
 	"github.com/davecgh/go-spew/spew"
 	"github.com/ethereum/go-ethereum/accounts"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/consensus/misc"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/crypto/secp256k1"
 	"github.com/ethereum/go-ethereum/node"
+	"github.com/ethereum/go-ethereum/rpc"
 )
 
 const (
@@ -270,7 +272,6 @@ func syncClient(cl *nodeClient) error {
 }
 
 func TestBasicRetrieval(t *testing.T) {
-	syncClient(ethClient)
 	t.Run("testBestBlockHash", testBestBlockHash)
 	t.Run("testBestHeader", testBestHeader)
 	t.Run("testBlock", testBlock)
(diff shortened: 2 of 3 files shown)


JoeGruffins:
I believe this does fix some random errors, but not any called out in the issues...

client/asset/eth/nodeclient.go
```diff
@@ -268,9 +270,12 @@ func (n *nodeClient) sendToAddr(ctx context.Context, addr common.Address, amt ui
...
 func (n *nodeClient) transactionReceipt(ctx context.Context, txHash common.Hash) (*types.Receipt, error) {
-	_, blockHash, _, index, err := n.leth.ApiBackend.GetTransaction(ctx, txHash)
+	tx, blockHash, _, index, err := n.leth.ApiBackend.GetTransaction(ctx, txHash)
 	if err != nil {
-		return nil, nil
+		return nil, err
+	}
+	if tx == nil {
+		return nil, fmt.Errorf("transaction %v not found", txHash)
 	}
```
JoeGruffins, line 279: Some convo in #1327

client/asset/eth/nodeclient.go
```diff
@@ -513,7 +518,7 @@ func (n *nodeClient) transactionConfirmations(ctx context.Context, txHash common
 	// tx pool, and when our peers are ready to supply the info. I saw a
 	// CoinNotFoundError in TestAccount/testSendTransaction, but haven't
 	// reproduced.
-	return 0, asset.CoinNotFoundError
+	return 0, fmt.Errorf("cannot find %v: %w", txHash, asset.CoinNotFoundError)
```
JoeGruffins, line 521: I see this randomly hit sometimes, but other nodes do see the tx on the other nodes. I don't know why it's being missed.

client/asset/eth/nodeclient.go
```diff
@@ -209,6 +209,8 @@ func (n *nodeClient) balance(ctx context.Context) (*Balance, error) {
 
 	for _, tx := range pendingTxs {
 		from, _ := ethSigner.Sender(tx) // zero Address on error
+		// Intentionally ignoring receives from addresses that do not
```
martonp, line 212: I don't think those would even be listed in pending transactions... I think this is more of a sanity check.

martonp (approved):
Looks good, tests passed on a fresh harness.

client/asset/eth/nodeclient.go
JoeGruffins: Is "those" receives from addresses that do not belong to the wallet? I think this will show all txs in the mempool?

buck54321: From **go-ethereum/light/txpool.go**

```Go
// TxPool implements the transaction pool for light clients, which keeps track
// of the status of locally created transactions, ...
```

More accurately, `ignoring transactions that are not from us`, but that's really what the code says, so I don't know that the doc adds much.

JoeGruffins: Removing the doc.

JoeGruffins:
Oh man, after rebase tests are failing in two spots consistently:
```
nodeclient_harness_test.go:966: test "ok before locktime": expected isRedeemable to be true, but got false
```
and
```
nodeclient_harness_test.go:1184: wrong refunder: refund error: insufficient funds for gas * price + value
```
What happened. anyone else? I updated to geth 1.10.13

martonp:
Looks like due to the key derivation change in https://github.com/decred/dcrdex/commit/c54f8d135aef57ff0fc1d9c576d96195a802cb4b , the `participantPrivKey` is being derived differently, and the `participantAddr` no longer has any funds.

chappjc:
Ah dang, my bad.  Didn't anticipate the hard coded eth keys in the harness.

JoeGruffins:
Ok, just need to change the keys then. Will do. Thx @martonp

client/asset/eth/nodeclient_harness_test.go
```diff
@@ -682,8 +689,12 @@ func testInitiate(t *testing.T) {
 			t.Fatalf("%s: balance error: %v", test.name, err)
 		}
 
+		gasPrice, err := feesAtBlk(ctx, ethClient, receipt.BlockNumber.Int64()-1)
+		if err != nil {
+			t.Fatalf("%s: feesAtBlk error: %v", test.name, err)
+		}
 		bigGasUsed := new(big.Int).SetUint64(receipt.GasUsed)
-		txFee := new(big.Int).Mul(bigGasUsed, baseFee) // DRAFT NOTE: Why don't I need to add the tip?
+		txFee := new(big.Int).Mul(bigGasUsed, gasPrice)
```
buck54321, line 697: I think @martonp had explained this to me, but I can't find it now. Everything I read seems to indicate that tx fees should be assessed at `(baseFeeRate + priorityFeeRate) * gas`, but these calculations seem to have no priority fee added.

client/asset/eth/nodeclient_harness_test.go
```diff
@@ -682,8 +689,12 @@ func testInitiate(t *testing.T) {
 			t.Fatalf("%s: balance error: %v", test.name, err)
 		}
 
+		gasPrice, err := feesAtBlk(ctx, ethClient, receipt.BlockNumber.Int64()-1)
```
buck54321, line 692: Does this mean that the transaction was not being included in the block we expected?

buck54321 (changes requested):

client/asset/eth/nodeclient_harness_test.go
chappjc: Ah, the `feesAtBlk` helper added in this PR for computing this `gasPrice` value is doing that.

BTW, the computation there is part of my motivation for https://github.com/decred/dcrdex/pull/1356 (so the needed ChainConfig can be loaded by server) and the linked [possible followup change](https://github.com/decred/dcrdex/commit/64a72b48fe3b05fcf8d5f4a7d9296496092967ac#r62517085) that builds on that PR to do the same calculation for `server/asset.(*Backend).FeeRate`, even though it's probably overkill for that need.

And I apologize in advance for the extremely long-winded issue https://github.com/decred/dcrdex/issues/1372.  I'm actually trying to get consensus about these eth fee issues there, with a focus on the prescribed match rates that get recorded and enforce.

buck54321: > the `feesAtBlk` helper added in this PR for computing this `gasPrice` value is doing that

Ah. Derp.

client/asset/eth/nodeclient_harness_test.go
JoeGruffins: I guess it just means that the base fee for the a block is encoded in (or calculated from) the previous block.

client/asset/eth/nodeclient_harness_test.go
```diff
@@ -441,11 +438,14 @@ func testSendSignedTransaction(t *testing.T) {
...
 	bal, _ := ethClient.balance(ctx)
-	if bal.PendingIn.Cmp(dexeth.GweiToWei(1)) != 0 { // We sent it to ourselves.
-		t.Fatalf("pending in not showing")
-	}
 
-	if bal.PendingOut.Cmp(dexeth.GweiToWei(1)) != 0 {
+	// Removed pending issue #1342 resolution.
+	//if bal.PendingIn.Cmp(dexeth.GweiToWei(1)) != 0 { // We sent it to ourselves.
+	//	t.Fatalf("pending in not showing")
+	//}
```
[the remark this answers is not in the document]
JoeGruffins: Will do

dex/testing/eth/harness.sh
```diff
@@ -87,6 +87,7 @@ cat > "${NODES_ROOT}/genesis.json" <<EOF
     "muirGlacierBlock": 0,
     "berlinBlock": 0,
     "londonBlock": 0,
+    "arrowGlacierBlock": 0,
```
JoeGruffins, line 90: Sneaking this in.

JoeGruffins:
Is there a issue for this error?

https://github.com/decred/dcrdex/runs/4654985696?check_suite_focus=true

chappjc:
> Is there a issue for this error?
> 
> https://github.com/decred/dcrdex/runs/4654985696?check_suite_focus=true

looks new

client/asset/eth/nodeclient_harness_test.go
JoeGruffins: Oh, it's in the header, the latest commit makes more sense for getting the base fee from mined blocks.

JoeGruffins:
> looks new
Is it #1268 ? Looks very similar.

client/asset/eth/nodeclient_harness_test.go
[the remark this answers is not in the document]
JoeGruffins: Removed.

JoeGruffins:
Just rebased.

buck54321 (approved):

chappjc:
Yeah, sorry @JoeGruffins.  This PR is g2g, but I merged @martonp's token nodeclient PR first.

PR #1338, opened 2021-12-08
[release-v0.2] maintenance backports

chappjc:
Since both the v0.3 and v0.4 stable branches of dcrdex are targeting the unreleased Decred 1.7 (dcrd/dcrwallet), this cherry-picks a handful of minor bug fixes for the 0.2 release branch.

- pg: remove old pseudo-upgrade (https://github.com/decred/dcrdex/pull/1226)
- client/webserver: do not check empty CustomSiteDir (https://github.com/decred/dcrdex/pull/1253)
- Update go:build tags and log more init data (https://github.com/decred/dcrdex/pull/1161)
- fix conns data races (https://github.com/decred/dcrdex/pull/1280)
- server/market: fix wrong asset for RequiredOrderFunds on market-buy (https://github.com/decred/dcrdex/pull/1323)

This also adds two new commits to facilitate this backporting:
- dcr harness: allow specifying binary paths (e.g. `DCRD=dcrd-1.6 DCRWALLET=dcrwallet-1.6 ./harness.sh`)
- bump versions to v0.2.3 in the applications and in the packaging script

PR #1351, opened 2021-12-14
server: validate taker's contract hash in audit

chappjc:
The server's audit of the taker's contract must also verify that the contract hash used in their contract script matches the secret hash from the maker's contract.  This check is already performed client-side, this just updates the server's auditing to do it as well.

This was included in a recent v0.2 backport PR: https://github.com/decred/dcrdex/pull/1338


chappjc:
Although this is the approach taken with the v0.2 fix, we can absolutely reconsider this.  But adding the `SecretHash` field seems logical to me, esp. considering it's the same kind of extracted audit data like locktime, swap address, and value, which we keep together in other places as well.

Naturally this is a v0.4 blocker, with the cherry-pick in https://github.com/decred/dcrdex/pull/1352

JoeGruffins (approved):

PR #1355, opened 2021-12-16
client/eth: Fix withdraw TODOs.

JoeGruffins:
Return txid bytes for printing.

    Reduce amount to withdraw by the max possible fee if value cannot
    otherwise be sent.

part of #1154

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -1116,11 +1116,6 @@ func (*ExchangeWallet) ValidateSecret(secret, secretHash []byte) bool {
 	return bytes.Equal(h[:], secretHash)
 }
 
-// Confirmations gets the number of confirmations for the specified coin ID.
-func (*ExchangeWallet) Confirmations(ctx context.Context, id dex.Bytes) (confs uint32, spent bool, err error) {
-	return 0, false, asset.ErrNotImplemented
-}
-
 // SyncStatus is information about the blockchain sync status.
 func (eth *ExchangeWallet) SyncStatus() (bool, float32, error) {
 	// node.SyncProgress will return nil both before syncing has begun and
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -1107,7 +1107,7 @@ func (eth *ExchangeWallet) Withdraw(addr string, value, _ uint64) (asset.Coin, e
 		return nil, err
 	}
 	txHash := tx.Hash()
-	return &coin{id: txHash[:], value: tx.Value().Uint64()}, nil
+	return &coin{id: txHash[:], value: dexeth.WeiToGwei(tx.Value())}, nil
 }
 
 // ValidateSecret checks that the secret satisfies the contract.
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -1081,16 +1081,33 @@ func (eth *ExchangeWallet) SwapConfirmations(ctx context.Context, _ dex.Bytes, c
 	return
 }
 
-// Withdraw withdraws funds to the specified address. Value is gwei.
-//
-// TODO: Return the asset.Coin.
-// TODO: Subtract fees from the value.
+// Withdraw withdraws funds to the specified address. Value is gwei. The fee is
+// subtracted from the total balance if it cannot be sent otherwise.
 func (eth *ExchangeWallet) Withdraw(addr string, value, _ uint64) (asset.Coin, error) {
-	_, err := eth.node.sendToAddr(eth.ctx, common.HexToAddress(addr), value)
+	eth.lockedMtx.Lock()
+	defer eth.lockedMtx.Unlock()
+
+	bal, err := eth.balance()
 	if err != nil {
 		return nil, err
 	}
-	return nil, nil
+	avail := bal.Available
+	if avail < value {
+		return nil, fmt.Errorf("not enough funds to withdraw: have %d gwei need %d gwei", avail, value)
+	}
+	maxFee := defaultSendGasLimit * eth.gasFeeLimit
+	if avail < maxFee {
+		return nil, fmt.Errorf("not enough funds to withdraw: cannot cover max fee of %d gwei", maxFee)
+	}
+	if avail < value+maxFee {
+		value -= maxFee
+	}
+	tx, err := eth.node.sendToAddr(eth.ctx, common.HexToAddress(addr), value)
+	if err != nil {
+		return nil, err
+	}
+	txHash := tx.Hash()
+	return &coin{id: txHash[:], value: tx.Value().Uint64()}, nil
 }
 
 // ValidateSecret checks that the secret satisfies the contract.
--- a/client/asset/eth/eth_test.go
+++ b/client/asset/eth/eth_test.go
@@ -2424,3 +2424,65 @@ func TestPayFee(t *testing.T) {
 		}
 	}
 }
+
+func TestWithdraw(t *testing.T) {
+	tx := tTx(0, 0, 0, &testAddressA, nil)
+	txHash := tx.Hash()
+	maxFee := uint64(defaultSendGasLimit * defaultGasFeeLimit)
+	tests := []struct {
+		name                  string
+		value                 uint64
+		bal                   *Balance
+		balErr, sendToAddrErr error
+		wantErr               bool
+	}{{
+		name:  "ok",
+		value: 10e9,
+		bal:   newBalance(10e9, 0, 0),
+	}, {
+		name:    "balance error",
+		balErr:  errors.New(""),
+		wantErr: true,
+	}, {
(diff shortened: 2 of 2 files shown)


client/asset/eth/eth.go
```diff
@@ -1060,16 +1060,33 @@ func (eth *ExchangeWallet) SwapConfirmations(ctx context.Context, _ dex.Bytes, c
...
-// Withdraw withdraws funds to the specified address. Value is gwei.
-//
-// TODO: Return the asset.Coin.
-// TODO: Subtract fees from the value.
+// Withdraw withdraws funds to the specified address. Value is gwei. The fee is
+// subtracted from the total balance if it cannot be sent otherwise.
 func (eth *ExchangeWallet) Withdraw(addr string, value, _ uint64) (asset.Coin, error) {
-	_, err := eth.node.sendToAddr(eth.ctx, common.HexToAddress(addr), value)
+	eth.lockedFundsMtx.Lock()
+	defer eth.lockedFundsMtx.Unlock()
```
chappjc, line 1067: Can these be RLock/RUnlock?

client/asset/eth/eth.go
```diff
@@ -1060,16 +1060,33 @@ func (eth *ExchangeWallet) SwapConfirmations(ctx context.Context, _ dex.Bytes, c
...
+	if avail < value {
+		return nil, fmt.Errorf("not enough funds to withdraw: have %d gwei need %d gwei", avail, value)
+	}
+	maxFee := defaultSendGasLimit * eth.gasFeeLimit
+	if avail < maxFee {
+		return nil, fmt.Errorf("not enough funds to withdraw: cannot cover max fee of %d gwei", maxFee)
+	}
+	if avail < value+maxFee {
+		value -= maxFee
+	}
```
chappjc, line 1083: You've documented this, but this gives `Withdraw` different behavior depending on if it looks like the user is trying to sweep their wallet or if they are trying to send a specific smaller amount.  Maybe we have no choice with EIP-1559?

We had generally agreed that it's best to have `Withdraw` consistently behave like the other implementations whereby `value` is subtracted from the wallet, not `value` sent to the destination `addr`.  Always subtracting a fee estimate from the value sent to the destination address (like on this line) would do that, at least with legacy transaction.  EIP-1559 transaction are gonna screw that up of course because the fee isn't decided until it is mined.

Even with this current approach of conditionally subtracting `maxFee`, I don't understand what happens if the resulting fee with EIP-1559 turns out to be lower.  The wallet is then left with dust?  How can we encourage the miner to take the dust in this case?  Increase the tip?

chappjc: I'm thinking that what you presently have is best from a UX perspective.  That is, have Withdraw use "SendToAddress" semantics with no fee subtraction unless it look like a sweep attempt (like you have).

If we see dust remaining in the wallet, we can look at either tweaking tip or using a legacy txn to sweep.

To address the Withdraw -> Send meaning change, we can look at wallet config opts or wallet definition stuff to communicate to the frontend how to display things.

chappjc: Well anyway, the main goal of this PR was to add the `asset.Coin` return, and it does that.

The second TODO did indicate that we should _universally_ subtract fees from the value though, the goal being to reduce balance by `value`.  Although I don't know  how to achieve that with EIP-1559, hence my suggestions above about changes for frontend to clarify the actual behavior and/or switches to enable potentially unknown amounts (less) sent to the destination address.

Curious what @martonp and @buck54321 think about this.

My feeling is we should not spend much time on this with everything else we want to get done, as long as we at least get the `asset.Coin` returned.

martonp: If we do a legacy transaction then the miner would take the dust, but I don't really see the point. I think we might as well leave the dust. Maybe they can use it in the future.

The one thing I would change is instead of using `eth.gasFeeLimit`, we should use `min(2 * BaseFee, eth.gasFeeLimit)` if `avail < value+maxFee`.

JoeGruffins: Removing the fee at all times and communicating that value better would be good I guess.

Normally wallets do not take the fee from the value, correct? When I send coins with dcr or btc, I expect that full amount to be set. And I suppose just an error if the fee cannot be covered.

chappjc: > If we do a legacy transaction then the miner would take the dust, but I don't really see the point. I think we might as well leave the dust. Maybe they can use it in the future.

Leaving dust than cannot be spent is a worse UX than sending a tiny bit extra to the miners IMO.

> Normally wallets do not take the fee from the value, correct

Both our btc and dcr implementations of this **do** take the fee from value.  By calling this Withdraw instead of Send, we are committing to a particular meaning.  Hence my suggestions about frontend changes to allow a deviation.

JoeGruffins: For now, using a legacy tx and always subtracting from the value sounds like the fastest solution.

chappjc: Well, fastest is to leave what you have and accepting that Withdraw does something different than what the name implies.  I honestly don't care right now.  It's probably a bigger discussion that involves frontend display changes and possibly options.

client/asset/eth/eth.go
JoeGruffins: The act of sending changes what is returned by `balance` so I think holding the write lock is optimal as it changes the amount locked.

client/asset/eth/eth.go
JoeGruffins: > If we do a legacy transaction then the miner would take the dust

Actually trying this out doesn't seem to be the case in the harness. Probably the miner would take as much as they can though, right? We choose the max, so they don't have to take it all I guess.

um...

JoeGruffins: image

😬

JoeGruffins:
I realized while doing this that pending out funds are not reflected in the balance on startup.

JoeGruffins:
This can't be 1/256 odds.

client/asset/eth/eth.go
chappjc: Sure.  100% guarantee we'll end up reverting the UI change, but we are not intending this commit for release. But with this PR I'm happy it's now returning a coin, and that it's at least possible to attempt to sweep the account.

chappjc (approved):
This PR updates `Withdraw` to returning a `Coin`, and makes it possible to attempt to sweep the account.
We have to work through making it consistently withdraw or send, rather than a mixed behavior, and I'm thinking more UI updates will be needed to do this right with probably checkboxes and tooltips for the ETH Withdraw dialog.

https://github.com/decred/dcrdex/pull/1355#discussion_r770762025

client/asset/eth/eth.go
```diff
@@ -1047,16 +1047,33 @@ func (eth *ExchangeWallet) SwapConfirmations(ctx context.Context, _ dex.Bytes, c
...
+	tx, err := eth.node.sendToAddr(eth.ctx, common.HexToAddress(addr), value)
+	if err != nil {
+		return nil, err
+	}
+	txHash := tx.Hash()
+	return &coin{id: txHash[:], value: tx.Value().Uint64()}, nil
```
chappjc, line 1076: Isn't `tx.Value()` wei instead of gwei?  If so, same in the PayFee PR.

JoeGruffins: yes

JoeGruffins: Changed.

client/asset/eth/eth.go
buck54321: UI change would be good. @chappjc's idea about a "sweep account" option is the best option at the UI level, imo. Maybe the correct API for the `asset.Wallet` would be to change from 

```Go
Withdraw(address string, value, feeSuggestion uint64) (Coin, error)
```

to

```Go
Send(address string, value, feeSuggestion uint64) (Coin, error)
Sweep(address string, feeSuggestion uint64) (Coin, error)
```
.

chappjc: I started experimenting with the notion of a `WalletTrait` added to `asset.WalletDefinition` to help convey variations in wallet functionality to wallet consumers like frontend.
I used `RescanWallet` as a test in https://github.com/decred/dcrdex/pull/1374.  The `WalletTrait`s and UI work is in the third commit there.

chappjc: OK, so @JoeGruffins, sounds like we can merge this if we drop https://github.com/decred/dcrdex/pull/1355/commits/88da8ed9a9ea2184e4e9485ca01fc2bc60a95337 and aim solve this with the WalletTraits proposed in #1374 .  When that PR gets ironed out, we can do as @buck54321 suggested and implement multiple optional methods, if you'd like to pick that up at that time.

JoeGruffins: Removed the second commit.

PR #1358, opened 2021-12-17
client/asset/eth: Add RecoverPubkey for no-cgo.

chappjc:
Add the `RecoverPubkey` function, which is just like geth's `crypto/secp256k1.RecoverPubKey`, except it does not use the cgo
secp256k1 library, instead using dcrd's `ecdsa.RecoverCompact` after transforming the signature from geth's compact format to
dcrd/btcd's compact format.

With geth signatures, it is always expected to return an uncompressed public key.

Build with `CGO_ENABLED=0` to ensure we aren't accidentally relying on cgo packages of functions.  See https://github.com/ethereum/go-ethereum/pull/19121


buck54321 (approved):

JoeGruffins (approved):

client/asset/eth/eth.go
```diff
@@ -752,6 +752,24 @@ func (eth *ExchangeWallet) Redeem(form *asset.RedeemForm) ([]dex.Bytes, asset.Co
 	return inputs, outputCoin, fundsRequired, nil
 }
 
+// RecoverPubkey recovers the uncompressed public key from the signature and the
+// message hash that wsa signed. The signature should be a compact signature
+// generated by a geth wallet, with the format [R || S || V], with the recover
+// bit V at the end. See go-ethereum/crypto.Sign.
+func RecoverPubkey(msgHash, sig []byte) ([]byte, error) {
```
JoeGruffins, line 759: It just hit me that this is exported. If will use it somewhere else, maybe goes in dex/networks/eth?

chappjc: Good point.  I suspected we'd want to use it server-side, but perhaps not.  Will just unexport for now.

For any VSCode users, this can go in settings.json to ensure the correct code is tested:

```
    "go.testEnvVars": {
        "CGO_ENABLED": 0,
    },
```

chappjc:
This appears to be needed for both https://github.com/decred/dcrdex/pull/1363 and https://github.com/decred/dcrdex/pull/1314, so merging ahead of those now.

PR #1360, opened 2021-12-18
dex/networks/erc20: ERC20Swap V0

### LINKED ISSUE #1345: Implement swap contract for erc20

martonp:
- Introduce the initial version of the ERC20Swap contract
- Add a script to compile the contract and generate the
go code needed to work with it
- Deploy the ERC20Swap contract and a test ERC20 token to
the ETH simnet using the ETH testing harness

Closes #1345


dex/networks/erc20/contracts/ERC20SwapV0.sol
```diff
@@ -0,0 +1,209 @@
+// SPDX-License-Identifier: BlueOak-1.0.0
+// pragma should be as specific as possible to allow easier validation.
+pragma solidity = 0.8.6;
+
+interface ERC20 {
+    function transfer(address recipient, uint256 amount) external returns (bool);
+    function transferFrom(address sender, address recipient, uint256 amount) external returns (bool);
```
chappjc, line 7: USDT's ERC20 is not quite an ERC20 since it doesn't have the returns from these functions.  https://etherscan.io/address/0xdac17f958d2ee523a2206206994597c13d831ec7#code  Will that he a challenge to accomodate with this `ERC20Swap` contract?

dex/networks/erc20/contracts/ERC20SwapV0.sol
```diff
@@ -0,0 +1,209 @@
...
+}
+
+
+contract ERC20Swap {
+    bytes4 private constant TRANSFER_FROM_SELECTOR = bytes4(keccak256("transferFrom(address,address,uint256)"));
+	bytes4 private constant TRANSFER_SELECTOR = bytes4(keccak256("transfer(address,uint256)"));
```
chappjc, line 13: I think these can be removed.  Leftovers from that stx template?

chappjc (commented):
Nice work, @martonp!

dex/networks/erc20/contracts/ERC20SwapV0.sol
chappjc: May be a good reference for other oddities: https://github.com/d-xo/weird-erc20

dex/networks/erc20/contracts/ERC20SwapV0.sol
martonp: Yes this needs to be removed.

dex/networks/erc20/contracts/ERC20SwapV0.sol
martonp: I don't think that will have an effect since the return type of a function is not part of its signature, but I'll test it out.

martonp: With my previous implementation, `transfer` and `transferFrom` actually don't work when the functions don't return anything. The current implementation fixes it.

I'm not sure if there's a general solution to checking a blocklist.. that's something we could only implement individually for each token.

dex/networks/erc20/contracts/ERC20SwapV0.sol
```diff
@@ -0,0 +1,207 @@
...
+
+            swapToUpdate.initBlockNumber = block.number;
+            swapToUpdate.refundBlockTimestamp = initiation.refundTimestamp;
+            swapToUpdate.initiator = msg.sender;
+            swapToUpdate.participant = initiation.participant;
+            swapToUpdate.value = initiation.value;
+            swapToUpdate.state = State.Filled;
+            swapToUpdate.token = token;
+
+            initVal += initiation.value;
```
JoeGruffins, line 97: If the user does not have enough of the token to satisfy `initVal`, will that error at `token.call`?

dex/networks/erc20/contracts/ERC20SwapV0.sol
```diff
@@ -0,0 +1,207 @@
...
+    // 256-bit slots as possible to reduce gas cost. In particular, the 160-bit
+    // address can pack with the 8-bit State.
+    struct Swap {
+        bytes32 secret;
+        uint256 value;
+        uint initBlockNumber;
+        uint refundBlockTimestamp;
+        address initiator;
+        address participant;
+        address token;
```
JoeGruffins, line 24: Is `token` in actuality the address of a contract?

dex/networks/erc20/contracts/ERC20SwapV0.sol
```diff
@@ -0,0 +1,207 @@
...
+    // isRedeemable returns whether or not a swap identified by secretHash
+    // can be redeemed using secret.
+    function isRedeemable(bytes32 secretHash, bytes32 secret)
+        public
+        view
+        returns (bool)
+    {
+        Swap storage swapToRedeem = swaps[secretHash];
+        return swapToRedeem.state == State.Filled &&
+               swapToRedeem.participant == msg.sender &&
```
JoeGruffins, line 120: Probably late to the party, but the `swapToRedeem.participant == msg.sender` would mean that this can only ever be checked by one wallet, right? I don't know if we wanted this to be observable from web browsers and stuff. It would just return false if not asked from the participant. No need to change here though, if it was not intended, because it's also like this in the eth version.

dex/testing/eth/harness.sh
```diff
@@ -296,10 +304,22 @@ do
   "${NODES_ROOT}/harness-ctl/mine-beta" "5"
 done
 
-CONTRACT_ADDR=$("${NODES_ROOT}/harness-ctl/alpha" "attach --preload ${NODES_ROOT}/harness-ctl/contractAddress.js --exec contractAddress(\"${CONTRACT_HASH}\")" | sed 's/"//g')
-echo "Contract address is ${CONTRACT_ADDR}. Saving to ${NODES_ROOT}/contract_hash.txt"
-cat > "${NODES_ROOT}/contract_addr.txt" <<EOF
-${CONTRACT_ADDR}
+ETH_SWAP_CONTRACT_ADDR=$("${NODES_ROOT}/harness-ctl/alpha" "attach --preload ${NODES_ROOT}/harness-ctl/contractAddress.js --exec contractAddress(\"${ETH_SWAP_CONTRACT_HASH}\")" | sed 's/"//g')
+echo "ETH SWAP contract address is ${ETH_SWAP_CONTRACT_ADDR}. Saving to ${NODES_ROOT}/eth_swap_contract_address.txt"
```
JoeGruffins, line 308: There are two tests still using the old file location, could you change them?
```
grep -rn contract_addr.txt
client/asset/eth/nodeclient_harness_test.go:64: contractAddrFile      = filepath.Join(homeDir, "dextest", "eth", "contract_addr.txt")
server/asset/eth/rpcclient_harness_test.go:29:  contractAddrFile = filepath.Join(homeDir, "dextest", "eth", "contract_addr.txt")
```

JoeGruffins (approved):
Makes sense to me and really doesn't deviate from the eth contract much at all. LGTM imo.

dex/networks/erc20/contracts/ERC20SwapV0.sol
JoeGruffins: As `token` in the contract would correspond to a known contract, we could make a list/map of blacklists later and apply if the address matches.

dex/networks/erc20/ERC20.go
```diff
@@ -0,0 +1,20 @@
+package erc20
```
JoeGruffins, line 1: Should this have the
```
// This code is available on the terms of the project LICENSE.md file,
// also available online at https://blueoakcouncil.org/license/1.0.0.

//go:build lgpl
// +build lgpl
```
At the top?

dex/networks/erc20/contracts/ERC20SwapV0.sol
```diff
@@ -0,0 +1,207 @@
...
+            require(sha256(abi.encodePacked(redemption.secret)) == redemption.secretHash);
+
+            swapToRedeem.state = State.Redeemed;
+            swapToRedeem.secret = redemption.secret;
+            amountToRedeem += swapToRedeem.value;
+        }
+		bool success;
+		bytes memory data;
+        (success, data) = token.call(abi.encodeWithSelector(TRANSFER_SELECTOR, msg.sender, amountToRedeem));
+        require(success && (data.length == 0 || abi.decode(data, (bool))), 'transfer failed');
```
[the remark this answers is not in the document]
martonp: You cannot see it from the transaction receipt, but you can see it if you call this function:
https://github.com/ethereum/go-ethereum/blob/e4f570fcc67aa2df489666cc08568c637024fd3c/ethclient/ethclient.go#L462

Yes, I'll add messages to the rest of them.

dex/networks/erc20/contracts/ERC20SwapV0.sol
```diff
@@ -0,0 +1,207 @@
...
+    // It is important to note that this calls an external contract which has no
+    // restrictions on gas used. This has the potential to open the contract up
+    // to a reentry attack.
+    //
+    // This is a writing function and requires gas. Failure or success should
+    // be guaged by querying the swap and checking state after being mined. Gas
+    // is expended either way.
+    function initiate(Initiation[] calldata initiations, address token)
+        public
+        senderIsOrigin()
```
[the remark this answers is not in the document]
martonp: Yes that's all correct. One big UX difference we will have to deal with is that for the swap contract to be able to spend the user's token's the user has to send an `approve` transaction to the token contract first which allows the swap contract to spend their tokens.

dex/networks/erc20/contracts/ERC20SwapV0.sol
martonp: Yes.. assuming the token contract doesn't give the user free tokens :). Also, it will fail if the user has not approved at least `initVal` tokens to be spent by the swap contract.

dex/networks/erc20/contracts/ERC20SwapV0.sol
martonp: Yes.

dex/networks/erc20/contracts/ERC20SwapV0.sol
martonp: No, any address can be used. When calling a `view` function a signature is required, so any address can be put into the `From` field.

chappjc: You mean a signature is **not** required?  No transaction or gas either, just an `accounts/abi/bind.(*BoundContract).Call` on the contract method with `&bind.CallOpts{From: whatever}`.

It's conceivable that this function could be written like `function isRedeemable(bytes32 secretHash, bytes32 secret, acct address) public view returns (bool)` so you could do it on say etherscan from the `#readContract` part of the page (e.g. https://etherscan.io/address/0xdac17f958d2ee523a2206206994597c13d831ec7#readContract).  However, since you would **never** want to execute this function outside of your own machine, except when actually redeeming, because of the `secret` argument, it need not be like that.  Also, we have the `swap(bytes32 secretHash) public view returns(Swap memory)` that presumably would be usable for free browser-based queries.

dex/networks/erc20/contracts/ERC20SwapV0.sol
JoeGruffins: Would relying on the `token.call` to fail on blacklisted addresses work? I would think a blacklisted address is prevented from sending or receiving funds?

dex/networks/erc20/contracts/ERC20SwapV0.sol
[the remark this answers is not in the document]
JoeGruffins: Very interesting. I did not know this.

dex/networks/erc20/contracts/ERC20SwapV0.sol
JoeGruffins: > When calling a view function a signature is required

Not required? Interesting.

> since you would never want to execute this function outside of your own machine

oh yeah, derp, wouldnt want to expose the secret

What was it then we wanted to be able to do from explorers, or am I making stuff up. Just view swap state?

JoeGruffins:
Thanks for the additions, looks good to me.

chappjc:
Looks good to me too, although note that with https://github.com/decred/dcrdex/pull/1362 merged the ethswapv0 contract bytes have changed and harness.sh has a conflict

dex/networks/erc20/contracts/ERC20SwapV0.sol
```diff
@@ -0,0 +1,233 @@
...
+            swapToUpdate.participant = initiation.participant;
+            swapToUpdate.value = initiation.value;
+            swapToUpdate.state = State.Filled;
+            swapToUpdate.token = token;
+
+            initVal += initiation.value;
+        }
+		bool success;
+		bytes memory data;
+        (success, data) = token.call(abi.encodeWithSelector(TRANSFER_FROM_SELECTOR, msg.sender, address(this), initVal));
```
buck54321, line 126: Would the swap contract need to approve itself on the token contract before use?

buck54321 (changes requested):

dex/networks/erc20/contracts/ERC20SwapV0.sol
```diff
@@ -0,0 +1,233 @@
...
+    // initiate initiates an array of swaps all for the same ERC20 token.
+    // It checks that all of the swaps have a non zero redemptionTimestamp
+    // and value, and that none of the secret hashes have ever been used
+    // previously. Once initiated, each swap's state is set to Filled. The
+    // tokens equal to the sum of each swap's value are now in the custody of
+    // the contract and can only be retrieved through redeem or refund.
+    //
+    // It is important to note that this calls an external contract which has no
+    // restrictions on gas used. This has the potential to open the contract up
+    // to a reentry attack.
```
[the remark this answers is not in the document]
chappjc: I think this was referring to the current advice not to use `send` or `transfer`, which have gas limits and as such were previously advised to mitigate reentrancy attacks, and to instead use `call` in combination with other patterns such as check-effects-interaction.  https://consensys.net/diligence/blog/2019/09/stop-using-soliditys-transfer-now/
This was my position at least when we were drafting the ETHSwapV0 with @JoeGruffins

dex/networks/erc20/contracts/ERC20SwapV0.sol
chappjc: Right, @martonp mentioned this in @JoeGruffins' review here: https://github.com/decred/dcrdex/pull/1360#discussion_r773220069
We'll have to have some code in the client wallet setup or connect methods that ensures it has approved this token swap contract to spend the particular token, and if it has not yet, the user account will have to make an `approve` txn on the token for this swap contract.  :(

buck54321: Ah. I get it now. Thanks.

dex/networks/erc20/contracts/ERC20SwapV0.sol
```diff
@@ -0,0 +1,233 @@
...
+    // 256-bit slots as possible to reduce gas cost. In particular, the 160-bit
+    // address can pack with the 8-bit State.
+    struct Swap {
+        bytes32 secret;
+        uint256 value;
+        uint initBlockNumber;
+        uint refundBlockTimestamp;
+        address initiator;
+        address participant;
+        address token;
```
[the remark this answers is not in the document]
martonp: The reason I prefer this way is to reduce the friction if adding new tokens. Is the plan to just support a very limited number of tokens?

dex/networks/erc20/contracts/ERC20SwapV0.sol
```diff
@@ -0,0 +1,233 @@
...
+
+    // constructor is empty. This contract has no connection to the original
+    // sender after deployed. It can only be interacted with by users
+    // initiating, redeeming, and refunding swaps.
+    constructor() {}
+
+    // senderIsOrigin ensures that this contract cannot be used by other
+    // contracts, which reduces possible attack vectors. There is some
+    // conversation in the eth community about removing tx.origin, which would
+    // make this check impossible.
```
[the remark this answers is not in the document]
martonp: I'll just remove it.. I do see some discussion, but seems like a lot of people are opposed to it. Either way, it won't be removed from the version of solidity that we are using.

dex/networks/erc20/contracts/ERC20SwapV0.sol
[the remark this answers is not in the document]
chappjc: The game is USDT and USDC, by trading volume with current market pairs, but supporting lots of ERC20s is also desirable.  We could make dedicated contracts for the two big ones, if it saves notable gas per swap.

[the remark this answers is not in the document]
buck54321: Presented with a choice of a one-time fee vs higher fees forever, the one-time fee seems favorable. It also makes the API identical to the eth swap contract, which has some potential benefits for us.

We'll need to hard-code asset IDs for each token anyway. I don't think it's crazy to add contract addresses too.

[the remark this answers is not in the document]
martonp: The friction isn't hard coding the address, it's having to deploy a new contract every time. I'll test out how much the gas savings are.. I guess it's around 20k extra to initiate, and just a minor difference in redemption.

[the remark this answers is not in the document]
martonp: 20k gas is ~$8 currently.. that's kind of worth it.

dex/networks/erc20/contracts/ERC20SwapV0.sol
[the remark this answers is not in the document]
martonp: I've removed the comments. I also removed some additional ones which are basically just documenting how solidity works.. does anyone mind I removed those?

dex/networks/erc20/contracts/ERC20SwapV0.sol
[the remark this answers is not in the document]
chappjc: Hmm, I suppose dedicated contracts for each token would allow going back to the interface instead of abi selectors.  Did that have any impact on gas?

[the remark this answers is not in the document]
martonp: That's true.. I'm not sure but I don't think that will be a substantial change.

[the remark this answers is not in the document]
martonp: Should we start with the generic one, and add the token specific contracts for the popular tokens later?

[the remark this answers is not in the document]
buck54321: I don't think we should do the generic one at all, but if anything I'd want to start with the token-bound contracts.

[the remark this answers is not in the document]
buck54321: Consider dropping something like this into **updatecontract.sh** too.

```bash
if [ "$VERSION" -eq "0" ]; then
  # Replace a few generated types with the ETH contract versions for interface compatibility.
  perl -0pi -e 's/go-ethereum\/event"/go-ethereum\/event"\n\tethv0 "decred.org\/dcrdex\/dex\/networks\/eth\/contracts\/v0"/' $CONTRACT_FILE

  perl -0pi -e 's/\/\/ ERC20SwapInitiation[^}]*}\n\n//' $CONTRACT_FILE
  perl -0pi -e 's/ERC20SwapInitiation/ethv0.ETHSwapInitiation/g' $CONTRACT_FILE

  perl -0pi -e 's/\/\/ ERC20SwapRedemption[^}]*}\n\n//' $CONTRACT_FILE
  perl -0pi -e 's/ERC20SwapRedemption/ethv0.ETHSwapRedemption/g' $CONTRACT_FILE

  perl -0pi -e 's/\/\/ ERC20SwapSwap[^}]*}\n\n//' $CONTRACT_FILE
  perl -0pi -e 's/ERC20SwapSwap/ethv0.ETHSwapSwap/g' $CONTRACT_FILE
fi
```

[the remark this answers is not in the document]
martonp: Nice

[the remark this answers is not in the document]
martonp: I've made the changes. For now, the contract just has the test token on simnet's address. We'll just need a script that takes a token contract's address, updates it in the solidity code, compiles, and deploys.

dex/networks/erc20/contracts/ERC20SwapV0.sol
```diff
@@ -18,18 +17,23 @@ pragma solidity = 0.8.6;
...
-// by compiling with the correct version of solidity and comparing the
-// resulting byte code to the data in the original transaction.
 contract ERC20Swap {
     bytes4 private constant TRANSFER_FROM_SELECTOR = bytes4(keccak256("transferFrom(address,address,uint256)"));
     bytes4 private constant TRANSFER_SELECTOR = bytes4(keccak256("transfer(address,uint256)"));
+    
+    // TOKEN_ADDRESS must be updated whenever deploying a swap contract for a new token.
+    // Currently it is set to the address where the test token contract on the simnet
+    // is deployed.
+    address public constant TOKEN_ADDRESS = 0x6b4368d3E41a60e20FF8539C843B3CDB38C8A507;
```
buck54321, line 36: Can we add it as a constructor argument?

martonp: It'll be 2k extra gas if we do that.

buck54321: Per transaction? Isn't the constructor only called once when the contract is deployed?

martonp: Yes, but in the address is a constant, then the address will be part of the on chain bytecode instead of part of the storage, and storage lookups cost gas. It's not crazy expensive, ~$1 extra per transaction at the current prices, but it's not so complicated to handle the deployment like this either.

buck54321: Should be good with an **immutable** value, no?

https://docs.soliditylang.org/en/v0.8.10/contracts.html#constant-and-immutable-state-variables

martonp: Ah nice, I didn't know about this. I'll try it.

martonp: This is good to go now, the deployment to simnet works.

dex/networks/erc20/ERC20.go
```diff
@@ -0,0 +1,26 @@
...
...",\"type\":\"string\"}],\"payable\":false,\"stateMutability\":\"view\",\"type\":\"function\"},{\"constant\":false,\"inputs\":[{\"name\":\"_to\",\"type\":\"address\"},{\"name\":\"_value\",\"type\":\"uint256\"}],\"name\":\"transfer\",\"outputs\":[{\"name\":\"\",\"type\":\"bool\"}],\"payable\":false,\"stateMutability\":\"nonpayable\",\"type\":\"function\"},{\"constant\":true,\"inputs\":[{\"name\":\"_owner\",\"type\":\"address\"},{\"name\":\"_spender\",\"type\":\"address\"}],\"name\":\"allowance\",\"outputs\":[{\"name\":\"\",\"type\":\"uint256\"}],\"payable\":false,\"stateMutability\":\"view\",\"type\":\"function\"},{\"payable\":true,\"stateMutability\":\"payable\",\"type\":\"fallback\"},{\"anonymous\":false,\"inputs\":[{\"indexed\":true,\"name\":\"owner\",\"type\":\"address\"},{\"indexed\":true,\"name\":\"spender\",\"type\":\"address\"},{\"indexed\":false,\"name\":\"value\",\"type\":\"uint256\"}],\"name\":\"Approval\",\"type\":\"event\"},{\"anonymous\":false,\"inputs\":[{\"indexed\":true,\"name\":\"from\",\"type\":\"address\"},{\"indexed\":true,\"name\":\"to\",\"type\":\"address\"},{\"indexed\":false,\"name\":\"value\",\"type\":\"uint256\"}],\"name\":\"Transfer\",\"type\":\"event\"}]"
```
chappjc, line 16: Where does the `erc20AbiStr` const come from?  I was expecting us to use the ABI from `ERC20SwapABI` in dex/networks/erc20/contracts/v0/contract.go.

chappjc: I see it used in the token wallet PR now.  A note about the source or method used to generate the JSON is warranted though.

martonp: I've added a comment there. This one is the result of running abigen on the OpenZeppelin ERC20 interface. For the one in the other PR I will run abigen on TestToken.sol.

chappjc: In this PR though, this file isn't used afaict.  Remove then?

martonp: Yeah, I can just add it in the next one. I've removed it.

chappjc (approved):

chappjc:
Good to merge and follow-up in the two PRs that require this, if you are ready @buck54321.

buck54321 (approved):

PR #1365, opened 2021-12-22
client/eth: Nodeclient token functionality

martonp:
This diff adds functions to interact with the token swap contract to the ETH `nodeclient`.

This is a partial implementation, and will be completed by changes in #1369.


client/asset/eth/nodeclient.go
```diff
@@ -531,6 +547,215 @@ func (n *nodeClient) sendSignedTransaction(ctx context.Context, tx *types.Transa
...
+		Pending: true,
+		From:    n.address(),
+		Context: ctx,
+	}
+	return n.tokenSwapContract.IsRefundable(callOpts, secretHash)
+}
+
+// estimateInitTokenGas checks the amount of gas that is used to redeem
+// token swaps.
+func (n *nodeClient) estimateInitTokenGas(ctx context.Context, tokenAddr common.Address, num int) (uint64, error) {
```
martonp, line 673: Unlike with ETH, the gas usage will vary depending on each token's implementation of `transfer` and `transferFrom`.

chappjc:
Two things to consider when you add back in ERC20.go with `ERC20_ABI`:
- The abi string field in the generated code from abigen is deprecated.

```go
// ERC20ABI is the input ABI used to generate the binding from.
// Deprecated: Use ERC20MetaData.ABI instead.
var ERC20ABI = ERC20MetaData.ABI
```

- Looking at `ERC20MetaData`, I noticed they [give is a method `bind.(*MetaData).GetAbi`](https://github.com/ethereum/go-ethereum/blob/d0bd5017ed0281ac1b66172008c694277e4e1ac1/accounts/abi/bind/base.go#L90-L96), which is pretty nifty because it does the abi unmarshalling and caches the resulting `*abi.ABI`.  I suppose we could have been using that already, but I actually like our lock-free solution for the ETH swap abis.   Just something to consider in other spots though, such as this.

buck54321 (approved):
This provides the base functionality needed to run with ERC20. There will obviously be some followup and refactoring per #1381, but I don't see any reason to hold this up.

chappjc (approved):

PR #1367, opened 2021-12-23
client/asset/btc: disallow descriptor wallets

chappjc:
Bitcoin Core v22 creates Descriptor wallets instead of Legacy wallets by default if the DB backend in the build is sqlite instead of bdb.  v0.21 still used Legacy wallets by default as descriptor wallets were experimental in v0.21.

This is a problem because descriptor wallets do not support the `dumpprivkey` RPC, which we need for several reasons, but most importantly for signing the redemption tx input with the private key for the address referenced by the contract script.

https://bitcoincore.org/en/releases/0.21.0/#experimental-descriptor-wallets
"New export RPCs for Descriptor Wallets have not yet been added"

This change adds a `getwalletinfo` check in `(*rpcClient).connect` to ensure that the `Descriptors bool` is false.

This should be cherry-picked to `release-v0.4` for RC2.

### DIFF
--- a/client/asset/btc/rpcclient.go
+++ b/client/asset/btc/rpcclient.go
@@ -34,7 +34,6 @@ const (
 	methodUnlock             = "walletpassphrase"
 	methodLock               = "walletlock"
 	methodPrivKeyForAddress  = "dumpprivkey"
-	methodSignMessage        = "signmessagewithprivkey"
 	methodGetTransaction     = "gettransaction"
 	methodSendToAddress      = "sendtoaddress"
 	methodSetTxFee           = "settxfee"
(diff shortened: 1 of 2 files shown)


chappjc:
For a solution for descriptor wallets, we could investigate if the `signrawtransactionwithwallet` RPC can do this, but it seems unlikely given the non-standard contract redeem script.  Possibly if we push all the [required data](https://github.com/decred/dcrdex/blob/6d5116985076a46595e07a9fa3d3eb0782f70a00/dex/networks/btc/script.go#L538-L544) (pubkey, secret, 0x1, contract) except the signature to the `RawTxWitnessInput.WitnessScript`, it will automagically work, but I'm pretty darn sure it does NOT work like that.  https://github.com/btcsuite/btcd/blob/63438c6d3661a7dec09013f7731035abdf1eb4bc/btcjson/walletsvrcmds.go#L730-L736

The other challenge is signing of other messages with address keys.  The `signmessage` RPC prepends a "Bitcoin Signed Message:\n" string, and the message is text, but the signatures we generate use `btcec.(*PrivateKey).Sign(hash)` of the bytes with no prefix, so this does not seem doable either.

chappjc:
More info. https://github.com/bitcoin/bitcoin/blob/22.x/doc/descriptors.md#including-private-keys

> Including private keys
> 
> Often it is useful to communicate a description of scripts along with the necessary private keys. For this reason, anywhere a public key or xpub is supported, a private key in WIF format or xprv may be provided instead. This is useful when private keys are necessary for hardened derivation steps, or for dumping wallet descriptors including private key material.

It's not clear where that "private key in WIF format" would be.  It is not in the output of `getdescriptorinfo`, `getaddressinfo`, or `listdescriptors`.  Yet?  https://github.com/bitcoin/bitcoin/pull/20226

> Currently the command only returns xpubs which is not very useful in itself, but there are multiples ways to extend it:
>
>   *   add an option to export xprv

That does not seem to have been added to either RPC.

JoeGruffins:
Works.

image

Makes sense to just wait for `dumpprivkey`

JoeGruffins (approved):

PR #1369, opened 2021-12-24
client/core,eth: Token Wallet

### LINKED ISSUE #1348: ethereum token Wallet design (client)
As smart contracts rather than regular externally-owned accounts, ERC-20 token control is tightly coupled with a regular account.

To receive tokens, the used address is the regular account address.  To transfer ERC-20 tokens, the user must have ETH in their regular account to call the `transfer` or `transferFrom` function.

As such, I believe the appropriate design for a `client/asset.Wallet` for an ERC-20 token involves the Ethereum `Wallet` being the **parent** for any token wallet instances.  One way this might work is to introduce a factory method:

```go
client/asset/eth.(*ExchangeWallet).NewTokenWallet(token TokenID, ...) Wallet
```

The token's backend would be tightly coupled with the ETH backend.  Namely, they would share a node, a block monitoring goroutine, account, etc.  The token `Wallet` would have direct access to the fund locking mechanisms for the purposes of reserving ETH for gas of the token txns, and it would have it's own token amount locking mechanisms.

The `Driver` system would still be used, and it can enforce that a regular ETH `Wallet` exist and be opened to power the requested token `Wallet`.

----

Alternatives?

martonp:
Still very rough draft but basic idea is here. 

Closes #1348


buck54321:
@martonp I didn't mean to steamroll you with #1381 -> #1399, but that's covers everything from this PR. Looking back at this, it's evident that we arrived at nearly the same ideas in a few places, so I'm confident we're on the right track. I'm going to recommend either closing this PR or working through the differences if you feel there is something here that is missed in the other work.

martonp:
@buck54321 Sounds good, I'll review those PRs.

PR #1370, opened 2021-12-26
client/dcr: log correct hash GetBlockVerbose err

jholdstock:


jholdstock:
Ugh, nevermind. Too early, not enough coffee.

PR #1371, opened 2021-12-26
client/core: use fee suggestion for Withdraw

chappjc:
**Must** go in v0.4

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -3339,12 +3339,12 @@ func (c *Core) MaxBuy(host string, base, quote uint32, rate uint64) (*MaxOrderEs
 		return nil, errors.New("cannot divide by lot size zero")
 	}
 
-	swapFeeSuggestion := c.feeSuggestion(dc, quote, false)
+	swapFeeSuggestion := c.feeSuggestion(dc, quote)
 	if swapFeeSuggestion == 0 {
 		return nil, fmt.Errorf("failed to get swap fee suggestion for %s at %s", unbip(quote), host)
 	}
 
-	redeemFeeSuggestion := c.feeSuggestion(dc, base, false)
+	redeemFeeSuggestion := c.feeSuggestion(dc, base)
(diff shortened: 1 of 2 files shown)


client/core/core.go
```diff
@@ -3637,7 +3660,7 @@ func (c *Core) Withdraw(pw []byte, assetID uint32, value uint64, address string)
 	if err != nil {
 		return nil, err
 	}
-	const feeSuggestion = 100
```
chappjc, line 3640: Nasty surprise paying $10 to withdraw when 50 cents is sufficient.

client/core/core.go
```diff
@@ -3607,9 +3607,32 @@ func (c *Core) notifyFee(dc *dexConnection, coinID []byte) error {
...
+			return feeSuggestion
+		}
+	}
+	return 0
+}
+
 // feeSuggestion gets the best fee suggestion, first from a synced order book,
 // and if not synced, directly from the server.
-func (c *Core) feeSuggestion(dc *dexConnection, assetID uint32, isBase bool) (feeSuggestion uint64) {
+func (c *Core) feeSuggestion(dc *dexConnection, assetID uint32, isBase /* ??? */ bool) (feeSuggestion uint64) {
```
chappjc, line 3635: @buck54321 I thought you had called this out as one seemingly pointless thing that was going to see use in order options work, but I don't see that changes in your work.  I'd like to remove this arg unless you have a reason for it since it's confusing.

JoeGruffins (approved):

client/core/core.go
buck54321: Oof. I don't know why tf I would have done this.

client/core/core.go
buck54321: I guess not. Go ahead.

chappjc: Removed.

buck54321 (approved):

PR #1374, opened 2021-12-28
client: Add RescanWallet for BTC SPV, with RPC and HTTP API handlers.

chappjc:
Requires https://github.com/decred/dcrdex/pull/1383 (the **first commit** in this PR).

Three other commits added in this work:

**`client/asset/btc: add (*spvWallet).rescanWallet`**

Based on btcwallet's current approach for forcing a rescan, this drops all transactions in the wallet db and restarts the wallet.

**`rescanwallet` RPC** 

Trigger wallet recsan from dexcctl.

**client/webserver: rescanwallet http api and button handler**

Add a `(*WebServer).apiRescanWallet` for calling core's rescan wallet method.  Use the `WalletTrait`s that are now in `core.WalletState`.

**IMPORTANT NOTE:**  When you click "Rescan" it is an async process, so the sync status icon will appear and start updating percent.  Due to the issue we have remaining with the wallet birthday and the chain server syncing, it shows 100% at first (although still not synced), but when it reaches the wallet birthday again it goes to sensible number and starts going up as the rescan progresses.  There is [existing commented code here](https://github.com/decred/dcrdex/blob/864f4d01daf040614c55b7b381011324f7a6af73/client/asset/btc/spv.go#L556) that was waiting on a [neutrino PR](https://github.com/lightninglabs/neutrino/pull/235), which was recently merged, so I will update that in another PR soon

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -536,6 +536,23 @@ type ExchangeWallet struct {
 	findRedemptionQueue map[outPoint]*findRedemptionReq
 }
 
+// ExchangeWalletSPV embeds an ExchangeWallet, and also provides the Rescan
+// method to implement asset.Rescanner.
+type ExchangeWalletSPV struct {
+	*ExchangeWallet
+}
+
+// Rescan satisfies the asset.Rescanner interface, and issues a rescan wallet
+// command if the backend is an SPV wallet.
+func (btc *ExchangeWalletSPV) Rescan(_ context.Context) error {
+	// This will panic if not an spvWallet, which would indicate that
+	// openSPVWallet was not used to construct this instance.
+	w := btc.node.(*spvWallet)
+	atomic.StoreInt64(&btc.tipAtConnect, 0) // for progress
+	// Caller should start calling SyncStatus on a ticker.
+	return w.rescanWalletAsync()
+}
+
 type block struct {
 	height int64
 	hash   chainhash.Hash
@@ -727,8 +744,10 @@ func newUnconnectedWallet(cfg *BTCCloneCFG, walletCfg *WalletConfig) (*ExchangeW
 	return w, nil
 }
 
+var _ asset.Wallet = (*ExchangeWallet)(nil)
+
 // openSPVWallet opens the previously created native SPV wallet.
-func openSPVWallet(cfg *BTCCloneCFG) (*ExchangeWallet, error) {
+func openSPVWallet(cfg *BTCCloneCFG) (*ExchangeWalletSPV, error) {
 	walletCfg := new(WalletConfig)
 	err := config.Unmapify(cfg.WalletCFG.Settings, walletCfg)
 	if err != nil {
@@ -747,10 +766,11 @@ func openSPVWallet(cfg *BTCCloneCFG) (*ExchangeWallet, error) {
 
 	btc.node = loadSPVWallet(cfg.WalletCFG.DataDir, cfg.Logger.SubLogger("SPV"), peers, cfg.ChainParams)
 
-	return btc, nil
+	return &ExchangeWalletSPV{btc}, nil
 }
 
-var _ asset.Wallet = (*ExchangeWallet)(nil)
+var _ asset.Wallet = (*ExchangeWalletSPV)(nil)
+var _ asset.Rescanner = (*ExchangeWalletSPV)(nil)
 
 // Info returns basic information about the wallet and asset.
 func (btc *ExchangeWallet) Info() *asset.WalletInfo {
--- a/client/asset/btc/spv.go
+++ b/client/asset/btc/spv.go
@@ -174,7 +174,7 @@ func createSPVWallet(privPass []byte, seed []byte, dbDir string, log dex.Logger,
 }
 
 var (
-	// loggingInited will be set when the log rotator has been initilized.
+	// loggingInited will be set when the log rotator has been initialized.
 	loggingInited uint32
 )
 
@@ -299,6 +299,10 @@ type spvWallet struct {
 	tipChan            chan *block
 	syncTarget         int32
 	lastPrenatalHeight int32
+
+	// rescanStarting is set while reloading the wallet and dropping
+	// transactions from the wallet db.
+	rescanStarting uint32 // atomic
 }
 
 var _ Wallet = (*spvWallet)(nil)
@@ -531,10 +535,10 @@ func (w *spvWallet) syncStatus() (*syncStatus, error) {
 	var blk *block
 	// Wallet address manager sync height.
 	if chainBlk.Timestamp.After(w.birthday) {
-		// After birthday, wallet address manager should be syncing. Although
-		// block time stamps are not necessarily monotonically increasing, this
-		// is a reasonable condition at which the wallet's sync height should be
-		// consulted instead of the chain service's height.
+		// After the wallet's birthday, the wallet address manager should begin
+		// syncing. Although block time stamps are not necessarily monotonically
+		// increasing, this is a reasonable condition at which the wallet's sync
+		// height should be consulted instead of the chain service's height.
 		walletBlock := w.wallet.syncedTo()
 		if walletBlock.Height == 0 {
 			// The wallet is about to start its sync, so just return the last
@@ -1093,6 +1097,7 @@ func (w *spvWallet) startWallet() error {
 		return errors.New("wallet not found")
 	}
 
+	w.log.Debug("Starting native BTC wallet...")
 	btcw, err := w.loader.OpenExistingWallet([]byte(wallet.InsecurePubPassphrase), false)
 	if err != nil {
 		return fmt.Errorf("couldn't load wallet: %w", err)
@@ -1100,7 +1105,7 @@ func (w *spvWallet) startWallet() error {
 
 	bailOnWallet := func() {
 		if err := w.loader.UnloadWallet(); err != nil {
-			w.log.Errorf("Error unloading wallet after loadChainClient error: %v", err)
+			w.log.Errorf("Error unloading wallet: %v", err)
 		}
 	}
 
@@ -1113,7 +1118,7 @@ func (w *spvWallet) startWallet() error {
 
 	bailOnWalletAndDB := func() {
 		if err := w.neutrinoDB.Close(); err != nil {
-			w.log.Errorf("Error closing neutrino database after loadChainClient error: %v", err)
+			w.log.Errorf("Error closing neutrino database: %v", err)
 		}
 		bailOnWallet()
 	}
@@ -1134,6 +1139,7 @@ func (w *spvWallet) startWallet() error {
 			w.connectPeers = []string{"localhost:20575"}
 		}
 	}
+	w.log.Debug("Starting neutrino chain service...")
 	chainService, err := neutrino.NewChainService(neutrino.Config{
 		DataDir:       w.netDir,
 		Database:      w.neutrinoDB,
@@ -1154,7 +1160,7 @@ func (w *spvWallet) startWallet() error {
 
 	bailOnEverything := func() {
 		if err := chainService.Stop(); err != nil {
-			w.log.Errorf("Error closing neutrino chain service after loadChainClient error: %v", err)
+			w.log.Errorf("Error closing neutrino chain service: %v", err)
 		}
 		bailOnWalletAndDB()
 	}
@@ -1166,14 +1172,71 @@ func (w *spvWallet) startWallet() error {
 
 	if err = w.chainClient.Start(); err != nil { // lazily starts connmgr
 		bailOnEverything()
-		if err != nil {
-			w.log.Errorf("error unloading wallet after chain client start error: %v", err)
-		}
 		return fmt.Errorf("couldn't start Neutrino client: %v", err)
 	}
 
+	w.log.Info("Synchronizing wallet with network...")
 	btcw.SynchronizeRPC(w.chainClient)
 
+	return nil
+}
+
+// rescanWalletAsync initiates a full wallet recovery (used address discovery
+// and transaction scanning) by stopping the btcwallet, dropping the transaction
+// history from the wallet db, resetting the synced-to height of the wallet
+// manager, restarting the wallet and its chain client, and finally commanding
+// the wallet to resynchronize, which starts asynchronous wallet recovery.
+// Progress of the rescan should be monitored with syncStatus. During the rescan
+// wallet balances and known transactions may not be reported accurately or
+// located. The neutrinoService is not stopped, so most spvWallet methods will
+// continue to work without error, but methods using the btcWallet will likely
+// return incorrect results or errors.
+func (w *spvWallet) rescanWalletAsync() error {
+	if !atomic.CompareAndSwapUint32(&w.rescanStarting, 0, 1) {
+		return errors.New("rescan already in progress")
+	}
+	defer atomic.StoreUint32(&w.rescanStarting, 0)
+
+	// Stop the wallet, but do not use w.loader.UnloadWallet because it also
+	// closes the database.
+	btcw, ok := w.loader.LoadedWallet()
+	if !ok {
+		return errors.New("wallet not loaded")
+	}
+	wdb := btcw.Database()
+
+	w.log.Info("Stopping wallet and chain client...")
+	btcw.Stop() // stops Wallet and chainClient (not chainService)
+	btcw.WaitForShutdown()
+	w.chainClient.WaitForShutdown()
+
+	// Force a full rescan with active address discovery on wallet restart by
+	// dropping the complete transaction history. See the
+	// btcwallet/cmd/dropwtxmgr app for more information.
+	w.log.Info("Dropping transaction history to perform full rescan...")
+	err := wallet.DropTransactionHistory(wdb, false)
+	if err != nil {
+		w.log.Errorf("Failed to drop wallet transaction history: %v", err)
+		// Continue to attempt restarting the wallet anyway.
+	}
+
+	err = walletdb.Update(wdb, func(dbtx walletdb.ReadWriteTx) error {
+		ns := dbtx.ReadWriteBucket([]byte("waddrmgr")) // it'll be fine
+		return btcw.Manager.SetSyncedTo(ns, nil)       // never synced, forcing recover from birthday
+	})
+	if err != nil {
+		w.log.Errorf("Failed to reset wallet manager sync height: %v", err)
+	}
+
+	w.log.Info("Starting wallet...")
+	btcw.Start()
+
+	if err = w.chainClient.Start(); err != nil {
+		return fmt.Errorf("couldn't start Neutrino client: %v", err)
+	}
+
+	w.log.Info("Synchronizing wallet with network...")
+	btcw.SynchronizeRPC(w.chainClient)
 	return nil
 }
 
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -62,8 +62,8 @@ var (
 	// The coin waiters will query for transaction data every recheckInterval.
 	recheckInterval = time.Second * 5
 	// When waiting for a wallet to sync, a SyncStatus check will be performed
-	// ever syncTickerPeriod. var instead of const for testing purposes.
-	syncTickerPeriod = 10 * time.Second
+	// every syncTickerPeriod. var instead of const for testing purposes.
+	syncTickerPeriod = 3 * time.Second
 	// serverAPIVers are the DEX server API versions this client is capable
 	// of communicating with.
 	//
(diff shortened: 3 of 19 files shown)


chappjc:
The third commit gives us the following "Rescan" button for an `asset.Rescanner`;

image

client/asset/interface.go
```diff
@@ -10,6 +10,13 @@ import (
 	"decred.org/dcrdex/dex"
 )
 
+type WalletTrait uint64
+
+const (
+	WalletTraitRescanner WalletTrait = 1 << iota
+	WalletTraitSender                // a Send method to complement Withdraw
```
chappjc, line 17: Here's where we could also have a `Sweeper`'s trait. See: https://github.com/decred/dcrdex/pull/1355#discussion_r775917934

buck54321: I love this, but I wish it were only used at a higher level, e.g. as a field of `core.xcWallet` that is populated when the wallet is opened based on interface type-checks. Having the assets self-report sets up potential inconsistencies between the UI and the available API.

chappjc: Can do, I think.  It would be better for the implementations to set these traits based on their methods.  Will just have to work out how to get this info to frontends then.  Because I'm doing the following to communicate this based on the walletState: https://github.com/decred/dcrdex/pull/1374/files#diff-a0c3b3a03517588f82d5bb3553e3e46ef5377c8bbcb4f8813ebb6d3cb0a6ee86R128-R139

chappjc: Hmm, I suppose the answer would be to put it in the `WalletState`s rather than the `WalletDefinition`s.  I think I considered that and ended up thinking it was not really "state" but then again we have Type there too.

buck54321: `WalletState` is what I was thinking.

buck54321: The bitfield could be created in `asset.OpenWallet` too and added as a return argument.

client/core/types.go
```diff
@@ -125,6 +125,19 @@ type SupportedAsset struct {
...
 }
 
+func (sa *SupportedAsset) IsRescanner() bool {
+	if sa.Wallet == nil {
+		return false
+	}
+	walletType := sa.Wallet.WalletType
+	for _, wd := range sa.Info.AvailableWallets {
+		if wd.Type == walletType {
+			return wd.Traits&asset.WalletTraitRescanner == asset.WalletTraitRescanner
```
chappjc, line 135: For background, this method was modeled similarly to this frontend method for accessing a def from a SupportedAsset based on Type:

https://github.com/decred/dcrdex/blob/fb7bb9da37a60f7cd6f4bea7707d11c31fd77610/client/webserver/site/src/js/app.js#L683-L687

If we go with a `WalletState` field instead, this would reduce to `return sa.Wallet.Traits.IsRescanner()` or similar.

buck54321: We can definitely echo the bitfield implementation in javascript.

chappjc:
Good lord it's a trick to get the `btcwallet/wallet.Wallet` to recover (address discovery and rescan), a process that's akin to dcrwallet's `discoverusage` followed by `rescanwallet`.  This is as close as you can get to reseeding the wallet without actually reseeding the wallet.

This works flawlessly by dropping all the txns and then creating a new instance of a btcwallet `*Wallet`, however that's all kinds of trouble in our `spvWallet` type where we have concurrency issues around the wallet and loader fields, in addition to it invalidating the transaction notification channel.   I believe the secret sauce that ultimately let me just do `btcw.Start()` after previously stopping it was to clear out that wallet manager's syced-to height via `btcw.Manager.SetSyncedTo` to instruct the `btcwallet/wallet.(*Wallet).recovery` method to start from the wallet's birthday block to discover address use.

I'll wrap up the traits and webserver business now.

client/asset/interface.go
chappjc: Moved it up to core (`xcWallet` and `WalletState`) in https://github.com/decred/dcrdex/pull/1375

chappjc:
Adding WalletTrait separately in https://github.com/decred/dcrdex/pull/1375.  That can get reviewed and merged first so both this and ETH Send/Sweep can be addressed.

For now the `(*spvWallet).rescanWallet` method added in the second commit is alright for review, as is the RPC server handler in the third.

The fourth commit where I add the webserver api handler and wire up a click action to the Rescan button could use some input too, but I'm thinking it still needs a form to show in click so the user can confirm ("This will perform a full rescan of the wallet, during which the wallet may not be able to perform other actions. Only attempt to rescan with no active orders or settling matches! [Do it anyway]".

Note that the current design is to have the wallet rescan be asynchronous, or at least the `spvWallet`'s implementation, with `(*Core).RescanWallet` starting a `SyncStatus` loop just like on wallet connect so that wallet state notes are sent to consumers.

client/asset/btc/spv.go
```diff
@@ -1196,6 +1197,81 @@ func (w *spvWallet) startWallet() error {
 	return nil
 }
 
+func (w *spvWallet) rescanWalletAsync() error {
```
chappjc, line 1200: This approach is all to keep the neutrino chain service running and only touch the wallet and its DB.  Otherwise we could have rescan handled with a full stop / drop / start, mostly coordinated higher up by the ExchangeWallet.
I'm still not certain this approach is flawless, but the wallet does go back to running as normal, eventually anyway.

chappjc:
Related, new wiki page: https://github.com/decred/dcrdex/wiki/Native-BTC-Wallet-Recovery-and-Rescanning

chappjc:
Rebased on the updated  that determines traits just from method signatures.  If in the future we decide it's not an easy approach, we can go back to defining the traits explicitly as in 92f0a6cb6a40e4aa481a59c85dd26ec5cf8c9d5c, and without the `ExchangeWalletSPV` type like it was in  https://github.com/decred/dcrdex/pull/1374/commits/278cc9a1daf202ffa8bb7489b26c2eebc4ecc47a

client/webserver/site/src/js/wallets.js
```diff
@@ -237,6 +240,20 @@ export default class WalletsPage extends BasePage {
     await this.newWalletForm.loadDefaults()
   }
 
+  async rescanWallet (assetID) {
+    if (app().haveAssetOrders(assetID)) {
+      const asset = app().assets[assetID]
+      app().notify(ntfn.make('Cannot rescan wallet.', `There are active orders involving ${asset.info.name}`, ntfn.ERROR)) // TODO: translate
+      return // TODO: show override prompt
```
chappjc, line 247: image

Users could still use dexcctl because that's more of a power user tool, but I don't want users clicking the "Rescan" button willy-nilly.  I'm thinking a form can open if there are active orders to allow the user to initiate the request anyway.

client/asset/btc/btc.go
```diff
@@ -536,6 +536,24 @@ type ExchangeWallet struct {
 	findRedemptionQueue map[outPoint]*findRedemptionReq
 }
 
+// ExchangeWalletSPV embeds a ExchangeWallet, but also provides the Rescan
+// method to implement asset.Rescanner.
+type ExchangeWalletSPV struct {
+	*ExchangeWallet
+}
```
chappjc, line 543: I don't know how often we'll get away with this kind of thing or the `NewAddress` dummy method, but we can try it out for a while and if it becomes hard to deal with we can switch back to explicitly set traits as described in https://github.com/decred/dcrdex/pull/1374#issuecomment-1016840825

client/asset/btc/spv.go
```diff
@@ -1238,6 +1238,81 @@ func (w *spvWallet) startWallet() error {
...
+		return errors.New("wallet not loaded")
+	}
+	wdb := btcw.Database()
+
+	w.log.Infof("Stopping wallet and chain client...")
+	btcw.Stop() // stops Wallet and chainClient (not chainService)
+	btcw.WaitForShutdown()
+	w.chainClient.WaitForShutdown()
+
+	// Force a full recsan with active address discovery on wallet restart by
```
JoeGruffins, line 1255: recsan -> rescan

client/asset/btc/spv.go
```diff
@@ -1238,6 +1238,81 @@ func (w *spvWallet) startWallet() error {
...
+func (w *spvWallet) rescanWalletAsync() error {
+	// Stop the wallet, but do not use w.loader.UnloadWallet because it also
+	// closes the database.
+	btcw, ok := w.loader.LoadedWallet()
+	if !ok {
+		return errors.New("wallet not loaded")
+	}
+	wdb := btcw.Database()
+
+	w.log.Infof("Stopping wallet and chain client...")
```
JoeGruffins, line 1250: Nothing to format.

client/asset/btc/spv.go
```diff
@@ -1238,6 +1238,81 @@ func (w *spvWallet) startWallet() error {
...
+		w.log.Errorf("Failed to reset wallet manager sync height: %v", err)
+	}
+
+	w.log.Info("Starting wallet...")
+	btcw.Start()
+
+	if err = w.chainClient.Start(); err != nil {
+		if err != nil {
+			w.log.Errorf("error unloading wallet after chain client start error: %v", err)
+		}
```
JoeGruffins, line 1279: `if err != nil` twice

client/asset/btc/spv.go
```diff
@@ -1238,6 +1238,81 @@ func (w *spvWallet) startWallet() error {
...
+	for {
+		select {
+		case <-ticker.C:
+			ss, err := w.syncStatus()
+			if err != nil {
+				w.log.Errorf("Unable to get btcwallet sync status (will try again): %v", err)
+				continue // keep trying since the loop will break if connection really drops
+			}
+			if !ss.Syncing {
+				w.log.Infof("Synchronized.")
```
JoeGruffins, line 1306: Nothing to format.

client/core/core.go
```diff
@@ -1923,6 +1911,68 @@ func (c *Core) WalletState(assetID uint32) *WalletState {
...
+			case <-ticker.C:
+				if c.walletCheckAndNotify(wallet) {
+					return
+				}
+
+			case <-c.ctx.Done():
+				return
+			}
+		}
+	}()
```
JoeGruffins, line 1971: Maybe fine, but I want to point out that both the wallet's `rescanWalletAsync` and core's `updateWalletBalance` have db writes and nothing is waiting for this goroutine to finish. Maybe golang will handle this fine, but what if it just exits the app in the middle of a db write? imo app shutdown should wait for this to finish.

client/rpcserver/handlers.go
```diff
@@ -431,6 +433,20 @@ func handleWithdraw(s *RPCServer, params *RawParams) *msgjson.ResponsePayload {
...
 
+func handleRescanWallet(s *RPCServer, params *RawParams) *msgjson.ResponsePayload {
+	assetID, err := parseRescanWalletArgs(params)
+	if err != nil {
+		return usage(withdrawRoute, err)
+	}
+	err = s.core.RescanWallet(assetID)
+	if err != nil {
+		errMsg := fmt.Sprintf("unable to rescan wallet: %v", err)
+		resErr := msgjson.NewError(msgjson.RPCWithdrawError, errMsg)
```
JoeGruffins, line 444: Need a new `msgjson.RPCRescanError` I guess.

client/asset/btc/spv.go
```diff
@@ -1238,6 +1238,81 @@ func (w *spvWallet) startWallet() error {
...
+	w.log.Info("Dropping transaction history to perform full rescan...")
+	err := wallet.DropTransactionHistory(wdb, false)
+	if err != nil {
+		w.log.Errorf("Failed to drop wallet transaction history: %v", err)
+		// Continue to attempt restarting the wallet anyway.
+	}
+
+	err = walletdb.Update(wdb, func(dbtx walletdb.ReadWriteTx) error {
+		ns := dbtx.ReadWriteBucket([]byte("waddrmgr")) // it'll be fine
+		return btcw.Manager.SetSyncedTo(ns, nil)       // never synched
```
buck54321, line 1267: synched

client/asset/btc/spv.go
```diff
@@ -1238,6 +1238,81 @@ func (w *spvWallet) startWallet() error {
...
+				w.log.Errorf("Unable to get btcwallet sync status (will try again): %v", err)
+				continue // keep trying since the loop will break if connection really drops
+			}
+			if !ss.Syncing {
+				w.log.Infof("Synchronized.")
+				return nil
+			}
+
+		case <-ctx.Done():
+			return ctx.Err()
```
buck54321, line 1311: So the caller can skip logging on `context.Canceled`?

client/core/core.go
```diff
@@ -1922,6 +1911,68 @@ func (c *Core) WalletState(assetID uint32) *WalletState {
...
+	}
+
+	// Begin potentially asynchronous wallet rescan operation.
+	if err = wallet.Rescan(c.ctx); err != nil {
+		return err
+	}
+
+	if c.walletCheckAndNotify(wallet) {
+		return nil // sync done
+	}
```
buck54321, line 1954: Would we ever expect this to be the case?

client/core/core.go
```diff
@@ -1922,6 +1911,68 @@ func (c *Core) WalletState(assetID uint32) *WalletState {
...
+
+	// Resync still running. Launch a status update goroutine.
+	go func() {
+		ticker := time.NewTicker(time.Second)
```
buck54321, line 1958: Why not `syncTickerPeriod`? If you did, the entire goroutine could be moved to a method used both here and `connectWallet`.

client/rpcserver/handlers.go
buck54321: Feels odd having a **msgjson** error code that doesn't correspond to any errors returned over the main API, but we're sort of stuck with our previous design choice here in **rpcserver**.

client/webserver/site/src/js/wallets.js
```diff
@@ -237,6 +240,20 @@ export default class WalletsPage extends BasePage {
     await this.newWalletForm.loadDefaults()
   }
 
+  async rescanWallet (assetID) {
+    if (app().haveAssetOrders(assetID)) {
```
buck54321, line 244: We usually just check this in **core**, where we iterate `(*dexConnection).hasActiveAssetOrders`. I guess by doing it this way, you allow the rescan to be forced via **rpcserver**?

buck54321 (changes requested):
Pretty much ready. A couple of questions.

client/core/core.go
chappjc: Depends if the `wallet.Rescan` implementation is async or not... or at least that was my thinking.  I wasn't ready to formalize `Rescan` as necessarily (a)synchronous.  Hmmm

client/asset/interface.go
```diff
@@ -233,6 +268,28 @@ type Wallet interface {
...
+// withdrawing a certain amount from the source wallet/account.
+type Sender interface {
+	Send(address string, value, feeSuggestion uint64) (Coin, error)
+}
+
+// Sweeper is a wallet that can clear the entire balance of the wallet/account
+// to an address. Similar to Withdraw, but no input value is required.
+type Sweeper interface {
+	Sweep(address string, feeSuggestion uint64) (Coin, error)
+}
```
chappjc, line 286: Will toast these unused interfaces.  This was just feeling out the uses.

client/core/core.go
chappjc: Good point.  I'll see about making it a supervised task that holds up shutdown.

client/rpcserver/handlers.go
chappjc: Yeah, it's really unfortunate at this point that the same enum is used for both sorts of errors in completely different domains.

client/webserver/site/src/js/wallets.js
chappjc: Correct.  I started with it in Core, but went to frontend because there are legit reasons to rescan even with swaps and I didn't want to close the door to doing that with rpcserver or even a "force rescan" dialog on the frontend.

chappjc: That reminds me, I've prepared a simplified version of the `dropwtxmgr` tool from the btcwallet repo: https://github.com/decred/dcrdex/commit/4a1fdb71cda620618085a874cdd1e5d648e3ff98

client/cmd/btcwdroptxs/main.go would be a fallback way to force a rescan after shutting down dexc.
It tests well and is built from our repo, so I might PR this.

buck54321: > Correct. I started with it in Core, but went to frontend because there are legit reasons to rescan even with swaps (might even be the only way to finish a swap, even to refund or auto-redeem if btcwallet failed to recognize your own swap, who knows?) and I didn't want to close the door to doing that with rpcserver or even a "force rescan" dialog on the frontend.

Maybe the most flexible way is to add a `force bool` argument to `(*Core).RescanWallet` and to the rpcserver method.

client/asset/interface.go
buck54321: Meh. The docs serve some purpose, even if not implemented. Nuke if you want, but I don't really mind.

client/webserver/site/src/js/wallets.js
chappjc: I rejected that idea in planning this because the caller still needs to provide the value for that bool when calling.  For frontend, that still means a prompt, so options there are (a) prompt based on known order info as done here, (b) just try rescan with force=false and prompt based on a _specific error_ returned, or (c) a combination of those.

I suppose an advantage to an arg on Core.RescanWallet would be that you could never accidentally rescan with an active  order unless you explicitly said so, which could happen if you had some automatic order placement routine going on concurrently.  It would also make the rpc a bit safer, so I'll make that change, but to avoid a round trip http api rescan request for the frontend, I'd still like to have the JS check.

buck54321: It can just be hard-coded to `force = false` for **webserver** and `force = true` for **rpcserver** for now, and the option added to the UI in the future without changes to the API.

chappjc: > hard-coded to `force = false` for **webserver**

This is what I'm doing now for webserver, but making it an arg defaulting to false for rpcserver

chappjc:
WIP with last commit, but working toward addressing reviews.

client/asset/btc/spv.go
```diff
@@ -1238,6 +1238,53 @@ func (w *spvWallet) startWallet() error {
 	return nil
 }
 
+func (w *spvWallet) rescanWalletAsync() error {
+	// Stop the wallet, but do not use w.loader.UnloadWallet because it also
+	// closes the database.
```
chappjc, line 1243: I need to add an atomic here to block repeated calls to `rescanWalletAsync`

chappjc:
Now rebased on https://github.com/decred/dcrdex/pull/1383.  Consider this draft for the moment.  I need to test out some changes.

client/core/core.go
chappjc: This is done in the latest commit on this PR now, but also note I've factored things a bit further in https://github.com/decred/dcrdex/pull/1445/commits/2f0a21db07fc5f39ed2971d07a7f21386c1bfeb9, which builds on this.

client/webserver/site/src/js/app.js
```diff
@@ -658,6 +658,19 @@ export default class Application {
...
+        if (!market.orders) continue
+        for (const ord of market.orders) {
+          if ((ord.baseID === assetID || ord.quoteID === assetID) &&
+            (ord.status < StatusExecuted || hasLiveMatches(ord))) return true
+        }
+      }
+    }
+    return false
+  }
+
```
buck54321, line 673: We no longer need `haveAssetOrders` since `*Core` does it now.

buck54321 (changes requested):

client/asset/btc/spv.go
```diff
@@ -507,6 +511,8 @@ func (w *spvWallet) syncHeight() int32 {
 	return maxHeight
 }
 
+var lastPrenatalHeight int32
```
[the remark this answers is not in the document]
chappjc: Can do that.  I really wish Go had `static` function scoped variables sometimes.

client/webserver/site/src/js/app.js
chappjc: Will removed.  Although I thought it was a good complement to the backend check to avoid a round trip api request looking for the error code, and would also be nice when a Force Rescan prompt is added.  It's no matter though, might as well have a single source of logic on this.

client/asset/btc/spv.go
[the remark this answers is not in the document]
chappjc: Oh, this is still being added in https://github.com/decred/dcrdex/pull/1383, which this PR is now rebased.  Sorry for changing that PR order up.  I'll resolve this variable scope in https://github.com/decred/dcrdex/pull/1383.  If you ended up reviewing that PR via the first commit of this PR, you might be able to drop an approval on that one when I fix up this variable.

chappjc:
Changes applied @buck54321 but not yet re-tested.

chappjc:
Retested on testnet and simnet after latest frontend change and `lastPrenatalHeight` move.

buck54321 (approved):

PR #1375, opened 2021-12-31
client/asset,webserver: wallet traits and hide eth new address button

chappjc:
Required for https://github.com/decred/dcrdex/pull/1374 and a future solution to the Send vs. Withdraw (or Sweep) issue in https://github.com/decred/dcrdex/pull/1355#discussion_r775917934

This PR adds a `client/core.WalletTrait` type that is a bitset communicating various optional functionality of a `client/asset.Wallet` in an `xcWallet` (via a new `core.WalletState` field).

As an initial application of the traits, this hides the ETH wallet's "New deposit address" button.

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -2553,13 +2553,19 @@ func (btc *ExchangeWallet) refundTx(txHash *chainhash.Hash, vout uint32, contrac
 	return msgTx, nil
 }
 
-// Address returns a new external address from the wallet.
+// Address returns an external address from the wallet.
 func (btc *ExchangeWallet) Address() (string, error) {
 	addr, err := btc.externalAddress()
 	if err != nil {
 		return "", err
 	}
 	return addr.String(), nil
+}
+
+// NewAddress returns a new address from the wallet. This satisfies the
+// NewAddresser interface.
+func (btc *ExchangeWallet) NewAddress() (string, error) {
+	return btc.Address()
 }
 
 // PayFee sends the dex registration fee. Transaction fees are in addition to
--- a/client/asset/interface.go
+++ b/client/asset/interface.go
@@ -10,6 +10,39 @@ import (
 	"decred.org/dcrdex/dex"
 )
 
+// WalletTrait is a bitset indicating various optional wallet features, such as
+// the presence of auxiliary methods like Rescan and Send.
+type WalletTrait uint64
+
+const (
+	WalletTraitRescanner    WalletTrait = 1 << iota // The Wallet is an asset.Rescanner.
+	WalletTraitNewAddresser                         // The Wallet can generate new addresses on demand with NewAddress.
+)
+
+// IsRescanner tests if the WalletTrait has the WalletTraitRescanner bit set.
+func (wt WalletTrait) IsRescanner() bool {
+	return wt&WalletTraitRescanner != 0
+}
+
+// IsNewAddresser tests if the WalletTrait has the WalletTraitNewAddresser bit
+// set, which indicates the presence of a NewAddress method that will generate a
+// new address on each call. If this method does not exist, the Address method
+// should be assumed to always return the same deposit address.
+func (wt WalletTrait) IsNewAddresser() bool {
+	return wt&WalletTraitNewAddresser != 0
+}
+
+// DetermineWalletTraits returns the WalletTrait bitset for the provided Wallet.
+func DetermineWalletTraits(w Wallet) (t WalletTrait) {
+	if _, is := w.(Rescanner); is {
+		t |= WalletTraitRescanner
+	}
+	if _, is := w.(NewAddresser); is {
+		t |= WalletTraitNewAddresser
+	}
+	return t
+}
+
 // CoinNotFoundError is returned when a coin cannot be found, either because it
 // has been spent or it never existed. This error may be returned from
 // AuditContract, Refund or Redeem as those methods expect the provided coin to
@@ -27,7 +60,9 @@ type WalletDefinition struct {
 	// seed that should be used to set the wallet key(s). This would be
 	// true for built-in wallets.
 	Seeded bool `json:"seeded"`
-	// Type is a string identifying the wallet type.
+	// Type is a string identifying the wallet type. NOTE: There should be a
+	// particular WalletTrait set for any given Type, but wallet construction is
+	// presently required to discern traits.
 	Type string `json:"type"`
 	// Tab is a displayable string for the wallet type. One or two words. First
 	// word capitalized. Displayed on a wallet selection tab.
@@ -111,7 +146,7 @@ type Wallet interface {
 	// It should be assumed that once disconnected, subsequent Connect calls
 	// will fail, requiring a new Wallet instance.
 	dex.Connector
-	// Info returns a set of basic information about the wallet.
+	// Info returns a set of basic information about the wallet driver.
 	Info() *WalletInfo
 	// Balance should return the balance of the wallet, categorized by
 	// available, immature, and locked. Balance takes a list of minimum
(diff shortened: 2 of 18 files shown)


client/webserver/site/src/html/wallets.tmpl
```diff
@@ -23,20 +23,22 @@
...
-  <button data-action="deposit"{{if not $ready}} class="d-hide"{{end}}>[[[Deposit]]]</button>
-  <button data-action="lock"{{if or (not $ready) (not $w.Encrypted)}} class="d-hide"{{end}}>[[[Lock]]]</button>
-  <button data-action="create" class="d-hide">[[[create_a_x_wallet]]]</button>
+    {{$ready := and $w.Running $w.Open}}
+    <button data-action="connect"{{if $w.Running}} class="d-hide"{{end}}>[[[Connect]]]</button>
+    <button data-action="unlock"{{if $w.Open}} class="d-hide"{{end}}>[[[Unlock]]]</button>
+    <button data-action="withdraw"{{if not $ready}} class="d-hide"{{end}}>[[[Withdraw]]]</button>
+    <button data-action="deposit"{{if not $ready}} class="d-hide"{{end}}>[[[Deposit]]]</button>
+    <button data-action="lock"{{if or (not $ready) (not $w.Encrypted)}} class="d-hide"{{end}}>[[[Lock]]]</button>
+    <button data-action="rescan"{{if not $w.Traits.IsRescanner}} class="d-hide"{{end}}>[[[Rescan]]]</button>
```
chappjc, line 32: The button never displays in this PR because no wallet's have this trait.  The button also has no bound action.  Both of those things are in https://github.com/decred/dcrdex/pull/1374.  this is just for illustration why we need a field in `WalletState` (or somewhere else in `SupportedAsset`).

chappjc:
Having changed from self-reporting traits in the WalletDefinition to determining traits based on interface type assertions, I now think I might go back.  Two reasons so far:
- both BTC wallets look like Rescanners, including the bitcoind RPC type, not just the integrated SPV wallet
- not all traits depend on the methods/interfaces.  For example, I would like a trait such as `WalletTraitSingleAddress` to signal that the "New Deposit Address" should not be shown (see https://github.com/decred/dcrdex/issues/1377) despite the ETH wallet's `Address()` method existing (but always returning the same address)

I'm considering giving `ExchangeWallet's an method `(*ExchangeWallet).ExtraTraits() WalletTrait` to report certain traits that should be merged with the ones determined from it's methods, but this nearly negates what we were trying to do by moving the traits into `core` and discerning them via interface checks.  Moving the `WalletTrait` definition back to `client/asset` isn't terrible, but I'm just no longer convinced that the method set is sufficient to determine every trait we'd want.

chappjc:
With commit https://github.com/decred/dcrdex/pull/1375/commits/92f0a6cb6a40e4aa481a59c85dd26ec5cf8c9d5c, traits are still in a field of `WalletState`, but the particular `WalletTrait` bit set for a particular wallet is determined on construction of the `asset.Wallet`, and it is provided to the consumer via the new `asset.(Wallet).Traits` method.  Notably, the determination of the `WalletTrait` on construction no longer involves interface assertion since as pointed out in my previous commit, wallet traits are more subtle than the methods provided.

With commit https://github.com/decred/dcrdex/pull/1375/commits/2be82db37ccee3858922b8d2ed2f1bd127d2f0ef, I'm demonstrating hiding the "New Deposit Address" button for an ETH wallet based on the new `WalletTraitSingleAddress` trait.

@buck54321 and @martonp Please see if the solution in this PR works for you for the general issue of signalling wallet traits beyond what the `asset.Wallet` implies, as well as the solution for ETH's deposit address issue https://github.com/decred/dcrdex/issues/1377, the planned Send/Sweep methods for ETH, and the Rescan method for the BTC SPV wallet (again noting that only the spvWallet should have the `WalletTraitRescanner` bit even though the `ExchangeWallet` must necessarily provide the `Rescan` method for both rpc and spv wallet `Type`s.

client/asset/interface.go
```diff
@@ -227,6 +263,23 @@ type Wallet interface {
 	RegFeeConfirmations(ctx context.Context, coinID dex.Bytes) (confs uint32, err error)
 }
 
+// Rescanner is a wallet implementation with rescan functionality.
+type Rescanner interface {
+	Rescan(ctx context.Context) error
+}
+
+// Sender is a wallet that can send funds to an address, as opposed to
```
martonp, line 271: I don't understand this comment.. isn't sending funds to an address the same as withdrawing?

client/asset/interface.go
```diff
@@ -10,6 +10,38 @@ import (
 	"decred.org/dcrdex/dex"
 )
 
+// WalletTrait is a bitset indicating various optional wallet features, such as
+// the presence of auxiliary methods like Rescan and Send.
+type WalletTrait uint64
+
+const (
+	WalletTraitRescanner     WalletTrait = 1 << iota // The Wallet is an asset.Rescanner.
+	WalletTraitSender                                // The Wallet is an asset.Sender.
```
martonp, line 19: Why do we need the sender trait? I would think every wallet would have the send functionality by default, and the additional information we would need to know is whether or not the wallet supports sweeping.

martonp (commented):
Looks good to me, just not sure about the sender trait.

client/asset/interface.go
chappjc: We can certainly make `Send` standard, part of `asset.Wallet`

client/asset/interface.go
chappjc: The difference is where fees come from.  Withdraw means the recipient gets less than the specified amount, but the dex wallet is reduced by exactly that amount).  Send means the recipient gets exactly the specified amount, but the dex wallet is reduced by a larger amount, determined after mining.  Which one to use depends entirely on the user's intended outcome.  Many wallet's support both.

A UI I imagine would be just a single Withdraw button, but the dialog/form for this would have a checkbox to indicate if the fees should be subtracted from the amount sent to the destination address or a larger amount should be consumed from the wallet to ensure the destination gets the exact amount.  If a wallet only supported Withdraw, it would be a disabled unchecked checkbox.  If it supported only Send, it would be a grayed/unchangable checked checkbox.  If it supported both, it would be user selectable.  To do this means there would have to be another trait to indicate if the wallet can do withdraw with fees subtracted as with BTC and DCR's

client/asset/btc/btc.go
```diff
@@ -730,12 +732,18 @@ func openSPVWallet(cfg *BTCCloneCFG) (*ExchangeWallet, error) {
...
 	return btc, nil
 }
 
 var _ asset.Wallet = (*ExchangeWallet)(nil)
 
+// Traits returns the traits for the type of wallet.
+func (btc *ExchangeWallet) Traits() asset.WalletTrait {
+	return btc.traits
+}
+
```
buck54321, line 746: I don't see why the wallet needs to know anything about the `asset.WalletTrait`. The wallet already knows what it's capabilities are.

buck54321 (commented):
How do you feel about this, @chappjc?

https://github.com/chappjc/dcrdex/compare/wallettraits...buck54321:wallettraits

chappjc:
@buck54321 In the first commit I had the basically the same `determineWalletTraits` function, but in general, I feel there will be other traits that aren't discernible by method presence.  Also, the `Rescanner` trait does not apply to both btc wallet types although that `ExchangeWallet` has a `Resan` method that only really works for the `spvWallet`.

buck54321:
Understood. We can roll with this. I wonder if the asset package could somehow validate the traits is knows against the implementation though?

chappjc:
@buck54321 On the RescanWallet work, trying out a different `asset.Wallet` impl for the spv `Type`: https://github.com/decred/dcrdex/pull/1374/commits/7b0f3e65ae380494a45a565f13aa0846ef4e040c
We could definitely make the interface assertions work with what we anticipate presently (the new address and rescan stuff).  Will keep thinking about how well that will work in the future.  Just concerned it could be hard to accommodate in other situations

client/asset/interface.go
martonp: I see, that makes sense.

chappjc:
Need to re-test this with https://github.com/decred/dcrdex/pull/1374, but it's now using the method assertion approach.  See comment https://github.com/decred/dcrdex/pull/1374#issuecomment-1016840825 if we decide to switch back for alt commits

client/webserver/site/src/js/wallets.js
```diff
@@ -345,6 +346,8 @@ export default class WalletsPage extends BasePage {
     await this.hideBox()
     page.depositName.textContent = asset.info.name
     page.depositAddress.textContent = wallet.address
+    if ((wallet.traits & traitNewAddresser) !== 0) Doc.hide(page.newDepAddrBttn)
```
chappjc, line 349: whoops botched this, pushing again shortly

chappjc:
image

image

client/core/wallet.go
```diff
@@ -177,7 +179,12 @@ func (w *xcWallet) refreshDepositAddress() (string, error) {
 			unbip(w.AssetID))
 	}
 
-	addr, err := w.Address()
+	na, is := w.Wallet.(asset.NewAddresser)
+	if !is {
+		return "", fmt.Errorf("wallet does not generate new addresses")
```
chappjc, line 184: Note that `(*Core).NewDepositAddress` will error for eth and any other asset that is not a `NewAddresser`, but the `(*Core).Wallets` method returns `WalletState` with an `Address` field for their current dep addr.

martonp (approved):
Looks great.

JoeGruffins (approved):

client/core/wallet.go
JoeGruffins: Nothing to format. I'm sorry can't resist.

buck54321 (approved):

PR #1376, opened 2021-12-31
client/asset/btc: work around PublishTransaction delays

chappjc:
Because of (a) what appears to be a neutrino bug, with a fix proposed in https://github.com/lightninglabs/neutrino/pull/240, and (b) that neutrino `wallet.PublishTransaction` is blocking while `inv` requests go to peers and `getdata` responses come back, this PR attempts to avoid an unnecessary delay with BTC SPV when broadcasting transactions.

In all my observations, `PublishTransaction` uses the **entire** `neutrino.Config.BroadcastTimeout`, which is nearly an intolerable delay for many operations.

With neutrino as the wallet's chain service, its `chainClient.SendRawTransaction` call is blocking for up to `neutrino.Config.BroadcastTimeout` while peers either respond to the `inv` request with a `getdata`, or time out. However, in virtually all cases, we just need to know that btcwallet was able to create and store the transaction record, and pass it to the chain service.  This is a yucky workaround to consider until https://github.com/lightninglabs/neutrino/pull/240 is accepted, maybe even complementary to it.

### DIFF
--- a/client/asset/btc/spv.go
+++ b/client/asset/btc/spv.go
@@ -61,6 +61,16 @@ const (
 	WalletTransactionNotFound = dex.ErrorKind("wallet transaction not found")
 	SpentStatusUnknown        = dex.ErrorKind("spend status not known")
 
+	// defaultBroadcastWait is long enough for btcwallet's PublishTransaction
+	// method to record the outgoing transaction and queue it for broadcasting.
+	// This rough duration is necessary since with neutrino as the wallet's
+	// chain service, its chainClient.SendRawTransaction call is blocking for up
+	// to neutrino.Config.BroadcastTimeout while peers either respond to the inv
+	// request with a getdata or time out. However, in virtually all cases, we
+	// just need to know that btcwallet was able to create and store the
+	// transaction record, and pass it to the chain service.
+	defaultBroadcastWait = 2 * time.Second
+
 	// see btcd/blockchain/validate.go
 	maxFutureBlockTime = 2 * time.Hour
 	neutrinoDBName     = "neutrino.db"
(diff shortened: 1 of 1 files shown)


chappjc:
The upstream fix is unlikely to happen this week, and even if it did, we would definitely not want broadcast to consistently block for even 4 seconds, which is appears to be most common even when all peers to respond with `getdata`.  Going to wrap this up and fix the tests, but I have been testing regularly in the past week.

In parallel, I'm looking to make certain requests async, with particular focus on not holding up preimage and match_proof requests.

client/asset/btc/spv.go
```diff
@@ -1160,12 +1195,17 @@ func (w *spvWallet) startWallet() error {
...
+		Database:      w.neutrinoDB,
+		ChainParams:   *w.chainParams,
+		PersistToDisk: true, // keep cfilter headers on disk for efficient rescanning
+		AddPeers:      addPeers,
+		ConnectPeers:  w.connectPeers,
+		// WARNING: PublishTransaction currently uses the entire duration
+		// because if an external bug, but even if the resolved, a typical
+		// inv/getdata round trip is ~4 seconds, so we set this so neutrino does
+		// not cancel queries too readily.
+		BroadcastTimeout: 6 * time.Second,
```
buck54321, line 1208: Looks like the default is 5? So we're actually giving neutrino **more** time.

buck54321 (approved):

client/asset/btc/spv.go
chappjc: Correct.  So 10 or even 30 would be fine with this approach, but I figured I would set it to something based on my empirical observations now that I've studies it more in practice.

chappjc:
We will definitely see a follow up to this in the future in the form of a neutrino update and/or a new btcwallet method that provides partially asynchronous PublishTransaction functionality without waiting for peer responses, only tx recording and queuing.

PR #1381, opened 2022-01-04
client token architecture proposal

buck54321:
This is not meant for merge as-is. This is built on top of an earlier commit of #1365 (itself on top of #1360), but not #1369. Those efforts should go in first, but hopefully informed by the discussion here. And after that, this can be broken up into smaller chunks.

This work provides a high-level client architecture for tokens. The approach taken for wallet initialization here is (almost) the one suggested by @chappjc in https://github.com/decred/dcrdex/issues/1348.

```
client/asset/eth:
Converts the ExchangeWallet to an AssetWallet that embeds a *baseWallet.
All wallets share the *baseWallet. Both ETH wallets and tokens are
*AssetWallets. This enables sharing of most logic. The relevant methods
of nodeClient are updated to take an asset ID. A new type of funding
coin is created to track both token value and fees reserved. Lots of
(as yet untested) updates to funding logic.

dex:
Token type is used to convey info about a token. To avoid lgpl
tag, addresses are stored as [20]byte. In the future this could
be generalized to a string, fmt.Stringer, or []byte.
DEXTT test token added to Bip map at index 60000. The token symbol
is denoted as dextt.eth, a convention I hope to establish and that
I am already making use of on the front end.

dex/networks/eth:
Package-level Tokens map[uint64]*dex.Token tracks eth tokens.

client/asset:
asset.Token embeds *dex.Token and adds a *WalletDefinition field.
RegisterToken function registers a *Token to the global tokens
map[uint64]*Token. RegisteredAsset gets a Tokens map[uint32]*Token
field too.

client/core:
SupportedAsset modified to add a Token *asset.Token field, which
will be populated instead of Info if the asset is a token. Nilness
is how consumers can differentiate asset type. A SupportedAsset
is generated for every asset whether base chain or token. The
asset.RegisteredAsset is translated to 1 + len(ra.Tokens)
SupportedAssets in (Core).assetMap.

client/webserver:
Live server updated to add a token.

ui:
Token wallets can be created directly with a single form, even if
the parent wallet doesn't exist yet. This pre-parent registration
is accomplished by merging the two WalletDefinition.ConfigOpts and
registering the two wallets sequentially, parent first. In some
important places, token asset symbols are denoted with their parent
asset's symbol in superscript. This should differentiate between a
token that might exist on two different chains.
```


chappjc:
Much appreciated, @buck54321.  These design decisions do need to be considered as a team.
I think with https://github.com/decred/dcrdex/pull/1360 and https://github.com/decred/dcrdex/pull/1365 merged this will be easier to reconcile with https://github.com/decred/dcrdex/pull/1369

chappjc:
Let's close

PR #1382, opened 2022-01-04
client/asset/{btc,dcr}: reset redeemScripts when changing min confs

chappjc:
In the UTXO selection methods for DCR and BTC, initially only UTXOs with >0 confs are tried, and if there are insufficient funds, it
attempts it again with unconfirmed utxos included (minus "untrusted" for BTC). However, on the second run, both methods neglect to reset the `redeemScripts` slices.  This fixes the issue.

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -1442,7 +1442,7 @@ func fund(utxos []*compositeUTXO, enough func(uint64, uint64) bool) (
 
 	tryUTXOs := func(minconf uint32) bool {
 		sum, size = 0, 0
-		coins, spents = nil, nil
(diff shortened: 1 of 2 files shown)


chappjc:
Any ideas if there's a bug around order funding that might lead to an error like `funding quantity <X sats> (N lots): not enough to cover requested funds (Y btc + tx fees). Z btc available`, where `Z` is ~50% larger than `Y`?  I have seen it in the past and my only guess is many small utxos, but there may yet be a bug.

buck54321 (approved):

JoeGruffins:
> Any ideas

no...
nitpick: The commit message is `{btc,dc}` dc missing the r

JoeGruffins (approved):

PR #1384, opened 2022-01-05
client/eth: Sign redeems.

JoeGruffins:
Add an AccountRedeemer interface that enables signing redeems for
    account based wallets when they will be a redeemer in a trade.

Preemptively adds the Account Redeemer interface in #1314 This does not attempt to solve the balance reserving problem, but allows signing redemptions so that we can use the eth wallet to redeem trades.


JoeGruffins:
I believe that #1314 also signs redeems. Will close this if that gets in first. This only allows for one possible redemption address, which may be a problem if account based wallets ever have more than one address. But for now, it means not much needs to change with the core.

JoeGruffins:
Closing in favor of #1314

PR #1392, opened 2022-01-07
eth: Funding coin encoding.

JoeGruffins:
depends on #1314 

This is what it would look like to encode the coin as a string. However, this changes the order ID, I believe, as the funding coins are all serialized and become part of the id
https://github.com/decred/dcrdex/blob/986c765be89df6e45cae5ef538675b2700c7f7a9/dex/order/order.go#L536-L541

I think maybe another field for `msgjson.Trade` might be best and leave the coins alone? Like `FundingAddress`.

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -10,6 +10,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/sha256"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"math/big"
@@ -471,6 +472,22 @@ func (c *coin) Value() uint64 {
 	return c.value
 }
 
+// fundingCoin is a coin that also implements asset.RecoveryCoin, enabling
+// a custom database record and special handling of the recovery ID for use in
+// later calls to FundingCoin.
+type fundingCoin struct {
+	*coin
+	recoveryID []byte
+}
+
+func (c *fundingCoin) RecoveryID() dex.Bytes {
+	return c.recoveryID[:]
+}
+
+func (c *fundingCoin) String() string {
+	return "{" + c.id.String() + ":" + hex.EncodeToString(c.recoveryID) + "}"
+}
+
 var _ asset.Coin = (*coin)(nil)
 
 // decodeFundingCoinID decodes a coin id into a coin object. This function ensures
@@ -493,11 +510,14 @@ func (eth *ExchangeWallet) decodeFundingCoinID(id []byte) (*coin, error) {
 	}, nil
 }
 
-func (eth *ExchangeWallet) createFundingCoin(amount uint64) *coin {
+func (eth *ExchangeWallet) createFundingCoin(amount uint64) *fundingCoin {
 	id := createFundingCoinID(eth.addr, amount)
-	return &coin{
-		id:    id.Encode(),
-		value: amount,
+	return &fundingCoin{
+		coin: &coin{
+			id:    []byte(eth.addr.String()),
+			value: amount,
+		},
+		recoveryID: id.Encode(),
 	}
 }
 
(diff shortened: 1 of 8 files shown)


martonp:
This looks good to me. Why is it an issue that the order ID will change due to this?

client/core/core.go
```diff
@@ -3910,19 +3930,24 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
 		return nil, 0, codedError(walletErr, fmt.Errorf("FundOrder error for %s, funding quantity %d (%d lots): %w",
 			wallets.fromAsset.Symbol, fundQty, lots, err))
 	}
+
 	coinIDs := make([]order.CoinID, 0, len(coins))
 	for i := range coins {
 		coinIDs = append(coinIDs, []byte(coins[i].ID()))
```
buck54321, line 3936: Here's where we need to insert the address. The `eth.coin` type will need to return the address here, but somehow the amounts for ID used for the database record. I recommend differentiating the two by adding a method to the `asset.Coin` interface, `RecoveryID() []byte`, which will be used for the database record and ultimately `FundingCoin`. For UTXO-based assets, the `RecoveryID()` will be identical to the `ID()`, but for eth, it will be different, so no upgrades are needed.

JoeGruffins: If I follow correctly, insert the address at index 0 and leave the rest of the coins as is?

chappjc: eth's `FundOrder` would only ever return a single coin right?

buck54321: @JoeGruffins what do you think of https://github.com/decred/dcrdex/compare/master...buck54321:coin-recovery-id?expand=1 ?

chappjc: That looks like a reasonable solution.

Although I will say that I am still open to keeping the encodings consistent and client code simpler by just sending the "dead weight" funding coin serialization that has the amount in it, changing the `FromAddress` method of `Trade` to `FromCoin`, and having the server's asset backend's `AssetBalancer` decode it internally.

chappjc: Eh, that suggestion affects `QuoteAccount` and `BaseAccount` and all the server and book tracking.  Too late to rework that.  NVM

chappjc: Current commits looking alright with RecoveryID.  You can pop out the redemption funding commit, but we'll also get the discoveraccount/testing seed PR in first.

client/asset/eth/eth.go
```diff
@@ -134,9 +134,9 @@ func (d *Driver) Open(cfg *asset.WalletConfig, logger dex.Logger, network dex.Ne
...
 // DecodeCoinID creates a human-readable representation of a coin ID for Ethereum.
 func (d *Driver) DecodeCoinID(coinID []byte) (string, error) {
-	id, err := dexeth.DecodeCoinID(coinID)
+	id, err := decodeFundingCoinID(coinID)
 	if err != nil {
-		return "", nil
+		return "<invalid coin>", err
 	}
 	return id.String(), nil
 }
```
JoeGruffins, line 142: Doing both in #1414

client/cmd/dexc/importlgpl.go
```diff
@@ -12,5 +12,5 @@ package main
 // import (
 // 	ETH is commented to prevent go-ethereum goroutines from launching via
 // 	it's init(). Uncomment this when ETH is ready:
-// 	_ "decred.org/dcrdex/client/asset/eth" // register eth asset
+//	_ "decred.org/dcrdex/client/asset/eth" // register eth asset
```
JoeGruffins, line 15: Unintended but leaving.

chappjc: Any time.

buck54321 (approved):

PR #1397, opened 2022-01-08
server/asset/eth: refactor coins

buck54321:
```
Separate swapCoin from redeemCoin. Switch (*rpcclient).swap method to
return a *dexeth.SwapState.
```

### DIFF
--- a/dex/networks/eth/params.go
+++ b/dex/networks/eth/params.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"decred.org/dcrdex/dex"
+	v0 "decred.org/dcrdex/dex/networks/eth/contracts/v0"
 	"github.com/ethereum/go-ethereum/common"
 )
 
@@ -120,7 +121,11 @@ func GweiToWei(v uint64) *big.Int {
 
 // GweiToWei converts *big.Int Wei to uint64 Gwei.
 func WeiToGwei(v *big.Int) uint64 {
-	return new(big.Int).Div(v, BigGweiFactor).Uint64()
+	vGwei := new(big.Int).Div(v, BigGweiFactor)
+	if vGwei.IsUint64() {
+		return vGwei.Uint64()
+	}
+	return 0
 }
 
 // SwapStep is the state of a swap and corresponds to values in the Solidity
@@ -248,4 +253,21 @@ func VersionedNetworkToken(assetID uint32, contractVer uint32, net dex.Network) 
 		return nil, common.Address{}, common.Address{}, fmt.Errorf("token %d version %d has no network %s token info", assetID, contractVer, net)
 	}
 	return token, addrs.Address, contractAddr, nil
+}
+
+// SwapStateFromV0 converts a v0.ETHSwapSwap to a *SwapState.
+func SwapStateFromV0(state *v0.ETHSwapSwap) *SwapState {
+	var blockTime int64
+	if state.RefundBlockTimestamp.IsInt64() {
+		blockTime = state.RefundBlockTimestamp.Int64()
+	}
+	return &SwapState{
+		BlockHeight: state.InitBlockNumber.Uint64(),
+		LockTime:    time.Unix(blockTime, 0),
+		Secret:      state.Secret,
+		Initiator:   state.Initiator,
+		Participant: state.Participant,
+		Value:       WeiToGwei(state.Value),
+		State:       SwapStep(state.State),
+	}
 }
--- a/dex/networks/eth/params.go
+++ b/dex/networks/eth/params.go
@@ -37,9 +37,6 @@ var (
 		},
 	}
 
-	// BigGweiFactor is the *big.Int form of the GweiFactor.
-	BigGweiFactor = big.NewInt(GweiFactor)
-
 	VersionedGases = map[uint32]*dex.Gases{
 		0: v0Gases,
 	}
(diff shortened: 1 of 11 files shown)


dex/networks/eth/params.go
```diff
@@ -130,7 +131,11 @@ func GweiToWei(v uint64) *big.Int {
 
 // GweiToWei converts *big.Int Wei to uint64 Gwei.
 func WeiToGwei(v *big.Int) uint64 {
-	return new(big.Int).Div(v, BigGweiFactor).Uint64()
+	vGwei := new(big.Int).Div(v, BigGweiFactor)
+	if vGwei.IsUint64() {
+		return vGwei.Uint64()
+	}
```
chappjc, line 137: We have a lot of redundancy with these helpers.

- server/asset/eth/common.go has `ToGwei` helpers that can be removed.  Also that one is bizarrely used in client (ExchangeWallet)
- dex/networks/eth/common.co has `ToGwei` and `ToWei`

Also, the docs on this function say `GweiToWei` instead of `WeiToGwei`.

server/asset/eth/coiner.go
```diff
@@ -176,109 +173,104 @@ func (backend *Backend) newSwapCoin(coinID []byte, contractData []byte, sct swap
...
+
+	bn, err := c.backend.node.blockNumber(ctx)
+	if err != nil {
+		return 0, fmt.Errorf("unable to fetch block number: %v", err)
+	}
+	return int64(bn - swap.BlockHeight + 1), nil
+}
 
-		bn, err := c.backend.node.blockNumber(c.backend.rpcCtx)
+func (c *redeemCoin) Confirmations(ctx context.Context) (int64, error) {
```
chappjc, line 219: Do we ever check redeem confs?  I'm drawing a blank on this for some reason

buck54321: I don't think so. I put added a `TODO` in the `(*Backend).newRedeemCoin` docs about it.

dex/networks/eth/params.go
JoeGruffins: Two things I was trying to avoid with the ToGwei.

- A package level BigGweiFactor. This is mutable.  Is this safe? For one a consumer could accidentally mutate it. `big` isn't the easiest package to use imo. Whenever you have `a := b.Add(c, d )` you might not expect b to change here, because we just want to add c and d, but b is set to those and returned as a. Also, could an asset that is not one of ours be added that maliciously changes this? I'm not sure if that is something we are worried about, but imo starting with a constant, or at least not exporting it, is safer.

As an example, add this test to the top of `eth_test.go` and another test will fail:
```
func TestMutate(t *testing.T) {
	fmt.Println("**************")
	fmt.Println(dexeth.BigGweiFactor)
	c := big.NewInt(2)
	d := big.NewInt(3)
	a := dexeth.BigGweiFactor.Add(c, d)
	fmt.Printf("%p %p\n", a, dexeth.BigGweiFactor)
	fmt.Println(dexeth.BigGweiFactor)
}
```

- A `big.Int` may not fit into a `uint64`. It could be too large or negative. It would be unexpected, but have untested consequences. Should we at least panic?

server/asset/eth/coiner.go
```diff
@@ -57,35 +47,77 @@ type swapCoin struct {
...
 		}
 		return nil, fmt.Errorf("unable to fetch transaction: %v", err)
 	}
 
-	txdata := tx.Data()
-	// Transactions that call contract functions must have extra data.
-	if len(txdata) == 0 {
-		return nil, errors.New("tx calling contract function has no extra data")
-	}
-
```
JoeGruffins, line 88: Was this check not accurate? I thought it was a good way to fail fast.

server/asset/eth/coiner.go
```diff
@@ -147,23 +154,13 @@ func (backend *Backend) newSwapCoin(coinID []byte, contractData []byte, sct swap
 		return nil, fmt.Errorf("unable to convert value: %v", err)
 	}
 
-	// For redemptions, the transaction should move no value.
-	if sct == sctRedeem && value != 0 {
-		return nil, fmt.Errorf("expected swapCoin value of zero for redeem but got: %d", value)
-	}
```
JoeGruffins, line 153: This check removed. Not needed? Any redeem tx with a value should fail to redeem because the solidity `redeem` function is not marked payable.

dex/networks/eth/params.go
buck54321: Good call on the `BigGweiFactor`. 

> * A `big.Int` may not fit into a `uint64`.

That's the purpose of both `ToGwei` and the new `IsUnit64` check in `WeiToGwei`. Which brings us back to @chappjc's original observation 

> We have a lot of redundancy with these helpers.

I have no problem having both `ToGwei` and `WeiToGwei`, the choice of which depends on if we want 1) to explicitly check for overflow/sign errors, or 2) just fall back to zero when not suitable for `uint64`. `WeiToGwei` clearly has some syntactic advantage, but whether it's appropriate needs to be considered in context.

I do think we should rename `ToGwei`.

server/asset/eth/coiner.go
buck54321: `baseCoin` is more generalized. The txdata length check is now performed (via `ParseInitiateData`) in `newSwapCoin` immediately after the `baseCoin` is created.

server/asset/eth/coiner.go
buck54321: Value is now being checked (to a tighter standard) in `newSwapCoin`.

server/asset/eth/coiner.go
```diff
@@ -176,109 +173,104 @@ func (backend *Backend) newSwapCoin(coinID []byte, contractData []byte, sct swap
...
+	// is the expected address and value was also as expected. Also
+	// not validating the locktime, as the swap is redeemed and
+	// locktime no longer relevant.
+	if swap.State == dexeth.SSRedeemed {
+		bn, err := c.backend.node.blockNumber(ctx)
 		if err != nil {
 			return 0, fmt.Errorf("unable to fetch block number: %v", err)
 		}
-		return int64(bn - swap.InitBlockNumber.Uint64()), nil
+		return int64(bn - swap.BlockHeight + 1), nil
```
chappjc, line 236: eek good catch

chappjc (approved):
LGTM, just a conflict to resolve.

JoeGruffins (approved):

server/asset/eth/coiner.go
JoeGruffins: Isn't that for inits, I mean redeems should be 0 value.

PR #1398, opened 2022-01-10
testing: Start app with seed.

### LINKED ISSUE #1393: testing: Init dexc from seed.
With external wallets we are able to send coins preemptively to simnet wallets that we later connect to and use during simnet testing.

The same can be done by using a predetermined seed with internal wallets. Specifically for eth, I propose we start the simnet testing app with a certain seed and send funds to whatever the wallet's address will be in the eth harness in order to always have funds for testing.

JoeGruffins:
rpcserver: Add discoveracct endpoint.

    Allows accounts that were previously paid for to be restored. This
    should be used as apposed to Register when restoring from seed.

    testing: Start app with seed. 

closes #1393


JoeGruffins:
Something must be missing when restoring an account. Looking into it...

JoeGruffins:
I believe we need to try and discover the account first. Adding that ability to the rpc server.

JoeGruffins:
Did not add exchanges as a result because that can be fetched separately with `getdexconfig`.

chappjc:
> I believe we need to try and discover the account first. Adding that ability to the rpc server.

For completeness, this makes sense to add.  For the testing harness, I don't believe `register` would have created errors since it is supposed to discover as well, but I want to verify that you didn't run into an issue there.

JoeGruffins:
> I don't believe register would have created errors

core's `Register` was giving an error on web login. Something about being unable to create an account bucket. I can look into it if that wasn't expected.

chappjc:
Not expected.  You should be able to use `Register` instead and it will do (remote) account discovery just like `Preregister`.  The difference between the two is just what happens after it does not find the account, where `Register` actually spends funds to register.

`Register` being unable to create an app bucket sounds like the account existed locally already in the dexc.db.

chappjc:
Well... after doing that it didn't show the DEX on the page.  Then I restarted and I can login but it wants me to setup a DEX server.  After I hard refreshed the page it showed though.

```
2022-01-12 19:08:19.822 [DBG] CORE: Authenticated connection to 127.0.0.1:17273, acct 71beeceddf39219fc2d5df285b1919e701d153c2dcfd91af628dc5fa884bd60e, 0 active orders, 0 active matches, score 0 (suspended = false)
```

JoeGruffins:
Oh, I clicked a lot of things and was able to get to trading. I was getting stuck at this
image

So, I guess it is an issue that the rpc server isn't setting some value web needs.

chappjc:
Hmm, yeah I think it changes state and the frontend doesn't know to do a fetchUser to get a completely updated state.  I had to hard refresh to get it to recognize things

JoeGruffins:
Well, I guess the rpc endpoint wasn't necessary, although it does let you login right after.

chappjc:
It's good to have for sure.  I'm curious you cannot login right after with Register.  We should get to the bottom of that.  If we do the dexcctl steps I described prior to opening the browser page, it should let you login.

chappjc:
@JoeGruffins pls try out https://github.com/decred/dcrdex/pull/1407 when you are able

chappjc (approved):
Looks good.  Would you please squash the first two rpcserver commits?  I'll merge this PR via rebase with two separate commits, one for the rpcserver and one for the harness.

JoeGruffins:
Merged the first two.

PR #1399, opened 2022-01-11
client/asset/eth: finish client wallet tokens

buck54321:
I've peeled a few other PRs off of this one, but I think I'll submit the rest as one. Lots of tests written or adapted and all are passing. Ready for a look.

```
Introduce tokenContractor interface to handle operations on the token
contract and token-bound swap contract.

Separate ExchangeWallet into baseWallet and AssetWallet.

Move contractor handling to AssetWallet. This simplifies the ethFetcher
and is a better division of labor, making future implementations of
EVM-family assets easier.

Implement detailed calculations of swap and approve gas, issuing
warnings when a configuration mismatch is detected.

dex/networks/eth,dex:
Simplify dex.Token and create dexeth.Token type to add gas and address
info. Move gas info to version-specific token data.

```

### DIFF
--- a/client/asset/eth/contractor.go
+++ b/client/asset/eth/contractor.go
@@ -50,7 +50,7 @@ type contractor interface {
 // contract.
 type tokenContractor interface {
 	contractor
-	tokenAddress(ctx context.Context) (common.Address, error)
+	tokenAddress() common.Address
 	balance(context.Context) (*big.Int, error)
 	allowance(context.Context) (*big.Int, error)
 	approve(*bind.TransactOpts, *big.Int) (*types.Transaction, error)
@@ -73,12 +73,6 @@ type contractV0 interface {
 	IsRefundable(opts *bind.CallOpts, secretHash [32]byte) (bool, error)
 }
 
-// tokenContractV0 extends the standard v0 swap contract with a method to
-// retrieve the token address bound to the contract.
-type tokenContractV0 interface {
-	TokenAddress(*bind.CallOpts) (common.Address, error)
-}
-
 // contractorV0 is the contractor for contract version 0.
 // Redeem and Refund methods of swapv0.ETHSwap already have suitable return types.
 type contractorV0 struct {
@@ -89,6 +83,7 @@ type contractorV0 struct {
 	acctAddr     common.Address
 	isToken      bool
 	evmify       func(uint64) *big.Int
+	atomize      func(*big.Int) uint64
 }
 
 var _ contractor = (*contractorV0)(nil)
@@ -113,6 +108,7 @@ func newV0Contractor(net dex.Network, acctAddr common.Address, cb bind.ContractB
 		contractAddr: contractAddr,
 		acctAddr:     acctAddr,
 		evmify:       dexeth.GweiToWei,
+		atomize:      dexeth.WeiToGwei,
 	}, nil
 }
 
@@ -191,13 +187,13 @@ func (c *contractorV0) swap(ctx context.Context, secretHash [32]byte) (*dexeth.S
 		Secret:      state.Secret,
 		Initiator:   state.Initiator,
 		Participant: state.Participant,
-		Value:       dexeth.WeiToGwei(state.Value),
+		Value:       c.atomize(state.Value),
 		State:       dexeth.SwapStep(state.State),
 	}, nil
 }
 
-// refund issues the refund command to the swap contract. Use isRedeemable first
-// to ensure the redemption will be accepted.
+// refund issues the refund command to the swap contract. Use isRefundable first
+// to ensure the refund will be accepted.
 func (c *contractorV0) refund(txOpts *bind.TransactOpts, secretHash [32]byte) (*types.Transaction, error) {
 	return c.contractV0.Refund(txOpts, secretHash)
 }
@@ -285,7 +281,7 @@ func (c *contractorV0) value(ctx context.Context, tx *types.Transaction) (in, ou
 		return v, 0, nil
 	}
 
-	return 0, c.outgoingValue(ctx, tx), nil
+	return 0, c.outgoingValue(tx), nil
 }
 
 // incomingValue calculates the value being redeemed for refunded in the tx.
@@ -313,7 +309,7 @@ func (c *contractorV0) incomingValue(ctx context.Context, tx *types.Transaction)
 }
 
 // outgoingValue calculates the value sent in swaps in the tx.
-func (c *contractorV0) outgoingValue(ctx context.Context, tx *types.Transaction) (swapped uint64) {
+func (c *contractorV0) outgoingValue(tx *types.Transaction) (swapped uint64) {
 	if inits, err := dexeth.ParseInitiateData(tx.Data(), 0); err == nil {
 		for _, init := range inits {
 			swapped += init.Value
@@ -326,9 +322,8 @@ func (c *contractorV0) outgoingValue(ctx context.Context, tx *types.Transaction)
 // methods, providing access to the methods of the token's ERC20 contract.
 type tokenContractorV0 struct {
 	*contractorV0
-	tokenContractor tokenContractV0
-	tokenAddr       common.Address
-	tokenContract   *erc20.IERC20
+	tokenAddr     common.Address
+	tokenContract *erc20.IERC20
 }
 
 var _ contractor = (*tokenContractorV0)(nil)
@@ -346,10 +341,19 @@ func newV0TokenContractor(net dex.Network, assetID uint32, acctAddr common.Addre
 		return nil, err
 	}
 
-	// tokenContract := bind.NewBoundContract(tokenAddr, *erc20.ERC20ABI, ec, ec, ec)
+	// tokenContract := bind.NewBoundContract(tokenAddr, *erc20.ERC20ABI, cb, cb, cb)
 	tokenContract, err := erc20.NewIERC20(tokenAddr, cb)
 	if err != nil {
 		return nil, err
+	}
+
+	if boundAddr, err := c.TokenAddress(&bind.CallOpts{
+		Pending: true,
+		Context: context.TODO(),
+	}); err != nil {
+		return nil, fmt.Errorf("error reading bound token address: %w", err)
+	} else if boundAddr != tokenAddr {
+		return nil, fmt.Errorf("wrong bound address. expected %s, got %s", tokenAddr, boundAddr)
 	}
 
 	return &tokenContractorV0{
@@ -361,17 +365,16 @@ func newV0TokenContractor(net dex.Network, assetID uint32, acctAddr common.Addre
 			acctAddr:     acctAddr,
 			isToken:      true,
 			evmify:       token.AtomicToEVM,
+			atomize:      token.EVMToAtomic,
 		},
-		tokenContractor: c,
-		tokenAddr:       tokenAddr,
-		tokenContract:   tokenContract,
+		tokenAddr:     tokenAddr,
+		tokenContract: tokenContract,
 	}, nil
 }
 
-// balance exposes the read-only balanceOf method of the swap contract.
+// balance exposes the read-only balanceOf method of the erc20 token contract.
 func (c *tokenContractorV0) balance(ctx context.Context) (*big.Int, error) {
 	callOpts := &bind.CallOpts{
-		Pending: true,
 		From:    c.acctAddr,
 		Context: ctx,
 	}
@@ -379,10 +382,9 @@ func (c *tokenContractorV0) balance(ctx context.Context) (*big.Int, error) {
 	return c.tokenContract.BalanceOf(callOpts, c.acctAddr)
 }
 
-// allowance exposes the read-only allowance method of the swap contract.
+// allowance exposes the read-only allowance method of the erc20 token contract.
 func (c *tokenContractorV0) allowance(ctx context.Context) (*big.Int, error) {
 	callOpts := &bind.CallOpts{
-		Pending: true,
 		From:    c.acctAddr,
 		Context: ctx,
 	}
@@ -395,7 +397,7 @@ func (c *tokenContractorV0) approve(txOpts *bind.TransactOpts, amount *big.Int) 
 	return c.tokenContract.Approve(txOpts, c.contractAddr, amount)
 }
 
-// transfer calls the transfer method of the token contract. Used for
+// transfer calls the transfer method of the erc20 token contract. Used for
 // sends or withdrawals.
 func (c *tokenContractorV0) transfer(txOpts *bind.TransactOpts, addr common.Address, amount *big.Int) (*types.Transaction, error) {
 	return c.tokenContract.Transfer(txOpts, addr, amount)
@@ -428,7 +430,9 @@ func (c *tokenContractorV0) estimateGas(ctx context.Context, method string, args
 }
 
 // value finds incoming or outgoing value for the tx to either the swap contract
-// or the token contract.
+// or the erc20 token contract. For the token contract, only transfer and
+// transferFrom are parsed. It is not an error if this tx is a call to another
+// method of the token contract, but no values will be parsed.
 func (c *tokenContractorV0) value(ctx context.Context, tx *types.Transaction) (in, out uint64, err error) {
 	to := *tx.To()
 	if to == c.contractAddr {
@@ -441,11 +445,11 @@ func (c *tokenContractorV0) value(ctx context.Context, tx *types.Transaction) (i
 	// Consider removing. We'll never be sending transferFrom transactions
 	// directly.
 	if sender, _, value, err := erc20.ParseTransferFromData(tx.Data()); err == nil && sender == c.acctAddr {
-		return 0, dexeth.WeiToGwei(value), nil
+		return 0, c.atomize(value), nil
 	}
 
 	if _, value, err := erc20.ParseTransferData(tx.Data()); err == nil {
-		return 0, dexeth.WeiToGwei(value), nil
+		return 0, c.atomize(value), nil
 	}
 
 	return 0, 0, nil
@@ -453,12 +457,8 @@ func (c *tokenContractorV0) value(ctx context.Context, tx *types.Transaction) (i
 
 // tokenAddress exposes the token_address immutable address of the token-bound
 // swap contract.
-func (c *tokenContractorV0) tokenAddress(ctx context.Context) (common.Address, error) {
-	return c.tokenContractor.TokenAddress(&bind.CallOpts{
-		Pending: true,
-		From:    c.acctAddr,
-		Context: ctx,
-	})
+func (c *tokenContractorV0) tokenAddress() common.Address {
+	return c.tokenAddr
 }
 
 var contractorConstructors = map[uint32]contractorConstructor{
--- a/client/asset/eth/contractor.go
+++ b/client/asset/eth/contractor.go
@@ -348,7 +348,6 @@ func newV0TokenContractor(net dex.Network, assetID uint32, acctAddr common.Addre
 	}
 
 	if boundAddr, err := c.TokenAddress(&bind.CallOpts{
-		Pending: true,
 		Context: context.TODO(),
 	}); err != nil {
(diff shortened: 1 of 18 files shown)


client/asset/eth/eth.go
```diff
@@ -507,68 +719,282 @@ func (eth *ExchangeWallet) createFundingCoin(amount uint64) *coin {
...
+			}
+			approveGas, err := w.approveGas(unlimitedAllowance, cfg)
+			if err != nil {
+				return nil, nil, fmt.Errorf("error estimating allowance gas: %w", err)
+			}
+			if (approveGas*cfg.MaxFeeRate)+ethToLock > ethBal.Available {
+				return nil, nil, fmt.Errorf("parent balance %d doesn't cover contract approval (%d) and tx fees (%d)",
+					ethBal.Available, approveGas*cfg.MaxFeeRate, ethToLock)
+			}
+			if _, err := w.approveToken(unlimitedAllowance, cfg.MaxFeeRate, cfg.Version); err != nil {
```
martonp, line 775: I haven't read the whole PR yet, but I'm assuming that on the first sell trade of the token, this is where the approval would happen. If this is the case then there would need to be additional handling on the server, because this trade cannot yet be booked. Having the unlimited approval as a manual step completely avoids this issue. I know you think this is a bad user experience, but anyone who has used an erc20 token is completely familiar with this step. It's just an extra click before you sell the token.

Also, I think these checks of the allowance going below 2^256/2 are not necessary. Other than maybe some absolute ridiculous token, there will be no tokens anywhere near 2^256 max supply, and even for that ridiculous token, the current user would have had to initiate swaps with the entire supply of the token before their allowance would run out, so I don't think we need to worry about that.

martonp (commented):
Initial Pass, looks good. I like the base/asset wallets and the contractors being moved out of the nodeclient. I think the biggest issue I found is regarding the unit conversion when initiating swaps.

client/asset/eth/eth.go
buck54321: > there would need to be additional handling on the server, because this trade cannot yet be booked

Can you expand on this? Doesn't the nonce order guarantee that the approval will be processed first.

From what I can tell, we're not checking the client's allowance server-side, and I didn't think we had any plans to do that. I was just going to relegate it to the client as part of configuration. Nothing we could do can prevent a client from sending an `approve(contractAddr, 0)` through after we accept their order anyway. 
The cost of approval is next to nothing, compartively, and correctly reported in the pre-order estimates.

If we did want to check allowance server-side, we would have to add latency handling though, which is what I think you were getting at.

To add manual approval, we would need to add a new UI dialog, new endpoint in webserver, new exported method from `*Core`, new exported method from `eth.ExchangeWallet`, and a new method or specialty interface in **interface.go**. And then we need to maintain all of that forever. And that's so we can force a user to click a button that they have no choice but to click in order to use the wallet that they just created. We also have to explain why we are forcing them to separately click this button to use the wallet. Why would they click "Create" if they were then going to say no to configuring the wallet for use? It makes for a bad UX, imo. Approval is not optional. It's part of trading with this asset. We don't need to pass the burden on to the user.

client/asset/eth/contractor.go
```diff
@@ -112,13 +140,14 @@ func (c *contractorV0) initiate(txOpts *bind.TransactOpts, contracts []*asset.Co
 			RefundTimestamp: big.NewInt(int64(contract.LockTime)),
 			SecretHash:      secretHash,
 			Participant:     common.HexToAddress(contract.Address),
-			Value:           new(big.Int).Mul(bigVal, big.NewInt(dexeth.GweiFactor)),
+			Value:           dexeth.GweiToWei(contract.Value),
```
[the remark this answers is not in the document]
buck54321: Ah. Damn.

client/asset/eth/eth.go
martonp: Hm I'm probably  biased because all the other apps I've used have had approval as a manual step.  I think you're right, we don't need to do any validation on the server side. 

I think the ideal way would be the approval happens automatically when the user is doing their first buy order.  In that case this code that does the approval should be in `ReserveN` instead of `FundOrder`.

chappjc:
I just noticed that dex/networks/eth/params_test.go is lacking an lgpl tag now.  Can add that too?

client/asset/eth/eth.go
```diff
@@ -593,101 +824,284 @@ func (c *coin) Value() uint64 {
...
+	// matters, just for estimates.
+	nSwap = oneSwap + uint64(n-1)*g.SwapAdd
+
+	// If a live estimate is greater than our estimate from configured values,
+	// use the live estimate with a warning.
+	if gasEst, err := w.estimateInitGas(w.ctx, n, dexCfg.Version); err != nil {
+		w.log.Errorf("(%d) error estimating swap gas: %v", w.assetID, err)
+		return 0, 0, err
+	} else if gasEst > nSwap {
+		w.log.Warnf("Swap gas estimate %d is greater than the server's configured value %d. Using live estimate + 10%.")
```
chappjc, line 972: The %d verbs don't have matching args.

client/asset/eth/eth.go
```diff
@@ -593,101 +824,284 @@ func (c *coin) Value() uint64 {
...
+
+	// Warn of mismatch in gas estimates.
+	if g.Swap > oneSwap {
+		w.log.Warnf("Server has a lower configured swap gas value, which is odd")
+		oneSwap = g.Swap
+	}
+
+	// Not gonna worry about accuracy for nSwap. It's never used where it
+	// matters, just for estimates.
+	nSwap = oneSwap + uint64(n-1)*g.SwapAdd
```
chappjc, line 964: What do you mean by worry about accuracy?  As in just using the hard coded and padded estimates for this contract version?

chappjc (changes requested):
Looking pretty slick so far.  Still going through it.

client/asset/eth/eth.go
```diff
@@ -593,101 +824,284 @@ func (c *coin) Value() uint64 {
...
 		}
-		if addr != eth.addr {
-			return fmt.Errorf("coin is not funded by this wallet. coin address %s != our address %s", addr, eth.addr)
+		if addr != w.addr {
+			return fmt.Errorf("coin is not funded by this wallet. coin address %s != our address %s", addr, w.addr)
 		}
 	}
-
-	eth.lockedFunds.mtx.Lock()
-	defer eth.lockedFunds.mtx.Unlock()
```
JoeGruffins, line 665: Still need to hold the lock to do `w.unlockFunds` I think.

client/asset/eth/eth.go
```diff
@@ -593,101 +824,284 @@ func (c *coin) Value() uint64 {
...
+	w.lockedFunds.mtx.Lock()
+	defer w.lockedFunds.mtx.Unlock()
+	// err := w.lockFunds(initiationFunds, initiationReserve)
+	// if err != nil {
+	// 	return nil, nil, fmt.Errorf("error estimating swap gas: %v", err)
+	// }
 
-	eth.lockedFunds.mtx.Lock()
-	defer eth.lockedFunds.mtx.Unlock()
-	err := eth.lockFunds(initiationFunds, initiationReserve)
```
JoeGruffins, line 638: Still need to hold the lock to do both locking and unlocking I think.

client/asset/eth/eth.go
```diff
@@ -1207,56 +1772,108 @@ func (eth *ExchangeWallet) Refund(_, contract dex.Bytes, feeSuggestion uint64) (
...
-func (eth *ExchangeWallet) EstimateRegistrationTxFee(feeRate uint64) uint64 {
-	return feeRate * defaultSendGasLimit
+func (w *AssetWallet) EstimateRegistrationTxFee(feeRate uint64) uint64 {
+	if w.assetID == BipID {
+		return feeRate * defaultSendGasLimit
+	}
+	g := w.gases(contractVersionNewest)
+	if g == nil {
+		w.log.Errorf("no gas table")
+		return math.MaxUint64
```
JoeGruffins, line 1868: Expensive.

JoeGruffins (commented):
Looks good. I hope that you can add a `TokenWallet` similar to #1404 because I think that would clean up these methods a lot and make them easier to reason about without all the `w.assetID == BipID` checks.

client/asset/eth/eth.go
```diff
@@ -214,17 +221,15 @@ type ethFetcher interface {
...
+// Check that AssetWallet satisfies the asset.Wallet interface.
+var _ asset.Wallet = (*AssetWallet)(nil)
+var _ asset.AccountLocker = (*AssetWallet)(nil)
+var _ asset.TokenMaster = (*AssetWallet)(nil)
 
-// ExchangeWallet is a wallet backend for Ethereum. The backend is how the DEX
-// client app communicates with the Ethereum blockchain and wallet. ExchangeWallet
-// satisfies the dex.Wallet interface.
-type ExchangeWallet struct {
+type baseWallet struct {
```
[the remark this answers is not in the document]
buck54321: There is only one `baseWallet` shared by all `AssetWallet`s. 

https://github.com/decred/dcrdex/blob/64b5c1e55e9a75937c4bd3d4f523bda17fff3813/client/asset/eth/eth.go#L512

I am going to look at `ETHWallet` and `TokenWallet` structs as per @JoeGruffins request though.

client/asset/eth/eth.go
buck54321: Whoops

client/asset/eth/eth.go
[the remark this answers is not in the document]
chappjc: > There is only one `baseWallet` shared by all `AssetWallet`s.

Yah, this was really what I was commenting about.  Made me uncertain on why there where two separate types.  Will see what happens with the next iteration.

client/asset/eth/contractor.go
```diff
@@ -259,6 +312,159 @@ func (c *contractorV0) incomingValue(ctx context.Context, tx *types.Transaction)
...
+		return c.contractorV0.value(ctx, tx)
+	}
+	if to != c.tokenAddr {
+		return 0, 0, nil
+	}
+
+	// Consider removing. We'll never be sending transferFrom transactions
+	// directly.
+	if sender, _, value, err := erc20.ParseTransferFromData(tx.Data()); err == nil && sender == c.acctAddr {
+		return 0, dexeth.WeiToGwei(value), nil
```
martonp, line 444: Use `evmify` instead of `WeiToGwei`

martonp (commented):
Started reviewing, will continue later.

client/asset/eth/eth.go
```diff
@@ -1022,78 +1067,102 @@ func (w *AssetWallet) approvalGas(newGas *big.Int, cfg *dex.Asset) (uint64, erro
...
+		fees += c.fees
 	}
 	if fees > 0 {
-		if w.token == nil {
-			return fmt.Errorf("tokenFundingCoin returned for non-token asset")
-		}
-		w.token.parent.unlockFunds(fees, initiationReserve)
+		ethWallet := w.parent
+		ethWallet.lockedFunds.mtx.Lock()
+		w.parent.unlockFunds(fees, initiationReserve)
```
chappjc, line 1117: Why not use `w.parent` when accessing the mutex too, and eliminate the `ethWallet` definition?

client/asset/interface.go
```diff
@@ -486,10 +486,10 @@ type Swaps struct {
 	// LockChange can be set to true if the change should be locked for
 	// subsequent matches.
 	LockChange bool
-	// AssetConfig is the swap protocol version, which may indicate a specific
-	// contract or form of contract. NOTE: Depending on the asset, batch swaps
-	// may force each Contract to use the same version, but conceptually this
-	// should perhaps be in Contract with SecretHash.
+	// AssetConfig contains the asset version and fee configuration for the DEX.
+	// NOTE: Only on Config field is supported, so only orders from the same
```
chappjc, line 490: Only *one*

chappjc (approved):
I really like the latest change of splitting `ETHWallet` and `TokenWallet`.  In general it makes things much easier to follow and less error prone, even if at the expense of a little extra code and boiler plate.  Worth it.

client/asset/eth/eth.go
```diff
@@ -1530,73 +2482,177 @@ func (eth *ExchangeWallet) balanceWithTxPool() (*Balance, error) {
...
+func (w *AssetWallet) estimateApproveGas(newGas *big.Int) (gas uint64, err error) {
+	return gas, w.withTokenContractor(w.assetID, contractVersionNewest, func(c tokenContractor) error {
+		gas, err = c.estimateApproveGas(w.ctx, newGas)
+		return err
+	})
+}
+
+// estimateTransferGas estimates the gas needed for a token transfer call to an
+// ERC20 contract.
+func (w *AssetWallet) estimateTransferGas(val uint64) (gas uint64, err error) {
```
JoeGruffins, line 2638: These three look like they should be off of `TokenWallet`.

client/asset/eth/eth.go
```diff
@@ -1605,29 +2661,204 @@ func (eth *ExchangeWallet) redeem(ctx context.Context, redemptions []*asset.Rede
...
 	})
 }
+
+// getGasEstimate is used to get a gas table for an asset's contract(s). The
+// provided gases, g, should be generous estimates of what the gas might be.
+// Errors are thrown if the provided estimates are too small by more than a
+// factor of 2. The account should already have a trading balance of at least
+// maxSwaps gwei (token or eth), and sufficient eth balance to cover the
+// requisite tx fees.
+func GetGasEstimates(ctx context.Context, cl *nodeClient, c contractor, maxSwaps int, g *dexeth.Gases, waitForMined func()) error {
```
JoeGruffins, line 2703: Comment is unexported, but this is exported. Will this be part of another tool? Does it belong in testing? Uses waitForMined() which is part of the harness tests. Not sure if you can use functions from test files in the build.

JoeGruffins (commented):
Just a few more comments.

The eth harness tests are failing with `participant newV0TokenContractor error: error reading bound token address: no contract code at given address`

client/asset/eth/eth.go
```diff
@@ -807,17 +1386,31 @@ func (eth *ExchangeWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin
 // redemption. All redemptions must be for the same contract version because the
 // current API requires a single transaction reported (asset.Coin output), but
 // conceptually a batch of redeems could be processed for any number of
-// different contract addresses with multiple transactions.
-func (eth *ExchangeWallet) Redeem(form *asset.RedeemForm) ([]dex.Bytes, asset.Coin, uint64, error) {
+// different contract addresses with multiple transactions. (buck: what would
+// the difference from calling Redeem repeatedly?)
```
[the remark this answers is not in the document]
chappjc: I think @buck54321 was posing that question.  I suppose the answer would just be convenience for core so it wouldn't have to figure out which ones to separate into different calls, if that's even something it should be aware of at all which I don't think it should.

client/asset/eth/eth.go
```diff
@@ -1207,56 +2003,139 @@ func (eth *ExchangeWallet) Refund(_, contract dex.Bytes, feeSuggestion uint64) (
...
 }
 
+// TODO: Lock, Unlock, and Locked should probably be part of an optional
+// asset.Authenticator interface that isn't implemented by token wallets.
+// This is easy to accomplish here, but would require substantial updates to
+// client/core.
+// The addition of an ETHWallet type that implements asset.Authenticator would
+// also facilitate the appropriate compartmentalization of our asset.TokenMaster
+// methods, which token wallets also won't need.
+
```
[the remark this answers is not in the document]
chappjc: I'm not sure I see much reason to allow locking an eth account at all.  As far as I'm concerned, they can be unlocked permanently when user logs in and the native eth wallet is loaded.  The app pass is still required to do various things, and always for withrdaw.

Alternatively, for a gentle solution to this `TODO` comment, how about overriding the `Lock` method with a dummy `func (*TokenWallet).Lock() error { return nil }`.

client/asset/eth/eth.go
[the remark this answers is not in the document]
martonp: It's because there's only one instance of `baseWallet` that holds the reference to the node. Then there are multiple  instances of `AssetWallet` each of which hold a reference to the one `baseWallet`. It's not that each `AssetWallet` is holding a reference to it's own `baseWallet`.

martonp (approved):

client/asset/eth/eth.go
```diff
@@ -807,17 +1386,31 @@ func (eth *ExchangeWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin
...
+	return w.AssetWallet.Redeem(form, nil)
+}
+
+// Redeem sends the redemption transaction, which may contain more than one
+// redemption.
+func (w *TokenWallet) Redeem(form *asset.RedeemForm) ([]dex.Bytes, asset.Coin, uint64, error) {
+	return w.AssetWallet.Redeem(form, w.parent)
+}
+
+func (w *AssetWallet) Redeem(form *asset.RedeemForm, feeWallet *AssetWallet) ([]dex.Bytes, asset.Coin, uint64, error) {
```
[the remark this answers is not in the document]
buck54321: There's already an unexported `redeem` method on `AssetWallet`.

client/asset/eth/eth.go
```diff
@@ -891,73 +1528,233 @@ func recoverPubkey(msgHash, sig []byte) ([]byte, error) {
...
-	eth.unlockFunds(reserves, redemptionReserve)
+func (w *TokenWallet) UnlockRedemptionReserves(reserves uint64) {
+	unlockRedemptionReserves(w.parent, reserves)
+}
+
+func unlockRedemptionReserves(w *AssetWallet, reserves uint64) {
+	w.lockedFunds.mtx.Lock()
+	defer w.lockedFunds.mtx.Unlock()
+	w.unlockFunds(reserves, redemptionReserve)
+}
```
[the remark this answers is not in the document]
buck54321: `AssetWallet` doesn't have access to the token parent.

client/asset/eth/eth.go
```diff
@@ -1441,41 +2378,48 @@ func (eth *ExchangeWallet) checkForNewBlocks() {
...
 	}
 }
 
-// withContractor runs the provided function with the versioned contractor.
-func (eth *ExchangeWallet) withContractor(ver uint32, f func(contractor) error) error {
-	contractor, found := eth.contractors[ver]
-	if !found {
-		return fmt.Errorf("no version %d contractor", ver)
+func (w *AssetWallet) balanceWithTxPool() (*Balance, error) {
+	isETH := w.assetID == BipID
```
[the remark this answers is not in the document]
buck54321: There was too much shared here. I didn't want to duplicate all of it.

client/asset/eth/eth.go
buck54321: `waitForMined` is passed in as an argument for now. This will be used to create the gas tables once a contract is launched on a network. I haven't written a utility for it yet, but I'm planning on it.

client/asset/eth/eth.go
buck54321: Damn. I can't do this because of some other shared code. Maybe it would be worth splitting more methods up, but I'm not feeling it. Can follow up.

client/asset/eth/eth.go
```diff
@@ -321,77 +375,195 @@ func NewWallet(assetCFG *asset.WalletConfig, logger dex.Logger, net dex.Network)
...
+		return nil, fmt.Errorf("token %d not found", tokenID)
+	}
+
+	cfg, err := parseTokenWalletConfig(settings)
+	if err != nil {
+		return nil, err
+	}
+
+	aw := &AssetWallet{
+		baseWallet:         w.baseWallet,
```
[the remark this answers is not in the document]
chappjc: I was being seriously dense about this too.

chappjc:
Thanks for the hard work guys.  I think we can move forward.  @JoeGruffins If there's anything else that pops out or looks to have slipped by, there's clearly gonna be followup.

PR #1403, opened 2022-01-12
client/asset/dcr: suppress harmless coin-not-found errors with SPV

chappjc:
When using dcrwallet in SPV mode and trading with another party (not yourself), there are certain errors and warnings logged that are not of cause for concern.

Consider the following logs of a Buy DCR order where the counterparty is another wallet and thus the dcr ExchangeWallet takes a different path to getting confirmations and spend status of their swap transaction:

> 2022-01-11 15:11:15.432 [DBG] CORE[dcr]: Contract output c4aeefdaa9dae7ab66e909d7107480287d638501bca8550fc1d53f1107f055bd:0 NOT yet found, will attempt finding it with block filters.
> 2022-01-11 15:11:15.432 [DBG] CORE[dcr]: Searching for tx c4aeefdaa9dae7ab66e909d7107480287d638501bca8550fc1d53f1107f055bd in blocks between block 852084 (000000004e22a09d0970a087ac07865d1ddd3f696940585e1039b6dc5a823dc7) to the block just before 2022-01-11 15:11:00 +0000 UTC.
> 2022-01-11 15:11:15.432 [DBG] CORE[dcr]: Tx c4aeefdaa9dae7ab66e909d7107480287d638501bca8550fc1d53f1107f055bd NOT found in blocks 852084 (000000004e22a09d0970a087ac07865d1ddd3f696940585e1039b6dc5a823dc7) to 852083 (00000000df11307a13f705c69368fa170f7fd90c6f5b0f4f4f6c363d044788b5).
> 2022-01-11 15:11:15.432 [ERR] CORE: Failed to get confirmations of the counter-party's swap c4aeefdaa9dae7ab66e909d7107480287d638501bca8550fc1d53f1107f055bd:0 (dcr) for match b620b29c634a491766dfb30d4d8b94d5b3e72cedfffe9d6f4c2f9f120c2f3562, order bf036c088a79c864fd397e0b63f9b24d2003005f828e7b112d1c8f634a49f214: coin not found

In the above, the ERR is logged because `asset.CoinNotFoundError` is returned from `SwapConfirmations` even though it is expected not to find the counterparty swap with an SPV wallet.  The BTC spvWallet does not return this error for this reason.
Also, the blocks scanned are reversed from a logical <low> to <high> order.

After repeating the error above until the counterparty contract is mined, it then finds it with a cfilter scan:

> 2022-01-11 15:13:30.431 [DBG] CORE[dcr]: tip change: 852084 (000000004e22a09d0970a087ac07865d1ddd3f696940585e1039b6dc5a823dc7) => 852085 (00000000509750fa562844e724b86af4e14f3880bbf804fbb7db7df3d01cf254)
> 2022-01-11 15:13:30.431 [TRC] CORE: processing tip change for dcr
> 2022-01-11 15:13:30.431 [TRC] CORE: Checking confirmations on COUNTERPARTY swap txn c4aeefdaa9dae7ab66e909d7107480287d638501bca8550fc1d53f1107f055bd:0 (dcr)...
> 2022-01-11 15:13:30.431 [DBG] CORE[dcr]: Contract output c4aeefdaa9dae7ab66e909d7107480287d638501bca8550fc1d53f1107f055bd:0 NOT yet found, will attempt finding it with block filters.
> 2022-01-11 15:13:30.432 [DBG] CORE[dcr]: Searching for tx c4aeefdaa9dae7ab66e909d7107480287d638501bca8550fc1d53f1107f055bd in blocks 852085 (00000000509750fa562844e724b86af4e14f3880bbf804fbb7db7df3d01cf254) to 852084 (000000004e22a09d0970a087ac07865d1ddd3f696940585e1039b6dc5a823dc7).
> 2022-01-11 15:13:30.439 [DBG] CORE[dcr]: Found mined tx c4aeefdaa9dae7ab66e909d7107480287d638501bca8550fc1d53f1107f055bd in block 852085 (00000000509750fa562844e724b86af4e14f3880bbf804fbb7db7df3d01cf254).
> 2022-01-11 15:13:30.439 [WRN] CORE[dcr]: Attempted to look for output spender in block 852086 but current tip is 852085

Here the WRN is unnecessary since it is expected to have already scanned the best block due to an optimization in `findTxInBlock` because it has the full block handy.

This addresses these cases with the following changes.

Do not return `asset.CoinNotFoundError` from `SwapConfirmations` with an SPV wallet since this is normal for mempool transactions in SPV.

For a non-SPV wallet, if the swap appears spent but it cannot be located in a block with a cfilters scan, this will return `asset.CoinNotFoundError`. For SPV wallets, it is not an error if the transaction cannot be located SPV wallets cannot see non-wallet transactions until they are mined.

Fix a few logs where arguments were ordered strangely.

Also silence a warning that happens each time a transaction is found in a block while looking for a spending transaction immediately afterward.

### DIFF
--- a/client/asset/btc/spv.go
+++ b/client/asset/btc/spv.go
@@ -957,6 +957,8 @@ func (w *spvWallet) swapConfirmations(txHash *chainhash.Hash, vout uint32, pkScr
 		if assumedMempool {
 			w.log.Tracef("swapConfirmations - scanFilters did not find %v:%d, assuming in mempool.",
 				txHash, vout)
+			// NOT asset.CoinNotFoundError since this is normal for mempool
+			// transactions with an SPV wallet.
 			return 0, false, nil
 		}
 		return 0, false, fmt.Errorf("output %s:%v not found with search parameters startTime = %s, pkScript = %x",
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -1862,14 +1862,14 @@ func (dcr *ExchangeWallet) lookupTxOutput(ctx context.Context, txHash *chainhash
 	output, err := dcr.wallet.UnspentOutput(ctx, txHash, vout, wire.TxTreeUnknown)
 	if err == nil {
 		return output.TxOut, output.Confirmations, false, nil
-	} else if err != asset.CoinNotFoundError {
+	} else if !errors.Is(err, asset.CoinNotFoundError) {
 		return nil, 0, false, err
 	}
 
 	// Check wallet transactions.
 	tx, err := dcr.wallet.GetTransaction(ctx, txHash)
 	if err != nil {
-		return nil, 0, false, err
+		return nil, 0, false, err // asset.CoinNotFoundError if not found
 	}
 	msgTx, err := msgTxFromHex(tx.Hex)
 	if err != nil {
@@ -2538,8 +2538,10 @@ func (dcr *ExchangeWallet) ValidateSecret(secret, secretHash []byte) bool {
 // the specified swap. The contract and matchTime are provided so that wallets
 // may search for the coin using light filters.
 //
-// If the swap was not funded by this wallet, and it is already spent, this
-// method may return asset.CoinNotFoundError. Compare dcr.externalTxOut.
+// For a non-SPV wallet, if the swap appears spent but it cannot be located in a
+// block with a cfilters scan, this will return asset.CoinNotFoundError. For SPV
+// wallets, it is not an error if the transaction cannot be located SPV wallets
+// cannot see non-wallet transactions until they are mined.
 //
 // If the coin is located, but recognized as spent, no error is returned.
 func (dcr *ExchangeWallet) SwapConfirmations(ctx context.Context, coinID, contract dex.Bytes, matchTime time.Time) (confs uint32, spent bool, err error) {
@@ -2552,7 +2554,7 @@ func (dcr *ExchangeWallet) SwapConfirmations(ctx context.Context, coinID, contra
 	_, confs, spent, err = dcr.lookupTxOutput(ctx, txHash, vout)
 	if err == nil {
 		return confs, spent, nil
-	} else if err != asset.CoinNotFoundError {
+	} else if !errors.Is(err, asset.CoinNotFoundError) {
 		return 0, false, err
 	}
 
@@ -2563,9 +2565,17 @@ func (dcr *ExchangeWallet) SwapConfirmations(ctx context.Context, coinID, contra
 	}
 	_, p2shScript := scriptAddr.PaymentScript()
 
-	// Find the contract and it's spend status using block filters.
+	// Find the contract and its spend status using block filters.
 	dcr.log.Debugf("Contract output %s:%d NOT yet found, will attempt finding it with block filters.", txHash, vout)
-	return dcr.lookupTxOutWithBlockFilters(ctx, newOutPoint(txHash, vout), p2shScript, matchTime)
+	confs, spent, err = dcr.lookupTxOutWithBlockFilters(ctx, newOutPoint(txHash, vout), p2shScript, matchTime)
+	// Don't trouble the caller if we're using an SPV wallet and the transaction
+	// cannot be located.
+	if errors.Is(err, asset.CoinNotFoundError) && dcr.wallet.SpvMode() {
+		dcr.log.Debugf("SwapConfirmations - cfilters scan did not find %v:%d. "+
+			"Assuming in mempool.", txHash, vout)
+		err = nil
+	}
+	return confs, spent, err
 }
 
 // RegFeeConfirmations gets the number of confirmations for the specified
@@ -3225,7 +3235,7 @@ func (dcr *ExchangeWallet) isMainchainBlock(ctx context.Context, block *block) (
 	}
 	nextBlockHash, err := chainhash.NewHashFromStr(blockHeader.NextHash)
 	if err != nil {
-		return false, fmt.Errorf("block %s has invalid nexthash value %s: %v",
+		return false, fmt.Errorf("block %s has invalid nexthash value %s: %w",
 			block.hash, blockHeader.NextHash, err)
 	}
 	nextBlockHeader, err := dcr.wallet.GetBlockHeaderVerbose(ctx, nextBlockHash)
--- a/client/asset/dcr/externaltx.go
+++ b/client/asset/dcr/externaltx.go
@@ -54,12 +54,12 @@ func (dcr *ExchangeWallet) lookupTxOutWithBlockFilters(ctx context.Context, op o
 
 	output, outputBlock, err := dcr.externalTxOutput(ctx, op, pkScript, earliestTxTime)
 	if err != nil {
-		return 0, false, err
+		return 0, false, err // may be asset.CoinNotFoundError
 	}
 
 	spent, err := dcr.isOutputSpent(ctx, output)
 	if err != nil {
-		return 0, false, fmt.Errorf("error checking if output %s is spent: %v", op, err)
+		return 0, false, fmt.Errorf("error checking if output %s is spent: %w", op, err)
 	}
 
 	// Get the current tip height to calculate confirmations.
@@ -82,7 +82,7 @@ func (dcr *ExchangeWallet) externalTxOutput(ctx context.Context, op outPoint, pk
 	tx := dcr.externalTxCache[op.txHash]
 	if tx == nil {
 		tx = &externalTx{hash: &op.txHash}
-		dcr.externalTxCache[op.txHash] = tx
+		dcr.externalTxCache[op.txHash] = tx // never deleted (TODO)
 	}
 	dcr.externalTxMtx.Unlock()
 
@@ -98,14 +98,14 @@ func (dcr *ExchangeWallet) externalTxOutput(ctx context.Context, op outPoint, pk
 	// First check if the tx block is cached.
 	txBlock, err := dcr.txBlockFromCache(ctx, tx)
 	if err != nil {
-		return nil, nil, fmt.Errorf("error checking if tx %s is known to be mined: %v", tx.hash, err)
+		return nil, nil, fmt.Errorf("error checking if tx %s is known to be mined: %w", tx.hash, err)
 	}
 
 	// Scan block filters to find the tx block if it is yet unknown.
 	if txBlock == nil {
 		txBlock, err = dcr.scanFiltersForTxBlock(ctx, tx, [][]byte{pkScript}, earliestTxTime)
 		if err != nil {
-			return nil, nil, fmt.Errorf("error checking if tx %s is mined: %v", tx.hash, err)
+			return nil, nil, fmt.Errorf("error checking if tx %s is mined: %w", tx.hash, err)
 		}
 		if txBlock == nil {
 			return nil, nil, asset.CoinNotFoundError
@@ -171,11 +171,11 @@ func (dcr *ExchangeWallet) scanFiltersForTxBlock(ctx context.Context, tx *extern
 	// to block just before earliestTxTime.
 	currentTip := dcr.cachedBestBlock()
 	if lastScannedBlock == nil {
-		dcr.log.Debugf("Searching for tx %s in blocks between block %d (%s) to the block just before %s.",
+		dcr.log.Debugf("Searching for tx %s in blocks between best block %d (%s) and the block just before %s.",
 			tx.hash, currentTip.height, currentTip.hash, earliestTxTime)
 	} else if lastScannedBlock.height < currentTip.height {
 		dcr.log.Debugf("Searching for tx %s in blocks %d (%s) to %d (%s).", tx.hash,
-			currentTip.height, currentTip.hash, lastScannedBlock.height, lastScannedBlock.hash)
+			lastScannedBlock.height, lastScannedBlock.hash, currentTip.height, currentTip.hash)
 	} else {
 		if lastScannedBlock.height > currentTip.height {
 			dcr.log.Warnf("Previous cfilters look up for tx %s stopped at block %d but current tip is %d?",
@@ -192,7 +192,7 @@ func (dcr *ExchangeWallet) scanFiltersForTxBlock(ctx context.Context, tx *extern
 	scanCompletedWithoutResults := func() (*block, error) {
 		tx.lastScannedBlock = currentTip.hash
 		dcr.log.Debugf("Tx %s NOT found in blocks %d (%s) to %d (%s).", tx.hash,
-			currentTip.height, currentTip.hash, iHeight, iHash)
+			iHeight, iHash, currentTip.height, currentTip.hash)
 		return nil, nil
 	}
 
@@ -272,8 +272,8 @@ func (dcr *ExchangeWallet) findTxInBlock(ctx context.Context, txHash *chainhash.
 		return nil, nil, fmt.Errorf("invalid hex for tx %s: %v", txHash, err)
 	}
 
-	// We have the txs in this block, check if any them spends an output
-	// from the original tx.
+	// We have the txs in this block, check if any of them spends an output from
+	// the original tx.
 	outputSpenders := make([]*outputSpenderFinder, len(msgTx.TxOut))
 	for i, txOut := range msgTx.TxOut {
 		outputSpenders[i] = &outputSpenderFinder{
@@ -286,7 +286,7 @@ func (dcr *ExchangeWallet) findTxInBlock(ctx context.Context, txHash *chainhash.
 	for t := range blockTxs {
 		blkTx := &blockTxs[t]
 		if blkTx.Txid == txHash.String() {
-			continue // oriignal tx, ignore
+			continue // original tx, ignore
 		}
 		for i := range blkTx.Vin {
 			input := &blkTx.Vin[i]
@@ -340,7 +340,12 @@ func (dcr *ExchangeWallet) isOutputSpent(ctx context.Context, output *outputSpen
 
 	bestBlock := dcr.cachedBestBlock()
 	if nextScanHeight >= bestBlock.height {
-		if nextScanHeight > bestBlock.height {
+		// When a transaction is initially found in a block, that block is also
+		// scanned for spending transactions (see findTxInBlock) as an
+		// optimization since the block was pulled. As such, it is normal for
+		// the lastScannedHeight for newly located transactions to be the best
+		// block, so there is no need to warn, only return.
+		if nextScanHeight > bestBlock.height+1 {
 			dcr.log.Warnf("Attempted to look for output spender in block %d but current tip is %d!",
 				nextScanHeight, bestBlock.height)
 		}


client/asset/dcr/dcr.go
```diff
@@ -2563,9 +2565,17 @@ func (dcr *ExchangeWallet) SwapConfirmations(ctx context.Context, coinID, contra
...
-	// Find the contract and it's spend status using block filters.
+	// Find the contract and its spend status using block filters.
 	dcr.log.Debugf("Contract output %s:%d NOT yet found, will attempt finding it with block filters.", txHash, vout)
-	return dcr.lookupTxOutWithBlockFilters(ctx, newOutPoint(txHash, vout), p2shScript, matchTime)
+	confs, spent, err = dcr.lookupTxOutWithBlockFilters(ctx, newOutPoint(txHash, vout), p2shScript, matchTime)
+	// Don't trouble the caller if we're using an SPV wallet and the transaction
+	// cannot be located.
+	if errors.Is(err, asset.CoinNotFoundError) && dcr.wallet.SpvMode() {
+		dcr.log.Debugf("SwapConfirmations - cfilters scan did not find %v:%d. "+
+			"Assuming in mempool.", txHash, vout)
```
chappjc, line 2575: Just like the btc spv wallet's handling.

client/asset/dcr/externaltx.go
```diff
@@ -340,7 +340,12 @@ func (dcr *ExchangeWallet) isOutputSpent(ctx context.Context, output *outputSpen
 
 	bestBlock := dcr.cachedBestBlock()
 	if nextScanHeight >= bestBlock.height {
-		if nextScanHeight > bestBlock.height {
+		// When a transaction is initially found in a block, that block is also
+		// scanned for spending transactions (see findTxInBlock) as an
+		// optimization since the block was pulled. As such, it is normal for
+		// the lastScannedHeight for newly located transactions to be the best
+		// block, so there is no need to warn, only return.
+		if nextScanHeight > bestBlock.height+1 {
```
chappjc, line 348: The caller could just not call `isOutputSpent` in certain conditions, but it takes a bunch more work and involved needless locking in places. This is simplest, and you get the `mainchainAncestor` checks just for safety if something changed after locating the txn moments ago.

client/asset/dcr/dcr.go
itswisdomagain, line 2575: I think this assumption is also valid for non-spv wallets. Is the idea that this assumption is less likely for non-spv wallets and so it's okay to alert the caller to the situation?

client/asset/dcr/externaltx.go
```diff
@@ -171,11 +171,11 @@ func (dcr *ExchangeWallet) scanFiltersForTxBlock(ctx context.Context, tx *extern
 	// to block just before earliestTxTime.
 	currentTip := dcr.cachedBestBlock()
 	if lastScannedBlock == nil {
-		dcr.log.Debugf("Searching for tx %s in blocks between block %d (%s) to the block just before %s.",
+		dcr.log.Debugf("Searching for tx %s in blocks between best block %d (%s) and the block just before %s.",
 			tx.hash, currentTip.height, currentTip.hash, earliestTxTime)
 	} else if lastScannedBlock.height < currentTip.height {
 		dcr.log.Debugf("Searching for tx %s in blocks %d (%s) to %d (%s).", tx.hash,
-			currentTip.height, currentTip.hash, lastScannedBlock.height, lastScannedBlock.hash)
+			lastScannedBlock.height, lastScannedBlock.hash, currentTip.height, currentTip.hash)
```
itswisdomagain, line 178: Hmm, was confusing I suppose. It does actually scan blocks in reverse order though, probably not necessarily worth pointing that out in log message.

itswisdomagain (changes requested):

client/asset/dcr/dcr.go
```diff
@@ -2538,8 +2538,10 @@ func (dcr *ExchangeWallet) ValidateSecret(secret, secretHash []byte) bool {
 // the specified swap. The contract and matchTime are provided so that wallets
 // may search for the coin using light filters.
 //
-// If the swap was not funded by this wallet, and it is already spent, this
-// method may return asset.CoinNotFoundError. Compare dcr.externalTxOut.
+// For a non-SPV wallet, if the swap appears spent but it cannot be located in a
+// block with a cfilters scan, this will return asset.CoinNotFoundError. For SPV
+// wallets, it is not an error if the transaction cannot be located SPV wallets
+// cannot see non-wallet transactions until they are mined.
```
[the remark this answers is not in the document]
chappjc: Right that's true that a full node's mempool could be missing it too.  The idea is that would be more cause for alarm than with and SPV wallet.

The alternative to this conditional return of `asset.CoinNotFoundError` on spv status is to adjust in `core` like you said.  Here:

https://github.com/decred/dcrdex/blob/622e702883f5fcefb96fd2410afd736db3b7d931/client/core/trade.go#L793-L808

That's the error log we get, and it does not record the 0 confs.  Fortunately, this function, `counterPartyConfirms`, does not return an error or anything else that will break anything.  If we change `counterPartyConfirms` (and any other callers of `SwapConfirmations`), we'd want to also change BTC so it starts returning `asset.CoinNotFoundError` instead of `nil`

client/asset/btc/spv.go
```diff
@@ -922,6 +922,8 @@ func (w *spvWallet) swapConfirmations(txHash *chainhash.Hash, vout uint32, pkScr
 		if assumedMempool {
 			w.log.Tracef("swapConfirmations - scanFilters did not find %v:%d, assuming in mempool.",
 				txHash, vout)
+			// NOT asset.CoinNotFoundError since this is normal for mempool
+			// transactions with an SPV wallet.
```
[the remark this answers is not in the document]
chappjc: Handling varies in `core`.  This is partly a philosophical question of if it is good to return non-nil errors when nothing unexpected happens.  At least one caller assumed what most callers would assume that a non-nil error is cause for concern.  I drafted the PR this way because it feels like the asset backend is in the position to say if it is cause for concern, not the caller.

client/asset/dcr/dcr.go
chappjc: :100: When this happens repeatedly with non-spv wallets, it is almost always a connectivity issue to fix promptly.

client/asset/btc/spv.go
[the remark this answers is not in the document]
itswisdomagain: But the asset backend can't really be sure, e.g. a spv wallet may return no error for a swap output that was not found because it could be in mempool when it's possible that the output really doesn't exist, yet no error is returned. Conversely, a non-spv wallet may return a `CoinNotFoundError` for a coin that does exist and may only subsequently be found when included in a block.

I guess one question is, is it really an error in the semantic sense of the word, for a non-spv wallet to not find a tx output? I think, if it's not an error for spv wallets, it shouldn't be for non-spv wallets either. It's just significantly more reason to worry for non-spv wallets, but is that what the error return here communicates? Or will callers so assume?

I think in terms of communicating that there's reason to worry because a coin was not found, a log should suffice, preferably on the asset backend since only the asset backend knows if there is a cause for worry (i.e. if !spvMode).

client/asset/dcr/dcr.go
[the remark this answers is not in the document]
itswisdomagain: Or we can adjust `SwapConfirmations` implementations to not return any error for coin not found scenarios unless the asset backend is sure the coin really does not exist. Could return -1 confs to signal that the coin is not yet found and may be subsequently found in a retry (spv or not). Then asset backends can additionally log a warning if the coin was not found but it really should have been (e.g. in non-spv modes).

client/asset/dcr/dcr.go
itswisdomagain: Is it necessary to add to the comment the reason why it's okay to trouble the caller if this is not an spv wallet?

client/asset/dcr/dcr.go
```diff
@@ -2538,8 +2538,10 @@ func (dcr *ExchangeWallet) ValidateSecret(secret, secretHash []byte) bool {
 // the specified swap. The contract and matchTime are provided so that wallets
 // may search for the coin using light filters.
 //
-// If the swap was not funded by this wallet, and it is already spent, this
-// method may return asset.CoinNotFoundError. Compare dcr.externalTxOut.
+// For a non-SPV wallet, if the swap appears spent but it cannot be located in a
+// block with a cfilters scan, this will return asset.CoinNotFoundError. For SPV
```
itswisdomagain, line 2542: Thinking about this again, it can't appear spent and not be located in a block. Not currently anyways. Per the code docs on dcrd's gettxout handler, dcrd treats an output as unspent (whether mined or unmined) as long as it isn't spent in a mined transaction. So if gettxout returns a nil response and the output isn't found by a block search, it's most likely in mempool, just not the node's mempool.

chappjc: > . So if gettxout returns a nil response and the output isn't found by a block search, it's most likely in mempool, just not the node's mempool

Either that or the tx really doesn't exist.  Either way, it's cause for concern.  A full node should pick up on it in short order and if it doesn't then there is connectivity trouble.  We've seen it happen like that

itswisdomagain: Right, I meant regarding the expression "if the swap appears spent but it cannot be located in a block with a cfilters scan" for non-spv wallets. It seems to me that a full node not being able to locate the swap in a block search means it's not spent. Maybe I'm not fully understanding what you meant by that line of comment.

Maybe... `For a non-SPV wallet, if the swap isn't found by the wallet and it cannot be located in block with a cfilters scan, ...`.

chappjc: Ah, yes the "appears spent" wording is confusing, but that's just referring to the `gettxout` part where a nil indicates *not unspent* (whether the output exists or not).  Will rephrase this.

client/asset/btc/spv.go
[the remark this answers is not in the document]
chappjc: > I guess one question is, is it really an error in the semantic sense of the word, for a non-spv wallet to not find a tx output? I think, if it's not an error for spv wallets, it shouldn't be for non-spv wallets either.

This depends on two things: (a) spv vs full node, **AND** (b) the caller's expectations.  The second part is important I think because if the caller is looking for confs/spend on **their own swap**, their wallet would certainly be expected to find it even in spv, whereas if it's checking the **counter-party swap**, it's expected not to find it in spv until it's mined.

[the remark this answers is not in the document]
chappjc: Trying it the other way now, but I'm concerned about big changes since we need to fix these ERR logs in v0.4.

client/asset/dcr/dcr.go
```diff
@@ -1061,7 +1061,7 @@ func (dcr *ExchangeWallet) FundOrder(ord *asset.Order) (asset.Coins, []dex.Bytes
 	if useSplit && !ord.Immediate {
 		// We apply the bumped fee rate to the split transaction when the
 		// PreSwap is created, so we use that bumped rate here too.
-		// But first, check that it's within bounds.
+		// But first, check that its within bounds.
```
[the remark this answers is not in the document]
chappjc: Just reverted this for now.

It's getting messy to change to a signed confs type, and check like `if err != nil || confs == -1 {` or just `err != nil` where appropriate in the caller.  For the next 0.4 release, probably final, I'm not keen to start btc returning a non-nil error where it was not before... just want to silence the errors in core for now.  Will come back to it this afternoon hopefully.

chappjc:
Closing in favor of https://github.com/decred/dcrdex/pull/1418, but can reopen this if we change our minds.

Moving from door number 1 toward door number 2 in the options list:

1. make DCR return nil just like BTC does when an SPV wallet does not find the transaction (this PR does this)
2. make BTC return asset.CoinNotFound and make the consumer (core) maybe ignore this error
3. make both never (regardless of SPV) return asset.CoinNotFound, only log this internal to the backend, but potentially deprive core of this info

Maybe door number 3 can happen in the next release cycle with a `found bool` return from `SwapConfirmations`.

chappjc:
https://github.com/decred/dcrdex/pull/1418/files#r785259509

chappjc:
buy DCR as taker, counterparty confs:

```
2022-01-15 03:17:18.564 [TRC] CORE: Checking confirmations on COUNTERPARTY swap txn a787501ee7e0468f5d1a37c52eadc031beb6d2261cbad48154dd5433fc65386a:0 (dcr)...
2022-01-15 03:17:18.565 [DBG] CORE[dcr]: Contract output a787501ee7e0468f5d1a37c52eadc031beb6d2261cbad48154dd5433fc65386a:0 NOT yet found, will attempt finding it with block filters.
2022-01-15 03:17:18.565 [DBG] CORE[dcr]: Searching for tx a787501ee7e0468f5d1a37c52eadc031beb6d2261cbad48154dd5433fc65386a in blocks between best block 854613 (0000000046692a5805e6453ff9d723be96ada0a8a55b778eeb7791e40c88c2cd) and the block just before 2022-01-15 03:16:20 +0000 UTC.
2022-01-15 03:17:18.566 [DBG] CORE[dcr]: Tx a787501ee7e0468f5d1a37c52eadc031beb6d2261cbad48154dd5433fc65386a NOT found in blocks 854613 (0000000046692a5805e6453ff9d723be96ada0a8a55b778eeb7791e40c88c2cd) to 854613 (0000000046692a5805e6453ff9d723be96ada0a8a55b778eeb7791e40c88c2cd).
2022-01-15 03:17:18.566 [DBG] CORE[dcr]: SwapConfirmations - cfilters scan did not find a787501ee7e0468f5d1a37c52eadc031beb6d2261cbad48154dd5433fc65386a:0. Assuming in mempool.
2022-01-15 03:17:18.566 [TRC] CORE: notify: |DATA| (match)
2022-01-15 03:17:18.566 [DBG] CORE: Match da0decc6132a09fd32938073a44467b444f20b40a28ddbc17efed10559f3ac2c not yet swappable: current confs = 0, required confs = 1
```
(didn't find confs on counter party swap, expected, but no ERR logged)

sell DCR as taker, counterparty confs:

```
022-01-15 03:22:48.564 [TRC] CORE: Checking match 783385febc8f405a9e204e4cbad4e2786011b549a0ba20490e1fd597ab7686e3 (Maker) in status MakerSwapCast. Order: 09e624ecf9f82044a6b5fb1cba80a7a3a4da3952909041a4ed40ca0f3730f0c8, Refund coin: , Script: 6382012088c020ba488575bf42665c52b92c5268c0b6d7dfb03700f818bfc3b8d01f9f1927c4f48876a91497581e91b25425f2168b459902ab9ddc334be7f16704f056e361b17576a9141250b07a815e96a26ccae73bdfc5ea64483572aa6888ac, Revoked: false
2022-01-15 03:22:48.564 [TRC] CORE: Checking confirmations on our OWN swap txn 7b2d08d9304453055ed7ad5b4fe8d9ea720c44a9cd70107c281a5dbc7d19a139:0 (dcr)...
2022-01-15 03:22:48.565 [TRC] CORE: notify: |DATA| (match)
2022-01-15 03:22:48.565 [INF] CORE: Contract for match 783385febc8f405a9e204e4cbad4e2786011b549a0ba20490e1fd597ab7686e3 with swap coin 7b2d08d9304453055ed7ad5b4fe8d9ea720c44a9cd70107c281a5dbc7d19a139:0 (DCR) expires at 2022-01-15 23:21:20 +0000 UTC (19h58m31s).
```
(found own swap confs)

itswisdomagain:
**TL;DR**
I'm not recommending an immediate diversion from the current solution, just noting some potentially harmless implications of this change. The biggest cause for concern with this change for me doesn't require immediate attention: i.e. the possibility that a caller of `SwapConfirmations` may fail to realize that a coin is not found (when using an SPV wallet) and thus take some major action with the assumption that the coin was found because `err == nil`.

**Potentially harmless implications of this change**

3 of the 4 callers of `SwapConfirmations` (`isSwappable`, `isRedeemable` and `shouldBeginFindRedemption`) in `client/asset/trade.go` use the method to look up own swap. In those 3 cases, it would definitely be an error if the wallet were unable to find the swap. SPV wallets would not return any error and those callers would assume the coin was found. In each of those 3 cases however, there is no major net effect. The client would only fail to see an error log which would ordinarily be ideal.

The other caller of `SwapConfirmations` in that file uses it to look up non-wallet swaps. There is a slight difference between how that method handles not finding the swap for SPV and non-SPV wallets. The similarity in handling is that the `match.counterConfirms` remain unchanged (whether the wallet returns an error or not). The difference is that there would be no log for the user to see if it's an SPV wallet. Given the possibility that the swap may actually not exist,  is it not worth it to notify the user at some point? The assumption currently is that the swap is in mempool and would be found eventually when mined. But if it were never found, there wouldn't be any `Failed to get confirmations of the counter-party's swap` for an SPV wallet.

The only advantage to having SPV wallets return a nil error that I see here is to avoid spamming the logs with what would most times be an harmless scenario of not finding a counter-swap. But the same change mean we also won't see error logs that may be important as described in the 2 previous paragraphs. Of course, the possibility of those errors actually occurring in practice is much, much lesser, perhaps negligible(?), they IMO are more worthy of being displayed IF they were to occur. Could we perhaps allow all logging when a coin is not found by `SwapConfirmations` but in the case of counter-party swaps, we also say that there might be no cause for concern? That would be fair to say because whether SPV or not, there really might be no cause for concern unless the logs continue for quite some time. For non-SPV wallets, the log itself is also harmless unless it occurs for quite a while, which would mean there's a need to double check wallet and node connections. For SPV wallets, the error would also be harmless unless it persists after a reasonable amount of time, in which case the log will serve its purpose of notifying the user to a potential problem.

chappjc:
I'd genuinely welcome for a change that reverts and replaces this with minimal code changes.

> Given the possibility that the swap may actually not exist, is it not worth it to notify the user at some point? The assumption currently is that the swap is in mempool and would be found eventually when mined. But if it were never found, there wouldn't be any `Failed to get confirmations of the counter-party's swap` for an SPV wallet

Only two reasons to notify the user if the SPV wallet continually fails to find it: blocks are mined and it is reasonable to assume it would have been included, or match is revoked.  Revocation is handled already.  So prior to revocation, we need a bunch more work to decide "should it have been mined by now".  We should not see a _single_ ERR or WRN logged from the get go.  @buck54321 suggested log throttling and I can see that although it feels a bit like working around a lack of info from the function returns, and the the throttling would *still depend* on what is "to be expected" from a particular wallet type because we want to log almost right away for a full node wallet.

Another thing to note is that the BTC SPV `SwapConfirmations` also does not appear to find the users *own* unconfirmed swaps.  This seems to be different from dcrwallet in SPV.  It's why I closed the alt PR: https://github.com/decred/dcrdex/pull/1418/files#r785259509

Revisiting the debate about `CoinNotFoundError`, I think the reason why are having so much trouble with this is because our SPV wallet implementations were simply not able to perform the task of locating unmined txns.  When they return `0` for confs, that is not wrong.  But it is also expected.  The problem comes in when a caller wants to use `SwapConfirmations` to also infer if the txn was located at all, and that is something that callers probably shouldn't be doing because it's only conditionally a problem.  The fact that no callers even check for `CoinNotFoundError` is salient.  They are all treating a non-nil error as "there is a problem".

Given the above, my feeling is that `SwapConfirmations` needs:
- new `found bool` output argument (or a -1 conf)
- only return `CoinNotFoundError` when it is *unexpected*, a decision only the wallet impl can make

This way, a caller can always log at ERR level any non-nil error, and the caller gets to know if it was found via the `found bool` even if the error was `nil`.

Issue #1328, opened 2021-12-06
contracts: Consider enforcing tx version.


JoeGruffins:
While script versions are checked to be something we expect, transaction versions are not. For the same reasons script versions are checked, I believe we should also check tx versions, to ensure the tx structure is what we expect.

For btc, only version one is ever used afaict. Have not looked into ltc, but can assume nothing changes there. So enforcing one would be good.

For dcr, I believe we can possibly have 1-3. The dex functions fine on one, and that's probably all we have used so far. Two enables sequence locks, and I'm not sure, but if the tx is version 2 or greater, I think we need to check that the tx sequence is a disabled value. 3 enables treasury actions, so, enabling that would probably be fine. I think we can still spend from tgen created outputs fine even at one though, similar to other stake outputs, votes and revocations. However, any versions that don't exist yet should probably be enabled as they come, so that we can consider if we need to change tx validation and raise the allowed version one at a time.

related comment https://github.com/decred/dcrdex/pull/1308#discussion_r755743244

JoeGruffins:
For bitcoin, I think sequence locks are enabled for version 1, so the sequence should also be checked for all tx.

https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki

chappjc:
I think we can check if nSequence on ANY input is _less than_ 0xfffffe (MaxUint32-1), meaning RBF is enabled.  Conversely, not rbf if all inputs have sequence >= 0xfffffe.  But totally final (locktime disabled too), if sequence == 0xffffff.
https://github.com/bitcoin/bitcoin/blob/7fcf53f7b4524572d1d0c9a5fdc388e87eb02416/src/util/rbf.cpp#L9-L17

PR #1404, opened 2022-01-12
server/asset/eth: token backends

buck54321:
```
Convert Backend to AssetBackend that embed *baseBackend similarly to
the AssetWallet in PR 1399. Add interfaces to interact with token
and token swap contracts.
```

### DIFF
--- a/server/asset/eth/eth.go
+++ b/server/asset/eth/eth.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	"decred.org/dcrdex/dex"
+	"decred.org/dcrdex/dex/config"
 	dexeth "decred.org/dcrdex/dex/networks/eth"
 	"decred.org/dcrdex/server/asset"
 	"github.com/decred/dcrd/dcrec/secp256k1/v4"
@@ -26,42 +27,96 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 )
 
+type registeredToken struct {
+	*dexeth.Token
+	drv *TokenDriver
+	ver uint32
+}
+
+var registeredTokens = make(map[uint32]*registeredToken)
+
+func networkToken(assetID uint32, net dex.Network) (token *registeredToken, netToken *dexeth.NetToken, contract *dexeth.SwapContract, err error) {
+	token, found := registeredTokens[assetID]
+	if !found {
+		return nil, nil, nil, fmt.Errorf("no token for asset ID %d", assetID)
+	}
+	netToken, found = token.NetTokens[net]
+	if !found {
+		return nil, nil, nil, fmt.Errorf("no addresses for %s on %s", token.Name, net)
+	}
+	contract, found = netToken.SwapContracts[token.ver]
+	if !found || contract.Address == (common.Address{}) {
+		return nil, nil, nil, fmt.Errorf("no version %d address for %s on %s", token.ver, token.Name, net)
+	}
+	return
+}
+
+func registerToken(assetID uint32, ver uint32) {
+	token, exists := dexeth.Tokens[assetID]
+	if !exists {
+		panic(fmt.Sprintf("no token constructor for asset ID %d", assetID))
+	}
+	drv := &TokenDriver{
+		driverBase: driverBase{
+			version:  ver,
+			unitInfo: token.UnitInfo,
+		},
+		token: token.Token,
+	}
+	asset.RegisterToken(assetID, drv)
+	registeredTokens[assetID] = &registeredToken{
+		Token: token,
+		drv:   drv,
+		ver:   ver,
+	}
+}
+
 func init() {
-	asset.Register(BipID, &Driver{})
+	asset.Register(BipID, &Driver{
+		driverBase: driverBase{
+			version:  version,
+			unitInfo: dexeth.UnitInfo,
+		},
+	})
+
+	// registerToken(testTokenID, 0)
 }
 
 const (
-	version   = 0
-	BipID     = 60
-	assetName = "eth"
+	BipID              = 60
+	ethContractVersion = 0
+	version            = 0
+	assetName          = "eth"
 	// The blockPollInterval is the delay between calls to bestBlockHash to
 	// check for new blocks.
-	blockPollInterval  = time.Second
-	ethContractVersion = 0
+	blockPollInterval = time.Second
 )
 
-var backendInfo = &asset.BackendInfo{
-	SupportsDynamicTxFee: true,
+var (
+	_ asset.Driver      = (*Driver)(nil)
+	_ asset.Driver      = (*Driver)(nil)
+	_ asset.TokenBacker = (*AssetBackend)(nil)
+
+	backendInfo = &asset.BackendInfo{
+		SupportsDynamicTxFee: true,
+	}
+
+	testTokenID, _ = dex.BipSymbolID("dextt.eth")
+)
+
+type driverBase struct {
+	unitInfo dex.UnitInfo
+	version  uint32
 }
-
-var _ asset.Driver = (*Driver)(nil)
-
-// Driver implements asset.Driver.
-type Driver struct{}
 
 // Version returns the Backend implementation's version number.
-func (d *Driver) Version() uint32 {
-	return version
-}
-
-// Setup creates the ETH backend. Start the backend with its Run method.
-func (d *Driver) Setup(configPath string, logger dex.Logger, network dex.Network) (asset.Backend, error) {
-	return NewBackend(configPath, logger, network)
+func (d *driverBase) Version() uint32 {
+	return d.version
 }
 
 // DecodeCoinID creates a human-readable representation of a coin ID for
 // Ethereum. This must be a transaction hash.
-func (d *Driver) DecodeCoinID(coinID []byte) (string, error) {
+func (d *driverBase) DecodeCoinID(coinID []byte) (string, error) {
 	txHash, err := dexeth.DecodeCoinID(coinID)
 	if err != nil {
 		return "", err
@@ -70,8 +125,27 @@ func (d *Driver) DecodeCoinID(coinID []byte) (string, error) {
 }
 
 // UnitInfo returns the dex.UnitInfo for the asset.
-func (d *Driver) UnitInfo() dex.UnitInfo {
-	return dexeth.UnitInfo
+func (d *driverBase) UnitInfo() dex.UnitInfo {
+	return d.unitInfo
+}
+
+// Driver implements asset.Driver.
+type Driver struct {
+	driverBase
+}
+
+// Setup creates the ETH backend. Start the backend with its Run method.
+func (d *Driver) Setup(configPath string, logger dex.Logger, network dex.Network) (asset.Backend, error) {
+	return NewBackend(configPath, logger, network)
+}
+
+type TokenDriver struct {
+	driverBase
+	token *dex.Token
+}
+
+func (d *TokenDriver) TokenInfo() *dex.Token {
+	return d.token
 }
 
 // ethFetcher represents a blockchain information fetcher. In practice, it is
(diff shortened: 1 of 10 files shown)


chappjc:
Awesome, will grind through this shortly!

server/asset/eth/tokener.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Swap error: %w", err)
+	}
+	return dexeth.SwapStateFromV0(&state), nil
+}
+
+// readOnlyCallOpts is the CallOpts used for read-only contract method calls.
+func readOnlyCallOpts(ctx context.Context) *bind.CallOpts {
+	return &bind.CallOpts{
+		Pending: true,
```
chappjc, line 147: With `bind.CallOpts.Pending` set to true, my understanding of their docs is that it will reflect a yet-to-be contract state that would include the execution of any pending transactions interacting with the contract.  Do we want that?  Seems if we update a match state and then that pending tx is replaced, we might have a problem.  What happens if we don't consider pending?

I can see it being desirable if it's retrieving the secret hash from maker's redeem, but if it's auditing the swap initiations, maybe that's not desirable.

chappjc, line 147: With `bind.CallOpts.Pending` set to true, my understanding of their docs is that it will reflect a yet-to-be contract state that would include the execution of any pending transactions interacting with the contract?  Do we want that?  Seems if we update a match state and then that pending tx is replaced, we might have a problem.  What happens if we don't consider pending?

server/asset/eth/tokener.go
```diff
@@ -0,0 +1,150 @@
...
+	BalanceOf(*bind.CallOpts, common.Address) (*big.Int, error)
+}
+
+// tokener is a contract data manager for a token.
+type tokener struct {
+	*registeredToken
+	swapContract
+	tokenAddresser
+	erc20Contract
+	ver                     uint32
```
chappjc, line 44: To clarify, `tokener.ver` here is the version specific to this token, changed if the `erc20Contract` changed, such as if USDT got a new contract address.  Whereas `registeredToken.ver` is the version of the ERC20 swap contract, changed if `swapContract` (and thus `tokenAddresser`) changed, such as if we redeployed a new dex/networks/erc20/contracts/ERC20SwapVx.sol (for either the same erc20 address or a new one).  Is that how the fields and versions are related?

chappjc, line 44: To clarify, `tokener.ver` here is the version specific to this token, changed if the `tokenAddresser` and `erc20Contract` changed.  Whereas `registeredToken.ver` is the version of the common ERC20 swap contract, changed if `swapContract` changed.  Is that how the fields and versions are related?

server/asset/eth/tokener.go
```diff
@@ -0,0 +1,150 @@
...
+	contractAddr, tokenAddr common.Address
```
chappjc, line 45: Related to the last comment, I'm a bit confused about the `tokenAddr` field here.  Would it be expected to be the same address from `tokenAddresser.TokenAddress()`, which says what token the swap contract was deployed for?

chappjc, line 45: Related to the last comment, I'm a bit confused about the `tokenAddr` field here.  Would it be expected to be the same address from `tokenAddresser.TokenAddress()`?

server/asset/eth/tokener.go
```diff
@@ -0,0 +1,150 @@
...
+		return nil, err
+	}
+
+	erc20, err := erc20.NewIERC20(netToken.Address, be)
+	if err != nil {
+		return nil, err
+	}
+
+	tkn := &tokener{
+		ver:             0,
```
chappjc, line 70: I'm wondering if `NetToken` should have a version field to set this, if my understanding of this version field is correct.

chappjc, line 70: I'm wondering if `NetToken` should have a version field if my understanding of this version is correct.

server/asset/eth/tokener.go
```diff
@@ -0,0 +1,150 @@
...
+		tokenAddresser:  es,
+		erc20Contract:   erc20,
+		contractAddr:    swapContract.Address,
+		tokenAddr:       netToken.Address,
+	}
+
+	if boundAddr, err := tkn.tokenAddress(ctx); err != nil {
+		return nil, fmt.Errorf("error retrieving bound address for %s version %d contract: %w",
+			token.Name, token.ver, err)
+	} else if boundAddr != netToken.Address {
```
chappjc, line 82: :+1: OK that jives. But this means we don't need one of either `tokener.tokenAddr` or `tokener.tokenAddresser` I think

chappjc, line 82: :+1: OK that jives. But this means we don't need the one if either `tokener.tokenAddr` or `tokener.tokenAddresser` I think

server/asset/eth/tokener.go
```diff
@@ -0,0 +1,150 @@
...
+	var v uint64
+	for _, init := range inits {
+		v += init.Value
+	}
+	return dexeth.GweiToWei(v)
+}
+
+// tokenAddress fetches the token address bound to the swap contract.
+func (t *tokener) tokenAddress(ctx context.Context) (common.Address, error) {
+	return t.tokenAddresser.TokenAddress(readOnlyCallOpts(ctx))
```
chappjc, line 115: `tokenAddresser` can be omitted like you did in `balanceOf` with `erc20Contract`

chappjc, line 115: `tokenAddresser` can be omitted like you did in `balanceOf` with `erc20Contract`

chappjc:
That was odd.  It duplicated all my comments.

server/asset/eth/tokener.go
buck54321: All versions refer to the swap contract version and swap contract address, updated by us only if a new swap contract is implemented or if gases need to be updated for some reason. We have no mechanism to update the token erc20 contract address.

buck54321:
> contractAddr, tokenAddr common.Address

It did it in my email notification too.

server/asset/eth/tokener.go
buck54321: RBF is possible in Bitcoin, but we still consider mempool transactions when looking for contract data. Isn't this the same?

server/asset/eth/tokener.go
chappjc: Ok. So what I'm seeing is that `tokener` has a `ver` from the embedded `registereToken` and another one at this comment, which overrides/shadows the embedded one.  They are intended to be the same from what I understand now.

I think you can get rid of this explicit `tokener.ver` field then.

server/asset/eth/tokener.go
chappjc: True.  Although arguably we should be disallowing RBF by checking the sequence bytes.
But I think I agree with you now.  Pending is kinda important to maintain the ability to locate the init/swap txns in mempool and get audits rolling before the first conf.

server/asset/eth/eth.go
```diff
@@ -26,42 +27,96 @@ import (
...
 func init() {
-	asset.Register(BipID, &Driver{})
+	asset.Register(BipID, &Driver{
+		driverBase: driverBase{
+			version:  version,
+			unitInfo: dexeth.UnitInfo,
+		},
+	})
+
+	// registerToken(testTokenID, 0)
```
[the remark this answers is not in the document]
buck54321: I was thinking we would add a line here for each supported testnet/mainnet token. At registration, we make no distinction between the networks, and it's of little consequence, but the `asset.Tokens` method will return the asset ID of all registered tokens, even if they don't have a mainnet configuration.

server/asset/eth/tokener.go
JoeGruffins: For my understanding:

We are just checking that the swap contract is for the correct token here correct? Is there a need to save the token's contract address beyond that? We will coordinate with contract address? If same contract must be same token correct, because the token address is hard coded?

server/asset/eth/tokener.go
JoeGruffins: > we should be disallowing RBF by checking the sequence bytes.

Related to  #1328 I guess.

JoeGruffins (commented):
Looks good. I would like it if you could break up the methods a bit more and do away with all the `eth.assetID != BipID` checks. You made `baseBackend` but it's only used by `AssetBackend` so I don't see why you couldn't go ahead and make a proper `TokenBackend` with different rpc calls as well.

server/asset/eth/eth.go
```diff
@@ -461,24 +642,22 @@ func (eth *Backend) ValidateSignature(addr string, pubkey, msg, sig []byte) erro
...
+		u := &asset.BlockUpdate{
+			Reorg: reorg,
+			Err:   err,
+		}
+
+		eth.sendBlockUpdate(u)
+
+		for _, be := range eth.tokens {
+			be.sendBlockUpdate(u)
 		}
```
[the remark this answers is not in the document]
buck54321: > Does `eth.tokens` also need a lock, or are all the tokens already expected to be loaded by now?

They are all expected to be loaded. https://github.com/decred/dcrdex/blob/105d3e730f5ac193e8fe91c30fb4615a18b260e5/server/dex/dex.go#L526-L541

> Also, could this potentially take a long time? Maybe depends on how many tokens we have and how many markets. Is it one send per coin per market it is used in?
> 
> Polling interval is one second. I don't know if we may just miss polling intervals if this method `poll` takes too long. That's probably fine though?

So I'm not **too** worried about it taking too long here. Maybe I can run in a goroutine? But you've reminded me of something I've been meaning to get into. I think we should meter block updates from all assets, since `Swapper` does a `processBlock` and a `checkInactionBlockBased` for every block. Ethereum blocks are coming every 10 seconds on average, so there'll be a ton of relatively simultaneous blocks. I think `Swapper` could use some kind of metering mechanism, ensuring that there is at least say 5 or 10 seconds between taking action on sequential block checks.

server/asset/eth/tokener.go
```diff
@@ -0,0 +1,150 @@
...
+	inits, err := dexeth.ParseInitiateData(txData, t.ver)
+	if err != nil {
+		return nil
+	}
+	var v uint64
+	for _, init := range inits {
+		v += init.Value
+	}
+	return dexeth.GweiToWei(v)
+}
```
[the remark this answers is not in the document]
buck54321: These are expected to fail more often than they pass. What would you do with the error?

server/asset/eth/eth.go
```diff
@@ -85,113 +159,152 @@ type ethFetcher interface {
...
+		}
+		var wg sync.WaitGroup
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			select {
+			case <-ctx.Done():
+			case <-eth.baseBackend.ctx.Done():
+			}
+		}()
```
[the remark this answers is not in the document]
buck54321: We still need to `wg.Done()` when the `Context` is canceled since this is a `dex.Connector`.

martonp (commented):
Looks good!

server/asset/eth/eth.go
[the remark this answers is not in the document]
chappjc: +1 to metering in Swapper

server/asset/eth/eth.go
```diff
@@ -26,42 +27,96 @@ import (
...
+	return
+}
+
+func registerToken(assetID uint32, ver uint32) {
+	token, exists := dexeth.Tokens[assetID]
+	if !exists {
+		panic(fmt.Sprintf("no token constructor for asset ID %d", assetID))
+	}
+	drv := &TokenDriver{
+		driverBase: driverBase{
```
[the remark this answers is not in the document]
buck54321: I didn't see any reason for a pointer.

server/asset/eth/eth.go
```diff
@@ -26,42 +27,96 @@ import (
...
+		return nil, nil, nil, fmt.Errorf("no version %d address for %s on %s", token.ver, token.Name, net)
+	}
+	return
+}
+
+func registerToken(assetID uint32, ver uint32) {
+	token, exists := dexeth.Tokens[assetID]
+	if !exists {
+		panic(fmt.Sprintf("no token constructor for asset ID %d", assetID))
+	}
```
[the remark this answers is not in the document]
buck54321: I'm not seeing it. I'd panic in `init` anyway, so it would just add a couple of lines per invocation of `registerToken`.

server/asset/eth/tokener.go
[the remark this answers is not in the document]
JoeGruffins: Log as trace?

server/asset/eth/eth.go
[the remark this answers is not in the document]
JoeGruffins: Just what we normally do. Wondering if there was a deeper meaning.

server/asset/eth/eth.go
```diff
@@ -231,29 +244,29 @@ func unconnectedETH(logger dex.Logger, net dex.Network) (*AssetBackend, error) {
...
 // Backend.
-func NewBackend(ipc string, logger dex.Logger, net dex.Network) (*AssetBackend, error) {
-
+func NewBackend(ipc string, logger dex.Logger, net dex.Network) (*ETHBackend, error) {
 	switch net {
 	case dex.Simnet:
 	case dex.Testnet:
 	case dex.Mainnet:
-		// TODO: Allow.
+		// TODO: Allow. When?
```
JoeGruffins, line 269: LOL

chappjc: Haha, deploy away!

server/asset/eth/eth.go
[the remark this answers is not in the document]
JoeGruffins: > Maybe I can run in a goroutine?

I was gonna suggest this, but we definitely wan the updates to happen sequentially. I don't think a goroutine would be good.

JoeGruffins (approved):
Thanks for addressing my comments.

martonp (approved):

Issue #1331, opened 2021-12-07
client/asset/btc: spurious TestReorg failure


chappjc:
```
2021-12-07 02:42:47.092 [INF] TEST: Reorg from 873cc0ea9b9ec18437f5198bb389ac728f5f4c45fcdd893a310facd9f5511802 (51) to 7d0f8a928585390908b802bdfc8df0dcf9f95edee4ae6d5c92e2b89ad1bdc176 (50) detected.
2021-12-07 02:42:47.092 [TRC] TEST: Notifying 0 btc asset consumers of new block at height 50
2021-12-07 02:42:47.142 [DBG] TEST: New block 873cc0ea9b9ec18437f5198bb389ac728f5f4c45fcdd893a310facd9f5511802 (51)
2021-12-07 02:42:47.142 [TRC] TEST: Notifying 0 btc asset consumers of new block at height 51
2021-12-07 02:42:47.250 [INF] TEST: Reorg from 873cc0ea9b9ec18437f5198bb389ac728f5f4c45fcdd893a310facd9f5511802 (51) to 0ce6e5cd00c4e533e9a3bb18c948b9e962e5dcfd887953e868571ddca6e10be5 (51) detected.
2021-12-07 02:42:47.250 [TRC] TEST: Notifying 0 btc asset consumers of new block at height 51
--- FAIL: TestReorg (0.56s)
    btc_test.go:1267: chain A block still mainchain for test 2:2 after reorg
2021-12-07 02:42:47.255 [ERR] TEST: CheckAddress error for btc : decoded address is of unknown format
2021-12-07 02:42:47.256 [ERR] TEST: CheckAddress error for btc 28Zpft83eov56iESWuPpV8XFLJ1b8gMZy7: checksum mismatch
2021-12-07 02:42:47.256 [ERR] TEST: CheckAddress error for btc 3GD2fSQxhkXDAW66i6JBwCqhLFSvhMNrtO: decoded address is of unknown format
2021-12-07 02:42:47.256 [ERR] TEST: CheckAddress error for btc 3GD2fSQx: checksum mismatch
FAIL
FAIL	decred.org/dcrdex/server/asset/btc	1.454s
```

chappjc:
just hit on a PR

PR #1405, opened 2022-01-12
[intl] Add Missing pt-br strings

vctt94:
This PR adds missing pt-br strings.


chappjc:
Looks like the last CI run hit https://github.com/decred/dcrdex/issues/1331, so I guess we should keep that on our radar.  I started a new CI run though.

chappjc (approved):

PR #1407, opened 2022-01-13
client/core: track connection on discover via Register

chappjc:
possibly resolving discussion in https://github.com/decred/dcrdex/pull/1398#issuecomment-1011641305

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -2712,6 +2712,11 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
 	}
 	if paid {
 		registrationComplete = true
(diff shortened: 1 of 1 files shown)


JoeGruffins (approved):
This does it.

chappjc:
It was just missing the `c.conns` map entry.  The goroutines (e.g. `listen`) were started in `connectDEX` because it wasn't a temp conn in `Register`, it just wasn't tracked in the map.
Re-tested it it's still working the same as when we tested it earlier except there aren't two `listen` goroutines running.

buck54321 (approved):

Issue #1408, opened 2022-01-13
btc TestReorg test panic


chappjc:
https://github.com/decred/dcrdex/runs/4798900012?check_suite_focus=true#step:5:1799

On master

PR #1409, opened 2022-01-13
server/asset/btc: fix panic in BTC TestReorg

LasTshaMAN:
Closes https://github.com/decred/dcrdex/issues/1408.

I can reproduce this panic by making `blockPollDelay` shorter in these unit-tests (see screenshot below). So extending this interval might help with encountering this `panic`.

From what I understand, the issue happens because these tests rely on timing (go-routine managing `time.Ticker` is not synchronised with the one running this test), and so `btc.blockCache` might not have the necessary block by the time `atHeight` gets called.

I've also checked tests for other assets, they shouldn't suffer from the same issue (DCR tests [explicitly sync block-cache](https://github.com/LasTshaMAN/dcrdex/blob/master/server/asset/dcr/dcr_test.go#L1245-L1269) and other coins don't have similar tests).

telegram-cloud-photo-size-2-5406630920015951656-y

### DIFF
--- a/server/asset/btc/btc_test.go
+++ b/server/asset/btc/btc_test.go
@@ -48,8 +48,8 @@ func TestMain(m *testing.M) {
 	// accommodate reorg testing.
 	blockPollInterval = time.Millisecond * 50
 	// blockPollDelay must be long enough to allow for block polling go-routine
-	// to get a single tick (must be at least as long as blockPollInterval and
-	// ideally significantly longer).
+	// to run at least a single tick (must be at least as long as blockPollInterval
+	// and ideally significantly longer).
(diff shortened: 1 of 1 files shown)


JoeGruffins (approved):
Makes sense.

I see the test also ignores the `found` return from `atHeight`. It could check that and fail with an error message rather than panicking if not found.

LasTshaMAN:
> I see the test also ignores the found return from atHeight. It could check that and fail with an error message rather than panicking if not found.

Right, would be nice to have meaningful error message for this failure case, I've added a commit on top to address that.

JoeGruffins (approved):

chappjc (approved):

Issue #1406, opened 2022-01-12
dcr: websocket client has already connected


chappjc:
We've seen this error in the past and the causes have been variable, but it's reported presently with 0.4 RC

![image](https://user-images.githubusercontent.com/9373513/149236205-21194417-934b-4c10-8ec4-e5a48b6f55a5.png)

High priority

previous causes
- https://github.com/decred/dcrdex/pull/1121
- https://github.com/decred/dcrdex/pull/1025

A related issue is that if an older dcrwallet is used, the RPC server version error gets logged, but this connect error happens too and that ends up in the notifications.  The wallet is generally hosed until restart.

chappjc:
I see the problem with dcr's `rpcWallet` approach to `Connect` and investigating a resolution.

chappjc:
Resolved by https://github.com/decred/dcrdex/pull/1472 and https://github.com/decred/dcrdex/pull/1424

PR #1413, opened 2022-01-14
client/core: reload wallet on connect fail

chappjc:
Resolves https://github.com/decred/dcrdex/issues/1406

You cannot do `wallet.Disconnect` and later `wallet.Connect`, at least not with dcr's `ExchangeWallet` using `rpcclient`.  I think that we should assume that once you do `asset.Wallet.Disconnect` you cannot do `asset.Wallet.Connect` although that would depend on the implementation.

This PR solves the issue by creating a new `xcWallet` to replace the one the `Core.conns` map if the connection fails.  This is obviously only required if the `xcWallet` was in the map in the first place.

A simple way to test this is with an counted error in `(*rpcWallet).checkRPCConnection`:

```diff
diff --git a/client/asset/dcr/rpcwallet.go b/client/asset/dcr/rpcwallet.go
index 314ebe08..df4b51aa 100644
--- a/client/asset/dcr/rpcwallet.go
+++ b/client/asset/dcr/rpcwallet.go
@@ -239,9 +239,15 @@ func (w *rpcWallet) checkRPCConnection(ctx context.Context) error {
                w.log.Infof("Connected to dcrwallet (JSON-RPC API v%s) in SPV mode", walletSemver)
        }
 
+       if fail < 3 {
+               fail++
+               return errors.New("asdf")
+       }
        return nil
 }
 
+var fail int
+
```

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1455,22 +1455,54 @@ func (c *Core) connectAndUpdateWallet(w *xcWallet) error {
 	return nil
 }
 
+// reloadWallet creates a fresh xcWallet via loadWallet from the provided
+// xcWallet. This creates a new asset.Wallet so that repeated connection
+// attempts are not met with confusing "already connected" errors if the
+// underlying asset.Wallet implementation is not capable of handling repeated
+// Connect request, which is a common limitation of wallet clients, especially
+// after a Disconnect.
+func (c *Core) reloadWallet(wallet *xcWallet) (*xcWallet, error) {
+	wallet.mtx.RLock()
+	defer wallet.mtx.RUnlock()
+	return c.loadWallet(&db.Wallet{
+		AssetID:     wallet.AssetID,
+		Type:        wallet.cfg.Type,
+		Settings:    wallet.cfg.Settings,
+		Balance:     wallet.balance.Balance,
+		EncryptedPW: wallet.encPass,
+		Address:     wallet.address,
+	})
+}
+
 // connectedWallet fetches a wallet and will connect the wallet if it is not
 // already connected. If the wallet gets connected, this also emits WalletState
 // and WalletBalance notification.
 func (c *Core) connectedWallet(assetID uint32) (*xcWallet, error) {
-	wallet, exists := c.wallet(assetID)
+	c.walletMtx.RLock()
+	defer c.walletMtx.RUnlock()
+	wallet, exists := c.wallets[assetID]
 	if !exists {
 		return nil, newError(missingWalletErr, "no configured wallet found for %s (%d)",
 			strings.ToUpper(unbip(assetID)), assetID)
 	}
-	if !wallet.connected() {
-		err := c.connectAndUpdateWallet(wallet)
-		if err != nil {
-			return nil, err
-		}
+	if wallet.connected() {
+		return wallet, nil
 	}
-	return wallet, nil
+
+	err := c.connectAndUpdateWallet(wallet)
+	if err == nil {
+		return wallet, nil
+	}
+
+	// When connect fails, assume the wallet is dead, so create a new xcWallet.
+	if newWallet, errLoad := c.reloadWallet(wallet); errLoad == nil {
+		c.wallets[assetID] = newWallet
+	} else {
+		// Leave the old xcWallet in the map for future attempt.
+		c.log.Errorf("Failed to reload %v wallet: %v", unbip(assetID), errLoad)
+	}
+
+	return nil, err
 }
 
 // connectWallet connects to the wallet and returns the deposit address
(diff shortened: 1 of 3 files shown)


client/core/wallet.go
```diff
@@ -197,34 +200,51 @@ func (w *xcWallet) connected() bool {
 // flag to true, and validates the deposit address. Use Disconnect to cleanly
 // shutdown the wallet.
 func (w *xcWallet) Connect() error {
+	reloadWallet := func() {
+		wallet, err := asset.OpenWallet(w.AssetID, w.cfg, w.logger, w.net)
```
chappjc, line 204: a connect error now requires on an asset.Driver being registered for this asset ID, and these new xcWallet fields being set

client/core/wallet.go
```diff
@@ -197,34 +200,51 @@ func (w *xcWallet) connected() bool {
...
+		if err != nil {
+			w.logger.Errorf("error opening wallet: %v", err)
+			return
+		}
+		w.Wallet = wallet
```
chappjc, line 209: Modifying this field would seem to be racy, but consumers won't be using any other wallet methods until the wallet has been connected.  I believe it's safe to assume that if correct access patterns with the `xcWallet` methods will be sufficient.  I don't seen any issues jumping out based on the code, have not observed any races in testing.

buck54321: This seems iffy. Almost certainly OK for our UI. Dims any prospects for concurrent use though. I'm just wondering if it'd be better to replace the entire `xcWallet`.

chappjc: Thought about that too, but you run into similar concurrency issues, just higher up.  Worth a shot though.

chappjc: Hmm, maybe `(c *Core) connectWallet(w *xcWallet)` could create a new `xcWallet` if `Connect` failed, and pop it into the `c.wallets` map...

chappjc: I think that actually worked out fairly well.  It's more involved because there are many more paths to `connectWallet` than just `connectedWallet`, but in the end I think this may have simplified things somewhat.

client/core/core.go
```diff
@@ -3111,6 +3114,23 @@ func (c *Core) Login(pw []byte) (*LoginResult, error) {
 				if err != nil {
 					c.log.Errorf("Unable to connect to %s wallet (start and sync wallets BEFORE starting dex!): %v",
 						unbip(wallet.AssetID), err)
+					// NOTE: Details for this topic is in the context of fee
+					// payment, but the subject pertains to a failure to connect
+					// to the wallet. This is better than just an ERR log.
+					subject, _ := c.formatDetails(TopicWalletConnectionWarning)
+					c.notify(newWalletConfigNote(TopicWalletConnectionWarning, subject, err.Error(),
+						db.ErrorLevel, wallet.state()))
```
chappjc, line 3122: `Login` was not communicating wallet connect errors except on the console as ERR logs, and that's one reason for the confusion in the issue this PR resolves.  Now you get a report of the connect error in the browser ntfns

image


2022-01-17 17:01:46.426 [INF] CORE: Connecting wallet for dcr
2022-01-17 17:01:46.426 [INF] CORE: Connecting wallet for btc
2022-01-17 17:01:46.427 [ERR] CORE: Unable to connect to dcr wallet (start and sync wallets BEFORE starting dex!): connectWallet: dcrwallet connect error: dial tcp 127.0.0.1:19572: connect: connection refused
2022-01-17 17:01:46.430 [ERR] CORE: notify: |ERROR| (walletconfig) Wallet connection warning - connectWallet: dcrwallet connect error: dial tcp 127.0.0.1:19572: connect: connection refused

JoeGruffins (approved):

client/core/core.go
```diff
@@ -1455,22 +1455,54 @@ func (c *Core) connectAndUpdateWallet(w *xcWallet) error {
...
+	})
+}
+
 // connectedWallet fetches a wallet and will connect the wallet if it is not
 // already connected. If the wallet gets connected, this also emits WalletState
 // and WalletBalance notification.
 func (c *Core) connectedWallet(assetID uint32) (*xcWallet, error) {
-	wallet, exists := c.wallet(assetID)
+	c.walletMtx.RLock()
+	defer c.walletMtx.RUnlock()
```
chappjc, line 1482: whoops needs to be a full write lock

client/core/core.go
chappjc: Note that this bit was already merged into release-v0.4 in https://github.com/decred/dcrdex/pull/1424

client/core/core.go
```diff
@@ -1455,22 +1455,54 @@ func (c *Core) connectAndUpdateWallet(w *xcWallet) error {
...
 	}
-	return wallet, nil
+
+	// When connect fails, assume the wallet is dead, so create a new xcWallet.
+	if newWallet, errLoad := c.reloadWallet(wallet); errLoad == nil {
+		c.wallets[assetID] = newWallet
+	} else {
+		// Leave the old xcWallet in the map for future attempt.
+		c.log.Errorf("Failed to reload %v wallet: %v", unbip(assetID), errLoad)
+	}
```
buck54321, line 1503: My instinct is to formalize this as a policy and make asset packages work with it, in which case the call to `connectAndUpdateWallet` wouldn't be made at all. Thoughts?

chappjc: Can you expand on that?  You mean make asset packages have working Connect after a Disconnect so that nothing needs to be reconstructed at this level or even inside of xcWallet?
Could happen, will require tons of mutex action in the asset.Wallet impls most likely.  Even trying this inside the xcWallet got nasty and it will be tougher inside each Wallet, but I'm open to that.
Will happily close this PR if there's a better solution.  I didn't find it.

buck54321: The opposite. Wallets packages should expect to be discarded after shutdown.

chappjc: Ahhh.  Lemme see about that.  I think the connect logic comes into play if you start core when the external wallet is down...

buck54321: There appears to be an assumption here that the wallet is disconnected after a proper shutdown, and therefore without any active state that might need to be rebuilt such as funding coins and queued redemption searches. We're also not replacing the `walletSet` of any existing orders that uses this wallet. It looks like the old way, in the case of an unintended disconnect, could potentially reconnect and resume trades. That wouldn't be the case anymore.

chappjc: Oh damn you're right the wallets are held in fields all over the place.

chappjc:
Replaced by https://github.com/decred/dcrdex/pull/1472

Issue #1372, opened 2021-12-26
server/asset/eth: max fee rate needs to be a max gas cap


chappjc:
As we have discussed previously, with ETH's eip-1559 "dynamic" transactions:

- dynamic eip-1559 transactions have two fee settings (`GasFeeCap` and `GasTipCap`) while legacy fee transactions have one (`GasPrice`)
- the `MaxFeeRate` asset const needs to be something quite high, and as with other asset params, **must be set in the server config**, and this must be high enough for any server-prescribed max fee.

Since a server's fee "suggestions" and prescribed match fee rate in our comms protocol only supports a single rate:

- The prescribed match fee rate must specify the desired **gas fee cap**, which sets the transactions limit of effective base fee + tip.
- Thus, a dynamic transaction's **gas fee cap** should be some combination of **gas tip cap** plus a _scaled_ **current base fee**.

A sensible formula for this is used in `go-ethereum/accounts/bind/(*BoundContract).createDynamicTx`:

```
gasFeeCap = gasTipCap + 2 * headBaseFee
```

https://github.com/ethereum/go-ethereum/blob/356bbe343a30789e77bb38f25983c8f2f2bfbb47/accounts/abi/bind/base.go#L250-L255

```go
gasFeeCap := opts.GasFeeCap
if gasFeeCap == nil {
	gasFeeCap = new(big.Int).Add(
		gasTipCap,
		new(big.Int).Mul(head.BaseFee, big.NewInt(2)),
	)
}
```

Presently the server gives a fee rate that is obtained from a "legacy" fee rate function that just sums current base fee rate with suggested tip.  This is far too low for a max gas fee for a "dynamic" transaction.  I started work on this, addressing prerequesites in https://github.com/decred/dcrdex/pull/1356
In that PR description I've linked a draft commit that will use the base fee and tip suggestion to compute a max gas fee for clients using a formula like the one I've given above.
https://github.com/decred/dcrdex/commit/64a72b48fe3b05fcf8d5f4a7d9296496092967ac

JoeGruffins:
imo. 

- The server only needs to send the max tip. 
This is because the base fee is not negotiable. It is part of consensus and not up to the miner to decide to include your tx based on this fee.

- We don't need to bother with the legacy fee type.
[...]

Issue #1336, opened 2021-12-08
client/eth: Respect user's max fee rate.


JoeGruffins:
eth should send transactions with a max fee chosen by the user. I really think this should just be changed to the chosen max fee as well: https://github.com/decred/dcrdex/blob/6bd4dc842fb5999c20c4c66d677dfa4cbc2fbaca/client/asset/eth/nodeclient.go#L471-L473
and there should be an error if the base fee plus miner tip would be over the chosen max fee. Maybe this is the only place that would need to change, besides recording the fee on construction of the `ExchangeWallet`.

martonp:
Should we just not allow a nil `MaxFeeRate` to be passed into this function?

JoeGruffins:
yes imo. If you do `maxFeeRate = new(big.Int).Mul(baseFee, big.NewInt(2))` Then that could really be anything depending on the network. There are some docs somewhere that says this specific 2x base is optimal for ensuring the tx gets included in the next six blocks. @buck54321 set this max so if he has thoughts about changing this...

JoeGruffins:
Will attempt to solve this.

JoeGruffins:
Related convo: https://github.com/decred/dcrdex/pull/1485#issuecomment-1049970496

The fee is being respected for initiations. It was never intended to block other transactions.

PR #1414, opened 2022-01-14
eth: Fix init and redeem.

JoeGruffins:
With these changes I was able to successfully swap.

A few of the current problems are pointed out.

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -134,9 +134,17 @@ func (d *Driver) Open(cfg *asset.WalletConfig, logger dex.Logger, network dex.Ne
 
 // DecodeCoinID creates a human-readable representation of a coin ID for Ethereum.
 func (d *Driver) DecodeCoinID(coinID []byte) (string, error) {
+	const invalid = "<invalid coin>"
+	if len(coinID) == common.HashLength {
+		id, err := dexeth.DecodeCoinID(coinID)
+		if err != nil {
+			return invalid, err
+		}
+		return id.String(), nil
+	}
 	id, err := decodeFundingCoinID(coinID)
 	if err != nil {
-		return "<invalid coin>", err
+		return invalid, err
 	}
 	return id.String(), nil
 }
@@ -689,7 +697,9 @@ func (eth *ExchangeWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin
 		return nil, nil, 0, fmt.Errorf("unfunded contract. %d < %d", totalInputValue, totalSpend)
 	}
 
-	tx, err := eth.node.initiate(eth.ctx, swaps.Contracts, swaps.FeeRate, swaps.AssetVersion)
+	// TODO: Fix fee rate. The current fee rate returned from the server
+	// will not allow this to be mined on simnet.
+	tx, err := eth.node.initiate(eth.ctx, swaps.Contracts, 200 /*swaps.FeeRate*/, swaps.AssetVersion)
 	if err != nil {
 		return fail(fmt.Errorf("Swap: initiate error: %w", err))
 	}
@@ -734,7 +744,6 @@ func (eth *ExchangeWallet) Redeem(form *asset.RedeemForm) ([]dex.Bytes, asset.Co
 	}
 
 	var contractVersion uint32 // require a consistent version since this is a single transaction
-	inputs := make([]dex.Bytes, 0, len(form.Redemptions))
 	var redeemedValue, unlocked uint64
 	for i, redemption := range form.Redemptions {
 		// NOTE: redemption.Spends.SecretHash is a dup of the hash extracted
@@ -773,7 +782,6 @@ func (eth *ExchangeWallet) Redeem(form *asset.RedeemForm) ([]dex.Bytes, asset.Co
 		}
 		redeemedValue += swapData.Value
 		unlocked += redemption.UnlockedReserves
-		inputs = append(inputs, redemption.Spends.Coin.ID())
 	}
 
 	outputCoin := eth.createFundingCoin(redeemedValue)
@@ -781,14 +789,22 @@ func (eth *ExchangeWallet) Redeem(form *asset.RedeemForm) ([]dex.Bytes, asset.Co
 
 	// TODO: make sure the amount we locked for redemption is enough to cover the gas
 	// fees. Also unlock coins.
-	_, err := eth.node.redeem(eth.ctx, form.Redemptions, form.FeeSuggestion, contractVersion)
+	// TODO: Fix fee rate. The current fee rate returned from the server
+	// will not allow this to be mined on simnet.
+	tx, err := eth.node.redeem(eth.ctx, form.Redemptions, 200 /*form.FeeSuggestion*/, contractVersion)
 	if err != nil {
 		return fail(fmt.Errorf("Redeem: redeem error: %w", err))
 	}
 
 	eth.UnlockReserves(unlocked)
 
-	return inputs, outputCoin, fundsRequired, nil
+	txHash := tx.Hash()
+	txs := make([]dex.Bytes, len(form.Redemptions))
+	for i := range txs {
+		txs[i] = txHash[:]
+	}
+
+	return txs, outputCoin, fundsRequired, nil
 }
 
 // recoverPubkey recovers the uncompressed public key from the signature and the
@@ -873,11 +889,11 @@ func (eth *ExchangeWallet) SignMessage(_ asset.Coin, msg dex.Bytes) (pubkeys, si
 // AuditContract retrieves information about a swap contract on the
 // blockchain. This would be used to verify the counter-party's contract
 // during a swap. coinID is expected to be the transaction id, and must
-// be the same as the hash of txData. contract is expected to be
+// be the same as the hash of serializedTx. contract is expected to be
 // (contractVersion|secretHash) where the secretHash uniquely keys the swap.
-func (eth *ExchangeWallet) AuditContract(coinID, contract, txData dex.Bytes, rebroadcast bool) (*asset.AuditInfo, error) {
+func (eth *ExchangeWallet) AuditContract(coinID, contract, serializedTx dex.Bytes, rebroadcast bool) (*asset.AuditInfo, error) {
 	tx := new(types.Transaction)
-	err := tx.UnmarshalBinary(txData)
+	err := tx.UnmarshalBinary(serializedTx)
 	if err != nil {
 		return nil, fmt.Errorf("AuditContract: failed to unmarshal transaction: %w", err)
 	}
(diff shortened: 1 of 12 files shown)


client/asset/eth/eth.go
```diff
@@ -134,9 +134,17 @@ func (d *Driver) Open(cfg *asset.WalletConfig, logger dex.Logger, network dex.Ne
...
+		}
+		return id.String(), nil
+	}
 	id, err := decodeFundingCoinID(coinID)
 	if err != nil {
-		return "<invalid coin>", err
+		return invalid, err
 	}
 	return id.String(), nil
 }
```
JoeGruffins, line 150: Sometime a tx hash, and we can gauge that by length. Otherwise a funding coin.

client/asset/eth/eth.go
```diff
@@ -689,7 +697,7 @@ func (eth *ExchangeWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin
 		return nil, nil, 0, fmt.Errorf("unfunded contract. %d < %d", totalInputValue, totalSpend)
 	}
 
-	tx, err := eth.node.initiate(eth.ctx, swaps.Contracts, swaps.FeeRate, swaps.AssetVersion)
+	tx, err := eth.node.initiate(eth.ctx, swaps.Contracts, 200 /*swaps.FeeRate*/, swaps.AssetVersion)
```
JoeGruffins, line 700: Here and in redeem, this was too low. Will fix this better.

client/asset/eth/eth.go
```diff
@@ -773,22 +781,28 @@ func (eth *ExchangeWallet) Redeem(form *asset.RedeemForm) ([]dex.Bytes, asset.Co
...
 	eth.UnlockReserves(unlocked)
 
-	return inputs, outputCoin, fundsRequired, nil
+	txHash := tx.Hash()
+	txs := make([]dex.Bytes, len(form.Redemptions))
+	for i, _ := range txs {
+		txs[i] = txHash[:]
+	}
+
+	return txs /*inputs*/, outputCoin, fundsRequired, nil
```
JoeGruffins, line 805: The server redeem coin wants the redeeming tx. Does this mess up balance calculations on the server side?

client/asset/eth/eth.go
```diff
@@ -876,23 +890,25 @@ func (eth *ExchangeWallet) SignMessage(_ asset.Coin, msg dex.Bytes) (pubkeys, si
...
 
-	txHash := tx.Hash()
-	if !bytes.Equal(coinID, txHash[:]) {
-		return nil, fmt.Errorf("AuditContract: coin id != txHash - coin id: %x, txHash: %x", coinID, tx.Hash())
-	}
+		txHash := tx.Hash()
+		if !bytes.Equal(coinID, txHash[:]) {
+			return nil, fmt.Errorf("AuditContract: coin id != txHash - coin id: %x, txHash: %x", coinID, tx.Hash())
+		}
+	*/
```
JoeGruffins, line 904: We are currently only sending the txdata as txData. This is not the whole transaction and shouldn't have enough info to recreate the tx. We could send the serialized tx instead, if it is necessary. It will be if we want to resend below.

client/asset/eth/eth.go
```diff
@@ -935,16 +952,21 @@ func (eth *ExchangeWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time,
...
 	if err != nil {
 		return false, time.Time{}, err
 	}
-	blockTime := time.Unix(int64(header.Time), 0)
 
-	swap, err := eth.node.swap(eth.ctx, secretHash, contractVer)
+	// Time is not yet set for uninitiated swaps.
+	if swap.State == dexeth.SSNone {
+		return false, time.Time{}, asset.ErrSwapNotInitiated
+	}
```
JoeGruffins, line 963: Useful for not logging this as an error, as it is expected until the init is mined.

client/asset/eth/nodeclient.go
```diff
@@ -69,10 +69,11 @@ type nodeClient struct {
 }
 
 func newNodeClient(dir string, net dex.Network, log dex.Logger) (*nodeClient, error) {
+	lggr := dex.StdOutLogger("node", dex.LevelError)
 	node, err := prepareNode(&nodeConfig{
 		net:    net,
 		appDir: dir,
-		logger: log.SubLogger("NODE"),
+		logger: lggr,
```
JoeGruffins, line 76: The eth node is really loud. This is not correct, but ok if we set it to error by default?

server/swap/swap.go
```diff
@@ -1550,7 +1550,7 @@ func (s *Swapper) processRedeem(msg *msgjson.Message, params *msgjson.Redeem, st
 	matchID := match.ID()
 	chain := stepInfo.asset.Backend
 	if !chain.ValidateSecret(params.Secret, cpContract) {
-		log.Errorf("Secret validation failed (match id=%v, maker=%v, secret=%x)",
+		log.Infof("Secret validation failed (match id=%v, maker=%v, secret=%v)",
 			matchID, actor.isMaker, params.Secret)
```
JoeGruffins, line 1554: %v prints hex and I think this is info. It is the fault of the user if the redemption secret is incorrect.

JoeGruffins:
Also, starting the node from seed every time isn't really making things faster. As commented in the nodeclient harness tests, the node has trouble figuring out the correct account nonce it seems, and always starts with zero, even if it cannot send txs using zero. Maybe there is a way fix for this, but often need a new eth harness if you really need to start over.

Also this does not touch refunds which may not work yet.

client/asset/eth/eth.go
JoeGruffins: I'm not sure how to fix this. For one there is #1336 which is still my opinion. Just use the user's max fee. There is also #1372 which is a book.

I think I'll leave it in this unfinished state for this pr and let that get solved in another pr.

client/asset/eth/eth.go
JoeGruffins: Changed it to the tx hashes.

client/asset/eth/eth.go
JoeGruffins: sending the serialized tx

client/asset/eth/nodeclient.go
JoeGruffins: Using "ITNL" for internal as a key. Will add something to the dexc config options I guess. We could just accept the level map string which would look like "key=lvl,key=lvl,...."

JoeGruffins: Added the beginnings of a way to control this per logger.

martonp (approved):
Looks good, I was able to successfully swap dcr<->eth on simnet.

client/asset/eth/eth.go
buck54321: I'm of the opinion that the server should just set the `Match.{FeeRateBase,FeeRateQuote}` to the `(dex.Asset).MaxFeeRate` for eth at match time. The current base fee rate is not relevant to the validation of the transaction. In theory, the client could just log the `MaxFeeRate` at order time and persist that info and use it as an input to the `Swap` method. But I don't see how that makes things easier for anyone. Instead, we can just think of the `Match.{FeeRateBase,FeeRateQuote}` as the "validation rate", or the rate that we are going to validate in their transaction. This fits into our current architecture without changes, and the relevant values are persisted.

chappjc: Agree completely with that @buck54321.  Ref, your comment in my issue on the matter: https://github.com/decred/dcrdex/issues/1372#issuecomment-1002273296

client/cmd/dexc/log.go
```diff
@@ -47,6 +47,20 @@ func initLogging(lvl string, utc bool) *dex.LoggerMaker {
...
+	var (
+		lvls      string
+		delimiter string
+	)
+	for k, v := range defaultLogLevelMap {
+		lvls = fmt.Sprintf("%s%s%s=%s", lvls, delimiter, k, v)
+		if delimiter == "" {
+			delimiter = ","
+		}
+	}
```
buck54321, line 59: This works, but you should feel free to add a method to the `LoggerMaker` so you can just pass the `defaultLogLevelMap` too..

client/core/trade.go
```diff
@@ -1007,8 +1011,10 @@ func (t *trackedTrade) isSwappable(ctx context.Context, match *matchTracker) boo
 			coinIDString(wallet.AssetID, match.MetaData.Proof.MakerSwap), unbip(wallet.AssetID))
 		confs, spent, err := wallet.SwapConfirmations(ctx, match.MetaData.Proof.MakerSwap,
 			match.MetaData.Proof.ContractData, match.MetaData.Stamp)
-		if err != nil {
-			t.dc.log.Errorf("error getting confirmation for our own swap transaction: %v", err)
+		if err != nil && !errors.Is(err, asset.ErrSwapNotInitiated) {
+			// No need to log an error if swap not initiated as this
+			// is expected for newly made swaps involving contracts.
+			t.dc.log.Errorf("isSwappable: error getting confirmation for our own swap transaction: %v", err)
 		}
```
buck54321, line 1018: This is closely related to #1425, and `ErrSwapNotInitiated` is analogous to `CoinNotFoundError`, as far as I can tell. There was [some discussion on this on Element too](https://app.element.io/#/room/!SFRQQFIHUUNXARfvew:decred.org/$RcebZlyaATD7J8ZtuI-PqAurABg2VFcdGTgwzBYKbGs). On that end, chappjc decided to just return `nil` error and zero confs in that case. Is there a strong argument to do it differently here?

buck54321 (approved):
Brilliant! My first swap with eth. Do whatever with the `ErrSwapNotInitiated` I guess, but we need a comprehensive plan there.

client/core/trade.go
chappjc: Just want to reiterate that https://github.com/decred/dcrdex/pull/1425 is just a stopgap for v0.4, and if we do keep it `nil` when the swap isn't found (either txn with utxo assets or an entry in the contracts map for eth), I suspect we'll want a `spent bool` or similar from `SwapConfirmations`, with an error reserved for the case where it is actually unexpected to not find it, but of course when unconfirmed it is expected (proper error semantics).

chappjc: meant to say **found** above, not spent, sorry

JoeGruffins: I am trying to avoid returning a false positive, I guess. Not finding the swap could be a problem or it might not be. It's not that there is absolutely no problem, there may be a problem. Does that make sense? We could use the initiation tx here instead and confidently return `CoinNotFoundError` like the server does.

client/asset/eth/eth.go
JoeGruffins: Will leave for not if that's ok.

client/cmd/dexc/log.go
JoeGruffins: Ok.

client/core/trade.go
JoeGruffins: Maybe it would be more correct to log as warn?

chappjc: It depends on (a) the capabilities of the wallet backend and (b) time and blocks that have elapsed.  I've beat this horse to death now, but I am adamantly opposed to logging a warning or error if it's normal for the wallet not to find the unconfirmed swap.

chappjc: No need to change this now, but I think the ultimate solution involves more work in Core to be intelligent about gauging severity.

client/cmd/dexc/log.go
JoeGruffins: Yes, this makes more sense.

client/asset/eth/eth.go
chappjc: OK, when we implement the changes discussed here and https://github.com/decred/dcrdex/issues/1372#issuecomment-1002273296 we will switch this back to `swaps.FeeRate`

server/swap/swap.go
chappjc: Good call on both accounts.  I fell into the %x-with-a-Stringer pitfall that I like to remind others about.

chappjc (approved):
Good to go after popping off the first two commits, rebasing and and squashing the last two.

PR #1417, opened 2022-01-14
ui: Fix wallet wait form issues

martonp:
- ui: Add a line on the wallet wait form that displays
  tells the user to send enough funds to cover tx fee in
  addition to the registration fee
- ui: Fix a bugs where balance updates for different assets
  were displayed on the wallet wait form, and one where
  if wallet was already synced when page loaded, but fee
  not yet paid, the confirm registration page would not load

This contains the same bug fixes as #1401, but adds a generic instruction to send enough to cover tx fees instead of trying to estimate how much they will be.

### DIFF
--- a/client/webserver/site/src/js/forms.js
+++ b/client/webserver/site/src/js/forms.js
@@ -623,13 +623,17 @@ export class WalletWaitForm {
     page.fee.textContent = Doc.formatCoinValue(fee.amount, asset.info.unitinfo)
 
     Doc.hide(page.syncUncheck, page.syncCheck, page.balUncheck, page.balCheck, page.syncRemainBox)
-    Doc.show(page.balanceBox)
+    Doc.show(page.balanceBox, page.sendEnough)
 
     Doc.show(wallet.synced ? page.syncCheck : wallet.syncProgress >= 1 ? page.syncSpinner : page.syncUncheck)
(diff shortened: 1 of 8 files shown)


client/webserver/site/src/js/forms.js
```diff
@@ -640,27 +644,27 @@ export class WalletWaitForm {
...
-  reportBalance (bal) {
-    if (this.funded || this.assetID === -1) return
+  reportBalance (bal, assetID) {
+    if (this.funded || this.assetID === -1 || this.assetID !== assetID) return
     const page = this.page
     const asset = app().assets[this.assetID]
-    const fee = this.regFee
 
-    if (bal.available <= fee.amount) {
+    if (bal.available <= this.regFee.amount) {
```
chappjc, line 659: Feel like it's ok to check the box when `avail == amt` rather than requiring a little more?  I know this doesn't generalize to tokens, but I feel like we'd catch a ton of use mistakes here that would lead to exact deposits and then it failing in `Register` on the backend.

martonp: The `<=` here leads to an early return statement. It will only check the box if the balance is greater than the registration fee.

chappjc: Ahh, read the flow backwards. Nice, thanks.

chappjc (approved):

PR #1418, opened 2022-01-14
client/asset,core: SwapConfirmations tx not found handling

chappjc:
Replacing https://github.com/decred/dcrdex/pull/1403. Not tested.
I find this solution a bit less palatable, but at least `SwapConfirmations` consistently returns `asset.CoinNotFound` for all wallet types when the swap cannot be located.

client/asset/btc: Also return `CoinNotFoundError` for SPV wallets when they cannot locate the transaction.  If the wallet is **not** SPV, log a warning when `SwapConfirmations` has this specific error.

client/asset/dcr: If the wallet is **not** SPV log a warning in `SwapConfirmations` when the result is `CoinNotFoundError`.

client/core: In `counterPartyConfirms`, do not log an error from `SwapConfirmations` if it is `CoinNotFoundError`.

Allow `CoinNotFoundError` to log an error when checking own/self swap.

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -2638,7 +2638,13 @@ func (btc *ExchangeWallet) SwapConfirmations(_ context.Context, id dex.Bytes, co
 	if err != nil {
 		return 0, false, err
 	}
-	return btc.node.swapConfirmations(txHash, vout, pkScript, startTime)
+	confs, spent, err := btc.node.swapConfirmations(txHash, vout, pkScript, startTime)
+	if _, isRPC := btc.node.(*rpcClient); isRPC && errors.Is(err, asset.CoinNotFoundError) {
+		// We'll assume caller is silent for this error, but it is a bad
+		// indication for full node wallets, so let's make extra noise here.
+		btc.log.Warnf("Failed to find swap transaction %v.", txHash)
+	}
+	return confs, spent, err
 }
 
 // RegFeeConfirmations gets the number of confirmations for the specified output
--- a/client/core/trade.go
+++ b/client/core/trade.go
@@ -793,8 +793,14 @@ func (t *trackedTrade) counterPartyConfirms(ctx context.Context, match *matchTra
 	have, spent, err = t.wallets.toWallet.SwapConfirmations(ctx, coin.ID(),
 		match.MetaData.Proof.CounterContract, match.MetaData.Stamp)
 	if err != nil {
-		t.dc.log.Errorf("Failed to get confirmations of the counter-party's swap %s (%s) for match %s, order %v: %v",
-			coin, t.wallets.toAsset.Symbol, match, t.UID(), err)
+		if !errors.Is(err, asset.CoinNotFoundError) {
+			// CoinNotFoundError may be expected for foreign transactions until
+			// it is mined. The wallet implementation may choose to log this if
+			// it has the capability to find unconfirmed external transactions.
+			t.dc.log.Errorf("Failed to get confirmations of the counter-party's swap "+
+				"%s (%s) for match %s, order %v: %v",
+				coin, t.wallets.toAsset.Symbol, match, t.UID(), err)
+		}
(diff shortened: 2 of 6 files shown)


client/asset/btc/btc.go
```diff
@@ -2638,7 +2638,13 @@ func (btc *ExchangeWallet) SwapConfirmations(_ context.Context, id dex.Bytes, co
...
 		return 0, false, err
 	}
-	return btc.node.swapConfirmations(txHash, vout, pkScript, startTime)
+	confs, spent, err := btc.node.swapConfirmations(txHash, vout, pkScript, startTime)
+	if _, isRPC := btc.node.(*rpcClient); isRPC && errors.Is(err, asset.CoinNotFoundError) {
+		// We'll assume caller is silent for this error, but it is a bad
+		// indication for full node wallets, so let's make extra noise here.
+		btc.log.Warnf("Failed to find swap transaction %v.", txHash)
+	}
+	return confs, spent, err
```
chappjc, line 2647: This is what bugs me about this approach of logging in the asset backend.  Logging an error or a warning is strange because we are about to return an error, so the question is why doesn't the error just convey all the needed info.

This leads me to other ideas for work on the next release cycle:
- `asset.UnexpectedCoinNotFoundError` (or `asset.ExpectedCoinNotFoundError`) vs `asset.CoinNotFoundError`
-  return a nil error for not found scenarios, and instead have a -1 confs or a `found bool` return

client/core/trade.go
```diff
@@ -1007,11 +1013,11 @@ func (t *trackedTrade) isSwappable(ctx context.Context, match *matchTracker) boo
 			coinIDString(wallet.AssetID, match.MetaData.Proof.MakerSwap), unbip(wallet.AssetID))
 		confs, spent, err := wallet.SwapConfirmations(ctx, match.MetaData.Proof.MakerSwap,
 			match.MetaData.Proof.ContractData, match.MetaData.Stamp)
-		if err != nil {
-			t.dc.log.Errorf("error getting confirmation for our own swap transaction: %v", err)
+		if err != nil { // unexpected error includes asset.CoinNotFoundError because this is our swap.
+			t.dc.log.Errorf("Unexpected error getting confirmation for our own swap transaction: %v", err)
```
chappjc, line 1017: OK, this isn't working as expected.  The BTC SPV wallet cannot find _it's own swap txn_, unlike dcrwallet that does.

```
2022-01-15 02:44:33.381 [TRC] CORE: Checking confirmations on our OWN swap txn b61818b246758e7c745047200974f78efbdcbefe790dcc0aec8ebdbd907b7f1a:0 (btc)...
2022-01-15 02:44:33.381 [TRC] CORE[btc][SPV]: swapConfirmations - SpentStatusUnknown: b61818b246758e7c745047200974f78efbdcbefe790dcc0aec8ebdbd907b7f1a:0 (block <nil>, confs 0)
2022-01-15 02:44:33.381 [TRC] CORE[btc][SPV]: swapConfirmations - scanFilters: b61818b246758e7c745047200974f78efbdcbefe790dcc0aec8ebdbd907b7f1a:0 (block <nil>, start time 2022-01-15 02:43:10 +0000 UTC)
2022-01-15 02:44:33.381 [DBG] CORE[btc][SPV]: Performing cfilters scan for b61818b246758e7c745047200974f78efbdcbefe790dcc0aec8ebdbd907b7f1a:0 from height 2135929
2022-01-15 02:44:33.381 [TRC] CORE[btc][SPV]: swapConfirmations - scanFilters did not find b61818b246758e7c745047200974f78efbdcbefe790dcc0aec8ebdbd907b7f1a:0, assuming in mempool.
2022-01-15 02:44:33.381 [ERR] CORE: Unexpected error getting confirmation for our own swap transaction: coin not found
```

Unless an alternate working solution goes up that does not modify huge amounts of code and APIs appears, I'm merging https://github.com/decred/dcrdex/pull/1403

PR #1419, opened 2022-01-14
client/ui: Display registration tx fee estimate

martonp:
- `client/core`: Add `EstimateRegistrationTxFee` function in core that determines the tx fee needed to register for the dex
- `client/asset`: Add `EstimateRegistrationTxFee` to wallet interface
- `client/webserver`: Add a new api endpoint `/regtxfee`
- `ui`: Each time the WalletWaitForm is loaded, this endpoint is called to retrieve an estimate

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -2573,6 +2573,34 @@ func (btc *ExchangeWallet) PayFee(address string, regFee, feeRateSuggestion uint
 	return newOutput(txHash, vout, sent), nil
 }
 
+// feeRateWithOracleFallback attempts to retrieve a fee rate from the wallet
+// but falls back to the oracle function if it fails.
+func (btc *ExchangeWallet) feeRateWithOracleFallback(confTarget uint64, oracle func() uint64) uint64 {
+	feeRate, err := btc.estimateFee(btc.node, confTarget)
+	if err == nil {
+		btc.log.Tracef("Obtained local estimate for %d-conf fee rate, %d", confTarget, feeRate)
+		return feeRate
+	}
+	feeSuggestion := oracle()
+	if feeSuggestion > 0 && feeSuggestion < btc.fallbackFeeRate && feeSuggestion < btc.feeRateLimit {
+		btc.log.Tracef("feeRateWithFallback using caller's suggestion for %d-conf fee rate, %d. Local estimate unavailable (%q)",
+			confTarget, feeSuggestion, err)
+		return feeSuggestion
+	}
+	btc.log.Warnf("Unable to get optimal fee rate, using fallback of %d: %v",
+		btc.fallbackFeeRate, err)
+	return btc.fallbackFeeRate
+}
+
+// EstimateRegistrationTxFee provides an estimate for the tx fee needed to
+// pay the registration fee. rateOracleFallback is used in the case where
+// the wallet fails to determine the current fee rate on its own.
+func (btc *ExchangeWallet) EstimateRegistrationTxFee(rateOracleFallback func() uint64) uint64 {
+	const inputCount = 5 // buffer so this estimate is higher than what PayFee uses
+	return (dexbtc.MinimumTxOverhead + 2*dexbtc.P2PKHOutputSize + inputCount*dexbtc.RedeemP2PKHInputSize) *
+		btc.feeRateWithOracleFallback(1, rateOracleFallback)
+}
+
 // Withdraw withdraws funds to the specified address. Fees are subtracted from
 // the value. feeRate is in units of atoms/byte.
 func (btc *ExchangeWallet) Withdraw(address string, value, feeSuggestion uint64) (asset.Coin, error) {
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -2582,7 +2582,7 @@ func (btc *ExchangeWallet) feeRateWithOracleFallback(confTarget uint64, oracle f
 		return feeRate
 	}
 	feeSuggestion := oracle()
-	if feeSuggestion > 0 && feeSuggestion < btc.fallbackFeeRate && feeSuggestion < btc.feeRateLimit {
+	if feeSuggestion > 0 && feeSuggestion < btc.feeRateLimit {
 		btc.log.Tracef("feeRateWithFallback using caller's suggestion for %d-conf fee rate, %d. Local estimate unavailable (%q)",
 			confTarget, feeSuggestion, err)
 		return feeSuggestion
--- a/client/asset/btc/btc_test.go
+++ b/client/asset/btc/btc_test.go
@@ -10,6 +10,7 @@ import (
 	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"math/rand"
 	"os"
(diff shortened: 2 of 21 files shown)


client/asset/btc/btc.go
```diff
@@ -2573,6 +2573,34 @@ func (btc *ExchangeWallet) PayFee(address string, regFee, feeRateSuggestion uint
...
+// feeRateWithOracleFallback attempts to retrieve a fee rate from the wallet
+// but falls back to the oracle function if it fails.
+func (btc *ExchangeWallet) feeRateWithOracleFallback(confTarget uint64, oracle func() uint64) uint64 {
+	feeRate, err := btc.estimateFee(btc.node, confTarget)
+	if err == nil {
+		btc.log.Tracef("Obtained local estimate for %d-conf fee rate, %d", confTarget, feeRate)
+		return feeRate
+	}
+	feeSuggestion := oracle()
+	if feeSuggestion > 0 && feeSuggestion < btc.fallbackFeeRate && feeSuggestion < btc.feeRateLimit {
```
JoeGruffins, line 2585: Why is `feeSuggestion < btc.fallbackFeeRate` a requirement? Not saying it's wrong, but asking for my understanding. If the network is busy it could be higher. As long as it isn't over `btc.feeRateLimit` I would think it's fine.

client/asset/btc/btc.go
```diff
@@ -2573,6 +2573,34 @@ func (btc *ExchangeWallet) PayFee(address string, regFee, feeRateSuggestion uint
...
+		return feeSuggestion
+	}
+	btc.log.Warnf("Unable to get optimal fee rate, using fallback of %d: %v",
+		btc.fallbackFeeRate, err)
+	return btc.fallbackFeeRate
+}
+
+// EstimateRegistrationTxFee provides an estimate for the tx fee needed to
+// pay the registration fee. rateOracleFallback is used in the case where
+// the wallet fails to determine the current fee rate on its own.
```
JoeGruffins, line 2597: `rateOracleFallback` the method is used regardless. Maybe just make it words i.e. `A rate oracle fallback is used ...` Here and in a few other places.

JoeGruffins (commented):
Looks and works well.

client/asset/btc/btc.go
martonp: Actually I'm not sure why either, but this is the logic in the existing code (`feeRateWithFallback`) as well. Does anyone else know why? @chappjc @buck54321

chappjc: Hmm, not really.  I would think that `feeRateLimit` would be adequate to compare.

client/asset/btc/btc.go
JoeGruffins: You are right, this is the name of the argument, I was reading it as `feeRateWithOracleFallback`.

JoeGruffins (approved):

buck54321:
Good stuff. I'm hoping #1430 can inform this as well. Part of the motivation in #1430 is to avoid the 'fee_rate' request when the wallet can provide one. I think that's part of your motivation here with the `rateOracleFallback func() uint64` argument too. Using a `feeSuggestion uint64` argument, even if it's procured from the wallet itself, is a better fit for our current use patterns and workflow, and should result in less code here. The `feeRateWithOracleFallback` methods seem to be doing some of the same logic as `(*Core).feeSuggestionAny`, so we can avoid some duplication and keep server comms as a *Core concern.

The idea would then be `(*ExchangeWallet).EstimateRegistrationTxFee(feeSuggestion uint64) uint64`, where the `feeSuggestion` was just procured from 1) the wallet itself if it implements `FeeRater` (primed full nodes, eth), 2) any connected server (multi-server config), or 3) a temporary connection to a server being registered (first-time users using spv wallet).

chappjc:
I think the thing is, @buck54321, that with many Wallet methods, the `feeSuggestion` argument is used as a fallback if internal fee rate estimation is not possible, so while it generally makes sense how you've put `rater.FeeRate()` into `feeSuggestionAny` in your other PR, what happens for certain wallets that are `FeeRaters` is that after you get a fee rate from the wallet when calling `feeSuggestionAny`, the method you call e.g. `Withdraw` again gets a fee rate from the wallet, meaning double RPCs often.  The oracle function would seem to solve that problem by providing a fallback that only incurs a cost of the wallet itself doesn't have a fee rate.  I'm not overly delighted with the oracle function although I had suggested that approach because of it's role as a fallback.  I'm very open to another solution, but unless I'm mistaken we'll be querying wallet twice back-to-back in many cases.

martonp:
If the functions such as `Withdraw` and `PayFee` are rewritten such that the `FeeSuggestion` parameter isn't a suggestion but the fee rate that must be used, then there wouldn't be double calls.
Personally I like the `rateOracleFallback` method slightly better because it seems a little strange to query something from the wallet only to pass it right back. Unless `core` will do something with the fee rates other than passing it right back to the wallet, what's the point of `core` knowing which wallets can and cannot get the fee rates? For me it's not a big deal either way though, whatever you guys prefer.

chappjc:
A related idea would be to provide a `rateOracleFallback` function on Wallet construction, and we could remove all the `feeSuggestion` arguments.  The wallet methods would still internally attempt to get a fee rate if they are able, and fallback to the oracle function.

My reservation with that and any oracle that makes a server request is that I don't really want a server request in just any Wallet method call, although not because it's crossing a line, but because of potential latency or even timeouts.  Otherwise I like an oracle function for avoiding the cost of getting any fallback/suggestion rate in the first place if wallet itself can get it.

If the oracle didn't involve a server request (only use the cached rates from the book) and/or it only made such requests from less time-sensitive methods, I'd be completely content with it.

buck54321:
> meaning double RPCs often.

> If the functions such as Withdraw and PayFee are rewritten such that the FeeSuggestion parameter isn't a suggestion but the fee rate that must be used, then there wouldn't be double calls.

Yes. Thank you, @martonp. It'll be the same amount of work for the `Wallet` without fewer `fee_rate` requests. in **core**. From the wallet's perspective, the `feeSuggestion` is either 1) a rate the wallet just provided, in which case no need to get a new one, or 2) a rate fetched from the server, because we can't provide one, again no need to try for a new one. As @martonp mentioned, it would no longer be a `feeSuggestion` as much as just a `feeRate` to use. 

In totality, both approaches result in the same number of wallet estimates and `fee_rate` requests, but

1) The `feeSuggestion`  is used identically by `PayFee`, `Withdraw`, and `EstimateRegistrationTxFee` (and future `Send`/`Sweep`). We should use a consistent approach for all three.
2) The oracle approach makes me uncomfortable. I would typically frame my objection around separation of concerns, which is the "line" I think @chappjc mentioned. He pointed to timeouts specifically, which I agree with. By keeping server comms in **core**, we can properly handle that error. It's also just generally easier to reason about from a development perspective, imo.

buck54321:
> 1. The `feeSuggestion`  is used identically by `PayFee`, `Withdraw`, and `EstimateRegistrationTxFee` (and future `Send`/`Sweep`). We should use a consistent approach for all three.

Regardless of which approach we use, once we're consistent we can remove the calls to `feeRateWithFallback` in those methods.

chappjc:
I guess I'm ok with having the argument be a requested rate instead of a fallback, but because core gets that rate from multiple sources, starting with the wallet itself, I think that's mixing up concerns a bit tbh.  With the fallback approach (either int or oracle function), the primary rate from wallet is gotten by wallet itself, and the fallback is gotten by core from book or server, so to me that approach still makes the most sense.  I don't actually seem much difference between an int and a function at the end of the day, so I really don't object to the approach on the grounds of separation of concerns.

My only hang up is that whatever we do we have to ensure `Swap`, `Redeem`, and `Refund` calls are not delayed by a server request in either core _or_ the oracle

martonp:
With `Redeem` we are currently not doing any server calls, I think that one can stay the same. With `Refund` we're currently using the `RequestWithTimeout` method in `fetchFeeRate`, and the same thing can be done in the refactor as well. All we're doing is avoiding unnecessary calls, there will be no case where there are more calls to the server than now.

buck54321:
If you're set on doing a function argument, go for it. If so, please update `PayFee` and `Withdraw` to use an oracle as well. Or I can open an issue, but we want to avoid extraneous calls for those methods too. I still think it's silly to implement `feeRateWithOracleFallback` in individual asset packages instead of leveraging the existing code in **core** and the existing patterns in `Wallet`, but I've registered my objection and I've already given this matter more attention than warranted. It'll be fine your way.

chappjc:
I'm not set on any one way, but I hate expanding scope to fix something with very narrow scope.  @martonp pls see how it works out to do everything in core and pass a demanded rate (although assets backends still need to apply their limits).  It's absolutely worth feeling out.

I've been commenting because I'm not at all offended by some methods working differently to suite their needs, I see a fallback _function_ instead of an int as a simple way to avoid incurring any cost unnecessarily, and there is not a need to shake up the rest of the interface.  I wish we could just solve this one issue of the registration fee, but if there's a change that improves everything over all, alright.  Personally, I see a solved issue with no particularly troublesome patterns introduced.

martonp:
#1444 is the new PR with the same functionality that builds on top of #1430.

martonp:
Closing in favor of #1444.

PR #1420, opened 2022-01-16
i18n: add pl-PL translations

artikozel:
This adds Polish (PL) translations, tag pl-PL, done by @artikozel.


artikozel:
was all kinds of wrong

PR #1425, opened 2022-01-18
[release-v0.4] client/asset/dcr: coin not found not an error for spv

chappjc:
Cherry-pick of https://github.com/decred/dcrdex/pull/1403 for `release-v0.4`

Do not return asset.CoinNotFoundError from SwapConfirmations with an
SPV wallet since this is normal for mempool transactions in SPV.

For a non-SPV wallet, if the swap appears spent but it cannot be
located in a block with a cfilters scan, this will return
asset.CoinNotFoundError. For SPV wallets, it is not an error if the
transaction cannot be located SPV wallets cannot see non-wallet
transactions until they are mined.

Fix a few logs where arguments were orderd strangely.

Also silence a warning that happens each time a transaction is
found in a block while looking for a spending transaction immediately
afterward.

### DIFF
--- a/client/asset/btc/spv.go
+++ b/client/asset/btc/spv.go
@@ -957,6 +957,8 @@ func (w *spvWallet) swapConfirmations(txHash *chainhash.Hash, vout uint32, pkScr
 		if assumedMempool {
 			w.log.Tracef("swapConfirmations - scanFilters did not find %v:%d, assuming in mempool.",
 				txHash, vout)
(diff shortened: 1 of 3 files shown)


buck54321 (approved):
We had some discussion on Element, and this is a good short-term solution to quiet the extraneous logging, with an eye towards a more sophisticated solution that will involve a change to the Wallet API.

PR #1426, opened 2022-01-19
eth swap contract v1

buck54321:
I was just playing with this trying to minimize fees. I think it's sound, but there are some tradeoffs and some new edges that need exploring. Leaving in draft for the foreseeable future and focusing on token stuff.

```
Don't store contract data on-chain, just commit to it in the key of
the swaps map, which maps to a "swap record", 32 bytes in length,
with a special interpretation scheme that allows us to track the
swap status. This reduces fees by a factor of ~2 for a single-match
round-trip trade, which would be similar to e.g. ETH-USDC or other
token trade. For multi-match trades, the savings are even better,
asymptoting at ~28% of the v0 fees. A single-match round trip trade
uses ~ 93k gas, which is notably less than a trade on UniSwap and
other Ethereum trading platforms which are typically > 135k, I think.
```


dex/networks/eth/contracts/ETHSwapV1.sol
```diff
@@ -0,0 +1,221 @@
...
+            Redemption calldata r = redemptions[i];
+
+            require(r.c.participant == msg.sender, "not authed");
+
+            (bytes32 k, bytes32 record, uint256 blockNum) = retrieveRecord(r.c);
+
+            // To be redeemable, the record needs to represent a valid block
+            // number.
+            require(blockNum > 0 && blockNum < block.number, "unfilled swap");
+            // require(blockNum >> 32 == 0, "invalid swap format");
```
JoeGruffins, line 174: Commented out, but does ethereum have this limitation of a uint32 block number?

JoeGruffins (commented):
Some comments on the contract.

dex/networks/eth/contracts/ETHSwapV1.sol
```diff
@@ -0,0 +1,221 @@
...
+    {
+        uint initVal = 0;
+        for (uint i = 0; i < contracts.length; i++) {
+            Contract calldata c = contracts[i];
+
+            require(c.value > 0, "0 val");
+            require(c.refundTimestamp > 0, "0 refundTimestamp");
+
+            // Is this needed?
+            // require(msg.sender == c.initiator)
```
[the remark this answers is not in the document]
JoeGruffins: This got me thinking so I made #1427

dex/networks/eth/contracts/ETHSwapV1.sol
buck54321: No, but it'll take a couple hundred years to get to that block height. This comment is left over from an earlier iteration on the record scheme, but `blockNum < block.number` is better.

dex/networks/eth/contracts/ETHSwapV1.sol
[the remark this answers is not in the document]
buck54321: Yeah, it doesn't seem strictly necessary to me. I see this as limiting to 1) third-party services, and 2) privacy-focused clients.

dex/networks/eth/contracts/ETHSwapV1.sol
```diff
@@ -0,0 +1,221 @@
...
+        uint64 refundTimestamp;
+        address participant;
+        uint64 value;
+    }
+
+    // contractKey generates a key hash which commits to the contract data. The
+    // generated hash is used as a key in the swaps map.
+    function contractKey(Contract calldata c) public pure returns (bytes32) {
+        return sha256(abi.encodePacked(c.secretHash, c.initiator, c.participant, c.value, c.refundTimestamp));
+    }
```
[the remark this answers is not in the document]
buck54321: > With version zero you could look at all the swaps and probably find your address as initiator with status refundable and wouldn't need the secret anymore.

Can you though? Swaps are keyed by secret hash, and there is no getter for all the swaps. If you nuke your db during settlement, aren't all bets are off? I feel like this is a just a general limitation of p2sh contracts that involve two parties.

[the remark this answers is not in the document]
JoeGruffins: > there is no getter for all the swaps

I think you can get all the swaps, it's public. I know I did in the begging while testing.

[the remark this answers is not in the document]
JoeGruffins: I think this used to return all the swaps https://github.com/decred/dcrdex/blob/4548c2ba32c9706250efc8c38b9b3403aa41cf33/dex/networks/eth/contracts/v0/contract.go#L335-L338

But looks like it doesn't now?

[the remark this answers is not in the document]
JoeGruffins: Maybe we don't even need the Swap getter now?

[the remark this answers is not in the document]
JoeGruffins: Oh, I think you're right and I'm miss-remembering: https://ethereum.stackexchange.com/questions/15337/can-we-get-all-elements-stored-in-a-mapping-in-the-contract

So, one wouldn't be able to find their swap without the key/secret hash

client/asset/eth/contractor.go
```diff
@@ -262,6 +277,151 @@ func (c *contractorV0) incomingValue(ctx context.Context, tx *types.Transaction)
...
+	for _, r := range redeems {
+		var initiator common.Address
+		copy(initiator[:], r.Spends.Coin.ID())
+		versionedRedemptions = append(versionedRedemptions, RedemptionToV1(initiator, c.acctAddr, r))
+	}
+	return c.Redeem(txOpts, versionedRedemptions)
+}
+
+// This one breaks contractor, since we need the entire *asset.Contract, not just
+// the secret hash.
```
chappjc, line 333: There are quite a few broken like this, including `swap`, which is used all over. e.g. https://github.com/decred/dcrdex/pull/1638/files#diff-a45b0f60d5cf2f6825eb9e9ed1f173fbf1df9eb53b09914ce6b17e42074653e5R101

I'm confused about how we're going to adapt if we can't look up the `Swap struct` by secret hash any more.

client/asset/interface.go
```diff
@@ -326,7 +330,7 @@ type Wallet interface {
 	// NOTE: This could potentially be a long and expensive operation if
 	// performed long after the swap is broadcast; might be better executed from
 	// a goroutine.
-	FindRedemption(ctx context.Context, coinID, contract dex.Bytes) (redemptionCoin, secret dex.Bytes, err error)
+	FindRedemption(ctx context.Context, coinID, contractData dex.Bytes, deets *dex.SwapContractDetails) (redemptionCoin, secret dex.Bytes, err error)
```
buck54321, line 333: I was on the fence about adding these arguments, because I think we could probably work it out without them, through the tx cache or something. But all of the info in the `SwapContractDetails` is required to be saved by the client already, so it actually seems more on-model to have `Core` supply it. Open to suggestions though.

server/asset/eth/tokener.go
```diff
@@ -44,11 +44,11 @@ func newTokener(ctx context.Context, assetID uint32, net dex.Network, be bind.Co
 		return nil, err
 	}
 
-	if token.ver != 0 {
+	if token.ver != 1 {
 		return nil, fmt.Errorf("only version 0 contracts supported")
 	}
 
-	es, err := erc20v0.NewERC20Swap(swapContract.Address, be)
+	es, err := erc20v1.NewERC20Swap(swapContract.Address, be)
```
buck54321, line 51: I had always intended for the server to only support a single version at a time, and I still do. But there's probably an argument for making it switchable via configuration. I don't know.

client/asset/interface.go
```diff
@@ -671,6 +676,8 @@ type Contract struct {
 type Redemption struct {
 	// Spends is the AuditInfo for the swap output being spent.
 	Spends *AuditInfo
+	// SwapDetails is the swap details.
+	SwapDetails *dex.SwapContractDetails
```
buck54321, line 680: I'd like to refactor this later. `AuditInfo` and  `SwapContractDetails` are identical except for an extra `Contract` field. But I didn't want to mess with the btc/dcr interfaces too much in the PR.

client/asset/eth/contractor.go
buck54321: > I'm confused about how we're going to adapt if we can't look up the Swap struct by secret hash any more.

We could do it with the `Swaps` and the `ContractKey` methods. We would just need to interpret ourselves. But I've added a little more detail to the return from state (renamed to status) to make things easier. Happy to go back if you're concerned with the contract change.

buck54321: > It also becomes super hard to use the refund-by-anyone path we are relying on as a backup. When we create the swap we can't a pre-signed refund, so we have been logging the secret hash so the user can just (e.g.) take that to etherscan and refund with metamask.

I'll look into a solution, but I don't see any reason that something like this should stop us.

server/asset/eth/coiner.go
```diff
@@ -197,78 +186,11 @@ func (be *AssetBackend) baseCoin(coinID []byte, contractData []byte) (*baseCoin,
...
-	}
-
-	// Uninitiated state is zero confs. It could still be in mempool.
-	// It is important to only trust confirmations according to the
-	// swap contract. Until there are confirmations we cannot be sure
-	// that initiation happened successfully.
-	if swap.State == dexeth.SSNone {
-		// Assume the tx still has a chance of being mined.
-		return 0, nil
-	}
```
buck54321, line 212: It wasn't quite clear to me why we wouldn't return `CoinNotFoundError` here. We have the latency queue running for `(Backend).Confirmations` in `(Swapper).processInit`.

server/asset/eth/coiner.go
```diff
@@ -197,78 +186,11 @@ func (be *AssetBackend) baseCoin(coinID []byte, contractData []byte) (*baseCoin,
 // and the same account and nonce, effectively voiding the transaction we
 // expected to be mined.
 func (c *swapCoin) Confirmations(ctx context.Context) (int64, error) {
-	swap, err := c.backend.node.swap(ctx, c.backend.assetID, c.secretHash)
-	if err != nil {
-		return -1, err
-	}
-
-	// Uninitiated state is zero confs. It could still be in mempool.
```
buck54321, line 205: I've moved these functions away from contract inspection and towards more transaction inspection. I think it simplifies things. Any reason I can't do that?

client/asset/eth/contractor.go
```diff
@@ -171,47 +187,47 @@ func (c *contractorV0) redeem(txOpts *bind.TransactOpts, redemptions []*asset.Re
 }
 
 // swap retrieves the swap info from the read-only swap method.
-func (c *contractorV0) swap(ctx context.Context, secretHash [32]byte) (*dexeth.SwapState, error) {
+func (c *contractorV0) status(ctx context.Context, contract *dex.SwapContractDetails) (step dexeth.SwapStep, secret [32]byte, blockNumber uint32, err error) {
```
chappjc, line 190: A preferable self-documenting design IMO would be to leave as many `contractorV0` methods unchanged as possible except for their name (e.g. `status` -> `statusImpl`) and have them wrapped with the more confusing function signature with the `SwapContractDetails` argument.

client/asset/ltc/ltc.go
```diff
@@ -119,6 +119,7 @@ var (
 			rpcWalletDefinition,
 			electrumWalletDefinition,
 		},
+		ProtocolVersions: []uint32{0},
```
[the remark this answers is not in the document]
chappjc: And what was `version`?

client/asset/eth/contractor.go
buck54321: mmm. OK.

client/asset/btc/btc.go
```diff
@@ -3319,12 +3320,8 @@ func (btc *baseWallet) AuditContract(coinID, contract, txData dex.Bytes, rebroad
 
 // LocktimeExpired returns true if the specified contract's locktime has
 // expired, making it possible to issue a Refund.
-func (btc *baseWallet) LocktimeExpired(_ context.Context, contract dex.Bytes) (bool, time.Time, error) {
-	_, _, locktime, _, err := dexbtc.ExtractSwapDetails(contract, btc.segwit, btc.chainParams)
-	if err != nil {
-		return false, time.Time{}, fmt.Errorf("error extracting contract locktime: %w", err)
-	}
-	contractExpiry := time.Unix(int64(locktime), 0).UTC()
+func (btc *baseWallet) LocktimeExpired(_ context.Context, deets *dex.SwapContractDetails) (bool, time.Time, error) {
```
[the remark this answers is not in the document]
chappjc: I mean it's way more than needed *for BTC*.

client/asset/ltc/ltc.go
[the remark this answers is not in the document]
chappjc: Whatever you prefer.  I was just confused that there was something else being tracked here.

client/core/trade.go
```diff
@@ -177,6 +177,55 @@ func (m *matchTracker) token() string {
 	return hex.EncodeToString(m.MatchID[:4])
 }
 
+func (t *trackedTrade) theirContract(m *matchTracker) *dex.SwapContractDetails {
```
chappjc, line 180: Deeply grateful that this function cannot error.

dex/networks/eth/contracts/ETHSwapV1.sol
```diff
@@ -0,0 +1,226 @@
...
+
+            require(c.value > 0, "0 val");
+            require(c.refundTimestamp > 0, "0 refundTimestamp");
+
+            bytes32 k = contractKey(c);
+            bytes32 record = swaps[k];
+            require(record == bytes32(0), "swap not empty");
+
+            record = bytes32(block.number);
+            require(!secretValidates(record, c.secretHash), "hash collision");
```
JoeGruffins, line 136: Although just as impossible as this collision, the secret can also not be `RefundRecord`. You could find out what that hashes to and do a straight != with the constant and record if you wanted to.

dex/networks/eth/contracts/ETHSwapV1.sol
```diff
@@ -0,0 +1,226 @@
...
+        senderIsOrigin()
+    {
+        // Is this contract even in a refundable state?
+        require(block.timestamp >= c.refundTimestamp, "locktime not expired");
+
+        // Retrieve the record.
+        (bytes32 k, bytes32 record, uint256 blockNum) = retrieveRecord(c);
+
+        // Is this swap initialized?
+        require(blockNum > 0 && blockNum <= block.number, "swap not active");
```
JoeGruffins, line 213: I think this makes the refunded check not reachable if refunded. It also stops most checks for is redeemed. Both of which are fine, but can remove the explicit refund check I guess, or move it and the redeem check up if the error is important.

dex/networks/eth/txdata.go
```diff
@@ -11,94 +11,50 @@ import (
...
-		return [32]byte{}, fmt.Errorf("contract version %v does not exist", contractVersion)
-	}
-
-	return txDataHandler.parseRefundData(calldata)
-}
+const (
+	initiateFuncName = "initiate"
+	redeemFuncName   = "redeem"
+	refundFuncName   = "refund"
+)
```
JoeGruffins, line 25: Maybe export these? they are used a bunch in the clients.

dex/networks/eth/txdata.go
```diff
@@ -206,3 +170,149 @@ func (t *txDataHandlerV0) parseRefundData(calldata []byte) ([32]byte, error) {
...
+	decoded, err := ParseCallData(calldata, ABIs[1])
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse call data: %v", err)
+	}
+	if decoded.Name != initiateFuncName {
+		return nil, fmt.Errorf("expected %v function but got %v", initiateFuncName, decoded.Name)
+	}
+	args := decoded.inputs
+	// Any difference in number of args and types than what we expect
+	// should be caught by parseCallData, but checking again anyway.
```
JoeGruffins, line 189: parseCallData -> ParseCallData

server/asset/eth/coiner.go
JoeGruffins: I wanted to ensure that this info comes from the contract and not our assumptions about the tx. Especially with the recent redemptions pr we may be looking at a bogus transaction. (I don't think this is used for redemptions, but maybe it will be some day.)

If you are sure we are not then fine, but I'm not sure we couldn't have a tx that passed our verification but does not do what we expect. **Our txdata parsing is basically a hack**, we had to copy paste from geth internals. I do not think it is intended to be done this way.

At least there needs to be another check besides confirmations that checks status is ok, as in it did what we expected onchain. But, I hope we can not trust the transaction except for the basic sanity check we were doing and start looking at the contract here.

server/asset/eth/coiner.go
JoeGruffins: I don't remember. `CoinNotFoundError` is probably correct.

server/asset/eth/tokener.go
JoeGruffins: Although it would seem crazy to us, a user will update in the middle of a swap. Or perhaps a redeem or refund that forgot about. It would be responsible to continue to support old versions and deprecate slowly.

server/asset/eth/coiner.go
chappjc: It's certainly not the eth way.  Things are meant to replaceable, and the focus is supposed to be on state change, and contract interactions.

I'd feel better about doing less tx inspection with the hacked call data parsing if possible.

I worry we're already dug in on something we shouldn't be doing as well.

server/asset/eth/coiner.go
chappjc: I believe it's because the contract state will not be reflecting txns in txpool.  As such, there is ambiguity here about whether it's really not found or if it is in fact in txpool.  What is the problem with returning `0, nil`?  That would be correct for a transaction not yet mined.

server/asset/eth/tokener.go
chappjc: I would expect the client would be able to refund or auto-redeem those on their own.

Slightly related, which I am still intending to address myself: https://github.com/decred/dcrdex/issues/1583

JoeGruffins: I want to add that we are continuously making changes with new wallets, and probably more, that add data to the db that makes downgrading impossible but does not error if you try. So, a user may just be really confused and out of luck if they accidentally upgraded with an old swap. They would need to restore from seed with an older version.

JoeGruffins: Oh yeah, we are handling both versions for the client. Please ignore my comment.

JoeGruffins: Oh, but yeah, I guess markets will just need to be suspended for makerlocktime to ensure no new initiations? And hopefully refunds will still be ok, I think ok. Should be ok.

client/asset/btc/btc.go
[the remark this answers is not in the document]
martonp: Why not stick all the data from `SwapContractDetails` into the `contract dex.Bytes`?

martonp (commented):
Looks good overall. I think the API would be more simple if all the `dex.SwapContractDetails` stuff was just in the contract bytes. All of the information in there is available when the contract bytes are created in `Swap`.

client/asset/btc/btc.go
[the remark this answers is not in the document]
chappjc: That sounds appealing to me too.

buck54321:
Closing in favor of #2038

PR #1431, opened 2022-01-20
client/{core,ui}: Fix unit info and asset config usage on markets page

### LINKED ISSUE #1428: eth lacks proper unitinfo
If you have an eth configured, you get an error when clicking the exchange to add

https://github.com/decred/dcrdex/blob/e42db6eeebf4a462ff84f567cbe6315eed1f90aa/client/webserver/site/src/js/forms.js#L476

```
 Uncaught (in promise) TypeError: can't access property "conventional", e is null
    r forms.js:476
    o forms.js:487
    i forms.js:536
    value forms.js:517
    e register.js:53
    Ut entry.js:1
    a entry.js:1
    zt entry.js:1
    zt entry.js:1
    dexAddrForm register.js:49
    n forms.js:888
    wt entry.js:1
```

How did you get past this @JoeGruffins ?

Note that I created the wallets first, then went back to /register to click this

buck54321:
```
We were expecting a SupportedAsset in some places where it wasn't guaranteed.
Update Application.unitInfo method to accept optional xc [core.Exchange] argument
from which to look for UnitInfo. Pass the arg in relevant contexts.
Use MarketsPage.market.{baseCfg,quoteCfg} [dex.Asset] instead of
{base,quote} [core.SupportedAsset] in book-related code. Fix omission in
(*Core).convertAssetInfo.
```

Resolves #1428

### DIFF
--- a/client/webserver/site/src/js/forms.js
+++ b/client/webserver/site/src/js/forms.js
@@ -480,9 +480,9 @@ export class FeeAssetSelectionForm {
       const marketTmpl = Doc.parseTemplate(marketNode)
 
       const baseAsset = xc.assets[mkt.baseid]
-      const baseUnitInfo = unitInfo(xc, mkt.baseid)
+      const baseUnitInfo = app().unitInfo(mkt.baseid, xc)
       const quoteAsset = xc.assets[mkt.quoteid]
-      const quoteUnitInfo = unitInfo(xc, mkt.quoteid)
+      const quoteUnitInfo = app().unitInfo(mkt.quoteid, xc)
 
       if (cFactor(baseUnitInfo) === 0 || cFactor(quoteUnitInfo) === 0) return null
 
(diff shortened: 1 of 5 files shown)


client/webserver/site/src/js/forms.js
```diff
@@ -1003,11 +1003,3 @@ export function bind (form, submitBttn, handler) {
 function isTruthyString (s) {
   return s === '1' || s.toLowerCase() === 'true'
 }
-
-function unitInfo (xc, assetID) {
-  const dexAsset = xc.assets[assetID]
-  if (dexAsset && dexAsset.unitInfo.conventional.conversionFactor > 0) return dexAsset.unitInfo
```
chappjc, line 1009: What was this >0 for?  That's a funny check.

chappjc (approved):
Very clean and working well.

Note the ETH market:

image

All other dialogs and forms still working as expected.

client/webserver/site/src/js/forms.js
buck54321: The `UnitInfo` is a value instead of a pointer in many places, notable `dex.Asset`. Similarly, the `Conventional Denomination` field is a value. So I was thinking it wouldn't be crazy for an incorrect configuration to have a `conversionFactor` of zero. I'm not checking for the condition in the combined `Application.unitInfo` method though.

PR #1439, opened 2022-01-22
client/ui/btc: Configure wallet bday

### LINKED ISSUE #1270: btc spv: don't use hard-coded wallet birthday unless restoring
A lot of the initial wallet sync is spent doing "restoration" = requesting and scanning filters from the hard-coded `walletBirthday`. But if the seed is being freshly generated, we don't need to restore. **core** will need to inform the wallet whether it is restoring from seed so that we can set the birthday appropriately.

We also don't know at wallet creation time whether the app seed was restored or created fresh. We'll need to persist that information in the client database app bucket.

martonp:
The wallet birthday for the bitcoin SPV wallet was previously
always set to a hardcoded value. This diff stores the time when
the app seed was generated in the db to give an earliest possible
time so scan from, and also adds options on the UI to specify the
desired wallet birthday. Also, the wallet birthday can now be updated
when reconfiguring the wallet

Closes #1270.

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -117,12 +117,15 @@ var (
 		Key:         "walletbirthday",
 		DisplayName: "Wallet Birthday",
 		Description: "This is the date the wallet start scanning the blockchain for " +
-			"transactions related to this wallet. When updating this date, a rescan " +
-			"must also be done for it to take effect.",
+			"transactions related to this wallet. A rescan of the will automatically" +
+			"be started if the new birthday is earlier than the old one. This option" +
+			" is disabled if there are currently active BTC trades.",
 		DefaultValue: defaultWalletBirthdayUnix,
 		MaxValue:     "now",
-		MinValue:     defaultWalletBirthdayUnix,
-		IsDate:       true,
+		// This MinValue must be removed if we start supporting importing private keys
+		MinValue:          defaultWalletBirthdayUnix,
+		IsDate:            true,
+		DisableWhenActive: true,
 	}}
 
 	commonOpts = []*asset.ConfigOption{
@@ -384,9 +387,10 @@ type WalletConfig struct {
 	FallbackFeeRate  float64 `ini:"fallbackfee"`
 	FeeRateLimit     float64 `ini:"feeratelimit"`
 	RedeemConfTarget uint64  `ini:"redeemconftarget"`
-	WalletName       string  `ini:"walletname"`     // RPC
-	Peer             string  `ini:"peer"`           // SPV
-	Birthday         uint64  `ini:"walletbirthday"` // SPV
+	ActivelyUsed     bool    `ini:"special:activelyUsed"` //injected by core
+	WalletName       string  `ini:"walletname"`           // RPC
+	Peer             string  `ini:"peer"`                 // SPV
+	Birthday         uint64  `ini:"walletbirthday"`       // SPV
 }
 
 // readRPCWalletConfig parses the settings map into a *RPCWalletConfig.
@@ -473,6 +477,7 @@ func (d *Driver) Create(params *asset.CreateWalletParams) error {
 
 	bday := time.Unix(int64(walletCfg.Birthday), 0)
 	now := time.Now()
+	// This check must be removed if we start supporting importing private keys
 	if defaultWalletBirthday.After(bday) {
 		bday = defaultWalletBirthday
 	} else if bday.After(now) {
@@ -561,8 +566,6 @@ type baseWallet struct {
 // method to implement asset.Rescanner.
 type ExchangeWalletSPV struct {
 	*baseWallet
-
-	walletBirthday time.Time
 }
 
 // ExchangeWalletFullNode implements Wallet and adds the FeeRate method.
@@ -585,7 +588,7 @@ func (btc *ExchangeWalletSPV) Rescan(_ context.Context) error {
 	w := btc.node.(*spvWallet)
 	atomic.StoreInt64(&btc.tipAtConnect, 0) // for progress
 	// Caller should start calling SyncStatus on a ticker.
-	return w.rescanWalletAsync(btc.walletBirthday)
+	return w.rescanWalletAsync()
 }
 
 // FeeRate satisfies asset.FeeRater.
@@ -806,8 +809,6 @@ func openSPVWallet(cfg *BTCCloneCFG) (*ExchangeWalletSPV, error) {
 		peers = append(peers, walletCfg.Peer)
 	}
 
-	btc.node = loadSPVWallet(cfg.WalletCFG.DataDir, cfg.Logger.SubLogger("SPV"), peers, cfg.ChainParams)
-
 	bday := time.Unix(int64(walletCfg.Birthday), 0)
 	now := time.Now()
 	if defaultWalletBirthday.After(bday) {
@@ -816,7 +817,10 @@ func openSPVWallet(cfg *BTCCloneCFG) (*ExchangeWalletSPV, error) {
 		bday = now
 	}
 
-	return &ExchangeWalletSPV{baseWallet: btc, walletBirthday: bday}, nil
+	allowAutomaticRescan := !walletCfg.ActivelyUsed
+	btc.node = loadSPVWallet(cfg.WalletCFG.DataDir, cfg.Logger.SubLogger("SPV"), peers, cfg.ChainParams, bday, allowAutomaticRescan)
+
+	return &ExchangeWalletSPV{baseWallet: btc}, nil
 }
 
 // Info returns basic information about the wallet and asset.
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -116,10 +116,10 @@ var (
 	spvOpts = []*asset.ConfigOption{{
 		Key:         "walletbirthday",
 		DisplayName: "Wallet Birthday",
-		Description: "This is the date the wallet start scanning the blockchain for " +
-			"transactions related to this wallet. A rescan of the will automatically" +
-			"be started if the new birthday is earlier than the old one. This option" +
-			" is disabled if there are currently active BTC trades.",
+		Description: "This is the date the wallet starts scanning the blockchain " +
+			"for transactions related to this wallet. If reconfiguring an existing " +
+			"wallet, this may start a rescan if the new birthday is older. This " +
+			"option is disabled if there are currently active BTC trades.",
 		DefaultValue: defaultWalletBirthdayUnix,
 		MaxValue:     "now",
 		// This MinValue must be removed if we start supporting importing private keys
@@ -394,6 +394,20 @@ type WalletConfig struct {
 	Birthday         uint64  `ini:"walletbirthday"`       // SPV
 }
 
+// adjustedBirthday converts WalletConfig.Birthday to a time.Time, and adjusts
+// it so that defaultWalletBirthday <= WalletConfig.Bithday <= now.
+func (cfg *WalletConfig) adjustedBirthday() time.Time {
+	bday := time.Unix(int64(cfg.Birthday), 0)
+	now := time.Now()
+	if defaultWalletBirthday.After(bday) {
+		return defaultWalletBirthday
+	} else if bday.After(now) {
+		return now
+	} else {
+		return bday
+	}
+}
+
 // readRPCWalletConfig parses the settings map into a *RPCWalletConfig.
 func readRPCWalletConfig(settings map[string]string, symbol string, net dex.Network, ports dexbtc.NetPorts) (cfg *RPCWalletConfig, err error) {
 	cfg = new(RPCWalletConfig)
@@ -476,16 +490,7 @@ func (d *Driver) Create(params *asset.CreateWalletParams) error {
 		return err
 	}
 
-	bday := time.Unix(int64(walletCfg.Birthday), 0)
-	now := time.Now()
-	// This check must be removed if we start supporting importing private keys
-	if defaultWalletBirthday.After(bday) {
-		bday = defaultWalletBirthday
-	} else if bday.After(now) {
-		bday = now
-	}
-
-	return createSPVWallet(params.Pass, params.Seed, bday, params.DataDir, params.Logger, chainParams)
+	return createSPVWallet(params.Pass, params.Seed, walletCfg.adjustedBirthday(), params.DataDir, params.Logger, chainParams)
 }
 
 // Open opens or connects to the BTC exchange wallet. Start the wallet with its
@@ -810,16 +815,8 @@ func openSPVWallet(cfg *BTCCloneCFG) (*ExchangeWalletSPV, error) {
 		peers = append(peers, walletCfg.Peer)
 	}
 
-	bday := time.Unix(int64(walletCfg.Birthday), 0)
-	now := time.Now()
-	if defaultWalletBirthday.After(bday) {
-		bday = defaultWalletBirthday
-	} else if bday.After(now) {
-		bday = now
-	}
-
 	allowAutomaticRescan := !walletCfg.ActivelyUsed
-	btc.node = loadSPVWallet(cfg.WalletCFG.DataDir, cfg.Logger.SubLogger("SPV"), peers, cfg.ChainParams, bday, allowAutomaticRescan)
+	btc.node = loadSPVWallet(cfg.WalletCFG.DataDir, cfg.Logger.SubLogger("SPV"), peers, cfg.ChainParams, walletCfg.adjustedBirthday(), allowAutomaticRescan)
 
 	return &ExchangeWalletSPV{baseWallet: btc}, nil
 }
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -109,6 +109,22 @@ var (
 		},
 	}
 
+	// 02 Jun 21 21:12 CDT
+	defaultWalletBirthdayUnix = 1622668320
+	defaultWalletBirthday     = time.Unix(int64(defaultWalletBirthdayUnix), 0)
+
+	spvOpts = []*asset.ConfigOption{{
+		Key:         "walletbirthday",
+		DisplayName: "Wallet Birthday",
+		Description: "This is the date the wallet start scanning the blockchain for " +
+			"transactions related to this wallet. When updating this date, a rescan " +
+			"must also be done for it to take effect.",
+		DefaultValue: defaultWalletBirthdayUnix,
+		MaxValue:     "now",
+		MinValue:     defaultWalletBirthdayUnix,
+		IsDate:       true,
+	}}
+
 	commonOpts = []*asset.ConfigOption{
 		{
 			Key:         "fallbackfee",
@@ -159,7 +175,7 @@ var (
 		Type:        walletTypeSPV,
 		Tab:         "Native",
 		Description: "Use the built-in SPV wallet",
-		ConfigOpts:  commonOpts,
+		ConfigOpts:  append(spvOpts, commonOpts...),
 		Seeded:      true,
 	}
 
@@ -174,7 +190,6 @@ var (
 		},
 		LegacyWalletIndex: 1,
 	}
-	walletBirthday time.Time
 )
 
 // TxInSigner is a transaction input signer.
@@ -369,8 +384,9 @@ type WalletConfig struct {
 	FallbackFeeRate  float64 `ini:"fallbackfee"`
 	FeeRateLimit     float64 `ini:"feeratelimit"`
 	RedeemConfTarget uint64  `ini:"redeemconftarget"`
-	WalletName       string  `ini:"walletname"` // RPC
-	Peer             string  `ini:"peer"`       // SPV
+	WalletName       string  `ini:"walletname"`     // RPC
+	Peer             string  `ini:"peer"`           // SPV
+	Birthday         uint64  `ini:"walletbirthday"` // SPV
 }
 
 // readRPCWalletConfig parses the settings map into a *RPCWalletConfig.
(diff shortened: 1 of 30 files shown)


martonp:
Screen Shot 2022-01-22 at 12 50 27 PM

If fresh wallet is selected, the current time will be used for the birthday. If earliest possible is selected, if the seed generation time is stored, that time will be used, otherwise the hardcoded date will be used. If the custom date is selected, that date will be used unless either the seed generation time is or the hardcoded date is later.

buck54321:
This seems like exactly what `WalletDefinition.ConfigOpts` are for.

martonp:
I was thinking the `ConfigOpts` were things that could change after the wallet has already been created, while this is something that is just needed for wallet creation. I guess it does make sense though to be able to update the wallet birthday and then do a rescan though.

client/core/core.go
```diff
@@ -1168,6 +1170,11 @@ func New(cfg *Config) (*Core, error) {
 		return nil, err
 	}
 
+	var seedGenerationTime uint64
+	if seedGenTime, err := boltDB.SeedGenerationTime(); err == nil {
+		seedGenerationTime = seedGenTime
+	}
+
```
JoeGruffins, line 1177: I would prefer if you didn't ignore the error. You could return a typed error for bucket not existing, and document that an error is expected for legacy wallets, or just return zero and no error, I guess. I would actually prefer a db upgrade, but maybe that's overkill? Are there any other bucket keys in the db that may or may not exist?

JoeGruffins (commented):
Working well.

martonp:
It's now using `WalletDefinition.ConfigOpts `. On the create wallet screen it looks the same as in the picture above, but when reconfiguring the wallet, there is just a regular date input. If the wallet is reconfigured with a new birthday, rescan will update the walletdb and use the new birthday.

client/asset/btc/spv.go
```diff
@@ -144,7 +144,7 @@ func createSPVWallet(privPass []byte, seed []byte, dbDir string, log dex.Logger,
 
 	pubPass := []byte(wallet.InsecurePubPassphrase)
 
-	_, err = loader.CreateNewWallet(pubPass, privPass, seed, walletBirthday)
+	_, err = loader.CreateNewWallet(pubPass, privPass, seed, bday)
```
chappjc, line 147: We need to consider what to do about/with the existing `spvWallet.birthday` field.

Here on initial wallet **creation** the passed birthday gets stored in the wallet Manager DB, and then on wallet startup (`startWallet`) it loads it like `spvWallet.birthday = btcw.Manager.Birthday()`.  This field was just used in the syncStatus method so far.

If someone later opens the wallet (or reconfigures it) with a different wallet birthday in the settings, `openSPVWallet` will record that in `ExchangeWalletSPV.walletBirthday`, but it does not seem to update the field in `spvWallet` or in the Manager's DB (unless rescan is done in the case of the DB).

Maybe I'm not seeing it all clearly yet, but I think we need to look at merging the two bday fields.  It does make sense though not to update the wallet db until a rescan has actually been applied with the new birthday.

martonp: It's definitely incorrect that `spvWallet.birthday` was not getting updated during the rescan, I've updated that now. The way I see it currently is that `spvWallet.birthday` is basically a cache for what is stored in the wallet db, both of which are updated when a rescan is done, and `ExchangeWalletSPV.walletBirthday` represents the latest selection by the user. I think that's fairly reasonable.. what's the reason they should be merged? Should we automatically start a rescan if the birthday is updated?

chappjc: > The way I see it currently is that `spvWallet.birthday` is a cache for what is stored in the wallet db, both of which are updated when a rescan is done, and `ExchangeWalletSPV.walletBirthday` represents the latest selection by the user. I think that's fairly reasonable.. what's the reason they should be merged?

That's a reasonable interpretation.  It was just that two birthday fields, one at high level and one low, stood out as a possible mistake.  But how I would answer your question is that there's a likelihood of them diverging accidentally, as happened in the initial draft.  We can run with what you've got now though IMO.

client/core/core.go
```diff
@@ -3073,6 +3082,14 @@ func (c *Core) InitializeClient(pw, restorationSeed []byte) error {
 	if err != nil {
 		return fmt.Errorf("SetPrimaryCredentials error: %w", err)
 	}
+
+	freshSeed := len(restorationSeed) == 0
+	if freshSeed {
+		now := uint64(time.Now().Unix())
```
chappjc, line 3088: Just noticed there's a `if len(restorationSeed) == 0` after `c.setCredentials` below.  Does this need to be up here?

client/asset/btc/btc.go
```diff
@@ -766,7 +792,15 @@ func openSPVWallet(cfg *BTCCloneCFG) (*ExchangeWalletSPV, error) {
 
 	btc.node = loadSPVWallet(cfg.WalletCFG.DataDir, cfg.Logger.SubLogger("SPV"), peers, cfg.ChainParams)
 
-	return &ExchangeWalletSPV{btc}, nil
+	bday := time.Unix(int64(walletCfg.Birthday), 0)
```
chappjc, line 795: Here's part of what's fishy about forcing the birthday set in the `ConfigOpts`.  Here we may be preparing to start an **existing** `spvWallet` with a birthday already stored in the walletdb (not using `createSPVWallet`), but `NewWallet` > `openSPVWallet` will just get whatever birthday value and set it in `ExchangeWalletSPV.walletBirthday`.

Then later when the wallet is actually started with `connect`>`startSPVWallet`, it pulls it out of wallet.Manager's walletdb.  It could differ from what the caller gave to `NewWallet`.

However, I acknowledge that it looks good the way the reconfigure dialog works w.r.t. birthday:

image

I just find it a little goofy that changing wallet settings required "reminding" the wallet what it's birthday is when it's stored.

Sorry I have no concrete suggestions to improve this other than logging here and in createSPVwallet what birthday is being set

chappjc:
In testing, I hit an existing data race in `ReconfigureWallet` that I believe can be resolved as follows:

```diff
diff --git a/client/core/core.go b/client/core/core.go
index 57bc9f2e..077dbae6 100644
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -2269,7 +2269,7 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, form *WalletForm) er
                c.log.Warnf("Error getting balance for wallet %s: %v", unbip(assetID), err)
                // Do not fail in case this requires an unlocked wallet.
        } else {
-               wallet.balance = balances           // update xcWallet's WalletBalance
+               wallet.setBalance(balances)         // update xcWallet's WalletBalance
                dbWallet.Balance = balances.Balance // store the db.Balance
        }
```

The race is because when `connectWallet` returns above this line, `monitorWalletSync` is running, calling `walletCheckAndNotify` > `wallet.state`.  I think we got away without this race until now because `syncTickerPeriod` had been 10 seconds, but now it's 3 seconds and it's easier to get a tick before reaching this part of `ReconfigureWallet`.

Would you mind applying this patch and seeing if you see any issues with a -race build on your end?

client/asset/btc/btc.go
martonp: In the beginning I had this birthday setting not a part of the wallet configuration, it was just another field in the `asset.CreateWalletParams` struct. We could put it back there. Then we could have the birthday be a parameter in the rescan request. On the UI, the rescan button would be on the wallet settings page as a separate section, and there would be a date input next to it. In this way there wouldn't be two dates in two separate databases. How do you feel about this?

client/core/core.go
martonp: No you're right, I'll combine them.

client/asset/btc/btc.go
chappjc: Hmm, that's an idea.  Since I commented I was imagining that a reconfigure with a new wallet birthday could detect a change an emit some kind of notification to suggest a rescan to have the new birthday take effect.

Let's see what @buck54321 thinks about that potential resolution to this UI/UX peculiarity since he requested doing this in `ConfigOpts` (for generally good reasons), although that was before your idea of putting the rescan button in a section of settings and tying birthday changes to a rescan.

I don't think the current solution is unacceptable, but it's somehow different from the other "settings", both in that it's awkward to constantly feed it back into wallet to change other settings, and in that it doesn't just "go into effect"

martonp: I think this will be more natural for the user. There's really no point in changing the birthday if you're not going to do a rescan.

martonp: @buck54321 Please let me know what's your opinion about this when you get a chance.

buck54321: If my read is right, we should pass in the `WalletConfig.Birthday` to `loadSPVWallet`, and then later check that in `(*spvWallet).startWallet` against the birthday stored in the btcwallet database. If there is a discrepancy, we would overwrite the btcwallet database value and (`if newBDay < oldBDay`) start a rescan. That's shouldn't be a problem, because we start a sync loop in `(*Core).connectWallet`, which is also called in `ReconfigureWallet`.

chappjc: Yeah, you're definitely on the same page.  My main hesitation with that was the last sentence in my previous comment on this, namely:

> We'd also have to worry about accidentally starting a wallet rescan with active swaps if it's triggered through ReconfigureWallet if the birthday appears to be changed.

I also still find it somewhat awkward that we're requiring a birthday when changing any other unrelated config change, but I think we can work around that by allowing a nil or zero-value birthday to indicate no change is intended.

Anyway, in this birthday-change-detected scenario, I'd sorta hope we could _not_ trigger an automatic rescan unless we tell the wallet that it is safe to do so.  Alternatively, possibly frontend could just advise to manually request a rescan seeing as it knows if the form was changed?

buck54321: > I also still find it somewhat awkward that we're requiring a birthday when changing any other unrelated config change, but I think we can work around that by allowing a nil or zero-value birthday to indicate no change is intended.

It'll be zero valued until the user specifies a custom birthday. Once the user specifies a custom birthday, the `WalletConfig.Settings` will always contain that birthday for us to compare, including during reconfiguration. So zero-value will be translated to default birthday by `ExchangeWallet`, but once a custom value is set it'll always be there. The "no change is intended" is signaled by equality of birthdays in `WalletConfig.Settings` and the btcwallet database.

> I'd sorta hope we could not trigger an automatic rescan unless we tell the wallet that it is safe to do so

> We'd also have to worry about accidentally starting a wallet rescan with active swaps if it's triggered through ReconfigureWallet if the birthday appears to be changed.

Does seem a little sketchy. Do you think that would end badly? If the risk is just a matter of some transient errors during the rescan, I'd be okay with it.

I'm with you 100% that wallet reconfiguration could use an overhaul, but I also don't have any solid suggestions for how to proceed. Maybe I'll open an issue. As it relates to this discussion, I guess a user shouldn't be allowed to set an earlier birthday if there are active orders because it requires a rescan? It's kind of odd though, because we already allow that the wallet may need to sync after a reconfigure, and we don't add any specific protections for the case of a long resync for other wallets. I guess we just assume that rpc wallet reconfigurations wouldn't result in a long sync? Any idea how long the rescan is with `PersistToDisk = true`?

I don't think a user should be able to set an earlier birthday without an automatic rescan.

chappjc: A normal sync up after reconfigure is really nothing compared to a rescan.  It could take an hour or more depending on peers and wallet history that might require fetching tons of actual blocks when filters match.  An automatic rescan would definitely be trouble.  Since the wallet's transaction history is dropped and rebuilt in the process, we really need to avoid initiating a rescan if there is anything active at all IMO.

buck54321: Hmm. Maybe you're right then that we should just skip the rescan and (somehow) send a notification for now, and work towards a better solution for reconfiguration later.

As a mid-term solution, we might consider a field in `ConfigOpt` like `DisableWhenActive bool` that would disable the input in the UI when the wallet has active orders.

chappjc: That `ConfigOption` could cover the UI.  Although the safest place to block a rescan would be in core when the new settings are actually submitted.  Perhaps `ReconfigureWallet` could inject `wcfg.Settings["special:permitAutoRescan"] = <bool>`.

buck54321: Ooh. I like that.

martonp: > In the beginning I had this birthday setting not a part of the wallet configuration, it was just another field in the `asset.CreateWalletParams` struct. We could put it back there. Then we could have the birthday be a parameter in the rescan request. On the UI, the rescan button would be on the wallet settings page as a separate section, and there would be a date input next to it. In this way there wouldn't be two dates in two separate databases. How do you feel about this?

What about this idea? It seems like it would solve all the complications. Just block rescans when there is anything active happening. It would be much more obvious for the user that once they click the rescan button, they will not be able to use the wallet for a while.

buck54321: Maybe. How does the user see what the currently set birthday is? 

I'm hesitant to add custom fields to `CreateWalletParams` that are specific to an asset. The `ConfigOpts`/`Settings` system was created to reign in the asset-specific configuration and keep the API sensible. 

I'm sort of on the fence here. We **should** just overhaul the wallet configuration system. @martonp why don't you choose whatever solution seems like the least work, and then we can bang our heads together on a new configuration system to squeeze into Phase III. I personally am leaning towards `ConfigOpts.DisableWhenActive` **+** `wcfg.Settings["special:permitAutoRescan"] = <bool>`, but just barely.

martonp: > Maybe. How does the user see what the currently set birthday is?

That's a good question, there's not really a mechanism to send it at the moment. I guess ideally there would be creation only settings, and then maybe some kind of way to retrieve some read only information from the wallet. I'll just go with `ConfigOpts.DisableWhenActive` + `wcfg.Settings["special:permitAutoRescan"] = <bool>`.

buck54321 (commented):
Just a couple of comments.

client/core/core.go
```diff
@@ -1896,6 +1909,17 @@ func (c *Core) WalletState(assetID uint32) *WalletState {
 	return wallet.state()
 }
 
+// AssetHasActiveOrders checks whether there are any active orders or
+// negotiating matches for the specified asset.
+func (c *Core) AssetHasActiveOrders(assetID uint32) bool {
```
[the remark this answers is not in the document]
chappjc: That reminds me, I had something similar in frontend for the rescan PR but we ended up moving into Core for the check.  https://github.com/decred/dcrdex/commit/3ab44ef9c33fe3b043265677dd9b6a2a5fb06feb  Can use that JS or something like it if you want to move the logic to frontend.

This is a different use case of course, but here's the previous discussion: https://github.com/decred/dcrdex/pull/1374#discussion_r788948033
I ended up agreeing that it made sense to check in Core, but I also appreciate that frontend can know too, even if it's check might turn out to be stale, with Core ultimately blocking it.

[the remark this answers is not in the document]
chappjc: Ohhhh, this method is double-purposed in `apiWalletSettings` as well as `loadWallet`.  Yah, it could be omitted from the wallet setting response.

client/webserver/site/src/localized_html/en-US/forms.tmpl
```diff
@@ -51,6 +55,16 @@
...
+  <div class="d-hide" data-tmpl="selectBirthday">
+    <span>Select wallet birthday: </span><span class="ico-info" data-tooltip="This wallet scans for transactions starting on a certain date. If you have never used this wallet, initialize it as fresh wallet. If you are not sure, use the earliest possible birthday. If you remember the first time you initialized this wallet, select a custom date."></span><br/>
+    <input type="radio" id="fresh" name="birthday" value="fresh" data-tmpl="freshButton">
+    <label for="fresh">Initialize as fresh wallet</label><br/>
+    <input type="radio" id="earliest" name="birthday" value="earliest" data-tmpl="earliestButton">
+    <label for="earliest">Use earliest possible birthday</label><br/>
+    <input type="radio" id="custom" name="birthday" value="custom" data-tmpl="customButton">
+    <label for="custom">Enter custom birthday</label>
+    <input type="date" data-tmpl="customDate">
+  </div>
```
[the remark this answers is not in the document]
martonp: Yeah you're right.

chappjc:
Silly conflict with btc.go.

client/asset/btc/btc.go
```diff
@@ -109,6 +109,26 @@ var (
...
 	}
 
+	// 02 Jun 21 21:12 CDT
+	defaultWalletBirthdayUnix = 1622668320
+	defaultWalletBirthday     = time.Unix(int64(defaultWalletBirthdayUnix), 0)
+
+	spvOpts = []*asset.ConfigOption{{
+		Key:         "walletbirthday",
+		DisplayName: "Wallet Birthday",
+		Description: "This is the date the wallet start scanning the blockchain for " +
```
chappjc, line 119: This is the date the wallet start**s** scanning...

client/asset/btc/btc.go
```diff
@@ -109,6 +109,26 @@ var (
...
+			"transactions related to this wallet. A rescan of the will automatically" +
```
chappjc, line 120: "rescan of the will"

client/asset/btc/btc.go
```diff
@@ -109,6 +109,26 @@ var (
...
+			"be started if the new birthday is earlier than the old one. This option" +
```
chappjc, line 121: Missing trailing space on this line and the previous.

image

chappjc: The part about starting a rescan if "earlier than the old one" is pretty confusing when viewed from the wallet **creation** dialog rather than settings/reconfig.  To make this sensible in both context, how about "If reconfiguring an existing wallet, this may start a rescan if the new birthday is older."

client/core/core.go
```diff
@@ -1169,6 +1171,11 @@ func New(cfg *Config) (*Core, error) {
 		return nil, err
 	}
 
+	seedGenerationTime, err := boltDB.SeedGenerationTime()
+	if err != nil && !errors.Is(err, db.ErrNoSeedGenTime) {
+		return nil, err
+	}
```
chappjc, line 1177: Instead of proceeding zero if `err` is `db.ErrNoSeedGenTime`, why don't we hard code the same value as used with btc since that's basically the earliest for an hd key.

client/core/core.go
```diff
@@ -3086,6 +3110,14 @@ func (c *Core) InitializeClient(pw, restorationSeed []byte) error {
 	if err != nil {
 		return fmt.Errorf("SetPrimaryCredentials error: %w", err)
 	}
+
+	freshSeed := len(restorationSeed) == 0
+	if freshSeed {
+		now := uint64(time.Now().Unix())
+		c.db.SetSeedGenerationTime(now)
```
chappjc, line 3117: error return is ignored

client/webserver/locales/en-us.go
```diff
@@ -213,4 +213,9 @@ var EnUS = map[string]string{
 	"Maximum Possible Swap Fees":  "Maximum Possible Swap Fees",
 	"max_fee_conditions":          "This is the most you would ever pay in fees on your swap. Fees are normally assessed at a fraction of this rate. The maximum is not subject to changes once your order is placed.",
 	"wallet_logs":                 "Wallet Logs",
+	"Initialize fresh":            "Initialize as fresh wallet",
+	"Earliest possible birthday":  "Use earliest possible birthday",
+	"Enter custom birthday":       "Enter custom birthday",
+	"Select wallet birthday":      "Select wallet birthday",
+	"w_birthday_tooltip":          "This wallet scans for transactions starting on a certain date. If you have never used this wallet, initialize it as fresh wallet. If you are not sure, use the earliest possible birthday. If you remember the first time you initialized this wallet, select a custom date.",
```
chappjc, line 220: These are all no longer used, correct?  Also the `selectBirthday` data-tmpl in forms.tmpl

chappjc (changes requested):
All working well.  Just a few formatting and grammatical issues with the birthday config opt description.

buck54321 (changes requested):

client/core/core.go
martonp: I'm actually checking if seedgentime = 0 over here:
https://github.com/martonp/dcrdex/blob/3d7963aa63a89df3551414e621c1ce0f3171234a/client/webserver/site/src/js/forms.js#L122

If the time the seed was generated is known, the default is the current date, so basically no scanning needs to be done. Otherwise, the default date that the wallet provides is used. The way it is currently used, `core.seedgentime` could just be a boolean that tells whether or not the current installation generated the seed.

client/asset/interface.go
```diff
@@ -130,10 +130,25 @@ type ConfigOption struct {
...
-	IsBoolean    bool        `json:"isboolean"`
+	// If MaxValue/MinValue are set to the string "now" for a date config, the
+	// UI will display the current date.
+	MaxValue          interface{} `json:"max"`
+	MinValue          interface{} `json:"min"`
+	NoEcho            bool        `json:"noecho"`
+	IsBoolean         bool        `json:"isboolean"`
+	IsDate            bool        `json:"isdate"`
+	DisableWhenActive bool        `json:"disablewhenactive"`
+	IsBirthdayConfig  bool        `json:"isBirthdayConfig"`
```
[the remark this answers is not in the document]
martonp: The behavior I wanted on the UI was to have the default value be "now" if the current installation created the seed, but otherwise be the default value passed in the config.

client/asset/btc/spv.go
```diff
@@ -1177,7 +1183,37 @@ func (w *spvWallet) startWallet() error {
 	w.cl = chainService
 	w.chainClient = chain.NewNeutrinoClient(w.chainParams, chainService)
 	w.wallet = &walletExtender{btcw, w.chainParams}
-	w.birthday = btcw.Manager.Birthday()
+
+	oldBday := btcw.Manager.Birthday()
+	wdb := btcw.Database()
+	performRescan := w.allowAutomaticRescan && w.birthday.Before(oldBday)
+
```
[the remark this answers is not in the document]
martonp: Let's do an error.

buck54321 (approved):

chappjc:
can merge, but I made conflicts

chappjc (approved):

PR #1430, opened 2022-01-20
client/{core,asset,btc,dcr}: Add FeeRater interface and WalletTraitFeeRater

buck54321:
```
A FeeRater can generate fee suggestions when they are not validated by the
server, specifically for inputs to Withdraw, PayFee, Send, and
possibly Refund. This corresponds to places where
(*Core).feeSuggestionAny would be used.
```

PR #1444, opened 2022-01-24
client/ui: Display registration tx fee estimate

martonp:
- `client/core`: Add `EstimateRegistrationTxFee` function in core that determines the tx fee needed to register for the dex
- `client/asset`: Add `EstimateRegistrationTxFee` to wallet interface
- `client/webserver`: Add a new api endpoint `/regtxfee`
- `ui`: Each time the WalletWaitForm is loaded, this endpoint is called to retrieve an estimate

### DIFF
--- a/client/asset/interface.go
+++ b/client/asset/interface.go
@@ -299,6 +299,9 @@ type Wallet interface {
 	// payment. This method need not be supported by all assets. Those assets
 	// which do no support DEX registration fees will return an ErrUnsupported.
 	RegFeeConfirmations(ctx context.Context, coinID dex.Bytes) (confs uint32, err error)
+	// EstimateRegistrationTxFee returns an estimate for the tx fee needed to
+	// pay the registration fee using the provided feeRate.
+	EstimateRegistrationTxFee(feeRate uint64) uint64
 }
 
 // Rescanner is a wallet implementation with rescan functionality.
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -2707,7 +2707,7 @@ func (c *Core) EstimateRegistrationTxFee(host string, certI interface{}, assetID
 		if r, err := rater.FeeRate(); err == nil {
 			rate = r
 		} else {
-			c.log.Debugf("failed to get fee suggestion from %s FeeRater: %v", unbip(assetID), err)
+			c.log.Warnf("failed to get fee suggestion from %s FeeRater: %v", unbip(assetID), err)
 		}
 	}
 
@@ -2720,7 +2720,7 @@ func (c *Core) EstimateRegistrationTxFee(host string, certI interface{}, assetID
 		if err == nil {
 			rate = dc.fetchFeeRate(assetID)
 		} else {
-			c.log.Debugf("failed to connect to dex: %v", err)
+			c.log.Warnf("failed to connect to dex: %v", err)
 		}
 	}
 
@@ -3776,7 +3776,7 @@ func (c *Core) feeSuggestionAny(assetID uint32, preferredConns ...*dexConnection
 			if r, err := rater.FeeRate(); err == nil {
 				return r
 			} else {
-				c.log.Debugf("failed to get fee suggestion from %s FeeRater: %v", unbip(assetID), err)
+				c.log.Warnf("failed to get fee suggestion from %s FeeRater: %v", unbip(assetID), err)
 			}
 		}
 	}
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -2692,6 +2692,42 @@ func (c *Core) DiscoverAccount(dexAddr string, appPW []byte, certI interface{}) 
 	return dc.exchangeInfo(), true, nil
 }
 
+// EstimateRegistrationTxFee provides an estimate for the tx fee needed to
+// pay the registration fee for a certain asset. The dex host is required
+// because the dex server is used as a fallback to determine the current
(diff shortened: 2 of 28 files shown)


client/asset/interface.go
```diff
@@ -270,6 +270,10 @@ type Wallet interface {
 	// payment. This method need not be supported by all assets. Those assets
 	// which do no support DEX registration fees will return an ErrUnsupported.
 	RegFeeConfirmations(ctx context.Context, coinID dex.Bytes) (confs uint32, err error)
+	// EstimateRegistrationTxFee provides an estimate for the tx fee needed to
+	// pay the registration fee. rateOracleFallback is used in the case where
```
chappjc, line 274: When this gets rebased after https://github.com/decred/dcrdex/pull/1430 is merged, please be sure to scour the comments for mentions of the old `rateOracleFallback`

chappjc: https://github.com/decred/dcrdex/pull/1430 is merged.  Sorry I held that up a few days longer than necessary.  I missed the update to that.

martonp: All good, I'll get this ready to review soon.

client/webserver/site/src/js/app.js
```diff
@@ -658,6 +658,18 @@ export default class Application {
     return o
   }
 
+  haveAssetOrders (assetID) {
```
chappjc, line 661: Snuck in latest commit in https://github.com/decred/dcrdex/pull/1439?

client/core/core.go
```diff
@@ -2692,6 +2692,42 @@ func (c *Core) DiscoverAccount(dexAddr string, appPW []byte, certI interface{})
...
+	if err != nil {
+		return 0, err
+	}
+
+	var rate uint64
+	if rater, is := wallet.Wallet.(asset.FeeRater); is {
+		if r, err := rater.FeeRate(); err == nil {
+			rate = r
+		} else {
+			c.log.Debugf("failed to get fee suggestion from %s FeeRater: %v", unbip(assetID), err)
```
chappjc, line 2710: This `Debugf` feels more like a `Warnf` to me.

client/core/core.go
```diff
@@ -2692,6 +2692,42 @@ func (c *Core) DiscoverAccount(dexAddr string, appPW []byte, certI interface{})
...
+	if rate == 0 {
+		dc, err := c.tempDexConnection(host, certI)
+		if dc != nil {
+			// Stop (re)connect loop, which may be running even if err != nil.
+			defer dc.connMaster.Disconnect()
+		}
+		if err == nil {
+			rate = dc.fetchFeeRate(assetID)
+		} else {
+			c.log.Debugf("failed to connect to dex: %v", err)
```
chappjc, line 2723: `rate` stays zero in this branch, so I think this should be `Warnf` since it's going to use the wallet's fallback fee rate in the method below.

chappjc (approved):
Looks great.

image

image

client/webserver/site/src/js/app.js
martonp: Oops

client/core/core.go
```diff
@@ -2692,6 +2692,42 @@ func (c *Core) DiscoverAccount(dexAddr string, appPW []byte, certI interface{})
...
+	}
+
+	var rate uint64
+	if rater, is := wallet.Wallet.(asset.FeeRater); is {
+		if r, err := rater.FeeRate(); err == nil {
+			rate = r
+		} else {
+			c.log.Warnf("failed to get fee suggestion from %s FeeRater: %v", unbip(assetID), err)
+		}
+	}
```
buck54321, line 2712: I'm on the fence on whether this should actually be a `feeSuggestionAny` to avoid a temp connection if there are existing connections from which we can get a fee rate.

client/webserver/api.go
```diff
@@ -46,6 +46,29 @@ func (s *WebServer) apiDiscoverAccount(w http.ResponseWriter, r *http.Request) {
...
+	}
+	cert := []byte(form.Cert)
+	txFee, err := s.core.EstimateRegistrationTxFee(form.Addr, cert, *form.AssetID)
+	if err != nil {
+		s.writeAPIError(w, err)
+		return
+	}
+	resp := struct {
+		OK       bool           `json:"ok"`
+		Exchange *core.Exchange `json:"xc,omitempty"`
```
buck54321, line 63: Looks like this can go.

buck54321 (approved):
Only change requested is removing the extra line in the response `struct` in `apiEstimateRegistrationTxFee`. Good to go, otherwise. Nice work!

chappjc (approved):
Re-approving with the updated `/regtxfee` endpoint response.

I'm content with the fee estimate using the current basic arithmetic if you guys are.

On the topic of `feeSuggestionAny` vs always doing `tempDexConnection`, my feeling is that if the vast majority of use cases will be people registering for their one and only dex server, so I think it's alright as is.  The cost of the temp connection is not great.

client/core/core.go
martonp: I was thinking it made sense just to get the rate from the server that the user is registering to, but both are fine. I'll update if you prefer `feeSuggestionAny`.

buck54321:
Yeah, those methods I mentioned wouldn't always work anyways, since they would require the wallet to have available inputs. Good as is.

buck54321 (approved):

PR #1447, opened 2022-01-25
server/eth: Match fee rate = MaxFeeRate

### LINKED ISSUE #1372: server/asset/eth: max fee rate needs to be a max gas cap

martonp:
- server/asset: Add `Info` and `ValidateFeeRate` functions
- server/dex: Add `SwapFeeRate` function to `FeeFetcher`. This will return the max fee rate for assets that support dynamic transaction fees, but the regular tx fee otherwise
- server/market: Use `SwapFeeRate` instead of the regular `FeeRate`
- server/swap: Use `ValidateFeeRate` function instead of just comparing the fee rates. This is to allow ETH to check the prority fee in addition to the `MaxFeeCap`

Closes #1372.


server/swap/swap.go
```diff
@@ -1386,8 +1386,8 @@ func (s *Swapper) processInit(msg *msgjson.Message, params *msgjson.Init, stepIn
 	if stepInfo.isBaseAsset {
 		reqFeeRate = stepInfo.match.FeeRateBase
 	}
-	if contract.FeeRate() < reqFeeRate {
```
chappjc, line 1389: A related change I've been meaning to make, probably next day or two, is to skip the fee rate check if the txn turns out to be mined already.   Just a note.

server/dex/feemgr.go
```diff
@@ -104,3 +104,14 @@ func (f *feeFetcher) LastRate() uint64 {
...
 }
+
+// SwapFeeRate returns the tx fee that needs to be used to initiate a swap.
+// This fee will be the max fee rate if the asset supports dynamic tx fees,
+// and otherwise it will be the current market fee rate.
+func (f *feeFetcher) SwapFeeRate(ctx context.Context) uint64 {
+	if f.Backend.SupportsDynamicTxFee() {
+		return f.MaxFeeRate()
+	} else {
+		return f.FeeRate(ctx)
```
chappjc, line 115: One thing to consider is for ETH does `eth.node.suggestGasPrice(ctx)` (part of `FeeRate` about 10 lines) even have much value anymore?  That function [gives a fee rate suitable for legacy (non-dynamic) txns](https://github.com/ethereum/go-ethereum/blob/015fde9a2c4b6e722a9de5bfea2a14226806a231/internal/ethapi/api.go#L63-L64), I had actually sorta figured we'd kill that and have `FeeRate` behave more like maxFeeRate.  I can see that being valuable as an oracle for clients, maybe, but we'd probably adjust the formula because I find it doesn't give high enough fee rates.

I think your approach works though.  Just scanned through once.

server/asset/eth/eth.go
```diff
@@ -238,6 +238,30 @@ func (eth *Backend) FeeRate(ctx context.Context) (uint64, error) {
...
+// be less than reqFeeRate, but for Eth, the gasTipCap must also be checked.
+func (eth *Backend) ValidateFeeRate(contract *asset.Contract, reqFeeRate uint64) bool {
+	coin := contract.Coin
+	sc, ok := coin.(*swapCoin)
+	if !ok {
+		eth.log.Error("eth contract coin type must be a swapCoin but got %T", sc)
+		return false
+	}
+
+	if sc.dynamicTx && sc.gasTipCap < dexeth.MinGasTipCap {
```
chappjc, line 258: This seems good, in that `ValidateFeeRate` will facilitate a more sophisticated `gasTipCap` check than just comparing `MinGasTipCap`, such as something based on a running average of recent blocks tips, but as discussed in planning this, I believe a constant will be quite adequate.

server/swap/swap.go
martonp: Good idea

server/dex/feemgr.go
martonp: Yeah I wanted to keep it so the `FeeRateRoute` wouldn't just return the max fee rate. How about 2 * previous base rate?

server/asset/eth/eth.go
martonp: I thought this would be adequate also. I generally 1 gwei is enough for the tip cap. If 2 gwei isn't enough it would be some kind of extreme network conditions, and even then the client still checks the suggested tip cap. It's just that the server doesn't do more complex verification.

server/dex/feemgr.go
chappjc: :+1: That seems to be geth's default in various API functions.

https://github.com/ethereum/go-ethereum/blob/015fde9a2c4b6e722a9de5bfea2a14226806a231/internal/ethapi/transaction_args.go#L95-L98

server/dex/feemgr.go
```diff
@@ -104,3 +104,14 @@ func (f *feeFetcher) LastRate() uint64 {
...
 	return f.Asset.MaxFeeRate
 }
+
+// SwapFeeRate returns the tx fee that needs to be used to initiate a swap.
+// This fee will be the max fee rate if the asset supports dynamic tx fees,
+// and otherwise it will be the current market fee rate.
+func (f *feeFetcher) SwapFeeRate(ctx context.Context) uint64 {
+	if f.Backend.SupportsDynamicTxFee() {
+		return f.MaxFeeRate()
+	} else {
```
JoeGruffins, line 114: Already returned in if so no need for the `else` clause.

JoeGruffins (approved):
Working well.

server/asset/eth/eth.go
```diff
@@ -231,11 +231,54 @@ func (eth *Backend) InitTxSizeBase() uint32 {
...
```
buck54321, line 277: We should probably just enforce dynamicTx = true for all of our transactions.

server/asset/btc/btc.go
```diff
@@ -493,6 +493,18 @@ func (btc *Backend) FeeRate(_ context.Context) (uint64, error) {
 	return btc.estimateFee(btc.node)
 }
 
+// SupportsDynamicTxFee returns true if the tx fee for this asset adjusts based
+// on market conditions
+func (*Backend) SupportsDynamicTxFee() bool {
+	return false
+}
```
buck54321, line 500: Wouldn't need to implement this here and in **dcr** if you made `SupportsDynamicTxFee` part of an optional interface like we're doing with the client.

buck54321 (commented):
Concept ACK. Untested and one suggestion, but feel free to merge when you're comfy, @chappjc.

server/asset/btc/btc.go
buck54321: Actually, I'm now thinking this might not work like I expected, since it's just a boolean return value. So we'd verify that the interface is a DynamicFeeSupporter or whatever, and then still check the boolean? What we really need is something akin to WalletInfo, but on the server side. I'm okay with this as is too

chappjc: Yeah, it works as is.  @martonp can try some stuff out I guess, but I can see the optional method being `SwapFeeRate`, which would make `feeFetcher` unaware of the whole dynamic<->max_fee relationship.  Maybe we like that dynamic/max concept front and center?

Specifically, changing

```go
// SwapFeeRate returns the tx fee that needs to be used to initiate a swap.
// This fee will be the max fee rate if the asset supports dynamic tx fees,
// and otherwise it will be the current market fee rate.
func (f *feeFetcher) SwapFeeRate(ctx context.Context) uint64 {
	if f.Backend.SupportsDynamicTxFee() {
		return f.MaxFeeRate()
	}
	return f.FeeRate(ctx)
}
```

into

```go
// SwapFeeRate returns the tx fee that needs to be used to initiate a swap.
// This will use the Backend's SwapFeeRate method if it exists, otherwise
// it will request the fee rate based on current network conditions.
func (f *feeFetcher) SwapFeeRate(ctx context.Context) uint64 {
	if fr, ok := f.Backend.(asset.SwapFeeRater); ok {
		return fr.SwapFeeRate()
	}
	return f.FeeRate(ctx)
}
```

The question is do we hide the whole dynamic-must-use-max concept in the Backend's methods or do we apply that notion at the higher levels like done currently in this PR?

server/swap/swap.go
chappjc: Done in https://github.com/decred/dcrdex/pull/1458

server/asset/btc/btc.go
martonp: This idea's not bad, maybe there will be another coin which requires some other technique.

chappjc: Ahh, that's right.  Max fee rate is a market/swapper concept, so just be done higher up.  Check.

martonp:
We don't really need to check for dynamic vs legacy, in the geth `Transaction` object, for a legacy transaction `GasFeeCap` and `GasTipCap` will both just return the gas price. `GasPrice` returns the fee cap in a dynamic transaction. Anyways, the code is more simple like this. I also added a test to make sure the coiner was populating the tip cap correctly.

chappjc:
@martonp I noticed that too:

https://github.com/ethereum/go-ethereum/blob/aaca58a7a1d9acbd24bbc74c49933efa2f1af183/core/types/legacy_tx.go#L99-L101

```go
func (tx *LegacyTx) gasPrice() *big.Int     { return tx.GasPrice }
func (tx *LegacyTx) gasTipCap() *big.Int    { return tx.GasPrice }
func (tx *LegacyTx) gasFeeCap() *big.Int    { return tx.GasPrice }
```

vs.

```go
func (tx *DynamicFeeTx) gasFeeCap() *big.Int    { return tx.GasFeeCap }
func (tx *DynamicFeeTx) gasTipCap() *big.Int    { return tx.GasTipCap }
func (tx *DynamicFeeTx) gasPrice() *big.Int     { return tx.GasFeeCap }
```

accessed via

```go
// GasPrice returns the gas price of the transaction.
func (tx *Transaction) GasPrice() *big.Int { return new(big.Int).Set(tx.inner.gasPrice()) }

// GasTipCap returns the gasTipCap per gas of the transaction.
func (tx *Transaction) GasTipCap() *big.Int { return new(big.Int).Set(tx.inner.gasTipCap()) }

// GasFeeCap returns the fee cap per gas of the transaction.
func (tx *Transaction) GasFeeCap() *big.Int { return new(big.Int).Set(tx.inner.gasFeeCap()) }
```

But your point is that if it's a legacy txn with a `gasPrice` set to our (very high) `gasFeeCap`, so be it, even though that's incredibly wasteful of the user?

So you are proposing setting `baseCoin.gasFeeCap` to either `gasPrice` if it is a legacy txn or actually `gasFeeCap` for dynamic.  Seems reasonable, and if we do decide to block legacy txns that have a high enough gas price, we can always go back to `(*Backend).baseCoin` to block it.

Finally, given the above code snippets, was our way of detecting legacy/dynamic tx type incorrect before?  Should we have been using `(tx *Transaction) Type() uint8`?

martonp:
> But your point is that if it's a legacy txn with a `gasPrice` set to our (very high) `gasFeeCap`, so be it, even though that's incredibly wasteful of the user?

Yes, I think the client should definitely use a dynamic transaction, but I don't see a reason to block legacy ones. 

> Finally, given the above code snippets, was our way of detecting legacy/dynamic tx type incorrect before? Should we have been using `(tx *Transaction) Type() uint8`?

Yeah, what we had was pointless lol.

server/asset/eth/coiner.go
```diff
@@ -145,17 +146,23 @@ func (eth *Backend) baseCoin(coinID []byte, contractData []byte) (*baseCoin, err
...
+		return nil, fmt.Errorf("Failed to parse gas fee cap from tx %s", txHash)
+	}
+	gasFeeCapGwei, err := dexeth.WeiToGweiUint64(gasFeeCap)
+	if err != nil {
+		return nil, fmt.Errorf("unable to convert gas fee cap: %v", err)
 	}
 
-	gasPrice, err := dexeth.WeiToGweiUint64(rate)
+	gasTipCap := tx.GasTipCap()
+	if gasTipCap == nil || gasTipCap.Cmp(zero) <= 0 {
```
chappjc, line 160: With a legacy transaction, which returns the same gas price for both `GasFeeCap` and `GasTipCap`, I think that the `swapCoin.gasTipCap` check in `(*Backend).ValidateFeeRate` is wrong or at least misleading.
Say we are debugging and we inspect a `swapCoin`, we'd see a `gasTipCap == gasFeeCap` (a huge `gasTipCap`).
I guess we can solve this with documentation spelling out the logic of why a legacy txn can be treated as a dynamic tx with a massive tip, but I do think we should spell that out in both methods (`baseCoin` construction and `ValidateFeeRate`)

martonp: I added some docs.

PR #1451, opened 2022-01-26
site: latest stylelint standard, other updates

chappjc:
The dependabot vuln monitoring zealots say it's time to fix some things.

We are essentially forced to update stylelint v13 to v14 to resolve a [postcss dependency vuln](https://github.com/advisories/GHSA-566m-qj78-rww5) (which does not really matter for us since it's not runtime).  This was no big deal but it does move us to the latest `stylelint-config-standard`.

Of note is that postcss used by stylelint now [requires separate packages for **s**css syntax](https://stylelint.io/migration-guide/to-14/#syntax-option-and-automatic-inferral-of-syntax), so we've added the [`stylelint-config-standard-scss`](https://github.com/stylelint-scss/stylelint-config-standard-scss) package to our `devDependencies`.  Without that stylelint spews errors for each SCSS file like `When linting something other than CSS, you should install an appropriate syntax`, and you resolve that by extending `stylelint-config-standard-scss` in the stylelintrc.

The latest standard configs are a big more strict.  I've ignored a couple silly rules, but have resolved the `string-quotes` rule demand for consistency, settling on double quotes (the default anyway) because there were far fewer single quotes to change.


client/webserver/site/src/css/icons.scss
```diff
@@ -1,9 +1,9 @@
 /* icons */
 
-[class^="ico-"],
-[class*=" ico-"] {
+[class^=ico-],
+[class*=ico-] {
```
chappjc, line 4: This change may not be exactly as intended, but I think in practice it will match what we want.

```
 3:9   ✖  Unexpected quotes around "ico-"   selector-attribute-quotes
 4:9   ✖  Unexpected quotes around " ico-"  selector-attribute-quotes
```

Alternatively we can modify the `selector-attribute-quotes` rule to go with `"always"` and then add double quotes to selector attributes in other places (or go with `null` to just disable this rule).  https://stylelint.io/user-guide/rules/list/selector-attribute-quotes/

client/webserver/site/.stylelintrc
```diff
@@ -1,7 +1,17 @@
...
     "rules": {
         "at-rule-no-unknown": null,
-        "no-descending-specificity": null
+        "font-family-name-quotes": "always-unless-keyword",
+        "no-descending-specificity": null,
+        "property-no-vendor-prefix": null,
+        "selector-attribute-quotes": "never",
+        "selector-id-pattern": null,
+        "value-keyword-case": null,
+        "scss/dollar-variable-pattern": null,
```
chappjc, line 11: default pattern for these variable names is "kebab" case (`'^(-?[a-z][a-z0-9]*)(-[a-z0-9]+)*$'`) and I didn't see a good reason to go changing them all.

JoeGruffins (approved):
Works fine. I don't see any new console errors.

chappjc:
Updating with the following:

```
 @babel/core           ^7.16.12  →  ^7.17.2     
 @babel/eslint-parser   ^7.16.5  →  ^7.17.0     
 css-loader              ^6.5.1  →   ^6.6.0     
 eslint                  ^8.7.0  →   ^8.8.0     
 sass                   ^1.49.0  →  ^1.49.7     
 webpack                ^5.67.0  →  ^5.68.0  
```

PR #1452, opened 2022-01-27
dexc-webview

chappjc:
Just something I was toying with, **not intended for merging**.  No plans to take this much further, but wanted share.

However, it would be nice to extend this further so it does an `os.Exec` of dcrwallet in whatever way it requires.

A Windows build may be created that does not open a console/term with `go build -ldflags "-H=windowsgui"`

We should look into using Flutter in a different graphical package.


client/cmd/dexc-webview/config.go
```diff
@@ -0,0 +1,263 @@
...
+		cfg.Net = dex.Mainnet
+		defaultDBPath = setNet(preCfg.AppData, "mainnet")
+	}
+	defaultHost := defaultHostByNetwork(cfg.Net)
+
+	// If web or RPC server addresses not set, use network specific
+	// defaults
+	if cfg.WebAddr == "" {
+		cfg.WebAddr = net.JoinHostPort(defaultHost, defaultWebPort)
+	}
```
buck54321, line 196: We should just allow this to be auto-generated, i.e. `localhost:0`. The `Server.addr` field will still be set properly (since we're reading it from the `net.Listener`), and there's no chance of conflict.

client/cmd/dexc-webview/main.go
```diff
@@ -0,0 +1,247 @@
...
+
+	<-ready
+
+	runWebview := func() {
+		w := webview.New(true)
+		defer w.Destroy()
+		w.SetTitle("DCRDEX")
+		w.SetSize(1280, 1024, webview.HintNone)
+		w.Navigate("http://" + webSrv.Addr())
+		w.Run()
```
buck54321, line 197: Well, that's not too bad.

Issue #1311, opened 2021-11-25
eth: Manual refunds


martonp:
For BTC and DCR, we log a signed refund transaction that the user can manually submit to the blockchain in case the dex fails to properly refund a swap. For ETH this is not possible, because the nonce might have been incremented by the time the user would need to do a refund. We need an different way to do manual refunds for ETH.

chappjc:
In the past I had made a couple quick and dirty tools to do refunds and redeems without dexc.

https://github.com/decred/dcrdex/commit/cb15c2fefe2eb44a24060fd196fc9e86a4214046

- The `redeem` tool works like `redeem <txid> <vout> <contract script> <secret>`, but it can be made to work like `redeem <raw tx hex> <contract script> <secret>`.  This tool is useful if you are the taker, and the maker has redeemed your contract, but your dexc is disabled before you can redeem theirs.  With logged data, you can do the redeem manually.
- The `refundall` tool is a bit different in that it chugs through the dexc.log file and attempts to refund any unspent contract that was logged.  Instead of scanning the log file, a new `refund` tool can be made to work like `redeem` by processing a specified contract of your own.

These are not well generalized at all currently, only working with DCR and BTC, but that can change.

martonp:
Those look nice.. I was thinking about something similar to the `refund` tool that you mentioned where you specify a contract of your own, but it would definitely be more convenient to have something that just scans the log files.

With ETH they are already deterministic as long as we only use a single swap contract. Definitely would be harder with DCR/BTC. I guess we would need the wallet to keep track of all the unredeemed swaps in the UTXO set.

chappjc:
> With ETH they are already deterministic as long as we only use a single swap contract

Address reuse and the central contract with ETH certainly makes things easier, even if crazier.

A complementary idea that @buck54321 had suggested was a secondary record of the essential swap data, which would be redundant with both the DB and the log file, but potentially provide a more concise and easily parseable record of the critical on-chain swap data.

- another file such as "critical.txt", synchronous and flushed after each write
- match ID on each line to link all logged data
- more structured than the .log file so that it is easily machine readable
- any data that would be required to perform refund or taker redeems such as contract scripts, transaction IDs, counter party secrets

It's all already in the log file as I've demonstrated, but this would still be valuable IMO.

chappjc:
[...]

PR #1455, opened 2022-01-28
eth: Remove msg.sender check for refunds.

### LINKED ISSUE #1427: eth/contract_V0 [discussion]: Remove msg.sender == * checks.
Currently the message sender is checked to be one of the initiator or the participant for init, redeem, and refund. I propose we remove these checks for a few reasons:

1. The main benefit is that this would allow anyone to redeem or refund a swap. As you know, it takes gas to work with a contract. A user could get stuck in a position where they cannot pay the fee to do one of these. This can be a time sensitive operation for the taker, who may need to redeem or refund before the maker's locktime is passed to avoid losing funds. Having no requirement on the sender would allow a "good Samaritan" to help the user out, however unlikely that may be.

2. It may reduce gas usage by a small amount for initiations and refunds. For redeems, if we want to send all the funds at once I guess there will still need to be a check that all participants are the same address.

3. It is not needed. We choose the initiator/refunder and participant/redeemer at initiation. As long as anyone knows the secret, we can make sure they can only redeem to the participant. If locktime is past, we can ensure the refund only goes to the initiator. I believe we check now because that's what UTXO coins do. They have the limitation that one cannot choose who a utxo is spent to when constructing a script to be signed. We ensure it can only be signed by the owner of a certain private key, and assume the owner will spend to themselves. But with a contract we can specify that the locked value is only paid to a specific party. 

Of course please let me know if I overlooked another obvious reason for the checks, or if point 3 is not true.

JoeGruffins:
In order to allow any party to refund swaps, remove the check that
    refunder is the initiator so that anyone may refund after locktime.

closes #1427

Even after making the changes, I'm still a little on the fence. The savings I thought we would get do not seem to be there. For some reason a different refunder also takes more gas, although it does not use it all, which was unexpected. As discussed, these changes still may make #1311 easier to solve.

Refunds do work, but the `feeSuggestion` is too low currently. Probably solved in  #1447

The locked funds are slightly off after refunding.


martonp (approved):
Even though it doesn't have any gas savings, I still support the change. It adds a bit of flexibility, and there doesn't seem to be any downsides. I think reason there's no change in the gas is because we're removing one SLOAD operation where we look inside the `swapToRedeem` storage variable in `isRedeemable`, but replacing it with a new one in `redeem`. SLOAD is 2100 gas, but checking the `msg.sender` is very cheap.

JoeGruffins:
Thanks for the explanation about he inner working @martonp !

> What do you mean the locked funds are slightly off?

The locked funds are off by a small amount after refunding or any failed swap. Here after a refund. It fixes itself on dexc restart.
image

There's also an unrelated problem with the funding coin encoding in the ui here:
image

chappjc:
> I think reason there's no change in the gas is because we're removing one SLOAD operation where we look inside the `swapToRedeem` storage variable in `isRedeemable`, but replacing it with a new one in `redeem`. SLOAD is 2100 gas, but checking the `msg.sender` is very cheap.

I still support this change too, but I'm confused by it using **more gas** rather than the same amount.  Based on this explanation, I would expect the gas use change to be minus the amount for the very cheap `memory` access of `msg.sender` (or is that considered `calldata`?)

@JoeGruffins' comment:

> For some reason a different refunder also takes more gas, although it does not use it all

Sounds like it takes more gas, which is also described in the added code comment, but what does "it does not use it all" mean?  Also do you mean with these current code changes any refund takes more gas, or with these changes there's actually a difference who does the refund?

martonp:
I think what ha means is changing msg.sender to swap.initiator adds to the gas cost, but as much as removing the`msg.sender == swap.initiator` reduced it by. 

I'll try to debug those bugs.

JoeGruffins:
> with these changes there's actually a difference who does the refund?

This. The initiator can refund with the current amount no problem. Another party (I used the participant) will take an upfront 46000 but use about 43103 or so, as can be viewed from the receipt. I'm guessing the msg.sender already has some form of verification done, so we can send there more easily, but just a guess.

chappjc (approved):

chappjc:
A bit mysterious till, but alright.

I'll make issue for the two things @JoeGruffins pointed out.

client/asset/eth/eth.go
```diff
@@ -1089,6 +1089,23 @@ func (eth *ExchangeWallet) Refund(_, contract dex.Bytes, feeSuggestion uint64) (
...
+		eth.unlockFunds(eth.gasFeeLimit * dexeth.RefundGas(version))
+	}
+
+	swap, err := eth.node.swap(eth.ctx, secretHash, version)
+	if err != nil {
+		return nil, err
+	}
+	// It's possible the swap was refunded by someone else. In that case we
+	// cannot know the refunding tx hash.
+	if swap.State == dexeth.SSRefunded {
```
chappjc, line 1102: BTW, I was sure the `Core`'s `isRefundable` checks would need something to account for this possibility, but it does not appear to matter.  Hopefully were not missing something.

PR #1456, opened 2022-01-31
client/core: allow wallet Lock to timeout

chappjc:
For various reasons, some unknown, it is apparently quite common for wallet lock to hang either client shutdown or just logout.
This is particularly problematic because it leaves the user no choice but to hard kill the process, which may have undesirable consequences.   Even when it is just slow to lock, most users decide it's hung forever much sooner.  I'm estimating this threshold at around 5-10 seconds.

This PR wraps the `wallet.Lock` calls with a timeout contraption.  The individual implementations could also be modified with say a `context.Context` or a `time.Duration` argument, but in cases where the backend is not context enabled it ends up like this anyway, just with the backend worrying about this and the caller (core) being ultimately at the mercy of the backends actually doing it, back to possible hangs.

Also of note is that this PR makes `wallet.Lock` errors no longer prevent full `Logout` (and thus clean shutdown) possible.  This was only encouraging dangerous killing of the process.

Related TODO: rework the **btc** backend so that the `call` method accepts a `context.Context`.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -75,7 +75,7 @@ var (
 	// orders.
 	ActiveOrdersLogoutErr = errors.New("cannot log out with active orders")
 
-	errWalletTimeout = errors.New("timeout")
+	errTimeout = errors.New("timeout")
 )
 
 type dexTicker struct {
@@ -1273,28 +1273,13 @@ func (c *Core) Run(ctx context.Context) {
 		}
 		symb := strings.ToUpper(unbip(assetID))
 		c.log.Infof("Locking %s wallet", symb)
-		if err := walletLockWithTimeout(5*time.Second, wallet); err != nil {
+		if err := wallet.Lock(5 * time.Second); err != nil {
 			c.log.Errorf("Failed to lock %v wallet: %v", symb, err)
 		}
 		wallet.Disconnect()
 	}
 
 	c.log.Infof("DEX client core off")
-}
-
-func walletLockWithTimeout(timeout time.Duration, wallet *xcWallet) error {
-	errChan := make(chan error, 1)
-	go func() {
-		defer close(errChan)
-		errChan <- wallet.Lock()
-	}()
-
-	select {
-	case err := <-errChan:
-		return err
-	case <-time.After(timeout):
-		return errWalletTimeout
-	}
 }
 
 // Ready returns a channel that is closed when Run completes its initialization
@@ -1760,7 +1745,7 @@ func (c *Core) CreateWallet(appPW, walletPW []byte, form *WalletForm) error {
 	}
 
 	initErr := func(s string, a ...interface{}) error {
-		_ = walletLockWithTimeout(2*time.Second, wallet) // just try, but don't confuse the user with an error
+		_ = wallet.Lock(2 * time.Second) // just try, but don't confuse the user with an error
 		wallet.Disconnect()
 		return fmt.Errorf(s, a...)
 	}
@@ -2039,7 +2024,7 @@ func (c *Core) CloseWallet(assetID uint32) error {
 	if err != nil {
 		return fmt.Errorf("wallet not found for %d -> %s: %w", assetID, unbip(assetID), err)
 	}
-	err = walletLockWithTimeout(5*time.Second, wallet)
+	err = wallet.Lock(5 * time.Second)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


client/core/core.go
```diff
@@ -1747,7 +1760,7 @@ func (c *Core) CreateWallet(appPW, walletPW []byte, form *WalletForm) error {
 	}
 
 	initErr := func(s string, a ...interface{}) error {
-		_ = wallet.Lock() // just try, but don't confuse the user with an error
+		_ = walletLockWithTimeout(2*time.Second, wallet) // just try, but don't confuse the user with an error
```
chappjc, line 1763: The error is ignored anyway and the wallet is immediately disconnected...

client/core/core.go
```diff
@@ -3283,8 +3296,10 @@ func (c *Core) Logout() error {
 	// Lock wallets
 	for _, w := range c.xcWallets() {
 		if w.connected() {
-			if err := w.Lock(); err != nil {
-				return err
```
chappjc, line 3287: This was a major contributor to hard kill user behavior

client/core/core.go
```diff
@@ -1272,18 +1273,30 @@ func (c *Core) Run(ctx context.Context) {
...
-	c.walletWait.Wait()
-
 	c.log.Infof("DEX client core off")
 }
 
+func walletLockWithTimeout(timeout time.Duration, wallet *xcWallet) error {
+	errChan := make(chan error, 1)
+	go func() {
+		defer close(errChan)
+		errChan <- wallet.Lock()
```
chappjc, line 1289: It's a little wild to just leave this hung if it takes too long, but I don't see much choice other than trusting the individual wallet implementations (and wiring them with a context or timeout which they could not respect).

I'm fairly comfortable with this sort of abandonment approach in the shutdown paths of `Run` and `Logout` (likely before shutdown).  The other uses are not at the end of `Core` lifetime, but they are equally problematic if there are hangs while, say, clicking the "Lock" button.

client/core/core.go
buck54321: Any ideas what the hold up was? I'm in favor of this secondary timeout, but I do wonder what the wallet thinks it's waiting on.

client/core/core.go
```diff
@@ -1272,18 +1273,30 @@ func (c *Core) Run(ctx context.Context) {
...
 		wallet.Disconnect()
 	}
 
-	// Let the wallet backends shut down cleanly.
-	c.walletWait.Wait()
-
 	c.log.Infof("DEX client core off")
 }
 
+func walletLockWithTimeout(timeout time.Duration, wallet *xcWallet) error {
```
buck54321, line 1285: It looks like you can just add a timeout argument to `(*xcWallet).Lock` and move this there, since `walletLockWithTimeout` is the only place it's used now.

client/core/core.go
buck54321: > giving asset.Wallet.Lock a context or timeout only gets us so far if those backends hang anyway.

That's the crux of it, I think. This is a reasonable move.

client/core/core.go
chappjc: Unclear, but personally I have had bitcoin hang on RPCs when there is peer trouble, and presumably some broadly-scoped internal mutex (in bitciond) is locked for a long time that holds up all RPCs.

client/core/core.go
chappjc: Gave this a thumbs up, but it's looking really quite smart to move it like you said because inside `(*xcWallet).Lock` has `xcWallet.mtx` is locked while it sits on `asset.Wallet.Lock`  Looks like we'd have ended up with deadlocks anyway if there were other `xcWallet` method calls.
Going to try to ensure the mutex is not locked while it's waiting on the backend.

client/core/wallet.go
```diff
@@ -86,34 +107,37 @@ func (w *xcWallet) refreshUnlock() (unlockAttempted bool, err error) {
...
 		return false, nil // unlocked
 	}
+
 	// Locked backend requires both encrypted and decrypted passwords.
-	if len(w.encPW()) == 0 {
-		return false, fmt.Errorf("%s wallet reporting as locked but no password has been set", unbip(w.AssetID))
-	}
 	w.mtx.RLock()
-	defer w.mtx.RUnlock()
-	if len(w.encPass) == 0 {
```
chappjc, line 95: These were duplicate checks.

client/core/wallet.go
```diff
@@ -86,34 +107,37 @@ func (w *xcWallet) refreshUnlock() (unlockAttempted bool, err error) {
...
+	pwUnset := len(w.encPass) == 0
+	locked := len(w.pw) == 0
+	w.mtx.RUnlock()
```
chappjc, line 115: This is the **un**locking method, but blocking other methods like `(*xcWallet).state` was probably an issue.  With bitcoind and huge numbers of keys, it is reported that Unlock can take several minutes to complete.

buck54321 (approved):

PR #1457, opened 2022-02-01
client/webserver/core: Link to wallet log file

### LINKED ISSUE #1437: link to neutrino.log, maybe from wallet settings form

martonp:
Adds a link to the wallet page to download the neutrino log file for the BTC SPV wallet.

Closes #1437

### DIFF
--- a/client/asset/interface.go
+++ b/client/asset/interface.go
@@ -17,6 +17,7 @@ type WalletTrait uint64
 const (
 	WalletTraitRescanner    WalletTrait = 1 << iota // The Wallet is an asset.Rescanner.
 	WalletTraitNewAddresser                         // The Wallet can generate new addresses on demand with NewAddress.
+	WalletTraitLogFiler                             // The Wallet allows for downloading of a log file.
 )
 
 // IsRescanner tests if the WalletTrait has the WalletTraitRescanner bit set.
@@ -32,6 +33,11 @@ func (wt WalletTrait) IsNewAddresser() bool {
 	return wt&WalletTraitNewAddresser != 0
 }
 
+// IsLogFiler tests if WalletTrait has the WalletTraitLogFiler bit set.
+func (wt WalletTrait) IsLogFiler() bool {
+	return wt&WalletTraitLogFiler != 0
+}
+
 // DetermineWalletTraits returns the WalletTrait bitset for the provided Wallet.
 func DetermineWalletTraits(w Wallet) (t WalletTrait) {
 	if _, is := w.(Rescanner); is {
@@ -39,6 +45,9 @@ func DetermineWalletTraits(w Wallet) (t WalletTrait) {
 	}
 	if _, is := w.(NewAddresser); is {
 		t |= WalletTraitNewAddresser
+	}
+	if _, is := w.(LogFiler); is {
+		t |= WalletTraitLogFiler
 	}
 	return t
 }
@@ -296,6 +305,11 @@ type Sweeper interface {
 // NewAddresser is a wallet that can generate new deposit addresses.
 type NewAddresser interface {
 	NewAddress() (string, error)
+}
+
+// LogFiler is a wallet that allows for downloading of its log file.
+type LogFiler interface {
+	LogFilePath() string
 }
 
 // TokenMaster is implemented by assets which support degenerate tokens.
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -6540,13 +6540,13 @@ func walletDefinition(assetID uint32, walletType string) (*asset.WalletDefinitio
 	return walletDef, nil
 }
 
-// GetWalletLogFilePath returns the path to the wallet's log file.
-func (c *Core) GetWalletLogFilePath(assetID uint32) (string, error) {
+// WalletLogFilePath returns the path to the wallet's log file.
+func (c *Core) WalletLogFilePath(assetID uint32) (string, error) {
 	wallet, exists := c.wallet(assetID)
 	if !exists {
 		return "", newError(missingWalletErr, "no configured wallet found for %s (%d)",
 			strings.ToUpper(unbip(assetID)), assetID)
 	}
 
-	return wallet.GetLogFilePath()
+	return wallet.LogFilePath()
 }
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -6539,3 +6539,14 @@ func walletDefinition(assetID uint32, walletType string) (*asset.WalletDefinitio
 	}
 	return walletDef, nil
 }
+
+// GetWalletLogFilePath returns the path to the wallet's log file.
+func (c *Core) GetWalletLogFilePath(assetID uint32) (string, error) {
+	wallet, exists := c.wallet(assetID)
+	if !exists {
+		return "", newError(missingWalletErr, "no configured wallet found for %s (%d)",
+			strings.ToUpper(unbip(assetID)), assetID)
+	}
+
+	return wallet.GetLogFilePath()
+}
--- a/client/webserver/http.go
+++ b/client/webserver/http.go
@@ -5,8 +5,10 @@ package webserver
 
 import (
 	"encoding/csv"
+	"fmt"
 	"io"
 	"net/http"
+	"os"
 	"sort"
 	"strconv"
 	"strings"
@@ -25,6 +27,7 @@ const (
 	loginRoute       = "/login"
 	marketsRoute     = "/markets"
 	walletsRoute     = "/wallets"
+	walletLogRoute   = "/wallets/logfile"
 	settingsRoute    = "/settings"
 	ordersRoute      = "/orders"
 	exportOrderRoute = "/orders/export"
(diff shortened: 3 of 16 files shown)


client/webserver/http.go
```diff
@@ -145,6 +148,81 @@ func (s *WebServer) handleWallets(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	logFilePath, err := s.core.GetWalletLogFilePath(uint32(assetID))
+	if err != nil {
+		log.Errorf("failed to get log file path %v", err)
+		http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
+		return
+	}
+
+	logFile, err := os.Open(logFilePath)
```
chappjc, line 188: I was totally thinking something lame like a `file:///` URL that would only work on the same machine, but this is great!

client/webserver/http.go
```diff
@@ -145,6 +148,81 @@ func (s *WebServer) handleWallets(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		log.Errorf("error opening log file: %v", err)
+		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+		return
+	}
```
chappjc, line 193: Need a defer `logFile.Close()` here.

client/webserver/http.go
```diff
@@ -145,6 +148,81 @@ func (s *WebServer) handleWallets(w http.ResponseWriter, r *http.Request) {
...
+	logFileName := fmt.Sprintf("dcrdex-%s-wallet.log", assetName)
+	w.Header().Set("Content-Disposition", "attachment; filename="+logFileName)
+	w.Header().Set("Content-Type", "text/plain")
+	w.WriteHeader(http.StatusOK)
+
+	bufferLength := int64(1024)
+	buffer := make([]byte, bufferLength)
+	offset := int64(0)
+	for {
+		n, readErr := logFile.ReadAt(buffer, offset)
```
chappjc, line 205: I'd typically go with a simple [`io.Copy`](https://pkg.go.dev/io#Copy) here since it essentially abstracts this buffered for loop pattern, with a default 32 kiB buffer.  Was there a particular reason for this approach?

client/webserver/http.go
```diff
@@ -145,6 +148,81 @@ func (s *WebServer) handleWallets(w http.ResponseWriter, r *http.Request) {
...
+	logFile, err := os.Open(logFilePath)
+	if err != nil {
+		log.Errorf("error opening log file: %v", err)
+		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+		return
+	}
+
+	assetName := dex.BipIDSymbol(uint32(assetID))
+	logFileName := fmt.Sprintf("dcrdex-%s-wallet.log", assetName)
+	w.Header().Set("Content-Disposition", "attachment; filename="+logFileName)
```
chappjc, line 197: If it weren't for the file potentially being written to concurrently I'd say we could even set the content length.  I wonder how this will work out if there's an exclusive lock on the log file.  Normally it's fine to read it though.

client/asset/interface.go
```diff
@@ -40,6 +46,9 @@ func DetermineWalletTraits(w Wallet) (t WalletTrait) {
 	if _, is := w.(NewAddresser); is {
 		t |= WalletTraitNewAddresser
 	}
+	if _, is := w.(LogFiler); is {
+		t |= WalletTraitLogFiler
```
chappjc, line 50: Glad we're getting use out of these traits.  I had my reservations about a bitset limited to 64 traits though, crossing fingers we have enough.  :laughing:

client/webserver/http.go
martonp: I'm glad you like it :)

client/webserver/http.go
martonp: No particular reason, I'll switch it

martonp: Actually I think it makes sense to keep it, because I'm not just copying the file to the `http.ResponseWriter`, but I'm also calling `flush` after each section of the file.

client/webserver/http.go
martonp: I think an exclusive lock that doesn't allow reads would be counterproductive for a log file.

client/webserver/http.go
chappjc: If you want.  Can't you just flush at the end though?  It's just that I never bother to do it the long way, never saw a reason.

chappjc: That would happen if we did ReadFile, but io.Copy is buffered.  After one section, it's cannot go on to the next until it's written to the ResponseWriter, and that itself has as a finite write buffer.  If I'm wrong I'd love to learn, but I don't see the file being loaded into memory with io.Copy without explicit flushes after each chunk.

martonp: Yeah I think you're right. But in that case what's the point of explicitly flushing when generating the csv file?  Wouldn't the `ResponseWriter` buffer fill up there as well, and then get flushed?

chappjc: So this was fun!  Out of curiosity I setup a little file server in a unit test and stepped into the code.  It got to this point in `buffio.(*Writer).Write`

image

image

That's from `io.Copy` > `net/http.(*response).ReadFrom` > `io.CopyBuffer` > ... > `net/http.(*response).write`.  The `b *Writer` there is a `net/http.chunkWriter` wrapping the actual `net/http.conn` that is alternately `Flush`ed and `Write`n to depending on the state of the buffer.

This was a fun check as I've never really confirmed what paths in the net/http code are actually used.

```go
func TestFile(t *testing.T) {
	f, err := os.Open("/home/jon/dexc-0.4.tar.gz")
	if err != nil {
		t.Fatal(err)
	}
	defer f.Close()

	var server *http.Server

	router := chi.NewRouter()
	router.Get("/", func(w http.ResponseWriter, r *http.Request) {
		io.Copy(w, f)
	})
	router.Get("/shutdown", func(w http.ResponseWriter, r *http.Request) {
		time.AfterFunc(time.Second, func() {
			server.Shutdown(context.Background())
		})
	})

	server = &http.Server{Addr: "127.0.0.1:8888", Handler: router}
	server.ListenAndServe()
}
```

martonp: Nice, I should do this more often. I'll just remove it from both places then.

JoeGruffins (approved):
Working well for me.

client/webserver/site/src/js/wallets.js
```diff
@@ -510,6 +517,15 @@ export default class WalletsPage extends BasePage {
     Doc.show(a.unlock)
   }
 
+  async downloadLogs () {
+    const search = new URLSearchParams('')
+    search.append('assetid', this.reconfigAsset + '')
```
chappjc, line 522: Is this a JS trick to make a number into a string?  I feel like it would be more expressive with `${this.reconfigAsset}` or even `String(this.reconfigAsset)` even if they aren't turbo fast.

client/core/core.go
```diff
@@ -6539,3 +6539,14 @@ func walletDefinition(assetID uint32, walletType string) (*asset.WalletDefinitio
 	}
 	return walletDef, nil
 }
+
+// GetWalletLogFilePath returns the path to the wallet's log file.
+func (c *Core) GetWalletLogFilePath(assetID uint32) (string, error) {
```
chappjc, line 6544: I'm interested in your opinion on this, but generally I think it's best to omit "get" function name prefixes unless it's alluding to some kind of actual remote fetch operation.  The getting goes on only in the webserver, so let's nuke the prefix from the Core and xcWallet methods.

client/webserver/http.go
```diff
@@ -145,6 +148,53 @@ func (s *WebServer) handleWallets(w http.ResponseWriter, r *http.Request) {
...
+	}
+	defer logFile.Close()
+
+	assetName := dex.BipIDSymbol(uint32(assetID))
+	logFileName := fmt.Sprintf("dcrdex-%s-wallet.log", assetName)
+	w.Header().Set("Content-Disposition", "attachment; filename="+logFileName)
+	w.Header().Set("Content-Type", "text/plain")
+	w.WriteHeader(http.StatusOK)
+
+	io.Copy(w, logFile)
```
chappjc, line 195: Although it's too late to do anything with an error from `io.Copy` (the header is already written), we can log the error if it's non-nil.

client/webserver/http.go
```diff
@@ -302,7 +344,6 @@ func (s *WebServer) handleExportOrders(w http.ResponseWriter, r *http.Request) {
 			log.Errorf("error writing CSV: %v", err)
 			return
 		}
-		wf.Flush()
```
chappjc, line 305: ACK.  Just reiterating that I don't think there's a need to `Flush` the `http.ResponseWriter` at the end of an `http.Handler`, so good to remove these.   Flushing the csvWriter on the other hand make sense to me, and that's done.

chappjc (changes requested):
Working very well for me.  A few minor suggested edits.

client/core/core.go
martonp: That makes sense, I agree.

client/webserver/site/src/js/wallets.js
martonp: I'll update to the template style.

chappjc (approved):

chappjc:
Good stuff. Leaving for @buck54321 though

buck54321 (approved):
Nice!

PR #1461, opened 2022-02-01
client/asset/eth: Fix decode coin id

### LINKED ISSUE #1459: site: eth coin encoding on match page is invalid
image

martonp:
The driver's DecodeCoinID function was attempting to parse the eth
coin's recovery address instead of the byte encoded address string.

Closes #1459.

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -133,20 +133,27 @@ func (d *Driver) Open(cfg *asset.WalletConfig, logger dex.Logger, network dex.Ne
 }
 
 // DecodeCoinID creates a human-readable representation of a coin ID for Ethereum.
+// In ETH there are 3 possible coin IDs:
+//   1. A transaction hash
+//   2. An encoded funding coin id which includes the account address and amount.
+//   3. A byte encoded string of the account address.
 func (d *Driver) DecodeCoinID(coinID []byte) (string, error) {
-	const invalid = "<invalid coin>"
-	if len(coinID) == common.HashLength {
-		id, err := dexeth.DecodeCoinID(coinID)
-		if err != nil {
-			return invalid, err
-		}
-		return id.String(), nil
+	txHashID, err := dexeth.DecodeCoinID(coinID)
+	if err == nil {
+		return txHashID.String(), nil
 	}
-	id, err := decodeFundingCoinID(coinID)
-	if err != nil {
-		return invalid, err
+
+	fundingCoinID, err := decodeFundingCoinID(coinID)
+	if err == nil {
+		return fundingCoinID.String(), nil
 	}
-	return id.String(), nil
+
+	addressID := string(coinID)
+	if len(addressID) == 42 && addressID[0:2] == "0x" {
+		return addressID, nil
+	}
+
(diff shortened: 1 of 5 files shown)


client/asset/eth/eth.go
```diff
@@ -142,11 +142,13 @@ func (d *Driver) DecodeCoinID(coinID []byte) (string, error) {
 		}
 		return id.String(), nil
 	}
-	id, err := decodeFundingCoinID(coinID)
-	if err != nil {
-		return invalid, err
+
+	id := string(coinID)
+	if len(id) != 42 || id[0:2] != "0x" {
+		return invalid, fmt.Errorf("coin is not byte encoded utf-8 string of ETH address: %x", coinID)
```
chappjc, line 148: Even now it's difficult to keep track of the coin ID variations for ETH.  Let me see if I can summarize.

1. a tx hash
2. just an address/account, utf-8 encoded
3. funding coin / recover encoding, the bytes of account || amount

So, the `eth.fundingCoin` type can give both the utf-8 encoded account via `ID()` or a funding coin encoding via `RecoveryID()`.   The plain utf-8 encoded address/account is used when placing orders for giving server an account without the amount.

Most places with server interaction for the actual swap will use the `eth.coin` type with a tx hash encoded.

What ends up in a `core.Order` returned by `(*Core).Order` (used by the http endoint for this page, I think) comes from `coreOrderFromTrade`, which gets it from `ord.Trade().Coins[]`.  These guys are set in `prepareTrackedTrade` from calling the `eth.(*fundingCoin).ID` method that returns the utf-8 encoded account as described above.
Finally, this `Driver` method you are fixing is used by `core.NewCoin` in `coreOrderFromTrade`?

:sweat: 

So I'm wondering what else uses this driver method and if it  needs to be designed to work with the recovery encoding (3) as well as the tx id and account utf-8.

martonp: Yeah you're right we need all three, my bad.

JoeGruffins: I wonder if encoding a bit in the eth coin id would make it more organized, jk

client/asset/eth/eth.go
```diff
@@ -133,20 +133,27 @@ func (d *Driver) Open(cfg *asset.WalletConfig, logger dex.Logger, network dex.Ne
...
+		return fundingCoinID.String(), nil
 	}
-	return id.String(), nil
+
+	addressID := string(coinID)
+	if len(addressID) == 42 && addressID[0:2] == "0x" {
+		return addressID, nil
+	}
+
+	return "", fmt.Errorf("%x is not a valid ETH coin id", coinID)
```
JoeGruffins, line 156: Some call sites still forward the string here, as we saw in the UI, so imo it is useful to continue returning "invalid" on error.

client/asset/eth/eth.go
chappjc: Lololol

JoeGruffins (approved):
Ok!

image

client/asset/eth/eth.go
martonp: OK, I updated BTC and DCR as well.

chappjc (approved):

PR #1462, opened 2022-02-02
web/markets: Remove comma from field int value.

### LINKED ISSUE #1442: ui: Unable to trade with 4 digit+ coin.
Testing on v0.4.
image

The dcr field is set to nothing with any value over 999. It is possible to make the trade via rpc, which looks something like `$ ./dexcctl --simnet trade 127.0.0.1:17273 false true 42 0 100000000000 1000000 true`

I originally thought this was just market trades but also looks true for limit trades. It is wierd that I haven't noticed before, so please check, maybe I've messed something up with different versions.

JoeGruffins:
Doc's formatCoinValue will stick a comma in the value that we save to an
    int field, causing it not to save at all, preventing four or more digit
    coin values from being traded with my language settings. Prevent this by
    using a simpler conversion that does not insert localized punctuation.

closes #1442

### DIFF
--- a/client/webserver/site/src/js/markets.js
+++ b/client/webserver/site/src/js/markets.js
@@ -687,8 +687,8 @@ export default class MarketsPage extends BasePage {
       sell: sell,
       base: market.base.id,
       quote: market.quote.id,
-      qty: convertConventional(qtyField.value, market.baseUnitInfo.conventional.conversionFactor),
-      rate: convertConventional(page.rateField.value, market.rateConversionFactor), // message-rate
+      qty: convertToAtoms(qtyField.value, market.baseUnitInfo.conventional.conversionFactor),
+      rate: convertToAtoms(page.rateField.value, market.rateConversionFactor), // message-rate
       tifnow: page.tifNow.checked,
       options: this.orderOpts
     }
@@ -1677,7 +1677,8 @@ export default class MarketsPage extends BasePage {
     }
     const lotSize = this.market.cfg.lotsize
     page.lotField.value = lots
-    page.qtyField.value = Doc.formatCoinValue(lots * lotSize, this.market.baseUnitInfo)
+    // Conversion factor must be a multiple of 10.
+    page.qtyField.value = lots * lotSize / this.market.baseUnitInfo.conventional.conversionFactor
     this.previewQuoteAmt(true)
   }
 
@@ -1698,7 +1699,8 @@ export default class MarketsPage extends BasePage {
     const adjusted = lots * lotSize
     page.lotField.value = lots
     if (!order.isLimit && !order.sell) return
-    if (finalize) page.qtyField.value = Doc.formatCoinValue(adjusted, this.market.baseUnitInfo)
+    // Conversion factor must be a multiple of 10.
+    if (finalize) page.qtyField.value = adjusted / this.market.baseUnitInfo.conventional.conversionFactor
     this.previewQuoteAmt(true)
   }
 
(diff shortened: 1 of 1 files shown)


client/webserver/site/src/js/markets.js
```diff
@@ -2340,8 +2340,13 @@ function makeMarket (host, base, quote) {
 /* marketID creates a DEX-compatible market name from the ticker symbols. */
 export function marketID (b, q) { return `${b}_${q}` }
 
-/* convertConventional converts the float string to atoms. */
+/* convertConventional converts the float string to conventional units. */
 function convertConventional (s, conversionFactor) {
+  return Math.round(parseFloat(s) / conversionFactor)
```
chappjc, line 2345: I don't think we want a `round` after dividing to get conventional units.

chappjc (changes requested):

client/webserver/site/src/js/markets.js
JoeGruffins: Should have been floor.

client/webserver/site/src/js/markets.js
```diff
@@ -2340,8 +2340,13 @@ function makeMarket (host, base, quote) {
 /* marketID creates a DEX-compatible market name from the ticker symbols. */
 export function marketID (b, q) { return `${b}_${q}` }
 
-/* convertConventional converts the float string to atoms. */
+/* convertConventional converts the float string to conventional units. */
 function convertConventional (s, conversionFactor) {
```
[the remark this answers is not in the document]
JoeGruffins: I see. I interpreted it as To I guess. I think the method in Doc is like that.

client/webserver/site/src/js/markets.js
chappjc: Wait, we want decimals in conventional units...

JoeGruffins: floor is right right?

client/webserver/site/src/js/markets.js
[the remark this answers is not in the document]
JoeGruffins: Change the naming to be like the Doc methods.

client/webserver/site/src/js/markets.js
chappjc: Don't think so.  Say there was a market with a lot size of 5000000 atoms (0.5 DCR).  This quantity field would need decimals

chappjc: I think there's some unnecessary back and forth conversions, but it works right.

image

image

So I kinda thought this would just be `return parseFloat(s) * conversionFactor`.
If we find we need to deal with floating point precision issues, we can do all this string->float->string business, but I think the precision would not be from the unitInfo's conversion factor but rather lot size in atoms.

JoeGruffins: Yeah, `return parseFloat(s) / conversionFactor` is fine right? Won't it always be 1000000 or something like that? I don't know why I would expect something different.

chappjc (approved):
Works for me.

JoeGruffins:
srry this is taking so much time. First argument should be an int correct?

JoeGruffins:
Just rebased.

client/webserver/site/src/js/markets.js
```diff
@@ -1677,7 +1677,7 @@ export default class MarketsPage extends BasePage {
     }
     const lotSize = this.market.cfg.lotsize
     page.lotField.value = lots
-    page.qtyField.value = Doc.formatCoinValue(lots * lotSize, this.market.baseUnitInfo)
+    page.qtyField.value = convertToConventional(lots * lotSize, this.market.baseUnitInfo.conventional.conversionFactor)
```
buck54321, line 1680: Seems like the function is unnecessary. It's just division.

buck54321 (approved):

client/webserver/site/src/js/markets.js
JoeGruffins: Will remove.

JoeGruffins: Removed.

PR #1463, opened 2022-02-02
charts: Make sure regions are defined in render.

### LINKED ISSUE #1357: eth/trading: Missing graph rates and cannot buy/sell.
After setting up the simnet harness and attempting to trade, a couple problems were encountered. Maybe simple misconfigurations.

Missing the graph rates, here is the dcr/btc graph:
image

And dcr/eth is missing the rates at the bottom and vertical lines:
image

The second problem happens when attempting to trade, buy or sell:
image

### LINKED ISSUE #1450: web: `e.zoomOutBttn is undefined` causing long hang.
As shown here, on master (ecc1f79), sometimes I get a long hang from this js error.

image

```
Uncaught TypeError: e.zoomOutBttn is undefined
    value charts.js:492
    value charts.js:1020
    value charts.js:457
    value charts.js:88
    value charts.js:332
    value markets.js:884
    value markets.js:1077
    P markets.js:184
    pr ws.js:37
    onmessage ws.js:101
```

JoeGruffins:
testing/dcrdex: Reduce dcr/eth rate step. 

    charts: Make sure regions are defined in render.

    There was an apparent race between the resized method and render. If
    render would happen first some elements that are initiated in resized
    were not yet available, causing a js error and hang. Set those fields in
    the constructor to avoid errors.

closes #1450,  closes #1357

I think this solves it.

I have not seen an error with the candle graph, so those changes may not be necessary. I think we se it with the normal graph because that is what we open the page to.

Also reduce the rate step in the simnet harness so the rates show.

### DIFF
--- a/client/webserver/site/src/js/charts.js
+++ b/client/webserver/site/src/js/charts.js
@@ -53,6 +53,7 @@ class Chart {
     this.ctx = this.canvas.getContext('2d')
     this.ctx.textAlign = 'center'
     this.ctx.textBaseline = 'middle'
+    this.setZoomBttns()
     // Mouse handling
     this.mousePos = null
     bind(this.canvas, 'mousemove', e => {
@@ -87,6 +88,10 @@ class Chart {
   draw () {
     this.render()
   }
+
+  // setZoomBttns is run before drawing and should be used for setup of zoom
+  // buttons.
+  setZoomBttns () {}
 
   /* click is the handler for a click event on the canvas. */
   click (e) {
@@ -277,11 +282,17 @@ export class DepthChart extends Chart {
     this.resize(parent.clientHeight)
   }
 
+  // setZoomBttns creates new regions for zoom in and zoom out buttons. It is
+  // used in initiation of the buttons and resizing.
+  setZoomBttns () {
+    this.zoomInBttn = new Region(this.ctx, new Extents(0, 0, 0, 0))
+    this.zoomOutBttn = new Region(this.ctx, new Extents(0, 0, 0, 0))
+  }
+
   /* resized is called when the window or parent element are resized. */
   resized () {
     // The button region extents are set during drawing.
-    this.zoomInBttn = new Region(this.ctx, new Extents(0, 0, 0, 0))
-    this.zoomOutBttn = new Region(this.ctx, new Extents(0, 0, 0, 0))
+    this.setZoomBttns()
     if (this.book) this.draw()
   }
 
--- a/dex/testing/dcrdex/harness.sh
+++ b/dex/testing/dcrdex/harness.sh
@@ -106,7 +106,7 @@ if [ $ETH_ON -eq 0 ]; then
             "base": "DCR_simnet",
             "quote": "ETH_simnet",
             "lotSize": 1000000000,
-            "rateStep": 1000000000,
+            "rateStep": 1000,
             "epochDuration": ${EPOCH_DURATION},
             "marketBuyBuffer": 1.2
 EOF


client/webserver/site/src/js/charts.js
```diff
@@ -692,13 +701,19 @@ export class CandleChart extends Chart {
     this.resize(parent.clientHeight)
   }
 
-  /* resized is called when the window or parent element are resized. */
-  resized () {
+  // setRegions creates new regions for candles and volume. It is used in
+  // initation of the regions and resizing.
```
chappjc, line 705: initation

client/webserver/site/src/js/charts.js
```diff
@@ -818,6 +833,9 @@ export class CandleChart extends Chart {
 
     // Draw the volume bars.
     const volDataExtents = new Extents(start(first), end(last), 0, highVol)
+    if (!this.volumeRegion) {
+      this.setRegions()
```
chappjc, line 837: As you suggested in the PR description, this probably isn't needed for the candle chart.  I bet this will be impossible to hit because `CandleChart`'s constructor calls `resize` (and then `resized` where `volumeRegion` is set), but this seems fine.

client/webserver/site/src/js/charts.js
```diff
@@ -277,11 +277,17 @@ export class DepthChart extends Chart {
...
+    this.zoomInBttn = new Region(this.ctx, new Extents(0, 0, 0, 0))
+    this.zoomOutBttn = new Region(this.ctx, new Extents(0, 0, 0, 0))
+  }
+
   /* resized is called when the window or parent element are resized. */
   resized () {
     // The button region extents are set during drawing.
-    this.zoomInBttn = new Region(this.ctx, new Extents(0, 0, 0, 0))
-    this.zoomOutBttn = new Region(this.ctx, new Extents(0, 0, 0, 0))
+    this.setZoomBttns()
```
chappjc, line 290: I suppose this makes sense you could get here before the `DepthChart` constructor is finished because the parent class (`Chart`) `resize` calls the child's `resized`?  A little confusing. I wonder if there's a better pattern to prevent this possibility.  But this works...

chappjc (changes requested):
Just the typo.  The other comments are just some thoughts about the class designs.

client/webserver/site/src/js/charts.js
JoeGruffins: I'm sure there's a better way to design this... but I don't know it offhand.

JoeGruffins: This seems more correct... to me. Make sure it is initiated before draw and we don't need to check later.

client/webserver/site/src/js/charts.js
JoeGruffins: Just removed since probably not needed.

client/webserver/site/src/js/charts.js
```diff
@@ -277,11 +281,22 @@ export class DepthChart extends Chart {
...
+  // used in initiation of the buttons and resizing.
+  setZoomBttns () {
+    this.zoomInBttn = new Region(this.ctx, new Extents(0, 0, 0, 0))
+    this.zoomOutBttn = new Region(this.ctx, new Extents(0, 0, 0, 0))
+  }
+
+  // initRegions sets up the zoom button regions.
+  initRegions () {
+    this.setZoomBttns()
+  }
```
chappjc, line 294: Seems like you can get rid one of these two methods now.

chappjc (approved):
That seems to do it.  But I think we can get rid of `initRegions` and just have `setZoomBttns` in the parent.

client/webserver/site/src/js/charts.js
JoeGruffins: Ok.

chappjc:
Sorry didn't want to squash.

PR #1465, opened 2022-02-04
client/asset/eth: Incorrect locked amount

### LINKED ISSUE #1460: eth: locked funds remain after match revoke
The locked funds are off by a small amount after refunding or any failed swap. Here after a refund. It fixes itself on dexc restart.
image

_Originally posted by @JoeGruffins in https://github.com/decred/dcrdex/issues/1455#issuecomment-1025305247_

martonp:
Pending outgoing funds were being previously incorrectly being added to the locked amount. This is no longer happening. Also, attempting to unlock more than is locked is no longer an error. The locked amount is just set to 0.

Closes #1460

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -379,9 +379,9 @@ func (eth *ExchangeWallet) balance() (*asset.Balance, error) {
 		return nil, err
 	}
 
-	locked := eth.locked + eth.redemptionReserve + dexeth.WeiToGwei(bal.PendingOut)
+	locked := eth.locked + eth.redemptionReserve
 	return &asset.Balance{
-		Available: dexeth.WeiToGwei(bal.Current) - locked,
+		Available: dexeth.WeiToGwei(bal.Current) - locked - dexeth.WeiToGwei(bal.PendingOut),
 		Locked:    locked,
 		Immature:  dexeth.WeiToGwei(bal.PendingIn),
 	}, nil
@@ -567,7 +567,8 @@ func (eth *ExchangeWallet) ReturnCoins(coins asset.Coins) error {
 			return fmt.Errorf("unsupported funding coin type for coin %[1]s: %[1]T", coins[i])
 		}
 	}
-	return eth.unlockFunds(amt)
+	eth.unlockFunds(amt)
+	return nil
 }
 
 // FundingCoins gets funding coins for the coin IDs. The coins are locked. This
@@ -612,17 +613,17 @@ func (eth *ExchangeWallet) lockFunds(amt uint64) error {
 }
 
 // unlockFunds unlocks funds held by the wallet.
-func (eth *ExchangeWallet) unlockFunds(amt uint64) error {
+func (eth *ExchangeWallet) unlockFunds(amt uint64) {
 	eth.lockedMtx.Lock()
 	defer eth.lockedMtx.Unlock()
 
 	if eth.locked < amt {
-		return fmt.Errorf("attempting to unlock more than is currently locked. %d > %d",
-			amt, eth.locked)
+		eth.log.Errorf("attempting to unlock more than is currently locked - %d > %d. "+
+			"clearing all locked funds", amt, eth.locked)
+		eth.locked = 0
+	} else {
+		eth.locked -= amt
 	}
-
-	eth.locked -= amt
-	return nil
 }
 
 // swapReceipt implements the asset.Receipt interface for ETH.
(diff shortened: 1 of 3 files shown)


JoeGruffins (commented):
Doesn't seem to change the locked amount for a refund.

Also, I get the error regardless `[ERR] CORE: attempting to unlock more than is reserved for order 349e68455da23ac663b8ba239c37405b160ed49581c254226abde5425abe7995. unlocking reserved amount instead: 0 reserved, unlocking 12600000`

It would be ideal if this did not error at all.

I have been observing refunds by reducing the locktimes in core, btw.

JoeGruffins:
Unrelated but...
While testing I hit this on server a few times:
```
2022-02-07 12:02:30.169 [DBG] SWAP: Allowing until 2022-02-07 03:02:30.005 +0000 UTC (-164.212885ms) to locate contract from maker (NewlyMatched), match cada9cbd5455183b422e2f6f98ec73bfbf6fcd8f19957587dbe4327fc66338f9
2022-02-07 12:02:30.169 [ERR] WAIT: wait.TickerQueue: Waiter given expiration before present
```
Which seems to correspond to a client:
```
2022-02-07 12:03:30.002 [TRC] CORE: notify: |DATA| (epoch) - Index: 109613534
2022-02-07 12:03:30.170 [ERR] CORE: notify: |ERROR| (order) Swap reporting error - Error notifying DEX of swap for match cada9cbd5455183b422e2f6f98ec73bfbf6fcd8f19957587dbe4327fc66338f9: error sending 'init' message: timed out waiting for "init" response - Order: ecfd8843b56258b052266e16069cfc965720ddb20c8d042ba9eeb1ac77374f24
```

chappjc:
> Unrelated but... While testing I hit this on server a few times:

I'll be addressing this with changes to Swapper to deal with "gray revokes".

But this error is the actor submitting their init shortly before the Swapper revokes it in either `checkInactionEventBased` or `checkInactionBlockBased`.  The `TickerQueue` could probably stand to be more courteous in the case of an expire time in the past, by at least calling the TryFunc first and then ExpireFunc if it fails... just letting the client's request timeout in this edge case is not really intended.

martonp:
The issue was actually that the redemption reserves were not being unlocked. When creating a new tracked trade, the `trackedTrade.redemptionLocked` was set to zero. If the match is revoked instead of being redeemed, then the reserves were not being unlocked.

> Also, I get the error regardless `[ERR] CORE: attempting to unlock more than is reserved for order 349e68455da23ac663b8ba239c37405b160ed49581c254226abde5425abe7995. unlocking reserved amount instead: 0 reserved, unlocking 12600000`

The reason for this issue is in #1466. The funds locked for refunds needs to be tracked separately, similarly to how it is done for redeems.

martonp:
Also snuck in a change to make refund use `gasFeeLimit` instead of the fee suggestion.

chappjc:
To clarify, are issues https://github.com/decred/dcrdex/issues/1466 and https://github.com/decred/dcrdex/issues/1460 the same root cause, same issue?
Does this PR resolve https://github.com/decred/dcrdex/issues/1460, which refers to refunds, or does that require tracking funds locked for refunds separately as you commented?

martonp:
What was observed in #1460 (having locked funds when there shouldn't be any) was due to the redemption funds not being unlocked. It wasn't actually due to a refund, it was due the fact that the match was revoked. It is just confusing because when testing on the simnet, both the buyer and seller are using the same wallet. This PR resolves this. 

#1466 is a separate issue, but it is reveled by the error logs added in this PR.

client/asset/eth/eth.go
```diff
@@ -379,9 +379,9 @@ func (eth *ExchangeWallet) balance() (*asset.Balance, error) {
 		return nil, err
 	}
 
-	locked := eth.locked + eth.redemptionReserve + dexeth.WeiToGwei(bal.PendingOut)
+	locked := eth.locked + eth.redemptionReserve
 	return &asset.Balance{
-		Available: dexeth.WeiToGwei(bal.Current) - locked,
+		Available: dexeth.WeiToGwei(bal.Current) - locked - dexeth.WeiToGwei(bal.PendingOut),
```
chappjc, line 384: Looks good as per discussion in https://github.com/decred/dcrdex/issues/1460#issuecomment-1029394466

client/asset/eth/eth.go
```diff
@@ -612,17 +613,17 @@ func (eth *ExchangeWallet) lockFunds(amt uint64) error {
 }
 
 // unlockFunds unlocks funds held by the wallet.
-func (eth *ExchangeWallet) unlockFunds(amt uint64) error {
+func (eth *ExchangeWallet) unlockFunds(amt uint64) {
```
chappjc, line 616: ACK the removal of the error return since it was being ignored everywhere except `ReturnCoins`, and there's really nothing to be done about the error in most code paths.

And if I understand correctly, we will continue to see an error logged until https://github.com/decred/dcrdex/issues/1466 is resolved not unlocking the refund-reserved funds until all swaps are completed.

chappjc (approved):
LGTM.

JoeGruffins (approved):
Locked refunds now being returned!

Rebased this on master and for some reason the max rate was only 100 gwei for initiations so was unable to trade without changes. For some reason I thought all the fee stuff was sorted.

martonp:
@JoeGruffins Either way that shouldn't be happening with this change.. I updated `Refund` to use `eth.gasFeeLimit` instead of the server's suggestion.

JoeGruffins:
The initiation was too low. I think was just an old server sending a bad fee.

PR #1471, opened 2022-02-08
client/core: reload wallet on connect fail

chappjc:
DRAFT  A starting point for a different solution to https://github.com/decred/dcrdex/pull/1413.  This was the inital solution, which is still in need of concurrency controls for the `xcWallet.Wallet` field.


chappjc:
Even this is going to lose any state that's kept in the asset.Wallet (e.g. ExchangeWallet), so I don't think this is going to fly either.

Maybe the resolution is to just address the issue with DCR only by replacing the `rpcclient.Client`, which will be messy, but possibly necessary.

PR #1472, opened 2022-02-08
client/asset/dcr: ensure rpcclient is fresh before Connect

chappjc:
Another alternative to https://github.com/decred/dcrdex/pull/1471, and the polar opposite approach to https://github.com/decred/dcrdex/pull/1413.  Unlike those, this just addresses the real issue, which is that dcrd's `rpcclient.Client` becomes unusable after `Shutdown`.  I had drafted this commit almost a month ago, but rejected it because I disliked the extra mutex, but given the state in an `xcWallet` or an `asset.Wallet` implementation like the `ExchangeWallet`s such as locked amounts, it's probably best just to solve it at the level of the `rpcWallet`.  If other assets have a similar constraint, they will have to adopt a similar solution.  e.g. BTC's wallet client does HTTP POST, and there's no issue there, but maybe eth would need special handling.

This allows the rpcclient to be reconstructed as needed before Connect, as is the case if it had been Disconnected previously. This is needed because dcrd's rpcclient.Client is not able to start up again after having shutdown.

### DIFF
--- a/client/asset/dcr/rpcwallet.go
+++ b/client/asset/dcr/rpcwallet.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"sync"
 	"sync/atomic"
 
 	"decred.org/dcrdex/client/asset"
@@ -46,16 +47,18 @@ const (
 type rpcWallet struct {
 	chainParams *chaincfg.Params
 	log         dex.Logger
-	spvMode     bool
+	rpcCfg      *rpcclient.ConnConfig
 
+	rpcMtx  sync.RWMutex
+	spvMode bool
 	// rpcConnector is a rpcclient.Client, does not need to be
 	// set for testing.
-	rpcConnector
+	rpcConnector rpcConnector
 	// rpcClient is a combined rpcclient.Client+dcrwallet.Client,
 	// or a stub for testing.
-	rpcClient
+	rpcClient rpcClient
 
-	connectCount uint32
+	connectCount uint32 // atomic
 }
 
 // Ensure rpcWallet satisfies the Wallet interface.
@@ -66,6 +69,13 @@ type walletClient = dcrwallet.Client
 type combinedClient struct {
 	*rpcclient.Client
 	*walletClient
+}
+
+func newCombinedClient(nodeRPCClient *rpcclient.Client, chainParams *chaincfg.Params) *combinedClient {
+	return &combinedClient{
+		nodeRPCClient,
+		dcrwallet.NewClient(dcrwallet.RawRequestCaller(nodeRPCClient), chainParams),
+	}
 }
 
 // Ensure combinedClient satisfies the rpcClient interface.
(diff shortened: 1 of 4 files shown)


client/asset/dcr/rpcwallet.go
```diff
@@ -580,6 +656,8 @@ func (w *rpcWallet) SyncStatus(ctx context.Context) (bool, float32, error) {
 // AddressPrivKey fetches the privkey for the specified address.
 // Part of the Wallet interface.
 func (w *rpcWallet) AddressPrivKey(ctx context.Context, address stdaddr.Address) (*dcrutil.WIF, error) {
+	w.rpcMtx.RLock()
+	defer w.rpcMtx.RUnlock()
```
chappjc, line 660: Maybe this mutex is just overly cautious.  It's not clear if there would ever be any of these normal method calls concurrent with a Connect or Disconnect call, so maybe this is solvable with documentation.

dex/runner.go
```diff
@@ -105,6 +105,9 @@ func (c *ConnectionMaster) Connect(ctx context.Context) (err error) {
 	c.mtx.Lock()
 	c.wg, err = c.connector.Connect(c.ctx)
 	c.mtx.Unlock()
+	if err != nil {
+		c.cancel() // otherwise On() says true
+	}
```
chappjc, line 110: Note sure how we didn't have trouble with this before.  This allows a wallet whose `ConnectionMaster` has returned a non-nil error from `Connect` to look like we expect:

image

Otherwise you just see the syncing circle-arrow icon and no "Connect" button

client/asset/dcr/rpcwallet.go
```diff
@@ -138,44 +143,46 @@ func newRPCWallet(cfg *Config, chainParams *chaincfg.Params, logger dex.Logger)
...
-	err := rpcw.setupRPCClient(cfg.RPCListen, cfg.RPCUser, cfg.RPCPass, cfg.RPCCert)
+	rpcw.rpcCfg = &rpcclient.ConnConfig{
+		Host:                cfg.RPCListen,
+		Endpoint:            "ws",
+		User:                cfg.RPCUser,
+		Pass:                cfg.RPCPass,
+		Certificates:        certs,
+		DisableConnectOnNew: true,
+	}
+	err = rpcw.setupRPCClient() // this rpcWallet instance is still locally scoped, no need to lock rpcMtx
```
buck54321, line 161: Because we have `DisableConnectOnNew` set, there's probably no way we're going to error in `setupRPCClient`, so doing it in the constructor is probably pointless in terms of checking configuration. Does it simplify things to defer all calls of `setupRPCClient` to `Connect`?

client/asset/dcr/rpcwallet.go
```diff
@@ -245,8 +255,20 @@ func (w *rpcWallet) checkRPCConnection(ctx context.Context) error {
 // Connect establishes a connection to the previously created rpc client.
 // Part of the Wallet interface.
 func (w *rpcWallet) Connect(ctx context.Context) error {
+	w.rpcMtx.Lock()
+	defer w.rpcMtx.Unlock()
+
```
buck54321, line 260: After successful `Connect`, is `rpcReadyToConnect` still `true`? So a subsequent call to `Connect` without a call to `Disconnect` would make it to `rpcConnector.Connect` for a second time on the same `rpcConnector`. In theory, the caller would never call `Connect` again before calling `Disconnect`. We can work under that assumption. I just wanted to make sure I'm understanding the intent.

client/asset/dcr/rpcwallet.go
```diff
@@ -188,14 +195,17 @@ func (w *rpcWallet) handleRPCClientReconnection() {
 	}
 
 	w.log.Debugf("dcrwallet reconnected (%d)", connectCount-1)
-	err := w.checkRPCConnection(context.TODO())
+	w.rpcMtx.RLock()
+	defer w.rpcMtx.RUnlock()
+	err := w.checkRPCConnection(context.Background())
```
buck54321, line 200: Wanna store the `ctx` in `(*rpcWallet).Connect`?

client/asset/dcr/rpcwallet.go
chappjc: My thinking here is that I'd rather have a second (incorrect) call to `Connect` with no prior `Disconnect` to be met with an "already connected" error than letting the rpcclient get recreated.  But yeah, never call `Connect` again before calling `Disconnect` is the idea.

client/asset/dcr/rpcwallet.go
```diff
@@ -188,14 +195,17 @@ func (w *rpcWallet) handleRPCClientReconnection() {
...
 }
 
-// isSpvMode uses the walletinfo rpc to determine if this wallet is
-// connected to the Decred network via SPV peers.
+// isSpvMode uses the walletinfo rpc to determine if this wallet is connected to
+// the Decred network via SPV peers. This method is not safe for concurrent
+// access, and the rpcMtx must be at least read locked.
 func (w *rpcWallet) checkRPCConnection(ctx context.Context) error {
 	// Reset spvMode to false, until we're sure we're connected to
 	// an SPV wallet below.
```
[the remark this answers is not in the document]
chappjc: Good catch.  The mutability of the `spvMode` field appears to have been neglected until now.

chappjc:
I'll have to test that last commit thoroughly again.  It's a big design change.
Got rid of the `rpcReadyToConnect` boolean, always creating an RPC client in `Connect`, passing the context from the Connect call where it applies, guarding `spvMode`, etc.
Only awkward thing is that I really didn't want to leave `rpcConnector` and `rpcClient` nil after construction so in the constructor I validate the config with `rpcclient.New` and store the results in those fields just to make any premature ordinary wallet method calls return a sensible "not connected" error instead of nil dereference a panic.

chappjc:
Still fixing some stuff up

chappjc:
I feel alright about this now.  Working in some interactive simnet tests with random errors in different parts of the connect code.

buck54321 (approved):

PR #1445, opened 2022-01-24
client: monitor wallet peers

chappjc:
This adds a `PeersChange` callback to `client/asset.WalletConfig`, similar to `TipChange`, for backends to report changes to the network peer count. The backends now implement a `monitorPeers` goroutine similar to `monitorBlocks` but checking and reporting on peer count changes.  Any changes in peer count result in a wallet state note.  Change in peer count from 0 to >0 also start a sync status monitoring goroutine, which runs until sync is restored.

This also updates each backend's `SyncStatus` method to ensure that the `synced bool` will never be true if there are peer count is 0.

Killing the "beta" node with the **SPV** beta wallet:

image

image

Starting up beta dcrd makes the signal-! icon change to a syncing arrow circle icon, which then goes away when it catches up with the network.

Killing "alpha" bitcoind with native btcwallet:

image

image

That also returns to normal after restarting alpha bitcoind.

Now the **full-node** wallets (still simnet):

https://user-images.githubusercontent.com/9373513/152836640-c8977ce5-bbf5-417c-add1-336326696b1b.mp4

```
2022-02-07 11:04:07.769 [ERR] CORE[dcr]: Failed to get peer count: rawrequest error: -9: wallet.NetworkBackend: Decred network is unreachable
2022-02-07 11:04:07.769 [WRN] CORE: Wallet for asset dcr has zero network peers!
2022-02-07 11:04:07.771 [WRN] CORE: notify: |WARNING| (walletconfig) Wallet network issue - Decred wallet has no network peers!
2022-02-07 11:04:07.771 [TRC] CORE: notify: |DATA| (walletstate)
2022-02-07 11:04:16.004 [TRC] CORE: notify: |DATA| (epoch) - Index: 109616897
2022-02-07 11:04:18.586 [ERR] CORE: btc wallet is reporting a failed state: failed to get best block hash from btc node
...
2022-02-07 11:04:22.586 [ERR] CORE[btc]: Failed to get peer count: rawrequest error: Post "http://127.0.0.1:20556/wallet/": dial tcp 127.0.0.1:20556: connect: connection refused
2022-02-07 11:04:22.586 [WRN] CORE: Wallet for asset btc has zero network peers!
2022-02-07 11:04:22.587 [ERR] CORE: btc wallet is reporting a failed state: failed to get best block hash from btc node
2022-02-07 11:04:22.589 [WRN] CORE: notify: |WARNING| (walletconfig) Wallet network issue - Bitcoin wallet has no network peers!
[...]

PR #759, opened 2020-10-17
server/auth: prevent multiple fee coin waiters for one account

chappjc:
This tracks actively running fee coin waiters by account ID.  An account is only permitted to start one waiter at a time.

Also fixes two leaks: a ticker leak in `dex/ws.WsLink`, and a context leak in `dex.ConnectionMaster`.  BTW, the `RWMutex` locking in the `ConnectionMaster` methods seems unnecessary if `Disconnect` and `Wait` are never called before `Connect`.

PR #1474, opened 2022-02-10
dex: redesign ConnectionMaster

chappjc:
See the discussion at https://github.com/decred/dcrdex/pull/1445#discussion_r803100088 for the motivation.

This redesigns `dex.ConnectionMaster`, adding a `Done() <-chan struct{}` method upon which the `On` and `Wait` methods are now based.

This also uses the new `Done` method in `(*Core).startWalletSyncMonitor` to quit if the wallet shuts down without an extra goroutine sitting on `Wait`.

### DIFF
--- a/dex/runner.go
+++ b/dex/runner.go
@@ -5,7 +5,10 @@ package dex
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"sync"
+	"sync/atomic"
 )
 
 // contextManager is used to manage a context and its cancellation function.
@@ -86,12 +89,15 @@ type Connector interface {
 
 // ConnectionMaster manages a Connector.
 type ConnectionMaster struct {
-	contextManager
-	wg        *sync.WaitGroup
 	connector Connector
+	cancel    context.CancelFunc
+	done      atomic.Value // chan struct{}
 }
 
-// NewConnectionMaster is the constructor for a new ConnectionMaster.
+// NewConnectionMaster creates a new ConnectionMaster. The Connect method should
+// be used before Disconnect. The On, Done, and Wait methods may be used at any
+// time. However, prior to Connect, Wait and Done immediately return and signal
+// completion, respectively.
 func NewConnectionMaster(c Connector) *ConnectionMaster {
 	return &ConnectionMaster{
 		connector: c,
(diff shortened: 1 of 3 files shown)


dex/runner.go
```diff
@@ -86,38 +87,94 @@ type Connector interface {
...
+	<-c.Done() // let the anon goroutine from Connect return
+}
+
+// Disconnect closes the connection and waits for shutdown. This must not be
+// used before or concurrently with Connect.
+func (c *ConnectionMaster) Disconnect() {
+	if !c.On() {
+		return // probably a bug in the consumer
+	}
+	c.cancel()
```
chappjc, line 178: See the old `Disconnect` method where the `cancel` field has always been ungaurded by the mutex.
Evidently we do have sane Connect/Disconnect access patterns, and the `On` method is the only one that really needs concurrency controls, which it gets via `Done`.

dex/runner.go
```diff
@@ -86,38 +87,94 @@ type Connector interface {
...
+	}
+	return c.done
+}
+
+// On indicates if the Connector is running. This returns false if never
+// connected, or if the Connector has completed shut down.
+func (c *ConnectionMaster) On() bool {
+	select {
+	case <-c.Done():
+		return false
```
chappjc, line 160: Although `Wait` has the same semantics as before, the `On` method now correlates to the `Wait` method's timing instead of just when the inner context is canceled.  This new way is actually a bit more consistent since depending on how the Connector died (inner context canceled vs. parent context canceled vs. internally-initiated shutdown) `On` would previously indicate slightly different stages of the `Connector`'s lifecycle (signaled to shut down vs actually shut down).

dex/runner.go
```diff
@@ -86,41 +87,94 @@ type Connector interface {
...
 // ConnectionMaster manages a Connector.
 type ConnectionMaster struct {
-	contextManager
-	wg        *sync.WaitGroup
 	connector Connector
+	cancel    context.CancelFunc
+
+	// To permit calls to Done/On before Connect, guard the done channel.
+	mtx  sync.RWMutex
+	done chan struct{}
```
buck54321, line 95: Could use an `atomic.Value` here too.

dex/runner.go
```diff
@@ -86,41 +87,94 @@ type Connector interface {
...
+	done := make(chan struct{})
+	c.mtx.Lock()
+	c.done = done // after setting cancel
+	c.mtx.Unlock()
+
+	go func() { // capture the local variables
+		wg.Wait()
+		cancel() // if the Connector just died on its own, don't leak the context
+		close(done)
+	}()
```
buck54321, line 137: No way to avoid the goroutine? Probably doesn't matter though.

Ah. The goroutine gets rid of the `wg sync.WaitGroup` field.

dex/runner.go
buck54321: > This must not be used before or concurrently with Connect.

I think this is an assumption we are under at various levels. Good to doc it, though.

dex/runner.go
```diff
@@ -86,41 +87,94 @@ type Connector interface {
...
+	if c.On() { // probably a bug in the consumer
+		return errors.New("already running")
+	}
+
+	// Attempt to start the Connector.
+	ctx, cancel := context.WithCancel(ctx)
+	wg, err := c.connector.Connect(ctx)
 	if err != nil {
-		c.cancel() // otherwise On() says true
+		cancel() // no context leak
```
chappjc, line 120: Turns out we cannot do this because at least for `client/comms.(*wsConn).Connect` it breaks the reconnect loop.

We'll just have to search for a different solution to the wallet Connectors not showing disconnected https://github.com/decred/dcrdex/pull/1472/commits/e8fd8acab23fbb253b649a66a95a5a63d4be6b75

chappjc:
Out of draft now since we keep running into issues and I'd kinda like to eliminate the fields that can be nil.  https://github.com/decred/dcrdex/pull/1602#discussion_r876074580, https://github.com/decred/dcrdex/pull/1616, https://github.com/decred/dcrdex/pull/1472#discussion_r802149239, https://github.com/decred/dcrdex/pull/759/commits/6aaa3c01391b0785ad5941c3b2e1ea26e7f95127, etc.

However, I suspect we still need to rethink the behavior of the `dex.Connector`:

```go
Connect(ctx context.Context) (*sync.WaitGroup, error)
```

Seems straightforward, but to facilitate the reconnect loop starting for DEX conns that fail their initial attempt, we have the case where `wg != nil` while also `err != nil`.  I'm not sure how best to fix or rework that.

dex/runner.go
```diff
@@ -107,13 +112,31 @@ func NewConnectionMaster(c Connector) *ConnectionMaster {
...
+		return errors.New("already running")
+	}
+
+	// Attempt to start the Connector.
+	ctx, cancel := context.WithCancel(ctx)
+	wg, err := c.connector.Connect(ctx)
+	if wg == nil {
+		cancel() // no context leak
+		return err
+	}
```
JoeGruffins, line 125: Should this have a custom error? Should just be checking `err != nil`? Does it need two checks?

JoeGruffins (approved):

dex/runner.go
chappjc: I'm thinking the Connector is buggy if both wg and err are nil.

JoeGruffins: To catch the bug a custom error would be better? i.e. "connection master connect error: %w, err"

martonp (commented):
`Connect(ctx context.Context) (*sync.WaitGroup, error)` could be updated to `Connect(ctx context.Context) (*sync.WaitGroup, bool, error)` where if the boolean is true, it means that the the initial connection failed, but the reconnect loop is running.

buck54321:
Kinda seems like we need different behavior depending on whether we are trying the connection for the first time (`(*Core).Register`) or we're restarting a connection that has already worked before (`(*Core).initialize`).

chappjc:
`Connect` vs. `ConnectOnce` maybe.
My feeling is that's a different PR as it touches all dex.Connectors like wallets not just server conns.  Main goal here was a `Done` method with a channel to wait on, and I just wanted to raise the design issue we're hitting with the waitgroup return.

BTW I've had thoughts about changing the `*WaitGroup` to something else more generic with a Wait method or even a `context.Context` with both an `Err` and `Done` method.

None of this seems high priority though compared to the mess with LTC

buck54321 (approved):

PR #1475, opened 2022-02-11
multi: Delete old records from client db.

### LINKED ISSUE #1448: core,db: need methods to clean/delete old orders and matches, purging them from the db
Perhaps optionally based on age.
Will likely need to be deleting keys/buckets in batches other wise the bolt txn will get too big and oom dex.

JoeGruffins:
client/rpcserver: Add deletearchivedrecords.
    
    Add a rpc enpoint that allows deleting archived order and match records
    with an optional older than time. Also allow specifying file names to
    save records in csv form.

    client: Add ability to delete archived records.
    
    Add a method that deletes archived matches and orders while optionally
    saving them to specified files in csv form.

    client/db: Add methods to delete archived data.
    
    Add two methods to delete archived orders and matches. They are designed
    to be run in batches and return all deleted material back to the caller.

    client: Shuffle some functions around.
    
    Move some useful functions from cmd and webserver to a new helper file
    in core so that they can be used in other places. The moved functions
    are used to parse file paths and orders and are helpful for creating
    comma separated values.

closes #1448


client/cmd/dexc/config.go
```diff
@@ -141,11 +142,17 @@ func configure() (*Config, error) {
...
+	// If the app directory has been changed, replace shortcut chars such
+	// as "~" with the full path.
+	if preCfg.AppData != defaultApplicationDirectory {
+		preCfg.AppData = core.CleanAndExpandPath(preCfg.AppData)
+		// If the app directory has been changed, but the config file path hasn't,
+		// reform the config file path with the new directory.
+		if preCfg.Config == defaultConfigPath {
+			preCfg.Config = filepath.Join(preCfg.AppData, configFilename)
+		}
 	}
```
JoeGruffins, line 154: I apologize for this unrelated change. I often attempt to use home or `~` to with dexc and alway end up creating a floder named `~` in the current directory which is not intended. Hopefully this is ok to squeeze in here.

client/core/core.go
```diff
@@ -6557,3 +6559,203 @@ func (c *Core) WalletLogFilePath(assetID uint32) (string, error) {
...
+		}
+		var ords []*db.MetaOrder
+		ords, finished, err = c.db.DeleteInactiveOrders(olderThan)
+		if err != nil {
+			return fmt.Errorf("unable to delete orders: %v", err)
+		}
+
+		if ordersFile != nil {
+			for _, ord := range ords {
+				cord := coreOrderFromTrade(ord.Order, ord.MetaData)
```
JoeGruffins, line 6620: Should this ignore cancel orders? I'm thinking it should because it is not very useful information. With these changes they look like
```
,77d91eda537bd722a64fa4d9a45c518a4dc77e3034518b3a5541a39a75c8239e,,,0,0,0,0,0,unknown,buy,immediate,unknown,NaN,NaN,1970-01-01T09:00:00+09:00
```
Which is not very helpful.

client/core/core.go
```diff
@@ -6557,3 +6559,203 @@ func (c *Core) WalletLogFilePath(assetID uint32) (string, error) {
 
 	return wallet.LogFilePath()
 }
+
+func (c *Core) DeleteArchivedRecords(olderThan *time.Time, matchesFile, ordersFile *os.File) error {
```
chappjc, line 6563: Please doc.

client/core/core.go
```diff
@@ -6557,3 +6559,203 @@ func (c *Core) WalletLogFilePath(assetID uint32) (string, error) {
...
+	c.wg.Add(1)
+	c.wg.Done()
```
chappjc, line 6565: wg copy pasta?

client/core/core.go
```diff
@@ -6557,3 +6559,203 @@ func (c *Core) WalletLogFilePath(assetID uint32) (string, error) {
...
+	var (
+		finished  bool
+		err       error
+		csvWriter *csv.Writer
+	)
+
+	// If provided a file to write the orders csv to, write the header and
+	// defer closing the file.
+	if ordersFile != nil {
+		defer ordersFile.Close()
```
chappjc, line 6575: I tend to think that the file should be closed by whatever opened it e.g. rpcserver, not here.

But I think `DeleteArchivedRecords` should take `io.Writer`s, not `*os.File`.  The caller could conceivably be the webserver and it would be desirable to write to an `http.ResponseWriter`...
although how that I say it, it's problematic that there are two different outputs, one for matches and one for orders.  Hmm.  Merging them would make for a nasty csv scheme.

client/db/bolt/db.go
```diff
@@ -1624,6 +1624,129 @@ func (idx *timeIndexNewest) add(t uint64, k []byte, b *bbolt.Bucket) {
 	})
 }
 
+// DeleteInactiveOrders deletes orders that no longer needed for normal
+// operations. Optionally accepts a time to delete orders with a later time
+// stamp.
+func (db *BoltDB) DeleteInactiveOrders(olderThan *time.Time) (deletedOrders []*dexdb.MetaOrder, finished bool, err error) {
```
chappjc, line 1630: While it is logical to have the delete and backup operations be facilitated by the same db method by returning a slice of deleted orders, I'm fairly certain that is going to be a ton of memory in some notable use cases.

A trick to consider is to pass a function to `DeleteInactiveOrders` like

```go
func (db *BoltDB) DeleteInactiveOrders(olderThan *time.Time, removed func(*dexdb.MetaOrder) error) (bool, error)
```

A caller like `(*Core).DeleteArchivedRecords` would then handle them in a stream.  A nil `removed` function could be an option if the caller does not want them.

I had to do something similar with `MarketMatchesStreaming` server-side.  See `server/admin.(*Server).apiMarketMatches` where it streams `MatchData` into JSON.

client/core/core.go
```diff
@@ -6557,3 +6559,203 @@ func (c *Core) WalletLogFilePath(assetID uint32) (string, error) {
...
+
+	// Delete matches while saving to csv if available until the database
+	// says that's all or context is canceled.
+	finished = false
+	for !finished {
+		if err := c.ctx.Err(); err != nil {
+			return fmt.Errorf("unable to delete matches: %v", err)
+		}
+		var mtchs []*db.MetaMatch
+		mtchs, finished, err = c.db.DeleteInactiveMatches(olderThan)
```
chappjc, line 6719: In case there's an error somewhere, here or in the orders loop above, I wonder if the DB would be in a better state if we deleted matches before deleting orders.  Maybe it doesn't matter.  Thoughts?

client/core/core.go
chappjc: Since the targeted order would have status cancelled, I tend to agree there's little value in recording the cancel order too.  Only thing I suppose we'd lose is the time of cancellation, right?

BTW, why in that example entry is the timestamp the epoch 0 value?

chappjc (changes requested):
Some initial feedback in case I don't do a complete review today.

Looks like you accidentally checked in an empty "client/rpcserver/123" file.

client/db/bolt/db.go
```diff
@@ -1624,6 +1624,129 @@ func (idx *timeIndexNewest) add(t uint64, k []byte, b *bbolt.Bucket) {
...
+		} else {
+			finished = true
+		}
+		for _, trio := range trios {
+			o, err := decodeOrderBucket(trio.k, trio.b)
+			if err != nil {
+				return fmt.Errorf("failed to decode order bucket: %v", err)
+			}
+			deletedOrders = append(deletedOrders, o)
+			if err := archivedOB.DeleteBucket(trio.k); err != nil {
```
chappjc, line 1695: Large bolt transactions always have me concerned about memory.  One would think that deleting buckets wouldn't balloon memory, but... wish we could be more certain about this first.

client/db/bolt/db.go
```diff
@@ -1624,6 +1624,129 @@ func (idx *timeIndexNewest) add(t uint64, k []byte, b *bbolt.Bucket) {
...
+	deletedOrders = make([]*dexdb.MetaOrder, 0, batchSize)
+	return deletedOrders, finished, db.Update(func(tx *bbolt.Tx) error {
+		// Some archived orders may still be needed for active matches.
+		// Put those order id's in a map to prevent deletion.
+		amb := tx.Bucket(activeMatchesBucket)
+		activeMatches := make(map[order.OrderID]struct{})
+		amb.ForEach(func(k, _ []byte) error {
+			mBkt := amb.Bucket(k)
+			if mBkt == nil {
+				return fmt.Errorf("match %x bucket is not a bucket", k)
```
chappjc, line 1648: Looks like this error is ignored. (`ForEach` error ignored)

client/core/core.go
JoeGruffins: should be defer

client/core/core.go
JoeGruffins: Can delete matches first.

client/db/bolt/db.go
JoeGruffins: Will try it with the function pass.

client/core/core.go
JoeGruffins: Can just pass the file name and not verify it in the rpcserver.

client/core/helpers.go
```diff
@@ -0,0 +1,570 @@
...
+
+	"decred.org/dcrdex/dex"
+	"decred.org/dcrdex/dex/calc"
+	"decred.org/dcrdex/dex/encode"
+	"decred.org/dcrdex/dex/order"
+)
+
+// CleanAndExpandPath expands environment variables and leading ~ in the
+// passed path, cleans the result, and returns it.
+func CleanAndExpandPath(path string) string {
```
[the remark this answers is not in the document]
JoeGruffins: Will move.

client/core/core.go
chappjc: I don't see the goroutine though

JoeGruffins: It's possible that the caller run this in a goroutine?

JoeGruffins: I can just comment and take out.

Looking at the current rpcserver code, I think it depends on if go-chi will wait for the handler to finish on its server shutdown method.

client/core/core.go
JoeGruffins: I beleive `coreOrderFromTrade` attempts to return nil for everything if it's a cancel order.

client/core/helpers.go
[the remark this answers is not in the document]
JoeGruffins: Moved.

client/db/bolt/db.go
JoeGruffins: Passing a function.

client/db/bolt/db.go
JoeGruffins: It should be only deleting the batch size per transaction. That can be adjusted.

client/db/bolt/db.go
JoeGruffins: Returning now if error.

client/core/core.go
JoeGruffins: Have not changed this yet.

client/core/core.go
JoeGruffins: Just passing the file name now. Core handles opening the file and clean up.

client/core/core.go
JoeGruffins: Deleting matches first.

client/db/bolt/db.go
```diff
@@ -1624,6 +1624,167 @@ func (idx *timeIndexNewest) add(t uint64, k []byte, b *bbolt.Bucket) {
...
+				}
+			}
+			db.log.Infof("Deleted %d orders from the database in %v.", len(trios), time.Since(start))
+			return nil
+		})
+		if err != nil {
+			return fmt.Errorf("unable to delete orders: %v", err)
+		}
+		if perBatchFn != nil {
+			if err := perBatchFn(deletedOrders); err != nil {
```
chappjc, line 1715: I had envisioned calling a `func(ords *db.MetaOrder) error` for each order _within_ the transaction (one at a time, no slice) so that it can be rolled back if processing the removed order(s) failed, and they would be more continuously streaming.  Down here it's too late to get those orders back in the DB without manually reinserting them, and `perBatchFn` failed to process them so they didn't make it to the consumer.

Do you think it would be problematic or less performant to have the function take a single order and to stream them continuously from within the db transaction that loads and removes them?  I haven't considered this as much as you so I could be mistaken.

client/db/bolt/db.go
```diff
@@ -1624,6 +1624,167 @@ func (idx *timeIndexNewest) add(t uint64, k []byte, b *bbolt.Bucket) {
...
+	for !finished {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		deletedOrders := make([]*dexdb.MetaOrder, 0, batchSize)
+		err := db.Update(func(tx *bbolt.Tx) error {
+			// Some archived orders may still be needed for active matches.
+			// Put those order id's in a map to prevent deletion.
+			amb := tx.Bucket(activeMatchesBucket)
+			activeMatches := make(map[order.OrderID]struct{})
```
chappjc, line 1651: This `activeMatches` map will generated identically on each call to `db.Update`.  Should there be a preceding `db.View`, outside of the `finished` loop, to create this map once?

client/rpcserver/handlers.go
```diff
@@ -745,6 +762,21 @@ var helpMsgs = map[string]helpMsg{
...
 		returns: `Returns:
     string: The message "` + initializedStr + `"`,
+	},
+	deleteArchivedRecordsRoute: {
+		argsShort: `("unix time milli") ("matches csv path") ("orders csv path")`,
+		cmdSummary: `Delete archived matches from the database. Optionally set a time to
+    delete records after and file paths to save deleted records as comma separated
+    values. Note that file locations are from the perspective of dexc and not the caller.`,
+		argsLong: `Args:
+    unix time milli (int): Optional. If set deletes records after the date in unix time
```
chappjc, line 772: after ->  before, right?  As in older than this time.

chappjc (changes requested):

client/cmd/dexc/config.go
```diff
@@ -15,6 +15,7 @@ import (
 	"decred.org/dcrdex/client/asset"
 	"decred.org/dcrdex/client/cmd/dexc/version"
 	"decred.org/dcrdex/dex"
+	"decred.org/dcrdex/dex/encode"
```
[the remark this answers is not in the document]
JoeGruffins: Moved.

client/cmd/dexcctl/config.go
```diff
@@ -142,57 +141,3 @@ func configure() (*config, []string, bool, error) {
 
 	return cfg, remainingArgs, false, nil
 }
-
-// cleanAndExpandPath expands environment variables and leading ~ in the
-// passed path, cleans the result, and returns it.
-func cleanAndExpandPath(path string) string {
```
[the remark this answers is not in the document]
JoeGruffins: I should have looked for more. All using the same one now.

client/db/bolt/db.go
```diff
@@ -1624,6 +1624,167 @@ func (idx *timeIndexNewest) add(t uint64, k []byte, b *bbolt.Bucket) {
...
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		deletedMatches := make([]*dexdb.MetaMatch, 0, batchSize)
+		if err := db.Update(func(tx *bbolt.Tx) error {
+			archivedMB := tx.Bucket(archivedMatchesBucket)
+			if archivedMB == nil {
+				return fmt.Errorf("failed to open %s bucket", string(archivedMatchesBucket))
+			}
+			// Retrieve one more than we will delete in order to guage if
```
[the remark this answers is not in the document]
JoeGruffins: Fixed.

client/db/bolt/db.go
JoeGruffins: imo the performance isn't such a big issue because this will not be used regularly, or by everyone. Will make this change so that we can roll back the tx if the passed function fails.

If we still do it inside the batch, however, we will return some orders a deleted that are later un-deleted with the roll back though. Will log this in case.

client/db/bolt/db.go
```diff
@@ -1624,6 +1624,167 @@ func (idx *timeIndexNewest) add(t uint64, k []byte, b *bbolt.Bucket) {
...
+		if perBatchFn != nil {
+			if err := perBatchFn(deletedOrders); err != nil {
+				return fmt.Errorf("problem performing batch function: %v", err)
+			}
+		}
+	}
+	return nil
+}
+
+// DeleteInactiveMatches deletes matches that no longer needed for normal
```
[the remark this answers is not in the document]
JoeGruffins: Fixed.

client/core/core.go
```diff
@@ -6621,3 +6623,234 @@ func (c *Core) WalletLogFilePath(assetID uint32) (string, error) {
...
+		return nil
+	}, matchesFile.Close, nil
+}
+
+// DeleteArchivedRecords deletes archived matches from the database. Optionally
+// set a time to delete records after and file paths to save deleted records as
+// comma separated values.
+func (c *Core) DeleteArchivedRecords(olderThan *time.Time, matchesFile, ordersFile string) error {
+	c.wg.Add(1)
+	defer c.wg.Done()
```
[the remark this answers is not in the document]
JoeGruffins: I see. wg.Add(1) could happen after all the other wg.Done() were called, and that would lead to unintended consequences.

Will remove.

client/db/bolt/db.go
chappjc: Those are good points.  Only make the changes if they feel right to you.

JoeGruffins: Doing per tx and logging the number of things that are rolled back if there is a problem.

client/db/bolt/db.go
JoeGruffins: Creating outside now.

client/core/core.go
```diff
@@ -6621,3 +6623,234 @@ func (c *Core) WalletLogFilePath(assetID uint32) (string, error) {
 
 	return wallet.LogFilePath()
 }
+
+func createFile(fileName string) (*os.File, error) {
+	if fileName == "" {
+		return nil, nil
```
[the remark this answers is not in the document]
JoeGruffins: Returning error.

client/core/core.go
```diff
@@ -6621,3 +6623,234 @@ func (c *Core) WalletLogFilePath(assetID uint32) (string, error) {
...
+	}
+	var f *os.File
```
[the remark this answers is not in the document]
JoeGruffins: Removed.

client/core/core.go
```diff
@@ -6621,3 +6623,234 @@ func (c *Core) WalletLogFilePath(assetID uint32) (string, error) {
...
+		err             error
+		mtchsPerBatchFn func([]*db.MetaMatch) error
+	)
+	// If provided a file to write the orders csv to, write the header and
+	// defer closing the file.
+	if matchesFile != "" {
+		var cleanup func() error
+		mtchsPerBatchFn, cleanup, err = deleteBatchMatchesFn(matchesFile)
+		if err != nil {
+			return fmt.Errorf("undable to set up orders csv: %v", err)
```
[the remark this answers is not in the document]
JoeGruffins: Fixed.

client/core/core.go
```diff
@@ -6621,3 +6623,234 @@ func (c *Core) WalletLogFilePath(assetID uint32) (string, error) {
...
+				return fmt.Errorf("error writing matches CSV: %v", err)
+			}
+		}
+		return nil
+	}, matchesFile.Close, nil
+}
+
+// DeleteArchivedRecords deletes archived matches from the database. Optionally
+// set a time to delete records after and file paths to save deleted records as
+// comma separated values.
```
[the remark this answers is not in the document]
JoeGruffins: Added.

client/core/core.go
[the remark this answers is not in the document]
JoeGruffins: Removed.

client/core/core.go
JoeGruffins: Removed.

client/rpcserver/handlers.go
JoeGruffins: Yes before....

JoeGruffins:
I apologize for rebasing without the incremental changes as separate comments.  It's much easier this way when updating multiple commits. What do you think? Dave has said there is a way on github to view the changes when rebased like this, but I'm not sure where it is.

client/core/core.go
JoeGruffins: I guess I will change `coreOrderFromTrade` to add more data for cancels. Will just tag on another commit...

JoeGruffins: Cancels now look like
```
Host,Order ID,Base,Quote,Base Quantity,Order Rate,Actual Rate,Base Fees,Quote Fees,Type,Side,Time in Force,Status,TargetOrderID,Filled (%),Settled (%),Time
127.0.0.1:17273,54d39189eb33224c5caf50ac759468b240a5a2d247bf433751f3ca87f069cf1a,dcr,eth,10,0.90865,0,0,0,limit,buy,standing,canceled,,0.0,0.0,2022-02-23T16:09:20.916+09:00
127.0.0.1:17273,26916d0a8244c1962b3f823e64c4db52f877de050da26fc1a702999bd535aa63,dcr,eth,0,0,0,0,0,cancel,buy,immediate,no match,54d39189eb33224c5caf50ac759468b240a5a2d247bf433751f3ca87f069cf1a,NaN,NaN,2022-02-23T16:09:31.881+09:00
127.0.0.1:17273,6acf10ae8c7c00411bed447fb51537a911a4cc61999b229a156170ed2d796bab,dcr,eth,10,0.90865,0,0,0,limit,sell,standing,canceled,,0.0,0.0,2022-02-23T16:09:06.299+09:00
127.0.0.1:17273,fe8f94a6c03bc6bf304f023e680f286a2a6dc6f138d0c3ff25d0f4cb3924d6d2,dcr,eth,0,0,0,0,0,cancel,buy,immediate,no match,6acf10ae8c7c00411bed447fb51537a911a4cc61999b229a156170ed2d796bab,NaN,NaN,2022-02-23T16:09:08.914+09:00
```

Added to the third commit.

JoeGruffins: I see the side's default is `buy` so wrong for cancels.

JoeGruffins: Now returning no string for cancel side. Also not attempting to parse filled and settled.

```
127.0.0.1:17273,4d4d35bd86e6c65f7616201b3274b357fcd048ec2c0ce9e9323fa65efc83c387,dcr,eth,0,0,0,0,0,cancel,,immediate,executed,ff99c97afe675dac465dda62b12d93af1198551fd0b8196c4c8ce950ed1bc267,,,2022-02-23T16:31:39.923+09:00
```

JoeGruffins:
I did not know first push is a hyperlink! I see what I should do though. Force push once just to rebase, then make changes. That way the second force push should just be my changes.

client/db/bolt/db.go
```diff
@@ -1624,6 +1624,181 @@ func (idx *timeIndexNewest) add(t uint64, k []byte, b *bbolt.Bucket) {
...
+					return fmt.Errorf("failed to delete order bucket: %v", err)
+				}
+				if perOrderFn != nil {
+					if err := perOrderFn(o); err != nil {
+						return fmt.Errorf("problem performing batch function: %v", err)
+					}
+				}
+				nDeletedOrders++
+			}
+			db.log.Infof("Deleted %d orders from the database in %v.", nDeletedOrders, time.Since(start))
```
chappjc, line 1720: This PR tests quite well, but one thing that stood out is these logs that show an increasing duration, but always 1000 orders. The log would make sense if it either showed cumulative deleted orders (not just the batch), or it showed the time for the last batch (not the cumulative time).

JoeGruffins: Ah. Will change. It is still working with the assumption that the caller is doing this method once at a time.

JoeGruffins: Ok https://github.com/decred/dcrdex/compare/05eec9ee0977dbaa577c007c96f1841688cb8a6f..7c46d81a71b020e6820091992722ad1e176207f6

chappjc (approved):
Works great.  Doing some more testing and a final pass through the code, but planning to merge today.

client/core/core.go
```diff
@@ -6670,3 +6672,228 @@ func (c *Core) WalletLogFilePath(assetID uint32) (string, error) {
...
+			dex.BipIDSymbol(mtch.MetaData.Quote),              // Quote
+			mtch.MatchID.String(),                             // Match ID
+			mtch.OrderID.String(),                             // Order ID
+			numToStr(mtch.Quantity),                           // Quantity
+			numToStr(mtch.Rate),                               // Rate
+			numToStr(mtch.FeeRateSwap),                        // Swap Fee Rate
+			mtch.Address,                                      // Swap Address
+			mtch.Status.String(),                              // Status
+			mtch.Side.String(),                                // Side
+			fmt.Sprintf("%x", mtch.MetaData.Proof.SecretHash), // Secret Hash
```
chappjc, line 6834: Should probably record the secret itself too.
What do you think about the contracts as well?  Hard to know how useful they would be if this is an archived match, meaning core completed it via either refund or redeem, if the swap was broadcast.

chappjc (changes requested):
Hit a couple snags in my final pass.  See what you think about the coin ID decoding idea.

client/db/bolt/db.go
```diff
@@ -1624,6 +1624,181 @@ func (idx *timeIndexNewest) add(t uint64, k []byte, b *bbolt.Bucket) {
...
+		finished   bool
+		olderThanB []byte
+	)
+	if olderThan != nil {
+		olderThanB = uint64Bytes(encode.UnixMilliU(*olderThan))
+	} else {
+		olderThanB = uint64Bytes(encode.UnixMilliU(time.Now()))
+	}
+
+	activeMatches := make(map[order.OrderID]struct{})
```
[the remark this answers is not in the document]
JoeGruffins: Are those inactive orders ever relevant again? I don't think it can hurt so adding just in case.

client/core/core.go
```diff
@@ -6670,3 +6672,228 @@ func (c *Core) WalletLogFilePath(assetID uint32) (string, error) {
...
+			mtch.MetaData.Proof.MakerSwap.String(),            // Maker Swap Coin ID
+			mtch.MetaData.Proof.MakerRedeem.String(),          // Maker Redeem Coin ID
+			mtch.MetaData.Proof.TakerSwap.String(),            // Taker Swap Coin ID
+			mtch.MetaData.Proof.TakerRedeem.String(),          // Taker Redeem Coin ID
+			mtch.MetaData.Proof.RefundCoin.String(),           // Refund Coin ID
```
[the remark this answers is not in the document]
JoeGruffins: Ok.... We could start saving from now, and do a big complicated update to add the sell bool? I won't try to do that here tho, if it's even worth it.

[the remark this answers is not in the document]
JoeGruffins: I guess not.

client/core/core.go
```diff
@@ -6670,3 +6672,228 @@ func (c *Core) WalletLogFilePath(assetID uint32) (string, error) {
...
+	csvWriter.Flush()
+	err = csvWriter.Error()
+	if err != nil {
+		ordersFile.Close()
+		return nil, nil, fmt.Errorf("error writing CSV: %v", err)
+	}
+	return func(ord *db.MetaOrder) error {
+		cord := coreOrderFromTrade(ord.Order, ord.MetaData)
+		unitInfo := func(assetID uint32, symbol string) dex.UnitInfo {
+			assetInfo, err := asset.Info(assetID)
```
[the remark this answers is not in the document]
JoeGruffins: Using.

client/core/core.go
```diff
@@ -6670,3 +6672,228 @@ func (c *Core) WalletLogFilePath(assetID uint32) (string, error) {
...
+			if err != nil {
+				xc, found := c.Exchanges()[cord.Host]
+				if !found {
+					return defaultUnitInfo(symbol)
+				}
+				asset, found := xc.Assets[assetID]
+				if !found || asset.UnitInfo.Conventional.ConversionFactor == 0 {
+					return defaultUnitInfo(symbol)
+				}
+			}
```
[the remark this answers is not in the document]
JoeGruffins: Just using `asset.UnitInfo()` should be fine?

JoeGruffins:
Thought this was right, but eth will give me `unable to delete records: unable to delete matches: unable to delete matches: problem performing batch function: unable to format maker's swap: coin ID wrong length. expected 36, got 32`
Will look at again tomorrow.

client/core/core.go
JoeGruffins: Sounds good, saving it all.

client/db/bolt/db.go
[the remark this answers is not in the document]
JoeGruffins: Checking now for active orders and not deleting matches with those order ID's.

JoeGruffins:
Ok, logic seems right now, if I didn't get confused.

https://github.com/decred/dcrdex/compare/c7696d9a1a6e2438f7f49f31713ebaba5e61c8d4..950b3fe3c6593f477799c71f538ae3624b09223e

client/core/core.go
chappjc: `ContractTxData` is the actual raw transaction hex.  Don't need that saved.

client/core/core.go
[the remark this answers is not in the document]
chappjc: Sure.  It would really be something to have no driver for this asset loaded, yet orders for that asset.

client/core/core.go
```diff
@@ -6670,3 +6672,275 @@ func (c *Core) WalletLogFilePath(assetID uint32) (string, error) {
...
+			return fmt.Sprintf("%d", n)
+		}
+		base, quote := mtch.MetaData.Base, mtch.MetaData.Quote
+		baseSymbol, quoteSymbol := dex.BipIDSymbol(base), dex.BipIDSymbol(quote)
+
+		// If we are maker and buying, we are buying base with quote.
+		makerBuyingAsset, makerSellingAsset := base, quote
+		// If we are either not maker or not buying, invert it. Double
+		// inverse would be no change.
+		if (mtch.Side == order.Taker) != isSell {
```
chappjc, line 6827: Tricky logic.  To check, if I write this out more verbosely:

- maker + selling = makerSwap is base asset (takerSwap is quote)
- maker + buying = makerSwap is quote asset (takerSwap is base)
- taker + selling = makerSwap is quote asset (takerSwap is base)
- taker + buying = makerSwap is base asset (takerSwap is quote)

```go
func matchAssets(base, quote uint32, sell, taker bool) (makerSwapAsset, takerSwapAsset uint32) {
	if sell {
		if taker {
			// maker buy, taker sell
			takerSwapAsset = base // me
			makerSwapAsset = quote
		} else {
			// maker sell, taker buy
			takerSwapAsset = quote
			makerSwapAsset = base // me
		}
	} else {
		if taker {
			// maker sell, taker buy
			takerSwapAsset = quote // me
			makerSwapAsset = base
		} else {
			// maker buy, taker sell
			takerSwapAsset = base
			makerSwapAsset = quote // me
		}
	}
	return
}
```

I'm having a hard time reasoning out if these are equal, partly because of the different variable meanings, where `makerBuyingAsset` refers to the maker's redeem and the taker's init.

Going to make some test cases.

chappjc: I got really bogged down today and didn't actually get to verifying this, but it seems to be the same.

client/core/core.go
```diff
@@ -6670,3 +6672,275 @@ func (c *Core) WalletLogFilePath(assetID uint32) (string, error) {
...
+		"Rate",
+		"Swap Fee Rate",
+		"Swap Address",
+		"Status",
+		"Side",
+		"Secret Hash",
+		"Secret",
+		"Contract Data",
+		"Counter Contract",
+		"Counter TxData",
```
chappjc, line 6798: After testing a bit with some big db files, I think we should omit the counter tx data.  It's just the raw byte of the counterparty's contract txn, which can be obtained elsewhere.

I'm on the fence about the "Contract Data" and "Counter Contract" too.

client/core/core.go
JoeGruffins: https://go.dev/play/p/3oLXnYUtWhQ

I can change the names to takerSwapAsset and makerSwapAsset

client/core/core.go
JoeGruffins: Can remove...

JoeGruffins:
https://github.com/decred/dcrdex/compare/950b3fe3c6593f477799c71f538ae3624b09223e..fbe776fd28935d1cc65a04681acd0d3d16a829ca

client/core/core.go
JoeGruffins: Changed the name.

client/core/core.go
JoeGruffins: Removed.

client/core/core.go
chappjc: Ty for the verification.

buck54321 (commented):
A couple of comments I had prepared but never submitted. Sorry for coming late to the party.

client/core/helpers.go
```diff
@@ -0,0 +1,522 @@
...
+
+	"decred.org/dcrdex/dex"
+	"decred.org/dcrdex/dex/calc"
+	"decred.org/dcrdex/dex/encode"
+	"decred.org/dcrdex/dex/order"
+)
+
+// OrderReader wraps a Order and provides methods for info display.
+// Whenever possible, add an OrderReader methods rather than a template func.
+type OrderReader struct {
```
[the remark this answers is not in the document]
JoeGruffins: It's moved to the core package now, is a different package better?

PR #1480, opened 2022-02-20
fidelity bonds foundation - DCR only, no frontend, manual bond replacement

chappjc:
### Summary

This begins advancing the proof-of-concept from https://github.com/decred/dcrdex/pull/1120 to a proper implementation in the product.  This covers the first several steps outlined in https://github.com/orgs/decred/projects/2
This will likely come out of draft in the next few days.  I have not tested it since the initial draft in #1120, which used different routes and has significant differences with the present work.

To make this transition smoother, the client and server retain all the legacy registration fee machinery, and the client's frontend continues to use the legacy `'register'` protocol, with the `'postbond'` method only accessible via the rpcserver (`dexcctl`).

This PR replaces the registration fee, which was paid to an address provided by the server operator, with a **time-locked fidelity bond, which is redeemable by the user who posted the bond after a certain time**. This creates a time cost to use DCRDEX instead of a monetary cost.

### Bond Transaction Structure

There must be at least two outputs: the bond output (P2SH) and an account commitment (nulldata).

See:
- [`dex/networks/dcr.ExtractBondDetails`](https://github.com/decred/dcrdex/blob/fa639fb00bbf09d736c81c51b48c42e94e99a7d7/dex/networks/dcr/script.go#L331)
- [`client/asset/dcr.(*ExchangeWallet).{MakeBondTx,RefundBond}`](https://github.com/decred/dcrdex/blob/dc2a5fe57b100629ad181e79edee97845139dcbe/client/asset/dcr/dcr.go#L2559)
- [`server/assset/dcr.ParseBondTx`](https://github.com/decred/dcrdex/blob/b6fdb9e72be7c6950784baea9a007ac5f07a3162/server/asset/dcr/dcr.go#L163)

#### Time-locked Bond (output 0)

Output 0 of the fidelity bond transaction must be a P2SH output paying to the bond script.

The redeem script looks similar to the refund path of an atomic swap script:

`<locktime[4]> OP_CHECKLOCKTIMEVERIFY OP_DROP <pubkey[33]> OP_CHECKSIG`

The above uses P2PK since nothing is gained from P2PKH on account of the need to reveal the script pubkey anyway for validation on the server.  (The user must demonstrate ownership of the locked funds by signing a message with the private key corresponding to the pubkey in the TL redeemscript.)

Server validates that this P2SH output actually pays to the provided redeem script.

The pubkey referenced by the script need not be controlled by the user's wallet, but in this implementation it comes from their wallet. The client could even use their account pubkey in the script, but for security reasons it is best not to reuse pubkeys in this bond outputs, which are unspent on the blockchain for long periods.  We may consider using a key pair from the client's HD keychain, but I'm leaning toward this being a wallet job.

The `lockTime` must be after a bond expiry time, the `lockTimeThreshold`. A bond becomes expired for a DEX account when duration until `lockTime` is less than a `bondExpiry` duration, which is on the order of a month. This, a user should create a bond with a `lockTime` at least `bondExpiry` in the future, but a practical `lockTime` would be more like 2x `bondExpiry` in the future so the account is active on the DEX for `bondExpiry`.

#### DEX Account commitment (output 1)

The account commitment OP_RETURN output should reference the account ID that corresponds to the account pubkey in the `postbond` request.

Having it in the raw like this allows the txn alone to identify the account, and without requiring the bond output pay to the account's private keys.

### Protocol

```go
type PostBond struct {
	Signature

}
```

Server must use provided account pubkey to verify signed `postbond` request, otherwise the user could be spamming or putting up a bond for a bogus account or someone else's account.

The response payload:

```go
type PostBondResult struct {
	Amt       uint64 `json:"amt"`
}
```

  - If it is not found, a `PostBondResult` is sent immediately with `Confs=-1` set.  The bond is NOT stored in the DB. This was a courtesy pre-validation. The user should then broadcast the transaction, wait for the required number of confirmations, and send `postbond` again.
  - Otherwise, a response is sent with the observed confirmation count, and a coin waiter is started to watch for the txn to reach the required number of confirmations. The bond is stored in DB as pending. When the transaction reaches the required confirmations, it is flagged active in the DB and a `bondconfirmed` notification is sent.

### TODOs

See the project board at https://github.com/orgs/decred/projects/2.

Main items:

Auto add bond in `client.Core`.  There are comments in the code regarding an option to automatically postbond when previous bonds are about to expire on DEX.  There are questions pertaining to timing, UI, and UX.  Presently this PR requires using dexcctl to call the `postbond` RPC (or a direct http API call to the endpoint of the same name, which is created for said UI plans).

Refining the bond amount.  It's pretty clear the bond amount has to be at least an order of magnitude more than the previous registration fee since you get it back, and it's only an opportunity cost having it locked up.  I'm thinking like 5 DCR per tier (the bond increment).

### Open Questions

What change can we make to facilitate recovery/discovery of bonds if the client's DB is lost?  What might be gained from deriving the private keys from the client's HD key chain instead of using keys from the wallet itself?  Can the bond scripts be deterministic?

When and if accounts go away, would output 1 commit to a magic DEX network key instead of account ID?

Is this transaction structure and account commitment output sufficient to prevent txns from being dual purposed for other services with similar TL output structure?

What architectural and protocol changes might facilitate adding ETH support for time locking funds in some basic contract that server could recognize as a bond?


buck54321:
> What change can we make to facilitate recovery/discovery of bonds if the client's DB is lost? What might be gained from deriving the private keys from the client's HD key chain instead of using keys from the wallet itself? 

Even if we have the keys, we don't have the lock time, so we can't generate the scripts. We could maybe ask the server for their records, but that's not great.

> Can the bond scripts be deterministic?

Could we base the script on a time offset or a block height offset, rather than an absolute time? Isn't that what [DCP0003](https://github.com/decred/dcps/blob/master/dcp-0003/dcp-0003.mediawiki) gave us?

chappjc:
I had given these questions a little more though myself, and another thing that makes it impossible is that the script includes an address from the counterparty, provided at match time.  Hard to see through that.

chappjc:
Broken up in to three pieces now:
- txn definition, creation, parsing in https://github.com/decred/dcrdex/pull/1818
- comms, account auth, and server-side bond tracking in https://github.com/decred/dcrdex/pull/1819
- client integration in https://github.com/decred/dcrdex/pull/1820

PR #1481, opened 2022-02-21
core/test: Use generic drivers.

### LINKED ISSUE #1477: cannot run core test without lgpl
core_test.go imports `dexeth "decred.org/dcrdex/dex/networks/eth"`, apparently just for `dexeth.InitGas`, although afaict this can just be some random but reasonable integer.  It's undesirable to have to specify `-tags lgpl` just to run the `core` unit tests.

Plus, I feel like core tests should ideally not require any of the networks or assets packages.  It requires some client/asset packages to get their drivers registered to power `coinIDString`, but the unit test can easily register fake drivers for this purpose, which we have done elsewhere even in core_test.go (see `tDriver` and calls to `asset.Register`).

JoeGruffins:
In order to avoid importing specific assets, create fake ones modeled
    after actual assets.

closes #1477

### DIFF
--- a/client/core/core_test.go
+++ b/client/core/core_test.go
@@ -44,18 +44,18 @@ import (
 
 func init() {
 	asset.Register(tUTXOAssetA.ID, &tDriver{
-		decodedCoinID: tUTXOAssetA.Symbol,
+		decodedCoinID: tUTXOAssetA.Symbol + "-coin",
 		winfo:         tWalletInfo,
 	})
 	asset.Register(tUTXOAssetB.ID, &tCreator{
 		tDriver: &tDriver{
-			decodedCoinID: tUTXOAssetB.Symbol,
+			decodedCoinID: tUTXOAssetB.Symbol + "-coin",
 			winfo:         tWalletInfo,
 		},
 	})
 	asset.Register(tACCTAsset.ID, &tCreator{
 		tDriver: &tDriver{
-			decodedCoinID: tACCTAsset.Symbol,
+			decodedCoinID: tACCTAsset.Symbol + "-coin",
 			winfo:         tWalletInfo,
 		},
 	})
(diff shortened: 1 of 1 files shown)


martonp (approved):

client/core/core_test.go
```diff
@@ -1432,50 +1445,52 @@ func TestBookFeed(t *testing.T) {
...
 	tILT := &a
 	tILT.Symbol = "ilt"
 	tILT.ID, _ = dex.BipSymbolID(tILT.Symbol)
 
 	// ILT is based on DCR, so use it's coinID decoder. If this changes, update
 	// the following decoder function:
-	decoder := func(coinID []byte) (string, error) {
-		return asset.DecodeCoinID(tDCR.ID, coinID) // using DCR decoder
-	}
+	decodedCoinID := "dcr"
```
chappjc, line 1493: `"ilt"` then?  Also the comment seems to be mostly obsolete now, at least the part about the function, but I don't see the reason to keep the `"dcr"` decoding either.

And does this need to be a variable at all or can we just set this directly in `asset.Register(tILT.ID, , &tDriver{... ` below?

client/core/core_test.go
```diff
@@ -48,36 +42,55 @@ import (
 	"golang.org/x/text/message"
 )
 
+func init() {
+	asset.Register(tUTXOAssetA.ID, &tDriver{
+		decodedCoinID: tUTXOAssetA.Symbol,
```
chappjc, line 47: What do you think about `tUTXOAssetA.Symbol + "-coin"` so it reads like `dcr-coin` in the logs or wherever.

chappjc (approved):
Nice work.  It's great to break these dependencies.  Just a few fairly inconsequential questions.

client/core/core_test.go
JoeGruffins: Yes.

client/core/core_test.go
JoeGruffins: Done.

PR #1482, opened 2022-02-21
eth/nodeclient: Set contract nonces manually.

### LINKED ISSUE #1467: eth: Redeems and refunds can be overwritten and lost without error.
Ethereum uses an account number plus nonce to prevent double spends. It is perfectly normal for a tx to be replaced by another and disappear before it is mined. In dex, this means that redeems and refunds with a low gas price may be silently replaced by the next transaction.

For example, here I have set a redemption gas cap too low to be mined. The ui will show everything is fine:
![image](https://user-images.githubusercontent.com/34998433/152728421-9cec2a25-3896-475c-8ec4-b974f26a124d.png)

Looking at the tx from the harness:
```
[joe@hyrule harness-ctl]$ ./alpha attach --exec 'eth.getTransaction("0x690ae52732c282b0960ed9fbcee827e9502d64c8806a2cb0c39434d2f5a3111a")'
{                                                                                                                                                                                                                                                                               
  accessList: [],                                                                                                                       
  blockHash: null,           
  blockNumber: null,               
  chainId: "0x2a",
  from: "0x1c8f908d99f62522f8e0ad9504ffcc187e5514ca",                                                                                   
  gas: 63000,                                                                                                                           
  gasPrice: 1000000000,                            
  hash: "0x690ae52732c282b0960ed9fbcee827e9502d64c8806a2cb0c39434d2f5a3111a",
  input: "0xf4fd17f900000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000001fa16bdf6e523e2a3bfbcb232ea9f4579023e2c404083137f9a3e03ea6319615b1ca0df49d8965ce5314405e43d819d02c3df33915869825072e97b72774d
ec7e",     
  maxFeePerGas: 1000000000,  
  maxPriorityFeePerGas: 1000000000,
  nonce: 16,              
  r: "0xcf2e9d9f0c440054bc27f7f2395c3d062c6a2e1fbf08a2bccff709972bb6eb77",
  s: "0x4ff989ee3495a90c237ea1b7268526a7d82db64e8588a1723033eca2fec99f7f",
  to: "0x2f68e723b8989ba1c6a9f03e42f33cb7dc9d606f",
  transactionIndex: null,
  type: "0x2",
  v: "0x0",
  value: 0
}
```

Then, I make another trade with a fee that can be mined:
![image](https://user-images.githubusercontent.com/34998433/152728688-53931b73-a1c8-4fe5-99ac-4d68572d10da.png)

The new init tx:
```
[joe@hyrule harness-ctl]$ ./alpha attach --exec 'eth.getTransaction("0x4744e0da9ada37bbaeb2c48f3c229d8a0ecf26371c624214e6733f8860ddb173")'
{
  accessList: [],
  blockHash: "0xff25a8e6b79c7b3a7df443d17dc66df78321c370902aeba2f72322a54c7acf5e",
  blockNumber: 395,
  chainId: "0x2a",
  from: "0x1c8f908d99f62522f8e0ad9504ffcc187e5514ca",
  gas: 135000,
  gasPrice: 2000000007,
  hash: "0x4744e0da9ada37bbaeb2c48f3c229d8a0ecf26371c624214e6733f8860ddb173",
  input: "0xa8793f9400000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000006201c315467667f7ea19ca8c5f0c2d2b60803e88b525430a8849d8ab83cdd509e51f
56a60000000000000000000000001c8f908d99f62522f8e0ad9504ffcc187e5514ca0000000000000000000000000000000000000000000000008ac7230489e80000",
  maxFeePerGas: 200000000000,
  maxPriorityFeePerGas: 2000000000,
  nonce: 16,
  r: "0x4848c3ea1baeab3d8afdbfa2a1a6742e758107e3d6b474ec080230b5b00f14ec",
  s: "0x72b5c445de072c67aa2c5130643d5fcf402454b920c15737f9d33a3f84927888",
  to: "0x2f68e723b8989ba1c6a9f03e42f33cb7dc9d606f",
  transactionIndex: 0,
  type: "0x2",
  v: "0x1",
  value: 10000000000000000000
}
```

And our last redeem:
```
[joe@hyrule harness-ctl]$ ./alpha attach --exec 'eth.getTransaction("0x690ae52732c282b0960ed9fbcee827e9502d64c8806a2cb0c39434d2f5a3111a")'
null
```

The funds appear "immature" until you restart the client, and then there is no trace they ever existed. No error or anything.

If you continue to make trades with a gas fee that is too low, you do see everything come to a stop, as the second low fee will use the next nonce and even if you replace the second tx and onward, all tx must be mined in order, so you are in another, slightly different, bad state.

I guess these tx, need to be saved and resent periodically, until they are mined, unless we are sure the fee will always be adequate?

I wonder if all coins could benefit from redeem and refund watching? There are more serious implications with eth, as the nonce must be in sync, but I would expect it's possible for other coins to have unmined and unnoticed txs? i.e. trade is not finished for client until final redeem/refund tx has a certain number of confirmations.

JoeGruffins:
The v0 contract will reuse account nonces if the tx is not yet
    confirmed. To avoid that get the next nonce from the light eth node pool
    api while holding the nodeClient nonceSendMtx and set manually. Hold the
    mutex until the tx has been sent or errors on sending.

closes  #1467

### DIFF
--- a/client/asset/eth/nodeclient.go
+++ b/client/asset/eth/nodeclient.go
@@ -410,6 +410,13 @@ func (n *nodeClient) initiate(ctx context.Context, contracts []*asset.Contract, 
 		val += c.Value
 	}
 	txOpts, _ := n.txOpts(ctx, val, gas, dexeth.GweiToWei(maxFeeRate))
+	n.nonceSendMtx.Lock()
+	defer n.nonceSendMtx.Unlock()
+	nonce, err := n.leth.ApiBackend.GetPoolNonce(ctx, n.creds.addr)
+	if err != nil {
+		return nil, fmt.Errorf("error getting nonce: %v", err)
+	}
+	txOpts.Nonce = new(big.Int).SetUint64(nonce)
 	if err := n.addSignerToOpts(txOpts); err != nil {
 		return nil, fmt.Errorf("addSignerToOpts error: %w", err)
 	}
@@ -449,6 +456,13 @@ func (n *nodeClient) estimateRefundGas(ctx context.Context, secretHash [32]byte,
 func (n *nodeClient) redeem(ctx context.Context, redemptions []*asset.Redemption, maxFeeRate uint64, contractVer uint32) (tx *types.Transaction, err error) {
 	gas := dexeth.RedeemGas(len(redemptions), contractVer)
 	txOpts, _ := n.txOpts(ctx, 0, gas, dexeth.GweiToWei(maxFeeRate))
+	n.nonceSendMtx.Lock()
+	defer n.nonceSendMtx.Unlock()
+	nonce, err := n.leth.ApiBackend.GetPoolNonce(ctx, n.creds.addr)
+	if err != nil {
+		return nil, fmt.Errorf("error getting nonce: %v", err)
+	}
+	txOpts.Nonce = new(big.Int).SetUint64(nonce)
 	if err := n.addSignerToOpts(txOpts); err != nil {
 		return nil, fmt.Errorf("addSignerToOpts error: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


JoeGruffins:
Tested with low fee txn and concurrent sends. Seems to always set the next nonce correctly. This runs the risk of a tx getting stuck ofc if the fee it too low, which is a bigger problem than with utxo based coins, and one that should be considered more, imo.

martonp (approved):
This LGTM. The code could go into a helper function though to avoid repeating it.

martonp (commented):
`sendToAddr` should also manually generate the nonce. Why not just put this code into `txOpts`?

JoeGruffins:
> Why not just put this code into txOpts?

I think the mtx needs to be held until after the tx has either made it into the pool or not, so after attempting to send. I'm not sure of the actual behavior, but if it will give you the next nonce before a tx is sent, then if one fails you would be skipping the nonce and no longer be able to send. If it will give you the same nonce until it has a tx in the pool using it, then you would be grabbing the same nonce for multiple txn until one is in the pool.

JoeGruffins:
I don't think the helper would be much shorter if I want to hold that mutex outside of the sending function.

chappjc:
Yeah let's run with it as is.  The key is holding the nonce mutex until the rx hits txpool.

chappjc (approved):

PR #1483, opened 2022-02-21
webserver/ui: Generate QR codes for deposit address

### LINKED ISSUE #1449: QR codes for deposit addresses
Where possible, we could generate QR codes to show with the deposit addresses.

Making our own QR code library seemed to be a major challenge last time I looked into it, so assuming we would use a library for this, I'd opt to do it on the backend in Go rather than in JS.  We've kept the JS free of all 3rd party packages (runtime), so we should try to keep it that way.  Unlike dcrdata where the webserver could be overloaded with QR code generation jobs, we don't really have that concern with the self-hosted dexc.

Best bet is probably https://github.com/skip2/go-qrcode, which introduces zero transitive dependencies and seems to be quite popular. Another option is https://github.com/rsc/qr.  I'd want to avoid https://github.com/yeqown/go-qrcode

Then again, if someone wants to spend a little time re-writing all the reed solomon and galois field stuff, I guess that's an option, but it doesn't strike me as a great use of time.

martonp:
Closes #1449

### DIFF
--- a/client/webserver/http.go
+++ b/client/webserver/http.go
@@ -14,8 +14,6 @@ import (
 	"strings"
 	"time"
 
-	"image/color"
-
 	"decred.org/dcrdex/client/asset"
 	"decred.org/dcrdex/client/core"
 	"decred.org/dcrdex/dex"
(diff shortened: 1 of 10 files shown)


martonp:
Screen Shot 2022-02-21 at 2 56 32 PM
Screen Shot 2022-02-21 at 2 56 52 PM

JoeGruffins (approved):
Working great.

I had a comment that disappeared.

client/webserver/http.go
```diff
@@ -198,6 +201,59 @@ func (s *WebServer) handleWalletLogFile(w http.ResponseWriter, r *http.Request)
...
+		return
+	}
+
+	if darkMode {
+		qr.BackgroundColor = color.Black
+		qr.ForegroundColor = color.White
+	} else {
+		qr.BackgroundColor = color.White
+		qr.ForegroundColor = color.Black
+	}
```
JoeGruffins, line 239: I have tried this with two scanners, and one is fine, but one cannot pick up the white qr code on black background. Can they always be white background with black foreground?

ukane-philemon: Well, I tried both we on two scanners and they work well. Maybe scanner specific issue?

chappjc: They both work for me too. If it turns out to be a more widespread issue, we can have the light mode version always displayed by using a white/light patch behind it.

martonp: They both work fine for me too. In the QR code library's docs there was an example with a black background and white foreground so I thought it would be fine, but now I'm doing some research and it's recommended that the foreground should be darker than the background.

martonp:
An error unrelated to the PR is causing the failed checks.

chappjc:
Restarted the CI run.   It was having trouble fetching the linter itself. :P

chappjc (approved):
LGTM.

I have some slight concerns about potential abuse of the route if the service or it's host are not secured properly.  e.g. With the /generateqrcode page not in a `requireLogin` router group, it could be called by anyone with network access to the listening port.  There's no sensitive info as with the wallets page, but it consumes cpu and memory resources to do the qr and png encoding.

A related concern is a max length of the address string.  A long enough string could be trouble for the qr encoder.  Not tested though.

Given the default to listen only on a loopback address, I'm not requesting any changes though.  Just something we should keep in mind.

Issue #1304, opened 2021-11-17
eth: Initiating 8 swaps fails on simnet


martonp:
When trying to initiate greater than 7 swaps in the simnet, the transaction will fail. The contract is fine as loaded the contract into remix and initiated 100, so the simnet is somehow setup incorrectly.

JoeGruffins:
I'll look into this if it's still a problem.

martonp:
Yeah it is, easy way to see it is in the `TestRedeemGas` where 7 swaps are initiated, just change it to 8.

JoeGruffins:
It actually seems to happen at nine initiations now with recent improvements to the efficiency.

The error is:`nodeclient_harness_test.go:706: Unable to initiate swap: tx fee (1.04 ether) exceeds the configured cap (1.00 ether)`

It seems this is due a a hard limit on fee+value. It can be seen as a configurable value here https://geth.ethereum.org/docs/interface/command-line-options `--rpc.txfeecap value                Sets a cap on transaction fee (in ether) that can be sent via the RPC APIs (0 = no cap) (default: 1)`

Do we just want to set to zero? 

Actually not sure where/how to set these RPC values. They don't seem to be accessible through our internal node's config. Maybe just haven't found them yet.

buck54321:
We adopt a bunch of defaults [here](https://github.com/decred/dcrdex/blob/4ee4c478fe7941bff575af1c39617d7945ee091e/client/asset/eth/node.go#L207), including [this one](https://github.com/ethereum/go-ethereum/blob/619a3e70858e60240ce1df75bdf65ba748387e57/eth/ethconfig/config.go#L95). Just need to `ethCfg.RPCTxFeeCap = 0`, I guess.

JoeGruffins:
The test uses a fee of 1000gwei. The math checks out gas used is 135000 + (113000 * 8) or 1039000 times fee gives you 1.039 ether. Normally fee wouldn't be that high.

`ethCfg.RPCTxFeeCap = 0` works. I was looking at the wrong config, where we create the node.

I think setting this is fine if we are monitoring the gas fee anyway and respect the user's max fee rate.

JoeGruffins:
It looks like we currently don't enforce the user's max fee prefs, so need to do that.

PR #1485, opened 2022-02-24
client/eth: Respect user's max fee.

### LINKED ISSUE #1336: client/eth: Respect user's max fee rate.

JoeGruffins:
Check initiation and redeem fee suggestions are not zero or too high. If
    so attempt to get our own gas limit value. If that also fails or is above
    our max limit, do not send the transaction.

closes #1336

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -71,7 +71,7 @@ var (
 		{
 			Key:         "gasfeelimit",
 			DisplayName: "Gas Fee Limit",
-			Description: "This is the highest network fee rate you are willing to " +
+			Description: "This is the highest gas fee rate you are willing to " +
 				"pay on swap transactions. If gasfeelimit is lower than a market's " +
 				"maxfeerate, you will not be able to trade on that market with this " +
 				"wallet.  Units: gwei / gas",
@@ -201,9 +201,10 @@ type ethFetcher interface {
 	locked() bool
 	unlock(pw string) error
 	signData(data []byte) (sig, pubKey []byte, err error)
-	sendToAddr(ctx context.Context, addr common.Address, val uint64) (*types.Transaction, error)
+	sendToAddr(ctx context.Context, addr common.Address, val, maxFeeRate uint64) (*types.Transaction, error)
 	transactionConfirmations(context.Context, common.Hash) (uint32, error)
 	sendSignedTransaction(ctx context.Context, tx *types.Transaction) error
+	netFeeState(ctx context.Context) (baseFees, tipCap *big.Int, err error)
 }
 
 // Check that ExchangeWallet satisfies the asset.Wallet interface.
@@ -426,7 +427,12 @@ func (eth *ExchangeWallet) MaxOrder(lotSize uint64, feeSuggestion uint64, nfo *d
 // PreSwap gets order estimates based on the available funds and the wallet
 // configuration.
 func (eth *ExchangeWallet) PreSwap(req *asset.PreSwapForm) (*asset.PreSwap, error) {
-	maxEst, err := eth.MaxOrder(req.LotSize, req.FeeSuggestion, req.AssetConfig)
+	gasFeeLimit, err := eth.feeLimitWithFallback(req.FeeSuggestion)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get pre swap fee rate limit: %v", err)
+	}
+
+	maxEst, err := eth.MaxOrder(req.LotSize, gasFeeLimit, req.AssetConfig)
 	if err != nil {
 		return nil, err
 	}
@@ -435,7 +441,7 @@ func (eth *ExchangeWallet) PreSwap(req *asset.PreSwapForm) (*asset.PreSwap, erro
 		return nil, fmt.Errorf("%d lots available for %d-lot order", maxEst.Lots, req.Lots)
 	}
 
-	est := eth.estimateSwap(req.Lots, req.LotSize, req.FeeSuggestion, req.AssetConfig)
+	est := eth.estimateSwap(req.Lots, req.LotSize, gasFeeLimit, req.AssetConfig)
 	return &asset.PreSwap{
 		Estimate: est,
 	}, nil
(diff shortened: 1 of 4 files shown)


JoeGruffins:
Actually the server is pretty strict about this, so I guess we can't use our estimation for initiations if the server's is too high.

client/asset/eth/eth.go
```diff
@@ -711,14 +726,14 @@ func (eth *ExchangeWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin
...
+	fees := dexeth.InitGas(1, swaps.AssetVersion) * uint64(len(swaps.Contracts)) * gasFeeLimit
 
 	totalSpend := fees + totalContractValue
 	if totalSpend > totalInputValue {
-		return nil, nil, 0, fmt.Errorf("unfunded contract. %d < %d", totalInputValue, totalSpend)
+		return fail(fmt.Errorf("unfunded contract. %d < %d", totalInputValue, totalSpend))
 	}
 
-	tx, err := eth.node.initiate(eth.ctx, swaps.Contracts, swaps.FeeRate, swaps.AssetVersion)
+	tx, err := eth.node.initiate(eth.ctx, swaps.Contracts, gasFeeLimit, swaps.AssetVersion)
```
buck54321, line 736: Can't do this part. This is validated by the server.

buck54321 (commented):
We realistically never have a fee suggestion from the server, since the wallet implements `FeeRater`.

client/asset/eth/eth.go
chappjc: More than that, the "fee limit" was only ever intended to **block order funding**, nothing else.  In fact, this is the setting that I fought to keep out of the wallet since Core's trade methods (prepareTrackedTrade) could have just as easily applied this user set limit in the one place it was intended.  https://github.com/decred/dcrdex/pull/874#discussion_r545420396

client/asset/eth/eth.go
```diff
@@ -1376,3 +1396,42 @@ func (eth *ExchangeWallet) checkFindRedemptions() {
...
+	}
+	// NOTE: It currently is impossible for the nodeclient to return zero
+	// for the local fee rate limit.
+	if lowerRate == 0 || localFeeRate < lowerRate {
+		lowerRate = localFeeRate
+	}
+	if lowerRate > 0 {
+		eth.log.Warnf("Eth gas rate limit too low! Set above %d gwei to complete current transaction!",
+			lowerRate)
+		return 0, fmt.Errorf("the needed gas fee rate limit (%d gwei) is higher than set max (%d gwei)", lowerRate, eth.gasFeeLimit)
```
chappjc, line 1434: No way we can just block inits, redeems, or refunds.

chappjc:
> We realistically never have a fee suggestion from the server, since the wallet implements `FeeRater`.

Looks like only server has `FeeRate` implemented, but I do think it's reasonable to have client be a `FeeRater` using the `tip +  2*base` formula also.

JoeGruffins:
> We cannot just not do the swap, redeem, or refund txn because the fee suggestion is "too high". Placing and funding the order commits us to executing the swap tnx, and we'd be insane not to refund or redeem on account of fee rates.

While I agree, the server can currently insert any value here and unless the total gas goes over 1 eth (#1304) the wallet will not reject it. On the other side, the server could send something too low and screw the user, which this pr wouldn't solve. There is currently no resolution for txn with fees that are too low with eth. Low fee tx are a bigger problem with eth. IMO client can not blindly trust the server.

Also, I guess I misundersood the meaning of the gas limit. It only applies to initiations, and not to anything else.

Can't we just not do the swap if fee is too high though? I mean initiation. I agree it could be catastrophic to not redeem/refund. I would agree that after you initiate, you kinda agreed to pay whatever fees you need to to complete the swap.

chappjc:
> can currently insert any value here and ... wallet will not reject it

It does not need to reject it because this `swaps.FeeRate` is checked at match time.  When the match is made, that is when the client checks the prescribed swap fee rate to make sure it's not higher than the limit that was recorded _at order placement time_ when FundOrder enforces the user's fee rate limit.  See `parseMatches` and it's check for `swapRate > tracker.metaData.MaxFeeRate`:  https://github.com/decred/dcrdex/blob/26b0c8244c1956c3beccc5dff28899802dc76b40/client/core/core.go#L577-L579 and https://github.com/decred/dcrdex/pull/898

chappjc:
As for too *low*, sure we can use any fee rate as long as its not over the MaxFeeRate because that would potentially eat into the reserves required for the swap(s) for the order.

On the high end limit though, if the wallet did block a swap/init because the fee was higher than it's configured fee limit, that could just be because the fee limit was changed after the order was placed (funded).  I wouldn't object to a second *very* high limit as a sanity check, but it would have to be super high, as in beyond any setting the user might switch their gasFeeLimit setting to.

chappjc:
> As for too _low_, sure we can use any fee rate minimum as long as its not over the MaxFeeRate because that would potentially eat into the reserves required for the swap(s) for the order.

In fact, I would 100% support changes to all asset impls to do this.  Particularly BTC where as taker you are using a fee rate prescribed at match time that was potentially a long time ago if the maker's swap took a while to confirm.  It should be fine to use any rate x that is `swapRate <= x <= db.OrderMetaData.MaxFeeRate`

JoeGruffins:
I think I see it now. The `ord.DEXConfig.MaxFeeRate` is saved to the order metadata and used for all the matches, as a max. So it's not per match I guess? Ignoring changes in the fee rate from time of placing order?

chappjc:
> I think I see it now. The `ord.DEXConfig.MaxFeeRate` is saved to the order metadata and used for all the matches, as a max. So it's not per match I guess? Ignoring changes in the fee rate from time of placing order?

Right, you have a certain configured fee rate limit at order time.  If the server's `MaxFeeRate` is less than that limit, it is recorded in the order's metadata.  That max is enforced for all matches made for that order.  When each match is made, it is checked against that limit.

JoeGruffins:
Ok, then we were respecting the max in the manner intended.

chappjc:
Yes although in this contorted way because the user's fee limit is configurable and this need to be recorded with the order metadata.

I like the idea of bumping the rate if it looks too low though

JoeGruffins:
But the redeem is not checked at all, and we are using the suggestion.

JoeGruffins:
Oh, I see, if it were a FeeRater, it would be using the local fee?

JoeGruffins:
I think @martonp said on matrix he would add that.

chappjc:
Ah, yeah, @buck54321 's point there is that the eth client should be a FeeRater, not needing the suggestion at all.  @martonp said he was going to make that change in his PR.  Would be good to coordinate with him on this so we're on the same page.

JoeGruffins:
I guess I will close this pr and the linked issue. Will make a new issue for suggested fees that are too low.

PR #1486, opened 2022-02-24
client/db: compact DB on shutdown

chappjc:
The latest revisions of bbolt expose DB compaction methods, which is necessary to reclaim any storage space from purged orders and matches from https://github.com/decred/dcrdex/pull/1475

https://github.com/etcd-io/bbolt/pull/248

This adds a compaction and custom backup file support to `client/db/boltdb.BoltDB`.

This also adds a `client/core.(*Core).BackupDB` method so Go consumers of `Core` may issue backup requests at any time.

```
2022-02-25 16:28:05.971 [INF] CORE[DB]: Backing up database...
2022-02-25 16:28:06.982 [INF] CORE[DB]: Compacting database...
2022-02-25 16:28:13.938 [INF] CORE[DB]: Compacted database from 1073156096 => 507797504 bytes (47.32% reduction)
```

### DIFF
--- a/client/db/bolt/db.go
+++ b/client/db/bolt/db.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"sort"
@@ -118,7 +119,7 @@ func NewDB(dbPath string, logger dex.Logger) (dexdb.DB, error) {
 	_, err := os.Stat(dbPath)
 	isNew := os.IsNotExist(err)
 
-	db, err := bbolt.Open(dbPath, 0600, &bbolt.Options{Timeout: 1 * time.Second})
+	db, err := bbolt.Open(dbPath, 0600, &bbolt.Options{Timeout: 3 * time.Second})
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


client/db/bolt/db.go
```diff
@@ -163,14 +164,55 @@ func NewDB(dbPath string, logger dex.Logger) (dexdb.DB, error) {
...
+	return stat.Size()
+}
+
 // Run waits for context cancellation and closes the database.
 func (db *BoltDB) Run(ctx context.Context) {
-	<-ctx.Done()
+	<-ctx.Done() // wait for shutdown to backup and compact
+
+	// Create a backup in the backups folder.
+	db.log.Infof("Backing up database...")
```
chappjc, line 181: I'm doing it this way with a vanilla backup first in case the compacted file has an issue.  But this could be changed to just to a BackupTo with compaction to the "backups" folder, and then copy the resulting file over the source file after closing the source db.

JoeGruffins (approved):
Looks working:
```
2022-02-28 18:25:21.138 [INF] CORE[DB]: Compacting database...
2022-02-28 18:25:21.143 [INF] CORE[DB]: Compacted database from 2097152 => 131072 bytes (6.25% reduction)
```

client/db/bolt/db.go
```diff
@@ -163,14 +164,55 @@ func NewDB(dbPath string, logger dex.Logger) (dexdb.DB, error) {
...
+		db.Close()
+		return
+	}
+
+	// Compact the database by writing into a temporary file, closing the source
+	// DB, and overwriting the original with the compacted temporary file.
+	db.log.Infof("Compacting database...")
+	srcPath := db.Path()                    // before db.Close
+	compFile := srcPath + ".tmp"            // deterministic on *same fs*
+	err = db.BackupTo(compFile, true, true) // overwrite and compact
```
buck54321, line 198: Would it be faster to compact the `Backup` and just copy it back?

buck54321 (approved):

client/db/bolt/db.go
buck54321: I guess if the compaction went bad, you wouldn't have a database left.

chappjc: We had the same thought it seems.  It's what I was getting at in https://github.com/decred/dcrdex/pull/1486#discussion_r814929231.  If we never see trouble with the compacted DBs when we restart dexc, I suppose we can declare it safe and do it directly.

buck54321: Okay as is.

client/db/bolt/db.go
```diff
@@ -1514,31 +1556,93 @@ func (db *BoltDB) withBucket(bkt []byte, viewer txFunc, f bucketFunc) error {
...
+	var dir string
+	if !filepath.IsAbs(dst) {
+		dir = filepath.Dir(db.Path())
+		dst = filepath.Join(dir, dst)
+	}
+	dst = filepath.Clean(dst)
+	dir = filepath.Dir(dst)
+	if dst == filepath.Clean(db.Path()) {
+		return errors.New("destination is the active DB")
+	}
```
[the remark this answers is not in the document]
chappjc: My view of `dex.CleanAndExpandPath` is that it should be used only at the highest levels, such as in the cli apps, rpc and web server handlers, etc. rather than internal to these lower level packages.

Issue #1148, opened 2021-08-03
ui: update to bootstrap 5


chappjc:
For any brave UI devs, we need to update from bootstrap 4 -> 5

Besides needing to get on the current release, it's going to fix all the sass `slash-div` warnings. https://github.com/twbs/bootstrap/issues/34051

chappjc:
For reference @thi4go just got done with dcrdata's Bootstrap upgrade https://github.com/decred/dcrdata/pull/1872
Wasn't terrible.  We still need to be more selective with the forms and helpers imports though because those are kind of meta imports with stuff that's not necessarily needed.  But that PR could be a good reference.

bgptr:
I'd like to work on this.

chappjc:
I think @thi4go was on this.  Is that right?

thi4go:
Go for it @bgptr , let me know if you need a hand. Had to put this one on hold, sorry mate @chappjc should have warned prior.

chappjc:
No problem.  For reference here's the bs 5 update pr for dcrdata https://github.com/decred/dcrdata/pull/1872

Issue #1441, opened 2022-01-23
ui: remove or fix password prompt for app-unlocked users


buck54321:
If the user has selected the option to "remember my password" (app-unlocked), when they attempt to unlock a wallet via the icon on the markets page balance widget, we don't need to display a button to click. We can just unlock. The only problem is that errors are displayed on the form, so for app-unlocked users, we'll need a new place to show errors. We could fall back to the notification icon, or only display the empty form with error message once an error is received. 

**A more immediate concern** though is that the form close button is not styled correctly, as per @jholdstock 

![image](https://user-images.githubusercontent.com/6109680/150678384-438685e5-c042-4f36-b319-d556409cd1a3.png)

jholdstock:
I actually think keeping the extra prompt is a good idea.

1. It isn't totally obvious that clicking on the padlock will unlock the wallet, so informing the user and giving them a chance to cancel is fair.
2. Given that clicking on the unlocked padlock icon does not re-lock the wallet, a confirmation prompt before unlocking is reasonable.
3. Displaying errors in a single place is likely simpler to implement and maintain.

ukane-philemon:
> A more immediate concern though is that the form close button is not styled correctly, as per @jholdstock

This has been fixed in [2fd8f3a](https://github.com/decred/dcrdex/pull/1487/commits/2fd8f3a11a28a22a448cdf2d99b23aaa2493a897)

Is this still an issue?

PR #1487, opened 2022-02-24
[ui] update to bootstrap 5

bgptr:
This diff updates bootstrap from `^4.6.1` to `^5.1.1` using https://github.com/decred/dcrdex/issues/1148 and https://getbootstrap.com/docs/5.0/migration/ as reference.

Changes:

- renamed utilities:
  - `.ml-*` to `.ms-*`
  - `.mr-*` to `.me-*`
  -  `.float-right*` to `.float-end*`
  - `.pl-*` to `.ps-*`
  - `.pr-*` to `.pe-*`
  - `.text-left` to `.text-start`
  - `.text-right` to `.text-end`
- updated bs import paths in `client/webserver/site/src/css/application.scss` and expanded with a new item: `~bootstrap/scss/utilities/api`, because there is a change in file structure in bs 5.  For example, without this import the `d-flex` utility is not defined.
- added `.form-label` to form labels which are now required.
- `media-breakpoint-down()` uses now the breakpoint itself instead of the next breakpoint in `client/webserver/site/src/css/market.scss`.
- checkboxes (and radio boxes) are now customs by default. Since they set SVG data as `background-image`, and because it violates the following Content Security Policy directive: `img-src 'self'` I updated it to `img-src 'self' data:` in `client/webserver/webserver.go`. As an alternative approach, we could force back the inputs to be shown as native eg: with `appearance: auto;`.
- fixed a few layout issues not related to the bs update:
  - layout of 'Authorize Export' and 'Disable Account' modals
          before
             Screenshot 2022-02-24 at 22 40 23
            Screenshot 2022-02-24 at 22 40 50
            Screenshot 2022-02-24 at 22 41 15
         after
             Screenshot 2022-02-24 at 22 46 14
            Screenshot 2022-02-24 at 22 46 19
           Screenshot 2022-02-24 at 22 46 24

  - background-color of tabs on the NewWalletForm and Create button's color in light mode.
        before
           Screenshot 2022-02-24 at 22 38 37
       after
           Screenshot 2022-02-24 at 22 44 10


chappjc:
Good stuff @bgptr.

The npm update workflow is something that I should document though, because there is a process to follow to minimize extraneous updates.  But the first thing is that you'll need to update npm because we've been on `lockfileVersion` 2 since npm 7 (I'm on 8 now).  This is the main reason for the huge deletion count in your current diff.

The first step when making changes to npm deps is to start with the package.json and package-lock.json as on `master` and do an **`npm clean-install` (`npm ci`) rather than `npm install`** so that it installs the versions from the package-lock.json file into node_modules.  Doing `npm install` first when there is no node_modules folder yet will fetch latest from the internet of whatever satisfies package.json (too much) and install those instead of what's in package-lock.json.

```
git restore --source master package.json package-lock.json
npm clean install                         # populate node_modules as per package-lock.json
git add package.json package-lock.json    # checkpoint
```

Then to update just one package, three approaches:
- Just edit the package.json manually then do `npm i` to update the package-lock.json.
- Use ncu (i.e. `ncu --target latest -f bootstrap -u`) followed by `npm i` to apply.  The `ncu` approach is usually most useful when updating multiple packages across major versions.
- `npm install bootstrap@5.1.3 -D`, which updates package.json and package-lock.json in one shot.  **I'd do this**.

Note that with the local node_modules folder populated after doing `npm clean-install` at the beginning, doing `npm install` subsequently will just update the package-lock.json rather than trying to get the latest acceptable versions of _everything_ from the npmjs repo.

```
$  npm install bootstrap@5.1.3 -D

added 1 package, removed 2 packages, changed 1 package, and audited 686 packages in 1s

104 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

$  git diff --stat
 client/webserver/site/package-lock.json | 68 ++++++++++++++++++++++++++------------------------------------------
 client/webserver/site/package.json      |  2 +-
 2 files changed, 27 insertions(+), 43 deletions(-)

$  git add package.json package-lock.json
```

Finally, follow it up with an `npm dedup`.  It may change nothing, but I always do this before committing.

You are welcome to do some other dep updates selectively, but they should be in a separate commit.  `npm out` says there's a few patch updates, but nothing looks required right now.

chappjc:
Thanks for resolving the misplaced "X" buttons too!  Does the fix apply to the unlock dialog as well?  i.e. https://github.com/decred/dcrdex/issues/1441

bgptr:
@chappjc 

Thanks for the instructions and sorry for the improper update. (The main problem was that I'm on node 14. It's best for decrediton dev now.)

I've fixed the misplaced close icon on unlock dialog too:

image

JoeGruffins (approved):
Just saw one part that looked different, the DCR here and for market orders:

image

Tested with:
node v17.6.0
npm 8.5.2

bgptr:
@JoeGruffins thanks for the review. Fixed the mentioned misplacement issue.

buck54321 (changes requested):
A few things I noticed.

1. The Balances table has extra lines between the rows.
2. Checkboxes are no longer aligned
3. Checkbox color has changed

new -> old
image

bgptr:
Thanks for the review @buck54321, nice catches. I've fixed them.

buck54321:
> added .form-label to form labels which are now required.

What do you mean "required"?

buck54321 (approved):
Looks great! I'm not so sure that we need the `form-label` class everywhere, but it's working as is.

bgptr:
It's a breaking change. I read it in the migration guide:
image

It means `margin-bottom: 0.5rem;`

PR #1490, opened 2022-02-25
client/eth: Remove tx fee cap.

### LINKED ISSUE #1304: eth: Initiating 8 swaps fails on simnet

JoeGruffins:
The default fee limit is one eth. Remove the limit so that more
    initations are possible in one transaction.

closes #1304

I can not think of any way to game this. The total expected fees are also visible in the UI, so I think it is save to remove this default limit.

image


client/asset/eth/node.go
```diff
@@ -224,6 +224,7 @@ func startNode(node *node.Node, network dex.Network) (*les.LightEthereum, error)
 	}
 
 	ethCfg.SyncMode = downloader.LightSync
+	ethCfg.RPCTxFeeCap = 0
```
JoeGruffins, line 227: I should add a comment.

JoeGruffins: Added https://github.com/decred/dcrdex/compare/3f016ca304b5a7da9e6b7470247c8aee95a17157..16184b74213accddb30fd34f94e06823ce877744

chappjc:
I asked in the linked issue, but asking again here: why on earth does the test use a gas price of 1000 gwei?  I'm fine with this PR's change I think, but fundamentally isn't it crazy to use 1000 gwei/gas?

JoeGruffins:
I don't know. Want to change it?

martonp (approved):

Issue #1491, opened 2022-02-28
client/btc: SPV balance way off during first trade.


JoeGruffins:
Only seems to happend during the first trade with a new bitcoin SPV wallet.

![image](https://user-images.githubusercontent.com/34998433/155928530-f1827b26-371c-4dc9-ae68-a894027b5051.png)

![image](https://user-images.githubusercontent.com/34998433/155928565-d91f9e11-38d9-41fb-bc53-d790e90d7446.png)

This wallet only had 400 btc sent to it, happens with any amount. To reproduce, make a new bitcoing SPV wallet, send it some funds, start a trade for one lot, and mine a block on each chain.

Does not seem to happen after the first trade. Also only happens during the trade. Locked will become the amount of the trade plus the total balance of the wallet. Available becomes an absurd number. Also happens if you send the wallet funds before creating it. Seems to have to be the first trade ever for the wallet?

chappjc:
Cannot reproduce.  I'm using the beta dcr wallet, and a native btcwallet that gets sent about a dozen 1.1 BTC outputs.
When did you first observe this?

JoeGruffins:
> When did you first observe this?

Just now.

> Any use of dexcctl while the browser ui is up?

No.

> Any browser console errors?

No.

> When the UI is showing this crazy balance, what does dexcctl wallets report?

```
  {
    "symbol": "btc",
    "assetID": 0,
    "version": 0,
    "type": "SPV",
    "traits": 7,
    "open": true,
    "running": true,
    "balance": {
      "available": 18446744072709551463,
      "immature": 0,
      "locked": 5000000000,
      "stamp": "2022-02-28T16:07:36.506214031+09:00",
      "orderlocked": 0,
      "contractlocked": 1000000000
    },
    "address": "bcrt1q9nuljmvrylcwu5f80nf8vcmuvtjkwa3lmd672x",
    "units": "Sats",
    "encrypted": true,
    "peerCount": 1,
    "synced": true,
    "syncProgress": 1
  },
```

Could you try a wallet with just one utxo?

chappjc:
18446744072709551463 is just shy of math.MaxUint64.  There's an underflow with the locked amount.  Will hunt down the issue.  Looks like it's related to what's in trusted vs untrusted balance categories.

PR #1497, opened 2022-02-28
client/asset/btc: unlock outpoint on sendRawTransaction

chappjc:
Resolves https://github.com/decred/dcrdex/issues/1491

```
When sendRawTransaction returns with a non-nil error, even if we hit the
PublishTransaction timeout (defaultBroadcastWait), we still need to
unlock the outpoint in the wallet.
Although listunspent no longer returns the spent outpoint, the
lockedSats computations reference the locked outpoints listed by
listlockunspent, and they are not removed when spent.
```


JoeGruffins (approved):
Looks fixed.

PR #1499, opened 2022-03-02
ui: Reset form when switching between markets

### LINKED ISSUE #1493: UI: Total value of trade's symbol does not change when switching.
When swapping markets, the UI should reflect this change. Total cost of the trade for the dcr_btc market is 10 btc, but for dcr_eth it should be 10 eth. The rate step does change accordingly. The fields could also just be cleared upon switching, I guess.

image
image

bgptr:
When swapping markets, this diff resets rate, lot and quantity input fields, and the total cost of trade text field.

Closes #1493

### DIFF
--- a/client/webserver/site/src/js/markets.js
+++ b/client/webserver/site/src/js/markets.js
@@ -536,7 +536,7 @@ export default class MarketsPage extends BasePage {
     const page = this.page
 
     // reset form inputs
-    page.lotField.value = 0
+    page.lotField.value = ''
     page.qtyField.value = ''
     page.rateField.value = ''
 
--- a/client/webserver/site/src/js/markets.js
+++ b/client/webserver/site/src/js/markets.js
@@ -534,6 +534,12 @@ export default class MarketsPage extends BasePage {
   async setMarket (host, base, quote) {
     const dex = app().user.exchanges[host]
     const page = this.page
+
+    // reset form inputs
+    page.lotField.value = 0
+    page.qtyField.value = ''
+    page.rateField.value = ''
+
     // If we have not yet connected, there is no dex.assets or any other
     // exchange data, so just put up a message and wait for the connection to be
     // established, at which time handleConnNote will refresh and reload.
@@ -589,6 +595,7 @@ export default class MarketsPage extends BasePage {
     this.resolveOrderFormVisibility()
     this.setOrderBttnText()
     this.setCandleDurBttns()
+    this.previewQuoteAmt(false)
   }
 
   /*
--- a/client/webserver/site/src/js/markets.js
+++ b/client/webserver/site/src/js/markets.js
@@ -1697,7 +1697,7 @@ export default class MarketsPage extends BasePage {
   quantityChanged (finalize) {
     const page = this.page
     const order = this.parseOrder()
-    if (order.qty <= 0) {
+    if (order.qty < 0) {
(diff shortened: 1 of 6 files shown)


JoeGruffins (approved):
Looks good to me.

client/webserver/site/src/js/markets.js
```diff
@@ -534,6 +534,12 @@ export default class MarketsPage extends BasePage {
   async setMarket (host, base, quote) {
     const dex = app().user.exchanges[host]
     const page = this.page
+
+    // reset form inputs
+    page.lotField.value = 0
```
buck54321, line 539: Should just set this to `page.lotField.value = ''` too.

buck54321 (changes requested):

buck54321 (approved):

client/webserver/site/src/js/markets.js
chappjc: If we set `lotField.value` to `''` instead of `0`, we can get an error in `lotChanged` if you delete the number:

image

image

`lotField.value` seems to be numeric everywhere else.  Looking at `lotChanged` and `quantityChanged` functions, there it seems intentional to have the other fields as text with this one as numeric.

bgptr: The warnings seem to be independent of the resetting. I see them on the master also. If you clear the lot or the qty field, they appear in the console.
I've pushed a possible solution. To remain consistent, I set the `lotField` field to `''` if it's empty or negative until @buck54321 takes a look at it.

bgptr: I noticed a possible downside of my solution. It disallows writing a floating number to the quantity field. I don't know whether it is meant to be possible, but on master, it is already a bit tricky. I can only type floating quantity numbers on master by starting typing `.` or `,` and the cost of the warning.

image

client/webserver/site/src/js/markets.js
```diff
@@ -1669,10 +1676,11 @@ export default class MarketsPage extends BasePage {
   /* lotChanged is attached to the keyup and change events of the lots input. */
   lotChanged () {
     const page = this.page
-    const lots = parseInt(page.lotField.value)
+    const lots = parseInt(page.lotField.value) || 0
     if (lots <= 0) {
-      page.lotField.value = 0
+      page.lotField.value = ''
```
buck54321, line 1681: I actually prefer to insert the zero in this case, since it helps indicate the problem with the user's input.

client/webserver/site/src/js/markets.js
```diff
@@ -1689,9 +1697,10 @@ export default class MarketsPage extends BasePage {
   quantityChanged (finalize) {
     const page = this.page
     const order = this.parseOrder()
-    if (order.qty <= 0) {
-      page.lotField.value = 0
+    if (order.qty < 0) {
+      page.lotField.value = ''
```
buck54321, line 1701: Same here.

client/webserver/site/src/js/markets.js
buck54321: To fix that warning, use

```Go
const lots = parseInt(page.lotField.value || 0)
```

buck54321: To fix that warning, use

```Go
const lots = parseInt(page.lotField.value || 0)
```

PR #1500, opened 2022-03-02
ui: Fix selecting prices over 1,000

### LINKED ISSUE #1494: UI: Unable to click prices over 1,000.
Undoubtedly the comma again. Unable to select prices in the graph over 1000 and have the automatically inserted into the form. Price is set to zero.

All clicks in this red area insert 0.
image

bgptr:
After clicking on the graph, no need to format the value from number to localized string because the rate input field has a number type.

Closes #1494


JoeGruffins (approved):
Am able to set from graph now.

buck54321 (approved):

chappjc (approved):

PR #1502, opened 2022-03-03
client: redeem cli tool

chappjc:
This is the `redeem` CLI app that is intended to help manually redeem a swap using the dcrdex wallet backends.  The main use case is taker redemption after maker has redeemed, but dexc is somehow hosed.

On the refund side of things we have the raw refund bytes logged that you just broadcast, although we'll need a refund tool for eth.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ node_modules/
 client/cmd/dexc/dexc
 client/cmd/dexcctl/dexcctl
 client/cmd/assetseed/assetseed
+client/cmd/redeem/redeem
 docs/examples/rpcclient/rpcclient
 wiki
 dex/testing/loadbot/loadbot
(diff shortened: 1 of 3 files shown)


buck54321:
> although we'll need a refund tool for eth

I've been thinking about this lately. I think for a clean dexc shutdown, we could dump transactions that will be valid until running again. I have no solution for an unclean shutdown except to just constantly store new refund transactions after nearly every action we take, which is terrible imo. Still hoping for something I don't yet know about ETH to save the day.

chappjc:
For the eth refunds, I'm feeling pretty comfy with the new ability to redeem any contract (by secret hash) *by anyone* (https://github.com/decred/dcrdex/pull/1455), so that you can do something super simple like connect metamask to etherscan and transact with the contract's redeem method and have it refund to the origin account.

Right around that time I thought it would be more or less "problem solved" if we just ensured both the contract address and secret hash were logged in place of signed refund txns: https://github.com/decred/dcrdex/commit/020ff9b906c4341e607490ce444b4e11cb35639f

PR #1503, opened 2022-03-03
client/db: Do not check cancel's trade.

JoeGruffins:
Fixes a panic introduced by #1475


chappjc (approved):
TY for patching my naive code.

PR #1505, opened 2022-03-03
client: resolve failure to start redemption search

chappjc:
This fix resolves a regression in 0.4 with the introduction of SPV wallets.

```
client/asset: Refund must indicate when spent

It is critical that asset.Wallet Refund signal to caller that it is
spent via asset.CoinNotFoundError so that it knows to begin looking
for the counterparty's redeem and move on to redeem as well.

This fixes dcr's refundTx so that it wraps asset.CoinNotFoundError
when lookupTxOutput indicates it is spent.

This also makes explicit the requirement for the asset.Wallet.Refund
method to return asset.CoinNotFoundError when spent.

client/core: allow more Refund errors to trigger redemption search

When locktime is expired, isRefundable will be true and thus block
the shouldBeginFindRedemption case in tick.  We handle this case
by having Refund errors signal when it is spent and then trigger a
redemption search with findMakersRedemption.  This improves
the docs on this handling and the need for Refund to return
asset.CoinNotFoundError when spent. This change also starts
findMakersRedemption for any non-nil error if it is a taker match
and the status is TakerSwapCast, but it does not block refund retries.
```

Note: I considered adding a `SwapConfirmations` call in `isRefundable` when locktime is found to be expired so that `isRefundable` will not block `shouldBeginFindRedemption`, but I am concerned about bugs in `SwapConfirmations`'s `spent bool` leading to failures to refund.  I think it is more foolproof to simply attempt `Refund` and on error, trigger `findMakersRedemption` depending on the error, match side, and match status/stage.

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -2487,7 +2487,8 @@ func (btc *baseWallet) tryRedemptionRequests(ctx context.Context, startBlock *ch
 }
 
 // Refund revokes a contract. This can only be used after the time lock has
-// expired.
+// expired. This MUST return an asset.CoinNotFoundError error if the coin is
+// spent.
 // NOTE: The contract cannot be retrieved from the unspent coin info as the
 // wallet does not store it, even though it was known when the init transaction
 // was created. The client should store this information for persistence across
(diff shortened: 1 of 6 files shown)


JoeGruffins (approved):
Does eth also need to return this error after a new check for `swap.State == dexeth.SSRedeemed` in its `Refund`?

buck54321 (approved):

chappjc:
> Does eth also need to return this error after a new check for `swap.State == dexeth.SSRedeemed` in its `Refund`?

Very good point.  Does this look alright? https://github.com/decred/dcrdex/compare/0a04249ba66533e2926c15297f4c4f9e2d82b5e5..1958940b24cfb431cf64d456f38948b269d0d800
Have not tested that, but it seems good to catch every swap state while it is known there.

JoeGruffins:
>  Does this look alright?

Looks perfect.

PR #1507, opened 2022-03-04
client/ui/btc: Recover BTC SPV wallet

### LINKED ISSUE #1438: similar to Rescan, need a way to Recreate a wallet for recovery
There are a myriad of ways the native btcwallet and/or it's chain service files may become corrupted, necessitating one or more of the manual recovery steps described on https://github.com/decred/dcrdex/wiki/Native-BTC-Wallet-Recovery-and-Rescanning

I was able to corrupt neutrino and prevent it from ever starting again by killing dexc hard a few times.  In my case `neutrino.NewChainService` was returning an EOF error, which through debugging I traced from `newBlockManager` > `cfg.BlockHeaders.ChainTip()` > `(*blockHeaderStore).readHeader` > `h.readRaw(seekDistance)`.  Almost none of the errors from the neutrino constructors supplement the error with any details about what caused the error, so this is nearly impossible for an end user to debug even with the help of a developer.  In such cases, you just need to toast the neutrino data files (and wallet.db) with dexc stopped, start again, and resubmit the wallet config to force it to reinitialize.  We should smooth out this "recreate" process.

martonp:
Adds the ability to delete all the files from the BTC SPV wallet and recreate it from scratch.

- `client/asset/interface`: A `Recoverer` trait is added that includes three functions
  - `GetRecoveryCfg`: Attempts to retrieve a `map[string]string` from the existing wallet that would help in putting the
     wallet back to the state is was in before all the wallet files were deleted.
  - `Destroy`: deletes all the wallet files
  - `ApplyRecoveryCfg`: passes the settings that were retrieved using `GetRecoveryCfg` to a newly created wallet.
- `client/asset/btc`: Implements the `Recoverer` interface. The recovery configuration includes the number of internal and number of external addresses that the wallet generated. This amount of addresses are regenerated in the new instance of the wallet. If this was not done and there were active orders being managed when the user recovered the wallet, the new wallet would not be able to redeem and swaps because it wouldn't have access to the required private keys.
- `client/core`: A `Recover` function is added that destroys the old wallet and creates a new one. It contains a force parameter similar to `Rescan` which unless is set to true, will not allow the user to recover a wallet if there are currently active deals requiring the wallet.  An `activeOrdersErr` is returned from the wallet if `force=false` and there are currently active deals.
- `app`: A button is added on the wallet's page to allow users to recover a wallet if the wallet has the `Recoverer` trait. The "force" workflow is implemented for both rescan and recover. If either of these requests return the `activeOrdersErr`, another popup will show up prompting the user to confirm that they would like to take this action even though there are active orders. If the user confirms, the same request will be resubmitted with `force=true`.

Closes #1438.

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -419,7 +419,6 @@ type WalletConfig struct {
 	RedeemConfTarget uint64  `ini:"redeemconftarget"`
 	ActivelyUsed     bool    `ini:"special:activelyUsed"` //injected by core
 	WalletName       string  `ini:"walletname"`           // RPC
-	Peer             string  `ini:"peer"`                 // SPV
 	Birthday         uint64  `ini:"walletbirthday"`       // SPV
 }
 
@@ -523,7 +522,14 @@ func (d *Driver) Create(params *asset.CreateWalletParams) error {
 		return err
 	}
 
-	return createSPVWallet(params.Pass, params.Seed, walletCfg.adjustedBirthday(), params.DataDir, params.Logger, chainParams)
+	recoveryCfg := new(RecoveryCfg)
+	err = config.Unmapify(params.Settings, recoveryCfg)
+	if err != nil {
+		return err
+	}
+
+	return createSPVWallet(params.Pass, params.Seed, walletCfg.adjustedBirthday(), params.DataDir,
+		params.Logger, recoveryCfg.NumExternalAddresses, recoveryCfg.NumInternalAddresses, chainParams)
 }
 
 // Open opens or connects to the BTC exchange wallet. Start the wallet with its
@@ -626,6 +632,49 @@ var _ asset.Accelerator = (*ExchangeWalletSPV)(nil)
 var _ asset.Rescanner = (*ExchangeWalletSPV)(nil)
 var _ asset.FeeRater = (*ExchangeWalletFullNode)(nil)
 var _ asset.LogFiler = (*ExchangeWalletSPV)(nil)
+var _ asset.Recoverer = (*ExchangeWalletSPV)(nil)
+
+// RecoveryCfg is the information that is transferred from the old wallet
+// to the new one when the wallet is recovered.
+type RecoveryCfg struct {
+	NumExternalAddresses uint32 `ini:"numexternaladdr"`
+	NumInternalAddresses uint32 `ini:"numinternaladdr"`
+}
+
+// GetRecoveryCfg returns information that will help the wallet get
+// back to its previous state after it is recreated. Part of the
+// Recoverer interface.
+func (btc *ExchangeWalletSPV) GetRecoveryCfg() (map[string]string, error) {
+	w := btc.node.(*spvWallet)
+
+	internal, external, err := w.numDerivedAddresses()
+	if err != nil {
+		return nil, err
+	}
+
+	reCfg := &RecoveryCfg{
+		NumInternalAddresses: internal,
+		NumExternalAddresses: external,
+	}
+	cfg, err := config.Mapify(reCfg)
+	if err != nil {
+		return nil, err
+	}
+
+	return cfg, nil
+}
+
+// Destroy will delete all the wallet files so the wallet can be recreated.
+// Part of the Recoverer interface.
+func (btc *ExchangeWalletSPV) Move(backupDir string) error {
+	w := btc.node.(*spvWallet)
+	err := w.moveWalletData(backupDir)
+	if err != nil {
+		return fmt.Errorf("unable to move wallet data: %w", err)
+	}
+
+	return nil
+}
 
 // Rescan satisfies the asset.Rescanner interface, and issues a rescan wallet
 // command if the backend is an SPV wallet.
@@ -892,13 +941,8 @@ func openSPVWallet(cfg *BTCCloneCFG) (*ExchangeWalletSPV, error) {
 		return nil, err
 	}
 
-	var peers []string
-	if walletCfg.Peer != "" {
-		peers = append(peers, walletCfg.Peer)
-	}
-
 	allowAutomaticRescan := !walletCfg.ActivelyUsed
-	btc.node = loadSPVWallet(cfg.WalletCFG.DataDir, cfg.Logger.SubLogger("SPV"), peers, cfg.ChainParams, walletCfg.adjustedBirthday(), allowAutomaticRescan)
+	btc.node = loadSPVWallet(cfg.WalletCFG.DataDir, cfg.Logger.SubLogger("SPV"), cfg.ChainParams, walletCfg.adjustedBirthday(), allowAutomaticRescan)
 
 	return &ExchangeWalletSPV{baseWallet: btc}, nil
 }
@@ -3585,7 +3629,7 @@ func (btc *baseWallet) watchBlocks(ctx context.Context) {
 				blockAllowance := walletBlockAllowance
 				syncStatus, err := btc.node.syncStatus()
 				if err != nil {
-					btc.log.Errorf("Error retreiving sync status before queuing polled block: %v", err)
+					btc.log.Errorf("Error retrieving sync status before queuing polled block: %v", err)
 				} else if syncStatus.Syncing {
 					blockAllowance *= 10
 				}
--- a/client/asset/btc/spv.go
+++ b/client/asset/btc/spv.go
@@ -79,6 +79,8 @@ const (
 	defaultAcctName    = "default"
 )
 
+var wAddrMgrBkt = []byte("waddrmgr")
+
 // btcWallet is satisfied by *btcwallet.Wallet -> *walletExtender.
 type btcWallet interface {
 	PublishTransaction(tx *wire.MsgTx, label string) error
@@ -127,25 +129,43 @@ type neutrinoService interface {
 
 var _ neutrinoService = (*neutrino.ChainService)(nil)
 
+func extendAddresses(extIdx, intIdx uint32, btcw *wallet.Wallet) error {
+	scopedKeyManager, err := btcw.Manager.FetchScopedKeyManager(waddrmgr.KeyScopeBIP0084)
+	if err != nil {
+		return err
+	}
+
+	return walletdb.Update(btcw.Database(), func(dbtx walletdb.ReadWriteTx) error {
+		ns := dbtx.ReadWriteBucket(wAddrMgrBkt)
+		if extIdx > 0 {
+			scopedKeyManager.ExtendExternalAddresses(ns, defaultAcctNum, extIdx)
+		}
+		if intIdx > 0 {
+			scopedKeyManager.ExtendInternalAddresses(ns, defaultAcctNum, intIdx)
+		}
+		return nil
+	})
+}
+
 // createSPVWallet creates a new SPV wallet.
-func createSPVWallet(privPass []byte, seed []byte, bday time.Time, dbDir string, log dex.Logger, net *chaincfg.Params) error {
+func createSPVWallet(privPass []byte, seed []byte, bday time.Time, dbDir string, log dex.Logger, extIdx, intIdx uint32, net *chaincfg.Params) error {
 	netDir := filepath.Join(dbDir, net.Name)
 
 	if err := logNeutrino(netDir); err != nil {
-		return fmt.Errorf("error initializing btcwallet+neutrino logging: %v", err)
+		return fmt.Errorf("error initializing btcwallet+neutrino logging: %w", err)
 	}
 
 	logDir := filepath.Join(netDir, logDirName)
 	err := os.MkdirAll(logDir, 0744)
 	if err != nil {
-		return fmt.Errorf("error creating wallet directories: %v", err)
+		return fmt.Errorf("error creating wallet directories: %w", err)
 	}
 
 	loader := wallet.NewLoader(net, netDir, true, 60*time.Second, 250)
 
 	pubPass := []byte(wallet.InsecurePubPassphrase)
 
-	_, err = loader.CreateNewWallet(pubPass, privPass, seed, bday)
+	btcw, err := loader.CreateNewWallet(pubPass, privPass, seed, bday)
 	if err != nil {
 		return fmt.Errorf("CreateNewWallet error: %w", err)
 	}
@@ -156,11 +176,20 @@ func createSPVWallet(privPass []byte, seed []byte, bday time.Time, dbDir string,
 		}
 	}
 
+	if extIdx > 0 || intIdx > 0 {
+		err = extendAddresses(extIdx, intIdx, btcw)
+		if err != nil {
+			bailOnWallet()
+			return fmt.Errorf("failed to set starting address indexes: %w", err)
+		}
+	}
+
+	// The chain service DB
 	neutrinoDBPath := filepath.Join(netDir, neutrinoDBName)
 	db, err := walletdb.Create("bdb", neutrinoDBPath, true, 5*time.Second)
 	if err != nil {
 		bailOnWallet()
-		return fmt.Errorf("unable to create wallet db at %q: %v", neutrinoDBPath, err)
+		return fmt.Errorf("unable to create neutrino db at %q: %w", neutrinoDBPath, err)
 	}
 	if err = db.Close(); err != nil {
 		bailOnWallet()
@@ -290,7 +319,6 @@ type spvWallet struct {
 	acctName             string
 	netDir               string
 	neutrinoDB           walletdb.DB
-	connectPeers         []string
 
 	txBlocksMtx sync.Mutex
 	txBlocks    map[chainhash.Hash]*hashEntry
@@ -314,7 +342,7 @@ var _ Wallet = (*spvWallet)(nil)
 var _ tipNotifier = (*spvWallet)(nil)
 
 // loadSPVWallet loads an existing wallet.
-func loadSPVWallet(dbDir string, logger dex.Logger, connectPeers []string, chainParams *chaincfg.Params, birthday time.Time, allowAutomaticRescan bool) *spvWallet {
+func loadSPVWallet(dbDir string, logger dex.Logger, chainParams *chaincfg.Params, birthday time.Time, allowAutomaticRescan bool) *spvWallet {
 	return &spvWallet{
 		chainParams:          chainParams,
 		acctNum:              defaultAcctNum,
@@ -323,7 +351,6 @@ func loadSPVWallet(dbDir string, logger dex.Logger, connectPeers []string, chain
 		txBlocks:             make(map[chainhash.Hash]*hashEntry),
 		checkpoints:          make(map[outPoint]*scanCheckpoint),
 		log:                  logger,
-		connectPeers:         connectPeers,
 		tipChan:              make(chan *block, 8),
 		allowAutomaticRescan: allowAutomaticRescan,
 		birthday:             birthday,
@@ -1169,15 +1196,14 @@ func (w *spvWallet) startWallet() error {
 	// mainet and testnet3, add a known reliable persistent peer to be used in
 	// addition to normal DNS seed-based peer discovery.
 	var addPeers []string
+	var connectPeers []string
 	switch w.chainParams.Net {
 	case wire.MainNet:
 		addPeers = []string{"cfilters.ssgen.io"}
 	case wire.TestNet3:
 		addPeers = []string{"dex-test.ssgen.io"}
 	case wire.TestNet, wire.SimNet: // plain "wire.TestNet" is regnet!
-		if len(w.connectPeers) == 0 {
-			w.connectPeers = []string{"localhost:20575"}
-		}
+		connectPeers = []string{"localhost:20575"}
 	}
 	w.log.Debug("Starting neutrino chain service...")
 	chainService, err := neutrino.NewChainService(neutrino.Config{
@@ -1186,7 +1212,7 @@ func (w *spvWallet) startWallet() error {
 		ChainParams:   *w.chainParams,
 		PersistToDisk: true, // keep cfilter headers on disk for efficient rescanning
 		AddPeers:      addPeers,
-		ConnectPeers:  w.connectPeers,
+		ConnectPeers:  connectPeers,
 		// WARNING: PublishTransaction currently uses the entire duration
 		// because if an external bug, but even if the resolved, a typical
 		// inv/getdata round trip is ~4 seconds, so we set this so neutrino does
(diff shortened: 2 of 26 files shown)


client/asset/btc/btc.go
```diff
@@ -775,12 +797,7 @@ func openSPVWallet(cfg *BTCCloneCFG) (*ExchangeWalletSPV, error) {
 		return nil, err
 	}
 
-	var peers []string
```
martonp, line 778: I removed the code related to peer configuration since there is no `ConfigOpt` for the peers.

JoeGruffins:
closes #1438 ?

JoeGruffins (commented):
Works well without active order, but if you have some can't redeem with `[ERR] CORE: 127.0.0.1:17273 tick: {redeemMatches order 39f776f7ae672bd2ca7fcad3ff7ad6b2f8700f5aa30d32fa7e5348db89ee08d5 - {error sending redeem transaction: unable to find key for addr bcrt1qk2aq3m7hl2k5shc33d7hujg5qh8qngvjua04j6}}`

I don't understand why this is, it is just an external address owned by the wallet?

martonp:
I've founds some issues with this, putting it back into draft for now.

client/core/core.go
```diff
@@ -2001,6 +2001,58 @@ func (c *Core) RescanWallet(assetID uint32, force bool) error {
...
+	if err != nil {
+		return newError(authErr, "RecoverWallet password error: %v", err)
+	}
+
+	if !force {
+		for _, dc := range c.dexConnections() {
+			if dc.hasActiveAssetOrders(assetID) {
+				return newError(activeOrdersErr, "active orders for %v", unbip(assetID))
+			}
+		}
```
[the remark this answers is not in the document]
martonp: Even if an order is not yet booked, it is considered active. The user would have to create an order while this request is still processing, and the UI will be spinning the whole time, so I don't think this is a concern.

client/asset/btc/spv.go
```diff
@@ -1244,6 +1241,55 @@ func (w *spvWallet) startWallet() error {
...
+	for _, f := range contents {
+		if f.Name() == "peers.json" {
+			continue
+		}
+		path := filepath.Join(w.netDir, f.Name())
+		if f.IsDir() {
+			_ = os.RemoveAll(path)
+		} else {
+			_ = os.Remove(path)
+		}
```
JoeGruffins, line 1261: I THINK `RemoveAll` is find for both files and directories. Is it not?

client/asset/btc/spv.go
```diff
@@ -1244,6 +1241,55 @@ func (w *spvWallet) startWallet() error {
...
+	}
+
+	wdb := btcw.Database()
+
+	err = walletdb.View(wdb, func(dbtx walletdb.ReadTx) error {
+		ns := dbtx.ReadBucket([]byte("waddrmgr"))
+		props, err := scopedKeyManager.AccountProperties(ns, w.acctNum)
+		internal = props.InternalKeyCount
+		external = props.ExternalKeyCount
+		return err
```
JoeGruffins, line 1287: Should return directly after the error if it is not nil. Otherwise props.Something may panic if props is a pointer. At the end you can return `nil` to satisfy the function return.

client/asset/interface.go
```diff
@@ -45,6 +46,12 @@ func (wt WalletTrait) IsFeeRater() bool {
 	return wt&WalletTraitFeeRater != 0
 }
 
+// IsFeeRater tests if the WalletTrait has the WalletTraitRecoverer bit set,
+// which indicates the presence of a Recover method.
+func (wt WalletTrait) IsRecoverer() bool {
```
JoeGruffins, line 51: IsFeeRater -> IsRecoverer

client/core/core.go
```diff
@@ -2035,6 +2035,101 @@ func (c *Core) RescanWallet(assetID uint32, force bool) error {
...
+// asset before initiating a rescan. WARNING: It is ill-advised to initiate
+// a wallet recovery with active orders unless the wallet db is definitely
+// corrupted and even a rescan will not save it.
+//
+// DO NOT MAKE CONCURRENT CALLS TO THIS FUNCTION WITH THE SAME ASSET.
+func (c *Core) RecoverWallet(assetID uint32, appPW []byte, force bool) error {
+	crypter, err := c.encryptionKey(appPW)
+	if err != nil {
+		return newError(authErr, "RecoverWallet password error: %v", err)
+	}
```
JoeGruffins, line 2052: Could you put a defer `crypter.Close()` after this?

client/core/core.go
```diff
@@ -2035,6 +2035,101 @@ func (c *Core) RescanWallet(assetID uint32, force bool) error {
...
+
+	dbWallet, err := c.db.Wallet(oldWallet.dbID)
+	if err != nil {
+		return fmt.Errorf("error retreiving DB wallet: %w", err)
+	}
+
+	seed, pw, err := c.assetSeedAndPass(assetID, crypter)
+	if err != nil {
+		return err
+	}
```
JoeGruffins, line 2076: Could you do a
```
defer func(){ 
    encode.ClearBytes(seed)
    encode.ClearBytes(pw)
}()
```
after this?

client/core/core.go
```diff
@@ -2035,6 +2035,101 @@ func (c *Core) RescanWallet(assetID uint32, force bool) error {
...
+			c.log.Errorf("RecoverWallet: failed to apply recovery config: %v", err)
+		}
+	}
+
+	c.updateAssetWalletRefs(assetID, newWallet)
+
+	err = newWallet.Unlock(crypter)
+	if err != nil {
+		c.log.Errorf("Error unlocking wallet after recovery: %v", err)
+	}
```
JoeGruffins, line 2126: Maybe just return error on these two errors? If we are unable to unlock, that's definitely a problem.

client/core/core.go
JoeGruffins: Actually, two defers are the thing now, so:
```
defer encode.ClearBytes(seed)
defer encode.ClearBytes(pw)
```

client/asset/btc/btc.go
```diff
@@ -535,14 +535,49 @@ func init() {
...
+	mtx        sync.RWMutex
+	inRecovery bool
+}
+
+// getWallet returns the wallet unless the wallet has gone into recovery.
+func (na *nodeAccessor) getNode() (Wallet, error) {
+	na.mtx.RLock()
+	if na.inRecovery {
+		return nil, fmt.Errorf("btc wallet is currently in recovery")
+	}
```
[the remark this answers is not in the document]
martonp: The reason I used the mutex was to make sure that any other action that started before recovery started would complete before the wallet files got deleted.

client/asset/btc/spv.go
martonp: Yes you're right.

client/core/core.go
martonp: For `Unlock` I agree, but shouldn't `Unlock` still be called if `ApplyRecoveryCfg` fails?

buck54321:
This approach seems off to me. Why not handle this more generally in `Core` and leave the wallets alone? Seems like we're relying on a wallet that has proven useless to fix itself for us.

martonp:
@buck54321 `GetRecoveryCfg` and `ApplyRecoveryCfg` are not required to work, but they make a best effort to recover some useful things from the wallet. If we are able to at least know how many addresses the wallet had created, then we can avoid reusing any addresses in the recovered wallet. Also, if there were active deals when the wallet was recovered, the wallet will have created the addresses needed to do redemptions. If these functions fail, then there's still no problem.. the wallet files are just deleted and the wallet is recreated from scratch. I think it's good to make an effort to try to recover what can be recovered though.

client/core/core.go
JoeGruffins: Well, ok. I guess maybe the old wallet gave us some bogus recovery config, then may error.

client/asset/btc/btc.go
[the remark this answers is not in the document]
JoeGruffins: I see. Makes sense.

JoeGruffins (commented):
Works great!

Deleted my neutrino db and then started a trade. Recovered in the middle of a trade and it finished fine!

I'm not so sure about the naming of the button. "Recover" sounds to me like you want to recover from seed or something. I don't have a better suggestion at the moment though.

Also, I think this screen should also have an "abort/cancel" button:
image

client/asset/btc/btc.go
```diff
@@ -618,21 +653,115 @@ var _ asset.Wallet = (*baseWallet)(nil)
...
 // FeeRate satisfies asset.FeeRater.
 func (btc *ExchangeWalletFullNode) FeeRate() uint64 {
-	rate, err := btc.estimateFee(btc.node, 1)
+	wallet, err := btc.nodeAccessor.getNode()
+	defer btc.nodeAccessor.returnNode()
+	if err != nil {
+		btc.log.Error(err)
+		return 0
+	}
+	rate, err := btc.estimateFee(wallet, 1)
```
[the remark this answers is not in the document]
martonp: Yeah there's no recovery in the `* ExchangeWalletFullNode`. Would it be bad to just leave it as it is and in the  `* ExchangeWalletFullNode` methods, just use `btc.nodeAccessor.node` instead of calling `getNode`?

client/asset/btc/spv.go
```diff
@@ -1244,6 +1241,57 @@ func (w *spvWallet) startWallet() error {
 	return nil
 }
 
+// deleteWalletData will delete all files except peers.json from the
+// wallet's directory.
+func (w *spvWallet) deleteWalletData() error {
```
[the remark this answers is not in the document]
chappjc: I have not reviewed this PR yet (very sorry), but  I would not use `os.Remove`(anything) on a file that's still opened.  Note that there's no file handle arg on that, it's just a syscall.  Before deleting anything I would close it.

client/asset/btc/btc.go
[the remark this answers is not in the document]
martonp: This is the only exclusively `* ExchangeWalletFullNode` node.

client/asset/btc/btc.go
```diff
@@ -563,14 +563,49 @@ func init() {
...
+// must be blocked.
+type nodeAccessor struct {
+	node       Wallet
+	mtx        sync.RWMutex
+	inRecovery bool
+}
+
+// getNode returns the node unless the node has gone into recovery.
+func (na *nodeAccessor) getNode() (Wallet, error) {
+	na.mtx.RLock()
```
chappjc, line 580: Because any node method callers will hang during recovery, I believe we'lll need to rethink this pattern.

It's also unfortunate that a non-spv / external wallet has to use the nodeAccessor when it has no purpose.

Will look closer and make an actionable suggestion.

chappjc:
> This approach seems off to me. Why not handle this more generally in `Core` and leave the wallets alone? Seems like we're relying on a wallet that has proven useless to fix itself for us.

I've reached the same conclusion, but I think we're talking about different aspects.  When I look at `(*Core).RecoverWallet`, it's pretty much what I would do, but the stuff I would avoid is the changes to `client/asset/btc.baseWallet`.

Very roughly speaking, `RecoverWallet` is very logical does what I think is enough and would leave alone `baseWallet` for the most part:
- grab the account indexes for faithful recovery, if possible
- shutdown the wallet (the `xcWallet.Wallet` via the `connector` field's `Disconnect`)
- clobber the old assetdb stuff
- create a new assetdb folder via CreateWallet
- load and connect the new one
- patch up all the dc.trades wallet pointers

AFAICT, there's no strong reason for the `nodeAccessor` business in `baseWallet`.  `Core` just needs to **remove it from the `Core.wallets` map** during the recovery process to prevent it from being reconnected automatically by a caller.  This would be just like what happens at the end of `(*Core).Run` on shutdown.

buck54321:
> * grab the account indexes for faithful recovery, if possible
> * shutdown the wallet (the `xcWallet.Wallet` via the `connector` field's `Disconnect`)
> * clobber the old assetdb stuff
> * create a new assetdb folder via CreateWallet (and I wonder if the recovery settings map could just be merged with the map used here)
> * load and connect the new one
> * patch up all the dc.trades wallet pointers

Exactly the suggestion I was going to write up today. Thanks, @chappjc.

martonp:
@chappjc @buck54321 Without the `nodeAccessor` in the wallet, I think there would need to be some concurrency control in the `trackedTrade`, because even if a wallet is disconnected and does not call `tipChange` anymore, there will still be calls to `tick` from a few other places. When `tick` is called, some methods on the `trackedTrade.fromWallet` and `trackedTrade.toWallet` are called, which causes panics because the files those methods need to use will no longer exist. So I think either there will need to be some added complexity in the `trackedTrade` or in the wallet implementation.

chappjc:
I have done no testing at all on the following, it's just for illustration: https://github.com/chappjc/dcrdex/commit/94a2f0808def33eee6ba36fecffd3f54d8bb0019

client/asset/btc/spv.go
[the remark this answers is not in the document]
JoeGruffins: It would probably be better to use a new database folder and not delete anything, if that is doable.

Or move/rename the old folder to something like `oldName.save_date`? Then can always recover from the recover.

Here's some info: https://www.baeldung.com/linux/open-file-handle-after-move-delete

Seems to suggest, on linux, that both deleting and renaming to a file on the same disc would do what we want. The file handles will still be pointing to the old files, even if they do not exist. Do not know what happens on Windows, but I suspect deletion or renaming would fail because file is in use.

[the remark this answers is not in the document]
martonp: > Here's some info: https://www.baeldung.com/linux/open-file-handle-after-move-delete

That's a good read. `spvWallet.stop()` closes all the file handles, and it is called before `Destroy`. Are you saying that in case `spvWallet.stop()` fails, I should find any file handles that were left and close them before calling `os.Remove`?

A backup folder is definitely a good idea though.

martonp:
@chappjc Those were nice changes, everything worked except `Unmapify` didn't work properly when embedding the `RecoveryCfg` inside `WalletConfig`. Calling `Unmapify` twice using both works though.

It's definitely nicer with `NodeAccessor` gone. I guess the error messages aren't a big deal.

martonp:
Screen Shot 2022-05-14 at 9 12 15 PM
 I've added a cancel button.

chappjc:
> @chappjc Those were nice changes, everything worked except `Unmapify` didn't work properly when embedding the `RecoveryCfg` inside `WalletConfig`. Calling `Unmapify` twice using both works though.
> 
> It's definitely nicer with `NodeAccessor` gone. I guess the error messages aren't a big deal.

Oh I was thinking `json` tags not `ini`.  That's annoying.  It might have an `inline` or `embed` directive (like `omitempty`), but whatever I don't like relying on quirks of imported community packages.

client/webserver/site/src/js/wallets.ts
```diff
@@ -576,6 +621,46 @@ export default class WalletsPage extends BasePage {
...
+    loaded()
+    if (res.code === activeOrdersErrCode) {
+      this.forceUrl = url
+      this.forceReq = req
+      this.showForm(page.confirmForce)
+    } else if (app().checkResponse(res)) {
+      this.closePopups()
+    } else {
+      page.recoverWalletErr.textContent = res.msg
+      Doc.show(page.recoverWalletErr)
```
buck54321, line 642: You'll want to `Doc.hide` this at the top of `recoverWallet` too, as a lingering message for an e.g. password typo might be confusing while the request is being processed.

client/webserver/site/src/js/wallets.ts
```diff
@@ -67,12 +69,22 @@ export default class WalletsPage extends BasePage {
   openAsset: number
   walletAsset: number
   reconfigAsset: number
+  forms: PageElement[]
+  forceReq: any
```
buck54321, line 73: Let's define a type here.

```Typescript
interface RescanRecoveryRequest {
  assetID: number
  appPW?: string
  force?: boolean
}
```

client/asset/interface.go
```diff
@@ -326,6 +336,15 @@ type Rescanner interface {
 	Rescan(ctx context.Context) error
 }
 
+// Recoverer is a wallet implementation with recover functionality.
+type Recoverer interface {
+	// GetRecoveryCfg returns information that will help the wallet get
+	// back to its previous state after it is recreated.
+	GetRecoveryCfg() (map[string]string, error)
+	// Destroy will delete all the wallet files so the wallet can be recreated.
+	Destroy() error
```
buck54321, line 345: I wish that `Destroy` wasn't needed. `*Core` should really just delete the `DataDir`. We can't though, because `*Core` doesn't set the network (mainnet/testnet/simnet) directory and relies on the wallets to do that part. Dang.

buck54321 (changes requested):
Good stuff. Tests well.

client/asset/btc/spv.go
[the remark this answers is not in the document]
JoeGruffins: > I should find any file handles that were left and close them before calling os.Remove?

If this was directed to me... no. It looks like from reading that that it doesn't matter if you rename or delete the file. The new folder will not be the same thing internally for linux, and Windows will probably just fail whatever you do.

The wallet could potentially not run if it is trying to listen on the same ports though, and the bad instance is still using those.

client/asset/interface.go
JoeGruffins: I suggest we change this to `Move()` at least, I think that is a better practice.

client/webserver/locales/en-us.go
```diff
@@ -215,4 +215,11 @@ var EnUS = map[string]string{
 	"Maximum Possible Swap Fees":  "Maximum Possible Swap Fees",
 	"max_fee_conditions":          "This is the most you would ever pay in fees on your swap. Fees are normally assessed at a fraction of this rate. The maximum is not subject to changes once your order is placed.",
 	"wallet_logs":                 "Wallet Logs",
+	"recover":                     "Recover",
+	"recover_wallet":              "Recover Wallet",
+	"recover_warning":             "Recovering your wallet will wipe all wallet data from the disk. You wil have to wait until the wallet resyncs with the network, which could potentially take a long time, before you can use the wallet again.",
+	"wallet_actively_used":        "Wallet actively used!",
+	"confirm_force_message":       "This wallet is activaly managing orders. After taking this action, it will take a long time to resync your wallet, potentially causing orders to fail. Only take this action if absolutely necessary!",
```
chappjc, line 222: activaly

client/asset/btc/spv.go
```diff
@@ -1244,6 +1268,56 @@ func (w *spvWallet) startWallet() error {
...
+	contents, err := os.ReadDir(w.netDir)
+	if err != nil {
+		return err
+	}
+
+	for _, f := range contents {
+		if f.Name() == "peers.json" {
+			continue
+		}
```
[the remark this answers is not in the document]
chappjc: Debatable w.r.t. peers.json, but I don't object to removing it.  Just seemed like something that could accelerate the re-bootstrapping

client/asset/btc/spv.go
```diff
@@ -127,25 +127,43 @@ type neutrinoService interface {
...
 var _ neutrinoService = (*neutrino.ChainService)(nil)
 
+func extendAddresses(extIdx, intIdx uint32, btcw *wallet.Wallet) error {
+	scopedKeyManager, err := btcw.Manager.FetchScopedKeyManager(waddrmgr.KeyScopeBIP0084)
+	if err != nil {
+		return err
+	}
+
+	return walletdb.Update(btcw.Database(), func(dbtx walletdb.ReadWriteTx) error {
+		ns := dbtx.ReadWriteBucket([]byte("waddrmgr"))
```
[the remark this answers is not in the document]
chappjc: I'm starting to wonder how we can make this cannon in exported btcwallet package. So much about the walletdb is exported, but the main namespace keys are not...

client/core/core.go
```diff
@@ -2049,6 +2049,123 @@ func (c *Core) RescanWallet(assetID uint32, force bool) error {
...
+		return errors.New("wallet is not a recoverer")
+	}
+	walletDef, err := walletDefinition(assetID, oldWallet.walletType)
+	if err != nil {
+		return err
+	}
+	// Unseeded wallets shouldn't implement the Recoverer interface. This
+	// is just an additional check for safety.
+	if !walletDef.Seeded {
+		return fmt.Errorf("cannot only recover a seeded wallet")
```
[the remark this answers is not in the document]
chappjc: This check is getting out ahead of the error you'd hit with `asset.CreateWallet` a few lines down

client/asset/driver.go:

```go
// CreateWallet creates a new wallet. Only use Create for seeded wallet types.
func CreateWallet(assetID uint32, seedParams *CreateWalletParams) error {
```

client/asset/btc/btc.go:

```go
// Create creates a new SPV wallet.
func (d *Driver) Create(params *asset.CreateWalletParams) error {
if params.Type != walletTypeSPV {
	return fmt.Errorf("SPV is the only seeded wallet type. required = %q, requested = %q", walletTypeSPV, params.Type)
}
if len(params.Seed) == 0 {
	return errors.New("wallet seed cannot be empty")
}
```

If "recovery" is modified to do something other than `Create`, I would agree it's an unnecessarily limitation, but it's presently central to the definition of recover.  And because we call `Destroy` before doing `Create`, the consequences for not checking this first are potentially severe.

client/core/core.go
```diff
@@ -2049,6 +2049,123 @@ func (c *Core) RescanWallet(assetID uint32, force bool) error {
...
+	}
+
+	dbWallet, err := c.db.Wallet(oldWallet.dbID)
+	if err != nil {
+		return fmt.Errorf("error retreiving DB wallet: %w", err)
```
chappjc, line 2105: retreiving -> retrieving

client/webserver/site/src/html/wallets.tmpl
```diff
@@ -225,7 +225,49 @@
...
+      </div>
+      <div class="d-flex justify-content-end mt-4">
+        <button id="cancelForce" type="button" class="justify-content-center fs15 bg2 danger mx-1">[[[cancel]]]</button>
+        <button id="confirmForceSubmit" type="button" class="justify-content-center fs15 bg2 selected mx-2">[[[confirm]]]</button>
+      </div>
+      <div class="fs15 pt-3 text-center d-hide errcolor" id="confirmForceErr"></div>
+    </form>
+
+    {{- /* RECOVER WALLET AUTHORIZATION */ -}}
+    <form class="d-hide" id="recoverWalletConfirm">
```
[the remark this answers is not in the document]
martonp: I just hadn't updated the cache buster.

client/asset/btc/spv.go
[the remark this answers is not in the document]
martonp: I have the whole folder moving to a backup location now, so peers.json doesn't stay. I could copy it back.. or is that overkill?

client/asset/interface.go
martonp: Updated it to `Move()`.

client/asset/btc/spv.go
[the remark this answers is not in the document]
chappjc: Not worth it

chappjc (approved):
Working well for me.

Only thing I noticed, which may just be from the bootstrap 5 migration, is that the `col-14` lacks left and right padding now.  From what I can tell, columns no longer have `position: relative` or a padding

image
(testnet obv)

vs 0.4.3 where it had 15 pixels

image

Maybe I'm just noticing what's on master now.

PR #1509, opened 2022-03-09
dex/networks/{eth,erc20}: common token stuff

buck54321:
Some of the token primitives used in #1399, offered separately here to ease the integration of that work.
```
dex: Add a simple Token type, outlining the parent-child link and
the UnitInfo. Move other token details to dex/networks/eth

dex/networks/erc20: Rename TOKEN_ADDRESS method to token_address
so that the Go method will be TokenAddress instead of TOKENADDRESS.
Add tx data parsers for transfer and transferFrom calls.

dex/networks/eth: Export calldata parsing machinery.
Add token details and Gases that used to be in /dex.
Add methods for handling tokens with decimals != 18.
```


dex/networks/erc20/txdata.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	if len(decoded.Args) != expArgs {
+		return nil, fmt.Errorf("wrong number of arguments. wanted %d, got %d", expArgs, len(decoded.Args))
+	}
+	return decoded.Args, nil
+}
+
+// ParseTransferFromData parses the calldata used to call the transferFrom
+// method of an ERC20 contract.
+func ParseTransferFromData(data []byte) (sender, recipient common.Address, amount *big.Int, err error) {
```
martonp, line 33: Amount is called value within the function.

dex/networks/erc20/txdata.go
```diff
@@ -0,0 +1,75 @@
+// This code is available on the terms of the project LICENSE.md file,
```
martonp, line 1: Should this file just go into `dex/networks/eth`? Neither client or server have an erc20 folder, so I think we should just get rid of the `dex/networks/erc20` folder.

martonp (commented):
LGTM, just some minor comments.

dex/networks/eth/tokens.go
```diff
@@ -0,0 +1,146 @@
...
+
+// Token is the definition of an ERC20 token, including all of its network and
+// version variants.
+type Token struct {
+	*dex.Token
+	// NetTokens is a mapping of token addresses for each network available.
+	NetTokens map[dex.Network]*NetToken `json:"netAddrs"`
+	// EVMFactor allows for arbitrary ERC20 decimals. For an ERC20 contract,
+	// the relation
+	//    UnitInfo.Conventional.ConversionFactor * Token.EVMFactor = decimals
```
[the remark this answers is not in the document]
buck54321: I see. For USDC/T, I'd want to give the full precision to the user, so I'd set `UnitInfo.Conventional.ConversionFactor = 1e6` and an `EVMFactor = 0`, which I guess means that my use of a default value won't work. Darn. 

That equation should read instead 

`log10(UnitInfo.Conventional.ConversionFactor) + Token.EVMFactor = decimals`

`decimals` is like the orders of magnitude that must be spanned from the EVM units to the display units.

dex/networks/eth/tokens.go
```diff
@@ -0,0 +1,146 @@
...
+	// should hold true.
+	// Since virtually every single asset will use a value of 9 here, a default
+	// value of 9 will be used in AtomicToEVM and EVMToAtomic if EVMFactor is
+	// not set.
+	EVMFactor int64 `json:"decimals"` // default 9
```
martonp, line 30: `json:decimals` will be confusing since decimals means something else.

dex/networks/eth/tokens.go
```diff
@@ -0,0 +1,145 @@
...
+
+// NetToken are the addresses associated with the token and it's versioned
+// swap contracts. The addresses are encoded as [20]byte since it works for
+// erc20 on ETH, but a custom interface or maybe an fmt.Stringer might be used
+// to generalize this in the future. We can't encode as common.Address because
+// it would require an lgpl build, and that would preclude the use of the type
+// in higher-level packages like core.
+type NetToken struct {
+	// Address is the token contract address.
+	Address common.Address `json:"address"`
```
JoeGruffins, line 64: I'm confused about the encoding comment. It is a common.Address here. I think the comment is just old maybe.

dex/networks/eth/tokens.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// SwapContract represents a single swap contract instance.
+type SwapContract struct {
+	Address common.Address
+	Gas     Gases
+}
+
+var Tokens = map[uint32]*Token{
+	// testTokenID = 'dextt.eth' is the used for the test token from
```
JoeGruffins, line 76: is the `ID` used, or just "is ~the~ used"

JoeGruffins (approved):
Makes sense.

martonp (approved):

dex/networks/eth/tokens.go
```diff
@@ -0,0 +1,145 @@
...
+	var evmFactor int64 = 9
+	if t.EVMFactor != nil {
+		evmFactor = *t.EVMFactor
+	}
+	return new(big.Int).Exp(big.NewInt(10), big.NewInt(evmFactor), nil)
+}
+
+// AtomicToEVM converts from DEX atomic units to EVM units.
+func (t *Token) AtomicToEVM(v uint64) *big.Int {
+	return new(big.Int).Mul(big.NewInt(int64(v)), t.factor())
```
chappjc, line 44: Can this be `return big.NewInt(int64(v)).Mul(t.factor())`?

dex/networks/eth/tokens.go
```diff
@@ -0,0 +1,145 @@
...
+// EVMToAtomic converts from raw EVM units to DEX atomic units.
+func (t *Token) EVMToAtomic(v *big.Int) uint64 {
+	vDEX := new(big.Int).Div(v, t.factor())
+	if vDEX.IsUint64() {
+		return vDEX.Uint64()
+	}
+	return 0
+}
+
+// NetToken are the addresses associated with the token and it's versioned
```
chappjc, line 56: its

Also Joe's comment about the encoding description stands out too.  I guess that was left from when there was more of this stuff in the `dex` package.

dex/networks/eth/tokens.go
```diff
@@ -0,0 +1,145 @@
...
+	*dex.Token
+	// NetTokens is a mapping of token addresses for each network available.
+	NetTokens map[dex.Network]*NetToken `json:"netAddrs"`
+	// EVMFactor allows for arbitrary ERC20 decimals. For an ERC20 contract,
+	// the relation
+	//    math.Log10(UnitInfo.Conventional.ConversionFactor) + Token.EVMFactor = decimals
+	// should hold true.
+	// Since most assets will use a value of 9 here, a default value of 9 will
+	// be used in AtomicToEVM and EVMToAtomic if EVMFactor is not set.
+	EVMFactor *int64 `json:"evmFactor"` // default 9
```
chappjc, line 29: I've read the discussion you had with @martonp on this, but I'm still a bit fuzzy on this.
`math.Log10(UnitInfo.Conventional.ConversionFactor) + Token.EVMFactor = decimals`
Is 9 still an expected `EVMFactor` default?  Sounded like you would have 6 and 0?

Take the `testTokenID`, defined below.  It has `ConversionFactor` of 1e8 and `nil` for `EVMFactor`.  The `factor()` method would return 10^9.  But log10(conversionFactor) is 8, so that makes `decimals` get computed as ... 17?

chappjc (commented):
Nice, basically ready to approve, but have a couple Qs.

dex/networks/eth/tokens.go
chappjc: Never mind this.  I hate the `math/big` API.

dex/networks/eth/tokens.go
buck54321: The `ConversionFactor` for `testTokenID` is `GweiFactor = 1e9`, so decimals is `18`, which is the default for a token if not specified.

dex/networks/eth/tokens.go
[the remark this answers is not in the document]
buck54321: I tried a few different names and this was the best I could come up with. `DEXToEVM` isn't really descriptive, `DEXAtomicToETHAtomic` is nonsense. This seemed to be a reasonable middle ground. We use "atomic" to refer to the DEX units throughout, so even though it's a misnomer in this context, I thought we'd all understand.

chappjc (approved):

dex/networks/erc20/txdata.go
buck54321: There's other stuff in the `dex/networks/erc20` folder, but do I understand that you mean that the new txdata stuff and existing params stuff could be moved to dexeth? I personally feel like we should keep the packages separate, but I really hadn't though much about it. I don't see any reason why the networks folder needs to match the structure of either client or server. I guess there might be some other advantages to combining though. I wonder how this choice might affect the future implementation of other EVM assets. If you think it's important enough, you could open an issue.

PR #1510, opened 2022-03-09
client/asset/dcr: simplify dcr.Wallet interface

buck54321:
```
Use wire types for the Wallet interface rather than rpc types. This
is a necessary step towards a built-in wallet. Previous discussion in
1278.
```

### DIFF
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -31,7 +31,6 @@ import (
 	"github.com/decred/dcrd/blockchain/v4"
 	"github.com/decred/dcrd/chaincfg/chainhash"
 	"github.com/decred/dcrd/chaincfg/v3"
-	"github.com/decred/dcrd/dcrec/secp256k1/v4"
 	"github.com/decred/dcrd/dcrec/secp256k1/v4/ecdsa"
 	"github.com/decred/dcrd/dcrutil/v4"
 	chainjson "github.com/decred/dcrd/rpc/jsonrpc/types/v3"
@@ -1171,7 +1170,7 @@ func (dcr *ExchangeWallet) spendableUTXOs() ([]*compositeUTXO, error) {
 		return nil, err
 	}
 	if len(unspents) == 0 {
-		return nil, fmt.Errorf("insufficient funds. 0 DCR available to spend")
+		return nil, fmt.Errorf("insufficient funds. 0 DCR available to spend in account %q", dcr.acctName)
 	}
 
 	// Parse utxos to include script size for spending input.
@@ -2129,9 +2128,6 @@ func (dcr *ExchangeWallet) findRedemptionsInMempool(contractOutpoints []outPoint
 			contractsCount-totalFound-totalCanceled, reason)
 	}
 
-	defer dcr.log.Debugf("%d redemptions found, %d canceled out of %d contracts in mempool",
-		totalFound, totalCanceled, contractsCount)
-
 	mempooler, is := dcr.wallet.(Mempooler)
 	if !is {
 		return
@@ -2156,6 +2152,9 @@ func (dcr *ExchangeWallet) findRedemptionsInMempool(contractOutpoints []outPoint
 			break
 		}
 	}
+
+	dcr.log.Debugf("%d redemptions found, %d canceled out of %d contracts in mempool",
+		totalFound, totalCanceled, contractsCount)
 }
 
 // findRedemptionsInBlockRange attempts to find spending info for the specified
@@ -2209,15 +2208,14 @@ rangeBlocks:
 		}
 		dcr.findRedemptionMtx.RUnlock()
 
-		// Get the cfilters for this block to check if any of the above p2sh scripts is
-		// possibly included in this block.
-		key, filter, err := dcr.wallet.BlockFilter(dcr.ctx, blockHash)
+		bingo, err := dcr.wallet.MatchAnyScript(dcr.ctx, blockHash, contractP2SHScripts)
 		if err != nil { // error retrieving a block's cfilters is a fatal error
-			err = fmt.Errorf("get cfilters error for block %d (%s): %w", blockHeight, blockHash, err)
+			err = fmt.Errorf("MatchAnyScript error for block %d (%s): %w", blockHeight, blockHash, err)
 			dcr.fatalFindRedemptionsError(err, contractOutpoints)
 			return
 		}
-		if !filter.MatchAny(key, contractP2SHScripts) {
+
+		if !bingo {
 			lastScannedBlockHeight = blockHeight
 			continue // block does not reference any of these contracts, continue to next block
 		}
@@ -2274,18 +2272,6 @@ func (dcr *ExchangeWallet) findRedemptionsInTx(scanPoint string, tx *wire.MsgTx,
 
 	redeemTxHash := tx.TxHash()
 
-	extractSecret := func(vin int, contractHash []byte, contractOutputScriptVer uint16) ([]byte, error) {
-		sigScript := tx.TxIn[vin].SignatureScript
-		if len(sigScript) == 0 {
-			return nil, fmt.Errorf("no sigScript")
-		}
-		secret, err := dexdcr.FindKeyPush(contractOutputScriptVer, sigScript, contractHash, dcr.chainParams)
-		if err != nil {
-			return nil, err
-		}
-		return secret, nil
-	}
-
 	for _, contractOutpoint := range contractOutpoints {
 		req, exists := dcr.findRedemptionQueue[contractOutpoint]
 		if !exists {
@@ -2305,7 +2291,7 @@ func (dcr *ExchangeWallet) findRedemptionsInTx(scanPoint string, tx *wire.MsgTx,
 			found++
 
 			scriptHash := dexdcr.ExtractScriptHash(req.contractOutputScriptVer, req.contractP2SHScript)
-			secret, err := extractSecret(i, scriptHash, req.contractOutputScriptVer)
+			secret, err := dexdcr.FindKeyPush(req.contractOutputScriptVer, txIn.SignatureScript, scriptHash, dcr.chainParams)
 			if err != nil {
 				dcr.log.Errorf("Error parsing contract secret for %s from tx input %s:%d in %s: %v",
 					contractOutpoint.String(), redeemTxHash, i, scanPoint, err)
@@ -2487,7 +2473,7 @@ func (dcr *ExchangeWallet) Lock() error {
 func (dcr *ExchangeWallet) Locked() bool {
 	unlocked, err := dcr.wallet.AccountUnlocked(dcr.ctx, dcr.acctName)
 	if err != nil {
-		dcr.log.Errorf("error unlocking: %v", err)
+		dcr.log.Errorf("error checking wallet lock status %v", err)
 		return false // assume wallet is unlocked?
 	}
 	return !unlocked
@@ -2849,7 +2835,7 @@ func (dcr *ExchangeWallet) signTxAndAddChange(baseTx *wire.MsgTx, feeRate uint64
 	// Sign the transaction to get an initial size estimate and calculate
 	// whether a change output would be dust.
 	sigCycles := 1
-	msgTx, err := dcr.wallet.SignRawTransaction(dcr.ctx, baseTx.Copy())
+	msgTx, err := dcr.wallet.SignRawTransaction(dcr.ctx, baseTx)
 	if err != nil {
 		return nil, nil, "", 0, err
 	}
@@ -2916,7 +2902,7 @@ func (dcr *ExchangeWallet) signTxAndAddChange(baseTx *wire.MsgTx, feeRate uint64
 			// Each cycle, sign the transaction and see if there is a need to
 			// raise or lower the fees.
 			sigCycles++
-			msgTx, err = dcr.wallet.SignRawTransaction(dcr.ctx, baseTx.Copy())
+			msgTx, err = dcr.wallet.SignRawTransaction(dcr.ctx, baseTx)
 			if err != nil {
 				return nil, nil, "", 0, err
 			}
@@ -3027,17 +3013,6 @@ func (dcr *ExchangeWallet) createSig(tx *wire.MsgTx, idx int, pkScript []byte, a
 	return sig, priv.PubKey().SerializeCompressed(), nil
 }
 
-// getKeys fetches the private/public key pair for the specified address.
-func (dcr *ExchangeWallet) getKeys(addr stdaddr.Address) (*secp256k1.PrivateKey, *secp256k1.PublicKey, error) {
-	priv, err := dcr.wallet.AddressPrivKey(dcr.ctx, addr)
-	if err != nil {
-		return nil, nil, fmt.Errorf("%w (is wallet locked?)", err)
-	}
-	defer priv.Zero()
-
-	return priv, priv.PubKey(), nil
-}
-
 func (dcr *ExchangeWallet) checkPeers() {
 	ctx, cancel := context.WithTimeout(dcr.ctx, 2*time.Second)
 	defer cancel()
@@ -3141,7 +3116,7 @@ func (dcr *ExchangeWallet) handleTipChange(newTipHash *chainhash.Hash, newTipHei
 	// Redemption search would typically resume from prevTipHeight + 1 unless the
 	// previous tip was re-orged out of the mainchain, in which case redemption
 	// search will resume from the mainchain ancestor of the previous tip.
-	prevTipHeader, isMainchain, err := dcr.blockHeader(dcr.ctx, prevTip.hash)
+	prevTipHeader, isMainchain, _, err := dcr.blockHeader(dcr.ctx, prevTip.hash)
 	if err != nil {
 		// Redemption search cannot continue reliably without knowing if there
 		// was a reorg, cancel all find redemption requests in queue.
@@ -3157,7 +3132,7 @@ func (dcr *ExchangeWallet) handleTipChange(newTipHash *chainhash.Hash, newTipHei
 		// that is the immediate ancestor to the previous tip.
 		ancestorBlockHash := &prevTipHeader.PrevBlock
 		for {
-			aBlock, isMainchain, err := dcr.blockHeader(dcr.ctx, ancestorBlockHash)
+			aBlock, isMainchain, _, err := dcr.blockHeader(dcr.ctx, ancestorBlockHash)
 			if err != nil {
 				notifyFatalFindRedemptionError("Error getting block header %s: %w", ancestorBlockHash, err)
 				return
@@ -3196,7 +3171,7 @@ func (dcr *ExchangeWallet) getBestBlock(ctx context.Context) (*block, error) {
 func (dcr *ExchangeWallet) mainchainAncestor(ctx context.Context, blockHash *chainhash.Hash) (*chainhash.Hash, int64, error) {
 	checkHash := blockHash
 	for {
-		checkBlock, isMainchain, err := dcr.blockHeader(ctx, checkHash)
+		checkBlock, isMainchain, _, err := dcr.blockHeader(ctx, checkHash)
 		if err != nil {
 			return nil, 0, fmt.Errorf("getblockheader error for block %s: %w", checkHash, err)
 		}
@@ -3212,38 +3187,38 @@ func (dcr *ExchangeWallet) mainchainAncestor(ctx context.Context, blockHash *cha
 }
 
 // blockHeader returns the *BlockHeader for the specified block hash, and a
-// bool indicating if the block is considered a mainchain block.
-func (dcr *ExchangeWallet) blockHeader(ctx context.Context, blockHash *chainhash.Hash) (*BlockHeader, bool, error) {
-	blockHeader, err := dcr.wallet.GetBlockHeader(ctx, blockHash)
+// validity bool indicating true when the block has not been stake-invalidated.
+func (dcr *ExchangeWallet) blockHeader(ctx context.Context, blockHash *chainhash.Hash) (blockHeader *BlockHeader, valid, mainchain bool, err error) {
+	blockHeader, err = dcr.wallet.GetBlockHeader(ctx, blockHash)
 	if err != nil {
-		return nil, false, fmt.Errorf("GetBlockHeader error for block %s: %w", blockHash, err)
+		return nil, false, false, fmt.Errorf("GetBlockHeader error for block %s: %w", blockHash, err)
 	}
 
 	checkHash, err := dcr.wallet.GetBlockHash(ctx, int64(blockHeader.Height))
 	if err != nil {
-		return nil, false, fmt.Errorf("error retreiving block hash: %w", err)
+		return nil, false, false, fmt.Errorf("error retreiving block hash: %w", err)
 	}
 
 	if blockHash != checkHash {
-		return nil, false, nil
+		return blockHeader, false, false, nil
 	}
 
 	// Check if there is a validating block
 	nextHash, err := dcr.wallet.GetBlockHash(ctx, int64(blockHeader.Height+1))
 	if err != nil {
 		if strings.Contains(err.Error(), "item does not exist") { // Assume we are the tip
-			return nil, true, nil
+			return blockHeader, true, true, nil
 		}
-		return nil, false, err
+		return nil, false, false, err
 	}
 
 	nextHeader, err := dcr.wallet.GetBlockHeader(ctx, nextHash)
 	if err != nil {
-		return nil, true, fmt.Errorf("error fetching validating block: %w", err)
+		return nil, false, false, fmt.Errorf("error fetching validating block: %w", err)
 	}
 
 	validated := nextHeader.VoteBits&1 != 0
-	return blockHeader, validated, nil
+	return blockHeader, validated, true, nil
 }
 
 func (dcr *ExchangeWallet) cachedBestBlock() block {
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -25,7 +25,6 @@ import (
 	"decred.org/dcrdex/dex/calc"
 	"decred.org/dcrdex/dex/config"
 	dexdcr "decred.org/dcrdex/dex/networks/dcr"
-	"decred.org/dcrwallet/v2/rpc/client/dcrwallet"
 	walletjson "decred.org/dcrwallet/v2/rpc/jsonrpc/types"
 	"github.com/btcsuite/btcwallet/wallet"
 	"github.com/decred/dcrd/blockchain/stake/v4"
@@ -386,7 +385,6 @@ type ExchangeWallet struct {
 	wallet           Wallet
 	chainParams      *chaincfg.Params
 	log              dex.Logger
-	acct             string
 	tipChange        func(error)
 	lastPeerCount    uint32
 	peersChange      func(uint32)
@@ -461,7 +459,7 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network) 
 			return nil, err
 		}
 	} else if makeCustomWallet, ok := customWalletConstructors[cfg.Type]; ok {
-		dcr.wallet, err = makeCustomWallet(cfg, chainParams, logger)
+		dcr.wallet, err = makeCustomWallet(cfg.Settings, chainParams, logger)
 		if err != nil {
 			return nil, fmt.Errorf("custom wallet setup error: %v", err)
 		}
@@ -504,7 +502,6 @@ func unconnectedWallet(cfg *asset.WalletConfig, dcrCfg *Config, chainParams *cha
 	return &ExchangeWallet{
 		log:                 logger,
 		chainParams:         chainParams,
-		acct:                dcrCfg.Account,
 		tipChange:           cfg.TipChange,
 		peersChange:         cfg.PeersChange,
 		fundingCoins:        make(map[outPoint]*fundingCoin),
@@ -555,14 +552,6 @@ func (dcr *ExchangeWallet) Connect(ctx context.Context) (*sync.WaitGroup, error)
 		}
 	}()
 
-	curnet, err := dcr.wallet.Network(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("unable to fetch wallet network: %w", err)
-	}
-	if curnet != dcr.chainParams.Net {
-		return nil, fmt.Errorf("unexpected wallet network %s, expected %s", curnet, dcr.chainParams.Net)
-	}
-
 	// Initialize the best block.
 	dcr.tipMtx.Lock()
 	dcr.currentTip, err = dcr.getBestBlock(ctx)
@@ -598,7 +587,11 @@ func (dcr *ExchangeWallet) Connect(ctx context.Context) (*sync.WaitGroup, error)
 
 // OwnsAddress indicates if an address belongs to the wallet.
 func (dcr *ExchangeWallet) OwnsAddress(address string) (bool, error) {
-	return dcr.wallet.AccountOwnsAddress(dcr.ctx, dcr.acct, address)
+	addr, err := stdaddr.DecodeAddress(address, dcr.chainParams)
+	if err != nil {
+		return false, err
+	}
+	return dcr.wallet.OwnsAddress(dcr.ctx, addr)
 }
 
 // Balance should return the total available funds in the wallet. Note that
@@ -611,7 +604,7 @@ func (dcr *ExchangeWallet) Balance() (*asset.Balance, error) {
 	if err != nil {
 		return nil, err
 	}
-	ab, err := dcr.wallet.AccountBalance(dcr.ctx, dcr.acct, 0)
+	ab, err := dcr.wallet.Balance(dcr.ctx, 0)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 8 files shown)


client/asset/dcr/wallet.go
```diff
@@ -69,19 +67,14 @@ type Wallet interface {
...
-	// EstimateSmartFeeRate returns a smart feerate estimate.
-	EstimateSmartFeeRate(ctx context.Context, confTarget int64, mode chainjson.EstimateSmartFeeMode) (float64, error)
-	// Unspents fetches unspent outputs for the specified account.
-	Unspents(ctx context.Context, account string) ([]walletjson.ListUnspentResult, error)
-	// GetChangeAddress returns a change address from the specified account.
-	GetChangeAddress(ctx context.Context, account string) (stdaddr.Address, error)
+	// OwnsAddress checks if the provided address belongs to the Wallet.
+	OwnsAddress(ctx context.Context, addr stdaddr.Address) (bool, error)
+	// Balance returns the balance breakdown for the Wallet.
+	Balance(ctx context.Context, confirms int32) (*walletjson.GetAccountBalanceResult, error)
```
itswisdomagain, line 73: When the `ExchangeWallet` calls `wallet.Balance`, it expects the balance for a specific account not the entire wallet. This seems to give a different meaning to `Wallet`. `wallet.AccountBalance` makes more sense to me.

Also, re mixed account support, balance reporting may have to deal with more than one account. With this change, implementations have to somehow understand how to properly report balance for mixed and unmixed accounts; and it won't be as simple as summing the sub balances for each account. This can be made simpler by letting the ExchangeWallet handle proper balance reporting while the `Wallet` does a basic wallet work - return the simple balance for a specific account.

itswisdomagain (changes requested):
The dcr.acct change called my attention right away. Will review further after discussing this change.

client/asset/dcr/rpcwallet.go
```diff
@@ -494,18 +529,52 @@ func (w *rpcWallet) SendRawTransaction(ctx context.Context, tx *wire.MsgTx, allo
...
+	if err != nil {
+		return false, fmt.Errorf("getblockheader error for block %s: %w", nextBlockHash, err)
+	}
+	validated := nextBlockHeader.VoteBits&1 != 0
+	return !validated, nil
 }
 
-// GetBlockVerbose returns information about a block, optionally including verbose
-// tx info.
+func (w *rpcWallet) GetBlockHeader(ctx context.Context, blockHash *chainhash.Hash) (*BlockHeader, error) {
```
JoeGruffins, line 559: Add new comment?

client/asset/dcr/rpcwallet.go
```diff
@@ -524,19 +593,19 @@ func (w *rpcWallet) GetTransaction(ctx context.Context, txHash *chainhash.Hash)
 	return tx, nil
 }
 
-// GetRawTransactionVerbose returns details of the tx with the provided hash.
-// Returns asset.CoinNotFoundError if the tx is not found.
-// Part of the Wallet interface.
-func (w *rpcWallet) GetRawTransactionVerbose(ctx context.Context, txHash *chainhash.Hash) (*chainjson.TxRawResult, error) {
-	tx, err := w.client().GetRawTransactionVerbose(ctx, txHash)
-	return tx, translateRPCCancelErr(err)
+func (w *rpcWallet) GetRawTransaction(ctx context.Context, txHash *chainhash.Hash) (*wire.MsgTx, error) {
```
JoeGruffins, line 596: Could use a new comment message.

client/asset/dcr/wallet.go
```diff
@@ -92,54 +85,65 @@ type Wallet interface {
...
 	// network.
 	SendRawTransaction(ctx context.Context, tx *wire.MsgTx, allowHighFees bool) (*chainhash.Hash, error)
-	// GetBlockHeaderVerbose returns block header info for the specified block hash.
-	GetBlockHeaderVerbose(ctx context.Context, blockHash *chainhash.Hash) (*chainjson.GetBlockHeaderVerboseResult, error)
-	// GetBlockVerbose returns information about a block, optionally including verbose
-	// tx info.
-	GetBlockVerbose(ctx context.Context, blockHash *chainhash.Hash, verboseTx bool) (*chainjson.GetBlockVerboseResult, error)
+	// GetBlockHeader returns block header info for the specified block hash.
+	GetBlockHeader(ctx context.Context, blockHash *chainhash.Hash) (*BlockHeader, error)
+	// IsValidMainchain returns true if the block is no orphaned or invalidated,
```
JoeGruffins, line 101: block is no orphaned -> not orphaned

client/asset/dcr/wallet.go
```diff
@@ -92,54 +85,65 @@ type Wallet interface {
...
-	WalletUnlocked(ctx context.Context) bool
-	// AccountUnlocked returns true if the specified account is unlocked.
-	AccountUnlocked(ctx context.Context, account string) (*walletjson.AccountUnlockedResult, error)
-	// LockAccount locks the specified account.
-	LockAccount(ctx context.Context, account string) error
-	// UnlockAccount unlocks the specified account.
-	UnlockAccount(ctx context.Context, account, passphrase string) error
+	// BlockFilter fetches the block filter info for the specified block.
+	BlockFilter(ctx context.Context, blockHash *chainhash.Hash) ([gcs.KeySize]byte, *gcs.FilterV2, error)
+	// Unlocked returns true if the Wallet unlocked.
```
JoeGruffins, line 120: if the Wallet *is* unlocked

client/asset/dcr/wallet.go
```diff
@@ -92,54 +85,65 @@ type Wallet interface {
...
+
+type FeeRateEstimator interface {
+	// EstimateSmartFeeRate returns a smart feerate estimate.
+	EstimateSmartFeeRate(ctx context.Context, confTarget int64, mode chainjson.EstimateSmartFeeMode) (float64, error)
+}
+
+type Mempooler interface {
+	// GetRawMempool returns hashes for all txs in a node's mempool.
+	GetRawMempool(ctx context.Context) ([]*chainhash.Hash, error)
 }
```
JoeGruffins, line 147: I guess the Interfaces themselves could use comments since exported. Do these need to be exported? Will core will use them in the future, or the wallets be split into different packages? The `dcr.Wallet` interface is also exported although I don't think it needs to be.

JoeGruffins (commented):
Unable to trade with as usual on simnet with `error placing order: btc connectAndUnlock error: failed to unlock dcr wallet: -4: encryption/decryption error: account is not encrypted with a unique passphrase`

JoeGruffins:
Unrelated but this log should change based on the wallet: https://github.com/decred/dcrdex/blob/c78ea331d226cd9cf6f192e256c74f15b49e53a8/client/core/core.go#L4086-L4089

client/asset/dcr/dcr.go
```diff
@@ -386,7 +385,6 @@ type baseWallet struct {
 	wallet           Wallet
 	chainParams      *chaincfg.Params
 	log              dex.Logger
-	acct             string
```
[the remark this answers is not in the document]
buck54321: >  Does this mean we no longer care about using a specific wallet account for dex operations?

Depends on what you mean by **we**. `baseWallet` no longer cares, but the `Wallet` implementation does. `rpcWallet` still needs the account number as an input to its constructor. It's still part of the `Config` `struct`, but I'll note that part of that `struct` will probably be split off into a separate `RPCConfig` `struct` when a built-in wallet is also an option. 

I really think it's better to leave account handling up to the implementation. For the built-in `Wallet`, for example, we're not going to ask the user to specify an account number. But I understand your concern about the account handling necessary for the privacy implementation. Looking at #1498, it appears that the account handling could also be done in the `Wallet` implementation, but that might also result in code duplication if we want to add the privacy feature to a built-in wallet later. On the flip side, for a built-in wallet privacy implementation, the user wouldn't be setting the account numbers manually. They would be internal constants. How does that change the picture?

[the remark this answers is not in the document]
itswisdomagain: I think wallets having accounts and having the ability to return balances, unspent outputs, locked utxos for an account are intrinsic to all dcr wallets, whether inbuilt or external. Instead of having `Wallet` implementations carry the burden of tracking an account number (or multiple account numbers, in the case of mixing-enabled wallets) for balance tracking and outputs selection, we can continue to have the `baseWallet` handle all that while leaving `Wallet` implementations to only do basic, generic wallet stuff.

> On the flip side, for a built-in wallet privacy implementation, the user wouldn't be setting the account numbers manually. They would be internal constants. How does that change the picture?

Just as how the btc inbuilt wallet doesn't require any details from users, the inbuilt dcr wallet doesn't need to request account information in the UI. The builtin constants can be used to populate the `ExchangeWallet.{acct,mixedAcct,changeAcct}` fields when the `ExchangeWallet` is being created.

buck54321:
I've reversed the account name argument changes. I'm yet to be convinced that there is any value to handling accounts at the `ExchangeWallet` level. To me, it just appears to complicate the API with extra unnecessary arguments. But since @itswisdomagain is knee-deep in a change that deals with a lot of account handling, we can punt on it until the bigger picture emerges.

buck54321:
I hoped to follow up today by addressing @JoeGruffins' review, but I got distracted. More coming.

client/asset/dcr/dcr.go
```diff
@@ -25,7 +25,6 @@ import (
 	"decred.org/dcrdex/dex/calc"
 	"decred.org/dcrdex/dex/config"
 	dexdcr "decred.org/dcrdex/dex/networks/dcr"
-	"decred.org/dcrwallet/v2/rpc/client/dcrwallet"
 	walletjson "decred.org/dcrwallet/v2/rpc/jsonrpc/types"
 	"github.com/btcsuite/btcwallet/wallet"
```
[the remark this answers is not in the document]
buck54321: Yep. Was for the `CreditReceive` constant.

buck54321:
Thanks for the detailed review, @JoeGruffins. Should be ready for another look.

chappjc:
Nice. Will have my review early this week too.

JoeGruffins (approved):
Thanks for addressing my comments!

client/asset/dcr/wallet.go
```diff
@@ -92,54 +86,64 @@ type Wallet interface {
...
-	UnlockAccount(ctx context.Context, account, passphrase string) error
+	// BlockFilter fetches the block filter info for the specified block.
+	BlockFilter(ctx context.Context, blockHash *chainhash.Hash) ([gcs.KeySize]byte, *gcs.FilterV2, error)
+	// AccountUnlocked returns true if the account is unlocked.
+	AccountUnlocked(ctx context.Context, acctName string) (bool, error)
+	// LockAccount locks the account.
+	LockAccount(ctx context.Context, acctName string) error
+	// UnlockAccount unlocks the Wallet. ExchangeWallet does not differentiate
+	// account locking vs wallet locking, but the underlying implementation may
+	// choose to unlock only the account.
```
itswisdomagain, line 125: Should this explanation also apply to `AccountUnlocked` and `LockAccount`?

client/asset/dcr/dcr.go
```diff
@@ -461,7 +460,7 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network)
 			return nil, err
 		}
 	} else if makeCustomWallet, ok := customWalletConstructors[cfg.Type]; ok {
-		dcr.wallet, err = makeCustomWallet(cfg, chainParams, logger)
+		dcr.wallet, err = makeCustomWallet(cfg.Settings, chainParams, logger)
```
itswisdomagain, line 463: What's the motivation for switching from `cfg` to `cfg.Settings`? This could affect other external implementations, although I haven't checked. A receiver of `cfg` can access `cfg.Settings` if that's all it cares about.

client/asset/dcr/dcr.go
```diff
@@ -1167,12 +1166,12 @@ func (dcr *ExchangeWallet) fund(enough func(sum uint64, size uint32, unspent *co
...
 // spendableUTXOs generates a slice of spendable *compositeUTXO.
 func (dcr *ExchangeWallet) spendableUTXOs() ([]*compositeUTXO, error) {
-	unspents, err := dcr.wallet.Unspents(dcr.ctx, dcr.acct)
+	unspents, err := dcr.wallet.Unspents(dcr.ctx, dcr.acctName)
 	if err != nil {
 		return nil, err
 	}
 	if len(unspents) == 0 {
-		return nil, fmt.Errorf("insufficient funds. 0 DCR available to spend in %q account", dcr.acct)
+		return nil, fmt.Errorf("insufficient funds. 0 DCR available to spend")
```
itswisdomagain, line 1174: If this error gets displayed to users, it would be helpful to include the fact that the unavailability of spendable dcr is specifically for a particular account, as users may have other accounts with spendable utxos. I think even if the error is only being logged, that additional info will be helpful.

client/asset/dcr/dcr.go
```diff
@@ -2469,78 +2464,33 @@ func (dcr *ExchangeWallet) NewAddress() (string, error) {
...
-	// First return locked status of the account, falling back to walletinfo if
-	// the account is not individually password protected.
-	encrypted, unlocked, err := dcr.accountUnlocked(dcr.ctx, dcr.acct)
+	unlocked, err := dcr.wallet.AccountUnlocked(dcr.ctx, dcr.acctName)
 	if err != nil {
-		dcr.log.Errorf("accountunlocked error: %v", err)
-		// return false // or try walletinfo???
-	} else if encrypted {
-		return !unlocked
+		dcr.log.Errorf("error unlocking: %v", err)
```
itswisdomagain, line 2490: error checking if locked?

client/asset/dcr/dcr.go
```diff
@@ -2932,7 +2849,7 @@ func (dcr *ExchangeWallet) signTxAndAddChange(baseTx *wire.MsgTx, feeRate uint64
 	// Sign the transaction to get an initial size estimate and calculate
 	// whether a change output would be dust.
 	sigCycles := 1
-	msgTx, err := dcr.signTx(baseTx)
+	msgTx, err := dcr.wallet.SignRawTransaction(dcr.ctx, baseTx.Copy())
```
itswisdomagain, line 2852: Maybe a comment here and other call sites why using `baseTx.Copy()` instead of just `baseTx`.

client/asset/dcr/dcr.go
```diff
@@ -3096,27 +3013,28 @@ func (dcr *ExchangeWallet) createSig(tx *wire.MsgTx, idx int, pkScript []byte, a
...
 func (dcr *ExchangeWallet) getKeys(addr stdaddr.Address) (*secp256k1.PrivateKey, *secp256k1.PublicKey, error) {
-	wif, err := dcr.wallet.AddressPrivKey(dcr.ctx, addr)
+	priv, err := dcr.wallet.AddressPrivKey(dcr.ctx, addr)
 	if err != nil {
 		return nil, nil, fmt.Errorf("%w (is wallet locked?)", err)
 	}
+	defer priv.Zero()
 
-	priv := secp256k1.PrivKeyFromBytes(wif.PrivKey())
 	return priv, priv.PubKey(), nil
```
itswisdomagain, line 3038: Will the returned `priv` still be useful if it's zeroed immediately after it is returned?

client/asset/dcr/dcr.go
```diff
@@ -3291,50 +3196,54 @@ func (dcr *ExchangeWallet) getBestBlock(ctx context.Context) (*block, error) {
...
+	if err != nil {
+		return nil, false, fmt.Errorf("error retreiving block hash: %w", err)
 	}
-	// Check if the next block invalidated this block's regular tree txs.
-	// This block checks out if there is no following block yet.
-	if blockHeader.NextHash == "" {
-		return true, nil
+
+	if blockHash != checkHash {
+		return nil, false, nil
```
itswisdomagain, line 3228:
```suggestion
		return blockHeader, false, nil
```

client/asset/dcr/dcr.go
```diff
@@ -3291,50 +3196,54 @@ func (dcr *ExchangeWallet) getBestBlock(ctx context.Context) (*block, error) {
...
 	}
-	nextBlockHash, err := chainhash.NewHashFromStr(blockHeader.NextHash)
+
+	// Check if there is a validating block
+	nextHash, err := dcr.wallet.GetBlockHash(ctx, int64(blockHeader.Height+1))
 	if err != nil {
-		return false, fmt.Errorf("block %s has invalid nexthash value %s: %w",
-			block.hash, blockHeader.NextHash, err)
+		if strings.Contains(err.Error(), "item does not exist") { // Assume we are the tip
+			return nil, true, nil
```
itswisdomagain, line 3235:
```suggestion
			return blockHeader, true, nil
```

client/asset/dcr/externaltx.go
```diff
@@ -219,79 +215,66 @@ func (dcr *ExchangeWallet) scanFiltersForTxBlock(ctx context.Context, tx *extern
...
 
-	var txHex string
-	blockTxs := append(blk.RawTx, blk.RawSTx...)
-	for t := range blockTxs {
-		blkTx := &blockTxs[t]
-		if blkTx.Txid == txHash.String() {
-			dcr.log.Debugf("Found mined tx %s in block %d (%s).", txHash, blk.Height, blk.Hash)
-			txHex = blkTx.Hex
+	var msgTx *wire.MsgTx
+	for _, tx := range blk.Transactions {
```
itswisdomagain, line 247: Should also check STransactions to be safe.

client/asset/dcr/externaltx.go
```diff
@@ -219,79 +215,66 @@ func (dcr *ExchangeWallet) scanFiltersForTxBlock(ctx context.Context, tx *extern
...
+			op:               newOutPoint(&txHash, uint32(i)),
 			tree:             determineTxTree(msgTx),
 			lastScannedBlock: blockHash,
 		}
 	}
-	for t := range blockTxs {
-		blkTx := &blockTxs[t]
-		if blkTx.Txid == txHash.String() {
-			continue // original tx, ignore
+	for _, tx := range blk.Transactions {
```
itswisdomagain, line 271: STransactions too.

client/asset/dcr/externaltx.go
```diff
@@ -388,32 +366,30 @@ func (dcr *ExchangeWallet) isOutputSpent(ctx context.Context, output *outputSpen
...
 
 	iHeight := startBlock.height
 	iHash := startBlock.hash
 	bestBlock := dcr.cachedBestBlock()
 	for {
-		blockFilter, err := dcr.getBlockFilterV2(ctx, iHash)
+		key, filter, err := dcr.wallet.BlockFilter(ctx, iHash)
 		if err != nil {
-			return nil, lastScannedHash, err
+			return nil, lastScannedHash, 0, err
```
itswisdomagain, line 378: 0 -> iHeight. There may have been a scan in a previous loop iteration, especially if `lastScannedHash != nil`.

client/asset/dcr/wallet.go
```diff
@@ -92,54 +86,64 @@ type Wallet interface {
...
-	// SignRawTransaction signs the provided transaction.
-	SignRawTransaction(ctx context.Context, txHex string) (*walletjson.SignRawTransactionResult, error)
+	// ExternalAddress returns a new external address,
+	ExternalAddress(ctx context.Context, acctName string) (stdaddr.Address, error)
+	// InternalAddress returns a change address from the Wallet.
+	InternalAddress(ctx context.Context, acctName string) (stdaddr.Address, error)
+	// SignRawTransaction signs the provided transaction. SignRawTransaction
+	// is not used for redemptions, so previous outpoints and scripts should
+	// be known by the wallet.
+	SignRawTransaction(context.Context, *wire.MsgTx) (*wire.MsgTx, error)
```
itswisdomagain, line 96: I think this doc should mention that the `msgTx` passed in could be modified, so callers should generally pass in a copy.

itswisdomagain (changes requested):
Another pass.

client/asset/dcr/externaltx.go
```diff
@@ -127,12 +123,12 @@ func (dcr *ExchangeWallet) txBlockFromCache(ctx context.Context, tx *externalTx)
 		return nil, nil
 	}
 
-	txBlockStillValid, err := dcr.isMainchainBlock(ctx, tx.block)
```
chappjc, line 130: We need to review all the previous calls to `isMainchainBlock`, and the new callers of `blockHeader` with the new bool arg because stake validation and main chain were being combined.

In places where we're doing reorg stuff or main chain ancestor identification, the stakeholder validation is wrong to consider.

In places like `isOutputSpent`, we *may* want to consider if the block's regular tree is invalidated.

Things have been conflated, in part because the previous name `isMainchainBlock` was not accurate.

client/asset/dcr/rpcwallet.go
```diff
@@ -64,6 +68,8 @@ type rpcWallet struct {
 
 // Ensure rpcWallet satisfies the Wallet interface.
 var _ Wallet = (*rpcWallet)(nil)
+var _ Mempooler = (*rpcWallet)(nil)
+var _ FeeRateEstimator = (*rpcWallet)(nil)
```
chappjc, line 72: I get where you're going with this, where a native/integrated dcrwallet would likely be known to be exclusively SPV, but the `rpcWallet` can be either full-node backed or SPV.  On top of that, we don't find out if it's SPV mode until it connects.

client/asset/dcr/rpcwallet.go
```diff
@@ -642,9 +723,13 @@ func (w *rpcWallet) PeerCount(ctx context.Context) (uint32, error) {
...
 // Part of the Wallet interface.
-func (w *rpcWallet) AddressPrivKey(ctx context.Context, address stdaddr.Address) (*dcrutil.WIF, error) {
-	wif, err := w.client().DumpPrivKey(ctx, address)
-	return wif, translateRPCCancelErr(err)
+func (w *rpcWallet) AddressPrivKey(ctx context.Context, address stdaddr.Address) (*secp256k1.PrivateKey, error) {
+	wif, err := w.rpcClient.DumpPrivKey(ctx, address)
+	if err != nil {
+		return nil, translateRPCCancelErr(err)
+	}
+	priv := secp256k1.PrivKeyFromBytes(wif.PrivKey())
```
chappjc, line 731: I know this was like this before, but it might be prudent to do the parsing manually so we can validate the private key bytes.

```go
var priv PrivateKey
if overflow := priv.Key.SetByteSlice(privKeyBytes); overflow || priv.Key.IsZero() {
	return nil, errors.New("invalid private key")
}
```

as described in the docs here https://pkg.go.dev/github.com/decred/dcrd/dcrec/secp256k1/v4#PrivKeyFromBytes

chappjc (changes requested):
I like the way this is going quite a bit.  Way less string conversions.

client/asset/dcr/dcr.go
```diff
@@ -555,14 +554,6 @@ func (dcr *ExchangeWallet) Connect(ctx context.Context) (*sync.WaitGroup, error)
 		}
 	}()
 
-	curnet, err := dcr.wallet.Network(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("unable to fetch wallet network: %w", err)
-	}
-	if curnet != dcr.chainParams.Net {
-		return nil, fmt.Errorf("unexpected wallet network %s, expected %s", curnet, dcr.chainParams.Net)
```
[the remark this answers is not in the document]
itswisdomagain: I thought it would be safer to make this check from the `ExchangeWallet` but I suppose since the `Network` method is only used here, maybe we can do without it and assume `Wallet` impls will do the right thing?

client/asset/dcr/rpcwallet.go
itswisdomagain: And on reconnection as well. I've toyed with shutting down dcrwallet after dexc initially connected to it and restarted dcrwallet with spv where it was previously using full node. The exchange wallet notices when this change occurs on reconnect.

client/asset/dcr/dcr.go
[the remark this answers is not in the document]
chappjc: Whatever really.

client/asset/dcr/wallet.go
buck54321: Just needs to be removed, I guess. Applied to previous revision with method `Unlock`.

client/asset/dcr/dcr.go
buck54321: Please take a look, but I think that based on the structure of `asset.WalletConfig`, it makes sense to restrain any `Wallet`-implementation-specific configuration to the `Settings` `map`.

client/asset/dcr/dcr.go
buck54321: Whoops

client/asset/dcr/wallet.go
chappjc: I'm not seeing any modifications made by `(*rpcWallet).SignRawTransaction` to the input `MsgTx`, so alternatively this interface comment should require the implementations do not modify the input.

client/asset/dcr/externaltx.go
```diff
@@ -219,79 +215,66 @@ func (dcr *ExchangeWallet) scanFiltersForTxBlock(ctx context.Context, tx *extern
...
 
-func (dcr *ExchangeWallet) findTxInBlock(ctx context.Context, txHash *chainhash.Hash, txScripts [][]byte, blockHash *chainhash.Hash) (*wire.MsgTx, []*outputSpenderFinder, error) {
-	blockFilter, err := dcr.getBlockFilterV2(ctx, blockHash)
+func (dcr *ExchangeWallet) findTxInBlock(ctx context.Context, txHash chainhash.Hash, txScripts [][]byte, blockHash *chainhash.Hash) (*wire.MsgTx, []*outputSpenderFinder, error) {
+	key, filter, err := dcr.wallet.BlockFilter(ctx, blockHash)
 	if err != nil {
 		return nil, nil, err
 	}
-	if !blockFilter.MatchAny(txScripts) {
+	if !filter.MatchAny(key, txScripts) {
```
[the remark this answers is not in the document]
buck54321: Miniscule difference is that we were only using `Match` at one location instead of `MatchAll`. Changed it to `MatchAnyScript` as suggested though.

client/asset/dcr/dcr.go
```diff
@@ -2130,16 +2129,24 @@ func (dcr *ExchangeWallet) findRedemptionsInMempool(contractOutpoints []outPoint
 			contractsCount-totalFound-totalCanceled, reason)
 	}
 
-	mempoolTxs, err := dcr.wallet.GetRawMempool(dcr.ctx, chainjson.GRMAll)
+	defer dcr.log.Debugf("%d redemptions found, %d canceled out of %d contracts in mempool",
+		totalFound, totalCanceled, contractsCount)
```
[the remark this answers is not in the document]
buck54321: Actually, these values are logged separately in `logAbandon`. I shouldn't have touched this

client/asset/dcr/rpcwallet.go
```diff
@@ -486,7 +486,8 @@ func (w *rpcWallet) ExternalAddress(ctx context.Context, acctName string) (stdad
 
 // SignRawTransaction signs the provided transaction using rpc RawRequest.
 // Part of the Wallet interface.
-func (w *rpcWallet) SignRawTransaction(ctx context.Context, baseTx *wire.MsgTx) (*wire.MsgTx, error) {
+func (w *rpcWallet) SignRawTransaction(ctx context.Context, inTx *wire.MsgTx) (*wire.MsgTx, error) {
+	baseTx := inTx.Copy()
```
chappjc, line 490: OK if this is just defensive, but `baseTx` is never modified by this method, only serialized, so the copy seems to be superfluous at present.

client/asset/dcr/dcr.go
```diff
@@ -3291,50 +3171,54 @@ func (dcr *ExchangeWallet) getBestBlock(ctx context.Context) (*block, error) {
...
 		}
+		checkHash = &checkBlock.PrevBlock
 	}
 }
 
-func (dcr *ExchangeWallet) isMainchainBlock(ctx context.Context, block *block) (bool, error) {
-	blockHeader, err := dcr.wallet.GetBlockHeaderVerbose(ctx, block.hash)
+// blockHeader returns the *BlockHeader for the specified block hash, and a
+// validity bool indicating true when the block has not been stake-invalidated.
+func (dcr *ExchangeWallet) blockHeader(ctx context.Context, blockHash *chainhash.Hash) (blockHeader *BlockHeader, valid, mainchain bool, err error) {
```
buck54321, line 3191: I could also add `Valid, Mainchain bool` fields to the `BlockHeader` and let the implementation decide how to fill them in.

chappjc: Hmm, yeah that might work out nicely.  Might also be worth documenting that it will never set valid = true where mainchain = false (if I'm reading the code right).

client/asset/dcr/externaltx.go
```diff
@@ -388,32 +366,34 @@ func (dcr *ExchangeWallet) isOutputSpent(ctx context.Context, output *outputSpen
...
 	bestBlock := dcr.cachedBestBlock()
 	for {
-		blockFilter, err := dcr.getBlockFilterV2(ctx, iHash)
+		bingo, err := dcr.wallet.MatchAnyScript(ctx, iHash, [][]byte{outputPkScript})
 		if err != nil {
-			return nil, lastScannedHash, err
+			if lastScannedHash != nil {
+				return nil, lastScannedHash, lastScannedHeight, err
+			}
+			return nil, lastScannedHash, lastScannedHeight, err
```
chappjc, line 382: Both of these return lines are identical.  I'm not clear on the intent.

chappjc (approved):
LGTM.

@itswisdomagain Please circle back on your review when you get a chance.

client/asset/dcr/wallet.go
```diff
@@ -92,54 +88,65 @@ type Wallet interface {
...
-	UnlockAccount(ctx context.Context, account, passphrase string) error
+	// MatchAnyScript looks for any of the provided scripts in the block specified.
+	MatchAnyScript(ctx context.Context, blockHash *chainhash.Hash, scripts [][]byte) (bool, error)
+	// AccountUnlocked returns true if the account is unlocked.
+	AccountUnlocked(ctx context.Context, acctName string) (bool, error)
+	// LockAccount locks the account.
+	LockAccount(ctx context.Context, acctName string) error
+	// UnlockAccount unlocks the Wallet. ExchangeWallet does not differentiate
+	// account locking vs wallet locking, but the underlying implementation may
+	// choose to unlock only the account.
```
itswisdomagain, line 128: Going through this again, the doc reads somehow. `UnlockAccount unlocks the Wallet`. Should that be `UnlockWallet`? I also notice that the `Un/lockWallet` methods are removed, so implementations have to understand that this single method may do either wallet unlocking or single account unlocking, whichever is necessary to provide access to the account priv keys. I think this should go in the doc for this and related methods.

client/asset/dcr/dcr.go
itswisdomagain: The other implementation I'm aware of only accesses the `Settings` map.

itswisdomagain (changes requested):
My previous review comments are addressed, except maybe 1. Just these nits from me for now.

client/asset/dcr/dcr.go
```diff
@@ -2932,7 +2835,7 @@ func (dcr *ExchangeWallet) signTxAndAddChange(baseTx *wire.MsgTx, feeRate uint64
 	// Sign the transaction to get an initial size estimate and calculate
 	// whether a change output would be dust.
 	sigCycles := 1
-	msgTx, err := dcr.signTx(baseTx)
+	msgTx, err := dcr.wallet.SignRawTransaction(dcr.ctx, baseTx)
```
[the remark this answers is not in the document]
chappjc: I don't understand the concerns.  We don't modify `baseTx`.  Also, we pass pointers to objects all the time without feeling the need to deep copy them.

In the implementation we have code like

```go
func (w *rpcWallet) SignRawTransaction(ctx context.Context, inTx *wire.MsgTx) (*wire.MsgTx, error) {
	baseTx := inTx.Copy()
	txHex, err := msgTxToHex(baseTx)

<code that does NOT modifiy inTx OR baseTx... why did we just copy it???>

}
```

I don't get the above any more than doing a `baseTx.Copy()` in the caller.

The *interface* docs have `// SignRawTransaction should not mutate the input transaction.`.
For completeness implementation docs could say `// SignRawTransaction does not mutate the input transaction.` but even that seems unnecessary.

client/asset/dcr/wallet.go
chappjc: I would even suggest just `Unlock`, but as I've said before, I would avoid that at all costs because it is ungrepable.  @itswisdomagain is more deep in multi-account support, so whatever he thinks I guess.  Or you could adjust this later as you see fit in your PR Wisdom.

client/asset/dcr/dcr.go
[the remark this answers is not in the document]
buck54321: Sort of a continuation of https://github.com/decred/dcrdex/pull/1510#discussion_r826947765

I opted to drop `Account` from method names at first, but reversed course to see how @itswisdomagain's privacy work would pan out.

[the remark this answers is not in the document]
chappjc: > I was on the fence on how to handle this. I went with with [@chappjc's suggestion](https://github.com/decred/dcrdex/pull/1510#discussion_r844196085)

I was really just suggesting I'd be more than happy with a doc on the interface but no Copy anywhere at all.  The implementation knows it doesn't modify it, so the copy it completely pointless.  I'd 100% propose removing it in a follow-up.

client/asset/dcr/rpcwallet.go
chappjc: As a partial resolution, at the time of assertion we could add the SpvMode condition:

```go
mempooler, is := dcr.wallet.(Mempooler)
if !is || dcr.wallet.SpvMode() {
	return

PR #1511, opened 2022-03-11
client/core: Make errors user oriented

### LINKED ISSUE #1488: Make ui error messages more oriented towards users
The error messages on the UI are just the error strings that are returned from functions in the code. These should be logged, but on the UI we should display just "Incorrect Password" instead of "error unlocking btc wallet: ReconfigureWallet password error: outer key deserialization error: incorrect password".

ukane-philemon:
This difff makes errors more user-oriented.
1. It modifies `core.Error` type to support error wrapping and unwrapping.
2. It adds an `Unwarp()` method to `core.Error` type.
3. Adds an ` UnwrapErr(err error) error` function that unwraps until it finds a non-wrapped error.
4. Adds tests for `core.Error` type.
5. Modify error strings in `webserver_test.go`.
6. Adds `%w` directive to errors constructed with `core` `newError(...)`.

Moving forward it would be nice to wrap errors that might end up being shown to users via client, detailed errors messages are logged to the terminal while errors shown via client are unwrapped. 

**Update**: Errors created using `core` `newError(...)` should be wrapped using the `%w` directive when necessary.
 
Now:
errorunwrap4
errorunwarp1
newdexerrafter

Closes #1488.

### DIFF
--- a/client/core/errors.go
+++ b/client/core/errors.go
@@ -49,34 +49,56 @@ const (
 	activeOrdersErr
 )
 
-// Error is an error message and an error code.
+// Error is an error message, an error code and a wrapped error.
 type Error struct {
-	s    string
-	code int
+	desc    string
+	code    int
+	wrapped error
 }
 
 // Error returns the error string. Satisfies the error interface.
 func (e *Error) Error() string {
-	return e.s
+	return e.desc
 }
 
+// Code returns the error code.
 func (e *Error) Code() *int {
 	return &e.code
 }
 
+// Unwrap returns the underlying wrapped error.
+func (e *Error) Unwrap() error {
+	return e.wrapped
+}
+
 // newError is a constructor for a new Error.
 func newError(code int, s string, a ...interface{}) error {
-	return &Error{
-		s:    fmt.Sprintf(s, a...),
-		code: code,
+	var err *Error
+	for _, v := range a {
+		e, ok := v.(error)
+		if ok {
+			err = &Error{
+				desc:    fmt.Sprintf(s, a...),
+				code:    code,
+				wrapped: e,
+			}
+			return err
+		}
 	}
+	err = &Error{
+		desc:    fmt.Sprintf(s, a...),
+		code:    code,
+		wrapped: fmt.Errorf(s, a...),
+	}
+	return err
 }
 
 // codedError converts the error to an Error with the specified code.
 func codedError(code int, err error) error {
 	return &Error{
-		s:    err.Error(),
-		code: code,
+		desc:    err.Error(),
+		code:    code,
+		wrapped: err,
 	}
 }
 
@@ -84,4 +106,19 @@ func codedError(code int, err error) error {
 func errorHasCode(err error, code int) bool {
 	var e *Error
 	return errors.As(err, &e) && e.code == code
+}
+
+// Unwrap returns the result of calling the Unwrap method on err, if err's
+// type contains an Unwrap method returning error.
+// Otherwise, Unwrap returns err.
+func Unwrap(err error) error {
+	for {
+		u, ok := err.(interface {
+			Unwrap() error
+		})
+		if !ok {
+			return err
+		}
+		err = u.Unwrap()
+	}
 }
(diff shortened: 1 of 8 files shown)


martonp (approved):
This looks good and definitely much better than it was before.
I agree we should wrap the messages more, and maybe we could add translations to them as well.

ukane-philemon:
> and maybe we could add translations to them as well.

They're know frequent errors that can possibly be translated, @martonp any suggestion on how to?

PS: are dex translated const done by a language native speaker?

client/webserver/api.go
```diff
@@ -853,9 +853,11 @@ func (s *WebServer) writeAPIError(w http.ResponseWriter, err error) {
 	if errors.As(err, &cErr) {
 		code = cErr.Code()
 	}
+
+	rawErr := core.Unwrap(err)
```
chappjc, line 857: In this context, you can see why `Unwrap` is not a great name for a `core` package function.  If there were an errors package it could work.

client/core/errors.go
```diff
@@ -49,34 +49,56 @@ const (
...
+				code:    code,
+				wrapped: e,
+			}
+			return err
+		}
+	}
+	err = &Error{
+		desc:    fmt.Sprintf(s, a...),
+		code:    code,
+		wrapped: fmt.Errorf(s, a...),
```
chappjc, line 91: I don't see why `Error` cannot be this:

```go
type Error struct {
	code int
	err  error
}
```

With `newError` as this:

```go
func newError(code int, s string, a ...interface{}) error {
	return &Error{
		code: code,
		err:  fmt.Errorf(s, a...), // s may contain a %w verb to wrap an error
	}
}
```

and `codedError` as this:

```go
func codedError(code int, err error) error {
	return &Error{
		code: code,
		err:  err,
	}
}
```

and these methods:

```go
func (e *Error) Error() string {
	return e.err.Error()
}
func (e *Error) Unwrap() error {
	return e.err
}
func (e *Error) Code() *int {
	return &e.code
}
```

chappjc (changes requested):

chappjc:
> errorunwrap4

Let's change this "zero withdraw" message.  Seeing "btc zero withdraw" is confusing.  Should be "zero BTC to withdraw" or "BTC balance is zero"

client/core/errors.go
```diff
@@ -85,3 +107,18 @@ func errorHasCode(err error, code int) bool {
...
 }
+
+// Unwrap returns the result of calling the Unwrap method on err, if err's
+// type contains an Unwrap method returning error.
+// Otherwise, Unwrap returns err.
+func Unwrap(err error) error {
+	for {
+		u, ok := err.(interface {
+			Unwrap() error
+		})
```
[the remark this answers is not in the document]
ukane-philemon: @chappjc thanks for looking into this.

martonp:
There's text all over the app including the errors, wallet configrations, and others that currently cannot be translated since they are not part of the templates. All of these texts should be in a central location where translations can be provided.

> PS: are dex translated const done by a language native speaker?

Yes translations are done by native speakers.

client/core/errors.go
ukane-philemon: Thanks for the sugestions, all refactoring addressed in latest commit [43ab36c](https://github.com/decred/dcrdex/pull/1511/commits/43ab36c4e82ed1baf1fa700dca613142664b58be)

ukane-philemon:
> Should then be "cannot withdraw zero BTC" or "withdraw amount must not be zero". The last one seems good since the symbol (BTC) isn't needed in this context IMO.

Now:
zerowithdrawerr
dcrwithdrawerr

chappjc:
So for translations, let's follow up in different work, but it would require the following:

- exported client/asset error types
- client/webserver/locates defines the translations 
- client/webserver recognizes these error types with errors.Is and gets the translation for the current locale for the user facing error

chappjc:
On second though, I think the translations in client/core would make more sense than in client/webserver since the ones in webserver are just for the html templates.
That would actually mean that these new `core.Error` changes and the unwrapping in `(*WebServer).writeAPIError` would become mostly irrelevant.
We can think it over more...

buck54321 (approved):

PR #1512, opened 2022-03-11
ui: line break long error messages

### LINKED ISSUE #1145: ui: Errors can extend beyond module.
Some errors can be extra long and not fit in the intended space. It's not really a problem but does seem unintended, although I don't have a suggestion to fix it. For example:

errtolong

ukane-philemon:
Closes #1145 .

### DIFF
--- a/client/webserver/site/src/css/main.scss
+++ b/client/webserver/site/src/css/main.scss
@@ -389,7 +389,6 @@ div.popup-notes {
 
 .errcolor {
   color: #e61c00;
-  word-break: break-all;
 }
 
 hr.dashed {
--- a/client/webserver/site/src/css/main.scss
+++ b/client/webserver/site/src/css/main.scss
@@ -389,6 +389,7 @@ div.popup-notes {
 
 .errcolor {
   color: #e61c00;
+  word-break: break-all;
 }
 
 hr.dashed {
--- a/client/webserver/site/src/html/forms.tmpl
+++ b/client/webserver/site/src/html/forms.tmpl
@@ -33,7 +33,7 @@
     <div data-tmpl="defaultSettings" class="dynamicopts"></div>
   </div>
 </div>
-<div data-tmpl="errMsg" class="fs15 pt-3 text-center d-hide errcolor"></div>
+<div data-tmpl="errMsg" class="fs15 pt-3 text-center d-hide errcolor text-break"></div>
 {{end}}
 
 {{define "newWalletForm"}}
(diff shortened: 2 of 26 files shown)


client/webserver/site/src/css/main.scss
```diff
@@ -389,6 +389,7 @@ div.popup-notes {
 
 .errcolor {
   color: #e61c00;
+  word-break: break-all;
```
chappjc, line 392: Let's see a couple examples of this please.  `break-all` breaks words very readily, perhaps too readily; on the other hand, `word-break: break-word`, or `word-break: normal` with `overflow-wrap: anywhere`, would only break the words that cannot fit on one line.

client/webserver/site/src/css/main.scss
```diff
@@ -389,6 +389,7 @@ div.popup-notes {
 
 .errcolor {
```
chappjc, line 390: Little odd to have the `word-break` style set in a color-related class.  I noticed there's a bootstrap class called `text-break` for this: https://getbootstrap.com/docs/5.0/utilities/text/#word-break
Probably we should make elements into the bootstrap `text-break` class where desired.  @buck54321 can decide though.

client/webserver/site/src/css/main.scss
ukane-philemon: > In contrast to `word-break: break-word` and `overflow-wrap: break-word`, `word-break: break-all` will create a break at the exact place where text would otherwise overflow its container. [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/word-breakl)

`word-break: break-word` is deprecated, using `overflow-wrap: anywhere` better?

chappjc: I say we let the bootstrap project decide. https://getbootstrap.com/docs/5.0/utilities/text/#word-break

client/webserver/site/src/css/main.scss
ukane-philemon: Yeah, figured out `.errcolor` is only applied on all errors. instead of adding the bootstrap `text-break` class to elements displaying errors.
> Probably we should make elements into the bootstrap text-break class where desired. I'm inclined to say we should use a bootstrap class where there is one. The bootstrap project can update their css as needed and we don't have to worry about it

This is a good reason to stick with bootstrap.

chappjc (approved):
LGTM, but leaving for @buck54321

JoeGruffins (approved):
Issue is solved:
image

buck54321 (approved):

PR #1519, opened 2022-03-15
update to Go 1.18

chappjc:
This updates for Go 1.18.

The supported Go versions are now 1.17 and 1.18.

CI is updated to new action versions and uses golangci-lint v1.45.0.

All the old `//+build` lines are removed now that all supported Go versions understand the new `//go:build` lines.

The go.mod now specifies `go 1.17`, and a `go mod tidy` is applied.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,22 +1,24 @@
 name: Build and Test
 on: [push, pull_request]
+permissions:
+  contents: read
 jobs:
   build-go:
     name: Go CI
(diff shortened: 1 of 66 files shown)


chappjc:
Working now, but golangci-lint is still sorting out go 1.18 issues: https://github.com/golangci/golangci-lint/issues/2649#issuecomment-1070528726

JoeGruffins (approved):
`grep -rn "+build"` will find three more usages in the `dex/networks` area

PR #1525, opened 2022-03-17
expand dcr wallet rpc cert

ukane-philemon:
This diff expands dcr wallet rpc cert.

### DIFF
--- a/client/asset/dcr/config.go
+++ b/client/asset/dcr/config.go
@@ -77,6 +77,8 @@ func loadConfig(settings map[string]string, network dex.Network) (*Config, *chai
 	}
 	if cfg.RPCCert == "" {
 		cfg.RPCCert = defaultRPCCert
+	} else {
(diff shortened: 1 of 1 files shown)


client/asset/dcr/config.go
```diff
@@ -78,6 +78,7 @@ func loadConfig(settings map[string]string, network dex.Network) (*Config, *chai
 	if cfg.RPCCert == "" {
 		cfg.RPCCert = defaultRPCCert
 	}
+	cfg.RPCCert = dex.CleanAndExpandPath(cfg.RPCCert)
```
JoeGruffins, line 81: Could be an else here, because you wouldn't need to try and clean the default path I don't think.

JoeGruffins (approved):

client/asset/dcr/config.go
ukane-philemon: Yes, wanting to keep things minimal.

buck54321 (approved):
I guess this is okay. Was there an issue for this?

PR #1528, opened 2022-03-17
start recognizing signatures of the hashed messages (sig fix stage 1)

chappjc:
This is stage 1 of the signature message truncation fix plan outlined https://github.com/decred/dcrdex/pull/1526.

In these commits, both client and server begin recognizing signatures of both the truncated messages or the message hash.  Neither side begins signing the message hash.  After this has been deployed to servers, the next stage (https://github.com/decred/dcrdex/pull/1529) has the client begin creating signatures of the hashed messages.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -6,6 +6,7 @@ package core
 import (
 	"bytes"
 	"context"
+	"crypto/sha256"
 	"encoding/binary"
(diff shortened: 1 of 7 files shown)


chappjc:
Working when rebased on release-v0.4

buck54321 (approved):

PR #1529, opened 2022-03-17
client: sign and verify hashed messages (sig fix stage 2)

chappjc:
<s>Rebased on https://github.com/decred/dcrdex/pull/1528</s>

This is stage 2 of the signature message truncation fix plan outlined in https://github.com/decred/dcrdex/pull/1526.

In these commits, client begins signing the hashed messages, but **continues to recognize the buggy signatures** from the server.  This is because the while the server understands the correct hashed message signatures [as of stage 1](https://github.com/decred/dcrdex/pull/1528), it is still sending the old signatures with its messages.

This work can be backported to a 0.4.3+ release if https://github.com/decred/dcrdex/pull/1528 has been deployed with 0.4.2.

In the next stage, the server begins sending the correct signatures.

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -2184,12 +2184,13 @@ func (btc *baseWallet) SignMessage(coin asset.Coin, msg dex.Bytes) (pubkeys, sig
 		return nil, nil, err
 	}
 	pk := privKey.PubKey()
-	sig, err := privKey.Sign(msg)
+	hash := chainhash.HashB(msg) // legacy servers will not accept this signature!
+	sig, err := privKey.Sign(hash)
 	if err != nil {
 		return nil, nil, err
 	}
 	pubkeys = append(pubkeys, pk.SerializeCompressed())
-	sigs = append(sigs, sig.Serialize())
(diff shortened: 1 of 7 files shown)


chappjc:
0.4 based branch at https://github.com/chappjc/dcrdex/tree/sig-fix-stage-2-v0.4

client/asset/btc/btc.go
```diff
@@ -2184,12 +2184,13 @@ func (btc *baseWallet) SignMessage(coin asset.Coin, msg dex.Bytes) (pubkeys, sig
 		return nil, nil, err
 	}
 	pk := privKey.PubKey()
-	sig, err := privKey.Sign(msg)
+	hash := sha256.Sum256(msg) // legacy servers will not accept this signature!
```
JoeGruffins, line 2187: btc test will use `chainhash.HashB`, just confirming you meant to use `sha256.Sum256` It is the same though https://github.com/btcsuite/btcd/blob/bf64c8bdbbbfc6db51312769742e1210b8ac8ed0/chaincfg/chainhash/hashfuncs.go#L10-L14

JoeGruffins (approved):

PR #1530, opened 2022-03-17
server: sign and verify hashed messages (sig fix stage 3)

chappjc:
This is stage 3 of the signature message truncation fix plan outlined in https://github.com/decred/dcrdex/pull/1526.

In these commits, server:
1. Begins signing the hashed messages
2. **Stops recognizing** the buggy signatures.

All clients must be running with the [stage 2 fix](https://github.com/decred/dcrdex/pull/1529) in v0.4.3 before this change can be deployed.

In the [next and final stage](https://github.com/decred/dcrdex/pull/1526), the client drops support for truncated message signatures from the server running at this point.

### DIFF
--- a/server/asset/btc/btc_test.go
+++ b/server/asset/btc/btc_test.go
@@ -524,7 +524,8 @@ func s256Auth(msg []byte) *testAuth {
 	if msg == nil {
 		msg = randomBytes(32)
 	}
-	sig, err := priv.Sign(msg)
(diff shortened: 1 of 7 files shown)


chappjc:
With stage 1 of the fix in v0.4.2 and evidently deployed, and stage 2 merged onto release-v0.4 for an upcoming v0.4.3, this stage of the fix can get on 0.5 assuming there's gonna be time before 0.5 is released.

PR #1532, opened 2022-03-18
server/wslink: Add SendRaw.

### LINKED ISSUE #1478: {dex/ws, server/market}: alternative send method for notifications
The `func (*WSLink) Send(msg *msgjson.Message) error` method JSON-encodes the passed `Message`. For notifications, that's not great, because the same notification is being sent to possibly many users, such as in `(*BookRouter).sendNote`.

https://github.com/decred/dcrdex/blob/0c75bd4ef78a0cde38eb558dd20b1ffe2a655683/server/market/bookrouter.go#L708-L709

We should offer a `SendRaw([]byte) error` alternative to avoid repeated encoding.

JoeGruffins:
In order to avoid marshalling the same message multiple times when
    sending to multiple clients, marshal once and send the raw bytes.

closes #1478

### DIFF
--- a/dex/ws/wslink.go
+++ b/dex/ws/wslink.go
@@ -105,6 +105,16 @@ func (c *WSLink) Send(msg *msgjson.Message) error {
 	return c.send(msg, nil)
 }
 
+// SendRaw sends the passed bytes to the websocket peer. The actual writing of
+// the message on the peer's link occurs asynchronously. As such, a nil error
+// only indicates that the link is believed to be up.
+func (c *WSLink) SendRaw(b []byte) error {
+	if c.Off() {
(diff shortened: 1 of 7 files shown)


chappjc (changes requested):
Good stuff.  Just one more spot for `SendRaw` is in `server/comms.(*Server).Broadcast`: https://github.com/decred/dcrdex/blob/24b2038b75158e224da3e0d3f7e1a4b0f7df47c9/server/comms/server.go#L632-L633

I suggest doing the marshalling in that method before locking the mutex.

buck54321 (approved):

chappjc (approved):

PR #1543, opened 2022-03-23
app: ignore zero-quantity book_order updates

buck54321:
I've been chasing down the residual orders in the front-end order book and depth chart. The orders that we see are zero-quantity orders that appear to be related to orders that are booked and unbooked on the same match cycle. In that case, we receive a `book_order` update followed by an `unbook_order` update. The `book_order` update contains a `BookOrderNote` with a zero-quantity order, because by the time the `order.LimitOrder` gets translated by `BookRouter`, the `Remaining` is already zero.

I still haven't figured this out. I've put feelers around to try to catch a zero-quantity `book_order` update with no subsequent `unbook_order` update, but I haven't seen one yet, even though I still see residual orders. 

Oddly, these leftover orders only appear on the buy side. 

This PR is here just to display a crude fix and for discussion. I've been banging my head against the wall on this one, and I need to switch gears.

The crude fix is to simply ignore zero-quantity orders in `OrderBook.add`. We already ignore unmatched `unbook_order` updates.

### DIFF
--- a/client/webserver/site/src/js/orderbook.ts
+++ b/client/webserver/site/src/js/orderbook.ts
@@ -23,7 +23,10 @@ export default class OrderBook {
 
   /* add adds an order to the order book. */
   add (ord: MiniOrder) {
-    if (ord.qtyAtomic === 0) return
+    if (ord.qtyAtomic === 0) {
+      window.log('zeroqty', 'zero quantity order encountered', ord)
+      return
+    }
     const side = ord.sell ? this.sells : this.buys
     side.splice(findIdx(side, ord.rate, !ord.sell), 0, ord)
   }
--- a/client/webserver/site/src/js/orderbook.ts
+++ b/client/webserver/site/src/js/orderbook.ts
@@ -24,6 +24,17 @@ export default class OrderBook {
   /* add adds an order to the order book. */
   add (ord: MiniOrder) {
     if (ord.qtyAtomic === 0) {
+      // TODO: Somebody, for the love of god, figure out why the hell this helps
+      // with the ghost orders problem. As far as I know, this order is a booked
+      // order that had more than one match in an epoch and completely filled.
+      // Because the first match didn't exhaust the order, there would be a
+      // 'update_remaining' notification scheduled for the order. But by the
+      // time OrderRouter generates the notification long after matching, the
+      // order has zero qty left to fill. It's all good though, kinda, because
+      // the notification is quickly followed with an 'unbook_order'
+      // notification. I have tried my damnedest to catch an update_remaining
+      // note without an accompanying unbook_order note, and have thus failed.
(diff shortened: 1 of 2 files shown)


chappjc:
I swear @JoeGruffins just commented about cancel orders being a related issue, but I cannot find it.

Also might be related is that I often see 0 qrt market buy orders in the table and I have a feeling they are not actually trade orders, but cancels that somehow got in.  Something about the zero-values for an order and the flags for force, sell, etc.?

chappjc:
In testing the expiry PRs I put up, I managed to get a ghost buy order on the depth chart.  I couldn't tell if it appeared after the match or when the revoke_order happened.  Might try it again.

buck54321:
I've come back to this a couple of times to try to figure why this helps, and I still have no idea. But it does seem to prevent the issue for me.

chappjc:
I've been saving the loadbot go.sum update for https://github.com/decred/dcrdex/pull/1656 if that's alright with you

client/webserver/site/src/js/orderbook.ts
```diff
@@ -23,6 +23,7 @@ export default class OrderBook {
 
   /* add adds an order to the order book. */
   add (ord: MiniOrder) {
+    if (ord.qtyAtomic === 0) return
```
chappjc, line 26: I say let's go with it, but in the name of figuring this out, how bout a console log with info about the order to help figure out why it's here?

buck54321: There are a lot of these.

chappjc: You mean it happens frequently or there are a lot of places in the code?

buck54321: They are orders that are booked, match partially (generating a `update_remaining`), then finish matching in the same epoch. By the time the `update_remaining` notification is generated by `BookRouter`, the order's `Remaining` value is `0`. It's of little consequence in most places because an `unbook_order` notification is sent immediately after. I've put a little cache with a timer in place to make sure that the `unbook_order` comes through for every zero-qty order within a second or so, and they all do. But still for some reason, if not ignored, a zero-qty order will end up persisting on the book somehow, screwing up the chart.

chappjc: I was trying to repro with something very similar.  Two lot order on book, matches with 1 lot taker, hitting `handleUpdateRemainingRoute` (which could use an `as RemainderUpdate` I think), and following with a cancel *in the same epoch* hitting `handleUnbookOrderRoute`, so very similar to what you described.  Still couldn't make this `ord.qtyAtomic` condition hit in this `add` function.  :man_shrugging:

buck54321:
You can watch loadbot trade on compound and enable the logger in the console.

```Javascript
enableLogger('zeroqty', true)
```

chappjc (approved):

PR #1545, opened 2022-03-24
client: bump node deps due to security alert.

dajohi:


chappjc (approved):
OK.

Just my usual note to the ether that these packages don't become part of the bundle and aren't used at runtime, only for the build system.  Still good to fix though.

Issue #1546, opened 2022-03-24
client/asset/btc: untrusted 0-conf funds counted as available in balance


chappjc:
https://github.com/decred/dcrdex/blob/fb76b59006d744e149a48ecc2218eaead4920819/client/asset/btc/spv.go#L596-L599

Which differs from funding code that identifies "safe" utxos that can be spent.  I have a fix in the works.

PR #1547, opened 2022-03-24
client/asset/btc: balance should consider untrusted UTXOs

chappjc:
Resolves https://github.com/decred/dcrdex/issues/1546

### DIFF
--- a/client/asset/btc/spv.go
+++ b/client/asset/btc/spv.go
@@ -586,23 +586,70 @@ func (w *spvWallet) syncStatus() (*syncStatus, error) {
 	}, nil
 }
 
-// Balances retrieves a wallet's balance details.
+// ownsInputs determines if we own the inputs of the tx.
+func (w *spvWallet) ownsInputs(txid string) bool {
+	txHash, err := chainhash.NewHashFromStr(txid)
+	if err != nil {
+		w.log.Warnf("Error decoding txid %q: %v", txid, err)
+		return false
+	}
+	txDetails, err := w.wallet.walletTransaction(txHash)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


buck54321:
Doesn't this mean that I can no longer fund my wallet and begin trading right away?

buck54321:
Oh. This is only SPV. In that case, the wallet already doesn't know about outside outputs until they already have confs = 1.

chappjc:
That's how it has always worked with bitcoind as a backend.

chappjc:
> Oh. This is only SPV. In that case, the wallet already doesn't know about outside outputs until they already have confs = 1.

It knows about a redeem.  That's the main issue.  You complete a swap for BTC, then try to withdraw it before the redeem is confirmed.  The page says you have x btc available, but it's not available to the coin selection code

client/asset/btc/spv.go
```diff
@@ -611,44 +658,20 @@ func (w *spvWallet) listUnspent() ([]*ListUnspentResult, error) {
...
-					if !errors.Is(err, wallet.ErrNotMine) {
-						w.log.Warnf("FetchInputInfo error: %v", err)
-					}
-					safe = false
-					break
-				}
-			}
-		}
+		// by seeing if we control the inputs of its transaction. TODO: Keep a
+		// cache of our redemption outputs and allow those as safe inputs.
```
chappjc, line 662: I meant to change this comment.  The TODO here is debatable.  The contract output being spent in the redeem is arguably untrusted.  The other party might still have a claim on it via redeem.  It all depends on network conditions and the lock time on the contract.  I think I'd rather play it safe with these.

JoeGruffins (approved):

buck54321 (approved):

PR #1548, opened 2022-03-24
client/core: consider contract expiry before acting

chappjc:
Complementary to https://github.com/decred/dcrdex/pull/1541, this includes client-side checks.

In particular, `(*trackedTrade).counterPartyConfirms` checks if the contract is expired and returns an `expired bool`.  The callers (`isSwappable` and `isRedeemable`) consider the expired status in the same way they already consider spent status -- self-revoke and return false to the spendable/redeemable check.  Specifically, this happens _only_ in the `order.TakerSwapCast` && `order.Maker` condition of `isRedeemable`, and in the `order.MakerSwapCast` && `order.Taker` condition of `isSwappable`.

Note that in the first scenario above of the maker deciding if the taker's contract is redeemable, it is quite important not to redeem an expired contract.  This is true for either actor since the contract could also be refunded at any time, but it's absolutely critical for maker since they are revealing the secret key to the taker.  That must not happen if the taker's contract is expired.

Another check added is for taker at `MakerSwapCast` when checking swappability, where it now also ensures that the taker's own contract that it _would publish_ would not have a lock time that is already in the past.

To ease readability, some order status `if` statements are converted to `switch` statements.  Also, an initial order status filter is added to `isSwappable` and `isRedeemable` for efficiency.

Finally, an minor bug with `wg.Done()` being called not at the end of a defer stack is fixed.

### DIFF
--- a/client/core/trade.go
+++ b/client/core/trade.go
@@ -844,7 +844,7 @@ func (t *trackedTrade) processCancelMatch(msgMatch *msgjson.Match) error {
 //
 // This method accesses match fields and MUST be called with the trackedTrade
 // mutex lock held for reads.
-func (t *trackedTrade) counterPartyConfirms(ctx context.Context, match *matchTracker) (have, needed uint32, changed, spent bool) {
+func (t *trackedTrade) counterPartyConfirms(ctx context.Context, match *matchTracker) (have, needed uint32, changed, spent, expired bool) {
 	// Counter-party's swap is the "to" asset.
 	needed = t.wallets.toAsset.SwapConf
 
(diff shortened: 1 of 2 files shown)


JoeGruffins (approved):

client/core/trade.go
```diff
@@ -1047,32 +1062,47 @@ func (t *trackedTrade) isSwappable(ctx context.Context, match *matchTracker) boo
...
 		return false
 	}
 
-	if match.Status == order.MakerSwapCast {
+	switch match.Status {
+	case order.NewlyMatched:
+		return match.Side == order.Maker
+	case order.MakerSwapCast:
 		// Get the confirmation count on the maker's coin.
 		if match.Side == order.Taker {
```
buck54321, line 1075: While we're improving here, maybe  a 

```Go
    return false
}
```

So we can lose a level of indentation?

client/core/trade.go
```diff
@@ -1104,11 +1132,25 @@ func (t *trackedTrade) isSwappable(ctx context.Context, match *matchTracker) boo
...
+		return false // all swaps not yet sent
+	}
+
 	if match.swapErr != nil || len(match.MetaData.Proof.RefundCoin) != 0 || match.tickGovernor != nil {
 		t.dc.log.Tracef("Match %s not redeemable: swapErr = %v, RefundCoin = %v, metered = %t",
 			match, match.swapErr, match.MetaData.Proof.RefundCoin, match.tickGovernor != nil)
 		return false
 	}
+	// NOTE: taker must be able to redeem when revoked! Arguably, maker should
+	// not either, and we can check here, but as long as taker contract's is not
```
buck54321, line 1147: ~~either~~

buck54321 (approved):
Some intermittent failures in core harness tests, specifically `TestNoMakerRedeem`, but I'm not convinced it has anything to do with this PR.

chappjc:
Holding off merging this until https://github.com/decred/dcrdex/pull/1541 is in since testing that without this client change is potentially more informative.  On the other hand, stacking that PR on top of this one is potentially helpful as well.

PR #1549, opened 2022-03-26
server/asset: Check if txindex is enabled.

### LINKED ISSUE #1535: server/dcr: Preemptively check for --txindex.
Recently, I set up a testnet server and forgot to run dcrd with the `--txindex` flag. The server asset started up fine and there was no problem until a user tried to pay the registration fee. This caused the error: `[WRN] AUTH: Unexpected error checking fee coin confirmations: -32603: The transaction index must be enabled to query the blockchain (specify --txindex)`

If possible, it would be better to error at start-up when transaction indexing is not enabled to avoid problems after the server has already been started.

The `getinfo` rpc command does provide a way to check:
```
$ dcrctl getinfo
{
  "version": 1080000,
  "protocolversion": 8,
  "blocks": 889610,
  "timeoffset": -1,
  "connections": 8,
  "proxy": "",
  "difficulty": 31.33877036,
  "testnet": true,
  "relayfee": 0.0001,
  "errors": "",
  "addrindex": false,
  "txindex": true
}
```

ukane-philemon:
This diff:
1. checks if  `-txindex` is enabled for bitcoind 
2. checks if  `--txindex` is enabled for dcrd.
Closes #1535

### DIFF
--- a/server/asset/btc/btc.go
+++ b/server/asset/btc/btc.go
@@ -254,6 +254,16 @@ func (btc *Backend) Connect(ctx context.Context) (*sync.WaitGroup, error) {
 		}
 	}
 
+	txindex, err := btc.node.CheckTxIndex()
+	if err != nil {
+		btc.shutdown()
+		return nil, fmt.Errorf("bitcoind getindexinfo check failed: %w", err)
+	}
+	if !txindex {
+		btc.shutdown()
+		return nil, errors.New("bitcoind transaction index is not enabled(specify -txindex)")
+	}
+
 	if _, err = btc.estimateFee(btc.node); err != nil {
 		btc.log.Warnf("%s backend started without fee estimation available: %v", btc.name, err)
 	}
--- a/server/asset/btc/btc.go
+++ b/server/asset/btc/btc.go
@@ -254,7 +254,7 @@ func (btc *Backend) Connect(ctx context.Context) (*sync.WaitGroup, error) {
 		}
 	}
 
-	if txindex, err := btc.node.CheckTxIndex(); err != nil {
+	if txindex, err := btc.node.checkTxIndex(); err != nil {
 		if !isMethodNotFoundErr(err) {
 			btc.shutdown()
 			return nil, fmt.Errorf("%s getindexinfo check failed: %w", btc.name, err)
--- a/server/asset/btc/btc.go
+++ b/server/asset/btc/btc.go
@@ -254,14 +254,19 @@ func (btc *Backend) Connect(ctx context.Context) (*sync.WaitGroup, error) {
 		}
 	}
 
-	txindex, err := btc.node.CheckTxIndex()
-	if err != nil {
+	if txindex, err := btc.node.CheckTxIndex(); err != nil {
+		if !isMethodNotFoundErr(err) {
+			btc.shutdown()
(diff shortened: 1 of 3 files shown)


server/asset/btc/rpcclient.go
```diff
@@ -22,6 +22,7 @@ const (
 	methodGetTxOut          = "gettxout"
 	methodGetRawTransaction = "getrawtransaction"
 	methodGetBlock          = "getblock"
+	methodGetIndexInfo      = "getindexinfo"
```
chappjc, line 25: Since this method was [introduced in Bitcoin Core v0.21](https://bitcoincore.org/en/releases/0.21.0/#new-rpcs), that officially breaks support for v0.20, which people still use.  It also is a problem for other forks of Bitcoin Core (e.g. Litecoin 0.18) that don't have this either.

This could be resolved in different ways.  Simplest is to just log a warning, which may or may not be seen by the operator.  We could detect a "method not found" error and ignore the error.  More complete would be to check the version like we do on client side.

chappjc (changes requested):

JoeGruffins (approved):

server/asset/btc/btc.go
```diff
@@ -254,6 +254,25 @@ func (btc *Backend) Connect(ctx context.Context) (*sync.WaitGroup, error) {
...
+		}
+		// Ignore and log err if getindexinfo method is not found.
+		// getindexinfo method is not currently supported by
+		// pre 0.21 versions of bitcoind, and some forks of
+		// bitcoin core (litecoin).
+		txindex = true
+		btc.log.Warnf("The getindexinfo RPC is unavailable for %s. Please ensure txindex is enabled in the node config.", btc.name)
+	}
+
+	if !txindex {
```
chappjc, line 271: Combine into an `else if` and remove the `txindex = true` in the `err != nil` branch above.

In fact, I would suggest limiting scope of the `txindex` variable like

```go
if txindex, err := btc.node.CheckTxIndex(); err != nil {
```

chappjc (changes requested):
Looks good. One last logic clean up please

ukane-philemon:
@chappjc thanks for the review, all changes applied.

server/asset/btc/rpcclient.go
```diff
@@ -75,6 +76,25 @@ func (rc *RPCClient) GetBlockChainInfo() (*GetBlockchainInfoResult, error) {
 	return chainInfo, nil
 }
 
+// TxIndexResult models the data returned from the getindexinfo command
+// for txindex.
+// TxIndexResult.Txindex is nil if the returned data is an empty json object.
+type TxIndexResult struct {
```
buck54321, line 82: Doesn't need to be exported.

server/asset/btc/rpcclient.go
```diff
@@ -75,6 +76,25 @@ func (rc *RPCClient) GetBlockChainInfo() (*GetBlockchainInfoResult, error) {
...
+	TxIndex *struct{} `json:"txindex"`
+}
+
+// CheckTxIndex checks if bitcoind transaction index is enabled.
+func (rc *RPCClient) CheckTxIndex() (bool, error) {
```
buck54321, line 87: Doesn't need to be exported.

buck54321 (approved):

PR #1550, opened 2022-03-29
client/eth: Allow harness testing on testnet.

JoeGruffins:
If testnet is set to true, create a new folder and wallets in the user's
    home directory and change set up to create testnet nodes. Testing is
    done pointing at testnet and the testnet contract.

    Tests currently often fail because a transaction randomly cannot be
    found to retrieve its receipt. This should not be an issue for swaps as
    we only check for txn in tests to verify gas used. It may, however, have
    implications for checking RegFeeConfirmations, as that similarly must
    find a tx.

Even waiting and trying many times to find some tx, they just seem to be randomly not existing. Maybe light nodes just can't be expected to find txn? See https://github.com/ethereum/go-ethereum/issues/24101

We should verify if this is normal or maybe we are doing something wrong. I'm guessing it's a problem on testnet and not so much on simnet because of the volume of transactions. Purely a guess at this point though.

Good news is that there doesn't seem to be any unexpected problems with the testnet contract.

### DIFF
--- a/client/asset/eth/nodeclient_harness_test.go
+++ b/client/asset/eth/nodeclient_harness_test.go
@@ -1,10 +1,7 @@
 //go:build harness && lgpl
 
-// This test requires that the testnet harness be running and the unix socket
-// be located at $HOME/dextest/eth/gamma/node/geth.ipc
-//
-// These tests are expected to be run in descending order as some depend on the
-// tests before. They cannot be run in parallel.
+// This test requires that the simnet harness be running. Some tests will
+// alternatively work on testnet.
 //
 // NOTE: These test reuse a light node that lives in the dextest folders.
 // However, when recreating the test database for every test, the nonce used
@@ -15,9 +12,13 @@
 // be a problem in the future if a user restores from seed. Punting on this
 // particular problem for now.
 //
-// TODO: Running these tests many times eventually results in all transactions
-// returning "unexpected error for test ok: exceeds block gas limit". Find out
-// why that is.
+// TODO: Running these tests many times on simnet eventually results in all
+// transactions returning "unexpected error for test ok: exceeds block gas
+// limit". Find out why that is.
+//
+// TODO: Running tests on the goerli testnet, specifically the contract tests,
+// will randomly fail because a transaction receipt cannot be retrieved. This
+// may be inevitable because we are using a light node. Verify this.
 
 package eth
 
@@ -33,7 +34,9 @@ import (
 	"math/big"
 	"os"
 	"os/exec"
+	"os/signal"
 	"path/filepath"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


client/asset/eth/nodeclient_harness_test.go
```diff
@@ -251,19 +393,41 @@ func setupWallet(walletDir, seed, listenAddress string) error {
...
+	}
 	for i := 0; ; i++ {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
 		prog := cl.syncProgress()
-		if prog.CurrentBlock >= prog.HighestBlock {
-			return nil
+		if testnet {
+			syncing := prog.CurrentBlock > prog.HighestBlock || prog.CurrentBlock == 0
```
chappjc, line 412: I'm confused by `prog.CurrentBlock > prog.HighestBlock` here.  For simnet we declare synced (not syncing) as `prog.CurrentBlock >= prog.HighestBlock && prog.CurrentBlock != 0`.
Should this here be a `<` to indicate styncing?

client/asset/eth/nodeclient_harness_test.go
```diff
@@ -56,22 +59,36 @@ import (
...
-	pw        = "bee75192465cef9f8ab1198093dfed594e93ae810ccbbf2b3b12e1771cc6cb19"
+	alphaNode         = "enode://897c84f6e4f18195413c1d02927e6a4093f5e7574b52bdec6f20844c4f1f6dd3f16036a9e600bd8681ab50fd8dd144df4a6ba9dd8722bb578a86aaa8222c964f@127.0.0.1:30304"
+	alphaAddr         = "18d65fb8d60c1199bb1ad381be47aa692b482605"
+	pw                = "bee75192465cef9f8ab1198093dfed594e93ae810ccbbf2b3b12e1771cc6cb19"
+	maxFeeRate uint64 = 200 // gwei per gas
+	// testnet can be set to true to perform tests on the goerli testnet.
+	// May need some setup including sending testnet coins to the addresses
+	// and a lengthy sync. Wallet addresses are the same as simnet. All
+	// wait and lock times are multiplied by testnetSecPerBlock. Tests may
+	// need to be run with a high --timeout=2h for the initial sync.
```
chappjc, line 70: So after the first time you run this harness, does the node data persist (and not get wiped) so another harness run can use it instead of syncing from scratch?

chappjc: Also, it would be swell if we could specify testnet bootnodes in case you have a local geth running to save internet bandwidth use.

chappjc (commented):
Cool.  Just a few comments to start.  Not tested net.

client/asset/eth/nodeclient_harness_test.go
JoeGruffins: > So after the first time you run this harness, does the node data persist (and not get wiped) so another harness run can use it instead of syncing from scratch?

This does not use the harness at all. So yes, the folder where these nodes live persists until you delete it manually.

> Also, it would be swell if we could specify testnet bootnodes

Will add this ability.

client/asset/eth/nodeclient_harness_test.go
JoeGruffins: I'm also a little confused here. I beleive `HighestBlock` is our local synced block, and `CurrentBlock` is the best block from the perspective of a peer. CurrentBlock always starts at zero I believe, when we have no peers. This is correct for syncing I think, after a bit of observing sure seems that way. If current block has reached highest block, we are synced and no longer syncing. I keep looking at SyncStatus thinking somethings wrong but I don't want to touch it in this pr.

Looking again, the simnet version should just be `==` I think as we dont want to worry about the header being old there.

chappjc: > I beleive HighestBlock is our local synced block, and CurrentBlock is the best block from the perspective of a peer

Pretty sure it's the opposite.  https://ethereum.stackexchange.com/a/424/90199

JoeGruffins: With `fmt.Printf("client: %d\ncurrent: %d\nhighest: %d\nstarting: %d\n", clientNum, prog.CurrentBlock, prog.HighestBlock, prog.StartingBlock)`

See:
```
Testnet nodes starting sync, this may take a while...
client: 2
current: 6619110
highest: 0
starting: 0
client: 1
current: 6619110
highest: 0
starting: 0
client: 2
...
client: 1
current: 6619110
highest: 6666234
starting: 6619110
INFO [04-06|14:21:27.332] Imported new block headers               count=192 elapsed=115.877ms number=6,619,302 hash=de8596..cfb8c4 age=1w1d3h
INFO [04-06|14:21:27.533] Imported new block headers               count=192 elapsed=32.827ms  number=6,619,494 hash=a119dc..f08c74 age=1w1d2h
INFO [04-06|14:21:27.818] Imported new block headers               count=192 elapsed=33.990ms  number=6,619,686 hash=9c59c7..50f588 age=1w1d2h
client: 2
current: 6619110
highest: 0
starting: 0
...
client: 1
current: 6666242
highest: 6666242
starting: 6619110
...
client: 2
current: 6619110
highest: 6666272
starting: 6619110
...
client: 2
current: 6666279
highest: 6666279
starting: 6619110
```

JoeGruffins: So, I think you are right. My way was just working because prog.CurrentBlock == prog.HighestBlock would still be not syncing... Otherwise the header check was stopping it?

JoeGruffins:
Running the syncing in two goroutines doesn't seem to matter as they do not sync at the same time for me.

client/asset/eth/nodeclient_harness_test.go
chappjc: > This does not use the harness at all.

Right right, there's only the light nodes for the dex wallets, and those just use testnet via internet or the specified bootnode.  Got it.

client/asset/eth/nodeclient_harness_test.go
```diff
@@ -158,129 +214,263 @@ func getContractAddrFromFile(fileName string) (common.Address, error) {
...
+	testTokenContractAddr, err = getContractAddrFromFile(testTokenContractAddrFile)
+	if err != nil {
+		return 1, err
+	}
+	fmt.Printf("Test token contract addr is %v\n", testTokenContractAddr)
+	err = setupWallet(simnetWalletDir, simnetWalletSeed, "localhost:30355", dex.Simnet)
+	if err != nil {
+		return 1, err
+	}
+	err = setupWallet(participantWalletDir, participantWalletSeed, "localhost:30356", dex.Simnet)
```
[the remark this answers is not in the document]
JoeGruffins: I think it's fine so everyone can test. Would need to be a fairly technical person to jack the testnet coins, and then I don't think they are worth real money... so no incentive?

I can drain the funds and do something else if you like. Some of the tests don't need funds and would work anyway. For instance `TestBasicRetrieval`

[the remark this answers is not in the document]
chappjc: I don't care about the funds that are in there presently.  But also what happens when two devs are using the same wallet at the same time?  I have a hard time seeing this work well on a public testnet.

I'm trying to understand if there's a way to have this work on testnet but with a seed that can be specified locally, either by editing the code or by an `-ldflags` switch.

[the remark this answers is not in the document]
chappjc: What if if the testnet seed/key could be specified locally, and there were a special test like `TestCreateTestnetWallet` that you could run to make the wallet and native light node, and it would spit out an address, then you could fund it, wait for confs, then run the other testnet tests now that there is funding there.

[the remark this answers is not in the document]
chappjc: Alternative the testnet tests could look for some text file in an expected location for the seed?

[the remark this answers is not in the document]
JoeGruffins: > But also what happens when two devs are using the same wallet at the same time?

Yeah, that would cause unexpected results.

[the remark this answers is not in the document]
JoeGruffins: Ok. Will add test that tests for funds and fails if there are not enough to run tests that need funds. It will also spit out the addresses that need funds.

And can the testnet seed just be a separate const? I already have the `testnet` switch as a constant... and the peer. So, following that pattern.

JoeGruffins:
Just rebased.

client/asset/eth/nodeclient_harness_test.go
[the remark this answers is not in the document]
JoeGruffins: https://github.com/decred/dcrdex/compare/e73f115596fee648b884e6bd6d1a955f83718482..271de4a504d8f0e444d37161b1b1cce0d0e2a9f8

chappjc (approved):
I'm fine with this unless there are irresolvable conflicts in https://github.com/decred/dcrdex/pull/1404 or https://github.com/decred/dcrdex/pull/1399
There will definitely be conflicts, but I don't think anything is conflicting conceptually.

JoeGruffins:
Oh, conflicts, will resolve.

PR #1553, opened 2022-03-30
client/asset/btc: attempt audit with empty txdata

chappjc:
This fixes the client btc (and clone) "livetest" simnet tests, and makes the `asset.Wallet` API usable without a server or oracle providing the txData of the counterparty contract.  `Core` should still always provide txData, but in general this is a needless limitation that I have had to confront in several contexts recently.

```
When AuditContract was switched to a hard requirement of txData
(the raw tx bytes), it broke certain simnet tests where no server was
available to provide the data, and made the asset.Wallet API somewhat
incomplete with no way to retrieve this txData in either full node or
SPV mode because there is no method to do so.  This makes a redeem
cli tool or other use case with no server or txdata oracle extremely
cumbersome because the raw bytes are required as an input instead of
just the coinid/txid.

Normally the server should communicate this txData, and the caller can
decide to require it. This adds (back) the ability to work with an
empty txData as a convenience for recovery tools and testing.  This MAY
change in the future if a GetTxData method is added for this purpose.

The baseWallet AuditContract now attempts retrieval of the transaction
output from the network, but it is only ensured to succeed for a full
node or, if the tx is confirmed, and SPV wallet.  If this is a full node
wallet, a simple gettxout RPC is sufficient with no pkScript or "since"
time. If this is an SPV wallet, only a confirmed counterparty contract
can be located, and only one within the last 48 hours. As such, this
mode of operation is not intended for normal server-coordinate
operation.

The above change fixes the "livetest" package, which is also used by
all btc "clone" assets. The test code is also updated with some extra
sleeps for SPV wallets.  A comment at the call site of AuditContract
is made noting that a nil txData must be provided because in normal
operation that data comes from the server coordinating the swap, and
that the Wallet API has no ability to get this data.
```

For reference, the commit that dropped support for empty txData: https://github.com/decred/dcrdex/commit/fe1995a42289f29e18bcccd9e537a2ba707a8ad1
That was fine if server or oracle is always providing it, but the `asset.Wallet` API is quite lacking without this capability.

I will similarly revitalize DCR's `AuditContract` after this.  It also appears to be a non-issue for eth to do this.

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -73,6 +73,11 @@ const (
 )
 
 var (
+	// ContractSearchLimit is how far back in time AuditContract in SPV mode
+	// will search for a contract if no txData is provided. This should be a
+	// positive duration.
+	ContractSearchLimit = 48 * time.Hour
+
 	// blockTicker is the delay between calls to check for new blocks.
 	blockTicker                  = time.Second
 	peerCountTicker              = 5 * time.Second
@@ -2190,7 +2195,13 @@ func (btc *baseWallet) SignMessage(coin asset.Coin, msg dex.Bytes) (pubkeys, sig
 
 // AuditContract retrieves information about a swap contract from the provided
 // txData. The extracted information would be used to audit the counter-party's
-// contract during a swap.
+// contract during a swap. The txData may be empty to attempt retrieval of the
+// transaction output from the network, but it is only ensured to succeed for a
+// full node or, if the tx is confirmed, an SPV wallet. Normally the server
+// should communicate this txData, and the caller can decide to require it. The
+// ability to work with an empty txData is a convenience for recovery tools and
+// testing, and it may change in the future if a GetTxData method is added for
+// this purpose.
 func (btc *baseWallet) AuditContract(coinID, contract, txData dex.Bytes, rebroadcast bool) (*asset.AuditInfo, error) {
 	txHash, vout, err := decodeCoinID(coinID)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


JoeGruffins (approved):

client/asset/btc/btc.go
```diff
@@ -2270,6 +2291,37 @@ func (btc *baseWallet) AuditContract(coinID, contract, txData dex.Bytes, rebroad
...
+	}
+
+	// In the off chance that the transaction is in mempool or txindex is
+	// enabled, try getrawtransaction.
+	txHash, _, err := decodeCoinID(coinID)
+	if err != nil {
+		return nil, err
+	}
+	tx, err := full.GetRawTransaction(txHash)
+	if err != nil {
```
buck54321, line 2315: This call is a little funny, because, as you mentioned, it will only work with mempool or txindex, but what we're really after is just the `*wire.TxOut`, and `gettxout` works for mempool and doesn't require txindex. The only difference is that now the tx will be rebroadcast, but that's primarily a concern for SPV clients who can't see mempool transactions, and at this point, we already know we're not SPV. You've added code to `(*baseWallet).AuditContract` to do the `gettxout` call already, and `rpcClient` already ignores the 48-hour filter and passes `true` for `includeMempool`. From my read, it looks like the entire `(*ExchangeWalletFullNode).AuditContract` method could be foregone without consequence.

chappjc: Huh, quite right.  If the `getrawtransaction` succeeds, there's no point in rebroadcasting it because the node would reject it as already existing.

chappjc:
For reference, I addressed DCR's simnet `AuditContract` issues in https://github.com/decred/dcrdex/commit/e42db6eeebf4a462ff84f567cbe6315eed1f90aa by using a GetTransaction call to the wallet that authored the swap I believe.  I was having trouble making a similar hack for BTC, but I really think `AuditContract` should at least try to retrieve the txout in the absence of txData.

buck54321 (approved):

PR #1554, opened 2022-03-30
ltc: switch to segwit

chappjc:
Rebased on https://github.com/decred/dcrdex/pull/1553 so the "livetest" simnet test can actually succeed.

```
This switches the LTC code to using and requiring segwit features.

This bumps the asset version to 1.  Note that we do not have a good
method for supporting multiple asset versions simultaneously, but
we can punt on that since LTC has not been deployed on mainnet
anywhere to my knowledge.
```

The client has asset version checks and will simply refuse an order if the server and client asset version does not match.

See discussion about possible client support for multiple asset versions here: https://github.com/decred/dcrdex/pull/1054#discussion_r616266089

### DIFF
--- a/dex/networks/ltc/params.go
+++ b/dex/networks/ltc/params.go
@@ -19,24 +19,31 @@ var (
 	}
 	// MainNetParams are the clone parameters for mainnet.
 	MainNetParams = btc.ReadCloneParams(&btc.CloneParams{
-		PubKeyHashAddrID: 0x30,
-		ScriptHashAddrID: 0x32,
+		Name:             "mainnet",
+		PubKeyHashAddrID: 0x30, // starts with L
+		ScriptHashAddrID: 0x32, // starts with M
 		Bech32HRPSegwit:  "ltc",
 		CoinbaseMaturity: 100,
 		Net:              0xdbb6c0fb,
+		HDPrivateKeyID:   [4]byte{0x04, 0x88, 0xad, 0xe4}, // starts with xprv
+		HDPublicKeyID:    [4]byte{0x04, 0x88, 0xb2, 0x1e}, // starts with xpub
 	})
 	// TestNet4Params are the clone parameters for testnet.
 	TestNet4Params = btc.ReadCloneParams(&btc.CloneParams{
-		PubKeyHashAddrID: 0x6f,
-		ScriptHashAddrID: 0x3a,
+		Name:             "testnet4",
+		PubKeyHashAddrID: 0x6f, // starts with m or n
+		ScriptHashAddrID: 0x3a, // starts with Q
 		Bech32HRPSegwit:  "tltc",
 		CoinbaseMaturity: 100,
 		Net:              0xf1c8d2fd,
+		HDPrivateKeyID:   [4]byte{0x04, 0x35, 0x83, 0x94}, // starts with tprv
+		HDPublicKeyID:    [4]byte{0x04, 0x35, 0x87, 0xcf}, // starts with tpub
 	})
 	// RegressionNetParams are the clone parameters for simnet.
 	RegressionNetParams = btc.ReadCloneParams(&btc.CloneParams{
-		PubKeyHashAddrID: 0x6f,
-		ScriptHashAddrID: 0x3a,
+		Name:             "regtest",
+		PubKeyHashAddrID: 0x6f, // starts with m or n
+		ScriptHashAddrID: 0x3a, // starts with Q
 		Bech32HRPSegwit:  "rltc",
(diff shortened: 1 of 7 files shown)


dex/networks/ltc/params.go
```diff
@@ -18,34 +18,54 @@ var (
 		},
 	}
 	// MainNetParams are the clone parameters for mainnet.
-	MainNetParams = btc.ReadCloneParams(&btc.CloneParams{
```
chappjc, line 21: I was having a really hard time seeing my way to using the `CloneParams` stuff here.  I'd rather do this directly and transparently.

chappjc:
The litecoin testnet chain is completely stuck right now, but I tested this yesterday.

JoeGruffins (approved):

dex/networks/ltc/params.go
buck54321: `CloneParams` was a reduction of the `Params` to only the fields that were actually used, in order to simplify and document implementation of clones. The entire dex Bitcoin RPC implementation only needs those five fields. Someone trying to copy this clone impl. will now end up doing more work than they need, which is not trivial when there is no btcd clone from which to copy the parameters.

dex/networks/ltc/params.go
```diff
@@ -18,34 +18,54 @@ var (
...
+	MainNetParams = &chaincfg.Params{
+		Name:                    "mainnet",
+		Net:                     wire.BitcoinNet(0xdbb6c0fb),
+		CoinbaseMaturity:        100,
+		Bech32HRPSegwit:         "ltc",
+		PubKeyHashAddrID:        0x30,                            // starts with L
+		ScriptHashAddrID:        0x32,                            // starts with M
+		PrivateKeyID:            0xB0,                            // starts with 6 (uncompressed) or T (compressed)
+		WitnessPubKeyHashAddrID: 0x06,                            // starts with p2
+		WitnessScriptHashAddrID: 0x0A,                            // starts with 7Xh
```
buck54321, line 30: For example, these two appear to be unused anywhere in the btcd, btcutil, or btcwallet. But someone trying to find these parameters from the C++ code of a non-ported project might waste a lot of time looking for them. They don't appear to exist there, from what I can tell.

dex/networks/ltc/params.go
chappjc: Great points.  I'll switch back.  What started me in this direction was partly looking to make the `Addresser` used for server fees generic enough for clones, which means the HD fields.  Also the Name struck me as one that is reasonable to have set to something for logging and/or debugging.  Will address the fee address issues subsequently and update `CloneParams` at that time.

buck54321: Please do improve as you see necessary. Just wanted to convey the intention. I probably should have doc'd better.

dex/networks/ltc/params.go
buck54321: To note, these lines were added after @davecgh had relinquished control of the repo.

https://github.com/btcsuite/btcd/blame/074266215c26f8c4731330abb515f92b905f8dfa/chaincfg/params.go#L261

buck54321 (approved):
LGTM. Inspired #1565 though.

client/asset/btc/btc.go
```diff
@@ -2202,19 +2208,34 @@ func (btc *baseWallet) AuditContract(coinID, contract, txData dex.Bytes, rebroad
...
+	// sufficient with no pkScript or "since" time. If this is an SPV wallet,
+	// only a confirmed counterparty contract can be located, and only one
+	// within the last 48 hours. As such, this mode of operation is not intended
+	// for normal server-coordinated operation.
+	var tx *wire.MsgTx
+	var txOut *wire.TxOut
+	if len(txData) == 0 {
+		// Fall back to gettxout, but we won't have the tx to rebroadcast.
+		pkScript, _ := btc.scriptHashScript(contract) // pkScript and since time are unused if full node
+		txOut, _, err = btc.node.getTxOut(txHash, vout, pkScript, time.Now().Add(-48*time.Hour))
```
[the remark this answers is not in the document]
chappjc: These two comments are for https://github.com/decred/dcrdex/pull/1553, which this PR is stacked on. Will make the changes there.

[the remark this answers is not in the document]
chappjc: @buck54321 As a half measure, I made this an exported `var` so such oddball use cases (recovery tools and tests) can at least adjust the limit.  I think if we want this used more commonly, we could go the route of adding a `GetTxData` method as mentioned in the comments, and that could have this since time as an arg.

Issue #1544, opened 2022-03-24
client/eth: attempting to unlock more than is reserved


JoeGruffins:
A random error testing on testnet. No idea currently what caused it. Full error is `2022-03-23 03:24:19.993 [ERR] CORE: attempting to unlock more than is reserved for order bea4ab86d2fb6d7585276cb1d10bbd04afdceb9e930c210d231b29b13c718fef. unlocking reserved amount instead: 0 reserved, unlocking 8600000`. Here are the client logs in that area. I can also provide server logs if needed.

https://pastebin.com/DCA6zSy3

JoeGruffins:
Looking into it.

JoeGruffins:
Can `t.findMakersRedemption` happen after a `t.processMakersRedemption`?

JoeGruffins:
Working on top of #1556 I see in core logs:
```
2022-04-04 07:25:04.752 [ERR] CORE: attempting to unlock more than is reserved for order fe6721132602847d649a07db09cac62735540ebe73f106efe6144334100ffecf. unlocking reserved amount instead: 0 reserved, unlocking 8600000
```
BUT, pretty sure this to wallet is DCR. 
![image](https://user-images.githubusercontent.com/34998433/161529082-0878d024-a510-4e99-8449-30fb6832f0e0.png)

Not sure if the `t.accountRedeemer()` is sending a wrong signal somewhere. Still looks pretty random, as I have a trade directly after that looks the same but does not try to unlock coins here.

chappjc:
There would be reserves for an ETH **refunds** when buying DCR with ETH, so this error is probably coming via `unlockRefundFraction`.

chappjc:
One thing that looks suspect is in `(*trackedTrade).revoke` we have `t.unlockRefundFraction(t.Trade().Remaining(), t.Trade().Quantity)`.  But revoking the trade does not mean you don't still have to refund or even redeem active matches.

JoeGruffins:
I think I've found where this is happening. Will hopefully have a pr up tomorrow.

chappjc:
Reopening as per findings reported in https://github.com/decred/dcrdex/pull/1721#issuecomment-1191042640

If it can be hit with loadbot on `master` (the compound program, @JoeGruffins?), we should be able to get to the bottom of it, and I can clean up https://github.com/decred/dcrdex/pull/1653 to see if it will resolve the issue without making more.

JoeGruffins:
> the compound program

Yes.

PR #1523, opened 2022-03-16
client/asset/eth: token prefactor

buck54321:
Preparing for tokens. Peeled off of #1399.

```
Moves contractor from nodeClient to ExchangeWallet. The changes the
ethFetcher API quite a bit, but is overall a simplification. To accomplish
this, I moved the nonceSendMtx to from nodeClient to ExchangeWallet too.
Nonce generation and signing is moved into nodeClient.txOpts to dedupe.

Use block headers instead of full blocks for tip monitoring.

Simplify fundingCoin and eliminate fundingCoinID.

Simplify tests using more consts and simpler structs.
```

PR #1556, opened 2022-04-01
client: Let core handle redeem and refund locking.

### LINKED ISSUE #1539: client/eth: 12600000 > 0 redemption funds reserved error.
I apologize if this is already an issue. I feel like we have had several issues/pr about this, but when trading on simnet or testnet, I still see `[ERR] CORE: retiring order 0d52237b5b799d416c78d35543ce3ef05c4a7fa118be701f2cd4600d8c23cc4d with 12600000 > 0 redemption funds reserved` If this is expected, we should not log as error. If not expected should fix.

To reproduce make a 1lot trade for dcr/eth and watch it to the end.

JoeGruffins:
For account based coins, let core use the Reserve and Unlock methods to
    handle redemption and refund reserves for assets that use them. This
    solves a minor bug where redemption funds are not unlocked for a tracked
    trade that is redeemed normally.

closes #1539

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -5948,10 +5948,14 @@ func (c *Core) listen(dc *dexConnection) {
 			for _, trade := range doneTrades {
 				// Log an error if redemption funds are still reserved.
 				trade.mtx.RLock()
-				reserved := trade.redemptionLocked
+				redeemLocked := trade.redemptionLocked
+				refundLocked := trade.refundLocked
 				trade.mtx.RUnlock()
-				if reserved > 0 {
-					dc.log.Errorf("retiring order %s with %d > 0 redemption funds reserved", trade.ID(), reserved)
+				if redeemLocked > 0 {
+					dc.log.Errorf("retiring order %s with %d > 0 redemption funds locked", trade.ID(), redeemLocked)
+				}
+				if refundLocked > 0 {
+					dc.log.Errorf("retiring order %s with %d > 0 refund funds locked", trade.ID(), refundLocked)
 				}
 
 				c.notify(newOrderNote(TopicOrderRetired, "", "", db.Data, trade.coreOrder()))
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -5309,6 +5309,13 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
 			}
 		}
 
+		if refundNum != 0 {
+			tracker.lockRefundFraction(refundNum, denom)
+		}
+		if redeemNum != 0 {
+			tracker.lockRedemptionFraction(redeemNum, denom)
+		}
+
 		// Active orders and orders with matches with unsent swaps need funding
 		// coin(s). If they are not found, block new matches and swap attempts.
 		needsCoins := len(matchesNeedingCoins) > 0
@@ -5354,12 +5361,8 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
 		tracker.recalcFilled()
 
 		if isActive {
-			if refundNum != 0 {
-				tracker.lockRefundFraction(refundNum, denom)
-			}
-			if redeemNum != 0 {
-				tracker.lockRedemptionFraction(redeemNum, denom)
-			}
+			tracker.lockRedemptionFraction(trade.Remaining(), trade.Quantity)
+			tracker.lockRefundFraction(trade.Remaining(), trade.Quantity)
 		}
 
 		// Balances should be updated for any orders with locked wallet coins,
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -5186,19 +5186,24 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
 
 		tracker.wallets = wallets
 
-		lockMatchRedemption := func(match *matchTracker) {
+		// Find the least common multiplier to use as the denom for adding
+		// reserve fractions.
+		denom, marketMult, limitMult := lcm(uint64(len(tracker.matches)), tracker.Trade().Quantity)
+		var refundNum, redeemNum uint64
+
+		addMatchRedemption := func(match *matchTracker) {
 			if tracker.isMarketBuy() {
-				tracker.lockRedemptionFraction(1, uint64(len(tracker.matches)))
+				redeemNum += marketMult // * 1
 			} else {
-				tracker.lockRedemptionFraction(match.Quantity, trade.Quantity)
+				redeemNum += match.Quantity * limitMult
 			}
 		}
 
-		lockMatchRefund := func(match *matchTracker) {
+		addMatchRefund := func(match *matchTracker) {
 			if tracker.isMarketBuy() {
-				tracker.lockRefundFraction(1, uint64(len(tracker.matches)))
+				refundNum += marketMult // * 1
 			} else {
-				tracker.lockRefundFraction(match.Quantity, trade.Quantity)
+				refundNum += match.Quantity * limitMult
 			}
 		}
 
@@ -5218,8 +5223,8 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
 					counterSwap = match.MetaData.Proof.TakerSwap
 				}
 				if match.Status < order.MakerRedeemed {
-					lockMatchRedemption(match)
-					lockMatchRefund(match)
+					addMatchRedemption(match)
+					addMatchRefund(match)
 				}
 			} else { // Taker
 				if match.Status < order.TakerSwapCast {
@@ -5230,10 +5235,10 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
 					counterSwap = match.MetaData.Proof.MakerSwap
 				}
 				if match.Status < order.MakerRedeemed {
-					lockMatchRefund(match)
+					addMatchRefund(match)
 				}
 				if match.Status < order.MatchComplete {
-					lockMatchRedemption(match)
+					addMatchRedemption(match)
 				}
 			}
 			c.log.Tracef("Trade %v match %v needs coins = %v, needs audit info = %v",
@@ -5349,8 +5354,12 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
 		tracker.recalcFilled()
 
 		if isActive {
-			tracker.lockRedemptionFraction(trade.Remaining(), trade.Quantity)
-			tracker.lockRefundFraction(trade.Remaining(), trade.Quantity)
+			if refundNum != 0 {
+				tracker.lockRefundFraction(refundNum, denom)
+			}
+			if redeemNum != 0 {
+				tracker.lockRedemptionFraction(redeemNum, denom)
+			}
 		}
 
 		// Balances should be updated for any orders with locked wallet coins,
--- a/client/core/core_test.go
+++ b/client/core/core_test.go
@@ -5157,6 +5157,7 @@ func TestResolveActiveTrades(t *testing.T) {
 			Proof:              db.OrderProof{},
 			ChangeCoin:         changeCoinID,
 			RedemptionReserves: redemptionReserves,
+			RefundReserves:     refundReserves,
 		},
 		Order: lo,
 	}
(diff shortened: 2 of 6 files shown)


martonp (approved):
Looks good, but why was it not working properly previously?

JoeGruffins:
> Looks good, but why was it not working properly previously?

Awesome!

The refunds should have spit out more errors, I just don't think we test them enough.

I think the accounting for the eth package was fine, just unlocking the same funds multiple times in some circumstances, because core is also tracking this for the trade. I don't guess the core locker serves a purpose besides logging? This does not solve #1544 which I guess I'll look into next...

chappjc:
> I think the accounting for the eth package was fine, just unlocking the same funds multiple times in some circumstances, because core is also tracking this for the trade. I don't guess the core locker serves a purpose besides logging?

Can you expand on this?  So both core and internally eth were unlocking stuff?  Are we following a consistent approach to the responsibility of coin locking / unlocking?

client/core/core.go
```diff
@@ -5948,10 +5948,14 @@ func (c *Core) listen(dc *dexConnection) {
 			for _, trade := range doneTrades {
 				// Log an error if redemption funds are still reserved.
 				trade.mtx.RLock()
-				reserved := trade.redemptionLocked
+				redeemLocked := trade.redemptionLocked
+				refundLocked := trade.refundLocked
```
chappjc, line 5952: Moving the refund unlocking does look more consistent, given we were doing it here for redeems...

This change set does feel better.

JoeGruffins:
> So both core and internally eth were unlocking stuff?

Yes.

Here when we redeem in core, we unlock the refund reserves in core and the eth package:
https://github.com/decred/dcrdex/blob/fb76b59006d744e149a48ecc2218eaead4920819/client/core/trade.go#L2038-L2042
but we unlock the redeem reserves in the eth package only, hence there being reserves left in the trade reserves at the end for a normal redeem:
https://github.com/decred/dcrdex/blob/fb76b59006d744e149a48ecc2218eaead4920819/client/asset/eth/eth.go#L878

If we want to lock/unlock the coins for both, the eth package cannot touch core, so it only makes sense to let core handle it imo. Core is already handling all the locking and most of the unlocking.

> Are we following a consistent approach to the responsibility of coin locking / unlocking?

No but this should change that as all use of the account locker methods inside the eth package is removed in this pr. That is `ReserveNRedemptions`, `UnlockRedemptionReserves`, `ReReserveRedemption`, `ReserveNRefunds`, `UnlockRedemptionReserves`, and `ReReserveRefund` are only used in core.

client/core/trade.go
```diff
@@ -2043,6 +2030,11 @@ func (c *Core) redeemMatchGroup(t *trackedTrade, matches []*matchTracker, errs *
 			match.Status = order.MakerRedeemed
 			proof.MakerRedeem = coinID
 		}
+		if t.isMarketBuy() {
+			t.unlockRedemptionFraction(1, uint64(len(t.matches)))
+		} else {
+			t.unlockRedemptionFraction(match.Quantity, t.Trade().Quantity)
+		}
```
buck54321, line 2037: How about just sum the `match.Quantity` while you loop so you can call your unlock functions just once at the end?

chappjc:
Hmm, last change [didn't work out](https://github.com/decred/dcrdex/runs/5882544593?check_suite_focus=true):

```
 2022-04-08 09:09:59.421 [ERR] TCORE: attempting to unlock more than is reserved for order a5aafd01d81f56054d7021ae9aa643adf30c5fad3f415275792169cd0e4de453. unlocking reserved amount instead: 0 reserved, unlocking 40
2022-04-08 09:09:59.422 [INF] TCORE: loaded 1 incomplete orders
--- FAIL: TestResolveActiveTrades (0.00s)
    core_test.go:5269: initial: side = Taker, order status = booked, match status = MakerSwapCast: redemption funds not reserved, 10 != 50
```

chappjc:
A few mostly trivial conflicts created by https://github.com/decred/dcrdex/pull/1523

JoeGruffins:
Yes, srry. I should run tests first. I'm looking into it.

Oh conflicts good. Srry I didn't re-review #1523, was getting around to it...

client/core/core.go
```diff
@@ -5349,8 +5354,12 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
 		tracker.recalcFilled()
 
 		if isActive {
-			tracker.lockRedemptionFraction(trade.Remaining(), trade.Quantity)
-			tracker.lockRefundFraction(trade.Remaining(), trade.Quantity)
```
JoeGruffins, line 5353: This was incorrect right? We were locking per match and here at the end. We should just do the targeted locks...

client/core/core_test.go
```diff
@@ -5240,10 +5241,12 @@ func TestResolveActiveTrades(t *testing.T) {
...
 		}
 
+		isActive := dbOrder.MetaData.Status == order.OrderStatusBooked || dbOrder.MetaData.Status == order.OrderStatusEpoch
+
 		if lo.T.Sell && ((match.Side == order.Taker && match.Status < order.MatchComplete) ||
 			(match.Side == order.Taker && match.Status < order.MakerRedeemed)) {
-			var reReserveQty uint64 = redemptionReserves
-			if dbOrder.MetaData.Status > order.OrderStatusBooked {
+			var reReserveQty uint64
+			if isActive {
```
JoeGruffins, line 5249: Not sure what was going on with this test. We only want locked if the matches are active correct?

client/core/core.go
JoeGruffins: Oh wait, this is locking unmatched funds.

client/core/core_test.go
JoeGruffins: Ok, I got it now. We want them all locked unless the match is no longer active then just the matched. Took a while to seep in.

client/core/trade.go
```diff
@@ -214,9 +214,9 @@ func (t *trackedTrade) accountRefunder() (asset.AccountLocker, bool) {
 	return ar, is
 }
 
-// lockRedemptionFraction locks the specified fraction of the available
+// lockRefundFraction locks the specified fraction of the available
 // refund reserves. Subsequent calls are additive. If a call to
-// lockRedemptionFraction would put the locked reserves > available reserves,
+// lockRefundFraction would put the locked reserves > available reserves,
```
chappjc, line 219: Not related to your changes, but I'm having a very hard time understanding why we're going to so much trouble of working with a rational number's numerator and denominator if we just end up doing this:

```
func applyFraction(num, denom, target uint64) uint64 {
	return uint64(math.Round(float64(num) / float64(denom) * float64(target)))
}
```

We're _are_ going to see precision loss and likely the occasional wrong result, but we still have to jump through all these hoops.  What am I missing?

I'm looking at `math/big.Rat` if we want to be rigorous w.r.t. precision.  Thoughts or clarification?

chappjc:
```go
func applyFraction(num, denom, target uint64) uint64 {
	frac := big.NewRat(int64(num), int64(denom))
	res, exact := new(big.Rat).Mul(frac, big.NewRat(int64(target), 1)).Float64()
	fmt.Println("**************** frac:", frac)
	fmt.Println("**************** val:", target)
	fmt.Printf("**************** scaled: %v (f) => %d (i) (exact = %v)\n", res, uint64(math.Round(res)), exact)
	simple := uint64(math.Round(float64(num) / float64(denom) * float64(target)))
	fmt.Println("**************** simple:", simple)
	return simple
}
```

Exposes some issues with `TestRedemptionReserves` and `TestRefundReserves`.  In particular trying to divide an order size by a value that the order cannot be split evenly.

Test issues aside, I'm still thinking that we an change the `trackedTrade` fields to a new system whereby for say, redeem reserves, one field is the multiple size and the other field is the number of reserved multiples.  Then the unlock methods just need to specify an integer multiple to unlock, which just decrements the locked integer.

chappjc: Nothing to change with this PR, @JoeGruffins .  Just clarifying

JoeGruffins: I agree. The current locker does deal with "dust" if a remainder slipped through, but there's probably a more strait forward answer. I'll make an issue?

buck54321: > I'll make an issue?

For what?

chappjc: The fraction system is turning out to be brittle, complex, and prone to precision loss and rounding error.  If we continue to have trouble with reserves locking, I'll propose some changes as I described in the last paragraph here: https://github.com/decred/dcrdex/pull/1556#discussion_r846523350
No need to divert effort yet.

buck54321: But as @JoeGruffin's pointed out, your comment 

> Exposes some issues with TestRedemptionReserves and TestRefundReserves. In particular trying to divide an order size by a value that the order cannot be split evenly.

is incorrect, since we deal with rounding errors in `reservesToUnlock`.

There've clearly been some hiccups in implementing refunds, but that doesn't mean the system is brittle or complex.

Not trying to stop you from trying your plan though.

chappjc: I don't mean to harshly criticize the design here; I just believe that it does not handle rounding well and it its use is error prone.

In `applyFraction`, the `math.Round` deals with floating point _precision_ errors, not errors in the input arguments that create an expression where the number cannot be divided evenly even on paper, never mind floating point precision.  The main issue is not when precision results in something like 3999.9999997, and that rounds to 4000 just fine; it's when num and denom result in something like 13333.4 or 666.6666666.  The result will just be wrong because the caller gave unsatisfiable inputs.

I _sort of_ understand the dust handling in `reservesToUnlock`, but why should there be dust.  Also what if the rounding in applyFraction goes the other way an is unlocking one extra gwei instead?  Is that all handled?  I cannot tell.

In the tests, run with the modified code I posted above.  First thing I noticed is this:

```
**************** frac: 2/3
**************** val: 100000
**************** scaled: 66666.66666666667 (f) => 66667 (i) (exact = false)
**************** simple: 66667
```

because of 
```
const reserves = 100_000
qty := lotSize * 10
matchQty := qty * 2 / 3
matchReserves := applyFraction(2, 3, reserves)
```

The math.Round isn't correcting for floating point precision with this; it's just not a whole number _on paper_.  Is this really expected?  I don't get it if so.

Even if this is all somehow "right", there's still little point in keeping num and denom separate because of `math.Round(float64(num) / float64(denom) * float64(target))`, which divides num/denom into a float first anyway.

chappjc: I'm sure a big part of this is just me being slow, but I am confused by how to use the functions and what the arguments mean, and I'm surprised at the acrobatics needed to deal with the reserves.

buck54321: I hear ya. I'll add some notes to explain some of the choices and hopefully save you some time in implementation.

- **Assuming that the redemption and refund reserves are evenly divisible by the number `n` passed to `ReserveNRedemptions`**, then for all orders but market buy orders, everything will work out evenly without rounding.
- The generalization of returns as fractional is to deal with market buy orders, where we can't know at order-time how many matches there will be. We **could** store the `n` value estimate we used to reserve and use that to eliminate rounding for market buy orders.
- I do think that the integer system that you're planning will be mathematically identical to this, just with more data persistence. Consider this. I could, instead of 
 
```Go
returnQty := applyFraction(matchQuantity, orderQuantity, reserves)
```
do
```Go
matchLots := matchQuantity / mktInfo.LotSize
orderLots := orderQuantity / mktInfo.LotSize
returnQty := applyFraction(matchLots, orderLots, reserves)
```

One problem is we're not saving the market info, but let's overlook that for a second.

If we were instead to save **a reserve quantity per lot** (in the spirit of your suggestion, I think), then we might have something like this

```Go
matchLots := matchQuantity / mktInfo.LotSize
returnQty := tracker.reserveQtyPerLot * matchLots
```

Which looks clean, but would require the persistence of more data.

Going back now, internally in `applyFraction`, I do roughly

```Go
return matchQuantity / orderQuantity * reserves
```
or equivalently
```Go
return matchLots / orderLots * reserves
```
Multiplying numerator and denominator by `1 / orderLots`
```Go
return matchLots * (reserves / orderLots)
```

which is exactly where we got with the alternative approach. Same result, more data persistence required. 

> Even if this is all somehow "right", there's still little point in keeping num and denom separate because of `math.Round(float64(num) / float64(denom) * float64(target))`, which divides num/denom into a float first anyway.

I'm not sure I understand. How else could I handle `num` and `denom`?

chappjc: Thanks for describing that all in detail.  I appreciate it.

> I'm not sure I understand. How else could I handle `num` and `denom`?

Because `num` and `denom` only ever get used in `applyFraction` (afaict), it could be: 

```go
// frac is on [0,1], target is an amount in atomic units.
func applyFraction(frac float64, target uint64) uint64 {
	return uint64(math.Round(frac * float64(target)))
}
```

Given the above, unless we got fancy with integer math or big.Rat, I'm not seeing a reason to make the caller worry about espressing the frac as a rational number.

buck54321: I was just trying to avoid math and conversions at the call sites.

chappjc (approved):
OK.  I'm still a bit perturbed that we have to compute a least common multiple via greatest common divisor via Euclidean algorithm just reservere/unreserve some funds...

PR #1558, opened 2022-04-03
multi: add support for Dogecoin

buck54321:
Adds support for Dogecoin. Implementation was straight-forward. A few differences in the RPC call arguments, but the clone framework adapted well.

My ~/.dogecoin/dogecoin.conf for mainnet looks like 

```
txindex=1
printtoconsole=1
rpcuser=buck
rpcpassword=somerandomness
```


client/asset/doge/doge.go
```diff
@@ -0,0 +1,160 @@
...
+	dexdoge "decred.org/dcrdex/dex/networks/doge"
+	"github.com/btcsuite/btcd/chaincfg"
+)
+
+const (
+	version = 0
+	BipID   = 3
+	// The default fee is passed to the user as part of the asset.WalletInfo
+	// structure.
+	defaultFee          = 500000
```
chappjc, line 22: such fees, much zeros

dex/testing/doge/harness.sh
```diff
@@ -0,0 +1,210 @@
...
+tmux rename-window -t $SESSION:0 'alpha'
+tmux send-keys -t $SESSION:0 "set +o history" C-m
+tmux send-keys -t $SESSION:0 "cd ${ALPHA_DIR}" C-m
+echo "Starting simnet alpha node"
+tmux send-keys -t $SESSION:0 "${DAEMON} -rpcuser=user -rpcpassword=pass \
+  -rpcport=${ALPHA_RPC_PORT} -datadir=${ALPHA_DIR} \
+  -debug=rpc -debug=net -debug=mempool -debug=walletdb -debug=addrman -debug=mempoolrej \
+  -whitelist=127.0.0.0/8 -whitelist=::1 \
+  -txindex=1 -regtest=1 -port=${ALPHA_LISTEN_PORT} -fallbackfee=0.00001 \
+  -printtoconsole; tmux wait-for -S alpha${SYMBOL}" C-m
```
chappjc, line 92: Madness that dogecoind requires `printtoconsole`.  I killed it several times in the last week or so thinking it was hung

buck54321: Zcash too, but for different reasons.

chappjc:
`(*rpcClient).ownsAddress` will need to check a boolean to decide to use `getaddressinfo` or `validateaddress`.  The latest release of docgoind/qt does not have `getaddressinfo`

chappjc:
Ah, *dcr* does the following in `parseUTXOs` when checking the returns from `InputInfo`

```go
if errors.Is(err, dex.UnsupportedScriptError) {
	continue
}
```

We gotta do the same in btc.

client/asset/btc/btc.go
```diff
@@ -238,6 +238,18 @@ type BTCCloneCFG struct {
...
+	OmitAddressType bool
+	// LegacySignTxRPC causes the RPC client to use the signrawtransaction
+	// endpoint instead of the signrawtransactionwithwallet endpoint.
+	LegacySignTxRPC bool
+	// BooleanGetBlockRPC causes the RPC client to use a boolean second argument
+	// for the getblock endpoint, instead of Bitcoin's numeric.
+	BooleanGetBlockRPC bool
+	// SingularWallet signals that the node software supports only one wallet,
+	// so the RPC endpoint does not have a /wallet/{walletname} path.
+	SingularWallet bool
```
chappjc, line 252: `estimatesmartfee` also seems to have a different syntax, namely, no "mode":

```
2022-04-12 02:17:14.867 [ERR] CORE[doge]: Failed to get fee rate: rawrequest (estimatesmartfee) error: -1: estimatesmartfee nblocks

WARNING: This interface is unstable and may disappear or change!

Estimates the approximate fee per kilobyte needed for a transaction to begin
confirmation within nblocks blocks if possible and return the number of blocks
for which the estimate is valid. Uses virtual transaction size as defined
in BIP 141 (witness data is discounted).

Arguments:
1. nblocks     (numeric)

Result:
{
  "feerate" : x.x,     (numeric) estimate fee-per-kilobyte (in BTC)
  "blocks" : n         (numeric) block number where estimate was found
}

A negative value is returned if not enough transactions and blocks
have been observed to make an estimate for any number of blocks.
However it will not return a value below the mempool reject fee.

Example:
> dogecoin-cli estimatesmartfee 6
```

chappjc:
Had to hack through the issues described in my comments, but got a swap on testnet done:
- DOGE redeem: https://blockexplorer.one/dogecoin/testnet/tx/eb475670e7e7762a6b001a6224e9d8c1fb22c4e3f5a00f46eaa81e2ca2f2664d
- DCR redeem: https://testnet.dcrdata.org/tx/8d5593b5fc0b36b656719316e6a968f7841b504d8b9598f45e498de08374c793/out/0

I wonder if we're testing different versions?  I'm using v1.14.5 from the `1.14-maint` branch (9Dogecoin Core Daemon version v1.14.5.0-31afd1331`)

client/asset/btc/btc.go
chappjc: Ah server has this set (`btc.BackendCloneConfig.FeeEstimator`), but not client.

Also, on server, the fee estimator never warmed up.  The following have been spamming logs for over 12 hrs:

```
2022-04-12 14:19:36.002 [ERR] DEX: Error retrieving fee rate for doge: fee could not be estimated
```

That is, no `err` back from `estimatefee`, just a rate of `<=0`.  On my machine is returns -1 (!?)

Actually, looks like client needs to use `estimatefee` instead of `estimatesmartfee`, which has that "unstable or may disappear or change" warning.

Maybe it always returns -1 or 0 because testnet does not have enough transactions.  In any case, I think we need to figure out a safe way to silence it.

client/asset/btc/btc.go
```diff
@@ -557,7 +573,7 @@ type baseWallet struct {
 	useSplitTx        bool
 	useLegacyBalance  bool
 	segwit            bool
-	legacyRawFeeLimit bool
+	legacyRawFeeLimit bool // wut dis?
```
chappjc, line 576: Looks only needed at the level of `rpcWallet`, not `ExchangeWallet`.

BTW, in `rpcWallet`, would be nice if `SendRawTransactionLegacy` and `SendRawTransaction` were situated remotely close to the caller `sendRawTransaction`

client/asset/btc/rpcclient.go
```diff
@@ -205,7 +202,13 @@ func (wc *rpcClient) callHashGetter(method string, args anylist) (*chainhash.Has
 // getBlock fetches the MsgBlock.
 func (wc *rpcClient) getBlock(h chainhash.Hash) (*wire.MsgBlock, error) {
 	var txB dex.Bytes
-	err := wc.call(methodGetBlock, anylist{h.String(), 0}, &txB)
+	args := anylist{h.String()}
+	if wc.booleanGetBlock {
+		args = append(args, false)
+	} else {
+		args = append(args, 0)
```
chappjc, line 209: I just caught this discrepancy server-side too, where we were passing `true` instead of `1` to get the `GetBlockVerboseResult`.  We were/are basically getting lucky there since bitcoind/litecoind were happy to treat `true` as `1` even though they really want 0,1, or 2.

I'm planning to make the necessary changes in https://github.com/decred/dcrdex/pull/1581 since I'm also fixing up the LiveUTXO btc tests and I wanted the mgsBlock... and I noticed this issue.

buck54321:
> Looks like we'll have to `lockunspent true ...` when we spend any (locked) coins. Maybe in `(*rpcClient).SendRawTransaction` in a similar way to how it's done in `(*spvWallet).sendRawTransaction`?

Do you think we should do it for all assets, or hide it behind a config flag?

chappjc:
> Do you think we should do it for all assets, or hide it behind a config flag?

Had the same question myself.  What are the downsides to unlocking anyway?  Just the extra rpcs?

buck54321:
> > Do you think we should do it for all assets, or hide it behind a config flag?
> 
> Had the same question myself. What are the downsides to unlocking anyway? Just the extra rpcs?

Yeah, I think so. Just one rpc, actually.

chappjc:
Seems alright to me, as long as it doesn't end up spamming logs with errors about already-unlocked coins.

client/asset/doge/doge.go
```diff
@@ -23,6 +25,7 @@ const (
 	defaultFeeRateLimit = 1_000_000
 	minNetworkVersion   = 1140400
 	walletTypeRPC       = "dogecoindRPC"
+	feeConfs            = 10
```
chappjc, line 28: 10? not 1?  Does it work more reliably with a larger number?

client/asset/btc/rpcclient.go
```diff
@@ -138,16 +142,47 @@ func (wc *rpcClient) SendRawTransactionLegacy(tx *wire.MsgTx) (*chainhash.Hash,
...
+	// TODO: lockUnspent should really just take a []*outPoint, since it doesn't
+	// need the value.
+	ops := make([]*output, 0, len(tx.TxIn))
+	for _, txIn := range tx.TxIn {
+		prevOut := &txIn.PreviousOutPoint
+		ops = append(ops, &output{pt: newOutPoint(&prevOut.Hash, prevOut.Index)})
 	}
-	return wc.SendRawTransaction(tx)
+	if err := wc.lockUnspent(true, ops); err != nil {
+		wc.log.Warnf("error unlocking spent outputs: %v", err)
```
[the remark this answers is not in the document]
chappjc: FWIW, bitcoind also shows an error including `expected locked output`.  I think we can ignore errors with that.
dogecoind doesn't error at all when repeatedly unlocking.  I don't feel like checking BCH.

chappjc:
Every 6 seconds on server, two of these are logged like this:

```
2022-04-13 14:24:48.001 [ERR] DEX: Error retrieving fee rate for doge: fee could not be estimated
2022-04-13 14:24:48.002 [ERR] DEX: Error retrieving fee rate for doge: fee could not be estimated
2022-04-13 14:24:54.001 [ERR] DEX: Error retrieving fee rate for doge: fee could not be estimated
2022-04-13 14:24:54.001 [ERR] DEX: Error retrieving fee rate for doge: fee could not be estimated
2022-04-13 14:25:00.001 [ERR] DEX: Error retrieving fee rate for doge: fee could not be estimated
2022-04-13 14:25:00.002 [ERR] DEX: Error retrieving fee rate for doge: fee could not be estimated
2022-04-13 14:25:06.001 [ERR] DEX: Error retrieving fee rate for doge: fee could not be estimated
2022-04-13 14:25:06.002 [ERR] DEX: Error retrieving fee rate for doge: fee could not be estimated
```

Really have to silence this somehow before this can go to production.

chappjc:
maybe

```diff
diff --git a/client/asset/doge/doge.go b/client/asset/doge/doge.go
index 78dcfa9e..5060163e 100644
--- a/client/asset/doge/doge.go
+++ b/client/asset/doge/doge.go
@@ -173,7 +173,7 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network)
                                return 0, err
                        }
                        if feeRate <= 0 {
-                               return 0, fmt.Errorf("fee could not be estimated")
+                               return 0, nil // fmt.Errorf("fee could not be estimated")
                        }
                        return uint64(math.Round(feeRate * 1e5)), nil
                },
diff --git a/server/dex/feemgr.go b/server/dex/feemgr.go
index 97e106ed..834abb15 100644
--- a/server/dex/feemgr.go
+++ b/server/dex/feemgr.go
@@ -85,6 +85,9 @@ func (f *feeFetcher) FeeRate(ctx context.Context) uint64 {
                log.Errorf("Error retrieving fee rate for %s: %v", f.Symbol, err)
                return 0 // Do not store as last rate.
        }
+       if r <= 0 {
+               return 0
+       }
        if r > f.Asset.MaxFeeRate {
                r = f.Asset.MaxFeeRate
        }
```

client/asset/doge/doge.go
chappjc: Makes sense

chappjc:
One last thing that comes to mind is setting the explorer links in order.ts.  Maybe [dogechain.info](https://dogechain.info/tx) for mainnet and https://blockexplorer.one/dogecoin/testnet/tx for testnet

client/asset/doge/doge.go
chappjc: Seems too high for the fallback fee.  500,000 sats/byte = 5 DOGE /kB >> 0.01 DOGE / kB  https://github.com/dogecoin/dogecoin/blob/master/doc/fee-recommendation.md

dex/testing/dcrdex/harness.sh
```diff
@@ -174,6 +191,18 @@ if [ $ETH_ON -eq 0 ]; then
 EOF
 fi
 
+if [ $DOGE_ON -eq 0 ]; then
+    cat << EOF >> "./markets.json"
+         },
+        "DOGE_simnet": {
+            "bip44symbol": "doge",
+            "network": "simnet",
+            "maxFeeRate": 20,
```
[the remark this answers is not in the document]
chappjc: A bit more background on the doge fee rate saga: https://github.com/dogecoin/dogecoin/discussions/2347
Proposed relay and dust values supposedly put into effect in 1.14.4:
https://github.com/dogecoin/dogecoin/pull/2440/files

Not entirely sure what the current right min relay fee is now.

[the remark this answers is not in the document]
chappjc: What I gather from that plan and the changes implemented in https://github.com/dogecoin/dogecoin/pull/2440/files back in 1.14.4 are:

### fee rates

- `DEFAULT_BLOCK_MIN_TX_FEE` (block inclusion fee limit) is now 0.01 DOGE / kB (1,000 atom / byte)
- `DEFAULT_MIN_RELAY_TX_FEE` (node relay fee limit) is now 0.001 DOGE / kB (100 atom / byte)

Essentially, the mining code (by default, but configurable) is more demanding than relay.  Sensibly, the *recommended* rate is `DEFAULT_FALLBACK_FEE` of 0.01 DOGE/kB or 1,000 atom/B.

So, it's probably a good idea for server's maxFeeRate to be maybe 5-10x of that.  Somewhat arbitrarily, I had testnet server set to 9000.  e.g. https://blockexplorer.one/dogecoin/testnet/tx/0c661154120cc35e778d0130b790de3bbd74707ef7f2203ec4c59e99ffeda6df

There should probably be some checks that maxFeeRate is not set below this limit.  Thoughts about that?

### dust

- The **hard** dust limit is 0.001 DOGE - "outputs under this value are invalid and rejected."
- The **soft** dust limit is 0.01 DOGE - "sending a transaction with outputs under this value, are required to add 0.01 DOGE for each such output, or else will be considered to have too low fee and be rejected."

The *recommended* dust limit (discard limit) is 1.00 DOGE.

Our dust checks should be sufficient since the function takes the min relay fee as an input: `func IsDust(txOut *wire.TxOut, minRelayTxFee uint64) bool`
In calling this function, we actually pass the targeted fee rate for the transaction being authored, so _as long as we check that that rate is at least the min relay fee_ (we aren't yet), that is adequate.

In summary, it sounds like there's use for a `dex/networks/doge.MinRelayFee` variable for both server and client for the above reasons.  Client can use it internally to `client/asset` in it's `targetFeeRateWithFallback` and/or `feeRateWithFallback` methods, while server can use it when setting up the backed asset (or wherever) to ensure maxFeeRate is at least that amount.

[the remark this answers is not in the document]
chappjc: Anyway, I guess we're fine on server side for as long as the operator picks a high enough maxFeeRate.

But I think we want to change client's (fallback) `defaultFee = 500000` to `50_000` or less, and the `defaultFeeRateLimit` to `200_000` or less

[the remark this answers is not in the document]
chappjc: Yup, we definitely have an issue with `IsDust`

client/asset/doge/doge.go
chappjc: On a mainnet node running for a while

| blocks |   fee rate |
|--------|-----------:|
| 1      |         -1 |
| 2      |         -1 |
| 3      | 5.46232920 |
| 4      | 0.01003906 |
| 5      | 0.01003906 |
| 6      | 0.01003906 |
| 7      | 0.01003906 |
| 8      | 0.01003906 |
| 9      | 0.01003906 |
| 10     | 0.01003906 |

:shrug:

dex/testing/dcrdex/harness.sh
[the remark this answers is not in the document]
chappjc: Long past time for me to change gears, but I arrived here: https://github.com/chappjc/dcrdex/commit/638ded8d2c259e7ec7a46594d80c8d37e460578a

client/webserver/site/src/js/order.ts
```diff
@@ -265,5 +265,9 @@ const CoinExplorers: Record<number, Record<number, (cid: string) => string>> = {
   2: { // ltc
     [Mainnet]: (cid: string) => `https://ltc.bitaps.com/${cid.split(':')[0]}`,
     [Testnet]: (cid: string) => `https://tltc.bitaps.com/${cid.split(':')[0]}`
+  },
+  3: { // doge
+    [Mainnet]: (cid: string) => `https://dogechain.info/tx/${cid.split(':')[0]}`,
```
chappjc, line 270: I take back my suggestion to use dogechain.info. Official or not, it's not so good.  CloudFlare constantly bugging and it the site itself isn't the best.  I think https://dogeblocks.com/tx/ is better.  No external JS and it actually shows the tech details we care about.  blockchair hurts my eyes but it almost works

client/asset/doge/doge.go
chappjc: Just got this on a mainnet redeem `2022-04-14 03:47:14.101 [TRC] CORE[doge]: Obtained local estimate for 2-conf fee rate, 537604`.  After that I opened a terminal to check, and  `estimatefee` is saying `5.37140538` for confs 2 through 10 (!)  :facepalm: So turns out `estimatefee` is complete shit.  I'm not sure what we want to do about that, but I'm almost thinking we go with a static fee equal to 3-5x of the "recommended" fee of 0.01 DOGE/kB

client/asset/btc/btc.go
```diff
@@ -1021,7 +1072,7 @@ func (a amount) String() string {
 // via feeRateWithFallback.
 func (btc *baseWallet) targetFeeRateWithFallback(confTarget, feeSuggestion uint64) uint64 {
 	feeRate, err := btc.estimateFee(btc.node, confTarget)
-	if err == nil {
+	if err == nil && feeRate > 0 {
 		btc.log.Tracef("Obtained local estimate for %d-conf fee rate, %d", confTarget, feeRate)
 		return feeRate
```
chappjc, line 1077: What do you think about an inner condition for `feeRate > feeRateLimit`?  To guard against the horrible anomalous behavior of `estimatefee` where it suggests more than 5 DOGE/kB.  https://github.com/decred/dcrdex/pull/1558#discussion_r850061882

server/asset/doge/doge.go
```diff
@@ -92,14 +92,22 @@ func NewBackend(configPath string, logger dex.Logger, network dex.Network) (asse
...
 		FeeEstimator: func(cl *btc.RPCClient) (uint64, error) {
+
 			var r float64
 			if err := cl.Call("estimatefee", []interface{}{feeConfs}, &r); err != nil {
 				return 0, err
 			}
 			if r <= 0 {
 				return 0, nil
 			}
-			return uint64(math.Round(r * 1e8)), nil
```
buck54321: This was wrong, right?

chappjc: Ohhhh.... yes.  Did not notice that.

chappjc (approved):

PR #1563, opened 2022-04-05
server/swap: reduce lock contention and fix memory leak

chappjc:
Not super polished, but OK to review.

There is a lot of lock contention in the swapper, with mutexes locked during expensive operations that can time out like confirmations checks.  This starts fixing that.

There also appeared to be a memory leak with the coinlocker map with "premature" init requests from clients/

### DIFF
--- a/server/swap/swap.go
+++ b/server/swap/swap.go
@@ -382,6 +382,12 @@ func (s *Swapper) deleteMatch(mt *matchTracker) {
 	mid := mt.ID()
 	delete(s.matches, mid)
 
+	// Unlock the maker and taker order coins. May be redundant if processBlock
+	// confirmed both swaps, but premature/quick counterparty actions that
+	// advance match status first prevent that.
+	s.unlockOrderCoins(mt.Maker)
+	s.unlockOrderCoins(mt.Taker)
+
 	// Remove the match from both maker's and taker's match maps.
 	maker, taker := mt.Maker.User(), mt.Taker.User()
 	for _, user := range []account.AccountID{maker, taker} {
@@ -2078,9 +2084,6 @@ func (s *Swapper) revoke(match *matchTracker) {
 	route := msgjson.RevokeMatchRoute
 	log.Infof("Sending a '%s' notification to each client for match %v",
 		route, match.ID())
-	// Unlock the maker and taker order coins.
-	s.unlockOrderCoins(match.Taker)
-	s.unlockOrderCoins(match.Maker)
 
 	sendRev := func(mid order.MatchID, ord order.Order) {
 		msg := &msgjson.RevokeMatch{
--- a/server/swap/swap.go
+++ b/server/swap/swap.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"decred.org/dcrdex/dex"
@@ -85,6 +86,9 @@ type swapStatus struct {
 	swapAsset   uint32
 	redeemAsset uint32
 
+	swapSearching   uint32 // atomic
+	redeemSearching uint32 // atomic
+
 	mtx sync.RWMutex
 	// The time that the swap coordinator sees the transaction.
 	swapTime time.Time
@@ -104,10 +108,32 @@ func (ss *swapStatus) String() string {
 		ss.swapAsset, ss.redeemAsset, ss.swapTime, ss.swap, ss.swapConfirmed, ss.redeemTime, ss.redemption)
 }
 
+func (ss *swapStatus) startSwapSearch() bool {
+	return atomic.CompareAndSwapUint32(&ss.swapSearching, 0, 1)
+}
+
+func (ss *swapStatus) endSwapSearch() {
+	atomic.StoreUint32(&ss.swapSearching, 0)
+}
+
+func (ss *swapStatus) startRedeemSearch() bool {
+	return atomic.CompareAndSwapUint32(&ss.redeemSearching, 0, 1)
+}
+
+func (ss *swapStatus) endRedeemSearch() {
+	atomic.StoreUint32(&ss.redeemSearching, 0)
+}
+
 func (ss *swapStatus) swapConfTime() time.Time {
 	ss.mtx.RLock()
 	defer ss.mtx.RUnlock()
 	return ss.swapConfirmed
+}
+
+func (ss *swapStatus) contractState() (known, confirmed bool) {
+	ss.mtx.RLock()
+	defer ss.mtx.RUnlock()
+	return ss.swap != nil, !ss.swapConfirmed.IsZero()
 }
 
 func (ss *swapStatus) redeemSeenTime() time.Time {
@@ -897,17 +923,14 @@ func bufferedTicker(ctx context.Context, dur time.Duration) chan struct{} {
 }
 
 func (s *Swapper) tryConfirmSwap(ctx context.Context, status *swapStatus, confTime time.Time) (final bool) {
-	status.mtx.RLock()
-	if status.swapTime.IsZero() {
-		status.mtx.RUnlock()
+	if known, confirmed := status.contractState(); !known {
 		return // no swap yet to confirm
-	}
-	if !status.swapConfirmed.IsZero() {
-		status.mtx.RUnlock()
+	} else if confirmed {
 		return true // already confirmed
 	}
-	status.mtx.RUnlock() // Contract.Confirmations can timeout (slow), causing lock contention
 
+	// Swap known means status.swap is set, and that it will not be replaced
+	// because we are gating processInit with the swapSearching semaphore.
 	confs, err := status.swap.Confirmations(ctx)
 	if err != nil {
 		log.Warnf("Unable to get confirmations for swap tx %v: %v", status.swap.TxID(), err)
@@ -1324,7 +1347,7 @@ func (s *Swapper) step(user account.AccountID, matchID order.MatchID) (*stepInfo
 	}
 
 	// Verify that the user specified is the actor for this step.
-	if actor.user != user {
+	if actor.user != user { // NOTE: self-trade slips past this
 		return nil, &msgjson.Error{
 			Code:    msgjson.SettlementSequenceError,
 			Message: "expected other party to act",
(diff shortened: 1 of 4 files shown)


server/swap/swap.go
```diff
@@ -340,6 +340,12 @@ func (s *Swapper) deleteMatch(mt *matchTracker) {
 	mid := mt.ID()
 	delete(s.matches, mid)
 
+	// Unlock the maker and taker order coins. May be redundant if processBlock
+	// confirmed both swaps, but premature/quick counterparty actions that
+	// advance match status first prevent that.
+	s.unlockOrderCoins(mt.Maker)
+	s.unlockOrderCoins(mt.Taker)
```
chappjc, line 347: moved from `revoke` to `deleteMatch` so this happens in the successful completion path too (see `processRedeem` for the taker)

chappjc:
FWIW, rebased with other swapper changes from https://github.com/decred/dcrdex/pull/1541 and the trade_simnet_test.go integration tests are passing, with a -race enabled server as well.  No deadlocks or races observed.

server/swap/swap.go
```diff
@@ -1063,27 +1090,29 @@ func (s *Swapper) checkInactionEventBased() {
...
-
 	for _, match := range s.matches {
 		checkMatch(match)
 	}
+	s.matchMtx.Unlock()
 
-	for _, match := range deletions {
-		s.deleteMatch(match)
+	// Record failed matches in the DB and auth mgr, unlock coins, and send
+	// revoke_match messsages.
```
LasTshaMAN, line 1113: typo `messsages` (and in another place in this PR),

btw, might be maybe get rid of duplicated bits by something like: https://github.com/chappjc/dcrdex/pull/4

LasTshaMAN (commented):
Minor suggestion

server/swap/swap.go
chappjc: Good suggestion!  I'll get this PR out of draft early next week and will include this.

server/swap/swap.go
```diff
@@ -1033,17 +1060,17 @@ func (s *Swapper) checkInactionEventBased() {
...
 
-		failMatch := func(fault bool) {
-			s.failMatch(match, fault)
-			deletions = append(deletions, match)
+		deleteMatch := func(fault bool) {
+			s.deleteMatch(match)
+			failures = append(failures, fail{match, fault}) // to process after map delete
 		}
 
 		switch match.Status {
```
LasTshaMAN, line 1068: while reading this code, noticed [fallthrough](https://github.com/LasTshaMAN/dcrdex/blob/8c190185d64795db9b698f20cccfe004dc8e6d13/server/swap/swap.go#L1082-L1084) usage here that seems a bit awkward to me,

although these checks (for [Maker](https://github.com/LasTshaMAN/dcrdex/blob/8c190185d64795db9b698f20cccfe004dc8e6d13/server/swap/swap.go#L1077-L1077), for [Taker](https://github.com/LasTshaMAN/dcrdex/blob/8c190185d64795db9b698f20cccfe004dc8e6d13/server/swap/swap.go#L1088-L1088)) are probably not mutually exclusive, [deleteMatch](https://github.com/LasTshaMAN/dcrdex/blob/8c190185d64795db9b698f20cccfe004dc8e6d13/server/swap/swap.go#L349-L349) might be called 2 times (because of that `fallthrough`),
I've reproduces this in tests (just add 2nd call to [deleteMatch here](https://github.com/LasTshaMAN/dcrdex/blob/8c190185d64795db9b698f20cccfe004dc8e6d13/server/swap/swap_test.go#L2144-L2144) and add panic under [corresponding if](https://github.com/LasTshaMAN/dcrdex/blob/8c190185d64795db9b698f20cccfe004dc8e6d13/server/swap/swap.go#L363-L363))

I don't have enough context in this area at this point, but maybe you'd wanna take a look

chappjc: Fair enough, it's confusing so worth tweaking.

Although, `match.expiredBy` checks if *either* contract is expired, so when calling that after `fallthrough` **in `order.MakerSwapCast` status**, that is checking only the maker's existing contract expiry, not taker's since taker's does not exist yet.

Because by definition the taker's contract MUST expire before maker's (by 12 hours on mainnet), the only way in practice it would failMatch twice is if dcrdex were suspended or shutdown for a long time and then restarted, only to find BOTH the maker's existing contract is expired (20 hours old) and the expected lock time of the non-existent taker contract would be expired.  So in practice it won't happen.  For the same reason, there's really no need to check the maker's own contract for expiry since as the comment above the `fallthrough` explains and I've stated, `expectedTakerLockTime` must be well before maker's contract expiry.

This was an overzealous attempt at reducing code duplication (calling `match.expiredBy` explicitly in this `case` instead of falling through).  I'll either put a `break` after the first `failMatch`, or simply remove the `fallthrough` since as I said the `match.expiredBy` check cannot hit if the `expectedTakerLockTime` did not.

LasTshaMAN: I see, thank you, the latest commit reads much nicer!

server/swap/swap.go
```diff
@@ -874,19 +880,30 @@ func bufferedTicker(ctx context.Context, dur time.Duration) chan struct{} {
...
+		status.mtx.RUnlock()
+		return // no swap yet to confirm
+	}
+	if !status.swapConfirmed.IsZero() {
+		status.mtx.RUnlock()
+		return true // already confirmed
 	}
+	status.mtx.RUnlock() // Contract.Confirmations can timeout (slow), causing lock contention
+
 	confs, err := status.swap.Confirmations(ctx)
```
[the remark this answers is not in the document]
chappjc: That's interesting.  We certainly don't want 3700 init requests for the same match all to start coin waiters.

Regarding the `status.swap` field, we could also disallow setting it if it's not nil, since it doesn't make sense to change it once it's set.  At least that was my thought process in doing this - The `swapTime.IsZero` check is what needs to be under lock, with the assumption that `status.swap` would never be modified after being initially set.  Yeah, I think that needs to be explicitly prevented.

I'll certainly look at this more closely.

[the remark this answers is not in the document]
chappjc: I think you identified a more fundamental and nasty issue.  I've tried to address that with https://github.com/decred/dcrdex/pull/1563/commits/4d82ea4ba00b8f80e29e42451856bd97c54f1205

server/swap/swap.go
```diff
@@ -1122,33 +1154,35 @@ func (s *Swapper) checkInactionBlockBased(assetID uint32) {
 		log.Tracef("checkInactionBlockBased: asset %d, match %v (%v)",
 			assetID, match.ID(), match.Status)
 
-		failMatch := func() {
+		deleteMatch := func() {
 			// Fail the match, and assign fault if lock times are not passed.
-			s.failMatch(match, !match.expiredBy(now))
-			deletions = append(deletions, match)
+			s.deleteMatch(match)
+			failures = append(failures, fail{match, !match.expiredBy(now)})
```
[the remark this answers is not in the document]
chappjc: Yeah, and `failMatch` only looks at the Status field.  Any updates to Status are intentionally done with the matches map locked in processInit and processRedeem.  It wouldn't hurt for failMatch to use the matchTracker mutex too but it wasn't needed and still isn't I'm fairly certain.

`processBlock` -> `tryConfirmSwap` could still have a reference to the match, but it's not touching the `matchTracker.Status` field that `matchTracker.mtx` is assigned to.  In retrospect, I think we could have put the mutex in `order.Match` instead of `swap.matchTracker`.

server/swap/swap.go
```diff
@@ -874,19 +906,27 @@ func bufferedTicker(ctx context.Context, dur time.Duration) chan struct{} {
...
-	if status.swapTime.IsZero() || !status.swapConfirmed.IsZero() {
-		return
+	if known, confirmed := status.contractState(); !known {
+		return // no swap yet to confirm
+	} else if confirmed {
+		return true // already confirmed
 	}
+
+	// Swap known means status.swap is set, and that it will not be replaced
+	// because we are gating processInit with the swapSearching semaphore.
```
chappjc, line 916: Note that an easy way to just avoid a data race would have been to just use a getter method like the other swapStatus methods, but as I'm pointing out it's is completely incorrect for the swap contract field to be changed once set so I wanted to fix that issue properly.

server/swap/swap.go
```diff
@@ -1525,18 +1589,20 @@ func (s *Swapper) processInit(msg *msgjson.Message, params *msgjson.Init, stepIn
 		log.Errorf("Contract txn located after match was revoked (match id=%v, maker=%v)",
 			matchID, actor.isMaker)
 		s.respondError(msg.ID, actor.user, msgjson.ContractError, "match already revoked due to inaction")
-		return wait.DontTryAgain
+		return dontTryAgain()
 	}
 
 	actor.status.mtx.Lock()
-	actor.status.swap = contract
+	actor.status.swap = contract // swap should not be set already (handleInit should gate)
```
chappjc, line 1596: If we were not confident that the searching semaphore is sufficient to prevent this as well as simultaneous coin waiters, we could add a final nil check here before assigning, but I don't think we need to code for every conceivable bug.  Besides, I'm not sure what an appropriate way to handle that would be other than log an error or panic.

JoeGruffins (approved):

server/swap/swap.go
```diff
@@ -1863,9 +1939,14 @@ func (s *Swapper) handleInit(user account.AccountID, msg *msgjson.Message) *msgj
 	s.latencyQ.Wait(&wait.Waiter{
 		Expiration: expireTime,
 		TryFunc: func() wait.TryDirective {
-			return s.processInit(msg, params, stepInfo)
+			res := s.processInit(msg, params, stepInfo)
+			if res == wait.DontTryAgain {
+				stepInfo.actor.status.endSwapSearch() // contract now recorded
+			}
```
buck54321, line 1945: Aren't you already doing this in `processInit`? I prefer this pattern if there's a choice. Obviously a slight functional difference, but I don't know if using this pattern (waiting to `endSwapSearch` until after acks are sent) introduces any bottlenecks.

chappjc: Oh right.  The code here is how I started, but then I felt weird about waiting to unblock to send the ack and counterparty audit requests.  Lemme look again.

chappjc: I agree this is way cleaner though here.  It's why I was tempted to to add that EndFun to Waiter mentioned here: https://github.com/decred/dcrdex/pull/1563/files#diff-c456beb63edc783dd7172356d814f0e516f0f7ec8a3ba4f3b180ff24464c2afdR40
I started down that path but it got messier for some reason.

I think this PR needs to be run with loadbot a bit.

chappjc: Ugh, well that's why I moved it into `processInit`... the darn tests use the ack as the indication to move on to the next request in the test.  I'll try to figure out how to modify the tests instead of making the swapper code weird

buck54321 (approved):

PR #1564, opened 2022-04-06
webserver/site: limit config file extensions supported

### LINKED ISSUE #182: client: config file errors should be sanitized
```
/api/newwallet
```
has a parameter "inipath" any errors from this config read should be sanitized since it might expose contents of sensitive files. 

```

 curl 'http://127.0.0.1:5758/api/newwallet'  -H 'content-type: application/json'  --data '{"assetID":42,"pass":"","account":"","inipath":"/etc/passwd"}'


{"ok":false,"msg":"error creating dcr wallet: error loading wallet for 42 -\u003e dcr: error creating wallet: error parsing DCR ini file: /etc/passwd:1: malformed key=value (root:x:0:0:root:/root:/bin/bash)"}
```

ukane-philemon:
This diff limits config file extensions supported to `.conf`, `.cnf`, `.cfg`, `.ini`. closes #182


buck54321 (approved):
Works well with Nemo on Linux.

chappjc (approved):

PR #1536, opened 2022-03-21
client/asset/ltc: work with and require v0.21

chappjc:
```
This updates the client's supported Litecoin Core version to v0.21.
In addition to bumping the version, this means:
- getbalances is supported now
- sendrawtransaction has the new syntax with maxfeerate not allowHighFee

This also adds the missing switches on wallet Type to both ltc and bch.
```

Trade test with v0.21.1:

```
2022-03-21 22:56:10.049 [DBG] CORE: Swappable match 05a3f5750314c35301cd42ea1100ff474324b9a05be4bb8423cde7c4500a42bd for order 88faca82d4c56c1f816cb8b406169e9524e15af6d974bfecb786d9eb760672cb (Maker)
2022-03-21 22:56:10.059 [DBG] CORE[ltc]: 1 signature cycles to converge on fees for tx 49bc0869333daf5da6cc04bcb61186114cb093a9e9a6607735cfcfc724df5a7d: min rate = 14, actual fee rate = 32 (6075 for 189 bytes), change = false
2022-03-21 22:56:10.060 [TRC] CORE[ltc]: Obtained local estimate for 2-conf fee rate, 21
2022-03-21 22:56:10.061 [INF] CORE: Broadcasted transaction with 1 swap contracts for order 88faca82d4c56c1f816cb8b406169e9524e15af6d974bfecb786d9eb760672cb. Fee rate = 14. Receipts (ltc): [49bc0869333daf5da6cc04bcb61186114cb093a9e9a6607735cfcfc724df5a7d:0].
2022-03-21 22:56:10.062 [INF] CORE: Contract coin 49bc0869333daf5da6cc04bcb61186114cb093a9e9a6607735cfcfc724df5a7d:0 (ltc), value = 100000000, refundable at 2022-03-22 18:56:10 +0000 UTC (script = 6382012088a82075430824f7abc3e19249e0dcfb3d47d77da940f04511a2b46343c8973ce08deb8876a914e440dde176c94584ec22b86762e96ef69bfd1ccf6704ca1b3a62b17576a914e187d6e1f195e1ac33fc6a9e23effb2dd7dc517e6888ac), match = 05a3f5750314c35301cd42ea1100ff474324b9a05be4bb8423cde7c4500a42bd
[...]

Issue #1515, opened 2022-03-14
core: time stamp redeem fee suggestion so it can be updated when stale


chappjc:
https://github.com/decred/dcrdex/blob/4e161a2872d3812f624834c8448e6506a0b05a76/client/core/core.go#L6503-L6505

Using the cached redeem fee can can become an issue with a non-`FeeRater` wallet in a headless client (that probably does not have an `orderbook` subscription to receive `epoch_report` notes that power `bestBookFeeSuggestion`).  It would only do the `fee_rate` request once and never get another one, forever.

_Originally posted by @chappjc in https://github.com/decred/dcrdex/pull/1501#discussion_r818178006_

To resolve this, we should store a time stamp with the cached rate, and get fetch a new rate if the stamp is too old, maybe 5 minutes.

chappjc:
Looks like this is a problem on v0.4, perhaps more so since there aren't FeeRaters on the 0.4 release branch.

Going to look into time stamping this cached redeem rate, as well as porting #1501 with `FeeRater`s to 0.4

PR #1570, opened 2022-04-07
multi: add support for ZCash

buck54321:
Built on #1558. [Here is the diff you're looking for](https://github.com/decred/dcrdex/compare/b44030f4f5360c86d2b73f8c720499e6fac8b420...buck54321:zec)

```
Implement block deserialization, tx deserialization/serialization
and input signing for ZCash and generalize those functions in the
client and server btc packages.

Implemenation notes
  1. zcashd does not support encrypted wallets. No passwords allowed.
  2. After starting the harness, it takes a few minutes for beta to
     get caught up.
  3. zcashd can take a very long time to get it's fee estimates primed.
```


buck54321:
Might want to keep an eye on https://github.com/zcash/lightwalletd too for a possible built-in light wallet. It doesn't appear to be suitable as is, but not certain.

chappjc:
Will review shorty.  Wanna rebase away the first two commits that are already merged (the first one without the `ExchangeWalletFullNode` override)?

dex/testing/zec/harness.sh
```diff
@@ -0,0 +1,232 @@
...
+tmux send-keys -t $SESSION:1 "cd ${BETA_DIR}" C-m
+
+echo "Starting simnet beta node"
+tmux send-keys -t $SESSION:1 "${DAEMON} -rpcuser=user -rpcpassword=pass \
+  -rpcport=${BETA_RPC_PORT} -datadir=${BETA_DIR} -txindex=1 -regtest=1 \
+  -debug=rpc -debug=net -debug=mempool -debug=walletdb -debug=addrman -debug=mempoolrej \
+  -whitelist=127.0.0.0/8 -whitelist=::1 \
+  -port=${BETA_LISTEN_PORT} -fallbackfee=0.00001 -printtoconsole; \
+  tmux wait-for -S beta${SYMBOL}" C-m
+sleep 3
```
martonp, line 126: I had to increase this to `sleep 30` because the rpc calls below would fail because the client was still "warming up".

martonp (commented):
Looks good, I was able to do a swap on simnet. Haven't tested with testnet yet. For some reason the dex client is constantly sending `getbestblockhash`, `getnetworkinfo`, and `getblockchaininfo` calls to the zcash client when I'm running on simnet. It seems a bit excessive.. there are at least 2-3 calls every second.

client/asset/zec/sign.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	var h chainhash.Hash
+	preImage := sigHash.Bytes()
+	sigHashKey := append([]byte(blake2BSigHash), cver[:]...)
+	if h, err = blake2bHash(preImage, sigHashKey); err != nil {
+		return nil, nil, err
+	}
+
+	return preImage, h.CloneBytes(), nil
```
chappjc, line 186: Shouldn't we just slice the hash directly instead of using `CloneBytes`.  i.e. `return preImage, h[:], nil`
Same Q above in the `SigHashSingle` conditional

client/asset/zec/sign.go
```diff
@@ -0,0 +1,256 @@
...
+
+	// nExpiryHeight
+	var expiryTime [4]byte
+	binary.LittleEndian.PutUint32(expiryTime[:], tx.ExpiryHeight)
+	sigHash.Write(expiryTime[:])
+
+	// valueBalance
+	if tx.Version >= dexzec.VersionSapling {
+		var valueBalance [8]byte
+		binary.LittleEndian.PutUint64(valueBalance[:], 0)
```
chappjc, line 147: Peculiar.  Always zero?  Looks like this block would be effectively `sigHash.Write(eightZeros[:])` i.e. removing this `PutUint64` that would seem to be a no-op

client/asset/zec/sign.go
```diff
@@ -0,0 +1,256 @@
...
+		sigHash.Write(valueBalance[:])
+	}
+
+	// hash type
+	var bHashType [4]byte
+	binary.LittleEndian.PutUint32(bHashType[:], uint32(hashType))
+	sigHash.Write(bHashType[:])
+
+	if idx != math.MaxUint32 {
```
chappjc, line 156: More `idx` strangeness.  The top of this function screens this out already, ensuring this code always executes, right?

client/asset/zec/sign.go
```diff
@@ -0,0 +1,256 @@
...
+		binary.LittleEndian.PutUint32(bIndex[:], tx.TxIn[idx].PreviousOutPoint.Index)
+		sigHash.Write(bIndex[:])
+
+		// scriptCode
+		if err = wire.WriteVarBytes(&sigHash, 0, subScript); err != nil {
+			return nil, nil, err
+		}
+
+		// value
+		if err = binary.Write(&sigHash, binary.LittleEndian, amt); err != nil {
```
chappjc, line 169: While `binary.Write` is shorter syntax, it's internally the same as what the rest of the function does manually with an array and `PutUint64`.

chappjc (commented):
Some thoughts on sign.go to start

dex/networks/zec/tx.go
```diff
@@ -0,0 +1,417 @@
...
+	// overWintered := (ver & (1 << 31)) > 0
+	ver &= overwinterMask // Clear the overwinter bit
+	tx.Version = int32(ver)
+
+	if ver > 4 {
+		return fmt.Errorf("unsupported tx version %d > 4", ver)
+	}
+
+	if ver > 3 {
+		_, err := readUint32(r)
```
chappjc, line 156: No need to declare new shadowing `err` here

dex/networks/zec/tx.go
```diff
@@ -0,0 +1,417 @@
...
+	if ver > 3 {
+		tx.ExpiryHeight, err = readUint32(r)
+		if err != nil {
+			return err
+		}
+	}
+
+	// Empty the rest of the buffer.
+	if ver < 2 {
+		return nil
```
chappjc, line 204: `// Empty the rest of the buffer.` meaning the rest of the buffer should be empty?  Or we should read and discard the rest before returning?  Or just, "done".

dex/networks/zec/tx.go
```diff
@@ -0,0 +1,417 @@
...
+			return fmt.Errorf("error reading joinSplitPubKey: %w", err)
+		}
+
+		var joinSplitSig [64]byte
+		if _, err := io.ReadFull(r, joinSplitSig[:]); err != nil {
+			return fmt.Errorf("error reading joinSplitSig: %w", err)
+		}
+	}
+
+	if ver == 4 && bindingSigRequired {
```
chappjc, line 258: Really not just `if bindingSigRequired {`?  Is it really intended that version 5 could set `bindingSigRequired = true`, not to hit this condition and read and discard these bytes?

dex/networks/zec/tx.go
```diff
@@ -0,0 +1,417 @@
...
+
+	}
+
+	if nJoinSplit, err := wire.ReadVarInt(r, pver); err != nil {
+		return fmt.Errorf("error reading nJoinSplit: %w", err)
+	} else if nJoinSplit > 0 {
+		// vJoinSplit - discard
+		sz := 1802 * nJoinSplit
+		if ver == 4 {
+			sz = 1698 * nJoinSplit
```
chappjc, line 242: Curious about the meaning of these magics.

Also, this function does a lot of discarding of fairly large slices of data, so to simply seek forward without allocating a junk slice, I'd consider using `buf := bytes.NewBuffer(blk)`, and then `buf.Next(N)` to throw out `N` bytes.  But you'd have to check the length of the returned slice.  There's also a `bufio.Reader` with a `Discard` method that returns an error, but I've not used that one and it might be inefficient for this purpose.

Another thought on that is that a `*bytes.Reader` (the concrete type, not the `io.Reader` interface) such as the one created by `DeserializeZecTx` is also an `io.ReedSeeker` that would just give us the `Seek` method, but that makes the `ZecDecode` input more specialized (e.g. the `io.Reader` returned by `hex.NewDecoder` is not also a `Seeker`).

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/btcsuite/btcwallet/walletdb v1.4.0
 	github.com/btcsuite/btcwallet/wtxmgr v1.3.0
 	github.com/davecgh/go-spew v1.1.1
+	github.com/dchest/blake2b v1.0.0
```
chappjc, line 16: I'm wondering about golang.org/x/crypto/blake2b.  It's maintained and has architecture specific implementations and assembly (may or may not be viewed a good thing though)

chappjc (commented):
dex/networks/zec

dex/networks/zec/params.go
```diff
@@ -0,0 +1,86 @@
...
+	// + 1 varint nOutputsSapling + 1 varint nJoinSplit
+	MinimumTxOverhead = 29
+
+	InitTxSizeBase = MinimumTxOverhead + btc.P2PKHOutputSize + btc.P2SHOutputSize // 29 + 34 + 32 = 95
+	InitTxSize     = InitTxSizeBase + btc.RedeemP2PKHInputSize                    // 95 + 149 = 244
+)
+
+var (
+	UnitInfo = dex.UnitInfo{
+		AtomicUnit: "Sats",
```
chappjc, line 26: Is it a "Zat"/zatoshi? I see stuff like `amountZat`, `inZat`, `valueZat` in RPCs and the zcashd code

https://github.com/zcash/zcash/commit/1805466d5c555778017bdb165688fd699bcbcdbb

dex/networks/zec/block_test.go
```diff
@@ -0,0 +1,191 @@
...
+		"9d3ac93ace45d86fca9a198913d8720805e4d77b0782a7e355ae4067856e0b7d7e462f" +
+		"7ba21517fc89e5561adb2129bbc8737d786f0f90df606a154955b845889283b74704ce" +
+		"c758fbce1bc9713c2f170569962567e27dcdf7f5f4bdee1ac2c3d877988e0104000080" +
+		"85202f8901000000000000000000000000000000000000000000000000000000000000" +
+		"0000ffffffff210387c9181c4d696e656420627920416e74506f6f6c36323361001f03" +
+		"20a3c16b65ffffffff0438c94d010000000017a914df37da178caa3ba0cc7feea71cd8" +
+		"63c6677f4ea587286bee000000000017a914d45cb1adffb5215a42720532a076f02c7c" +
+		"778c908740787d010000000017a914931fec54c1fea86e574462cc32013f5400b89129" +
+		"8780b2e60e000000001976a9145c38e5e20b62bbb5683dd68677ac3715047341ea88ac" +
+		"00000000000000000000000000000000000000")
```
chappjc, line 60: I'd have crack at `go:embed` for the blocks.  See my previous doge block deserializer PR and the current ltc 0.21 pr https://github.com/decred/dcrdex/pull/1536

You can write straight into a .dat file suitable for a `[]byte` var using `xxd`:
`./zcash-cli getblock <hash> 0 | xxd -p -r - blockX.dat`

client/asset/btc/btc.go
```diff
@@ -2350,13 +2416,48 @@ func (btc *baseWallet) AuditContract(coinID, contract, txData dex.Bytes, rebroad
...
 	}, nil
 }
 
+// AuditContract retrieves information about a swap contract from the provided
+// txData. The extracted information would be used to audit the counter-party's
+// contract during a swap. Since this wallet is backed by a full node, txData
+// may be empty to attempt retrieval of the transaction output from the network.
+func (btc *ExchangeWalletFullNode) AuditContract(coinID, contract, txData dex.Bytes, rebroadcast bool) (*asset.AuditInfo, error) {
+	if len(txData) != 0 {
+		return btc.baseWallet.AuditContract(coinID, contract, txData, rebroadcast)
```
chappjc, line 2432: I think this `ExchangeWalletFullNode` method was left over from rejected code from a [previous PR of mine](https://github.com/decred/dcrdex/pull/1553#discussion_r843386210) (on iteration https://github.com/decred/dcrdex/commit/bae9942753de9d56596434db0c34ce89da7faeec).  I guess this zcash PR was based on top of it at one point, and then you rebased the whole branch on master instead of using `--onto` and just your commits.

Or did this method actually turn out to be helpful?

chappjc (changes requested):

client/asset/zec/sign.go
```diff
@@ -0,0 +1,256 @@
...
+
+		sigHash.Write(sigHashes.HashSequence[:])
+	} else {
+		sigHash.Write(zeroHash[:])
+	}
+
+	// hashOutputs
+	if hashType&sigHashMask != txscript.SigHashSingle && hashType&sigHashMask != txscript.SigHashNone {
+		sigHash.Write(sigHashes.HashOutputs[:])
+	} else if hashType&sigHashMask == txscript.SigHashSingle && idx < len(tx.TxOut) {
```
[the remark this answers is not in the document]
buck54321: See e.g. https://wiki.bitcoinsv.io/index.php/SIGHASH_flags

**SIGHASH_SINGLE**: Sign all inputs and the output with the same index

client/asset/zec/sign.go
buck54321: Oh yeah. Hmm.

buck54321:
Updated for NU5. I expect that ZCash won't be live on DEX when the upgrade goes live on May 31st, but we're ready-ish for the upgrade if it is. For a ZCash upgrade, an operator should be expected to shut down for the upgrade, since old version txs won't be accepted on the new chain, per [this upgrade guide](https://zcash.readthedocs.io/en/latest/rtd_pages/nu_dev_guide.html#canopy).

I'm syncing testnet and then I'll search for some NU5 txs to use for tests.

dex/networks/zec/tx.go
```diff
@@ -0,0 +1,611 @@
...
+	b := make([]byte, count)
+	_, err = io.ReadFull(r, b)
+	if err != nil {
+		return nil, err
+	}
+	return b, nil
+}
+
+func discardBytes(r io.Reader, n int64) error {
+	m, err := io.CopyN(io.Discard, r, n)
```
chappjc, line 567: Nice, no need for a buffer or other type, just a reader. io.Discard didn't cross my mind

go.mod
buck54321: I did look try this, and couldn't get tests to pass using our personalization keys. Not sure if key handling is different somehow.

dex/networks/zec/tx.go
```diff
@@ -0,0 +1,417 @@
...
+	return b, nil
+}
+
+// CalcTxSize calculates the size of a ZCash transparent transaction. CalcTxSize
+// won't return accurate results for shielded or blended transactions.
+func CalcTxSize(tx *wire.MsgTx) uint64 {
+	var sz uint64 = 4 // header
+	ver := tx.Version
+	if ver > 3 {
+		sz += 4 // nVersionGroup
```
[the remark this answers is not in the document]
buck54321: I think the tx version issues are resolved now.

dex/networks/zec/tx.go
buck54321: I guess I just meant to discard the rest before returning. I've removed the confusing message now.

buck54321:
> For some reason the dex client is constantly sending `getbestblockhash`, `getnetworkinfo`, and `getblockchaininfo` calls to the zcash client when I'm running on simnet. It seems a bit excessive.. there are at least 2-3 calls every second.

I put some feelers in, and it seems to me like the calls to `getbestblockhash` are our normal block polling. I'm not sure why we're seeing two log messages for each in the zcashd/bitcoind terminals. Might be worth looking into further, but I don't see any issues on the client side.

The `getnetworkinfo` is coming from our `monitorPeers` loop.

I'm not seeing a lot of `getblockchaininfo`, but that would be expected with a new block.

I do see some `estimatesmartfee` pretty regularly from `(*Core).tick` → `(*Core).cacheRedemptionFeeSuggestion`. Related: #1515.

go.mod
chappjc: Oh... https://github.com/golang/go/issues/32447
personalization != MAC key

chappjc (approved):
Looks great!  :rocket: :lock: :fireworks: 

Just need a resolve.

JoeGruffins (commented):
Wow. Thats a lot of zcash. My eyes started to glaze over at `tx.go`

I've not been able to trade so far with :
```
2022-05-19 23:23:46.860 [ERR] CORE: notify: |ERROR| (order) Swap send error - Error encountered sending a swap output(s) worth 1.00000000 ZEC on order 117411dd - Order: 117411dd464a5b5f40e3fc40c32f5d45cf0fbc88d1fa38fca69ab9ec668b1090
2022-05-19 23:23:46.860 [ERR] CORE: 127.0.0.1:17273 tick: {swapMatches order 117411dd464a5b5f40e3fc40c32f5d45cf0fbc88d1fa38fca69ab9ec668b1090 - {error sending swap transaction: signing error: tx signing error: rawrequest (signrawtransaction) error: -22: TX decode failed, raw tx: 0500000001ce67eab2739d1db57d16dcc8457cd0f1028c424459a5052ef4d65ca59cf841150100000000ffffffff0100e1f5050000000017a914c83a74347ff867cade02a48197b39345b8635b018700000000}}
```

I also attempted to get the core simnet tests running but they failed. Can do in another pr though.

Noticed a couple things maybe unrelated to this pr.

If I start to make a ZEC wallet but I put in a password, it errors, which is fine. But if I close that box then I see this for the wallet:
image

And I can click on that box to unlock the wallet that isnt there.

Also, I was not able to pay the fee with bitcoin for the simnet tests. May be a problem with the simnet tests.

```
$ zcashd --version
Zcash Daemon version v5.0.0
```

dex/networks/btc/clone.go
```diff
@@ -12,6 +12,11 @@ import (
 // AddressDecoder decodes a string address to a btcutil.Address.
 type AddressDecoder func(addr string, net *chaincfg.Params) (btcutil.Address, error)
 
+// AddressStringer converts the btcutil.Address into a string address. This may
+// be different from using (btcutil.Address).String when there is a special
+// encoding such as a BCH "Cash Address".
+type AddressStringer func(btcutil.Address, *chaincfg.Params) (string, error)
```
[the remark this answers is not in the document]
chappjc: "**such as** BCH"
;)

chappjc:
Oh! zcash 5.0 was released a week ago.  https://github.com/zcash/zcash/releases/tag/v5.0.0  I've been working with 4.6

dex/testing/zec/harness.sh
```diff
@@ -0,0 +1,232 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
buck54321: I keep doing it, but git's not remembering for some reason. Hmm.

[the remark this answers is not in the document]
chappjc: You can cherry-pick https://github.com/decred/dcrdex/commit/d1ec7f4d73615a0bde695fc9141a33e0b2d3851b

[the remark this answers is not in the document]
buck54321: Lemme try `git update-index --chmod=+x <file>`.

go.mod
```diff
@@ -51,6 +52,8 @@ require (
 	gopkg.in/ini.v1 v1.62.0
 )
 
+require github.com/decred/base58 v1.0.3
```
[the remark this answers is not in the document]
chappjc: There's also a v1.0.4 https://github.com/decred/base58/releases/tag/v1.0.4

client/asset/zec/sign.go
```diff
@@ -0,0 +1,248 @@
...
+	if err != nil {
+		return err
+	}
+	w.Write(h[:])
+	return nil
+}
+
+func validateSigHashType(v txscript.SigHashType) bool {
+	return v == txscript.SigHashAll || v == txscript.SigHashNone || v == txscript.SigHashSingle ||
+		v == txscript.SigHashAnyOneCanPay || v == 0x82 || v == 0x83
```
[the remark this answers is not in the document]
chappjc: Would that be the bit-or of SIGHASH_ANYONECANPAY with either SIGHASH_NONE or SIGHASH_SINGLE?

Now that I say that, I think we're missing 0x81 too?

chappjc:
How about https://zcashblockexplorer.com/ for the mainnet explorer? No external JS, displays fine with no JS, seems to be reference deployment of https://github.com/nighthawk-apps/zcash-explorer, has an onion address.

I don't see a good testnet option though

dex/networks/btc/clone.go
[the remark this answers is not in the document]
JoeGruffins: oh, has such as, srry

go.mod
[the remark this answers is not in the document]
chappjc: Sorry to harp on this, but we should really use v1.0.4.  Thanks for moving it up though.

dex/networks/zec/tx.go
```diff
@@ -0,0 +1,896 @@
...
+// see https://zips.z.cash/protocol/protocol.pdf section 7.1
+func DeserializeTx(b []byte) (*Tx, error) {
+	tx := &Tx{MsgTx: new(wire.MsgTx)}
+	r := bytes.NewReader(b)
+	if err := tx.ZecDecode(r); err != nil {
+		return nil, err
+	}
+
+	remains, _ := io.ReadAll(r)
+	if len(remains) > 0 {
```
chappjc, line 437: I believe this is `remains := r.Len() > 0`, so no need for the `ReadAll` to determine this.

dex/networks/zec/tx.go
```diff
@@ -471,19 +725,33 @@ func writeOutPoint(w io.Writer, version int32, op *wire.OutPoint) error {
...
+func uint64Bytes(v uint64) []byte {
 	b := make([]byte, 8)
 	binary.LittleEndian.PutUint64(b, v)
-	_, err := w.Write(b)
+	return b
+}
+
+func int64Bytes(v int64) []byte {
+	buf := new(bytes.Buffer)
+	binary.Write(buf, binary.LittleEndian, v)
```
chappjc, line 748: This just does `order.PutUint16(bs, uint16(v))` internally, with it's own slice alloc.  I'd do it directly to skip the extra alloc, the err return, and overhead of `Write`.

dex/networks/zec/tx.go
```diff
@@ -471,19 +725,33 @@ func writeOutPoint(w io.Writer, version int32, op *wire.OutPoint) error {
...
+	return buf.Bytes()
+}
+
+// putUint32 writes a little-endian encoded uint64 to the Writer.
+func putUint64(w io.Writer, v uint64) error {
+	_, err := w.Write(uint64Bytes(v))
```
chappjc, line 754: Doc says 32.

dex/networks/zec/tx.go
```diff
@@ -471,19 +725,33 @@ func writeOutPoint(w io.Writer, version int32, op *wire.OutPoint) error {
 	return putUint32(w, op.Index)
```
chappjc, line 725: `version` input seems to have no use with `writeOutPoint` and `writeTxIn`.

chappjc (changes requested):

chappjc:
#1597 made some conflicts

JoeGruffins (commented):
This server log seems off: `Blocks are too empty to calculate %!d(string=zec) median fees. Using no-competition rate.`

Also, I can't seem to get refunds to happen. Is client checking the locktime correctly?

It looks like will need a couple more wallets for simnet tests just like doge. Would you like to add or shall I?

chappjc (approved):
OK.  Just one silly conflict with the new `txscript.NewTxSigHashes` function signature.  Also @JoeGruffins review

dex/testing/dcrdex/harness.sh
```diff
@@ -130,6 +133,20 @@ EOF
...
 fi
 
+if [ $ZEC_ON -eq 0 ]; then
+    cat << EOF >> "./markets.json"
+        },
+        {
+            "base": "ZEC_simnet",
+            "quote": "BTC_simnet",
+            "lotSize": 100000000,
+            "rateStep": 100000,
```
chappjc, line 143: This rate step is too high to reflect realistic rates.  No a big deal on simnet, just not realistic.
Currently it's 0.00293146 BTC/ZEC, but I can only use 0.002 or 0.001

chappjc:
> Also, I can't seem to get refunds to happen. Is client checking the locktime correctly?

@JoeGruffins is right.  It can never refund.  It seems the reason is that the `getblockheader` response does not include `MedianTime`, so it is always treated as 0 in `(*baseWallet).LocktimeExpired` and thus cannot ever be after contract expiry.

```
$  ./alpha getblockheader 0b5d2de16ab1f74674f4ad52f2096d90a293c61b9e552c49433a33fafc9541a9
{
  "hash": "0b5d2de16ab1f74674f4ad52f2096d90a293c61b9e552c49433a33fafc9541a9",
  "confirmations": 1,
  "height": 427,
  "version": 4,
  "merkleroot": "0abafddb3bd8053a633e64a50e7876b663371c328e7f31036524a23dccf5fb4e",
  "finalsaplingroot": "3e49b5f954aa9d3545bc6c37744661eea48d7c34e3000d82b7f0010c30f4c2fb",
  "time": 1654015360,
  "nonce": "00005ab9fbe1195127e17d82a2929c18e0c5037e44836c4a2d3ccdf7ef2c002e",
  "solution": "0c6203f9b1fd8888ab255da63a4435b577c910339676e1a30d45cc3cba2fd9f4b45e95b6",
  "bits": "200f0f0f",
  "difficulty": 1,
  "chainwork": "0000000000000000000000000000000000000000000000000000000000001c6c",
  "previousblockhash": "05a282582126e141821e885428d852e5756a471bacc122aec282392fcca4cc2b"
}
```

chappjc:
I suppose we'll just have to pull the last 11 block headers to compute mediantime.  The only RPCs I can see that might give it to use is `getblockdeltas`, which requires some crazy switches like `-insightexplorer`, or `getblocktemplate`->`mintime`-1, and it's probably a terrible idea to use mining functions.

Issue #1574, opened 2022-04-08
client UI and backend: server setting and dialog


chappjc:
We will need backend methods to change various client settings pertaining to configured DEX servers, and a UI redesign to collect all operations related to a given server.

This would house the existing export/disable buttons, as well as the following new features:
- updating a server's TLS certificate https://github.com/decred/dcrdex/issues/1571
- changing a server's host name e.g. to/from regular internet and .onion https://github.com/decred/dcrdex/issues/1569
- configuring fidelity bond settings such as tier level to maintain, showing a list of active bonds, etc. https://github.com/decred/dcrdex/pull/1480

I suggest starting a UI redesign with these things in mind.  It will take some time to implement the backend bits described above, so let's just plan on needing a place for them.

martonp:
I'll start on this as well since I'm working on #1571. I'm thinking to put a settings gear icon next to the server address on the settings page. Clicking the gear will open a new page which will have all the settings specific to that server.

martonp:
@chappjc I'm looking into updating the server's host name from regular internet to .onion. Currently the host name is used as the identifier in the client to identify the server, and the host name is stored with both the orders and matches. Since the host name can change, the server's public key should be the identifier for a server, and the client db needs to be updated to use the public key as the foreign key in the match/order buckets. So, there must be an upgrade process for the database when the user's upgrade their client. Do you agree with this, or do you think there's an easier way?

chappjc:
> Since the host name can change, the server's public key should be the identifier for a server, and the client db needs to be updated to use the public key as the foreign key in the match/order buckets. So, there must be an upgrade process for the database when the user's upgrade their client. Do you agree with this, or do you think there's an easier way?

Yes, related: Related: https://github.com/decred/dcrdex/issues/433

In the sense of account recovery, that issue was closed.  However, in terms of the client's operation, your assessment is correct that we're keying on the host name when we probably shouldn't.

That said, my feeling is that changing that is not worth the effort and upheaval.  Since a host name change can be worked around by disabling the existing account that's keyed by host name A and re-registering (via account discovery and no fee) with host name B as long as they share the same public key, I think there needs to be a strong reason to resolve this the hard way. (As an aside, we should perhaps see what happens when you try to add the same dex with two host names simultaneously, like is possible with dex-test, and see if there need to be any sanity checks to prevent it.)

You could briefly investigate this in case it's not as extensive as it seems.  Or if perhaps there's a more compelling reason to go to those lengths.

[...]

PR #1575, opened 2022-04-10
hidden service / .onion dex support

chappjc:
This introduces special handling for dex hosts with .onion host names.

Deployed to the dex-test.ssgen.io (**testnet**) server under the host name **dextesttaggpdj27s26wklczemovhqdzyvgokyumey33uemlw4imawqd.onion**.

No TLS is used with the dedicated .onion listener, and as such the client should not specify a TLS certificate.  The onion protocol provides end-to-end encryption and identification via the ed25512 key pair.  The hostname itself is all the certification that is needed AFAICT.

You can either make a new account or disable any existing dex-test.ssgen.io account and add the .onion host in its place.  If you had an account already, account discovery works as expected despite the changed host name (server is identified by it's signing key pair, not the TLS junk used for plain internet connections).

Bad thinks are likely to happen if you have an active dex-test.ssgen.io account and you try to add it again under the .onion host without first disabling the dex-test.ssgen.io version.  Related issue: https://github.com/decred/dcrdex/issues/1574

This PR is mainly the backend bits, but it all works fine via the browser, albeit with some ugly display:

image

image

A related change to server that I am planning is a keypair *slice* so that the TLS listener can server **virtual hosts**.

### DIFF
--- a/server/comms/server.go
+++ b/server/comms/server.go
@@ -164,6 +164,14 @@ func RegisterHTTP(route string, handler HTTPHandler) {
 // The RPCConfig is the server configuration settings and the only argument
 // to the server's constructor.
 type RPCConfig struct {
+	// HiddenServiceAddr is the local address to which connections from the
+	// local hidden service will connect, e.g. 127.0.0.1:7252. This is not the
+	// .onion address of the hidden service. The TLS key pairs do not apply to
+	// these connections since TLS is not used on the hidden service's listener.
+	// This corresponds to the last component of a HiddenServicePort line in a
+	// torrc config file. e.g. HiddenServicePort 7232 127.0.0.1:7252. Clients
+	// would specify the port preceding this address in the above statement.
+	HiddenServiceAddr string
 	// ListenAddrs are the addresses on which the server will listen.
 	ListenAddrs []string
 	// The location of the TLS keypair files. If they are not already at the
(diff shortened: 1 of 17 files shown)


server/comms/server.go
```diff
@@ -353,7 +388,9 @@ func (s *Server) Run(ctx context.Context) {
 		Handler:      mux,
 		ReadTimeout:  rpcTimeoutSeconds * time.Second, // slow requests should not hold connections opened
 		WriteTimeout: rpcTimeoutSeconds * time.Second, // hung responses must die
-		//BaseContext:  func(net.Listener) context.Context { return ctx },
+		BaseContext: func(l net.Listener) context.Context {
+			return context.WithValue(ctx, ctxListener, &onionListener{l}) // the actual listener is not really useful, maybe drop it
```
martonp, line 392: Shouldn't the value just be `l` instead of `&onionListener{l}`?

martonp (commented):
Nice, I was able to connect.

chappjc:
In preparation to port this onto 0.4.3, together with **LTC segwit**:

image

chappjc:
Well, there's plenty of room to expand on this, but it's working.

buck54321 (approved):
I would add these changes for some UI sanity. 
https://github.com/chappjc/dcrdex/compare/toriness...buck54321:review?expand=1

chappjc:
Much appreciated @buck54321

martonp (approved):

PR #1578, opened 2022-04-11
test: use `T.TempDir` to create temporary test directory

Juneezee:
A testing cleanup.

This pull request replaces `os.MkdirTemp` with `t.TempDir`. We can use the `T.TempDir` function from the `testing` package to create temporary directory. The directory created by `T.TempDir` is automatically removed when the test and all its subtests complete. 

Reference: https://pkg.go.dev/testing#T.TempDir

### DIFF
--- a/client/core/trade_simnet_test.go
+++ b/client/core/trade_simnet_test.go
@@ -81,8 +81,7 @@ var (
 	tLockTimeTaker = 30 * time.Second
 	tLockTimeMaker = 1 * time.Minute
 
-	tLog   = dex.StdOutLogger("TEST", dex.LevelTrace)
-	tmpDir string
+	tLog = dex.StdOutLogger("TEST", dex.LevelTrace)
 )
 
 func readWalletCfgsAndDexCert() error {
@@ -128,9 +127,9 @@ func readWalletCfgsAndDexCert() error {
 	return err
 }
 
-func startClients(ctx context.Context) error {
+func startClients(t testing.TB, ctx context.Context) error {
 	for _, c := range clients {
-		err := c.init(ctx)
+		err := c.init(t, ctx)
 		if err != nil {
 			return err
 		}
@@ -197,13 +196,13 @@ func startClients(ctx context.Context) error {
 	return nil
 }
 
-func setup() (context.CancelFunc, error) {
+func setup(t testing.TB) (context.CancelFunc, error) {
 	err := readWalletCfgsAndDexCert()
 	if err != nil {
 		return func() {}, fmt.Errorf("error reading wallet cfgs or dex cert (harnesses running?): %w", err)
 	}
 	ctx, cancelCtx := context.WithCancel(context.Background())
-	err = startClients(ctx)
+	err = startClients(t, ctx)
 	if err != nil {
 		return cancelCtx, fmt.Errorf("error starting clients: %w", err)
 	}
@@ -216,12 +215,6 @@ func teardown(cancelCtx context.CancelFunc) {
 		c.wg.Wait()
 		c.log("Client %d done", c.id)
 	}
-	if client1.core != nil && client1.core.cfg.DBPath != "" {
-		os.RemoveAll(client1.core.cfg.DBPath)
-	}
-	if client2.core != nil && client2.core.cfg.DBPath != "" {
-		os.RemoveAll(client2.core.cfg.DBPath)
-	}
 }
 
 var sleepFactor time.Duration = 1
@@ -230,8 +223,6 @@ func TestMain(m *testing.M) {
 	if race {
 		sleepFactor = 3
 	}
-	tmpDir, _ = os.MkdirTemp("", "")
-	defer os.RemoveAll(tmpDir)
 	os.Exit(m.Run())
 }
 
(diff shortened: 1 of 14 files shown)


chappjc:
Thanks for this!  Looks like a well crafted PR, much appreciated.
Only thing is that you mention `ioutil.TempDir` but we [ditched ioutil recently](https://github.com/decred/dcrdex/commit/b294a00be0960821d131357a3ad19adde4e7cbab), so the commit message is inaccurate.

JoeGruffins (approved):
Wow! Can't believe I didn't know this existed. See two more usages of MkdirTemp
```
$ grep -rn MkdirTemp
client/db/bolt/db_test.go:64:           tDir, err = os.MkdirTemp("", "dbtest")
client/core/trade_simnet_test.go:233:   tmpDir, _ = os.MkdirTemp("", "")
```

Juneezee:
> Wow! Can't believe I didn't know this existed. See two more usages of MkdirTemp
> 
> ```
> $ grep -rn MkdirTemp
> client/db/bolt/db_test.go:64:           tDir, err = os.MkdirTemp("", "dbtest")
> client/core/trade_simnet_test.go:233:   tmpDir, _ = os.MkdirTemp("", "")
> ```

@JoeGruffins  These two cannot be replaced because they are both under `TestMain(m *testing.M)`

```go
func TestMain(m *testing.M) {
	if race {
		sleepFactor = 3
	}
	tmpDir, _ = os.MkdirTemp("", "")
	defer os.RemoveAll(tmpDir)
	os.Exit(m.Run())
}
```

```go
func TestMain(m *testing.M) {
	defer os.Stdout.Sync()
	doIt := func() int {
		var err error
		tDir, err = os.MkdirTemp("", "dbtest")
		if err != nil {
			fmt.Printf("error creating temporary directory: %v\n", err)
			return -1
		}
		defer os.RemoveAll(tDir)
		return m.Run()
	}
	os.Exit(doIt())
}
```

JoeGruffins:
Oh, sorry. I should have looked closer at where they were.

> After a second look, I think they can be replaced.

Really?

JoeGruffins:
Actually, both the tests look to use a counter to increment database names. So, don't need to share and should probably also be changed. imo

JoeGruffins:
Thank you!

I'm pretty sure you can replace all the `fmt.Sprintf("dex_%d.db", cNum)` and `fmt.Sprintf("db%d.db", tCounter)` with just `dex.db` and `db.db` or whatever now and there's no need for the increment variable. They can be the same name now as they live in different folders, correct?

Juneezee:
@JoeGruffins Yup! You are correct.

client/core/trade_simnet_test.go
```diff
@@ -197,13 +196,13 @@ func startClients(ctx context.Context) error {
...
 
-func setup() (context.CancelFunc, error) {
+func setup(t testing.TB) (context.CancelFunc, error) {
 	err := readWalletCfgsAndDexCert()
 	if err != nil {
 		return func() {}, fmt.Errorf("error reading wallet cfgs or dex cert (harnesses running?): %w", err)
 	}
 	ctx, cancelCtx := context.WithCancel(context.Background())
-	err = startClients(ctx)
+	err = startClients(t, ctx)
```
chappjc, line 205: Please see the `teardown` function about 7 lines down from here.   Those `...cfg.DBPath` fields are what we're now creating in `init` with the new `t.TempDir()`.  I believe what was removed in `TestMain` was really just for the temp _folder_ assuming each test cleaned up their actual DB file via `teardown`.  If each `t.TempDir` works as intended, `teardown` won't need those `os.RemoveAll` anymore.
I will test however.

client/core/trade_simnet_test.go
```diff
@@ -1358,11 +1355,12 @@ func (client *tClient) log(format string, args ...interface{}) {
...
 var clientCounter uint32
 
-func (client *tClient) init(ctx context.Context) error {
+func (client *tClient) init(t testing.TB, ctx context.Context) error {
+	tmpDir := t.TempDir()
 	cNum := atomic.AddUint32(&clientCounter, 1)
 	var err error
 	client.core, err = New(&Config{
-		DBPath: filepath.Join(tmpDir, fmt.Sprintf("dex_%d.db", cNum)),
+		DBPath: filepath.Join(tmpDir, "dex_db"),
```
chappjc, line 1363: dex_db not dex.db?  I guess this doesn't matter, but let's use an extension.

client/core/trade_simnet_test.go
Juneezee: I have removed those `os.RemoveAll` lines in `teardown` function.

`t.TempDir` ensures that the temporary directory is removed when the test finishes. It will fail the test if it could not delete the directory https://github.com/golang/go/blob/0c6d8bb109adfa188190b2e6de59b88853ded378/src/testing/testing.go#L1099

chappjc (approved):

client/db/bolt/db_test.go
```diff
@@ -59,17 +56,7 @@ func newTestDB(t *testing.T) (*BoltDB, func()) {
...
-		if err != nil {
-			fmt.Printf("error creating temporary directory: %v\n", err)
-			return -1
-		}
-		defer os.RemoveAll(tDir)
-		return m.Run()
-	}
-	os.Exit(doIt())
+	os.Exit(m.Run())
 }
```
JoeGruffins, line 60: I think you can just remove `TestMain` now. I'm unsure of what `os.Stdout.Sync()` accomplishes... Or you can leave it I guess, don't think it hurts anything.

JoeGruffins (approved):

PR #1581, opened 2022-04-12
client/asset/{btc,dcr}: support p2pk and fix unsupported btc utxos

chappjc:
Addressing issues revealed in https://github.com/decred/dcrdex/pull/1558#issuecomment-1095680799

This updates the DCR and BTC packages to understand P2PK outputs.

This also fixes the BTC client packages UTXO conversion code so that
unsupported output scripts in the listunspent result are merely skipped
instead of breaking the ability to select UTXOs at all.

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -3414,7 +3414,16 @@ func convertUnspent(confs uint32, unspents []*ListUnspentResult, chainParams *ch
 
 			nfo, err := dexbtc.InputInfo(txout.ScriptPubKey, txout.RedeemScript, chainParams)
 			if err != nil {
+				if errors.Is(err, dex.UnsupportedScriptError) {
+					continue
+				}
 				return nil, nil, 0, fmt.Errorf("error reading asset info: %w", err)
+			}
+			if nfo.ScriptType == dexbtc.ScriptUnsupported || nfo.NonStandardScript {
+				// InputInfo sets NonStandardScript for P2SH with non-standard
+				// redeem scripts. Don't return these since they cannot fund
+				// arbitrary txns.
+				continue
 			}
 
 			utxo := &compositeUTXO{
--- a/dex/networks/dcr/script_test.go
+++ b/dex/networks/dcr/script_test.go
@@ -105,31 +105,43 @@ func TestScriptType(t *testing.T) {
 		}
 	}
 
+	parse(addrs.pk1, 0)
+	check("p2pk-IsP2PK", scriptType.IsP2PK(), true)
+	check("p2pk-IsP2PKH", scriptType.IsP2PKH(), false)
+	check("p2pk-IsP2SH", scriptType.IsP2SH(), false)
+	check("p2pk-IsStake", scriptType.IsStake(), false)
+	check("p2pk-IsMultiSig", scriptType.IsMultiSig(), false)
+
 	parse(addrs.pkh, 0)
+	check("p2pkh-IsP2PK", scriptType.IsP2PK(), false)
 	check("p2pkh-IsP2PKH", scriptType.IsP2PKH(), true)
 	check("p2pkh-IsP2SH", scriptType.IsP2SH(), false)
 	check("p2pkh-IsStake", scriptType.IsStake(), false)
 	check("p2pkh-IsMultiSig", scriptType.IsMultiSig(), false)
 
 	parse(addrs.pkh, txscript.OP_SSGEN)
+	check("stakePKH-IsP2PK", scriptType.IsP2PK(), false)
 	check("stakePKH-IsP2PKH", scriptType.IsP2PKH(), true)
 	check("stakePKH-IsP2SH", scriptType.IsP2SH(), false)
 	check("stakePKH-IsStake", scriptType.IsStake(), true)
 	check("stakePKH-IsMultiSig", scriptType.IsMultiSig(), false)
 
 	parse(addrs.edwards, 0)
+	check("edwards-IsP2PK", scriptType.IsP2PK(), false)
 	check("edwards-IsP2PKH", scriptType.IsP2PKH(), true)
 	check("edwards-IsP2SH", scriptType.IsP2SH(), false)
 	check("edwards-IsStake", scriptType.IsStake(), false)
 	check("edwards-IsMultiSig", scriptType.IsMultiSig(), false)
 
 	parse(addrs.schnorrPK, 0)
+	check("schnorrPK-IsP2PK", scriptType.IsP2PK(), true)
 	check("schnorrPK-IsP2PKH", scriptType.IsP2PKH(), false)
 	check("schnorrPK-IsP2SH", scriptType.IsP2SH(), false)
 	check("schnorrPK-IsStake", scriptType.IsStake(), false)
 	check("schnorrPK-IsMultiSig", scriptType.IsMultiSig(), false)
 
 	pkScript, scriptType := parse(addrs.sh, 0)
+	check("p2sh-IsP2PK", scriptType.IsP2PK(), false)
 	check("p2sh-IsP2PKH", scriptType.IsP2PKH(), false)
 	check("p2sh-IsP2SH", scriptType.IsP2SH(), true)
 	check("p2pkh-IsStake", scriptType.IsStake(), false)
@@ -149,6 +161,7 @@ func TestScriptType(t *testing.T) {
 	check("p2pkh-IsStake", scriptType.IsStake(), false)
 
 	pkScript, scriptType = parse(addrs.sh, txscript.OP_SSGEN)
+	check("stake-p2sh-IsP2PK", scriptType.IsP2PK(), false)
 	check("stake-p2sh-IsP2PKH", scriptType.IsP2PKH(), false)
 	check("stake-p2sh-IsP2SH", scriptType.IsP2SH(), true)
 	check("stake-p2pkh-IsStake", scriptType.IsStake(), true)
--- a/server/asset/btc/btc.go
+++ b/server/asset/btc/btc.go
@@ -130,6 +130,10 @@ type Backend struct {
 	log         dex.Logger
 	decodeAddr  dexbtc.AddressDecoder
 	estimateFee func(*RPCClient) (uint64, error)
+	// booleanGetBlockRPC corresponds to BackendCloneConfig.BooleanGetBlockRPC
+	// field and is used by RPCClient, which is constructed on Connect.
+	booleanGetBlockRPC bool
+	blockDeserializer  func([]byte) (*wire.MsgBlock, error)
 }
 
 // Check that Backend satisfies the Backend interface.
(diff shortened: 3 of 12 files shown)


client/asset/btc/btc.go
```diff
@@ -3349,8 +3349,17 @@ func convertUnspent(confs uint32, unspents []*ListUnspentResult, chainParams *ch
...
+					continue
+				}
 				return nil, nil, 0, fmt.Errorf("error reading asset info: %w", err)
 			}
+			if nfo.ScriptType == dexbtc.ScriptUnsupported || nfo.NonStandardScript {
+				// InputInfo sets NonStandardScript for P2SH with non-standard
+				// redeem scripts. Don't return these since they cannot fund
+				// arbitrary txns.
+				continue
+			}
```
chappjc, line 3362: This now mirror's DCR's client code.

JoeGruffins (approved):

buck54321 (commented):
Maybe add p2pk tracking in `TestLiveUTXO` and `LiveUTXOStats`?

chappjc:
> Maybe add p2pk tracking in `TestLiveUTXO` and `LiveUTXOStats`?

That was a minor rabbit hole.  I ended up fixing some bugs in the btc `LiveUTXOStats` functions (the way it checked segwit was incorrect, maybe once before, but not now).  Also got the serialized block instead of a discrete getrawtransaction for each.  Lastly, added the `booleanVerboseGetBlock` flag for server's btc backend.

server/asset/btc/utxo.go
```diff
@@ -287,7 +287,7 @@ func (utxo *UTXO) Confirmations(context.Context) (int64, error) {
 		// See if we can find the utxo in another block.
 		newUtxo, err := utxo.btc.utxo(&utxo.tx.hash, utxo.vout, utxo.redeemScript)
 		if err != nil {
-			return -1, fmt.Errorf("utxo block is not mainchain")
+			return -1, fmt.Errorf("utxo error: %w", err)
```
chappjc, line 290: `immature output` was one of the ones left after the bug fixes.  Nothing to do with mainchain anyway.  We might rethink the `utxo` method a little because there is a `maturity` field so I'm not sure it should error instead of deferring to the caller to decide if immature is a problem (server is probably checking funding coins or swap txns, so it does make sense, but in terms of general operation `utxo` failing to provide a return for immature outputs even though it has a maturity field is strange)

server/asset/btc/btc.go
```diff
@@ -130,6 +130,9 @@ type Backend struct {
 	log         dex.Logger
 	decodeAddr  dexbtc.AddressDecoder
 	estimateFee func(*RPCClient) (uint64, error)
+	// booleanGetBlockRPC corresponds to BackendCloneConfig.BooleanGetBlockRPC
+	// field and is used by RPCClient, which is constructed on Connect.
+	booleanGetBlockRPC bool
```
buck54321, line 135: I don't see this being set anywhere.

server/asset/btc/btc.go
```diff
@@ -197,6 +200,9 @@ type BackendCloneConfig struct {
 	// FeeEstimator provides a way to get fees given an RawRequest-enabled
 	// client and a confirmation target.
 	FeeEstimator func(*RPCClient) (uint64, error)
+	// BooleanGetBlockRPC causes the RPC client to use a boolean second argument
+	// for the getblock endpoint, instead of Bitcoin's numeric.
+	BooleanGetBlockRPC bool
```
buck54321, line 205: I don't see this being used anywhere.

buck54321 (changes requested):

server/asset/btc/btc.go
chappjc: whoops, half checked in changes.

This seemingly pointless field in `Backend` is because `RPCClient` isn't constructed until connect, and the `dexbtc.RPCConfig` is pretty much intended just to be for ini parsing of the .conf RPC settings.

chappjc: https://github.com/decred/dcrdex/compare/e1902e7bd60af9021a622a2e80575d2d94b4bf2f..e859d9fe6307a035f74ac11be1c88428d7991483

dex/networks/dcr/script_test.go
```diff
@@ -105,31 +105,43 @@ func TestScriptType(t *testing.T) {
...
+
 	parse(addrs.pkh, 0)
+	check("p2pkh-IsP2PK", scriptType.IsP2PK(), false)
 	check("p2pkh-IsP2PKH", scriptType.IsP2PKH(), true)
 	check("p2pkh-IsP2SH", scriptType.IsP2SH(), false)
 	check("p2pkh-IsStake", scriptType.IsStake(), false)
 	check("p2pkh-IsMultiSig", scriptType.IsMultiSig(), false)
 
 	parse(addrs.pkh, txscript.OP_SSGEN)
+	check("stakePKH-IsP2PKH", scriptType.IsP2PK(), false)
```
itswisdomagain, line 123:
```suggestion
	check("stakePKH-IsP2PK", scriptType.IsP2PK(), false)
```

dex/networks/dcr/script_test.go
```diff
@@ -105,31 +105,43 @@ func TestScriptType(t *testing.T) {
...
 	check("stakePKH-IsP2PKH", scriptType.IsP2PKH(), true)
 	check("stakePKH-IsP2SH", scriptType.IsP2SH(), false)
 	check("stakePKH-IsStake", scriptType.IsStake(), true)
 	check("stakePKH-IsMultiSig", scriptType.IsMultiSig(), false)
 
 	parse(addrs.edwards, 0)
+	check("edwards-IsP2PKH", scriptType.IsP2PK(), false)
```
itswisdomagain, line 130:
```suggestion
	check("edwards-IsP2PK", scriptType.IsP2PK(), false)
```

dex/networks/dcr/script_test.go
```diff
@@ -105,31 +105,43 @@ func TestScriptType(t *testing.T) {
...
 	check("edwards-IsP2PKH", scriptType.IsP2PKH(), true)
 	check("edwards-IsP2SH", scriptType.IsP2SH(), false)
 	check("edwards-IsStake", scriptType.IsStake(), false)
 	check("edwards-IsMultiSig", scriptType.IsMultiSig(), false)
 
 	parse(addrs.schnorrPK, 0)
+	check("schnorrPK-IsP2PKH", scriptType.IsP2PK(), true)
```
itswisdomagain, line 137:
```suggestion
	check("schnorrPK-IsP2PK", scriptType.IsP2PK(), true)
```

dex/networks/dcr/script_test.go
```diff
@@ -105,31 +105,43 @@ func TestScriptType(t *testing.T) {
...
 	check("schnorrPK-IsP2PKH", scriptType.IsP2PKH(), false)
 	check("schnorrPK-IsP2SH", scriptType.IsP2SH(), false)
 	check("schnorrPK-IsStake", scriptType.IsStake(), false)
 	check("schnorrPK-IsMultiSig", scriptType.IsMultiSig(), false)
 
 	pkScript, scriptType := parse(addrs.sh, 0)
+	check("p2sh-IsP2PKH", scriptType.IsP2PK(), false)
```
itswisdomagain, line 144:
```suggestion
	check("p2sh-IsP2PK", scriptType.IsP2PK(), false)
```

dex/networks/dcr/script_test.go
```diff
@@ -149,6 +161,7 @@ func TestScriptType(t *testing.T) {
 	check("p2pkh-IsStake", scriptType.IsStake(), false)
 
 	pkScript, scriptType = parse(addrs.sh, txscript.OP_SSGEN)
+	check("stake-p2sh-IsP2PKH", scriptType.IsP2PK(), false)
```
itswisdomagain, line 164:
```suggestion
	check("stake-p2sh-IsP2PK", scriptType.IsP2PK(), false)
```

server/asset/btc/rpcclient.go
```diff
@@ -128,10 +131,43 @@ func (rc *RPCClient) GetRawTransactionVerbose(txHash *chainhash.Hash) (*btcjson.
...
-// GetBlockVerbose fetches verbose block data for the specified hash.
+// GetBlock fetches raw block data, and the "verbose" block header, for the
+// block with the given hash. The verbose block header return is separate
+// because it contains other useful info like the height and median time that
+// the wire type does not contain.
+func (rc *RPCClient) GetBlock(blockHash *chainhash.Hash) (*wire.MsgBlock, *btcjson.GetBlockHeaderVerboseResult, error) {
+	arg := interface{}(0)
+	if rc.booleanVerboseGetBlock {
+		arg = false
+	}
```
itswisdomagain, line 142: Just to ask, I'm sure you've considered this, doesn't the backend parse `0` as `false` or `false` as `0` as needed?

itswisdomagain (changes requested):

server/asset/btc/rpcclient.go
chappjc: bitcoind, yes, dogecoind, no.

```
$  ./dogecoin-cli -testnet getblock 64acaa2347aafe30e8898dbe3c7dd3cef118ba690c79dedc7b463a5c01b0ef8e 0
error code: -1
error message:
JSON value is not a boolean as expected
```

We could assume boolean would always work fine for bitcoind and clones, but now that we know we were glossing over this, we should be explicit.

Plus, there's a "2" if we ever want to do that, so we should have the integer nature of the arg front and center.

chappjc: https://github.com/decred/dcrdex/pull/1558#discussion_r848878728

client/asset/btc/btc.go
```diff
@@ -3349,8 +3349,17 @@ func convertUnspent(confs uint32, unspents []*ListUnspentResult, chainParams *ch
 
 			nfo, err := dexbtc.InputInfo(txout.ScriptPubKey, txout.RedeemScript, chainParams)
 			if err != nil {
+				if errors.Is(err, dex.UnsupportedScriptError) {
+					continue
+				}
 				return nil, nil, 0, fmt.Errorf("error reading asset info: %w", err)
 			}
+			if nfo.ScriptType == dexbtc.ScriptUnsupported || nfo.NonStandardScript {
```
buck54321, line 3357: Can we check these internally to `InputInfo` and return a `dex.UnsupportedScriptError`?

It'll affect `*(*Backend).utxo` server-side, I'm wondering if we're handling unsupported script types properly there.

server/asset/btc/rpcclient.go
```diff
@@ -128,10 +131,43 @@ func (rc *RPCClient) GetRawTransactionVerbose(txHash *chainhash.Hash) (*btcjson.
 		true}, res)
 }
 
-// GetBlockVerbose fetches verbose block data for the specified hash.
+// GetBlock fetches raw block data, and the "verbose" block header, for the
+// block with the given hash. The verbose block header return is separate
+// because it contains other useful info like the height and median time that
+// the wire type does not contain.
+func (rc *RPCClient) GetBlock(blockHash *chainhash.Hash) (*wire.MsgBlock, *btcjson.GetBlockHeaderVerboseResult, error) {
```
buck54321, line 138: Doesn't need to be exported.

client/asset/btc/btc.go
chappjc: > It'll affect `*(*Backend).utxo` server-side, I'm wondering if we're handling unsupported script types properly there.

That's the reason I opted to just mirror dcr.go instead of modifying `InputInfo`.  On server, the `(*Backend).utxo` method is designed to return info about all utxos, at least looking at the return:

```go
out := &Output{
	TXIO: TXIO{
		btc:        btc,
		tx:         tx,
		height:     blockHeight,
		blockHash:  blockHash,
		maturity:   int32(maturity),
		lastLookup: lastLookup,
	},
	vout:              vout,
	scriptType:        scriptType,
	nonStandardScript: inputNfo.NonStandardScript,
	pkScript:          pkScript,
	redeemScript:      redeemScript,
	numSigs:           inputNfo.ScriptAddrs.NRequired,
	spendSize:         inputNfo.VBytes(),
	value:             toSat(txOut.Value),
}
```

Specifically, that `nonStandardScript: inputNfo.NonStandardScript` part.

Maybe it should/could be changed, but I opted for the path of least resistance with the least likelihood of unintended consequences. 

Still worth investigating though.

chappjc: Ah, yeah, it requires non-standard to be not an error because of `(*Backend) Contract(coinID []byte, redeemScript []byte)` -> `(*Backend).output` -> `InputInfo(pkScript, redeemScript)`, where `redeemScript` is the swap contract (non-standard)

buck54321: Oh right.

server/asset/btc/rpcclient.go
chappjc: Unexported.  Will stay wait for DOGE to go in first as per discussion in https://github.com/decred/dcrdex/pull/1581#discussion_r849513634

buck54321 (approved):

PR #1582, opened 2022-04-12
client/dcr: Unlock wallet if acct not encrypted.

### LINKED ISSUE #1576: core/simnet_tests: account is not encrypted with a unique passphrase
As of commit 95c9a38b it seems the client core simnet tests fail with `error starting clients: dcr wallet authentication error: -4: encryption/decryption error: account is not encrypted with a unique passphrase`

JoeGruffins:
closes #1576

### DIFF
--- a/client/asset/dcr/rpcwallet.go
+++ b/client/asset/dcr/rpcwallet.go
@@ -688,10 +688,19 @@ func (w *rpcWallet) LockAccount(ctx context.Context, acctName string) error {
 	return translateRPCCancelErr(err)
 }
 
-// UnlockAccount unlocks the specified account.
-// Part of the Wallet interface.
+// UnlockAccount unlocks the specified account or the wallet if account is not
+// encrypted. Part of the Wallet interface.
 func (w *rpcWallet) UnlockAccount(ctx context.Context, pw []byte, acctName string) error {
-	return translateRPCCancelErr(w.rpcClient.UnlockAccount(ctx, acctName, string(pw)))
+	var res *walletjson.AccountUnlockedResult
+	res, err := w.rpcClient.AccountUnlocked(ctx, acctName)
+	if err != nil {
+		return err
+	}
+	if res.Encrypted {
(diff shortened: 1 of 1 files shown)


client/asset/dcr/rpcwallet.go
```diff
@@ -688,10 +688,19 @@ func (w *rpcWallet) LockAccount(ctx context.Context, acctName string) error {
...
+// encrypted. Part of the Wallet interface.
 func (w *rpcWallet) UnlockAccount(ctx context.Context, pw []byte, acctName string) error {
-	return translateRPCCancelErr(w.rpcClient.UnlockAccount(ctx, acctName, string(pw)))
+	var res *walletjson.AccountUnlockedResult
+	res, err := w.rpcClient.AccountUnlocked(ctx, acctName)
+	if err != nil {
+		return err
+	}
+	if res.Encrypted {
+		return translateRPCCancelErr(w.rpcClient.UnlockAccount(ctx, acctName, string(pw)))
```
chappjc, line 700: Thank you for restoring this logic.  I suppose it makes sense to restore it down here in the `rpcWallet` code instead of up at the `ExchangeWallet` level where it was.

Also, I can confirm that `(*rpcWallet).LockAccount` does have this `accountunlocked` check.

client/core/core.go
```diff
@@ -1700,6 +1700,7 @@ func (c *Core) CreateWallet(appPW, walletPW []byte, form *WalletForm) error {
 	if err != nil {
 		return err
 	}
+	defer crypter.Close()
```
chappjc, line 1703: Hmm, we don't do this *anywhere*, and there seem to be numerous places we should.  Can we split this out and address them all in another PR?

JoeGruffins: ok.

JoeGruffins: diff https://github.com/decred/dcrdex/pull/1582/commits/0fdbd4e6688a9ac47b0812c22ede529f297575bc

chappjc (approved):

PR #1584, opened 2022-04-13
multi: Clear private data asap.

JoeGruffins:
In core methods and consumers, clear passwords, private keys, and seeds
    when possible to protect against some memory scraping attacks.

lightly discussed in #1582

This pr makes it so the crypter cannot be used in goroutines. I think we're not doing so atm, but should be careful going forward.

### DIFF
--- a/client/rpcserver/handlers.go
+++ b/client/rpcserver/handlers.go
@@ -127,9 +127,7 @@ func handleInit(s *RPCServer, params *RawParams) *msgjson.ResponsePayload {
 	}
 	defer func() {
 		appPass.Clear()
-		if len(seed) > 0 {
-			seed.Clear()
-		}
+		seed.Clear()
 	}()
 	if err := s.core.InitializeClient(appPass, seed); err != nil {
 		errMsg := fmt.Sprintf("unable to initialize client: %v", err)
@@ -206,9 +204,9 @@ func handleOpenWallet(s *RPCServer, params *RawParams) *msgjson.ResponsePayload 
 	if err != nil {
 		return usage(openWalletRoute, err)
 	}
+	defer form.appPass.Clear()
 
 	err = s.core.OpenWallet(form.assetID, form.appPass)
-	form.appPass.Clear() // AppPass not needed after this, clear
 	if err != nil {
 		errMsg := fmt.Sprintf("error unlocking %s wallet: %v",
 			dex.BipIDSymbol(form.assetID), err)
@@ -609,6 +607,7 @@ func handleAppSeed(s *RPCServer, params *RawParams) *msgjson.ResponsePayload {
 	}
 	defer appPass.Clear()
 	seed, err := s.core.ExportSeed(appPass)
+	defer encode.ClearBytes(seed)
 	if err != nil {
 		errMsg := fmt.Sprintf("unable to retrieve app seed: %v", err)
 		resErr := msgjson.NewError(msgjson.RPCExportSeedError, errMsg)
@@ -616,12 +615,6 @@ func handleAppSeed(s *RPCServer, params *RawParams) *msgjson.ResponsePayload {
 	}
 	// Zero seed and hex representation after use.
 	seedHex := fmt.Sprintf("%x", seed[:])
-	defer func() {
-		for i := range seed {
-			seed[i] = 0
-		}
-		seedHex = ""
-	}()
 
 	return createResponse(appSeedRoute, seedHex, nil)
 }
--- a/client/webserver/api.go
+++ b/client/webserver/api.go
@@ -17,9 +17,12 @@ import (
 	"decred.org/dcrdex/dex/encode"
 )
 
+var zero = encode.ClearBytes
+
 // apiDiscoverAccount is the handler for the '/discoveracct' API request.
 func (s *WebServer) apiDiscoverAccount(w http.ResponseWriter, r *http.Request) {
 	form := new(registrationForm)
+	defer form.Password.Clear()
 	if !readPost(w, r, form) {
 		return
 	}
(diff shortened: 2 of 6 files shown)


client/webserver/api.go
```diff
@@ -701,7 +721,13 @@ func (s *WebServer) apiWithdraw(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
-	coin, err := s.core.Withdraw(form.Pass, form.AssetID, form.Value, form.Address)
+	pass, err := s.resolvePass(form.Pass, r)
+	if err != nil {
+		s.writeAPIError(w, fmt.Errorf("password error: %w", err))
+		return
+	}
+	defer zero(pass)
+	coin, err := s.core.Withdraw(pass, form.AssetID, form.Value, form.Address)
```
JoeGruffins, line 730: Sort of unrelated, but looked like an oversight. We can used a cached password here same as other places. Could even hide the password box if it is cached. Unless this was intended to always need a password?

client/rpcserver/handlers.go
```diff
@@ -617,9 +616,7 @@ func handleAppSeed(s *RPCServer, params *RawParams) *msgjson.ResponsePayload {
 	// Zero seed and hex representation after use.
 	seedHex := fmt.Sprintf("%x", seed[:])
 	defer func() {
-		for i := range seed {
-			seed[i] = 0
-		}
+		encrypt.Zero(seed)
 		seedHex = ""
```
JoeGruffins, line 620: I'm not sure that clearing a string this way is effectual.

https://stackoverflow.com/questions/39968084/is-it-possible-to-zero-a-golang-strings-memory-safely

The method to clear string memory seems overkill. It's probably fine.

client/rpcserver/handlers.go
```diff
@@ -127,9 +128,7 @@ func handleInit(s *RPCServer, params *RawParams) *msgjson.ResponsePayload {
 	}
 	defer func() {
 		appPass.Clear()
-		if len(seed) > 0 {
-			seed.Clear()
-		}
```
JoeGruffins, line 132: Often confused about slice being nil and not really nil. Confirming this will not panic https://go.dev/play/p/DS8Zp9uQFCy

client/rpcserver/handlers.go
```diff
@@ -13,6 +13,7 @@ import (
 	"decred.org/dcrdex/client/core"
 	"decred.org/dcrdex/dex"
 	"decred.org/dcrdex/dex/encode"
+	"decred.org/dcrdex/dex/encrypt"
```
chappjc, line 16: `encode.ClearBytes` should suffice in this fine too

chappjc (changes requested):
Thanks for this.  I'm pretty sure we don't need the new `encrypt.Zero` given `encode.ClearBytes`.
I would not be opposed to dex/encrypt/encrypt.go defining `type Zero = encode.ClearBytes` to smooth some stuff out, but if a package is already importing `dex/encode`, there's really no need to use encrypt.

client/webserver/api.go
```diff
@@ -15,11 +15,15 @@ import (
 	"decred.org/dcrdex/dex"
 	"decred.org/dcrdex/dex/config"
 	"decred.org/dcrdex/dex/encode"
+	"decred.org/dcrdex/dex/encrypt"
 )
 
+var zero = encrypt.Zero
```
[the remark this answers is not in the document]
JoeGruffins: Oh, I missed that. Will use it instead.

client/rpcserver/handlers.go
JoeGruffins: OH! I've been misunderstanding when things panic for a while then.

JoeGruffins:
Just rebased.

client/rpcserver/handlers.go
JoeGruffins: Just removed because ineffectual.

client/webserver/api.go
```diff
@@ -423,7 +438,10 @@ func (s *WebServer) apiCloseWallet(w http.ResponseWriter, r *http.Request) {
 // apiInit is the handler for the '/init' API request.
 func (s *WebServer) apiInit(w http.ResponseWriter, r *http.Request) {
 	init := new(initForm)
-	defer init.Pass.Clear()
+	defer func() {
+		init.Pass.Clear()
+		zero(init.Seed)
+	}()
```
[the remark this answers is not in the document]
JoeGruffins: Using defer twice now.

client/webserver/api.go
```diff
@@ -648,6 +666,7 @@ func (s *WebServer) apiChangeAppPass(w http.ResponseWriter, r *http.Request) {
 		} else {
 			setCookie(pwKeyCK, hex.EncodeToString(key), w)
 		}
+		zero(key)
```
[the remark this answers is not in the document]
JoeGruffins: Moved to under the else.

client/webserver/api.go
```diff
@@ -701,7 +721,13 @@ func (s *WebServer) apiWithdraw(w http.ResponseWriter, r *http.Request) {
 		s.writeAPIError(w, fmt.Errorf("no wallet found for %s", unbip(form.AssetID)))
 		return
 	}
-	coin, err := s.core.Withdraw(form.Pass, form.AssetID, form.Value, form.Address)
+	pass, err := s.resolvePass(form.Pass, r)
```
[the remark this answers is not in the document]
JoeGruffins: Reverted.

client/webserver/api.go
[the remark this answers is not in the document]
JoeGruffins: Removed the new function and using `encode.ClearBytes`

client/rpcserver/handlers.go
JoeGruffins: Ok.

JoeGruffins:
https://github.com/decred/dcrdex/compare/85b65b73f865060c7299ce663055b8ef50d3f48f..7ab5425390ffe0ed8fbc450cbf6787efeca844e9

chappjc (approved):

Issue #1587, opened 2022-04-15
dcr simnet harness tests fails: mainchain / validation issues


chappjc:
> 2022-04-15 12:15:52.677 [WRN] CORE:2[dcr]: Block 657daa7f6df971f6e3428566760c7b700f44ada4209000837e7ab7ae74d918e8 found to contain tx c85b06032c8e641d7a7a63fdc74ee9b976ce8dc811696120af3ba69f45e6aea2 **has been invalidated**.

> 2022-04-15 12:16:06.182 [ERR] CORE:2: Failed to get confirmations of the counter-party's swap c85b06032c8e641d7a7a63fdc74ee9b976ce8dc811696120af3ba69f45e6aea2:0 (dcr) for match 67f39ef033b6ad713d29ed088e7e0c951b96e03878ffa24f6fd45d3db23936df, order 9864e737855f2d2ea1cef0daa4bfdef6ca30f5c84c523ead52584dc7a2883e96: error checking if output c85b06032c8e641d7a7a63fdc74ee9b976ce8dc811696120af3ba69f45e6aea2:0 is spent: **no mainchain ancestor** for block 657daa7f6df971f6e3428566760c7b700f44ada4209000837e7ab7ae74d918e8

Probably related to mainchain/validation changes in https://github.com/decred/dcrdex/pull/1510

chappjc:
Seems to be fine on `release-v0.4` and `staging-v0.4.3`

PR #1588, opened 2022-04-15
client/asset/dcr: fix side chain and approval checks

chappjc:
Resolves https://github.com/decred/dcrdex/issues/1587

```
The mainchain and valid returns from (*ExchangeWallet).blockHeader were
switched.  This flips them correctly.

This also changes the way a block is detected as side chain by checking
the Confirmations field of the getblockverbose result instead of
requesting the hash of the block at the same height and comparing
the hashes.  A side chain block will have -1 for Confirmations.

This also changes the way the next block is retrieved for stakeholder
approval checking.  Instead of requesting the hash of the block at the
next height, use the NextHash field returned by getblockverbose.
This saves an RPC, but it is also more robust since there is no need
to use strings.Contains on the error returned from getblockhash
to detect if there is no next block (we are at tip).  If we are at tip,
the NextHash field is empty.  With getblockhash, the wallet says
"-1: Block number out of range: N" while the node says
"-4: wallet.BlockInfo: item does not exist: no block at height N in main
chain".  Although client only ever checks with wallet, these strings are
fairly arbitrary  human readable messages that are likely to change
without warning.
```

### DIFF
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -3193,40 +3193,35 @@ func (dcr *ExchangeWallet) mainchainAncestor(ctx context.Context, blockHash *cha
 	}
 }
 
-// blockHeader returns the *BlockHeader for the specified block hash, and a
-// validity bool indicating true when the block has not been stake-invalidated.
-// validity will always be false if mainchain is false.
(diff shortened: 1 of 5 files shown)


JoeGruffins (approved):

itswisdomagain (approved):

chappjc:
Merging so we can test other stuff properly, e.g. https://github.com/decred/dcrdex/pull/1578

PR #1589, opened 2022-04-17
ui: more long host line breaks

chappjc:
Just some more ui tweaks for long onion host names.

Just enough to for a v3 .onion address to go on two lines on the reg fee asset selection step:

image

This also updates the register wallet wait form to clarify the following points:

1. says the precise amount, not just a vague mention of network fees (this was actually done previously by @martonp)
2. refers to the address as _your_ wallet's address
3. makes it clear you can deposit as much as you want because it's your wallet and only the required amount will be used to register
4. now implies that after the deposit there is a "next step" where the fee is actually paid and you become registered. Before I think this form may have suggested that by sending to the show address that you were making the fee payment then
5. there's a little "i" circle defining the purpose of the unfortunately named "registration fee" as really a small one-time anti-spam fee to discourages people from fucking up trades
6. After you send funds to the address, nothing happens until the tx is confirmed. That was a repeated point of confusion before

image


chappjc:
Pushed a commit to clarify some of the confusing points around funding the registration fee, based on recent conversations in matrix with community members.

image

JoeGruffins:
My "i" message looks different. Any idea why that is?

image

$ node --version
v17.9.0
image

JoeGruffins (approved):

chappjc:
> My "i" message looks different. Any idea why that is?

Must be that `flex-center` on the div around the two spans.  Will fix.

JoeGruffins:
Looks ok now.

image

JoeGruffins (approved):

buck54321 (approved):
That form is getting out of hand. Too many words. We need to either say it more concisely or use layout to separate the info.

This tests fine, and we need to get the info on there, so this is a start.

chappjc:
I'm open to changes.  Obviously I think it needs to be this verbose to be sufficiently clear, but I'm open to suggestions as long as nothing is left to the imagination.

chappjc:
I say that because I have been through numerous conversations with relatively informed community members, and this form in general is extremely confusing and the user almost always does it wrong.

PR #1590, opened 2022-04-19
multi: update version string system and build metadata

### LINKED ISSUE #1518: update version string system and build metadata
Latest version string handling: https://github.com/decred/dcrd/pull/2651
Build metadata including vcs commit on go 1.18: https://github.com/decred/dcrd/pull/2841

ukane-philemon:
This diff updates the version string system and build metadata using the following as a reference.
1. Latest version string handling: [decred/dcrd#2651](https://github.com/decred/dcrd/pull/2651)
2. Build metadata including vcs commit on go 1.18: [decred/dcrd#2841](https://github.com/decred/dcrd/pull/2841)
Closes #1518


dex/version/version_buildinfo.go
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2013-2014 The btcsuite developers
+// Copyright (c) 2015-2022 The Decred developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+//go:build go1.18
+// +build go1.18
```
chappjc, line 7: We don't want the "+build" lines any more now that Go 1.17 is the minimum.

chappjc (changes requested):
Thanks for porting this over from dcrd/dcrwallet!  A couple minor tweaks and I think it's good.

dex/version/version.go
```diff
@@ -0,0 +1,109 @@
+// Copyright (c) 2013-2014 The btcsuite developers
+// Copyright (c) 2015-2022 The Decred developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
```
JoeGruffins, line 4: Not sure what is best here. The LICENSE file in our root directory is a different license. Should either include the LICENSE file this is referencing or include a link or change this to ours. Not sure at all... And what if we change this file will we just up the date?

JoeGruffins (approved):
```
$ ./dexc --version
dexc version 0.5.0-pre+ee33a8f11 (Go version go1.18 linux/amd64)
```

Looks good.

client/cmd/dexc/version.go
```diff
@@ -0,0 +1,49 @@
...
+	// It is defined as a variable so it can be overridden during the build
+	// process with:
+	// '-ldflags "-X main.Version=fullsemver"'
+	// if needed.
+	//
+	// It MUST be a full semantic version per the semantic versioning spec or
+	// the package will panic at runtime.  Of particular note is the pre-release
+	// and build metadata portions MUST only contain characters from
+	// semanticAlphabet.
+	// NOTE: The Version string is overidden on init.
```
chappjc, line 43: spelling: overridden, here and the other two files

dex/version/version.go
chappjc: Part of me wants to just change this to the blueoak license but I suspect it's more complicated than that even though we are the "Decred developers".  So we could do one of the following:
- a link to the license in the dcrd repo (https://github.com/decred/dcrd/blob/master/LICENSE)
- the actual 15 or so lines of the license from that file in this comment, just without the Lightning Network line since that does not pertain to this file
- put a LICENSE file in this dex/version folder

I think a URL would be just fine personally.

dex/version/version_buildinfo.go
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2013-2014 The btcsuite developers
+// Copyright (c) 2015-2022 The Decred developers
```
chappjc, line 2: I don't get these lines.  Surely this `vcsCommitID` didn't even exist until recently let alone back when it was btcsuite work.

Just 2021-2022 and the one author:
https://github.com/decred/dcrd/blob/master/internal/version/version_buildinfo.go

dex/version/version_test.go
```diff
@@ -0,0 +1,370 @@
+// Copyright (c) 2013-2014 The btcsuite developers
+// Copyright (c) 2015-2022 The Decred developers
```
chappjc, line 2: Also has wrong  dates and authors for this file.

chappjc (changes requested):
Works, just a few nits.

```sh
$  ./dexc -V  # from pkg.sh
dexc version 0.5.0+release (Go version go1.18.1 linux/amd64)
$  ./dexc -V  # from go build
dexc version 0.5.0-pre+48d3a3e4d (Go version go1.18.1 linux/amd64)
```

client/cmd/dexc/version.go
ukane-philemon: @chappjc thanks for looking into this.

README.md
```diff
@@ -1,8 +1,9 @@
 # <img src="docs/images/logo_wide_v1.svg" alt="DCRDEX" width="250">
 
 [![Build Status](https://github.com/decred/dcrdex/workflows/Build%20and%20Test/badge.svg)](https://github.com/decred/dcrdex/actions)
-[![ISC License](https://img.shields.io/badge/license-Blue_Oak-007788.svg)](http://copyfree.org)
+[![ISC License](https://img.shields.io/badge/license-Blue_Oak-007788.svg)](https://blueoakcouncil.org/license/1.0.0)
 [![GoDoc](https://img.shields.io/badge/go.dev-reference-blue.svg?logo=go&logoColor=lightblue)](https://pkg.go.dev/decred.org/dcrdex)
+[![](https://img.shields.io/badge/license-ISC%20License-informational.svg)](https://github.com/decred/dcrd/blob/master/LICENSE)
```
chappjc, line 6: We don't want two license badges.  We can have a link in the relevant source code files with a different license, but this one should be representative of the the product as a whole baring a few exceptions like the version info files.

dex/version/version.go
```diff
@@ -1,7 +1,5 @@
-// Copyright (c) 2013-2014 The btcsuite developers
 // Copyright (c) 2015-2022 The Decred developers
-// Use of this source code is governed by an ISC
-// license that can be found in the LICENSE file.
+// Use of this source code is governed by an ISC license.
```
chappjc, line 2: I meant put a link here in these files, not the README

chappjc (changes requested):

chappjc (approved):

PR #1591, opened 2022-04-20
multi: show fee rate on withdrawal form

### LINKED ISSUE #970: UI: on Withdraw page, show current fee rate estimate
Since bitcoind's fee rate estimation can be wildly inaccurate if bitcoind has just started and hasn't had a chance to really warm up (the gray area where estimatesmartfee does not throw an error but instead returns poor estimates), the withdraw page could show the backend's current fee rate estimate to help the user decide if it's going to be be appropriate given external sources such as mempool.space.

ukane-philemon:
This diff displays the estimated fee rate on the withdrawal form and adds a `FallbackFeeRate() uint64` method to default if the fee rate is unavailable.
feeratedcr
btcfeerate

Closes #970

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -520,6 +520,11 @@ func (eth *ExchangeWallet) MaxOrder(lotSize uint64, feeSuggestion uint64, nfo *d
 	return est, nil
 }
 
+// FallbackFeeRate returns the fallback fee rate.
+func (eth *ExchangeWallet) FallbackFeeRate() uint64 {
+	return defaultGasFee
+}
+
 // PreSwap gets order estimates based on the available funds and the wallet
 // configuration.
 func (eth *ExchangeWallet) PreSwap(req *asset.PreSwapForm) (*asset.PreSwap, error) {
@@ -1357,11 +1362,12 @@ func (eth *ExchangeWallet) RegFeeConfirmations(ctx context.Context, coinID dex.B
 }
 
 // FeeRate satisfies asset.FeeRater.
-func (eth *ExchangeWallet) FeeRate() (uint64, string) {
+// FeeRate returns the current optimal fee rate in gwei / gas.
+func (eth *ExchangeWallet) FeeRate() uint64 {
 	base, tip, err := eth.node.currentFees(eth.ctx)
 	if err != nil {
 		eth.log.Errorf("Error getting net fee state: %v", err)
-		return 0, ""
+		return 0
 	}
 
 	feeRate := new(big.Int).Add(
@@ -1371,10 +1377,10 @@ func (eth *ExchangeWallet) FeeRate() (uint64, string) {
 	feeRateGwei, err := dexeth.WeiToGweiUint64(feeRate)
 	if err != nil {
 		eth.log.Errorf("Failed to convert wei to gwei: %v", err)
-		return 0, ""
+		return 0
 	}
 
-	return feeRateGwei, "gwei"
+	return feeRateGwei
 }
 
 func (eth *ExchangeWallet) checkPeers() {
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -1357,11 +1357,11 @@ func (eth *ExchangeWallet) RegFeeConfirmations(ctx context.Context, coinID dex.B
 }
 
 // FeeRate satisfies asset.FeeRater.
-func (eth *ExchangeWallet) FeeRate() uint64 {
+func (eth *ExchangeWallet) FeeRate() (uint64, string) {
 	base, tip, err := eth.node.currentFees(eth.ctx)
 	if err != nil {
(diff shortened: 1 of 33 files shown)


client/webserver/site/src/js/wallets.ts
```diff
@@ -454,10 +454,20 @@ export default class WalletsPage extends BasePage {
...
+
+    Doc.hide(page.FeeRate)
+    const res = await postJSON('/api/getfeerate', {
+      assetID: assetID
+    })
+    if (!app().checkResponse(res) || res.feeRate === 0) {
+      page.withdrawFeeRate.textContent = 'unavailable'
+    } else if (res.feeRate > 0) {
+      page.withdrawFeeRate.textContent = res.feeRate
+      page.withdrawUnit.textContent = wallet.units + '/' + res.unit
```
JoeGruffins, line 468: Would it be better if the fee rate was also just a constant connected to the wallet? It would not ever change, just like the units. I see dcr does right now, but can just divide by 1000 and use B.

JoeGruffins (commented):
Looks ok except for eth should be gwei/gas I think.

image

client/asset/eth/eth.go
```diff
@@ -1371,10 +1377,10 @@ func (eth *ExchangeWallet) FeeRate() uint64 {
 	feeRateGwei, err := dexeth.WeiToGweiUint64(feeRate)
 	if err != nil {
 		eth.log.Errorf("Failed to convert wei to gwei: %v", err)
-		return 0
+		return 0, ""
 	}
 
-	return feeRateGwei
+	return feeRateGwei, "gwei"
```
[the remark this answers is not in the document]
ukane-philemon: Thanks for looking into this, I was not able to check with eth at the time.

JoeGruffins:
Actually I guess that's also Sats/B for bitcoin correct?

client/webserver/site/src/js/wallets.ts
ukane-philemon: AFAICT, the fee rate changes and we would want to show the user a current fee rate?

JoeGruffins: The rate changes but not the units. I mean the units are passed as a string, and don't need to be because we already know what they are per asset.

JoeGruffins: > Would it be better if the fee rate

Oh, sorry. I just meant the unit. Not the rate...

JoeGruffins: eth is gwei/gas dcr is atom/byte and btc is sat/byte and doesn't change.

client/asset/interface.go
```diff
@@ -319,6 +319,8 @@ type Wallet interface {
 	// EstimateRegistrationTxFee returns an estimate for the tx fee needed to
 	// pay the registration fee using the provided feeRate.
 	EstimateRegistrationTxFee(feeRate uint64) uint64
+	// FallbackFeeRate returns the fallback fee rate.
+	FallbackFeeRate() uint64
```
buck54321, line 323: I think we should flip the script on this, and instead add a method like 
```Go
EstimateWithdrawalFees(amt, feeSuggestion uint64) (fees, feeRate uint64, err error)
```
This would allow the wallet to use whatever internal methods it already uses to find and apply a fee rate, and you'd get an accurate value for total fees (that considers available inputs for utxo-based). It's also a similar pattern to that already used for estimating swaps, redeems, and the registration fee, so fits the interface nicely.

buck54321 (commented):
I question the usefulness of displaying a fee rate to the user. Virtually nobody will know how to use it. Even if you showed them the expected tx size, they wouldn't know what it means.

What I'd want to see during a withdraw is the total fees paid and the fees as a percentage of withdrawn amount.




If you search around for the withdraw fees listed for various exchanges, they are all listed as absolute values.

chappjc:
Oh dear, I recalled that getting such a table with amounts and fees right depends on the asset unfortunately because ETH doesn't do withdraws the way the others do.  ETH's wallet instead sends that amount to the address rather than subtracting fees from it (except if it looks like a sweep!).  This goes back to the `WalletTraitSender` idea that would correspond to if the wallet offered a `Send` method with the "no subtract" semantics.

I wonder if we need to get ETH's implementations of Withdraw and Send right before we can do this properly. (properly meaning show the three amounts in the little table in my comment)

chappjc:
Refs for the Send vs. Withdraw conversations: https://github.com/decred/dcrdex/pull/1355#discussion_r775917934 and https://github.com/decred/dcrdex/pull/1375/files#r778363477

ukane-philemon:
@chappjc, I just marked #1688 ready for review. So I was wondering if I can/should close one.

chappjc:
Superceded by https://github.com/decred/dcrdex/pull/1688

Issue #1561, opened 2022-04-05
trading: timed out waiting for "init" response


JoeGruffins:
Have been seeing this a lot while testing. Not sure if eth related. Today I hit for the first time on testnet so decided to just go ahead and make an issue.

It ultimately leads to a confusing state for the client. For example:

![image](https://user-images.githubusercontent.com/34998433/161673032-11b53aaf-1568-4d7e-82ac-fc8610a4c61c.png)

If you refresh, it does show refunding for a few seconds... I guess until something happens, a tick?
![image](https://user-images.githubusercontent.com/34998433/161676647-62eb6354-5e11-4427-a46d-63469551d984.png)

The client did get an error from the server that matches were revoked. But what are we doing now? Waiting for refund? The other client seems sure that they are refunding:

![image](https://user-images.githubusercontent.com/34998433/161675501-c5aa7fec-0c2a-438e-8617-7921e49fa2cc.png)

But besides the UI uncertainty here, isn't it odd the init failed? Just one of four of the same trade was ok? This is on the same machine too, so probably not a communication issue.

This is on #1552 with the `confCheckTimeout` set to 10 seconds but I've see the same init response timeout error on master as well on simnet. Can also be seen in #1538 so maybe solved by #1541 ? If so there is still the wierd UI issue for the errored client.

From logs it looks like the three failed matches were Taker, and the success was Maker. It looks like all three of the failed matches sent their inits at `02:28:26`  and server got them all. Then server searched for a minute and failed them for timeout. Then it found them all the next second!? Seems suspicious. Is it possible `s.matchMtx.Lock()` is being held somewhere preventing `processInit` from continuing?

Then the client doesn't know how to interpret the message: `[ERR] CORE[wss://127.0.0.1:17273/ws]: No handler found for response: {"type":2,"id":18,"payload":{"result":null,"error":{"code":25,"message":"match already revoked due to inaction"}}}` which I think is an old issue.

Errored client log:
[dexc.log](https://github.com/decred/dcrdex/files/8414668/dexc.log)

Other client log:
[dexc.log](https://github.com/decred/dcrdex/files/8414676/dexc.log)

Server log:
[dcrdex.log](https://github.com/decred/dcrdex/files/8414679/dcrdex.log)

chappjc:
Thank you for dog-fooding so hard.  It is absolutely critical and you should get a medal when this is done.

[...]

PR #1593, opened 2022-04-21
client/asset/eth: Fix eth receipt locktime.

### LINKED ISSUE #1592: eth: Bad locktime.
Have seen this lately. Not sure why, but a recent change seems to have caused locktime to not be set sometimes on simnet. 

```
2022-04-21 10:38:15.015 [INF] CORE: Broadcasted transaction with 1 swap contracts for order 7ea3b9ba72cb6ba2e736dee946ac048a2e6a8d21b855f75fd7c97f360277de15. Assigned fee rate = 200. Receipts (eth): [{ tx hash: 9e2c1a54758baff2cdd332c3f33da13758fd4647788d756f1cd14af70f7f7a72, contract address: 0x2f68E723b8989BA1C6a9F03E42F33cb7Dc9D606f, secret hash: 24b95c6d09bfa306db01ace478b8bc21c955f83e8f5354927bfd8c95f81020bc }].
2022-04-21 10:38:15.016 [INF] CORE: Contract coin 9e2c1a54758baff2cdd332c3f33da13758fd4647788d756f1cd14af70f7f7a72 (eth), value = 10000000000, refundable at 1970-01-20 02:29:37.095 +0000 UTC (script = 0000000024b95c6d09bfa306db01ace478b8bc21c955f83e8f5354927bfd8c95f81020bc), match = 287e1435c2465670e4a2911994a39683f53c922f32483e9a1b6213e762dee3ed
```

refundable at 1970-01-20 02:29:37.095 +0000 UTC

So, not completely nil? I guess locktime taker added to zero time rather than now somewhere.

JoeGruffins:
Parse locktime as seconds because it is seconds and not milliseconds.

closes #1592

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -780,7 +780,7 @@ func (eth *ExchangeWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin
 		copy(secretHash[:], swap.SecretHash)
 		receipts = append(receipts,
 			&swapReceipt{
-				expiration:   encode.UnixTimeMilli(int64(swap.LockTime)),
+				expiration:   time.Unix(int64(swap.LockTime), 0),
 				value:        swap.Value,
 				txHash:       txHash,
 				secretHash:   secretHash,
--- a/client/cmd/dexcctl/simnet-setup.sh
+++ b/client/cmd/dexcctl/simnet-setup.sh
@@ -2,6 +2,9 @@
 # Set up DCR and BTC wallets and register with the DEX.
 # dcrdex, dexc, and the wallet simnet harnesses should all be running before
 # calling this script.
+#
+# dexc can be built with -ldflags "-X 'decred.org/dcrdex/dex.testLockTimeTaker=30s' -X 'decred.org/dcrdex/dex.testLockTimeMaker=1m'"
+# in order to set simnet locktimes.
 
 set +e
 
--- a/client/asset/eth/eth_test.go
+++ b/client/asset/eth/eth_test.go
@@ -1152,9 +1152,9 @@ func TestSwap(t *testing.T) {
 		for i, contract := range swaps.Contracts {
 			// Check that receipts match the contract inputs
 			receipt := receipts[i]
-			if encode.UnixMilliU(receipt.Expiration()) != contract.LockTime {
+			if uint64(receipt.Expiration().Unix()) != contract.LockTime {
 				t.Fatalf("%v: expected expiration %v != expiration %v",
-					testName, encode.UnixTimeMilli(int64(contract.LockTime)), receipts[0].Expiration())
+					testName, time.Unix(int64(contract.LockTime), 0), receipts[0].Expiration())
 			}
 			if receipt.Coin().Value() != contract.Value {
 				t.Fatalf("%v: receipt coin value: %v != expected: %v",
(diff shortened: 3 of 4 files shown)


JoeGruffins:
With shorter locktimes, eth swaps never complete. Even if the tx has been mined 20 blocks ago and the tx can for sure be fetched by the console. Related to  #1561

client/asset/eth/eth.go
```diff
@@ -780,7 +780,7 @@ func (eth *ExchangeWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin
 		copy(secretHash[:], swap.SecretHash)
 		receipts = append(receipts,
 			&swapReceipt{
-				expiration:   encode.UnixTimeMilli(int64(swap.LockTime)),
+				expiration:   time.Unix(int64(swap.LockTime), 0),
```
chappjc, line 783: Wow, good catch @JoeGruffins. Yikes.  I can confirm the other assets like DCR are indeed treating this `Contract.LockTime` field as seconds.   Also that `(*Core).swapMatchGroup` is preparing it in **seconds** like `matchTime.Add(t.lockTimeTaker).UTC().Unix()`.

So, righteous fix!  Only thing I think we should change is the docs on the `client/asset.Contract.LockTime` field so that it states the units.

client/cmd/dexcctl/simnet-setup.sh
```diff
@@ -2,6 +2,9 @@
 # Set up DCR and BTC wallets and register with the DEX.
 # dcrdex, dexc, and the wallet simnet harnesses should all be running before
 # calling this script.
+#
+# dexc should be built with -ldflags "-X 'decred.org/dcrdex/dex.testLockTimeTaker=30s' -X 'decred.org/dcrdex/dex.testLockTimeMaker=1m'"
+# in order to mirror the default server simnet locktimes.
```
chappjc, line 7: BTW, when doing interactive testing, I usually rebuild server with longer lock times.
With eth and propagation often taking a minute on testnet, I suspect we'll want to change increase this.

client/asset/eth/eth.go
buck54321: > Also, looks to have been broken in November, so it was once correct anyway.

Damn. That was me.

client/cmd/dexcctl/simnet-setup.sh
buck54321: Any thoughts on using a Makefile for this stuff?

chappjc: Could.  I like your `build` and `run` scripts you added for server though.  But to start up the harness without the short lock times, I can see how a make file would be nice, like `make interactive` for longer lock times suitable for interactive testing vs `make scripted` suitable for the scripted harness trade tests.

@JoeGruffins I'm not sure what everyone's workflow looks like, but I should clarify that in general, I advise against any kind of interactive tests with these 30sec/1min lock times.  They are only good for for the *scripted* simnet tests.

JoeGruffins: Changed the wording to only inform it can be changed in that manner.

chappjc (approved):

client/asset/eth/eth.go
JoeGruffins: is "UNIX seconds" correct English

PR #1597, opened 2022-04-22
server/asset/{btc,doge,ltc,bch}: median-fee based fee rate estimation backup

buck54321:
```
Add methods for calculating the median fee of the most recent block(s).
Ensure that there is sufficient data by requiring M transactions in the
last N blocks. If not enough data, fallback to a "no-competition" fee
rate. Implement a cache to prevent repeated scans between blocks.
Some chains provide the getblockstats RPC. Others require manual
scanning.
```

### DIFF
--- a/server/asset/btc/btc.go
+++ b/server/asset/btc/btc.go
@@ -29,6 +29,8 @@ import (
 	"github.com/decred/dcrd/rpcclient/v7"
 )
 
+const defaultNoCompetitionRate = 10
+
 // Driver implements asset.Driver.
 type Driver struct{}
 
@@ -109,7 +111,8 @@ func netParams(network dex.Network) (*chaincfg.Params, error) {
 // cache of block data for quick lookups. Backend implements asset.Backend, so
 // provides exported methods for DEX-related blockchain info.
 type Backend struct {
-	cfg *dexbtc.RPCConfig
+	rpcCfg *dexbtc.RPCConfig
+	cfg    *BackendCloneConfig
 	// The asset name (e.g. btc), primarily for logging purposes.
 	name string
 	// segwit should be set to true for blockchains that support segregated
@@ -127,13 +130,20 @@ type Backend struct {
 	chainParams *chaincfg.Params
 	// A logger will be provided by the dex for this backend. All logging should
 	// use the provided logger.
-	log         dex.Logger
-	decodeAddr  dexbtc.AddressDecoder
-	estimateFee func(*RPCClient) (uint64, error)
-	// booleanGetBlockRPC corresponds to BackendCloneConfig.BooleanGetBlockRPC
-	// field and is used by RPCClient, which is constructed on Connect.
-	booleanGetBlockRPC bool
-	blockDeserializer  func([]byte) (*wire.MsgBlock, error)
+	log               dex.Logger
+	decodeAddr        dexbtc.AddressDecoder
+	blockDeserializer func([]byte) (*wire.MsgBlock, error)
+	// fee estimation configuration
+	feeConfs          int64
+	noCompetitionRate uint64
+
+	// The feeCache prevents repeated calculations of the median fee rate
+	// between block changes when estimate(smart)fee is unprimed.
+	feeCache struct {
+		sync.Mutex
+		fee  uint64
+		hash chainhash.Hash
+	}
 }
 
 // Check that Backend satisfies the Backend interface.
@@ -163,29 +173,35 @@ func NewBackend(configPath string, logger dex.Logger, network dex.Network) (asse
 	})
 }
 
-func newBTC(cloneCfg *BackendCloneConfig, cfg *dexbtc.RPCConfig) *Backend {
-	feeEstimator := feeRate
-	if cloneCfg.FeeEstimator != nil {
-		feeEstimator = cloneCfg.FeeEstimator
-	}
-
+func newBTC(cloneCfg *BackendCloneConfig, rpcCfg *dexbtc.RPCConfig) *Backend {
 	addrDecoder := btcutil.DecodeAddress
 	if cloneCfg.AddressDecoder != nil {
 		addrDecoder = cloneCfg.AddressDecoder
 	}
 
+	noCompetitionRate := cloneCfg.NoCompetitionFeeRate
+	if noCompetitionRate == 0 {
+		noCompetitionRate = defaultNoCompetitionRate
+	}
+
+	feeConfs := cloneCfg.FeeConfs
+	if feeConfs == 0 {
+		feeConfs = 1
+	}
+
 	return &Backend{
-		cfg:                cfg,
-		name:               cloneCfg.Name,
-		blockCache:         newBlockCache(),
-		blockChans:         make(map[chan *asset.BlockUpdate]struct{}),
-		chainParams:        cloneCfg.ChainParams,
-		log:                cloneCfg.Logger,
-		segwit:             cloneCfg.Segwit,
-		decodeAddr:         addrDecoder,
-		estimateFee:        feeEstimator,
-		booleanGetBlockRPC: cloneCfg.BooleanGetBlockRPC,
-		blockDeserializer:  cloneCfg.BlockDeserializer,
+		rpcCfg:            rpcCfg,
+		cfg:               cloneCfg,
+		name:              cloneCfg.Name,
+		blockCache:        newBlockCache(),
+		blockChans:        make(map[chan *asset.BlockUpdate]struct{}),
+		chainParams:       cloneCfg.ChainParams,
+		log:               cloneCfg.Logger,
+		segwit:            cloneCfg.Segwit,
+		decodeAddr:        addrDecoder,
+		noCompetitionRate: noCompetitionRate,
+		feeConfs:          feeConfs,
+		blockDeserializer: cloneCfg.BlockDeserializer,
 	}
 }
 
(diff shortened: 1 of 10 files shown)


chappjc:
Very cool for a fallback if `estimate(smart)fee` is bombing.

Although honestly if I'm picking a mainnet BTC fee rate, I barely even look at the recent past blocks, instead looking at the state of *mempool*.  Say the next block is shaping up like the following:

image

Even if that mined block 733097 was full or nearly full, I'd make my decision based primarily on the next block's worth of mempool on the left:
- If mempool is barely used (like above), I'd pick a value that was similar to the previous block's median fee rate or lower if the median was inexplicably high. Could just be 1 (the min rate) but that is risky because the next block might takes a while.  So here's where I think the median of past blocks is helpful, although subject to manipulation if there aren't many txns.
- If that chunk of mempool is full (more than one "next blocks" on the left) or almost full, then I'd look closer at the histogram of the fee rates in that section of mempool:

image

I'd probably pick something around 7 or 8 in the above case case.  Regardless, in this second case, I would barely consider the last mined block because the fees in mempool could be going up rapidly, and if I picked based on the last block, I could be immediately several blocks deep in mempool.

Do you think there's a way to bring mempool into the calculus you've got so far?

buck54321:
I'm happy to work some mempool in, but the whole point here is that mempool is not developed enough for `estimatesmartfee`. I'll take a look.

chappjc:
> I'm happy to work some mempool in, but the whole point here is that mempool is not developed enough for `estimatesmartfee`. I'll take a look.

Ah, for e.g. with doge mempool is probably fairly empty 
Is the assumption that BTC isn't gonna hit this code?  It only would if the node is new and hasn't observed several blocks being mined.

chappjc:
Another example from electrum x server, which can target a depth in the memory pool, which is conceptually what I do mentally
https://github.com/spesmilo/electrum/blob/59c1d03f018026ac301c4e74facfc64da8ae4708/RELEASE-NOTES#L34-L46

At the end of the day, if mempool is fairly empty, one could even go with the min relay fee, possibly plus some extra if blocks have historically been fairly full.  If mempool is of significant size, I'd _only_ look at mempool, almost completely disregarding past blocks.

chappjc:
Sounds reasonable.  It's a sound backup.

For ref, here's a link to me early experimentation with `estimatesmartfee` in its warm-up stages: https://github.com/decred/dcrdex/issues/501 In short, it look bitcoind observing 5 blocks getting mined before it declared estimatesmartfee to have "sufficient" data.  I think it was because it has to observer (a) txns entering its mempool and (b) those same txns getting mined, and observing those times and rates.

server/asset/btc/rpcclient.go
```diff
@@ -130,8 +164,195 @@ func (rc *RPCClient) GetRawTransactionVerbose(txHash *chainhash.Hash) (*btcjson.
...
+func (rc *RPCClient) medianFeesTheHardWay() (uint64, error) {
+	const numTxs = 101
+
+	iHash, err := rc.GetBestBlockHash()
+	if err != nil {
+		return 0, err
+	}
+
+	txs := make([]*wire.MsgTx, 0, numTxs)
+	prevOuts := make(map[chainhash.Hash]map[int]int64, numTxs)
```
chappjc, line 260: pls doc the keys and values: first map key is prevout hash (funding tx), the second map is prev out idx => value

server/asset/btc/rpcclient.go
```diff
@@ -130,8 +164,195 @@ func (rc *RPCClient) GetRawTransactionVerbose(txHash *chainhash.Hash) (*btcjson.
...
+			tx := utilTx.MsgTx()
+
+			for _, vin := range tx.TxIn {
+				prevOut := vin.PreviousOutPoint
+				prevs := prevOuts[prevOut.Hash]
+				if len(prevs) == 0 {
+					prevs = make(map[int]int64, 1)
+					prevOuts[prevOut.Hash] = prevs
+				}
+				prevs[int(prevOut.Index)] = 0 // placeholder
```
chappjc, line 303: `// value will be retrieved subsequently, after all prevouts for this funding txhash are recorded`

server/asset/btc/rpcclient.go
```diff
@@ -130,8 +164,195 @@ func (rc *RPCClient) GetRawTransactionVerbose(txHash *chainhash.Hash) (*btcjson.
...
+		}
+		fees := in - out
+		if fees < 0 {
+			return 0, fmt.Errorf("fees < 0 for tx %s", tx.TxHash())
+		}
+		sz := tx.SerializeSize()
+		if sz == 0 {
+			return 0, fmt.Errorf("size 0 tx %s", tx.TxHash())
+		}
+		rates = append(rates, uint64(fees)/uint64(sz))
```
chappjc, line 352: Under non-error conditions you will always append to length `numTxs`, so I'd be tempted to do `for i, tx := range txs` and `rates[i] = ...`  to spare modifying len in the slice header on each iteration, but nbd.

server/asset/btc/btc.go
```diff
@@ -127,9 +130,19 @@ type Backend struct {
 	chainParams *chaincfg.Params
 	// A logger will be provided by the dex for this backend. All logging should
 	// use the provided logger.
-	log         dex.Logger
-	decodeAddr  dexbtc.AddressDecoder
-	estimateFee func(*RPCClient) (uint64, error)
```
chappjc, line 132: :heart: 
I'm looking to make a similar change in the future to client where we have the awkward issue of a `client/asset.Wallet` having to be a `RawRequester`, something that stands out given it's RPC-oriented.

server/asset/btc/btc.go
```diff
@@ -506,7 +555,7 @@ func (btc *Backend) InitTxSizeBase() uint32 {
 
 // FeeRate returns the current optimal fee rate in sat / byte.
 func (btc *Backend) FeeRate(_ context.Context) (uint64, error) {
```
chappjc, line 557: Once we get around to updating `(*RPCClient).call` to accept a context we should pass it through `estimateFee`, esp. if we hit the "hard way" path.  In fact, a ctx could still be quite helpful in those loops juts by doing `if ctx.Err() != nil { return ctx.Err() }` in case it's shutdown or canceled so it doesn't keep trying `getrawtransaction`s e.g.

server/asset/btc/btc.go
```diff
@@ -1084,6 +1133,61 @@ out:
...
+func (btc *Backend) estimateFee() (satsPerB uint64, err error) {
+	if btc.cfg.DumbFeeEstimates {
+		satsPerB, err = btc.node.EstimateFee(btc.feeConfs)
+	} else {
+		satsPerB, err = btc.node.EstimateSmartFee(btc.feeConfs, &btcjson.EstimateModeConservative)
+	}
+	if err == nil && satsPerB > 0 {
+		return satsPerB, nil
+	} else if err != nil {
+		btc.log.Tracef("Estimatefee error for %s: %v", btc.name, err)
```
chappjc, line 1150: Just noticing that this could show `<nil>` for `err` if `satsPerB` were just zero.  Could be confusing, but meh, it's a TRC.

server/asset/btc/btc.go
```diff
@@ -127,9 +130,19 @@ type Backend struct {
...
+	// fee estimation configuration
+	feeConfs          int64
+	noCompetitionRate uint64
+
+	// The feeCache prevents repeated calculations of the median fee rate
+	// between block changes when estimate(smart)fee is unprimed.
+	feeCache struct {
+		sync.Mutex
+		fee  uint64
+		hash chainhash.Hash
```
chappjc, line 144: Nice pattern.

server/asset/btc/btc.go
```diff
@@ -1084,6 +1133,61 @@ out:
...
+	}
+
+	// Need to revert to the median fee calculation.
+	if btc.cfg.ManualMedianFee {
+		satsPerB, err = btc.node.medianFeesTheHardWay()
+	} else {
+		satsPerB, err = btc.node.medianFeeRate()
+	}
+	if err != nil {
+		if errors.Is(err, ErrNoCompetition) {
```
chappjc, line 1173: unexport `ErrNoCompetition`?

server/asset/btc/btc.go
```diff
@@ -194,9 +214,30 @@ type BackendCloneConfig struct {
...
+	DumbFeeEstimates bool
+	// Argsless fee estimates are for assets who don't take an argument for
+	// number of blocks to estimatefee.
+	ArglessFeeEstimates bool
+	// FeeConfs specifies the target number of confirmations to use for
+	// estimate(smart)fee. If not set, default value is 1,
+	FeeConfs int64
+	// MaxFeeBlocks is the maximum number of blocks that can be evaluated for
+	// median fee calculations. If > 100 txs are not seen in the last
+	// MaxFeeBlocks, an ErrNoCompetition is returned. Default value is 5.
```
chappjc, line 237: Makes sense, but the `ErrNoCompetition` seems to be internal, not really something of concern for consumer of `BackendCloneConfig`

chappjc (commented):
tACK.

Just needs a rebase to deal with conflicts from when I added the block deserializaer.  
We can go with your bool name, but I'd prefer my `anylist` approach.
Rebase branch I'm evaluating: https://github.com/chappjc/dcrdex/commit/abb0db65b0043d9e3c8265d180ba062603cf7bac

chappjc:
Oh, you could also update the hard way to use the msgblock, but it's fine
as is

chappjc (approved):

PR #1598, opened 2022-04-24
client: add electrum client

chappjc:
To be the RPC client for a future `walletTypeElectrum` wallet type.

This creates a zero-dep Electrum wallet RPC client package.
An "ExchangeWalletElectrum" will be created that uses this client.  Checking foreign/counterparty transactions is possible with the `getaddresshistory`, `getaddressunspent`, and `gettransaction` RPCs, so this solves confirmation checks as well as finding redemption.  The redemption search appears to be as easy as checking `getaddresshistory` for the swap contract address, as it lists any spending txns as well.

This is meant to support external Electrum light wallets, since there are rougly 30 altcoins with their own forks of Electrum.  Namely LTC.  Maybe DOGE will get on the stick and make an Electrum fork work too.

This is designed for the RPCs available in 4.0.9 since that's where Litecoin's fork is at.  4.2.0 has some other RPCs including freeeze_utxo, so maybe in the future that will be available.

### DIFF
--- a/client/asset/btc/electrum/client.go
+++ b/client/asset/btc/electrum/client.go
@@ -1,0 +1,172 @@
+// This code is available on the terms of the project LICENSE.md file,
+// also available online at https://blueoakcouncil.org/license/1.0.0.
+
+package electrum
+
+import (
+	"bytes"
+	"context"
+	"encoding/base64"
+	"encoding/json"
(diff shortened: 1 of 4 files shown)


chappjc:
The full shebang will replace this shortly.

PR #1599, opened 2022-04-25
client/eth: Get receipt before checking status.

### LINKED ISSUE #1577: client/eth: checkTxStatus returns a receipt on error sometimes.
Hit a panic while testing:
```
--- FAIL: TestContract (985.33s)
    --- PASS: TestContract/testSwap (2.10s)
    --- PASS: TestContract/testInitiate (102.21s)
    --- FAIL: TestContract/testRedeem (881.02s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x180 pc=0xc172a0]

goroutine 1973 [running]:
testing.tRunner.func1.2({0xcf6b20, 0x15662d0})
	/usr/lib/go/src/testing/testing.go:1389 +0x24e
testing.tRunner.func1()
	/usr/lib/go/src/testing/testing.go:1392 +0x39f
panic({0xcf6b20, 0x15662d0})
	/usr/lib/go/src/runtime/panic.go:838 +0x207
decred.org/dcrdex/client/asset/eth.testRedeem(0xc02e17ed00)
	/home/joe/git/dcrdex/client/asset/eth/nodeclient_harness_test.go:1214 +0x22a0
testing.tRunner(0xc02e17ed00, 0xf8f098)
	/usr/lib/go/src/testing/testing.go:1439 +0x102
created by testing.(*T).Run
	/usr/lib/go/src/testing/testing.go:1486 +0x35f
exit status 2
FAIL	decred.org/dcrdex/client/asset/eth	1069.039s
```

And here is this area:
https://github.com/decred/dcrdex/blob/11fe415b1218463254f5e9ee2cdce6e2c4480416/client/asset/eth/nodeclient_harness_test.go#L998-L1004

So, the method errored, but it was not the error that returns a receipt, then we tried to get the used gas which panic'd. I think the design of `checkTxStatus` needs to be changed to take a receipt and return only an error to avoid pitfalls like what was seen in the test.

JoeGruffins:
checkTxStatus would sometimes return a receipt on error, which caused a
    panic in tests sometimes. Get the receipt separately to avoid this.

closes #1577

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -2663,6 +2663,19 @@ func (w *assetWallet) isRefundable(secretHash [32]byte, contractVer uint32) (ref
 	})
 }
 
+func checkTxStatus(receipt *types.Receipt, gasLimit uint64) error {
+	if receipt.Status != types.ReceiptStatusSuccessful {
+		return fmt.Errorf("transaction status failed")
+
+	}
+
+	if receipt.GasUsed > gasLimit {
+		return fmt.Errorf("gas used, %d, appears to have exceeded limit of %d", receipt.GasUsed, gasLimit)
+	}
+
+	return nil
+}
+
 // getGasEstimate is used to get a gas table for an asset's contract(s). The
 // provided gases, g, should be generous estimates of what the gas might be.
 // Errors are thrown if the provided estimates are too small by more than a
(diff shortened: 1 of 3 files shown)


chappjc:
I got the impression this was used in upcoming work: https://github.com/decred/dcrdex/pull/1523#discussion_r842008980
@buck54321 you gave the thumbs up, but it wasn't clear if you meant "yeah, will be used" or "right, will remove".

chappjc:
NVM, I see it used in https://github.com/decred/dcrdex/pull/1399

JoeGruffins:
Oh shoot. Well... I'll go comment over there and convert this to draft for the time being.

JoeGruffins:
Forgot to comment on this. Fixing here.

JoeGruffins:
> could you point out the panic. 

https://github.com/decred/dcrdex/blob/e500d1c4f886dc775a95896d28693a73956dbc15/client/asset/eth/nodeclient.go#L383-L400

Imagine `n.transactionReceipt` fails here, as it usually does when testing on testnet. We get an error and nil receipt return, which is good. The second two errors will return a populated receipt and an error.

Now here in tests:
https://github.com/decred/dcrdex/blob/e500d1c4f886dc775a95896d28693a73956dbc15/client/asset/eth/nodeclient_harness_test.go#L1065-L1072

So if a test where is `test.success` was false and there was an error it goes to the second condition and tries to take the gas used which causes the panic.

The test was expected to fail, but failed for an unexpected reason. I think that simply sticking to the "rule" that if error expect bad returns is best and solves the problem.

PR #1600, opened 2022-04-28
multi: support fiat conversion for assets

### LINKED ISSUE #1579: client/core: fiat conversions
It'd be nice to display fiat conversions next to values in some places.

- Next to their balance on the wallets page. 
- Order quantity on order confirmation form
- Withdraw form

Ideally, we'd have busy stablecoin markets and we could do this without oracles. But in the meantime, we could grab the prices through available public APIs. For Decred and Bitcoin, we can use [dcrdata](https://explorer.dcrdata.org/api/exchangerate), which already does a weighted average of markets. For others though, we'll need to find new endpoint.

#### Design considerations

Public APIs will have rate limits and API calls will incur significant delays. We'll need to keep cached values and only update the cache periodically. 

On the front-end, we'll need to be flexible, only showing fiat conversion when they're available.

~~We **could** implement through the `asset.Wallet`s as an optional `interface`~~ (nah, we should use **core**, see below), e.g.

```Go
type PriceFetcher interface {
    FetchPrice(ctx context.Context) (uint64, error)
}
```

We would have to define a canonical fiat unit, almost certainly USD, and wallets will return values in `USD/XYZ`. Or maybe `FetchPrice` should take a `fiat string` argument. 

I could also be a mostly **core**-level system, which makes more sense if we only rely on one or two APIs for the prices.

#### Where do we get data for other coins?

We **could** try to calculate it ourselves. For example, if we know the fiat exchange rate of BTC and DCR, then we can **kind of** know the exchange rate for any assets that share any markets with those, via a conversion using the mid-gap rate. For example, if we have a BTC-ETH market, then we can get an approximation for ETH by taking `ethPriceUSD = btcPriceUSD / midGap`. It won't be the best price, especially if the market is dead. And attempting to chain that logic by, say, figuring the `BCH` price from our `ETH-BCH` using the aforementioned ETH price would potentially chain and magnify the "error".

I've used [coinpaprika](https://api.coinpaprika.com/) before, and as long as you meter your requests, it seems pretty solid.
It looks like [messari.io](https://messari.io/api/docs) has a non-authenticated API that allows up to 20 requests per minute.

ukane-philemon:
This enables the conversion of assets value to fiat equivalent. Configuring fiat conversion is UI only. closes #1579

- [x] Modify wallets page to display fiat value for the asset if fiat price is available.
- [x] Show fiat value for send amount on the send form.
- [x] Show fiat value for order amount on the order confirmation form
- [x] Modify the settings page to display exchange rate sources and configuration options for fiat conversion.

Wallet view
usddisplaybal

Market view
usdmk1
usdmk2
usdmk3

Withdraw/Send Form view
usdwithdrawval

Exchange rate source settings view
Screenshot from 2022-06-28 20-07-47

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1135,6 +1135,12 @@ type Core struct {
 
 	sentCommitsMtx sync.Mutex
 	sentCommits    map[order.Commitment]chan struct{}
+
+	ratesMtx        sync.RWMutex
+	fiatRateSources map[string]*commonRateSource
+	// stopFiatRateFetching will be used to shutdown fetchFiatExchangeRates
+	// goroutine when all rate sources have been disabled.
+	stopFiatRateFetching context.CancelFunc
 }
 
 // New is the constructor for a new Core.
@@ -1227,6 +1233,8 @@ func New(cfg *Config) (*Core, error) {
 		locale:             locale,
 		localePrinter:      message.NewPrinter(lang),
 		seedGenerationTime: seedGenerationTime,
+
+		fiatRateSources: make(map[string]*commonRateSource),
 	}
 
 	// Populate the initial user data. User won't include any DEX info yet, as
@@ -1258,6 +1266,39 @@ func (c *Core) Run(ctx context.Context) {
 		defer c.wg.Done()
 		c.latencyQ.Run(ctx)
 	}()
+
+	// Skip rate fetch setup if on simnet. Rate fetching maybe enabled if
+	// desired.
+	if c.cfg.Net != dex.Simnet {
+		c.ratesMtx.Lock()
+		// Retrieve disabled fiat rate sources from database.
+		disabledSources, err := c.db.DisabledRateSources()
+		if err != nil {
+			c.log.Errorf("Unable to retrieve disabled fiat rate source: %v", err)
+		}
+
+		// Construct enabled fiat rate sources.
+		for token, rateFetcher := range fiatRateFetchers {
+			var isDisabled bool
+			for _, v := range disabledSources {
+				if token == v {
+					isDisabled = true
+					break
+				}
+			}
+			if !isDisabled {
+				c.fiatRateSources[token] = newCommonRateSource(rateFetcher)
+			}
+		}
+
+		// Start goroutine for fiat rate fetcher's if we have at least one source.
+		if len(c.fiatRateSources) != 0 {
+			c.fetchFiatExchangeRates()
+		} else {
+			c.log.Debug("no fiat rate source initialized")
+		}
+		c.ratesMtx.Unlock()
+	}
 
 	c.wg.Wait() // block here until all goroutines except DB complete
 
@@ -1707,6 +1748,7 @@ func (c *Core) User() *User {
 		Exchanges:          c.Exchanges(),
 		Initialized:        c.IsInitialized(),
 		SeedGenerationTime: c.seedGenerationTime,
+		FiatRates:          c.fiatConversions(),
 	}
 }
 
@@ -7722,4 +7764,224 @@ func (c *Core) findActiveOrder(oid order.OrderID) (*trackedTrade, error) {
 		}
 	}
 	return nil, fmt.Errorf("could not find active order with order id: %s", oid)
+}
+
+// fetchFiatExchangeRates starts the fiat rate fetcher goroutine and schedules
+// refresh cycles. Use under ratesMtx lock.
+func (c *Core) fetchFiatExchangeRates() {
+	if c.stopFiatRateFetching != nil {
+		c.log.Debug("Fiat exchange rate fetching is already enabled")
+		return
+	}
+	ctx, cancel := context.WithCancel(c.ctx)
+	c.stopFiatRateFetching = cancel
+
+	c.wg.Add(1)
+	go func() {
+		defer c.wg.Done()
+		tick := time.NewTimer(0) // starts rate fetching immediately.
+		for {
+			select {
+			case <-ctx.Done():
+				tick.Stop()
+				return
+			case <-tick.C:
+				c.refreshFiatRates(ctx)
+			}
+			tick = c.nextRateSourceTick()
+		}
+	}()
+}
+
+// refreshFiatRates refreshes the fiat rates for rate sources whose values have
+// not been updated since fiatRateRequestInterval.
+func (c *Core) refreshFiatRates(ctx context.Context) {
+	ctx, cancel := context.WithTimeout(ctx, 4*time.Second)
+	defer cancel()
+
+	var wg sync.WaitGroup
+	supportedAssets := c.SupportedAssets()
+	timeNow := time.Now()
+	c.ratesMtx.Lock()
+	for _, source := range c.fiatRateSources {
+		if time.Since(source.lastRequest) > fiatRateRequestInterval {
+			wg.Add(1)
+			go func(source *commonRateSource) {
+				source.lastRequest = timeNow
+				defer wg.Done()
+				source.refreshRates(ctx, c.log, supportedAssets)
+			}(source)
+		}
+	}
+	c.ratesMtx.Unlock()
+	wg.Wait()
+
+	fiatRatesMap := c.fiatConversions()
+	if len(fiatRatesMap) != 0 {
+		c.notify(newFiatRatesUpdate(fiatRatesMap))
+	}
+}
+
+// nextRateSourceTick checks the fiat rate source's last request, and calculates
+// when the next cycle should run.
+func (c *Core) nextRateSourceTick() *time.Timer {
+	c.ratesMtx.RLock()
+	defer c.ratesMtx.RUnlock()
+
+	minTick := 10 * time.Second
+	tOldest := time.Now()
+	for _, source := range c.fiatRateSources {
+		t := source.lastRequest
+		if t.Before(tOldest) {
+			tOldest = t
+		}
+	}
+	tilNext := fiatRateRequestInterval - time.Since(tOldest)
+	if tilNext < minTick {
+		tilNext = minTick
+	}
+	return time.NewTimer(tilNext)
+}
+
+// FiatRateSources returns a list of fiat rate sources and their individual
+// status.
+func (c *Core) FiatRateSources() map[string]bool {
+	c.ratesMtx.RLock()
+	defer c.ratesMtx.RUnlock()
+	rateSources := make(map[string]bool, len(fiatRateFetchers))
+	for token := range fiatRateFetchers {
+		rateSources[token] = c.fiatRateSources[token] != nil
+	}
+	return rateSources
+}
+
+// fiatConversions returns fiat rate for all supported assets that have a
+// wallet. It also checks if fiat rates are expired and does some clean-up.
+func (c *Core) fiatConversions() map[uint32]float64 {
+	supportedAssets := c.SupportedAssets()
+
+	c.ratesMtx.RLock()
+	defer c.ratesMtx.RUnlock()
+	disabledCount := 0
+	fiatRatesMap := make(map[uint32]float64, len(supportedAssets))
+	for assetID := range supportedAssets {
+		var rateSum float64
+		var sources int32
+		for token, source := range c.fiatRateSources {
+			// Remove fiat rate source with expired exchange rate data.
+			if source.isExpired(fiatRateDataExpiry) {
+				delete(c.fiatRateSources, token)
+				disabledCount++
+				continue
+			}
+			rateInfo := source.assetRate(assetID)
+			if rateInfo != nil && time.Since(rateInfo.lastUpdate) < fiatRateDataExpiry {
+				sources++
+				rateSum += rateInfo.rate
+			}
+		}
+		if rateSum != 0 {
+			fiatRatesMap[assetID] = rateSum / float64(sources) // get average rate.
+		}
+	}
+
+	// Ensure disabled fiat rate fetchers are saved to database.
+	if disabledCount > 0 {
+		c.saveDisabledRateSources()
+	}
+
+	return fiatRatesMap
+}
+
+// ToggleRateSourceStatus toggles a fiat rate source status. If disable is true,
+// the fiat rate source is disabled, otherwise the rate source is enabled.
+func (c *Core) ToggleRateSourceStatus(source string, disable bool) error {
+	if disable {
+		return c.disableRateSource(source)
+	}
+	return c.enableRateSource(source)
+}
+
+// enableRateSource enables a fiat rate source.
+func (c *Core) enableRateSource(source string) error {
+	c.ratesMtx.Lock()
+	defer c.ratesMtx.Unlock()
+
+	// Check if it's an invalid rate source or it is already enabled.
+	rateFetcher, found := fiatRateFetchers[source]
+	if !found {
+		return errors.New("cannot enable unknown fiat rate source")
+	} else if c.fiatRateSources[source] != nil {
+		return nil // already enabled.
+	}
+
+	// Build fiat rate source.
+	c.fiatRateSources[source] = newCommonRateSource(rateFetcher)
+
+	// If this is our first fiat rate source, start fiat rate fetcher goroutine,
+	// else fetch rates.
+	if len(c.fiatRateSources) == 1 {
+		c.fetchFiatExchangeRates()
+	} else {
+		source := c.fiatRateSources[source]
+		supportedAssets := c.SupportedAssets()
+		go func(source *commonRateSource) {
+			ctx, cancel := context.WithTimeout(c.ctx, 4*time.Second)
+			defer cancel()
+			source.refreshRates(ctx, c.log, supportedAssets)
+		}(source)
+	}
+
+	// Update disabled fiat rate source.
+	c.saveDisabledRateSources()
+
+	c.log.Infof("Enabled %s to fetch fiat rates.", source)
+	return nil
+}
+
+// disableRateSource disables a fiat rate source.
+func (c *Core) disableRateSource(source string) error {
+	c.ratesMtx.Lock()
+	defer c.ratesMtx.Unlock()
+
+	// Check if it's an invalid fiat rate source or it is already
+	// disabled.
+	_, found := fiatRateFetchers[source]
+	if !found {
+		return errors.New("cannot disable unknown fiat rate source")
+	} else if c.fiatRateSources[source] == nil {
+		return nil // already disabled.
+	}
+
+	// Remove fiat rate source.
+	delete(c.fiatRateSources, source)
+
+	// Save disabled fiat rate sources to database.
+	c.saveDisabledRateSources()
+
+	c.log.Infof("Disabled %s from fetching fiat rates.", source)
+	return nil
+}
+
+// saveDisabledRateSources saves disabled fiat rate sources to database and
+// shuts down rate fetching if there are no exchange rate source. Use under
+// ratesMtx lock.
+func (c *Core) saveDisabledRateSources() {
+	var disabled []string
+	for token := range fiatRateFetchers {
+		if c.fiatRateSources[token] == nil {
+			disabled = append(disabled, token)
+		}
+	}
+
+	// Shutdown rate fetching if there are no exchange rate source.
+	if len(c.fiatRateSources) == 0 && c.stopFiatRateFetching != nil {
+		c.stopFiatRateFetching()
+		c.stopFiatRateFetching = nil
+	}
+
+	err := c.db.SaveDisabledRateSources(disabled)
+	if err != nil {
+		c.log.Errorf("Unable to save disabled fiat rate source to database: %v", err)
+	}
 }
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1278,17 +1278,14 @@ func (c *Core) Run(ctx context.Context) {
 		}
 
 		// Construct enabled fiat rate sources.
+	fetchers:
 		for token, rateFetcher := range fiatRateFetchers {
-			var isDisabled bool
 			for _, v := range disabledSources {
 				if token == v {
-					isDisabled = true
-					break
+					continue fetchers
 				}
 			}
-			if !isDisabled {
-				c.fiatRateSources[token] = newCommonRateSource(rateFetcher)
-			}
+			c.fiatRateSources[token] = newCommonRateSource(rateFetcher)
 		}
 
 		// Start goroutine for fiat rate fetcher's if we have at least one source.
@@ -7776,71 +7773,53 @@ func (c *Core) fetchFiatExchangeRates() {
 	ctx, cancel := context.WithCancel(c.ctx)
 	c.stopFiatRateFetching = cancel
 
+	c.log.Debug("starting fiat rate fetching")
+
 	c.wg.Add(1)
 	go func() {
 		defer c.wg.Done()
-		tick := time.NewTimer(0) // starts rate fetching immediately.
+		tick := time.NewTimer(fiatRateRequestInterval)
+		defer tick.Stop()
 		for {
+
+			c.refreshFiatRates(ctx)
+
 			select {
 			case <-ctx.Done():
-				tick.Stop()
 				return
 			case <-tick.C:
-				c.refreshFiatRates(ctx)
 			}
-			tick = c.nextRateSourceTick()
 		}
 	}()
 }
 
 // refreshFiatRates refreshes the fiat rates for rate sources whose values have
-// not been updated since fiatRateRequestInterval.
+// not been updated since fiatRateRequestInterval. It also checks if fiat rates
+// are expired and does some clean-up.
 func (c *Core) refreshFiatRates(ctx context.Context) {
 	ctx, cancel := context.WithTimeout(ctx, 4*time.Second)
 	defer cancel()
 
 	var wg sync.WaitGroup
 	supportedAssets := c.SupportedAssets()
-	timeNow := time.Now()
-	c.ratesMtx.Lock()
+	c.ratesMtx.RLock()
 	for _, source := range c.fiatRateSources {
-		if time.Since(source.lastRequest) > fiatRateRequestInterval {
-			wg.Add(1)
-			go func(source *commonRateSource) {
-				source.lastRequest = timeNow
-				defer wg.Done()
-				source.refreshRates(ctx, c.log, supportedAssets)
-			}(source)
-		}
+		wg.Add(1)
+		go func(source *commonRateSource) {
+			defer wg.Done()
+			source.refreshRates(ctx, c.log, supportedAssets)
+		}(source)
 	}
-	c.ratesMtx.Unlock()
+	c.ratesMtx.RUnlock()
 	wg.Wait()
+
+	// Remove expired rate source if any.
+	c.removeExpiredRateSources()
 
 	fiatRatesMap := c.fiatConversions()
 	if len(fiatRatesMap) != 0 {
 		c.notify(newFiatRatesUpdate(fiatRatesMap))
 	}
-}
-
-// nextRateSourceTick checks the fiat rate source's last request, and calculates
-// when the next cycle should run.
-func (c *Core) nextRateSourceTick() *time.Timer {
-	c.ratesMtx.RLock()
-	defer c.ratesMtx.RUnlock()
-
-	minTick := 10 * time.Second
-	tOldest := time.Now()
-	for _, source := range c.fiatRateSources {
-		t := source.lastRequest
-		if t.Before(tOldest) {
-			tOldest = t
-		}
-	}
-	tilNext := fiatRateRequestInterval - time.Since(tOldest)
-	if tilNext < minTick {
-		tilNext = minTick
-	}
-	return time.NewTimer(tilNext)
 }
 
 // FiatRateSources returns a list of fiat rate sources and their individual
@@ -7856,24 +7835,17 @@ func (c *Core) FiatRateSources() map[string]bool {
 }
 
 // fiatConversions returns fiat rate for all supported assets that have a
-// wallet. It also checks if fiat rates are expired and does some clean-up.
+// wallet.
 func (c *Core) fiatConversions() map[uint32]float64 {
-	supportedAssets := c.SupportedAssets()
+	supportedAssets := asset.Assets()
 
 	c.ratesMtx.RLock()
 	defer c.ratesMtx.RUnlock()
-	disabledCount := 0
 	fiatRatesMap := make(map[uint32]float64, len(supportedAssets))
 	for assetID := range supportedAssets {
 		var rateSum float64
-		var sources int32
-		for token, source := range c.fiatRateSources {
-			// Remove fiat rate source with expired exchange rate data.
-			if source.isExpired(fiatRateDataExpiry) {
-				delete(c.fiatRateSources, token)
-				disabledCount++
-				continue
-			}
+		var sources int
+		for _, source := range c.fiatRateSources {
 			rateInfo := source.assetRate(assetID)
 			if rateInfo != nil && time.Since(rateInfo.lastUpdate) < fiatRateDataExpiry {
 				sources++
@@ -7884,12 +7856,6 @@ func (c *Core) fiatConversions() map[uint32]float64 {
 			fiatRatesMap[assetID] = rateSum / float64(sources) // get average rate.
 		}
 	}
-
-	// Ensure disabled fiat rate fetchers are saved to database.
-	if disabledCount > 0 {
-		c.saveDisabledRateSources()
-	}
-
 	return fiatRatesMap
 }
 
@@ -7904,32 +7870,33 @@ func (c *Core) ToggleRateSourceStatus(source string, disable bool) error {
 
 // enableRateSource enables a fiat rate source.
 func (c *Core) enableRateSource(source string) error {
-	c.ratesMtx.Lock()
-	defer c.ratesMtx.Unlock()
-
 	// Check if it's an invalid rate source or it is already enabled.
 	rateFetcher, found := fiatRateFetchers[source]
 	if !found {
 		return errors.New("cannot enable unknown fiat rate source")
-	} else if c.fiatRateSources[source] != nil {
+	}
+
+	c.ratesMtx.Lock()
+	defer c.ratesMtx.Unlock()
+	if c.fiatRateSources[source] != nil {
 		return nil // already enabled.
 	}
 
 	// Build fiat rate source.
-	c.fiatRateSources[source] = newCommonRateSource(rateFetcher)
+	rateSource := newCommonRateSource(rateFetcher)
+	c.fiatRateSources[source] = rateSource
 
 	// If this is our first fiat rate source, start fiat rate fetcher goroutine,
 	// else fetch rates.
 	if len(c.fiatRateSources) == 1 {
 		c.fetchFiatExchangeRates()
 	} else {
-		source := c.fiatRateSources[source]
-		supportedAssets := c.SupportedAssets()
-		go func(source *commonRateSource) {
+		go func() {
+			supportedAssets := c.SupportedAssets() // not with ratesMtx locked!
 			ctx, cancel := context.WithTimeout(c.ctx, 4*time.Second)
 			defer cancel()
-			source.refreshRates(ctx, c.log, supportedAssets)
-		}(source)
+			rateSource.refreshRates(ctx, c.log, supportedAssets)
+		}()
 	}
 
 	// Update disabled fiat rate source.
@@ -7941,15 +7908,17 @@ func (c *Core) enableRateSource(source string) error {
 
 // disableRateSource disables a fiat rate source.
 func (c *Core) disableRateSource(source string) error {
-	c.ratesMtx.Lock()
-	defer c.ratesMtx.Unlock()
-
 	// Check if it's an invalid fiat rate source or it is already
 	// disabled.
 	_, found := fiatRateFetchers[source]
 	if !found {
 		return errors.New("cannot disable unknown fiat rate source")
-	} else if c.fiatRateSources[source] == nil {
+	}
+
+	c.ratesMtx.Lock()
+	defer c.ratesMtx.Unlock()
+
+	if c.fiatRateSources[source] == nil {
 		return nil // already disabled.
 	}
 
@@ -7961,6 +7930,27 @@ func (c *Core) disableRateSource(source string) error {
 
 	c.log.Infof("Disabled %s from fetching fiat rates.", source)
 	return nil
+}
+
+// removeExpiredRateSources disables expired fiat rate source.
+func (c *Core) removeExpiredRateSources() {
+	c.ratesMtx.Lock()
+	defer c.ratesMtx.Unlock()
+
+	// Remove fiat rate source with expired exchange rate data.
+	var disabledSources []string
+	for token, source := range c.fiatRateSources {
+		if source.isExpired(fiatRateDataExpiry) {
+			delete(c.fiatRateSources, token)
+			disabledSources = append(disabledSources, token)
+		}
+	}
+
+	// Ensure disabled fiat rate fetchers are saved to database.
+	if len(disabledSources) > 0 {
+		c.saveDisabledRateSources()
+		c.log.Warnf("Expired rate source(s) has been disabled: %v", strings.Join(disabledSources, ", "))
+	}
 }
 
 // saveDisabledRateSources saves disabled fiat rate sources to database and
@@ -7978,6 +7968,7 @@ func (c *Core) saveDisabledRateSources() {
 	if len(c.fiatRateSources) == 0 && c.stopFiatRateFetching != nil {
 		c.stopFiatRateFetching()
 		c.stopFiatRateFetching = nil
+		c.log.Debug("shutting down rate fetching")
 	}
 
 	err := c.db.SaveDisabledRateSources(disabled)
--- a/client/core/core_test.go
+++ b/client/core/core_test.go
@@ -32,7 +32,6 @@ import (
 	"decred.org/dcrdex/dex/encrypt"
 	"decred.org/dcrdex/dex/msgjson"
 	"decred.org/dcrdex/dex/order"
-	"decred.org/dcrdex/dex/order/test"
 	ordertest "decred.org/dcrdex/dex/order/test"
 	"decred.org/dcrdex/dex/wait"
 	"decred.org/dcrdex/server/account"
@@ -530,7 +529,12 @@ func (tdb *TDB) SetSeedGenerationTime(time uint64) error {
 func (tdb *TDB) SeedGenerationTime() (uint64, error) {
 	return 0, nil
 }
-
+func (tdb *TDB) DisabledRateSources() ([]string, error) {
+	return nil, nil
+}
+func (tdb *TDB) SaveDisabledRateSources(disableSources []string) error {
+	return nil
+}
 func (tdb *TDB) Recrypt(creds *db.PrimaryCredentials, oldCrypter, newCrypter encrypt.Crypter) (
 	walletUpdates map[uint32][]byte, acctUpdates map[string][]byte, err error) {
 
@@ -1084,6 +1088,13 @@ func randomMsgMarket() (baseAsset, quoteAsset *msgjson.Asset) {
 	return randomAsset(), randomAsset()
 }
 
+func tFetcher(_ context.Context, log dex.Logger, _ map[uint32]*SupportedAsset) map[uint32]float64 {
+	fiatRates := make(map[uint32]float64)
+	fiatRates[tUTXOAssetA.ID] = 45
+	fiatRates[tUTXOAssetB.ID] = 32000
+	return fiatRates
+}
+
 type testRig struct {
 	shutdown func()
 	core     *Core
@@ -1158,6 +1169,8 @@ func newTestRig() *testRig {
 
 			locale:        enUS,
 			localePrinter: message.NewPrinter(language.AmericanEnglish),
+
+			fiatRateSources: make(map[string]*commonRateSource),
 		},
 		db:      tdb,
 		queue:   queue,
@@ -4839,7 +4852,7 @@ func TestReconcileTrades(t *testing.T) {
 			clientOrders: []*trackedTrade{},
 			serverOrders: []*msgjson.OrderStatus{
 				{
-					ID:     test.RandomOrderID().Bytes(),
+					ID:     ordertest.RandomOrderID().Bytes(),
 					Status: uint16(order.OrderStatusBooked),
 				},
 			},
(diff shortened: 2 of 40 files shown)


client/core/exchangeratefetcher.go
```diff
@@ -0,0 +1,276 @@
...
+func (fetcher *commonFetcher) fetch(request *http.Request, response interface{}) error {
+	resp, err := fetcher.client.Do(request)
+	if err != nil {
+		return fmt.Errorf(fmt.Sprintf("Request failed: %v", err))
+	}
+	defer resp.Body.Close()
+	// Read the raw bytes and close the response.
+	err = json.NewDecoder(resp.Body).Decode(response)
+	if err != nil {
+		return fmt.Errorf(fmt.Sprintf("Failed to decode json from %s: %v", request.URL.String(), err))
```
JoeGruffins, line 124: No need for `fmt.Sprintf`

client/core/exchangeratefetcher.go
```diff
@@ -0,0 +1,276 @@
...
+		return
+	}
+	err = fetcher.fetch(request, res)
+	if err != nil {
+		fetcher.log.Errorf("%s: Fetch %v\n", fetcher.token, err)
+		return
+	}
+
+	fetcher.updateRate(btcBipId, res.BtcPrice)
+	fetcher.updateRate(dcrBipId, res.DcrPrice)
```
JoeGruffins, line 226: I guess it's possible the client only has one of these wallets. Could not do both in that case.

client/db/bolt/db.go
```diff
@@ -2075,6 +2076,40 @@ func (db *BoltDB) DeleteInactiveMatches(ctx context.Context, olderThan *time.Tim
...
+		if bkt == nil {
+			return fmt.Errorf("failed to open %s bucket", string(appBucket))
+		}
+		disabledByte := []byte(strings.Join(disabledFetchers, ","))
+		return bkt.Put(disabledRateFetchersKey, disabledByte)
+	})
+}
+
+// DisabledRatefetchers retreives a map of disabled exchange rate fetchers.
+func (db *BoltDB) DisabledRatefetchers() (disabledFetchers []string, err error) {
```
JoeGruffins, line 2092: In other places Fetcher will be capitalized, so I suggest `DisabledRateFetchers` here as well.

client/db/bolt/db.go
```diff
@@ -2075,6 +2076,40 @@ func (db *BoltDB) DeleteInactiveMatches(ctx context.Context, olderThan *time.Tim
...
+		}
+		disabledByte := bkt.Get(disabledRateFetchersKey)
+		disabledString := string(disabledByte)
+		if disabledString == "" {
+			return nil
+		}
+		disabled := strings.Split(disabledString, ",")
+		for _, token := range disabled {
+			if token != "" {
+				disabledFetchers = append(disabledFetchers, token)
```
JoeGruffins, line 2106: Could you create this slice before the loop? `disabledFetchers = make([]string, len(disabled))`

JoeGruffins (commented):
Just some comments on the code. Concept looks fine to me.

client/core/core.go
```diff
@@ -7406,3 +7439,233 @@ func (c *Core) findActiveOrder(oid order.OrderID) (*trackedTrade, error) {
...
+	}
+	tilNext := fiatRateRequestInterval - time.Since(tOldest)
+	if tilNext < minTick {
+		tilNext = minTick
+	}
+	return time.NewTimer(tilNext)
+}
+
+// ExchangeRateFetchers returns a list of exchange rate fetchers and their
+// individual status.
```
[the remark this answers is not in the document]
ukane-philemon: > Is status here disabled or not?

disabled and enabled rate fetchers.
The frontend use this map to display the settings checkbox for ratefetchers.
True `check`
False `uncheck`

client/core/core.go
```diff
@@ -7406,3 +7439,233 @@ func (c *Core) findActiveOrder(oid order.OrderID) (*trackedTrade, error) {
...
+	// Build exchange rate fetcher.
+	core.exchangeRateFetchers[source] = fetcherConstructor(core.log)
+
+	// If this is our first rate fetcher, start exchange rate fetcher
+	// gorountine, else fetch rates.
+	if len(core.exchangeRateFetchers) == 1 {
+		core.fetchExchangeRates()
+	} else {
+		go core.exchangeRateFetchers[source].FetchRates(core.assetMap())
+	}
```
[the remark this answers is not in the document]
ukane-philemon: > Which is maybe supposed to stop the loop there, but could a fetcher be removed and added in the interval so that the length of the fetchers is never 0 for the already running core.fetchExchangeRates()?
Maybe simplest to just run the one loop regardless and for the life of core.


Yes, this allows users to toggle the rate fetchers as they deem fit without shutting down dexc. Also, we don't want to keep running a goroutine that does nothing?, assuming the user disabled all rate fetchers on the fly and then wants to enable rate fetching after some time (could be a very long time)?.

client/core/exchangeratefetcher.go
```diff
@@ -0,0 +1,276 @@
...
+
+type dcrDataSource struct {
+	*commonFetcher
+}
+
+type dcrdataResponse struct {
+	Currency string  `json:"btcIndex"`
+	DcrPrice float64 `json:"dcrPrice"`
+	BtcPrice float64 `json:"btcPrice"`
+}
```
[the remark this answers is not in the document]
ukane-philemon: Maybe it's not.

client/core/core.go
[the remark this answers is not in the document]
JoeGruffins: It's my opinion that simplest is always best to reduce bugs. Not only bugs now, but bugs later when someone else touches the code.

This looks like it has a bug to me but I'm not entirely sure as I have not tested it irl.

I don't think that a loop that sometimes returns immediately every ten seconds is that bad.

Turning on/off the loop is also fine imo, as long as it works as intended.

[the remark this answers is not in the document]
ukane-philemon: @JoeGruffins, thanks for reviewing.

client/core/exchangeratefetcher.go
```diff
@@ -0,0 +1,297 @@
...
+// and response information.
+type coinpaprikaResponse struct {
+	Name   string `json:"name"`
+	Symbol string `json:"symbol"`
+	Quotes struct {
+		Currency struct {
+			Price float64 `json:"price"`
+		} `json:"USD"`
+	} `json:"quotes"`
+}
```
buck54321, line 165: If these response types are only used in `FetchRates`, lets define them there.

client/core/notification.go
```diff
@@ -14,20 +14,21 @@ import (
...
-	NoteTypeUpgrade      = "upgrade"
-	NoteTypeDEXAuth      = "dex_auth"
+	NoteTypeFeePayment       = "feepayment"
+	NoteTypeWithdraw         = "withdraw"
+	NoteTypeOrder            = "order"
+	NoteTypeMatch            = "match"
+	NoteTypeEpoch            = "epoch"
+	NoteTypeConnEvent        = "conn"
+	NoteTypeBalance          = "balance"
+	NoteTypeAssetRatesUpdate = "assetRatesUpdate"
```
buck54321, line 24: Make the name shorter, please. Maybe `NoteTypeFiatRates` or something. And let's keep with the trend here and avoid camel-case for the value. Maybe `"fiatrateupdate"`.

client/core/exchangeratefetcher.go
```diff
@@ -0,0 +1,297 @@
...
+	// Tokens. Used to identify exchange rate source.
+	messari      = "messari"
+	coinpaprika  = "coinpaprika"
+	dcrataDotOrg = "dcrdata"
+)
+
+var (
+	dcrDataURL     = "https://explorer.dcrdata.org/api/exchangerate"
+	coinpaprikaURL = "https://api.coinpaprika.com/v1/tickers/%s"
+	messariURL     = "https://data.messari.io/api/v1/assets/%s/metrics"
```
buck54321, line 35: `https://data.messari.io/api/v1/assets/{assetKey}/metrics/market-data` seems more efficient.

buck54321 (changes requested):
First pass. Lots of good stuff here.

chappjc:
No rush @ukane-philemon, just wanted to bump since the ball is in your court with current reviews, in case you missed.

client/core/exchangeratefetcher.go
```diff
@@ -0,0 +1,297 @@
...
+	"strings"
+	"sync"
+	"time"
+
+	"decred.org/dcrdex/dex"
+)
+
+const (
+	// DefaultFiatCurrency is the currency for displaying assets fiat value.
+	DefaultFiatCurrency = "USD"
```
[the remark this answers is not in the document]
ukane-philemon: we need `DefaultFiatCurrency` in the settings view. Any thoughts?

client/core/exchangeratefetcher.go
ukane-philemon: @buck54321, thanks for looking into this. Just noticed `https://data.messari.io/api/v1/assets/{assetKey}/metrics` returns too many unused information.

ukane-philemon:
Send Form:
sendValue

client/webserver/site/src/js/app.ts
```diff
@@ -549,15 +559,37 @@ export default class Application {
...
-        if (w) w.balance = n.balance
-        break
+        if (!w) break
+        w.balance = n.balance
+        const rate = this.fiatRatesMap[w.assetID]
+        const fiatDiv = this.main.querySelector(`[data-conversion-target="${n.assetID}"]`) as PageElement
+        if (!fiatDiv || !fiatDiv.firstElementChild) return
+        fiatDiv.firstElementChild.textContent = Doc.formatAssetValue(n.balance.available, rate, this.unitInfo(n.assetID))
+        if (rate && !this.disableConversion) Doc.show(fiatDiv)
+        else Doc.hide(fiatDiv)
```
buck54321, line 590: This 5-line block belongs in `wallets.ts`, since that's the only place that has the `data-conversion-target` elements. You already have it there, but you can delete it from **app.ts** in two places.

client/webserver/site/src/js/doc.ts
```diff
@@ -235,6 +235,18 @@ export default class Doc {
     return fullPrecisionFormatter(prec).format(v)
   }
 
+  /*
+   * formatAssetValue formats the value in atomic units to its representation in
+   * conventional units and returns the fiat value as a string.
+   */
+  static formatAssetValue (vAtomic: number, rate: number, unitInfo?: UnitInfo): string {
```
buck54321, line 242: Maybe `formatFiatConversion`.

client/webserver/site/src/js/wallets.ts
```diff
@@ -199,13 +201,39 @@ export default class WalletsPage extends BasePage {
...
     })
 
+    for (const [assetID, wallet] of (Object.entries(app().walletMap) as [any, WalletState][])) {
+      if (!wallet) continue
+      const rate = app().fiatRatesMap[assetID]
+      const fiatDiv = this.page.walletTable.querySelector(`[data-conversion-target="${assetID}"]`) as PageElement
+      if (!fiatDiv || !fiatDiv.firstElementChild) continue
+      fiatDiv.firstElementChild.textContent = Doc.formatAssetValue(wallet.balance.available, rate, app().unitInfo(assetID))
+      if (rate && !app().disableConversion) Doc.show(fiatDiv)
+      else Doc.hide(fiatDiv)
```
buck54321, line 223: This block is repeated 3 times in `MarketsPage` methods. How about a little helper method.

client/webserver/live_test.go
```diff
@@ -1612,6 +1612,12 @@ func (c *TCore) UpdateCert(string, []byte) error {
 func (c *TCore) UpdateDEXHost(string, string, []byte, interface{}) (*core.Exchange, error) {
 	return nil, nil
 }
```
buck54321, line 1614: Do me a favor, and add these changes 

https://github.com/buck54321/dcrdex/compare/8b805d37e403882a0f0732b21e8d481b110b0540...buck54321:ukane-fiat

client/webserver/site/src/html/wallets.tmpl
```diff
@@ -71,9 +71,16 @@
...
           </td>
-          <td data-balance-target="{{.ID}}">
+         <td>
             {{if .Wallet}}
+            <div data-balance-target="{{.ID}}">
               {{.Info.UnitInfo.ConventionalString .Wallet.Balance.Available}}
+            </div>
+            <div
+             data-conversion-target={{.ID}}
+             class="d-inline d-hide"> &dollar;
```
buck54321, line 81: Let's get rid of the dollar sign and write USD instead.

client/webserver/site/src/html/markets.tmpl
```diff
@@ -449,17 +449,20 @@
             <span class="grey fs17 flex-grow-1 text-start">[[[Quantity]]]</span>
             <span id="vQty" class="fs18 demi"></span>
             <span data-unit="base" class="grey fs14 ms-2"></span>
+            <span id="vFiatQty" class="d-hide"> / &dollar; <span></span></span>
           </div>
           <div class="d-flex align-items-center mt-1">
             <span class="grey fs17 flex-grow-1 text-start">[[[Total]]]</span>
             <span id="vTotal" class="fs18 demi"></span>
             <span data-unit="quote" class="grey fs14 ms-2"></span>
+            <span id="vFiatTotal" class="d-hide"> / &dollar; <span></span></span>
```
buck54321, line 458: These two figures should be relatively equivalent. We only need to show the approximate fiat value once. How about a new row? Or just put the figure in small grey text at the bottom right of this section. And prefer "USD" to a dollar sign.

buck54321 (changes requested):

client/core/core.go
```diff
@@ -7446,3 +7481,215 @@ func (c *Core) findActiveOrder(oid order.OrderID) (*trackedTrade, error) {
...
+	_, found := exchangeRateFetchers[source]
+	if !found {
+		return errors.New("cannot disable unknown exchange rate source")
+	} else if core.exchangeRateSources[source] == nil {
+		return nil // already disabled.
+	}
+
+	// Remove exchange rate source.
+	delete(core.exchangeRateSources, source)
+
```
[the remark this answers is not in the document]
ukane-philemon: `fetchExchangeRates` loop should shutdown when the next tick fires. But we could run multiple routines when a user do that(toggle btw 0 and 1 multiple times).

> Maybe simplest to just run the one loop regardless and for the life of core.

Quoting @JoeGruffins. 
However, it means if no fiat source was initialized when core starts, we should still start the goroutine? If we don't then users won't be able to toggle fiat source status on the fly. @buck54321 any suggestions?

client/webserver/site/src/html/markets.tmpl
ukane-philemon: > These two figures should be relatively equivalent. We only need to show the approximate fiat value once. How about a new row?

Does this mean we should only display fiat value for `Total`?

client/core/exchangeratefetcher.go
```diff
@@ -0,0 +1,279 @@
...
+
+// fiatRateInfo holds the fiat exchange rate and the last update time for an
+// asset.
+type fiatRateInfo struct {
+	rate       float64
+	lastUpdate time.Time
+}
+
+// rateFetcher can fetch exchange rates for assets from an API.
+type rateFetcher func(context context.Context, logger dex.Logger, assets map[uint32]*SupportedAsset) map[uint32]float64
```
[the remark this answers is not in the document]
ukane-philemon: > You can drop the assets map[uint32]*SupportedAsset argument and just use asset.Assets() internally.

Are we considering fetching fiat rates for assets without wallets?

client/core/exchangeratefetcher.go
```diff
@@ -0,0 +1,279 @@
...
+			expiredCount++
+		}
+	}
+	totalFiatRate := len(source.fiatRates)
+	return expiredCount == totalFiatRate && totalFiatRate != 0
+}
+
+// assetRate returns the exchange rate information for the assetID specified.
+// found is true is the rate source has a value for the assetID specified.
+func (source *commonSource) assetRate(assetID uint32) (fiatRateInfo, bool) {
```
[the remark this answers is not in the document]
ukane-philemon: Maybe even do my check in here and have `assetRate(assetID uint32) float64 {` instead.

ukane-philemon:
@buck54321, I've made changes and updated the UI display in my first comment.

client/core/exchangeratefetcher.go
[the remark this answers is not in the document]
buck54321: Derp. Nevermind me.

buck54321 (changes requested):

client/core/types.go
```diff
@@ -117,6 +117,7 @@ type User struct {
 	Initialized        bool                       `json:"inited"`
 	SeedGenerationTime uint64                     `json:"seedgentime"`
 	Assets             map[uint32]*SupportedAsset `json:"assets"`
+	FiatRates          map[uint32]float64         `json:"fiatRates"`
```
buck54321, line 120: If we can figure out a slightly different encoding here, something like

```Go
FiatRates    map[uint32]*float64
```

and then make a `nil` entry for disabled sources, then we can get rid of the `(*Core).FiatRateSources` method and the changes to `(*WebServer).handleSettings` (all page templates already have access to the `core.User`).

A similar approach might be to provide the rate as part of a struct, and similarly use `nil` to indicate disabled.

```Go
FiatRates map[uint32]*struct{
    Rate float64
}
```
Either way, it's a cleaner interface, I think.

buck54321 (changes requested):
Almost ready.

client/core/types.go
ukane-philemon: Uhm, currently the `map[string]bool` returned by `(*Core).FiatRateSources` method is useful for sending the rate source `name/token` to the frontend `settings page`, where js can manipulate it for use in enabling and disabling rate source.

ukane-philemon:
update send form UI:
usdvform

JoeGruffins (commented):
Currently unable to load the webserver on simnet with a panic:
<details closed>
<summary>panic</summary>

```
2022/07/06 11:11:25 http: panic serving 127.0.0.1:34718: runtime error: slice bounds out of range [-1:]
goroutine 2172 [running]:
net/http.(*conn).serve.func1()
        /usr/lib/go/src/net/http/server.go:1825 +0xbf
panic({0x141c1a0, 0xc001cae2e8})
        /usr/lib/go/src/runtime/panic.go:844 +0x258
github.com/go-chi/chi/v5/middleware.prettyStack.decorateFuncCallLine({}, {0xc000862e5b, 0x1d}, 0x4?, 0x8)
        /home/joe/go/pkg/mod/github.com/go-chi/chi/v5@v5.0.1/middleware/recoverer.go:130 +0x545
github.com/go-chi/chi/v5/middleware.prettyStack.decorateLine({}, {0xc000862e5b?, 0xc00027a800?}, 0x90?, 0x1?)
        /home/joe/go/pkg/mod/github.com/go-chi/chi/v5@v5.0.1/middleware/recoverer.go:106 +0x15d
github.com/go-chi/chi/v5/middleware.prettyStack.parse({}, {0xc001cfa000, 0xaf4, 0xc00222b1d8?}, {0x12fc920, 0x21a0700})
        /home/joe/go/pkg/mod/github.com/go-chi/chi/v5@v5.0.1/middleware/recoverer.go:89 +0x4e5
github.com/go-chi/chi/v5/middleware.PrintPrettyStack({0x12fc920, 0x21a0700})
        /home/joe/go/pkg/mod/github.com/go-chi/chi/v5@v5.0.1/middleware/recoverer.go:46 +0x45
github.com/go-chi/chi/v5/middleware.(*defaultLogEntry).Panic(0x3c?, {0x12fc920?, 0x21a0700?}, {0x8?, 0x0?, 0x0?})
        /home/joe/go/pkg/mod/github.com/go-chi/chi/v5@v5.0.1/middleware/logger.go:165 +0x29
github.com/go-chi/chi/v5/middleware.Recoverer.func1.1()
        /home/joe/go/pkg/mod/github.com/go-chi/chi/v5@v5.0.1/middleware/recoverer.go:28 +0xcc
panic({0x12fc920, 0x21a0700})
        /usr/lib/go/src/runtime/panic.go:838 +0x207
decred.org/dcrdex/client/core.(*Core).saveDisabledRateSources(0xc0003d96c0)
        /home/joe/git/dcrdex/client/core/core.go:7689 +0x19a
decred.org/dcrdex/client/core.(*Core).fiatConversions(0xc0003d96c0)
        /home/joe/git/dcrdex/client/core/core.go:7602 +0x1a5
decred.org/dcrdex/client/core.(*Core).User(0xc0003d96c0)
        /home/joe/git/dcrdex/client/core/core.go:1749 +0x53
decred.org/dcrdex/client/webserver.(*WebServer).authMiddleware.func1({0x7f10c05291c8, 0xc001c991c0}, 0xc00160b100)
        /home/joe/git/dcrdex/client/webserver/middleware.go:43 +0x99
net/http.HandlerFunc.ServeHTTP(0x10?, {0x7f10c05291c8?, 0xc001c991c0?}, 0x4167b9?)
        /usr/lib/go/src/net/http/server.go:2084 +0x2f
github.com/go-chi/chi/v5/middleware.Recoverer.func1({0x7f10c05291c8?, 0xc001c991c0?}, 0xc0016278c0?)
        /home/joe/go/pkg/mod/github.com/go-chi/chi/v5@v5.0.1/middleware/recoverer.go:37 +0x83
net/http.HandlerFunc.ServeHTTP(0x13e86e0?, {0x7f10c05291c8?, 0xc001c991c0?}, 0xe?)
        /usr/lib/go/src/net/http/server.go:2084 +0x2f
decred.org/dcrdex/client/webserver.(*WebServer).securityMiddleware.func1({0x7f10c05291c8, 0xc001c991c0}, 0xc001ce8400?)
        /home/joe/git/dcrdex/client/webserver/middleware.go:34 +0x4ed
net/http.HandlerFunc.ServeHTTP(0xc00160b000?, {0x7f10c05291c8?, 0xc001c991c0?}, 0x203000?)
        /usr/lib/go/src/net/http/server.go:2084 +0x2f
github.com/go-chi/chi/v5/middleware.RequestLogger.func1.1({0x197e468, 0xc001cb0380}, 0xc00160b000)
        /home/joe/go/pkg/mod/github.com/go-chi/chi/v5@v5.0.1/middleware/logger.go:57 +0x3aa
net/http.HandlerFunc.ServeHTTP(0x197ef78?, {0x197e468?, 0xc001cb0380?}, 0x21a0410?)
        /usr/lib/go/src/net/http/server.go:2084 +0x2f
github.com/go-chi/chi/v5.(*Mux).ServeHTTP(0xc0001a8780, {0x197e468, 0xc001cb0380}, 0xc00160af00)
        /home/joe/go/pkg/mod/github.com/go-chi/chi/v5@v5.0.1/mux.go:88 +0x442
net/http.serverHandler.ServeHTTP({0xc001ce8360?}, {0x197e468, 0xc001cb0380}, 0xc00160af00)
        /usr/lib/go/src/net/http/server.go:2916 +0x43b
net/http.(*conn).serve(0xc001cbe140, {0x197f020, 0xc000140a50})
        /usr/lib/go/src/net/http/server.go:1966 +0x5d7
created by net/http.(*Server).Serve
        /usr/lib/go/src/net/http/server.go:3071 +0x4db

```
</details>

And after that I am unable to stop dexc.

JoeGruffins (approved):
Working well on testnet for me.

chappjc (changes requested):
Test well and looks right.  A few initial comments, but full review still coming.

client/webserver/site/src/html/settings.tmpl
```diff
@@ -16,6 +16,18 @@
         [[[Show pop-up notifications]]]
       </label>
     </div>
+    <div {{if not .UserInfo.Authed}} class="d-hide"{{end}}>
+    <span>Fiat Currency: </span><span id="fiatCurrency">{{.FiatCurrency}}</span>
+    </div>
+    <div id="fiatRateSources" {{if not .UserInfo.Authed}} class="d-hide"{{end}}>
+    <span class="mb-2">[[[fiat_exchange_rate_sources]]]:</span>
+    {{range $source, $enabled := .FiatRateSources}}
```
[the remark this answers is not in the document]
ukane-philemon: > Fallback? Do they all get requested and only one used?

They all get requested and an average fiat rate is used as the final rate for assets, if any expires/stops to return rates, others are used as a fallback. Do users need to read it somewhere? 

> Which does it use for, say BTC-USD?

Screenshot 2022-07-09 at 1 09 08 AM

`Fiat Currency` shows the fiat rate currency `USD` e.g(BTC-USD, ETH-USD, DCR-USD), any thoughts on the display?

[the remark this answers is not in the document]
chappjc: > They all get requested and an average fiat rate is used as the final rate for assets, if any expires/stops to return rates, others are used as a fallback if enabled. Do users need to read it somewhere?

They need some kind of explanation that rates are averaged if on multiple sources.  That's not obvious.  I can see an "i" in circle with a tooltip, as short as possible.

> Fiat Currency shows the fiat rate currency USD e.g(BTC-USD, ETH-USD, DCR-USD), any thoughts on the display?

My question "Which does it use for, say BTC-USD?" was meant to show a user's thought process.  Say I'm trading something against BTC (or just withdrawing BTC), which source has that exchange rate?  Do they all have the same asset rates available or do some have assets that others do not?  If the idea is that they will all have the same info, just different values, OK, but that's not a given.

e.g. Does dcrdata really provide exchange rates for LTC and BCH?  If I want rates for these, what sources do I pick?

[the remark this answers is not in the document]
chappjc: I can live with this as-is if the general guidance is something like: "Sources may provide fiat exchange rates for different subsets of assets.  You should select all sources that are acceptable to get the fiat exchange rates for the most assets.  Assets with fiat rates from multiple sources will use the average from all of those sources."

Is that generally accurate?  We should revise the wording, but I'm trying to understand how it's meant to be interpreted.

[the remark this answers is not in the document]
ukane-philemon: @buck54321 any thoughts on this? I agree users need to know what fiat exchange rates are available per source, and a clear UI interpretation of this information is important.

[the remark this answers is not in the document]
chappjc: We don't have to go down the rabbit hole to Lodi them all now.  I just want to understand.
If my understanding is right, let's just put a little text in a tooltip, otherwise we just end up answering the question in chat and GitHub issues.

client/webserver/site/src/html/settings.tmpl
```diff
@@ -16,6 +16,18 @@
...
+    <div class="form-check">
+      <input class="form-check-input" type="checkbox" value="{{$source}}" id="{{$source}}" {{if $enabled}} checked {{end}}>
+      <label class="form-check-label" for="{{$source}}">{{toTitle $source}}</label>
```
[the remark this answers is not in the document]
ukane-philemon: > "Dcrdata" is weird (it's just "dcrdata"), and we can just have the other sources capitalized as needed in FiatRateSources.

@chappjc, is it ok to have all sources displayed in lowercase (I think it's odd to have one lowercase and the others capitalized)?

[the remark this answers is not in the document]
chappjc: Can't the correct case just be in the map?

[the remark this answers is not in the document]
ukane-philemon: > Can't the correct case just be in the map? 

they can but we end up having only `dcrdata` displayed in lowercase.

ukane-philemon:
We might need to revise the tooltip infomation.
Screenshot 2022-07-11 at 12 58 26 PM

buck54321 (changes requested):
When I enable dcrdata on simnet, navigate to the markets page, and prepare an order, the fiat rate is not shown. I have to refresh. 

I can `enableLogger('notes', true)` and see that the notification is coming through.

client/webserver/site/src/html/settings.tmpl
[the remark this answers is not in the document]
buck54321: > @buck54321 any thoughts on this? I agree users need to know what fiat exchange rates are available per source, and a clear UI interpretation of this information is important.

It's not too bad and it doesn't have to be perfect right away. The new tooltip is good enough for now.

client/core/exchangeratefetcher.go
```diff
@@ -0,0 +1,270 @@
...
+// assetRate returns the fiat rate information for the assetID specified.
+func (source *commonRateSource) assetRate(assetID uint32) *fiatRateInfo {
+	source.mtx.Lock()
+	defer source.mtx.Unlock()
+	rateInfo := source.fiatRates[assetID]
+	var fiatRateInfo fiatRateInfo
+	if rateInfo != nil {
+		fiatRateInfo = *rateInfo
+	}
+	return &fiatRateInfo
```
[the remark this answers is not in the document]
ukane-philemon: I think this will no longer be protected by the lock if the caller gets the pointer to the struct in the map.

> it won't be modified, will it?

Maybe.

ukane-philemon:
> I have to refresh.

maybe the fiat rate was not available at the time you prepared the order. A refresh requests the rates directly from the client-server via `fetchUser()`

client/core/core_test.go
```diff
@@ -9060,3 +9073,107 @@ func TestLCM(t *testing.T) {
...
+
+	// Expects assets fiat rate values.
+	fiatRates = tCore.fiatConversions()
+	if len(fiatRates) != 2 {
+		t.Fatal("Expected assets fiat rate for two assets")
+	}
+
+	// fiat rates for assets can expire, and fiat rate fetchers can be
+	// removed if expired.
+	time.Sleep(2 * time.Second)
```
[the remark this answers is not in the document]
ukane-philemon: > Can you just manually set the lastRequest?

`source.isExpired(expiry duration)` does not check expiry using `source.lastRequest`. A source is considered expired when the source's fiat rates are all obsolete.  Each `fiatRateInfo` has a `lastUpdate` used for checking if the `fiatRateInfo` is expired. will manually set the `fiatRateInfo.lastUpdate`.

client/core/exchangeratefetcher.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+// refreshRates updates the last update time and the rate information for assets.
+func (source *commonRateSource) refreshRates(ctx context.Context, assets map[uint32]*SupportedAsset) {
+	source.logRequest()
+	fiatRates := source.fetchRates(ctx, source.log, assets)
```
chappjc, line 109: In general, I strongly dislike giving level structs like `commonRateSource` a log field.  If it's inconvenient to pass errors to the caller, we can just as easily pass the logger with `ctx` when calling `refreshRates` (just like the individual `rateFetcher`s).

client/core/core.go
```diff
@@ -7446,3 +7488,218 @@ func (c *Core) findActiveOrder(oid order.OrderID) (*trackedTrade, error) {
...
+	if len(fiatRatesMap) != 0 {
+		core.notify(newFiatRatesUpdate(fiatRatesMap))
+	}
+}
+
+// nextRateSourceTick checks the fiat rate source's last request, and calculates
+// when the next cycle should run.
+func (core *Core) nextRateSourceTick() *time.Timer {
+	core.ratesMtx.Lock()
+	defer core.ratesMtx.Unlock()
```
chappjc, line 7549: `core.fiatRateSources` will be used _far_ more frequently for read-only purposes, so we'd benefit from a `sync.RWMutex` instead.

client/core/exchangeratefetcher.go
```diff
@@ -0,0 +1,270 @@
...
+type fiatRateInfo struct {
+	rate       float64
+	lastUpdate time.Time
+}
+
+// rateFetcher can fetch fiat rates for assets from an API.
+type rateFetcher func(context context.Context, logger dex.Logger, assets map[uint32]*SupportedAsset) map[uint32]float64
+
+type commonRateSource struct {
+	mtx         sync.RWMutex
```
chappjc, line 58: Should be clear from the struct definition what fields the mutex guards.  Please reformat so that the mutex is above all fields it protects and others are separate, as we do in other parts of the code.

client/core/exchangeratefetcher.go
```diff
@@ -0,0 +1,270 @@
...
+	source.mtx.RLock()
+	defer source.mtx.RUnlock()
+	return source.lastRequest
+}
+
+// isExpired checks the last update time for all fiat rates against the
+// provided expiryTime.
+func (source *commonRateSource) isExpired(expiryTime time.Duration) bool {
+	source.mtx.Lock()
+	defer source.mtx.Unlock()
```
chappjc, line 83: RLock?

client/core/exchangeratefetcher.go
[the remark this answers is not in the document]
chappjc: > I think this will no longer be protected by the lock if the caller gets the pointer to the struct in the map.

What does the mutex aim to protect?  The map, the elements in the map, or both?  This is an extremely important distinction.  It should be documented clearly.  Being unclear about this is a very common source of bugs.

> Maybe.

Be clear and specific please.  It should be known and documented.  I see the map entry being replcaed entirely in `refreshRates`, not modified.

client/core/exchangeratefetcher.go
```diff
@@ -0,0 +1,270 @@
...
+	var fiatRateInfo fiatRateInfo
+	if rateInfo != nil {
+		fiatRateInfo = *rateInfo
+	}
+	return &fiatRateInfo
+}
+
+// refreshRates updates the last update time and the rate information for assets.
+func (source *commonRateSource) refreshRates(ctx context.Context, assets map[uint32]*SupportedAsset) {
+	source.logRequest()
```
chappjc, line 108: This is the one and only place `logRequest` is called, and you lock the mutex again one line down, so why not remove the `logRequest` method and just do `source.lastRequest = time.Now()` before the `for` loop.

Also, don't call `time.Now` in a loop, especially with a mutex locked.  In general, accessing system clock should be considered an "expensive" operation compared with, say, arithmetic.  Do `time.Now` once, prior to obtaining the lock, and assign it wherever.

chappjc (changes requested):

client/core/exchangeratefetcher.go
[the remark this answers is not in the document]
JoeGruffins: I think this was my suggestion. The returned value could be altered changing the value in the map if we just return the pointer. Currently the value is only read so this is not necessary... now. I think either doing it this way or leaving a note in the comments that the value should not be altered is good to prevent possible bugs in the future.

Apologies if this is overthinking it.

client/core/core.go
```diff
@@ -7446,3 +7488,218 @@ func (c *Core) findActiveOrder(oid order.OrderID) (*trackedTrade, error) {
...
+	defer cancel()
+
+	var wg sync.WaitGroup
+	core.ratesMtx.Lock()
+	for _, source := range core.fiatRateSources {
+		if time.Since(source.lastTry()) > fiatRateRequestInterval {
+			wg.Add(1)
+			go func(source *commonRateSource) {
+				defer wg.Done()
+				source.refreshRates(ctx, core.assetMap())
```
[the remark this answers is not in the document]
ukane-philemon: > I'm also not convinced that Core.fiatRateSources needs to be mutable at all. I don't see why it cannot be filled to the size of fiatRateFetchers on construction of the Core instance, and the commonRateSource be given an enabled flag.

Sounds good but we don't want to keep disabled/unused sources in `core.fiatRateSources` if they won't be needed?

client/core/exchangeratefetcher.go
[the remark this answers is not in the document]
ukane-philemon: > What does the mutex aim to protect? 

The map, in `refreshRates` the`fiatRatesInfo` for an asset  in `source.fiatRates` is replaced if a new rate was returned.

client/core/core.go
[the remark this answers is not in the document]
chappjc: Yeah maybe not.  That last idea was just a feeling.  I hadn't explorered it.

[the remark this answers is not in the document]
buck54321: I had a smell here too.

https://github.com/decred/dcrdex/pull/1600#discussion_r906711551

But my read was a little off. Probably worth some optimization, but that can also happen later.

client/core/exchangeratefetcher.go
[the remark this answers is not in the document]
buck54321: We don't need the copy. You would only need a copy if you expected the `fiatRateInfo` fields to be modified, but once created, the `fiatRateInfo` fields are read-only. You're only protecting the map. IMO, we don't even need to document that. It's sort of the default. But if other's want a doc, you should.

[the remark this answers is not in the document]
buck54321: Sorry @chappjc. I wasn't calling out your request for docs as unusual or off-base. On the contrary, I think as soon as there appears to be some confusion amongst developers around any topic, more docs are righteous.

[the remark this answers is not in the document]
ukane-philemon: made changes to `source.assetRate` and dropeda comment.

chappjc:
Feel free to squash before rebasing to make resolution simpler. i.e. `git rebase -i $(git merge-base master HEAD)` [squash via fixup commits], then `git rebase master` [resolve one commit instead of 3]

client/core/core.go
```diff
@@ -7437,3 +7479,223 @@ func (c *Core) findActiveOrder(oid order.OrderID) (*trackedTrade, error) {
...
+	// else fetch rates.
+	if len(c.fiatRateSources) == 1 {
+		c.fetchFiatExchangeRates()
+	} else {
+		source := c.fiatRateSources[source]
+		supportedAssets := c.SupportedAssets()
+		go func(source *commonRateSource) {
+			ctx, cancel := context.WithTimeout(c.ctx, 4*time.Second)
+			defer cancel()
+			source.refreshRates(ctx, c.log, supportedAssets)
```
chappjc, line 7645: On this path, it looks like there won't be a fiatRateUpdate note until the next regular tick.  Intended?

client/core/core.go
```diff
@@ -7437,3 +7479,223 @@ func (c *Core) findActiveOrder(oid order.OrderID) (*trackedTrade, error) {
...
+			}
+		}
+		if rateSum != 0 {
+			fiatRatesMap[assetID] = rateSum / float64(sources) // get average rate.
+		}
+	}
+
+	// Ensure disabled fiat rate fetchers are saved to database.
+	if disabledCount > 0 {
+		c.saveDisabledRateSources()
```
chappjc, line 7604: This should have a WRN log I think.  It means something is failing, although hopefully remotely, and the user's source selections were changed for them.

chappjc: Philosophically, pruning dead rate sources is a job only for `refreshFiatRates` (after it completes `wg.Wait`).
`fiatConversions()` is primarily a getter method, called from the `(*Core).User` method that is hit quite frequently by external consumers and should generally by fast and not modify state.

client/core/core_test.go
```diff
@@ -1072,6 +1076,13 @@ func randomMsgMarket() (baseAsset, quoteAsset *msgjson.Asset) {
 	return randomAsset(), randomAsset()
 }
 
+func tFetcher(_ context.Context, log dex.Logger, _ map[uint32]*SupportedAsset) map[uint32]float64 {
+	fiatRates := make(map[uint32]float64)
+	fiatRates[tUTXOAssetA.ID] = 45
+	fiatRates[tUTXOAssetB.ID] = 32000
+	return fiatRates
```
chappjc, line 1083: Map literal would be best here:

```go
	return map[uint32]float64{
		tUTXOAssetA.ID: 45,
		tUTXOAssetB.ID: 32000,
	}
```

client/core/exchangeratefetcher.go
```diff
@@ -0,0 +1,254 @@
...
+	defer source.mtx.RUnlock()
+	var expiredCount int
+	for _, rateInfo := range source.fiatRates {
+		if time.Since(rateInfo.lastUpdate) > expiryTime {
+			expiredCount++
+		}
+	}
+	totalFiatRate := len(source.fiatRates)
+	return expiredCount == totalFiatRate && totalFiatRate != 0
+}
```
chappjc, line 79: Embrace early returns for readability and simplicity:

```diff
 // isExpired checks the last update time for all fiat rates against the
-// provided expiryTime.
+// provided expiryTime. This only returns true if all rates are expired.
 func (source *commonRateSource) isExpired(expiryTime time.Duration) bool {
+       now := time.Now()
        source.mtx.RLock()
        defer source.mtx.RUnlock()
-       var expiredCount int
+       if len(source.fiatRates) == 0 {
+               return false
+       }
        for _, rateInfo := range source.fiatRates {
-               if time.Since(rateInfo.lastUpdate) > expiryTime {
-                       expiredCount++
+               if now.Sub(rateInfo.lastUpdate) < expiryTime {
+                       return false // one not expired is enough
                }
        }
-       totalFiatRate := len(source.fiatRates)
-       return expiredCount == totalFiatRate && totalFiatRate != 0
+       return true
 }
```

client/core/exchangeratefetcher.go
```diff
@@ -0,0 +1,254 @@
...
+
+// refreshRates updates the last update time and the rate information for assets.
+func (source *commonRateSource) refreshRates(ctx context.Context, logger dex.Logger, assets map[uint32]*SupportedAsset) {
+	fiatRates := source.fetchRates(ctx, logger, assets)
+	source.mtx.Lock()
+	defer source.mtx.Unlock()
+	for assetID, fiatRate := range fiatRates {
+		source.fiatRates[assetID] = &fiatRateInfo{
+			rate:       fiatRate,
+			lastUpdate: time.Now(),
```
chappjc, line 97:
```diff
 func (source *commonRateSource) refreshRates(ctx context.Context, logger dex.Logger, assets map[uint32]*SupportedAsset) {
        fiatRates := source.fetchRates(ctx, logger, assets)
+       now := time.Now()
        source.mtx.Lock()
        defer source.mtx.Unlock()
        for assetID, fiatRate := range fiatRates {
                source.fiatRates[assetID] = &fiatRateInfo{
                        rate:       fiatRate,
-                       lastUpdate: time.Now(),
+                       lastUpdate: now,
```

client/core/core.go
```diff
@@ -7437,3 +7479,223 @@ func (c *Core) findActiveOrder(oid order.OrderID) (*trackedTrade, error) {
...
+	c.ratesMtx.Lock()
+	defer c.ratesMtx.Unlock()
+
+	// Check if it's an invalid fiat rate source or it is already
+	// disabled.
+	_, found := fiatRateFetchers[source]
+	if !found {
+		return errors.New("cannot disable unknown fiat rate source")
+	} else if c.fiatRateSources[source] == nil {
+		return nil // already disabled.
```
chappjc, line 7667: Always be judicious with what needs to go in critical sections (mutex locked):

```diff
 func (c *Core) disableRateSource(source string) error {
+       // Check if it's an invalid fiat rate source or it is already disabled.
+       if _, found := fiatRateFetchers[source]; !found {
+               return errors.New("cannot disable unknown fiat rate source")
+       }
+
        c.ratesMtx.Lock()
        defer c.ratesMtx.Unlock()
-
-       // Check if it's an invalid fiat rate source or it is already
-       // disabled.
-       _, found := fiatRateFetchers[source]
-       if !found {
-               return errors.New("cannot disable unknown fiat rate source")
-       } else if c.fiatRateSources[source] == nil {
+       if c.fiatRateSources[source] == nil {
                return nil // already disabled.
```

chappjc (changes requested):

client/core/core.go
ukane-philemon: > On this path, it looks like there won't be a fiatRateUpdate note until the next regular tick. Intended?

Yes, fetching rates for the newly enabled source will ensure the next tick includes the source rates when computing the average fiat rate for an asset.

client/webserver/site/src/html/wallets.tmpl
```diff
@@ -71,9 +71,13 @@
             <span class="fs20">{{.Info.Name}}</span>
             <span class="fs15">({{toUpper .Symbol}})</span>
           </td>
-          <td data-balance-target="{{.ID}}">
+         <td>
             {{if .Wallet}}
+            <div data-balance-target="{{.ID}}">
               {{.Info.UnitInfo.ConventionalString .Wallet.Balance.Available}}
+            </div>
+             <div class="d-hide d-flex"><span data-conversion-target={{.ID}}></span>USD</span>
```
chappjc, line 79: Stray space here.

client/webserver/site/src/js/app.ts
```diff
@@ -566,15 +572,16 @@ export default class Application {
...
         const asset = this.assets[wallet.assetID]
         asset.wallet = wallet
         this.walletMap[wallet.assetID] = wallet
+        const bal = wallet.balance.available
         const balances = this.main.querySelectorAll(`[data-balance-target="${wallet.assetID}"]`)
-        balances.forEach(el => { el.textContent = Doc.formatFullPrecision(wallet.balance.available, asset.info.unitinfo) })
-        break
+        balances.forEach(el => { el.textContent = Doc.formatFullPrecision(bal, asset.info.unitinfo) })
       }
+        break
```
chappjc, line 584: This move outside the block is odd compared to the other cases.

client/webserver/site/src/js/app.ts
```diff
@@ -566,15 +572,16 @@ export default class Application {
       case 'walletstate':
       case 'walletconfig': {
         // assets can be null if failed to connect to dex server.
-        if (!this.assets) return
+        if (!this.assets) break
```
chappjc, line 575: I'm not sure this was a bug to fix.  I expect that the walletconfig note is unhandleable if `this.assets` is null.

As an aside, I don't know that the comment or this condition are even possible.  `User.Assets` can only be undefined if fetchUser has never been called, and I don't see that.  @buck54321 do you remember this?  https://github.com/decred/dcrdex/pull/989#discussion_r602382449

chappjc (approved):
Working well.  Good stuff!  A few minor nits remaining.  Will wait on @buck54321 for a closer look at frontend bits.

client/core/core.go
```diff
@@ -7723,3 +7765,214 @@ func (c *Core) findActiveOrder(oid order.OrderID) (*trackedTrade, error) {
...
+	}
+
+	// Build fiat rate source.
+	rateSource := newCommonRateSource(rateFetcher)
+	c.fiatRateSources[source] = rateSource
+
+	// If this is our first fiat rate source, start fiat rate fetcher goroutine,
+	// else fetch rates.
+	if len(c.fiatRateSources) == 1 {
+		c.fetchFiatExchangeRates()
```
buck54321, line 7893: How about a `c.log.Debug("starting fiat rate fetching")` here

client/core/core.go
```diff
@@ -7723,3 +7765,214 @@ func (c *Core) findActiveOrder(oid order.OrderID) (*trackedTrade, error) {
...
+	var disabled []string
+	for token := range fiatRateFetchers {
+		if c.fiatRateSources[token] == nil {
+			disabled = append(disabled, token)
+		}
+	}
+
+	// Shutdown rate fetching if there are no exchange rate source.
+	if len(c.fiatRateSources) == 0 && c.stopFiatRateFetching != nil {
+		c.stopFiatRateFetching()
```
buck54321, line 7970: Maybe `c.log.Debug("shutting down rate fetching")`

buck54321 (changes requested):
These should be my last requests.

client/webserver/site/src/js/app.ts
buck54321: It's not clear to me what's gained with this change either. Is there a specific issue you were addressing, @ukane-philemon?

I'm sure this line was written with blood, but if you think it's worth pursuing, we'd need to audit the `walletstate`and `walletconfig` notifications to ensure that they can't arrive async after we connect but before we receive our `fetchUser` result.

ukane-philemon: > Is there a specific issue you were addressing, @ukane-philemon?

No.

chappjc: OK no sweat.  Let's just keep this part the same and I think we can get this PR merged.

ukane-philemon:
@buck54321, @chappjc, @JoeGruffins for thanks for the reviews and making this look shiny :)

PR #1602, opened 2022-04-30
client: support server TLS certificate change

### LINKED ISSUE #1571: client: support server TLS certificate change

martonp:
- `client/comms`: Instead of just keeping track of the connection
  being conneccted or not, a `ConnectionStatus` type is introduced
  which includes `InvalidCert` in addition to `Connected` and
  `Disconnected`. This is then passed to core and the UI through
  the `ConnEventNote`.

- `client/core`: A new function `UpdateCert` is added which uses
  TLS certificate supplied by the user to connect to a server. If
  the connection is successful, the new certificate is stored in
  the db.

- `client/webserver`: A new api request `/api/updatecert` is added
  to allow the user to update the TLS Certificate for a host.

- `ui`: The settings page is updated. Instead of all the options for
  each server being on the settings page, there is just a link to a
  dex settings page specific to a server. This contains all the options
  that previously existed on the settings page in addition to the new
  option to update the TLS Certificate.

Closes #1571


buck54321 (changes requested):
We can probably downsize the connection status and icon a bit



After the certificate is accepted, there's really no indication that things went correctly. I wonder if we could just display a little message or checkmark next to the button or something.

client/webserver/http.go
```diff
@@ -244,6 +244,34 @@ func (s *WebServer) handleSettings(w http.ResponseWriter, r *http.Request) {
...
+func (s *WebServer) handleDexSettings(w http.ResponseWriter, r *http.Request) {
+	host, err := getHostCtx(r)
+	if err != nil {
+		log.Errorf("error getting host ctx: %v", err)
+		http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
+		return
+	}
+
+	exchanges := s.core.Exchanges()
+	exchange, ok := exchanges[host]
```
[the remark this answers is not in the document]
martonp: A new method sounds good, but what do you mean by extra encoding?

client/comms/wsconn.go
```diff
@@ -195,11 +204,13 @@ func (conn *wsConn) connect(ctx context.Context) error {
 	if err != nil {
 		var e x509.UnknownAuthorityError
 		if errors.As(err, &e) {
+			conn.setConnectionStatus(InvalidCert)
 			if conn.tlsCfg == nil {
 				return ErrCertRequired
 			}
 			return ErrInvalidCert
 		}
+		conn.setConnectionStatus(Disconnected)
```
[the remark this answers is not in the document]
martonp: What if the server goes offline, updates its certificate, and turns back on? The user will not know that the reason the client isn't able to connect is that the TLS certificate is incorrect, and not that the server is still offline.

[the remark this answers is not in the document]
buck54321: Oh, right. Because this could happen during `keepAlive` too. Thanks for clarifying.

client/webserver/http.go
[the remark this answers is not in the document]
buck54321: The `dexConnection` → `Exchange` conversions.

client/comms/wsconn.go
```diff
@@ -165,18 +174,18 @@ func NewWsConn(cfg *WsCfg) (WsConn, error) {
...
-// setConnected updates the connection's connected state and runs the
+// setConnectionStatus updates the connection's status and runs the
 // ConnectEventFunc in case of a change.
-func (conn *wsConn) setConnected(connected bool) {
+func (conn *wsConn) setConnectionStatus(status ConnectionStatus) {
 	conn.connectedMtx.Lock()
-	statusChange := conn.connected != connected
-	conn.connected = connected
+	statusChange := conn.connectionStatus != status
+	conn.connectionStatus = status
```
JoeGruffins, line 185: It looks like you could change this to an atomic an remove the lock if you wanted.

JoeGruffins (commented):
Working well, but, if you change the cert while dexc is offline, it will only say `Disconnected` after you start up dexc and navigate to the settings screen, even if the cert is the problem.

client/comms/wsconn.go
martonp: Obviously it's very unlikely, but wouldn't it be possible to call `ConnectEventFunc` twice for the same status update, if two goroutines are updating to the same status it the same time?

martonp:
While fixing the issue of the dex settings screen showing `Disconnected` if the cert is updated while the client is offline, I also noticed that the client will not do keep alive if the initial connection failed. If the initial connection fails, the `ConnectionMaster` will cancel the context, causing the keep alive loop to end. This is a bug isn't it?

chappjc:
> While fixing the issue of the dex settings screen showing `Disconnected` if the cert is updated while the client is offline, I also noticed that the client will not do keep alive if the initial connection failed. If the initial connection fails, the `ConnectionMaster` will cancel the context, causing the keep alive loop to end. This is a bug isn't it?

Haven't tried this pr yet, but the reconnect loop starts if the server is unreachable on dexc startup. https://github.com/decred/dcrdex/commit/c782ffb0a8e4b8dd11360c8a6394a6b64c91cd3e

I would expect it to do the same on server reconfigure, but I have to look into that

martonp:
The loop starts, but it is stopped immediately because in `ConnectionMaster.Connect`, if `c.connector.Connect` returns an error, then the context is cancelled.

chappjc:
OK, I'll look at it.  On master and 0.4, the reconnect loops until connection succeeds.  I can start dexc with network down and it starts connection attempts, backing off, and when network comes up it connects fine.

martonp:
In this commit here: https://github.com/decred/dcrdex/commit/f222e12769e9f7c4d2da39b5c03dd66296b45290

Do you not see the "no longer keeping alive" message right after it fails to connect? For me, using master, it does not reconnect to the dex if it fails to connect at startup.

chappjc:
Oh WTF.  It's broken on master.
I'll hunt down what broke it

chappjc:
Hmm https://github.com/decred/dcrdex/pull/1567

chappjc:
Oh, it was this: https://github.com/decred/dcrdex/pull/1472/commits/e8fd8acab23fbb253b649a66a95a5a63d4be6b75

I'll try to figure out what needs to be done there.

chappjc:
Let's consider this: https://github.com/decred/dcrdex/pull/1616
See if you can get this PR working with that.

martonp:
It works well after rebasing. I've gotten rid of the last unnecessary commit.

JoeGruffins (approved):
I guess we don't need to do the reconnect loop if we failed for the certificate. We would expect it to continue failing until the certificate is changed. I don't know if this is a big change though. Working well.

client/core/core.go
```diff
@@ -328,16 +330,18 @@ func (dc *dexConnection) exchangeInfo() *Exchange {
 		feeAssets["dcr"] = dcrAsset
 	}
 
+	connectionStatus = comms.ConnectionStatus(
+		atomic.LoadUint32(&dc.connectionStatus))
```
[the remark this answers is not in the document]
JoeGruffins: I used to do this with a combination of vps and ssh while on the train. Not recently though.

client/core/account.go
```diff
@@ -190,3 +191,26 @@ func (c *Core) AccountImport(pw []byte, acct Account) error {
...
+// the connection is successful, then the cert in the database is updated.
+func (c *Core) UpdateCert(host string, cert []byte) error {
+	accountInfo, err := c.db.Account(host)
+	if err != nil {
+		return err
+	}
+
+	accountInfo.Cert = cert
+
+	_, connected := c.connectAccount(accountInfo)
```
chappjc, line 205: If there's a dexConnection already in the map, it will be orphaned.  You can observe this if a failing reconnect loop is going for the old connection as it never stops even though the conn was replaced with a working one with a new TLS cert.

I believe `connectAccount` needs a check that stops any existing connection (prior to connecting the new one otherwise the frontend gets the notifications in the wrong order and it looks disconnected until you refresh the page):

```diff
diff --git a/client/core/core.go b/client/core/core.go
index 48b1002f..4dfcf333 100644
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -4824,7 +4824,8 @@ func (c *Core) initialize() {
 // connectAccount makes a connection to the DEX for the given account. If a
 // non-nil dexConnection is returned, it was inserted into the conns map even if
 // the initial connection attempt failed (connected == false), and the connect
-// retry / keepalive loop is active.
+// retry / keepalive loop is active. If there was already a dexConnection, it is
+// first stopped.
 func (c *Core) connectAccount(acct *db.AccountInfo) (dc *dexConnection, connected bool) {
        if !acct.Paid && len(acct.FeeCoin) == 0 {
                // Register should have set this when creating the account that was
@@ -4839,6 +4840,14 @@ func (c *Core) connectAccount(acct *db.AccountInfo) (dc *dexConnection, connecte
                c.log.Errorf("skipping loading of %s due to address parse error: %v", host, err)
                return
        }
+
+       c.connMtx.RLock()
+       if dc := c.conns[host]; dc != nil {
+               dc.connMaster.Disconnect()
+               dc.acct.lock()
+       } // leave it in the map so it remains listed if connectDEX fails
+       c.connMtx.RUnlock()
+
        dc, err = c.connectDEX(acct)
        if dc == nil {
                c.log.Errorf("Cannot connect to DEX %s: %v", host, err)
```

chappjc (changes requested):
One more fix.  Otherwise working well:

image

image

client/core/account.go
chappjc: Should close the book feeds to I think

```go
	c.connMtx.RLock()
	if dc := c.conns[host]; dc != nil {
		dc.connMaster.Disconnect()
		dc.acct.lock()
		dc.booksMtx.Lock()
		for _, b := range dc.books {
			b.closeFeeds()
			if b.closeTimer != nil {
				b.closeTimer.Stop()
			}
		}
		dc.booksMtx.Unlock()
	} // leave it in the map so it remains listed if connectDEX fails
	c.connMtx.RUnlock()
```

chappjc: https://github.com/decred/dcrdex/commit/ed9284e99e17aef82f8af4d3ef7a5d0527bad5ed

client/core/core.go
```diff
@@ -1395,7 +1400,22 @@ func (c *Core) Network() dex.Network {
...
+	for _, dc := range dcs {
+		infos[dc.acct.host] = dc.exchangeInfo()
+	}
+	return infos
+}
+
+// Exchange returns an exchange with a certain host. It returns an error if
+// no exchange exists at that host.
+func (c *Core) Exchange(host string) (*Exchange, error) {
+	exchange, ok := c.Exchanges()[host]
```
buck54321, line 1414: The point of the new method is to avoid all the extra calls to `exchangeInfo` for hosts we don't care about.


```Go
func (c *Core) Exchange(host string) (*Exchange, error) {
	dc, _, err := c.dex(host)
	if err != nil {
		return nil, err
	}
	return dc.exchangeInfo(), nil
}
```

client/webserver/site/src/font/icomoon.svg
```diff
@@ -19,6 +19,7 @@
...
+<glyph unicode="&#xe91b;" glyph-name="connection" horiz-adv-x="1280" d="M640 256c105.87 0 201.87-43.066 271.402-112.598l-90.468-90.468c-46.354 46.356-110.356 75.066-180.934 75.066s-134.578-28.71-180.934-75.066l-90.468 90.468c69.532 69.532 165.532 112.598 271.402 112.598zM187.452 324.548c120.88 120.88 281.598 187.452 452.548 187.452s331.668-66.572 452.55-187.452l-90.51-90.508c-96.706 96.704-225.28 149.96-362.040 149.96-136.762 0-265.334-53.256-362.038-149.962l-90.51 90.51zM988.784 697.562c106.702-45.132 202.516-109.728 284.782-191.996v0l-90.508-90.508c-145.056 145.056-337.92 224.942-543.058 224.942-205.14 0-398-79.886-543.058-224.942l-90.51 90.51c82.268 82.266 178.082 146.862 284.784 191.994 110.504 46.738 227.852 70.438 348.784 70.438s238.278-23.7 348.784-70.438zM576-64c0 35.346 28.654 64 64 64s64-28.654 64-64c0-35.346-28.654-64-64-64s-64 28.654-64 64z" />
```
buck54321, line 22: Please put it at the bottom and change the value to `e90e` to stay on pattern.

client/webserver/site/src/js/app.ts
```diff
@@ -578,7 +580,9 @@ export default class Application {
       case 'conn': {
         const n = note as ConnEventNote
         const xc = this.user.exchanges[n.host]
-        if (xc) xc.connected = n.connected
+        if (xc) {
+          xc.connectionStatus = n.connectionStatus
+        }
```
buck54321, line 585: Can still be one line.

client/webserver/site/src/html/dexsettings.tmpl
```diff
@@ -0,0 +1,41 @@
...
+{{template "top" .}}
+{{$passwordIsCached := .UserInfo.PasswordIsCached}}
+<div id="main" data-handler="dexsettings" data-host="{{.Exchange.Host}}" class="text-center py-5 overflow-y-auto">
+  <span class="settings-gear ico-settings"></span>
+  <h2 class="text-break">{{.Exchange.Host}}</h2>
+  <div class="d-flex justify-content-center">
+    <span class="fs20 mx-2 connection ico-connection d-hide" id="connectedIcon"></span>
+    <span class="fs20 mx-2 disconnected ico-disconnected d-hide" id="disconnectedIcon"></span>
+    <h4 id="connectionStatus"></h4>
+  </div>
```
buck54321, line 11: Need an `align-items-center` on the connection status div. Might as well just use `flex-center`. Proper alignment is also hindered by the `margin-bottom: 0.5rem;` that bootstrap adds to `h4`.

image

The font sizes used are pretty large. Without a compelling reason to go larger, `.fs28` is the largest font-size I'd like to use throughout.

```html
 <div class="flex-center fs28">{{.Exchange.Host}}</div>
 <div class="flex-center fs16 mb-2">
    <span class="me-2 connection ico-connection d-hide" id="connectedIcon"></span>
    <span class="disconnected ico-disconnected d-hide" id="disconnectedIcon"></span>
    <span id="connectionStatus"></span>
 </div>
```
image

buck54321 (changes requested):

client/core/account.go
martonp: Nice, the changes work well.

client/core/core.go
martonp: Haha my bad.

client/core/types.go
```diff
@@ -498,15 +499,15 @@ type PendingFeeState struct {
...
-	Connected  bool                  `json:"connected"`
-	Fee        *FeeAsset             `json:"feeAsset"` // DEPRECATED. DCR.
-	RegFees    map[string]*FeeAsset  `json:"regFees"`
-	PendingFee *PendingFeeState      `json:"pendingFee,omitempty"`
-	CandleDurs []string              `json:"candleDurs"`
+	Host             string                 `json:"host"`
+	AcctID           string                 `json:"acctID"`
+	Markets          map[string]*Market     `json:"markets"`
+	Assets           map[uint32]*dex.Asset  `json:"assets"`
+	ConnectionStatus comms.ConnectionStatus `json:"connectionStatus"`
```
chappjc, line 506: Ok because we are still at v0.x but note that when we reach v1, we're making a promise to Go consumers not to break the API.  So doing something like this in the future would cause the module to become `core/v2`. 

However, if we left `Connected bool`, it would not be a breaking change even if it appears redundant with `ConnectionStatus`.

Just pointing it out so we can get used to stabilizing the API.

chappjc:
All still looking good to me with the latest update. LOL re the `Exchange` method.  That one went right over my head.

buck54321 (approved):

chappjc (approved):

PR #1603, opened 2022-05-03
client/testing: Make simnet tests modular.

### LINKED ISSUE #1565: testing: core harness and loadbot tests need to be generalized
If we host the implementation and claim support for the asset, it needs to go through the full range of testing. **trade_simnet_test.go** must be generalized for arbitrary asset pairs. Loadbot too.

JoeGruffins:
client/testing: Make simnet tests modular.
    
    Change the simnet harness tests to work with all currently supported
    assets.

    testing/doge: Add delta and gamma nodes.
    
    Because doge does not distinguish between different wallets, it is
    difficult to test balance changes when testing, especially if the node
    is mining. Add two wallets that don't do anything special to use in
    tests.

part of #1565

edit: Not sure if it closes the issue because it also mentions loadbot. Removed the closing phrase.

### DIFF
--- a/client/core/trade_simnet_test.go
+++ b/client/core/trade_simnet_test.go
@@ -1,4 +1,4 @@
-//go:build harness
+//go:build harness && lgpl
 
 package core
 
@@ -24,11 +24,12 @@ package core
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"math"
 	"os"
 	"os/exec"
-	"os/user"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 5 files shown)


JoeGruffins:
Tests for all assets are passing. eth sometimes fails, but I expect it's because transactions are not mined sometimes. Added a block multiplier to try and solve this.

Two wallets added to the doge harness in order to test balance changes.

client/core/trade_simnet_test.go
```diff
@@ -1293,34 +1348,207 @@ func tryUntil(ctx context.Context, tryDuration time.Duration, tryFn func() bool)
...
+}
+
+func (hc *harnessCtrl) tmuxRun(cmd string) error {
+	cmd += "; tmux wait-for -S harnessdone"
+	err := exec.Command("tmux", "send-keys", "-t", hc.tmuxWindow, cmd, "C-m").Run() // ; wait-for harnessdone
+	if err != nil {
+		return nil
+	}
+	return exec.Command("tmux", "wait-for", "harnessdone").Run()
+}
```
buck54321, line 1377: Instead of using tmux, with the harnessdone signal, it's easier to just call the command directly from the appropriate dextest directory. 

See e.g. `func harnessCtl` in **loadbot.go**.

client/core/trade_simnet_test.go
```diff
@@ -1293,34 +1348,207 @@ func tryUntil(ctx context.Context, tryDuration time.Duration, tryFn func() bool)
...
-	daemon     string // indicates conf (beta.conf)
-	account    string // for dcr wallets
-	walletName string // for btc wallets, put into config map
-	pass       []byte
-	config     map[string]string
-	_type      string // type is a keyword
-	created    bool
+	pass   []byte
+	config map[string]string
+	_type  string // type is a keyword
```
buck54321, line 1423: While you're here, can you change this to `walletType` or something.

buck54321 (approved):
This works. I'll want to come back and make this easier to set the assets later. Just made a couple of suggestions.

PR #1605, opened 2022-05-05
client: Update server host

martonp:
DEX servers are able to advertise on multiple host names. This update gives users of the dex client the ability to switch between the different host names of the same dex.

- client/core: Add `UpdateDexHost` method which takes
  a current dex host and a new host name, and if the dex
  at both host names use the same dex pub key, the host
  name is updated. Also, adding a dex with the same pub
  key as a dex which is already registered is now blocked.

- ui: Adds a button on the dex settings page to allow updating
  the dex host. The DEXAddressForm is reused for this purpose.
  The DEXAddressPopup is also updated to not show known
  exchanges that the user is already registered for.

### DIFF
--- a/client/core/account.go
+++ b/client/core/account.go
@@ -9,6 +9,31 @@ import (
 	"github.com/decred/dcrd/dcrec/secp256k1/v4"
 )
 
+// disconnectDEX unsubscribes from the dex's orderbooks, ends the connection
+// with the dex, and removes it from the connection map.
+func (c *Core) disconnectDEX(dc *dexConnection) {
+	// Stop dexConnection books.
+	dc.cfgMtx.RLock()
+	for _, m := range dc.cfg.Markets {
+		// Empty bookie's feeds map, close feeds' channels & stop close timers.
+		dc.booksMtx.Lock()
+		if b, found := dc.books[m.Name]; found {
+			b.closeFeeds()
+			if b.closeTimer != nil {
+				b.closeTimer.Stop()
+			}
+		}
+		dc.booksMtx.Unlock()
+		dc.stopBook(m.Base, m.Quote)
+	}
+	dc.cfgMtx.RUnlock()
+	// Disconnect and delete connection from map.
+	dc.connMaster.Disconnect()
+	c.connMtx.Lock()
+	delete(c.conns, dc.acct.host)
+	c.connMtx.Unlock()
+}
+
 // AccountDisable is used to disable an account by given host and application
 // password.
 func (c *Core) AccountDisable(pw []byte, addr string) error {
@@ -33,27 +58,8 @@ func (c *Core) AccountDisable(pw []byte, addr string) error {
 	if err != nil {
 		return newError(accountDisableErr, "error disabling account: %w", err)
 	}
-	// Stop dexConnection books.
-	dc.cfgMtx.RLock()
-	for _, m := range dc.cfg.Markets {
-		// Empty bookie's feeds map, close feeds' channels & stop close timers.
-		dc.booksMtx.Lock()
-		if b, found := dc.books[m.Name]; found {
-			b.closeFeeds()
-			if b.closeTimer != nil {
-				b.closeTimer.Stop()
-			}
-		}
-		dc.booksMtx.Unlock()
 
-		dc.stopBook(m.Base, m.Quote)
-	}
-	dc.cfgMtx.RUnlock()
-	// Disconnect and delete connection from map.
-	dc.connMaster.Disconnect()
-	c.connMtx.Lock()
-	delete(c.conns, dc.acct.host)
-	c.connMtx.Unlock()
+	c.disconnectDEX(dc)
 
 	return nil
 }
@@ -213,4 +219,73 @@ func (c *Core) UpdateCert(host string, cert []byte) error {
 	}
 
 	return nil
+}
+
+// UpdateDEXHost updates the host for a connection to a dex. The dex at oldHost
+// and newHost must be the same dex, which means that the dex at both hosts use
+// the same public key.
+func (c *Core) UpdateDEXHost(oldHost, newHost string, appPW []byte, certI interface{}) (*Exchange, error) {
+	if oldHost == newHost {
+		return nil, errors.New("old host and new host are the same")
+	}
+
+	crypter, err := c.encryptionKey(appPW)
+	if err != nil {
+		return nil, codedError(passwordErr, err)
+	}
+
+	oldDc, _, err := c.dex(oldHost)
+	if err != nil {
+		return nil, err
+	}
+
+	if oldDc.hasActiveOrders() {
+		return nil, fmt.Errorf("cannot update host while dex has active orders")
+	}
+
+	if oldDc.acct.dexPubKey == nil {
+		return nil, fmt.Errorf("cannot update host if dex public key is nil")
+	}
+
+	if oldDc.getPendingFee() != nil {
+		// The notify fee code will have to be completely refactored to allow
+		// this.
+		return nil, fmt.Errorf("host cannot be updated while registration fee is pending")
+	}
+
+	var updatedHost bool
+	newDc, err := c.tempDexConnection(newHost, certI)
+	if newDc != nil { // (re)connect loop may be running even if err != nil
+		defer func() {
+			// Either disconnect or promote this connection.
+			if !updatedHost {
+				newDc.connMaster.Disconnect()
+				return
+			}
+			c.upgradeConnection(newDc)
+		}()
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	if !newDc.acct.dexPubKey.IsEqual(oldDc.acct.dexPubKey) {
+		return nil, fmt.Errorf("the dex at %s does not have the same public key as %s",
+			oldHost, newHost)
+	}
+
+	c.disconnectDEX(oldDc)
+
+	_, err = c.discoverAccount(newDc, crypter)
+	if err != nil {
+		return nil, err
+	}
+
+	err = c.db.DisableAccount(oldDc.acct.host)
+	if err != nil {
+		return nil, newError(accountDisableErr, "error disabling account: %w", err)
+	}
+
+	updatedHost = true
+	return newDc.exchangeInfo(), nil
 }
--- a/client/core/account_test.go
+++ b/client/core/account_test.go
@@ -10,6 +10,7 @@ import (
 
 	"decred.org/dcrdex/client/db"
 	"decred.org/dcrdex/dex/encode"
+	"decred.org/dcrdex/dex/msgjson"
 	"decred.org/dcrdex/dex/order"
 	"decred.org/dcrdex/server/account"
 	"github.com/decred/dcrd/dcrec/secp256k1/v4"
@@ -145,9 +146,11 @@ func TestAccountDisable(t *testing.T) {
 		defer rig.shutdown()
 		tCore := rig.core
 		rig.crypter.(*tCrypter).recryptErr = test.recryptErr
+		rig.db.disabledHost = nil
 		rig.db.disableAccountErr = test.disableAcctErr
 		tCore.connMtx.Lock()
 		tCore.conns[tDexHost].trades = test.activeTrades
+
 		if test.loseConns {
 			// Lose the dexConnection
 			delete(tCore.conns, tDexHost)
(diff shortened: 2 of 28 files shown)


buck54321 (changes requested):

client/core/account_test.go
```diff
@@ -256,6 +257,102 @@ func TestUpdateCert(t *testing.T) {
...
+	newPrivKey, _ := secp256k1.GeneratePrivateKey()
+	newPubKey := newPrivKey.PubKey()
+
+	tests := []struct {
+		name        string
+		oldHost     string
+		newHost     string
+		feePending  bool
+		expectError bool
+		newPubKey   *secp256k1.PublicKey
```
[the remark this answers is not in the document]
martonp: `newPubKey` isn't always the same, but you're right about `newHost`.

client/webserver/site/src/js/forms.ts
```diff
@@ -1123,8 +1124,13 @@ export class DEXAddressForm {
...
     else Doc.show(page.appPWBox, page.auth)
-    Doc.hide(page.customBox)
-    Doc.show(page.showCustom)
+    if (this.knownExchanges.length === 0) {
+      Doc.show(page.customBox)
+      Doc.hide(page.showCustom, page.knownXCs, page.pickServerMsg, page.addCustomMsg)
+    } else {
+      Doc.hide(page.customBox)
+      Doc.show(page.showCustom)
+    }
```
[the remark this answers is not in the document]
martonp: Sorry about that, I think it must be because I forgot to update the cache buster. I've updated it now.

JoeGruffins (approved):
Testing well for me.

client/webserver/site/src/localized_html/en-US/bodybuilder.tmpl
```diff
@@ -5,10 +5,10 @@
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   {{- /* The above 2 meta tags *must* come first in the head; any other head content must come *after* these tags */ -}}
-  <link rel="icon" href="/img/favicon.png?v=z8A71BD">
+  <link rel="icon" href="/img/favicon.png?v=h877Tgq">
   <meta name="description" content="Decred DEX Client Web Portal">
   <title>{{.Title}}</title>
-  <link href="/css/style.css?v=lDtQNP" rel="stylesheet">
+  <link href="/css/style.css?v=h877Tgq" rel="stylesheet">
```
chappjc, line 11: No changes to favicon or css afaict.  No complaints though if you wanna keep the js and css matching, but the favicon can stay.

server/cmd/dcrdex/main.go
```diff
@@ -31,7 +31,7 @@ func mainCore(ctx context.Context) error {
 	// Parse the configuration file, and setup logger.
 	cfg, opts, err := loadConfig()
 	if err != nil {
-		fmt.Printf("Failed to load dcrdata config: %s\n", err.Error())
+		fmt.Printf("Failed to load dcrdex config: %s\n", err.Error())
```
chappjc, line 34: LOL, that one sure went unnoticed for a while!

client/webserver/api.go
```diff
@@ -418,6 +418,40 @@ func (s *WebServer) apiUpdateCert(w http.ResponseWriter, r *http.Request) {
 	writeJSON(w, simpleAck(), s.indent)
 }
 
+func (s *WebServer) apiUpdateDEXHost(w http.ResponseWriter, r *http.Request) {
+	form := new(struct {
```
chappjc, line 422: All the other literals in this file are defined with `&struct {` rather than `new(struct {`, so just for the sake of consistency lets use `&` here too.

client/core/account.go
```diff
@@ -214,3 +214,66 @@ func (c *Core) UpdateCert(host string, cert []byte) error {
...
+			}
+			c.upgradeConnection(newDc)
+		}()
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	if newDc.acct.dexPubKey == nil || oldDc.acct.dexPubKey == nil {
+		return nil, fmt.Errorf("cannot update host if dex public key is nil")
```
chappjc, line 259: Not sure if or how this would happen, but you might as well do just the following above, right after `c.dex(oldHost)` (before bothering with a temp connection to `newHost`:

```go
if oldDc.acct.dexPubKey == nil {
	return nil, fmt.Errorf("cannot update host if dex public key is nil")
}
```

Then the following check that the old and new pubkey are equivalent is sufficient.

client/core/core.go
```diff
@@ -2802,6 +2802,38 @@ func (c *Core) discoverAccount(dc *dexConnection, crypter encrypt.Crypter) (bool
...
+		if dexConnection.acct.dexPubKey == nil {
+			continue
+		}
+
+		if dexConnection.acct.dexPubKey.IsEqual(pubKey) {
+			return true, host, nil
+		}
+	}
+
+	return false, "", nil
```
chappjc, line 2821: This function never errors.  Also I'd suggest the following:

```go
func (c *Core) dexWithPubKeyExists(pubKey *secp256k1.PublicKey) (bool, string) {
		if dc.acct.dexPubKey == nil {
			continue
		}

		if dc.acct.dexPubKey.IsEqual(pubKey) {
			return true, dc.acct.host
		}
	}

	return false, ""
}
```

client/core/core.go
```diff
@@ -2802,6 +2802,38 @@ func (c *Core) discoverAccount(dc *dexConnection, crypter encrypt.Crypter) (bool
...
+}
+
+// upgradeConnection promotes a temporary dex connection and starts listening
+// to the messages it recieves.
```
chappjc, line 2825: receives

client/core/core.go
```diff
@@ -2802,6 +2802,38 @@ func (c *Core) discoverAccount(dc *dexConnection, crypter encrypt.Crypter) (bool
...
+func (c *Core) upgradeConnection(dc *dexConnection) {
+	atomic.StoreUint32(&dc.reportingConnects, 1)
+	c.wg.Add(1)
+	go c.listen(dc)
+	go dc.subPriceFeed()
```
chappjc, line 2830: If you switch to the following, I believe you can use `upgradeConnection` from `(*Core).register` as well:

```go
if atomic.CompareAndSwapUint32(&dc.reportingConnects, 0, 1) {
	c.wg.Add(1)
	go c.listen(dc)
	go dc.subPriceFeed()
}
c.connMtx.Lock()
c.conns[dc.acct.host] = dc
c.connMtx.Unlock()
```

client/webserver/api.go
```diff
@@ -418,6 +418,40 @@ func (s *WebServer) apiUpdateCert(w http.ResponseWriter, r *http.Request) {
...
+	defer form.Pass.Clear()
+	if !readPost(w, r, form) {
+		return
+	}
+	pass, err := s.resolvePass(form.Pass, r)
+	if err != nil {
+		s.writeAPIError(w, fmt.Errorf("password error: %w", err))
+		return
+	}
+	cert := []byte(form.Cert)
```
chappjc, line 437: Need a `defer zero(pass)` too I think.

client/webserver/api.go
```diff
@@ -418,6 +418,40 @@ func (s *WebServer) apiUpdateCert(w http.ResponseWriter, r *http.Request) {
...
+	exchange, err := s.core.UpdateDEXHost(form.OldHost, form.NewHost, pass, cert)
+	if err != nil {
+		s.writeAPIError(w, fmt.Errorf("error updating host: %w", err))
+		return
+	}
+
+	resp := struct {
+		OK       bool           `json:"ok"`
+		Exchange *core.Exchange `json:"xc,omitempty"`
+		Paid     bool           `json:"paid"`
```
chappjc, line 447: `Paid` is copy pasta or do you want to have `UpdateDEXHost` provide that detail?

chappjc (changes requested):

client/core/account.go
martonp: Would the pub key be nil if the DEX was created before the pub key scheme was implemented? I saw some code referring to that.

client/core/account.go
```diff
@@ -214,3 +214,66 @@ func (c *Core) UpdateCert(host string, cert []byte) error {
...
+
+	err = c.AccountDisable(appPW, oldHost)
+	if err != nil {
+		return nil, fmt.Errorf("failed to disable old account: %w", err)
+	}
+
+	_, err = c.discoverAccount(newDc, crypter)
+	if err != nil {
+		return nil, err
+	}
```
buck54321, line 275: I'm wondering if the order of these `AccountDisable` and `discoverAccount` should be switched.

client/webserver/http.go
```diff
@@ -88,19 +88,24 @@ type registerTmplData struct {
...
-		KnownExchanges:  s.knownExchanges(),
+		CommonArguments: *common,
+		KnownExchanges:  s.KnownUnregisteredExchanges(common.UserInfo.Exchanges),
 	})
 }
 
-func (s *WebServer) knownExchanges() []string {
+// KnownUnregisteredExchanges returns all the known exchanges that
+// the user has not registered for.
+func (s *WebServer) KnownUnregisteredExchanges(registeredExchanges map[string]*core.Exchange) []string {
```
buck54321, line 100: No need to export.

client/webserver/site/src/js/forms.ts
```diff
@@ -1175,20 +1193,24 @@ export class DEXAddressForm {
...
       await app().fetchUser()
       app().loadPage('markets')
       return
     }
-
     if (this.pwCache) this.pwCache.pw = pw
     this.success(res.xc, cert)
   }
 
+  makeDexUpdater (dexToUpdate: string) {
```
buck54321, line 1207: Please comment here that this is a one way function. The changes made during `makeDexUpdater` are not reversed upon `refresh`. So the same `DEXAddressForm` can't do double duty for new registration and host name updates.


I would actually prefer that `makeDexUpdater` was used internally based on an argument to the constructor.

client/webserver/site/src/html/dexsettings.tmpl
```diff
@@ -22,6 +22,9 @@
       <button id="updateCertBtn" class="bg2 selected">[[[Update TLS Certificate]]]</button>
```
buck54321, line 22: Not from this PR, but from https://github.com/decred/dcrdex/pull/1602#discussion_r878660443, I think only part of solution made it into the final product. 

Can you get rid of the `<h4>` up above and do something like

```HTML
<div class="flex-center fs16 mb-2">
  <span class="me-2 connection ico-connection d-hide" id="connectedIcon"></span>
  <span class="me-2 disconnected ico-disconnected d-hide" id="disconnectedIcon"></span>
  <span id="connectionStatus"></span>
</div>
```

client/core/account.go
```diff
@@ -214,3 +214,66 @@ func (c *Core) UpdateCert(host string, cert []byte) error {
...
+
+	if !newDc.acct.dexPubKey.IsEqual(oldDc.acct.dexPubKey) {
+		return nil, fmt.Errorf("the dex at %s does not have the same public key as %s",
+			oldHost, newHost)
+	}
+
+	err = c.AccountDisable(appPW, oldHost)
```
buck54321, line 267: If there are active orders, we'll fail here, but should we check that condition higher up to prevent making an unnecessary connection?

buck54321 (changes requested):

client/core/account.go
martonp: I changed it so that it disconnects the dex first, then does `discoverAccount`, then finally removes the old dex from the db.

client/core/account.go
chappjc: Oh now I remember. A dex server has always been identified by the pubkey, but it wasn't always returned in the `config` response, just the `register` result.
But the DB has always stored the dex pubkey so I don't quite know.  It's fine to check.

client/core/core.go
```diff
@@ -3214,15 +3245,7 @@ func (c *Core) register(dc *dexConnection, assetID uint32) (regRes *msgjson.Regi
 
 			// If called from DiscoverAccount, where a temporary connection is used,
 			// promote this dexConnection and start listening for messages.
```
chappjc, line 3247: Sorry, I'm just realizing that this comment is not right anymore since `register` hasn't been called from `DiscoverAccount` since https://github.com/decred/dcrdex/commit/d8e46bb21e6dcd299343f69257270529aaee6c7e.

Although the CAS will never be true as long as `Register` is calling `register` because `Register`'s `dexConnection` is not temporary, insertion into `c.conns` is still needed, so let's either replace the comment with something that describes the current flow, or remove this `upgradeConnection` and instead put the map insertion in the `if paid {` branch of the caller, `Register`.

client/core/account.go
```diff
@@ -9,6 +9,31 @@ import (
 	"github.com/decred/dcrd/dcrec/secp256k1/v4"
 )
 
+// disconnectDEX unsubsribes from the dex's orderbooks, ends the connection
```
chappjc, line 12: unsubsribes

client/webserver/site/src/js/forms.ts
chappjc: Remove `makeDexUpdater` now?  I believe you have it done in `if (dexToUpdate) {` now.

chappjc (changes requested):
A few little things left, but seems g2g.

chappjc (approved):
LGTM.  Working very well.

Might as well rebase while waiting for @buck54321 to circle back since your merge base is quite old.  I tested a rebased branch though.

buck54321 (approved):

PR #1606, opened 2022-05-06
bch: switch to testnet4 and use custom address stringer

chappjc:
Pulled out of a larger Electrum branch.

This does:
- Switch to testnet4 now that it's first class with bchd.  See https://github.com/gcash/bchd/releases/tag/v0.19.0 and https://read.cash/@bitcoincashnode/bch-testnet4-for-may-2022-network-upgrade-updated-2021-11-23-95a21d51
- creates a custom address stringer field of the btc client's `baseWallet` to resolve what appears to be incorrect conversion of `btcutil.Address` types to strings in places.  See commit message below.

```
client/asset: custom address stringer

The exchange wallets had an address decoder function for converting
a string to a btcutil.Address, which is required for assets like BCH
with special address formats. However, it is incorrect to then use
the btcutil.Address String method go get a human readable address.

This adds a complementary helper function for converting from a
btcutil.Address to a correct string. Normally this is simply using
the Address as a Stringer, but for BCH it is the new EncodeCashAddress
function, which powers the existing RecodeCashAddress.

This updates the btc baseWallet methods to use the new stringAddr
to achieve the reverse of decodeAddr. This ensures that the string
address stored in the fundingCoins is correct for the asset. This
also ensures the recipient address in AuditContract is correct.
I believe the main issue resolved by this is providing the correct
address to the node.privKeyForAddress method when signing.
This also provides the correct string to the getaddressinfo RPC.
It is not clear if bitcoincashd was previously ok with the incorrect
address strings when handling RPCs.

This also removes the unused (*auditInfo).Recipient method, and
fixes the documentation for the other methods since there is no longer
an asset.AuditInfo *interface*, just a struct.
```


client/asset/btc/btc.go
```diff
@@ -332,27 +336,19 @@ func (op *output) wireOutPoint() *wire.OutPoint {
...
 
-// Recipient returns a base58 string for the contract's receiving address. Part
-// of the asset.AuditInfo interface.
-func (ci *auditInfo) Recipient() string {
-	return ci.recipient.String()
-}
-
 // Expiration returns the expiration time of the contract, which is the earliest
-// time that a refund can be issued for an un-redeemed contract. Part of the
-// asset.AuditInfo interface.
```
chappjc, line 349: At one point there was an `asset.AuditInfo` **interface**, but it it became a concrete type at one point and these remained.

client/asset/btc/btc.go
```diff
@@ -3343,8 +3364,11 @@ func (btc *baseWallet) createSig(tx *wire.MsgTx, idx int, pkScript []byte, addr
 // input and the pubkey associated with the address.
 func (btc *baseWallet) createWitnessSig(tx *wire.MsgTx, idx int, pkScript []byte,
 	addr btcutil.Address, val uint64, sigHashes *txscript.TxSigHashes) (sig, pubkey []byte, err error) {
-
-	privKey, err := btc.node.privKeyForAddress(addr.String())
+	addrStr, err := btc.stringAddr(addr, btc.chainParams)
+	if err != nil {
+		return nil, nil, err
+	}
+	privKey, err := btc.node.privKeyForAddress(addrStr)
```
chappjc, line 3371: Not sure, but perhaps `bitcoincashd` (Bitcoin Cash Node)` was able to pull out the pubkey from the bitcoin addresses and locate the private key.  The chain params were the clone params with the correct bch address ID values, but this wasn't making Cash Addresses.

chappjc: Yeah, that's what it was doing.

```
$  ./beta getaddressinfo mgDzcoxdBrhu435knkdUEP5UGVYGKUJJgN
{
  "address": "bchreg:qqru9lns98ugq85h4cl3p8srf4y9w8cyt5eux55xmn",
  "scriptPubKey": "76a91407c2fe7029f8801e97ae3f109e034d48571f045d88ac",
  "ismine": true,
  "iswatchonly": false,
  "isscript": false,
  "pubkey": "0310892969dde0c7f6e177ce0f1a8ba807fbbbbc05135e9c47479f0aa1fa2555a5",
...
```

buck54321:
`AddressStringer` part is fixed in #1570 too.

chappjc:
> `AddressStringer` part is fixed in #1570 too.

Ah, dang.  Hopefully we did things similarly.
I guess I'll put the zcash pr at the top of my queue

chappjc:
We're close.  I can minimize conflicts because having this fix separate makes sense IMO, but some points:
- I can go with `stringifyAddress` if you prefer that to `stringAddr`
- I'd prefer the `EncodeCashAddress` helper extracted from `RecodeCashAddress` because it avoids a needless encode->decode round trip before getting to the switch
- Can you convince me that it's OK to ignore the error? It shouldn't happen I guess, and if it fails then it would be an empty string, and that could be caught or allowed to error in the next call I suppose
- Good catch that `BCHWallet` no longer needs to exist.

buck54321:
Your way is better. I'll adapt.

chappjc:
Shall I merge this then?

chappjc:
Or just discard the stringer commit?  Whatever

client/asset/btc/btc.go
buck54321: Interesting.

buck54321 (approved):

buck54321:
Any leads on getting some testnet4 BCH?

chappjc:
Trade on testnet for the record:

image

https://testnet.dcrdata.org/tx/280fb46b603bc0dbb188516af28e14174ac0534ef57d216a8ee584c59193ab02/in/0
https://tbch4.loping.net/tx/17d3476d3e152d8eaa28f95aae127c3493ae966481bc6062c93ef7b8bf085e5e
https://tbch4.loping.net/tx/be067cc99cac47a9f19da556a664cbb4b364214bae3984eb3ce47a9e0249597c

PR #1607, opened 2022-05-06
Electrum exchange wallet (BTC, LTC, BCH)

chappjc:
This adds support for the common Electrum SPV wallet and forks.  This includes BTC, LTC, BCH, and others.  BCH needs an [upstream fix](https://github.com/Electron-Cash/Electron-Cash/pull/2399), but it should work soon.  This PR has been tested on both BTC and LTC forks (testnet). Use at least version 4.2.x of both.  Litecoin testnet is forked right now because of MWEB, so that's a pain to test now although I tested it before - you can force your Electrum-LTC to use the "dex-test.ssgen.io:52002" server in the "Network" settings, and you will see a message that there is a chain split with the other two public servers on the dead non-MWEB chain.

First, some background on the Electrum model:
- There are Electrum **servers** ("electrumx") that use a [well-defined protocol](https://electrumx.readthedocs.io/en/latest/protocol.html) that is designed to be usable for [numerous assets](https://github.com/spesmilo/electrumx/blob/master/electrumx/lib/coins.py) (Decred included!).  This is primarily because address info is requested based on a hash of the payment/output script (pkScript).  It is also a simple JSON-RPC protocol that is commonly implemented on a variety of transports.
- Electrum **wallet** forks are common: BTC (https://electrum.org/), LTC (https://electrum-ltc.org/), BCH (https://electroncash.org/), etc.
- The Electrum wallet software has its own [JSON-RPC server](https://electrum.readthedocs.io/en/latest/jsonrpc.html) that uses http post requests.
- You sacrifice significant privacy in using an Electrum wallet since it reveals your entire address set to the servers.
- Certain operations are dead simple because it is much like requesting data from a block explorer.  e.g. listing unspent outputs or even entire use history for any address, not just your own.  This makes finding and locating counterparty transactions, including swaps and redemptions, simple.

This PR is broken into several commits to ease review.  Only three of the commits deal directly with Electrum.
- `electrum: add wallet client` Adds the HTTP JSON-RPC client for making requests to the local Electrum wallet instance.
- `electrum: add network/server client` Adds a TCP JSON-RPC client for making requests to remote ElectrumX servers.
- `client/asset/btc: add ElectrumExchangeWallet` Uses both clients to define an `ElectrumExchangeWallet` that satisfies the `client/asset.Wallet` interface.  It creates new wallet definitions for the assets that have usable Electrum wallets.

The message from the final **`client/asset/btc: add ElectrumExchangeWallet`** commit:

```
This creates the ExchangeWalletElectrum type for using an external
Electrum wallet along with an ElectrumX server.

This is intended to support Electrum forks for LTC, BCH, and BTC.

The new walletTypeElectrum is defined as "electrumRPC".

The ExchangeWalletElectrum type overrides some baseWallet methods:
- FindRedemption since the process is very different (and simpler)
  from that used by block filter-enabled wallets.
- feeRate, to avoid pointless conversions and btcjson types
- watchBlocks and reportNewTip, since it is not necessary to retarget a
  starting point for redemption searches.
- Connect, since it sets estimateFee to the overridden feeRate method,
  and starts the overridden watchBlocks instead of the baseWallet
  method.

It also satisfies the Sweeper and FeeRater interfaces.

To power ExchangeWalletElectrum, we define an electrumWallet type
that satisfies the asset.Wallet interface similar to rpcWallet and
spvWallet.

The electrumWallet type is powered by an electrumWalletClient, which
communicates with the external Electrum wallet via RPC, and an
electrumNetworkClient, which communicates with a remote Electrum(X)
server for block headers and verbose transaction data.

The general design of electrumWallet is to govern electrumNetworkClient
so that it tracks the electrum server in use by the Electrum *wallet*.
This is done for a couple of reasons:
- ensure consistent results between the two
- allow the Electrum wallet to handle server selection
```

To enable an Electrum wallet's JSON-RPC server, open the Console in Electrum (you may need to enable the console from the "View" menu), and run the following commands, with a port and password of your choice:

```
setconfig("rpcuser","user")
setconfig("rpcpassword","pass")
setconfig("rpcport","6789")
setconfig("rpchost","127.0.0.1")
```

Then restart Electrum.  Test: `curl --data-binary '{"jsonrpc":"2.0","id":"curltext","method":"getbalance","params":[]}' http://user:pass@127.0.0.1:6789`

**Caveats and TODOs**

There is no `listlockunspent` RPC, so the `electrumWallet` just keeps a map of outpoints.

The `listunspent` RPC result includes locked/frozen utxos, so we have to manually filter them out in the application when funding transactions.  All coin locking does is prevent the wallet's `payTo` RPC from spending those locked coins.

<s>UTXO locking is not implemented in the LTC fork, instead only offering _address_ locking.  This only a possible issue for available vs. locked balance display if there are several utxos for the same address but not all of them are actually locked, meaning withdraw will not be able to spend them until the address is unlocked.</s> `freeze_utxo` is now available in the latest [Electrum-LTC release 4.2.2.1](https://github.com/pooler/electrum-ltc/releases/tag/4.2.2.1), available at https://electrum-ltc.org/

Test coverage is sparse because it's just a ton of stubbing stuff out.  There's no easy way to test on simnet either, only testnet.

image

image


LasTshaMAN (commented):
Hi there, spotted couple of typos passing by, awesome work!

chappjc:
Just rebased, but not retested. **Most recently tested commit** is https://github.com/decred/dcrdex/pull/1607/commits/07cceb5380c8a0021f2408ebad906da62579f5e6

chappjc:
I left a couple of things unfinished w.r.t. coin locking.  Hoping for electrum ltc to get a release with `freeze_utxo`.  Even with that though, I realized I left a few TODOs around this stuff.  Otherwise OK to review.

client/asset/btc/electrum_client.go
```diff
@@ -0,0 +1,1289 @@
...
+	ew.passMtx.Lock()
+	ew.pass = ""
+	ew.passMtx.Unlock()
+	return nil
+}
+
+func (ew *electrumWallet) locked() bool {
+	ew.passMtx.RLock()
+	defer ew.passMtx.RUnlock()
+	return ew.pass == ""
```
martonp, line 871: I was able to create the electrum wallet without a password.

client/asset/btc/electrum_client.go
```diff
@@ -0,0 +1,1289 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error listing unspent outputs: %w", err)
+	}
+
+	// Filter out locked outpoints since listUnspent includes them.
+	unspentMap := make(map[RPCOutpoint]*ListUnspentResult, len(unspents))
+	for _, unspent := range unspents {
+		unspentMap[RPCOutpoint{unspent.TxID, unspent.Vout}] = unspent
+	}
+	locked, _ := ew.listLockUnspent() // does not error
```
martonp, line 1063: The implementation might change later and this would still not be checking for an error.

martonp (commented):
Looks good, haven't tested yet though. Is the plan to use this with the lightning network later?

client/asset/btc/electrum/network.go
```diff
@@ -0,0 +1,646 @@
...
+			err = json.Unmarshal(msg, &ntfnParams)
+			if err != nil {
+				fmt.Println("Unmarshal error:", err)
+				continue
+			}
+			for _, c := range sc.subChans(jsonResp.Method) {
+				select {
+				case c <- ntfnParams.Params:
+				default:
+				}
```
JoeGruffins, line 87: I'm not sure but do you need to hold the `ntfnHandlers` here? You hold it when getting with `subChans` but it looks like they could be closed while sending here.

JoeGruffins (commented):
Partial review.

client/asset/btc/btc.go
```diff
@@ -4113,9 +4138,12 @@ func (btc *baseWallet) spendableUTXOs(confs uint32) ([]*compositeUTXO, map[outPo
 		// unlocked something or even restarted the wallet software.
 		pt := newOutPoint(utxo.txHash, utxo.vout)
 		if btc.fundingCoins[pt] != nil {
+			// This may happen Electrum, which returns even locked utxos in the
+			// listunspent RPC. Maybe we re-lockunspent if we hit this so
+			// electrumWallet's lockedOutpoints map is updated to prevent the
+			// listUnspent method from returning this again. Also, we should
+			// probably remove this outpoint from the slice and not return it!
 			btc.log.Warnf("Known order-funding coin %s returned by listunspent!", pt)
-			// TODO: Consider relocking the coin in the wallet.
```
chappjc, line 4079: eh, removing the utxo from the return was what I was considering here.  Part of the unresolved coin locking issues for electrum...

client/asset/btc/electrum_client.go
```diff
@@ -0,0 +1,1289 @@
...
+			ew.log.Warnf("Output (%v:%d) with bad value %v found: %v",
+				utxo.PrevOutHash, utxo.PrevOutIdx, val, err)
+			continue
+		}
+		var confs uint32
+		if height := int32(utxo.Height); height > 0 {
+			if height > chainHeight {
+				confs = 1 // conf after getinfo call
+			} else {
+				confs = uint32(chainHeight - height)
```
chappjc, line 655: Needs a +1

client/asset/btc/electrum_client.go
```diff
@@ -0,0 +1,1289 @@
...
+		val, err := strconv.ParseFloat(utxo.Value, 64)
+		if err != nil {
+			ew.log.Warnf("Output (%v:%d) with bad value %v found: %v",
+				utxo.PrevOutHash, utxo.PrevOutIdx, val, err)
+			continue
+		}
+		var confs uint32
+		if height := int32(utxo.Height); height > 0 {
+			if height > chainHeight {
+				confs = 1 // conf after getinfo call
```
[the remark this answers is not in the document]
chappjc: This is detecting the race in the two RPCs.  First `chainHeight = getBestBlockHeight()`, then `utxos = listunspent()`.
If `utxo.Height` is set at all, that indicates it is in a block, not unconfirmed.  But I do a quick check to ensure this height is not greater than `chainHeight` we got from `getBestBlockHeight`. If it is, the tx was confirmed between the two RPCs.  The purpose is to ensure we don't underflow confs

client/asset/btc/electrum_client.go
```diff
@@ -0,0 +1,1289 @@
...
+						return err
+					}
+					ew.lockedOutpointsMtx.Lock()
+					delete(ew.lockedOutpoints, op)
+					ew.lockedOutpointsMtx.Unlock()
+					delete(opMap, op)
+					continue unspents
+				}
+
+				if err = ew.wallet.FreezeAddress(utxo.Address); err != nil {
```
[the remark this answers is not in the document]
chappjc: Exactly.  The ability to use `payTo` as expected is hindered until `freeze_utxo` is available for LTC.  

I called this out in different language in the PR description:

> UTXO locking is not implemented in the LTC fork, instead only offering address locking. This only a possible issue for available vs. locked balance display if there are several utxos for the same address but all of them are actually locked, meaning withdraw will not be able to spend them until the address is unlocked.

client/asset/btc/electrum_client.go
chappjc: Oh!  I guess I'll change this scheme

client/asset/btc/electrum/network.go
```diff
@@ -0,0 +1,646 @@
...
+		for _, feat := range peer.Feats {
+			if len(feat) < 2 {
+				continue
+			}
+			switch []rune(feat)[0] {
+			case 't':
+				if !isOnion {
+					continue
+				}
+				port := feat[1:]
```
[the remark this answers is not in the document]
chappjc: Good point.  I think I might make a comment that we are going to ignore those.  In my observations it is always set, so it shouldn't limit the server selection much.  The reason I'm hesitant to define a default port is because the default is both ill-defined (seeming to be organic convention but not documented) and different between assets.

client/asset/btc/electrum_client.go
```diff
@@ -0,0 +1,1289 @@
...
+}
+
+func (ew *electrumWallet) balances() (*GetBalancesResult, error) {
+	eBal, err := ew.wallet.GetBalance()
+	if err != nil {
+		return nil, err
+	}
+	return &GetBalancesResult{
+		Mine: Balances{
+			Trusted: eBal.Confirmed + eBal.Unconfirmed,
```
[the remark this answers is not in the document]
chappjc: Unconfirmed does not necessarily imply untrusted.  Since we don't have a good indication of what part of the unconfirmed balance is trusted or not, I put it here so that it will be `Available` in `(*baseWallet).Balance`

For DCR and BTC at least, we allow use of unconfirmed utxos (but preferring confirmed), so to stick with that I just did this.  The important rule for the `btc.Wallet` backends is that the result of `balances` must agree with `listUnspent` in terms of the trusted/untrusted balance and the "safe" field of `ListUnspentResult`.

So, the options are: (a) declare all unconfirmed utxos **unsafe** in both methods, diverging from how we operate with the other BTC wallet backends, (b) make no attempt at determining "safeness" (see DOGE approach that leaves SafePtr unset and uses legacyBalance with no notion of trustedness and simply combining confirmed+unconfirmed balances), or (c) make like `spvWallet` and see if we can make an [`ownsInputs` method](https://github.com/decred/dcrdex/commit/55c890c4815c8baa28476c820225d10a45e99010) without making electrumx server requests.

I think I may be able to do (c) without expensive operations, but I'm not hopeful.

client/asset/btc/electrum/network.go
```diff
@@ -0,0 +1,646 @@
...
+}
+
+func (sc *ServerConn) pinger(ctx context.Context) {
+	t := time.NewTicker(pingInterval)
+	defer t.Stop()
+
+	for {
+		// listen => ReadBytes cannot wait forever. Reset the read deadline for
+		// the next ping's response, as the ping loop is running.
+		err := sc.conn.SetReadDeadline(time.Now().Add(pingInterval * 5 / 4))
```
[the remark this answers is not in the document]
chappjc: It's global, just like with gorilla.  We design around that because when the last-set deadline is exceeded, all read operations fail instead of blocking, allowing the read loop to break as well.

If the ping takes over 5 seconds to get a response, we should cut off everything and fail the connection.

client/asset/btc/electrum/network.go
```diff
@@ -0,0 +1,646 @@
...
+	return sc.done
+}
+
+func (sc *ServerConn) send(msg []byte) error {
+	err := sc.conn.SetWriteDeadline(time.Now().Add(7 * time.Second))
+	if err != nil {
+		return err
+	}
+	_, err = sc.conn.Write(msg)
+	return err
```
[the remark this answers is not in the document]
chappjc: It's also in the [design for `net.Conn` that it's safe for concurrent use](https://pkg.go.dev/net#Conn). We won't get data races or interleaved packets or anything else nasty.

> Multiple goroutines may invoke methods on a Conn simultaneously.

client/asset/btc/electrum/network.go
chappjc: They are closed in the defer of this function, thus synchronized with these sends.  I'm adding docs to both `deleteSubscriptions` and `cancelRequests` to make this clearer.
The rule of thumb with closing channels is to only close them in the goroutine that _sends_ on the channel, which is why I had a comment on the defer.  It can be clearer though.

client/asset/btc/btc.go
```diff
@@ -3059,13 +3113,9 @@ func (btc *baseWallet) LocktimeExpired(contract dex.Bytes) (bool, time.Time, err
 // This method blocks until the redemption is found, an error occurs or the
 // provided context is canceled.
 func (btc *baseWallet) FindRedemption(ctx context.Context, coinID, _ dex.Bytes) (redemptionCoin, secret dex.Bytes, err error) {
-	exitError := func(s string, a ...interface{}) (dex.Bytes, dex.Bytes, error) {
-		return nil, nil, fmt.Errorf(s, a...)
-	}
```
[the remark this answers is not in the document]
chappjc: Doesn't add anything IMO.  It just makes me look up at what `exitError` is instead of just seeing right at the `return` what it is.  I think this pattern also leads us to not wrap errors for this reason - we could use %w but we don't because we're not giving it directly to `fmt.Errorf` it it feels weird, I guess.

I understand the purpose of this kind of thing in certain cases where you want to be certain to zero out the returns on error, esp. when using named return parameters, but I don't sense we're getting that benefit here.

client/asset/btc/electrum.go
```diff
@@ -0,0 +1,330 @@
...
+
+// watchBlocks pings for new blocks and runs the tipChange callback function
+// when the block changes.
+func (btc *ExchangeWalletElectrum) watchBlocks(ctx context.Context) {
+	const electrumBlockTick = 5 * time.Second
+	ticker := time.NewTicker(electrumBlockTick)
+	defer ticker.Stop()
+
+	// TODO: Why is currentTip a field, esp. one guarded by a mutex? It's use
+	// can be limited to the watchBlocks loop.
```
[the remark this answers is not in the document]
chappjc: Oh it's getting initialized in `(*baseWallet).connect` before starting this loop.  I don't think that has to be the case though, if we don't mind a second getbestblock call when starting the monitoring loop right after actually the same RPC when connecting to the backend in `connect`.
I think I'll leave this design for now, but it's awkward.

client/asset/btc/electrum/network.go
```diff
@@ -0,0 +1,646 @@
...
+				}
+			}
+			continue
+		}
+		// fmt.Println(string(msg))
+
+		c := sc.responseChan(jsonResp.ID)
+		if c == nil {
+			fmt.Printf("Received response for unknown request ID %d", jsonResp.ID)
+			continue
```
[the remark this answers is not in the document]
chappjc: > If the connection while there are some pending responses, wouldn't this error message print for every response received after it is closed?

Can you expand on this?  I'm not seeing it.  The `defer sc.cancelRequests()` on this function comes _after_ this loop breaks.

client/asset/btc/electrum/network.go
```diff
@@ -0,0 +1,646 @@
...
+// SubscribeHeadersResult is the contents of a block header notification.
+type SubscribeHeadersResult struct {
+	Height int32  `json:"height"`
+	Hex    string `json:"hex"`
+}
+
+// SubscribeHeaders subscribes for block header notifications. There seems to be
+// no guarantee that we will be notified of all new blocks, such as when there
+// are blocks in rapid succession.
+func (sc *ServerConn) SubscribeHeaders(ctx context.Context) (*SubscribeHeadersResult, <-chan *SubscribeHeadersResult, error) {
```
chappjc, line 616: So I ended up not using the subscription machinery, but I'd really like to keep it in place in case it's needed.  Plus it rounds out ServerConn as an electrumx client.

client/asset/btc/electrum/network.go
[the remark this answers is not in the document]
JoeGruffins: So, setting the deadline also sets if for requests we are already waiting for?

client/asset/btc/electrum/network.go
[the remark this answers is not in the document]
JoeGruffins: Understood. Thank you.

client/asset/btc/electrum/network.go
JoeGruffins: Ok I see, `deleteSubscriptions` is only used at the end of this function in the defer.

client/asset/btc/electrum/network.go
JoeGruffins: This pr pending changes? Will continue review when they are in.

chappjc: I have revisions based on your guys' review, but nothing for this method

chappjc:
Still working on revisions.  Just pushing my work to change gears.

client/asset/btc/electrum_client.go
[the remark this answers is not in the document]
chappjc: I'll make an option for forks that do have `freeze_utxo`, like BTC's Electrum.

client/asset/btc/electrum/network.go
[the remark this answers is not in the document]
martonp: Nvm I must have misread it.

client/asset/btc/electrum_client.go
```diff
@@ -0,0 +1,1289 @@
...
+				}
+				delete(failing, info.Server) // clean slate now that wallet has just gotten on it
+				ew.log.Infof("Electrum wallet changed server to %v", info.Server)
+			}
+			lastWalletServer = info.Server
+
+			tryAddr := info.Server
+			if fails := failing[tryAddr]; fails > maxFails {
+				ew.log.Warnf("Server %q has failed to connect %d times. Trying a random one...", tryAddr, fails)
+				tryAddr = "" // try a random one instead
```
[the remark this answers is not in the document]
martonp: I would've thought it would do the same thing each time if the map is exactly the same. I think you should at least keep track of which ones failed when doing `failing[tryAddr]++` though, instead of just saying that picking a random one failed.

client/asset/btc/electrum_client.go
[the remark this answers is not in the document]
martonp: Yeah that makes sense. I was just thinking that a confirmation means that the mining of a tx in a block is confirmed by another block on top.

client/asset/btc/electrum/wallet_types.go
```diff
@@ -0,0 +1,69 @@
...
+	// Sighash "sighash": null,
+	// "unknown_psbt_fields": {},
+	// "utxo": null,
+	// "witness_utxo": null
+}
+
+// Balance is the result of the balance RPC.
+type Balance struct {
+	Confirmed   float64 // includes spent value that is not yet confirmed
+	Unconfirmed float64 // will be negative for sends
```
martonp, line 67: The comments for `Confirmed` and `Unconfirmed` seem to be contradictory. They both are saying that unconfirmed sent value is included.

client/asset/btc/electrum/network.go
```diff
@@ -0,0 +1,674 @@
...
+		if err != nil {
+			sc.debug("response Unmarshal error:", err)
+			continue
+		}
+
+		if jsonResp.Method != "" { // notification
+			var ntfnParams ntfnData // the ntfn payload is in the params field of a request object (!)
+			err = json.Unmarshal(msg, &ntfnParams)
+			if err != nil {
+				sc.debug("notification Unmarshal error:", err)
```
LasTshaMAN, line 104: %v is missing ?

client/asset/btc/electrum/network.go
```diff
@@ -0,0 +1,674 @@
...
+				sc.debug("ReadBytes: %v", err)
+			}
+			sc.cancel()
+			return
+		}
+
+		var jsonResp response
+		err = json.Unmarshal(msg, &jsonResp)
+		if err != nil {
+			sc.debug("response Unmarshal error:", err)
```
LasTshaMAN, line 96: %v is missing ?

client/asset/btc/electrum_client.go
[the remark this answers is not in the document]
LasTshaMAN: Probably off topic,

> as long as "not specified" means random, but that's probably a bad idea

Found it curious, that 1st element of the map gets chosen roughly 50% of the time (using range over map): https://go.dev/play/p/s1wrj_UY41Q?v=gotip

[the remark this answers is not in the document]
chappjc: That's quite the bias.  Yeah, definitely not an approach to take. Was being exceedingly lazy while hacking on this logic

chappjc:
@buck54321 Lemme know if you think it would be good to pop off the last commit and get the prerequisite commits reviewed and merged first.  Most of the TODOs, testing, and issues are likely to be in the ExchangeWallet bits.

client/asset/btc/electrum/wallet_types.go
chappjc: Yeah, it's madness, but that's how Electrum handles these fields.  If you send funds, the Confirmed field will not be reduced by that amount until the tx is confirmed ("includes spent value that is not yet confirmed"), while the Unconfirmed field includes this amount, but as negative.  Perhaps the comment on the Confirmed field is unclear.
I'll try this again from the Electrum console, but that's what I observed.

chappjc:
Electrum-LTC 4.2.2.1 released with `freeze_utxo`.  https://github.com/pooler/electrum-ltc/issues/288  Updated accordingly.  Not retested.

chappjc:
image

chain split is huge

chappjc:
Getting rid of the "falseMedianTime" now that the ZEC PR conveniently removed the median time field from `blockHeader` by making the `rpcBlockHeader`.  That was a great change @buck54321

chappjc:
Consider draft for now...

client/asset/btc/electrum/network.go
```diff
@@ -0,0 +1,646 @@
...
+
+	for {
+		if ctx.Err() != nil {
+			return
+		}
+		msg, err := reader.ReadBytes(newline)
+		if err != nil {
+			if ctx.Err() == nil { // unexpected
+				fmt.Println(err)
+			}
```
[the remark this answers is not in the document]
chappjc: It's indirect.  When the context is cancelled, there's a goroutine started in `ConnectServer` that closes the connection, which results in errors from all conn read/write calls.

```go
go func() {
	<-ctx.Done()
	conn.Close()
	close(sc.done)
}()
```

So if it's a shutdown, we don't want to log the error from `reader.ReadBytes`.  If the connection just died and `ReadBytes` is the first to find out (context not canceled yet), then we can log the error and initiate a shutdown of the other `ServerConn` goroutines.

client/asset/btc/electrum/network.go
```diff
@@ -0,0 +1,646 @@
...
+	Witness   []string   `json:"txinwitness,omitempty"`
+	Sequence  uint32     `json:"sequence"`
+	Coinbase  string     `json:"coinbase,omitempty"`
+}
+
+// PkScript represents the pkScript/scriptPubKey of a transaction output
+// returned by a transaction requests.
+type PkScript struct {
+	Asm       string   `json:"asm"`
+	Hex       string   `json:"hex"`
```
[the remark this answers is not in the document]
chappjc: Ah, true.  Although it would be a very misleading struct definition though.  The `Asm` and `Hex` in this `PkScript` type are the decode pkScript and the hex of the pkScript, neither the `SigScript`, which is something that belongs to an input, not an output.

Here's part of a `blockchain.transaction.get` result:

```json
"vout": [
    {
      "value": 1.21,
      "n": 0,
      "scriptPubKey": {
        "asm": "0 dd8fc00271f7543a7f85960fa94fabd49e451c8d",
        "hex": "0014dd8fc00271f7543a7f85960fa94fabd49e451c8d",
        "reqSigs": 1,
        "type": "witness_v0_keyhash",
        "addresses": [
          "tltc1qmk8uqqn37a2r5lu9jc86jnat6j0y28ydmfuadd"
        ]
      }
    },
```

chappjc:
Still a bit drafty, haven't re-tested.

Also will look into making electrumx server and Electrum wallets work on simnet.

chappjc:
To start a simnet electrumx server for the Bitcoin "alpha" harness:

```sh
$ mkdir -p /mnt/electrumx/btc_regtest
# go run github.com/decred/dcrd/cmd/gencerts@release-v1.7 -L ...
$ gencerts -L /mnt/electrumx/btc_regtest/ssl.cert /mnt/electrumx/btc_regtest/ssl.key
$ COIN="Bitcoin" NET="regtest" DB_ENGINE="leveldb" DB_DIRECTORY="/mnt/electrumx/btc_regtest" \
DAEMON_URL="http://user:pass@127.0.0.1:20556" SERVICES="tcp://localhost:54001,ssl://localhost:54002,rpc://" \
SSL_CERTFILE="/mnt/electrumx/btc_regtest/ssl.cert" SSL_KEYFILE="/mnt/electrumx/btc_regtest/ssl.key" \
PEER_DISCOVERY="off" PEER_ANNOUNCE= ./electrumx_server
```

To start Electrum wallet in regtest mode:

```
electrum --regtest
```

Make a new seed to ensure it is the latest features (segwit), then open the Network dialog and enter in your local server:

image

> Note: if you ever re-create the ssl certificate used by electrumx_server, the Electrum wallet will log an error that it has changed and refuse to connect to it.  If that happens, stop Electrum and `rm ~/.electrum/regtest/certs/127.0.0.1` to make it forget the old one.

Next go to the Console tab (if hidden, View -> Console) and set up the RPC server:

image

Restart Electrum and test the RPC server with curl:

```
$ curl -s --data-binary '{"jsonrpc":"2.0","id":"curltest","method":"getinfo","params":[]}' http://user:pass@127.0.0.1:6789 | jq '.'
{
  "id": "curltest",
  "jsonrpc": "2.0",
  "result": {
    "path": "/home/user/.electrum/regtest",
    "server": "127.0.0.1",
    "blockchain_height": 435,
    "server_height": 435,
    "spv_nodes": 1,
    "connected": true,
    "auto_connect": false,
    "version": "4.2.0",
    "default_wallet": "/home/user/.electrum/regtest/wallets/default_wallet",
    "fee_per_kb": 180000
  }
}
```

chappjc:
New block updates are laggy, so be sure to build the dcrdex server harness with longer lock times (`./build-lock && ./run`).

client/asset/btc/electrum_client.go
```diff
@@ -0,0 +1,1342 @@
...
+}
+
+func (ew *electrumWallet) pass() (pw string, unlocked bool) {
+	pw, unlocked = ew.pw.Load().(string)
+	return
+}
+
+// walletLock locks the wallet. Part of the btc.Wallet interface.
+func (ew *electrumWallet) walletLock() error {
+	ew.pw.Store(nil)
```
chappjc, line 878: ugh, this was dumb

client/asset/btc/electrum_client.go
```diff
@@ -0,0 +1,1338 @@
...
+	BlockHeaders(ctx context.Context, startHeight, count uint32) (*electrum.GetBlockHeadersResult, error)
+}
+
+// electrumNetworkManager provides a thread-safe way to use and reset the
+// mutable electrumNetworkClient field. Could replace this with a getter/setter
+// on electrumWallet, but that gets a little messy with defers capturing a
+// potentially old/replaced electrumNetworkClient.
+type electrumNetworkManager struct {
+	sync.RWMutex
+	electrumNetworkClient
```
buck54321, line 78: How about dropping the mutex in favor of an `atomic.Value` and a getter method?

```Go
type electrumWallet struct {
    ...
    chainV atomic.Value // electrumNetworkClient
    ...
}

func (ew *electrumWallet) chain() electrumNetworkClient {
	return ew.chainV.Load().(electrumNetworkClient)
}
```

client/asset/btc/electrum_client.go
```diff
@@ -0,0 +1,1338 @@
...
+func (ew *electrumWallet) externalAddress() (btcutil.Address, error) {
+	addr, err := ew.wallet.CreateNewAddress()
+	if err != nil {
+		return nil, err
+	}
+	return ew.decodeAddr(addr, ew.chainParams)
+}
+
+func (ew *electrumWallet) changeAddress() (btcutil.Address, error) {
+	return ew.externalAddress() // sadly, cannot request internal addresses
```
buck54321, line 796: That sucks.

buck54321 (changes requested):
Have gotten eyes on most of it now. Just some first thoughts.

client/asset/btc/btc.go
```diff
@@ -1966,34 +1998,62 @@ outer:
...
+		if len(txRaw) == 0 {
+			tx, err := btc.node.getWalletTransaction(txHash) // maybe getTxOut()? we don't need block info or the full tx
+			if err != nil {
+				return nil, err
 			}
-
-			continue outer
+			txRaw = tx.Hex
+		}
+		msgTx, err := msgTxFromBytes(txRaw)
```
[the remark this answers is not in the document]
chappjc: Already updated this.  I think you're a few commits behind

client/asset/btc/spv.go
```diff
@@ -1823,13 +1825,17 @@ func (w *spvWallet) getTransaction(txHash *chainhash.Hash) (*GetTransactionResul
 	// This is what the JSON-RPC does (and has since at least May 2018).
 	details, err := w.wallet.walletTransaction(txHash)
 	if err != nil {
+		if errors.Is(err, WalletTransactionNotFound) {
+			return nil, asset.CoinNotFoundError // for the asset.Wallet interface
+		}
```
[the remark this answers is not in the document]
chappjc: Yeah.  I had this as a separate commit since it was mostly unrelated: https://github.com/decred/dcrdex/pull/1607/commits/cb6fe1908b02e3e0ff65bcc9de10063b333a2678

client/asset/ltc/ltc.go
```diff
@@ -175,5 +189,15 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network)
 		BlockDeserializer:   dexltc.DeserializeBlockBytes,
 	}
 
-	return btc.BTCCloneWallet(cloneCFG)
+	switch cfg.Type {
+	case walletTypeRPC, walletTypeLegacy:
+		return btc.BTCCloneWallet(cloneCFG)
+	case walletTypeElectrum:
```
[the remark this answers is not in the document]
chappjc: Whoops, rebase slop.

client/asset/btc/btc.go
```diff
@@ -816,6 +827,10 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, net dex.Network) (ass
 			return nil, err
 		}
 		return &ExchangeWalletAccelerator{rpcWallet}, nil
+	case walletTypeElectrum:
+		cloneCFG.SingularWallet = true     // not used, but no path on RPC endpoint
```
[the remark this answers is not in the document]
chappjc: Sorry, this comment was confusing.  The field name is still pretty good, my comment just makes no sense.  I'll just remove this line since the actual code is ineffectual because the request code does not use this variable, instead just doing `"http://"+rpcCfg.RPCBind` in `func ElectrumWallet`.

client/asset/btc/wallet.go
```diff
@@ -41,12 +41,26 @@ type Wallet interface {
...
 	swapConfirmations(txHash *chainhash.Hash, vout uint32, contract []byte, startTime time.Time) (confs uint32, spent bool, err error)
-	getBlockHeader(blockHash *chainhash.Hash) (*blockHeader, error)
-	ownsAddress(addr btcutil.Address) (bool, error)
+	getBestBlockHeader() (*blockHeader, error)
+	ownsAddress(addr btcutil.Address) (bool, error) // this should probably just take a string
 	getWalletTransaction(txHash *chainhash.Hash) (*GetTransactionResult, error)
+
+	// WARNING: The following methods may not be implemented by all Wallet
+	// types. Care should be taken before using these in any method in which
+	// they are not already used.
```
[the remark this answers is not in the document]
chappjc: Ah that's nice.  The multiple levels of embedding don't bug me at all.  It's usually only too many interfaces that I have a though time with.

chappjc:
Debugging an issue with utxos being declared spent incorrectly.  Testing the refund path nicely at least.

chappjc:
I think I ironed out the coin locking/unlocking and spent check issues.  All seems to be working with LTC.

If you want to run an LTC electrumx server (as you would with regtest) with litecoind 0.21.2, you will need to use this branch that tolerates MWEB: https://github.com/chappjc/electrumx/tree/litecoin-mweb-tx

client/asset/btc/btc.go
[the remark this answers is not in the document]
buck54321: Yep. I think I had started a review some time ago and then lost track. My bad.

client/asset/btc/btc.go
```diff
@@ -192,6 +195,13 @@ var (
 		ConfigOpts:  append(spvOpts, commonOpts...),
 		Seeded:      true,
 	}
+	electrumWalletDefinition = &asset.WalletDefinition{
+		Type:        walletTypeElectrum,
+		Tab:         "Electrum (external)",
+		Description: "Use an external Electrum Wallet",
+		// json: DefaultConfigPath: filepath.Join(btcutil.AppDataDir("electrum", false), "config"), // e.g. ~/.electrum/config
+		ConfigOpts: append(rpcOpts, commonOpts...),
```
buck54321, line 203: The `Description` field of the `rpcOpts` is not accurate for electrum.

buck54321 (commented):
I wrote some harness scripts. https://github.com/buck54321/dcrdex/compare/351c9898cd738567451d84f8d89942802cfcf653...buck54321:dcrdex:electrum-harness-scripts

Can we get a flag for the core simnet tests?

JoeGruffins:
I guess the wallet name needs to be "default"? I had named one "trading" and get `code 1: "wallet not loaded"`

JoeGruffins (approved):
I see in the logs often:
```
2022-07-11 04:55:14.901 [INF] CORE[ltc]: Sent split transaction ef8c731dabb270a56ff01af7aa21a1aee5c9b8b26cc9455d77759edad8acf7f7 to accommodate swap of size 0.309 ltc + fees = 0.3090459
2022-07-11 04:55:14.902 [ERR] CORE[ltc]: error locking unspent outputs: failed to lock some utxos
2022-07-11 04:55:19.552 [DBG] CORE: notify: |POKE| (order) Order placed - buying 0.000300000 ETH, rate = 1030 (c97e3faa) - Order: c97e3faa826a58854a0e42267f817c3cfe425a6358efabbe4511d8ca4ceb992a
```

I see lots of connection errors to electrum, but may be because I'm using the one server.

This may be normal, but see this in the electrum logs:
```
Exception: Transaction not in wallet.
E | daemon.CommandsServer | internal error while executing RPC
Traceback (most recent call last):
  File "/tmp/.mount_electr7K3LjB/usr/lib/python3.9/site-packages/electrum_ltc/daemon.py", line 237, in handle
    response['result'] = await f(*params)
  File "/tmp/.mount_electr7K3LjB/usr/lib/python3.9/site-packages/electrum_ltc/commands.py", line 151, in func_wrapper
    return await func(*args, **kwargs)
  File "/tmp/.mount_electr7K3LjB/usr/lib/python3.9/site-packages/electrum_ltc/commands.py", line 1020, in get_tx_status
    raise Exception("Transaction not in wallet.")
```

Have tested on btc simnet and ltc testnet and can't find any problems. Working great.

chappjc:
> 2022-07-11 04:55:14.901 [INF] CORE[ltc]: Sent split transaction ef8c731dabb270a56ff01af7aa21a1aee5c9b8b26cc9455d77759edad8acf7f7 to accommodate swap of size 0.309 ltc + fees = 0.3090459
> 2022-07-11 04:55:14.902 [ERR] CORE[ltc]: error locking unspent outputs: failed to lock some utxos

Ah good catch @JoeGruffins.  This is some subtlety with split txns I did not handle.  Will come back to this after reviews.

client/core/simnet_trade.go
```diff
@@ -933,7 +954,12 @@ func testResendPendingRequests(s *simulationTest) error {
 	// has been spent but the spending tx is still in mempool. This
 	// will cause the txout to be included in the wallet's locked
 	// balance, causing a higher than actual balance report.
-	time.Sleep(4 * sleepFactor * time.Second)
+	sleep := 4 * time.Second
+	if s.client1.BaseWalletType == WTElectrum || s.client1.QuoteWalletType == WTElectrum ||
+		s.client2.BaseWalletType == WTElectrum || s.client2.QuoteWalletType == WTElectrum {
```
buck54321, line 959: Might as well make these two lines (4 repeats) a method on `simulationTest`.

```Go
func (s *simulationTest) anyElectrum() bool {
	return s.client1.BaseWalletType == WTElectrum || s.client1.QuoteWalletType == WTElectrum ||
		s.client2.BaseWalletType == WTElectrum || s.client2.QuoteWalletType == WTElectrum
}
```

Or even do all 5 lines.

```Go
func (s *simulationTest) waitALittleBit() {
	sleep := 4 * time.Second
	if s.client1.BaseWalletType == WTElectrum || s.client1.QuoteWalletType == WTElectrum ||
		s.client2.BaseWalletType == WTElectrum || s.client2.QuoteWalletType == WTElectrum {
		sleep = 7 * time.Second
	}
	time.Sleep(sleep * sleepFactor)
}
```

client/asset/btc/electrum_client.go
```diff
@@ -0,0 +1,1298 @@
...
+func (ew *electrumWallet) balances() (*GetBalancesResult, error) {
+	eBal, err := ew.wallet.GetBalance()
+	if err != nil {
+		return nil, err
+	}
+	// NOTE: Nothing from the Electrum wallet's response indicates trusted vs.
+	// untrusted. To allow unconfirmed coins to be spent, we treat both
+	// confirmed and unconfirmed as trusted. This is like dogecoind's handling
+	// of balance. TODO: listunspent -> checkWalletTx(txid) -> for each
+	// input, checkWalletTx(prevout) and ismine(addr)
```
buck54321, line 600: > TODO: listunspent -> checkWalletTx(txid) -> for each input, checkWalletTx(prevout) and ismine(addr)

Seems like you're checking that we actually own the utxos?

client/asset/btc/electrum_client.go
```diff
@@ -0,0 +1,1298 @@
...
+	for _, op := range ops {
+		opMap[op.pt] = struct{}{}
+	}
+	// For the ones that appear in listunspent, use (un)freeze_utxo also.
+unspents:
+	for _, utxo := range eUnspent {
+		for op := range opMap {
+			if op.vout == utxo.PrevOutIdx && op.txHash.String() == utxo.PrevOutHash {
+				// FreezeAddress and UnfreezeAddress do not error when called
+				// repeatedly for the same address.
```
buck54321, line 698: Outdated

client/asset/btc/electrum_client.go
```diff
@@ -0,0 +1,1298 @@
...
+		return nil, err
+	}
+	wif, err := btcutil.DecodeWIF(wifStr)
+	if err != nil {
+		return nil, err
+	} // wif.PrivKey is the result
+
+	// Sanity check that PrivKey corresponds to the pubkey(hash).
+	var pkh []byte
+	switch addrT := addrDec.(type) { // maybe switch on interfaces like Hash160() and PubKey() instead of concrete types
```
buck54321, line 826: Can remove `maybe` comment, it appears.

client/asset/btc/electrum_test.go
```diff
@@ -0,0 +1,165 @@
+// This code is available on the terms of the project LICENSE.md file,
+// also available online at https://blueoakcouncil.org/license/1.0.0.
+
+//go:build electrumlive
```
buck54321, line 4: Should you add `&& !electrumlive` to the top of **btc_test.go**?

client/asset/btc/electrum_client_test.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/txscript"
+)
+
+func Test_electrumWallet(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	const walletPass = "walletpass" // set me
+	ewc := electrum.NewWalletClient("user", "pass", "http://127.0.0.1:5678")
```
buck54321, line 29: Should this port should be `16789` now?

---
When using on simnet, this test is failing with `no headers retrieved`.

client/asset/btc/electrum_test.go
```diff
@@ -0,0 +1,165 @@
...
+		},
+		PeersChange: func(num uint32) {
+			t.Log("peer count: ", num)
+		},
+	}
+	cfg := &BTCCloneCFG{
+		WalletCFG:           walletCfg,
+		Symbol:              "ltc",
+		Logger:              tLogger,
+		ChainParams:         dexltc.TestNet4Params,
```
buck54321, line 75: This confused me a little. Should this test go in **client/assets/ltc**? Either way is fine.

buck54321 (changes requested):
Testing well. Thanks for adding the LTC simnet harness. Pretty much ready, just some small stuff.

client/asset/btc/electrum_client.go
```diff
@@ -0,0 +1,1298 @@
...
+	if !info.Connected || info.Server == "" {
+		return errors.New("Electrum wallet has no server connections")
+	}
+
+	// Determine if segwit expectation is met. Request and decode an address,
+	// then compare with the segwit config field.
+	addr, err := ew.wallet.GetUnusedAddress()
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
chappjc: I couldn't find the info in any RPCs, but fortunately `GetUnusedAddress` doesn't generate new ones like `CreateNewAddress`.

client/asset/btc/electrum/example/server/main.go
```diff
@@ -0,0 +1,159 @@
...
+	addrs := map[string]bool{ // addr => ssl
+		// "127.0.0.1:51002": true,
+		// "electrum.ltc.xurious.com:51002": true,
+		// "electrum-ltc.bysh.me:51002":     true,
+		// "electrum3.hodlister.co:50002": true,
+		// "node.degga.net:50002": true,
+		// "kittycp2gatrqhlwpmbczk5rblw62enrpo2rzwtkfrrr27hq435d4vid.onion:50001": false,
+		"kittycp2gatrqhlwpmbczk5rblw62enrpo2rzwtkfrrr27hq435d4vid.onion:50002": true,
+		// "hodlers.beer:50002": true,
+		// "electrum.neocrypto.io:50001": false,
```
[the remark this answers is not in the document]
chappjc: I'll uncomment them all.  For some reason I wanted the kitty one for a specific test.

client/asset/btc/electrum_client_test.go
chappjc: Ah, this little playground of a test is targeting testnet (`params: dexltc.TestNet4Params` below) because the txids are historical and you can investigate on the block explorer.  Although now that I revisit this, I had done this test on the "bad" non-mweb fork so I should regenerate the txids and and explorer links

client/asset/btc/electrum_client.go
chappjc: Right, this TODO is alluding to an "ownsInputs" kind of function like we have for `spvWallet`.  In this iteration it's just like DOGE where there's no distinction between trusted vs. untrusted.  I definitely want to try this before release.

client/asset/btc/electrum/wallet.go
```diff
@@ -0,0 +1,102 @@
...
+// result is marshaled into result if it is non-nil, otherwise the result is
+// discarded.
+func (ec *WalletClient) Call(method string, args interface{}, result interface{}) error {
+	reqMsg, err := prepareRequest(ec.nextID(), method, args)
+	if err != nil {
+		return err
+	}
+
+	bodyReader := bytes.NewReader(reqMsg)
+	ctx, cancel := context.WithTimeout(context.Background(), ec.Timeout)
```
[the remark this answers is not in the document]
chappjc: Bit the bullet and plumbed every exported method with a Context.  Welcome to 2022 I guess

chappjc:
testnet test updated for BTC instead of LTC, new swap used in findRedemption test:
- https://mempool.space/testnet/tx/1e99930f76638e3eddd79de94bf0ff574c7a400d1e6986cd61b3b5fd8212b1a3
- https://mempool.space/testnet/tx/4283c1fefa4898eb1bd2041547cd6361f8167dec004890b58ba1817914dc8541

client/asset/btc/electrum_client.go
```diff
@@ -350,6 +355,24 @@ func (ew *electrumWallet) connect(ctx context.Context, wg *sync.WaitGroup) error
...
+
+	// Check the RPC configuration.
+	var parsedCfg RPCWalletConfig // {RPCConfig, WalletConfig} - former may not change
+	err = config.Unmapify(cfg.Settings, &parsedCfg)
+	if err != nil {
+		return false, fmt.Errorf("error parsing rpc wallet config: %w", err)
+	}
+
+	// Changing RPC settings is not supported without restart.
+	return parsedCfg.RPCConfig != *ew.rpcCfg, nil
```
chappjc, line 373: maybe later

buck54321 (approved):

chappjc:
Ready to merge but holding off for a bit longer so https://github.com/decred/dcrdex/pull/1600 can go in first, possibly https://github.com/decred/dcrdex/pull/1638 too but the conflicts there should be minimal.

PR #1611, opened 2022-05-10
multi: differentiate send and withdraw

ukane-philemon:
This adds Send to asset.Wallet API which supports sending an amount without subtracting the fees from the amount.
- Add an `asset.Withdrawer`.
- Remove `asset.Sender`.
- Implement `asset.Withdrawer` for btc and dcr.
- Add `WalletTraitSweeper`.
- Add a send route to rpcserver.
- Change core's `Withdraw` method to  `Send`.
- Modify UI to support send.
- Return coins if send/withdraw fails.
- Remove `PayFee` Method in favor of the general-purpose `Send` method.

UI send btn
sendbtn

UI send/withdraw
senddcr
sendbitcoin
UI send only :
sendeth


client/asset/btc/btc.go
```diff
@@ -2782,6 +2780,19 @@ func (btc *baseWallet) Withdraw(address string, value, feeRate uint64) (asset.Co
...
 }
 
+// Send sends the exact value to the specified address. This is different from
+// Withdraw, which subtracts the tx fees from the amount sent. feeRate is in
+// units of sats/byte.
+// Send satisfies asset.Sender.
+func (btc *baseWallet) Send(address string, value, feeRate uint64) (asset.Coin, error) {
+	txHash, vout, sent, err := btc.send(address, value, btc.feeRateWithFallback(feeRate), false)
+	if err != nil {
+		btc.log.Errorf("Send error - address = '%s', amount = %s: %v", address, amount(value), err)
```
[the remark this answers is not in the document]
ukane-philemon: @JoeGruffins thanks for reviewing. This will log once, but should probably not log since the error is UI facing.

client/core/core.go
```diff
@@ -3920,8 +3920,22 @@ func (c *Core) Withdraw(pw []byte, assetID uint32, value uint64, address string)
...
+		if sender, isSender := wallet.Wallet.(asset.Sender); isSender {
+			coin, err = sender.Send(address, value, feeSuggestion)
+		} else {
+			return nil, fmt.Errorf("wallet does not support sending")
+		}
+	} else {
+		if withdrawer, isWithdrawer := wallet.Wallet.(asset.Withdrawer); isWithdrawer {
+			coin, err = withdrawer.Withdraw(address, value, feeSuggestion)
+		} else {
+			return nil, fmt.Errorf("wallet does not support withdrawing")
```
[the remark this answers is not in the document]
ukane-philemon: > That's an interesting error for a user to see. wallet does not support withdrawing. forced HODL

`wallet does not support subtracting network fee from withdraw amount.` sounds better?

> I kinda feel like don't need the sender trait if all the wallets can do that anyway.

the sender trait is used when displaying the withdraw form.

client/asset/dcr/dcr.go
```diff
@@ -2534,9 +2536,9 @@ func (dcr *ExchangeWallet) PayFee(address string, regFee, feeRate uint64) (asset
 	if err != nil {
 		return nil, err
 	}
-	// TODO: Evaluate SendToAddress and how it deals with the change output
+	// TODO: Evaluate SendToAddress rpc and how it deals with the change output
 	// address index to see if it can be used here instead.
```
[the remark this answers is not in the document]
ukane-philemon: > Does this solve the TODO?

It doesn't.

client/core/core.go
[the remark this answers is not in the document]
JoeGruffins: I think can base both messages on whether or not `isWithdrawer`? If all are senders.

client/asset/dcr/dcr.go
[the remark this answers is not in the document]
JoeGruffins: I see a new TODO: `TODO: Just use the sendtoaddress rpc since dcrwallet respects locked utxos.` is that talking about the same thing?

buck54321 (commented):
On a cursory scan, this looks pretty good. I'd like to offer some tweaks to the APIs. On the front end, I think we should just go with the **send** terminology, with an option to **subtract fees from amount sent**. Similary, **core** can just have `(*Core).Send`, with a `subtract bool` argument. 

Wallet APIs look about right, but now that I'm looking at it, I'm tempted to say that `Send` should remain a required part of the `asset.Wallet` API. Every asset should support that.

client/asset/dcr/dcr.go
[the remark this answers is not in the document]
ukane-philemon: > I see a new TODO: `TODO: Just use the sendtoaddress rpc since dcrwallet respects locked utxos.` is that talking about the same thing?

Don't think so. This todo is about using the sendToAddress rpc instead of creating the tx and then broadcasting.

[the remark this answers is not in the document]
chappjc: I think you guys are talking about the same thing now.

But no, the TODO is not resolved here.  There's still the possibility of just using the RPC to author the transaction for us.  I honestly don't remember what I was getting at with the "change output address index", but I think I made that comment.

ukane-philemon:
> I'm tempted to say that Send should remain a required part of the asset.Wallet API. Every asset should support that.

@chappjc what do you think?

chappjc:
It's required to do PayFee, so yes, I don't see why an asset would not support that.
https://github.com/decred/dcrdex/pull/1607/commits/823bfb3c064b5d73b30636b8b77d76fe6bf0cc39#diff-1b07e494f4bdd99b2a617ac110fbadc081e5b6ae1abf23f64032bbee5ebebef1R3441-R3445

buck54321:
> It's required to do PayFee, so yes, I don't see why an asset would not support that.

In fact, I think we can get rid of `PayFee` in favor of the general-purpose `Send` method.

chappjc:
We definitely can.  It's just a wrapper as in that linked code, so I'd be OK removing PayFee.  I had just kept it to minimize disruption, but if we all agree...

ukane-philemon:
Updated the images to reflect the modified UI.

client/core/notification.go
```diff
@@ -197,19 +197,19 @@ func newFeePaymentNoteWithConfirmations(topic Topic, subject, details string, se
...
+// SendNote is a notification regarding a requested send or withdraw.
+type SendNote struct {
 	db.Notification
 }
 
 const (
-	TopicWithdrawError Topic = "WithdrawError"
-	TopicWithdrawSend  Topic = "WithdrawSend"
+	TopicSendError  Topic = "SendError"
+	TopicSendSucess Topic = "SendSuccess"
```
LasTshaMAN, line 207: typo: TopicSendSucess -> TopicSendSuccess

client/core/core.go
```diff
@@ -4058,16 +4060,26 @@ func (c *Core) Withdraw(pw []byte, assetID uint32, value uint64, address string)
...
+	var coin asset.Coin
 	feeSuggestion := c.feeSuggestionAny(assetID)
-	coin, err := wallet.Withdraw(address, value, feeSuggestion)
+	if !subtract {
+		coin, err = wallet.Wallet.Send(address, value, feeSuggestion)
+	} else {
+		if withdrawer, isWithdrawer := wallet.Wallet.(asset.Withdrawer); isWithdrawer {
+			coin, err = withdrawer.Withdraw(address, value, feeSuggestion)
+		} else {
+			return nil, fmt.Errorf("wallet does not support subtracting network fee from withdraw amount")
```
LasTshaMAN, line 4071: this return prevents the notification below from being sent, is this intentional ? I don't know who exactly expects what notifications, 
but I think it makes sense to send error notification here as well, since other errors are notified on as well

```go
subject, details := c.formatDetails(TopicSendError, unbip(assetID), err)
c.notify(newSendNote(TopicSendError, subject, details, db.ErrorLevel))
```

LasTshaMAN (commented):
Mostly nits, don't mean to be very pedantic, just pointing out things that caught my eye

feel free to ignore, if comments seem too spammy

client/asset/btc/btc.go
```diff
@@ -3429,33 +3430,33 @@ func (btc *baseWallet) NewAddress() (string, error) {
...
 
 // Withdraw withdraws funds to the specified address. Fees are subtracted from
-// the value. feeRate is in units of atoms/byte.
+// the value. feeRate is in units of sats/byte.
+// Withdraw satisfies asset.Withdrawer.
 func (btc *baseWallet) Withdraw(address string, value, feeRate uint64) (asset.Coin, error) {
 	txHash, vout, sent, err := btc.send(address, value, btc.feeRateWithFallback(feeRate), true)
 	if err != nil {
-		btc.log.Errorf("Withdraw error - address = '%s', amount = %s: %v", address, amount(value), err)
+		return nil, err
```
JoeGruffins, line 3449: nit: A custom error return is still good, just without capitalization.

client/asset/interface.go
```diff
@@ -59,6 +61,18 @@ func (wt WalletTrait) IsRecoverer() bool {
...
+// which indicates the presence of a Withdraw method.
+func (wt WalletTrait) IsWithdrawer() bool {
+	return wt&WalletTraitWithdrawer != 0
+}
+
+// IsSweeper test if the WalletTrait has the WalletTraitSweeper bit set, which
+// indicates the presence of a Sweep method.
+func (wt WalletTrait) IsSweeper() bool {
+	return wt&WalletTraitSweeper != 0
+}
```
JoeGruffins, line 74: I still think the Sweeper interface can be removed as the purpose of making it was to deal with the issue that this pr solves. All `Withdrawer`s are also `Sweepers`.

ukane-philemon:
@chappjc do you think we would need translation for the new send success/error message before this is merged?

chappjc:
> @chappjc do you think we would need translation for the new send success/error message before this is merged?

No, we usually don't let that hold up development.

client/asset/interface.go
ukane-philemon: > I still think the Sweeper interface can be removed as the purpose of making it was to deal with the issue that this pr solves. All `Withdrawer`s are also `Sweepers`.


@chappjc, what do you think?

chappjc: `Sweeper` is an optional interface.  It does no harm to have it, and I'm pretty sure I would eventually add it back if it's not added now.

The method is more convenient than `Balance` check -> `Withdraw` -> check error in case balance was too low -> back to `Balance` check.

We're not talking about using it in the frontend in this PR are we?  I'm pretty sure the frontend demands a **UX** that is more in line with `Withdraw(specificAmount)`.  But I would absolutely use a Go `Sweep` method if I were writing automation logic to regularly send whatever is presently available to send (without the fuss of a logic race given active orders).

In many cases, it simplifies the backend handling if the API is (send whatever and tell me the amount sent) vs (spend exactly X and fail if I cannot, possibly leaving extra funds if more landed while I was studying the UI that displayed some available balance).  e.g. a `sweepaccount` RPC or special amount values like `"!"` for electrum can avoid the complex dexc application logic of picking utxos and dealing with the "subtract from amount" semantics required.

JoeGruffins: Thanks for the explanation.

JoeGruffins (commented):
Working well except there seems to be a little bug when withdrawing. If you try to withdraw the max amount from eth, it will error. If you change the value to something lower, it will continue to error. It only lets you send if you refresh the page.

This screen shot is after trying to withdraw full amt and then changing and clicking send:

image

client/asset/dcr/dcr.go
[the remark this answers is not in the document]
buck54321: Probably talking about randomization of the change output position. I think dcrwallet randomizes it by default, and we assume it's at `0` below.

client/webserver/locales/en-us.go
```diff
@@ -106,134 +106,135 @@ var EnUS = map[string]string{
...
+	"Withdraw":                               "Withdraw",
+	"Send":                                   "Send",
+	"Deposit":                                "Deposit",
+	"Lock":                                   "Lock",
+	"New Deposit Address":                    "New Deposit Address",
+	"Address":                                "Address",
+	"Amount":                                 "Amount",
+	"Authorize send with your app password.": "Authorize send with your app password.",
+	"Reconfigure":                            "Reconfigure",
+	"pw_change_instructions":                 "Changing the password below does not change the password for your wallet software. Use this form to update the DEX client after you have changed your password with the wallet software directly.",
```
buck54321, line 155: I'm not sure what going on with the formatting here. This new formatting seems to make sense, but if I'm on master and I run `gofmt`, this isn't what I get.

Did you make changes to this file? If you added entries, please put them at the bottom. If you updated existing entries, see if you can do it without changing the indentation everywhere.

client/asset/interface.go
ukane-philemon: > We're not talking about using it in the frontend in this PR are we? I'm pretty sure the frontend demands a UX that is more in line with Withdraw(specificAmount).

This can come in a separate PR? the `sweep` feature should fit the current UX/UI.

client/webserver/locales/en-us.go
ukane-philemon: Just edited "Authorize the withdraw" to "Authorize send" which is the longest string on that map.

client/asset/interface.go
ukane-philemon: >  the sweep feature should fit the current UX/UI.

Like a button on the send form? the current display might not be the total balance (assuming the user receives extra funds).

chappjc: Let's not worry about the frontend with respect to sweep right now.  Let's just address the send/withdraw issues you're after.
Sweeper can stay, but no need to implement or use any of it.  Maybe we'll get rid of it, but I don't think we should.

Maybe I'm in the minority, but a sweep feature accessible via a sweep RPC or a Sweep Go method is a very helpful thing to have even if that's not how we're designing our browser frontend.  It does no harm to leave it.

client/core/core.go
```diff
@@ -4039,16 +4039,18 @@ func (c *Core) feeSuggestion(dc *dexConnection, assetID uint32) (feeSuggestion u
 	return dc.fetchFeeRate(assetID)
 }
 
-// Withdraw initiates a withdraw from an exchange wallet. The client password
-// must be provided as an additional verification.
-func (c *Core) Withdraw(pw []byte, assetID uint32, value uint64, address string) (asset.Coin, error) {
```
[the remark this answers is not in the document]
ukane-philemon: Does this mean `core.Withdraw` should send (subtract `false`) for eth to maintain existing behavior or error?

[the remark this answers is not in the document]
chappjc: Oh right.  The answer would be, just allow it to error.  However, I think I'm changing my mind on this since Withdraw is currently unsatisfiable for ETH.  Legacy consumers that were using Withdraw for UTXO assets can just switch to Send.

Nevermind then.

client/webserver/webserver.go
```diff
@@ -342,7 +342,7 @@ func New(cfg *Config) (*WebServer, error) {
 			apiAuth.Post("/walletsettings", s.apiWalletSettings)
 			apiAuth.Post("/orders", s.apiOrders)
 			apiAuth.Post("/order", s.apiOrder)
-			apiAuth.Post("/withdraw", s.apiWithdraw)
+			apiAuth.Post("/send", s.apiSend)
```
chappjc, line 345: I don't think there are any other http api users besides our browser frontend, but if there are, they'll get 404.  It's fine I guess, but generally I'd keep the `/withdraw` route and have it wrap some code shared with `apiSend`.  Then we'd mark it deprecated and make that clear in the release notes.

client/core/core.go
```diff
@@ -4033,16 +4033,18 @@ func (c *Core) feeSuggestion(dc *dexConnection, assetID uint32) (feeSuggestion u
...
+// Send initiates either send or withdraw from an exchange wallet. if subtract
+// is true, fees are subtracted from the value else fees are taken from the
+// exchange wallet. The client password must be provided as an additional
+// verification.
+func (c *Core) Send(pw []byte, assetID uint32, value uint64, address string, subtract bool) (asset.Coin, error) {
```
chappjc, line 4040: We need to first deprecate `Core`'s `Withdraw` before removing it.  There are Go consumers that use it, but I was less certain about the http api method.  It's trivial for `Withdraw` to call `Send` with `subtract=true`.  The comment for `Withdraw` should include `// Deprecated: ...` and we will note this in the release notes.

I don't mind killing off the method from the `client/asset.Wallet` interface, but we should be a bit more cautious with breaking changes to `Core`.

client/core/locale_ntfn.go
```diff
@@ -346,12 +346,12 @@ var ptBR = map[Topic]*translation{
 		template: "Conectado com carteira para completar o registro em %s, mas falha ao destrancar: %v",
 	},
 	// [ticker, error]
-	TopicWithdrawError: {
+	TopicSendError: {
 		subject:  "Erro Retirada",
 		template: "Erro encontrado durante retirada de %s: %v",
```
chappjc, line 351: We don't have much of a system for this, but maybe comment with `// RETRANSLATE`.  If the message were completely changed, we'd just go back to English, but I actually think we're better off keeping the previous translations even if they still say "withdraw" because it still gets the point across and isn't technically incorrect.

client/core/locale_ntfn.go
```diff
@@ -60,14 +60,14 @@ var enUS = map[Topic]*translation{
...
+		subject:  "Send error",
+		template: "Error encountered while sending %s: %v",
 	},
 	// [ticker, coin ID]
-	TopicWithdrawSend: {
-		subject:  "Withdraw sent",
-		template: "Withdraw of %s has completed successfully. Coin ID = %s",
+	TopicSendSuccess: {
+		subject:  "Send Successful",
+		template: "Sending %s has completed successfully. Coin ID = %s",
```
chappjc, line 70: For another day, but I don't quite understand why these are notifications at all.  User clicks and the HTTP response contains the success/error message, which can go on the form, not the notifications.   Strange...

I'm also not entirely sure we need to change these note type and topic variable names, or the string literals assigned to them.  I guess there's no special handling for `NoteTypeWithdraw = "withdraw"` (now `NoteTypeSend = "send"`) anywhere on the frontend, so I suppose it's fine.  I guess just check closely and keep the new names and strings, but it feels like unneeded churn since none of it is user-facing and it was all still clear an applicable despite the updated capability (optionally subtracting fees).

Perhaps @buck54321 has some thoughts about both of the above points.

client/asset/interface.go
```diff
@@ -358,10 +375,12 @@ type Recoverer interface {
...
-// Sender is a wallet that can send funds to an address, as opposed to
-// withdrawing a certain amount from the source wallet/account.
-type Sender interface {
-	Send(address string, value, feeSuggestion uint64) (Coin, error)
+// Withdrawer is a wallet that can withdraw a certain amount from the
+// source wallet/account.
+type Withdrawer interface {
+	// Withdraw withdraws funds to the specified address. Fees are subtracted
+	// from the value.
+	Withdraw(address string, value, feeSuggestion uint64) (Coin, error)
```
chappjc, line 383: Why back to `feeSuggestion` on these?  These methods used to do fee estimation internally and use the suggestion as a fallback, but they were changed to take the `feeRate` that they should just use, if not zero: https://github.com/decred/dcrdex/commit/e41ecdfe57045cd3bf6b22aa7a6054740f0dca05
I see the eth backend still said `feeSuggestion` on the impl, but I think that is just a minor deviation in the naming.

That was a somewhat contentious change, but that's what we did.  https://github.com/decred/dcrdex/pull/1419

chappjc (changes requested):

dex/testing/loadbot/go.sum
```diff
@@ -42,24 +42,14 @@ cloud.google.com/go/storage v1.8.0/go.mod h1:Wv1Oy7z6Yz3DshWRJFhqM/UCfaWIRTdp0RX
 cloud.google.com/go/storage v1.10.0/go.mod h1:FLPqc6j+Ki4BU591ie1oL6qBQGu2Bl/tZ9ullr3+Kg0=
 collectd.org v0.3.0/go.mod h1:A/8DzQBkF6abtvrT2j/AU/4tiBgJWYyh0y/oB/4MlWE=
 contrib.go.opencensus.io/exporter/stackdriver v0.13.4/go.mod h1:aXENhDJ1Y4lIg4EUaVTwzvYETVNZk10Pu26tevFKLUc=
-decred.org/cspp/v2 v2.0.0-20211122173608-ee00e4952d5f h1:7gQImobDtPhW0BpK1h3KSb6fXelpzEhWlHqmz1gvDes=
-decred.org/cspp/v2 v2.0.0-20211122173608-ee00e4952d5f/go.mod h1:USyJS44Kqxz2wT/VaNsf9iTAONegO/qKXRdLg1nvrWI=
-decred.org/dcrwallet/v2 v2.0.0-20211206163037-9537363becbb h1:vio6o+nzszBrdj2Yi3/gifDa5ocfy49A9SqYPlbUjXo=
-decred.org/dcrwallet/v2 v2.0.0-20211206163037-9537363becbb/go.mod h1:rbFJaCuXCfDhYoI5ZdeZr8TmF4A4Sb1zE7jQAwtaFMo=
+decred.org/cspp/v2 v2.0.0 h1:b4fZrElRufz30rYnBZ2shhC8AjNVTN4i6TMzDi+hk44=
```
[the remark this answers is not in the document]
ukane-philemon: Great, didn't notice he had taken care of this. Though unrelated, I noticed the need for a go.sum update when updating `m.Withdraw` method -> `m.Send`.

client/webserver/locales/en-us.go
```diff
@@ -106,134 +106,135 @@ var EnUS = map[string]string{
...
+	"reg_ssl_needed":                         "Looks like we don't have an SSL certificate for this DEX. Add the server's certificate to continue.",
+	"Dark Mode":                              "Dark Mode",
+	"Show pop-up notifications":              "Show pop-up notifications",
+	"Account ID":                             "Account ID",
+	"Export Account":                         "Export Account",
+	"simultaneous_servers_msg":               "The Decred DEX Client supports simultaneous use of any number of DEX servers.",
+	"Change App Password":                    "Change App Password",
+	"Build ID":                               "Build ID",
+	"Connect":                                "Connect",
+	"Withdraw":                               "Withdraw",
```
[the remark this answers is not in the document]
ukane-philemon: Yes.

dex/testing/loadbot/go.sum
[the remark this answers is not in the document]
chappjc: Yah, good catch.  I always forget to update go.sum and I think I'm the one who left it like this when updating go.mod most recently.

client/webserver/locales/en-us.go
```diff
@@ -106,134 +106,135 @@ var EnUS = map[string]string{
...
+	"Send":                                   "Send",
+	"Deposit":                                "Deposit",
+	"Lock":                                   "Lock",
+	"New Deposit Address":                    "New Deposit Address",
+	"Address":                                "Address",
+	"Amount":                                 "Amount",
+	"Authorize send with your app password.": "Authorize send with your app password.",
```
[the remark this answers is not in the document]
ukane-philemon: I think the diff is due to `go fmt ./... `. Looks like it formats differently from `gofmt -s`.

client/asset/interface.go
ukane-philemon: Thanks for referencing #1419 and `e41ecdf` didn't know `feeSuggestion` had special meaning.

client/core/core.go
itswisdomagain: Wouldn't any go code consuming `Core` fail to build if `Core` no longer has a `Withdraw` method? I suppose a rebuild would be necessary since the code will need to update the dcrdex module version after this is merged to access this change.

chappjc: Yes, that's what I'm getting at.  We shouldn't immediately remove a method without first deprecating it and marking it as such.

client/core/locale_ntfn.go
```diff
@@ -345,12 +345,12 @@ var ptBR = map[Topic]*translation{
 		subject:  "Erro ao Destravar Carteira",
 		template: "Conectado com carteira para completar o registro em %s, mas falha ao destrancar: %v",
 	},
-	// [ticker, error]
+	// [ticker, error], RETRANSLATE.
```
chappjc, line 348: Just a note to other reviewers.  This was my suggestion for keeping a reasonable although potentially outdated translation. https://github.com/decred/dcrdex/pull/1611/commits/fb983f188d305770368fc649391b3c6f65613eb1#r896098359

client/core/core.go
```diff
@@ -4034,15 +4034,24 @@ func (c *Core) feeSuggestion(dc *dexConnection, assetID uint32) (feeSuggestion u
 }
 
 // Withdraw initiates a withdraw from an exchange wallet. The client password
-// must be provided as an additional verification.
+// must be provided as an additional verification.This method is DEPRECATED. Use
```
JoeGruffins, line 4037: nit: Space before new sentence `verification.This method`

JoeGruffins (commented):
I still cant send with eth after clicking max amount until a refresh.
image

This may be on master too, but unable to withdraw all my decred.
image

Probably unrelated also on master and unrelated to these changes,  but this error does not have a proper explanation.
image

ukane-philemon:
> I still cant send with eth after clicking max amount until a refresh. 

oops, thought I already applied a fix. 
> This may be on master too, but unable to withdraw all my decred. 

Can't reproduce this on my end. please can you reproduce this?

> Probably also on master and unrelated to these changes, but this error does not have a proper explanation.

Yes, the error type returned has an `Unwrap()` method that returns its [errorKind](https://github.com/decred/dcrd/blob/master/txscript/stdaddr/error.go#L12-L14) `ErrUnsupportedAddress` string instead of an expected `error` type. related #1488.

ukane-philemon:
Improved address error message
addrErrdcr
addrErrbtc

client/core/locale_ntfn.go
buck54321: These are sent at level `Success`. I think the idea is that it's nice to have a notification for certain important actions.

chappjc: Yeah, that's true about the success ntfn.  I like to have the txid in the notification menu to be copied at any time.

About the error one though, we've been moving to have those on the form itself, so isn't that a bit of a relic?

chappjc:
> Yes, the error type returned has an `Unwrap()` method that returns its [errorKind](https://github.com/decred/dcrd/blob/master/txscript/stdaddr/error.go#L12-L14) `ErrUnsupportedAddress` string instead of an expected `error` type. related #1488.

Regarding that related issue, I think I want to try a combination of [my last comment there](https://github.com/decred/dcrdex/issues/1488#issuecomment-1153432115) and an expanded `standardError` struct to assist the frontend with providing various levels of error detail (https://github.com/decred/dcrdex/commit/4453e5d3cca8a0c2df46951f741126c62c6b38d3).

It's unfortunate that those stdaddr errors don't have a human friendly string, and also an illustration why fully unwrapping is not always the best answer.

I'm keen to try out some ideas for this as described, but the last commit from @ukane-philemon to just do `fmt.Errorf("invalid address: %s", address)` instead of wrapping the unreadable error is a good approach for now.

client/core/locale_ntfn.go
ukane-philemon: > we've been moving to have those on the form itself, so isn't that a bit of a relic?

Are we removing the error notification for withdraw/send? True, the form displays the error if any. But it looks like it would be nice to have withdraw/send errors in the notifications (at least they don't just disappear, so users have a persistent record of errors related to withdrawing/sending their funds.)

JoeGruffins (approved):
> please confirm we don't need a refresh to send eth after clicking send available.

Yes ok now.

I've seen a couple weird withdrawing dcr now. Probably one is due to the massive amount of utxo. Will look into it and make an issue.

client/core/locale_ntfn.go
chappjc: Leaving I guess.  Personally, I think it's bizarre for the error notification to show up in the tray _and_ on the form.  What's the point of a persistent record of a failed withdraw request if nothing was withdrawn.  The user will resolve the error and try again, and the success message can stick around so they can find the txid.

chappjc:
Conflicts, sorry.

chappjc (approved):
All working for me.  Tested with all supported assets with variations of send/withdraw/full on with the simnet harnesses:

image

Two things to consider for future:
- ETH sweep being impossible presently without guestimating or trial and error with the fees (https://github.com/decred/dcrdex/pull/1591)
- I couldn't believe this was possible, but I sent all of my simnet LTC to an address from my DOGE wallet.  Is this only possible for simnet addresses?  I sure hope so.  If not, we might consider some warnings on the send/withdraw form for these two assets.

JoeGruffins:
> I couldn't believe this was possible, but I sent all of my simnet LTC to an address from my DOGE wallet. Is this only possible for simnet addresses? I sure hope so. If not, we might consider some warnings on the send/withdraw form for these two assets.

Will also be an issue for eth tokens I guess.

JoeGruffins:
> recovered given it's the same key.

True... So would also be same Private key for utxo coins?

chappjc:
> True... So would also be same Private key for utxo coins?

I only meant ETH tokens - same chain and same keys, just different contract to interact with.

But in this case of DOGE and LTC, they were identical pkScripts, so yes I could have imported the private key to the other chain's wallet.

chappjc:
@ukane-philemon the locales file has conflicts now.  I suggest you squash first before rebasing so you don't have to repeatedly resolve at each commit in this current branch.  After pulling latest master and checking out this branch, I suggest: `git rebase -i $(git merge-base master HEAD)` and `f` al but the first commit, and when that is done then `git rebase master` so just a single commit needs to be resolved.

buck54321 (approved):

PR #1613, opened 2022-05-12
dex client desktop app with flutter

itswisdomagain:
Early stages, basic app working for Mac OS using the `client/core.Core` Go library for DEX functionality.


buck54321:
Recommend closing. The development and maintenance involved with this effort would be considerable. I appreciate the effort and the intention. Perhaps as a very-highly simplified interface, there's some potential. But the UI for a widget-based app (esp. a newer framework) would be difficult to keep up with. The amount of boilerplate required is also unattractive. 

Open to counter-arguments.

itswisdomagain:
I share your thoughts @buck54321. Although, I think introducing a desktop GUI app would also require considerable development and maintenance. Unless we're scrapping desktop GUI implementation, are there any alternatives you've come across that would require less effort?

PR #1616, opened 2022-05-15
dex,client: restore dex reconnect loop

chappjc:
This reverts e8fd8acab23fbb253b649a66a95a5a63d4be6b75, which broke the
reconnect loop that the dex.Connector implemented by
client/comms.(*wsConn) starts.

This also adds a call to Disconnect when the connector for a client
wallet returns a non-nil error so that the ConnectionMaster's On method
does not say true, which is what e8fd8ac had aimed to resolve.


martonp (approved):
This looks good. Let's merge this and I'll rebase on top of it?

Issue #1105, opened 2021-06-11
ws: Handle messages that are too large more gracefully.


JoeGruffins:
Sending a message that is too large will cause the ws to disconnect temporarily. It would be better to stop these and error before sending. It looks like from the logs that it is reading, so I guess the server is at fault? Perhaps sending messages in parts if too big is the better answer...

Can reproduce with the simnet harness by selling 200,000 or so dcr.

```
2021-06-11 11:38:19.610 [ERR] CORE[wss://127.0.0.1:17273/ws]: read error (websocket: close 1009 (message too big)), attempting reconnection
2021-06-11 11:38:19.610 [DBG] CORE: notify: |POKE| (conn) DEX disconnected - DEX at 127.0.0.1:17273 has disconnected
2021-06-11 11:38:19.610 [INF] CORE[wss://127.0.0.1:17273/ws]: Attempting to reconnect to wss://127.0.0.1:17273/ws...
2021-06-11 11:38:19.640 [DBG] CORE: notify: |POKE| (conn) DEX connected - DEX at 127.0.0.1:17273 has connected
```

chappjc:
Impressive.  The read limit for authenticated connections is 64 KiB, so you must have been funding an order with a shit ton of UTXOs.

https://github.com/decred/dcrdex/blob/db3df625eb736442a464dcc3aea424e8a8491a78/server/comms/link.go#L107-L109

Not sure the right resolution at this time though.

ukane-philemon:
> It looks like from the logs that it is reading, so I guess the server is at fault? Perhaps sending messages in parts if too big is the better answer...

Two possible fix for this includes
1. Increase read limit for authenticated connections.
2. Split large messages and send in parts.
3. Send an error message but don't disconnect websocket.

@chappjc which way to go?

chappjc:
(1) was done in https://github.com/decred/dcrdex/commit/fb7bb9da37a60f7cd6f4bea7707d11c31fd77610
Not sure about the other options.

ukane-philemon:
> (1) was done in fb7bb9d

This solves error with large messages for now(until the next larger message :) )

Adding other options should wait?

chappjc:
You may propose other solutions.  Not disconnecting sounds like a reasonable first step though.

ukane-philemon:
> Not disconnecting sounds like a reasonable first step though, unless it gets too messy.

seems increasing the read limit is our best shot for now since it immediately reconnects if ws was disconnected due to read error.

[...]

PR #1620, opened 2022-05-17
dex/encode: allow data longer than 65535 bytes

chappjc:
Relating to the issue hit in https://github.com/decred/dcrdex/issues/1105#issuecomment-1128595137, where the counter tx data was too large on account of the tx having many funding coins.

```
Retrofit the blob encoding and decoding to allow byte slices up to
0x00fe_ffff (16711679) long.  The decoding scheme detects a uint32
length when the top two bytes decoded as big endian uint16 indicate a
value less than 255.
```

This is a bit of a cludge, but it seems to be working, so I figure we could discuss it.

### DIFF
--- a/dex/encode/encode.go
+++ b/dex/encode/encode.go
@@ -10,19 +10,21 @@ import (
 	"encoding/binary"
 	"fmt"
 	"io"
+	"math"
 	"os"
 	"time"
 )
 
 var (
-	// IntCoder is the DEX-wide integer byte-encoding order.
+	// IntCoder is the DEX-wide integer byte-encoding order. IntCoder must be
+	// BigEndian so that variable length data encodings work as intended.
 	IntCoder = binary.BigEndian
 	// A byte-slice representation of boolean false.
 	ByteFalse = []byte{0}
 	// A byte-slice representation of boolean true.
-	ByteTrue = []byte{1}
-	maxU16   = int(^uint16(0))
-	bEqual   = bytes.Equal
+	ByteTrue   = []byte{1}
+	bEqual     = bytes.Equal
+	maxDataLen = 0x00fe_ffff // top two bytes in big endian stop at 254, signalling 32-bit len
 )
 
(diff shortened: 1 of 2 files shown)


dex/encode/encode.go
```diff
@@ -125,7 +126,14 @@ func ExtractPushes(b []byte, preAlloc ...int) ([][]byte, error) {
 				return nil, fmt.Errorf("2 bytes not available for uint16 data length")
 			}
 			l = int(IntCoder.Uint16(b[:2]))
-			b = b[2:]
+			if l < 255 {
+				// This indicates it's really a uint32 capped at 0x00fe_ffff, and
+				// we are looking at the top two bytes.
+				l = int(IntCoder.Uint32(b[:4]))
```
JoeGruffins, line 132: This works because any length under 255 would have just been encoded as the first byte correct?

There's a check for if `len(b) < 2` before this to make sure there's enough data left for the two byte uint16. Maybe another check for the two more bytes of the uint32 is prudent?

Also the error message "2 bytes not available for uint16 data length" may be wrong because may be uint32

JoeGruffins (approved):
Looks good.

martonp (approved):
LGTM

dex/encode/encode.go
```diff
@@ -172,12 +184,25 @@ type BuildyBytes []byte
...
+	if l >= 0xff {
+		if l > maxDataLen {
+			panic("cannot use addData for pushes > 16711679 bytes")
+		}
+		var i []byte
+		if l > math.MaxUint16 { // not >= since that is historically in 2 bytes
+			// We are retrofitting for data longer than 65535 bytes, so we
+			// cannot switch to uint32 at 65535 itself since it is possible
+			// there is data of exactly that length already stored using just
+			// two bytes to encode the length. Thus, the decoder should inspect
```
buck54321, line 196: Ah crap. It it wasn't for this, we could just us the varint stuff from wire.

buck54321 (approved):

Issue #1241, opened 2021-10-19
app: unhandleable spot price notes prior to login


chappjc:
When dexc is started up and automatically connects to any configure DEX servers, it starts sending spot price notes.  But when the pages have not logged in yet:

![image](https://user-images.githubusercontent.com/9373513/137991313-b447f9de-7a5a-4e93-b68a-00f30062bfac.png)

Each note results in this:

![image](https://user-images.githubusercontent.com/9373513/137991130-83b555b1-144c-4d0d-b5e7-5987488ed184.png)

https://github.com/decred/dcrdex/blob/f7bb867dda5193987280848224796b26220405c2/client/webserver/site/src/js/app.js#L521-L524

chappjc:
Bump as I think we should fix for next patch.

On a somewhat related note, I'm not concerned about the logged `[ERR] CORE: subPriceFeed: unable to fetch market overview: rpc error: error code 2: unknown route` since that will go away when server updates, but this JS error will not.

chappjc:
bump in case anyone is looking for work.  This is still a console error with 0.4 release

It can also happen at some point in the registration sequence

![image](https://user-images.githubusercontent.com/9373513/152226116-9daf6e51-8879-4134-8305-c9bb5e43f3bd.png)

chappjc:
For another day

chappjc:
Resolved by https://github.com/decred/dcrdex/pull/1624

PR #1624, opened 2022-05-25
app: ignore balance and spot notifications during login

buck54321:
Prevents a couple of the errors we see when logging in.


chappjc:
Resolves https://github.com/decred/dcrdex/issues/1241

chappjc (approved):
I can see the spot notes in the browser's WS monitoring, but no longer console errors.

PR #1626, opened 2022-05-25
eth: update solc and abigen workflow

chappjc:
Wtih go-ethereum 1.10.18, abigen no longer has solidity compiler support, and we must use solc directly.


martonp (approved):
Works well.

JoeGruffins (approved):
I have changes when doing `./updatecontract.sh 0`

I'm using:
```
$ solc --version
solc, the solidity compiler commandline interface
Version: 0.8.6+commit.11564f7e.Linux.g++
$ abigen --version
abigen version 1.10.2-stable-97d11b01
```

Maybe can add something about the versions in the README? Also we have a testnet address now you could add.

chappjc:
> I have changes when doing

Changes other than what's in the PR?  To `Bin` or other parts of contract.go?

```
$  solc --version
solc, the solidity compiler commandline interface
Version: 0.8.6+commit.11564f7e.Linux.g++
$  abigen --version
abigen version 1.10.18-stable
```

Running an old abigen?

JoeGruffins:
> Running an old abigen?

Yes. No changes after updating to `1.10.18-stable`

buck54321 (approved):

PR #1629, opened 2022-05-26
server/swap: metered block-triggered conf and penalty checks

buck54321:
```
Ethereum has very short block times. And just for conservation of
resources, tightly-grouped blocks need not trigger tightly-grouped
conf and penalty checks. This adds a minimum delay window to block
notifications. If blocks comes farther than the minimum delay apart,
nothing changes from the current behavior. If 1 or more blocks
arrives within the delay window of another, they will be collected into a
single block notification that occurs once the min delay has passed.

The newly implemented SignalMeterer is only used in the swapper for
block notifications for now, but might also be useful for the client.
```

### DIFF
--- a/dex/meter/meterer.go
+++ b/dex/meter/meterer.go
@@ -1,0 +1,66 @@
+// This code is available on the terms of the project LICENSE.md file,
+// also available online at https://blueoakcouncil.org/license/1.0.0.
+
+package meter
+
+import (
+	"context"
+	"time"
+)
+
+// DelayedRelay creates a simple error signal pipeline that delays and
+// aggregates the relaying of nil errors. Non-nil errors received on the in
+// channel are immediately send on the out channel without delay. If a nil error
+// arrives within minDelay of the previous one, it will be scheduled for later
+// to respect the configured delay. If multiple arrive within minDelay, they
+// will be grouped into a single delayed signal.
+func DelayedRelay(ctx context.Context, minDelay time.Duration, n int) (out <-chan error, in chan<- error) {
+	inC := make(chan error, n)
+	outC := make(chan error, n)
+
+	go func() {
+		defer close(outC) // signal shutdown to the consumer
+
+		var last time.Time
+		var scheduled *time.Timer
+		now := make(chan struct{})
+
+		for {
+			select {
+			case err := <-inC:
+				if err != nil {
+					outC <- err
+					continue
+				}
+
+				if scheduled != nil {
+					continue
+				}
+
+				if time.Since(last) > minDelay {
+					outC <- nil
+					last = time.Now().UTC()
+					continue
+				}
+
+				delay := time.Until(last.Add(minDelay))
+				scheduled = time.AfterFunc(delay, func() {
+					now <- struct{}{}
+				})
+
+			case <-now:
+				outC <- nil
+				scheduled = nil
+				last = time.Now().UTC()
+
+			case <-ctx.Done():
+				if scheduled != nil && !scheduled.Stop() {
+					<-scheduled.C
+				}
+				return
+			}
+		}
+	}()
+
+	return outC, inC
+}
--- a/server/swap/swap.go
+++ b/server/swap/swap.go
@@ -14,6 +14,7 @@ import (
 	"decred.org/dcrdex/dex"
 	"decred.org/dcrdex/dex/calc"
 	"decred.org/dcrdex/dex/encode"
+	"decred.org/dcrdex/dex/meter"
 	"decred.org/dcrdex/dex/msgjson"
 	"decred.org/dcrdex/dex/order"
 	"decred.org/dcrdex/dex/wait"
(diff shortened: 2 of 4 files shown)


chappjc:
Cool.  I guess I should get https://github.com/decred/dcrdex/pull/1563 out of draft.  Lock contention can be just as bad as blocking the ws inHandler or backing up conf checks depending on where it is blocking.

dex/meter/meterer.go
```diff
@@ -0,0 +1,72 @@
...
+
+// Stop cancels any scheduled signals.
+func (s *SignalMeterer) Stop() {
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+	if s.scheduled != nil {
+		if !s.scheduled.Stop() {
+			<-s.scheduled.C
+		}
+	}
```
LasTshaMAN, line 71: maybe also need to `s.scheduled = nil` inside of this `if` (assuming `Stop` method is only meant to `cancels any scheduled signals` and not completely prevent further usage of `SignalMeterer`),

as the following condition will be hit, it seems:

image

server/swap/swap.go
```diff
@@ -34,6 +35,7 @@ var (
 	// txWaitExpiration is the longest the Swapper will wait for a coin waiter.
 	// This could be thought of as the maximum allowable backend latency.
 	txWaitExpiration = 2 * time.Minute
+	minBlockPeriod   = time.Second * 10
```
LasTshaMAN, line 38: might be useful to describe how/why this value was chosen

LasTshaMAN (commented):
Minor things

JoeGruffins (approved):

dex/meter/meterer.go
buck54321: > and not completely prevent further usage of `SignalMeterer`

I always just assumed this was a one-`Done` type.

server/swap/swap.go
LasTshaMAN: maybe just some explanation regarding why this delay is needed in the first place would be nice to have (idk, one can search through git history and find this PR with the answer as well),

can suggest something like:
```
// minBlockPeriod helps with limiting notification bursts when blocks are generated too fast (e.g. in Ethereum occasionally several blocks can are generated in a matter of several seconds).
```

dex/meter/meterer.go
```diff
@@ -0,0 +1,72 @@
...
+	"sync"
+	"time"
+)
+
+// SignalMeterer meters a simple signal to prevent tight grouping. The user
+// specifies a minimum delay (minDelay). If a signal arrives within minDelay of
+// the previous signal, that signal will be scheduled for later to respect the
+// configured delay. If multiple signals arrive within minDelay, they will be
+// grouped into a single delayed signal. Error signals are never delayed.
+type SignalMeterer struct {
```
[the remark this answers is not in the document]
JoeGruffins: Thanks. Understood.

dex/meter/meterer.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+}
+
+// Signal sends a signal, with a delay if appropriate.
+func (s *SignalMeterer) Signal(err error) {
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+	// Errors are passed immediately.
+	if err != nil {
+		go s.recv(err)
```
chappjc, line 39: I think you might as well replace `recv  func(error)` with `recv  chan error`, and doing `s.recv <- err` instead of `go s.recv(err)`.

The goroutine allocations and runtime overhead seem avoidable, and it feels tailored for a channel anyway.  I get the sense you're using goroutines as a way to use the runtime for effectively unlimited buffering, which could backfire in addition to being a bit inefficient.

Plus in `Swapper`, I think a channel is more natural since you've got a select loop going already.  illustration: https://github.com/decred/dcrdex/commit/124ec646226926a7b3a70ee290751ba90fb5fc8c

And as usual lately I'm looking at patterns that eliminate locks and fields with event loops and closure, so I'm toying with this alternate mechanism: https://github.com/decred/dcrdex/commit/817b80063fe33f491999155ae21c27ab148279c9
Just toying around, don't feel like you need to take either suggestion.

dex/meter/meterer.go
```diff
@@ -0,0 +1,72 @@
...
+		recv:     recv,
+	}
+}
+
+// Signal sends a signal, with a delay if appropriate.
+func (s *SignalMeterer) Signal(err error) {
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+	// Errors are passed immediately.
+	if err != nil {
```
chappjc, line 38: Can move the locking until after this first early return

server/swap/swap.go
```diff
@@ -780,6 +790,8 @@ func (s *Swapper) Run(ctx context.Context) {
 	bcastBlockTrigger := make(chan uint32, 32*len(s.coins))
 	scheduleInactionCheck := func(assetID uint32) {
 		time.AfterFunc(s.bTimeout, func() {
+			// TODO: This pattern would still send the block trigger half of the
+			// time if the ctxMaster is canceled.
```
chappjc, line 794: Good catch.  That's annoying.  Seems harmless, just not really intended.

server/swap/swap.go
chappjc: Agree a sentence would be good here.

As far as the choice of 10s goes, I suppose that's short enough that it would not be an annoying delay, but also long enough that checks shouldn't pile up too much. (it works for me)

chappjc (changes requested):
Cool little utility.  Overall seems like good logic and it works.

I don't totally love the goroutine use as a means to get unlimited buffering courtesy of the runtime, and I have some ideas there, but nothing that really has to change.

dex/meter/meterer.go
buck54321: https://github.com/decred/dcrdex/commit/817b80063fe33f491999155ae21c27ab148279c9 is beautiful 😍.

PR #1632, opened 2022-05-29
utility for running core harness tests

buck54321:
They're not all passing yet, but this makes it easier to run different combos for the **core** simnet harness tests. Still figuring out the failing tests, but the error is always `balance change not in expected range`. Could be the same on master.


client/core/simnet_trade.go
```diff
@@ -1719,99 +1688,126 @@ func (n *notificationReader) find(ctx context.Context, waitDuration time.Duratio
...
 
 	ord, err := client.core.Trade(client.appPass, tradeForm)
 	if err != nil {
 		return "", err
 	}
 
-	client.log("placed order %sing %s at %s (%s)", sellString(client.isSeller), qtyStr, rateStr, ord.ID[:8])
+	client.lastOrder = ord.ID
+
+	client.log.Infof("placed order %sing %s at %s (%s)", sellString(client.isSeller), qtyStr, rateStr, ord.ID[:8])
```
JoeGruffins, line 1722: Must be an old bug but log doesn't look right `2022-05-31 18:24:25.360 [INF] T[1]: placed order selling 20.00000000 dcr at 0.000150000 eth%!(EXTRA string=dcr) (7e764dabd4977779)`

JoeGruffins, line 1722: Must be an old bug but log doesn't look right `2022-05-31 18:24:25.360 [INF] T[1]: placed order selling 20.00000000 dcr at 0.000150000 eth%!(EXTRA string=dcr) (7e764dabd4977779)`

client/cmd/simnet-trade-tests/run
```diff
@@ -0,0 +1,45 @@
...
+    ./simnet-trade-tests --base1node trading1 --base2node trading2 --quote eth ${@:2}
+    ;;
+
+  list)
+    ./simnet-trade-tests --list
+    ;;
+
+  help)
+    ./simnet-trade-tests --help
+    ;;
```
JoeGruffins, line 40: nit: This help should be help for the run file, like maybe the options above, and list isn't so helpful.

JoeGruffins, line 40: nit: This help should be help for the run file, like maybe the options above, and list isn't so helpful.

JoeGruffins (commented):
Very interesting approach.

I see an error when runnin `ltcdcr`: `rawrequest (sendrawtransaction) error: -3: Amount is not a number or string`

I see the balance errors for `dcrdoge`. doge must use the gamma and delta nodes because it has no accounts so balance tests will always fail when mining blocks. https://github.com/decred/dcrdex/pull/1603/commits/9fd40540121d4f9f00680579d6c6feb2f6a8a66c

I also see balance errors for `dcreth`. Do you think it's because maybe you have included the fee? Eth's estimation or `swaps.FeeRate` is 200 gwei here. Looking up the actual for the tx I find `effectiveGasPrice: 2000000007,` which I think is 2 gwei per gas actual.

chappjc:
> I see an error when runnin `ltcdcr`: `rawrequest (sendrawtransaction) error: -3: Amount is not a number or string`

Are you using 0.21 in this test?  The 0.21 PR isn't merged yet

buck54321:
I'm going to go through and try to resolve some errors. I'm aware of a few. But I'm pretty sure they aren't all mine.

chappjc:
Just for reference, on `master` the main trade tests pass for me: `TestTradeSuccess`, `TestNoTakerSwap`, `TestNoMakerRedeem`, `TestNoTakerRedeem`, and `TestMakerGhostingAfterTakerRedeem`  :heavy_check_mark:
Not tested this PR yet.

JoeGruffins:
> Are you using 0.21 in this test? The 0.21 PR isn't merged yet

Oh yeah... Using 0.21.2, so that`s probably it. Apologies.

chappjc:
We should probably get that merged too though...  Just infuriating that ltc testnet is messed up still (explorers on wrong chain, etc.).  I had to make large changes to that PR to to complete deserialization of all the MW tx types though, so it kinda got bigger like the zcash PR.

client/core/simnet_trade.go
```diff
@@ -1836,21 +1832,28 @@ func (client *tClient) disableWallets() {
...
+		return float64(amt) / float64(s.conversionFactor(assetID))
 	}
 	panic(fmt.Sprintf("invalid call to fmtAmt with %v", anyAmt))
 }
+
+func init() {
+	if race {
+		sleepFactor = 3
+	}
+}
```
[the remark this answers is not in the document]
buck54321: Things just take longer with the race detector, so sleeps needed more time, it appears. But yes, it should work to `go build -tags lgpl,simnet_trade_test -race` and then run `./simnet-trade-tests`.

dex/testing/doge/harness.sh
```diff
@@ -277,11 +277,18 @@ tmux send-keys -t $SESSION:4 "./alpha generatetoaddress 400 ${ALPHA_MINING_ADDR}
 # Send beta some coin
 ################################################################################
 
+DELTA_ADDR=`./delta getnewaddress`
+echo "delta address = '${DELTA_MINING_ADDR}'"
+GAMMA_ADDR=`./gamma getnewaddress`
+echo "gamma address = '${GAMMA_MINING_ADDR}'"
```
chappjc, line 283: Do you mean to use the *mining* address in these echos?  looks like no

buck54321:
I think this is ready. If there's any tweaks to be made for Ethereum, I'll do it in #1622.

PR #1633, opened 2022-05-29
client/asset/dcr: native Decred SPV wallet

buck54321:
Use dcrwallet's `*wallet.Wallet` in a similar fashion to the `spvWallet` in **client/btc**.

This is ready for review, but I don't want to merge until this is rebased on top of #1632 and tests implemented.


client/asset/dcr/dcr.go
```diff
@@ -452,27 +536,35 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network)
 		return nil, err
 	}
 
-	// Set dcr.wallet using either the default rpcWallet or a custom wallet.
-	if cfg.Type == walletTypeDcrwRPC || cfg.Type == walletTypeLegacy {
-		dcr.wallet, err = newRPCWallet(walletCfg, chainParams, logger)
+	switch {
+	case cfg.Type == walletTypeDcrwRPC || cfg.Type == walletTypeLegacy:
```
chappjc, line 540:
```go
switch cfg.Type {
case walletTypeDcrwRPC, walletTypeLegacy:
...
case walletTypeSPV:
...
default:
...
```

chappjc (commented):
Awesome.  I just did a turbo scan through.  Very excited for this.  Looks straightforward based on the BTC SPV so I bet this can get in short order.

JoeGruffins (commented):
I was able to get the client into a totally unresponsive state at one point, but no idea if it was because of these changes.

Otherwise testing well.

Do you think a version somewhere needs to be upped for these changes? If I were to make an spv wallet and go back a version it would no longer be visible. I don't know what would happen if I then attempted to make another wallet and then went back to this diff.

The simnet tests are not working.

chappjc:
> Do you think a version somewhere needs to be upped for these changes? If I were to make an spv wallet and go back a version it would no longer be visible.

I think an asset version is a value coordinated between client and server as an indication of more fundamental changes like contract structure or block/tx serializations.

chappjc:
Not sure which way you are going to go for conflict resolution, but clearly the account names for the SPV wallet should stay RPC-only since the SPV wallet should manage the accounts even when we tackle mixing support for the native DCR wallet.
Reviewing as-is though.

client/asset/dcr/spv.go
```diff
@@ -0,0 +1,1009 @@
...
+	if len(txd.Debits) == len(txd.MsgTx.TxIn) {
+		var outputTotal dcrutil.Amount
+		for _, output := range txd.MsgTx.TxOut {
+			outputTotal += dcrutil.Amount(output.Value)
+		}
+		fee = debitTotal - outputTotal
+		negFeeF64 = (-fee).ToCoin()
+	}
+	ret.Amount = (creditTotal - debitTotal).ToCoin()
+	ret.Fee = negFeeF64
```
chappjc, line 687: **TLDR**: I'm generally uneasy about computing anything reliably from the `Credits` and `Debits` slices (namely `Fee` or `Amount`), but fortunately the full `GetTransactionResult` is mostly unsued, so can we consider paring down the interface to make it both easier to satisfy and more reliable?

----

Lately I have concluded we should avoid relying on both the `Credits` and `Debits` slices from `TxDetails` (or the `Amount` and `Fee` fields that we compute from them or get from `gettransaction` on the RPC side) because they are very often incomplete or incorrect.  (Both BTC and DCR)

As you noted above, `Fee can only be determined if every input is a debit`, and this is indeed trouble with mix transactions, just as an example.  Inspecting one of those in my wallet, the `"fee"` field is omitted, and `"amount"` is something negative like `"amount": -0.002399` which is presumably indicative of the net change to my wallet's balance.
The help for `gettransaction` hints at more general conditions `"fee": n.nnn,                     (numeric)         The total input value minus the total output value, or 0 if 'txid' is not a sent transaction`.  Mixed transactions are just one instance.

I chatted with @martonp about this here: https://github.com/decred/dcrdex/pull/1555#discussion_r871745203

To be clear, I don't think the issue is what we are doing with the output from `dcrWallet.TxDetails`, but that it has weak assurances about what it will include in `Credits` and `Debits`, plus the conditions are different between unconfirmed and confirmed txns.  In both `unminedTxDetails` and `minedTxDetails`, it just grabs what credits and debits it has, and looking through real wallet transactions they are often missing.

Anyway, for the purpose of tx fee, Marton did it the hard but reliable way by pulling prevout tnxs for the input values, and in the Electrum PR I've [pared down the `GetTransactionResult` type](https://github.com/decred/dcrdex/commit/2e857bdac9f61721a05e854b0a12efba924c4133#diff-8b451093a1f3c33c83eee1f754caa0125c399e7764c35ec57eb6b57a254be39d) because it's neigh impossible to completely satisfy.  I think this is OK since consumers generally just want (1) confs, (2) block info, and (3) raw tx hex.  Scanning through this PR, I think that's still the case.

client/asset/dcr/spv.go
```diff
@@ -0,0 +1,1009 @@
...
+	}
+
+	medianTime, err := w.medianTime(ctx, blockHash)
+	if err != nil {
+		return nil, err
+	}
+
+	_, tipHeight := w.MainChainTip(ctx)
+	if tipHeight < int32(hdr.Height) {
+		return nil, fmt.Errorf("sumpin's wrong with our tip")
```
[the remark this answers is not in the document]
buck54321: I'm conflicted on this. I'm going to fiddle with the directory structure though.

client/asset/dcr/spv.go
```diff
@@ -0,0 +1,1009 @@
...
+		return fmt.Errorf("checkCreateDir error: %w", err)
+	}
+
+	// At this point it is asserted that there is no existing database file, and
+	// deleting anything won't destroy a wallet in use.  Defer a function that
+	// attempts to remove any written database file if this function errors.
+	defer func() {
+		if err != nil {
+			_ = os.Remove(dbPath)
+		}
```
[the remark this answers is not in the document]
buck54321: I'm conflicted on this. I'm going to fiddle with the directory structure though.

client/asset/dcr/spv.go
```diff
@@ -0,0 +1,1009 @@
...
+	defer w.blockCache.Unlock()
+	for blockHash, cb := range w.blockCache.blocks {
+		if time.Since(cb.lastAccess) > time.Hour*2 {
+			delete(w.blockCache.blocks, blockHash)
+		}
+	}
+}
+
+func newSpvSyncer(w *wallet.Wallet, netDir string, connectPeers []string) *spv.Syncer {
+	addr := &net.TCPAddr{IP: net.ParseIP("::1"), Port: 0}
```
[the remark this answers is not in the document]
chappjc: This code mirrors what dcrwallet does internally, but interestingly this address seems to be unused by dcrwallet anyway.  If you trace into p2p/peering.go, and look for where `extaddr` is used, it seems to go nowhere. It gets passed to `(*LocalPeer).newMsgVersion`, but it is unused there, using `net.Conn.LocalAddr()` instead.

Almost as if `nil` would work.  Maybe we keep it like this with ipv6 to match dcrwallet, but make a comment because it's clearly confusing multiple people.

client/asset/dcr/spv.go
```diff
@@ -0,0 +1,1020 @@
...
+	blockCache blockCache
+
+	cancel context.CancelFunc
+	wg     sync.WaitGroup
+}
+
+var _ Wallet = (*spvWallet)(nil)
+var _ tipNotifier = (*spvWallet)(nil)
+
+func createSPVWallet(pw, seed []byte, dataDir string, log dex.Logger, extIdx, intIdx uint32, chainParams *chaincfg.Params) error {
```
chappjc, line 136: The `log` arg is unused, can remove.

client/asset/dcr/spv.go
```diff
@@ -0,0 +1,1020 @@
...
+		}
+	}
+
+	return nil
+}
+
+// walletExists returns whether a file exists at the loader's database path.
+// This may return an error for unexpected I/O failures.
+func walletExists(dbDir string) (bool, error) {
+	exists, err := fileExists(filepath.Join(dbDir, walletDbName))
```
chappjc, line 941: You wrote `fileExists` right below so you know it always returns false for non-nil error.  Hence,
`return fileExists(filepath.Join(dbDir, walletDbName))`

client/asset/dcr/spv.go
```diff
@@ -0,0 +1,1020 @@
...
+			MsgBlock:   block,
+			lastAccess: time.Now(),
+		}
+		w.blockCache.blocks[blockHash] = cb
+	} else {
+		cached.lastAccess = time.Now()
+	}
+}
+
+// cachedBlock retreives the MsgBlock from the cache, if it's been cached, else
```
chappjc, line 833: retreives

client/asset/dcr/spv.go
```diff
@@ -0,0 +1,1020 @@
...
+	for _, cred := range txd.Credits {
+		creditTotal += cred.Amount
+	}
+	// Fee can only be determined if every input is a debit.
+	if len(txd.Debits) == len(txd.MsgTx.TxIn) {
+		var outputTotal dcrutil.Amount
+		for _, output := range txd.MsgTx.TxOut {
+			outputTotal += dcrutil.Amount(output.Value)
+		}
+	}
```
chappjc, line 677: Can delete these 17 lines now.

client/asset/dcr/spv.go
```diff
@@ -0,0 +1,1020 @@
...
+// GetBlockHeader generates a *BlockHeader for the specified block hash. The
+// returned block header is a wire.BlockHeader with the addition of the block's
+// median time and other auxiliary information.
+func (w *spvWallet) GetBlockHeader(ctx context.Context, blockHash *chainhash.Hash) (*BlockHeader, error) {
+	hdr, err := w.dcrWallet.BlockHeader(ctx, blockHash)
+	if err != nil {
+		return nil, err
+	}
+
+	medianTime, err := w.medianTime(ctx, blockHash)
```
chappjc, line 557: The first thing that `medianTime` does is `w.dcrWallet.BlockHeader(ctx, blockHash)`.  Might consider having `medianTime` return that header although it's a minor redundancy.

client/asset/dcr/spv.go
```diff
@@ -0,0 +1,1020 @@
...
+// Part of the Wallet interface.
+func (w *spvWallet) InternalAddress(ctx context.Context, _ string) (stdaddr.Address, error) {
+	return w.NewInternalAddress(ctx, w.acctNum, wallet.WithGapPolicyIgnore())
+}
+
+// SignRawTransaction signs the provided transaction.
+// Part of the Wallet interface.
+func (w *spvWallet) SignRawTransaction(ctx context.Context, baseTx *wire.MsgTx) (*wire.MsgTx, error) {
+	tx := baseTx.Copy()
+	sigErrs, err := w.dcrWallet.SignTransaction(ctx, tx, txscript.SigHashAll, nil, nil, nil)
```
chappjc, line 528: :star_struck: After all that zec signing pain, this is a breath of fresh air.

client/asset/dcr/spv.go
```diff
@@ -0,0 +1,1020 @@
...
+		Addresses:     []string{addrStr},
+		Confirmations: uint32(txd.Block.Height - tipHeight + 1),
+	}, nil
+}
+
+// ExternalAddress returns an external address using GapPolicyIgnore.
+// Part of the Wallet interface.
+//
+// NOTE: Why GapPolicyIgnore instead of GapPolicyWrap? Can we recover from an
+// ignored gap policy without extreme measures?
```
chappjc, line 513: Maybe we are trying a little to hard to avoid address re-use.  In the Electrum PR, I spent some time looking at where we are burning through addresses, and I decided to [add a `refundAddress` method to `client/asset/btc.Wallet`](https://github.com/decred/dcrdex/pull/1607/commits/c63ef108d423cd5ffe64de25a8d8b5fd68f38915#diff-1d84027a7ea77f3c9b4b3adea610bde93e9a489d8e8f1f977b8121699c8af0fdR32-R34) where the addresses are very rarely used and we can tolerate some reuse.

The other big source of wasted addresses is the address provided in orders, when the address really [could be provided closer to match time](https://github.com/decred/dcrdex/pull/1607/commits/c63ef108d423cd5ffe64de25a8d8b5fd68f38915#diff-f750305f8da9a4959a035c59e067874c849530184462549bbe658a8a79a89e5fR424) although with more messaging, so that is not as easy.

Regarding recovery, when we implement Rescanner, we can use the DiscoverAddresses method with a provided gap limit.  Also, I haven't gotten far enough with the review, but we should ensure the seed can be used in an external dcrwallet for manual recovery.

chappjc (changes requested):
Feeling great about this.  About 2/3 of it reviewed.

JoeGruffins (approved):

JoeGruffins:
It looks like loadbot needs a `go mod tidy`

chappjc:
> It looks like loadbot needs a `go mod tidy`

That settles it.  We're updating CI to tidy loadbot too.

client/asset/dcr/spv.go
```diff
@@ -0,0 +1,976 @@
...
+// Part of the Wallet interface.
+func (w *spvWallet) AccountOwnsAddress(ctx context.Context, addr stdaddr.Address, _ string) (bool, error) {
+	ka, err := w.KnownAddress(ctx, addr)
+	if err != nil {
+		return false, fmt.Errorf("KnownAddress error: %w", err)
+	}
+	if ka.AccountName() != w.acctName {
+		return false, nil
+	}
+	own, err := w.HaveAddress(ctx, addr)
```
chappjc, line 391: I was curious about these two `Wallet` methods:

```go
func (w *Wallet) HaveAddress(ctx context.Context, a stdaddr.Address) (bool, error) {
	const op errors.Op = "wallet.HaveAddress"
	err := walletdb.View(ctx, w.db, func(tx walletdb.ReadTx) error {
		addrmgrNs := tx.ReadBucket(waddrmgrNamespaceKey)
		_, err := w.manager.Address(addrmgrNs, a)
		return err
	})
	if err != nil {
		if errors.Is(err, errors.NotExist) {
			return false, nil
		}
		return false, errors.E(op, err)
	}
	return true, nil
}
```

```go
func (w *Wallet) KnownAddress(ctx context.Context, a stdaddr.Address) (KnownAddress, error) {
	const op errors.Op = "wallet.KnownAddress"

	var ma udb.ManagedAddress
	var acct uint32
	var acctName string
	err := walletdb.View(ctx, w.db, func(tx walletdb.ReadTx) error {
		addrmgrNs := tx.ReadBucket(waddrmgrNamespaceKey)
		var err error
		ma, err = w.manager.Address(addrmgrNs, a)
		if err != nil {
			return err
		}
		acct = ma.Account()
		acctName, err = w.manager.AccountName(addrmgrNs, acct)
		return err
	})
	if err != nil {
		return nil, errors.E(op, err)
	}

	var acctKind AccountKind
	switch {
	case acct < udb.ImportedAddrAccount:
		acctKind = AccountKindBIP0044
	case acct == udb.ImportedAddrAccount:
		acctKind = AccountKindImported
	default:
		acctKind = AccountKindImportedXpub
	}
	return wrapManagedAddress(ma, acctName, acctKind)
}
```

Some observations:
- `HaveAddress` seems to be a subset of `KnownAddress`
- `HaveAddress` doesn't actually do say if it is the "owner"
- we should check `ka.AccountKind() == wallet.AccountKindBIP0044` to ensure we actually have the private keys (the other kinds are `AccountKindImported` and `AccountKindImportedXpub`).  Maybe just exclude imported xpub?

client/asset/dcr/spv.go
```diff
@@ -0,0 +1,976 @@
...
+func (w *spvWallet) LockUnspent(ctx context.Context, unlock bool, ops []*wire.OutPoint) error {
+	if unlock {
+		for _, op := range ops {
+			w.UnlockOutpoint(&op.Hash, op.Index)
+		}
+		return nil
+	}
+	for _, op := range ops {
+		w.LockOutpoint(&op.Hash, op.Index)
+	}
```
chappjc, line 441: All good, but they have the same signature, so another approach:
```go
	fun := w.LockOutpoint
	if unlock {
		fun = w.UnlockOutpoint
	}
	for _, op := range ops {
		fun(&op.Hash, op.Index)
	}
	return nil
```

client/asset/dcr/spv.go
```diff
@@ -0,0 +1,976 @@
...
+}
+
+func (w *spvWallet) notesLoop(ctx context.Context, dcrw *wallet.Wallet) {
+	txNotes := dcrw.NtfnServer.TransactionNotifications()
+	defer txNotes.Done()
+	// removeTxNotes := dcrw.NtfnServer.RemovedTransactionNotifications()
+	// defer removeTxNotes.Done()
+	// acctNotes := dcrw.NtfnServer.AccountNotifications()
+	// defer acctNotes.Done()
+	// tipNotes := dcrw.NtfnServer.MainTipChangedNotifications()
```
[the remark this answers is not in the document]
buck54321: `TransactionNotifications` gives us a notification for every block, regardless of whether there are wallet transactions present.

client/asset/dcr/spv.go
buck54321: Important to accept `AccountKindImported` too?

chappjc: I suppose that would be needed for an address for an imported private key.
I commented on that other pr so hopefully it never means imported public key

chappjc (approved):
All good.  Manually tested: (1) success, (2) refund maker DCR contract, and (3) find maker redemption with hacked ghosting.

PR #1639, opened 2022-06-01
ui: Add button to copy address on wallets page

martonp:
Screen Shot 2022-06-02 at 11 45 52 AM


client/webserver/site/src/js/wallets.ts
```diff
@@ -232,6 +233,10 @@ export default class WalletsPage extends BasePage {
     Doc.hide(this.page.forms)
   }
 
+  copyAddress () {
+    navigator.clipboard.writeText(this.page.depositAddress.textContent || '')
```
chappjc, line 237: Looks good.  https://caniuse.com/mdn-api_clipboard_writetext
I'm assuming `'click'` counts as a "user gesture event handlers such as pointerdown or pointerup." so it will work on Safari.  Can anyone test on Safari?

martonp: It works on Safari. I put the icon inside the box.. I think it looks better.

buck54321 (commented):
image

https://stackoverflow.com/questions/51805395/navigator-clipboard-is-undefined

chappjc:
What host makes that happen @buck54321 ?  A non-loopback addr?

chappjc:
So is that a chrome bug with ipv6?  That looks like it should qualify as a secure origin.

buck54321:
Looks like this does the trick.

https://github.com/martonp/dcrdex/compare/copyAddress...buck54321:martonp-copy-addr?expand=1

chappjc:
Ah, yes!  I read it as `[::1]` which basically the opposite of `[::]`.

buck54321:
Actually, this would prevent me from copying if I access the client from, say, my tablet on the same network.

buck54321:
Is it unsafe to use `execCommand` with a hard-coded argument? If it's safe, might as well do it that way, no?

chappjc:
`execCommand` and friends are deprecated: https://developer.mozilla.org/en-US/docs/Web/API/document/execCommand

> Deprecated: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the [compatibility table](https://developer.mozilla.org/en-US/docs/Web/API/document/execCommand#browser_compatibility) at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.

martonp:
I pulled your commit in @buck54321

JoeGruffins (approved):
Working well.

I think it would be nice if there was indication that the copy worked. Like a message that says `Copied!` or something. dcrdata has this if you click on the copy icon https://dcrdata.decred.org/address/DsZJ3hUqfaGR4GbwvLDLaSxTGJvYk6ypk4K

PR #1640, opened 2022-06-01
site: make cancel button respond with spinner

chappjc:
This makes the button on the cancel form behave like the order submission button in that when clicked, the button is replaced with a spinner, and any error from the cancel request is shown on the form instead of the notification menu.


buck54321 (commented):
What do you think about this?

https://github.com/chappjc/dcrdex/compare/cancel-spinner...buck54321:chappjc-cancel-loading?expand=1

chappjc:
Sure that looks fine.  The order submit button works with the spinner though, so different.

JoeGruffins (approved):

PR #1642, opened 2022-06-02
client: Base fraction unlocking on total lots.

### LINKED ISSUE #1544: client/eth: attempting to unlock more than is reserved

JoeGruffins:
closes #1544

The function wasn't a problem on the default simnet, but when used with smaller lot sizes `isDust` does not scale well. You can try out the way in master here: https://go.dev/play/p/QNqo2EDUiUX

I think using the ratio of one lot to max lots and multiplying that times the max reserved is more straightforward.

Also adding a log for client to see the net and locktimes on start up.

### DIFF
--- a/client/cmd/dexc/main.go
+++ b/client/cmd/dexc/main.go
@@ -80,6 +80,9 @@ func mainCore() error {
 		log.Infof("Logging with UTC time stamps. Current local time is %v",
 			time.Now().Local().Format("15:04:05 MST"))
 	}
+	log.Infof("dexc starting for network: %s", cfg.Net)
+	log.Infof("Swap locktimes config: maker %s, taker %s",
+		dex.LockTimeMaker(cfg.Net), dex.LockTimeTaker(cfg.Net))
 
 	defer func() {
 		if pv := recover(); pv != nil {
--- a/client/core/core_test.go
+++ b/client/core/core_test.go
@@ -2966,7 +2966,8 @@ func TestRefundReserves(t *testing.T) {
 	rate := dcrBtcRateStep * 100
 
 	lo, dbOrder, preImg, _ := makeLimitOrder(dc, true, qty, rate)
-	lo.BaseAsset = tACCTAsset.ID
+	lo.BaseAsset = tUTXOAssetB.ID
+	lo.QuoteAsset = tACCTAsset.ID
 	lo.Force = order.StandingTiF
 	loid := lo.ID()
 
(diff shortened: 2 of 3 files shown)


chappjc:
Isn't there a way to just pass a `final bool` to unlock all remaining?  Surely either the caller or the trackedTrade can determine that based on some counters.
Just the fact that we're trying to detect left over reserves is super confusing and a little unsettling tbh.

chappjc:
Another problem that relates to refund reserves is that it's never being saved to the DB, only `RedemptionReserves`

https://github.com/decred/dcrdex/blob/a1d4690e5837c5882a26ff31034a657987bb00ba/client/core/core.go#L4534-L4536

But there's a `RefundReserves` too:

https://github.com/decred/dcrdex/blob/a1d4690e5837c5882a26ff31034a657987bb00ba/client/db/types.go#L280-L287

chappjc:
Definitely fix these things, but I'm investigating a reworked reserves system.  If it pans out I'll share.  I don't want to spend more than a day on it.

JoeGruffins:
just rebased

martonp (commented):
The change makes sense. I guess this was happening when a significant portion of the reserved funds were being used for the tx fees?

JoeGruffins:
> I guess this was happening when a significant portion of the reserved funds were being used for the tx fees?

Yes, toggling the reserved amount can also make `isDust` true or false. Trying to think of a way to indicate this is the last trade so we can not worry about dust... or something.

chappjc:
> Trying to think of a way to indicate this is the last trade so we can not worry about dust... or something.

I'll put something up in the morning.  Almost ready.
We can merge this now though.

chappjc (approved):

chappjc:
@JoeGruffins Lots of test code to update, also completely untested with ETH, but this is the way I'm going: https://github.com/decred/dcrdex/commit/a96a980496426be176dcc61cb92ba6cda7e8df12  Please peruse to see the concept though.  The goal is to make it usable by the consumer (Core) and make rounding errors and the dust issue go away.  It does handle a remainder if the reserves given by the wallet are not evenly divisible by the given N, but that is not sensible and should always be zero.

In doing this, I realized the LCM stuff was unnecessary because it was looping through matches for a given trackedTrade, for which the result of `trade.isMarketBuy` would be the same for all matches.  But that's not important anymore if this concept works.

JoeGruffins:
```
type reservesCounter struct {
	done uint32 // mutable atomic

	num  uint32 // max lot / swap count
	mult uint64 // amount reserved per swap
	rem  uint64 // ideally zero, but handled if not
}
```
This is also what I had imagined we need to do, so pr it up...

PR #1643, opened 2022-06-02
docs: commit contract audit results

buck54321:
As far as I can tell, it's a perfect result. Congrats @martonp.


chappjc (approved):
OK.  Don't forget @JoeGruffins started the contract!

PR #1644, opened 2022-06-02
client: fix confirm when adding new dex server

### LINKED ISSUE #1531: ui: js errors registering for second dex from settings page
When you click Submit here *from the settings page* for a secondary DEX server:

image

it shows this in the console

image

and the Submit prompt remains although dexc has sent the reg txn.

vctt94:
this PR closes #1531

### DIFF
--- a/client/webserver/site/src/js/settings.ts
+++ b/client/webserver/site/src/js/settings.ts
@@ -291,15 +291,6 @@ export default class SettingsPage extends BasePage {
     form.style.right = '0'
   }
 
-  /* clearCertFile cleanup certFile value and selectedCert text */
-  clearCertFile () {
-    const page = this.page
-    page.certFile.value = ''
-    page.selectedCert.textContent = this.defaultTLSText
-    Doc.hide(page.removeCert)
-    Doc.show(page.addCert)
-  }
-
   /* gets the contents of the cert file */
   async getCertFile () {
     let cert = ''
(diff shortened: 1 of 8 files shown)


JoeGruffins (approved):
Working well.

client/webserver/site/src/js/settings.ts
```diff
@@ -319,8 +310,6 @@ export default class SettingsPage extends BasePage {
   /* Called after successful registration to a DEX. */
   async registerDEXSuccess () {
     const page = this.page
-    page.dexAddr.value = ''
-    this.clearCertFile()
```
chappjc, line 323: @martonp Would you comment on this?  You added this in https://github.com/decred/dcrdex/pull/1135

There's also a `clearCertFile` in forms.ts.

I'm not clear on the root cause of this bug or why the fix is appropriate.  @vctt94 Can you comment on what you observed going wrong, and why we do not need to clear this?

vctt94:
I believe the reason of the bug is the refactor which changed how forms were, to the newer version with modals. So when it gets to the point of register success, the page.addr component is already `''` and it throws an error.

Right now the `refresh()` method on forms is taking care of removing the `page.addr` value, but it is not calling the `clearCertFile()` method, making the cert file keep its value, if closing the form. 

I also added the `clearCertFile()` now on the refresh() method and should be working as expected, even though the password input still keep its value, should it also be cleaned?

client/webserver/site/src/js/settings.ts
martonp: In `forms.tmpl`, `dexAddr` used to be the id of the element, but it got updated to be a `data-tmpl` at some point, so it wasn't getting loaded into the `page` object on the settings page.

martonp (approved):
LGTM

chappjc (approved):
LGTM.  A glance from @buck54321 and we can merged.

buck54321 (approved):

PR #1648, opened 2022-06-04
client/eth/ui: Export ETH wallet to Metamask

martonp:
- `client/asset`: Adds a `WalletRestorer` interface which requires
  the `RestorationInfo` method to be implemented. `RestorationInfo`
  returns the wallet seed and instructions for the user to restore
  the wallet in various external wallets.
- client/eth: Implements the `WalletRestorer` interface. Returns
  information about how to restore the wallet in MetaMask.
- ui: Displays a new button on the wallet settings page if the wallet
  implements `WalletRestorer`. The user is promped to copy a disclaimer
  confirming that they understand the risks of restoring their wallet
  in an external wallet software. After doing so, they are able to see
  the information returned from the wallet's `RestorationInfo` function.

Screen Shot 2022-07-06 at 10 37 40 PM
Screen Shot 2022-07-06 at 10 37 54 PM

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -9,6 +9,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/sha256"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"math"
@@ -118,6 +119,12 @@ var (
 	unlimitedAllowanceReplenishThreshold = new(big.Int).Div(unlimitedAllowance, big.NewInt(2))
 
 	findRedemptionCoinID = []byte("FindRedemption Coin")
+
+	seedDerivationPath = []uint32{hdkeychain.HardenedKeyStart + 44,
+		hdkeychain.HardenedKeyStart + 60,
+		hdkeychain.HardenedKeyStart,
+		0,
+		0}
 )
 
 // WalletConfig are wallet-level configuration settings.
@@ -238,6 +245,7 @@ var _ asset.Wallet = (*TokenWallet)(nil)
 var _ asset.AccountLocker = (*ETHWallet)(nil)
 var _ asset.AccountLocker = (*TokenWallet)(nil)
 var _ asset.TokenMaster = (*ETHWallet)(nil)
+var _ asset.WalletRestorer = (*assetWallet)(nil)
 
 type baseWallet struct {
 	ctx         context.Context // the asset subsystem starts with Connect(ctx)
@@ -333,13 +341,7 @@ func CreateWallet(createWalletParams *asset.CreateWalletParams) error {
 		return err
 	}
 
-	// m/44'/60'/0'/0/0
-	path := []uint32{hdkeychain.HardenedKeyStart + 44,
-		hdkeychain.HardenedKeyStart + 60,
-		hdkeychain.HardenedKeyStart,
-		0,
-		0}
-	extKey, err := keygen.GenDeepChild(createWalletParams.Seed, path)
+	extKey, err := keygen.GenDeepChild(createWalletParams.Seed, seedDerivationPath)
 	defer extKey.Zero()
 	if err != nil {
 		return err
@@ -2033,7 +2035,29 @@ func (w *TokenWallet) EstimateRegistrationTxFee(feeRate uint64) uint64 {
 		return math.MaxUint64
 	}
 	return g.Transfer * feeRate
+}
 
+// RestorationInfo returns information about how to restore the wallet in
+// various external wallets.
+func (w *assetWallet) RestorationInfo(seed []byte) ([]*asset.WalletRestoration, error) {
+	extKey, err := keygen.GenDeepChild(seed, seedDerivationPath)
+	defer extKey.Zero()
+	if err != nil {
+		return nil, err
+	}
+	privateKey, err := extKey.SerializedPrivKey()
+	defer encode.ClearBytes(privateKey)
+	if err != nil {
+		return nil, err
+	}
+
+	return []*asset.WalletRestoration{
+		&asset.WalletRestoration{
+			Target:       "MetaMask",
+			Seed:         hex.EncodeToString(privateKey),
+			Instructions: "1. Open the settings menu\n2. Select \"Import Account\"\n3. Make sure \"Private Key\" is selected and paste the seed into the box",
+		},
+	}, nil
 }
 
 // SwapConfirmations gets the number of confirmations and the spend status
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -2053,7 +2053,7 @@ func (w *assetWallet) RestorationInfo(seed []byte) ([]*asset.WalletRestoration, 
 	}
 
 	return []*asset.WalletRestoration{
-		&asset.WalletRestoration{
+		{
 			Target:   "MetaMask",
 			Seed:     hex.EncodeToString(privateKey),
 			SeedName: "Private Key",
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -120,11 +120,13 @@ var (
 
 	findRedemptionCoinID = []byte("FindRedemption Coin")
 
-	seedDerivationPath = []uint32{hdkeychain.HardenedKeyStart + 44,
-		hdkeychain.HardenedKeyStart + 60,
-		hdkeychain.HardenedKeyStart,
-		0,
-		0}
+	seedDerivationPath = []uint32{
+		hdkeychain.HardenedKeyStart + 44, // purpose 44' for HD wallets
+		hdkeychain.HardenedKeyStart + 60, // eth coin type 60'
+		hdkeychain.HardenedKeyStart,      // account 0'
+		0,                                // branch 0
+		0,                                // index 0
+	}
 )
 
 // WalletConfig are wallet-level configuration settings.
@@ -342,11 +344,10 @@ func CreateWallet(createWalletParams *asset.CreateWalletParams) error {
 	}
 
 	extKey, err := keygen.GenDeepChild(createWalletParams.Seed, seedDerivationPath)
-	defer extKey.Zero()
 	if err != nil {
 		return err
 	}
-
+	defer extKey.Zero()
 	privateKey, err := extKey.SerializedPrivKey()
 	defer encode.ClearBytes(privateKey)
 	if err != nil {
@@ -2041,10 +2042,10 @@ func (w *TokenWallet) EstimateRegistrationTxFee(feeRate uint64) uint64 {
 // various external wallets.
 func (w *assetWallet) RestorationInfo(seed []byte) ([]*asset.WalletRestoration, error) {
 	extKey, err := keygen.GenDeepChild(seed, seedDerivationPath)
-	defer extKey.Zero()
 	if err != nil {
 		return nil, err
 	}
+	defer extKey.Zero()
 	privateKey, err := extKey.SerializedPrivKey()
 	defer encode.ClearBytes(privateKey)
 	if err != nil {
@@ -2053,9 +2054,14 @@ func (w *assetWallet) RestorationInfo(seed []byte) ([]*asset.WalletRestoration, 
 
 	return []*asset.WalletRestoration{
 		&asset.WalletRestoration{
-			Target:       "MetaMask",
-			Seed:         hex.EncodeToString(privateKey),
-			Instructions: "1. Open the settings menu\n2. Select \"Import Account\"\n3. Make sure \"Private Key\" is selected and paste the seed into the box",
+			Target: "MetaMask",
+			Seed:   hex.EncodeToString(privateKey),
+			Instructions: "Accounts can be imported by private key only if MetaMask has already be initialized. " +
+				"If this is your first time installing MetaMask, create a new wallet and secret recovery phrase. " +
+				"Then, to import your DEX account into MetaMask, follow the steps below:\n" +
+				`1. Open the settings menu
+				 2. Select "Import Account"
+				 3. Make sure "Private Key" is selected, and paste the private key above into the box`,
 		},
 	}, nil
 }
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -2054,14 +2054,15 @@ func (w *assetWallet) RestorationInfo(seed []byte) ([]*asset.WalletRestoration, 
 
 	return []*asset.WalletRestoration{
 		&asset.WalletRestoration{
-			Target: "MetaMask",
-			Seed:   hex.EncodeToString(privateKey),
+			Target:   "MetaMask",
+			Seed:     hex.EncodeToString(privateKey),
+			SeedName: "Private Key",
 			Instructions: "Accounts can be imported by private key only if MetaMask has already be initialized. " +
 				"If this is your first time installing MetaMask, create a new wallet and secret recovery phrase. " +
 				"Then, to import your DEX account into MetaMask, follow the steps below:\n" +
 				`1. Open the settings menu
 				 2. Select "Import Account"
-				 3. Make sure "Private Key" is selected, and paste the private key above into the box`,
+				 3. Make sure "Private Key" is selected, and enter the private key above into the box`,
 		},
 	}, nil
 }
--- a/client/asset/interface.go
+++ b/client/asset/interface.go
@@ -23,6 +23,7 @@ const (
 	WalletTraitRecoverer                            // The wallet is an asset.Recoverer.
 	WalletTraitWithdrawer                           // The Wallet can withdraw a specific amount from an exchange wallet.
 	WalletTraitSweeper                              // The Wallet can sweep all the funds, leaving no change.
+	WalletTraitRestorer                             // The wallet is an asset.WalletRestorer
 )
 
 // IsRescanner tests if the WalletTrait has the WalletTraitRescanner bit set.
@@ -73,6 +74,12 @@ func (wt WalletTrait) IsSweeper() bool {
 	return wt&WalletTraitSweeper != 0
 }
 
+// IsRestorer test if the WalletTrait has the WalletTraitRestorer bit set, which
+// indicates the wallet implements the WalletRestorer interface.
+func (wt WalletTrait) IsRestorer() bool {
+	return wt&WalletTraitRestorer != 0
+}
+
 // DetermineWalletTraits returns the WalletTrait bitset for the provided Wallet.
 func DetermineWalletTraits(w Wallet) (t WalletTrait) {
 	if _, is := w.(Rescanner); is {
@@ -98,6 +105,9 @@ func DetermineWalletTraits(w Wallet) (t WalletTrait) {
 	}
 	if _, is := w.(Sweeper); is {
 		t |= WalletTraitSweeper
+	}
+	if _, is := w.(WalletRestorer); is {
+		t |= WalletTraitRestorer
 	}
 	return t
 }
@@ -414,6 +424,22 @@ type LogFiler interface {
 // suggestion for swap operations.
 type FeeRater interface {
 	FeeRate() uint64
+}
+
+// WalletRestoration contains all the information needed for a user to restore
+// their wallet in an external wallet.
+type WalletRestoration struct {
+	Target       string `json:"target"`
+	Seed         string `json:"seed"`
+	Instructions string `json:"instructions"`
+}
+
+// WalletRestorer is a wallet which gives information about how to restore
+// itself in external wallet software.
+type WalletRestorer interface {
+	// RestorationInfo returns information about how to restore the wallet in
+	// various external wallets.
+	RestorationInfo(seed []byte) ([]*WalletRestoration, error)
 }
 
 // EarlyAcceleration is returned from the PreAccelerate function to inform the
--- a/client/asset/interface.go
+++ b/client/asset/interface.go
@@ -429,8 +429,11 @@ type FeeRater interface {
 // WalletRestoration contains all the information needed for a user to restore
 // their wallet in an external wallet.
 type WalletRestoration struct {
-	Target       string `json:"target"`
-	Seed         string `json:"seed"`
+	Target string `json:"target"`
+	Seed   string `json:"seed"`
+	// SeedName is the name of the seed used for this particular wallet, i.e
+	// Private Key.
+	SeedName     string `json:"seedName"`
 	Instructions string `json:"instructions"`
 }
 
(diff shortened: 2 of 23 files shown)


client/webserver/locales/en-us.go
```diff
@@ -236,4 +236,9 @@ var EnUS = map[string]string{
 	"Update TLS Certificate":      "Update TLS Certificate",
 	"registered dexes":            "Registered Dexes:",
 	"successful_cert_update":      "Successfully updated certificate!",
+	"export_wallet":               "Export Wallet",
+	"pw_for_wallet_seed":          "Enter your app password to show the wallet seed. Make sure nobody else can see your screen. If anyone gets accecss to the wallet seed, they will be able to steal all of your funds.",
+	"export_wallet_disclaimer":    "You WILL LOSE FUNDS if export your wallet and use the external wallet while you have active trades running in the DEX. It is recommended that you do not export your wallet unless you are an experienced user and you know what are doing.",
```
LasTshaMAN, line 241: maybe phrase like this:

if export your wallet and use the external wallet -> if you export your wallet and try using it

client/core/core.go
```diff
@@ -7023,6 +7023,39 @@ func (c *Core) WalletLogFilePath(assetID uint32) (string, error) {
...
+// WalletRestorationInfo returns information about how to restore the currently
+// loaded wallet for assetID in various external wallet software. This function
+// will return an error if the currently loaded wallet for assetID does not
+// implement the WalletRestorer interface.
+func (c *Core) WalletRestorationInfo(pw []byte, assetID uint32) ([]*asset.WalletRestoration, error) {
+	crypter, err := c.encryptionKey(pw)
+	if err != nil {
+		return nil, fmt.Errorf("WalletRestorationInfo password error: %w", err)
+	}
+
```
JoeGruffins, line 7035: Can you do `defer crypter.Close()` after checking err?

client/core/core.go
```diff
@@ -7023,6 +7023,39 @@ func (c *Core) WalletLogFilePath(assetID uint32) (string, error) {
...
+	seed, _, err := c.assetSeedAndPass(assetID, crypter)
+	if err != nil {
+		return nil, fmt.Errorf("assetSeedAndPass error: %w", err)
+	}
```
JoeGruffins, line 7039: Can you do `defer encode.ClearBytes(seed)` before or after the error check?

client/webserver/api.go
```diff
@@ -418,6 +419,33 @@ func (s *WebServer) apiUpdateCert(w http.ResponseWriter, r *http.Request) {
 	writeJSON(w, simpleAck(), s.indent)
 }
 
+// apiRestoreWalletInfo is the handler for the '/restorewalletinfo' API
+// request.
+func (s *WebServer) apiRestoreWalletInfo(w http.ResponseWriter, r *http.Request) {
+	form := &struct {
+		AssetID uint32
+		Pass    encode.PassBytes
+	}{}
```
JoeGruffins, line 428: Can you do a `defer form.Pass.Clear()` after this?

Unrelated, but in the function below I see a `defer form.Pass.Clear()` and a `zero(form.Pass)` which is redundant. I wonder if I did that.

client/asset/eth/eth.go
```diff
@@ -2099,7 +2101,29 @@ func (w *TokenWallet) EstimateRegistrationTxFee(feeRate uint64) uint64 {
...
+	privateKey, err := extKey.SerializedPrivKey()
+	defer encode.ClearBytes(privateKey)
+	if err != nil {
+		return nil, err
+	}
 
+	return []*asset.WalletRestoration{
+		&asset.WalletRestoration{
+			Target:       "MetaMask",
+			Seed:         hex.EncodeToString(privateKey),
```
JoeGruffins, line 2123: I don't know what terminology is better to use, but not really a seed. It's an address's private key. Is this what it will be called when using in metamask?

Checking out metamask on firefox browser, it looks like it wants a word phrase:
image

Should we convert to words? We would rather have the user copy the words with pencil and paper than save the hex or copy paste it, I think.

client/asset/eth/eth.go
```diff
@@ -2099,7 +2101,29 @@ func (w *TokenWallet) EstimateRegistrationTxFee(feeRate uint64) uint64 {
...
+}
+
+// RestorationInfo returns information about how to restore the wallet in
+// various external wallets.
+func (w *assetWallet) RestorationInfo(seed []byte) ([]*asset.WalletRestoration, error) {
+	extKey, err := keygen.GenDeepChild(seed, seedDerivationPath)
+	defer extKey.Zero()
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
JoeGruffins: Looks same in `GenDeepChild`

client/asset/eth/eth.go
martonp: Yes, typing the private key in that box works. I should update the instructions to say that the wallet already needs to be initialized before the account can be added.

Going from seed phrase to seed is a one way operation. The link you shared was for generating the mnemonic, but the process for deriving the seed is below that:
https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#from-mnemonic-to-seed

LasTshaMAN: ah, right, because BIP39 mnemonic is encoded represenatation for "initial entropy" (not seed),

and, looks like having word-based seed requires some rework of `seed` concept (and related `PrimaryCredentials` structure) and its probably difficult to implement it in a backward-compatible manner ...

JoeGruffins: > DEX requires some rework of seed concept

Do you mean the app seed?

Since this seed is something new and not used by anything but eth, we could do whatever we want with this "seed" I think, unless I'm missing something.

But the hex is fine with me. It's an emergency feature for refunds correct? Probably not used regularly.

We can't convert bytes -> mnemonic?

JoeGruffins:
Working on gorli testnet. Imported straight from quark.
image

client/asset/eth/eth.go
LasTshaMAN: > Do you mean the app seed?

Yes, I believe it is called `dex app seed`, this seed stored in this variable - https://github.com/decred/dcrdex/blob/a1d4690e5837c5882a26ff31034a657987bb00ba/client/db/types.go#L69-L71

let me try to lay out what happens next (maybe for future reference, pls correct me if I get it wrong)

`dex app seed` gets generated [here](https://github.com/decred/dcrdex/blob/a1d4690e5837c5882a26ff31034a657987bb00ba/client/core/core.go#L3373-L3380) and then **asset-specific seed** is derived from it as one-way operation [like this](https://github.com/decred/dcrdex/blob/a1d4690e5837c5882a26ff31034a657987bb00ba/client/core/core.go#L1872-L1880),
then **asset-specific seed** is used as `BIP 32 master-seed` when [deriving eth private key from it](https://github.com/decred/dcrdex/blob/a1d4690e5837c5882a26ff31034a657987bb00ba/client/asset/eth/eth.go#L336-L346) at `m/44'/60'/0'/0/0` path (and this **eth private key** gets exported by functionality implemented in this PR.

Now, `BIP 39 seed` would be the equivalent of DEX **asset-specific seed**, 
the problem is one can't convert `BIP 39 seed` (and hence DEX **asset-specific seed**) back into words, because it is [a one-way operation](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#from-mnemonic-to-seed).

LasTshaMAN: assuming I've layed out all bits correctly,

its probably possible to use `asset-specific seed` not as `BIP 39 seed` but as 
[BIP 39 entropy](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#generating-the-mnemonic),

and then we'll be able to recover this entropy in `RestorationInfo` func, and construct 24 word phrase from it,

I'm not suggesting that's must be done, just laying out possible options (I also don't have much context on why this feature gets implemented, but assume its nice to be able to take your DEX-generated wallet-phrase, for example, and use it in another app for whatever reason, better UI etc)

chappjc: That's the flow, yes.  This is almost entirely for recovery IMO.  We don't really want people messing in the wallet.
This is precisely why I made the stanalone assetseed app: https://github.com/decred/dcrdex/blob/master/client/cmd/assetseed/main.go

Because if all you have is your dex app seed and you want to access your funds without fixing dexc (presumably dexc is broken and you cannot use the feature added in this PR!), you just run the app and it gives your seed. In this case, I'd modify assetseed to take one more step for eth (deriving the private key from the asset seed since it's not a mneumonic that is recognized anywhere).

I don't really care much for all this generalization and expansion of the interfaces.  I'm more than happy with a recovery tool with asset specific code.

chappjc: > and, looks like having word-based seed in DEX requires some rework of `seed` concept (and related `PrimaryCredentials` structure) and its probably difficult to implement it in a backward-compatible manner

It's easy if we don't bother with BIP39 for the app seed <-> mneumonic at the top level.
https://github.com/decred/dcrwallet/blob/d1343d034a6431d90dc2500bc2d52a9ef4e9e029/walletseed/seed.go#L55
We can just adopt the same pgpwordlist and use ByteToMnemonic to make a mneumonic from our current app seed bytes.  This is unrelated to the metamask seed or other assets that use bip39 or other mneumonic schemes of course.

LasTshaMAN: Since app seed is not meant to be imported in any other app besides DEX I think it doesn't make much sense to make it BIP39-compliant,
representing app seed as words (from pgpwordlist linked above) in UI is preferable IMO (possibly, with an option to show it as binary seed),

it you agree that it's better UX, I can try looking into it and sketching up something working

chappjc: For the app seed, I think it would be somewhat of an improvement to show words using our (decred) scheme as we've discussed.  For making a paper wallet it's harder to mess up at least.

client/webserver/api.go
martonp: I'll remove the `zero(form.Pass)`.

client/asset/eth/eth.go
```diff
@@ -2099,7 +2101,29 @@ func (w *TokenWallet) EstimateRegistrationTxFee(feeRate uint64) uint64 {
...
+	defer encode.ClearBytes(privateKey)
+	if err != nil {
+		return nil, err
+	}
 
+	return []*asset.WalletRestoration{
+		&asset.WalletRestoration{
+			Target:       "MetaMask",
+			Seed:         hex.EncodeToString(privateKey),
+			Instructions: "1. Open the settings menu\n2. Select \"Import Account\"\n3. Make sure \"Private Key\" is selected and paste the seed into the box",
```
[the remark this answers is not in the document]
martonp: Do you mean in the instructions or for the header? Exporting the wallet might not use private key for other wallets.

chappjc:
Reviewing, but I see there's a scss conflict with your address clipboard PR merged.

client/asset/eth/eth.go
chappjc: Regarding the suggestion to use the `createWalletParams.Seed` as entropy to BIP39 mnemonic and seed generation, we could consider doing that even if we only give the account private key as in this PR.  Just to leave the door opened...

https://go.dev/play/p/vaMC6nA4k4P

https://iancoleman.io/bip39/
- Check the box "Show entropy details"
- Paste "f32a6d5737296e7d0dda7b78dfe5cf139a000374f7e32c886a19bfe183dcbe1b" into "Entropy" text box
- Select "12 Words" for "Mnemonic Length"
- Verify "BIP39 Mnemonic" and "BIP39 Seed" match the playground

I just don't know if that would be useful if other wallets do it a bit differently or use a password with the salt.

This obviously changes the account keys derived from a given app seed.

client/webserver/locales/en-us.go
JoeGruffins, line 241: I don't know if it's worth mentioning, but copying anything will put that plaintext into your clipboard which any program can easily access. That's why words copied with pencil and paper are potentially better. But, I think it's fine to let advanced users decide for themselves if they want to take the risk.

JoeGruffins (approved):
Looks good to me, but there are some conflicts.

client/asset/eth/eth.go
chappjc: Just want to point out one more time that taking the `m/44'/60'/0'/0/0` derivation steps to get an account private key doesn't have much point if there's no usable seed words corresponding to `m`.  This is not related to the app seed discussion, to be clear.

We just want to go from `createWalletParams.Seed` -> account private key, which we can get with `m/0'/0` for example.

My [previous comment](https://github.com/decred/dcrdex/pull/1648#discussion_r898428136) would go the other direction and insert another step so there is a seed to go with the `m/44'/60'/0'/0` path, just in case in the future we'd like to obtain a seed phrase that will derive the account private key we are exporting now.  I still prefer just exporting the private key though!

Am I missing something or is the `seedDerivationPath` used just because it's familiar?

client/asset/interface.go
```diff
@@ -59,6 +60,10 @@ func (wt WalletTrait) IsRecoverer() bool {
 	return wt&WalletTraitRecoverer != 0
 }
 
+func (wt WalletTrait) IsExporter() bool {
```
buck54321, line 63: `IsRestorer`?

client/webserver/site/src/js/wallets.ts
```diff
@@ -79,12 +88,16 @@ export default class WalletsPage extends BasePage {
...
   currentForm: PageElement
+  restoreInfoCard: HTMLElement
 
   constructor (body: HTMLElement) {
     super()
     this.body = body
     const page = this.page = Doc.idDescendants(body)
 
+    this.restoreInfoCard = page.restoreInfoCard.cloneNode(true) as HTMLElement
+    page.restoreInfoCard.remove()
```
buck54321, line 99: Use `Doc.cleanTemplates` so that the `id` is removed too.

client/webserver/site/src/js/wallets.ts
```diff
@@ -647,6 +665,75 @@ export default class WalletsPage extends BasePage {
     window.open(url.toString())
   }
 
+  // displayExportWalletDisclaimer displays a popup which prompts the user to
+  // retype a text confirming that they understand the risks of exporting the
+  // wallet.
+  async displayExportWalletDisclaimer () {
```
buck54321, line 671: > prompts the user to retype a text confirming that they understand the risks of exporting the wallet

Mixed feelings about this new security step. It seems a little random. We allow exporting of the application seed, which gives access to all funds from all native wallets, with just a password check. Why would this need stronger security? Or maybe the app seed needs stronger security. I'm also feeling like a password and good messaging is sufficient though.

buck54321 (changes requested):

client/webserver/site/src/js/wallets.ts
martonp: > Why would this need stronger security?

When exporting the funds to an external wallet, there is the additional risk of losing funds when using the wallet at the same time as using the DEX. Most people are aware that if they lose their seed the cannot recover their funds, but they are not aware of this additional risk.

chappjc: No comment yet about the actual text here, but the risk is pronounced with ETH and tokens where:
- ETH needs to be available for txns, including redeems and refunds, which can be time sensitive
- it is very easy to clobber a dex-managed txn by creating txn in the external non-managed wallet that uses the same nonce.  Even if it is just a speed-up, our dex client probably isn't hardened enough to deal with a tx vanishing even if the desired contract interaction is still affected.  The flip side is that DEX can create a txn using a nonce that may have been used by the external wallet clobbering that one.

client/webserver/locales/en-us.go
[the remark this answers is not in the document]
martonp: Hows this:

`Using an externally restored wallet while you have active trades running in the DEX could result in failed trades and LOST FUNDS.`

client/webserver/site/src/js/wallets.ts
martonp: If the external wallet uses a locked UTXO, the trade will be screwed up as well.

client/asset/eth/eth.go
JoeGruffins: I think just because it's familiar.

Working on some Trevor stuff which also involved just needing an arbitrary, deterministic private key, the argument was made that starting with the seed and doing bip 39 derivation made it easier to derive addresses with other software, if you ever needed to.

Yes we only use the one address atm, but maybe in the future there's more we want to do here.

> no usable seed words corresponding to m.

I guess I don't understand what this means. We COULD export `m` in the future?

JoeGruffins: Oh, we would need "seed words" for most wallets in order to import the seed? because we cannot go from seed hex -> words apparently?huh...

chappjc: Right.  I get the reasoning of making it "easier to derive addresses with other software, if you ever needed to", but that means words, not the seed bytes that come from words, which are presently non-existent.  Hence the playground link I posted, if we want to have yet another dep "just in case" we want words available in the future for this purpose.  I dunno

martonp: I think it's a good idea to be able to generate the words if needed. I can do a separate PR for that though, that doesn't really affect anything in this one.

client/webserver/locales/en-us.go
martonp: How much more risky is copying to the clipboard than just typing the password? Is accessing someone's clipboard that much easier than installing a keylogger?

chappjc:
The dialog looks like this:

image

"Seed" should be "Private Key", right?  Also, the formatting can be improved with `mono` and `fs14:

image

chappjc:
When I select and copy the private key, it seems to include a trailing space even though I don't see one selected.  When I paste into metamask I saw this:

image

I pasted into a text editor and figured out the trailing space.

client/webserver/site/src/js/wallets.ts
chappjc: > We allow exporting of the application seed, which gives access to all funds from all native wallets, with just a password check

@buck54321  You mean by initializing another dexc instance with the app seed or via the `assetseed` CLI app (or advanced/manual coding)?  

> Or maybe the app seed needs stronger security.

Possibly a sentence in the view seed dialog advising not to operate two instances of dexc or the native wallets simultaneously.

> I'm also feeling like a password and good messaging is sufficient though.

No objection to that either I suppose, if it's fairly strong language, even with bold and/or red font for the sentence that says what not to do and what can happen in the worst case scenario.

client/webserver/locales/en-us.go
LasTshaMAN: It seems clipboard peeking is far worse,

I got curious and tried to find some info on it, but didn't seems to find good examples right away, however I tired the following:
- I've installed [a simple go binary that works with clipboard](https://github.com/golang-design/clipboard) (no root permissions required for this I think)
- then I just ran `gclip --paste` (no root needed as well) and basically got clipboard contents

^ I think this means pretty much any binary on my system (Mac, but I think this is applicable to most OS) running in user space can peek secrets in clipboard at any time

client/webserver/site/src/js/wallets.ts
martonp: I've updated to using red text, (you can see in the PR description) and removed the extra security step.

martonp:
> When I select and copy the private key, it seems to include a trailing space even though I don't see one selected. When I paste into metamask I saw this:

Were you using Firefox? I put the span element into a div and it seems to be working well now.

chappjc:
> > When I select and copy the private key, it seems to include a trailing space even though I don't see one selected. When I paste into metamask I saw this:
> 
> Were you using Firefox? I put the span element into a div and it seems to be working well now.

Yup, Firefox.  Is that a quirk of FF you've seen?  Selecting the <br> too?

martonp:
> Yup, Firefox. Is that a quirk of FF you've seen? Selecting the `<br>` too?

Yeah I'm guessing that's what it was. There's some discussion online about extra spaces at the end of text after double clicking and copying in Firefox, but after putting it into the `<div>` the way it is now, it seems to be working. In Chrome it works the previous way as well.

chappjc (approved):
LGTM.
Other reviewers should circle back though.

client/core/core.go
```diff
@@ -7057,6 +7057,41 @@ func (c *Core) WalletLogFilePath(assetID uint32) (string, error) {
...
+	defer encode.ClearBytes(seed)
+
+	wallet, found := c.wallet(assetID)
+	if !found {
+		return nil, fmt.Errorf("no wallet configured for asset %d", assetID)
+	}
+
+	restorer, ok := wallet.Wallet.(asset.WalletRestorer)
+	if !ok {
+		return nil, fmt.Errorf("wallet for asset %d cannot be restored", assetID)
```
LasTshaMAN, line 7084: might wanna be specific and say that "wallet doesn't support exporting functionality"

client/asset/eth/eth.go
```diff
@@ -2016,7 +2019,35 @@ func (w *TokenWallet) EstimateRegistrationTxFee(feeRate uint64) uint64 {
...
+		&asset.WalletRestoration{
+			Target:   "MetaMask",
+			Seed:     hex.EncodeToString(privateKey),
+			SeedName: "Private Key",
+			Instructions: "Accounts can be imported by private key only if MetaMask has already be initialized. " +
+				"If this is your first time installing MetaMask, create a new wallet and secret recovery phrase. " +
+				"Then, to import your DEX account into MetaMask, follow the steps below:\n" +
+				`1. Open the settings menu
+				 2. Select "Import Account"
+				 3. Make sure "Private Key" is selected, and paste the private key above into the box`,
```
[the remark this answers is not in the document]
martonp: I'll just replace "paste" with "enter". I already have the warning on the top, I don't think it needs to be mentioned over and over.

buck54321 (approved):
I haven't tested on testnet, but I was able to import keys and got the correct address when I set up MetaMask on simnet. As discussed on chat.decred.org, I wasn't able to see my balance on simnet for one reason or another. Regardless, it seemed obvious that the key was imported correctly. I'm approving, but if you know how to get MetaMask connected to simnet, we could potentially test out some of the security issues raised.

chappjc:
> I wasn't able to see my balance on simnet for one reason or another.

I couldn't make metamask work with simnet either.  But to get mm to at least connect to geth I had to add this to alpha:

`--http --http.corsdomain "*" --allow-insecure-unlock --http.api "eth,miner,web3,net,personal,rpc,txpool,db,admin,rpc"`

Then I could get it to use a custom RPC endpoint at http://127.0.0.1:8545, chain ID 42 (0x2a), and I observed geth fielding a bunch of RPC requests for best block, but no address requests strangely.

PR #1651, opened 2022-06-05
dexc/core: remove redundant validation

LasTshaMAN:
While reading this part of code I've noticed a duplicate check (same check is done several lines of codes above, I've moved it slightly higher because it can be done earlier)

looks like it can be removed ?


JoeGruffins (approved):

chappjc (approved):
ok

Issue #1618, opened 2022-05-16
client: using an external API as a fee rate fallback


chappjc:
I honestly wouldn't mind using an external API as a fallback.

https://explorer.dcrdata.org/insight/api/utils/estimatefee?nbBlocks=2

_Originally posted by @buck54321 in https://github.com/decred/dcrdex/pull/1498#discussion_r857720422_

buck54321:
This should be implemented in the wallet packages behind the `FeeRate` method, I think.

vctt94:
I can work on that

buck54321:
Our Litecoin wallet doesn't assign `BTCCloneCFG.FeeEstimator`, so the wallet falls back to `(*baseWallet).feeRate`, which itself falls back to an external API with hard-coded endpoints. Does this mean that Litecoin can end up with a Bitcoin fee estimate?

vctt94:
Okay, will do them next. Do you guys have any api's to recommend?

For bitcoin cash I was able to find: https://docs.bitcoincashnode.org/doc/json-rpc/estimatefee/
For zcash I am looking into: https://getblock.io/docs/available-nodes-methods/ZEC/JSON-RPC/estimatefee/

For litecoin and doge I still have not found a json-rpc api for them, with estimate smart fee, do you guys have any suggestions?

chappjc:
For an external API, we'd need a service, not the node's JSON-RPC API, unless I'm misunderstanding those suggestions @vctt94.

An option is to use the bitcore API (was insight) that provides many alts including BCH, LTC, DOGE, and ETH:
- https://api.bitcore.io/api/BCH/mainnet/fee/1
- https://api.bitcore.io/api/LTC/mainnet/fee/1
- https://api.bitcore.io/api/DOGE/mainnet/fee/1
- https://api-eth.bitcore.io/api/ETH/mainnet/fee/1

I'll look for a ZEC oracle too.

chappjc:
ZEC is interesting

https://github.com/zcash/zcash/issues/2942

`relayFee` from https://api.zcha.in/v2/mainnet/network:

```json
{
  "name": "mainnet",
  "accounts": 7236146,
  "transactions": 11589222,
  "blockHash": "0000000000ad94b0ae7b01b62ae9ced52f6fc16d959c6a449276e12a20fc1c97",
  "blockNumber": 1812084,
  "difficulty": 81371313.48398675,
  "hashrate": 10421641389,
  "meanBlockTime": 81.00833333333334,
[...]

PR #1654, opened 2022-06-07
client/wip: Add external api request to get fee rate in case of fallback

vctt94:
This PR still a WIP.

I added the method `externalFeeEstimator` and I am calling it at the `feeRate` method, as suggested by buck. 

I also added a `externalApiAllowed` var, so we can toggle external requests, but I am not sure where we can add it, so we can keep its value, any suggestions?


client/asset/dcr/dcr.go
```diff
@@ -748,7 +756,13 @@ func (dcr *ExchangeWallet) feeRate(confTarget uint64) (uint64, error) {
...
 	estimatedFeeRate, err := feeEstimator.EstimateSmartFeeRate(dcr.ctx, int64(confTarget), chainjson.EstimateSmartFeeConservative)
 	if err != nil {
-		return 0, err
+		if externalApiAllowed {
+			estimatedFeeRate, err = externalFeeEstimator(confTarget)
+			if err != nil {
+				dcr.log.Errorf("Failed to get fee rate from external API: %v", err)
+				return 0, err
+			}
+		}
```
LasTshaMAN, line 765: the original error that got us here is ignored, looks like

client/asset/dcr/dcr.go
```diff
@@ -759,6 +773,27 @@ func (dcr *ExchangeWallet) feeRate(confTarget uint64) (uint64, error) {
...
+	if err != nil {
+		return 0, err
+	}
+	body, err := ioutil.ReadAll(r.Body)
+	r.Body.Close()
+	if err != nil {
+		return 0, err
+	}
+	c := make(map[uint64]json.RawMessage)
+	json.Unmarshal(body, &c)
```
LasTshaMAN, line 789: looks like error is ingored

client/asset/dcr/dcr.go
vctt94: I am checking for error now.

client/asset/dcr/dcr.go
vctt94: I am checking for error now.

client/asset/dcr/config.go
```diff
@@ -45,6 +45,9 @@ type Config struct {
 	// Context should be canceled when the application exits. This will cause
 	// some cleanup to be performed during shutdown.
 	Context context.Context `ini:"-"`
+	// AllowExternalAPIRequest is a flag that allows the client to make requests
+	// to external apis.
+	AllowExternalAPIRequest bool `ini:"allowexternalapirequest"`
```
buck54321, line 50: `Config` items are set by the user via a corresponding `ConfigOpt` in the `WalletDefinition.ConfigOpts`. I agree that this is a wallet-level configuration option, and this is the right place for this field, but setting the option should not involve **core**, **asset**, or **db** in any way. We already have the system in place to populate `Config`.

```Go
{
	Key:         "allowexternalapirequest",
	DisplayName: "External fee rate estimates",
	Description: "Allow fee rate estimation from a block explorer API. " +
		"This is useful as a fallback for SPV wallets and RPC wallet th" +
		"at have just recently been started.",
	IsBoolean: true,
},
```

vctt94: Removed changes on core, asset and db and left them only on as wallet option. Much simpler indeed.

client/asset/dcr/dcr.go
```diff
@@ -164,6 +170,14 @@ var (
 				"limit orders without immediate time-in-force.",
 			IsBoolean: true,
 		},
+		{
+			Key:         "allowexternalapirequest",
+			DisplayName: "External fee rate estimates",
+			Description: "Allow fee rate estimation from a block explorer API. " +
+				"This is useful as a fallback for SPV wallets and RPC wallet th" +
+				"at have just recently been started.",
```
chappjc, line 178: Although this wrapping is valid and buck suggested it, it's a pain to grep for and confusing to read.  Longish code lines is just fine when they help readability. Let's just keep words together.

Also "RPC wallet" should be "RPC wallets", and <s>just</s>.

client/asset/dcr/dcr.go
```diff
@@ -748,7 +764,16 @@ func (dcr *ExchangeWallet) feeRate(confTarget uint64) (uint64, error) {
 	}
 	estimatedFeeRate, err := feeEstimator.EstimateSmartFeeRate(dcr.ctx, int64(confTarget), chainjson.EstimateSmartFeeConservative)
 	if err != nil {
-		return 0, err
+		dcr.log.Errorf("Failed to get fee rate with estimate smart fee rate: %v", err)
+		if !dcr.allowExternalAPIRequest {
+			return 0, err
+		}
+		dcr.log.Info("trying to estimate fee with externeral API")
```
chappjc, line 771: `"Retrieving fee rate from external API"`

`Debug`

client/asset/dcr/dcr.go
```diff
@@ -759,6 +784,30 @@ func (dcr *ExchangeWallet) feeRate(confTarget uint64) (uint64, error) {
...
 }
 
+// externalFeeEstimator gets the fee rate from the external API
+func externalFeeEstimator(nb uint64) (float64, error) {
+	url := externalApiUrl + "/utils/estimatefee?nbBlocks=" + strconv.Itoa(int(nb))
+	r, err := http.Get(url)
+	if err != nil {
+		return 0, err
+	}
+	body, err := ioutil.ReadAll(r.Body)
```
chappjc, line 794: `ioutil` is deprecated.

Also, we shouldn't use `ReadAll` without a limit on the read size (never trust a remote source).  Maybe consider a `io.LimitReader` with a fairly small limit, like a few kB, `1 << 14` for a high limit for this response.

client/asset/dcr/dcr.go
```diff
@@ -759,6 +784,30 @@ func (dcr *ExchangeWallet) feeRate(confTarget uint64) (uint64, error) {
...
+	r.Body.Close()
+	if err != nil {
+		return 0, err
+	}
+	c := make(map[uint64]json.RawMessage)
```
chappjc, line 799: The responses look like `{"1":0.0001}`, so why not a `map[uint64]float64` and then

```go
estimatedFeeRate, ok := c[nb]
if !ok {
	return 0, errors.New("no fee rate for requested number of blocks")
```

https://go.dev/play/p/m3WtFtbKIKK

Does the value sometimes include a string or other non-float value?

client/asset/dcr/dcr.go
```diff
@@ -759,6 +784,30 @@ func (dcr *ExchangeWallet) feeRate(confTarget uint64) (uint64, error) {
...
+	err = json.Unmarshal(body, &c)
```
chappjc, line 800: `resp.Body` is a `io.Reader`, so could just use `json.NewDecoder(resp.Body).Decode` directly instead of `ReadAll` (still with a read limit).

client/asset/dcr/dcr.go
```diff
@@ -759,6 +784,30 @@ func (dcr *ExchangeWallet) feeRate(confTarget uint64) (uint64, error) {
 	return 1 + uint64(atomsPerKB)/1000, nil // dcrPerKB * 1e8 / 1e3
 }
 
+// externalFeeEstimator gets the fee rate from the external API
+func externalFeeEstimator(nb uint64) (float64, error) {
+	url := externalApiUrl + "/utils/estimatefee?nbBlocks=" + strconv.Itoa(int(nb))
```
chappjc, line 789: `strconv.FormatUint(nb, 10)`

chappjc (changes requested):

vctt94:
Thanks for the review chapp. Adressed your comments.

client/asset/dcr/dcr.go
```diff
@@ -785,25 +786,26 @@ func (dcr *ExchangeWallet) feeRate(confTarget uint64) (uint64, error) {
 }
 
 // externalFeeEstimator gets the fee rate from the external API
-func externalFeeEstimator(nb uint64) (float64, error) {
-	url := externalApiUrl + "/utils/estimatefee?nbBlocks=" + strconv.Itoa(int(nb))
-	r, err := http.Get(url)
+func (dcr *ExchangeWallet) externalFeeEstimator(nb uint64) (float64, error) {
```
chappjc, line 789: Not a problem but I liked this as a plain function and it would be idiomatic like `externalFeeEstimator(ctx context.Context, nb uint64) (float64, error)`.

JoeGruffins:
Still WIP?

client/asset/dcr/dcr.go
```diff
@@ -759,6 +785,31 @@ func (dcr *ExchangeWallet) feeRate(confTarget uint64) (uint64, error) {
 	return 1 + uint64(atomsPerKB)/1000, nil // dcrPerKB * 1e8 / 1e3
 }
 
+// externalFeeEstimator gets the fee rate from the external API
+func externalFeeEstimator(ctx context.Context, nb uint64) (float64, error) {
+	url := externalApiUrl + "/utils/estimatefee?nbBlocks=" + strconv.FormatUint(nb, 10)
+	r, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
```
chappjc, line 791: Still need a timeout set.  The `dcr.ctx` is just to cancel requests on shutdown.  Can either wrap this `ctx` using `context.WithTimeout`, or you can change the client definition to `client http.Client{Timeout: t}`.  I'm not sure if there's a practical difference TBH.

client/asset/dcr/config.go
```diff
@@ -31,17 +31,18 @@ var (
...
+	TradingAccount          string  `ini:"tradingaccount"`
+	RPCUser                 string  `ini:"username"`
+	RPCPass                 string  `ini:"password"`
+	RPCListen               string  `ini:"rpclisten"`
+	RPCCert                 string  `ini:"rpccert"`
+	UseSplitTx              bool    `ini:"txsplit"`
+	FallbackFeeRate         float64 `ini:"fallbackfee"`
+	FeeRateLimit            float64 `ini:"feeratelimit"`
+	RedeemConfTarget        uint64  `ini:"redeemconftarget"`
+	AllowExternalAPIRequest bool    `ini:"allowexternalapirequest"`
```
buck54321, line 45: About half of this diff is just because you picked a long name here and in `ExchangeWallet`. Can we not find a shorter way to say this? How about `ApiFeeFallback`? 

`AllowExternalAPIRequest` doesn't mention fees anyway.

client/asset/dcr/dcr.go
```diff
@@ -72,6 +75,9 @@ const (
 	// hierarchical deterministic key derivation for the internal branch of an
 	// account.
 	acctInternalBranch uint32 = 1
+
+	// externalApiUrl is the URL of the external API in case of fallback.
+	externalApiUrl = "https://explorer.dcrdata.org/insight/api"
```
JoeGruffins, line 80: Testnet also has one. `https://testnet.dcrdata.org/insight/api` would help with testing.

JoeGruffins (commented):
Looks good to me, but does it close #1618 ? I'm assuming this issue is for all assets.

chappjc:
> Looks good to me, but does it close #1618 ? I'm assuming this issue is for all assets.

It does not.  It's of particular importance for BTC SPV and even DOGE and ZEC full node that can't figure out how to get fee rates.

client/asset/dcr/dcr.go
```diff
@@ -759,6 +788,39 @@ func (dcr *ExchangeWallet) feeRate(confTarget uint64) (uint64, error) {
...
 }
 
+// externalFeeEstimator gets the fee rate from the external API
+func externalFeeEstimator(ctx context.Context, nb uint64) (float64, error) {
+	var url string
+	if dex.Testnet == 1 {
+		url = testnetExternalApiUrl
+	} else {
+		url = externalApiUrl
+	}
```
buck54321, line 798: `if dex.Testnet == 1` is always `true`. You'll need to work with the `dex.Network` passed to `NewWallet` somehow. You **could** inspect the `*chaincfg.Params` to figure out what network you're on, but using the `dex.Network` would be better.

vctt94: my bad. Now I am adding a new global var `isTestnet`, which is stored on `NewWallet` and check when calling the `externalFeeEstimator`

client/asset/dcr/dcr.go
```diff
@@ -548,6 +568,12 @@ type findRedemptionResult struct {
 // NewWallet is the exported constructor by which the DEX will import the
 // exchange wallet.
 func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network) (*ExchangeWallet, error) {
+	// store the network because we might need it for external api calls.
+	if network == dex.Testnet {
+		isTestnet = true
+	} else {
+		isTestnet = false
+	}
```
buck54321, line 576: No globals here, please. Let's pass it in to `unconnectedWallet` and save it as a field on the `ExchangeWallet`.

JoeGruffins (approved):

client/core/core.go
```diff
@@ -1914,6 +1914,7 @@ func (c *Core) loadWallet(dbWallet *db.Wallet) (*xcWallet, error) {
 			}
 		},
 		DataDir: c.assetDataDirectory(assetID),
+		Network: c.Network(),
```
chappjc, line 1917: See the call to `asset.OpenWallet` about 6 lines down.  Then see:

```go
func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network) (*ExchangeWallet, error)
```

There is no need to add `Network` to `asset.WalletConfig` because it is provided directly to `NewWallet`.

chappjc (changes requested):

client/asset/dcr/dcr.go
```diff
@@ -885,6 +939,39 @@ func (dcr *ExchangeWallet) feeRate(confTarget uint64) (uint64, error) {
 	return 1 + uint64(atomsPerKB)/1000, nil // dcrPerKB * 1e8 / 1e3
 }
 
+// externalFeeEstimator gets the fee rate from the external API
+func externalFeeEstimator(ctx context.Context, net dex.Network, nb uint64) (float64, error) {
+	var url string
+	if net == dex.Testnet {
+		url = testnetExternalApiUrl
+	} else {
+		url = externalApiUrl
```
chappjc, line 948: Note that for simnet it's gonna use mainnet's fee estimates.  This isn't bad TBH as it allows some more fee rate variability (maybe not with DCR, but in general it would).

client/core/core.go
vctt94: should I just pass `network` directly to `unconnectedWallet()` instead?

like:
`dcr, err := unconnectedWallet(cfg, walletCfg, chainParams, logger, network)`

chappjc: Yeah, or set `dcr.network = network` in `NewWallet`.  I honestly don't care which, but `unconnectedWallet` is probably right.

Also, I don't think the `ExchangeWallet` fields should be exported.

client/asset/dcr/dcr.go
```diff
@@ -850,11 +873,25 @@ func (dcr *ExchangeWallet) Balance() (*asset.Balance, error) {
...
+		if err != nil {
+			dcr.log.Errorf("Failed to get fee rate from external API: %v", err)
+			return 0
+		}
+		uintFee, err := convertFeeToUint(estimatedFeeRate)
+		if err != nil {
+			dcr.log.Errorf("Failed to convert fee rate: %v", err)
+			return 0
+		}
+		return uintFee
```
buck54321, line 891: What do you think of this refactor to reduce duplication?

```Go
func (dcr *ExchangeWallet) FeeRate() uint64 {
	const confTarget = 2
	rate, err := dcr.feeRate(confTarget)
	if err != nil {
		dcr.log.Errorf("feeRate error: %v", err)
	}
	return rate
}

// FeeRate returns the current optimal fee rate in atoms / byte.
func (dcr *ExchangeWallet) feeRate(confTarget uint64) (uint64, error) {
	// estimatesmartfee 1 returns extremely high rates on DCR.
	if confTarget < 2 {
		confTarget = 2
	}
	if feeEstimator, is := dcr.wallet.(FeeRateEstimator); is {
		ratePerKB, err := feeEstimator.EstimateSmartFeeRate(dcr.ctx, int64(confTarget), chainjson.EstimateSmartFeeConservative)
		if err == nil {
			return convertFeeToUint(ratePerKB)
		} else {
			dcr.log.Errorf("Failed to get fee rate with estimate smart fee rate: %v", err)
		}
	}
	// Either SPV wallet or EstimateSmartFeeRate failed.
	if !dcr.apiFeeFallback {
		return 0, fmt.Errorf("fee rate estimation unavailable")
	}
	dcr.log.Debug("Retrieving fee rate from external API")
	ratePerKB, err := externalFeeEstimator(dcr.ctx, dcr.network, confTarget)
	if err != nil {
		dcr.log.Errorf("Failed to get fee rate from external API: %v", err)
		return 0, err
	}
	return convertFeeToUint(ratePerKB)
}
```

buck54321 (approved):

chappjc (approved):

client/asset/dcr/dcr.go
```diff
@@ -850,32 +873,42 @@ func (dcr *ExchangeWallet) Balance() (*asset.Balance, error) {
...
+		} else {
+			dcr.log.Errorf("Failed to get fee rate with estimate smart fee rate: %v", err)
+		}
+	}
+	// Either SPV wallet or EstimateSmartFeeRate failed.
+	if !dcr.apiFeeFallback {
+		return 0, fmt.Errorf("fee rate estimation unavailable")
+	}
+	dcr.log.Debug("Retrieving fee rate from external API")
+	ratePerKB, err := externalFeeEstimator(dcr.ctx, dcr.network, confTarget)
```
chappjc, line 903: I'm just realizing that we're not doing any sanity checking or thresholding on the result from the external source.  They could be malicious or maybe they changed the API from DCR/kB -> atoms/kB (or whatever) without warning.
The external source might also say 0.

I think after calling `externalFeeEstimator` we need to both catch zero and filter the result of `convertFeeToUint` through `feeRateWithFallback`

client/asset/dcr/dcr.go
```diff
@@ -885,6 +918,39 @@ func (dcr *ExchangeWallet) feeRate(confTarget uint64) (uint64, error) {
...
+	if err != nil {
+		return 0, err
+	}
+	estimatedFeeRate, ok := c[nb]
+	if !ok {
+		return 0, errors.New("no fee rate for requested number of blocks")
+	}
+	return estimatedFeeRate, nil
+}
+
```
chappjc, line 953: A few lines down there's still a `if dcr.wallet.SpvMode() {` that skips calling `feeRate` where the external API would be called.  Thus we're still falling back in a lot of cases.

chappjc (changes requested):
Took a closer look and I think we've neglected two things.

client/asset/dcr/dcr.go
vctt94: lol I completely missed it. Removed it now, thanks for catching.

client/asset/dcr/dcr.go
vctt94: Yeah, you are right, it was returning 1, which would not go into `feeRateWithFallback`. Returning 0 now if it is 0.

I am not checking for a upper limit though, should I?

vctt94: sounds reasonable. Checking if it is greater than  `dcr.feeRateLimit` now

chappjc (approved):

PR #1657, opened 2022-06-09
client: fix native btc wallet creation with fee limit smaller than 1sat/B

### LINKED ISSUE #1609: btc: creating native wallet with too-low fee limit wrecks the wallet
image

```
[INF] CORE: Initializing a built-in btc wallet
[TRC] CORE[btc]: Fallback fees set at 100 Sats/vbyte
[ERR] WEB: error creating btc wallet: error loading wallet for 0 -> btc: error opening wallet: fee rate limit is smaller than smallest unit: 2e-06
"POST http://127.0.0.3:5758/api/newwallet HTTP/1.1" from 127.0.0.1:58922 - 200 73B in 1.888988083s
```

Then when trying again with a higher limit:

image

`[ERR] WEB: error creating btc wallet: Error creating wallet: CreateNewWallet error: wallet already exists`

vctt94:
closes #1609

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -446,14 +446,6 @@ type WalletConfig struct {
 	Birthday         uint64  `ini:"walletbirthday"`       // SPV
 }
 
-// baseWalletConfig is the config params for the base wallet.
-type baseWalletConfig struct {
-	fallbackFeeRate  uint64 // atoms/byte
-	feeRateLimit     uint64 // atoms/byte
-	redeemConfTarget uint64
-	useSplitTx       bool
-}
-
 // adjustedBirthday converts WalletConfig.Birthday to a time.Time, and adjusts
 // it so that defaultWalletBirthday <= WalletConfig.Bithday <= now.
 func (cfg *WalletConfig) adjustedBirthday() time.Time {
@@ -607,6 +599,14 @@ func init() {
 	asset.Register(BipID, &Driver{})
 }
 
+// baseWalletConfig is the config params for the base wallet.
+type baseWalletConfig struct {
+	fallbackFeeRate  uint64 // atoms/byte
+	feeRateLimit     uint64 // atoms/byte
+	redeemConfTarget uint64
+	useSplitTx       bool
+}
+
 // baseWallet is a wallet backend for Bitcoin. The backend is how the DEX
 // client app communicates with the BTC blockchain and wallet. baseWallet
 // satisfies the dex.Wallet interface.
@@ -623,7 +623,6 @@ type baseWallet struct {
 	lastPeerCount     uint32
 	peersChange       func(uint32)
 	minNetworkVersion uint64
-	config            *baseWalletConfig
 	dustLimit         uint64
 	useLegacyBalance  bool
 	zecStyleBalance   bool
@@ -637,6 +636,7 @@ type baseWallet struct {
 	hashTx            func(*wire.MsgTx) *chainhash.Hash
 	stringAddr        dexbtc.AddressStringer
 	txVersion         func() int32
+	*baseWalletConfig
 
 	tipMtx     sync.RWMutex
 	currentTip *block
@@ -886,7 +886,7 @@ func newRPCWallet(requester RawRequesterWithContext, cfg *BTCCloneCFG, walletCon
 }
 
 func readBaseWalletConfig(walletCfg *WalletConfig) (*baseWalletConfig, error) {
-	baseWallet := &baseWalletConfig{}
+	cfg := &baseWalletConfig{}
 	// if values not specified, use defaults. As they are validated as BTC/KB,
 	// we need to convert first.
 	if walletCfg.FallbackFeeRate == 0 {
@@ -900,30 +900,27 @@ func readBaseWalletConfig(walletCfg *WalletConfig) (*baseWalletConfig, error) {
 	}
 	// If set in the user config, the fallback fee will be in conventional units
 	// per kB, e.g. BTC/kB. Translate that to sats/byte.
-	baseWallet.fallbackFeeRate = toSatoshi(walletCfg.FallbackFeeRate / 1000)
-	if baseWallet.fallbackFeeRate == 0 {
-		return nil, fmt.Errorf("fallback fee rate limit is smaller than smallest unit(1e-5 coins/kB): %v",
+	cfg.fallbackFeeRate = toSatoshi(walletCfg.FallbackFeeRate / 1000)
+	if cfg.fallbackFeeRate == 0 {
+		return nil, fmt.Errorf("fallback fee rate limit is smaller than the minimum 1000 sats/byte: %v",
 			walletCfg.FallbackFeeRate)
 	}
 	// If set in the user config, the fee rate limit will be in units of BTC/KB.
 	// Convert to sats/byte & error if value is smaller than smallest unit.
-	baseWallet.feeRateLimit = toSatoshi(walletCfg.FeeRateLimit / 1000)
-	if baseWallet.feeRateLimit == 0 {
-		return nil, fmt.Errorf("fee rate limit is smaller than smallest unit(1e-5 coins/kB): %v",
+	cfg.feeRateLimit = toSatoshi(walletCfg.FeeRateLimit / 1000)
+	if cfg.feeRateLimit == 0 {
+		return nil, fmt.Errorf("fee rate limit is smaller than the minimum 1000 sats/byte: %v",
 			walletCfg.FeeRateLimit)
 	}
-	baseWallet.redeemConfTarget = walletCfg.RedeemConfTarget
-	if walletCfg.RedeemConfTarget == 0 {
-		return nil, fmt.Errorf("redeem conf target can not be 0: %v",
-			walletCfg.RedeemConfTarget)
-	}
-	baseWallet.useSplitTx = walletCfg.UseSplitTx
 
-	return baseWallet, nil
+	cfg.redeemConfTarget = walletCfg.RedeemConfTarget
+	cfg.useSplitTx = walletCfg.UseSplitTx
+
+	return cfg, nil
 }
 
 func newUnconnectedWallet(cfg *BTCCloneCFG, walletCfg *WalletConfig) (*baseWallet, error) {
-	baseWalletConfig, err := readBaseWalletConfig(walletCfg)
+	baseCfg, err := readBaseWalletConfig(walletCfg)
 	if err != nil {
 		return nil, err
 	}
@@ -977,22 +974,21 @@ func newUnconnectedWallet(cfg *BTCCloneCFG, walletCfg *WalletConfig) (*baseWalle
 		fundingCoins:        make(map[outPoint]*utxo),
 		findRedemptionQueue: make(map[outPoint]*findRedemptionReq),
 		minNetworkVersion:   cfg.MinNetworkVersion,
-		config:              baseWalletConfig,
 		dustLimit:           cfg.ConstantDustLimit,
-
-		useLegacyBalance: cfg.LegacyBalance,
-		zecStyleBalance:  cfg.ZECStyleBalance,
-		segwit:           cfg.Segwit,
-		signNonSegwit:    nonSegwitSigner,
-		estimateFee:      cfg.FeeEstimator,
-		decodeAddr:       addrDecoder,
-		stringAddr:       addrStringer,
-		walletInfo:       cfg.WalletInfo,
-		deserializeTx:    txDeserializer,
-		serializeTx:      txSerializer,
-		hashTx:           txHasher,
-		calcTxSize:       txSizeCalculator,
-		txVersion:        txVersion,
+		useLegacyBalance:    cfg.LegacyBalance,
+		zecStyleBalance:     cfg.ZECStyleBalance,
+		segwit:              cfg.Segwit,
+		signNonSegwit:       nonSegwitSigner,
+		estimateFee:         cfg.FeeEstimator,
+		decodeAddr:          addrDecoder,
+		stringAddr:          addrStringer,
+		walletInfo:          cfg.WalletInfo,
+		deserializeTx:       txDeserializer,
+		serializeTx:         txSerializer,
+		hashTx:              txHasher,
+		calcTxSize:          txSizeCalculator,
+		txVersion:           txVersion,
+		baseWalletConfig:    baseCfg,
 	}
 
 	if w.estimateFee == nil {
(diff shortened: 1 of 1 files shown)


chappjc:
Hmm there a whole bunch of walletconf checks in unconnectedWallet that I figure should be shared.  Maybe we make a shared function for the common stuff

vctt94:
I agree it shouldn't be possible to create a wallet which can not be opened. 

Now I am creating a method `validateBtcWalletConf(walletCfg *WalletConfig)` which I use to compare the fee rate limit, fallback fee rate, and conf target.

When the validation fails on `newUnconnectedWallet`, I set the default values to keep old behavior. 

I also added two new params into WalletConfig struct, `FeeRateLimitPerByte` and `FallbackFeeRatePerByte` so we can pass them to the baseWallet, after creating a new `baseWallet`.

client/asset/btc/btc.go
```diff
@@ -875,36 +881,46 @@ func newRPCWallet(requester RawRequesterWithContext, cfg *BTCCloneCFG, walletCon
...
+func newUnconnectedWallet(cfg *BTCCloneCFG, walletCfg *WalletConfig) (*baseWallet, error) {
+	err := validateBtcWalletConf(walletCfg)
+	if err != nil {
+		cfg.Logger.Tracef(err.Error())
+		// check which value failed, and add the default value to it
+		if walletCfg.FallbackFeeRatePerByte == 0 {
+			walletCfg.FallbackFeeRatePerByte = cfg.DefaultFallbackFee
+		}
+		if walletCfg.FeeRateLimitPerByte == 0 {
+			walletCfg.FeeRateLimitPerByte = cfg.DefaultFeeRateLimit
```
chappjc, line 918: If the previous value, `FallbackFeeRate`, failed validation, then `validateBtcWalletConf` would have returned before it set `walletCfg.FeeRateLimitPerByte`, which may have been valid, but it would go with the default instead.

I suggest were rework this a little, making `validateBtcWalletConf` not mutate the struct, and only applying the defaults to the fields that failed validation.

chappjc (changes requested):

client/asset/btc/btc.go
vctt94: I removed the early return from the validate method. 

I also noticed this values being re-applied on error were actually doing nothing, as they would be overwritten by the forms input, so I removed them.

vctt94: oh, actually they need to be overwritten on error, otherwise tests fail. 

Okay, undoing that

client/asset/btc/btc.go
```diff
@@ -875,36 +881,46 @@ func newRPCWallet(requester RawRequesterWithContext, cfg *BTCCloneCFG, walletCon
...
 }
 
-func newUnconnectedWallet(cfg *BTCCloneCFG, walletCfg *WalletConfig) (*baseWallet, error) {
+func validateBtcWalletConf(walletCfg *WalletConfig) error {
+	var err error
 	// If set in the user config, the fallback fee will be in conventional units
 	// per kB, e.g. BTC/kB. Translate that to sats/byte.
-	fallbackFeesPerByte := toSatoshi(walletCfg.FallbackFeeRate / 1000)
-	if fallbackFeesPerByte == 0 {
-		fallbackFeesPerByte = cfg.DefaultFallbackFee
```
buck54321, line 883: Isn't the only thing we need to do is to check that 

```Go
if walletCfg.FallbackFeeRate > 0 && walletCfg.FallbackFeeRate < 1000 {
    return nil, fmt.Errorf("minimum fee rate is 1000 sats/kB")
}
```

vctt94: If we just check for `FallbackFeeRate` it will still be possible to create a wallet with "invalid" values, for example: 

someone tries to create a wallet with Highest acceptable fee rate less than the minimum value allowed (anything less than 1 sat/B), would create the wallet directories and db, but the form would throw an error making a wallet which can not be opened, like described in the issue.

Also, with `toSatoshi()` It rounds for the minimum value possible, so we just need to check if it is not 0, considering we are not checking for a maximum value.

JoeGruffins (commented):
Solves the issue. I think should error more though.

client/asset/btc/btc.go
```diff
@@ -437,13 +437,15 @@ type RPCWalletConfig struct {
...
-	ActivelyUsed     bool    `ini:"special:activelyUsed"` //injected by core
-	WalletName       string  `ini:"walletname"`           // RPC
-	Birthday         uint64  `ini:"walletbirthday"`       // SPV
+	UseSplitTx             bool    `ini:"txsplit"`
+	FallbackFeeRate        float64 `ini:"fallbackfee"`
+	FeeRateLimit           float64 `ini:"feeratelimit"`
+	FeeRateLimitPerByte    uint64
+	FallbackFeeRatePerByte uint64
+	RedeemConfTarget       uint64 `ini:"redeemconftarget"`
+	ActivelyUsed           bool   `ini:"special:activelyUsed"` //injected by core
```
JoeGruffins, line 446: nit: space missing in comment

client/asset/btc/btc.go
vctt94: > If we want default values if values not set can we set those first and then do validate and error if that errors?

What do you mean by set those value first? Right now the default values are populated from the `configOpts` and if changing its value on the form, it will not be the default values anymore.

> Isn't it correct to return error if values are wrong here? This happens when opening an existing wallet correct? We don't want to silently ignore what the user thinks they set, overwrite it and use some values they don't want?

In the UI an error will be thrown, when creating the wallet, but I agree with you about returning the error, the reason I am not returning it, is to keep the old behavior: 

Right now on the walletTypeSPV test it passes an empty wallet config struct, so it will fail if we return the errors here

```
	case walletTypeSPV:
		w, err := newUnconnectedWallet(cfg, &WalletConfig{})
		if err == nil {
			wallet = &ExchangeWalletFullNode{w}
			neutrinoClient := &tNeutrinoClient{data}
			wallet.node = &spvWallet{
				chainParams: &chaincfg.MainNetParams,
				wallet:      &tBtcWallet{data},
				cl:          neutrinoClient,
				tipChan:     make(chan *block, 1),
				chainClient: nil,
				acctNum:     0,
				txBlocks:    data.dbBlockForTx,
				checkpoints: data.checkpoints,
				log:         cfg.Logger.SubLogger("SPV"),
				loader:      nil,
			}
		}
```

client/asset/btc/btc.go
chappjc: Notice that in @buck54321's example code it permits 0, allowing the default to be applied elsewhere (caller).

The validator function can still check each of the settings that can result in an unopenable wallet.

vctt94: So it does not set values on error, it sets if values are not specified. 

Thanks for clarifying chapp, I had misunderstood this. I fixed it now.

client/asset/btc/btc.go
JoeGruffins: This looks correct now. But I don't think we need the trace log as it will log as error later.

client/asset/btc/btc.go
```diff
@@ -875,36 +881,48 @@ func newRPCWallet(requester RawRequesterWithContext, cfg *BTCCloneCFG, walletCon
...
+func validateBtcWalletConf(walletCfg *WalletConfig) error {
+	var err error
 	// If set in the user config, the fallback fee will be in conventional units
 	// per kB, e.g. BTC/kB. Translate that to sats/byte.
-	fallbackFeesPerByte := toSatoshi(walletCfg.FallbackFeeRate / 1000)
-	if fallbackFeesPerByte == 0 {
-		fallbackFeesPerByte = cfg.DefaultFallbackFee
+	walletCfg.FallbackFeeRatePerByte = toSatoshi(walletCfg.FallbackFeeRate / 1000)
+	if walletCfg.FallbackFeeRatePerByte == 0 {
+		err = fmt.Errorf("fallback fee rate limit is smaller than smallest unit(1 sat/byte): %v",
```
JoeGruffins, line 890: While true, the form has values of coin/kb. The error is also like this and in scientific notation. So maybe `1e-5 coin/kb`?

client/asset/btc/btc.go
```diff
@@ -875,36 +881,46 @@ func newRPCWallet(requester RawRequesterWithContext, cfg *BTCCloneCFG, walletCon
...
+	var err error
 	// If set in the user config, the fallback fee will be in conventional units
 	// per kB, e.g. BTC/kB. Translate that to sats/byte.
-	fallbackFeesPerByte := toSatoshi(walletCfg.FallbackFeeRate / 1000)
-	if fallbackFeesPerByte == 0 {
-		fallbackFeesPerByte = cfg.DefaultFallbackFee
+	walletCfg.FallbackFeeRatePerByte = toSatoshi(walletCfg.FallbackFeeRate / 1000)
+	if walletCfg.FallbackFeeRatePerByte == 0 {
+		err = fmt.Errorf("fallback fee rate limit is smaller than smallest unit: %v",
+			walletCfg.FallbackFeeRate)
```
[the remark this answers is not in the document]
JoeGruffins: This is still not returning on error. It's not wrong, just not standard.

client/asset/btc/btc.go
buck54321: > Isn't the only thing we need to do is to check that
> 
> ```go
> if walletCfg.FallbackFeeRate > 0 && walletCfg.FallbackFeeRate < 1000 {
>     return nil, fmt.Errorf("minimum fee rate is 1000 sats/kB")
> }
> ```

I'm saying this entire PR should be just these three lines. You didn't need to touch `WalletConfig` or create any new functions. We already had a pattern established for validating input values, We just weren't checking this condition.

JoeGruffins (approved):

client/asset/btc/btc.go
vctt94: okay, but the problem is actually on the `Create` method, which does not validate the spv wallet allowing it to create a wallet which can not be opened. Should I copy this validations over that method?

buck54321: I see what you mean. It wasn't failing on creation but failed on open, which tanked the creation process. I guess wallets need to basically guarantee that the settings given to `Create` are valid. 

We should also re-consider the sequence in `CreateWallet` that allows this to happen. Should we somehow delete the wallet if the full `CreateWallet` sequence doesn't complete? Or should we store the wallet (in `db` and `wallets` map) right after `Create`, so users could have recovered by adjusting configuration settings.

client/asset/btc/btc.go
```diff
@@ -875,36 +881,46 @@ func newRPCWallet(requester RawRequesterWithContext, cfg *BTCCloneCFG, walletCon
...
+	// we need to convert first.
+	if walletCfg.FallbackFeeRate == 0 {
+		walletCfg.FallbackFeeRate = float64(cfg.DefaultFallbackFee) * 1000 / 1e8
+	}
+	if walletCfg.FeeRateLimit == 0 {
+		walletCfg.FeeRateLimit = float64(cfg.DefaultFeeRateLimit) * 1000 / 1e8
+	}
+	if walletCfg.RedeemConfTarget == 0 {
+		walletCfg.RedeemConfTarget = defaultRedeemConfTarget
+	}
```
buck54321, line 919: If you're going to assign defaults here, you'll need want to do this in `Create` also, no? Maybe just move it in to `validateBtcWalletConf`.

client/asset/btc/btc.go
```diff
@@ -437,13 +437,15 @@ type RPCWalletConfig struct {
...
-	Birthday         uint64  `ini:"walletbirthday"`       // SPV
+	UseSplitTx             bool    `ini:"txsplit"`
+	FallbackFeeRate        float64 `ini:"fallbackfee"`
+	FeeRateLimit           float64 `ini:"feeratelimit"`
+	FeeRateLimitPerByte    uint64
+	FallbackFeeRatePerByte uint64
+	RedeemConfTarget       uint64 `ini:"redeemconftarget"`
+	ActivelyUsed           bool   `ini:"special:activelyUsed"` // injected by core
+	WalletName             string `ini:"walletname"`           // RPC
+	Birthday               uint64 `ini:"walletbirthday"`       // SPV
```
buck54321, line 448: Rather than `WalletConfig` pulling double duty, I suggest having a second struct

```Go
type baseWalletConfig struct {
	fallbackFeeRate  uint64 // atoms/byte
	feeRateLimit     uint64 // atoms/byte
	redeemConfTarget uint64
	useSplitTx       bool
}
```

and modifying `validateBtcWalletConf` to be something like...

```Go
func readBaseWalletConfig(walletCfg *WalletConfig) (*baseWalletConfig, error)
```

Set the defaults for zeros inside. Then you can embed `baseWalletConfig` in `baseWallet`.

```Go
type baseWallet {
    baseWalletConfig
    ...
}
```

and delete the existing fields.

buck54321 (changes requested):

client/asset/btc/btc.go
chappjc: > We should also re-consider the sequence in `CreateWallet` that allows this to happen. Should we somehow delete the wallet if the full `CreateWallet` sequence doesn't complete?

If in `(*Core).CreateWallet` it creates the wallet fine but then opening it fails, I agree it would be ideal to "uncreate" the wallet.  I think that can be handled **if** the wallet is a `Recoverer` that allows it to be destroyed.  We can follow-up with that I think.  Ideally the `Move` method that started out as `Destroy` if I recall correctly, would take an empty string to signal deletion instead of backup.

client/asset/btc/btc.go
```diff
@@ -441,11 +441,19 @@ type WalletConfig struct {
...
 	FeeRateLimit     float64 `ini:"feeratelimit"`
 	RedeemConfTarget uint64  `ini:"redeemconftarget"`
-	ActivelyUsed     bool    `ini:"special:activelyUsed"` //injected by core
+	ActivelyUsed     bool    `ini:"special:activelyUsed"` // injected by core
 	WalletName       string  `ini:"walletname"`           // RPC
 	Birthday         uint64  `ini:"walletbirthday"`       // SPV
 }
 
+// baseWalletConfig is the config params for the base wallet.
+type baseWalletConfig struct {
```
buck54321, line 450: Could you move this down to right above the `baseWallet` definition.

client/asset/btc/btc.go
```diff
@@ -611,11 +623,8 @@ type baseWallet struct {
 	lastPeerCount     uint32
 	peersChange       func(uint32)
 	minNetworkVersion uint64
-	fallbackFeeRate   uint64
-	feeRateLimit      uint64
+	config            *baseWalletConfig
```
buck54321, line 626: Could you please embed this so you don't have to add `.config.` everywhere. It'll actually save me a bunch of conflicts in #1686 if you do.

client/asset/btc/btc.go
```diff
@@ -876,36 +885,48 @@ func newRPCWallet(requester RawRequesterWithContext, cfg *BTCCloneCFG, walletCon
...
+		walletCfg.RedeemConfTarget = defaultRedeemConfTarget
+	}
 	// If set in the user config, the fallback fee will be in conventional units
 	// per kB, e.g. BTC/kB. Translate that to sats/byte.
-	fallbackFeesPerByte := toSatoshi(walletCfg.FallbackFeeRate / 1000)
-	if fallbackFeesPerByte == 0 {
-		fallbackFeesPerByte = cfg.DefaultFallbackFee
+	baseWallet.fallbackFeeRate = toSatoshi(walletCfg.FallbackFeeRate / 1000)
+	if baseWallet.fallbackFeeRate == 0 {
+		return nil, fmt.Errorf("fallback fee rate limit is smaller than smallest unit(1e-5 coins/kB): %v",
```
buck54321, line 905:
```suggestion
		return nil, fmt.Errorf("fallback fee rate limit is smaller than the minimum 1000 sats/byte: %v",
```

buck54321 (changes requested):

client/asset/btc/btc.go
```diff
@@ -876,36 +885,45 @@ func newRPCWallet(requester RawRequesterWithContext, cfg *BTCCloneCFG, walletCon
...
-	if redeemConfTarget == 0 {
-		redeemConfTarget = defaultRedeemConfTarget
+	cfg.redeemConfTarget = walletCfg.RedeemConfTarget
+	cfg.useSplitTx = walletCfg.UseSplitTx
+
+	return cfg, nil
+}
+
+func newUnconnectedWallet(cfg *BTCCloneCFG, walletCfg *WalletConfig) (*baseWallet, error) {
+	baseWalletConfig, err := readBaseWalletConfig(walletCfg)
```
buck54321, line 923: You're variable name is shadowing the type name. How about just `baseCfg`?

buck54321 (approved):

PR #1658, opened 2022-06-10
dex/testing/btc: do not create descriptor wallets

chappjc:
```
The default wallet type with v23 of Bitcoin Core using the createwallet
RPC is a "descriptor" wallet, which does not work with DEX.

This updates the harness script to specify descriptors=false with the
named form of cli commands, which seem to be available since at least
0.19.  The latest litecoin is also good with this and will also switch
to named arguments with 0.21.  For now, just BTC.
```

@JoeGruffins I ran into this when messing with loadbot using bitcoind v23.
I see in https://github.com/decred/dcrdex/pull/1656 you are removing the new-wallet shell script, but I'm not sure what your plan is there yet since BTC needs `createwallet` while DCR needs `createnewaccount`, plus as we see now the args can be different between Bitcoin clones.

This PR is required until https://github.com/decred/dcrdex/pull/1659 is done

### DIFF
--- a/dex/testing/btc/base-harness.sh
+++ b/dex/testing/btc/base-harness.sh
@@ -168,10 +168,17 @@ sleep 2
 EOF
 chmod +x "./reorg"
 
+if [ "$LONG_CREATEWALLET" ] ; then
 cat > "./new-wallet" <<EOF
 #!/usr/bin/env bash
-./\$1 createwallet \$2
+./\$1 -named createwallet wallet_name=\$2 descriptors=false
 EOF
+else
+cat > "./new-wallet" <<EOF
+#!/usr/bin/env bash
+./\$1 createwallet \$
+EOF
+fi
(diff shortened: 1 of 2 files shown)


martonp (approved):
LGTM, although when I run the old harness script from master with bitcond 23.0 and bitcoin-cli 23.0, it's still creating a wallet without descriptors...

martonp:
Yes, it works after deleting the tarball.

JoeGruffins (approved):
> I see in https://github.com/decred/dcrdex/pull/1656 you are removing the new-wallet shell script

I can update #1656 to use this. It's doing the same thing now, just I'm setting the whole create string in the `harness.sh` portion rather than a switch.

chappjc:
Taking a different approach with #1656 merged and https://github.com/decred/dcrdex/pull/1658 going to create a descriptor wallet.

PR #1660, opened 2022-06-11
dexc/ui: show app seed as words from pgpwordlist

LasTshaMAN:
Following up on https://github.com/decred/dcrdex/pull/1648#discussion_r892762495 this is a basic implementation for handling app seed as seed phrase (instead of binary string), this PR:
- adds ability to export app seed as seed phrase (with words from `pgpwordlist` used in Decred ecosystem)
- removes ability to export app seed in binary form in UI (I guess that's no longer needed, no need to clutter UI), server API still returns seed along with seed phrase
- adds ability to import app seed as seed phrase (tested on testnet, works as I would expect), importing binary seed also still works (left this feature intact for backward-compatibility)

Exporting seed phrase:
image

Importing seed phrase:
image
image

Also updated Docker file a bit (to support running with `testnet`, for example), and instructions on how to run it.


LasTshaMAN:
Unrelated to this PR, but I've noticed that [webserver API unwraps error](https://github.com/decred/dcrdex/blob/649e91c5bf1356ba1f85fe2fc180ad5e0ecc14e5/client/webserver/api.go#L1026-L1031), and I'm not sure why,

to get the full error stack trace I have to look into `dexc` logs:
```
2022-06-11 07:26:53.756 [ERR] WEB: seed phrase can't be decoded: walletseed.DecodeUserInput: invalid encoding: encoding/hex: invalid byte: U+006B 'k'
```
while in UI we only get:
image

also, looks like [rpcserver doesn't do that](https://github.com/decred/dcrdex/blob/649e91c5bf1356ba1f85fe2fc180ad5e0ecc14e5/client/webserver/api.go#L1026-L1031),

so, should this unwrapping be gotten rid of ?

client/webserver/locales/zh-cn.go
```diff
@@ -162,7 +162,9 @@ var ZhCN = map[string]string{
 	"Balance":                "余额", // alt. 平衡
 	"Actions":                "操作", // alt. 动作
 	"Restoration Seed":       "恢复种子",
+	"Restoration Seed Phrase": "Restoration Seed Phrase",
 	"Restore from seed":      "从种子恢复",
+	"Restore from seed phrase": "Restore from seed phrase",
```
LasTshaMAN, line 167: I didn't update/added translations, should I use `google translate` or invite someone to check it out ?

seems like overkill doing it every pull request, maybe better do it in batches (every so often invite someone to update translations) ?

client/webserver/locales/en-us.go
```diff
@@ -158,12 +158,14 @@ var EnUS = map[string]string{
 	"seed_implore_msg":            "You should carefully write down your application seed and save a copy. Should you lose access to this machine or the critical application files, the seed can be used to restore your DEX accounts and native wallets. Some older accounts cannot be restored from seed, and whether old or new, it's good practice to backup the account keys separately from the seed.",
```
LasTshaMAN, line 158: > Some older accounts cannot be restored from seed, and whether old or new, it's good practice to backup the account keys separately from the seed.

I wonder whether this part is still relevant ? Or maybe this warning has already served its purpose and everyone migrated their accounts ?

And, I'm not sure what `account keys` mentioned here are (is it also an old artifact new users shouldn't worry about ?)

my main concern is - the simpler (and shorter) the text is, the better UX is for the user

chappjc: I don't think the message ever made much sense.  The seed gives your account keys, which are is the key pair that uniquely identifies you to the server.  The seed is all anyone needs, so why is is good practice to backup/export every DEX account?  I don't think it is.

chappjc:
> Unrelated to this PR, but I've noticed that [webserver API unwraps error](https://github.com/decred/dcrdex/blob/649e91c5bf1356ba1f85fe2fc180ad5e0ecc14e5/client/webserver/api.go#L1026-L1031), and I'm not sure why,
> 
> to get the full error stack trace I have to look into `dexc` logs:
> 
> ```
> 2022-06-11 07:26:53.756 [ERR] WEB: seed phrase can't be decoded: walletseed.DecodeUserInput: invalid encoding: encoding/hex: invalid byte: U+006B 'k'
> ```
> 
> while in UI we only get: image
> 
> also, looks like [rpcserver doesn't do that](https://github.com/decred/dcrdex/blob/649e91c5bf1356ba1f85fe2fc180ad5e0ecc14e5/client/webserver/api.go#L1026-L1031),
> 
> so, should this unwrapping be gotten rid of ?

Follow the `git blame` for that line: https://github.com/decred/dcrdex/pull/1511
Meant to address https://github.com/decred/dcrdex/issues/1488

In most cases, the inner most error is what sheds light on the actual problem. e.g. `Incorrect Password` vs. `error unlocking btc wallet: ReconfigureWallet password error: outer key deserialization error: incorrect password`

In this seed error, what would be most helpful? The unwrapped error drills down to the problem: `k` is not a valid hex character.  IF the full message were displayed, what would it add?  

I think the solution is to tweak _what_ errors are and _where_ they are wrapped with `%w` to ensure the most helpful message makes it to the frontend, not to give the user the full message with layers of extra context to sift through.

client/webserver/locales/en-us.go
LasTshaMAN: okay, I've slightly updated texts for `seed_implore_msg` and `pw_for_seed` to address that,

let me know if you want different phrasing for these

LasTshaMAN:
> In this seed error, what would be most helpful? The unwrapped error drills down to the problem: k is not a valid hex character. IF the full message were displayed, what would it add?

For this particular case, I think `encoding/hex: invalid byte: U+006B 'k'` is way too scary for ordinary user, showing `seed phrase can't be decoded` instead is probably better way to handle this

> I think the solution is to tweak what errors are and where they are wrapped ...

Makes sense, I'm just used to slightly different way of mapping API errors, but core idea is pretty much the same, so I'll adjust

chappjc:
> For this particular case, I think `encoding/hex: invalid byte: U+006B 'k'` is way too scary for ordinary user, showing `seed phrase can't be decoded` instead is probably better way to handle this

Possibly.  We can just not wrap the low level error and instead wrap it only higher in the stack.  Or we can replace the error with something more user friendly.  Personally, if it said "seed phrase can't be decoded" to me, I'd be annoyed that it doesn't say why.    While `encoding/hex: invalid byte: U+006B 'k'` might be scary, it says exactly why.

>  Makes sense, I'm just used to slightly different way of mapping API errors, but core idea is pretty much the same, so I'll adjust

Open to better ideas.  Please expand on how it would be best to do this.  I think we've just been lazy about passing the errors up.

client/webserver/locales/en-us.go
```diff
@@ -155,10 +155,10 @@ var EnUS = map[string]string{
 	"New Wallet Password":         "New Wallet Password",
 	"pw_change_warn":              "Note: Changing to a different wallet while having active trades might cause funds to be lost.",
 	"Show more options":           "Show more options",
-	"seed_implore_msg":            "You should carefully write down your application seed and save a copy. Should you lose access to this machine or the critical application files, the seed can be used to restore your DEX accounts and native wallets. Some older accounts cannot be restored from seed, and whether old or new, it's good practice to backup the account keys separately from the seed.",
+	"seed_implore_msg":            "Write down your application seed phrase and make a copy.",
 	"View Application Seed":       "View Application Seed",
 	"Remember my password":        "Remember my password",
-	"pw_for_seed":                 "Enter your app password to show your seed phrase. Make sure nobody else can see your screen.",
+	"pw_for_seed":                 "Application seed can be used to restore your DEX accounts and native wallets. Make sure nobody else can see your screen.",
```
chappjc, line 161: We now we have a untranslated text that's going to take forever to get retranslated.  This `pw_for_seed` message was fine, so please revert.

All that needed to be done, if anything, was to remove from `seed_implore_msg` the part: `and whether old or new, it's good practice to backup the account keys separately from the seed` part of the other message.`
Ideally do that without breaking any of the translations please.

client/Dockerfile
```diff
@@ -1,15 +1,6 @@
-#
-# Build the docker image
-# $ docker build -t user/dcrdex -f client/Dockerfile .
-#
-# Create docker volume to store client data
-# $ docker volume create --name=dcrdex_data
```
chappjc, line 6: Why remove the build and volume create commands?  Were they incorrect?

client/webserver/locales/zh-cn.go
chappjc: > seems like overkill doing it every pull request, maybe better do it in batches (every so often invite someone to update translations) ?

Yes, that's what I said previously.  This should not hold up dev work.

LasTshaMAN:
> Please expand on how it would be best to do this. I think we've just been lazy about passing the errors up. (e.g. If hex.DecodeString gave an error, we could replace the details with "not a hexadecimal string", but I still would like the exact reason)

> Removing the unwrapping to show the full error message would only make it more scary. Please suggest a solution that improves things.

I don't think there is a simple solution, maybe in ideal world every error needs to be mapped appropriately:
- some extended with additional context to interpret properly - in the example above  we have just a single argument, if we had many ... there would be no way for the user to tell which argument is wrong
- some replaced completely for a message crafted specifically for non-technical user
- some are fine as is

specific-enough and not-too-scary often lie on the opposite sides, the goal is probably to find right balance

A different approach to currently used (unwrapping to root cause) would be to define a limited set of errors that API user (or Application user) is supposed to understand (e.x. bad_argumet, unauthorized, ... internal_server_error) - lets call these **domain errors**; when error is encountered anywhere in the code, developer is supposed to map it to **domain erorr** (adding  necessary context),
that **domain erorr** is returned to API user (basically **unwrapping to root cause** behavior changes to **unwrapping to domain error**), while error stack trace is logged fully (like it currently is). 
This approach is pretty cumbersome and becomes messy if not strictly followed (its hard to enforce using it everywhere, and pick best places for errors to be converted into domain errors), so while its more user-oriented I wouldn't say its better and it likely isn't worth it (at least not until project development turns into mostly polishment).

client/Dockerfile
LasTshaMAN: removed in favor of this link - https://github.com/LasTshaMAN/dcrdex/blob/607675366f8116e01e05dd755883a4cec6940bc4/client/Dockerfile#L1-L1

thought it's not necessary to duplicate them in Docker file, and instructions themselves needed an update (after I've changed `CMD`->`ENTRYPOINT` existing `docker` commands need to pass arguments to `dexc` now)

chappjc: I had seen the readme change, but I guess I didn't understand what happened to the `volume create`.  Apologies, I don't use docker

chappjc:
> So, to sum up, I don't think current approach is bad/wrong, I've just never seen it before implemented this way, and wanted to ask about the reasoning (I need to use git history more often).
> 
> **edit**: actually the approach with **domain errors** I've described above is pretty much the same as **unwrapping to root cause** if necessary context gets added at the place of "root cause", so they are pretty much equivalent

I agree it's clear that just showing the root cause it's not good for all cases.  I've reopened the referenced issue.  Further discussion can continue there.

client/Dockerfile
LasTshaMAN: regarding `volume create`, 

it seems `-v dcrdex_data:/root/.dexc` covers volume creation in `docker run`.

client/webserver/locales/zh-cn.go
LasTshaMAN: so, I see 2 options here:
- leaving it English (until somebody can pick it up and translate properly)
- using google translate, and then someone will spot discrepancies and improve upon it

which approach do you usually follow ? I can see examples of [1st option being used](https://github.com/LasTshaMAN/dcrdex/blob/607675366f8116e01e05dd755883a4cec6940bc4/client/webserver/locales/pt-br.go#L32-L32), I guess this is the way to go

ukane-philemon: Yeah, just leave it English. You can remove it from this file and other locale except the **en-us.go** locale. `go generate` will use the default value English when preparing the HTML template files.

ukane-philemon: Yeah, just leave it English. You can remove it from this file and other locale except the **en-us.go** locale. `go generate` will use the default value English when preparing the HTML template files.

client/webserver/locales/en-us.go
LasTshaMAN: Okay, followed up on this in my latest commit to decrease breakage in existing translations

client/webserver/locales/zh-cn.go
LasTshaMAN: > Yeah, just leave it English. You can remove it from this file and other locale except the en-us.go locale.

applied this

martonp:
I think this would be better if when displaying the words they were numbered. The way it is now, it is convenient if you copy paste, but not convenient if you want to write it onto a piece of paper and then copy it back in. When entering the words there should be a separate box for each word, and if you misspell one of the words, it should let you know. Something similar to how they have in metamask:
Screen Shot 2022-06-14 at 1 05 35 PM

Also, I think this should be something that is shown to the user right after they initialize the client. Otherwise, I people will put off doing it, and eventually forget. it would be nice if there was another screen that helped the user double check that they copied the words down correctly.

LasTshaMAN:
> ... When entering the words there should be a separate box for each word ...

Agree, I was also thinking maybe we can re-use existing JS code from Decrediton (or maybe even Metamask/Keplr or something similar),
although I'm not that experienced with JS and not sure whether I'll be able to implement it reasonably fast, let me try to see what I can do about it and get back to you (we can also consider doing this improvement in a separate follow-up PR to not hold this one hanging here unmerged).

> Also, I think this should be something that is shown to the user right after they initialise the client. Otherwise, I people will put off doing it, and eventually forget. it would be nice if there was another screen that helped the user double check that they copied the words down correctly.

Also, agree, also will try adding this (could also be done as a follow-up PR though).

chappjc:
> > Also, I think this should be something that is shown to the user right after they initialise the client. Otherwise, I people will put off doing it, and eventually forget. it would be nice if there was another screen that helped the user double check that they copied the words down correctly.
> 
> Also, agree, also will try adding this (could also be done as a follow-up PR though).

Agree as well.  Definitely can be a follow-up PR

chappjc:
> Agree, I was also thinking maybe we can re-use existing JS code from Decrediton (or maybe even Metamask/Keplr or something similar),
> although I'm not that experienced with JS and not sure whether I'll be able to implement it reasonably fast, let me try to see what I can do about it and get back to you (we can also consider doing this improvement in a separate follow-up PR to not hold this one hanging here unmerged).

Which way are you going with this @LasTshaMAN?  Do avoid third party code though please.

And personally I'd be fire with a basic multi-line text box with a regular layout of the words.

If we converge on something that does not allow me to copy either the seed or the words somewhere, instead making me write them down, that's a non-starter for me.

LasTshaMAN:
> Which way are you going with this @LasTshaMAN? 
> Do avoid third party code though please.

I've checked Decrediton/Metamask/Keplr and looks like all of them are using some version of React, and I don't know how to convert it easily into plain JS yet (even if its possible).

I don't think the implementation for this feature needs any 3rd party packages, and given that we are managing seed-phrase here I understand how critical this is.

> For displaying the generated seed, if we converge on something that does not allow me to copy either the seed or the words somewhere, instead making me write them down, that's a non-starter for me.

yeah, I also thought that having an option to copy-paste seed (or phrase) is a useful feature to have (at least for testing), so it should be supported IMO.

> And personally I'd be fire with a basic multi-line text box with a regular layout of the words.

do you mean like already currently implemented in this PR ? Frankly, I think this PR does all that's necessary for now with respect to word-phrase, there are other things in DEX I'd rather be looking into and spending time on right now :)

chappjc:
> do you mean like already currently implemented in this PR ? Frankly, I think this PR does all that's necessary for now with respect to word-phrase, there are other things in DEX I'd rather be looking into and spending time on right now :)

Oh, I haven't retested.  Initially it looked like this:

image

All the words on one line, not visible.

raedah:
Its been discussed in the past that the pgp word list is less than ideal to use for the seed. The list was compiled as ideal words for speaking over a voice channel. "The words were carefully chosen for their phonetic distinctiveness". This looks like a great opportunity for a fresh start with a better selected word list.

chappjc:
@jrick Opinions about pgpwordlist and suitability of `DecodeMnemonics` and `ByteToMnemonic` https://github.com/decred/dcrwallet/blob/4ed5a4c677f9bdb1b8e70ceccf476fbff413cd79/pgpwordlist/pgpwordlist.go#L36
I wasn't aware of any fundamental issues with it.

LasTshaMAN:
> Its been discussed in the past that the pgp word list is less than ideal to use for the seed.

I thought about 24 word-phrase, shorter lenght is better user UX, but I see at least 2 potential (maybe minor) issues here:
- these days, 24 word-phrase pretty much = BIP39 + BIP32 + BIP44, given DEX supports managed(seeded) wallets users might expect that using DEX-24-word-phrase in Metamask, for example, should result into managed(seeded) Ethereum wallet import (so users can access the funds they have in DEX-managed Ethereum wallet via Metamask); this isn't yet implemented, but is possible to do so, and I think it might even improve UX
- most (if not all) Decred wallet apps/software uses pgp word list, and this inconsistency might bother people

jrick:
> @jrick Opinions about pgpwordlist and suitability of `DecodeMnemonics` and `ByteToMnemonic`? https://github.com/decred/dcrwallet/blob/4ed5a4c677f9bdb1b8e70ceccf476fbff413cd79/pgpwordlist/pgpwordlist.go#L36 I wasn't aware of any fundamental issues with it. This was recently proposed in dex channel: https://matrix.to/#/!mlRZqBtfWHrcmgdTWB:decred.org/$XukqwhoK9ICpwjVHW6VTFPrTOY68XkEZxUKjn9qEtmk?via=decred.org&via=matrix.org&via=planetdecred.org

as @raedah mentioned, pgpwordlist was designed to be unambiguous when spoken over a voice channel such as radio.  Obviously it can be used for other purposes as well like the dcrwallet seed, but other word lists can provide other and more useful advantages for seed encoding.  Off the top of my head, some of the missing properties from pgpwordlist that would be nice to have would be

* a larger word list that can encode the seed more compactly
* words chosen such that just typing (or even recording) the first few characters of each word is sufficient information to restore
* error correction (as opposed to just a single byte for error detection)

raedah:
Having a seed birthday encoded is useful as well (I believe lightning wallets have this feature), which can potentially be used to speed up initial syncs on associated wallets. Regarding inconsistency with other decred wallet software, the seed improvement would ideally get pushed to dcrwallet as a newly supported seed version. The current long 33 word seed is the most common complaint we get when onboarding new users to the decred wallet mobile apps.

matheusd:
> Having a seed birthday encoded is useful as well (I believe lightning wallets have this feature),

Correct. The [aezeed](https://github.com/lightningnetwork/lnd/tree/master/aezeed) seed scheme encodes+encrypts the seed  using a passphrase, so that you can both add more data to the seed, other then it strictly representing the initial entropy, plus it allows for re-encoding the same entropy (i.e. the same wallet) into a different mnemonic seed by changing the passphrase.

buck54321:
We need a larger plan. Let's continue discussion via #1670. In the meantime, this can be closed or at least moved to draft.

buck54321:
Closing in favor of linked issues. If we ever do this, we'll use **aezeed** anyway. Thanks of the effort, @LasTshaMAN.

PR #1662, opened 2022-06-14
ui: Hide old errors on market page forms

martonp:
Some of the forms on the markets page would show an old error when opening the form. This is now fixed.

### DIFF
--- a/client/webserver/site/src/js/markets.ts
+++ b/client/webserver/site/src/js/markets.ts
@@ -340,7 +340,7 @@ export default class MarketsPage extends BasePage {
     bindForm(page.cancelForm, page.cancelSubmit, async () => { this.submitCancel() })
     // Order detail view
     Doc.bind(page.vFeeDetails, 'click', () => this.showForm(page.vDetailPane))
-    Doc.bind(page.closeDetailPane, 'click', () => this.showForm(page.verifyForm))
+    Doc.bind(page.closeDetailPane, 'click', () => this.showVerifyForm())
     // Bind active orders list's header sort events.
     page.liveTable.querySelectorAll('[data-ordercol]')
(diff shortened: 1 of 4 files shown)


client/webserver/site/src/js/markets.ts
```diff
@@ -1439,6 +1439,13 @@ export default class MarketsPage extends BasePage {
     }
   }
 
+  /* showVerifyForm displays form to verify an order */
+  async showVerifyForm () {
+    const page = this.page
+    Doc.hide(page.vErr)
+    this.showForm(page.verifyForm)
```
chappjc, line 1446: Strikes me as a bit backwards to clean up when opening the form instead of when closing the form.

Would the following not be targeted well enough?

```diff
diff --git a/client/webserver/site/src/js/markets.ts b/client/webserver/site/src/js/markets.ts
index f7a791e0..c4a1ff03 100644
--- a/client/webserver/site/src/js/markets.ts
+++ b/client/webserver/site/src/js/markets.ts
@@ -381,8 +381,9 @@ export default class MarketsPage extends BasePage {
     setOrdersSortColClasses()
 
     const closePopups = () => {
-      Doc.hide(page.forms)
+      Doc.hide(page.forms, page.vErr)
       page.vPass.value = ''
+      page.vErr.textContent = ''
       page.cancelPass.value = ''
     }
```

martonp: Both of these work, but I was thinking of it more of setting up the form to be displayed instead of cleaning it up. I actually see now that all of the forms on this page have the same issue, so I'll update those as well.

For a lot of the forms we have a `refresh` method that sets up the form. How about we make it the standard from now on to always have a refresh function for every popup form which must be called whenever you open it? Then, when forms are reused, these cleanup things don't have to be repeated in the `closePopups` function of every page. These `closePopups` methods can get very big when you need to handle the cleanup for each popup on the page each time you close one of them.

PR #1663, opened 2022-06-15
multi: show recently matched orders

### LINKED ISSUE #1143: ui: show recently matched orders
Would be nice to be able to see a list of the recently matched orders like they have on Binance.

vctt94:
This PR closes #1143 

I am adding a basic match struct into `MatchProofNote` so we can be notified on the client.

On the client I added a new recent matches table in our market page. Here how it looks right now:

This is how it looks right now:

image

smaller screen: 

image

### DIFF
--- a/client/core/bookie.go
+++ b/client/core/bookie.go
@@ -192,8 +192,9 @@ func (b *bookie) logEpochReport(note *msgjson.EpochReportNote) error {
 	}
 
 	marketID := marketName(b.base, b.quote)
-	if note.MatchesSummary != nil {
-		b.SetMatchesSummary(note.MatchesSummary, note.EndStamp)
+	b.SetMatchesSummary(note.MatchesSummarySell, note.EndStamp, true)
+	b.SetMatchesSummary(note.MatchesSummaryBuy, note.EndStamp, false)
+	if note.MatchesSummaryBuy != nil || note.MatchesSummarySell != nil {
 		b.send(&BookUpdate{
 			Action:   EpochMatchSummary,
 			MarketID: marketID,
--- a/client/core/bookie.go
+++ b/client/core/bookie.go
@@ -191,6 +191,14 @@ func (b *bookie) logEpochReport(note *msgjson.EpochReportNote) error {
 		return fmt.Errorf("epoch report has zero-valued candle end stamp")
 	}
 
+	marketID := marketName(b.base, b.quote)
+	b.SetMatchesSummary(note.MatchesSummary, note.EndStamp)
+	fmt.Printf("matches summary: %+v\n\n", b.GetMatchesSummary())
+	b.send(&BookUpdate{
+		Action:   EpochMatchSummary,
+		MarketID: marketID,
+		Payload:  b.GetMatchesSummary(),
+	})
 	for durStr, cache := range b.candleCaches {
 		c := cache.addCandle(&note.Candle)
 		if c == nil {
@@ -200,7 +208,7 @@ func (b *bookie) logEpochReport(note *msgjson.EpochReportNote) error {
 		b.send(&BookUpdate{
 			Action:   CandleUpdateAction,
 			Host:     b.dc.acct.host,
-			MarketID: marketName(b.base, b.quote),
+			MarketID: marketID,
 			Payload: CandleUpdate{
 				Dur:          durStr,
 				DurMilliSecs: uint64(dur.Milliseconds()),
@@ -1053,7 +1061,7 @@ func handleEpochReportMsg(_ *Core, dc *dexConnection, msg *msgjson.Message) erro
 
 // handleEpochOrderMsg is called when an epoch_order notification is
 // received.
-func handleEpochOrderMsg(c *Core, dc *dexConnection, msg *msgjson.Message) error {
+func handleEpochOrderMsg(_ *Core, dc *dexConnection, msg *msgjson.Message) error {
 	note := new(msgjson.EpochOrderNote)
 	err := msg.Unmarshal(note)
 	if err != nil {
--- a/client/core/bookie.go
+++ b/client/core/bookie.go
@@ -192,13 +192,14 @@ func (b *bookie) logEpochReport(note *msgjson.EpochReportNote) error {
 	}
 
 	marketID := marketName(b.base, b.quote)
-	b.SetMatchesSummary(note.MatchesSummary, note.EndStamp)
-	fmt.Printf("matches summary: %+v\n\n", b.GetMatchesSummary())
-	b.send(&BookUpdate{
-		Action:   EpochMatchSummary,
-		MarketID: marketID,
-		Payload:  b.GetMatchesSummary(),
-	})
+	if note.MatchesSummary != nil {
+		b.SetMatchesSummary(note.MatchesSummary, note.EndStamp)
+		b.send(&BookUpdate{
+			Action:   EpochMatchSummary,
+			MarketID: marketID,
+			Payload:  b.GetMatchesSummary(),
+		})
+	}
 	for durStr, cache := range b.candleCaches {
 		c := cache.addCandle(&note.Candle)
 		if c == nil {
--- a/client/core/types.go
+++ b/client/core/types.go
@@ -596,6 +596,7 @@ const (
 	UnbookOrderAction     = "unbook_order"
 	UpdateRemainingAction = "update_remaining"
 	CandleUpdateAction    = "candle_update"
+	EpochMatchSummary     = "epoch_match_summary"
 )
 
 // BookUpdate is an order book update.
--- a/client/orderbook/orderbook.go
+++ b/client/orderbook/orderbook.go
@@ -614,9 +614,9 @@ func (ob *OrderBook) BestFillMarketBuy(qty, lotSize uint64) ([]*Fill, bool) {
 	return ob.sells.bestFill(qty, true, lotSize)
 }
 
+// SetMatchesSummary set matches summary. If the matches summary length grows
+// bigger than 100, it will slice out the ones first added.
 func (ob *OrderBook) SetMatchesSummary(matches map[uint64]uint64, ts uint64) {
-	ob.matchSummaryMtx.Lock()
-	defer ob.matchSummaryMtx.Unlock()
 	newMatchesSummary := make([]*MatchSummary, len(matches))
 	i := 0
 	for rate, qty := range matches {
@@ -627,16 +627,26 @@ func (ob *OrderBook) SetMatchesSummary(matches map[uint64]uint64, ts uint64) {
 		}
 		i++
 	}
+
+	ob.matchSummaryMtx.Lock()
+	defer ob.matchSummaryMtx.Unlock()
 	if ob.matchesSummary == nil {
 		ob.matchesSummary = newMatchesSummary
 		return
 	}
 	ob.matchesSummary = append(newMatchesSummary, ob.matchesSummary...)
+	maxLength := 1000
+	// if maxLength is greater than 100, slice it and left the first 100.
+	if len(ob.matchesSummary) > maxLength+100 {
+		ob.matchesSummary = ob.matchesSummary[maxLength:]
+	}
 }
 
 func (ob *OrderBook) GetMatchesSummary() []*MatchSummary {
+	ob.matchSummaryMtx.Lock()
 	if ob.matchesSummary == nil {
 		ob.matchesSummary = make([]*MatchSummary, 0)
 	}
+	ob.matchSummaryMtx.Unlock()
 	return ob.matchesSummary
 }
--- a/client/orderbook/orderbook.go
+++ b/client/orderbook/orderbook.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"decred.org/dcrdex/dex"
 	"decred.org/dcrdex/dex/msgjson"
@@ -55,9 +56,10 @@ type rateSell struct {
 }
 
 type MatchSummary struct {
-	Rate uint64
-	Qty  uint64
-	Age  uint64
+	Rate uint64    `json:"rate"`
+	Qty  uint64    `json:"qty"`
+	Age  time.Time `json:"age"`
+	Sell bool      `json:"sell"`
 }
 
 // OrderBook represents a client tracked order book.
@@ -616,14 +618,18 @@ func (ob *OrderBook) BestFillMarketBuy(qty, lotSize uint64) ([]*Fill, bool) {
 
 // SetMatchesSummary set matches summary. If the matches summary length grows
 // bigger than 100, it will slice out the ones first added.
-func (ob *OrderBook) SetMatchesSummary(matches map[uint64]uint64, ts uint64) {
+func (ob *OrderBook) SetMatchesSummary(matches map[uint64]uint64, ts uint64, sell bool) {
+	if matches == nil {
+		return
+	}
 	newMatchesSummary := make([]*MatchSummary, len(matches))
 	i := 0
 	for rate, qty := range matches {
 		newMatchesSummary[i] = &MatchSummary{
 			Rate: rate,
 			Qty:  qty,
-			Age:  ts,
+			Age:  time.UnixMilli(int64(ts)),
+			Sell: sell,
 		}
 		i++
 	}
@@ -643,10 +649,5 @@ func (ob *OrderBook) SetMatchesSummary(matches map[uint64]uint64, ts uint64) {
 }
 
 func (ob *OrderBook) GetMatchesSummary() []*MatchSummary {
-	ob.matchSummaryMtx.Lock()
-	if ob.matchesSummary == nil {
-		ob.matchesSummary = make([]*MatchSummary, 0)
-	}
-	ob.matchSummaryMtx.Unlock()
 	return ob.matchesSummary
 }
--- a/client/orderbook/orderbook.go
+++ b/client/orderbook/orderbook.go
@@ -641,13 +641,32 @@ func (ob *OrderBook) SetMatchesSummary(matches map[uint64]uint64, ts uint64, sel
 		return
 	}
 	ob.matchesSummary = append(newMatchesSummary, ob.matchesSummary...)
-	maxLength := 1000
-	// if maxLength is greater than 100, slice it and left the first 100.
-	if len(ob.matchesSummary) > maxLength+100 {
-		ob.matchesSummary = ob.matchesSummary[maxLength:]
+	maxLength := 100
+	// if ob.matchesSummary length is greater than max length, we slice the array
+	// to maxLength, removing values first added.
+	if len(ob.matchesSummary) > maxLength {
+		ob.matchesSummary = ob.matchesSummary[:maxLength]
 	}
 }
 
-func (ob *OrderBook) GetMatchesSummary() []*MatchSummary {
-	return ob.matchesSummary
+// GetMatchesSummary returns a deep copy of MatchesSummary
+func (ob *OrderBook) GetMatchesSummary() []MatchSummary {
+	ob.matchSummaryMtx.Lock()
+	defer ob.matchSummaryMtx.Unlock()
+	matchesSummary := make([]MatchSummary, len(ob.matchesSummary))
+	for i := 0; i < len(ob.matchesSummary); i++ {
+		matchesSummary[i] = ob.matchesSummary[i].copy()
+	}
+	return matchesSummary
+}
+
+// makes a copy of a MatchSummary
+func (ob *MatchSummary) copy() MatchSummary {
+	matchSummary := MatchSummary{
+		Rate: ob.Rate,
+		Qty:  ob.Qty,
+		Age:  ob.Age,
+		Sell: ob.Sell,
+	}
+	return matchSummary
 }
--- a/client/orderbook/orderbook.go
+++ b/client/orderbook/orderbook.go
@@ -54,6 +54,12 @@ type rateSell struct {
 	sell bool
 }
 
+type MatchSummary struct {
+	Rate uint64
+	Qty  uint64
+	Age  uint64
+}
+
 // OrderBook represents a client tracked order book.
 type OrderBook struct {
 	log      dex.Logger
@@ -78,6 +84,9 @@ type OrderBook struct {
 	currentEpoch uint64
 	proofedEpoch uint64
 	epochQueues  map[uint64]*EpochQueue
+
+	matchSummaryMtx sync.Mutex
+	matchesSummary  []*MatchSummary
 
 	// feeRates is a separate struct to account for atomic field alignment in
 	// 32-bit systems. See also https://golang.org/pkg/sync/atomic/#pkg-note-BUG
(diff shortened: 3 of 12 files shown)


chappjc:
> I noticed the MakerOID might be a parameter which causes confusion, as it is possible to have many makers into a limit order.

You're referring to the distinction between `Match` and `MatchSet` (both `dex/order` types), where a `MatchSet` has one taker order and N maker orders.  Note that a `Match` has it's own `Quantity` that is not necessarily the amount of either maker or taker *order*.  I think the `[]BasicMatch` should be distilled from all the `Match`s from the `MatchSets`s.  In other words, if `(*MatchSet).Matches` returns 4 matches, there should be 4 `BasicMatch`s all with the same taker and potentially different makers.

chappjc:
rougly:

```go
var basicMatches []*order.BasicMatch
for _, matchSet := range matches {
	for i, match := range matchSet.Matches() {
		basicMatches = append(basicMatches, &order.BasicMatch{
			MatchID:  match.ID(),
			MakerOID: match.Maker.ID(),
			TakerOID: match.Taker.ID(),
			Quantity: match.Quantity,
			Rate:     match.Rate,
		})
	}
}
```

client/webserver/site/src/js/markets.ts
```diff
@@ -187,6 +197,7 @@ export default class MarketsPage extends BasePage {
     // that the screen is updated with the most recent one.
     this.maxOrderUpdateCounter = 0
     this.metaOrders = {}
+    this.recentMatches = []
```
vctt94, line 200: should this be stored at the webserver so we can keep this information in case of refresh?

client/webserver/site/src/js/markets.ts
```diff
@@ -1749,6 +1796,90 @@ export default class MarketsPage extends BasePage {
...
+    console.log(this.recentMatchesSortKey)
+    switch (this.recentMatchesSortKey) {
+      case 'rate':
+        return (a: BasicMatches, b: BasicMatches) => this.recentMatchesSortDirection * (a.rate - b.rate)
+      case 'qty':
+        return (a: BasicMatches, b: BasicMatches) => this.recentMatchesSortDirection * (a.qty - b.qty)
+      case 'stamp':
+        return (a: BasicMatches, b: BasicMatches) => this.recentMatchesSortDirection * (b.age.getTime() - a.age.getTime())
+
+    // Sort by taker and maker?
```
vctt94, line 1813: Does sorting by taker and maker make sense?

chappjc: My suggestion to include the OIDs was really just for the server->client messaging, which gives the backend some more capability to audit the matches based on the match_proof and other orderbook subscription data.  The frontend can be very simple.

dex/msgjson/types.go
```diff
@@ -845,14 +845,25 @@ type OrderBook struct {
 	QuoteFeeRate uint64           `json:"quoteFeeRate"`
 }
 
+type BasicMatch struct {
+	MatchID  Bytes  `json:"matchid"`
+	MakerOID Bytes  `json:"maker"`
+	TakerOID Bytes  `json:"taker"`
```
buck54321, line 851: > @chappjc: My suggestion to include the OIDs in the BasicMatch was really just for the server->client messaging, which gives the backend some more capability to audit the matches based on the match_proof and other orderbook subscription data.

Can you be more specific?

This is > 200 bytes **per match** sent to every consumer. I think we'd need a pretty compelling reason to justify that.

buck54321:
This whole thing is a hard sell for me. I think showing matches live makes sense when matches can occur at any time, but when there's epochs, we're really just displaying the results of the last match cycle, which is better summarized with statistics rather than specific information about the matches.

vctt94:
By statistics you mean the percentage of matches in some price range?

buck54321:
How many matches and what was the average and the range of buy and sell prices. I'd like standard deviation too, but most users won't know what to do with it. That info can be encoded in tens of bytes, as opposed to individual matches which may be more like tens of **kilo**bytes. 

Note that the bandwidth requirements of additional notification data compounds with busier markets, since we have more activity (= larger match data reports) **and** more subscribers.

dex/msgjson/types.go
chappjc: The `MatchID` itself can be omitted, but what I'm describing is how an observer can determine what orders were matched in an epoch.  This can only be done at present by:
- subscribe to orderbook
- observe all the epoch order notes in a full epoch
- at epoch close use the match_proof note with the preimages listed to collect the orders that should be matched
- **run the matching algorithm** against book - major burden, not for the casual user

The idea of just sending the matches (this PR) is to avoid the above if we just want matches for display purposes.  If someone really wants to audit the matches, they can go to the trouble the steps above with an orderbook subscription.

chappjc:
> This whole thing is a hard sell for me. I think showing matches live makes sense when matches can occur at any time, but when there's epochs, we're really just displaying the results of the last match cycle, which is better summarized with statistics rather than specific information about the matches.

It's just that literally every exchange shows all fills.

How I can see the information distilled is a table of cumulative amounts filled at different prices.  Each amount can be the result of multiple fills, but I think it's fine to gloss over that and just report how much was filled at a certain price.

I think that reducing an epochs matches to a couple numbers like avg, range, etc. it losing too much information.  However, I'm on board to scrap sending every match made separately and condensing the info instead, just not that much.

dex/msgjson/types.go
buck54321: I don't agree that a casual user needs that much detail, but I'm just one guy. How about we just do a slice of tuples? 

```Go
Matches [][2]uint64 `json:"matches"` // [rate, qty]
```

and I guess we should send it in the `EpochReportNote`. Is it relevant to the match proof?

chappjc: That's actually what I'm suggesting in my other comment https://github.com/decred/dcrdex/pull/1663#issuecomment-1199709874 Except I think we can aggregate all matches for a given rate and sum the quantity.

I do concede that every match in an epoch isn't quite necessary, but I do think that giving the amounts and rates, even if aggregated over many matches, is important.  No exchange I have ever seen does not give you this info.

vctt94: So, I think we can remove `BasicMatch`, and add 

```
	// quantity of matches by rate: {rate}qty
	Matches map[uint64]uint64 `json:"matches"`
```

into `MatchCycleStats` so it is reported on `EpochReportNote`?

chappjc: That seems reasonable.  Using a map to aggregate quantities for a rate is certainly convenient (e.g. `ms[rate]+=qty` just works even before the first entry for a given `rate`).

As for the JSON in the epoch report, a slice of tuples might end up better for unmarshalling.

```json
"matchSummary": [
  [11000, 20],
  [11400, 40],
  [12000, 60],
]
```

vs.

```json
"matchSummary": {
  "11000": 20,
  "11400": 40,
  "12000": 60,
}
```

Because when you marshal a `map[uint64]uint64`, it makes the key into a string.  But as long as the quoted key represents a number it unmarshals fine, so maybe it's fine.  https://go.dev/play/p/uJO6JL64CuV  Go ahead and feel it out.

vctt94: I pushed the sugested changes, aggregating on a map [rate]quantity. 

Still need to fix the site build, to show this information properly, but pushed it before changing the front, so I can get your opinions.

vctt94: The problem with a tuple would be that we would need to run the array in order to find the rate so we can sum new matches from the server, while with a map we would simple sum it up, like you said: `ms[rate]+=qty`. 

So I believe it would be simpler to just convert the matchSummary from string to uint, but I am fine with either way. 

Which one do you prefer @chappjc ?

chappjc: > The problem with a tuple would be that we would need to run the array in order to find the rate so we can sum new matches from the server, while with a map we would simple sum it up, like you said: `ms[rate]+=qty`.

I guess I have to review your current code, but I thought that aggregation happens only on the server to reduce an epochs matches, and we can send anything else to the client although it might require converting and sorting I suppose.

> So I believe it would be simpler to just convert the matchSummary from string to uint, but I am fine with either way.
> 
> Which one do you prefer @chappjc ?

No preference yet.  I'd have to get into the code and feel it out.  I don't feel like it's a critical design decision except that updating a msgjson type commits us to live with that choice.

client/core/core.go
```diff
@@ -217,6 +217,7 @@ func (dc *dexConnection) marketMap() map[string]*Market {
 			EpochLen:        mkt.EpochLen,
 			StartEpoch:      mkt.StartEpoch,
 			MarketBuyBuffer: mkt.MarketBuyBuffer,
+			RecentMatches:   mkt.RecentMatches,
```
chappjc, line 220: Recent matches won't be be coming from a `'config'` response (in the `msgjson.Market`).  I wasn't sure the best place to keep the data, so I reviewed what we're doing for other `'orderbook'` and `'spots'` subscription notifications.

Currently we decided to use the `EpochReportNote`, which is part of the `orderbook` subscription, to communicate recent matches from the server in notifications.  Time-stamped recent matches for a market would be tracked by some new fields in `dexConnection` (e.g. the `spots` and `spotsMtx` fields), or possibly in the `bookie` for a given market (see how `(*bookie).logEpochReport(*msgjson.EpochReportNote)` already does this for this note but for the `Candle` data).

Now note how we send `BookUpdate`s to consumers like the frontend, with various "actions":
https://github.com/decred/dcrdex/blob/3a476f110fe1e295b20a400edb34f6e08fe1fc74/client/core/types.go#L570-L578

So, I imagine we'd follow that pattern.

chappjc (changes requested):

client/core/core.go
vctt94: I am using now `epoch_report` to update recent matches

vctt94:
Thanks for the review chapp. I am now storing the recent matches at bookies, and adding age by epoch, so we can show them diferently.

One thing about this solution, is that it is not possible to know if it  is a sell or not, as we are only sending number of qty and rate.

So I am showing everything with the same color: 

image

client/core/types.go
```diff
@@ -403,20 +403,28 @@ func coreOrderFromTrade(ord order.Order, metaData *db.OrderMetaData) *Order {
...
+	QuoteID         uint32         `json:"quoteid"`
+	QuoteSymbol     string         `json:"quotesymbol"`
+	LotSize         uint64         `json:"lotsize"`
+	RateStep        uint64         `json:"ratestep"`
+	EpochLen        uint64         `json:"epochlen"`
+	StartEpoch      uint64         `json:"startepoch"`
+	MarketBuyBuffer float64        `json:"buybuffer"`
+	Orders          []*Order       `json:"orders"`
+	SpotPrice       *msgjson.Spot  `json:"spot"`
+	RecentMatches   []*recentMatch `json:"recentmatches"`
```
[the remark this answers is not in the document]
chappjc: Agree about that.  I don't think I was explicit this point in https://github.com/decred/dcrdex/pull/1663#discussion_r936023097, but this `Market` struct doesn't deal with order book things.  The Orders that are in this struct are your own.  The only possible violation there is the spot price, but we didn't include that in the OrderBook types.

client/core/bookie.go
```diff
@@ -191,6 +191,13 @@ func (b *bookie) logEpochReport(note *msgjson.EpochReportNote) error {
 		return fmt.Errorf("epoch report has zero-valued candle end stamp")
 	}
 
+	marketID := marketName(b.base, b.quote)
+	b.SetRecentMatches(note.MatchesSummary, note.EndStamp)
+	b.send(&BookUpdate{
+		Action:   "epoch_report",
```
chappjc, line 197: This action would need to be defined in a const, near where `FreshBookAction` is defined.  But let's not use `"epoch_report"` since the recent matches is just a subset of the epoch report message from the server.  Perhaps `epoch_match_summary`

client/core/types.go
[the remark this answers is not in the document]
chappjc: Also before we get too committed to the naming, `RecentMatches` and `RecentMatch" suggests that each entry is an actual match, but this is a distillation of an epoch's matches into a summary.  This is an epoch match summary.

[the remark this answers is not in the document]
buck54321: > The only possible violation there is the spot price, but we didn't include that in the OrderBook types.

We get the spot price from the `"price_feed"` subscription instead of the order book feed, and we expect to have it for all markets all of the time.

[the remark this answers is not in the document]
chappjc: Indeed!  I was thinking of candles.  My bad.

client/core/core.go
```diff
@@ -206,6 +206,21 @@ func (dc *dexConnection) marketMap() map[string]*Market {
...
+			matchesSummary := book.GetRecentMatches()
+			for _, m := range matchesSummary {
+				recentMatches = append(recentMatches, &recentMatch{
+					Age:  m.Age,
+					Qty:  m.Qty,
+					Rate: m.Rate,
+				})
+			}
+		}
+		dc.booksMtx.RUnlock()
```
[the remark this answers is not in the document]
vctt94: removed. The reason I did that aws to have recent matches information on after refreshing the page, but I guess it is fine waiting for the next epoch

client/orderbook/orderbook.go
```diff
@@ -604,3 +613,30 @@ func (ob *OrderBook) BestFill(sell bool, qty uint64) ([]*Fill, bool) {
...
+		return
+	}
+	ob.matchesSummary = append(newMatchesSummary, ob.matchesSummary...)
+}
+
+func (ob *OrderBook) GetMatchesSummary() []*MatchSummary {
+	if ob.matchesSummary == nil {
+		ob.matchesSummary = make([]*MatchSummary, 0)
+	}
+	return ob.matchesSummary
```
JoeGruffins, line 641: I think you want to hold the lock for this.

client/orderbook/orderbook.go
```diff
@@ -604,3 +613,30 @@ func (ob *OrderBook) BestFill(sell bool, qty uint64) ([]*Fill, bool) {
...
+			Qty:  qty,
+			Age:  ts,
+		}
+		i++
+	}
+	if ob.matchesSummary == nil {
+		ob.matchesSummary = newMatchesSummary
+		return
+	}
+	ob.matchesSummary = append(newMatchesSummary, ob.matchesSummary...)
```
JoeGruffins, line 634: This grows forever. Maybe a good idea to prune sometimes? Does not look very big, but not sure whats best. Maybe is len is more than 100_000 chop off the first 10_000?

JoeGruffins (commented):
Looks ok to me.

vctt94:
Thanks for the review joe

client/orderbook/orderbook.go
vctt94: Now if matchesSummary is bigger than 100, it will slice out the first ones added to the queue

client/orderbook/orderbook.go
```diff
@@ -627,16 +627,25 @@ func (ob *OrderBook) SetMatchesSummary(matches map[uint64]uint64, ts uint64) {
...
 		ob.matchesSummary = newMatchesSummary
 		return
 	}
 	ob.matchesSummary = append(newMatchesSummary, ob.matchesSummary...)
+	// if len is greater than 100, slice the first 100.
+	if len(ob.matchesSummary) > 100 {
+		ob.matchesSummary = ob.matchesSummary[0:100]
+	}
+	ob.matchSummaryMtx.Unlock()
 }
```
JoeGruffins, line 642: This will only save tho older ones correct? I think you want to throw away the older ones, so `ob.matchesSummary = ob.matchesSummary[100:]`

https://go.dev/play/p/_Dg-K0wEnNA

JoeGruffins: also, this would make more sense if len is over 1000 or so, otherwise would delete almost everything whenever it hits 100.

vctt94: Yeah 100 is too small. 

Now I am doing 1000 and if it is greater than 1100.

Thanks for the suggestion

client/orderbook/orderbook.go
```diff
@@ -54,6 +54,12 @@ type rateSell struct {
 	sell bool
 }
 
+type MatchSummary struct {
+	Rate uint64
+	Qty  uint64
+	Age  uint64
```
buck54321, line 60: Time stamp is much more flexible. An age is stale as soon as you write it.

client/orderbook/orderbook.go
```diff
@@ -604,3 +613,40 @@ func (ob *OrderBook) BestFill(sell bool, qty uint64) ([]*Fill, bool) {
...
+}
+
+func (ob *OrderBook) GetMatchesSummary() []*MatchSummary {
+	ob.matchSummaryMtx.Lock()
+	if ob.matchesSummary == nil {
+		ob.matchesSummary = make([]*MatchSummary, 0)
+	}
+	ob.matchSummaryMtx.Unlock()
+	return ob.matchesSummary
+}
```
buck54321, line 652: I don't see a great reason to ensure we are returning a non nil slice. Caller should handle it properly.

client/webserver/site/src/js/markets.ts
```diff
@@ -1789,6 +1839,58 @@ export default class MarketsPage extends BasePage {
...
+  updateRecentMatchRow (tr: HTMLElement, match: RecentMatch) {
+    app().bindTooltips(tr)
+    updateDataCol(tr, 'rate', Doc.formatCoinValue(match.Rate / this.market.rateConversionFactor))
+    // change rate color based if is sell or not.
+    updateDataCol(tr, 'qty', Doc.formatCoinValue(match.Qty, this.market.baseUnitInfo))
+    updateDataCol(tr, 'age', match.Age.toLocaleDateString())
+    updateToolTipCol(tr, 'age', match.Age.toLocaleTimeString())
+  }
+
+  bindToRecentMatches (matches: RecentMatch[]) {
```
buck54321, line 1882: You're not binding anything

client/webserver/site/src/js/markets.ts
```diff
@@ -1789,6 +1839,58 @@ export default class MarketsPage extends BasePage {
...
+    for (const match of recentMatches) {
+      const row = page.recentMatchesTemplate.cloneNode(true) as HTMLElement
+
+      this.updateRecentMatchRow(row, match)
+
+      page.recentMatchesLiveList.append(row)
+    }
+  }
+
+  updateRecentMatchRow (tr: HTMLElement, match: RecentMatch) {
```
buck54321, line 1873: Only used once. Wanna just do it inline?

dex/msgjson/types.go
```diff
@@ -1176,6 +1176,8 @@ type EpochReportNote struct {
 	Epoch        uint64 `json:"epoch"`
 	BaseFeeRate  uint64 `json:"baseFeeRate"`
 	QuoteFeeRate uint64 `json:"quoteFeeRate"`
+	// quantity of matches by rate: [rate]qty
+	MatchesSummary map[uint64]uint64 `json:"matchessumary"`
```
buck54321, line 1180: 1. You're missing an **m** in summary
2. This should be camel-case

client/webserver/site/src/js/registry.ts
```diff
@@ -314,6 +315,12 @@ export interface OrderNote extends CoreNote {
   order: Order
 }
 
+export interface RecentMatch {
+  Rate: number
+  Qty: number
+  Age: Date
```
buck54321, line 321: Why are these upper-case? Add JSON tags or do type conversion as needed to fix this.

dex/msgjson/types.go
buck54321, line 1180: How about separating into buys and sells so we can apply colors eventually.

chappjc: The is one issue with aggregating.  Some are takers buying some are takers selling, combined in quantity at a given rate.

buck54321: You can keep separate maps/slices all the way down.

chappjc: Works for me.  To be clear you mean in `processEpochReady`, two maps server-side, one for aggregated taker sells and one for aggregated taker buys?
`match.Taker.Trade().Sell`

buck54321: Hmm. I guess we lose order information that way.

buck54321: Let's roll with this as is. If we want to iterate later, we can.

client/orderbook/orderbook.go
JoeGruffins: I still think you don't need to chop off so much. I suggest 
```
if len(ob.matchesSummary) > maxLength {
        ob.matchesSummary = ob.matchesSummary[len(ob.matchesSummary)-maxLength+100:]
}
```
So, if the thing is over 1000 set it to 900.

vctt94:
Thanks for the review.

That's how it looks after last review:

image

chappjc:
Nice, thanks.  That should be good to merge and move to next steps.  I'm pretty certain we don't want recent matches taking up nearly that much real estate, more like a small column carved out of where the current depth chart is, but we can refine this now.

chappjc:
A couple things wrong.

image

The "Age" column shows just a date.  The column should be "Time", and the time should be more precise than just the date.
  Please see the model screen captures in https://github.com/decred/dcrdex/pull/1663#discussion_r922409282

Other think is that the colors (or the sell bool) is backwards.  Sell=true means the taker order was a sell, but it seems to be showing the opposite.

server/market/market.go
```diff
@@ -2589,16 +2589,29 @@ func (m *Market) processReadyEpoch(epoch *readyEpoch, notifyChan chan<- *updateS
 		log.Errorf("Error updating API data collector: %v", err)
 	}
 
+	basicMatchesSell := make(map[uint64]uint64)
+	basicMatchesBuy := make(map[uint64]uint64)
+	for _, matchSet := range matches {
+		for _, match := range matchSet.Matches() {
+			if match.Maker.Sell {
```
chappjc, line 2596: `if match.Taker.Sell`.

I'm 99% certain that the buy/sell for a match considers the taker.

vctt94: Thanks, I was confused by this. Believe it is properly now.

server/market/market.go
```diff
@@ -2589,16 +2589,29 @@ func (m *Market) processReadyEpoch(epoch *readyEpoch, notifyChan chan<- *updateS
...
 	}
 
+	basicMatchesSell := make(map[uint64]uint64)
+	basicMatchesBuy := make(map[uint64]uint64)
+	for _, matchSet := range matches {
+		for _, match := range matchSet.Matches() {
+			if match.Taker.Trade() != nil && match.Taker.Trade().Sell {
+				basicMatchesSell[match.Rate] += match.Quantity
+			} else {
+				basicMatchesBuy[match.Rate] += match.Quantity
```
buck54321, line 2599: Will this create a `0 -> 0` bucket for cancel order matches?

buck54321 (approved):

chappjc:
Thanks.  I think we'll iterate on the frontend parts a bit.  

Perhaps also we'll look at having `SetMatchesSummary` take both buys and sells, and then return the new match summary so we don't have to explicity go get it.

We also will look into different aggregation server-side, but this is a basis to experiment.

PR #1665, opened 2022-06-16
webpack: use npm exec to run tsc in node_module

chappjc:
Calling manually ./node_modules/.bin/tsc is not portable, and fails
on Windows with a syntax error.  Siwtch to using npm exec to run
the tsc that is installed locally.


davecgh (approved):
The commit message only calls out `tsc`, but technically it's both `tsc` and `eslint`.  Tested that the code changes work as intended.

PR #1667, opened 2022-06-17
client/core,webserver,rpcserver: order submit time

chappjc:
```
Use order submission time, which is a local time stamp, instead of the
server's time stamp, which is from the server's system clock, when
displaying order age and time on the frontend.

This adds the core.Order.SubmitTime field to complement the Stamp
field.  This marshals to json as "submitTime".

The frontend is updated to use "submitTime" instead of "stamp".
```

Should resolves orders showing with a huge order age for a little while if the server's clock is off (behind) from the local client system clock.

image

In the above reported case, it stays like that for a minute or so, then starts at 0s, presumably when the clock discrepancy is passed.


client/db/types.go
```diff
@@ -358,6 +358,7 @@ type MatchMetaData struct {
 	// Stamp is the match time (ms UNIX), according to the server's 'match'
 	// request timestamp.
 	Stamp uint64
+	// TODO: ReceiveTime uint64 -- local time stamp for match age and time display
```
chappjc, line 361: I wanted to sleep on the match stamp issue because unlike order there's no times stamp from the client's machine, only the server stamp.

One possibility is when making the `core.Match` for the frontend, this `MatchMetaData.Stamp` can be compared to the order's `Stamp` to get a time-since-order offset and add that to the order's `SubmitTime` to get a `ReceiveTime` approximation for the match.

Obviously we could also start recording the ReceiveTime as well when the match is received, but I'm not sure that's worth storing in the DB.

client/webserver/site/src/html/order.tmpl
```diff
@@ -82,7 +82,7 @@
       </div>
       <div class="order-datum">
         <div>[[[Age]]]</div>
-        <div data-stamp="{{$ord.Stamp}}"></div>
+        <div data-stamp="{{$ord.SubmitTime}}"></div>
```
chappjc, line 85: Left as `data-stamp` so it is one of the `stampers` used in order.ts with `setStamp` to covert these fields into ages based on this attribute's presence

client/webserver/site/src/js/order.ts
```diff
@@ -94,7 +94,7 @@ export default class OrderPage extends BasePage {
 
     const setStamp = () => {
       for (const span of stampers) {
-        span.textContent = Doc.timeSince(parseInt(span.dataset.stamp || ''))
+        span.textContent = Doc.timeSince(parseInt(span.dataset.submitTime || span.dataset.stamp || '')) // order and match stamps
```
chappjc, line 97: whoops this is wrong

buck54321 (approved):

client/db/types.go
buck54321: Fine either way, but I would put it in the `MatchAuth`.

martonp (approved):

PR #1668, opened 2022-06-17
multi: use the UnixMilli functions in the std lib

chappjc:
Because my brain needed a coffee break.


chappjc:
Not too valuable, but it's working like before afaict.

buck54321 (approved):

PR #1669, opened 2022-06-18
ui: Fix order options bugs

martonp:
This diff fixes multiple bugs related to the order options:
- The order options weren't being passed to the client
- An additional range selection UI element was being displayed in each of the order option boxes
- If the acceleration form was displayed, the order options would no longer work


client/webserver/site/src/html/markets.tmpl
```diff
@@ -527,32 +527,31 @@
                   <span data-tmpl="tooltip" class="ico-info me-2" data-tooltip="Prevent temporary overlocking of funds"></span>
                   <img data-tmpl="chainIcon" class="micro-icon me-1">
                 </div>
+              </div>
+            </div>
 
-                <div id="booleanOptTmpl" class="d-flex justify-content-between align-items-center fs14 ps-2">
```
martonp, line 531: I just moved the `rangeOptTmpl` and the `boolOptTmpl` to be outside of the `orderOptTmpl` here, there are no other changes.

client/webserver/site/src/js/markets.ts
```diff
@@ -1786,12 +1795,9 @@ export default class MarketsPage extends BasePage {
   async submitOrder () {
     const page = this.page
     Doc.hide(page.orderErr, page.vErr)
-    const order = this.parseOrder()
+    const order = this.currentOrder
```
martonp, line 1798: This was the reason the order options weren't being passed to the client.

client/webserver/site/src/js/orderutil.ts
```diff
@@ -123,9 +123,19 @@ export function matchStatusString (status: number, side: number) {
 // life easier.
 let orderOptTmpl: HTMLElement, booleanOptTmpl: HTMLElement, rangeOptTmpl: HTMLElement
 
-// setOptionTemplates sets the package vars for the templates and application.
-export function setOptionTemplates (page: Record<string, PageElement>) {
-  [booleanOptTmpl, rangeOptTmpl, orderOptTmpl] = [page.booleanOptTmpl, page.rangeOptTmpl, page.orderOptTmpl]
+// setRangeOptionTempalate sets the package range option template variable.
+export function setRangeOptionTempalate (tmpl: PageElement) {
```
martonp, line 127: I changed this because on the `markets.tmpl` we had two instances of the `rangeOptTmpl`, one in the acceleration popup and one on the verify order popup. By doing this we can avoid having two elements with the same ID on the page.

client/webserver/site/src/js/forms.ts
```diff
@@ -1027,7 +1027,7 @@ export class AccelerateOrderForm {
     this.currencyUnit = preAccelerate.suggestedRange.yUnit
     page.accelerateAvgFeeRate.textContent = `${preAccelerate.swapRate} ${preAccelerate.suggestedRange.yUnit}`
     page.accelerateCurrentFeeRate.textContent = `${preAccelerate.suggestedRate} ${preAccelerate.suggestedRange.yUnit}`
-    OrderUtil.setOptionTemplates(page)
+    OrderUtil.setRangeOptionTempalate(page.accelerationRangeOptTmpl)
```
buck54321, line 1030: If the `accelerationRangeOptTmpl` is different than the `rangeOptTmpl`, then please make a new `class`. Maybe it can inherit `XYRangeHandler`.

client/webserver/site/src/js/markets.ts
```diff
@@ -215,7 +215,7 @@ export default class MarketsPage extends BasePage {
     const success = () => { /* do nothing */ }
     // Do not call cleanTemplates before creating the AccelerateOrderForm
     this.accelerateOrderForm = new AccelerateOrderForm(page.accelerateForm, success)
-    Doc.cleanTemplates(page.rangeOptTmpl)
+    Doc.cleanTemplates(page.accelerationRangeOptTmpl)
```
buck54321, line 218: You can just do this with the other templates a few lines down.

client/webserver/site/src/js/markets.ts
```diff
@@ -241,7 +241,13 @@ export default class MarketsPage extends BasePage {
...
     // Prepare templates for the buy and sell tables and the user's order table.
     Doc.cleanTemplates(page.rowTemplate, page.liveTemplate, page.durBttnTemplate, page.booleanOptTmpl, page.rangeOptTmpl, page.orderOptTmpl)
-    OrderUtil.setOptionTemplates(page)
+
+    const setOrderOptionTemplates = () => {
+      OrderUtil.setRangeOptionTempalate(page.rangeOptTmpl)
+      OrderUtil.setBooleanOptionTemplate(page.booleanOptTmpl)
+      OrderUtil.setOrderOptionTempalate(page.orderOptTmpl)
+    }
+    setOrderOptionTemplates()
```
buck54321, line 250: Don't need to separate these if you're not changing them (which you shouldn't be). Just pull out the new template in `setOptionTemplates`.

client/webserver/site/src/js/forms.ts
martonp: It wasn't different. I've done some refactoring.. all the templates are now in a template in `forms.tmpl`, and included in `markets.tmpl` and `order.tmpl`. I think it's much cleaner now.

JoeGruffins:
Difficult to explain, but the first time I slide the sliders they pop back to zero now.

chappjc (approved):
LGTM.  Tests well.

JoeGruffins:
Sorry to hijack but what program do you use for screen video captures? @chappjc

chappjc:
[Peek](https://github.com/phw/peek)

client/webserver/site/src/js/order.ts
```diff
@@ -70,8 +70,9 @@ export default class OrderPage extends BasePage {
       this.refreshOnPopupClose = true
     }
     // Do not call cleanTemplates before creating the AccelerateOrderForm
+    OrderUtil.setOptionTemplates(page)
```
buck54321, line 73: The `setOptionTemplates` things is kinda a hack anyway. We should really just pass the `page.*Tmpl` elements to the form constructors. But no reason to mix it up now.

buck54321 (approved):

PR #1675, opened 2022-06-20
ui: Copy deposit address alert

martonp:
Adds a disappearing message confirming that the deposit address has been copied.

### DIFF
--- a/client/webserver/site/src/css/wallets.scss
+++ b/client/webserver/site/src/css/wallets.scss
@@ -92,3 +92,9 @@ table.wallets {
   cursor: pointer;
   color: #a8a8a8;
 }
+
+#copyAlert {
+  margin-left: 9px;
+  position: absolute;
+  z-index: 1;
+}
--- a/client/webserver/site/src/css/wallets.scss
+++ b/client/webserver/site/src/css/wallets.scss
@@ -94,7 +94,6 @@ table.wallets {
 }
 
 #copyAlert {
-  margin-left: 9px;
+  margin-left: 12px;
   position: absolute;
-  z-index: 1;
 }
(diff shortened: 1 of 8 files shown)


ukane-philemon:
Nice, I was actually looking out for this :).

client/webserver/site/src/js/wallets.ts
```diff
@@ -233,8 +233,13 @@ export default class WalletsPage extends BasePage {
     Doc.hide(this.page.forms)
   }
 
-  copyAddress () {
-    navigator.clipboard.writeText(this.page.depositAddress.textContent || '')
+  async copyAddress () {
+    const page = this.page
+    navigator.clipboard.writeText(page.depositAddress.textContent || '')
+    Doc.show(page.copyAlert)
```
chappjc, line 239: We should only show the "Copied!" alert if `writeText` says it succeeded.

```js
navigator.clipboard.writeText(textContent).then(() => {
	Doc.show(page.copyAlert)
	...
}, (reason) => {
	console.error('Unable to copy:', reason) // or a different alert, whatever
})
```

https://github.com/decred/dcrdata/commit/c9521242ea01902ed3452250306d67ac54932824

client/webserver/site/src/css/wallets.scss
```diff
@@ -92,3 +92,9 @@ table.wallets {
   cursor: pointer;
   color: #a8a8a8;
 }
+
+#copyAlert {
+  margin-left: 9px;
```
chappjc, line 97: With 9x it's right up against the `div` for me.  Maybe 12?

chappjc (approved):
Working for me, but the alter text is a little close to the address text div.  A couple more pixels of margin and I think it would be a little better.

https://user-images.githubusercontent.com/9373513/175057130-35bae9f6-7bbd-4752-a9dc-d7bb42c73d9a.mp4

client/webserver/site/src/css/wallets.scss
martonp: 12 looks good.

PR #1679, opened 2022-06-23
eth: Update Solidity contract versions to 0.8.15.

### LINKED ISSUE #1646: eth: Update sol compiler version and recompile.
Looking through https://github.com/decred/dcrdex/blob/master/docs/InterFi-swap-contract-audit-v0v1-Q2-2022.pdf it seems that our only  problem is an "Outdated Compiler Version". So, let's update it. I guess `v0.8.14` is newest. Currently we are using `v0.8.6`

JoeGruffins:
closes #1646


chappjc (approved):
I get the same bins with solc 0.8.15

buck54321 (approved):

Issue #1671, opened 2022-06-20
TestTaperingQueue fails sometimes


martonp:
Failed during a CI run, but passes usually.

chappjc:
Try to capture the CI log if it happens again.
I tried to reproduce but unable:
`parallel -P 4 -N0 --halt soon,fail=1 go test -race -count=1 -short -run TestTaperingQueue ./dex/wait  :::: <(seq 1 2000)`

chappjc:
I think 5 ms is just too short for these CI VMs

chappjc:
Hmm, I think the "little optimization if this waiter would fire immediately" that I suggested allows the waiters to be executed out of order given the right circumstances.  Maybe we remove that and just let it append even if it should fire immediately, just to keep the order stable.

ukane-philemon:
https://github.com/decred/dcrdex/runs/7903466734?check_suite_focus=true fail log for #1766 
```
--- FAIL: TestTaperingQueue (0.01s)
    queue_test.go:123: wrong result order expected [+1 +3 +4 +2 +0], got [+1 +4 +3 +2 +0]
```

chappjc:
Thx.  I triggered a re-run of those jobs for now.

ukane-philemon:
Uhmm, I've hit a related fail here https://github.com/decred/dcrdex/actions/runs/3479377426/jobs/5817848849

chappjc:
The link to that failed CI run will be dead before too long, so here's the failure for posterity:

```
 --- FAIL: TestTaper (0.04s)
    queue_test.go:58: first intervals are out of bound: 11.752457ms
FAIL
FAIL	decred.org/dcrdex/dex/wait	0.092s
```

PR #1681, opened 2022-06-23
dex/wait: try to keep waiters in order

chappjc:
Potentially resolves https://github.com/decred/dcrdex/issues/1671.
However, As long as `runWaiter` is asynchronous (`go runWaiter(w)`) we cannot guarantee the `TryFunc`s will be called in the order that the waiters are popped off the waiter queue.

### DIFF
--- a/dex/wait/queue.go
+++ b/dex/wait/queue.go
@@ -238,17 +238,11 @@ func (q *TaperingTickerQueue) Run(ctx context.Context) {
 			w := waiters[0]
 			waiters = waiters[1:]
 			wg.Add(1)
-			go runWaiter(w)
+			go runWaiter(w) // note: async, so the next one could run its TryFunc first!
 
 		case w := <-q.queueWaiter:
-			// A little optimization if this waiter would fire immediately, but
-			// it works to append regardless.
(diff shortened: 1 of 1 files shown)


JoeGruffins (approved):

dex/wait/queue.go
```diff
@@ -238,17 +238,11 @@ func (q *TaperingTickerQueue) Run(ctx context.Context) {
...
 
 		case w := <-q.queueWaiter:
-			// A little optimization if this waiter would fire immediately, but
-			// it works to append regardless.
-			if time.Until(w.nextTick) <= 0 {
-				wg.Add(1)
-				go runWaiter(w)
-				continue
-			}
-
```
buck54321, line 251: I actually want to keep this. Is there any way we can adjust the test instead? What about doing this in `TestTaperingQueue`?

```Go
if numTrys > numTryAgains*2 {
```

chappjc: I'd like to keep it too.  Your test seem to enforce an implicit design goal of maintaining order, which I was trying to keep, but I think that can (needs to?) change too.

PR #1682, opened 2022-06-24
server/{market,swap,db}: free cancels

buck54321:
```
server/db:
Add epoch_gap column to cancel order table. When set (not EpochGapNA),
the epoch_gap is the number of epochs between the limit order placement
and the cancel order placement. Update InsertEpoch method to accept an
epoch gap argument (or EpochGapNA).

server/auth:
RecordCancel now accepts an epoch gap argument. Epoch gap is loaded from
the db via ExecutedCancelsForUser with a modified return value.
(*latestOrders).counts now only counts cancel orders if 0 <= epochGap < 2.

server/market:
processOrder now reports the epoch gap to the AuthManager for matched
cancel orders.
```

### DIFF
--- a/server/auth/cancel.go
+++ b/server/auth/cancel.go
@@ -12,12 +12,20 @@ import (
 	"decred.org/dcrdex/dex/order"
 )
 
+// freeCancelThreshold is the minimum number of epochs a user should wait before
+// placing a cancel order, if they want to avoid penalization. It is set to 2,
+// which means if a user places a cancel order in the same epoch as its limit
+// order, or the next epoch, the user will be penalized. This ensures that the
+// order remains booked for at least one full epoch and one full match cycle.
+const freeCancelThreshold = 2
+
 // oidStamped is a time-stamped order ID, with a field for target order ID if
 // the order is a cancel order.
 type oidStamped struct {
 	order.OrderID
-	time   int64
-	target *order.OrderID
+	time     int64
+	target   *order.OrderID
+	epochGap int
 }
 
 // ordsByTimeThenID is used to sort an ord slice in ascending order by time and
@@ -116,7 +124,7 @@ func (lo *latestOrders) counts() (total, cancels int) {
 
 	total = len(lo.orders)
 	for _, o := range lo.orders {
-		if o.target != nil {
+		if o.target != nil && o.epochGap >= 0 && o.epochGap < freeCancelThreshold {
 			cancels++
 		}
 	}
--- a/server/auth/cancel.go
+++ b/server/auth/cancel.go
@@ -15,8 +15,9 @@ import (
 // freeCancelThreshold is the minimum number of epochs a user should wait before
 // placing a cancel order, if they want to avoid penalization. It is set to 2,
 // which means if a user places a cancel order in the same epoch as its limit
-// order, or the next epoch, the user will be penalized. This ensures that the
-// order remains booked for at least one full epoch and one full match cycle.
+// order, or the next epoch, the user will be penalized. This value is chosen
+// because it is the minimum value such that the order remains booked for at
+// least one full epoch and one full match cycle.
 const freeCancelThreshold = 2
 
 // oidStamped is a time-stamped order ID, with a field for target order ID if
@@ -25,7 +26,7 @@ type oidStamped struct {
 	order.OrderID
 	time     int64
 	target   *order.OrderID
-	epochGap int
+	epochGap int32
 }
 
 // ordsByTimeThenID is used to sort an ord slice in ascending order by time and
--- a/server/db/driver/pg/orders.go
+++ b/server/db/driver/pg/orders.go
@@ -183,7 +183,7 @@ func (status pgOrderStatus) active() bool {
 
 // NewEpochOrder stores the given order with epoch status. This is equivalent to
 // StoreOrder with OrderStatusEpoch.
-func (a *Archiver) NewEpochOrder(ord order.Order, epochIdx, epochDur int64, epochGap int) error {
+func (a *Archiver) NewEpochOrder(ord order.Order, epochIdx, epochDur int64, epochGap int32) error {
 	return a.storeOrder(ord, epochIdx, epochDur, epochGap, orderStatusEpoch)
 }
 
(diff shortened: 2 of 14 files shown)


server/auth/cancel.go
```diff
@@ -116,7 +124,7 @@ func (lo *latestOrders) counts() (total, cancels int) {
 
 	total = len(lo.orders)
 	for _, o := range lo.orders {
-		if o.target != nil {
+		if o.target != nil && o.epochGap >= 0 && o.epochGap < freeCancelThreshold {
```
buck54321, line 127: This is the line that matters. This is a rule change. Free cancels after the order sits for one full epoch. If the order is submitted in epoch 10, and the cancel order submitted in epoch 10 or 11, that's a "short" cancel (not a free cancel), and the cancellation can hurt to the user's score.

chappjc:
Have not looked at the code, but as discussed, this is a very reasonable compromise to outright eliminating the cancellation ratio limit that is overly restrictive.  I think a very small `epochGap`, which you have proposed here, would not create much more of a challenge for mesh than already exists with the current order history knowledge requirements since it's necessarily very recent history.  Good stuff!

However, I still view the cancellation ratio limit as largely unnecessary given the epoch-shuffle matching algorithm, which makes it risky to cancel rapidly for deceptive purposes, combined with the need to have unique coins funding every order and the various order quantity limits, which throttle the ability of a user to place an excessive number of (presumably bogus) orders.

buck54321:
I think this accomplishes a few important things that will improve the character of our order books. Human users should have a fair chance to match with orders that they see. Whereas before, an order could be 1) placed and canceled in the same epoch, or 2) placed near the end of an epoch, and canceled in the next epoch. In either of those cases, there's a chance that I could see that order, place my order immediately, and then that order is canceled before my order comes up for matching. It's like probabilistic spoofing. But if you wait for the free cancel, that's eliminated.

If you truly make a mistake or something and want to cancel quickly, you still have that choice. Cancellations are relatively low impact in terms of penalization. Human traders will never run up against the cancellation ratio. But bots are encouraged to trade in a way that makes the book more stable and usable.

chappjc:
Well put.  I think this is a great balance.

JoeGruffins (approved):

server/db/interface.go
```diff
@@ -148,7 +148,7 @@ type OrderArchiver interface {
 	// NewEpochOrder stores a new order with epoch status. Such orders are
 	// pending execution or insertion on a book (standing limit orders with a
 	// remaining unfilled amount).
-	NewEpochOrder(ord order.Order, epochIdx, epochDur int64) error
+	NewEpochOrder(ord order.Order, epochIdx, epochDur int64, epochGap int) error
```
chappjc, line 151:
```go
// For trade orders, the epoch gap should be db.EpochGapNA, while for cancel orders
// it is the number of epochs since the targeted order was placed, as described in the
// docs for CancelRecord.
```

server/db/driver/pg/orders.go
```diff
@@ -183,8 +183,8 @@ func (status pgOrderStatus) active() bool {
 
 // NewEpochOrder stores the given order with epoch status. This is equivalent to
 // StoreOrder with OrderStatusEpoch.
-func (a *Archiver) NewEpochOrder(ord order.Order, epochIdx, epochDur int64) error {
-	return a.storeOrder(ord, epochIdx, epochDur, orderStatusEpoch)
+func (a *Archiver) NewEpochOrder(ord order.Order, epochIdx, epochDur int64, epochGap int) error {
+	return a.storeOrder(ord, epochIdx, epochDur, epochGap, orderStatusEpoch)
```
chappjc, line 187: Since the DB is using `INT4`, we should use `int32` instead of `int`, at least on the `Archiver` interface.  It's better to use a specific size integer anyway since `int` varies on different platforms.

server/market/market.go
```diff
@@ -2137,7 +2141,7 @@ func (m *Market) prepEpoch(orders []order.Order, epochEnd time.Time) (cSum []byt
 		// Change the order status from orderStatusEpoch to orderStatusRevoked.
 		coid, revTime, err := m.storage.RevokeOrder(ord)
 		if err == nil {
-			m.auth.RecordCancel(ord.User(), coid, ord.ID(), revTime)
+			m.auth.RecordCancel(ord.User(), coid, ord.ID(), db.EpochGapNA, revTime)
```
chappjc, line 2144: I'm thinking this one should perhaps use a gap of 0.  Yes, the user will get the preimage miss strike too, but this is effectively cancelling an order in the same epoch.  Redundant with the preimage violation?

server/db/driver/pg/upgrades.go
```diff
@@ -39,6 +39,10 @@ var upgrades = []func(db *sql.Tx) error{
 	// v4 upgrade updates the markets tables to use a integer type that can
 	// accommodate a 32-bit unsigned integer.
 	v4Upgrade,
+
+	// v5 upgrade adds an epoch_gap column to the cancel order tables in order
```
chappjc, line 43: <s>in order</s> extra words

server/db/driver/pg/upgrades.go
```diff
@@ -291,6 +295,30 @@ func v4Upgrade(tx *sql.Tx) (err error) {
...
+	if err != nil {
+		return fmt.Errorf("failed to read markets table: %w", err)
+	}
+
+	doTable := func(tableName string) error {
+		_, err = tx.Exec(fmt.Sprintf("ALTER TABLE %s ADD COLUMN epoch_gap INT4 DEFAULT -1;", tableName))
+		return err
+	}
+
+	log.Info("Adding epoch_gap column to %d market tables", len(mkts)*2)
```
chappjc, line 309: Info**f**

Also, this reads a bit like you have N markets.  But since the active cancels table will be instant compared to the _archived_ table, I'd prefer if this read `log.Infof("Adding epoch_gap column to cancel tables for %d markets", len(mkts))`

server/auth/cancel.go
```diff
@@ -12,12 +12,20 @@ import (
 	"decred.org/dcrdex/dex/order"
 )
 
+// freeCancelThreshold is the minimum number of epochs a user should wait before
+// placing a cancel order, if they want to avoid penalization. It is set to 2,
+// which means if a user places a cancel order in the same epoch as its limit
+// order, or the next epoch, the user will be penalized. This ensures that the
+// order remains booked for at least one full epoch and one full match cycle.
```
chappjc, line 19: `This ensures that the order remains booked` I think should be rephrased in terms of incentivizing this behavior rather than ensuring it.

chappjc (changes requested):
Looks great.  Some minor nits.

server/auth/cancel.go
buck54321: This was sloppy, but I wasn't commenting on the intention of free cancel period, but rather the reason this particular value was chosen. I'm really trying to say `This value is chosen because it is the minimum value such that the order remains booked for at least one full epoch and one full match cycle`.

Will update.

PR #1684, opened 2022-06-24
multi: optimize MaxOrder

### LINKED ISSUE #1612: max buy/sell endpoints/methods still use too much CPU
EDIT: The heavy CPU use was coming from dcrwallet's handing of listunspent (huge testnet wallet), but there's still a mystery of why frontend fires off many concurrent maxsell / order estimate requests.  I'm positive they weren't each from key presses when typing, but even if they were that would be undesirable without a delay after typing.

2022-05-12 15:57:31.791 [TRC] CORE[dcr_doge][book]: Validating match proof note for epoch 275395174 (dcr_doge) with 1 preimages and 0 misses.
2022-05-12 15:57:34.801 [TRC] CORE[doge]: no 2-conf feeRate available: <nil>
2022-05-12 15:57:34.801 [TRC] CORE[doge]: feeRateWithFallback using caller's suggestion for fee rate, 40000. Local estimate unavailable
2022/05/12 10:57:34 "POST http://127.0.0.2:5758/api/maxsell HTTP/1.1" from 127.0.0.1:32862 - 200 238B in 6.841225498s
2022-05-12 15:57:35.796 [TRC] CORE[doge]: no 2-conf feeRate available: <nil>
2022-05-12 15:57:35.796 [TRC] CORE[doge]: feeRateWithFallback using caller's suggestion for fee rate, 40000. Local estimate unavailable
2022/05/12 10:57:35 "POST http://127.0.0.2:5758/api/maxsell HTTP/1.1" from 127.0.0.1:32864 - 200 238B in 7.637165629s
2022-05-12 15:57:35.882 [TRC] CORE[doge]: no 2-conf feeRate available: <nil>
2022-05-12 15:57:35.882 [TRC] CORE[doge]: feeRateWithFallback using caller's suggestion for fee rate, 40000. Local estimate unavailable
2022/05/12 10:57:35 "POST http://127.0.0.2:5758/api/maxsell HTTP/1.1" from 127.0.0.1:32866 - 200 238B in 7.66689064s
2022-05-12 15:57:36.129 [DBG] CORE: Unsubscribing from the dcr_btc order book for dextesttaggpdj27s26wklczemovhqdzyvgokyumey33uemlw4imawqd.onion:7232
2022-05-12 15:57:36.529 [TRC] CORE: notify: |DATA| (epoch) - Index: 275395176
2022-05-12 15:57:36.639 [TRC] CORE[doge]: no 2-conf feeRate available: <nil>
2022-05-12 15:57:36.639 [TRC] CORE[doge]: feeRateWithFallback using caller's suggestion for fee rate, 40000. Local estimate unavailable
2022/05/12 10:57:36 "POST http://127.0.0.2:5758/api/maxsell HTTP/1.1" from 127.0.0.1:32868 - 200 238B in 7.122970715s
2022-05-12 15:57:36.805 [DBG] CORE[dcr]: Change output size = 36, addr = TsRGtZRG6xpA7FF8HLTxmBxhNL9zeFrGwBU

buck54321:
Resolves #1612

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -1756,18 +1756,23 @@ func fund(utxos []*compositeUTXO, enough func(uint64, uint64) bool) (
 			if len(okUTXOs) == 0 {
 				return false
 			}
-			// On each loop, find the smallest UTXO that is enough.
-			for _, txout := range okUTXOs {
-				if isEnoughWith(txout) {
-					addUTXO(txout)
-					return true
-				}
+
+			// Check if the largest output is too small.
+			lastUTXO := okUTXOs[len(okUTXOs)-1]
+			if !isEnoughWith(lastUTXO) {
+				addUTXO(lastUTXO)
+				okUTXOs = okUTXOs[0 : len(okUTXOs)-1]
+				continue
 			}
-			// No single UTXO was large enough. Add the largest (the last
-			// output) and continue.
-			addUTXO(okUTXOs[len(okUTXOs)-1])
-			// Pop the utxo.
-			okUTXOs = okUTXOs[:len(okUTXOs)-1]
+
+			// We only need one then. Find it.
+			idx := sort.Search(len(okUTXOs), func(i int) bool {
+				return isEnoughWith(okUTXOs[i])
+			})
+			// No need to check idx == len(okUTXOs). We already verified that the last
+			// utxo passes above.
+			addUTXO(okUTXOs[idx])
+			return true
 		}
 	}
 
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -1457,6 +1457,10 @@ func (dcr *ExchangeWallet) tryFund(utxos []*compositeUTXO, enough func(sum uint6
 		return nil
 	}
 
+	isEnoughWith := func(utxo *compositeUTXO) bool {
+		return enough(sum, size, utxo)
+	}
+
 	tryUTXOs := func(minconf int64) (ok bool, err error) {
 		sum, size = 0, 0
 		coins, spents, redeemScripts = nil, nil, nil
@@ -1473,22 +1477,25 @@ func (dcr *ExchangeWallet) tryFund(utxos []*compositeUTXO, enough func(sum uint6
 			if len(okUTXOs) == 0 {
 				return false, nil
 			}
-			// On each loop, find the smallest UTXO that is enough.
-			for _, txout := range okUTXOs {
-				if enough(sum, size, txout) {
-					if err = addUTXO(txout); err != nil {
-						return false, err
-					}
-					return true, nil
-				}
+
+			// Check if the largest output is too small.
+			lastUTXO := okUTXOs[len(okUTXOs)-1]
+			if !isEnoughWith(lastUTXO) {
+				addUTXO(lastUTXO)
+				okUTXOs = okUTXOs[0 : len(okUTXOs)-1]
+				continue
 			}
-			// No single UTXO was large enough. Add the largest (the last
-			// output) and continue.
-			if err = addUTXO(okUTXOs[len(okUTXOs)-1]); err != nil {
+
+			// We only need one then. Find it.
+			idx := sort.Search(len(okUTXOs), func(i int) bool {
+				return isEnoughWith(okUTXOs[i])
+			})
+			// No need to check idx == -1. We already verified that the last
+			// utxo passes above.
+			if err = addUTXO(okUTXOs[idx]); err != nil {
 				return false, err
 			}
-			// Pop the utxo.
-			okUTXOs = okUTXOs[:len(okUTXOs)-1]
+			return true, nil
 		}
 	}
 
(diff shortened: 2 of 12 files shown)


client/asset/btc/btc.go
```diff
@@ -1755,18 +1790,24 @@ func fund(utxos []*compositeUTXO, enough func(uint64, uint64) bool) (
...
-					return true
-				}
+
+			// Check if the largest output is too small.
+			lastUTXO := okUTXOs[len(okUTXOs)-1]
+			if !isEnoughWith(lastUTXO) {
+				addUTXO(lastUTXO)
+				okUTXOs = okUTXOs[0 : len(okUTXOs)-1]
+				continue
 			}
```
buck54321, line 1800: Moving this check up up here was the critical improvement. Before, we iterated the utxos from smallest to largest to find the smallest utxo that satisfied our needs. If none worked, we'd add the last element and try again. But for `MaxOrder`, we're using all the utxos, so we'd run the enough function `N + (N - 1) + (N - 2) + ... + 1` = `N * (N + 1) / 2`, or roughly `N²/2`. With the new pattern, it's just `N`.

chappjc:
These are smart changes, but I don't think it quite resolves https://github.com/decred/dcrdex/issues/1612, at least in my case.  Initially I thought it must be all this UTXO selection code you're fixing up, but it turned out to be a combination of:
- the [one listunspent call taking forever](https://github.com/decred/dcrdex/issues/1612#issuecomment-1143911878), a wallet bottleneck
- frontend hitting the maxorder endpoint too liberally, almost simultaneously

To see what I mean for the first bullet you really need a testnet wallet with hundreds of thousands of utxos.

buck54321:
> These are smart changes, but I don't think it quite resolves #1612, at least in my case. Initially I thought it must be all this UTXO selection code you're fixing up, but it turned out to be a combination of:
> 
> * the [one listunspent call taking forever](https://github.com/decred/dcrdex/issues/1612#issuecomment-1143911878), a wallet bottleneck
> * frontend hitting the maxorder endpoint too liberally, almost simultaneously
> 
> To see what I mean for the first bullet you really need a testnet wallet with hundreds of thousands of utxos.

Try it out to be sure. I tested with 8000 utxos from the decred simnet wallet, and it shortened the times by a few orders of magnitude. Do you know how many utxos you had?

re frontend spamming: I've observed that too, but was unable to reproduce after the fix. I can test some more.

chappjc:
Yeah it was definitely sitting on listunspent in that issue, although maxSell, not maxBuy.

Regarding the frontend metering, I think it would  ideally be the kind of request that gets fired after nothing has been typed for say 250 ms, similar to many live search boxes on the internet.  Instead, how it behaves is if you want to type out 0.001125 in the rate, each digit (after it becomes a valid rate) pressed _immediately_ sends the maxsell request, and then it'll be metering by the time you get to "5".

chappjc:
That last push solved the maxSell hammering for me.

The maxbuy backend times are also still good.

buck54321:
[608921a](https://github.com/decred/dcrdex/pull/1684/commits/608921a310dbd5860356805776c6a84b79753359) should make the behavior as you've requested. We needed to update the `maxsell` on rate field change anyway, since we're using that rate for limit orders now.

chappjc:
Perfection! `take 2` makes me sooo much less confused about those things.

chappjc (approved):
Test well.  Leaving up for a bit for other reviews.

client/webserver/site/src/js/markets.ts
```diff
@@ -884,8 +885,11 @@ export default class MarketsPage extends BasePage {
       this.setMaxOrder(mkt.maxSell.swap)
       return
     }
+    if (mkt.maxSellRequested) return
```
chappjc, line 888: This is the critical bit that really resolves https://github.com/decred/dcrdex/issues/1612 for me

client/asset/dcr/dcr.go
```diff
@@ -1348,22 +1352,25 @@ func (dcr *ExchangeWallet) tryFund(utxos []*compositeUTXO, enough func(sum uint6
...
-					if err = addUTXO(txout); err != nil {
-						return false, err
-					}
-					return true, nil
-				}
+
+			// Check if the largest output is too small.
+			lastUTXO := okUTXOs[len(okUTXOs)-1]
+			if !isEnoughWith(lastUTXO) {
+				addUTXO(lastUTXO)
```
chappjc, line 1359: We check the error from `addUTXO` below, and both places previously.

Although thinking about this, one utxo with a screwed up txid or redeem script would ideally not break utxo selection entirely.  How about at the top of `tryFund` we do a last minute check and re-parse of the `compositeUTXO` slice before the `tryUTXOs` calls, generating a `[]parsedUTXO` with a local type like

```go
type parsedUTXO struct {
	hash         *chainhash.Hash
	vout         uint32
	redeemScript []byte
	amt          uint64
	tree         int8
	inSize       uint32
}
```

And refactor `addUTXO` to take that type and return no error.

chappjc: We could even go all the way back to `spendableUTXOs`, but I would kinda like to avoid doing that.

JoeGruffins (approved):

client/asset/dcr/dcr.go
buck54321: Just checked the error for now.

PR #1685, opened 2022-06-24
app: refactor frontend notification flow

buck54321:
I've wanted to do this for a while, but its need became more apparent in #1622. This prevents the need to pass notifications to forms.


chappjc (approved):
Works great.

Issue #1492, opened 2022-02-28
client: all state, including locked funds, are lost upon reconfiguring wallet.


JoeGruffins:
When our wallet is in control of tracking locked funds, they are lost on re-configuring the wallet. To reproduce, make an order selling eth or btc, then go to wallet options and save. The funds will be unlocked. Attempting to use the funds afterwards will result in an error from the server's coin locker.

Before:
![image](https://user-images.githubusercontent.com/34998433/155936173-4fa82dff-2856-4f26-b2c5-9b3188884ef6.png)

After:
![image](https://user-images.githubusercontent.com/34998433/155936260-6341b12b-f731-4f7b-8847-c561b220e965.png)

chappjc:
I'm gonna prototype some methods as described in my last comment, but it would be great if this issue had proposed resolutions for multiple people, with reference to each of the wallet types (dcr full, dcr spv, btc full, btc spv, eth light)

chappjc:
I wanted to get to this, but got swamped.  Fair game for all

buck54321:
I'm gonna take a swing at this.

chappjc:
Mostly resolved by `(*Core).reReserveFunding` added in https://github.com/decred/dcrdex/pull/1686.
Live reconfig support can be added for the other assets, but I think this issue is closed because it refers to the state loss that is resolved for all assets.

PR #1686, opened 2022-06-25
implement live wallet reconfiguration

buck54321:
Starts resolving #1492 . Just sketching this out for now. I can add other wallets here, or that can be followup work. The btc changes are in anticipation of changes coming in #1657. 

```
Our current scheme for configuration changes requires killing the
existing wallet and creating a new one. Unfortunately, even though
we allow reconfiguration with active matches/orders, we've made
no accommodations for re-establishing locked coins on the new wallet.

client/asset:
Introduce the LiveReconfigurer interface with a single
method, Reconfigure(WalletConfig) (restartRequired bool, err error).
If a wallet implements LiverReconfigurer,

client/core:
Core will call Reconfigure early in (*Core).ReconfigureWallet, and
if there is no error and restartRequired is false, then the
reconfiguration is considered complete and no new wallet is created.
If, on the other hand, the LiveReconfigurer indicates that a restart
is required, we use our existing ReconfigureWallet code without
modification, but add a step at the end to re-reserve funding coin
and refund/redeem reserves. This new method (*Core).reReserveFunding
is modeled closely after resumeTrades, but optimized for only 1 asset
an no audits.

client/asset/btc:
Implement Reconfigure. Puts user-configurable baseWallet configuration
in a seperate struct, stored as an atomic.Value baseWallet field, with
the settings themselves readable through new getters. rpcWallet does
something similar with its an atomic.Value for its requester
RawRequesterWithContext field. The only thing that requires a restart
now is a wallet type change, though an error is generated if there is
an attempt to reduce the spv wallet birthday with special:activelyUsed set.
```


chappjc:
Looks like you're building on https://github.com/decred/dcrdex/pull/1657, in spirit.  I'm fine with that one so wanna merge that so this can get rebased?

dex/config/config_test.go
```diff
@@ -262,3 +262,60 @@ func DataConversion(t *testing.T) {
...
+
+	settings := map[string]string{
+		"a": "1",
+		"b": "2",
+	}
+
+	// Embedded value.
+	cfg := &struct {
+		A int `ini:"a"`
+		BB
```
chappjc, line 279: I think the `ini` package supports this out of the box with `ini:",extends"` as the tag.

i.e.

```go
cfg := &struct {
	A  int `ini:"a"`
	BB `ini:",extends"`
}{}
```

and we can keep `ParseInto` unchanged, calling `section.MapTo(obj)` once directly.

buck54321: How did you find this?

chappjc: Had a vague recollection of https://github.com/go-ini/ini/commit/5e97220809ffaa826f787728501264e9114cb834 while reviewing the commits when I bummed it to 1.62 in https://github.com/decred/dcrdex/pull/1606/commits/9c461e34c72b3446950f495f6196e2a1a31544f7#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R51

buck54321: I do kinda with they would have just done it my way.

buck54321:
I'm opening this up for review. I can add other assets here, but following up is good too.

dex/config/config_test.go
chappjc: I hear you about the that.  Not sure why a tag is needed.

Regardless, I don't think this is the kind of thing we should get side tracked with even if they didn't have this tag. It's more than fine to just call `Unmapify` twice, once on the parent and once on the embedded field.  Either that or we just don't compose a data-only struct like this with embedding to avoid extra complexity of the pre-processing.  A little repetition to avoid complexity and non-trivial code to maintain is fine.  The flatten function with the reflection code calls `section.MapTo` repeatedly anyway right? 

New issue because I think we have some unhealthy patterns established that create the need for these acrobatics in the first place: https://github.com/decred/dcrdex/issues/1698

martonp (commented):
Works well. I just see two minor UI issues:

1. If I try to switch wallet types while I have an active order, it fails with an error message as it should, but on the UI the wallet is shown as locked and unsynced. This fixes itself after refreshing the screen.

3. If I'm trying to switch from the SPV wallet to the external wallet while I have a trade running, it fails as it should, but if I navigate away to the markets page, then come back to the reconfigure screen, the RPC wallet settings are still showing even though I'm not using an SPV wallet. When I'm using the RPC wallet and I fail at switching to the SPV wallet, then I navigate away and come back to the reconfigure screen, the RPC wallet configs are still showing as they are supposed to. The issue only happens when attempting to switch from SPV to RPC.

JoeGruffins (commented):
Looks fine and tests well.

dex/config/config_test.go
buck54321: I've used the ``ini:",extends"`` tag. Do you not want to use it?

chappjc: No please use it!  I'm just thinking we need to look to fix our use of Unmapify and the ini package in the future.

client/asset/btc/btc.go
```diff
@@ -1636,16 +1700,16 @@ func (btc *baseWallet) FundOrder(ord *asset.Order) (asset.Coins, []dex.Bytes, er
...
 	}
 	// Check wallets fee rate limit against server's max fee rate
-	if btc.feeRateLimit < ord.DEXConfig.MaxFeeRate {
+	if btc.feeRateLimit() < ord.DEXConfig.MaxFeeRate {
 		return nil, nil, fmt.Errorf(
 			"%v: server's max fee rate %v higher than configued fee rate limit %v",
 			ord.DEXConfig.Symbol,
 			ord.DEXConfig.MaxFeeRate,
-			btc.feeRateLimit)
+			btc.feeRateLimit())
```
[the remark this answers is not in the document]
buck54321: I'm okay with it if it's just for the error log.

client/core/core.go
```diff
@@ -2336,6 +2351,84 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, form *WalletForm) er
...
+		}
+
+		c.notify(newBalanceNote(assetID, balances)) // redundant with wallet config note?
+		subject, details := c.formatDetails(TopicWalletConfigurationUpdated, unbip(assetID), w.address)
+		c.notify(newWalletConfigNote(TopicWalletConfigurationUpdated, subject, details, db.Success, w.state()))
+
+		return nil
+	}
+
+	clearTickGovernors := func() {
```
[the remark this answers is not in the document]
buck54321: The tick governor meters attempts at swapping when there is repeating error. A new wallet might fix the issue, so we allow a tick to occur immediately.

go.mod
```diff
@@ -54,7 +54,7 @@ require (
 	golang.org/x/term v0.0.0-20210927222741-03fcf44c2211
 	golang.org/x/text v0.3.7
 	golang.org/x/time v0.0.0-20220411224347-583f2d630306
-	gopkg.in/ini.v1 v1.62.0
+	gopkg.in/ini.v1 v1.66.4
```
chappjc, line 57: conflicts on the go.mod.

These updates seem fine, but curious if they was a specific motivation for the bumps?

chappjc: Oh, nvm no conflicts.  I had an old branch, sorry.

client/asset/btc/rpcclient.go
```diff
@@ -136,9 +144,64 @@ func (wc *rpcClient) connect(ctx context.Context, _ *sync.WaitGroup) error {
...
+			if err := call(wc.ctx, cl, method, anylist{currentAddress}, ai); err != nil {
+				return false, fmt.Errorf("error getting address info with new RPC credentials: %w", err)
+			} else if !ai.IsMine {
+				return false, errors.New("cannot reconfigure to a new RPC wallet during active use")
+			}
+		}
+
+		wc.requesterV.Store(cl)
+
+		// No restart required
```
chappjc, line 197: Seems like we need to update `wc.rpcConfig` though.  Fine to do it without concurrency controls because construction and calls to reconfigure can simply be declared not thread safe.

client/asset/btc/rpcclient.go
```diff
@@ -136,9 +144,64 @@ func (wc *rpcClient) connect(ctx context.Context, _ *sync.WaitGroup) error {
...
+	if err = dexbtc.CheckRPCConfig(&newCfg.RPCConfig, wc.cloneParams.WalletInfo.Name, wc.cloneParams.Network, wc.cloneParams.Ports); err != nil {
+		return
+	}
+
+	// If the RPC configuration has changed, try to update the client.
+	oldCfg := wc.rpcConfig
+	if newCfg.RPCBind != oldCfg.RPCBind ||
+		newCfg.RPCUser != oldCfg.RPCUser ||
+		newCfg.RPCPass != oldCfg.RPCPass ||
+		newCfg.WalletName != oldCfg.WalletName {
```
chappjc, line 174: I think we can directly compare the structs, never have to worry about added fields as long as it's a plain-old-data struct with no pointers. `if *oldCfg == *newCfg { return }`?

client/asset/btc/rpcclient.go
```diff
@@ -136,9 +144,64 @@ func (wc *rpcClient) connect(ctx context.Context, _ *sync.WaitGroup) error {
...
+		}
+
+		// If the wallet is in active use, check the supplied address.
+		if parsedCfg.ActivelyUsed {
+			method := methodGetAddressInfo
+			if wc.legacyValidateAddressRPC {
+				method = methodValidateAddress
+			}
+			ai := new(GetAddressInfoResult)
+			if err := call(wc.ctx, cl, method, anylist{currentAddress}, ai); err != nil {
```
chappjc, line 188: I guess you're not using `ownsAddress` to skip making a `btcutil.Address`?  Seems alright, just stands out.

client/asset/btc/rpcclient.go
```diff
@@ -136,9 +144,64 @@ func (wc *rpcClient) connect(ctx context.Context, _ *sync.WaitGroup) error {
...
+	}
+
+	parsedCfg := new(RPCWalletConfig)
+	if err = config.Unmapify(cfg.Settings, parsedCfg); err != nil {
+		return
+	}
+
+	// Check the RPC configuration.
+	newCfg := &parsedCfg.RPCConfig
+	if err = dexbtc.CheckRPCConfig(&newCfg.RPCConfig, wc.cloneParams.WalletInfo.Name, wc.cloneParams.Network, wc.cloneParams.Ports); err != nil {
```
chappjc, line 165: Would you please insert a newline after `.cloneParams.WalletInfo.Name`?

client/asset/btc/btc.go
```diff
@@ -599,23 +607,61 @@ func init() {
...
+// baseWalletConfig is the validated, unit-converted, user-configurable wallet
+// settings.
 type baseWalletConfig struct {
 	fallbackFeeRate  uint64 // atoms/byte
 	feeRateLimit     uint64 // atoms/byte
 	redeemConfTarget uint64
 	useSplitTx       bool
 }
 
+func readBaseWalletConfig(walletCfg *WalletConfig) (*baseWalletConfig, error) {
```
chappjc, line 619: Since you're moving this function, I'd prefer if you put it above `readRPCWalletConfig`, before its first use.

client/asset/btc/btc.go
```diff
@@ -1075,6 +1107,37 @@ func (btc *baseWallet) shutdown() {
...
+	// when the caller attempts to open the wallet. If no restart is required,
+	// we'll swap out the configuration parameters right away.
+	newCfg, err := readBaseWalletConfig(walletCfg)
+	if err != nil {
+		return false, err
+	}
+
+	if !restart {
+		// No restart required. Just update the configuration.
+		btc.cfgV.Store(newCfg)
```
chappjc, line 1135: So we're on the same page, `restart` means this `baseWallet` is going to the trash. (right?) Which means we don't *have* to store the new `baseWalletConfig`.
I guess what I'm getting at is: it could be always:

```go
btc.cfgV.Store(newCfg) // probably trash if restart/reinit required
return restart, nil
```

My thinking is that if we've returned a nil error, and the caller is in charge of restarting... in the meantime... should we apply the new config items assuming the wallet is still live or the caller might not want to restart fully but still take the new settings they requested?

client/core/wallet.go
```diff
@@ -55,6 +55,8 @@ type xcWallet struct {
 	hookedUp     bool
 	synced       bool
 	syncProgress float32
+
+	broadcasting *uint32
```
chappjc, line 59: Will you doc this field please?  I've been looking at core.go and I can't really tell the intend.  Let wallet state notes out?

client/asset/btc/rpcclient.go
buck54321: It's because `ownsAddress` is a `rpcClient` method that uses the stored `RawRequesterWithContext`. We don't update the `requesterV` until after this validation.

chappjc: Ahhh, I see now.

chappjc (approved):
LGTM.  @JoeGruffins please circle back on your review when convenient.  Thx.

client/core/core.go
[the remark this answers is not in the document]
JoeGruffins: Ok. I didn't realize this method is expected to be used if something was erroring. No problem.

JoeGruffins:
Switching wallet types, rpc to spv looks fine, but spv to rpc will say I have no peers and not let me trade until I restart dexc.

JoeGruffins (approved):

PR #1692, opened 2022-07-01
client/eth: Increase gas fee cap for redemptions when base fee is too high

martonp:
This diff increases the gas fee cap for redemptions to 2x base fee if the base fee > the fee suggestion. If the wallet does not have enough available funds for this, then the funds that are available will be used.

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -1439,17 +1439,23 @@ func (w *assetWallet) Redeem(form *asset.RedeemForm, feeWallet *assetWallet) ([]
 
 	g := w.gases(contractVer)
 	if g == nil {
-		return nil, nil, 0, fmt.Errorf("no gas table")
+		return fail(fmt.Errorf("no gas table"))
 	}
 
-	// Get a redemption gas estimate.
+	if feeWallet == nil {
+		feeWallet = w
+	}
+	bal, err := feeWallet.Balance()
+	if err != nil {
+		return nil, nil, 0, fmt.Errorf("error getting balance in excessive gas fee recovery: %v", err)
+	}
+
+	gasLimit, gasFeeCap := g.Redeem*n, form.FeeSuggestion
+	originalFundsReserved := gasLimit * gasFeeCap
 	gasEst, err := w.estimateRedeemGas(w.ctx, secrets, contractVer)
 	if err != nil {
-		return nil, nil, 0, fmt.Errorf("error getting redemption estimate: %v", err)
+		return fail(fmt.Errorf("error getting redemption estimate: %w", err))
 	}
-
-	// If our estimate is higher than the calculated limit
-	gasLimit := g.Redeem * n
 	if gasEst > gasLimit {
 		// This is sticky. We only reserved so much for redemption, so accepting
 		// a gas limit higher than anticipated could potentially mess us up. On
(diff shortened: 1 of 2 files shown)


martonp:
This should go in before #1638.

client/asset/eth/eth.go
```diff
@@ -1439,17 +1439,23 @@ func (w *assetWallet) Redeem(form *asset.RedeemForm, feeWallet *assetWallet) ([]
...
+	}
+	bal, err := feeWallet.Balance()
 	if err != nil {
-		return nil, nil, 0, fmt.Errorf("error getting redemption estimate: %v", err)
+		return nil, nil, 0, fmt.Errorf("error getting balance in excessive gas fee recovery: %v", err)
 	}
 
-	// If our estimate is higher than the calculated limit
-	gasLimit := g.Redeem * n
+	var gasLimit, gasFeeCap uint64 = g.Redeem * n, form.FeeSuggestion
```
chappjc, line 1453: Shorthand is fine with this multiple assignment.

client/asset/eth/eth.go
```diff
@@ -1459,24 +1465,35 @@ func (w *assetWallet) Redeem(form *asset.RedeemForm, feeWallet *assetWallet) ([]
 		candidateLimit := gasEst * 11 / 10 // Add 10% for good measure.
 		// Cannot be more than double.
 		if candidateLimit > gasLimit*2 {
-			return nil, nil, 0, fmt.Errorf("cannot recover from excessive gas estimate %d > 2 * %d", candidateLimit, gasLimit)
+			return fail(fmt.Errorf("cannot recover from excessive gas estimate %d > 2 * %d", candidateLimit, gasLimit))
```
chappjc, line 1468: What was this double rule about?

chappjc (commented):
Concept ACK.  Untested.

client/asset/eth/eth.go
martonp: If the estimate for the amount of gas the redemption will use is more than double the hardcoded value, then it will fail.

chappjc: Fail where? Why?

martonp: It won't submit the redemption transaction. The redemption function just returns an error. This is the current logic.

martonp: This shouldn't happen though since there's only one code path in the contract.

martonp: I meant that this function would fail. It is completely arbitrary for sure.. do you think it should be higher than 2x?

chappjc: Ohhhh.  Ok that's fine.  The comments suggested to me that there's something special about this threshold.  It's fine.  Could be less even, let's just keep it this way though.

JoeGruffins (approved):
Sorry, was in the middle of looking at this so leaving what I had. Looks good to me.

client/asset/eth/eth.go
```diff
@@ -1459,24 +1465,35 @@ func (w *assetWallet) Redeem(form *asset.RedeemForm, feeWallet *assetWallet) ([]
...
 		gasLimit = candidateLimit
 	}
 
-	tx, err := w.redeem(w.ctx, w.assetID, form.Redemptions, form.FeeSuggestion, gasLimit, contractVer)
+	// If the base fee is higher than the FeeSuggestion we attempt to increase
+	// the gasFeeCap to 2*baseFee. If we don't have enough funds, we use the
+	// funds we have available.
+	baseFee, _, err := w.node.currentFees(w.ctx)
+	if err != nil {
+		return fail(fmt.Errorf("Error getting net fee state: %w", err))
```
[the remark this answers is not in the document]
chappjc: Good points @JoeGruffins.  Sorry to interrupt your review.  @martonp can fold these revisions into https://github.com/decred/dcrdex/pull/1638

PR #1694, opened 2022-07-03
rpcserver: parse options on TradeForm

LasTshaMAN:


client/rpcserver/types.go
```diff
@@ -208,6 +208,14 @@ func checkBoolArg(arg, name string) (bool, error) {
 	return b, nil
 }
 
+func checkMapArg(arg, name string) (map[string]string, error) {
+	m, err := config.Parse([]byte(arg))
+	if err != nil {
+		return nil, fmt.Errorf("%w: %s must be a map[string]string: %v", errArgs, name, err)
```
chappjc, line 214: Is `arg` really a `map[string]string` as this error indicates?  I think the input to `config.Parse` is an *ini* text file data (when not a file name).

If the arg were a text representation of a map, it would be like `{"gas_price": 23}`, not `gas_price=23`.

With the ini format, is it even possible to provide more then one `k=v` option?

LasTshaMAN: This is modelled after `newWalletForm.config` which also expects `map[string]string`, the _ini_ format might not be the best fit for it, but I thought it is better to keep to the same format for the same data types (`map[string]string` in this case),

but if you want it to be JSON (like in your example `{"gas_price": 23}`, I can change it of course, just think the discrepancy with `newWalletForm.config` isn't worth it)

> With the ini format, is it even possible to provide more then one k=v option? Would have to use \n and it would look funny I think.

yeah, probably should have covered multi-value case right away (just was somewhat lazy)

chappjc: Look at the handler for the `newwallet` RPC.  `parseNewWalletArgs` gets the `map[string]string` using `json.Unmarshal` _from JSON_, not `config.Parse` from ini formatted data.

See [how it's called](https://github.com/decred/dcrdex/blob/add7978ed055df3fc983420c86f8852b6820435e/client/cmd/dexcctl/sample-dexcctl.conf#L15) with JSON for the options:

`dexcctl newwallet 42 ~/dextest/dcr/beta/beta.conf '{"account":"default"}'``

So by expecting ini text, this is actually creating a discrepancy.

chappjc: Ah, fair enough.  Looks like `TestParseNewWalletArgs` was never updated for the config options map (4th arg).  With respect to the ini arg (3rd), I think the test was written that way because it was more friendly for the tests than an .conf/.ini file, but the intended use for that arg is a path, while the settings are JSON:

`newwallet assetID walletType ("path" "settings")`

```
path (string): Optional. The path to a configuration file.
settings (string): A JSON-encoded string->string mapping of additional
```

client/rpcserver/types.go
```diff
@@ -400,6 +409,10 @@ func parseTradeArgs(params *RawParams) (*tradeForm, error) {
 	if err != nil {
 		return nil, err
 	}
+	options, err := checkMapArg(params.Args[8], "options")
```
chappjc, line 412: Please update the help doc in handlers.go for `tradeRoute:`.  It can read similarly to the `settings` arg description for `newWalletRoute`.

chappjc (changes requested):

client/rpcserver/handlers.go
```diff
@@ -812,8 +812,8 @@ var helpMsgs = map[string]helpMsg{
     assetID (int): The asset's BIP-44 registered coin index. e.g. 42 for DCR.
       See https://github.com/satoshilabs/slips/blob/master/slip-0044.md
     walletType (string): The wallet type.
-     path (string): Optional. The path to a configuration file.
-     settings (string): A JSON-encoded string->string mapping of additional
+    path (string): Optional. The path to a configuration file.
+    settings (string): A JSON-encoded string->string mapping of additional
```
LasTshaMAN, line 816: Looks like formatting here was slightly off: (took liberty to fix it, since I noticed)

image

client/rpcserver/types.go
LasTshaMAN: Should be fixed now

chappjc (approved):

PR #1699, opened 2022-07-06
CI: enforce fmt simplification

chappjc:
In catching up with recent PRs I noticed some churn because `gofmt` has a "simplify" option that removes some superfluous identifiers, but CI does not enforce its use and only some developers have it set.  I don't see a way to enforce "not simple" so this PR enforces simple for consistency.  This is done by enabling the `gofmt` linter in golangci-lint:

`gofmt: Gofmt checks whether code was gofmt-ed. By default this tool runs with -s option to check for code simplification [fast: true, auto-fix: true]`

Developers may choose to add that as well, or just add `-s` to their gofmt flags.

This also modifies the the erc20 updatecontract.sh script so that it reorders the imports after rewriting the import lines with perl.  Recent PRs have tried to modify contract.go, but since it is generated it should not be touched (except by the script that generates it).

### DIFF
--- a/client/asset/btc/btc_test.go
+++ b/client/asset/btc/btc_test.go
@@ -800,7 +800,7 @@ func testAvailableFund(t *testing.T, segwit bool, walletType string) {
 	blockHash, _ := node.addRawTx(blockHeight, msgTx)
 
 	node.getTransactionMap = map[string]*GetTransactionResult{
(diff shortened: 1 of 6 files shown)


martonp (commented):
Shouldn't this be updated in `run_tests.sh`?

martonp (approved):

PR #1700, opened 2022-07-07
ui: restyle wallets page

buck54321:
The wallets page is restyled. The new style is responsive.

**Desktop**

**Tablet**

**Mobile**


client/webserver/locales/pl-pl.go
```diff
@@ -163,13 +163,11 @@ var PlPL = map[string]string{
...
 	"Import Account":              "Importuj konto",
 	"no_wallet":                   "brak portfela",
-	"create_a_x_wallet":           "Utwórz portfel {{.Name}}", // .Name will be an asset symbol like DCR
+	"create_a_x_wallet":           "Utwórz portfel <span data-asset-name=1></span>",
 	"dont_share":                  "Nie udostępniaj nikomu. Nie zgub go.",
 	"Show Me":                     "Pokaż",
 	"Wallet Settings":             "Ustawienia portfela",
 	"add_a_x_wallet":              `Dodaj portfel <img data-tmpl="assetLogo" class="asset-logo mx-1"> <span data-tmpl="assetName"></span>`,
-	"ready":                       "gotowy",
-	"off":                         "wyłączony",
```
chappjc, line 172: Surely these off and ready are still good, capitalization aside, and I think we can feel out if caps can be done on a per-language basis (pl and pt-br, but not zh?)

client/webserver/site/src/html/bodybuilder.tmpl
```diff
@@ -5,10 +5,10 @@
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   {{- /* The above 2 meta tags *must* come first in the head; any other head content must come *after* these tags */ -}}
-  <link rel="icon" href="/img/favicon.png?v=AK4XS4">
+  <link rel="icon" href="/img/favicon.png?v=AK4XS55">
```
chappjc, line 8: I don't see any changes to favicon.png

client/webserver/site/src/js/app.ts
```diff
@@ -615,6 +613,8 @@ export default class Application {
         f(note)
       } catch (error) {
         console.error('note feeder error:', error.message ? error.message : error)
+        console.log(note)
+        console.log(error.stack)
```
chappjc, line 617: leftovers or you want this because it shouldn't happen?

client/webserver/site/src/js/doc.ts
```diff
@@ -368,13 +357,70 @@ export default class Doc {
   }
 }
 
+export class Animation {
+  done: (() => void) | undefined
+  endAnimation: boolean
+  thread: Promise<void>
+  Forever: number
+
+  constructor (duration: number, f: (progress: number) => void, easingAlgo?: string, done?: () => void) {
```
chappjc, line 366: Looks like a neat reusable class.  Perhaps a little docs?

client/webserver/site/src/js/wallets.ts
```diff
@@ -346,58 +257,255 @@ export default class WalletsPage extends BasePage {
...
-    this.animation = this.showBox(box)
+    page.successMessage.textContent = msg
+    this.currentForm = page.checkmarkForm
+    this.forms.forEach(form => Doc.hide(form))
+    Doc.show(page.forms, page.checkmarkForm)
+    page.checkmarkForm.style.right = '0'
+    page.checkmark.style.fontSize = '0px'
+
+    const [startR, startG, startB] = State.isDark() ? [223, 226, 225] : [51, 51, 51]
+    const [endR, endG, endB] = [16, 163, 16]
```
chappjc, line 270: I'm not one to nitpick frontend, but I would expect you'd pull these values out of a document element with scss dictating these values.
That said, this is pretty darn simple, but screams style rather than function.

client/webserver/site/src/js/wallets.ts
```diff
@@ -346,58 +257,255 @@ export default class WalletsPage extends BasePage {
...
+        tmpl.volumeUnit.textContent = assetID === baseid ? basesymbol.toUpperCase() : quotesymbol.toUpperCase()
+      } else Doc.hide(tmpl.priceBox, tmpl.volumeBox)
+      Doc.bind(row, 'click', () => app().loadPage('markets', { host, base: baseid, quote: quoteid }))
+    }
+    page.marketsOverviewBox.classList.remove('invisible')
+  }
+
+  async showRecentActivity (assetID: number) {
+    const page = this.page
+    // const loaded = app().loading(this.main)
```
chappjc, line 457: Planned to use?

client/webserver/site/src/js/wallets.ts
```diff
@@ -346,58 +257,255 @@ export default class WalletsPage extends BasePage {
...
+        }
+      }
+
+      tmpl.fromLogo.src = Doc.logoPath(from)
+      Doc.empty(tmpl.fromSymbol, tmpl.toSymbol)
+      tmpl.fromSymbol.appendChild(Doc.symbolize(from))
+      tmpl.toLogo.src = Doc.logoPath(to)
+      tmpl.toSymbol.appendChild(Doc.symbolize(to))
+      tmpl.status.textContent = OrderUtil.statusString(ord)
+      tmpl.filled.textContent = `${(ord.filled / ord.qty * 100).toFixed(1)}%`
```
chappjc, line 502: Need to use the new `OrdreUtil.filled` for this percent, from https://github.com/decred/dcrdex/pull/1744

client/webserver/site/src/js/wallets.ts
```diff
@@ -346,58 +257,255 @@ export default class WalletsPage extends BasePage {
...
+      // tmpl.settled.textContent = `${(OrderUtil.settled(ord) / ord.qty * 100).toFixed(1)}%`
```
chappjc, line 503: I think it's reasonable to exclude settled from the wallet's page recent activity since the status would say settling.

client/webserver/site/src/js/wallets.ts
```diff
@@ -571,14 +667,7 @@ export default class WalletsPage extends BasePage {
   async showSendForm (assetID: number) {
     const page = this.page
     const box = page.sendForm
-    const asset = this.sendAsset = app().assets[assetID]
-    this.lastFormAsset = assetID
-    const wallet = app().walletMap[assetID]
-    if (!wallet) {
-      app().notify(ntfn.make('Cannot send/withdraw.', `No wallet found for ${asset.name}`, ntfn.ERROR))
```
chappjc, line 578: Was this impossible to hit before?

client/webserver/site/src/js/wallets.ts
```diff
@@ -595,65 +684,51 @@ export default class WalletsPage extends BasePage {
...
   }
 
   /* send submits the send form to the API. */
-  async send () {
+  async send (): Promise<void> {
     const page = this.page
     Doc.hide(page.sendErr)
-    const assetID = parseInt(page.sendForm.dataset.assetID || '')
-    const subtract = page.subtractCheckBox.checked || false
+    const assetID = parseInt(page.sendForm.dataset.assetID ?? '')
```
chappjc, line 724: Eeeek have we broadly misusing `||` when we should have been using the nullish coalescing operator?

client/asset/interface.go
```diff
@@ -568,6 +568,8 @@ type Balance struct {
 	// Locked is the total amount locked in the wallet which includes but
 	// is not limited to funds locked for swap but not actually swapped yet.
 	Locked uint64 `json:"locked"`
+	// Other is a place to list custom balance categories.
+	Other map[string]uint64 `json:"other"`
```
chappjc, line 572: I can see the need for this.  Has come up already.  What's the immediate first use that you have in mind out of curiosity? (none that I see in this PR)

chappjc (changes requested):
Looks great just based on code.  Will actually test next.

I see what you mean about the localized_html bloat. Agree we should look into those done on-the-fly after we get the initial `go:embed` stuff in.

client/asset/interface.go
chappjc: Oh, the fee balance for tokens?

buck54321: I was actually thinking about Zcash shielded balances and Decred mixing wallets.

client/webserver/site/src/localized_html/en-US/wallets.tmpl
```diff
@@ -220,39 +313,37 @@
...
       <hr class="dashed my-2">
-      <div class="d-flex flex-row">
-        <button id="downloadLogs" type="button" class="w-25 mt-1 mx-1 justify-content-center fs15 bg2">Wallet Logs</button>
+      <div class="sans-light fs17">Other Actions</div>
+      <div class="d-flex flex-row fs15 pt-1">
+        <button id="downloadLogs" type="button" class="p-1 me-2 hoverbg">Wallet Logs</button>
         <button id="exportWallet" type="button" class="w-25 mx-1 mt-1 justify-content-center fs15 bg2">Export Wallet</button>
-        <button id="recoverWallet" type="button" class="danger w-25 mx-1 mt-1 justify-content-center fs15 bg2">Recover</button>
+        <button id="recoverWallet" type="button" class="danger p-1 me-2 hoverbg">Recover</button>
+        <button id="rescanWallet" type="button" class="danger p-1 me-2 hoverbg">Rescan</button>
```
chappjc, line 344: When I click Rescan, there's no indication that it's rescanning.  Maybe close the reconfig form, sending them back to see the sync status icon on the wallets list?

chappjc: I'm seeing a Rescan button for wallets that don't support rescanning.

client/webserver/site/src/localized_html/en-US/wallets.tmpl
```diff
@@ -1,154 +1,222 @@
...
+                    <td data-tmpl="status"></td>
+                    <td data-tmpl="filled" class="d-none d-md-table-cell"></td>
+                    <td data-tmpl="age" class="text-nowrap"></td>
+                    <td><a class="ico-open fs14 pointer plainlink" data-tmpl="link"></a></td>
+                  </tr>
+                </tbody>
+              </table>
+            </div>
+          </div>
+          <div class="m-auto flex-center flex-column w-100" id="noActivity">
```
chappjc, line 196: This div looks better with a `py-3` like the `orderActivity` div that replaces it once there's activity.

chappjc (changes requested):
Tests well overall, just a few minor things.

And the following is not an issue with this PR, but there's inconsistent terminology with the built-in wallets.  We call them "Native" for DCR and BTC, but "Internal" for ETH:

image

image

This is just something we don't want the user scratching their head over.

Also in that BTC screen, it's a whole number so there are no decimals, not even a single ".0".  I guess that's fine if intended, but looks a little unusual for wallet displays.

client/webserver/site/src/js/wallets.ts
buck54321: Meh. I just discovered this, really. Will try to use it more in the future, but it's already bitten me. I did something like

```Javascript
const v = parseFloat(inputElement.value ?? '0')
```

which yields `NaN`, because `value` is not `undefined` or `null`, just empty, which is falsy but not nullish.

I think we should use it, but I'm not going to enforce it **too** strictly.

---

I've also been using the **optional chaining operator**, `?.`.

```Javascript
const thing = { a: 1 }
let v = thing.b.c    // error thrown cuz b undefined
v = thing?.b?.c      // undefined

```

chappjc:
Thanks for the review followup.  Just want to make sure you saw that I had a prior review: https://github.com/decred/dcrdex/pull/1700#pullrequestreview-1070071646  No rush but I got the sense you just were only addressing the second review from my testing.  The order filled percent is the main outstanding thing.

client/webserver/site/src/js/wallets.ts
buck54321: Getting r,g,b values from css into javascript is not straightforward. I didn't think it was worth the string parsing required.

chappjc: Figured.  Don't waste time.

client/webserver/site/src/js/wallets.ts
buck54321: I believe so.

chappjc (approved):
OK.  I thought we could just capitalize the other languages where it looks safe, but whatever

PR #1701, opened 2022-07-07
client: add peers restored message and improve error

chappjc:
1. Add a peers restored message to match a peers lost message
 
image

2. Distinguish peers lost from unable to determine peer count (e.g. rpc timeout)

image

3. Fix a bug in the DCR block monitoring goroutine.  When native SPV was added, some code that made sense in its own function (defer and return) no longer worked properly when put directly into `monitorBlocks`.  The return was killing the wallet if connectivity was lost, instead of it being a temporary issue.  This fixes those issues.

image

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -621,7 +621,7 @@ type baseWallet struct {
 	symbol            string
 	tipChange         func(error)
 	lastPeerCount     uint32
-	peersChange       func(uint32)
+	peersChange       func(uint32, error)
 	minNetworkVersion uint64
 	dustLimit         uint64
 	useLegacyBalance  bool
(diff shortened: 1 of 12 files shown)


client/asset/dcr/dcr.go
```diff
@@ -3498,52 +3498,62 @@ func (dcr *ExchangeWallet) monitorBlocks(ctx context.Context) {
...
 
-	for {
-		select {
-		case <-ticker.C:
-			ctx, cancel := context.WithTimeout(dcr.ctx, 2*time.Second)
-			defer cancel()
-			newTip, err := dcr.getBestBlock(ctx)
-			if err != nil {
-				dcr.handleTipChange(nil, 0, fmt.Errorf("failed to get best block: %w", err))
-				return
```
chappjc, line 3509: This return was the bug that resulted in "client has been shutdown" and unable to reconnect to the wallet without dexc restart.  The `defer cancel()` was also not working as intended, but also filling up the defer stack leading to memory exhaustion.

client/asset/btc/btc.go
```diff
@@ -1809,11 +1809,11 @@ func fund(utxos []*compositeUTXO, enough func(uint64, uint64) bool) (
...
-// - overhead from 1 transaction
-// - 1 extra signed p2wpkh-spending input. The split tx has the fundingCoins as
-//   inputs now, but we'll add the input that spends the sized coin that will go
-//   into the first swap if the split tx does not add excess baggage
-// - 2 additional p2wpkh outputs for the split tx sized output and change
+//   - overhead from 1 transaction
+//   - 1 extra signed p2wpkh-spending input. The split tx has the fundingCoins as
+//     inputs now, but we'll add the input that spends the sized coin that will go
+//     into the first swap if the split tx does not add excess baggage
+//   - 2 additional p2wpkh outputs for the split tx sized output and change
```
chappjc, line 1816: Hmm, I must have done go fmt with Go 1.19 rc1 which  has new formatting.  https://tip.golang.org/doc/go1.19#go-doc

Since older gofmt versions don't try to change it back, I think I'll leave it.

buck54321 (approved):

PR #1702, opened 2022-07-08
client/eth: Use wallet seed as entropy for BIP-39 mnemonic

martonp:
Previously, the wallet seed from core was used as the master seed of the an HD wallet from which a private key was derived. Instead of this, the seed from core is used as an entropy for a BIP-39 mnemonic phrase, then the seed derived from the mnemonic phrase is used to derive the private key used in the wallet. The mnemonic phrase is currently not used for anything, but this will keep our options open for more features in the future.

This will come on top of #1648, d0f77dc22900b26c24226c2f347015a143581d94 is the diff for this PR.

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -37,6 +37,7 @@ import (
 	ethmath "github.com/ethereum/go-ethereum/common/math"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/params"
+	"github.com/tyler-smith/go-bip39"
 )
 
 func registerToken(tokenID uint32, desc string) {
@@ -328,6 +329,20 @@ func (w *TokenWallet) Info() *asset.WalletInfo {
 	}
 }
 
+// genWalletSeed uses the wallet seed passed from core as the entropy for
+// generating a BIP-39 mnemonic. Then it returns the wallet seed generated
+// from the mnemonic which can be used to derive a private key.
+func genWalletSeed(entropy []byte) ([]byte, error) {
+	if len(entropy) != 32 {
+		return nil, fmt.Errorf("wallet entropy must be 32 bytes long")
+	}
+	mnemonic, err := bip39.NewMnemonic(entropy)
+	if err != nil {
+		return nil, fmt.Errorf("error deriving mnemonic: %w", err)
+	}
+	return bip39.NewSeed(mnemonic, ""), nil
+}
+
 // CreateWallet creates a new internal ETH wallet and stores the private key
 // derived from the wallet seed.
 func CreateWallet(createWalletParams *asset.CreateWalletParams) error {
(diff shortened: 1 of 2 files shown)


chappjc:
Looks to be as discussed https://github.com/decred/dcrdex/pull/1648#discussion_r898428136

https://github.com/decred/dcrdex/pull/1648 tests well and is ready to merge for me, as soon as @buck54321 circles back on his review there.

JoeGruffins (commented):
Looks fine.. Who is Tyler Smith?

client/asset/eth/eth.go
```diff
@@ -328,6 +329,21 @@ func (w *TokenWallet) Info() *asset.WalletInfo {
...
 
+// genWalletSeed uses the wallet seed passed from core as the entropy for
+// generating a BIP-39 mnemonic. Then it returns the wallet seed generated
+// from the mnemonic which can be used to derive a private key.
+func genWalletSeed(entropy []byte) ([]byte, error) {
+	if len(entropy) != 32 {
+		return nil, fmt.Errorf("wallet seed must be 32 bytes long")
+	}
+	mnemonic, err := bip39.NewMnemonic(entropy)
+	defer func() { mnemonic = "" }()
```
JoeGruffins, line 340: This is probably ineffectual. Strings can't be forcefully cleared from memory apparently. https://stackoverflow.com/questions/39968084/is-it-possible-to-zero-a-golang-strings-memory-safely

client/asset/eth/eth.go
```diff
@@ -328,6 +329,21 @@ func (w *TokenWallet) Info() *asset.WalletInfo {
 	}
 }
 
+// genWalletSeed uses the wallet seed passed from core as the entropy for
+// generating a BIP-39 mnemonic. Then it returns the wallet seed generated
+// from the mnemonic which can be used to derive a private key.
+func genWalletSeed(entropy []byte) ([]byte, error) {
+	if len(entropy) != 32 {
+		return nil, fmt.Errorf("wallet seed must be 32 bytes long")
```
LasTshaMAN, line 337: maybe `wallet seed` -> `wallet entropy`

client/asset/eth/eth.go
```diff
@@ -328,6 +329,21 @@ func (w *TokenWallet) Info() *asset.WalletInfo {
 	}
 }
 
+// genWalletSeed uses the wallet seed passed from core as the entropy for
```
LasTshaMAN, line 332: maybe elaborate a bit: `uses the wallet seed passed from core` -> `uses the wallet seed core derives from app seed`

LasTshaMAN (approved):
just some nits

martonp:
> I looked through tyler's repo for a bit and looks fine to me. Also, out of curiosity I tried using the mnemonic we get here in metamask and ended up with the same account.

Yeah, this seems to be the standard go bip-39 library that everyone uses. It hasn't been updated for 2 years.

client/asset/eth/eth.go
```diff
@@ -2041,7 +2063,13 @@ func (w *TokenWallet) EstimateRegistrationTxFee(feeRate uint64) uint64 {
 // RestorationInfo returns information about how to restore the wallet in
 // various external wallets.
 func (w *assetWallet) RestorationInfo(seed []byte) ([]*asset.WalletRestoration, error) {
```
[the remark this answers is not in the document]
martonp: These are the bytes used by core to seed the wallet though. It's just that the ETH wallet happens to use it as the entropy for a BIP-39 mnemonic.

chappjc (approved):
Works as expected.  This will leave the door open to exporting both the seed words and the account private key, if needed.  I don't think we need to change any UI aspects however.

PR #1709, opened 2022-07-13
ui/markets: banish 1e8 for good

buck54321:
I don't know how these have evaded us for so long.

### DIFF
--- a/client/webserver/site/src/js/app.ts
+++ b/client/webserver/site/src/js/app.ts
@@ -811,15 +811,7 @@ export default class Application {
     const supportedAsset = this.assets[assetID]
     if (supportedAsset) return supportedAsset.info.unitinfo
     if (!xc) {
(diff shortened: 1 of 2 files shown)


client/webserver/site/src/js/app.ts
```diff
@@ -811,15 +811,7 @@ export default class Application {
...
-      console.error(`no supported asset info for id = ${assetID}, and no exchange info provided`)
-      return {
-        atomicUnit: '',
-        conventional: {
-          unit: '',
-          conversionFactor: 1e8
-        },
-        denominations: []
-      }
+      throw Error(`no supported asset info for id = ${assetID}, and no exchange info provided`)
```
buck54321, line 814: Now that we actually have `conversionFactors`s that aren't `1e8`, I don't think we should return a default value here anymore.

JoeGruffins (approved):

chappjc (approved):

PR #1711, opened 2022-07-14
core: Fix OrderReader percentage helper for market buys

### LINKED ISSUE #775: app: order details page shows incorrect filled amount for market buy
testnet example:

image

800% is wrong, but it's because it was a 2.5 BTC market buy.

The maker's rate that was matched is 0.1 btc/dcr, so 2.5 btc covered just one 20 DCR lot.

martonp:
The percent filled/settled calculation for market buys was incorrectly using the "to" asset as the numerator. It is now updated to use the "from asset".

Closes #775

### DIFF
--- a/client/core/helpers.go
+++ b/client/core/helpers.go
@@ -179,7 +179,7 @@ func (ord *OrderReader) SideString() string {
 
 func (ord *OrderReader) percent(filter func(match *Match) bool) string {
 	var sum uint64
-	if ord.Sell {
+	if ord.Sell || ord.IsMarketBuy() {
 		sum = ord.sumFrom(filter)
 	} else {
 		sum = ord.sumTo(filter)


client/core/helpers.go
```diff
@@ -179,7 +179,7 @@ func (ord *OrderReader) SideString() string {
 
 func (ord *OrderReader) percent(filter func(match *Match) bool) string {
 	var sum uint64
-	if ord.Sell {
+	if ord.Sell || ord.IsMarketBuy() {
```
chappjc, line 182: Are `FilledFrom` and `FilledTo` OK?  They both call `sumTo`/`sumFrom`.  Should this `IsMarketBuy` condition be in the `sum` functions instead?

chappjc (approved):
Confirmed that Filled is now a percentage of the quote asset:

image

Looks like `FilledFrom` and `FilledTo` are fine.

JoeGruffins (approved):

Issue #1696, opened 2022-07-06
eth: MaxSell panic with no AssetConfig set


JoeGruffins:
On master 1961b68a38a4c886f9708166f7b64eff438a2f30

Get a panic when trading on simnet. Have been trading on dcr_eth so probably related. I see something about redeem gas.:
<details closed>
<summary>panic</summary>

```
2022/07/06 11:39:23 http: panic serving 127.0.0.1:37618: runtime error: slice bounds out of range [-1:]
goroutine 8408 [running]:
net/http.(*conn).serve.func1()
        /usr/lib/go/src/net/http/server.go:1825 +0xbf
panic({0x1415620, 0xc001c9e120})
        /usr/lib/go/src/runtime/panic.go:844 +0x258
github.com/go-chi/chi/v5/middleware.prettyStack.decorateFuncCallLine({}, {0xc000ab8c7b, 0x1d}, 0xc1?, 0x8)
        /home/joe/go/pkg/mod/github.com/go-chi/chi/v5@v5.0.1/middleware/recoverer.go:130 +0x545
github.com/go-chi/chi/v5/middleware.prettyStack.decorateLine({}, {0xc000ab8c7b?, 0xc000601000?}, 0x90?, 0x1?)
        /home/joe/go/pkg/mod/github.com/go-chi/chi/v5@v5.0.1/middleware/recoverer.go:106 +0x15d
github.com/go-chi/chi/v5/middleware.prettyStack.parse({}, {0xc001dd0000, 0x141a, 0xc004bfce40?}, {0x12f6d00, 0x21956d0})
        /home/joe/go/pkg/mod/github.com/go-chi/chi/v5@v5.0.1/middleware/recoverer.go:89 +0x4e5
github.com/go-chi/chi/v5/middleware.PrintPrettyStack({0x12f6d00, 0x21956d0})
        /home/joe/go/pkg/mod/github.com/go-chi/chi/v5@v5.0.1/middleware/recoverer.go:46 +0x45
github.com/go-chi/chi/v5/middleware.(*defaultLogEntry).Panic(0x900000000?, {0x12f6d00?, 0x21956d0?}, {0x19bca8a79bbc?, 0x203000?, 0xc001d83800?})
        /home/joe/go/pkg/mod/github.com/go-chi/chi/v5@v5.0.1/middleware/logger.go:165 +0x29
github.com/go-chi/chi/v5/middleware.Recoverer.func1.1()
[...]

PR #1713, opened 2022-07-16
client/core: Popuplate AssetConfig on PreRedeem call

martonp:
The asset config was not populated when calling `PreRedeem` in `MaxSell`. This was causing a panic.


chappjc:
Resolves, https://github.com/decred/dcrdex/issues/1696, although @buck54321 has this in PR #1622

JoeGruffins (approved):

chappjc (approved):
OK, but I'm going to focus on https://github.com/decred/dcrdex/pull/1622 now so we can get other fixes in with that.

Issue #1717, opened 2022-07-18
core: PreRedeem (via maxsell,maxbuy,preorder) hammers feeRate


chappjc:
`PreRedeem` still internally requests it's own fee rate (and now the external API like dcrdata) however its callers (`MaxSell`, `MaxBuy`, and `PreOrder`) _also_ request a fee rate from the "to"/redeeming wallet which they feed to `PreRedeem`.  As a result, these methods (and http api endpoints) always request fee rates twice.  This is not new, but it's now intolerable because it hits a remote HTTP source with DCR SPV.

This issue is worsened because the frontend still has a glitch where it hits the `/api/maxbuy` endpoint in rapid succession.  Typing "0.00115" on the Buy price for the DCR-BTC market:

```
2022-07-18 16:32:21.837 [DBG] CORE[dcr]: Retrieving fee rate from external API
2022-07-18 16:32:21.878 [DBG] CORE[dcr]: Retrieving fee rate from external API
2022-07-18 16:32:22.037 [DBG] CORE[dcr]: Retrieving fee rate from external API
2022-07-18 16:32:22.155 [TRC] CORE[dcr]: Obtained local estimate for 1-conf fee rate, 11
2022/07/18 16:32:22 "POST http://127.0.0.3:5758/api/maxbuy HTTP/1.1" from 127.0.0.1:55876 - 200 200B in 332.722562ms
2022-07-18 16:32:22.545 [DBG] CORE[dcr]: Retrieving fee rate from external API
2022-07-18 16:32:22.598 [TRC] CORE[dcr]: Obtained local estimate for 1-conf fee rate, 11
2022/07/18 16:32:22 "POST http://127.0.0.3:5758/api/maxbuy HTTP/1.1" from 127.0.0.1:59526 - 200 200B in 729.31227ms
```

Four times just for the frontend's maxbuy requests, then when you hit "Place order to buy DCR" it hits the `/api/preorder` endpoint, requesting the fee rate two more times.

Related issue, with initial resolution in comments: https://github.com/decred/dcrdex/issues/1718#issuecomment-1189380283

chappjc:
Note to self, with a fix, we also need to make the following fixes/tweaks to DCR's external fee rate handling:

```diff
diff --git a/client/asset/dcr/rpcwallet.go b/client/asset/dcr/rpcwallet.go
index b59bc63b..34a14453 100644
--- a/client/asset/dcr/rpcwallet.go
[...]

Issue #1718, opened 2022-07-18
with an active order, fee rates are requested each epoch


chappjc:
```
2022-07-18 16:46:22.678 [DBG] CORE[dcr]: Retrieving fee rate from external API
2022-07-18 16:46:30.177 [DBG] CORE[dcr]: Retrieving fee rate from external API
2022-07-18 16:46:37.677 [DBG] CORE[dcr]: Retrieving fee rate from external API
2022-07-18 16:46:45.177 [DBG] CORE[dcr]: Retrieving fee rate from external API
2022-07-18 16:46:52.677 [DBG] CORE[dcr]: Retrieving fee rate from external API
2022-07-18 16:47:00.177 [DBG] CORE[dcr]: Retrieving fee rate from external API
2022-07-18 16:47:07.677 [DBG] CORE[dcr]: Retrieving fee rate from external API
```

The `feeRate` method in the above come from:

```
decred.org/dcrdex/client/asset/dcr.(*ExchangeWallet).FeeRate(0xc0002d97a0)
	/home/jon/github/decred/dcrdex/client/asset/dcr/dcr.go:878 +0x27
decred.org/dcrdex/client/core.(*Core).cacheRedemptionFeeSuggestion(0xc000322000, 0xc000170000)
	/home/jon/github/decred/dcrdex/client/core/core.go:7013 +0x53
decred.org/dcrdex/client/core.(*Core).tick(0xc000322000, 0xc000170000)
	/home/jon/github/decred/dcrdex/client/core/trade.go:1348 +0x189
decred.org/dcrdex/client/core.(*Core).listen.func3()
	/home/jon/github/decred/dcrdex/client/core/core.go:6476 +0x88a
decred.org/dcrdex/client/core.(*Core).listen.func4()
	/home/jon/github/decred/dcrdex/client/core/core.go:6506 +0x234
created by decred.org/dcrdex/client/core.(*Core).listen
	/home/jon/github/decred/dcrdex/client/core/core.go:6491 +0x3d0
```

I don't recall why, but `cacheRedemptionFeeSuggestion` calls the `FeeRate` method if the wallet is a `FeeRater`.  If the wallet is a `FeeRater`, why does it need to cache it's own suggestion?
[...]

PR #1653, opened 2022-06-06
client: redo reserves

chappjc:

PR #1721, opened 2022-07-20
client/{core,asset}: fee rate caching and expiry

chappjc:
This resolves https://github.com/decred/dcrdex/issues/1717, https://github.com/decred/dcrdex/issues/1515, and https://github.com/decred/dcrdex/issues/1718

```
In client/core, this updates trackedTrade's redeemFeeSuggestion field
as a new type, feeStamped, with a time stamp paired with the fee rate.

cacheRedemptionFeeSuggestion is made into a trackedTrade method (always
should have been), and now checks the time stamp on the
redeemFeeSuggestion before retrieving an updated rate from any of the
sources: wallet (may include external oracle request), best book fee
suggestion, and on demand fee_rate server RPC. This expiry check
resolves a long standing stale fee rate issue, and it throttles the
frequency with which we request fee rates from any source.

In client/asset/dcr, fee rates from an external API / oracle are now
cached and time stamped. The feeRate method is in charge of checking
and saving these cached rates.

PreRedeem now uses any non-zero FeeSuggestion directly, only making a
feeRate request if the provided rate is zero.  Callers in Core,
including MaxSell, MaxBuy, and Preorder, already obtain a rate.

Redeem remains a method that internally gets a fee rate because some
wallets have a redeem confirm target blocks setting to respect. It is
now documented on dcr's Redeem method that FeeSuggestion is a fallback.

Also, the Refund method now treats the fee rate argument as the rate to
use rather than a fallback, which matches the methods updated in
e41ecdfe57045cd3bf6b22aa7a6054740f0dca05.
dcr's Refund now only requests a rate if the provided rate is zero,
where previously the feeSuggestion input was a fallback.

btc's Refund and PreRedeem methods are similarly updated to use the
provided fee rate directly rather than as a fallback.

eth's Refund and PreRedeem methods already used the fee rates directly.
```

### DIFF
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -83,6 +83,10 @@ const (
 	// testnetExternalApiUrl is the URL of the testnet external API in case of
 	// fallback.
 	testnetExternalApiUrl = "https://testnet.dcrdata.org/insight/api"
+
+	// freshFeeAge is the expiry age for cached fee rates of external origin,
+	// past which fetchFeeFromOracle should be used to refresh the rate.
+	freshFeeAge = time.Minute
 )
 
 var (
@@ -497,6 +501,11 @@ type redeemOptions struct {
 	FeeBump *float64 `ini:"redeemfeebump"`
 }
 
+type feeStamped struct {
+	rate  uint64
+	stamp time.Time
+}
+
 // ExchangeWallet is a wallet backend for Decred. The backend is how the DEX
 // client app communicates with the Decred blockchain and wallet. ExchangeWallet
 // satisfies the dex.Wallet interface.
@@ -505,6 +514,7 @@ type ExchangeWallet struct {
 	wallet         Wallet
 	chainParams    *chaincfg.Params
 	log            dex.Logger
+	network        dex.Network
 	primaryAcct    string
 	unmixedAccount string // mixing-enabled wallets only
 	// tradingAccount (mixing-enabled wallets only) stores utxos reserved for
@@ -519,7 +529,10 @@ type ExchangeWallet struct {
 	redeemConfTarget uint64
 	useSplitTx       bool
 	apiFeeFallback   bool
-	network          dex.Network
+
+	oracleFeesMtx sync.Mutex
+	oracleFees    map[uint64]feeStamped // conf target => fee rate
+	oracleFailing bool
 
 	tipMtx     sync.RWMutex
 	currentTip *block
@@ -665,6 +678,7 @@ func unconnectedWallet(cfg *asset.WalletConfig, dcrCfg *walletConfig, chainParam
 	return &ExchangeWallet{
 		log:                 logger,
 		chainParams:         chainParams,
+		network:             network,
 		primaryAcct:         dcrCfg.PrimaryAccount,
 		unmixedAccount:      dcrCfg.UnmixedAccount,
 		tradingAccount:      dcrCfg.TradingAccount,
@@ -678,7 +692,7 @@ func unconnectedWallet(cfg *asset.WalletConfig, dcrCfg *walletConfig, chainParam
 		redeemConfTarget:    redeemConfTarget,
 		useSplitTx:          dcrCfg.UseSplitTx,
 		apiFeeFallback:      dcrCfg.ApiFeeFallback,
-		network:             network,
+		oracleFees:          make(map[uint64]feeStamped),
 	}, nil
 }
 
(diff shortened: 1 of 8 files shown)


client/core/trade.go
```diff
@@ -2324,24 +2390,24 @@ func (c *Core) refundMatches(t *trackedTrade, matches []*matchTracker) (uint64,
...
+			feeRate = t.metaData.MaxFeeRate
 		}
-		if feeSuggestion == 0 {
-			feeSuggestion = c.feeSuggestionAny(refundAsset.ID)
+		if feeRate == 0 {
+			feeRate = c.feeSuggestionAny(refundAsset.ID) // includes wallet itself
 		}
 
-		refundCoin, err := refundWallet.Refund(swapCoinID, contractToRefund, feeSuggestion)
+		refundCoin, err := refundWallet.Refund(swapCoinID, contractToRefund, feeRate)
```
chappjc, line 2410: We're already getting a proper fee rate for Refund, so we can just solidify the new semantics and avoid a redundant internal fee rate request.

chappjc:
Briefly tested interactively, so ready for review, but I haven't tested all trade scenarios.

JoeGruffins (approved):
Running with loadbot:
For eth, will often see in the logs:
```
2022-07-20 13:08:40.533 [ERR] CORE:CMPD:SNIPER:0: attempting to unlock more than is reserved for order 8172267c7353ac0b503948a75d80e47dfbcae64ab652ebad6b5129d5d579e3dd. unlocking reserved amount instead: 45866666 reserved, unlocking 45866667
```
Which I think is new.

And loadbot seems to fail with a dcr error:
```
2022-07-20 13:52:15.177 [ERR] CORE:CMPD:STACKER:1: notify: |ERROR| (send) Send error - Error encountered while sending dcr: not enough funds to cover minimum fee rate. 0.0004658 > 0.0004651
```
Which may or may not be new.

client/asset/dcr/dcr.go
```diff
@@ -873,75 +886,87 @@ func (dcr *ExchangeWallet) Balance() (*asset.Balance, error) {
...
+
+	dcr.oracleFeesMtx.Lock()
+	defer dcr.oracleFeesMtx.Unlock()
+	oracleFee := dcr.oracleFees[confTarget]
+	if now.Sub(oracleFee.stamp) < freshFeeAge {
+		return oracleFee.rate, nil
+	}
+
+	dcr.log.Debugf("Retrieving fee rate from external fee oracle for %d target blocks", confTarget)
+	dcrPerKB, err := fetchFeeFromOracle(dcr.ctx, dcr.network, confTarget)
```
buck54321, line 926: This sequence is a little scary. Consider the case when the oracle is somehow non-responsive. The first rate caller is stuck here for 4 seconds. Any other callers within that 4 seconds will line up. When the timeout occurs, subsequent requesters will still wait in line to issue requests themselves, each taking 4 seconds in turn.

I think we need an error mode, so that subsequent requests are blocked for some time after an error or two. On the `FeeRate` route, the client may still be able to get a fee suggestion from the DEX server.

client/asset/dcr/dcr.go
```diff
@@ -873,75 +886,87 @@ func (dcr *ExchangeWallet) Balance() (*asset.Balance, error) {
...
-		dcr.log.Errorf("Failed to convert fee to atoms: %v", err)
 		return 0, err
 	}
-	if atomsPerKB > dcr.feeRateLimit {
-		dcr.log.Errorf("Fee rate greater than fee rate limit: %v", atomsPerKB)
-		return 0, err
+	if atomsPerByte > dcr.feeRateLimit {
+		return 0, fmt.Errorf("fee rate from external API greater than fee rate limit: %v > %v",
+			atomsPerByte, dcr.feeRateLimit)
 	}
```
buck54321, line 941: Was bug?

Much better variable and function names, btw.

client/asset/dcr/dcr.go
```diff
@@ -2788,13 +2817,17 @@ func (dcr *ExchangeWallet) fatalFindRedemptionsError(err error, contractOutpoint
...
+// otherwise it will be used directly.
 // NOTE: The contract cannot be retrieved from the unspent coin info as the
 // wallet does not store it, even though it was known when the init transaction
 // was created. The client should store this information for persistence across
 // sessions.
-func (dcr *ExchangeWallet) Refund(coinID, contract dex.Bytes, feeSuggestion uint64) (dex.Bytes, error) {
-	msgTx, err := dcr.refundTx(coinID, contract, 0, nil, feeSuggestion)
+func (dcr *ExchangeWallet) Refund(coinID, contract dex.Bytes, feeRate uint64) (dex.Bytes, error) {
+	if feeRate == 0 {
+		feeRate = dcr.targetFeeRateWithFallback(2, 0)
```
buck54321, line 2828: I had to really work through this. The way the `feeRate` is generated in **core**, if `feeRate == 0` here, that means that `(*ExchangeWallet).FeeRate` just failed (as well as any attempt to get a rate from the server or the order book). So this is likely to fail again, and we probably end up with the `fallbackFeeRate`. Is that your read too?

chappjc: Yeah, with core's current use pattern, that's likely to happen.  I thought about just `dcr.feeRateWithFallback(feeRate)` here, but figured what the heck, might as well try an internal rate assuming we don't know what caller is doing.

client/asset/dcr/dcr.go
chappjc: > When the timeout occurs, subsequent requesters will still wait in line to issue requests themselves, each taking 4 seconds in turn

Good call.  That's an ugly outcome.
I think you're right that there needs to be a "failing" state that expires after a bit.

client/asset/dcr/dcr.go
buck54321: Your call. Even just a comment.

client/asset/dcr/dcr.go
```diff
@@ -3205,10 +3251,11 @@ func (dcr *ExchangeWallet) withdraw(addr stdaddr.Address, val, feeRate uint64) (
 
 // sendToAddress sends an exact amount to an address. Transaction fees will be
 // in addition to the sent amount, and the output will be the zeroth output.
-// TODO: Just use the sendtoaddress rpc since dcrwallet respects locked utxos.
+// TODO: Just use the sendtoaddress rpc since dcrwallet respects locked utxos!
 func (dcr *ExchangeWallet) sendToAddress(addr stdaddr.Address, amt, feeRate uint64) (*wire.MsgTx, uint64, error) {
+	baseSize := uint32(dexdcr.MsgTxOverhead + dexdcr.P2PKHOutputSize*2) // may be extra if change gets omitted (see signTxAndAddChange)
 	enough := func(sum uint64, size uint32, unspent *compositeUTXO) bool {
-		txFee := uint64(size+unspent.input.Size()) * feeRate
+		txFee := uint64(baseSize+size+unspent.input.Size()) * feeRate
```
buck54321, line 3258: It's curious that this has evaded tests and has not popped up in the wild. Or has it? Regardless, it's probably worth asking if our approach to testing is insufficient.

chappjc: But regarding testing, I don't think "insufficient" is really the problem so much as very cumbersome testing procedures on account of our product code design, specifically the need to coax certain outcomes by stubbing things out with fake data.  It's hard to do, so we don't get thorough unit tests.

In the case of this `enough` closure inside of the `sendToAddress`, it's a closure so we have to test it _indirectly_, which means actually making the following other methods work: `fund`, `spendableUTXOs`, `tryFund`, `lockFundingCoins`, `sendCoins`, `addInputCoins`, `sendWithReturn`, and probably more.  It's understandable that we didn't have narrow unit tests for _just_ these few lines of code in `sendToAddress`.

Perhaps if we defined `enough` at the package level as `enoughForRegularSend` we'd write short and direct unit test without the fuss of stubbing out `dcr.Wallet` and hacking in responses to get the desired effect.

That is all to say I think the general solution is making the bits and pieces more testable on their own, and doing less high level testing of methods that call dozens of other methods.  Integration tests are important, but our unit tests are written like integration tests and fairly difficult to maintain.

JoeGruffins:
> Hard to see how the ETH "attempting to unlock more than is reserved" error log can be from this diff

I assumed it had something to do with `IntDivUp` being used somewhere and not somewhere else. I don't think I've ever seen this one gwei off every time before.

chappjc:
> > Hard to see how the ETH "attempting to unlock more than is reserved" error log can be from this diff
> 
> I assumed it had something to do with `IntDivUp` being used somewhere and not somewhere else. I don't think I've ever seen this one gwei off every time before.

So it is every time? And always off by 1?

JoeGruffins:
Oh yeah, only used in btc and dcr, huh...

JoeGruffins:
> So it is every time? And always off by 1? And not on master?

No, yes, not seen yet but maybe just not lucky yet. Will run for longer on master, but here are some full logs on this diff. One time in an hour.

loadbot logs:
[dcreth.tar.gz](https://github.com/decred/dcrdex/files/9156152/dcreth.tar.gz)

JoeGruffins:
Was able to hit on master.

Can reopen reopen #1544 or make a new issue. This seems to be a rounding issue. Maybe fixed by #1653

buck54321 (approved):

PR #1722, opened 2022-07-20
eth/harness: Uncomment init overflow test.

### LINKED ISSUE #1720: eth/harness: Add test for contract overflow in init.
I was looking at this: https://gist.github.com/chappjc/350aafb9031f7a66986967bf8ab67d38

And there's a comment at the top:
```
 * Arithmetic operations in Solidity wrap on overflow. This can easily result
 * in bugs, because programmers usually assume that an overflow raises an
 * error, which is the standard behavior in high level programming languages.
 * `SafeMath` restores this intuition by reverting the transaction when an
 * operation overflows.
 ```

So I went to check our contract to see if overflow could affect us somehow. It looks like one could init several swaps together in such a way that they overflow and the sum is equal to the input value, but the values we store in the map could be anything. Later, this person could refund any amount at all.

But, this should not be the case since v0.8.0 https://docs.soliditylang.org/en/v0.8.13/080-breaking-changes.html

The transaction should be reverted if there is overflow for us currently.

We should at least make a harness test to ensure we are safe. Maybe even add checks for overflow to the contract(s)? Or at least some comments that version must be over v0.8.0 Adding more checks would up the transaction costs.

The value here is a `uint256`.

https://github.com/decred/dcrdex/blob/78dc76557a24fe798732c03f7ad9cceda8a0f6c6/dex/networks/eth/contracts/ETHSwapV0.sol#L92-L117

https://github.com/decred/dcrdex/blob/78dc76557a24fe798732c03f7ad9cceda8a0f6c6/dex/networks/eth/contracts/ETHSwapV0.sol#L171-L187

### PR DESCRIPTION
**Pull request by JoeGruffins**
closes #1720

So, there was already a test! I guess @buck54321 or @martonp noticed the possible vulnerability before me. The commented out test just never registered for me....

Anyway, I tried this with a contract build with 0.7.5 and can confirm the init succeeds and test fails.

### DIFF
--- a/client/asset/eth/nodeclient_harness_test.go
+++ b/client/asset/eth/nodeclient_harness_test.go
@@ -44,6 +44,7 @@ import (
 	"decred.org/dcrdex/dex/testing/eth/reentryattack"
 	"github.com/davecgh/go-spew/spew"
 	"github.com/ethereum/go-ethereum/accounts"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by buck54321** [comment]
I think @martonp wrote the test, and I commented it out, because I though there was no way to actually overflow with our architecture. I'm not fully caught up on this convo, but from what I gather, I was wrong? Anyway, more tests and better security is always welcome.

**Comment by chappjc** [comment]
Yah, good to check these things anyway.  But why is the contract bin changing?

File: dex/networks/eth/contracts/ETHSwapV0.sol [hunk 1]
```diff
@@ -1,5 +1,9 @@
 // SPDX-License-Identifier: BlueOak-1.0.0
+
 // pragma should be as specific as possible to allow easier validation.
+//
+// WARNING: pragma MUST be above v0.8.0 so that addition overflow will cause
+// transactions to be reverted.
```
chappjc, line 6: If changing the comments really changes the bin, let's just skip this.  I don't think we'll ever consider downgrading to before 0.8.0.

JoeGruffins:
> I'm not fully caught up on this convo, but from what I gather, I was wrong?

No, just wanted to make sure. The contract is fine afaik.

dex/networks/eth/contracts/ETHSwapV0.sol
JoeGruffins: Ok.

chappjc: So comments really change the output?  That's kinda nuts

JoeGruffins: I guess there's a hash of the entire thing and a checksum somewhere. I'm guessing it's only a checksum changing. Can check how much changes because that's good info to have.

JoeGruffins: At the end of the hex, this part changed:
9c8d8d137b7639af24fe8215dd5976b16323176cb4b9be076c90eeb32eaf7446 -> ef281bf04bfe702e408f8eba5336858e62259885e954f238248ad567172b2764

So, a 32 byte hash, probably of the entire contract including comments.

chappjc: Probably.  Geeze it should be a hash of a language preprocessor's output (or optionally omitted), not the file contents.  Oh well

JoeGruffins: Removed warning https://github.com/decred/dcrdex/compare/c158bd437e5a1e7d821444f449eac232641e5ae3..15e4e83718b35379398615078cba21e48729e006

martonp (approved):
I guess the point of test is to make sure no one reduces the solidity version below 8.0 and to make sure the solidity overflow check is working properly.

JoeGruffins:
> I guess the point of test

There could also be a change reverting the behavior in the future. However unlikely that may be.

chappjc (approved):

PR #1723, opened 2022-07-21
Update README.md to use consistent "JavaScript"

andsiu:
Changes mentions of "javascript" to "JavaScript" in README.md

No reference to issue as the problem is too small to justify creating an issue.


chappjc (approved):
Thanks.

But just a note to everyone that I'll be doing a README.md update soon for the 0.5 release.  There are a myriad of changes needed: node 16/18, mention native BTC SPV and not bitcoind straight away, DCR mixing support and account settings, etc.

PR #1732, opened 2022-07-25
address client wallet hanging issues

chappjc:
We're clearly fighting some limitations of our `asset.Wallet` interface (no ctx args) and Go (no `context.Merge`), but there are some smaller things we can do.

The first commit adds some more context passing to avoid things getting hung and unable to cancel them.  Some sensible timeouts are also established.

The second commit is a bit of a hack to prevent shutdown from becoming impossible if an RPC or something hangs and causes deadlock on the `trackedTrade` mutex.

This PR is the basis for more extensive refactoring in https://github.com/decred/dcrdex/pull/1739

### DIFF
--- a/client/core/trade.go
+++ b/client/core/trade.go
@@ -919,7 +919,7 @@ func (t *trackedTrade) counterPartyConfirms(ctx context.Context, match *matchTra
 	wallet := t.wallets.toWallet
 	coin := match.counterSwap.Coin
 
-	_, lockTime, err := wallet.LocktimeExpired(match.MetaData.Proof.CounterContract)
+	_, lockTime, err := wallet.LocktimeExpired(ctx, match.MetaData.Proof.CounterContract)
 	if err != nil {
 		return fail(fmt.Errorf("error checking if locktime has expired on taker's contract on order %s, "+
 			"match %s: %w", t.ID(), match, err))
@@ -1191,7 +1191,7 @@ func (t *trackedTrade) isSwappable(ctx context.Context, match *matchTracker) boo
 // broadcast.
 //
 // This method accesses match fields and MUST be called with the trackedTrade
-// mutex lock held for reads.
+// mutex lock held for writes.
 func (t *trackedTrade) isRedeemable(ctx context.Context, match *matchTracker) bool {
 	// Quick status check before we bother with the wallet.
 	switch match.Status {
@@ -1290,7 +1290,7 @@ func (t *trackedTrade) isRedeemable(ctx context.Context, match *matchTracker) bo
 //
 // This method modifies match fields and MUST be called with the trackedTrade
 // mutex lock held for writes.
-func (t *trackedTrade) isRefundable(match *matchTracker) bool {
+func (t *trackedTrade) isRefundable(ctx context.Context, match *matchTracker) bool {
 	if match.refundErr != nil || len(match.MetaData.Proof.RefundCoin) != 0 {
 		t.dc.log.Tracef("Match %s not refundable: refundErr = %v, RefundCoin = %v",
 			match, match.refundErr, match.MetaData.Proof.RefundCoin)
@@ -1313,7 +1313,7 @@ func (t *trackedTrade) isRefundable(match *matchTracker) bool {
 	}
 
 	// Issue a refund if our swap's locktime has expired.
-	swapLocktimeExpired, contractExpiry, err := wallet.LocktimeExpired(match.MetaData.Proof.ContractData)
+	swapLocktimeExpired, contractExpiry, err := wallet.LocktimeExpired(ctx, match.MetaData.Proof.ContractData)
 	if err != nil {
 		if !errors.Is(err, asset.ErrSwapNotInitiated) {
 			// No need to log an error as this is expected for newly
@@ -1405,8 +1405,7 @@ func (c *Core) tick(t *trackedTrade) (assetMap, error) {
 	t.mtx.Lock()
 	defer t.mtx.Unlock()
 	var swaps, redeems, refunds []*matchTracker
-	assets := make(assetMap)
-	errs := newErrorSet(t.dc.acct.host + " tick: ")
+	assets := make(assetMap) // callers expect non-nil map even on error
 
 	// Check all matches for and resend pending requests as necessary.
 	c.resendPendingRequests(t)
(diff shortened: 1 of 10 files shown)


client/core/trade.go
```diff
@@ -1405,60 +1405,95 @@ func (c *Core) tick(t *trackedTrade) (assetMap, error) {
...
+			return nil // either refunded or revoked requiring no action on this side of the match
 		}
-		switch {
-		case t.isSwappable(c.ctx, match):
+
+		// The trackedTrade mutex is locked, so we must not hang forever. Give
+		// this a generous timeout because it may be necessary to retrieve full
+		// blocks, and catch timeout/shutdown after each check.
+		ctx, cancel := context.WithTimeout(c.ctx, 30*time.Second)
+		defer cancel()
```
chappjc, line 1434: This part makes me a touch nervous.  We could just use `c.ctx` and still get the shutdown fix, but really there should be a timeout I think.  But what?

chappjc:
Ultimately we probably should move toward adding `ctx` args on the interfaces like [client/asset.Wallet](https://github.com/decred/dcrdex/blob/fb05984dc0559f91f0c9d09c3fa82460e5615fae/client/asset/interface.go#L246) and [client/asset/btc.Wallet](https://github.com/decred/dcrdex/blob/fb05984dc0559f91f0c9d09c3fa82460e5615fae/client/asset/btc/wallet.go#L20).  Things would be a whole lot easier if there were a [context.Merge](https://github.com/golang/go/issues/36503), but that doesn't exists so I guess we should give up on trying to use an asset's own subsystem context and just accept the context from the caller i.e. `Core`.

Another challenge I can't really solve easily right now is how we do a **TON** with the `trackedTrade` mutex locked in `(*Core).tick`.  Any wallet request hang is deadly because of that, not just because of logout, but many other handlers. That'll take a bit more careful refactoring, which we can do after branching `release-v0.5`.
I have a separate branch feeling out solutions there.

client/core/trade.go
```diff
@@ -1405,60 +1405,95 @@ func (c *Core) tick(t *trackedTrade) (assetMap, error) {
...
 			c.log.Debugf("Refundable match %s for order %v (%v)", match, t.ID(), side)
 			refunds = append(refunds, match)
+			return nil
+		}
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
 
-		case t.shouldBeginFindRedemption(c.ctx, match):
+		if t.shouldBeginFindRedemption(ctx, match) {
```
chappjc, line 1471: I've done a lot more refactoring to `tick` in https://github.com/decred/dcrdex/pull/1739

JoeGruffins (approved):

client/core/trade.go
```diff
@@ -1405,60 +1405,95 @@ func (c *Core) tick(t *trackedTrade) (assetMap, error) {
...
+		if t.isSwappable(ctx, match) {
 			c.log.Debugf("Swappable match %s for order %v (%v)", match, t.ID(), side)
 			swaps = append(swaps, match)
 			sent += match.Quantity
 			quoteSent += calc.BaseToQuote(match.Rate, match.Quantity)
+			return nil
+		}
+		if ctx.Err() != nil { // may be here because of timeout or shutdown
+			return ctx.Err()
+		}
```
buck54321, line 1445: These checks seem unnecessary. The `ctx` will only be **used** in one of these checks, and match metadata checks will preclude rpc calls on the others.

chappjc: > The `ctx` will only be **used** in one of these checks, and match metadata checks will preclude rpc calls on the others

Can you expand on this?  The ctx is passed to each check and a timeout or shutdown can happen at any stage of these checks regardless of which checks actually make use of the ctx.

But it's not obvious or guaranteed that only one of these checks will actually use the ctx.

buck54321: | Method  | Context User | Pre-filters 
| ------------- | ------------- | ------------- |
| `isSwappable`  | `counterPartyConfirms`  | no swapErr or revocation, `Taker` in `MakerSwapCast`  |
| `isSwappable`  | `swapConfirmations`  | no swapErr or revocation, `Maker` in `MakerSwapCast`  |
| `isRedeemable`  | `counterPartyConfirms`  | no swapErr or revocation, `Maker` in `TakerSwapCast`  |
| `shouldBeginFindRedemption`  | `swapConfirmations`  | no search running, `Maker` in `MakerSwapCast`  |

So you're right. It **is** possible to get past the pre-filters of `isSwappable` and `shouldBeginFindRedemption` both on the same `tick`.

chappjc: Wow, thanks for the nice summation table!

But what I'm thinking is that even if you didn't get to any wallet requests for more than one of those *is* checks, I think we'd still want the `ctx.Err()` check after each one, because we don't know in which one(s) there might be a timeout.

buck54321: Haha, that's pretty much **my** argument [here](https://github.com/decred/dcrdex/pull/1751#discussion_r935751867). As the caller, `fetchFiatExchangeRates` doesn't know that `refreshFiatRates` has a timeout, so should use the safer pattern of creating a new timer every tick.

I don't know. You can leave it.

client/core/trade.go
```diff
@@ -1405,60 +1405,95 @@ func (c *Core) tick(t *trackedTrade) (assetMap, error) {
...
 
-		case t.shouldBeginFindRedemption(c.ctx, match):
+		if t.shouldBeginFindRedemption(ctx, match) {
 			c.log.Debugf("Ready to find counter-party redemption for match %s, order %v (%v)", match, t.ID(), side)
-			t.findMakersRedemption(match)
+			t.findMakersRedemption(c.ctx, match) // async search, not on the timeout
+			return nil
+		}
+
+		return ctx.Err()
```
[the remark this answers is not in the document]
chappjc: This comes down to what were arguing about in https://github.com/decred/dcrdex/pull/1732#discussion_r935502352

The whole reason this looks and feels weird is because these `is` functions aren't written to return and error.  Normally we'd have a pattern like this where the error from the function may be one of `context.DeadlineExceeded` or `context.Canceled`:

```go
err := doSomething(ctx)
if err != nil {
	if errors.Is(err, context.DeadlineExceeded) {
		 fmt.Println("timeout")
	} else if errors.Is(err, context.Canceled) {
		fmt.Println("shutdown")
	}
	return err
}
```

I didn't want to refactor those `is` helpers so much, so I'm just checking if the reason for the `false` return is because of either context timeout/shutdown **rather than** a legitimate `false` from successful checks.

In the previous revision with the `switch`-`case` it was very possible that one of the cases was `false` for one of these bad reasons (real error) but the `switch` let it go to check the next `case` because that's how a `switch` works.  These changes are just a minor improvement to prevent that erroneous flow and instead bail out to try again.

client/core/trade.go
chappjc: Final note on this issue in the following related comment: https://github.com/decred/dcrdex/pull/1732/files#r937073616

A normal pattern would have a function return an error, and if that function took a context that error could be a context error,  however these just have a bool.  A false can mean either it confidently determined that result, or there was simply an error (timeout, shutdown, or other exception).  Instead of refactoring these functions, we at least have the capability to see if the `false` was because of a timeout or shutdown rather than it actually being false **and moving on to the next check**.  Maybe we can refactor this even more in the future so it isn't quite so awkward, but this is the reason.

client/core/trade.go
chappjc: Good point.  I was actually basing this on some logged results from `checkTrades`, but that is reporting tick time for all trades:

https://github.com/decred/dcrdex/blob/3a476f110fe1e295b20a400edb34f6e08fe1fc74/client/core/core.go#L6461-L6466

I'll add some trace logs like that inside tick

chappjc: Added in https://github.com/decred/dcrdex/pull/1732/commits/eaf22a07aaadbee16b7cb03bb5f765faef6292f6

client/core/trade.go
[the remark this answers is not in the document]
buck54321: It looks like that would be a change in behavior and that we couldn't return any of the errors that we're currently logging in these methods. The reason those errors are only logged is because an error on one match can't stop others from proceeding, which it appears would no longer be the case in this new scheme. That's why the single boolean return makes sense here. There was no error that should have change the behavior of the caller, so there was no need to return one.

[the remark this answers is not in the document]
chappjc: > The reason those errors are only logged is because an error on one match can't stop others from proceeding, which it appears would no longer be the case in this new scheme.

Well that's another reason I didn't make that change.  But I didn't say every internal error would need to return an error, just that the pattern of checking the context in the caller instead of an error exists because of this current design.  Nothing needs to change IMO.

PR #1735, opened 2022-07-26
client: upgrade btc min version required

vctt94:
This PR upgrades bitcoin min version required from v0.19 to v0.21. It also adds a min descriptor version, if using descriptor wallets.

This was discussed at matrix.

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -53,8 +53,10 @@ const (
 	// redeem transaction.
 	defaultRedeemConfTarget = 2
 
-	minNetworkVersion  = 190000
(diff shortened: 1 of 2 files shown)


chappjc:
Nice.  Looks like what we need for 0.21, which had descriptor wallets, but not the `listdescriptors` RPC.

vctt94:
just noted the error does not look good:

image

fixing it

JoeGruffins (approved):
Good except for maybe the comment as @chappjc said.

chappjc (approved):
The comment is still incorrect, but the code works as required.

PR #1736, opened 2022-07-26
client: small fixes

vctt94:
This PR is a couple of small fixes. 

 - the `x` to close the btc wallet page is a little inside the div, 
 - it might throw an error, when clicking on some order if not having a wallet. It will throw an error when trying to popula the max order: 
image


client/webserver/site/src/css/main.scss
```diff
@@ -427,10 +427,9 @@ hr.dashed {
   background-color: #7775;
 }
 
-div.form-closer,
-div.corner-bttn {
```
chappjc, line 431: `corner-bttn` is clearly a relic, but for the future, it's helpful to comment on the PR to state that it's unused.

It does seem to have been orphaned in https://github.com/decred/dcrdex/commit/ebfcff744ad90754aba651c0a5d0b998543771a3#diff-9b6521c407894cf405fa6e27af8fb8ae95c2e629d563180d39c05d0cc0563f17L66

client/webserver/site/src/css/main.scss
```diff
@@ -427,10 +427,9 @@ hr.dashed {
...
+div.form-closer {
   position: absolute;
-  right: -15px;
```
chappjc, line 433: Before

image

After

image

After, hovering mouse over the X

image

I can live with that.  But I wonder if we need to reserve space for the X _above_ the rest of the form elements, or have it over top anything else in the form with nice styling.  Right now it is firmly outside the form.

chappjc (approved):
LGTM.

chappjc:
Please also bump the js and css bundle cache busters.

JoeGruffins (approved):

buck54321:
I think there's more to this. There are a couple of places where we apply style to `form-closer`. We do this at the top level of **main.scss**, which is where you've applied changes. There's also style applied in **forms.scss**.

```SCSS
#forms {
  & > form {
    div.form-closer {
      right: 20px;
      top: 35px;
    }
  }
}
```

The problem is that since the second definition selects for `#forms`, it's not applying to the forms when they're on the **wallets.tmpl** inside of `#rightBox`. We need to consolidate those styles.

vctt94:
I removed the declaration of `.form-closer` from #forms, and removed its padding, which give the gap between the `x` and the form.

buck54321:
The 50px padding is to provide some buffer for clicks around the form, since a click often closes the form.

vctt94:
ok, returned the padding. I removed the element as `position: absolute`, and making it a flex-row and now it will be alway on top, at the right of the div's parent.

Non modal: 

image

modal: 

image

with hover: 

image

chappjc:
The X is now in an awkward spot on some forms, like Send and Deposit and Reconfigure wallets:

image

image

It's more reasonable on the top for others though:

image

image

image

vctt94:
Thanks for catching thoses. Fixed.

chappjc (approved):
Good. Dragging those `div`s up was what help for me.  Leaving for @buck54321 to iterate with you now.

buck54321 (approved):

Issue #1730, opened 2022-07-25
server/swap/test: TestMalformedSwap test failure.


JoeGruffins:
Random test failure on pr #1729 but probably unrelated to those changes. A second try by github passed and tests pass locally. Can be seen here: https://github.com/JoeGruffins/dcrdex/runs/7492614387?check_suite_focus=true

```
--- FAIL: TestMalformedSwap (0.01s)
    swap_test.go:1976: error sending maker swap transaction: maker swap rpc error. code: 64, msg: already received a swap contract, search in progress
```

LasTshaMAN:
Looking into this, if nobody minds

chappjc:
OK.

The fundamental issue is described here: https://github.com/decred/dcrdex/pull/1563#discussion_r899257787

The semaphore isn't toggled until after `processInit` sends it's response and returns to the `latencyQ` with `DontTryAgain` thus finally calling `endSwapSearch`.  In the original PR I had `endSwapSearch` called in `processInit` earlier but it wasn't half as clean of a solution.  I don't know if it's realistic outside of a unit test to have this be a race the client can win with a second init request, but this is how we would solve it.

Also, `ensureErr` needs a `t.Helper()` so we know where the error actually happened within

chappjc:
Here's the alternate approach I am referring to: https://github.com/decred/dcrdex/commit/4d82ea4ba00b8f80e29e42451856bd97c54f1205#diff-6af7c56e40de4eb4dba203d67b1b07ccfc43ed88e99811b9697e9d754d01e21cR1469-R1473

LasTshaMAN:
Thanks for additional context on this, 
- I'm unsure of what implementation of `Swapper` (current or the alternative you've suggested) is preferable for all intents and purposes, I think master version is good enough for now (and can be upgraded in the future), but can dig deeper into it if you think it's worth doing so
- and it's the flaky tests that needs improvements anyway, which I'm basically focusing on in the linked PR

chappjc:
Thanks, I think working on the test is best for now.

If a client sends a second init request so fast, they are effectively hammering the server.  All they need to do is slow down for a microsecond for the semaphore to flip and they can submit again, assuming they haven't tripped the global or user request rate limiters.

BTW, the other reason for a "search in progress" response is if they had a previous request that timed out client side, but hadn't yet timed out server-side.  And again that is a try again scenario.

buck54321:
Can close?

JoeGruffins:
[...]

PR #1737, opened 2022-07-26
server/swap/test: fix intermittent TestMalformedSwap failures

LasTshaMAN:
I also pulled out problematic part of `TestMalformedSwap` test into separate `TestRetriesDuringSwap` since it doesn't seem to be related to "testing malformed" (and I've discovered quite a lot of things this way about this functionality).

Closes https://github.com/decred/dcrdex/issues/1730.

### DIFF
--- a/server/market/market.go
+++ b/server/market/market.go
@@ -2039,8 +2039,8 @@ func (m *Market) handlePreimageResp(msg *msgjson.Message, reqData *piData) {
 		log.Errorf("StorePreimage: %v", err)
 		// Fatal backend error. New swaps will not begin, but pass the preimage
 		// along so that it does not appear as a miss to collectPreimages.
-		m.respondError(msg.ID, reqData.ord.User(), msgjson.UnknownMarketError,
-			fmt.Sprintf("internal server error: %v", err))
+		m.respondError(msg.ID, reqData.ord.User(), msgjson.RPCInternalError,
+			"internal server error")
 	}
 
 	sendPI(&pi)
--- a/server/market/market.go
+++ b/server/market/market.go
@@ -884,14 +884,14 @@ func (m *Market) processMatchAcksForCancel(user account.AccountID, msg *msgjson.
 	err := msg.UnmarshalResult(&acks)
 	if err != nil {
 		m.respondError(msg.ID, user, msgjson.RPCParseError,
-			"error parsing match request acknowledgment")
+			fmt.Sprintf("error parsing match request acknowledgment: %v", err))
 		return
 	}
 	// The acknowledgment for both the taker and maker should come from the same user
 	expectedNumAcks := 2
 	if len(acks) != expectedNumAcks {
 		m.respondError(msg.ID, user, msgjson.AckCountError,
-			fmt.Sprintf("expected %d acknowledgements, got %d", len(acks), expectedNumAcks))
+			fmt.Sprintf("expected %d acknowledgements, got %d", expectedNumAcks, len(acks)))
 		return
 	}
 	log.Debugf("processMatchAcksForCancel: 'match' ack received from %v", user)
@@ -1995,7 +1995,7 @@ func (m *Market) handlePreimageResp(msg *msgjson.Message, reqData *piData) {
 	if err != nil {
 		sendPI(nil)
 		m.respondError(msg.ID, reqData.ord.User(), msgjson.RPCParseError,
-			"error parsing preimage notification response")
+			fmt.Sprintf("error parsing preimage notification response: %v", err))
 		return
 	}
 	if resp.Error != nil {
@@ -2007,7 +2007,7 @@ func (m *Market) handlePreimageResp(msg *msgjson.Message, reqData *piData) {
 	if err != nil {
 		sendPI(nil)
 		m.respondError(msg.ID, reqData.ord.User(), msgjson.RPCParseError,
-			"error parsing preimage response payload result")
+			fmt.Sprintf("error parsing preimage response payload result: %v", err))
 		return
 	}
 
@@ -2038,8 +2038,9 @@ func (m *Market) handlePreimageResp(msg *msgjson.Message, reqData *piData) {
 	if err != nil {
 		log.Errorf("StorePreimage: %v", err)
 		// Fatal backend error. New swaps will not begin, but pass the preimage
-		// along so it does not appear as a miss to collectPreimages.
-		m.respondError(msg.ID, reqData.ord.User(), msgjson.UnknownMarketError, "internal server error")
+		// along so that it does not appear as a miss to collectPreimages.
+		m.respondError(msg.ID, reqData.ord.User(), msgjson.UnknownMarketError,
+			fmt.Sprintf("internal server error: %v", err))
 	}
 
 	sendPI(&pi)
--- a/server/swap/swap.go
+++ b/server/swap/swap.go
@@ -986,12 +986,11 @@ func (s *Swapper) processBlock(ctx context.Context, block *blockNotification) {
 			return
 		}
 
-		// Lock the matchTracker so the following checks and updates are atomic
-		// with respect to Status.
 		match.mtx.RLock()
-		defer match.mtx.RUnlock()
+		status := match.Status
+		match.mtx.RUnlock()
 
-		switch match.Status {
+		switch status {
 		case order.MakerSwapCast:
 			if match.makerStatus.swapAsset != block.assetID {
 				break
@@ -1486,7 +1485,7 @@ func (s *Swapper) processAck(msg *msgjson.Message, acker *messageAcker) {
 }
 
 // processInit processes the `init` RPC request, which is used to inform the DEX
-// of a newly broadcast swap transaction. Once the transaction is seen and and
+// of a newly broadcast swap transaction. Once the transaction is seen and
 // audited by the Swapper, the counter-party is informed with an 'audit'
 // request. This method is run as a coin waiter, hence the return value
 // indicates if future attempts should be made to check coin status.
@@ -1649,7 +1648,7 @@ func (s *Swapper) processInit(msg *msgjson.Message, params *msgjson.Init, stepIn
 	s.authMgr.Sign(auditParams)
 	notification, err := msgjson.NewRequest(comms.NextID(), msgjson.AuditRoute, auditParams)
 	if err != nil {
-		// This is likely an impossibly condition.
+		// This is likely an impossible condition.
 		log.Errorf("error creating audit request: %v", err)
 		return wait.DontTryAgain
 	}
@@ -1665,8 +1664,6 @@ func (s *Swapper) processInit(msg *msgjson.Message, params *msgjson.Init, stepIn
 	// Send the 'audit' request to the counter-party.
 	log.Debugf("processInit: sending contract 'audit' request to counterparty %v (%s) "+
 		"for match %v", ack.user, makerTaker(ack.isMaker), matchID)
-	// Send the request.
-
 	// The counterparty will audit the contract by retrieving it, which may
 	// involve them waiting for up to the broadcast timeout before responding,
 	// so the user gets at least s.bTimeout to the request.
--- a/server/swap/swap.go
+++ b/server/swap/swap.go
@@ -2122,7 +2122,7 @@ func (s *Swapper) processMatchAcks(user account.AccountID, msg *msgjson.Message,
 	}
 	if len(matches) != len(acks) {
 		s.respondError(msg.ID, user, msgjson.AckCountError,
-			fmt.Sprintf("expected %d acknowledgements, got %d", len(acks), len(matches)))
+			fmt.Sprintf("expected %d acknowledgements, got %d", len(matches), len(acks)))
 		return
 	}
 
--- a/server/swap/swap.go
+++ b/server/swap/swap.go
@@ -1477,8 +1477,8 @@ func (s *Swapper) processAck(msg *msgjson.Message, acker *messageAcker) {
 	if !acker.isMaker {
 		acker.match.Sigs.TakerRedeem = ack.Sig
 		if err = s.storage.SaveRedeemAckSigB(mktMatch, ack.Sig); err != nil {
-			s.respondError(msg.ID, acker.user, msgjson.UnknownMarketError,
-				fmt.Sprintf("internal server error: %v", err))
+			s.respondError(msg.ID, acker.user, msgjson.RPCInternalError,
+				"internal server error")
 			log.Errorf("SaveRedeemAckSigB failed for match %v: %v", mktMatch.String(), err)
 			return
 		}
@@ -1595,8 +1595,8 @@ func (s *Swapper) processInit(msg *msgjson.Message, params *msgjson.Init, stepIn
 	if err != nil {
 		log.Errorf("saving swap contract (match id=%v, maker=%v) failed: %v",
 			matchID, actor.isMaker, err)
-		s.respondError(msg.ID, actor.user, msgjson.UnknownMarketError,
-			fmt.Sprintf("internal server error: %v", err))
+		s.respondError(msg.ID, actor.user, msgjson.RPCInternalError,
+			"internal server error")
 		// TODO: revoke the match without penalties instead of retrying forever?
 		return wait.TryAgain
 	}
@@ -1705,7 +1705,7 @@ func (s *Swapper) processRedeem(msg *msgjson.Message, params *msgjson.Redeem, st
 	if !chain.ValidateSecret(params.Secret, cpContract) {
 		log.Infof("Secret validation failed (match id=%v, maker=%v, secret=%v)",
 			matchID, actor.isMaker, params.Secret)
-		s.respondError(msg.ID, actor.user, msgjson.UnknownMarketError, "secret validation failed")
+		s.respondError(msg.ID, actor.user, msgjson.InvalidRequestError, "secret validation failed")
 		return wait.DontTryAgain
 	}
 	redemption, err := chain.Redemption(params.CoinID, cpSwapCoin, cpContract)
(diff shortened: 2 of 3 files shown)


server/swap/swap.go
```diff
@@ -2122,7 +2119,7 @@ func (s *Swapper) processMatchAcks(user account.AccountID, msg *msgjson.Message,
 	}
 	if len(matches) != len(acks) {
 		s.respondError(msg.ID, user, msgjson.AckCountError,
-			fmt.Sprintf("expected %d acknowledgements, got %d", len(acks), len(matches)))
+			fmt.Sprintf("expected %d acknowledgements, got %d", len(matches), len(acks)))
```
LasTshaMAN, line 2122: Looks like these 2 args should be switched

server/swap/swap_test.go
```diff
@@ -747,9 +748,25 @@ func (rig *testRig) checkMatchNotification(msg *msgjson.Message, oid order.Order
...
+// helper to check the swap status for specified user.
+func (rig *testRig) checkSwapStatusChange(wantStatus order.MatchStatus, user *tUser) error {
+	// Swap counterparty ultimately gets an audit request. Synchronize with that.
+	if err := rig.waitChans("counterparty received our swap -> counterparty audit request", nil, rig.auth.swapReceived, rig.auth.auditReq); err != nil {
+		return err
+	}
+	tracker := rig.getTracker()
+	tracker.mtx.RLock()
+	status := tracker.Status
+	tracker.mtx.RUnlock()
```
LasTshaMAN, line 760: Seems like there are a bunch of places in these tests where mutex should be used, like this one,
if reviewers agree, I'd add the required mutex calls I've spotted in this test-file

server/swap/swap.go
```diff
@@ -983,12 +983,11 @@ func (s *Swapper) processBlock(ctx context.Context, block *blockNotification) {
...
 
-		// Lock the matchTracker so the following checks and updates are atomic
-		// with respect to Status.
 		match.mtx.RLock()
-		defer match.mtx.RUnlock()
+		status := match.Status
+		match.mtx.RUnlock()
 
-		switch match.Status {
+		switch status {
```
LasTshaMAN, line 990: Does it make sense to use `defer` here ? my linter complains about it, and it seems that `match.Status` is all that we really need to access in this `for` loop (based on `matchTracker.mtx` comment), so I've trimmed down mutex usage in this case,

unless there is another reason to hold this mutex until this function terminates ?

server/swap/swap_test.go
LasTshaMAN: and I've actually encountered a race couple of times while running these tests locally:

```
==================
WARNING: DATA RACE
Write at 0x00c0004be348 by goroutine 38:
  decred.org/dcrdex/server/swap.(*Swapper).processInit()
      /Users/iurii/dcrdex/server/swap/swap.go:1618 +0x1c07
  decred.org/dcrdex/server/swap.(*Swapper).handleInit.func1()
      /Users/iurii/dcrdex/server/swap/swap.go:1957 +0x6b
  decred.org/dcrdex/dex/wait.(*TaperingTickerQueue).Run.func1()
      /Users/iurii/dcrdex/dex/wait/queue.go:197 +0xfa
  decred.org/dcrdex/dex/wait.(*TaperingTickerQueue).Run·dwrap·5()
      /Users/iurii/dcrdex/dex/wait/queue.go:248 +0x47

Previous read at 0x00c0004be348 by goroutine 91:
  decred.org/dcrdex/server/swap.(*testRig).sendSwap_maker()
      /Users/iurii/dcrdex/server/swap/swap_test.go:777 +0x384
  decred.org/dcrdex/server/swap.TestRetrySwapInit()
      /Users/iurii/dcrdex/server/swap/swap_test.go:2041 +0x2de
  testing.tRunner()
      /Users/iurii/go/src/testing/testing.go:1259 +0x22f
  testing.(*T).Run·dwrap·21()
      /Users/iurii/go/src/testing/testing.go:1306 +0x47
```

server/swap/swap_test.go
```diff
@@ -823,11 +816,13 @@ func (rig *testRig) sendSwap(user *tUser, oid order.OrderID, recipient string) (
 	} else {
 		rig.abcNode.setContract(swap.coin, false)
 	}
+	rig.auth.mtx.Lock()
 	rig.auth.swapID = swap.req.ID
+	rig.auth.mtx.Unlock()
```
LasTshaMAN, line 821: Also seems like there is a race with `TAuthManager.swapID` field here (and this ^ should fix it):
```
WARNING: DATA RACE
Write at 0x00c0003c1cd0 by goroutine 97:
  decred.org/dcrdex/server/swap.(*testRig).sendSwap()
      /Users/iurii/dcrdex/server/swap/swap_test.go:829 +0x2f9
  decred.org/dcrdex/server/swap.(*testRig).sendSwap_maker()
      /Users/iurii/dcrdex/server/swap/swap_test.go:765 +0x127
  decred.org/dcrdex/server/swap.TestRetrySwapInit()
      /Users/iurii/dcrdex/server/swap/swap_test.go:2045 +0x2de
  testing.tRunner()
      /Users/iurii/go/src/testing/testing.go:1259 +0x22f
  testing.(*T).Run·dwrap·21()
      /Users/iurii/go/src/testing/testing.go:1306 +0x47

Previous read at 0x00c0003c1cd0 by goroutine 42:
  decred.org/dcrdex/server/swap.(*TAuthManager).Send()
      /Users/iurii/dcrdex/server/swap/swap_test.go:159 +0x3ac
  decred.org/dcrdex/server/swap.(*Swapper).respondSuccess()
      /Users/iurii/dcrdex/server/swap/swap.go:1258 +0x1c3
  decred.org/dcrdex/server/swap.(*Swapper).processInit()
      /Users/iurii/dcrdex/server/swap/swap.go:1631 +0x2345
  decred.org/dcrdex/server/swap.(*Swapper).handleInit.func1()
      /Users/iurii/dcrdex/server/swap/swap.go:1956 +0x6b
  decred.org/dcrdex/dex/wait.(*TaperingTickerQueue).Run.func1()
      /Users/iurii/dcrdex/dex/wait/queue.go:197 +0xfa
  decred.org/dcrdex/dex/wait.(*TaperingTickerQueue).Run·dwrap·5()
      /Users/iurii/dcrdex/dex/wait/queue.go:248 +0x47
```

server/swap/swap.go
chappjc: Just for the reason in the comment, to make sure that any other goroutine viewing `matchTracker.Status` and `swapStatus.swapConfirmed` (i.e. `checkInactionEventBased` and `checkInactionBlockBased`) has a consistent view of those things.

It probably doesn't matter, otherwise what's the point of a separate mutex for the `swapStatus`, but it needs to be examined carefully.

server/swap/swap_test.go
chappjc: Maybe another test already had the same issue, but `TAuthManager.swapID` being modified while there is as swap presently being handled with `handleInit` doesn't make sense.  (If a subsequent `sendSwap` modified it while a previous `handleInit` was already running, then the `Send` for the response on the first one would reference the wrong `swapID`! Thus the test rig won't work right unless the swapID happens to be the same like it is in your test).

In other words, this `swapID` field is just a single value because the test rig isn't designed to work properly with multiple concurrent `handleInit` calls, and all of the prior tests respected that.

I think you're only seeing this race now because of what `TestRetrySwapInit` is trying to test.
OK to locking in `sendSwap`, but I don't think other tests were hitting this race, and any that do including this new one are potentially in error.

server/swap/swap_test.go
```diff
@@ -2003,18 +1998,54 @@ func TestMalformedSwap(t *testing.T) {
...
+	// adequate fee and 0 confs).
+	startWaitingTime := time.Now()
+	for {
+		// We might get a couple of duplicate init request errors here, in that case
+		// we simply retry request later.
+		err := rig.sendSwap_maker(false)
+		if err == nil {
+			// We are done, finally got a retry attempt that worked.
+			ensureNilErr(rig.checkSwapStatusChange(order.MakerSwapCast, matchInfo.maker))
+			return
```
chappjc, line 2038: Currently we are indeed done after this, but I'd prefer to `break` the loop instead and achieve the same outcome just by running to the end of the test function.  It'll allow any coded that may eventually get added after the `for` to work.  Linters will call that out, but `go` will still compile and run the test with unreachable code.

chappjc (commented):
Looks pretty good over all.  Thanks for the good clean up and fixing my broken test.

Sorry I just noticed this was draft, but I wanted to see.  :)

server/swap/swap_test.go
LasTshaMAN: Interesting ... haven't thought about increasing test "coverage" this way, makes sense; so I see 2 options then:
- leave other tests as is (maybe I can go over those to see whether I can spot any potential concurrency issues and raise if any), preserving this "coverage", letting race detector do its job for us (and address these, if any, on a need-to basis)
- or (lazy approach I've suggested above) use mutex blindly, basically giving up on this additional "coverage" at the benefit of not having to worry about writing / maintaining precise tests with respect to concurrent access

I can see the benefits of either of these (personally I'm used to option 2) and I'm onboard with whichever you prefer

server/market/market.go
```diff
@@ -876,14 +876,14 @@ func (m *Market) processMatchAcksForCancel(user account.AccountID, msg *msgjson.
...
-			"error parsing match request acknowledgment")
+			fmt.Sprintf("error parsing match request acknowledgment: %v", err))
 		return
 	}
 	// The acknowledgment for both the taker and maker should come from the same user
 	expectedNumAcks := 2
 	if len(acks) != expectedNumAcks {
 		m.respondError(msg.ID, user, msgjson.AckCountError,
-			fmt.Sprintf("expected %d acknowledgements, got %d", len(acks), expectedNumAcks))
+			fmt.Sprintf("expected %d acknowledgements, got %d", expectedNumAcks, len(acks)))
```
LasTshaMAN, line 886: This is mixed as well (similar to one I've spotted above)

server/market/market.go
```diff
@@ -1992,8 +1992,9 @@ func (m *Market) handlePreimageResp(msg *msgjson.Message, reqData *piData) {
 	if err != nil {
 		log.Errorf("StorePreimage: %v", err)
 		// Fatal backend error. New swaps will not begin, but pass the preimage
-		// along so it does not appear as a miss to collectPreimages.
-		m.respondError(msg.ID, reqData.ord.User(), msgjson.UnknownMarketError, "internal server error")
+		// along so that it does not appear as a miss to collectPreimages.
+		m.respondError(msg.ID, reqData.ord.User(), msgjson.UnknownMarketError,
+			fmt.Sprintf("internal server error: %v", err))
```
LasTshaMAN, line 1997: I'm assuming here there is "nothing to hide" from the client; in other words, this error shouldn't contain sensitive information, thoughts ?

server/swap/swap.go
```diff
@@ -1420,7 +1419,7 @@ func (s *Swapper) processAck(msg *msgjson.Message, acker *messageAcker) {
 	ack := new(msgjson.Acknowledgement)
 	err := msg.UnmarshalResult(ack)
 	if err != nil {
-		s.respondError(msg.ID, acker.user, msgjson.RPCParseError, "error parsing acknowledgment")
+		s.respondError(msg.ID, acker.user, msgjson.RPCParseError, fmt.Sprintf("error parsing acknowledgment: %v", err))
```
LasTshaMAN, line 1422: Otherwise `err` is swallowed (updated a couple of similar places with the same issue, can find them all by searching `fmt.Sprintf` in this PR):
```
2022-07-27 11:57:10.586 [DBG] TEST: Error going to user 224cbaaafa80bf3bd1ff731590bcbdda5a76f91e60a156994634e91c0000002e, code: 1, msg: error parsing acknowledgment
```

server/market/market.go
LasTshaMAN: Also, seems like `msgjson.UnknownMarketError` code here (and in some other similar places) isn't the best fit ? Thought I'd call this out, and maybe fix.

server/swap/swap_test.go
chappjc: If there's a place where you can reason out a race even if it's not detected, use the mutex.

We're sort of used to the lazy approach in much of the product code, and I can say that it bites us.   It's very important to have a good understanding of access patterns because over using mutexes can be bad for not only performance but it can get you into deadlocks and undesirable mutex contention.

server/swap/swap.go
```diff
@@ -1475,15 +1474,16 @@ func (s *Swapper) processAck(msg *msgjson.Message, acker *messageAcker) {
 	if !acker.isMaker {
 		acker.match.Sigs.TakerRedeem = ack.Sig
 		if err = s.storage.SaveRedeemAckSigB(mktMatch, ack.Sig); err != nil {
-			s.respondError(msg.ID, acker.user, msgjson.UnknownMarketError, "internal server error")
+			s.respondError(msg.ID, acker.user, msgjson.UnknownMarketError,
+				fmt.Sprintf("internal server error: %v", err))
```
chappjc, line 1478: For any of the errors from an Archivist, I'm inclined to log server-side and omit the actual error from the response to client.  Unless we are explicitly checking the error type (see the `Is*` helpers in `server/db/errors.go`), we should assume the error is an internal failure and that's all the client needs to know.

server/swap/swap.go
```diff
@@ -1502,7 +1502,7 @@ func (s *Swapper) processInit(msg *msgjson.Message, params *msgjson.Init, stepIn
 			stepInfo.match.ID(), actor, params.CoinID, params.Contract, err)
 		actor.status.mtx.RUnlock()
 		s.respondError(msg.ID, actor.user, msgjson.ContractError,
-			"redemption error")
+			fmt.Sprintf("contract error encountered: %v", err))
```
chappjc, line 1505: Good catch!  That's not a redemption error lol.  In this case I don't see the harm in including the full `err`.

server/swap/swap_test.go
chappjc: On this note, while I think it's good to be acutely aware of what goroutines are running and why it is or isn't safe to access some field, I'm no longer fond of the general approach in our tests of being so deeply involved with internals and implementation details rather than testing the API.  The fact that we have to do this is partly a consequence of the need to stub out connections and databases, but it's also a hint that we've not factored out enough logic into smaller functions so we can test them in isolation, among other things like wanting to reuse structures instead of making fresh ones, doing more repeated teardown/bringup, etc.

It's frustrating to change some small detail in the implementation that does not affect the overall function of the system but ends up completely wrecking a test that expects some irrelevant order of operations or some message that we hack a channel into via a stub, etc.

This is not a comment that you should change anything, just saying that while I always prefer a developer to think hard about access patterns and if a mutex is even needed, I generally do not love our overall testing approach.  This is an evolving view of mine.

server/market/market.go
LasTshaMAN: So, would it be ok to replace `msgjson.UnknownMarketError` (and in similar cases where appropriate) with `msgjson. RPCInternalError` ? and revert back `fmt.Sprintf("internal server error: %v", err)` -> `"internal server error"`,

> The error is likely to be from postgres here, and sending it blindly to the client might leak info about server config

since looks like this is applicable only to "internal" cases calling `respondError` I've modified in this PR

chappjc: That sounds good.  The other unmarshal errors and similar can still go back to the client though I think.

server/swap/swap_test.go
LasTshaMAN: Revised this a bit, I think I was observing this race while `TestRetrySwapInit` wasn't "complete", now it can locking on this mutex can be safely removed I think (done).

server/swap/swap_test.go
LasTshaMAN: I've looked a bit more at how `tracker.Status` is accessed, (in `TestRetrySwapInit`) it gets synchronized on `rig.auth.swapReceived` (or for redeems it would be `rig.auth.redeemReceived`) with the concurrent Go-routine that is running `processInit`, thus locking `tracker.mtx` isn't really necessary (so, I removed locking).
In other tests there seems to be no concurrent access at all.

server/swap/swap_test.go
```diff
@@ -2003,18 +1994,54 @@ func TestMalformedSwap(t *testing.T) {
...
-	// Back to NewlyMatched
 	tracker := rig.getTracker()
+	// We're "rewinding time" back NewlyMatched status to be able to retry sending
+	// the same init request again.
+	tracker.mtx.Lock()
 	tracker.Status = order.NewlyMatched
-	// We're rewinding time to send the same init again. To support this hack,
-	// wait for swapSearching to clear.
-	time.Sleep(200 * time.Millisecond)
+	tracker.mtx.Unlock()
```
LasTshaMAN, line 2021: Here, I still think we need to lock in order for next `processInit` to be able to read this change (it will also lock on this `tracker.mtx` for reading),

still, I couldn't reproduce this race (without locking)

LasTshaMAN: Ah well, looking at it afresh I can see why there is no race, calling `s.latencyQ.Wait` establishes "happens after" via [this chan](https://github.com/LasTshaMAN/dcrdex/blob/1201862002cab39d58b5d4366bd6e0e251367341/dex/wait/queue.go#L183-L183),

go-routine that will be working on `processInit` will have to receive on this chan first

server/market/market.go
LasTshaMAN: Done.

LasTshaMAN:
Addressed all the things discussed above, and also added retry-test for redeems (similar to init), hopefully you don't mind slight  code rearrangements in tests.

Ready for another review round now.

server/swap/swap.go
chappjc: Revert this please and we can get this in.

LasTshaMAN: Done.

chappjc (approved):

PR #1738, opened 2022-07-27
multi: implement a market-maker bot

buck54321:
```
core:
Add a market-maker bot to **core**. The bot uses some internal and
(optionally) external signals to calculate a ideal buy and sell
price and a "break-even spread", which is the spread at which
a buy-sell sequence's tx fees equal its profit. These values can
be used as inputs into one of five "gap strategies", which determine
the target spread.

The bot can be created, started, updated, paused, and retired. Each
of those actions generates a notification.

db:
New BotProgram type is not specific to market-maker bots. Everything
is structured with a mind towards future expansion with new automated
trading routines.

ui:
New /mm view enables creation, re-configuration, and monitoring of
existing market-maker bots.
```


buck54321:
image

buck54321:
I've made the simnet test into a full-fledged cli bot. This means I could've foregone the **simharness** package move, but I do plan to use that package other places, so I'd love to keep that change.

client/db/bolt/db.go
```diff
@@ -1114,36 +1103,42 @@ func (db *BoltDB) UpdateOrderMetaData(oid order.OrderID, md *dexdb.OrderMetaData
...
+		put(statusKey, uint16Bytes(uint16(md.Status))).
+		put(updateTimeKey, uint64Bytes(timeNow())).
+		put(proofKey, md.Proof.Encode()).
+		put(changeKey, md.ChangeCoin).
+		put(linkedKey, linkedB).
+		put(swapFeesKey, uint64Bytes(md.SwapFeesPaid)).
+		put(maxFeeRateKey, uint64Bytes(md.MaxFeeRate)).
+		put(redemptionFeesKey, uint64Bytes(md.RedemptionFeesPaid)).
+		put(fromVersionKey, uint32Bytes(md.FromVersion)).
+		put(toVersionKey, uint32Bytes(md.ToVersion)).
```
chappjc, line 1134: I think at least the two max fee rates and to/from versions are immutable and thus best suited for teh `put`s back in `UpdateOrder`.

client/cmd/mmbot/main.go
```diff
@@ -0,0 +1,216 @@
...
+	flag.StringVar(&logLevel, "log", "info", "universal or per-logger log level. e.g. log=trace, or log=CORE=trace,BOT=trace")
+	flag.Float64Var(&manualRate, "emptyrate", 0, "(optional) a rate to use if the book is empty an there are no oracles available")
+	flag.Parse()
+
+	if coreDir != defaultCoreDir && configFile == defaultConfigFilePath {
+		configFile = filepath.Join(coreDir, defaultConfigFileName)
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
chappjc, line 56: Not that it matters, but this defer will never run on account of the `os.Exit`s

client/cmd/mmbot/main.go
```diff
@@ -0,0 +1,216 @@
...
+		return fmt.Errorf("no market %q", mktName)
+	}
+
+	if _, err := c.Login(appPW); err != nil {
+		return fmt.Errorf("login error: %w", err)
+	}
+
+	printStartMessage := func() {
+		fmt.Println(":::::: Base Wallet Address: ", b.Wallet.Address)
+		fmt.Println(":::::: Quote Wallet Address:", q.Wallet.Address)
```
chappjc, line 176: Why not use `b.Symbol` and `q.Symbol` instead of the "Base" and "Quote" literals.

chappjc:
> I've made the simnet test into a full-fledged cli bot.

Glad to see it.

> This means I could've foregone the **simharness** package move, but I do plan to use that package other places, so I'd love to keep that change.

In that case, will you please put the simharness.go creation and the loadbot changes in a separate commit?  There's no clear connection to the mm bot work.

client/core/bookie.go
```diff
@@ -385,7 +385,7 @@ func (dc *dexConnection) bookie(marketID string) *bookie {
 // syncBook subscribes to the order book and returns the book and a BookFeed to
 // receive order book updates. The BookFeed must be Close()d when it is no
 // longer in use. Use stopBook to unsubscribed and clean up the feed.
-func (dc *dexConnection) syncBook(base, quote uint32) (BookFeed, error) {
+func (dc *dexConnection) syncBook(base, quote uint32) (*orderbook.OrderBook, BookFeed, error) {
```
chappjc, line 388: This is funny because originally we have the *OrderBook returned, then we removed it and had the orderbook snap as the first send on the feed (`FreshBookAction` I believe). It is certainly most convenient this way.

client/core/core.go
```diff
@@ -4619,11 +4702,15 @@ func (c *Core) PreOrder(form *TradeForm) (*OrderEstimate, error) {
 // Trade is used to place a market or limit order.
 func (c *Core) Trade(pw []byte, form *TradeForm) (*Order, error) {
 	// Check the user password.
-	crypter, err := c.encryptionKey(pw)
-	if err != nil {
-		return nil, fmt.Errorf("Trade password error: %w", err)
+	var crypter encrypt.Crypter
+	if len(pw) > 0 {
```
chappjc, line 4706: Will you please comment on this and the related change in prepareTrackedTrade?

Also, the diff in prepareTrackedTrade would be been a lot tighter if `prepareWallet` simply returned early if `crypter == nil`.

client/core/bot.go
```diff
@@ -0,0 +1,1712 @@
...
+	fiatConversions() map[uint32]float64
+}
+
+// basisPrice is the oracle-weighted price on which to base the target price and
+// spread. The oracle weighting and bias is applied, but no other adjustments.
+func basisPrice(host string, mkt *Market, oracleBias, oracleWeighting float64, midGap uint64, bp basisPricer, log dex.Logger) uint64 {
+	basisPrice := float64(midGap) // float64 message-rate units
+
+	log.Tracef("basisPrice: mid-gap price = %d", midGap)
+
```
buck54321, line 598: We should be a little more careful with using the current mid-gap here. If the book is sparse, someone could manipulate the bot into providing a favorable price by placing orders far from the gap. We should try to get some metric of order book health before using the mid-gap as an input, and we should probably only allow the mid-gap to vary so much from the oracle rate.

martonp:
Looks really cool. I'm just starting to look through, but I found an issue.. if you press enter instead of clicking the submit button when creating a bot, it will create two bots instead of just one. If you click the button, only one button is created.

Also, I created a bot, and I have funds in both wallets, but it is only creating orders on the buy side, not the sell side. Isn't it supposed to create for both?

Screen Shot 2022-08-14 at 12 18 52 PM
Screen Shot 2022-08-14 at 12 19 03 PM

Also, there should be a header on that table with the CEX info. And a K after the volume values.

buck54321:
> if you press enter instead of clicking the submit button when creating a bot, it will create two bots instead of just one.

When you say "press enter", do you mean in the lot number input? I've tried on Chrome and Firefox, and can't seem to repro, which makes sense because it's not part of a form and there's no handler on that input to submit. 

Maybe I've misunderstood. Please verify.

> but it is only creating orders on the buy side

Not really sure about this, but I squashed some minor bugs that may have contributed. If you still see it, please run with 
`--log=BOT=trace,CORE=trace`
and see if anything seems off.

martonp:
> When you say "press enter", do you mean in the lot number input? I've tried on Chrome and Firefox, and can't seem to repro, which makes sense because it's not part of a form and there's no handler on that input to submit.

I meant press enter when submitting the password.

I'll check about the other issue later.

client/core/bot_live_test.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func TestMarketAveragedPrice(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	m := &makerBot{
+		Core: &Core{},
+		base: &makerAsset{
+			Asset: &dex.Asset{Symbol: "dcr"},
```
chappjc, line 51: Looks like it's working on an old definition of `makerAsset`.  There's no `dex.Asset`, but there is a `core.SupportedAsset`

client/core/core.go
```diff
@@ -237,6 +215,66 @@ func (dc *dexConnection) marketMap() map[string]*Market {
...
+		if m.Name == mktName {
+			msgMkt = m
+			break
+		}
+	}
+	if msgMkt == nil {
+		return nil
+	}
+
+	mkt := coreMarketFromMsgMarket(dc, msgMkt)
```
chappjc, line 238:
```suggestion
	var mkt *Market
	for _, m := range cfg.Markets {
		if m.Name == mktName {
			mkt = coreMarketFromMsgMarket(dc, m)
			break
		}
	}
	if mkt == nil {
		return nil
	}
```

client/core/notification.go
```diff
@@ -84,11 +88,23 @@ func (c *Core) notify(n Notification) {
 // channel has capacity 1024, and should be monitored for the lifetime of the
 // Core. Blocking channels are silently ignored.
 func (c *Core) NotificationFeed() <-chan Notification {
```
chappjc, line 90: How 'bout we export this pattern for webserver, et. al?  Except it could be simplified:

```go
func (c *Core) NotificationFeed() (feed <-chan Notification, done func()) {
	ch := make(chan Notification, 1024)
	cid := atomic.AddUint64(&noteChanCounter, 1)
	c.noteMtx.Lock()
	c.noteChans[cid] = ch
	c.noteMtx.Unlock()
	return ch, func() {
		c.noteMtx.Lock()
		delete(c.noteChans, cid)
		c.noteMtx.Unlock()
	}
}
```

client/core/simharness/simharness.go
```diff
@@ -0,0 +1,474 @@
...
+	p := &process{
+		cmd:    command,
+		stopFn: stopFn,
+	}
+	if err := command.Start(); err != nil {
+		return fmt.Errorf("unable to start process: %v", err)
+	}
+	processes = append(processes, p)
+	return nil
+}
```
chappjc, line 372: Some fairly significant duplication from loadbot.  More trouble than it's worth to find a common home for these kinds of harness interactions?

client/core/simharness/simharness.go
```diff
@@ -0,0 +1,474 @@
...
+			}
+		}
+	}
+	return
+}
+
+var zecSendMtx sync.RWMutex
+
+// Mine will mine a single block on the node and asset indicated.
+func Mine(ctx context.Context, symbol, node string) <-chan *HarnessResult {
```
chappjc, line 149: Unused.  Thinking to use this in other spots like loadbot and eth/nodeclient_harness_test.go?

About 90 loc in `RPCWalletForm` also unused.  Should all of this simharness stuff be in this PR?

client/cmd/mmbot/main.go
```diff
@@ -0,0 +1,215 @@
...
+	"decred.org/dcrdex/server/admin"
+
+	_ "decred.org/dcrdex/client/asset/btc"
+	_ "decred.org/dcrdex/client/asset/dcr"
+)
+
+var (
+	userDir, _            = os.UserHomeDir()
+	defaultConfigFileName = "mm.json"
+	defaultCoreDir        = filepath.Join(userDir, ".dexc")
```
chappjc, line 30: Why not dcrutil.AppData?  Using home/.dexc isn't going to jive with Windows and probably macos

client/cmd/mmbot/main.go
```diff
@@ -0,0 +1,215 @@
+package main
```
chappjc, line 1: copyright lines pls

client/cmd/mmbot/main.go
```diff
@@ -0,0 +1,215 @@
...
+	go func() {
+		for range killChan {
+			if c != nil && pgmID > 0 {
+				if err := c.RetireBot(pgmID); err != nil {
+					fmt.Fprintf(os.Stderr, "error retiring bot: %v \n", err)
+				}
+
+			}
+			log.Infof("Shutdown signal received")
+			cancel()
```
chappjc, line 82: And `return` or `break`?  Do you want to allow subsequent interrupts to call RetireBot repeatedly?

chappjc (changes requested):
A first pass.

The new `core/simharness` package looks like a cool reusability refactor of some of the code in loadbot and elsewhere, but I don't see how it's related to this PR.  Shouldn't it be delayed for when we're ready to put it into action in loadbot and eth/node_harness_test.go?

The `cmd/mmbot` app looks fine.  I'm fairly certain that in the near future we'll want to hack in some paths for automated init/wallet-creation/register to suit existing workflows.
Also, it looks like it won't be a problem to start multiple bot programs within the one app, but if I'm not seeing an possible issue with that, let me know.

client/core/bot.go
```diff
@@ -0,0 +1,1733 @@
...
+	GapStrategyAbsolute GapStrategy = "absolute"
+	// GapStrategyAbsolutePlus sets the spread to the rate difference plus the
+	// break-even gap.
+	GapStrategyAbsolutePlus GapStrategy = "absolute-plus"
+	// GapStrategyPercent sets the spread as a ratio of the mid-gap rate.
+	// 0 <= r <= 0.1
+	GapStrategyPercent GapStrategy = "percent"
+	// GapStrategyPercentPlus sets the spread as a ratio of the mid-gap rate
+	// plus the break-even gap.
+	GapStrategyPercentPlus GapStrategy = "percent-plus"
```
JoeGruffins, line 66: I don't have a great grasp on what is needed here, but seems like `Plus` categories could just be an option and don't need separate strategies. Also, why add the break-even? Could set to break even if gap is lower also work?

Asking for my understanding.

JoeGruffins (approved):
I was able to get banned on btc/dcr on simnet because the btc native wallet was too slow seemingly and missed a lot of redemptions. But anyway, still says running after I was banned.

image

The Oracle Bias slider is visible when weight is zero if you move the bar on another bot and then go to a bot with zero. Probably unintended.

image

Working well and looks good to me.

client/core/bot.go
buck54321: Gaps above the break-even are "profitable". A gap below the break-even guarantees losses. Having break-even as a minimum allows the bot to adapt to changing market conditions, particularly changes in tx fee rates, without user input.

buck54321:
I thought I had deleted the simharness stuff already, but I guess I never `git rm`'d. It's gone now.

client/core/bot.go
buck54321: > Could set to break even if gap is lower also work?

I'm not sure I understand this question.

chappjc:
> I was actually kinda hoping you would iterate on the cli app after this.

Gladly.

> Here are the three big enhancements I'd like to see going forward.

Yes, no question.  We need these, particularly arb.

client/core/bot.go
JoeGruffins: > I'm not sure I understand this question.

If break even is .07 and the program would set to .1 adding sets it to .17 but really .1 was already enough to bring it above.

client/core/bot.go
```diff
@@ -0,0 +1,1734 @@
...
+			log.Errorf("Could not calculate break-even spread: %v", err)
+			return
+		}
+	}
+
+	// Apply the base strategy.
+	var halfSpread uint64
+	switch pgm.GapStrategy {
+	case GapStrategyMultiplier:
+		halfSpread = uint64(math.Round(float64(breakEven) * (1 + pgm.GapFactor)))
```
martonp, line 930: Why is this `1 + pgm.GapFactor`? It seems correct if it is exactly break even when the `GapFactor` is 1.

martonp (commented):
Looking really nice.

A few UI issues:
- After pausing a bot, it still shows running
- The settings icon on the top right corner is not visible when starting the app. If I type `/mm` into the address bar, the settings icon shows up. 
- The drift tolerance minimum is actually  0.1%, but you can enter 0 in the option selector.

client/core/bot.go
```diff
@@ -0,0 +1,1734 @@
...
+			msgOracleRate *= 1 + oracleBias
+
+			log.Tracef("basisPrice: biased oracle price = %.0f", msgOracleRate)
+		}
+
+		if basisPrice == 0 { // no mid-gap available. Just use the oracle value straight.
+			basisPrice = msgOracleRate
+			log.Tracef("basisPrice: using basis price %.0f from oracle because no mid-gap was found in order book", basisPrice)
+		} else if oracleWeighting > 0 {
+			basisPrice = msgOracleRate*oracleWeighting + basisPrice*(1-oracleWeighting)
```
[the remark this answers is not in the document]
buck54321: For a low-depth book, it would be super easy to drain a bot at a favorable price by manipulating the mid-gap rate.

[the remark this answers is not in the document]
martonp: So the oracle weight makes it harder to do this type of manipulation, but if someone is able to do a major manipulation, then even a 50% weight wouldn't save the victim bots. I think it would make sense to have the oracle weight combined with another setting that says if the difference between the mid-gap rate and the oracle rate is greater than some percentage, then don't place any orders.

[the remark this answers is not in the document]
martonp: > I think I've captured at least part of what you're asking with maxOracleMismatch

Ah yeah that's right. 

Another feature could be for the bot to turn off if it keeps losing money. I'd definitely have more peace of mind to leave it on after going to sleep like that.

client/core/bot.go
```diff
@@ -0,0 +1,1734 @@
...
+	OracleBias float64 `json:"oracleBias"`
+
+	// EmptyMarketRate can be set if there is no market data available, and is
+	// ignored if there is market data available.
+	EmptyMarketRate float64 `json:"manualRate"`
+}
+
+// validateProgram checks the sensibility of a *MakerProgram's values and sets
+// some defaults.
+func validateProgram(pgm *MakerProgram) error {
```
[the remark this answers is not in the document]
buck54321: We just can't know here whether it's required or not. For many or most markets, it probably isn't needed.

client/core/bot.go
```diff
@@ -0,0 +1,1734 @@
...
+//  3. The break-even spread serves as a hard minimum, and is used to determine
+//     the target spread based on the specified gap strategy, giving the target
+//     buy and sell prices.
+//  4. Scan existing orders to determine if their prices are still valid,
+//     within DriftTolerance of the buy or sell price. If not, schedule them
+//     for cancellation.
+//  5. Calculate how many lots are needed to be ordered in order to meet the
+//     2 x L commitment. If low balance restricts the maintenance of L lots on
+//     one side, allow the difference in lots to be added to the opposite side.
+//  6. Place orders, cancels first, then buys and sells.
```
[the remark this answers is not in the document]
buck54321: This was more about just mapping out the bot game plan. You're correct that we can't control the matching sequence of orders placed in the same epoch.

martonp (approved):
I think it's looking really good. I see no issues when the bot is trading against the loadbot. 

I think it would be nice to have a report of all of a bot's actions and how it performed. Also, something like the loadbot that simulates different market conditions would be cool.

client/core/bot.go
```diff
@@ -0,0 +1,1744 @@
...
+	r.BasisPrice = mkt.MsgRateToConventional(basisPrice(host, mkt, zeroOracleBias, zeroOracleWeight, midGap, c, c.log))
+	return r, setBreakEven()
+}
+
+// Spreader is a function that can generate market spread data for a known
+// exchange.
+type Spreader func(ctx context.Context, baseSymbol, quoteSymbol string) (sell, buy float64, err error)
+
+var spreaders = map[string]Spreader{
+	"binance.com":  fetchBinanceSpread,
```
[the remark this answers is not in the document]
chappjc: Pls skip gate, not hi fidelity numbers.

client/core/bot.go
```diff
@@ -0,0 +1,1744 @@
...
+	}
+
+	var maxBuyLots, maxSellLots uint64
+	maxBuy, err := c.MaxBuy(pgm.Host, pgm.BaseID, pgm.QuoteID, rate)
+	if err == nil {
+		maxBuyLots = maxBuy.Swap.Lots
+	} else {
+		m.log.Errorf("Bot MaxBuy error: %v", err)
+	}
+	maxSell, err := c.MaxSell(pgm.Host, pgm.BaseID, pgm.QuoteID)
```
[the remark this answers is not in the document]
buck54321: I think this should be part of a larger effort to handle "inventory management". Essentially, bots need to have have limits to how much funds they can access. I think `Core` will unfortunately need to be involved and it might be a pita.

client/core/bot.go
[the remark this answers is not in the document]
buck54321: Not adding here, but can do after.

client/core/bot.go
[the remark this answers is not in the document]
martonp: Yeah I was thinking the same thing, definitely sounds like a pita.

PR #1740, opened 2022-07-28
simnet_trade_tests: Fix eth and doge.

### LINKED ISSUE #1731: client/simnet-trade-tests: Some fail.
Running the simnet test with dcreth fail.
```
[simnet-trade-tests]$ ./run dcreth
...
2022-07-25 14:16:27.033 [INF] T[1]: updating balances
2022-07-25 14:16:27.033 [INF] T[1]: dcr available 1502.99994980, immature 0.00000000, locked 0.00000000
2022-07-25 14:16:27.034 [INF] T[1]: eth available 8.402879083, immature 0.000000000, locked 0.000000000
2022-07-25 14:16:27.034 [INF] T[1]: dcr balance change -20.00002500 is in expected range of -21.00002625 - -19.00002375
ERROR: test "success" failed: eth balance change not in expected range -0.010080000 - -0.009120000, got 0.002879083
```

also dcrdoge
```
[simnet-trade-tests]$ ./run dcrdoge
...
2022-07-25 14:51:43.774 [INF] CORE:1[dcr]: Funding 2 DCR order with coins [5f16297da4e0fbfc1fdb754bfb4cace94d36699215c4bb29aad274767572bfc9:1] worth 2.99995
2022-07-25 14:51:43.780 [INF] T[1]: placed order selling 2.00000000 dcr at 1.500000 doge/dcr (51e1b062a7e6513f)
ERROR: test "success" failed: client2 place buy order error: FundOrder error for doge, funding quantity 300000000 (2 lots): suggested fee > configured limit. 4000 > 1400
```

JoeGruffins:
closes #1731

### DIFF
--- a/client/core/simnet_trade.go
+++ b/client/core/simnet_trade.go
@@ -49,6 +49,7 @@ import (
 	"decred.org/dcrdex/dex/calc"
 	"decred.org/dcrdex/dex/config"
 	"decred.org/dcrdex/dex/encode"
+	dexdoge "decred.org/dcrdex/dex/networks/doge"
 	"decred.org/dcrdex/dex/order"
 	dexsrv "decred.org/dcrdex/server/dex"
 	"golang.org/x/sync/errgroup"
@@ -1452,6 +1453,11 @@ var cloneTypes = map[uint32]string{
 	133: "zcashdRPC",
 }
 
+// accountBIPs is a map of account based assets. Used in fee estimation.
+var accountBIPs = map[uint32]bool{
+	eth.BipID: true,
+}
+
 func dcrWallet(wt SimWalletType, node string) (*tWallet, error) {
 	switch wt {
 	case WTSPVNative:
@@ -1532,7 +1538,7 @@ func btcCloneWallet(assetID uint32, node string, wt SimWalletType) (*tWallet, er
 
 	switch assetID {
 	case doge.BipID, zec.BipID:
-	// dogecoind doesn't support > 1 wallet, so gamma and delta
+	// dogecoind and zcash don't support > 1 wallet, so gamma and delta
 	// have their own nodes.
 	default:
 		switch node {
@@ -1550,6 +1556,13 @@ func btcCloneWallet(assetID uint32, node string, wt SimWalletType) (*tWallet, er
 
 	if parentNode != node {
 		cfg["walletname"] = node
+	}
+
+	// doge fees are slightly higher than others. Leaving this as 0 will
+	// apply bitcoin limits.
+	if assetID == doge.BipID {
+		cfg["fallbackfee"] = fmt.Sprintf("%f", dexdoge.DefaultFee*1000/1e8)
+		cfg["feeratelimit"] = fmt.Sprintf("%f", dexdoge.DefaultFeeRateLimit*1000/1e8)
 	}
 
 	return &tWallet{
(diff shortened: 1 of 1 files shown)


client/core/simnet_trade.go
```diff
@@ -1857,6 +1870,15 @@ func (s *simulationTest) assertBalanceChanges(client *simulationClient) error {
 			quoteFees = int64(fees.Swap)
 			baseFees = int64(fees.Redemption)
 		}
+		// eth based assets will grossly overestimate swap fees here.
```
chappjc, line 1873: Why is this do you think?  Is it not giving fees using the current base fee, which we cannot go below?

Also, using too high fees was causing the balance failures?  Does that mean the expected balance change was computed based on different fee rates than these?

buck54321: For ethereum, the max fee rate is the only value we have when we set these fee values. This is in contrast to BTC, where the server reports the fee rate that will ultimately be used.

We **should** figure out how to update these or to populate them when we check our confirmations somehow.

JoeGruffins: > Why is this do you think?

I think we just decided to use the max here because it was easiest. There was a pr recently that made this fail I think but cannot find it atm... we were probably using the 6 blocks thing before.

200 doesn't seem too ridiculous... https://ycharts.com/indicators/ethereum_average_gas_price Even above that at times on mainnet.

Don't all coins lock enough fees for `ord.DEXConfig.MaxFeeRate` which is the max fee from the dex config file? I'm actually not sure atm... I mean when orders are placed, before matched.

JoeGruffins: Oh, these are the fees at match time.

JoeGruffins: Yeah, different because https://github.com/decred/dcrdex/blob/fb05984dc0559f91f0c9d09c3fa82460e5615fae/server/dex/feemgr.go#L111-L119

JoeGruffins: I guess the test would fail since #1447 as we were using `SuggestGasPrice` before.

JoeGruffins: I don't guess the server's FeeRate is ever used for eth?

chappjc: I recall https://github.com/decred/dcrdex/pull/1447 and I believe I may have suggested that change because with dynamic fees it's not a problem at all.  200 is ok, and even higher is fine as MaxFeeRate (and the match's prescribed fee rate) for the same reason.

@buck54321 stated the real issue in his comment.  When the dynamic eth txns are mined, they use a different fee than what was specified when it was authored and broadcasted.  i.e. with ETH, the prescribed swap fee is not what the transaction actually consumes in fees, so if we want to assert balance change we should base it on the _real_ fees.

Therefore, the actual bug is that `core.Order.FeesPaid` does not have the correct fees paid as it should.   If we run an ETH swap and look at the completed order in the order page, it'll show it paid higher fees than actual because of this.  As per Buck's comment, we need to update `trackedTrade.metaData.{SwapFeesPaid,RedemptionFeesPaid}` once we observe the txns are mined.

To make the test "work" for now I'm ok with this PR's change, but we'd want to open an issue to do the needed update.

chappjc: Computing the fees paid on an arbitrary txns can be quite tough actually.  For UTXO assets it often means fetching the prevout values of the inputs, so for whatever solution we reach, we should make it easy on ourselves and skip the fee computation and update if it's not `SupportsDynamicTxFee` because it's not needed.  Smells like another `WalletTrait`, like `DynamicTxFeeGetter`

JoeGruffins: look ok? #1741

chappjc: Thanks.  I started exploring the idea here https://github.com/decred/dcrdex/commit/0896ffd993856889ae42f7b7378439bcf09ab2bc
Anyone is welcome to pick it up.  I should focus one reviewing existing PRs.

chappjc (approved):

PR #1742, opened 2022-07-30
dcrdex/dex/wait: intermittent test failure

LasTshaMAN:
Stumbled upon this test failure on master, looks like `TestTaperingQueue` should be somewhat more resilient to go-scheduler.
```
--- FAIL: TestTaperingQueue (0.01s)
    queue_test.go:123: wrong result order expected [+1 +3 +4 +2 +0], got [+1 +3 +4 +0 +2]
FAIL
FAIL	decred.org/dcrdex/dex/wait	0.824s
```
With the changes from in this PR, problematic test still runs in under 2.5 seconds on CI, which I think is ok for flakiness trade-off (although it could be tweaked of course).

Closes https://github.com/decred/dcrdex/issues/1671.

### DIFF
--- a/dex/wait/queue.go
+++ b/dex/wait/queue.go
@@ -185,8 +185,6 @@ func (q *TaperingTickerQueue) Wait(waiter *Waiter) {
 
 // Run runs the primary wait loop until the context is canceled.
 func (q *TaperingTickerQueue) Run(ctx context.Context) {
-	taper := float64(q.slowestInterval - q.fastestInterval)
-
 	waiters := make([]*taperingWaiter, 0, 100)
 
 	var wg sync.WaitGroup
@@ -194,6 +192,7 @@ func (q *TaperingTickerQueue) Run(ctx context.Context) {
 
 	runWaiter := func(w *taperingWaiter) {
 		defer wg.Done()
+
 		if w.TryFunc() == DontTryAgain {
 			return
 		}
@@ -204,31 +203,27 @@ func (q *TaperingTickerQueue) Run(ctx context.Context) {
 			return
 		}
 
-		var nextTick time.Time
-		switch {
-		case w.tick <= fullSpeedTicks:
-			nextTick = time.Now().Add(q.fastestInterval)
-		case w.tick < fullyTapered: // ramp up the interval
-			prog := float64(w.tick-fullSpeedTicks) / (fullyTapered - fullSpeedTicks)
-			interval := q.fastestInterval + time.Duration(math.Round(prog*taper))
-			nextTick = time.Now().Add(interval)
-		default:
-			nextTick = time.Now().Add(q.slowestInterval)
-		}
-		if nextTick.After(w.Expiration) {
-			nextTick = w.Expiration
-		}
-
-		w.nextTick = nextTick
 		w.tick++
+		w.nextTick = nextTick(w.tick, q.slowestInterval, q.fastestInterval,
+			time.Now(), w.Expiration)
 
 		q.queueWaiter <- w // send it back to the queue
 	}
+	maybeStop := func(timer *time.Timer) {
+		if timer != nil {
+			timer.Stop()
+		}
+	}
 
+	var (
+		tick  <-chan time.Time
+		timer *time.Timer
+	)
 	for {
-		var tick <-chan time.Time
 		if len(waiters) > 0 {
-			tick = time.After(time.Until(waiters[0].nextTick))
+			maybeStop(timer)
+			timer = time.NewTimer(time.Until(waiters[0].nextTick))
+			tick = timer.C
 		}
 
 		select {
@@ -253,14 +248,37 @@ func (q *TaperingTickerQueue) Run(ctx context.Context) {
 			sort.Slice(waiters, func(i, j int) bool {
 				return waiters[i].nextTick.Before(waiters[j].nextTick) // ascending, next tick first
 			})
-			// NOTE: timer leaked until it fires - consider NewTimer and Stop here
 
 		case <-ctx.Done():
+			maybeStop(timer)
 			for _, w := range waiters {
 				w.ExpireFunc() // early, but still ending prior to DontTryAgain
 			}
 			return
 		}
 	}
+}
 
+func nextTick(
+	nextTick int,
+	slowestInterval, fastestInterval time.Duration,
+	now, expiration time.Time,
+) time.Time {
+	var nextTickTime time.Time
+	switch {
+	case nextTick < fullSpeedTicks:
+		nextTickTime = now.Add(fastestInterval)
+	case nextTick < fullyTapered: // ramp up the interval
+		prog := float64(nextTick+1-fullSpeedTicks) / (fullyTapered - fullSpeedTicks)
+		taper := float64(slowestInterval - fastestInterval)
+		interval := fastestInterval + time.Duration(math.Round(prog*taper))
+		nextTickTime = now.Add(interval)
+	default:
+		nextTickTime = now.Add(slowestInterval)
+	}
+
+	if nextTickTime.After(expiration) {
+		return expiration
+	}
+	return nextTickTime
 }
--- a/dex/wait/queue.go
+++ b/dex/wait/queue.go
@@ -185,8 +185,6 @@ func (q *TaperingTickerQueue) Wait(waiter *Waiter) {
 
 // Run runs the primary wait loop until the context is canceled.
 func (q *TaperingTickerQueue) Run(ctx context.Context) {
-	waiters := make([]*taperingWaiter, 0, 100)
-
 	var wg sync.WaitGroup
 	defer wg.Wait()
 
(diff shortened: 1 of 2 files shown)


dex/wait/queue.go
```diff
@@ -253,14 +257,15 @@ func (q *TaperingTickerQueue) Run(ctx context.Context) {
 			sort.Slice(waiters, func(i, j int) bool {
 				return waiters[i].nextTick.Before(waiters[j].nextTick) // ascending, next tick first
 			})
-			// NOTE: timer leaked until it fires - consider NewTimer and Stop here
```
LasTshaMAN, line 256: Attempting to address this while we are here (it's in a separate commit - could easily revert it or change).

chappjc:
Linking issue https://github.com/decred/dcrdex/issues/1671.

Previous proposed resolution: https://github.com/decred/dcrdex/pull/1681/files
The fundamental issue is that the test is designed to ensure the waiters run in the order in which they are added to the queue.  However that's simply not possible because of `go runWaiter(w) // note: async, so the next one could run its TryFunc first!`  The "little optimization" is another source of potential out of order execution.

I think the conclusion was we should change the test so it doesn't test this undocumented design goal, but I'm not sure.

dex/wait/queue.go
```diff
@@ -193,7 +193,9 @@ func (q *TaperingTickerQueue) Run(ctx context.Context) {
 	defer wg.Wait()
 
 	runWaiter := func(w *taperingWaiter) {
+		wg.Add(1)
```
chappjc, line 196: You always `Add` before starting the goroutine, not at the beginning of the goroutine you are starting.

LasTshaMAN:
> The fundamental issue is that the test is designed to ensure the **waiters run** in the order in which they are added to the queue.

From what I understood by looking at the relevant code, the fundamental issue is that the test is designed to ensure the **waiters terminate** in the order in which they are added to the queue.

However, when the difference between `Waiter`s is just 1 retry-cycle, go-scheduler can sometimes mix them up.

Thus, extending retry-cycle (the way I'm doing it is by increasing retry-attempts, however increasing waiting-interval is probably another fitting way to do it) should sufficiently split **termination** of each `Waiter` across time (so we can check exact **termination** order).

chappjc:
> Thus, extending retry-cycle (the way I'm doing it is by increasing retry-attempts x100 times per `Waiter`, however increasing waiting-interval is probably another fitting way to do it) should sufficiently split **termination** of each `Waiter` across time (so we can check exact **termination** order in very-likely-reliable manner).

I see that, and it does work.  However, the test is trying to enforce a property of the queue that does not really exists, or cannot be guaranteed.

One reason the go scheduler can "mix them up" is just because of how they are started:

```go
go thingA()
go thingB()
x = 1
```

In the above code there are no guarantees about what gets executed first, anything _inside_ either goroutine or the assignment to `x` in the parent goroutine.  It's simply out of our control. (BTW, this is also why you have to `wg.Add` _before_ launching the goroutine, because you don't know how long it's going to be before Go decides to run that statement.)

I acknowledge what you said, that the other problem is the execution inside the `runWaiter` and the termination because execution is free to switch between goroutines.

I guess I'm fine with just tweaking the test so it "always" passes in CI, but IMO it's implicitly describing a property that it cannot guarantee.  Really I think we need to define a better test for `TaperingTickerQueue`: something that tests what it is we want it to do.  I think we can write such a test with just a single waiter that we rig to signal each time it's called back to the test goroutine so we can observe that it starts **tapering**.  Maintaining order is not a documented property afaict.

dex/wait/queue.go
LasTshaMAN: Ah, right, thanks for calling this out, thought I'd simplify it but instead I got caught :)

Relevant doc section for `WaitGroup.Add`
```
// Note that calls with a positive delta that occur when the counter is zero
// must happen before a Wait. Calls with a negative delta, or calls with a
// positive delta that start when the counter is greater than zero, may happen
// at any time.
// Typically this means the calls to Add should execute before the statement
// creating the goroutine or other event to be waited for.
```

Fixed.

dex/wait/queue.go
```diff
@@ -253,14 +248,37 @@ func (q *TaperingTickerQueue) Run(ctx context.Context) {
...
+	nextTick int,
+	slowestInterval, fastestInterval time.Duration,
+	now, expiration time.Time,
+) time.Time {
+	var nextTickTime time.Time
+	switch {
+	case nextTick < fullSpeedTicks:
+		nextTickTime = now.Add(fastestInterval)
+	case nextTick < fullyTapered: // ramp up the interval
+		prog := float64(nextTick+1-fullSpeedTicks) / (fullyTapered - fullSpeedTicks)
```
LasTshaMAN, line 272: minor change that makes this behavior slightly more precise with respect to the comment on:
```go
    // fullSpeedTicks is the number of attempts that will be made with the
	// configured fastestInterval delay. After fullSpeedTicks, the retry speed
	// will be tapered off.
	fullSpeedTicks = 3
```
since `tick` starts at 0, not 1

LasTshaMAN (commented):
I've rewrote TestTaperingQueue to check **tapering** instead of **execution order**, even if to see how it can look like,

however I didn't get as good coverage as I expected from this (e.g. try changing some code in TaperingTickerQueue and see whether test breaks) ... maybe I picked sub-optimal approach for this test;

lets see what you think of it

dex/wait/queue.go
```diff
@@ -204,31 +203,27 @@ func (q *TaperingTickerQueue) Run(ctx context.Context) {
...
-			nextTick = time.Now().Add(q.slowestInterval)
-		}
-		if nextTick.After(w.Expiration) {
-			nextTick = w.Expiration
-		}
-
-		w.nextTick = nextTick
 		w.tick++
+		w.nextTick = nextTick(w.tick, q.slowestInterval, q.fastestInterval,
+			time.Now(), w.Expiration)
```
LasTshaMAN, line 208: pulled `nextTick` in a separate func to be able to unit-test it;

**note:** `w.tick++` now happens before next tick value is calculated; this is to make it fit with the 1-off change related to `fullSpeedTicks` I've mentioned above

dex/wait/queue_test.go
```diff
@@ -63,41 +64,45 @@ func TestTaper(t *testing.T) {
...
-					wg.Done()
+					wgWaiters.Done()
 					return DontTryAgain
 				}
 				return TryAgain
 			},
-			ExpireFunc: func() {},
+			// We don't expect expire func being called in this test, leaving it
+			// undefined so that we'll get a panic in case it gets called.
+			//ExpireFunc: func() {},
```
LasTshaMAN, line 96: maybe worth covering as well, maybe in a separate test ?
- on one hand, this test is already quite complex
- on the other, when testing concurrency its better to combine lots of different go-routines executing different code-paths

JoeGruffins:
If this is intended to close #1730 could you add that to the first githup comment (got the git commit)? https://docs.devart.com/studio-for-sql-server/source-controlling-databases/associating-commits-with-github-issues.html

LasTshaMAN:
> could you add that to the first githup comment (not the git commit)?

Added to PR description, 

I thought **chappjc** already linked it:
image

JoeGruffins (approved):
Looks ok to me.

Changes the test to look at individual tapering is working correctly, rather than the order in which they fire.

I think this is what was wanted.

JoeGruffins:
> I thought chappjc already linked it:

Link yeah but I don't think it closes unless you put that in the comment, if this is intended to close. We have issues left for very long periods of time after being solved just because everyone forgets. This ensures it is closed.

LasTshaMAN:
Closes https://github.com/decred/dcrdex/issues/1671 (to make sure this PR will close it, because I'm not sure whether "editing description" will count)

chappjc:
Editing the description totally counts, but I do appreciate the effort!  :smile: 

image

dex/wait/queue.go
LasTshaMAN: I think `maybeStop` will be called on next `for` iteration (since there will be at least 1 elem in `waiters`), so no need to call it here as well

dex/wait/queue.go
```diff
@@ -204,31 +203,27 @@ func (q *TaperingTickerQueue) Run(ctx context.Context) {
...
+			timer.Stop()
+		}
+	}
 
 	for {
-		var tick <-chan time.Time
+		var (
+			tick  <-chan time.Time
+			timer *time.Timer
+		)
```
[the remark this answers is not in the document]
LasTshaMAN: yeah, that was the intention, thanks for spotting this for me, fixed

dex/wait/queue_test.go
LasTshaMAN: Added simple check to `TestTaper` test.

dex/wait/queue.go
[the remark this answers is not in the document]
LasTshaMAN: I'm thinking, we might even optimise this even further via using single `Timer` with `Reset`,
although it's a tricky business, check for example this discussion https://github.com/cockroachdb/cockroach/pull/4367, they have implemented a wrapper for `time.Timer` in `util/timer.go` we could re-use, however (if we want to do it) I'm not sure how to pull this dependency in respecting their licence

dex/wait/queue_test.go
```diff
@@ -110,17 +121,108 @@ func TestTaperingQueue(t *testing.T) {
...
 }
+
+// expTickSchedule returns expected tick schedule with a certain startTime.
+func expTickSchedule(startTime time.Time, fastestInterval, slowestInterval time.Duration) []time.Time {
+	expectedTicks := [21]time.Time{} // 21 element should be enough for all our needs in these tests.
+	expectedTicks[0] = startTime
+	expectedTicks[1] = expectedTicks[0].Add(fastestInterval)
+	expectedTicks[2] = expectedTicks[1].Add(fastestInterval)
+	expectedTicks[3] = expectedTicks[2].Add(fastestInterval + time.Duration(math.Round((float64(1)/12)*float64(slowestInterval-fastestInterval))))
+	expectedTicks[4] = expectedTicks[3].Add(fastestInterval + time.Duration(math.Round((float64(2)/12)*float64(slowestInterval-fastestInterval))))
```
chappjc, line 210: So these 12s are fullyTapered (15) - fullSpeedTicks (3)?

LasTshaMAN: Yes, updated to show this explicitly.

chappjc (approved):

Issue #1734, opened 2022-07-26
sever/auth/auth_test.go: TestHandleResponse test failure


ukane-philemon:
This is the first time I'm seeing this. 
https://github.com/decred/dcrdex/runs/7524088585?check_suite_focus=true
```
2022-07-26 15:56:14.652 [DBG] AUTH_TEST: (*AuthManager).handleResponse: unknown msg ID 26
--- FAIL: TestHandleResponse (0.02s)
    auth_test.go:1296: expected 0 response handlers, found 1
2022-07-26 15:56:14.677 [ERR] AUTH_TEST: AccountInfo error for ID e94e8d95338b67a417650538b5250d3939d5eb81c4497677f4263ef0d4e45885: general failure
```

ukane-philemon:
```
client.logReq(newID, func(comms.Link, *msgjson.Message) {},
		0, func() { t.Log("expired (ok)") })
	time.Sleep(time.Millisecond) // expire Timer func run in goroutine
```
There's a `time.Sleep(time.Millisecond)` right above the fail line which should have ensured response handlers expire and are removed. Unfortunately, the whole clean-up process is done in a separate goroutine and maybe `time.Sleep(time.Millisecond)` is not always enough.

LasTshaMAN:
Also saw this fail, attaching [full test output](https://pastebin.com/KJ6y68EC) just in case it will become useful during debugging.

Will try looking into it, if nobody has started on this yet.

ukane-philemon:
> Unfortunately, the whole clean-up process is done in a separate goroutine and maybe time.Sleep(time.Millisecond) is not always enough.

Increasing `time.Sleep` a little more, maybe a few more Millisecond will allow enough time to ensure the clean-up is complete.

PR #1743, opened 2022-07-30
dcrdex/server/auth: TestHandleResponse is flaky

LasTshaMAN:
Closes https://github.com/decred/dcrdex/issues/1734.

### DIFF
--- a/server/auth/auth_test.go
+++ b/server/auth/auth_test.go
@@ -352,6 +352,19 @@ func makeEnsureErr(t *testing.T) func(rpcErr *msgjson.Error, tag string, code in
 	}
 }
 
+func waitFor(pred func() bool, timeout time.Duration) (fail bool) {
+	tStart := time.Now()
+	for {
+		if pred() {
+			return false
+		}
+		if time.Since(tStart) > timeout {
+			return true
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
+}
+
 var (
 	tCheckFeeAddr         = "DsaAKsMvZ6HrqhmbhLjV9qVbPkkzF5daowT"
 	tCheckFeeVal   uint64 = 500_000_000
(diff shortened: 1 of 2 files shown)


server/auth/auth_test.go
```diff
@@ -1290,21 +1290,34 @@ func TestHandleResponse(t *testing.T) {
 	newID := comms.NextID()
 	client.logReq(newID, func(comms.Link, *msgjson.Message) {},
 		0, func() { t.Log("expired (ok)") })
-	time.Sleep(time.Millisecond) // expire Timer func run in goroutine
-	client.mtx.Lock()
-	if len(client.respHandlers) != 0 {
-		t.Fatalf("expected 0 response handlers, found %d", len(client.respHandlers))
+	// Wait in a loop until response handler expires.
```
LasTshaMAN, line 1293: I think this is often a better approach to test cases like these (compared to just single sleep)

chappjc: I tend to agree.  I think one thing that would encourage taking this extra step if there was a simple helper function:

```go
func waitFor(pred func() bool, timeout time.Duration) (fail bool) {
	tStart := time.Now()
	for {
		if pred() {
			return false
		}
		if time.Since(tStart) > timeout {
			return true
		}
		time.Sleep(10 * time.Millisecond)
	}
}
```

Used in tests like like

```go
	if waitFor(func() bool {
		client.mtx.Lock()
		defer client.mtx.Unlock()
		return len(client.respHandlers) != 0
	}, 10*time.Second) {
		t.Fatal("no response handlers found")
	}
```

LasTshaMAN: yeah, I went for helper-func `retryUntilSuccess` in https://github.com/decred/dcrdex/pull/1737 to get rid of code-duplication,

can apply here as well, I guess (done! only seems like predicate should be inverted @chappjc).

chappjc (approved):

PR #1744, opened 2022-07-30
multi: exclude cancel matches for filled amount and percent on UI display

### LINKED ISSUE #1608: When order is manually canceled, the "Filled" field becomes 100%
Should not be 0%?

Screen Shot 2022-05-07 at 4 27 56 PM

ukane-philemon:
Exclude canceled matches when displaying filled amount and percent on the order page and orders table. Closes #1608
Screenshot 2022-07-30 at 6 31 02 PM
Screenshot 2022-07-30 at 6 30 24 PM

### DIFF
--- a/client/core/helpers.go
+++ b/client/core/helpers.go
@@ -140,29 +140,31 @@ func (ord *OrderReader) SettledPercent() string {
 	return ord.percent(settledFilter)
 }
 
-// FilledFrom is the sum filled in units of the outgoing asset.
+// FilledFrom is the sum filled in units of the outgoing asset. Excludes cancel
+// matches.
 func (ord *OrderReader) FilledFrom() string {
 	if ord.Sell {
-		return formatQty(ord.sumFrom(filledFilter), ord.BaseUnitInfo)
+		return formatQty(ord.sumFrom(filledNonCancelFilter), ord.BaseUnitInfo)
 	}
-	return formatQty(ord.sumFrom(filledFilter), ord.QuoteUnitInfo)
+	return formatQty(ord.sumFrom(filledNonCancelFilter), ord.QuoteUnitInfo)
 }
 
-// FilledTo is the sum filled in units of the incoming asset.
+// FilledTo is the sum filled in units of the incoming asset. Excludes cancel
+// matches.
 func (ord *OrderReader) FilledTo() string {
 	if ord.Sell {
-		return formatQty(ord.sumTo(filledFilter), ord.QuoteUnitInfo)
+		return formatQty(ord.sumTo(filledNonCancelFilter), ord.QuoteUnitInfo)
 	}
-	return formatQty(ord.sumTo(filledFilter), ord.BaseUnitInfo)
+	return formatQty(ord.sumTo(filledNonCancelFilter), ord.BaseUnitInfo)
 }
 
 // FilledPercent is the percent of the order that has filled, without percent
-// sign.
+// sign. Excludes cancel matches.
 func (ord *OrderReader) FilledPercent() string {
 	if ord.Type == order.CancelOrderType {
 		return ""
 	}
-	return ord.percent(filledFilter)
+	return ord.percent(filledNonCancelFilter)
 }
 
 // SideString is "sell" for sell orders, "buy" for buy orders, and "" for
@@ -370,8 +372,6 @@ func settlingFilter(match *Match) bool {
 	return (match.Side == order.Taker && match.Status < order.MatchComplete) ||
 		(match.Side == order.Maker && match.Status < order.MakerRedeemed)
 }
-
-func filledFilter(match *Match) bool { return true }
 
 func filledNonCancelFilter(match *Match) bool {
 	return !match.IsCancel
--- a/client/webserver/site/src/js/markets.ts
+++ b/client/webserver/site/src/js/markets.ts
@@ -1105,7 +1105,7 @@ export default class MarketsPage extends BasePage {
           ((OrderUtil.settled(a) * 100 / a.qty) - (OrderUtil.settled(b) * 100 / b.qty))
       case 'filled':
         return (a: Order, b: Order) => this.ordersSortDirection *
-          ((a.filled * 100 / a.qty) - (b.filled * 100 / b.qty))
+          ((OrderUtil.filled(a) * 100 / a.qty) - (OrderUtil.filled(b) * 100 / b.qty))
     }
   }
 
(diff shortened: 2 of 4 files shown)


chappjc:
Good stuff.  To confirm, if there's a two lot order with a single lot match and the order then gets cancelled, that would show as 50% filled now?  Looks like it based on a scan of the code.

ukane-philemon:
Yes, all cancelled matches are excluded from the UI display.

JoeGruffins (approved):

client/core/helpers.go
```diff
@@ -141,28 +141,40 @@ func (ord *OrderReader) SettledPercent() string {
 }
 
 // FilledFrom is the sum filled in units of the outgoing asset.
-func (ord *OrderReader) FilledFrom() string {
+func (ord *OrderReader) FilledFrom(excludeCancels bool) string {
+	filter := filledFilter
+	if excludeCancels {
+		filter = filledNonCancelFilter
```
chappjc, line 147: Interestingly, the fix to use `filledNonCancelFilter` was applied long ago in https://github.com/decred/dcrdex/pull/816/commits/890ab64885907a2416e780e2f1e2d91bd89d894b (https://github.com/decred/dcrdex/pull/816) and then reverted in an apparent accidental regression in https://github.com/decred/dcrdex/commit/08a72d1cd085af6a89dc535ab03184efa6083ad9

Looking at the history of this issue, I don't think `false` would ever be the correct arg for `FilledFrom` or `FilledTo`.  Can you comment on the need to ever include cancels in the computation, other than retaining the same value in the CSV output in `(*Core).deleteOrderFn`?

ukane-philemon: We also export orders using `(s *WebServer).handleExportOrders` and it's the same with  `(*Core).deleteOrderFn`. These are the only two places we'd want to include cancels to match internal behaviour.

buck54321:
Terminology is confused here. A match cannot be cancelled. This is a successful match **to a** cancel order. Looking around **core**, we usually call this a **cancel match**.

client/core/helpers.go
chappjc: Thanks.  In both of those cases it's for the `FilledPercent`, and that percent in the exported data should be just like everywhere else -- based only on trade matches not cancel matches.

ukane-philemon: > Unless I'm missing something else, the bool arg should be removed from all of these and the correct filter used consistently.

If this is what we want, then we shouldn't include cancel matches when setting `trade.Filled`?

chappjc: Yeah, I think that was a mistake.  But I think it's going to be hard to change now because for *internal* accounting purposes and other logic, we probably expect canceled orders to be 100% filled.  Again, it doesn't make sense to me, but lets just fix the display.

ukane-philemon: Alright.

chappjc: e.g. 

```go
func (t *Trade) Remaining() uint64 {
	t.fillAmtMtx.RLock()
	defer t.fillAmtMtx.RUnlock()
	return t.Quantity - t.FillAmt
}
```

If we redefine `FillAmt`, lots will break.  Maybe we could separately check a canceled bool, etc.  It's going to be a can of worms.

ukane-philemon: If it is worth it (which I don't think it is), you open an issue. I be glad to look into it.

client/core/helpers.go
```diff
@@ -140,29 +140,31 @@ func (ord *OrderReader) SettledPercent() string {
...
-// FilledFrom is the sum filled in units of the outgoing asset.
+// FilledFrom is the sum filled in units of the outgoing asset. Excludes cancel
+// matches.
 func (ord *OrderReader) FilledFrom() string {
 	if ord.Sell {
-		return formatQty(ord.sumFrom(filledFilter), ord.BaseUnitInfo)
+		return formatQty(ord.sumFrom(filledNonCancelFilter), ord.BaseUnitInfo)
 	}
-	return formatQty(ord.sumFrom(filledFilter), ord.QuoteUnitInfo)
+	return formatQty(ord.sumFrom(filledNonCancelFilter), ord.QuoteUnitInfo)
```
chappjc, line 149: Looks good.  Reapplies the fix in https://github.com/decred/dcrdex/commit/890ab64885907a2416e780e2f1e2d91bd89d894b that was accidentally reverted later.

client/core/core.go
```diff
@@ -3733,8 +3733,7 @@ func (c *Core) Orders(filter *OrderFilter) ([]*Order, error) {
 func (c *Core) coreOrderFromMetaOrder(mOrd *db.MetaOrder) (*Order, error) {
 	corder := coreOrderFromTrade(mOrd.Order, mOrd.MetaData)
 	oid := mOrd.Order.ID()
-	excludeCancels := false // maybe don't include cancel order matches?
-	matches, err := c.db.MatchesForOrder(oid, excludeCancels)
+	matches, err := c.db.MatchesForOrder(oid, true)
```
chappjc, line 3736: I think we need to keep this so that the cancel matches are show on the order details page. (Make an order, settle 50% and cancel the remaining -> the order details page should show both of those matches.  We have special handling for the "cancel match" added in https://github.com/decred/dcrdex/pull/816)

image

chappjc: I mean I guess we can get rid of the need for this special display if we all agree.  @buck54321 Referring to your fixes from https://github.com/decred/dcrdex/pull/816 being mostly restored in this PR.

chappjc: On second thought, excluding cancels in this method this would result in a breaking change on exported the `Core` API, and the `core.Order` generated by this method would be inconsistent with the `core.Order` generated by `(*trackedTrade).coreOrder`.

Let's just stick to the goal of fixing the filled amounts and percentages.

ukane-philemon: changes reverted.

chappjc:
Good:

image

image

image

chappjc:
The remaining amount on the confirm cancel form is correct, and the `statusString` is correct because `order.filled` is only used in `StatusExecuted` (not `StatusCanceled` or booked while cancelling).

However, the sort in the your orders table of the markets page looks wrong:

image

It looks like `ordersSortCompare` needs to use `OrderUtil.filled(a)` instead of `a.filled` in `case 'filled':`.  Same for `b` of course.

chappjc (approved):

buck54321 (approved):

Issue #1330, opened 2021-12-06
client/asset/btc: spurious TestTryBlocksWithNotifier failure


chappjc:
CI got an error with `TestTryBlocksWithNotifier` on a PR that didn't touch client/asset/btc.  It appears spurious, but needs investigation:

```
2021-12-06 18:07:47.809 [TRC] TEST: Fallback fees set at 100 Sats/vbyte
2021-12-06 18:07:47.809 [TRC] TEST: Fees rate limit set at 1400 sats/byte
2021-12-06 18:07:47.809 [TRC] TEST: Redeem conf target set to 2 blocks
2021-12-06 18:07:47.812 [TRC] TEST: Fallback fees set at 100 Sats/vbyte
2021-12-06 18:07:47.812 [TRC] TEST: Fees rate limit set at 1400 sats/byte
2021-12-06 18:07:47.812 [TRC] TEST: Redeem conf target set to 2 blocks
2021-12-06 18:07:48.179 [WRN] TEST: Reporting a block found in polling that the wallet apparently never reported: 1 bf9af9b3627ebd5f2a533b88aa9f3ca9f298769ce5664012cf2dd4854d8821d4. If you see this message repeatedly, it may indicate an issue with the wallet.
2021-12-06 18:07:48.179 [DBG] TEST: tip change: 0 (000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f) => 1 (bf9af9b3627ebd5f2a533b88aa9f3ca9f298769ce5664012cf2dd4854d8821d4)
2021-12-06 18:07:48.212 [WRN] TEST: Reporting a block found in polling that the wallet apparently never reported: 2 9e05825c69d16bd4586415ce44b093bfd07faad96389603ac7b8e569cf30c333. If you see this message repeatedly, it may indicate an issue with the wallet.
2021-12-06 18:07:48.213 [DBG] TEST: tip change: 1 (bf9af9b3627ebd5f2a533b88aa9f3ca9f298769ce5664012cf2dd4854d8821d4) => 2 (9e05825c69d16bd4586415ce44b093bfd07faad96389603ac7b8e569cf30c333)
2021-12-06 18:07:48.213 [DBG] TEST: tip change: 2 (9e05825c69d16bd4586415ce44b093bfd07faad96389603ac7b8e569cf30c333) => 3 (71bd9d48b67aeee050feee76fd4c5131241f4bd79e8bb4b11c4de8ff8a757538)
[...]

PR #1745, opened 2022-07-31
client/asset/btc: TestTryBlocksWithNotifier is flaky

LasTshaMAN:
Couldn't see any issues with the code being tested, however I think test waiting-intervals need to be tweaked (in master, `blockTicker / 2` is `2.5ms` in this test, which doesn't seem too generous for go-runtime to finish its business with handling go-routines / channels),

for tests relying on go-scheduler to "work in time" (like this one), I think, it's better to over-allocate waiting times (than under-allocate these), even at the cost of long-running tests (cause flakieness will distract devs); although, it's a trade-off ...

Maybe worth closing https://github.com/decred/dcrdex/issues/1330 and re-open it if  the issue reproduces with higher waiting-intervals.

### DIFF
--- a/client/asset/btc/spv_test.go
+++ b/client/asset/btc/spv_test.go
@@ -755,8 +755,8 @@ func TestTryBlocksWithNotifier(t *testing.T) {
 	defaultWalletBlockAllowance := walletBlockAllowance
 	defaultBlockTicker := blockTicker
 
-	walletBlockAllowance = 300 * time.Millisecond
-	blockTicker = 50 * time.Millisecond
+	walletBlockAllowance = 50 * time.Millisecond
+	blockTicker = 20 * time.Millisecond
 
 	defer func() {
 		walletBlockAllowance = defaultWalletBlockAllowance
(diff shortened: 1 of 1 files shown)


chappjc:
> better to over-allocate waiting times (than under-allocate these), even at the cost of long-running tests

In this case, nACK.  The test goes from 0.47sec to 4.7 sec.

I think the long run time is mainly because for two of the tests the `walletBlockAllowance` gets multiplied by 10 in the `watchBlock` method (see the `else if syncStatus.Syncing` block`).  I suspect any flakiness is primarily coming from the 5ms `blockTicker`, so maybe we that to 20ms and only change `walletBlockAllowance` to 50ms.

LasTshaMAN:
> I suspect any flakiness is primarily coming from the 5ms blockTicker, so maybe we change that to 20ms and only change walletBlockAllowance to 50ms.

I guess that should be enough, changed.

chappjc:
I'll re-run CI a few times to see.
But I don't see these errors often TBH.

chappjc (approved):
LGTM.  Not seeing any failures, although they were rare in the first place, and the duration of `TestTryBlocksWithNotifier` is still under a second.

PR #1746, opened 2022-07-31
server/apidata: fix 24-hour calcs

buck54321:
```
Fixes the calculation of 24 hour volume and price change. I was using
the same cache for every market before. duh.
```

### DIFF
--- a/server/apidata/apidata.go
+++ b/server/apidata/apidata.go
@@ -55,7 +55,6 @@ type DataAPI struct {
 
 	cacheMtx     sync.RWMutex
 	marketCaches map[string]map[uint64]*candles.Cache
-	cache5min    *candles.Cache
 }
 
 // NewDataAPI is the constructor for a new DataAPI.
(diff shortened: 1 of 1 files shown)


server/apidata/apidata.go
```diff
@@ -118,17 +111,19 @@ func (s *DataAPI) ReportEpoch(base, quote uint32, epochIdx uint64, stats *matche
...
+		defer s.cacheMtx.Unlock()
+		mktCaches := s.marketCaches[mktName]
+		if mktCaches == nil {
+			return 0, 0, fmt.Errorf("unknown market %q", mktName)
+		}
+		epochDur := s.epochDurations[mktName]
+		startStamp := epochIdx * epochDur
+		endStamp := startStamp + epochDur
+		var cache5min *candles.Cache
+		const fiveMins = uint64(time.Minute * 5 / time.Millisecond)
```
chappjc, line 125: Not sure if this is clearer than `(5*time.Minute).Milliseconds()` or just `5 * 60 * 1000`, but ok

chappjc (approved):

PR #1751, opened 2022-08-02
client/core: throttle with Ticker instead of Timer

LasTshaMAN:
Noticed this while exploring `time.Timer` quirks [here](https://github.com/decred/dcrdex/pull/1742#discussion_r935345226).

Looks like we need a `Ticker` (not `Timer`), otherwise we'll throttle `refreshFiatRates` only once ?

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -7743,10 +7743,9 @@ func (c *Core) fetchFiatExchangeRates() {
 	c.wg.Add(1)
 	go func() {
 		defer c.wg.Done()
-		tick := time.NewTimer(fiatRateRequestInterval)
+		tick := time.NewTicker(fiatRateRequestInterval)
 		defer tick.Stop()
 		for {
-
 			c.refreshFiatRates(ctx)
 
 			select {


chappjc (approved):
Good catch

client/core/core.go
```diff
@@ -7743,10 +7743,9 @@ func (c *Core) fetchFiatExchangeRates() {
 	c.wg.Add(1)
 	go func() {
 		defer c.wg.Done()
-		tick := time.NewTimer(fiatRateRequestInterval)
+		tick := time.NewTicker(fiatRateRequestInterval)
 		defer tick.Stop()
 		for {
-
 			c.refreshFiatRates(ctx)
```
buck54321, line 7750: Generally, I think it's better to create a new timer on every loop, so that it is X time **between the end** of one call and the start of the other. This prevents any chance that calls can pile up. In this case, since `fiatRequestInterval` is much greater than a 4 second timeout internally in `refreshFiatRates`, I don't think we'll have any problems with this. 

I actually thought that @ukane-philemon was creating a new timer every loop here. Wasn't there a method calculated the next tick based on the current state somehow? Maybe I'm confused.

Anyway, this is fine, I guess, Just wanted to comment on the different patterns.

ukane-philemon:
My bad. I was creating a new timer after every loop but removed that method in subsequent reviews.

client/core/core.go
chappjc: Yeah, this was just how I thought it should be (consistent time between requests, not time between end of one call and start of another).  But I'd be fine with a new timer on each iteration if you guys really think it makes a difference.  This is just simple and regular request intervals.

chappjc:
> My bad. I was creating a new timer after every loop but removed that method in subsequent reviews.

No, nothing is wrong.

client/core/core.go
chappjc: For ref: https://github.com/decred/dcrdex/pull/1600#discussion_r921498759
`nextRateSourceTick` added nothing but complexity so we removed it.  It's still straightforward to make a new *timer* in the loop though.  No need for a helper function because all the last request recording for the different sources was simplified too.

buck54321: As long as that 4 second timeout is effective, it's fine as is.

chappjc: That's not how a ticker works though.  It won't spew out multiple signals if no one was listening for a while.  The internal channel sent on a ticker (or timer) is non-blocking.

chappjc: See https://pkg.go.dev/time#NewTicker

> The ticker will adjust the time interval or drop ticks to make up for slow receivers

It doesn't matter if the timeout internal to `refreshFiatRates` is effective.

PR #1752, opened 2022-08-03
trade: Update dynamic fees after they are mined.

JoeGruffins:
Fees for some trades cannot be accurately calculated until after the tx
    is mined. For those assets, start a go routine to check on the txn until
    it can be found and the actual gas used be added to the swap and
    redemption fees paid.

~depends on #1739~

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -16,6 +16,7 @@ import (
 	"math"
 	"math/big"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -71,7 +72,8 @@ const (
 	// confCheckTimeout is the amount of time allowed to check for
 	// confirmations. Testing on testnet has shown spikes up to 2.5
 	// seconds. This value may need to be adjusted in the future.
-	confCheckTimeout = 4 * time.Second
+	confCheckTimeout                 = 4 * time.Second
+	dynamicSwapOrRedemptionFeesConfs = 2
 )
 
 var (
@@ -229,6 +231,7 @@ type ethFetcher interface {
 	connect(ctx context.Context) error
 	peerCount() uint32
 	contractBackend() bind.ContractBackend
+	headerByHash(txHash common.Hash) *types.Header
 	lock() error
 	locked() bool
 	pendingTransactions() ([]*types.Transaction, error)
@@ -243,7 +246,7 @@ type ethFetcher interface {
 	currentFees(ctx context.Context) (baseFees, tipCap *big.Int, err error)
 	unlock(pw string) error
 	getConfirmedNonce(context.Context, int64) (uint64, error)
-	transactionReceipt(context.Context, common.Hash) (*types.Receipt, error)
+	transactionReceipt(ctx context.Context, txHash common.Hash) (*types.Receipt, *types.Transaction, error)
 }
 
 // monitoredTx is used to keep track of redemption transactions that have not
@@ -330,6 +333,8 @@ var _ asset.LiveReconfigurer = (*ETHWallet)(nil)
 var _ asset.LiveReconfigurer = (*TokenWallet)(nil)
 var _ asset.TxFeeEstimator = (*ETHWallet)(nil)
 var _ asset.TxFeeEstimator = (*TokenWallet)(nil)
+var _ asset.DynamicSwapOrRedemptionFeeChecker = (*ETHWallet)(nil)
+var _ asset.DynamicSwapOrRedemptionFeeChecker = (*TokenWallet)(nil)
 
 type baseWallet struct {
 	ctx  context.Context // the asset subsystem starts with Connect(ctx)
@@ -2542,6 +2547,92 @@ func (eth *baseWallet) SyncStatus() (bool, float32, error) {
 	return eth.node.peerCount() > 0, 1.0, nil
 }
 
+// DynamicSwapFeesPaid returns fees for initiation transactions. Part of the
+// asset.DynamicSwapOrRedemptionFeeChecker interface.
+func (eth *assetWallet) DynamicSwapFeesPaid(ctx context.Context, coinID, contractData dex.Bytes) (fee uint64, secretHashes [][]byte, err error) {
+	return eth.swapOrRedemptionFeesPaid(ctx, coinID, contractData, true)
+}
+
+// DynamicRedemptionFeesPaid returns fees for redemption transactions. Part of
+// the asset.DynamicSwapOrRedemptionFeeChecker interface.
+func (eth *assetWallet) DynamicRedemptionFeesPaid(ctx context.Context, coinID, contractData dex.Bytes) (fee uint64, secretHashes [][]byte, err error) {
+	return eth.swapOrRedemptionFeesPaid(ctx, coinID, contractData, false)
+}
+
+// swapOrRedemptionFeesPaid returns exactly how much gwei was used to send an
+// initiation or redemption transaction. It also returns the secret hashes
+// included with this init or redeem. Secret hashes are sorted so returns are
+// always the same, but the order may not be the same as they exist in the
+// transaction on chain. The transaction must be already mined for this
+// function to work. Returns asset.CoinNotFoundError for unmined txn. Returns
+// asset.ErrNotEnoughConfirms for txn with too few confirmations. Will also
+// error if the secret hash in the contractData is not found in the transaction
+// secret hashes.
+func (eth *baseWallet) swapOrRedemptionFeesPaid(ctx context.Context, coinID, contractData dex.Bytes,
+	isInit bool) (fee uint64, secretHashes [][]byte, err error) {
+	contractVer, secretHash, err := dexeth.DecodeContractData(contractData)
+	if err != nil {
+		return 0, nil, err
+	}
+	var txHash common.Hash
+	copy(txHash[:], coinID)
+	receipt, tx, err := eth.node.transactionReceipt(ctx, txHash)
+	if err != nil {
+		return 0, nil, err
+	}
+
+	hdr := eth.node.headerByHash(receipt.BlockHash)
+	if hdr == nil {
+		return 0, nil, fmt.Errorf("header for hash %v not found", receipt.BlockHash)
+	}
+
+	bestHdr, err := eth.node.bestHeader(ctx)
+	if err != nil {
+		return 0, nil, err
+	}
+
+	confs := bestHdr.Number.Int64() - hdr.Number.Int64() + 1
+	if confs < dynamicSwapOrRedemptionFeesConfs {
+		return 0, nil, asset.ErrNotEnoughConfirms
+	}
+
+	effectiveGasPrice := new(big.Int).Add(hdr.BaseFee, tx.EffectiveGasTipValue(hdr.BaseFee))
+	bigFees := new(big.Int).Mul(effectiveGasPrice, big.NewInt(int64(receipt.GasUsed)))
+	if isInit {
+		inits, err := dexeth.ParseInitiateData(tx.Data(), contractVer)
+		if err != nil {
+			return 0, nil, fmt.Errorf("invalid initiate data: %v", err)
+		}
+		secretHashes = make([][]byte, 0, len(inits))
+		for k := range inits {
+			copyK := k
+			secretHashes = append(secretHashes, copyK[:])
+		}
+	} else {
+		redeems, err := dexeth.ParseRedeemData(tx.Data(), contractVer)
+		if err != nil {
+			return 0, nil, fmt.Errorf("invalid redeem data: %v", err)
+		}
+		secretHashes = make([][]byte, 0, len(redeems))
+		for k := range redeems {
+			copyK := k
+			secretHashes = append(secretHashes, copyK[:])
+		}
+	}
+	sort.Slice(secretHashes, func(i, j int) bool { return bytes.Compare(secretHashes[i], secretHashes[j]) < 0 })
+	var found bool
+	for i := range secretHashes {
+		if bytes.Equal(secretHash[:], secretHashes[i]) {
+			found = true
+			break
+		}
+	}
+	if !found {
+		return 0, nil, fmt.Errorf("secret hash %x not found in transaction", secretHash)
+	}
+	return dexeth.WeiToGwei(bigFees), secretHashes, nil
+}
+
 // RegFeeConfirmations gets the number of confirmations for the specified
 // transaction.
 func (eth *baseWallet) RegFeeConfirmations(ctx context.Context, coinID dex.Bytes) (confs uint32, err error) {
@@ -2553,7 +2644,7 @@ func (eth *baseWallet) RegFeeConfirmations(ctx context.Context, coinID dex.Bytes
 // recommendedMaxFeeRate finds a recommended max fee rate using the somewhat
 // standard baseRate * 2 + tip formula.
 func (eth *baseWallet) recommendedMaxFeeRate(ctx context.Context) (*big.Int, error) {
-	base, tip, err := eth.node.currentFees(eth.ctx)
+	base, tip, err := eth.node.currentFees(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("Error getting net fee state: %v", err)
 	}
@@ -3018,7 +3109,7 @@ func (w *assetWallet) confirmRedemptionWithoutMonitoredTx(txHash common.Hash, re
 		}
 	}
 	if confirmations >= txConfsNeededToConfirm {
-		receipt, err := w.node.transactionReceipt(w.ctx, txHash)
+		receipt, _, err := w.node.transactionReceipt(w.ctx, txHash)
 		if err != nil {
 			return nil, err
 		}
@@ -3113,7 +3204,7 @@ func (w *assetWallet) confirmRedemption(coinID dex.Bytes, redemption *asset.Rede
 		}
 	}
 	if confirmations >= txConfsNeededToConfirm {
-		receipt, err := w.node.transactionReceipt(w.ctx, txHash)
+		receipt, _, err := w.node.transactionReceipt(w.ctx, txHash)
 		if err != nil {
 			return nil, err
 		}
@@ -3573,7 +3664,7 @@ func GetGasEstimates(ctx context.Context, cl *nodeClient, c contractor, maxSwaps
 			return fmt.Errorf("initiate error for %d swaps: %v", n, err)
 		}
 		waitForMined()
-		receipt, err := cl.transactionReceipt(ctx, tx.Hash())
+		receipt, _, err := cl.transactionReceipt(ctx, tx.Hash())
 		if err != nil {
 			return err
 		}
@@ -3583,7 +3674,7 @@ func GetGasEstimates(ctx context.Context, cl *nodeClient, c contractor, maxSwaps
 		stats.swaps = append(stats.swaps, receipt.GasUsed)
 
 		if isToken {
-			receipt, err = cl.transactionReceipt(ctx, approveTx.Hash())
+			receipt, _, err = cl.transactionReceipt(ctx, approveTx.Hash())
 			if err != nil {
 				return err
 			}
@@ -3591,7 +3682,7 @@ func GetGasEstimates(ctx context.Context, cl *nodeClient, c contractor, maxSwaps
 				return err
 			}
 			stats.approves = append(stats.approves, receipt.GasUsed)
-			receipt, err = cl.transactionReceipt(ctx, transferTx.Hash())
+			receipt, _, err = cl.transactionReceipt(ctx, transferTx.Hash())
 			if err != nil {
 				return err
 			}
@@ -3626,7 +3717,7 @@ func GetGasEstimates(ctx context.Context, cl *nodeClient, c contractor, maxSwaps
 			return fmt.Errorf("redeem error for %d swaps: %v", n, err)
 		}
 		waitForMined()
-		receipt, err = cl.transactionReceipt(ctx, tx.Hash())
+		receipt, _, err = cl.transactionReceipt(ctx, tx.Hash())
 		if err != nil {
 			return err
 		}
--- a/client/asset/eth/nodeclient.go
+++ b/client/asset/eth/nodeclient.go
@@ -134,6 +134,10 @@ func (n *nodeClient) bestHeader(ctx context.Context) (*types.Header, error) {
 	return n.leth.ApiBackend.HeaderByNumber(ctx, rpc.LatestBlockNumber)
 }
 
+func (n *nodeClient) headerByHash(txHash common.Hash) *types.Header {
+	return n.leth.BlockChain().GetHeaderByHash(txHash)
+}
+
 func (n *nodeClient) stateAt(ctx context.Context, bn rpc.BlockNumber) (*state.StateDB, error) {
 	state, _, err := n.leth.ApiBackend.StateAndHeaderByNumberOrHash(ctx, rpc.BlockNumberOrHashWithNumber(bn))
 	if err != nil {
@@ -174,29 +178,29 @@ func (n *nodeClient) locked() bool {
 }
 
 // transactionReceipt retrieves the transaction's receipt.
-func (n *nodeClient) transactionReceipt(ctx context.Context, txHash common.Hash) (*types.Receipt, error) {
+func (n *nodeClient) transactionReceipt(ctx context.Context, txHash common.Hash) (*types.Receipt, *types.Transaction, error) {
 	tx, blockHash, _, index, err := n.leth.ApiBackend.GetTransaction(ctx, txHash)
 	if err != nil {
 		if errors.Is(err, ethereum.NotFound) {
-			return nil, asset.CoinNotFoundError
+			return nil, nil, asset.CoinNotFoundError
 		}
-		return nil, err
+		return nil, nil, err
 	}
 	if tx == nil {
-		return nil, fmt.Errorf("%w: transaction %v not found", asset.CoinNotFoundError, txHash)
+		return nil, nil, fmt.Errorf("%w: transaction %v not found", asset.CoinNotFoundError, txHash)
 	}
 	receipts, err := n.leth.ApiBackend.GetReceipts(ctx, blockHash)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	if len(receipts) <= int(index) {
-		return nil, fmt.Errorf("no receipt at index %d in block %s for tx %s", index, blockHash, txHash)
+		return nil, nil, fmt.Errorf("no receipt at index %d in block %s for tx %s", index, blockHash, txHash)
 	}
 	receipt := receipts[index]
 	if receipt == nil {
-		return nil, fmt.Errorf("nil receipt at index %d in block %s for tx %s", index, blockHash, txHash)
+		return nil, nil, fmt.Errorf("nil receipt at index %d in block %s for tx %s", index, blockHash, txHash)
 	}
-	return receipt, nil
+	return receipt, tx, nil
 }
 
 // pendingTransactions returns pending transactions.
@@ -358,6 +362,9 @@ func (n *nodeClient) getTransaction(ctx context.Context, txHash common.Hash) (*t
 	}
 	tx, _, blockHeight, _, err := n.leth.ApiBackend.GetTransaction(ctx, txHash)
 	if err != nil {
+		if errors.Is(err, ethereum.NotFound) {
+			return nil, 0, asset.CoinNotFoundError
+		}
 		return nil, 0, err
 	}
 	if tx != nil {
--- a/client/asset/interface.go
+++ b/client/asset/interface.go
@@ -143,6 +143,9 @@ const (
 	// ErrSwapRefunded is returned from ConfirmRedemption when the swap has
 	// been refunded before the user could redeem.
 	ErrSwapRefunded = dex.ErrorKind("swap refunded")
+	// ErrNotEnoughConfirms is returned when a transaction is confirmed,
+	// but does not have enough confirmations to be trusted.
+	ErrNotEnoughConfirms = dex.ErrorKind("transaction does not have enough confirmations")
 
 	// InternalNodeLoggerName is the name for a logger that is used to fine
 	// tune log levels for only loggers using this name.
@@ -455,6 +458,24 @@ type NewAddresser interface {
 // LogFiler is a wallet that allows for downloading of its log file.
 type LogFiler interface {
 	LogFilePath() string
+}
+
+// DynamicSwapOrRedemptionFeeChecker defines methods that accept an initiation
+// or redemption coinID and returns the fee spent on the transaction along with
+// the secrets included in the tx. Returns asset.CoinNotFoundError for unmined
+// txn. Returns asset.ErrNotEnoughConfirms for txn with too few confirmations.
+// Will also error if the secret in the contractData is not found in the
+// transaction secrets.
+type DynamicSwapOrRedemptionFeeChecker interface {
+	// RedemptionConfirmer methods are not currently used. However, a
+	// RedemptionConfirmer adds the necessity to wait for redemption
+	// confirmations that the DynamicSwapOrRedemptionFeeChecker depends on
+	// and will not work properly without them.
+	RedemptionConfirmer
+	// DynamicSwapFeesPaid returns fees for initiation transactions.
+	DynamicSwapFeesPaid(ctx context.Context, coinID, contractData dex.Bytes) (fee uint64, secretHashes [][]byte, err error)
+	// DynamicRedemptionFeesPaid returns fees for redemption transactions.
+	DynamicRedemptionFeesPaid(ctx context.Context, coinID, contractData dex.Bytes) (fee uint64, secretHashes [][]byte, err error)
 }
 
 // FeeRater is capable of retrieving a non-critical fee rate estimate for an
--- a/client/core/trade.go
+++ b/client/core/trade.go
@@ -1286,6 +1286,134 @@ func (t *trackedTrade) isSwappable(ctx context.Context, match *matchTracker) (re
 	return false, false
 }
 
+// checkSwapFeeConfirms returns whether the swap fee confirmations should be
+// checked.
+//
+// This method accesses match fields and MUST be called with the trackedTrade
+// mutex lock held for reads.
+func (t *trackedTrade) checkSwapFeeConfirms(match *matchTracker) bool {
+	if match.MetaData.Proof.SwapFeeConfirmed {
+		return false
+	}
+	_, dynamic := t.wallets.fromWallet.Wallet.(asset.DynamicSwapOrRedemptionFeeChecker)
+	if !dynamic {
+		// Confirmed will be set in the db.
+		return true
+	}
+	if match.Side == order.Maker {
+		return match.Status >= order.MakerSwapCast
+	}
+	return match.Status >= order.TakerSwapCast
+}
+
+// checkRedemptionFeeConfirms returns whether the swap fee confirmations should
+// be checked.
+//
+// This method accesses match fields and MUST be called with the trackedTrade
+// mutex lock held for reads.
+func (t *trackedTrade) checkRedemptionFeeConfirms(match *matchTracker) bool {
+	if match.MetaData.Proof.RedemptionFeeConfirmed {
+		return false
+	}
+	_, dynamic := t.wallets.toWallet.Wallet.(asset.DynamicSwapOrRedemptionFeeChecker)
+	if !dynamic {
+		// Confirmed will be set in the db.
+		return true
+	}
+	if match.Side == order.Maker {
+		return match.Status >= order.MakerRedeemed
+	}
+	return match.Status >= order.MatchComplete
+}
+
+// updateDynamicSwapOrRedemptionFeesPaid updates the fees used for dynamic fee
+// checker transactions. We do not know the exact fees the tx will use until
+// they are mined, so this waits until they are mined and updates the value for
+// the entire trade.
+//
+// NOTE: As long as init and redemption confirms add up to more than two this
+// method will fire as expected before the swap is determined in Confirmed
+// status. The asset.DynamicSwapOrRedemptionFeeChecker interface requires the
+// asset.RedemptionConfirmer interface to ensure this for redemptions.
+func (t *trackedTrade) updateDynamicSwapOrRedemptionFeesPaid(ctx context.Context, match *matchTracker, isInit bool) {
+	wallet := t.wallets.fromWallet
+	if !isInit {
+		wallet = t.wallets.toWallet
+	}
+	stopChecks := func() {
+		if isInit {
+			match.MetaData.Proof.SwapFeeConfirmed = true
+		} else {
+			match.MetaData.Proof.RedemptionFeeConfirmed = true
+		}
+		err := t.db.UpdateOrderMetaData(t.ID(), t.metaData)
+		if err != nil {
+			t.dc.log.Errorf("Error updating order metadata for order %s: %v", t.ID(), err)
+		}
+	}
+	feeChecker, dynamic := wallet.Wallet.(asset.DynamicSwapOrRedemptionFeeChecker)
+	if !dynamic {
+		stopChecks()
+		return
+	}
+	txType := "swap"
+	if !isInit {
+		txType = "redemption"
+	}
+	var coinID, contractData []byte
+	// Check if a swap or redeem coin id has been populated in the
+	// match tracker. If it has we ask the wallet for the fees paid
+	// and add that to either the total swap or redeem fees for the
+	// trade.
+	if isInit {
+		coinID = []byte(match.MetaData.Proof.MakerSwap)
+		if match.Side != order.Maker {
+			coinID = []byte(match.MetaData.Proof.TakerSwap)
+		}
+		contractData = match.MetaData.Proof.ContractData
+	} else {
+		coinID = []byte(match.MetaData.Proof.MakerRedeem)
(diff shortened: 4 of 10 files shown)


JoeGruffins:
Will squash the last two commits later I guess. Also will add a test.

@chappjc what do you think about these changes?

It's still just going from zero fees to actual. If we wanted to put in a better estimate at first, and also if we wanted to put in the correct values if the client was shut down before it found the actual values, maybe we need to save the amounts per match. Otherwise we don't know which was checked/updated and which not.

chappjc:
> Will squash the last two commits later I guess. Also will add tests.
> 
> @chappjc what do you think about these changes?

Cool, I like, based on a quick scan.

> It's still just going from zero fees to actual. If we wanted to put in a better estimate at first, and also if we wanted to put in the correct values if the client was shut down before it found the actual values, maybe we need to save the amounts per match. Otherwise we don't know which was checked/updated and which not.

I don't feel as strongly about this.  Its only while unconfirmed so should be brief. Besides, 0 or max are both incorrect -- it's simply not known until mined.
But it looks like we could pass the reserved/max fees into `updateFeesPaid` when it gets launched so that it can `-=` the difference once it figures out actual fees paid.

JoeGruffins:
> But it looks like we could pass the reserved/max fees into updateFeesPaid when it gets launched so that it can -= the difference once it figures out actual fees paid.

Ah, very obvious solution. Did not even think of that. Anyway, will rebase after #1739 gets in.

client/asset/eth/nodeclient.go
```diff
@@ -174,29 +174,39 @@ func (n *nodeClient) locked() bool {
...
+}
+
+func (n *nodeClient) transactionFeesPaid(ctx context.Context, txHash common.Hash) (*big.Int, error) {
+	receipt, tx, err := n.transactionReceipt(ctx, txHash)
+	if err != nil {
+		return nil, err
 	}
-	return receipt, nil
+	hdr := n.leth.BlockChain().GetHeaderByHash(receipt.BlockHash) // n.leth.ApiBackend.HeaderByHash with no error
+	effectiveGasPrice := tx.EffectiveGasTipValue(hdr.BaseFee)
```
chappjc, line 208: @buck54321 pointed this out, but I think what I suggested was incomplete.  What I was basing this on was https://github.com/ethereum/go-ethereum/blob/aa69d36152becdc819f323b5ae05e249d984d609/internal/ethapi/api.go#L1572-L1573
`gasPrice := new(big.Int).Add(header.BaseFee, tx.EffectiveGasTipValue(header.BaseFee))`

JoeGruffins:
Just rebased.

client/asset/eth/nodeclient.go
JoeGruffins: Ok. I also took out a check in this function for the tx status. I think that some gas may be used regardless if it is mined so couldn't think of a good reason to fail if status is not OK. Maybe it was just a copy pasta or if there was a reason can add back.

buck54321:
I was kind of wondering if we might add a `feeUpdate uint64` return value to `(asset.Wallet).SwapConfirmations`.

chappjc:
> I was kind of wondering if we might add a `feeUpdate uint64` return value to `(asset.Wallet).SwapConfirmations`.

The thing about that is it can be quite inconvenient to get the tx fees paid for UTXO assets (input values not available without pulling prevout values), where it's not necessary for the fee update anyway.  That's why the optional interface was convenient.

chappjc:
However, if the goals is to get rid of the ongoing goroutine, it certainly could just get the actual fee at the same place where swap confs are checked (isSwappable, etc) and update on the first conf.

buck54321:
> > I was kind of wondering if we might add a `feeUpdate uint64` return value to `(asset.Wallet).SwapConfirmations`.
> 
> The thing about that is it can be quite inconvenient to get the tx fees paid for UTXO assets (input values not available without pulling prevout values), where it's not necessary for the fee update anyway. That's why the optional interface was convenient.

In the changes I recommended for #1638, I had a zero value signaling no update.

chappjc:
> In the changes I recommended for #1638, I had a zero value signaling no update

Can you clarify, because I see that suggestion on the new `ConfirmRedemption` method, not `SwapConfirmations`.

However, I don't think I would be comfortable with an output from `SwapConfirmations` that is just not used on the UTXO assets.  It would be awkward and confusing to the implementors ("ok so I can just return 0 or I need to compute fees"... checks interface docs... "ok just return zero")

buck54321:
I though I could edit faster. 

> However, I don't think I would be comfortable with an output from `SwapConfirmations` that is just not used on the UTXO assets.

Understandable. I was on the fence too. Let's just roll with the interface.

JoeGruffins:
> However, if the goal is to get rid of the ongoing goroutine, it certainly could just get the actual fee at the same place where swap confs are checked (isSwappable, etc) and update on the first conf.
> 
> i.e.
> 
> ```go
> 	lastConfs := match.setSwapConfirms(int64(confs))
> 	justMined := lastConfs < 1 && confs > 0
> 
> 	if feeChecker, ok := wallet.Wallet.(asset.DynamicTxFeeChecker); ok && justMined {
> 		actualSwapFees, err := feeChecker.TransactionFeesPaid(ctx, []byte(match.MetaData.Proof.MakerSwap))
> 
> ... <then update t.metaData.SwapFeesPaid later when the mutex situation permits>
> ```

Isn't this just for maker inits? Does this also work for taker inits and redemptions?

chappjc:
> Isn't this just for maker inits? Does this also work for taker inits and redemptions? I think maker and init is the only time we current wait for our own swap to have confs. Otherwise we are only waiting for the counterparty. No?

True, so it would need similar code in the other code like isRedeemable and elsewhere.  I'm personally not opposed to the goroutine to do this in part because of that.  But also it eases lock contention by keeping the work done under mutex lock in `tick` less.

JoeGruffins:
> But also it eases lock contention by reducing the work done under mutex lock in tick.

The goroutine?

Maybe I don't quite get how you would implement in isSwappable/isRedeemable, but seems like you would need to change quite a bit for non-eth coins as well. Do you want me to move the checks there and remove the goroutine?

chappjc:
> Do you want me to move the checks there and remove the goroutine?

No no.  Just conversing about it.  I'll dig in here shortly, but it sits fine with me.

client/core/trade.go
```diff
@@ -1526,6 +1628,7 @@ func (c *Core) tick(t *trackedTrade) (assetMap, error) {
 			swaps = append(swaps, match)
 			sent += match.Quantity
 			quoteSent += calc.BaseToQuote(match.Rate, match.Quantity)
+			t.updateFeesPaid(c.ctx, match, true)
```
martonp, line 1631: Redeems are batched, so multiple matches may use the same transaction. This needs to be checked per CoinID instead of per match.

martonp (changes requested):

client/asset/eth/eth.go
```diff
@@ -2292,6 +2294,19 @@ func (eth *baseWallet) SyncStatus() (bool, float32, error) {
 	return eth.node.peerCount() > 0, 1.0, nil
 }
 
+// TransactionFeesPaid returns exactly how much gwei was used to send a
+// transaction. The transaction must be already mined for this function to work.
+// Returns asset.CoinNotFoundError for unmined txn.
+func (eth *baseWallet) TransactionFeesPaid(ctx context.Context, coinID dex.Bytes) (uint64, error) {
+	var txHash common.Hash
+	copy(txHash[:], coinID)
+	fees, err := eth.node.transactionFeesPaid(ctx, txHash)
```
martonp, line 2303: I think this should wait for a few additional confirmations on the transaction just in case.

client/core/trade.go
JoeGruffins: I probably don't understand correctly, but it does check per coinID I think. It check's on the `match.MetaData.Proof.TakerSwap` or `match.MetaData.Proof.TakerRedeem` coin ID

Do you mean that the tx fee burden may be spaced over several matches?

Hmm, yeah. Should we make an effort to divide it up based on weight if one tx would init or redeem several matches? Currently it might end up in a larger than actual value if you were doing strict accounting across all of your swaps.

client/asset/eth/eth.go
JoeGruffins: Sure, how many confirms?

client/core/trade.go
buck54321: > Redeems are batched, so multiple matches may use the same transaction. This needs to be checked per CoinID instead of per match.

I want to avoid `*Core` not doing any coin ID inspection. Just running each match separately.

client/asset/eth/eth.go
buck54321: Let's use 10. Same as redeems.

client/core/trade.go
martonp: If two matches are redeemed by the same coin ID, then the current code will count double the fees. If we're avoiding any CoinID inspection in Core then the I guess the only way is to check how many redemptions were done by the tx in the wallet, and then divide the transaction fee by that amount.

chappjc: The double counting applies to all assets, not just dynamic tx assets, so it's a bug that would need solving for say DCR and BTC too.

chappjc: Oh nevermind it's not an issue in the other assets.

client/core/trade.go
```diff
@@ -1239,14 +1244,111 @@ func (t *trackedTrade) isSwappable(ctx context.Context, match *matchTracker) (re
...
+// updateFeesPaid starts a goroutine that updates the fees used for dynamic fee
+// checker transactions. We do not know the exact fees the tx will use until
+// they are mined, so this waits until they are mined and updates the value for
+// the entire trade.
+//
+// TODO: This method does not persist if the client is shut down. It probably
+// should because payed fees will be incorrect forever for this trade if not.
+//
+// This method accesses match fields and MUST be called with the trackedTrade
+// mutex lock held for reads.
```
chappjc, line 1266: Looks like the goroutine it starts has it's own locking, while `updateFeesPaid` does not access match fields itself.

client/core/trade.go
chappjc: Chatting on Matrix a bit, a suggestion might be to update `client/db.MatchProof` with a `SwapBatchSize uint32` that we set in `swapMatchGroup` where we know the batch size (and similar for redeem).

I think it lets us continue to iterate through matches like now and just `+= actualFees / batchSize`.   Iterating over `batchSize` matches, we get the desired effect of just adding `actualFees` instead of `actualFees * batchSize`.

client/core/trade.go
```diff
@@ -1239,14 +1244,111 @@ func (t *trackedTrade) isSwappable(ctx context.Context, match *matchTracker) (re
...
+				if errors.Is(err, asset.CoinNotFoundError) {
+					return false
+				}
+				t.dc.log.Errorf("Failed to determine actual %s transaction fees paid for "+
+					"match %s: %v", txType, match, err)
+				return true
+			}
+			t.mtx.Lock()
+			if isInit {
+				t.metaData.SwapFeesPaid += actualSwapFees
```
buck54321, line 1317: Didn't we already `+=` in `swapMatchGroup`. We're not subtracting that original value, are we?

I think we might need to move the `SwapFeesPaid` and `RedemptionFeesPaid` fields to the `MatchMetadata` if we're going to update the fees?

chappjc: There's a `!dynamic` guarding the `+=` in `swapMatchGroup`

client/core/trade.go
JoeGruffins: Oh, I see now. The order may have the same swap counting multiple times even though it was only one transaction.

chappjc: Right.  Long convo in matrix ongoing, but I think we're gravitating toward:

```go
TransactionFeesPaid(ctx context.Context, coinID dex.Bytes) (fees uint64, secretHashes [][]byte, err error)
```

And then the caller knows which match index in the batch they are, plus the number of swaps in the batch.

Then the fees can be added different ways:
- add the full amount for just the leader swap (matching first secret hash) and zero for the rest
- OR do division by the batch size and have the leader also add any remainder

client/core/trade.go
JoeGruffins: It takes the trade wallets, but I guess those never change so doesn't need the read lock. Will remove comment.

JoeGruffins:
Just rebased.

client/core/trade.go
JoeGruffins: If it's the same coinid can we just use that to ensure that tx has only been added once for the order?

I forget if matches from different orders can be grouped together and have the same init/redeem tx

chappjc: > If it's the same coinid can we just use that to ensure that tx has only been added once for the order?

@buck54321 doesn't want core interpreting coin IDs, and I agree 

> I forget if matches from different orders can be grouped together and have the same init/redeem tx

They cannot.

JoeGruffins: Ok. So will need to check on the tx each time? no cache?

JoeGruffins: We can pretend geth is cacheing everything.

client/asset/eth/eth.go
JoeGruffins: Set to 10.

JoeGruffins:
Changes https://github.com/decred/dcrdex/compare/59d20c75ffbb5c38d04f51adcf688adea03c3505..a6d6ce3181780487f4ca022aebd172b218e7e01a

Now only adding fees for one match if any share a transaction.

client/core/trade.go
```diff
@@ -1239,14 +1245,124 @@ func (t *trackedTrade) isSwappable(ctx context.Context, match *matchTracker) (re
...
+					return false
+				}
+				t.dc.log.Errorf("Failed to determine actual %s transaction fees paid for "+
+					"match %s: %v", txType, match, err)
+				return true
+			}
+			if confs < updateFeesConfs {
+				return false
+			}
+			actualSwapFees, _, position, err := feeChecker.TransactionFeesPaid(ctx, coinID, contractData, isInit)
```
martonp, line 1317: It works like this, but why are you returning both the position and the list of secret hashes?

JoeGruffins: It was suggested to do so, but I wasn't sure why. Can remove the secrets return. Coin id is the secret here, so could also go with just the secrets. Probably that was the thinking.

client/core/trade.go
```diff
@@ -1239,14 +1245,124 @@ func (t *trackedTrade) isSwappable(ctx context.Context, match *matchTracker) (re
...
+			}
+			t.mtx.Unlock()
+			err = t.db.UpdateOrderMetaData(t.ID(), t.metaData)
+			if err != nil {
+				t.dc.log.Errorf("Error updating order metadata for order %s: %v", t.ID(), err)
+			}
+			return true
+		}
+
+		// Check once before entering the loop.
```
JoeGruffins, line 1343: Probably no point to this first check as we need 10 confs now.

client/core/trade.go
chappjc: It was suggested to just return the secret hashes and let the caller match with the known `MatchProof.SecretHash`, but not also the position.  However, just the position does seem to be fine, so yeah let's ditch the secrets return.

Also, it's a strangely named method now that it only works on swap contract interactions and not all txns.  I feel like we should rename it or have it work in the absence of provided contract data.

Finally, I'm wondering if init vs redeem cannot be determined from the tx call data instead of requiring that `isInit`.

martonp: Yeah I think just submitting a CoinID and getting back the list of secret hashes and the fee is possible.

JoeGruffins: > MatchProof.SecretHash, but not also the position.

Oh yeah, that would have been the thing to do.

> Finally, I'm wondering if init vs redeem cannot be determined from the tx call data instead of requiring that isInit.

Can try one then the other and only error if both fail.

> Also, it's a strangely named method 

`TransactionFeesPaid` or `updateFeesPaid`?

JoeGruffins:
Just rebased.

client/core/trade.go
JoeGruffins: RIP geth

client/core/trade.go
JoeGruffins: https://github.com/decred/dcrdex/compare/4ad09b10f27bede4e765edaaa5772927a5ef303e..1eb6260a25e50f9e4fb902322d53ab87757558aa

Decided to just use the secrets so removed the position. It felt weird to just return the position. Caller is checking the returns against the secret hash.

No longer taking the isInit bool.

Added `Dynamic` to everything.

Or did I missread.

>  I feel like we should rename it or have it work in the absence of provided contract data.

It needs the contract data for the version. change to `SwapFeesPaid` ok?

JoeGruffins: renamed things to `Swap` https://github.com/decred/dcrdex/compare/1eb6260a25e50f9e4fb902322d53ab87757558aa..7ea288bb8c61cfdb74190cd34d8c3c36c5370135

JoeGruffins: The core tests wait a bit now without the first check...

client/asset/interface.go
```diff
@@ -420,6 +420,14 @@ type LogFiler interface {
 	LogFilePath() string
 }
 
+// DynamicSwapFeeChecker accepts an initiation or redemption coinID and returns
+// the fee spent on the transaction along with the secrets included in the tx.
+// Returns asset.CoinNotFoundError for unmined txn. Will also error if the
+// secret in the contractData is not found in the transaction secrets.
+type DynamicSwapFeeChecker interface {
+	DynamicSwapFeesPaid(ctx context.Context, coinID, contractData dex.Bytes) (fee uint64, secretHashes [][]byte, err error)
+}
```
buck54321, line 429: Names and functionality don't really match. We typically use "swap" to signify the "initiate" or "participate" txs, not the redemption. If we want this to pull double duty, we should name it more generally, but I'd prefer separate methods for swaps and redemptions since they will be handled differently internally with regards to data parsing.

client/core/trade.go
```diff
@@ -1239,14 +1245,120 @@ func (t *trackedTrade) isSwappable(ctx context.Context, match *matchTracker) (re
...
 		match.setSwapConfirms(int64(confs))
+
 		t.notify(newMatchNote(TopicConfirms, "", "", db.Data, t, match))
 		return false, false
 	}
 
 	return false, false
 }
 
+// updateDynamicSwapFeesPaid starts a goroutine that updates the fees used for
```
buck54321, line 1258: > starts a goroutine

Do we have to? This seems like it would be simpler and more on-pattern without a goroutine.

client/core/trade.go
```diff
@@ -1542,6 +1655,7 @@ func (c *Core) tick(t *trackedTrade) (assetMap, error) {
 			redeems = append(redeems, match)
 			received += match.Quantity
 			quoteReceived += calc.BaseToQuote(match.Rate, match.Quantity)
+			t.updateDynamicSwapFeesPaid(c.ctx, match, false)
```
buck54321, line 1658: Feels weird scheduling a fees watcher before we even successfully send the tx. Can this be reactive instead, more like `isSwappable` and `isRedeemable` which examine the current state to determine what actions are needed?

client/core/trade.go
JoeGruffins: I'll see what it looks like.

client/asset/eth/eth.go
```diff
@@ -2322,6 +2326,72 @@ func (eth *baseWallet) SyncStatus() (bool, float32, error) {
 	return eth.node.peerCount() > 0, 1.0, nil
 }
 
+// DynamicSwapFeesPaid returns exactly how much gwei was used to send an
+// initiation or redemtion transaction. It also returns the secret hashes
```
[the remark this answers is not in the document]
JoeGruffins: fixed

JoeGruffins:
just rebased

client/asset/eth/eth.go
```diff
@@ -2322,6 +2326,72 @@ func (eth *baseWallet) SyncStatus() (bool, float32, error) {
...
+// included with this init or redeem. Secret hashes are sorted so returns are
+// always the same, but the order may not be the same as they exist in the
+// transaction on chain. The transaction must be already mined for this
+// function to work. Returns asset.CoinNotFoundError for unmined txn. Will also
+// error if the secret in the contractData is not found in the transaction
```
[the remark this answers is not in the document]
JoeGruffins: fixed

client/asset/interface.go
JoeGruffins: Ok. Names are rather long but the interface implements two methods now. https://github.com/decred/dcrdex/compare/426f1421d142e92123f332020cea28bbfd8e74df..6780dde689d4493e5288e966d3b97557ab4c60e3

JoeGruffins:
Rebased on top of #1638 because that has a few changes to when a match is complete.

client/core/trade.go
chappjc: If it blocks in the main tick loop causing more lock contention that's a no-go from me.

client/core/trade.go
JoeGruffins: Not a goroutine https://github.com/decred/dcrdex/compare/0f8028f67ae1b69f3bf1563dce2027543d06322c..e411b4e266045aa22009185c9b6bede40b02bc58

client/core/trade.go
chappjc: Very good point though about it starting here before even sending the tx.  Seems more appropriate in swapMatchGroup and redeemMatchGroup.  Maybe I need to check out this branch and poke around.

chappjc:
I like that last commit @JoeGruffins, but you don't even need the `check` closure anymore.

client/core/trade.go
```diff
@@ -1239,14 +1269,107 @@ func (t *trackedTrade) isSwappable(ctx context.Context, match *matchTracker) (re
...
+			if errors.Is(err, asset.CoinNotFoundError) {
+				return false
+			}
+			t.dc.log.Errorf("Failed to determine actual %s transaction fees paid for "+
+				"match %s: %v", txType, match, err)
+			return true
+		}
+		if confs < updateDynamicSwapOrRedemptionFeesConfs {
+			return false
+		}
```
buck54321, line 1335: Why wait until 10 confirmations to get a fee update? Is this in case there's a reorg and the fee changes? In that case, it's up to the wallet to determine that a value is not yet safe to report.

JoeGruffins: Will move to wallet. We still want to wait 10 confs though right? Going by this comment https://github.com/decred/dcrdex/pull/1752#discussion_r952515671

chappjc: Personally I think it's a little silly to wait for more than 2 given this is just ancillary info that has no bearing on the success of the swap.  But whatever you guys think...

JoeGruffins: Two sounds good to me.

JoeGruffins:
Moved checking number of confirms to the asset and only checking for 2 confirms. Removed the check function. https://github.com/decred/dcrdex/compare/e411b4e266045aa22009185c9b6bede40b02bc58..8e6bdb89088e7b850c5c54e09f56fea941194fec

chappjc:
Can rebase with https://github.com/decred/dcrdex/pull/1638 merged.

client/core/trade.go
JoeGruffins: I don't think I've made this any more reactive. It is still queuing the check here. If I made `needsDynamicSwapFeeConfirms` to `dynamicSwapFeeConfirmed` and checked those off for all the coins it could be more like that. Let me know if that's the route I should go. The check later would then be pretty much the same plus determining if this is an initiation or redemption which we already know this way.

buck54321 (commented):
Looks much better, but I'm getting an error 

```Failed to determine actual redemption transaction fees paid for match 3827c6098b5638f6cd625bc41d026ff55bd29f77857c882f1c9a8a0f70aa372a: invalid swap data```

File: client/core/trade.go [hunk 14]
```diff
@@ -1270,6 +1276,90 @@ func (t *trackedTrade) isSwappable(ctx context.Context, match *matchTracker) (re
...
+		if match.Side != order.Maker {
+			coinID = []byte(match.MetaData.Proof.TakerSwap)
+		}
+	} else {
+		coinID = []byte(match.MetaData.Proof.MakerRedeem)
+		if match.Side != order.Maker {
+			coinID = []byte(match.MetaData.Proof.TakerRedeem)
+		}
+	}
+	contractData := match.MetaData.Proof.ContractData
```
chappjc, line 1322: Seems to need `CounterContract` instead for the redeem.  Otherwise this would be e.g. your BTC contract's redeem script.

JoeGruffins:
Just rebased.

client/core/trade.go
JoeGruffins: Oh yeah... randomly didn't hit this testing... fixed: https://github.com/decred/dcrdex/compare/56872cb137011870f79609b4d065794b8035f930..e04a5d3ef44253041f2a1f81a64caf4dbdfa1bbf

JoeGruffins:
@buck54321 error hopefully resolved by https://github.com/decred/dcrdex/compare/56872cb137011870f79609b4d065794b8035f930..e04a5d3ef44253041f2a1f81a64caf4dbdfa1bbf as I should be looking at the counterContract and not our init for redeem.

client/core/trade.go
```diff
@@ -1116,6 +1119,9 @@ func (t *trackedTrade) matchIsRevoked(match *matchTracker) bool {
 // If redeemMustConfirm is true, a match that has completed but the redemption
 // has not yet been confirmed will still be considered active.
 func (t *trackedTrade) matchIsActive(match *matchTracker, redeemMustConfirm bool) bool {
+	if match.needsDynamicSwapFeeConfirms || match.needsDynamicRedemptionFeeConfirms {
+		return true
+	}
```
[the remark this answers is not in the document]
JoeGruffins: There is a TODO related to this https://github.com/decred/dcrdex/blob/e04a5d3ef44253041f2a1f81a64caf4dbdfa1bbf/client/core/trade.go#L1290-L1292

client/core/trade.go
```diff
@@ -201,10 +202,12 @@ func (fs *feeStamped) get() uint64 {
...
 
-// freshRedeemFeeAge is the expiry age for cached redeem fee rates, past which
-// fetchFeeFromOracle should be used to refresh the rate. See
-// cacheRedemptionFeeSuggestion.
-const freshRedeemFeeAge = time.Minute
+const (
+	// freshRedeemFeeAge is the expiry age for cached redeem fee rates, past which
+	// fetchFeeFromOracle should be used to refresh the rate. See
+	// cacheRedemptionFeeSuggestion.
+	freshRedeemFeeAge = time.Minute
```
chappjc, line 209: OK but why?  This just looks like needless churn to me.

client/core/trade.go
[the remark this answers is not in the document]
chappjc: If the swap is complete and we've confirmed the redeem as required by `redeemMustConfirm` (babysit ETH redemptions), then the order can be retired.

The swap and redeem fees would only be used when viewing the order in the order details page (or exporting order history I suppose).  The way that is done for inactive orders is by loading the order from DB just for that purpose (no longer in `dc.trades`).

So, after on-demand loading the inactive order and it's matches, can we check something as a heuristic to suggest if we need to recompute real fees instead of trusting `OrderMetaData.{SwapFeesPaid,RedemptionFeesPaid}`?  Even if that means adding a flag or something to the DB, at least that means we don't have to adjust our definition of order/match activeness.

chappjc (approved):

chappjc:
I thought this was working now, but I'm seeing 0 fees for the buyer's redeem:

image

Seller's init looks good though:

image

```js
> eth.getTransactionReceipt('0xd165ee4e5989679fdf760f4f899cd6dcd98e6b6454988a91e97783a468ab7a03')
{
  blockHash: "0x4456835c67e7e89b3dfb48f362d516d8f46dba4cb868d0263b91d6991c6e73b9",
  blockNumber: 43,
  contractAddress: null,
  cumulativeGasUsed: 134500,
  effectiveGasPrice: 2003394124,
  from: "0x909c21ddb713beea4ce2ad0d070c5bdb162dfb23",
  gasUsed: 134500,
...
```

2003394124×134500 = 2.69456509678×10¹⁴ wei = 269456.509678 gwei

Although now I'm wondering about how we're rounding to gwei and they that's not 26945**7**

chappjc:
Ah the redemption fees were correct after the redeem confirmed fully.

`[INF] CORE: notify: |SUCCESS| (match) Redemption Confirmed - Your redemption for match d6948f40 in order 168105ea was confirmed`

image

I don't know why it had to get to 10 confs instead of 2 to update those fees though.

client/core/trade.go
```diff
@@ -1276,6 +1282,91 @@ func (t *trackedTrade) isSwappable(ctx context.Context, match *matchTracker) (re
...
+	if isInit {
+		t.metaData.SwapFeesPaid += actualSwapFees
+	} else {
+		t.metaData.RedemptionFeesPaid += actualSwapFees
+	}
+	err = t.db.UpdateOrderMetaData(t.ID(), t.metaData)
+	if err != nil {
+		t.dc.log.Errorf("Error updating order metadata for order %s: %v", t.ID(), err)
+	}
+	stopChecks()
```
chappjc, line 1367: Ah at the end we need to do:

```go
note := newMatchNote(TopicConfirms, "", "", db.Data, t, match)
t.notify(note)
```

Otherwise we have to wait for `(*trackedTrade) confirmRedemption` to send a `MatchNote` with the updated fees.

client/core/trade.go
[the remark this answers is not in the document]
chappjc: Well, I think I realized that as long as the redeem confirmed count is greater or equal to the number for updating the fees, we can remove this `match.needsDynamicSwapFeeConfirms || match.needsDynamicRedemptionFeeConfirms` condition and the DB reload issue is already solved.

client/core/trade.go
```diff
@@ -120,7 +120,8 @@ type matchTracker struct {
 	// send an initialization until it is confirmed with the server (see
 	// authDEX) that the match is not revoked. This should be set on reconnect
 	// for all taker matches in MakerSwapCast. Guarded by exceptionMtx.
-	checkServerRevoke bool
+	checkServerRevoke                                              bool
+	needsDynamicSwapFeeConfirms, needsDynamicRedemptionFeeConfirms bool
```
buck54321, line 124: I think we can get rid of these fields. In `(*Core).tick`, they can be local variables. And as @chappjc pointed out, our `MatchConfirmed` flag should preclude any `MatchIsActive` checks.

buck54321 (changes requested):

client/core/trade.go
JoeGruffins: This must be leftover from a bad rebase. Will remove.

JoeGruffins:
just rebased

JoeGruffins:
Running tests in `/` ran into an error twice:
```2022-09-19 11:55:29.591 [WRN] TEST: Contract error encountered for match 06314483d3227d5d0985ec799c1989851d2027e86e1f0ca252fa09bf724579c3, actor user: 224cbaaafa80bf3bd1ff731590bcbdda5a76f91e60a156994634e91c00000022, swapAsset: 123, isMaker: true, order: 67d2d1b95e4fc0c115b71ade059215aa442e7f5227c3c753e5d936bc0e7bad9f, status: {swapAsset: 123, redeemAsset: 789, swapTime: 0001-01-01 00:00:00 +0000 UTC, swap: <nil>, swapConfirmed: 0001-01-01 00:00:00 +0000 UTC, redeemTime: 0001-01-01 00:00:00 +0000 UTC, redemption: <nil>} using coin ID b9f97d046296124621928739a86671cc180152b953e3bf9d19f825c3dd54ae1688e49efb and contract 01234567ab00000022: test error
2022-09-19 11:55:29.591 [DBG] TEST: Error going to user 224cbaaafa80bf3bd1ff731590bcbdda5a76f91e60a156994634e91c00000022, code: 25, msg: redemption error
2022-09-19 11:55:29.591 [DBG] TEST: handleInit: 'init' received from user 224cbaaafa80bf3bd1ff731590bcbdda5a76f91e60a156994634e91c00000022 for match 06314483d3227d5d0985ec799c1989851d2027e86e1f0ca252fa09bf724579c3, order 67d2d1b95e4fc0c115b71ade059215aa442e7f5227c3c753e5d936bc0e7bad9f
--- FAIL: TestTxWaiters (0.01s)
    swap_test.go:1610: error sending maker swap transaction: maker swap rpc error. code: 64, msg: already received a swap contract, search in progress
```

```
--- FAIL: TestTaper (0.03s)
    queue_test.go:61: last interval wasn't ~5 ms: 1.088372ms

```

client/core/trade.go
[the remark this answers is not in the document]
JoeGruffins: > DB reload issue is already solved.

Currently needs isSwappable or isRedeemable to return with ready true. If loading from db after the status is changed I think the funds will never be added still. But, other than that two confirm window I guess this issue is solved.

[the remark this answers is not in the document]
JoeGruffins: Currently, isSwappable or isRedeemable need to return with ready ok once before we start looking for for the fee transaction. I think there is a two confirm window after that where if the client was shut down we would loose the check flags and the balance not updated if we restored from db because the swap state has already moved on.

client/asset/interface.go
```diff
@@ -135,11 +135,14 @@ const (
...
 	// in the future.
 	ErrSwapNotInitiated = dex.ErrorKind("swap not yet initiated")
-	CoinNotFoundError   = dex.ErrorKind("coin not found")
-	ErrRequestTimeout   = dex.ErrorKind("request timeout")
-	ErrConnectionDown   = dex.ErrorKind("wallet not connected")
-	ErrNotImplemented   = dex.ErrorKind("not implemented")
-	ErrUnsupported      = dex.ErrorKind("unsupported")
+	// ErrNotEnoughConfirms is returned when a transaction is confirmed,
+	// but does not have enough confirmations to be trusted.
+	ErrNotEnoughConfirms = dex.ErrorKind("transaction does not have enough confirmations")
```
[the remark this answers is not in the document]
JoeGruffins: Moved to the bottom.

client/core/trade.go
```diff
@@ -1276,6 +1282,91 @@ func (t *trackedTrade) isSwappable(ctx context.Context, match *matchTracker) (re
...
+		contractData = match.MetaData.Proof.ContractData
+	} else {
+		coinID = []byte(match.MetaData.Proof.MakerRedeem)
+		if match.Side != order.Maker {
+			coinID = []byte(match.MetaData.Proof.TakerRedeem)
+		}
+		contractData = match.MetaData.Proof.CounterContract
+	}
+	secretHash := match.MetaData.Proof.SecretHash
+	if coinID == nil {
```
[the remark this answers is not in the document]
JoeGruffins: Checking not zero.

client/core/trade.go
JoeGruffins: trying...

JoeGruffins: All I can think of is adding a flag to the db for whether this match's fee's have been accounted for, which is probably the best answer so will do that. Hope I'm not missing something obvious here.

client/core/trade.go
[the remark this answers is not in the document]
JoeGruffins: Saving to db... but not changing the confirm logic. So, it`ll find it if it finds it while active for other reasons. Should be fine.

client/core/trade.go
JoeGruffins: Removed them https://github.com/decred/dcrdex/pull/1752/commits/66115041bde7a69977cbb435870f8889077d82a9

JoeGruffins:
Oh yeah, can't write in the check.

client/core/trade.go
```diff
@@ -1276,6 +1276,145 @@ func (t *trackedTrade) isSwappable(ctx context.Context, match *matchTracker) (re
...
+// mutex lock held for reads.
+func (t *trackedTrade) checkRedemptionFeeConfirms(match *matchTracker) bool {
+	if match.MetaData.Proof.RedemptionFeeConfirmed {
+		return false
+	}
+	_, dynamic := t.wallets.toWallet.Wallet.(asset.DynamicSwapOrRedemptionFeeChecker)
+	if !dynamic {
+		// Confirmed will be set in the db.
+		return true
+	}
```
martonp, line 1318:
``` golang
if (match.Side == order.Maker && match.Status < order.MakerRedeemed) || match.Status < order.MatchComplete {
    return false
}
```

martonp (commented):
LGTM

martonp:
This doesn't matter now, but if a wallet does not implement `RedemptionConfirmer` but implements `DynamicSwapOrRedemptionFeeChecker`, then this will not work. Maybe the interfaces should be combined.

client/core/trade.go
```diff
@@ -1545,14 +1684,15 @@ func (c *Core) tick(t *trackedTrade) (assetMap, error) {
...
+	var swaps, redeems, refunds, revokes, searches, redemptionConfirms,
+		dynamicSwapFeeConfirms, dynamicRedemptionFeeConfirms []*matchTracker
 	var sent, quoteSent, received, quoteReceived uint64
 
 	checkMatch := func(match *matchTracker) error { // only errors on context.DeadlineExceeded or context.Canceled
-		side := match.Side
 		if match.Status == order.MatchConfirmed {
 			return nil
 		}
+		side := match.Side
```
JoeGruffins, line 1695: unintended move

client/core/trade.go
```diff
@@ -1592,7 +1732,6 @@ func (c *Core) tick(t *trackedTrade) (assetMap, error) {
 			redeems = append(redeems, match)
 			received += match.Quantity
 			quoteReceived += calc.BaseToQuote(match.Rate, match.Quantity)
-			return nil
```
JoeGruffins, line 1595: unintended removal I think

client/db/types.go
```diff
@@ -480,7 +496,10 @@ func DecodeMatchProof(b []byte) (*MatchProof, uint8, error) {
 		return nil, 0, err
 	}
 	switch ver {
-	case 2: // MatchProofVer
+	case 3: // MatchProofVer
```
JoeGruffins, line 499: wrong name

client/core/trade.go
JoeGruffins: Notifying now anyway.

client/core/trade.go
JoeGruffins: Can shorted below? Will change.

client/core/trade.go
chappjc: I'm not seeing the need for the new MatchProof fields stored in the DB.
I thought it was fine to simply require the redemption conf requirement to be no less than the fee update conf requirement.

client/core/trade.go
```diff
@@ -1276,6 +1276,145 @@ func (t *trackedTrade) isSwappable(ctx context.Context, match *matchTracker) (re
...
+	if match.Side == order.Maker {
+		if match.Status < order.MakerSwapCast {
+			return false
+		}
+	} else {
+		if match.Status < order.TakerSwapCast {
+			return false
+		}
+	}
+	return true
```
chappjc, line 1302: Normally wouldn't fuss about this, but I think this is far easier to follow:

```go
	return match.Status >= order.MakerSwapCast
}
return match.Status >= order.TakerSwapCast
```

client/core/trade.go
chappjc: See related suggestion in https://github.com/decred/dcrdex/pull/1752/files#r974331842
We should be returning the result of an expression rather than boolean literals to simplify these.

client/core/trade.go
chappjc: Oh if we restarted, or didn't have a flag attached to the matchTracker somewhere, we'd repeatedly recompute and add the fees.  Dang.
New DB data it is then.  That's unfortunate.

client/core/trade.go
chappjc: Perhaps you wanted to always get to `checkSwapFeeConfirms`?  Could move the call to `checkSwapFeeConfirms` to happen between `isSwappable` and `isRedeemable` instead of below.

client/db/types.go
JoeGruffins: No, name is ok.

client/core/trade.go
JoeGruffins: Doing this.

client/db/types.go
```diff
@@ -447,6 +453,14 @@ func (p *MatchProof) Encode() []byte {
 	if p.SelfRevoked {
 		selfRevoked = encode.ByteTrue
 	}
+	swapFeeConfirmed := encode.ByteFalse
+	if p.SwapFeeConfirmed {
+		swapFeeConfirmed = encode.ByteTrue
+	}
+	redemptionFeeConfirmed := encode.ByteFalse
+	if p.RedemptionFeeConfirmed {
+		redemptionFeeConfirmed = encode.ByteTrue
```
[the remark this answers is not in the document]
JoeGruffins: Ok.

client/core/trade.go
JoeGruffins: I don't know now. I don't think it need be removed but I did move up the `checkSwapFeeConfirms`

JoeGruffins:
Ensuring this new interface is also a `RedemptionConfirmer` https://github.com/decred/dcrdex/compare/152e5650dea6a74abd68233f872bde80314a92e3..2c64ad6c7baf42283c8b7a902aa754889cd3abd7

JoeGruffins:
just rebased

chappjc:
I don't think there's anything holding this up.  Will test and briefly re-review tomorrow to get this in.

PR #1756, opened 2022-08-03
client/asset/btc: avoid a bbolt deadlock with nested View

chappjc:
This resolves a useless nested `walletdb.View` that causes a reproducible deadlock with a concurrent  `Update`.

**Repro**: Get three DCR buy orders funded by BTC on the book, match a 3 lot DCR sell with those three orders.  When matched, this triggers `swapMatchGroup` to run all three BTC `Swap` calls concurrently (which by design).  I used a `-race` build with Go 1.19.

image

**Cause:**  In my tests, two of them are doing `Updates` like from `refundAddress` > `(*Wallet).NewAddress` (or similar for `changeAddress`), and another that does a `View` for `signTxAndAddChange`.  This is where the problem is since `(*walletExtender).signTransaction` does the nested `View`, where the inner `View` deadlocks with one of the other `Updates`.

This is [called out in the bbolt docs](https://github.com/etcd-io/bbolt/blob/master/README.md#transactions) as a no-no:

> Transactions should not depend on one another and generally shouldn't be opened simultaneously in the same goroutine. This can cause a deadlock as the read-write transaction needs to periodically re-map the data file but it cannot do so while any read-only transaction is open. **Even a nested read-only transaction can cause a deadlock, as the child transaction can block the parent transaction from releasing its resources.**

https://github.com/etcd-io/bbolt/pull/59

The deadlock was exactly as described.  The inner `View` was hung on `db.mmaplock.RLock()` because the `Update` was waiting on `db.mmaplock.Lock()` (in `(*Tx).Commit` > `(*DB).mmap`) and the outer view already had it read-locked.

**Resolution:**  I recently added a comment that the outer `View` seemed to have no purpose, and indeed it did not.  We can just remove the outer `View`.

----

`refundAddress` goroutine hung getting `db.mmaplock.Lock()` in `Update` > `Commit` > `mmap`

image

`signTransaction` goroutine hung in nested `View`, outer `View` already read-locked `mmaplock` and inner view trying to `RLock` again in `beginTx` (but cannot because the `refundAddress` goroutine is waiting for write lock):

image

`changeAddress` goroutine hung getting `db.rwlock.Lock()` in `Update` > `beginRWTx`:

image

### DIFF
--- a/client/asset/btc/spv.go
+++ b/client/asset/btc/spv.go
@@ -1997,9 +1997,7 @@ func (w *walletExtender) signTransaction(tx *wire.MsgTx) error {
 		txIn.SignatureScript = nil
 		txIn.Witness = nil
 	}
-	return walletdb.View(w.Database(), func(dbtx walletdb.ReadTx) error { // what is the db tx for?
-		return txauthor.AddAllInputScripts(tx, prevPkScripts, inputValues, &secretSource{w, w.chainParams})
-	})
+	return txauthor.AddAllInputScripts(tx, prevPkScripts, inputValues, &secretSource{w, w.chainParams})
 }
 
 // txNotifications gives access to the NotificationServer's tx notifications.


client/asset/btc/spv.go
```diff
@@ -1997,9 +1997,7 @@ func (w *walletExtender) signTransaction(tx *wire.MsgTx) error {
 		txIn.SignatureScript = nil
 		txIn.Witness = nil
 	}
-	return walletdb.View(w.Database(), func(dbtx walletdb.ReadTx) error { // what is the db tx for?
```
chappjc, line 2000: I added the comment in the Electrum PR, but didn't investigate further.  I think an early draft of native SPV BTC started with a different approach, but the `secretSource` now manages the DB accesses (calls to `btcwallet/wallet.(*Wallet).AddressInfo`).

JoeGruffins (approved):
Just looks like a mistake to me. I can't imagine a purpose for the dbtx. Must have been left over from an earlier version.

buck54321:
I'll don't know why this is the way it is, but I'll just point out that **btcwallet**'s (*Wallet).txToOutputs` also has `AddAllInputScripts` nested.

buck54321 (approved):

chappjc:
> I'll don't know why this is the way it is, but I'll just point out that **btcwallet**'s `(*Wallet).txToOutputs` also has `AddAllInputScripts` nested.

That's because their `secretSource` does not start a new db transaction, rather using the parent's `ReadBucket`:

```go
type secretSource struct {
	*waddrmgr.Manager
	addrmgrNs walletdb.ReadBucket
}

func (s secretSource) GetKey(addr btcutil.Address) (*btcec.PrivateKey, bool, error) {
	ma, err := s.Address(s.addrmgrNs, addr)
	if err != nil {
		return nil, false, err
	}

```

```
err = walletdb.Update(w.db, func(dbtx walletdb.ReadWriteTx) error {
	addrmgrNs, changeSource, err := w.addrMgrWithChangeSource(
		dbtx, keyScope, account,
	)

...

err = tx.AddAllInputScripts(
	secretSource{w.Manager, addrmgrNs},
)
```

That's no problem because

PR #1757, opened 2022-08-03
reload wallets page when new wallets are created

### LINKED ISSUE #1754: ui: note feeder error
On the Wallets page after creating a native DCR wallet.

image

ukane-philemon:
This diff modifies `createWalletSuccess` to reload `wallets` page as part of the success callback for wallet creation. Closes #1754 and maybe #1763.

### DIFF
--- a/client/webserver/site/src/js/wallets.ts
+++ b/client/webserver/site/src/js/wallets.ts
@@ -615,13 +615,8 @@ export default class WalletsPage extends BasePage {
   async createWalletSuccess () {
     const rowInfo = this.rowInfos[this.walletAsset]
     this.showMarkets(rowInfo.assetID)
-    const a = rowInfo.actions
-    Doc.hide(a.create)
-    Doc.show(a.send, a.deposit, a.settings)
     await app().fetchUser()
-    if (app().walletMap[rowInfo.assetID].encrypted) {
-      Doc.show(a.lock)
-    }
+    await app().loadPage('wallets')
   }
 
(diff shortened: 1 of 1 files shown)


buck54321:
What was the problem? This seems like we're just ignoring some underlying issue.

chappjc:
Various notes arrive and we're looking for some `data-balance-target` and data-conversion-target` elements that don't exist.  These only seem to be in the dom if assetinfo has a wallet set.

chappjc:
I suppose you'd want a hidden html template somewhere that gets cloned and displayed for the created wallet.  Personally I think this is a rare thing to do, and we might as well refresh everything on the page.

buck54321:
We should be creating the `data-balance-target` elements for every asset and just hiding the unused ones. That's consistent with what we do elsewhere.

This is all going to be different with #1700. Are you looking for a solution for 0.5?

chappjc:
It's not horrible 0.5.  But it's just lame that user has to refresh the page or it's incorrect looking.  You wanna do a quick proper fix?
If it's too much or introduced new bugs, I'd really just assume we refresh the page for 0.5.  It doesn't hurt and it's very rare to do as I said.

buck54321:
OK to refresh page for 0.5.

chappjc:
OK.  For this PR then, should the `fetchUser` come **before** `loadPage`.  That's the pattern everywhere else.

buck54321:
> OK. For this PR then, should the `fetchUser` come **before** `loadPage`? That's the pattern everywhere else.

Yes

PR #1758, opened 2022-08-03
client: fix markets page

ukane-philemon:
Each time the markets page is loaded I see these errors. Some elements might not be made available if we run `app().loadPage('markets')` without `await` which will result in `notes` and other methods trying to access elements that are not available. Don't know if there is more to these errors.

Screenshot 2022-08-03 at 11 28 17 PM
Screenshot 2022-08-03 at 11 29 42 PM

EDIT:  Also this.
Screenshot 2022-08-05 at 12 34 36 AM

EDIT 2: Looks like noteFeeder is trying to access a property on an object that is not yet available. Just like in #1754. It goes away after a refresh. I think these errors are all related to not waiting for `app().loadPage('markets')`.

Screenshot 2022-08-05 at 12 59 32 AM

### DIFF
--- a/client/webserver/site/src/js/charts.ts
+++ b/client/webserver/site/src/js/charts.ts
@@ -1207,6 +1207,7 @@ function makeLabels (
   valFmt = valFmt || formatLabelValue
   const n = screenW / spacingGuess
   const diff = max - min
+  if (n < 1 || diff <= 0) return { lbls: [] }
   const tickGuess = diff / n
   // make the tick spacing a multiple of the step
   const tick = tickGuess + step - (tickGuess % step)
--- a/client/webserver/site/src/js/forms.ts
+++ b/client/webserver/site/src/js/forms.ts
@@ -1466,7 +1466,7 @@ export class DEXAddressForm {
     }
     if (!this.dexToUpdate && res.paid) {
       await app().fetchUser()
-      app().loadPage('markets')
+      await app().loadPage('markets')
       return
     }
     if (this.pwCache) this.pwCache.pw = pw
--- a/client/webserver/site/src/js/login.ts
+++ b/client/webserver/site/src/js/login.ts
@@ -18,6 +18,6 @@ export default class LoginPage extends BasePage {
   /* login submits the sign-in form and parses the result. */
   async loggedIn () {
     await app().fetchUser()
-    app().loadPage('markets')
+    await app().loadPage('markets')
   }
 }
--- a/client/webserver/site/src/js/markets.ts
+++ b/client/webserver/site/src/js/markets.ts
@@ -2125,7 +2125,7 @@ export default class MarketsPage extends BasePage {
       // or this may be the first opportunity to get the server's config, so
       // fetch it all before reloading the markets page.
       await app().fetchUser()
-      app().loadPage('markets')
+      await app().loadPage('markets')
     }
   }
 
--- a/client/webserver/site/src/js/register.ts
+++ b/client/webserver/site/src/js/register.ts
@@ -214,7 +214,7 @@ export default class RegistrationPage extends BasePage {
   /* Called after successful registration to a DEX. */
   async registerDEXSuccess () {
     await app().fetchUser()
-    app().loadPage('markets')
+    await app().loadPage('markets')
   }
 
   async newWalletCreated (assetID: number) {


chappjc:
Haven't seen those errors and I usually run with the dev console open in the browser.  Specific steps?  Browser, etc?

ukane-philemon:
> Haven't seen those errors and I usually run with the dev console open in the browser. Specific steps? Browser, etc?

I use chrome. Repo: start dexc -> login -> will redirect to markets page. Dev console should be open to see errors.

AFAICT, all redirects to the `markets` page cause these errors.

chappjc:
Interesting.  I cannot reproduce with Chromium 103.0.5060.134 or FireFox 103.0b9.

Have you always seen them or just recently?

chappjc:
Despite being unable to reproduce the issue, I did test this PR and it works.

ukane-philemon:
I use Chrome version 104.0.5112.79 (Official Build) (x86_64)

buck54321:
Can't repro myself either. Maybe put a `console.log` in **charts.ts** at that line

```Javascript
x = Number(x.toPrecision(sigFigs))
```
and see what's happening.

ukane-philemon:
These errors are random but I've seen them frequently. AFAICT, waiting for `app().loadPage('markets')` has solved this for me.

JoeGruffins:
Maybe related to network latency? Is the dex server local or is this testnet and looking at the testnet server maybe?

ukane-philemon:
>is this testnet and looking at the testnet server maybe?

Yes, it is.

chappjc:
Trying to introduce server latency with the following, and it does slow the responses, but still no JS

```diff
diff --git a/server/comms/link.go b/server/comms/link.go
index 94d8eb06..27c1851d 100644
--- a/server/comms/link.go
+++ b/server/comms/link.go
@@ -4,6 +4,7 @@
 package comms
 
 import (
+       "math/rand"
        "sync"
        "time"
 
@@ -118,6 +119,7 @@ func (c *wsLink) Authorized() {
 
 // The WSLink.handler for WSLink.inHandler
 func handleMessage(c *wsLink, msg *msgjson.Message) *msgjson.Error {
+       time.Sleep(time.Duration(rand.Intn(3)) * time.Second)
        switch msg.Type {
        case msgjson.Request:
                if msg.ID == 0 {
```

martonp (approved):
I cannot reproduce this either, but I don't see any problem arising from this change.

chappjc:
I can reproduce this error if I doc the browser dev console at the bottom like you have it and scrunch up the visible part of the page to the point where the depth chart is zero height.

image

The only way I can see this error is if in `makeLabels` `screenW` or `max`-`min` are crazy.  That's what happens when the depth chart is not a reasonable size:

image

Reloading the page with this windows size results in the error coming from the `"book"` notification from the websocket message handler, just like you've shown:

image

Chromium 104.0.5112.79

I do not see any way that the markets page cannot be loaded before that first notification with the order book arrives and is passed to `handleBook` because the request for the subscription is made at the end of the market page's constructor.
I'm still looking into it, but it doesn't all add up yet.

chappjc:
You may be onto a deeper issue, so I'd like to know if other errors still pop out on your machine with the following diff:

```diff
diff --git a/client/webserver/site/src/js/charts.ts b/client/webserver/site/src/js/charts.ts
index d04e0eac..20e90022 100644
--- a/client/webserver/site/src/js/charts.ts
+++ b/client/webserver/site/src/js/charts.ts
@@ -1207,6 +1207,7 @@ function makeLabels (
   valFmt = valFmt || formatLabelValue
   const n = screenW / spacingGuess
   const diff = max - min
+  if (n < 1 || diff <= 0) return { lbls: [] }
   const tickGuess = diff / n
   // make the tick spacing a multiple of the step
   const tick = tickGuess + step - (tickGuess % step)
```

ukane-philemon:
@chappjc, thanks for the investigation. I can see in this case `min` is zero hence the error from `x.toPrecision(...)` because `x` is zero.

chappjc:
Although I don't see a lot of the errors you've posted in this thread, there is a problem on the 0.5 release branch where whenever there's an automatic reconnect with the server I get a `setExtents` property error

image

I encourage people to get on mainnet to try.

I will try this PR's fixes on mainnet.

chappjc (approved):
I can confirm that at least the errors I see on reconnect go away with this PR when backported to release-v0.5.  Seems we need this.

I wish I understood how you get all the other errors, but I see no issue with this PR.

I also think we should incorporate the bounds check suggested in https://github.com/decred/dcrdex/pull/1758#issuecomment-1217276712

I'm testing these on a 0.5 branch at https://github.com/chappjc/dcrdex/tree/markets-fix with an active mainnet server to watch chart changes and reconnects when the internet or proxies hiccup.

ukane-philemon:
> I also think we should incorporate the bounds check suggested in

Yes, I was still doing some testing due to this 

> Also, I'm beginning to encounter other errors in my console. I will apply a fix in this PR

Saw an issue (maybe @vctt94 was trying to fix) that caused the errors on my last error image. Maybe should make another PR for that?

JoeGruffins (approved):

PR #1761, opened 2022-08-04
prevent concurrent login in Core

chappjc:
Concurrent attempts take a short path when the existing one unblocks.
Subsequent attempts do full Login, as before.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1098,6 +1098,7 @@ type Core struct {
 	ctx           context.Context
 	wg            sync.WaitGroup
 	ready         chan struct{}
+	loginSlot     chan struct{}
 	cfg           *Config
 	log           dex.Logger
 	db            db.DB
@@ -1217,6 +1218,7 @@ func New(cfg *Config) (*Core, error) {
 		credentials:   creds,
 		ready:         make(chan struct{}),
 		log:           cfg.Logger,
+		loginSlot:     make(chan struct{}, 1),
 		db:            boltDB,
 		conns:         make(map[string]*dexConnection),
 		wallets:       make(map[uint32]*xcWallet),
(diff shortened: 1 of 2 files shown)


chappjc:
In the comments I've spelled out the changes that need to happen eventually:

```go
// The design issue that requires this is that we require multiple Login
// calls to succeed without a Logout between them. To properly solve this,
// Login would actually error if already logged in. However:
//  1. Login gives "restart" capabilities for loading and resuming trades.
//  2. We lack a mechanism for repeated callers (e.g. impatient HTTP
//     requests or sessions from other browsers) to get the result from the
//     already-running Login.
//
// Fundamentally, we need to separate the following notions:
//  1. login to Core (credentials check, get cookies)
//  2. Core initialization (prep internal state, e.g. loading data from DB)
//  3. login to the server (initializeDEXConnections > authDEX).
```

JoeGruffins (approved):

martonp (approved):

client/core/core.go
```diff
@@ -3563,6 +3565,54 @@ func (c *Core) Login(pw []byte) (*LoginResult, error) {
...
+
+	// Connecting and loading wallets might take time, but we don't want
+	// concurrent wallet connect attempts running, especially for native wallets
+	// that use the same file system resources. NOTE: The following channel
+	// mechanism similar to the following, and it does suggest some design
+	// issues, described below.
+	//
+	// !c.loginMtx.TryLock() {
+	//  c.loginMtx.Lock()
+	//  defer c.loginMtx.Lock()
```
buck54321, line 3591: It's commented, but it should be  `Unlock()`.

client/core/core.go
```diff
@@ -3563,6 +3565,54 @@ func (c *Core) Login(pw []byte) (*LoginResult, error) {
...
+	// }
+	defer func() { <-c.loginSlot }() // remove the peg from the slot when done
+	select {
+	case c.loginSlot <- struct{}{}: // full login
+	default:
+		c.log.Warnf("Login already running! Waiting for it to complete.")
+		// Just wait and take the short path.
+		c.loginSlot <- struct{}{}
+		return loginResult(), err
+	}
```
buck54321, line 3602: Well this is interesting.

buck54321 (approved):

client/core/core.go
chappjc: buffered channel tricks y'all

PR #1764, opened 2022-08-05
client/asset/btc: remove incorrect error return

chappjc:
The fixes a regression in `client/asset/btc.(*baseWallet).FundingCoins` introduced in https://github.com/decred/dcrdex/pull/1607/files#diff-1b07e494f4bdd99b2a617ac110fbadc081e5b6ae1abf23f64032bbee5ebebef1

This error return made sense before because it meant it had found the UTXO in the `listunspent` results but it did not find the vout in the `gettransaction` result `details` array.

Now if you have reached this spot, it means you have successfully found the output and added it to `btc.fundingCoins[pt] = utxo`.


martonp (approved):

chappjc:
Updated `TestFundingCoins` with a test that would fail with the misplaced error.
This is the last 0.5 change.

JoeGruffins (approved):

Issue #1769, opened 2022-08-09
UI: Orders screen shows refunding incorrectly sometimes.


JoeGruffins:
When I first visited the page:
![image](https://user-images.githubusercontent.com/34998433/183633127-5175bd2f-2304-4796-82f4-a271c23a1227.png)

After some time, and I don't think it was a refresh:
![image](https://user-images.githubusercontent.com/34998433/183633681-77dda5ec-0616-4120-9a8d-a0327867a2d5.png)

JoeGruffins:
The locktimes for the order are past which may be the reason. Unable to redeem because the output would be dust, which is human error I suppose.

chappjc:
I have also seen it switch from "Revoked - X" to another status.

chappjc:
Ok so `OrderUtil.matchStatusString` needs to match `(*matchReader).StatusString` is all I think.

buck54321:
> Ok so `OrderUtil.matchStatusString` needs to match `(*matchReader).StatusString` is all I think.

related: https://github.com/decred/dcrdex/issues/1179#issuecomment-920056750

Issue #1760, opened 2022-08-04
ui: Accelerate button pops up for wrong trade.


JoeGruffins:
You can be looking at a completed order and when another order goes from epoch to maker cast, the `Accelerate` button will pop up where you are looking.

To reproduce:
Start the simnet harness and complete one trade on the dcr_btc market.
Make another trade and click on the executed order and wait a bit. The accelerate button will pop up. You can click it and accelerate some order.
Refreshing the page makes it go away.

![image](https://user-images.githubusercontent.com/34998433/182830474-b0fb4ea5-2698-4e82-abc8-9e8836c6dd84.png)

I often see some weird things happen on this screen. Maybe some investigation can get to the bottom of it all.

martonp:
I think the reason is that the updates happen based on the match ID. And since both sides of the match are you, this causes issues.

JoeGruffins:
Oh, could be... But this is a different match. I will try with separate dexc.

JoeGruffins:
I have confirmed it happens even when using separate clients.

martonp:
OK I'll take a look.

chappjc:
Resolved by https://github.com/decred/dcrdex/pull/1772

PR #1772, opened 2022-08-09
ui: correct match status for revoked matches

chappjc:
Resolves https://github.com/decred/dcrdex/issues/1769

### DIFF
--- a/client/webserver/site/src/js/order.ts
+++ b/client/webserver/site/src/js/order.ts
@@ -255,8 +255,6 @@ export default class OrderPage extends BasePage {
       Doc.hide(swapSpan, cSwapSpan)
     }
 
-    console.log(m)
-
     Doc.tmplElement(card, 'status').textContent = OrderUtil.matchStatusString(m)
   }
 }
--- a/client/webserver/site/src/js/order.ts
+++ b/client/webserver/site/src/js/order.ts
@@ -256,7 +256,7 @@ export default class OrderPage extends BasePage {
       Doc.hide(swapSpan, cSwapSpan)
     }
 
-    Doc.tmplElement(card, 'status').textContent = OrderUtil.matchStatusString(m.status, m.side)
+    Doc.tmplElement(card, 'status').textContent = OrderUtil.matchStatusString(m)
   }
 }
 
--- a/client/webserver/site/src/js/order.ts
+++ b/client/webserver/site/src/js/order.ts
@@ -192,12 +192,11 @@ export default class OrderPage extends BasePage {
   handleOrderNote (note: OrderNote) {
     const page = this.page
     const order = note.order
+    if (order.id !== this.orderID) return
     this.order = order
     const bttn = page.cancelBttn
-    if (bttn && order.id === this.orderID) {
-      if (bttn && order.status > OrderUtil.StatusBooked) Doc.hide(bttn)
-      page.status.textContent = OrderUtil.statusString(order)
-    }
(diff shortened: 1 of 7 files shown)


chappjc:
Oh, `note.id` is a note token, not the order ID. Last commit uses the CoreOrder id.  I suspect that'll fix https://github.com/decred/dcrdex/issues/1760

A related issue I noticed is that self revokes like:

```
[TRC] CORE: Revoked match 07f3b7f146f247de0414241dcf957e844887429d3c5547b6e27fc85efb2772ab (Taker) in status MakerSwapCast considered inactive.
[INF] CORE: Retiring inactive order 3e7a77531f4afdb470e90b43df901b4a0f76b7a17b8e37c2968dc2232ab82421 in status executed
```

get retired sometimes before the server revoke that triggers proper revoke handling.  So if you try to look at these retired orders and matches they don't look revoked.  When you restart dexc they seem to get reconciled again and they look proper.

Going to switch gears for now.

client/webserver/site/src/js/order.ts
```diff
@@ -256,6 +255,8 @@ export default class OrderPage extends BasePage {
       Doc.hide(swapSpan, cSwapSpan)
     }
 
+    console.log(m)
```
buck54321, line 258: ↑

chappjc:
Just observed this when testing Decrediton update for v0.5 (not including this PR):

image

Having the order page open while a swap is settling where you are both sides highlights some of these issues.  I think this one is also resolved by the order ID filter in this PR.

JoeGruffins (approved):
Also resolves #1760 for me.

martonp (approved):

client/webserver/site/src/js/order.ts
```diff
@@ -192,12 +192,11 @@ export default class OrderPage extends BasePage {
   handleOrderNote (note: OrderNote) {
     const page = this.page
     const order = note.order
+    if (order.id !== this.orderID) return
     this.order = order
     const bttn = page.cancelBttn
-    if (bttn && order.id === this.orderID) {
-      if (bttn && order.status > OrderUtil.StatusBooked) Doc.hide(bttn)
-      page.status.textContent = OrderUtil.statusString(order)
```
chappjc, line 199: I don't know why this line would be conditional on `bttn`.

buck54321 (approved):

client/webserver/site/src/js/order.ts
buck54321: Because we're omitting the element in the Go template instead of using `class="d-hide"`. It's a silly way to do it.

chappjc: Oh, so `page.status` doesn't exist unless `page.cancelBttn` exists too?

chappjc:
```
      <div class="order-datum">
        <div>[[[Status]]]</div>
        <div>
          <span id="status">{{$ord.StatusString}}</span>
          {{if $ord.Cancelable}}
            <span id="cancelBttn" class="ico-cross fs12 ms-1 pointer hoverbg" data-tooltip="cancel order"></span>
          {{end}
```

Looks like `page.status` would be there even if `page.cancelBttn` is not.

buck54321: I wasn't commenting on why setting the status string was inside of the bttn conditional. Not sure on that, tbh.

chappjc: Gotcha.  I highlighted too many lines with the github comment.  The `page.status` condition must have been an error.

PR #1773, opened 2022-08-09
client/asset/dcr: update native dcrwallet for new testnet3 rules

chappjc:
This update is needed to get dcrwallet SPV updated to use the new testnet3 rules: https://github.com/decred/dcrwallet/pull/2182

I'm also sneaking in an update to the testnet LTC explorer because the bitaps one is broken.  Keeping the mainnet one though.


JoeGruffins (approved):
Working on testnet.

buck54321 (approved):

PR #1780, opened 2022-08-10
client/db/bolt: store asset versions with order

chappjc:
`UpdateOrder` must store the asset versions for both to/from assets.
Otherwise we wait for the first `UpdateOrderMetaData` data to actually store the real values.
This can be bad for LTC where we are on v1 already when we switched to segwit.

This fix was pulled from the draft https://github.com/decred/dcrdex/pull/1653#discussion_r916910330, and in that PR I also remove a bunch of stuff from `UpdateOrderMetaData` that should not be written except on order creation by `UpdateOrder`, but I'm not doing that here.


buck54321 (approved):
Yikes

PR #1781, opened 2022-08-11
server/{matcher,market}: base start and end rate calcs on actual match rates

buck54321:
We have always used the mid-gap at the beginning and end of the epoch for our candle's start rate and end rate. In a traditional market, you would use the first and last match price. Use the expected numbers instead. The candle should also not go to zero on an empty book anymore.


server/db/driver/pg/internal/epochs.go
```diff
@@ -18,6 +18,11 @@ const (
 	InsertEpoch = `INSERT INTO %s (epoch_idx, epoch_dur, match_time, csum, seed, revealed, missed)
 		VALUES ($1, $2, $3, $4, $5, $6, $7);`
 
+	SelectLastEpochRate = `SELECT end_rate
+		FROM %s
+		ORDER BY epoch_end DESC
```
chappjc, line 23: Is `PRIMARY KEY` => :+1:

server/db/driver/pg/internal/epochs.go
```diff
@@ -18,6 +18,11 @@ const (
...
+		LIMIT 1;`
+
 	// CreateEpochReportTable creates an epoch_reports table that holds
 	// epoch-end reports that can be used to construct market history data sets.
 	CreateEpochReportTable = `CREATE TABLE IF NOT EXISTS %s (
```
chappjc, line 28: The comments for `start_rate` and `end_rate` still refer to mid-gap.

server/matcher/match.go
```diff
@@ -152,10 +152,24 @@ func (m *Matcher) Match(book Booker, queue []*OrderRevealed) (seed []byte, match
 
 	updates = new(OrdersUpdated)
 	stats = new(MatchCycleStats)
-	startRate := midGap(book)
-	stats.StartRate = startRate
-	stats.LowRate = startRate
-	stats.HighRate = startRate
+
+	defer func() {
```
chappjc, line 156: I don't have a problem with this `defer`, but why is it a `defer`?  I see only one return this method.

server/db/driver/pg/internal/epochs.go
chappjc: For the testnet `dcr_btc.epoch_reports` table with 5,809,377 rows:

```
 Limit  (cost=0.43..0.47 rows=1 width=16) (actual time=0.032..0.032 rows=1 loops=1)
   Output: end_rate, epoch_end
   ->  Index Scan Backward using epoch_reports_pkey on dcr_btc.epoch_reports  (cost=0.43..205329.76 rows=5726653 width=16) (actual time=0.030..0.030 rows=1 loops=1)
         Output: end_rate, epoch_end
 Planning Time: 0.103 ms
 Execution Time: 0.048 ms
```

That's good.

However, it returns 0 because it has not been running this code yet to ensure `end_rate` won't be stored zero.  We could add `WHERE end_rate>0` for a small penalty:

```
 Limit  (cost=0.43..0.50 rows=1 width=16) (actual time=4.713..4.715 rows=1 loops=1)
   Output: end_rate, epoch_end
   ->  Index Scan Backward using epoch_reports_pkey on dcr_btc.epoch_reports  (cost=0.43..219646.39 rows=3139871 width=16) (actual time=4.711..4.712 rows=1 loops=1)
         Output: end_rate, epoch_end
         Filter: (epoch_reports.end_rate > 0)
         Rows Removed by Filter: 15225
 Planning Time: 0.127 ms
 Execution Time: 4.737 ms
 ```

This would of course introduce the possibility of returning no rows, which you have covered already.

chappjc: Or we could keep the fastest query possible and just wait for the first epoch with a trade match and it would never be an issue.  I'm thinking keep it without the condition to stay fast.

chappjc (changes requested):
Looks good and tests well on testnet DB.  Just remove the defer, moving contents down near `bookVolumes`, and this is good to merge.

server/db/driver/pg/internal/epochs.go
buck54321: This was my thinking too. Once it's non-zero now, it will never be zero again.

chappjc (approved):

PR #1798, opened 2022-08-18
client: fix markets page JS errors

ukane-philemon:
To reproduce: 
1. start dexc without internet connection -> login -> markets page [here JS will complain `cannot read property of undefined (reading dex)`] 
Screenshot 2022-08-18 at 6 42 42 PM

2. connect to the internet while on the markets page. This will trigger a page reload when the user account auto connects to a dex. While here, the page reload will hang(forever until manual refresh) because spots and balance updates keep coming when the market page is not fully set up.  This resolves some of the issues mentioned in #1758 

Screenshot 2022-08-18 at 7 49 01 PM
Screenshot 2022-08-17 at 9 17 36 AM
Screenshot 2022-08-18 at 6 46 14 PM


JoeGruffins (approved):
I can confirm the js errors and behavior and that this pr fixes them.


There is another js error I see if I start not connected to the dex and navigate to the wallets page, since you are fixing all these. You don't have to fix in this pr though.

image

ukane-philemon:
> There is another js error I see if I start not connected to the dex and navigate to the wallets page, since you are fixing all these. You don't have to fix in this pr though.

Alright, I will look into it.

chappjc:
Related, which I'm assigning to myself: https://github.com/decred/dcrdex/issues/1583

client/webserver/site/src/html/bodybuilder.tmpl
```diff
@@ -90,4 +90,3 @@
 </body>
 </html>
 {{end}}
-5
```
chappjc, line 93: For reference and other reviewers, the reasoning:
- https://unix.stackexchange.com/a/18789/133297
- https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline
- https://thoughtbot.com/blog/no-newline-at-end-of-file
- https://stackoverflow.com/questions/5813311/whats-the-significance-of-the-no-newline-at-end-of-file-log

ukane-philemon: is the "5" intentional?

chappjc: WTF.  No that has to be an accident.  I was just referring to the newline at the end of the file, but that "5" makes no sense there.

chappjc (approved):
LGTM.

buck54321 (approved):
5

PR #1799, opened 2022-08-18
site: set zoom buttons in DepthChart ctor

chappjc:
This resolves the error `Uncaught TypeError: can't access property "setExtents", e.zoomOutBttn is undefined`, which happens often on DEX server reconnect events.  Probably more so if you have more than one server configured.
The best way to reproduce is:
- get the dex window opened to the markets page
- switch to another tab in the same window, effectively hiding the tab
- minimize the window for good measure, but neither of these steps are really required
- wait for a reconnect, or wait a minute or two and force a reconnect (e.g. restart tor, toggle vpn, etc).

The root of the issue seems to be that the parent constructor was not setting the zoom buttons as intended.  And the `requestAnimation` callback that actually was setting them via the `resized` reporter is not called until the browser decides to actually draw on the canvas.  This is asynchronous, and the steps above try to make it a longer delay.

```
The zoom buttons were not being initialized in the constructor as
indended. This calls setZoomBttns directly from the constructor of the
inheriting class, DepthChart.

This also fixes a missing ws.deregisterRoute for updateRemainingRoute.
```

### DIFF
--- a/client/webserver/site/src/html/bodybuilder.tmpl
+++ b/client/webserver/site/src/html/bodybuilder.tmpl
@@ -86,7 +86,7 @@
 {{end}}
 
 {{define "bottom"}}
-<script src="/js/entry.js?v=GHTDEG56"></script>
+<script src="/js/entry.js?v=YKVNRn4"></script>
 </body>
 </html>
 {{end}}
--- a/client/webserver/site/src/js/charts.ts
+++ b/client/webserver/site/src/js/charts.ts
@@ -169,7 +169,6 @@ class Chart {
     this.ctx = ctx
     this.ctx.textAlign = 'center'
     this.ctx.textBaseline = 'middle'
-    this.setZoomBttns()
     // Mouse handling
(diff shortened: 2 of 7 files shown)


buck54321 (approved):
I spent a good deal of time trying to repro this and was unable to, but this change is righteous. Better to not address child-only properties in the parent class constructor.

chappjc:
Looks like with `webpack --watch`, the parent constructor does seem to affect the child class, but not when fully compiled into a webpack bundle.  The dev.js config with watch doesn't use babel, so maybe transpiling changes something.

My guess is that it has something to do with how the TypeScript class field declarations are made into js.  https://jsfiddle.net/bf9w410v/

JoeGruffins (approved):
I was able to reproduce the error with a method suggested by chapp on simnet turning off the server while not looking at the tab and letting it reconnect:

image


However, I cannot seem to make it happen with these changes.

PR #1800, opened 2022-08-19
site: fix wallet page JS errors

ukane-philemon:
Quoting @JoeGruffins 
> There is another js error I see if I start not connected to the dex and navigate to the wallets page, since you are fixing all these. You don't have to fix in this pr though.

This resolves: 
Screenshot 2022-08-19 at 1 59 07 PM

To reproduce: start dexc without internet connection -> login -> wallets page. While here the first auto-selected wallet will only display the wallet detail table, while the markets table and recent activity table remain blank(selecting another wallet loads the markets table and recent activity table but still produce errors).
btcmarketserror

### DIFF
--- a/client/webserver/site/src/js/wallets.ts
+++ b/client/webserver/site/src/js/wallets.ts
@@ -411,6 +411,7 @@ export default class WalletsPage extends BasePage {
     const page = this.page
     const markets: [string, Market][] = []
     for (const xc of Object.values(app().user.exchanges)) {
+      if (!xc.markets) continue
       for (const mkt of Object.values(xc.markets)) {
         if (mkt.baseid === assetID || mkt.quoteid === assetID) markets.push([xc.host, mkt])
       }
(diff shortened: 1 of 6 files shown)


chappjc (approved):
Looks good.  Sorry for the dumb template conflicts though.

client/webserver/site/src/js/wallets.ts
```diff
@@ -411,6 +411,7 @@ export default class WalletsPage extends BasePage {
     const page = this.page
     const markets: [string, Market][] = []
     for (const xc of Object.values(app().user.exchanges)) {
+      if (!xc.markets) continue
```
chappjc, line 414: Just confirming that 0.5 has this check (twice, redundantly, actually) in it's `showMarkets`.

ukane-philemon: `setSelectedAsset` will run when the wallets page loads hence the error from `showAvailableMarkets`

chappjc: I understand.  I'm saying the bug is only on master (does not need backport for release).

JoeGruffins (approved):

Issue #1782, opened 2022-08-12
meterer_test: Random local failure.


JoeGruffins:
Running tests locally, one randomly failed. This is on a commit one off from current master, but nothing was touched in the `dcrdex/dex` area. Test was run using the script in root `./run_tests.sh`. Sorry that I don't have more info.
```
--- FAIL: TestDelayedRelay (0.10s)
    meterer_test.go:53: last delay was less than minimum. 19.662726ms < 20ms
FAIL
FAIL    decred.org/dcrdex/dex/meter     0.131s
```

PR #1802, opened 2022-08-20
dex/meter: fix flaky TestDelayedRelay test

LasTshaMAN:
Left a comment in the test describing the root cause (**edit**: see 1st commit in this PR; actually, this test can be easily re-designed to avoid having to deal with Go scheduler - which is the latest version of this PR, i.e. 2nd commit).

Closes https://github.com/decred/dcrdex/issues/1782.

### DIFF
--- a/dex/meter/meterer.go
+++ b/dex/meter/meterer.go
@@ -39,7 +39,7 @@ func DelayedRelay(ctx context.Context, minDelay time.Duration, n int) (out <-cha
 
 				if time.Since(last) > minDelay {
 					outC <- nil
-					last = time.Now().UTC()
+					last = time.Now()
 					continue
 				}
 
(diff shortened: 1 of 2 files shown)


dex/meter/meterer.go
```diff
@@ -51,7 +51,7 @@ func DelayedRelay(ctx context.Context, minDelay time.Duration, n int) (out <-cha
 			case <-now:
 				outC <- nil
 				scheduled = nil
-				last = time.Now().UTC()
+				last = time.Now()
```
LasTshaMAN, line 54: I believe there is no need for .UTC() here, so I removed it to simplify a bit.

dex/meter/meterer_test.go
```diff
@@ -43,22 +39,24 @@ func TestDelayedRelay(t *testing.T) {
 		t.Fatalf("wrong signal count. expected 1, got %d", n)
 	}
 
-	// Sending two signals simultaneously should end up with two signals, at
-	// least minDelay apart.
+	// 2nd signal should be delayed at least for minDelay.
+	sentTime := time.Now()
```
chappjc, line 43: This *is* the root of the issue with the test. The signal receiver can receive as lazily as it likes.

chappjc (approved):

PR #1804, opened 2022-08-20
multi: Implement disable wallet feature

### LINKED ISSUE #1794: disable wallet function
When you start and login to dexc, it tries to connect to all previously configured wallets.  These might be external wallets like Bitcoin Core, Electrum-LTC, etc.  Or maybe you just don't want the native wallet for asset X getting online and establishing peers anymore.

You should be able to "disable" a wallet into Wallets page so that dexc will not attempt to connect or start the wallet until it is re-enabled.

ukane-philemon:
- Implement a `disable` feature for wallets.
- Add a new `RPC` route `togglewalletstatus` to toggle wallet status. Closes #1794 

UI changes:
Screenshot 2022-08-20 at 9 21 42 PM

Screenshot 2022-08-20 at 9 21 27 PM


client/db/bolt/db.go
```diff
@@ -104,6 +104,7 @@ var (
 	byteTrue               = encode.ByteTrue
 	backupDir              = "backup"
 	disabledRateSourceKey  = []byte("disabledRateSources")
+	walletStatusKey        = []byte("walletStatus")
```
chappjc, line 107: This is too vague of a name and string.  Why not `walletDisabledKey`?

client/webserver/locales/en-us.go
```diff
@@ -262,4 +262,9 @@ var EnUS = map[string]string{
 	"No Recent Activity":          "No Recent Activity",
 	"Recent asset_name Activity":  "Recent <span data-asset-name=1></span> Activity",
 	"Other Actions":               "Other Actions",
+	"disable":                     "Disable Wallet",
+	"enable":                      "Enable Wallet",
```
chappjc, line 266: `disable_wallet` and `enable_wallet` for the keys pls

chappjc (changes requested):
Thanks for tackling this.  Seems reasonable overall.  Will take a closer look at the mechanisms in core though.

client/webserver/locales/en-us.go
ukane-philemon: > disable_wallet and enable_wallet for the keys pls

Alright, but does it mean other  multiple word `key` without `_` in between  will && should be changed to `xx_xx`?

chappjc: Don't need to change them all, but ideally we would have established a convention earlier.  I just prefer a key with no spaced and which is clearly not intended for display.

client/db/bolt/db.go
ukane-philemon: Does your suggestion apply to other use of `walletstatus`? e.g `togglewalletstatus` and `updateWalletStatus`?

chappjc: Just this probably.

client/rpcserver/handlers.go
```diff
@@ -840,6 +868,19 @@ var helpMsgs = map[string]helpMsg{
...
+	toggleWalletStatusRoute: {
+		pwArgsShort: "appPass",
+		argsShort:   `assetID disable`,
+		cmdSummary:  `Disable or enable an existing wallet.`,
+		pwArgsLong: `Password Args:
+    appPass (string): The DEX client password.`,
+		argsLong: `Args:
+    assetID (int): The asset's BIP-44 registered coin index. e.g. 42 for DCR.
+      See https://github.com/satoshilabs/slips/blob/master/slip-0044.md
+	disable (bool): The wallet's status. e.g To disable a wallet set to "true", to enable set to "false"`,
```
JoeGruffins, line 881: Could you move disable to the left more? I think it's just four spaces and not tab.
Looks like this now.
```
$ ./dexcctl --simnet help togglewalletstatus
togglewalletstatus assetID disable

Disable or enable an existing wallet.

Args:
    assetID (int): The asset's BIP-44 registered coin index. e.g. 42 for DCR.
      See https://github.com/satoshilabs/slips/blob/master/slip-0044.md
        disable (bool): The wallet's status. e.g To disable a wallet set to "true", to enable set to "false"

Returns:
    string: The message "[coin symbol] wallet has been [wallet status]"
```

JoeGruffins (approved):
Working well, just a few nits.

Would be nice if there was a way to tell here that the btc wallet is disabled:
image

And it would be nice is this error message told me which wallet is disabled, and that I should disable in settings:
image

ukane-philemon:
> and that I should enable in settings

The `walletDisabled` error message is generic since we use it for the `rpcserver` and `webserver`.

chappjc:
> The `walletDisabled` error message is generic since we use it for the `rpcserver` and `webserver`.

But you can wrap it.  `return fmt.Errorf("%v %w", strings.ToUpper(unbip(assetID)), walletDisabled)`
Or use a coded error using `newError` with a message that contains the symbol,  and recognize the code on the frontend (e.g. `activeOrdersErrCode` in wallets.tx).

ukane-philemon:
> But you can wrap it. return fmt.Errorf("%s %w", strings.ToUpper(unbip(assetID)), walletDisabled)

Yeah, I am updating this PR with some changes which include having a `walletDisabledErrStr`, i.e `return fmt.Errorf(walletDisabledErrStr, unbip(assetID))`

ukane-philemon:
> Would be nice if there was a way to tell here that the btc wallet is disabled:

> And it would be nice is this error message told me which wallet is disabled,

I've added an icon for disabled wallets on the markets page, through which users can easily enable the wallet and also improved the error message to include the asset name.

client/webserver/locales/en-us.go
```diff
@@ -262,4 +262,10 @@ var EnUS = map[string]string{
 	"No Recent Activity":          "No Recent Activity",
 	"Recent asset_name Activity":  "Recent <span data-asset-name=1></span> Activity",
 	"Other Actions":               "Other Actions",
+	"disable_wallet":              "Disable Wallet",
+	"enable_wallet":               "Enable Wallet",
+	"change_wallet_status":        "Change Wallet Status",
+	"disable_wallet_warning":      "Note: This wallet will not be connected to when you start the DEX client software and cannot be used until it is enabled. Ensure you do not have active trades.",
```
chappjc, line 268: Backend checks `assetHasActiveOrders` anyway, so that last sentence seems unnecessary.  Plus we could also consider `app().haveAssetOrders(assetID)` on the frontend.

chappjc (changes requested):

client/webserver/api.go
```diff
@@ -676,6 +676,25 @@ func (s *WebServer) apiWalletSettings(w http.ResponseWriter, r *http.Request) {
 	}, s.indent)
 }
 
+// apiToggleWalletStatus updates the wallet's status.
+func (s *WebServer) apiToggleWalletStatus(w http.ResponseWriter, r *http.Request) {
+	form := new(walletStatusForm)
+	if !readPost(w, r, form) {
+		return
+	}
+	err := s.core.ToggleWalletStatus(form.Pass, form.AssetID, form.Disable)
```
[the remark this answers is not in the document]
ukane-philemon: > Intentionally disallowing cached pass with s.resolvePass here and {{$passwordIsCached := .UserInfo.PasswordIsCached}} in forms.tmpl?

I'm not. I no longer think the password is required.

chappjc:
Gotta delete the localized_html on this one too

chappjc:
Clicking the red X for the BCH wallet **from the markets page** gives the confirm form:

image

Then clicking confirm:

image

I can re-enable from the wallets page, but not the markets page.

client/webserver/site/src/html/forms.tmpl
```diff
@@ -544,3 +544,18 @@
   </div>
 </div>
 {{end}}
+
+{{define "toggleWalletStatusConfirm"}}
+<div class="form-closer hoverbg"><span class="ico-cross"></span></div>
+<div class="px-2 py-1 text-center position-relative fs22 sans-light">
+   [[[change_wallet_status]]]
```
chappjc, line 551: "Change Wallet Status" is a bit confusing.  Why not get rid of `change_wallet_status` and instead use `[[[disable_wallet]]]` or `[[[enable_wallet]]]` spans that get enabled or disabled like the message text below this?

ukane-philemon:
> I can re-enable from the wallets page, but not the markets page.

The issue here is fixed.

> Also, after re-enabling from the wallets page, it still looks disabled until I refresh the page.

I don't think you'll need a refresh. The wallets page gets updated via the `WalletState` note and there could be a minor delay before the `note` hits the frontend.

client/webserver/site/src/html/forms.tmpl
```diff
@@ -544,3 +544,18 @@
   </div>
 </div>
 {{end}}
+
+{{define "toggleWalletStatusConfirm"}}
+<div class="form-closer hoverbg"><span class="ico-cross"></span></div>
+<div class="px-2 py-1 text-center position-relative fs22 sans-light">
+   <span id="toggleWalletStatus" class="d-hide"></span>
```
chappjc, line 551: About this last change, I'm confused why you chose to do this with a javascript locale const but three lines down you did the spans with the html template localizations and show/hide the intended span. 
Is there a reason? Because the inconsistency stands out.

chappjc:
> I don't think you'll need a refresh. The wallets page gets updated via the `WalletState` note and there could be a minor delay before the `note` hits the frontend.


https://user-images.githubusercontent.com/9373513/188765054-78f2b4d2-b14e-457e-8114-4afb368eadbc.mp4

client/webserver/site/src/html/forms.tmpl
ukane-philemon: > Also we already have separate translations for in the html template strings vs. js locale consts.

I'm now utilizing the translations in `locale.go`.

chappjc:
I can't see what part of the diff fixed that refresh issue.  Going forward, will you try to make commits or push in such a way that it's easy to click something to see what changed?

ukane-philemon:
> Going forward, will you try to make commits or push in such a way that it's easy to click something to see what changed?

Alright, I didn't want to make a new commit as it was really minor.

chappjc:
> Alright, I didn't want to make a new commit as it was really minor.

Doesn't matter to me as we squash the PR before merging.  But you can still squash on your end as long as when you force  push you don't change the merge base commit in the same push.  Because now when I click the link to compare the changes in the last force push, it shows all the changes that happened on master mixed in with whatever tiny change you made.

What was it though?  I still can't find it.

chappjc:
Oh I see your comment now https://github.com/decred/dcrdex/commit/af12ccf43f4768b943436c818bd97db2c2d8cfad#r83326609

client/webserver/api.go
```diff
@@ -733,6 +733,25 @@ func (s *WebServer) apiWalletSettings(w http.ResponseWriter, r *http.Request) {
 	}, s.indent)
 }
 
+// apiToggleWalletStatus updates the wallet's status.
+func (s *WebServer) apiToggleWalletStatus(w http.ResponseWriter, r *http.Request) {
+	form := new(walletStatusForm)
+	if !readPost(w, r, form) {
+		return
+	}
+	err := s.core.ToggleWalletStatus(form.AssetID, form.Disable, form.Force)
```
buck54321, line 742: Since we're not allowing forced disables through the UI, it's clearer, imo, to hard code the value `false` here and drop the `Force bool` field. Otherwise, someone has to go look in javascript to see that there is no real option.

client/webserver/site/src/js/markets.ts
```diff
@@ -243,10 +243,15 @@ export default class MarketsPage extends BasePage {
       bind(quoteIcons.sleeping, 'click', () => { this.showOpen(this.market.quote, this.walletUnlocked) })
       bind(baseIcons.locked, 'click', () => { this.showOpen(this.market.base, this.walletUnlocked) })
       bind(quoteIcons.locked, 'click', () => { this.showOpen(this.market.quote, this.walletUnlocked) })
+      bind(baseIcons.disabled, 'click', () => { this.showToggleWalletStatus(this.market.base) })
+      bind(quoteIcons.disabled, 'click', () => { this.showToggleWalletStatus(this.market.quote) })
```
buck54321, line 247: Do we need this functionality on the markets view?

ukane-philemon: I think so. Just like every other wallet action like unlocking etc... Plus, It is convenient to enable wallets in one click without having to go to walletsPage -> Setting.

client/webserver/api.go
ukane-philemon: > Since we're not allowing forced disables through the UI, it's clearer, imo, to hard code the value false here and drop the Force bool field.

We allow force disable just like `Rescan` && `Recovery`. Unless we no longer want force disable, we can remove the `Force bool` Field and parameter.

client/webserver/site/src/js/markets.ts
chappjc: I figure the "X" has to be shown, so might as well make it clickable.

client/webserver/api.go
chappjc: Well, what's the use case for force disable?  Rescan and Recovery have clear use cases, when something is broken while you have an active swap, and a rescan or recover if the only recourse.  But why would disabling the wallet be useful with active orders?

chappjc:
After enabling the wallet *from the markets page* and then trying to place an order:

image

ukane-philemon:
> After enabling the wallet _from the markets page_ and then trying to place an order:

Thanks, for the review. The issue there has been resolved.

client/webserver/api.go
ukane-philemon: A` Force` shutdown could be useful if a market order has been placed and there is no match yet(also, market orders do not have a cancel UI like limit orders). However, the current behaviour I've noticed is that a limit order will return an `activeOrder` error when trying to disable a wallet but a market order(without a match) does not.

chappjc: > A` Force` shutdown could be useful if a market order has been placed and there is no match yet

Do you mean when the market order is still in epoch status?  That's definitely a time NOT to disable the wallet.  Facilitating that would mean giving the user a tool to break the rules.  Not a good use case.  If they are set on doing that, they're going to have to kill dexc.

> (market orders do not have a cancel UI like limit orders, I think).

There is no cancelling an epoch status market order because such an order is never booked.

> However, the current behaviour I've noticed is that a booked limit order(without a match) will return an `activeOrder` error when trying to disable a wallet but a market order(without a match) does not (is this an expected behaviour?).

An epoch status market order?  Or a market order that went through the epoch matching cycle and did not match with a maker trade (OrderStatusExecuted)?
The first should absolutely be considered and active order, the latter not.

ukane-philemon: > There is no cancelling an epoch status market order because such an order is never booked.

This is clear. My order could not have completed the epoch matching cycle. I Will get rid of the force disable. I thought you meant we could provide special handling for `acvtiveOrders` error ([comment](https://github.com/decred/dcrdex/pull/1804#discussion_r958823665))

chappjc: Oh, yeah, sorry for the misleading you there.  I just meant we could recognize the code and display a more clear, and perhaps translated, error than we'd get from the go code's `error`

chappjc (approved):

PR #201, opened 2020-03-09
db: Add MatchData.Active field and active column

chappjc:
The matches table has a `status` column with values corresponding to
`order.MatchStatus`, but this did not indicate if match negotiation was
still active.  A failed and revoked match would likely not have `MatchComplete`
as `status`, so a separate `active` flag is introduced to handle this case.
The `active` column also helps identify matches with status `MatchComplete`
that have both redeem txns fully confirmed, but which have *not yet**
received redeem acks from both counterparties.

The swap negotiator will set a match as inactive on (1) revoke, or
(2) when both `SaveRedeemAckSigA` and `SaveRedeemAckSigB` have
stored the acknowledgements of the counterparty's redemption txn.

`MatchArchiver.ActiveMatches` now returns matches for a user
according to the active flag rather than `status <> MatchComplete`.
This helps the server and user identify which matches have **pending
user actions**.  Otherwise the this would list failed swaps that never made
it to `MatchComplete` and swaps in `MatchComplete` status that still
require redemption acks from one or both parties.

PR #1811, opened 2022-08-23
site: fix firefox bug

### LINKED ISSUE #1808: site: change wallet type settings, close form on firefox
If changing the form wallet type on firefox, the form closes and it is not possible to change the wallet type:

image

It seems to happen only on firefox.

ukane-philemon:
I investigated the issue resulting in #1808 and found out firefox and chrome behave differently on `mousedown` events.
I think the issue lies in the fact that Firefox triggers 3 events (onmousedown, onmousedown, onchange) instead of 2 (onmousedown, onchange) like Chrome does. The last `onmousedown` will have the screen coordination value of the event set to zero(cause the last mouse `event`  is actually right in the `Element`?). Closes #1808.

Reference 1: [w3c uievents #201](https://github.com/w3c/uievents/issues/201)

### DIFF
--- a/client/webserver/site/src/js/doc.ts
+++ b/client/webserver/site/src/js/doc.ts
@@ -110,9 +110,7 @@ export default class Doc {
    * the bounds of the specified element.
    */
   static mouseInElement (e: MouseEvent, el: HTMLElement): boolean {
-    const rect = el.getBoundingClientRect()
-    return e.pageX >= rect.left && e.pageX <= rect.right &&
-      e.pageY >= rect.top && e.pageY <= rect.bottom
+    return el.contains(e.target as Node)
   }
 
   /*


chappjc:
This only happens on master, not release-v0.5, so it seems clear it's the recent wallet page restyling, but I don't understand why. There doesn't seem to be any difference with `mouseInElement` or the `mousedown` binding in wallets.ts.

chappjc:
Also, this says to me the bind should be to `onchange`

chappjc:
I'm going to investigate too, but @buck54321 your input is probably needed to discern why this is a problem on master but not `release-v0.5`, or if this is the best fix.

ukane-philemon:
> It's actually because firefox fires a mousedown event for the click on the <option> element,

So I guess this behaviour only affects the `<option>` element? Firefox fires one `mousedown` event for the click on the <option> element, and another for `select`.

buck54321:
In my tests, it was the option element that had the zeroed coordinates. Please do verify.

ukane-philemon:
> In my tests, it was the option element that had the zeroed coordinates. Please do verify.

Yes, I can verify. That was why I added this line. 
```
|| (e.pageX === 0 && e.pageY === 0)
```

client/webserver/site/src/js/doc.ts
```diff
@@ -110,6 +110,7 @@ export default class Doc {
    * the bounds of the specified element.
    */
   static mouseInElement (e: MouseEvent, el: HTMLElement): boolean {
+    if (e.target && (e.target as HTMLElement).tagName === 'OPTION') return true
     const rect = el.getBoundingClientRect()
     return e.pageX >= rect.left && e.pageX <= rect.right &&
       e.pageY >= rect.top && e.pageY <= rect.bottom
```
chappjc, line 116: Seems like we're overthinking this.

```js
  static mouseInElement (e: MouseEvent, el: HTMLElement): boolean {
    return el.contains(e.target as Node)
  }
```

buck54321: Nice find. That should work, with a couple of caveats.

1. You probably need to check if the target is the element itself, otherwise that click padding would become ineffective. Something like
```Javascript
return el === e.target || el.contains(e.target as Node)
```
2. This would fail if we had an element that's not a descendant absolutely positioned over the form. I'm not sure that we do that with anything but tooltips, so it probably won't matter to us.

chappjc: > 1. You probably need to check if the target is the element itself, otherwise that click padding would become ineffective. Something like

https://developer.mozilla.org/en-US/docs/Web/API/Node/contains

contains covers self apparently:

> The contains() method of the [Node](https://developer.mozilla.org/en-US/docs/Web/API/Node) interface returns a boolean value indicating whether a node is a descendant of a given node, that is the node itself, one of its direct children ([childNodes](https://developer.mozilla.org/en-US/docs/Web/API/Node/childNodes)), one of the children's direct children, and so on.
> 
> Note: A node is contained inside itself.

ukane-philemon: > contains covers self, apparently

yeah, your suggestion is efficient.

JoeGruffins (approved):
Can confirm the bug and this fixes it.

chappjc (approved):

PR #1812, opened 2022-08-23
client: add positional fee rate param

### LINKED ISSUE #1706: client/asset/btc: do not use settxfee RPC anymore
https://github.com/decred/dcrdex/blob/b7f667d97c3f59e6d5e4a0ad42fa7911a06fb87d/client/asset/btc/rpcclient.go#L671-L684

This is racy, as we knew when initially coding this.  However, recent versions of Bitcoin Core (and other coin's clones) support a `fee_rate` argument for `sendtoaddress` that we should use instead.

`10. fee_rate                (numeric or string, optional, default=not set, fall back to wallet fee estimation) Specify a fee rate in sat/vB.`

I'm not certain if we can use our current rpcclient easily though because it does not support named args (json object instead of positional args).  We can try to just give all the defaults or empty args for the ones we don't want to set.

The `settxfee` approach may still be required for ancient forks like DOGE however.

vctt94:
This PR closes #1706

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -352,6 +352,9 @@ type BTCCloneCFG struct {
 	ManualMedianTime bool
 	// OmitRPCOptionsArg is for clones that don't take an options argument.
 	OmitRPCOptionsArg bool
+	// LegacySendToAddr sents legacy raw tx which does not have positional fee
+	// rate param.
+	LegacySendToAddr bool
 }
 
 // outPoint is the hash and output index of a transaction output.
@@ -1015,6 +1018,7 @@ func newRPCWallet(requester RawRequesterWithContext, cfg *BTCCloneCFG, parsedCfg
 		chainParams:              cfg.ChainParams,
 		omitAddressType:          cfg.OmitAddressType,
 		legacySignTx:             cfg.LegacySignTxRPC,
+		legacySendToAddr:         cfg.LegacySendToAddr,
 		booleanGetBlock:          cfg.BooleanGetBlockRPC,
 		unlockSpends:             cfg.UnlockSpends,
 		deserializeTx:            btc.deserializeTx,
--- a/client/asset/btc/rpcclient.go
+++ b/client/asset/btc/rpcclient.go
@@ -125,6 +125,7 @@ type rpcCore struct {
 	legacyValidateAddressRPC bool
 	manualMedianTime         bool
 	omitRPCOptionsArg        bool
+	legacySendToAddr         bool
 }
 
 func (c *rpcCore) requester() RawRequesterWithContext {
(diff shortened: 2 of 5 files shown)


client/asset/btc/rpcclient.go
```diff
@@ -770,6 +770,24 @@ func (wc *rpcClient) walletLock() error {
 // sendToAddress sends the amount to the address. feeRate is in units of
 // sats/byte.
 func (wc *rpcClient) sendToAddress(address string, value, feeRate uint64, subtract bool) (*chainhash.Hash, error) {
+	// 1e-5 = 1e-8 for satoshis * 1000 for kB.
+	var txid string
+	// if legacy sign, use legacy sendToAddress.
+	if wc.rpcCore.legacySignTx {
```
chappjc, line 776: This is an interesting inference.  Can it be justified with release notes that show the introduction of the new sign method coincides with the addition of the fee_rate arg for sendtoaddress?
Even if it can, why don't wee make a new `BTCCloneCFG` field for this?  There's a lot of RPC variation with zcashd, dogecoind, litecoind, etc. and we are generally checking each one explicitly.

vctt94: Ok sounds fine. How should it be named? `legacyForks`?

chappjc: I mean more specific, not general.  See the existing bools in `BTCCloneCFG`.  An appropriate name would be `LegacySentRawTx

client/asset/btc/btc.go
```diff
@@ -333,6 +333,9 @@ type BTCCloneCFG struct {
 	TxVersion func() int32
 	// ManualMedianTime causes the median time to be calculated manually.
 	ManualMedianTime bool
+	// LegacySentRawTx sents legacy raw tx which does not have positional fee
+	// rate param.
+	LegacySentRawTx bool
```
chappjc, line 338: OMG, I'm sorry I don't know why I said `LegacySentRawTx` when it's `LegacySentToAddr`

client/asset/zec/zec.go
```diff
@@ -173,6 +173,7 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, net dex.Network) (ass
 		Segwit:                   false,
 		OmitAddressType:          true,
 		LegacySignTxRPC:          true,
+		LegacySentRawTx:          true,
```
chappjc, line 176: :+1: confirmed that zcashd is legacy w.r.t. this:

`sendtoaddress "zcashaddress" amount ( "comment" "comment-to" subtractfeefromamount )`

client/asset/doge/doge.go
```diff
@@ -154,6 +154,7 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network)
 		Segwit:                   false,
 		OmitAddressType:          true,
 		LegacySignTxRPC:          true,
+		LegacySentRawTx:          true,
```
chappjc, line 157: :+1: confirmed that dogecoind is legacy w.r.t. this:

`sendtoaddress "address" amount ( "comment" "comment_to" subtractfeefromamount )`

chappjc (changes requested):
BCH (Core) needs to be legacy too

`sendtoaddress "address" amount ( "comment" "comment_to" subtractfeefromamount coinsel include_unsafe )` (no fee rate)

vctt94:
mb, forgot about bch

client/asset/btc/rpcclient.go
```diff
@@ -768,8 +769,26 @@ func (wc *rpcClient) walletLock() error {
 }
 
 // sendToAddress sends the amount to the address. feeRate is in units of
-// sats/byte.
+// sats/byte. If there is not a fee rate positional param, it is used
+// legacySendToAddress instead.
```
JoeGruffins, line 773: This comment does not look correct. If wc.rpcCore.legacySentToAddr is true legacySendToAddress is used instead?

client/asset/btc/rpcclient.go
```diff
@@ -768,8 +769,26 @@ func (wc *rpcClient) walletLock() error {
...
 func (wc *rpcClient) sendToAddress(address string, value, feeRate uint64, subtract bool) (*chainhash.Hash, error) {
+	// 1e-5 = 1e-8 for satoshis * 1000 for kB.
+	var txid string
```
JoeGruffins, line 776: nit: Can move this down and not declare until after the possible return.

JoeGruffins (approved):

client/asset/btc/rpcclient.go
chappjc: Had typos in my suggestion from the other day, sorry. `legacySendToAddress` indeed looks most appropirate

client/asset/btc/btc.go
```diff
@@ -333,6 +333,9 @@ type BTCCloneCFG struct {
 	TxVersion func() int32
 	// ManualMedianTime causes the median time to be calculated manually.
 	ManualMedianTime bool
+	// LegacySentToAddr sents legacy raw tx which does not have positional fee
+	// rate param.
+	LegacySentToAddr bool
```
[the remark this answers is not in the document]
chappjc: Was a typo in my suggestion comment.  My bad

buck54321 (approved):

client/asset/btc/rpcclient.go
chappjc: I still think this declaration should be moved after the conditional.  It looks sloppy up here.

vctt94: oh sorry, skiped this comment. mb

chappjc (approved):

PR #1813, opened 2022-08-23
client: fix error before spot price

vctt94:
If going to the wallet page before the spot price being calculated, an error will be thrown:

image

This PR fixes it.

### DIFF
--- a/client/webserver/site/src/html/wallets.tmpl
+++ b/client/webserver/site/src/html/wallets.tmpl
@@ -141,13 +141,13 @@
                       </span>
                     </td>
                     <td class="d-none d-md-table-cell d-lg-none d-xxl-table-cell"><div data-tmpl="host" class="short-host"></div></td>
-                    <td>
+                    <td data-tmpl="priceBox">
                       <span data-tmpl="price"></span>
                       <span class="fs13 grey">
                         <sup data-tmpl="priceQuoteUnit" class></sup>/<sub data-tmpl="priceBaseUnit"></sub>
                       </span>
                     </td>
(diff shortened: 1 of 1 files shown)


client/webserver/site/src/js/wallets.ts
```diff
@@ -450,10 +450,10 @@ export default class WalletsPage extends BasePage {
         tmpl.priceBaseUnit.textContent = basesymbol.toUpperCase()
         tmpl.volume.textContent = fourSigFigs(spotVolume(assetID, mkt))
         tmpl.volumeUnit.textContent = assetID === baseid ? basesymbol.toUpperCase() : quotesymbol.toUpperCase()
-      } else Doc.hide(tmpl.priceBox, tmpl.volumeBox)
+        page.marketsOverviewBox.classList.remove('invisible')
+      }
       Doc.bind(row, 'click', () => app().loadPage('markets', { host, base: baseid, quote: quoteid }))
     }
-    page.marketsOverviewBox.classList.remove('invisible')
```
chappjc, line 456: Why should the `marketsOverviewBox` depend on the spot prices being available?  We still want the markets listed whether or not the price and volume are available.

vctt94: No reason. So it should be fine like that:

image

JoeGruffins (approved):
Can confirm the error in console and that this fixes it.

buck54321 (approved):

buck54321:
Looks good. Just gotta rebase away those deleted translations.

PR #1815, opened 2022-08-23
add DE translations

karamble:
this adds translations for the latest release-v0.5 branch strings and adds the language to locales.go


chappjc:
Awesome, thanks!  We're chatting on Matrix, but echoing some things here for future reference:

https://github.com/decred/dcrdex/wiki/Localization-and-Translation
There are currently three different scopes that need translated strings: Go (notifications), html (static page components), js (live page stuff)

chappjc (approved):
Thanks!

Although we're getting rid of localized_html in https://github.com/decred/dcrdex/pull/1826, merging now so it can be applied cleanly to a 0.5 patch.

PR #1817, opened 2022-08-24
client/core: allow changing dex host for down dex

chappjc:
When using `UpdateDEXHost`, it will panic if the DEX server has been down since startup since there is no `dc.cfg` (the config response).  Checking `dc.cfg != nil` is a practice throughout `Core`, just forgotten here.


martonp (approved):
LGTM, it works now.

Issue #1765, opened 2022-08-08
key mesh design decisions


chappjc:
Some key concepts of the current DEX may be difficult to translate to a decentralized config with a simple design, and we need to rethink several things, particularly regarding conduct.

Just going to a mesh configuration with the current designs increases complexity, so we should simplify things as much as possible so that it is robust _and_ maintainable.  In particular: account scoring, the registration fee, matching, and book synchronization.  Mesh will be most feasible if we look at really chopping things down when it looks like something does not move from centralized -> decentralized without lots of complexity, edge cases, and assumptions.  Not just complexity, but potential for loss of consensus between nodes on fundamental things like status of a user/account.

We should start with a few notable challenges in mind and build the solution up from them.  Namely, replacing the reg fee, the issue of perpetual history for user scoring, the problem of consensus on the history, the matching and epoch cycle issues, etc.

### Incentive Mechanism (fee -> bonds)

We will replace the **registration fee** with **fidelity bonds** to solve the issues with registration fees paid to a server operator.  The general problem solved is that in mesh, if you consider a user can be running a server/node themselves, the conduct has to come from _independent_ evidence of misconduct.  A user could have paid a legacy reg fee _to themselves_, the operator of a server. Bonds will form the basis for new solutions in order authorization and/or account scoring/tiers.  Obviously a fee paid to a third party is a blemish on dex too, but this is primarily a mesh solution for this discussion.

### Conduct

As long as we keep the notion of **conduct scores** for each account, nodes have to agree on these scores. Challenges:
- Nodes agreeing on conduct scores for a given account would ideally avoid synchronizing lots of history (complex, error prone, resource intensive).
- How do _new_ nodes joining the mesh trust history being provided, and how to form consensus.  I believe we should avoid all solutions that lead to some kind of consensus algorithm, and ideally move away from history-based solutions as much as possible.

Our current notion of account history that we use to compute a score is unlikely to be translated unchanged into mesh.  It's entirely too much state, which is baggage not only for storage and transfer to other nodes, but forming consensus on what is the true state (order and match outcomes).  If a node joins the mesh, how do they get an account's state?  Just get it from a peer and let it be true?  After a few years, how big is this total mesh "state".  e.g. 1 million accounts, N latest history items per account, etc.  Since we will not be creating an auxiliary blockchain to do help with this, we need to simplify things and get clever to deal with this.

[...]

PR #1819, opened 2022-08-24
fidelity bonds foundation: comms, account tiering, server tracking

chappjc:
Further breaking up https://github.com/decred/dcrdex/pull/1480 for feasible review, this piece deals with:

- Updating and creating new comms routes and payloads 
- Changing the server's account management from a binary active/suspended to a tiered system
- Updating server's DB to track client bonds

To make this transition smoother, the **client and server retain all the legacy registration fee machinery**, and the client's continues to use the legacy registration system.

### Background from https://github.com/decred/dcrdex/pull/1480:

This creates a time-locked fidelity bond, which is redeemable by the user who posted the bond after a certain time. This creates a opportunity cost to use DCRDEX instead of a simple monetary cost.  It also is a prerequisite for [mesh](https://github.com/decred/dcrdex/issues/1765).

### Protocol

NOTE: the following payload and the prevalidation scheme have changed.  See the code.

```go
type PostBond struct {
	Signature
	BondVer    uint16 `json:"bondVersion"`

}
```

Server must use provided account pubkey to verify signed `postbond` request, otherwise the user could be spamming or putting up a bond for a bogus account or someone else's account.  **EDIT**: This `BondSig` will be removed in from this PR since the second output has the account ID and this verification of bond output ownership doesn't seem to matter.

**EDIT:**  May nuke `LegacyFeeRefundAddr`.

The response payload:

```go
type PostBondResult struct {
	Amt       uint64 `json:"amt"`
}
```

  - If it is not found, a `PostBondResult` is sent immediately with `Confs=-1` set.  The bond is NOT stored in the DB. This was a courtesy pre-validation. The user should then broadcast the transaction, wait for the required number of confirmations, and send `postbond` again.  Note that assuming the lowest confs ever accepted will be 1, not unconfirmed, this works because it's virtually certain to have at least propagated to the server before it got mined, but we could also have a flag in the request indicating to just pre-validate and skip trying to locate and activate the bond.
  - Otherwise a coin waiter starts and responds only when the required confirmations are reached.  As such, this last `postbond` request should not be made until the client has observed the required confirmations, and the waiter is only needed for network latency.
<s>  - Otherwise, a response is sent with the observed confirmation count, and a coin waiter is started to watch for the txn to reach the required number of confirmations. The bond is stored in DB as pending. When the transaction reaches the required confirmations, it is flagged active in the DB and a `bondconfirmed` notification is sent.</s>

### TODOs

Refining the bond amount.  It's pretty clear the bond amount has to be at least an order of magnitude more than the previous registration fee since you get it back, and it's only an opportunity cost having it locked up.  I'm thinking like 5 DCR per tier (the bond increment).

See the project board at https://github.com/orgs/decred/projects/2.


server/db/driver/pg/upgrades.go
```diff
@@ -319,6 +324,39 @@ func v5Upgrade(tx *sql.Tx) (err error) {
...
+		log.Warnf("Unexpected existing %q table!", bondsTableName)
+	}
+
+	namespacedBondsTable := publicSchema + "." + bondsTableName
+	err = createIndexStmt(tx, internal.CreateBondsAcctIndexV0, indexBondsOnAccountName, namespacedBondsTable)
+	if err != nil {
+		return fmt.Errorf("failed to index bonds table on account: %w", err)
+	}
+
+	err = createIndexStmt(tx, internal.CreateBondsLockTimeIndexV0, indexBondsOnLockTimeName, namespacedBondsTable)
```
martonp, line 345: Why is there an index on locktime? Isn't it enough to just grab all the bonds for a certain user?

server/auth/auth.go
```diff
@@ -688,25 +802,85 @@ func integrateOutcomes(matchOutcomes *latestMatchOutcomes, preimgOutcomes *lates
...
-	score, _, _ = integrateOutcomes(auth.matchOutcomes[user], auth.preimgOutcomes[user])
+	score, _, _ = auth.integrateOutcomes(auth.matchOutcomes[user], auth.preimgOutcomes[user], auth.orderOutcomes[user])
 	return score
 }
 
+// tier computes a user's tier from their conduct score and the provided active
+// bond total. NOTE: This will be a short-lived helper function while bond
+// amounts are all in the same units (Decred atoms).
+func (auth *AuthManager) tier(bondTotal int64, score int32) int64 {
+	bondTier := bondTotal / int64(auth.minBond) // each asset will have a different minBond / increment
```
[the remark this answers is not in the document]
chappjc: I've come back to this point myself a couple times.  It's in the case of an offline user that you don't have a `client`: https://github.com/decred/dcrdex/pull/1819/files/33986669e42f7ddc35d53b2eb80d27f3ee6aebd0#diff-db497632f7ebb518251e95d0f5e3e66e4527e986708df27c05ce409cbe00d55fR847

server/auth/auth.go
```diff
@@ -434,52 +517,59 @@ func (auth *AuthManager) ExpectUsers(users map[account.AccountID]struct{}, withi
...
+	}
 }
 
 // RecordCompletedOrder records a user's completed order, where completed means
 // a swap involving the order was successfully completed and the order is no
 // longer on the books if it ever was.
 func (auth *AuthManager) RecordCompletedOrder(user account.AccountID, oid order.OrderID, t time.Time) {
-	auth.recordOrderDone(user, oid, nil, db.EpochGapNA, t.UnixMilli())
+	score := auth.recordOrderDone(user, oid, nil, db.EpochGapNA, t.UnixMilli())
+	auth.computeUserTier(user, score) // may raise tier, but no action needed
```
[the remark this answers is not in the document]
chappjc: Ah, yes!  I defined a `TierChangedRoute`, but marked it's use as a TODO
https://github.com/decred/dcrdex/blob/33986669e42f7ddc35d53b2eb80d27f3ee6aebd0/dex/msgjson/types.go#L177-L179

Might as well wire that up in this work.

server/auth/registrar.go
```diff
@@ -25,7 +30,372 @@ var (
...
+	}
+
+	// Must be equal to account ID computed from pubkey in the PayFee message.
+	if commitAcct != acctID {
+		return msgjson.NewError(msgjson.BondError, "invalid bond transaction - account commitment does not match pubkey")
+	}
+
+	// Verify that the user has the private keys to the bond pubkey by verifying
+	// the BondSig against the message (acct ID) and bond script's pubkey hash.
+	msgHash := sha256.Sum256(acctID[:])
```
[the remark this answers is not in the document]
chappjc: This is how the user shows that the bond *output* is actually theirs.  The `commitAcct` comes from the second output with the OP_RETURN, but the first output (or the corresponding redeem script) does not directly reference the account to which it should apply.

server/db/driver/pg/upgrades.go
chappjc: `SelectActiveBondsForUser` uses `lockTime` in a condition to only choose bonds that haven't expired.

Another thought is to DELETE bonds occasionally to prune them when they become spendable (not just expired in the eyes of the DEX), but we still need to select bonds that are actually active even if they aren't refundable yet.

server/auth/registrar.go
```diff
@@ -25,7 +30,372 @@ var (
...
+			LockTime: lockTime,
+			Pending:  pending,
+		}
+
+		// The user may be posting their initial bond (new account) or topping up.
+		newAcct := dbAcct == nil
+		if newAcct {
+			log.Infof("Creating new user account %v from %v, posted first bond in %v (%s) / %v",
+				acctID, conn.Addr(), bondStr, bondAssetSym, bondAddr)
+			err = auth.storage.CreateAccountWithBond(acct, dbBond)
```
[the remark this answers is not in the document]
chappjc: I'll reconsider this, but the interface docs are stale.

Current design is to create the account entry (still tier 0) _as long as the txn is found by the node_, even if unconfirmed.  But it has to be found on the network, thus, the signed and validated bond txn is published.  The idea is that the user can `connect` (tier 0), and then later server can send a `bondconfirmed` notification to the authenticated user once it confirms and `waitBondConf` sees that.

But it must be found by the node.  It won't create accounts based solely on message data anymore like with legacy `'register'`. (See the `auth.checkFee` call above.)

I will reconsider this however.  The client should and does wait until it is fully confirmed before sending the final `postbond` request, but server's node could just be lagging the client's,  so `waitBondConf` is intended to deal with this latency.
I think it should be possible to delay replying with a `PostBondResult` until it's confirmed in the `waitBondConf` goroutine.  I suppose this would make the `bondconfirmed` notification obsolete.

Good suggestion.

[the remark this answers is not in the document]
martonp: My main worry was that there could be accounts in the database that never get a bond confirmed. Then they just live on forever. Also there is the possibility of someone creating many transactions that spend the same input but commit to different accounts, and flooding the database with pointless records. Although it would probably be hard to create so many that it causes any issue.

Even if the account is only created when the bond is confirmed, I think there should be a way to clear accounts that have had no active bonds for a while. No reason to keep them around.

server/auth/registrar.go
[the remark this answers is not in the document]
JoeGruffins: When there is mesh, the user will need to supply this extra signature for every dex? Also, this is always the same, so a server could impersonate a user after seeing this once.

I kind of don't think it is necessary to check or require this. Does it really matter who made the bond? Someone is losing out on opportunity costs for this unique ID.

JoeGruffins (commented):
partial review

server/auth/auth.go
```diff
@@ -688,25 +802,85 @@ func integrateOutcomes(matchOutcomes *latestMatchOutcomes, preimgOutcomes *lates
...
 	return score
 }
 
+// tier computes a user's tier from their conduct score and the provided active
+// bond total. NOTE: This will be a short-lived helper function while bond
+// amounts are all in the same units (Decred atoms).
+func (auth *AuthManager) tier(bondTotal int64, score int32) int64 {
+	bondTier := bondTotal / int64(auth.minBond) // each asset will have a different minBond / increment
+	tierAdj := int64(score) / int64(auth.banScore)
+	return bondTier - tierAdj
```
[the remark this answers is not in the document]
chappjc: > A user could overcome a ban buy locking more funds?

Absolutely, although there's no more "ban" concept any more, just tier.

server/auth/auth.go
```diff
@@ -956,6 +1107,105 @@ func (auth *AuthManager) conn(conn comms.Link) *clientInfo {
...
+func (auth *AuthManager) checkBonds() {
+	lockTimeThresh := time.Now().Add(auth.bondExpiry).Unix()
+
+	checkClientBonds := func(client *clientInfo) ([]*db.Bond, int64, int64) {
+		client.mtx.Lock()
+		defer client.mtx.Unlock()
+		pruned, bondTotal := client.pruneBonds(lockTimeThresh)
+		if len(pruned) == 0 {
+			return nil, bondTotal, -1 // no tier change
+		}
```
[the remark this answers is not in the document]
chappjc: That's taken care of in `(*AuthManager).addBond`, which is called on-demand rather than on a ticker.

server/auth/registrar.go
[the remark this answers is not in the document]
chappjc: > When there is mesh, the user will need to supply this extra signature for every dex? Also, this is always the same, so a server could impersonate a user after seeing this once.

That's true.  It could be a signature of something else though.

> I kind of don't think it is necessary to check or require this. Does it really matter who made the bond? Someone is losing out on opportunity costs for this unique ID.

OK, I think you guys convinced me this does little. I suppose I'll remove `BondSig`.

server/auth/registrar.go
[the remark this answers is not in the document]
chappjc: I do see it your way now.  I did make the changes.

Not sure exactly why I felt the extra complexity of the notifications and the notion of pending bond was worth it, but one challenge deals with the edge case of lost connections after a postbond request is submitted but before a response is received.  I think we can work around that, esp. since if the request is made when client thinks it's fully confirmed, there will be a fairly short period where client is waiting on server to also observe that it's fully confirmed.

> Although it would probably be hard to create so many that it causes any issue.

Next to impossible, I think.  They would be double spends.

> Even if the account is only created when the bond is confirmed, I think there should be a way to clear accounts that have had no active bonds for a while. No reason to keep them around.

That's a somewhat separate issue, but I agree, as long as that account is not deep in the negative tiers on account of major penalties.  We can prune accounts for sure.

chappjc:
Offline convo with @JoeGruffins @buck54321 and @martonp about some possible bond output script changes.

- We can put all the relevant data pushes (lock time and pubkey[hash]) in the `OP_RETURN` or an additional output so that the redeem script does not need to be provided (or in a mesh, stored and sent between servers).  How bonds are handles in mesh is not yet clear, but this would certainly simplify things.

- If we don't require the pubkey in the bond script to be revealed for verification of some signature, and [we're clearly not seeing a compelling reason](https://github.com/decred/dcrdex/pull/1819#discussion_r958342218), we can change back to a `p2pkh` script in the redeem script and gain security.

Ref for both points above:

`<locktime[4]> OP_CHECKLOCKTIMEVERIFY OP_DROP OP_DUP OP_HASH160 <pubkeyhash[20]> OP_EQUALVERIFY OP_CHECKSIG`

dex/msgjson/types.go
```diff
@@ -941,6 +978,92 @@ func (n *PenaltyNote) Serialize() []byte {
...
+// Client should send bond info when their bond tx is fully-confirmed. Server
+// should start waiting for required confs when it receives the 'postbond'
+// request if the txn is found. Client is responsible for submitting 'postbond'
+// for their bond txns when they reach required confs, or when the client
+// restarts and finds bonds recorded in their DB that is flagged as neither
+// accepted nor expired (also maybe if not listed in the 'connect' response).
+
+// PostBond should include the unsigned bond transaction for validation prior to
+// broadcasting the signed transaction so the client hasn't needlessly locked
+// funds if the transaction is rejected.
```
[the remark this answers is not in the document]
chappjc: > If the transaction is rejected then the funds wouldn't be locked

The comment means if there was no unsigned txn being checked first.  We don't want a published+signed txn rejected.

> Server should check so that the bond is not rejected by the server?

Server first validates the transaction structure with the unsigned bond txns, then client submits it again with the confidence that it's good.

Also as you noted elsewhere, this initial comms round can be used to agree on other bond parameters.

JoeGruffins (commented):
> ... relevant data pushes (lock time and pubkey[hash]) ...

And the bond version I suppose.

client/core/core.go
```diff
@@ -3132,20 +3154,24 @@ func (c *Core) discoverAccount(dc *dexConnection, crypter encrypt.Crypter) (bool
 			}
 			return false, newError(authErr, "unexpected authDEX error: %w", err)
 		}
-		if dc.acct.isSuspended {
-			c.log.Infof("HD account key for %s was reported as suspended. Deriving another account key.", dc.acct.host)
+		// do not skip key if tier is 0 and bonds will be used
+		if dc.acct.tier < 0 || (dc.acct.tier < 1 && dc.apiVersion() < serverdex.BondAPIVersion) {
+			c.log.Infof("HD account key for %s was has tier %d (not able to trade). Deriving another account key.",
+				dc.acct.host, dc.acct.tier)
```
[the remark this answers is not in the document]
chappjc: > I guess a client could go far into bad behavior territory now, and have a very negative score. They would need more bond to overcome this. Maybe they would rather make a new account than try to keep up with the higher "fees" they have incurred.
> 
> Conversely, maybe the user would like to lock more bonds rather than skip this account.

Correct that there's actually a choice here because no account is permanently suspended anymore... no more "ban" concept.  In this PR, I'm trying to preserve current behavior of the client, just using the new messages and types.  In this case I'm considering negative tier to mean suspended permanently (not worth posting more bond to get out of the hole).  In following bond PRs we'll certainly get into this more.

> Can tier 0 mean that either the account has no bonds or has enough negative credit that it is shifted to zero?

Yes, can mean either.  Tier is just a balance of active bond strength with penalizations.

> Does a tier less than zero mean definite bad behavior?

That's the only way for the formula to go negative, yes.

https://github.com/decred/dcrdex/blob/72603a9c7e6f668ef8dbb75ceb1f51b3d910d6d1/server/auth/auth.go#L846-L853

server/auth/registrar.go
```diff
@@ -25,7 +30,344 @@ var (
...
+	if confs >= reqConfs {
+		// No need to call checkFee again in a waiter.
+		postBondRes.Active = true
+		auth.storeBondAndRespond(conn, dbBond, acct, newAcct, msg.ID, postBondRes, refundAddr)
+		auth.removeBondWaiter(bondIDKey) // after storing it
+		return nil
+	}
+
+	auth.latencyQ.Wait(&wait.Waiter{
+		Expiration: time.Now().Add(time.Hour),
```
chappjc, line 249: Need to change this to a few minutes since this is now just network latency we're accounting for here.

server/auth/registrar.go
```diff
@@ -25,7 +30,344 @@ var (
...
+		})
+		return
+	}
+
+	if refundAddr != "" {
+		log.Warnf("Account %v posting bond requests a legacy fee refund to %v",
+			acctID, refundAddr) // signature checked above
+		// To actually do this, the operator would need to manually send funds
+		// to the address, then use an admin endpoint to somehow delete the old
+		// fee coin ID from the table.
```
chappjc, line 298: I don't think this legacy fee refund business is worth it any more, regardless of the level of automation.

buck54321 (commented):
For discussion: I think we should drop the pre-validation step. I don't think it's necessary or amenable to mesh. The server version should advertise all we need to know about the bond output structure needed/supported, and all members of the network should be expected to adhere to the protocol without help from a central authority. 

Is there some instance where the server would accept a `config` request, advertise their version, but not accept a new properly constructed bond?

server/auth/registrar.go
```diff
@@ -25,7 +30,344 @@ var (
...
+		}
+	}
+
+	bondStr := coinIDString(assetID, bondCoinID)
+	bondAssetSym := dex.BipIDSymbol(assetID)
+
+	// Determine if it exists and is confirmed.
+	chkAddr, chkAmt, confs, err := auth.checkFee(assetID, bondCoinID)
+	if err != nil {
+		// Not found is expected if this is a pre-broadcast validation request.
```
chappjc, line 196: Alternative to always checking and interpreting a coin-not-found to mean it's a pre-validation request, the request payload could simply indicate this intention.

chappjc:
> For discussion: I think we should drop the pre-validation step. I don't think it's necessary or amenable to mesh. The server version should advertise all we need to know about the bond output structure needed/supported, and all members of the network should be expected to adhere to the protocol without help from a central authority.

It's true that it's not *necessary*, and in fact if the client skips it, server doesn't know or care.

It's also correct that the config response must and does provide all that is needed to form the proper bond.

However, I don't think mesh enters into this at all.  The "central authority" you refer to is the same one where we get the config response, so I don't see anything particularly problematic in mesh with a courtesy pre-check.  This is just a way of ensuring there's not something plainly incorrect, and it should be accepted.  If the real thing is rejected, it would have been rejected on pre-validation, so it's just prudent to do so afaict.

> Is there some instance where the server would accept a `config` request, advertise their version, but not accept a new properly constructed bond?

Should not be.  But it's not hard to imagine reasons why it could happen in reality.  These include stale config responses on the client end, bugs on client and or server with serving or interpreting the config, bugs in bond creation or server-side parsing, etc.  And although we're versioning the bond formats, something subtle could change that makes a new bond incompatible.

But, we can certainly remove this unsigned bond transaction check and it wouldn't be a problem.  I would just be more comfortable with being done before locking considerable amounts for a duration on the order of months.

server/auth/registrar.go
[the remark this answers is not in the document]
chappjc: Circling back on this for clarify.  This change was made in https://github.com/decred/dcrdex/pull/1819/commits/b6bafa324d7c7bf0d6c2e6e9888b2a37e1c7ac1a

server/auth/auth.go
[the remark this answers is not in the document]
chappjc: I think this might be the last pending change I have for this PR.  Will wire up a `tierchange` handler on client that notifies, and have server send the ntfn spots like the one on which we're commenting presently.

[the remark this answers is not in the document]
chappjc: Still planning to wire up the `tierchange` notes

dex/msgjson/types.go
```diff
@@ -942,6 +974,86 @@ func (n *PenaltyNote) Serialize() []byte {
...
+	sz := len(pbr.AccountID) + len(pbr.BondID)
+	b := make([]byte, 0, sz)
+	b = append(b, pbr.AccountID...)
+	return append(b, pbr.BondID...)
+}
+
+// BondConfirmedNotification is a courtesy notification when a bond tx reaches
+// the required number of confirmations. This will not be sent for postbond
+// requests for fully-confirmed txns.
+type BondConfirmedNotification struct {
```
buck54321, line 1035: Looks like you're using a latency queue instead now? Can delete?

dex/msgjson/types.go
```diff
@@ -942,6 +974,86 @@ func (n *PenaltyNote) Serialize() []byte {
...
+// broadcasting the signed transaction so the client hasn't needlessly locked
+// funds if the transaction is rejected.
+type PostBond struct {
+	Signature
+	AcctPubKey Bytes  `json:"acctPubKey"` // acctID = blake256(blake256(acctPubKey))
+	AssetID    uint32 `json:"assetID"`
+	BondVer    uint16 `json:"bondVersion"`
+	BondTx     Bytes  `json:"bondTx"`
+	BondScript Bytes  `json:"bondScript"`
+}
```
buck54321, line 994: My instinct says we should just send the transaction ID, but since this pulls double duty as pre-validation, that doesn't work.

server/auth/registrar.go
```diff
@@ -25,7 +27,320 @@ var (
...
+// user's public key provided in the payload. If these requirements are
+// satisfied, the client will receive a PostBondResult in the response.
+//
+// Once the raw transaction is validated, the relevant asset node is used to
+// locate the transaction on the network.
+//
+//   - If it is not found, a PostBondResult is sent immediately with
+//     Confs=-1 set.  The bond is NOT stored in the DB. This was a courtesy
+//     pre-validation. The user should then broadcast the transaction, wait for
+//     the required number of confirmations, and send 'postbond' again.
```
buck54321, line 70: Why have a single route perform two rather different functions? Wouldn't it be more clear to have a separate route for pre-validation?

buck54321 (commented):
Finally got all the way through. Sorry for taking so long. First pass with some initial comments. More to come.

server/auth/auth.go
```diff
@@ -70,10 +75,24 @@ type Signer interface {
...
+// The returned account ID is from the second output. This will become a
+// multi-asset checker.
+//
+// NOTE: For DCR, and possibly all assets, the bond script is reconstructed from
+// the null data output, and it is verified that the bond output pays to this
+// script. As such, the provided bondData (redeem script for UTXO assets) is not
+// used. Maybe remove it from the function signature if no asset requires any
+// additional data.
+type BondTxParser func(assetID uint32, version uint16, rawTx, bondData []byte) (bondCoinID []byte,
+	amt int64, bondAddr string, lockTime int64, acct account.AccountID, err error)
```
buck54321, line 92: Why should `AuthManager` care about the bond address? It seems asset-specific and unrelated to the task of validation.

dex/msgjson/types.go
```diff
@@ -942,6 +974,86 @@ func (n *PenaltyNote) Serialize() []byte {
...
+// PostBond should include the unsigned bond transaction for validation prior to
+// broadcasting the signed transaction so the client hasn't needlessly locked
+// funds if the transaction is rejected.
+type PostBond struct {
+	Signature
+	AcctPubKey Bytes  `json:"acctPubKey"` // acctID = blake256(blake256(acctPubKey))
+	AssetID    uint32 `json:"assetID"`
+	BondVer    uint16 `json:"bondVersion"`
+	BondTx     Bytes  `json:"bondTx"`
+	BondScript Bytes  `json:"bondScript"`
```
buck54321, line 993: The `BondScript` can now be reconstructed from the `OP_RETURN` data, no? I don't think `AuthManager` needs this.

buck54321 (commented):
Some thoughts on the API.

dex/msgjson/types.go
chappjc: This is change on the integ branch.  It's more like BondData now, thinking about account/contract assets.  Indeed it's not used for UTXO bond assets.
When the bond-assets PR is merged I'll update both this comms PR and the integ PR.

client/core/bookie.go
```diff
@@ -799,38 +799,43 @@ func handleTradeResumptionMsg(c *Core, dc *dexConnection, msg *msgjson.Message)
...
+		return nil, fmt.Errorf("unable to fetch server config: %w", err)
+	}
+
+	// (V0PURGE) Infer if the server advertises API v0 but supports the upcoming
+	// v1 routes such as postbond. Server can only advertise a single version
+	// presently so we have to detect this indirectly.
+	apiVer := int32(cfg.APIVersion)
+	if apiVer == 0 && cfg.BondExpiry > 0 {
+		apiVer = 1
 	}
```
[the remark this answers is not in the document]
chappjc: Still breaks v0 clients, but that's something to consider.

chappjc:
Briefly going to draft until https://github.com/decred/dcrdex/pull/1818 goes in first.

chappjc:
Back out of draft with https://github.com/decred/dcrdex/pull/1818 merged.  Not re-tested with the integ branch, and I still want to add some `TierChanged` notifications server-side, but it should be alright.

server/auth/auth.go
```diff
@@ -488,15 +583,39 @@ func (auth *AuthManager) Run(ctx context.Context) {
...
+
+		for {
+			select {
+			case <-t.C:
+				auth.checkBonds()
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
```
buck54321, line 600: I wonder if it would be more efficient and accurate to prune the user's bonds at order time?

server/auth/auth.go
buck54321: Is it just so we can `if chkAddr != bondAddr ` and also use the same `FeeChecker` interface?

server/auth/registrar.go
```diff
@@ -25,7 +27,320 @@ var (
...
+			return msgjson.NewError(msgjson.BondError, "unable to check bond with node: %v", err)
+		}
+
+		// OK. Respond with active=false. They may now broadcast, wait for the
+		// required confirms, and postbond again.
+		log.Debugf("Validated prospective bond txn output %s (%s) paying %d for user %v",
+			bondStr, bondAssetSym, amt, acctID)
+		postBondRes.Active = false
+		return sendResp()
+	}
```
buck54321, line 193: This makes the latency queue somewhat useless. The latency queue is convenient because the user can submit their transaction and the communicate with the server immediately, expecting that if the server does not see the tx yet, they will wait for it. Now, if the server doesn't see the tx yet, they will assume this is a pre-validation and won't start the queue.

buck54321: Oh. Nevermind. Client must wait for confs before sending. Doof.

server/auth/auth.go
chappjc: I think I meant to remove the address return.  Need to follow up now that the bond tx structure is reworked and merged.  Will get to asap.

buck54321:
> The "central authority" you refer to is the same one where we get the config response, so I don't see anything particularly problematic in mesh with a courtesy pre-check.

When I connect my dcrd node to the P2P network, I ask every node what version they're running. When I later go to send a transaction, I don't ask other nodes if the transaction looks okay first.

chappjc:
> When I later go to send a transaction, I don't ask other nodes if the transaction looks okay first (pre-validation).

So we can't do better?  There's little cost for dex to do this, and it's handling for a very specific case.  I don't want to support locking up user funds without this.  When something doesn't work as expected and if we didn't have a precheck, and people ask why we didn't just verify they transaction was acceptable before broadcasting it, I would have no good answer.

server/auth/registrar.go
chappjc: Have given this honest consideration, and I honestly don't think so.  The machinery of the handlers would be virtually identical and whether signed or unsigned, your intend is still to *post bond*.  One less route sounds appealing to me, hence this approach.  I won't die on this hill, but I just don't feel particularly convinced or motivated to add a route.

chappjc:
In the Decred node network + tx broadcast analogy, consider the consequences of a node ultimately rejecting a transaction in the `INV` -> `GETDATA` -> `DATA` sequence.  The UTXOs aren't spend.  The broadcast failed because something is wrong with your transaction.  In our case, if the bond is rejected, we have already broadcasted (and spent UTXOs).  Now, if bond structure were somehow part of decred network consensus rules, this could be an atomic operation, but since it's not, I think we have a different situation that warrants this check before spending (and locking) funds.

server/auth/registrar.go
buck54321: Still thinking this over. One thing that stands out to me is the way we're using `PostBond`

```Go
type PostBond struct {
	Signature
	BondVer    uint16 `json:"bondVersion"`
}
```

I think that structure would make sense for pre-validation, but for actually posting the bond, it would be much more "on-model" for us to do 

```Go
type PostBond struct {
	Signature
	CoinID     Bytes  `json:"coinID"` // txid
}
```

since we can pull all of the bond details from the authoritative on-chain record. I'd almost say it's been an unwritten rule for us that we send only the minimum information required to perform the DEX functions. I've referred to this before as "minimizing the API", and I do think it's important. For one, it prevents mistakes. We can't accidentally trust what the user says. Less importantly, it reduces bandwidth.

Anyway. We can chat more here or in Element.

chappjc: I see where you're going with that.  Indented BondVer still might be important there, esp. if it indicates which contract to check (e.g. eth), but that is more on-pattern.

chappjc:
Only rebased, no substantive changes.

server/auth/auth.go
chappjc: Will look into pruning at order time, but issues I see there are (1) many orders in a narrow time period like an epoch and not checking too fast, and (2) we can't send tier change notifications closer to the time when their tier actually changes and they find out when trying to place an order.

server/db/driver/pg/system.go
```diff
@@ -199,6 +200,16 @@ type sqlQueryExecutor interface {
 	sqlExecutor
 }
 
+func createIndexStmt(db sqlQueryExecutor, fmtStmt, indexName, fullTableName string) error {
+	stmt := fmt.Sprintf(fmtStmt, indexName, fullTableName)
+	log.Debugf("Creating index %q on the %q table.", indexName, fullTableName)
+	_, err := db.Exec(stmt)
+	if errors.Is(err, sql.ErrNoRows) {
+		err = nil
```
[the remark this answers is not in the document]
chappjc: True, but this is on-pattern with the rest of the `if errors.Is(err, sql.ErrNoRows)` checks in the server/db/driver/pg package, so going to leave for now.

server/db/driver/pg/tables.go
```diff
@@ -18,6 +18,11 @@ const (
 	metaTableName     = "meta"
 	feeKeysTableName  = "fee_keys"
 	accountsTableName = "accounts"
+	bondsTableName    = "bonds"
+
+	indexBondsOnAccountName  = "idx_bonds_on_acct"
+	indexBondsOnLockTimeName = "idx_bonds_on_locktime"
+	indexBondsOnCoinIDName   = "idx_bonds_on_coinid"
```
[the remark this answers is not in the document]
chappjc: Just trying to keep the pattern as `idx_{table token}_on_{column token}` with just those 4 underscores.  Not sure exactly why, but it feels right.

chappjc:
Just rebased and added some fixup commits, but not ready for re-review.  Need to split the postbond route as per https://github.com/decred/dcrdex/pull/1819#discussion_r988311753

server/auth/registrar.go
chappjc: Suggested changes made, routes separate.

dex/msgjson/types.go
chappjc: BondData/BondScript removed from comms (and server-side code) now, although I suspect we'll need it back for ETH, but that remains to be seen.

dex/msgjson/types.go
```diff
@@ -904,6 +925,17 @@ func (c *Connect) Serialize() []byte {
...
 
+// Bond is information on a fidelity bond. This is part of the ConnectResult and
+// PostBondResult payloads.
+type Bond struct {
+	Version uint16 `json:"version"`
+	Amount  uint64 `json:"amount"`
+	Expiry  uint64 `json:"expiry"` // when it expires, not the lock time
+	CoinID  Bytes  `json:"coinID"` // NOTE: ID capitalization not consistent with other payloads, but internally consistent with assetID
+	AssetID uint32 `json:"assetID"`
+	Pending bool   `json:"pending,omitempty"` // omit when at required confs
```
chappjc, line 936: Need to remove `Pending` field

dex/msgjson/types.go
```diff
@@ -942,6 +996,131 @@ func (n *PenaltyNote) Serialize() []byte {
...
+// true, the bond was applied to the account; if false it is not confirmed, but
+// was otherwise validated.
+type PostBondResult struct {
+	Signature        // message is BondID | AccountID
+	AccountID Bytes  `json:"accountID"`
+	AssetID   uint32 `json:"assetID"`
+	Amount    uint64 `json:"amount"`
+	Expiry    uint64 `json:"expiry"` // not locktime, but time when bond expires for dex
+	BondID    Bytes  `json:"bondID"`
+	Active    bool   `json:"active"`
```
chappjc, line 1091: Need to remove the `Active field.

dex/msgjson/types.go
```diff
@@ -1080,19 +1259,41 @@ type FeeAsset struct {
 	Amt   uint64 `json:"amount"`
 }
 
+// BondAsset describes an asset for which fidelity bonds are supported.
+type BondAsset struct {
+	Version uint16 `json:"version"` // latest version supported
+	ID      uint32 `json:"id"`
+	Confs   uint32 `json:"confs"`
+	Amt     uint64 `json:"amount"` // to be implied by bond version?
```
chappjc, line 1267: Food for thought, but given variations in exchange rate, I think this has to be a variable.

dex/msgjson/types.go
```diff
@@ -1080,19 +1259,41 @@ type FeeAsset struct {
...
+	BondAssets map[string]*BondAsset `json:"bondAssets"`
+	// BondExpiry defines the duration of time remaining until lockTime below
+	// which a bond is considered expired. As such, bonds should be created with
+	// a considerably longer lockTime. NOTE: BondExpiry in the config response
+	// is temporary, removed when APIVersion reaches BondAPIVersion and we have
+	// codified the expiries for each network (main,test,sim). Until then, the
+	// value will be considered variable, and we will communicate to the clients
+	// what we expect at any given time. BondAsset.Amt may also become implied
+	// by bond version.
+	BondExpiry uint64 `json:"DEV_bondExpiry"`
```
chappjc, line 1292: Please note this BondExpiry comment.  My feeling is we can settle on a value by network consensus, perhaps implied by the API version or something, but that's not practical while we figure this out. Testnet bond specs will very likely change frequently prior to release.

server/asset/dcr/dcr.go
```diff
@@ -628,6 +631,74 @@ func (dcr *Backend) VerifyUnspentCoin(ctx context.Context, coinID []byte) error
...
+func (dcr *Backend) BondVer() uint16 {
+	return BondVersion
+}
+
+// ParseBondTx makes the package-level ParseBondTx pure function accessible via
+// a Backend instance. This performs basic validation of a serialized
+// time-locked fidelity bond transaction given the bond's P2SH redeem script.
+func (*Backend) ParseBondTx(ver uint16, rawTx []byte) (bondCoinID []byte, amt int64, bondAddr string,
+	bondPubKeyHash []byte, lockTime int64, acct account.AccountID, err error) {
+	return ParseBondTx(ver, rawTx)
```
chappjc, line 644: Pure functions are more usable in general, but for server/dex, this is convenient.

server/dex/dex.go
```diff
@@ -74,11 +77,11 @@ type DexConf struct {
 	Markets           []*dex.MarketInfo
 	Assets            []*AssetConf
 	Network           dex.Network
+	BondIncrement     uint64
```
chappjc, line 80: Need to remove this `BondIncrement` field since bond parameters a per-asset and included in `AssetConf` from markets.json

client/core/types.go
```diff
@@ -506,16 +506,21 @@ func (m *Market) marketName() string {
...
+	Confs   uint32 `json:"confs"`
+	Amt     uint64 `json:"amount"`
 }
 
+// FeeAsset is deprecated (V0PURGE), but the same as BondAsset.
+type FeeAsset BondAsset
+
 // PendingFeeState conveys a pending registration fee's asset and current
-// confirmation count.
+// confirmation count.  Deprecated (V0PURGE).
```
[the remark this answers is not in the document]
chappjc: Just occurred to me that you might have been commenting on the fields below, which are actually in the `Exchange` struct, but github's code folding kinda hid that.

dex/msgjson/types.go
```diff
@@ -942,6 +974,86 @@ func (n *PenaltyNote) Serialize() []byte {
 	return append(b, []byte(p.Details)...)
 }
 
+// Client should send bond info when their bond tx is fully-confirmed. Server
+// should start waiting for required confs when it receives the 'postbond'
+// request if the txn is found. Client is responsible for submitting 'postbond'
+// for their bond txns when they reach required confs, or when the client
+// restarts and finds bonds recorded in their DB that is flagged as neither
+// accepted nor expired (also maybe if not listed in the 'connect' response).
```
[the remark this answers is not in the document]
chappjc: Revised the comment, but not that this block comment is not part of the exported API docs.  It is intended to inform a developer.

dex/msgjson/types.go
```diff
@@ -1076,19 +1188,36 @@ type FeeAsset struct {
...
+}
+
+// ConfigResult is the successful result for the ConfigRoute. A bond expires
+// when the time remaining until lockTime is less than BondExpiry. As such,
+// bonds should be created with a considerably longer lockTime.
 type ConfigResult struct {
+	// APIVersion is the server's communications API version, but we may
+	// consider APIVersions []uint16, with versioned routes e.g. "initV2".
+	// APIVersions []uint16 `json:"apivers"`
+	APIVersion       uint16    `json:"apiver"`
```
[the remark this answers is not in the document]
chappjc: I don't have a clear picture of how this would work.  Let's address it subsequently, or in parallel if you see how it would play out now.

server/auth/auth.go
chappjc: Also worth noting that bond check/prune involves no nodes requests, just a loop through the bond checking expiry times, so it's lightning fast.

server/auth/auth_test.go
```diff
@@ -721,6 +747,15 @@ func TestConnect(t *testing.T) {
...
 	defer rig.storage.setRatioData(&ratioData{}) // clean slate
 
+	// TODO: update tests now that there is are close/ban and unban operations,
+	// instead an integral tier.
+
+	// TODO: update tests now that cancel ratio is part of the score equation
+	// rather than a hard close operation. But the cancel ratio is stupid and
+	// generally disabled in production anyway so maybe screw it.
+
+	/* cancel ratio stuff we should probably just remove
```
chappjc, line 757: Dang I forgot we decided not remove this after all.  I'll work on these tests in #1820

server/auth/auth.go
```diff
@@ -956,6 +1148,119 @@ func (auth *AuthManager) conn(conn comms.Link) *clientInfo {
...
+			log.Infof("Pruned %d expired bonds for user %v, new bond tier = %d, new trading tier = %d",
+				len(pruned), acct, bondTier, client.tier)
+			expiredBonds[acct] = checkRes{newTier, pruned}
+		}
+	}
+
+	go func() { // godspeed
+		for acct, prunes := range expiredBonds {
+			for _, bond := range prunes.bonds {
+				auth.sendBondExpired(acct, bond, prunes.tier)
```
chappjc, line 1235: Here's probably a good place to call a new `Archiver` method to DELETE rows in the bonds table.

client/core/types.go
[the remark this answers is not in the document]
buck54321: Oh yeah. That's exactly what I did.

dex/msgjson/types.go
chappjc: Note the entire solution, but in the right direction: https://github.com/decred/dcrdex/pull/1819/commits/e80d4a7ded182e6b63cdce5148ebccd7fee34c41

Still would need some kind of overlap handling like you said to avoid rejecting bonds that were confirming at the time of the Amt change.

JoeGruffins (approved):

chappjc:
Re-tested with `bond-client-integ` branch, and still functioning (UI gap aside).

buck54321 (approved):

chappjc:
Squashed and re-wrote all commit messages.  Merging after CI completes.

PR #1823, opened 2022-08-27
app: Fix profile and notification dropdowns

martonp:
The previous update to `Doc.mouseInElement` broke the profile and notification dropdowns. Using the new `Doc.mouseInElementBounds` function, which is the original implementation of `Doc.mouseInElement`, fixes them.


chappjc:
Whoops.  So the click targets were are not always descendants after all...

chappjc (approved):

PR #1827, opened 2022-08-28
core: Fix swap/redemption fee info

### LINKED ISSUE #1814: frontend reports same fees for both swap/redeem assets
image

The from/swap asset is correct, but the to/redeem asset's fees is just the other value.

martonp:
Previously, swap fees were displayed in place of redemption fees. Also, for tokens, the token asset was displayed as the fee asset instead of the parent. Both of these issues are fixed. Also, in the exported orders CSV, two new columns are added, one for the base asset's fee asset, and one for the quote asset's fee asset.

Closes #1814

### DIFF
--- a/client/asset/driver.go
+++ b/client/asset/driver.go
@@ -154,16 +154,6 @@ func DecodeCoinID(assetID uint32, coinID []byte) (cid string, err error) {
 		tkn.ParentID, dex.BipIDSymbol(tkn.ParentID), assetID, dex.BipIDSymbol(assetID))
 }
 
-// IsToken checks if the asset ID is for a token and returns the token's parent
-// ID.
-func IsToken(assetID uint32) (is bool, parentID uint32) {
-	token, is := tokens[assetID]
-	if !is {
-		return
-	}
-	return true, token.ParentID
-}
-
 // A registered asset is information about a supported asset.
 type RegisteredAsset struct {
 	ID     uint32
--- a/client/asset/driver.go
+++ b/client/asset/driver.go
@@ -154,6 +154,16 @@ func DecodeCoinID(assetID uint32, coinID []byte) (cid string, err error) {
 		tkn.ParentID, dex.BipIDSymbol(tkn.ParentID), assetID, dex.BipIDSymbol(assetID))
 }
 
+// IsToken checks if the asset ID is for a token and returns the token's parent
+// ID.
+func IsToken(assetID uint32) (is bool, parentID uint32) {
+	token, is := tokens[assetID]
+	if !is {
+		return
+	}
+	return true, token.ParentID
+}
+
 // A registered asset is information about a supported asset.
 type RegisteredAsset struct {
 	ID     uint32
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -7734,12 +7734,11 @@ func (c *Core) deleteOrderFn(ordersFileStr string) (perOrderFn func(*db.MetaOrde
 
 		baseFeeAssetSymbol := unbip(cord.BaseID)
 		baseFeeUnitInfo := baseUnitInfo
-		baseIsToken, baseParent := asset.IsToken(cord.BaseID)
-		if baseIsToken {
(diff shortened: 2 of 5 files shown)


client/core/helpers.go
```diff
@@ -38,6 +40,34 @@ func (ord *OrderReader) ToSymbol() string {
...
+
+// BaseFeeSymbol is the symbol of the asset used to pay the base asset's
+// network fees.
+func (ord *OrderReader) BaseFeeSymbol() string {
+	return ord.BaseFeeUnitInfo.Conventional.Unit
+}
+
+// QuoteFeeSymbol is the symbol of the asset used to pay the quote asset's
+// network fees.
+func (ord *OrderReader) QuoteFeeSymbol() string {
```
LasTshaMAN, line 67: maybe use these funcs to simplify `FromFeeSymbol` and `FromFeeSymbol` a bit

client/core/helpers.go
```diff
@@ -303,17 +333,17 @@ func (ord *OrderReader) AverageRateString() string {
...
+	return formatQty(ord.FeesPaid.Swap, ord.QuoteFeeUnitInfo)
 }
 
 // RedemptionFeesString is a formatted string of the paid redemption fees.
 func (ord *OrderReader) RedemptionFeesString() string {
 	if ord.Sell {
-		return formatQty(ord.FeesPaid.Swap, ord.QuoteUnitInfo)
+		return formatQty(ord.FeesPaid.Redemption, ord.QuoteFeeUnitInfo)
 	}
-	return formatQty(ord.FeesPaid.Swap, ord.BaseUnitInfo)
```
chappjc, line 316: Amazing that went unnoticed so long

JoeGruffins (commented):
Working well. These symbols are not appearing for me howerver:
image

martonp:
> Working well. These symbols are not appearing for me howerver:

The file names are case sensitive when using the embedded files but not when using the external site dir.

client/asset/driver.go
```diff
@@ -154,6 +154,16 @@ func DecodeCoinID(assetID uint32, coinID []byte) (cid string, err error) {
...
 
+// IsToken checks if the asset ID is for a token and returns the token's parent
+// ID.
+func IsToken(assetID uint32) (is bool, parentID uint32) {
+	token, is := tokens[assetID]
+	if !is {
+		return
+	}
+	return true, token.ParentID
+}
```
chappjc, line 165: Will you please move this down next to `func TokenInfo` around line 230?

Also, I think you need to follow the pattern in that function of using `driversMtx` to access the `tokens` map.

Although I think you should just use `asset.TokenInfo`.  See my other comment in core.go.

client/core/core.go
```diff
@@ -7729,14 +7731,42 @@ func (c *Core) deleteOrderFn(ordersFileStr string) (perOrderFn func(*db.MetaOrde
 		if err != nil {
 			return fmt.Errorf("unable to get base unit info for %v: %v", cord.BaseSymbol, err)
 		}
+
+		baseFeeAssetSymbol := unbip(cord.BaseID)
+		baseFeeUnitInfo := baseUnitInfo
+		baseIsToken, baseParent := asset.IsToken(cord.BaseID)
+		if baseIsToken {
```
chappjc, line 7738: Could consider just using `asset.TokenInfo` and tightening scope a little:

```go
	baseFeeAssetSymbol := unbip(cord.BaseID)
	baseFeeUnitInfo := baseUnitInfo
	if baseToken := asset.TokenInfo(cord.BaseID); baseToken != nil {
		baseFeeAssetSymbol = unbip(baseToken.ParentID)
		baseFeeUnitInfo, err = asset.UnitInfo(baseToken.ParentID)
		if err != nil {
			return fmt.Errorf("unable to get base fee unit info for %v: %v", baseToken.ParentID, err)
		}
	}
```

chappjc (approved):

chappjc:
Merging now unless you're mid review @buck54321

buck54321:
Go for it

PR #1828, opened 2022-08-28
dexc in system tray

chappjc:
image

Works in Linux, Windows, Mac OS/X (supposedly).  Only tested on Linux so far.  No cgo required in Linux.

On Windows, it can be build with `-ldflags "-H=windowsgui"` no not even open a terminal window.

Now that we have site resources embedded, this means that a Windows user can just download and extract "dexc.exe" and just double click it without any terminal interactions required.


chappjc:
The icon for the menu item (not the tray icon itself) is a little wonky on windows 7 but otherwise this is working fine.  Just need https://github.com/decred/dcrdex/pull/1829 to go first.

chappjc:
Works alright on Windows 11 too:

dex-tray-w11

TODO:
- get rid of the ugly icon on the "Launch browser" menu item because it does not respect alpha/transparency like the tray icon itself
- when clicking "Open config file" it needs logic to create it if it does not exist.  Also rename to "Edit config file"

chappjc:
> It's been a while, but can't seem to build correctly on Windows 10. I don't see any errors. attempting to run the executable does nothing. Ill try building on linux and sending the bin to windows. Any flags I need besides the `-ldflags "H=windowsgui"`?

Could you be missing the site dir before build?

JoeGruffins:
Actually it is running! I can go to `http://127.0.0.1:5758` and see the site. I however do not see it in the system tray.

chappjc:
Is it in the tray as one of the disabled/hidden icons that you have to whitelist?
That's what I saw on Windows 11.  I had to click the little "^" icon to see all the hidden tray icons

JoeGruffins:
No, I do not see it there. Sorry for slow response. Where is my task manager...

JoeGruffins:
image

chappjc:
Sadly I only have Linux, Windows 7 and Windows 11, but not Windows 10.

My only guess is that there's a W10 bug, which does seem to be common:
https://answers.microsoft.com/en-us/windows/forum/all/icons-from-deleted-programs-apppeared-in-task-bar/9fa65ff1-ee90-4935-8073-abc301eceeb5
https://winaero.com/fix-notification-area-system-tray-icons-are-messed-up-in-windows-10/

If it is really super common, that could be a showstopper though.

JoeGruffins:
Sorry, I goofed up. It is working.

image

I didn't realize there was a new dex-tray folder with a different executalbe to build. Building there works! Apologies.

chappjc:
> I didn't realize there was a new dex-tray folder with a different executalbe to build

Ahhhh!  I should have noticed that from your task manager you shared offline

JoeGruffins:
Why does there need to be a new folder? We need to keep up with the two now. Things here don't really change that much over time, but seems like we could have build flags if you wanted to build the tray support like `--tags tray` and put the initiation logic in a separate file?

chappjc:
> Why does there need to be a new folder? We need to keep up with the two now. Things here don't really change that much over time, but seems like we could have build flags if you wanted to build the tray support like `--tags tray` and put the initiation logic in a separate file?

I'm trying, perhaps a little too hard, to avoid putting these two new depends in the main module's go.mod.  I'll see how it feels merged with build tags instead.

buck54321 (approved):
Brilliant! Working beautifully on Linux. May fire up Windows later, but don't wait on me.

JoeGruffins (approved):
Working well on windows.

chappjc:
Mac:

image

**Open logs folder** works.

**Edit config file** only works if you first associate the text editor with .conf files.  Otherwise the console says:

> No application knows how to open URL file:///Users/jon/Library/Applicatoin%20Support/Dexc/dexc.conf (Error Domain=NSOSStatusErrorDomain Code=-10814 "kLSApplicationNotFoundErr: E.g. no application claims the file" UserInfo={_LSLine=1489, _LSFunction=runEvaluator}).

Perhaps a .ini extension would open without having to change file associations.  Not worth the effort now though.

PR #1829, opened 2022-08-28
windows embed paths

chappjc:
v0.5.0 on Windows is broken when using embedded files because as per the `embed` pkg docs:

> The path separator is a forward slash, even on Windows systems.

So doing `filepath.Join`, which is usually "the right thing", fails on Windows.  Surprising to say the least.  https://github.com/golang/go/issues/45230 https://github.com/golang/go/issues/44305 https://pkg.go.dev/io/fs#ValidPath

For an immediate patch.

### DIFF
--- a/client/webserver/template.go
+++ b/client/webserver/template.go
@@ -46,7 +46,8 @@ func newTemplates(folder, lang string, embedded bool) *templates {
 		t.addErr = fmt.Errorf("no translation dictionary found for lang %q", lang)
 		return t
 	}
-	if embedded {
+	if embedded { // slashes even on Windows, see embed pkg docs
+		t.folder = filepath.ToSlash(t.folder)
 		return t
 	}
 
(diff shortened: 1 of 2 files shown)


client/webserver/template.go
```diff
@@ -60,6 +61,9 @@ func newTemplates(folder, lang string, embedded bool) *templates {
 
 // filepath constructs the template path from the template ID.
 func (t *templates) filepath(name string) string {
+	if t.embedded { // slashes even on Windows, see embed pkg docs
+		return t.folder + "/" + name + ".tmpl"
```
chappjc, line 65: The general advice would be to use `path.Join` for embedded paths, instead of `filepath.Join`, but this is a very simple operation and `path.Join` is over complex for the job in this case.

buck54321 (approved):

chappjc:
One more review would be good.  Ideally, another linux, plus a mac.  Wanna double check everything this time so 0.5.1 is workable on all platforms.

PR #1830, opened 2022-08-29
client:{app,core}: Make order page more intuitive

### LINKED ISSUE #1790: UI: Orders page order is unintuitive.
This issue has two parts:

1. The order screen's swaps' order is not intuitive for taker. It will show your swap and redemption first, when in reality the other side sent first. The order it shows is 2, 1, 4, 3. Maker on the other hand will show the order 1, 2, 3.

Comments from @jholdstock on matrix:
```
future steps might be helpful as well actually, eg.

Step 1: BTC swap broadcast (them)
Step 2: DCR swap broadcast (you) (1/4 confs)
Step 3: DCR redeem (them) (pending)
Step 4 BTC redeem (you) (pending)

perhaps pending steps can be grayed or something
i think making the process a little more visible would help with user understanding and potentially reduce support questions
```

2. Refreshing will randomized the order of matches, which is disorienting. Should be oldest to newest, and if time the same alphabetical. This way the order will be the same every time you visit the page.

martonp:
Previously, the orders page displayed the actions out of order if you were the taker, and did not display the pending future events. It is now updated to show each of the swap actions in the proper order.
Screen Shot 2022-08-29 at 9 19 21 PM
Screen Shot 2022-08-29 at 9 18 43 PM

Closes #1790

### DIFF
--- a/client/core/helpers.go
+++ b/client/core/helpers.go
@@ -682,17 +682,11 @@ func (m *matchReader) ConfirmingTakerRedeem() bool {
 		m.Match.Status < order.MatchConfirmed && m.Match.Status >= order.MatchComplete
 }
 
-// InConfirmingRedeem returns true if the match has completed but the user's
-// redemption has not yet been confirmed.
-func (m *matchReader) InConfirmingRedeem() bool {
-	return m.ConfirmingMakerRedeem() || m.ConfirmingTakerRedeem()
-}
-
 // ConfirmRedeemString returns a string indicating the current confirmation
 // progress of the user's redemption. An empty string is returned if the
 // user has not yet submit their redemption, or if it is already confirmed.
 func (m *matchReader) ConfirmRedeemString() string {
-	if !m.InConfirmingRedeem() || m.Redeem == nil {
+	if !(m.ConfirmingMakerRedeem() || m.ConfirmingTakerRedeem()) || m.Redeem == nil {
 		return ""
 	}
 
--- a/client/core/helpers.go
+++ b/client/core/helpers.go
@@ -432,6 +432,105 @@ func (ord *OrderReader) MatchReaders() []*matchReader {
 	return readers
 }
 
+// IsMaker returns if the user is the maker in this match.
+func (m *matchReader) IsMaker() bool {
+	return m.Side == order.Maker
+}
+
+// MakerSwapID reutrns the ID of the asset used in the maker's swap.
+func (m *matchReader) MakerSwapID() uint32 {
+	if m.Side == order.Maker {
+		return m.ord.FromID()
+	}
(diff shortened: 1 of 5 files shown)


client/core/helpers.go
```diff
@@ -400,6 +400,105 @@ func (ord *OrderReader) MatchReaders() []*matchReader {
...
+
+// MakerSwapSymbol reutrns the symbol of the asset used in the maker's swap.
+func (m *matchReader) MakerSwapSymbol() string {
+	if m.Side == order.Maker {
+		return m.ord.FromSymbol()
+	}
+	return m.ord.ToSymbol()
+}
+
+// TakerSwapSymbol reutrns the symbol of the asset used in the taker's swap.
```
JoeGruffins, line 432: For these four comments reutrns -> returns

client/core/helpers.go
```diff
@@ -400,6 +400,105 @@ func (ord *OrderReader) MatchReaders() []*matchReader {
...
+		return m.CounterRedeem
+	}
+	return m.Redeem
+}
+
+// ShowMakerSwap returns whether or not to display the maker swap section
+// on the match card.
+func (m *matchReader) ShowMakerSwap() bool {
+	return m.MakerSwap() != nil || !m.Revoked
+}
```
JoeGruffins, line 476: I think should also not show all of these for cancel orders

image

JoeGruffins (approved):
Looks great. Just a couple comments.

client/webserver/site/src/js/order.ts
```diff
@@ -236,37 +236,68 @@ export default class OrderPage extends BasePage {
...
+      const pendingSpan = Doc.tmplElement(card, pendingName)
+      if (!coin) {
+        Doc.show(pendingSpan)
+        Doc.hide(coinLink)
+        return
+      }
       coinLink.textContent = coin.stringID
       coinLink.dataset.explorerCoin = coin.stringID
       setCoinHref(coinLink)
+      Doc.hide(Doc.tmplElement(card, pendingName))
```
chappjc, line 251: `Doc.hide(pendingSpan)`?

chappjc (approved):
LGTM.  Not merging until Buck gets eyes on it.

buck54321 (commented):
Nice. Pick up @chappjc's one comment, rebase, and its ready to go.

chappjc (approved):
Works great.  Here's the revoked pending refund change up for maker:

image

image

image

One thing I noticed was that if you have the order page opened for an epoch status order, which gets matched, the Matches section and any match boxes don't appear until refresh.  Not a new issue, at least I don't think.  Something for a follow-up.

PR #1838, opened 2022-09-06
loadboat: missing quotes

vctt94:
loadboat stopped working for me and this PR seems to fix it.


chappjc (approved):
OK, since beta_node-ctl.conf uses quotes on the `rpccert` arg like this too, but I'm curious what the error was?
I'm assuming you have spaces in your path.  Is that right? I didn't realize ini values required quoting .

JoeGruffins (approved):

Issue #1641, opened 2022-06-01
Master Issue: Documentation


buck54321:
Updates for the below items include both textual updates and more technical updates e.g. to message structures. Potential editors should feel free to update just the technical portions or to update just the text portions, as their aptitudes permit.

**If you're feeling motivated, please claim more than one at a time.**
## README
I'd like the README to be more marketing and less technical. 

**These sections should be made into wiki pages**

- [x] Client Quick Start Installation (@JoeGruffins #1715)
- [x] Client Configuration (@chappjc #1747)
- [x] Advanced Client Installation (@chappjc #1747)
- [x] Server Installation (@chappjc #1747)

## Specification
- [ ] README.mediawiki
- [ ] comms.mediawiki
- `connect` response
- [ ] fundamentals.mediawiki
- Ban score added to `config` response
- `maxFeeRate` and `swapSizeBase` updates fields added to `config` response
- [ ] admin.mediawiki
- [ ] accounts.mediawiki
- [ ] orders.mediawiki
- Order status reconciliation
- Negotiation acknowledgement changes
- [ ] api.mediawiki
- [ ] community.mediawiki  (#238)
- [ ] Need to describe trade suspension somewhere (#354)

## Wiki
[See the wiki homepage](https://github.com/decred/dcrdex/wiki) to see how to contribute here. 

- [ ] New Page: New asset integration
- [ ] New page: DEX account management. add/disable/discover/re-enable (@ukane-philemon #2468)
- [ ] DEX client backup and restoration. Can be part of clients page moved from README
- [ ] New page: Bonds, user score
- [ ] Update **Server Admin**
- Lot size / max fee rate arithmetic
- Market suspension / resumption
- [ ] **Testnet Testing** needs updating. It's tailored to RPC.
- [ ] New page: Bitcoin
- Wallet types
- Wallet configuration
- Fees and fee spikes
- [x] New page: Ethereum (@chappjc #2233)
- Wallet description
- Fees
- [ ] New page: Decred
- Mixed account setup
- [ ] New page: Tokens
- [ ] New pages maybe: Litecoin, Bitcoin Cash, Dogecoin, ZCash, Firo, Digibyte

Subsumes #707, #830, #354, #521, #238

**Please discuss. Will update based on feedback.**

xaur:
[...]

PR #1839, opened 2022-09-07
spec/comm: Update connect result.

JoeGruffins:
spec/comm: Update connect result.

    wiki: Simplify contribution instructions.

part of #1641

comm page can be viewed at https://github.com/JoeGruffins/dcrdex/blob/wikihelplesshelpful/spec/comm.mediawiki

### DIFF
--- a/spec/comm.mediawiki
+++ b/spec/comm.mediawiki
@@ -191,9 +191,23 @@ of their penalization.
 |-
 | matches || &#91;object&#93; || list of [[orders.mediawiki/#Match_negotiation|Match objects]]
 |-
-| penalty || object || a [[community.mediawiki/#Penalty_Object|Penalty Object]]. omitted if in good standing
+| order statuses || <nowiki>[object]</nowiki> || list of <code>Order Status Object</code>
 |-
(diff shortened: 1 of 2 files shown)


spec/comm.mediawiki
```diff
@@ -191,9 +191,23 @@ of their penalization.
...
+|-
+| suspended || bool || true if suspended. will be deprecated in a future version
+|-
+| sig || string || hex-encoded server's signature of the serialized connection data
+|}
+
+'''Order Status Object'''
+
+{|
+| field || type || description
```
chappjc, line 206: Should be headers like the other tables with `!` instead of `|`

chappjc (changes requested):

spec/comm.mediawiki
JoeGruffins: Ok https://github.com/decred/dcrdex/compare/76f17fac855decc14288d37c0865e68329fed2d8..b2ea39a42dca73ae454cdadd3fa003d01b02de5e

chappjc (approved):

PR #1840, opened 2022-09-07
site: reject invalid address for send or withdraw tx

ukane-philemon:
Quoting @chappjc.
>At least something. You could argue fully validated address, but really it should not go to confirm for an empty address because that's obviously crazy
Sorry, I know we've been over this, but in testing all the assets' Send functionality, this really stuck out as bad UX.  (Says "Provide a valid address" on the "Confirm" form where it's too late to provide a valid address, so you have to cancel and go back.  So why does not just just stop you from getting to "Confirm"?)

This address a UI concern and the issue on the image below.
I was trying to simulate ZEC's edge case with BCH and noticed send amount was not displayed properly.
Screenshot 2022-09-07 at 3 48 42 PM

### DIFF
--- a/client/webserver/site/src/html/wallets.tmpl
+++ b/client/webserver/site/src/html/wallets.tmpl
@@ -311,7 +311,6 @@
       <span class="d-flex d-hide justify-content-end grey fs14">
        ~<span id="balanceAfterSendFiat" class="mx-1"></span>USD
       </span>
-      <div class="fs14  errcolor mt-3" id="vSendAddrMsg">[[[invalid_address]]]</div>
       <hr class="dashed my-4">
        <div class="fs16 px-4 text-center">[[[Authorize the transfer with your app password.]]]</div>
         <div class="input-group mt-2">
--- a/client/webserver/site/src/html/wallets.tmpl
+++ b/client/webserver/site/src/html/wallets.tmpl
@@ -231,7 +231,7 @@
       <div class="mt-3">
         <label for="sendAddr" class="form-label ps-1 mb-1">[[[Address]]]</label>
         <div class="d-flex">
-         <input type="text" class="form-control col-1 select" id="sendAddr">
+         <input type="text" class="form-control col-1 select" id="sendAddr" spellcheck="false">
          <span id="validAddr" class="ico-check mx-2 pt-2 d-hide"></span>
         </div>
       </div>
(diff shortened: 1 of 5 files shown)


JoeGruffins (approved):

ukane-philemon:
I've updated this PR with a new commit that closes #1878. 
Screenshot 2022-10-01 at 8 59 19 PM
Screenshot 2022-10-01 at 8 59 10 PM

client/webserver/site/src/html/wallets.tmpl
```diff
@@ -292,26 +292,30 @@
...
       <hr class="dashed mt-2">
-      <div class="d-flex align-items-center justify-content-between">[[[estimated_fee]]]: 
-       <span id="vSendFee"></span>
+      <div id="vFeeEstimateDisplay" class="d-hide">
+        <div class="d-flex align-items-center justify-content-between">[[[estimated_fee]]]: 
+         <span id="vSendFee"></span>
+        </div>
+        <span class="d-flex d-hide justify-content-end grey fs14">
+         ~<span id="vSendFeeFiat" class="mx-1"></span>USD
+       </span>
```
buck54321, line 301: Indentation is funny here.

buck54321 (changes requested):
Please tweak the cache busters in bodybuilder.tmpl.

---

I just noticed that spell checking displays a red line under the address in the input for me. Is there a way to disable that?
image

client/webserver/site/src/js/locales.ts
```diff
@@ -123,7 +125,9 @@ export const enUS: Locale = {
   [ID_WALLET_ENABLED_MSG]: '{{ assetName }} Wallet Enabled',
   [ID_WALLET_DISABLED_MSG]: '{{ assetName }} Wallet Disabled',
   [ID_DISABLED_MSG]: 'wallet is disabled',
-  [ID_ACTIVE_ORDERS_ERR_MSG]: '{{ assetName }} wallet is actively managing orders'
+  [ID_ACTIVE_ORDERS_ERR_MSG]: '{{ assetName }} wallet is actively managing orders',
+  [ID_INVALID_ADDRESS_MSG]: 'Invalid address: {{ address }}',
+  [ID_TXFEE_UNAVAILABLE]: 'Fee estimation for this {{ assetName }} transaction is unavailable at the moment. This could be due to an error while calculating transaction fees or wallet does not currently support fee estimation.'
```
chappjc, line 130: This PR can add some brief messages on top of the changes in https://github.com/decred/dcrdex/pull/1870, but I want to go ahead with what I have there that considers the trait and skips the /txfee request entirely when it is know the wallet does not support that.

Also, with that trait check in mind, the handling changes and we don't need a long message like this that alludes to different possible reasons:
- if trait not detected -> form just says "fee estimation is not supported for this wallet type"
- if trait detected but an error is returned from /api/txfee -> show the actual error, "fee estimation failed: <reason>"

ukane-philemon: > EDIT: you can build on the traits check in https://github.com/decred/dcrdex/commit/1dbeb380fd3d95e53c85143ccf7888b02d44bd46, which I've removed from https://github.com/decred/dcrdex/pull/1870

Okay.

client/webserver/site/src/js/wallets.ts
```diff
@@ -227,59 +225,74 @@ export default class WalletsPage extends BasePage {
...
-      return
-    }
+    const addr = page.sendAddr.value || ''
+    if (addr === '') return Doc.showFormError(page.sendErr, intl.prep(intl.ID_INVALID_ADDRESS_MSG, { address: addr }))
+    const { wallet, unitInfo: ui, symbol } = app().assets[assetID]
+
+    // txfee will not be available if wallet is not a fee estimator or the
+    // request failed.
+    let txfee = 0
+    if (((wallet.traits & traitTxFeeEstimator) !== 0)) {
```
buck54321, line 240: Can remove one outside set of parentheses.

buck54321 (approved):

PR #1841, opened 2022-09-08
client: hide "load from file" if needed

### LINKED ISSUE #1805: UI: Hide load from file for spv wallets.
There is the `load from file` button on the eth reconfigure wallet form which has no function currently. It is also there when looking at the btc and dcr SPV type wallets. Is it possible to hide it whenever one is looking at an spv/built-in wallet form?

image

vctt94:
Closes #1805

### DIFF
--- a/client/webserver/site/src/js/forms.ts
+++ b/client/webserver/site/src/js/forms.ts
@@ -337,6 +337,9 @@ export class NewWalletForm {
 
     if (this.subform.dynamicOpts.children.length) Doc.show(page.walletSettingsHeader)
     else Doc.hide(page.walletSettingsHeader)
+    // A seeded or token wallet is internal to the dex client and as such does
+    // not have an external config file to select.
+    if (walletDef.seeded || Boolean(this.current.asset.token)) Doc.hide(this.subform.fileSelector)
 
     this.refresh()
     await this.loadDefaults()
--- a/client/webserver/site/src/js/wallets.ts
+++ b/client/webserver/site/src/js/wallets.ts
@@ -719,10 +719,10 @@ export default class WalletsPage extends BasePage {
 
   updateDisplayedReconfigFields (walletDef: WalletDefinition) {
     if (walletDef.seeded || walletDef.type === 'token') {
-      Doc.hide(this.page.showChangePW)
+      Doc.hide(this.page.showChangePW, this.reconfigForm.fileSelector)
       this.changeWalletPW = false
       this.setPWSettingViz(false)
-    } else Doc.show(this.page.showChangePW)
+    } else Doc.show(this.page.showChangePW, this.reconfigForm.fileSelector)
   }
 
   /* Display a deposit address. */


client/webserver/site/src/js/wallets.ts
```diff
@@ -723,6 +726,13 @@ export default class WalletsPage extends BasePage {
       this.changeWalletPW = false
       this.setPWSettingViz(false)
     } else Doc.show(this.page.showChangePW)
+
+    // if wallet not loadable from file, hide file selector, else shows it.
+    if (walletNotLoadableFromFile.includes(walletDef.type)) {
+      Doc.hide(this.reconfigForm.fileSelector)
+    } else {
+      Doc.show(this.reconfigForm.fileSelector)
```
chappjc, line 734: I figured we'd do this above in the `if (walletDef.seeded || walletDef.type === 'token')` condition that we use to hide tint wait password form.
Is this conceptually different in some way that would require a separate whitelist?

vctt94: I don't believe so, the only reason I did not implemented it, inside the other if is to avoid nesting ifs, but I guess if it the wallet is seeded or token, it is enough for hiding the "load from file" link?

chappjc: At present, "seeded" implies native/built-in, which means there's never going to be an external wallet config file.  So, yes, I think `seeded` and the token condition are what we're after.

JoeGruffins (commented):
Looking at my rpc wallet on simnet, it seems to default it to SPV in the dropdown which is unexpected and incorrect I believe as the current wallet is rpc. I have to click rpc once and then SPV to get the SPV form. It should default on what the current wallet is:
image

vctt94:
Thanks joe, should be fixed now.

JoeGruffins (approved):

client/webserver/site/src/js/wallets.ts
```diff
@@ -722,7 +722,11 @@ export default class WalletsPage extends BasePage {
       Doc.hide(this.page.showChangePW)
       this.changeWalletPW = false
       this.setPWSettingViz(false)
-    } else Doc.show(this.page.showChangePW)
+      Doc.hide(this.reconfigForm.fileSelector)
+    } else {
+      Doc.show(this.page.showChangePW)
```
martonp, line 727: You can do `Doc.show(this.page.showChangePW, this.reconfigForm.fileSelector)`.

martonp (approved):

client/webserver/site/src/js/wallets.ts
chappjc: Good point.  @vctt94 if you combine them like this, we can merge the PR now.

client/webserver/site/src/js/forms.ts
```diff
@@ -87,8 +87,6 @@ export class NewWalletForm {
     // WalletConfigForm will set the global app variable.
     this.subform = new WalletConfigForm(page.walletSettings, true)
 
-    Doc.bind(this.subform.showOther, 'click', () => Doc.show(page.walletSettingsHeader))
```
chappjc, line 90: This is slightly too heavy handed to hide the load file icon because that's just a _part_ of the "Wallet Settings" header, which is still applicable when you expand the additional settings that seeded wallets still have.

```diff
@@ -336,6 +338,10 @@ export class NewWalletForm {
     if (this.subform.dynamicOpts.children.length) Doc.show(page.walletSettingsHeader)
     else Doc.hide(page.walletSettingsHeader)
 
+    // A seeded or token wallet is internal to the dex client ad as such does
+    // not have an external config file to select.
+    if (walletDef.seeded || Boolean(this.current.asset.token)) Doc.hide(this.subform.fileSelector)
+
     this.refresh()
     await this.loadDefaults()
   }
```

client/webserver/site/src/js/wallets.ts
```diff
@@ -722,7 +722,10 @@ export default class WalletsPage extends BasePage {
       Doc.hide(this.page.showChangePW)
       this.changeWalletPW = false
       this.setPWSettingViz(false)
-    } else Doc.show(this.page.showChangePW)
+      Doc.hide(this.reconfigForm.fileSelector)
```
chappjc, line 725: can combine the `Doc.hide`s too

PR #1844, opened 2022-09-14
dcrdex Ar

Insaf01:


chappjc:
Thanks @Insaf01!  This mostly covers step 1 (HTML strings) described https://github.com/decred/dcrdex/wiki/Localization-and-Translation

For the other two language domains with translated strings:
- https://github.com/decred/dcrdex/blob/master/client/core/locale_ntfn.go
- https://github.com/decred/dcrdex/blob/master/client/webserver/site/src/js/locales.ts

Sorry to have the translations in three different spots.  We'll eventually simplify that, but it's a challenge.

Insaf01:
Thank you @chappjc, sorry for the late reply, I've add the translation for the two other files

chappjc:
> Thank you @chappjc, sorry for the late reply, I've add the translation for the two other files

Can you check your git workspace?  This branch doesn't include changes to the other two files.  https://github.com/decred/dcrdex/pull/1844/files

chappjc:
Thanks @Insaf01.  I'm going to put these changes on a different branch and open a new PR with some minor changes.  I think this is everything that's needed.

chappjc:
All set.  Rebased and squashed into https://github.com/decred/dcrdex/pull/1898.  You've done all the hard work @Insaf01 and you're the git commit is yours in that PR.  I'll just have to tweak as needed so it builds and runs as intended.  I'll message you if there's anything else to add.

PR #1845, opened 2022-09-14
client: No hide header when having additional settings

vctt94:
This PR is build on top of #1841

This PR is based in this matrix chat: https://matrix.to/#/!SFRQQFIHUUNXARfvew:decred.org/$Yy4D-nIYqum6RkurYGH6RDPz3WnvbzRBh85xZjow06o?via=decred.org&via=matrix.org&via=planetdecred.org

Where @chappjc mentioned the `Wallet Settings` header appearing and not disappearing was a small bug, but later said he feels like it should always show, when availble. 

I think that's reasonable, so this PR removes hiding the wallet settings header. The wallet setting header starts as hided. It will always show, if available. 

New behavior: 

Peek 14-09-2022 16-42

### DIFF
--- a/client/webserver/site/src/js/forms.ts
+++ b/client/webserver/site/src/js/forms.ts
@@ -87,8 +87,6 @@ export class NewWalletForm {
     // WalletConfigForm will set the global app variable.
     this.subform = new WalletConfigForm(page.walletSettings, true)
 
-    Doc.bind(this.subform.showOther, 'click', () => Doc.show(page.walletSettingsHeader))
-
     bind(form, page.submitAdd, () => this.submit())
     bind(form, page.oneBttn, () => this.submit())
 
@@ -316,6 +314,8 @@ export class NewWalletForm {
     const noWalletPWNeeded = walletDef.seeded || Boolean(this.current.asset.token)
     if (appPwCached && noWalletPWNeeded) {
       Doc.show(page.oneBttnBox)
+      // no required configuration for wallet.
+      Doc.hide(page.walletSettingsHeader)
     } else if (noWalletPWNeeded) {
       Doc.show(page.auth)
       page.newWalletPass.value = ''
@@ -335,7 +335,7 @@ export class NewWalletForm {
       })
     } else this.subform.update(configOpts, {})
 
-    if (this.subform.dynamicOpts.children.length) Doc.show(page.walletSettingsHeader)
+    if (this.subform.dynamicOpts.attributes.length) Doc.show(page.walletSettingsHeader)
     else Doc.hide(page.walletSettingsHeader)
     // A seeded or token wallet is internal to the dex client and as such does
     // not have an external config file to select.
--- a/client/webserver/site/src/js/forms.ts
+++ b/client/webserver/site/src/js/forms.ts
@@ -335,7 +335,9 @@ export class NewWalletForm {
       })
     } else this.subform.update(configOpts, {})
 
-    if (this.subform.dynamicOpts.attributes.length) Doc.show(page.walletSettingsHeader)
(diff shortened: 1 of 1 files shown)


buck54321:
This is for the `oneBttn`, not shown in your video because you don't have the password cached. That's the primary case for hiding it, since we are trying to super-simplify the UI in that case to avoid any chance of user confusion. In any case, this doesn't need to be changed, IMO.

chappjc:
What he's trying to address (in the second commit, first is already merged) is something else.  Here's the explanation in a vid:

https://user-images.githubusercontent.com/9373513/190867799-08d87c68-b7fa-4d1d-a0c2-51b33dedbb2c.mp4

It's hardly a bug really, just a quirk.

The "Wallet Settings" header only appears when you expand the additional settings.  When you collapse/hide them, the Wallet Settings header stays.  I don't really understand why it's not show always, but there are two ways to make it consistent: always show it or hide it when you collapse the additional settings.

buck54321:
```Javascript
if (this.subform.dynamicOpts.children.length) Doc.show(page.walletSettingsHeader)
```

It's hidden when there are no fields shown. So we wouldn't `hide it when you collapse the additional settings` unless there were no fields in the visible section.

chappjc:
> unless there were no fields in the visible section.

yeah... exactly.  See the vid.

buck54321:
> > unless there were no fields in the visible section.
> 
> yeah... exactly. See the vid to the end when the additional settings are collapsed. It is not re-hidden.
> 
> Although, I think it's perfectly fine to show the header always as long as there are additional settings, collapsed or not.

I was replying to this comment

> always show it, or hide it when you collapse the additional settings.

to point out that you wouldn't hide it if there are visible fields.

As long as it's hidden when the `oneBttn` is shown (= no required configuration), I'm good. I see the inconsistency.

chappjc:
OK.  For ref, here's the create wallet form from the wallet's page when logged in *without saving the password* followed by a page refresh (the `oneBttn` not shown because *app pass*, not wallet pass, is needed):

image

-->

image

-->

image

So, I think `noWalletPWNeeded` **plus** `this.subform.dynamicOpts.children.length` is the condition to consider.

client/webserver/site/src/js/forms.ts
```diff
@@ -335,7 +335,7 @@ export class NewWalletForm {
       })
     } else this.subform.update(configOpts, {})
 
-    if (this.subform.dynamicOpts.children.length) Doc.show(page.walletSettingsHeader)
+    if (this.subform.dynamicOpts.attributes.length) Doc.show(page.walletSettingsHeader)
```
vctt94, line 338: Children might be 0, as they will be hidden.

buck54321: `children` doesn't consider visibility or display, and `attributes` does not count descendants. Wanna check this again?

vctt94: You are correct. The Reason is because the children is at defaultSettings, because gap limit have a default value.

So now I am also considerent defaultSettings children for showing the wallet settings header.

client/webserver/site/src/js/forms.ts
```diff
@@ -316,6 +314,8 @@ export class NewWalletForm {
     const noWalletPWNeeded = walletDef.seeded || Boolean(this.current.asset.token)
     if (appPwCached && noWalletPWNeeded) {
       Doc.show(page.oneBttnBox)
+      // no required configuration for wallet.
+      Doc.hide(page.walletSettingsHeader)
```
buck54321, line 318: I don't think this does anything because it'll be redone below anyway.

buck54321 (approved):

PR #1846, opened 2022-09-14
dex: use spec defined epoch duration if provided value is not valid

### LINKED ISSUE #300: dcrdex config should set a default epoch duration
Fair enough, will check for 0 there.  Also seems reasonable to replace 0/-1/omitted with a the specd default.

_Originally posted by @chappjc in https://github.com/decred/dcrdex/pull/283_

ukane-philemon:
This diff ensures a market's epoch duration is set to the [spec-defined epoch duration](https://github.com/decred/dcrdex/blob/master/spec/admin.mediawiki#exchange-variables) if the config epoch duration is invalid. Closes #300.

### DIFF
--- a/dex/market.go
+++ b/dex/market.go
@@ -9,6 +9,12 @@ import (
 	"strings"
 )
 
+const (
+	// defaultEpochDuration is the spec defined default markets epoch duration
+	// in milliseconds.
+	defaultEpochDuration uint64 = 60000
+)
+
 // MarketInfo specifies a market that the Archiver must support.
 type MarketInfo struct {
 	Name                   string
@@ -53,6 +59,12 @@ func NewMarketInfo(base, quote uint32, lotSize, rateStep, epochDuration uint64, 
 	if err != nil {
 		return nil, err
 	}
+
+	// Check for sensible epoch duration.
+	if epochDuration <= 0 {
+		epochDuration = defaultEpochDuration
+	}
+
 	return &MarketInfo{
 		Name:                   name,
 		Base:                   base,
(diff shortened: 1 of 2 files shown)


dex/market.go
```diff
@@ -53,6 +59,12 @@ func NewMarketInfo(base, quote uint32, lotSize, rateStep, epochDuration uint64,
 	if err != nil {
 		return nil, err
 	}
+
+	// Check for sensible epoch duration.
+	if epochDuration <= 0 {
```
JoeGruffins, line 64: nit: uint64 cannot be < 0

JoeGruffins (approved):

dex/market.go
ukane-philemon: > nit: uint64 cannot be < 0

You mean to change the comment, right?

chappjc: He means you only need to check for `== 0`.

dex/market.go
```diff
@@ -9,6 +9,12 @@ import (
 	"strings"
 )
 
+const (
+	// defaultEpochDuration is the spec defined default markets epoch duration
+	// in milliseconds.
+	defaultEpochDuration uint64 = 60000
```
chappjc, line 15: Turns out that 60 sec is an extremely long epoch.  I think we should revise the spec to make it 10 sec, 20 tops.

ukane-philemon: > I think we should revise the spec to make it 10 sec, 20 tops.

20 sec should be good.

chappjc (approved):

PR #1847, opened 2022-09-15
client: hide accelerate button after first maker's swap

### LINKED ISSUE #1797: accelerate icon remains on your orders table of markets page after confs
The rocket icon for a settling order remains even after the first confirmation.  It goes away when you refresh.

vctt94:
closes #1797

### DIFF
--- a/client/webserver/site/src/js/markets.ts
+++ b/client/webserver/site/src/js/markets.ts
@@ -57,7 +57,8 @@ import {
   ConnEventNote,
   OrderOption,
   ConnectionStatus,
-  RecentMatch
+  RecentMatch,
+  MatchNote
 } from './registry'
 
 const bind = Doc.bind
@@ -471,6 +472,7 @@ export default class MarketsPage extends BasePage {
     // Notification filters.
     app().registerNoteFeeder({
       order: (note: OrderNote) => { this.handleOrderNote(note) },
+      match: (note: MatchNote) => { this.handleMatchNote(note) },
       epoch: (note: EpochNote) => { this.handleEpochNote(note) },
       conn: (note: ConnEventNote) => { this.handleConnNote(note) },
       balance: (note: BalanceNote) => { this.handleBalanceNote(note) },
@@ -1835,6 +1837,13 @@ export default class MarketsPage extends BasePage {
     // update local dex
     this.market.dex = app().exchanges[dexAddr]
     this.setRegistrationStatusVisibility()
+  }
+
+  handleMatchNote (note: MatchNote) {
+    const mord = this.metaOrders[note.orderID]
+    if (!mord) return this.refreshActiveOrders()
+    if (app().canAccelerateOrder(mord.ord)) Doc.show(mord.details.accelerateBttn)
+    else Doc.hide(mord.details.accelerateBttn)
   }
 
   /*


client/core/types.go
```diff
@@ -253,9 +253,7 @@ func matchFromMetaMatchWithConfs(ord order.Order, metaMatch *db.MetaMatch, swapC
 	}
 	if len(swapCoin) > 0 {
 		swap = NewCoin(fromID, swapCoin)
-		if (side == order.Taker && status == order.TakerSwapCast) ||
```
vctt94, line 256: Is there a reason for not sending swap information if not on this condition? 

On this issue, after sending the first swap as a maker, the status will be TakerSwapCast and the information will not arrive at the site.

vctt94:
I noticed some inconsistent behavior sometimes with a bitcoin block being mined, but the swap confirmations of the match not been updated, which causes the accelerate button not to be hidden. 

I am still not sure why this happens not how, but would appreciate if someone else notices it, let me know.

client/core/types.go
chappjc: > On this issue, after sending the first swap as a maker, the status will be TakerSwapCast and the information will not arrive at the site.

It would be MakerSwapCast.

I always see the conf count update, so I'm confused.

vctt94: After the first maker swap, the order status is not changed to TakerSwapCast?

I am trying to find a reasonable way to reproduce the conf update bug

vctt94: nvm, probably some state I got into. Wasn't able to reproduce the issue on master

client/webserver/site/src/js/app.ts
```diff
@@ -930,12 +921,25 @@ function setSeverityClass (el: HTMLElement, severity: number) {
   el.classList.add(severityClassMap[severity])
 }
 
+// Updates given order in market's orders list if it finds it.
+// Returns a bool which indicates if order was found.
+function updateOrder (mkt: Market, ord: Order) {
```
JoeGruffins, line 926: I'm a noob at TypeScript but should this have the return type `boolean` at the end? Does it work correctly without it?

JoeGruffins (approved):
I thought there was an accelerate button on the orders page, but I also don't see it on master. Was it removed?

client/webserver/site/src/js/app.ts
vctt94: typescript is able to tell by itself the return type by the function return. From the docs:

`We can add types to each of the parameters and then to the function itself to add a return type. TypeScript can figure the return type out by looking at the return statements, so we can also optionally leave this off in many cases.`

https://www.typescriptlang.org/docs/handbook/functions.html

But return type added. Thanks for the catch

buck54321 (commented):
I'm seeing some odd behavior. 

1. If I have ***/order*** open before the maker matches, the button will be displayed when the match occurs, but won't be hidden when the first confirmation happens.
2. If I **don't** have ***/order*** open **before** the maker matches, the button won't be displayed at all once I navigate there, even though there are zero confs and the little button in ***/markets*** is visible.

JoeGruffins:
> I'm seeing some odd behavior.

I think this pr is only addressing the rocket button on markets. But yeah the other button on the orders page is messed up. I also commented here to @martonp https://github.com/decred/dcrdex/pull/1864#issuecomment-1257548313 since he is working in that area.

chappjc:
Yeah I think #1864 should aim to resolve those issues with the order page.

client/webserver/site/src/js/app.ts
```diff
@@ -576,8 +565,10 @@ export default class Application {
       }
       case 'match': {
         const n = note as MatchNote
-        const ord = this.order(n.orderID)
-        if (ord) updateMatch(ord, n.match)
+        const order = this.order(n.orderID)
+        if (!order) return
```
buck54321, line 569: Are we certain that we don't want to pass this note along to the `noteReceivers` anyway?

vctt94: I believe we can, but right now nothing is done by it. Should we send it to the bell, or do you have something in mind for it?

buck54321: > Should we send it to the bell,

I'm not sure what you mean. I'm talking about subscribers to order notes. It is expected that all notes received are passed the `noteReceivers`. I don't see any reason to break that assumption here.

chappjc: @vctt94 Try to push changes in such a way that we can always see what changed in just this PR.  The force-push diff includes a mass of other changes unrelated to this PR.  I can see this `return` has become a `break` if I look at the full PR diff again, but it's much more convenient for each update to have a diff viewable somewhere.

client/webserver/site/src/js/markets.ts
```diff
@@ -458,6 +458,7 @@ export default class MarketsPage extends BasePage {
     // Notification filters.
     app().registerNoteFeeder({
       order: (note: OrderNote) => { this.handleOrderNote(note) },
+      match: () => { this.refreshActiveOrders() },
```
buck54321, line 461: This is what fixes it, but you can't do this. This replaces every user order div, which means if the user has details expanded, they will all close on every single update. It's also just way to blunt. You'll need to find the relevant `data-tmpl` element.

client/webserver/site/src/js/app.ts
```diff
@@ -934,12 +925,25 @@ function setSeverityClass (el: HTMLElement, severity: number) {
...
+}
+
 /* updateMatch updates the match in or adds the match to the order. */
-function updateMatch (order: Order, match: Match) {
+function updateMatch (order: Order, match: Match, mkt: Market) {
   for (const i in order.matches) {
     const m = order.matches[i]
     if (m.matchID === match.matchID) {
       order.matches[i] = match
+      updateOrder(mkt, order)
```
buck54321, line 946: Hmm. Does this do anything? `updateMatch` is only used once, in the `'match'` notification handler. The order there is from 
```Javascript
const order = this.order(n.orderID)
```
which itself does, 
```Javascript
order (oid: string): Order | null {
    for (const xc of Object.values(this.user.exchanges)) {
      for (const market of Object.values(xc.markets)) {
        if (!market.orders) continue
        for (const ord of market.orders) {
          if (ord.id === oid) return ord
        }
      }
    }
    return null
  }
```
, and the market passed in is similarly from the map
```Javascript
const mkt = this.user.exchanges[order.host].markets[order.market]
```

These are all reference values, so you're simply replacing an order with itself, with no net effect, I think.

The thing is, this line is really the only change here, so if this does nothing, all of the changes in this file do nothing.

buck54321 (changes requested):

client/webserver/site/src/js/app.ts
vctt94: That's correct. I thought the reason the value wasn't updating could be because it's reference was not updating, but I was wrong.

Thanks for pointing that out.

client/webserver/site/src/js/markets.ts
```diff
@@ -1165,9 +1167,7 @@ export default class MarketsPage extends BasePage {
   /* refreshActiveOrders refreshes the user's active order list. */
   refreshActiveOrders () {
     const page = this.page
-    const metaOrders = this.metaOrders
     const market = this.market
-    for (const oid in metaOrders) delete metaOrders[oid]
```
vctt94, line 1170: deleting metaOrders[oid] reference now on `refreshActiveOrder`

vctt94 (commented):
Thanks for the review buck.

I added now a new `refreshActiveOrder` function, which update orders, without changing other order's properties.

client/webserver/site/src/js/markets.ts
```diff
@@ -1837,6 +1853,19 @@ export default class MarketsPage extends BasePage {
...
 
+  /*
+   * handleMatchNote is the handler for the 'match'-type notification, which are
+   * used to update a user's match's status.
+   */
+  handleMatchNote (note: MatchNote) {
+    const market = this.market
+    const orders = app().orders(market.dex.host, marketID(market.baseCfg.symbol, market.quoteCfg.symbol))
+    const order = orders.find((ord: Order) => ord.id === note.orderID)
+    if (!order) return
```
martonp, line 1864: The only thing that would change on this page due to a match note is showing/hiding the acceleration button. The only change you need is adding this new `handleMatchNote` function, and adding the code below:
``` js
    const mord = this.metaOrders[order.id]
    if (!mord) return this.refreshActiveOrders()
    mord.ord = order
    if (app().canAccelerateOrder(order)) Doc.show(mord.details.accelerateBttn)
    else Doc.hide(mord.details.accelerateBttn)
```

The better way to solve this would probably be to send an order note when the order becomes able to be accelerated, or stops being able to be accelerated. I think that actually makes more sense, since we're dealing with orders on this page, and this is a situation where the state of an order has changed.

vctt94: Thanks for the suggestion @martonp . I am now sending a new order notification after mining a block.

Right now I am keeping the match notification, because it is useful on the order page. 

Do you think would it be good to remove the match notification and keep only the new order notification when confirming new blocks?

client/webserver/site/src/js/app.ts
vctt94: sorry for the latest force push, but I am taking now a different approach as suggested by marton.

client/core/trade.go
```diff
@@ -1513,6 +1515,8 @@ func (t *trackedTrade) isRedeemable(ctx context.Context, match *matchTracker) (r
 				" be network latency.", match)
 		}
 		match.setSwapConfirms(int64(confs))
+		cord := t.coreOrder()
```
chappjc, line 1518: I think these are going to deadlock.  `coreOrderInternal`?

Also can skip the `cord` variable.

And could use a few words in a trailing comment about the intent.

client/webserver/site/src/js/markets.ts
chappjc: If we go with the backend solution, consider:

An order **becomes** able to be accelerated when user's own swap/init is just broadcasted. For both maker and taker, it looks like an order note is already sent after the call to `c.swapMatches(t, swaps)` in `tick`.  :heavy_check_mark: 

An order **stops** being able to be accelerated when the user's own swap/init first reaches 1 confirm.
- For maker, this is detected in `isSwappable` in the `case order.MakerSwapCast:` with `match.Side == order.Maker`.
- For taker, this is detected in `isRedeemable`in the `case order.TakerSwapCast` with `match.Side == order.Taker`.

The counterparty's swap or redeem is not relevant in either case.

So it seems like your previous commit was more correct.  I think what is missing is just a condition to only send the order note when it hits 1 conf.

Does it appear to be working correctly with your latest revision.  If so, I'm confused, because confs on the counterparty's swap does not seem relevant to this.

chappjc: > I think what is missing is just a condition to only send the order note when it hits 1 conf.

Regarding this, perhaps a small modification to `(*matchTracker).setSwapConfirms`:

```diff
diff --git a/client/core/trade.go b/client/core/trade.go
index 0cca3002..ec7da7ab 100644
--- a/client/core/trade.go
+++ b/client/core/trade.go
@@ -164,8 +164,8 @@ func (m *matchTracker) confirms() (mine, theirs int64) {
        return atomic.LoadInt64(&m.swapConfirms), atomic.LoadInt64(&m.counterConfirms)
 }
 
-func (m *matchTracker) setSwapConfirms(mine int64) {
-       atomic.StoreInt64(&m.swapConfirms, mine)
+func (m *matchTracker) setSwapConfirms(mine int64) (was int64) {
+       return atomic.SwapInt64(&m.swapConfirms, mine)
 }
 
 func (m *matchTracker) setCounterConfirms(theirs int64) (was int64) {
@@ -1283,10 +1283,12 @@ func (t *trackedTrade) isSwappable(ctx context.Context, match *matchTracker) (re
                                "but we have not seen the counter-party's redemption yet. This could just"+
                                " be network latency.", match)
                }
-               match.setSwapConfirms(int64(confs))
-               cord := t.coreOrder()
-               t.notify(newOrderNote(TopicConfirms, "", "", db.Data, cord))
-               t.notify(newMatchNote(TopicConfirms, "", "", db.Data, t, match))
+               was := match.setSwapConfirms(int64(confs))
+               if was == 0 && confs > 0 {
+                       t.notify(newOrderNote(TopicConfirms, "", "", db.Data, t.coreOrderInternal())) // includes all current match data
+               } else {
+                       t.notify(newMatchNote(TopicConfirms, "", "", db.Data, t, match))
+               }
                return false, false
        }
 
@@ -1514,10 +1516,12 @@ func (t *trackedTrade) isRedeemable(ctx context.Context, match *matchTracker) (r
                                "but we have not seen the counter-party's redemption yet. This could just"+
                                " be network latency.", match)
                }
-               match.setSwapConfirms(int64(confs))
-               cord := t.coreOrder()
-               t.notify(newOrderNote(TopicConfirms, "", "", db.Data, cord))
-               t.notify(newMatchNote(TopicConfirms, "", "", db.Data, t, match))
+               was := match.setSwapConfirms(int64(confs))
+               if was == 0 && confs > 0 {
+                       t.notify(newOrderNote(TopicConfirms, "", "", db.Data, t.coreOrderInternal())) // includes all current match data
+               } else {
+                       t.notify(newMatchNote(TopicConfirms, "", "", db.Data, t, match))
+               }
                return false, false
 
        case order.MakerRedeemed:
```

Not tested, just thinking.

chappjc: Still, an order note that includes all match data feels a little brute force.  I'm not totally convinced that this is better than updating the frontend's match note handling as per https://github.com/decred/dcrdex/pull/1847#discussion_r1001917323

chappjc: I believe what @martonp was getting at is this:

```diff
diff --git a/client/webserver/site/src/js/markets.ts b/client/webserver/site/src/js/markets.ts
index e0eebf21..0a520319 100644
--- a/client/webserver/site/src/js/markets.ts
+++ b/client/webserver/site/src/js/markets.ts
@@ -50,6 +50,7 @@ import {
   SpotPriceNote,
   FeePaymentNote,
   OrderNote,
+  MatchNote,
   EpochNote,
   BalanceNote,
   MiniOrder,
@@ -471,6 +472,7 @@ export default class MarketsPage extends BasePage {
     // Notification filters.
     app().registerNoteFeeder({
       order: (note: OrderNote) => { this.handleOrderNote(note) },
+      match: (note: MatchNote) => { this.handleMatchNote(note) },
       epoch: (note: EpochNote) => { this.handleEpochNote(note) },
       conn: (note: ConnEventNote) => { this.handleConnNote(note) },
       balance: (note: BalanceNote) => { this.handleBalanceNote(note) },
@@ -1860,6 +1862,17 @@ export default class MarketsPage extends BasePage {
       (oldStatus === OrderUtil.StatusBooked && order.status > OrderUtil.StatusBooked)) this.setDepthMarkers()
   }
 
+  /* handleMatchNote handles a 'match' notification. The sole purpose of this is
+   * to show or hide the accelerate order icon, which depends on the status of
+   * an order's matches.
+   */
+  handleMatchNote (note: MatchNote) {
+    const mord = this.metaOrders[note.orderID]
+    if (!mord) return this.refreshActiveOrders()
+    if (app().canAccelerateOrder(mord.ord)) Doc.show(mord.details.accelerateBttn)
+    else Doc.hide(mord.details.accelerateBttn)
+  }
+
   /*
    * handleEpochNote handles notifications signalling the start of a new epoch.
    */
```

chappjc: I don't think we should send full order notes for this anymore though.  I think we should just handle the match notes that are already sent to the UI properly on the /markets, as we do on the /order page.  https://github.com/decred/dcrdex/pull/1847#discussion_r1007414247

martonp (approved):
LGTM

PR #1849, opened 2022-09-18
client/core: order status sync after preimage miss 

chappjc:
When reconciling an order that was in epoch status when the connection was lost and not restored before the preimage request (a preimage miss), be sure to unlock the funding coins when reconciling the status of the order to revoked.

There are two variants of this failure mode:
- Client reconnects after that and the order status is already **revoked**.  The first commit that adds a missing `trade.maybeReturnCoins` to the `updateOrder` closure addresses this.
- Client reconnects while the preimage collection is still active and the order status is still **epoch**, but they've still missed their preimage request (server doesn't re-send it).  This is addressed by flagging epoch orders in handleReconnect for a status check after a delay.

This PR also adds some additional edge case handling to `reconcileTrades`.  One fix handles a malicious or malfunctioning server that returns different orders than requested in the order status result.  The other deals with orders that get no order status result from the server.

PR #1850, opened 2022-09-19
[release-v0.5] backports for v0.5.3

chappjc:
These plus https://github.com/decred/dcrdex/pull/1849 will be v0.5.3.

Note that this PR also bumps the version strings to [v0.5.3](https://github.com/decred/dcrdex/milestone/23).
It also regenerates the static site resources, including the webpack bundle, since this is the release branch.


client/core/core.go
```diff
@@ -3565,6 +3567,54 @@ func (c *Core) Login(pw []byte) (*LoginResult, error) {
...
+	//  return loginResult(), err
+	// }
+	defer func() { <-c.loginSlot }() // remove the peg from the slot when done
+	select {
+	case c.loginSlot <- struct{}{}: // full login
+	default:
+		c.log.Warnf("Login already running! Waiting for it to complete.")
+		// Just wait and take the short path.
+		c.loginSlot <- struct{}{}
+		return loginResult(), err
```
chappjc, line 3606: This contraption has not been giving any trouble on master since it was merged on Aug 22, so I think we can use this on the release branch as well to prevent concurrent logins from doing bad things.

client/asset/btc/btc.go
```diff
@@ -1344,7 +1372,21 @@ func (btc *baseWallet) legacyBalance() (*asset.Balance, error) {
...
 	feeResult, err := btc.node.estimateSmartFee(int64(confTarget), &btcjson.EstimateModeConservative)
 	if err != nil {
-		return 0, err
+		btc.log.Errorf("Failed to get fee rate with estimate smart fee rate: %v", err)
+
+		if !btc.apiFeeFallback() {
+			return 0, err
+		}
+		btc.log.Debug("Retrieving fee rate from external API: ", externalApiUrl)
+		estimatedFee, err := externalFeeEstimator(btc.ctx, btc.Network)
```
chappjc, line 1381: Well.... considering https://github.com/decred/dcrdex/issues/1618#issuecomment-1248122923 (@buck54321) we may want to drop this commit from the backport batch.

client/asset/dcr/dcr.go
```diff
@@ -506,29 +506,36 @@ type feeStamped struct {
...
 	redeemConfTarget uint64
-	useSplitTx       bool
 	apiFeeFallback   bool
+}
+
+// ExchangeWallet is a wallet backend for Decred. The backend is how the DEX
+// client app communicates with the Decred blockchain and wallet. ExchangeWallet
+// satisfies the dex.Wallet interface.
+type ExchangeWallet struct {
+	cfgV atomic.Value // *exchangeWalletConfig
```
chappjc, line 529: This particular change might be too bold for the release branch.  Thoughts?

buck54321 (commented):
I'm good with all of this. Might consider #1852 too. I'll test more thoroughly later today.

chappjc:
> Might consider #1852 too

Turns out the bug fixed by https://github.com/decred/dcrdex/pull/1852 does not apply to release-v0.5 because it exported `btc.CommonConfigOpts` helper was introduced in the BCH SPV work that was merged after 5fed412bfa07e56b1a5d4f3d298dd7ae134e19d4

So it looks like we just need to get https://github.com/decred/dcrdex/pull/1849 in v0.5.3 and it's g2g.
My fix for https://github.com/decred/dcrdex/issues/952 should be in the next release.

PR #1859, opened 2022-09-22
build: Use targets and loop in pkg.sh.

davecgh:
This modifies the packaging script to make use of a TARGETS variable to control which platforms are built rather than having to manually specify the OS and ARCH multiple times for each command.

This should be less error prone when modifying the target platforms and is certainly easier to add/remove platforms as desired.

### DIFF
--- a/pkg.sh
+++ b/pkg.sh
@@ -6,12 +6,7 @@ set -e
 VER="0.6.0-pre" # pre, beta, rc1, etc.
 META= # "release"
 
-rm -rf bin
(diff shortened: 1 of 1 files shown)


chappjc:
Working.  Thanks!

davecgh:
Moved the `TARGETS` definition up to the top so it's more obvious that it's essentially a config option.

JoeGruffins (approved):

chappjc (approved):

PR #1860, opened 2022-09-22
server: fix purge markets with wrong name

### LINKED ISSUE #1853: server/config: Changing DEXTT lot size causes start up to fail once.
I assume this is also a problem on mainnet but have only tried it on simnet. Probably not anything serious but is weird. To reproduce, start up the harness with eth enabled. Stop it once and edit markets.json and attempt to start up again. It seems only the token will cause an error if lot size is changed:
```
db.Open: unrecognized market dcr_dextt.eth
```

Trying a second time makes the server start.

Also, lot sizes under 1e8 show up as 0e8 and should probably show the amount. USDC will only have 1e6 units per dollar.
```
2022-09-21 15:31:58.806 [DBG] MAIN: -------------------- BEGIN parsed markets.json --------------------
2022-09-21 15:31:58.806 [DBG] MAIN: MARKETS
2022-09-21 15:31:58.806 [DBG] MAIN:                   Base         Quote    LotSize     EpochDur
2022-09-21 15:31:58.806 [DBG] MAIN: Market 0:   DCR_simnet    BTC_simnet       10e8     15000 ms
2022-09-21 15:31:58.806 [DBG] MAIN: Market 1:   DCR_simnet    ETH_simnet       10e8     15000 ms
2022-09-21 15:31:58.806 [DBG] MAIN: Market 2:   DCR_simnet  DEXTT_simnet        0e8     15000 ms
2022-09-21 15:31:58.806 [DBG] MAIN: 
```

vctt94:
fixes #1853

### DIFF
--- a/server/db/driver/pg/tables.go
+++ b/server/db/driver/pg/tables.go
@@ -206,7 +206,9 @@ func prepareMarkets(db *sql.DB, mktConfig []*dex.MarketInfo) ([]string, error) {
 				if err != nil {
 					return nil, fmt.Errorf("unable to update lot size for %s: %w", mkt.Name, err)
 				}
-				purgeMarkets = append(purgeMarkets, mkt.Name)
+				// archiver.markets use market schema name.
+				schema := marketSchema(mkt.Name)
+				purgeMarkets = append(purgeMarkets, schema)
 			}
 		}
 


server/db/driver/pg/tables.go
```diff
@@ -206,7 +206,9 @@ func prepareMarkets(db *sql.DB, mktConfig []*dex.MarketInfo) ([]string, error) {
 				if err != nil {
 					return nil, fmt.Errorf("unable to update lot size for %s: %w", mkt.Name, err)
 				}
-				purgeMarkets = append(purgeMarkets, mkt.Name)
+				// archiver.markets use market schema name.
+				schema := marketSchema(mkt.Name)
+				purgeMarkets = append(purgeMarkets, schema)
```
chappjc, line 211: Wow good catch. Can you do a cursory audit of some of the other pg code to ensure we're not taking the market name directly anywhere else?

vctt94: Sure I can. Should I look all methods which have a call to `db.Exec`?

chappjc: Anywhere really. Query, Exec, helper functions that feed into other query methods, etc.

PR #1864, opened 2022-09-23
app: Dynamically create order page match cards

martonp:
Updates the order page to dynamically create match cards instead of creating them with the template. This allows the user to see new matches without having to reload the page.


client/webserver/site/src/js/order.ts
```diff
@@ -40,6 +40,9 @@ export default class OrderPage extends BasePage {
     // Find the order
     this.orderID = main.dataset.oid || ''
 
+    page.matchCardTmpl.removeAttribute('id')
+    page.matchCardTmpl.remove()
```
buck54321, line 44: Can also `Doc.cleanTemplates`.

client/webserver/site/src/js/order.ts
```diff
@@ -104,7 +107,11 @@ export default class OrderPage extends BasePage {
       match: (note: MatchNote) => { this.handleMatchNote(note) }
     })
 
-    this.start()
+    this.start().then(
+      () => {
+        this.showMatchCards()
+      }
+    )
```
buck54321, line 114: Why not just put `this.showMatchCards()` at the end of `start`?

client/webserver/site/src/js/order.ts
```diff
@@ -133,6 +140,223 @@ export default class OrderPage extends BasePage {
...
+   * changed.
+   */
+  setImmutableMatchCardElements (matchCard: HTMLElement, match: Match) {
+    Doc.tmplElement(matchCard, 'matchID').textContent = match.matchID
+
+    const time = new Date(match.stamp)
+    Doc.tmplElement(matchCard, 'matchTime').textContent = time.toLocaleTimeString('en-GB', {
+      year: 'numeric',
+      month: 'short',
+      day: 'numeric'
```
buck54321, line 154:
```Javascript
const tmpl = Doc.parseTemplate(matchCard)
tmpl.matchTime.textContent = ...
```

buck54321 (commented):
Nice work! A couple of observations from a cursory scan.

JoeGruffins (approved):
Looks good.

client/webserver/site/src/js/order.ts
martonp: True

JoeGruffins:
The accelerate button does not seem to be showing up for the orders page most of the time. Do you also see, or not see, this? I can make an issue or you can if so. Unless its an easy fix.

chappjc:
> I can't remember, but do the accelerations handle accelerations for different matches for the same order?

Acceleration is at the level of the order, not the match, because they are all in a chain, spending change to fund subsequent matches, as you noted.

client/core/helpers.go
```diff
@@ -432,105 +432,6 @@ func (ord *OrderReader) MatchReaders() []*matchReader {
 	return readers
 }
 
-// IsMaker returns if the user is the maker in this match.
```
buck54321, line 435: Looks to me like the rest of `matchReader` can go now. There are no invocations of `(*OrderReader).MatchReaders` remaining, and the remaining methods aren't used (some share names with `OrderReader` methods, but the `matchReader` variants aren't used.)

client/webserver/site/src/js/order.ts
```diff
@@ -133,6 +137,212 @@ export default class OrderPage extends BasePage {
...
+    if (showMakerSwap) Doc.show(tmpl.makerSwap)
+    else Doc.hide(tmpl.makerSwap)
+    if (showTakerSwap) Doc.show(tmpl.takerSwap)
+    else Doc.hide(tmpl.takerSwap)
+    if (showMakerRedeem) Doc.show(tmpl.makerRedeem)
+    else Doc.hide(tmpl.makerRedeem)
+    if (showTakerRedeem) Doc.show(tmpl.takerRedeem)
+    else Doc.hide(tmpl.takerRedeem)
+    if (showRefund) Doc.show(tmpl.refund)
+    else Doc.hide(tmpl.refund)
```
buck54321, line 320: Some of the existing code here is clearly from very early development. There is also `Doc.setVis` now, so you could do 1 or 2 lines per element, e.g.

```Javascript
Doc.setVis(!m.isCancel && (makerSwapCoin(m) || !m.revoked), tmpl.makerSwap)
```

martonp:
I'm still fixing the acceleration button issue. =

buck54321 (approved):

PR #1865, opened 2022-09-24
server/market: send revoke order notes on suspend with purge

chappjc:
The market `'suspension'` route is a somewhat mixed purpose notification in that all connected clients receive a notification of scheduled future suspension, but only `'orderbook'` subscribers get the _sequenced_ suspension notification at the time the market actually clears the book.

As such, if you have orders on the book, but do *not* have an orderbook subscription (e.g. you closed the browser window and it unsubscribed after 30 seconds or so), you will not get the final suspension notification that performs the "auto-revokes" and sends the kinder `TopicOrderAutoRevoked` notes to the frontend.  Instead, your client thinks the orders are still booked **until you reconnect with the server after market resumption** and order status reconciles from booked to revoked.  This is a bug.

To ensure that non-subscribers do not keep their orders in booked status incorrectly, this modifies `(*Market).PurgeBook` to send a `revoke_order` notification to each (user,order) that was purged.  These do not require an `orderbook` subscription.  These revokes are less friendly looking because (a) they don't explain why it was revoked other than by observing the prior suspension scheduled notification, and (b) they are sent to the frontend at error level severity instead of warn as the auto-revokes do.  However, this fixes the bug of clients without an orderbook subscription being unaware that their orders were unbooked.

Finally, this makes the revoke_order notifications come *after* the trade suspension notification so that clients that did have a subscription will get the friendlier message on the frontend, and then simply ignore the explicit redundant revoke_order notes that follow.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -6847,6 +6847,12 @@ func handleRevokeOrderMsg(c *Core, dc *dexConnection, msg *msgjson.Message) erro
 		return fmt.Errorf("no order found with id %s", oid.String())
 	}
 
+	if tracker.status() == order.OrderStatusRevoked {
+		// Already revoked is expected if entire book was purged in a suspend
+		// ntfn, which emits a gentler and more informative notification.
+		// However, we may not be subscribed to orderbook notifications.
+		return nil
+	}
 	tracker.revoke()
 
 	subject, details := c.formatDetails(TopicOrderRevoked, tracker.token(), tracker.mktID, dc.acct.host)
--- a/server/market/market.go
+++ b/server/market/market.go
@@ -1273,17 +1273,17 @@ func (m *Market) Book() (epoch int64, buys, sells []*order.LimitOrder) {
 // to status revoked.
 func (m *Market) PurgeBook() {
 	// Clear booked orders from the DB and the in-memory book.
-	buysRemoved, sellsRemoved := m.purgeBook()
+	removed := m.purgeBook()
 
 	// Send individual revoke order notifications. These are not part of the
 	// orderbook subscription, so the users will receive them whether or not
 	// they are subscribed for book updates.
-	for _, lo := range append(sellsRemoved, buysRemoved...) {
-		m.sendRevokeOrderNote(lo.ID(), lo.AccountID)
+	for oid, aid := range removed {
+		m.sendRevokeOrderNote(oid, aid)
 	}
 }
 
-func (m *Market) purgeBook() (buysRemoved, sellsRemoved []*order.LimitOrder) {
+func (m *Market) purgeBook() (removed map[order.OrderID]account.AccountID) {
 	m.bookMtx.Lock()
 	defer m.bookMtx.Unlock()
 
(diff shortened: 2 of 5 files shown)


JoeGruffins (approved):

buck54321:
Can't the client look at the `Persist` and `FinalEpoch` fields of the `TradeSuspension` to know when to self-revoke their orders?

chappjc:
> Can't the client look at the `Persist` and `FinalEpoch` fields of the `TradeSuspension` to know when to self-revoke their orders?

That's one thing I considered, but rejected, for a few reasons:

Doing that is not simple or accurate.  _If_ they even get that note, they'd have to set some kind of timer to go off when they _think_ that final epoch was reached, but the user's system clock is not exactly the servers (recall that epoch progress/countdown idea). Then scheduled suspensions can change or be lifted, so we'd have to cancel or reschedule the revokes.  If they did not receive the initial suspend message that scheduled the book purge, they'd have to reference the schedule in the config message on reconnect.  And if all that was coded well, it would still be racy and inaccurate.

Those issues aside, I believe it is generally proper for user-directed revoke_order notes to accompany all changes to a user order's status, _at the time the change is made_.

The suspend notifications are really only for orderbook subscribers to keep their book right.  Doing this batch auto-revoke when the final suspend note arrives still makes sense and it gives a better message, but it was also hiding these missing revoke_order notes.  If we supplement the revoke_order payload with a "reason", we could probably sunset this legacy handling via the orderbook.

chappjc:
Related, 5230c19f46702f1253390c597bfb428ccef95ea6, recently merged with https://github.com/decred/dcrdex/pull/1849

client/core/core.go
```diff
@@ -6847,6 +6847,12 @@ func handleRevokeOrderMsg(c *Core, dc *dexConnection, msg *msgjson.Message) erro
 		return fmt.Errorf("no order found with id %s", oid.String())
 	}
 
+	if tracker.status() == order.OrderStatusRevoked {
+		// Already revoked is expected if entire book was purged in a suspend
+		// ntfn, which emits a gentler and more informative notification.
+		// However, we may not be subscribed to orderbook notifications.
+		return nil
```
chappjc, line 6854: I'd also note that that this is a sensible check anyway to silence error messages from `tracker.revoke()` on account of the myriad of other sefl-revoke causes.

server/market/market.go
```diff
@@ -1272,39 +1272,65 @@ func (m *Market) Book() (epoch int64, buys, sells []*order.LimitOrder) {
...
+
+	// Send individual revoke order notifications. These are not part of the
+	// orderbook subscription, so the users will receive them whether or not
+	// they are subscribed for book updates.
+	for _, lo := range append(sellsRemoved, buysRemoved...) {
+		m.sendRevokeOrderNote(lo.ID(), lo.AccountID)
+	}
+}
+
+func (m *Market) purgeBook() (buysRemoved, sellsRemoved []*order.LimitOrder) {
```
buck54321, line 1286: You **could** get away with a simpler return value, 
```Go
func (m *Market) purgeBook() map[order.OrderID]account.AccountID
```
but meh.

buck54321 (approved):

PR #1866, opened 2022-09-24
client/core: swap completion without server coordination

chappjc:
Resolves https://github.com/decred/dcrdex/issues/952 (part 1 anyway, because the wallet locked issue is fairly different and partially resolved by the backup refund paths)

```
client/core: self-governed trades when server is down

When the client has active swaps on a server that is either down or
mysteriously looses the market on which the trades was started, it is
still necessary to complete these orders if a swap has been broadcast.
In particular, the refund and auto-redeem/find-redemption paths need
to work as expected.

This change allows such trades to be loaded on startup even when
the asset configurations are unavailable from the server for any
reason. The lack of a market configuration or a DEX connection are also
conditions for special handling. Trades in this state are flagged as
"self-governed".

In addition to flagging self-governed trades on startup and login
(resumeTrades), the DEX connect and reconnect handlers now toggle the
flag. The reconnect handler only removes the self-governed flag if the
fresh market config includes the trade's market.

To support safely redeeming a counter-party swap and other actions
that depend on the asset configs (e.g. SwapConf, MaxFeeRate, etc).,
this supplements the db.OrderMetaData struct with both to/from SwapConf
values as they were when the order was placed. This also adds EpochDur
to the struct to help with existing logic for handling stale cancels
and other epoch orders, as well as epoch index computations when
considered with the order's ServerTime stamp.

While it is generally ill-advised to perform a client upgrade with
active orders, this also tries during order loading to get these
values if they were not stored (load as zero).
```

### DIFF
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -216,9 +216,10 @@ var (
 
 	// WalletInfo defines some general information about a Decred wallet.
 	WalletInfo = &asset.WalletInfo{
-		Name:     "Decred",
-		Version:  version,
-		UnitInfo: dexdcr.UnitInfo,
+		Name:              "Decred",
+		Version:           version,
+		SupportedVersions: []uint32{version},
+		UnitInfo:          dexdcr.UnitInfo,
 		AvailableWallets: []*asset.WalletDefinition{
 			{
 				Type:        walletTypeSPV,
@@ -1121,14 +1122,14 @@ func (a amount) String() string {
 // calculate lotSize based on a rate conversion from the base asset's lot size.
 // lotSize must not be zero and will cause a panic if so.
 func (dcr *ExchangeWallet) MaxOrder(ord *asset.MaxOrderForm) (*asset.SwapEstimate, error) {
-	_, est, err := dcr.maxOrder(ord.LotSize, ord.FeeSuggestion, ord.AssetConfig)
+	_, est, err := dcr.maxOrder(ord.LotSize, ord.FeeSuggestion, ord.MaxFeeRate)
 	return est, err
 }
 
 // maxOrder gets the estimate for MaxOrder, and also returns the
 // []*compositeUTXO and network fee rate to be used for further order estimation
 // without additional calls to listunspent.
-func (dcr *ExchangeWallet) maxOrder(lotSize, feeSuggestion uint64, nfo *dex.Asset) (utxos []*compositeUTXO, est *asset.SwapEstimate, err error) {
+func (dcr *ExchangeWallet) maxOrder(lotSize, feeSuggestion, maxFeeRate uint64) (utxos []*compositeUTXO, est *asset.SwapEstimate, err error) {
 	if lotSize == 0 {
 		return nil, nil, errors.New("cannot divide by lotSize zero")
 	}
@@ -1143,11 +1144,11 @@ func (dcr *ExchangeWallet) maxOrder(lotSize, feeSuggestion uint64, nfo *dex.Asse
 	}
 
 	// Start by attempting max lots with a basic fee.
-	basicFee := nfo.SwapSize * nfo.MaxFeeRate
+	basicFee := dexdcr.InitTxSize * maxFeeRate
 	lots := avail / (lotSize + basicFee)
 	cfg := dcr.config()
 	for lots > 0 {
-		est, _, _, err := dcr.estimateSwap(lots, lotSize, feeSuggestion, utxos, nfo, cfg.useSplitTx, 1.0)
+		est, _, _, err := dcr.estimateSwap(lots, lotSize, feeSuggestion, maxFeeRate, utxos, cfg.useSplitTx, 1.0)
 		// The only failure mode of estimateSwap -> dcr.fund is when there is
 		// not enough funds, so if an error is encountered, count down the lots
 		// and repeat until we have enough.
@@ -1162,8 +1163,8 @@ func (dcr *ExchangeWallet) maxOrder(lotSize, feeSuggestion uint64, nfo *dex.Asse
 }
 
 // estimateSwap prepares an *asset.SwapEstimate.
-func (dcr *ExchangeWallet) estimateSwap(lots, lotSize, feeSuggestion uint64, utxos []*compositeUTXO,
-	nfo *dex.Asset, trySplit bool, feeBump float64) (*asset.SwapEstimate, bool /*split used*/, uint64 /* locked */, error) {
+func (dcr *ExchangeWallet) estimateSwap(lots, lotSize, feeSuggestion, maxFeeRate uint64, utxos []*compositeUTXO,
+	trySplit bool, feeBump float64) (*asset.SwapEstimate, bool /*split used*/, uint64 /* locked */, error) {
 
 	var avail uint64
 	for _, utxo := range utxos {
@@ -1172,7 +1173,7 @@ func (dcr *ExchangeWallet) estimateSwap(lots, lotSize, feeSuggestion uint64, utx
 
 	// If there is a fee bump, the networkFeeRate can be higher than the
 	// MaxFeeRate
-	bumpedMaxRate := nfo.MaxFeeRate
+	bumpedMaxRate := maxFeeRate
 	bumpedNetRate := feeSuggestion
 	if feeBump > 1 {
 		bumpedMaxRate = uint64(math.Round(float64(bumpedMaxRate) * feeBump))
@@ -1180,18 +1181,21 @@ func (dcr *ExchangeWallet) estimateSwap(lots, lotSize, feeSuggestion uint64, utx
 	}
 
 	val := lots * lotSize
-	sum, inputsSize, _, _, _, err := tryFund(utxos, orderEnough(val, lots, bumpedMaxRate, nfo))
+	sum, inputsSize, _, _, _, err := tryFund(utxos, orderEnough(val, lots, bumpedMaxRate))
 	if err != nil {
 		return nil, false, 0, err
 	}
 
-	reqFunds := calc.RequiredOrderFundsAlt(val, uint64(inputsSize), lots, nfo.SwapSizeBase, nfo.SwapSize, bumpedMaxRate)
+	reqFunds := calc.RequiredOrderFundsAlt(val, uint64(inputsSize), lots,
+		dexdcr.InitTxSizeBase, dexdcr.InitTxSize, bumpedMaxRate)
 	maxFees := reqFunds - val
 
-	estHighFunds := calc.RequiredOrderFundsAlt(val, uint64(inputsSize), lots, nfo.SwapSizeBase, nfo.SwapSize, bumpedNetRate)
+	estHighFunds := calc.RequiredOrderFundsAlt(val, uint64(inputsSize), lots,
+		dexdcr.InitTxSizeBase, dexdcr.InitTxSize, bumpedNetRate)
 	estHighFees := estHighFunds - val
 
-	estLowFunds := calc.RequiredOrderFundsAlt(val, uint64(inputsSize), 1, nfo.SwapSizeBase, nfo.SwapSize, bumpedNetRate)
+	estLowFunds := calc.RequiredOrderFundsAlt(val, uint64(inputsSize), 1,
+		dexdcr.InitTxSizeBase, dexdcr.InitTxSize, bumpedNetRate)
 	estLowFunds += dexdcr.P2SHOutputSize * (lots - 1) * bumpedNetRate
 	estLowFees := estLowFunds - val
 
@@ -1231,7 +1235,7 @@ func (dcr *ExchangeWallet) PreSwap(req *asset.PreSwapForm) (*asset.PreSwap, erro
 	// The utxo set is only used once right now, but when order-time options are
 	// implemented, the utxos will be used to calculate option availability and
 	// fees.
-	utxos, maxEst, err := dcr.maxOrder(req.LotSize, req.FeeSuggestion, req.AssetConfig)
+	utxos, maxEst, err := dcr.maxOrder(req.LotSize, req.FeeSuggestion, req.MaxFeeRate)
 	if err != nil {
 		return nil, err
 	}
@@ -1260,7 +1264,8 @@ func (dcr *ExchangeWallet) PreSwap(req *asset.PreSwapForm) (*asset.PreSwap, erro
 	}
 
 	// Get the estimate for the requested number of lots.
-	est, splitUsed, _, err := dcr.estimateSwap(req.Lots, req.LotSize, req.FeeSuggestion, utxos, req.AssetConfig, split, bump)
+	est, splitUsed, _, err := dcr.estimateSwap(req.Lots, req.LotSize, req.FeeSuggestion,
+		req.MaxFeeRate, utxos, split, bump)
 	if err != nil {
 		return nil, fmt.Errorf("estimation failed: %v", err)
 	}
@@ -1280,7 +1285,8 @@ func (dcr *ExchangeWallet) PreSwap(req *asset.PreSwapForm) (*asset.PreSwap, erro
 	var maxBump float64
 	var maxBumpEst *asset.SwapEstimate
 	for maxBump = 2.0; maxBump > 1.01; maxBump -= 0.1 {
-		tryEst, splitUsed, _, err := dcr.estimateSwap(req.Lots, req.LotSize, req.FeeSuggestion, utxos, req.AssetConfig, split, maxBump)
+		tryEst, splitUsed, _, err := dcr.estimateSwap(req.Lots, req.LotSize,
+			req.FeeSuggestion, req.MaxFeeRate, utxos, split, maxBump)
 		// If the split used wasn't the configured value, this option is not
 		// available.
 		if err == nil && split == splitUsed {
@@ -1290,7 +1296,8 @@ func (dcr *ExchangeWallet) PreSwap(req *asset.PreSwapForm) (*asset.PreSwap, erro
 	}
 
 	if maxBumpEst != nil {
-		noBumpEst, _, _, err := dcr.estimateSwap(req.Lots, req.LotSize, req.FeeSuggestion, utxos, req.AssetConfig, split, 1.0)
+		noBumpEst, _, _, err := dcr.estimateSwap(req.Lots, req.LotSize, req.FeeSuggestion,
+			req.MaxFeeRate, utxos, split, 1.0)
 		if err != nil {
 			// shouldn't be possible, since we already succeeded with a higher bump.
 			return nil, fmt.Errorf("error getting no-bump estimate: %w", err)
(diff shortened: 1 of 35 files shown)


client/db/types.go
```diff
@@ -257,6 +257,21 @@ type OrderMetaData struct {
...
+	// swap confirmations could change too:
+	//
+	// FromSwapConf uint64
+	// ToSwapConf   uint64
+	//
+	// Without EpochIdx stored explicitly, we currently infer based on the epoch
+	// duration in the *current* market config (actually variable) and
+	// the order's ServerTime stamp. It might be wise to record it too:
+	//
+	// EpochIdx     uint64
```
chappjc, line 273: Before bringing out of draft, I'm going to go forward with storing these data in the client DB, except I'm going to store epoch **duration** instead of index.

JoeGruffins (commented):
It seems if I am taker and an dex goes offline and comes back on later and we revoke, the order stays in settling until I restart the app, then is shows as executed.

image

It also seems that if server goes offline before maker redeem, we always wait til taker locktime is up to check for maker's redeem? This completes exactly at the three minute mark/ taker's locktime.

image


I realized while looking at this pr that swap confs can be zero and the server and client are fine with that. I think this is a problem for eth and hope client can enforce a mimum swap conf?

client/core/core.go
```diff
@@ -6230,10 +6293,55 @@ func (c *Core) resumeTrades(dc *dexConnection, trackers []*trackedTrade) assetMa
...
+		mktConf := dc.marketConfig(tracker.mktID)
+
+		if tracker.metaData.EpochDur == 0 { // upgraded with live orders... smart :/
+			if mktConf != nil { // may remain zero if market also vanished
+				tracker.metaData.EpochDur = mktConf.EpochLen
+			}
+		}
+		if tracker.metaData.FromSwapConf == 0 {
+			tracker.metaData.FromSwapConf = walletSet.fromAsset.SwapConf // possibly still zero if !haveAssetConfigs
+		}
```
JoeGruffins, line 6328: Is seems can just be zero naturally if server sets to zero.

client/core/core.go
```diff
@@ -5274,35 +5311,46 @@ func (w *walletSet) trimmedConventionalRateString(r uint64) string {
...
+			ID:       baseWallet.AssetID,
+			Symbol:   unbip(baseWallet.AssetID),
+			Version:  baseWallet.Info().Version,
+			UnitInfo: baseWallet.Info().UnitInfo,
+		}
+		c.log.Warnf("Base asset server configuration not available for %s (asset %s).",
+			dc.acct.host, unbip(baseID))
+	} else if ver := baseWallet.Info().Version; baseAsset.Version != ver {
+		return nil, false, newError(walletErr, "wallet asset %d version %d does not match server asset version %d",
 			baseID, ver, baseAsset.Version)
```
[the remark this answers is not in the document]
chappjc: You're right.  This should not get in the way.  I don't think this method is even the right place to check version mismatch anyway.

client/core/trade.go
```diff
@@ -1780,6 +1824,18 @@ func (c *Core) tick(t *trackedTrade) (assetMap, error) {
...
+		// For certain "self-governed" trades where the market or server has
+		// vanished, we should revoke the match to allow it to retire without
+		// having sent any pending redeem requests. Note that self-governed is
+		// not necessarily a permanent state, so we delay this action.
+		if !revoked && t.isSelfGoverned() && time.Since(match.matchTime()) > t.lockTimeTaker {
+			c.log.Warnf("Revoking old self-governed match %v for market %v, host %v.",
+				match, t.mktID, t.dc.acct.host)
+			revokes = append(revokes, match)
+			// NOTE: If the trade is in booked status, the order still won't
+			// retire. We need a way to force-cancel such orders.
```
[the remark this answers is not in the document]
chappjc: Yep.  Even after https://github.com/decred/dcrdex/pull/1889, where I note the need for this at the end of the second commit message.

chappjc:
> It also seems that if server goes offline before maker redeem, we always wait til taker locktime is up to check for maker's redeem? This completes exactly at the three minute mark/ taker's locktime.

That's correct, but this deserves more discussion, because it's not clear when it should take the action.  The reason it's not "right away" when it sees their swap is spent, is because doing auto-redeem implies self-revoking the order.

However, the self-governed/down-dex state is transient unlike revocation.  What if the server connection is restored and we've already revoked, completed, and retired the order?  We get penalized for violating the settlement sequence unless we can respond with our redeem ack.  How can we work around that?  Can we (a) keep these orders from retiring so we can handle any future redeem request that might come? (b) load them as needed when we get a redeem request rather than simply erroring because the match isn't active?

Note that its common for a tick to see one's own swap as spent before receiving the `'redeem'` request from the server that formally notifies us of the action.  But we don't immediately jump to the auto-redeem path because we don't want to get ahead.

Probably we need a way for **taker** to freely redeem at any time regardless of if server relayed the details to us.  The trick there is probably just being able to respond to a future redeem request even if we've completed the swap.

chappjc:
It has become clear to me that there are some considerable flaws around `dex.Asset` and how it is used and provided by server.

The only `"configurable asset variables"` we should get from server are: (1) asset version, (2) max fee rate, and (3) swap conf.
The other fields are either unused or redundant with _internal_ wallet parameters.

I've put this in draft because in in the middle of some more significant refactoring to resolve these issues.

client/core/core.go
```diff
@@ -4899,10 +4939,16 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
 		return nil, nil, newError(orderParamsErr, "zero-rate order not allowed")
 	}
 
-	wallets, err := c.walletSet(dc, form.Base, form.Quote, form.Sell)
+	wallets, assetConfigs, versionsMatch, err := c.walletSet(dc, form.Base, form.Quote, form.Sell)
 	if err != nil {
 		return nil, nil, err
 	}
+	// maybe we should just let the wallets decide in FundOrder if the server's
+	// version works? We'd have to confirm with the redeeming wallet too...
```
buck54321, line 4947: Maybe change to a `SupportedVersions []uint32` field in `asset.WalletInfo`?

buck54321 (commented):
Approach ACK. It's beautiful! 😍

chappjc:
Growing in scope, but there's a lot of cool stuff we can do now with these fundamental changes to the order tracking and interfaces.

client/core/core.go
chappjc: I had started looking asset-specific settings to be used by Core but not the wallets, such as a min swap conf setting, but didn't find an nice solution.  We should do this soon though.
Related to a min confs setting per asset, I was looking into a swap conf override (i.e. used to swap or redeem *earlier* than server requires, at your own discretion).  I think we should do that too.  The user should be able to decide to e.g. redeem taker's BTC swap tx after 1 conf instead of maybe a higher server setting of 3, if they want to.

chappjc:
This darn server/swap commit is making the tests fail sporadically, and I can't understand why.  I might drop that commit since it's not required for these client changes to work properly; it's just a bonus.

client/core/core.go
chappjc: Done.  Not entirely show how that's gonna jive with your contract version PR.

JoeGruffins (approved):
Working well. I notice if settling happens offline, the ui will tell me refunding when it is redeeming:
image

Not sure if related to these changes, but a token seems to have a coin id display problem here:
image

chappjc:
That "Refund PENDING" thing has been an issue for a while, but it's potentially less transient now.  Will look into that after this.

I can't seem to identify the source of the token `<invalid coin>` for the maker redeem.  The length of that hex string doesn't even look like a hash.  In fact my count is 39 hexadecimal characters, so I'm guessing it's a 20 byte address (?) with a zero padding stripped???

chappjc:
Just rebased, not re-tested.
Also, https://github.com/decred/dcrdex/pull/1909 should get in first

chappjc:
Just rebased with the partially fixed simnet trade tests from https://github.com/decred/dcrdex/pull/1909
Needs re-testing and I'm sure there will be some changes required to the simnet tests now.

JoeGruffins:
> I can't seem to identify the source of the token <invalid coin> for the maker redeem on the UI. The length of that hex string doesn't even look like a hash. In fact my count is 39 hexadecimal characters, so I'm guessing it's a 20 byte address (?) with a zero padding stripped???

The client should not have the txid, so we need to put some other info in there I guess. Will make an issue about it.

client/asset/eth/eth.go
```diff
@@ -1248,13 +1253,12 @@ func (w *assetWallet) PreRedeem(req *asset.PreRedeemForm) (*asset.PreRedeem, err
 // SingleLotRedeemFees is a fallback for PreRedeem that uses estimation when
 // funds aren't available. The returned fees are the RealisticWorstCase.  The
 // Lots field of the PreSwapForm is ignored and assumed to be a single lot.
-func (w *assetWallet) SingleLotRedeemFees(form *asset.PreSwapForm) (fees uint64, err error) {
-	redeemSize := form.AssetConfig.RedeemSize
-	if redeemSize == 0 {
-		g := w.gases(form.AssetConfig.Version)
-		redeemSize = g.Redeem
+func (w *assetWallet) SingleLotRedeemFees(form *asset.PreRedeemForm) (fees uint64, err error) {
```
chappjc, line 1256: Seems that eth wasn't a BotWallet because of the input arg type.  Added an assertion and fixed the input type.

client/core/bot.go
```diff
@@ -1246,27 +1247,29 @@ func (c *Core) feeEstimates(form *TradeForm) (swapFees, redeemFees uint64, err e
...
 		FeeSuggestion:   redeemFeeSuggestion,
 		SelectedOptions: form.Options,
-		AssetConfig:     toAsset,
 	}
 	redeemEstimate, err := toWallet.PreRedeem(preRedeemForm)
 	if err == nil {
 		redeemFees = redeemEstimate.Estimate.RealisticWorstCase
 	} else {
-		if bw, is := fromWallet.Wallet.(asset.BotWallet); is {
+		if bw, is := toWallet.Wallet.(asset.BotWallet); is {
```
chappjc, line 1266: Pretty sure the `fromWallet` part was copy pasta from the `PreSwap`->`SingleLotSwapFees` fallback pattern above.

chappjc:
What a rabbit hole.  Sorry for the additional commits.

client/asset/dcr/dcr.go
```diff
@@ -2530,69 +2529,72 @@ func determineTxTree(msgTx *wire.MsgTx) int8 {
...
-	// but still not found by wallet.UnspentOutput.
-	var outputPaysToWallet bool
-	for _, details := range tx.Details {
-		if details.Vout == vout {
-			outputPaysToWallet = details.Category == wallet.CreditReceive.String()
-			break
-		}
-	}
-	if outputPaysToWallet {
-		dcr.log.Debugf("Output %s:%d was not reported as unspent, pays to the wallet and is considered SPENT.",
```
chappjc, line 2577: As the comments in the revisions say, this was never happening in the dex use case.  Verified searching lots of logs for this log statement.

client/asset/dcr/dcr.go
```diff
@@ -3506,9 +3508,12 @@ func (dcr *ExchangeWallet) SwapConfirmations(ctx context.Context, coinID, contra
...
 
 	// Check if we can find the contract onchain without using cfilters.
-	_, confs, spent, err = dcr.lookupTxOutput(ctx, txHash, vout)
+	var spendFlag int8
+	_, confs, spendFlag, err = dcr.lookupTxOutput(ctx, txHash, vout)
 	if err == nil {
-		return confs, spent, nil
+		if spendFlag != -1 {
+			return confs, spendFlag > 0, nil
+		} // else go on to block filters scan
```
chappjc, line 3516: Without this distinction, all of the below code to go to `lookupTxOutWithBlockFilters` NEVER happened.  Also checked lots of logs for the "SwapConfirmations - cfilters scan did not find" string... nothing.

client/core/trade.go
```diff
@@ -1359,10 +1359,8 @@ func (t *trackedTrade) isSwappable(ctx context.Context, match *matchTracker) (re
 			// is expected for newly made swaps involving contracts.
 			t.dc.log.Errorf("isSwappable: error getting confirmation for our own swap transaction: %v", err)
 		}
-		if spent {
-			t.dc.log.Debugf("Our (maker) swap for match %s is being reported as spent, "+
-				"but we have not seen the counter-party's redemption yet. This could just"+
-				" be network latency.", match)
+		if spent { // This should NEVER happen for maker in MakerSwapCast unless revoked and refunded!
+			t.dc.log.Errorf("Our (maker) swap for match %s is being reported as spent before taker's swap was broadcast!", match)
```
chappjc, line 1363: Also verified that this (incorrect log) never happened.  If it had, it would have meant a major problem.

client/asset/dcr/dcr.go
buck54321: Wouldn't that mean that `lookupTxOutput -> GetTransaction` never returned `asset.CoinNotFoundError`? There does seem to be a clear path to the code below.

client/core/core.go
```diff
@@ -4396,17 +4413,19 @@ func (c *Core) MaxBuy(host string, base, quote uint32, rate uint64) (*MaxOrderEs
 	maxBuy, err := quoteWallet.MaxOrder(&asset.MaxOrderForm{
 		LotSize:       quoteLotEst,
 		FeeSuggestion: swapFeeSuggestion,
-		AssetConfig:   quoteAsset,
-		RedeemConfig:  baseAsset,
+		AssetVersion:  quoteAsset.Version, // using the server's asset version, when our wallets support multiple vers
```
buck54321, line 4416: Eventually, we'll get the UI figured out where we hide order forms for unsupported versions, and will also check version compatibility before we get this far. This note is good enough for now.

client/core/core.go
```diff
@@ -4547,6 +4568,16 @@ func (c *Core) initializeDEXConnections(crypter encrypt.Crypter) []*DEXBrief {
 			continue
 		}
 
+		// If the connection is down, authDEX will fail on Send.
```
buck54321, line 4571: We were already sending a notification when authDEX failed. Is this about setting the connection icon on the front end?

client/core/core.go
```diff
@@ -5262,21 +5310,28 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
 		return nil, nil, fmt.Errorf("validateOrderResponse error: %w", err)
 	}
 
+	// TODO: Need xcWallet fields for acceptable SwapConf values: a min
+	// acceptable for security, and even a max confs override to act sooner.
```
buck54321, line 5314: via `WalletInfo`, I guess?

client/core/core.go
```diff
@@ -6750,6 +6882,47 @@ func (c *Core) handleReconnect(host string) {
...
+	dc.cfgMtx.RUnlock()
+
+	// Update the orders' selfGoverned flag according to the configured markets.
+	for _, trade := range dc.trackedTrades() {
+		// If the server's market is gone, we're on our own, otherwise we are
+		// now free to swap for this order.
+		auto := mkts[trade.mktID] == nil
+		if !auto { // market exists, now check asset config and version
+			baseCfg := dc.assetConfig(trade.Base())
+			auto = baseCfg == nil || baseCfg.Version != trade.wallets.baseWallet.version // TODO: consider allowing, handling version check in wallets
```
buck54321, line 6908: I guess `xcWallet` will need a `supportedVersions []uint32`  field.

client/core/core.go
```diff
@@ -7572,16 +7735,138 @@ func handleRedemptionRoute(c *Core, dc *dexConnection, msg *msgjson.Message) err
...
+		c.schedTradeTick(tracker)
+		return nil
 	}
-	err = tracker.processRedemption(msg.ID, redemption)
+
+	sigMsg := redemption.Serialize()
+	err = dc.acct.checkSig(sigMsg, redemption.Sig)
+	if err != nil {
+		c.log.Warnf("Server redemption signature error: %v", err) // just warn
+	}
```
buck54321, line 7756: Does this need to be moved up now?

client/core/trade.go
```diff
@@ -128,6 +132,26 @@ func (m *matchTracker) matchTime() time.Time {
 	return time.UnixMilli(int64(m.MetaData.Proof.Auth.MatchStamp)).UTC()
 }
 
+func (m *matchTracker) swapSpentAgo() time.Duration {
+	now := time.Now() // avoid system clock with mutex locked
```
buck54321, line 136: > `// avoid system clock with mutex locked`

Why?

buck54321 (commented):
Partial review. Mostly done, but I gotta run for the morning. Here are just a few notes.

client/asset/dcr/dcr.go
chappjc: I mean it never happened when the swap was actually confirmed (and found by GetTransaction).  No spending transaction was ever found by `SwapConfirmations` as a result.  Also note that for you own swap, `GetTransaction` always returns a result even when unconfirmed because it's a wallet txn, hence we never determined spend status of our own contract, always relying on revoke to allow find redemption.

client/core/core.go
chappjc: Yeah it's not yet ideal to just get an error from the API, but at least it's caught here and in `PreOrder` when it calls `walletSet` and returns for `!versCompat`.

client/core/core.go
chappjc: This is mainly about avoiding a confusing error from authDEX about a broken connection or something when we know it won't be possible.  Nothing goes terribly wrong without these lines, but it's a smoother UX. (`TopicDEXDisconnected` with host name vs. `TopicDexAuthError` with an error message from the comms methods)

client/core/core.go
chappjc: It has `baseWallet.Info().SupportedVersions`. I need to do here what I did in `(*Core).walletSet` to check `versCompat`.

client/core/core.go
chappjc: These are the settings that would be wallet/asset-specific, but not internal to to the wallet.  They are used at the level of `Core` (and even higher like the UI).  https://github.com/decred/dcrdex/issues/1900

chappjc: Oh but the default *min* confs would be appropriate to get from the asset code via `WalletInfo` or whatever, yeah.

client/core/core.go
chappjc: Before the call to `processRedemption`? Yah, I suppose could do that, and remove the redundant `checkSig` from that method?

client/core/trade.go
chappjc: It's a relatively expensive syscall, and there have been Go bugs that have made it worse.  This isn't a hot path, so perhaps an unnecessary optimization.

https://github.com/golang/go/issues/22601
https://tpaschalis.github.io/golang-time-now/

client/asset/estimation.go
```diff
@@ -61,6 +59,17 @@ type PreSwapForm struct {
...
+	// The following fields are only used for some assets where the redeemed/to
+	// asset may require funds in this "from" asset. For example, buying ERC20
+	// tokens with ETH. The funds may be used for withdraw approval for the
+	// redeemed token contract, or just the redeem tx gas.
+
+	// RedeemVersion is the asset version of the "to" asset with the redeem
+	// transaction. Most backends only support one version.
+	RedeemVersion uint32
+	// RedeemVersionAssetID is the asset
+	RedeemVersionAssetID uint32
```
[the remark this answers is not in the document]
chappjc: haha, I'm racking my brain to figure out what I might have been thing with the name `RedeemVersionAssetID`, but I guess this is just strange copy pasta.

client/core/core.go
buck54321: Ah. I didn't notice the check in `processRedemption`. But yeah, better the way you have it in your latest commit.

client/webserver/site/src/js/app.ts
```diff
@@ -802,6 +802,7 @@ export default class Application {
   walletIsActive (assetID: number): boolean {
     if (this.haveAssetOrders(assetID)) return true
     for (const xc of Object.values(this.user.exchanges)) {
+      if (!xc) continue
       if (xc.pendingFee && xc.pendingFee.assetID === assetID) {
```
buck54321, line 806: So there can be nil entries in the map now?

buck54321 (approved):

PR #1868, opened 2022-09-25
server: reject empty password from the terminal

ukane-philemon:
This fixes an issue where the `PasswordPrompt` method accepts an empty string as a valid password.

### DIFF
--- a/server/admin/prompt.go
+++ b/server/admin/prompt.go
@@ -21,6 +21,19 @@ type passwordReadResponse struct {
 // PasswordPrompt prompts the user to enter a password. Password must not be an
 // empty string.
 func PasswordPrompt(ctx context.Context, prompt string) ([]byte, error) {
+	var password []byte
+	var err error
+	for len(password) == 0 {
+		password, err = passwordPrompt(ctx, prompt)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return password, nil
+}
+
+// passwordPrompt prompts the user to enter a password.
+func passwordPrompt(ctx context.Context, prompt string) ([]byte, error) {
 	// Get the initial state of the terminal.
 	initialTermState, err := term.GetState(int(os.Stdin.Fd()))
 	if err != nil {


JoeGruffins (approved):

server/admin/prompt.go
```diff
@@ -21,6 +21,19 @@ type passwordReadResponse struct {
...
 // empty string.
 func PasswordPrompt(ctx context.Context, prompt string) ([]byte, error) {
+	var password []byte
+	var err error
+	for len(password) == 0 {
+		password, err = passwordPrompt(ctx, prompt)
+		if err != nil {
+			return nil, err
+		}
+	}
```
chappjc, line 31: Can you comment on the desirability of a retry loop vs returning an error?

chappjc (commented):
Is it incorrect to have a zero-length password, not considering security implications?

server/admin/prompt.go
ukane-philemon: > Can you comment on the desirability of a retry loop vs returning an error?
if users don't enter a password, we use the retry loop to give them another chance.

buck54321:
> This fixes an issue where the `PasswordPrompt` method accepts an empty password as a valid password.
> 
> Note: this fix does not check for (`""`). However, (`""`) will eventually fail when trying to convert password bytes to type `encode.PassBytes`.

Can you describe the problem a little more. What do you mean by "accepts an empty password"? Surely you don't mean that we were allowing authed actions to be performed with an empty password. Wouldn't an empty password result an an error before?

ukane-philemon:
> Can you describe the problem a little more.

If on the terminal and you're required to provide a password to a password prompt. The current behavior is that, if you click `Enter` key without providing a password, the prompt will accept it and passed it on as a valid password. This fixes the issue.

chappjc:
That doesn't sound like a problem to me though.  It gets passed on and then whatever gets that empty string decides if it is correct or not.  Doesn't clearly seem like a job for the password prompt.

ukane-philemon:
Hmm, the password prompt is used in places where a non-zero-length password is a MUST, I think. We could let callers do the checking as they see fit but I was looking at a simpler way to ensure the users provide a non-zero-length password and do it efficiently.

JoeGruffins:
> an empty string is a valid password for Electrum

That's not something I considered and a good point where you might use the password prompt for dexcctl.

ukane-philemon:
> Motion to close on account of an empty password being valid

It turns out that whilst the password from a password Prompt shouldn't be empty like in the case of:
 - creating and supplying dexcctl app password 
 - creating server signing key password
 It is valid for cases where the password `can` be empty(i.e for wallet password).
 
 This can't be classified as a `real` issue, just a minor improvement. With a few more lines of code, we could ensure empty passwords are not supplied where they shouldn't and handle exceptional cases(e.g walletpass). 
 
 Users providing an empty password when they shouldn't, fail & cause unexpected issues eventually but we could prevent that. 
 
If it's worth it, I could make the improvement.

Issue #1867, opened 2022-09-24
litecoin electrum rpc issue with walletpath when using multiple wallets


karamble:
i have created a 2nd litecoin wallet in my electrum wallet gui.
from that moment on i am not able to connect my wallet trough the dexc gui, always giving me:
'wallet not loaded' errors.

Research shows, that the rpc query needs to specify the walletpath, see:
https://pythontechworld.com/issue/spesmilo/electrum/7391

I am not able to enter a wallet name in the 
Electrum-Ltc (external) - tab when adding a new litecoin wallet, therefore i think we need to add the form field for wallet name and add it to the rpc query. see:
https://github.com/decred/dcrdex/search?q=default_wallet

chappjc:
It's documented that it must be named default_wallet.
https://github.com/decred/dcrdex/wiki/Client-Installation-and-Configuration#sync-blockchains
We will eventually support multiple like you indicated, but not at present.

chappjc:
It's a legit issue, to be clear.

It's just an irritating rpc design because it requires the full path to the wallet, not just it's name.  Plus it's in every request payload rather than in the endpoint.

chappjc:
Solution brewing.  Not too bad, just annoying API.

PR #1870, opened 2022-09-26
client/asset/btc: provide Electrum wallet path in RPCs if specified

chappjc:
Resolves https://github.com/decred/dcrdex/issues/1867

To use a non-default wallet, it is necessary to:
- Include the full path to the wallet file loaded in Electrum in all RPCs that pertain to a specific wallet. This string is passed in the RPCs without interpreting the path.  See https://github.com/spesmilo/electrum/issues/7391
- Use named/dict `params` because positional arguments do not work for the wallet argument.  See https://github.com/spesmilo/electrum/issues/6221

Tested with: (1) the `electrumlive` BTC wallet "test", (2) the example/wallet/main.go app for LTC, and (3) interactive BTC testing with dexc and the UI with both an empty wallet arg for the default and a path to the wallet file `~/.electrum/testnet/wallets/wallet_1`. Tested on [a 0.5 branch](https://github.com/chappjc/dcrdex/tree/electrum-wallet-args-0.5), but almost identical.

image

### DIFF
--- a/client/asset/btc/electrum_client.go
+++ b/client/asset/btc/electrum_client.go
@@ -372,6 +372,7 @@ func (ew *electrumWallet) reconfigure(cfg *asset.WalletConfig, currentAddress st
 	if err != nil {
 		return false, fmt.Errorf("error parsing rpc wallet config: %w", err)
 	}
+	dexbtc.StandardizeRPCConf(&parsedCfg.RPCConfig.RPCConfig, "")
 
 	// Changing RPC settings is not supported without restart.
 	return parsedCfg.RPCConfig != *ew.rpcCfg, nil
--- a/client/asset/btc/electrum_client.go
+++ b/client/asset/btc/electrum_client.go
@@ -274,6 +274,10 @@ func (ew *electrumWallet) connect(ctx context.Context, wg *sync.WaitGroup) error
 	ew.resetChain(chain)
 	ew.ctx = ctx // for requests via methods that lack a context arg
 
+	// This wallet may not be "protected", in which case we omit the password
+	// from the requests. Detect this now and flag the wallet as unlocked.
+	_ = ew.walletUnlock([]byte{})
+
 	// Start a goroutine to keep the chain client alive and on the same
 	// ElectrumX server as the external Electrum wallet if possible.
 	wg.Add(1)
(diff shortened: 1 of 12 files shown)


JoeGruffins (approved):
It seems I cannot send funds from the wallet with the error `wallet does not support fee estimation`. I am running this rebased on master if that matters.

Also, you can switch the wallet to electrum with no password, and the symbol says unlocked, but when you try to trade you get the error `wallet reporting as locked, but no password has been set`. You can then go to settings and set the password but maybe better to check if it needs a password when re-configuring the wallet?

Otherwise testing great.

chappjc:
> It seems I cannot send funds from the wallet with the error `wallet does not support fee estimation`. I am running this rebased on master if that matters.

You should always be able to send without fee estimation. @ukane-philemon We cannot block sending because /txfee doesn't work.

chappjc:
> Also, you can switch the wallet to electrum with no password, and the symbol says unlocked, but when you try to trade you get the error `wallet reporting as locked, but no password has been set`. You can then go to settings and set the password but maybe better to check if it needs a password when re-configuring the wallet?

Hmm, very long standing TODO: https://github.com/decred/dcrdex/blob/94a9aef6442c77cd7a55d5b136a1a56f125f09b8/client/core/core.go#L3097-L3102

Seems we make an assumption about the meaning of an empty password at the level of `Core` that is not true for all wallets.  Not sure of the best solution right now.  Will consider more.

chappjc:
I believe I've resolved both the empty password and fee estimate issues.

Also fixed an issue with live reconfigure thinking the RPC config changed when it didn't.

chappjc:
@ukane-philemon @buck54321 I can pop off this last commit dealing with using `traitTxFeeEstimator` on the frontend, and you can build on that in https://github.com/decred/dcrdex/pull/1840.  I only tagged it on this PR because it was relevant to Electrum.
Probably best not to churn on the resolution in two separate PRs.  Will pop it off now.  For ref it is https://github.com/decred/dcrdex/pull/1870/commits/fe20a0cf609870a6d6fc5a49da481110dfb64bd2

client/core/core.go
```diff
@@ -3112,6 +3112,21 @@ func (c *Core) setWalletPassword(wallet *xcWallet, newPW []byte, crypter encrypt
...
 	} else {
+		// Test that the wallet is actually good with no password. At present,
+		// this means the backend either cannot be locked or unlocks with an
+		// empty password. This might change. We will ignore several errors in
+		// this process since if the external wallet's password was removed
+		// without a reconnect, the backend state may be inconsistent without a
+		// full reconnect with a fresh wallet.
+		backend := wallet.Wallet
+		_ = backend.Lock()           // may error, but will not actually lock for unprotected wallet
+		_ = backend.Unlock([]byte{}) // some backends may error, others require to avoid a reconnect
```
buck54321, line 3123: > others require to avoid a reconnect

Is this referring to Electrum?

client/core/core.go
```diff
@@ -3112,6 +3112,21 @@ func (c *Core) setWalletPassword(wallet *xcWallet, newPW []byte, crypter encrypt
 		}
 		wallet.setEncPW(encNewPW)
 	} else {
+		// Test that the wallet is actually good with no password. At present,
```
buck54321, line 3115: TODO above can be deleted?

client/core/core.go
chappjc: I suppose it would be any wallet that lets you go from protected->unprotected (or the reverse) while the wallet is running _and DEX is already connected to it_.  Off hand, I'm not sure if Bitcoin Core lets you remove the passphrase or not.  But yes, this currently addresses the possibility with Electrum wallets how I can remove password protection and our code won't find out until reconnect unless we do some prodding here at time of reconfigure.

image

For reference, we discover this on wallet connect for Bitcoin Core by inspecting a `getwalletinfo` response field, while for Electrum we test with a getprivatekey RPC to see if it works with no pass.  AFAICT we had a bit of a hole in the reconfigure logic that would always let you remove the password even if that was not correct for the wallet.

client/asset/btc/electrum_client.go
```diff
@@ -895,10 +900,35 @@ func (ew *electrumWallet) pass() (pw string, unlocked bool) {
...
+func (ew *electrumWallet) testPass(pw []byte) error {
+	addr, err := ew.wallet.GetUnusedAddress(ew.ctx)
+	if err != nil {
+		return err
+	}
+	wifStr, err := ew.wallet.GetPrivateKeys(ew.ctx, string(pw), addr)
+	if err != nil {
+		if strings.Contains(strings.ToLower(err.Error()), "incorrect padding") {
+			return errors.New("incorrect password (no password required?)")
+		}
```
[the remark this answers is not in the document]
chappjc: Ah, yeah that's not at all clear.  I discovered that if the wallet was not password protected, but you provided a password, it would return the cryptic "incorrect padding" response.  I'm pretty sure I got "incorrect padding" in another case as well, so I didn't want to state definitively that it was "no password required", rather that's a possibility and it's definitely not the right password.

JoeGruffins:
Password issue looks resolved. Unrelated to these changes, but I noticed you can't disable the rpc wallet if it's not connected which makes disable for those types not as useful.

chappjc:
> Unrelated to these changes, but I noticed you can't disable the rpc wallet if it's not connected which makes disable for those types not as useful.

Aw geeze, yeah that's the primary use case.  I have a dumb dogecoind testnet wallet that I rarely feel like starting up, and if I cannot disable without connecting it first...

chappjc:
Cleaned up around number of comments mentioned in reviews.  Also applied and tested a trivial fix to allow disabling the disconnected wallet.

PR #1863, opened 2022-09-23
zh translation update

DominicTing:

PR #1871, opened 2022-09-27
client: updates for zh-CN

chappjc:
Updates for zh-CN translations.

This PR is a massaged version of https://github.com/decred/dcrdex/pull/1863


client/webserver/locales/zh-cn.go
```diff
@@ -12,79 +12,81 @@ var ZhCN = map[string]string{
...
-	"cancel_pw":                      "输入密码以取消剩余的请求", // alt. 输入密码以取消订单未成交的
-	"cancel_no_pw":                   "取消剩余未成交订单",
-	"cancel_remain":                  "剩余数量可在取消请求一致之前更改。",
+	"Cancel Order":                   "取消订单",
+	"cancel_pw":                      "输入您的密码以取消订单",
+	"cancel_no_pw":                   "取消剩余订单",
+	"cancel_remain":                  "在取消订单之前，剩余金额可能会发生变化。",
 	"Log In":                         "登录",
-	"epoch":                          "时间",
+	"epoch":                          "epoch",
```
chappjc, line 57: These don't translate properly?

chappjc, line 57: These don't translate properly?

chappjc:
Slightly behind because the frontend is such a moving target, but I don't think we need to hold up this update as long as it doesn't break anything.

PR #1872, opened 2022-09-27
[v0.5] dex/candles: delta overflow bug

buck54321:
whoops

### DIFF
--- a/dex/candles/candles.go
+++ b/dex/candles/candles.go
@@ -130,8 +130,8 @@ func (c *Cache) Delta(since time.Time) (changePct float64, vol uint64) {
 		if candle.StartStamp <= cutoff && candle.StartRate != 0 {
 			// Interpret the point linearly between the start and end stamps
 			cut := float64(cutoff-candle.StartStamp) / float64(candle.EndStamp-candle.StartStamp)
-			rateDelta := candle.EndRate - candle.StartRate
-			r := candle.StartRate + uint64(cut*float64(rateDelta))
+			rateDelta := float64(candle.EndRate) - float64(candle.StartRate)
+			r := candle.StartRate + uint64(cut*rateDelta)
 			if r > 0 {
 				startRate = r
 			}


JoeGruffins (approved):

dex/candles/candles.go
```diff
@@ -130,8 +130,8 @@ func (c *Cache) Delta(since time.Time) (changePct float64, vol uint64) {
 		if candle.StartStamp <= cutoff && candle.StartRate != 0 {
 			// Interpret the point linearly between the start and end stamps
 			cut := float64(cutoff-candle.StartStamp) / float64(candle.EndStamp-candle.StartStamp)
-			rateDelta := candle.EndRate - candle.StartRate
-			r := candle.StartRate + uint64(cut*float64(rateDelta))
+			rateDelta := float64(candle.EndRate) - float64(candle.StartRate)
```
chappjc, line 133: Oh signed integers.

Trying to gauge impact.  There going to be candles in the epoch_reports with crazy spot rate changes, and there are going to be crazy candles from the `/candles` endpoint, but on restart will all that correct itself when `AddMarketSource` >  `LoadEpochStats` runs?

chappjc: Can you also update `TestDelta` with a `Candle` where `EndRate` is less than `StartRate`?

buck54321: > Oh unsigned integers.
> 
> Trying to gauge impact. There going to be candles in the epoch_reports with crazy spot rate changes, and there are going to be crazy candles from the `/candles` endpoint, but on restart will all that correct itself when `AddMarketSource` > `LoadEpochStats` runs?

That's right. Nothing is wrong in the DB, so everything should correct itself on restart.

chappjc (approved):

chappjc:
OK, merged to `release-v0.5` and operators can simply deploy from that branch.  No client updates require AFAICT.

We'll have to put this on master too.

PR #1875, opened 2022-09-28
dex/candles: delta overflow bug

buck54321:
#1872, but for master

### DIFF
--- a/dex/candles/candles_test.go
+++ b/dex/candles/candles_test.go
@@ -154,10 +154,10 @@ func TestDelta(t *testing.T) {
 	})
 	c.Add(&Candle{
 		MatchVolume: 150,
-		StartStamp:  now - 2*fiveMins,
-		EndStamp:    now - fiveMins,
+		StartStamp:  aDayAgo,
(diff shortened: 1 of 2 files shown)


dex/candles/candles_test.go
```diff
@@ -167,41 +167,63 @@ func TestDelta(t *testing.T) {
...
 	// Basic function test.
 	check24(0.75, 200)
 
+	// Test halfway through the candle time. 125 -> 175 = +0.4, vol = 75 + 50
+	testTime = tNow.Add(time.Minute * 5 / 2)
+	check24(0.4, 125)
+
+	// Larger start rate tests undeflow handling.
+	startCandle.StartRate, startCandle.EndRate = startCandle.EndRate, startCandle.StartRate
+	check24(0.4, 125)
```
chappjc, line 194: Confirmed it fails without the patch with `wrong delta24. expected 0.400, got, -1.000`.
"underflow" spelled wrong, but np. :)

chappjc (approved):

PR #1877, opened 2022-09-29
[v0.5] server/db: Move cancel fields explicitly.

JoeGruffins:
This resolves a possible issue where fields may shift when moving.


buck54321:
This is necessary because the column order of upgraded DBs is not the same as new, and we were explicitly specifying an order of return for the columns in the subquery, but not for insertion.

buck54321:
Oh crap. I lied, @JoeGruff. The free cancels stuff didn't make it into 0.5 yet. Sorry.

Issue #1874, opened 2022-09-28
ui: hide advanced order options by default?


buck54321:
There was some interest in hiding the advanced order options by default. I'm open to that. We would want to take care with the design and the messaging.

Just an option to consider. I'm find with leaving it too.

chappjc:
I get hiding the fee bump sliders since most of the time they shouldn't be used, and they are even a touch misleading (Faster Swaps and Faster Redemption is not accurate if blocks have free space anyway, making it just a waste of funds).

On the other hand, of come to really like seeing the current selection for the txsplit. In fact, it really highlights that a default for that at order time might be to unselect it if the unlock savings is a tiny tiny fraction of lot size).

vctt94:
Can work on that

PR #1885, opened 2022-10-03
client: hide advanced pre order options if not showing by default

vctt94:
This PR closes: #1874.

I added a `ShowByDefault` config option into `OrderOption struct`, which defines value to show by default. 

I am showing the txsplit option by default and hiding the others, as chapp showed interest in keep showing it. 

How it looks right now: 

hiding:
image

showing:
image

### DIFF
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -1325,7 +1325,6 @@ func (dcr *ExchangeWallet) PreSwap(req *asset.PreSwapForm) (*asset.PreSwap, erro
 				Description:  desc,
 				DefaultValue: 1.0,
 			},
-			ShowByDefault: false,
 			XYRange: &asset.XYRange{
 				Start: asset.XYRangePoint{
 					Label: "1X",
@@ -1426,13 +1425,13 @@ func (dcr *ExchangeWallet) splitOption(req *asset.PreSwapForm, utxos []*composit
 	cfg := dcr.config()
 	return &asset.OrderOption{
 		ConfigOption: asset.ConfigOption{
-			Key:          splitKey,
-			DisplayName:  "Pre-size Funds",
-			Description:  desc,
-			DefaultValue: cfg.useSplitTx,
-			IsBoolean:    true,
+			Key:           splitKey,
+			DisplayName:   "Pre-size Funds",
+			Description:   desc,
+			DefaultValue:  cfg.useSplitTx,
+			IsBoolean:     true,
+			ShowByDefault: true,
 		},
-		ShowByDefault: true,
 		Boolean: &asset.BooleanConfig{
 			Reason: reason,
 		},
@@ -1481,7 +1480,6 @@ func (dcr *ExchangeWallet) PreRedeem(req *asset.PreRedeemForm) (*asset.PreRedeem
 			Description:  "Bump the redemption transaction fees up to 2x for faster confirmations on your redemption transaction.",
 			DefaultValue: 1.0,
 		},
-		ShowByDefault: false,
 		XYRange: &asset.XYRange{
 			Start: asset.XYRangePoint{
 				Label: "1X",
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -1325,6 +1325,7 @@ func (dcr *ExchangeWallet) PreSwap(req *asset.PreSwapForm) (*asset.PreSwap, erro
 				Description:  desc,
 				DefaultValue: 1.0,
 			},
+			ShowByDefault: false,
 			XYRange: &asset.XYRange{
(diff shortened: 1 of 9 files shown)


client/asset/options.go
```diff
@@ -40,4 +40,6 @@ type OrderOption struct {
 	// Range indicates a numeric input where the user can adjust the value
 	// between a pre-defined low and high.
 	XYRange *XYRange `json:"xyRange,omitempty"`
+
+	ShowByDefault bool `json:"showByDefault,omitempty"`
```
buck54321, line 44: We've typically expanded functionality by adding fields to the `ConfigOption`. Any reason that we can't continue to follow that pattern?

client/asset/dcr/dcr.go
```diff
@@ -1414,6 +1416,7 @@ func (dcr *ExchangeWallet) PreRedeem(req *asset.PreRedeemForm) (*asset.PreRedeem
 			Description:  "Bump the redemption transaction fees up to 2x for faster confirmations on your redemption transaction.",
 			DefaultValue: 1.0,
 		},
+		ShowByDefault: false,
```
buck54321, line 1419: No need to specify unless `true`, I don't think.

client/asset/dcr/dcr.go
```diff
@@ -1366,6 +1367,7 @@ func (dcr *ExchangeWallet) splitOption(req *asset.PreSwapForm, utxos []*composit
 			DefaultValue: cfg.useSplitTx,
 			IsBoolean:    true,
 		},
+		ShowByDefault: true,
```
buck54321, line 1370: Just Decred? Why no Bitcoin & clones?

buck54321 (commented):
It's not showing split option by default for me.
image

client/asset/options.go
vctt94: moved showByDefault to `ConfigOption`

client/webserver/site/src/js/markets.ts
```diff
@@ -881,6 +884,20 @@ export default class MarketsPage extends BasePage {
     page.candleVol.textContent = Doc.formatCoinValue(candle.matchVolume, this.market.baseUnitInfo)
   }
 
+  showPreOrderAdvancedOptions () {
+    this.preOrder(this.parseOrder(), true)
```
chappjc, line 888: @buck54321 I'm just checking out this PR for the first time now since https://github.com/decred/dcrdex/issues/1874 was your issue, but does it seem right to you that toggling the advanced options should trigger a `fetchPreorder` (esp. if hiding)?  I'm not seeing why the "advanced" options cannot just be un-hidden.

buck54321: No, we shouldn't need to do this just when showing/hiding.

buck54321 (changes requested):
Need to bump the cache buster.

chappjc:
Isn't this still doing a `fetchPreorder` when just toggling these advanced options *display*?

client/webserver/site/src/js/markets.ts
```diff
@@ -1659,6 +1659,33 @@ export default class MarketsPage extends BasePage {
...
+    Doc.hide(page.showAdvancedOptions)
+    Doc.show(page.hideAdvancedOptions)
+    this.reloadOrderOpts(order, swap, redeem, true, changed)
+  }
+
+  hidePreOrderAdvancedOptions (order: TradeForm, swap: PreSwap, redeem: PreRedeem, changed: ()=>void) {
+    const page = this.page
+    Doc.hide(page.hideAdvancedOptions)
+    Doc.show(page.showAdvancedOptions)
+    this.reloadOrderOpts(order, swap, redeem, false, changed)
```
buck54321, line 1673: We shouldn't need this. What were you seeing that required a reload?

vctt94: Right now we only keep `page.vOrderOpts` in memory, so I have no way to know if it should be showed by default or not if I only hide or show the child node

buck54321: I see now. A preferable pattern would be to create and insert all of the elements, and selectively `Doc.{hide,show}` them based on the user's selection.

vctt94: that user selection would be some settings like "show by default" with a checkbox? What do you have in mind?

buck54321: check out https://github.com/buck54321/dcrdex/compare/97f7f3ab9ed3a05282e7098d637e9af9d2c28fa9...buck54321:dcrdex:defaultopts-alt

vctt94: Thanks for the suggestion, pushed it on my PR

buck54321 (approved):

PR #1887, opened 2022-10-05
multi: implement order queueing

### LINKED ISSUE #1862: ui,core: order placement could be partly async
Especially when using a split tx, in which the funding output must propagate to the server node while we wait, the ordering sequence could be broken into an synchronous "funding" part and an asynchronous "comms" part. The benefit being that from the client's perspective, the ordering would appear to happen quickly, and the part most likely to error is the `FundOrder` part anyway.

The flow would be something like this.

1. **frontend**: User initiates order via `(*WebServer).apiTrade`.
2. **core** :Run `FundOrder` and other order preparation up to but not including sending the order.
3.  **core**: Generate a temporary ID for the order (since we don't have a server timestamp yet).
4. **core**: Start sending the order asynchronously.
5. **core**: Return the temporary ID to the caller waiting on `apiTrade`
6. **frontend**: Receive temporary ID. Close forms. Show order in "my orders" table, with pending indicator.
7. **core**: When ordering is complete, send result to client via core notification, including both temporary ID and newly finalized order ID
8. **frontend**: Receive result, updating order ID on succes. On failure, error goes to notification bell but also show some kind of modal error dialog too.

ukane-philemon:
This enables support for order queueing only for orders placed from the client UI. Closes #1862
Screenshot 2022-10-27 at 9 27 25 PM

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -141,6 +141,9 @@ type dexConnection struct {
 	tradeMtx sync.RWMutex
 	// trades tracks outstanding orders issued by this client.
 	trades map[order.OrderID]*trackedTrade
+	// inFlightOrders tracks orders issued by this client that have not been
+	// processed by a dex server.
+	inFlightOrders map[uint64]*InFlightOrder
 
 	epochMtx sync.RWMutex
 	epoch    map[string]uint64
@@ -282,11 +285,35 @@ func coreMarketFromMsgMarket(dc *dexConnection, msgMkt *msgjson.Market) *Market 
 		AtomToConv:      float64(qconv) / float64(bconv),
 	}
 
-	for _, trade := range dc.marketTrades(mkt.marketName()) {
+	trades, inFlight := dc.marketTrades(mkt.marketName())
+	mkt.InFlightOrders = inFlight
+
+	for _, trade := range trades {
 		mkt.Orders = append(mkt.Orders, trade.coreOrder())
 	}
 
 	return mkt
+}
+
+// temporaryOrderIDCounter for inflight orders.
+var temporaryOrderIDCounter uint64
+
+// storeInFlightOrder stores an inflight order and returns a generated ID.
+func (dc *dexConnection) storeInFlightOrder(ord *Order) uint64 {
+	tempID := atomic.AddUint64(&temporaryOrderIDCounter, 1)
+	dc.tradeMtx.Lock()
+	dc.inFlightOrders[tempID] = &InFlightOrder{
+		Order:       ord,
+		TemporaryID: tempID,
+	}
+	dc.tradeMtx.Unlock()
+	return tempID
+}
+
+func (dc *dexConnection) deleteInFlightOrder(tempID uint64) {
+	dc.tradeMtx.Lock()
+	delete(dc.inFlightOrders, tempID)
+	dc.tradeMtx.Unlock()
 }
 
 func (dc *dexConnection) trackedTrades() []*trackedTrade {
@@ -299,8 +326,9 @@ func (dc *dexConnection) trackedTrades() []*trackedTrade {
 	return allTrades
 }
 
-// marketTrades is a slice of active trades in the trades map.
-func (dc *dexConnection) marketTrades(mktID string) []*trackedTrade {
+// marketTrades returns a slice of active trades in the trades map and a slice
+// of inflight orders in the inFlightOrders map.
+func (dc *dexConnection) marketTrades(mktID string) ([]*trackedTrade, []*InFlightOrder) {
 	// Copy trades to avoid locking both tradeMtx and trackedTrade.mtx.
 	allTrades := dc.trackedTrades()
 	trades := make([]*trackedTrade, 0, len(allTrades)) // may over-allocate
@@ -310,7 +338,16 @@ func (dc *dexConnection) marketTrades(mktID string) []*trackedTrade {
 		}
 		// Retiring inactive orders is presently the responsibility of ticker.
 	}
-	return trades
+
+	dc.tradeMtx.RLock()
+	inFlight := make([]*InFlightOrder, 0, len(dc.inFlightOrders)) // may over-allocate
+	for _, ord := range dc.inFlightOrders {
+		if ord.MarketID == mktID {
+			inFlight = append(inFlight, ord)
+		}
+	}
+	dc.tradeMtx.RUnlock()
+	return trades, inFlight
 }
 
 func (dc *dexConnection) setPendingFee(asset, confs uint32) {
@@ -425,6 +462,12 @@ func (dc *dexConnection) hasActiveAssetOrders(assetID uint32) bool {
 	dc.tradeMtx.RLock()
 	defer dc.tradeMtx.RUnlock()
 	familial := assetFamily(assetID)
+	for _, inFlight := range dc.inFlightOrders {
+		if familial[inFlight.BaseID] || familial[inFlight.QuoteID] {
+			return true
+		}
+	}
+
 	for _, trade := range dc.trades {
 		if (familial[trade.Base()] || familial[trade.Quote()]) &&
 			trade.isActive() {
@@ -445,7 +488,7 @@ func (dc *dexConnection) hasActiveOrders() bool {
 			return true
 		}
 	}
-	return false
+	return len(dc.inFlightOrders) > 0
 }
 
 // findOrder returns the tracker and preimage for an order ID, and a boolean
@@ -1101,7 +1144,7 @@ func (c *Core) dex(addr string) (*dexConnection, bool, error) {
 	return dc, dc.status() == comms.Connected, nil
 }
 
-// Get the *dexConnection for the the host. Return an error if the DEX is not
+// Get the *dexConnection for the host. Return an error if the DEX is not
 // connected.
 func (c *Core) connectedDEX(addr string) (*dexConnection, error) {
 	dc, connected, err := c.dex(addr)
@@ -4332,16 +4375,12 @@ func (c *Core) Order(oidB dex.Bytes) (*Order, error) {
 	if err != nil {
 		return nil, err
 	}
-	// See if its an active order first.
-	var tracker *trackedTrade
+	// See if it's an active order first.
 	for _, dc := range c.dexConnections() {
-		tracker, _, _ = dc.findOrder(oid)
+		tracker, _, _ := dc.findOrder(oid)
 		if tracker != nil {
-			break
+			return tracker.coreOrder(), nil
 		}
-	}
-	if tracker != nil {
-		return tracker.coreOrder(), nil
 	}
 	// Must not be an active order. Get it from the database.
 	mOrd, err := c.db.Order(oid)
@@ -4987,6 +5026,72 @@ func (c *Core) PreOrder(form *TradeForm) (*OrderEstimate, error) {
 
 // Trade is used to place a market or limit order.
 func (c *Core) Trade(pw []byte, form *TradeForm) (*Order, error) {
+	dc, crypter, err := c.parseTradeCredentials(pw, form)
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := c.prepareTrackedTrade(dc, form, crypter)
+	if err != nil {
+		return nil, err
+	}
+
+	corder, err := c.sendTradeRequest(req)
+	if err != nil {
+		return nil, err
+	}
+
+	return corder, nil
+}
+
+// TradeAsync is like Trade but a temporary order is returned before order
+// server validation. This helps handle some issues related to UI/UX where
+// server response might take a fairly long time (15 - 20s).
+func (c *Core) TradeAsync(pw []byte, form *TradeForm) (*InFlightOrder, error) {
+	dc, crypter, err := c.parseTradeCredentials(pw, form)
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := c.prepareTrackedTrade(dc, form, crypter)
+	if err != nil {
+		return nil, err
+	}
+
+	// Prepare and store the inflight order.
+	corder := coreOrderFromTrade(req.dbOrder.Order, req.dbOrder.MetaData)
+	tempID := dc.storeInFlightOrder(corder)
+	req.tempID = tempID
+
+	// Send silent note for the async order. This improves the UI/UX, so
+	// users don't have to wait for orders especially split tx orders.
+	c.notify(newOrderNoteWithTempID(TopicAsyncOrderSubmitted, "", "", db.Data, corder, tempID))
+
+	c.wg.Add(1)
+	go func() { // so core does not shut down while processing this order.
+		defer func() {
+			// Cleanup when the inflight order has been processed.
+			dc.deleteInFlightOrder(tempID)
+			c.wg.Done()
+		}()
+
+		_, err := c.sendTradeRequest(req)
+		if err != nil {
+			// Send async order error note.
+			topic := TopicAsyncOrderFailure
+			subject, details := c.formatDetails(topic, tempID, err)
+			c.notify(newOrderNoteWithTempID(topic, subject, details, db.ErrorLevel, corder, tempID))
+		}
+	}()
+
+	return &InFlightOrder{
+		corder,
+		tempID,
+	}, nil
+}
+
+// parseTradeCredentials ensures that trade requirements are met.
+func (c *Core) parseTradeCredentials(pw []byte, form *TradeForm) (*dexConnection, encrypt.Crypter, error) {
 	// Check the user password. A Trade can be attempted with an empty password,
 	// which should work if both wallets are unlocked. We use this feature for
 	// bots.
@@ -4995,55 +5100,61 @@ func (c *Core) Trade(pw []byte, form *TradeForm) (*Order, error) {
 		var err error
 		crypter, err = c.encryptionKey(pw)
 		if err != nil {
-			return nil, fmt.Errorf("Trade password error: %w", err)
+			return nil, nil, fmt.Errorf("Trade password error: %w", err)
 		}
 		defer crypter.Close()
 	}
 	dc, err := c.connectedDEX(form.Host)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	if dc.acct.suspended() {
-		return nil, newError(suspendedAcctErr, "may not trade while account is suspended")
+		return nil, nil, newError(suspendedAcctErr, "may not trade while account is suspended")
 	}
-
-	corder, updatedAssets, err := c.prepareTrackedTrade(dc, form, crypter)
-	if err != nil {
-		return nil, err
-	}
-
-	for assetID := range updatedAssets {
-		c.updateAssetBalance(assetID)
-	}
-
-	return corder, nil
+	return dc, crypter, nil
 }
 
-// Send an order, process result, prepare and store the trackedTrade.
-func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter encrypt.Crypter) (*Order, assetMap, error) {
+// tradeRequest hold all the information required to send a trade request to a
+// server.
+type tradeRequest struct {
+	mktID, route string
+	dc           *dexConnection
+	preImg       order.Preimage
+	form         *TradeForm
+	dbOrder      *db.MetaOrder
+	msgOrder     msgjson.Stampable
+	coins        asset.Coins
+	recoveryCoin asset.Coin
+	wallets      *walletSet
+	errCloser    *dex.ErrorCloser
+	tempID       uint64
+}
+
+// prepareTrackedTrade prepares a trade request.
+func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter encrypt.Crypter) (*tradeRequest, error) {
 	mktID := marketName(form.Base, form.Quote)
 	mktConf := dc.marketConfig(mktID)
 	if mktConf == nil {
-		return nil, nil, newError(marketErr, "order placed for unknown market %q", mktID)
+		return nil, newError(marketErr, "order placed for unknown market %q", mktID)
 	}
 
 	// Proceed with the order if there is no trade suspension
 	// scheduled for the market.
 	if !dc.running(mktID) {
-		return nil, nil, newError(marketErr, "%s market trading is suspended", mktID)
+		return nil, newError(marketErr, "%s market trading is suspended", mktID)
 	}
 
 	rate, qty := form.Rate, form.Qty
 	if form.IsLimit && rate == 0 {
-		return nil, nil, newError(orderParamsErr, "zero-rate order not allowed")
+		return nil, newError(orderParamsErr, "zero-rate order not allowed")
 	}
 
 	wallets, assetConfigs, versCompat, err := c.walletSet(dc, form.Base, form.Quote, form.Sell)
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 	if !versCompat { // also covers missing asset config, but that's unlikely since there is a market config
-		return nil, nil, fmt.Errorf("client and server asset versions are incompatible for %v", dc.acct.host)
+		return nil, fmt.Errorf("client and server asset versions are incompatible for %v", dc.acct.host)
 	}
 
 	fromWallet, toWallet := wallets.fromWallet, wallets.toWallet
@@ -5075,18 +5186,18 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
 
 	err = prepareWallet(fromWallet)
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 
 	err = prepareWallet(toWallet)
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 
 	// Get an address for the swap contract.
 	redeemAddr, err := toWallet.RedemptionAddress()
 	if err != nil {
-		return nil, nil, codedError(walletErr, fmt.Errorf("%s RedemptionAddress error: %w",
+		return nil, codedError(walletErr, fmt.Errorf("%s RedemptionAddress error: %w",
 			assetConfigs.toAsset.Symbol, err))
 	}
 
@@ -5127,16 +5238,16 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
 							"qty = %d %s, mid-gap = %d, base-qty = %d %s, lot size = %d",
 						qty, assetConfigs.quoteAsset.Symbol, midGap, baseQty,
 						assetConfigs.baseAsset.Symbol, lotSize)
-					return nil, nil, err
+					return nil, err
 				}
 			} else if isAccountRedemption {
-				return nil, nil, newError(orderParamsErr, "cannot estimate redemption count")
+				return nil, newError(orderParamsErr, "cannot estimate redemption count")
 			}
 		}
 	}
 
 	if lots == 0 {
-		return nil, nil, newError(orderParamsErr, "order quantity < 1 lot. qty = %d %s, rate = %d, lot size = %d",
+		return nil, newError(orderParamsErr, "order quantity < 1 lot. qty = %d %s, rate = %d, lot size = %d",
 			qty, assetConfigs.baseAsset.Symbol, rate, lotSize)
 	}
 
@@ -5152,7 +5263,7 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
 		RedeemAssetID: assetConfigs.toAsset.ID,
 	})
 	if err != nil {
-		return nil, nil, codedError(walletErr, fmt.Errorf("FundOrder error for %s, funding quantity %d (%d lots): %w",
+		return nil, codedError(walletErr, fmt.Errorf("FundOrder error for %s, funding quantity %d (%d lots): %w",
 			assetConfigs.fromAsset.Symbol, fundQty, lots, err))
 	}
 
@@ -5175,16 +5286,15 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
 
 	// The coins selected for this order will need to be unlocked
 	// if the order does not get to the server successfully.
-	var success bool
-	defer func() {
-		if success {
-			return
-		}
+	errCloser := dex.NewErrorCloser()
+	defer errCloser.Done(c.log)
+	errCloser.Add(func() error {
 		err := fromWallet.ReturnCoins(coins)
 		if err != nil {
-			c.log.Warnf("Unable to return %s funding coins: %v", unbip(fromWallet.AssetID), err)
+			return fmt.Errorf("Unable to return %s funding coins: %v", unbip(fromWallet.AssetID), err)
 		}
-	}()
+		return nil
+	})
 
 	// Construct the order.
 	preImg := newPreimage()
@@ -5227,12 +5337,12 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
 	}
 	err = order.ValidateOrder(ord, order.OrderStatusEpoch, lotSize)
 	if err != nil {
-		return nil, nil, fmt.Errorf("ValidateOrder error: %w", err)
+		return nil, fmt.Errorf("ValidateOrder error: %w", err)
 	}
 
 	msgCoins, err := messageCoins(wallets.fromWallet, coins, redeemScripts)
 	if err != nil {
-		return nil, nil, fmt.Errorf("%v wallet failed to sign coins: %w", assetConfigs.fromAsset.Symbol, err)
+		return nil, fmt.Errorf("%v wallet failed to sign coins: %w", assetConfigs.fromAsset.Symbol, err)
 	}
 
 	// Everything is ready. Send the order.
@@ -5244,25 +5354,24 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
 	if isAccountRedemption {
 		pubKeys, sigs, err := toWallet.SignMessage(nil, msgOrder.Serialize())
 		if err != nil {
-			return nil, nil, codedError(signatureErr, fmt.Errorf("SignMessage error: %w", err))
+			return nil, codedError(signatureErr, fmt.Errorf("SignMessage error: %w", err))
 		}
 		if len(pubKeys) == 0 || len(sigs) == 0 {
-			return nil, nil, newError(signatureErr, "wrong number of pubkeys or signatures, %d & %d", len(pubKeys), len(sigs))
+			return nil, newError(signatureErr, "wrong number of pubkeys or signatures, %d & %d", len(pubKeys), len(sigs))
 		}
 		redemptionReserves, err = accountRedeemer.ReserveNRedemptions(redemptionRefundLots,
 			assetConfigs.toAsset.Version, assetConfigs.toAsset.MaxFeeRate)
 		if err != nil {
-			return nil, nil, codedError(walletErr, fmt.Errorf("ReserveNRedemptions error: %w", err))
+			return nil, codedError(walletErr, fmt.Errorf("ReserveNRedemptions error: %w", err))
 		}
 		msgTrade.RedeemSig = &msgjson.RedeemSig{
 			PubKey: pubKeys[0],
 			Sig:    sigs[0],
 		}
-		defer func() {
-			if !success {
-				accountRedeemer.UnlockRedemptionReserves(redemptionReserves)
-			}
-		}()
+		errCloser.Add(func() error {
+			accountRedeemer.UnlockRedemptionReserves(redemptionReserves)
+			return nil
+		})
 	}
 
 	// If the from asset is an AccountLocker, we need to lock up refund funds.
@@ -5271,13 +5380,12 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
 		refundReserves, err = accountRefunder.ReserveNRefunds(redemptionRefundLots,
 			assetConfigs.fromAsset.Version, assetConfigs.fromAsset.MaxFeeRate)
 		if err != nil {
-			return nil, nil, codedError(walletErr, fmt.Errorf("ReserveNRefunds error: %w", err))
+			return nil, codedError(walletErr, fmt.Errorf("ReserveNRefunds error: %w", err))
 		}
-		defer func() {
-			if !success {
-				accountRefunder.UnlockRefundReserves(refundReserves)
-			}
-		}()
+		errCloser.Add(func() error {
+			accountRefunder.UnlockRefundReserves(refundReserves)
+			return nil
+		})
 	}
 
 	// A non-nil changeID indicates that this is an account based coin. The
@@ -5285,7 +5393,7 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
 	// be signed with that address's private key.
 	if changeID != nil {
 		if _, msgTrade.Coins[0].Sigs, err = fromWallet.SignMessage(nil, msgOrder.Serialize()); err != nil {
-			return nil, nil, fmt.Errorf("%v wallet failed to sign for redeem: %w",
+			return nil, fmt.Errorf("%v wallet failed to sign for redeem: %w",
 				assetConfigs.fromAsset.Symbol, err)
 		}
 	}
@@ -5296,41 +5404,10 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
 	c.sentCommits[prefix.Commit] = commitSig
 	c.sentCommitsMtx.Unlock()
 
-	// Send and get the result.
-	result := new(msgjson.OrderResult)
-	err = dc.signAndRequest(msgOrder, route, result, fundingTxWait+DefaultResponseTimeout)
-	if err != nil {
-		// At this point there is a possibility that the server got the request
-		// and created the trade order, but we lost the connection before
-		// receiving the response with the trade's order ID. Any preimage
-		// request will be unrecognized. This order is ABANDONED.
-		return nil, nil, fmt.Errorf("new order request with DEX server %v market %v failed: %w", dc.acct.host, mktID, err)
-	}
-
-	// If we encounter an error, perform some basic logging.
-	logAbandon := func(err interface{}) {
-		c.log.Errorf("Abandoning order. preimage: %x, server time: %d: %v",
-			preImg[:], result.ServerTime, err)
-	}
-
-	err = validateOrderResponse(dc, result, ord, msgOrder) // stamps the order, giving it a valid ID
-	if err != nil {
-		logAbandon(fmt.Sprintf("order response validation failure: %v", err))
-		return nil, nil, fmt.Errorf("validateOrderResponse error: %w", err)
-	}
-
-	// TODO: Need xcWallet fields for acceptable SwapConf values: a min
-	// acceptable for security, and even a max confs override to act sooner.
-
-	// Store the order.
+	// Prepare order meta data.
 	dbOrder := &db.MetaOrder{
 		MetaData: &db.OrderMetaData{
-			Status: order.OrderStatusEpoch,
-			Host:   dc.acct.host,
-			Proof: db.OrderProof{
-				DEXSig:   result.Sig,
-				Preimage: preImg[:],
-			},
+			Host:               dc.acct.host,
 			EpochDur:           mktConf.EpochLen, // epochIndex := result.ServerTime / mktConf.EpochLen
 			FromSwapConf:       assetConfigs.fromAsset.SwapConf,
 			ToSwapConf:         assetConfigs.toAsset.SwapConf,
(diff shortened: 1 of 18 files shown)


client/core/core.go
```diff
@@ -5098,112 +5114,191 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
...
+	}
+
+	if form.Queue {
+		meta := &db.OrderMetaData{
+			Status:     order.OrderStatusPending,
+			Host:       form.Host,
+			ChangeCoin: changeID,
+		}
+
+		corder := coreOrderFromTrade(ord, meta, true)
```
chappjc, line 5194: Having `coreOrderFromTrade` assign a random temp ID isn't desirable.  `coreOrderFromTrade` is intended to be a transform function and should be deterministic.  If it's called again elsewhere in the code, the ID is different.

I need to consider more carefully, but I believe the temp ID needs to be stored with the `OrderMetaData`

dex/order/status.go
```diff
@@ -55,6 +55,9 @@ const (
 	// standing limit orders that were matched but have failed to swap (neither
 	// executed nor canceled), and preimage misses.
 	OrderStatusRevoked
+
+	// OrderStatusPending is for queued orders.
+	OrderStatusPending
```
chappjc, line 60: We (unfortunately) do a lot of inequalities when deciding about order status.  e.g. of `if ord.Status > order.OrderStatusBooked { /* order is done */ }`.

However, prepending a new status is no good because that changes the value of all the other status and the orders stored in the DB are all different statuses effectively, and we absolutely do not want to solve that with an enormous DB upgrade that patches all existing order statuses.

Need to think about this more, but maybe we just root out all inequality based comparisons in favor of explicitly naming status. and make docs clear about the perils of doing that.

Grep the code for:
- `>= order.OrderStat`
- `> order.OrderStat`
- `<= order.OrderStat`
- `< order.OrderStat`

client/webserver/site/src/js/markets.ts
```diff
@@ -1219,8 +1219,15 @@ export default class MarketsPage extends BasePage {
         header.expander.classList.remove('ico-arrowdown')
         header.expander.classList.add('ico-arrowup')
       })
-      details.link.href = `order/${ord.id}`
-      app().bindInternalNavigation(div)
+      if (ord.status === OrderUtil.StatusPending) {
+        Doc.show(details.accelerateBttn)
```
chappjc, line 1223: Did you mean to *hide*? the accelerate button?  There's no swap yet to accelerate if the order is pending.

client/core/types.go
```diff
@@ -422,6 +421,13 @@ func coreOrderFromTrade(ord order.Order, metaData *db.OrderMetaData) *Order {
 		AccelerationCoins: accelerationCoins,
 	}
 
+	if !temp {
+		corder.Stamp = uint64(prefix.ServerTime.UnixMilli())
+		corder.ID = ord.ID().Bytes()
+	} else {
+		corder.ID = encode.RandomBytes(order.OrderIDSize)
```
chappjc, line 428: Related to the other comment about not having `coreOrderFromTrade` generate the random ID (it needs to be a deterministic function), it's not good to have the `ID` field change when the order goes from pending -> concrete.

How come `coreOrderFromTrade` leaves the `TempID` field zero?

dex/order/status.go
chappjc: I think we should consider a different approach that does not involve a new status defined here (which wouldn't have any meaning for server BTW).  The "pending" nature of the order could be **inferred** by other OrderMetaData fields, or perhaps just by the lack of the real ID and instead only a TempID.

chappjc (commented):
Good start.  I know it's in draft, but I want to point out some pitfalls and possible alternatives before you polish it up.

dex/order/status.go
ukane-philemon: > That logic sounds simple enough for frontend (isPending = !o.id && o.tempid or whatever is proper)

I think this is beter too.

ukane-philemon:
The tests are flaky, but I'd appreciate a review.

chappjc:
This looks like a reasonable approach, with the caveat that simply refreshing the page while waiting on a queued/async order would cause the order to vanish from the table until the actual accepted order notification arrived and made the table entry again.
I'm thinking this is a reasonable compromise because making these transient orders persistent requires the tempID to be in `OrderMetaData`, and stored with a `db.MetaOrder`, and a special container for these types of partial orders in the `User` object.  I think it's wise not to try to mix these incomplete orders with the others.

I haven't looked at the tests at all yet.

Regarding the changes to `Trade`, I had envisioned breaking `prepareTrackedTrade` into two methods: one to prep for submission by making a OrderMetaData skeleton, and another to do the submission and make+insert the `trackedTrade`.  But the channel approach you took also seems to make sense.

client/core/core.go
```diff
@@ -5244,9 +5298,12 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
...
 	}
 
-	success = true
+	for assetID := range updated {
+		c.updateAssetBalance(assetID)
+	}
 
-	return corder, updated, nil
+	success = true
+	return corder, err
```
chappjc, line 5306: `nil` instead of `err`?  If you don't know if `err` will be nil here, there should be an `if` statement to return differently depending on that.  It's ill-advised to return non-zero values when the error can be non-nil.

client/core/notification.go
```diff
@@ -269,6 +269,8 @@ const (
 	TopicFailedCancel         Topic = "FailedCancel"
 	TopicOrderLoaded          Topic = "OrderLoaded"
 	TopicOrderRetired         Topic = "OrderRetired"
+	TopicOrderError           Topic = "OrderError"
```
chappjc, line 272: I think this should be more specific.  `TopicQueueOrderFailure` or `TopicFailedOrderSubmission`, etc.
In the future something as generic sounding as `OrderError` will be misused.

client/core/core.go
ukane-philemon: yeah, should be `nil`.

ukane-philemon:
> This looks like a reasonable approach, with the caveat that simply refreshing the page while waiting on a queued/async order would cause the order to vanish from the table until the actual accepted order notification arrived and made the table entry again.

> and a special container for these types of partial orders in the User object

Yes, I thought about storing the `queued` orders somehow until a server response or failure and a page refresh seem to be one scenario that would need the `queued` orders until server response. However, the server response appears to be instant.

chappjc:
> However, the server response appears to be instant

Ah, we would not be doing this if this were always the case.

Especially with txsplit, order submission can take a long time, routinely 10 seconds or more.  This can be especially slow with a high latency asset network because propagation of this transaction can take a long time.

In general, if a method request, like `Trade` involves a server request with a long timeout (it is), we should plan for the request to take some time.

ukane-philemon:
> we should plan for the request to take some time.

Okay, will look into storing `queued` orders somewhere.

chappjc:
Having thought some more about it, I think we do want the pending order reflected in the User struct somewhere.  Consider a headless client that is an app that uses Core without the browser based ui, if it wanted to call the Trade method with the async submit option, it would be reasonable to expect that the returned pending order would be accounted for in the list of orders from User.  In an ideal world, no notifications would ever be missed and there'd be no races with the notifications arriving before recording the pending order, etc. but that's not how apps get written.

chappjc:
> > This looks like a reasonable approach, with the caveat that simply refreshing the page while waiting on a queued/async order would cause the order to vanish from the table until the actual accepted order notification arrived and made the table entry again.
> 
> > and a special container for these types of partial orders in the User object
> 
> Yes, I thought about storing the `queued` orders somehow until a server response or failure and a page refresh seem to be one scenario that would need the `queued` orders until server response. However, the server response appears to be instant.

Did you try a new field in User for pending orders in a different type?  I don't feel great about using a trackedTrade or a new order status for pending orders, but if it didn't work out... I guess

ukane-philemon:
> Did you try a new field in User for pending orders in a different type?

User is not stored on `core`. Plus having these temp orders stored like `trackedTrade`  reduces code change and the possibility of unexpected errors on the frontend, I think.

chappjc:
OK, I see.  Will review.  It's good that the pending orders are isolated in the `queuedTrades` maps.

client/core/core.go
```diff
@@ -5203,6 +5241,82 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
...
+		// request will be unrecognized. This order is ABANDONED.
+		return nil, failedOrderError(fmt.Errorf("new order request with DEX server %v market %v failed: %w", dc.acct.host, mktID, err))
+	}
+
+	// If we encounter an error, perform some basic logging.
+	logAbandon := func(err error) {
+		c.log.Errorf("Abandoning order. preimage: %x, server time: %d: %v", preImg[:], result.ServerTime, err)
+	}
+
+	err = validateOrderResponse(dc, result, ord, msgOrder) // stamps the order, giving it a valid ID
```
chappjc, line 5299: An issue with having the `order.Order` accessible to the app before this point via the `queuedTrades` map is that the `order.Order` is not ready, and only in `validateOrderResponse` do we stamp the order and compute+cache the order ID.  You have conditions in `coreOrder()`, but it's still a data race with the modifications here.

In the `trackedTrade`, it the `order.Order` is intended to be immutable, but it's is no longer.

If the `queuedTrades` map were a `map[order.OrderID]*db.OrderMetaData`, we wouldn't have that issue.

client/core/core.go
```diff
@@ -5203,6 +5241,82 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
...
+	if err != nil {
+		logAbandon(fmt.Errorf("validateOrderResponse error: %v", err))
+		return nil, failedOrderError(fmt.Errorf("order response validation failure: %v", err))
+	}
+
+	// Update tracker order meta.
+	tracker.metaData.Status = order.OrderStatusEpoch
+	tracker.metaData.Proof = db.OrderProof{
```
chappjc, line 5307: Need to lock `tracker.mtx` to access these fields because it's already in the `queuedTrades` map.

client/core/core.go
```diff
@@ -5203,6 +5241,82 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
...
+	// Store the validated and verified order.
+	err = c.db.UpdateOrder(dbOrder)
+	if err != nil {
+		logAbandon(fmt.Errorf("UpdateOrder error: %w", err))
+		return nil, failedOrderError(fmt.Errorf("failed to store order in database: %v", err))
+	}
+
+	// Store the tracker.
 	dc.tradeMtx.Lock()
 	dc.trades[tracker.ID()] = tracker
```
chappjc, line 5321: Very briefly, the `trackedTrade` is in both maps.  Maybe a single mutex for these maps, and the insert and delete in the same critical section.

client/core/core.go
```diff
@@ -230,6 +233,10 @@ func (dc *dexConnection) marketMap() map[string]*Market {
 		}
 		mktID := mkt.marketName()
 
+		for _, trade := range dc.marketQueuedTrades(mktID) {
+			mkt.Orders = append(mkt.Orders, trade.coreOrder())
```
chappjc, line 237: This is what I was referring to regarding a new field somewhere in the `User` struct (a new `Market.PendingOrders` field that is within the bigger `User` struct).  In https://github.com/decred/dcrdex/pull/1887#issuecomment-1279277001 you say "User is not stored on core", but I don't understand the meaning.

And if the queued orders are stored in a different type, we don't use `trade.coreOrder` and hopefully don't have to deal with the mutability issues mentioned in my other comment.  It can literally just be *copies* of an immutable `db.OrderMetaData` object. No need to modify `marketEpoch`, no other edge cases to check with the zero order time, etc.

And by putting pending orders in a completely separate field, there's no need for the new order status.

client/core/core.go
```diff
@@ -5203,6 +5241,82 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
 		tracker.changeLocked = true
 	}
 
+	if form.Queue {
+		tempID, err := order.IDFromBytes(encode.RandomBytes(order.OrderIDSize))
+		if err != nil {
+			return nil, fmt.Errorf("unable to generate temporary order ID: %w", err)
+		}
+		tracker.metaData.TempID = tempID
```
chappjc, line 5249: Just `rand.Read(tracker.metaData.TempID[:])` (using `crypto/rand`).  No need to use these helpers or another variable.

chappjc (changes requested):
Re-reviewed, and I still think it would be wise for many reasons not to use a `trackedTrade` for these partially-ready orders.

client/core/core.go
ukane-philemon: > And by putting pending orders in a completely separate field, there's no need for the new order status.

Putting them a `Market.PendingOrders` field works. I was looking at adding support for viewing temp orders on the `order/oid` page but no longer think it is necessary.

chappjc: You could get away with just one Orders slice and no new order status.
I think they important changes are too avoid trackedTrades in the queuedOrders map

buck54321:
Here's roughly what I was after for the **core** bits.

ukane-philemon:
> Here's roughly what I was after for the **core** bits. https://github.com/decred/dcrdex/compare/master...buck54321:async-login?expand=1

Thanks, nice suggestions using the `dex.NewErrorCloser`.

client/core/core.go
```diff
@@ -406,6 +431,30 @@ func (dc *dexConnection) hasActiveOrders() bool {
...
+
+func (dc *dexConnection) findQueuedOrder(oid order.OrderID) *Order {
+	dc.queuedOrdersMtx.RLock()
+	defer dc.queuedOrdersMtx.RUnlock()
+	return dc.queuedOrders[oid]
+}
+
+func (dc *dexConnection) hasQueuedOrders() bool {
+	dc.queuedOrdersMtx.RLock()
+	defer dc.queuedOrdersMtx.RLock()
```
chappjc, line 454: You've got `RLock` deferred instead of `RUnlock`

chappjc:
> Here's roughly what I was after for the **core** bits. https://github.com/decred/dcrdex/compare/master...buck54321:async-login?expand=1

I'm afraid we don't get the ability to list queued/processing orders with `(c *Core) TradeAsync(pw []byte, form *TradeForm) (uint64, error)`.
Whether it's a UI refresh or just a Go consumer listing their orders via the User or Exchange methods, the existence of such orders needs to knowable.

I believe the `queuedOrders.queuedOrders` map solves these issues, plus it informs methods like `(*dexConnection).hasActiveOrders`.

I think we were getting close to a workable solution at https://github.com/decred/dcrdex/commit/ab3a90266b6aec3639e19ee012cc521bb6a0c1ce.  I'm reluctant to brew a suggested diff of my own too, but with the current PR at  https://github.com/decred/dcrdex/commit/0f6577e402406faeffa915eccae7ac247097756f, the `(*Core).trade` doesn't have much purpose.

I'm fine changing from a `OrderForm.Queued` flag to two different methods (`Trade` and `TradeAsync`), but we just need to have theses pending orders tracked and enumerated.

ukane-philemon:
> I'm fine changing from a OrderForm.Queued flag to two different methods (Trade and TradeAsync), 

Do we really need to have another exported method?

buck54321:
> Whether it's a UI refresh or just a Go consumer listing their orders via the User or Exchange methods, the existence of such orders needs to knowable.

Even now with a fully synchronous order submission, I can hit refresh right after hitting submit and the order would not be in the `User` struct. I never saw a problem with that.

ukane-philemon:
> Even now with a fully synchronous order submission, I can hit refresh right after hitting submit and the order would not be in the User struct.

Well, I haven't done this for fully synchronous orders but if a response to the order API request has been received by the frontend we should be able to account for it in the interim.

chappjc:
> > Whether it's a UI refresh or just a Go consumer listing their orders via the User or Exchange methods, the existence of such orders needs to knowable.
> 
> Even now with a fully synchronous order submission, I can hit refresh right after hitting submit and the order would not be in the `User` struct. I never saw a problem with that. It's temporary and hard to hit in practice.

In that case, the Trade call hadn't even completed yet.  Now the request is fully handled and the caller has a promise.  There should not be a period after a response where the backend can provide no information about such orders-in-process.

This is not so hard to hit in practice, and "temporary" is up to 30 seconds.

We can go ahead with the simplified version, but I think we'll end up tracking these orders in the future.  It will lead to some very bad UX issues, including resubmitting orders repeatedly. What is the downside of doing it properly?  AFAICT, it's done.

client/core/core.go
```diff
@@ -4853,6 +4904,25 @@ func (c *Core) PreOrder(form *TradeForm) (*OrderEstimate, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	submissionResult := <-resultChan
+	if submissionResult.err != nil {
+		return nil, submissionResult.err
+	}
+
+	return submissionResult.ord, nil
+}
```
buck54321, line 4917: I'm not following. From the caller's perspective, it would seem that the `Queue bool` has no effect whatsoever.

buck54321:
> In that case, the Trade call hadn't even completed yet. 

Isn't this just semantics. I could say the trade hasn't completed until the notification is received too.

> Now the request is fully handled and the caller has a promise. There should not be a period after a response where the backend can provide no information about such orders-in-process.

I understand. The queued orders thing is just a different problem than I thought we were after here, but I guess it just evolved to address both issues. Looks like @ukane-philemon already adapted some bits. I do think `Core` should use two exported functions and roughly the API that I've suggested.

I'll note that I have only commented on **core** so far, but the UI also is not correct, because it gives no indication of a pending order after submission. So it's actually worse than before, when at least you had a spinner to tell you something was happening. The goal is to return the user to the markets view, yes, but we must show the user that something is waiting. But I wanted to work through the core API first since how that's done depends on the API and notification scheme.

chappjc:
> I could say the trade hasn't completed until the notification is received too.

The notifications are ancillary.  When the existing synchronous Trade returns, it is guaranteed that there will be an Order in all the Core methods that return such info.

Yes, the async order capability solves this when done right, but I don't think this is an evolution so much as a central issue with the existing synchronous order submission API.

> the UI also is not correct, because it gives no indication of a pending order after submission

I haven't tested this one bit, so I guess that's not done, but the goal would certainly be to put a row in the my orders table, which gets updated when the order submission response is received.

client/core/core.go
chappjc: Yes, I think the PR is WIP at the moment.  The diff suggested in https://github.com/decred/dcrdex/compare/master...buck54321:async-login?expand=1 is not compatible with the modified TradeForm

ukane-philemon:
Moving back to draft to add some changes related to suggestions.

ukane-philemon:
> the UI also is not correct, because it gives no indication of a pending order after submission

@buck54321, have you tested this on your end? 

Pending orders show on the markets page with URL to view the pending order but not on the `/orders` page. i.e they can't be retrieved via `core.Orders(...)`

However, the goal of this PR is to ensure order placement on the UI provides a better UX. Users can always view pending orders on the `markets` page (even if the page is refreshed).

client/core/core.go
```diff
@@ -4867,40 +4930,47 @@ func (c *Core) Trade(pw []byte, form *TradeForm) (*Order, error) {
...
 
-	corder, updatedAssets, err := c.prepareTrackedTrade(dc, form, crypter)
+	resultChan, err := c.prepareTrackedTrade(dc, form, crypter, queue)
 	if err != nil {
 		return nil, err
 	}
 
-	for assetID := range updatedAssets {
-		c.updateAssetBalance(assetID)
+	submissionResult := <-resultChan
```
chappjc, line 4938: This pattern *looks* fully-synchronous because it appears to wait for an "order submission result" regardless of `queue`, but really if `queue=true`, it's not an order submission result - the submission server request is still in process.

In other words, `prepareTrackedTrade` might as well just always return (`core.Order`, `error).  The channel appears pointless, misleading at best.

`prepareTrackedTrade` should just be split in two pieces so that when `queue` is `true`, it launches part 2 in a goroutine, and when `false`, it runs part 2 blocking.

ukane-philemon: >  /api/trade to return before the order submission request is fully handled.

I had tested about 10 sec delay after `err = dc.signAndRequest(msgOrder, route, result, fundingTxWait+DefaultResponseTimeout)` but not more than 10sec unless the order don't show up on orderbook.

`/api/trade` returns with temp order and the frontend get updated subsequently via note.

client/core/core.go
```diff
@@ -5146,106 +5214,164 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
...
-			ui.ConventionalString(corder.Qty), ui.Conventional.Unit, tracker.token())
-		c.notify(newOrderNote(TopicYoloPlaced, subject, details, db.Poke, corder))
-	} else {
-		rateString := "market"
-		if form.IsLimit {
-			rateString = wallets.trimmedConventionalRateString(corder.Rate)
+			// Return a temp order and send silent note for the queued order.
+			// This improves the UI/UX, so users don't have to wait for orders
+			// especially split tx orders.
+			resultChan <- &orderSubmissionResult{ord: corder}
```
chappjc, line 5273: How is the caller supposed to receive this unless they get the actual channel from the `return`?

ukane-philemon: With the current design which will eventually change, all callers of `prepareTrackedTrade` get the channel.

chappjc: But that's after the server request has completed, making this fully-synchronous.

ukane-philemon: The server request is launched in a go routine, then callers wait for the result sent via the channel. However, I updated the PR at https://github.com/decred/dcrdex/commit/a439114460eb1e4ba4ef599e33aebc573d92dabe

buck54321:
The primary goals then are to

1. Have an order submission path that returns before the order is submitted to the server. This allows us to hide the order form and the form overlay and return the user to the markets view. There will need to be a placeholder to show the pending order, probably just in the "user orders" list somehow.
2. Keep a record of orders that are being submitted, so that they don't disappear on refresh.

Based on these goals, I've taken another swing at **core**. I tried to keep the diff a little more readable in this one.

https://github.com/buck54321/dcrdex/compare/d606490ca154807d5895acbcee9ee4a38d3c7b4f...buck54321:dcrdex:async-trade

client/core/core.go
chappjc: I see. The commit just pushed makes more sense.

chappjc:
> Based on these goals, I've taken another swing at **core**. I tried to keep the diff a little more readable in this one.
> 
> [buck54321/dcrdex@d606490...buck54321:dcrdex:async-trade](https://github.com/buck54321/dcrdex/compare/d606490ca154807d5895acbcee9ee4a38d3c7b4f...buck54321:dcrdex:async-trade)

That's excellent, @buck54321.  I think @ukane-philemon has actually reached the proper behavior with his last push in https://github.com/decred/dcrdex/commit/a439114460eb1e4ba4ef599e33aebc573d92dabe, but what you have has some superior design elements, including splitting `prepareTrackedTrade` and embedding core.Order with temp ID field.

buck54321:
> [buck54321/dcrdex@d606490...buck54321:dcrdex:async-trade](https://github.com/buck54321/dcrdex/compare/d606490ca154807d5895acbcee9ee4a38d3c7b4f...buck54321:dcrdex:async-trade)

I probably wouldn't do this thing
```Go
msgOrder, dbOrder, route, errCloser, unstampedOrder := s.msgOrder, s.dbOrder, s.route, s.errCloser, s.unstampedOrder
recoveryCoin, form, isAccountRefund, isAccountRedemption := s.recoveryCoin, s.form, s.isAccountRefund, s.isAccountRedemption
ord, tempID := dbOrder.Order, unstampedOrder.TemporaryID
```
in practice, but it makes the diff cleaner.

chappjc:
> in practice, but it makes the diff cleaner.

Yah, I'm also a little squeamish about constructing a `trackedTrade` before a submission response is received, but since it's carefully scoped inside the `pendingSubmission struct`, which itself is carefully scoped to just the call stack of `trade`/`tradeAsync` (right?)  I can live with that.

ukane-philemon:
> https://github.com/buck54321/dcrdex/compare/d606490ca154807d5895acbcee9ee4a38d3c7b4f...buck54321:dcrdex:async-trade

Thanks for the suggestions. The `InFlightOrder ` type looks cool since it presents a better way to keep the `TempID` out of `core.Order`. But aside from tracking, we could have minor inconsistency if the order information is requested via the `api/order` endpoint in the interim and there's no `ID`, and no `TempID` on it.

RE: splitting prepareTrackedTrade

I'm trying to keep the changes minimal. However, if splitting `prepareTrackedTrade` is a priority, (which will require us to move a lot of things around and am not sure it is). I will make more improvements.

chappjc:
> However, if splitting `prepareTrackedTrade` is a priority, (which will require us to move a lot of things around and am not sure it is). I will make more improvements.

Well TBH, since in your last commit you solved this with a *closure* (`submitAndValidateOrder`), I actually think it keeps things a little cleaner by avoiding to have to package things in that `pendingSubmission` struct because the closure just captures everything.  I'll let you and @buck54321 sort out the differences, but I rather like how you've used the closure that way.

buck54321:
Take it or leave it. Mix-and-match. It's whatever. Just trying to keep things moving in the right direction.

ukane-philemon:
Alright, other notable changes I see @buck54321 suggestion address, include naming `queuedOrders > InFlightOrders` and keeping TempID out of `core.Order`.  I'd know if there's an APPROACH ACK especially about having `TempID` on `core.Order`.

buck54321:
> But aside from tracking, we could have minor inconsistency if the order information is requested via the api/order endpoint in the interim and there's no ID, and no TempID on it.

We wouldn't request an in-flight order from api/order, since we don't have an ID yet and the order is not in the `trades` map yet.

ukane-philemon:
> We wouldn't request an in-flight order from api/order, since we don't have an ID yet and the order is not in the trades map yet.

I added support for viewing inFlight orders with their TempID on the `/order/oid` page which will auto-redirect when the actual order arrives.

ukane-philemon:
> We wouldn't request an in-flight order from api/order, since we don't have an ID yet and the order is not in the trades map yet.

I added support for viewing inFlight orders with their TempID on the `/order/oid` page which will auto-redirect when the actual order arrives.

ukane-philemon:
> We wouldn't request an in-flight order from api/order, since we don't have an ID yet and the order is not in the trades map yet.

I added support for viewing inFlight orders with their TempID on the `/order/oid` page which will auto-redirect when the actual order arrives.

JoeGruffins (approved):

client/core/core.go
```diff
@@ -4867,40 +4916,36 @@ func (c *Core) Trade(pw []byte, form *TradeForm) (*Order, error) {
...
 	if err != nil {
 		return nil, err
 	}
 
-	for assetID := range updatedAssets {
-		c.updateAssetBalance(assetID)
-	}
-
 	return corder, nil
 }
```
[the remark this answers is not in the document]
ukane-philemon: Yeah, thanks for the review.

ukane-philemon:
@buck54321, what do you think this PR could improve on, with regards to your previous suggestions?

client/core/core.go
```diff
@@ -5146,106 +5183,144 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
...
+			Order:    ord,
+		}
 
-	tracker.redemptionLocked = tracker.redemptionReserves
-	tracker.refundLocked = tracker.refundReserves
+		err = c.db.UpdateOrder(dbOrder)
+		if err != nil {
+			logAbandon(fmt.Errorf("UpdateOrder error: %w", err))
+			return nil, fmt.Errorf("failed to store order in database: %v", err)
+		}
```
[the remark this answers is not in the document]
ukane-philemon: > This is the primary reason I broke this into two functions. Git diffs are nearly unreadable when you indent large blocks.

Ah, I see. Looking at how both methods are similar and only with the goroutine difference, I'm thinking if we can let this pass? https://github.com/decred/dcrdex/pull/1887#issuecomment-1292123164

client/webserver/site/src/js/app.ts
```diff
@@ -528,8 +529,33 @@ export default class Application {
...
+            return false
+          }
+          for (const i in mkt.inflight || []) {
+            if (mkt.inflight[i].tempID === tempID) {
+              mkt.inflight = mkt.inflight.filter((ord) => ord.tempID !== tempID)
+              return true
+            }
+          }
+        }
+        if (!updateInFlightOrders()) break
```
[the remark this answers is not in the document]
ukane-philemon: okay, I'd rather use an `if-else` block then.

client/core/core.go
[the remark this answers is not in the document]
chappjc: There a pros and cons to each approach.  I'll suggest some minor changes shortly to find a middle ground.  I started hacking and it looks like we can have a split method approach without the heavy duty structs only used to pass data on this one call stack.  Sorry for the delay.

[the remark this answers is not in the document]
chappjc: Well the changes wouldn't be so minor as I hoped... lots of data needed by the submit method. The suggested `pendingSubmission struct` really is just about as minimal as you can get.  I can work with either approach, the closure or the split methods with the new type to bus data.

Please just iterate on the other review requests, particularly the InflightOrder return, unless that complicates things somehow.

You may squash to a single commit before rebasing if you like.  Conflict resolution for this PR is much harder with the individual commits.

client/core/core.go
```diff
@@ -5146,106 +5183,144 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
...
+		tempID := dc.storeInFlightOrder(corder)
+		defer dc.deleteInFlightOrder(tempID)
 
-	return corder, updated, nil
+		// Send silent note for the async order. This improves the UI/UX, so
+		// users don't have to wait for orders especially split tx orders.
+		topic := TopicAsyncOrderSubmitted
+		c.notify(newOrderNoteWithTempID(topic, "", "", db.Data, corder, tempID))
+
+		_, err := submitAndValidateOrder(asyncErrorCloser, tempID)
```
chappjc, line 5312: I think you should get rid of both args to `submitAndValidateOrder`.  Declare `var tempID uint64` before defining the closure, and do `tempID = dc.storeInFlightOrder(corder)` (no `:=`) in the async branch, and just don't touch `tempID` in the synchronous branch.

chappjc (changes requested):

ukane-philemon:
I've moved some things around in [8839474](https://github.com/decred/dcrdex/pull/1887/commits/88394741b4484b47f450f2647a0a379c83b74eba), splitng `core.prepareTrackedTrade`.

client/core/core.go
```diff
@@ -4987,55 +5092,61 @@ func (c *Core) Trade(pw []byte, form *TradeForm) (*Order, error) {
...
+	recoveryCoin asset.Coin
+	wallets      *walletSet
+	errCloser    *dex.ErrorCloser
+	tempID       uint64
 }
 
-// Send an order, process result, prepare and store the trackedTrade.
-func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter encrypt.Crypter) (*Order, assetMap, error) {
+// prepareTrackedTrade prepares a trade request.
+func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter encrypt.Crypter) (*tradeRequest, error) {
```
chappjc, line 5126: `prepareTradeRequest`

client/core/core.go
```diff
@@ -4979,6 +5018,72 @@ func (c *Core) PreOrder(form *TradeForm) (*OrderEstimate, error) {
 
 // Trade is used to place a market or limit order.
 func (c *Core) Trade(pw []byte, form *TradeForm) (*Order, error) {
+	dc, crypter, err := c.parseTradeCredentials(pw, form)
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := c.prepareTrackedTrade(dc, form, crypter)
```
chappjc, line 5026: There no longer seems to be a reason to factor `parseTradeCredentials` out of `prepareTrackedTrade`.

However, an unintended side effect is that `crypter.Close()` seems to run far too soon now that it is deferred in `parseTradeCredentials` helper instead of after the crypter is finished in `prepareTrackedTrade`.

Maybe eliminate `parseTradeCredentials` and do this:

```diff
@@ -5026,12 +5030,7 @@ func (c *Core) PreOrder(form *TradeForm) (*OrderEstimate, error) {
 
 // Trade is used to place a market or limit order.
 func (c *Core) Trade(pw []byte, form *TradeForm) (*Order, error) {
-       dc, crypter, err := c.parseTradeCredentials(pw, form)
-       if err != nil {
-               return nil, err
-       }
-
-       req, err := c.prepareTrackedTrade(dc, form, crypter)
+       req, err := c.prepareTrackedTrade(form, pw)
        if err != nil {
                return nil, err
        }
@@ -5048,15 +5047,11 @@ func (c *Core) Trade(pw []byte, form *TradeForm) (*Order, error) {
 // server validation. This helps handle some issues related to UI/UX where
 // server response might take a fairly long time (15 - 20s).
 func (c *Core) TradeAsync(pw []byte, form *TradeForm) (*InFlightOrder, error) {
-       dc, crypter, err := c.parseTradeCredentials(pw, form)
-       if err != nil {
-               return nil, err
-       }
-
-       req, err := c.prepareTrackedTrade(dc, form, crypter)
+       req, err := c.prepareTrackedTrade(form, pw)
        if err != nil {
                return nil, err
        }
+       dc := req.dc
 
        // Prepare and store the inflight order.
        corder := coreOrderFromTrade(req.dbOrder.Order, req.dbOrder.MetaData)
@@ -5131,7 +5102,27 @@ type tradeRequest struct {
 }
 
 // prepareTrackedTrade prepares a trade request.
-func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter encrypt.Crypter) (*tradeRequest, error) {
+func (c *Core) prepareTrackedTrade(form *TradeForm, pw []byte) (*tradeRequest, error) {
... <paste contents of parseTradeCredentials> 
```

client/core/core.go
```diff
@@ -4987,55 +5092,61 @@ func (c *Core) Trade(pw []byte, form *TradeForm) (*Order, error) {
 		var err error
 		crypter, err = c.encryptionKey(pw)
 		if err != nil {
-			return nil, fmt.Errorf("Trade password error: %w", err)
+			return nil, nil, fmt.Errorf("Trade password error: %w", err)
 		}
 		defer crypter.Close()
```
chappjc, line 5097: Looks like `crypter.Close()` runs way too soon now that this code is put in the `parseTradeCredentials` helper.

Maybe you tested in a scenario where on-the-fly unlocking the wallet was not needed?

client/core/core.go
```diff
@@ -5288,114 +5396,142 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
...
+
+	errCloser.Success()
+
+	return tradeRequest, nil
+}
+
+// sendTradeRequest sends an order, processes the result, then prepares and
+// stores the trackedTrade.
+func (c *Core) sendTradeRequest(tr *tradeRequest) (*Order, error) {
+	defer tr.errCloser.Done(c.log)
```
chappjc, line 5441: This diff becomes quite a bit tamer with this at the start of `sendTradeRequest`:

```go
dc, dbOrder, wallets, form := tr.dc, tr.dbOrder, tr.wallets, tr.form
preImg, recoveryCoin := tr.preImg, tr.recoveryCoin
```

client/core/core.go
```diff
@@ -5288,114 +5396,142 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
...
 		// and created the trade order, but we lost the connection before
 		// receiving the response with the trade's order ID. Any preimage
 		// request will be unrecognized. This order is ABANDONED.
-		return nil, nil, fmt.Errorf("new order request with DEX server %v market %v failed: %w", dc.acct.host, mktID, err)
+		return nil, fmt.Errorf("new order request with DEX server %v market %v failed: %w", tr.dc.acct.host, tr.mktID, err)
 	}
 
 	// If we encounter an error, perform some basic logging.
-	logAbandon := func(err interface{}) {
+	logAbandon := func(err error) {
```
chappjc, line 5454: This was fine before, but if we were to make a change, we'd make it `msg string` instead, and keep the `fmt.Sprintf` lines.  There's no reason this needs to be a type that implements the `error` interface.  It's just logging.

Although TBH, I don't seen the point in `logAbandon` at all.  We should just log when needing to log (only in two places presently), not bounce around the call stack, worrying about argument types etc.

client/core/core.go
```diff
@@ -5288,114 +5396,142 @@ func (c *Core) prepareTrackedTrade(dc *dexConnection, form *TradeForm, crypter e
...
+		DEXSig:   result.Sig,
+		Preimage: tr.preImg[:],
 	}
-	err = c.db.UpdateOrder(dbOrder)
+
+	err = c.db.UpdateOrder(tr.dbOrder)
 	if err != nil {
-		logAbandon(fmt.Sprintf("failed to store order in database: %v", err))
-		return nil, nil, fmt.Errorf("Order abandoned due to database error: %w", err)
+		logAbandon(fmt.Errorf("UpdateOrder error: %w", err))
```
chappjc, line 5478: This message is now less informative. Before it was at least clear it was the DB.  Let's leave both `logAbandon` calls the same please, messages and Sprintf... or remove `logAbandon` as per my other comment.

client/core/core_test.go
```diff
@@ -226,7 +226,8 @@ func testDexConnection(ctx context.Context, crypter *tCrypter) (*dexConnection,
 			tUTXOAssetB.ID: tUTXOAssetB,
 			tACCTAsset.ID:  tACCTAsset,
 		},
-		books: make(map[string]*bookie),
+		inFlightOrders: make(map[uint64]*InFlightOrder),
```
chappjc, line 229: Strange position for this, should be down next to `trades:`

client/core/core_test.go
```diff
@@ -2689,7 +2701,8 @@ func TestSend(t *testing.T) {
 	}
 }
 
-func TestTrade(t *testing.T) {
+func trade(t *testing.T, async bool) {
+	t.Helper()
```
chappjc, line 2705: Isn't `t.Helper()` actually _undesirable_ in this use case?  The calling function has just one line, so we are interested in the line number in this function that failed, not the caller.

client/core/notification.go
```diff
@@ -242,7 +242,8 @@ func newSendNote(topic Topic, subject, details string, severity db.Severity) *Se
 // OrderNote is a notification about an order or a match.
 type OrderNote struct {
 	db.Notification
-	Order *Order `json:"order"`
+	Order       *Order `json:"order"`
+	TemporaryID uint64 `json:"tempID,omitempty"`
```
chappjc, line 246: OK with the caveat that we can never use 0

client/core/types.go
```diff
@@ -396,6 +403,12 @@ func coreOrderFromTrade(ord order.Order, metaData *db.OrderMetaData) *Order {
 		accelerationCoins = append(accelerationCoins, NewCoin(fromID, coinID))
 	}
 
+	// For in-flight orders, we'll set the order ID as a zero-hash.
```
chappjc, line 406: OK but will you comment why this is necessary instead of leaving nil? What specifically goes wrong if we don't do this?  Is it just for display?  Does a consumer recognize a zero hash vs zero length?

client/webserver/site/src/js/locales.ts
```diff
@@ -149,7 +150,8 @@ export const enUS: Locale = {
   [ID_INVALID_DATE_ERR_MSG]: 'error: invalid date or time',
   [ID_NO_ARCHIVED_RECORDS]: 'No archived records found',
   [ID_DELETE_ARCHIVED_RECORDS_RESULT]: 'Message: {{ nRecords }} archived records has been deleted',
-  [ID_ARCHIVED_RECORDS_PATH]: 'File Location: {{ path }}'
+  [ID_ARCHIVED_RECORDS_PATH]: 'File Location: {{ path }}',
+  [ID_ORDER_PENDING]: 'pending'
```
chappjc, line 154: Having tested this, I think it would be better to show this as "submitting" rather than "pending".  In a different sense, "epoch" orders are also "pending", so let's be more specific.

client/webserver/api.go
```diff
@@ -373,14 +373,14 @@ func (s *WebServer) apiTrade(w http.ResponseWriter, r *http.Request) {
...
+	ord, err := s.core.TradeAsync(pass, form.Order)
 	if err != nil {
 		s.writeAPIError(w, fmt.Errorf("error placing order: %w", err))
 		return
 	}
 	resp := &struct {
-		OK    bool        `json:"ok"`
-		Order *core.Order `json:"order"`
+		OK    bool                `json:"ok"`
+		Order *core.InFlightOrder `json:"order"`
```
chappjc, line 383: Instead of breaking the HTTP API by changing the return type and meaning of `/api/trade`, we should leave `/api/trade` and add `/api/tradeasync` or `/api/submittrade`

client/webserver/site/src/html/bodybuilder.tmpl
```diff
@@ -8,7 +8,7 @@
   <link rel="icon" href="/img/favicon.png?v=AK4XS4">
   <meta name="description" content="Decred DEX Client Web Portal">
   <title>{{.Title}}</title>
-  <link href="/css/style.css?v=LESXCTh9d" rel="stylesheet">
+  <link href="/css/style.css?v=GHSXZ15s" rel="stylesheet">
```
chappjc, line 11: There style bundle is unchanged.

chappjc (changes requested):
Tested and seems to be working, with some caveats.  The current design is overall quite good, I think.  Some minor changes needed.

client/core/core.go
ukane-philemon: > Maybe you tested in a scenario where on-the-fly unlocking the wallet was not needed?

Yeah, the testing wallet was always unlocked.

client/core/notification.go
ukane-philemon: Yes, we can never use zero.

client/webserver/site/src/js/locales.ts
ukane-philemon: okay.

client/core/notification.go
chappjc: Then should doc that at the site of the atomic declaration so that never happens.

client/core/core_test.go
ukane-philemon: oh, yeah.

client/webserver/site/src/html/bodybuilder.tmpl
ukane-philemon: I think this must have been during my last rebase.

client/core/types.go
ukane-philemon: Hmm, I think we can leave it empty. Instead of having to loop through every order ID string character.

ukane-philemon:
Screenshot 2022-11-16 at 1 32 20 PM

client/webserver/api.go
```diff
@@ -359,7 +359,8 @@ func (s *WebServer) apiConnectWallet(w http.ResponseWriter, r *http.Request) {
 	writeJSON(w, simpleAck(), s.indent)
 }
 
-// apiTrade is the handler for the '/trade' API request.
+// apiTrade is the handler for the '/trade' API request. This end-point is
+// Deprecated. Use the 'tradeasync' end-point.
```
chappjc, line 363: This endpoint is NOT deprecated.

client/core/locale_ntfn.go
```diff
@@ -87,6 +87,11 @@ var originLocale = map[Topic]*translation{
 		subject:  "Send Successful",
 		template: "Sending %s %s to %s has completed successfully. Coin ID = %s",
 	},
+	//[order ID, error]
+	TopicAsyncOrderFailure: {
+		subject:  "Inflight Order Error",
+		template: "Inflight order with ID %v failed: %v",
```
chappjc, line 93: For user-facing messages, it should be "in-flight".  I don't believe "inflight" is proper English, at best an idiom.

https://dictionary.cambridge.org/dictionary/english/in-flight

chappjc: Although now that I say that, I suspect "in-flight" will prove to be too colloquial to translate directly while keeping the intended meaning.  We can keep "in-flight", but I bet for other languages we'll want translators to give the equivalent of "Submitting Order Error".  Just a thought.

This PR is g2g with the two comments resolved.

buck54321 (approved):
Nice! Looks great.

client/webserver/site/src/js/app.ts
```diff
@@ -771,14 +793,15 @@ export default class Application {
...
-      o = []
-      this.user.exchanges[host].markets[mktID].orders = o
-    }
-    return o
+    let orders: Order[] = []
+    const mkt = this.user.exchanges[host].markets[mktID]
+    if (mkt.orders) orders = orders.concat(mkt.orders)
+    if (mkt.inflight) orders = orders.concat(mkt.inflight)
+    return orders
   }
```
ukane-philemon, line 806: @buck54321, I think initializing `mkt.orders` or `mkt.inflight` with an empty array is not useful.

ukane-philemon:
> Nice! Looks great.

@buck54321, @chappjc, thanks for the reviews.

chappjc (approved):
Re-tested and still working.

client/core/locale_ntfn.go
chappjc: You really didn't need to change anything in the code, just the *user-facing* text, but it's fine.

I'm unresolving this conversation so that it has visibility when we decide to look back during translation.

ukane-philemon: Alright.

PR #1888, opened 2022-10-05
client/doge: Add external fee estimator for doge

vctt94:
This PR is a work towards #1618.

On this PR I added new params to the FeeEstimator btc function: `func(context.Context, RawRequester, uint64, bool, dex.Network) (uint64, error)` and added a external fee estimator for doge clients.

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -341,11 +341,12 @@ type BTCCloneCFG struct {
 	// NonSegwitSigner can be true if the transaction signature hash data is not
 	// the standard for non-segwit Bitcoin. If nil, txscript.
 	NonSegwitSigner TxInSigner
-	// FeeEstimator provides a way to get fees given a context, a RawRequest-enabled
-	// client, a confirmation target, a bool checking if the client is allowed
-	// or not to make external requests and the network.
-	FeeEstimator func(ctx context.Context, rawRequester RawRequester, confTarget uint64,
-		externalAllowed bool, net dex.Network) (fee uint64, err error)
+	// FeeEstimator provides a way to get fees given an RawRequest-enabled
+	// client and a confirmation target.
+	FeeEstimator func(RawRequester, uint64) (uint64, error)
+	// ExternalFeeEstimator should be supplied if the clone provides the
+	// apifeefallback ConfigOpt.
+	ExternalFeeEstimator func(context.Context, dex.Network) (uint64, error)
 	// OmitAddressType causes the address type (bech32, legacy) to be omitted
 	// from calls to getnewaddress.
 	OmitAddressType bool
@@ -789,7 +790,8 @@ type baseWallet struct {
 	zecStyleBalance   bool
 	segwit            bool
 	signNonSegwit     TxInSigner
-	estimateFee       func(context.Context, RawRequester, uint64, bool, dex.Network) (uint64, error) // TODO: resolve the awkwardness of an RPC-oriented func in a generic framework
+	estimateFee       func(RawRequester, uint64) (uint64, error) // TODO: resolve the awkwardness of an RPC-oriented func in a generic framework
+	fetchExternalFee  func(context.Context, dex.Network) (uint64, error)
 	decodeAddr        dexbtc.AddressDecoder
 	deserializeTx     func([]byte) (*wire.MsgTx, error)
 	serializeTx       func(*wire.MsgTx) ([]byte, error)
@@ -1006,6 +1008,8 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, net dex.Network) (ass
 		DefaultFallbackFee:  defaultFee,
 		DefaultFeeRateLimit: defaultFeeRateLimit,
 		Segwit:              true,
+		// Add our fee rate fetcher.
+		ExternalFeeEstimator: externalFeeEstimator,
 	}
 
 	switch cfg.Type {
@@ -1174,6 +1178,7 @@ func newUnconnectedWallet(cfg *BTCCloneCFG, walletCfg *WalletConfig) (*baseWalle
 		initTxSizeBase:      uint64(initTxSizeBase),
 		signNonSegwit:       nonSegwitSigner,
 		estimateFee:         cfg.FeeEstimator,
+		fetchExternalFee:    cfg.ExternalFeeEstimator,
 		decodeAddr:          addrDecoder,
 		stringAddr:          addrStringer,
 		walletInfo:          cfg.WalletInfo,
@@ -1474,8 +1479,12 @@ func (btc *baseWallet) feeRate(ctx context.Context, _ RawRequester, confTarget u
 			btc.log.Warnf("Failed to get local fee rate estimate: %v", err)
 			return 0, err
 		}
+		if btc.fetchExternalFee == nil {
+			btc.log.Debug("API fallback allowed but no fetcher provided for %s", btc.symbol)
+			return 0, err
+		}
 		btc.log.Debug("Retrieving fee rate from external API: ", externalApiUrl)
-		estimatedFee, err := externalFeeEstimator(ctx, net)
+		estimatedFee, err := btc.fetchExternalFee(btc.ctx, btc.Network)
 		if err != nil {
 			btc.log.Errorf("Failed to get fee rate from external API: %v", err)
 			return 0, err
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -343,7 +343,7 @@ type BTCCloneCFG struct {
 	NonSegwitSigner TxInSigner
 	// FeeEstimator provides a way to get fees given an RawRequest-enabled
 	// client and a confirmation target.
-	FeeEstimator func(RawRequester, uint64) (uint64, error)
+	FeeEstimator func(context.Context, RawRequester, uint64, bool, dex.Network) (uint64, error)
 	// OmitAddressType causes the address type (bech32, legacy) to be omitted
 	// from calls to getnewaddress.
 	OmitAddressType bool
@@ -787,7 +787,7 @@ type baseWallet struct {
 	zecStyleBalance   bool
 	segwit            bool
 	signNonSegwit     TxInSigner
-	estimateFee       func(RawRequester, uint64) (uint64, error) // TODO: resolve the awkwardness of an RPC-oriented func in a generic framework
+	estimateFee       func(context.Context, RawRequester, uint64, bool, dex.Network) (uint64, error) // TODO: resolve the awkwardness of an RPC-oriented func in a generic framework
 	decodeAddr        dexbtc.AddressDecoder
 	deserializeTx     func([]byte) (*wire.MsgTx, error)
 	serializeTx       func(*wire.MsgTx) ([]byte, error)
@@ -921,7 +921,7 @@ func (btc *ExchangeWalletSPV) Rescan(_ context.Context) error {
 
 // FeeRate satisfies asset.FeeRater.
 func (btc *ExchangeWalletFullNode) FeeRate() uint64 {
-	rate, err := btc.estimateFee(btc.node, 1)
+	rate, err := btc.estimateFee(btc.ctx, btc.node, 1, btc.apiFeeFallback(), btc.Network)
 	if err != nil {
 		btc.log.Tracef("Failed to get fee rate: %v", err)
 		return 0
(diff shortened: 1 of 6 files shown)


client/asset/doge/doge.go
```diff
@@ -160,32 +192,102 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network)
...
+		return nil, err
+	}
+	return &DogeWallet{
+		*cloneWallet,
+		doge,
+	}, nil
+}
+
+func (w *DogeWallet) Connect(ctx context.Context) (*sync.WaitGroup, error) {
+	w.ctx = ctx
```
vctt94, line 212: should this be done?

client/asset/doge/doge.go
```diff
@@ -160,32 +192,102 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network)
...
+	err = json.Unmarshal(resp, &feeRate)
+	if err != nil {
+		return 0, err
+	}
+	if feeRate <= 0 {
+		return 0, nil
+	}
+	// estimatefee is f#$%ed
+	// https://github.com/decred/dcrdex/pull/1558#discussion_r850061882
+	if feeRate > dexdoge.DefaultFeeRateLimit/1e5 {
```
vctt94, line 247: should we try a external fee here?

client/asset/doge/doge.go
```diff
@@ -160,32 +192,102 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network)
...
 		UnlockSpends:             true,
 		ConstantDustLimit:        dustLimit,
-		FeeEstimator: func(cl btc.RawRequester, _ uint64) (uint64, error) {
-			confArg, err := json.Marshal(feeConfs)
-			if err != nil {
-				return 0, err
-			}
-			resp, err := cl.RawRequest("estimatefee", []json.RawMessage{confArg})
-			if err != nil {
-				return 0, err
```
chappjc, line 170: Can't this FeeEstimator closure stay the same and just call the external API on errors like these?  I'm not seeing the need for `type DogeWallet struct` or why it's composed the way it is.

vctt94: The reason I am not calling it at the FeeEstimator is because I do not have access to the context. 

So I created the DogeWallet struct, so I can call the `Connect` method and initialize the context.

client/asset/btc/btc.go
```diff
@@ -335,7 +335,7 @@ type BTCCloneCFG struct {
 	NonSegwitSigner TxInSigner
 	// FeeEstimator provides a way to get fees given an RawRequest-enabled
 	// client and a confirmation target.
-	FeeEstimator func(RawRequester, uint64) (uint64, error)
+	FeeEstimator func(context.Context, RawRequester, uint64, bool, dex.Network) (uint64, error)
```
chappjc, line 338: Docs for `FeeEstimator` could use an update now.

client/asset/btc/btc.go
```diff
@@ -1212,7 +1212,7 @@ func (btc *baseWallet) connect(ctx context.Context) (*sync.WaitGroup, error) {
 		return nil, fmt.Errorf("invalid best block hash from %s node: %v", btc.symbol, err)
 	}
 	// Check for method unknown error for feeRate method.
-	_, err = btc.estimateFee(btc.node, 1)
+	_, err = btc.estimateFee(btc.ctx, btc.node, 1, btc.apiFeeFallback(), btc.Network)
```
chappjc, line 1215: Same thing, but let's use the `ctx` var directly here.

client/asset/doge/doge.go
```diff
@@ -160,22 +180,36 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network)
...
 			if feeRate <= 0 {
-				return 0, nil
+				if !allowExternal {
+					return 0, nil
+				}
+				feeRate, err := externalFeeEstimator(ctx, net)
+				if err != nil {
+					return 0, err
+				}
+				return feeRate, nil
```
chappjc, line 212: We can do `return externalFeeEstimator(ctx, net)` for these two places in this closure.  We define the function below, so we know it's going to return 0 for a non-nil error.

chappjc (changes requested):
Feels much better now.  I think this opens some doors to transitioning to `RawRequesterWithContext` if we decide to keep going with this rather than switching to the bool approach of modifying RPC behavior, and given the external functions are different I think we'll probably keep it this way for a while.

JoeGruffins (commented):
Unrelated to this pr but:

I went to test this on testnet and none of my txn were being mined. Now dogecoind thinks I have a lot of utxo that never happened apparently, and there is no rescan so funds are gone.

So I went to simnet and our harness says:
```
+ echo 'Generating 400 blocks for alpha'
Generating 400 blocks for alpha
+ tmux send-keys -t doge-harness:4 './alpha generatetoaddress 400 mjzB71SzfAi8BwNvobPZ983d8vCdxuQD2i; tmux wait-for -S doge' 'C-m;' wait-for doge
++ ./delta getnewaddress
error code: -28
error message:
Loading wallet...
+ DELTA_ADDR=
```

I'm pretty sure this was working before. Not sure what is going on. Doge version is `v1.14.5.0-31afd1331`

Everyone else's doge fine?

JoeGruffins (approved):
Working well on simnet.

Is testnet ok for others on master?

chappjc:
> ```
> + echo 'Generating 400 blocks for alpha'
> Generating 400 blocks for alpha
> + tmux send-keys -t doge-harness:4 './alpha generatetoaddress 400 mjzB71SzfAi8BwNvobPZ983d8vCdxuQD2i; tmux wait-for -S doge' 'C-m;' wait-for doge
> ++ ./delta getnewaddress
> error code: -28
> error message:
> Loading wallet...
> + DELTA_ADDR=
> ```
> 
> I'm pretty sure this was working before. Not sure what is going on. Doge version is `v1.14.5.0-31afd1331`
> 
> Everyone else's doge fine?
> 
> I was using the harness with the last tests pr so it was working then.

ok on simnet for me

```
+ echo 'Generating 400 blocks for alpha'
Generating 400 blocks for alpha
+ tmux send-keys -t doge-harness:4 './alpha generatetoaddress 400 mjzB71SzfAi8BwNvobPZ983d8vCdxuQD2i; tmux wait-for -S doge' 'C-m;' wait-for doge
++ ./delta getnewaddress
+ DELTA_ADDR=mz2AAWNoxgHzYZnMRxZ5nL6XDejd9xYiaZ
+ echo 'delta address = '\''mz2AAWNoxgHzYZnMRxZ5nL6XDejd9xYiaZ'\'''
delta address = 'mz2AAWNoxgHzYZnMRxZ5nL6XDejd9xYiaZ'
++ ./gamma getnewaddress
+ GAMMA_ADDR=miKNQzWeCTvtWJWfNYtSpVUY7EtZTrxnbg
+ echo 'gamma address = '\''miKNQzWeCTvtWJWfNYtSpVUY7EtZTrxnbg'\'''
gamma address = 'miKNQzWeCTvtWJWfNYtSpVUY7EtZTrxnbg'
+ echo 'Sending 8,400,000 DOGE to lazy wallets in 8 blocks'
Sending 8,400,000 DOGE to lazy wallets in 8 blocks
...
```

```
$  dogecoind --version
Dogecoin Core Daemon version v1.14.6
```

> I went to test this on testnet and none of my txn were being mined. Now dogecoind thinks I have a lot of utxo that never happened apparently, and there is no rescan so funds are gone.

testnet seems alright too.

v0.14.6 does have rescan RPC.  Maybe you need to update.

client/asset/doge/doge.go
```diff
@@ -189,3 +215,44 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network)
...
+	}
+	var resp map[string]float64
+	reader := io.LimitReader(httpResponse.Body, 1<<20)
+	err = json.NewDecoder(reader).Decode(&resp)
+	if err != nil {
+		return 0, err
+	}
+	httpResponse.Body.Close()
+
+	feeBtcByte, ok := resp["feerate"]
```
chappjc, line 247: Right now the testnet URL is returning `{"feerate":-1,"blocks":25}`.

Need the sanity check, i.e. run it through the `dexdoge.DefaultFeeRateLimit` check that we do back in `FeeEstimator`

client/asset/doge/doge.go
```diff
@@ -189,3 +215,44 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network)
...
+	feeInSat := toSatoshi(feeBtcByte / 1000)
```
chappjc, line 248: Are you sure the units are `feeBtcByte`?  I think the bitcore result uses exactly the same as the units from `estimatesmartfee`.  DOGE/kB?

vctt94: You are correct, I didn't pay attention to the `DefaultFeeRateLimit` comment and thought 6 doge/kb was too high to be right. Fixed it

JoeGruffins:
> v0.14.6 does have rescan RPC. Maybe you need to update.

Will update.

JoeGruffins:
I have not looked into it too much, but I think something is wrong with the fees and that's why my testnet txn are not being mined, maybe not related to this pr. In the logs I will see:

```
2022-10-29 04:07:15.073 [DBG] CORE[doge]: 3 signature cycles to converge on fees for tx 83e3eb0c60ffff1057c3ac23373bedf6243ffe5079a4dc47021c2d1a8373a69a: min rate = 10, actual fee rate = 10 (2230 for 223 bytes), change = true
```
fee rate is 10sat/byte, that can't be right. Should be at least 400 for doge no? https://github.com/decred/dcrdex/blob/5550e8443a01309244fa12d9f9f2333bf4620503/dex/networks/doge/params.go#L22

JoeGruffins:
I see what it is now. I have 10 set as the max fee rate for doge in my `markets.conf`:
```
        "DOGE_testnet": {
            "bip44symbol": "doge",
            "network": "testnet",
            "maxFeeRate": 10,
            "swapConf": 2,
            "configPath": "/home/joe/finalbig/doge/dogecoin.conf"
        }
```
Again, unrelated to this pr, but it seems server could make clients fail by setting this too low. Will make an issue since I am also worried about the `swapConf` value.

vctt94:
Thanks for the review Joe

buck54321:
What do you think of this alternative approach. 

https://github.com/buck54321/dcrdex/compare/ccd5edf1ad9615c97f4af2912e8096277d90edc3...buck54321:dcrdex:doge-xcrate-alt

It leaves the existing API alone and adds functionality via the `BTCCloneCFG` instead, which has become a somewhat standard way of doing things.

chappjc:
> It leaves the existing API alone and adds functionality via the `BTCCloneCFG` instead, which has become a somewhat standard way of doing things.

That is a lot cleaner.  Don't know why I didn't see that option.

vctt94:
I think that's a good approach, should I cherry-pick and push it here, or will you open another PR?

buck54321:
One last thing, if you would.
https://github.com/buck54321/dcrdex/compare/d25a4e54f452bd48bb8f36965cd5fb552ca096f9...buck54321:dcrdex:add-fee-rate-test-for-doge

chappjc:
Pushing a patch shortly.  Please hold the branch where it is.

chappjc:
> What do you think of this alternative approach.
> 
> [buck54321/dcrdex@ccd5edf...buck54321:dcrdex:doge-xcrate-alt](https://github.com/buck54321/dcrdex/compare/ccd5edf1ad9615c97f4af2912e8096277d90edc3...buck54321:dcrdex:doge-xcrate-alt)
> 
> It leaves the existing API alone and adds functionality via the `BTCCloneCFG` instead, which has become a somewhat standard way of doing things.

I'm afraid this didn't work because `(*baseWallet).feeRate` is only used with BTC.  The clones override it, which is why it https://github.com/decred/dcrdex/pull/1888/commits/76c2937aab8056dfe347129a8cfdf35b04d69df1 worked.

I'm looking at a reworked approach, but I think we overlooked a complication.

chappjc:
Thanks for the update @buck54321 but I meant to close this PR.
I kept with your approach in https://github.com/decred/dcrdex/pull/1967, but addressed a number of other issues.  The resolution there is in line with what you just proposed, I believe.

buck54321:
> The resolution there is in line with what you just proposed, I believe.

Yes. Seems equivalent.

PR #1889, opened 2022-10-05
client/core: blind cancel of unknown orders reported on connect

chappjc:
<s>Requires https://github.com/decred/dcrdex/pull/1866, this PR is the last two commits:</s>

**client/core: blind cancel of unknown orders reported on connect**

```
There are a variety of pathological circumstances in which the
client can be unaware of orders that the server claims are active
(in epoch or booked status on the server). These include using
an old or corrupt DB file that lacks current orders, errors saving
submitted orders, and such trades having been self-revoked such
as if the server were down past the taker's locktime. Each of these
scenarios has been encountered in the wild.

This change introduces a simple "blind cancel" mechanism for submitting
cancel orders for any of the orders reported by the server on connect
that are not presently "active" according to the client. In submitting
these cancels, no accompanying tracked trade is present, only an entry
in the new blindCancels map. This map facilitates handling of the
preimage request for the order (if is accepted), as well as silencing
errors that would occur in the match/nomatch handlers.

Future work should introduce a method for manually self-revoking orders
while the DEX connection is down. Presently, this code will only be
used for the anomalous cases described above.
```

**client/core: purge legacy preimage request handling** 

```
The initial preimage request did not include the order commitment for
the preimage being requested.  This had us doing some tricks because
the request only provided the order ID, which is not known to the client
until the order submission response was received, creating a race.

Since this flawed design was resolved in 850e8a6, in v0.2.0 about 18
months ago, by including the commitment in the preimage request,
we can safely remove the legacy client-side handling for requests
that do not include the commitment.  This change removes the piSyncers
map and legacy code path from handlePreimageRequest.
```

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -145,6 +145,9 @@ type dexConnection struct {
 	// processed by a dex server.
 	inFlightOrders map[uint64]*InFlightOrder
 
+	blindCancelsMtx sync.Mutex
+	blindCancels    map[order.OrderID]order.Preimage
+
 	epochMtx sync.RWMutex
 	epoch    map[string]uint64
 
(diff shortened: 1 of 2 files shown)


client/core/core.go
```diff
@@ -430,6 +433,55 @@ func (dc *dexConnection) findOrder(oid order.OrderID) (tracker *trackedTrade, pr
 	return
 }
 
+func (c *Core) sendCancelOrder(dc *dexConnection, oid order.OrderID, base, quote uint32) (order.Preimage, *order.CancelOrder, []byte, chan struct{}, error) {
+	preImg := newPreimage()
+	co := &order.CancelOrder{
+		P: order.Prefix{
+			AccountID:  dc.acct.ID(),
+			BaseAsset:  base,
+			QuoteAsset: quote,
```
chappjc, line 442: The requirement to have base/quote asset IDs for a cancel order is kind of annoying since just `TargetOrderID` is enough for server to figure it out.  I have a stash that modifies the server's `handleCancel` to loop through the market tunnels, but that doesn't change deployed servers...

chappjc:
Alright for review I suppose.  Just the last two commits.  Haven't retested recently and there have been numerous rebases since, but it's what I intend in concept.

JoeGruffins (approved):

client/core/core_test.go
```diff
@@ -7558,7 +7539,8 @@ func TestPreimageSync(t *testing.T) {
 		lo := convertMsgLimitOrder(msgOrder)
 		resp := orderResponse(msg.ID, msgOrder, lo, false, false, false)
 		limitRouteProcessing <- lo.ID()
-		<-time.NewTimer(time.Millisecond * 100).C
+		commitSig <- lo.Commit
+		time.Sleep(200 * time.Millisecond)
```
buck54321, line 7543: I commented out this line and ran 
`parallel -P 4 -N0 --halt soon,fail=1 go test -race -count=1 -short -run TestPreimageSync :::: <(seq 1 50)`
and had no fails.

client/core/core_test.go
```diff
@@ -7558,7 +7539,8 @@ func TestPreimageSync(t *testing.T) {
 		lo := convertMsgLimitOrder(msgOrder)
 		resp := orderResponse(msg.ID, msgOrder, lo, false, false, false)
 		limitRouteProcessing <- lo.ID()
-		<-time.NewTimer(time.Millisecond * 100).C
+		commitSig <- lo.Commit
```
buck54321, line 7542: I think because of the sequencing provided by the `errChan` that you can skip the channel and just  directly set a 
```Go
var commit order.Commitment
```

buck54321 (approved):

PR #1892, opened 2022-10-06
client: add external gas estimation for eth as fallback

vctt94:
This is a work towards #1618

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -11,10 +11,13 @@ import (
 	"crypto/sha256"
 	"encoding/binary"
 	"encoding/hex"
+	"encoding/json"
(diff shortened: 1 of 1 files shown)


JoeGruffins:
https://github.com/decred/dcrdex/issues/1618#issuecomment-1248131384

This is not needed for eth because the fee is not negotiated. It is part of their consensus, now. We almost always just use the max we are willing to pay.

The places you have changed in the code are not dealing with the price of gas, but the amount of gas used for the tx, which an external api can't tell us, probably, but it's always about the same for the same type of tx (swap, redeem, refund, normal send).

Eth is different and a little confusing at first.

vctt94:
Thanks for point it out Joe

PR #1893, opened 2022-10-06
asset/eth: create node client on Connect

### LINKED ISSUE #1883: client: failed to enable wallet due to datadir error
While investigating #1881, I disabled my ETH wallet. Then I started recovery for my btc spv wallet. While still recovering the wallet, I was unable to enable my ETH wallet because it kept failing with `datadir already used by another process` error(I guess because of the recovery?).  But now am done with the recovery, btc spv wallet is 100% synced still the `datadir already used by another process` error persists.

EDIT 1: After some investigation, the issue was never caused or related to the btc spv recovery.

EDIT 2: To reproduce: disable ETH wallet -> shutdown and start dexc -> try to enable ETH wallet.
What happens is that ETH opens a db and places a lock on the eth `datadir`(even if the wallet is disabled) then trying to load the wallet after it has been enabled will return an error: 

```
2022-10-07 10:50:00.274 [INF] CORE: Starting DEX client core
2022-10-07 10:50:00.275 [INF] CORE: Loaded btc wallet configuration.
2022-10-07 10:50:00.276 [INF] CORE: Loaded dcr wallet configuration.
2022-10-07 10:50:00.338 [INF] CORE[eth][TXDB]: All 0 tables opened in 0s // wallet is disabled.

2022-10-07 10:50:00.355 [INF] CORE[eth][TXDB]: Replaying file id: 0 at offset: 0

2022-10-07 10:50:00.356 [INF] CORE[eth][TXDB]: Replay took: 14.343µs

2022-10-07 10:50:00.356 [DBG] CORE[eth][TXDB]: Value log discard stats empty
2022-10-07 10:50:00.356 [INF] CORE: Loaded eth wallet configuration.
```

Screenshot 2022-10-03 at 8 56 22 AM

ukane-philemon:
`ETH` `NewWallet` initializes a node client and database and when initializing the node client, a lock is placed on the wallet's `datadir`. As a result, trying to enable the wallet will return `datadir already used by another process` because `loadWallet` will eventually call `NewWallet` which will try to initialize the node client and place a lock on the already locked `datadir`. Closes #1883

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -479,6 +479,11 @@ func CreateWallet(createWalletParams *asset.CreateWalletParams) error {
 // NewWallet is the exported constructor by which the DEX will import the
 // exchange wallet.
 func NewWallet(assetCFG *asset.WalletConfig, logger dex.Logger, net dex.Network) (*ETHWallet, error) {
+	cl, err := newNodeClient(getWalletDir(assetCFG.DataDir, net), net, logger.SubLogger("NODE"))
+	if err != nil {
+		return nil, err
+	}
+
 	cfg, err := parseWalletConfig(assetCFG.Settings)
 	if err != nil {
 		return nil, err
@@ -492,6 +497,8 @@ func NewWallet(assetCFG *asset.WalletConfig, logger dex.Logger, net dex.Network)
 	eth := &baseWallet{
 		log:          logger,
 		net:          net,
+		node:         cl,
+		addr:         cl.address(),
 		dir:          assetCFG.DataDir,
 		gasFeeLimitV: gasFeeLimit,
 		wallets:      make(map[uint32]*assetWallet),
@@ -556,22 +563,8 @@ func loadMonitoredTxs(db kvdb.KeyValueDB) (map[common.Hash]*monitoredTx, error) 
 
 // Connect connects to the node RPC server. Satisfies dex.Connector.
 func (w *ETHWallet) Connect(ctx context.Context) (*sync.WaitGroup, error) {
-	cl, err := newNodeClient(getWalletDir(w.dir, w.net), w.net, w.log.SubLogger("NODE"))
-	if err != nil {
-		return nil, err
-	}
-
-	db, err := kvdb.NewFileDB(filepath.Join(w.dir, "tx.db"), w.log.SubLogger("TXDB"))
-	if err != nil {
-		return nil, err
-	}
-
-	w.node = cl
-	w.addr = cl.address()
-	w.monitoredTxDB = db
 	w.ctx = ctx
-
-	err = w.node.connect(ctx)
+	err := w.node.connect(ctx)
 	if err != nil {
 		return nil, err
 	}
@@ -584,6 +577,12 @@ func (w *ETHWallet) Connect(ctx context.Context) (*sync.WaitGroup, error) {
 		w.contractors[ver] = c
 	}
 
+	db, err := kvdb.NewFileDB(filepath.Join(w.dir, "tx.db"), w.log.SubLogger("TXDB"))
+	if err != nil {
+		return nil, err
+	}
+
+	w.monitoredTxDB = db
 	w.monitoredTxs, err = loadMonitoredTxs(w.monitoredTxDB)
 	if err != nil {
 		return nil, err
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -342,6 +342,7 @@ type baseWallet struct {
 	node ethFetcher
 	addr common.Address
 	log  dex.Logger
+	dir  string
 
 	gasFeeLimitV uint64 // atomic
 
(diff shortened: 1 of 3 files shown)


JoeGruffins:
Are you able to reproduce the original error where you can't enable? I can't make it happen on master. It looks like eth's shutdown will release any locks on the data dir and db. If it doesn't then I guess the problem is there. Are you sure your wallet just wasn't able to shutdown properly there?

I could be wrong, but my assumption is that the asset driver's Open only ever happens once, and then Connect also only happens once and that wallet is dead after connect's ctx is canceled and needs another Open and Connect. The difference being Connect stuff needs a context.Context. If this assumption is correct then moving between the two should not matter.

ukane-philemon:
> Are you able to reproduce the original error where you can't enable? I can't make it happen on master.

I've updated my comment here : https://github.com/decred/dcrdex/issues/1883#issue-1394313936

chappjc:
Huh, I would think the same issue of loading an on-disk db twice exists for all the native wallets.

That suggests to me the problem is with `(*Core).ToggleWalletStatus`.

In the `!disable` branch there, why do we do this:

```go
		dbWallet, err := c.db.Wallet(wallet.dbID)
		if err != nil {
			return fmt.Errorf("error retrieving DB wallet: %w", err)
		}

		wallet, err = c.loadWallet(dbWallet)
		if err != nil {
			return newError(walletErr, "error loading wallet for %d -> %s: %w",
				assetID, unbip(assetID), err)
		}

		wallet.setDisabled(false)
```

Instead of just `wallet.setDisabled(false)`?  We've loaded the xcWallet from the map at the start of the `ToggleWalletStatus` so don't we just have to flag it enabled and connect?

chappjc:
Even if what I said about `ToggleWalletStatus` is true, I think you're right that the `NewWallet` constructor shouldn't be doing these disk activities.

I'm looking at the DCR and BTC constructors and they only ready the asset.Wallet, but they don't mess with the datadirs.

Will test both ETH and the native utxo asset wallets

chappjc (commented):
Concept ACK.  I believe it is correct to perform these on-disk activities in connect rather than the constructor.  It is more in line with the pattern used in the DCR and BTC wallets.

However, please also consider what I noted about not calling `loadWallet` in `ToggleWalletStatus`.  I don't see a reason to re-create an `xcWallet` when enabling.  I don't think connect->disconnect->connect cycles are prohibited, and we do that in the ReconfigureWallet method, but will give consideration if it is indeed safest to make a new instance.  EDIT: yes, I'm pretty sure the wallets are ok with connect/disconnect cycles, and we went to some length to make sure the dcr websocket RPC client worked with this in https://github.com/decred/dcrdex/commit/abc07a3c8cabe0f064e938abab32e247efd8551b for v0.5.0

Also, I suspect we'll hit a random panic somewhere because `baseWallet.node` is nil before connect, but I don't recall precisely which methods are accessed before connect.  I'm thinking about the SyncStatus, NewAddress, OwnsDepositAddress, and DepositAddress methods.  Probably OK, but I recall some pitfall.

ukane-philemon:
Re: using `loadWallet` in `ToggleWalletStatus` 

I think the issue then was from `ETH`, I might not have understood why I had issues with ETH and was loading all disabled wallet to be sure everything will be in shape for the newly enabled wallet but now It's clear, disconnecting an ETH wallet will close the `monitorTxdb` which `Connect` won't open causing unexpected behaviour. There could more issue related to this. Will investigate properly and improve `ToggleWalletStatus`.

chappjc (approved):
Tests well with ETH, plus both DCR and BTC each in SPV and RPC modes.

Tested with the suggested `ToggleWalletStatus` change that removes `db.Wallet` and `c.loadWallet` in favor of re-using the existing `xcWallet`,

chappjc:
Oh, I also squeezed a fix for disabling an unconnected wallet into #1870 at https://github.com/decred/dcrdex/pull/1870/commits/9525880298c8b4cb418ff6bb33821975cf8b1c9e

Plus a check in `connectWallet` to block connecting a disabled wallet (in case a method is called without referencing WalletState.Disabled). https://github.com/decred/dcrdex/pull/1870/commits/b06003b0f6f87656855c33e7849c061da78018a2

Have a look at #1870 when you get a chance please.

chappjc:
> However, `loadWallet` should be able to re-create new instance of any wallet without gotchas, right? Assuming all other instance are not connected.

_Should_ be fine, yeah.

ukane-philemon:
> Have a look at https://github.com/decred/dcrdex/pull/1870 when you get a chance please.

Okay, I will merge the changes there.

ukane-philemon:
> Should be fine, yeah

So what do we think? We could assign `baseWallet.node` at connect and fix instances where `baseWallet.node` could be used before Connect or allow NewWallet assign a node and deal with ETH placing a lock for an unconected node.

chappjc:
> > Should be fine, yeah
> 
> So what do we think? We could assign `baseWallet.node` at connect and fix instances where `baseWallet.node` could be used before Connect or allow NewWallet assign a node and deal with ETH placing a lock for an unconected node.

Commented in matrix.  I think your original PR made sense, yes.  I just meant that I _also_ tested these loadWallet changes and they worked too.

buck54321 (approved):

JoeGruffins (approved):
Am able to reproduce not being able to enable the wallet if I start disabled on mater. This pr allows me to enable it.

nit: Commit message should be wrapped at 72 chars for f27c12f423c96dbba8f00bd56c33c4c05c2073ae

PR #1894, opened 2022-10-09
core/eth: Handle block gas limit

### LINKED ISSUE #1771: eth: handle block size max
> I does seem limited, but considering that a regular tx is 21,000 gas, I think it's correct.

Back to the gas estimate issue, I think this means we should place a cap on the number of inits.

More broadly, it sounds like we need to place a cap on the number of lots otherwise there will be potentially unswappable orders, however huge they'd be.

_Originally posted by @chappjc in https://github.com/decred/dcrdex/issues/1622#issuecomment-1209423904_

___

Just spitballin'. Either...
1) Place a cap on number of lots in the dex protocol or at the server configuration level
2) Break up large swap internally to eth, which will carry a lot of complexity especially around errors.
3) Handle it in `*Core`, which may end up being the easiest, implementation-wise, but it's really just an ethereum thing.

martonp:
Two new fields are added to `asset.WalletInfo`: `MaxSwapsInTx` and `MaxRedeemsInTx`. `Core` uses these fields to call `Swap` and `Redeem` in batches if needed. These fields are currently only populated by ETH and ETH token wallets. They check each of the versions of the contract, and the fields are populated so that at most 1/4 of the gas limit of a block can be filled up by a swap or redeem transaction.

Closes #1771

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -39,6 +39,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	ethmath "github.com/ethereum/go-ethereum/common/math"
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/eth/ethconfig"
 	"github.com/ethereum/go-ethereum/params"
 	"github.com/tyler-smith/go-bip39"
 )
@@ -384,6 +385,9 @@ type assetWallet struct {
 
 	evmify  func(uint64) *big.Int
 	atomize func(*big.Int) uint64
+
+	maxSwapsInTx   uint64
+	maxRedeemsInTx uint64
 }
 
 // ETHWallet implements some Ethereum-specific methods.
@@ -408,16 +412,22 @@ type TokenWallet struct {
 	token    *dexeth.Token
 }
 
+const maxProportionOfBlockGasLimitToUse = 4
+
 // Info returns basic information about the wallet and asset.
-func (*ETHWallet) Info() *asset.WalletInfo {
+func (w *ETHWallet) Info() *asset.WalletInfo {
+	WalletInfo.MaxSwapsInTx = w.maxSwapsInTx
+	WalletInfo.MaxRedeemsInTx = w.maxRedeemsInTx
 	return WalletInfo
 }
 
 // Info returns basic information about the wallet and asset.
 func (w *TokenWallet) Info() *asset.WalletInfo {
 	return &asset.WalletInfo{
-		Name:     w.token.Name,
-		UnitInfo: w.token.UnitInfo,
+		Name:           w.token.Name,
+		UnitInfo:       w.token.UnitInfo,
+		MaxSwapsInTx:   w.maxSwapsInTx,
+		MaxRedeemsInTx: w.maxRedeemsInTx,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


client/asset/eth/eth.go
```diff
@@ -408,17 +409,62 @@ type TokenWallet struct {
...
+			maxSwapGas = gases.Swap
+		}
+		if gases.Redeem > maxRedeemGas {
+			maxRedeemGas = gases.Redeem
+		}
+	}
+
+	WalletInfo.MaxSwapsInTx = int(gasCeil / maxProportionOfBlockGasLimitToUse / maxSwapGas)
+	WalletInfo.MaxRedeemsInTx = int(gasCeil / maxProportionOfBlockGasLimitToUse / maxRedeemGas)
+
```
JoeGruffins, line 430: It looks like all of these numbers are constant and we don't need to calculate it every time. Could also use the swapAdd or redeemAdd to be more precise.

I guess the max number would be something like 30000000/4/135000 or 55 swaps for eth?

JoeGruffins (approved):
Is there an issue for this? I see that it solves the problem of a tx being too large. Looks good to me.

client/asset/eth/eth.go
martonp: I wanted to do a calculation so that this doesn't need to be updated if a new version of the ETH contract is added. I guess I could do the calculation in `init` to avoid doing it on each call to `Info`.

> I guess the max number would be something like 30000000/4/135000 or 55 swaps for eth?

Yeah, you can see the exact values in `eth_test.go`.

client/core/trade.go
```diff
@@ -2090,7 +2090,18 @@ func (c *Core) swapMatches(t *trackedTrade, matches []*matchTracker) error {
 		}
 	}
 	if len(groupables) > 0 {
-		c.swapMatchGroup(t, groupables, errs)
+		maxSwapsInTx := t.wallets.fromWallet.Info().MaxSwapsInTx
+		if maxSwapsInTx <= 0 || len(groupables) < maxSwapsInTx {
```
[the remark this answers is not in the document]
martonp: No, but it feels safer to have that when it's a signed int.

client/asset/eth/eth.go
martonp: I've updated it to be calculated just once when the wallet is created. I wanted to do a calculation instead of hardcoding the value to avoid having to update this when new versions of the contract are created.

> I guess the max number would be something like 30000000/4/135000 or 55 swaps for eth?
Yes its 55 swaps and 119 redeems for eth.

client/asset/eth/eth.go
```diff
@@ -408,16 +412,22 @@ type TokenWallet struct {
 	token    *dexeth.Token
 }
 
+const maxProportionOfBlockGasLimitToUse = 4
```
chappjc, line 415: ACK that 25% of a block's gas limit is a reasonable default.  Just please comment here to explain.

chappjc (changes requested):
Nice work, thanks.

client/asset/interface.go
```diff
@@ -207,6 +207,12 @@ type WalletInfo struct {
 	// UnitInfo is the information about unit names and conversion factors for
 	// the asset.
 	UnitInfo dex.UnitInfo `json:"unitinfo"`
+	// MaxSwapsInTx is the max amount of swaps that this wallet can do in a
+	// single transaction.
+	MaxSwapsInTx uint64
+	// MaxRedeemsInTx is the max amount of redemptions that this wallet can do
+	// in a single transaction.
+	MaxRedeemsInTx uint64
```
chappjc, line 215: Pretty soon these will be slices/maps for multiple supported contracts versions (if we indeed support multiple at once).  Reasonable for now though.

chappjc (approved):

PR #1902, opened 2022-10-15
fix typos and add hardcoded tooltips to translation

ukane-philemon:


client/webserver/locales/en-us.go
```diff
@@ -217,7 +218,8 @@ var EnUS = map[string]string{
 	"Maximum Possible Swap Fees":  "Maximum Possible Swap Fees",
 	"max_fee_conditions":          "This is the most you would ever pay in fees on your swap. Fees are normally assessed at a fraction of this rate. The maximum is not subject to changes once your order is placed.",
 	"wallet_logs":                 "Wallet Logs",
-	"accelerate_order":            "Accelerate Order",
+	"Accelerate_Order":            "Accelerate Order",
+	"accelerate_order":            "accelerate order",
```
chappjc, line 222: Interesting issue, and not the first time we've had to do this.

I'm considering something like:

```
	"Accelerate_Order":            ":title:accelerate_order",
	"accelerate_order":            "accelerate order",
```

With something like this:

```diff
diff --git a/client/webserver/template.go b/client/webserver/template.go
index 554ecf3c..e30e0b24 100644
--- a/client/webserver/template.go
+++ b/client/webserver/template.go
@@ -12,6 +12,8 @@ import (
        "strings"
 
        "decred.org/dcrdex/client/webserver/locales"
+       "golang.org/x/text/cases"
+       "golang.org/x/text/language"
 )
 
 // pageTemplate holds the information necessary to process a template. Also
@@ -27,6 +29,7 @@ type templates struct {
        fs           fs.FS // must contain tmpl files at root
        reloadOnExec bool
        dict         map[string]string
+       titler       cases.Caser
 
        addErr error
 }
@@ -44,6 +47,7 @@ func newTemplates(folder, lang string) *templates {
                t.addErr = fmt.Errorf("no translation dictionary found for lang %q", lang)
                return t
        }
+       t.titler = cases.Title(language.MustParse(lang))
 
        if embedded {
                t.fs = htmlTmplSub
@@ -77,6 +81,12 @@ func (t *templates) translate(name string) (string, error) {
                                return "", fmt.Errorf("warning: no translation text for key %q", key)
                        }
                }
+               if titleKey := strings.TrimPrefix(replacement, ":title:"); titleKey != replacement {
+                       toTitle, found := t.dict[titleKey]
+                       if found {
+                               replacement = t.titler.String(toTitle)
+                       }
+               }
                rawTmpl = bytes.ReplaceAll(rawTmpl, token, []byte(replacement))
        }
```

dex/networks/ltc/block_test.go
```diff
@@ -16,29 +16,29 @@ import (
...
 
 	// Block 2319633 is version 20000000 with a MWEB, 2 txns, one coinbase and
 	// one integration.
 	// e9fe2c6496aedefa8bf6529bdc5c1f9fd4af565ca4c98cab73e3a1f616fb3502
-	//go:embed testnet4Block2319633.dat
+	//go:embed test-data/testnet4Block2319633.dat
 	block2319633 []byte
 
-	//go:embed block2215586testnet4.dat
+	//go:embed test-data/block2215586testnet4.dat
```
chappjc, line 41: Looks like I checked in this block twice with different names.  For consistency with the other block files, we should use the one that starts with `testnet4Block`

server/db/driver/pg/matches.go
```diff
@@ -136,7 +136,7 @@ func activeSwaps(ctx context.Context, dbe *sql.DB, tableName string) (matches []
 // successfully completed by the specified user, or (2) failed with the user
 // being the at-fault party. Note that the MakerRedeemed match status may be
 // either a success or failure depending on if the user was the maker or taker
-// in the swap, respectively, and the MatchOutcome.Fail flag disambiguates this.
+// in the swap, respectively, and the MatchOutcome.Fail flag disambiguate's this.
```
chappjc, line 139: `disambiguates` is right I believe.  It's a verb.

client/webserver/locales/en-us.go
chappjc: A different approach would use that `:title:` prefix in the .tmpl. That might be more versatile.

ukane-philemon: > A different approach would use that :title: prefix in the .tmpl.

Yeah, I think this works. But in another work, we might need to have a consistent key format for translations.

client/webserver/locales/en-us.go
```diff
@@ -217,7 +218,8 @@ var EnUS = map[string]string{
 	"Maximum Possible Swap Fees":  "Maximum Possible Swap Fees",
 	"max_fee_conditions":          "This is the most you would ever pay in fees on your swap. Fees are normally assessed at a fraction of this rate. The maximum is not subject to changes once your order is placed.",
 	"wallet_logs":                 "Wallet Logs",
-	"accelerate_order":            "Accelerate Order",
+	":title:accelerate_order":     "Accelerate Order",
+	"accelerate_order":            "accelerate order",
```
chappjc, line 222: The second suggestion in https://github.com/decred/dcrdex/pull/1902#discussion_r996459520 is to avoid needing two definitions at all, only needing the non-title one.

In the locales files we'd only have `"accelerate_order":            "accelerate order",` (for each lang), the tmpl would use `[[[:title:accelerate_order]]]` when title case is desired, and then use `golang.org/x/text/cases.Caser` for the title case version.

```diff
diff --git a/client/webserver/template.go b/client/webserver/template.go
index 554ecf3c..ca98aacb 100644
--- a/client/webserver/template.go
+++ b/client/webserver/template.go
@@ -12,6 +12,7 @@ import (
        "strings"
 
        "decred.org/dcrdex/client/webserver/locales"
+       "golang.org/x/text/cases"
 )
 
 // pageTemplate holds the information necessary to process a template. Also
@@ -27,6 +28,7 @@ type templates struct {
        fs           fs.FS // must contain tmpl files at root
        reloadOnExec bool
        dict         map[string]string
+       titler       cases.Caser
 
        addErr error
 }
@@ -70,6 +72,11 @@ func (t *templates) translate(name string) (string, error) {
                        return "", fmt.Errorf("can't parse match group: %v", matchGroup)
                }
                token, key := matchGroup[0], string(matchGroup[1])
+               var toTitle bool
+               if titleKey := strings.TrimPrefix(key, ":title:"); titleKey != key {
+                       toTitle = true
+                       key = titleKey
+               }
                replacement, found := t.dict[key]
                if !found {
                        replacement, found = locales.EnUS[key]
@@ -77,6 +84,9 @@ func (t *templates) translate(name string) (string, error) {
                                return "", fmt.Errorf("warning: no translation text for key %q", key)
                        }
                }
+               if toTitle {
+                       replacement = t.titler.String(replacement)
+               }
                rawTmpl = bytes.ReplaceAll(rawTmpl, token, []byte(replacement))
```

ukane-philemon: In another work, I will update duplicated translations that could use the `:title:` prefix.

client/webserver/locales/en-us.go
```diff
@@ -95,9 +95,10 @@ var EnUS = map[string]string{
 	"Settled":                        "Settled",
 	"Status":                         "Status",
 	"view order history":             "view order history",
-	"cancel order":                   "cancel order",
+	"cancel_order":                   "cancel order",
 	"order details":                  "order details",
 	"verify_order":                   `Verify <span id="vSideHeader"></span>  Order`,
+	"prevent_temporary_overlocking":  "Prevent temporary over locking of funds",
```
chappjc, line 101: I realize we're sorta making a new word with "overlocking", but I would prefer that or "over-locking" to two words.

dex/marshal.go
```diff
@@ -49,7 +49,7 @@ func (b *Bytes) UnmarshalJSON(encHex []byte) (err error) {
 	if encHex[0] != '"' || encHex[len(encHex)-1] != '"' {
 		return fmt.Errorf("marshalled Bytes, %q, not quoted", string(encHex))
 	}
-	// DecodeString overallocates by at least double, and it makes a copy.
+	// DecodeString over allocates by at least double, and it makes a copy.
```
chappjc, line 52: This is where hyphenation is appropriate.

dex/networks/btc/script_test.go
```diff
@@ -424,7 +424,7 @@ func TestInputInfo(t *testing.T) {
 	if err == nil {
 		t.Fatalf("no error for missing redeem script")
 	}
-	// Redeem script must be parseable.
+	// Redeem script must be parsable.
```
chappjc, line 427: Not sure about this one either, but I think "parseable" is way more common in programming context.

dex/order/match.go
```diff
@@ -177,7 +177,7 @@ type Match struct {
 
 // A UserMatch is similar to a Match, but contains less information about the
 // counter-party, and it clarifies which side the user is on. This is the
-// information that might be provided to the client when they are resyncing
```
chappjc, line 180: eh, ok, but this is also a common word in the computer industry

chappjc (changes requested):

dex/networks/btc/script_test.go
ukane-philemon: Yeah, just a false positive typo. Didn't notice.

ukane-philemon:
@chappjc, thanks for your suggestions.

chappjc (approved):

PR #1903, opened 2022-10-15
client/core: Refactor Login

### LINKED ISSUE #1762: client/core: rethink scope of Login, expand API, separate concerns
In the comments for the changes proposed in https://github.com/decred/dcrdex/pull/1761, I've described the heart of the issue as I see it.

The comment refers to the added mechanism to deal with concurrent Login requests at all:

```go
// The design issue that requires this is that we require multiple Login
// calls to succeed without a Logout between them. To properly solve this,
// Login would actually error if already logged in. However:
//  1. Login gives "restart" capabilities for loading and resuming trades.
//  2. We lack a mechanism for repeated callers (e.g. impatient HTTP
//     requests or sessions from other browsers) to get the result from the
//     already-running Login.
//
// Fundamentally, we need to separate the following notions:
//  1. login to Core (credentials check, get cookies)
//  2. Core startup (prep internal state, e.g. loading data from DB, DB upgrades, etc.)
//  3. login to configured servers (initializeDEXConnections > authDEX).
```

Purpose 1 would also do 2 if not already done.  If 2 already done, 1 would just check pw and get cookies for the other session.

A separate Core method may be needed to force re-running of 2 and/or 3.

Wallet connection could fall into either 2. or 3.

_Originally posted by @chappjc in https://github.com/decred/dcrdex/issues/1761#issuecomment-1205559884_

martonp:
`client/core`: 
- Login is updated not return a `LoginResult`. It only returns an error if the password is incorrect. On the first login after startup or after a logout, the wallets are connected, active trades are resolved, and the dex connections are authed. On subsequent logins, only the password is checked.
- Two new functions, `GetNotifications` and `AuthDEXConnections` are added.
- `loadDBTrades` now adds all trades for which a `walletSet` is able to be generated to the `dexConnection`'s trades map. A field `readyToTick` is added to `trackedTrade`, and `tick` returns early for trades where it is set to false. `resumeTrades` will set `readyToTick` to true for all trades where both wallets can be unlocked. `resumeTrades` is also called whenever a wallet is unlocked or connected in order to possibly resume trades that were previously unable to have both wallets unlocked.
 
`client/webserver`:
- `/api/getnotes` and `/api/authdexconns` are added.

`app`:
- `/api/getnotes` is called after login to retrieve existing notifications since this is now removed from login.
- `/api/authdexconns` is called instead of `/api/login` after a dex account is imported.
- Active trades that have `readyToTick` will have their header displayed in red on the market's page. Also a message asking the user to unlock their wallets is displayed.

Screen Shot 2022-11-22 at 9 28 04 AM

Closes #1762

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1307,13 +1307,15 @@ type Core struct {
 	ctx           context.Context
 	wg            sync.WaitGroup
 	ready         chan struct{}
-	loginSlot     chan struct{}
 	cfg           *Config
 	log           dex.Logger
 	db            db.DB
 	net           dex.Network
 	lockTimeTaker time.Duration
 	lockTimeMaker time.Duration
+
+	loginMtx sync.Mutex
+	loggedIn bool
 
 	locale        map[Topic]*translation
 	localePrinter *message.Printer
@@ -1436,7 +1438,6 @@ func New(cfg *Config) (*Core, error) {
 		credentials:   creds,
 		ready:         make(chan struct{}),
 		log:           cfg.Logger,
-		loginSlot:     make(chan struct{}, 1),
 		db:            boltDB,
 		conns:         make(map[string]*dexConnection),
 		wallets:       make(map[uint32]*xcWallet),
@@ -1737,7 +1738,10 @@ func (c *Core) storeDepositAddress(wdbID []byte, addr string) error {
 	return c.db.UpdateWallet(dbWallet)
 }
 
-func (c *Core) connectAndUpdateWallet(w *xcWallet) error {
+// connectAndUpdateWalletResumeTrades creates a connection to a wallet and
+// updates the balance. If resumeTrades is set to true, an attempt to resume
+// any trades that were unable to be resumed at startup will be made.
+func (c *Core) connectAndUpdateWalletResumeTrades(w *xcWallet, resumeTrades bool) error {
 	assetID := w.AssetID
 
 	token := asset.TokenInfo(assetID)
@@ -1747,7 +1751,7 @@ func (c *Core) connectAndUpdateWallet(w *xcWallet) error {
 			return fmt.Errorf("token %s wallet has no %s parent?", unbip(assetID), unbip(token.ParentID))
 		}
 		if !parentWallet.connected() {
-			if err := c.connectAndUpdateWallet(parentWallet); err != nil {
+			if err := c.connectAndUpdateWalletResumeTrades(parentWallet, resumeTrades); err != nil {
 				return fmt.Errorf("failed to connect %s parent wallet for %s token",
 					unbip(token.ParentID), unbip(assetID))
 			}
@@ -1756,7 +1760,7 @@ func (c *Core) connectAndUpdateWallet(w *xcWallet) error {
 
 	c.log.Infof("Connecting wallet for %s", unbip(assetID))
 	addr := w.currentDepositAddress()
-	newAddr, err := c.connectWallet(w)
+	newAddr, err := c.connectWalletResumeTrades(w, resumeTrades)
 	if err != nil {
 		return fmt.Errorf("connectWallet: %w", err) // core.Error with code connectWalletErr
 	}
@@ -1778,6 +1782,12 @@ func (c *Core) connectAndUpdateWallet(w *xcWallet) error {
 	return nil
 }
 
+// connectAndUpdateWallet creates a connection to a wallet and updates the
+// balance.
+func (c *Core) connectAndUpdateWallet(w *xcWallet) error {
+	return c.connectAndUpdateWalletResumeTrades(w, true)
+}
+
 // connectedWallet fetches a wallet and will connect the wallet if it is not
 // already connected. If the wallet gets connected, this also emits WalletState
 // and WalletBalance notification.
@@ -1796,11 +1806,13 @@ func (c *Core) connectedWallet(assetID uint32) (*xcWallet, error) {
 	return wallet, nil
 }
 
-// connectWallet connects to the wallet and returns the deposit address
-// validated by the xcWallet after connecting. If the wallet backend is still
-// syncing, this also starts a goroutine to monitor sync status, emitting
-// WalletStateNotes on each progress update.
-func (c *Core) connectWallet(w *xcWallet) (depositAddr string, err error) {
+// connectWalletResumeTrades connects to the wallet and returns the deposit
+// address validated by the xcWallet after connecting. If the wallet backend
+// is still syncing, this also starts a goroutine to monitor sync status,
+// emitting WalletStateNotes on each progress update. If resumeTrades is set to
+// true, an attempt to resume any trades that were unable to be resumed at
+// startup will be made.
+func (c *Core) connectWalletResumeTrades(w *xcWallet, resumeTrades bool) (depositAddr string, err error) {
 	if w.isDisabled() {
 		return "", fmt.Errorf(walletDisabledErrStr, strings.ToUpper(unbip(w.AssetID)))
 	}
@@ -1808,6 +1820,12 @@ func (c *Core) connectWallet(w *xcWallet) (depositAddr string, err error) {
 	err = w.Connect() // ensures valid deposit address
 	if err != nil {
 		return "", codedError(connectWalletErr, err)
+	}
+
+	// This may be a wallet that does not require a password, so we can attempt
+	// to resume any active trades.
+	if resumeTrades {
+		go c.resumeTrades(nil)
 	}
 
 	w.mtx.RLock()
@@ -1823,16 +1841,18 @@ func (c *Core) connectWallet(w *xcWallet) (depositAddr string, err error) {
 	return
 }
 
-// Connect to the wallet if not already connected. Unlock the wallet if not
-// already unlocked.
-func (c *Core) connectAndUnlock(crypter encrypt.Crypter, wallet *xcWallet) error {
-	if !wallet.connected() {
-		c.log.Infof("Connecting wallet for %s", unbip(wallet.AssetID))
-		err := c.connectAndUpdateWallet(wallet)
-		if err != nil {
-			return err
-		}
-	}
+// connectWallet connects to the wallet and returns the deposit address
+// validated by the xcWallet after connecting. If the wallet backend is still
+// syncing, this also starts a goroutine to monitor sync status, emitting
+// WalletStateNotes on each progress update.
+func (c *Core) connectWallet(w *xcWallet) (depositAddr string, err error) {
+	return c.connectWalletResumeTrades(w, true)
+}
+
+// unlockWalletResumeTrades will unlock a wallet if it is not yet unlocked. If
+// resumeTrades is set to true, an attempt to resume any trades that were
+// unable to be resumed at startup will be made.
+func (c *Core) unlockWalletResumeTrades(crypter encrypt.Crypter, wallet *xcWallet, resumeTrades bool) error {
 	// Unlock if either the backend itself is locked or if we lack a cached
 	// unencrypted password for encrypted wallets.
 	if !wallet.unlocked() {
@@ -1850,8 +1870,44 @@ func (c *Core) connectAndUnlock(crypter encrypt.Crypter, wallet *xcWallet) error
 		}
 		// Notify new wallet state.
 		c.notify(newWalletStateNote(wallet.state()))
+
+		if resumeTrades {
+			go c.resumeTrades(crypter)
+		}
 	}
+
 	return nil
+}
+
+// unlockWallet will unlock a wallet if it is not yet unlocked.
+func (c *Core) unlockWallet(crypter encrypt.Crypter, wallet *xcWallet) error {
+	return c.unlockWalletResumeTrades(crypter, wallet, true)
+}
+
+// connectAndUnlockResumeTrades will connect to the wallet if not already
+// connected, and unlock the wallet if not already unlocked. If the wallet
+// backend is still syncing, this also starts a goroutine to monitor sync
+// status, emitting WalletStateNotes on each progress update. If resumeTrades
+// is set to true, an attempt to resume any trades that were unable to be
+// resumed at startup will be made.
+func (c *Core) connectAndUnlockResumeTrades(crypter encrypt.Crypter, wallet *xcWallet, resumeTrades bool) error {
+	if !wallet.connected() {
+		c.log.Infof("Connecting wallet for %s", unbip(wallet.AssetID))
+		err := c.connectAndUpdateWalletResumeTrades(wallet, resumeTrades)
+		if err != nil {
+			return err
+		}
+	}
+
+	return c.unlockWalletResumeTrades(crypter, wallet, resumeTrades)
+}
+
+// connectAndUnlock will connect to the wallet if not already connected,
+// and unlock the wallet if not already unlocked. If the wallet backend
+// is still syncing, this also starts a goroutine to monitor sync status,
+// emitting WalletStateNotes on each progress update.
+func (c *Core) connectAndUnlock(crypter encrypt.Crypter, wallet *xcWallet) error {
+	return c.connectAndUnlockResumeTrades(crypter, wallet, true)
 }
 
 // walletBalance gets the xcWallet's current WalletBalance, which includes the
@@ -2267,7 +2323,7 @@ func (c *Core) createWalletOrToken(crypter encrypt.Crypter, walletPW []byte, for
 		return nil, fmt.Errorf(s, a...)
 	}
 
-	err = wallet.Unlock(crypter) // no-op if !wallet.Wallet.Locked() && len(encPW) == 0
+	err = c.unlockWallet(crypter, wallet) // no-op if !wallet.Wallet.Locked() && len(encPW) == 0
 	if err != nil {
 		wallet.Disconnect()
 		return nil, fmt.Errorf("%s wallet authentication error: %w", symbol, err)
@@ -2794,7 +2850,7 @@ func (c *Core) RecoverWallet(assetID uint32, appPW []byte, force bool) error {
 
 		c.updateAssetWalletRefs(newWallet)
 
-		err = newWallet.Unlock(crypter)
+		err = c.unlockWallet(crypter, newWallet)
 		if err != nil {
 			return err
 		}
@@ -2816,7 +2872,7 @@ func (c *Core) OpenWallet(assetID uint32, appPW []byte) error {
 	if err != nil {
 		return fmt.Errorf("OpenWallet: wallet not found for %d -> %s: %w", assetID, unbip(assetID), err)
 	}
-	err = wallet.Unlock(crypter)
+	err = c.unlockWallet(crypter, wallet)
 	if err != nil {
 		return newError(walletAuthErr, "failed to unlock %s wallet: %w", unbip(assetID), err)
 	}
@@ -3118,7 +3174,7 @@ func (c *Core) ReconfigureWallet(appPW, newWalletPW []byte, form *WalletForm) er
 	// setWalletPassword since it would use connectAndUpdateWallet, which
 	// performs additional deposit address validation and balance updates that
 	// are redundant with the rest of this function.
-	dbWallet.Address, err = c.connectWallet(wallet)
+	dbWallet.Address, err = c.connectWalletResumeTrades(wallet, false)
 	if err != nil {
 		return fmt.Errorf("connectWallet: %w", err)
 	}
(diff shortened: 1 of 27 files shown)


client/core/core.go
```diff
@@ -4067,13 +3995,82 @@ func (c *Core) Login(pw []byte) (*LoginResult, error) {
...
+	}
+
+	return c.initializeDEXConnections(crypter), nil
+}
+
+// Login logs the user in. On the first login after startup or after a logout,
+// this function will connect wallets, resolve active trades, and decrypt
+// account keys for all known DEXes. Otherwise, it will only check whether or
+// not the app pass is correct.
+func (c *Core) Login(pw []byte) error {
```
buck54321, line 4015: Could you move `Login` back above `connectWallets`? Much of this is unchanged.

client/core/core.go
```diff
@@ -142,6 +142,12 @@ type dexConnection struct {
 	// trades tracks outstanding orders issued by this client.
 	trades map[order.OrderID]*trackedTrade
 
+	failedTradesMtx sync.RWMutex
+	// failedTrades tracks trades that were unable to load at startup.
+	// Attempts to resume these trades are made when wallets are unlocked.
+	// Trades that are successfully resumed will be removed from this slice.
+	failedTrades []*trackedTrade
```
buck54321, line 149: Can we do this differently? If my read is right, these orders wouldn't show up in the e.g. `User` until they connect their wallet. I'd really like all orders to make it in to the `dexConnection.trades` map ASAP so that the user can at least see them.

martonp: Yeah the behavior is unchanged for the user, except instead of having to re-login after unlocking the wallets, it will automatically load the trades when the wallets it needs are unlocked. I thought the whole point of the logic currently in `loadDBTrades` (which I have now moved to `checkIfTradesAreReady`) was to avoid having trades that have wallets which are unable to be unlocked to be in `dexConnection.trades`, so that `tick` is not constantly called with them.

I was thinking to use the `failedTrades` slice in another PR to create a popup after the user logs in to notify them that some trades could not be loaded due to whichever asset's wallet not being connected. Also, the "Your Orders" section could have some sort of red button saying "Active Trades failed to Load" that allows the user to see the popup again.

client/core/core.go
```diff
@@ -6172,15 +6188,38 @@ func (c *Core) loadDBTrades(dc *dexConnection, crypter encrypt.Crypter, failed m
 		}
 		if baseFailed {
 			errs.add("could not complete order %s because the wallet for %s cannot be used", trade.token(), unbip(base))
-			continue
 		}
 		if quoteFailed {
 			errs.add("could not complete order %s because the wallet for %s cannot be used", trade.token(), unbip(quote))
-			continue
```
chappjc, line 6179: Happy this is being addressed.  Just not loading the orders for wallets that aren't connected or cannot unlock was poor.

I'm realizing that with my latest merged work where `walletSet` is now just the `xcWallets`, we seem to be able to set `trackedTrade.wallets` in `loadDBTrades` when initially constructing with `newTrackedTrade` instead of later in `resumeTrades` (and not failing on connect or unlock errors, rather just using `c.wallet` instead of `c.connectedWallet` and `unlock`, which can still be attempted before `resumeTrades`).  That would mean that "ready" trades include any trade with a complete `walletSet` **regardless of if there are wallet connection or unlocking issues.**  And really, don't we want all trades listed in the trades map so the user is aware and can fix the issue?

chappjc: Oh, I just read your comment in https://github.com/decred/dcrdex/pull/1903#discussion_r1019525129 saying you intend to list the failed ones.  Still see if my comment makes sense w.r.t. setting the `walletSet` field of the `trackedTrades` in `loadDBTrades` if possible.  Would that improve anything, or just make spam the logs with errors on each `tick`?

martonp: Yeah, there will be a constant spam of error logs for each trade that failed to load on each `tick`. Does the approach I wrote in that comment sound good?

chappjc: I'm not sure yet, but I think we want `trackedTrade`s with wallet issues to still be loaded so you can cancel the orders.  There are three general conditions in `checkIfTradesAreReady` that make a trade "not ready" or "failed to load":

1. No wallet definition for one of the assets, hence cannot make an `xcWallet`.  This is when `c.wallet` errors in `c.connectedWallet`.
2. Wallet not connected, but there was a wallet definition.  This is when `c.connectAndUpdateWallet` errors in `c.connectedWallet`.
3. `(*xcWallet).Unlock` errors

The first is resolvable by creating the wallet, but I don't know how this would happen.

The second is resolvable by starting the external wallet, or perhaps "recovering" a busted native wallet.

The third is resolvable by reconfiguring the wallet with the new password.

Each of the resolutions would need to trigger a re-check of the `failedTrades`.  I don't know that `(*Core).OpenWallet` would cover any of the above scenarios.

In all but the first of those three causes we can make a complete `walletSet` for the `trackedTrade.wallet` fields.  And only trades with actively negotiating swaps would actually spam with errors on `tick` I think.  If the order is still booked, regardless of whether it has active swaps, being able to cancel that order without resolving the wallet situation would be desirable.

buck54321: > In all but the first of those three causes we can make a complete `walletSet` for the `trackedTrade.wallet` fields.

That first one would be a doozy.

As long as we can make a `walletSet`, I think the order can go in the `trades` map. Where in tick would we be logging? After `refreshUnlock`? Could also set a flag on the `trackedTrade` to bail early in `tick`, right after `t.mtx.RLock()`, I guess.

martonp: These types of logs would show up in every tick: `not checking if order %s, match %s is redeemable because %s wallet is locked or disabled`.

I've added a flag to `trackedTrade` that blocks the ticks. If the `walletSet` cannot be created, we get the same behavior as before, but all trades are added to the trades map now. Let me know what you guys think.

client/core/core.go
```diff
@@ -4170,21 +4192,26 @@ func (c *Core) Login(pw []byte) (*LoginResult, error) {
...
+	if err != nil {
+		return nil, err
 	}
 
-	return loginResult(), nil
+	return c.initializeDEXConnections(crypter), nil
+}
+
+// GetNotifications loads the latest notifications from the db.
+func (c *Core) GetNotifications(n int) ([]*db.Notification, error) {
```
chappjc, line 4209: `s/GetNotifications/Notifications/g`

A "get" prefix on method names is rarely valuable, unless is has really needed to convey there is some remote fetching going on, but usually not even in that case.

It's idiomatic to elide the get [for "getters"](https://go.dev/doc/effective_go#Getters), and the reasoning applies to most other kinds of methods too.

client/webserver/api.go
```diff
@@ -1129,13 +1129,7 @@ func (s *WebServer) actuallyLogin(w http.ResponseWriter, r *http.Request, login
...
 	}
 
-	writeJSON(w, struct {
-		OK    bool               `json:"ok"`
-		Notes []*db.Notification `json:"notes"`
-	}{
-		OK:    true,
-		Notes: loginResult.Notifications,
-	}, s.indent)
+	writeJSON(w, simpleAck(), s.indent)
```
chappjc, line 1132: Although `(*Core).Login` no longer returns the notes, the `/api/login` http handler could retrieve the notifications after login and return that in the response, with the identical response structure.
Looking at the JS, it looks like it always does `/api/getnotes` after `/api/login`.  Is there a use case for having them separate?  Seems it could be made more efficient for the frontend.

client/webserver/api.go
```diff
@@ -1153,6 +1147,48 @@ func (s *WebServer) apiUser(w http.ResponseWriter, r *http.Request) {
...
+		Notes []*db.Notification `json:"notes"`
+		OK    bool               `json:"ok"`
+	}{
+		Notes: notes,
+		OK:    true,
+	}
+	writeJSON(w, response, s.indent)
+}
+
+// apiAuthDexConnections handles the 'authdexconns' API request.
```
chappjc, line 1167: Maybe state when this would be used.  The use case is... importing an account but not needing to re-login to dexc?

Depending on the answer, that might imply that `apiAccountImport` should be updated to make the `(*Core).AuthDEXConnections` call rather than troubling the frontend with sequential api requests.

client/webserver/site/src/js/forms.ts
```diff
@@ -1598,10 +1598,11 @@ export class LoginForm {
       Doc.show(page.errMsg)
       return
     }
-    if (res.notes) {
-      res.notes.reverse()
+    const notesRes = await getJSON('api/getnotes')
```
chappjc, line 1601: Looks like a leading `/` is missing, but consider if this separate endpoint is even needed as per my other comment.

client/core/core.go
```diff
@@ -4170,21 +4192,26 @@ func (c *Core) Login(pw []byte) (*LoginResult, error) {
...
-	// may be accessible this time.
-	c.resolveActiveTrades(crypter)
-	// Unlock account keys and attempt to authorize with each dexConnection.
-	// Even if connect fails, this prepares the account private keys so authDEX
-	// triggered by the reconnect handler can succeed.
-
-	if err := c.loadBotPrograms(); err != nil {
-		c.log.Errorf("Error loading bot programs: %v", err)
+// AuthDEXConnections connects to the each of the DEX servers in the conns map
+// and authenitcates the connection.
```
chappjc, line 4198: authenitcates

chappjc (changes requested):
Some quick thoughts about the webserver/middleware/frontend bits.  Moving to backend...

client/webserver/api.go
chappjc: Or perhaps `(*Core).AccountImport` should be doing `(*Core).initializeDEXConnections` and there's no need at all for `(*Core).AuthDEXConnections`?   It's not clear when a consumer would want to use `AuthDEXConnections` more generally.

client/webserver/api.go
martonp: Good idea.

client/webserver/api.go
martonp: I've removed `AuthDexConnections`. The only thing that is missing now that we had before, is users of `dexcctl` seeing which servers the client was able to successfully connect to at login. They also saw a huge list of 100 notifications previously, which was a bit distracting.

client/core/core.go
```diff
@@ -1660,7 +1661,10 @@ func (c *Core) storeDepositAddress(wdbID []byte, addr string) error {
 	return c.db.UpdateWallet(dbWallet)
 }
 
-func (c *Core) connectAndUpdateWallet(w *xcWallet) error {
+// connectAndUpdateWalletResumeTrades creates a connection to a wallet and
+// updates the balance. If resumeTrades is set to true, an attempt to resume
+// and trades that were unable to be resumed at startup will be made.
+func (c *Core) connectAndUpdateWalletResumeTrades(w *xcWallet, resumeTrades bool) error {
```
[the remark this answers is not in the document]
martonp: Yeah I meant to call `connectWalletResumeTrades`.

client/rpcserver/handlers.go
```diff
@@ -399,19 +399,20 @@ func handleExchanges(s *RPCServer, _ *RawParams) *msgjson.ResponsePayload {
...
-// handleLogin sets up the dex connection and returns core.LoginResult.
+// handleLogin sets up the dex connections.
 func handleLogin(s *RPCServer, params *RawParams) *msgjson.ResponsePayload {
 	appPass, err := parseLoginArgs(params)
 	if err != nil {
 		return usage(loginRoute, err)
 	}
 	defer appPass.Clear()
-	res, err := s.core.Login(appPass)
+	err = s.core.Login(appPass)
```
buck54321, line 409: As far as I can tell, this leaves no way to get notifications through the RPC server.

client/webserver/site/src/js/markets.ts
```diff
@@ -1252,6 +1258,7 @@ export default class MarketsPage extends BasePage {
       })
       app().bindTooltips(div)
     }
+    Doc.setVis(unreadyOrders, page.unreadyOrdersMsg)
```
buck54321, line 1261: I wonder if we'll want a way to indicate a problem even if the user is on a different market. Don't have to do it now. Could be a `db.ErrorLevel` notification maybe. Or maybe just well-placed messaging. Either way, I'd hate it if my orders on one market were not processing but I'm unaware because I'm looking at a different market.

client/core/core.go
```diff
@@ -1850,10 +1870,46 @@ func (c *Core) connectAndUnlock(crypter encrypt.Crypter, wallet *xcWallet) error
...
+	return c.unlockWalletResumeTrades(crypter, wallet, true)
+}
+
+// connectAndUnlockResumeTrades will connect to the wallet if not already
+// connected, and unlock the wallet if not already unlocked. If the wallet
+// backend is still syncing, this also starts a goroutine to monitor sync
+// status, emitting WalletStateNotes on each progress update. If resumeTrades
+// is set to true, an attempt to resume and trades that were unable to be
+// resumed at startup will be made.
+func (c *Core) connectAndUnlockResumeTrades(crypter encrypt.Crypter, wallet *xcWallet, resumeTrades bool) error {
```
buck54321, line 1893: One thing that smells is the `resumeTrades bool` arguments being passed around. My intuition is that this hides an underlying deficiency where we should be doing a more sophisticated form of resolution once new wallets come online, i.e. we should be able to tell if trades need to be resumed without the caller's input.

martonp: The reason for passing it around is to avoid recursive calls from happening when `connectAndUnlockWallet` is called from `resumeTrade`. I guess it's not completely necessary..

client/webserver/site/src/js/markets.ts
martonp: I was thinking to do a popup that shows up on the markets page after login in another PR.

client/rpcserver/handlers.go
martonp: New `notifications` command?

client/core/core.go
```diff
@@ -6116,80 +6145,76 @@ func (c *Core) dbTrackers(dc *dexConnection) (map[order.OrderID]*trackedTrade, e
...
-			} else if !quoteWallet.unlocked() {
-				err = quoteWallet.Unlock(crypter)
-				if err != nil {
-					quoteFailed = true
-					failed[quote] = struct{}{}
-					c.log.Errorf("Unlock wallet %s failed: %v", unbip(quote), err)
-				}
+
+		tracker.wallets = walletSet
+		dc.trades[tracker.ID()] = tracker
```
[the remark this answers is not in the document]
chappjc: I haven't digested the logic yet as it pertains to the `resumeTrades bool`, but we should be sure never to overwrite an existing map entry.  If there are two `trackedTrade` instances for the same orders, bad things can and will happen.

client/rpcserver/handlers.go
chappjc: I'm not convinced its needed for the RPC server.  Even headless Go consumers likely just dump the notifications.

client/core/core.go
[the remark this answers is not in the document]
martonp: I don't think an overwrite could happen because this function is only called at login, but I've added a check to make sure.

The `resumeTrades bool` arguments have anything to do with adding trades to the map. It's only to avoid recursive calls back to `resumeTrades` when wallets are being unlocked in `resumeTrade`.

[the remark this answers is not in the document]
chappjc: In `(*Core).resumeTrade`, there is a large amount of work being done even for trades that don't need to be resumed i.e. are already running fine.  We should really leave alone trades that are fine.  Calling FindingCoins repeatedly, messing with account-based wallet reserves, etc.  Need to avoid all that somehow.  It looks to me like all orders that would load fine are going to go through all this, return true from `resumeTrade`, and emit an order note.

We talked about having the lame trades in the trades map for reasons, but that doesn't mean we can't add a tracking structure as a field to `dexConnections` to assist.

That's potentially separate, but I also thing we should look at the need for the bool because it does hint that something is amiss in the design.

A related thing that stands out is when wallet A is unlocked, we try to resume all trades even if they have nothing to do with wallet A.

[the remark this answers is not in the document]
chappjc: Ohh, `readyToTick` blocks calling these again?  OK.  Please consider the other points though.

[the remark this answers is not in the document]
martonp: > Seems like resumeTrade could check this too and return true.

Good call, I'll add a check. 

> That's potentially separate, but I also think we should reconsider at the need for the bool because it does hint that something is amiss in the design.

I was considering putting the lame trades into a separate map, and then having `cancelOrder` search through those as well, but it seemed cleaner to do it with the bool. Why is this a bad design? The `trades` map has all of the active trades for a dex, but some of them could not yet complete pre-processing, so we mark this with a flag. It seems decently straightforward.

> A related thing that stands out is when wallet A is unlocked, we try to resume all trades even if they have nothing to do with wallet A.

I considered this also.. we could pass an assetID filter to `resumeTrades`, but the max cost of not doing it is one failed call to per wallet that has active trades.
Maybe it's far fetched, but there could be a case where the user has an active trade with asset A-B, and one with A-C, and both A and B initially failed to connect. Then they start both the A and B RPC wallets, unlock the B wallet, and A is able to automatically unlock. If only resume trades involving wallet B, then the A-C trade will not be resumed. Unless we allow recursive calls back to `resumeTrades`.

chappjc (changes requested):
Looking pretty good to me!

client/core/core.go
```diff
@@ -6324,80 +6323,82 @@ func (c *Core) dbTrackers(dc *dexConnection) (map[order.OrderID]*trackedTrade, e
...
-					quoteFailed = true
-					failed[quote] = struct{}{}
-					c.log.Errorf("Unlock wallet %s failed: %v", unbip(quote), err)
-				}
+
+		tracker.wallets = walletSet
+		dc.tradeMtx.Lock()
+		if _, found := dc.trades[tracker.ID()]; found {
+			dc.tradeMtx.Unlock()
+			return nil
```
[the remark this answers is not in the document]
martonp: `Logout()` won't work if there are active deals in the map. I don't see any way this will get hit currently unless there is a bug causing duplicates to be returned from the db, but I 100% agree the check should be there. As you say, `loadDBTrades` could be called differently in the future.

client/core/core.go
```diff
@@ -6324,80 +6323,82 @@ func (c *Core) dbTrackers(dc *dexConnection) (map[order.OrderID]*trackedTrade, e
...
-					failed[base] = struct{}{}
-					c.log.Errorf("Unlock wallet %s failed: %v", unbip(base), err)
-				}
-			}
+
+		trade := tracker.Trade()
+
+		walletSet, assetConfigs, versCompat, err := c.walletSet(dc, tracker.Base(), tracker.Quote(), trade.Sell)
+		if err != nil {
+			errs.add("failed to load wallets for trade ID %s: %w", tracker.ID(), err)
```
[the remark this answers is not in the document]
martonp: Yeah makes sense. A notification is definitely better than an unreadable error message.

client/webserver/site/src/js/markets.ts
```diff
@@ -1897,10 +1897,16 @@ export default class MarketsPage extends BasePage {
...
-    // If metaOrder doesn't exist for the given order it means it was created
-    // via dexcctl and the GUI isn't aware of it or it was an inflight order.
-    // Call refreshActiveOrders to grab the order.
-    if (!mord || note.topic === 'AsyncOrderFailure') return this.refreshActiveOrders()
+    // - If metaOrder doesn't exist for the given order it means it was created
+    //  via dexcctl and the GUI isn't aware of it or it was an inflight order.
+    //  refreshActiveOrders must be called to grab this order.
+    // - If an OrderLoaded notification is recieved, it means an order that was
+    //   previously not "ready to tick" (due to its wallets not being connected
+    //   and unlocked) has now become ready to tick. The active orders section
```
chappjc, line 1905: Doesn't this also happen when the user just logs in with active orders?  Won't it emit one of these OrderLoaded topic notes for each active order on initial login, causing it to refresh active orders multiple times?

martonp: In those cases `readyToTick` will be false and the market page will not be loaded yet. I'll add a new notification topic though.

martonp: Yes, but those will be before login is complete, so the market page will not yet be loaded.

chappjc: > Yes, but those will be before login is complete, so the market page will not yet be loaded.

Ah, OK, perfect.

chappjc:
I noticed a nil context panic in `btc.externalFeeEstimator`, but I'll get to that in https://github.com/decred/dcrdex/pull/1967 right after this goes in.  Saw a duplicate log statement too, but otherwise working as expected.

I had forgotten that "unfunded" orders get automatically canceled in `authDEX`.  Since these never get `tracker.coinsLocked = true` in `resumeTrade`, booked orders that don't have a connected or unlockable wallet get canceled.  I think that's good, I just forgot. Yeah, very good... before these orders wouldn't have even been loaded into the trades map. :grimacing:

Issue #1895, opened 2022-10-09
ui: Orders with pending refund not shown among active orders


martonp:
This is confusing when you try to logout and cannot due to active orders. Also, there is no response on the UI when you try to log out when there are active orders. There should be some sort of popup notifying the user the reason they were unable to log out.

chappjc:
They show as "executed" in the your orders list.  I agree that's not sufficient, and there should be a clear message when logout is blocked.

PR #1904, opened 2022-10-16
app: Logout with active orders error

martonp:
Previously if the user tried to log out while there were active deals, there would be no feedback. Now, an error message is displayed.

Screen Shot 2022-10-16 at 7 54 11 AM

Closes: #1895


client/webserver/site/src/js/wallets.ts
```diff
@@ -962,7 +962,11 @@ export default class WalletsPage extends BasePage {
     const loaded = app().loading(page.newWalletForm)
     const res = await postJSON('/api/closewallet', { assetID: assetID })
     loaded()
-    if (!app().checkResponse(res)) return
+    if (!app().checkResponse(res)) {
+      page.walletErr.textContent = `Error locking wallet: ${res.msg}`
```
ukane-philemon, line 966: might need to add this hardcoded string to translation at `locales.ts`.

client/webserver/site/src/js/app.ts
```diff
@@ -887,9 +888,13 @@ export default class Application {
   async signOut () {
     const res = await postJSON('/api/logout')
     if (!this.checkResponse(res)) {
-      // TODO: Esp. for active orders error, let the user know what's going on.
-      // Maybe give them an option to force.
-      Doc.hide(this.page.profileBox)
+      const activeOrdersErrCode = 35
```
chappjc, line 891: We have this defined at https://github.com/decred/dcrdex/blob/34732a8b7e52346d86e52a03610845e450125059/client/webserver/site/src/js/wallets.ts#L37

Seems like we should have all http API error codes that we need defined somewhere central.  We had a constants.js at one point, but that got nuked.  Thoughts?  Maybe in http.ts or at the top of app.ts?

client/core/core.go
```diff
@@ -4126,7 +4126,7 @@ func (c *Core) Logout() error {
 	conns := c.dexConnections()
 	for _, dc := range conns {
 		if dc.hasActiveOrders() {
-			return ActiveOrdersLogoutErr
+			return newError(activeOrdersErr, "%w", ActiveOrdersLogoutErr)
```
chappjc, line 4129: Without a message, just `codedError(activeOrdersErr, ActiveOrdersLogoutErr)`

chappjc (changes requested):
Working as expected.  Just two minor changes please.

client/webserver/site/src/js/app.ts
martonp: I think that's a good idea. I put them in http.ts.

Issue #1768, opened 2022-08-09
zec: Incorrect available/locked amounts.


JoeGruffins:
Testing on testnet zec_btc because dcr testnet is down. Here trying to make a market sell (limit gives same error) and the error message is confusing. It looks like I should have 1.0119302 zec available, but the error message will tell me I only have 0.0019302 so exactly 1.01 less.

![image](https://user-images.githubusercontent.com/34998433/183612473-fe056162-08d9-4ca4-a6f6-bf479639ce77.png)

![image](https://user-images.githubusercontent.com/34998433/183611241-27bca0f3-c44a-477e-888a-a24723ba35f1.png)

Also, I have about 0.006 zec in sell orders already, so I think that should be showing up in `locked` but it is not.

JoeGruffins:
It would make sense that that the 1.01 is two utxo. One from the testnet wallet and the other send from my other node. I guess it should be showing 1.01 locked is the real problem.

chappjc:
https://github.com/decred/dcrdex/blob/50cf04fcf09919d96c801cae5c03ef4430eb1923/client/asset/btc/btc.go#L1393-L1416

For ZEC, it looks like we just don't set the `Locked` field at all.  @buck54321 Do you recall a reason why we don't also use `btc.lockedSats()` in this `zecStyleBalance` path?  Seems like a bug.

Also, I think we should look again at `(*baseWallet).lockedSats` and how it takes the result of `gettransaction` and does ` uint64(txOut.Value)`.  If it is negative (like it can be with `listtransactions`), we'll have an underflow.

PR #1909, opened 2022-10-18
client/tests: Fix simnet tests.

### LINKED ISSUE #1783: simnet-trade-tests/dcrbtc: Some fail with dcr unlocking error if run together.
To run all the tests, I think, you do `./run dcrbtc -t success -t nomakerswap -t notakerswap -t nomakerredeem -t makerghost -t orderstatus -t resendpending` and this results in the error:
```
ERROR: test "notakerswap" failed: client1 place sell order error: dcr connectAndUnlock error: failed to unlock dcr wallet: dcr unlockWallet decryption error: expected 2 pushes. got 0
```
for a few of the tests.

They seem to be fine if run alone.

Also a command like `--all` to just run all the tests would be cool.

JoeGruffins:
Fix numerous balance calculating issues. Run most tests twice,
    alternating the maker and taker between clients. Add all, except, and
    runonce commands. Cause requests to fail using communications rather
    than altering match data.

closes #1783 and fixes some other problems with them not running.

### DIFF
--- a/client/cmd/simnet-trade-tests/run
+++ b/client/cmd/simnet-trade-tests/run
@@ -86,7 +86,7 @@ dcrdextt - Decred RPC wallet and Ethereum wallet with dextt.eth test token on ma
 
 ---------------------
 
-One or more of the following tests can be run by specifying one or more -t flags. Default is ['success'].
+One or more of the following tests can be run by specifying one or more -t flags. Default is ['success']. All tests for an asset pair can be run with --all. When using --all, tests can be excluded with --except=test_name.
 
 EOF
 
--- a/client/core/simnet_trade.go
+++ b/client/core/simnet_trade.go
@@ -25,7 +25,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"math/rand"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -44,14 +43,17 @@ import (
 	"decred.org/dcrdex/client/asset/ltc"
 	"decred.org/dcrdex/client/asset/zec"
 	"decred.org/dcrdex/client/comms"
-	"decred.org/dcrdex/client/db"
 	"decred.org/dcrdex/dex"
 	"decred.org/dcrdex/dex/calc"
 	"decred.org/dcrdex/dex/config"
-	"decred.org/dcrdex/dex/encode"
+	"decred.org/dcrdex/dex/msgjson"
+	dexbtc "decred.org/dcrdex/dex/networks/btc"
 	dexdoge "decred.org/dcrdex/dex/networks/doge"
+	dexeth "decred.org/dcrdex/dex/networks/eth"
 	"decred.org/dcrdex/dex/order"
 	dexsrv "decred.org/dcrdex/server/dex"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
 	"golang.org/x/sync/errgroup"
 )
 
@@ -124,21 +126,23 @@ type SimulationConfig struct {
 	Client2           *SimClient
 	Tests             []string
 	Logger            dex.Logger
+	RunOnce           bool
 }
 
 type simulationTest struct {
-	ctx        context.Context
-	cancel     context.CancelFunc
-	log        dex.Logger
-	base       *assetConfig
-	quote      *assetConfig
-	regAsset   uint32
-	marketName string
-	lotSize    uint64
-	rateStep   uint64
-	client1    *simulationClient
-	client2    *simulationClient
-	clients    []*simulationClient
+	ctx            context.Context
+	cancel         context.CancelFunc
+	log            dex.Logger
+	base           *assetConfig
+	quote          *assetConfig
+	regAsset       uint32
+	marketName     string
+	lotSize        uint64
+	rateStep       uint64
+	client1        *simulationClient
+	client2        *simulationClient
+	clients        []*simulationClient
+	client1IsMaker bool
 }
 
 func (s *simulationTest) waitALittleBit() {
@@ -147,20 +151,7 @@ func (s *simulationTest) waitALittleBit() {
 		s.client2.BaseWalletType == WTElectrum || s.client2.QuoteWalletType == WTElectrum {
 		sleep = 7 * time.Second
 	}
-	sleep *= sleepFactor
-	feeSleep := time.Millisecond * time.Duration(s.client1.dc().cfg.BroadcastTimeout/2)
-	if sleep < feeSleep {
-		// eth based assets need to wait for txn to be mined for fees
-		// to be populated here.
-		for bipID := range s.client1.wallets {
-			if accountBIPs[bipID] {
-				sleep = feeSleep
-				s.client1.log.Infof("Waiting for paid order fees to be populated (%s).", sleep)
-				break
-			}
-		}
-	}
-	time.Sleep(sleep)
+	time.Sleep(sleep * sleepFactor)
 }
 
 // RunSimulationTest runs one or more simulations tests, based on the provided
@@ -248,13 +239,21 @@ func RunSimulationTest(cfg *SimulationConfig) error {
 	}
 
 	for _, testName := range cfg.Tests {
-		s.log.Infof("Running %q", testName)
+		s.log.Infof("Running %q with cliet 1 as maker", testName)
 		f, ok := testLookup[testName]
 		if !ok {
 			return fmt.Errorf("no test named %q", testName)
 		}
+		s.client1IsMaker = true
 		if err := f(s); err != nil {
-			return fmt.Errorf("test %q failed: %w", testName, err)
+			return fmt.Errorf("test %q failed with client 1 as maker: %w", testName, err)
+		}
+		s.client1IsMaker = false
+		if !cfg.RunOnce && testName != "orderstatus" {
+			s.log.Infof("Running %q with client 2 as maker", testName)
+			if err := f(s); err != nil {
+				return fmt.Errorf("test %q failed with client 2 as maker: %w", testName, err)
+			}
 		}
 	}
 
@@ -404,6 +403,8 @@ func (s *simulationTest) setup(cl1, cl2 *SimClient) (err error) {
 	if err != nil {
 		return fmt.Errorf("error starting clients: %w", err)
 	}
+	s.client1.replaceConns()
+	s.client2.replaceConns()
 	return nil
 }
 
@@ -414,7 +415,7 @@ var sleepFactor time.Duration = 1
 func testTradeSuccess(s *simulationTest) error {
 	var qty, rate uint64 = 2 * s.lotSize, 150 * s.rateStep
 	s.client1.isSeller, s.client2.isSeller = true, false
-	return s.simpleTradeTest(qty, rate, order.MatchComplete)
+	return s.simpleTradeTest(qty, rate, order.MatchConfirmed)
 }
 
 // TestNoMakerSwap runs a simple trade test and ensures that the resulting
@@ -445,6 +446,38 @@ func testNoTakerSwap(s *simulationTest) error {
 func testNoMakerRedeem(s *simulationTest) error {
 	var qty, rate uint64 = 1 * s.lotSize, 250 * s.rateStep
 	s.client1.isSeller, s.client2.isSeller = true, false
+
+	var bits uint8
+	go func() {
+		<-time.After(tLockTimeTaker)
+		s.client1.filteredConn.requestFilter.Store(func(string) error { return nil })
+		s.client2.filteredConn.requestFilter.Store(func(string) error { return nil })
+		if bits == 0b1 {
+			s.client1.enableWallets()
+		} else {
+			s.client2.enableWallets()
+		}
+	}()
+	preFilter1 := func(route string) error {
+		// Fail every first try of init and redeem. Second try will be
+		// passed on to the real comms.
+		if route == msgjson.InitRoute && bits == 0 {
+			bits = 0b1
+			s.client1.disableWallets()
+		}
+		return nil
+	}
+	preFilter2 := func(route string) error {
+		if route == msgjson.InitRoute && bits == 0 {
+			bits = 0b10
+			s.client2.disableWallets()
+		}
+		return nil
+	}
+
+	s.client1.filteredConn.requestFilter.Store(preFilter1)
+	s.client2.filteredConn.requestFilter.Store(preFilter2)
+
 	return s.simpleTradeTest(qty, rate, order.TakerSwapCast)
 }
 
@@ -461,93 +494,42 @@ func testMakerGhostingAfterTakerRedeem(s *simulationTest) error {
 	var qty, rate uint64 = 1 * s.lotSize, 250 * s.rateStep
 	s.client1.isSeller, s.client2.isSeller = true, false
 
-	c1OrderID, c2OrderID, err := s.placeTestOrders(qty, rate)
-	if err != nil {
-		return err
-	}
-
-	// Monitor trades and stop at order.TakerSwapCast
-	monitorTrades, ctx := errgroup.WithContext(context.Background())
-	monitorTrades.Go(func() error {
-		return s.monitorOrderMatchingAndTradeNeg(ctx, s.client1, c1OrderID, order.TakerSwapCast)
-	})
-	monitorTrades.Go(func() error {
-		return s.monitorOrderMatchingAndTradeNeg(ctx, s.client2, c2OrderID, order.TakerSwapCast)
-	})
-	if err = monitorTrades.Wait(); err != nil {
-		return err
-	}
-
-	// Resume trades but disable Maker's ability to notify the server
-	// after redeeming Taker's swap.
-	resumeTrade := func(ctx context.Context, client *simulationClient, orderID string) error {
-		tracker, err := client.findOrder(orderID)
-		if err != nil {
-			return err
-		}
-		finalStatus := order.MatchComplete
-		var disconnectClient bool
-		tracker.mtx.Lock()
-		for _, match := range tracker.matches {
-			side, status := match.Side, match.Status
-			client.log.Infof("trade %s paused at %s", token(match.MatchID[:]), status)
-			if side == order.Maker {
-				// Disconnecting the client will lock the
-				// tracker mutex, and so cannot be done with it already locked.
-				disconnectClient = true
-			} else {
-				client.log.Infof("%s: resuming trade negotiations to audit Maker's redeem", side)
-				client.noRedeemWait = true
+	var bits uint8
+	anErr := errors.New("intentional error from test")
+	// Prevent the first redeemer, who must be maker, from sending
+	// redeem info to the server.
+	preFilter1 := func(route string) error {
+		if route == msgjson.RedeemRoute {
+			if bits == 0 {
+				bits = 0b1
 			}
-			// Resume maker to redeem even though the redeem request to server
-			// will fail (disconnected) after the redeem bcast.
-			match.swapErr = nil
+			if bits&0b1 != 0 {
+				return anErr
+			}
 		}
-		tracker.mtx.Unlock()
-		if disconnectClient {
-			client.log.Infof("Maker: disconnecting DEX before redeeming Taker's swap")
-			client.dc().connMaster.Disconnect()
-			finalStatus = order.MakerRedeemed // maker shouldn't get past this state
-		}
-		// force next action since trade.tick() will not be called for disconnected dcs.
-		if _, err = client.core.tick(tracker); err != nil {
-			client.log.Infof("tick failure: %v", err)
-		}
-
-		// Propagation to miners can take some time after the send RPC
-		// completes, especially with SPV wallets, so wait a bit before mining
-		// blocks in monitorTrackedTrade.
-		time.Sleep(sleepFactor * time.Second)
-
-		return s.monitorTrackedTrade(client, tracker, finalStatus)
+		return nil
 	}
-	resumeTrades, ctx := errgroup.WithContext(context.Background())
-	resumeTrades.Go(func() error {
-		return resumeTrade(ctx, s.client1, c1OrderID)
-	})
-	resumeTrades.Go(func() error {
-		return resumeTrade(ctx, s.client2, c2OrderID)
-	})
-	if err = resumeTrades.Wait(); err != nil {
-		return err
+	preFilter2 := func(route string) error {
+		if route == msgjson.RedeemRoute {
+			if bits == 0 {
+				bits = 0b10
+			}
+			if bits&0b10 != 0 {
+				return anErr
+			}
+		}
+		return nil
 	}
 
-	// Allow some time for balance changes to be properly reported.
-	// There is usually a split-second window where a locked output
-	// has been spent but the spending tx is still in mempool. This
-	// will cause the txout to be included in the wallets locked
-	// balance, causing a higher than actual balance report.
-	s.waitALittleBit()
+	s.client1.filteredConn.requestFilter.Store(preFilter1)
+	s.client2.filteredConn.requestFilter.Store(preFilter2)
+	defer s.client1.filteredConn.requestFilter.Store(func(string) error { return nil })
+	defer s.client2.filteredConn.requestFilter.Store(func(string) error { return nil })
 
-	for _, client := range s.clients {
-		if err = s.assertBalanceChanges(client); err != nil {
-			return err
-		}
-	}
-
-	s.log.Infof("Trades completed. Maker went dark at %s, Taker continued till %s.",
-		order.MakerRedeemed, order.MatchComplete)
-	return nil
+	// TODO: Currently taker redeemer will not have its redeem confirmed
+	// when match is server revoked. That should be changed and this
+	// changed to order.MatchConfirmed.
+	return s.simpleTradeTest(qty, rate, order.MatchComplete)
 }
 
 // TestOrderStatusReconciliation simulates a few conditions that could cause a
@@ -824,6 +806,7 @@ func testOrderStatusReconciliation(s *simulationTest) error {
 		return fmt.Errorf("client 2 dex not disconnected after %v", disconnectTimeout)
 	}
 
+	s.client2.enableWallets()
 	// Disconnect the wallets, they'll be reconnected when Login is called below.
 	// Login->connectWallets will error for btc spv wallets if the wallet is not
 	// first disconnected.
@@ -877,6 +860,19 @@ func testOrderStatusReconciliation(s *simulationTest) error {
 		return fmt.Errorf("client 2 locked funds not returned: locked before trading %v, locked after trading %v, "+
 			"locked after reconnect %v", preTradeLockedBalance, preDisconnectLockedBalance, c2Balance.Locked)
 	}
+	s.client2.enableWallets()
+
+	for _, c := range s.clients {
+		if err := c.mineMedian(context.TODO(), s.quote.id); err != nil {
+			return err
+		}
+		if err := c.mineMedian(context.TODO(), s.base.id); err != nil {
+			return err
+		}
+	}
+
+	s.waitALittleBit()
+
 	return nil
 }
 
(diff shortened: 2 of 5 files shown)


chappjc:
> Taker will skip confirming the redemption if they do not have server's ack of their redeem request, as stated here:

Yeah, looks like revoked matches slip through the cracks w.r.t. redeem confirmation.  Good catch.

Thanks for fixing the lock time issue I created too.

client/core/simnet_trade.go
```diff
@@ -90,7 +91,7 @@ type assetConfig struct {
 	isToken          bool
 }
 
-var testLookup = map[string]func(s *simulationTest) error{
+var TestLookup = map[string]func(s *simulationTest) error{
```
chappjc, line 94: This is a little awkward since it's exporting a func that takes an unexported type (`simulationTest`), but this is all simnet bidness, so I don't much care.  It's just unfortunate to have this kind of stuff end up in `core`'s public API.

JoeGruffins: Exporting a getter now if that's better?

client/core/simnet_trade.go
```diff
@@ -1193,91 +1172,98 @@ func (s *simulationTest) monitorTrackedTrade(client *simulationClient, tracker *
...
 	// run a repeated check for match status changes to mine blocks as necessary.
 	maxTradeDuration := 2 * time.Minute
 
 	tryUntil(s.ctx, maxTradeDuration, func() bool {
 		var completedTrades int
+		mineAssets := make(map[uint32]uint32)
+		// Don't spam.
+		time.Sleep(time.Second * 2 * sleepFactor)
 		tracker.mtx.Lock()
-		defer tracker.mtx.Unlock()
```
JoeGruffins, line 1203: Sorry for the delay, holding this lock too much was causing odd failures and took a while to figure it out. Trades can't continue if you hold it forever, obviously.

JoeGruffins:
Still need to be able to choose maker/taker... Or run once with each on each asset.

client/core/simnet_trade.go
```diff
@@ -146,20 +154,7 @@ func (s *simulationTest) waitALittleBit() {
...
-		for bipID, _ := range s.client1.wallets {
-			if accountBIPs[bipID] {
-				sleep = feeSleep
-				s.client1.log.Infof("Waiting for paid order fees to be populated (%s).", sleep)
-				break
-			}
-		}
-	}
-	time.Sleep(sleep)
+	time.Sleep(sleep * sleepFactor)
```
JoeGruffins, line 157: We don't need to wait really for this, just mine 2 blocks for inits, so this is moved.

client/core/simnet_trade.go
```diff
@@ -403,6 +398,10 @@ func (s *simulationTest) setup(cl1, cl2 *SimClient) (err error) {
 	if err != nil {
 		return fmt.Errorf("error starting clients: %w", err)
 	}
+	s.client1.replaceConns()
+	s.client1.fakeConns.open()
+	s.client2.replaceConns()
+	s.client2.fakeConns.open()
```
JoeGruffins, line 404: Now abusing conns rather than altering match info to cause request errors.

client/core/simnet_trade.go
```diff
@@ -876,154 +838,173 @@ func testOrderStatusReconciliation(s *simulationTest) error {
...
+	s.client2.enableWallets()
+
+	for _, c := range s.clients {
+		if err := c.mineMedian(context.TODO(), s.quote.id); err != nil {
+			return err
+		}
+		if err := c.mineMedian(context.TODO(), s.base.id); err != nil {
+			return err
+		}
+	}
```
JoeGruffins, line 850: Sometimes the clients will send inits, so this allows the next test to not have a refund in the middle. Could probably be more specific but seems ok like this...

client/core/simnet_trade.go
```diff
@@ -1193,91 +1172,98 @@ func (s *simulationTest) monitorTrackedTrade(client *simulationClient, tracker *
...
+		if match.Side == order.Maker {
+			if finalStatus >= order.MakerSwapCast {
+				recordBalanceChanges(true, match.Quantity, match.Rate)
+			}
+			if finalStatus >= order.MakerRedeemed {
+				recordBalanceChanges(false, match.Quantity, match.Rate)
+			}
+		}
+	}
+	return nil
```
JoeGruffins, line 1193: This fixes the btc_dcr balance change issue. Before it was added as the swap progressed, but sometimes it would miss a match status change. We already know what balances to expect before the swap progresses so add it up at the beginning.

client/core/simnet_trade.go
```diff
@@ -1193,91 +1172,98 @@ func (s *simulationTest) monitorTrackedTrade(client *simulationClient, tracker *
...
-				// instantaneos and other times the server's node takes a while.
-				// to see the txns (TODO: diagnose this).
-				// Tests that don't expect a redeem can set noRedeemWait to true.
-				if finalStatus == order.MatchComplete &&
-					len(match.MetaData.Proof.Auth.RedeemSig) == 0 &&
-					!client.noRedeemWait {
-					client.log.Infof("Completed match waiting for async redeem response...")
-					continue // check again later
-				}
-				// Prevent further action by blocking the match with a swapErr.
```
JoeGruffins, line 1218: order.MatchConfirmed now signifies the Redeem completed, so this check is not necessary.

client/core/simnet_trade.go
```diff
@@ -1331,7 +1336,8 @@ func (s *simulationTest) checkAndWaitForRefunds(ctx context.Context, client *sim
 		if !client.isSeller {
 			swapAmt = calc.BaseToQuote(match.Rate, match.Quantity)
 		}
-		refundAmts[tracker.wallets.fromAsset.ID] += int64(swapAmt)
+		assetID := tracker.wallets.fromAsset.ID
+		refundAmts[assetID] += int64(swapAmt)
```
JoeGruffins, line 1340: unintended change

client/core/simnet_trade.go
```diff
@@ -1985,7 +1977,7 @@ func (s *simulationTest) assertBalanceChanges(client *simulationClient) error {
 	}
 
 	var baseFees, quoteFees int64
-	if fees := ord.FeesPaid; fees != nil {
+	if fees := ord.FeesPaid; fees != nil && !isRefund {
```
JoeGruffins, line 1980: This caused account tests to fail and was incorrect for utxo refunds as well, but the amount was not enough to matter for them.

client/core/simnet_trade.go
```diff
@@ -2059,17 +2059,33 @@ func (client *simulationClient) findOrder(orderID string) (*trackedTrade, error)
 func (client *simulationClient) disableWallets() {
```
JoeGruffins, line 2059: I wanted to get rid of this, but it looks needed for some tests like `testNoMakerSwap`

JoeGruffins:
All zec tests fail, possibly related to #1768

JoeGruffins:
Most tests are run with each client as maker/taker now, so twice by default. This ensures that each asset has the same tests applied, rather than randomly choosing the maker/taker.

client/cmd/simnet-trade-tests/run
```diff
@@ -86,7 +86,7 @@ dcrdextt - Decred RPC wallet and Ethereum wallet with dextt.eth test token on ma
 
 ---------------------
 
-One or more of the following tests can be run by specifying one or more -t flags. Default is ['success'].
+One or more of the following tests can be run by specifying one or more -t flags. Default is ['success']. All tests can be run win --all.
```
chappjc, line 89: win -> with

I like the `except` switch you added.  Will you add a short sentence about that too?

client/core/simnet_trade.go
```diff
@@ -877,154 +849,173 @@ func testOrderStatusReconciliation(s *simulationTest) error {
...
-	// Monitor trades and stop at order.MakerSwapCast.
-	monitorTrades, ctx := errgroup.WithContext(context.Background())
-	monitorTrades.Go(func() error {
-		return s.monitorOrderMatchingAndTradeNeg(ctx, s.client1, c1OrderID, order.MakerSwapCast)
-	})
-	monitorTrades.Go(func() error {
-		return s.monitorOrderMatchingAndTradeNeg(ctx, s.client2, c2OrderID, order.MakerSwapCast)
-	})
-	if err = monitorTrades.Wait(); err != nil {
+	return tc.realConn.Request(msg, respHandler)
```
chappjc, line 894: The MITM approach is great.   I think we should have done this from the start.  :hearts:

dex/testing/zec/harness.sh
```diff
@@ -33,10 +33,10 @@ mkdir -p "${HARNESS_DIR}"
...
 WALLET_PASSWORD="abc"
 
-ALPHA_CLI_CFG="-rpcport=${ALPHA_RPC_PORT} -regtest=1 -rpcuser=user -rpcpassword=pass"
-BETA_CLI_CFG="-rpcport=${BETA_RPC_PORT} -regtest=1 -rpcuser=user -rpcpassword=pass"
-DELTA_CLI_CFG="-rpcport=${DELTA_RPC_PORT} -regtest=1 -rpcuser=user -rpcpassword=pass"
-GAMMA_CLI_CFG="-rpcport=${GAMMA_RPC_PORT} -regtest=1 -rpcuser=user -rpcpassword=pass"
+ALPHA_CLI_CFG="-rpcport=${ALPHA_RPC_PORT} -regtest=1 -rpcuser=user -rpcpassword=pass -conf=${ALPHA_DIR}/alpha.conf"
+BETA_CLI_CFG="-rpcport=${BETA_RPC_PORT} -regtest=1 -rpcuser=user -rpcpassword=pass -conf=${BETA_DIR}/beta.conf"
+DELTA_CLI_CFG="-rpcport=${DELTA_RPC_PORT} -regtest=1 -rpcuser=user -rpcpassword=pass -conf=${DELTA_DIR}/delta.conf"
+GAMMA_CLI_CFG="-rpcport=${GAMMA_RPC_PORT} -regtest=1 -rpcuser=user -rpcpassword=pass -conf=${GAMMA_DIR}/gamma.conf"
```
chappjc, line 39: What is missing from zcash-cli command that requires the `-conf`?  They work without this for me, although I'm still testing with zcashd 5.0.0 and 5.3.0 is out now...

client/core/simnet_trade.go
chappjc: So TODO? The change is fine to me unless you meant it shouldn't do this increment?

client/core/simnet_trade.go
```diff
@@ -1998,6 +2054,16 @@ func (s *simulationTest) assertBalanceChanges(client *simulationClient) error {
 		}
 	}
 
+	// Account assets require a refund fee in addition to the swap amount.
+	if isRefund {
+		if ord.Sell && accountBIPs[s.base.id] {
+			baseFees = int64(dexeth.RefundGas(0 /*version*/)) * 2
```
chappjc, line 2060: So here we have the fee rate as 2 gwei/gas.  Is that generally expected or dictated for eth simnet? Can we doc?

Also, I think we do want to account for refund fees for UTXO assets too.  Perhaps we can just approximate the refund fee as the same as the swap fee for UTXO assets, given the rather large [0.95, 1.05] buffer we add below.

Or do we need to this info to the `FeeBreakdown`?

client/core/simnet_trade.go
chappjc: If it's refunded, `fees.Swap` was paid though (plus refund fees for the same asset), just not `fees.Redemption`.

client/core/simnet_trade.go
```diff
@@ -1151,35 +1211,18 @@ func (s *simulationTest) monitorOrderMatchingAndTradeNeg(ctx context.Context, cl
...
 
-	runMiner(s.base, 1)
-	runMiner(s.quote, 1)
-
-	return cancel
+	return s.monitorTrackedTrade(client, tracker, finalStatus)
 }
 
-func (s *simulationTest) monitorTrackedTrade(client *simulationClient, tracker *trackedTrade, finalStatus order.MatchStatus) error {
+func (s *simulationTest) calcBalanceChanges(client *simulationClient, orderID string, finalStatus order.MatchStatus) error {
```
chappjc, line 1221: `calcBalanceChanges` is only used from within `monitorOrderMatchingAndTradeNeg`, where we've already done `findOrder`, and do a `range tracker.matches` loop with proper `tracker.mtx` use.

I think `monitorOrderMatchingAndTradeNeg` should look like:

```go
func (s *simulationTest) monitorOrderMatchingAndTradeNeg(ctx context.Context, client *simulationClient, orderID string, finalStatus order.MatchStatus) error {
	errs := newErrorSet("[client %s] ", client.name)

	tracker, err := client.findOrder(orderID)
	if err != nil {
		return errs.addErr(err)
	}

	// Wait up to 2 times the epoch duration for this order to get matched.
	maxMatchDuration := 2 * time.Duration(tracker.epochLen) * time.Millisecond
	client.log.Infof("Waiting up to %v for matches on order %s", maxMatchDuration, tracker.token())
	matched := client.notes.find(ctx, maxMatchDuration, func(n Notification) bool {
		orderNote, isOrderNote := n.(*OrderNote)
		isMatchedTopic := n.Topic() == TopicBuyMatchesMade || n.Topic() == TopicSellMatchesMade
		return isOrderNote && isMatchedTopic && orderNote.Order.ID.String() == orderID
	})
	if ctx.Err() != nil { // context canceled
		return nil
	}
	if !matched {
		return errs.add("order %s not matched after %s", tracker.token(), maxMatchDuration)
	}

	recordBalanceChanges := func(isSwap bool, qty, rate uint64) {
		amt := int64(qty)
		a := s.base
		if client.isSeller != isSwap {
			// use quote amt for seller redeem and buyer swap
			amt = int64(calc.BaseToQuote(rate, qty))
			a = s.quote
		}
		if isSwap {
			amt *= -1
		}
		client.log.Infof("updated %s balance diff with %s", a.symbol, a.valFmt(amt))
		client.expectBalanceDiffs[a.id] += amt
	}

	tracker.mtx.RLock()
	client.log.Infof("%d match(es) received for order %s", len(tracker.matches), tracker.token())
		client.log.Infof("%s on match %s, amount %s %s", match.Side.String(), token(match.MatchID.Bytes()),
			s.base.valFmt(match.Quantity), s.base.symbol)
		if match.Side == order.Taker {
			if finalStatus >= order.TakerSwapCast {
				recordBalanceChanges(true, match.Quantity, match.Rate)
			}
			if finalStatus >= order.MatchComplete {
				recordBalanceChanges(false, match.Quantity, match.Rate)
			}
		}
			if finalStatus >= order.MakerSwapCast {
				recordBalanceChanges(true, match.Quantity, match.Rate)
			}
			if finalStatus >= order.MakerRedeemed {
				recordBalanceChanges(false, match.Quantity, match.Rate)
			}
		}
	}
	tracker.mtx.RUnlock()

	return s.monitorTrackedTrade(client, tracker, finalStatus)
}
```

chappjc (changes requested):

JoeGruffins:
just rebased

client/core/simnet_trade.go
JoeGruffins: I just mean about declaring the assetID variable when it is only used once.

client/core/simnet_trade.go
JoeGruffins: The expected balances are checked after the swap and then after the refunds, so `fees.Swap` was already accounted for. I think.

client/cmd/simnet-trade-tests/run
JoeGruffins: Corrected and added.

dex/testing/zec/harness.sh
JoeGruffins: I'm on 5.2.0. On master I see this when I attempt to run the zec harness:
```
+ sleep 1
+ tmux send-keys -t zec-harness:4 './alpha z_importwallet /home/joe/git/dcrdex/dex/testing/zec/alphawallet ; tmux wait-for -S zec' 'C-m;' wait-for zec
+ tmux send-keys -t zec-harness:4 './beta z_importwallet /home/joe/git/dcrdex/dex/testing/zec/betawallet ; tmux wait-for -S zec' 'C-m;' wait-for zec
+ echo 'Generating 400 blocks for alpha'
Generating 400 blocks for alpha
+ tmux send-keys -t zec-harness:4 './alpha generate 400; tmux wait-for -S zec' 'C-m;' wait-for zec
+ ALPHA_ADDR=tmEgW8c44RQQfft9FHXnqGp8XEcQQSRcUXD
+ BETA_ADDR=tmSog4freWuq1aC13yf1996fy4qXPmv3GTB
++ ./delta getnewaddress
Error reading configuration file: Missing zcash.conf
+ DELTA_ADDR=
```

JoeGruffins: Updating to 5.3.0 If thats out anyway.

client/core/simnet_trade.go
```diff
@@ -101,6 +101,14 @@ var testLookup = map[string]func(s *simulationTest) error{
 	"resendpending": testResendPendingRequests,
 }
 
+func SimnetTestNames() []string {
```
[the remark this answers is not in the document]
JoeGruffins: Oh... didn't see that 😂

Removing `SimnetTestNames`

client/core/simnet_trade.go
```diff
@@ -1109,14 +1115,68 @@ func (s *simulationTest) placeTestOrders(qty, rate uint64) (string, string, erro
...
+		client = s.client2
 	}
-	c2OrderID, err := s.placeOrder(s.client2, qty, rate, false)
+
+	tracker, err := client.findOrder(orderID)
 	if err != nil {
-		return "", "", fmt.Errorf("client2 place %s order error: %v", sellString(s.client2.isSeller), err)
+		return "", "", err
+	}
+	// Wait the epoch duration for this match to get booked.
```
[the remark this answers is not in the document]
JoeGruffins: Corrected.

client/core/simnet_trade.go
```diff
@@ -1066,6 +1055,23 @@ func (s *simulationTest) simpleTradeTest(qty, rate uint64, finalStatus order.Mat
 		return err
 	}
 
+	// Some coins will need confirmations on the init to calculate swap
```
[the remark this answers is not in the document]
JoeGruffins: Removed.

client/core/simnet_trade.go
```diff
@@ -877,154 +849,173 @@ func testOrderStatusReconciliation(s *simulationTest) error {
...
-	resumeTrades.Go(func() error {
-		return resumeTrade(ctx, s.client1, c1OrderID)
-	})
-	resumeTrades.Go(func() error {
-		return resumeTrade(ctx, s.client2, c2OrderID)
-	})
-	if err = resumeTrades.Wait(); err != nil {
-		return err
-	}
+func (sc *simulationClient) replaceConns() {
```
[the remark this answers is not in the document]
JoeGruffins: Moved and named `client`

client/core/simnet_trade.go
```diff
@@ -1761,6 +1816,7 @@ type simulationClient struct {
 	// change validation. Set to nil to NOT perform balance checks.
 	expectBalanceDiffs map[uint32]int64
 	lastOrder          []byte
+	fakeConns          *tConn
```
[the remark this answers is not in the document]
JoeGruffins: Removed.

client/core/simnet_trade.go
JoeGruffins: > So here we have the fee rate as 2 gwei/gas. Is that generally expected or dictated for eth simnet?

It's just expected from simnet. It is the tip cap. The, whats it called, consensus enforced fee is much smaller. I think I've seen that fee rise above a few gwei after running loadbot for a while though.

> Or do we need to add this info to the FeeBreakdown?

Maybe.

JoeGruffins: Added utxo refunds, based off of the btc price of a refund.

client/core/simnet_trade.go
```diff
@@ -877,154 +849,173 @@ func testOrderStatusReconciliation(s *simulationTest) error {
...
 
-	c1OrderID, c2OrderID, err := s.placeTestOrders(qty, rate)
+type tConn struct {
+	routeCh      chan string
+	errCh        chan error
+	on           uint32
+	closeAndWait func()
+
+	realConn comms.WsConn
+}
```
buck54321, line 877: I was toying with a different way.

https://github.com/buck54321/dcrdex/compare/a7ddea84e2ac6593270a496582474a765f675fde...buck54321:dcrdex:jg-simnet-test-simplification

I think I got the `bits` part wrong? I haven't quite wrapped my head around the `bits` yet, but the approach seems simpler.

chappjc: That's not bad.  Not sure about bits either.
Would there be a data race on these assignments, or are comms at this points?

```go
	s.client1.filteredConn.requestFilter = preFilter
	s.client2.filteredConn.requestFilter = preFilter

	defer func() {
		s.client1.filteredConn.requestFilter = nil
		s.client2.filteredConn.requestFilter = nil
	}()
	}()
```

buck54321: I think you're right. I pushed an update.

client/core/simnet_trade.go
```diff
@@ -461,93 +470,55 @@ func testMakerGhostingAfterTakerRedeem(s *simulationTest) error {
...
+						continue
+					}
+				}
+				c1FakeConns.errCh <- nil
+			case route := <-c2FakeConns.routeCh:
+				if route == msgjson.RedeemRoute {
+					if bits == 0 {
+						bits |= 0x01
+					}
+					if bits&0x01 != 0 {
```
chappjc, line 506: I'm not quite getting the intent of the bits.  You have `0x1` in the c1 case, and `0x01` in the c2 case, but those are the same number (same bit).  Did you want it to be `0x2` down here.  Is it needed to distinguish c1 from c2 in any test?  Could this be done with one (or two) bools?

client/core/simnet_trade.go
chappjc: Seems like feasibility of this alt approach depends on intent of bits.  Commented here: https://github.com/decred/dcrdex/pull/1909#discussion_r1004991954

client/core/simnet_trade.go
JoeGruffins: Oh, shoot. I meant to be using the `0b` prefix. And it would progress 0b1 -> 0b10 -> 0b100 but I'll change to what buck suggests since I was also confusing it.

JoeGruffins:
https://github.com/decred/dcrdex/compare/a7ddea84e2ac6593270a496582474a765f675fde..1942e61260a268dee172a24b4c6a1d56983bf3fd

It's difficult to balance which side is mined when the mining is asynchronous. Also, geth light client is very slow to update its best block. So, if taker is selling eth, the match is always confirmed before the taker's init can get enough confs. It has enough confs, but geth will not see that for a while. I guess it is polling on an interval for new blocks. So, added more logic to wait and not mine on one side if eth should be searching for an init on the other.

Also disabling the wallets can just be the new 'setDisabled' which is very useful.

The bits are now in line with the vision, but I will still change to buck's suggestion.

nomakerswap randomly fails but I think it will be fine once the TODO is fixed and all matches that make it to MakerRedeemed also make it to SwapConfirmed.

buck54321:
> The bits are now in line with the vision, but I will still change to buck's suggestion.

I don't think mine was working right. Was hoping to get your input.

JoeGruffins:
> OK we'll move forward with the current fakeConns approach and follow up with @buck54321's adaptation after the bits business is sorted out.

Oh sorry, did the buck thing https://github.com/decred/dcrdex/compare/1942e61260a268dee172a24b4c6a1d56983bf3fd..1bf948917c28e140aefd83209e695bf5800e6a81

chappjc:
> Oh sorry, did the buck thing

That's great!  I just thought you were out for the day

JoeGruffins:
Small fix https://github.com/decred/dcrdex/compare/1bf948917c28e140aefd83209e695bf5800e6a81..3834128531c45724b67d44d557820f1481ec6dee

I'm still not sure this is good enough for eth. Not sure if this is the best way or if it's possible to sleep the right amount every time like this. Probably could use smarter mining but I shouldn't spend any more time on it.

chappjc:
So I ran `./run dcrbtc -t all`, forgetting that it was supposed to be `./run dcrbtc -all` and it of course bombed with `ERROR: no test named "all"`.
- should probably have main.go interpret `all` as a test in addition to the `-all` flag
- it had to actually start the harnesses **and register on the servers!** before realizing "all" isn't a know test.  Need to screen out earlier.

```
2022-10-26 11:19:11.270 [INF] CORE:2: notify: |SUCCESS| (feepayment) Account registered - You may now trade at 127.0.0.1:17273
2022-10-26 11:19:11.353 [INF] T[2]: Fee payment confirmed
2022-10-26 11:19:11.353 [INF] T: Running "all" with cliet 1 as maker
ERROR: no test named "all"
```

chappjc:
Hmm, running `dcrbtc -all`, it failed `makerghost`.  I'm sure it worked yesterday

```
2022-10-26 11:34:26.698 [WRN] CORE:2: notify: |WARNING| (order) Match revoked - Match ff79639c has been revoked - Order: 82819b9017f3d404b7777887b8fb16480029ea2c19af7cf069b56eb21bd6bdef
2022-10-26 11:34:26.698 [WRN] CORE:2: Match ff79639cedac2da3aaede9f58b7de950a4b4b86d1522e3d41a3dfd023f2f3da3 revoked in status TakerSwapCast for order 82819b9017f3d404b7777887b8fb16480029ea2c19af7cf069b56eb21bd6bdef
2022-10-26 11:34:27.551 [INF] T[2]: Mining 1 btc blocks for Maker's swap, match ff79639c
2022-10-26 11:34:27.732 [INF] CORE:2: Refunding btc contract cbb7c34bb2a35713c43799966eeaf96e544d5f6b1ecd91d2fcb6c8b4f7b6d1b4:0 for match ff79639cedac2da3aaede9f58b7de950a4b4b86d1522e3d41a3dfd023f2f3da3 (no valid redemption received from Maker)
2022-10-26 11:34:27.733 [WRN] CORE:2: notify: |WARNING| (order) Matches Refunded - Refunded 0.00000000 BTC on order 82819b90 - Order: 82819b9017f3d404b7777887b8fb16480029ea2c19af7cf069b56eb21bd6bdef
2022-10-26 11:34:27.734 [INF] CORE:2[btc]: FindRedemption - Checking block 41ef0e7072cbb693593cb46f04af60e66b41b5713a107ae177cd164d9b9f564f for swap cbb7c34bb2a35713c43799966eeaf96e544d5f6b1ecd91d2fcb6c8b4f7b6d1b4:0
2022-10-26 11:34:27.738 [INF] CORE:2: Found redemption of contract cbb7c34bb2a35713c43799966eeaf96e544d5f6b1ecd91d2fcb6c8b4f7b6d1b4:0 (btc) for order 82819b9017f3d404b7777887b8fb16480029ea2c19af7cf069b56eb21bd6bdef, match ff79639cedac2da3aaede9f58b7de950a4b4b86d1522e3d41a3dfd023f2f3da3. Redeem: eee7ada034224949ebbdbabd933cea79919bcb2fad89610963a4595914d7ac19:0
2022-10-26 11:34:29.561 [INF] T[2]: NOW =====> MakerRedeemed
2022-10-26 11:34:31.562 [INF] T[2]: trade for order 82819b90, match ff79639c monitored successfully till MakerRedeemed, side Taker
2022-10-26 11:34:35.563 [INF] T[1]: updating balances
2022-10-26 11:34:35.565 [INF] T[1]: dcr available 1430.99961330, immature 0.00000000, locked 0.00000000
2022-10-26 11:34:35.566 [INF] T[1]: btc available 1360.01449285, immature 0.00000000, locked 0.00000000
2022-10-26 11:34:35.566 [INF] T[1]: dcr balance change -10.00002500 is in expected range of -10.50002625 - -9.50002375
2022-10-26 11:34:35.566 [INF] T[1]: btc balance change 0.00249857 is in expected range of 0.00237364 - 0.00262349
2022-10-26 11:34:35.566 [INF] T[2]: updating balances
2022-10-26 11:34:35.567 [INF] T[2]: dcr available 1600.99978450, immature 0.00000000, locked 0.00000000
2022-10-26 11:34:35.568 [INF] T[2]: btc available 443.98548371, immature 0.00000000, locked 0.00000000
ERROR: test "makerghost" failed with client 1 as maker: dcr balance change not in expected range 9.50000000 - 10.50000000, got 0.00000000
```

Will run again.

JoeGruffins:
https://github.com/decred/dcrdex/compare/3834128531c45724b67d44d557820f1481ec6dee..fa8a129a6eab01a6ebfaf65c8b20851e2eb43fbc

Fixes orderstatus

> Hmm, running dcrbtc -all, it failed makerghost. I'm sure it worked yesterday

I had the bits mixed up yesterday. It seems that taker will only ever get to "order.MakerRedeemed" so it's difficult to monitor it to the end without making more special cases. It should work fine after https://github.com/decred/dcrdex/issues/1924 is solved

JoeGruffins:
oh shoot, added per your last commit but was too slow, will make another pr

PR #1910, opened 2022-10-18
client: remove duplicate translations

ukane-philemon:
This is a follow-up for #1902


client/webserver/locales/zh-cn.go
```diff
@@ -248,16 +242,13 @@ var ZhCN = map[string]string{
 	"wallet_wait_synced":          "同步后会创建钱包",
 	"Create a Wallet":             "创建钱包",
 	"Receive":                     "接收",
-	"Locked":                      "锁定",
-	"Ready":                       "已准备",
-	"Off":                         "关",
+	"ready":                       "已准备",
```
chappjc, line 245: Unfortunately, the `cases.Caser` does not work for "zh"
https://go.dev/play/p/VMH6czbu43r
Does not go from "已准备" to "准备就绪".  I have no clue why there's such a difference, or if it's correct, but some languages simply may not be so easy.

Maybe we need to adjust the code to first check the map for a key entry that actually has `:title:` in the key, as an indicator **not** to use the `Caser`.

ukane-philemon: > Unfortunately, the cases.Caser does not work for "zh"

Yes, about this. I noticed "准备就绪" was incorrectly tied to "ready" but it translates to "Ready" while ''已准备' retranslates to "ready" .

> Maybe we need to adjust the code to check the map for a key entry that actually has :title: in the key (the prefix is indeed part of the text token), as an indicator not to use the Caser.

Makes sense.

buck54321 (approved):

PR #1913, opened 2022-10-20
client: missing configOpt at input

vctt94:
This PR fixes an error when creating an external wallet.

If going to external wallet form the following error will be thrown: 

image

The reason is because the configOpt is not added into the input when the input html element `ConfigOptionInput` is created. This PR fixes it.

### DIFF
--- a/client/webserver/site/src/js/forms.ts
+++ b/client/webserver/site/src/js/forms.ts
@@ -700,7 +700,7 @@ export class WalletConfigForm {
     const inputs: Record<string, HTMLElement[]> = {}
     box.querySelectorAll('input').forEach((input: ConfigOptionInput) => {
       const k = input.dataset.configKey
-      if (!k) return // typescript
+      if (!k) return // TS2538
       const els = []
       for (const [opt, el] of this.configElements) if (opt.key === k) els.push(el)
       inputs[k] = els
--- a/client/webserver/site/src/js/forms.ts
+++ b/client/webserver/site/src/js/forms.ts
@@ -520,7 +520,7 @@ export class WalletConfigForm {
     } else el = this.textInputTmpl.cloneNode(true) as HTMLElement
     this.configElements.push([opt, el])
     const input = el.querySelector('input') as ConfigOptionInput
-    input.configOpt = opt
+    input.dataset.configKey = opt.key
     input.id = elID
     const label = Doc.safeSelector(el, 'label')
     label.htmlFor = elID // 'for' attribute, but 'for' is a keyword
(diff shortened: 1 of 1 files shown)


chappjc:
Affects 0.5 release?  Glancing at code, it seems to, haven't confirmed.

vctt94:
nope, looks like it was introduced at https://github.com/decred/dcrdex/commit/921f7f6a2c4e34229b733d5d6d1f0287d9613bc7#diff-3a4a16c018f4bf3f4ca85090174b8c06fddc8566445458a01a0edaa5c5f97681R508

chappjc:
Alright, good.  Didn't want to backport. I don't see the JS error on the release branch now that I test.

chappjc (approved):
I see the error routinely.  Resolution looks good.  Leaving up for now, for other eyes.

client/webserver/site/src/js/forms.ts
```diff
@@ -520,6 +520,7 @@ export class WalletConfigForm {
     } else el = this.textInputTmpl.cloneNode(true) as HTMLElement
     this.configElements.push([opt, el])
     const input = el.querySelector('input') as ConfigOptionInput
+    input.configOpt = opt
```
buck54321, line 523: If the `opt.key` is the only thiing we need, can we store just the key as `input.dataset.configKey` instead?

client/webserver/site/src/js/forms.ts
```diff
@@ -698,7 +699,8 @@ export class WalletConfigForm {
   reorder (box: HTMLElement) {
     const inputs: Record<string, HTMLElement[]> = {}
     box.querySelectorAll('input').forEach((input: ConfigOptionInput) => {
-      const k = input.configOpt.key
+      const k = input.dataset.configKey
+      if (!k) return // typescript
```
chappjc, line 703: What does this comment indicate?

vctt94: That I am doing this check because of typescript, can remove if you'd like

chappjc: Docs are always good, I'm just still confused.  You mean this is something typescript-specific and not javascript?  
What does it indicate if `!k`?

vctt94: that's correct, it is typescript-specific and if not added it will throw the following error:

```
src/js/forms.ts:705:14 - error TS2538: Type 'undefined' cannot be used as an index type.

705       inputs[k] = els
```

It happens because undefined is a valid value for `input.dataset.configkey`

vctt94: I can also be more specific at the comment

chappjc: When and why would it be undefined?

vctt94: It is a valid value, it does not mean it will be. It shouldn't, that's why the comment

chappjc: Ohhh, it's just `tsc` code analysis being obnoxiou.  It's not that typescript and javascript syntax are different, or even that babel will fail to transpile this to JavaScript without this check, but that we hit `TS2538`.

JoeGruffins:
Can confirm the error and that this fixes it.

client/webserver/site/src/js/forms.ts
chappjc: If you change this comment to `// TS2538` we can merge it.

chappjc (approved):

Issue #1912, opened 2022-10-19
client/core: intermittent TestWalletSyncing failure


chappjc:
I have been seeing this a lot recently

```
 --- FAIL: TestWalletSyncing (0.12s)
    core_test.go:9174: expected 9 progress notes, got 4
```

ukane-philemon:
Yeah, I have seen this once. When during one of #1887 CI test.

PR #1914, opened 2022-10-20
client/core: fix intermittent TestWalletSyncing failure

LasTshaMAN:
How about relaxing testing conditions a little bit so that there is no chance for intermittent failure due to reasons it happens in the example from issue description ?

Closes https://github.com/decred/dcrdex/issues/1912.

### DIFF
--- a/client/core/core_test.go
+++ b/client/core/core_test.go
@@ -37,6 +37,7 @@ import (
 	"decred.org/dcrdex/server/account"
 	serverdex "decred.org/dcrdex/server/dex"
 	"github.com/decred/dcrd/crypto/blake256"
+	"github.com/decred/dcrd/dcrec/secp256k1/v4"
 	"golang.org/x/text/language"
 	"golang.org/x/text/message"
 )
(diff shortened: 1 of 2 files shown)


client/core/simnet_trade.go
```diff
@@ -152,7 +152,7 @@ func (s *simulationTest) waitALittleBit() {
 	if sleep < feeSleep {
 		// eth based assets need to wait for txn to be mined for fees
 		// to be populated here.
-		for bipID, _ := range s.client1.wallets {
+		for bipID := range s.client1.wallets {
```
LasTshaMAN, line 155: that's just something `gofmt` has picked up

client/core/core_test.go
```diff
@@ -9168,12 +9168,12 @@ out:
...
 		}
 	}
-	// Should get 9 notes, but just make sure we get at least half of them to
-	// avoid github vm false positives.
-	if progressNotes < 5 /* 9? */ {
-		t.Fatalf("expected 9 progress notes, got %d", progressNotes)
+	// We must get at least 1 progress note.
+	// By the time we've got 9th note it should signal that the wallet has been
+	// synced (due to how we've set up testDuration and syncTickerPeriod values).
+	if progressNotes < 1 || progressNotes > 9 {
```
chappjc, line 9174: If we got here at all, that mean's we got at least one note (otherwise we'd hit the `t.Fatal` above on timeout), and the last note had `walletNote.Wallet.Synced = true` (otherwise no `break out`).

Since this test has degenerated completely now, we might as well remove this note count check entirely.

PR #1916, opened 2022-10-21
client: Add form to delete archived records

### LINKED ISSUE #1911: ui: form to archive orders
The `(c *Core) DeleteArchivedRecords(olderThan *time.Time, matchesFile, ordersFile string)` method could use an http api endpoint and a form accessible from the /orders page.  Three inputs: (1) time before which to purge data, (2) optional, purged matches csv file name, (3) optional, purged orders csv file name

ukane-philemon:
This adds a form to delete archived records. Closes #1911 

UI changes:
form
Screenshot 2022-10-29 at 12 50 07 AM

`/orders` page
Screenshot 2022-10-22 at 4 27 30 AM


client/webserver/api.go
```diff
@@ -1170,6 +1170,26 @@ func (s *WebServer) apiToggleRateSource(w http.ResponseWriter, r *http.Request)
...
+func (s *WebServer) apiDeleteArchivedRecords(w http.ResponseWriter, r *http.Request) {
+	form := new(deleteRecordsForm)
+	if !readPost(w, r, form) {
+		return
+	}
+
+	var olderThan time.Time
+	if form.OlderThanMs > 0 {
+		olderThan = time.UnixMilli(form.OlderThanMs)
+	}
```
ukane-philemon, line 1183: was thinking if we should reject an invalid date time here.

chappjc:
Initial take in the UI is that the added button suggests some relation to the checkboxes above, but it's actually unrelated.  Needs some kind of visual partition or header

ukane-philemon:
@chappjc, I'd be glad to get a review from you. Then I can make relevant changes.

chappjc:
Uh oh, with https://github.com/decred/dcrdex/pull/1910 merged, I'm seeing

image

Looks like we didn't test that one well enough.

client/webserver/locales/en-us.go
```diff
@@ -262,7 +262,7 @@ var EnUS = map[string]string{
 	"Volume":                      "Volume",
 	"No Recent Activity":          "No Recent Activity",
 	"Recent asset_name Activity":  "Recent <span data-asset-name=1></span> Activity",
-	"Other Actions":               "Other Actions",
+	"other_actions":               "other_actions",
```
chappjc, line 265: Looks like this just needs a tweak to `"Other Actions"`

client/webserver/locales/en-us.go
```diff
@@ -283,4 +283,11 @@ var EnUS = map[string]string{
 	"Disabled":                    "Disabled",
 	"txfee_not_available":         "Transaction fee currently unavailable",
 	"fiat_exchange_rate_msg":      "Sources may provide fiat exchange rates for different subsets of assets. You should select all sources that are acceptable to get fiat exchange rates for the most assets. Assets with fiat rates from multiple sources will use the average from all of those sources. Note: dcrdate provides fiat exchange rates for only BTC and DCR.",
+	"delete_archived_records":     "Delete Archived Records",
+	"date_time":                   "Date and Time",
+	"archived_date_time_tooltip":  "Archived orders and matches created before your specified and date and time will be deleted from the database. Note: If an invalid date and time is provided, all archived records till date will be deleted.",
+	"matches_file_path":           "Matches CSV File Path",
+	"matches_file_path_tooltip":   "Optional: Specify a file path to save deleted matches as a CSV file. File must not already exist.",
+	"orders_file_path":            "Orders CSV File Path",
```
chappjc, line 291: Because of how `(*Core).DeleteArchivedRecords` takes file paths and writes to the disk of the machine where dexc is running, I think we have to do it like this. But it's a notable difference from how the order data export works by serving the exported file over the web browser.

To make it work like that, we'd have to ask Core to write to a temp folder and then serve the written files somehow.  Not sure it's worth it though.  Most people will either be running dexc on the same machine as the browser, or have easy access to the machine.

ukane-philemon: > takes file paths and writes to the disk of the machine where dexc is running, I think we have to do it like this.

Yeah, if a user provides just a filename, Core will write to `pwd` i.e the `/dexc` folder. I added a check for that in `core.go` `createFile(fileName string)`.

chappjc: I'm not commenting about the cwd issue raised in matrix, but the fact that this can be a different machine.

ukane-philemon: > I'm not commenting about the cwd issue raised in matrix, but the fact that this can be a different machine.

Oh, okay. We could provide the `datadir` to Core so we can save files like these to `datadir` instead?

client/core/core.go
```diff
@@ -8146,19 +8155,26 @@ func deleteMatchFn(matchesFileStr string) (perMatchFn func(*db.MetaMatch, bool)
...
+func (c *Core) archivedRecordsDataDirectory() string {
+	return filepath.Join(filepath.Dir(c.cfg.DBPath), "archived-records")
+}
+
 // DeleteArchivedRecords deletes archived matches from the database. Optionally
-// set a time to delete records after and file paths to save deleted records as
+// set a time to delete older records and file paths to save deleted records as
 // comma separated values. If a nil *time.Time is provided, current time is used.
-func (c *Core) DeleteArchivedRecords(olderThan *time.Time, matchesFile, ordersFile string) error {
+func (c *Core) DeleteArchivedRecords(olderThan *time.Time, saveMatchesToFile, saveOrdersToFile bool) error {
```
buck54321, line 8167: re https://github.com/decred/dcrdex/pull/1916#discussion_r1005017938:

Recommend leaving this method alone. If you want different functionality, add a new method. I think for a first pass, we can just write to a system location like you've done, so I guess the new method is something like. 

```Go
func (c *Core) DeleteArchiveWithBackup(olderThan *time.Time, saveMatchesToFile, saveOrdersToFile bool) error {
	var matchesFile, ordersFile string
	if saveMatchesToFile {
		matchesFile = filepath.Join(c.archivedRecordsDataDirectory(), fmt.Sprintf("archived-matches-%d", time.Now().Unix()))
	}
	if saveOrdersToFile {
		ordersFile = filepath.Join(c.archivedRecordsDataDirectory(), fmt.Sprintf("archived-orders-%d", time.Now().Unix()))
	}
	return c.DeleteArchivedRecords(olderThan, matchesFile, ordersFile)
}
```

chappjc: That's an option.  In matrix I suggested also leaving the method alone, but instead detecting a `"."` as either file name as the indicator to use an automatic location in `datadir`.  A new method is OK too, but we can keep the interface smaller this way.

buck54321: > That's an option. In matrix I suggested also leaving the method alone, but instead detecting a `"."` as either file name as the indicator to use an automatic location in `datadir`. A new method is OK too, but we can keep the interface smaller this way.

That's fine too. Just don't wanna change the existing interface.

chappjc: I see we have a new `DeleteArchivedRecordsWithBackup` right now. Let's leave it unless other things come up in re-testing.

ukane-philemon: > I see we have a new DeleteArchivedRecordsWithBackup right now.

Yes, I think 99% of users might not be comfortable with providing a non-existent file-path and that cuts out the process of dealing with `.` ,`./`. The `cmdSummary` for existing RPC `deleteArchivedRecordsRoute` says `that file locations are from the perspective of dexc and not the caller` so it makes sense to use the `dexc` `datadir` without having to require it from the user who might not be on the machine running `dexc` atm.

server/admin/server.go
```diff
@@ -94,16 +94,10 @@ func UseLogger(logger slog.Logger) {
...
-// filesExists reports whether the named file or directory exists.
-func fileExists(name string) bool {
-	_, err := os.Stat(name)
-	return !os.IsNotExist(err)
-}
-
 // NewServer is the constructor for a new Server.
 func NewServer(cfg *SrvConfig) (*Server, error) {
 	// Find the key pair.
-	if !fileExists(cfg.Key) || !fileExists(cfg.Cert) {
```
chappjc, line 106: To be perfectly honest, it's this kind of thing that I would keep any day to avoid further tightening the coupling between packages. `dex` is about the lowest level, so OK, but please do not assume it's always best to add an import just remove a few lines of boilerplate. OK to keep these changes, but next time save this for a PR to discuss.

ukane-philemon: Alright, I understand minor duplication > importing in some scenarios and will keep this in mind.

chappjc: Thanks. In this case it's really NBD, but in the very near future we are going to break the repo into at least two different modules, and that could get hairy.

chappjc (approved):
Looking good, but now that https://github.com/decred/dcrdex/pull/1923 is merged, would you please rebase and squash?  Thanks.

ukane-philemon:
> Looking good, but now that #1923 is merged, would you please rebase and squash? Thanks.

Done.

client/webserver/api.go
```diff
@@ -1170,6 +1170,26 @@ func (s *WebServer) apiToggleRateSource(w http.ResponseWriter, r *http.Request)
...
+
+	err := s.core.DeleteArchivedRecordsWithBackup(&olderThan, form.SaveMatchesToFile, form.SaveOrdersToFile)
```
chappjc, line 1185: When you don't set the time on the frontend, the payload is this: `{"olderThanMs":null,"saveMatchesToFile":false,"saveOrdersToFile":false}`

That's OK I suppose, but the issue is that in this handler code, we leave `olderThan` as the zero value for `time.Time`.  i.e. `0001-01-01 00:00:00 +0000 UTC`, which has a `UnixMilli()` value of `-62135596800000`.  Then in the `BoltDB` code, we do `olderThanB = uint64Bytes(uint64(olderThan.UnixMilli()))`, which first converts the negative `int64` to an `uint64` and then to bytes.  Not intended.

The docs for `(*Core) DeleteArchivedRecords` state: `If a nil *time.Time is provided, current time is used.`.  So, this `WebServer` handler should pass nil if `OlderThanMs` is not set.

So, let's use a `*time.Time` (pointer), and consider further if we want the form to be sumitted with `null` for `olderThanMs`.

chappjc: In addition to changing the http handler as described, a related change to the `BoltDB` code:

```diff
diff --git a/client/db/bolt/db.go b/client/db/bolt/db.go
index e23ad98b..6e8f9374 100644
--- a/client/db/bolt/db.go
+++ b/client/db/bolt/db.go
@@ -1867,7 +1867,7 @@ func (db *BoltDB) DeleteInactiveOrders(ctx context.Context, olderThan *time.Time
                finished   bool
                olderThanB []byte
        )
-       if olderThan != nil {
+       if olderThan != nil && !olderThan.IsZero() {
                olderThanB = uint64Bytes(uint64(olderThan.UnixMilli()))
        } else {
                olderThanB = uint64Bytes(timeNow())
@@ -2008,7 +2008,7 @@ func (db *BoltDB) DeleteInactiveMatches(ctx context.Context, olderThan *time.Tim
                finished   bool
                olderThanB []byte
        )
-       if olderThan != nil {
+       if olderThan != nil && !olderThan.IsZero() {
                olderThanB = uint64Bytes(uint64(olderThan.UnixMilli()))
        } else {
                olderThanB = uint64Bytes(timeNow())
```

ukane-philemon: Oh, shoot. `var olderThan time.Time` was supposed to be a pointer `var olderThan *time.Time`

client/webserver/locales/en-us.go
```diff
@@ -274,4 +274,11 @@ var EnUS = map[string]string{
 	"Disabled":                    "Disabled",
 	"txfee_not_available":         "Transaction fee currently unavailable",
 	"fiat_exchange_rate_msg":      "Sources may provide fiat exchange rates for different subsets of assets. You should select all sources that are acceptable to get fiat exchange rates for the most assets. Assets with fiat rates from multiple sources will use the average from all of those sources. Note: dcrdate provides fiat exchange rates for only BTC and DCR.",
+	"delete_archived_records":     "Delete Archived Records",
+	"date_time":                   "Date and Time",
+	"archived_date_time_tooltip":  "Archived orders and matches created before your specified and date and time will be deleted from the database. Note: If an invalid date or time is provided, all archived records till date will be deleted.",
```
martonp, line 279: The text should be "all archived records will be deleted" instead of "all archived records **till date** will be deleted." However, I think the user should be able to select to either delete all the records, or if the date is invalid, then there should be an error message. I don't think an invalid date should delete everything.

martonp (commented):
Looks good.

I think there should be another screen showing the path of the archived files and the number of deleted records after submit is clicked. Right now there's no feedback after doing the deletion.

client/webserver/locales/en-us.go
chappjc: Agree it is unexpected to have that behavior when leaving the date field of the UI empty.  I only did that accidentally.  A "Specify date of latest records to keep" checkbox could work (or reverse wording).  By default, unchecked, and no data selection/entry box shown, and `null` sent to backend; if checked, show the date entry, and reject submission if no valid date entered.

I'm a little less concerned with summarizing after the purge, but ideally it would report on the location etc.

ukane-philemon: Okay.

ukane-philemon: > Right now there's no feedback after doing the deletion.

@chappjc,  do you think we need to modify the existing API to return `no of records deleted`?

chappjc: The mod to the db api is minor, and since we added a core method it's ok to change that.
Don't spend a lot of time on it, but it's worth a shot at showing some kind of result.

ukane-philemon: Screenshot 2022-10-29 at 9 53 28 AM

ukane-philemon: My recent commit at https://github.com/decred/dcrdex/pull/1916/commits/d64637e9bf22dda58c09350f1e0a62f044e55b8a also fixed an issue where files are created even if no record was deleted.

martonp (approved):
LGTM

client/db/interface.go
```diff
@@ -118,16 +118,17 @@ type DB interface {
...
-	// supplied the current time is used. Accepts an optional function to
-	// perform on deleted matches that includes if it was a sell order.
-	DeleteInactiveMatches(ctx context.Context, olderThan *time.Time, perMatchFn func(mtch *MetaMatch, isSell bool) error) error
+	// DeleteInactiveOrders deletes inactive orders from the database that are
+	// older than the supplied time and returns the total number of orders
+	// deleted. If no time is supplied, the current time is used. Accepts an
+	// optional function to perform on deleted orders.
+	DeleteInactiveOrders(ctx context.Context, olderThan *time.Time, perOrderFn func(ord *MetaOrder) error) (int, error)
+	// DeleteInactiveMatches deletes inactive matches from the database that are
+	// older than the supplied time and return total number of matches deleted.
```
JoeGruffins, line 127: return -> returns

JoeGruffins (approved):
Working great.

Issue #1908, opened 2022-10-17
UI: inaccurate Price warning (when placing limit order)


norwnd:
The warning on screenshot below (it only appears if I'm hovering over Price field that needs to be filled) seems to suggest that price should be an integer number.

On the other hand, when I was trying to set price at `0.0000099` BTC per DCR I was able to click on `Place order to buy DCR` button and it silently cut this price off to `0.000009` and let me successfully place an order (I didn't even notice that price was rounded down until later; it might be just sub-optimal UX when buying DCR, it is even worse when selling DCR because rounding down means I'm selling for less than I intended to if I don't double-check).

Instead I think it would be better if UI didn't allow me to enter a Price value that doesn't satisfy `Rate Step` in the first place (possibly highlighting `Rate Step` value in red  when I try to, so that I can see why it doesn't).

<img width="365" alt="image" src="https://user-images.githubusercontent.com/112318969/196187668-3ef3c6dc-6e7f-4a93-9aa1-0d81f0b045ff.png">

I'm running latest `master` of this repo on testnet.

chappjc:
Instead of js validation when you change focus, we should look at using the `step` attribute. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#step

norwnd:
Thanks for the hint, lets see if I can fix/improve this in a PR of my own.

chappjc:
I tried the `step` attribute manually shortly after suggesting to check it out, and it didn't seem to actually enforce anything, although it did change the message that was displayed on hover.

I'm not sure what changes silently though.  As soon as the price field looses focus, it applies the rate step.  When you click "Place order..." the confirmation screen shows the rate with any rate step interval applied.  How can we make the price field more responsive?

ukane-philemon:
> I tried the step attribute manually shortly after suggesting to check it out, and it didn't seem to actually enforce anything,

> Note: When the data entered by the user doesn't adhere to the stepping configuration, the [user agent](https://developer.mozilla.org/en-US/docs/Glossary/User_agent) may round to the nearest valid value, preferring numbers in the positive direction when there are two equally close options.

norwnd:
I've been fiddling with this for last couple of days, will have something to show soon.

norwnd:
Same thing with rounding to integer number is present for Market Buy Order as well (`mktBuyField` field), 

[...]

PR #1921, opened 2022-10-25
UI: make limit order form more responsive

norwnd:
In this PR I'm trying to address the following:
- `Price`, `Quantity` fields should have reasonable defaults when `/markets` page loads (and when garbage user input gets reset), so that less clicks/actions are required by the user to place his limit order
- validation for `Price`, `Quantity` fields should be more responsive, so that the user has a better sence of why his inputs don't make sense
- `Place order to buy ...` button press should fail right away when `Total` exceeds `Max Buy`, instead of allowing the user to proceed only to fail later on

Here is a demo to get the general idea - https://skynetfree.net/AADCen1qLW87koj0oLmJpgLg9cOMInagd1W0CS8dOQdtkQ (couldn't make reasonably small recording to fit 10MB Github limit with `Kap` to embed it here)

Closes https://github.com/decred/dcrdex/issues/1908.


client/webserver/site/src/js/markets.ts
```diff
@@ -1972,14 +2023,11 @@ export default class MarketsPage extends BasePage {
       order: wireOrder(order),
       pw: pw
     }
-    if (!this.validateOrder(order)) return
```
norwnd, line 1975: This validation has no effect since user can't modify any of the values on `verifyForm` page and `orderForm` page already calls this func

client/webserver/site/src/js/locales.ts
```diff
@@ -93,6 +94,7 @@ export const enUS: Locale = {
   [ID_ESTIMATE_UNAVAILABLE]: 'estimate unavailable',
   [ID_NO_ZERO_RATE]: 'zero rate not allowed',
   [ID_NO_ZERO_QUANTITY]: 'zero quantity not allowed',
+  [ID_NO_QUANTITY_EXCEEDS_MAX]: 'not enough funds',
```
norwnd, line 97: I'm not sure what needs to be done about translations, or will it pick English by default if the local one isn't present ?

client/webserver/site/src/js/markets.ts
```diff
@@ -2081,17 +2151,30 @@ export default class MarketsPage extends BasePage {
    * input.
    */
   rateFieldChanged () {
-    // Truncate to rate step. If it is a market buy order, do not adjust.
```
norwnd, line 2084: I think this comment is out of date since for market order `rate` gets handled by different func.

client/webserver/site/src/js/locales.ts
ukane-philemon: Yes, it will.

chappjc:
When pushing updates to reviewable PRs, please try to ensure there is always a way to see just what was changed since the last push. Either add commits, which we will squash before merge, or force push once without changing the merge-base so that when we view the "force-pushed" diff it shows only what was changed in your PR.  If the merge-base changes, as with the latest force-push, it is mixed up with whatever was changed on master, making it more difficult to review.

JoeGruffins (commented):
Looks really cool! Noticing a couple things while testing:

Can you also prevent negative values?

image

Is the max order calculation correct? I don't have this much btc to sell.

image

client/webserver/site/src/js/markets.ts
```diff
@@ -785,9 +811,22 @@ export default class MarketsPage extends BasePage {
...
 
+    // reset form inputs
+    const lot = '1'
+    page.lotField.setAttribute('min', lot)
+    page.lotField.setAttribute('step', lot)
+    page.lotField.value = lot
+    const lotSize = Doc.formatCoinValue(this.market.cfg.lotsize, this.market.baseUnitInfo)
+    page.qtyField.setAttribute('min', lotSize)
+    page.qtyField.setAttribute('step', lotSize)
+    page.qtyField.value = lotSize
```
buck54321, line 822: I do not agree that we should have a default here. It is not standard on any exchange's order form that I've used.

client/webserver/site/src/js/markets.ts
```diff
@@ -94,6 +94,8 @@ const percentFormatter = new Intl.NumberFormat(document.documentElement.lang, {
 
 const parentIDNone = 0xFFFFFFFF
 
+const colorPink = 'pink'
```
buck54321, line 97: I'm not feeling this color.

client/webserver/site/src/js/markets.ts
```diff
@@ -2904,3 +2996,38 @@ function hostColor (host: string): string {
...
+    // If we are here, the element is already undergoing highlight transition,
+    // we must not interfere with it - if we do, it might break.
+    return
+  }
+
+  const previousBColor = element.style.backgroundColor
+  const prevTransition = element.style.transition
+
+  element.style.backgroundColor = color
+  element.style.transition = 'background 500ms'
```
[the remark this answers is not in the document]
buck54321: Another thing to consider is that the `transition` attribute applies both direction, which means we can only fade in -> fade out, we can't say flash on -> fade out, which I think would be a better indicator. 

A couple of suggestions:
1) Use `Doc.animate` to do a flash -> fade
2) For inputs, at least, animate the element's `outline` attribute rather than the background color.

buck54321 (commented):
Please tweak the "cache busters" in the url queries [here](https://github.com/decred/dcrdex/blob/4dc0bc4accb65dc1878240b030ddfc7dae5e6d4a/client/webserver/site/src/html/bodybuilder.tmpl#L11) and [here](https://github.com/decred/dcrdex/blob/4dc0bc4accb65dc1878240b030ddfc7dae5e6d4a/client/webserver/site/src/html/bodybuilder.tmpl#L91). They just need to be changed in some way to force reloading of web assets.

client/webserver/site/src/js/markets.ts
norwnd: Some reasons for why having defaults there might be a good idea:
- it provides **an example**, emphasizing the Lot concept (present in Dex, and not on most other exchanges, at least not so explicitely) = more streamlined experience for users trying out Dex for the very first time
- many first-time users will probably want to **start with 1 lot**, so less clicks overall = good for onboarding (and it doesn't affect those who want to increase this to 2, 3, ... except, as is implemented right now - one has to erase those defaults first to put in another value, this is probably a deal-breaker, but I hope this can be solved somehow)
- having those defaults there allows for `Total` calculated on page load, showing the user all the order-form parts in complete (and also happens to show min trade size which is neat)
- some manual UI testing is simpler with defaults, since I can place a bunch of limit order faster

If that's convincing enough, then I can try to address that "deal-breaker" thing, if not - I'll just remove defaults.

buck54321: Other than illustrating the lot concept, all of these points could be made about any exchange interface, yet nobody does this. I don't think it's a good idea to pre-fill critical fields for people.

chappjc: Agree.  Plus there are events that can force reload the page, and if a valid order size replaces what you had entered before, you may not notice, and end up submitting the wrong order.

norwnd: Interesting, ok - will remove.

How do you feel about slider, perhaps in the next PR ? Seems like a popular feature these days.

image

norwnd:
> Is the max order calculation correct? I don't have this much btc to sell.

- The calculation of `Total` value (for large values) seems off (`Total` value on your screenshot is not what `1e+50 / 10` should be); It seems to me the reason is `parseFloat` is used to parse all those order-form `inputs` which returns `number` that's represented as `float` / `double` (32 / 64 bit). There seems to be no native decimal of arb precision in JS based on brief googling, there are some libs out there,  for example: https://github.com/MikeMcl/decimal.js/ or https://github.com/MikeMcl/big.js/, however I don't know what your opinion of using a lib for this is, this particular issue probably isn't that nasty to warrant such usage, maybe there are other places where UI might benefit from using precise decimals though.

- The fact that "you don't have this much BTC to sell" is actually on purpose, allowing the user to fill in whatever (positive) numbers he likes so that he can estimate `Total` without having the actual funds in his wallet; the actual balance-check will happen when `Place order ...` button is pressed.

@JoeGruffins, guys ^ if you have any suggestions on this

client/webserver/site/src/js/markets.ts
buck54321: Love the slider!
I also use these buttons sometimes
image

client/webserver/site/src/js/markets.ts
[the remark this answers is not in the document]
norwnd: > why not specify it in CSS?

My knowledge of CSS and some front-end stuff is a bit rusty, I've rewrote this piece with `Doc.animate`; please let me know if you had a different solution in mind.

> The prevTransition stuff is likely unnecessary. Did you have a conflict with another transition value?

Not really, the example I was following was doing that - my guess would be it's just "general purpose code", sort of best practice; removed.

I've applied those suggestions, seem - looks better indeed!

client/webserver/site/src/js/markets.ts
norwnd: > I'm not feeling this color.

Changed to red, or do you have something else in mind ?

norwnd:
I've addressed reviews above, and in the process had to rewrite `change` & `keyup` event handling on limit-order form (also touching other parts of markets page a bit), hopefully you don't mind that.

Also cleaned up Market-Order form tiny bit (e.g. `Rate step` and `Total` have no business being present there). Hope to improve it slightly in a separate PR later.

Couldn't quite figure out yet why linter fails though, on line 571 I don't even have 7th column ...
```
Error: src/js/markets.ts(571,7): error TS2532: Object is possibly 'undefined'.
Error: Process completed with exit code 2.
```

client/webserver/site/src/js/markets.ts
```diff
@@ -554,6 +551,38 @@ export default class MarketsPage extends BasePage {
...
+  }
+
+  // defaultRate return default exchange rate (aka price).
+  defaultRate (): number {
+    // Current exchange rate would be reasonable default Price value.
+    return app().conventionalRate(
+      this.market.cfg.baseid,
+      this.market.cfg.quoteid,
+      this.market.cfg.spot.rate,
+      this.market.dex
```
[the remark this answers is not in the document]
ukane-philemon: @norwnd, maybe you should do this instead. Or check that `market` || `cfg` is not `undefined` before accessing those fields. That's why CI run keeps complaining.

client/webserver/site/src/js/markets.ts
```diff
@@ -557,6 +559,39 @@ export default class MarketsPage extends BasePage {
...
```
ukane-philemon, line 572:
```js
    const { market: { cfg: { baseid, quoteid, spot }, dex } } = this
    let rate = 0
    if (spot) rate = spot.rate
    return app().conventionalRate(baseid, quoteid, rate, dex)
```
`spot` could be `undefined`.

norwnd: Changed, and it fixed linter, thanks!

JoeGruffins:
> The fact that "you don't have this much BTC to sell" is actually on purpose

Oh. I thought the purpose was to not allow spending more than available.

JoeGruffins (approved):
Looks good to me.

client/webserver/site/src/js/markets.ts
norwnd: Haven't seen those buttons (mid, bid, ...) before, but I'll give a try to slider later on in a separate PR.

norwnd:
On `/order` page that imprecise rounding I've mentioned above also shows - `Rate` should probably be trimmed to "rate-step" there as well ?

image

JoeGruffins:
@norwnd I have a change in another pr that may fix this https://github.com/decred/dcrdex/pull/1733#issuecomment-1303080513

Just multiplying before dividing in the js `conventionalRate`

client/webserver/site/src/html/markets.tmpl
```diff
@@ -233,12 +233,12 @@
 
                   {{- /* MARKET CONFIG */ -}}
                   <div class="d-flex justify-content-between my-2 fs14">
-                    <div>
+                    <div id="lostSizeBox">
```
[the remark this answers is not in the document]
buck54321: You have resolved this, but have not changed anything.

client/webserver/site/src/js/markets.ts
buck54321: I'd start by seeing if any of the constants in **application.scss** would work. Two that jump out at me are

```SCSS
$sellcolor_dark: #e95e5e;
$sellcolor_light: #99302b;
```

And then set the outline color in both **market.scss**.

```SCSS
div[data-handler=markets] {
  input {
    outline: 2px solid $sellcolor_light;
  }
}
```

and **market_dark.scss**.

```SCSS
body.dark {
  div[data-handler=markets] {
    input {
      outline: 2px solid $sellcolor_dark;
    }
  }
}
```

buck54321: Actually, I don't know that would work with the animation. I suppose you would need to define the rgb values in **markets.ts** after all. You can pick which color (dark/light) using `State.isDark()`.

client/webserver/site/src/js/registry.ts
```diff
@@ -376,6 +376,7 @@ export interface PageElement extends HTMLElement {
   checked?: boolean
   href?: string
   htmlFor?: string
+  animating?: boolean
```
buck54321, line 379: Please don't add custom attributes to elements.

client/webserver/site/src/js/markets.ts
```diff
@@ -2914,3 +3145,65 @@ function hostColor (host: string): string {
...
+}
+
+// highlightOutlineRed will add red outline to the element and will smoothly remove
+// it afterwards.
+function highlightOutlineRed (element: PageElement) {
+  if (element.animating) {
+    // Don't want to interfere with already running animation.
+    return
+  }
+  element.animating = true
```
buck54321, line 3180: I know we're doing it in a couple of places still, but let's avoid adding custom attributes to native elements. Instead, use `Animation`, keep a list of running `Animation`s and cancel all of them when running a new one. Roughly...
```Typescript
import Doc, { WalletIcons, Animation } from './doc'
```

```Typescript
const errorAnimations: Animation[] = []

async function highlightOutlineRed (el: PageElement) {
  for (const ani of errorAnimations) await ani.stopAndWait()
  while (errorAnimations.length) errorAnimations.pop()
  const [r, g, b] = State.isDark() ? [233, 94, 94] : [153, 48, 43]
  /* Set the outline width in market.scss */

  errorAnimations.push(new Animation(animationLength, (progress: number) => {
    el.style.outlineColor = `rgba(${r}, ${g}, ${b}, ${1 - progress})`
  }, 'easeIn'))
}
```

client/webserver/site/src/js/markets.ts
```diff
@@ -2027,51 +2077,175 @@ export default class MarketsPage extends BasePage {
...
-    // Conversion factor must be a multiple of 10.
-    page.qtyField.value = String(lots * lotSize / this.market.baseUnitInfo.conventional.conversionFactor)
-    this.previewQuoteAmt(true)
+
+    const [inputInvalid, adjusted,, adjQty] = this.parseLotInput()
+    if (inputInvalid || adjusted) {
+      // Let the user know that lot value he's entered was rounded down to the
+      // nearest integer number.
+      highlightOutlineRed(page.lotField)
+      highlightBackgroundRed(page.lostSizeBox)
```
buck54321, line 2096: Sorry. I meant to do **just** the outline, at least for input elements. We don't need both.

client/webserver/site/src/js/markets.ts
```diff
@@ -2027,51 +2077,175 @@ export default class MarketsPage extends BasePage {
...
+
+    this.previewTotal()
   }
 
-  /*
-   * quantityChanged is attached to the keyup and change events of the quantity
-   * input.
-   */
-  quantityChanged (finalize: boolean) {
+  lotFieldChangeEvent () {
```
buck54321, line 2111: The input `--5` in the lots input does not show an error indication and does not self-adjust to zero.

On master, a negative sign is removed immediately after entering. That feels better.

buck54321 (changes requested):
Sorry for all the requests. Most of these are just adapting the code to our existing style.

client/webserver/site/src/js/markets.ts
```diff
@@ -913,44 +952,36 @@ export default class MarketsPage extends BasePage {
   }
 
   /**
-   * previewQuoteAmt shows quote amount when rate or quantity input are changed
+   * previewTotal calculates Total value (in quote asset) for the order.
    */
-  previewQuoteAmt (show: boolean) {
+  previewTotal () {
     const page = this.page
-    if (!this.market.base || !this.market.quote) return // Not a supported asset
```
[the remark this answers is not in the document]
buck54321: Actually, can't test this part because of #1949.

client/webserver/site/src/js/markets.ts
buck54321: Actually, I don't know that would work with the animation. I suppose you would need to define the rgb values in **markets.ts** after all. You can pick which color (dark/light) using `State.isDark()`.

client/webserver/site/src/js/markets.ts
```diff
@@ -206,7 +206,14 @@ export default class MarketsPage extends BasePage {
...
 
     const depthReporters = {
-      click: (x: number) => { this.reportDepthClick(x) },
+      // When the user clicks on depth-chart area, this click func gets called with x
+      // being the value user chose (value along x-axis).
+      click: (x: number) => {
+        // Must adjust chosen value here since it might not be conforming to rate-step.
+        this.page.rateField.value = String(x)
+        const [,, adjRate] = this.parseRateInput()
+        this.reportDepthClick(adjRate)
```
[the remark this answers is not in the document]
norwnd: Previously `rateFieldChanged` was doing several things (being both `change` event handler and being called from `reportDepthClick` for example),

once I started modifying the way `change` event handler works, it no longer made sense to call same event handler code in `reportDepthClick`, hence I had to move rate-adjustment here.

client/webserver/site/src/js/markets.ts
```diff
@@ -913,44 +952,36 @@ export default class MarketsPage extends BasePage {
...
-        color: order.sell ? this.depthChart.theme.sellLine : this.depthChart.theme.buyLine
-      }]
-    }
-    this.drawChartLines()
-    if (!show || !adjusted || !order.qty) {
-      page.orderPreview.textContent = ''
-      this.drawChartLines()
-      return
-    }
-    const quoteAsset = app().assets[order.quote]
```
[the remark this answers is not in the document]
norwnd: It was moved to `rateFieldKeyUpEvent` func now, where [this.drawChartLineInputRate()](https://github.com/norwnd/dcrdex/blob/3d3067e05f4dbd88dd0801e7a287e2824a5ce9f9/client/webserver/site/src/js/markets.ts#L2278) is called for the same purpose, I believe.

client/webserver/site/src/js/markets.ts
```diff
@@ -913,44 +952,36 @@ export default class MarketsPage extends BasePage {
...
+
+    const quoteAsset = app().assets[market.quote.id]
     const quoteQty = order.qty * order.rate / OrderUtil.RateEncodingFactor
-    const total = Doc.formatCoinValue(quoteQty, this.market.quoteUnitInfo)
+    const total = Doc.formatCoinValue(quoteQty, market.quoteUnitInfo)
 
-    page.orderPreview.textContent = intl.prep(intl.ID_ORDER_PREVIEW, { total, asset: quoteAsset.symbol.toUpperCase() })
-    if (this.isSell()) this.preSell()
-    else this.preBuy()
```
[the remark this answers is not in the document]
norwnd: With `this.previewTotal()` actually, since there is no need to re-calculate `Max` when quantity/lots field changes.

And similar to `rateFieldChanged` func (like I mentioned above), `previewQuoteAmt` was doing a lot of different things which I also split into smaller ones to combine as needed.

client/webserver/site/src/js/markets.ts
```diff
@@ -1081,16 +1112,35 @@ export default class MarketsPage extends BasePage {
...
     const page = this.page
-    if (order.isLimit && !order.rate) {
+
+    const showError = function (err: string) {
+      page.orderErr.textContent = intl.prep(err)
       Doc.show(page.orderErr)
-      page.orderErr.textContent = intl.prep(intl.ID_NO_ZERO_RATE)
+      setTimeout(function () {
+        Doc.hide(page.orderErr)
+      }, 8000) // 8s should be enough for user to consume error message.
```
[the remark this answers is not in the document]
norwnd: Yeah, I guess I went for a shortcut here ... will try to do that ^

client/webserver/site/src/js/markets.ts
norwnd: I highlight 2 different elements here (not same element twice):
- for `input` outline is highlighted
- for others it's background,

I think it makes sense to highlight `page.lostSizeBox` in addition to the field where user types his input, so that he gets a better sense of what's wrong,

maybe for **lot-input** of particular it's not very helpful, but when user enters **quantity=15** (and **lot size = 20**) highlighting lot size makes sense to me

client/webserver/site/src/js/markets.ts
```diff
@@ -2027,51 +2077,175 @@ export default class MarketsPage extends BasePage {
...
+    page.qtyField.value = String(adjQty)
+
+    this.previewTotal()
+  }
+
+  // parseLotInput parses lot input and returns whether there are any
+  // parsing issues, whether rounding(adjustment) to it had happened, and
+  // returns adjusted lot value. If lot value couldn't be parsed (parsing
+  // issues), 0 values, not defaults, are returned for 2nd, 3rd, 4th value.
+  parseLotInput (): [boolean, boolean, number, number] {
```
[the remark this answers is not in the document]
norwnd: Agree, I was just hesitant to change that while refactoring other things, now is a good time to do that, changed.

client/webserver/site/src/js/markets.ts
```diff
@@ -2027,51 +2077,175 @@ export default class MarketsPage extends BasePage {
     this.balanceWgt.updateAsset(this.openAsset.id)
   }
 
-  /* lotChanged is attached to the keyup and change events of the lots input. */
-  lotChanged () {
+  lotFieldKeyUpEvent () {
```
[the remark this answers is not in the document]
norwnd: Yes, updated.

client/webserver/site/src/js/markets.ts
```diff
@@ -557,6 +559,37 @@ export default class MarketsPage extends BasePage {
         setPriceAndChange(s.tmpl, xc, mkt)
       }
     }
+    this.page.rateField.value = fourSigFigs(this.defaultRate())
```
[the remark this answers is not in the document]
norwnd: I think this just sets "default" value for rate when market page loads or re-loads (based on current price),
actually I'm already doing it in `resolveOrderFormVisibility`, so I don't need it here, removed.

> In fact, we should never be setting the rate field for the user.

or do you mean it like [no defaults for lots/quantity](https://github.com/decred/dcrdex/pull/1921#discussion_r1011652574) we discussed above ? I think almost all exchanges suggest some sort of default price, and that's very convenient feature IMO because typing in decimal point value with lots of digits is error-prone and pretty much unnecessary burden on the user (compared to suggested one). I can see how having defaults for lots/quantity could be problematic (having all fields: rate, lots/quantity set allows for placing order  mistakenly, "without looking"), but since we don't set lots/quantity defaults now I don't see how setting default rate could result into issues, so why not ?

client/webserver/site/src/js/markets.ts
```diff
@@ -557,6 +559,37 @@ export default class MarketsPage extends BasePage {
...
+    const adjRateAtoms = this.adjustedRateAtoms(this.page.rateField.value)
+    let maxEstimate: MaxOrderEstimate | null = this.market.maxBuys[adjRateAtoms]
+    if (this.isSell()) {
+      maxEstimate = this.market.maxSell
+    }
+    if (!maxEstimate) {
+      return 0
+    }
+    return maxEstimate.swap.lots
+  }
```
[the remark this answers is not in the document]
norwnd: Seems to be working, changed.

client/webserver/site/src/html/markets.tmpl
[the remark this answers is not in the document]
norwnd: Hmm, I guess got distracted by something else and then forgot, thanks for catching it again!

client/webserver/site/src/js/markets.ts
norwnd: - maybe keep those animations on `page` (not global) scope ?
- and we probably can just do `page.runningAnimations = []` instead of doing `pop()` for each element; or would this prevent GC from collecting array  elements ?

----------------------------------------------------------------------------

also note, based on the testing I did, using `ani.stop()` (or even `ani.stopAndWait()`) by itself doesn't address the issue when 2 or more  animations are started in parallel (which happens when I change Price field very fast, say `0.000003` -> `0.00000321` in under half-a-second), resulting in highlighting getting stuck like this:

image

I've pushed [this commit](https://github.com/decred/dcrdex/pull/1921/commits/c7524e95be16e7f66eda80208d43fb0cb3a8596c) in case you want to test it for yourself;

so, some kind of event de-duplication is needed anyway, not that my previous solution fully solved it:
```typescript
function highlightBackgroundRed (element: PageElement) {
  if (element.animating) {
    // Don't want to interfere with already running animation.
    return
  }
  element.animating = true
```
I think it ^ just hid it very well ... Don't know what would the proper solution would look like yet, any ideas ?

norwnd: [This simple adjustment](https://github.com/decred/dcrdex/pull/1921/commits/f8be54fbd74144f034b03fe72bbf494fedc84c2f) "hides" the issue pretty well by the way (in a similar way as my original implementation does), but it doesn't fully solve it:

image

in console there are 4 "started" prints, while I expect 2 at most.

norwnd: The issue pretty much boils down to `async` / `await` usage in `Animation` class.

I found a good explanation [here](https://stackoverflow.com/a/74276307) that states that `events` execute to completion without interruptions (unless it runs into `await`).

norwnd: > Don't know what would the proper solution would look like yet, any ideas ?

**In this particular case**, I guess, I can define `done` on `Animation` in such a way that it doesn't depend on some "current" value, since it is pretty much known ahead of time, 

that is, I can do this (in `done` func):
```
element.style.backgroundColor = 'none'
```
instead of:
```
const prevColor = element.style.backgroundColor
...
element.style.backgroundColor = prevColor
```

**that seems to be working OK ^**, but it is something to keep an eye our for in the future, I guess

client/webserver/site/src/js/markets.ts
norwnd: Yeah, I don't think I can make css can work with animation, or at least don't know how to make it workable and simple,

I went the route you've suggested [below](https://github.com/decred/dcrdex/pull/1921#discussion_r1017804446)

client/webserver/site/src/js/markets.ts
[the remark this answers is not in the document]
buck54321: Mmkay. I think you convinced me.

client/webserver/site/src/js/markets.ts
norwnd: Should work better now,

but I still have quite a bit of minor things to address to make sure different  edge-cases are covered and overall experience is smooth.

client/webserver/site/src/js/registry.ts
norwnd: Changed as was suggested in https://github.com/decred/dcrdex/pull/1921#discussion_r1017804446.

client/webserver/site/src/js/markets.ts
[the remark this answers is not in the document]
norwnd: I tried
```
go test -tags lgpl,live -timeout 60m
```
and I don't seem to hit the issue described in https://github.com/decred/dcrdex/issues/1949

but from what I understand, that line executes when something on order-form gets changed (`previewQuoteAmt` name kinda implies that ?), but when we are viewing a market that `dexc` (UI) doesn't support we won't be even seeing order-form in the first place,

MONA-KMD market seems to be looking fine to me (I see no errors in console, similar and everything is looking similar to what I see running same test on **master**)

image

I also tried to comment out this line on **master** just to compare:
```
// if (!this.market.base || !this.market.quote) return // Not a supported asset
```
and then I **do** see this error in console (which doesn't appear with the code from this PR) you were probably addressing by having this line there:

image

client/webserver/site/src/js/markets.ts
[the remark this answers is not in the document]
norwnd: Applied the suggestion.

norwnd:
@buck54321 I think I've addressed everything you mentioned ?

there are probably still some minor things to improve - I might get to those later (maybe as a follow-up PR).

buck54321 (changes requested):
Working nicely.

client/webserver/site/src/js/markets.ts
```diff
@@ -968,16 +1043,21 @@ export default class MarketsPage extends BasePage {
       mkt.maxSellRequested = false
       mkt.maxSell = res.maxSell
       mkt.sellBalance = baseWallet.balance.available
-      this.setMaxOrder(res.maxSell.swap)
+      // Since this executes in callback quite a bit of time could have passed, and the
+      // user might have switched to another form already (from buy to sell order form,
+      // or vise-versa), update max value only if user is on the relevant form.
+      if (this.isSell()) {
+        this.setMaxOrder(res.maxSell.swap)
+      }
```
norwnd, line 1051: I actually hit this while testing. This wouldn't be an issue were it not for the fact that single `Max` view gets re-used for both `buy` / `sell`. I was able to split many of Limit and Market related stuff into separate view, but splitting `buy` and `sell` just because of this issue doesn't seem worth it.

client/webserver/site/src/js/markets.ts
```diff
@@ -601,19 +631,16 @@ export default class MarketsPage extends BasePage {
   setOrderVisibility () {
     const page = this.page
     if (this.isLimit()) {
-      Doc.show(page.priceBox, page.tifBox, page.qtyBox, page.maxBox)
+      Doc.show(page.priceBox, page.tifBox, page.qtyBox, page.maxBox, page.rateStepBox, page.orderTotalPreview)
       Doc.hide(page.mktBuyBox)
-      this.previewQuoteAmt(true)
     } else {
-      Doc.hide(page.tifBox, page.maxBox, page.priceBox)
+      Doc.hide(page.priceBox, page.tifBox, page.maxBox, page.rateStepBox, page.orderTotalPreview)
```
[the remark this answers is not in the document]
norwnd: Initially I wanted to keep the scope of this PR to improving limit-order form, but since that didn't quite work out this way and I changed market-order form already - might as well address it too here.

I've created split event-handlers on market order form (mostly for sell side) and now also validate lot/quantity inputs the same way limit order form does; also added "Total: ~ estimate" the user will receive.

buck54321 (commented):
I feel like the warning is a little agressive. If I'm typing in a valid number but just haven't finished typing, it really shouldn't be telling me I'm wrong already. For instance, in the video below, with a rate step of 0.0000001, if I'm typing the number 0.001, I see three red flashes while I'm typing. Can we wait for the `change` event to flash?

https://user-images.githubusercontent.com/6109680/204164171-156a79c7-73b0-4d0e-b4b7-89fa598fb2a8.mp4

norwnd:
Added [another commit](https://github.com/decred/dcrdex/pull/1921/commits/00cfca44e244c4a3ac019bf90736d05892a3af46) that seemingly strikes a good balance between interactivity and validation aggressiveness.

client/webserver/site/src/js/markets.ts
```diff
@@ -2062,54 +2203,251 @@ export default class MarketsPage extends BasePage {
...
+
+  // parseLotInput parses lot input and returns:
+  // 1) whether there are any parsing issues (true if none, false when
+  //    parsing fails)
+  // 2) whether rounding(adjustment) had happened (true when did)
+  // 3) adjusted lot value
+  // 4) adjusted quantity value
+  //
+  // If lot value couldn't be parsed (parsing issues), the following
+  // values are returned: [false, false, 0, 0].
```
buck54321, line 2306: For multi-line docstrings, use 
```
/*
 * This function ...
 * ... last line.
 */
```

buck54321 (changes requested):
Managed to get into a weird state. 
1. Start from scratch. Register with dcr rpc alpha node
2. On markets page, create btc wallet, using rpc alpha node
3. From dcr harness, `./mine-alpha 1`
The Max Buy estimate is stuck on "calculating".
image
There are some warnings in the console.
image
Max sell estimate is working fine. Going back and forth between buy and sell does not fix the problem.

---

I'm seeing that the rate field is being populated with a `0`. Is that intended?

---

For the market sell 'Total:', the div should have a fixed height so that the form elements don't jump around when the input fields are first used or cleared.
image

client/webserver/site/src/js/markets.ts
```diff
@@ -761,14 +821,14 @@ export default class MarketsPage extends BasePage {
     const [bui, qui] = [app().unitInfo(base, dex), app().unitInfo(quote, dex)]
 
     const rateConversionFactor = OrderUtil.RateEncodingFactor / bui.conventional.conversionFactor * qui.conventional.conversionFactor
-    Doc.hide(page.maxOrd, page.chartErrMsg)
```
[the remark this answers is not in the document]
norwnd: I think I was after `Doc.hide(page.maxOrd` part (with the interactive approach this PR takes there is no need to hide/show it), and accidentally removed `page.chartErrMsg` piece as well,

changed to `Doc.hide(page.chartErrMsg)` now, thanks!

chappjc:
I don't have the time to review this PR now, but I want to verify that this does not make API requests any _more frequently_ than before.  Is that true?  It can be very computationally costly to do things like `/api/prebuy` and `/api/maxsell`, how costly depending on the wallet and factors like size of its UTXO set.  Even slight increases in latency on these endpoints result in a net negative UX if we start hitting them more often.

norwnd:
> I want to verify that this does not make API requests any more frequently than before. Is that true?

It didn't seem to overload my machine, but I don't have lots of UTXOs in my wallets, and there is some caching working in our favor.

**For `maxbuy` and `maxsell`:**

From what I'm seeing in current `master` it looks like `Max Buy` is re-calculated on each price change, that's same behavior as in this PR except that for this PR we do 1  additional `maxbuy` request when `/markets` page loads or when user chooses a particular market from the list because from user-set  values it depends only on rate/price which gets initialised to default value of "current market price" (or at least it **should** with [these minor adjustments](https://github.com/decred/dcrdex/pull/1921/commits/a6cec235c592d15033b1f6aeafca7479a4dcd2c1) that removed unnecessary additional requests for `maxbuy`, based on my testing).

**For `pre-` types of requests:**

I don't think this PR changes anything.

client/webserver/site/src/js/markets.ts
```diff
@@ -632,7 +677,31 @@ export default class MarketsPage extends BasePage {
...
 
     if (feePaid && assetsAreSupported && hasWallets) {
+      const lot = '1'
+      const lotSize = String(this.market.cfg.lotsize / this.market.baseUnitInfo.conventional.conversionFactor)
+      const rateStep = String(this.market.cfg.ratestep / this.market.rateConversionFactor)
+      const rate = this.defaultRate()
+
+      // Reset limit-order form inputs to defaults.
+      page.lotField.min = lot
+      page.lotField.step = lot
```
buck54321, line 687: It's not clear to me that setting `min` and `step` actually do anything for us. If I comment these out, it doesn't seem to change my ability to type, say, a decimal number in the lots input.

buck54321 (approved):

client/webserver/site/src/js/markets.ts
chappjc: Same.  Browser ignores it except for changing the little hover message.

chappjc:
Why does the rate field have "0" pre-filled?  It is not valid under any circumstance, and it is annoying.

https://user-images.githubusercontent.com/9373513/208001701-0f1a3574-c023-437c-a2b1-ca6726168490.mp4

chappjc:
If it now has logic to pre-fill with mid gap rate, OK, I guess.  But if the book is empty, don't pre-fill.
I'm not at all convinced we should put any rate in there by default.  There is literally no precedent for doing that.

chappjc:
JS error refreshing market page:

image

Also, the default rate seems to be 0 regardless of book.

client/webserver/site/src/js/markets.ts
norwnd: > except for changing the little hover message

yes, that's exactly the reason to change this, because otherwise the message shown doesn't make sense to me

norwnd:
> Why does the rate field have "0" pre-filled? It is not valid under any circumstance, and it is annoying.

It gets pre-filled with the "default" value which I'm defining as "current market rate/price", I've outlined my reasoning to have it for **buck54321** [here](https://github.com/decred/dcrdex/pull/1921#discussion_r1020769681).

> if the book is empty, don't pre-fill.

Okay, that can be fixed - I've done another iteration on that [here](https://github.com/decred/dcrdex/pull/1921/commits/75e615875e5ef6e11670402621f467aa908feddd) along with some other minor improvements, does it behave more like what you'd expect ?

> JS error refreshing market page

Shouldn't happen now, since I've added a check for non-zero rate before sending `maxbuy` request in UI, and re-tested.

norwnd:
> Also, the default rate seems to be 0 regardless of book.

That's caused by another issue (probably unrelated to this PR) I've observed -  **mid-gap-rate** (`0.001354` displayed on depth-chart on your screenshot above) and **Price** (`0.000` top left corner on your screenshot above) values shown in UI diverge,

- testing on `testnet` several weeks ago, I observed that **mid-gap-rate** changed immidiately once order has been placed on the book, but **Price** didn't change accordingly until I would refresh web-page
- testing on `testnet` last week, **Price** didn't update at all (stayed at `0.000`) regardless of whether order-book was empty or filled

So, currently in this PR for **default rate** I'm using same value that's used to fill **Price** value. I was hoping to look into this issue and fix it in a follow-up PR, thinking maybe it's just DEX-server on `testnet` having some difficulties delivering updates for **Price** value (I'm currently observing "constant" `0.000001000` value on DCR-BTC market and `0.000` on DCR-ETH, **and these values don't change** when I'm putting order on the book on `testnet` with `master`, attached screenshot below).

image

buck54321:
> Why does the rate field have "0" pre-filled? It is not valid under any circumstance, and it is annoying.

> It gets pre-filled with the "default" value which I'm defining as "current market rate/price", I've outlined my reasoning to have it for buck54321 https://github.com/decred/dcrdex/pull/1921#discussion_r1020769681.

I had my reservations too, clearly. Let's scrap the pre-fill in this PR, and [add little buttons](https://github.com/decred/dcrdex/pull/1921#discussion_r1014357395) instead, in a followup PR. That would capture the spirit and get us closer to expected behavior by wider convention.

> That's caused by another issue (probably unrelated to this PR) I've observed - mid-gap-rate (0.001354 displayed on depth-chart on your screenshot above) and Price (0.000 top left corner on your screenshot above) values shown in UI diverge,

Since #1781, spot price is based on last match, not mid-gap rate.

norwnd:
> Let's scrap the pre-fill in this PR

Okay, https://github.com/decred/dcrdex/pull/1921/commits/1873926bc2adec303a09e08b3abeaf6f65b7b268 should do it.

> spot price is based on last match, not mid-gap rate.

I see, Thanks for clarification!

client/webserver/site/src/js/markets.ts
norwnd: I couldn't find way to remove that hover message, so having `min` & `step` is probably best trade-off here to at least correct what it displays.

And there is another reason as well, I just remembered, which is handling up/down key-press handling correctly:
- for `lotField ` the values I'm setting just happen to match the defaults (but sometimes it is worth to specify defaults explicitly so we know what they are, and that it is intentional that we are using these defaults) 
- for `qtyField` in makes sense to increment/decrement in, say, 20s than by 1 when `lotSize=20` (when using up/down keys on keyboard)
- and similar applies for `rateField` ^

Added [commit](https://github.com/decred/dcrdex/pull/1921/commits/3d5afd10867b39c237c2e4a5eedcb95085f06b83) to document some reasoning for this.

buck54321:
I think there's a lot of good stuff here, but there have been quite a few regressions, and I think I may have failed to see the forest through the trees. I want to step back and examine the goals and outcomes of this effort. Goals:
> 1. `Price`, `Quantity` fields should have reasonable defaults when `/markets` page loads (and when garbage user input gets reset), so that less clicks/actions are required by the user to place his limit order
> 2. validation for `Price`, `Quantity` fields should be more responsive, so that the user has a better sence of why his inputs don't make sense
> 3. `Place order to buy ...` button press should fail right away when `Total` exceeds `Max Buy`, instead of allowing the user to proceed only to fail later on

Arguably, those could have been separate PRs, but if they're small enough, catching a few at a time is OK.

I think the general consensus on **number 1** is that it's questionable to prefill any fields for the user, and we've seen a number of regressions from the attempts to resolve this mostly abandoned goal (https://github.com/decred/dcrdex/pull/1921#discussion_r1017824753, https://github.com/decred/dcrdex/pull/1921#issuecomment-1354054340).

Effects:
1. **Goal number 2** is complete. This is the primary benefit afaiac. We probably could have done this without reworking all of the event handlers so completely.
2. **Goal number 3** works, but we're changing user fields beyond just rounding here, which is questionable.
3. Pre-parses input to the `reportDepthClick`, but `reportDepthClick` already ran `rateFieldChanged`, so there's no net change here. 
4. Parsing the input and change events for input value normalization, instead of, in the case of the rate field, the change as a trigger for normalization, and keyup as a trigger for max order estimate updates. I think the goal here was to give live feedback for invalid values, but we're not actually providing visual feedback on input anymore. Now we're just check input validity to e.g. hide the max order estimate value. There is a lot of refactoring here, and I think there might be some critical improvements, but at this point, it's not entirely clear if any large-scale changes were needed.
5. Fixed a bug where the `orderTotalPreview` for market orders was not being set correctly
6. The market sell order form is separated from the limit form, meaning a bunch of new inputs and essentially duplicated elements and the code that goes with that. A lot of the code changes come from this decision, and it's not clear to me that it was needed in the first place. Maybe you can remind me of the reasoning here. Regardless, no net changes to the UI/UX from this, right?
7. Sets the min and step attributes of various number inputs, resulting in better messaging and enabling the use of arrow keys on number inputs.

The net effects are actually minimal, and there's net +600 LOC. I don't know, maybe they are all needed, but with the labyrinth this PR has become, it's hard to see. 

I hate to ask, but I think the best path forward is to pull a few of the most important pieces out as separate, highly-focused, **minimally-implemented** PRs. I stress **minimally-implemented**. Challenge yourself to find solutions that slide right in to the existing system. Keep the diff small. We can see what's left after that.

norwnd:
While there have been some regressions during development, by this time they all should already have been fixed, whatever slipped through shouldn't be critical since I tested placing orders quite a bit. The thing is, pulling apart this PR into smaller ones will probably introduce more bugs than prevent (and, each of them will require more-or-less similar testing efforts, multiplying them immensely). Besides, all the code is targeting roughly same things, only their different _aspects_ (validation, "smart" adjustment, highlighting feedback, occasional bug, etc), that means they **all** depend on roughly same refactoring of those `handlers` (which is the bulk of this PR, and is why it turned out to be soo long; I had to untangle some of the existing code to make it extendable to support added "features").

Answering more specificaly for each goal,

0) Pre-filling fields (**number 1**) is already removed, that's not a deal-breaker and can easily be added later on, if the need arises.

1) Maybe there is a way to implement ALL those minor adjustments(improvements) I've made building directly on existing code in `master`, but that would likely be more of a add-hoc solution riddled with pitfalls and work-arounds, I don't like putting square peg in round hole (it's not that I didn't try, you can probably see in the commit-history how it all unfolded) - so, I don't think it makes much difference whether its 100 or 1000 lines of code, as long as complexity of that code is sufficiently managed.

2) I probably wasn't clear about it, but **Goal number 3** was just **an example** of the behavior that could be improved in an **obvious** way. It doesn't end there, and was just meant to show there are things I'd like to improve (things I can add, that I could easily remove if they are contentious - just like we did with "input defaults"). That would be another reason for the bloat in this PR. I wouldn't be able (probably not worth spending time on it) to find and list all the "minor bugs / suboptimal behavior" that I stumbled upon when writing this code, here is [one other example](https://github.com/decred/dcrdex/pull/1921#discussion_r1030506507) though, I think there were more.

3) I think that was just necessary to adapt it to the refactoring I've mentioned above

4) similar to 3),

> I think the goal here was to give live feedback for invalid values

Correct, currently in this PR I don't do error-reporting for `input` events. But, that potentially could change in the future, having a code that allows for it is just "a bonus" imo.

> but we're not actually providing visual feedback on input anymore.
> Now we're just check input validity to e.g. hide the max order estimate value.

We do check that, and while on its own it might not justify the whole refactoring, it is just another "bonus" we get, along with the option to easily extend this behavior in the future.

5) Yes, that's probably expected when you fiddle with forms like this (many different handles to adjust, quite a bit of inter-dependant code) for a while.

6) That was kicked-off by the bug mentioned in 5), I think, maybe there were more - I just couldn't remember. The reasoning was, keep code-complexity manageable (having all 4 types of order managed by same set of fields can get out of hand; so I thought splitting Market-related stuff into its own code would be optimal, I didn't change much of "Market Buy" because the scope started bloating already, but it probably could benefit from this as well in the future), address that 5) bug, also added `Max Sell` section for "Market Sell" order type, which is probably just as useful as on limit form.

7) Correct. Up-key doesn't even work on `Quantity` for me on `master`.

> I hate to ask, but I think the best path forward is to pull a few of the most important pieces out as separate, highly-focused, minimally-implemented PRs. I stress minimally-implemented. Challenge yourself to find solutions that slide right in to the existing system.

If there is a particular concern you'd wanna addressed, I'd be happy to.

I don't think I realistically can/want to go this route. Pulling this code apart is gonna accomplish nothing that's worth the effort. There is nothing in it for me, I don't want  to "trade" for each tiny piece being included or not, if **my vision overall** doesn't **match yours** then there is no point, if it does - asking me to work with my hands tied around my back is a waste.

norwnd:
Rebased & force-pushed, 

> When pushing updates to reviewable PRs, please try to ensure there is always a way to see just what was changed since the last push. Either add commits, which we will squash before merge

I think I should have asked about force-push first, still learning how Github treats diffs, didn't think it will update all the commits before [the last one I added](https://github.com/decred/dcrdex/pull/1921/commits/21862fd8882f408b517b52f69dadc319106afb88) (since none of them changed) after rebase.

buck54321:
It probably seems like I’m on a power trip, but I assure you that’s not the case. There are some nice improvements here that I really want to get in. But we’re not going to use them as a Trojan horse to get a bunch of unnecessary changes in just because you want it styled your way. That's a recipe for disaster. I won’t merge a 600 LOC PR without additional approvals, and I can’t speak for others, but my guess is that this isn’t getting reviewed because nobody can pin down what all of the changes are fixing. Most of them aren’t fixing anything.

I'm happy to spin off a minimalist PR that has all the fixes, if that helps. I swear that my only goal is to get your fixes in.

chappjc (commented):
Now that we've been through several iterations and it's functionally what we want (afaict), I'm happy to approve after we squash away the fixups and stuff to a few commits with model commit messages.  The way I'd like to merge is with multiple independent commits (assuming there are independent bits here).

I agree with @buck54321 that it has been hard to see through all the extraneous changes, and that's why review has been sluggish on this.  I *think* it's more minimal now, so let's massage the branch to the point that it's ready to click "Rebase and merge" and we'll end up with clear and well commented commits on master.

norwnd:
The only thing that's not highly-tied/related to Limit-form changes in this PR is Market-order form being split-off into it's own view (HTML + separate JS handling), the rest of the changes are pretty much adjustments needed to make Limit-form work smoothly.

The risk of making a mistake (introducing bugs) splitting off this Market-order stuff into separate commit is pretty high compared to the benefit of simpler-diff, I think, 
and to do that I'd need to make **old** Market-form work with **new** Limit-form (which is a challenge in itself, if even possible to do in a reasonably clean way) because we probably don't want non-working commit in **master**; and we'll be just throwing away this effort in a follow-up Market-order stuff commit, assuming you are on board with this split-off, better described [here](https://github.com/decred/dcrdex/pull/2043):

> having all 4 order types: Limit/Market+Buy/Sell managed by same view seems have resulted into tight-coupling of some code in markets.ts. To keep code-complexity manageable, I've split off Market-Sell order stuff into its own separate view (Limit Buy/Sell forms are almost identical, can't say same for Market orders). Market-Buy form was already handled by separate view (see `MARKET BUY ORDER` in `markets.tmpl`)

So, while I do appreciate the security concerns you guys have raised, I think the most practical way to go about it is to take our time and carefully review what's implemented here in one piece. Since (from what I see in master) you don't squash PR-commits before merging, I've pushed a new PR: https://github.com/decred/dcrdex/pull/2043 to combine these 23 commits into 1, summarising in commit message all the things it changes (this PR can be closed in favor of https://github.com/decred/dcrdex/pull/2043, I just didn't want to force-rewrite commit history in this one, in case you'd want to preserve it for some reason).

norwnd:
Superseded by https://github.com/decred/dcrdex/pull/2043.

PR #1923, opened 2022-10-25
UI: template parsing doesn't work with dash

norwnd:
I've noticed this template-parsing issue, and it seems it's been introduced recently (since in Decrediton 1.7.5 I don't observe it),

**On master:**

image

**With the changes from this PR:**

image

### DIFF
--- a/client/webserver/locales/parse.go
+++ b/client/webserver/locales/parse.go
@@ -2,7 +2,7 @@ package locales
 
 import "regexp"
 
-var keyRegExp = regexp.MustCompile(`\[\[\[([a-zA-Z0-9 _.-:]+)\]\]\]`)
+var keyRegExp = regexp.MustCompile(`\[\[\[([a-zA-Z0-9 _.\-:]+)\]\]\]`)
 
 // Tokens returns a slice of tuples (token, key), where token is just key with
 // triple square brackets, and could also be constructed simply as
--- a/client/webserver/locales/parse_test.go
+++ b/client/webserver/locales/parse_test.go
@@ -12,35 +12,18 @@ func TestTokens(t *testing.T) {
 {{template "top" .}}
 {{$passwordIsCached := .UserInfo.PasswordIsCached}}
 <div id="main" data-handler="settings" class="text-center py-5 overflow-y-auto">
-  <span class="settings-gear ico-settings"></span><br>
   <div class="settings">
-    <div class="form-check">
-      <input class="form-check-input" type="checkbox" value="" id="darkMode"{{if .UserInfo.DarkMode}} checked{{end}}>
-      <label class="form-check-label" for="darkMode">
-        [[[Dark Mode]]]
-      </label>
-    </div>
     <div class="form-check">
       <input class="form-check-input" type="checkbox" value="" id="showPokes"{{if .UserInfo.ShowPopups}} checked{{end}}>
       <label class="form-check-label" for="showPokes">
         [[[Show pop-up notifications]]]
       </label>
     </div>
-    <div {{if not .UserInfo.Authed}} class="d-hide"{{end}}>
-    <span>Fiat Currency: </span><span id="fiatCurrency">{{.FiatCurrency}}</span>
-    </div>
     <div id="fiatRateSources" {{if not .UserInfo.Authed}} class="d-hide"{{end}}>
     <span class="mb-2" data-tooltip="[[[fiat_exchange_rate_msg]]]">
      [[[fiat_exchange_rate_sources]]]:
     <span class="ico-info"></span>
     </span>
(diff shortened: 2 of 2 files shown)


client/webserver/locales/parse.go
```diff
@@ -2,7 +2,7 @@ package locales
 
 import "regexp"
 
-var keyRegExp = regexp.MustCompile(`\[\[\[([a-zA-Z0-9 _.-:]+)\]\]\]`)
+var keyRegExp = regexp.MustCompile(`\[\[\[([a-zA-Z0-9 _.\-:]+)\]\]\]`)
```
chappjc, line 5: Good catch.  With the `:` at the end it (added last week), it must have started interpreting the `-` to mean a character range.

https://regex101.com/r/aDB2hS/1

image

image

client/webserver/locales/parse_test.go
```diff
@@ -0,0 +1,210 @@
...
+		"pw_for_seed",
+		"Password",
+		"Show Me",
+		"dont_share",
+	}
+	sort.Slice(wantKeys, func(i, j int) bool {
+		return wantKeys[i] < wantKeys[j]
+	})
+
+	got := Tokens(tmplFile)
```
chappjc, line 184: I'm not so sure the settings html template is the best test.  There's are dozens of lines of irrelevant text, plus the tokens being tested don't include all the characters in `keyRegExp`.  How about a more concise and contrived template, with more keys to test the full intended char set?

client/webserver/locales/parse.go
ukane-philemon: > https://regex101.com/r/aDB2hS/1

phew, didn't notice the new interpretation. Good catch.

client/webserver/locales/parse_test.go
norwnd: I used this html template because it is a ready-made real-world example, I don't think for this particular test it's an issue to have all this text included,

added a harder pattern as per your suggestion, 

but if you have some specific pattern in mind - I can update to that, of course

client/webserver/locales/parse.go
norwnd: I don't think all those rules can (or should) be remembered : ) that's what tests are for

chappjc: Indeed.  Thanks for the harder pattern.  Now we do test all the rules.

client/webserver/locales/parse_test.go
```diff
@@ -0,0 +1,213 @@
...
+
+    {{- /* NEW WALLET */ -}}
+    <form class="d-hide" id="newWalletForm">
+      {{template "newWalletForm" }}
+    </form>
+
+    {{- /* CHANGE APP PASSWORD */ -}}
+    <form class="d-hide" id="changeAppPWForm">
+      {{template "changeAppPWForm"}}
+    </form>
```
chappjc, line 100: This entire stretch is what makes no sense to be in the test data IMO.  Sure it's "real world" but I don't see how it adds anything.  The pattern just repeats 6 times.

client/webserver/locales/parse.go
chappjc: However, we don't have a negative test (not matching when something not in the intended set is included), which would have caught that this was being treated as a range.

client/webserver/locales/parse_test.go
norwnd: Ok, trimmed it down

client/webserver/locales/parse_test.go
```diff
@@ -0,0 +1,102 @@
...
+{{define "settings"}}
+{{template "top" .}}
+{{$passwordIsCached := .UserInfo.PasswordIsCached}}
+<div id="main" data-handler="settings" class="text-center py-5 overflow-y-auto">
+  <div class="settings">
+    <div class="form-check">
+      <input class="form-check-input" type="checkbox" value="" id="showPokes"{{if .UserInfo.ShowPopups}} checked{{end}}>
+      <label class="form-check-label" for="showPokes">
+        [[[Show pop-up notifications]]]
+      </label>
```
ukane-philemon, line 20: For this test, the only necessary HTML elements we need are those with `[[[whatever-token-here]]]`, the rest can be trimmed off, I think.

norwnd: Well, the reason why I want to leave some HTML code in is to make sure this pattern doesn't catch any extra HTML code that it shouldn't,

I really see no benefit to trimming it down even further

client/webserver/locales/parse.go
chappjc: Great, thanks for adding a negative test that does include the triple brackets.  The error prone bits are clearly what's inside the brackets.

PR #1928, opened 2022-10-28
client/ws: fix panic in ws communications

LasTshaMAN:
I've encountered this `panic` once (`dexc` doesn't crash since there is a `recover()` in place, but it did show up in logs, and I had to refresh web page to get it to working state):

```
2022-10-19 12:07:10.545 [CRT] WEB[WS][[::1]:52108]: Uh-oh! Panic while handling message from [::1]:52108.
​
Message:
​
&msgjson.Message{Type:0x1, Route:"loadcandles", ID:0x2, Payload:json.RawMessage{0x7b, 0x22, 0x68, 0x6f, 0x73, 0x74, 0x22, 0x3a, 0x22, 0x64, 0x65, 0x78, 0x2d, 0x74, 0x65, 0x73, 0x74, 0x2e, 0x73, 0x73, 0x67, 0x65, 0x6e, 0x2e, 0x69, 0x6f, 0x3a, 0x37, 0x32, 0x33, 0x32, 0x22, 0x2c, 0x22, 0x62, 0x61, 0x73, 0x65, 0x22, 0x3a, 0x34, 0x32, 0x2c, 0x22, 0x71, 0x75, 0x6f, 0x74, 0x65, 0x22, 0x3a, 0x30, 0x2c, 0x22, 0x64, 0x75, 0x72, 0x22, 0x3a, 0x22, 0x35, 0x6d, 0x22, 0x7d}}
​
Panic:
​
runtime error: invalid memory address or nil pointer dereference
​
Stack:
​
goroutine 181 [running]:
runtime/debug.Stack()
	/usr/local/go/src/runtime/debug/stack.go:24 +0x65
decred.org/dcrdex/dex/ws.(*WSLink).handleMessage.func1()
	/Users/iurii/dcrdex/dex/ws/wslink.go:228 +0x58
panic({0x1d30860, 0x2f218e0})
	/usr/local/go/src/runtime/panic.go:838 +0x207
decred.org/dcrdex/client/websocket.wsLoadCandles(0xc0005d2000, 0xc000146c00, 0xc00166ebc0)
	/Users/iurii/dcrdex/client/websocket/websocket.go:334 +0x1a0
decred.org/dcrdex/client/websocket.(*Server).handleMessage(0xc0005d2000, 0xc00166ebc0?, 0xc00166ebc0)
	/Users/iurii/dcrdex/client/websocket/websocket.go:196 +0x1bf
decred.org/dcrdex/client/websocket.(*Server).connect.func1(0xc00166ebc0?)
	/Users/iurii/dcrdex/client/websocket/websocket.go:132 +0x2a
decred.org/dcrdex/dex/ws.(*WSLink).handleMessage(0xc0005a4bd0, 0xc00166ebc0)
	/Users/iurii/dcrdex/dex/ws/wslink.go:234 +0x73
decred.org/dcrdex/dex/ws.(*WSLink).inHandler(0xc0005a4bd0, {0x27bab90, 0xc0004bba80})
	/Users/iurii/dcrdex/dex/ws/wslink.go:285 +0x1f7
created by decred.org/dcrdex/dex/ws.(*WSLink).Connect
	/Users/iurii/dcrdex/dex/ws/wslink.go:189 +0x2a5
```

Here is my full log (expires in 21 days) - https://pastebin.mozilla.org/bEXK5B56. Decoded message looks like this, if you need it: https://go.dev/play/p/ndDCmerq8HE

Here is how I see what happens (pulled out relevant log lines, added comments):
```
...
2022-10-19 12:06:40.544 [TRC] WEB[WS]: message of type 1 received for route loadmarket
# but 30s after
2022-10-19 12:07:10.545 [ERR] WEB[WS]: error getting order feed: timed out waiting for 'orderbook' response
# Because of this error ^ `cl.feed` will be nil, and when `loadcandles` msg below gets handled (UI sends it immediately after `loadmarket`, but it gets queued on the dexc side since all WS messages are sequential)  it results in nil pointer dereference
2022-10-19 12:07:10.545 [TRC] WEB[WS]: message of type 1 received for route loadcandles
# panic happens here
...
```

### DIFF
--- a/client/websocket/websocket.go
+++ b/client/websocket/websocket.go
@@ -54,6 +54,17 @@ func newWSClient(addr string, conn ws.Connection, hndlr func(msg *msgjson.Messag
 	}
 }
 
+func (cl *wsClient) shutDownFeed() {
+	cl.feedMtx.Lock()
+	defer cl.feedMtx.Unlock()
+
+	if cl.feed != nil {
+		cl.feed.loop.Stop()
+		cl.feed.loop.WaitForShutdown()
+		cl.feed = nil
+	}
+}
+
 // Core specifies the needed methods for Server to operate. Satisfied by *core.Core.
 type Core interface {
 	SyncBook(dex string, base, quote uint32) (core.BookFeed, error)
@@ -125,7 +136,7 @@ func (s *Server) HandleConnect(ctx context.Context, w http.ResponseWriter, r *ht
 func (s *Server) connect(ctx context.Context, conn ws.Connection, addr string) {
 	s.log.Debugf("New websocket client %s", addr)
 	// Create a new websocket client to handle the new websocket connection
-	// and wait for it to shutdown.  Once it has shutdown (and hence
+	// and wait for it to shut down.  Once it has shutdown (and hence
 	// disconnected), remove it.
 	var cl *wsClient
 	cl = newWSClient(addr, conn, func(msg *msgjson.Message) *msgjson.Error {
@@ -151,12 +162,7 @@ func (s *Server) connect(ctx context.Context, conn ws.Connection, addr string) {
 	s.clientsMtx.Unlock()
 
 	defer func() {
-		cl.feedMtx.Lock()
-		if cl.feed != nil {
-			cl.feed.loop.Stop()
-			cl.feed.loop.WaitForShutdown()
-		}
-		cl.feedMtx.Unlock()
+		cl.shutDownFeed()
 
 		s.clientsMtx.Lock()
 		delete(s.clients, cl.cid)
@@ -303,12 +309,9 @@ func loadMarket(s *Server, cl *wsClient, req *marketLoad) (*bookFeed, *msgjson.E
 		return nil, msgjson.NewError(msgjson.RPCOrderBookError, errMsg)
 	}
 
+	cl.shutDownFeed()
+
 	cl.feedMtx.Lock()
-	defer cl.feedMtx.Unlock()
-	if cl.feed != nil {
-		cl.feed.loop.Stop()
-		cl.feed.loop.WaitForShutdown()
-	}
 	cl.feed = &bookFeed{
 		BookFeed: feed,
 		loop:     newMarketSyncer(cl, feed, s.log.SubLogger(name)),
@@ -316,6 +319,7 @@ func loadMarket(s *Server, cl *wsClient, req *marketLoad) (*bookFeed, *msgjson.E
 		base:     req.Base,
 		quote:    req.Quote,
 	}
+	cl.feedMtx.Unlock()
 
 	return cl.feed, nil
 }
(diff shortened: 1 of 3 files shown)


client/websocket/websocket.go
```diff
@@ -331,7 +335,12 @@ func wsLoadCandles(s *Server, cl *wsClient, msg *msgjson.Message) *msgjson.Error
...
 	cl.feedMtx.RUnlock()
-	if feed.host != req.Host || feed.base != req.Base || feed.quote != req.Quote {
+	// If market hasn't been initialized/chosen yet (client should do it in a separate
+	// 'loadmarket' request), or if client wants to change currently chosen market (requesting
+	// candles for market that's different from currently chosen implies that) - we can
+	// try to load it here.
+	if feed == nil ||
+		(feed.host != req.Host || feed.base != req.Base || feed.quote != req.Quote) {
 		var msgErr *msgjson.Error
 		feed, msgErr = loadMarket(s, cl, &req.marketLoad)
```
LasTshaMAN, line 345: One additional consideration, 

with the "lazy market loading/reloading logic" implemented in this PR do we even need to support (or issue from UI-client) `loadmarket` msg ?

client/websocket/websocket.go
```diff
@@ -303,19 +309,17 @@ func loadMarket(s *Server, cl *wsClient, req *marketLoad) (*bookFeed, *msgjson.E
 		return nil, msgjson.NewError(msgjson.RPCOrderBookError, errMsg)
 	}
 
+	cl.shutDownFeed()
+
 	cl.feedMtx.Lock()
```
chappjc, line 314: nACK.  This introduces an extra mutex lock/unlock cycle, and makes the feed shutdown and reassignment no longer atomic.
Avoiding three lines of code duplication is a poor trade off with the loss of inefficiency and atomicity.
Also, the `return cl.feed, nil` is now outside of the critical section.

chappjc (changes requested):
Thanks for looking into it.

client/websocket/websocket.go
LasTshaMAN: When I remove [this line](https://github.com/LasTshaMAN/dcrdex/blob/86db1f4e2a8ebb52c5041c3c16b7618ab742bd03/client/webserver/site/src/js/markets.ts#L795-L795), `/markets` page seems to be working same way without it,

that's because it loads candlesticks as well, and I thought maybe having markets page without candlesticks doesn't make much sense - hence maybe remove it; but maybe there is a use-case for a market without candlesticks (like showing order-book depth without candles)

client/websocket/websocket.go
LasTshaMAN: > This introduces an extra mutex lock/unlock cycle

Didn't think this would make any difference since it's not in hot path.

> makes the feed shutdown and reassignment no longer atomic

Didn't really think this could result into issues here, since all requests by client are sequentially ordered; however that's a good point to keep in mind

> Also, the return cl.feed, nil is now outside of the critical section.

I should probably do self-reviews for my own PRs first ;)

LasTshaMAN: Updated to address stuff mentioned above, hopefully without introducing new issues.

client/websocket/websocket.go
chappjc: Oh that's an interesting point!  Before the latest markets page redesign, candlesticks weren't shown unless you toggled to them.  But now it does `this.loadCandles()` -> `this.requestCandles()` -> `ws.request('loadcandles',...` which would be seem to be redundant with the 'loadmarket' you linked to.  Will defer to @buck54321 on that, but I think you're correct.

chappjc (approved):

JoeGruffins (approved):

Issue #1816, opened 2022-08-23
client/asset/dcr: restart required after restoring native dcr wallet with funds


chappjc:
When I reinitialized dexc with a native DCR wallet that had funds, it showed 0 balance even after scan/recover had completed.  I had to restart dexc to have the proper balance.

Refreshing the UI did not help.
<s>New DCR blocks did not change the balance either.</s> a new block seems to trigger the proper balance.

chappjc:
A related issue is that sync status for the native DCR wallet seems to neglect the wallet rescan that comes after block header and filter sync.  That is, when you create the wallet it will get to 100% synced at this point:

```
2022-08-24 09:31:42.154 [INF] SPV: Connected 2000 blocks, new tip 000000001477f201395d9fdbdbf08b4576235f17ef8dbe3dd0c0fee47385f33c, height 974000, date 2022-08-20 06:27:58 -0500 CDT
2022-08-24 09:31:42.438 [DBG] SPV: Fetched 2000 new header(s) ending at height 976000 from [2001:19f0:5:33c5:5400:2ff:fee0:b1c3]:19108
2022-08-24 09:31:42.706 [INF] SPV: Connected 2000 blocks, new tip 00000000208e62d3981c876c06f8983b8e3cc4ff528d13eac2b83bd5e7110660, height 976000, date 2022-08-23 02:55:29 -0500 CDT
2022-08-24 09:31:42.921 [DBG] SPV: Fetched 952 new header(s) ending at height 976952 from [2001:19f0:5:33c5:5400:2ff:fee0:b1c3]:19108
2022-08-24 09:31:43.046 [INF] SPV: Connected 952 blocks, new tip 00000000ba45d21fdc8eea2e3c40977545304887973e6ac3d9c8476e8adc50eb, height 976952, date 2022-08-24 09:30:09 -0500 CDT
```

However, this is where it's recsan starts:

```
2022-08-24 09:31:43.046 [INF] SPV: Connected 952 blocks, new tip 00000000ba45d21fdc8eea2e3c40977545304887973e6ac3d9c8476e8adc50eb, height 976952, date 2022-08-24 09:30:09 -0500 CDT
2022-08-24 09:31:43.102 [INF] WLLT: Discovering used addresses for 1 account(s)
2022-08-24 09:32:12.445 [INF] WLLT: Account 0 next child indexes: external:2 internal:1
2022-08-24 09:32:12.445 [INF] WLLT: Updating DB with discovered addresses...
[...]

PR #1931, opened 2022-10-28
core,btc,app: Manage SPV wallet peers

martonp:
This change allows the user to see the peers the BTC SPV wallets are connected to, and add new peers.
- `client/asset`: Adds a new `PeerManager` interface with three methods: `Peers`, `AddPeer` and `RemovePeer`. Wallets that implement this interface have the trait `WalletTraitPeerManager`.
- `client/core`: Adds new methods that call the `PeerManager` methods.
- `client/btc`: 
  - A `SPVPeerManager` type is added which manages a key value db and a neutrino chain service. When the user adds a new peer, it is stored in the key value db, and will be automatically connected the next time the wallet is opened. 
  - The `BTCWallet` interface now has three new functions,  `Peers`, `AddPeer` and `RemovePeer`, and the BTC, BCH, and LTC SPV wallets each use a `SPVPeerManager` in order to handle these new functions.
- `app`:
Screen Shot 2022-10-28 at 1 21 28 PM

Screen Shot 2022-11-03 at 6 07 58 PM


go.mod
```diff
@@ -73,6 +73,8 @@ require (
 	lukechampine.com/blake3 v1.1.7
 )
 
+replace github.com/lightninglabs/neutrino v0.14.2 => /Users/marci/git/neutrino
```
chappjc, line 76: OK to proceeding with an update to the required to point to 792af8548c14a112dafba06f4d8081436d9ec8ca
It's just your commit past the v0.14.2 tag.

chappjc:
I really like the Manage Peers view.  Good stuff @martonp!

martonp:
@chappjc This will be ready to review once https://github.com/dcrlabs/neutrino-bch/pull/9 and https://github.com/dcrlabs/neutrino-ltc/pull/6 are merged.

client/asset/btc/spv_peer_manager.go
```diff
@@ -0,0 +1,236 @@
...
+
+	defaultPeers []string
+
+	log dex.Logger
+}
+
+// Peers returns the list of peers that the wallet is connected to. It also
+// returns the peers that the user added that the wallet may not currently
+// be connected to.
+func (w *SPVPeerManager) Peers() ([]*asset.WalletPeer, error) {
```
buck54321, line 72: Why `w`?

client/asset/btc/spv_wrapper.go
```diff
@@ -158,10 +165,25 @@ func (s *btcChainService) Peers() []SPVPeer {
...
+	if err != nil {
+		return err
+	}
+	conn, err := net.Dial("tcp", peer.Addr())
+	if err != nil {
+		return err
+	}
+	peer.AssociateConnection(conn)
+	serverPeer.Peer = peer
+	return nil
```
buck54321, line 180: I see this part repeated in **bch** and **ltc**. Perhaps an exported helper function that accepts the proper interfaces would be usefuil?

buck54321 (commented):
Approach ACK, for sure. This is really well done.

client/asset/btc/spv_peer_manager.go
martonp: I'll update, I had these methods on the btcSPVWallet type earlier.

client/asset/btc/spv_wrapper.go
martonp: All the functions on these `spvService` types are completely repeated. There are so many different types used in these functions from the three neutrino libraries that are all the same, I feel like it's simpler just to repeat them. I can try to do something if you feel strongly about it though.

JoeGruffins (commented):
I think the red X could use a column name or a hover over message. It's probably obvious for most but I was confused if maybe it meant I was not connected at first, until I clicked it.

image

JoeGruffins:
Probably unrelated to these changes, but client won't shut down as bch is trying to sync.
```
2022-11-01 07:59:04.627 [TRC] CORE[btc][SPV]: Stopping neutrino DB.
2022-11-01 07:59:04.644 [INF] CORE[btc][SPV]: SPV wallet closed
2022-11-01 07:59:04.644 [INF] CORE: Locking DCR wallet
2022-11-01 07:59:04.644 [ERR] CORE: Failed to lock DCR wallet: wallet locked: account is already locked
2022-11-01 07:59:04.644 [INF] CORE[dcr][SPV]: Unloading wallet
2022-11-01 07:59:04.647 [INF] CORE[dcr][SPV]: SPV wallet closed
2022-11-01 07:59:04.647 [INF] CORE: Locking BCH wallet
2022-11-01 07:59:04.647 [INF] CORE[bch][SPV]: Unloading wallet
```

10 minutes now... giving up

chappjc:
> Unrelated to this pr. Will make an issue after some investigation.

Will keep an eye out for the issue.  There are a number of neutrino-related issues that only really pop out with testnet or mainnet where sync can take actual time. 

1. Wallet sync progress hits 100% way too soon: https://github.com/decred/dcrdex/issues/1816#issuecomment-1225816026  It was hung during that too of course.
2. When the wallets appear to be hung, I always find that in the assetdb neutrino.log it is requesting cfilters, nice and slow, usually 1-by-1 instead of batch.  That's something that can be improved with https://github.com/lightninglabs/neutrino/pull/236, which I tested at one point.
3. Even with the above fix, neutrino sync is **non-cancellable**.  We'd have to figure out why it doesn't respond to shutdown requests when in sync.

client/asset/btc/spv_wrapper.go
buck54321: Oh yeah. It'd be messy.

client/asset/btc/spv_peer_manager.go
```diff
@@ -0,0 +1,236 @@
...
+		for _, peerResolvedName := range peer.resolvedNames {
+			for _, newPeerResolvedName := range resolvedNames {
+				if peerResolvedName == newPeerResolvedName {
+					return name, true
+				}
+			}
+		}
+	}
+	return "", false
+}
```
buck54321, line 125: I think address resolution is unnecessary once you do @JoeGruffins suggestion. Should be able to clean things up a little bit.

client/webserver/site/src/html/wallets.tmpl
```diff
@@ -2,6 +2,9 @@
 {{$passwordIsCached := .UserInfo.PasswordIsCached}}
 {{template "top" .}}
 <div id="main" data-handler="wallets" class="walletspage flex-grow-1 position-relative">
+  <span class="me-2 connection ico-connection d-hide" style="padding-left:30%" id="connectedIconTmpl"></span>
+  <span class="me-2 errcolor ico-disconnected d-hide" style="padding-left:30%" id="disconnectedIconTmpl"></span>
+  <span class="me-2 errcolor ico-cross d-hide pointer" style="padding-left:30%" id="removeIconTmpl"></span>
```
buck54321, line 7: No inline style.
image

buck54321 (changes requested):
Wallet is ready, but I'm not seeing the button.

image

client/asset/btc/spv_peer_manager.go
```diff
@@ -0,0 +1,236 @@
...
+		return nil
+	})
+}
+
+// NewSPVPeerManager creates a new SPVPeerManager.
+func NewSPVPeerManager(cs PeerManagerChainService, defaultPeers []string, dir string, log dex.Logger) (*SPVPeerManager, error) {
+	db, err := kvdb.NewFileDB(filepath.Join(dir, "peers.db"), log.SubLogger("PEERDB"))
+	if err != nil {
+		return nil, fmt.Errorf("failed to create peers db: %w", err)
+	}
```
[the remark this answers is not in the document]
chappjc: Me neither.  A json.peers file that we  open/write/close the JSON on each call (AddPeer, RemovePeer, Peers) rather than holding it open would be preferable.  If any of those fail, the call was a no-op and it logs an error instead of causing a fatal wallet error.  Also it keeps `OpenSPVWallet` free of one error possibility if `walletConstructor` cannot error.

client/asset/btc/spv_wrapper.go
```diff
@@ -158,10 +165,25 @@ func (s *btcChainService) Peers() []SPVPeer {
 	return peers
 }
 
+func (s *btcChainService) AddPeer(addr string) error {
```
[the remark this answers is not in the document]
martonp: Yeah thanks, I've used `ConnectNode` now.

client/asset/btc/spv_peer_manager.go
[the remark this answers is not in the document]
martonp: I've updated to use a json file.

client/asset/btc/spv_peer_manager.go
martonp: I still have the address resolution.. I wrote the reason in the comment here:
https://github.com/decred/dcrdex/blob/52d2b69fd4128c3d69206224be7c02e78e9c0cba/client/asset/btc/spv_peer_manager.go#L103

Another option would be to update the neutrino code to return the resolved address it is using from `ConnectNode`.

martonp:
@buck54321 I hadn't updated the cache busters. You should be able to see the button now.

JoeGruffins (commented):
Hmm, it was working well before the last changes. Not sure what happened but now unable to add peers.

The neutrino logs show:
```
2022-11-04 12:23:22.381 [DBG] NTRNO: Peer 127.0.0.1:28333 is banned for another 23h4m36.618483257s
2022-11-04 12:23:22.382 [DBG] NTRNO: Peer 127.0.0.1:28336 is banned for another 23h4m36.617422658s
2022-11-04 12:23:22.384 [DBG] NTRNO: Peer 127.0.0.1:28335 is banned for another 23h4m36.615500595s
2022-11-04 12:29:13.708 [DBG] NTRNO: Peer 127.0.0.1:28337 is banned for another 22h58m45.291805946s
```

And my bchd has logs that correspond with attempting to add:
```
2022/11/04 12:29:13 http: TLS handshake error from 127.0.0.1:32816: EOF
```

Also, removing peers shows an error, but also seems to work if I reload the page:
image

martonp:
@JoeGruffins Seems like your node was banned by neutrino for some reason, maybe due to the node not serving the compact filters? I'll add a reconnect button on the screen that will unban and attempt to connect.

client/asset/btc/spv_peer_manager.go
```diff
@@ -0,0 +1,308 @@
...
+	err = json.Unmarshal(content, &peers)
+	if err != nil {
+		return nil, err
+	}
+
+	return peers, nil
+}
+
+// loadSavedPeersFromFile replaces the contents of dexc-peers.json.
+func (w *SPVPeerManager) writeSavedPeersToFile(peers map[string]peerSource) error {
```
JoeGruffins, line 187: loadSavedPeersFromFile -> writeSavedPeersToFile

JoeGruffins (commented):
It looks like my suggestion of using `s.ChainService.ConnectNode(addr, true)` is not working for me. It is not adding the node for some reason. I can look into it more.

JoeGruffins (approved):
My bch issue does not appear to be related to these changes.

chappjc:
> My bch issue does not appear to be related to these changes.

So is `ConnectNode` working for you now?

JoeGruffins:
> So is ConnectNode working for you now?

No. But @martonp was looking into it and it looks like it is a communications issue.

From @martonp:
I've synced a bch testnet node as well.. It's unable to connect because I'm getting an error here: https://github.com/gcash/bchd/blob/5964724713ae9aef9ad334dee56496f4e3bce7e6/peer/peer.go#L2321

I think this warrants a separate issue. I am planning to make one after these changes are finalized, or marton can.

chappjc:
> I've synced a bch testnet node as well.. It's unable to connect because I'm getting an error here: https://github.com/gcash/bchd/blob/5964724713ae9aef9ad334dee56496f4e3bce7e6/peer/peer.go#L2321
> 
> I think this warrants a separate issue. I am planning to make one after these changes are finalized, or marton can.

I don't understand.  The BCH SPV wallet has worked fine with the bchd harness in the past.  Are you saying that it's now broken, or it's just broken when manually adding peers in this PR?

chappjc:
> > 
> 
> I don't understand. The BCH SPV wallet has worked fine with the bchd harness in the past. Testnet too. Are you saying that it's now broken, or it's just broken when manually adding peers in this PR?

I guess you were just talking about this PR.  BCH SPV wallet still seems fine to me:

```
2022-11-17 12:24:36.890 [INF] SYNC: New valid peer 127.0.0.1:49296 (inbound) (neutrino:0.0.4-beta/)                
2022-11-17 12:24:36.891 [DBG] PEER: Received getheaders (locator 0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206, stop 00000000
00000000000000000000000000000000000000000000000000000000) from 127.0.0.1:49296 (inbound)                                                        
2022-11-17 12:24:36.891 [DBG] PEER: Sending headers (num 413) to 127.0.0.1:49296 (inbound)
2022-11-17 12:24:36.897 [DBG] PEER: Received getcfheaders (start_height=1, stop_hash=6528106e9d3a34525c603d68818aa24ad412d2c7980b617b595bcedc81e
4ee20) from 127.0.0.1:49296 (inbound)                                                                                                           
2022-11-17 12:24:36.898 [DBG] PEER: Sending cfheaders (stop_hash=6528106e9d3a34525c603d68818aa24ad412d2c7980b617b595bcedc81e4ee20, num_filter_ha
shes=413) to 127.0.0.1:49296 (inbound)                                                                                                          
2022-11-17 12:24:36.901 [DBG] PEER: Received getcfilters from 127.0.0.1:49296 (inbound)                                                         
2022-11-17 12:24:36.901 [DBG] PEER: Sending cfilter to 127.0.0.1:49296 (inbound)
```

However, bch neutrino still seems to try to discover other peers on regnet and I end up with it connecting to the same bitcoin-cash-node twice and seeing lots of apparently harmless DBG level errors.  Silenced with:

```diff
diff --git a/dex/testing/btc/base-harness.sh b/dex/testing/btc/base-harness.sh
index afa05287d..b9c565194 100755
--- a/dex/testing/btc/base-harness.sh
+++ b/dex/testing/btc/base-harness.sh
@@ -393,7 +393,7 @@ if [ ! -z "$GODAEMON" ]; then
       OMEGA_RPC_PORT=21558
 
 cat > "${NODE_CONF}" <<EOF
-addpeer=127.0.0.1:${ALPHA_LISTEN_PORT}
+connect=127.0.0.1:${ALPHA_LISTEN_PORT}
 listen=:21577
 rpcuser=user
 rpcpass=pass
@@ -401,7 +401,7 @@ rpclisten=0.0.0.0:${OMEGA_RPC_PORT}
 regtest=1
 rpccert=${OMEGA_DIR}/rpc.cert
 rpckey=${OMEGA_DIR}/rpc.key
-debuglevel=trace
+debuglevel=debug
 EOF
 
 cat > "${CLIENT_CONF}" <<EOF
```

client/asset/btc/spv_peer_manager.go
```diff
@@ -0,0 +1,308 @@
...
+
+	port, err := strconv.Atoi(strPort)
+	if err != nil {
+		return "", err
+	}
+
+	return (&net.TCPAddr{
+		IP:   ips[0],
+		Port: port,
+	}).String(), nil
```
chappjc, line 151: Without `net.TCPAddr.Zone` set, this is just `net.JoinHostPort(IP, Itoa(Port))`.  Since we've already done `SplitHostPort` at the top, it seems like we have a needless Atoi -> Itoa in the mix.

`return net.JoinHostPort(ips[0], strPort), nil`?

client/asset/btc/spv_peer_manager.go
```diff
@@ -0,0 +1,308 @@
...
+			strPort = w.defaultPort
+		default:
+			return "", err
+		}
+	}
+
+	// Tor addresses cannot be resolved to an IP, so just return onionAddr
+	// instead.
+	if strings.HasSuffix(host, ".onion") {
+		return host, nil
```
chappjc, line 130: `return addr, nil`  Otherwise we've lost any custom port, right?

client/asset/btc/spv_peer_manager.go
```diff
@@ -0,0 +1,308 @@
...
+
+	resolvedAddr, err := w.resolveAddress(addr)
+	if err != nil {
+		if initialLoad {
+			// If this is the initial load, we still want to add peers that are
+			// not able to be connected to the peers map, in order to display them
+			// to the user. If a user previously added a peer that originally connected
+			// but now the address cannot be resolved to an IP, it should be displayed
+			// that the wallet was unable to connect to that peer.
+			w.peers[addr] = &walletPeer{source: source, resolvedName: resolvedAddr}
```
chappjc, line 207: Won't `resolvedAddr` be an empty string here?

client/asset/btc/spv_peer_manager.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+
+	return nil
+}
+
+// ConnectToInitialWalletPeers connects to the default peers and the peers
+// that were added by the user and persisted in the db.
+func (w *SPVPeerManager) ConnectToInitialWalletPeers() {
+	for _, peer := range w.defaultPeers {
+		w.addPeer(peer, defaultPeer, true)
```
chappjc, line 281: Please comment on reasoning for ignoring error, if intended.

chappjc (commented):
Partial review because I want to respond to latest comment from @martonp

chappjc:
> I was able to connect to bchd on testnet as well. I removed the default peer now to avoid confusion.. the default peer was "localhost:28333" which attempted to connect to [::1]:28333, and would fail if bchd was set to listen to 127.0.0.1:2883.

Made a relevant comment in my review here https://github.com/decred/dcrdex/pull/1931#discussion_r1025644939

But ACK to removing the localhost peer, regardless of the resolution issues. Since BCH testnet4 at least has a DNS seeder (testnet4.imaginary.cash), unlike LTC testnet.  It only returns 1 node unfortunately, but it's something.

client/asset/btc/spv_wrapper.go
```diff
@@ -158,10 +163,14 @@ func (s *btcChainService) Peers() []SPVPeer {
...
+	return s.ChainService.ConnectNode(addr, true)
+}
+
 var _ SPVService = (*btcChainService)(nil)
 
 // BTCWalletConstructor is a function to construct a BTCWallet.
-type BTCWalletConstructor func(dir string, cfg *WalletConfig, chainParams *chaincfg.Params, log dex.Logger) BTCWallet
+type BTCWalletConstructor func(dir string, cfg *WalletConfig, chainParams *chaincfg.Params, log dex.Logger) (BTCWallet, error)
```
chappjc, line 173: I think we decided we could keep this signature error-free, right?

client/asset/btc/spv_peer_manager.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+
+	// Tor addresses cannot be resolved to an IP, so just return onionAddr
+	// instead.
+	if strings.HasSuffix(host, ".onion") {
+		return host, nil
+	}
+
+	// Attempt to look up an IP address associated with the parsed host.
+	ips, err := net.LookupIP(host)
```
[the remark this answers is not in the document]
martonp: We won't get any error if it resolves to the IPv6 address and there's nothing listening over there, neutrino will just keep retrying. So we'd have to separately make a connection to try it out. I think just manually resolving to `127.0.0.1` is fine.. do you think we should put a message on the screen stating that `localhost` resolves to `127.0.0.1`, and users should type `[::1]` if that's what they need?

client/asset/btc/spv_wrapper.go
martonp: Yeah I forgot to update this part.

client/asset/btc/spv_peer_manager.go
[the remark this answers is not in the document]
chappjc: No need to add messaging about how we resolve `localhost` IMO.  Resolvers are very inconsistent about `localhost`, and given we won't get an error if it incorrectly resolves to IPv6 and neutrino will retry and fail forever, we should take steps to prevent that.

> do you think we should put a message on the screen stating that localhost resolves to 127.0.0.1, and users should type [::1] if that's what they need?

Totally.  If they really want [::1], they can type that.

client/asset/btc/btc.go
```diff
@@ -1218,6 +1231,9 @@ func OpenSPVWallet(cfg *BTCCloneCFG, walletConstructor BTCWalletConstructor) (*E
 	}
 
 	spvw.wallet = walletConstructor(spvw.dir, spvw.cfg, spvw.chainParams, spvw.log)
+	if err != nil {
+		return nil, err
+	}
```
chappjc, line 1236: pls rm the leftover if return branch

client/asset/btc/spv.go
```diff
@@ -174,40 +177,38 @@ func (w *btcSPVWallet) Start() (SPVService, error) {
...
-		AddPeers:      addPeers,
-		ConnectPeers:  connectPeers,
+		// AddPeers:      addPeers,
+		// ConnectPeers:  connectPeers,
 		// WARNING: PublishTransaction currently uses the entire duration
 		// because if an external bug, but even if the resolved, a typical
 		// inv/getdata round trip is ~4 seconds, so we set this so neutrino does
 		// not cancel queries too readily.
 		BroadcastTimeout: 6 * time.Second,
+		NameResolver:     net.LookupIP,
```
chappjc, line 193: This is interesting.  What is it if not specified?  We were giving it fqdns that required resolution.

client/asset/interface.go
```diff
@@ -687,6 +697,31 @@ type RedemptionConfirmer interface {
...
+	Discovered
+)
+
+type WalletPeer struct {
+	Addr      string     `json:"addr"`
+	Source    PeerSource `json:"source"`
+	Connected bool       `json:"connected"`
+}
+
+type PeerManager interface {
```
chappjc, line 714: Could use some brief docs on these exported types and consts

client/rpcserver/handlers.go
```diff
@@ -691,6 +698,66 @@ func handleDeleteArchivedRecords(s *RPCServer, params *RawParams) *msgjson.Respo
...
+	assetID, err := checkUIntArg(params.Args[0], "assetID", 32)
+	if err != nil {
+		return usage(walletPeersRoute, err)
+	}
+
+	addr := params.Args[1]
+
+	err = s.core.RemoveWalletPeer(uint32(assetID), addr)
+	if err != nil {
+		errMsg := fmt.Sprintf("unable to get wallet peers: %v", err)
```
chappjc, line 753: The message applies to `handleWalletPeers`.  All the route constants are `walletPeersRoute`, but I think you want `addWalletPeerRoute` or `removeWalletPeerRoute` for these to latter handlers.

client/rpcserver/handlers.go
```diff
@@ -691,6 +698,66 @@ func handleDeleteArchivedRecords(s *RPCServer, params *RawParams) *msgjson.Respo
...
+func handleRemoveWalletPeer(s *RPCServer, params *RawParams) *msgjson.ResponsePayload {
+	if len(params.Args) != 2 {
+		return usage(walletPeersRoute, errors.New("expected 2 args"))
+	}
+	assetID, err := checkUIntArg(params.Args[0], "assetID", 32)
+	if err != nil {
+		return usage(walletPeersRoute, err)
+	}
+
+	addr := params.Args[1]
```
chappjc, line 749: These arg checks are a little off-pattern, where we generally make funcs in types.go like `func parseWalletPeersArgs`.  If it feels just feels like more boilerplate code, that's fine.  Just wanted to make sure you were aware.

client/asset/btc/spv.go
martonp: If it's not specified, it will still use net.LookupIP. I was just trying to make sure it didn't change in a future release of neutrino.

This is leftover from the previous version of the PR where I was trying to make sure that the resolver used in neutrino was the same as what is used in our code. I'll remove this since we're doing the resolving before we call `CreateNode`.

JoeGruffins:
> I don't understand. The BCH SPV wallet has worked fine with the bchd harness in the past. Testnet too. Are you saying that it's now broken, or it's just broken when manually adding peers in this PR?

BCH spv has never worked for me, besides one iteration of this pr.

chappjc:
> Could it be because I have ipv6 disabled on my computer? I see some discussion about that above.

That's likely.  Now that @martonp has it going to 127.0.0.1, I bet you'll be good now.

chappjc (approved):

chappjc:
There are a few unaddressed comments from other reviews, but otherwise good.  Please circle back on the comments from @JoeGruffins and @buck54321, thanks.

Issue #1933, opened 2022-11-01
client/tests: TestTryBlocksWithNotifier test failure.


JoeGruffins:
Seen on github, not locally. Not on master, but the section of the code that failed is not touched in the pr.

```
--- FAIL: TestTryBlocksWithNotifier (0.10s)
    spv_test.go:832: got block that should've been cached
```

https://github.com/decred/dcrdex/actions/runs/3368044348/jobs/5586144957

PR #1939, opened 2022-11-01
asset/btc: make spv test more robust

norwnd:
I've studied code related to this test and think the issue happens due to this line not executing "in time":

```go
// Prime the sync target to avoid the syncStatus tip send here.
atomic.StoreInt32(&spv.syncTarget, 1)
```

since [notification comes in concurrently]((https://github.com/decred/dcrdex/blob/bcfe43c8ef7c7917bb1dcce40f7ce8a63cbd9ca6/client/asset/btc/spv_wrapper.go#L552-L554)) to it, it is not guaranteed to execute before it.

Possibly closes https://github.com/decred/dcrdex/issues/1933.

### DIFF
--- a/client/asset/btc/spv_test.go
+++ b/client/asset/btc/spv_test.go
@@ -9,7 +9,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


client/asset/btc/spv_test.go
```diff
@@ -817,19 +816,22 @@ func TestTryBlocksWithNotifier(t *testing.T) {
 	// Start with no blocks so that we're not synced.
 	node.blockchainMtx.Lock()
 	node.getBlockchainInfo = &getBlockchainInfoResult{
-		Headers: 2,
+		Headers: 3,
```
norwnd, line 819: Checking 0th block case I've added to this test requires this increase to `3`.

client/asset/btc/spv_test.go
```diff
@@ -817,19 +816,22 @@ func TestTryBlocksWithNotifier(t *testing.T) {
...
+		t.Fatalf("0th block didn't send tip change update")
+	}
+
+	addBlock()
 
-	// It should not come through on the block tick, since it will be cached.
+	// It should not come through on the block tick, since it will be queued.
 	if getNote(blockTicker * 2) {
-		t.Fatalf("got block that should've been cached")
+		t.Fatalf("got non-0th block that should've been queued")
```
norwnd, line 834: `queue` is somewhat more consistent with the terminology used in code, I think

JoeGruffins (approved):
Yeah, it looks to me too like the test failed due to a race between `watchBlocks` which contains:
```
if target > 0 && atomic.SwapInt32(&w.syncTarget, target) == 0 {
		w.tipChan <- blk
	}
```
and the `atomic.StoreInt32(&spv.syncTarget, 1)` in tests to prevent the syncTarget from being the zero that triggers send on `w.tipChan`.

This fix looks good to me.

chappjc (approved):

Issue #1932, opened 2022-10-31
client: Check server config values are not too low.


JoeGruffins:
Some server config values are blindly trusted by the client when they could be detrimental to the user if too low.

They are `swapConf` and `maxFeeRate` and to a lesser extent `lotSize` and `rateStep`.

`swapConf` can be set to zero. This is sort of ok for utxo assets but no good for account assets as transactions are very easy to replace and our contracts do not work without confirmations.

`maxFeeRate` can be set too low causing swaps to fail, and possibly redeems if things align, since it is a fallback.

To a lesser extent, `lotSize` and `rateStep` being too small can cause swaps to fail due to our own code checking for dust. Related to #1610

chappjc:
Closely related, mainly pertaining to the `swapConf` values: https://github.com/decred/dcrdex/issues/1900
A global sanity check for `swapConf > 0` is reasonable in addition to user settings for the upper and lower bounds.

While we have `feeRateLimit` user setting as the _upper_ bound for the acceptable `maxFeeRate`, a lower bound can also be defined.  That value can be a user setting too, but an internal threshold enforced by the wallet seems warranted.

Strict internal lower bounds enforced internally to the wallet also make sense for `lotSize` to prevent dust swaps.  I'm not sure about `rateStep` though.  Also not sure if user-configurable settings for these two make sense since you are looking right at these values when placing an orders, so just internal wallet limits are probably good for these.

vctt94:
I'd like to work on this as it is partially included on https://github.com/decred/dcrdex/pull/1941

norwnd:
I recently looked into 1 aspect of this that bothered me most - which is "user-specified max fee rate is not respected" (meaning regardless of whatever value I enter in there - chosen fee / gas price will be different),

there are a bunch of places in code where "a fee is chosen, but not checked against user-specified max fee rate" - I was able to find/isolate the most impactful ones with this change https://github.com/decred/dcrdex/commit/1b50a4afb4babe99e07ed6f04a8f6d9c79cfab80 (it seems to be working now for sends/swaps/redeems at least),

leaving it here for future reference.

PR #1941, opened 2022-11-04
client: start adding asset config at core

### LINKED ISSUE #1900: core,ui: need asset-specific settings that are not wallet settings
I had started looking asset-specific settings to be **used by Core but not the wallets**, such as a min swap conf setting, but didn't find an nice solution.  We should do this soon though.
Related to a min confs setting per asset, I was looking into a swap conf override (i.e. used to swap or redeem *earlier* than server requires, at your own discretion).  I think we should do that too.  The user should be able to decide to e.g. redeem taker's BTC swap tx after 1 conf instead of maybe a higher server setting of 3, if they want to.

https://github.com/decred/dcrdex/pull/1866/commits/a37e6ff3eec7804b77555a8e8fc606d0f8cf868c

_Originally posted by @chappjc in https://github.com/decred/dcrdex/pull/1866#discussion_r992561844_

There are other possible settings that we don't want as part of the ConfigOpts that are defined by the wallet's package, yet are still only specific to a certain asset.

I think these settings are most logically shown on the UI with or near the wallet settings since the user doesn't care about the distinction as we understand it under-the-hood.  They just want to change a setting that applies to a given "wallet" when it's really how they wish to handle certain things pertaining to that asset.

vctt94:
Closes #1900.

I am using @chappjc 's commit https://github.com/decred/dcrdex/commit/bdb379f6ca343ce5768b7d374920272795a2da57 as reference.

I am adding `makerSwapConfOverride` and `takerSwapConfOverride` params at core, as I am using chapp's commit as base, but I believe we might want to add a trading asset config opt, which can have more fields, so we can use it to fix https://github.com/decred/dcrdex/issues/1932 as well, without the need to add each param individually.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -2152,12 +2152,13 @@ func (c *Core) assetMap() map[uint32]*SupportedAsset {
 			wallet = w.state()
 		}
 		assets[assetID] = &SupportedAsset{
-			ID:       assetID,
-			Symbol:   asset.Symbol,
-			Wallet:   wallet,
-			Info:     asset.Info,
-			Name:     asset.Info.Name,
-			UnitInfo: asset.Info.UnitInfo,
+			ID:          assetID,
+			Symbol:      asset.Symbol,
+			Wallet:      wallet,
+			Info:        asset.Info,
+			Name:        asset.Info.Name,
+			UnitInfo:    asset.Info.UnitInfo,
+			TradingOpts: tradingConfigOpt,
 		}
 		for tokenID, token := range asset.Tokens {
 			wallet = nil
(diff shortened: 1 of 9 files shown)


client/webserver/site/src/js/forms.ts
```diff
@@ -302,11 +302,12 @@ export class NewWalletForm {
     this.current.selectedDef = walletDef
     const appPwCached = State.passwordIsCached() || (this.pwCache && this.pwCache.pw)
     Doc.hide(page.auth, page.oneBttnBox, page.newWalletPassBox)
-    const configOpts = walletDef.configopts || []
+    const { asset, parentAsset, winfo } = this.current
+    const configOpts = [...walletDef.configopts, ...asset.tradingOpts] || []
     // If a config represents a wallet's birthday, we update the default
     // selection to the current date if this installation of the client
     // generated a seed.
-    configOpts.map((opt) => {
```
vctt94, line 309: no need to be a map as it is not returning any value

client/core/core.go
```diff
@@ -2445,11 +2446,33 @@ func (c *Core) createWallet(crypter encrypt.Crypter, walletPW []byte, assetID ui
 		}
 	}
 
+	var makerswapconfoverride, takerswapconfoverride int64
+	if form.Config["makerswapconfoverride"] != "" {
```
vctt94, line 2450: not good, maybe make a config const keys at core?

buck54321:
There's value here, but we need to re-evaluate the architecture. Re-open as needed.

PR #1946, opened 2022-11-08
asset: create btcwallet+neutrino log dir and file without initialization

### LINKED ISSUE #1944: wallets/internal: Cannot view logs after recover.
If you click `wallet logs` after doing `recover` you are given an internal server error:

image

To reproduce, go on simnet and change btc to SPV. Recover then click wallet logs.

ukane-philemon:
This fixes an issue where wallet recovery deletes the existing neutrino log file but logging initialization is a one-time operation. Closes #1944.

### DIFF
--- a/client/asset/bch/spv.go
+++ b/client/asset/bch/spv.go
@@ -115,7 +115,8 @@ func openSPVWallet(dir string, cfg *btc.WalletConfig, btcParams *chaincfg.Params
 
 // createSPVWallet creates a new SPV wallet.
 func createSPVWallet(privPass []byte, seed []byte, bday time.Time, dbDir string, log dex.Logger, extIdx, intIdx uint32, net *bchchaincfg.Params) error {
-	netDir := filepath.Join(dbDir, net.Name, "spv")
+	netDir := filepath.Join(dbDir, net.Name)
+	walletDir := filepath.Join(netDir, "spv")
 
 	if err := logNeutrino(netDir, log); err != nil {
 		return fmt.Errorf("error initializing bchwallet+neutrino logging: %w", err)
@@ -127,7 +128,7 @@ func createSPVWallet(privPass []byte, seed []byte, bday time.Time, dbDir string,
 		return fmt.Errorf("error creating wallet directories: %w", err)
 	}
 
-	loader := wallet.NewLoader(net, netDir, true, 250)
+	loader := wallet.NewLoader(net, walletDir, true, 250)
 
 	pubPass := []byte(wallet.InsecurePubPassphrase)
 
@@ -148,7 +149,7 @@ func createSPVWallet(privPass []byte, seed []byte, bday time.Time, dbDir string,
 	}
 
 	// The chain service DB
-	neutrinoDBPath := filepath.Join(netDir, neutrinoDBName)
+	neutrinoDBPath := filepath.Join(walletDir, neutrinoDBName)
 	db, err := walletdb.Create("bdb", neutrinoDBPath, true)
 	if err != nil {
 		return fmt.Errorf("unable to create neutrino db at %q: %w", neutrinoDBPath, err)
@@ -168,7 +169,8 @@ func createSPVWallet(privPass []byte, seed []byte, bday time.Time, dbDir string,
 // Start initializes the *bchwallet.Wallet and its supporting players and starts
 // syncing.
 func (w *bchSPVWallet) Start() (btc.SPVService, error) {
-	if err := logNeutrino(w.dir, w.log); err != nil {
+	netDir := filepath.Dir(w.dir)
+	if err := logNeutrino(netDir, w.log); err != nil {
 		return nil, fmt.Errorf("error initializing bchwallet+neutrino logging: %v", err)
 	}
 	// recoverWindow arguments borrowed from bchwallet directly.
@@ -960,7 +962,7 @@ func (w logWriter) Write(p []byte) (n int, err error) {
 // logRotator initializes a rotating file logger.
 func logRotator(netDir string) (*rotator.Rotator, error) {
 	const maxLogRolls = 8
-	logDir := filepath.Join(filepath.Dir(netDir), logDirName)
+	logDir := filepath.Join(netDir, logDirName)
 	if err := os.MkdirAll(logDir, 0744); err != nil {
 		return nil, fmt.Errorf("error creating log directory: %w", err)
 	}
--- a/client/asset/bch/spv.go
+++ b/client/asset/bch/spv.go
@@ -122,12 +122,6 @@ func createSPVWallet(privPass []byte, seed []byte, bday time.Time, dbDir string,
 		return fmt.Errorf("error initializing bchwallet+neutrino logging: %w", err)
 	}
 
-	logDir := filepath.Join(netDir, logDirName)
-	err := os.MkdirAll(logDir, 0744)
-	if err != nil {
-		return fmt.Errorf("error creating wallet directories: %w", err)
-	}
-
 	loader := wallet.NewLoader(net, walletDir, true, 250)
 
 	pubPass := []byte(wallet.InsecurePubPassphrase)
--- a/client/asset/bch/spv.go
+++ b/client/asset/bch/spv.go
@@ -53,7 +53,7 @@ import (
 
 const (
 	DefaultM       uint64 = 784931 // From bchutil. Used for gcs filters.
-	logDirName            = "logs"
+	logDirName            = "spvlogs"
 	neutrinoDBName        = "neutrino.db"
 	defaultAcctNum        = 0
 )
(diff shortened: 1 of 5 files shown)


client/asset/dcr/spv.go
```diff
@@ -138,9 +138,10 @@ var _ tipNotifier = (*spvWallet)(nil)
 
 func createSPVWallet(pw, seed []byte, dataDir string, extIdx, intIdx uint32, chainParams *chaincfg.Params) error {
 	dir := filepath.Join(dataDir, chainParams.Name, "spv")
-
-	if err := initLogging(dir); err != nil {
-		return fmt.Errorf("error initializing dcrwallet logging: %w", err)
+	// Create log path and file.
+	_, err := logRotator(dir)
```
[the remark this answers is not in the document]
ukane-philemon: > It also creates a log rotator but doesn't use it.

Yeah, I'm letting the log rotator create the log file with its file permission...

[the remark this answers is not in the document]
chappjc: We can't just do `rotator.New` and throw away the `*rotator.Rotator`.  It opens the file for writing, and there's an active file descriptor.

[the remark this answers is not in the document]
ukane-philemon: oops, missed closing the rotator.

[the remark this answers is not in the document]
ukane-philemon: oops, missed closing the rotator.

client/asset/bch/spv.go
```diff
@@ -477,6 +476,8 @@ func (w *bchSPVWallet) Stop() {
 	}
 
 	w.log.Info("SPV wallet closed")
+	// Allow the logger to be re-initiated when the wallet reconnects.
+	atomic.CompareAndSwapUint32(&loggingInited, 1, 0)
```
JoeGruffins, line 480: nit: Can just do [StoreUint32(addr *uint32, val uint32)](https://pkg.go.dev/sync/atomic#StoreUint32) since you are not using the return.

JoeGruffins (approved):
Looks fixed to me.

client/asset/bch/spv.go
chappjc: Same thought.  Not using the return has the smell of a bug

buck54321:
Looks like the `spvWallet.dir` is
```Go
dir: filepath.Join(cfg.WalletCFG.DataDir, cfg.ChainParams.Name, "spv"),
```
and the log file is created in a subdirectory of that directory. 

If we instead created the log files in, say, 
```Go
filepath.Join(cfg.WalletCFG.DataDir, cfg.ChainParams.Name, "spvlogs")
```
this wouldn't be an issue.

ukane-philemon:
> If we instead created the log files in, say,
filepath.Join(cfg.WalletCFG.DataDir, cfg.ChainParams.Name, "spvlogs")
this wouldn't be an issue.

I think the issue is that only the log dir is created while the `neutrino.log` file is not, causing an error when trying to retrieve the log file. This PR allows `rotator.New(...)` to create the  `neutrino.log`  file when the wallet is created so that calls to download wallet logs will find the `neutrino.log` file.

buck54321:
My read might be wrong, but it seems to me that the problem is that `(*spvWallet).moveWalletData` is moving log files in the first place.

ukane-philemon:
> My read might be wrong, but it seems to me that the problem is that `(*spvWallet).moveWalletData` is moving log files in the first place.

What is the expected behaviour for new wallets(the create wallet method does not know it's a recovery)? And does it mean the wallet log should be untouched in any scenario?

buck54321:
I don't see a reason to start new log files when we perform recovery. We'll want the record to be continuous, no?

ukane-philemon:
> I don't see a reason to start new log files when we perform recovery. We'll want the record to be continuous, no?

Hmm, maybe not, because we ought to keep all details regarding the old wallet including the logs on the `assetdb-backup` directory. So the new wallet will have to start clean?

chappjc:
I think Buck is saying we should just not delete the logs.  We can still copy them.

Also with the current PR, dcrwallet loggers keep using the old log rotator.  I don't see how it's working.

ukane-philemon:
> I think Buck is saying we should just not delete the logs. We can still copy them.

Okay, I had assumed the expected behaviour for recovery is to remove everything related to the previous wallet(including logs) from the main wallet dir. However, @buck54321 suggestion will keep changes minimal, I think.

ukane-philemon:
@buck54321, RE: https://github.com/decred/dcrdex/pull/1946#issuecomment-1309975817, I had to move some things around in https://github.com/decred/dcrdex/commit/1e56f6dbde69ab75b08234fe9db2d84e14f36aea

client/asset/btc/spv_wrapper.go
```diff
@@ -191,15 +191,44 @@ var (
...
+		if !exist {
+			if err := os.MkdirAll(logDir, 0744); err != nil {
+				return nil, fmt.Errorf("error creating log directory: %w", err)
+			}
+		} else {
+			// Move old logs to new log path.
+			if err := os.Rename(oldDir, logDir); err != nil {
+				return nil, err
+			}
+		}
```
buck54321, line 214: I personally think it'd be fine to skip this "upgrade" step.

client/asset/bch/spv.go
```diff
@@ -942,20 +942,49 @@ func (w logWriter) Write(p []byte) (n int, err error) {
...
+	exist, err := fileExists(logDir)
+	if err != nil {
+		return nil, err
+	}
+	if !exist {
+		oldDir := filepath.Join(netDir, "logs")
+		exist, err := fileExists(oldDir)
+		if err != nil {
+			return nil, err
+		}
```
buck54321, line 955: As mentioned in other comment, I'm not certain we need this upgrade check, but if we do, we probably wouldn't bail on an error here. 

We also have `dex.FileExists`, which returns just a bool.

ukane-philemon: Okay. I didn't want to use the dex pkg in the assets pkg to reduce dependency. But maybe it's okay?

buck54321: We're already importing it.

client/asset/btc/spv_wrapper.go
chappjc: Was thinking the same, especially since it's a perpetual upgrade.  I'd want to remove this stuff eventually, but when?  Let's just leave things

ukane-philemon: >  Let's just leave the old log alone

Alright.

client/asset/dcr/spv.go
```diff
@@ -939,7 +938,7 @@ var (
 // logRotator initializes a rotating file logger.
 func logRotator(netDir string) (*rotator.Rotator, error) {
 	const maxLogRolls = 8
-	logDir := filepath.Join(netDir, logDirName)
+	logDir := filepath.Join(filepath.Dir(netDir), logDirName)
```
chappjc, line 941: Should be just telling `logRotator` the right parent dir to use, not have `logRotator` strip a path.  Here's the right place:

https://github.com/decred/dcrdex/blob/441eeb3df87ada71afe0565d624a581793403a76/client/asset/dcr/spv.go#L141-L142

Don't pass that `dir` and expect `logRotator` to strip off `"spv"`.  Instead have an intermediate `netDir := filepath.Join(dataDir, chainParams.Name)` That you give to `logRotator`

client/asset/btc/spv.go
```diff
@@ -58,18 +58,20 @@ var _ BTCWallet = (*btcSPVWallet)(nil)
...
+	netDir := filepath.Join(dataDir, net.Name)
+	walletDir := filepath.Join(netDir, "spv")
+
+	if err := logNeutrino(netDir); err != nil {
 		return fmt.Errorf("error initializing btcwallet+neutrino logging: %w", err)
 	}
 
-	logDir := filepath.Join(dir, logDirName)
+	logDir := filepath.Join(netDir, logDirName)
 	err := os.MkdirAll(logDir, 0744)
```
chappjc, line 69: Not something with this PR per se, but what are these two `Join`/`MkdirAll` lines doing here after `logNeutrino` anyway?

ukane-philemon: Ah, yeah. I think I wanted to remove this. Because `logNrutrino` will create the directory.

client/asset/btc/spv_wrapper.go
```diff
@@ -1158,7 +1158,25 @@ func (w *spvWallet) moveWalletData(backupDir string) error {
...
+
+func (w *spvWallet) copyWalletLogFile(dstDir string) error {
+	if err := os.MkdirAll(dstDir, 0744); err != nil {
+		return err
+	}
+	logBytes, err := os.ReadFile(w.logFilePath())
+	if err != nil {
+		return err
+	}
+	return os.WriteFile(filepath.Join(dstDir, logFileName), logBytes, 0744)
```
buck54321, line 1179: Is this sufficient? We have a log rotator that gzips logs periodically and performs some amount of truncation (starts from scratch?). For example, here is what my current mainnet **assetdb/btc/mainnet/logs** looks like.
image

chappjc: Aren't we also going to have a collision between network at **assetdb/btc/spvlogs**?  That's the new log path right?

ukane-philemon: > Is this sufficient?

Ah, maybe not. We would need to copy the whole directory content.

client/asset/btc/spv_wrapper.go
```diff
@@ -1157,8 +1158,54 @@ func (w *spvWallet) moveWalletData(backupDir string) error {
...
+// destination directory without preserving the original file permissions.
+func copyDir(srcDir, dstDir string) error {
+	dirFiles, err := os.ReadDir(srcDir)
+	if err != nil {
+		return err
+	}
+
+	if err := os.MkdirAll(dstDir, 0744); err != nil {
+		return err
+	}
```
chappjc, line 1188: The docs for `copyDir` should also say `The destination folder must not exist.` if that's gonna be the restriction.

client/asset/btc/spv_wrapper.go
```diff
@@ -1157,8 +1158,54 @@ func (w *spvWallet) moveWalletData(backupDir string) error {
...
+
+	for _, file := range dirFiles {
+		srcFile := filepath.Join(srcDir, file.Name())
+		dstFile := filepath.Join(dstDir, file.Name())
+		if file.IsDir() {
+			if err := copyDir(srcFile, dstFile); err != nil {
+				return err
+			}
+		} else if file.Type().IsRegular() {
+			logBytes, err := os.ReadFile(srcFile)
```
chappjc, line 1198: We should never do `os.ReadFile` to copy files.  Please see the design here: https://github.com/decred/dcrdata/blob/master/stakedb/ticketpool.go#L174-L220

In addition to using a streaming buffered implementation of `copyFile`, it also `Stat`s the dest dir and only makes it if needed.  We can keep your current restriction that it must not exist, but note that please.

What your code has that the linked code does not seems to be recursion, which is fine to keep, but we should change how we copy the files.

chappjc (changes requested):
Works so that the wallet file is accessible (and continuous) after recover.

Just need to rework the file copy code so it is buffered, not loading a whole file into memory.

client/asset/btc/spv_wrapper.go
ukane-philemon: Alright, will look into the referenced code.

client/asset/btc/spv_wrapper.go
ukane-philemon: I think it should be a restriction.

chappjc (approved):

PR #1948, opened 2022-11-09
multi: Ensure redeemed matches are confirmed.

### LINKED ISSUE #1924: redeems do not require confs if match revoked
Taker will skip confirming the redemption if they do not have server's ack of their redeem request, as stated here:
https://github.com/decred/dcrdex/blob/1619a9ccfefe12c15ffa8e0b788013d6432431b8/client/db/types.go#L336-L337

I think the logic should be changed to check all redemptions.

_Originally posted by @JoeGruffins in https://github.com/decred/dcrdex/issues/1909#issuecomment-1282496285_

This really refers to **revoked** matches skipping the redeem confirm requirement.

This might be a simple fix to the logic, but it needs to be considered carefully when we change the canonical definition of match "activeness".  Will likely require updates to `(*trackedTrade).matchIsActive` to consider the requires confirms bool

JoeGruffins:
multi: Ensure redeemed matches are confirmed.

    Remove conditional checks for revoked matches that allow some to skip
    redemption confirms. All matches that are redeemed now must make it to
    status order.MatchConfirmed to be considered inactive.

closes #1924

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -803,7 +803,15 @@ func (dc *dexConnection) compareServerMatches(srvMatches map[order.OrderID]*serv
 	dc.tradeMtx.RLock()
 	defer dc.tradeMtx.RUnlock()
 	for oid, trade := range dc.trades {
-		activeMatches := trade.activeMatches(false)
+		var activeMatches []*matchTracker
+		for _, m := range trade.activeMatches() {
+			// Server is not expected to report matches that have been fully
+			// redeemed or are revoked. Only client cares about redeem confs.
+			if m.Status >= order.MatchComplete || m.MetaData.Proof.IsRevoked() {
+				continue
+			}
+			activeMatches = append(activeMatches, m)
+		}
 		if len(activeMatches) == 0 {
 			continue
 		}
@@ -7674,7 +7682,7 @@ func (c *Core) schedTradeTick(tracker *trackedTrade) {
 		}
 	}
 
-	numMatches := len(tracker.activeMatches(true))
+	numMatches := len(tracker.activeMatches())
 	switch numMatches {
 	case 0:
 		return
--- a/client/core/trade.go
+++ b/client/core/trade.go
@@ -1189,7 +1189,7 @@ func (t *trackedTrade) isActive() bool {
 		// 	"Order: %v, Refund coin: %v, ContractData: %x, Revoked: %v", match,
 		// 	match.Side, match.Status, t.ID(),
 		// 	proof.RefundCoin, proof.ContractData, proof.IsRevoked())
-		if t.matchIsActive(match, true) {
+		if t.matchIsActive(match) {
 			return true
 		}
 	}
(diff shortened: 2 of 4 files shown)


JoeGruffins:
0ba6063 stolen from #1866

JoeGruffins:
@martonp is it ok to remove `redeemMustConfirm`?

client/core/core.go
```diff
@@ -791,7 +791,7 @@ func (dc *dexConnection) compareServerMatches(srvMatches map[order.OrderID]*serv
 	dc.tradeMtx.RLock()
 	defer dc.tradeMtx.RUnlock()
 	for oid, trade := range dc.trades {
-		activeMatches := trade.activeMatches(false)
```
chappjc, line 794: So this is the one and only spot where we said `false` to not list matches that have at least reached a redeemed status, whether or not we hit `order.MatchConfirmed`, right?  I believe we did this here in `compareServerMatches` where we are trying to `"Identify active matches that are missing from server's response"` because we do not expect the server to report matches that have been redeemed. (server doesn't know or care about the `MatchConfirmed` status.

Lets look at what happens with these "missing" matches and if we can or should make an attempt to exclude (just here) matches that are at least redeemed by us in this use case.

JoeGruffins:
Just rebased and first commit removed.

client/core/core.go
JoeGruffins: Does this create the wanted behavior https://github.com/decred/dcrdex/compare/d7224a25890c5e37ebccf8cd14d05219e7392f16..0b32eda28689ae01f458e9bdbc3faef404f32e44

We don't expect the server to have matches waiting for redeem confirms. I guess we also do not expect revoked matches? Server revoked matches for sure, but can we assume self revoked matches are also not being tracked by the server anymore? Do we care what the server says about those still?

martonp: > Does this create the wanted behavior https://github.com/decred/dcrdex/compare/d7224a25890c5e37ebccf8cd14d05219e7392f16..0b32eda28689ae01f458e9bdbc3faef404f32e44

I think that diff is almost good, but the following line, `found` will be false:
https://github.com/JoeGruffins/dcrdex/blob/0b32eda28689ae01f458e9bdbc3faef404f32e44/client/core/core.go#L810

If you have a function which filters out the unnecessary matches from `activeMatches` before the `if len(activeMatches) == 0` check, it will work.

JoeGruffins: Ok. How about this one https://github.com/decred/dcrdex/compare/0b32eda28689ae01f458e9bdbc3faef404f32e44..9a979de88a0769006305e65985fa370a3ce3cf96

chappjc (changes requested):

JoeGruffins:
Removed activeMatches argument, again

https://github.com/decred/dcrdex/compare/9a979de88a0769006305e65985fa370a3ce3cf96..1b7e5815c18465a0b573b49c78fe90a45ca1d84d

client/db/types.go
```diff
@@ -354,13 +354,9 @@ func MatchIsActive(match *order.UserMatch, proof *MatchProof) bool {
 		// - NewlyMatched requires no further action from either side
 		// - MakerSwapCast requires no further action from the taker
 		// - (TakerSwapCast requires action on both sides)
-		// - MakerRedeemed requires no further action from the maker
-		// - MatchComplete requires no further action. This happens if taker
-		//   does not have server's ack of their redeem request (RedeemSig).
```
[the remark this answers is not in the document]
JoeGruffins: Adding a comment https://github.com/decred/dcrdex/compare/1b7e5815c18465a0b573b49c78fe90a45ca1d84d..4929d13c6f2d4988f5b01778386db8f3ae980683

client/core/trade.go
```diff
@@ -1203,35 +1203,33 @@ func (t *trackedTrade) matchIsRevoked(match *matchTracker) bool {
...
-
-	status, side := match.Status, match.Side
-	return redeemMustConfirm ||
-		(side == order.Maker && status < order.MakerRedeemed) ||
-		(side == order.Taker && status < order.MatchComplete)
+	return isActive
 }
 
-func (t *trackedTrade) activeMatches(redeemMustConfirm bool) []*matchTracker {
+// activeMatches returns active matches. Accepts a bool that ingores matches we
```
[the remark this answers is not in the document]
JoeGruffins: bool and comment removed

client/core/core.go
martonp: Yes this is what I meant, sorry I wasn't clear enough.

martonp (approved):

chappjc (approved):

PR #1951, opened 2022-11-09
client/db: log database path on startup

norwnd:
Would it be useful to have DB path logged on startup ? to be able to back it up somewhere of a separate device, nuke it, etc.

I couldn't easily find where `dexc` stores its data.

### DIFF
--- a/client/db/bolt/db.go
+++ b/client/db/bolt/db.go
@@ -124,6 +124,8 @@ func NewDB(dbPath string, logger dex.Logger) (dexdb.DB, error) {
 	_, err := os.Stat(dbPath)
 	isNew := os.IsNotExist(err)
 
+	logger.Infof("opening database at: %s", dbPath)
+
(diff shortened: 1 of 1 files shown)


client/db/bolt/db.go
```diff
@@ -124,6 +124,8 @@ func NewDB(dbPath string, logger dex.Logger) (dexdb.DB, error) {
 	_, err := os.Stat(dbPath)
 	isNew := os.IsNotExist(err)
 
+	logger.Infof("opening database at: %s", dbPath)
+
```
ukane-philemon, line 128: Nice, we could have it after opening and doing the necessary tasks. i.e right before we return `bdb` and even include the version.

```
bdb.log.Infof("Opened database (version = %d, file = %s)", dbVersion, dbPath)
```

norwnd: Ok, adjusted logged messages a bit.

JoeGruffins (approved):

chappjc (approved):
OK

PR #1954, opened 2022-11-09
client/asset/dcr: Unlock must unlock unmixed account

chappjc:
All accounts must be unlocked, including any unmixed account, which is used to supply keys to the refund path of the swap contract script. If using a mixed account configuration with individually-encrypted accounts, AND the unmixed account is not unlocked, preparing the backup refund tx in Swap will fail (good and early!).


itswisdomagain (approved):

buck54321:
utACK

chappjc:
Ran through every scenario I could think of.  Everything flows as expected, and unlocking works with either full wallet encryption or with individual encryption of just the unmixed account.

PR #1955, opened 2022-11-10
client/asset/dcr: determine sync status with SpvSyncer.Synced() method

### LINKED ISSUE #1816: client/asset/dcr: restart required after restoring native dcr wallet with funds

itswisdomagain:
Resolves #1816.

After headers are fetched, other processes occur before the wallet is considered
synced and balance is updated. Use `SpvSyncer.Synced()` to accurately report
sync completion.

Reporting sync completed only after block headers are fetched results in dexc
reporting an outdated balance, which will only be corrected when a block is mined
_after_ wallet synchronisation (including account discovery and blocks rescan)
fully completes. Thus, new blocks won't necessarily update the balance unless
the new block notification is reported after the sync actually completes.

### DIFF
--- a/client/asset/dcr/spv.go
+++ b/client/asset/dcr/spv.go
@@ -774,7 +774,12 @@ func (w *spvWallet) SyncStatus(ctx context.Context) (bool, float32, error) {
 		return false, 0, nil
 	}
 
-	return w.spv.Synced(), float32(height) / float32(bestHeight), nil
+	synced, progress := w.spv.Synced(), float32(height)/float32(bestHeight)
+	if progress > 0.999 && !synced {
+		progress = 0.999
+	}
+
+	return synced, progress, nil
 }
 
 // bestPeerInitialHeight is the highest InitialHeight recorded from our peers.
--- a/client/asset/dcr/spv.go
+++ b/client/asset/dcr/spv.go
@@ -88,6 +88,7 @@ type dcrWallet interface {
 // Interface for *spv.Syncer so that we can test with a stub.
 type spvSyncer interface {
 	wallet.NetworkBackend
+	Synced() bool
 	GetRemotePeers() map[string]*p2p.RemotePeer
 }
 
@@ -773,7 +774,7 @@ func (w *spvWallet) SyncStatus(ctx context.Context) (bool, float32, error) {
 		return false, 0, nil
 	}
 
-	return bestHeight == height, float32(height) / float32(bestHeight), nil
+	return w.spv.Synced(), float32(height) / float32(bestHeight), nil
 }
 
 // bestPeerInitialHeight is the highest InitialHeight recorded from our peers.
--- a/client/asset/dcr/spv_test.go
+++ b/client/asset/dcr/spv_test.go
@@ -174,6 +174,10 @@ func (w *tDcrWallet) TxDetails(ctx context.Context, txHash *chainhash.Hash) (*ud
 	return w.txDetails, w.txDetailsErr
 }
 
+func (w *tDcrWallet) Synced() bool {
+	return true
+}
+
 func (w *tDcrWallet) GetRemotePeers() map[string]*p2p.RemotePeer {
 	return w.remotePeers
 }


client/asset/dcr/spv.go
```diff
@@ -773,7 +774,7 @@ func (w *spvWallet) SyncStatus(ctx context.Context) (bool, float32, error) {
 		return false, 0, nil
 	}
 
-	return bestHeight == height, float32(height) / float32(bestHeight), nil
+	return w.spv.Synced(), float32(height) / float32(bestHeight), nil
```
chappjc, line 777: Good find!  Although, we get no visibility into the wallet's sync progress, so we sit with `false, 1.0, nil` for a long time on initial sync, which is still quite confusing even if it now correctly says `false`.

There seems to be a `RescanProgress` notification in the `spv.Syncer` we can jack into.
```go
	spvSyncer := newSpvSyncer(dcrw, w.dir, connectPeers)
	spvSyncer.SetNotifications(&spv.Notifications{
		RescanProgress: func(rescannedThrough int32) {
		// something
		},
	})
// now can start syncer
```

Have a look in `dcrwallet/v2/spv.(*Syncer)startupSync` to see what notifications we can and should use.

itswisdomagain: My initial checks show that the blocks rescan is actually pretty fast, but what takes some time is address discovery. I'm not sure if it's possible to estimate the address discovery progress but will look into that. If we account for block rescan progress without accounting for address discovery progress, the percentage progress will sit at some number < 1, synced = false, for a while too; and then soon after address discovery completes, percentage progress will go to 1, especially for a wallet that does not have much transactions.

Some logs to clarify. This tentatively assumes blocks fetching to be half of the total sync progress, so total progress report is 0.5 (50%) when block fetching completes.
```
2022-11-11 20:13:39.394 [INF] CORE[dcr][SPV]: wallet tip 1033902, target height 1033902, blocks fetch progress 1.00, rescan height 1, rescan progress 0.00, overall progress 0.50
/// 5 minutes later, before blocks rescan starts, progress report sits at 0.5 for this 5 minutes period
2022-11-11 20:18:00.393 [INF] CORE[dcr][SPV]: wallet tip 1033903, target height 1033903, blocks fetch progress 1.00, rescan height 16001, rescan progress 0.02, overall progress 0.51
/// block rescan completes within 40-42 seconds, and overall progress jumps to 1.0 in that time frame
2022-11-11 20:18:39.393 [INF] CORE[dcr][SPV]: wallet tip 1033904, target height 1033904, blocks fetch progress 1.00, rescan height 992001, rescan progress 0.96, overall progress 0.98
2022-11-11 20:18:42.396 [INF] CORE: Wallet synced for asset dcr
```

chappjc: ACK.  I thought the period at 100% but sync=false was much longer, including getdata block requests, but that all comes before, at least in a "recover" I just tested.

chappjc (approved):

client/asset/dcr/spv.go
chappjc: Oh, but my balance was still zero after the recover.  When I click "Unlock" the logs at least reflect the non-zero balance:

```
[TRC] CORE[btc][SPV]: Bals: spendable = x.abc BTC (x.abc BTC trusted, 0 BTC untrusted, 0 BTC assumed locked), immature = 0 BTC
TRC] CORE: notify: |DATA| (balance) balance updated
[INF] CORE: Connected to and unlocked btc wallet. Balance available = NNNNN / locked = 0 / locked in contracts = 0, Deposit address = bc1xzzzzzzzzzzzzzzzzz
"POST http://127.0.0.6:5758/api/openwallet HTTP/1.1" from 127.0.0.1:58592 - 200 12B in 1.019914551s
```

But the frontend (wallets page) said 0 until I clicked rescan.  It wasn't the rescan itself that did it, because the balance updated the instant I submitted the "Rescan" but rather some notification triggered by the action.

Seems we're still missing a wallet state or balance note.

client/asset/dcr/spv.go
```diff
@@ -773,7 +774,7 @@ func (w *spvWallet) SyncStatus(ctx context.Context) (bool, float32, error) {
 		return false, 0, nil
 	}
```
chappjc, line 776: Related, above we use `w.dcrWallet.MainChainTip` and `bestPeerInitialHeight`, but should we be using this?

```go

// EstimateMainChainTip returns an estimated height for the current tip of the
// blockchain. The estimate is made by comparing the initial height reported by
// all connected peers and the wallet's current tip. The highest of these values
// is estimated to be the mainchain's tip height.
func (s *Syncer) EstimateMainChainTip() int32 {
	_, chainTip := s.wallet.MainChainTip(context.Background())
	s.forRemotes(func(rp *p2p.RemotePeer) error {
		if rp.InitialHeight() > chainTip {
			chainTip = rp.InitialHeight()
		}
		return nil
	})
	return chainTip
}
```

It's not clear why we have the `bestPeerInitialHeight()` helper defined given

client/asset/dcr/spv.go
chappjc: I don't see a reason looking at `RecoverWallet` or `walletCheckAndNotify` why the updated balance wouldn't come through when `synced` becomes true.  Unless you see something, we'll get this fix merged shortly.

chappjc: Oh derp I was doing *BTC* recover, not DCR.  Apparently BTC has some issue with balances after recover or restore too.

chappjc: Here's a Decred mainnet re-init timing for a wallet with a birthday about 6 months ago:

```
2022-11-15 17:06:49.147 [DBG] SPV: Fetched 2000 new header(s) ending at height 712000 from 34.225.205.218:9108
2022-11-15 17:06:49.319 [INF] SPV: Connected 2000 blocks, new tip 000000000000000088e21481a2dcd8d1fc05a70586eab677aa63bb14970956c1, height 712000, date 2022-11-15 15:31:26 -0600 CST
2022-11-15 17:06:49.556 [DBG] SPV: Fetched 17 new header(s) ending at height 712017 from 107.152.33.208:9108
2022-11-15 17:06:49.560 [INF] SPV: Connected 17 blocks, new tip 000000000000000008d020c02c42f61ef399d5ff1ecfe57278dbc7801096b249, height 712017, date 2022-11-15 16:57:52 -0600 CST
```

Here it is at 100% with synced=false:

image

```
2022-11-15 17:06:49.742 [INF] WLLT: Discovering used addresses for 1 account(s)
2022-11-15 17:07:21.582 [INF] WLLT: Account 0 next child indexes: external:ABC internal:DEF
2022-11-15 17:07:21.582 [INF] WLLT: Updating DB with discovered addresses...
2022-11-15 17:07:21.618 [INF] WLLT: Finished address discovery
2022-11-15 17:07:21.618 [INF] WLLT: Loading active addresses and unspent outputs...
2022-11-15 17:07:21.646 [INF] WLLT: Registered for transaction notifications for XYZ HD address(es)
2022-11-15 17:07:21.646 [INF] WLLT: Registered for transaction notifications for all relevant outputs
2022-11-15 17:07:21.647 [INF] WLLT: Rescanning block range [1, 2000]...
2022-11-15 17:07:21.770 [INF] WLLT: Rescanning block range [2001, 4000]...
2022-11-15 17:07:21.784 [INF] WLLT: Rescanning block range [4001, 6000]...
...
2022-11-15 17:09:07.195 [INF] WLLT: Rescanning block range [712001, 712017]...
2022-11-15 17:09:07.197 [INF] WLLT: Rescan complete
```

Then it becomes synced and reflects the correct balance.

So for a about 2 minutes it sits with 100% but not synced.  Only thing I think we could improve is perhaps capping progress at 99.9% until synced=true.

This commit is going on v0.5.6 as is.

client/asset/dcr/spv.go
itswisdomagain: Thought about making that change but it's basically the same thing anyways, so decided against it.

My next commit will cap progress percentage at 99.9% while synced == false.

client/asset/dcr/spv.go
itswisdomagain: Aha. Was going to retest this to confirm but glad you've caught it!

client/asset/dcr/spv.go
```diff
@@ -773,7 +774,12 @@ func (w *spvWallet) SyncStatus(ctx context.Context) (bool, float32, error) {
...
 
-	return bestHeight == height, float32(height) / float32(bestHeight), nil
+	synced, progress := w.spv.Synced(), float32(height)/float32(bestHeight)
+	if progress == 1 && !synced {
```
chappjc, line 778: Sorry to nit, but to avoid any backward progress, even if small, I think this should be

`if progress > 0.99 & !synced {`

Also that's 99%, so shouldn't we use 0.999?

PR #1961, opened 2022-11-16
core: Refactor MM Bot

martonp:
Refactoring in preparation to add an arbitrage bot.

The `MakerBot` type handles all the interactions core and the DEX markets. The logic of when to buy, sell, and cancel existing orders is handled in a `BotEngine`, which is a property of the `MakerBot`. The `MakerBot` sends notifications to the to a `BotEngine` using its `Notify` function, currently only to let it know that a new epoch has begun. The `BotEngine` exposes a `TradeFeed` which the `MakerBot` listens to in order to know when to submit orders to the dex. Additional information, such as the current orders the bot has on the order books, the current mid-gap price of the market, are available to the `BotEngine` to query through an interface that is provided when the engine is created.

Currently there is one `BotEngine` type, `GapEngine`, which has the same functionality that currently exists in the `MakerBot`.


buck54321 (commented):
@martonp and I have spoken about this change on Matrix, and I've expressed some early approval of the overall approach. This looks good overall. I do have a couple of questions regarding the overall architecture.

1. Passing in an `GapEngineInputs` that is satisfied by `makerBot` and specifies so many methods has advantages for testing, but also hints that an embed pattern may be more appropriate. I think @chappjc has a better grasp on embed  vs. interface patterns. I'd love to get his feedback here.
2. Moving things to a new package causes us to do some acrobatics with type conversions. I personally don't mind if it stays in the **core** package and remains unexported. New file is as good as new package in this case. I can be convinced though. 
3. I'm still not clear on the goals of this PR in terms of integrating an arbitrage function. Will a single `makerBot` control multiple engines, such that a single "master bot" handles all engines for a market? Or will the arbitrage bot engine be assigned to a separate `makerBot`?

martonp:
> 1. Passing in an `GapEngineInputs` that is satisfied by `makerBot` and specifies so many methods has advantages for testing, but also hints that an embed pattern may be more appropriate. I think @chappjc has a better grasp on embed  vs. interface patterns. I'd love to get his feedback here.

`GapEngineInputs` is satisfied by `makerBot`, but I don't think every engine's "inputs" necessarily need to be satisfied by `makerBot`. Ideally there would be a single `Inputs` interface that all engines use. After implementing the arbitrage bot, and maybe some type of other strategy, it will be clear what interface pretty much every strategy could use. My goal was to have it be super easy to implement any type of strategy and then easily unit test it.

> * Moving things to a new package causes us to do some acrobatics with type conversions. I personally don't mind if it stays in the **core** package and remains unexported. New file is as good as new package in this case. I can be convinced though.

The main reason for this is to avoid cluttering core with too many files. Also, I think eventually other people will want to implement their own market making strategies, and if the bot engines are in their own package behind an interface, then it will be much easier.

> 3. I'm still not clear on the goals of this PR in terms of integrating an arbitrage function. Will a single makerBot control multiple engines, such that a single "master bot" handles all engines for a market? Or will the arbitrage bot engine be assigned to a separate makerBot?

My original plan was to have the arbitrage engine be assigned to a separate `makerBot`, but I think having multiple engines in the same bot, and only having one bot per market would allow the `makerBot` to make sure there are no self-matches across the engines. Regarding inventory management, however, I think the user should be able to allocate a certain amount of lots for each strategy, so for this purpose, one bot per engine makes it easier. I still need to think more about these issues.

martonp:
@buck54321 Moved this to draft as I see some changes need to be made based on the arbitrage bot implementation. No need to get this in unless you're also doing bot related work right now.

PR #1965, opened 2022-11-18
server/asset/doge: use feeConfs and reduce maxFeeBlocks

chappjc:
This resolves a couple minor bugs introduced in https://github.com/decred/dcrdex/pull/1597

The `feeConfs` arg that was previously used for `estimatefee` was neglected when the `FeeEstimator` config item was replaced with a more sophisticated `(*Backend).estimateFee` method.  This setting was important for DOGE and BCH where a value of 1 almost always returns an estimate of -1 (failure), resulting in one of the median fee estimates.  In the case it was the "hard" way that is extremely demanding in terms of RPCs to the node, making a remote dogecoind instance virtually infeasible.

A second change is reducing the max blocks in the hard median fee estimator from 50 to 16, which is plenty because if it doesn't find 101 txns in 16 blocks it should just go with the "no competition" rate.

### DIFF
--- a/server/asset/doge/doge.go
+++ b/server/asset/doge/doge.go
@@ -14,7 +14,7 @@ import (
 	"github.com/btcsuite/btcd/chaincfg"
 )
 
-var maxFeeBlocks = 50
+var maxFeeBlocks = 16
 
 // Driver implements asset.Driver.
 type Driver struct{}
@@ -49,7 +49,7 @@ const (
 	version   = 0
 	BipID     = 3
 	assetName = "doge"
-	feeConfs  = 10
+	feeConfs  = 8
 )
 
 // NewBackend generates the network parameters and creates a ltc backend as a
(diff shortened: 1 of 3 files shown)


server/asset/doge/doge.go
```diff
@@ -92,7 +92,8 @@ func NewBackend(configPath string, logger dex.Logger, network dex.Network) (asse
 		Net:                  network,
 		ChainParams:          params,
 		Ports:                ports,
-		DumbFeeEstimates:     true,
+		DumbFeeEstimates:     true, // dogecoind actually has estimatesmartfee though...
```
chappjc, line 95: https://github.com/decred/dcrdex/pull/1558#discussion_r847882970

server/asset/doge/doge.go
```diff
@@ -49,7 +49,7 @@ const (
 	version   = 0
 	BipID     = 3
 	assetName = "doge"
-	feeConfs  = 10
+	feeConfs  = 8
```
chappjc, line 52: https://github.com/decred/dcrdex/pull/1558#discussion_r849436237

The table I pasted there is still accurate even today.

server/asset/doge/doge.go
```diff
@@ -14,7 +14,7 @@ import (
 	"github.com/btcsuite/btcd/chaincfg"
 )
 
-var maxFeeBlocks = 50
```
chappjc, line 17: In practice, it hung and context times out with a remote node.  This just needs to make an quick effort and fall back to the "no competition" rate it can't get 101 txns in 16 blocks or less.

buck54321 (approved):

JoeGruffins (approved):

PR #1966, opened 2022-11-18
client/asset/btc: remove SendOutputs and cleanup

chappjc:
<s>Updating for changes in https://github.com/btcsuite/btcwallet/pull/833

An update to btcwallet causes `SendOutputs` and `CreateSimpleTx` to use P2TR addresses for change unless we specify a keyscope to use for change (new arg for `CreateSimpleTx`, but none available for `SendOutputs`).  To keep the previous behavior where a P2WPKH address is used for change, we use an overriding `SendToOutputs` method.  Note that I've used a different method name, and excluded the unused label arg from the signature, to make it extra clear that it's not a pass-through.</s>

EDIT: `sendToAddress` is removed and the `SendOutputs` method of the btc wallet interface is no longer used (see BTC bonds PR), so this is now a clean-up PR.


client/asset/btc/spv.go
```diff
@@ -408,6 +409,35 @@ func (w *btcSPVWallet) SignTx(tx *wire.MsgTx) error {
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// Create the transaction and broadcast it to the network. The
+	// transaction will be added to the database in order to ensure that we
+	// continue to re-broadcast the transaction upon restarts until it has
+	// been confirmed.
+	changeTypeOpt := wallet.WithCustomChangeScope(&waddrmgr.KeyScopeBIP0084) // p2wpkh change
```
chappjc, line 426: This is why we implement `SendToOutputs`  ourselves now.  btcwallet might make a similar variadic arg change to `SendOutputs`, but we'll still have to override it to match the interface that the ltc and bch packages use.

go.mod
```diff
@@ -5,14 +5,15 @@ go 1.18
...
-	github.com/btcsuite/btcd/btcec/v2 v2.2.1
-	github.com/btcsuite/btcd/btcutil v1.1.2
+	github.com/btcsuite/btcd v0.23.4
+	github.com/btcsuite/btcd/btcec/v2 v2.2.2
+	github.com/btcsuite/btcd/btcutil v1.1.3
 	github.com/btcsuite/btcd/btcutil/psbt v1.1.5
 	github.com/btcsuite/btcd/chaincfg/chainhash v1.0.1
 	github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f
-	github.com/btcsuite/btcwallet v0.16.1
+	github.com/btcsuite/btcwallet v0.16.5
```
chappjc, line 14: https://github.com/btcsuite/btcwallet/compare/v0.16.1...v0.16.5

chappjc:
`sendToAddress` has since been removed and the `SendOutputs` method of the btc wallet interface is no longer used (see BTC bonds PR), so this is now a clean-up PR

PR #1971, opened 2022-11-23
client/core: do not revoke trade when the cancel is revoked

chappjc:
Related to https://github.com/decred/dcrdex/pull/1889, this fixes one such bug that can lead to retired/non-existent orders on the client that are still booked on the server.

```
client/core: do not revoke trade when the cancel is revoked

This fixes a bug where a revoke_order notification from the server that
refers to a cancel order ID would cause the client to revoke the
targeted trade order instead.

Most of the time a cancel order is declared stale and deleted after it
becomes 2 epochs old, a heuristic we use to infer when we've missed the
preimage request for the cancel order.  This was necessary since unlike
trade orders, we don't expect to get status of cancel orders.  However,
we neglected the possibility that we could miss the preimage request
but reconnect in time to receive the revoke_order notification at the
end of the epoch collection period.

When handleRevokeOrderMsg detects that the order ID is a cancel order,
we now unlink the cancel order and send a TopicFailedCancel order ntfn.
The TopicFailedCancel topic previously was only used when cancel orders
were declared "stale", meaning they were more than 2 epochs old.
```

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -7301,9 +7301,19 @@ func handleRevokeOrderMsg(c *Core, dc *dexConnection, msg *msgjson.Message) erro
 	var oid order.OrderID
 	copy(oid[:], revocation.OrderID)
 
-	tracker, _, _ := dc.findOrder(oid)
+	tracker, _, isCancel := dc.findOrder(oid)
 	if tracker == nil {
 		return fmt.Errorf("no order found with id %s", oid.String())
+	}
+
+	if isCancel {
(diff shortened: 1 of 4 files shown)


chappjc:
This bug is recently more severe because of https://github.com/decred/dcrdex/commit/26fa5a1e4f4b9820a3681de35e6aab4482a676d3
Need to merge and cut v0.5.7 asap

chappjc:
Multiple parties running in production fine.  Already backported.

PR #1973, opened 2022-11-24
server: fail fast if txindex is not enabled for btc clones

### LINKED ISSUE #1778: server/zec: The getindexinfo rpc endpoint does not exist.
zec requires that the `-txindex` flag be set when starting as seen in #1767. Our current way of checking this is the `getindexinfo` rpc endpoint. However, it does not appear that zec has this. If there is a way to check we should as it is fatal for zec markets if this is not enabled. Maybe we can use `getrawtransaction` for a known transaction hash on test/mainnet, or could be an easier way. Also, maybe this should be an error and stop server start up?

ukane-philemon:
This ensures `txindex` is enabled by fetching information for a `coinbase` transaction of a given block which is essential for dex activities as it could be fatal for `btc` clone markets if this is not enabled. Closes #1778 

Litecoin Core version v0.21.2.1 has `getindexinfo` rpc command. Tested with bitcoin cash, dogecoin, litecoin and should work for zcash(my zec refuses to sync).
PS: @JoeGruffins, I'd be glad if you can find time to test this with zcash.

### DIFF
--- a/server/asset/btc/rpcclient.go
+++ b/server/asset/btc/rpcclient.go
@@ -29,6 +29,7 @@ const (
 	methodGetIndexInfo      = "getindexinfo"
 	methodGetBlockHeader    = "getblockheader"
 	methodGetBlockStats     = "getblockstats"
+	methodGetBlockHash      = "getblockhash"
 
 	errNoCompetition = dex.ErrorKind("no competition")
 	errNoFeeRate     = dex.ErrorKind("fee rate could not be estimated")
@@ -96,8 +97,8 @@ func (rc *RPCClient) GetBlockChainInfo() (*GetBlockchainInfoResult, error) {
 	return chainInfo, nil
 }
 
-// txIndexResult models the data returned from the getindexinfo command
-// for txindex.
+// txIndexResult models the data returned from the getindexinfo command for
+// txindex.
 // txIndexResult.Txindex is nil if the returned data is an empty json object.
 type txIndexResult struct {
 	TxIndex *struct{} `json:"txindex"`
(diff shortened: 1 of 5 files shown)


JoeGruffins (approved):
Working for zcash.

JoeGruffins:
Unrelated to these changes, but if you send an rpc request to zcash while it is still doing the initial scan, the request hangs forever.

server/asset/btc/rpcclient.go
```diff
@@ -107,12 +108,46 @@ type txIndexResult struct {
...
-	if err != nil {
-		return false, err
+	if err == nil {
+		// Return early if there is no error. bitcoind returns an empty json
+		// object if txindex is not enabled. It is safe to conclude txindex is
+		// enabled if res.Txindex is not nil.
+		return res.TxIndex != nil, nil
+	}
+
+	if isMethodNotFoundErr(err) {
```
chappjc, line 118: Let's negate the condition, unindent the new code, and `return false, err` in the branch instead.

server/asset/btc/rpcclient.go
```diff
@@ -249,6 +284,22 @@ func (rc *RPCClient) GetBlockVerbose(blockHash *chainhash.Hash) (*GetBlockVerbos
...
+	if err != nil {
+		return nil, err
+	}
+
+	blockHash, err := chainhash.NewHashFromStr(blockHashStr)
+	if err != nil {
+		return nil, err
+	}
+
+	return blockHash, nil
```
chappjc, line 300: The boiler plate isn't really needed here. `return chainhash.NewHashFromStr(blockHashStr)`.

Actually, this whole function could be `return rc.callHashGetter(methodGetBlockHash, anylist{index})`, but I can't stand that `callHashGetter` helper TBH.  :laughing:

server/asset/btc/rpcclient.go
```diff
@@ -249,6 +284,22 @@ func (rc *RPCClient) GetBlockVerbose(blockHash *chainhash.Hash) (*GetBlockVerbos
 	return res, rc.call(methodGetBlock, anylist{blockHash.String(), arg}, res)
 }
 
+// GetBlockHash fetches the block hash for the block at the given index.
+func (rc *RPCClient) GetBlockHash(index int64) (*chainhash.Hash, error) {
+	var blockHashStr string
+	err := rc.Call(methodGetBlockHash, []interface{}{index}, &blockHashStr)
```
chappjc, line 290: Pattern is to use `rc.call`, the non-exported version.
Also `anylist{index}`, but it's the same.

chappjc (changes requested):
Looks pretty reasonable to me.  A few little nits.

server/asset/btc/rpcclient.go
ukane-philemon: > Pattern is to use rc.call, the non-exported version.

Yeah, thanks for the review.

server/asset/btc/rpcclient.go
ukane-philemon: I could leave it all in `checkTxIndex` but I'd like to keep it cleaner.

ukane-philemon:
> Unrelated to these changes, but if you send an rpc request to zcash while it is still doing the initial scan, the request hangs forever.

Hmm, other btc clones return an error almost immediately. I think using a new `ctx`  with timeout when making requests with an `rpcClient` should cover such a scenario.

chappjc (approved):

PR #1974, opened 2022-11-24
client: check for zero fiat rates and handle fiat rates request limit

### LINKED ISSUE #1968: fiat values are way off
just use mainnet for a while.

BTC typically ends up showing roughly half the USD after a while.

Showstopper for 0.6

ukane-philemon:
This addresses a bug where a fiat source rate returns zero rates after some time (~19hrs). This source has been removed in https://github.com/decred/dcrdex/commit/0210b57f513676daa32f071a7e67a53de2060095 and a sanity check added in  https://github.com/decred/dcrdex/commit/f7e7006c8e7e144e0946801d63ab62e1bd8a1060. Closes #1968.

PS:  I'd appreciate any suggestions of good sources to replace with.

### DIFF
--- a/client/core/bot.go
+++ b/client/core/bot.go
@@ -742,7 +742,7 @@ func oracleMarketReport(ctx context.Context, b, q *SupportedAsset, log dex.Logge
 	// running on the same market.
 	var rawMarkets []*coinpapMarket
 	url := fmt.Sprintf("https://api.coinpaprika.com/v1/coins/%s/markets", baseSlug)
-	if err := getInto(ctx, url, &rawMarkets); err != nil {
+	if err := getRates(ctx, url, &rawMarkets); err != nil {
 		return nil, err
 	}
 
@@ -1657,7 +1657,7 @@ func fetchBinanceSpread(ctx context.Context, baseSymbol, quoteSymbol string) (se
 		BidPrice float64 `json:"bidPrice,string"`
 		AskPrice float64 `json:"askPrice,string"`
 	}
-	return resp.AskPrice, resp.BidPrice, getInto(ctx, url, &resp)
+	return resp.AskPrice, resp.BidPrice, getRates(ctx, url, &resp)
 }
 
 func fetchCoinbaseSpread(ctx context.Context, baseSymbol, quoteSymbol string) (sell, buy float64, err error) {
@@ -1669,7 +1669,7 @@ func fetchCoinbaseSpread(ctx context.Context, baseSymbol, quoteSymbol string) (s
 		Bid float64 `json:"bid,string"`
 	}
 
-	return resp.Ask, resp.Bid, getInto(ctx, url, &resp)
+	return resp.Ask, resp.Bid, getRates(ctx, url, &resp)
 }
 
 func fetchBittrexSpread(ctx context.Context, baseSymbol, quoteSymbol string) (sell, buy float64, err error) {
@@ -1679,7 +1679,7 @@ func fetchBittrexSpread(ctx context.Context, baseSymbol, quoteSymbol string) (se
 		AskRate float64 `json:"askRate,string"`
 		BidRate float64 `json:"bidRate,string"`
 	}
-	return resp.AskRate, resp.BidRate, getInto(ctx, url, &resp)
+	return resp.AskRate, resp.BidRate, getRates(ctx, url, &resp)
 }
 
 func fetchHitBTCSpread(ctx context.Context, baseSymbol, quoteSymbol string) (sell, buy float64, err error) {
@@ -1690,7 +1690,7 @@ func fetchHitBTCSpread(ctx context.Context, baseSymbol, quoteSymbol string) (sel
 		Ask [][2]json.Number `json:"ask"`
 		Bid [][2]json.Number `json:"bid"`
 	}
-	if err := getInto(ctx, url, &resp); err != nil {
+	if err := getRates(ctx, url, &resp); err != nil {
 		return 0, 0, err
 	}
 	if len(resp.Ask) < 1 || len(resp.Bid) < 1 {
@@ -1719,7 +1719,7 @@ func fetchEXMOSpread(ctx context.Context, baseSymbol, quoteSymbol string) (sell,
 		BidTop float64 `json:"bid_top,string"`
 	}
 
-	if err := getInto(ctx, url, &resp); err != nil {
+	if err := getRates(ctx, url, &resp); err != nil {
 		return 0, 0, err
 	}
 
--- a/client/core/exchangeratefetcher.go
+++ b/client/core/exchangeratefetcher.go
@@ -20,7 +20,7 @@ const (
 	// DefaultFiatCurrency is the currency for displaying assets fiat value.
 	DefaultFiatCurrency = "USD"
 	// fiatRateRequestInterval is the amount of time between calls to the exchange API.
-	fiatRateRequestInterval = 5 * time.Minute
+	fiatRateRequestInterval = 12 * time.Minute
 	// fiatRateDataExpiry : Any data older than fiatRateDataExpiry will be discarded.
 	fiatRateDataExpiry = 60 * time.Minute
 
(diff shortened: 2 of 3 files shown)


JoeGruffins:
In my opinion, the first commit makes a lot of sense and is good. But, maybe we don't need to remove Messari if it's ok half of the time? Do the rates ever come back online? Maybe they are even having some temporary problems? Is it just from that ip, like we are being rate limited, or for everyone it's zero?

Or maybe it's a waste of bandwidth if it down more than half the time. If we are being rate limited, maybe we can just request less.

Here is something:
https://help.messari.io/hc/en-us/articles/8499986549147-What-are-the-key-limits-for-the-API-

ukane-philemon:
Just checked out the article. Messari has a 1000 request limit per day which we exceed even as we request rates every 5min. We eventually make 288 request every 24hr although we are only eligible for 1000/5(200). That means for the first 200 requests Messari will be auto disabled.

JoeGruffins:
> We eventually make ~288 request every 24hr although we are only eligible for 1000/5(200).

What is the `/5`? If we can have 1000 per day and only use 288 should be ok?

ukane-philemon:
> What is the /5?

Oh, that's 5min. The current request interval is 5min

chappjc:
> Messari returns zero rates even if we clearly are not exceeding 1000 per day or 20 per min rate limit.
> Here is a link to my findings [#1968 (comment)](https://github.com/decred/dcrdex/issues/1968#issuecomment-1327031885)

I saw the pastebin you gave for Messari showing zero, but I do not thing the JSON response is actually giving 0 fiat exchange rates.  Normally it would be a 429 http response code (are we not checking for 200 before using the data?) and a different payload, if any.

```
{
  "status": {
    "elapsed": 1,
    "timestamp": "2022-11-29T14:58:05.99596483Z",
    "error_code": 429,
    "error_message": "Rate limit exceeded, retry in 3275 seconds. To increase the limit, create an account on messari.io."
  }
}
```

The other thing, which this PR may already do but I haven't reviewed yet, is to ignore zero rates instead of combining them with other values.

client/core/exchangeratefetcher.go
```diff
@@ -212,7 +215,7 @@ func coinpapSlug(symbol, name string) string {
 	return strings.ToLower(strings.ReplaceAll(slug, " ", "-"))
 }
 
-func getInto(ctx context.Context, url string, thing interface{}) error {
+func getRates(ctx context.Context, url string, thing interface{}) error {
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
```
chappjc, line 184: This function is where we need to check the response code.  Messari will be returning 429 instead of 200, but we don't seem to catch that at all.

client/core/exchangeratefetcher.go
```diff
@@ -196,7 +199,7 @@ func fetchMessariRates(ctx context.Context, log dex.Logger, assets map[uint32]*S
 		slug := strings.ToLower(asset.Symbol)
 		reqStr := fmt.Sprintf(messariURL, slug)
 
-		if err := getInto(ctx, reqStr, res); err != nil {
+		if err := getRates(ctx, reqStr, res); err != nil {
```
chappjc: The reason we think we're getting zero from Messari is this:

```go
		res := new(struct {
			Data struct {
				MarketData struct {
					Price float64 `json:"price_usd"`
				} `json:"market_data"`
			} `json:"data"`
		})
```

When we unmarshal some other JSON into that, it just leaves that `data.market_data.price_usd` as zero, but it's not an unmarshal error, so we end up putting that in `fiatRates[assetID] = res.Data.MarketData.Price` below.

client/core/exchangeratefetcher.go
chappjc: It's just rate limiting.  If you are sharing a public IP, you will get limited faster.
What can we do?  Just return an error and prevent some invalid value from being used.  We can still have the `<=0` check you added, but we know right here if it's valid or not.
I'm not keen to spam the logs though.

client/core/exchangeratefetcher.go
ukane-philemon: > We should only try to unmarshal if the response code was 200 (http.StatusOK).

I agree, but I will have to run Messari for a longer period and see what other `resp.body` we are receiving.

chappjc: The response is here: https://github.com/decred/dcrdex/pull/1974#issuecomment-1330782812

But only the status code matters

client/core/exchangeratefetcher.go
ukane-philemon: Darn, I think we are hitting the rate limit. Assuming we make one request each for all 7 assets every 5min, we end up making ~2016 (7 * 12 req per hour * 24hr) requests in 24hrs although we have a 1000 daily limit.

And then this also affects `Coinparika` because for a user with 7 assets, we end up making `~60480` requests per month which exceed `Coinparika` monthly limit of `25_000` requests.

chappjc: I think we should just increase the request interval a bit, and handle non-200-status code responses properly.  Also cases where all sources fail to return a rate.

This affects all the sources, not just Messari.  As I mentioned, if you are sharing a public IP (e.g. on a VPN or Tor where hundreds or thousands of other users have the same source IP), you'll have failed API requests frequently.

We can also consider backing off in the individual rate fetchers if 429 codes are received.  For example, if `fetchDcrdataRates` gets a 429, it could toggle some global atomic flag assigned to the fetcher to skip future requests until the flag is cleared, and when setting that flag start a time.AfterFunc to clear it after say 5 minutes.

client/core/exchangeratefetcher.go
```diff
@@ -196,7 +199,7 @@ func fetchMessariRates(ctx context.Context, log dex.Logger, assets map[uint32]*S
...
 			log.Error(err)
```
chappjc, line 203: We'll have to see how this goes.  I think it's gonna be irritating (to me) to see the expected 429 code all the time, but perhaps with 12 minute intervals it won't be a big deal.

ukane-philemon:
> I think the spirit of the "Credits" is to just make clear Messari is being used, and we do that. The problem would be if we baked it into the software and had no mention of Messari on the UI.

Yeah, adding a link won't be necessary. I was just trying just to consider all possibilities.

client/core/exchangeratefetcher.go
ukane-philemon: > We can also consider backing off in the individual rate fetchers if 429 codes are received. For example, if fetchDcrdataRates gets a 429, it could toggle some global atomic flag assigned to that fetcher to skip future requests until the flag is cleared, and when setting that flag start a time.AfterFunc to clear it after say 5 minutes.

Yeah, this will make more sense if the rate limiting is based on a minute-by-minute throttling but then, we are never hitting any of the per-minute limits(at least not until we add support for up to 20 assets).

client/core/bot.go
```diff
@@ -742,7 +742,7 @@ func oracleMarketReport(ctx context.Context, b, q *SupportedAsset, log dex.Logge
 	// running on the same market.
 	var rawMarkets []*coinpapMarket
 	url := fmt.Sprintf("https://api.coinpaprika.com/v1/coins/%s/markets", baseSlug)
-	if err := getInto(ctx, url, &rawMarkets); err != nil {
```
ukane-philemon, line 745: We could even hit the rate limit faster if there is a running bot using any of the fiat rate sources. We can consider sources that place limits only on a minute-by-minute basis.

client/core/exchangeratefetcher.go
chappjc: Depends if the service keeps counting failed requests against your rate limit.  Anyway, just a thought.  Let's go with what you have now.

client/core/exchangeratefetcher.go
```diff
@@ -224,6 +227,10 @@ func getInto(ctx context.Context, url string, thing interface{}) error {
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("unexpected response, got status code %d", resp.StatusCode)
```
chappjc, line 231: No need to update, but a helpful tool is the `http.StatusText` function.

```go
		return fmt.Errorf("unexpected response, got status code %d (%s)",
			resp.StatusCode, http.StatusText(resp.StatusCode))
```

So you see "429 (Too Many Requests)" or "403 (Forbidden)" e.g.

chappjc (approved):

client/core/exchangeratefetcher.go
ukane-philemon: Okay.

PR #1975, opened 2022-11-25
ui: store markets/wallets user preferences

norwnd:
This PR implements saving (between page switch/refresh or `dexc` restarts) the following UI preferences a user can set:
- hide/show left market dock on `/markets` page
- select chosen wallet on `/wallets` page

Relevant UI views affected:

image image

### DIFF
--- a/client/webserver/site/src/js/app.ts
+++ b/client/webserver/site/src/js/app.ts
@@ -108,7 +108,7 @@ export default class Application {
     this.commitHash = process.env.COMMITHASH || ''
     this.noteReceivers = []
     this.fiatRatesMap = {}
-    this.showPopups = State.getCookie('popups') === '1'
+    this.showPopups = State.getCookie(State.PopupsCK) === '1'
     console.log('Decred DEX Client App, Build', this.commitHash.substring(0, 7))
 
     // Loggers can be enabled by setting a truthy value to the loggerID using
--- a/client/webserver/site/src/js/app.ts
+++ b/client/webserver/site/src/js/app.ts
@@ -48,8 +48,6 @@ const bind = Doc.bind
 const unbind = Doc.unbind
 
 const notificationRoute = 'notify'
-const loggersKey = 'loggers'
-const recordersKey = 'recorders'
 const noteCacheSize = 100
 
 interface Page {
@@ -114,18 +112,18 @@ export default class Application {
     // Loggers can be enabled by setting a truthy value to the loggerID using
     // enableLogger. Settings are stored across sessions. See docstring for the
     // log method for more info.
-    this.loggers = State.fetch(loggersKey) || {}
+    this.loggers = State.fetch(State.LoggersLK) || {}
     window.enableLogger = (loggerID, state) => {
       if (state) this.loggers[loggerID] = true
       else delete this.loggers[loggerID]
-      State.store(loggersKey, this.loggers)
+      State.store(State.LoggersLK, this.loggers)
       return `${loggerID} logger ${state ? 'enabled' : 'disabled'}`
     }
     // Enable logging from anywhere.
     window.log = (loggerID, ...a) => { this.log(loggerID, ...a) }
 
     // Recorders can record log messages, and then save them to file on request.
-    const recorderKeys = State.fetch(recordersKey) || []
+    const recorderKeys = State.fetch(State.RecordersLK) || []
     this.recorders = {}
     for (const loggerID of recorderKeys) {
       console.log('recording', loggerID)
@@ -134,7 +132,7 @@ export default class Application {
     window.recordLogger = (loggerID, on) => {
       if (on) this.recorders[loggerID] = []
       else delete this.recorders[loggerID]
-      State.store(recordersKey, Object.keys(this.recorders))
+      State.store(State.RecordersLK, Object.keys(this.recorders))
       return `${loggerID} recorder ${on ? 'enabled' : 'disabled'}`
     }
     window.dumpLogger = loggerID => {
--- a/client/webserver/site/src/js/app.ts
+++ b/client/webserver/site/src/js/app.ts
@@ -931,9 +931,8 @@ export default class Application {
   }
 
   /**
-   * signOut call to /api/logout, if response with no errors occurred clear all
-   * store, remove auth, darkMode cookies and reload the page, otherwise will
-   * show a notification
+   * signOut call to /api/logout, if response with no errors occurred remove
+   * auth cookie and reload the page, otherwise show a notification.
    */
   async signOut () {
     const res = await postJSON('/api/logout')
@@ -946,7 +945,6 @@ export default class Application {
       Doc.show(this.page.logoutErr)
       return
     }
-    State.clearAllStore()
     State.removeAuthCK()
     window.location.href = '/login'
   }
--- a/client/webserver/site/src/js/app.ts
+++ b/client/webserver/site/src/js/app.ts
@@ -106,7 +106,7 @@ export default class Application {
     this.commitHash = process.env.COMMITHASH || ''
     this.noteReceivers = []
     this.fiatRatesMap = {}
-    this.showPopups = State.fetchLocal(State.popupsCK) === '1'
+    this.showPopups = State.fetchLocal(State.popupsLK) === '1'
 
     console.log('Decred DEX Client App, Build', this.commitHash.substring(0, 7))
 
--- a/client/webserver/site/src/js/app.ts
+++ b/client/webserver/site/src/js/app.ts
@@ -112,18 +112,18 @@ export default class Application {
     // Loggers can be enabled by setting a truthy value to the loggerID using
     // enableLogger. Settings are stored across sessions. See docstring for the
     // log method for more info.
-    this.loggers = State.fetch(State.LoggersLK) || {}
+    this.loggers = State.fetchLocal(State.LoggersLK) || {}
     window.enableLogger = (loggerID, state) => {
       if (state) this.loggers[loggerID] = true
       else delete this.loggers[loggerID]
-      State.store(State.LoggersLK, this.loggers)
+      State.storeLocal(State.LoggersLK, this.loggers)
       return `${loggerID} logger ${state ? 'enabled' : 'disabled'}`
     }
     // Enable logging from anywhere.
     window.log = (loggerID, ...a) => { this.log(loggerID, ...a) }
 
     // Recorders can record log messages, and then save them to file on request.
-    const recorderKeys = State.fetch(State.RecordersLK) || []
+    const recorderKeys = State.fetchLocal(State.RecordersLK) || []
     this.recorders = {}
     for (const loggerID of recorderKeys) {
       console.log('recording', loggerID)
@@ -132,7 +132,7 @@ export default class Application {
     window.recordLogger = (loggerID, on) => {
       if (on) this.recorders[loggerID] = []
       else delete this.recorders[loggerID]
-      State.store(State.RecordersLK, Object.keys(this.recorders))
+      State.storeLocal(State.RecordersLK, Object.keys(this.recorders))
       return `${loggerID} recorder ${on ? 'enabled' : 'disabled'}`
     }
     window.dumpLogger = loggerID => {
@@ -184,7 +184,7 @@ export default class Application {
     this.attachCommon(this.header)
     this.attach({})
     // Load recent notifications from Window.localStorage.
-    const notes = State.fetch(State.NotificationsLK)
+    const notes = State.fetchLocal(State.NotificationsLK)
     this.setNotes(notes || [])
     // Connect the websocket and register the notification route.
     ws.connect(getSocketURI(), this.reconnected)
@@ -434,7 +434,7 @@ export default class Application {
    * actual stored list is stripped of information not necessary for display.
    */
   storeNotes () {
-    State.store(State.NotificationsLK, this.notes.map(n => {
+    State.storeLocal(State.NotificationsLK, this.notes.map(n => {
       return {
         subject: n.subject,
         details: n.details,
@@ -944,11 +944,11 @@ export default class Application {
       Doc.show(this.page.logoutErr)
       return
     }
-    State.removeAuthCK()
-    State.removePwKeyCK()
-    State.remove(State.LastMarketLK)
-    State.remove(State.LastMMMarketLK)
-    State.remove(State.NotificationsLK)
+    State.removeCookie(State.AuthCK)
+    State.removeCookie(State.PwKeyCK)
+    State.removeLocal(State.LastMarketLK)
+    State.removeLocal(State.LastMMMarketLK)
+    State.removeLocal(State.NotificationsLK)
     window.location.href = '/login'
   }
 }
--- a/client/webserver/site/src/js/app.ts
+++ b/client/webserver/site/src/js/app.ts
@@ -184,7 +184,7 @@ export default class Application {
     this.attachCommon(this.header)
     this.attach({})
     // Load recent notifications from Window.localStorage.
-    const notes = State.fetch('notifications')
+    const notes = State.fetch(State.NotificationsLK)
     this.setNotes(notes || [])
     // Connect the websocket and register the notification route.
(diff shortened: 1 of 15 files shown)


client/webserver/site/src/js/markets.ts
```diff
@@ -469,6 +466,9 @@ export default class MarketsPage extends BasePage {
         this.setMarket(row.mkt.xc.host, row.mkt.baseid, row.mkt.quoteid)
       })
     }
+    if (!State.showLeftMarketDock()) { // It is shown by default, hiding if necessary.
+      closeMarketsList()
+    }
```
norwnd, line 471: Alternatively we can generate template without left-market-doc on backend, but doing it like this was just easier.
How do you choose between using back-end-templating / javascript for things like these ? Here it doesn't seem to make any difference.

norwnd: Also, if we leave it in JS, I believe [local storage can be used instead of cookies](https://stackoverflow.com/questions/3220660/local-storage-vs-cookies), what do you think ?

chappjc: signOut presently does `window.localStorage.clear()` for security reasons but I suppose that could change.

buck54321: > Alternatively we can generate template without left-market-doc on backend, but doing it like this was just easier. How do you choose between using back-end-templating / javascript for things like these ? Here it doesn't seem to make any difference.

It's better how you have it.

client/webserver/site/src/js/settings.ts
```diff
@@ -48,9 +48,7 @@ export default class SettingsPage extends BasePage {
     })
 
     Doc.bind(page.showPokes, 'click', () => {
-      const show = page.showPokes.checked || false
-      State.setCookie('popups', show ? '1' : '0')
-      app().showPopups = show
```
buck54321, line 53: Deleting this line seems wrong. The changes won't apply until refresh... 

Verified. Doesn't update without refresh.

buck54321 (changes requested):
Good stuff.

client/webserver/site/src/js/markets.ts
norwnd: > It's better how you have it.

What about cookies vs local storage ? I think I should change to local storage.

buck54321: > > It's better how you have it.
> 
> What about cookies vs local storage ? I think I should change to local storage.

I suppose anything the server doesn't use would be better off in local storage.

client/webserver/site/src/js/settings.ts
norwnd: Phew, I got a bit confused by who/how is using `app().showPopups`, 

now I think I understand better how interfaces work in typescript, fixed.

client/webserver/site/src/js/markets.ts
norwnd: Moved to local storage.

norwnd:
Force-pushed last commit there (here is [the diff](https://github.com/decred/dcrdex/compare/5af9ef98fbe4a529756f32ba162b2472e4e5fc3c..d0df2493f239f0c6d9ea13df5b3dd76a556732b7)) to fix commit message formatting issues.

client/webserver/site/src/js/markets.ts
chappjc: So it's meant to be lost on logout?

norwnd: Oh man, I forgot about your comment above! Gotta address that, we don't want to loose it on logout - [changed](https://github.com/decred/dcrdex/pull/1975/commits/11b05754d7dc905ce8baec53b9985ec46a7819ae) and re-tested (seems to be working).

So, now on logout only `dexauth` cookie gets cleaned up, I understand this is the behavior that fits best.

client/webserver/site/src/js/app.ts
```diff
@@ -946,7 +945,6 @@ export default class Application {
       Doc.show(this.page.logoutErr)
       return
     }
-    State.clearAllStore()
```
chappjc, line 949: Alright I suppose.

The local storage includes a string of `notifications` that includes coin IDs and account IDs, etc. as well as `selectedMarket`, which shows what dex host and market was last selected.  Historically we have been clearing these.

1. If it no longer cleans up on logout, this can leave potentially sensitive info in the browser's user data.  Are we OK with that?
2. Are there consequences of leaving `notifications`?  On next login, do they get duplicated?  Were we relying on this to clean out this storage value, and will it now grow forever?

norwnd: > If it no longer cleans up on logout, this can leave potentially sensitive info in the browser's user data.

Just my thoughts. As casual DEX user I don't even know that logout cleans up after me, so I wouldn't count on it as a user and if need to I would try cleaning-up browser data through means browser has (but, I'm not sure what the equivalent actions for Decredtion-embedded DEX would be ...).

If cleaning-up user data is a desirable feature to have, one option I can suggest is to have special button in `/settings` section to "Erase **ALL** data (**seed**, trading history, metadata)" that would clean up browser storage (local/cookies) and drop `dexc` database allowing user to start from scratch. From what I understand, there is no such functionality currently (`dexc` database has to be deleted manually in order to "start from scratch") ? I'd be happy to take a crack at it.

> Are there consequences of leaving notifications? will it now grow forever?

I've looked a bit into how notifications are handled in UI, it seems only latest ones are shown / saved to local storage (there is a limit `const noteCacheSize = 100`), so there shouldn't be any issues with that. And in `dexc` database only notifications of `severity >= dexdb.Success` are stored, I don't see any obvious clean-up procedures there, but probably it won't occupy too much space in practice.

> On next login, do they get duplicated?

I don't think so. Don't observe any duplicates while testing.
From what I see in the code, `/login` request just replaces whatever there is locally stored under `notifications`.

norwnd: Useful note about "resetting DEX client to factory settings", looks like [this  code](https://github.com/decred/dcrdex/blob/391f2fdbf95bb7772f34f87ff875d33f9bca3d35/client/webserver/site/src/js/register.ts#L174-L178) takes care of cleaning-up local storage from  notifications left there.

client/webserver/site/src/js/state.ts
```diff
@@ -1,12 +1,22 @@
...
+  static PwKeyCK = 'sessionkey'
+  // Local storage keys (for data that we don't need at the server).
+  static LoggersLK = 'loggers'
+  static RecordersLK = 'recorders'
+  static LastMarketLK = 'selectedMarket'
+  static DepthZoomLK = 'depthZoom'
+  static LastMMMarketLK = 'mmMarket'
+  static OptionsExpansionLK = 'mmOptsExpand'
+  static LeftMarketDockLK = 'leftmarketdock'
+  static SelectedAssetLK = 'selectedasset'
```
norwnd, line 18: After @chappjc mentioned [those 2 local storage values I was unaware of](https://github.com/decred/dcrdex/pull/1975#discussion_r1051113367) I though it might be worth grouping all local storage / cookie entries for easier tracking, like this ^.

Which I kinda started doing already in this PR, and [now covered all the rest](https://github.com/decred/dcrdex/pull/1975/commits/5c25bcf46ef2426f5e6ed2d90cc99cec33e82c00).

client/webserver/site/src/js/app.ts
chappjc: > > If it no longer cleans up on logout, this can leave potentially sensitive info in the browser's user data.
> 
> Just my thoughts. As casual DEX user I don't even know that logout cleans up after me, so I wouldn't count on it as a user. If need to, I would try cleaning-up browser data through means browser has (but, I'm not sure what the equivalent actions for Decredtion-embedded DEX would be ...).

Of course a user isn't aware!  This is just a best practice that developers should follow.  Whether or not a user is aware of it is immaterial.  "Sign out" functionality should clean up as much as reasonable to avoid leaving junk *in the browser* that is ripe for exploits.  Obviously some stuff should persist (dark mode, etc.), but not the stuff with transaction details.

> If cleaning-up user data is a desirable feature to have, one option I can suggest is to have special button in `/settings` section to "Erase **ALL** data (**seed**, trading history, metadata)" that would clean up browser storage (local/cookies) and drop `dexc` database allowing user to start from scratch. From what I understand, there is no such functionality currently (`dexc` database has to be deleted manually in order to "start from scratch") ? I'd be happy to take a crack at it.

No, that's not needed or related to the issue at hand.  First of all, the dexc data (DB etc) is not the issue. 1. Resetting the _backend_ is not something that's often required. 2. It's easy to delete (open folder, delete file).  3. A browser's "local storage" is the real issue because it's not an isolated and permissioned ".dexc" folder used by the backend.  Browsers are shared by countless other web apps and sites, and while browsers are better these days about sandboxing things, it's just smart to limit what gets stored in the browser's data.

norwnd: I see, sounds like we can delete only privacy-critical values, leaving the rest alone upon sign out for user convenience ? Added [commit that does just that](https://github.com/decred/dcrdex/pull/1975/commits/4b81cf98c1cda187e62b5fe2376e13cb8a9d9985).

client/webserver/site/src/js/state.ts
norwnd: Whoops, forgot to add `static NotificationsLK = 'notifications'`, now done.

client/webserver/site/src/js/markets.ts
buck54321: > How do you choose between using back-end-templating / javascript for things like these ?

When there isn't a clear choice, I lean Javascript.

client/webserver/site/src/js/state.ts
```diff
@@ -26,55 +37,68 @@ export default class State {
...
+    return !!this.getCookie(State.PwKeyCK)
   }
 
-  /* store puts the key-value pair into Window.localStorage. */
-  static store (k: string, v: any) {
-    window.localStorage.setItem(k, JSON.stringify(v))
+  /* showLeftMarketDock returns whether or not user wants left market doc shown. */
+  static showLeftMarketDock () {
+    return this.fetchLocal(State.LeftMarketDockLK) === '1'
   }
```
buck54321, line 64: Only used one place and is specific to **/markets**. I'd say just do this in **markets.ts** and forego the method.

client/webserver/site/src/js/state.ts
```diff
@@ -26,55 +37,68 @@ export default class State {
...
+  /*
+   * selectedAsset returns selected asset ID or null if none user hasn't selected one
+   * yet.
+   */
+  static selectedAsset (): number | null {
+    const assetIDStr = State.fetchLocal(State.SelectedAssetLK)
+    if (!assetIDStr) {
+      return null
+    }
+    return Number(assetIDStr)
```
buck54321, line 75: This one is also specific to **/wallets** and only used one place. Any reason to have this logic here instead of in **markets.ts**?

client/webserver/site/src/js/state.ts
```diff
@@ -1,12 +1,23 @@
...
-const pwKeyCK = 'sessionkey'
-
 // State is a set of static methods for working with the user state. It has
 // utilities for setting and retrieving cookies and storing user configuration
 // to localStorage.
 export default class State {
+  // Cookie keys.
+  static DarkModeCK = 'darkMode'
+  static AuthCK = 'dexauth'
+  static PopupsCK = 'popups'
```
buck54321, line 8: Can be local?

client/webserver/site/src/js/state.ts
```diff
@@ -1,12 +1,23 @@
...
-const authCK = 'dexauth'
-const popupsCK = 'popups'
-const pwKeyCK = 'sessionkey'
-
 // State is a set of static methods for working with the user state. It has
 // utilities for setting and retrieving cookies and storing user configuration
 // to localStorage.
 export default class State {
+  // Cookie keys.
+  static DarkModeCK = 'darkMode'
```
buck54321, line 6: We could **almost** move this to local storage. It's only used in the backend template (via `UserInfo.DarkMode`) in two places, both of which could be set with Javascript. I think the reasoning of sending the value from the client is that if you're in dark mode, and you set the CSS class via Javascript, you see a flash of light mode while the script is loading. Maybe there's a way around that, but I feel like it was a thorny issue. 

Anyway, nothing to change here.

buck54321 (changes requested):

client/webserver/site/src/js/state.ts
norwnd: Looks like yes, converted to local.

client/webserver/site/src/js/state.ts
norwnd: yeah, makes sense, applied

client/webserver/site/src/js/state.ts
norwnd: > Any reason to have this logic here instead of in markets.ts?

Probably none, unlike `State.isDark` or `State.passwordIsCached` this method is unlikely to be re-used (used > 1 time), applied this suggestion.

client/webserver/site/src/js/state.ts
norwnd: I guess that's a good example for [my question from above about where to use template/javascript](https://github.com/decred/dcrdex/pull/1975#discussion_r1032065037).

client/webserver/site/src/js/wallets.ts
```diff
@@ -217,7 +217,13 @@ export default class WalletsPage extends BasePage {
       createwallet: (note: WalletCreationNote) => { this.handleCreateWalletNote(note) }
     })
 
-    this.setSelectedAsset(firstAsset.id)
+    const firstAsset = this.sortAssetButtons()
+    let selectedAsset = firstAsset.id
+    const assetIDStr = State.fetchLocal(State.selectedAssetLK)
+    if (assetIDStr) {
+      selectedAsset = Number(assetIDStr)
+    }
```
buck54321, line 225: One line will do

client/webserver/site/src/js/state.ts
```diff
@@ -1,12 +1,23 @@
...
-const pwKeyCK = 'sessionkey'
-
 // State is a set of static methods for working with the user state. It has
 // utilities for setting and retrieving cookies and storing user configuration
 // to localStorage.
 export default class State {
+  // Cookie keys.
+  static darkModeCK = 'darkMode'
+  static authCK = 'dexauth'
+  static popupsCK = 'popups'
```
buck54321, line 8: This can be a local key?

client/webserver/site/src/js/app.ts
```diff
@@ -946,8 +944,11 @@ export default class Application {
...
-    State.removeAuthCK()
+    State.removeCookie(State.authCK)
+    State.removeCookie(State.pwKeyCK)
+    State.removeLocal(State.lastMarketLK)
+    State.removeLocal(State.lastMMMarketLK)
```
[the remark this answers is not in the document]
norwnd: Probably not, unless somebody is extremely worried about his privacy, removed.

client/webserver/site/src/js/wallets.ts
norwnd: updated

client/webserver/site/src/js/state.ts
norwnd: yes, I think so, moved to local

client/webserver/site/src/js/state.ts
```diff
@@ -5,9 +5,9 @@ export default class State {
   // Cookie keys.
   static darkModeCK = 'darkMode'
   static authCK = 'dexauth'
-  static popupsCK = 'popups'
   static pwKeyCK = 'sessionkey'
   // Local storage keys (for data that we don't need at the server).
+  static popupsCK = 'popups'
```
buck54321, line 10: LK

buck54321:
Looks good. Needs a rebase.

norwnd:
Rebased with necessary adjustments on top: https://github.com/decred/dcrdex/pull/1975/commits/0c4e85940e28cabeb242e76ab382c9a63d31e35c.

buck54321 (approved):

PR #1977, opened 2022-11-26
ui: Enter key usage on verify order view

norwnd:
When the final form for posting order (to confirm all the data) shows up after typing password and (instinctively) hitting `Enter` nothing happens, unlike with `/login` page (or when cancelling order) where it works. That's because focus doesn't change automatically to `Sell DCR` / `Buy DCR` button when `Enter` button is pressed. This PR addresses this.

Form affected:

image

### DIFF
--- a/client/webserver/site/src/html/forms.tmpl
+++ b/client/webserver/site/src/html/forms.tmpl
@@ -67,11 +67,11 @@
       <label for="newWalletPass" class="form-label pt-3 ps-1 mb-1">[[[Wallet Password]]]
         <span class="ico-info" data-tooltip="[[[w_password_tooltip]]]"></span>
       </label>
-      <input type="password" class="form-control select" data-tmpl="newWalletPass" autocomplete="current-password">
+      <input type="password" class="form-control select" data-tmpl="newWalletPass" autocomplete="new-password">
     </div>
     <div class="hide-pw col-11 p-0">
       <label for="nwAppPass" class="form-label pt-3 ps-1 mb-1">[[[App Password]]]</label>
-      <input type="password" class="form-control select" id="nwAppPass" data-tmpl="appPass" autocomplete="current-password">
+      <input type="password" class="form-control select" id="nwAppPass" data-tmpl="appPass" autocomplete="new-password">
     </div>
     <span class="col-3 hide-pw"></span> {{/* flex dummy */}}
     <div class="d-grid align-items-end col-10 p-0 mt-4">
--- a/client/webserver/site/src/html/forms.tmpl
+++ b/client/webserver/site/src/html/forms.tmpl
@@ -67,11 +67,11 @@
       <label for="newWalletPass" class="form-label pt-3 ps-1 mb-1">[[[Wallet Password]]]
         <span class="ico-info" data-tooltip="[[[w_password_tooltip]]]"></span>
       </label>
-      <input type="password" class="form-control select" data-tmpl="newWalletPass" autocomplete="off">
+      <input type="password" class="form-control select" data-tmpl="newWalletPass" autocomplete="current-password">
     </div>
     <div class="hide-pw col-11 p-0">
       <label for="nwAppPass" class="form-label pt-3 ps-1 mb-1">[[[App Password]]]</label>
-      <input type="password" class="form-control select" id="nwAppPass" data-tmpl="appPass" autocomplete="off">
+      <input type="password" class="form-control select" id="nwAppPass" data-tmpl="appPass" autocomplete="current-password">
     </div>
     <span class="col-3 hide-pw"></span> {{/* flex dummy */}}
     <div class="d-grid align-items-end col-10 p-0 mt-4">
@@ -110,7 +110,7 @@
 <div id="uwAppPassBox" class="my-3">
   <label for="uwAppPass" class="form-label ps-1">[[[App Password]]]</label>
   <div class="fs14 px-1 mb-1">[[[app_password_reminder]]]</div>
-  <input type="password" class="form-control select" id="uwAppPass" autocomplete="off">
+  <input type="password" class="form-control select" id="uwAppPass" autocomplete="current-password">
 </div>
 <div id="submitUnlockDiv" class="d-flex justify-content-end">
   <button id="submitUnlock" type="submit" class="col-8 justify-content-center fs15 bg2 selected">[[[Unlock]]]</button>
@@ -405,7 +405,7 @@
 <div class="d-flex flex-row align-items-end {{if $passwordIsCached}}justify-content-end{{end}} pb-4">
   <div class="col-12 p-0 {{if $passwordIsCached}}d-hide{{end}}">
     <label for="cancelPass" class="form-label pt-3 mb-1">[[[Password]]]</label>
-    <input type="password" class="form-control select" id="cancelPass" autocomplete="off">
+    <input type="password" class="form-control select" id="cancelPass" autocomplete="current-password">
   </div>
   <div class="col-12 py-1 {{if not $passwordIsCached}}ps-5{{end}} position-relative">
     <button id="cancelSubmit" type="button" class="w-100 fs15 bg2 selected position-relative d-inline-block">[[[Submit]]]</button>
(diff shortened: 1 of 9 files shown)


norwnd:
One thing related to this feature is probably worth noting, when `Remember my password` is checked by the user on verify order view there is no longer a password field where focus was placed perviously, so pressing Enter just results in form-refresh and not in form-submission. I don't really have any preference with respect to that, but the trade-off looks like this to me:
- user can't accidentally press `Enter` and create order before he had time to take a look at the "verify order view"
- but he'll need extra click (or `tab`) to proceed

client/webserver/site/src/html/wallets.tmpl
```diff
@@ -323,7 +323,7 @@
        <div class="fs16 px-4 text-center">[[[Authorize the transfer with your app password.]]]</div>
         <div class="input-group mt-2">
           <label for="vSendPw" class="form-label m-2">[[[Password]]]: </label>
-          <input type="password" class="form-control select" id="vSendPw" autocomplete="off">
+          <input type="password" class="form-control select" id="vSendPw" autocomplete="current-password">
```
norwnd, line 326: I noticed there are several places where `autocomplete="off"` is used instead of `autocomplete="current-password"`, if it isn't on-purpose - I assumed it probably needs changing.

client/webserver/site/src/js/markets.ts
```diff
@@ -415,6 +415,9 @@ export default class MarketsPage extends BasePage {
     bind(page.mktBuyField, 'keyup', () => { this.marketBuyChanged() })
     bind(page.rateField, 'change', () => { this.rateFieldChanged() })
     bind(page.rateField, 'keyup', () => { this.previewQuoteAmt(true) })
+    // Verify order view.
+    bind(page.vPass, 'input', () => { this.vPassInputHandler() })
+    bind(page.vUnlockPass, 'input', () => { this.vUnlockPassInputHandler() })
```
norwnd, line 420: Kinda unrelated to this PR, but we discussed a similar issue (about hiding error) in another one so I decided to fix it as well while I'm here.

client/webserver/site/src/html/wallets.tmpl
chappjc: https://github.com/decred/dcrdex/issues/414

Allowing autocomplete also tells the browser to remember what you entered.  We don't want that.

https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#the_autocomplete_attribute_and_login_fields

> 
> Setting autocomplete="off" on fields has two effects:
> 
>    * It tells the browser **not to save data inputted by the user** for later autocompletion on similar forms, though heuristics for complying vary by browser.
>    * It **stops the browser from caching form data in the session history**. When form data is cached in session history, the information filled in by the user is shown in the case where the user has submitted the form and clicked the Back button to go back to the original form page.

chappjc:
Lots of unrelated changes in one commit here.  Try to push separate commits to show what does what.

client/webserver/site/src/html/wallets.tmpl
norwnd: Okay, then I suppose instead better replace `autocomplete=current-password` and `autocomplete="new-password"` with `autocomplete="off"` on all views ? Since it is present on some in `master` which is why I wanted to change it in the first place (to make it uniform everywhere).

Just to mention, using password-autocomplete also allows for seamless integration with password manager (such as LastPass). But I suppose when dealing with money reducing attack surface is preferable.

chappjc: lastpass can and does ignore it in my experience.  It's quite aggressive... an irritating about it actually

norwnd: so, about other places on `master`, should I `autocomplete="off"` them as well ?

buck54321:
It's not unusual for sensitive forms to have no enter-submission. Have you compared this behavior to order forms on other exchanges?

buck54321:
> That's because focus doesn't change automatically to Sell DCR / Buy DCR button when Enter button is pressed. This PR addresses this.

Shouldn't require an focus changes. We aren't doing that with e.g. login.

buck54321:
> Lots of unrelated changes in one commit here. Try to push separate commits to show what does what. e.g.
> 
> * fix focus and submit action on password field
> * rename stuff
> * fix autocomplete
> * change some html style and formatting
> * fix something with vPass/vUnlockPass, see previous issue X

Yeah. Fixing the described problem can be done in a few lines, if we even want to.

@norwnd can you minimize this diff please. Don't change HTML indentation, don't rename unrelated things, etc.

buck54321:
> It's not unusual for sensitive forms to have no enter-submission. Have you compared this behavior to order forms on other exchanges?

On second thought, this isn't a field of the actual order form, and it's the only field on the "form", so enter submit probably makes sense.

buck54321:
@norwnd check out https://github.com/buck54321/dcrdex/compare/4685064b998de6e5a7dea26e89f8c1ffb258d10e...buck54321:dcrdex:enter-submit-order-pw to see how I'd do it

norwnd:
> It's not unusual for sensitive forms to have no enter-submission. Have you compared this behavior to order forms on other exchanges?

Yeah, I thought about that. I think I've seen both implementations in different products.

For this particular case user probably can't "accidentally proceed" unless he has entered proper password, but if he did then he probably meant it. So upon closer look it seems like non-issue to me.

> @norwnd can you minimize this diff please. Don't change HTML indentation, don't rename unrelated things, etc.

Let me pull apart this PR in separate ones (would be easier to decide for each one if  changes there make sense), since there are couple of unrelated things I was trying to address (I started on the issue mentioned in PR description, but stumbled upon couple of other ones, I think).

chappjc:
> For this particular case user probably can't "accidentally proceed" unless he has entered proper password, but if he did then he probably meant it. So upon closer look it seems like non-issue to me.

Most people click the "remember password' checkbox on login.

client/webserver/site/src/html/forms.tmpl
```diff
@@ -67,11 +67,11 @@
       <label for="newWalletPass" class="form-label pt-3 ps-1 mb-1">[[[Wallet Password]]]
         <span class="ico-info" data-tooltip="[[[w_password_tooltip]]]"></span>
       </label>
-      <input type="password" class="form-control select" data-tmpl="newWalletPass" autocomplete="off">
+      <input type="password" class="form-control select" data-tmpl="newWalletPass" autocomplete="new-password">
```
buck54321, line 70: But this isn't a new password, e.g. it won't benefit from use of a password manager. I think `off` is the best choice here.

client/webserver/site/src/html/forms.tmpl
```diff
@@ -67,11 +67,11 @@
...
     </div>
     <div class="hide-pw col-11 p-0">
       <label for="nwAppPass" class="form-label pt-3 ps-1 mb-1">[[[App Password]]]</label>
-      <input type="password" class="form-control select" id="nwAppPass" data-tmpl="appPass" autocomplete="off">
+      <input type="password" class="form-control select" id="nwAppPass" data-tmpl="appPass" autocomplete="new-password">
```
buck54321, line 74: This one is definitely not new. Could be `current-password` maybe.

client/webserver/site/src/js/markets.ts
```diff
@@ -949,6 +952,14 @@ export default class MarketsPage extends BasePage {
...
   }
 
+  vPassInputHandler () {
+    Doc.hide(this.page.verifyAuthFormErr)
+  }
+
+  vUnlockPassInputHandler () {
+    Doc.hide(this.page.vPreorderErr)
+  }
+
```
buck54321, line 962: These one-line functions are only used once. Might as well just inline it.

buck54321 (changes requested):
What is the goal of this PR? Are you catching an enter key on a password input, or are you "fixing" the html. Please undo anything not related to the goal of this PR. You can follow up with separate PRs for 1) setting autocomplete, 2) improving html structure and changing class names and ids.

norwnd:
> What is the goal of this PR?

It isn't very high on my priorities list right now, but I eventually want to re-visit it and see whether anything can be pulled out of it (like [I mentioned previously](https://github.com/decred/dcrdex/pull/1977#issuecomment-1339448037)),

I think there isn't a need to keep it open anymore, closing.

Issue #1927, opened 2022-10-28
UI/order details: invalid coin when finding eth redemption.


JoeGruffins:
For eth based assets, when something happens that prevents one client from sending their redemption, and the other client finds the redemption, there is no txid. This is because the redemption is really a contract without any traceable connection to the redeeming tx (with our tools). This is fine. But when we show this nonexistent txid in the UI it does not look good.

![image](https://user-images.githubusercontent.com/34998433/198552794-4808562a-2577-4814-9637-2efe2ca7be76.png)

This is really this https://github.com/decred/dcrdex/blob/e4bdbf020465b7def3a9691dc9890b406f6d546e/client/asset/eth/eth.go#L150

Something more informative should be returned here. Or the UI should deal with it better. I don't know that there is anything we can point to on chain that makes sense in an explorer. If there is something we can point to that would be best.

Maybe just a message that reads something like "Redemption txid not accessible."?

norwnd:
> I don't know that there is anything we can point to on chain that makes sense in an explorer. If there is something we can point to that would be best.

Googled about this a bit, it seems it isn't possible to pull out the actual smart-contract call transaction hash (which would be ideal), see
- the (closed) issue I've linked to above
- and current list of exposed variables:  https://docs.soliditylang.org/en/v0.8.17/cheatsheet.html#global-variables 

Note, `nonce` isn't exposed, that would not allow for `address` + `nonce` lookup  either (which would have been another option).

> "Redemption txid not accessible."?

That exact wording doesn't answer the "why" question, which is probably the most important one to answer,

maybe something like `"%s redeemed"` (where `%s` is counterparty's address) 
since `SwapState.Initiator` and `Swap.Participant` are already [exposed](https://github.com/decred/dcrdex/blob/d606490ca154807d5895acbcee9ee4a38d3c7b4f/dex/networks/eth/params.go#L207-L208) ?

chappjc:
> maybe something like `"%s redeemed"` (where `%s` is counterparty's address)
> since `SwapState.Initiator` and `Swap.Participant` are already [exposed](https://github.com/decred/dcrdex/blob/d606490ca154807d5895acbcee9ee4a38d3c7b4f/dex/networks/eth/params.go#L207-L208), and that would allow for manual best-guess lookup at least ?

[...]

PR #1978, opened 2022-11-28
ui: display Maker redeem found by Taker

norwnd:
Resolves https://github.com/decred/dcrdex/issues/1927.
Depends on / extends a bit https://github.com/decred/dcrdex/pull/1733 (`ethExplorers` func).

Maker/Taker sides:

image image

Here are corresponding redeem [transaction](https://goerli.etherscan.io/tx/0x999fc57cf90bd9a4221e494ccc1534d2b90a9f1e90f4e9072cc1990b704b8d58) and Maker [Ethereum address](https://goerli.etherscan.io/address/0x9eEd1b4A72B4Bf7768A852c48715531b1Ee4a26c) where **0x9eEd1b4A72B4Bf7768A852c48715531b1Ee4a26c redeemed** link highlighted on Maker-side screenshot leads to.


client/core/types.go
```diff
@@ -919,8 +919,7 @@ func token(id []byte) string {
 }
 
 // coinIDString converts a coin ID to a human-readable string. If an error is
-// encountered, the error is logged at Warning level, and the value
-// "<invalid coin>" is returned.
+// encountered the value starting with "<invalid coin>:" prefix is returned.
```
norwnd, line 922: Looks like this needed correction since recently.

client/asset/eth/eth.go
```diff
@@ -218,7 +224,16 @@ func (d *Driver) DecodeCoinID(coinID []byte) (string, error) {
...
 		}
 		return common.HexToAddress(hexAddr).String(), nil
+	case len(coinIDTakerFoundMakerRedemption) + common.AddressLength*2 + 2:
+		hexCoinID := string(coinID)
+		if !strings.HasPrefix(hexCoinID, coinIDTakerFoundMakerRedemption) {
+			return "", fmt.Errorf("coinID %q has no %s prefix", coinID, coinIDTakerFoundMakerRedemption)
+		}
+		coinIDCopy := make([]byte, len(hexCoinID))
+		copy(coinIDCopy, hexCoinID)
+		return string(coinIDCopy), nil
```
norwnd, line 234: Looks like `DecodeCoinID` strives to return a copy of its argument, so it won't be modified explicitely ? Doing same here.

client/core/types.go
JoeGruffins: I wish we logged the error here rather than throwing it away...

JoeGruffins (approved):
Looks good and working well for me.

client/asset/eth/eth.go
JoeGruffins: I think the weirdness you see is because the go-ethereum package will capitalize some hex for some reason that I forget atm. @martonp what was it again?

I do not that performing `string(hexCoinID)` would alter the slice anyway.

JoeGruffins: Oh, this was it https://medium.com/@danielque/why-ethereum-addresses-have-some-capitalized-letters-b8e55728ecaa

client/core/types.go
norwnd: Looks like @buck54321 removed it, if there is no objections I can put it back, I also believe it's best to not loose the original error message when possible
```
log.Warnf("invalid coin ID %x for asset %d -> %s: %v", coinID, assetID, unbip(assetID), err)
```

client/asset/eth/eth.go
norwnd: Yeah, [looks like](https://go.dev/play/p/k8JpRvQLAda) wrapping `[]byte` with `string()` protects underlying slice from implicit modification,

so, no need to do deep-copy, thanks, simplified!

client/asset/eth/eth.go
```diff
@@ -2291,9 +2305,15 @@ type findRedemptionRequest struct {
 
 // sendFindRedemptionResult sends the result or logs a message if it cannot be
 // sent.
-func (eth *baseWallet) sendFindRedemptionResult(req *findRedemptionRequest, secretHash [32]byte, secret []byte, err error) {
+func (eth *baseWallet) sendFindRedemptionResult(
```
martonp, line 2308: You could split this into 2 lines instead of 1 line per arg. That's what I see elsewhere in the codebase.

martonp (approved):
LGTM

client/asset/eth/eth.go
norwnd: Ok, applied.

client/webserver/site/src/js/order.ts
```diff
@@ -257,14 +259,21 @@ export default class OrderPage extends BasePage {
     tmpl.status.textContent = OrderUtil.matchStatusString(m)
 
     const setCoin = (pendingName: string, linkName: string, coin: Coin) => {
+      const formatCoinID = (cid: string) => {
+        if (cid.startsWith(coinIDTakerFoundMakerRedemption)) {
+          const makerAddr = cid.substring(coinIDTakerFoundMakerRedemption.length)
+          return `${makerAddr} redeemed`
```
buck54321, line 265: Should format via **intl**. Recommend format `redeemed by: [addr]`

buck54321 (changes requested):

client/core/types.go
chappjc: Unfortunate consequence of losing the global logger.
My stance is that fmt.Print is not a sin compared with dropping the error entirely or poisoning the function signatures with a logger argument (terrible IMO).
If we've been fine with dropping it, what's so bad about printing to stdout.

client/webserver/site/src/js/order.ts
norwnd: Now re-tested (also checked that similar flow isn't broken, if done in reverse for the same trading pair - that is when Taker finds Maker DCR redemption on its own):

image

client/webserver/site/src/js/order.ts
```diff
@@ -257,14 +259,21 @@ export default class OrderPage extends BasePage {
     tmpl.status.textContent = OrderUtil.matchStatusString(m)
 
     const setCoin = (pendingName: string, linkName: string, coin: Coin) => {
+      const formatCoinID = (cid: string) => {
+        if (cid.startsWith(coinIDTakerFoundMakerRedemption)) {
+          const makerAddr = cid.substring(coinIDTakerFoundMakerRedemption.length)
+          return intl.prep(intl.ID_TAKER_FOUND_MAKER_REDEMPTION, { makerAddr: makerAddr })
+        }
+        return cid
+      }
```
buck54321, line 268: Front end is taking on a new role and I'm not certain it's necessary. Why not just set the coin ID to `fmt.Sprintf("redeemed by %s", makerAddr)` in `FindRedemption`?  It's still a constant length string for `(Driver),DecodeCoinID`. Or you could leave `FindRedemption` as you have it, and format the `redeemed by` string in `DecodeCoinID`, which we be more amenable to internationalization.

norwnd: > Why not just set the coin ID to fmt.Sprintf("redeemed by %s", makerAddr) in FindRedemption?

I considered that, but we'd still need to take that same role when extracting `address` to build a link to Etherscan (see `const ethExplorers` definition), hence I though it is better to define this CoinID format **explicitly** instead of relying on the exact phrasing **implicitly** (which by the way changed [as per your suggestion](https://github.com/decred/dcrdex/pull/1978#discussion_r1040278273): `${makerAddr} redeemed` -> `redeemed by ${makerAddr}` - which would have broken Etherscan-url-building code if not adjusted accordingly),

> Or you could leave FindRedemption as you have it, and format the redeemed by string in DecodeCoinID, which we be more amenable to internationalization.

and that same reasoning I outlined above also applies here.

client/webserver/site/src/js/locales.ts
```diff
@@ -93,6 +93,7 @@ export const ID_MATCH_STATUS_REFUNDED = 'MATCH_STATUS_REFUNDED'
 export const ID_MATCH_STATUS_REFUND_PENDING = 'MATCH_STATUS_REFUND_PENDING'
 export const ID_MATCH_STATUS_REDEEM_PENDING = 'MATCH_STATUS_REDEEM_PENDING'
 export const ID_MATCH_STATUS_COMPLETE = 'MATCH_STATUS_COMPLETE'
+export const ID_TAKER_FOUND_MAKER_REDEMPTION = 'ID_TAKER_FOUND_MAKER_REDEMPTION'
```
buck54321, line 96: Remove the `ID_` prefix from the value.

buck54321 (changes requested):
Testing well. One last nit.

client/webserver/site/src/js/locales.ts
norwnd: Hmm, you are sharp! Should be consistent with the rest now.

client/webserver/site/src/js/order.ts
chappjc: Yeah, the explorer link builder is a pickle.  If we think of a better solution later, it shouldn't be a problem to change it since this encoding exception exists only on the client.

chappjc (approved):

client/webserver/site/src/js/order.ts
chappjc: Well, aside from adding a new field to the `core` package's `type Coin struct` and modifying `NewCoin` somehow

buck54321 (approved):

PR #1980, opened 2022-12-02
client: add static strings to translation

ukane-philemon:
This removes a handful of statics strings and adds them to the translation in `locales.ts`.


client/webserver/site/src/js/locales.ts
```diff
@@ -78,6 +78,23 @@ export const ID_ARCHIVED_RECORDS_PATH = 'ARCHIVED_RECORDS_PATH'
...
+export const ID_UNSUPPORTED_ASSET_INFO_ERR_MSG = 'UNSUPPORTED_ASSET_INFO_ERR_MSG'
+export const ID_LIMIT_ORDER = 'LIMIT_ORDER'
+export const ID_LIMIT_ORDER_IMMEDIATE_TIF = 'LIMIT_ORDER_IMMEDIATE_TIF'
+export const ID_MARKET_ORDER = 'MARKET_ORDER'
+export const ID_MATCH_STATUS_NEWLY_MATCHED = 'MATCH_STATUS_NEWLY_MATCHED'
+export const ID_MATCH_STATUS_MAKER_SWAP_CAST = 'MATCH_STATUS_MAKER_SWAP_CAST'
+export const ID_MATCH_STATUS_TAKER_SWAP_CAST = 'MATCH_STATUS_TAKER_SWAP_CAST'
+export const ID_MATCH_STATUS_MAKER_REDEEMED = 'MATCH_STATUS_MAKER_REDEEMED'
+export const ID_MATCH_STATUS_REDEMPTION_SENT = 'MATCH_STATUS_REDEMPTION_SENT'
+export const ID_MATCH_STATUS_MATCH_COMPLETE = 'MATCH_STATUS_MATCH_COMPLETE'
```
JoeGruffins, line 90: `ID_MATCH_STATUS_MATCH_COMPLETE` does not have an entry below. It also looks unused in the codebase.

client/webserver/site/src/js/orderutil.ts
```diff
@@ -135,41 +137,41 @@ export function matchStatusString (m: Match) {
...
     case MakerRedeemed:
       if (m.side === Maker) {
-        return 'Redemption Sent'
+        return intl.ID_MATCH_STATUS_REDEMPTION_SENT
       }
-      return 'Maker Redeemed'
+      return intl.ID_MATCH_STATUS_MAKER_REDEEMED
     case MatchComplete:
-      return 'Redemption Sent'
+      return intl.ID_MATCH_STATUS_REDEMPTION_SENT
```
JoeGruffins, line 170: I guess this one should be the `ID_MATCH_STATUS_MATCH_COMPLETE`

client/webserver/site/src/js/locales.ts
JoeGruffins: But there is also `ID_MATCH_STATUS_COMPLETE` so maybe only need one or the other?

client/webserver/site/src/js/orderutil.ts
JoeGruffins: Are these revoked? It looks like all `m.revoked` already returned.

JoeGruffins: Maybe I don't understand what you are saying, but `Complete` looks correct here for status `MatchComplete`

ukane-philemon: > but Complete looks correct here for status MatchComplete

Yes, we can use `Complete` was just thinking why we have been using `Redemption Sent` for `MatchComplete`.

chappjc: > I was just thinking why we have been using `Redemption Sent` for `MatchComplete`.

Because we have `MatchConfirmed` now, otherwise it's just "Redemption Sent". @martonp Will you check out this PR?

martonp: We updated it because in the `MatchComplete` state, an order is still active (waiting for confirmation). It would be confusing to see "Match Complete" on the screen but you cannot shut the client off because an order is still active.

martonp (approved):

client/webserver/site/src/js/orderutil.ts
JoeGruffins: I guess that makes sense.

JoeGruffins (approved):

client/webserver/site/src/js/orderutil.ts
```diff
@@ -135,41 +137,41 @@ export function matchStatusString (m: Match) {
     // When revoked, match status is less important than pending action if still
     // active, or the outcome if inactive.
     if (m.active) {
-      if (m.redeem) return 'Revoked - Redemption Sent' // must require confirmation if active
+      if (m.redeem) return intl.prep(intl.ID_MATCH_STATUS_REVOKED, { status: intl.ID_MATCH_STATUS_REDEMPTION_SENT }) // must require confirmation if active
```
buck54321, line 140: Seems like you'd need to `intl.prep` the `status` value too, no?

```Typescript
intl.prep(intl.ID_MATCH_STATUS_REVOKED, { status: intl.prep(intl.ID_MATCH_STATUS_REDEMPTION_SENT) })
```

client/webserver/site/src/js/orderutil.ts
```diff
@@ -135,41 +137,41 @@ export function matchStatusString (m: Match) {
...
+      return intl.prep(intl.ID_MATCH_STATUS_REVOKED, { status: intl.ID_MATCH_STATUS_REDEMPTION_CONFIRMED })
     }
-    return 'Revoked - Complete' // i.e. we sent no swap
+    return intl.prep(intl.ID_MATCH_STATUS_REVOKED, { status: intl.ID_MATCH_STATUS_COMPLETE }) // i.e. we sent no swap
   }
 
   switch (m.status) {
     case NewlyMatched:
-      return 'Newly Matched'
+      return intl.ID_MATCH_STATUS_NEWLY_MATCHED
```
buck54321, line 159: These also need to be `prep`'d, right?

client/webserver/site/src/js/orderutil.ts
ukane-philemon: yeah, thanks for the review.

client/webserver/site/src/js/orderutil.ts
```diff
@@ -135,41 +137,41 @@ export function matchStatusString (m: Match) {
...
-      return 'Maker Redeemed'
+      return intl.prep(intl.ID_MATCH_STATUS_MAKER_REDEEMED)
     case MatchComplete:
-      return 'Redemption Sent'
+      return intl.prep(intl.ID_MATCH_STATUS_REDEMPTION_SENT)
     case MatchConfirmed:
-      return 'Redemption Confirmed'
+      return intl.prep(intl.ID_MATCH_STATUS_REDEMPTION_CONFIRMED)
   }
-  return 'Unknown Match Status'
```
chappjc, line 172: Some thoughts about this one.
- This is in a Match card, and there's "Status" heading.  I think this can just be "Unknown"
- There's an existing `ID_UNKNOWN` that is simply `'unknown'`
- we've never seen this happen, which is a good sign, no egregious bugs with the status values

All considered, I don't think we need yet another string to translate.

image :heavy_check_mark: 

vs

image :x:

client/webserver/site/src/js/orderutil.ts
```diff
@@ -135,41 +137,41 @@ export function matchStatusString (m: Match) {
...
     if (m.refund) {
-      return 'Revoked - Refunded'
+      return intl.prep(intl.ID_MATCH_STATUS_REVOKED, { status: intl.prep(intl.ID_MATCH_STATUS_REFUNDED) })
     }
     if (m.redeem) {
-      return 'Revoked - Redemption Confirmed'
+      return intl.prep(intl.ID_MATCH_STATUS_REVOKED, { status: intl.prep(intl.ID_MATCH_STATUS_REDEMPTION_CONFIRMED) })
     }
-    return 'Revoked - Complete' // i.e. we sent no swap
+    return intl.prep(intl.ID_MATCH_STATUS_REVOKED, { status: intl.prep(intl.ID_MATCH_STATUS_COMPLETE) }) // i.e. we sent no swap
```
chappjc, line 154: NBD, but the pattern repeats for all the revoked statuses, so I think this could all be more readable with a helper letting us do e.g. `return revokedMatchStatus(intl.ID_MATCH_STATUS_COMPLETE)`

PR #1981, opened 2022-12-02
rpc: Add default config options to settings.

JoeGruffins:
If user does not supply some default values over rpc when creating
    wallets, use default values. Add the eth test token, zec, and doge to
    the simnet client set up script.

This is mostly to assist testing on simnet.


client/core/core.go
```diff
@@ -8403,9 +8403,9 @@ func parseCert(host string, certI interface{}, net dex.Network) ([]byte, error)
 	return nil, fmt.Errorf("not a valid certificate type %T", certI)
 }
 
-// walletDefinition gets the registered WalletDefinition for the asset and
+// WalletDefinition gets the registered WalletDefinition for the asset and
 // wallet type.
-func walletDefinition(assetID uint32, walletType string) (*asset.WalletDefinition, error) {
+func WalletDefinition(assetID uint32, walletType string) (*asset.WalletDefinition, error) {
```
chappjc, line 8408: Kinda funky that this is a `core` functions.  It only uses the `client/asset` package internally, and it returns a `client/asset` type.

JoeGruffins:
`WalletDefinition` moved to client/asset https://github.com/decred/dcrdex/compare/611f2faaad4d2d1902a97224b722d510a4567f23..78da72663ca4d753712a380555465ec54b947217

client/rpcserver/handlers.go
```diff
@@ -192,6 +193,22 @@ func handleNewWallet(s *RPCServer, params *RawParams) *msgjson.ResponsePayload {
...
+		errMsg := fmt.Sprintf("error creating %s wallet: unable to get wallet definition: %v",
+			dex.BipIDSymbol(form.assetID), err)
+		resErr := msgjson.NewError(msgjson.RPCWalletDefinitionError, errMsg)
+		return createResponse(newWalletRoute, nil, resErr)
+	}
+
+	// Apply default config options if they exist.
+	for _, opt := range walletDef.ConfigOpts {
+		if _, has := form.config[opt.Key]; !has && opt.DefaultValue != nil {
+			form.config[opt.Key] = fmt.Sprintf("%v", opt.DefaultValue)
```
chappjc, line 208: `fmt.Sprintf("%v", ...)` should do the trick for all the types we give in `DefaultValue`, but in our other uses of this field (`WalletConfigForm.map`) we consider the other flags like `isdate`and `repeatable`.  I don't feel like we need to replicate all that at this time though.

chappjc (approved):
LGTM, with the one noted caveat regarding the correct way to go from a configopt item to a string in the `map[string]string` given to the wallet form.

client/rpcserver/handlers.go
chappjc: To clarify, nothing to change unless another reviewer sees an issue related to this.

buck54321 (approved):

PR #1986, opened 2022-12-05
client: allow skipping registration when adding a dex (view-only)

### LINKED ISSUE #726: dexc: view only support
It would be nice if dexc could be "watch only" --  watch trade activity without needing wallets.

### LINKED ISSUE #1882: client: allow adding a dexConnection with no dexAccount
The current flow to "Add a DEX" is tied to registering an account on that DEX, however, this flow could go differently:
- You do Add a DEX, specify the host name, click OK
- Without registering, it adds the DEX host, and takes you **straight to the Markets page** where you see the depth chart, order book, etc. just like you do now if you are registered
- Instead of the order submission form, there is a Create Account button and some simple text explaining that to trade you need to create an identity with the server.
- The markets list on the left will allow you to explorer the markets for all servers regardless of registration.
- Clicking the Create Account button will take you to the form we have now that shows what assets that server accepts for the fee. When register and then you come back to the markets page, it would show the confirmation countdown and eventually the actual order form.

This should be fairly straightforward in `client.Core` with perhaps a little extra logic and fields on the `dexConnection` and methods that process the `Core.conns` map to allow servers with no accounts to be handled without issue.  Namely, don't attempt `authDEX`, block orders, etc.

This would require a new `Core` method to add a dex without registering (yet).  The existing register method would need to learn to work with both new/unknown dex hosts as well as an accountless dex already in the `hosts` map.

Somewhat related, but I don't see the bandwidth as an issue anymore: https://github.com/decred/dcrdex/issues/142  Solving the present issue would be straightforward if "accountless" "dexConnections" behaved the same way on backend and frontend w.r.t. order book subscriptions and all other notifications.

On a related note, there could be an "Add a DEX" on the left on the Markets page:

image
(that looks terrible, but for illustration...)

itswisdomagain:
Resolves #1882. Also resolves #726.

This adds a new flow for connecting to a DEX.
- The registration page shows a checkbox to connect a DEX in view-only mode (no registration required).
Screenshot 2022-12-05 at 22 15 06

- The markets page shows a "Create Account" button for view-only DEX connections.
Screenshot 2022-12-05 at 22 15 26

- The "Create Account" button from the markets page redirects to the register page with a confirmation form that performs account discovery on submission, and if necessary, starts the fee payment process.
Screenshot 2022-12-05 at 22 17 17


client/webserver/locales/en-us.go
```diff
@@ -25,6 +25,7 @@ var EnUS = map[string]string{
 	"remove":                         "remove",
 	"add a file":                     "add a file",
 	"Submit":                         "Submit",
+	"Skip Registration":              "Skip Registration (view only mode)",
```
buck54321, line 28: view-only should be hyphenated

buck54321 (changes requested):
Damn good work! Looks and tests great. Just a couple little things.

client/core/types.go
```diff
@@ -735,6 +736,14 @@ func (a *dexAccount) ID() account.AccountID {
 	return a.id
 }
 
+// isRegistered is true if this account has a feeCoin. Every registered account
+// must have a feeCoin set.
+func (a *dexAccount) isRegistered() bool {
+	a.keyMtx.RLock()
+	defer a.keyMtx.RUnlock()
+	return len(a.feeCoin) > 0
```
chappjc, line 744: `a.feeCoin` is not an good indicator of an unconfigured `dexAccount` (bonds, but also payment status is not a direct indicator of the existence of an account identity/key).

What are the characteristics of a `dexAccount` created by the `AddDEX` method?  I think the answer is that it has not been "initialized" via `setupCryptoV2`, i.e. the fields guarded by `keyMtx` are still zero-valued.  The fee, bond, and auth related fields guarded by `authMtx` are not what we are interested in I think.

The previous `(*Core).isRegistered` had the appropriate check, although now it's expected to have a `dexConnection` in the map with a `dexAccount` with no configured identity.

client/webserver/webserver.go
```diff
@@ -317,6 +318,7 @@ func New(cfg *Config) (*WebServer, error) {
 		// initial setup and settings is used to register more DEXs
 		// after initial setup.
 		web.Get(registerRoute, s.handleRegister)
+		web.With(dexHostCtx).Get(registerRoute+"/{host}", s.handleRegister)
```
chappjc, line 321: Would suggest a sub-router for expressiveness:

```go
web.Route(registerRoute, func(rr chi.Router) {
	rr.Get("/", s.handleRegister)
	rr.With(dexHostCtx).Get("/{host}", s.handleRegister)
})
```

chappjc (changes requested):
I think we're conflating two concepts: (a) being registered at the DEX host, and (b) having a generated account identity/keys for authentication/authorization purposes regardless of registration and fee/bond status.

In `Register` of course we are concerned with both, particularly the first one so we don't pay a fee twice.

In many other places, we really don't need to know about fee coins or bonds or anything like that; we just want to know if we've run `setupCryptoV2` and created account keys.  I think that is what distinguishes a view-only account from one intended for authorization (regardless of payment details).

See my comments on the new `(*dexAccount).isRegistered` method.

client/core/types.go
itswisdomagain: That was my first approach too, checking `a.id`, or `a.encKey`/`a.privKey`. But realized that those fields would be generated (by `DiscoverAccount -> setupCryptoV2`) if a registration is attempted but not completed. Thought about zeroing those fields if `DiscoverAccount` does not find a paid account and `Register` does not pay the required fee, but I worry things may fall through the crack with that approach. At least for a brief period, those keys won't be empty and the account will be thought to be registered (by `a.isRegistered`).

Another thought was to NOT use the dexConnection added by `AddDEX` in `DiscoverAccount` and `Register`, to ensure that that dexAccount has empty keys but doing this will cause `DiscoverAccount` and `Register` to create a new dexConnection which would override the initial one in the conns map if all goes well. Going this route, I'd have to properly close the existing connection (stop book subscriptions etc) before replacing in the conns map. Dislike this approach primarily because we'd be setting up (and potentially promoting) a new connection when we have a connection that is almost g2g.

So to answer directly,
> What are the characteristics of a `dexAccount` created by the `AddDEX` method?

It has not been initialized via setupCryptoV2, but could be at a later time even though the dexAccount is still not ready for trading.

Perhaps, I should rename `a.isRegistered` to `a.readyToTrade` or `a.canTrade`? But is it really registered if the keys are generated by setupCryptoV2 but the fee isn't paid yet? I agree this check will need to change with bonds, but pre-bonds, it does make sense to me to rely more on whether a fee has been paid or not rather than on whether keys have been generated.

> The fee, bond, and auth related fields guarded by `authMtx` are not what we are interested in, I think (at least in most of the places you are using `a.isRegistered`).

I'll take another look at the call sites to confirm, but could any of such callers only care about the presence of generated keys even if the account may not have paid the required fee?

itswisdomagain:
> we really don't need to know about fee coins or bonds or anything like that; we just want to know if we've run `setupCryptoV2` and created account keys. I think that is what distinguishes a view-only account from one intended for authorization (regardless of payment details).

Hmm. A dexAccount that started as a view-only account (no keys setup) but later has keys setup is still a view-only account unless the fee is paid? If not, how would the markets page UI look for such a dexAccount?

client/core/types.go
chappjc: Things have changed quite a bit recently, and yes, it really is the account keys/identity that matters.  `feeCoin` is a deprecated field, and it is expected to be empty if you create an account with a bond, and the field will eventually be removed.  But _even if_ we only had the legacy registration fee system, it's still not the appropriate indicator of whether a `dexConnection` and it's associated `dexAccount` is a **view-only connection** (let's call the opposite an "active" conn/account).

You can have a conn/account that is converted to "active" from "view-only" that is not fully registered.  I will try to give reasoning for this.

> That was my first approach too, checking `a.id`, or `a.encKey`/`a.privKey`. But realized that those fields would be generated (by `DiscoverAccount -> setupCryptoV2`) if a registration is attempted but not completed. Thought about zeroing those fields if `DiscoverAccount` does not find a paid account and `Register` does not pay the required fee, but I worry things may fall through the crack with that approach. At least for a brief period, those keys won't be empty and the account will be thought to be registered (by `a.isRegistered`).

Calling either `Register` or `DiscoverAccount` should be understood to permanently convert a "view-only" account/connection to an "active"/"trading" account/connection (wording TBD).  This should be documented, but it is a reasonable expectation for the caller of these methods.  There's not much ambiguity in the intent of the caller of these methods.

If `Register` is "not completed" for whatever reason, that does not change the fact that the dex/account has been setup and is destined for activation on the DEX via either the legacy fee or posting bonds.  If we want it to revert to a "view-only" account on error, then yes, we'd have to zero this stuff out.  But depending on the failure, there is or should be handling like `reFee` to complete the registration without calling again, but I don't think that's central to the issue.

> > What are the characteristics of a `dexAccount` created by the `AddDEX` method?
> 
> It has not been initialized via setupCryptoV2, but could be at a later time even though the dexAccount is still not ready for trading.
> 
> Perhaps, I should rename `a.isRegistered` to `a.readyToTrade` or `a.canTrade`? But is it really registered if the keys are generated by setupCryptoV2 but the fee isn't paid yet? I agree this check will need to change with bonds, but pre-bonds, it does make sense to me to rely more on whether a fee has been paid or not rather than on whether keys have been generated.

Bond are *now*.  It's not a future concept anymore.  The browser UI doesn't do it yet, but you can create a new account with a bond via `dexcctl --simnet -p a postbond 127.0.0.1:17273 1000000000 42` today and `feeCoin` will always be empty.

Depending on the context, it does not matter how you registered (fee vs. bond) or the current account status (tier), just that you've configured the account and it's now intended for trading (`!viewOnly`) regardless of whether your account is tier>0 and thus "readyToTrade".

So, regarding "`a.readyToTrade` or `a.canTrade`", these aren't always the pertinent concepts.  Sometimes we want to know if the `dexConnection`+`dexAccount` a "view-only" or "trading" connection, and sometimes we want to know if tier>0 and thus can place orders, and sometimes like in `Register` we want to know if there is already a `feeCoin` (but the legacy fee is now a deprecated account creation method).  `(*Core).PostBond` does not care at all if `feeCoin` is set; only if the account identity and keys are initialized (not a view-only conn).  `authDEX` doesn't care if there's a `feeCoin` or live bonds, but it will fail if there's no `feeCoin` and the account's very first bond hasn't yet been activated (when server first makes the account entry in the DB), but after that `authDEX` will succeed with no `feeCoin` and even if there are no more active bonds -- tier will just be <1.

> > The fee, bond, and auth related fields guarded by `authMtx` are not what we are interested in, I think (at least in most of the places you are using `a.isRegistered`).
> 
> I'll take another look at the call sites to confirm, but could any of such callers only care about the presence of generated keys even if the account may not have paid the required fee?

We have to forget about the registration fee.  But even if that were the only method of account registration for the forseable future, yes, we would still not be concerned with the `feeCoin` but rather if the `dexAccount` were given an identity.  Only `Register` cares about `feeCoin`.  `(*dexConnection).exchangeInfo` should show the account ID if it's non-zero.  `AddDEX` should fail if there's a `dexAccount` with just the keys set because that could be a bonded account with no fee coin.  `DiscoverAccount` should fail if there's a `dexAccount` with keys *and* `authed()`, otherwise it's OK to try the `authDEX` loop.  `initializeDEXConnections` and `handleReconnect` can attempt `authDEX` for any account with keys for the same reasons.  `Trade` (via `connectedDEX`), etc. should only fail if tier<1.

chappjc:
> > we really don't need to know about fee coins or bonds or anything like that; we just want to know if we've run `setupCryptoV2` and created account keys. I think that is what distinguishes a view-only account from one intended for authorization (regardless of payment details).
> 
> Hmm. A dexAccount that started as a view-only account (no keys setup) but later has keys setup is still a view-only account unless the fee is paid? If not, how would the markets page UI look for such a dexAccount?

I think I just answered this in https://github.com/decred/dcrdex/pull/1986#discussion_r1041569456, but **no**, it is no longer a view-only account if it has it's keys set.  Fee is a legacy concept, and `feeCoin` is deprecated.  It is expected for accounts created via postbond to have no `feeCoin` ever set.  Further, it's expected to have no live bonds if they've all expired.  Neither of these things means the account is not intended for trading, or non-existent on the server.  Trading should only be blocked if tier<1.

As far as how the UI looks, not that with the legacy registration fee, there's a "pending" state between when you broadcast the fee transaction and when it gets confirmed and then submitted to the DEX.  Similarly, with bonds, if you create an account with a bond (`'postbond'`), there's a period where the client waits for confirmations before submitting the bond for acceptance.  In both cases, the UI has to have special handling.  For the legacy fee, it's the confirmation countdown that takes the place of the order form.  With bonds, it needs to be something indicating "tier zero, you must post new bonds", and that state can come and go as bond are posted and expire.

itswisdomagain:
Your explanations above clarify things for me @chappjc, thanks.

> I think we're conflating two concepts: (a) being registered at the DEX host, and (b) having a generated account identity/keys for authentication/authorization purposes regardless of registration and fee/bond status.

So, this PR adds a third concept (c) view-only DEX (not registered, no keys).

Among other changes in my last commit,
- authDEX continues to be done for (a) and (b), excluding only (c)
- markets page shows "Create account to trade" prompt even if keys have been generated as long as the account is not yet registered with the server.

JoeGruffins (commented):
Starting fresh on simnet not working

image

itswisdomagain:
> Starting fresh on simnet not working

Silly if statement bug. Fixed.

chappjc:
> Also, a "tier zero, bond posted, waiting confirmations" state.
> 
> Is the UI bits something you're working on in your next bond PR @chappjc?

I'm starting the next "in progress" item on the project page here: https://github.com/orgs/decred/projects/2/views/1

chappjc:
Next bond steps include UI, updated messaging (suspended -> sub-tier or whatever), client bond maintenance, settings, switches, etc.

I'm starting client db and core work for maintaining a certain tier with limits on total bonding amount.

For the UI, what we do depends on if we intend to support both the legacy fee and bonds in the next release.  I don't think we should give the choice, only bonds.

So, while we transition in development, I think we should start by showing two options at /register - one that uses the legacy fee and one that uses a bond.   But since it's only for developers to use either as needed, I don't think we should make it a single polished form with a selector.  In fact, two separate forms shown simultaneously where you just use one of the forms might make the most sense.  Otherwise if we make a nice selector on a single form, we're going to have to change that when the legacy fee is dropped.  If there are just two forms shown, we just remove one when it's time to drop the fee.

client/core/types.go
```diff
@@ -735,6 +737,25 @@ func (a *dexAccount) ID() account.AccountID {
...
+// if the payment/bond is yet to be confirmed.
+func (a *dexAccount) isRegistered() bool {
+	a.authMtx.RLock()
+	defer a.authMtx.RUnlock()
+
+	return a.isAuthed || // 'connect' succeeded, so fee is paid or initial bond is posted+confirmed
+		len(a.feeCoin) > 0 || // fee is paid but either yet to 'connect' or is waiting confirmations
+		len(a.bonds) > 0 || // initial bond is posted+confirmed, probably hasn't 'connect'ed yet
+		len(a.pendingBonds) > 0 || // initial bond is posted, waiting confirmations
+		len(a.expiredBonds) > 0 // initial bond _was_ posted and confirmed
```
chappjc, line 756: Refunded bonds appear in none of these slices, so this `isRegistered` can have false negatives.  I'm toying with a `registered` flag for `dexAccount` that is set according to DB info.  See this `AccountInfo` method and toggling to true when register or bondconfirmed succeeds: https://github.com/decred/dcrdex/commit/91d612432c553c786d5d3647d358ec12beace507

itswisdomagain: I initially thought that was a commit from one of your bond in-progress work but later realized it was a commit on top of my last commit in this PR. Tried adapting it and ensuring there are no cracks in the logic when I realized that the purpose of the `acct.isRegistered()` method is to prevent duplicate legacy registration and possibly to also prevent needless account discovery.

The only 3 callers of that method use the returned value to determine if to
- display the "create account to trade" button on the frontend
- prevent account discovery attempts
- prevent legacy registration attempts

For each of these cases, the `isRegistered` value would need to also consider pending payments (fee / bond txs that have been broadcast but not yet confirmed).

There's also a potential for false negatives with the db bonds check, for a registered account that was recovered with no active bonds, but that's less of a concern. The db `AccountInfo.Registered()` method would return false if the account has no legacy fee payment and no active bonds reported by the server on recovery, until and unless the user posts another bond.

JoeGruffins (approved):
Working great.

One nit is that the viewing dex will appear in the settings under `Registered Dexes`:
image

None of the functions there seem to work, so maybe just don't show it there if possible.

client/webserver/site/src/html/forms.tmpl
```diff
@@ -185,6 +193,20 @@
 <div class="fs15 pt-3 text-center d-hide errcolor text-break" data-tmpl="err"></div>
 {{end}}
 
+{{define "discoverAcctForm"}}
+<div class="py-1 text-center fs28 sans-light offwhite lh1" data-tmpl="header">[[[Create Account]]]</div>
+<div class="fs16 text-center mb-4">on <span data-tmpl="dexHost"></span></div>
+<div data-tmpl="appPWBox" class="pb-2">
+  <label for="discoverAcctAppPW" class="form-label ps-1 mb-1">[[[App Password]]]</label>
+  <div class="fs14 px-1 mb-1">Your app password is always required when performing sensitive operations.</div>
```
chappjc, line 201: If we keep this message, it needs a locale template entry.  However, I don't think it's needed.  Regarding the message, other operations only show the password text box if the pass is not cached, so why is this form more sensitive than e.g. registration?

chappjc (changes requested):

client/db/types.go
```diff
@@ -184,7 +184,19 @@ type AccountInfo struct {
 	LegacyFeeCoin    []byte
 	LegacyFeeAssetID uint32
 	// LegacyFeePaid should be set on retrieval if there is an AccountProof set.
-	LegacyFeePaid bool // DEPRECATED
+	LegacyFeePaid bool // DEPRECATED LegacyFeeConfirmed
```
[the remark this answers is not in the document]
itswisdomagain: Ah, creeped in. Will remove.

client/db/types.go
```diff
@@ -184,7 +184,19 @@ type AccountInfo struct {
...
+// Registered indicates if the account should exist server-side. This is when
+// the legacy fee txn or at least one bond txn has been broadcast even if
+// neither has received the required confirmations. This may return a false
+// negative for a recovered account that had no active bonds at recovery time,
+// unless new bond(s) have been posted since recovery.
+func (ai *AccountInfo) Registered() bool {
+	if ai.LegacyFeePaid || len(ai.LegacyFeeCoin) > 0 {
+		return true
+	}
+	return len(ai.Bonds) > 0
```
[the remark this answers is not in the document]
itswisdomagain: I see. Yeah, there's a need to discern if a bond tx or the legacy fee tx has been broadcast to prevent re-attempting a new legacy tx. But that's about it, and since legacy registration will soon be a thing of the past, I'm beginning to wonder if there's any need to account for this.

The only thing we'd need to know is if there's NO active/pending bond and then display a prompt on the UI to post a bond to trade. If there's a pending bond, we display a confirmations counter.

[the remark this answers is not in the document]
chappjc: Alright, sounds fine for now.  Will test it all out today.

chappjc:
Sorry, didn't mean to request review from you @buck54321.  Feel free to review of course, but no urgent need.

client/core/account.go
```diff
@@ -209,9 +209,28 @@ func (c *Core) AccountImport(pw []byte, acct *Account, bonds []*db.Bond) error {
 // UpdateCert attempts to connect to a server using a new TLS certificate. If
 // the connection is successful, then the cert in the database is updated.
 func (c *Core) UpdateCert(host string, cert []byte) error {
+	var viewOnlyConn bool
 	accountInfo, err := c.db.Account(host)
 	if err != nil {
-		return err
+		if err != db.ErrAcctNotFound {
```
buck54321, line 215: `errors.Is`?

client/webserver/site/src/html/forms.tmpl
```diff
@@ -140,8 +140,16 @@
 {{define "dexAddrForm"}}
 <div class="form-closer hoverbg"><span class="ico-cross"></span></div>
 <div data-tmpl="addDexHdr" class="flex-center fs24 sans-light"><span class="ico-plus fs12 me-2"></span> [[[Add a DEX]]]</div>
+<div class="pt-2 fs14">
+  <hr class="dashed my-2">
+  <input class="form-check-input" type="checkbox" data-tmpl="skipRegistration" id="skipRegistration">
+  <label for="skipRegistration" class="ps-1">[[[Skip Registration]]]</label>
+</div>
```
buck54321, line 147: Can we move this below the quick-select buttons?

client/core/core.go
```diff
@@ -3534,6 +3545,59 @@ func (c *Core) GetDEXConfig(dexAddr string, certI interface{}) (*Exchange, error
...
+
+	// Don't allow adding another dex with the same pubKey. There can only be
+	// one dex connection per pubKey. UpdateDEXHost must be called to connect to
+	// the same dex using a different host name.
+	exists, host := c.dexWithPubKeyExists(dc.acct.dexPubKey)
+	if exists {
+		dc.connMaster.Disconnect()
+		return newError(dupeDEXErr, "already connected to DEX at %s but with different host name %s", dexAddr, host)
+	}
+
```
buck54321, line 3593: Ohhhh. We're not storing the watch-only account in the DB. So if the user had a registered account on server A, and wanted to also see server B view-only, they would need to go to the settings page and add it after starting dexc every time. Hmm.

buck54321 (changes requested):

client/core/core.go
chappjc: Haven't actually tested this PR yet, but that would be less than ideal UX.
Two options:
- remember watch-only in DB
- use a cookie to remember the desired view-only servers, and they get re-added on-demand

itswisdomagain: Didn't like the idea of storing watch-only account in DB because once added, how's it removed? Although, disable account can be used for that perhaps? Will look into adding to DB, but first, since legacy reg will likely remain past 0.6, I will update the relevant parts of this to consider both legacy and bond payments.

chappjc: > Will look into adding to DB,

Would it be messy to have the frontend re-add the view-only hosts after login, using cookies to remember which view-only hosts were previously added?

itswisdomagain: Maybe not. Was thinking db since it's likely lesser work. But will look into using cookies.

chappjc: That's fine.

I'm also pondering if we eventually want to take this view-only idea one step further and not even require app initialization.  If you just download the client to check out the markets and it wants you to start setting passwords you might walk away.   Obviously that would mean not adding the view-only DEX host to a DB.  Completely stateless on the backend.

buck54321: > I'm also pondering if we eventually want to take this view-only idea one step further and not even require app initialization. If you just download the client to check out the markets and it wants you to start setting passwords you might walk away. Obviously that would mean not adding the view-only DEX host to a DB. Completely stateless on the backend.

That's f****** brilliant!

buck54321: > Obviously that would mean not adding the view-only DEX host to a DB. Completely stateless on the backend.

Does it?

chappjc: Oh, right.  `InitializeClient` is just for the credentials.  Just constructing `Core` with `core.New` does indeed ready the DB file.  Alright then.

itswisdomagain: I'll start with saving to DB. Not requiring init for view-only connection will follow.

client/webserver/http.go
```diff
@@ -83,16 +83,19 @@ func (s *WebServer) handleLogin(w http.ResponseWriter, r *http.Request) {
 // registerTmplData is template data for the /register page.
 type registerTmplData struct {
 	CommonArguments
-	Initialized    bool
-	Authed         bool
 	KnownExchanges []string
+	// Host is optional. If provided, the register page will not display the add
+	// dex form, instead this host will be pre-selected for registration.
+	Host string
```
itswisdomagain, line 89: If #2025 goes in first (I think it should), I'd use the dexsettings page for this instead of the register page. It would already contain the mechanics for posting bond for an existing dex connection and would naturally fit.

buck54321 (commented):
Testing well.

client/webserver/site/src/js/forms.ts
```diff
@@ -1558,6 +1582,74 @@ export class DEXAddressForm {
...
+    bind(form, page.submit, () => this.checkDEX())
+
+    this.refresh()
+  }
+
+  refresh () {
+    const page = this.page
+    page.appPW.value = ''
+    Doc.hide(page.err)
+    const hidePWBox = State.passwordIsCached() || (this.pwCache && this.pwCache.pw)
```
buck54321, line 1610: Could we skip this form altogether if the password is cached?

buck54321 (approved):

PR #1987, opened 2022-12-07
client/db/bolt: more aggressive DB compaction

chappjc:
```
client/db/bolt: more aggressive DB compaction

This makes the DB compaction smarter, considering all free/unused
bytes regardless of their sources. Previously it was only considering
FreePageN, which can be was only a small part of the unused space,
making compaction very unlikely for most DBs in practice.

This also creates a relative free space threshold (5%) in addition to
an absolute bytes free space threshold (256 KiB). Free space must
exceed *both* thresholds to trigger a compaction. This is to spare
long and large compactions happening too frequently for very large
DB files.

The logging that reports the percent space reclaimed is also fixed
since it was reporting 1-pct instead of pct.

Example of new checks and logs with a large testnet DB:

CORE[DB]: Total DB size 1057017856 bytes, 328340544 bytes unused (31.06%)
CORE[DB]: Compacting database to reclaim at least 328340544 bytes...
CORE[DB]: Compacted database file from 1057017856 => 507797504 bytes (51.96% reduction)

In this case, the actual reclaimed amount was larger than the
estimate from the unused amount reported by bolt, presumably because
of orphaned junk sectors of the file created due to crashes.
However, testing with repeated compactions and other databases
indicates the actual reduction amounts are almost exactly the
estimates. Also, the percentages logged are now correct.
```

### DIFF
--- a/client/db/bolt/db.go
+++ b/client/db/bolt/db.go
@@ -220,8 +220,26 @@ func (db *BoltDB) Run(ctx context.Context) {
 		db.log.Errorf("Unable to backup database: %v", err)
 	}
 
-	// Only compact the current DB file if there are excessive free pages.
-	if db.Stats().FreePageN < 32 { // 128 KiB for 4096 page size
+	// Only compact the current DB file if there is excessive free space, in
+	// terms of bytes AND relative to total DB size.
+	const byteThresh = 1 << 18 // 256 KiB free
+	const pctThresh = 0.05     // 5% free
+	var dbSize int64
+	_ = db.View(func(tx *bbolt.Tx) error {
+		dbSize = tx.Size() // db size including free bytes
+		return nil
+	})
+	dbStats := db.Stats()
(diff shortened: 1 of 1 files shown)


client/db/bolt/db.go
```diff
@@ -220,8 +220,26 @@ func (db *BoltDB) Run(ctx context.Context) {
 		db.log.Errorf("Unable to backup database: %v", err)
 	}
 
-	// Only compact the current DB file if there are excessive free pages.
-	if db.Stats().FreePageN < 32 { // 128 KiB for 4096 page size
+	// Only compact the current DB file if there is excessive free space, in
+	// terms of bytes AND relative to total DB size.
+	const byteThresh = 1 << 18 // 256 KiB free
+	const pctThresh = 0.05     // 5% free
```
chappjc, line 226: Note that `pctThresh` kicks in, effectively taking the place of `byteThresh`, at ~5 MB DB size.

JoeGruffins (approved):

Issue #1959, opened 2022-11-15
underflow in overlock savings estimate


chappjc:
![image](https://user-images.githubusercontent.com/9373513/202008823-9dafa5e0-5fb5-46ec-825a-77b7d77e5428.png)

chappjc:
Taking this one.

PR #1988, opened 2022-12-08
client: fix split option handling

chappjc:
This fixes a bug in the asset backends' `estimateSwap` handling when trying a split transaction.

It was possible for the estimates to permit/allow a split even when it would consume more than without a split transaction. The symptom of this was that the "locked" amount estimates obtained when requesting a split could be higher than if not using a split, resulting in an integer underflow in the caller when comparing the two amounts.

Resolves https://github.com/decred/dcrdex/issues/1959

[client/asset: fix estimateSwap split rejection](https://github.com/decred/dcrdex/commit/dac81568edf388767266f0bb71de745f6c06115f)

```
The enough func in estimateSwap does not account for a split transaction
at the start, so it is possible that funding for trySplit would actually
choose more UTXOs. Actual order funding accounts for this. For this
estimate, we will just not use a split tx if the split-adjusted required
funds exceeds the total value of the UTXO selected with this enough
closure.

The fix is to reject a spit transaction if the output amount plus split tx
fees is less than the sum of the amounts of the utxos selected by fund,
rather than the total of the available utxos provided to fund initially.
There is no point to calling fund again with a different enough func
that accounts for the cost split tx because this indicates it would
reqiure an additional UTXO, thus making the spit txn wasteful.
```

--- 

This also changes how the split order option is emitted from `PreSwap` so that it is never omitted, instead given a default of false and a reason explaining why it is not effective (wasteful) for the proposed order.  This is better than the option not appearing since the UI should be consistent for a given form and wallet. (What happened to the Pre-size option?  Sometimes it's just not there.)

[client/asset: show split option when ineffectual with reason](https://github.com/decred/dcrdex/commit/c209a419269384e004d51f73de610551567823d6)

image

image

---

Finally, this also fixes a bug in how core was obtaining `PreOrder` results for market orders with the spit option enabled.  Split transactions are supposed to be disabled/skipped for any order that is immediate (_not_ standing), which includes both market orders and limit orders with immediate time-in-force, as described by the `PreSwapForm` docs.  This fixes how `PreSwapForm.Immediate` is set by Core so that it is true for market orders as well as immediate limit orders.

[client/core: market orders are immediate (PreOrder)](https://github.com/decred/dcrdex/commit/ef0f1c57e1b42e4aa7be7e7ae0ac2b08ec3cfa79)

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -1706,7 +1706,7 @@ func (btc *baseWallet) PreSwap(req *asset.PreSwapForm) (*asset.PreSwap, error) {
 	}
 
 	// Get the estimate using the current configuration.
-	est, splitUsed, _, err := btc.estimateSwap(req.Lots, req.LotSize, req.FeeSuggestion,
+	est, _, _, err := btc.estimateSwap(req.Lots, req.LotSize, req.FeeSuggestion,
 		req.MaxFeeRate, utxos, split, bump)
 	if err != nil {
 		return nil, fmt.Errorf("estimation failed: %v", err)
@@ -1714,9 +1714,8 @@ func (btc *baseWallet) PreSwap(req *asset.PreSwapForm) (*asset.PreSwap, error) {
 
 	var opts []*asset.OrderOption
 
-	// If the used split isn't the requested split, the other split option was
-	// unavailable, so there is no option to offer.
-	if !req.Immediate && splitUsed == split {
+	// Only offer the split option for standing orders.
+	if !req.Immediate {
 		if splitOpt := btc.splitOption(req, utxos, bump); splitOpt != nil {
 			opts = append(opts, splitOpt)
 		}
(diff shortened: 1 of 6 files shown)


client/asset/btc/btc.go
```diff
@@ -1935,23 +1939,21 @@ func (btc *baseWallet) estimateSwap(lots, lotSize, feeSuggestion, maxFeeRate uin
...
 	// Math for split transactions is a little different.
 	if trySplit {
-		_, extraMaxFees := btc.splitBaggageFees(bumpedMaxRate)
+		_, splitMaxFees := btc.splitBaggageFees(bumpedMaxRate)
 		_, splitFees := btc.splitBaggageFees(bumpedNetRate)
-		locked := val + maxFees + extraMaxFees
-
-		if avail >= reqFunds+extraMaxFees {
+		reqTotal := reqFunds + splitMaxFees // ~ rather than actually fund()ing again
+		if reqTotal <= sum {
```
chappjc, line 1951: Comparing against `avail` instead of `sum` was the primary bug.  If the required amount is more than the amount in the funding coins selected with no split transaction, it is pointless to do a split because it would require another UTXO to be selected.   This is described in the multi-line comment around line 1911.

client/asset/btc/btc.go
```diff
@@ -1935,23 +1939,21 @@ func (btc *baseWallet) estimateSwap(lots, lotSize, feeSuggestion, maxFeeRate uin
...
 			return &asset.SwapEstimate{
 				Lots:               lots,
 				Value:              val,
-				MaxFees:            maxFees + extraMaxFees,
+				MaxFees:            maxFees + splitMaxFees,
 				RealisticBestCase:  estLowFees + splitFees,
 				RealisticWorstCase: estHighFees + splitFees,
-			}, true, locked, nil
+			}, true, reqFunds, nil // requires reqTotal, but locks reqFunds in the split output
```
chappjc, line 1958: This was another bug, but not very important.  The amount locked by the funding coins for an order funded by a split transaction is precisely the value of the output of the split transaction i.e. reqFunds.  The "extra" fees are just burned in making the split transaction, so it's not correct to also call them locked since they are included in the various fee estimates as well.

The tests didn't check for this because we don't test estimateSwap directly, and this locked amount goes nowhere except the test messages in the generated config opt for the UI.  The `checkSwapEstimate` helper hand an unused `locked` arg and I can see why it was unused.  I'm not planning to add new tests for this however.

client/asset/btc/btc.go
```diff
@@ -1853,40 +1852,43 @@ func (btc *baseWallet) splitOption(req *asset.PreSwapForm, utxos []*compositeUTX
...
-			Reason: reason,
-		},
+		Boolean: &asset.BooleanConfig{},
 	}
+
+	if !splitUsed || splitLocked >= noSplitLocked { // locked check should be redundant
+		opt.Boolean.Reason = fmt.Sprintf("avoids no %s overlock for this order (ignored)", symbol)
+		opt.Description = fmt.Sprintf("A split transaction for this order avoids no %s overlock, "+
+			"but adds additional fees.", symbol)
+		return opt // do not show by default, but explain why
```
chappjc, line 1872: Rather than having a missing option, I believe it is a better UX to show it, unchecked, and with an appropriate reason and description.  Otherwise, the user is wondering where the option went that they normally use.

chappjc:
Not tested as thoroughly as I'd normally do before making ready for review, but it does address the underflow bug.

JoeGruffins (approved):
Working well.

martonp (approved):

chappjc:
Pls see if you're OK with the conceptual changes to the frontend regarding this order option @buck54321 .  Reasoning for not just vanishing the option from the frontend when likely inapplicable: https://github.com/decred/dcrdex/pull/1988#discussion_r1042858820

PR #1994, opened 2022-12-12
spec/comms: Update connect response.

JoeGruffins:
spec/comms: Update connect response.

    README: Remove mediawiki version.

Part of #1641

### DIFF
--- a/spec/comm.mediawiki
+++ b/spec/comm.mediawiki
@@ -56,7 +56,10 @@ with unit ''atoms Y / unit Z''.
 In order to demonstrate control of unspent value on the blockchain, a user must
 provide its location. For Bitcoin-based blockchains, value is located by
 pointing to an unspent transaction output (UTXO), identified by its transaction
-ID and output index (vout).
+ID and output index (vout). For account based assets, the coin id is different
+in different situations. Pre-swap it is simply an account address. While swapping,
+the value is transferred to a swap contract's address and can be found there in
+a map keyed by the swap's secret hash.
 
 In an effort to stay blockchain-protocol agnostic, the DEX accepts
 and recognizes the locating information as a single byte-array called the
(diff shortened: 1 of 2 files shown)


JoeGruffins:
Removing the `README.mediawiki` in the root dir because I don't think it has a purpose. We do not reference it anywhere and the github homepage points to `README.md`. Maybe there is another reason for it existing? Can pop off the commit if so.

spec/comm.mediawiki
```diff
@@ -189,13 +192,19 @@ of their penalization.
...
-| matches || &#91;object&#93; || list of [[orders.mediawiki/#Match_negotiation|Match objects]]
+| activematches || <nowiki>[object]</nowiki> || list of active [[orders.mediawiki/#Match_negotiation|Match objects]]
 |-
-| order statuses || <nowiki>[object]</nowiki> || list of <code>Order Status Object</code>
+| activeorderstatuses || <nowiki>[object]</nowiki> || list of active <code>Order Status Object</code>
 |-
 | score || int || The user's score. The math behind this is currently being adjusted
 |-
-| suspended || bool || true if suspended. will be deprecated in a future version
+| tier || int || Bond tier. One or more indicates the account can trade. Not set for legacy accounts.
```
chappjc, line 201: This would be "Not set by legacy servers." because it is set even for legacy accounts if the server has updated.

A legacy account (that paid the old reg fee) gets a free and perpetual tier +1, as you've noted below on `legacyFeePaid`.

spec/comm.mediawiki
```diff
@@ -189,13 +192,19 @@ of their penalization.
...
+|-
+| activeBonds || <nowiki>[object]</nowiki> || list of active <code>Bond Object</code>
+|-
+| legacyFeePaid || bool || Whether the legacy fee was paid. This currently has the internal value of one bond tier. Not set for legacy servers.
+|-
+| suspended || bool || DEPRECATED. For legacy servers true if suspended.
```
chappjc, line 207: Let's also add that this also implies `tier<1`, and it just means that the user cannot trade without posting new bond.

chappjc (changes requested):
Thanks for updating the connect response for bonds!

spec/comm.mediawiki
JoeGruffins: If the server can handle bonds, will this be set? They can't post bonds on the legacy server.

spec/comm.mediawiki
JoeGruffins: Updated the wording.

spec/comm.mediawiki
JoeGruffins: Updated the wording.

JoeGruffins:
updated wording https://github.com/decred/dcrdex/compare/fe812e5a8d81d4a0ab1a9748678d3d937d7e5e3f..d8665a5270c669917d7809ca705404f806663d73

spec/comm.mediawiki
chappjc: > If the server can handle bonds, will this be set? They can't post bonds on the legacy server.

Yes, for backward compatibility with legacy clients.

However, it's looking more and more like the release with bonds will fully kill the legacy fee, and this backward compatibility is just a transition.

chappjc (approved):

PR #1995, opened 2022-12-13
meterer: Do not wait for stopped timer.

### LINKED ISSUE #1984: testing: go ci test timeout in `server/swap`
https://github.com/decred/dcrdex/actions/runs/3616423483/jobs/6094364116

JoeGruffins:
closes #1984

please see https://go.dev/play/p/U9d_pvnMBu6

As called out in the docs https://pkg.go.dev/time#Timer.Stop It does not close the channel. If the timer has already fired, nothing happens. The example says "assuming the program has not received from t.C already"

So anyway, we are not reusing the timer, so I don't think we need to worry about it.

### DIFF
--- a/server/swap/swap.go
+++ b/server/swap/swap.go
@@ -799,7 +799,7 @@ func (s *Swapper) Run(ctx context.Context) {
 						return
 					}
 					select {
-					case <-mainLoop:
+					case <-ctxHelpers.Done():
 						return
 					case blockNotes <- &blockNotification{
 						time:    time.Now().UTC(),
--- a/dex/meter/meterer.go
+++ b/dex/meter/meterer.go
@@ -10,7 +10,7 @@ import (
 
 // DelayedRelay creates a simple error signal pipeline that delays and
 // aggregates the relaying of nil errors. Non-nil errors received on the in
-// channel are immediately send on the out channel without delay. If a nil error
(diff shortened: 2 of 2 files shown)


server/swap/swap.go
```diff
@@ -799,7 +799,7 @@ func (s *Swapper) Run(ctx context.Context) {
 						return
 					}
 					select {
-					case <-mainLoop:
+					case <-ctxHelpers.Done():
```
JoeGruffins, line 802: This seems unreachable since `mainLoop` is not closed until after all the helpers have returned

chappjc:
Good catch.  We're following documented pattern too blindly.  `"This cannot be done concurrent to other receives from the Timer's channel or other calls to the Timer's Stop method."` is the relevant bit.  While at a glance it may seem there's no concurrent receive since we just have `Timer` used in the loop, but the `AfterFunc` receives on it in it's own goroutine.

As you said, we're not reusing the timer after stopping it, so this is a good fix.

server/swap/swap.go
chappjc: Looks right. `mainLoop` is intended to be closing the, uh, main loop, not this helper that must die *first*.

chappjc (approved):

PR #1997, opened 2022-12-16
client: add login notes to improve UX

### LINKED ISSUE #1434: app: login spinner needs accompanying text
This is not an error, but it can do this much longer now because it's starting up the built-in wallets.  Even though this isn't syncing the wallets, it can still take time to fire things up.

image

Also if auth or some other network request times out, this can take a long time. Users may be tempted to kill dexc.

There should be some kind of information here to communicate what is going on at this point and to wait up to a minute or so if there are built-in wallets.

ukane-philemon:
This adds `LoginNote` which is sent to the UI during login. These `LoginNote`s improve the UX during login, especially during the first login. Closes #1434.
After:
Screenshot 2022-12-16 at 1 47 05 AM

Screenshot 2022-12-16 at 1 47 23 AM

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -4268,6 +4268,7 @@ func (c *Core) Login(pw []byte) error {
 		return fmt.Errorf("cannot log in because app has not been initialized")
 	}
 
+	c.notify(newLoginNote("Verifying login credentials..."))
 	if len(creds.EncInnerKey) == 0 {
 		err := c.initializePrimaryCredentials(pw, creds.OuterKeyParams)
 		if err != nil {
@@ -4293,8 +4294,11 @@ func (c *Core) Login(pw []byte) error {
 		// resolveActiveTrades. We won't try to unlock here, but if the wallet
 		// is needed for active trades, it will be unlocked in resolveActiveTrades
 		// and the balance updated there.
+		c.notify(newLoginNote("Connecting wallets..."))
 		c.connectWallets()
+		c.notify(newLoginNote("Resolving active trades..."))
 		c.resolveActiveTrades(crypter)
+		c.notify(newLoginNote("Initializing DEX connections..."))
 		c.initializeDEXConnections(crypter)
 		if err = c.loadBotPrograms(); err != nil {
 			c.log.Errorf("Error loading bot programs: %v", err)
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -4268,7 +4268,7 @@ func (c *Core) Login(pw []byte) error {
 		return fmt.Errorf("cannot log in because app has not been initialized")
 	}
 
-	c.notify(newLoginNote("Verifying login credentials..."))
+	c.notify(newLoginNote("Verifying credentials..."))
 	if len(creds.EncInnerKey) == 0 {
 		err := c.initializePrimaryCredentials(pw, creds.OuterKeyParams)
 		if err != nil {
@@ -4296,9 +4296,9 @@ func (c *Core) Login(pw []byte) error {
 		// and the balance updated there.
 		c.notify(newLoginNote("Connecting wallets..."))
 		c.connectWallets()
-		c.notify(newLoginNote("Resolving active trades..."))
+		c.notify(newLoginNote("Resuming active trades..."))
 		c.resolveActiveTrades(crypter)
-		c.notify(newLoginNote("Initializing DEX connections..."))
+		c.notify(newLoginNote("Connecting to DEX servers..."))
 		c.initializeDEXConnections(crypter)
 		if err = c.loadBotPrograms(); err != nil {
 			c.log.Errorf("Error loading bot programs: %v", err)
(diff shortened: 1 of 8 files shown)


client/core/core.go
```diff
@@ -4268,6 +4268,7 @@ func (c *Core) Login(pw []byte) error {
 		return fmt.Errorf("cannot log in because app has not been initialized")
 	}
 
+	c.notify(newLoginNote("Verifying login credentials..."))
```
chappjc, line 4271: Just "Verifying credentials..." is good.

client/core/core.go
```diff
@@ -4293,8 +4294,11 @@ func (c *Core) Login(pw []byte) error {
 		// resolveActiveTrades. We won't try to unlock here, but if the wallet
 		// is needed for active trades, it will be unlocked in resolveActiveTrades
 		// and the balance updated there.
+		c.notify(newLoginNote("Connecting wallets..."))
 		c.connectWallets()
+		c.notify(newLoginNote("Resolving active trades..."))
```
chappjc, line 4299: "Resolving" is unclear in meaning.  For the user, I think we should just say "Loading" or "Preparing" or "Resuming".  I'm leaning toward "Resuming".

client/core/core.go
```diff
@@ -4293,8 +4294,11 @@ func (c *Core) Login(pw []byte) error {
...
 		c.resolveActiveTrades(crypter)
+		c.notify(newLoginNote("Initializing DEX connections..."))
```
chappjc, line 4301: "Initializing" is also vague to the user.  "Connecting to DEX servers.."

chappjc (changes requested):
Works great!  A few rewording requests.

Also it would be best if we could somehow have the loader and the new message not on top of other stuff, so it can be more legible.

chappjc:
Also bump cache the JS buster please.

ukane-philemon:
> Also it would be best if we could somehow have the loader and the new message not on top of other stuff, so it can be more legible.

The loader prevents interaction with the page until its done, but I think we can blur the background to make the text stand out.

JoeGruffins (approved):

chappjc (approved):

client/core/notification.go
```diff
@@ -613,3 +614,18 @@ func newBotNote(topic Topic, subject, details string, severity db.Severity, repo
...
+	db.Notification
+	Message string `json:"msg"`
+}
+
+const TopicLoginStatus Topic = "LoginStatus"
+
+func newLoginNote(message string) *LoginNote {
+	return &LoginNote{
+		Notification: db.NewNotification(NoteTypeLogin, TopicLoginStatus, "", "", db.Data),
+		Message:      message,
```
chappjc, line 629: Doesn't look like we need to expand on the `db.Notification`.

```go
// LoginNote is a notification with the recent login status.
type LoginNote struct {
	db.Notification
}

const TopicLoginStatus Topic = "LoginStatus"

func newLoginNote(message string) *LoginNote {
	return &LoginNote{
		db.NewNotification(NoteTypeLogin, TopicLoginStatus, message, "", db.Data),
	}
}
```

Frontend can just use `CoreNote`.

client/webserver/site/src/js/login.ts
```diff
@@ -13,6 +13,16 @@ export default class LoginPage extends BasePage {
     Doc.show(this.form)
     this.loginForm = new LoginForm(this.form, () => { this.loggedIn() })
     this.loginForm.focus()
+
+    app().registerNoteFeeder({
+      login: (note: LoginNote) => { this.handleLoginNote(note) }
+    })
```
chappjc, line 19: It took me a while to figure out why I wasn't getting the notes when using Login from `/register`.
I think we need to have this handler registered from `LoginForm` instead of `LoginPage`.

```diff
diff --git a/client/webserver/site/src/js/forms.ts b/client/webserver/site/src/js/forms.ts
index 5f738eb5a..cea6f1a2b 100644
--- a/client/webserver/site/src/js/forms.ts
+++ b/client/webserver/site/src/js/forms.ts
@@ -22,7 +22,8 @@ import {
   WalletStateNote,
   WalletInfo,
   Token,
-  WalletCreationNote
+  WalletCreationNote,
+  CoreNote
 } from './registry'
 import { XYRangeHandler } from './opts'
 
@@ -1573,12 +1574,21 @@ export class LoginForm {
     this.headerTxt = page.header.textContent || ''
 
     bind(form, page.submit, () => { this.submit() })
+
+    app().registerNoteFeeder({
+      login: (note: CoreNote) => { this.handleLoginNote(note) }
+    })
   }
 
   focus () {
     this.page.pw.focus()
   }
 
+  handleLoginNote (n: CoreNote) {
+    const loginMsg = Doc.idel(this.form, 'loaderMsg')
+    if (loginMsg) loginMsg.textContent = n.subject
+  }
+
   async submit () {
     const page = this.page
     Doc.hide(page.errMsg)
```

Any down sides to that?

ukane-philemon: > It took me a while to figure out why I wasn't getting the notes when using Login from /register.

Oops, thanks.

> Any down sides to that?

Nah, all good. The login note handler will be deregistered when a new page loads.

PR #1998, opened 2022-12-16
spec: Update README.

JoeGruffins:
This only updates the readme for the current spec. Some places will still need to change like registration but hopefully that is ok as incremental changes.

Unexpectedly, none of the links with spaces scrolled. I am pretty sure that these worked before, so perhaps browser handling changed at some point? I have tested on firefox and chrome and it seems that underscores do not work, they need to be hyphens. Please confirm this behavior. Here are the changes on my branch https://github.com/JoeGruffins/dcrdex/tree/updatespecadmin/spec

Part of #1641

### DIFF
--- a/spec/README.mediawiki
+++ b/spec/README.mediawiki
@@ -23,7 +23,8 @@ While they remove the trusted third party (TTP), they insert their own products
 as a means to capture trading fees, which replaces the TTP friction with a
 new platform friction.
 
-A more robust solution is based on atomic swap technology [https://github.com/decred/atomicswap &#91;2&#93;],
+A more robust solution is based on [https://github.com/decred/atomicswap atomic swap technology]
+[[references.mediawiki|<nowiki>[2]</nowiki>]],
 which enables trustless exchange directly between wallets.
 Until recently, few blockchains supported atomic swaps, but the past few years
 have seen widespread adoption, and such an exchange is now possible.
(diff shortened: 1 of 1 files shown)


spec/README.mediawiki
```diff
@@ -38,10 +39,10 @@ censorship and unfair matching difficult.
 
 While trades are fully trustless, steps are taken to discourage malicious
 clients from hindering normal trade activity.
-All clients pay a non-refundable registration fee.
+All clients must lock coins in bonds to trade.
 Registered clients are then bound to the
 [[community.mediawiki|rules of community conduct]], violation of which typically
-results in loss of trading privileges up to and including a permanent ban.
+results in loss of trading privileges.
```
JoeGruffins, line 45: Correct? I think the concept of permanent ban is out?

spec/README.mediawiki
```diff
@@ -23,7 +23,8 @@ While they remove the trusted third party (TTP), they insert their own products
 as a means to capture trading fees, which replaces the TTP friction with a
 new platform friction.
 
-A more robust solution is based on atomic swap technology [https://github.com/decred/atomicswap &#91;2&#93;],
+A more robust solution is based on [https://github.com/decred/atomicswap atomic swap technology]
+[[references.mediawiki|<nowiki>[2]</nowiki>]],
```
JoeGruffins, line 27: changing unreadable code to `<nowiki></nowiki>` as it makes more sense for humans.

chappjc (approved):

PR #2002, opened 2022-12-20
ui: uniform responsive checkbox

norwnd:
Trying to improve checkbox UX a bit in this PR, namely:
- define common style for checkbox-type inputs so that it is consistent across different forms (if necessary, particular form can adjust/redefine it ofc)
- lower check-box brightness in dark-mode (it's way too bright/white compared to everything else, sticking out)
- highlight (currently just making font bold, maybe there is a better approach for this) **label** when hovering over ; this helps significantly with signalling to the user that he **can click on the text**, which is much easier than trying to pin the actual **box**; I think there is a room to improve checkboxes overall (redesign them somehow), can't suggest anything better at the moment though

Demo: 

https://user-images.githubusercontent.com/112318969/208713669-bc8f2b50-b4b0-4f44-99fb-ec64cdc2ae76.mp4

### DIFF
--- a/client/webserver/site/src/css/main.scss
+++ b/client/webserver/site/src/css/main.scss
@@ -507,3 +507,17 @@ span.token-aware-symbol sup {
     }
   }
 }
+
+.form-check-input[type=checkbox] {
+  background-color: #ebebeb;
+
+  &:checked {
+    background-color: #3095db;
+  }
+}
+
+.form-check-label {
+  &:hover {
+    font-weight: bold;
+  }
+}
--- a/client/webserver/site/src/css/main.scss
+++ b/client/webserver/site/src/css/main.scss
@@ -510,10 +510,7 @@ span.token-aware-symbol sup {
 
 .form-check-input[type=checkbox] {
   background-color: #ebebeb;
-
-  &:hover {
-    cursor: pointer;
-  }
+  cursor: pointer;
 
   &:checked {
     background-color: #3095db;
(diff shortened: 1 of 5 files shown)


norwnd:
[Added](https://github.com/decred/dcrdex/pull/2002/commits/81f991e7e325b93b7823d76a8e380ed55156c2b3) `cursor: pointer;` on hover, looks like it is being used already for pretty much all clickable elements

client/webserver/site/src/css/main_dark.scss
```diff
@@ -132,4 +132,18 @@ body.dark {
...
+
+    &:checked {
+      background-color: #27278d;
+    }
+  }
+
+  .form-check-label {
+    &:hover {
+      font-weight: bold;
+    }
```
buck54321, line 147: Styling from **main.scss** still applies. This isn't needed.

buck54321 (changes requested):
Looks sharp!

client/webserver/site/src/css/main_dark.scss
norwnd: Oh yeah, right, removed.

client/webserver/site/src/css/main.scss
```diff
@@ -507,3 +507,22 @@ span.token-aware-symbol sup {
     }
   }
 }
+
+.form-check-input[type=checkbox] {
+  background-color: #ebebeb;
+
+  &:hover {
+    cursor: pointer;
+  }
```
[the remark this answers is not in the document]
norwnd: Probably not, seems to be working same with your suggestion applied.

chappjc:
Could make the cut for 0.6.  I'm ambivalent about the changes personally, but it seems to be an improvement.

buck54321 (approved):

PR #2003, opened 2022-12-20
ui: order form compacting

norwnd:
Builds on https://github.com/decred/dcrdex/pull/2002.

I think before [attempting to implement quantity-slider](https://github.com/decred/dcrdex/pull/1921#discussion_r1014318998) it's worth to take a shot at cutting limit-order size overall. Limit-order form is already pretty big, adding slider on top likely won't look/feel good.

Side by side comparison (before / after),

limit, light:

imageimage

limit, dark:

imageimage

market, light:

imageimage

market, dark:

imageimage

Here is new design in action: 

https://user-images.githubusercontent.com/112318969/208711529-8981e2cd-2af1-4e18-bda9-b951e503c83a.mp4

I think I managed to keep the overall order-form style in line with what I see on other UI forms.


client/webserver/locales/en-us.go
```diff
@@ -63,14 +63,15 @@ var EnUS = map[string]string{
...
+	"buy_switch":                     "⇄ Buy",
+	"sell_switch":                    "⇄ Sell",
 	"lot_size":                       "Lot Size",
 	"Rate Step":                      "Rate Step",
 	"Max":                            "Max",
 	"lot":                            "lot",
 	"min trade is about":             "min trade is about",
 	"immediate_explanation":          "If the order doesn't fully match during the next match cycle, any unmatched quantity will not be booked or matched again. Taker-only order.",
-	"Immediate or cancel":            "Immediate or cancel",
+	"Immediate only":                 "Immediate only",
```
norwnd, line 74: This seems to be slightly more accurate, since "Immediate or cancel" feels like whole order will be canceled if something didn't match, as opposed to "only immediate part" will be filled. 

And, short text looks better in UI.

client/webserver/site/src/css/main_dark.scss
```diff
@@ -35,19 +35,19 @@ body.dark {
...
   }
 
   .brdrtop {
-    border-top: 1px solid dark_border_color;
+    border-top: 1px solid $dark_border_color;
   }
 
   .brdrbottom {
-    border-bottom: 1px solid dark_border_color;
+    border-bottom: 1px solid $dark_border_color;
```
norwnd, line 50: Not sure whether this affected anything or not, but looked like a typo that needed fixing.

client/webserver/site/src/css/market.scss
```diff
@@ -247,15 +297,6 @@ div[data-handler=markets] {
...
   }
 
-  .markettab {
-    height: 30px;
-    border-style: none;
-
-    &.selected {
-      background-color: #fffe;
-    }
-  }
```
norwnd, line 257: Was never used - cleaning up.

client/webserver/site/src/css/market_dark.scss
```diff
@@ -17,31 +17,11 @@ body.dark {
...
-      border-right: 1px solid $dark_border_color;
-    }
-
-    .brdrtop {
-      border-top: 1px solid $dark_border_color;
-    }
-
-    .brdrbottom {
-      border-bottom: 1px solid $dark_border_color;
-    }
```
norwnd, line 43: All these seems to have already been defined in `main_dark.scss`, as I understood it, after removing these - the ones in `main` will be used.

client/webserver/site/src/css/main.scss
```diff
@@ -71,8 +71,7 @@ input:invalid {
 
 header.maintop {
   width: 100%;
-  height: 70px;
-  min-height: 70px;
+  min-height: 60px;
```
norwnd, line 74: It seems that 60px is sufficient for the header section ? I don't see any obvious issues with lowering it to 60 (to save more space) with Brave/Chrome/Firefox.

I'm also not sure why both are defined, isn't it the case that `min-height` always supersedes `height` ? That's [what  I found](https://css-tricks.com/almanac/properties/m/min-height/) when googled.

It's in a separate commit, could be easily removed in case it breaks anything.

chappjc:
I appreciate the videos on your PRs! However, they won't play for me in Chromium or FF.  What's your capture app?

norwnd:
> However, they won't play for me in Chromium or FF. What's your capture app?

I'm using [Kap on Mac](https://github.com/wulkano/kap), just learning it. Firefox won't play it for me either, hmm, opens easily in standard Chrome/Brave browsers. 

Will note this for the future, but let me know if you want/need me to re-take demo-capture for this PR.

client/webserver/site/src/css/market_dark.scss
norwnd: Maybe they were set because [these didn't work](https://github.com/decred/dcrdex/pull/2003#discussion_r1053494915) properly ?

chappjc:
> > However, they won't play for me in Chromium or FF. What's your capture app?
> 
> I'm using [Kap on Mac](https://github.com/wulkano/kap), just learning it. Firefox won't play it for me either, hmm, opens easily in standard Chrome/Brave browsers.
> 
> Will note this for the future, but let me know if you want/need me to re-take demo-capture for this PR.

I downloaded it and it's H.265 (HEVC), and it plays fine in vlc, so that's probably why browsers are picky about it.  If there's h.264 or webm, that's probably the best compatibility options.

norwnd:
This PR is ready for preview, but maybe hold off proper reviews until https://github.com/decred/dcrdex/pull/2002 and https://github.com/decred/dcrdex/pull/1921 are merged (since it's easier to re-check/test visual HTML/CSS stuff to see if anything broke after rebase).

Maybe I should have put it in Draft ? I didn't because other than that ^ it is ready to go.

buck54321:
We need to be consistent with our margins and padding. Does this match other forms used. I also think it looks worse compacted. What's the problem exactly with keeping it the way it was? If you add the slider is it too tall to fit on screen or something?

norwnd:
> We need to be consistent with our margins and padding. Does this match other forms used?

visually, to me, it looks consistent enough with the rest of the forms I see in UI, or do you mean there is something else to margins/padding (like on different screen sizes it becomes more obvious, or something) ?

I guess margins / padding could be tweaked a bit to find the most optimal values, if you feel like currently they are too tight.

>  I also think it looks worse compacted. What's the problem exactly with keeping it the way it was? If you add the slider is it too tall to fit on screen or something?

I was hoping you'd like it :) Consider it "an alternative design" I'd want to propose. I think I've managed to remove some unnecessary clutter (like having 2 buy/sell indicators when 1 is enough) / spacing (around `Immediate or cancel`) without sacrificing information contents.

And done a couple of quality-of-life modifications, like:
- in light mode, on **master** it is not clear which of the tabs: Market / Limit is open (unless you look at the actual contents, the tab itself isn't highlighted - it is highlighted in dark-mode)
- trimmed unnecessary text, to me it looks cleaner when shorter
- the most important fields are highlighting in bold (quantity/price/total) and are **placed closer together so it is easier** on the eyes **to keep track how changing one value affects others**; this is the 2nd reason why compaction is desirable, the 1st being removing unnecessary bulk

client/webserver/site/src/html/markets.tmpl
```diff
@@ -215,22 +208,24 @@
...
-                      <span data-unit="base"></span>
-                    </button>
-                    <button id="sellBttn" type="button" class="sellred-bg ms-2">
-                      [[[Sell]]]
-                      <span data-unit="base"></span>
-                    </button>
-                    <span></span>
+              <form id="orderForm" class="p-1 d-hide">
+                {{- /* ORDER TYPE BUTTONS */ -}}
+                <div id="orderTypeBttns" class="fs14 d-flex justify-content-between">
```
buck54321, line 213: Nice! Moving these into the form will give us tab navigation as discussed in #2022.

client/webserver/site/src/js/markets.ts
```diff
@@ -182,6 +182,8 @@ export default class MarketsPage extends BasePage {
   recentMatchesSortDirection: 1 | -1
   stats: [StatsDisplay, StatsDisplay]
   loadingAnimations: { candles?: Wave, depth?: Wave }
+  isSell: boolean
+  isLimit: boolean
```
buck54321, line 187: Good move.

buck54321 (changes requested):
I appreciate you looking at this. There's a lot of good stuff here. Let's just iterate on the style again.
I see a lot of `mt-3` or `mt-4` going to `mt-1`. What's it look like with a more consistent `mt-2` everywhere?

buck54321:
Looking at other exchanges, it seems to be pretty standard to display the buy-sell selector similar to the way it was.

image
image
image
image

If anything, we should move the buy-sell selector to the tabs at the top and de-emphasize the order-type selectors. We don't have to be exactly like others, but the goal is to create a comfortable experience for someone migrating from a CEX.

chappjc:
> If anything, we should move the buy-sell selector to the tabs at the top and de-emphasize the order-type selectors. We don't have to be exactly like others, but we want to create a comfortable experience for users familiar with CEX interfaces.

And if we do make buy/sell higher in the selector hierarchy of this form, it needs to continue to be the most prominent to prevent accidentally doing a trade in the wrong direction.  That is, if we make a change here, we cannot just move buy/sell to where limit/market were without some appropriate color-based style to keep it distinct, as show in @buck54321's example screens in the previous comment.

norwnd:
If you feel like keeping to "standards" is important here (or better-looking), what about something like this ([added a commit to address that concern](https://github.com/decred/dcrdex/pull/2003/commits/b3abc93cc417665b052981b1e5d1ebdbd2c492da), along with minor adjustments) ? having 4 selectors on 1 line looks much cleaner to me, compared to what typical exchanges are doing:

imageimage

and [another commit](https://github.com/decred/dcrdex/pull/2003/commits/d43a186b3debc8695f21fc49c7b7cc65ba37e9af) on top of that with @buck54321's suggestions on setting margins to `mt-2` (screenshots above ^ were taken with that applied as well).

norwnd:
Fixed couple of other issues (like relying on padding for element placement) [here](https://github.com/decred/dcrdex/pull/2003/commits/6b04b3d450d4e7a1fd45806b36c696734ef52efd). Here is the updated version:

imageimage

chappjc:
This last iteration looks nice.

chappjc:
I still like this form in the latest revision, but in chat we agreed it's a bit unusual to have the Limit/Market selectors flanking the Buy/Sell.  IMO we can just put the big colored Buy/Sell buttons on the left, with the smaller and less prominent Limit/Market buttons on the right.  Really minor change.  Foresee issues with that?  I suspect we can get this merged after that.

norwnd:
> IMO we can just put the big colored Buy/Sell buttons on the left, with the smaller and less prominent Limit/Market buttons on the right. Foresee issues with that?

That approach crossed my mind, I just think it might look a bit clunky ... at least the way I imagine it would look, if @buck54321 is onboard with such an approach I can make these adjustments and we'll be able to compare before/after screenshots side by side for real.

chappjc:
> > IMO we can just put the big colored Buy/Sell buttons on the left, with the smaller and less prominent Limit/Market buttons on the right. Foresee issues with that?
> 
> That approach crossed my mind, I just think it might look a bit clunky ... at least the way I imagine it would look, if @buck54321 is onboard with such an approach I can make these adjustments and we'll be able to compare before/after screenshots side by side for real.

It was his idea to change this (matrix chat), but I'll let him propose a specific change.

buck54321:
If it fits nicely and doesn't look weird. To me, it seems much more natural to have the groups stacked vertically.them

chappjc:
Removing from 0.6 milestone since we cannot quite settle on the top selection buttons for buy/sell/market/limit.  I think the rest of the compaction changes are straightforward and nice though.

chappjc:
Closing as deadlocked.  Can reopen later.

PR #2004, opened 2022-12-20
server: allow server operators to provide custom logo

### LINKED ISSUE #1256: server: custom logo
The server should provide a custom logo. 

1. Min size 32x32. 1:1 aspect ratio.
2. Served over HTTP
3. Logo is configurable, but we need a default icon. I can see two ways of dealing with this. a) have a default icon for the default server configuration. Probably just the dcrdex logo. Or b) have no default. If none is provided, client can fall back to letter coins. e.g. **/img/coins/a.png**

ukane-philemon:
This makes it possible for server operators to specify their custom logos. Closes #1256 

-- Add new server HTTP `logo` endpoint.
-- Add server custom logo to config response.

Added logo display on the `settings` and `dexsettings` pages. Will it look better if the logos had a rounded border?
Screenshot 2022-12-20 at 5 42 35 PM
Screenshot 2022-12-20 at 5 42 00 PM

Screenshot 2022-12-20 at 5 55 30 PM
Screenshot 2022-12-20 at 5 42 17 PM

### DIFF
--- a/client/comms/tls.go
+++ b/client/comms/tls.go
@@ -1,0 +1,35 @@
+package comms
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"net/url"
+)
+
+// TLSConfig prepares a *tls.Config struct using the provided cert.
+func TLSConfig(URL string, cert []byte) (*tls.Config, error) {
+	if len(cert) == 0 {
+		return nil, nil
+	}
+
+	uri, err := url.Parse(URL)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing URL: %w", err)
+	}
+
+	rootCAs, _ := x509.SystemCertPool()
+	if rootCAs == nil {
+		rootCAs = x509.NewCertPool()
+	}
+
+	if ok := rootCAs.AppendCertsFromPEM(cert); !ok {
+		return nil, ErrInvalidCert
+	}
+
+	return &tls.Config{
+		RootCAs:    rootCAs,
+		MinVersion: tls.VersionTLS12,
+		ServerName: uri.Hostname(),
+	}, nil
+}
--- a/client/comms/wsconn.go
+++ b/client/comms/wsconn.go
@@ -12,7 +12,6 @@ import (
 	"fmt"
 	"net"
 	"net/http"
-	"net/url"
 	"regexp"
 	"strings"
 	"sync"
@@ -165,28 +164,9 @@ func NewWsConn(cfg *WsCfg) (WsConn, error) {
 		return nil, fmt.Errorf("ping wait cannot be negative")
 	}
 
-	var tlsConfig *tls.Config
-	if len(cfg.Cert) > 0 {
-
-		uri, err := url.Parse(cfg.URL)
-		if err != nil {
-			return nil, fmt.Errorf("error parsing URL: %w", err)
-		}
-
-		rootCAs, _ := x509.SystemCertPool()
-		if rootCAs == nil {
-			rootCAs = x509.NewCertPool()
-		}
-
-		if ok := rootCAs.AppendCertsFromPEM(cfg.Cert); !ok {
-			return nil, ErrInvalidCert
-		}
-
-		tlsConfig = &tls.Config{
-			RootCAs:    rootCAs,
-			MinVersion: tls.VersionTLS12,
-			ServerName: uri.Hostname(),
-		}
+	tlsConfig, err := TLSConfig(cfg.URL, cfg.Cert)
+	if err != nil {
+		return nil, fmt.Errorf("TLSConfig error: %w", err)
 	}
 
 	return &wsConn{
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -12,6 +12,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"io"
 	"math"
 	"net"
 	"net/http"
@@ -10122,11 +10123,11 @@ func (c *Core) fetchDEXLogo(host string) []byte {
 		return nil
 	}
 
-	buf := new(bytes.Buffer)
-	if _, err = buf.ReadFrom(res.Body); err != nil {
+	logoBytes, err := io.ReadAll(io.LimitReader(res.Body, 1<<20))
+	if err != nil {
 		c.log.Errorf("%s: error reading response body: %v", funcName, err)
 		return nil
 	}
 
-	return buf.Bytes()
+	return logoBytes
 }
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -14,6 +14,7 @@ import (
 	"fmt"
 	"math"
 	"net"
+	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
@@ -1404,6 +1405,11 @@ type Config struct {
 	UnlockCoinsOnLogin bool
 }
 
+type cachedLogo struct {
+	lastRefreshed time.Time
+	logo          []byte
+}
+
 // Core is the core client application. Core manages DEX connections, wallets,
 // database access, match negotiation and more.
 type Core struct {
@@ -1470,6 +1476,9 @@ type Core struct {
 			prices map[string]*stampedPrice
 		}
 	}
+
+	dexLogoMtx   sync.Mutex
+	dexLogoCache map[string]*cachedLogo
 }
 
 // New is the constructor for a new Core.
@@ -1568,6 +1577,8 @@ func New(cfg *Config) (*Core, error) {
 
 		fiatRateSources: make(map[string]*commonRateSource),
 		pendingWallets:  make(map[uint32]bool),
+
+		dexLogoCache: make(map[string]*cachedLogo),
 	}
 	c.mm.bots = make(map[uint64]*makerBot)
 	c.mm.cache.prices = make(map[string]*stampedPrice)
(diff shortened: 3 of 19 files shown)


dex/msgjson/types.go
```diff
@@ -1270,6 +1272,9 @@ type BondAsset struct {
 
 // ConfigResult is the successful result for the ConfigRoute.
 type ConfigResult struct {
+	// Logo is an optional custom logo that can be be provided by server
+	// operators.
+	Logo []byte `json:"logo"`
```
chappjc, line 1277: As I said in https://github.com/decred/dcrdex/issues/1256#issuecomment-1358479967, adding the logo bytes to the `ConfigResult` is a **no-go** for me.

The HTTP endpoint (`/logo`) that you added will serve this purpose.

I don't even think the dexc backend needs to concern itself with the logo.  The frontend not being able to request the logo because it's missing the cert is solvable by putting the cert in `core.Exchanges`.  However, if you really want the backend to be retrieving the logo, not the frontend (the only consumer of the logo), then core should fetch it via HTTP regardless of whether it's a configured exchange or not.  @buck54321 noted it must be HTTP in the issue.

client/webserver/template.go
```diff
@@ -189,14 +194,81 @@ var templateFuncs = template.FuncMap{
...
+	u.Host = host
+	u.Scheme = scheme
+	u.Path = serverLogoPath
+
+	req, err := http.NewRequest(http.MethodGet, u.String(), nil)
+	if err != nil {
+		return nil, fmt.Errorf("http.NewRequest error: %w", err)
+	}
+
+	tlsConfig, err := comms.TLSConfig(u.String(), cert)
```
martonp, line 247: Shouldn't the tlsConfig only be used if it is https?

client/comms/tls.go
```diff
@@ -0,0 +1,35 @@
...
+	"crypto/x509"
+	"fmt"
+	"net/url"
+)
+
+// TLSConfig prepares a *tls.Config struct using the provided cert.
+func TLSConfig(URL string, cert []byte) (*tls.Config, error) {
+	if len(cert) == 0 {
+		return nil, nil
+	}
```
ukane-philemon, line 14: https://github.com/decred/dcrdex/pull/2004/files#r1054897301

@martonp, you are correct and it is already handled.

docs/wiki/Server-Admin.md
```diff
@@ -2,6 +2,15 @@
 
 ## Exchange Settings
 
+### Logo Image
+
+Server operators can optionally provide their preferred logo image when starting
+`dcrdex` using the `--logo` flag, e.g `dcrdex --logo path/to/logo.png`. Logo
+image should be preferably encoded as a `png`, `jpeg`, or `jpeg` file.
```
buck54321, line 9: Thanks for updating docs. We should word the instructions towards a configuration file, not CLI flags.

server/comms/server.go
```diff
@@ -448,6 +449,7 @@ func (s *Server) Run(ctx context.Context) {
 		rr.Use(s.limitRate)
 		rr.Get("/config", routeHandler(msgjson.ConfigRoute))
 		rr.Get("/spots", routeHandler(msgjson.SpotsRoute))
+		rr.Get("/logo", routeHandler(msgjson.LogoRoute))
```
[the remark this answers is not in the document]
chappjc: It does not.  It does `w.Header().Set("Content-Type", "application/json; charset=utf-8")` and `json.Marshal`, sending a base64 *text* encoding of the binary data.

client/core/types.go
```diff
@@ -594,6 +594,7 @@ type PendingFeeState struct {
 // Exchange represents a single DEX with any number of markets.
 type Exchange struct {
 	Host             string                 `json:"host"`
+	Cert             []byte                 `json:"cert"`
```
chappjc, line 597: Just noting one more time that this means the frontend has the TLS certificate and could make the XHR request itself.  That said, I'm also fine with the Go html template execution doing `fetchExchangeLogo` as you have it.  It's probably more proper to keep all server communications to the dexc backend.

server/comms/server.go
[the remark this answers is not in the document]
ukane-philemon: > It does not. It does `w.Header().Set("Content-Type", "application/json; charset=utf-8")` and `json.Marshal`, sending a base64 _text_ encoding of the binary data.

Oh, right.

ukane-philemon:
Something weird went off [here](https://github.com/decred/dcrdex/actions/runs/3824469529/jobs/6506641057).

chappjc:
> Something weird went off [here](https://github.com/decred/dcrdex/actions/runs/3824469529/jobs/6506641057).

Dang that was supposed to have been resolved by https://github.com/decred/dcrdex/pull/1995

chappjc:
Resolving that unrelated Swapper deadlock in https://github.com/decred/dcrdex/pull/2021
Thanks for bringing that to my attention @ukane-philemon

client/core/types.go
chappjc: Hmm, one issue with this is if the DEX is down, loading the html page hangs (times out), whereas it could be async on the frontend.

Another issue I'm seeing is that there's no caching of the logo anywhere, so whenever the user loads one of the three pages that would show the logo, it makes a server request.  That's a bit much.

ukane-philemon: > Hmm, one issue with this is if the DEX is down, loading the html page hangs (times out), whereas it could be async on the frontend.

I would set a timeout for the request.

chappjc: > I would set a timeout for the request.

Doesn't matter.  This makes a fast *local* page load have the latency of a remote server.

Also, it's not necessary to involve the config response to achieve caching of the logo.

Another issue not yet mentioned is the network route used to retrieve the logo.  Say the user has a proxy (tor or otherwise) configured for their authenticated server comms, it would be a major information leak to fetch the logo on the open internet.

ukane-philemon: > Another issue not yet mentioned is the network route used to retrieve the logo. Say the user has a proxy (tor or otherwise) configured for their authenticated server comms, it would be a major information leak to fetch the logo on the open internet.

The webserver could have this set when it is started so we can use it when need(which would be just for this).

ukane-philemon: @chappjc, I have re-implemented my suggestion in https://github.com/decred/dcrdex/pull/2004/commits/7838b89117045ce9eaeb7a9503a019e933ac25e3 and I believe it handles major concerns (caching and proxy) adequately. Please take a look and let me know if I missed anything.

client/comms/wsconn.go
```diff
@@ -150,28 +149,9 @@ func NewWsConn(cfg *WsCfg) (WsConn, error) {
...
-		}
-
-		tlsConfig = &tls.Config{
-			RootCAs:    rootCAs,
-			MinVersion: tls.VersionTLS12,
-			ServerName: uri.Hostname(),
-		}
+	tlsConfig, err := TLSConfig(cfg.URL, cfg.Cert)
+	if err != nil {
+		return nil, fmt.Errorf("TlsConfig error: %w", err)
```
buck54321, line 154: `TlsConfig` capitalization does not match function name.

client/core/core.go
```diff
@@ -9510,3 +9513,90 @@ func (c *Core) saveDisabledRateSources() {
...
+	} else {
+		cert = CertStore[c.Network()][host]
+	}
+
+	logoBytes, err := c.fetchDEXLogo(host, cert)
+	if err != nil {
+		c.log.Error("fetchDEXLogo error: %v", err)
+		return nil
+	}
+	return logoBytes
```
buck54321, line 9534: The http cache would save us from hitting `fetchDEXLogo` too many times if the user is only using one device, and as long as the user's client is sending the headers. If they sign in with multiple devices, or have the cache disabled,   or are using **Core** outside of **webserver**, we would generate repeated requests. We should cache the logo bytes locally.

server/cmd/dcrdex/config.go
```diff
@@ -414,6 +417,23 @@ func loadConfig() (*dexConf, *procOpts, error) {
 		cfg.LogDir = filepath.Join(cfg.AppDataDir, cfg.LogDir)
 	}
 
+	if cfg.Logo == "" {
+		cfg.Logo = filepath.Join(cfg.AppDataDir, defaultLogoName)
+	} else if !filepath.IsAbs(cfg.Logo) {
+		cfg.Logo = filepath.Join(cfg.AppDataDir, cfg.Logo)
```
buck54321, line 423: Docs say 
```
logo=path/to/logo.png
```
 and don't mention that it will be interpreted as a subpath of `AppDataDir`.

server/comms/server.go
```diff
@@ -448,6 +449,25 @@ func (s *Server) Run(ctx context.Context) {
...
+			if handler == nil {
+				writeServerError(w, map[string]string{"error": "no handler"})
+				return
+			}
+
+			logo, err := handler(r.Context().Value(ctxThing))
+			if err != nil { // err will always be nil, but this might change.
+				writeServerError(w, map[string]string{"error": err.Error()})
+				return
+			}
```
buck54321, line 463: I do think that we should inform the client about the **availability** of a logo in the `msgjson.ConfigResult`, so they can avoid the request altogether if we don't have a file.

buck54321 (changes requested):

server/comms/server.go
chappjc: The logo business will begin life deprecated, because it'll have no place in a mesh.  That's why I just don't think we should bother the important message payloads.  That said, I don't really mind what happens as long as we don't add kilobytes to the config response for this.  An extra http get with a 404 is fine imo, when there is a server that doesn't have a logo, when compared against technical baggage.  If that slows down page rendering, that says this was not engineered correctly anyway because this feature should add no latency in any place but the loading of the logo itself.

client/core/core.go
ukane-philemon: I've added a DEX logo cache in `Core` and also a means to clean up the cache every 24 hours (https://github.com/decred/dcrdex/commit/a3ebb22156e7148b1ad36c401aeeee9b5f392697)

server/comms/server.go
```diff
@@ -448,6 +449,25 @@ func (s *Server) Run(ctx context.Context) {
...
+
+			logoBytes := logo.([]byte)
+			w.WriteHeader(http.StatusOK)
+			w.Header().Set("Content-Type", http.DetectContentType(logoBytes))
+			w.Header().Set("Content-Length", fmt.Sprintf("%d", len(logoBytes)))
+			w.Write(logoBytes)
```
buck54321, line 469: Unless I'm reading things wrong, it's possible for logoBytes to be empty. I think we should 404 in that case.

client/webserver/site/src/html/settings.tmpl
```diff
@@ -35,7 +35,16 @@
       <div id="exchanges" {{if eq (len .UserInfo.Exchanges) 0}} class="d-hide"{{end}}>
         <h5>[[[registered dexes]]]</h5>
         {{range $host, $xc := .UserInfo.Exchanges}}
-          <a href="/dexsettings/{{$host}}"><button class="bg2 selected"><div class=text-break>{{$host}}<span class="dex-settings-icon ico-settings"></span></div></button></a>
+          <a href="/dexsettings/{{$host}}">
+            <button class="bg2 selected">
+              <div class=text-break>
+                <img class="mini-icon m-1" src="/dexlogo/{{$host}}">
```
buck54321, line 41: Logo is a little big relative to the text size. 
image
Could increase text size or decrease logo size or both.

buck54321 (changes requested):

server/comms/server.go
chappjc, line 469: - WriteHeader cannot come before w.Header().Set() https://cs.opensource.google/go/go/+/refs/tags/go1.19.5:src/net/http/server.go;l=100-102
- WriteHeader isn't even needed because Write will do it for you https://cs.opensource.google/go/go/+/refs/tags/go1.19.5:src/net/http/server.go;l=120-121

ukane-philemon: okay.

client/webserver/site/src/html/settings.tmpl
ukane-philemon: I intentionally made the logo bigger so it's more visible but seems it should match the text size.

client/core/core.go
```diff
@@ -9510,3 +9519,127 @@ func (c *Core) saveDisabledRateSources() {
...
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
+		c.log.Errorf("%s: received unexpected response: %s", funcName, res.Status)
+		return nil
+	}
+
+	var logoBytes []byte
+	if err = json.NewDecoder(res.Body).Decode(&logoBytes); err != nil {
```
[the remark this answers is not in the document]
ukane-philemon: > JSON? This seems like it shouldn't work. Go json encodes bytes as a base64 string.

It worked in my initial testing, but am now using a `bytes.Buffer` in my recent commit.

server/comms/server.go
```diff
@@ -450,20 +450,30 @@ func (s *Server) Run(ctx context.Context) {
...
+			writeErr := func(code int, msg string) {
+				w.WriteHeader(code)
+				w.Write([]byte(msg))
+			}
+
 			handler := httpRoutes[msgjson.LogoRoute]
 			if handler == nil {
-				writeServerError(w, map[string]string{"error": "no handler"})
+				log.Critical("No handler for %s", msgjson.LogoRoute)
+				writeErr(500, "Something unexpected happened")
```
buck54321, line 461: `http.StatusInternalServerError`

Could also use `http.StatusText(http.StatusInternalServerError)` for the text.

server/comms/server.go
```diff
@@ -450,20 +450,30 @@ func (s *Server) Run(ctx context.Context) {
...
 			if err != nil { // err will always be nil, but this might change.
-				writeServerError(w, map[string]string{"error": err.Error()})
+				writeErr(404, err.Error())
+				return
+			}
+
+			if logo == nil {
+				writeErr(404, "no server logo")
 				return
 			}
```
buck54321, line 474: `http.StatusNotFound` or even `http.NotFound(w, r)`.

ukane-philemon:
Had to squash the multiple commits.

buck54321 (approved):

ukane-philemon:
Just rebased...

chappjc:
Will review.  Please rebase again when you get a chance.

client/core/core.go
```diff
@@ -10013,3 +10024,109 @@ func (c *Core) saveDisabledRateSources() {
...
+		return nil
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
+		c.log.Errorf("%s: received unexpected response: %s", funcName, res.Status)
+		return nil
+	}
+
+	buf := new(bytes.Buffer)
```
chappjc, line 10125: Since we're reading into memory, we should use a `io.LimitReader` to be safe.

ukane-philemon: Done.

chappjc:
This PR needs another advocate.  There is no point to this in the future and it adds a lot of baggage.  I do not want this.

chappjc:
Closing tomorrow.

buck54321:
It just doesn't make sense to do this anymore now that we're focusing on mesh. Thanks for the effort @ukane-philemon.

PR #2008, opened 2022-12-26
client/asset/{btc,dcr}: maintain sync progress after losing peers

### LINKED ISSUE #2006: Decred SPV 99.9%(or 100%) Sync progress on connectivity loss
During initial wallet sync (I was setting up brand-new `dexc` instance on latest **master** on **testnet**) when Decred SPV (Native) wallet looses connectivity "Sync progress" jumps to 99.9% on Wallet view (or 100% on Registration view). When connectivity is restored, it changes back to the value it should have been showing all along (assuming we can keep showing last reasonable known value, or maybe something like "?" / "Disconnected" in its place).
I also saw similar thing on **mainnet**.

I think this is the relevant piece in the logs:
```
2022-12-22 08:02:46.138 [WRN] CORE: Wallet for asset dcr has zero network peers!
2022-12-22 08:02:46.189 [WRN] CORE: notify: |WARNING| (walletconfig) Wallet network issue - Decred wallet has no network peers!
2022/12/22 11:03:00 "GET http://127.0.0.1:5758/wallets HTTP/1.1" from 127.0.0.1:63150 - 200 38282B in 68.283776ms
2022/12/22 11:03:00 "POST http://127.0.0.1:5758/api/orders HTTP/1.1" from 127.0.0.1:63157 - 200 24B in 76.255337ms
2022/12/22 11:03:01 "GET http://127.0.0.1:5758/img/coins/doge.png HTTP/1.1" from 127.0.0.1:63150 - 200 3075B in 213.372409ms
2022/12/22 11:03:01 "GET http://127.0.0.1:5758/img/coins/zec.png HTTP/1.1" from 127.0.0.1:63156 - 200 4061B in 244.051191ms
2022-12-22 08:03:21.139 [TRC] CORE: New peer count for asset dcr: 4
2022-12-22 08:03:21.198 [INF] CORE: notify: |SUCCESS| (walletconfig) Wallet connectivity restored - Decred wallet has reestablished connectivity.
2022-12-22 08:03:26.138 [TRC] CORE: New peer count for asset dcr: 7
2022-12-22 08:03:43.045 [TRC] WEB[WS][127.0.0.1:61904]: Sent 623 and dropped 0 messages to 127.0.0.1:61904 before shutdown.
2022-12-22 08:03:43.046 [TRC] WEB[WS]: Disconnected websocket client 127.0.0.1:61904
```

**Besides the fact that it will try to re-connect to Decred wallet**,  restarting DEX doesn't seem to affect/change this behavior, it stays reproducible between restarts.

How it looks in UI (2 min presentation):

[Kapture 2022-12-22 at 11.30.48.webm](https://user-images.githubusercontent.com/112318969/209094130-f39ee953-aa4d-40b1-bd24-e3162384849c.webm)

itswisdomagain:
Resolves #2006.

Sync target height becomes 0 (for btc) or the same as the wallet tip (for dcr) when the correct value cannot be inferred from connected peers when the connection to all peers is lost.

Fix: cache sync target height so the cached value is used when calculating sync progress even when there are no peers connected.

### DIFF
--- a/client/asset/btc/spv_wrapper.go
+++ b/client/asset/btc/spv_wrapper.go
@@ -536,8 +536,15 @@ func (w *spvWallet) syncStatus() (*syncStatus, error) {
 	if err != nil {
 		return nil, err
 	}
-	target := w.syncHeight()
+
 	currentHeight := chainBlk.Height
+
+	var target int32
+	if len(w.cl.Peers()) > 0 {
+		target = w.syncHeight()
+	} else { // use cached value if available
+		target = atomic.LoadInt32(&w.syncTarget)
+	}
 
 	var synced bool
 	var blk *block
--- a/client/asset/dcr/spv.go
+++ b/client/asset/dcr/spv.go
@@ -118,15 +118,16 @@ func (w *extendedWallet) TxDetails(ctx context.Context, txHash *chainhash.Hash) 
 // spvWallet is a Wallet built on dcrwallet's *wallet.Wallet running in SPV
 // mode.
 type spvWallet struct {
-	dcrWallet   // *wallet.Wallet
-	db          wallet.DB
-	acctNum     uint32
-	acctName    string
-	dir         string
-	chainParams *chaincfg.Params
-	log         dex.Logger
-	spv         spvSyncer // *spv.Syncer
-	tipChan     chan *block
+	dcrWallet         // *wallet.Wallet
+	db                wallet.DB
+	acctNum           uint32
+	acctName          string
+	dir               string
+	chainParams       *chaincfg.Params
+	log               dex.Logger
+	spv               spvSyncer // *spv.Syncer
+	bestSpvPeerHeight int32     // atomic
+	tipChan           chan *block
 
 	blockCache blockCache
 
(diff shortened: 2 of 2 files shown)


client/asset/btc/spv_wrapper.go
```diff
@@ -506,7 +506,7 @@ func (w *spvWallet) removePeer(addr string) error {
 
 // syncHeight is the best known sync height among peers.
 func (w *spvWallet) syncHeight() int32 {
-	var maxHeight int32
+	maxHeight := atomic.LoadInt32(&w.syncTarget)
```
chappjc, line 509: OK, but it stands out that `syncStatus()` is the only caller of `syncHeight()` and that the only update to this via `atomic.StoreInt32` is in `syncStatus()`.  Sorta suggest that one or the other method should be concerned with the atomic field, but not both.  Otherwise the fix makes sense.

chappjc (approved):

client/asset/dcr/spv.go
```diff
@@ -765,33 +766,39 @@ func (w *spvWallet) UnlockAccount(ctx context.Context, pw []byte, _ string) erro
...
 		progress = 0.999
 	}
 
 	return synced, progress, nil
 }
 
-// bestPeerInitialHeight is the highest InitialHeight recorded from our peers.
+// bestPeerInitialHeight is the highest InitialHeight recorded from connected
+// spv peers. This value is maintained even if the peer that reported the value
+// later becomes disconnected.
```
chappjc, line 793: I wonder if this is an attack vector.  Is the spv syncer smart enough not to include peers with bogus chain heights?

itswisdomagain: Probably not something we should rely on. Another thought I had, which now sounds more reasonable, is to only use the cached value if there are no more connected peers; otherwise the value reported by connected peers takes precedence. Will update.

JoeGruffins (approved):

chappjc (approved):

martonp (approved):

Issue #2009, opened 2022-12-27
client does not recognize manually broadcast refund


norwnd:
This is a rare scenario, but I think might be worth looking into in case there is more to it.

I was testing manual refunds [as per these discussions](https://github.com/decred/dcrdex/issues/952), on `testnet` I've done this:
- created DCR-USDC order
- took/matched against it with another **dexc** instance (**Taker** on screenshot below), here is [the relevant DCR swap contract](https://testnet.dcrdata.org/tx/f403a9b58ad27a1d00df7c801263548a63f2842f0349898c1af186a69564678f/out/0)
- shut down my **dexc** instances to trigger refunding scenarios
- manually broadcasted **Taker** [refund transaction](https://testnet.dcrdata.org/tx/a0d0adf9f7132672067b7497cc49df377e9801de38b8fb9c0de3788f994baeee/out/0) I pulled out from logs, it got mined after lock-time expired
- restarted my **Taker dexc** instance, and it discovered [this non-existent transaction](https://testnet.dcrdata.org/tx/4639d4dc9bef6977d59c20681f10dcec5f754387a47bb875bafc6ac9a73e5a76) as my refund, see **Refund (dcr, you)** below

<img width="400" alt="image" src="https://user-images.githubusercontent.com/112318969/209617022-411e7376-edf1-49c7-9773-540057b4ec98.png">

norwnd:
In the logs I don't see any mentions of [non-existent transaction](https://testnet.dcrdata.org/tx/4639d4dc9bef6977d59c20681f10dcec5f754387a47bb875bafc6ac9a73e5a76), or the correct [manually broadcasted transaction](https://testnet.dcrdata.org/tx/4639d4dc9bef6977d59c20681f10dcec5f754387a47bb875bafc6ac9a73e5a76). The most relevant bits about refunding this match seem to be:
```
2022-12-27 05:13:38.321 [DBG] CORE: Refundable match 8038a42b394ff7250fce0317ce421ac9701b6bb569b0e94ebc1c5b593cf806f9 for order 5bba41c5e2f61fc6793193eace7bc56308c6a3eddbdf693a48d63f33a53c8d84 (Taker)
[...]

PR #2011, opened 2022-12-28
client/asset/dcr: refund method recognizes spends in refunds

chappjc:
Resolves https://github.com/decred/dcrdex/issues/2009

The first commit fixes a bug whereby "unknown" spend status causes the DCR backend to attempt to refund even if it was already spent in certain undetectable situations.  Normally not an issue since for taker the maker redemption search would have been running prior to locktime expiry that triggers the refund attempt.

The second commit adds the ability to recognize when a located spending transaction is a refund, in which case that transaction is returned without authoring a new one.

In draft because it may be necessary to skip the broadcast step if the refund transaction was locate in this manner.

### DIFF
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -3069,14 +3069,40 @@ func (dcr *ExchangeWallet) refundTx(coinID, contract dex.Bytes, val uint64, refu
 		if utxo == nil {
 			return nil, asset.CoinNotFoundError
 		}
-		if spent == 1 {
+		val = uint64(utxo.Value)
+
+		switch spent {
+		case 0: // unspent, proceed to create refund tx
+		case 1: // spent!
 			// Refund MUST signal to caller that it is spent via
 			// asset.CoinNotFoundError so that it knows to begin looking for the
-			// counterparty's redeem and move on to redeem too.
+			// counterparty's redeem and move on to redeem too. TODO: attempt to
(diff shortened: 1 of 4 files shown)


chappjc:
Scoped for 0.6 since the first commit is a somewhat important fix to dcr's Refund method.

@norwnd This is in draft because I haven't tested it on simnet, but please give it a shot.

norwnd:
> @norwnd This is in draft because I haven't tested it on simnet, but please give it a shot.

Seems to be working with this PR, in logs (for manually refunded txn, while **dexc** was offline, then **dexc** successfully found txn I broadcasted through `dcrdata.org`):
```
2022-12-31 14:27:45.645 [DBG] CORE[dcr]: Found spender for output 8e1a680a9ca7754851aed580e0f100091aca30f592943551e7f4893d2deb3e29:0 in block 1055418 (00000000ac3a2290e02f685c929d71b55fcffac67babf2befde6ba6004a656bd), spender tx hash cf41a44c2049ec443c6e1c6e54751fa809a8e813b2bf308bd8f1728777423bcd.
```

image

chappjc:
Thanks for testing @norwnd, and raising the issue.

Next steps could do similarly for BTC with a `dexbtc.IsRefundScript` for use in the case where `Refund` discovers the contract is spent.

JoeGruffins (approved):
Code looks good and tested on simnet refunding and finding a refunded contract.

Issue #2010, opened 2022-12-27
eth and erc20: enable mainnet


chappjc:
One of the last main tasks to release ETH and ERC20 wallet support is to switch it on for mainnet and direct the v0 assets to the deployed swap contracts.  They are here:

ETHSwap: https://etherscan.io/address/0x8c17e4968b6903e1601be82ca989c5b5e2c7b400#code
ERC20Swap: https://etherscan.io/address/0x1bbd020ddd6dc01f974aa74d2d727b2a6782f32d#code

Assigning myself.

For 0.6 (see the current milestone and scope [here](https://github.com/decred/dcrdex/milestone/22)), the eth and token wallets will be fully functional with RPC providers, as will swap functionality but in a beta/experimental status initially.

PR #2013, opened 2022-12-30
enable eth and usdc.eth on mainnet

chappjc:
<s>Rebased on https://github.com/decred/dcrdex/pull/2111</s>

Resolves https://github.com/decred/dcrdex/issues/2010

ETHSwap: https://etherscan.io/address/0x8c17e4968b6903e1601be82ca989c5b5e2c7b400#code
ERC20Swap: https://etherscan.io/address/0x1bbd020ddd6dc01f974aa74d2d727b2a6782f32d#code

All code verified by etherscan

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -455,11 +455,11 @@ type baseWallet struct {
 // assetWallet satisfies the dex.Wallet interface.
 type assetWallet struct {
 	*baseWallet
-	assetID    uint32
-	tipChange  func(error)
-	log        dex.Logger
-	atomicUnit string
-	connected  atomic.Bool
+	assetID   uint32
+	tipChange func(error)
+	log       dex.Logger
+	ui        dex.UnitInfo
+	connected atomic.Bool
 
 	lockedFunds struct {
 		mtx                sync.RWMutex
@@ -684,7 +684,7 @@ func NewWallet(assetCFG *asset.WalletConfig, logger dex.Logger, net dex.Network)
 		contractors:        make(map[uint32]contractor),
 		evmify:             dexeth.GweiToWei,
 		atomize:            dexeth.WeiToGwei,
-		atomicUnit:         dexeth.UnitInfo.AtomicUnit,
+		ui:                 dexeth.UnitInfo,
 		maxSwapsInTx:       perTxGasLimit / maxSwapGas,
 		maxRedeemsInTx:     perTxGasLimit / maxRedeemGas,
 	}
@@ -988,7 +988,7 @@ func (w *ETHWallet) OpenTokenWallet(tokenCfg *asset.TokenConfig) (asset.Wallet, 
 		contractors:        make(map[uint32]contractor),
 		evmify:             token.AtomicToEVM,
 		atomize:            token.EVMToAtomic,
-		atomicUnit:         token.UnitInfo.AtomicUnit,
+		ui:                 token.UnitInfo,
 		maxSwapsInTx:       perTxGasLimit / maxSwapGas,
 		maxRedeemsInTx:     perTxGasLimit / maxRedeemGas,
 	}
@@ -1017,6 +1017,10 @@ func (eth *baseWallet) OwnsDepositAddress(address string) (bool, error) {
 	}
 	addr := common.HexToAddress(address)
 	return addr == eth.addr, nil
+}
+
+func (w *assetWallet) amtString(amt uint64) string {
+	return fmt.Sprintf("%s %s", w.ui.ConventionalString(amt), w.ui.Conventional.Unit)
 }
 
 // fundReserveType represents the various uses for which funds need to be locked:
@@ -1066,7 +1070,7 @@ func (w *assetWallet) lockFunds(amt uint64, t fundReserveType) error {
 
 	if balance.Available < amt {
 		return fmt.Errorf("attempting to lock more %s for %s than is currently available. %d > %d %s",
-			dex.BipIDSymbol(w.assetID), t, amt, balance.Available, w.atomicUnit)
+			dex.BipIDSymbol(w.assetID), t, amt, balance.Available, w.ui.AtomicUnit)
 	}
 
 	w.lockedFunds.mtx.Lock()
@@ -1390,6 +1394,8 @@ func (w *ETHWallet) FundOrder(ord *asset.Order) (asset.Coins, []dex.Bytes, error
 	// some work for the caller as well. We can't just always do it that way and
 	// remove RedeemN, since we can't guarantee that the redemption asset is in
 	// our fee-family. though it could still be an AccountRedeemer.
+	w.log.Debugf("Locking %s to swap %s in up to %d swaps at a fee rate of %d gwei/gas using up to %d gas per swap",
+		w.amtString(ethToLock), w.amtString(ord.Value), ord.MaxSwapCount, ord.MaxFeeRate, g.Swap)
 
 	coin := w.createFundingCoin(ethToLock)
 
@@ -1431,6 +1437,8 @@ func (w *TokenWallet) FundOrder(ord *asset.Order) (asset.Coins, []dex.Bytes, err
 		}
 	}()
 
+	w.log.Debugf("Locking %s to swap %s in up to %d swaps at a fee rate of %d gwei/gas using up to %d gas per swap",
+		w.parent.amtString(ethToLock), w.amtString(ord.Value), ord.MaxSwapCount, ord.MaxFeeRate, g.Swap)
 	if err := w.parent.lockFunds(ethToLock, initiationReserve); err != nil {
 		return nil, nil, err
 	}
@@ -1557,7 +1565,7 @@ func (w *assetWallet) swapGas(n int, ver uint32) (oneSwap, nSwap uint64, approve
 		}
 	}
 	if gasEst > nSwap {
-		w.log.Warnf("Swap gas estimate %d is greater than the server's configured value %d. Using live estimate + 10%.", gasEst, nSwap)
+		w.log.Warnf("Swap gas estimate %d is greater than the server's configured value %d. Using live estimate + 10%%.", gasEst, nSwap)
 		nSwap = gasEst * 11 / 10 // 10% buffer
 		if n == 1 && nSwap > oneSwap {
 			oneSwap = nSwap
@@ -2747,7 +2755,7 @@ func (w *TokenWallet) canSend(value uint64, isPreEstimate bool) (uint64, *big.In
 		}
 		avail := bal.Available
 		if avail < value {
-			return 0, nil, fmt.Errorf("not enough tokens: have %[1]d %[3]s need %[2]d %[3]s", avail, value, w.atomicUnit)
+			return 0, nil, fmt.Errorf("not enough tokens: have %[1]d %[3]s need %[2]d %[3]s", avail, value, w.ui.AtomicUnit)
 		}
 
 		ethBal, err := w.parent.Balance()
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -689,6 +689,9 @@ func NewWallet(assetCFG *asset.WalletConfig, logger dex.Logger, net dex.Network)
 		maxRedeemsInTx:     perTxGasLimit / maxRedeemGas,
 	}
 
+	logger.Infof("ETH wallet will support a maximum of %d swaps and %d redeems per transaction.",
+		aw.maxSwapsInTx, aw.maxRedeemsInTx)
+
 	aw.wallets = map[uint32]*assetWallet{
 		BipID: aw,
 	}
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -3420,7 +3420,7 @@ func (w *TokenWallet) ConfirmRedemption(coinID dex.Bytes, redemption *asset.Rede
 }
 
 const (
-	txConfsNeededToConfirm               = 10
+	txConfsNeededToConfirm               = 3
 	blocksToWaitBeforeCoinNotFound       = 10
 	blocksToWaitBeforeCheckingIfReplaced = 10
 )
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -2150,7 +2150,7 @@ func (w *TokenWallet) maybeApproveTokenSwapContract(ver uint32, maxFeeRate, swap
 		return fmt.Errorf("parent balance %d doesn't cover contract approval (%d) and tx fees (%d)",
 			ethBal.Available, approveGas*maxFeeRate, swapReserves)
 	}
-	tx, err := w.approveToken(unlimitedAllowance, maxFeeRate, ver)
+	tx, err := w.approveToken(unlimitedAllowance, maxFeeRate, approveGas, ver)
 	if err != nil {
 		return fmt.Errorf("token contract approval error (using max fee rate %d): %w", maxFeeRate, err)
 	}
@@ -2178,10 +2178,10 @@ func (w *assetWallet) tokenAllowance() (allowance *big.Int, err error) {
 
 // approveToken approves the token swap contract to spend tokens on behalf of
 // account handled by the wallet.
-func (w *assetWallet) approveToken(amount *big.Int, maxFeeRate uint64, contractVer uint32) (tx *types.Transaction, err error) {
+func (w *assetWallet) approveToken(amount *big.Int, maxFeeRate, gasLimit uint64, contractVer uint32) (tx *types.Transaction, err error) {
 	w.nonceSendMtx.Lock()
 	defer w.nonceSendMtx.Unlock()
-	txOpts, err := w.node.txOpts(w.ctx, 0, approveGas, dexeth.GweiToWei(maxFeeRate), nil)
+	txOpts, err := w.node.txOpts(w.ctx, 0, gasLimit, dexeth.GweiToWei(maxFeeRate), nil)
 	if err != nil {
 		return nil, fmt.Errorf("addSignerToOpts error: %w", err)
 	}
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -59,7 +59,7 @@ func init() {
 	asset.Register(BipID, &Driver{})
 	// Test token
 	registerToken(simnetTokenID, "A token wallet for the DEX test token. Used for testing DEX software.", dex.Simnet)
-	registerToken(usdcTokenID, "The USDC Ethereum ERC20 token.", dex.Testnet)
+	registerToken(usdcTokenID, "The USDC Ethereum ERC20 token.", dex.Mainnet, dex.Testnet)
 }
 
 const (
--- a/client/asset/eth/eth_test.go
+++ b/client/asset/eth/eth_test.go
@@ -559,6 +559,7 @@ func newTestNode(assetID uint32) *testNode {
 		approveTx:   types.NewTransaction(4, common.Address{0x34}, big.NewInt(1e9), defaultGasFeeLimit, big.NewInt(2e9), nil),
 	}
 	if assetID != BipID {
+		ttc.tContractor.gasEstimates = &tokenGases
 		c = ttc
 	}
 
--- a/client/asset/eth/eth_test.go
+++ b/client/asset/eth/eth_test.go
@@ -3779,10 +3779,10 @@ func testConfirmRedemption(t *testing.T, assetID uint32) {
 					blockSubmitted: 9,
 				},
 			},
-			bestBlock: 13,
+			bestBlock: 11, // tx.height + txConfsNeededToConfirm - 2
 			expectedResult: &asset.ConfirmRedemptionStatus{
-				Confs:  4,
-				Req:    10,
+				Confs:  txConfsNeededToConfirm - 1,
+				Req:    txConfsNeededToConfirm,
 				CoinID: toEthTxCoinID(3, 200, redeem0Data),
 			},
 			baseFee: dexeth.GweiToWei(100),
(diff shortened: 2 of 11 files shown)


chappjc:
Need to resolve https://github.com/decred/dcrdex/issues/2014 first.

chappjc:
@itswisdomagain and I will be a bit busy with https://github.com/decred/dcrdex/pull/1986, https://github.com/decred/dcrdex/pull/2036, and https://github.com/decred/dcrdex/pull/2025 for the next several days, so if anyone wants to push this branch to their remote and work on it, that's fine with me.

chappjc:
Hit the deadlock resolved by https://github.com/decred/dcrdex/pull/2021 when testing.  Should merge that soon

dex/networks/eth/tokens.go
```diff
@@ -161,8 +161,21 @@ var Tokens = map[uint32]*Token{
 		},
 		NetTokens: map[dex.Network]*NetToken{
 			dex.Mainnet: {
-				Address:       common.Address{},
-				SwapContracts: map[uint32]*SwapContract{},
+				Address: common.HexToAddress("0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48"), // https://etherscan.io/address/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48
+				SwapContracts: map[uint32]*SwapContract{
+					0: { // https://etherscan.io/address/0xa7af47db3296206ea543a82ffbf7fc312698e6c9
+						Address: common.HexToAddress("0xA7Af47DB3296206eA543A82ffBF7Fc312698E6C9"),
+						Gas: Gases{ // Same as Goerli gases. DRAFT PR: verify!
```
buck54321, line 168: This is where `GetGasEstimates` gets real. What do you suggest we use for `maxSwaps`. cc @JoeGruffins

chappjc:
Just pushed https://github.com/decred/dcrdex/pull/2013/commits/6aed724fae653731cacd0623dea747a6f84be0c6 to address observations from final review of https://github.com/decred/dcrdex/pull/1733#pullrequestreview-1218454607
Namely, extremely tight gas limits on the swap contract txns.  Targeting actual use equal to 97% of reserves was not nearly adequate.  Some wallets will target 66.6% (150% buffer).  This commit changes ours to about 130% (we should see about actual of about 77% the limit in etherscan).

chappjc:
Also need to review and probably further pad the `v0Gasses` used for ETH swaps.  Last comment and commit pertains to ERC20.

chappjc:
Last commit https://github.com/decred/dcrdex/pull/2013/commits/0abeb4e75bca10b071088e44ca879094aab805bc is in regard to https://github.com/decred/dcrdex/pull/1733#discussion_r1047421051 and the new reality of the `asset.Wallet` interface established in https://github.com/decred/dcrdex/commit/f950443754abe7aa302404590be3476a3bfbbbf5 (reasoning also stated at the time of review in https://github.com/decred/dcrdex/pull/1866#issuecomment-1270331373)

dex/networks/eth/tokens.go
martonp: Do you mean mean `maxSwapsInTx`?

chappjc: That one gets computed automatically on wallet construction, based on the hard coded gases, right?

martonp: Yes.

martonp: Not sure what `maxSwaps` is.

buck54321: `maxSwaps int` is an argument to `GetGasEstimates` that specifies how many initiations and redemptions will be sent.

dex/testing/eth/create-node.sh
```diff
@@ -32,7 +32,7 @@ chmod +x "${NODES_ROOT}/harness-ctl/${NAME}"
 
 HTTP_OPT=""
 if [ "$NAME" = "delta" ]; then
-  HTTP_OPT="--http --http.port ${DELTA_HTTP_PORT} --ws --ws.port ${DELTA_WS_PORT} --ws.api \"db,eth,net,web3,personal,txpool,admin\""
+  HTTP_OPT="--http --http.port ${DELTA_HTTP_PORT} --ws --ws.port ${DELTA_WS_PORT} --ws.api \"eth,net,web3,personal,txpool,admin\""
```
chappjc, line 35: `db` was not a thing: `ERROR[01-26|14:00:13.226] Unavailable modules in HTTP API list     unavailable=[db] available="[admin debug web3 eth txpool personal clique net les vflux engine]"`

chappjc:
Dropped the [commit](5796ff1d354148b422a8c7d82264f027329be1ec) for block poll interval fiddling with routable hosts (likely public provider).  Leaving the compiler settable interval commit.

chappjc:
Pushing small fixes as I test.

**ETHSWap mainnet of 0.0278 ETH** for 2 DCR
- swap (134,500 gas used @ 40.27 gwei => 0.005416 ETH): https://etherscan.io/tx/0x56c5b7fc3a636b4dc18f6b7acb08972e3ddcf13eac2b1389d76c73abcf5f8bff
- redeem (60,456 gas used @ 35.58 gwei => 0.002151 ETH): https://etherscan.io/tx/0x4711a6af579044f28354ee75f477de2a9e4e3862a56ee231a5eae1a8d70d1378

Very small lot size of 2 DCR, so high rate of fee loss.

Gas caps were revised after that test.

client/asset/eth/eth.go
```diff
@@ -1392,6 +1398,8 @@ func (w *TokenWallet) FundOrder(ord *asset.Order) (asset.Coins, []dex.Bytes, err
 		}
 	}()
 
+	w.log.Debugf("Locking %s to swap %s in up to %d swaps at a fee rate of %d gwei/gas using up to %d gas per swap",
+		w.parent.amtString(ethToLock), w.amtString(ord.Value), ord.MaxSwapCount, ord.MaxFeeRate, g.Swap)
```
chappjc, line 1402: `[DBG] CORE[eth][USDC.ETH]: Locking 0.045400000 ETH to swap 46.120000 USDC in up to 1 swaps at a fee rate of 200 gwei/gas using up to 227000 gas per swap`

Super high gas price (200 gwei/gas) there, but informative.

chappjc:
Re-tested with re-deployed contracts and USDC (ERC20)

New `ETHSwapV0`: https://etherscan.io/address/0x8c17e4968b6903e1601be82ca989c5b5e2c7b400
New `ERC20SwapV0`: https://etherscan.io/address/0x1bbd020ddd6dc01f974aa74d2d727b2a6782f32d

lot sizes tested are stupidly wasteful with fees though

client/asset/eth/eth_test.go
```diff
@@ -4254,6 +4255,7 @@ func testConfirmRedemption(t *testing.T, assetID uint32) {
 	}
 
 	for _, test := range tests {
+		fmt.Println(test.name)
```
buck54321, line 4258: OK to leave, but how about some nice formatting. `fmt.Printf("###### %s ###### \n", test.name)`

dex/networks/eth/params.go
```diff
@@ -242,9 +242,6 @@ var testTokenID, _ = dex.BipSymbolID("dextt.eth")
 var usdcTokenID, _ = dex.BipSymbolID("usdc.eth")
 
 // Gases lists the expected gas required for various DEX and wallet operations.
-// ★★ IMPORTANT ★★  By policy, clients should allow servers to adjust Swap and
-// Redeem gas values in their *dex.Asset up to but not over 2x the value listed
-// in VersionedGases or Tokens.
```
buck54321, line 247: No?

dex/networks/eth/tokens.go
```diff
@@ -199,18 +215,20 @@ var Tokens = map[uint32]*Token{
...
 							//   [65500 48400 48400 48400 48400]
-							Swap:      175_000,
-							SwapAdd:   115_000,
-							Redeem:    87_000,
-							RedeemAdd: 33_000,
-							Refund:    67_000,
-							Approve:   65_000,
-							Transfer:  70_000,
+							//
+							// Then buffered by about 130%...
```
buck54321, line 219: Buffered by about **30%**

dex/networks/eth/tokens.go
```diff
@@ -199,18 +215,20 @@ var Tokens = map[uint32]*Token{
...
+							Swap:      242_000, // actual ~187,880 -- https://goerli.etherscan.io/tx/0x352baccafa96bb09d5c118f8dcce26e34267beb8bcda9c026f8d5353abea50fd, verified on mainnet at 188,013 gas
+							SwapAdd:   146_400, // actual ~112,639 (300,519 for 2) -- https://goerli.etherscan.io/tx/0x97f9a1ed69883a6e701f37883ef74d79a709e0edfc4a45987fa659700663f40e
+							Redeem:    102_700, // actual ~79,012 -- https://goerli.etherscan.io/tx/0x96f007036b01eb2e44615dc67d3e99748bc133496187348b2af26834f46bfdc8, verified on mainnet at 79,113 gas
```
buck54321, line 222: I saw redeem gas of 83,850 while testing #2050.

dex/networks/eth/tokens.go
```diff
@@ -199,18 +215,20 @@ var Tokens = map[uint32]*Token{
...
-							Swap:      175_000,
-							SwapAdd:   115_000,
-							Redeem:    87_000,
-							RedeemAdd: 33_000,
-							Refund:    67_000,
-							Approve:   65_000,
-							Transfer:  70_000,
+							//
+							// Then buffered by about 130%...
+							Swap:      242_000, // actual ~187,880 -- https://goerli.etherscan.io/tx/0x352baccafa96bb09d5c118f8dcce26e34267beb8bcda9c026f8d5353abea50fd, verified on mainnet at 188,013 gas
```
buck54321, line 220: I wonder why this is higher than either I or @JoeGruffins saw before. We both got 170k.

buck54321: Wait. Are you just using the mainnet values on testnet now?

chappjc: No, I did it on both networks.  The comment just means I verified it's the same(ish) on mainnet but without pasting my txns

dex/networks/eth/params.go
chappjc: No, the `dex.Asset` doesn't even make it into the wallet backends anymore:

https://github.com/decred/dcrdex/commit/f950443754abe7aa302404590be3476a3bfbbbf5
https://github.com/decred/dcrdex/pull/1866#issuecomment-1270331373

Server can specify fee rates, but not gas caps to use the in transactions.  That would be like telling the BTC backend what size in bytes when planning the transaction.  The client is in charge of authoring transactions the way they think is right.

chappjc: We actually have the size fields deprecated from `msgjson.Asset` now: https://github.com/decred/dcrdex/blob/f950443754abe7aa302404590be3476a3bfbbbf5/dex/msgjson/types.go#L1068-L1073

dex/networks/eth/tokens.go
chappjc: I bet that is because it was before the redeployment in https://github.com/decred/dcrdex/pull/2111
The contract didn't change much, but it was also a different solc version.  Not sure.

dex/networks/eth/tokens.go
chappjc: > I wonder why this is higher than either I or @JoeGruffins saw before. We both got 170k.

Same guess as in https://github.com/decred/dcrdex/pull/2013#discussion_r1100638399 with the updated contract.

dex/networks/eth/params.go
```diff
@@ -46,19 +46,19 @@ var (
...
-	Swap:      135000,
-	SwapAdd:   113000,
-	Redeem:    63000,
-	RedeemAdd: 32000,
-	Refund:    43000,
+	Swap:      174500, // 134,500 actual -- https://goerli.etherscan.io/tx/0xa17b6edeaf79791b5fc9232dc05a56d43f3a67845f3248e763b77162fae9b181, verified on mainnet
+	SwapAdd:   146400, // 112,600 actual (247,100 for 2) -- https://goerli.etherscan.io/tx/0xa4fc65b8001bf8c44f1079b3d97adf42eb1097658e360b9033596253b0cbbd04, verified on mainnet
+	Redeem:    78600,  // 60,456 actual -- https://goerli.etherscan.io/tx/0x5b22c48052df4a8ecd03a31b62e5015e6afe18c9ffb05e6cdd77396dfc3ca917, verified on mainnet
+	RedeemAdd: 41000,  // 31,672 actual (92,083 for 2, 123,724 for 3) -- https://goerli.etherscan.io/tx/0xae424cc9b0d43bf934112245cb74ab9eca9c2611eabcd6257b6ec258b071c1e6, https://goerli.etherscan.io/tx/0x7ba7cb945da108d39a5a0ac580d4841c4017a32cd0e244f26845c6ed501d2475, verified on mainnet
+	Refund:    68000,
```
chappjc, line 61: Did a manual refund on a short locktime swap, got 43,014.
https://goerli.etherscan.io/tx/0x586ed4cb7dab043f98d4cc08930d9eb291b0052d140d949b20232ceb6ad15f25
That was the last one I needed to verify.

dex/networks/eth/tokens.go
chappjc: Oh, when an account receives a token for the first time, I think that requires more gas to allocate storage space for the new non-zero balance, the next redeem will be cheaper (not referring to batches either).

I'm finding it's the same with token *transfer* too.  When an account receives USDC for the first time, it seems to be a 65,524 gas, but after that it's 48,424)

chappjc:
Verified and revised gas caps for refund, approve, and transfer.

Also revised up the token redeem gas cap to assume it's your first time receiving that token, thus requiring extra gas.  However, the redeems I have observed (with accounts already having USDC), were 79,012, not 83,850.

JoeGruffins (approved):
I have verified that the contracts at the mainnet addresses is the same as in our repo. Unfortunately unable to test on mainnet, but looks good.

JoeGruffins:
`Swap` will still fail if the gas used looks like more than we reserved in `FundOrder`. Is that ok?

chappjc:
> `Swap` will still fail if the gas used looks like more than we reserved in `FundOrder`. Is that ok?

Let me see if I got it right: `FundOrder` will reserve for swaps at a rate up to `MaxFeeRate`, but only using the `g.Swap` gas value from `(*assetWallet).swapGas`, which is either the know gas we're hard-coding now or 110% of a live estimate if it happens to be higher than the hard-coded gas.

Then when we make the swap, we are setting the fee limit (eip1559) to `MaxFeeRate`.  That means we have absolutely no wiggle room at `Swap` time.  This is exactly what happened to me before I revised up the token swap gas a few days ago, which you helped with.

```go
	if swapVal+fees > reservedVal {
		return fail("unfunded swap: %d < %d", reservedVal, swapVal+fees)
	}
```

So the question is, if there is more available balance that could be reserved, do we still fail?  Is that what we're trying to decide @JoeGruffins ?  Or something more straightforward?

**This implies there's no point in doing a live estimate at `Swap` time if we cannot use it.**

We're buffering these hard-coded values a good amount, so perhaps we shouldn't buffer more or in other places, but another option would be to buffer `ethToLock` in `FundOrder`, but that redundant.

Regardless of the above issues, I think it's more important that `Redeem` and `Refund` work regardless of reserves and balance.  Indeed we don't block in those methods.

dex/networks/eth/tokens.go
chappjc: >   First redeem used 79024 gas  

Here's where we need to consider if the receiving account has ever received that token.  What I found is that it's about ~83,850 for a redeem by an account that has no balance for that token yet, but ~79,012 for any redeem after that.

There's a similar effect with transfer.  ~65,524 for an initial receive, subsequent 48,424

> Average of 5 refunds: 64299 

Not sure how it's using more gas in the estimate utility vs real life: https://goerli.etherscan.io/tx/0xc5692ad0e6d86b721af75ff3b4b7c2e17d939918db030ebf5444ccf840c7a90b

I keep observing ~59,152 when actually swapping.  Is it possible that gas used here is different if the tx sender is the same vs. different than the token recipient of the refund tx?

dex/networks/eth/params.go
```diff
@@ -46,19 +46,19 @@ var (
...
-	Swap:      135000,
-	SwapAdd:   113000,
-	Redeem:    63000,
-	RedeemAdd: 32000,
-	Refund:    43000,
+	Swap:      174500, // 134,500 actual -- https://goerli.etherscan.io/tx/0xa17b6edeaf79791b5fc9232dc05a56d43f3a67845f3248e763b77162fae9b181, verified on mainnet
+	SwapAdd:   146400, // 112,600 actual (247,100 for 2) -- https://goerli.etherscan.io/tx/0xa4fc65b8001bf8c44f1079b3d97adf42eb1097658e360b9033596253b0cbbd04, verified on mainnet
+	Redeem:    78600,  // 60,456 actual -- https://goerli.etherscan.io/tx/0x5b22c48052df4a8ecd03a31b62e5015e6afe18c9ffb05e6cdd77396dfc3ca917, verified on mainnet
+	RedeemAdd: 41000,  // 31,672 actual (92,083 for 2, 123,724 for 3) -- https://goerli.etherscan.io/tx/0xae424cc9b0d43bf934112245cb74ab9eca9c2611eabcd6257b6ec258b071c1e6, https://goerli.etherscan.io/tx/0x7ba7cb945da108d39a5a0ac580d4841c4017a32cd0e244f26845c6ed501d2475, verified on mainnet
+	Refund:    57000,  // 43,014 actual -- https://goerli.etherscan.io/tx/0x586ed4cb7dab043f98d4cc08930d9eb291b0052d140d949b20232ceb6ad15f25
```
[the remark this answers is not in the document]
chappjc: The refund discrepancy here is smaller, but still there

https://goerli.etherscan.io/tx/0x586ed4cb7dab043f98d4cc08930d9eb291b0052d140d949b20232ceb6ad15f25
43,014

chappjc:
Another question I have pertains to the following code in `Swap`:

```go
	oneSwap, _, _, err := w.swapGas(1, swaps.Version)
	if err != nil {
		return fail("error getting gas fees: %v", err)
	}

	gasLimit := oneSwap * uint64(len(swaps.Contracts))
	fees := gasLimit * swaps.FeeRate

	if swapVal+fees > reservedVal {
		return fail("unfunded swap: %d < %d", reservedVal, swapVal+fees)
	}

	tx, err := w.initiate(w.ctx, w.assetID, swaps.Contracts, swaps.FeeRate, gasLimit, swaps.Version)
	if err != nil {
		return fail("Swap: initiate error: %w", err)
	}
```

We are calling `w.initiate` with `len(swaps.Contracts)` batched swaps, so why do we use `gasLimit := oneSwap * uint64(len(swaps.Contracts))` instead of just `nSwap` (the second output from `w.swapGas`)?

JoeGruffins:
> So the question is, if there is more available balance that could be reserved, do we still fail? Is that what we're trying to decide @JoeGruffins ? Or something more straightforward?

As you experienced on testnet, if the constants are too low this may happen. So, we make our best guess, but it may be wrong come initiate time. It does a live estimate for funding too though, so not sure why the gas amount changed so much between order time and swap time.

chappjc:
> so not sure why the gas amount changed so much between order time and swap time.

I think `FundOrder` skipped the estimate because the token approval was still pending.

chappjc:
> We are calling `w.initiate` with `len(swaps.Contracts)` batched swaps, so why do we use `gasLimit := oneSwap * uint64(len(swaps.Contracts))` instead of `gasLimit := nSwap` (the second output from `w.swapGas(n,...`)?
> I suppose because we've reserved that much so why not?

On this topic, I've added two commits to the other PR https://github.com/decred/dcrdex/pull/2118#issuecomment-1424468349

dex/networks/eth/params.go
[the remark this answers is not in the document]
JoeGruffins: > @JoeGruffins I'm confused because I don't see your **refunds** on the swap contract: https://goerli.etherscan.io/address/0x198463496037754564e9bea5418bf4117db0520c

I hadn't noticed til now, but it seems the gas estimate app does not refund, it only does `estimateRefundGas`

chappjc:
There are live gas estimate issues begin resolved in https://github.com/decred/dcrdex/pull/2129 and https://github.com/decred/dcrdex/pull/2118, but this PR has no more changes pending.

All the gases are verified with actual on-chain transactions, not just estimates.

chappjc:
Rebased and re-tested an ETH swap with DCR.  Again the current ETHSwap contract is https://etherscan.io/address/0x8c17e4968b6903e1601be82ca989c5b5e2c7b400

martonp:
I started a fresh dex, looked at the simnet exchange in view-only mode, had not yet tried to post any bonds, and I start seeing this log: `Bond asset 0 not supported by DEX 127.0.0.1:17273`

chappjc:
There's some hangover from view-only.  That looks like another thing.  Will look into that on the bond-reserves branch.
The other view-only related fix is in https://github.com/decred/dcrdex/pull/2025/commits/ae92a07a713cf85965152f03dce117010f55b79d

chappjc:
I don't quite get that error though if you didn't try to post any bonds (presumably didn't touch bond options either), I wonder why the view-only dex conn had `targetTier > 0`...

chappjc:
I understand mainnet testing is a burden, so if there are troubles on that front we can chat on matrix and I can provide guidance.
Unless there are specific concerns about the contracts or the updates in this PR, I'd like to have this merged by EOW.

chappjc:
Ran another USDC swap today.  Bad timing with gas (9am central is peak it seems), but otherwise it was fine.

@martonp Thanks for testing.  You reported on matrix that it worked smoothly, correct?

@JoeGruffins has verified the contract source code via etherscan matches the contract in this PR.  If one more person can also verify that, we can merge this.

https://etherscan.io/address/0x8c17e4968b6903e1601be82ca989c5b5e2c7b400#code#L1
https://etherscan.io/address/0x1bbd020ddd6dc01f974aa74d2d727b2a6782f32d#code#L1

martonp (approved):
I tested and it looks good to me!

PR #2015, opened 2022-12-30
webserver: Zero key after storing.

### LINKED ISSUE #2007: App password change isn't synched with Native wallets
When I use **Remember my password** on login, and then change App password to a different one Native wallets seem to be not aware of the change (while other parts of the app are aware: e.g. when asking to show seed new password works fine):

[password-change-issues.webm](https://user-images.githubusercontent.com/112318969/209271544-1f77c1c3-0da5-4638-bd7f-68d2a177273d.webm)

Note:
- the error ^ is different when I'm changing password for the 2nd time (from 8 symbols to 2). Or maybe it's just a different flavor of the same issue
- restarting **dexc** seems to be resolving the issue (logout probably too, I don't use it very often, didn't check)
- the issue doesn't seem to appear when **Remember my password** on login isn't checked

JoeGruffins:
closes #2007

### DIFF
--- a/client/webserver/api.go
+++ b/client/webserver/api.go
@@ -1086,8 +1086,8 @@ func (s *WebServer) apiChangeAppPass(w http.ResponseWriter, r *http.Request) {
 			log.Errorf("unable to cache password: %w", err)
 			clearCookie(pwKeyCK, w)
 		} else {
-			zero(key)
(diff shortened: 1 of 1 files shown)


chappjc (approved):
Whether or not this closes https://github.com/decred/dcrdex/issues/2007, this fix is righteous.

PR #2018, opened 2022-12-30
client/asset: use compatible wallet dir and rework moveWalletData

chappjc:
This reverts the BTC SPV wallet data directory so that there is no trailing `"spv"` folder.  This allows existing native BTC wallets created with 0.5 to be usable with 0.6, otherwise 0.6/master does not find them.

This also reworks how the wallet backups are done so that all files in the native wallet directory are moved one at a time, skipping the "logs" folder so that the log rotator used by the package-level loggers is not interrupted.  There is no longer an `"spvlogs"` folder; it is back to the `"logs"` folder adjacent to the other neutrino and wallet db files.

### DIFF
--- a/client/asset/bch/bch.go
+++ b/client/asset/bch/bch.go
@@ -127,9 +127,9 @@ func (d *Driver) Exists(walletType, dataDir string, settings map[string]string, 
 	if err != nil {
 		return false, err
 	}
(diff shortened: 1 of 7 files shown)


JoeGruffins (approved):
Can confirm that v0.5.8 -> master does not find spv wallets and this pr does.

buck54321 (approved):

PR #2019, opened 2023-01-01
core: reestablish old dex connection if new one is unsuccessful

ukane-philemon:
Issue:
Updating a dex cert to an invalid causes `core` to run a continuous retry using the bad cert (until dex is restarted). The old connection was disconnected and replaced even though the new connection failed.

Fix:
 This fix allows `core.UpdateCert` to specify that failed connection(while updating a dex's cert) should not replace the previous connection.

-- Modified `core.connectAccount` to accept an `addFailedConn bool` value which is used to control behaviour concerning saving a failed connection to the conns map. 

I should have opened an issue first but I think it was better to deal with it.

### DIFF
--- a/client/core/account.go
+++ b/client/core/account.go
@@ -1,10 +1,12 @@
 package core
 
 import (
+	"bytes"
 	"encoding/hex"
 	"errors"
 	"fmt"
 
+	"decred.org/dcrdex/client/comms"
 	"decred.org/dcrdex/client/db"
 	"decred.org/dcrdex/server/account"
 	"github.com/decred/dcrd/dcrec/secp256k1/v4"
@@ -208,23 +210,56 @@ func (c *Core) AccountImport(pw []byte, acct *Account, bonds []*db.Bond) error {
 
 // UpdateCert attempts to connect to a server using a new TLS certificate. If
 // the connection is successful, then the cert in the database is updated.
+// Updating cert for already connected dex will return an error.
 func (c *Core) UpdateCert(host string, cert []byte) error {
-	accountInfo, err := c.db.Account(host)
+	c.connMtx.RLock()
+	dc, found := c.conns[host]
+	c.connMtx.RUnlock()
+	if found && dc.status() == comms.Connected {
+		return errors.New("dex is already connected")
+	}
+
+	acct, err := c.db.Account(host)
 	if err != nil {
 		return err
 	}
 
-	accountInfo.Cert = cert
-
-	_, connected := c.connectAccount(accountInfo)
-	if !connected {
-		return errors.New("failed to connect using new cert")
+	// Ensure user provides a new cert.
+	if bytes.Equal(acct.Cert, cert) {
+		return errors.New("provided cert is the same with the old cert")
 	}
 
-	err = c.db.UpdateAccountInfo(accountInfo)
+	// Stop reconnect retry for previous dex connection first but leave it in
+	// the map so it remains listed incase we need it in the interim.
+	if found {
+		dc.connMaster.Disconnect()
+		dc.acct.lock()
+		dc.booksMtx.Lock()
+		for m, b := range dc.books {
+			b.closeFeeds()
+			if b.closeTimer != nil {
+				b.closeTimer.Stop()
+			}
+			delete(dc.books, m)
+		}
+		dc.booksMtx.Unlock()
+	}
+
+	acct.Cert = cert
+	dc, err = c.connectDEX(acct)
+	if err != nil {
+		if dc != nil {
+			dc.connMaster.Disconnect() // stop any retry loop for this new connection.
+		}
+		return fmt.Errorf("failed to connect using new cert (will attempt to restore old connection): %v", err)
+	}
+
+	err = c.db.UpdateAccountInfo(acct)
 	if err != nil {
 		return fmt.Errorf("failed to update account info: %w", err)
 	}
+
+	c.addDexConnection(dc)
 
 	return nil
 }
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1193,6 +1193,16 @@ func (c *Core) dex(addr string) (*dexConnection, bool, error) {
 	return dc, dc.status() == comms.Connected, nil
 }
 
+// addDexConnection is a helper used to add a dex connection.
+func (c *Core) addDexConnection(dc *dexConnection) {
+	if dc == nil {
+		return
+	}
+	c.connMtx.Lock()
+	c.conns[dc.acct.host] = dc
+	c.connMtx.Unlock()
+}
+
 // Get the *dexConnection for the host. Return an error if the DEX is not
 // connected.
 func (c *Core) connectedDEX(addr string) (*dexConnection, error) {
@@ -3723,9 +3733,7 @@ func (c *Core) upgradeConnection(dc *dexConnection) {
 		go c.listen(dc)
 		go dc.subPriceFeed()
 	}
-	c.connMtx.Lock()
-	c.conns[dc.acct.host] = dc
-	c.connMtx.Unlock()
+	c.addDexConnection(dc)
 }
 
 // DiscoverAccount fetches the DEX server's config, and if the server supports
@@ -3944,9 +3952,7 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
 	if paid {
 		registrationComplete = true
 		// The listen goroutine is already running, now track the conn.
-		c.connMtx.Lock()
-		c.conns[dc.acct.host] = dc
-		c.connMtx.Unlock()
+		c.addDexConnection(dc)
 
 		feeCoinStr := coinIDString(dc.acct.feeAssetID, dc.acct.feeCoin)
 		return &RegisterResult{FeeID: feeCoinStr, ReqConfirms: 0}, nil
@@ -3975,10 +3981,7 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
 		return nil, err
 	}
 	if paid { // would have gotten this from discoverAccount
-		c.connMtx.Lock()
-		c.conns[dc.acct.host] = dc
-		c.connMtx.Unlock()
-
+		c.addDexConnection(dc)
 		registrationComplete = true
 		// register already promoted the connection
 		feeCoinStr := coinIDString(dc.acct.feeAssetID, dc.acct.feeCoin)
@@ -4020,9 +4023,7 @@ func (c *Core) Register(form *RegisterForm) (*RegisterResult, error) {
 
 	// Registration complete.
 	registrationComplete = true
-	c.connMtx.Lock()
-	c.conns[host] = dc
-	c.connMtx.Unlock()
+	c.addDexConnection(dc)
 
 	err = c.db.CreateAccount(&db.AccountInfo{
 		Host:             dc.acct.host,
@@ -6341,8 +6342,7 @@ func (c *Core) initialize() {
 // connectAccount makes a connection to the DEX for the given account. If a
 // non-nil dexConnection is returned, it was inserted into the conns map even if
 // the initial connection attempt failed (connected == false), and the connect
-// retry / keepalive loop is active. If there was already a dexConnection, it is
-// first stopped.
+// retry / keepalive loop is active.
 func (c *Core) connectAccount(acct *db.AccountInfo) (dc *dexConnection, connected bool) {
 	// if !acct.Paid && len(acct.FeeCoin) == 0 {
 	// 	// Register should have set this when creating the account that was
@@ -6358,22 +6358,6 @@ func (c *Core) connectAccount(acct *db.AccountInfo) (dc *dexConnection, connecte
 		return
 	}
 
-	c.connMtx.RLock()
-	if dc := c.conns[host]; dc != nil {
-		dc.connMaster.Disconnect()
-		dc.acct.lock()
-		dc.booksMtx.Lock()
-		for m, b := range dc.books {
-			b.closeFeeds()
-			if b.closeTimer != nil {
-				b.closeTimer.Stop()
-			}
-			delete(dc.books, m)
-		}
-		dc.booksMtx.Unlock()
-	} // leave it in the map so it remains listed if connectDEX fails
-	c.connMtx.RUnlock()
-
 	dc, err = c.connectDEX(acct)
 	if dc == nil {
 		c.log.Errorf("Unable to prepare DEX %s: %v", host, err)
@@ -6386,9 +6370,7 @@ func (c *Core) connectAccount(acct *db.AccountInfo) (dc *dexConnection, connecte
 		connected = true
 	}
 
-	c.connMtx.Lock()
-	c.conns[host] = dc
-	c.connMtx.Unlock()
+	c.addDexConnection(dc)
 
 	return
 }
--- a/client/core/account_test.go
+++ b/client/core/account_test.go
@@ -307,9 +307,7 @@ func TestUpdateDEXHost(t *testing.T) {
 		rig.db.acct.LegacyFeeCoin = encode.RandomBytes(32)
 		rig.db.acct.Host = tDexHost
 
-		tCore.connMtx.Lock()
-		tCore.conns[rig.acct.host] = rig.dc
-		tCore.connMtx.Unlock()
+		tCore.addDexConnection(rig.dc)
 
 		rig.dc.pendingFee = nil
 		if test.feePending {
--- a/client/core/bond.go
+++ b/client/core/bond.go
@@ -860,9 +860,7 @@ func (c *Core) PostBond(form *PostBondForm) (*PostBondResult, error) {
 		if paid {
 			success = true
 			// The listen goroutine is already running, now track the conn.
-			c.connMtx.Lock()
-			c.conns[dc.acct.host] = dc
-			c.connMtx.Unlock()
+			c.addDexConnection(dc)
 			return &PostBondResult{ /* no new bond */ }, nil
 		}
 	}
@@ -997,9 +995,7 @@ func (c *Core) makeAndPostBond(dc *dexConnection, acctExists bool, wallet *xcWal
 	dc.acct.authMtx.Unlock()
 
 	if !acctExists { // *after* setting pendingBonds for rotateBonds accounting if targetTier>0
-		c.connMtx.Lock()
-		c.conns[dc.acct.host] = dc
-		c.connMtx.Unlock()
+		c.addDexConnection(dc)
 		// NOTE: it's still not authed if this was the first bond
 	}
 
--- a/client/core/core_test.go
+++ b/client/core/core_test.go
@@ -2066,9 +2066,7 @@ func TestRegister(t *testing.T) {
 	form.Addr = tDexHost
 
 	// account already exists
-	tCore.connMtx.Lock()
-	tCore.conns[tDexHost] = dc
-	tCore.connMtx.Unlock()
+	tCore.addDexConnection(dc)
 	_, err = tCore.Register(form)
 	if !errorHasCode(err, dupeDEXErr) {
 		t.Fatalf("wrong account exists error: %v", err)


chappjc:
Good catch.  I believe you've identified a regression.

> Updating a dex cert to an invalid causes `core` to run a continuous retry using the bad cert (until dex is restarted). The old connection was disconnected and replaced even though the new connection failed.

In https://github.com/decred/dcrdex/pull/1602 we attempted to prevent just this with handling in `(*wsConn).connect` and `(*wsConn).Connect`:  
https://github.com/decred/dcrdex/blob/63cbc663040c307d9fbd3a6d157f287112f7310e/client/comms/wsconn.go#L429-L437

https://github.com/decred/dcrdex/blob/63cbc663040c307d9fbd3a6d157f287112f7310e/client/comms/wsconn.go#L213-L221

Rather than adding bool args to one of the already-contorted connect helpers (which is only used in two places btw), let's identify why `wsConn` is no longer behaving as intended with respect to connect and cert errors.

I see there is removed code from `Connect` on master, so that could be it.  Also, I'm curious what precisely was invalid about the cert in your test (expired, wrong host, etc.?).

Also requesting review from @martonp too, because I believe there was reason to disconnect the old connection before making a new one (hence not using `disconnectDEX`), but I don't recall for certain.

chappjc:
I'm guessing that if you print the error from `(*wsConn).connect`, it would be of type `x509.CertificateInvalidError`.  Please supplement `connect` with a `fmt.Printf(%v, (%T)", err, err)` so we can inspect.

ukane-philemon:
>  I'm curious what precisely was invalid about the cert in your test (expired, wrong host, etc.?)

I selected a non-cert file and expected it to fail, but the old connection was removed and the new one will never make.

ukane-philemon:
I'm thinking we could just handle the cert update without using `core.connectAccount`.  Then all cert update logic will live in `core.UpdateCert`. I believe there is some edge cases `core. connectAccount` did not originally care about and should stay that way.

martonp:
> Also requesting review from @martonp too, because I believe there was reason to disconnect the old connection before making a new one (hence not using disconnectDEX), but I don't recall for certain. Context: https://github.com/decred/dcrdex/pull/1602#discussion_r878270518

I can't recall the reason for this either. 

> I selected a non-cert file and expected it to fail, but the old connection was removed and the new one will never make it.

@ukane-philemon What do you mean by this? When I select a non-cert file when trying to update the cert, I just get an error that it failed to connect. Also, I'm not able to reproduce this issue. I don't see the reconnect loop running when the connection failed due to invalid certificate. Could you tell me the exact steps you took?

chappjc:
> > Also requesting review from @martonp too, because I believe there was reason to disconnect the old connection before making a new one (hence not using disconnectDEX), but I don't recall for certain. Context: [#1602 (comment)](https://github.com/decred/dcrdex/pull/1602#discussion_r878270518)
> 
> I can't recall the reason for this either.

Seems the issue was order of notifications.  If we disconnect the old one after connecting the new one, it looks like we connect and then disconnect, confusing the user.

I'm also confused about what's not working here.  The error handling in `wsConn` appears to be working as intended unless you've hit a new error that we aren't catching with `errors.As`.

ukane-philemon:
Sorry for the confusion here(can't recall how I got to the retry loop initially), but I had mentioned that the retry loop will not run with my recent investigation. See https://github.com/decred/dcrdex/pull/2019#issuecomment-1368567473. 

But the main issue I intend to fix is maintaining the old dex connection when the new cert is invalid(maybe rejected by dex server) and also the confusion that could arise from connecting the new one before disconnecting the old one.

chappjc:
> Sorry for the confusion here(can't recall how I got to the retry loop initially), but I had mentioned that the retry loop will not run with my recent investigation. See [#2019 (comment)](https://github.com/decred/dcrdex/pull/2019#issuecomment-1368567473).
> 
> But the main issue I intend to fix is maintaining the old dex connection when the new cert is invalid(maybe rejected by dex server) and also the confusion that could arise from connecting the new one before disconnecting the old one.

Ok, still some unclear points here, but thanks for clarifying.

> I'm thinking we could just handle the cert update without using core.connectAccount. Then all cert update logic will live in core.UpdateCert. I believe there is some edge cases core. connectAccount did not originally care about and should stay that way.

Sounds better.  See about this idea please.  But we do need to have it disconnect the old connection before attempting a new connection because of the notification ordering issue, or at least find a way to prevent the confusing notification issue.  Another resolution would be to re-establish the prior connection (if it had been properly connected before).

martonp:
> But the main issue I intend to fix is maintaining the old dex connection when the new cert is invalid(maybe rejected by dex server) and also the confusion that could arise from connecting the new one before disconnecting the old one.

So you're saying if there is a good connection, then you try to update the cert to an invalid one, it will fail?

ukane-philemon:
> So you're saying if there is a good connection, then you try to update the cert to an invalid one, it will end the original connection?

Correct and there's currently no way to reconnect.

ukane-philemon:
> Another resolution would be to re-establish the prior connection (if it had been properly connected before).

Yeah, will look into a better solution to cover these:

1. Ensure the former connection is re-established if anything went wrong(especially if cert was rejected by the server).
2. Ensure the cert update flow does not confuse the user(i.e connecting and disconnection notes).

JoeGruffins:
I also cannot seem to replicate the issue on master. If I update to a bad cert it disconnects and does not seem to retry, although the log does say it will `2023-01-03 16:43:03.045 [ERR] CORE: Trouble establishing connection to 127.0.0.1:17273 (will retry). Error: connect failure: invalid certificate`

The log seems to be incorrect about retrying.

JoeGruffins (approved):
This fixes the log when it will not retry to reconnect. If the user is trying to change the cert, then probably the old one is not working so I don't think there is a great reason to try to connect with the old one again. But ok.

martonp (approved):

client/core/core.go
```diff
@@ -6221,22 +6221,6 @@ func (c *Core) connectAccount(acct *db.AccountInfo) (dc *dexConnection, connecte
...
-		dc.booksMtx.Lock()
-		for m, b := range dc.books {
-			b.closeFeeds()
-			if b.closeTimer != nil {
-				b.closeTimer.Stop()
-			}
-			delete(dc.books, m)
-		}
-		dc.booksMtx.Unlock()
-	} // leave it in the map so it remains listed if connectDEX fails
```
chappjc, line 6237: With `UpdateCert` using `c.disconnectDEX` now instead of this:
- the `dexConnection` vanishes from the `c.conns` map for a brief period (potentially not so brief if the connection takes time to establish, which can be several seconds over tor)
- if `connectDEX`  returns a nil conn, it does not get added back to the map at all, and if you refresh the page or get Exchanges again, you've completely lost the DEX until you restart or login again

client/core/account.go
```diff
@@ -208,24 +211,51 @@ func (c *Core) AccountImport(pw []byte, acct *Account, bonds []*db.Bond) error {
...
 	}
 
-	err = c.db.UpdateAccountInfo(accountInfo)
+	err = c.db.UpdateAccountInfo(acct)
 	if err != nil {
 		return fmt.Errorf("failed to update account info: %w", err)
 	}
 
+	connected = true
+	c.addDexConnection(dc)
```
chappjc, line 257: In other words, these lines are the danger zone where you've lost the `dexConnection` from the `Core.conns` map, and it may not come back at all, and if it does it come back, it was gone for a notable period.

ukane-philemon: > In other words, these lines are the danger zone where you've lost the dexConnection from the Core.conns map

We can stop the previous conn and leave it in the map if this brief duration is going to be an issue.

client/core/core.go
ukane-philemon: I can restore this in `UpdateCert` since it seems this brief duration could be an issue.

ukane-philemon: > if connectDEX returns a nil conn, it does not get added back to the map at all, 

If the connection is not successful it is never added but we will still have our old connection no matter what.

client/core/account.go
ukane-philemon: @chappjc, I believe your concern has been adequately handled in https://github.com/decred/dcrdex/pull/2019/commits/e776e8336885970a833233f970b2e12d962cdc3d, correct?

chappjc: Yes, thanks.

An unrelated/existing issue I think is what happens if you try to use `UpdateCert` on a host on which you are authorized (`authDEX` run).  Am I missing something or does this wipe out most of the important fields of the `dexConnection` when it is reconstructed freshly by `connectDEX`?  Have we just not tested doing a cert change on a functional DEX host?  I know the cert change function was added to deal with a DEX that has connectivity issues because of an expired certificate or something similar, but...  how was this supposed to work after it succeeds updating?

ukane-philemon: > wipe out most of the important fields of

What `important` fields are you specifically referring to?

chappjc: Most of the `acct` (`dexAccount`) fields, and everything set by `authDEX`.
`trades` and everything set by `resolveActiveTrades`.

`connectDEX` makes a nerfed `dexConnection`, and `connectAccount` is a very narrowly purposed helper generally only used from `initialize` (pre-login).

This PR is fine though.  We can look at this in a follow-up.

ukane-philemon: Ah, I see.

> We can look at this in a follow-up. 

Alright.

chappjc:
Ok, this makes the cut for 0.6, but I have to test briefly.

client/core/account.go
```diff
@@ -208,24 +211,61 @@ func (c *Core) AccountImport(pw []byte, acct *Account, bonds []*db.Bond) error {
...
-		return errors.New("failed to connect using new cert")
+	acct.Cert = cert
+	dc, err := c.connectDEX(acct)
+	if dc == nil {
+		return fmt.Errorf("failed to connect using new cert (will attempt to restore old connection): %v", err)
+	}
+	if err != nil {
+		dc.connMaster.Disconnect() // stop any retry loop for this new connection.
+		return fmt.Errorf("failed to connect using new cert (will attempt to restore old connection): %v", err)
 	}
```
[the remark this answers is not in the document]
ukane-philemon: Yeah.

client/core/account.go
```diff
@@ -13,6 +13,9 @@ import (
 // disconnectDEX unsubscribes from the dex's orderbooks, ends the connection
 // with the dex, and removes it from the connection map.
 func (c *Core) disconnectDEX(dc *dexConnection) {
+	if dc == nil {
+		return
+	}
```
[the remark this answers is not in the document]
ukane-philemon: I somehow passed a nil dc to this method while testing this fix. So I added a safety check. But can remove if not necessary.

client/core/account.go
```diff
@@ -250,11 +245,10 @@ func (c *Core) UpdateCert(host string, cert []byte) error {
 
 	acct.Cert = cert
 	dc, err := c.connectDEX(acct)
-	if dc == nil {
-		return fmt.Errorf("failed to connect using new cert (will attempt to restore old connection): %v", err)
-	}
 	if err != nil {
-		dc.connMaster.Disconnect() // stop any retry loop for this new connection.
+		if dc == nil {
```
buck54321, line 249: Should be `if dc != nil`.

ukane-philemon: Darn, I missed this. Thanks for the review.

client/core/account.go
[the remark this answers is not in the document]
buck54321: Just noting the **TODO** in `connectDEX`.

> TODO: split the above code into a dexConnection constructor, and the below into a startDexConnection function so we don't have the anti-pattern of returning a non-nil object with a non-nil error and requiring the caller to check both!

client/core/account.go
```diff
@@ -208,24 +208,58 @@ func (c *Core) AccountImport(pw []byte, acct *Account, bonds []*db.Bond) error {
...
+		}
+	}(acct.Cert)
+
+	// Disconnect previous dex connection first but leave it in the map so it
+	// remains listed incase we need it in the interim.
+	c.connMtx.RLock()
+	if dc := c.conns[host]; dc != nil {
+		dc.connMaster.Disconnect()
+		dc.acct.lock()
+		dc.booksMtx.Lock()
```
buck54321, line 234: Looking around, I notice that this would be the only place with the nested `c.connMtx.RLock -> dc.booksMtx.Lock`. As a general rule, I like to avoid nested mutex locking if there's not a compelling reason. Any reason we can't do 👇 ?

```Go
c.connMtx.RLock()
dc := c.conns[host]
c.connMtx.RUnlock()
if dc != nil {
```

buck54321 (approved):
Consider the mutex comment, but looks good.

client/core/account.go
ukane-philemon: Yes we can do that. No compelling reason except to save a LOC.

client/core/account.go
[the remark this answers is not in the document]
ukane-philemon: Right, this would even resolve https://github.com/decred/dcrdex/pull/2019#discussion_r1062027655

client/core/account.go
chappjc: Well I think we can do `dc, _ := c.dex(host)` as well.

client/core/account.go
chappjc: We could also only allow updating unless there are no active orders, so nothing would break.  Similar to the logout checks.

chappjc: Even worse, every `trackedTrade` carries a reference to the previous `dexConnection`.  I think I'm convinced we should prevent a cert update unless the connection is down or their are no active trades.  And updating in the first case would still require a restart to get any trades moving again.

ukane-philemon: > A more correct change would be to simply deny an update if the dexConnection is already connected

I agree.

ukane-philemon: > Even worse, every trackedTrade carries a reference to the previous dexConnection. I think I'm convinced we should prevent a cert update unless the connection is down or there are no active trades. And updating in the first case would still require a restart to get any trades moving again.

The issues related to this seem to be a rabbit hole. However, we could do the following:

1. If we have an already connected dex, reject the update cert request. https://github.com/decred/dcrdex/pull/2019/commits/8ccc00a85b2f2e732611683bf9631e3004298066
2. If not connected, use a new `*dexConnection`(even if host is found in the conns map), initialize the dex connection(i.e `authDEX`), then load and resume active trades. This would mean `core.UpdateCert` will require the app's password so we can resume trades without requiring a login to trigger trade resumption for the newly connected DEX.  Any thoughts? WIP.

client/core/account.go
```diff
@@ -208,24 +209,52 @@ func (c *Core) AccountImport(pw []byte, acct *Account, bonds []*db.Bond) error {
...
+		return fmt.Errorf("failed to connect using new cert (will attempt to restore old connection): %v", err)
 	}
 
-	err = c.db.UpdateAccountInfo(accountInfo)
+	err = c.db.UpdateAccountInfo(acct)
 	if err != nil {
 		return fmt.Errorf("failed to update account info: %w", err)
 	}
 
+	c.addDexConnection(dc)
```
chappjc, line 256: Between here and `connectDEX` above is where we could consider patching up the `dexAccount` and anything else (which is a lot of maps actually).  But as in our other comment thread, this is a major can of worms.  A different approach might be to patch the previous `dexConnection`'s `WsConn` and `connectionMaster`, but those are meant to be immutable so that sounds much worse.

client/core/account.go
ukane-philemon: > However, I'd like to first correct an unfortunate design pattern, which is having the "parent" types like dexConnection be fields of member types like trackedTrade. I've just completed a commit on a branch I can push shortly to illustrate.

Okay, sounds good.

client/core/account.go
ukane-philemon: > A different approach might be to patch the previous dexConnection's WsConn and connectionMaster, but those are meant to be immutable so that sounds much worse.

Yeah, thought of this too, but it didn't fit in with the existing design.

client/core/account.go
```diff
@@ -208,24 +209,52 @@ func (c *Core) AccountImport(pw []byte, acct *Account, bonds []*db.Bond) error {
...
 	if err != nil {
 		return err
 	}
 
-	accountInfo.Cert = cert
+	// Stop reconnect retry for previous dex connection first but leave it in
+	// the map so it remains listed incase we need it in the interim.
+	if found {
+		dc.connMaster.Disconnect()
+		dc.acct.lock()
```
buck54321, line 230: I see you disconnecting and locking the account, and I see you're leaving the original `dexConnection` in the map, but if we error at `connectDEX` below, this `dexConnection` will be pretty useless.

ukane-philemon: So, we are not supposed to lock the account, correct?

buck54321: Maybe? Right now, it seems like this PR has virtually no net effect. We've moved the disconnect/account lock/book unsub from `connectAccount` to `UpdateCert` and call `connectDEX` instead. But `connectAccount` was just `connectDEX` + disconnect/lock/unsub and adding to `c.conns`. And `connectAccount` is only called one other place, in `(*Core).initialize` where the `dexConnection` will certainly not be in `c.conns`, so no net effect there either. So I guess the only net effect is that the `dexConnection` is added to the map **after** we store the `db.AccountInfo` instead of before.

ukane-philemon: I agree. The new connection is not trade ready(even if we did connect) untill the user re-login. In another fix, we should prepare the new connection to be trade ready, just want to keep the scope of this fix small (i.e prevent disconnecting an already connected DEX).

chappjc (approved):
OK.  The `addDexConnection` helper looks useful.  Also, I like that `connectAccount` is no longer expected to deal with existing connections, only to be used as an initialize helper, and I'm almost certainly going to remove `connectAccount` in a bonds PR shortly, esp. with this change in first.

The issue of requiring a restart/re-login to get the restored dex connection working is indeed not new, but this is not needed immediately for 0.6.

From this discussion came https://github.com/decred/dcrdex/pull/2073, which is likely a prerequisite for being able to recreate an active `dexConnnection` used by trades, which would clearly be needed if the connection died because the cert expired when trades were running.

I don't want to continue along this line yet, because it's probably tempting to just call `initializeDEXConnections` again, but that's too heavy handed as it runs for all `dexConnections`.  Plus I'm formally designating that method as ONLY for use in Login, once.

client/core/account.go
```diff
@@ -208,24 +209,52 @@ func (c *Core) AccountImport(pw []byte, acct *Account, bonds []*db.Bond) error {
...
+			}
+			delete(dc.books, m)
+		}
+		dc.booksMtx.Unlock()
+	}
 
-	_, connected := c.connectAccount(accountInfo)
-	if !connected {
-		return errors.New("failed to connect using new cert")
+	acct.Cert = cert
```
chappjc, line 242: Thought.  Before disconnecting old, do a `bytes.Equal(cert, acct.Cert)` check so we don't do anything if they didn't change anything.

client/core/account.go
```diff
@@ -208,24 +209,52 @@ func (c *Core) AccountImport(pw []byte, acct *Account, bonds []*db.Bond) error {
...
 // UpdateCert attempts to connect to a server using a new TLS certificate. If
 // the connection is successful, then the cert in the database is updated.
+// Updating cert for already connected dex will return an error.
 func (c *Core) UpdateCert(host string, cert []byte) error {
-	accountInfo, err := c.db.Account(host)
+	c.connMtx.RLock()
+	dc, found := c.conns[host]
+	c.connMtx.RUnlock()
+	if found && dc.status() == comms.Connected {
+		return errors.New("dex is already connected")
```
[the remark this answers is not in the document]
ukane-philemon: Using the current flow will allow confirming from the `db `that the host actually exists(even if it does not exist in `c.conns`) but `c.dex(host)` will return an error if not found.

ukane-philemon:
@chappjc, Thanks for the rebase.

PR #2020, opened 2023-01-02
eth/server: Dumb down finding balance.

### LINKED ISSUE #2016: eth/server: Problem getting balances.
Seeing the error `2022-12-30 17:20:21.193 [ERR] MKT: (*DEXBalancer).CheckBalance: error getting account balance for %q: %v 0x946dfaB1AD7caCFeF77dE70ea68819a30acD4577 accountBalance error: contentFrom error: the method txpool_contentFrom does not exist/is not available` on master. Probably due to a recent geth version change. I do not think I noticed this in the pr, however.

JoeGruffins:
eth/server: Dumb down finding balance.

    eth/jwt: Add eth module check.

closes #2016

### DIFF
--- a/server/asset/eth/rpcclient.go
+++ b/server/asset/eth/rpcclient.go
@@ -18,7 +18,6 @@ import (
 	swapv0 "decred.org/dcrdex/dex/networks/eth/contracts/v0"
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/ethereum/go-ethereum/rpc"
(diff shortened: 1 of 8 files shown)


JoeGruffins:
Also ok to rewind the jwt endpoint. Just offering this up as another option.

chappjc:
Are we absolutely certain that specifying the apis via `--http.api=...` (and `--ws.api=...`) does not affect the apis provided on the `authrpc` interface as well?  The Prysm docs seem to suggest so.

chappjc:
Seeking clarification from geth devs: https://github.com/ethereum/go-ethereum/issues/26406

JoeGruffins:
Looks like #2023 is preferred, so closing.

server/asset/eth/rpcclient.go
```diff
@@ -183,92 +182,14 @@ func (c *rpcclient) transaction(ctx context.Context, hash common.Hash) (tx *type
...
-	// We need to subtract and pending outgoing value, but ignore any pending
-	// incoming value since that can't be spent until mined. So we can't using
-	// PendingBalanceAt or BalanceAt by themselves.
-	// We'll iterate tx pool transactions and subtract any value and fees being
-	// sent from this account. The rpc.Client doesn't expose the
-	// txpool_contentFrom => (*TxPool).ContentFrom RPC method, for whatever
-	// reason, so we'll have to use CallContext and copy the mimic the
-	// internal RPCTransaction type.
-	var txs map[string]map[string]*RPCTransaction
-	err = c.caller.CallContext(ctx, &txs, "txpool_contentFrom", addr)
```
chappjc, line 200: I haven't found any publicly available (paid or not) RPC providers that have any methods from the `txpool` namespace available.  I'm thinking we can have server identify which method to take on connect via the `CheckAPIModules` function you added, however I'm not seeing `rpc_modules` supported anywhere either!  Perhaps using a third-party RPC provider with server is gonna be trickier than I'd like.

JoeGruffins: Yeah, when I tested the modules were only available from geth itself.

chappjc: Been stewing on this, and I think we should revive it with a mini project to support an RPC provider that does not provide it, but use the smart balance if it does.

The aim would then be to allow the operator to use a third party RPC provider that they trust, ideally with their account's own API key so rate limits aren't so prevalent.

This is going to be key to make testnet or just mainnet "testing" less of burden.  At present, it's ridiculously difficult, time consuming, and expensive to get geth+prysm running on either networkk.

PR #2024, opened 2023-01-04
server: notify clients with unstable websocket connection

### LINKED ISSUE #778: server: characterize poor client connectivity and notify client
When the server observes unexpected websocket errors (e.g. i/o timeouts every ~ping interval), server could mark the anomaly in an IP-keyed map, and when that IP has many anomalies the server could send a notification that the client could display to the user to check their connectivity.

It has already been observed that a user's connectivity can be degraded in such a way that messages sent from server->client succeed, but the client apparently cannot respond.  This can be due to any number of things, from client's firewall to poor upstream connectivity (e.g. sat).

ukane-philemon:
This enables a DEX server to track and notify websocket clients with unstable connection i.e frequent disconnects/timeouts every ping interval. Closes #778 

Websocket clients with more than 3 disconnects within a one hour period since last connect will receive a message from the DEX server.

### DIFF
--- a/client/websocket/websocket.go
+++ b/client/websocket/websocket.go
@@ -172,7 +172,7 @@ func (s *Server) connect(ctx context.Context, conn ws.Connection, addr string) {
 	s.log.Tracef("Disconnected websocket client %s", addr)
 }
 
-// Notify sends a notification to the websocket client.
+// Notify sends a notification to the websocket clients.
 func (s *Server) Notify(route string, payload interface{}) {
 	msg, err := msgjson.NewNotification(route, payload)
 	if err != nil {
--- a/server/comms/comms_test.go
+++ b/server/comms/comms_test.go
@@ -33,11 +33,12 @@ var (
 
 func newServer() *Server {
 	return &Server{
-		clients:     make(map[uint64]*wsLink),
-		wsLimiters:  make(map[dex.IPKey]*ipWsLimiter),
-		v6Prefixes:  make(map[dex.IPKey]int),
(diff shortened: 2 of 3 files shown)


chappjc:
Thanks for tackling this long standing issue.  Still, I'm wondering if https://github.com/decred/dcrdex/issues/778 was a bad idea I had.  I can't think of a reason why the client itself wouldn't do this, instead of putting the burden on the server... and trying to send notifications to a client with a known bad link.  Also if it's the server with a bad uplink, it spams all clients.  I was bulk-labeling issues the other day and should have noticed this one.

Some thoughts:
- The server side machinery you added generally looks well-done, so perhaps there other ways servers can utilize this info.  Not sure.
- Can some of the logic be ported to client instead?

So sorry this issue was up so long, but I just don't get what I was thinking at the time (2.5 years ago now).  Thoughts from others about this? @buck54321 @JoeGruffins @martonp

ukane-philemon:
> Can some of the logic be ported to client instead?

I can see to this.

> so perhaps there other ways servers can utilize this info. Not sure.

Except for this current purpose, can't think of anything yet. ws clients with bad connectivity should not be prevented from placing trades, right?

chappjc:
> Except for this current purpose, can't think of anything yet. ws clients with bad connectivity should not be prevented from placing trades, right?

Same thought, would be extreme.

martonp:
I agree, the client should be able to handle this.

ukane-philemon:
Since there's no need for this feature on the server, I've implemented them client-side in #2028.

Issue #1990, opened 2022-12-08
Client-side bonding level/tier maintenance - auto posting bonds to keep specified tier


chappjc:
With all the foundational piece in place for bonds, this is one of the next steps https://github.com/orgs/decred/projects/2

This issue pertains the backend work in client to maintaining a certain desired tier with limits on total bonding amount.

> Next steps will be multifaceted. UI, updated messaging (suspended -> sub-tier or whatever), client bond maintenance, settings, switches, etc.
> 
> I will begin client db and core work for maintaining a certain tier with limits on total bonding amount.

https://github.com/decred/dcrdex/pull/1820#issuecomment-1335511711

chappjc:
Unassigning this from myself temporarily so I can review and merge Joe's 2 ETH PRs first, which are top priority.
I intend to pick this up, but it's fair game in the meantime.

PR #2025, opened 2023-01-04
app: replace legacy fee registration with postbond

### LINKED ISSUE #1991: Switch client browser frontend to postbond instead of legacy fee register
With all the foundational piece in place for bonds, this is one of the next steps https://github.com/orgs/decred/projects/2

For the UI, what we do depends on if we intend to support both the legacy fee and bonds in the next release. I don't think we should give the choice, only bonds.

So, while we transition in development, I think we should start by showing two options at /register - one that uses the legacy fee and one that uses a bond. But since it's only for developers to use either as needed, I don't think we should make it a single polished form with a selector. In fact, two separate forms shown simultaneously where you just use one of the forms might make the most sense. Otherwise if we make a nice selector on a single form, we're going to have to change that when the legacy fee is dropped. If there are two separate forms shown side-by-side, we just remove one when it's time to drop the fee for the next release.

itswisdomagain:
Resolves #1991.

Registration flow is largely maintained, except the /postbond API endpoint
is used instead of /register; and the confirm registration form allows users
to specify a bond strength.

Markets page shows confirmation counter for pending bonds which is only
displayed if the user's account tier is < 1 and there are pending bonds.

Also adds a Post Bond feature to the dexsettings page for manual bond posting.


chappjc:
Thanks!  I'm rushing to resolve https://github.com/decred/dcrdex/issues/1990 so we can iron out the UI elements of bond options, namely target tier and max bonded amount.

itswisdomagain:
The reg asset selector and other forms used in posting a bond still contain "registration fee" terminology. Variable, class, method names as well. I believe those need to be updated but want to put the core stuff out here first.

client/core/bond.go
```diff
@@ -1016,18 +1019,29 @@ func (c *Core) makeAndPostBond(dc *dexConnection, acctExists bool, wallet *xcWal
...
 	c.monitorBondConfs(dc, bond, reqConfs)
 
 	return bond.CoinID, nil
 }
 
+func (c *Core) updatePendingBondConfs(dc *dexConnection, assetID uint32, coinID []byte, confs uint32) {
+	dc.acct.authMtx.Lock()
+	defer dc.acct.authMtx.Unlock()
+	bondIDStr := coinIDString(assetID, coinID)
+	dc.acct.pendingBondsConfs[bondIDStr] = confs
```
martonp, line 1034: Got this error here: `panic: assignment to entry in nil map`
It happened after I clicked register.

martonp (commented):
I wasn't able to post the bond because I kept getting the panic because `pendingBondConfs` was nil.

client/core/bond.go
itswisdomagain: Ah, crap. Will fix and push shortly.

client/core/bond.go
```diff
@@ -908,9 +906,15 @@ func (c *Core) PostBond(form *PostBondForm) (*PostBondResult, error) {
...
+		maxBondedAmt := 4 * form.Bond // default
+		if form.MaxBondedAmt != nil {
+			maxBondedAmt = *form.MaxBondedAmt
+		}
 		dc.acct.authMtx.Lock()
+		dc.acct.bondAsset = bondAssetID
 		dc.acct.targetTier = form.Bond / bondAsset.Amt
+		dc.acct.maxBondedAmt = maxBondedAmt
 		dc.acct.authMtx.Unlock()
 		// maxBondedAmt and bondAsset were set by connectDEX
```
chappjc, line 919: no longer correct if view-only acct?

client/core/bond.go
```diff
@@ -807,12 +809,14 @@ func (c *Core) PostBond(form *PostBondForm) (*PostBondResult, error) {
...
-		if form.MaintainTier != nil || form.MaxBondedAmt != nil {
-			return nil, fmt.Errorf("maintain tier and max bonded amount may only be set when registering " +
-				"(use UpdateBondOptions to change bond maintenance settings)")
+		if acctExists {
+			if dc.acct.locked() { // require authDEX first to reconcile any existing bond statuses
+				return nil, newError(acctKeyErr, "acct locked %s (login first)", form.Addr)
+			}
+			if form.MaintainTier != nil || form.MaxBondedAmt != nil {
+				return nil, fmt.Errorf("maintain tier and max bonded amount may only be set when registering " +
+					"(use UpdateBondOptions to change bond maintenance settings)")
```
chappjc, line 818: whoops, yeah, these checks must all be under `acctExists`.  Missed this in view-only: https://github.com/decred/dcrdex/commit/02374f771e944b2f89c7ff81348571cb23456876#diff-6327e534faf5bc61df70b157ad5474582e26a4fd99bceb53316d758e9a5fe9bdR807

client/core/bond.go
itswisdomagain: No longer correct in all cases. They’re only set here now, if maintain is true and it’s a fresh registration attempt (!acctExists).

ukane-philemon:
> Thanks! I'm rushing to resolve #1990 so we can iron out the UI elements of bond options, namely target tier and max bonded amount.

About max bonded amount, assuming a user has zero trading tier(and maybe the expired tier has not been refunded) and then decides to set bond options to (target tier := 1, max bonded amount := target tier * bond amount) there would be an issue when maintaining bonds since the previous bond has not been refunded but we want to post a bond which would exceed the current `maxBondAmt`.

`core.UpdateBondOptions` and `core.PostBond` currently does not do any validations but considering the above scenario and ensuring the provided `MaxBondAmount` is sufficient to maintain the target tier, we might have to do some validation there so users are not stuck waiting for a new bond/tier increase and do not know they need to increase the max bond amount to `current bonded amount(including expired but not refunded) + amtNeeded(target tier * bond amount) + amtNeeded * 4(or maybe more or less?)`. The last variable is excluded if `maintainTier != true`.

I guess since #1990 is merged, we can discuss this further.

chappjc:
Referring to https://github.com/decred/dcrdex/pull/2103/files#diff-6327e534faf5bc61df70b157ad5474582e26a4fd99bceb53316d758e9a5fe9bdR912

chappjc:
For UI, we generally do NOT want the user setting `maxBondAmt`.  This setting should be an advanced option, and hidden from users.  Info text should say the max bonded amount is automatically set unless the user picks a larger value (to compensate for possible tier drops).  The backend enforces the "floor" according to the code I linked.

ukane-philemon:
> Referring to https://github.com/decred/dcrdex/pull/2103/files#diff-6327e534faf5bc61df70b157ad5474582e26a4fd99bceb53316d758e9a5fe9bdR912

Right. I agree the bonds reserves PR might need to go in before finally resolving the UI elements regarding bonds, since it has extensive changes.

Regarding having an advance bonds settings section(esp. for `maxBondAmount`), that makes sense so regular users need not fret over settings they do not need to meddle with.

chappjc:
> Right. I agree the bonds reserves PR might need to go in before finally resolving the UI elements regarding bonds, since it has extensive changes.

You may rebase on the `bond-reserves` branch.  I think that would aid in testing both PRs.  I do `dexcctl bondopts` all the time instead.

chappjc:
@ukane-philemon If you are taking this over, do you want to make a new PR or can you push on this branch?

chappjc:
@itswisdomagain Did you wanna do the rebase and a couple more tweaks before handing it over?  Or did I misunderstand matrix chat?

ukane-philemon:
> The reg asset selector and other forms used in posting a bond still contain "registration fee" terminology. Variable, class, method names as well. I believe those need to be updated but want to put the core stuff out here first.

While at it, we could update these too.

chappjc:
I want to keep the footprint of this PR minimal to get it in fast.  If we update that stuff (for completeness, but unnecessary), we need to isolate it into a separate commit, but I still expect breakage doing it.

itswisdomagain:
> @itswisdomagain Did you wanna do the rebase and a couple more tweaks before handing it over? Or did I misunderstand matrix chat?

I'll rebase and cherry-pick @ukane-philemon's commit for the bond maintenance option, so this can go in without further delay.

JoeGruffins (approved):
Working well. Just some UI nits.
I still see "Registration Fee" used everywhere. Is that still accurate? Not only here, but for example:
image

The bond amount inputs allow negatives and letters, can restrict to whole numbers? Examples:
This does error.
image

This one does not error.
image

JoeGruffins:
Some warnings and errors, not sure if related to these changes:
```
2023-02-23 20:35:26.112 [INF] CORE: Connected to DEX server at 127.0.0.1:17273 but NOT listening for messages.
2023-02-23 20:35:26.124 [INF] CORE: Authenticated connection to 127.0.0.1:17273, acct ca230ff6a2c6cae19902cd5dd8cc4a7505b0ecb36a85611d45d779ce379bba86, 1 active bonds, 0 active orders, 0 active matches, score -1, tier 6
2023-02-23 20:35:26.124 [WRN] CORE: Unknown bond reported by server: 6c1194ccea72666667de5590d575ae164b833a9fabd5792c3508d89f103ddfe5:0 (dcr)
2023/02/23 20:35:26 "POST http://127.0.0.3:5758/api/discoveracct HTTP/1.1" from 127.0.0.1:46216 - 200 2340B in 58.78178ms
2023/02/23 20:35:26 "GET http://127.0.0.3:5758/api/user HTTP/1.1" from 127.0.0.1:46216 - 200 49550B in 492.624µs
2023/02/23 20:35:26 "GET http://127.0.0.3:5758/markets HTTP/1.1" from 127.0.0.1:46216 - 200 49175B in 236.284µs
2023-02-23 20:35:26.176 [TRC] WEB[WS]: message of type 1 received for route loadmarket
2023-02-23 20:35:26.176 [DBG] CORE: Subscribing to the dcr_btc order book for 127.0.0.1:17273
2023-02-23 20:35:26.176 [DBG] CORE[dcr_btc][book]: Processing 0 cached order notes
2023-02-23 20:35:26.176 [TRC] WEB[WS]: message of type 1 received for route loadcandles
2023/02/23 20:35:26 "GET http://127.0.0.3:5758/img/coins/btc.png HTTP/1.1" from 127.0.0.1:46216 - 200 3636B in 123.582µs
2023/02/23 20:35:26 "GET http://127.0.0.3:5758/img/coins/dextt.eth.png HTTP/1.1" from 127.0.0.1:46216 - 200 7041B in 99.156µs
2023/02/23 20:35:26 "GET http://127.0.0.3:5758/img/coins/eth.png HTTP/1.1" from 127.0.0.1:46216 - 200 2593B in 80.24µs
2023-02-23 20:35:29.352 [INF] CORE: Newly expired bond found: 6c1194ccea72666667de5590d575ae164b833a9fabd5792c3508d89f103ddfe5:0 (dcr)
```
```
2023-02-23 20:40:09.352 [ERR] CORE: Failed to broadcast bond refund txn []: EOF
2023-02-23 20:40:15.012 [TRC] CORE[dcr_btc][book]: Validating match proof note for epoch 111810160 (dcr_btc) with 6 preimages and 0 misses.
2023-02-23 20:40:17.532 [DBG] CORE[dcr]: tip change: 658 (0d8c1715d1d8e4164fa7d92eb6008ecea5d9376c1518b2609b5c4f72b0ff6286) => 659 (2168d66c145e3f9fbe93d2200f7e71a91d44ff3e9e813e3839f0c79a2cf3d320)
2023-02-23 20:40:17.532 [TRC] CORE: Processing tip change for dcr
2023-02-23 20:40:22.903 [DBG] CORE[btc]: tip change: 466 (71585b7a4adc898afdb79d57bb511f32259c39b3f3ee82cf7c6545412c1ae836) => 467 (677c638a3201f6ac8fb1aa8e8fb53f0198e6417c5c532b55d2a37e92c89f3653)
2023-02-23 20:40:22.903 [TRC] CORE: Processing tip change for btc
2023-02-23 20:40:22.903 [TRC] CORE[btc][SPV]: Bals: spendable = 0.0086014 BTC (0.0086014 BTC trusted, 0 BTC untrusted, 0 BTC assumed locked), immature = 0 BTC
2023-02-23 20:40:29.352 [ERR] CORE: Failed to broadcast bond refund txn []: EOF
2023-02-23 20:40:30.017 [TRC] CORE[dcr_btc][book]: Validating match proof note for epoch 111810161 (dcr_btc) with 7 preimages and 0 misses.
2023-02-23 20:40:39.189 [DBG] CORE[dcr]: tip change: 659 (2168d66c145e3f9fbe93d2200f7e71a91d44ff3e9e813e3839f0c79a2cf3d320) => 660 (1666c692efd31650091a7d7c39aa9f5a9bdcc0f41f9bd0220c1c61f662c56666)
2023-02-23 20:40:39.189 [TRC] CORE: Processing tip change for dcr
2023-02-23 20:40:45.018 [TRC] CORE[dcr_btc][book]: Validating match proof note for epoch 111810162 (dcr_btc) with 7 preimages and 0 misses.
2023-02-23 20:40:49.272 [DBG] CORE[btc]: tip change: 467 (677c638a3201f6ac8fb1aa8e8fb53f0198e6417c5c532b55d2a37e92c89f3653) => 468 (646fef046fdb89e3bd30362c3b5efeca9aecaac8885340950677c2162015237e)
2023-02-23 20:40:49.272 [TRC] CORE: Processing tip change for btc
2023-02-23 20:40:49.272 [TRC] CORE[btc][SPV]: Bals: spendable = 0.0086014 BTC (0.0086014 BTC trusted, 0 BTC untrusted, 0 BTC assumed locked), immature = 0 BTC
2023-02-23 20:40:49.352 [ERR] CORE: Failed to broadcast bond refund txn []: EOF
```

I was deleting the account in options and then connecting again. Maybe related...

JoeGruffins:
Oh, yeah, maybe unrelated, but issue a bond to a dex, then "Disable Account" in options then connect to the dex again and you get those logs. It appears to brick the refund.

chappjc:
> Oh, yeah, maybe unrelated, but issue a bond to a dex, then "Disable Account" in Settings then connect to the dex again and you get those logs. It appears to brick the refund.

Looks like we should block Disable Account if there are active bonds!

chappjc:
> > The reg asset selector and other forms used in posting a bond still contain "registration fee" terminology. Variable, class, method names as well. I believe those need to be updated but want to put the core stuff out here first.
> 
> While at it, we could update these too.

To clarify, we still need to update the UI.  I just meant let's not churn on "Variable, class, method names as well".

"How will you post the registration fee?"  => "What asset would you like to use for fidelity bonds?"

"...funds will be spent from your wallet to play the registration fee." => "...funds will be locked in time-locked fidelity bonds, which you may redeem in the future."

etc.

chappjc:
> Looks like we should block Disable Account if there are active bonds!

Just did this in https://github.com/decred/dcrdex/pull/2103

I think this PR should rebase on that one.

itswisdomagain:
> Rebased and force pushed on the remote.

Thanks @chappjc. Sorry for the delays.

itswisdomagain:
My last force push is a [minimal change](https://github.com/decred/dcrdex/compare/812f66b5228f5901742d796038f558aa17334e9d..4b86a8ddd7c44607c79f2a4d0ae2aa36c5bc638b) to how the `Update Bond Options` button is hidden for view-only connections.

buck54321 (changes requested):

client/core/bond.go
```diff
@@ -1199,13 +1203,14 @@ func (c *Core) makeAndPostBond(dc *dexConnection, acctExists bool, wallet *xcWal
...
-			BondAsset:  dc.acct.bondAsset,
-			TargetTier: dc.acct.targetTier,
+			Host:         dc.acct.host,
+			Cert:         dc.acct.cert,
+			DEXPubKey:    dc.acct.dexPubKey,
+			EncKeyV2:     dc.acct.encKey,
+			Bonds:        []*db.Bond{dbBond},
+			TargetTier:   dc.acct.targetTier,
+			MaxBondedAmt: dc.acct.maxBondedAmt,
+			BondAsset:    dc.acct.bondAsset,
```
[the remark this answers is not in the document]
chappjc: Good catch.  I'd use bondAsset, targetTier, maxBondedAmt := dc.bondopts()` just before.

buck54321 (approved):

chappjc (approved):

PR #2028, opened 2023-01-04
client: send notification for unstable websocket connection

ukane-philemon:
This should supersede #2024.

UI view: wording changed though.
Screenshot 2023-01-05 at 12 19 02 AM

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -66,11 +66,12 @@ const (
 	// or a config response field if it should be considered variable.
 	preimageReqTimeout = 20 * time.Second
 
-	// maxClientAnomaliesCount is the maximum websocket connection anomaly
-	// before client is sends a notification to check their connection.
-	maxClientAnomaliesCount = 3
-	// If websocket client disconnects before wsAnomalyDuration since last
-	// connect time, the websocket client anomaly count is increased.
+	// wsMaxAnomalyCount is the maximum websocket connection anomaly after which
+	// a client receives a notification to check their connectivity.
+	wsMaxAnomalyCount = 3
+	// If a client's websocket connection to a server disconnects before
+	// wsAnomalyDuration since last connect time, the client's websocket
+	// connection anomaly count is increased.
 	wsAnomalyDuration = 60 * time.Minute
 )
 
@@ -7374,7 +7375,7 @@ func (c *Core) handleReconnect(host string) {
 	}
 
 	anomalies := atomic.LoadUint32(&dc.anomaliesCount)
-	if anomalies > maxClientAnomaliesCount {
+	if anomalies > wsMaxAnomalyCount {
 		// Send notification to check connectivity.
 		subject, details := c.formatDetails(TopicDexConnectivity, host)
 		c.notify(newConnEventNote(TopicDexConnectivity, subject, host, dc.status(), details, db.Poke))
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -7531,8 +7531,7 @@ func (c *Core) handleConnectEvent(dc *dexConnection, status comms.ConnectionStat
 		dc.lastConnectMtx.RLock()
 		lastConnect := dc.lastConnect
 		dc.lastConnectMtx.RUnlock()
-
-		if atomic.CompareAndSwapUint32(&dc.anomaliesCount, wsMaxAnomalyCount, 0 /* reset anomalies count */) {
+		if atomic.LoadUint32(&dc.anomaliesCount)%wsMaxAnomalyCount == 0 {
 			// Send notification to check connectivity.
 			subject, details := c.formatDetails(TopicDexConnectivity, dc.acct.host)
 			c.notify(newConnEventNote(TopicDexConnectivity, subject, dc.acct.host, dc.status(), details, db.Poke))
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -7376,8 +7376,8 @@ func (c *Core) handleReconnect(host string) {
 	anomalies := atomic.LoadInt64(&dc.anomaliesCount)
 	if anomalies > maxClientAnomaliesCount {
 		// Send notification to check connectivity.
-		subject, details := c.formatDetails(TopicCheckConnectivity)
-		c.notify(newClientNotifyNote(TopicCheckConnectivity, subject, details, db.Poke))
+		subject, details := c.formatDetails(TopicDexConnectivity, host)
+		c.notify(newConnEventNote(TopicDexConnectivity, subject, host, dc.status(), details, db.Poke))
 		atomic.StoreInt64(&dc.anomaliesCount, 0) // reset anomalies count.
 	}
 
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -172,7 +172,7 @@ type dexConnection struct {
 	spots    map[string]*msgjson.Spot
 
 	// anomaliesCount tracks client's connection anomalies.
-	anomaliesCount int64 // atomic
+	anomaliesCount uint32 // atomic
 	lastConnect    time.Time
 }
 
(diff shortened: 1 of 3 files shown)


client/core/notification.go
```diff
@@ -556,7 +556,7 @@ const (
 
 func newServerNotifyNote(topic Topic, subject, details string, severity db.Severity) *ServerNotifyNote {
 	return &ServerNotifyNote{
-		Notification: db.NewNotification(NoteTypeServerNotify, topic, subject, details, severity),
+		Notification: db.NewNotification(NoteTypeNotify, topic, subject, details, severity),
 	}
 }
```
ukane-philemon, line 561: I was thinking about having a single `NotifyNote` for this and the new `newClientNotifyNote` below . Any thoughts?

chappjc: No.  All these notes are originating from the client, and only the _one_ that is forwarded from the server requires any qualification on the source.
`ClientNotifyNote` doesn't make sense to me.  This PR is adding a notification regarding connectivity with a server.  That is the kind of note to define (e.g. `DexConnectivityNote`), if it has to be a new type, which is questionable.  I suspect there's already a fairly generic  note type that can be given an appropriate `subject` for this case.

ukane-philemon: Okay, there's `ConnEventNote`  which I think will be suitable. I had intended not to bind connectivity messages to a specific dex connection.

client/core/core.go
```diff
@@ -163,6 +170,10 @@ type dexConnection struct {
 
 	spotsMtx sync.RWMutex
 	spots    map[string]*msgjson.Spot
+
+	// anomaliesCount tracks client's connection anomalies.
+	anomaliesCount int64 // atomic
```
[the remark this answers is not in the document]
ukane-philemon: Oh, right.

client/core/core.go
```diff
@@ -7511,7 +7531,13 @@ func (c *Core) handleConnectEvent(dc *dexConnection, status comms.ConnectionStat
 	topic := TopicDEXDisconnected
 	if status == comms.Connected {
 		topic = TopicDEXConnected
+		dc.lastConnect = time.Now()
 	} else {
+		if time.Since(dc.lastConnect) < wsAnomalyDuration {
+			// Increase anomalies count for this connection.
+			atomic.StoreUint32(&dc.anomaliesCount, atomic.LoadUint32(&dc.anomaliesCount)+1)
```
chappjc, line 7538: `atomic.AddUint32(&dc.anomaliesCount, 1)`?

client/core/core.go
```diff
@@ -7511,7 +7530,13 @@ func (c *Core) handleConnectEvent(dc *dexConnection, status comms.ConnectionStat
 	topic := TopicDEXDisconnected
 	if status == comms.Connected {
 		topic = TopicDEXConnected
+		dc.lastConnect = time.Now()
 	} else {
+		if time.Since(dc.lastConnect) < wsAnomalyDuration {
```
buck54321, line 7535: Since `lastConnect` has no mutex protection but is only used in `handleConnectEvent`, there is an assumption that `(*wsConn).setConnectionStatus` will never be called concurrently. Are we sure about that?

ukane-philemon: > Are we sure about that?

Double-checked, and  I think we need mutex protection for `lastConnect` because `(*wsConn).setConnectionStatus` is called from `(*wsConn).keepAlive`. Added in https://github.com/decred/dcrdex/pull/2028/commits/5c8f10f8bf3f47007fc056d5900cf2231af7679f

client/core/core.go
```diff
@@ -7362,6 +7375,13 @@ func (c *Core) handleReconnect(host string) {
 		return
 	}
 
+	if atomic.LoadUint32(&dc.anomaliesCount) > wsMaxAnomalyCount {
+		// Send notification to check connectivity.
+		subject, details := c.formatDetails(TopicDexConnectivity, host)
+		c.notify(newConnEventNote(TopicDexConnectivity, subject, host, dc.status(), details, db.Poke))
+		atomic.StoreUint32(&dc.anomaliesCount, 0) // reset anomalies count.
+	}
+
```
buck54321, line 7384: I'm wondering if we should do this on the **disconnect** instead of the connect. Down where you're doing 
```Go
atomic.AddUint32(&dc.anomaliesCount, 1)
```

chappjc: That would also be nice because with mods done in a single place, it could be properly atomic with `if atomic.CompareAndSwapUint32(&dc.anomaliesCount, wsMaxAnomalyCount, 0) { /* notify */ }`

chappjc: Although now that I say that, it could just be a `atomic.AddUin32(&dc.anomaliesCount, 1) % wsMaxAnomalyCount == 0`

ukane-philemon:
It looks like there's an issue with the linter. https://github.com/ukane-philemon/dcrdex/actions/runs/4028889667/jobs/6926242030

chappjc:
> It looks like there's an issue with the linter. https://github.com/ukane-philemon/dcrdex/actions/runs/4028889667/jobs/6926242030

rerunning

chappjc:
Looks like github started ignoring the content-type request header that goreleaser needs
curl -H 'content-type: application/json' https://github.com/golangci/golangci-lint/releases/v1.50.1

chappjc (approved):
LGTM.  We'll have to see how this feels on a variety of real internet hosts including hidden services. I think we can merge and get this running while I'm testing long term bond rotation.  May tweak `wsMaxAnomalyCount`, but looks good to start.

chappjc:
Oh, I just noticed you killed `lastConnect`.  Virtually every user will get a scary unstable connection warning after running long enough. What was the reason for removing that since last review?

ukane-philemon:
> What was the reason for removing that since last review?

This https://github.com/decred/dcrdex/pull/2028#discussion_r1089527870.

chappjc:
Oh.  I didn't expect that would preclude checking duration since last, or
at least last nag. Huh

client/core/core.go
```diff
@@ -7511,7 +7524,22 @@ func (c *Core) handleConnectEvent(dc *dexConnection, status comms.ConnectionStat
...
+		lastConnect := dc.lastConnect
+		dc.lastConnectMtx.RUnlock()
+		if atomic.LoadUint32(&dc.anomaliesCount)%wsMaxAnomalyCount == 0 {
+			// Send notification to check connectivity.
+			subject, details := c.formatDetails(TopicDexConnectivity, dc.acct.host)
+			c.notify(newConnEventNote(TopicDexConnectivity, subject, dc.acct.host, dc.status(), details, db.Poke))
+		} else if time.Since(lastConnect) < wsAnomalyDuration {
+			// Increase anomalies count for this connection.
+			atomic.AddUint32(&dc.anomaliesCount, 1)
+		}
```
buck54321, line 7541: Actually, I wonder if it should be something like
```Go
dc.lastConnectMtx.RLock()
lastConnect := dc.lastConnect
dc.lastConnectMtx.RUnlock()
if time.Since(lastConnect) < wsAnomalyDuration {
	// Increase anomalies count for this connection.
	count := atomic.AddUint32(&dc.anomaliesCount, 1)
	if count%wsMaxAnomalyCount == 0 {
		// Send notification to check connectivity.
		subject, details := c.formatDetails(TopicDexConnectivity, dc.acct.host)
		c.notify(newConnEventNote(TopicDexConnectivity, subject, dc.acct.host, dc.status(), details, db.Poke))
	}
} else {
	atomic.StoreUint32(&dc.anomaliesCount, 0)
}
```

ukane-philemon: > atomic.StoreUint32(&dc.anomaliesCount, 0)

@chappjc, had suggested we should not reset to zero https://github.com/decred/dcrdex/pull/2028#discussion_r1089527870. But I just noticed there is something I was reminded of rn.

chappjc: That comment was to avoid a sequential (non-atomic) load/store that you previously had depending on `wsMaxAnomalyCount`.  Resetting to zero based on time since last anomaly is a different matter.

ukane-philemon: >  Resetting to zero based on time since last anomaly is a different matter.

I agree.  I was referring to the part of your comment that mentioned  `(never reset to 0)` but maybe since that was in a different context I might be wrong.

chappjc: Right, different context.  Before it was just about wrapping the counter when a modulo would do the job.  Here's it's about resetting the counter when it has been a long time since last fail.

ukane-philemon: #2235 is up to address a nit, do y'all want these changes there?

chappjc: > #2235 is up to address a nit, do y'all want these changes there?

Up to @buck54321, but that would be the time and place if it makes sense and tests well.

buck54321: Let's do it

ukane-philemon: Done.

PR #2034, opened 2023-01-06
eth: Always check header time.

### LINKED ISSUE #2017: geth: Sync status always %100.
Server will start the markets before chain is synced. Client just says %100 and will trade when it should no be able to yet.

JoeGruffins:
Post merge geth will report both the current and highest block as the same number. Check the header time even when we would have been synced before the merge.

closes #2017

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -2756,10 +2756,32 @@ func (*baseWallet) ValidateSecret(secret, secretHash []byte) bool {
 }
 
 // SyncStatus is information about the blockchain sync status.
+//
+// TODO: Since the merge, the sync status from a geth full node, namely the
+// prog.CurrentBlock prog.HighestBlock, always seem to be the same number.
+// Initial sync will always be zero. Later when restarting the node they move
+// together but never indicate the highest known block on the chain. Further
+// more, requesting the best block header starts to fail after a few tries
+// during initial sync. Investigate how to get correct sync progress.
 func (eth *baseWallet) SyncStatus() (bool, float32, error) {
 	prog, err := eth.node.syncProgress(eth.ctx)
 	if err != nil {
 		return false, 0, err
+	}
+	checkHeaderTime := func() (bool, error) {
+		bh, err := eth.node.bestHeader(eth.ctx)
+		if err != nil {
+			return false, err
+		}
+		// Time in the header is in seconds.
+		timeDiff := time.Now().Unix() - int64(bh.Time)
+		if timeDiff > dexeth.MaxBlockInterval && eth.net != dex.Simnet {
+			eth.log.Infof("Time since last eth block (%d sec) exceeds %d sec."+
+				"Assuming not in sync. Ensure your computer's system clock "+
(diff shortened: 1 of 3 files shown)


JoeGruffins:
Now sync status is always zero, so still not perfect. Not sure how to determine percent synced when talking to a geth node. The user can always look at their geth node however. Server is just wrong, Unix is already seconds.

JoeGruffins:
This page may have some insight into the process https://docs.ava.do/packages/geth/. There is probably a span during initial sync where the prog.CurrentBlock and prog.HighestBlock are different. After that, it looks like their dashboard will just say 99% until the headers are new. We could do that I suppose.

chappjc:
I don't quite get why sync progress is useless after merge, but I agree that as longs as we don't say synced when it's not, that's what's important.

Good catch on the server units.

client/asset/eth/eth.go
```diff
@@ -2756,17 +2756,45 @@ func (*baseWallet) ValidateSecret(secret, secretHash []byte) bool {
...
+	checkHeaderTime := func() (bool, error) {
+		bh, err := eth.node.bestHeader(eth.ctx)
+		if err != nil {
+			return false, err
+		}
+		// Time in the header is in seconds.
+		timeDiff := time.Now().Unix() - int64(bh.Time)
+		if timeDiff > dexeth.MaxBlockInterval && eth.net != dex.Simnet {
+			eth.log.Debugf("Time since last block (%d sec) exceeds %d sec."+
+				"Assuming not in sync.", timeDiff, dexeth.MaxBlockInterval)
```
chappjc, line 2780: Seems like our only option.  We may want to add a note to the message to check the system clock too, because that could wreck things.

chappjc (approved):

client/asset/eth/eth.go
JoeGruffins: Added a note to the log and made it info level. Should it be a note in the ui? https://github.com/decred/dcrdex/compare/7e5a0222d0c2847c90137004201e11f777013285..a417333106afbfe8320818c2dc719bcd8aca6308

chappjc: > Should it be a note in the ui?

Don't think so.  They already get an implied "just wait longer" by it saying false for synced.  We could clamp it at 99.9% too though, if the best header is old, which we do for BTC as of recently.

client/asset/eth/eth.go
```diff
@@ -2756,17 +2756,46 @@ func (*baseWallet) ValidateSecret(secret, secretHash []byte) bool {
 }
 
 // SyncStatus is information about the blockchain sync status.
+//
+// TODO: Since the merge, the sync status from a geth full node, namely the
+// prog.CurrentBlock prog.HighestBlock, always seem to be the same number.
+// Initial sync will always be zero. Later when restarting the node they move
+// together but never indicate the highest know block on the chain. Further
```
buck54321, line 2763: know**n**

buck54321 (approved):

client/asset/eth/eth.go
JoeGruffins: Corrected.

Issue #2037, opened 2023-01-10
testing: simnet-trade-tests failing at fee payment


buck54321:
```
2023-01-10 11:17:08.097 [INF] CORE:1: notify: |SUCCESS| (feepayment) Fee payment in progress - Waiting for 1 confirmations before trading at 127.0.0.1:17273
2023-01-10 11:17:08.097 [INF] T[1]: Sent registration fee to DEX 127.0.0.1:17273
2023-01-10 11:17:08.097 [INF] T[1]: Mined 1 dcr blocks for fee payment confirmation
2023-01-10 11:17:08.097 [INF] T[1]: Waiting 1m12s for fee confirmation notice
ERROR: setup error: error starting clients: fee payment not confirmed after 1m12s
```

JoeGruffins:
`./run dcrbtc --all` will pass on master for me.

Are you sure this is not a temporary problem with the dcr harness? From the beginning, sometimes that thing just gets stuck. It's very random and I am unable to reproduce on command.

chappjc:
I think this issue can be closed.

About `PATH`, that's the case for all the node daemons too.

> In the reverse case some livetest test cases mine on demand but may sometimes get an extra block mined by the watch miner -- invalidating the expected number of confirmations, for example.
> 
> I usually switch off the watch miner in such a case

Yes, I always kill the miners before running the simnet trade tests.

PR #2038, opened 2023-01-10
eth: v1 contract

buck54321:
Replaces #1426. Not tested with simnet-trade-tests (#2037) or loadbot, so leaving in draft for now. Need to get some reviews done.

```
Implements the version 1 contracts for ethereum and tokens. Based
on feedback in #1426, everything is now encoded in the
"contract data". This "contract data", is the msgjson.Init.Contract
-> msgjson.Audit.Contract -> MatchMetaData.Proof.CounterContract,
AuditInfo.Contract -> Redemption.Spends.Contract.

A few new terms are introduced to differentiate various encodings and
data sets. The aforementioned contract data did encode a version
and a secret hash. It now encodes a version and a "locator", which is
a []byte whose length and content depend on the version. For
version 0, the locator is still just the secretHash[:]. For v1,
the locator encodes all of the immutable data that defines the
swap. This immutable data is now collected in something called
a "vector" (dexeth.SwapVector). For version 0, some vector data
is stored on-chain indexed by the secret hash. For version 1, all
vector data is encoded in the locator.

I've also made an effort to standardize the use of status/step,
and eliminated the use of ambiguous "ver" variables throughout.
A "status" is now the collection of mutable contract data: the step,
the init block height, and the secret. The status and vector
collectively fully characterize the swap state.

client/asset/eth:
New contractV1 and tokenContractorV1 interfaces. To avoid duplication,
the ERC20 parts of the tokenContractors are separated into a new type
erc20Contractor that is embedded by both versions. Getters for
status and vector are added in place of the old method "swap".

assetWallet and embedding types are updated to work with the new
version-dependent locators and the status and vector model.

dex/networks/{eth,erc20}:
New contracts added. New methods for dealing with locators. Simnet
entries added for eth and dextt.eth in the ContractAddresses and Tokens
maps. txDataHandler interace is replaced with versioned package-level
functions.

server/asset/eth:
Server is fully switched to version 1. No option to use version 0.
Translation to new version was straightforward, with one notable
difference that we can no longer get a block height from the
contract once the swap is redeemed.```

### CODE REVIEW

File: client/asset/eth/eth.go [hunk 1]
```diff
@@ -2689,17 +2794,35 @@ func (w *assetWallet) SwapConfirmations(ctx context.Context, coinID dex.Bytes, c
...
 	}
 
-	spent = swapData.State >= dexeth.SSRedeemed
-	confs = uint32(hdr.Number.Uint64() - swapData.BlockHeight + 1)
+	spent = status.Step >= dexeth.SSRedeemed
+	confs = uint32(hdr.Number.Uint64() - status.BlockHeight + 1)
+
+	// NOTE: confs will equal to the block number for the version 1 contract if spent == true,
+	// because BlockHeight will be zero. This is probably fine, since the caller
+	// will examine spent first? Otherwise, we could do this.
```
chappjc, line 2811: Interesting problem to have here, but I think it's right to be using what's recorded in the contract state like you're doing in this version of the PR, at least before it's spent.  What you suggest about getting the tx confirmations after it is spent might be alright unless the tx was replaced.

client/asset/eth/eth.go
```diff
@@ -2689,17 +2794,35 @@ func (w *assetWallet) SwapConfirmations(ctx context.Context, coinID dex.Bytes, c
...
+	// 	var txHash common.Hash
+	// 	copy(txHash[:], coinID)
+	// 	confs, err = w.node.transactionConfirmations(ctx, txHash)
+	// 	if err != nil {
+	// 		return 0, false, fmt.Errorf("error finding swap state: %w", err)
+	// 	}
+	// }
+	//
+	// Or maybe it'd be better to set confs to some constant, large, round
+	// number so it doesn't look dumb in logs.
```
chappjc, line 2825: Yah, maybe like 32 or something?

martonp:
There is a new ETH functionality that would allow us to have ETH redemptions for accounts without any ETH:
https://eips.ethereum.org/EIPS/eip-4337

Would be cool to have it in the v1 contract.

norwnd:
> There is a new ETH functionality that would allow us to have ETH redemptions for accounts without any ETH:
https://eips.ethereum.org/EIPS/eip-4337

It might also allow for removal of additional "approve" step since multiple user actions can be executed within single transaction (from https://youtu.be/iLf8qpOmxQc?t=2028).

But who knows how long it will take to get this Account Abstraction design working, I bet it's 1+ year down the road.

martonp (commented):
Swaps seem to be working fine, but I'm getting some errors on the server:

Screenshot 2024-03-13 at 8 19 23 PM

JoeGruffins (commented):
Glad you rebased all of this. The gas savings on everything but Refund is substantial. I think I have a valid comment about the RefundRecord being used as the Secret and maybe some unintended things can happen with that.

dex/networks/eth/contracts/ETHSwapV1.sol
```diff
@@ -0,0 +1,231 @@
...
+            // Can't already be redeemed.
+            require(!secretValidates(record, r.v.secretHash), "already redeemed");
+
+            // Are they presenting the correct secret?
+            require(secretValidates(r.secret, r.v.secretHash), "invalid secret");
+
+            swaps[k] = r.secret;
+
+
+            // DRAFT TODO: NOOOOOO! This doesn't account for decimals. This is
```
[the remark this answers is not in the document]
buck54321: This `TODO` should have been deleted. I was initially passing in value as a `uint64` and multiplying by `1 gwei` internally. That works for eth, but not for some tokens, so I switched to `uint256` (for both contracts to maintain API compatibility). This does incur a slight bump in gas because there's more call data blocks, but they're not stored so it's not a lot.

client/asset/eth/eth.go
```diff
@@ -3324,41 +3417,80 @@ func (eth *baseWallet) swapOrRedemptionFeesPaid(ctx context.Context, coinID, con
...
+	case 0:
+		if isInit {
+			inits, err := dexeth.ParseInitiateDataV0(tx.Data())
+			if err != nil {
+				return nil, nil, fmt.Errorf("invalid initiate data: %v", err)
+			}
+			locators = make([][]byte, 0, len(inits))
+			for k := range inits {
+				copyK := k // TODO: Is this really necessary?
+				locators = append(locators, copyK[:])
```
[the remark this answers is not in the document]
buck54321: You're playground test makes sense, but for some reason it's not working the same in our code and I do need the copy. Confusing.

dex/networks/eth/contracts/ETHSwapV1.sol
```diff
@@ -0,0 +1,231 @@
...
+        uint256 value;
+        address initiator;
+        uint64 refundTimestamp;
+        address participant;
+    }
+
+    // contractKey generates a key hash which commits to the contract data. The
+    // generated hash is used as a key in the swaps map.
+    function contractKey(Vector calldata v) public pure returns (bytes32) {
+        return sha256(bytes.concat(v.secretHash, bytes20(v.initiator), bytes20(v.participant), bytes32(v.value), bytes8(v.refundTimestamp)));
```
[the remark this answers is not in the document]
buck54321: I was originally using `abi.encodePacked`, but we decided to use `bytes.encode` because of some ambiguity in the implementation of abi encoding, iirc. I can't find the convo. I think it was on Matrix.

server/asset/eth/coiner.go
```diff
@@ -55,30 +55,72 @@ func (be *AssetBackend) newSwapCoin(coinID []byte, contractData []byte) (*swapCo
...
+
+		vector = &dexeth.SwapVector{
+			// From: ,
+			To:         init.Participant,
+			Value:      init.Value,
+			SecretHash: secretHash,
+			LockTime:   uint64(init.LockTime.Unix()),
+		}
+
+		// if value < bc.vector.Value {
```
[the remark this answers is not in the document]
buck54321: The value check is moved and the others are completely pointless since we just pulled (v0) or validated (v1) the data on-chain. This is also not the purview of the backend, which reports chain data to the swapper for validation.

client/asset/eth/contractor.go
```diff
@@ -164,7 +190,73 @@ func (c *contractorV0) redeem(txOpts *bind.TransactOpts, redemptions []*asset.Re
...
+	if err != nil {
+		return nil, err
+	}
+	swap, err := c.swap(ctx, secretHash)
+	if err != nil {
+		return nil, err
+	}
+	vector := &dexeth.SwapVector{
+		From:       swap.Participant,
+		To:         swap.Initiator,
```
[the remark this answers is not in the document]
buck54321: Whoops.

client/asset/eth/contractor.go
```diff
@@ -370,79 +525,401 @@ func newV0TokenContractor(net dex.Network, token *dexeth.Token, acctAddr common.
...
-func (c *tokenContractorV0) estimateGas(ctx context.Context, method string, args ...any) (uint64, error) {
-	data, err := erc20.ERC20ABI.Pack(method, args...)
+// func (c *contractorV1) record(v *dexeth.SwapVector) (r [32]byte, err error) {
+// 	abiVec := dexeth.SwapVectorToAbigen(v)
+// 	ck, err := c.ContractKey(&bind.CallOpts{From: c.acctAddr}, abiVec)
+// 	if err != nil {
+// 		return r, fmt.Errorf("ContractKey error: %v", err)
+// 	}
+// 	return c.Swaps(&bind.CallOpts{From: c.acctAddr}, ck)
+// }
```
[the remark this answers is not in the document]
buck54321: I though this might come in handy in an emergency for troubleshooting. The underlying `Swaps` method is generated automatically by abigen anyway.

dex/networks/eth/contracts/ETHSwapV1.sol
```diff
@@ -0,0 +1,231 @@
...
+    // status returns the current state of the swap.
+    function status(Vector calldata v)
+        public view returns(Status memory)
+    {
+        (, bytes32 record, uint256 blockNum) = retrieveStatus(v);
+        Status memory r;
+        if (blockNum == 0) {
+            r.step = Step.Empty;
+        } else if (record == RefundRecord) {
+            r.step = Step.Refunded;
```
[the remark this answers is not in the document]
buck54321: As per off-github convo, went with a check in `initiate` for the refund record hash, as well as a check server-side to prevent the maker choosing the refund record as the secret.

server/asset/eth/coiner.go
```diff
@@ -55,30 +55,72 @@ func (be *AssetBackend) newSwapCoin(coinID []byte, contractData []byte) (*swapCo
...
 		for _, in := range inits {
-			sum.Add(sum, in.Value)
+			sum = +be.atomize(in.Value)
 		}
-		if bc.value.Cmp(sum) < 0 {
-			return nil, fmt.Errorf("tx %s value < sum of inits. %d < %d", bc.txHash, bc.value, sum)
+
+		vector = &dexeth.SwapVector{
+			// From: ,
+			To:         init.Participant,
```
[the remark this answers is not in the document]
buck54321: We simply don't have this information at this point, and luckily the server doesn't need it. I wasn't sure what else to do but leave it empty.

server/cmd/dcrdex/evm-protocol-overrides.json
```diff
@@ -0,0 +1,8 @@
+{
+    "eth": 0,
+    "usdc.eth": 0,
+    "polygon": 0,
+    "usdc.polygon": 0,
+    "wbtc.polygon": 0,
+    "weth.polygon": 0
```
buck54321, line 7: This is annoying, but until we launch contracts, you'll have to delete these overrides to test v1 contracts.

dex/networks/eth/contracts/ETHSwapV1.sol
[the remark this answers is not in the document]
buck54321: I still need to check out keccak

JoeGruffins (approved):

JoeGruffins:
Is it possible to add tests for the contracts like @martonp has done in #2717 ?

buck54321:
All v1 contracts launched and gases validated on mainnet, and contracts launched for all available tokens on testnet.

dex/networks/eth/contracts/ETHSwapV1.sol
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    bytes32 constant RefundRecord = 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF;
+    bytes32 constant RefundRecordHash = 0xAF9613760F72635FBDB44A5A0A63C39F12AF30F950A6EE5C971BE188E89C4051;
+
+    // swaps is a map of contract hashes to the "swap record". The swap record
+    // has the following interpretation.
+    //   if (record == bytes32(0x00)): contract is uninitiated
+    //   else if (uint256(record) < block.number && sha256(record) != contract.secretHash):
+    //      contract is initiated and redeemable by the participant with the secret.
```
martonp, line 42: Could we use 20 bytes here for the the token address and the remaining 12 for the block number? Then we wouldn't need to create a new contract for each token.

buck54321:
Switched up to using a single contract for eth and tokens. Non-live unit tests are passing. Getgas utility is working. But not tested beyond that. This should be close to the final form though.

dex/networks/eth/contracts/ETHSwapV1.sol
```diff
@@ -0,0 +1,261 @@
...
+
+    // Redemption is the information necessary to redeem a Vector. Since we
+    // don't store the Vector itself, it must be provided as part of the
+    // redemption.
+    struct Redemption {
+        Vector v;
+        bytes32 secret;
+    }
+
+    function secretValidates(bytes32 secret, bytes32 secretHash) public pure returns (bool) {
```
dev-warrior777, line 85: nit: I would prefer `canSecreteBeValidated` or even `isSecretValidated` for readabilty

dev-warrior777 (approved):

dex/networks/eth/contracts/ETHSwapV1.sol
```diff
@@ -0,0 +1,261 @@
...
+        payable
+        senderIsOrigin()
+    {
+        uint initVal = 0;
+        for (uint i = 0; i < contracts.length; i++) {
+            Vector calldata v = contracts[i];
+
+            require(v.value > 0, "0 val");
+            require(v.refundTimestamp > 0, "0 refundTimestamp");
+            require(v.secretHash != RefundRecordHash, "illegal secret hash (refund record hash)");
```
martonp, line 142: 0 val and refund hash can be checked by client. They can't result in the initiator cheating.

dex/networks/eth/contracts/ETHSwapV1.sol
```diff
@@ -0,0 +1,261 @@
...
+
+            bytes32 k = contractKey(token, v);
+            bytes32 record = swaps[k];
+            require(record == bytes32(0), "swap not empty");
+
+            record = bytes32(block.number);
+            require(!secretValidates(record, v.secretHash), "hash collision");
```
martonp, line 149: This check also doesn't prevent misbehavior. Can be on the client.

dex/networks/eth/contracts/ETHSwapV1.sol
```diff
@@ -0,0 +1,261 @@
...
+
+    // swaps is a map of contract hashes to the "swap record". The swap record
+    // has the following interpretation.
+    //   if (record == bytes32(0x00)): contract is uninitiated
+    //   else if (uint256(record) < block.number && sha256(record) != contract.secretHash):
+    //      contract is initiated and redeemable by the participant with the secret.
+    //   else if (sha256(record) == contract.secretHash): contract has been redeemed
+    //   else if (record == RefundRecord): contract has been refunded
+    //   else: invalid record. Should be impossible by construction
+    mapping(bytes32 => bytes32) public swaps;
```
martonp, line 48: Why don't we just make this a 
```
mapping(bytes32 => bool) public swaps;
```

True if the swap is initiated. Then when redeeming / refunding just delete it. For keeping records of which block swaps are redeemed / refunded, we can use event logs.

ukane-philemon (changes requested):
nits...will continue reviewing later.

ukane-philemon:
Overall, the new concepts are easy to understand.

client/asset/eth/eth.go
```diff
@@ -5949,17 +6067,23 @@ func getGasEstimates(ctx context.Context, cl, acl ethFetcher, c contractor, ac t
 		fmt.Printf("    %+v \n", stats.transfers)
 	}()
 
+	logTx := func(tag string, n int, tx *types.Transaction) {
+		log.Infof("%s %d tx, hash = %s, nonce = %d, maxFeeRate = %s, tip cap = %s",
+			tag, n, tx.Hash(), tx.Nonce(), tx.GasFeeCap(), tx.GasTipCap())
+	}
```
JoeGruffins, line 6073: Maybe better as debug or trace?

client/core/trade.go
```diff
@@ -2295,36 +2295,75 @@ func (c *Core) swapMatches(t *trackedTrade, matches []*matchTracker) error {
...
+		freshRate := t.wallets.fromWallet.feeRate()
+		if freshRate == 0 { // either not a FeeRater, or FeeRate failed
+			freshRate = t.dc.bestBookFeeSuggestion(t.wallets.fromWallet.AssetID)
+		}
+		if freshRate > t.metaData.MaxFeeRate {
+			freshRate = t.metaData.MaxFeeRate
+		}
+		if highestFeeRate < freshRate {
+			t.dc.log.Infof("Prescribed %v fee rate %v looks low, using %v",
+				t.wallets.fromWallet.Symbol, highestFeeRate, freshRate)
```
JoeGruffins, line 2344: Maybe debug is better?

JoeGruffins (commented):
Sorry going through again and taking a bit, this is halfway.

buck54321:
Hey all. What I need from this PR right now is to make sure that I haven't missed something dumb that will break v1.0 branch clients that are still using the v0 contracts. Once I have that, I'll merge, and we can all open our own PRs to address outstanding questions.

JoeGruffins:
Ok, I will be able to do better testing after the geth updates are in and this is rebased.

martonp (approved):
Testing will with a v0 client.

JoeGruffins (approved):
Testing fine with v0 server and trading there. A couple of my newer comments are valid I think so should update.

Issue #1947, opened 2022-11-08
UI: Maker Swap - display expected Refund time


norwnd:
When this happens:

![image](https://user-images.githubusercontent.com/112318969/200620148-5e18cd04-3f8e-4987-880a-a9cff375fde5.png)

user might get uneasy, starting at this message for hours (might even contact support asking for help):
```
Refund (eth, you)
<Pending>
```

The info about **when refund will happen** would be useful to show:

```
Refund (eth, you) will happen after: ~%s
<Pending>
```

I guess `%s` should reflect the value that can be seen in this log:
```
2022-11-08 12:17:01.277 [INF] CORE: Contract coin 0x2ef2c3bab324b45da3a2be2fba46890f68ac91318250f5eddfe2e0b74118fbbf (eth), value = 4000, refundable at 2022-11-09 11:16:57 +0300 MSK (receipt = { tx hash: 0x2ef2c3bab324b45da3a2be2fba46890f68ac91318250f5eddfe2e0b74118fbbf, contract address: 0xa483b6166dA8Da6748B29Af35f96C4F9388c456C, secret hash: ac4587cb9eb12f99f0ce0b5ae8c7b969c64aa680da4f03d85ffb1297dae5355c }), match = dbfb6b3b5f295319a890c6967ecb70e28b20b56baee9e3a636caa3c475591016
```

Could similar happen to `taker` ? when he also needs to refund.

norwnd:
Maybe there is no need to use exact `receipt.Expiration()` value (that will require adding it to `Match`, webserver API, and storing it in DB probably) and just take `Match.stamp` + `20h` as rough estimate ? Even if this `20h` value changes in the future on the back-end, and we forget to adjust it in UI - that won't be a big deal I think.

chappjc:
> Maybe there is no need to use exact `receipt.Expiration()` value (that will require adding it to `Match`, webserver API, and storing it in DB probably) and just take `Match.stamp` + `20h` as rough estimate ? Even if this `20h` value changes in the future on the back-end, and we forget to adjust it in UI - that won't be a big deal I think.

[...]

PR #2042, opened 2023-01-12
ui: add expected refund time to match-card

norwnd:
Resolves https://github.com/decred/dcrdex/issues/1947.

With changes in this PR (note, these screenshots were indeed made in November 2022, just waited for https://github.com/decred/dcrdex/pull/1978 to go in first; I didn't re-test it myself after rebase, but you'll probably be doing your own testing and I'm optimistic that it still works fine):

imageimageimage

### DIFF
--- a/client/webserver/site/src/js/locales.ts
+++ b/client/webserver/site/src/js/locales.ts
@@ -109,6 +109,8 @@ export const ID_SELECT_WALLET_FOR_FEE_PAYMENT = 'SELECT_WALLET_FOR_FEE_PAYMENT'
 export const ID_UNAVAILABLE = 'UNAVAILABLE'
 export const ID_WALLET_SYNC_FINISHING_UP = 'WALLET_SYNC_FINISHING_UP'
 export const ID_CONNECT_WALLET_ERR_MSG = 'CONNECTING_WALLET_ERR_MSG'
+export const ID_REFUND_IMMINENT = 'REFUND_IMMINENT'
+export const ID_REFUND_WILL_HAPPEN_AFTER = 'REFUND_WILL_HAPPEN_AFTER'
 
 export const enUS: Locale = {
   [ID_NO_PASS_ERROR_MSG]: 'password cannot be empty',
@@ -219,7 +221,10 @@ export const enUS: Locale = {
   [ID_EMPTY_DEX_ADDRESS_MSG]: 'DEX address cannot be empty',
   [ID_SELECT_WALLET_FOR_FEE_PAYMENT]: 'Select a valid wallet to post a bond',
   [ID_WALLET_SYNC_FINISHING_UP]: 'finishing up',
-  [ID_CONNECT_WALLET_ERR_MSG]: 'Failed to connect {{ assetName }} wallet: {{ errMsg }}'
+  [ID_CONNECT_WALLET_ERR_MSG]: 'Failed to connect {{ assetName }} wallet: {{ errMsg }}',
+  [ID_TAKER_FOUND_MAKER_REDEMPTION]: 'Redeemed by {{ makerAddr }}',
+  [ID_REFUND_IMMINENT]: 'Will happen in the next few blocks',
+  [ID_REFUND_WILL_HAPPEN_AFTER]: 'Refund will happen after {{ refundAfterTime }}'
 }
 
 export const ptBR: Locale = {
--- a/client/webserver/site/src/js/order.ts
+++ b/client/webserver/site/src/js/order.ts
@@ -19,6 +19,11 @@ const Mainnet = 0
 const Testnet = 1
 // const Regtest = 3
 
+// lockTimeMakerMs must match the value returned from LockTimeMaker func in dexc.
+const lockTimeMakerMs = 20 * 60 * 60 * 1000
+// lockTimeTakerMs must match the value returned from LockTimeTaker func in dexc.
+const lockTimeTakerMs = 8 * 60 * 60 * 1000
+
 const coinIDTakerFoundMakerRedemption = 'TakerFoundMakerRedemption:'
 
 const animationLength = 500
@@ -155,7 +160,7 @@ export default class OrderPage extends BasePage {
     tmpl.matchID.textContent = match.matchID
 
     const time = new Date(match.stamp)
-    tmpl.matchTime.textContent = time.toLocaleTimeString('en-GB', {
+    tmpl.matchTime.textContent = time.toLocaleTimeString(navigator.languages as string[], {
       year: 'numeric',
       month: 'short',
       day: 'numeric'
@@ -260,40 +265,59 @@ export default class OrderPage extends BasePage {
    * updated on each update to the match.
    */
   setMutableMatchCardElements (matchCard: HTMLElement, m: Match) {
-    if (m.isCancel) {
-      return
-    }
+    if (m.isCancel) return
 
     const tmpl = Doc.parseTemplate(matchCard)
     tmpl.status.textContent = OrderUtil.matchStatusString(m)
 
-    const setCoin = (pendingName: string, linkName: string, coin: Coin) => {
-      const formatCoinID = (cid: string) => {
-        if (cid.startsWith(coinIDTakerFoundMakerRedemption)) {
-          const makerAddr = cid.substring(coinIDTakerFoundMakerRedemption.length)
-          return intl.prep(intl.ID_TAKER_FOUND_MAKER_REDEMPTION, { makerAddr: makerAddr })
-        }
-        return cid
+    const formatCoinID = (cid: string) => {
+      if (cid.startsWith(coinIDTakerFoundMakerRedemption)) {
+        const makerAddr = cid.substring(coinIDTakerFoundMakerRedemption.length)
+        return intl.prep(intl.ID_TAKER_FOUND_MAKER_REDEMPTION, { makerAddr: makerAddr })
       }
-      const coinLink = tmpl[linkName]
-      const pendingSpan = tmpl[pendingName]
+      return cid
+    }
+
+    const tryShowCoin = (pendingEl: PageElement, coinLink: PageElement, coin: Coin) => {
       if (!coin) {
-        Doc.show(tmpl[pendingName])
-        Doc.hide(tmpl[linkName])
+        Doc.hide(coinLink)
+        Doc.show(pendingEl)
         return
       }
       coinLink.textContent = formatCoinID(coin.stringID)
       coinLink.dataset.explorerCoin = coin.stringID
       setCoinHref(coin.assetID, coinLink)
-      Doc.hide(pendingSpan)
       Doc.show(coinLink)
+      Doc.hide(pendingEl)
     }
 
-    setCoin('makerSwapPending', 'makerSwapCoin', makerSwapCoin(m))
-    setCoin('takerSwapPending', 'takerSwapCoin', takerSwapCoin(m))
-    setCoin('makerRedeemPending', 'makerRedeemCoin', makerRedeemCoin(m))
-    setCoin('takerRedeemPending', 'takerRedeemCoin', takerRedeemCoin(m))
-    setCoin('refundPending', 'refundCoin', m.refund)
+    tryShowCoin(tmpl.makerSwapPending, tmpl.makerSwapCoin, makerSwapCoin(m))
+    tryShowCoin(tmpl.takerSwapPending, tmpl.takerSwapCoin, takerSwapCoin(m))
+    tryShowCoin(tmpl.makerRedeemPending, tmpl.makerRedeemCoin, makerRedeemCoin(m))
+    tryShowCoin(tmpl.takerRedeemPending, tmpl.takerRedeemCoin, takerRedeemCoin(m))
+    if (!m.refund) {
+      // Special messaging for pending refunds.
+      let lockTime = lockTimeMakerMs
+      if (m.side === OrderUtil.Taker) lockTime = lockTimeTakerMs
+      const refundAfter = new Date(m.stamp + lockTime)
+      if (Date.now() > refundAfter.getTime()) tmpl.refundPending.textContent = intl.prep(intl.ID_REFUND_IMMINENT)
+      else {
+        const refundAfterStr = refundAfter.toLocaleTimeString(navigator.languages as string[], {
+          year: 'numeric',
+          month: 'short',
+          day: 'numeric'
+        })
+        tmpl.refundPending.textContent = intl.prep(intl.ID_REFUND_WILL_HAPPEN_AFTER, { refundAfterTime: refundAfterStr })
+      }
+      Doc.hide(tmpl.refundCoin)
+      Doc.show(tmpl.refundPending)
+    } else {
+      tmpl.refundCoin.textContent = formatCoinID(m.refund.stringID)
+      tmpl.refundCoin.dataset.explorerCoin = m.refund.stringID
+      setCoinHref(m.refund.assetID, tmpl.refundCoin)
+      Doc.show(tmpl.refundCoin)
+      Doc.hide(tmpl.refundPending)
+    }
 
     if (m.status === OrderUtil.MakerSwapCast && !m.revoked && !m.refund) {
       const c = makerSwapCoin(m)
@@ -317,14 +341,64 @@ export default class OrderPage extends BasePage {
       Doc.hide(tmpl.makerSwapMsg, tmpl.takerSwapMsg, tmpl.makerRedeemMsg, tmpl.takerRedeemMsg)
     }
 
-    Doc.setVis(!m.isCancel && (makerSwapCoin(m) || !m.revoked), tmpl.makerSwap)
-    Doc.setVis(!m.isCancel && (takerSwapCoin(m) || !m.revoked), tmpl.takerSwap)
-    Doc.setVis(!m.isCancel && (makerRedeemCoin(m) || !m.revoked), tmpl.makerRedeem)
-    // When revoked, there is uncertainty about the taker redeem coin. The taker
-    // redeem may be needed if maker redeems while taker is waiting to refund.
-    Doc.setVis(!m.isCancel && (takerRedeemCoin(m) || (!m.revoked && m.active) || ((m.side === OrderUtil.Taker) && m.active && (m.counterRedeem || !m.refund))), tmpl.takerRedeem)
-    // The refund placeholder should not be shown if there is a counter redeem.
-    Doc.setVis(!m.isCancel && (m.refund || (m.revoked && m.active && !m.counterRedeem)), tmpl.refund)
+    if (!m.revoked) {
+      // Match is still following the usual success-path, it is desirable for the
+      // user to see it in full (even if to learn how atomic swap is supposed to
+      // work).
+
+      Doc.setVis(makerSwapCoin(m) || m.active, tmpl.makerSwap)
+      Doc.setVis(takerSwapCoin(m) || m.active, tmpl.takerSwap)
+      Doc.setVis(makerRedeemCoin(m) || m.active, tmpl.makerRedeem)
+      // When maker isn't aware of taker redeem coin, once the match becomes inactive
+      // (nothing else maker is expected to do in this match) just hide taker redeem.
+      Doc.setVis(takerRedeemCoin(m) || m.active, tmpl.takerRedeem)
+      // Refunding isn't a usual part of success-path, but don't rule it out.
+      Doc.setVis(m.refund, tmpl.refund)
+    } else {
+      // Match diverged from the usual success-path, since this could have happened
+      // at any step it is hard (maybe impossible) to predict the final state this
+      // match will end up in, so show only steps that already happened plus all
+      // the possibilities on the next step ahead.
+
+      // If we don't have swap coins after revocation, we won't show the pending message.
+      Doc.setVis(makerSwapCoin(m), tmpl.makerSwap)
+      Doc.setVis(takerSwapCoin(m), tmpl.takerSwap)
+      const takerRefundsAfter = new Date(m.stamp + lockTimeTakerMs)
+      const takerLockTimeExpired = Date.now() > takerRefundsAfter.getTime()
+      // When match is revoked and both swaps are present, maker redeem might still show up:
+      // - as maker, we'll try to redeem until taker locktime expires (if taker refunds
+      //   we won't be able to redeem; even if taker hasn't refunded just yet - it
+      //   becomes too dangerous to redeem after taker locktime expired because maker
+      //   reveals his secret when redeeming, and taker might be able to submit both
+      //   redeem and refund transactions before maker's redeem gets mined), so we'll
+      //   have to show redeem pending element until maker redeem shows up, or until
+      //   we give up on redeeming due to taker locktime expiry.
+      // - as taker, we should expect maker redeeming any time, so we'll have to show
+      //   redeem pending element until maker redeem shows up, or until we refund.
+      Doc.setVis(makerRedeemCoin(m) || (takerSwapCoin(m) && m.active && !m.refund && !takerLockTimeExpired), tmpl.makerRedeem)
+      // When maker isn't aware of taker redeem coin, once the match becomes inactive
+      // (nothing else maker is expected to do in this match) just hide taker redeem.
+      Doc.setVis(takerRedeemCoin(m) || (makerRedeemCoin(m) && m.active && !m.refund), tmpl.takerRedeem)
+      // As taker, show refund placeholder only if we have outstanding swap to refund.
+      // There is no need to wait for anything else, we can show refund placeholder
+      // (to inform the user that it is likely to happen) right after match revocation.
+      let expectingRefund = Boolean(takerSwapCoin(m)) // as taker
+      if (m.side === OrderUtil.Maker) {
+        // As maker, show refund placeholder only if we have outstanding swap to refund.
+        // If we don't have taker swap there is no need to wait for anything else, we
+        // can show refund placeholder (to inform the user that it is likely to happen)
+        // right after match revocation.
+        expectingRefund = Boolean(makerSwapCoin(m))
+        // If we discover taker swap we'll be trying to redeem it (instead of trying
+        // to refund our own swap) until taker refunds, so start showing refund
+        // placeholder only after taker is expected to start his refund process in
+        // this case.
+        if (takerSwapCoin(m)) {
+          expectingRefund = expectingRefund && takerLockTimeExpired
+        }
+      }
+      Doc.setVis(m.refund || (m.active && !m.redeem && !m.counterRedeem && expectingRefund), tmpl.refund)
+    }
   }
 
   /*
@@ -458,7 +532,7 @@ export default class OrderPage extends BasePage {
 
 /*
  * confirmationString is a string describing the state of confirmations for a
- * coin
+ * coin.
  * */
 function confirmationString (coin: Coin) {
   if (!coin.confs || coin.confs.required === 0) return ''
--- a/client/db/types.go
+++ b/client/db/types.go
@@ -491,12 +491,12 @@ func (m *MetaMatch) MatchOrderUniqueID() []byte {
 
 // MatchIsActive returns false (i.e. the match is inactive) if any: (1) status
 // is MatchConfirmed OR InitSig unset, signaling a cancel order match, which is
-// never active), (2) the match is refunded, or (3) it is revoked and this side
+// never active, (2) the match is refunded, or (3) it is revoked and this side
 // of the match requires no further action like refund or auto-redeem.
 func MatchIsActive(match *order.UserMatch, proof *MatchProof) bool {
 	// MatchComplete only means inactive if: (a) cancel order match or (b) the
 	// redeem request was accepted for trade orders. A cancel order match starts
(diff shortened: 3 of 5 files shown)


client/webserver/site/src/js/order.ts
```diff
@@ -258,33 +263,86 @@ export default class OrderPage extends BasePage {
...
+          const refundAfterStr = refundAfter.toLocaleTimeString('en-GB', {
+            year: 'numeric',
+            month: 'short',
+            day: 'numeric'
+          })
+          tmpl.refundPending.textContent = intl.prep(intl.ID_REFUND_WILL_HAPPEN_AFTER, { refundAfterTime: refundAfterStr })
+        }
+
+        Doc.show(tmpl.refundPending)
+      },
```
norwnd, line 339: `setCoin` func (renamed to `tryShowCoin` now) refactoring in this PR is mostly to accommodate this change here ^

client/webserver/site/src/js/locales.ts
```diff
@@ -189,7 +191,9 @@ export const enUS: Locale = {
   [ID_MATCH_STATUS_REDEEM_PENDING]: 'Redeem PENDING',
   [ID_MATCH_STATUS_REDEMPTION_CONFIRMED]: 'Redemption Confirmed',
   [ID_MATCH_STATUS_COMPLETE]: 'Complete',
-  [ID_TAKER_FOUND_MAKER_REDEMPTION]: 'redeemed by {{ makerAddr }}'
+  [ID_TAKER_FOUND_MAKER_REDEMPTION]: 'redeemed by {{ makerAddr }}',
+  [ID_REFUND_COMING_YOUR_WAY]: 'Coming your way, just a little patience',
```
chappjc, line 195: Fine in English but a direct translation will likely be bonkers.  Maybe add a trailing comment like `// alt: Will happen in the next few blocks `

client/webserver/site/src/js/locales.ts
```diff
@@ -94,6 +94,8 @@ export const ID_MATCH_STATUS_REFUND_PENDING = 'MATCH_STATUS_REFUND_PENDING'
 export const ID_MATCH_STATUS_REDEEM_PENDING = 'MATCH_STATUS_REDEEM_PENDING'
 export const ID_MATCH_STATUS_COMPLETE = 'MATCH_STATUS_COMPLETE'
 export const ID_TAKER_FOUND_MAKER_REDEMPTION = 'TAKER_FOUND_MAKER_REDEMPTION'
+export const ID_REFUND_COMING_YOUR_WAY = 'REFUND_COMING_YOUR_WAY'
```
chappjc, line 97: Let's go with `REFUND_IMMINENT` so in the future it's clear that this is the stage where lockTime is passed and it's gonna happen any minute.  It's a bit colloquial for the code IMO.  The other one (`REFUND_WILL_HAPPEN_AFTER`) is clearly during the lockTime countdown.

chappjc (changes requested):
Looks nice.  The one message is potentially too colloquial, so at least for the code I have some suggestions, although I'm tentatively OK with the user-facing message since I think most native English speakers will get the message.

client/webserver/site/src/js/locales.ts
norwnd: Sure, applied.

client/webserver/site/src/js/locales.ts
norwnd: Totally open to rephrase it however you like, added `alt` comment for now.

JoeGruffins:
This is not new, but since you are here, can you remove the "Taker Redemption" line here? If we are pass the taker refund time that is no longer pending and we are trying to refund. Also, the three becomes a four here. Not sure if this is in master but probably is.
image

client/webserver/site/src/js/locales.ts
```diff
@@ -189,7 +191,9 @@ export const enUS: Locale = {
   [ID_MATCH_STATUS_REDEEM_PENDING]: 'Redeem PENDING',
   [ID_MATCH_STATUS_REDEMPTION_CONFIRMED]: 'Redemption Confirmed',
   [ID_MATCH_STATUS_COMPLETE]: 'Complete',
-  [ID_TAKER_FOUND_MAKER_REDEMPTION]: 'redeemed by {{ makerAddr }}'
+  [ID_TAKER_FOUND_MAKER_REDEMPTION]: 'redeemed by {{ makerAddr }}',
+  [ID_REFUND_IMMINENT]: 'Coming your way, just a little patience', // alt: Will happen in the next few blocks
+  [ID_REFUND_WILL_HAPPEN_AFTER]: 'Will happen after {{ refundAfterTime }}'
```
JoeGruffins, line 196: imo:
(Capitalization)
redeemed by -> Redeemed by
(More professional)
Coming your way, just a little patience -> Will happen in the next few blocks
(More easily understandable with the noun)
Will happen after -> Refund will happen after

client/webserver/site/src/js/order.ts
```diff
@@ -310,10 +368,13 @@ export default class OrderPage extends BasePage {
 
     Doc.setVis(!m.isCancel && (makerSwapCoin(m) || !m.revoked), tmpl.makerSwap)
     Doc.setVis(!m.isCancel && (takerSwapCoin(m) || !m.revoked), tmpl.takerSwap)
-    Doc.setVis(!m.isCancel && (makerRedeemCoin(m) || !m.revoked), tmpl.makerRedeem)
+    // When revoked, there is no need to show maker redeem if we are maker (since
+    // it won't happen), but if we are taker, maker redeem might still show up
+    // if maker "doesn't play by server rules" (so we have to account for that).
+    Doc.setVis(!m.isCancel && (makerRedeemCoin(m) || !m.revoked || (m.revoked && m.side === OrderUtil.Taker)), tmpl.makerRedeem)
```
norwnd, line 374: I wonder if this part (it seems, this code makes this assumption) I documented here: `When revoked, there is no need to show maker redeem if we are maker (since it won't happen)` is always **true** ?

Could the following happen ? Maker `dexc` publishes maker redeem transaction roughly at the moment of match revocation, such that redeem transaction will be mined **after** match is already revoked, resulting in:
- maker actually redeemed
- but (as Maker) we don't show it in UI

that would be bad ^, it seems to me ...

norwnd:
I've looked a bit at the issue @JoeGruffins [raised above](https://github.com/decred/dcrdex/pull/2042#issuecomment-1401551040), and given:

> The maker could redeem at any time, forcing us to go the redeem path instead of refund.

I think (if we are Taker) we should show both **Maker Redemption** and **Taker Redemption** (along with **Refund** message, after Revocation kicks in) up until the moment when Taker gives up on searching for maker redemption (which happens only after successful refund, from what I figured). That is to take into account the case when Maker redeems **after** Taker starts trying to broadcast his refund transaction but **before** he manages to do so (I assume this is what @chappjc meant by "it's passed the taker refund time"):

> but only when it's passed the taker refund time (not just revoked and counting down)

added a commit: https://github.com/decred/dcrdex/pull/2042/commits/c61a71ef5ed1f61745c8c73359f662eed60c3afa to implement just that ^ (can remove/update if I got it wrong),

and I [left another relevant comment](https://github.com/decred/dcrdex/pull/2042/files#r1086553970) for similar (yet unsolved) "unfortunate timing" issue that Maker might encounter

client/webserver/site/src/js/order.ts
norwnd: Or maybe (for Maker) when Maker sent redeem transaction and it is waiting in mempool - `makerRedeemCoin(m)` will already have it for display ?

client/webserver/site/src/js/locales.ts
norwnd: > Will happen after -> Refund will happen after

I started with this one actually, but repeated Refund (in title, and actual body) kinda scared me off.

Will apply, if nobody has other suggestions.

norwnd: Updated: https://github.com/decred/dcrdex/pull/2042/commits/3dc2b757f598339ce4cb241afc66fef6b28f387b

JoeGruffins:
Before redeeming is different and looks ok, but the third line does not go away after redeem. If refunded we don't need to show pending right?
image
image

norwnd:
> If refunded we don't need to show pending right?

Yeah, **after** successful refund it shouldn't show up, I didn't have a chance to fully  test this PR yet (after latest changes) - planning on testing it more soon.

norwnd:
Updated **bodybuilder.tmpl** and rebased (in latest force-push), more changes coming.

norwnd:
> > If refunded we don't need to show pending right?
> 
> Yeah, **after** successful refund it shouldn't show up, I didn't have a chance to fully test this PR yet (after latest changes) - planning on testing it more soon.

I've played around with match-card some more, 

to keep the complexity manageable, and get some flexibility when showing **Refund** we might want to display match slightly differently depending on whether it has been revoked. Trying to simply extend conditions that are in place in **master** gets quite messy, so instead I split match displaying in two sections  https://github.com/decred/dcrdex/pull/2042/commits/7440e286d404a72d00038d26658b6344c5e56014 : **success-path**, **alternative-path** (based on `m.revoked` flag) to handle each of them separately.

Did quite a bit of testing on this also, and it seems to be working same way as in **master** except it should fix issues @JoeGruffins mentioned above.

client/webserver/site/src/js/order.ts
```diff
@@ -308,14 +366,67 @@ export default class OrderPage extends BasePage {
...
+      // we'll show it in that case).
+      Doc.setVis(!m.isCancel && (takerSwapCoin(m)), tmpl.takerSwap)
+      // When match is revoked and both swaps are present, maker redeem might still show up:
+      // - as maker, we'll try to redeem until it isn't possible due to refund by taker,
+      //   so we'll have to show redeem coin until we either redeem or refund
+      //   because there isn't enough data on match card to figure out whether taker
+      //   refunded already
+      // - as taker, we should expect maker redeeming any time, so we'll have to show
+      //   redeem coins until make redeem shows up, or until we refund.
+      Doc.setVis(!m.isCancel && (makerRedeemCoin(m) || (makerSwapCoin(m) && takerSwapCoin(m) && m.active && !m.refund)), tmpl.makerRedeem)
```
buck54321, line 401: Can drop `makerSwapCoin(m)` from the conditional. if `takerSwapCoin(m)` evals `true`, the the maker swap coin will definitely be there.

client/webserver/site/src/js/order.ts
```diff
@@ -308,14 +366,67 @@ export default class OrderPage extends BasePage {
...
+      // When maker isn't aware of taker redeem coin, once the match becomes inactive
+      // (nothing else maker is expected to do in this match) just hide taker redeem.
+      Doc.setVis(!m.isCancel && (takerRedeemCoin(m) || (makerRedeemCoin(m) && m.active && !m.refund)), tmpl.takerRedeem)
+      // As taker, show refund placeholder only if we have outstanding swap to refund.
+      // There is no need to wait for anything else, we can show refund placeholder
+      // (to inform the user that it is likely to happen) right after match revocation.
+      let expectingRefund = !!takerSwapCoin(m) // as taker
```
buck54321, line 408: I prefer the semantics of `Boolean(takerSwapCoin(m))`, but not a biggie.

client/webserver/site/src/js/order.ts
```diff
@@ -308,14 +366,67 @@ export default class OrderPage extends BasePage {
...
+        // this case.
+        if (takerSwapCoin(m)) {
+          const takerRefundsAfter = new Date(m.stamp + lockTimeTakerMs)
+          expectingRefund = expectingRefund && Date.now() > takerRefundsAfter.getTime()
+        }
+      }
+      // We need to show refund placeholder if the match is still active after it has
+      // been revoked (to provide some feedback to the user), but refund placeholder
+      // should not be shown if there is a known redeem (or counter redeem, depending
+      // on our side).
```
buck54321, line 427: I think this message repeats and detracts from the more detailed docs above. Recommend delete this one.

dex/asset.go
```diff
@@ -59,19 +59,23 @@ func init() {
...
 }
 
-// LockTimeTaker returns the taker locktime value that should be used by both
-// client and server for the specified network. Mainnet uses a constant value
-// while test networks support setting a custom value during build.
+// LockTimeTaker returns the taker locktime value (i.e. how long funds are locked
+// for in swap contract before they can be pulled out - which happens on refund
+// paths) that client and server will use for the specified network. Mainnet uses
+// a constant value while test networks support setting a custom value
+// during build.
```
buck54321, line 66: how long funds are locked for in swap contract before they can be ~~pulled out - which happens on refund paths~~ **refunded**

here and below

buck54321 (changes requested):

client/webserver/site/src/js/order.ts
```diff
@@ -308,14 +366,67 @@ export default class OrderPage extends BasePage {
...
+      // After revocation, we don't expect swap coin to appear (it could though, and
+      // we'll show it in that case).
+      Doc.setVis(!m.isCancel && (takerSwapCoin(m)), tmpl.takerSwap)
+      // When match is revoked and both swaps are present, maker redeem might still show up:
+      // - as maker, we'll try to redeem until it isn't possible due to refund by taker,
+      //   so we'll have to show redeem coin until we either redeem or refund
+      //   because there isn't enough data on match card to figure out whether taker
+      //   refunded already
+      // - as taker, we should expect maker redeeming any time, so we'll have to show
+      //   redeem coins until make redeem shows up, or until we refund.
```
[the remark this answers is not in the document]
norwnd: yeah, I think the original phrasing (on master) is like `setCoin`, so I just kept it same, will update

client/webserver/site/src/js/order.ts
```diff
@@ -258,33 +263,86 @@ export default class OrderPage extends BasePage {
...
     tmpl.status.textContent = OrderUtil.matchStatusString(m)
 
-    const setCoin = (pendingName: string, linkName: string, coin: Coin) => {
-      const formatCoinID = (cid: string) => {
-        if (cid.startsWith(coinIDTakerFoundMakerRedemption)) {
-          const makerAddr = cid.substring(coinIDTakerFoundMakerRedemption.length)
-          return intl.prep(intl.ID_TAKER_FOUND_MAKER_REDEMPTION, { makerAddr: makerAddr })
-        }
-        return cid
-      }
```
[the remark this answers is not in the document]
norwnd: This is analogous to `setCoinHref` which could also be inlined (since it is used only once),

so since `setCoinHref` was moved outside seems natural to do same for `formatCoinID`, to simplify the `setMutableMatchCardElements` overall, or would you prefer to inline both ?

dex/asset.go
norwnd: updated

client/webserver/site/src/js/order.ts
norwnd: changed to `Boolean`

client/webserver/site/src/js/order.ts
norwnd: removed

client/webserver/site/src/js/order.ts
norwnd: removed

client/webserver/site/src/js/order.ts
```diff
@@ -308,14 +366,67 @@ export default class OrderPage extends BasePage {
...
+      Doc.setVis(!m.isCancel && m.refund, tmpl.refund)
+    } else {
+      // Match diverged from the usual success-path, since this could have happened
+      // at any step it is hard (maybe impossible) to predict the final state this
+      // match will end up in, so show only steps that already happened plus all
+      // the possibilities on the next step ahead.
+
+      // After revocation, we don't expect swap coin to appear (it could though, and
+      // we'll show it in that case).
+      Doc.setVis(!m.isCancel && (makerSwapCoin(m)), tmpl.makerSwap)
```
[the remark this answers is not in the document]
norwnd: updated

client/webserver/site/src/js/order.ts
```diff
@@ -308,14 +366,67 @@ export default class OrderPage extends BasePage {
...
+      // After revocation, we don't expect swap coin to appear (it could though, and
+      // we'll show it in that case).
+      Doc.setVis(!m.isCancel && (takerSwapCoin(m)), tmpl.takerSwap)
+      // When match is revoked and both swaps are present, maker redeem might still show up:
+      // - as maker, we'll try to redeem until it isn't possible due to refund by taker,
+      //   so we'll have to show redeem coin until we either redeem or refund
+      //   because there isn't enough data on match card to figure out whether taker
+      //   refunded already
```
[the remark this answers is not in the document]
norwnd: Meant `data in m` (also updated comment), but I admit I have limited understanding of this subject - clarifications are welcome!

client/webserver/site/src/js/order.ts
```diff
@@ -308,14 +366,67 @@ export default class OrderPage extends BasePage {
...
+
+      // After revocation, we don't expect swap coin to appear (it could though, and
+      // we'll show it in that case).
+      Doc.setVis(!m.isCancel && (makerSwapCoin(m)), tmpl.makerSwap)
+      // After revocation, we don't expect swap coin to appear (it could though, and
+      // we'll show it in that case).
+      Doc.setVis(!m.isCancel && (takerSwapCoin(m)), tmpl.takerSwap)
+      // When match is revoked and both swaps are present, maker redeem might still show up:
+      // - as maker, we'll try to redeem until it isn't possible due to refund by taker,
+      //   so we'll have to show redeem coin until we either redeem or refund
```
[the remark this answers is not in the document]
norwnd: updated

client/webserver/site/src/js/order.ts
```diff
@@ -258,33 +263,86 @@ export default class OrderPage extends BasePage {
...
 
-    const setCoin = (pendingName: string, linkName: string, coin: Coin) => {
-      const formatCoinID = (cid: string) => {
-        if (cid.startsWith(coinIDTakerFoundMakerRedemption)) {
-          const makerAddr = cid.substring(coinIDTakerFoundMakerRedemption.length)
-          return intl.prep(intl.ID_TAKER_FOUND_MAKER_REDEMPTION, { makerAddr: makerAddr })
-        }
-        return cid
-      }
+    const tryShowCoin = (showPlaceholder: () => void, showCoin: () => void, linkName: string, coin: Coin) => {
```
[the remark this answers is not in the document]
norwnd: Applied, however I'd also suggest minor improvement on top: https://github.com/decred/dcrdex/pull/2042/commits/17542c2c5a6995bf8d62183c17ee776a24ec33f8 to keep all the code related to `refund` together in one place, instead of spread out (with that extra `if (m.refund) {`), thoughts ?

**update:** it should be also `if (m.refund) {` -> `if (!m.refund) {` in your example, right ? I noticed that only after I've done that refactoring https://github.com/decred/dcrdex/pull/2042/commits/17542c2c5a6995bf8d62183c17ee776a24ec33f8 (pulling this code close together)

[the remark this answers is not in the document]
norwnd: Applied, however I'd also suggest minor improvement on top: https://github.com/decred/dcrdex/pull/2042/commits/17542c2c5a6995bf8d62183c17ee776a24ec33f8 to keep all the code related to `refund` together in one place, instead of spread out (with that extra `if (m.refund) {`), thoughts ?

**update:** it should be also `if (m.refund) {` -> `if (!m.refund) {` in your example, right ? I noticed that only after I've done that refactoring https://github.com/decred/dcrdex/pull/2042/commits/17542c2c5a6995bf8d62183c17ee776a24ec33f8 (pulling this code close together)

client/webserver/site/src/js/order.ts
```diff
@@ -308,14 +329,60 @@ export default class OrderPage extends BasePage {
...
+        // to refund our own swap) until taker refunds, so start showing refund
+        // placeholder only after taker is expected to start his refund process in
+        // this case.
+        if (takerSwapCoin(m)) {
+          const takerRefundsAfter = new Date(m.stamp + lockTimeTakerMs)
+          expectingRefund = expectingRefund && Date.now() > takerRefundsAfter.getTime()
+        }
+      }
+      Doc.setVis(!m.isCancel && (m.refund || (m.active && !m.redeem && !m.counterRedeem && expectingRefund)), tmpl.refund)
+    }
```
buck54321, line 385: I think most of this could be done with few changes to the existing code. It's somewhat self-documenting.

```Typescript
if (m.isCancel) return
Doc.setVis(makerSwapCoin(m) || (m.active && !m.revoked), tmpl.makerSwap)
Doc.setVis(takerSwapCoin(m) || (m.active && !m.revoked), tmpl.takerSwap)
Doc.setVis(makerRedeemCoin(m) || (!m.revoked && m.active) || (m.revoked && takerSwapCoin(m) && m.active && !m.refund), tmpl.makerRedeem)
Doc.setVis(takerRedeemCoin(m) || (!m.revoked && m.active) || (m.revoked && makerRedeemCoin(m) && m.active && !m.refund), tmpl.takerRedeem)
```

dex/asset.go
```diff
@@ -59,19 +59,21 @@ func init() {
 	}
 }
 
-// LockTimeTaker returns the taker locktime value that should be used by both
-// client and server for the specified network. Mainnet uses a constant value
-// while test networks support setting a custom value during build.
+// LockTimeTaker returns the taker locktime value (i.e. how long funds are locked
+// for in swap contract before they can be refunded) that client and server will
```
buck54321, line 63: There's no need to add the definition of locktime to the docs for these functions. Locktime is a fundamental concept for atomic swaps, and is well-defined via other documentation.

norwnd: Does it hurt to be slightly repetitive with such documentation ? The locktime concept is obvious to someone who maintained this code for a while, for a newcomer looking at this code for the very first time it might not be 100% clear whether it is The locktime (or he might not even be familiar with the concept yet and is just browsing through the code, that's how I started).

client/webserver/site/src/js/order.ts
norwnd: Haven't checked whether this would be exact 1-to-1 equivalent to what I currently have here (if I remember correctly it addresses >= 1 issue that we currently have in master), 

but it kind of defeats the goal I tried to accomplish here - which is **to break down and document** each non-trivial path we might have here, having spent quite some time on this I'd say there are plenty of caveats here (and we might need to adjust this code in the future for some reason, so having it all documented could save effort).

dex/asset.go
chappjc: it's perfectly clear

chappjc:
Generally OK.  Please revive with reviewer comments resolved, otherwise please express intent to abandon.  Will close tomorrow.

client/webserver/site/src/js/order.ts
```diff
@@ -308,14 +329,60 @@ export default class OrderPage extends BasePage {
...
+        // can show refund placeholder (to inform the user that it is likely to happen)
+        // right after match revocation.
+        expectingRefund = Boolean(makerSwapCoin(m))
+        // If we discover taker swap we'll be trying to redeem it (instead of trying
+        // to refund our own swap) until taker refunds, so start showing refund
+        // placeholder only after taker is expected to start his refund process in
+        // this case.
+        if (takerSwapCoin(m)) {
+          const takerRefundsAfter = new Date(m.stamp + lockTimeTakerMs)
+          expectingRefund = expectingRefund && Date.now() > takerRefundsAfter.getTime()
```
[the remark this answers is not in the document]
norwnd: > we have no way of differentiating here whether the taker has actually refunded or if we're waiting on confirmations or wallet connectivity or something else, so this approach is maybe the best we can do with the information available?

Yes, this is my understanding too.

dex/asset.go
norwnd: Okay, reverted my updates to those comments.

client/webserver/site/src/js/order.ts
```diff
@@ -308,14 +329,60 @@ export default class OrderPage extends BasePage {
       Doc.hide(tmpl.makerSwapMsg, tmpl.takerSwapMsg, tmpl.makerRedeemMsg, tmpl.takerRedeemMsg)
     }
```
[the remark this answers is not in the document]
norwnd: > do a if (m.isCancel) return here and get rid of all of the !m.isCancel checks below.

Yeah, that would simplify it a bit, applied. 

> Maybe we should hide all of the coin elements by default (add d-hide class)

Added.

norwnd:
Noticed and fixed in https://github.com/decred/dcrdex/pull/2042/commits/6240b94495cd8d314c9536f05a342c086b5fdd36:

- I introduced a bug (caused by rogue `return`) when inlining `func`
- we already check `m.isCancel` at the very first line when entering `setMutableMatchCardElements` func, so no need to check it 2nd time

client/webserver/site/src/js/order.ts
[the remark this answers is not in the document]
chappjc: > If we're the maker revoked in `TakerSwapCast`, we'll redeem up until we can refund, as long as the taker doesn't refund first. **core** does not care if the taker's locktime has passed, only whether the taker has actually refunded.

As maker, `core` won't redeem the taker's swap if it's locktime has passed. That's extremely important because maker is revealing the secret in this step, and the taker could quickly attempt to double spend their own swap contract in a refund _and_ redeem the maker's contract since they now have the secret.  Since the first seen policies are going by the wayside recently, it's even riskier.  https://github.com/decred/dcrdex/pull/1548 added this to `isRedeemable`:

https://github.com/decred/dcrdex/blob/d68ae847f8a8992ce248eeaec4c13275c3d7e6d7/client/core/trade.go#L1554-L1581

We can be sure that as maker there will be no redeem of the taker swap if the taker's swap contract has expired.

What was being displayed with the previous logic for show/hiding these fielids?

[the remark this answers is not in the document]
norwnd: > What was being displayed incorrectly with the previous logic for show/hiding these fielids?

**Maker**, ~2h since matched, before/after:
imageimage

- so we should be waiting Maker redeeming, not refunding until Taker is able to refund right ? This is fixed here.

**Taker**, ~2h since matched, before/after:
imageimage

- in this PR, we display now expected refund time (if refund path gets taken)
- in this PR, we display Maker redemption (3rd line) but not Taker (our) redemption (4rd line), while on **master** we don't display 3rd line at all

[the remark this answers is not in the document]
norwnd: The weird thing I encountered just now is: Maker doesn't publish redeem transaction due to:
```
2023-04-17 05:32:47.103 [INF] CORE: Match 1bf6ef2071d48b6e2f9201770fbcd4605c7a5c9c75a149d5e8da82769f3c0c63 not yet redeemable: current confs = 0, required confs = 1
```
even though it already has 49 confirmations (according to **dcrdata**):
image

My DCR wallet is 100% synched:
image

I think it did always publish it in the past, maybe some kind of regression ?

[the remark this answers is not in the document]
chappjc: Smells like another reorg issue related to "external txns" / block filters scan.  Notice the Included in Block on dcrdata shows an orphaned block.

client/webserver/site/src/js/order.ts
```diff
@@ -260,40 +265,59 @@ export default class OrderPage extends BasePage {
...
+    tryShowCoin(tmpl.makerRedeemPending, tmpl.makerRedeemCoin, makerRedeemCoin(m))
+    tryShowCoin(tmpl.takerRedeemPending, tmpl.takerRedeemCoin, takerRedeemCoin(m))
+    if (!m.refund) {
+      // Special messaging for pending refunds.
+      let lockTime = lockTimeMakerMs
+      if (m.side === OrderUtil.Taker) lockTime = lockTimeTakerMs
+      const refundAfter = new Date(m.stamp + lockTime)
+      if (Date.now() > refundAfter.getTime()) tmpl.refundPending.textContent = intl.prep(intl.ID_REFUND_IMMINENT)
+      else {
+        const refundAfterStr = refundAfter.toLocaleTimeString('en-GB', {
```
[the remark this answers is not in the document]
norwnd: I just copied it from here without thinking much about it:
```go
tmpl.matchTime.textContent = time.toLocaleTimeString('en-GB', {
```
I guess both need changing ? [passing in `navigator.languages as string[]` there now](https://github.com/decred/dcrdex/pull/2042/commits/c643bafcdb569a12dc0f0668f9c70f680f9b9b52) (like we do for formatters I've worked with recently) and it seems to be working fine.

client/webserver/site/src/js/order.ts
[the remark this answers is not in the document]
norwnd: > Smells like another reorg issue related to "external txns" / block filters scan. Notice the Included in Block on dcrdata shows an orphaned block.

Hm, I can look into this a bit later, it has no relation to this PR I believe.

> We can be sure that as maker there will be no redeem of the taker swap if the taker's swap contract has expired.

If that's the case then there is no need to keep showing Maker redeem (3rd line) like I'm currently doing in this PR, we can hide it like that (it is also hidden on `master` but as can be seen on screenshots in my comment above it's overly aggressive and hides it regardless Taker locktime): https://github.com/decred/dcrdex/pull/2042/commits/aa25f0db8d81a129909bbd4e449a3bb22aca9c15

chappjc:
Very tentatively in the v0.6.1 milestone.

buck54321 (approved):

PR #2043, opened 2023-01-12
ui: order-form responsive validation

norwnd:
Replaces https://github.com/decred/dcrdex/pull/1921.

To implement responsive order-form validation HTML layout and JS event handlers need to be rewritten. Since there is quite a bit of JS-code  relying on the way event handlers were written, those code-bits also needed updates/rewrites to fit in with the new event-handlers, hopefully, these changes overall make order-form handling code more explicit / thorougly documented, and less tightly coupled.

Additionally, some bugs were discovered (when testing this functionality) and fixed here.

List of most notable changes:
- validation for `Price`, `Quantity` fields should be more responsive, so that the user has a better sense of why his inputs don't make sense (this is mostly accomplished by highlighting background/border for relevant order-form element; and maybe hiding some elements like `Total` when its not possible to calculate based on current user-inputs)
- `Place order to buy ...` button press should fail right away when `Total` exceeds `Max Buy`, instead of allowing user to proceed - only to fail later on
- `Place order to ...` animates now; that 1) allows for more snappy user-feedback and more importantly 2) prevents user from clicking on it right after he's entered invalid `Rate` / `Quantity` inputs (otherwise user might not notice the rounding or will learn about his mistake in delayed manner)
- fixed a bug where the `orderTotalPreview` for market orders was not being set correctly
- added `min` and `step` attributes for various number inputs on order-form, resulting in better messaging and enabling the use of arrow keys on number inputs, and fixing a bug with Up-key not working with `Quantity` input
- fixed a bug with applying the result of `Max Buy/Sell` calculation in callback; by the time this callback displays the result, the user might have switched from Buy to Sell side, or vice versa (resulting in order-form showing him `Max Buy/Sell` for the opposite order side, not the one he switched to)
- having all 4 order types: Limit/Market+Buy/Sell managed by same view seems have resulted into tight-coupling of some code in `markets.ts`. To keep code-complexity manageable, I've split off Market-Sell order stuff into its own separate view (Limit Buy/Sell forms are almost identical, can't say same for Market orders). Market-Buy form was already handled by separate view (see `MARKET BUY ORDER` in `markets.tmpl`)
- added `Max Sell` section for "Market Sell" order type, which is probably just as useful as on limit form

### DIFF
--- a/client/webserver/site/src/css/main_dark.scss
+++ b/client/webserver/site/src/css/main_dark.scss
@@ -84,7 +84,6 @@ body.dark {
   input,
   input:focus {
     border: none;
-    outline: none;
     color: $font-color-dark;
   }
 
--- a/client/webserver/site/src/css/market.scss
+++ b/client/webserver/site/src/css/market.scss
@@ -56,10 +56,6 @@ div[data-handler=markets] {
         font-size: 14px;
       }
 
-      input:focus {
-        outline: none;
-      }
-
       span.unitbox {
         position: absolute;
         font-size: 14px;
@@ -100,7 +96,8 @@ div[data-handler=markets] {
         color: #1e7d11;
       }
 
-      #orderPreview,
+      #orderTotalPreview,
+      #mktSellTotalPreview,
       .h21 {
         height: 21px;
       }
--- a/client/webserver/site/src/html/bodybuilder.tmpl
+++ b/client/webserver/site/src/html/bodybuilder.tmpl
@@ -9,7 +9,7 @@
   <link rel="icon" href="/img/favicon.png?v=AK4XS4">
   <meta name="description" content="Decred DEX Client Web Portal">
   <title>{{.Title}}</title>
-  <link href="/css/style.css?v=XYXY183gH" rel="stylesheet">
+  <link href="/css/style.css?v=LSBFflh8d" rel="stylesheet">
 </head>
 <body {{if .UserInfo.DarkMode}} class="dark"{{end}}>
   <div class="popup-notes" id="popupNotes">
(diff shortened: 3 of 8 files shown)


norwnd:
I've re-tested this recently and found 1 unpleasant side-effect of responsive validation, the way it is currently implemented in https://github.com/decred/dcrdex/pull/2043/commits/4f31cdd71e06f78b1e5ed0e141f2fcdc649f5f05 - validation won't allow placing order (namely, switching to verify order form) until `Max Buy/Sell` has been calculated and is available for comparison against `Quantity`,

[that next commit solves it](https://github.com/decred/dcrdex/pull/2043/commits/0ed13558794e071fd7c1d9b2683f4b51ac2f432f), skipping this comparison step in case `Max Buy/Sell` isn't readily available in UI for some reason (e.g. it could be a long-running HTTP request), so, the user will be able to place an order and in case he doesn't have enough funds to cover it he'll just receive `error` response from **dexc** later on.

buck54321:
> I've re-tested this recently and found 1 unpleasant side-effect of responsive validation, the way it is currently implemented in [4f31cdd](https://github.com/decred/dcrdex/commit/4f31cdd71e06f78b1e5ed0e141f2fcdc649f5f05) - validation won't allow placing order (namely, switching to verify order form) until `Max Buy/Sell` has been calculated and is available for comparison against `Quantity`

I don't want to be that guy, but this is exactly the type of thing that concerns me. Your previous statement

> So, while I do appreciate the security concerns you guys have raised, I think the most practical way to go about it is to take our time and carefully review what's implemented here in one piece.

seems to me to be misguided. I think the most prudent way to go about this is to break off the responsive form feedback into a new, focused PR that addresses the problem minimally. That way we reviewers can focus on the problem-at-hand, and potentially catch these bugs.

chappjc:
Closing as deadlocked.  Can reopen later.

Issue #2045, opened 2023-01-12
eth rpc: timeouts and failing connections need better handling


chappjc:
For several hours DNS trouble prevented RPC requests, and they were timing out, but also apparently stacking up.

```
2023-01-12 14:43:54.609 [ERR] CORE: eth wallet is reporting a failed state: failed to get best hash: HeaderByNumber error: Post "https://rpc.ankr.com/eth_goerli": context deadline exceeded
2023-01-12 14:43:56.609 [TRC] CORE[eth][ETH][RPC]: all providers in a failed state, so acting like none are
2023-01-12 14:43:56.609 [TRC] CORE[eth][ETH][RPC]: fetching fresh header from "rpc.ankr.com"
2023-01-12 14:43:56.609 [ERR] CORE: eth wallet is reporting a failed state: failed to get best hash: HeaderByNumber error: Post "https://rpc.ankr.com/eth_goerli": context deadline exceeded
2023-01-12 14:43:58.610 [TRC] CORE[eth][ETH][RPC]: all providers in a failed state, so acting like none are
2023-01-12 14:43:58.610 [TRC] CORE[eth][ETH][RPC]: fetching fresh header from "rpc.ankr.com"
2023-01-12 14:43:58.611 [ERR] CORE: eth wallet is reporting a failed state: failed to get best hash: HeaderByNumber error: Post "https://rpc.ankr.com/eth_goerli": context deadline exceeded
2023-01-12 14:44:00.611 [TRC] CORE[eth][ETH][RPC]: all providers in a failed state, so acting like none are
2023-01-12 14:44:00.611 [TRC] CORE[eth][ETH][RPC]: fetching fresh header from "rpc.ankr.com"
2023-01-12 14:44:00.611 [ERR] CORE: eth wallet is reporting a failed state: failed to get best hash: HeaderByNumber error: Post "https://rpc.ankr.com/eth_goerli": context deadline exceeded
2023-01-12 14:44:02.495 [TRC] CORE: New peer count for asset btc: 7
2023-01-12 14:44:02.611 [TRC] CORE[eth][ETH][RPC]: all providers in a failed state, so acting like none are
2023-01-12 14:44:02.611 [TRC] CORE[eth][ETH][RPC]: fetching fresh header from "rpc.ankr.com"
[...]

PR #2047, opened 2023-01-12
server/eth: Allow geth connections using http and ws

### LINKED ISSUE #2026: eth: add plain http/ws support
With https://github.com/decred/dcrdex/pull/2023 reverting the authrpc changes that allowed server to use an alternative to IPC, we will need to add in support for http/ws RPC for server.

I believe client is fine with it's "multirpc" to use geth's http/ws interfaces without changes.  Only server needs updates.
Is that correct @JoeGruffins?

martonp:
In order to test the connection using ws, in the eth backend's
rpc_harness_test, the delta node (the only one that has enabled
http and ws) is used to send funds instead of the alpha node.

The rpc_harness_test had pre-existing issues:
- panicing due to registering the test token twice
- incorrectly checking the difference in token balance

These issues are now fixed.

Closes #2026


server/asset/eth/rpcclient.go
```diff
@@ -48,24 +49,34 @@ type rpcclient struct {
...
+	var client *rpc.Client
+	var err error
+	if strings.HasSuffix(c.endpoint, ".ipc") {
+		client, err = rpc.DialIPC(ctx, c.endpoint)
+		if err != nil {
+			return fmt.Errorf("unable to dial ipc: %v", err)
+		}
+	} else {
+		log.Debugf("dialing endpoint: %s", c.endpoint)
+		client, err = rpc.Dial(c.endpoint)
```
chappjc, line 72: `DialContext(ctx, ...` please.  And will you also change the dial in multirpc while doing this?  Related: https://github.com/decred/dcrdex/issues/2045

server/asset/eth/rpcclient_harness_test.go
```diff
@@ -153,9 +153,18 @@ func testAccountBalance(t *testing.T, assetID uint32) {
...
-	if diff.Cmp(dexeth.GweiToWei(vGwei)) <= 0 {
-		t.Fatalf("account balance changed by %d. expected > %d", dexeth.WeiToGwei(diff), uint64(vGwei))
+	if assetID == BipID {
+		diff := new(big.Int).Sub(balBefore, balAfter)
+		if diff.Cmp(dexeth.GweiToWei(vGwei)) <= 0 {
+			t.Fatalf("account balance changed by %d. expected > %d", dexeth.WeiToGwei(diff), uint64(vGwei))
+		}
+	}
+
+	if assetID == testTokenID {
```
chappjc, line 163: Will you beef up the commit message with notes about these other fixes, and changes like using delta, etc.?

dex/networks/eth/common.go
```diff
@@ -55,7 +55,7 @@ func CheckAPIModules(c *rpc.Client, endpoint string, log dex.Logger, reqModules
 		for v := range reqModulesMap {
 			reqs = append(reqs, v)
 		}
-		return fmt.Errorf("needed apis not present: %v", strings.Join(reqs, " "))
+		return fmt.Errorf("needed apis not present: %v. some APIs may only be available over websocket and not http.", strings.Join(reqs, " "))
```
chappjc, line 58: I'm curious if there was a specific case where you found ws had it but on http.  I assume that's solvable with the `--http.api=` flag, unless you were referring to a third-party RPC provider.

I'm afraid we may have to go with the "dumbed down" balance check https://github.com/decred/dcrdex/pull/2020 because I haven't seen any RPC providers (other than geth itself) that provide `txpool`.  e.g. https://docs.alchemy.com/reference/ethereum-api-faq#what-methods-does-alchemy-support-for-the-ethereum-api
Or we could at least fall back to that method when txpool is not available.  Not needed for this PR, but have been thinking.

chappjc (changes requested):

chappjc:
Good stuff.  dex-test:

```
[INF] DEX: Starting asset backend "eth"...
[DBG] ASSET[eth][ETH]: dialing endpoint: ws://127.0.0.1:8546
[DBG] ASSET[eth][ETH]: API endpoints supported by ws://127.0.0.1:8546: admin:1.0 eth:1.0 net:1.0 rpc:1.0 txpool:1.0 web3:1.0
```

dex/networks/eth/common.go
martonp: Yeah you're right, I just had to add -http.api.

chappjc (approved):

PR #2049, opened 2023-01-13
client/assets: Add redeem confirms to utxo assets.

### LINKED ISSUE #1962: utxo assets need redeemed swaps monitored through confirmation too
Although less critical than ETH where txns get clobbered very easily by nonce replacement, Bitcoin's mempool can get so full that transactions are purged (and never mined).  In the last couple days, this threshold has been as high as 11 sat/vB.  With no redeem rebroadcasting for UTXO assets, that was nearly a problem with client configs with very high redeem confirmation targets, like 6.

In addition to txns being evicted from mempool prior to being mined, broadcasting can fail with errors like `error sending redeem transaction: rawrequest (sendtransaction) error: -26: mempool fee not met, 143 < 1924`.  While an error like that from a full-node wallet like bitcoind is good because it will keep trying to rebroadcast, the light clients more readily declare success if peers decide not to respond in error, or we simply move on assuming the publish succeeded.  In this light client case, the neutrino broadcaster should continually rebroadcast, eventually succeeding when network conditions permit, but this could be too late.  It would be best to have logic to change the fee and try a new redeem.

JoeGruffins:
closes #1962

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -82,6 +82,11 @@ const (
 	// testnetExternalApiUrl is the URL of the testnet external API in case of
 	// fallback.
 	testnetExternalApiUrl = "https://mempool.space/testnet/api/"
+
+	// requiredRedeemConfirms is the amount of confirms a redeem transaction
+	// needs before the trade is considered confirmed. The redeem is
+	// monitored until this number of confirms is reached.
+	requiredRedeemConfirms = 1
 )
 
 const (
@@ -5796,4 +5801,83 @@ func deserializeMsgTx(r io.Reader) (*wire.MsgTx, error) {
 // msgTxFromBytes creates a wire.MsgTx by deserializing the transaction.
 func msgTxFromBytes(txB []byte) (*wire.MsgTx, error) {
 	return deserializeMsgTx(bytes.NewReader(txB))
+}
+
+// ConfirmRedemption returns how many confirmations a redemption has. Normally
+// this is very straitforward. However, with fluxuating fees, there's the
+// possibility that the tx is never mined and eventually purged from the
+// mempool. In that case we use the provided fee suggestion to create and send
+// a new redeem transaction, returning the new transactions hash.
+func (btc *baseWallet) ConfirmRedemption(coinID dex.Bytes, redemption *asset.Redemption, feeSuggestion uint64) (*asset.ConfirmRedemptionStatus, error) {
+	txHash, _, err := decodeCoinID(coinID)
+	if err != nil {
+		return nil, err
+	}
+
+	_, confs, err := btc.rawWalletTx(txHash)
+	// redemption transaction found, return its confirms.
+	//
+	// TODO: Investigate the case where this redeem has been sitting in the
+	// mempool for a long amount of time, possibly requiring some action by
+	// us to get it unstuck.
+	if err == nil {
+		return &asset.ConfirmRedemptionStatus{
+			Confs:  uint64(confs),
+			Req:    requiredRedeemConfirms,
+			CoinID: coinID,
+		}, nil
+	}
+
+	if !errors.Is(err, WalletTransactionNotFound) {
+		return nil, fmt.Errorf("problem searching for redemption transaction %s: %w", txHash, err)
+	}
+
+	// Redemption transaction is missing from the point of view of our node!
+	// Unlikely, but possible it was redeemed by another transaction. Check
+	// if the contract is still an unspent output.
+
+	pkScript, err := btc.scriptHashScript(redemption.Spends.Contract)
+	if err != nil {
+		return nil, fmt.Errorf("error creating contract script: %w", err)
+	}
+
+	swapHash, vout, err := decodeCoinID(redemption.Spends.Coin.ID())
+	if err != nil {
+		return nil, err
+	}
+
+	utxo, _, err := btc.node.getTxOut(swapHash, vout, pkScript, time.Now().Add(-ContractSearchLimit))
+	if err != nil {
+		return nil, fmt.Errorf("error finding unspent contract %s with swap hash %v vout %d: %w", redemption.Spends.Coin.ID(), swapHash, vout, err)
+	}
+	if utxo == nil {
+		// TODO: Spent, but by who. Find the spending tx.
+		btc.log.Warnf("Contract coin %v with swap hash %v vout %d spent by someone but not sure who.", redemption.Spends.Coin.ID(), swapHash, vout)
+		// Incorrect, but we will be in a loop of erroring if we don't
+		// return something.
+		return &asset.ConfirmRedemptionStatus{
+			Confs:  requiredRedeemConfirms,
+			Req:    requiredRedeemConfirms,
+			CoinID: coinID,
+		}, nil
+	}
+
+	// The contract has not yet been redeemed, but it seems the redeeming
+	// tx has disappeared. Assume the fee was too low at the time and it
+	// was eventually purged from the mempool. Attempt to redeem again with
+	// a currently reasonable fee.
+
+	form := &asset.RedeemForm{
+		Redemptions:   []*asset.Redemption{redemption},
+		FeeSuggestion: feeSuggestion,
+	}
+	_, coin, _, err := btc.Redeem(form)
+	if err != nil {
+		return nil, fmt.Errorf("unable to re-redeem %s with swap hash %v vout %d: %w", redemption.Spends.Coin.ID(), swapHash, vout, err)
+	}
+	return &asset.ConfirmRedemptionStatus{
+		Confs:  0,
+		Req:    requiredRedeemConfirms,
+		CoinID: coin.ID(),
+	}, nil
 }
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -88,6 +88,12 @@ const (
 	// freshFeeAge is the expiry age for cached fee rates of external origin,
 	// past which fetchFeeFromOracle should be used to refresh the rate.
 	freshFeeAge = time.Minute
+
+	// requiredRedeemConfirms is the amount of confirms a redeem transaction
+	// needs before the trade is considered confirmed. The redeem is
+	// monitored until this number of confirms is reached. Two to make sure
+	// the block containing the redeem is stakeholder-approved
+	requiredRedeemConfirms = 2
 )
 
 var (
@@ -101,6 +107,13 @@ var (
 	peerCountTicker              = 5 * time.Second
 	conventionalConversionFactor = float64(dexdcr.UnitInfo.Conventional.ConversionFactor)
 	walletBlockAllowance         = time.Second * 10
+
+	// maxRedeemMempoolAge is the max amount of time the wallet will let a
+	// redeem transaction sit in mempool from the time it is first seen
+	// until it attempts to abandon it and try to send a new transaction.
+	// This is necessary because transactions with already spent inputs may
+	// be tried over and over with wallet in SPV mode.
+	maxRedeemMempoolAge = time.Hour * 2
 
 	walletOpts = []*asset.ConfigOption{
 		{
@@ -537,6 +550,11 @@ type exchangeWalletConfig struct {
 	apiFeeFallback   bool
 }
 
+type mempoolRedeem struct {
+	txHash    chainhash.Hash
+	firstSeen time.Time
+}
+
 // ExchangeWallet is a wallet backend for Decred. The backend is how the DEX
 // client app communicates with the Decred blockchain and wallet. ExchangeWallet
 // satisfies the dex.Wallet interface.
@@ -585,6 +603,10 @@ type ExchangeWallet struct {
 
 	externalTxMtx   sync.RWMutex
 	externalTxCache map[chainhash.Hash]*externalTx
+
+	// TODO: Consider persisting mempool redeems on file.
+	mempoolRedeemsMtx sync.RWMutex
+	mempoolRedeems    map[[32]byte]*mempoolRedeem // keyed by secret hash
 }
 
 func (dcr *ExchangeWallet) config() *exchangeWalletConfig {
@@ -813,6 +835,7 @@ func unconnectedWallet(cfg *asset.WalletConfig, dcrCfg *walletConfig, chainParam
 		findRedemptionQueue: make(map[outPoint]*findRedemptionReq),
 		externalTxCache:     make(map[chainhash.Hash]*externalTx),
 		oracleFees:          make(map[uint64]feeStamped),
+		mempoolRedeems:      make(map[[32]byte]*mempoolRedeem),
 	}
 
 	w.cfgV.Store(walletCfg)
@@ -2703,10 +2726,14 @@ func (dcr *ExchangeWallet) Redeem(form *asset.RedeemForm) ([]dex.Bytes, asset.Co
 		return nil, nil, 0, err
 	}
 	coinIDs := make([]dex.Bytes, 0, len(form.Redemptions))
+	dcr.mempoolRedeemsMtx.Lock()
 	for i := range form.Redemptions {
 		coinIDs = append(coinIDs, toCoinID(txHash, uint32(i)))
+		var secretHash [32]byte
+		copy(secretHash[:], form.Redemptions[i].Spends.SecretHash)
+		dcr.mempoolRedeems[secretHash] = &mempoolRedeem{txHash: *txHash, firstSeen: time.Now()}
 	}
-
+	dcr.mempoolRedeemsMtx.Unlock()
 	return coinIDs, newOutput(txHash, 0, uint64(txOut.Value), wire.TxTreeRegular), fee, nil
 }
 
@@ -4908,4 +4935,201 @@ func float64PtrStr(v *float64) string {
 		return "nil"
 	}
 	return strconv.FormatFloat(*v, 'f', 8, 64)
+}
+
+// ConfirmRedemption returns how many confirmations a redemption has. Normally
+// this is very straitforward. However there are two situations that have come
+// up that this also handles. One is when the wallet can not find the redemption
+// transaction. This is most likely because the fee was set too low and the tx
+// was removed from the mempool. In the case where it is not found, this will
+// send a new tx using the provided fee suggestion. The second situation
+// this watches for is a transaction that we can find but has been sitting in
+// the mempool for a long time. This has been observed with the wallet in SPV
+// mode and the transaction inputs having been spent by another transaction. The
+// wallet will not pick up on this so we could tell it to abandon the original
+// transaction and, again, send a new one using the provided feeSuggestion, but
+// only warning for now. This method should not be run for the same redemption
+// concurrently as it need to watch a new redeem transaction before finishing.
+func (dcr *ExchangeWallet) ConfirmRedemption(coinID dex.Bytes, redemption *asset.Redemption, feeSuggestion uint64) (*asset.ConfirmRedemptionStatus, error) {
+	txHash, _, err := decodeCoinID(coinID)
+	if err != nil {
+		return nil, err
+	}
+
+	var secretHash [32]byte
+	copy(secretHash[:], redemption.Spends.SecretHash)
+	dcr.mempoolRedeemsMtx.RLock()
+	mRedeem, have := dcr.mempoolRedeems[secretHash]
+	dcr.mempoolRedeemsMtx.RUnlock()
+
+	var deleteMempoolRedeem bool
+	defer func() {
+		if deleteMempoolRedeem {
+			dcr.mempoolRedeemsMtx.Lock()
+			delete(dcr.mempoolRedeems, secretHash)
+			dcr.mempoolRedeemsMtx.Unlock()
+		}
+	}()
+
+	tx, err := dcr.wallet.GetTransaction(dcr.ctx, txHash)
+	if err != nil && !errors.Is(err, asset.CoinNotFoundError) {
+		return nil, fmt.Errorf("problem searching for redemption transaction %s: %w", txHash, err)
+	}
+	if err == nil {
+		if have && mRedeem.txHash == *txHash {
+			if tx.Confirmations == 0 && time.Now().After(mRedeem.firstSeen.Add(maxRedeemMempoolAge)) {
+				// Transaction has been sitting in the mempool
+				// for a long time now.
+				//
+				// TODO: Consider abandoning.
+				redeemAge := time.Since(mRedeem.firstSeen)
+				dcr.log.Warnf("Redemption transaction %v has been in the mempool for %v which is too long.", txHash, redeemAge)
+			}
+		} else {
+			if have {
+				// This should not happen. Core has told us to
+				// watch a new redeem with a different transaction
+				// hash for a trade we were already watching.
+				return nil, fmt.Errorf("tx were were watching %s for redeem with secret hash %x being "+
+					"replaced by tx %s. core should not be replacing the transaction. maybe ConfirmRedemption "+
+					"is being run concurrently for the same redeem", mRedeem.txHash, secretHash, *txHash)
+			}
+			// Will hit this if dexc was restarted with an actively
+			// redeeming swap.
+			dcr.mempoolRedeemsMtx.Lock()
+			dcr.mempoolRedeems[secretHash] = &mempoolRedeem{txHash: *txHash, firstSeen: time.Now()}
+			dcr.mempoolRedeemsMtx.Unlock()
+		}
+		if tx.Confirmations >= requiredRedeemConfirms {
+			deleteMempoolRedeem = true
+		}
+		return &asset.ConfirmRedemptionStatus{
+			Confs:  uint64(tx.Confirmations),
+			Req:    requiredRedeemConfirms,
+			CoinID: coinID,
+		}, nil
+	}
+
+	// Redemption transaction is missing from the point of view of our wallet!
+	// Unlikely, but possible it was redeemed by another transaction. We
+	// assume a contract past its locktime cannot make it here, so it must
+	// not be refunded. Check if the contract is still an unspent output.
+
+	swapHash, vout, err := decodeCoinID(redemption.Spends.Coin.ID())
+	if err != nil {
+		return nil, err
+	}
+
+	_, _, spentStatus, err := dcr.lookupTxOutput(dcr.ctx, swapHash, vout)
+	if err != nil {
+		return nil, fmt.Errorf("error finding unspent contract: %w", err)
+	}
+
+	switch spentStatus {
+	case -1, 1:
+		// First find the block containing the output itself.
+		scriptAddr, err := stdaddr.NewAddressScriptHashV0(redemption.Spends.Contract, dcr.chainParams)
+		if err != nil {
+			return nil, fmt.Errorf("error encoding contract address: %w", err)
+		}
+		_, pkScript := scriptAddr.PaymentScript()
+		outFound, block, err := dcr.externalTxOutput(dcr.ctx, newOutPoint(swapHash, vout),
+			pkScript, time.Now().Add(-60*24*time.Hour)) // search up to 60 days ago
+		if err != nil {
+			return nil, err // possibly the contract is still in mempool
+		}
+		spent, err := dcr.isOutputSpent(dcr.ctx, outFound)
+		if err != nil {
+			return nil, fmt.Errorf("error checking if contract %v:%d is spent: %w", *swapHash, vout, err)
+		}
+		if !spent {
+			break
+		}
+		vin := -1
+		spendTx := outFound.spenderTx
+		for i := range spendTx.TxIn {
+			sigScript := spendTx.TxIn[i].SignatureScript
+			sigScriptLen := len(sigScript)
+			if sigScriptLen < dexdcr.SwapContractSize {
+				continue
+			}
+			// The spent contract is at the end of the signature
+			// script. Lop off the front half.
+			script := sigScript[sigScriptLen-dexdcr.SwapContractSize:]
+			_, _, _, sh, err := dexdcr.ExtractSwapDetails(script, dcr.chainParams)
+			if err != nil {
+				// This is not our script, but not necessarily
+				// a problem.
+				dcr.log.Tracef("Error encountered searching for the input that spends %v, "+
+					"extracting swap details from vin %d of %d. Probably not a problem: %v.",
+					spendTx.TxHash(), i, len(spendTx.TxIn), err)
+				continue
+			}
+			if bytes.Equal(sh[:], secretHash[:]) {
+				vin = i
+				break
+			}
+		}
+		if vin >= 0 {
+			_, height, err := dcr.wallet.GetBestBlock(dcr.ctx)
+			if err != nil {
+				return nil, err
+			}
+			confs := uint64(height - block.height)
+			hash := spendTx.TxHash()
+			if confs < requiredRedeemConfirms {
+				dcr.mempoolRedeemsMtx.Lock()
+				dcr.mempoolRedeems[secretHash] = &mempoolRedeem{txHash: hash, firstSeen: time.Now()}
+				dcr.mempoolRedeemsMtx.Unlock()
+			}
+			return &asset.ConfirmRedemptionStatus{
+				Confs:  confs,
+				Req:    requiredRedeemConfirms,
+				CoinID: toCoinID(&hash, uint32(vin)),
+			}, nil
+		}
+		dcr.log.Warnf("Contract coin %v spent by someone but not sure who.", redemption.Spends.Coin.ID())
+		// Incorrect, but we will be in a loop of erroring if we don't
+		// return something. We were unable to find the spender for some
+		// reason.
+
+		// May be still in the map if abandonTx failed.
+		deleteMempoolRedeem = true
+
+		return &asset.ConfirmRedemptionStatus{
+			Confs:  requiredRedeemConfirms,
+			Req:    requiredRedeemConfirms,
+			CoinID: coinID,
+		}, nil
+	}
+
+	// The contract has not yet been redeemed, but it seems the redeeming
+	// tx has disappeared. Assume the fee was too low at the time and it
+	// was eventually purged from the mempool. Attempt to redeem again with
+	// a currently reasonable fee.
+
+	form := &asset.RedeemForm{
+		Redemptions:   []*asset.Redemption{redemption},
+		FeeSuggestion: feeSuggestion,
+	}
+	_, coin, _, err := dcr.Redeem(form)
+	if err != nil {
+		return nil, fmt.Errorf("unable to re-redeem %s: %w", redemption.Spends.Coin.ID(), err)
+	}
+
+	coinID = coin.ID()
+	newRedeemHash, _, err := decodeCoinID(coinID)
+	if err != nil {
+		return nil, err
+	}
+
+	dcr.mempoolRedeemsMtx.Lock()
+	dcr.mempoolRedeems[secretHash] = &mempoolRedeem{txHash: *newRedeemHash, firstSeen: time.Now()}
+	dcr.mempoolRedeemsMtx.Unlock()
+
+	return &asset.ConfirmRedemptionStatus{
+		Confs:  0,
+		Req:    requiredRedeemConfirms,
+		CoinID: coinID,
+	}, nil
 }
--- a/client/asset/dcr/dcr_test.go
+++ b/client/asset/dcr/dcr_test.go
@@ -202,8 +202,7 @@ type tRPCClient struct {
 	signFunc       func(tx *wire.MsgTx) (*wire.MsgTx, bool, error)
 	privWIF        *dcrutil.WIF
 	privWIFErr     error
-	walletTx       *walletjson.GetTransactionResult
-	walletTxErr    error
+	walletTxFn     func() (*walletjson.GetTransactionResult, error)
 	lockErr        error
 	passErr        error
 	disconnected   bool
@@ -250,6 +249,7 @@ func (blockchain *tBlockchain) addRawTx(blockHeight int64, tx *wire.MsgTx) (*cha
 				PrevBlock: *prevBlockHash,
 				Height:    uint32(blockHeight),
 				VoteBits:  1,
+				Timestamp: time.Now(),
 			},
 		}
 	}
@@ -474,8 +474,8 @@ func (c *tRPCClient) DumpPrivKey(_ context.Context, address stdaddr.Address) (*d
 }
 
 func (c *tRPCClient) GetTransaction(_ context.Context, txHash *chainhash.Hash) (*walletjson.GetTransactionResult, error) {
-	if c.walletTx != nil || c.walletTxErr != nil {
-		return c.walletTx, c.walletTxErr
+	if c.walletTxFn != nil {
+		return c.walletTxFn()
 	}
 	return nil, dcrjson.NewRPCError(dcrjson.ErrRPCNoTxInfo, "no test transaction")
 }
@@ -1754,7 +1754,7 @@ func TestFindRedemption(t *testing.T) {
 	if err != nil {
 		t.Fatalf("error generating hex for contract tx: %v", err)
 	}
-	node.walletTx = &walletjson.GetTransactionResult{
+	walletTx := &walletjson.GetTransactionResult{
 		BlockHash:     blockHash.String(),
 		Confirmations: 1,
 		Details: []walletjson.GetTransactionDetailsResult{
@@ -1765,6 +1765,10 @@ func TestFindRedemption(t *testing.T) {
 			},
 		},
 		Hex: txHex,
+	}
+
+	node.walletTxFn = func() (*walletjson.GetTransactionResult, error) {
+		return walletTx, nil
 	}
 
 	// Add an intermediate block for good measure.
@@ -1795,12 +1799,16 @@ func TestFindRedemption(t *testing.T) {
 	}
 
 	// gettransaction error
-	node.walletTxErr = tErr
+	node.walletTxFn = func() (*walletjson.GetTransactionResult, error) {
+		return walletTx, tErr
+	}
 	_, _, err = wallet.FindRedemption(tCtx, coinID, nil)
 	if err == nil {
 		t.Fatalf("no error for gettransaction rpc error")
 	}
-	node.walletTxErr = nil
+	node.walletTxFn = func() (*walletjson.GetTransactionResult, error) {
+		return walletTx, nil
+	}
 
 	// getcfilterv2 error
 	node.rawErr[methodGetCFilterV2] = tErr
@@ -1838,12 +1846,12 @@ func TestFindRedemption(t *testing.T) {
 	redeemBlock.Transactions[0].TxIn[1].SignatureScript = redemptionScript
 
 	// Wrong script type for output
-	node.walletTx.Hex, _ = makeTxHex(inputs, []dex.Bytes{otherScript, otherScript})
+	walletTx.Hex, _ = makeTxHex(inputs, []dex.Bytes{otherScript, otherScript})
 	_, _, err = wallet.FindRedemption(tCtx, coinID, nil)
 	if err == nil {
 		t.Fatalf("no error for wrong script type")
 	}
-	node.walletTx.Hex = txHex
+	walletTx.Hex = txHex
 
 	// Sanity check to make sure it passes again.
 	_, _, err = wallet.FindRedemption(tCtx, coinID, nil)
@@ -2203,7 +2211,10 @@ func TestLookupTxOutput(t *testing.T) {
 
 	// gettransaction error
 	delete(node.txOutRes, op)
-	node.walletTxErr = tErr
+	walletTx := &walletjson.GetTransactionResult{}
+	node.walletTxFn = func() (*walletjson.GetTransactionResult, error) {
+		return walletTx, tErr
+	}
 	_, _, spent, err = wallet.lookupTxOutput(context.Background(), &op.txHash, op.vout)
 	if err == nil {
 		t.Fatalf("no error for gettransaction error")
@@ -2211,7 +2222,9 @@ func TestLookupTxOutput(t *testing.T) {
 	if spent != 0 {
 		t.Fatalf("spent is not 0 with gettransaction error")
 	}
-	node.walletTxErr = nil
+	node.walletTxFn = func() (*walletjson.GetTransactionResult, error) {
+		return walletTx, nil
+	}
 
 	// wallet.lookupTxOutput will check if the tx is confirmed, its hex
 	// is valid and contains an output at index 0, for the output to be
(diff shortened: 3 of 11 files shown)


client/asset/btc/btc.go
```diff
@@ -5089,3 +5091,71 @@ func deserializeMsgTx(r io.Reader) (*wire.MsgTx, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	utxo, _, err := btc.node.getTxOut(txHash, vout, pkScript, time.Now().Add(-ContractSearchLimit))
+	if err != nil {
+		return nil, fmt.Errorf("error finding unspent contract %s: %w", redemption.Spends.Coin.ID(), err)
+	}
+	if utxo == nil {
+		// TODO: Spent, but by who. Find the spending tx.
```
chappjc, line 5135: I actually ran into a similar situation with someone in production where their client has actually redeemed, but because of a dcrd timeout with dcrwallet, Core's call to Redeem returned an error, and Core didn't record the redeem that really did broadcast.  As a result it retried a redeem repeatedly only to fail because the contract output was already spent.

However in the case I'm addressing we wouldn't be trying to confirm a known redemption yet, so it would have to be caught in one of the other ticking checks (isRedeemable).  I recently made a comment in the bonds pr diff where we'd do this:  https://github.com/decred/dcrdex/pull/2036/commits/8e0761761256a94ca17f2a1b59450bd05aab1086#diff-8962b4c526755f0f98ed9f46ae05827e8edbc43ed5671ea4a2624bdf1617496cR1565

Anyway, finding our own (forgotten) redeem would of course be similar to the recent refund search like in https://github.com/decred/dcrdex/pull/2011/files
I had tentatively planned to do similar to that if no one else go to that first.  I suppose the same code would be used for this TODO in `ConfirmRedemption`

client/asset/btc/btc.go
```diff
@@ -5089,3 +5091,71 @@ func deserializeMsgTx(r io.Reader) (*wire.MsgTx, error) {
...
+			Confs:  reqiredRedeemConfirms,
+			Req:    reqiredRedeemConfirms,
+			CoinID: coinID,
+		}, nil
+	}
+
+	// The contract has not yet been redeemed, but it seems the redeeming
+	// tx has disappeared. Assume the fee was too low at the time and it
+	// was eventually purged from the mempool. Attempt to redeem again with
+	// a currently reasonable fee.
```
chappjc, line 5149: Nice.  This is the key thing solved in this PR.

client/asset/dcr/dcr.go
```diff
@@ -4527,3 +4529,124 @@ func float64PtrStr(v *float64) string {
...
+
+	tx, err := dcr.wallet.GetTransaction(dcr.ctx, txHash)
+	if err == nil {
+		spew.Dump(tx)
+		return &asset.ConfirmRedemptionStatus{
+			Confs:  uint64(tx.Confirmations),
+			Req:    reqiredRedeemConfirms,
+			CoinID: coinID,
+		}, nil
+	}
```
JoeGruffins, line 4547: It does not seem like dcr spv will ever let go of a bad transaction. I have one in my testnet node that I planted, and now am stuck here as it will always find the fake transaction even though it is spending a spent contract.

So, for dcr, I'm not sure we can ever lose a transaction that is not mined. It seems as if there could be the opposite problem and we hang on to an impossible transaction forever.

We could check if the contract is spent yet first.

But I wonder if this should either check the redeem's fee vs the current fee and/or how long the transaction has been sitting and we `abandontransaction` if it looks like it is being ignored by miners and send a new one.

JoeGruffins: Adding a map for dcr that will abandon transactions if they were first seen two hours ago and still have no confirmations.

client/asset/dcr/externaltx.go
```diff
@@ -328,7 +328,7 @@ func (dcr *ExchangeWallet) isOutputSpent(ctx context.Context, output *outputSpen
 	nextScanHeight := lastScannedHeight + 1
 
 	bestBlock := dcr.cachedBestBlock()
-	if nextScanHeight >= bestBlock.height {
+	if nextScanHeight > bestBlock.height {
```
JoeGruffins, line 331: This was awkward during testing and I think it was incorrect. It skipped the best block, which has one confirm, so transactions were not searched there. I do not see an comments about a reason for this.

chappjc: @itswisdomagain See if this is correct please.

itswisdomagain: Yeah, looks correct per the comment below this if statement. I think that was supposed to be and I think it should be changed to `if lastScannedHeight >= bestBlock.height`.

The `nextScanHeight := lastScannedHeight + 1` should be moved to after this if statement.

The idea, per the comment inside the if statement, is ...
```
if lastScannedHeight == bestBlock.height {
	// this is fine, just return, no more blocks to scan
	return ...
}
if lastScannedHeight > bestBlock.height {
	// not fine, how did it scan a height above the best height?
	// log a warning, should never happen anyways
	// return too, since there're obviously no more blocks to scan
	return ...
}
```

JoeGruffins: This look right? https://github.com/decred/dcrdex/compare/db62034bae674721bf653a8b666faf8e1d4b5ad3..53bd2f1384b8ba7c09ac0ff40302d7c1085de111

client/asset/btc/btc.go
```diff
@@ -5090,3 +5092,71 @@ func deserializeMsgTx(r io.Reader) (*wire.MsgTx, error) {
...
+	}
+
+	txHash, vout, err := decodeCoinID(redemption.Spends.Coin.ID())
+	if err != nil {
+		return nil, err
+	}
+
+	utxo, _, err := btc.node.getTxOut(txHash, vout, pkScript, time.Now().Add(-ContractSearchLimit))
+	if err != nil {
+		return nil, fmt.Errorf("error finding unspent contract %s: %w", redemption.Spends.Coin.ID(), err)
```
norwnd, line 5133: Just a nit: I noticed that here you are using `%w` to wrap this error while in other places you use `%v`, 

I'm curious whether you have some convention for this ? It seems both styles are used all across `client/asset/btc/btc.go` (I thought `%w` is usually preferred, unless you explicitly want to "disable" unwrapping for some reason)

JoeGruffins: Probably %w is always correct for errors. I am just in the habit of using %v. Will change the error %v's to %w's

chappjc: %w is almost always the "right thing to do" but we've had an annoying issue where we unwrap some errors recursively and only report the inner error ...  https://github.com/decred/dcrdex/issues/1488
One of the most annoying DEX (non)issues TBH.

JoeGruffins: v to w for errors https://github.com/decred/dcrdex/compare/53bd2f1384b8ba7c09ac0ff40302d7c1085de111..10f3413dbd2f2fdc612d6129e891a5792f2bb893

client/asset/dcr/dcr.go
```diff
@@ -4527,3 +4546,178 @@ func float64PtrStr(v *float64) string {
...
+	if !abandonTx && !errors.Is(err, asset.CoinNotFoundError) {
+		return nil, fmt.Errorf("problem searching for redemption transaction %s: %w", txHash, err)
+	}
+
+	// Redemption transaction is missing from the point of view of our node!
+	// Unlikely, but possible it was redeemed by another transaction. We
+	// assume a contract past its locktime cannot make it here, so it must
+	// not be refunded. Check if the contract is still an unspent output.
+
+	txHash, vout, err := decodeCoinID(redemption.Spends.Coin.ID())
```
buck54321, line 4614: `txHash` is used above for the redemption tx. How about `swapHash` or something here to differentiate? Same in **btc**.

client/asset/dcr/dcr.go
```diff
@@ -88,6 +88,8 @@ const (
 	// freshFeeAge is the expiry age for cached fee rates of external origin,
 	// past which fetchFeeFromOracle should be used to refresh the rate.
 	freshFeeAge = time.Minute
+
+	reqiredRedeemConfirms = 2
```
buck54321, line 92: Missing a `u`. req**u**iredRedeemConfirms

client/asset/dcr/dcr.go
```diff
@@ -4527,3 +4546,178 @@ func float64PtrStr(v *float64) string {
...
+				_, height, err := dcr.wallet.GetBestBlock(dcr.ctx)
+				if err != nil {
+					return nil, err
+				}
+				hash := spendTx.TxHash()
+				return &asset.ConfirmRedemptionStatus{
+					Confs:  uint64(height - block.height),
+					Req:    reqiredRedeemConfirms,
+					CoinID: toCoinID(&hash, uint32(vin)),
+				}, nil
```
buck54321, line 4679: If `Confs` is `< reqiredRedeemConfirms` should we add this redemption to `mempoolRedeems` so we can avoid future calls to `externalTxOutput`?

client/asset/dcr/dcr.go
```diff
@@ -4527,3 +4546,178 @@ func float64PtrStr(v *float64) string {
...
+			}
+		}
+		dcr.log.Warnf("Contract coin %v spent by someone but not sure who.", redemption.Spends.Coin.ID())
```
buck54321, line 4682: Any point in examining the output address(es) for the tx to see if it's us?

client/asset/dcr/dcr_test.go
```diff
@@ -182,38 +182,39 @@ func signFunc(msgTx *wire.MsgTx, scriptSize int) (*wire.MsgTx, bool, error) {
...
+	passErr               error
+	disconnected          bool
+	rawRes                map[string]json.RawMessage
+	rawErr                map[string]error
+	blockchain            *tBlockchain
+	lluCoins              []walletjson.ListUnspentResult // Returned from ListLockUnspent
+	lockedCoins           []*wire.OutPoint               // Last submitted to LockUnspent
+	listLockedErr         error
+	estFeeErr             error
+	abandonTransactionErr error
```
buck54321, line 217: How about `abandonTxErr` so we don't blow up the diff?

client/asset/dcr/dcr.go
chappjc: I know you're asking @JoeGruffins, but I have this question some thought in context of the recently-added refund detection code (https://github.com/decred/dcrdex/pull/2011). Ultimately, I decided the client primarily needs a path forward instead of retrying and failing forever.  If we did detect it was paid to address X, we could log it, maybe even try to see if wallet owns the address, but aside from detecting if we were hacked, I didn't see it being helpful.  User can bring up the tx on an explorer, so maybe we log the located refund txid?

buck54321: Good points. A good log message is probably the best we can do. This will probably never happen anyway. No change needed.

JoeGruffins:
Just rebased.

client/asset/btc/btc.go
```diff
@@ -5090,3 +5092,71 @@ func deserializeMsgTx(r io.Reader) (*wire.MsgTx, error) {
...
+			CoinID: coinID,
+		}, nil
+	}
+
+	// The contract has not yet been redeemed, but it seems the redeeming
+	// tx has disappeared. Assume the fee was too low at the time and it
+	// was eventually purged from the mempool. Attempt to redeem again with
+	// a currently reasonable fee.
+
+	form := &asset.RedeemForm{Redemptions: []*asset.Redemption{redemption}}
```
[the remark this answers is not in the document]
JoeGruffins: Adding a fee rate argument.

client/asset/dcr/dcr.go
JoeGruffins: Changed to `swapHash` in both places.

client/asset/dcr/dcr.go
JoeGruffins: Corrected.

client/asset/dcr/dcr.go
JoeGruffins: Good idea, done.

client/asset/dcr/dcr.go
JoeGruffins: Leaving as is for now.

client/asset/dcr/dcr_test.go
JoeGruffins: shortened

JoeGruffins:
buck's review changes https://github.com/decred/dcrdex/compare/570ea76accfebd5203d28b86a3b1242e473c7121..d9084f18b7795eddd92fa5afecd0c82b111d9d1a

client/asset/dcr/dcr.go
```diff
@@ -4521,3 +4540,190 @@ func float64PtrStr(v *float64) string {
...
+	var abandonTx bool
+	tx, err := dcr.wallet.GetTransaction(dcr.ctx, txHash)
+	if err == nil {
+		if have && mRedeem.txHash == *txHash {
+			if tx.Confirmations == 0 && time.Now().After(mRedeem.firstSeen.Add(maxRedeemMempoolAge)) {
+				// Transaction has been sitting in the mempool
+				// for a long time now. Attempt to abandon.
+				abandonTx = true
+				if err := dcr.wallet.AbandonTransaction(dcr.ctx, txHash); err != nil {
+					dcr.log.Warnf("Unable to abandon redemption mempool transaction %v: %v.", txHash, err)
```
JoeGruffins, line 4574: %w was not correct for errors that are logged.

client/core/trade.go
```diff
@@ -2854,8 +2854,14 @@ func (t *trackedTrade) confirmRedemption(match *matchTracker) {
...
-		return
+
+	// In some cases the wallet will need to send a new redeem transaction.
+	didUnlock, err := t.wallets.toWallet.refreshUnlock()
+	if err != nil { // Just log it and try anyway.
+		t.dc.log.Errorf("refreshUnlock error checking redeem %s: %v", t.wallets.toWallet.Symbol, err)
+	}
+	if didUnlock {
+		t.dc.log.Warnf("Unexpected unlock needed for the %s wallet to check a redemption", t.wallets.toWallet.Symbol)
 	}
```
JoeGruffins, line 2865: Since this may need to create a new redemption, for eth too, I think we need this check.

client/core/trade.go
```diff
@@ -2867,10 +2873,24 @@ func (t *trackedTrade) confirmRedemption(match *matchTracker) {
...
+	// one if we don't have one cached.
+	var feeSuggestion uint64
+	if _, is := t.accountRedeemer(); is {
+		feeSuggestion = t.metaData.RedeemMaxFeeRate
+	} else {
+		feeSuggestion = t.redeemFeeSuggestion.get()
+	}
+	if feeSuggestion == 0 {
+		feeSuggestion = t.dc.bestBookFeeSuggestion(t.wallets.toWallet.AssetID)
+	}
```
norwnd, line 2888: seems like copy-paste ? maybe wrap it in helper **func** like `trackedTrade. redeemFeeSuggestion()` ?

JoeGruffins: I is. Sure. I think used in 3 places so good idea.

JoeGruffins: or 2... added redemptionFee method https://github.com/decred/dcrdex/compare/d9084f18b7795eddd92fa5afecd0c82b111d9d1a..b946db78fa3409c28bd67c3b2df81498f5ed8e9e

client/asset/dcr/dcr.go
```diff
@@ -4521,3 +4540,190 @@ func float64PtrStr(v *float64) string {
...
+	swapHash, vout, err := decodeCoinID(redemption.Spends.Coin.ID())
+	if err != nil {
+		return nil, err
+	}
+
+	var spent bool
+	_, err = dcr.wallet.UnspentOutput(dcr.ctx, swapHash, vout, 0)
+	if err != nil {
+		if errors.Is(err, asset.CoinNotFoundError) {
+			spent = true
```
chappjc, line 4617: The naming of `spent` here is misleading here since in SPV mode we can't know it's spent based on `UnspentOutput`.  It's more like `maybeSpent`.  In SPV mode, it's _expected_ to not know if the contract is spent until you do a cfilters scan.  See the weak guarantees documented for `UnspentOutput` in the `asset.Wallet` interface docs.

Also, you do follow this with a block filters scan to determine, but why don't we use `lookupTxOutput` instead of using `UnspentOutput` directly before the scan like in the other places we do this check like in `refundTx`?  It's a bit smarter when not in spv mode.

client/asset/dcr/dcr.go
```diff
@@ -4521,3 +4540,190 @@ func float64PtrStr(v *float64) string {
...
+			}
+			return &asset.ConfirmRedemptionStatus{
+				Confs:  uint64(tx.Confirmations),
+				Req:    requiredRedeemConfirms,
+				CoinID: coinID,
+			}, nil
+		}
+	}
+
+	if !abandonTx && !errors.Is(err, asset.CoinNotFoundError) {
```
chappjc, line 4599: It's a bit challenging to reason out if it's possible to have a `nil` `err` and `abantonTx=false`.  Can that happen?
I keep thinking that before this it should do:

```go
if err == nil {
    if have && ...
    ...
} else if !errors.Is(err, asset.CoinNotFoundError) {
    return ...
}

// We are here if the tx not found, or found but abandoning...
````
?

File: client/asset/dcr/dcr.go [hunk 17]
```diff
@@ -4521,3 +4540,190 @@ func float64PtrStr(v *float64) string {
...
+					dcr.mempoolRedeemsMtx.Unlock()
+				}
+				return &asset.ConfirmRedemptionStatus{
+					Confs:  confs,
+					Req:    requiredRedeemConfirms,
+					CoinID: toCoinID(&hash, uint32(vin)),
+				}, nil
+			}
+		}
+		dcr.log.Warnf("Contract coin %v spent by someone but not sure who.", redemption.Spends.Coin.ID())
```
chappjc, line 4685: You'd also be here if `isOutputSpent` returned `spent = false`... meaning the cfilters scan just failed to fine the spender (it's actually unspent).

chappjc (changes requested):
We seem to be drawing incorrect conclusions about the returns from `UnspentOutput`.  If it finds an output it is definitely unspent; if it does not find it, it's spend status is *unknown* and we must scan to know spent or not.  If the filters scan then says **un**spent, that's not incorrect or contradicting `UnspentOutput`, it's the real result.

JoeGruffins:
Just rebased.

client/asset/dcr/dcr.go
```diff
@@ -102,6 +104,8 @@ var (
 	conventionalConversionFactor = float64(dexdcr.UnitInfo.Conventional.ConversionFactor)
 	walletBlockAllowance         = time.Second * 10
 
+	maxRedeemMempoolAge = time.Hour * 2
```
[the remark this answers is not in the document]
JoeGruffins: Added a comment.

client/asset/dcr/dcr.go
```diff
@@ -569,6 +578,10 @@ type ExchangeWallet struct {
 
 	externalTxMtx   sync.RWMutex
 	externalTxCache map[chainhash.Hash]*externalTx
+
+	// TODO: Consider persisting mempool redeems on file.
+	mempoolRedeemsMtx sync.RWMutex
+	mempoolRedeems    map[[32]byte]*mempoolRedeem
```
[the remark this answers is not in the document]
JoeGruffins: Added the doc.

client/asset/dcr/dcr.go
```diff
@@ -4521,3 +4540,190 @@ func float64PtrStr(v *float64) string {
 	}
 	return strconv.FormatFloat(*v, 'f', 8, 64)
 }
+
+func (dcr *ExchangeWallet) ConfirmRedemption(coinID dex.Bytes, redemption *asset.Redemption, feeSuggestion uint64) (*asset.ConfirmRedemptionStatus, error) {
```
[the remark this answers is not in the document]
JoeGruffins: Added a doc.

client/asset/dcr/dcr.go
```diff
@@ -4521,3 +4540,190 @@ func float64PtrStr(v *float64) string {
...
+				if err := dcr.wallet.AbandonTransaction(dcr.ctx, txHash); err != nil {
+					dcr.log.Warnf("Unable to abandon redemption mempool transaction %v: %v.", txHash, err)
+				} else {
+					dcr.log.Infof("Abandoned redemption mempool transaction: %v.", txHash)
+					deleteMempoolRedeem = true
+				}
+			}
+		} else {
+			// Will hit this if node was restarted with an actively
+			// redeeming swap.
```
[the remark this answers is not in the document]
JoeGruffins: I mean dexc, yes. Correcting the doc.

The `mempoolRedeems` are not persisted and are added in `Redeem` so restarting after starting the redeem process would remove them.

client/asset/dcr/dcr.go
JoeGruffins: Using `lookupTxOutput` now.

client/asset/dcr/dcr.go
JoeGruffins: I'm not sure what I was thinking here. I moved `!errors.Is(err, asset.CoinNotFoundError)` up because if it is not that error we should fail anyway.

client/asset/dcr/dcr.go
JoeGruffins: Now if `isOutputSpent` return false with no error, breaking to the bottom and sending a new tx

JoeGruffins:
Now trusting the cfilters scan https://github.com/decred/dcrdex/compare/38be0018222f04592e9e964b27980deb94897f25..4160ae204bc60e9dceaf37dfb54aae553bf85861

client/asset/dcr/dcr.go
```diff
@@ -4910,3 +4934,199 @@ func float64PtrStr(v *float64) string {
...
+			}
+		}
+		if vin >= 0 {
+			if spentStatus == -1 {
+				_, height, err := dcr.wallet.GetBestBlock(dcr.ctx)
+				if err != nil {
+					return nil, err
+				}
+				confs = uint64(height - block.height)
+			}
```
JoeGruffins, line 5075: Wait, that can't be right. These are the swap tx confirms.

JoeGruffins:
Use correct confirms https://github.com/decred/dcrdex/compare/4160ae204bc60e9dceaf37dfb54aae553bf85861..a70b12c14044dfd1b46e04dd03484629e7c5096c

client/asset/dcr/dcr_test.go
```diff
@@ -2779,3 +2790,214 @@ func TestEstimateSendTxFee(t *testing.T) {
...
+		txRes:          txFn([]bool{false}),
+		abandonTxErr:   errors.New(""),
+		mempoolRedeems: map[[32]byte]*mempoolRedeem{secretHash: {txHash: txHash, firstSeen: time.Now().Add(-maxRedeemMempoolAge - time.Second)}},
+		txOutRes:       map[outPoint]*chainjson.GetTxOutResult{newOutPoint(&txHash, 0): makeGetTxOutRes(0, 5, nil)},
+	}, {
+		name:       "ok and spent",
+		coinID:     coinID,
+		txRes:      txFn([]bool{true, false}),
+		redemption: redemption,
+		wantConfs:  1, // one confirm becasue this tx is in the best block
```
JoeGruffins, line 2926: because

chappjc:
Priority feature for 1.0.

JoeGruffins:
just rebased and fixed a typo

client/asset/dcr/dcr.go
```diff
@@ -4908,3 +4932,196 @@ func float64PtrStr(v *float64) string {
...
+				// for a long time now. Attempt to abandon.
+				abandonTx = true
+				if err := dcr.wallet.AbandonTransaction(dcr.ctx, txHash); err != nil {
+					dcr.log.Warnf("Unable to abandon redemption mempool transaction %v: %v.", txHash, err)
+				} else {
+					dcr.log.Infof("Abandoned redemption mempool transaction: %v.", txHash)
+					deleteMempoolRedeem = true
+				}
+			}
+		} else {
```
buck54321, line 4986: `else if !have`?

client/asset/dcr/rpcwallet.go
```diff
@@ -890,6 +891,16 @@ func (w *rpcWallet) rpcClientRawRequest(ctx context.Context, method string, args
 	return nil
 }
 
+// AbandonTransaction attempts to abandon a mempool transaction.
+// Part of the Wallet interface.
+func (w *rpcWallet) AbandonTransaction(ctx context.Context, hash *chainhash.Hash) error {
+	err := w.rpcClientRawRequest(ctx, methodAbandonTransaction, anylist{hash.String()}, nil)
```
chappjc, line 897: Hopefully this won't be a problem:

```
abandontransaction "hash"

Remove an unconfirmed transaction and all dependent transactions
```

Say the redeem created a UXTO that while unconfirmed was then selected to fund a new swap...
Detecting that might be a challenge though.

client/asset/dcr/dcr_test.go
```diff
@@ -1766,6 +1770,10 @@ func TestFindRedemption(t *testing.T) {
 		},
 		Hex: txHex,
 	}
+	walletTxErr := new(error)
```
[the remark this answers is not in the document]
JoeGruffins: So later we can set the value inside the function without rewriting the entire function.

But since it is confusing will set the entire function.

The reason for changing to a function is because the new method I added will need two or more walletTx there so this way tests can produce different txn based on some logic.

client/asset/dcr/dcr.go
JoeGruffins: If `have` is true here, I think it is an error. Core has given this method a different redeem for this secret hash. It is this method's job to assign a new transaction so something is off. Perhaps the method was run concurrently for the same redeem and this is the result of a race if hit. Adding an error return here so we do not erase what the first call to this method must have done.

client/asset/dcr/rpcwallet.go
JoeGruffins: I can remove this if you would like as it is not really part of the issue, I guess it's a separate issue but something I have hit by trying to redeem dust transactions with dcrwallet in spv mode. Then the wallet uses those same inputs for something else. But then it still keeps the bad transaction around in its db.

Or I could be misremembering.

Maybe this is a bug in dcrwallet that it keeps the tx around even after the inputs were spent by something else. I think I will just remove abandoning here, try to discern if it is a bug in dcrwallet, and try to get it fixed there if possible.

Incidentally, the vspd fee payment logic has AbandonTransaction as part of its wallet interface, so may have the same worry there.

JoeGruffins:
https://github.com/decred/dcrdex/compare/7b4d93d04f02584c39ddae9ff7baad167ede33f5..dc96c6466c669306bd604983df7583217ea63326

buck fixes a couple more pushes though hold

JoeGruffins:
just rebased

JoeGruffins:
Removed abandon transaction for now, probably needs a separate issue and may be a dcrwallet bug that I was seeing. https://github.com/decred/dcrdex/compare/4059a0e1b331e280a0cdd55a48068ea2f00d1af6..78f763286de1192acf5fcc6bab092d618149716f

JoeGruffins:
fixing doc https://github.com/decred/dcrdex/compare/78f763286de1192acf5fcc6bab092d618149716f..13c04a34166b9f4ba0217285af84d41e6ac61f5f

client/asset/dcr/dcr.go
```diff
@@ -4909,3 +4933,200 @@ func float64PtrStr(v *float64) string {
 	}
 	return strconv.FormatFloat(*v, 'f', 8, 64)
 }
+
+// ConfirmRedemption returns how many confirmations a redemption has. Normally
+// this is very straitforward. However there are two situations that have come
+// up that this also handles. One is when the wallet can not find the redemption
+// transaction. This is most likely because the fee was set too low and the tx
+// was removed from the mempool. In the case where it is not found, this will
+// send a new fee tx using the provided fee suggestion. The second situation
```
martonp, line 4942: "new fee tx" -> "new tx"

client/asset/dcr/dcr.go
```diff
@@ -4909,3 +4933,200 @@ func float64PtrStr(v *float64) string {
...
+		}
+	}()
+
+	tx, err := dcr.wallet.GetTransaction(dcr.ctx, txHash)
+	if err != nil && !errors.Is(err, asset.CoinNotFoundError) {
+		return nil, fmt.Errorf("problem searching for redemption transaction %s: %w", txHash, err)
+	}
+	if err == nil {
+		if have && mRedeem.txHash == *txHash {
+			if tx.Confirmations == 0 && time.Now().After(mRedeem.firstSeen.Add(maxRedeemMempoolAge)) {
```
martonp, line 4977: In this case, shouldn't we check to see if the contract output is still unspent, and resubmitting if it isn't, instead of just returning 0 confirmations?

client/asset/btc/btc.go
```diff
@@ -5792,3 +5794,74 @@ func deserializeMsgTx(r io.Reader) (*wire.MsgTx, error) {
...
 	return deserializeMsgTx(bytes.NewReader(txB))
 }
+
+func (btc *baseWallet) ConfirmRedemption(coinID dex.Bytes, redemption *asset.Redemption, feeSuggestion uint64) (*asset.ConfirmRedemptionStatus, error) {
+	txHash, _, err := decodeCoinID(coinID)
+	if err != nil {
+		return nil, err
+	}
+
+	_, confs, err := btc.rawWalletTx(txHash)
```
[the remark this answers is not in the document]
chappjc: > to see if the fee needs to be bumped

Maybe could treat the redeem tx, which has a single output, as a transaction comprising a single change output.

I'm not excited about having Core start tracking a new set of "acceleration coins" for the redeem though.

We can make a TODO here unless it's simpler than I think.

However, if it doesn't get purged from mempool, I think it's alright.  The user just has to wait longer.  Still, there's a subtle possible vuln if you are the maker because your redeem really shouldn't hang around unmined after the taker's contract has expired.  But we've survived without this so far.

client/asset/btc/btc.go
```diff
@@ -5792,3 +5794,74 @@ func deserializeMsgTx(r io.Reader) (*wire.MsgTx, error) {
...
+		return nil, err
+	}
+
+	utxo, _, err := btc.node.getTxOut(swapHash, vout, pkScript, time.Now().Add(-ContractSearchLimit))
+	if err != nil {
+		return nil, fmt.Errorf("error finding unspent contract %s: %w", redemption.Spends.Coin.ID(), err)
+	}
+	if utxo == nil {
+		// TODO: Spent, but by who. Find the spending tx.
+		btc.log.Warnf("Contract coin %v spent by someone but not sure who.", redemption.Spends.Coin.ID())
```
chappjc, line 5839: Let's use `swapHash, vout` in this log and the two `error`s that use it.

chappjc (changes requested):

client/asset/dcr/dcr.go
chappjc: We just found the spending transaction (the redeem) here.  Do you mean go look for a different mined spender of the contract output as we do in the flow later in this function in the case where `dcr.wallet.GetTransaction` could not find the redeem?

The wallet should periodically rebroadcast the unconfirmed redeem that it initially broadcasted.  Abandoning it and making an entirely new transaction is fairly perilous since the outputs from the initial redeem may be used already.

client/asset/btc/btc.go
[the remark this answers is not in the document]
JoeGruffins: I don't think neutrino has a purge transaction option like dcrwallet spv does either. So even if you thought a tx needed to go I don't think there's anything you can do about it?

client/asset/dcr/dcr.go
```diff
@@ -4909,3 +4933,200 @@ func float64PtrStr(v *float64) string {
...
+
+// ConfirmRedemption returns how many confirmations a redemption has. Normally
+// this is very straitforward. However there are two situations that have come
+// up that this also handles. One is when the wallet can not find the redemption
+// transaction. This is most likely because the fee was set too low and the tx
+// was removed from the mempool. In the case where it is not found, this will
+// send a new fee tx using the provided fee suggestion. The second situation
+// this watches for is a transaction that we can find but has been sitting in
+// the mempool for a long time. This has been observed with the wallet in SPV
+// mode and the transaction inputs having been spent by another transaction. The
```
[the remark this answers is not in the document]
JoeGruffins: I have not observed it so far, but I do test with dcr and other coins more than btc... because the testnet btc block times are longer.

client/asset/dcr/dcr.go
JoeGruffins: Resubmit the same transaction?

client/asset/btc/btc.go
```diff
@@ -5792,3 +5794,74 @@ func deserializeMsgTx(r io.Reader) (*wire.MsgTx, error) {
 func msgTxFromBytes(txB []byte) (*wire.MsgTx, error) {
 	return deserializeMsgTx(bytes.NewReader(txB))
 }
+
+func (btc *baseWallet) ConfirmRedemption(coinID dex.Bytes, redemption *asset.Redemption, feeSuggestion uint64) (*asset.ConfirmRedemptionStatus, error) {
```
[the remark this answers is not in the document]
JoeGruffins: Oh I forgot this in both assets, adding.

client/asset/dcr/dcr.go
chappjc: > Resubmit the same transaction?

It would surely just say "already have transaction" or similar.  `GetTransaction` just reported it.

client/asset/dcr/dcr.go
```diff
@@ -662,6 +680,7 @@ var _ asset.LiveReconfigurer = (*ExchangeWallet)(nil)
 var _ asset.TxFeeEstimator = (*ExchangeWallet)(nil)
 var _ asset.Bonder = (*ExchangeWallet)(nil)
 var _ asset.Authenticator = (*ExchangeWallet)(nil)
+var _ asset.RedemptionConfirmer = (*ExchangeWallet)(nil)
```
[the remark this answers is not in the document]
JoeGruffins: Oh yeah! Great! Doing that. The wallet traits are just a runtime thing right? They don't need to keep order do they?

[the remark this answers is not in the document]
chappjc: Just check the values on the frontend

client/asset/dcr/dcr.go
JoeGruffins: Fixed.

client/asset/btc/btc.go
```diff
@@ -82,6 +82,8 @@ const (
 	// testnetExternalApiUrl is the URL of the testnet external API in case of
 	// fallback.
 	testnetExternalApiUrl = "https://mempool.space/testnet/api/"
+
+	requiredRedeemConfirms = 2
```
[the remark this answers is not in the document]
JoeGruffins: Setting to one for btc and adding docs.

client/core/trade.go
JoeGruffins: Adding. For the peerCount, the xcWallet has this field but should it be checked live or good enough to use the field?

chappjc: Current field value is good.

client/asset/btc/btc.go
```diff
@@ -5792,3 +5794,74 @@ func deserializeMsgTx(r io.Reader) (*wire.MsgTx, error) {
...
+		}, nil
+	}
+
+	if !errors.Is(err, WalletTransactionNotFound) {
+		return nil, fmt.Errorf("problem searching for redemption transaction %s: %w", txHash, err)
+	}
+
+	// Redemption transaction is missing from the point of view of our node!
+	// Unlikely, but possible it was redeemed by another transaction. We
+	// assume a contract past its locktime cannot make it here, so it must
```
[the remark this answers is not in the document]
JoeGruffins: Removing.

client/asset/btc/btc.go
JoeGruffins: Adding them to the log and two error returns following obtaining it.

JoeGruffins:
I think covering all recent comments. Most notably removing the `RedemptionConfirmer` interface and adding that to the main interface. https://github.com/decred/dcrdex/compare/13c04a34166b9f4ba0217285af84d41e6ac61f5f..0819a9455112cde1c7f836c0a8d89d6117844334

client/asset/interface.go
```diff
@@ -16,19 +16,18 @@ import (
...
-	WalletTraitRescanner           WalletTrait = 1 << iota // The Wallet is an asset.Rescanner.
-	WalletTraitNewAddresser                                // The Wallet can generate new addresses on demand with NewAddress.
-	WalletTraitLogFiler                                    // The Wallet allows for downloading of a log file.
-	WalletTraitFeeRater                                    // Wallet can provide a fee rate for non-critical transactions
-	WalletTraitAccelerator                                 // This wallet can accelerate transactions using the CPFP technique
-	WalletTraitRecoverer                                   // The wallet is an asset.Recoverer.
-	WalletTraitWithdrawer                                  // The Wallet can withdraw a specific amount from an exchange wallet.
-	WalletTraitSweeper                                     // The Wallet can sweep all the funds, leaving no change.
-	WalletTraitRestorer                                    // The wallet is an asset.WalletRestorer
-	WalletTraitRedemptionConfirmer                         // The wallet has a process to confirm a redemption.
```
JoeGruffins, line 28: These do not need to be kept in order, correct?

client/core/trade.go
```diff
@@ -2844,24 +2830,50 @@ func (c *Core) sendRedeemAsync(t *trackedTrade, match *matchTracker, coinID, sec
...
+	if !toWallet.synchronized() {
+		t.dc.log.Errorf("Unable to confirm redemption. %s still syncing.", unbip(toWallet.AssetID))
 		return
 	}
-	if !t.wallets.toWallet.connected() {
+
+	didUnlock, err := toWallet.refreshUnlock()
+	if err != nil { // Just log it and try anyway.
+		t.dc.log.Errorf("refreshUnlock error checking redeem %s: %v", t.wallets.toWallet.Symbol, err)
 		return
```
JoeGruffins, line 2872: This return is new. I think correct to return if unable to unlock? Maybe not necessary though if the redeem only needs to check confirms and not send a new transaction, which is most of the time. But I think wallet is expected to be unlocked with active trades anyhow.

client/asset/interface.go
chappjc: They do not afaik, but I mean check and adjust the frontend where they are mirrored.

client/core/trade.go
chappjc: Why not let the action requiring an unlocked wallet throw the error then?

client/asset/interface.go
JoeGruffins: Ok sorry, missed that. Will do.

JoeGruffins:
If the RedemptionConfirmer was 9 then the two after that need to come down two. Hopefully this is what was needed in the frontend https://github.com/decred/dcrdex/compare/0819a9455112cde1c7f836c0a8d89d6117844334..ec9b60f984a62538efe4b35bfb91d0539b35ffd8

client/core/trade.go
JoeGruffins: Ok. Will remove the return then.

chappjc: Ok.  I mean you could still try to `refreshUnlock` here but if there's something downstream that really needed it to succeed, it can bomb.

JoeGruffins:
Oh yeah it has the doc `// Just log it and try anyway.`

Not returning on unlock error. https://github.com/decred/dcrdex/compare/ec9b60f984a62538efe4b35bfb91d0539b35ffd8..f8022cad997135328e28e73cd0730c0045a87cb5

client/asset/dcr/dcr.go
chappjc: > it says that if the transaction has been sitting in the mempool for a long time, it means that the inputs have been spent by another transaction

That has to be an insane wallet bug.  Sitting in mempool for a long time does not in general mean the inputs were double spent.

> Which basically means another transaction redeemed the swap somehow.  In this case shouldn't we check to see if the contract output is no longer unspent, and if so, return that the redemption is confirmed?

"no longer unspent" doesn't make sense to me.  It will definitely not be unspent the second the redeem tx enters mempool via the wallet.

client/asset/btc/btc.go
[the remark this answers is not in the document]
chappjc: So a kvdb for dcr to track a chain of accelerations of the redeem.  I guess so.  Feels messy and error prone.

client/asset/dcr/dcr.go
martonp: This is what's in the comment:

```
The second situation
// this watches for is a transaction that we can find but has been sitting in
// the mempool for a long time. This has been observed with the wallet in SPV
// mode and the transaction inputs having been spent by another transaction. 
```

chappjc: I don't believe that comment is correct in general.  What I recall is that Joe hit a pathological case where the core startup match resolution actions fired of a refund or redeem before the wallet was synced, tricking wallet to double spend.  We've since resolved this with synchronized checks

chappjc: > If it's not this strange wallet issue, then the reason the tx has been sitting in the mempool for 2 hours must be that the fee is too low. In which case I think we should resubmit with a higher fee.

What do you mean be "resubmit with a higher fee"?  That sounds like RBF, but a Decred node will not relay a double spend, and the wallet will not publish the transaction either (it only got to that failure mode for the reason I described in my previous comment).

JoeGruffins: Ok to punt on this for now? Maybe whatever I experienced is fixed. Maybe better to go further down this road if we can reliably reproduce something.

client/asset/btc/btc.go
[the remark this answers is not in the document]
JoeGruffins: Does the accelerator drop the old tx or does the wallet handle that? Need to look.

client/asset/interface.go
```diff
@@ -593,11 +598,6 @@ type LogFiler interface {
 // Will also error if the secret in the contractData is not found in the
 // transaction secrets.
 type DynamicSwapOrRedemptionFeeChecker interface {
```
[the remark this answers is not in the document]
JoeGruffins: Will rename.

client/asset/btc/btc.go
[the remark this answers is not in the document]
chappjc: It's doesn't drop or replace.  It spends the change. CPFP.
This is what I was getting at a few comments above.

JoeGruffins:
DynamicSwapOrRedemptionFeeChecker -> DynamicSwapper

https://github.com/decred/dcrdex/compare/f8022cad997135328e28e73cd0730c0045a87cb5..b3965a95e4e4e8b8e1d3b52a4de8232faa2e1954

client/asset/btc/btc.go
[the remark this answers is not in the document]
martonp: I was proposing to do RBF. I think the server might have to be updated for this to start searching for the replacement tx instead of the original one though.

> So a kvdb for dcr to track a chain of accelerations of the redeem. I guess so. Feels messy and error prone.

I agree, it's kind of annoying in ETH, but probably worth it since BTC can have fee spikes. Since it also could require server changes I think it can be done later as a separate piece of work.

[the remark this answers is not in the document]
martonp: ETH is doing RBF/replace nonce, not CPFP. Couldn't we do the same thing here?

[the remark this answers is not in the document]
chappjc: RBF is not a thing in Decred.  And in BTC it's not trivial (need to fiddle with sequence values) or without consequences (new txid and any spenders of the replaced 0-conf UTXOs are hosed so we have to prevent that like our lives depend on it).

client/asset/dcr/dcr.go
chappjc: > Ok to punt on this for now? Maybe whatever I experienced is fixed. Maybe better to go further down this road if we can reliably reproduce something.

Yes, I think so.  I think it was an edge case, but even if we want to consider abandoning, that's just a wallet concept and the rest of the network probably already has the initial tx (unless it was this wallet double spend bug) and a new one that tries to spend the already-spent contract output will be rejected by the network.

client/asset/btc/btc.go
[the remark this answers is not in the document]
JoeGruffins: `CPFP` oh yeah it ups the entire fee of the chain of txn. I remember now.

client/asset/btc/btc.go
```diff
@@ -5792,3 +5797,82 @@ func deserializeMsgTx(r io.Reader) (*wire.MsgTx, error) {
...
+		}, nil
+	}
+
+	if !errors.Is(err, WalletTransactionNotFound) {
+		return nil, fmt.Errorf("problem searching for redemption transaction %s: %w", txHash, err)
+	}
+
+	// Redemption transaction is missing from the point of view of our node!
+	// Unlikely, but possible it was redeemed by another transaction. Check
+	// if the contract is still an unspent output.
```
chappjc, line 5832: For both DCR and BTC, in the case where the wallet/node gives a result for the transaction, and we declare "mempool for a long amount of time" (the case from about 15 lines above), it could jump down here to attempt to locate a different spender.

If it found a different spender, that would really be indicative of a wallet bug because the wallet should automatically abandon/remove a transaction if its inputs are spent, but it's something we might want to try in the future, for robustness.

Regarding doing an explicit `AbandonTransaction` (with DCR) or otherwise attempting to replace the original, I still maintain that is dangerous because there is a chance that:

1. It is still on the network waiting to be mined, and any replacement may be rejected by the network on broadcast
2. The outputs of the original redeem may be spent, and abandoning/replacing the redeem will have adverse consequences for those 0-conf spenders
3. The redeem may **actually be mined** and our node/wallet could just have lost connection or gotten stuck or corrupted or whatever, and we're falsely concluding it's not mined.  We could be creating tons of unnecessary issues for ourselves if we start abanonding, replacing, trying to bump with cpfp, etc.  A restart or rescan of the wallet may very well be the best and only real resolution.

client/asset/dcr/externaltx.go
```diff
@@ -328,13 +328,13 @@ func (dcr *ExchangeWallet) isOutputSpent(ctx context.Context, output *outputSpen
...
 		// When a transaction is initially found in a block, that block is also
 		// scanned for spending transactions (see findTxInBlock) as an
 		// optimization since the block was pulled. As such, it is normal for
 		// the lastScannedHeight for newly located transactions to be the best
 		// block, so there is no need to warn, only return.
-		if nextScanHeight > bestBlock.height+1 {
+		if lastScannedHeight > bestBlock.height {
 			dcr.log.Warnf("Attempted to look for output spender in block %d but current tip is %d!",
 				nextScanHeight, bestBlock.height)
 		}
```
chappjc, line 340: This new logic looks right, although the pseudocode from Wisdom still reads easier to me even though there are separate `return`s

```go
if lastScannedHeight == bestBlock.height {
	// this is fine, just return, no more blocks to scan
	return false, nil
}
if lastScannedHeight > bestBlock.height {
	// not fine, how did it scan a height above the best height?
	// log a warning, should never happen anyways
	dcr.log.Warn("shouldnt happen asdf")
	// return too, since there're obviously no more blocks to scan
	return false, nil
}
```

chappjc (approved):
All looks fine to me.  Just wanted to take the opportunity to comment one more time about the meaning of the "in mempool for hours" scenario, and what we could and shouldn't do in that case.

chappjc:
Will merge as soon as the interface conflict is resolved.

client/asset/dcr/externaltx.go
JoeGruffins: Will change to the more readable code here.

JoeGruffins:
Make if more readable https://github.com/decred/dcrdex/compare/ad9d460f5e0865e33c88467660aa690aedc0dce7..52b2fff0fc3bd69f857dfb31c7f4a35e8a294201

PR #2050, opened 2023-01-14
eth: add gas estimate utility

buck54321:
Add a program to get gas estimates for ETH or tokens.


chappjc:
Great!  There's no way to move the `GetGasEstimates` function (and the other non-product code like `runSimnetMiner` and `quickNode`) out of the client/asset/eth package it seems?  Looks like we can only do what we want with internals from the eth package.

Also, how about client/asset/eth/cmd/get gas?  I was really only thinking that the client/cmd stuff that needs to be consolidated is product/user oriented tools, while this is a dev tool.

buck54321:
> There's no way to move the GetGasEstimates function (and the other non-product code like runSimnetMiner and quickNode) out of the client/asset/eth package it seems?

I didn't see an easy way without sumoning interface hell.

> Also, how about client/asset/eth/cmd/getgas?

I almost put it there. Will do.

client/asset/eth/nodeclient_harness_test.go
```diff
@@ -716,17 +716,17 @@ func prepareTokenClients(t *testing.T) {
 		t.Fatalf("unexpected error while waiting to mine approval block: %v", err)
 	}
 
-	receipt1, err := waitForReceipt(ethClient, tx1)
+	_, err = waitForReceipt(ethClient, tx1)
 	if err != nil {
 		t.Fatal(err)
 	}
-	spew.Dump(receipt1)
+	// spew.Dump(receipt1)
```
buck54321, line 723: `waitForReceipt` also dumped the receipt, so they were dumped twice, and without any context. I'm personally not a fan of using `spew.Dump` the way we do in these tests, but if we do, we need to add some additional messaging to indicate what is being shown.

chappjc:
> I didn't see an easy way without summoning interface hell.

Yah let's avoid that.  No one will try to use the function anyway.

chappjc:
> > Also, how about client/asset/eth/cmd/getgas?
> 
> I almost put it there. Will do.

Thanks.  Also I see you have the binary accidentally checked in.

JoeGruffins (commented):
On testnet I see:
```
$ ./getgas --testnet --n 3
2023-01-19 14:35:09.576 [INF] GG: getting gas estimates for up to 3 swaps of asset eth, contract version 0 on eth
2023-01-19 14:35:11.222 [INF] GG: Initiator address: 0x0482199652949Df7cb4304B95c084286f1dF8E68
2023-01-19 14:35:11.222 [ERR] GG: error getting tip cap suggestion from "goerli.infura.io": client is closed
2023-01-19 14:35:11.222 [INF] GG: 10080006 gwei eth required for fees and swaps
error getting eth balance: client is closed
```

The harness tests are ok with this infura endpoint so I don't think it's that.

Working well on simnet.

client/asset/eth/eth.go
```diff
@@ -3886,15 +3889,266 @@ func checkTxStatus(receipt *types.Receipt, gasLimit uint64) error {
...
+func runSimnetMiner(ctx context.Context, log dex.Logger) {
+	log.Infof("Starting the simnet miner")
+	go func() {
+		tick := time.NewTicker(time.Second * 5)
+		u, _ := user.Current()
+		for {
+			select {
+			case <-tick.C:
+				log.Debugf("Mining a simnet block")
+				mine := exec.CommandContext(ctx, "./mine-alpha", "1")
```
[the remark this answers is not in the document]
buck54321: Oh really? I thought that was just kinda random because you can only specify to start/stop the miner with eth. But it is (almost?) always 2 blocks. Is that a geth thing?

client/asset/eth/eth.go
```diff
@@ -3950,28 +4204,58 @@ func GetGasEstimates(ctx context.Context, cl ethFetcher, c contractor, maxSwaps
...
-			if err != nil {
-				return fmt.Errorf("error estimating transfer gas: %v", err)
-			}
+	// Estimate approve for tokens.
+	if isToken {
+		txOpts, err := cl.txOpts(ctx, 0, g.Approve*2, nil, nil)
+		if err != nil {
+			return fmt.Errorf("error constructing signed tx opts for approve: %w", err)
+		}
+		log.Debugf("Sending approval transaction")
```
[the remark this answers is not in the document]
buck54321: I guess we'd need to create a random account and send it some funds for gas. Oof.

client/asset/eth/eth.go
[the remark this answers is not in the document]
JoeGruffins: iirc we set the block time to one second.

Geth has no way to say "mine one block."

So we mining.Start, then mining.Stop later. If we don't wait one second before we stop, it might not mine one at all. If we wait a second before calling mining.Stop, it might mine another or it might not. So, stopping after it at least mined one was the safest bet, and not shooting for one.

Sorry, not really a precision of two, just the first block can't guarantee one, and the rest are also sort of one off at times.

JoeGruffins (approved):
Working well. Testnet usdc:
```
  First swap used 170853 gas 
    4 additional swaps averaged 112583 gas each
    [170853 283439 396025 508611 621185] 
  First redeem used 83862 gas 
    4 additional redeems averaged 31641 gas each
    [83862 115518 147162 178783 210427] 
  Average of 5 refunds: 64337 
    [64337 64337 64337 64337 64337] 
  Average of 2 approvals: 50240 
    [60190 40290] 
  Average of 1 transfers: 65500 
    [65500] 
```

client/asset/eth/cmd/getgas/README.md
```diff
@@ -0,0 +1,39 @@
...
+        "simnet": "http://localhost:38556",
+        "testnet": "https://goerli.infura.io/v3/<API key>",
+        "mainnet": "https://mainnet.infura.io/v3/<API key>"
+    }
+}
+```
+
+- Decide the maximum number of swaps you want in the largest initiation transaction, `--n`. Minimum is 2. `getgas` will check initiations with from 1 up to `n` swaps. There is a balance between cost and precision. Using more than 2 generates an average over `n - 1` intiations to calculate the cost of additional swaps (`Gases.SwapAdd`) and redeems (`Gases.RedeemAdd`).
+
+- If you run `getgas` with insufficient or zero ETH and/or token balance on the seed, no transactions will be sent and you'll get a message indicating the amount of funding needed to run.
```
chappjc, line 27: How do you the leftover funding out when done with the tool?

buck54321: I didn't think we would return funds. For ETH, I would just leave any remainder in the wallet for future estimation. For tokens, we're only funding a few atoms, almost definitely not worth the tx fees to send back.

chappjc:
> I did not notice before, but this does not actually refund, just estimates it. It seems like the refund estimate is a smidge off because of this [#2013 (comment)](https://github.com/decred/dcrdex/pull/2013#discussion_r1101519013)

I guess we'll address that in a subsequent issue.  This explains the confusion in that PR about the refund gas discrepancy.

client/asset/eth/cmd/getgas/README.md
chappjc: How do you deposit funds for a given seed?  You run it once and it fails with this zero balance error but it gives you an address?

client/asset/eth/cmd/getgas/README.md
```diff
@@ -0,0 +1,39 @@
+## Gas Estimator Utility
+
+`getgas` is a program for getting gas estimates for DEX-critical operations
+on Ethereum. Use `getgas` to generate gas tables (`dexeth.Gases`) when adding a
+new token or a new contract version.
```
chappjc, line 5: This suggests the gases from this tool can be inserted directly with no additional buffering or even rounding up.  Maybe this should say "Use `getgas` results as a basis for the gas tables when..."

We've targeted an actual observed gas use of ~75% of the gas cap.  For example, to get a `v0Gases` for a new L1 or the `Gases` in the `Tokens` map for a token, you'd take the gasses from this tool and multiply by something like 1.3 and use those values.

client/asset/eth/eth.go
```diff
@@ -4383,19 +4385,107 @@ func runSimnetMiner(ctx context.Context, log dex.Logger) {
...
+		return "", "", err
+	}
+	defer zero()
+	privateKey, err := crypto.ToECDSA(privB)
+	if err != nil {
+		return "", "", err
+	}
+
+	publicKey := privateKey.Public()
+	publicKeyECDSA, ok := publicKey.(*goecdsa.PublicKey)
```
chappjc, line 4412: The public key is an exported field, so we can just grab it:

```go
	privateKey, err := crypto.ToECDSA(privB)
	if err != nil {
		return "", "", err
	}
	addr = crypto.PubkeyToAddress(privateKey.PublicKey).String()
```

client/asset/eth/eth.go
```diff
@@ -579,6 +580,29 @@ func genWalletSeed(entropy []byte) ([]byte, error) {
...
+	}
+	// defer extKey.Zero()
+	pk, err = extKey.SerializedPrivKey()
+	if err != nil {
+		extKey.Zero()
+		return nil, nil, err
+	}
+	return pk, func() {
+		extKey.Zero()
+		defer encode.ClearBytes(pk)
```
chappjc, line 602: Not new in this commit, but turns out we shouldn't be doing `encode.ClearBytes(pk)`.  The docs for `SerializedPrivKey` say `The returned bytes must not be modified.` and it is because it is a shallow copy of the byte slice (`extKey.Zero()` handles it).

client/asset/eth/eth.go
```diff
@@ -4383,19 +4385,107 @@ func runSimnetMiner(ctx context.Context, log dex.Logger) {
...
-// funding needed to run.
-func GetGasEstimates(ctx context.Context, net dex.Network, assetID, contractVer uint32, maxSwaps int, credentialsPath string, log dex.Logger) error {
+type getGas byte
+
+// GetGas provides access to the gas estimation utilities.
+var GetGas getGas
+
+// ReadCredentials reads the credentials for the network from the credentials
+// file.
+func (getGas) ReadCredentials(credentialsPath string, net dex.Network) (addr, provider string, err error) {
```
chappjc, line 4395: Interesting.  Using methods with no receiver variables to sorta make a namespace for these functions rather than prefix `GetGas*` to all the vanilla functions?

chappjc (commented):
Nice added functionality.

client/asset/eth/eth.go
buck54321: Exactly.

client/asset/eth/eth.go
buck54321: I noticed it was a shallow copy, but I didn't want to assume anything. Will just use `Zero`.

chappjc: Understandable, but since `SerializedPrivKey` docs are explicit about it, I think we should not zero the slice.

chappjc (approved):

PR #2051, opened 2023-01-14
eth/multirpc: add timeouts to requests

buck54321:
Re #2045.

### DIFF
--- a/client/asset/eth/multirpc.go
+++ b/client/asset/eth/multirpc.go
@@ -50,6 +50,7 @@ const (
 	tipCapSuggestionExpiration   = time.Hour
 	brickedFailCount             = 100
 	providerDelimiter            = " "
+	defaultRequestTimeout        = time.Second * 10
 )
 
 // nonceProviderStickiness is the minimum amount of time that must pass between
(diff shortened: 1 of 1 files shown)


chappjc:
Nice.  Do you happen to know what default or built-in rpcclient timeout I was hitting in https://github.com/decred/dcrdex/issues/2045?  I can't find any timeout in geth's `httpConn` that it would have been using for the requests.
Is that the issue you meant to reference?  I don't see the link with PR 2025.

buck54321:
Yes. Typo. Should have been #2045. Edited.

> Do you happen to know what default or built-in rpcclient timeout I was hitting in https://github.com/decred/dcrdex/issues/2045?

Hmm. I was really trying to address the hang at `ChainID` and carrying that through, but I similarly could find no timeouts in geth for http requests, so I'm not sure where the context deadline errors are coming from. I'll take another look.

JoeGruffins:
> Nice. Do you happen to know what default or built-in rpcclient timeout I was hitting in https://github.com/decred/dcrdex/issues/2045? I can't find any timeout in geth's httpConn that it would have been using for the requests.

Could it be these default timeouts? https://github.com/ethereum/go-ethereum/blob/a35b654f25fdac518d5775b55aa064025040b01a/rpc/http.go#L114-L121

JoeGruffins (approved):
Looks right and working well.

chappjc:
> > Nice. Do you happen to know what default or built-in rpcclient timeout I was hitting in #2045? I can't find any timeout in geth's httpConn that it would have been using for the requests.
> 
> Could it be these default timeouts? https://github.com/ethereum/go-ethereum/blob/a35b654f25fdac518d5775b55aa064025040b01a/rpc/http.go#L114-L121

I had followed those, and they seem only to be used when geth makes it's RPC *server*.  No matter.

chappjc (approved):

PR #2053, opened 2023-01-17
client: fix missing settings button for erc20 tokens

### LINKED ISSUE #2040: ui: wallet settings button (and disable button in that view) missing for ERC20 tokens

vctt94:
closes #2040.

I am always showing the settings button for tokens now, even though they don't have configopts defined. This way it is possible to enable or disable and export them.

### DIFF
--- a/client/webserver/site/src/js/wallets.ts
+++ b/client/webserver/site/src/js/wallets.ts
@@ -629,18 +629,15 @@ export default class WalletsPage extends BasePage {
       page.balanceBox, page.fiatBalanceBox, page.createWalletBox, page.walletDetails,
       page.sendReceive, page.connectBttnBox, page.statusLocked, page.statusReady,
       page.statusOff, page.unlockBttnBox, page.lockBttnBox, page.connectBttnBox,
-      page.reconfigureBox, page.peerCountBox, page.syncProgressBox, page.statusDisabled
+      page.peerCountBox, page.syncProgressBox, page.statusDisabled
     )
     if (wallet) {
       this.updateDisplayedAssetBalance()
 
       const walletDef = app().walletDefinition(assetID, wallet.type)
       page.walletType.textContent = walletDef.tab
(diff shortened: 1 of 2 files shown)


vctt94:
Thanks for the review Marton. Changed for your proposed solution and I believe it makes more sense.

JoeGruffins (approved):

client/webserver/site/src/js/wallets.ts
```diff
@@ -636,8 +636,11 @@ export default class WalletsPage extends BasePage {
 
       const walletDef = app().walletDefinition(assetID, wallet.type)
       page.walletType.textContent = walletDef.tab
+      Doc.show(page.reconfigureBox)
       const configurable = assetIsConfigurable(assetID)
-      if (configurable) Doc.show(page.reconfigureBox)
+      // if it is not configurable, we do not show the password form.
+      if (configurable) Doc.show(page.passwordWrapper)
```
martonp, line 642:
``` javascript
Doc.setVis(configurable, page.passwordWrapper)
```

client/webserver/site/src/js/wallets.ts
```diff
@@ -636,8 +636,11 @@ export default class WalletsPage extends BasePage {
 
       const walletDef = app().walletDefinition(assetID, wallet.type)
       page.walletType.textContent = walletDef.tab
+      Doc.show(page.reconfigureBox)
```
martonp, line 639: I don't think you need to explicitly show this. Just remove `page.reconfigureBox` from the `Doc.hide` above.

martonp (approved):

chappjc (approved):
utACK.  In the absence of a token wallet definition with any configopts (which is the case presently), hiding the Submit button is the right move so that `showReconfig` is accessible and shows the disable button among other things.

buck54321 (approved):

PR #2054, opened 2023-01-17
client: hide order form if asset version is not supported

### LINKED ISSUE #2039: ui: hide order form for incompatible asset versions
Need to check `asset.WalletInfo.SupportedVersions` against `Core.Exchange.Assets[assetID].Version` for both assets and hide the order form if we can't trade.

vctt94:
closes #2039

### DIFF
--- a/client/webserver/site/src/js/markets.ts
+++ b/client/webserver/site/src/js/markets.ts
@@ -605,8 +605,11 @@ export default class MarketsPage extends BasePage {
    * supported
    */
   assetsAreSupported () {
-    const [b, q] = [this.market.base, this.market.quote]
-    return b && q
+    const [base, quote] = [this.market.base, this.market.quote]
+    if (!base || !quote) return false
+    const { baseCfg: b, quoteCfg: q } = this.market
+    // check if versions are supported
+    return base.info?.versions.includes(b.version) && quote.info?.versions.includes(q.version)
   }
 
   /*
--- a/client/webserver/site/src/js/registry.ts
+++ b/client/webserver/site/src/js/registry.ts
@@ -191,6 +191,7 @@ export interface WalletInfo {
   name: string
   version: number
   availablewallets: WalletDefinition[]
+  versions: number[]
   emptyidx: number
   unitinfo: UnitInfo
 }


client/webserver/site/src/js/markets.ts
```diff
@@ -605,8 +605,11 @@ export default class MarketsPage extends BasePage {
    * supported
    */
   assetsAreSupported () {
-    const [b, q] = [this.market.base, this.market.quote]
-    return b && q
+    const [base, quote] = [this.market.base, this.market.quote]
+    if (!base || !quote) return false
+    const { baseCfg: b, quoteCfg: q } = this.market
+    // check if versions are supported
+    return base.info?.versions.includes(b.version) && quote.info?.versions.includes(q.version)
```
martonp, line 612: Could you test this with tokens? WalletInfo will not be defined for tokens, so I think this will throw an error.

client/webserver/site/src/js/markets.ts
```diff
@@ -605,8 +605,11 @@ export default class MarketsPage extends BasePage {
    * supported
    */
   assetsAreSupported () {
-    const [b, q] = [this.market.base, this.market.quote]
-    return b && q
+    const [base, quote] = [this.market.base, this.market.quote]
+    if (!base || !quote) return false
+    const { baseCfg: b, quoteCfg: q } = this.market
```
martonp, line 610: I would just leave this as `baseCfg` and `quoteCfg` to be able to more clearly differentiate in the check below.

client/webserver/site/src/js/markets.ts
vctt94: It will not throw an error because of the optional chaining operator, but it will also not work. Fixed in the next commit, thanks for spotting it.

vctt94:
Thanks for the review marton, addressed your comments.

ukane-philemon:
Thanks @vctt94 for picking this up. 

Don't we want display a message to user if we eventually hide the order form due to this new check? I think,  `setLoaderMsgVisibility()` displays a message for unsupported assets but does not consider assets with unsupported version.

vctt94:
Hey Ukane, my pleasure. 

I am not sure I understand your point, right now it does show a message if assets aren't supported, which is: `{{ asset }} is not supported`.

Do you mean it should be a more descreptive message?

ukane-philemon:
> Do you mean it should be a more descreptive message?

Yeah.

vctt94:
Thanks for the suggestion, but I am not sure I agree with you.

Right now the `assetsAreSupported()` method returns a boolean, which indicates if assets are supported or not. So to be more descriptive about the message we would need to change this method in order to know what is the reason of this asset not being supported so we can show the right message at `setLoaderMsgVisibility()`.

Which I believe `{{ asset }} is not supported` already covers the point.

chappjc:
Can you share a screenshot of current view when not supported?

vctt94:
This is how it currently looks:

image

chappjc:
So is what's missing the unsupported *version*?

vctt94:
Yeah I think so. If everyone agrees I can change `assetsAreSupported`, to show the unsupported version as well.

vctt94:
Addressed @ukane-philemon comment on my last commit.

This is how it looks right now:

image

client/webserver/site/src/js/locales.ts
```diff
@@ -111,6 +112,7 @@ export const enUS: Locale = {
   [ID_BUY]: 'Buy',
   [ID_SELL]: 'Sell',
   [ID_NOT_SUPPORTED]: '{{ asset }} is not supported',
+  [ID_VERSION_NOT_SUPPORTED]: 'version: {{ version }} of {{ asset }} is not supported',
```
chappjc, line 115: Let's rephrase this.  The "version: " prefix there is a bit odd.  How about:

> {{ asset }} version {{ version }} is not supported

e.g. "USDC.ETH version 0 is not supported"

or

> {{ asset }} (v{{version}}) is not supported

e.g. "USDC.ETH (v0) is not supported"

Any preferences?

I think the point of the issue this PR is aiming to resolve is that without naming the version it might be confusing if the user can see that the server does indeed have a market for this asset, so we have to state the version so there's some kind of explanation for why it's not supported.

client/webserver/site/src/js/markets.ts
```diff
@@ -604,9 +604,74 @@ export default class MarketsPage extends BasePage {
...
+      }
+    }
+    const { baseCfg, quoteCfg } = this.market
+    // check if versions are supported. If asset is a token, we check if its
+    // parent supports the version.
+    let text = ''
+    if (base.token && quote.token) {
+      const bParent = app().assets[base.token.parentID]
+      const qParent = app().assets[quote.token.parentID]
+      if (!bParent.info?.versions.includes(baseCfg.version)) {
```
martonp, line 627: I don't think this is correct, you're checking the parent's version against the token version. The token also has its own version independent of the parent asset, but it doesn't seem to be exposed through the `SupportedAsset` type. I think the token wallet's version also needs to be exposed through the `SupportedAsset` type. @buck54321  Am I correct about this?

buck54321: Sorry for the late response. Yes. We are not properly exposing supported versions for tokens. @vctt94, check out https://github.com/buck54321/dcrdex/compare/97a8cb2593588862c0ac7774f83fbe1e84dae952...buck54321:dcrdex:supported-versions-addon

buck54321: Just opened #2094 too, so if that goes in first, you can just rebase to get the client/asset parts.

buck54321: I've changed tacts on this. See the updated #2094, but I think we'll have the server maintain a single protocol version and token assets will share both a version and supported versions with the parent asset. The code here is roughly correct, though I'd recommend the following refactor https://github.com/buck54321/dcrdex/compare/97a8cb2593588862c0ac7774f83fbe1e84dae952...buck54321:dcrdex:parent-vers-pattern

The only difference is that it will show the token's symbol instead of the parent asset, but I think that makes sense anyway.

chappjc:
https://github.com/decred/dcrdex/pull/2094 is in, can rebase that commit out of your branch

vctt94:
rebased

buck54321 (approved):

PR #2057, opened 2023-01-20
app: Fix swap fee summary

### LINKED ISSUE #2041: erc20: token fee estimates on ui are wild
image

Noted in https://github.com/decred/dcrdex/pull/1733#pullrequestreview-1218454607 when buying DCR with USDC (testnet)

martonp:
The fee estimates for tokens on the UI was not accounting for the existance of tokens. When trading tokens, the fee is paid in the parent asset, so it doesn't make sense to display the fee as a percentage of the order amount. The fee summary pane now displays the actual fee estimate ranges instead of the percentage, and the fee details pane only shows percentages for non-token assets.

Screen Shot 2023-01-20 at 2 21 14 PM
Screen Shot 2023-01-20 at 2 21 05 PM

Closes #2041


JoeGruffins (commented):
On the form, I think there could be an explanation for what the (%) is displaying for non-tokens. I don't think it is very obvious that it is the percent of total trade.

On simnet, eth will say "Order estimates and options unavailable" because "gas required exceeds allowance (big number)". Why was this?

JoeGruffins:
Trying this on testnets and seeing a little weirdness:
build ID is correct
image

No fees selling eth on eth-dcr market here:
image

And the fees have double parenthesis:
image

martonp:
> Trying this on testnets and seeing a little weirdness:

It seems like somehow the updates to the `markets.tmpl` did not make it into your build.

JoeGruffins:
Let me try building again?

client/webserver/locales/zh-cn.go
```diff
@@ -195,7 +195,7 @@ var ZhCN = map[string]string{
 	"details":                     "细节",
 	"to":                          "到",
 	"Options":                     "选项",
-	"fee_projection_tooltip":      "如果在您的订单匹配之前网络条件没有改变，则总费用（作为交易的百分比）应该在这个范围内。",
+	"fee_projection_tooltip":      "如果在您的订单匹配之前网络条件没有改变，则总费用应该在这个范围内。",
```
buck54321, line 198: No space?

client/webserver/site/src/html/markets.tmpl
```diff
@@ -563,12 +563,19 @@
...
-            <span id="vFeeSummaryLow"></span>
-            <span class="fs15">%</span>
-            <span class="fs15 px-2">[[[to]]]</span>
-            <span id="vFeeSummaryHigh"></span>
-            <span class="fs15">%</span>
+          <div class="py-1 flex-column align-items-center justify-content-center fs18" id="vFeeSummary">
+            <div class="d-flex flex-row align-items-center justify-content-center">
+              <img class="micro-icon mx-1" data-icon="from">
+              <span id="summarySwapFeesLow"></span>
+              -
```
buck54321, line 570: How about `<span class="px-1">&ndash;</span>`

client/webserver/locales/zh-cn.go
martonp: Haha no spaces in Chinese.

client/webserver/site/src/html/markets.tmpl
```diff
@@ -646,8 +653,8 @@
                   <td><img class="micro-icon" data-icon="to"> [[[Redeem]]]</td>
                 </tr>
                 <tr>
-                  <td><span id="vSwapFeesLow"></span> (<span id="vSwapFeesLowPct"></span>)</td>
-                  <td><span  id="vRedeemFeesLow"></span> (<span  id="vRedeemFeesLowPct"></span>)</td>
+                  <td><span id="vSwapFeesLow"></span> <span id="vSwapFeesLowPct"></span></td>
+                  <td><span  id="vRedeemFeesLow"></span> <span  id="vRedeemFeesLowPct"></span></td>
```
chappjc, line 657: If you are doing another commit, can you remove the weird extra space after `span` in this element tag?

martonp:
> On the form, I think there could be an explanation for what the (%) is displaying for non-tokens. I don't think it is very obvious that it is the percent of total trade.

I don't have time to fix this right now. I think this PR can go in without it though, because that issue was already existing.

> On simnet, eth will say "Order estimates and options unavailable" because "gas required exceeds allowance (big number)". Why was this?

This is happening because when doing `MaxOrder`, we are attempting to estimate gas for a huge amount of lots, and the transaction will surpass the block gas limit.

chappjc:
> I don't have time to fix this right now. I think this PR can go in without it though, because that issue was already existing.

Yup, all good here.

client/webserver/site/src/js/markets.ts
```diff
@@ -1529,17 +1529,36 @@ export default class MarketsPage extends BasePage {
...
-    for (const icon of Doc.applySelector(page.vDetailPane, '[data-icon]')) {
+    const setIcon = (icon: PageElement) => {
       switch (icon.dataset.icon) {
         case 'from':
-          icon.src = Doc.logoPath(fromAsset.symbol)
+          if (fromAsset.token) {
+            const parentAsset = app().assets[fromAsset.token.parentID]
+            icon.src = Doc.logoPath(parentAsset.symbol)
+          } else {
+            icon.src = Doc.logoPath(fromAsset.symbol)
```
chappjc, line 1539: This new `setIcon` assumes the icon will only be used to represent fees.  If the order verification form wants to put the icons for the traded assets, this helper won't be of help.  Seems like we need `[data-icon]` for the traded asset and `[data-fee-icon]` for the fee asset icon.  Fine in this PR though.

chappjc (approved):
LGTM.  @buck54321 pls merge when you're good with it, or we can push a fix to it if needed.

buck54321 (approved):

PR #2059, opened 2023-01-21
server/eth: Do not use txpool when calculating balance if not supported

martonp:
The server's rpcClient will only take into account pending transactions if it can use the rpc_getmethods method to check the available methods, and if the txpool namespace is available.


JoeGruffins (approved):
I like the names!

Was able to trade on testnet using infura with this.

chappjc (approved):
Thank you.

chappjc:
Server can now use regular RPC providers.

PSA: try to use your own full node in production, or at least a provider with an API key assigned to you, preferably with a service agreement.

PR #2060, opened 2023-01-21
client/assets: add feeRate sanity check for swaps and minor refactoring

### LINKED ISSUE #1585: client/core: don't send zero-fee swaps
While testing doge in a primitive stage, @chappjc managed to get a redeem wedged since it was sent with a fee rate of zero. The bug there is fixed, but maybe we should have caught that in a sanity check somewhere. It's also related to #1573, because the redemption was assumed sent when in reality it never propagated.

ukane-philemon:
Resolves #1585.

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -2881,7 +2881,7 @@ func accelerateOrder(btc *baseWallet, swapCoins, accelerationCoins []dex.Bytes, 
 	if err != nil {
 		return nil, "", err
 	}
-	err = btc.broadcastTx(signedTx)
+	_, err = btc.broadcastTx(signedTx)
 	if err != nil {
 		return nil, "", err
 	}
@@ -3303,7 +3303,7 @@ func (btc *baseWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin, ui
 	}
 
 	// Refund txs prepared and signed. Can now broadcast the swap(s).
-	err = btc.broadcastTx(msgTx)
+	_, err = btc.broadcastTx(msgTx)
 	if err != nil {
 		return nil, nil, 0, err
 	}
@@ -3468,14 +3468,9 @@ func (btc *baseWallet) Redeem(form *asset.RedeemForm) ([]dex.Bytes, asset.Coin, 
 	}
 
 	// Send the transaction.
-	checkHash := btc.hashTx(msgTx)
-	txHash, err := btc.node.sendRawTransaction(msgTx)
+	txHash, err := btc.broadcastTx(msgTx)
 	if err != nil {
 		return nil, nil, 0, err
-	}
-	if *txHash != *checkHash {
-		return nil, nil, 0, fmt.Errorf("redemption sent, but received unexpected transaction ID back from RPC server. "+
-			"expected %s, got %s", *txHash, checkHash)
 	}
 	// Log the change output.
 	coinIDs := make([]dex.Bytes, 0, len(form.Redemptions))
@@ -3951,14 +3946,9 @@ func (btc *baseWallet) Refund(coinID, contract dex.Bytes, feeRate uint64) (dex.B
 		return nil, fmt.Errorf("error creating refund tx: %w", err)
 	}
 
-	checkHash := btc.hashTx(msgTx)
-	refundHash, err := btc.node.sendRawTransaction(msgTx)
+	refundHash, err := btc.broadcastTx(msgTx)
 	if err != nil {
-		return nil, fmt.Errorf("sendRawTransaction: %w", err)
-	}
-	if *refundHash != *checkHash {
-		return nil, fmt.Errorf("refund sent, but received unexpected transaction ID back from RPC server. "+
-			"expected %s, got %s", *refundHash, checkHash)
+		return nil, fmt.Errorf("broadcastTx: %w", err)
 	}
 	return toCoinID(refundHash, 0), nil
 }
@@ -4519,7 +4509,7 @@ func (btc *baseWallet) sendWithReturn(baseTx *wire.MsgTx, addr btcutil.Address,
 		return nil, err
 	}
 
-	err = btc.broadcastTx(signedTx)
+	_, err = btc.broadcastTx(signedTx)
 	return signedTx, err
 }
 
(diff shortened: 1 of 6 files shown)


client/asset/btc/btc.go
```diff
@@ -3299,7 +3303,7 @@ func (btc *baseWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin, ui
 	}
 
 	// Refund txs prepared and signed. Can now broadcast the swap(s).
-	err = btc.broadcastTx(msgTx)
+	_, err = btc.broadcastTx(msgTx)
```
chappjc, line 3306: I'm totally fine with the `btc.broadcastTx` refactoring (I've thought about doing just this before), but this is the kind of thing I'd like to see in a separate commit in the future.  There are three lines in btc.go that pertain to what this PR aims to resolve.

chappjc: Workflow suggestion: experiment with `git add -p`, or similar "State Selected Ranges" functionality of some IDEs.  It helps add/stage only certain changes in a file.

ukane-philemon: Yeah, totally agree. This one slipped through.

client/asset/btc/btc.go
```diff
@@ -3190,6 +3190,10 @@ func preAccelerate(btc *baseWallet, swapCoins, accelerationCoins []dex.Bytes, ch
 // are NOT manually unlocked because they're auto-unlocked when the transaction
 // is broadcasted.
 func (btc *baseWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin, uint64, error) {
+	if swaps.FeeRate == 0 {
+		return nil, nil, 0, fmt.Errorf("cannot send swap with with zero fee rate")
+	}
```
buck54321, line 3195: I wish I remembered better the details of #1587, but I think it was an issue on the server. In that case, it would be fine to check just once at https://github.com/decred/dcrdex/blob/69ace7bdafad0a8b70714be4e5b7153f4c83efed/client/core/trade.go#L2329-L2330

```Go
if highestFeeRate == 0 {
    return fmt.Errorf(...)
}
```

ukane-philemon: Yes, we can do that and save a few LOC. I initially assumed the check would be better at the asset level.

chappjc: I don't have a problem with it in both places TBH, but core also shouldn't request impossible swaps.

ukane-philemon: > I don't have a problem with it in both places TBH

I left the check at the asset level as an extra precaution, since it is very crucial to the swap process. Or do we really want to remove the check from the asset level? @chappjc

ukane-philemon:
It seems the new check is effective. https://github.com/decred/dcrdex/actions/runs/3988636125/jobs/6840053009

<details>
2023-01-23 16:54:39.188 [ERR] TCORE: somedex.tld:7232 tick error: somedex.tld:7232 tick: {swapMatches order b2c0e3807fc2b2819a125b672c7bea91c991b2f8d5cf74e338332fd72f6415d9 - {swap cannot proceed with a zero fee rate}}
[1547](https://github.com/decred/dcrdex/actions/runs/3988636125/jobs/6840053009#step:5:1548)
2023-01-23 16:54:39.188 [ERR] TCORE: Error fetching fee rate for btc: no handler for route "fee_rate"
[1548](https://github.com/decred/dcrdex/actions/runs/3988636125/jobs/6840053009#step:5:1549)
panic: test timed out after 10m0s
[1549](https://github.com/decred/dcrdex/actions/runs/3988636125/jobs/6840053009#step:5:1550)

[1550](https://github.com/decred/dcrdex/actions/runs/3988636125/jobs/6840053009#step:5:1551)
goroutine 426 [running]:
[1551](https://github.com/decred/dcrdex/actions/runs/3988636125/jobs/6840053009#step:5:1552)
testing.(*M).startAlarm.func1()
[1552](https://github.com/decred/dcrdex/actions/runs/3988636125/jobs/6840053009#step:5:1553)
	/opt/hostedtoolcache/go/1.19.5/x64/src/testing/testing.go:2036 +0xbb
[1553](https://github.com/decred/dcrdex/actions/runs/3988636125/jobs/6840053009#step:5:1554)
created by time.goFunc
[1554](https://github.com/decred/dcrdex/actions/runs/3988636125/jobs/6840053009#step:5:1555)
	/opt/hostedtoolcache/go/1.19.5/x64/src/time/sleep.go:176 +0x48
[1555](https://github.com/decred/dcrdex/actions/runs/3988636125/jobs/6840053009#step:5:1556)
</details>

chappjc:
> It seems the new check is effective

Haha, yes it seems.  But let's fix _the tests_ that neglect to set a fee rate to use for these fake swaps.  Changing the product code to use MaxFeeRate if the server prescribes zero and the wallet gives zero doesn't feel right.

ukane-philemon:
> Changing the product code to use MaxFeeRate if the server prescribes zero and the wallet gives zero doesn't feel right.

But does it mean we don't need to set the fee rate to MaxFeeRate if the server and wallet returned zero?

chappjc:
> > Changing the product code to use MaxFeeRate if the server prescribes zero and the wallet gives zero doesn't feel right.
> 
> But does it mean we don't need to set the fee rate to MaxFeeRate if the server and wallet returned zero?

Not sure, but it's rather questionable to fall back to MaxFeeRate.  If we're getting no informed fee rate from any source, what does that mean?  Is the server working?  Is the swap even going to negotiate?  Will the BTC swap cost a fortune?  Will the problem clear up after retrying after some time?

Backing up further, when we get the `msgjson.Match` we could see that the `swapRate` is zero at that point (in `(*dexConnection) parseMatches`).  Why should we even register a match with invalid data in the first place?  We reject a match with a rate above `MaxFeeRate`, so why not reject one with a zero rate?  Are there any conceivable use cases for a zero fee rate being valid?

I'm not positive what's best right now, but the red flag can be seen much earlier.

ukane-philemon:
> Are there any conceivable use cases for a zero fee rate being valid?

Right, this is one thing am not so sure about but It makes sense to reject swaps with a zero fee rate atm if defaulting to MaxFeeRate in this scenario is questionable. Input from others would be valuable @buck54321, @martonp.

JoeGruffins:
This works well. The client will continue to try even if max fee is zero in the servers config file. But this is #1932 and #1900

If the server sets max fee to zero with this pr:
image

JoeGruffins (approved):

chappjc (approved):
Ok, but I think in the future we'll just wanna block acceptance of such matches early on in `(*dexConnection).parseMatches`.  With this PR we won't make an impossible swap txn, but we may retry until revoke as @JoeGruffins pointed out, or we'll use our own rates in a situation where we know the server is misconfigured / match data is invalid.

chappjc:
@ukane-philemon To merge, will you prep this branch by squashing the review fixes into the "sanity check" commit, keeping the `broadcastTx` refactoring commit separate.  We'll rebase merge this with 2 commits.

PR #2062, opened 2023-01-21
client/core: Only check for fees when they are definitely available

martonp:
The fee checks for dynamic fee assets were happening many times before the transaction was confirmed, unnecessarily slowing the ticks down.

### DIFF
--- a/client/core/trade.go
+++ b/client/core/trade.go
@@ -1386,10 +1386,12 @@ func (t *trackedTrade) checkSwapFeeConfirms(match *matchTracker) bool {
 		// Confirmed will be set in the db.
 		return true
 	}
+	// Waiting until the swap is definitely confirmed in order to not
+	// keep calling the fee checker before the swap is confirmed.
 	if match.Side == order.Maker {
-		return match.Status >= order.MakerSwapCast
+		return match.Status >= order.TakerSwapCast
 	}
-	return match.Status >= order.TakerSwapCast
+	return match.Status >= order.MakerRedeemed
 }
 
 // checkRedemptionFeeConfirms returns whether the swap fee confirmations should
@@ -1406,9 +1408,8 @@ func (t *trackedTrade) checkRedemptionFeeConfirms(match *matchTracker) bool {
 		// Confirmed will be set in the db.
 		return true
 	}
-	if match.Side == order.Maker {
-		return match.Status >= order.MakerRedeemed
-	}
+	// Checking taker redemption fees might be a bit early, but there will be
+	// no more ticks after the status is set to MatchConfirmed.
 	return match.Status >= order.MatchComplete
 }
 


chappjc (approved):

client/core/trade.go
```diff
@@ -1386,10 +1386,12 @@ func (t *trackedTrade) checkSwapFeeConfirms(match *matchTracker) bool {
 		// Confirmed will be set in the db.
 		return true
 	}
+	// Waiting until the swap is definitely confirmed in order to not
+	// keep calling the fee checker before the swap is confirmed.
 	if match.Side == order.Maker {
-		return match.Status >= order.MakerSwapCast
+		return match.Status >= order.TakerSwapCast
```
chappjc, line 1392: we can always check `mine, _ := match.confirms()` rather than waiting for potentially many confs (the next action happening)

client/core/trade.go
```diff
@@ -1386,10 +1386,13 @@ func (t *trackedTrade) checkSwapFeeConfirms(match *matchTracker) bool {
 		// Confirmed will be set in the db.
 		return true
 	}
+	// Waiting until the swap is definitely confirmed in order to not
+	// keep calling the fee checker before the swap is confirmed.
+	mine, _ := match.confirms()
 	if match.Side == order.Maker {
-		return match.Status >= order.MakerSwapCast
+		return match.Status > order.MakerSwapCast || (match.Status == order.MakerSwapCast && mine > 1)
```
buck54321, line 1393: Why not...

```Go
return match.Status >= order.MakerSwapCast && mine > 0
```

client/core/trade.go
```diff
@@ -1386,10 +1386,13 @@ func (t *trackedTrade) checkSwapFeeConfirms(match *matchTracker) bool {
 		// Confirmed will be set in the db.
 		return true
 	}
+	// Waiting until the swap is definitely confirmed in order to not
+	// keep calling the fee checker before the swap is confirmed.
+	mine, _ := match.confirms()
```
buck54321, line 1391: A more descriptive variable would be nice. `mySwapConfs` or something.

client/core/trade.go
```diff
@@ -1406,9 +1409,8 @@ func (t *trackedTrade) checkRedemptionFeeConfirms(match *matchTracker) bool {
 		// Confirmed will be set in the db.
 		return true
 	}
-	if match.Side == order.Maker {
-		return match.Status >= order.MakerRedeemed
-	}
+	// Checking taker redemption fees might be a bit early, but there will be
+	// no more ticks after the status is set to MatchConfirmed.
 	return match.Status >= order.MatchComplete
```
buck54321, line 1414: It's not clear what the intention is here. For maker, the difference between `MakerRedeemed` and `MatchComplete` is just a successful `sendRedeemAsync`, which should come immediately after the status is set to `MakerRedeemed` anyway, and seems unrelated to whether or not we want to check confirmations.

client/core/trade.go
chappjc: Unless I'm mistaken the `confirms()` value is all that matters now.  If it's confirmed, does it matter what the match status is?

martonp: The reason I did it like this is the confirmation updates seemed very irregular using the MultiRPC. On Etherscan I would see many confirmations and it would still show 0. I was worried that maybe in some case `setSwapConfirms` would not be called.

How about 

``` golang
    return match.Status > order.MakerSwapCast || mine > 0
```

client/core/trade.go
martonp: Yeah you're right, this change is unnecessary. For the redemption I don't think there's anything we can do due to the irregular confirmation updates I mentioned https://github.com/decred/dcrdex/pull/2062#discussion_r1083591489 . If one updates takes us from 0 to 10, and the status is set to `MatchConfirmed`, `tick` will no longer run. Unless we update `matchIsActive` to take into account these fees.

JoeGruffins (approved):

client/core/trade.go
```diff
@@ -1386,10 +1386,13 @@ func (t *trackedTrade) checkSwapFeeConfirms(match *matchTracker) bool {
 		// Confirmed will be set in the db.
 		return true
 	}
+	// Waiting until the swap is definitely confirmed in order to not
+	// keep calling the fee checker before the swap is confirmed.
+	mySwapConfs, _ := match.confirms()
 	if match.Side == order.Maker {
-		return match.Status >= order.MakerSwapCast
+		return match.Status > order.MakerSwapCast || mySwapConfs > 0
```
buck54321, line 1393: Why `>` instead of `>=`? Why should maker wait to see the taker's swap before checking for fees on its own swap?

client/core/trade.go
chappjc: > On Etherscan I would see many confirmations while still showing 0 in the dex client. I was worried that maybe in some case `setSwapConfirms` would not be called.

In this case, surely the dynamic fee check would also failed to find it on the chain?

client/core/trade.go
chappjc: I believe this is fine since unlike with the swap transactions, there should be no delays after the maker broadcasts their txn.

client/core/trade.go
chappjc: > We are not strictly tracking the confirmations on the swap transaction, and due to the irregular way the confirmations were coming in when testing on testnet, I thought this would be safer.

Hmm, that's an interesting point.  I suppose there can be a case where no `tick` gets your own swap's confirms before moving on to the next status, which could preclude the confs check.

OK, I think this is safest since tick does not make explicit guarantees about this.

Actually, I think the counter party could even make their transaction without waiting for a single confirmation on ours.  Nothing stopping that.

chappjc (approved):

buck54321 (approved):

PR #2065, opened 2023-01-23
client: add static strings to translation

ukane-philemon:
This removes a handful of statics strings and adds them to the `Locale` map.


JoeGruffins (approved):

chappjc (approved):
Looks good, but leaving for other review first.

PR #2069, opened 2023-01-24
 multi: implement asset.Authtenticator for assets that require authentication

ukane-philemon:
This resolves an existing TODO:

>   TODO: Use an asset.Authenticator interface and WalletTraits to do this instead.

- Implements an `asset.Authenticator` for wallets that require auth.
- Added a `btc.ExchangeWalletAuthenticator` the implements the `asset.Authenticator` interface.

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -871,12 +871,18 @@ type intermediaryWallet struct {
 // method to implement asset.Rescanner.
 type ExchangeWalletSPV struct {
 	*intermediaryWallet
+	*authAddOn
 
 	spvNode *spvWallet
 }
 
 // ExchangeWalletFullNode implements Wallet and adds the FeeRate method.
 type ExchangeWalletFullNode struct {
+	*intermediaryWallet
+	*authAddOn
+}
+
+type ExchangeWalletNoAuth struct {
 	*intermediaryWallet
 }
 
@@ -898,6 +904,9 @@ var _ asset.Recoverer = (*ExchangeWalletSPV)(nil)
 var _ asset.PeerManager = (*ExchangeWalletSPV)(nil)
 var _ asset.TxFeeEstimator = (*intermediaryWallet)(nil)
 var _ asset.Bonder = (*baseWallet)(nil)
+var _ asset.Authenticator = (*ExchangeWalletSPV)(nil)
+var _ asset.Authenticator = (*ExchangeWalletFullNode)(nil)
+var _ asset.Authenticator = (*ExchangeWalletAccelerator)(nil)
 
 // RecoveryCfg is the information that is transferred from the old wallet
 // to the new one when the wallet is recovered.
@@ -1077,22 +1086,42 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, net dex.Network) (ass
 // conjunction with ReadCloneParams, to create a ExchangeWallet for other assets
 // with minimal coding.
 func BTCCloneWallet(cfg *BTCCloneCFG) (*ExchangeWalletFullNode, error) {
+	iw, err := btcCloneWallet(cfg)
+	if err != nil {
+		return nil, err
+	}
+	return &ExchangeWalletFullNode{iw, &authAddOn{iw.node}}, nil
+}
+
+// BTCCloneWalletNoAuth is like BTCCloneWallet but the wallet created does not
+// implement asset.Authenticator.
+func BTCCloneWalletNoAuth(cfg *BTCCloneCFG) (*ExchangeWalletNoAuth, error) {
+	iw, err := btcCloneWallet(cfg)
+	if err != nil {
+		return nil, err
+	}
+	return &ExchangeWalletNoAuth{iw}, nil
+}
+
+// btcCloneWallet creates a wallet backend for a set of network parameters and
+// default network ports.
+func btcCloneWallet(cfg *BTCCloneCFG) (*intermediaryWallet, error) {
 	clientCfg, client, err := parseRPCWalletConfig(cfg.WalletCFG.Settings, cfg.Symbol, cfg.Network, cfg.Ports, cfg.SingularWallet)
 	if err != nil {
 		return nil, err
 	}
 
-	btc, err := newRPCWallet(client, cfg, clientCfg)
+	iw, err := newRPCWallet(client, cfg, clientCfg)
 	if err != nil {
-		return nil, fmt.Errorf("error creating %s ExchangeWallet: %v", cfg.Symbol,
+		return nil, fmt.Errorf("error creating %s exchange wallet: %v", cfg.Symbol,
 			err)
 	}
 
-	return btc, nil
+	return iw, nil
 }
 
 // newRPCWallet creates the ExchangeWallet and starts the block monitor.
-func newRPCWallet(requester RawRequester, cfg *BTCCloneCFG, parsedCfg *RPCWalletConfig) (*ExchangeWalletFullNode, error) {
+func newRPCWallet(requester RawRequester, cfg *BTCCloneCFG, parsedCfg *RPCWalletConfig) (*intermediaryWallet, error) {
 	btc, err := newUnconnectedWallet(cfg, &parsedCfg.WalletConfig)
 	if err != nil {
 		return nil, err
@@ -1131,12 +1160,10 @@ func newRPCWallet(requester RawRequester, cfg *BTCCloneCFG, parsedCfg *RPCWallet
 	core.requesterV.Store(requester)
 	node := newRPCClient(core)
 	btc.node = node
-	return &ExchangeWalletFullNode{
-		intermediaryWallet: &intermediaryWallet{
-			baseWallet:     btc,
-			txFeeEstimator: node,
-			tipRedeemer:    node,
-		},
+	return &intermediaryWallet{
+		baseWallet:     btc,
+		txFeeEstimator: node,
+		tipRedeemer:    node,
 	}, nil
 }
 
@@ -1288,7 +1315,8 @@ func OpenSPVWallet(cfg *BTCCloneCFG, walletConstructor BTCWalletConstructor) (*E
 			txFeeEstimator: spvw,
 			tipRedeemer:    spvw,
 		},
-		spvNode: spvw,
+		authAddOn: &authAddOn{spvw},
+		spvNode:   spvw,
 	}, nil
 }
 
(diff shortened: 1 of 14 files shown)


JoeGruffins (approved):
Behavior of everything seems to be the same as before.

client/asset/btc/btc.go
```diff
@@ -140,13 +140,15 @@ var (
 		Description:       "Connect to bitcoind",
 		DefaultConfigPath: dexbtc.SystemConfigPath("bitcoin"),
 		ConfigOpts:        append(RPCConfigOpts("Bitcoin", "8332"), CommonConfigOpts("BTC", true)...),
+		Traits:            asset.DetermineWalletTraits(&ExchangeWalletAccelerator{}),
```
chappjc, line 143: Can't do this.  Only the actual instance can have it's traits identified.

client/asset/btc/btc.go
```diff
@@ -856,10 +859,15 @@ type ExchangeWalletFullNode struct {
 	*intermediaryWallet
 }
 
+// ExchangeWalletAuthenticator implements the asset.Authenticator.
+type ExchangeWalletAuthenticator struct {
+	*ExchangeWalletFullNode
+}
+
```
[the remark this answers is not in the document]
chappjc: No need for a new type

client/asset/btc/btc.go
```diff
@@ -2540,19 +2584,19 @@ func (btc *baseWallet) FundingCoins(ids []dex.Bytes) (asset.Coins, error) {
...
+// lock locks the ExchangeWallet and the underlying bitcoind wallet.
+func (btc *baseWallet) lock() error {
 	return btc.node.walletLock()
 }
 
-// Locked will be true if the wallet is currently locked.
-func (btc *baseWallet) Locked() bool {
+// locked will be true if the wallet is currently locked.
+func (btc *baseWallet) locked() bool {
 	return btc.node.locked()
```
chappjc, line 2600: Pretty sure this was all fine in `baseWallet`.  Comes back to @JoeGruffins' comment

client/webserver/site/src/js/forms.ts
```diff
@@ -330,7 +332,7 @@ export class NewWalletForm {
       page.submitAdd.textContent = intl.prep(intl.ID_CREATE)
     } else {
       Doc.show(page.auth)
-      if (!walletDef.noauth) Doc.show(page.newWalletPassBox)
+      if ((walletDef.traits & traitAuthenticator) !== 0) Doc.show(page.newWalletPassBox)
```
chappjc, line 335: The traits are in `WalletState`.  They are incorrect in the static package-level `WalletDefinition`s.
We have `noWalletPWNeeded`, which considers exactly the case (token wallet) to which the TODO refers.  Why would we need `noauth` or another traits field?

client/webserver/site/src/js/registry.ts
```diff
@@ -212,7 +212,7 @@ export interface WalletDefinition {
   description: string
   configpath: string
   configopts: ConfigOption[]
-  noauth: boolean
+  traits: number
```
chappjc, line 215: Line 177 above in the `WalletState` is where we already have the traits.

chappjc (changes requested):
Let's rewind most of the client/asset changes.  There is no need to make any new wallet types.  The existing types implemented the new interface.  The only client/asset changes that are needed are in the _eth package_.

Adding the `WalletDefinition.Traits` field does not work either, esp. for the btc clones that can embed or expand upon types from btc.  You can only determine wallet traits from the instance you have constructed.

client/core/wallet.go
```diff
@@ -109,8 +109,13 @@ func (w *xcWallet) Unlock(crypter encrypt.Crypter) error {
 	if !w.connected() {
 		return errWalletNotConnected
 	}
+	if !w.traits.IsAuthenticator() {
+		return nil
+	}
+
+	a := w.Wallet.(asset.Authenticator)
```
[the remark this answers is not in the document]
chappjc: This is definitely safest in case the `xcWallet.traits` field somehow becomes in accurate we can avoid a panic.

client/asset/btc/btc.go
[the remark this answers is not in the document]
ukane-philemon: Thanks for the review @JoeGruffins, the reason for the new type was because `zec` had the `NoAuth` field set to `true` and letting `baseWallet` keep the exported methods means it will be called for `zec`.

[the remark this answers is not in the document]
chappjc: Ugh, I missed the zec issue.  That's really unfortunate fallout.

client/webserver/site/src/js/forms.ts
ukane-philemon: The reason for having the triats available here was to accurately tell which wallet is an `asset.Authenticator`(exculding `zec`). Which was the original use for the `noauth` field.

client/asset/btc/btc.go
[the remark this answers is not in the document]
ukane-philemon: Even with the new `asset.Authenticator` we still need to be able to tell the frontend that the wallet deos not require password(esp. for BTC clones like `zec` since seeded and token wallets ain't an issue). We can either do it with `noauth` or  `traits`.

client/asset/eth/eth.go
```diff
@@ -53,6 +53,7 @@ func registerToken(tokenID uint32, desc string, nets ...dex.Network) {
 	asset.RegisterToken(tokenID, token.Token, &asset.WalletDefinition{
 		Type:        "token",
 		Description: desc,
+		NoAuth:      true,
```
buck54321, line 56: Let's work to deprecate this field.

buck54321 (commented):
@ukane-philemon check out this alternative composition for the btc parts.

https://github.com/buck54321/dcrdex/compare/8d99975b34e4ed372953ee39f5fbc659b2302918...buck54321:dcrdex:authenticator

client/asset/eth/eth.go
ukane-philemon: Thanks, for the suggestion. I thought about having a separate method `noauth` clones can use but I was hesitant to make many code changes.

buck54321:
> @ukane-philemon check out this alternative composition for the btc parts.
> 
> [buck54321/dcrdex@8d99975...buck54321:dcrdex:authenticator](https://github.com/buck54321/dcrdex/compare/8d99975b34e4ed372953ee39f5fbc659b2302918...buck54321:dcrdex:authenticator)

I made a couple of changes to this suggestions. The one thing I'm not sure about is the `FeeRater` stuff. ZCash has a constant fee rate (duh), but isn't  a `FeeRater` in my suggestion.

client/asset/eth/eth.go
buck54321: > I thought about having a separate method noauth clones can use 

Hmm. A method for what?

ukane-philemon: For returning a full node that does not support auth, just like you have in your suggestion `btc.NoAuthClone(...)`.

ukane-philemon:
> @ukane-philemon check out this alternative composition for the btc parts.
> 
> [buck54321/dcrdex@8d99975...buck54321:dcrdex:authenticator](https://github.com/buck54321/dcrdex/compare/8d99975b34e4ed372953ee39f5fbc659b2302918...buck54321:dcrdex:authenticator)

Are we just going to fall through for this part? We will no longer have access to the `btc.locked` method.
```
if btc.Locked() {
    return addrStr, nil
}
```

buck54321:
> Are we just going to fall through for this part? We will no longer have access to the `btc.locked` method.
> 
> ```
> if btc.Locked() {
>     return addrStr, nil
> }
> ```

I figured we could just fall through, but I haven't tested anything.

ukane-philemon:
Btw, we are still keeping the `NoAuth` field so the frontend should not show the password field for `zec`, no?

buck54321:
> Btw, we are still keeping the `NoAuth` field so the frontend does not show the password field for `zec`, no?

Ooh. That's what you and @chappjc were talking about. Yeah, I guess we need to keep it.

client/asset/eth/eth.go
chappjc: What's the main goal of this undertaking?  I thought it was the useless button on UI, but we have the exact same solution for that, a flag.  If it's the useless Go Lock/Unlock methods, that's not even a problem to solve IMO, unless it turned out to be super trivial. Is there something problematic in core that having the separate Authenticator type solves for the backend, or is this all about the frontend/walletdefinition?

ukane-philemon: > What's the main goal of this undertaking? 

1. Remove the `noauth` field returned in the wallet definition. But seems there's no better way to hide the wallet password input for `zec` for now without explicitly specifying the `noauth` field.
2. Remove unless Lock/Unlock methods. This could seem to be trivial but calling Go methods on wallets that clearly don't implement them didn't match our existing design, hence the `TODO` in `asset/eth/eth.go`.
3. Future wallet implementations that do not require authentications will not have to implement a method they clearly don't need.

ukane-philemon:
> I figured we could just fall through, but I haven't tested anything.

Brief testing indicates we are going to fall if we fall through.
```
error recovering btc wallet: btc Wallet.Address error: private key unavailable for address tb1qkr3dnuf953mlxfv0n8leflgnv6fa2vgkp2wr7m: address manager is locked
```

client/asset/eth/eth.go
```diff
@@ -133,7 +133,6 @@ var (
 				Description: "Infrastructure providers (e.g. Infura) or local nodes",
 				ConfigOpts:  append(RPCOpts, WalletOpts...),
 				Seeded:      true,
```
buck54321, line 135: Up above in `registerToken`, need to set `NoAuth` in the `WalletDefinition`.

buck54321 (changes requested):

ukane-philemon:
Rebased...

client/asset/zec/zec.go
```diff
@@ -160,7 +160,6 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, net dex.Network) (ass
 		Testnet: "18232",
 		Simnet:  "18232",
 	}
-	var w *btc.ExchangeWalletFullNode
```
buck54321, line 163: With newly merged changes, you'll need to leave this now, and just change it to `var w *btc.ExchangeWalletNoAuth`.

buck54321 (changes requested):

client/asset/btc/btc.go
```diff
@@ -1077,22 +1086,42 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, net dex.Network) (ass
...
 // with minimal coding.
 func BTCCloneWallet(cfg *BTCCloneCFG) (*ExchangeWalletFullNode, error) {
+	iw, err := btcCloneWallet(cfg)
+	if err != nil {
+		return nil, err
+	}
+	return &ExchangeWalletFullNode{iw, &authAddOn{iw.node}}, nil
+}
+
+// BTCCloneWalletNoAuth is like BTCCloneWallet but the wallet created deos not
```
martonp, line 1096: does

martonp (commented):
The ETH and ERC20 pages now contain a field for adding a wallet password, and if you add a wallet password it will fail to create.

martonp:
> I'm wondering what the `containsRequired` part has to do with whether we show the password field? Should it be
> 
> ```ts
> const noWalletPWNeeded = walletDef.noauth || walletDef.seeded || Boolean(this.current.asset.token)
> if (appPwCached && noWalletPWNeeded && !containsRequired) {
> ```

Not sure where that came from. The only `Required` config is ETH's Provider config. I think your replacement looks good.

ukane-philemon:
> The create ETH and ERC20 pages now contain a field for adding a wallet password,

The password field only shows up if the parent wallet has not been created. Since the child wallet will create the parent wallet if it does not already exists, it makes sense to give users a chance to provide a password and used that password for the parent wallet only.

buck54321:
> > The create ETH and ERC20 pages now contain a field for adding a wallet password,
> 
> The password field only shows up if the parent wallet has not been created. Since the child wallet will create the parent wallet if it does not already exists, it makes sense to give users a chance to provide a password and used that password for the parent wallet only.

Seeded wallets don't take user passwords.

ukane-philemon:
Oh, the RPC ETH wallet type is actually seeded, right?

chappjc:
Good to go pending rebase and resolve.

client/asset/btc/btc.go
ukane-philemon: fixed

ukane-philemon:
rebased and squashed.

PR #2070, opened 2023-01-24
ui: normalize input borders

buck54321:
I noticed after #2002 that the checkbox is hard to see in dark mode. 
image
Also, we didn't have consistent input borders site-wide. 
Now all inputs, selects, and checkboxes have a light border in both dark and light mode.
image

### DIFF
--- a/client/webserver/site/src/css/application.scss
+++ b/client/webserver/site/src/css/application.scss
@@ -10,6 +10,8 @@ $light_body_bg: white;
 $dark_body_bg: #091a28;
 $light_border_color: #ddd;
 $dark_border_color: #383f4b;
+$light_input_border: #999;
+$dark_input_border: #555;
 $sans: "source-sans", sans-serif;
 $demi-sans: "demi-sans", sans-serif;
 $mono: "mono", monospace;
--- a/client/webserver/site/src/css/main.scss
+++ b/client/webserver/site/src/css/main.scss
@@ -184,6 +184,13 @@ input[type=number] {
   -moz-appearance: textfield;
 }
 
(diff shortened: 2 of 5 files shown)


client/webserver/site/src/css/application.scss
```diff
@@ -10,6 +10,8 @@ $light_body_bg: white;
 $dark_body_bg: #091a28;
 $light_border_color: #ddd;
 $dark_border_color: #383f4b;
+$light_input_border: #999;
+$dark_input_border: #555;
```
chappjc, line 14: OK, this seems like the established place for these vars, but it would be nice (in the future) if the variables declared here _before_ the bootstrap imports were just the ones intended to customize bootstrap variables (https://getbootstrap.com/docs/5.0/customize/sass/#variable-defaults)  Any others that are dex specific would be better defined after the bootstrap imports, just before the `./` imports below.
I can no longer tell which of these 18 variables fall into that category.

chappjc (approved):

chappjc:
@ukane-philemon @norwnd I see thumbs up from you.  Please submit a review when you get a chance.

ukane-philemon (approved):
Looks good.

norwnd (approved):
Looks good, and seems to be working as expected.

client/webserver/site/src/css/application.scss
chappjc: Merged but I think we should look at the placement of these vars to avoid ambiguity about whether they are local to the application, or doing sass customization of vars used internal to bootstrap.

PR #2071, opened 2023-01-24
asset: display asset symbol in failed connect message

### LINKED ISSUE #2066: Messaging for missing RPC wallets isn't great
When a wallet can't be found on startup, the error message doesn't even inform the user of which wallet it was.

image

ukane-philemon:
Resolves #2066.

Wallet connection warning

### DIFF
--- a/client/core/wallet.go
+++ b/client/core/wallet.go
@@ -340,37 +340,40 @@ func (w *xcWallet) Connect() error {
 	// ConnectOnce so that the ConnectionMaster's On method will report false.
 	err := w.connector.ConnectOnce(context.Background())
 	if err != nil {
-		return fmt.Errorf("%s: %w", unbip(w.AssetID), err)
+		return err
 	}
-	// Now that we are connected, we must Disconnect if any calls fail below
-	// since we are considering this wallet not "hookedUp".
+
+	var ready bool
+	defer func() {
+		// Now that we are connected, we must Disconnect if any calls fail below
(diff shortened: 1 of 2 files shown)


client/core/wallet.go
```diff
@@ -340,15 +340,15 @@ func (w *xcWallet) Connect() error {
 	// ConnectOnce so that the ConnectionMaster's On method will report false.
 	err := w.connector.ConnectOnce(context.Background())
 	if err != nil {
-		return err
+		return fmt.Errorf("%s: %w", unbip(w.AssetID), err)
```
chappjc, line 343: Would be `w.Symbol` for these.
However, the caller knows what wallet it's using, so instead of decorating every error returned from this method with a symbol, it would be better to have the caller do it.  AFAICT, that's only `(*Core).connectWalletResumeTrades`, so change `codedError` there to `newError` and it's done.  That's the pattern in the other methods that surround it.

chappjc (changes requested):

client/core/wallet.go
ukane-philemon: I see we currently do `unbip(w.AssetID)` even when we have access to `w.Symbol`, would you want me to change those too to use `w.Symbol`?

chappjc: Just here is fine.  That field happens to be new is all.

chappjc (approved):

PR #2073, opened 2023-01-25
client/core: remove the immutable trackedTrade.dc field

chappjc:
We have had difficulty with smaller/child types (like `trackedTrade`) carrying references to the larger/parent types (like `dexConnection`, `xcWallet`, etc.).  In addition, having these higher level types available to lower level types encourages some bad patterns that are difficult to break out of.

This removes the `*dexConnection` field from the `trackedTrade` type.

In doing this, I've also _provisionally_ created a default package-level logger for `core` since having logger fields has soured on me. The many `t.dc.log(...` occurrences is also a sign of the first issue this PR is addressing.  If we really have a problem with the package-level logger, we can make a logger field for `trackedTrade`.  However, I feel we have been bending over backwards to accommodate load testing utilities, while there are virtually no conceivable uses for an application with multiple instances of `Core` other than load testing utilities.  Even a production bot should have a single `Core` instances.  If it needs two, there should be two standalone apps.

@ukane-philemon re https://github.com/decred/dcrdex/pull/2019#discussion_r1086041439

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -89,7 +89,13 @@ var (
 	walletDisabledErrStr = "%s wallet is disabled"
 
 	errTimeout = errors.New("timeout")
+
+	log = dex.StdOutLogger("core", dex.DefaultLogLevel) // default logger
 )
+
+func UseLogger(l dex.Logger) {
+	log = l
+}
 
 type dexTicker struct {
 	dur int64 // atomic
@@ -1205,7 +1211,7 @@ func (c *Core) tickAsset(dc *dexConnection, assetID uint32) assetMap {
 		}
 		trade := trade // bad go, bad
 		go func() {
-			newUpdates, err := c.tick(trade)
+			newUpdates, err := c.tick(&dexTrade{trade, dc})
 			if err != nil {
 				c.log.Errorf("%s tick error: %v", dc.acct.host, err)
 			}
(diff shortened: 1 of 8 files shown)


chappjc:
Haven't actually tried using this, but conceptual acks (or nacks) welcome on this idea.
I don't really mind adding a logger field to the `trackedTrade`, but it really doesn't help anything, just extra code imo.

This change is similar to how we had issues with old/replaced wallets in the walletSet field, and we ended up having to write an update wallet method, but with the dexConnection.  Instead of trying to solve it the same way, with updaters and even more concurrency controls, I decided to try breaking out of the somewhat backwards pattern of holding a reference to a parent type.  When I see that, I think that responsibilities have been reversed (dexConnection methods operating on trades rather than trade method doing higher level operations via their dexConnection pointer), and it seem to be the case in resolving this.

client/core/core.go
```diff
@@ -1512,6 +1512,12 @@ func (c *Core) Run(ctx context.Context) {
 	// Store the context as a field, since we will need to spawn new DEX threads
 	// when new accounts are registered.
 	c.ctx = ctx
+	defer func() {
+		if pv := recover(); pv != nil {
+			c.log.Critical(pv)
+			close(c.ready) // unblock <-Ready()
```
chappjc, line 1518: mm, maybe I will just add and error return to `initialize` since it's possible the channel was already closed if it panicked *after* initialize rather than in it... meh

client/core/core.go
```diff
@@ -1397,7 +1403,7 @@ type Core struct {
 // New is the constructor for a new Core.
 func New(cfg *Config) (*Core, error) {
 	if cfg.Logger == nil {
-		return nil, fmt.Errorf("Core.Config must specify a Logger")
+		cfg.Logger = log // use the package level-logger, may be stdout
```
buck54321, line 1406: Should we panic if `log == nil` here?

buck54321 (commented):
Concept ACK

chappjc:
May revive, but not essential to DEX evolution.

PR #2074, opened 2023-01-25
client/eth: Check provider header times.

### LINKED ISSUE #2064: client/eth: Not synced geth nodes are used.
I was testing on testnet with two providers, an unsynced geth and infura. The client was failing to find confirmations and later to send redemptions until I removed the un-synced geth node. Client should not use providers it can tell are not synced.

JoeGruffins:
When fetching a new or cached header with a provider, do a basic check on the header's time to determine if the header, and so the provider, are up to date.

closes #2064

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -325,7 +325,7 @@ type ethFetcher interface {
 	sendSignedTransaction(ctx context.Context, tx *types.Transaction) error
 	sendTransaction(ctx context.Context, txOpts *bind.TransactOpts, to common.Address, data []byte) (*types.Transaction, error)
 	signData(data []byte) (sig, pubKey []byte, err error)
-	syncProgress(context.Context) (*ethereum.SyncProgress, error)
+	syncProgress(context.Context) (progress *ethereum.SyncProgress, tipTime uint64, err error)
 	transactionConfirmations(context.Context, common.Hash) (uint32, error)
 	getTransaction(context.Context, common.Hash) (*types.Transaction, int64, error)
 	txOpts(ctx context.Context, val, maxGas uint64, maxFeeRate, nonce *big.Int) (*bind.TransactOpts, error)
@@ -622,7 +622,7 @@ func createWallet(createWalletParams *asset.CreateWalletParams, skipConnect bool
 			}
 
 			if len(unknownEndpoints) > 0 && createWalletParams.Net == dex.Mainnet {
-				providers, err := connectProviders(ctx, unknownEndpoints, createWalletParams.Logger, big.NewInt(chainIDs[createWalletParams.Net]))
+				providers, err := connectProviders(ctx, unknownEndpoints, createWalletParams.Logger, big.NewInt(chainIDs[createWalletParams.Net]), createWalletParams.Net)
 				if err != nil {
 					return err
 				}
@@ -2885,17 +2885,13 @@ func (*baseWallet) ValidateSecret(secret, secretHash []byte) bool {
 // more, requesting the best block header starts to fail after a few tries
 // during initial sync. Investigate how to get correct sync progress.
 func (eth *baseWallet) SyncStatus() (bool, float32, error) {
-	prog, err := eth.node.syncProgress(eth.ctx)
+	prog, tipTime, err := eth.node.syncProgress(eth.ctx)
 	if err != nil {
 		return false, 0, err
 	}
 	checkHeaderTime := func() (bool, error) {
-		bh, err := eth.node.bestHeader(eth.ctx)
-		if err != nil {
-			return false, err
-		}
 		// Time in the header is in seconds.
-		timeDiff := time.Now().Unix() - int64(bh.Time)
+		timeDiff := time.Now().Unix() - int64(tipTime)
 		if timeDiff > dexeth.MaxBlockInterval && eth.net != dex.Simnet {
 			eth.log.Infof("Time since last eth block (%d sec) exceeds %d sec."+
 				"Assuming not in sync. Ensure your computer's system clock "+
--- a/client/asset/eth/eth_test.go
+++ b/client/asset/eth/eth_test.go
@@ -97,6 +97,8 @@ type testNode struct {
 	bestHdr         *types.Header
 	bestHdrErr      error
 	syncProg        ethereum.SyncProgress
+	syncProgT       uint64
+	syncProgErr     error
 	bal             *big.Int
 	balErr          error
 	signDataErr     error
@@ -206,8 +208,8 @@ func (n *testNode) lock() error {
 func (n *testNode) locked() bool {
 	return false
 }
-func (n *testNode) syncProgress(context.Context) (*ethereum.SyncProgress, error) {
-	return &n.syncProg, nil
+func (n *testNode) syncProgress(context.Context) (prog *ethereum.SyncProgress, bestBlockUNIXTime uint64, err error) {
+	return &n.syncProg, n.syncProgT, n.syncProgErr
 }
 func (n *testNode) peerCount() uint32 {
 	return 1
@@ -465,8 +467,8 @@ func TestSyncStatus(t *testing.T) {
 	tests := []struct {
 		name                string
 		syncProg            ethereum.SyncProgress
+		syncProgErr         error
 		subSecs             uint64
-		bestHdrErr          error
 		wantErr, wantSynced bool
 		wantRatio           float32
 	}{{
@@ -492,22 +494,22 @@ func TestSyncStatus(t *testing.T) {
 		},
 		subSecs: dexeth.MaxBlockInterval + 1,
 	}, {
-		name:       "best header error",
-		bestHdrErr: errors.New(""),
+		name: "sync progress error",
 		syncProg: ethereum.SyncProgress{
 			CurrentBlock: 25,
 			HighestBlock: 0,
 		},
-		wantErr: true,
+		syncProgErr: errors.New(""),
+		wantErr:     true,
 	}}
 
 	for _, test := range tests {
 		nowInSecs := uint64(time.Now().Unix())
 		ctx, cancel := context.WithCancel(context.Background())
 		node := &testNode{
-			syncProg:   test.syncProg,
-			bestHdr:    &types.Header{Time: nowInSecs - test.subSecs},
-			bestHdrErr: test.bestHdrErr,
+			syncProg:    test.syncProg,
+			syncProgT:   nowInSecs - test.subSecs,
+			syncProgErr: test.syncProgErr,
 		}
 		eth := &baseWallet{
 			node: node,
--- a/client/asset/eth/multirpc.go
+++ b/client/asset/eth/multirpc.go
@@ -42,6 +42,10 @@ const (
 	// failQuarantine is how long we will wait after a failed request before
 	// trying a provider again.
 	failQuarantine = time.Minute
+	// headerCheckInterval is the time between header checks. Slightly less
+	// than the fail quarantine to ensure providers with old headers stay
+	// quarantined.
+	headerCheckInterval = time.Second * 50
 	// receiptCacheExpiration is how long we will track a receipt after the
 	// last request. There is no persistent storage, so all receipts are cached
 	// in-memory.
@@ -83,7 +87,9 @@ type provider struct {
 	host    string
 	ec      *combinedRPCClient
 	ws      bool
+	net     dex.Network
 	tipCapV atomic.Value // *cachedTipCap
+	stop    func()
 
 	// tip tracks the best known header as well as any error encountered
 	tip struct {
@@ -95,13 +101,22 @@ type provider struct {
 	}
 }
 
-func (p *provider) setTip(header *types.Header) {
+func (p *provider) shutdown() {
+	p.stop()
+	p.ec.Close()
+}
+
+func (p *provider) setTip(header *types.Header, log dex.Logger) {
 	p.tip.Lock()
 	p.tip.header = header
 	p.tip.headerStamp = time.Now()
 	p.tip.failStamp = time.Time{}
+	unfailed := p.tip.failCount != 0
 	p.tip.failCount = 0
 	p.tip.Unlock()
+	if unfailed {
+		log.Debugf("Provider at %s was failed but is now useable again.", p.host)
+	}
 }
 
 // cachedTip retrieves the last known best header.
@@ -117,6 +132,7 @@ func (p *provider) cachedTip() *types.Header {
 
 	p.tip.RLock()
 	defer p.tip.RUnlock()
+
 	if time.Since(p.tip.failStamp) < failQuarantine || time.Since(p.tip.headerStamp) > stale {
 		return nil
 	}
@@ -142,27 +158,30 @@ func (p *provider) failed() bool {
 // bestHeader get the best known header from the provider, cached if available,
 // otherwise a new RPC call is made.
 func (p *provider) bestHeader(ctx context.Context, log dex.Logger) (*types.Header, error) {
-	ctx, cancel := context.WithTimeout(ctx, defaultRequestTimeout)
-	defer cancel()
 	// Check if we have a cached header.
 	if tip := p.cachedTip(); tip != nil {
-		log.Tracef("using cached header from %q", p.host)
+		log.Tracef("Using cached header from %q", p.host)
 		return tip, nil
 	}
 
-	log.Tracef("fetching fresh header from %q", p.host)
+	log.Tracef("Fetching fresh header from %q", p.host)
 	hdr, err := p.ec.HeaderByNumber(ctx, nil /* latest */)
 	if err != nil {
 		p.setFailed()
 		return nil, fmt.Errorf("HeaderByNumber error: %w", err)
 	}
-	p.setTip(hdr)
+	timeDiff := time.Now().Unix() - int64(hdr.Time)
+	if timeDiff > dexeth.MaxBlockInterval && p.net != dex.Simnet {
+		p.setFailed()
+		return nil, fmt.Errorf("time since last eth block (%d sec) exceeds %d sec. "+
+			"Assuming provider %s is not in sync. Ensure your computer's system clock "+
+			"is correct.", timeDiff, dexeth.MaxBlockInterval, p.host)
+	}
+	p.setTip(hdr, log)
 	return hdr, nil
 }
 
 func (p *provider) headerByHash(ctx context.Context, h common.Hash) (*types.Header, error) {
-	ctx, cancel := context.WithTimeout(ctx, defaultRequestTimeout)
-	defer cancel()
 	hdr, err := p.ec.HeaderByHash(ctx, h)
 	if err != nil {
 		p.setFailed()
@@ -174,8 +193,6 @@ func (p *provider) headerByHash(ctx context.Context, h common.Hash) (*types.Head
 // suggestTipCap returns a tip cap suggestion, cached if available, otherwise a
 // new RPC call is made.
 func (p *provider) suggestTipCap(ctx context.Context, log dex.Logger) *big.Int {
-	ctx, cancel := context.WithTimeout(ctx, defaultRequestTimeout)
-	defer cancel()
 	if cachedV := p.tipCapV.Load(); cachedV != nil {
 		rec := cachedV.(*cachedTipCap)
 		if time.Since(rec.stamp) < tipCapSuggestionExpiration {
@@ -200,6 +217,32 @@ func (p *provider) suggestTipCap(ctx context.Context, log dex.Logger) *big.Int {
 	})
 
 	return tipCap
+}
+
+// refreshHeader fetches a header every headerCheckInterval. This keeps the
+// cached header up to date or fails the provider if there is a problem getting
+// the header.
+func (p *provider) refreshHeader(ctx context.Context, log dex.Logger) {
+	log.Tracef("handling header refreshes for %q", p.host)
+	ticker := time.NewTicker(headerCheckInterval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ticker.C:
+			// Fetching the best header will check that either the
+			// provider's cached header is not too old or that a
+			// newly fetched header is not too old. If it is too
+			// old that indicates the provider is not in sync and
+			// should not be used.
+			innerCtx, cancel := context.WithTimeout(ctx, defaultRequestTimeout)
+			if _, err := p.bestHeader(innerCtx, log); err != nil {
+				log.Warnf("Problem getting best header from provider %s: %s.", p.host, err)
+			}
+			cancel()
+		case <-ctx.Done():
+			return
+		}
+	}
 }
 
 // subscribeHeaders starts a listening loop for header updates for a provider.
@@ -256,7 +299,7 @@ func (p *provider) subscribeHeaders(ctx context.Context, sub ethereum.Subscripti
 		select {
 		case hdr := <-h:
 			log.Tracef("%q reported new tip at height %s (%s)", p.host, hdr.Number, hdr.Hash())
-			p.setTip(hdr)
+			p.setTip(hdr, log)
 		case err, ok := <-sub.Err():
 			if !ok {
 				// Subscription cancelled
@@ -300,6 +343,7 @@ type multiRPCClient struct {
 	creds   *accountCredentials
 	log     dex.Logger
 	chainID *big.Int
+	net     dex.Network
 
 	providerMtx sync.RWMutex
 	endpoints   []string
@@ -336,6 +380,7 @@ func newMultiRPCClient(dir string, endpoints []string, log dex.Logger, cfg *para
 	}
 
 	m := &multiRPCClient{
+		net:       net,
 		cfg:       cfg,
 		log:       log,
 		creds:     creds,
@@ -352,14 +397,14 @@ func newMultiRPCClient(dir string, endpoints []string, log dex.Logger, cfg *para
 // list of providers that were successfully connected. It is not an error for a
 // connection to fail. The caller can infer failed connections from the length
 // and contents of the returned provider list.
-func connectProviders(ctx context.Context, endpoints []string, log dex.Logger, chainID *big.Int) ([]*provider, error) {
+func connectProviders(ctx context.Context, endpoints []string, log dex.Logger, chainID *big.Int, net dex.Network) ([]*provider, error) {
 	providers := make([]*provider, 0, len(endpoints))
 	var success bool
 
 	defer func() {
 		if !success {
 			for _, p := range providers {
-				p.ec.Close()
+				p.shutdown()
 			}
 		}
 	}()
@@ -467,18 +512,38 @@ func connectProviders(ctx context.Context, endpoints []string, log dex.Logger, c
 		p := &provider{
 			host: host,
 			ws:   sub != nil,
+			net:  net,
 			ec: &combinedRPCClient{
 				Client: ec,
 				rpc:    rpcClient,
 			},
 		}
-		p.setTip(hdr)
-		providers = append(providers, p)
+		p.setTip(hdr, log)
+
+		ctx, cancel := context.WithCancel(ctx)
+		var wg sync.WaitGroup
 
 		// Start websocket listen loop.
 		if sub != nil {
-			go p.subscribeHeaders(ctx, sub, h, log)
+			wg.Add(1)
+			go func() {
+				p.subscribeHeaders(ctx, sub, h, log)
+				wg.Done()
+			}()
 		}
+		wg.Add(1)
+		go func() {
+			p.refreshHeader(ctx, log)
+			wg.Done()
+		}()
+
+		p.stop = func() {
+			cancel()
+			wg.Wait()
+		}
+
+		providers = append(providers, p)
+
 		return nil
 	}
 
@@ -499,7 +564,7 @@ func connectProviders(ctx context.Context, endpoints []string, log dex.Logger, c
 }
 
 func (m *multiRPCClient) connect(ctx context.Context) (err error) {
-	providers, err := connectProviders(ctx, m.endpoints, m.log, m.chainID)
+	providers, err := connectProviders(ctx, m.endpoints, m.log, m.chainID, m.net)
 	if err != nil {
 		return err
 	}
@@ -524,7 +589,7 @@ func (m *multiRPCClient) connect(ctx context.Context) (err error) {
 	go func() {
 		<-ctx.Done()
 		for _, p := range m.providerList() {
-			p.ec.Close()
+			p.shutdown()
 		}
 	}()
 
@@ -575,7 +640,7 @@ func (m *multiRPCClient) reconfigure(ctx context.Context, settings map[string]st
 		return errors.New("no providers specified")
 	}
 	endpoints := strings.Split(providerDef, " ")
-	providers, err := connectProviders(ctx, endpoints, m.log, m.chainID)
+	providers, err := connectProviders(ctx, endpoints, m.log, m.chainID, m.net)
 	if err != nil {
 		return err
 	}
@@ -585,7 +650,7 @@ func (m *multiRPCClient) reconfigure(ctx context.Context, settings map[string]st
 	m.endpoints = endpoints
 	m.providerMtx.Unlock()
 	for _, p := range oldProviders {
-		p.ec.Close()
+		p.shutdown()
 	}
 	return nil
 }
@@ -639,9 +704,7 @@ func (m *multiRPCClient) transactionReceipt(ctx context.Context, txHash common.H
 	}
 
 	// Fetch a fresh one.
-	if err = m.withPreferred(func(p *provider) error {
-		ctx, cancel := context.WithTimeout(ctx, defaultRequestTimeout)
-		defer cancel()
+	if err = m.withPreferred(ctx, func(ctx context.Context, p *provider) error {
 		r, err = p.ec.TransactionReceipt(ctx, txHash)
 		return err
 	}); err != nil {
@@ -690,8 +753,6 @@ func (tx *rpcTransaction) UnmarshalJSON(b []byte) error {
 }
 
 func getRPCTransaction(ctx context.Context, p *provider, txHash common.Hash) (*rpcTransaction, error) {
-	ctx, cancel := context.WithTimeout(ctx, defaultRequestTimeout)
-	defer cancel()
 	var resp *rpcTransaction
 	err := p.ec.rpc.CallContext(ctx, &resp, "eth_getTransactionByHash", txHash)
 	if err != nil {
@@ -708,7 +769,7 @@ func getRPCTransaction(ctx context.Context, p *provider, txHash common.Hash) (*r
 }
 
 func (m *multiRPCClient) getTransaction(ctx context.Context, txHash common.Hash) (tx *types.Transaction, h int64, err error) {
-	return tx, h, m.withPreferred(func(p *provider) error {
+	return tx, h, m.withPreferred(ctx, func(ctx context.Context, p *provider) error {
 		resp, err := getRPCTransaction(ctx, p, txHash)
 		if err != nil {
 			if isNotFoundError(err) {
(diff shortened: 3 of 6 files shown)


client/asset/eth/eth.go
```diff
@@ -295,7 +295,7 @@ type ethFetcher interface {
 	sendSignedTransaction(ctx context.Context, tx *types.Transaction) error
 	sendTransaction(ctx context.Context, txOpts *bind.TransactOpts, to common.Address, data []byte) (*types.Transaction, error)
 	signData(data []byte) (sig, pubKey []byte, err error)
-	syncProgress(context.Context) (*ethereum.SyncProgress, error)
+	syncProgress(context.Context) (progress *ethereum.SyncProgress, bestHeaderUNIXTime uint64, err error)
```
JoeGruffins, line 298: The providers fetch the block header anyway, so this is an optimization that cuts off one header fetch for providers.

client/asset/eth/eth.go
```diff
@@ -2798,18 +2798,14 @@ func (*baseWallet) ValidateSecret(secret, secretHash []byte) bool {
...
 		return false, 0, err
 	}
 	checkHeaderTime := func() (bool, error) {
-		bh, err := eth.node.bestHeader(eth.ctx)
-		if err != nil {
-			return false, err
-		}
 		// Time in the header is in seconds.
-		timeDiff := time.Now().Unix() - int64(bh.Time)
-		if timeDiff > dexeth.MaxBlockInterval && eth.net != dex.Simnet {
```
JoeGruffins, line 2812: Removing the simnet check and mining some blocks in the harness tests. It seems reasonable to just mine some blocks when testing. May be a preference thing so I can revert.

JoeGruffins:
I guess it's possible for a provider to get out of sync and still not be found to have an old header if that provider randomly doesn't call best header but later is randomly asked to do something else. Need to look at this a little bit more.

chappjc:
Heck I think we'll have to be able to work without even `eth_syncing`.  With blastapi.io:

`[ERR] MKT: Not starting dcr_usdc.eth market because of ChainsSynced error: error checking sync status for 60001: Method not found`

chappjc:
Wait a sec, we don't use `eth_syncing` at all, do we?

In fact, I think `providerAnkr` should be moved to the `compliantProviders` map because there's no eth_syncing use at all right?

Similarly, I think we can add `providerBlast         = "blastapi.io"` to client's multirpc compliant providers.

chappjc:
For server, is there an issue with this?

```diff
diff --git a/server/asset/eth/eth.go b/server/asset/eth/eth.go
index e541304ef..8e11349a7 100644
--- a/server/asset/eth/eth.go
+++ b/server/asset/eth/eth.go
@@ -565,10 +565,7 @@ func (eth *baseBackend) Synced() (bool, error) {
        // node.SyncProgress will return nil both before syncing has begun and
        // after it has finished. In order to discern when syncing has begun,
        // check that the best header came in under MaxBlockInterval.
-       sp, err := eth.node.syncProgress(eth.ctx)
-       if err != nil {
-               return false, err
-       }
+       sp, _ := eth.node.syncProgress(eth.ctx)
        if sp != nil {
                return false, nil
        }
```

If there's no `eth_syncing` support, fall back to header age.  We should do that.

JoeGruffins:
> Wait a sec, we don't use eth_syncing at all, do we?

It seems pretty useless and confusing from the very beginning. We can just do the header time everywhere. Thats what it will always come down to currently.

Server only recently (this week) became useable with providers so probably some things we weren't thinking about pop up.

client/asset/eth/multirpc.go
```diff
@@ -784,11 +781,26 @@ func errorFilter(err error, matches ...interface{}) bool {
 }
 
 // withOne runs the provider function against the providers in order until one
-// succeeds or all have failed.
-func (m *multiRPCClient) withOne(providers []*provider, f func(*provider) error, acceptabilityFilters ...acceptabilityFilter) (superError error) {
+// succeeds or all have failed. The context used to run functions has a time
+// limit equal to defaultRequestTimeout for all requests to return. If
+// operations are expected to run longer than that the calling function should
+// not use the altered context.
+func (m *multiRPCClient) withOne(ctx context.Context, providers []*provider, f func(context.Context, *provider) error, acceptabilityFilters ...acceptabilityFilter) (superError error) {
```
JoeGruffins, line 788: So, I needed a ctx to check the header time... and so I thought it might be better to set the context deadline in this function to ensure all the functions that use providers have a deadline. I think it is easier to keep up with than setting them all separately. You can see a couple cases where this timeout is not used, but in the vast majority we are just running one rpc quest with these so I think this is cleaner. This isn't strictly needed to solve the issue however.

JoeGruffins:
> In fact, I think providerAnkr should be moved to the compliantProviders map because there's no eth_syncing use at all right?
>
> Similarly, I think we can add providerBlast         = "blastapi.io" to client's multirpc compliant providers.

Moved these two to the compliant providers.

client/asset/eth/multirpc.go
```diff
@@ -1031,14 +1043,16 @@ func (m *multiRPCClient) signData(data []byte) (sig, pubKey []byte, err error) {
...
+// syncProgress: Current and Highest blocks are not very useful for the caller,
+// but the best header's time in seconds can be used to determine if the
+// provider is out of sync.
+func (m *multiRPCClient) syncProgress(ctx context.Context) (prog *ethereum.SyncProgress, bestHeaderUNIXTime uint64, err error) {
+	return prog, bestHeaderUNIXTime, m.withAny(ctx, func(ctx context.Context, p *provider) error {
 		tip, err := p.bestHeader(ctx, m.log)
 		if err != nil {
 			return err
 		}
+		bestHeaderUNIXTime = tip.Time
```
chappjc, line 1055: Good call returning this instead of doing bestHeader again in the caller.

client/asset/eth/multirpc.go
```diff
@@ -1242,10 +1244,11 @@ const (
 	providerInfura      = "infura.io"
 	providerRivetCloud  = "rivet.cloud"
 	providerAlchemy     = "alchemy.com"
+	providerAnkr        = "ankr.com"
+	providerBlast       = "blastapi.io"
 
 	// Non-compliant providers
 	providerCloudflareETH = "cloudflare-eth.com" // "SuggestGasTipCap" error: Method not found
-	providerAnkr          = "ankr.com"           // "SyncProgress" error: the method eth_syncing does not exist/is not available
```
chappjc, line 1248: I keep meaning to come back to this, but even before this diff, I had been using ankr and it was never treated as non-compliant nor did it have to run a compliance check (I searched the logs, and I don't have a compliant-providers.json anywhere on my drive).  Not sure how it was workin.  The host was https://rpc.ankr.com/eth_goerli

client/asset/eth/multirpc.go
```diff
@@ -113,7 +113,9 @@ func (p *provider) cachedTip() *types.Header {
 
 	p.tip.RLock()
 	defer p.tip.RUnlock()
-	if time.Since(p.tip.failStamp) < failQuarantine || time.Since(p.tip.headerStamp) > stale {
+	if time.Since(p.tip.failStamp) < failQuarantine ||
+		time.Since(p.tip.headerStamp) > stale ||
+		time.Now().Unix()-int64(p.tip.header.Time) > dexeth.MaxBlockInterval {
```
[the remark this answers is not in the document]
JoeGruffins: I got the impression on matrix that @buck54321 is going to do some things around here.

JoeGruffins:
Just rebased.

JoeGruffins:
I broke the the multirpc live test by not calling `m.Run` in `TestMain`. Fixed now https://github.com/decred/dcrdex/compare/589fcfe6e187f26a46b18b7ccb20e36017c8a6bb..6f3d98ec538934215f917ad05aa846e16e2c69e1

But, tests are not passing. Looking into that and the `compliant providers` testing on testnet.

JoeGruffins:
Removed a bad comment. https://github.com/decred/dcrdex/compare/6f3d98ec538934215f917ad05aa846e16e2c69e1..c7015102c8a5f64156f30728162adb9fd2e46803

JoeGruffins:
The test failures in the multirpc live test are solved by #2091 so nothing to do here about that.

JoeGruffins:
> I had been using ankr and it was never treated as non-compliant nor did it have to run a compliance check

It looks like nothing happens if a provider is on the non-compliant list atm, it only checks unknown providers. Should it just not use providers on the bad list, or maybe check them again?

chappjc:
> Should it just not use providers on the bad list, or maybe check them again?

I don't think there's a point in the non-compliant providers really.  Why not have it check all unknown?  Ankr worked for a long time for me and it was on the non-compliant providers list.  Providers can change and the list can become stale... or the RPCs we use can change (no eth_syncing anymore) and the list can become stale.

JoeGruffins:
> Providers can change and the list can become stale... or the RPCs we use can change (no eth_syncing anymore) and the list can become stale.

Maybe saving a check timestamp with the providers, and check every month or so even if known? Could also add a version for forced checks when we changed things, although keeping up with versions is a pain point. Can also just check all providers every start up, May be worth it.

chappjc:
Potentially all good ideas, but I bet the compliant providers will stay compliant.  IMO, for now, just checking providers that are not on the compliant list is a start though.  I don't see any upside to the non-compliant list given we have a check.  Maybe @buck54321 had something in mind that we aren't thinking of.

client/asset/eth/multirpc.go
```diff
@@ -784,11 +781,26 @@ func errorFilter(err error, matches ...interface{}) bool {
...
+			// provider's cached header is not too old or that a
+			// newly fetched header is not too old. If it is too
+			// old that indicates the provider is not in sync and
+			// should not be used.
+			_, err := p.bestHeader(ctx, m.log)
+			cancel()
+			if err != nil {
+				m.log.Warnf("Problem getting best header: %s.", err)
+				continue
+			}
```
buck54321, line 803: `bestHeader` is called on a random provider from the list every second. I think we can rely on that to keep `failed` up to date.

buck54321 (changes requested):

JoeGruffins:
Just rebsed.

client/asset/eth/eth.go
```diff
@@ -2806,18 +2806,14 @@ func (*baseWallet) ValidateSecret(secret, secretHash []byte) bool {
 // more, requesting the best block header starts to fail after a few tries
 // during initial sync. Investigate how to get correct sync progress.
 func (eth *baseWallet) SyncStatus() (bool, float32, error) {
-	prog, err := eth.node.syncProgress(eth.ctx)
+	prog, bestHeaderUNIXTime, err := eth.node.syncProgress(eth.ctx)
```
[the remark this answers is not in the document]
JoeGruffins: changed bestHeaderUNIXTime -> tipTime

client/asset/eth/eth_test.go
```diff
@@ -91,43 +91,45 @@ type tGetTxRes struct {
...
-	tokenContractor *tTokenContractor
-	contractor      contractor
-	tokenParent     *assetWallet // only set for tokens
+	acct                       *accounts.Account
+	addr                       common.Address
+	connectErr                 error
+	bestHdr                    *types.Header
+	bestHdrErr                 error
+	syncProg                   ethereum.SyncProgress
+	syncProgBestHeaderUNIXTime uint64
```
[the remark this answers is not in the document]
JoeGruffins: changed bestHeaderUNIXTime -> tipTime

client/asset/eth/multirpc.go
JoeGruffins: You mean in `checkForNewBlocks`?

I guess it depends on how many providers... it would not be guaranteed unless one provider.

client/asset/eth/multirpc.go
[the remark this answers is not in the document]
JoeGruffins: Removing the stale check.

client/asset/eth/multirpc.go
JoeGruffins: Ok. The extra ctx in `currentFees` and `transactionConfirmations` were just because two requests were being done with the same limit. But, they probably only should take milliseconds, so probably fine. I was thinking `nextNonce` was using the same context the whole time, but now that I look again it is using a new `withPreferred` every time so a new timeout. I was just confused it seems.

client/asset/eth/eth.go
JoeGruffins: Well, ok. Also passing the net to prividers `bestHeader` so it can also not care.

client/asset/eth/multirpc.go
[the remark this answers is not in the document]
JoeGruffins: Or, actually,  with no stale check it only fetches a new header after `dexeth.MaxBlockInterval`... So maybe one block interval after that should fetch a new one.

JoeGruffins:
Fixing most of @buck54321 comments https://github.com/decred/dcrdex/compare/eddd3d3e88d6cccb0ee35b2b3cc4b691992d6924..8c4ae5ffd8e49be3f28046513e57594b9a1b6453

I don't know if this is breaking what was intended with the stale check? It's basically a shorter stale check. We want the current header without requesting needlessly I guess. So, requesting after a block should have been mined seems prudent. However, if it is the same block, then it will be pulled again the next check. The previous check was from when it was pulled, so time is different and behavior is different after header time is too old.

I like the header check in `withOne`. With the cache, if it is not too old it does not make any requests.

client/asset/eth/multirpc.go
buck54321: Can we do this check down in the `for _, p := range readyProviders` loop right before we actually use the provider, so that if the first one passes we're not wasting calls on the rest?

client/asset/eth/eth_test.go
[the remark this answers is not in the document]
buck54321: Can you change this one too so that it doesn't blow up the diff

client/asset/eth/multirpc.go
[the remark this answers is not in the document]
buck54321: After seeing it now, I think maybe we did need the `stale` check. It's the `MaxBlockInterval` check that could have been deleted, since `stale` was always more restrictive. But we definitely don't want to check after 13 seconds when we have a websocket connection.

client/asset/eth/multirpc.go
JoeGruffins: Moved to the loop.

client/asset/eth/multirpc.go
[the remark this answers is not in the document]
JoeGruffins: Put it back. The header can still be 180 seconds old though. Thats a few blocks.

JoeGruffins:
Moving the header check down and putting the stale check back as it was https://github.com/decred/dcrdex/compare/954492242bdf7bcce95f40daff83fd94aecf3a7d..fb774e1cdf0f11f977aea4945b04b3c6a2216353

JoeGruffins:
Header can still be 180 seconds old. Is that good enough? The provider could be 13 blocks behind and still be ok.

chappjc:
> Header can still be 180 seconds old. Is that good enough? The provider could be 13 blocks behind and still be ok.

It's not great, but as long as it doesn't fall further behind, that's enough time buffer to do what it needs to do, I believe.

client/asset/eth/multirpc.go
```diff
@@ -803,7 +806,19 @@ func (m *multiRPCClient) withOne(providers []*provider, f func(*provider) error,
...
+		if _, err := p.bestHeader(ctx, m.log, m.net); err != nil {
+			cancel()
+			m.log.Warnf("Problem getting best header: %s.", err)
+			continue
+		}
+		err := f(ctx, p)
+		cancel()
 		if err == nil {
 			break
 		}
```
buck54321, line 824: Since a `readyProvider` could potentially never run now, we need to do two things. 1) change this to 

```Go
if err == nil {
    return nil
}
```

and then 2) if we make it out of the `readyProviders` loop, it now means that all providers failed or weren't up to date, so we would return an error.

I actually think that `break`ing when `err == nil` was incorrect anyway, since `superError` could have been set in a previous iteration.

client/asset/eth/multirpc.go
```diff
@@ -141,28 +144,31 @@ func (p *provider) failed() bool {
 
 // bestHeader get the best known header from the provider, cached if available,
 // otherwise a new RPC call is made.
-func (p *provider) bestHeader(ctx context.Context, log dex.Logger) (*types.Header, error) {
-	ctx, cancel := context.WithTimeout(ctx, defaultRequestTimeout)
-	defer cancel()
+func (p *provider) bestHeader(ctx context.Context, log dex.Logger, net dex.Network) (*types.Header, error) {
```
buck54321, line 147: You could also make `net` a field on `provider`, in which case you would add the argument to `connectProviders`.

buck54321 (changes requested):

JoeGruffins:
Just rebased.

client/asset/eth/multirpc.go
JoeGruffins: Done.

client/asset/eth/multirpc.go
JoeGruffins: Doing that.

client/asset/eth/eth_test.go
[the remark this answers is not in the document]
JoeGruffins: syncProgBestHeaderUNIXTime -> syncProgT

JoeGruffins:
Addressing @buck54321 last review https://github.com/decred/dcrdex/compare/1f4d011c4244de96bed7b410f540c6295b2117fe..e598ecda67878df8bc302fd6e88ba60803bfed76

client/asset/eth/multirpc.go
```diff
@@ -106,6 +107,9 @@ func (p *provider) setTip(header *types.Header) {
 
 // cachedTip retrieves the last known best header.
 func (p *provider) cachedTip() *types.Header {
+	p.tip.RLock()
+	defer p.tip.RUnlock()
```
chappjc, line 111: The RLock/RUnlock are still below though.  Mean to move them up here (remove them from below)?

client/asset/eth/multirpc.go
```diff
@@ -803,9 +808,21 @@ func (m *multiRPCClient) withOne(providers []*provider, f func(*provider) error,
...
 	}
 	for _, p := range readyProviders {
-		err := f(p)
+		ctx, cancel := context.WithTimeout(ctx, defaultRequestTimeout)
+		// Fetching the best header will check that either the
+		// provider's cached header is not too old or that a
+		// newly fetched header is not too old. If it is too
+		// old that indicates the provider is not in sync and
+		// should not be used.
+		if _, err := p.bestHeader(ctx, m.log); err != nil {
```
chappjc, line 817: So in `(*multiRPCClient).bestHeader`, it's going to do: https://github.com/decred/dcrdex/blob/e598ecda67878df8bc302fd6e88ba60803bfed76/client/asset/eth/multirpc.go#L960-L963

which means that through `withAny` -> `withOne`, we're going to get here and do `(*provider).bestHeader` twice in a row?

I kinda wish something other than `withOne` were in charge of flagging providers as stale.  `(*ETHWallet).checkForNewBlocks` is going to be calling `bestHeader` regularly anyway, so it seems like either `(*provider) bestHeader` or `(*multiRPCClient) bestHeader`would be a good place to set some flag, which `withOne` could then use in a similar manner to how it gets `readyProviders` with `!p.failed()`.

client/asset/eth/multirpc.go
JoeGruffins: Yes this looks like a mistake. Fixing.

client/asset/eth/multirpc.go
JoeGruffins: Can I stick the check in a goroutine? I'm afraid if you just put the check in a random call, it randomly might not get called while another random call is called and the stale provider is used.

JoeGruffins: While adding the goroutine, it seems like the other goroutine `subscribeHeaders` will continue to run on old providers if you reconfigure, although it does not do anything, it's just waiting. Maybe should be stopped when the rpc client is closed.

JoeGruffins: Is this ok? Just check every 50 seconds. https://github.com/decred/dcrdex/compare/e598ecda67878df8bc302fd6e88ba60803bfed76..f981b0e14dea6c63eee50728992d19cf321694cd

JoeGruffins:
Added a timeout to the periodic header fetches https://github.com/decred/dcrdex/compare/f981b0e14dea6c63eee50728992d19cf321694cd..843c7aeeeb3c860741be39f7d629ad35c08929fa

client/asset/eth/multirpc.go
chappjc: Thanks for removing the double lock, but... why does it need to lock up here?  Isn't `p.ws` immutable?

client/asset/eth/multirpc.go
JoeGruffins: Oh, maybe I was mistaken.

JoeGruffins: It makes sense to wait for it to finish during shutdown?

JoeGruffins:
Just rebased.

client/asset/eth/multirpc.go
JoeGruffins: I think so, moving it down.

client/asset/eth/multirpc.go
JoeGruffins: I tested some more and you are right, closing the client causes the 'ok' part of `case err, ok := <-sub.Err():` being true.

I still think the extra monitoring is sane. Can revert though.

JoeGruffins:
Move the lock down https://github.com/decred/dcrdex/compare/28d85e4080d3cbeba5d8f544aa7d645878682f4a..b738ccd108eb25ff7634ac1a20991f55e62a3aac

chappjc:
Just worked as intended for me right now as blastio fell behind, got flagged as bad, and then caught up really fast:

```
2023-02-17 23:15:17.366 [TRC] CORE[eth][ETH][RPC]: Fetching fresh header from "eth-mainnet.blastapi.io"
2023-02-17 23:15:17.489 [WRN] CORE[eth][ETH][RPC]: Problem getting best header: time since last eth block (222 sec) exceeds 180 sec. Assuming provider eth-mainnet.blastapi.io is not in sync. Ensure your computer's system clock is correct..
2023-02-17 23:15:22.115 [TRC] CORE: New peer count for asset eth: 1
2023-02-17 23:15:22.115 [TRC] CORE: New peer count for asset usdc.eth: 1
2023-02-17 23:15:25.154 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.g.alchemy.com" reported new tip at height 16651611 (0x095e7b51572a9da8dd735dff1dcac429274c736e18713652d082934824598450)
2023-02-17 23:15:26.115 [DBG] CORE[eth][ETH]: tip change: 16651610 (0x209a203ea224de8e366261db80b92b4fb427c084520bf0489204c1fc20ea6d2e) => 16651611 (0x095e7b51572a9da8dd735dff1dcac429274c736e18713652d082934824598450)
2023-02-17 23:15:26.115 [TRC] CORE: Processing tip change for eth
2023-02-17 23:15:26.115 [TRC] CORE: Processing tip change for usdc.eth
2023-02-17 23:15:31.250 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16651593 (0x8e37c44fdca54686b245cad17e114d8807e359fd2944d1eb0967df6caf9b7610)
2023-02-17 23:15:31.539 [TRC] CORE: New peer count for asset btc: 7
2023-02-17 23:15:31.804 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16651594 (0x893fb6ec2d4dbc0e60c0e529076478167c8b7f3ca142817be6067e946169ceee)
2023-02-17 23:15:32.116 [TRC] CORE: New peer count for asset eth: 2
2023-02-17 23:15:32.116 [TRC] CORE: New peer count for asset usdc.eth: 2
2023-02-17 23:15:32.470 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16651595 (0x8eb5d0eb1a0779a8bffabf4bed760f1b691fc4d9456f12c0d53b2c9c598e4dbe)
2023-02-17 23:15:34.784 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16651596 (0xcd59b55834a73e887383e5c8a3b20e4b5c45846e0999f014c79ba02f04065c9d)
2023-02-17 23:15:35.504 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16651597 (0x6ab097173e9207edb7c79c692ad16d4b3680dcb8be0d99e62a9f6817622863c7)
2023-02-17 23:15:36.161 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16651598 (0x3eca4a799bf3563ae8ab44799a67a5c72352a589c9a0b8ca34398df459b2c6ec)
2023-02-17 23:15:36.538 [TRC] CORE: New peer count for asset btc: 8
2023-02-17 23:15:36.762 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16651599 (0xc5340788b2f1f146f075898e5ac37072a572d8b50207046cd822772ef6250e12)
2023-02-17 23:15:37.342 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16651600 (0xd79d53be055fee5221e04556d96b421c28ece5fc1bb7897487bee72f69a0d69f)
2023-02-17 23:15:38.166 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16651601 (0x796459da04c9c84b757df7cfda4fd2b2a76c50c795ca14225254e10b231f37ff)
2023-02-17 23:15:38.820 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16651602 (0x5abb38b8ff4b49c3ce3f0692799167bf485e3faa8122925bde7badd0faaa5676)
2023-02-17 23:15:39.018 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.g.alchemy.com" reported new tip at height 16651612 (0x59a491808f230dc8495ce5b05cd9aa77a54a7452c6272374d8be17d89b2ad8e8)
2023-02-17 23:15:39.116 [DBG] CORE[eth][ETH]: tip change: 16651611 (0x095e7b51572a9da8dd735dff1dcac429274c736e18713652d082934824598450) => 16651612 (0x59a491808f230dc8495ce5b05cd9aa77a54a7452c6272374d8be17d89b2ad8e8)
2023-02-17 23:15:39.116 [TRC] CORE: Processing tip change for usdc.eth
2023-02-17 23:15:39.116 [TRC] CORE: Processing tip change for eth
2023-02-17 23:15:39.485 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16651603 (0xc5859eb5a0df52ae0300b4494b8803229ff6df9112703b8dff945f97a0512b6b)
2023-02-17 23:15:40.112 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16651604 (0xddec1707767e2b271d44a6e0b5744ae902983c93ae2d2a85443d053a200487b3)
2023-02-17 23:15:40.860 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16651605 (0x518b45072927637cd51378308636102adb037e590da736014a8b4679ba4e8cba)
2023-02-17 23:15:41.539 [TRC] CORE: New peer count for asset btc: 9
2023-02-17 23:15:41.634 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16651606 (0x7b6e68242bc3fc531b0ac2769f426e391e80db65c7fdfa0f8efbd79d577eba8a)
2023-02-17 23:15:48.924 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.g.alchemy.com" reported new tip at height 16651613 (0x7dfd3052ffecf1b0413021d7d1c18b67b6300ce74290cc06c630147299fbae88)
2023-02-17 23:15:49.116 [DBG] CORE[eth][ETH]: tip change: 16651612 (0x59a491808f230dc8495ce5b05cd9aa77a54a7452c6272374d8be17d89b2ad8e8) => 16651613 (0x7dfd3052ffecf1b0413021d7d1c18b67b6300ce74290cc06c630147299fbae88)
2023-02-17 23:15:49.116 [TRC] CORE: Processing tip change for usdc.eth
2023-02-17 23:15:49.116 [TRC] CORE: Processing tip change for eth
2023-02-17 23:15:59.252 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16651607 (0x0238ee7ae410401bed6ad7f1cf5d1fd747b75d6f58a570467f7bb19f2d976a6d)
2023-02-17 23:16:00.013 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16651608 (0xfea6a20072a76aa9214d2a99cc74cde5fec3b572facbb3f7fb02d4ac5fc7bab9)
2023-02-17 23:16:01.930 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.g.alchemy.com" reported new tip at height 16651614 (0xf78cc3cfcbe757d8ce5bf34b8767a74aba77b111f505dc0369df3653fbf252e2)
2023-02-17 23:16:02.116 [DBG] CORE[eth][ETH]: tip change: 16651613 (0x7dfd3052ffecf1b0413021d7d1c18b67b6300ce74290cc06c630147299fbae88) => 16651614 (0xf78cc3cfcbe757d8ce5bf34b8767a74aba77b111f505dc0369df3653fbf252e2)
2023-02-17 23:16:02.116 [TRC] CORE: Processing tip change for usdc.eth
2023-02-17 23:16:02.116 [TRC] CORE: Processing tip change for eth
2023-02-17 23:16:03.317 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16651609 (0x9364d6592d7dc4a07984030f130a3c9cea05de314197357234345f79c463c185)
2023-02-17 23:16:04.119 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16651610 (0x209a203ea224de8e366261db80b92b4fb427c084520bf0489204c1fc20ea6d2e)
2023-02-17 23:16:04.691 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16651611 (0x095e7b51572a9da8dd735dff1dcac429274c736e18713652d082934824598450)
2023-02-17 23:16:05.313 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16651612 (0x59a491808f230dc8495ce5b05cd9aa77a54a7452c6272374d8be17d89b2ad8e8)
2023-02-17 23:16:06.210 [TRC] CORE[eth][ETH][RPC]: Using cached header from "eth-mainnet.g.alchemy.com"
2023-02-17 23:16:07.490 [TRC] CORE[eth][ETH][RPC]: Using cached header from "eth-mainnet.blastapi.io"
2023-02-17 23:16:12.934 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.g.alchemy.com" reported new tip at height 16651615 (0x07b2377dc57f72ee77f141839d9a54f67844835de03a7641809ddde8917ccfac)
2023-02-17 23:16:13.116 [DBG] CORE[eth][ETH]: tip change: 16651614 (0xf78cc3cfcbe757d8ce5bf34b8767a74aba77b111f505dc0369df3653fbf252e2) => 16651615 (0x07b2377dc57f72ee77f141839d9a54f67844835de03a7641809ddde8917ccfac
```

But, when and were did it "unfail" the provider?

client/asset/eth/multirpc.go
```diff
@@ -202,6 +215,30 @@ func (p *provider) suggestTipCap(ctx context.Context, log dex.Logger) *big.Int {
...
 }
 
+// refreshHeader fetches a header every headerCheckInterval. This keeps the
+// cached header up to date or fails the provider if there is a problem getting
+// the header.
+func (p *provider) refreshHeader(ctx context.Context, log dex.Logger) {
+	log.Tracef("handling header refreshes for %q", p.host)
+	for {
+		select {
+		case <-time.After(headerCheckInterval):
```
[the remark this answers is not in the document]
JoeGruffins: With a time limit on the context, there should be no way to overlap. Will change to a ticker.

JoeGruffins:
Just rebased.

client/asset/eth/multirpc.go
```diff
@@ -202,6 +215,30 @@ func (p *provider) suggestTipCap(ctx context.Context, log dex.Logger) *big.Int {
...
+			// Fetching the best header will check that either the
+			// provider's cached header is not too old or that a
+			// newly fetched header is not too old. If it is too
+			// old that indicates the provider is not in sync and
+			// should not be used.
+			innerCtx, cancel := context.WithTimeout(ctx, defaultRequestTimeout)
+			if _, err := p.bestHeader(innerCtx, log); err != nil {
+				log.Warnf("Problem getting best header: %s.", err)
```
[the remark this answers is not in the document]
JoeGruffins: Yeah not helpful. Adding the host.

JoeGruffins:
> But, when and were did it "unfail" the provider?

If it is working how I expect, it is not unfailed until a minute passes. Unusuable until a minute after `2023-02-17 23:15:17.489`. The logs are from new headers coming in. So... hmm. If it is caught up it should unfail I guess? There are different reasons for fail currently though, so would need a different stamp for just headers.

JoeGruffins:
Appling the `brickedFailCount` to out of date headers actually may be a bit extreme. Probably should be separate.

JoeGruffins:
Or wait, `setTip` does reset the fail timer.

JoeGruffins:
Adding a log for when a provider was unusable, but is not usable. https://github.com/decred/dcrdex/compare/37ea7c0e8ef4c5de4cdca9be80dbd8843177ad4e..813fc242776c1ab3338fd6f598dcfc3bc5bfbfe8

client/asset/eth/multirpc_live_test.go
```diff
@@ -47,7 +49,8 @@ func mine(ctx context.Context) error {
 }
 
 func testEndpoint(endpoints []string, syncBlocks uint64, tFunc func(context.Context, *multiRPCClient)) error {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	var cancel context.CancelFunc
```
buck54321, line 52: Can delete this declaration.

client/asset/eth/multirpc_live_test.go
```diff
@@ -102,6 +105,33 @@ func testEndpoint(endpoints []string, syncBlocks uint64, tFunc func(context.Cont
...
+	}
+	time.Sleep(time.Second)
+	err := mine(ctx)
+	if err != nil {
+		fmt.Println(err)
+		exit(1)
+	}
+	time.Sleep(time.Second)
+	exit(m.Run())
+}
```
buck54321, line 133: Why need all this?

buck54321 (approved):
Seems to be working.

buck54321:
Hmm. Looks like a ci error.
```
--- FAIL: TestTxWaiters (0.31s)
    swap_test.go:1689: error sending taker swap request: taker swap rpc error. code: 64, msg: already received a swap contract, search in
```
Seems unrelated though.

chappjc:
Retriggered CI.  Can merge with final iteration for nits in https://github.com/decred/dcrdex/pull/2074#pullrequestreview-1304592428

client/asset/eth/multirpc_live_test.go
JoeGruffins: I was mining two blocks before tests when I took out the Simnet check for old headers, but then put the Simnet check back. So.. can revert.

JoeGruffins:
Removed some needless mining in tests since we have a special case for Simnet https://github.com/decred/dcrdex/compare/813fc242776c1ab3338fd6f598dcfc3bc5bfbfe8..324705a2d1192fa5f3d491a8df2274e3bb496244

JoeGruffins:
I thought that listening for ctrl-c was necessary in tests, but it appears not. The tests stop and fail on ctrl-c. Maybe I was thinking about test frameworks like the loadbot.
```
^Csignal: interrupt
FAIL    decred.org/dcrdex/client/asset/eth      3.861s
```

chappjc (approved):

PR #2078, opened 2023-01-27
client/asset: eliminate excessive locking on driver maps

chappjc:
`withDriver` held the mutex for the entire duration of the provided function, AND it was write locking. This blocked other things such as `Assets()` and more importantly, `DecodeCoinID`. Deadlock. The mutex only protects the map, so release it after retrieving the requested driver, and only use a read lock.

### DIFF
--- a/client/asset/driver.go
+++ b/client/asset/driver.go
@@ -52,9 +52,9 @@ type Creator interface {
 }
 
 func withDriver(assetID uint32, f func(Driver) error) error {
-	driversMtx.Lock()
-	defer driversMtx.Unlock()
+	driversMtx.RLock()
 	drv, ok := drivers[assetID]
+	driversMtx.RUnlock()
 	if !ok {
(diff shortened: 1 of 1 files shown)


client/asset/driver.go
```diff
@@ -136,8 +136,8 @@ func OpenWallet(assetID uint32, cfg *WalletConfig, logger dex.Logger, net dex.Ne
 // asset with a corresponding driver registered with this package. For now,
 // token id decoding is deferred to the parent asset.
 func DecodeCoinID(assetID uint32, coinID []byte) (cid string, err error) {
```
chappjc, line 138: DecodeCoinID is used everywhere, probably even from within some of the creator functions.  I experienced a deadlock and I think it was this.  Assets() was waiting on the lock and part of the stack trace was missing.

JoeGruffins (approved):

chappjc:
Have also had this as part of my latest bonds branch.

PR #2079, opened 2023-01-27
client/core: fix newMatchNote swapConfirms data race

chappjc:
The matchTracker.swapConfirms is an atomic, with the confirms() method as it's acccessor. newMatchNote was neglecting this.


JoeGruffins (approved):

JoeGruffins (approved):

JoeGruffins:
Oh, twice approved, sorry about that.

PR #2082, opened 2023-01-29
eth: ctx initialization race

norwnd:
I've started experimenting with using `-race` when building `dexc` binary (when playing around with different stuff) and one race jumped at me right away:

```
WARNING: DATA RACE
Write at 0x00c0001f42d0 by goroutine 303:
  decred.org/dcrdex/client/asset/eth.(*TokenWallet).Connect()
      /Users/norwnd/dcrdex/client/asset/eth/eth.go:827 +0x89
  decred.org/dcrdex/dex.(*ConnectionMaster).Connect()
      /Users/norwnd/dcrdex/dex/runner.go:122 +0xc4
  decred.org/dcrdex/dex.(*ConnectionMaster).ConnectOnce()
      /Users/norwnd/dcrdex/dex/runner.go:153 +0x44
  decred.org/dcrdex/client/core.(*xcWallet).Connect()
      /Users/norwnd/dcrdex/client/core/wallet.go:341 +0xb4
  decred.org/dcrdex/client/core.(*Core).connectWalletResumeTrades()
      /Users/norwnd/dcrdex/client/core/core.go:1859 +0x7d
  decred.org/dcrdex/client/core.(*Core).connectAndUpdateWalletResumeTrades()
      /Users/norwnd/dcrdex/client/core/core.go:1802 +0x215
  decred.org/dcrdex/client/core.(*Core).connectAndUpdateWallet()
      /Users/norwnd/dcrdex/client/core/core.go:1827 +0xf7
  decred.org/dcrdex/client/core.(*Core).connectWallets.func1()
      /Users/norwnd/dcrdex/client/core/core.go:4325 +0xdd
  decred.org/dcrdex/client/core.(*Core).connectWallets.func3()
      /Users/norwnd/dcrdex/client/core/core.go:4361 +0x47

Previous read at 0x00c0001f42d0 by goroutine 259:
  decred.org/dcrdex/client/asset/eth.(*ETHWallet).checkForNewBlocks()
      /Users/norwnd/dcrdex/client/asset/eth/eth.go:3035 +0x9e
  decred.org/dcrdex/client/asset/eth.(*ETHWallet).monitorBlocks()
      /Users/norwnd/dcrdex/client/asset/eth/eth.go:3024 +0xde
  decred.org/dcrdex/client/asset/eth.(*ETHWallet).Connect.func1()
      /Users/norwnd/dcrdex/client/asset/eth/eth.go:813 +0xee
```

Full log (if you need it): [eth-race.txt](https://github.com/decred/dcrdex/files/10530048/eth-race.txt)

Taking a shot at resolving it in this PR. The root cause is that:
- `ETHWallet` and `TokenWallet` are using same `baseWallet.ctx` instance (since they share same instance of `baseWallet`)
- when `TokenWallet.Connect` is called it tries to modify `baseWallet.ctx` (while `ETHWallet` has already been connected a while ago, initialized `baseWallet.ctx`, and is already using it for different purposes)

There doesn't seem to be an easy way to initialize `baseWallet.ctx` once `baseWallet` instance is created (which I think is probably preferrable in general for cases like this one) since a lot of code will likely need a change for that (and [passing it as func argument is probably even better anyway](https://github.com/decred/dcrdex/pull/2082#issuecomment-1407692645)).

So instead, there seems to be no need to re-initialize `baseWallet.ctx` in `TokenWallet` ? becase it doesn't seem to add anything on top of just re-using parent's `ctx` (and `Connect(ctx)` for parent is called before it is called on TokenWallet, from what I see in the code).

It does resolve this race for me, thoughts ?

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -394,7 +394,9 @@ var _ asset.DynamicSwapOrRedemptionFeeChecker = (*TokenWallet)(nil)
 var _ asset.BotWallet = (*assetWallet)(nil)
 
 type baseWallet struct {
-	ctx        context.Context // the asset subsystem starts with Connect(ctx)
+	// The asset subsystem starts with Connect(ctx). This ctx will be initialized
+	// in parent ETHWallet once and re-used in child TokenWallet instances.
+	ctx        context.Context
 	net        dex.Network
 	node       ethFetcher
 	addr       common.Address
@@ -768,7 +770,7 @@ func (w *ETHWallet) Connect(ctx context.Context) (_ *sync.WaitGroup, err error) 
 
 	w.node = cl
 	w.addr = cl.address()
-	w.ctx = ctx
+	w.ctx = ctx // TokenWallet will re-use this ctx.
 	err = w.node.connect(ctx)
 	if err != nil {
 		return nil, err
@@ -824,10 +826,8 @@ func (w *ETHWallet) Connect(ctx context.Context) (_ *sync.WaitGroup, err error) 
 // Connect waits for context cancellation and closes the WaitGroup. Satisfies
 // dex.Connector.
 func (w *TokenWallet) Connect(ctx context.Context) (*sync.WaitGroup, error) {
-	w.ctx = ctx
-
 	if w.parent.ctx == nil || w.parent.ctx.Err() != nil {
-		return nil, fmt.Errorf("parent not connected")
+		return nil, fmt.Errorf("parent wallet not connected")
 	}
 
 	err := w.loadContractors()
@@ -841,7 +841,7 @@ func (w *TokenWallet) Connect(ctx context.Context) (*sync.WaitGroup, error) {
 		defer wg.Done()
 		select {
 		case <-ctx.Done():
-		case <-w.baseWallet.ctx.Done():
+		case <-w.parent.ctx.Done():
 		}
 	}()
 	return &wg, nil


client/asset/eth/eth.go
```diff
@@ -841,7 +841,7 @@ func (w *TokenWallet) Connect(ctx context.Context) (*sync.WaitGroup, error) {
 		defer wg.Done()
 		select {
 		case <-ctx.Done():
-		case <-w.baseWallet.ctx.Done():
+		case <-w.parent.ctx.Done():
```
norwnd, line 844: To be a bit more explicit that `w.baseWallet.ctx` is just parent `ctx`.

chappjc:
I used to run with race more frequently.  Thanks for doing this.

I'm guessing it's much more difficult to eliminate the context *field*?

norwnd:
> I'm guessing it's much more difficult to eliminate the context field?

On the first glance, it isn't a trivial change. Maybe `type Wallet interface` should accept `ctx` in many methods it has (to achieve this, and just seems like a good convention to follow for methods that should be interruptable - similar in spirit to this [note I saw](https://github.com/decred/dcrdex/blob/cacff25f33f17d82c76eb2f23f618d8ad23e2e4c/server/asset/common.go#L38-L40)).

Long term, it's probably desirable to refactor and pass as argument 👍

chappjc:
> > I'm guessing it's much more difficult to eliminate the context field?
> 
> On the first glance, it isn't a trivial change. Maybe `type Wallet interface` should accept `ctx` in many methods it has (to achieve this, and just seems like a good convention to follow for methods that should be interruptable - similar in spirit to this [note I saw](https://github.com/decred/dcrdex/blob/cacff25f33f17d82c76eb2f23f618d8ad23e2e4c/server/asset/common.go#L38-L40)).
> 
> Long term, it's probably desirable to refactor and pass as argument +1

Yup, sadly, I agree.  Sadly because I was initially very resistant to doing this, but I can see clearly the hardships it causes by keeping context fields.
OK, for now let's review what you've got.  Thanks again.

client/asset/eth/eth.go
```diff
@@ -394,7 +394,9 @@ var _ asset.DynamicSwapOrRedemptionFeeChecker = (*TokenWallet)(nil)
 var _ asset.BotWallet = (*assetWallet)(nil)
 
 type baseWallet struct {
-	ctx        context.Context // the asset subsystem starts with Connect(ctx)
+	// The asset subsystem starts with Connect(ctx). This ctx will be initialized
+	// in parent ETHWallet once and re-used in child TokenWallet instances.
+	ctx        context.Context
```
norwnd, line 399: It seems `baseWallet.ctx` is the only field on `baseWallet` suffering from this kind of race.

JoeGruffins (approved):

buck54321 (approved):

PR #2091, opened 2023-01-30
eth: don't cancel ws header subscription context

### LINKED ISSUE #2087: eth: simnet harness websocket blocks not seen
Not seeing `TipChange` from **ws://localhost:38557**.

buck54321:
Resolves #2087

### DIFF
--- a/client/asset/eth/multirpc.go
+++ b/client/asset/eth/multirpc.go
@@ -108,7 +108,11 @@ func (p *provider) setTip(header *types.Header) {
 func (p *provider) cachedTip() *types.Header {
 	stale := time.Second * 10
 	if p.ws {
-		stale = time.Minute * 2
+		// We want to avoid requests, and we expect that our notification feed
+		// is working. Setting this too low would result in unnecessary requests
+		// when notifications are working right. Setting this too high will
+		// inevitably result in long tip change intervals if notifications fail.
+		stale = time.Minute
 	}
 
 	p.tip.RLock()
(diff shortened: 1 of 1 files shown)


client/asset/eth/multirpc.go
```diff
@@ -362,7 +367,7 @@ func connectProviders(ctx context.Context, endpoints []string, log dex.Logger, c
 	// addEnpoint only returns errors that should be propagated immediately.
 	addEndpoint := func(endpoint string) error {
 		// Give ourselves a limited time to resolve a connection.
-		ctx, cancel := context.WithTimeout(ctx, defaultRequestTimeout)
+		timedCtx, cancel := context.WithTimeout(ctx, defaultRequestTimeout)
```
chappjc, line 370: Good catch.  I can see the problem was `go p.subscribeHeaders(ctx, sub, h, log)` below, and causing it to kill that subscription maintenance goroutine way too early.

We might run into trouble into the future if we assume the context to `connectProviders` just limits the time to connect, which is often the case with such functions in the standard library e.g. `DialContext`, but it also kills the goroutine.  I think we should document the context parameter on `(*multiRPCClient).connect` and `connectProviders`, so that callers aren't tempted to put their own timeout on it, or anywhere higher up.

Use via `(*ETHWallet).Connect` => `(*multiRPCClient).connect` seems fine though because our Connector pattern is such that it shuts down the long running process rather than the initial connect attempt.

JoeGruffins (approved):

PR #2093, opened 2023-01-31
client/rpcserver: Always add a new wallet config.

### LINKED ISSUE #2084: Decred SPV wallet creation not working via RPC
From https://github.com/decred/dcrdex/pull/2036#pullrequestreview-1272197515
> ... creating an dcr SPV wallet over rpc doesn't seem to be working for some reason. I'm sure unrelated.

JoeGruffins:
closes #2084

An oversight. There should always be a config. We check for keys later.

### DIFF
--- a/client/rpcserver/types.go
+++ b/client/rpcserver/types.go
@@ -312,6 +312,7 @@ func parseNewWalletArgs(params *RawParams) (*newWalletForm, error) {
 		walletType: params.Args[1],
 		walletPass: params.PWArgs[1],
 		assetID:    uint32(assetID),
+		config:     make(map[string]string),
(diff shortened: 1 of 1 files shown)


chappjc (approved):
Core's `createWallet` does:

```go
	if form.Config == nil {
		form.Config = make(map[string]string)
	}
```

So I assume the issue was the RPC handler, `handleNewWallet`? panic at the following?

```go
form.config[opt.Key] = fmt.Sprintf("%v", opt.DefaultValue)
```

This PR seems fine since `parseNewWalletArgs` is designed to prepare the args for `handleNewWallet`.

PR #2097, opened 2023-02-01
client/btc: fix racing test

norwnd:
Maybe resolves https://github.com/decred/dcrdex/issues/1268.

Looks like a typo that needs fixing regardless of whether or not it resolves 1268 (but it probably does as far as I can tell).

### DIFF
--- a/client/asset/btc/spv_test.go
+++ b/client/asset/btc/spv_test.go
@@ -357,7 +357,7 @@ func (c *tNeutrinoClient) GetBlockHash(blockHeight int64) (*chainhash.Hash, erro
 func (c *tNeutrinoClient) BestBlock() (*headerfs.BlockStamp, error) {
 	c.blockchainMtx.RLock()
 	if c.getBestBlockHashErr != nil {
-		c.blockchainMtx.RUnlock()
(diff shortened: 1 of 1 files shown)


client/asset/btc/spv_test.go
```diff
@@ -357,7 +357,7 @@ func (c *tNeutrinoClient) GetBlockHash(blockHeight int64) (*chainhash.Hash, erro
 func (c *tNeutrinoClient) BestBlock() (*headerfs.BlockStamp, error) {
 	c.blockchainMtx.RLock()
 	if c.getBestBlockHashErr != nil {
-		c.blockchainMtx.RUnlock()
+		defer c.blockchainMtx.RUnlock() // reading c.getBestBlockHashErr below
```
chappjc, line 360: Hit this too.  Had put a fix in a [draft PR for tangential work](https://github.com/decred/dcrdex/pull/1966) and forgotten about it: https://github.com/decred/dcrdex/pull/1966/commits/bf8a4e1fcb465abb3ab5532d8a4d74f80f701c94

We can go with your defer though.

chappjc (approved):

PR #2098, opened 2023-02-02
tests: repo-wide bandaid for rand.Seed badness

chappjc:
Early on we got in the habit of building tests on data with reproducible random sequences using the `math/rand` package's global `*math.Rand` instance, seeding it with some arbitrary hard coded numbers to get certain sequence.  That proved quite problematic for test maintenance, concurrency, and now the standard library rightly has deprecated the use of `rand.Seed` and they have made the default initialization actually random instead of with "1" as the seed.

> The [math/rand](https://go.dev/pkg/math/rand/) package now automatically seeds the global random number generator (used by top-level functions like Float64 and Int) with a random value, and the top-level [Seed](https://go.dev/pkg/math/rand/#Seed) function has been deprecated. Programs that need a reproducible sequence of random numbers should prefer to allocate their own random source, using rand.New(rand.NewSource(seed)).
> 
> Programs that need the earlier consistent global seeding behavior can set GODEBUG=randautoseed=0 in their environment.
> 
> The top-level [Read](https://go.dev/pkg/math/rand/#Read) function has been deprecated. In almost all cases, [crypto/rand.Read](https://go.dev/pkg/crypto/rand/#Read) is more appropriate.

This PR is a bandaid to get the test passing given the new random seeding of the package level `math/rand.Rand`.
We should look at rewriting some of these tests to not have unrecorded but expected magic values to produce certain test data set.

<s>We cannot support Go 1.20 yet, at least not in CI, because golangci-lint is getting patched up for a bug `ERRO [runner] Panic: buildir: package "netip" (isInitialPkg: false, needAnalyzeSource: true): in net/netip.AddrFromSlice: cannot convert Load <[]byte> t0 ([]byte) to [4]byte: goroutine 9489 [running]:`.  There's a PR up with a fix so we might be good to go soon.</s>  Go 1.20 PR in https://github.com/decred/dcrdex/pull/2100

### DIFF
--- a/client/db/bolt/db.go
+++ b/client/db/bolt/db.go
@@ -2123,7 +2123,7 @@ func (db *BoltDB) DeleteInactiveOrders(ctx context.Context, olderThan *time.Time
 			return 0, err
 		}
 		start := time.Now()
-		nDeletedOrders = 0
+		nDeletedBatch := 0
 		err := db.Update(func(tx *bbolt.Tx) error {
 			// Run through the archived order bucket storing id's keys for
 			// returning the order data and deletion later.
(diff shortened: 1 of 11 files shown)


JoeGruffins (commented):
Tests passing on 1.20

except...

Do the tests in `client/db/bolt` pass for you?

chappjc:
> Do the tests in `client/db/bolt` pass for you?

No!  Those were only working with `-short` interestingly.  Fixed bug in https://github.com/decred/dcrdex/pull/2098/commits/d9a84eaf3b7d7ccc2fe6bb1fff5f81d0f25e9c24

Not sure how it ever passed or what changed.

JoeGruffins (approved):
Thats strange...

buck54321 (approved):

PR #2102, opened 2023-02-03
client/eth: Remove non compliant providers.

JoeGruffins:
Check all providers that are not known to be good. Check them on
    reconfigure as well. Also use WebSocket with full nodes rather than the
    light node.

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -596,7 +596,6 @@ func createWallet(createWalletParams *asset.CreateWalletParams, skipConnect bool
 			return errors.New("no providers specified")
 		}
 		endpoints := strings.Split(providerDef, providerDelimiter)
-		n := len(endpoints)
 
 		// TODO: This procedure may actually work for walletTypeGeth too.
 		ks := keystore.NewKeyStore(filepath.Join(walletDir, "keystore"), keystore.LightScryptN, keystore.LightScryptP)
@@ -607,36 +606,9 @@ func createWallet(createWalletParams *asset.CreateWalletParams, skipConnect bool
 		}
 
 		if !skipConnect {
-			ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
-			defer cancel()
-
-			var unknownEndpoints []string
-
-			for _, endpoint := range endpoints {
-				known, compliant := providerIsCompliant(endpoint)
-				if known && !compliant {
-					return fmt.Errorf("provider %q is known to have an insufficient API for DEX", endpoint)
-				} else if !known {
-					unknownEndpoints = append(unknownEndpoints, endpoint)
-				}
-			}
-
-			if len(unknownEndpoints) > 0 && createWalletParams.Net == dex.Mainnet {
-				providers, err := connectProviders(ctx, unknownEndpoints, createWalletParams.Logger, big.NewInt(chainIDs[createWalletParams.Net]), createWalletParams.Net)
-				if err != nil {
-					return err
-				}
-				defer func() {
-					for _, p := range providers {
-						p.ec.Close()
-					}
-				}()
-				if len(providers) != n {
-					return fmt.Errorf("Could not connect to all providers")
-				}
-				if err := checkProvidersCompliance(ctx, walletDir, providers, createWalletParams.Logger); err != nil {
-					return err
-				}
+			if err := createAndCheckProviders(context.Background(), walletDir, endpoints,
+				createWalletParams.Net, createWalletParams.Logger); err != nil {
+				return fmt.Errorf("create and check providers problem: %v", err)
 			}
 		}
 		return importKeyToKeyStore(ks, priv, createWalletParams.Pass)
@@ -893,7 +865,8 @@ func (w *ETHWallet) Reconfigure(ctx context.Context, cfg *asset.WalletConfig, cu
 	}
 
 	if rpc, is := w.node.(*multiRPCClient); is {
-		if err := rpc.reconfigure(ctx, cfg.Settings); err != nil {
+		walletDir := getWalletDir(w.dir, w.net)
+		if err := rpc.reconfigure(ctx, cfg.Settings, walletDir); err != nil {
 			return false, err
 		}
 	}
--- a/client/asset/eth/multirpc.go
+++ b/client/asset/eth/multirpc.go
@@ -84,12 +84,13 @@ type provider struct {
 	// host is the domain and tld of the provider, and is used as a identifier
 	// in logs and as a unique, path- and subdomain-independent ID for e.g. map
 	// keys.
-	host    string
-	ec      *combinedRPCClient
-	ws      bool
-	net     dex.Network
-	tipCapV atomic.Value // *cachedTipCap
-	stop    func()
+	host         string
+	endpointAddr string
+	ec           *combinedRPCClient
+	ws           bool
+	net          dex.Network
+	tipCapV      atomic.Value // *cachedTipCap
+	stop         func()
 
 	// tip tracks the best known header as well as any error encountered
 	tip struct {
@@ -395,8 +396,8 @@ func newMultiRPCClient(dir string, endpoints []string, log dex.Logger, cfg *para
 
 // connectProviders attempts to connect to the list of endpoints, returning a
 // list of providers that were successfully connected. It is not an error for a
-// connection to fail. The caller can infer failed connections from the length
-// and contents of the returned provider list.
+// connection to fail, unless all endpoints fail. The caller can infer failed
+// connections from the length and contents of the returned provider list.
 func connectProviders(ctx context.Context, endpoints []string, log dex.Logger, chainID *big.Int, net dex.Network) ([]*provider, error) {
 	providers := make([]*provider, 0, len(endpoints))
 	var success bool
@@ -409,12 +410,12 @@ func connectProviders(ctx context.Context, endpoints []string, log dex.Logger, c
 		}
 	}()
 
-	// addEnpoint only returns errors that should be propagated immediately.
+	// addEndpoint only returns errors that should be propagated immediately.
 	addEndpoint := func(endpoint string) error {
 		// Give ourselves a limited time to resolve a connection.
 		timedCtx, cancel := context.WithTimeout(ctx, defaultRequestTimeout)
 		defer cancel()
-		// First try to get a websocket connection. Websockets have a header
+		// First try to get a websocket connection. WebSockets have a header
 		// feed, so are much preferred to http connections. So much so, that
 		// we'll do some path inspection here and make an attempt to find a
 		// websocket server, even if the user requested http.
@@ -510,9 +511,10 @@ func connectProviders(ctx context.Context, endpoints []string, log dex.Logger, c
 		}
 
 		p := &provider{
-			host: host,
-			ws:   sub != nil,
-			net:  net,
+			host:         host,
+			endpointAddr: endpoint,
+			ws:           sub != nil,
+			net:          net,
 			ec: &combinedRPCClient{
 				Client: ec,
 				rpc:    rpcClient,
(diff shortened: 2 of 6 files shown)


dex/testing/eth/harness.sh
```diff
@@ -20,6 +20,8 @@ ALPHA_NODE_KEY="71d810d39333296b518c846a3e49eca55f998fd7994998bb3e5048567f2f073c
 ALPHA_ENODE="897c84f6e4f18195413c1d02927e6a4093f5e7574b52bdec6f20844c4f1f6dd3f16036a9e600bd8681ab50fd8dd144df4a6ba9dd8722bb578a86aaa8222c964f"
 ALPHA_NODE_PORT="30304"
 ALPHA_AUTHRPC_PORT="8552"
+ALPHA_HTTP_PORT="38556"
+ALPHA_WS_PORT="38557"
```
JoeGruffins, line 24: Would like to move this to the full nodes, since that is what would be used irl.

dex/testing/eth/create-node.sh
```diff
@@ -135,8 +132,10 @@ if [ "${SYNC_MODE}" = "snap" ]; then
   tmux send-keys -t "$TMUX_WIN_ID" "${NODES_ROOT}/harness-ctl/${NAME} --nodiscover " \
 	  "--config ${NODE_DIR}/eth.conf --unlock ${CHAIN_ADDRESS} " \
 	  "--password ${GROUP_DIR}/password --light.serve 25 --datadir.ancient " \
-	  "${NODE_DIR}/geth-ancient --verbosity 5 --vmdebug 2>&1 | tee " \
-	  "${NODE_DIR}/${NAME}.log" C-m
+	  "${NODE_DIR}/geth-ancient --verbosity 5 --vmdebug --http --http.port " \
+	  "${HTTP_PORT} --ws --ws.port ${WS_PORT} --ws.api " \
+	  "eth --allow-insecure-unlock " \
```
JoeGruffins, line 137: Removed all the modules except eth. This should be all we need. Also "db" was apparently not a module? This was for the light client harness node.
```
ERROR[02-03|11:23:46.895] Unavailable modules in HTTP API list     unavailable=[db] available="[admin debug web3 eth txpool personal clique net les vflux engine]"
```

JoeGruffins:
Hmm, actually seems bugged. Doesn't seem to be checking unknown providers either. Will figure it out here.

dex/testing/eth/create-node.sh
chappjc: Noticed this too: https://github.com/decred/dcrdex/pull/2013#discussion_r1088346355
No clue why it was ever there.  Good to have it here too.

However, didn't we still want txpool? Is the idea that we want it to use the dumbed down fallback path?  Or that beta is just used for server?

JoeGruffins: > However, didn't we still want txpool? Is the idea that we want it to use the dumbed down fallback path? Or that beta is just used for server?

I guess that server operators will use providers, either by choice or because the eth node can not be kept up %100. I also guess that the txpool type operations will not be available for providers. Maybe we can turn txpool on for one and off for the other? So we can use both for testing.

chappjc: This is the simnet harness, so I mean if you're testing dexc and want to use the beta node?  Would that just be pointless and now expected not to work?  I'm fine with that, if that's intended.

JoeGruffins: > This is the simnet harness, so I mean if you're testing dexc (client, not server) and want to use the beta node? Would that just be pointless and now expected not to work?

I didn't think the client needed anything but "eth" and only server needed "txpool" (doesn't need now)

I think client is fine with with either "eth" or "eth,txpool"?

chappjc:
Oh.  Sorry about that.  You're right

client/asset/eth/eth.go
```diff
@@ -849,7 +823,8 @@ func (w *ETHWallet) Reconfigure(ctx context.Context, cfg *asset.WalletConfig, cu
 	}
 
 	if rpc, is := w.node.(*multiRPCClient); is {
-		if err := rpc.reconfigure(ctx, cfg.Settings); err != nil {
+		walletDir := getWalletDir(w.dir, w.net)
```
JoeGruffins, line 826: The `getWalletDir` path seems unnecessary. It will end up being something like `$HOME/.dexc/simnet/assetdb/eth/simnet` and I don't think the last `siment` is necessary. The `keystore` folder already lives there though, is it worth changing now?

dex/testing/eth/create-node.sh
JoeGruffins: Now alpha is `eth` and beta is `eth,txpool`

client/asset/eth/multirpc.go
```diff
@@ -569,12 +569,82 @@ func (m *multiRPCClient) voidUnusedNonce() {
...
+		localCP[d] = true
+		if _, known := compliantProviders[d]; known {
+			continue
+		}
+		unknownEndpoints = append(unknownEndpoints, p)
+	}
+
+	providers := []*provider{}
+	if len(unknownEndpoints) > 0 {
+		ctx, cancel := context.WithTimeout(ctx, time.Second*10)
```
JoeGruffins, line 607: I wonder if 10 seconds is long enough if there are multiple providers to check.

client/asset/eth/eth.go
chappjc: I've noticed that too.  Not sure if we should change it though.  However, nobody should be using this on mainnet, and we can move our testnet folders around if we do change it.
Maybe someone else has a more decisive opinion.

client/asset/eth/multirpc.go
chappjc: Multiply timeout by number of providers?

client/asset/eth/multirpc.go
```diff
@@ -569,12 +569,82 @@ func (m *multiRPCClient) voidUnusedNonce() {
...
+		}
+		writeLocalCP = true
+		localCP[d] = true
+		if _, known := compliantProviders[d]; known {
+			continue
+		}
+		unknownEndpoints = append(unknownEndpoints, p)
+	}
+
+	providers := []*provider{}
```
chappjc, line 605: I think this slice should just be declared by the connectProviders call.

client/asset/eth/eth.go
buck54321: This has always bugged me too. On one hand, the asset package shouldn't make assumptions about what the caller is going to give. On the other hand, as you pointed out, it's stupid with the extra directory.

Anyway, won't change this bork existing wallets?

client/asset/eth/multirpc.go
```diff
@@ -569,12 +569,82 @@ func (m *multiRPCClient) voidUnusedNonce() {
...
+		d := domain(p)
+		if localCP[d] {
+			continue
+		}
+		writeLocalCP = true
+		localCP[d] = true
+		if _, known := compliantProviders[d]; known {
+			continue
+		}
+		unknownEndpoints = append(unknownEndpoints, p)
```
buck54321, line 602:
```Go
if _, known := compliantProviders[d]; !known {
	unknownEndpoints = append(unknownEndpoints, p)
}
```

client/asset/eth/multirpc.go
buck54321: There is already a per-connection default timeout in `connectProviders`. I think what needs to happen is to change the call signature of `rpcTest.f` to `func(context.Context, *provider) error`, and create a new timed context for each request in `checkProvidersCompliance`.

JoeGruffins:
Just rebased.

client/asset/eth/multirpc.go
JoeGruffins: Changed the `rpcTest.f` function signature as suggested.

client/asset/eth/multirpc.go
JoeGruffins: Oh yeah, I was thinking about returning this and not disconnecting sometimes, but it was messy so I didn't do that. Removing this unneeded declaration.

client/asset/eth/multirpc.go
JoeGruffins: Updated.

client/asset/eth/eth.go
JoeGruffins: > Anyway, won't this change bork existing wallets?

Yes. Existing testnet eth multiprc wallets.

JoeGruffins: Just leaving for now, can be addressed in another pr if necessary.

JoeGruffins:
Use per funciton contexts https://github.com/decred/dcrdex/compare/fb3e0e3bf030c7a704417dfd75b3a66aa7f911c8..1fd0935a1ab32e48bb4bfad9a7d5da3123d2526c

client/asset/eth/eth.go
```diff
@@ -580,36 +579,11 @@ func createWallet(createWalletParams *asset.CreateWalletParams, skipConnect bool
...
-				}()
-				if len(providers) != n {
-					return fmt.Errorf("Could not connect to all providers")
-				}
-				if err := checkProvidersCompliance(ctx, walletDir, providers, createWalletParams.Logger); err != nil {
-					return err
-				}
+			if err := createAndCheckProviders(walletDir, endpoints,
+				createWalletParams.Net, createWalletParams.Logger, ctx); err != nil {
+				return fmt.Errorf("providers problem: %v", err)
```
norwnd, line 586: Maybe to be more specific: **providers problem** -> **create and check providers problem**

client/asset/eth/multirpc.go
```diff
@@ -569,12 +569,80 @@ func (m *multiRPCClient) voidUnusedNonce() {
...
+func createAndCheckProviders(walletDir string, endpoints []string, net dex.Network,
+	log dex.Logger, ctx context.Context) error {
+	var localCP map[string]bool
+	path := filepath.Join(walletDir, "compliant-providers.json")
+	b, err := os.ReadFile(path)
+	if err == nil {
+		if err := json.Unmarshal(b, &localCP); err != nil {
+			log.Warnf("Couldn't parse compliant providers file: %v", err)
+		}
+	}
```
norwnd, line 584: Looks like we are dropping the original error from `os.ReadFile`, maybe worth logging also (similar to how parsing error from `json.Unmarshal` is logged here) ?

client/asset/eth/multirpc.go
```diff
@@ -569,12 +569,80 @@ func (m *multiRPCClient) voidUnusedNonce() {
...
+	if writeLocalCP {
+		// All unknown providers were checked.
+		b, err := json.Marshal(localCP)
+		if err != nil {
+			return err
+		}
+		if err := os.WriteFile(path, b, 0644); err != nil {
+			log.Errorf("Failed to write compliant providers file: %v", err)
+		}
```
norwnd, line 632: Maybe return `error` here instead of just logging it ? So the user can retry saving his configuration (in case it's temporary failure), because he'll have to refill it upon restart ... I can't think of a valid use case to log here instead of erroring right away, thougths ?

norwnd (commented):
Minor suggestions, just possible improvements.

client/asset/eth/multirpc.go
```diff
@@ -569,12 +569,80 @@ func (m *multiRPCClient) voidUnusedNonce() {
 	m.lastNonce.stamp = time.Time{}
 }
 
-func (m *multiRPCClient) reconfigure(ctx context.Context, settings map[string]string) error {
+// createAndCheckProviders creates and connects to providers. It checks that
+// unknown providers have a sufficient api to trade and saves good providers to
+// file. One bad provider or connect problem will cause this to error.
+func createAndCheckProviders(walletDir string, endpoints []string, net dex.Network,
+	log dex.Logger, ctx context.Context) error {
```
buck54321, line 576: It bugs me that the `context.Context` is not the first argument.

chappjc: bugs my linters too

client/asset/eth/eth.go
```diff
@@ -580,36 +579,11 @@ func createWallet(createWalletParams *asset.CreateWalletParams, skipConnect bool
 		}
 
 		if !skipConnect {
-			ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+			ctx, cancel := context.WithCancel(context.Background())
```
chappjc, line 582: What's the purpose of this cancelable context?

client/asset/eth/multirpc.go
chappjc: If it's something than os.ErrNotExists or whaterver

client/asset/eth/multirpc.go
JoeGruffins: Ah, sorry, will move.

JoeGruffins:
Just rebased.

client/asset/eth/eth.go
JoeGruffins: Should I just use `context.Background()`?

client/asset/eth/multirpc.go
```diff
@@ -569,12 +569,82 @@ func (m *multiRPCClient) voidUnusedNonce() {
...
+		if err != nil {
+			return err
+		}
+		defer func() {
+			for _, p := range providers {
+				p.ec.Close()
+			}
+		}()
+		if len(providers) != len(unknownEndpoints) {
+			return errors.New("Could not connect to all unknown providers.")
```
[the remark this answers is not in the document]
JoeGruffins: There are logs in the `connectProviders` function, but yeah, they don't currently make it to the UI.

client/asset/eth/eth.go
JoeGruffins: Ok.

client/asset/eth/multirpc.go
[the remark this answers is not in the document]
norwnd: > Is it easiest to stick the failed providers in that error?

As a user, by looking at that error in UI I'd want to know which providers I should remove/adjust. Might be easier to "fail fast" and show only one bad  provider at a time, so user can repeat adjusting until all supplied configuration is acceptable  good.

client/asset/eth/eth.go
JoeGruffins: Using `context.Background()`

client/asset/eth/multirpc.go
JoeGruffins: Hope I haven't done this anywhere else.

client/asset/eth/multirpc.go
JoeGruffins: Now printing the error with warn severity unless os.ErrNotExists

client/asset/eth/multirpc.go
JoeGruffins: He won't need to refill it, the providers are saved somewhere else. This file is only for a list of user added providers we didn't know about, like random full nodes. If it fails writing, it would mean the provider needs to be checked again if unknown if doing reconfigure. So, it only saves a little time and not every start up. I can't imagine why this would error here under normal circumstances.

client/asset/eth/multirpc.go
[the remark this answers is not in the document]
JoeGruffins: Sticking the failing provider addresses in the error.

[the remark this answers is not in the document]
JoeGruffins: image

client/asset/eth/multirpc.go
norwnd: I misunderstood it then, thanks for clarifying.

JoeGruffins:
Put all the failed providers in the error so it makes it to the frontend https://github.com/decred/dcrdex/compare/6ff317b8054100e06e059373e230680ab0979217..eae7c4f018735fd1214eefd814a70655bfc27824

norwnd (approved):

JoeGruffins:
Was a hairy rebase.

client/asset/eth/multirpc.go
JoeGruffins: Forgot to remove this ctx it seems.

JoeGruffins:
Removed unneeded timed context. Both `connectProviders` and  `checkProvidersCompliance` will use their own timed contexts  https://github.com/decred/dcrdex/compare/15c6d1d974d643cb55d2652000417febcdbde14d..5bd1190a3b1b9471ad0474d3bb6895e4627a3acd

chappjc:
> Removed unneeded timed context. Both `connectProviders` and `checkProvidersCompliance` will use their own timed contexts https://github.com/decred/dcrdex/compare/15c6d1d974d643cb55d2652000417febcdbde14d..5bd1190a3b1b9471ad0474d3bb6895e4627a3acd

But you're calling with `context.Background()` in `createWallet`: https://github.com/decred/dcrdex/blob/5bd1190a3b1b9471ad0474d3bb6895e4627a3acd/client/asset/eth/eth.go#L609
Probably OK, but it's not timed on that path.

chappjc:
`git rm client/asset/eth/cmd/getgas` pls

chappjc (approved):

chappjc:
Merged, but can you clarify your comment about light node for me?

JoeGruffins:
> git rm client/asset/eth/cmd/getgas/getgas pls

Ah srry, this is like twice now. I guess you removed it.

PR #2104, opened 2023-02-05
add multi-rpc client for server

buck54321:


dex/testing/dcrdex/harness.sh
```diff
@@ -217,14 +217,19 @@ EOF
 fi
 
 if [ $ETH_ON -eq 0 ]; then
+
+ETH_CONFIG_PATH=${HARNESS_DIR}/eth.conf
+cat << EOF >> $ETH_CONFIG_PATH
+ws://localhost:38557
+EOF
```
chappjc, line 224: Since the markets.json in dcrdex/harness.sh tends to be demonstrative, would you mind adding a `#`-prefixed comment with a second provider as an example?  Also, could we actually have the harness use two nodes so it's more than an example?

server/asset/eth/eth.go
```diff
@@ -276,15 +278,34 @@ func NewBackend(endpoint string, logger dex.Logger, net dex.Network) (*ETHBacken
...
+		return nil, err
+	}
+	defer file.Close()
+
+	var endpoints []string
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := strings.Trim(scanner.Text(), " ")
+		if line == "" || strings.HasPrefix(line, "#") {
+			continue
```
chappjc, line 292: Very nice to avoid ini/conf dependency since this is trivial. ty

server/asset/eth/rpcclient.go
```diff
@@ -33,61 +34,67 @@ type ContextCaller interface {
...
+		if err == nil {
+			return nil
 		}
-	} else {
-		log.Debugf("dialing endpoint: %s", c.endpoint)
-		client, err = rpc.DialContext(ctx, c.endpoint)
-		if err != nil {
-			return fmt.Errorf("unable to dial rpc: %v", err)
+		if len(propagateNotFound) > 0 && propagateNotFound[0] && strings.Contains(err.Error(), "not found") {
+			return err
```
chappjc, line 80: Digesting this variadic bool, it's more like `haltOnNotFound`, I think, like it short-circuits if it hits a not found from provider A so it doesn't bother with provider B because it's not really an error, just an expected outcome.

server/asset/eth/rpcclient.go
```diff
@@ -33,61 +34,67 @@ type ContextCaller interface {
...
 		}
+		c.log.Errorf("Unpropagated error from %q: %v", c.endpoints[idx], err)
+		// Try the next client.
+		c.idxMtx.Lock()
+		// Only advance it if another thread hasn't.
+		if c.endpointIdx == idx && len(c.endpoints) > 0 {
+			c.endpointIdx = (c.endpointIdx + 1) % len(c.endpoints)
```
chappjc, line 87: Clever condition and circular rotation.

server/asset/eth/rpcclient.go
```diff
@@ -97,74 +104,122 @@ func (c *rpcclient) connect(ctx context.Context, log dex.Logger) error {
 		return fmt.Errorf("no contract address for eth version %d on %s", ethContractVersion, c.net)
 	}
 
-	es, err := swapv0.NewETHSwap(contractAddr, c.ec)
-	if err != nil {
-		return fmt.Errorf("unable to find swap contract: %v", err)
+	c.clients = make([]*ethConn, len(c.endpoints))
+	for i, endpoint := range c.endpoints {
+		client, err := rpc.Dial(endpoint)
```
chappjc, line 109: Would `DialContext` not be preferable, so caller can cancel this instead of hang?

server/asset/eth/rpcclient.go
```diff
@@ -175,27 +230,28 @@ func (c *rpcclient) swap(ctx context.Context, assetID uint32, secretHash [32]byt
 // transaction gets the transaction that hashes to hash from the chain or
 // mempool. Errors if tx does not exist.
 func (c *rpcclient) transaction(ctx context.Context, hash common.Hash) (tx *types.Transaction, isMempool bool, err error) {
-	return c.ec.TransactionByHash(ctx, hash)
+	return tx, isMempool, c.withClient(func(ec *ethConn) error {
+		tx, isMempool, err = ec.TransactionByHash(ctx, hash)
+		return err
+	}, true)
```
chappjc, line 236: Maybe a trailing comment because this is easy to miss: `}, true) // stop on first provider with "not found", because this shoudl be an error if tx does not exist`

server/asset/eth/rpcclient.go
```diff
@@ -33,61 +34,67 @@ type ContextCaller interface {
...
+		err = f(ec)
+		if err == nil {
+			return nil
 		}
-	} else {
-		log.Debugf("dialing endpoint: %s", c.endpoint)
-		client, err = rpc.DialContext(ctx, c.endpoint)
-		if err != nil {
-			return fmt.Errorf("unable to dial rpc: %v", err)
+		if len(propagateNotFound) > 0 && propagateNotFound[0] && strings.Contains(err.Error(), "not found") {
```
chappjc, line 79: Looks like `ethclient.(*Client)TransactionByHash` (and `TransactionReceipt` and others) will return a `github.com/ethereum/go-ethereum.NotFound` (https://github.com/ethereum/go-ethereum/blob/877d2174fb898daefeb7e9c944a68d1c3a98c923/interfaces.go#L29-L30)

`errors.Is` has annoying pitfalls, but then so does string matching (they might change that string).  Think we can check the error type first and then the string too?

chappjc (commented):
utACK, a few minor things.

I'm going to rebase https://github.com/decred/dcrdex/pull/2013 on this PR and test today.

buck54321:
I addressed your first review, @chappjc. One thing I'm still thinking about doing is advancing the endpoint index if a header appears outdated.

chappjc:
BTW, how's https://github.com/decred/dcrdex/pull/2074 for you, for client
ofc

server/asset/eth/rpcclient.go
```diff
@@ -97,74 +105,122 @@ func (c *rpcclient) connect(ctx context.Context, log dex.Logger) error {
...
-	}
-	return c.ec.HeaderByNumber(ctx, big.NewInt(int64(bn)))
+func (c *rpcclient) bestHeader(ctx context.Context) (hdr *types.Header, err error) {
+	return hdr, c.withClient(func(ec *ethConn) error {
+		bn, err := ec.BlockNumber(ctx)
+		if err != nil {
+			return err
+		}
+		hdr, err = ec.HeaderByNumber(ctx, big.NewInt(int64(bn)))
+		return err
```
chappjc, line 179: Doing some testing with Blast (blastapi.io) and I got a method not found, but I can't tell what method:

```
[ERR] ASSET[eth][RPC]: Unpropagated error from "wss://eth-mainnet.blastapi.io/xxxxx-xyyyy-adsfasdf": Method not found
[INF] ASSET[eth][RPC]: Switching RPC endpoint to "ws://127.0.0.1:8546"
```

On startup it did the balance method check and as expected it went the dumb way:

```
[DBG] ASSET[eth]: Parsed 2 endpoints from the ETH config file                                                                                                                      
[WRN] ASSET[eth][RPC]: Error checking required modules at "wss://eth-mainnet.blastapi.io/xxxxx-xyyyy-adsfasdf": unable to check supported modules: Method not found
[WRN] ASSET[eth][RPC]: Will not account for pending transactions in balance calculations at "wss://eth-mainnet.blastapi.io/xxxxx-xyyyy-adsfasdf"                   
[DBG] ASSET[eth][RPC]: API endpoints supported by ws://127.0.0.1:8546: admin:1.0 engine:1.0 eth:1.0 net:1.0 rpc:1.0 txpool:1.0          
```

So possibly there's another method missing and blast isn't gonna work, or the smart/dumb switch didn't work as expected.  I don't think it was checking a balance at the time, so it's probably another method sadly.

chappjc:
> I addressed your first review, @chappjc. One thing I'm still thinking about doing is advancing the endpoint index if a header appears outdated.

That seems like a good idea.  Also, on startup when it checks each endpoint it could compare the best header times (even hash and height) to be sure one doesn't look dead or stalled.

server/asset/eth/rpcclient.go
buck54321: You could add blastapi to your `~/dextest/providers.json` and run `go test -tags rpclive,lgpl -run TestMainnetCompliance` in **client/asset/eth** to see if anything pops out. I've seen `"Method not found"` for `SuggestGasTipCap` before. Not sure how that aligns with API availability.

server/asset/eth/rpcclient.go
```diff
@@ -171,11 +197,18 @@ func (c *rpcclient) withTokener(assetID uint32, f func(*tokener) error) error {
 // bestHeader gets the best header at the time of calling.
 func (c *rpcclient) bestHeader(ctx context.Context) (hdr *types.Header, err error) {
```
chappjc, line 198: To clarify discussion on matrix happening now, I think `poll` will work best without calling `bestHeader`:

```diff
diff --git a/server/asset/eth/eth.go b/server/asset/eth/eth.go
index 38dbfe683..34fa87db3 100644
--- a/server/asset/eth/eth.go
+++ b/server/asset/eth/eth.go
@@ -708,13 +708,19 @@ func (eth *ETHBackend) poll(ctx context.Context) {
                        be.sendBlockUpdate(u)
                }
        }
-       bhdr, err := eth.node.bestHeader(ctx)
+       blockNum, err := eth.node.blockNumber(ctx)
        if err != nil {
-               send(false, fmt.Errorf("error getting best block header from geth: %w", err))
+               send(false, fmt.Errorf("error getting best block number: %w", err))
+               return
+       }
+       if blockNum != best.height {
+               // Same height, nothing to do.
                return
        }
-       if bhdr.Hash() == best.hash {
-               // Same hash, nothing to do.
+
+       bhdr, err := eth.node.bestHeader(ctx)
+       if err != nil {
+               send(false, fmt.Errorf("error getting best block header from geth: %w", err))
                return
        }
        update := func(reorg bool, fastBlocks bool) {
```

That should drastically reduce RPC usage because `eth_blockNumber` is very cheap, but `eth_getBlockByNumber` is relatively costly in terms of bandwidth.

chappjc: Oh, whoops I meant `==`, not `!=`

server/asset/eth/rpcclient.go
chappjc: This turned out to be the `eth_syncing` method, which is expected (with https://github.com/decred/dcrdex/pull/2013/commits/06345a5ffcdcedfc1218a35ca8e58289892071bf).  Given this provider switching that happens as a result, I think I'll update that commit in #2013 to just not even try `eth_syncing` since it's next to useless anyway.

chappjc: #2013 revised with https://github.com/decred/dcrdex/pull/2013/commits/168cb574f8b128ea12b426b753e4261889ec5849
No more `eth_syncing` on client either.

PR #2105, opened 2023-02-06
server/asset/eth: forget reorg handling

buck54321:
re: https://github.com/decred/dcrdex/pull/2104#discussion_r1096819950

We don't use reorg. No need for complex tip change handling.

### DIFF
--- a/server/asset/eth/eth.go
+++ b/server/asset/eth/eth.go
@@ -172,11 +172,6 @@ type ethFetcher interface {
 	accountBalance(ctx context.Context, assetID uint32, addr common.Address) (*big.Int, error)
 }
 
-type hashN struct {
-	height uint64
-	hash   common.Hash
-}
-
 type baseBackend struct {
 	// A connection-scoped Context is used to cancel active RPCs on
 	// connection shutdown.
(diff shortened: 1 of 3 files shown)


chappjc (approved):
Great.  Less RPCs.  An hour of `eth_getBlockByNumber` at 20 req/min seemed to be about ~7MiB, so this is good to have.

JoeGruffins (approved):
According to https://www.paradigm.xyz/2021/07/ethereum-reorgs-after-the-merge#post-merge-ethereum-with-proof-of-stake

```
Finally, long reorgs are not possible because all blocks that are deeper than 2 epochs in the past are considered "finalized", i.e. it is impossible to revert past them. If an attacker caused two conflicting blocks to be finalized (e.g by controlling 67% of the stake), the system would need to fall back to social intervention to recover.
```

So, I think, if confirms are 3+ we are fine.

chappjc:
Good find @JoeGruffins.  I wonder what "social intervention to recover" is supposed to mean, panic mode and politicking I guess.

server/asset/eth/eth.go
```diff
@@ -687,57 +678,18 @@ func (eth *ETHBackend) poll(ctx context.Context) {
...
-		send(false, nil)
+	if bn == eth.bestHeight {
+		// Same height, nothing to do.
 		return
 	}
-	// The block for our recorded hash was forked off and the chain had a
-	// reorganization.
-	update(true, false)
-	send(true, nil)
+	eth.log.Debugf("Tip change from %d to %d.", eth.bestHeight, bn)
```
chappjc, line 690: Final note that tip change log no longer includes the block hashes, and I'm fine with that.
If we really need the reorg bool and the best block hashes back, we can keep polling using `eth_blockNumber`, and put back the `eth.node.headerByHeight` call if the height has changed.  At least the frequency will be more like 12 seconds instead of 1 sec.

PR #2106, opened 2023-02-06
server/assets/eth: refactor options for rpclient bestHeader

buck54321:
**There are 2 options here**

The `(*rpcclient).bestHeader` method was not optimized to reduce rpc calls. I've given two options to make it better.

1) We don't need provide a block number to `HeaderByNumber`. The arg is a `*big.Int` and if we pass `nil`, it is parsed to `"latest"` which has a special meaning for **geth**. **This is the first commit.**
2) The load of getting a header is large. We could still call `BlockNumber`, but only call `HeaderByNumber` if the tip height has changed. **This is the second commit.**

### DIFF
--- a/server/asset/eth/rpcclient.go
+++ b/server/asset/eth/rpcclient.go
@@ -133,11 +133,7 @@ func (c *rpcclient) withTokener(assetID uint32, f func(*tokener) error) error {
 
 // bestHeader gets the best header at the time of calling.
 func (c *rpcclient) bestHeader(ctx context.Context) (*types.Header, error) {
-	bn, err := c.ec.BlockNumber(ctx)
-	if err != nil {
-		return nil, err
-	}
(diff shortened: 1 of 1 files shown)


chappjc:
With https://github.com/decred/dcrdex/pull/2105, I think we're good without the cache in the second commit.

I think your use of the `nil` arg in https://github.com/decred/dcrdex/pull/2104 is good.

Seems like we already have all the bases covered in those.

chappjc:
Still, I don't see a problem with the cache commit.  It could make the difference when RPCs really need to be light :shrug:

buck54321:
`bestHeader` is called once per market on startup via `(Swapper).ChainsSynced -> (*baseBackend).Synced`. It's also still called in `FeeRate`, but it's not clear to my why we even have that method on **eth**, since it's not really used by the `feeFetcher`. I guess just cause it's part of the `Backend` interface.

Anyway, no cache needed, I think. Closing.

PR #2108, opened 2023-02-06
TestTaper is flaky

LasTshaMAN:
Closes https://github.com/decred/dcrdex/issues/1671.

I probably should have addressed this one when fixing adjacent test (in https://github.com/decred/dcrdex/pull/1742), but decided to leave `TestTaper` as is since it didn't complain, and unlike `TestTaperingQueue` did check that `Expiration` functionality works properly. 

It seems intervals used in `TestTaper` (on the order of 10ms) are low enough to sometimes result in this false-positive failure, especially if run with `-race` which is what I see in `run_tests.sh`.

I can suggest 2 ways to address this flakiness:
- change `TestTaper` so that it will test `Expiration` functionality only, which is what I'm doing in this PR
- or bumping 10ms up to maybe ~30ms ... this would be just as shaky imo

### DIFF
--- a/dex/wait/queue_test.go
+++ b/dex/wait/queue_test.go
@@ -8,29 +8,21 @@ import (
 	"time"
 )
 
-func TestTaper(t *testing.T) {
+func TestTaperingQueueExpiration(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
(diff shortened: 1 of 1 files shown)


chappjc:
> * change `TestTaper` so that it will test `Expiration` functionality only, which is what I'm doing in this PR

Good.  No need to repeat that.

chappjc (approved):

PR #2110, opened 2023-02-06
app: Fee summary touch-ups

### LINKED ISSUE #2095: ui: swap fee summary followup
#2057 solves some big issues with fee summaries for tokens. There were a couple of more things I think could use some tweaking.

1) We changed the fee range from percent to absolute values
image
which isn't as informative, imo. Can we still display percents for non-tokens? We could even go back to percents for tokens, **IF** we have a fiat rate source for the necessary assets.

2) When a fee rate rounds to 0.0 (< 0.05) %, we should show "< 0.1%"
image

martonp:
The fee summary panel now shows the fees as a percentage of the order for token assets, if the fiat rate is available for both the token and the parent. Also, if the percentage is less than 0.05%, "< 0.1" is displayed.

Closes #2095

### DIFF
--- a/client/webserver/site/src/js/markets.ts
+++ b/client/webserver/site/src/js/markets.ts
@@ -1768,8 +1768,7 @@ export default class MarketsPage extends BasePage {
   /* setFeeEstimates sets all of the pre-order estimate fields */
   setFeeEstimates (swap: PreSwap, redeem: PreRedeem, order: TradeForm) {
     const { page, market } = this
-    let { baseUnitInfo, quoteUnitInfo, rateConversionFactor } = market
-    const swapped = swap.estimate.value || 0
+    const { baseUnitInfo, quoteUnitInfo, rateConversionFactor } = market
     const fmtPct = (value: number) => {
       if (value < 0.05) return '< 0.1'
       return percentFormatter.format(value)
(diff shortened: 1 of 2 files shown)


JoeGruffins:
The `Fee Projection` and the breakdown in the details are very different. Should the fee projection be the lowest and highest on the details page? On testnet:
image
But details are twice the values. These do look correct though:
image

martonp:
> The `Fee Projection` and the breakdown in the details are very different. Should the fee projection be the lowest and highest on the details page? On testnet:

Previously we were adding the fee percentages, but now I started averaging them. I was thinking that 3% of the swap + 3% of the redemption is 3% of the entire order, but now I realize that the entire order amount is just one side of the order. I'll fix.

client/webserver/site/src/js/markets.ts
```diff
@@ -1770,51 +1770,98 @@ export default class MarketsPage extends BasePage {
...
+      const tokenFiatRate = app().fiatRatesMap[market.base.id]
+      const parentFiatRate = app().fiatRatesMap[parent.id]
+      if (tokenFiatRate && parentFiatRate) {
+        baseExchangeRate = tokenFiatRate / parentFiatRate
+      } else {
+        baseExchangeRate = 0
+      }
     }
+
+    console.log('baseExchangeRate', baseExchangeRate)
```
buck54321, line 1799: Can delete? Here and three other places.

client/webserver/site/src/js/markets.ts
```diff
@@ -1770,51 +1770,98 @@ export default class MarketsPage extends BasePage {
...
+      [fromExchangeRate, toExchangeRate] = [toExchangeRate, fromExchangeRate]
     }
 
+    console.log('fromExchangeRate', fromExchangeRate)
+    console.log('toExchangeRate', toExchangeRate)
+
     // Set swap fee estimates in the details pane.
-    const bestSwapPct = swap.estimate.realisticBestCase / swapped * 100
-    page.vSwapFeesLowPct.textContent = fromIsToken ? '' : `(${fmtPct(bestSwapPct)}%)`
+    const bestSwapPct = swap.estimate.realisticBestCase * fromExchangeRate / swapped * 100
```
buck54321, line 1827: Hmm. This line seems wrong and the units don't pan out for a percent. Is this maybe the reason @JoeGruffins saw a crazy value on testnet.
image

buck54321: Oooh. Nevermind. You're using `1` for non-tokens. The exchange rate actually needs to be applied to `swapped`, but the effect is the same. Carry on.

client/webserver/site/src/js/markets.ts
```diff
@@ -1770,51 +1770,98 @@ export default class MarketsPage extends BasePage {
...
+    let quoteExchangeRate = 1
 
     if (market.base.token) {
       const parent = app().assets[market.base.token.parentID]
       baseUnitInfo = parent.unitInfo
+
+      const tokenFiatRate = app().fiatRatesMap[market.base.id]
+      const parentFiatRate = app().fiatRatesMap[parent.id]
+      if (tokenFiatRate && parentFiatRate) {
+        baseExchangeRate = tokenFiatRate / parentFiatRate
```
buck54321, line 1793: I think you need a `* quoteUnitInfo.Conventional.ConversionFactor / baseUnitInfo.Conventional.ConversionFactor` here. Fiat rates use conventional units, but we're operating on atoms below. 

Say, for instance, we're trading on ETH-USDC, and the going rate is 1 ETH for 1 USDC, and that's the exact order we place. Fees are 1e4 gwei, miniscule in comparison to trade size. In that case, `baseExchangeRate` here will be 1, and applying that below to `swapped` would do nothing, resulting in 
```Go
pct = 1e4 /* fees */ / (1e4 /* swapped atoms */ * 1 /* xc rate */ ) * 100 = 100%
```
which is clearly wrong.

JoeGruffins (approved):

buck54321 (approved):

chappjc (approved):

PR #2111, opened 2023-02-06
eth: remove isRedeemable contract method, disable estimateRedeemGas

chappjc:
The `isRedeemable` method made more sense when the client was using a private node, however this is a vulnerability with public RPC providers, which is what we are forced to use since The Merge (boo).

This removes the `isRedeemable` public method from the Solidity contracts `ETHSwapV0` and `ERC20SwapV0`.

This also updates the solc version to 0.8.18 ([github download](https://github.com/ethereum/solidity/releases/tag/v0.8.18)), the current latest. This may or may not be required, but there are two solc bugs prior to 0.8.17:
- https://blog.soliditylang.org/2022/09/08/storage-write-removal-before-conditional-termination/
- https://blog.soliditylang.org/2022/08/08/calldata-tuple-reencoding-head-overflow-bug/

Since we are having to re-deploy the contracts, we are updating the solc version to 0.8.18+commit.87f61d96.  Note that `abigen` **v1.10.26** was used to regenerate the contract ABI bindings.

`(*assetWallet).isRedeemable` now performs the redeemable check itself. That is:
- the `swap()` contractor method can retrieve the swap state
- the swap state is `SSInitiated`
- the `secret` hashes to the `secretHash`

Finally this disables the use of `estimateRedeemGas`, which also reveals secret ahead of the redeem transaction.  There is a possibly we could use this method if we recognize a private full node as the backend, or if we ensure that `redeem` always occurs even if `estimateRedeemGas` fails or returns nonsense.  Further, this would be safe for taker (participant), but not maker (initiator).


chappjc:
Redeployed Goerli contracts, but dex-test server is still expecting the old ones!
Will comment when that's updated.

chappjc:
New contracts are now in use on dex-test.

**re-test on goerli**
- initiate: https://goerli.etherscan.io/tx/0xa17b6edeaf79791b5fc9232dc05a56d43f3a67845f3248e763b77162fae9b181
- redeem: https://goerli.etherscan.io/tx/0x9c8e03a69aa90a6e011e772d46083a617f75419c31dc7d01a989871fd142721f

chappjc:
Caution!  Still using the bonkers gas limits

image
image

Limits are revised up in https://github.com/decred/dcrdex/pull/2013.  I'll rebase that PR shortly, on this after redeploying on mainnet as well.

client/asset/eth/eth.go
```diff
@@ -1909,11 +1909,15 @@ func (w *assetWallet) Redeem(form *asset.RedeemForm, feeWallet *assetWallet, non
 
 	gasLimit, gasFeeCap := g.Redeem*n, form.FeeSuggestion
 	originalFundsReserved := gasLimit * gasFeeCap
-	gasEst, err := w.estimateRedeemGas(w.ctx, secrets, contractVer)
-	if err != nil {
+
+	/* We could get a gas estimate via RPC, but this will reveal the secret key
+	   before submitting the redeem transaction. This is not OK for maker. This
+	   could be acceptable if any error from estimateRedeemGas is not fatal, and
+	   we redeem regardless (using hard-coded gas). Disable for now.
```
buck54321, line 1916: > This could be acceptable if any error from estimateRedeemGas is not fatal, and we redeem regardless

I question whether even that is appropriate. If the risk is that the provider is in on the scame, then they could just deny our redemption broadcast.

buck54321 (approved):
Still an `isRedeemable` method in the reentry test contract at **VulnerableToReentryAttack.sol**. Do we care?

martonp (approved):

chappjc:
> Still an `isRedeemable` method in the reentry test contract at **VulnerableToReentryAttack.sol**. Do we care?

Not really.  I see that as a proof of concept.  Could really just delete it, let it be in git history for reference.

client/asset/eth/eth.go
chappjc: I agree with that.  I'll revise the comment so it doesn't suggest it as a possibility.

buck54321: I do still want to leave this code in case we one day have a privacy-preserving client wallet.🤞

buck54321 (approved):

PR #2113, opened 2023-02-07
eth: Do not fail for high gas fees.

JoeGruffins:
If our gas constants are too low, current estimates will be used that
    are more than we reserved. We need to raise our constants in this
    situation, but do not fail swaps for this reason.

Resolves an issue brought up in matrix when trading on testnet.

### DIFF
--- a/dex/networks/eth/tokens.go
+++ b/dex/networks/eth/tokens.go
@@ -199,13 +199,13 @@ var Tokens = map[uint32]*Token{
 							// map.
 							// Average of 5 transfers: 51820
 							//   [65500 48400 48400 48400 48400]
-							Swap:      175_000,
-							SwapAdd:   115_000,
-							Redeem:    87_000,
-							RedeemAdd: 33_000,
-							Refund:    67_000,
-							Approve:   65_000,
-							Transfer:  70_000,
+							Swap:      188_000,
+							SwapAdd:   124_000,
+							Redeem:    92_000,
+							RedeemAdd: 35_000,
+							Refund:    71_000,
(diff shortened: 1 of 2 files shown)


dex/networks/eth/tokens.go
```diff
@@ -199,13 +199,13 @@ var Tokens = map[uint32]*Token{
...
-							Refund:    67_000,
-							Approve:   65_000,
-							Transfer:  70_000,
+							Swap:      188_000,
+							SwapAdd:   124_000,
+							Redeem:    92_000,
+							RedeemAdd: 35_000,
+							Refund:    71_000,
+							Approve:   66_000,
+							Transfer:  72_000,
```
JoeGruffins, line 208: all are about 10% than the estimates now

chappjc: I'm revising these up a lot in https://github.com/decred/dcrdex/pull/2013/commits/52ae7d6cf31535793e22bb68c2261dc42f4977b2
IMO we shouldn't be having transactions use >80% of their reserves.

chappjc: You were right about the cause of `unfunded token swap fees: 35000000 < 41559000` that I reported in matrix.  It was indeed these gases being too low, not a reserves issue.
I think we'll close this because I'm updating and validating all the gases for both ETH and USDC on both goerli and mainnet in https://github.com/decred/dcrdex/pull/2013, which I'll take out of draft shortly

JoeGruffins:
closing in favor of #2013

PR #2116, opened 2023-02-07
app: Fix candle chart Y-Axis

### LINKED ISSUE #2114: ui: candles y-axis and high/low rates for ETH and ERC20 markets are shifted
Fresh market with two trades: one at 0.0139 ETH/DCR and a second at 0.0142 ETH/DCR:

image

The second trade (order pair) still in my latest history:

image

martonp:
`RateEncodingFactor * quoteFactor / baseFactor` gives the conventional rate, but
`RateEncodingFactor * baseFactor / quoteFactor` was being used.

Closes #2114

### DIFF
--- a/client/webserver/site/src/js/charts.ts
+++ b/client/webserver/site/src/js/charts.ts
@@ -1018,7 +1018,7 @@ export class CandleChart extends Chart {
     if (!data.candles) return
     this.market = market
     const [qFactor, bFactor] = [quoteUnitInfo.conventional.conversionFactor, baseUnitInfo.conventional.conversionFactor]
-    this.rateConversionFactor = RateEncodingFactor * bFactor / qFactor
+    this.rateConversionFactor = RateEncodingFactor * qFactor / bFactor
     let n = 25
     this.zoomLevels = []
     const maxCandles = Math.max(data.candles.length, 1000)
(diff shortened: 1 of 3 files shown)


client/webserver/site/src/js/charts.ts
```diff
@@ -1018,7 +1018,7 @@ export class CandleChart extends Chart {
     if (!data.candles) return
     this.market = market
     const [qFactor, bFactor] = [quoteUnitInfo.conventional.conversionFactor, baseUnitInfo.conventional.conversionFactor]
-    this.rateConversionFactor = RateEncodingFactor * bFactor / qFactor
+    this.rateConversionFactor = RateEncodingFactor * qFactor / bFactor
```
chappjc, line 1021: Does `bconv` in `setHighLow` also need to change or is that inheriting this?

chappjc: The USDC market is a good illustration of the shifts:

image

Looks like in the case of `bconv` it's neglecting the `RateEncodingFactor`

martonp: It was using `bconv`, but it should be `qconv`.

JoeGruffins (approved):

buck54321 (approved):

chappjc:
Not tested but merging with current reviews.  I was sure the `RateEncodingFactor` would be needed instead of `conventional.conversionFactor` in `setHighLow`, but looks like it's working.

Issue #2068, opened 2023-01-24
ui/markets: active orders section could use some work


buck54321:
1. Resolved orders should either disappear or there should be an option to clean it up. I'm imagining a little broom icon or something.
2. I wonder if the dropdown section could be displayed automatically on hover instead. Not inline, but as an absolutely positioned element with high z-index. This could potentially be troublesome on mobile though. Not sure.

![image](https://user-images.githubusercontent.com/6109680/214273337-a6a8cc86-32ee-4ed8-a507-b31061603d54.png)

norwnd:
Another minor thing we might wanna look into is how to arrange elements such that they don't push into each other, for example, here seems like **Age** column could be moved to the right slightly (**Rate** and **Filled** are also getting mixed up in undesirable way, but could be slightly alleviated by https://github.com/decred/dcrdex/issues/2174):

<img width="200" alt="image" src="https://user-images.githubusercontent.com/112318969/221482801-919a5102-7216-4a4e-aef0-813d6ae7561d.png">

Maybe, for all practical purposes, this ^ could be solved by simply adjusting padding/margins a bit (as opposed to significantly reworking table structure such that it becomes adaptive to its contents).

buck54321:
> Another minor thing we might wanna look into is how to arrange elements such that they don't push into each other, for example, here seems like **Age** column could be moved to the right slightly (**Rate** and **Filled** are also getting mixed up in undesirable way, but could be slightly alleviated by #2174):
> 
> <img alt="image" width="200" src="https://user-images.githubusercontent.com/112318969/221482801-919a5102-7216-4a4e-aef0-813d6ae7561d.png">
> 
> Maybe, for all practical purposes, this ^ could be solved by simply adjusting padding/margins a bit (as opposed to significantly reworking table structure such that it becomes fully adaptive to its contents).

I'm using `display: grid`, so I added `column-gap: 5px`.

buck54321:
> One other related feature-request I can think of is making `cancel order` and `order details` buttons a bit more prominent (in a similar spirit to #2117)

Added some x padding and margin.

PR #2117, opened 2023-02-07
ui: make order detailsDiv slightly easier to drop down

chappjc:
This makes the order `status` text as well as any "active" light indicator expand the order details div, in addition to clicking the `expander`. This is just to make it a bit easier to get the order details to expand, without having the entire header element under user-order trigger expanding it, although that might be ok too.

To clarify, you can click the status text in addition to the expander icon to the order `detailsDiv` to expand down.

image
(click) =>
image

### DIFF
--- a/client/webserver/site/src/html/bodybuilder.tmpl
+++ b/client/webserver/site/src/html/bodybuilder.tmpl
@@ -104,7 +104,7 @@
 {{end}}
 
 {{define "bottom"}}
-<script src="/js/entry.js?v=ZZS892G"></script>
+<script src="/js/entry.js?v=3KW3Tiwr"></script>
 </body>
 </html>
 {{end}}
--- a/client/webserver/site/src/js/markets.ts
+++ b/client/webserver/site/src/js/markets.ts
@@ -1281,7 +1281,7 @@ export default class MarketsPage extends BasePage {
         app().bindInternalNavigation(div)
       }
 
-      Doc.bind(header.expander, 'click', () => {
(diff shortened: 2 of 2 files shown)


martonp (commented):
I think you may as well go for the whole row.

chappjc:
Whole header row clickable now.  The `expander` is not simply informative and suggestive.

norwnd:
Working well! What do you think of additionally highlighting header background, similar to how markets (on left dock) are highlighted on hover (here is [a simple change to do that](https://github.com/norwnd/dcrdex/commit/210224eff61885f9b4391a9d9ba7d1ce110ee84e)) ?

[header-highlighting.webm](https://user-images.githubusercontent.com/112318969/219596183-864f1e73-0b78-4b16-b8d8-ba01fe40f0fb.webm)

--------------------------------------------------------

And on top of that ^ looks like it would be cleaner, more snappy if you will, to remove "extra" highlighting (double-highlighting) **on the arrow itself** (here is [the commit to do that](https://github.com/norwnd/dcrdex/commit/5a9e505da4ea04392161f40dd418cc451135013b)):

[remove-double-highlighting.webm](https://user-images.githubusercontent.com/112318969/219597789-9e936f64-a211-4140-9329-2ff131fe1d6c.webm)

buck54321:
Related #2068

buck54321 (approved):

buck54321:
I'm going to follow up and take a look at #2068. I'll look at highlighting there, @norwnd.

PR #2118, opened 2023-02-07
client/asset/eth: don't fail in swapGas on estimateInitGas RPC failure

chappjc:
Testing a USDC swap, my maker failed (at first) to create the first swap in the sequence with the following error:

```
[DBG] CORE: Swappable match 6cdfd20fe1b2bbb168f1c0071ef625d71c1bfc04b4afb6c653a06d02de094788 for order a595439a976bbc4681297efbbbf306e33020fea9e763eec8b90d1ab0137ae2f5 (Maker)
[DBG] CORE: Using stored change coin address: ..., amount:47320000, fees:43560000 (usdc.eth) for order a595439a976bbc4681297efbbbf306e33020fea9e763eec8b90d1ab0137ae2f5 matches
[TRC] CORE[eth][ETH][RPC]: using cached header from "eth-mainnet.blastapi.io"
[ERR] CORE[eth][USDC.ETH]: (60001) error estimating swap gas: execution reverted: transfer from failed
[ERR] CORE: notify: |ERROR| (order) Swap send error - Error encountered sending a swap output(s) worth 47.320000 USDC on order a595439a - Order: a595439a976bbc4681297efbbbf306e33020fea9e763eec8b90d1ab0137ae2f5
[TRC] CORE: notify: |DATA| (balance) balance updated
[ERR] CORE: Route 'match' request handler error (DEX ...): runMatches - {tick of order a595439a976bbc4681297efbbbf306e33020fea9e763eec8b90d1ab0137ae2f5 failed: ... tick: {swapMatches order a595439a976bbc4681297efbbbf306e33020fea9e763eec8b90d1ab0137ae2f5 - {error sending swap transaction: error getting gas fees: execution reverted: transfer from failed}}}
```

This was not fatal because it retried on the next tick about a minute later, but for some reason it appears that `c.cb.EstimateGas` for the init tx payload returned `execution reverted: transfer from failed`.  May just be a provider error, but I don't think we need to fail to swap because we can use the known gas cap values (just like we do in `redeemGas`).

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -1454,11 +1454,12 @@ func (w *assetWallet) swapGas(n int, ver uint32) (oneSwap, nSwap uint64, approve
 	// If a live estimate is greater than our estimate from configured values,
 	// use the live estimate with a warning.
 	if gasEst, err := w.estimateInitGas(w.ctx, n, ver); err != nil {
-		w.log.Errorf("(%d) error estimating swap gas: %v", w.assetID, err)
+		w.log.Errorf("(%d) error estimating swap gas (using expected gas cap instead): %v", w.assetID, err)
 		// TODO: investigate "gas required exceeds allowance".
-		return 0, 0, false, err
+		// TODO: investigate "error getting gas fees: execution reverted: transfer from failed"
+		// return 0, 0, false, err
 	} else if gasEst > nSwap {
-		w.log.Warnf("Swap gas estimate %d is greater than the server's configured value %d. Using live estimate + 10%.", gasEst, nSwap)
+		w.log.Warnf("Swap gas estimate %d is greater than the server's configured value %d. Using live estimate + 10%%.", gasEst, nSwap)
 		nSwap = gasEst * 11 / 10 // 10% buffer
 		if n == 1 && nSwap > oneSwap {
 			oneSwap = nSwap
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -1389,7 +1389,9 @@ type gasEstimate struct {
 	// IF the redeem asset is a fee-family asset. Otherwise Redeem and RedeemAdd
 	// are zero.
 	dexeth.Gases
-	// Additional fields are based on live estimates of the swap.
+	// Additional fields may be based on live estimates of the swap. Both oneGas
+	// and nGas will include both swap and redeem gas, but note that the redeem
+	// gas may be zero if the redeemed asset is not ETH or an ETH token.
 	oneGas, nGas, nSwap, nRedeem uint64
 }
 
@@ -1412,14 +1414,16 @@ func (w *assetWallet) initGasEstimate(n int, initVer, redeemVer, redeemAssetID u
 		return nil, fmt.Errorf("error calculating swap gas: %w", err)
 	}
 
-	est.oneGas = est.Swap + est.Redeem
-	est.nGas = est.nSwap + est.nRedeem
+	est.oneGas = est.Swap
+	est.nGas = est.nSwap
 
 	if redeemW := w.wallet(redeemAssetID); redeemW != nil {
 		est.Redeem, est.nRedeem, err = redeemW.redeemGas(n, redeemVer)
 		if err != nil {
 			return nil, fmt.Errorf("error calculating fee-family redeem gas: %w", err)
 		}
+		est.oneGas += est.Redeem
+		est.nGas += est.nRedeem
 	}
 
 	return
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -1667,8 +1667,6 @@ func (w *ETHWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin, uint6
 		return nil, nil, 0, fmt.Errorf(s, a...)
 	}
 
-	receipts := make([]asset.Receipt, 0, len(swaps.Contracts))
-
 	var reservedVal uint64
 	for _, input := range swaps.Inputs { // Should only ever be 1 input, I think.
 		c, is := input.(*fundingCoin)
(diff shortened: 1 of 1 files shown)


martonp (approved):

JoeGruffins (approved):
Can you make the log under this `10%%` so the `10%` prints correctly?

buck54321 (approved):

client/asset/eth/eth.go
```diff
@@ -1454,11 +1454,12 @@ func (w *assetWallet) swapGas(n int, ver uint32) (oneSwap, nSwap uint64, approve
 	// If a live estimate is greater than our estimate from configured values,
 	// use the live estimate with a warning.
 	if gasEst, err := w.estimateInitGas(w.ctx, n, ver); err != nil {
-		w.log.Errorf("(%d) error estimating swap gas: %v", w.assetID, err)
+		w.log.Errorf("(%d) error estimating swap gas (using expected gas cap instead): %v", w.assetID, err)
 		// TODO: investigate "gas required exceeds allowance".
-		return 0, 0, false, err
+		// TODO: investigate "error getting gas fees: execution reverted: transfer from failed"
+		// return 0, 0, false, err
```
chappjc, line 1460: I believe whether we fail depends on the caller.  `FundOrder` should probably fail, whilst `Swap` should not.
This will take some refactoring unfortunately.

chappjc: Proposed solution in https://github.com/decred/dcrdex/pull/2118/commits/6101a6cb554437df45c334aef7dcbe3d5593e949

chappjc:
Pushed two new commits.  https://github.com/decred/dcrdex/pull/2118/commits/0bea9010b405f0901dc38f4e29660bf7979d206d uses the batched gas limits as a fallback if for some reason reserves are insufficient for the naive `oneSwap * n` limit that does not consider batch savings.

https://github.com/decred/dcrdex/pull/2118/commits/6101a6cb554437df45c334aef7dcbe3d5593e949 lets the live estimate remain fatal for order funding and pre-order estimates, but not for the swap when negotiation is already underway. 

Need to re-test, but putting back to ready-for-review.

JoeGruffins:
Will still fail in swap if we did not reserve enough, but it is sure to get an estimate or fail in `FundOrder` so less likely the second estimate is higher.

What if there was one swap, and the second live estimate was, for whatever reason, one gas higher than the initial. Would that fail?

chappjc:
> What if there was one swap, and the second live estimate was, for whatever reason, one gas higher than the initial. Would that fail?

Really makes me question the utility of doing a live estimate at swap time (for n=1) at all.  This would be an especially frustrating failure mode if the wallet has plenty of available balance and it's just the reserved `amt`s in the funding coins that seem insufficient.

chappjc:
Sorry for changing this one so much after your reviews @martonp and @buck54321.  https://github.com/decred/dcrdex/pull/2118#issuecomment-1424468349

The question @JoeGruffins posed remains: if it's a single lot swap (n=1), and order funding got a live estimate of X and used that for reserves, then the live estimate at time of swap got X+1, we'd fail to swap because it exceeds reserves.  Options:
- that's correct, block the swap to avoid a failed transaction
- Swap anyway with the "third fallback that only uses our known gases"  (see code comments, but apply that in the `n=1` case too).

client/asset/eth/eth.go
```diff
@@ -1682,16 +1687,32 @@ func (w *ETHWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin, uint6
 		swapVal += contract.Value
 	}
 
-	oneSwap, _, _, err := w.swapGas(1, swaps.Version)
+	// Set the gas limit as high as reserves will allow.
+	n := len(swaps.Contracts)
+	oneSwap, nSwap, _, err := w.swapGas(n, swaps.Version, false) // allow live est to fail, using our own gas values
```
martonp, line 1692: If the live estimate fails here, and our hardcoded estimates are higher than the actual live estimates, then wouldn't `swapVal+fees > reservedVal` always be true?

chappjc: I thought they would be equal.  `reservedVal` should turn out to be `swapVal + g.Swap * n`.  Basing that on my understanding on how the reserves (via the `fundingCoin`) are initially set.

chappjc:
I don't think I have the bandwidth to follow through on this swapGas PR, but there are some issues that need to get resolved for 0.6.  Maybe these change (or the gist) can be absorbed into https://github.com/decred/dcrdex/pull/2129?

client/asset/eth/eth.go
```diff
@@ -1395,18 +1395,19 @@ type gasEstimate struct {
...
+	if g := w.gases(initVer); g == nil {
 		return nil, fmt.Errorf("no gas table")
+	} else { // scoping g
+		est.Refund = g.Refund
 	}
 
-	est.Refund = g.Refund
-
-	est.Swap, est.nSwap, _, err = w.swapGas(n, initVer)
+	est.Swap, est.nSwap, _, err = w.swapGas(n, initVer, true) // fail if live gas estimate fails
```
buck54321, line 1410: While you're here, a few lines down, it says
```Go
est.oneGas = est.Swap + est.Redeem
est.nGas = est.nSwap + est.nRedeem
```

Adding the redeem value there is wrong and pointless. They're both zero. I think those lines needs to come after the redeem wallet check.

chappjc: Good catch.  This would only affect self-trades with same asset or ETH+TOKEN, right?

client/asset/eth/eth.go
```diff
@@ -1454,11 +1458,14 @@ func (w *assetWallet) swapGas(n int, ver uint32) (oneSwap, nSwap uint64, approve
...
 	// use the live estimate with a warning.
 	if gasEst, err := w.estimateInitGas(w.ctx, n, ver); err != nil {
-		w.log.Errorf("(%d) error estimating swap gas: %v", w.assetID, err)
+		w.log.Errorf("(%d) error estimating swap gas (using expected gas cap instead): %v", w.assetID, err)
+		if requireLiveEst {
+			return 0, 0, false, err
+		} // else go with what we know
 		// TODO: investigate "gas required exceeds allowance".
-		return 0, 0, false, err
+		// TODO: investigate "error getting gas fees: execution reverted: transfer from failed" (seems to be approval not actually ready)
```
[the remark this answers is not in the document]
chappjc: Oh, it was also testnet or mainnet, cannot recall.  I may have commented on the mainnet eth PR.  Almost positive it was the very first trade with the token and it did the approval on the fly when order was created.

client/asset/eth/eth.go
chappjc: It's generally a sign of major problems coming if the live estimates are higher than the hard-coded estimates.  In other words, it is (and has been) normal to have the hard coded estimates be what comes out of `swapGas`.
Consider the code on `master`, the checks below haven't changed, only added a fallback to the more realistic batched gas limits.

client/asset/eth/eth.go
chappjc: Like this? https://github.com/decred/dcrdex/pull/2118/commits/ba8b9fa1fa5f5e2e911fd023e1088fe030abfa92

I'm now wondering if `maxOrder` needs to include refund in it's `oneFee` value.

chappjc: > I'm now wondering if `maxOrder` needs to include refund in it's `oneFee` value.

Oh, @JoeGruffins is doing this in 7222c01dd512108a1e337b42d33a9b4e6617ce09.  Good.

chappjc:
Closing in favor of https://github.com/decred/dcrdex/pull/2143

PR #2119, opened 2023-02-07
client/core: return from Run if db.Accounts fail in initialize

chappjc:
After testing bonds PRs with my testnet dexc db, I went back to master or another PR, and that failed to load the Accounts because it didn't know v3 of `db.AccountInfo`.  However, i barely noticed this because it's not a fatal error.

This acts on a long standing comment to make this error fatal.  <s>I'm just using panic/recover as suggested by the comment because I didn't really want to add and `error` return to `initialize`, although that wouldn't be horrible now that I look again.</s> Now returning an error from `initialize`.


chappjc:
Switched to returning an error instead of panic.

JoeGruffins (approved):

PR #2124, opened 2023-02-10
core: remove anti-pattern in `core.connectDEX`

ukane-philemon:
This resolves:
>	TODO: split the above code into a dexConnection constructor, and the below into
>      a startDexConnection function so we don't have the anti-pattern of
>	returning a non-nil object with a non-nil error and requiring the caller
>	to check both!

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -3719,10 +3719,6 @@ func (c *Core) tempDexConnection(dexAddr string, certI interface{}) (*dexConnect
 // as either a string filename, or []byte file contents.
 func (c *Core) GetDEXConfig(dexAddr string, certI interface{}) (*Exchange, error) {
 	dc, err := c.tempDexConnection(dexAddr, certI)
-	if dc != nil {
-		// Stop (re)connect loop, which may be running even if err != nil.
-		defer dc.connMaster.Disconnect()
-	}
 	if err != nil {
 		return nil, err
 	}
@@ -3989,20 +3985,19 @@ func (c *Core) DiscoverAccount(dexAddr string, appPW []byte, certI interface{}) 
 	var ready bool
 	if !existingConn {
 		dc, err = c.tempDexConnection(host, certI)
-		if dc != nil { // (re)connect loop may be running even if err != nil
-			defer func() {
-				// Either disconnect or promote this connection.
-				if !ready {
-					dc.connMaster.Disconnect()
-					return
-				}
-
-				c.upgradeConnection(dc)
-			}()
-		}
 		if err != nil {
 			return nil, false, err
 		}
+
+		defer func() {
+			// Either disconnect or promote this connection.
+			if !ready {
+				dc.connMaster.Disconnect()
+				return
+			}
+
+			c.upgradeConnection(dc)
+		}()
 	}
 
 	// Older DEX server. We won't allow registering without an HD account key,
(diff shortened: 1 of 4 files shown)


client/core/core.go
```diff
@@ -7383,18 +7389,19 @@ func (c *Core) connectDEX(acctInfo *db.AccountInfo, temporary ...bool) (*dexConn
...
+// startDexConnection attempts to connect the provided dexConnection. dc must be
+// a new dexConnection returned from newDEXConnection above. Callers can choose
+// to stop reconnect retries and any current goroutine for the provided
+// dexConnection using dc.connMaster.Disconnect().
+func (c *Core) startDexConnection(acctInfo *db.AccountInfo, dc *dexConnection) error {
 	// Start listening for messages. The listener stops when core shuts down or
 	// the dexConnection's ConnectionMaster is shut down. This goroutine should
 	// be started as long as the reconnect loop is running. It only returns when
 	// the wsConn is stopped.
+	listen := dc.reportingConnects == 1
```
JoeGruffins, line 7404: Should be `listen := dc.broadcastingConnect()` ?

JoeGruffins (approved):
Looks good to me.

client/core/core.go
ukane-philemon: Yeah, that works too.

ukane-philemon:
Just rebased...

buck54321 (approved):

ukane-philemon:
@buck54321, sorry I didn't notice this needed a rebase. Just did.

client/core/core.go
```diff
@@ -6686,6 +6674,48 @@ func (c *Core) initialize() error {
 	return nil
 }
 
+// connectAccount makes a connection to the DEX for the given account. If a
+// non-nil dexConnection is returned, it was inserted into the conns map even if
+// the initial connection attempt failed (connected == false), and the connect
+// retry / keepalive loop is active.
+func (c *Core) connectAccount(acct *db.AccountInfo) (dc *dexConnection, connected bool) {
```
chappjc, line 6681: Can remove this method.  We don't use it anymore.  It was only designed for `initialized`, and is simpler doing just what's needed there.

client/core/core.go
```diff
@@ -7635,12 +7665,29 @@ const (
...
 func (c *Core) connectDEXWithFlag(acctInfo *db.AccountInfo, flag connectDEXFlag) (*dexConnection, error) {
+	dc, err := c.newDEXConnection(acctInfo, flag)
+	if err != nil {
+		return nil, err
+	}
+
+	err = c.startDexConnection(acctInfo, dc)
+	if err != nil {
+		dc.connMaster.Disconnect() // stop any retry loop for this new connection.
+		return nil, err
```
chappjc, line 7683: This is right for 90% of the use cases, but it breaks critical functionality of `initialize`.  The `initialize` method will have to work differently (just logging errors from `c.startDexConnection` but not stopping the conn master because the reconnect loop is there for the initial connection, not just reestablishing a dropped connection.

chappjc (changes requested):

client/core/core.go
chappjc: Ohh, you had this PR up when `initialize` was using `connectAccount`... which would have allowed the connect loop to run.  Sorry. You can either wire `connectAccount` back into `initialize` or change `initialize` to do it.

ukane-philemon: Okay.

client/core/core.go
```diff
@@ -7630,17 +7638,39 @@ const (
...
 )
 
-// connectDEX establishes a ws connection to a DEX server using the provided
-// account info, but does not authenticate the connection through the 'connect'
-// route. If temporary is provided and true, the c.listen(dc) goroutine is not
-// started so that associated trades are not processed and no incoming requests
-// and notifications are handled. A temporary dexConnection may be used to
-// inspect the config response or check if a (paid) HD account exists with a DEX.
+// connectDEX is like connectDEXWithFlag but will not use any flag thereby
+// ensuring a retry loop is started for failed connections. Use
```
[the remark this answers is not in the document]
ukane-philemon: Derp, thanks for the review.

chappjc (approved):
Really so much cleaner and less confusing now.  Just revise that command and it's g2g.

ukane-philemon:
Uhmm, afk atm.

chappjc:
Ah, no worries.  I'll tweak real quick and merge.

ukane-philemon:
Thanks.

PR #2129, opened 2023-02-13
client/eth: Fix init estimates.

### LINKED ISSUE #2123: eth: Consider block gas limit.
If you try to get a swap estimate on our simnet with over 40ish swaps, you git an error `gas required exceeds allowance (n).` This n is the simnet harness's block gas limit. That value starts as `12,487,783` in our simnet genesis file. On mainnet it is currently `30,000,000` so pretty close.

I think ultimitely `MaxOrder` needs to consider the current block gas limit and either restrict based on some percent of the max, or the orders need to be broken up into multiple txn.

It seems like the mainnet block gas limit is just set to 30 million at the moment? I'm also not sure what percent of of one block's gas we can expect to use comfortably.

We might also need to check that too many redeems would not be called in the same transaction.

I remember seeing some code in our repo about the block gas limit, but I cannot remember where it was now. I don't think the client is currently handling this properly however.

Also, as an aside, for `EstimateGas`, although the limit ends up being the block gas limit, there is also a configurable rpc limit, with the default being `50,000,000`. If some providers adjust this it may unexpectedly fail. This gas limit appears to affect all rpc calls.

JoeGruffins:
Make estimates aware of the max gas limit for transactions that we later
    use when trading. Use one fourth of the block gas limit or the tx fee
    limit per max fee per gas as the top end, whichever is lower. Set
    simnet's block gas limit to 30 million, the same as test and mainnet.

closes #2123


JoeGruffins:
I thought this was right but worst case is less that best case in the UI. Must be missing something.

client/asset/eth/eth.go
```diff
@@ -1451,9 +1452,20 @@ func (w *assetWallet) swapGas(n int, ver uint32) (oneSwap, nSwap uint64, approve
 	// If we've approved the contract to transfer, we can get a live
 	// estimate to double check.
 
+	// The amount we can estimate and ultimately the amount we can use in a
+	// single transaction is limited by the block gas limit. We attempt to
+	// use one fourth of that. Estimate the number of swaps we can safely
+	// estimate here.
+	maxN := n
+	if n > 1 && nSwap > blockGasLimit {
```
chappjc, line 1460: My instinct says this should be an error since the inputs are unsatisfiable, and we don't return the reduced value of `n` to the caller so they know what the estimate actually describes.

I had guessed the resolution to https://github.com/decred/dcrdex/issues/2123 would be to modify code a little higher up in `(*assetWallet).estimateSwap` (or perhaps `maxOrder`/`preSwap`) so that they become aware of when `n` is too large, and only return a `asset.SwapEstimate` with the largest possible `lots`

JoeGruffins: Is is bad to scale the estimate? In practice all the SwapAdd are almost the same. The code later will split the swaps at the same break points.

chappjc: It's fine I think to scale up the output of `w.estimateInitGas`, I just thought the problem to solve was higher level callers asking for estimates on impossible swaps...  But now I suppose the thinking it they are only impossible when *batched* but that's not to prevent the user from making a 100 lot order (because it will be split into multiple batches)?

JoeGruffins: I didn't think there were any impossible swaps. Will test this more and see if there is something obviously wrong. Until now large orders were not possible because of the error when trying to get an estimate.

JoeGruffins:
hmm, seeing some weirdness in logs. Made a large swap on simnet:
image

Everything seems to have gone ok:
image

But there are some troubling points in the log. For one I expected it to break the order up. The estimate will break up about 66 swaps with the current code, but this seems to have tried to send everything in one tx.

Some logs.
```
2023-02-14 12:33:48.097 [DBG] CORE: Using stored change coin address: 0x946dfaB1AD7caCFeF77dE70ea68819a30acD4577, amount:27b4244380 (eth) for order 2149e90ff3662477f230505c916bd6d5b5eda513ba2bc461ac678c856a13cbb4 matches
2023-02-14 12:33:48.100 [TRC] CORE[eth][ETH][RPC]: using cached header from "IPC"
2023-02-14 12:33:48.101 [TRC] CORE[eth][ETH][RPC]: Sending signed tx via "IPC"
2023-02-14 12:33:48.102 [DBG] CORE: Using stored change coin address: 0x946dfaB1AD7caCFeF77dE70ea68819a30acD4577, amount:27b4244380 (eth) for order 2149e90ff3662477f230505c916bd6d5b5eda513ba2bc461ac678c856a13cbb4 matches
2023-02-14 12:33:48.105 [TRC] CORE[eth][ETH][RPC]: using cached header from "IPC"
2023-02-14 12:33:48.106 [TRC] CORE[eth][ETH][RPC]: Sending signed tx via "IPC"
2023-02-14 12:33:48.108 [ERR] CORE: notify: |ERROR| (order) Swap send error - Error encountered sending a swap output(s) worth 103.528210000 ETH on order 2149e90f - Order: 2149e90ff3662477f230505c916bd6d5b5eda513ba2bc461ac678c856a13cbb4
2023-02-14 12:33:48.108 [ERR] CORE: 127.0.0.1:17273 tick: {swapMatches order 2149e90ff3662477f230505c916bd6d5b5eda513ba2bc461ac678c856a13cbb4 - {error sending swap transaction: Swap: initiate error: tx fee (1.49 ether) exceeds the configured cap (1.00 ether), error sending swap transaction: Swap: initiate error: tx fee (1.13 ether) exceeds the configured cap (1.00 ether)}}
2023-02-14 12:33:48.108 [TRC] CORE[eth][ETH][RPC]: using cached header from "IPC"
```
Notice the amount `27b4244380` hopefully just not printing correctly. Then it looks like it tried to send the whole sum, or all the swaps. And then the error `Swap: initiate error: tx fee (1.13 ether) exceeds the configured cap (1.00 ether)`. I remember we did something to solve this with the internal node, but not sure what we need to do now. Maybe consider this 1.00 ether in the max swaps. If this error is from the provider we would have no choice.

But then it seems to break up all the swaps and send them one at a time, and that works but I didn't expect it.

```
2023-02-14 12:33:55.525 [INF] CORE: Broadcasted transaction with 1 swap contracts for order 2149e90ff3662477f230505c916bd6d5b5eda513ba2bc461ac678c856a13cbb4. Assigned fee rate = 200. Receipts (eth): [{ tx hash: 0x527970de89b482ecef3b82eb3d99b894300c87c35c86c3d898934b8055563054, contract address: 0x2f68E723b8989BA1C6a9F03E42F33cb7Dc9D606f, secret hash: 7e9333fbb08c314b846cfc746b8c49ac0804a05b2772f02b9dde8e7ccbce5889 }].
2023-02-14 12:33:55.525 [DBG] CORE: Saving change coin address: 0x946dfaB1AD7caCFeF77dE70ea68819a30acD4577, amount:2776248770 (eth) to DB for order 2149e90ff3662477f230505c916bd6d5b5eda513ba2bc461ac678c856a13cbb4
2023-02-14 12:33:55.526 [INF] CORE: Contract coin 0x527970de89b482ecef3b82eb3d99b894300c87c35c86c3d898934b8055563054 (eth), value = 1013170000, refundable at 2023-02-14 12:36:45 +0900 JST (receipt = { tx hash: 0x527970de89b482ecef3b82eb3d99b894300c87c35c86c3d898934b8055563054, contract address: 0x2f68E723b8989BA1C6a9F03E42F33cb7Dc9D606f, secret hash: 7e9333fbb08c314b846cfc746b8c49ac0804a05b2772f02b9dde8e7ccbce5889 }), match = f7e2880b6da7c1db4d865184b3ed7d4a19e696a990065e15a33cafbde9095618
2023-02-14 12:33:55.527 [INF] CORE: Notifying DEX 127.0.0.1:17273 of our eth swap contract 0x527970de89b482ecef3b82eb3d99b894300c87c35c86c3d898934b8055563054 for match f7e2880b6da7c1db4d865184b3ed7d4a19e696a990065e15a33cafbde9095618
2023-02-14 12:33:55.527 [DBG] CORE: Using stored change coin address: 0x946dfaB1AD7caCFeF77dE70ea68819a30acD4577, amount:2776248770 (eth) for order 2149e90ff3662477f230505c916bd6d5b5eda513ba2bc461ac678c856a13cbb4 matches
2023-02-14 12:33:55.530 [TRC] CORE[eth][ETH][RPC]: using cached header from "IPC"
2023-02-14 12:33:55.531 [TRC] CORE[eth][ETH][RPC]: Sending signed tx via "IPC"
2023-02-14 12:33:55.532 [INF] CORE: Broadcasted transaction with 1 swap contracts for order 2149e90ff3662477f230505c916bd6d5b5eda513ba2bc461ac678c856a13cbb4. Assigned fee rate = 200. Receipts (eth): [{ tx hash: 0xb7fc9666199b98be7ef92eb022bc15ba77aac270fd094771aa088ac41b284fa5, contract address: 0x2f68E723b8989BA1C6a9F03E42F33cb7Dc9D606f, secret hash: febf60da143e02c290723fd6453ddfbecf740c9b823fcdd0b420bbdc59d56c18 }].

.....

2023-02-14 12:33:56.192 [DBG] CORE: notify: |POKE| (order) Swaps initiated - Sent swaps worth 103.528210000 ETH on order 2149e90f - Order: 2149e90ff3662477f230505c916bd6d5b5eda513ba2bc461ac678c856a13cbb4
2023-02-14 12:33:56.192 [DBG] CORE: Slow tick: trade 2149e90ff3662477f230505c916bd6d5b5eda513ba2bc461ac678c856a13cbb4 processed in 763.464344ms, blocked for 270ns
```

JoeGruffins:
Seems to be working after taking 1 eth max fee into account. Just one random error:
```
2023-02-14 15:27:16.213 [ERR] CORE: attempting to unlock more than is reserved for order 5030c1f342aa9c0871c884e776db18f41fa97e8c36896f9da6778a8c9572082c. unlocking reserved amount instead: 168627449 reserved, unlocking 168627451
```

JoeGruffins:
Seems ok.

JoeGruffins:
Also adjust nSwap because oneSwap will be used more if there are more transacitons, and it uses more gas. https://github.com/decred/dcrdex/compare/91a0de7ca5ad00592d29a47feee51c1019cd1b14..0a64ba6c49a253564674ae6f973be45190e4a8e6

client/asset/eth/eth.go
```diff
@@ -1451,13 +1470,39 @@ func (w *assetWallet) swapGas(n int, ver uint32) (oneSwap, nSwap uint64, approve
...
 	// use the live estimate with a warning.
-	if gasEst, err := w.estimateInitGas(w.ctx, n, ver); err != nil {
+	gasEst, err := w.estimateInitGas(w.ctx, maxN, ver)
+	if err != nil {
 		w.log.Errorf("(%d) error estimating swap gas: %v", w.assetID, err)
-		// TODO: investigate "gas required exceeds allowance".
 		return 0, 0, false, err
-	} else if gasEst > nSwap {
+	}
+	if maxN != n {
```
martonp, line 1498: This will overestimate unless `n % maxN == 0`, but I guess that's fine.

client/asset/eth/eth.go
```diff
@@ -84,6 +84,10 @@ const (
 	// coinIDTakerFoundMakerRedemption is a prefix to identify one of CoinID formats,
 	// see DecodeCoinID func for details.
 	coinIDTakerFoundMakerRedemption = "TakerFoundMakerRedemption:"
+
+	// maxTxFeeGwei is the default max amout of eth that can be used in one
+	// transaction.
+	maxTxFeeGwei = 1_000_000_000
```
martonp, line 90: Why do we need this extra limit?

JoeGruffins: There is default max set in the geth host, possibly. It is set in our nodes by default. I do not know if there is a way to test for it or if it is present for providers or is the same max. https://github.com/ethereum/go-ethereum/blob/dbd6c1324dda8bb93c7a8297446fce7998cab4e1/internal/ethapi/api.go#L2084-L2097

We dealt with this with the internal node here https://github.com/decred/dcrdex/pull/1490

JoeGruffins: I'll add more to the comment anyway.

JoeGruffins: The only api that I can find that gives some info about the host is `admin.nodeInfo` and we probably cannot access admin and it doesn't even have the value. If anyone knows how to sniff this value out would be helpful. There is also another config variable the rpc gas cap we really should know, but the default is 50 million so its probably ok for now: https://github.com/ethereum/go-ethereum/blob/dbd6c1324dda8bb93c7a8297446fce7998cab4e1/internal/ethapi/api.go#L1050

JoeGruffins: I guess all the compliant providers use geth?

JoeGruffins: Updated the comment https://github.com/decred/dcrdex/compare/0a64ba6c49a253564674ae6f973be45190e4a8e6..f004787d95fc2874562e8c873571bcbd5f6c7721

JoeGruffins:
Amended the commit message.

client/asset/eth/eth.go
chappjc: To restate the reason:

`--rpc.txfeecap value                Sets a cap on transaction fee (in ether) that can be sent via the RPC APIs (0 = no cap) (default: 1)`

And it's probably safe to assume that at least some providers will be using geth with the default.

chappjc (approved):
Untested, but LGTM.

client/asset/eth/eth.go
```diff
@@ -1451,13 +1475,39 @@ func (w *assetWallet) swapGas(n int, ver uint32) (oneSwap, nSwap uint64, approve
...
-		// TODO: investigate "gas required exceeds allowance".
 		return 0, 0, false, err
-	} else if gasEst > nSwap {
+	}
+	if maxN != n {
+		// If we needed to adjust the max earlier, and the estimate did
+		// not error, multiply the estimate by the fraction of swaps we
+		// were able to estimate.
+		multBy := float64(n) / float64(maxN)
+		gasEst = uint64(float64(gasEst) * multBy)
```
buck54321, line 1508: I don't think this math checks, and I can't think of a way to do this without getting an additional `estimateInitGas` for the remainder. Something like...

```Go
if maxN != n {
	nFull := n / maxN
	gasEst *= uint64(nFull)
	if nRemain := n % maxN; nRemain > 0 {
		remainEst, err := w.estimateInitGas(w.ctx, nRemain, ver)
		if err != nil {
			w.log.Errorf("(%d) error estimating swap gas for remainder: %v", w.assetID, err)
			return 0, 0, false, err
		}
		gasEst += remainEst
	}
}
```

buck54321:
> But then it seems to break up all the swaps and send them one at a time, and that works but I didn't expect it.

Probably because they were marked `suspect` after a failed swap.

chappjc:
> > But then it seems to break up all the swaps and send them one at a time, and that works but I didn't expect it.
> 
> Probably because they were marked `suspect` after a failed swap.

That's what I figured too, broken up at the level of `Core`.  Should have been broken up intelligently by the eth backend.

client/asset/eth/eth.go
buck54321: Actually, I think if you did it this way, you can avoid the `fold` math up above.

buck54321:
> Should have been broken up intelligently by the eth backend.

We've already added machinery to handle this in core. I'd rather not re-implement in eth if we can get that working right. It hinges on correctly setting the `WalletInfo.MaxSwapsInTx`, which this work seems to fix.

client/asset/eth/eth.go
```diff
@@ -1451,13 +1475,39 @@ func (w *assetWallet) swapGas(n int, ver uint32) (oneSwap, nSwap uint64, approve
...
+	if n > 1 && nSwap > perTxGasLimit {
+		maxN = 1
+		gas := perTxGasLimit - g.Swap
+		maxN += int(gas / g.SwapAdd)
+		// nSwap will use extra gas for every fold.
+		folds := uint64(n / maxN)
+		if n%maxN == 0 {
+			folds -= 1
+		}
+		// Assume SwapAdd is less that Swap.
```
martonp, line 1491: than*

chappjc:
Ah yes, you're right, sorry.  `swapMatchGroup(t, groupables[i:i+maxSwapsInTx], errs)` is what this is all about.

martonp (approved):

JoeGruffins:
Just rebased.

client/asset/eth/eth.go
JoeGruffins: Oh, that works. Get another estimate for the remainder.

JoeGruffins: Better to just recalculate for nSwap? Yeah that probably reads better.

client/asset/eth/eth.go
JoeGruffins: Removed this comment.

JoeGruffins:
Wait, should be `nRemain = n % nMax`

JoeGruffins:
Running live estimate on the remainder for a more accurate estimate. https://github.com/decred/dcrdex/compare/0740897845469b475a827f3873b7e0dc7a95e839..f97fd3b6f0043986a20727873d86d07b628b1a9f

However there still seems to be an issue with the max estimate not being the same as the funding code. If you click on `MaxBuy` in the UI you would expect to be able to buy that much dcr on the dcr_eth market, but instead you are met with an error upon confirming:
image

Should I try to fix here? I guess it's higher up that I have been medling.

JoeGruffins:
Including refund fees in the MaxOrder estimate https://github.com/decred/dcrdex/compare/f97fd3b6f0043986a20727873d86d07b628b1a9f..7222c01dd512108a1e337b42d33a9b4e6617ce09

I notice that that the max orders for non-eth assets do not check the other side for redemption funds. Eth and tokens will check each other. So a max order to Sell dcr on the dcr_eth market will fail at funding time when trying to do max sell if you do not have enough eth. Should this be fixed?

chappjc:
> I notice that that the max orders do not check the other side for redemption funds. So a max order to Sell dcr on the dcr_eth market will fail at preparing the trade when trying to do max sell if you do not have enough eth for redemption fees. Should this be fixed? I believe only a max order problem, when trading redeem funds are reserved.

Ugh, not for 0.6.  We need to ship this thing, and this is getting really hung up on some UX issue, which they will still get feed back when trying to fund as you say.

chappjc:
> I notice that that the max orders do not check the other side for redemption funds. So a max order to Sell dcr on the dcr_eth market will fail at preparing the trade when trying to do max sell if you do not have enough eth for redemption fees. Should this be fixed? I believe only a max order problem, when trading redeem funds are reserved.

Fixed in https://github.com/decred/dcrdex/commit/ba8b9fa1fa5f5e2e911fd023e1088fe030abfa92, I belive

client/asset/eth/eth.go
```diff
@@ -1119,17 +1142,18 @@ func (w *assetWallet) maxOrder(lotSize uint64, feeSuggestion, maxFeeRate uint64,
 		return nil, fmt.Errorf("gasEstimate error: %w", err)
 	}
 
+	refundCost := g.Refund * maxFeeRate
 	oneFee := g.oneGas * maxFeeRate
 	var lots uint64
 	if feeWallet == nil {
-		lots = balance.Available / (lotSize + oneFee)
+		lots = balance.Available / (lotSize + oneFee + refundCost)
```
buck54321, line 1149: Ouch, but yeah. I guess there's no other way.

client/asset/eth/eth.go
```diff
@@ -1451,13 +1475,45 @@ func (w *assetWallet) swapGas(n int, ver uint32) (oneSwap, nSwap uint64, approve
...
+	gasEst, err := w.estimateInitGas(w.ctx, nMax, ver)
+	if err != nil {
 		w.log.Errorf("(%d) error estimating swap gas: %v", w.assetID, err)
-		// TODO: investigate "gas required exceeds allowance".
 		return 0, 0, false, err
-	} else if gasEst > nSwap {
+	}
+	if nMax != n {
+		// If we needed to adjust the max earlier, and the estimate did
+		// not error, mulitply the estimates and add the estimate of the
```
buck54321, line 1504: mul**it**ply

client/asset/eth/eth.go
```diff
@@ -1451,13 +1475,45 @@ func (w *assetWallet) swapGas(n int, ver uint32) (oneSwap, nSwap uint64, approve
...
 
+	// The amount we can estimate and ultimately the amount we can use in a
+	// single transaction is limited by the block gas limit or the tx gas
+	// limit. Estimate the number of swaps we can safely estimate here.
+	nMax := n
+	var nRemain, nFull int
+	if n > 1 && nSwap > perTxGasLimit {
+		nMax = 1
+		gas := perTxGasLimit - g.Swap
+		nMax += int(gas / g.SwapAdd)
```
buck54321, line 1486: Shouldn't we use `nMax = w.maxSwapsInTx` since that's what **core** is going to do.

client/asset/eth/eth.go
JoeGruffins: Fixed.

client/asset/eth/eth.go
JoeGruffins: I'm not sure. This is versioned...

JoeGruffins: Why isn't `maxSwapGas` the lowest of all `gases.Swap`?

JoeGruffins:
Just fixing the typo https://github.com/decred/dcrdex/compare/6c42b1f746a9e819ba49cdee0ca34f11846270a7..f69354724604e4deb0926c832745533167658d67

client/asset/eth/eth.go
JoeGruffins: Oh I see, the highest `maxSwapGas` creates the lowest `maxSwapsInTx`

Yeah I guess it should.

JoeGruffins:
As @buck54321 pointed out, the swaps per tx is already decided for core. This value is always safe to estimate as it will always use the gas of the highest swap among versions. There is only one version so it's not so impactful right now but I agree it is correct https://github.com/decred/dcrdex/compare/f69354724604e4deb0926c832745533167658d67..18addba9f4f533686b5b274fb4da70875688c20d

chappjc (approved):

chappjc:
Will merge in within the hour based on my testing and review, unless there are comments from others.

buck54321 (approved):

PR #2130, opened 2023-02-15
client/comms: check for non-standard compliant error

ukane-philemon:
This seems to be a known issue on some versions of `MacOS` or `OS X` where an error: `x509: “invalid-cert-name” certificate is not standards compliant` is returned. Since we cannot check the error type using `errors.Is` or `errors.As`, this PR checks if the returned error contains the known error message `certificate is not standards compliant`.

Users could run varying versions of `macOS` and could be affected by this issue. This PR will ensure that a retry loop is not started because the cert is considered `Invalid`.

Issue: https://github.com/golang/go/issues/51991

For more context, I use `macOS` and was a bit flustered that `TestWsConn` kept failing, so I had to investigate.

### DIFF
--- a/client/comms/wsconn.go
+++ b/client/comms/wsconn.go
@@ -213,7 +213,7 @@ func (conn *wsConn) connect(ctx context.Context) error {
 	ws, _, err := dialer.DialContext(ctx, conn.cfg.URL, nil)
 	if err != nil {
 		var e x509.UnknownAuthorityError
-		if errors.As(err, &e) {
+		if errors.As(err, &e) || strings.Contains(err.Error(), "certificate is not standards compliant") {
 			conn.setConnectionStatus(InvalidCert)
 			if conn.tlsCfg == nil {
 				return ErrCertRequired
--- a/client/comms/wsconn.go
+++ b/client/comms/wsconn.go
@@ -13,6 +13,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -62,6 +63,20 @@ func (cs ConnectionStatus) String() string {
 	default:
 		return "unknown status"
 	}
+}
+
+// invalidCertRegexp is a regexp that helps check for non-typed x509 errors
+// caused by or related to an invalid cert.
+var invalidCertRegexp = regexp.MustCompile(".*(unknown authority|not standards compliant|not trusted)")
+
+// IsErrorInvalidCert checks if the provided error is one of the different
+// variant of an invalid cert error returned from the x509 package or is
+// ErrInvalidCert.
+func IsErrorInvalidCert(err error) bool {
+	var invalidCert x509.CertificateInvalidError
+	var unknownCertAuth x509.UnknownAuthorityError
+	return errors.Is(err, ErrInvalidCert) || errors.Is(err, invalidCert) ||
+		errors.Is(err, unknownCertAuth) || invalidCertRegexp.MatchString(err.Error())
 }
 
 // ErrInvalidCert is the error returned when attempting to use an invalid cert
@@ -212,8 +227,8 @@ func (conn *wsConn) connect(ctx context.Context) error {
 	}
 	ws, _, err := dialer.DialContext(ctx, conn.cfg.URL, nil)
 	if err != nil {
-		var e x509.UnknownAuthorityError
-		if errors.As(err, &e) || strings.Contains(err.Error(), "certificate is not standards compliant") {
+		var e x509.HostnameError // No need to retry...
+		if IsErrorInvalidCert(err) || errors.Is(err, e) {
 			conn.setConnectionStatus(InvalidCert)
 			if conn.tlsCfg == nil {
 				return ErrCertRequired
--- a/client/comms/wsconn.go
+++ b/client/comms/wsconn.go
@@ -70,13 +70,12 @@ func (cs ConnectionStatus) String() string {
 var invalidCertRegexp = regexp.MustCompile(".*(unknown authority|not standards compliant|not trusted)")
 
 // isErrorInvalidCert checks if the provided error is one of the different
-// variant of an invalid cert error returned from the x509 package or is
-// ErrInvalidCert.
+// variant of an invalid cert error returned from the x509 package.
 func isErrorInvalidCert(err error) bool {
 	var invalidCertErr x509.CertificateInvalidError
 	var unknownCertAuthErr x509.UnknownAuthorityError
 	var hostNameErr x509.HostnameError
-	return errors.Is(err, ErrInvalidCert) || errors.As(err, &invalidCertErr) || errors.As(err, &hostNameErr) ||
+	return errors.As(err, &invalidCertErr) || errors.As(err, &hostNameErr) ||
 		errors.As(err, &unknownCertAuthErr) || invalidCertRegexp.MatchString(err.Error())
 }
 
@@ -231,9 +230,9 @@ func (conn *wsConn) connect(ctx context.Context) error {
 		if isErrorInvalidCert(err) {
 			conn.setConnectionStatus(InvalidCert)
 			if conn.tlsCfg == nil {
-				return ErrCertRequired
+				return dex.NewError(ErrCertRequired, err.Error())
 			}
-			return ErrInvalidCert
+			return dex.NewError(ErrInvalidCert, err.Error())
 		}
 		conn.setConnectionStatus(Disconnected)
 		return err
--- a/client/comms/wsconn.go
+++ b/client/comms/wsconn.go
@@ -69,14 +69,15 @@ func (cs ConnectionStatus) String() string {
 // caused by or related to an invalid cert.
 var invalidCertRegexp = regexp.MustCompile(".*(unknown authority|not standards compliant|not trusted)")
 
-// IsErrorInvalidCert checks if the provided error is one of the different
+// isErrorInvalidCert checks if the provided error is one of the different
 // variant of an invalid cert error returned from the x509 package or is
 // ErrInvalidCert.
-func IsErrorInvalidCert(err error) bool {
-	var invalidCert x509.CertificateInvalidError
-	var unknownCertAuth x509.UnknownAuthorityError
-	return errors.Is(err, ErrInvalidCert) || errors.Is(err, invalidCert) ||
-		errors.Is(err, unknownCertAuth) || invalidCertRegexp.MatchString(err.Error())
+func isErrorInvalidCert(err error) bool {
+	var invalidCertErr x509.CertificateInvalidError
+	var unknownCertAuthErr x509.UnknownAuthorityError
+	var hostNameErr x509.HostnameError
+	return errors.Is(err, ErrInvalidCert) || errors.As(err, &invalidCertErr) || errors.As(err, &hostNameErr) ||
+		errors.As(err, &unknownCertAuthErr) || invalidCertRegexp.MatchString(err.Error())
 }
 
 // ErrInvalidCert is the error returned when attempting to use an invalid cert
@@ -227,8 +228,7 @@ func (conn *wsConn) connect(ctx context.Context) error {
 	}
 	ws, _, err := dialer.DialContext(ctx, conn.cfg.URL, nil)
 	if err != nil {
-		var e x509.HostnameError // No need to retry...
-		if IsErrorInvalidCert(err) || errors.Is(err, e) {
+		if isErrorInvalidCert(err) {
 			conn.setConnectionStatus(InvalidCert)
 			if conn.tlsCfg == nil {
 				return ErrCertRequired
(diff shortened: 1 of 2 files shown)


chappjc:
What precise go version?
https://github.com/golang/go/issues/52010#issuecomment-1418743370

chappjc:
See also https://go-review.googlesource.com/c/go/+/460896/2/src/crypto/x509/root_darwin.go#61

ukane-philemon:
> What precise go version? [golang/go#52010 (comment)](https://github.com/golang/go/issues/52010#issuecomment-1418743370)

I saw the issue with `macOS Catalina 10.15.7`

ukane-philemon:
> See also https://go-review.googlesource.com/c/go/+/460896/2/src/crypto/x509/root_darwin.go#61 So additional error types, plus https://github.com/kubernetes/kubernetes/pull/109102/files#diff-4a1205c4eb7e8f2b8cf6feb57eba6df5472e88fc58302e7c89c01924231243a1R53 (more strings)

`case macOS.ErrSecNotTrusted` does not seem to catch this error. 

What do you think?

chappjc:
> What do you think? Do you suggest we do something like this instead?

We should do a combination of things.:
1. The three different substrings from that kubes PR
2. The other `x509` errors: `CertificateInvalidError`, `HostnameError`, and `UnknownAuthorityError`... probably more from https://pkg.go.dev/crypto/x509
3. figure out what about the certificate that macOS doesn't like
4. see if there's a reason the Mac certificate resolver is involved instead of the Go resolver

chappjc:
> 4\. see if there's a reason the Mac certificate resolver is involved instead of the Go resolver

What I mean about this is, why are we even bothering with `SystemCertPool` in this:

https://github.com/decred/dcrdex/blob/c6a4c340d1d94a47e30ce518b476f27b399a6c37/client/comms/wsconn.go#L161-L171

I suspect all would work fine for you if we juts used `rootCAs = x509.NewCertPool()` always.

ukane-philemon:
> see if there's a reason the Mac certificate resolver is involved instead of the Go resolver

I think this is the reason: https://github.com/golang/go/blob/master/src/crypto/x509/verify.go#L767-#L784

`SystemCertPool` returns the `*CertPool` type with `systemPool` set to true which will force two verifications("one using the roots provided by the caller, and one using the system platform verifier.") according to the docs. However, since we are not even providing a cert(in the test that failed for me), Go will always attempt to use platform verifiers.

While `x509.NewCertPool()` could force Go to use its verifier, we currently don't set the `*tlsConfig` values used by a connection if a cert is not provided. Do we want to change this? i.e set the tlsConfig values regardless of whether a cert was provided. Only the `RootCAs` value would be modified in the `if` block when/if a cert is provided.

chappjc:
>  we currently don't set the `*tlsConfig` values used by a connection if a cert is not provided. Do we want to change this? i.e set the tlsConfig values regardless of whether a cert was provided

```
// RootCAs defines the set of root certificate authorities
// that clients use when verifying server certificates.
// If RootCAs is nil, TLS uses the host's root CA set.
RootCAs *[x509](https://pkg.go.dev/crypto/x509).[CertPool](https://pkg.go.dev/crypto/x509#CertPool)
```

`nil` is correct to use the hosts root CA set if you specify no certificate, meaning perhaps it's not self-signed.

Anyway, can you link to the test that uses no cert and what it's expected to do?

ukane-philemon:
> Anyway, can you link to the test that uses no cert and what it's expected to do?

We are currently unable to catch extra `invalid cert` errors but we depend on our `ErrInvalidCert` for some actions like stopping a conn retry loop. 
https://github.com/decred/dcrdex/blob/master/client/comms/wsconn_test.go#L258

chappjc:
Right, I just wanted to see why we're giving a nil cert.  In this case what "test no cert error" really means is see if the system cert store has it.  So this test really depends on the system.

We still need to do the fixes in https://github.com/decred/dcrdex/pull/2130#issuecomment-1431523812.

ukane-philemon:
Does `HostnameError` really fall within `InvalidCert` errors? I don't think so.

chappjc:
> Does `HostnameError` really fall within `InvalidCert` errors? I don't think so.

Yup:

> HostnameError results when the set of authorized names doesn't match the requested name. 

Regardless, it's an error from x509 and retrying isn't going to help.

ukane-philemon:
Okay.

client/comms/wsconn.go
```diff
@@ -64,6 +65,20 @@ func (cs ConnectionStatus) String() string {
...
+var invalidCertRegexp = regexp.MustCompile(".*(unknown authority|not standards compliant|not trusted)")
+
+// IsErrorInvalidCert checks if the provided error is one of the different
+// variant of an invalid cert error returned from the x509 package or is
+// ErrInvalidCert.
+func IsErrorInvalidCert(err error) bool {
+	var invalidCert x509.CertificateInvalidError
+	var unknownCertAuth x509.UnknownAuthorityError
+	return errors.Is(err, ErrInvalidCert) || errors.Is(err, invalidCert) ||
+		errors.Is(err, unknownCertAuth) || invalidCertRegexp.MatchString(err.Error())
```
chappjc, line 79: This is what `errors.As` is for.  You don't need to create a dummy instance, and I'm not sure that would even work because you'd expect `err` to be an instance created elsewhere.

client/comms/wsconn.go
```diff
@@ -212,8 +227,8 @@ func (conn *wsConn) connect(ctx context.Context) error {
 	}
 	ws, _, err := dialer.DialContext(ctx, conn.cfg.URL, nil)
 	if err != nil {
-		var e x509.UnknownAuthorityError
-		if errors.As(err, &e) {
+		var e x509.HostnameError // No need to retry...
+		if IsErrorInvalidCert(err) || errors.As(err, &e) {
```
chappjc, line 231: Why isn't `var e x509.HostnameError` part of `IsErrorInvalidCert`?  It's not actually saying there's a problem with the host name but that it doesn't match the cert.  Hence why it is a cert error defined by the `x509` package.

chappjc: e.g. https://github.com/nats-io/nats-server/pull/1264

norwnd:
On a related note, I've just observed this issue happening for the first time today on my Mac too:

```
2023-02-16 09:33:41.488 [INF] CORE: Connecting wallet for eth
2023-02-16 09:33:42.606 [DBG] CORE[eth][ETH][RPC]: couldn't get a websocket connection for "wss://goerli.infura.io/ws/v3/e068872d3aa044aebbcf96f8e31104cc" (original scheme: "https") (OK)
2023-02-16 09:33:43.784 [ERR] CORE[eth][ETH][RPC]: Failed to get chain ID from "https://goerli.infura.io/v3/e068872d3aa044aebbcf96f8e31104cc": Post "https://goerli.infura.io/v3/e068872d3aa044aebbcf96f8e31104cc": x509: “*.infura.io” certificate is not standards compliant
2023-02-16 09:33:43.784 [ERR] WEB: error placing order: eth connectAndUnlock error: connectWallet: failed to connect eth wallet: connect failure: failed to connect
```

It seems to have disappeared [after restarting Mac](https://github.com/golang/go/issues/51991#issuecomment-1431990039), but **not right after** restart (I tried, it was still there) - more like 1h after. Looks more like Mac-specific (or maybe golang-specific) issue anyway, nothing to do here, just leaving it for anyone hitting in the future.

I'm running `dexc` with `go version go1.19.2 darwin/amd64` on `Monterey 12.6 (21G115)`.

chappjc:
> On a related note, I've just observed this issue happening for the first time today on my Mac too:
> 
> ```
> 2023-02-16 09:33:41.488 [INF] CORE: Connecting wallet for eth
> 2023-02-16 09:33:42.606 [DBG] CORE[eth][ETH][RPC]: couldn't get a websocket connection for "wss://goerli.infura.io/ws/v3/e068872d3aa044aebbcf96f8e31104cc" (original scheme: "https") (OK)
> 2023-02-16 09:33:43.784 [ERR] CORE[eth][ETH][RPC]: Failed to get chain ID from "https://goerli.infura.io/v3/e068872d3aa044aebbcf96f8e31104cc": Post "https://goerli.infura.io/v3/e068872d3aa044aebbcf96f8e31104cc": x509: “*.infura.io” certificate is not standards compliant
> 2023-02-16 09:33:43.784 [ERR] WEB: error placing order: eth connectAndUnlock error: connectWallet: failed to connect eth wallet: connect failure: failed to connect
> ```
> 
> It seems to have disappeared [after restarting Mac](https://github.com/golang/go/issues/51991#issuecomment-1431990039), but **not right after** restart (I tried, it was still there) - more like 1h after. Looks more like Mac-specific (or maybe golang-specific) issue anyway, nothing to do here, just leaving it for anyone hitting in the future.
> 
> I'm running `dexc` with `go version go1.19.2 darwin/amd64` on `Monterey 12.6 (21G115)`.

Wow, so it affects host names with certificates signed by trusted root CAs.  That's messed up.  Literally nothing we can do about that, esp. within the eth `rpcclient`.  

I suggest updating your Go version though.

client/comms/wsconn.go
ukane-philemon: updated `IsErrorInvalidCert` in my latest push.

client/comms/wsconn.go
```diff
@@ -64,6 +65,21 @@ func (cs ConnectionStatus) String() string {
...
+var invalidCertRegexp = regexp.MustCompile(".*(unknown authority|not standards compliant|not trusted)")
+
+// isErrorInvalidCert checks if the provided error is one of the different
+// variant of an invalid cert error returned from the x509 package or is
+// ErrInvalidCert.
+func isErrorInvalidCert(err error) bool {
+	var invalidCertErr x509.CertificateInvalidError
+	var unknownCertAuthErr x509.UnknownAuthorityError
+	var hostNameErr x509.HostnameError
+	return errors.Is(err, ErrInvalidCert) || errors.As(err, &invalidCertErr) || errors.As(err, &hostNameErr) ||
```
chappjc, line 79: OK to include `ErrInvalidCert` here for completeness I suppose, but this  helpers purpose is to detect when the caller should return `ErrInvalidCert`.  It won't ever get `ErrInvalidCert` itself, or am I missing something?

chappjc:
Will you also please try this on your end?

```diff
diff --git a/client/comms/wsconn.go b/client/comms/wsconn.go
index fb1a8123e..b118873a0 100644
--- a/client/comms/wsconn.go
+++ b/client/comms/wsconn.go
@@ -174,11 +174,7 @@ func NewWsConn(cfg *WsCfg) (WsConn, error) {
                        return nil, fmt.Errorf("error parsing URL: %w", err)
                }
 
-               rootCAs, _ := x509.SystemCertPool()
-               if rootCAs == nil {
-                       rootCAs = x509.NewCertPool()
-               }
-
+               rootCAs := x509.NewCertPool()
                if ok := rootCAs.AppendCertsFromPEM(cfg.Cert); !ok {
                        return nil, ErrInvalidCert
                }
```

If the user is specifying a certificate to use, I don't see a point in appending it to the SystemCertPool.
I'm hesitant to actually make that change though, but I am curious if it changes anything for you with Mac

client/comms/wsconn.go
ukane-philemon: Oh, I had intend to use it in the tests too.

chappjc:
> I will check. But I don't think it would make any difference. I haven't have any issues with the system verifier when a cert is provided.

I wouldn't expect it to affect the nil cert test case.

ukane-philemon:
> but I am curious if it changes anything for you with Mac

Well, since it does not change anything for mac(the invalid cert errors have been handled in this PR already). I think it's best we continue to use `SystemCertPool` as default and  `x509.NewCertPool()` as a fallback as we've been doing. 

Unless there's a reason a particular reason we shouldn't.

chappjc:
Labeling for 0.6 since it's a show stopper for a subset of mac users.

client/comms/wsconn.go
chappjc: I don't see it being useful in the tests.  The test are testing the exported `(*wsConn).Connect` method, ensuring that they return the *exported* error types (`ErrInvalidCert` and `ErrCertRequired`) as expected.

chappjc: The error messages from connect can also be supplemented with the verbose error message while maintaining the wrapped type:

```diff
diff --git a/client/comms/wsconn.go b/client/comms/wsconn.go
index fb1a8123e..43d3d599a 100644
--- a/client/comms/wsconn.go
+++ b/client/comms/wsconn.go
@@ -231,9 +231,9 @@ func (conn *wsConn) connect(ctx context.Context) error {
                if isErrorInvalidCert(err) {
                        conn.setConnectionStatus(InvalidCert)
                        if conn.tlsCfg == nil {
-                               return ErrCertRequired
+                               return dex.NewError(ErrCertRequired, err.Error())
                        }
-                       return ErrInvalidCert
+                       return dex.NewError(ErrInvalidCert, err.Error())
                }
                conn.setConnectionStatus(Disconnected)
                return err
```

Otherwise with this PR's change, you'd never see the "not standards compliant" text even though we would catch it as a cert error and not retry.

ukane-philemon: Yeah, the extra error detail would be useful for maybe future debugging.

chappjc (approved):

PR #2131, opened 2023-02-15
client/core: improve `TestRegister`

### LINKED ISSUE #2128: TestRegister occasionally fails
--- FAIL: TestRegister (1.35s)
    core_test.go:2256: timed out waiting for feepayment or balance notification

ukane-philemon:
This improves `TestRegister` to use a `time.Ticker(10 * time.Microsecond)` instead of `time.After(10 * time.Millisecond)` when waiting on notifications and to free timer resources when done with tests. Notice the change in `tick duration` reduced to `10 Microseconds`.  While I am unable to reproduce the rare error described by #2128 this an attempt to resolve #2128.

### DIFF
--- a/client/core/core_test.go
+++ b/client/core/core_test.go
@@ -1943,7 +1943,7 @@ func TestRegister(t *testing.T) {
 			defer wg.Done()
 			timeout := time.NewTimer(time.Second * 2)
 			defer timeout.Stop()
-			ticker := time.NewTicker(10 * time.Microsecond)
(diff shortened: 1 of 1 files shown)


ukane-philemon:
>I would expect that it needs a longer timeout, not more frequent checks.

I was reluctant in increasing the timeout since the  fail is rare and we don't know why the current timeout was not enough. But your comment makes more sense. I was hoping a more frequent check ensure we move to the next thing in the test since the notes are sent almost instantly.

Issue #2080, opened 2023-01-28
ui: registration, better messaging when almost synced


norwnd:
On `/register` page, when progress gets to 100% (it can stay like this for minutes, in my experience) I can see 99.9% on `/wallets` page (which is more accurate, because showing it as 100% might screw with user expectations about his balance - see more on this below, so it should probably be 99.9% on `/register` page also):

<img width="250" alt="image" src="https://user-images.githubusercontent.com/112318969/215254332-1a940311-51a1-43ad-98fd-1c18e26265e8.png"><img width="250" alt="image" src="https://user-images.githubusercontent.com/112318969/215254354-1b381641-c2cb-4ee8-b81a-367f7910d80a.png">

**Additionally**, note that it shows `> 1 day remaining` estimate even with 100% sync progress. This is probably inaccurate even if sync progress is changed to display 99.9%, and might be a **separate issue entirely**.

The address (shown on registration form) has been funded ~14 mins ago (so it should probably show balance of 40 DCR, not 0 DCR if wallet indeed is 100% synced):
<img width="250" alt="image" src="https://user-images.githubusercontent.com/112318969/215254191-1e28259f-8fc3-49ea-8a77-520dac4056e3.png"><img width="250" alt="image" src="https://user-images.githubusercontent.com/112318969/215254143-0c174e4c-4869-43f5-aad5-332ad36996d8.png">

Not sure whether it is a dcr/spv specific issue, or more of a UI display problem.

chappjc:
> On `/register` page, when progress gets to 100% (it can stay like this for minutes, in my experience) I can see 99.9% on `/wallets` page (which is more accurate, because showing it as 100% might screw with user expectations about his balance - see more on this below, so it should probably be 99.9% on `/register` page also):

Yeah this is just UI rounding it I think.  We've ensured 99.9% and not 100% if the synced flag is false:

https://github.com/itswisdomagain/dcrdex/blob/73816f5d4ed5d20374a462e0e89f9b0af23b2bfd/client/asset/dcr/spv.go#L784-L785

If you follow the issues that are linked with that you'll see we have limited visibility to what dcrwallet (or btcwallet) is doing in this phase, and there's literally no way to discern how long it's going to need to get done.

chappjc:
> The address (shown on registration form) has been funded ~14 mins ago (so it should probably show balance of 40 DCR, not 0 DCR if wallet indeed is 100% synced):

[...]

PR #2133, opened 2023-02-15
ui: wallet sync progress improvements (register page)

norwnd:
Resolves https://github.com/decred/dcrdex/issues/2080.

This should address boths parts of https://github.com/decred/dcrdex/issues/2080:
- progress stuck at 100% for more than one would expect (while wallet is still undergoing some final stages of syncing; this also makes sync progress behavior  consistent with 99.9% shown on `/wallets` page)

current behavior seems to result from rounding 99.9% to 100% (currently on **master**), while `.toFixed(1)` will not round and just trim all the extra digits instead, which is probably what we want here too,

- `> 1 day remaining` message isn't very accurate

I've reimplemented some edge-case handling to address that.

### DIFF
--- a/client/webserver/site/src/html/forms.tmpl
+++ b/client/webserver/site/src/html/forms.tmpl
@@ -508,7 +508,8 @@
     <span class="fs16 grey">%</span>
   </div>
   <div class="mt-1 flex-center flex-row fs14 d-hide" data-tmpl="syncRemainBox">
-    <span data-tmpl="syncRemain" class="me-1"></span> [[[remaining]]]
+    <span data-tmpl="syncRemaining"><span data-tmpl="syncRemain" class="me-1"></span> [[[remaining]]]</span>
+    <span data-tmpl="syncFinishingUp" class="me-1"></span>
   </div>
 </div>
 
--- a/client/webserver/site/src/js/forms.ts
+++ b/client/webserver/site/src/js/forms.ts
@@ -213,7 +213,7 @@ export class NewWalletForm {
       // Not synced, so create a syncer to update the parent sync pane.
       this.parentSyncer = (w: WalletState) => {
         if (w.assetID !== parentAsset.id) return
-        page.parentSyncPct.textContent = String(Math.round(w.syncProgress * 100))
+        page.parentSyncPct.textContent = (w.syncProgress * 100).toFixed(1)
       }
       // Handle the async result.
       this.createUpdater = (note: WalletCreationNote) => {
@@ -1058,7 +1058,7 @@ export class WalletWaitForm {
     Doc.show(wallet.synced ? page.syncCheck : wallet.syncProgress >= 1 ? page.syncSpinner : page.syncUncheck)
     Doc.show(wallet.balance.available > fee.amount ? page.balCheck : page.balUncheck)
 
-    page.progress.textContent = String(Math.round(wallet.syncProgress * 100))
+    page.progress.textContent = (wallet.syncProgress * 100).toFixed(1)
 
     if (wallet.synced) {
       this.progressed = true
@@ -1124,26 +1124,43 @@ export class WalletWaitForm {
       Doc.hide(page.syncSpinner)
       Doc.show(page.syncUncheck)
     }
-    page.progress.textContent = String(Math.round(prog * 100))
+    page.progress.textContent = (prog * 100).toFixed(1)
 
-    // The remaining time estimate must be based on more than one progress
-    // report. We'll cache up to the last 20 and look at the difference between
-    // the first and last to make the estimate.
+    if (prog >= 0.999) {
+      Doc.hide(page.syncRemaining)
+      Doc.show(page.syncFinishingUp)
+      Doc.show(page.syncRemainBox)
+      // The final stage of wallet sync process can take a while (it might hang
+      // at 99.9% for many minutes, indexing addresses for example), the simplest
+      // way to handle it is to keep displaying "finishing up" message until the
+      // sync is finished, since we can't reasonably show it progressing over time.
+      page.syncFinishingUp.textContent = intl.prep(intl.ID_WALLET_SYNC_FINISHING_UP)
+      return
+    }
+    // Before we get to 99.9% the remaining time estimate must be based on more
+    // than one progress report. We'll cache up to the last 20 and look at the
+    // difference between the first and last to make the estimate.
     const cacheSize = 20
     const cache = this.progressCache
     cache.push({
       stamp: new Date().getTime(),
       progress: prog
     })
+    if (cache.length < 2) {
+      // Can't meaningfully estimate remaining until we have at least 2 data points.
+      return
+    }
     while (cache.length > cacheSize) cache.shift()
-    if (cache.length === 1) return
-    Doc.show(page.syncRemainBox)
     const [first, last] = [cache[0], cache[cache.length - 1]]
     const progDelta = last.progress - first.progress
     if (progDelta === 0) {
-      page.syncRemain.textContent = '> 1 day'
+      // Having no progress for a while likely means we are experiencing network
+      // issues, can't reasonably estimate time remaining in this case.
       return
     }
+    Doc.hide(page.syncFinishingUp)
+    Doc.show(page.syncRemaining)
+    Doc.show(page.syncRemainBox)
     const timeDelta = last.stamp - first.stamp
     const progRate = progDelta / timeDelta
     const toGoProg = 1 - last.progress
--- a/client/webserver/site/src/html/bodybuilder.tmpl
+++ b/client/webserver/site/src/html/bodybuilder.tmpl
@@ -104,7 +104,7 @@
 {{end}}
 
 {{define "bottom"}}
-<script src="/js/entry.js?v=1yeumsqty"></script>
+<script src="/js/entry.js?v=N2HhoiFV"></script>
 </body>
 </html>
 {{end}}
--- a/client/webserver/site/src/js/locales.ts
+++ b/client/webserver/site/src/js/locales.ts
@@ -107,6 +107,7 @@ export const ID_MAKER = 'MAKER'
 export const ID_EMPTY_DEX_ADDRESS_MSG = 'EMPTY_DEX_ADDRESS_MSG'
 export const ID_SELECT_WALLET_FOR_FEE_PAYMENT = 'SELECT_WALLET_FOR_FEE_PAYMENT'
 export const ID_UNAVAILABLE = 'UNAVAILABLE'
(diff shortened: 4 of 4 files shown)


client/webserver/site/src/js/forms.ts
```diff
@@ -1124,26 +1124,39 @@ export class WalletWaitForm {
...
+
+    if (prog >= 0.999) {
+      Doc.show(page.syncRemainBox)
+      // The final stage of wallet sync process can take a while (it might hang
+      // at 99.9% for many minutes, indexing addresses for example), the simplest
+      // way to handle it is to keep displaying constant rough estimate until
+      // sync is finished.
+      page.syncRemain.textContent = '~1 m'
+      return
+    }
```
norwnd, line 1137: Might also create a separate message for **>= 99.9% case** (instead of displaying `~1 m remaining` like I'm currently doing here), but that would require more code/translations, seems like not worth the effort ?

Or maybe just not show this estimate altogether for **>= 99.9% case** since it won't always be accurate ?

client/webserver/site/src/js/forms.ts
```diff
@@ -1058,7 +1058,7 @@ export class WalletWaitForm {
     Doc.show(wallet.synced ? page.syncCheck : wallet.syncProgress >= 1 ? page.syncSpinner : page.syncUncheck)
     Doc.show(wallet.balance.available > fee.amount ? page.balCheck : page.balUncheck)
 
-    page.progress.textContent = String(Math.round(wallet.syncProgress * 100))
+    page.progress.textContent = String((wallet.syncProgress * 100).toFixed(1))
```
chappjc, line 1061: Nice.  Is there any remaining purpose of the `String` *object* constructor here instead of the string primitive that we get from `toFixed`?

client/webserver/site/src/js/forms.ts
chappjc: Tough call.  "~1 m" isn't terrible, and it's probably ok not to require translation.  Showing nothing isn't bad either IMO.  Something like "finishing up" could be best, but it's really not that necessary and its more to translate. So, meh

client/webserver/site/src/js/forms.ts
```diff
@@ -1124,26 +1124,39 @@ export class WalletWaitForm {
...
+      // Can't meaningfully estimate remaining until we have at least 2 data points.
+      return
+    }
     while (cache.length > cacheSize) cache.shift()
-    if (cache.length === 1) return
-    Doc.show(page.syncRemainBox)
     const [first, last] = [cache[0], cache[cache.length - 1]]
     const progDelta = last.progress - first.progress
     if (progDelta === 0) {
-      page.syncRemain.textContent = '> 1 day'
```
chappjc, line 1144: lol :+1:

client/webserver/site/src/js/forms.ts
ukane-philemon: IMO, if with will help with passing a more correct message to users, another translatable wording won't hurt. If it hangs there for a much longer time ~1 m is still confusing. It is 1 minute, a month? 🤔☺️ And even if users can tell what 1 m is but we are not done in that time frame, it could become a thing of concern.

norwnd:
Just removed some debugging stuff there with force-push.

client/webserver/site/src/js/forms.ts
norwnd: Changed to "finishing up":
image

client/webserver/site/src/js/forms.ts
norwnd: Not that I'm aware of, I just didn't notice `toFixed` returned **string**, removed this conversion.

client/webserver/site/src/html/forms.tmpl
```diff
@@ -485,7 +485,8 @@
     <span class="fs16 grey">%</span>
   </div>
   <div class="mt-1 flex-center flex-row fs14 d-hide" data-tmpl="syncRemainBox">
-    <span data-tmpl="syncRemain" class="me-1"></span> [[[remaining]]]
+    <span data-tmpl="syncRemaining"><span data-tmpl="syncRemain" class="me-1"></span> [[[remaining]]]</span>
+    <span data-tmpl="syncFinishingUp" class="me-1"></span>
```
chappjc, line 489: Reasons not to do `[[[finishing_up]]]` and instead setting `syncFinishingUp.textContent` in JS?  There's nothing dynamic about it and we show/hide the entire `span`?

client/webserver/site/src/js/forms.ts
```diff
@@ -1124,26 +1124,43 @@ export class WalletWaitForm {
...
+    // difference between the first and last to make the estimate.
     const cacheSize = 20
     const cache = this.progressCache
     cache.push({
       stamp: new Date().getTime(),
       progress: prog
     })
+    if (cache.length < 2) {
+      // Can't meaningfully estimate remaining until we have at least 2 data points.
+      return
```
chappjc, line 1151: But we can still show progress right? Just not rate/remaining.  It could be at say 35% after a one or two updates.

chappjc: Oh, NVM read that wrong

client/webserver/site/src/html/forms.tmpl
norwnd: > we show/hide the entire span?

yes, we hide one of **syncRemaining**/**syncFinishingUp** and show the other,

yeah, seems like could be easily moved to go html templates, done (and re-tested, working same way).

norwnd:
Also tested with BTC, all is same except it doesn't seem to be hanging at 99.9% for nearly as long as DCR wallet (and currently does not, but might in the future, require [similar trick we do for DCR wallet](https://github.com/itswisdomagain/dcrdex/blob/73816f5d4ed5d20374a462e0e89f9b0af23b2bfd/client/asset/dcr/spv.go#L784-L785)).

chappjc (approved):
utACK.  Will try to test today.

client/webserver/locales/en-us.go
```diff
@@ -321,4 +321,5 @@ var EnUS = map[string]string{
 	"connected":              "Connected",
 	"Remove":                 "Remove",
 	"unready_wallets_msg":    "Your wallets must be connected and unlocked before trades can be processed. Resolve this ASAP!",
+	"finishing_up":           "finishing up",
```
buck54321, line 324: How about "finishing sync"

client/webserver/site/src/js/forms.ts
```diff
@@ -1058,7 +1058,7 @@ export class WalletWaitForm {
     Doc.show(wallet.synced ? page.syncCheck : wallet.syncProgress >= 1 ? page.syncSpinner : page.syncUncheck)
     Doc.show(wallet.balance.available > fee.amount ? page.balCheck : page.balUncheck)
 
-    page.progress.textContent = String(Math.round(wallet.syncProgress * 100))
+    page.progress.textContent = (wallet.syncProgress * 100).toFixed(1)
```
buck54321, line 1061: I honestly don't think the third sig fig does anything for us or the users. How about changing `Math.round` to `Math.floor`. In combination with your `if (prog >= 0.999)` block, we should get the desired behavior.

client/webserver/site/src/js/forms.ts
```diff
@@ -1124,26 +1124,42 @@ export class WalletWaitForm {
...
     cache.push({
       stamp: new Date().getTime(),
       progress: prog
     })
+    if (cache.length < 2) {
+      // Can't meaningfully estimate remaining until we have at least 2 data points.
+      return
+    }
     while (cache.length > cacheSize) cache.shift()
-    if (cache.length === 1) return
```
buck54321, line 1139: This change is curious. What's the purpose?

client/webserver/site/src/js/forms.ts
```diff
@@ -1124,26 +1124,42 @@ export class WalletWaitForm {
...
-    Doc.show(page.syncRemainBox)
     const [first, last] = [cache[0], cache[cache.length - 1]]
     const progDelta = last.progress - first.progress
     if (progDelta === 0) {
-      page.syncRemain.textContent = '> 1 day'
```
buck54321, line 1144: So the `syncRemaining` field will just say "remaining"?

buck54321 (changes requested):

client/webserver/site/src/js/forms.ts
chappjc: On mainnet and test, it sometimes slows to a crawl for whatever reason, like one crap peer slowing things down, and 45.2 changing to 45.7 gives you a sign of life.
Maybe most users just want a whole number?  I feel like one decimal is fine personally, but whatever you guys decided is fine with me.

norwnd: I personally like 3rd figure because of the sense of progress **chap** mentioned (I think it is noticeable on slow network), plus it is already implemented and working (according to my testing), so I wouldn't wanna change it again unless you insist ;)

client/webserver/site/src/js/forms.ts
norwnd: It's equivalent, from what I understand, but
- failing fast (before `while (cache.length > cacheSize) cache.shift()`)
- and having a doc-comment for what exactly the intent is

makes it more readable/explicit for me.

client/webserver/site/src/js/forms.ts
norwnd: We would if I were to leave `Doc.show(page.syncRemainBox)` where it was on **master**, I moved it below (so that `page.syncRemainBox` is shown only when we are about to update anything on it, and not otherwise) - again to make the behavior more explicit and straightforward, so we don't have surprises like this one (at the expense of few more lines of code, though, I believe it's more often a better trade-off).

client/webserver/locales/en-us.go
norwnd: Updated

norwnd:
Since we changed some JS code in this PR I'm guessing JS cache busters need an update too (added a [commit on top](https://github.com/decred/dcrdex/pull/2133/commits/e33278903690802c24df69aa9db01b637d73f620) for that) ?

JoeGruffins (approved):

chappjc:
Looks like final nits here are resolved. bodybuilder.tmpl needs a bump, but I can do that if norwnd is unavailable at the moment.  Last revisions good for you @buck54321?

client/webserver/site/src/js/forms.ts
chappjc: The > 1 day thing didn't seem reasonable to me either.  There's almost no chance that that estimate is correct.  Even if it were, user would almost certainly intervene (shutdown and restart) to give it a kick.

buck54321 (approved):
Ugh. Cache busters are annoying.

PR #2138, opened 2023-02-16
client/eth: Only call tipChange and peersChange for connected wallets

### LINKED ISSUE #2137: balance error: contractor for 60001 <nil> is not a tokenContractor
I saw this shortly after startup, after first tip change, I think:

```
2023-02-16 17:37:36.507 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.blastapi.io" reported new tip at height 16642828 (0x4cd54bf01f547a77c4213c3e63ce7442d968c9a9d3f88f6c2e1d7ec12
fddb082)
2023-02-16 17:37:36.530 [TRC] CORE[eth][ETH][RPC]: "eth-mainnet.g.alchemy.com" reported new tip at height 16642828 (0x4cd54bf01f547a77c4213c3e63ce7442d968c9a9d3f88f6c2e1d7ec
12fddb082)
2023-02-16 17:37:37.456 [DBG] CORE[eth][ETH]: tip change: 16642827 (0x865b09a4afd8a179630a2c26cd83f7ccbb32464dccaa2cb37a862ce308b8b399) => 16642828 (0x4cd54bf01f547a77c4213c
3e63ce7442d968c9a9d3f88f6c2e1d7ec12fddb082)
2023-02-16 17:37:37.456 [TRC] CORE: Processing tip change for eth
2023-02-16 17:37:37.456 [TRC] CORE: Processing tip change for usdc.eth
2023-02-16 17:37:37.456 [ERR] CORE: error updating "usdc.eth" balance: pending balance error: balance error: contractor for 60001 <nil> is not a tokenContractor
```

Definitely build with `-tags lgpl` because I just swapped eth

martonp:
`TipChange` was being called for token wallets that were not yet connected. This caused core to call the token wallet to retrieve its balance, but because it was not yet connected, it would return an error.

Closes #2137.

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -434,6 +434,7 @@ type assetWallet struct {
 	tipChange  func(error)
 	log        dex.Logger
 	atomicUnit string
+	connected  atomic.Bool
 
 	lockedFunds struct {
 		mtx                sync.RWMutex
@@ -797,6 +798,8 @@ func (w *ETHWallet) Connect(ctx context.Context) (_ *sync.WaitGroup, err error) 
 	atomic.StoreInt64(&w.tipAtConnect, height.Int64())
 	w.log.Infof("Connected to geth, at height %d", height)
 
+	w.connected.Store(true)
+
 	var wg sync.WaitGroup
 	wg.Add(1)
 	go func() {
@@ -809,6 +812,12 @@ func (w *ETHWallet) Connect(ctx context.Context) (_ *sync.WaitGroup, err error) 
 		defer wg.Done()
 		w.monitorPeers(ctx)
 	}()
+
+	go func() {
+		<-ctx.Done()
+		w.connected.Store(false)
+	}()
+
 	return &wg, nil
 }
 
@@ -824,6 +833,8 @@ func (w *TokenWallet) Connect(ctx context.Context) (*sync.WaitGroup, error) {
 		return nil, err
 	}
 
+	w.connected.Store(true)
+
 	var wg sync.WaitGroup
 	wg.Add(1)
 	go func() {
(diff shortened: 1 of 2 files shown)


client/asset/eth/eth.go
```diff
@@ -434,6 +434,7 @@ type assetWallet struct {
 	tipChange  func(error)
 	log        dex.Logger
 	atomicUnit string
+	connected  atomic.Bool
```
chappjc, line 437: `atomic.Bool` woot!

chappjc:
On the topic of the `wallets` map, from what I can tell, the `TokenWallet[.assetWallet].wallets` map is never populated, so just a potential source of errors.  For instance, I just tried moving the wallets map and mtx to `ETHWallet` fields, but realized there are there are the `(*assetWallet).allowanceGasRequired` and `(*assetWallet).initGasEstimate` methods that do `w.wallet`.  So now I'm worried that those are bugged and returning 0 because they may be using the empty map.

martonp:
> On the topic of the `wallets` map, from what I can tell, the `TokenWallet[.assetWallet].wallets` map is never populated (only the parent one), so just a potential source of errors. For instance, I just tried moving the wallets map and mtx to `ETHWallet` fields, but realized there are there are the `(*assetWallet).allowanceGasRequired` and `(*assetWallet).initGasEstimate` methods that do `w.wallet(assetID)`. So now I'm worried that those are bugged and returning 0 because they may be using the empty map.

The wallets map is on the `baseWallet` type of which there is only one instance. It is referenced by all the `assetWallet`s.

chappjc:
> The wallets map is on the `baseWallet` type of which there is only one instance. It is referenced by all the `assetWallet`s.

```
type TokenWallet struct {
	*assetWallet

	cfg      *tokenWalletConfig
	parent   *assetWallet
```

Only the `parent` instance has entries in it's map.

chappjc:
Oh, apologies, I see in `(*ETHWallet) OpenTokenWallet`

```go
	aw := &assetWallet{
		baseWallet:         w.baseWallet,
```

chappjc (approved):
Been working well on mainnet, stacked on a bunch of other eth PRs

chappjc:
Will merge in within the hour based on my testing and review, unless there are comments from others.

buck54321 (approved):

Issue #2136, opened 2023-02-16
ETH PreSwap feeSuggestion does not seem to be using current base fee, but the maxFeeRate


chappjc:

chappjc:
For some reason I have `PreOrder` doing this: https://github.com/decred/dcrdex/blob/998ddd23602f7bf67c9fef5050cf6c0d44ccba26/client/core/core.go#L5221-L5226

That is, `c.feeSuggestion` instead of `c.feeSuggestionAny`, which would ask the wallet.  The comment says `server rates only for the swap init`, but this is an estimate and shouldn't it be realistic?

chappjc:
Although it gets trickier because ETH's `FeeRate()` method is based on the `baseRate * 2 + tip` formula used to set a recommended max fee rate.  However, if we're interested in an estimate of what the swap fees are likely to be, that's almost double.

chappjc:
Let's have at least `PreSwap` ignore the input and use base + tip it has internally.

PR #2139, opened 2023-02-16
client/asset/eth: use currentFeeRate for preSwap, not max

chappjc:
Resolves https://github.com/decred/dcrdex/issues/2136 <s>https://github.com/decred/dcrdex/issues/2137</s>

Best/worst fee rate estimates from PreSwap (to PreOrder on frontend) are now considering the current effective (likely) fee rate, not the server's MaxFeeRate or the "recommended" max fee when authoring txns that would be more than double current base fee.

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -1146,18 +1146,18 @@ func (w *assetWallet) balance() (*asset.Balance, error) {
 // associated with nfo.MaxFeeRate. For quote assets, the caller will have to
 // calculate lotSize based on a rate conversion from the base asset's lot size.
 func (w *ETHWallet) MaxOrder(ord *asset.MaxOrderForm) (*asset.SwapEstimate, error) {
-	return w.maxOrder(ord.LotSize, ord.FeeSuggestion, ord.MaxFeeRate, ord.AssetVersion,
+	return w.maxOrder(ord.LotSize, ord.MaxFeeRate, ord.AssetVersion,
 		ord.RedeemVersion, ord.RedeemAssetID, nil)
 }
 
 // MaxOrder generates information about the maximum order size and associated
 // fees that the wallet can support for the given DEX configuration.
 func (w *TokenWallet) MaxOrder(ord *asset.MaxOrderForm) (*asset.SwapEstimate, error) {
-	return w.maxOrder(ord.LotSize, ord.FeeSuggestion, ord.MaxFeeRate, ord.AssetVersion,
+	return w.maxOrder(ord.LotSize, ord.MaxFeeRate, ord.AssetVersion,
 		ord.RedeemVersion, ord.RedeemAssetID, w.parent)
 }
 
-func (w *assetWallet) maxOrder(lotSize uint64, feeSuggestion, maxFeeRate uint64, ver uint32,
+func (w *assetWallet) maxOrder(lotSize uint64, maxFeeRate uint64, ver uint32,
 	redeemVer, redeemAssetID uint32, feeWallet *assetWallet) (*asset.SwapEstimate, error) {
 	balance, err := w.Balance()
 	if err != nil {
@@ -1193,7 +1193,7 @@ func (w *assetWallet) maxOrder(lotSize uint64, feeSuggestion, maxFeeRate uint64,
 	if lots < 1 {
 		return &asset.SwapEstimate{}, nil
 	}
-	return w.estimateSwap(lots, lotSize, feeSuggestion, maxFeeRate, ver, redeemVer, redeemAssetID)
+	return w.estimateSwap(lots, lotSize, maxFeeRate, ver, redeemVer, redeemAssetID)
 }
 
 // PreSwap gets order estimates based on the available funds and the wallet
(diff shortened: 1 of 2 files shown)


chappjc:
Hmm, it's much much lower now, but still turns out that the tx uses something like half the estimate.

chappjc:
Oh derp that's because we're padding the heck out of the gas estimates to set safe limits on the txn.
Seems like we need estimateSwap to use the lower of the estimate or the hard coded gas... opposite of tx param setting.

buck54321 (commented):
Shouldn't we do this in `(*assetWallet).maxOrder` too? Can even remove the `feeSuggestion` argument there.

chappjc:
Good call.  I'm presently trying to decide what we should return if token approval hasn't done yet.  The issue is as I said in my comment above we want these estimates to be as accurate as possible, unlike the hard-coded gases that are set to be used in txOpts to prevent failed transactions.

buck54321:
> Oh derp that's because we're padding the heck out of the gas estimates to set safe limits on the txn.

Because we're using `defaultGasFeeLimit` to calculate `maxSwapsInTx`?

chappjc:
> > Oh derp that's because we're padding the heck out of the gas estimates to set safe limits on the txn.
> 
> Because we're using `defaultGasFeeLimit` to calculate `maxSwapsInTx`?

No I'm referring to how `swapGas` first does `w.gases` and then gets the (hopefully lower) live estimate.
The former is important to set conservative parameters in the transactions (and worst case scenarios), but for "reasonable" estimates that we're trying to report to the user, we wanna only use the live estimates.

buck54321:
> Good call. I'm presently trying to decide what we should return if token approval hasn't done yet. The issue is as I said in my comment above we want these estimates to be as accurate as possible, unlike the hard-coded gases that are set to be used in txOpts to prevent failed transactions.

Live estimates will never bring the realistic estimates **down**, only **up**. We can skip live estimates if there is no approval. It'll only happen for the first order. That unfortunately has to be calculated at the `TokenWallet` level, so new arguments will be needed in some `assetWallet` methods.

chappjc:
> Live estimates will never bring the realistic estimates **down**, only **up**. We can skip live estimates if there is no approval. It'll only happen for the first order. That unfortunately has to be calculated at the `TokenWallet` level, so new arguments will be needed in some `assetWallet` methods.

The point is they only bring it up with the current `swapGas` function, but that is exactly the opposite of what we want in the context.
The live estimates had better be lower than hard coded ones, because if they are higher, something bad has changed.

chappjc:
Broke the tests (again), but now:
- made fee rate request internal to `estimateSwap` (no input param)
- fixed "best case" fees to reflect single swap for multi lot order
- fall back on dexeth.Gases if token approval not done yet

Fee previews look accurate on mainnet now.

client/asset/eth/eth.go
```diff
@@ -1267,8 +1292,8 @@ func (w *assetWallet) estimateSwap(lots, lotSize, feeSuggestion uint64, maxFeeRa
 		Lots:               lots,
 		Value:              value,
 		MaxFees:            maxFees,
-		RealisticWorstCase: oneGasMax * feeSuggestion,
-		RealisticBestCase:  (nSwap + allowanceGas) * feeSuggestion,
+		RealisticWorstCase: oneGasMax * rate,
+		RealisticBestCase:  (oneSwap + allowanceGas) * rate, // not even batch, just perfect match
```
JoeGruffins, line 1296: I think this is fine.

nit: I was looking at this the other day and was thinking about saying something, but isn't using nSwap what we do for other coins? In dcr the Best case is really one output isn't it? https://github.com/decred/dcrdex/blob/f9531e8709c7a80130bb73dbd36bcfbc704ccabf/client/asset/dcr/dcr.go#L1213

chappjc: And one input.  We can certainly change the UI and code docs, but: https://github.com/decred/dcrdex/blob/f9531e8709c7a80130bb73dbd36bcfbc704ccabf/client/asset/estimation.go#L20-L22
That's "best case", even better than batch.

chappjc: UI: https://github.com/decred/dcrdex/blob/f9531e8709c7a80130bb73dbd36bcfbc704ccabf/client/webserver/locales/en-us.go#L209

Consumed in single match also means single output, no need for batch.

JoeGruffins: So is the dcr best case wrong or am I confused?

chappjc: Oh, yeah, I'm agreeing with you.  I guess we decided to consider batch because it was just there as an option, but it's hard to call it "best" when you can simply match with another order of the same number of lots.

JoeGruffins: Yeah I think the definition of best here, as you have changed eth to, is correct.

client/asset/eth/eth_test.go
```diff
@@ -2479,60 +2493,63 @@ func TestMaxOrder(t *testing.T) {
...
 	for _, test := range tests {
-		var assetID uint32 = BipID
-		dexAsset := tETH
-		if test.token {
-			assetID = simnetTokenID
-		}
-
-		w, _, node, shutdown := tassetWallet(assetID)
-		defer shutdown()
+		t.Run(test.name, func(t *testing.T) {
```
buck54321, line 2496: In `TestPreSwap` you did this thing `runTest := func(t *testing.T, test testData) {`. Can you do the same here?

client/asset/eth/eth.go
buck54321: Yeah, seems like the `estLowFunds += dexdcr.P2SHOutputSize * (lots - 1) * bumpedNetRate ` can be deleted in **dcr** and **btc**.

client/asset/eth/eth_test.go
chappjc: The two tests mirror eachother closely now with that `runTest` pattern

client/asset/eth/eth.go
chappjc: Done for dcr and btc in https://github.com/decred/dcrdex/pull/2147

JoeGruffins (approved):

buck54321 (approved):

Issue #1945, opened 2022-11-08
UI: Order status discrepancy between different views


norwnd:
Out of 3 different forms that show order status I can see:
- `executed` on 2 of them

<img width="350" alt="image" src="https://user-images.githubusercontent.com/112318969/200587784-9b453340-8a11-4eca-84c2-39535476b4ae.png">

<img width="350" alt="image" src="https://user-images.githubusercontent.com/112318969/200587902-e01d1208-47fa-4bfc-b189-a7aa56843c4e.png">

- `settling` on this one

<img width="350" alt="image" src="https://user-images.githubusercontent.com/112318969/200588090-50684388-29ad-4d9d-8eec-017d4c5ef059.png">

On current `master`, I created 2 limit orders that matched (trading with myself basically), and it seems corresponding ETH Maker Swap didn't get mined in time, DEX server revoked it:
```
2022-11-08 12:35:38.391 [ERR] CORE: notify: |ERROR| (order) Swap reporting error - Error notifying DEX of swap for match dbfb6b3b5f295319a890c6967ecb70e28b20b56baee9e3a636caa3c475591016: error sending 'init' message: rpc error: error code 24: failed to find contract coin 0x2ef2c3bab324b45da3a2be2fba46890f68ac91318250f5eddfe2e0b74118fbbf - Order: dd9386a5ed4cefadce933cfa4315285fb2d888f12b4d6ad5dfc44028b8457577

...

2022-11-08 12:39:15.848 [WRN] CORE: Match dbfb6b3b5f295319a890c6967ecb70e28b20b56baee9e3a636caa3c475591016 revoked in status MakerSwapCast for order dd9386a5ed4cefadce933cfa4315285fb2d888f12b4d6ad5dfc44028b8457577
```

(according to Etherscan txn is still pending several hours after; it's probably because 32 Gwei is much lower than current base-fee):

<img width="350" alt="image" src="https://user-images.githubusercontent.com/112318969/200590703-178b9056-7e2e-4b2a-a260-0fb30687463d.png">

I think `settling` status is the correct one, because this Maker Swap txn could still be mined, and if that happens `dexc` should probably perform refund afterwards.

--------------------------------------------------------------------------------

[...]

PR #2140, opened 2023-02-17
ui: fix order status discrepancy on different views

norwnd:
Resolves https://github.com/decred/dcrdex/issues/1945.

For the case described in the linked issue (and I haven't seen other cases) now  status is same (and correct one) for all navigated views:

[status-discrepancy.webm](https://user-images.githubusercontent.com/112318969/219619838-0382951b-f727-4ebb-837e-514555db75ca.webm)

### DIFF
--- a/client/core/helpers.go
+++ b/client/core/helpers.go
@@ -268,6 +268,9 @@ func (ord *OrderReader) hasActiveMatches() bool {
 }
 
 // StatusString is the order status.
+//
+// IMPORTANT: we have similar function in JS for UI, it must match this one exactly,
+// when updating make sure to update both!
 func (ord *OrderReader) StatusString() string {
 	isLive := ord.hasActiveMatches()
 	switch ord.Status {
--- a/client/core/helpers.go
+++ b/client/core/helpers.go
@@ -256,8 +256,9 @@ func (ord *OrderReader) sumTo(filter func(match *Match) bool) uint64 {
 	return v
 }
 
-// hasLiveMatches will be true if there are any matches < MakerRedeemed.
-func (ord *OrderReader) hasLiveMatches() bool {
+// hasActiveMatches will be true if order has any matches that still require some
+// actions to complete settlement.
+func (ord *OrderReader) hasActiveMatches() bool {
 	for _, match := range ord.Matches {
 		if match.Active {
 			return true
@@ -268,7 +269,7 @@ func (ord *OrderReader) hasLiveMatches() bool {
 
 // StatusString is the order status.
 func (ord *OrderReader) StatusString() string {
-	isLive := ord.hasLiveMatches()
+	isLive := ord.hasActiveMatches()
 	switch ord.Status {
 	case order.OrderStatusUnknown:
 		return "unknown"
--- a/client/webserver/site/src/js/app.ts
+++ b/client/webserver/site/src/js/app.ts
@@ -9,7 +9,7 @@ import OrdersPage from './orders'
 import OrderPage from './order'
 import DexSettingsPage from './dexsettings'
 import MarketMakerPage from './mm'
-import { RateEncodingFactor, StatusExecuted, hasLiveMatches } from './orderutil'
+import { RateEncodingFactor, StatusExecuted, hasActiveMatches } from './orderutil'
 import { getJSON, postJSON, Errors } from './http'
 import * as ntfn from './notifications'
 import ws from './ws'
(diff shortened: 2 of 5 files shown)


client/webserver/site/src/js/app.ts
```diff
@@ -807,31 +807,20 @@ export default class Application {
    * haveActiveOrders returns whether or not there are active orders involving a
    * certain asset.
    */
-  haveAssetOrders (assetID: number): boolean {
+  haveActiveOrders (assetID: number): boolean {
```
norwnd, line 810: Renaming since looks like a typo (comment actually called `haveActiveOrders` which makes more sense to me).

client/webserver/site/src/js/app.ts
```diff
@@ -807,31 +807,20 @@ export default class Application {
...
-  walletIsActive (assetID: number): boolean {
-    if (this.haveAssetOrders(assetID)) return true
-    for (const xc of Object.values(this.user.exchanges)) {
-      if (!xc) continue
-      if (xc.pendingFee && xc.pendingFee.assetID === assetID) {
-        return true
-      }
-    }
-    return false
-  }
```
norwnd, line 833: This **func** doesn't seem to be used, hence removed.

client/webserver/site/src/js/orderutil.ts
```diff
@@ -59,21 +59,21 @@ export function isMarketBuy (ord: Order) {
...
- * hasLiveMatches returns true if the order has matches that have not completed
+ * hasActiveMatches returns true if the order has matches that have not completed
  * settlement yet.
  */
-export function hasLiveMatches (order: Order) {
+export function hasActiveMatches (order: Order) {
   if (!order.matches) return false
   for (const match of order.matches) {
-    if (!match.revoked && match.status < MakerRedeemed) return true
+    if (match.active) return true
```
norwnd, line 68: The reason for status showing properly only on `/order/{id}` page on current **master** seems to have been the reliance on `!match.revoked` here in `function hasActiveMatches`. 

Best I can tell `function statusString` in JS must be same as `OrderReader.StatusString` in Golang (and it will be with the changes from this PR).

client/core/helpers.go
```diff
@@ -256,8 +256,9 @@ func (ord *OrderReader) sumTo(filter func(match *Match) bool) uint64 {
 	return v
 }
 
-// hasLiveMatches will be true if there are any matches < MakerRedeemed.
-func (ord *OrderReader) hasLiveMatches() bool {
+// hasActiveMatches will be true if order has any matches that still require some
+// actions to complete settlement.
+func (ord *OrderReader) hasActiveMatches() bool {
```
norwnd, line 261: Terms **live** and **active** seem to be used interchangeably, it isn't obvious though so I'd prefer to have single name for the same thing when possible, hence renaming to **active**.

client/webserver/site/src/js/app.ts
chappjc: I'd be surprised if we don't need this back but with a bonds check instead of the pending fee.  Can re-create though.

client/webserver/site/src/js/orderutil.ts
```diff
@@ -59,21 +59,26 @@ export function isMarketBuy (ord: Order) {
...
```
buck54321, line 68: Nice. I just fixed this in a local branch I'm working on too.

client/webserver/site/src/js/app.ts
buck54321: A mixed bag really. We've lost "asset" from the name, so less descriptive there, but we've added "active", so better on that part. I'm okay with the function signature providing some context for use, so this is fine, really. `haveActiveAssetOrders` would also be acceptable.

buck54321 (approved):
The live/active ambiguity has irked be before.

JoeGruffins (approved):

chappjc:
OK, nits can be resolved in subequent UI work if we want.

client/webserver/site/src/js/orderutil.ts
chappjc: https://github.com/decred/dcrdex/issues/2164

Need another solution for this.

chappjc: https://github.com/decred/dcrdex/pull/2140

Not clear about the root cause of that though.

PR #2143, opened 2023-02-17
client/asset/eth: batched gas limit fallback

chappjc:
Replaces https://github.com/decred/dcrdex/pull/2118

I think we've come to understand some of the live estimate failures better, and such an error is likely to be suggestive that the transaction could fail, which we don't handle well.  Also, retries will generally be resolved as provider's errors tend to be transient.  Thus, the live estimate errors remain fatal to `Swap`.  This PR just does:

- Before failing to Swap on account of insufficient reserves, use the the realistic gas limit when it is a batch swap.
- Actually include any non-zero redeem gas estimate.  See `initGasEstimate`
- When a token is not approved, we return our known gases, but this change now also ensures we return the correct `nSwap` in the new case of `n > w.maxSwapsInTx`

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -1443,7 +1443,9 @@ type gasEstimate struct {
 	// IF the redeem asset is a fee-family asset. Otherwise Redeem and RedeemAdd
 	// are zero.
 	dexeth.Gases
-	// Additional fields are based on live estimates of the swap.
+	// Additional fields may be based on live estimates of the swap. Both oneGas
+	// and nGas will include both swap and redeem gas, but note that the redeem
+	// gas may be zero if the redeemed asset is not ETH or an ETH token.
 	oneGas, nGas, nSwap, nRedeem uint64
 }
 
@@ -1453,26 +1455,28 @@ type gasEstimate struct {
 func (w *assetWallet) initGasEstimate(n int, initVer, redeemVer, redeemAssetID uint32) (est *gasEstimate, err error) {
 	est = new(gasEstimate)
 
-	g := w.gases(initVer)
-	if g == nil {
+	// Get the refund gas.
+	if g := w.gases(initVer); g == nil {
 		return nil, fmt.Errorf("no gas table")
+	} else { // scoping g
+		est.Refund = g.Refund
 	}
-
-	est.Refund = g.Refund
 
 	est.Swap, est.nSwap, _, err = w.swapGas(n, initVer)
 	if err != nil {
 		return nil, fmt.Errorf("error calculating swap gas: %w", err)
 	}
 
-	est.oneGas = est.Swap + est.Redeem
-	est.nGas = est.nSwap + est.nRedeem
+	est.oneGas = est.Swap
+	est.nGas = est.nSwap
 
 	if redeemW := w.wallet(redeemAssetID); redeemW != nil {
 		est.Redeem, est.nRedeem, err = redeemW.redeemGas(n, redeemVer)
 		if err != nil {
 			return nil, fmt.Errorf("error calculating fee-family redeem gas: %w", err)
 		}
+		est.oneGas += est.Redeem
+		est.nGas += est.nRedeem
 	}
 
 	return
@@ -1480,7 +1484,8 @@ func (w *assetWallet) initGasEstimate(n int, initVer, redeemVer, redeemAssetID u
 
 // swapGas estimates gas for a number of initiations. swapGas will error if we
 // cannot get a live estimate from the contractor, which will happen if the
-// wallet has no balance.
+// wallet has no balance. A live gas estimate will always be attempted, and used
+// if our expected gas values are lower (anomalous).
 func (w *assetWallet) swapGas(n int, ver uint32) (oneSwap, nSwap uint64, approved bool, err error) {
 	g := w.gases(ver)
 	if g == nil {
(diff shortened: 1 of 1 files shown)


client/asset/eth/eth.go
```diff
@@ -1768,16 +1778,32 @@ func (w *ETHWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin, uint6
...
-
 	if swapVal+fees > reservedVal {
-		return fail("unfunded swap: %d < %d", reservedVal, swapVal+fees)
+		if n == 1 {
+			return fail("unfunded swap: %d < %d", reservedVal, swapVal+fees)
+		}
+		w.log.Warnf("Unexpectedly low reserves for %d swaps: %d < %d", n, reservedVal, swapVal+fees)
+		// Since this is a batch swap, attempt to use the realistic limits.
+		gasLimit = nSwap
+		fees = gasLimit * swaps.FeeRate
```
chappjc, line 1796: `fees` is a return from `Swap`, so arguably we should be returning this correctly batched fees always.
What does everyone think about how presently we call `w.initiate` below with the naive unbatched gas limit?  Is this a bug or just conservative?  This reserves condition I'm adding here could potentially be the normal path...

chappjc: Actually the dynamic fees update is quite jarring with batched swap because of this quirk / conservative tx authoring with batched swaps.

chappjc: hmm, considering this more, if we do *not* use the properly batched gas limits, the tx might be rejected if it even approaches the max swaps per tx limit

martonp (approved):
LGTM

client/asset/eth/eth.go
martonp: > hmm, considering this more, if we do not use the properly batched gas limits, the tx might be rejected if it even approaches the max swaps per tx limit

The `maxSwapsInTx` is calculated using the individual `Swap` value not the batched value, so I don't think this would happen. The downside of using the batched estimate is that the real value would be more than 10% higher than the estimate, which is unlikely, so might as well go with that if only to just make the dynamic fee update less jarring.

client/asset/eth/eth.go
```diff
@@ -1453,34 +1455,37 @@ type gasEstimate struct {
...
+	est.oneGas = est.Swap
+	est.nGas = est.nSwap
 
 	if redeemW := w.wallet(redeemAssetID); redeemW != nil {
 		est.Redeem, est.nRedeem, err = redeemW.redeemGas(n, redeemVer)
 		if err != nil {
 			return nil, fmt.Errorf("error calculating fee-family redeem gas: %w", err)
 		}
+		est.oneGas += est.Redeem
+		est.nGas += est.nRedeem
```
chappjc, line 1479: So this is supposed to be a bug fix, rightly pointed out by @buck54321 as doing nothing when it was above before getting est.Redeem/Redeemn, but isn't this inconsistent with the `oneSwap` that `swapGas` gets from `w.gases`?  Why should the init gas estimate include redeem gas at all?

client/asset/eth/eth.go
chappjc: Consider if it was a batch of say, _half_ of `maxSwapsInTx`.  Here in `Swap` we call `oneSwap, nSwap, _, err := w.swapGas(n, swaps.Version)`, and assume it gives us the expected hard-coded gas values for both `oneSwap` and `nSwap`.  Why would we use `oneSwap * n` when we know `nSwap` is correct for this batch init tx?
Because we do use `oneSwap * n`, that could go over the `maxTxFeeGwei` limit even though we're far less than `maxSwapsInTx`.  The transaction would not actually consume that gas, but I'm wondering at what level the limit is enforced?  The EVM could easily decide not to even try executing the txn if the gas limit were set too high.

chappjc: Just saw your response.

> The maxSwapsInTx is calculated using the individual Swap value not the batched value, so I don't think this would happen

Why would we do that though?  Shouldn't it be based on the batched value?

client/asset/eth/eth.go
chappjc: Oh lordy, I keep mixing up `estimateInitGas` and `initGasEstimate`.  NVM this concern about including Redeem gas.   This method is in fact just for funding, not tx authoring.

client/asset/eth/eth.go
buck54321: > Why would we use `oneSwap * n` when we know `nSwap` is correct for this batch init tx?
> Because we do use `oneSwap * n` (to be conservative?), that could go over the `maxTxFeeGwei` limit even though we're far less than `maxSwapsInTx`.

Good point. We used `oneSwap * n` because that was what we planned for in the worst case, but since we are batching here and we better understand the gas limit now, `nSwap` seems appropriate.

JoeGruffins (approved):

client/asset/eth/eth.go
chappjc: Asking because we've testing as-is and there are two approves, and I hate to fundamentally change PRs at that stage.

chappjc: @martonp has rightly pointed out that the batch savings is not huge, and because of how `maxSwapsInTx` is computed with the individual Swap value (that's ver questionable too BTW), this is fine as-is.  Perhaps we should look again at this later, perhaps with 0.6.1.

buck54321: Sounds good.

buck54321 (approved):

PR #2144, opened 2023-02-17
client/db/bolt: ignore notes with bad IDs

chappjc:
The db.Notification serialization and thus ID changed in https://github.com/decred/dcrdex/commit/c994f032fe0e9a36e58a9a984dd924734c659e36#diff-592072666205a423f50bc3061a74c2cda51bfe4689d4044315ff285d29645bc3R832
The ID was used as the bucket key.
Therefore any notes prior to the upgrade will be stored in a bucket with a different key than it's new ID.  This isn't a big deal except that acknowledging such notes will fail for eternity.

In draft because I'm considering auto-acking such notes on load.

### DIFF
--- a/client/db/bolt/db.go
+++ b/client/db/bolt/db.go
@@ -1835,7 +1835,7 @@ func (db *BoltDB) AckNotification(id []byte) error {
 // NotificationsN reads out the N most recent notifications.
 func (db *BoltDB) NotificationsN(n int) ([]*dexdb.Notification, error) {
 	notes := make([]*dexdb.Notification, 0, n)
-	return notes, db.notesView(func(master *bbolt.Bucket) error {
+	return notes, db.notesUpdate(func(master *bbolt.Bucket) error {
 		trios := newestBuckets([]*bbolt.Bucket{master}, n, stampKey, nil)
 		for _, trio := range trios {
 			note, err := dexdb.DecodeNotification(getCopy(trio.b, noteKey))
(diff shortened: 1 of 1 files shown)


client/db/bolt/db.go
```diff
@@ -1844,6 +1844,16 @@ func (db *BoltDB) NotificationsN(n int) ([]*dexdb.Notification, error) {
 			}
 			note.Ack = bEqual(trio.b.Get(ackKey), byteTrue)
 			note.Id = note.ID()
+			if !bytes.Equal(note.Id, trio.k) {
+				if !note.Ack {
+					// This notification was initially stored when the serialization
+					// and thus note ID did not include the TopicID. Ignore it.
+					db.log.Tracef("Ignoring stored note with bad key: %x != %x \"%s\"",
+						[]byte(note.Id), trio.k, note.String())
+					// trio.b.Put(ackKey, byteTrue)
```
chappjc, line 1853: We could even move it to a new bucket with the correct key, but these are 1-1.5 year old notifications and I kinda don't wanna bother.

buck54321: I'm good with auto-acking these.

chappjc: Acking.  The two unacknowledgeable notes logged once, and are now gone after restart.  I think this should be good for this fairly rare situation (upgraded with unacked notes, over a year ago).

chappjc: Also worth noting that these notes would fall off the latest 100 list eventually anyway, but that can take a while for light use.

JoeGruffins (approved):

client/db/bolt/db.go
chappjc: Auto-acking in last force push.

buck54321 (approved):

PR #2148, opened 2023-02-18
core: Trim URL before parsing

### LINKED ISSUE #2146: Add DEX Page Problem
When I try to add a DEX server with the UI, if I leave a space after the server's URL, it will attempt to connect to localhost. When I remove the space, it works.

martonp:
Additional space at the end was causing the port to not parse properly.

Closes #2146

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1655,6 +1655,7 @@ const defaultDEXPort = "7232"
 
 // addrHost returns the host or url:port pair for an address.
 func addrHost(addr string) (string, error) {
+	addr = strings.TrimSpace(addr)
 	const defaultHost = "localhost"
 	const missingPort = "missing port in address"
 	// Empty addresses are localhost.


client/core/core.go
```diff
@@ -1655,6 +1655,7 @@ const defaultDEXPort = "7232"
 
 // addrHost returns the host or url:port pair for an address.
 func addrHost(addr string) (string, error) {
+	addr = strings.TrimSpace(addr)
```
chappjc, line 1658: bizzare: https://play.golang.com/p/VL3fC6d8E_y

chappjc (approved):

client/core/core.go
ukane-philemon: > bizzare: https://play.golang.com/p/VL3fC6d8E_y

Interesting. Does did not qualify as a bug upstream?

chappjc: I don't know.  They'd probably say it has to be that way because the port part can be "service" like "ftp" or "dns" or whatever, so they don't validate whatever is after the last ":".

PR #2149, opened 2023-02-19
server/auth: use `misstep` parameter of `*AuthManager.Inaction(...)` instead of global `SwapSuccess` variable

### LINKED ISSUE #2072: server/auth: something seems wrong with this AuthManager method
https://github.com/decred/dcrdex/blob/8d99975b34e4ed372953ee39f5fbc659b2302918/server/auth/auth.go#L965-L967

I think maybe that first line is supposed to be 

```Go
violation := misstep.Violation()
```

ukane-philemon:
Closes #2072


chappjc (approved):
Thanks.  I could have sworn I fixed this but must have forgotten.  Glad you checked.

PR #2150, opened 2023-02-20
testing: add miners to harnesses

buck54321:
Just a thought. We don't have to auto-mine, and I haven't considered how this might affect others tests, but it seems simple enough and I think we'll want something reasonable for bond maintenance.

### DIFF
--- a/dex/testing/btc/base-harness.sh
+++ b/dex/testing/btc/base-harness.sh
@@ -201,6 +201,7 @@ cat > "${HARNESS_DIR}/quit" <<EOF
 tmux send-keys -t $SESSION:0 C-c
 tmux send-keys -t $SESSION:1 C-c
 tmux send-keys -t $SESSION:3 C-c
+tmux send-keys -t $SESSION:4 C-c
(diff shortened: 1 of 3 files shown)


chappjc:
Seems reasonable to me.  The rotateBonds method is called every 20 seconds, so I found on simnet that I had to set me watch interval to be less than that (settled on 15) otherwise you occasionally rotate too late and get a tier drop.

ukane-philemon:
Nice. Makes sense and is a useful addition.

buck54321:
E2E tests seem to be working. Loadbot runs a miner that targets 1 block per roughly two epochs. If it's important to maintain that target (I don't think it is), we can always shut the miners down.

JoeGruffins (approved):
Probably fine. Can always turn them off. Some tests mine though... I feel like dcr gets funky if you try to mine from two places. But again can just turn off and let the test mine.

The dcr harness goes to the mining session rather than the ctrl-harness after these changes.

JoeGruffins:
The dcr harness still defaults to a different session.

PR #2152, opened 2023-02-20
client/core: simplify view-only backend machinery, remove connectAccount helper

chappjc:
The one place a view-only flag needs to exist is `core.Exchange.ViewOnly` so that when marshaled to JSON for the frontend, it is possible to recognize when a DEX is view-only.

Within core (`dexAccount`, `db.AccountInfo`, etc.) a view-only account is characterized by the lack of account keys, which are the user's identity.  These values are set in `discoverAccount` > `setupCryptoV2`, and implicitly convert a view-only dex connection.  There seems to be no need in `Register` to distinguish a dex account that is *intended* to be for trading (not view-only) for the brief period between `connectDEX` and `addDexConnection`, so this commit simplifies everything.

### DIFF
--- a/client/core/types.go
+++ b/client/core/types.go
@@ -700,11 +700,10 @@ type dexAccount struct {
 	cert      []byte
 	dexPubKey *secp256k1.PublicKey
 
-	keyMtx   sync.RWMutex
-	viewOnly bool // true, unless account keys are generated AND saved to db
-	encKey   []byte
-	privKey  *secp256k1.PrivateKey
-	id       account.AccountID
+	keyMtx  sync.RWMutex
+	encKey  []byte
+	privKey *secp256k1.PrivateKey
+	id      account.AccountID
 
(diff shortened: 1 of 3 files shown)


chappjc:
nevermind.  I wish for this though.

client/core/types.go
```diff
@@ -750,13 +748,6 @@ func (a *dexAccount) ID() account.AccountID {
 	return a.id
 }
 
-// isViewOnly is true if account keys have not been generated AND saved to db.
```
chappjc, line 753: I think the reason this PR isn't quite OK is this sorta confusing phrase "AND saved to db."

I'm thinking all these changes would be fine if we made `dbCreateOrUpdateAccount` unnecessary by changing `c.db.UpdateAccountInfo` in to and **upsert** operation, and removed `db.CreateAccount`.  Then things would work fine without the extra flags and such.

PR #2157, opened 2023-02-22
multi: build with eth code by default

chappjc:
This negates the `lgpl` build tag, making it `nolgpl`.

The **importlgpl.go** files in the CLI app (`main`) packages are updated to indicate that the `nolgpl` build tag should be used to exclude such code if it is not acceptable.

Use of any of the eth-specific packages now _implies_ lgpl. That is, we gate at the import of these packages in the cmd apps that would use them via their **importlgpl.go** files that require the `!nolgpl` build tag.

All docs and scripts are updated for the new default.


martonp (commented):
Why aren't you putting`!nolgpl` in all the eth files? Was that just never needed?

chappjc:
> Why aren't you putting`!nolgpl` in all the eth files?

> Use of any of the eth-specific packages now implies lgpl. That is, we gate at the import of these packages in the cmd apps that would use them via their importlgpl.go files that require the !nolgpl build tag.

Seems reasonable since if you import the packages explicitly.  There's a doc.go in each of those that basically states "by using this package use are using LGPL code"

chappjc:
The other unfortunate thing I noticed is that if you use the gopkg.dev docs page (https://pkg.go.dev/decred.org/dcrdex), it does not even show the packages with no default-built code.  I think it should show it all, and the doc.go explains the license of that package.

martonp (approved):

JoeGruffins (approved):

PR #2160, opened 2023-02-22
docker: on release branches we skip npm build

chappjc:
The webpack build output is checked in on release branches for reasons, so the Dockerfile can/should skip that.


chappjc:
Gonna add a special dockerfile instead

Issue #2164, opened 2023-02-23
old orders (pre-matchconfirmed) show as settling after upgrading to 0.6


chappjc:
They aren't loaded on login or treated as "active", but from the orders page they are shown with status settling.

chappjc:
Will push a couple possible resolutions we can consider.

PR #2165, opened 2023-02-24
client/db/bolt: patch archived match status based on bucket

chappjc:
Option 1 of 2 that resolves https://github.com/decred/dcrdex/issues/2164
See other option at https://github.com/decred/dcrdex/pull/2166

I have confirmed this PR works as expected.


chappjc:
This option is working fine, and it covers matches that were saved on master, while the other one will only be good after merged.

PR #2166, opened 2023-02-24
client/db,dex/order: patch archived match status based on enc version

chappjc:
Option 2 of 2 that resolves https://github.com/decred/dcrdex/issues/2164
See other option at https://github.com/decred/dcrdex/pull/2165

I **have** confirmed that this PR works yet.


JoeGruffins (approved):
I prefer this method. It sets up a way to do upgrades like this in the future. I'd even do it once as a db upgrade.

chappjc:
> I prefer this method. It sets up a way to do upgrades like this in the future. I'd even do it once as a db upgrade.

OK will test this method out to make sure it works as expected. I don't want to rewrite all archived matches though.

chappjc:
Got to testing it and it works as expected for my old testnet setup

martonp (approved):
I think this is better too.

PR #2170, opened 2023-02-26
client/dcr: Incorrect excess return from sendEnough

martonp:
The txFee was being incorrectly subtracted from the excess value when `subtract == true`. This does not need to be subtracted from the `excess`, as it is subtracted from the actual value being sent.


martonp:
I noticed this by trying to send the full available amount from the wallet by clicking the available balance number on the send popup. It then failed due to insufficient funds.

chappjc (approved):

PR #2172, opened 2023-02-26
ui: fix sort direction and cursor on recent matches table

buck54321:
Cursor was set to pointer for the entire table, but only the headers are interactive. 

Sort direction showed oldest first. Changed to newest first.

### DIFF
--- a/client/webserver/site/src/css/market.scss
+++ b/client/webserver/site/src/css/market.scss
@@ -757,10 +757,6 @@ div[data-handler=markets] {
       }
     }
   }
-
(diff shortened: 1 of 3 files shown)


norwnd (approved):
Works nicely!

JoeGruffins (approved):

chappjc (approved):

chappjc:
NOTE: style.scss and index.js cache busters are not bumped, so just hard refresh after this.  There will be more frontend fixes coming anyway.

PR #2179, opened 2023-02-27
site: wipe orders on OrderRetired topic order note

chappjc:
Pertaining to https://github.com/decred/dcrdex/pull/2171, the `OrderRetired` order notification topic is sent when the backend "retires" the order, meaning it deletes it from the `trackedTrades` map.  This is when an order ceases to be listed in the "Your Orders" table of the markets page, and is only accessible via the historical orders methods.

As such, this change recognizes the `OrderRetired` topic on the frontend and removes it from the "Your Orders" table automatically (since a page refresh would remove it too).

We could use `setTimeout` to remove it after another minute or so, which would be helpful for simnet where ticks are every 7.5 by default, and even mainnet if the timing of order completion is such that the tick that retires it is not long after.

### DIFF
--- a/client/webserver/site/src/js/markets.ts
+++ b/client/webserver/site/src/js/markets.ts
@@ -2050,6 +2050,10 @@ export default class MarketsPage extends BasePage {
     mord.ord = order
     if (note.topic === 'MissedCancel') Doc.show(mord.details.cancelBttn)
     if (order.filled === order.qty) Doc.hide(mord.details.cancelBttn)
+    if (note.topic === 'OrderRetired') {
+      delete this.metaOrders[order.id]
+      mord.div.remove()
+    }
     if (app().canAccelerateOrder(order)) Doc.show(mord.details.accelerateBttn)
     else Doc.hide(mord.details.accelerateBttn)
     this.updateMetaOrder(mord)
--- a/client/webserver/site/src/html/bodybuilder.tmpl
+++ b/client/webserver/site/src/html/bodybuilder.tmpl
@@ -104,7 +104,7 @@
 {{end}}
 
 {{define "bottom"}}
-<script src="/js/entry.js?v=qAsSUzx7"></script>
+<script src="/js/entry.js?v=NhPu0J3"></script>
 </body>
 </html>
 {{end}}


buck54321:
I'm typically weary of removing elements from the page without user interaction. It's not great UX to be monitoring an active order an then it just disappears, forcing the user to go find it in the orders page if they want to ensure that it settled correctly.

chappjc:
> I'm typically weary of removing elements from the page without user interaction. It's not great UX to be monitoring an active order an then it just disappears, forcing the user to go find it in the orders page if they want to ensure that it settled correctly.

But if you refresh the page after it is retired, the the order won't be there either.

The flip side is https://github.com/decred/dcrdex/pull/2171 where you can manually remove them, but they just come right back.

client/webserver/site/src/js/markets.ts
```diff
@@ -2050,6 +2050,10 @@ export default class MarketsPage extends BasePage {
     mord.ord = order
     if (note.topic === 'MissedCancel') Doc.show(mord.details.cancelBttn)
     if (order.filled === order.qty) Doc.hide(mord.details.cancelBttn)
+    if (note.topic === 'OrderRetired') {
```
chappjc, line 2053: @buck54321 How about in https://github.com/decred/dcrdex/pull/2171, you catch this order note topic to set a `canClear` flag for the order, only showing the sweep icon for such orders.  That would solve the reappearing order issue.

The alternative, which I'd really rather avoid, is to redefine the Your Orders to include more than what's in the `trades` map, because the intent there is to report only active (and very recently completed but not yet retired) orders.

chappjc:
> > I'm typically weary of removing elements from the page without user interaction. It's not great UX to be monitoring an active order an then it just disappears, forcing the user to go find it in the orders page if they want to ensure that it settled correctly.
> 
> But if you refresh the page after it is retired, the the order won't be there either.
> 
> The flip side is #2171 where you can manually remove them, but they just come right back.

Point taken though.  I think I'd like my executed orders to be there when I check back, at least until I refresh the page (or it refreshes itself).  Closing this PR, but as I said in https://github.com/decred/dcrdex/pull/2179#discussion_r1119215068, I think #2171 may need to consider if the order has been retired, otherwise sweeping it away is not going to have the desired effect.

buck54321:
> But if you refresh the page after it is retired, the the order won't be there either.

I consider that a **user interaction**. 

> I think https://github.com/decred/dcrdex/pull/2171 may need to consider if the order has been retired, otherwise sweeping it away is not going to have the desired effect.

Definitely. Looking at this.

chappjc:
> I consider that a "user interaction".

I'm sure you will say we shouldn't be doing the following, but presently we refresh everything when a dex server reconnects because we've broken the sequence of notifications and we have to resync all state (server config, live orders, match statuses, everything).  So this is unfortunately something that happens all by itself now.

buck54321:
> I'm sure you will say we shouldn't be doing the following

I think I even removed that refresh in prior work, but there were some unintended consequences so we put it back. But yes, I think using hard refresh is unsophisticated and shows that we have shortcomings in how we maintain state.

Issue #2173, opened 2023-02-27
canceled orders showing as canceled/settling on markets page


chappjc:
Not one of the legacy historical orders.

/markets page:

![image](https://user-images.githubusercontent.com/9373513/221465922-0479f544-e6a0-4e03-b94a-c3337a736647.png)

That's the issue, although it didn't start or stay that way.  It starts as "cancelling", then changes to "canceled", but as soon as you submit a new trade order, this one changes to "canceled/settling", presumably when it refreshes the table for the new order you just booked.  Then the moment backend retires the order (`[INF] CORE: Retiring inactive order xxx in status canceled`), it changes to "canceled".
Perhaps the active matches loop is considering a cancel match as active?

norwnd:
Observing **possibly related behavior** with recently placed orders, in `booked` status, at some point "Your Orders" table starts to display `booked/settling` status (presumably during matches executing against those orders, which is expected) but doesn't eventually update `booked/settling` -> `booked` automatically (only does after I manually refresh the page) like it does on `/order/{id}` page.

<img width="200" alt="image" src="https://user-images.githubusercontent.com/112318969/221480110-ca3cd7af-994d-4b44-b9ad-08a24f307b7b.png">

I can dig into it if nobody is already.

chappjc:
Please do look.
I think the ".../settling" appears when the table is refreshed when a new order is inserted, regardless of whether that order has any fills.

PR #2180, opened 2023-02-27
client/core: fix cancel match ntfn being active/new

chappjc:
This resolves https://github.com/decred/dcrdex/issues/2173.

When the match cards on the order details page became fully dynamic, a new match notification was added for the cancel match.  However, the match was sent as `Active=true` and with `Status=order.NewlyMatched`.  This does not agree with the matches that get listed by `(*Core).coreOrderFromMetaOrder` (which serves the `/api/order` and `/api/orders` endpoints) because they pull the match data from the DB, which was patched in the `*db.MetaMatch` when storing it.

This resolves the bug by using the `*db.MetaMatch` in `(*trackedTrade).negotiate` when constructing this match ntfn.

Side note: the `trackedTrade.matches` map intentionally does not include cancel matches, and thus the `core.Order`s returned by `(*trackedTrade).coreOrderInternal` do not include the cancel matches in the `Matches []*core.Match` slice field.  It may or may not be worth resolving this inconsistency with `(*Core).coreOrderFromMetaOrder` that includes cancel matches.

### DIFF
--- a/client/core/trade.go
+++ b/client/core/trade.go
@@ -807,6 +807,7 @@ func (t *trackedTrade) negotiate(msgMatches []*msgjson.Match) error {
 	}
 
 	// Record any cancel order Match and update order status.
+	var metaCancelMatch *db.MetaMatch
 	if cancelMatch != nil {
 		t.dc.log.Infof("Maker notification for cancel order received for order %s. match id = %s",
 			t.ID(), cancelMatch.MatchID)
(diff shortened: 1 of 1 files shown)


norwnd:
> Note sure if this resolves the booked/settling case reported in @norwnd's comment https://github.com/decred/dcrdex/issues/2173#issuecomment-1445723166, but I confirmed this resolves the canceled/settling issue.

I'll try to reproduce it again with/without changes in this PR a bit later. And let you know.

JoeGruffins (approved):
I found it rather difficult to reproduce the issue. I could if I switched between markets with a cancel order, and more orders, but I don't see it with these changes.

Maybe because simnet times are faster.

chappjc:
https://user-images.githubusercontent.com/9373513/221742248-98e8de6f-0a61-4559-8518-454db73bb5f2.mp4

JoeGruffins:
I see on master:

https://user-images.githubusercontent.com/34998433/221749367-e3d776ba-c349-49fc-9adb-31b1079cdb24.mp4

chappjc:
Call me baffled.  This effect happens every time for me.  Most of the time it stays with settled too

chappjc:
Happens on mainnet for me too. Below it stayed that way:

image

The previous screen video was testnet at commit https://github.com/decred/dcrdex/pull/2181/commits/9d7f92c5a60645821a2a5fdf235ee617714239bb

Regardless of the reproduction, the root cause is fairly clear from the code.  Hopefully this also resolves the `booked/settling` issue observed by @norwnd.

JoeGruffins:
I was indeed able to reproduce on master on mainnet just now. Just making orders and canceling.

buck54321 (approved):

Issue #2175, opened 2023-02-27
ui: funding coin on order details page


norwnd:
While I typically do appreciate all the extra info I can get (about order execution process and whatnot), a typical DEX user won't probably know or care about `Funding Coins` section on order details page:

<img width="200" alt="image" src="https://user-images.githubusercontent.com/112318969/221493093-33b31134-2840-484e-bd8b-78052f932a71.png">

personally, fiddling with DEX for ~ half a year now I didn't use this `Funding Coins` data in UI even once (and it is available in logs for debugging purposes, for interested users anyway), but it does add cognitive load (and minor bloat) when digesting order details page, so I would suggest to remove it altogether from UI.

On the off chance I just haven't encountered a use-case for it (while some exist), then instead of removing, it is probably worth to
- make it a clickable link (similar to how transactions on Match cards are leading to blockchain explorers)
- maybe supply a tooltip that explains to the user what `Funding Coins` are and when he might need those

jholdstock:
I'm in favour of removing funding coins from the UI so long as it remains available in logs. Despite running into various issues and corner cases with erroneously locked outputs, I have never found that value in the UI useful. Agree its just bloat for 99% of users.

chappjc:
Darn, I personally use it quite often.

> On the off chance I just haven't encountered a use-case for it (while some exist), then instead of removing, it is probably worth to
> 
> * make it a clickable link (similar to how transactions on Match cards are leading to blockchain explorers)

They used to be!  This is a recent regression.

> * maybe supply a tooltip that explains to the user what `Funding Coins` are and when he might need those

OK, "the coins that are used to fund this order"?

chappjc:
Tagging this issue for 0.6, for at least making the links clickable again, possibly to somehow hide the funding coins section with "show funding coins" thing if we want to hide this text by default.

JoeGruffins:
I would rather not have the funding coins removed. I'll look at the pr though...

PR #2181, opened 2023-02-27
ui: funding/acceleration coins missing links regression

norwnd:
This is part 1 for addressing https://github.com/decred/dcrdex/issues/2175, it resolves regression with `Funding Coins` being clickable previously https://github.com/decred/dcrdex/issues/2175#issuecomment-1446343544 but no longer on the current **master**.

Seems like regression originates here: https://github.com/buck54321/dcrdex/commit/b9aaef2c280e2cb4ca1c4dbca99ac7b1fa939aee

Tested `Acceleration Coins` links too, seem to be working for BTC on **testnet**.

### DIFF
--- a/client/webserver/site/src/html/order.tmpl
+++ b/client/webserver/site/src/html/order.tmpl
@@ -203,7 +203,7 @@
       <div class="text-start">[[[Funding Coins]]]</div>
       <div class="fs14 text-start">
         {{range $ord.FundingCoins}}
-          <a target="_blank" class="plainlink mono" data-asset-id="{{$ord.FromID}}" data-explorer-coin="{{.StringID}}">{{.StringID}}</a><br>
+          <a target="_blank" class="plainlink mono" data-explorer-id="{{$ord.FromID}}" data-explorer-coin="{{.StringID}}">{{.StringID}}</a><br>
         {{end}}
       </div>
     </div>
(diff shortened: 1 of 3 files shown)


client/webserver/site/src/html/order.tmpl
```diff
@@ -212,7 +212,7 @@
       <div class="text-start">[[[acceleration_transactions]]]</div>
       <div class="fs14 text-start">
         {{range $ord.AccelerationCoins}}
-          <a target="_blank" class="plainlink mono" data-explorer-id="{{$ord.FromID}}" data-explorer-coin="{{.StringID}}">{{.StringID}}</a><br>
+          <a target="_blank" class="plainlink mono" data-asset-id="{{$ord.FromID}}" data-explorer-coin="{{.StringID}}">{{.StringID}}</a><br>
```
norwnd, line 215: Renaming to better reflect the assumptions we make (**explorer-id**=**asset-id**).

buck54321: What if we need to tag other elements with an asset ID. **This** asset ID is specifically for block explorer links. The attribute name reflected that.

buck54321 (approved):
Recommend leaving the attribute name alone, but whatever. I have no idea why I deleted that block in #1890.

client/webserver/site/src/html/order.tmpl
chappjc: Tend to agree.  We've named this because if its application, not the assumption of the data type.  Let's just leave it.

chappjc (approved):
Funding coin links are back on testnet, so LGTM, naming issue aside.

client/webserver/site/src/html/order.tmpl
norwnd: Renamed so it's like it was previously: https://github.com/decred/dcrdex/pull/2181/commits/2b6ba04c7aa2f5e07c741f8c5a5be16fff208340

PR #2183, opened 2023-02-27
core: Update wallet balance after funding order

martonp:
The wallet balance was not being updated after funding an order, causing the available balance to be too high and locked balance to be too low.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -5693,6 +5693,11 @@ func (c *Core) prepareTradeRequest(pw []byte, form *TradeForm) (*tradeRequest, e
 		return nil, codedError(walletErr, fmt.Errorf("FundOrder error for %s, funding quantity %d (%d lots): %w",
 			assetConfigs.fromAsset.Symbol, fundQty, lots, err))
 	}
+	defer func() {
+		if _, err := c.updateWalletBalance(fromWallet); err != nil {
+			c.log.Errorf("updateWalletBalance error: %v", err)
+		}
+	}()
 
 	coinIDs := make([]order.CoinID, 0, len(coins))
 	for i := range coins {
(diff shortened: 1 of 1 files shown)


client/core/core.go
```diff
@@ -5693,6 +5693,11 @@ func (c *Core) prepareTradeRequest(pw []byte, form *TradeForm) (*tradeRequest, e
 		return nil, codedError(walletErr, fmt.Errorf("FundOrder error for %s, funding quantity %d (%d lots): %w",
 			assetConfigs.fromAsset.Symbol, fundQty, lots, err))
 	}
+	defer func() {
+		if _, err := c.updateWalletBalance(fromWallet); err != nil {
```
chappjc, line 5697: Verified the bug on master, and that it's not on 0.5 (checked both on mainnet).

There was a `c.updateAssetBalance(...)` loop in `(*Core).Trade`, which we seem to have lost with the refactoring for `TradeAsync`.

However, the following was previously making the updated asset ID list, and considering parent wallets because possible locked fee balance.

```go
	updated := assetMap{fromWallet.AssetID: struct{}{}}
	if isAccountRefund && fromWallet.parent != nil {
		updated[fromWallet.parent.AssetID] = struct{}{}
	}
	if isAccountRedemption && toWallet.parent != nil {
		updated[toWallet.parent.AssetID] = struct{}{}
	}
```

I'm seeing the *parent* parts of that logic are now in `sendTradeRequest`, but I think all the balance updates belong where you are putting them here in `prepareTradeRequest`.

JoeGruffins (approved):

client/core/core.go
chappjc: I let this slip by, but do we also need to consider any parent asset for fee reserves?

martonp: You're probably right. I'll fix that.

Issue #2188, opened 2023-02-28
client/dex: candles race


norwnd:
Continuing race crusade:
```
WARNING: DATA RACE
Write at 0x00c00ae1c508 by goroutine 41:
  decred.org/dcrdex/dex/candles.(*Cache).combineCandles()
      /Users/norwnd/dcrdex/dex/candles/candles.go:174 +0x679
  decred.org/dcrdex/dex/candles.(*Cache).Add()
      /Users/norwnd/dcrdex/dex/candles/candles.go:62 +0x249
  decred.org/dcrdex/client/core.(*candleCache).addCandle()
      /Users/norwnd/dcrdex/client/core/bookie.go:101 +0xee
  decred.org/dcrdex/client/core.(*bookie).logEpochReport()
      /Users/norwnd/dcrdex/client/core/bookie.go:204 +0x484
  decred.org/dcrdex/client/core.handleEpochReportMsg()
      /Users/norwnd/dcrdex/client/core/bookie.go:1068 +0x152
  decred.org/dcrdex/client/core.(*Core).listen.func1()
      /Users/norwnd/dcrdex/client/core/core.go:8240 +0x156
  decred.org/dcrdex/client/core.(*Core).listen.func2()
      /Users/norwnd/dcrdex/client/core/core.go:8254 +0xcc

Previous read at 0x00c00ae1c508 by goroutine 13681:
  reflect.Value.Uint()
      /usr/local/go/src/reflect/value.go:2662 +0x169
  encoding/json.uintEncoder()
      /usr/local/go/src/encoding/json/encode.go:561 +0x45
  encoding/json.structEncoder.encode()
      /usr/local/go/src/encoding/json/encode.go:759 +0x2ba
  encoding/json.structEncoder.encode-fm()
      <autogenerated>:1 +0xdb
  encoding/json.ptrEncoder.encode()
      /usr/local/go/src/encoding/json/encode.go:943 +0x382
  encoding/json.ptrEncoder.encode-fm()
      <autogenerated>:1 +0x90
  encoding/json.structEncoder.encode()
      /usr/local/go/src/encoding/json/encode.go:759 +0x2ba
  encoding/json.structEncoder.encode-fm()
      <autogenerated>:1 +0xdb
  encoding/json.(*encodeState).reflectValue()
      /usr/local/go/src/encoding/json/encode.go:358 +0x88
  encoding/json.interfaceEncoder()
      /usr/local/go/src/encoding/json/encode.go:714 +0x17b
[...]

Issue #2168, opened 2023-02-24
gaps in candlesticks


chappjc:
This is not a new issue for 0.6, just reporting now.

![image](https://user-images.githubusercontent.com/9373513/221195307-76b0cd28-9f5d-4e54-818e-08402930e47d.png)

![image](https://user-images.githubusercontent.com/9373513/221195322-d863107f-9174-4f19-bca1-5e964c9200a0.png)

Usually this happens after you go away for a while and come back.  When you switch to the chart it looks alright for a couple seconds except that it's out of date, and then it redraws, but you are missing candles.

Refreshing the browser does not help, suggesting something is wrong on the backend.  I wonder if there's a gap in subscribe -> unsubscribe -> resubscribe, where it fails to get the data it missed while unsubscribed?

JoeGruffins:
I noticed this on mainnet the other day. I'm on 88692db99c7a4b11fe08efd587c4283368a09ad6

chappjc:
This has been a symptom since as long as I can remember having candlesticks.

JoeGruffins:
It seems I can reproduce this. Log in and view mainnet as normal and wait a bit. Then log in with a private tab. The candles are deleted from where you first logged in with the first terminal up until present, even for the first terminal.

ukane-philemon:
> It seems I can reproduce this. Log in and view mainnet as normal and wait a bit. Then log in with a private tab. The candles are deleted from where you first logged in with the first terminal up until present, even for the first terminal.

Yeah, I see it too and was able to reproduce it by just running mainnet for some minutes while on another tab.

chappjc:
I didn't know about the other private tab trick.  It also just happens on its own with a single tab logged in.

chappjc:
Because candlesticks are show by default, this is much harder to ignore:

![image](https://user-images.githubusercontent.com/9373513/221729817-39aeaca8-42d7-4abf-9d51-4fecb388a803.png)

I'm planning to look into this, but I think we should consider this high priority.

PR #2189, opened 2023-02-28
client/dex: resolve candles race

norwnd:
Addresses a part of https://github.com/decred/dcrdex/issues/2188 specifically pertaining to candles issue: https://github.com/decred/dcrdex/issues/2168.

### DIFF
--- a/client/core/bookie.go
+++ b/client/core/bookie.go
@@ -64,6 +64,9 @@ func (f *bookFeed) Candles(durStr string) error {
 // candleCache adds synchronization and an on/off switch to *candles.Cache.
 type candleCache struct {
 	*candles.Cache
+	// candleMtx protects the integrity of candles.Cache (e.g. we can't update
+	// it while making copy at the same time), so it represents a consistent
+	// data snapshot.
 	candleMtx sync.RWMutex
 	on        uint32
 }
@@ -91,15 +94,16 @@ func (c *candleCache) init(in []*msgjson.Candle) {
 	}
 }
 
-// addCandle adds the candle using candles.Cache.Add.
-func (c *candleCache) addCandle(msgCandle *msgjson.Candle) *msgjson.Candle {
+// addCandle adds the candle using candles.Cache.Add. It returns most recent candle
+// in cache.
+func (c *candleCache) addCandle(msgCandle *msgjson.Candle) (recent msgjson.Candle, ok bool) {
 	if atomic.LoadUint32(&c.on) == 0 {
-		return nil
+		return msgjson.Candle{}, false
 	}
 	c.candleMtx.Lock()
 	defer c.candleMtx.Unlock()
 	c.Add(msgCandle)
-	return c.Last()
+	return *c.Last(), true
 }
 
 // bookie is a BookFeed manager. bookie will maintain any number of order book
@@ -201,12 +205,11 @@ func (b *bookie) logEpochReport(note *msgjson.EpochReportNote) error {
 		})
 	}
 	for durStr, cache := range b.candleCaches {
-		c := cache.addCandle(&note.Candle)
-		if c == nil {
+		c, ok := cache.addCandle(&note.Candle)
+		if !ok {
 			continue
 		}
 		dur, _ := time.ParseDuration(durStr)
-		cCopy := *c
 		b.send(&BookUpdate{
 			Action:   CandleUpdateAction,
 			Host:     b.dc.acct.host,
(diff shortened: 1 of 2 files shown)


client/core/bookie.go
```diff
@@ -213,7 +213,17 @@ func (b *bookie) logEpochReport(note *msgjson.EpochReportNote) error {
 			Payload: CandleUpdate{
 				Dur:          durStr,
 				DurMilliSecs: uint64(dur.Milliseconds()),
-				Candle:       c,
+				// Have to copy msgjson.Candle data here since it will be used concurrently.
```
chappjc, line 216: `Candle: *c,`?

client/core/bookie.go
```diff
@@ -213,7 +213,8 @@ func (b *bookie) logEpochReport(note *msgjson.EpochReportNote) error {
...
+				Candle: &*c,
```
chappjc, line 217: `SA4001: &*x will be simplified to x. It will not copy x. (staticcheck)`

Sorry, I made my suggestion assuming you had change the `CandleUpdate` definition to have a value.

Let's just make the copy a few lines up, before `b.send`

client/core/bookie.go
```diff
@@ -206,14 +206,16 @@ func (b *bookie) logEpochReport(note *msgjson.EpochReportNote) error {
 			continue
 		}
 		dur, _ := time.ParseDuration(durStr)
+		cCopy := *c
```
buck54321, line 209: Shouldn't we make the copy in `addCandle` while we have the `candleMtx` locked?

client/core/bookie.go
```diff
@@ -91,15 +94,16 @@ func (c *candleCache) init(in []*msgjson.Candle) {
 	}
 }
 
-// addCandle adds the candle using candles.Cache.Add.
-func (c *candleCache) addCandle(msgCandle *msgjson.Candle) *msgjson.Candle {
+// addCandle adds the candle using candles.Cache.Add. It returns most recent candle
+// in cache.
+func (c *candleCache) addCandle(msgCandle *msgjson.Candle) (recent msgjson.Candle, ok bool) {
```
buck54321, line 99: Why are we changing the return type? Just return a pointer.

norwnd: I find pointers are harder to think about (specifically when they appear in concurrent code), so just follow a convention of trying to limit the scope of their usage when possible. Could change it back to pointer - it's not really affecting anything here.

chappjc (approved):
Been a few hours and I have not noticed the candlestick gaps.

client/core/bookie.go
JoeGruffins: I'm still trying to wrap my head around the issue, but pointers are passed around to save on memory allocations and speed up the app. And, it can be just as confusing to copy structs. Take this for example, some types from the copied struct are not actually copied but passed by reference themselves: https://go.dev/play/p/TAYp0ruB6pL

JoeGruffins:
> Been a few hours and I have not noticed the candlestick gaps.

I am still able to induce them by logging in on another tab.

JoeGruffins:
Ah, my review was too slow.

chappjc:
> I am still able to induce them by logging in on another tab.

Sorry, I was actually hoping to merge it before you put time into it, but good additions to the discussion anyway!

> I am still able to induce them by logging in on another tab.

This is a pretty darn good clue about the root of the problem.  I was crossing my fingers this race some how was related, but I wasn't seeing how.  Oh well.  https://github.com/decred/dcrdex/issues/2168 is still opened.

client/core/bookie.go
chappjc: interfaces also like to sneak past our intuition as under the hood they are basically data pointers with a type label in tow.  Fun fact - maps (if initialized) are literally pointers to unexported types the implementation. https://dave.cheney.net/2017/04/30/if-a-map-isnt-a-reference-variable-what-is-it

Still, there are good points in this discussion that it's arguably more expressive to be returning a value.  However, I don't think we should follow that rigidly at all, especially with objects that are many times larger than a pointer, but there are merits.

norwnd: > And, it can be just as confusing to copy structs.

True, but all else being equal, I'd use whatever primitives language provides to lower the probability of making error.

> However, I don't think we should follow that rigidly at all, especially with objects that are many times larger than a pointer, but there are merits.

I didn't refactor it because of refactor, changing return type there was the first thing that came to mind to solve the problem and I didn't think it is a change that could rub anyone in a wrong way (it's couple of line, and straightforward). Everyone has different preferences for small details, I too don't think we should be spending much time discussing/fighting over them.

PR #2191, opened 2023-02-28
client/asset/eth: break immediately on shutdown when timing out

chappjc:
Resolves https://github.com/decred/dcrdex/issues/2045

Turns out the timeouts there before shutdown were (wait for it)... DNS.  The only actual issue there was that shutting down when `eth.checkForNewBlocks` was timing out would involve a random chance of receiving from the canceled context instead of the ticker.

This change checks for a shutdown that may have occurred while `eth.checkForNewBlocks` timed out.  I've decided to keep the ticker because with health providers, it is desirable to keep a regular check interval in terms of wall time rather than relative to the previous check.  Tickers don't backup indefinitely, but they will buffer a single tick, so a single tick may be delayed when there is a timeout.

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -3151,8 +3151,11 @@ func (eth *ETHWallet) monitorBlocks(ctx context.Context, reportErr func(error)) 
 	for {
 		select {
 		case <-ticker.C:
-			eth.checkForNewBlocks(reportErr)
+			eth.checkForNewBlocks(ctx, reportErr)
 		case <-ctx.Done():
+			return
+		}
+		if ctx.Err() != nil { // shutdown during last check, disallow chance of another tick
 			return
 		}
 	}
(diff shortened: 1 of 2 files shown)


client/asset/eth/eth.go
```diff
@@ -3151,18 +3151,21 @@ func (eth *ETHWallet) monitorBlocks(ctx context.Context, reportErr func(error))
...
 	}
 }
 
 // checkForNewBlocks checks for new blocks. When a tip change is detected, the
 // tipChange callback function is invoked and a goroutine is started to check
 // if any contracts in the findRedemptionQueue are redeemed in the new blocks.
-func (eth *ETHWallet) checkForNewBlocks(reportErr func(error)) {
-	ctx, cancel := context.WithTimeout(eth.ctx, 2*time.Second)
+func (eth *ETHWallet) checkForNewBlocks(ctx context.Context, reportErr func(error)) {
+	ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
```
chappjc, line 3168: Same actual context, but I we're moving toward best practices.

buck54321 (approved):

JoeGruffins (approved):

PR #2194, opened 2023-03-01
client/core: bad log msg formatting

norwnd:
Noticed this in logs, fixing:

```
2023-02-28 08:16:01.156 [WRN] CORE: Failed to validate commitment checksum for %!s(uint64=239653162) epoch %!d(string=dex-test.ssgen.io:7232) at checksum #߄�
         ���ONƾ���� (I�	� != trade order preimage request checksum  for trade order 699b4ba88c8db764c84785d096c803fd012a12901fea64e25d61793c63a0c37d: %!v(MISSING)dcr_eth
```

### DIFF
--- a/client/core/trade.go
+++ b/client/core/trade.go
@@ -582,7 +582,7 @@ func (t *trackedTrade) cancelEpochIdx() uint64 {
 	return uint64(t.cancel.Prefix().ServerTime.UnixMilli()) / epochLen
 }
 
-func (t *trackedTrade) verifyCSum(csum []byte, epochIdx uint64) error {
+func (t *trackedTrade) verifyCSum(csum dex.Bytes, epochIdx uint64) error {
 	t.mtx.RLock()
 	defer t.mtx.RUnlock()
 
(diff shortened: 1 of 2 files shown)


client/core/trade.go
```diff
@@ -582,7 +582,7 @@ func (t *trackedTrade) cancelEpochIdx() uint64 {
 	return uint64(t.cancel.Prefix().ServerTime.UnixMilli()) / epochLen
 }
 
-func (t *trackedTrade) verifyCSum(csum []byte, epochIdx uint64) error {
+func (t *trackedTrade) verifyCSum(csum dex.Bytes, epochIdx uint64) error {
```
norwnd, line 585: While testing updates for logged message above, found this:
```
2023-03-01 08:27:57.333 [WRN] CORE: Failed to validate commitment checksum for dcr_eth epoch 239665610 at dex-test.ssgen.io:7232: checksum ���$aS��Qw8�Ȟ!�p�X@��}EZ�E�� != trade order preimage request checksum 038519cbe72461538df4517738a5c89e2196708d58408ebe7d05455a8b45ccc5 for trade order 25ffdbb388c7b7e67a4f0335cbbab50ec3e4aed7eb01c2b70f99e993bf9161ee
```

should also be fixed ? This change seems to help it.

chappjc: Yes, thank you.

A side note about `dex.Bytes` though -- in this case the caller is a message handler and the input was already a `dex.Bytes` so this is fine to change like this, but in most other contexts `dex.Bytes` should be limited for use in comms applications where its json marshal/unmarshal methods are required, otherwise we should stick to native types like `[]byte`.

chappjc (approved):

PR #2196, opened 2023-03-01
{client,server}/btc: BTC Fidelity Bonds

martonp:
This enables BTC fidelity bonds on both the client and the server. The implementation is almost identical to DCR. Notably, on the client, the `sendToAddress` wallet API is no longer used because sends need to respect the bond reserves. Instead, the transaction is constructed in the client and passed to the wallet through `sendRawTransaction`.

An `UnsignedCoinID` field is added to the `asset.Bond` struct in order to support BTC wallets that have segwit turned off.

This is still untested in the electrum wallet and BTC clones.


dex/networks/btc/script.go
```diff
@@ -601,6 +627,205 @@ func RefundP2SHContract(contract, sig, pubkey []byte) ([]byte, error) {
...
+		if templateOffset >= len(template) {
+			return 0, nil, errors.New("too many script elements")
+		}
+
+		op, data := tokenizer.Opcode(), tokenizer.Data()
+		tplEntry := &template[templateOffset]
+		if tplEntry.expectInt {
+			switch {
+			case data != nil:
+				val, err := dcrtxscript.MakeScriptNum(data, tplEntry.maxIntBytes)
```
martonp, line 782: These "dcrtxscript" functions are identical in the btcd code, but unexported. I'll do a PR in the btcd repo to export them so we can use the correct ones.

client/core/bond.go
```diff
@@ -499,7 +499,7 @@ func (c *Core) preValidateBond(dc *dexConnection, bond *asset.Bond) error {
 		return fmt.Errorf("account keys not decrypted")
 	}
 
-	assetID, bondCoin := bond.AssetID, bond.CoinID
+	assetID, bondCoin := bond.AssetID, bond.UnsignedCoinID
```
chappjc, line 502: Since we only store a `db.Bond` after `preValidateBond` has succeeded, I suspect we don't need to modify any of the DB code to store this.

Looking at how this coinID is used, does it seem kinda worthless?  Is the only place it's used just below with `if !bytes.Equal(preBondRes.BondID, bondCoin) {`?

client/asset/btc/coin_selection.go
```diff
@@ -0,0 +1,212 @@
...
+
+	"decred.org/dcrdex/dex/calc"
+	dexbtc "decred.org/dcrdex/dex/networks/btc"
+)
+
+// sendEnough generates a function that can be used as the enough argument to
+// the fund method when creating transactions to send funds. If fees are to be
+// subtracted from the inputs, set subtract so that the required amount excludes
+// the transaction fee. If change from the transaction should be considered
+// immediately available (not mixing), set reportChange to indicate this and the
```
chappjc, line 19: The "not mixing" and mentions of mixing can probably get removed.

client/asset/btc/btc.go
```diff
@@ -388,9 +388,6 @@ type BTCCloneCFG struct {
 	ManualMedianTime bool
 	// OmitRPCOptionsArg is for clones that don't take an options argument.
 	OmitRPCOptionsArg bool
-	// LegacySendToAddr sents legacy raw tx which does not have positional fee
-	// rate param.
-	LegacySendToAddr bool
```
chappjc, line 393: That reminds me, I think there was a comment in the dcr code that says something like "TODO: just use sendtoaddress?".  Haha, guess that's out of the question now.

client/core/bond.go
chappjc: Say we remove `CoinID` from `msgjson.PreValidateBondResult`, it somewhat devalues the message's signature as it doesn't necessarily indicate they've validated a specific transaction, but we can have the signature be an amalgamation of the request and the response fields.  e.g.

```go
	preBondRes := &msgjson.PreValidateBondResult{
		AccountID: acctID[:],
		AssetID:   assetID,
		Amount:    uint64(amt),
		Expiry:    uint64(expireTime.Unix()),
	}
	preBondRes.SetSig(auth.SignMsg(append(preBondRes.Serialize(), preBond.RawTx...)))
```

In some places we actually have the signature pertain to the request, which is admittedly generally expected, but it could work I think.

Client would obviously have to change their interpretation of `msgjson.PreValidateBondResult.Signature` so that could be a pain, but at least we aren't released yet.

martonp: How about instead of returning the `BondID` in the `PreValidateBondResult`, it just returns a `RawTransactionHash` which is always a sha256 hash of the raw transaction sent in the request.

chappjc: I gave this a thought too, but it's basically what `SignMsg` is doing, in that it does a `sha256` hash of some data.  If this data includes `RawTx`, you're validating that by validating the signature.  No need to actually send the sha256 of RawTx

martonp: Yeah they both work the same way, I just thought it would be good to keep it consistent that the signature in every response is just signing the response object.

chappjc: So: (1) don't return `BondID` in the `PreValidateBondResult`, (2) don't return `UnsignedCoinID` in `asset.Bond`, and (3) don't store it in `db.Bond` and the DB.

chappjc: > Yeah they both work, I just thought it would be good to keep it consistent that the signature in every response is just signing the response object.

Understood.  We're already not consistent on this though. (See connect response signature)
We could be more conventional however.  I just thing this coin ID is essentially meaningless, and it's quite a burden to pass around.  
I wouldn't be totally opposed to returning a `RawTxHash` in the manner you've described though.
Mainly I want to not have to deal with `bond.UnsignedCoinID` at any level.

client/asset/btc/btc.go
```diff
@@ -1458,9 +1480,49 @@ func (btc *baseWallet) Balance() (*asset.Balance, error) {
...
+	bal, err := btc.balance()
+	if err != nil {
+		return nil, err
+	}
+
+	reserves := btc.reserves()
+	if reserves > bal.Available {
+		btc.log.Warnf("Available balance is below configured reserves: %f < %f",
+			toBTC(bal.Available), toBTC(reserves))
+		bal.Other["Reserves Deficit"] = reserves - bal.Available
```
ukane-philemon, line 1498: > "Reserves Deficit"

Can this be a global var? For translation purpose and whatnots.

client/asset/btc/btc.go
```diff
@@ -1458,9 +1480,49 @@ func (btc *baseWallet) Balance() (*asset.Balance, error) {
...
+		reserves = bal.Available
+	}
+	bal.Other["Bond Reserves (locked)"] = reserves
```
ukane-philemon, line 1501: same as my comment above. I don't know if we do the same for dcr atm but I think it's best to have them declared somewhere.

client/asset/btc/btc.go
```diff
@@ -1779,7 +1837,7 @@ func (btc *baseWallet) PreSwap(req *asset.PreSwapForm) (*asset.PreSwap, error) {
 	}
 
 	return &asset.PreSwap{
-		Estimate: est,
+		Estimate: est, // may be nil so we can present options, which in turn affect estimate feasibility
```
ukane-philemon, line 1840: Are other consumers aware of this change? (FE, etc)

client/asset/btc/btc.go
```diff
@@ -1955,6 +2013,20 @@ func (btc *baseWallet) estimateSwap(lots, lotSize, feeSuggestion, maxFeeRate uin
 		}
 	}
 
+	if sum > avail-reserves {
+		if trySplit {
+			return nil, false, 0, errors.New("eats bond reserves")
+		}
```
ukane-philemon, line 2019: We could use a more description error message.

server/auth/registrar.go
```diff
@@ -158,15 +154,11 @@ func (auth *AuthManager) handlePostBond(conn comms.Link, msg *msgjson.Message) *
 		return msgjson.NewError(msgjson.BondError, "error parsing postbond request")
 	}
 
-	// TODO: allow different assets for bond, switching parse functions, etc.
 	assetID := postBond.AssetID
 	bondAsset, ok := auth.bondAssets[assetID]
 	if !ok {
 		return msgjson.NewError(msgjson.BondError, "only DCR bonds supported presently")
```
ukane-philemon, line 160: Error message needs an update.

ukane-philemon (changes requested):
utAck

server/auth/registrar.go
```diff
@@ -123,9 +123,8 @@ func (auth *AuthManager) handlePreValidateBond(conn comms.Link, msg *msgjson.Mes
 		AssetID:   assetID,
 		Amount:    uint64(amt),
 		Expiry:    uint64(expireTime.Unix()),
-		BondID:    bondCoinID,
 	}
-	auth.Sign(preBondRes)
+	preBondRes.SetSig(auth.SignMsg(append(preBondRes.Serialize(), preBond.RawTx...)))
```
chappjc, line 127: Just linking back to our previous discussion of the options a https://github.com/decred/dcrdex/pull/2196#discussion_r1122413838
The approach implicitly hashes the RawTx but the signature's message is a little non-intuitive, whereas the suggestion from @martonp in the discussion with to put an explicit hash of this sent tx in the response payload.  I preferred to omit it from the payload, but wanted to doc here.

client/asset/btc/btc.go
chappjc: When making the same change for DCR recently, the `setFeeEstimates` function for the markets page was updated: https://github.com/decred/dcrdex/pull/2103/files#diff-1e4f8de7bbcd6448186a65982ea5e8fe3d7b64b982ed09520ff884829db5221aR1806
I didn't come across any other places to change at the time.

client/asset/btc/electrum_client.go
```diff
@@ -1111,179 +1111,6 @@ func (ew *electrumWallet) tryRemoveLocalTx(ctx context.Context, txid string) {
...
-	}
-	_, err = ew.wallet.Broadcast(ctx, txRaw)
-	if err != nil {
-		ew.tryRemoveLocalTx(ctx, msgTx.TxHash().String())
-		return nil, err
-	}
-
-	return txRaw, nil
-}
-
```
chappjc, line 1286: Looks like we can also remove three of the methods form the `electrumWalletClient interface`, but I kinda wanna keep the `WalletClient` methods in wallet_methods.go (it's the separate `btc/electrum` package, so it's more complete with those left).  You've not removed them, just noting that they'll look unused when the interface in this file is cleaned up.

chappjc:
New plan.  rc1 tomorrow, rc2 next week with this

client/asset/interface.go
```diff
@@ -784,11 +784,12 @@ type PeerManager interface {
...
 type Bond struct {
-	Version uint16
-	AssetID uint32
-	Amount  uint64
-	CoinID  []byte
-	Data    []byte // additional data to interpret the bond e.g. redeem script, bond contract, etc.
+	Version        uint16
+	AssetID        uint32
+	Amount         uint64
+	UnsignedCoinID []byte
```
chappjc, line 790: Can remove this field now.

dex/msgjson/types.go
```diff
@@ -1041,18 +1041,16 @@ type PreValidateBondResult struct {
 	AssetID   uint32 `json:"assetID"`
 	Amount    uint64 `json:"amount"`
 	Expiry    uint64 `json:"expiry"` // not locktime, but time when bond expires for dex
-	BondID    Bytes  `json:"bondID"`
```
chappjc, line 1044: Please doc on the `Signature` field that the preimage is the result serialization concatenated with the `RawTx` in the request.

server/asset/btc/btc.go
```diff
@@ -509,6 +514,173 @@ func (btc *Backend) VerifyUnspentCoin(_ context.Context, coinID []byte) error {
...
 
+// ParseBondTx performs basic validation of a serialized time-locked fidelity
+// bond transaction given the bond's P2SH or P2WSH redeem script.
+//
+// The transaction must have at least two outputs: out 0 pays to a P2SH address
+// (the bond), and out 1 is a nulldata output that commits to an account ID.
+// There may also be a change output.
+//
+// Returned: The bond's coin ID (i.e. encoded UTXO) of the bond output. The bond
+// output's amount and PWSH/P2WSH address. The lockTime and pubkey hash data pushes
```
chappjc, line 525: PWSH -> P2SH

server/asset/btc/btc.go
```diff
@@ -509,6 +514,173 @@ func (btc *Backend) VerifyUnspentCoin(_ context.Context, coinID []byte) error {
...
+		}
+		return
+	}
+	if txOut == nil { // spent == invalid bond
+		err = asset.CoinNotFoundError
+		return
+	}
+
+	_, amt, _, _, lockTime, acct, err = ParseBondTx(ver, rawTx, btc.chainParams)
+	return
```
chappjc, line 681: I feel like it should enforce segwit/non-segwit according to the `btc.segwit` flag.

client/asset/btc/btc.go
```diff
@@ -4739,6 +4925,474 @@ func (btc *intermediaryWallet) EstimateSendTxFee(address string, sendAmount, fee
...
+	msgTx.AddTxIn(txIn)
+
+	// Calculate fees and add the refund output.
+	var outputSize int
+	if btc.segwit {
+		outputSize = dexbtc.P2WPKHOutputSize
+	} else {
+		outputSize = dexbtc.P2PKHOutputSize
+	}
+	redeemSize := msgTx.SerializeSize() + dexbtc.RedeemBondSigScriptSize + outputSize
```
chappjc, line 5313: `dexbtc.RedeemBondSigScriptSize` does not apply to segwit, where the sigscript would be 1+scripthash, and the witness data would contain the signatures and script itself.  We'll have to do all the **vbytes** computations to be precise about this. `MsgTxVBytes` or like https://github.com/decred/dcrdex/blob/4d5c1cc7fa25b46bcb3c4565f69d25f93c0f8071/client/asset/btc/btc.go#L3550-L3558?

chappjc (changes requested):

server/asset/btc/btc.go
```diff
@@ -573,6 +573,10 @@ func ParseBondTx(ver uint16, rawTx []byte, chainParams *chaincfg.Params) (bondCo
 		err = fmt.Errorf("bad bond pkScript")
 		return
 	}
+	if !segwit && len(scriptHash) == 32 {
+		err = fmt.Errorf("%s backend does not support segwit bonds", chainParams.Name)
```
chappjc, line 577: Maybe the strictness is ultimately pointless, but I was thinking for a given bondver and assetID, we should expect either segwit or not segwit for consistency and to keep it standardized for a given asset.

```go
switch len(scriptHash) {
case 32:
// require segwit
case 20:
// require !segwit
}
```

chappjc (approved):
testnet working well too

`/api/updatebondoptions` to switch from DCR with target tier 3 to BTC with target tier 5:

```
2023-03-29 17:00:05.619 [INF] CORE: Attempting to update bond reserves by 0.08000000 BTC (0.00000000 BTC currently locked in bonds with 0.00000000 BTC in-force) for target tier change from 3 to 5, with bond increment 0.02000000 BTC
2023-03-29 17:00:05.622 [TRC] CORE[btc][SPV]: Bals: spendable = 14.05559527 BTC (14.05559527 BTC trusted, 0 BTC untrusted, 0 BTC assumed locked), immature = 0 BTC
2023-03-29 17:00:05.622 [TRC] CORE[btc]: ReserveBondFunds(0.08, true): enforced = 0.15556 / bonded = 0 / nominal = 0.12  ==>  enforced = 0.23556 / bonded = 0 / nominal = 0.2
2023-03-29 17:00:05.622 [INF] CORE: Total reserved for dextesttaggpdj27s26wklczemovhqdzyvgokyumey33uemlw4imawqd.onion:7232 is now 0.20000000 BTC (0.08000000 BTC more in future bonds)
2023-03-29 17:00:05.622 [DBG] CORE: Bond options for dextesttaggpdj27s26wklczemovhqdzyvgokyumey33uemlw4imawqd.onion:7232: target tier 5, bond asset 0, maxBonded 40000000
2023-03-29 17:00:05.629 [TRC] CORE[btc][SPV]: Bals: spendable = 14.05559527 BTC (14.05559527 BTC trusted, 0 BTC untrusted, 0 BTC assumed locked), immature = 0 BTC
2023/03/29 12:00:05 "POST http://127.0.0.2:5758/api/updatebondoptions HTTP/1.1" from 127.0.0.1:49192 - 200 12B in 21.99913ms
```

Side note: I think when switching to a new asset (BTC) with live bonds in the old asset (DCR), it's not reserving enough in the new asset for when the old asset's bonds refund, but that's not an issue with this PR.  It should re-reserve the right amounts on restart after the DCR bonds are refunded.

Shortly after updating bond options, `rotateBonds` triggers:

```
2023-03-29 17:00:22.918 [INF] CORE: Gotta post 1 bond increments now. Target tier 5, current tier 4 (0 weak, 0 pending)
2023-03-29 17:00:22.918 [TRC] CORE: Bond lifetime = 7h40m0s (lockTime = 2023-03-29 19:40:22 -0500 CDT)
2023-03-29 17:00:22.919 [TRC] CORE[btc]: feeRateWithFallback using caller's suggestion for fee rate, 1.
2023-03-29 17:00:22.921 [TRC] CORE[btc]: bondLocked (0.02): enforced 0.23556 ==> 0.21556 (with bonded = 0.02 / nominal = 0.2)
2023-03-29 17:00:22.921 [DBG] CORE[btc]: New bond reserves (new post) = 0.215560 BTC with 0.020000 in unspent bonds
2023-03-29 17:00:22.923 [DBG] CORE[btc]: Change output size = 31, addr = tb1q8c6kkz7234j2w90p0ldtkec9e4w9d8kvyk8rd2
2023-03-29 17:00:22.923 [DBG] CORE[btc]: 2 signature cycles to converge on fees for tx 2b1f8005471f6cfefd7b43b1b1cb598095c37b19de411d6e963f3ba1a0d7aed6: min rate = 1, actual fee rate = 1 (222 for 222 bytes), change = true
```

https://mempool.space/testnet/tx/2b1f8005471f6cfefd7b43b1b1cb598095c37b19de411d6e963f3ba1a0d7aed6

prevalidate:

```
2023-03-29 17:00:26.373 [INF] CORE: DEX dextesttaggpdj27s26wklczemovhqdzyvgokyumey33uemlw4imawqd.onion:7232 has validated our bond 2b1f8005471f6cfefd7b43b1b1cb598095c37b19de411d6e963f3ba1a0d7aed6:0 (btc) with strength 1. 1 confirmations required to trade.
2023-03-29 17:00:26.374 [INF] CORE: Broadcasting bond 2b1f8005471f6cfefd7b43b1b1cb598095c37b19de411d6e963f3ba1a0d7aed6:0 (btc) with lock time 2023-03-29 19:40:22 -0500 CDT, data = 0476da2464b17576a914e82704b22735f58423bbb881abb1226c36bcc18188ac.

BACKUP refund tx paying to current wallet: 01000000000101d6...
...
2023-03-29 17:00:28.374 [DBG] CORE[btc][SPV]: No error from PublishTransaction after 2s for txn 2b1f8005471f6cfefd7b43b1b1cb598095c37b19de411d6e963f3ba1a0d7aed6. Assuming wallet accepted it.
2023-03-29 17:00:28.378 [TRC] CORE[btc][SPV]: Bals: spendable = 14.03559305 BTC (14.03559305 BTC trusted, 0 BTC untrusted, 0 BTC assumed locked), immature = 0 BTC
2023-03-29 17:00:28.380 [INF] CORE: notify: |SUCCESS| (bondpost) BondConfirming - Waiting for 1 confirmations to post bond 2b1f8005471f6cfefd7b43b1b1cb598095c37b19de411d6e963f3ba1a0d7aed6:0 (btc) to dextesttaggpdj27s26wklczemovhqdzyvgokyumey33uemlw4imawqd.onion:7232
```

confirmed and accepted:

```
2023-03-29 17:14:30.135 [INF] CORE: DEX dextesttaggpdj27s26wklczemovhqdzyvgokyumey33uemlw4imawqd.onion:7232 bond txn 2b1f8005471f6cfefd7b43b1b1cb598095c37b19de411d6e963f3ba1a0d7aed6:0 now has 1 confirmations. Submitting postbond request...
2023-03-29 17:14:30.769 [INF] CORE: Bond confirmed 2b1f8005471f6cfefd7b43b1b1cb598095c37b19de411d6e963f3ba1a0d7aed6:0 (btc) with expire time of 2023-03-29 16:40:22 -0500 CDT
2023-03-29 17:14:30.770 [INF] CORE: Bond 2b1f8005471f6cfefd7b43b1b1cb598095c37b19de411d6e963f3ba1a0d7aed6:0 (btc) confirmed.
2023-03-29 17:14:30.770 [INF] CORE: notify: |SUCCESS| (bondpost) BondConfirmed - New tier = 5 (target = 5).
```

Refund will happen tonight.

chappjc:
Refund happened fine, as expected: https://mempool.space/testnet/tx/90e8e134493666073bf348d1c3fea8423c5c21661be063c7e5cd1441de1088a7#vin=0

We just need to confident that all the sends that now no longer use the `sendtoaddress` RPC are still gonna work in all cases.

chappjc:
...and this will make the cut for 0.6.  On that note, I don't see why #1993 shouldn't land on 0.6 either.

PR #2199, opened 2023-03-02
server/eth: RPC client priority

### LINKED ISSUE #2161: server/eth: Prioritize RPC connections
The rpc providers file should allow setting the order of priority for connections.

martonp:
This change allows server operators to rank RPC entpoints by priority. The healthy connection with the highest priority will always be attempted first. If no priority is specified for an endpoint, 0 is assumed.

Closes #2161

### DIFF
--- a/server/asset/eth/eth.go
+++ b/server/asset/eth/eth.go
@@ -284,16 +284,19 @@ func NewBackend(configPath string, log dex.Logger, net dex.Network) (*ETHBackend
 		if line == "" || strings.HasPrefix(line, "#") || strings.HasPrefix(line, ";") {
 			continue
 		}
+		ethCfgInstructions := "invalid eth config line: \"%s\". " +
+			"Each line must contain URL and optionally a priority (between 0-65535) " +
+			"separated by a comma. Example: \"https://www.infura.io/,2\""
 		parts := strings.Split(line, ",")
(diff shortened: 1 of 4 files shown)


server/asset/eth/eth.go
```diff
@@ -275,16 +276,35 @@ func NewBackend(configPath string, log dex.Logger, net dex.Network) (*ETHBackend
...
+			continue
+		}
+		parts := strings.Split(line, ",")
+		if len(parts) < 1 || len(parts) > 2 {
+			return nil, fmt.Errorf("invalid eth config line %q", line)
+		}
+		url := strings.TrimSpace(parts[0])
+		var priority uint16
+		if len(parts) == 2 {
+			priority64, err := strconv.ParseUint(strings.TrimSpace(parts[1]), 10, 16)
```
chappjc, line 294: `strconv.ParseUint: parsing "226135": value out of range`
Borderline clear enough, but maybe have the error say the expected range or max.

chappjc (approved):
Untested but LGTM.  Just a thought about clarifying the allowable prio values, at least in the error message.

server/asset/eth/eth.go
chappjc: Or allow mega numbers?

chappjc:
On dex-test now

PR #2201, opened 2023-03-03
client/core: Update token parent balances

martonp:
When an order involving a token was cancelled, the locked amount in the parent's balance would not be updated. Now, in `updateBalances`, when a token's balance is update is parents is also updated.

This also moves the updates for the parent asset's balances from `sendTradeRequest` to `prepareTradeRequest`.


JoeGruffins (approved):
Makes sense.

chappjc (approved):

PR #2202, opened 2023-03-04
remove deprecated linter: structcheck

norwnd:
Removes linter warning:

```
WARN [runner] The linter 'structcheck' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter. Replaced by unused.
```


chappjc:
It's deprecated, not useless.
We'll remove it when we get around to cleaning up the code to the point where we can use the `unused` linter.  I will do that careful, please refrain

PR #2204, opened 2023-03-04
remove unnecessary build step

norwnd:
Looks like this build step is no longer needed/relevant ? If it does serve some implicit purpose I'm not aware of, I'd like to document it with a comment here instead.


chappjc (approved):
Git blame makes it clear.
I've removed this in another branch but fine

PR #2205, opened 2023-03-04
client/db clarify DB timeout error

norwnd:
Pretty much a nit, but could help a not-very-technical user administering `dexc`, because currently when I have 1 instance running, starting another gives me this error:
```
error creating client core: database initialization error: timeout
```
it becomes this instead:
```
error creating client core: database initialization error: timeout, could happen when database is already being used by another process
```


chappjc (approved):
Ok, but it's only one of many possible causes, so potentially misleading.  It's the most common case, so fine

JoeGruffins (approved):

PR #2207, opened 2023-03-04
ui: token wallet is missing type

norwnd:
I'm reaching into "unknown" area here a bit, hope this doesn't rub you in wrong way (don't want to step on someone's toes if it's already being worked on, still, I would like to figure out how it works for myself anyway).

Currently **Token** wallet displays its `Wallet Type` as empty. This is either unnecessary to see for the user, or it could be a bug (and this PR treats it this way),

before / after:

imageimage

Also tested that ETH/USDC wallet creation isn't broken with this change, however affected code is a bit hairy and I might have missed something.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -2568,7 +2568,7 @@ func (c *Core) createTokenWallet(tokenID uint32, token *asset.Token, form *Walle
 	}
 
 	return &db.Wallet{
-		Type:     "token",
+		Type:     form.Type,
 		AssetID:  tokenID,
 		Settings: form.Config,
 		// EncryptedPW ignored because we assume throughout that token wallet
--- a/client/webserver/site/src/js/forms.ts
+++ b/client/webserver/site/src/js/forms.ts
@@ -155,11 +155,7 @@ export class NewWalletForm {
     let parentForm
     let walletType = selectedDef.type
     if (parentAsset) {
-      if (!asset.token) {
-        this.setError('token is absent from asset with parent, should never happen!')
-        return
-      }
-      walletType = asset.token.definition.type
+      walletType = (asset.token as Token).definition.type
       parentForm = {
         assetID: parentAsset.id,
         config: this.subform.map(parentAsset.id),
--- a/client/webserver/site/src/js/forms.ts
+++ b/client/webserver/site/src/js/forms.ts
@@ -153,7 +153,13 @@ export class NewWalletForm {
     const { asset, parentAsset } = this.current
     const selectedDef = this.current.selectedDef
     let parentForm
+    let walletType = selectedDef.type
     if (parentAsset) {
+      if (!asset.token) {
+        this.setError('token is absent from asset with parent, should never happen!')
+        return
+      }
+      walletType = asset.token.definition.type
       parentForm = {
         assetID: parentAsset.id,
         config: this.subform.map(parentAsset.id),
@@ -161,7 +167,7 @@ export class NewWalletForm {
       }
     }
     // Register the selected asset.
-    const res = await this.createWallet(asset.id, selectedDef.type, pw, parentForm)
+    const res = await this.createWallet(asset.id, walletType, pw, parentForm)
     if (!app().checkResponse(res)) {
       this.setError(res.msg)
       return
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -50,7 +50,8 @@ func registerToken(tokenID uint32, desc string, nets ...dex.Network) {
 		panic("token " + strconv.Itoa(int(tokenID)) + " not known")
 	}
 	asset.RegisterToken(tokenID, token.Token, &asset.WalletDefinition{
-		Type:        "token",
+		Type:        walletTypeToken,
+		Tab:         "Token",
 		Description: desc,
 	}, nets...)
 }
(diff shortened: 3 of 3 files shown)


client/core/core.go
```diff
@@ -2568,7 +2568,7 @@ func (c *Core) createTokenWallet(tokenID uint32, token *asset.Token, form *Walle
 	}
 
 	return &db.Wallet{
-		Type:     "token",
+		Type:     form.Type,
```
norwnd, line 2571: This makes it a bit more consistent with how other wallets are created, I think.

chappjc:
Very sensible indeed.

I've also been tempted to change `walletTypeRPC` from `"External"` to `"RPC"`.  I think external has previously implied the entire wallet outside like dcrwallet or Electrum.  Native arguably could still work, but with RPC providers it's a stretch.

JoeGruffins (commented):
Are the previous types Native or External? I feel like that was the important part, whether the wallet was a separate process. Can the wallet type we show include Native or External? For example, show `Token/External` for the eth token wallets, since they are both.

chappjc:
> Are the previous types Native or External?

Ethereum was an is still **External**.  Tokens were an empty string.

>  I feel like that was the important part, whether the wallet was a separate process

That's my feeling too.  Ethereum's `walletTypeRPC` wallet definition should be changed by this logic.  The wallet is not a separate process, although all chain data comes from an RPC provider.

I'm still a little uncertain about what to show for tokens, but any thing is better than an empty string.

norwnd:
> I'm still a little uncertain about what to show for tokens, but any thing is better than an empty string.

Btw, a typical user is probably not interested in details that low level, those who are interested will find this info under "Reconfigure wallet settings", so maybe just remove it altogether from that view ? This info there isn't actionable anyway.

chappjc:
> > I'm still a little uncertain about what to show for tokens, but any thing is better than an empty string.
> 
> Btw, a typical user is probably not interested in details that low level, those who are interested will find this info under "Reconfigure wallet settings", so maybe just remove it altogether from that view ? This info there isn't actionable anyway.

Let's do it.

norwnd:
I've added commit on top to remove "Wallet Type" from that view on screenshots above:

image

leaving `walletTypeToken` type related changes (1st commit) to differentiate token wallets because we'll probably need to come back to it later with a better definition,  once we have it.

chappjc:
Eh, I was thinking to remove Wallet Type only if it is just an empty string, but now that I see it with it removed entirely from the view, I kinda see it's _perhaps_ best to remove it like you did.  You can click the Settings button to reconfigure the wallet.
Let's see what the others think.

JoeGruffins:
> those who are interested will find this info under "Reconfigure wallet settings"

I don't see this anywhere. Do you mean infer the type of wallet by the form? Or I guess in the "change the wallet type" collapsed area? I think it would be good to have the type somewhere obvious in case of a support scenario where the person giving help needs to know what type of wallet they are using.

I don't feel super strong about this and I don't hate these changes so approving.

JoeGruffins (approved):

norwnd:
> Or I guess in the "change the wallet type" collapsed area?

This.

martonp (commented):
Why should this be removed? It can just be `${parentWalletType} / token` for tokens. I guess it's not crucial for it to be there, but I don't see any reason to remove it either. It's nice to look at it sometimes especially when doing development.

norwnd:
> I guess it's not crucial for it to be there, but I don't see any reason to remove it either.

My motivation (same as for hiding "funding coins") to remove comes from the idea of reducing cognitive load the user will experience to the lowest possible level. It isn't all black and white but typically devs / tech people have bias towards optionality/customizability which average folks don't appreciate.

> It can just be ${parentWalletType} / token for tokens.

Yeah, that's better than just Token but **might not fit** (need to keep an eye out for  that) into space allocated for it. I guess this is the next best option I can try out.

norwnd:
> it's not an "External" wallet.

Regarding **chappjc** comment ^, should it be "Native" then ?

buck54321:
The ethereum wallet type can be called "RPC". I'm still holding out hope a p2p light wallet which we'll call "Native". Token wallet type can just be "token". Doesn't need the parent type specifier. I don't have a strong opinion whether we show the wallet type for tokens in the wallets view or not. Though it might make sense to call it e.g. "Ethereum token".

norwnd:
Popped off the latest commit to keep showing `Wallet Type` in UI (as per feedback above), and renamed as **buck54321** [suggested](https://github.com/decred/dcrdex/pull/2207#issuecomment-1482813590).

imageimage

client/webserver/site/src/js/forms.ts
```diff
@@ -153,15 +153,21 @@ export class NewWalletForm {
     const { asset, parentAsset } = this.current
     const selectedDef = this.current.selectedDef
     let parentForm
+    let walletType = selectedDef.type
     if (parentAsset) {
+      if (!asset.token) {
+        this.setError('token is absent from asset with parent, should never happen!')
+        return
+      }
```
buck54321, line 161: Like you said, it should never happen. We don't need defensive code for this. If you're just trying to deal with TypeScript complaining about the optional field, you can do `if (parentAsset && asset.token)`

buck54321 (changes requested):

client/webserver/site/src/js/forms.ts
chappjc: Maybe just do `walletType = (asset.token as Token).definition.type` then?

buck54321: > Maybe just do `walletType = (asset.token as Token).definition.type` then?

That's probably better.

buck54321 (approved):

PR #2208, opened 2023-03-07
client/core,app: Fix rate/volume decimal precision

### LINKED ISSUE #2174: ui: trim Rate precision to Market rate step
There are multiple UI elements displaying Rate/Price with higher digit precision than Market actually supports (seemingly for no good reason, and at the expense of UI bloat), for example:

imageimageimage

Among Markets I tested (DCR-ETH, DCR-USDC, DCR-BTC) it seems to only affect DCR-ETH one.

martonp:
Closes #2174

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -319,7 +319,7 @@ func coreMarketFromMsgMarket(dc *dexConnection, msgMkt *msgjson.Market) *Market 
 		EpochLen:        msgMkt.EpochLen,
 		StartEpoch:      msgMkt.StartEpoch,
 		MarketBuyBuffer: msgMkt.MarketBuyBuffer,
-		AtomToConv:      float64(qconv) / float64(bconv),
+		AtomToConv:      float64(bconv) / float64(qconv),
 	}
 
 	trades, inFlight := dc.marketTrades(mkt.marketName())
--- a/client/webserver/site/src/js/wallets.ts
+++ b/client/webserver/site/src/js/wallets.ts
@@ -1299,13 +1299,12 @@ function assetIsConfigurable (assetID: number) {
 }
 
 const FourSigFigs = new Intl.NumberFormat((navigator.languages as string[]), {
-  minimumSignificantDigits: 4,
   maximumSignificantDigits: 4
 })
 
 const intFormatter = new Intl.NumberFormat((navigator.languages as string[]))
 
 function fourSigFigs (v: number): string {
-  if (v < 100) return FourSigFigs.format(v)
+  if (v < 1000) return FourSigFigs.format(v)
   return intFormatter.format(Math.round(v))
 }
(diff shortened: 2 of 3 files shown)


client/webserver/site/src/js/wallets.ts
```diff
@@ -743,10 +742,15 @@ export default class WalletsPage extends BasePage {
 
       if (spot) {
         const convRate = app().conventionalRate(baseid, quoteid, spot.rate, exchanges[host])
-        tmpl.price.textContent = fourSigFigs(convRate)
+        tmpl.price.textContent = withNumDecimals(convRate, mkt.convRateNumDecimals)
```
buck54321, line 745: I think the point is to lop off extra zeros in the decimals. In many cases, this will force the showing of decimal places we don't care about. For example, if the current rate is `1,284.0`, and the rate step is `1e4`, this will show `1,284.0000`, which is clearly not what we want. Sig figs is the way to go. We just need to clean up the max number of decimals too.

martonp: You're right, no need for exact values.

JoeGruffins (approved):

client/core/core.go
```diff
@@ -309,17 +342,19 @@ func coreMarketFromMsgMarket(dc *dexConnection, msgMkt *msgjson.Market) *Market
...
+		BaseID:                base.ID,
+		BaseSymbol:            base.Symbol,
+		QuoteID:               quote.ID,
+		QuoteSymbol:           quote.Symbol,
+		LotSize:               msgMkt.LotSize,
+		RateStep:              msgMkt.RateStep,
+		EpochLen:              msgMkt.EpochLen,
+		StartEpoch:            msgMkt.StartEpoch,
+		MarketBuyBuffer:       msgMkt.MarketBuyBuffer,
+		AtomToConv:            float64(bconv) / float64(qconv),
```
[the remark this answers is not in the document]
chappjc: Not incorrect?

client/webserver/site/src/js/wallets.ts
chappjc: Isn't the intent then to be using {min,max}imum**Fraction**Digits, not {min,max}imum**Significant**Digits?

Also, regardless of value, I feel like we'd want at least one decimal e.g. `1,284.0` is Buck's example.

chappjc: For instance:

```js
const FourSigFigsSmall = new Intl.NumberFormat('en-US', {
  minimumFractionDigits: 1,
  maximumSignificantDigits: 5
})

const FourSigFigsBig = new Intl.NumberFormat('en-US', {
  minimumFractionDigits: 1,
  maximumFractionDigits: 1
})

function fourSigFigs (v) {
  if (v<1000) return FourSigFigsSmall.format(v)
  return FourSigFigsBig.format(v)
}

console.log(fourSigFigs(123456.789))
console.log(fourSigFigs(1234.789))
console.log(fourSigFigs(123.789))
console.log(fourSigFigs(0.01357))
```

Output:

```
> "123,456.8"
> "1,234.8"
> "123.79"
> "0.01357"
```

PR #2209, opened 2023-03-07
client/core: Do not disconnect token wallets on shutdown

### LINKED ISSUE #2058: Token wallet error on shutdown
I see these logs on shutdown: 

```
2023-01-21 16:30:55.003 [INF] CORE: Locking USDC.ETH wallet
2023-01-21 16:30:55.003 [ERR] CORE: Failed to lock USDC.ETH wallet: wallet not connected
```

martonp:
This was causing an unnecessary error message, and there is no need for this as disconnecting the parent wallet is sufficient.

Closes #2058

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -1674,7 +1674,7 @@ func (c *Core) Run(ctx context.Context) {
 	defer c.walletMtx.Unlock()
 	for assetID, wallet := range c.wallets {
 		delete(c.wallets, assetID)
-		if !wallet.connected() {
+		if !wallet.connected() || wallet.parent != nil {
 			continue
 		}
 		if !c.cfg.NoAutoWalletLock {


ukane-philemon (approved):

client/core/core.go
```diff
@@ -1674,7 +1674,7 @@ func (c *Core) Run(ctx context.Context) {
 	defer c.walletMtx.Unlock()
 	for assetID, wallet := range c.wallets {
 		delete(c.wallets, assetID)
-		if !wallet.connected() {
+		if !wallet.connected() || wallet.parent != nil {
```
buck54321, line 1677: If I'm reading https://github.com/decred/dcrdex/issues/2058#issuecomment-1399829244 correctly, we should actually be deleting this block and moving the `wallet.parent != nil` check to `Disconnect`.

PR #2210, opened 2023-03-08
server/orderbook: consider atomic seq updates with respect to book contents

norwnd:
`seq` updates on server for `msgBook` are not atomic with respect to order book contents, this means 2 clients (or even 1 client sending 2 requests in parallel) can see different order book contents for the same `seq` number (e.g. one order book snapshot already including last **book_order**/**update_remaining**/**unbook_order** update effects, while the other one lacking those because that update technically belongs to the next `seq`)

which, in turn, means:
1) client has to handle concurrent subscribe/resubscribe requests with extra care to avoid applying the effects of the same update tiwice (e.g. client got snapshot with latest update applied already, but he also can get same update notification with `seq+1`  afterwards); currently **dexc** resolves this by ignoring such redundant update notification ? It also prevents resubscribe running concurrent subscribe/resubscribe by locking mutex 
2) if somebody takes periodic server order book snapshots, and say counts tracks (how much orders there is, for debug purposes or whatnot) for each seq order book update then he can't rely on this giving him exact results

It's not an issue per se? It's a tradeoff that leaves more room for shooting yourself in the foot for seemingly no good reason.

This PR adds aforementioned atomicity property.

### DIFF
--- a/server/market/bookrouter.go
+++ b/server/market/bookrouter.go
@@ -150,7 +150,6 @@ type BookSource interface {
 type subscribers struct {
 	mtx   sync.RWMutex
 	conns map[uint64]comms.Link
-	seq   uint64
 }
 
 // add adds a new subscriber.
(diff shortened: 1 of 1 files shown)


norwnd:
Whoops, sorry I meant to open this one against my repo-fork and maybe mention it some time later.

PR #2216, opened 2023-03-12
ui: funding coins hidden by default under offering section

norwnd:
Completes 2nd part and resolves https://github.com/decred/dcrdex/issues/2175.

We can get the best of both: still having info about funding coins, not overwhelming users with that

[funding-coins-under-offering.webm](https://user-images.githubusercontent.com/112318969/224531320-cf6030b6-5ca9-440e-ad75-5cffccf58c5f.webm)

-----
I don't think we need to implement a switch-back **funding coins** -> **offering amount** since page refresh does that.

Regarding that tool tip for **funding coins** I suggested in #2175, since it's an advanced and hidden feature now (with the changes in this PR) I no longer think it's something we want/need to add there.

### DIFF
--- a/client/webserver/site/src/css/order.scss
+++ b/client/webserver/site/src/css/order.scss
@@ -8,7 +8,7 @@ div.order-datum {
   font-size: 15px;
   margin: 5px 10px 5px 0;
 
-  div:first-child {
+  .order-datum-hdr {
     padding: 2px 15px;
     border-bottom: 1px solid #777a;
     font-family: $demi-sans;
--- a/client/webserver/locales/ar.go
+++ b/client/webserver/locales/ar.go
@@ -124,7 +124,6 @@ var Ar = map[string]string{
 	"them":                      "هم",
 	"Redemption":                "اِستِردَاد",
 	"Refund":                    "المبالغ المردودة",
-	"Funding Coins":             "عملات التمويل",
 	"Exchanges":                 "منصات التداول",
 	"apply":                     "طبق",
 	"Assets":                    "الأصول",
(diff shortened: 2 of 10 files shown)


client/webserver/site/src/css/order.scss
```diff
@@ -8,7 +8,7 @@ div.order-datum {
   font-size: 15px;
   margin: 5px 10px 5px 0;
 
-  div:first-child {
+  .order-datum-hdr {
```
norwnd, line 11: This caused over-selection for elements under those with `order-datum` class label on `/orders/{id}` page:

image

So, changing here, but could also be moved to a separate PR of its own (just don't think it's worth bothering with when changes aren't huge).

chappjc:
Clicking "Offering" is sensible, but I don't love it modifying the Offering box contents.  I'd rather it just _unhide_ the previous Funding Coins div.

Any other opinions from the rest of the team?

In any case, we don't need to modify this now, certainly won't for 0.6.

JoeGruffins:
I can't say I like this change very much. Maybe just my opinion. I think clicking `offering` is confusing. You wouldn't know it's there, then you can't close it. I don't know why the information at the bottom was a problem.

I have used them often. Also consider the support situation, where the helper needs to know everything about the transaction. Having the info in the ui would be better than digging though logs in that scenario. You could argue it's likely they will need to dig through logs anyway, but I think it depends on the problem. I haven't been paying much attention to the support channel, has this info been useful so far anyone that knows?

chappjc:
> I can't say I like this change very much. Maybe just my opinion. I think clicking `offering` is confusing. You wouldn't know it's there, then you can't close it. I don't know why the information at the bottom was a problem.
> 
> I have used them often. Also consider the support situation, where the helper needs to know everything about the transaction. Having the info in the ui would be better than digging though logs in that scenario. You could argue it's likely they will need to dig through logs anyway, but I think it depends on the problem. I haven't been paying much attention to the support channel, has this info been useful so far anyone that knows?

I have the same point of view.  Not sure https://github.com/decred/dcrdex/issues/2175 is an issue to resolve at all.

chappjc:
Closing as not planned.

Issue #2184, opened 2023-02-28
client/core: suspicious assertion failure in updateBalances


norwnd:
While observing/debugging some weirdness with ETH balances (~~more on that later~~ see https://github.com/decred/dcrdex/issues/2185, https://github.com/decred/dcrdex/issues/2186, https://github.com/decred/dcrdex/issues/2187) on **testnet**, latest **master** I noticed this statement being logged:

https://github.com/decred/dcrdex/blob/80b0531a64a806ac8901d812b1e322418118cac1/client/core/core.go#L2105-L2108

Probably something to investigate, attaching full **dexc** log: 
[suspicious-assertion-fail.txt](https://github.com/decred/dcrdex/files/10847859/suspicious-assertion-fail.txt)

norwnd:
I looked but didn't find anything suspicious related to this assertion, attaching a [PR to log it as `Trace`](https://github.com/decred/dcrdex/pull/2218) instead. 

I did find [something interesting in the area nearby](https://github.com/decred/dcrdex/issues/2076#issuecomment-1465916999) though, might want to take a look at that while on the same topic.

norwnd:
I also seen somewhat similar `Error` in the logs which doesn't seem to be critical/error actually (at least from `CORE` point of view) because **dexc** continued to work fine regardless:

```
2023-03-19 08:23:30.495 [DBG] CORE[eth][RPC]: couldn't get a websocket connection for "wss://goerli.infura.io/ws/v3/e068872d3aa044aebbcf96f8e31104cc" (original scheme: "https") (OK)
2023-03-19 08:23:30.495 [ERR] CORE[eth][RPC]: Failed to get chain ID from "https://goerli.infura.io/v3/e068872d3aa044aebbcf96f8e31104cc": Post "https://goerli.infura.io/v3/e068872d3aa044aebbcf96f8e31104cc": context deadline exceeded
2023-03-19 08:23:30.495 [ERR] CORE: Unable to connect to eth wallet (start and sync wallets BEFORE starting dex!): connectWallet: failed to connect eth wallet: connect failure: failed to connect
2023-03-19 08:23:30.543 [ERR] CORE: notify: |ERROR| (walletconfig) Wallet connection warning - connectWallet: failed to connect eth wallet: connect failure: failed to connect
```

JoeGruffins:
[...]

PR #2218, opened 2023-03-14
client/core: investigate failing assertion

norwnd:
Resolves: https://github.com/decred/dcrdex/issues/2184

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -2102,9 +2102,9 @@ func (c *Core) updateBalances(assets assetMap) {
 	for assetID := range assets {
 		w, exists := c.wallet(assetID)
 		if !exists {
-			// This should never be the case, but log an error in case I'm
-			// wrong or something changes.
-			c.log.Errorf("non-existent %d wallet should exist", assetID)
+			// Could happen if we are here when handling "tip change" before Core
+			// managed to initialize its wallets map (or already removed it from it).
+			c.log.Tracef("can't update %q balance, Core isn't tracking this wallet yet/already", unbip(assetID))
 			continue
 		}
 		_, err := c.updateWalletBalance(w)


client/core/core.go
```diff
@@ -2102,9 +2102,9 @@ func (c *Core) updateBalances(assets assetMap) {
 	for assetID := range assets {
 		w, exists := c.wallet(assetID)
 		if !exists {
-			// This should never be the case, but log an error in case I'm
-			// wrong or something changes.
-			c.log.Errorf("non-existent %d wallet should exist", assetID)
+			// Could happen if we are here when handling "tip change" before Core
+			// managed to initialize its wallets map (or already removed it from it).
+			c.log.Tracef("can't update %q balance, Core isn't tracking this wallet yet/already", unbip(assetID))
```
chappjc, line 2107: Arguably papering over the issue, but OK.

"already" -> "still"?

norwnd: No rush to merge it, we **could** to document the current understanding until someone familiar with the code has time to address the issue properly (I'd love that too ofc).

JoeGruffins:
So we expect this is getting called before the wallet is created. I assume the newly created wallet would not need to update the balance. This change seems reasonable to me. It could be smarter but I don't know if that would be worth it.

JoeGruffins (approved):

JoeGruffins:
Or, `tipChange` in core could check to see if the wallet is in the map first and not do `updateBalance` on that asset if it is not, then the error log would still be true by our understanding.

client/core/core.go
```diff
@@ -2102,9 +2102,9 @@ func (c *Core) updateBalances(assets assetMap) {
 	for assetID := range assets {
 		w, exists := c.wallet(assetID)
 		if !exists {
-			// This should never be the case, but log an error in case I'm
-			// wrong or something changes.
-			c.log.Errorf("non-existent %d wallet should exist", assetID)
+			// Could happen if we are here when handling "tip change" before Core
+			// managed to initialize its wallets map (or removed this wallet already).
+			c.log.Tracef("can't update %q balance, Core isn't tracking this wallet yet/still", unbip(assetID))
```
buck54321, line 2107: I'd like to better understand the circumstances under which this can happen? It doesn't see like it's possible during core initialization. Does this happen during new wallet creation?

chappjc: Can also happen on shutdown it seems:

```
2023-03-29 02:44:39.572 [INF] CORE[dcr][SPV]: Unloading wallet
2023-03-29 02:44:39.573 [INF] CORE[dcr][SPV]: SPV wallet closed
2023-03-29 02:44:39.573 [INF] CORE: Locking ETH wallet
2023-03-29 02:44:39.633 [INF] CORE[eth][TXDB]: Got compaction priority: {level:0 score:1.73 dropPrefixes:[]}
2023-03-29 02:44:39.634 [INF] CORE: Locking USDC.ETH wallet
2023-03-29 02:44:39.634 [ERR] CORE: Failed to lock USDC.ETH wallet: wallet not connected
2023-03-29 02:44:39.634 [INF] CORE: DEX client core off
2023-03-29 02:44:39.634 [INF] DEXC: Exiting dexc main.
2023-03-29 02:44:39.634 [ERR] CORE: non-existent 60 wallet should exist
2023-03-29 02:44:39.634 [ERR] CORE: non-existent 60001 wallet should exist
2023-03-29 02:44:39.634 [ERR] CORE: non-existent 2 wallet should exist
2023-03-29 02:44:39.634 [ERR] CORE: non-existent 20 wallet should exist
```

chappjc:
Should convert to issue so we can properly characterize what if anything needs to change.

PR #2222, opened 2023-03-14
client/core: Unlock wallet during PostBond

martonp:
Instead of doing a `refreshUnlock` do an `unlockWallet` when calling `PostBond`.

It is very frustrating to have to keep unlocking during development. Probably not a huge issue during regular use, but definitely more convenient so why not?

### DIFF
--- a/client/core/bond.go
+++ b/client/core/bond.go
@@ -973,21 +973,21 @@ func (c *Core) PostBond(form *PostBondForm) (*PostBondResult, error) {
 	if _, ok := wallet.Wallet.(asset.Bonder); !ok { // will fail in MakeBondTx, but assert early
 		return nil, fmt.Errorf("wallet %v is not an asset.Bonder", bondAssetSymbol)
 	}
-	_, err = wallet.refreshUnlock()
-	if err != nil {
-		// TODO: Unlock with form.AppPass?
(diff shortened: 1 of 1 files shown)


chappjc:
It's good to have this separate, but I had this with a few other changes in #2200.  https://github.com/decred/dcrdex/pull/2200/commits/08e35a7cf73bc1e0a6d13592bafb55f873b04946
It's still in draft, but pls have a look when you get a chance.

client/core/bond.go
```diff
@@ -973,21 +973,21 @@ func (c *Core) PostBond(form *PostBondForm) (*PostBondResult, error) {
...
+
+	err = c.unlockWallet(crypter, wallet)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unlock bond asset wallet %v", unbip(bondAssetID))
+	}
+	if !wallet.synchronized() { // otherwise we might double spend if the wallet keys were used elsewhere
+		return nil, fmt.Errorf("wallet %v is not synchronized", unbip(bondAssetID))
+	}
+
 	if form.Addr == "" {
```
chappjc, line 991: Part of what I was trying to do in my alternative commit was to have the cheap and non-mutating checks before unlocking

chappjc, line 991: Part of what I was trying to do in my alternative commit was to have the cheap and non-mutating checks before unlocking

PR #2223, opened 2023-03-14
multi: ETH Fidelity Bonds

martonp:
Still a few small things to fix, but it's ready for a review. There's a 13k line package.lock file included in this commit. 

```
Add ETHBondV0.sol contract

This adds a solidity contract for managing ETH fidelity bonds. In addition
to the usual functionalities of creating and refunding bonds, it also
allows using funds locked in existing bonds to create a new bond. If
a bond with more funds than the existing bonds is being created, the user
has to send additional funds, and if a bond is being created with less
funds than the existing bonds, a "change" bond is also created with the
remainder that maintains the earliest lockTime of the existing bonds. This
process will only allow the user to lock funds for a longer period than
they were previously locked for, never less.

A Hardhat project is also created for this contract to allow easy testing
of the smart contract logic. To run the tests, go to the
`dex/networks/eth/bondcontracts/v0` folder and run the following:
`npm install`
`npx harhdat test`
```

```
client/asset/eth: Add BondUpdater interface and implement it in ETH Client

Adds a BondUpdater interface which contains a single method, `UpdateBondsTx`.
Similar to MakeBondTx, this generates an unsigned transaction, but
instead of creating a new bond from scratch, it uses the funds locked
in existing bonds to create a new bond.

For ETH, it is not possible to create a signed transaction that will
be submitted later because if the nonce is used before the transaction
is submitted, the transaction will no longer be valid. Therefore, for ETH,
the SignedTx and UnsignedTx in the `asset.Bond` returned from `MakeBondTx`
and `UpdateBondsTx` contain the same data. The ETH wallet backend populates
the nonce and signs the transaction on the call to `SendTransaction`.
```

```
client/core: Support wallets that implement 'BondUpdater`

This updates the client to support wallets that implement `BondUpdater`.
In `rotateBonds`, if an account is set to maintain a certain tier, and it
already has some weak or expired bonds posted in an asset that supports
updating bonds, it will use the funds in those bonds to create a new bond
with a sufficiently long lockTime.
```

```
server/asset/eth: Add BondUpdater interface and implement in ETH

This adds a BondUpdater interface in the server that requires a backend
to implement an `AllAccountBonds` method. in addition to the Bonder
interface's methods. This method returns all the currenlty active bonds
for an account. This new interface is implemented by the ETH backend. This
is required for assets where it is possible to use existing unexpired
bonds to fund new bonds, because the server needs to know if bonds that it
was previously using the calculates a user's tier no longer exist.
```

```
server/{auth,db}: Update server to support BondUpdaters

When a new bond is posted for an asset that implements `BondUpdater`,
all of the currently active bonds for that asset are retrieved from
the blockchain, and the the bonds for this asset that were stored in
the database and the `AuthManager` are replaced by what was retreived.
```


dex/networks/eth/bond_test.go
```diff
@@ -0,0 +1,56 @@
+//go:build lgpl
```
chappjc, line 1: Should remove these build tags now

dex/networks/eth/bondcontracts/ETHBondV0.sol
buck54321, line 1: Interestingly, with eth, we have the ability to "renew" a bond, negating the need to lock up > 1 bond amount. We could easily add a method to extend the locktime. This would be a huge improvement to UX, but would obviously require a lot of changes in **core** and UI.

martonp: Yeah that's a good idea.

buck54321:
I'm planning on resurrecting this effort soon.

martonp:
> I'm planning on resurrecting this effort soon.

Sounds good, I'll rebase and wrap it up. It would be good if you could do a basic review of `ETHBondV0.sol` first though.

PR #2219, opened 2023-03-14
client/asset/dcr: spv block header confs is -1 for orphan

chappjc:
This fixes `(*spvWallet).GetBlockHeader` so that it correctly considers the main chain status of the block when computing confirmations and retrieving `NextHash`.

The new function matches the behavior of dcrd and dcrwallet's getblockheader handlers.

PR #2225, opened 2023-03-15
client/asset/btc: fix SPV reorg handling

norwnd:
This PR targeting BTC SPV wallet being able to spot reorgs, as per https://github.com/decred/dcrdex/pull/2219#pullrequestreview-1340616893 (functionally mirroring https://github.com/decred/dcrdex/pull/2219).

While it is possible to just fix the issue and move on I think some [code clean up is warranted here](https://github.com/decred/dcrdex/pull/2227).

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -4349,7 +4349,7 @@ func (btc *intermediaryWallet) reportNewTip(ctx context.Context, newTip *block) 
 	// Redemption search would typically resume from prevTipHeight + 1 unless the
 	// previous tip was re-orged out of the mainchain, in which case redemption
 	// search will resume from the mainchain ancestor of the previous tip.
-	prevTipHeader, err := btc.tipRedeemer.getBlockHeader(&prevTip.hash)
+	prevTipHeader, isMainchain, err := btc.tipRedeemer.getBlockHeader(&prevTip.hash)
 	switch {
 	case err != nil:
 		// Redemption search cannot continue reliably without knowing if there
@@ -4358,7 +4358,7 @@ func (btc *intermediaryWallet) reportNewTip(ctx context.Context, newTip *block) 
 			prevTip.hash, err)
 		return
 
-	case prevTipHeader.Confirmations < 0:
+	case !isMainchain:
 		// The previous tip is no longer part of the mainchain. Crawl blocks
 		// backwards until finding a mainchain block. Start with the block
 		// that is the immediate ancestor to the previous tip.
@@ -4368,12 +4368,12 @@ func (btc *intermediaryWallet) reportNewTip(ctx context.Context, newTip *block) 
 			return
 		}
 		for {
-			aBlock, err := btc.tipRedeemer.getBlockHeader(ancestorBlockHash)
+			aBlock, isMainchain, err := btc.tipRedeemer.getBlockHeader(ancestorBlockHash)
 			if err != nil {
 				notifyFatalFindRedemptionError("getBlockHeader error for block %s: %w", ancestorBlockHash, err)
 				return
 			}
-			if aBlock.Confirmations > -1 {
+			if isMainchain {
 				// Found the mainchain ancestor of previous tip.
 				startPoint = &block{height: aBlock.Height, hash: *ancestorBlockHash}
 				btc.log.Debugf("reorg detected from height %d to %d", aBlock.Height, newTip.height)
--- a/client/asset/btc/spv_wrapper.go
+++ b/client/asset/btc/spv_wrapper.go
@@ -1726,7 +1726,7 @@ func (w *spvWallet) getWalletTransaction(txHash *chainhash.Hash) (*GetTransactio
 		TimeReceived: uint64(details.Received.Unix()),
 	}
 
-	if details.Block.Height > 0 {
+	if details.Block.Height >= 0 {
 		ret.BlockHash = details.Block.Hash.String()
 		ret.BlockTime = uint64(details.Block.Time.Unix())
 		// ret.BlockHeight = uint64(details.Block.Height)
--- a/client/asset/btc/spv_wrapper.go
+++ b/client/asset/btc/spv_wrapper.go
@@ -1055,8 +1055,10 @@ func (w *spvWallet) getBlockHeader(blockHash *chainhash.Hash) (header *blockHead
 	mainchain = w.blockIsMainchain(blockHash, blockHeight)
 	if mainchain {
 		confirmations = int64(confirms(blockHeight, tip.Height))
-	}
-	if tip.Height < blockHeight { // if tip is less, may be rolling back, so just mock dcrd/dcrwallet
+	} else if tip.Height < blockHeight {
+		// Gotta treat this block as mainchain, until mainchain tip catches up.
+		mainchain = true
+		// If tip is less, may be rolling back, so just mock dcrd/dcrwallet.
 		confirmations = 0
 	}
 
--- a/client/asset/btc/spv_wrapper.go
+++ b/client/asset/btc/spv_wrapper.go
@@ -1034,7 +1034,7 @@ func (w *spvWallet) walletUnlock(pw []byte) error {
 
 // getBlockHeader gets the *blockHeader for the specified block hash. It also
 // returns a bool value to indicate whether this block is a part of main chain.
-// For orphaned blocks header.Confirmations is negative (typically -1).
+// For orphaned blocks header.Confirmations is negative.
 func (w *spvWallet) getBlockHeader(blockHash *chainhash.Hash) (header *blockHeader, mainchain bool, err error) {
 	hdr, err := w.cl.GetBlockHeader(blockHash)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


client/asset/btc/btc.go
```diff
@@ -3719,7 +3719,7 @@ func (btc *intermediaryWallet) FindRedemption(ctx context.Context, coinID, _ dex
 		return nil, nil, fmt.Errorf("queueFindRedemptionRequest error for redemption %s: %w", outPt, err)
 	}
 
-	go btc.tryRedemptionRequests(ctx, nil, []*findRedemptionReq{req})
+	btc.tryRedemptionRequests(ctx, nil, []*findRedemptionReq{req})
```
norwnd, line 3722: There seems to be no need to run this as a separate go-routine (and dcr wallet doesn't) because we are waiting on the result below anyway.

norwnd: One other thing of note here is that we set `startBlock` as `nil` here, which prompts us to find lowest height that we want to start with like this:
```
		iHeight = math.MaxInt32
		for _, req := range undiscovered {
			if req.blockHash != nil && req.blockHeight < iHeight {
				iHeight = req.blockHeight
				iHash = req.blockHash
			}
		}
```
and I'm not entirely sure that this way ^ we can't end up in a situation where `iHeight` will be == `0` which will force us to search through all blocks starting with 0th while we should have searched mempool only for this case instead, which is what DCR wallet does:
```
// Run initial search for redemption. If this contract is unmined,
// only scan mempool transactions as mempool contracts can only be
// spent by another mempool tx. If the contract is mined, scan all
// mined tx inputs starting from the block in which the contract is
// mined, up till the current best block. If the redemption is not
// found in that block range, proceed to check mempool.
if contractBlock == nil {
    dcr.findRedemptionsInMempool([]outPoint{contractOutpoint})
} else {
    dcr.tipMtx.RLock()
    bestBlock := dcr.currentTip
    dcr.tipMtx.RUnlock()
    dcr.findRedemptionsInBlockRange(contractBlock.height, bestBlock.height, []outPoint{contractOutpoint})
}
```
do we need to do something like this ^ for BTC also ?

norwnd:
Will also probably want to add more unit tests (similar to https://github.com/decred/dcrdex/pull/2219) here, but existing ones are at least passing.

chappjc:
Thanks.
- you may squash the fixup commit for blockheight
- the find redemption changes should have been a separate commit

norwnd:
> you may squash the fixup commit for blockheight

Yep, just squashed, originally decided to do a commit on top in case somebody was already reviewing the 1st commit.

chappjc:
Conditionally tagging this for 0.6, but again I would prefer the unrelated find redemption changes in a separate commit.  It needs discussion.

client/asset/btc/btc.go
norwnd: moved here: https://github.com/decred/dcrdex/pull/2227/files#r1138098479

norwnd:
> but again I would prefer the unrelated find redemption changes in a separate commit. It needs discussion.

Okay, turned out simple enough to do, moved redemptions search refactor [here](https://github.com/decred/dcrdex/pull/2227) with that last force-push.

client/asset/btc/spv_wrapper.go
```diff
@@ -1030,36 +1032,50 @@ func (w *spvWallet) walletUnlock(pw []byte) error {
...
-		Hash:          hdr.BlockHash().String(),
-		Confirmations: int64(confirms(blockHeight, tip.Height)),
-		Height:        int64(blockHeight),
-		Time:          hdr.Timestamp.Unix(),
-	}, nil
+		Hash:              hdr.BlockHash().String(),
+		Confirmations:     confirmations,
+		Height:            int64(blockHeight),
+		Time:              hdr.Timestamp.Unix(),
+		PreviousBlockHash: hdr.PrevBlock.String(),
```
norwnd, line 1068: Looks like without initialising `PreviousBlockHash` here we won't be able to to crawl back the chain to find common ancestor with mainchain (instead, we'd nil pointer dereference).

chappjc (approved):

client/asset/btc/spv_wrapper.go
```diff
@@ -1711,7 +1726,7 @@ func (w *spvWallet) getWalletTransaction(txHash *chainhash.Hash) (*GetTransactio
 		TimeReceived: uint64(details.Received.Unix()),
 	}
 
-	if details.Block.Height != -1 {
+	if details.Block.Height > 0 {
```
itswisdomagain, line 1729: Does not block height == 0 mean tx is in the genesis block and shouldn't we set the block details if so?

client/asset/btc/spv_wrapper.go
```diff
@@ -1030,36 +1032,50 @@ func (w *spvWallet) walletUnlock(pw []byte) error {
...
+	}
+
+	confirmations := int64(-1)
+	mainchain = w.blockIsMainchain(blockHash, blockHeight)
+	if mainchain {
+		confirmations = int64(confirms(blockHeight, tip.Height))
+	}
+	if tip.Height < blockHeight { // if tip is less, may be rolling back, so just mock dcrd/dcrwallet
+		confirmations = 0
 	}
```
itswisdomagain, line 1061: Even if it's not mainchain? If it's mainchain and `tip.Height < blockHeight`, `confirmations` would be set to 0 by the `confirms` fn.

client/asset/btc/spv_wrapper.go
norwnd: Yes, that's my understanding too, it shouldn't be an issue in practice because currently SPV wallets are dexc-managed (so there no way such a wallet could have BTC transactions in genesis block); that could change in the future though, so maybe better change this now too, **done**.

client/asset/btc/spv_wrapper.go
norwnd: I've re-checked with https://github.com/decred/dcrdex/pull/2219 once again,  there is slight discrepancy which should be addressed by https://github.com/decred/dcrdex/pull/2225/commits/3ded90408babb40df8969e0bc893c85e2add313c, thanks for pointing it out!

> If it's mainchain and tip.Height < blockHeight, confirmations would be set to 0 by the confirms fn.

I believe when we have `tip.Height < blockHeight` we can't also have `mainchain=true`, right ? So this shouldn't be an issue. But to be absolutely sure I've changed this condition a bit to only trigger on `mainchain=false`.

client/asset/btc/spv_wrapper.go
```diff
@@ -1030,36 +1032,52 @@ func (w *spvWallet) walletUnlock(pw []byte) error {
 	return w.Unlock(pw)
 }
 
-func (w *spvWallet) getBlockHeader(blockHash *chainhash.Hash) (*blockHeader, error) {
+// getBlockHeader gets the *blockHeader for the specified block hash. It also
+// returns a bool value to indicate whether this block is a part of main chain.
+// For orphaned blocks header.Confirmations is negative (typically -1).
```
chappjc, line 1037: The "typically" wording only applies to the RPC code.

client/asset/btc/spv_wrapper.go
chappjc: > I've changed this condition a bit to only trigger on `mainchain=false`.

I think @itswisdomagain's point was that this `tip.Height < blockHeight` check is not required at all because the `confirms` func handles this, unlike with dcr where we don't have that helper.  If `blockIsMainchain` says `false`, I don't believe we would overrule that.

In other words, with no knowledge about `confirms` functionality in this `<` situation, the code would be:

```go
	if mainchain {
		if tip.Height < blockHeight { // if tip is less, may be rolling back, so just mock dcrd/dcrwallet
			confirmations = 0
		} else {	
			confirmations = int64(confirms(blockHeight, tip.Height))
		}
	}

	return &blockHeader{
```

But `confirms` handles that already, so it's just

```go
	if mainchain {
		confirmations = int64(confirms(blockHeight, tip.Height))
	}

	return &blockHeader{
```

client/asset/btc/spv_wrapper.go
```diff
@@ -1745,12 +1762,12 @@ func (w *spvWallet) getWalletTransaction(txHash *chainhash.Hash) (*GetTransactio
...
-func confirms(txHeight, curHeight int32) int32 {
+func confirms(txHeight, curHeight int32) uint32 {
 	switch {
-	case txHeight == -1, txHeight > curHeight:
-		return 0
+	case curHeight >= txHeight:
+		return uint32(curHeight - txHeight + 1) // positive numbers here, no overflow
 	default:
-		return curHeight - txHeight + 1
+		return 0 // undefined for all other cases
```
chappjc, line 1770: One thing we're missing with these changes to `confirms` is special handling when `txHeight` is -1.  When this helper is used with `txHeight, _ := w.cl.GetBlockHeight(blockHash)` as in our `getBlockHeader` method, I don't think we'd ever get -1, **however** this helper with `wtxmgr.TxDetails` types like `txDetails.Block.Height` where -1 is expected in caess and we don't want to change any behavior there e.g. `getTxOut`.

Note that this `confirms` helper is lifted directly from dcrwallet:  https://github.com/decred/dcrwallet/blob/0aeb02cde83ebb552b802b4eebcc151fe371325c/wallet/wallet.go#L4446-L4456

Let's revert the changes to this method please.

chappjc (changes requested):

buck54321 (approved):

Issue #2076, opened 2023-01-25
dcr/spv: auto-discover redemption when blockchain forks (without restart)


norwnd:
Stumbled upon interesting case, using `testnets`. I was testing refunds, initiated DCR-ETH trade with myself, waited until Maker/Taker swaps have been broadcasted, and shut down `dexc`, waited ~30 mins and started `dexc`. Maker published redeem txn, it soon got mined, but Maker couldn't notify server about his redeem (because server doesn't keep track of revoked matches ?)

<img width="200" alt="image" src="https://user-images.githubusercontent.com/112318969/214648006-65527ded-d860-4636-9bd6-5fb6270c2da7.png"><img width="200" alt="image" src="https://user-images.githubusercontent.com/112318969/214649003-5448b892-9763-4cf7-8cc6-03cbee3eb8d9.png">

but 2 hours later Taker still can't find Maker redeem:

<img width="200" alt="image" src="https://user-images.githubusercontent.com/112318969/214648954-bf390b79-42c7-4ea5-8b08-a1370fc25bf9.png">

[Redemption txn](https://testnet.dcrdata.org/tx/57c6505bc9e3342a1b8ea199b978e52b2e06bde89dab0d8d0f51de51f134375c/out/0) got mined in 1066748 (0000000022d52e3e33287ebce49fa96a7d5e3adea5dce02c50b144424ac32a94) block (on main-chain).

digging through the logs, here is what I think happened:
```
2023-01-25 15:54:27.095 [DBG] CORE[dcr]: finding redemptions for 1 contracts in blocks 1066738 - 1066748
2023-01-25 15:54:27.096 [DBG] CORE[dcr]: 0 redemptions found, 0 canceled out of 1 contracts in blocks 1066738 to 1066748
```
here ^ my DCR SPV wallet was looking at 1066748 on side-chain (see more on this below; probably on side-chain txn hasn't been mined in 1066748 yet, while it has been on main-chain), 
then search continues, **never re-checking 1066748 block** (which **contains** redemption txn on main-chain)
```
2023-01-25 16:00:59.881 [DBG] CORE[dcr]: finding redemptions for 1 contracts in blocks 1066749 - 1066749
2023-01-25 16:00:59.882 [DBG] CORE[dcr]: 0 redemptions found, 0 canceled out of 1 contracts in blocks 1066749 to 1066749
```
and here is another relevant bit from my logs (looks like it was searching side-chain block 1066748, not main-chain one):
```
[...]

PR #2226, opened 2023-03-15
client/asset/dcr: dedup find mainchain ancestor block

norwnd:
Found duplicate block of code when researching https://github.com/decred/dcrdex/issues/2076.

### DIFF
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -4689,12 +4689,15 @@ func (dcr *ExchangeWallet) handleTipChange(ctx context.Context, newTipHash *chai
 		return
 	}
 
+	// Lock to avoid concurrent handleTipChange execution for simplicity.
 	dcr.tipMtx.Lock()
 	defer dcr.tipMtx.Unlock()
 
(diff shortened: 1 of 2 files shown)


JoeGruffins (approved):
Looks correct. Will test a bit.

client/asset/dcr/dcr.go
```diff
@@ -4710,12 +4713,18 @@ func (dcr *ExchangeWallet) handleTipChange(ctx context.Context, newTipHash *chai
 		return
 	}
 
+	startHeight := prevTip.height + 1
```
buck54321, line 4716: Why is this moved up to here?

norwnd: Just to keep `isMainchain` check as close as possible to `blockHeader` (where we basically got it from) - so it looks as single logical block of code. While for `startHeight` it doesn't really matter where it will be placed, all else equal typically prefer not to break those logical blocks.

image

buck54321: Seems pretty subjective. Try not to fix things that aren't broken. You can leave it now though.

Issue #2224, opened 2023-03-14
Documentation for ETH Wallet


martonp:
We need some sort of documentation or instructions on how to use the ETH wallet. I don't think people will know how to setup a provider.

chappjc:
I agree.  I had some recommendations and advice I wanted to put somewhere too.  I kinda wish we could just edit the wiki without using the repo

PR #2233, opened 2023-03-20
wiki: add Ethereum page

chappjc:
Resolves https://github.com/decred/dcrdex/issues/2224

Click "View file" from the menu for docs/wiki/Ethereum.md to render Markdown.


docs/wiki/Ethereum.md
```diff
@@ -0,0 +1,102 @@
...
+Casual ETH users may be familiar with Infura, as an example, since this is the
+default provider for the common MetaMask wallet. When creating your Ethereum
+wallet in the DEX application, you will need to specify the full path to an
+Ethereum RPC provider, i.e. an RPC endpoint.
+
+## Wallet Setup
+
+From the "Wallets" page of the DEX application, you select Ethereum from the
+left menu, which will show you your the ETH balance if you already have a
+wallet, or you will get a "Create a Ethereum Wallet" button:
```
ukane-philemon, line 25: > or you will get 

nit: or you will see

ukane-philemon (approved):
Thanks, the doc is well written but others could have something to say.

docs/wiki/Ethereum.md
```diff
@@ -0,0 +1,102 @@
...
+wallets for BTC and DCR, but requires some manual configuration.
+
+Like the other native wallets, the private keys (and accounts) are derived from
+the DEX "application seed", so that reinitializing the DEX app from seed will
+restore the same Ethereum wallet and any ERC20 wallets. The keys are maintained
+exclusively by the application process, and all transaction signing is handled
+by the application.
+
+Unlike the other native wallets, the Ethereum wallet uses "RPC providers"
+specified by the user. The is the wallet's gateway to the Ethereum network.
```
redtobster, line 15: >The is

I believe it should be "This is" ?

docs/wiki/Ethereum.md
```diff
@@ -0,0 +1,102 @@
...
+  allow the DEX app to use it. However, typically even a full node will expose a
+  WebSocket interface.
+
+## RPC Provider Suggestions
+
+Consider the following tips when configuring your Ethereum wallet:
+
+- Use **multiple providers**, even if one of them is your own full node. This
+  provides redundancy. Often a provider will fall behind or even stall, so
+  multiple providers adds robustness to your wallet.
```
JoeGruffins, line 54: multiple providers adds -> multiple providers add

docs/wiki/Ethereum.md
```diff
@@ -0,0 +1,102 @@
...
+  particularly important if you use a VPN or other proxy where you share an
+  outbound IP address with many other people.
+
+  Further, taking the time to get an endpoint with your personal API key often
+  gives you access to a "dashboard" with the provider so you can monitor your
+  API requests.
+
+- Use **WebSockets** (wss:// URL scheme) when using a third-party RPC provider.
+  It is much more efficient than HTTP, the wallet will become aware of new
+  blocks sooner, and it will consume less of your quota with the provider.
```
JoeGruffins, line 72: I believe the client will attempt to upgrade even if the user specifies http. Could be noted.

JoeGruffins (approved):

martonp (approved):

docs/wiki/Ethereum.md
```diff
@@ -0,0 +1,104 @@
...
+  allow the DEX app to use it. However, typically even a full node will expose a
+  WebSocket interface.
+
+## RPC Provider Suggestions
+
+Consider the following tips when configuring your Ethereum wallet:
+
+- Use **multiple providers**, even if one of them is your own full node. This
+  provides redundancy. Often a provider will fall behind or even stall, so
+  use of multiple providers adds robustness to your wallet.
```
norwnd, line 54: I think most users would ignore this recommendation go for single provider (and btw will also be annoyed by that need to configure it on their own unlike with Metamask) without fully realising the risks they are taking. Users might also ignore the recommendation about **private API key** because "it is just a recommendation, and requires additional effort".

Since ignoring these recommendations (providers go down, req limit could be reached) could result in failure to redeem (if all providers fail while user is trying to redeem), how about:
- (bad UX) require at least 2 providers to be configured by user instead of just 1
- (much better UX) allow just 1 provider, but have 2 or more hardcoded fallback providers that will not affect current flow in any way except that these will be used as a measure of last resort when none of user-configured providers are available

norwnd:
Is there a way to embed the link to this wiki page somewhere in UI (somewhere visible, like on create and reconfigure ETH/Token wallet forms) so that user is prompted to read this tutorial ? Expecting users to find it on their own is probably a set up for countless help-requests in decred support channels.

ukane-philemon:
We could introduce a help page that contains links to guides or content for each of the different assets supported in the UI but that would require careful planning and positioning so it fits the client design and is easily accessible.

norwnd:
> A separate issue can be opened if need be to discuss how we can introduce a help page in the client or somewhere easily accessible by non-techy users.

True, just seems like this doc is a prime example of this occasional need for user-guidance. Moving discussion to https://github.com/decred/dcrdex/issues/2245.

docs/wiki/Ethereum.md
norwnd: > But what if I do not what those providers used? What if I'm trying to use one for whatever reason (privacy, security, etc)?

Could potentially have additional opt-in flag "never use default providers" (with the description about this being more privacy-oriented). I think opt-in features are better suited to those people who like to configure their env exactly the way they want, not for average user who might not even want to understand why it is there (assuming DEX at least remotely tries to target such users).

Here is a ticket to continue this discussion:https://github.com/decred/dcrdex/issues/2246.

PR #2234, opened 2023-03-21
client/eth: Add send tx fee estimate buffer

### LINKED ISSUE #2186: eth: sending full ETH balance isn't resilient to small gas price changes
Noticed that trying to send out all ETH funds often errors like this for me on **testnet**, latest **master** (either send attempt itself is rejected, or a warning about fee estimation pops up):

[sending-eth-issues.webm](https://user-images.githubusercontent.com/112318969/221803972-a2253bd9-5f24-4efe-b212-90317c40f871.webm)

a couple of related points/questions on this to discuss:
- I observed this issue today quite consistently, that's probably not a behavior users are expected to hit often or maybe ever; as you can see on the vid ^ the difference in gas price (when estimates taken 1st to decide how much to send, and when 2nd verification check happens) isn't huge (it isn't small there either, but I saw sub 3% difference to prevent send attempt), should it handle small differences without error (what the threshold/wiggle-room should be, something on the order of 10-20% maybe) ?
- I expect this issue might happen for ERC20s too when user has barely enough ETH to cover token transfer fees, such that slight gas price increase could render his balance lacking when 2nd check performed
- on the vid above I was sending ETH (which has a fixed 21 000 gas limit, and it is spent 100% unlike with ERC20s where it **gets much worse**) and it seems the dust left after successful attempt in ETH account wasn't an insignificant amount; the main reason seems to be `maxFee` set to 2x of `baseFee` - which effectively (very often)  means we are leaving roughly same amount of ETH as we spend on transaction in the wallet as dust for the use-case when user just wants his funds out (maybe never touching wallet again); any reason to use `maxFee` that high here instead of letting user to rebroadcast instead (if necessary) for sending out of his wallet (when trades are processing setting `maxFee` high is probably a warranted safety margin, no questions there) ?

martonp:
This adds a 20% buffer to the send tx fee estimate in order to not return an error from the send function when the base fee rises between the time the estimate is done and the ETH is sent.

Closes #2186

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -2768,7 +2768,7 @@ func isValidSend(addr string, value uint64, subtract bool) error {
 // canSend ensures that the wallet has enough to cover send value and returns
 // the fee rate and max fee required for the send tx. If isPreEstimate is false,
 // wallet balance must be enough to cover total spend.
-func (w *ETHWallet) canSend(value uint64, isPreEstimate bool) (uint64, *big.Int, error) {
+func (w *ETHWallet) canSend(value uint64, verifyBalance, isPreEstimate bool) (uint64, *big.Int, error) {
 	maxFeeRate, err := w.recommendedMaxFeeRate(w.ctx)
 	if err != nil {
 		return 0, nil, fmt.Errorf("error getting max fee rate: %w", err)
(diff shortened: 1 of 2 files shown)


chappjc (approved):
utACK.  Code reads well.

JoeGruffins (approved):

Issue #2192, opened 2023-03-01
wallet birthday at creation assumes GMT (just UI or backend too?)


chappjc:
It's 6:02pm CST on Feb 28 at the time of this, which is just past midnight GMT:

![image](https://user-images.githubusercontent.com/9373513/222010972-45b91476-b9c5-4730-905c-f6665a66ad64.png)

Seems like the future.  In fact, I'm not sure it's even handled correctly on the backend.

ukane-philemon:
> Seems like the future. In fact, I'm not sure it's even handled correctly on the backend.

The backend uses the default location for time values including wallet birthdays (i.e UTC), are you suggesting we change that to use the local timezone?

chappjc:
Looks to just be UI.  Docs for `toISOString` state "The timezone is always zero UTC offset, as denoted by the suffix Z.".  Fixing so we get the locale date.

chappjc:
This is way stranger than I thought.  Assigning to an input's value (or valueAsNumber or valueAsDate) and trying to get it back is even not so straightforward.
https://stackoverflow.com/questions/53032953/what-is-the-correct-way-to-set-and-get-htmlinputelement-valueasdate-using-local
https://stackoverflow.com/questions/16825363/how-do-you-parse-a-date-from-an-html5-date-input?answertab=scoredesc#tab-top
I might move on from this.

PR #2236, opened 2023-03-21
app: fix wallet birthday is UTC date

chappjc:
Resolves https://github.com/decred/dcrdex/issues/2192

When using a `<input type='date'>` element, it is tricky to get the time zone correct when using the `value`, `valueAsNumber`, and `valueAsDate` fields.  This is because the browser will always display the date as UTC, regardless of how you have set `valueAsDate`.  Also, when getting the date out with `new Date(input.value)`, where the value is `yyyy-mm-dd`, the `Date` parses as in UTC (12am UTC instead of local).

When the user is interacting with the input element, they are thinking in local time.  However, unless we offset the time that we set, it can show an incorrect date.  This change sets the `value` date string manually to the local date. We could also offset the `Date` that we assign to `valueAsDate` so that it will show a local date, or similarly with `valueAsNumber`, but this is harder to follow and we do not benefit from finer precision than simply 12am of the current day.

When reading the selected date back into javascript, we also have to interpret it as a local date string.  We do this by forcing the `Date` constructor to interpret the date as local by appending `"T00:00"` to the date string from `input.value`.  We also must use this trick when reading the `min` and `max` date strings.  We could instead use `valueAsDate` and manually shift in the opposite direction as when we set it, but this seems better.

See https://dev.to/mendyberger/input-and-js-dates-2lhc for a better explanation

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -550,7 +550,7 @@ type WalletConfig struct {
 }
 
 // AdjustedBirthday converts WalletConfig.Birthday to a time.Time, and adjusts
-// it so that defaultWalletBirthday <= WalletConfig.Bithday <= now.
+// it so that defaultWalletBirthday <= WalletConfig.Birthday <= now.
 func (cfg *WalletConfig) AdjustedBirthday() time.Time {
 	bday := time.Unix(int64(cfg.Birthday), 0)
 	now := time.Now()
--- a/client/webserver/site/src/html/bodybuilder.tmpl
+++ b/client/webserver/site/src/html/bodybuilder.tmpl
@@ -104,7 +104,7 @@
 {{end}}
 
 {{define "bottom"}}
-<script src="/js/entry.js?v=YYNQOy1"></script>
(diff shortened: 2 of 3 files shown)


chappjc:
Seems to be working as expected, but I'm going to try again in a couple hours when it's tomorrow in GMT.

chappjc:
No longer getting a future date when it's tomorrow in GMT but not locally.  Wallet backend gets the right date as well, logged as `2023-03-21 00:00:00 -0500 CDT 1679374800` (beginning of current day).

JoeGruffins (approved):
If you change the btc or dcr wallet from rpc to spv the date is not current. It seems like something random. Is this date intended?

Just created rpc wallet on simnet changing to spv:

image

chappjc:
> If you change the btc or dcr wallet from rpc to spv the date is not current. It seems like something random. Is this date intended?

Yeah, sorta intended.  @martonp would probably be most familiar with the logic, but that may 27, 2022 day is the default  birthday under certain conditions.  Perhaps the workflow you described of switching wallet types evades the logic that detects the "is seeded" check that should use a more recent time.

PR #2248, opened 2023-03-23
server/asset: trade address validation consistency

chappjc:
The trade address provided with an order is checked for validity, but it also must be checked that it is the proper type for inclusion in a swap contract.  This renames `server.asset.CheckAddress` to `CheckSwapAddress`, and updates it with the same type checks used client-side in `dex/networks/{dcr,btc}.MakeContract` as well as in `ExtractSwapDetails`.


chappjc:
image

image

JoeGruffins (approved):

PR #2250, opened 2023-03-23
client/asset: add methods for ZCash shielded pools

buck54321:
Add methods for working with shielded pools. Only the client/asset parts are implemented. **client/core** and UI will follow in separate efforts. As such, this API may change.

### DIFF
--- a/client/asset/interface.go
+++ b/client/asset/interface.go
@@ -74,36 +74,38 @@ func (wt WalletTrait) IsWithdrawer() bool {
 	return wt&WalletTraitWithdrawer != 0
 }
 
-// IsSweeper test if the WalletTrait has the WalletTraitSweeper bit set, which
+// IsSweeper tests if the WalletTrait has the WalletTraitSweeper bit set, which
 // indicates the presence of a Sweep method.
 func (wt WalletTrait) IsSweeper() bool {
 	return wt&WalletTraitSweeper != 0
 }
 
-// IsRestorer test if the WalletTrait has the WalletTraitRestorer bit set, which
+// IsRestorer tests if the WalletTrait has the WalletTraitRestorer bit set, which
 // indicates the wallet implements the WalletRestorer interface.
 func (wt WalletTrait) IsRestorer() bool {
 	return wt&WalletTraitRestorer != 0
 }
 
-// IsTxFeeEstimator test if the WalletTrait has the WalletTraitTxFeeEstimator
+// IsTxFeeEstimator tests if the WalletTrait has the WalletTraitTxFeeEstimator
 // bit set, which indicates the wallet implements the TxFeeEstimator interface.
 func (wt WalletTrait) IsTxFeeEstimator() bool {
 	return wt&WalletTraitTxFeeEstimator != 0
 }
 
-// IsPeerManager test if the WalletTrait has the WalletTraitPeerManager bit
+// IsPeerManager tests if the WalletTrait has the WalletTraitPeerManager bit
 // set, which indicates the wallet implements the PeerManager interface.
 func (wt WalletTrait) IsPeerManager() bool {
 	return wt&WalletTraitPeerManager != 0
 }
 
-// IsAuthenticator test if WalletTrait has WalletTraitAuthenticator bit set,
+// IsAuthenticator tests if WalletTrait has WalletTraitAuthenticator bit set,
 // which indicates authentication is required by wallet.
 func (wt WalletTrait) IsAuthenticator() bool {
 	return wt&WalletTraitAuthenticator != 0
 }
 
+// IsShielded tests if the WalletTrait has the WalletTraitShielded bit
+// set, which indicates the wallet implements the ShieldedWallet interface.
 func (wt WalletTrait) IsShielded() bool {
 	return wt&WalletTraitShielded != 0
 }
@@ -855,7 +857,7 @@ type ShieldedWallet interface {
 	// account.
 	ShieldedStatus() (*ShieldedStatus, error)
 	// NewShieldedAddress creates a new shielded address. A shielded address can
-	// be be reused without sacrifice of privacy on-chain, but that doesn't stop
+	// be reused without sacrifice of privacy on-chain, but that doesn't stop
 	// meat-space coordination to reduce privacy.
 	NewShieldedAddress() (string, error)
 	// ShieldFunds moves funds from the transparent account to the shielded
--- a/client/asset/zec/shielded_rpc.go
+++ b/client/asset/zec/shielded_rpc.go
@@ -3,11 +3,20 @@
 
 package zec
 
+import (
+	"decred.org/dcrdex/dex"
+	"github.com/btcsuite/btcd/btcutil"
+)
+
 const (
 	methodZGetAddressForAccount = "z_getaddressforaccount"
 	methodZListUnifiedReceivers = "z_listunifiedreceivers"
 	methodZListAccounts         = "z_listaccounts"
 	methodZGetNewAccount        = "z_getnewaccount"
+	methodZGetBalanceForAccount = "z_getbalanceforaccount"
+	methodZSendMany             = "z_sendmany"
+	methodZValidateAddress      = "z_validateaddress"
+	methodZGetOperationResult   = "z_getoperationresult"
 )
 
 type zListAccountsResult struct {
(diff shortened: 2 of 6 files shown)


client/asset/zec/zec.go
```diff
@@ -185,6 +193,12 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, net dex.Network) (ass
 		SingularWallet:           true,
 		UnlockSpends:             true,
 		FeeEstimator:             estimateFee,
+		ConnectFunc: func() error {
+			return connect(w)
+		},
+		AddrFunc: func() (btcutil.Address, error) {
+			return transparentAddress(w, addrParams, btcParams)
```
chappjc, line 200: I will look closer at this, but I think I love this capture of the rpcCaller vs the awkwardness we ensure with `estimateFee(ctx context.Context, node btc.RawRequester, ...)`.
Seems like in an other commit we could follow this pattern and eliminate a bunch of cruft like `btc.RawRequester` itself.

client/asset/interface.go
```diff
@@ -814,6 +822,37 @@ type BotWallet interface {
...
+	NewShieldedAddress() (string, error)
+	// ShieldFunds moves funds from the transparent account to the specified
+	// shielded address.
+	ShieldFunds(ctx context.Context, shieldedAddr string, amt uint64) (dex.Bytes, error)
+	// UnshieldFunds moves funds from the specified shielded address to the
+	// transparent account.
+	UnshieldFunds(ctx context.Context, shieldedAddr string, amt uint64) (dex.Bytes, error)
+	// SendShielded sends funds from the specified shielded account to the
+	// provided shielded or transparent address.
+	SendShielded(ctx context.Context, shieldedAddr, toAddr string, amt uint64) (dex.Bytes, error)
```
chappjc, line 853: Since the `Core` changes come next, could you describe the intended flow of funds from deposits and trades into and out of shielded addresses, to effectively achieve mixing presumably?

buck54321: > Since the Core changes come next, could you describe the intended flow of funds from deposits and trades into and out of shielded addresses, to effectively achieve mixing presumably?

Great question! In the lingo I'm trying to normalize, the model goes as such. Every wallet has 2 parts. There's the transparent part, which is the only part that we've known so far. And there's the shielded part = the new secret part. All swap activity will take place on the transparent part. That's why the server will never be involved with this effort. But we can also transfer value into and out of the shielded part (on-chain). Transferring value from the transparent part to the shielded part does not reveal the shielded address. There is no way to view the balance of a shielded address by viewing blockchain data only. Tranferring value from, say, the Orchard shielded pool to another address in the Orchard shielded pool does not reveal the sender, receiver, or amount. This is the highest-security type of value transfer. Any other type of transfer, e.g. Orchard-to-Sapling, Orchard-to-transparent, transparent-to-Sapling, etc. will always reveal the transfer amount, which itself could potentially be used to reduce your anonymity. Any value transfer from or to a transparent address further reduces the anonymity, by revealing a pubkey hash (btc-like security).

Since the shielded address is never parseable from on-chain data, the shielded address can be reused (with some security implications for meat-world collaboration). For this reason, a shielded address is more akin to an "account". We don't need to generate a new address for every shielded transaction. We **can** generate new shielded addresses=accounts at will, but **we can't send value from more that one shielded address=account in a single transaction**. The address=account paradigm has serious implications for UI.


> And also a comment about why the consumer must be aware of this instead of the wallet handling everything itself, with things just being immature for a while, like with DCR.

With DCR in mixing mode, the wallet is assumed to be a properly configured remote dcrwallet instance accessed via RPC. It requires substantial external setup, but is expected to be self-sustaining after that. (When we implement Decred mixing natively to DEX, we will also need to expose more information to the consumer and to add UI elements.)

With ZCash, the situation is quite different, just because of the difference in the models. There is no "automatic" mode. Any use of privacy will require explicit interaction from the user.

martonp: When you say all of the swap activity takes place in the transparent part, do you mean that someone has to unshield manually, do a swap, and then reshield manually? I think creating the swap transaction could itself be an unsheild, and the redeem could be a reshield transaction.

buck54321: I'm not opposed to a shielded-by-default mode, but I don't think it's right for a first approach. It would require modifying virtually every step of the trade process and a deeper integration into the BTC package. It would not be a small undertaking. It's also not what I promised in the prop, so any time put into such an effort would potentially detract from our ability to deliver on other promises.

martonp: Wouldn't the only difference be to change the Swap to get inputs from the shielded pool and the redeem output to be to a shielded pool? Can definitely do it later though if there's time.

martonp: Ah I guess proving you have control of funds to the server would be a challenge.

client/asset/zec/shielded_test.go
```diff
@@ -0,0 +1,136 @@
+//go:build shielded
```
martonp, line 1: You should add this test to the `run_tests.sh` file.

client/asset/zec/shielded_test.go
```diff
@@ -0,0 +1,136 @@
...
+	return string(op), err
+}
+
+func getDeltaWallet(t *testing.T) (*zecWallet, func()) {
+	wi, err := NewWallet(&asset.WalletConfig{
+		Type: walletTypeRPC,
+		Settings: map[string]string{
+			"rpcuser":     "user",
+			"rpcpassword": "pass",
+			"rpcport":     "33770",
```
martonp, line 40: I successfully ran the tests with the `alpha` wallet, but when running with the other wallets I got 
```
rawrequest (getwalletinfo) error: -28: This wallet operation is disabled while reindexing.
```

Might just be an issue on my end.

client/asset/zec/zec.go
```diff
@@ -224,7 +238,229 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, net dex.Network) (ass
...
+	addrRes, err := zGetAddressForAccount(w, shieldedAcctNumber, []string{orchardAddressType})
+	if err != nil {
+		return "", err
+	}
+	return addrRes.Address, nil
+}
+
+// ShieldFunds moves funds from the transparent account to the specified
+// shielded address.
+func (w *zecWallet) ShieldFunds(ctx context.Context, shieldedAddr string, transparentVal uint64) (dex.Bytes, error) {
```
martonp, line 385: Since we're only doing one account, why should `core` have to deal with the various shielded addresses? It seems like this should just be managed by the wallet and there is no need for a `shieldedAddr` argument.

buck54321: One account can create multiple addresses, but shielded addresses are not single-use like trasnparent addresses, so a shielded address itself acts like an account. Users will be able to create new shielded addresses.

martonp: I assumed you would be able to use the funds from multiple addresses in the same account in the same transaction, since they are all controlled by the same private key. What's the point of having the same public key control the funds of multiple addresses if you can't send them in the same transaction?

client/asset/zec/shielded_test.go
buck54321: That error usually means that a node has not finished its initial sync. Non-alpha zec harness nodes take a while to get caught up after you run the harness. Is that possible?

client/asset/zec/zec.go
```diff
@@ -285,14 +298,171 @@ func connect(c rpcCaller) error {
...
+}
+
+// Balance adds a sum of shielded pool balances to the transparent balance info.
+func (w *zecWallet) Balance() (*asset.Balance, error) {
+	bal, err := w.ExchangeWalletFullNode.Balance()
+	if err != nil {
+		return nil, err
+	}
+
+	bals, err := w.ShieldedBalances()
```
martonp, line 350: Over here you could do a z_getAccountBalance to reduce the amount of rpc calls.

client/asset/zec/zec.go
martonp: I was talking to ChatGPT and it told me that all of the addresses in an account are derived from the same private key, so I assumed you could send funds from multiple addresses in the same tx. Gotta stop talking to ChatGPT lol.

client/asset/zec/shielded_test.go
martonp: Could be, but it was definitely running for a while before I ran the tests. I'll try again though.

client/asset/zec/zec.go
buck54321: So I was wrong. Balances and spending are associated with the account, not the particular address. When using `z_sendmany`, you can provide any address from the account as the `fromaddress` argument. Same thing for the deprecated `z_getbalance` method. Doesn't matter which address you specify, as long as it's from the right account. Had me all confused. Fixed now.

client/asset/zec/shielded_rpc.go
```diff
@@ -52,7 +67,153 @@ type unifiedReceivers struct {
...
+// favoredAddressAndPolicy returns the best address and the associated privacy
+// policy for a unified address.
+func (r *unifiedReceivers) favoredAddressAndPolicy() (string, privacyPolicy, error) {
+	switch {
+	case r.Orchard != "":
+		return r.Orchard, FullPrivacy, nil
+	case r.Sapling != "":
+		return r.Sapling, AllowRevealedRecipients, nil
+	case r.Transparent != "":
+		return r.Transparent, AllowRevealedAmounts, nil
```
chappjc, line 79: Also the context for this method is the *to* address for `z_sendmany`, correct?  
And only when sending *from* a shielded (orchard) address?
So while it feels nice as a method, it seems like a pure helper function for just `(*zecWallet).SendShielded` in zec.go i.e. `func favoredToAddressAndSendPolicy(r *unifiedReceivers)`.

But I think the sapling and transparent policies here are switched.

The docs for `z_sendmany` describe these as

```
- "FullPrivacy": Only allow fully-shielded transactions (involving a single shielded value pool).
- "AllowRevealedAmounts": Allow funds to cross between shielded value pools, revealing the amount
  that crosses pools.
- "AllowRevealedRecipients": Allow transparent recipients. This also implies revealing
  information described under "AllowRevealedAmounts".
- "AllowRevealedSenders": Allow transparent funds to be spent, revealing the sending
  addresses and amounts. This implies revealing information described under "AllowRevealedAmounts".
```

(`AllowRevealedRecipients` and  `AllowRevealedSenders` both imply `AllowRevealedAmounts`.)

If we are sending _to_ a transparent address, I would expect `AllowRevealedRecipients`.  And if we are sending to Sapling, I would expect just `AllowRevealedAmounts`.

client/asset/zec/shielded_rpc.go
```diff
@@ -52,7 +67,153 @@ type unifiedReceivers struct {
...
+				return nil, fmt.Errorf("z_sendmany operation failed: %s", res.Error.Message)
+			}
+			if res.Status == "success" {
+				txid, err := chainhash.NewHashFromStr(res.Result.TxID)
+				if err != nil {
+					return nil, fmt.Errorf("error decoding txid: %w", err)
+				}
+				return txid, nil
+			}
+			fmt.Println("--###########$$$$$$$$$$$$$$$$$$ operation status", res.Status)
```
chappjc, line 194: `Debugf` or remove?  Have you seen this print though?  If it never seems to happen, maybe `Warnf`.

client/asset/zec/zec.go
```diff
@@ -285,14 +294,207 @@ func connect(c rpcCaller) error {
...
+		return nil, asset.ErrInsufficientBalance
+	}
+
+	res, err := zValidateAddress(w, toAddr)
+	if err != nil {
+		return nil, fmt.Errorf("error validating address: %w", err)
+	}
+
+	var priv privacyPolicy
+	if res.AddressType == unifiedAddressType {
```
chappjc, line 477: `switch res.AddressType {` looks cleaner for this.

client/asset/zec/zec.go
```diff
@@ -285,14 +294,207 @@ func connect(c rpcCaller) error {
...
+			return nil, fmt.Errorf("error getting unified receivers: %w", err)
+		}
+		toAddr, priv, err = receivers.favoredAddressAndPolicy()
+		if err != nil {
+			return nil, fmt.Errorf("error parsing unified receiver: %w", err)
+		}
+	} else if res.AddressType == transparentAddressType {
+		priv = AllowRevealedRecipients
+	} else if res.AddressType == saplingAddressType {
+		priv = AllowRevealedAmounts
```
chappjc, line 490: These two policies look right, so I guess they are switched in `receivers.favoredAddressAndPolicy`

client/asset/interface.go
```diff
@@ -814,6 +822,34 @@ type BotWallet interface {
...
+	// NewShieldedAddress creates a new shielded address. A shielded address can
+	// be be reused without sacrifice of privacy on-chain, but that doesn't stop
+	// meat-space coordination to reduce privacy.
+	NewShieldedAddress() (string, error)
+	// ShieldFunds moves funds from the transparent account to the shielded
+	// account.
+	ShieldFunds(ctx context.Context, amt uint64) (dex.Bytes, error)
+	// UnshieldFunds moves funds from the shielded account to the transparent
+	// account.
+	UnshieldFunds(ctx context.Context, amt uint64) (dex.Bytes, error)
```
chappjc, line 847: Why use `dex.Bytes` for these returns?  We should prefer native types like `[]byte` unless there's a good reason, like if we have a struct field as a `dex.Bytes` to it marshalls to hex instead of base64, but I don't see it on these.

client/asset/zec/shielded_test.go
chappjc: On the beta node I saw `This wallet operation is disabled while reindexing` on my first manual RPC to the wallet (a `z_getbalanceforaccount`) even though it had been running for quite a long time.  I wonder if we need to give each node a "tick" from the harness shell script somehow?

client/asset/zec/zec.go
martonp: So ChatGPT was right.

chappjc: :robot:

client/asset/zec/zec.go
```diff
@@ -285,14 +294,207 @@ func connect(c rpcCaller) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return status, nil
+}
+
+// Balance adds a sum of shielded pool balances to the transparent balance info.
+func (w *zecWallet) Balance() (*asset.Balance, error) {
+	bal, err := w.ExchangeWalletFullNode.Balance()
```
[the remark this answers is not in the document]
buck54321: Wallets that were used before `getnewaddress` was deprecated will have addresses in a "legacy" account. `getbalance` returns balance of **all** transparent outputs, regardless of account, and including the "legacy" account. I wouldn't even know how to get just the legacy account balance to add to the `z_getbalanceforaccount` result for transparent account 0. Similarly, `listunspent` returns all transparent outputs, regardless of account. The only reason we need an account 0 is because `getnewaddress` is deprecated. I've taken care that addresses produced from the shielded account 1 don't include a transparent receiver, so barring external RPC sabotage, all t addrs will be from legacy or 0. 

Side note: It is unfortunate that you even need to include a shielded address in a unified address, but zcashd requires it, so our account 0 addresses **do** have a shielded receiver, though neither the unified address or the Orchard receiver are ever exposed to the caller.

> EDIT: Oh, I see your impl of zGetBalanceForAccount only gets Orchard balance, ignoring transparent value.

Generalizing that return output would be no problem. I just didn't have a reason yet.

client/asset/zec/shielded_rpc.go
```diff
@@ -52,7 +67,153 @@ type unifiedReceivers struct {
...
+		return "", "", fmt.Errorf("no known receiver types")
+	}
+}
+
 // z_listunifiedreceivers unified_address
 func zGetUnifiedReceivers(c rpcCaller, unifiedAddr string) (receivers *unifiedReceivers, err error) {
 	return receivers, c.CallRPC(methodZListUnifiedReceivers, []interface{}{unifiedAddr}, &receivers)
 }
+
+func zGetBalanceForAccount(c rpcCaller, acct uint32) (uint64, error) {
```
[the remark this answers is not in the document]
buck54321: No.

client/asset/zec/shielded_test.go
chappjc: No luck tweaking zec/harness.sh to fix this long delay.  Hopefully we find a solution, but we might have to live with it.

client/asset/zec/zec.go
```diff
@@ -286,14 +298,253 @@ func connect(c rpcCaller) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return status, nil
+}
+
+// Balance adds a sum of shielded pool balances to the transparent balance info.
+func (w *zecWallet) Balance() (*asset.Balance, error) {
+	bal, err := w.ExchangeWalletNoAuth.Balance()
```
chappjc, line 353: I would not object to a brief comment here touching on the "legacy" account concept you mention in https://github.com/decred/dcrdex/pull/2250#discussion_r1174427285, and the desire to aggregate transparent balance from any and all accounts, to match `listunspent`

client/asset/zec/zec.go
```diff
@@ -286,14 +298,253 @@ func connect(c rpcCaller) error {
...
+	return w.sendOne(ctx, anyTAddr, oneAddr, transparentVal, AllowRevealedSenders)
+}
+
+// UnshieldFunds moves funds from the shielded account to the transparent
+// account.
+func (w *zecWallet) UnshieldFunds(ctx context.Context, amt uint64) ([]byte, error) {
+	// lastAddr, err := w.lastShieldedAddress()
+	// if err != nil {
+	// 	return nil, err
+	// }
```
chappjc, line 414: rm?

chappjc (approved):
LGTM!  Just some nits and cleanup, plus a rebase/resolve with the other interface PR in.

client/asset/zec/zec.go
```diff
@@ -507,7 +514,7 @@ func (w *zecWallet) SendShielded(ctx context.Context, toAddr string, amt uint64)
 	if err != nil {
 		return nil, err
 	}
-	const fees = 1000
+	const fees = 1000 // TODO: Update after v4.5.0 which includes ZIP137
```
chappjc, line 517: I think you meant 5.5 :grin:

PR #2253, opened 2023-03-25
ui: prevent negative amt on send form

norwnd:
When transaction fee is comparable to amount left in user wallet we might get into negative territory calculating max send value.

[send-negative-amt-ui.webm](https://user-images.githubusercontent.com/112318969/227706401-7bdb39db-46ef-4b50-b8df-9366643f31be.webm)

### DIFF
--- a/client/webserver/site/src/js/wallets.ts
+++ b/client/webserver/site/src/js/wallets.ts
@@ -986,6 +986,9 @@ export default class WalletsPage extends BasePage {
         let canSend = wallet.balance.available
         if (!token) {
           canSend -= res.txfee
+          if (canSend < 0) {
+            canSend = 0
+          }
         }
         this.maxSend = canSend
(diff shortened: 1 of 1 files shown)


JoeGruffins (approved):

chappjc (approved):

client/webserver/site/src/js/wallets.ts
```diff
@@ -986,6 +986,9 @@ export default class WalletsPage extends BasePage {
         let canSend = wallet.balance.available
         if (!token) {
           canSend -= res.txfee
+          if (canSend < 0) {
+            canSend = 0
+          }
```
buck54321, line 991: One line please.

buck54321 (changes requested):

client/webserver/site/src/js/wallets.ts
norwnd: Changed.

Issue #2245, opened 2023-03-22
ui: eth/token wallet setup hints


norwnd:
Embed the link to wiki page from https://github.com/decred/dcrdex/pull/2233 somewhere in UI (somewhere visible, like on create and reconfigure ETH/Token wallet forms highlighted with red) so that user is prompted to read Ethereum tutorial. 

Expecting users to find it on their own is probably a set up for countless help-requests in decred support channels.

I think overall approach of hint-on-need-to-basis is more practical than having a separate wiki and expecting users to read it before using DEX.

buck54321:
Can add a field for a link to the `WalletDefinition` and display the link on the create/reconfig form (or hide link element if field is empty). Can I assign you, @norwnd?

norwnd:
Yep!

Issue #2246, opened 2023-03-22
client/asset/eth: fallback providers of last resort


norwnd:
Ignoring Ethereum recommendations listed in https://github.com/decred/dcrdex/blob/master/docs/wiki/Ethereum.md could result in failure to redeem (if all user-specified providers fail while user is trying to redeem),

one solution that doesn't degrade UX would be to allow user to configure just 1 provider (like we currently do), but have 2 or more hardcoded fallback providers (e.g. https://rpc.ankr.com/eth) that will not affect current flow in any way except that these will be used as a measure of last resort when none of user-configured providers are available.

For additional context/concerns see [this discussion](https://github.com/decred/dcrdex/pull/2233#discussion_r1144288254).

norwnd:
Would like to work on this, if this is a desirable feature.

norwnd:
> Lock times are long for a reason. User has plenty of time to fix the problem if their provider is really tanked for 8 hours.

Time isn't the main issue here (but even 8h might not be enough, can't user set up some trades and go afk for a while ? or have additional network issues to deal with on top), 

not providing a fallback here is basically equivalent to stating "**users who don't understand and remember the following**: 1) ETH provider can fall out of sync for whatever reason 2) that means your wallet falls out of sync 3) that in turn means trades might fail to complete (you likely loose funds) **shouldn't use DEX**"
- DEX doesn't target those users who don't understand this chain of events ?
- they might still use DEX and encounter this "issue" on their own, dumb as they are, and go in flames on media about DEX being unreliable; and quite rightfully so  because this could've been prevented quite easily
- even if I do understand this chain of events, I might not want to keep these low level details in my head when trading or doing something else

chappjc:
Well the least we can do is:
- have two provider text boxes already, so it's obvious that's suggested
- have a link to the wiki page where they can see a list of providers and learn what they are picking

On the philosophical points, yes, I think DEX users need to be more aware than the average user.  At least a little.  Sorry.  Further, MetaMask users:
- have blindly been unaware and as a consequence they have seen downtime when Infura experiences [downtime](https://decrypt.co/47846/ethereum-backbone-infura-suffers-major-damage), [repeatedly], [...](https://decrypt.co/98457/metamask-ethereum-apps-down-infura-outage)(https://coinmarketcap.com/alexandria/article/ethereums-infura-iating-outage-revives-decentralization-concerns).
[...]

PR #2261, opened 2023-03-27
ui/eth: wallet setup hints

norwnd:
Resolves https://github.com/decred/dcrdex/issues/2245, https://github.com/decred/dcrdex/issues/2246.

image

There are 2 providers now there ^ (with 2nd commit) actually, screenshot taken based on 1st commit from this PR).

### DIFF
--- a/client/asset/eth/eth.go
+++ b/client/asset/eth/eth.go
@@ -118,9 +118,10 @@ var (
 			Key:         providersKey,
 			DisplayName: "Provider",
 			Description: "Specify one or more providers. For infrastructure " +
-				"providers, use an https address. Only url-based authentication " +
+				"providers, prefer using wss address. Only url-based authentication " +
 				"is supported. For a local node, use the filepath to an IPC file.",
 			Repeatable: providerDelimiter,
+			RepeatN:    2,
 			Required:   true,
 		},
 	}
@@ -152,6 +153,7 @@ var (
 				ConfigOpts:  append(RPCOpts, WalletOpts...),
 				Seeded:      true,
 				NoAuth:      true,
+				GuideLink:   "https://github.com/decred/dcrdex/blob/master/docs/wiki/Ethereum.md",
 			},
 			// MaxSwapsInTx and MaxRedeemsInTx are set in (Wallet).Info, since
 			// the value cannot be known until we connect and get network info.
--- a/client/webserver/site/src/html/forms.tmpl
+++ b/client/webserver/site/src/html/forms.tmpl
@@ -43,6 +43,13 @@
 <div data-tmpl="errMsg" class="fs15 pt-3 text-center d-hide errcolor text-break"></div>
 {{end}}
 
+{{define "walletCfgGuideTemplate"}}
+<a data-tmpl="walletCfgGuide" class="d-flex justify-content-start align-items-center py-2 mt-2 pointer hoverbg" target="_blank">
+  <span class="ico-textfile fs18"></span>
+  <span class="pointer ps-1">[[[configuration guide]]]</span>
+</a>
+{{end}}
+
 {{define "newWalletForm"}}
 <div class="form-closer hoverbg"><span class="ico-cross"></span></div>
 <div class="wallet-tabs" data-tmpl="walletTypeTabs">
@@ -61,6 +68,7 @@
   <div data-tmpl="walletSettings">
     {{template "walletConfigTemplates"}}
   </div>
+  {{template "walletCfgGuideTemplate"}}
   <hr class="dashed my-2">
   <div class="d-flex justify-content-between align-items-stretch flex-wrap" data-tmpl="auth">
     <div class="col-11 p-0" data-tmpl="newWalletPassBox">
@@ -74,7 +82,7 @@
       <input type="password" class="form-control select" id="nwAppPass" data-tmpl="appPass" autocomplete="off">
     </div>
     <span class="col-3 hide-pw"></span> {{/* flex dummy */}}
-    <div class="d-grid align-items-end col-10 p-0 mt-4">
+    <div class="d-grid align-items-end col-10 p-0">
       <button data-tmpl="submitAdd" type="submit" class="fs15 bg2 selected">[[[Add]]]</button>
     </div>
   </div>
--- a/client/webserver/site/src/js/forms.ts
+++ b/client/webserver/site/src/js/forms.ts
@@ -67,7 +67,7 @@ export class NewWalletForm {
   current: CurrentAsset
   pwHiders: HTMLElement[]
   subform: WalletConfigForm
-  currentWalletType: string
+  walletCfgGuide: PageElement
   parentSyncer: null | ((w: WalletState) => void)
   createUpdater: null | ((note: WalletCreationNote) => void)
 
@@ -89,6 +89,8 @@ export class NewWalletForm {
 
     // WalletConfigForm will set the global app variable.
     this.subform = new WalletConfigForm(page.walletSettings, true)
+
+    this.walletCfgGuide = Doc.tmplElement(form, 'walletCfgGuide')
 
     bind(form, page.submitAdd, () => this.submit())
     bind(form, page.oneBttn, () => this.submit())
@@ -304,6 +306,7 @@ export class NewWalletForm {
     this.current.selectedDef = walletDef
     const appPwCached = State.passwordIsCached() || (this.pwCache && this.pwCache.pw)
     Doc.hide(page.auth, page.oneBttnBox, page.newWalletPassBox)
+    const guideLink = walletDef.guidelink
     const configOpts = walletDef.configopts || []
     // If a config represents a wallet's birthday, we update the default
     // selection to the current date if this installation of the client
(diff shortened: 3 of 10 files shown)


client/asset/eth/eth.go
```diff
@@ -121,6 +121,17 @@ var (
 			Repeatable: providerDelimiter,
 			Required:   true,
 		},
+		// Allow for single Provider setup, still nag the user to specify 2nd one for
+		// additional safety in case 1st provider fails during trade settlement.
+		{
+			Key:         providersKey,
+			DisplayName: "Provider",
+			Description: "Specify one or more providers. For infrastructure " +
+				"providers, use an https address. Only url-based authentication " +
```
chappjc, line 130: "use an https address"?  We're supposed to be recommending wss

client/webserver/site/src/html/forms.tmpl
```diff
@@ -1,4 +1,7 @@
 {{define "walletConfigTemplates"}}
+<div id="walletCfgGuide" class="py-2 d-hide">
+  Make sure you follow:  <a id="walletCfgGuideLink" target="_blank" class="py-2 hyperlink">configuration guide</a><br>
```
chappjc, line 3: Any text like this unfortunately needs to go in the locales.js like `[[[wallet_follow_the_guide]]]`

Regarding the text itself, why the colon?  Wouldn't it read better as "Make sure you follow the configuration guide."

chappjc (commented):
@buck54321 Do you think we need the guide link in the backend as part of the wallet definition?  I don't see a reason why the frontend shouldn't just have some map of guide links per asset, similar to how we do the explorer links.

client/webserver/site/src/html/forms.tmpl
buck54321: Better with no colon.

buck54321 (changes requested):
Please consider these changes.
https://github.com/buck54321/dcrdex/commit/a58ab2b544b767303d692e95435fc3f4ae47dcfd
image

buck54321:
> @buck54321 Do you think we need the guide link in the backend as part of the wallet definition? I don't see a reason why the frontend shouldn't just have some map of guide links per asset, similar to how we do the explorer links.

I had made this recommendation and I stand by it. Storing it in the front end code is hacky, and precludes non-UI consumers from accessing the information.

chappjc:
> Storing it in the front end code is hacky, and precludes non-UI consumers from accessing the information.

Alright, but I don't get why it's hacky, or why non-UI consumers would access a docs URL.  Filling up the backend code with this makes me cringe because it is in no way required to make or even configure the backend -- it's entirely docs i.e. UI.

norwnd:
Upon retesting I realised 2 providers by default doesn't quite work and needs further adjustments:

[add-2nd-provider-not-working.webm](https://user-images.githubusercontent.com/112318969/229025783-53871df6-d2b0-4fa1-8499-a0c1d1979e7e.webm)

possible way to resolve it: https://github.com/decred/dcrdex/pull/2261/commits/2f0fc860acf53e986b9c75a95b189026a40bd11c

client/asset/eth/eth.go
norwnd: Changed this text a bit.

norwnd:
> Please consider these changes.
https://github.com/buck54321/dcrdex/commit/a58ab2b544b767303d692e95435fc3f4ae47dcfd

It seems these need to be adjusted such that they also apply to "Reconfigure" form ? Basically similar how I have it now.

chappjc (approved):
Looks good to me.  I personally don't care for either `RepeatN` or `GuideLink` in the `ConfigOpts` defined on the backend, but I'm overruled.  It's certainly wrong to just duplicate the `providersKey` config opt entry, so `RepeatN` is right in that regard.  I think both the guide link and duplicating a form entry are entirely frontend concerns.  A headless bot wouldn't have any use for these config opt fields.

buck54321:
> I think both the guide link and duplicating a form entry are entirely frontend concerns. A headless bot wouldn't have any use for these config opt fields.

What about RPC users? They need to know wallet configuration instructions too.

client/webserver/site/src/js/forms.ts
```diff
@@ -557,12 +562,18 @@ export class WalletConfigForm {
   /*
    * update creates the dynamic form.
    */
-  update (configOpts: ConfigOption[] | null, activeOrders: boolean) {
+  update (guideLink: string, configOpts: ConfigOption[] | null, activeOrders: boolean) {
     this.assetHasActiveOrders = activeOrders
     this.configElements = []
     this.configOpts = configOpts || []
     Doc.empty(this.dynamicOpts, this.defaultSettings, this.loadedSettings)
```
buck54321, line 569: As I [suggested before](https://github.com/decred/dcrdex/pull/2261#pullrequestreview-1359740617), please move the `walletCfgGuide` out of the `walletSettings` and into the `mainForm`. This is not the purview of the dynamic form section.

chappjc:
> > I think both the guide link and duplicating a form entry are entirely frontend concerns. A headless bot wouldn't have any use for these config opt fields.
> 
> What about RPC users? They need to know wallet configuration instructions too.

These config options are not available via the RPC server.  The `newwallet` syntax is just to specify the settings:

```
settings (string): A JSON-encoded string->string mapping of additional
       configuration settings. These settings take precedence over any settings
       parsed from file. e.g. '{"account":"default"}' for Decred accounts
...
```

While that could be changed, and a walletdefinitions RPC added, I don't think it's worth it.

buck54321:
Huh. I thought you could access `SupportedAsset`s through RPC. Regardless, I don't view this information as being specific to our user interface implementation. It's asset-related information that could be useful to other users of `Core`. `Core` is not only for headless bots. More than anything though, I don't think this is worth arguing about any more. If it's important to you to move it to the front end, that's fine.

`RepeatN`, on the other hand, **is** extremely specific to our UI implementation, but I can see an argument for keeping it with the `ConfigOption` parameters too, so it's whatever. I don't want to keep dragging @norwnd into this philosophical debate. Let's just move forward.

chappjc:
> If it's important to you to move it to the front end, that's fine.

No, let's move on.

norwnd:
>  I don't want to keep dragging @norwnd into this philosophical debate. Let's just move forward.

Yeah, I don't think it makes much difference currently whether it's defined on backend or front end.

client/webserver/site/src/js/forms.ts
```diff
@@ -67,7 +67,7 @@ export class NewWalletForm {
   current: CurrentAsset
   pwHiders: HTMLElement[]
   subform: WalletConfigForm
-  currentWalletType: string
```
norwnd, line 70: Removed `currentWalletType` since unused.

client/webserver/site/src/js/forms.ts
norwnd: Moved now to a separate `template` that gets embedded into **both** `mainForm` and `reconfigForm` https://github.com/decred/dcrdex/pull/2261/commits/26c65ba6e867c76f3a7e0d1c77dee5cf11ac0aa8, see also [why I couldn't apply your change as is](https://github.com/decred/dcrdex/pull/2261#issuecomment-1491302624).

But some JS code is duplicated now (to serve 2 different forms), not sure if you'd want to dedup it by introducing additional entity that combines settings+guide into single section or something like that, so leaving like that for now.

buck54321: Sorry. I see what you're saying now. Less duplication to put it in the already templated section. The new arguments to `update` stuck out to me as a smell though. You can put it back if you want, but I'm fine either way.

buck54321 (changes requested):
This looks good, but I think I found some improvements. Lemme know what you think.
https://github.com/buck54321/dcrdex/compare/26c65ba...buck54321:dcrdex:repeat-addopt-internal
1. The `repeatableCounter` business was apparently only to avoid duplicate IDs for javascript `id => from` attribute matching. That's simplified.
2. Handling of `RepeatN` is internal to `addOpt`, reducing duplication
3. An extra blank field(s) is added to the reconfig form

client/webserver/site/src/js/forms.ts
norwnd: I think it's fine to leave it like it currently is for now.

norwnd:
> This looks good, but I think I found some improvements. Lemme know what you think. [buck54321/dcrdex@26c65ba...buck54321:dcrdex:repeat-addopt-internal](https://github.com/buck54321/dcrdex/compare/26c65ba...buck54321:dcrdex:repeat-addopt-internal)
> 
> 1. The `repeatableCounter` business was apparently only to avoid duplicate IDs for javascript `id => from` attribute matching. That's simplified.
> 2. Handling of `RepeatN` is internal to `addOpt`, reducing duplication
> 3. An extra blank field(s) is added to the reconfig form

Yep ty, looks cleaner and seems to be working, applied https://github.com/decred/dcrdex/pull/2261/commits/f0068475323aad5d40fcb8de9590322c714d8860

norwnd:
And now rebased.

chappjc:
Would like to ship this in the coming patch that adds zcash 5.4.2 support and dgb

buck54321 (approved):

chappjc:
Thanks.  `GuideLink` will be nice the Electrum wallet's too.  I'll get around to making a wiki page for that showing screens like [this](https://github.com/decred/dcrdex/wiki/images/electrum-rpc-config.png)

chappjc (approved):

PR #2266, opened 2023-03-29
client/asset/eth: broadcast to all providers

chappjc:
This adds `(*multiRPCClient).withAll`, which is used by `sendSignedTransaction` to broadcast a transaction to all providers, preferred first.

### DIFF
--- a/client/asset/eth/multirpc.go
+++ b/client/asset/eth/multirpc.go
@@ -102,6 +102,11 @@ type provider struct {
 	}
 }
 
+// String returns the provider host name.
+func (p *provider) String() string {
+	return p.host
+}
+
 func (p *provider) shutdown() {
 	p.stop()
 	p.ec.Close()
@@ -976,7 +981,7 @@ func (m *multiRPCClient) withOne(ctx context.Context, providers []*provider, f f
 		} else if err.Error() != superError.Error() {
 			superError = fmt.Errorf("%v: %w", superError, err)
 		}
-		for _, f := range acceptabilityFilters {
+		for _, f := range acceptabilityFilters { // use case for more than one? is it just variadic to allow 0?
 			discard, propagate, fail := f(err)
 			if discard {
 				return nil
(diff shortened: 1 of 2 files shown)


chappjc:
Regular send:

```
2023-04-06 00:32:43.773 [TRC] CORE[eth][RPC]: Sending signed tx via "IPC"
2023-04-06 00:32:43.773 [TRC] CORE[eth][RPC]: Sending signed tx via "127.0.0.1:38559"
2023-04-06 00:32:43.774 [TRC] CORE[eth][RPC]: non-fatal provider error: already known (*rpc.jsonError / <nil>)
2023-04-06 00:32:43.774 [INF] CORE: notify: |SUCCESS| (send) Send Successful - Sending 11.957361500 eth to 0x5173C33514731bBaF84a082be2354DfF9e01282a has completed successfully. Coin ID = 0x3c730a514d5fb0190465a5d6bec9dade185b85bd8c944f68dc69e5bc5162610d
2023-04-06 00:32:43.776 [TRC] WEB: "POST http://127.0.0.3:5758/api/send HTTP/1.1" from 127.0.0.1:37768 - 200 88B in 141.780817ms
```

Broadcast new contract:

```
2023-04-06 00:36:45.014 [TRC] CORE[eth][RPC]: Using cached header from "127.0.0.1:38559"
2023-04-06 00:36:45.015 [TRC] CORE[eth][RPC]: Sending signed tx via "127.0.0.1:38559"
2023-04-06 00:36:45.016 [TRC] CORE[eth][RPC]: Sending signed tx via "IPC"
2023-04-06 00:36:45.016 [TRC] CORE[eth][RPC]: non-fatal provider error: already known (*rpc.jsonError / <nil>)
2023-04-06 00:36:45.016 [INF] CORE: Broadcasted transaction with 1 swap contracts for order a9e499e6e871ea858d87f3170536d590bbb3f4c6d03cbc5d07955f2c7014cb2f. Assigned fee rate = 200. Receipts (eth): [{ tx hash: 0x1ede6be5afab509a09a2d94789441284034b2912c08b904ca7097532df5fb0b8, ...
```

Counterparty rebroadcasting before audit:

```
2023-04-06 00:36:45.021 [TRC] CORE[eth][RPC]: Sending signed tx via "IPC"
2023-04-06 00:36:45.021 [TRC] CORE[eth][RPC]: non-fatal provider error: already known (*rpc.jsonError / <nil>)
2023-04-06 00:36:45.021 [TRC] CORE[eth][RPC]: Sending signed tx via "127.0.0.1:38559"
2023-04-06 00:36:45.022 [TRC] CORE[eth][RPC]: non-fatal provider error: already known (*rpc.jsonError / <nil>)
2023-04-06 00:36:45.022 [INF] CORE: Audited contract (eth: 0x1ede6be5afab509a09a2d94789441284034b2912c08b904ca7097532df5fb0b8) ...
```

Redeem broadcast:

```
2023-04-06 00:37:19.827 [TRC] CORE[eth][RPC]: Sending signed tx via "127.0.0.1:38559"
2023-04-06 00:37:19.827 [TRC] CORE[eth][RPC]: Sending signed tx via "IPC"
2023-04-06 00:37:19.827 [TRC] CORE[eth][RPC]: non-fatal provider error: already known (*rpc.jsonError / <nil>)
2023-04-06 00:37:19.829 [INF] CORE: Broadcasted redeem transaction spending 1 contracts for order 878aafe3b6bc132a14f9b79e9ac46bc50b80561ccf062953415154e131c367e5, paying to 0x479e7d80c300e02a093be968b0066f2c06f9da11859876d5366447ab36135cc9 (eth)
```

chappjc:
While this PR overall should help ensure the transactions quickly propagate and are mined, while helping to keep the providers reporting the same next nonce, my only concern is that some providers might deviate from the RPC error "already known".

buck54321:
It is straight from geth, so I think we have a chance.
```
https://github.com/ethereum/go-ethereum/blob/b946b7a13b749c99979e312c83dce34cac8dd7b1/core/txpool/txpool.go#L61
```
Regardless, it'd be hard to get stuck because at least one provider would not error on our first successful broadcast.

buck54321 (approved):

chappjc:
> Regardless, it'd be hard to get stuck because at least one provider would not error on our first successful broadcast.

While each provider gets a try, one "non-discarded" error is enough to return a non-nil error to the caller.
So my concern is that this might make it more likely for say `Swap` to error via `(*BoundContract).transact` using `(*multiRPCClient).SendTransaction`.  I think that would be bad if one provider did broadcast because it would retry `Swap` and it never would work again.

Talking that out, I think that if one provider succeeds, `sendSignedTransaction` should not error from this.

buck54321:
> I think that if one provider succeeds, `sendSignedTransaction` should not error from this.

I agree.

client/asset/eth/multirpc.go
```diff
@@ -985,6 +990,61 @@ func (m *multiRPCClient) withOne(ctx context.Context, providers []*provider, f f
...
+		err := f(ctx, p)
+		cancel()
+		if err == nil {
+			atLeastOne = true
+			continue
+		}
+		var discarded bool
+		for i, f := range acceptabilityFilters {
+			discard, propagate, fail := f(err)
+			discarded = discard && (discarded || i == 0)
```
chappjc, line 1011: I will revise this so `atLeastOne` means what it suggests, and that no error is returned from it.

`propagate` itself might be problematic in this method as well as it contradicts our intent with `atLeastOne`.

chappjc: Leaving as-is, with docs to clarify the function of `withAll`.  A nil error sets `atLeastOne=true`, causing `nil` to be returned to the caller of `withAll`, unless a subsequent provider hits an explicitly "propagated" error.  There are no such current uses, but it should be sensible for a future use case that has a propagateable error in mind.

On a related note, I've concluded that there is no valid use case for more than 1 filter, and the argument is variadic simply to allow 0 or 1.  I'm still running all filters if there are more than one, but it's not clear why a given error would match more than one filter or why it would even be required to define more than a single filter at all, hence my conclusion.

buck54321 (approved):

Issue #2271, opened 2023-03-30
eth: domain helper is misused, and does not handle IP addresses


chappjc:
https://github.com/decred/dcrdex/blob/ff852e67bed8961a5dd8f13e5906ef6cb15049f3/client/asset/eth/multirpc.go#L663-L666

The `domain` helper function is only intended for use with a host name (e.g. `provider.host`), but recently it was applied to the full endpoint address, which could be a URL.  To use `domain` for compliant provider matching, we would need to parse the URL first, as we did in `connectProviders`, which used to precede this check.  @JoeGruffins 

A different issue is that `domain` does not work as expected with IP addresses like "127.0.0.1" where it says the domain is "0.1".

Lastly, `func failedProviders(succeeded []*provider, tried []string) string` also misused `domain` since the `tried` slice is the full endpoints, not the hosts.

JoeGruffins:
Will attempt to fix.

PR #2273, opened 2023-03-30
client/asset/eth: fix failedProviders string misusing domain helper

chappjc:
Related to https://github.com/decred/dcrdex/issues/2271

The stops using `domain` to shorten the failed providers string, which is purely informational, since `domain` only works for a hostname, not a full endpoint address, and there are IP address issues to resolve.


buck54321 (approved):

chappjc:
**Before**

image

**After**

image

PR #2275, opened 2023-03-30
client/core: Refactor rotateBonds

martonp:
This changes no functionality, just extracts the login in rotateBonds into multiple smaller functions.

It was a bit daunting to add the logic for updating existing ETH bonds. This will make it simpler.

### DIFF
--- a/client/core/bond.go
+++ b/client/core/bond.go
@@ -275,7 +275,7 @@ type acctUpdate struct {
 
 // refundExpiredBonds refunds expired bonds and returns the list of bonds that
 // have been refunded and their assetIDs.
-func (c *Core) refundExpiredBonds(ctx context.Context, dc *dexConnection, state *dexAcctBondState, now int64) (map[uint32]struct{}, []*bondID, error) {
+func (c *Core) refundExpiredBonds(ctx context.Context, acct *dexAccount, cfg *dexBondCfg, state *dexAcctBondState, now int64) (map[uint32]struct{}, int64, error) {
 	spentBonds := make([]*bondID, 0, len(state.expiredBonds))
 	assetIDs := make(map[uint32]struct{})
 
@@ -295,7 +295,7 @@ func (c *Core) refundExpiredBonds(ctx context.Context, dc *dexConnection, state 
 			continue
 		}
 		if _, ok := wallet.Wallet.(asset.Bonder); !ok { // will fail in RefundBond, but assert here anyway
-			return nil, nil, fmt.Errorf("Wallet %v is not an asset.Bonder", unbip(bond.AssetID))
+			return nil, 0, fmt.Errorf("Wallet %v is not an asset.Bonder", unbip(bond.AssetID))
 		}
 
 		expired, err := wallet.LockTimeExpired(ctx, time.Unix(int64(bond.LockTime), 0))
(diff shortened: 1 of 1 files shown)


client/core/bond.go
```diff
@@ -150,337 +150,304 @@ func minBondLifetime(net dex.Network, bondExpiry int64) time.Duration {
...
+		assetIDs[assetID] = struct{}{}
+	}
+
+	return assetIDs, spentBonds, nil
+}
+
+// clearSpentBonds removes all bonds that have been successfully refunded from
+// a dexConnection's expiredBonds list and returns the strength of the remaining
+// bonds.
+func (c *Core) clearSpentBonds(dc *dexConnection, bondCfg *dexBondCfg, spentBonds []*bondID) int64 {
```
chappjc, line 395: Why does this one need to be separate, rather than part of `refundExpiredBonds`?

client/core/bond.go
```diff
@@ -150,337 +150,304 @@ func minBondLifetime(net dex.Network, bondExpiry int64) time.Duration {
...
+		// tierChange field, but it could change, so we remember it.
+		actuatedTierChanges[state.bondAssetID] = append(actuatedTierChanges[state.bondAssetID],
+			acctUpdate{dc.acct, state.tierChange, reserveDelta},
+		)
+	}
+}
 
+// applyReserveAdjustments notifies wallets of the aggregate reserve
+// adjustments over all dex connections.
+func (c *Core) applyReserveAdjustments(netReserveDeltas map[uint32]int64, actuatedTierChanges map[uint32][]acctUpdate) {
```
chappjc, line 450: This body of this reserves helper and `recordReserveAdjustments` look better suited to stay in `rotateBonds`.  Is this also important for ETH bond renewal/extension?

chappjc (approved):
I honestly miss the linearly flowing code, but this looks fine and I think I understand the need.
There don't appear to be any issues or change of behavior with this.

client/core/bond.go
martonp: No need, I've moved it.

client/core/bond.go
martonp: I've moved it back there. The point if this was that this function was already ~500 lines and the update bond code will also add a good amount of additional logic. Personally I think it's very hard to wrap my head around these huge functions. If the details are extracted into smaller helper functions, it's easier to quickly know what is happening and make changes. Why do you like the linear flowing code better?

chappjc: > Why do you like the linear flowing code better?

Just readable for me is all.

Regarding the reserves adjustments, when I see variables like `netReserveDeltas` and `actuatedTierChanges` (and these two methods that need them) that are only instantiated or used in one other method, it stands out.  The functionality in either of these methods is useless unless applied exactly how `rotateBonds` applies it, and they'd always be used in succession.  So for me it's confusing (bounce around the code and wonder why they are separate methods), and it adds more the `Core`'s method set, which is not such a good thing even if they are not exported.

client/core/bond.go
```diff
@@ -580,11 +573,38 @@ func (c *Core) rotateBonds(ctx context.Context) {
...
+			if err != nil { // we grabbed it above so shouldn't happen
+				c.log.Errorf("%v wallet not available for bond: %v", unbip(assetID), err)
+				continue
+			}
+			c.log.Infof("Updating bond reserves by %s (automatic tier change adjustments)",
+				wallet.amtStringSigned(reserveDelta))
+			wallet.ReserveBondFunds(reserveDelta, false)
+		}
+		// Update the unactuated tier change counter.
+		for _, s := range actuatedTierChanges[assetID] {
```
chappjc: Thanks for switching this back.  The `netReserveDeltas` and `actuatedTierChanges` variables are just naturally locally scoped, and the helper methods shouldn't be used in any other context, so they are baggage and make it hard to see what's really being done.  I think these lines are not so long that it makes `rotateBonds` unreadable for you.  I respect the differences of opinion, and I do think the other refactoring mostly helps.

chappjc: Oh, looks like you need to remove the two helper functions.

chappjc (approved):

PR #2278, opened 2023-03-31
ui: consistent rate formatting

norwnd:
As a follow-up to https://github.com/decred/dcrdex/issues/2174 we still have some rates in order book (& depth/candle charts) with trailing zeros:

image

in this PR I'm basically truncating those leading 0s, which works because those rates are already complying with their respective rate steps, so it looks more like this:

image

### DIFF
--- a/client/webserver/site/src/js/charts.ts
+++ b/client/webserver/site/src/js/charts.ts
@@ -385,7 +385,7 @@ export class DepthChart extends Chart {
   book: OrderBook
   zoomLevel: number
   lotSize: number
-  rateStep: number
+  conventionalRateStep: number
   lines: DepthLine[]
   markers: Record<string, DepthMarker[]>
   zoomInBttn: Region
@@ -454,11 +454,10 @@ export class DepthChart extends Chart {
   }
 
   // set sets the current data set and draws.
-  set (book: OrderBook, lotSize: number, rateStep: number, baseUnitInfo: UnitInfo, quoteUnitInfo: UnitInfo) {
+  set (book: OrderBook, lotSize: number, rateStepEnc: number, baseUnitInfo: UnitInfo, quoteUnitInfo: UnitInfo) {
     this.book = book
     this.lotSize = lotSize / baseUnitInfo.conventional.conversionFactor
-    const [qFactor, bFactor] = [quoteUnitInfo.conventional.conversionFactor, baseUnitInfo.conventional.conversionFactor]
-    this.rateStep = rateStep / RateEncodingFactor * qFactor / bFactor
+    this.conventionalRateStep = Doc.conventionalRateStep(rateStepEnc, baseUnitInfo, quoteUnitInfo)
     this.baseUnit = baseUnitInfo.conventional.unit
     this.quoteUnit = quoteUnitInfo.conventional.unit
     if (!this.zoomLevel) {
@@ -588,7 +587,7 @@ export class DepthChart extends Chart {
 
     // Print the x labels
     const xLabels = makeLabels(ctx, this.plotRegion.width(), dataExtents.x.min,
-      dataExtents.x.max, 100, this.rateStep, '')
+      dataExtents.x.max, 100, this.conventionalRateStep, '')
 
     this.plotXLabels(xLabels, low, high, [`${this.quoteUnit}/`, this.baseUnit])
 
@@ -611,7 +610,7 @@ export class DepthChart extends Chart {
       ctx.textBaseline = 'middle'
       ctx.fillStyle = this.theme.value
       const y = 0.5 * dataExtents.y.max
-      ctx.fillText(formatLabelValue(midGap), tools.x(midGap), tools.y(y))
+      ctx.fillText(Doc.formatFourSigFigs(midGap), tools.x(midGap), tools.y(y))
       ctx.font = '12px \'sans\', sans-serif'
       // ctx.fillText('mid-market price', tools.x(midGap), tools.y(y) + 24)
       ctx.fillText(`${(gapWidth / midGap * 100).toFixed(2)}% spread`,
@@ -928,7 +927,7 @@ export class CandleChart extends Chart {
 
     // Apply labels.
     const rFactor = this.rateConversionFactor
-    this.doYLabels(this.candleRegion, rateStep, this.market.quotesymbol, v => formatLabelValue(v / rFactor))
+    this.doYLabels(this.candleRegion, rateStep, this.market.quotesymbol, v => Doc.formatFourSigFigs(v / rFactor))
     this.candleRegion.extents.x.min = this.yRegion.extents.x.max
     this.volumeRegion.extents.x.min = this.yRegion.extents.x.max
 
@@ -1325,7 +1324,7 @@ function makeLabels (
   unit: string,
   valFmt?: (v: number) => string
 ): LabelSet {
-  valFmt = valFmt || formatLabelValue
+  valFmt = valFmt || Doc.formatFourSigFigs
   const n = screenW / spacingGuess
   const diff = max - min
   if (n < 1 || diff <= 0) return { lbls: [] }
@@ -1435,17 +1434,6 @@ function clamp (v: number, min: number, max: number): number {
   if (v < min) return min
   if (v > max) return max
   return v
-}
-
-/* labelSpecs is specifications for axis tick labels. */
-const labelSpecs = {
-  minimumSignificantDigits: 4,
-  maximumSignificantDigits: 5
-}
-
-/* formatLabelValue formats the provided value using the labelSpecs format. */
-function formatLabelValue (x: number) {
-  return x.toLocaleString('en-us', labelSpecs)
 }
 
 /* floatCompare compares two floats to within a tolerance of  1e-8. */
--- a/client/webserver/site/src/js/doc.ts
+++ b/client/webserver/site/src/js/doc.ts
@@ -5,6 +5,7 @@ import {
   WalletState,
   PageElement
 } from './registry'
+import { RateEncodingFactor } from './orderutil'
 
 const parser = new window.DOMParser()
 
@@ -27,20 +28,22 @@ const BipIDs: Record<number, string> = {
 
 const BipSymbols = Object.values(BipIDs)
 
+const log10RateEncodingFactor = Math.round(Math.log10(RateEncodingFactor))
+
 const intFormatter = new Intl.NumberFormat((navigator.languages as string[]))
 
-const threeSigFigs = new Intl.NumberFormat((navigator.languages as string[]), {
-  minimumSignificantDigits: 3,
-  maximumSignificantDigits: 3
+const fourSigFigs = new Intl.NumberFormat((navigator.languages as string[]), {
+  minimumSignificantDigits: 4,
+  maximumSignificantDigits: 4
 })
 
-const fiveSigFigs = new Intl.NumberFormat((navigator.languages as string[]), {
-  minimumSignificantDigits: 5,
-  maximumSignificantDigits: 5
+const oneFractionalDigit = new Intl.NumberFormat((navigator.languages as string[]), {
+  minimumFractionDigits: 1,
+  maximumFractionDigits: 1
 })
 
 /* A cache for formatters used for Doc.formatCoinValue. */
-const decimalFormatters = {}
+const decimalFormatters: Record<number, Intl.NumberFormat> = {}
 
 /*
  * decimalFormatter gets the formatCoinValue formatter for the specified decimal
@@ -51,25 +54,25 @@ function decimalFormatter (prec: number) {
 }
 
 /* A cache for formatters used for Doc.formatFullPrecision. */
-const fullPrecisionFormatters = {}
+const fullPrecisionFormatters: Record<number, Intl.NumberFormat> = {}
 
 /*
  * fullPrecisionFormatter gets the formatFullPrecision formatter for the
  * specified decimal precision.
  */
-function fullPrecisionFormatter (prec: number) {
-  return formatter(fullPrecisionFormatters, prec, prec)
+function fullPrecisionFormatter (prec: number, locales?: string | string[]) {
+  return formatter(fullPrecisionFormatters, prec, prec, locales)
 }
 
 /*
  * formatter gets the formatter from the supplied cache if it already exists,
  * else creates it.
  */
-function formatter (formatters: Record<string, Intl.NumberFormat>, min: number, max: number): Intl.NumberFormat {
+function formatter (formatters: Record<string, Intl.NumberFormat>, min: number, max: number, locales?: string | string[]): Intl.NumberFormat {
   const k = `${min}-${max}`
   let fmt = formatters[k]
   if (!fmt) {
-    fmt = new Intl.NumberFormat((navigator.languages as string[]), {
+    fmt = new Intl.NumberFormat(locales || navigator.languages as string[], {
       minimumFractionDigits: min,
       maximumFractionDigits: max
     })
@@ -259,15 +262,20 @@ export default class Doc {
     return decimalFormatter(prec).format(v)
   }
 
-  static formatThreeSigFigs (v: number): string {
-    if (v >= 1000) return intFormatter.format(Math.round(v))
-    return threeSigFigs.format(v)
+  /*
+   * formatRateFullPrecision formats rate to represent it exactly at rate step
+   * precision, trimming non-effectual zeros if there are any.
+   */
+  static formatRateFullPrecision (encRate: number, bui: UnitInfo, qui: UnitInfo, rateStepEnc: number) {
+    const r = bui.conventional.conversionFactor / qui.conventional.conversionFactor
+    const convRate = encRate * r / RateEncodingFactor
+    const rateStepDigits = log10RateEncodingFactor - Math.floor(Math.log10(rateStepEnc)) -
+      Math.floor(Math.log10(bui.conventional.conversionFactor) - Math.log10(qui.conventional.conversionFactor))
+    return fullPrecisionFormatter(rateStepDigits).format(convRate)
   }
 
-  static formatFiveSigFigs (v: number, prec?: number): string {
-    if (v >= 10000) return intFormatter.format(Math.round(v))
-    else if (v < 1e5) return fullPrecisionFormatter(prec ?? 8 /* rate encoding factor */).format(v)
-    return fiveSigFigs.format(v)
+  static formatFourSigFigs (n: number): string {
+    return formatSigFigsWithFormatters(oneFractionalDigit, fourSigFigs, n)
   }
 
   /*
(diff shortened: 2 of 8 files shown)


client/webserver/site/src/js/charts.ts
```diff
@@ -1437,17 +1437,6 @@ function clamp (v: number, min: number, max: number): number {
...
-/* labelSpecs is specifications for axis tick labels. */
-const labelSpecs = {
-  minimumSignificantDigits: 4,
-  maximumSignificantDigits: 5
-}
-
-/* formatLabelValue formats the provided value using the labelSpecs format. */
-function formatLabelValue (x: number) {
-  return x.toLocaleString('en-us', labelSpecs)
-}
```
norwnd, line 1449: I didn't quite get the idea behind formatting those values this way, but the way I'm formatting now in this PR seems reasonable from what testing I've done.

client/webserver/site/src/js/doc.ts
```diff
@@ -263,6 +272,11 @@ export default class Doc {
...
   }
 
+  static formatFourSigFigs (v: number) {
+    if (v >= 1000 || Math.round(v) === v) return oneFractionalDigit.format(v)
+    return fourSigFigs.format(v)
+  }
+
   static formatFiveSigFigs (v: number, prec?: number): string {
     if (v >= 10000) return intFormatter.format(Math.round(v))
     else if (v < 1e5) return fullPrecisionFormatter(prec ?? 8 /* rate encoding factor */).format(v)
```
norwnd, line 282: I moved `fourSigFigs` func into `doc.ts` (and renamed it to `formatFourSigFigs `) so it is accessible both in `markets.ts` and `charts.ts`, but turns out there already are `formatThreeSigFigs` and `formatFiveSigFigs ` that are pretty similar but use slightly different formatting rules ... which seems somewhat weird to me. Can't we use same formatting for all of these ?

norwnd:
Note, there is also a room to improve `qty` displayed in order book (which on depth chart is trimmed to 1 digit after the dot, is it integer-only in practice ?):

image

but that won't work while we are showing Market orders in the book with `qty` in opposite units: 

image

which is counterintuitive on its own and probably needs changing.

norwnd:
Also note, switching to using `fourSigFigs` in this PR makes those rates displayed in the book subject to the same [rounding caveat mentioned here](https://github.com/decred/dcrdex/pull/2281#issuecomment-1492852434).

client/webserver/site/src/js/doc.ts
buck54321: You're right. It is weird. We really need consistent guidelines for display of numerical values. As I've pointed out [elsewhere](https://github.com/decred/dcrdex/pull/2281#issuecomment-1498181498), I typically draw a distinction between "display" values and "data" values. For "display" values, I think sig figs is appropriate, especially for numbers that have decimals (but for large numbers, I wouldn't, for instance, truncate 1,234,567 to 1,230,000). I personally feel like nobody ever needs more than 3 significant figures for display values, but I would entertain arguments for 4 or 5 sig figs too. For "data" values, for which it might be necessary to validate the number against e.g. an order, the full value should always be displayed, out to the number of decimals derived from `UnitInfo` and/or `RateStep`.

buck54321: One problem though is that Javascript seems to prioritize `minimumSignificantDigits` over `maximumFractionalDigits`. That's why we end up with values like `0.00000001000` for `fourSigFigs.format(0.00000001)`. I guess a question I have is whether we can omit the `minimumSignificantDigits` specifier in all cases.

chappjc: Right, that's exactly why we unfortunately have to do this `(v >= 1000)` else if `(v < 1e5)` business.  The ability to use the *significant* digits and *fraction* digits options together effectively is considered an [experiential option](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#roundingpriority).  https://github.com/decred/dcrdex/pull/2208#discussion_r1134733483

Generally though I find 4 sig figs for smaller values to be the sweet spot, but as you said whole numbers precision should not be lost, only fractional values.

norwnd: I agree with all of the above,

> For "data" values the full value should always be displayed

we can't address this ^ in general by simply changing how we are using/combining different rounding modifiers (_significant_ digits and _fraction_ digits) because [JS Number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number#number_encoding) itself can't represent what is actually **arbitrary precision Decimal**.

Maybe in practice we don't need to (if those cases when we don't get precise representation for order rate/qty/... are very rare on DEX, and we can adjust those _significant_ digits and _fraction_ digits modifiers such that they fit our use-cases most closely. Otherwise we need to figure out how to handle Decimals precisely (use some 3rd party lib to handle it for us, for example), but that's .

chappjc: As you point out, this is why we use integer representations of quantities and rates on the backend.  These numbers find also their way to the frontend.  Particularly with the rates, these whole numbers do not exceed 2^53 – 1 so there is no precision loss because the values fit into the IEEE 754 64-bit float's significand.

Only when we start _doing math_ with these on the frontend can we run into trouble.  When we divide the whole number "message rate" by the "rate encoding factor" (1e8), we get the decimal value.  If the float string formatters are used with the appropriate "rate step" or even the full 8 fractional digits of precision, it is rounded so that there is no incorrect or misleading information presented (the order of precision loss is far smaller than the number of digits we retain and the formatters do not simply truncate).  We only need to request the appropriate number of digits for a rate based on the market's rate step.

norwnd:
Reworked things in that last force-push, 

1st commit applies these changes https://github.com/buck54321/dcrdex/compare/2e0968c...buck54321:dcrdex:four-sig-figs

2nd commit: 
- is using those ^ formatters **buck54321** wrote for "display" values (when rounding is acceptable), also changed 3 and 5 figure rounding to 4 everywhere
- using "to rate-step" representation for "Your Orders" section (effectively addressing https://github.com/decred/dcrdex/pull/2281#issuecomment-1492852434 issue)
- applies exact "to-rate-step" representation to trim non-effectual 0s in orderbook table for **rate** (leaving **qty** formatting there as is)

in UI it looks like this now:

image

client/webserver/site/src/js/doc.ts
```diff
@@ -291,6 +296,15 @@ export default class Doc {
     return fullPrecisionFormatter(prec).format(value)
   }
 
+  static decodeRateStep (rateStepEnc: number, baseUnitInfo: UnitInfo, quoteUnitInfo: UnitInfo) {
+    const [qFactor, bFactor] = [quoteUnitInfo.conventional.conversionFactor, baseUnitInfo.conventional.conversionFactor]
+    return rateStepEnc / (RateEncodingFactor * qFactor / bFactor)
+  }
```
norwnd, line 302: Maybe we could "decode" rate step in a simpler manner ? Just asking, in case I missed something.

chappjc: I wish this were easier.  I'm constantly forgetting that all asset's conversion factors are not 1e8 anymore.

Although I really don't know if this is the best order of operations.  Elsewhere we've found this was best: https://github.com/decred/dcrdex/pull/1733/files#diff-8466f141b8e33a7e492b165ae7b6509808343604c759f3a0cc36e06fc33f13efR892  https://github.com/decred/dcrdex/pull/1733#issuecomment-1303080513

client/webserver/site/src/js/charts.ts
```diff
@@ -454,11 +454,10 @@ export class DepthChart extends Chart {
   }
 
   // set sets the current data set and draws.
-  set (book: OrderBook, lotSize: number, rateStep: number, baseUnitInfo: UnitInfo, quoteUnitInfo: UnitInfo) {
+  set (book: OrderBook, lotSize: number, rateStepEnc: number, baseUnitInfo: UnitInfo, quoteUnitInfo: UnitInfo) {
     this.book = book
     this.lotSize = lotSize / baseUnitInfo.conventional.conversionFactor
-    const [qFactor, bFactor] = [quoteUnitInfo.conventional.conversionFactor, baseUnitInfo.conventional.conversionFactor]
-    this.rateStep = rateStep / RateEncodingFactor * qFactor / bFactor
+    this.rateStep = Doc.decodeRateStep(rateStepEnc, baseUnitInfo, quoteUnitInfo)
```
buck54321, line 460: A couple of things here. 1) it appears that this was wrong? See other comment on `bFactor / qFactor`. 2) `this.rateStep` should really be called `this.conventionalRateStep`, but not really your problem here, so whatever.

client/webserver/site/src/js/doc.ts
```diff
@@ -258,15 +259,19 @@ export default class Doc {
...
   }
 
-  static formatThreeSigFigs (v: number): string {
-    if (v >= 1000) return intFormatter.format(Math.round(v))
-    return threeSigFigs.format(v)
+  /*
+   * formatRate formats rate to represent it exactly at rate step precision,
+   * trimming non-effectual zeros if there are any.
+   */
+  static formatRateToRateStep (encRate: number, bui: UnitInfo, qui: UnitInfo, rateStepEnc: number) {
```
buck54321, line 266: We are formatting an **encoded rate** to a **conventional rate**. Can the method name reflect that? This is also full-precision, so maybe a name like `formatRateFullPrecision` would be more apt.

client/webserver/site/src/js/doc.ts
```diff
@@ -291,6 +296,15 @@ export default class Doc {
...
   }
 
+  static decodeRateStep (rateStepEnc: number, baseUnitInfo: UnitInfo, quoteUnitInfo: UnitInfo) {
+    const [qFactor, bFactor] = [quoteUnitInfo.conventional.conversionFactor, baseUnitInfo.conventional.conversionFactor]
+    return rateStepEnc / (RateEncodingFactor * qFactor / bFactor)
+  }
+
+  static rateStepDigits (rateStepDec: number) {
+    return Math.round(Math.log10(1 / rateStepDec))
+  }
```
buck54321, line 306: You might be onto something here using the conventional ratestep to determine the number of decimals to show. But, the argument is not just decoded (`rateStepDec`), it's conventional, so the argument name should reflect that. Also, this is only used once. Do we really need a method for this?

client/webserver/site/src/js/doc.ts
buck54321, line 266: I've explored some tests to see if we're getting what we expect from this method. 

https://github.com/buck54321/dcrdex/compare/1a17cc4...buck54321:dcrdex:full-precision-rate-tests

buck54321 (changes requested):

client/webserver/site/src/js/doc.ts
norwnd: Should we also adjust `ConventionalRateAlt` then to do something like this ?
```
return float64(msgRate) * (float64(baseFactor) / float64(quoteFactor)) / RateEncodingFactor
```
It's doing this currently:
```
return float64(msgRate) / RateEncodingFactor * float64(baseFactor) / float64(quoteFactor)
```

client/webserver/site/src/js/doc.ts
norwnd: Nice, I was lazy to figure out how to run those, now I did,

I'm not sure how to use `npm run watch` properly (and it seems to be putting significant load on my laptop), so [I added `npm run build_dev`](https://github.com/decred/dcrdex/pull/2278/commits/cea84c2bb2c2b71282ff30c8e288e7737e226494) which is similar to `npm run build` but allows for accessing those tests ... if you don't have better suggestions for me on this.

norwnd: > so maybe a name like formatRateFullPrecision would be more apt.

Yep, sounds good, full precision also implies that it is a **conventional** rate. [Applied](https://github.com/decred/dcrdex/pull/2278/commits/fc82f2b2fa856572a36d9ee64a0cd168bc78e0a6).

client/webserver/site/src/js/charts.ts
norwnd: > 1) it appears that this was wrong? See other comment on bFactor / qFactor

Yep, it contradicts this:
```
func ConventionalRateAlt(msgRate uint64, baseFactor, quoteFactor uint64) float64 {
	return float64(msgRate) / RateEncodingFactor * float64(baseFactor) / float64(quoteFactor)
}
```

> 2) this.rateStep should really be called this.conventionalRateStep

I would like to rename it, while we are at it, [done here](https://github.com/decred/dcrdex/pull/2278/commits/fc82f2b2fa856572a36d9ee64a0cd168bc78e0a6).

client/webserver/site/src/js/doc.ts
```diff
@@ -291,6 +296,15 @@ export default class Doc {
     return fullPrecisionFormatter(prec).format(value)
   }
 
+  static decodeRateStep (rateStepEnc: number, baseUnitInfo: UnitInfo, quoteUnitInfo: UnitInfo) {
+    const [qFactor, bFactor] = [quoteUnitInfo.conventional.conversionFactor, baseUnitInfo.conventional.conversionFactor]
+    return rateStepEnc / (RateEncodingFactor * qFactor / bFactor)
```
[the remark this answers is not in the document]
norwnd: [Fixed](https://github.com/decred/dcrdex/pull/2278/commits/fc82f2b2fa856572a36d9ee64a0cd168bc78e0a6).

client/webserver/site/src/js/doc.ts
```diff
@@ -291,6 +296,15 @@ export default class Doc {
     return fullPrecisionFormatter(prec).format(value)
   }
 
+  static decodeRateStep (rateStepEnc: number, baseUnitInfo: UnitInfo, quoteUnitInfo: UnitInfo) {
```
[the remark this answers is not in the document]
norwnd: Got a bit confused by `conventional` and `decoded`, kinda assuming both to be same thing. Seems clear now, [applied](https://github.com/decred/dcrdex/pull/2278/commits/fc82f2b2fa856572a36d9ee64a0cd168bc78e0a6).

client/webserver/site/src/js/doc.ts
```diff
@@ -625,3 +639,51 @@ function timeMod (t: number, dur: number) {
...
+      ['fr-FR', '123.45678', undefined, '123,5'], // oneFractionalDigit
+      ['fr-FR', '1234.5', undefined, '1 234,5'], // U+202F for thousands separator
+      // For Arabic, https://www.saitak.com/number is useful, but seems to use
+      // slightly different unicode points and no thousands separator. I think
+      // the Arabic decimal separator is supposed to be more like a point, not
+      // a comma, but Google Chrome uses U+066B (Arabic Decimal Separator),
+      // which looks like a comma to me. ¯\_(ツ)_/¯
+      ['ar-EG', '123.45678', undefined, '١٢٣٫٥'],
+      ['ar-EG', '1234', undefined, '١٬٢٣٤٫٠'],
+      ['ar-EG', '0.12345', 3, '٠٫١٢٣']
```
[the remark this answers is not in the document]
norwnd: So it is test-specific issue (and using `navigator.languages` is what we want users to use in practice), right ? We can use something like this to accommodate this test (it passes now): https://github.com/decred/dcrdex/pull/2278/commits/2577606fed7cf1bc034d1165da05462ed5ebf977

The reason I wanted to remove `preciseFormatter` from there is because it kinda duplicates `maxDecimals` param (according to my understanding) and it isn't clear what happens if `preciseFormatter` has precision set to anything different from `maxDecimals` value.

image

client/webserver/site/src/js/doc.ts
```diff
@@ -258,15 +259,19 @@ export default class Doc {
...
-    return threeSigFigs.format(v)
+  /*
+   * formatRate formats rate to represent it exactly at rate step precision,
+   * trimming non-effectual zeros if there are any.
+   */
+  static formatRateToRateStep (encRate: number, bui: UnitInfo, qui: UnitInfo, rateStepEnc: number) {
+    const r = bui.conventional.conversionFactor / qui.conventional.conversionFactor
+    const convRate = encRate * r / RateEncodingFactor
+    const rateStepDigits = Doc.rateStepDigits(Doc.decodeRateStep(rateStepEnc, bui, qui))
+    return convRate.toFixed(rateStepDigits)
```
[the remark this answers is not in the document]
norwnd: Should be [fixed here](https://github.com/decred/dcrdex/pull/2278/commits/2577606fed7cf1bc034d1165da05462ed5ebf977).

client/webserver/site/src/js/doc.ts
chappjc: Not sure, but if we change the backend we should use big.Float for the math.

client/webserver/site/src/js/doc.ts
chappjc: > I'm not sure how to use `npm run watch` properly

I think you have to use it with `--no-embed-site`, but I'm not sure either tbh.  It has been a while since I've done anything other than a full build.

buck54321 (changes requested):

chappjc:
Closing EOW.  Please finalize.

client/webserver/site/src/js/doc.ts
```diff
@@ -258,15 +262,20 @@ export default class Doc {
...
-    return threeSigFigs.format(v)
+  /*
+   * formatRateFullPrecision formats rate to represent it exactly at rate step
+   * precision, trimming non-effectual zeros if there are any.
+   */
+  static formatRateFullPrecision (encRate: number, bui: UnitInfo, qui: UnitInfo, rateStepEnc: number) {
+    const r = bui.conventional.conversionFactor / qui.conventional.conversionFactor
+    const convRate = encRate * r / RateEncodingFactor
+    const conventionalRateStep = Doc.conventionalRateStep(rateStepEnc, bui, qui)
+    const rateStepDigits = Math.round(Math.log10(1 / conventionalRateStep))
```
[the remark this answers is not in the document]
norwnd: Right, applied.

client/webserver/site/src/js/doc.ts
```diff
@@ -625,3 +639,92 @@ function timeMod (t: number, dur: number) {
...
 }
+
+function formatSigFigsWithFormatters (sigFigs: number, intFormatter: Intl.NumberFormat, sigFigFormatter: Intl.NumberFormat, n: number, maxDecimals?: number, locales?: string | string[]): string {
+  if (n >= Math.round(Math.pow(10, sigFigs - 1))) return intFormatter.format(n)
+  const s = sigFigFormatter.format(n)
+  if (typeof maxDecimals !== 'number') return s
+  const fractional = sigFigFormatter.formatToParts(n).filter((part: Intl.NumberFormatPart) => part.type === 'fraction')[0].value
+  if (fractional.length <= maxDecimals) return s
+  return fullPrecisionFormatter(maxDecimals, locales).format(n)
+}
```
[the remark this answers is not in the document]
norwnd: Like that https://github.com/decred/dcrdex/pull/2278/commits/02523369448861666fe34d4648709e77504b6cfe ?

client/webserver/site/src/js/doc.ts
```diff
@@ -50,25 +51,28 @@ function decimalFormatter (prec: number) {
...
  */
-function formatter (formatters: Record<string, Intl.NumberFormat>, min: number, max: number): Intl.NumberFormat {
+function formatter (formatters: Record<string, Intl.NumberFormat>, min: number, max: number, locales?: string | string[]): Intl.NumberFormat {
   const k = `${min}-${max}`
   let fmt = formatters[k]
   if (!fmt) {
-    fmt = new Intl.NumberFormat((navigator.languages as string[]), {
+    if (!locales) {
+      locales = navigator.languages as string[]
+    }
```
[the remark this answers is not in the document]
norwnd: Applied.

chappjc (approved):
Whenever it's good with you, @buck54321.
Ideally a squash and rebase from @norwnd before merging, but I can handle that if need be.

buck54321 (approved):

PR #2280, opened 2023-04-01
ui: price/change header/row discrepancy

norwnd:
`mkt.spot` might not be initialized when we are setting up left Market doc - we leave **Price**/**Change** values set at `-`. This is probably fine as long as these will be updated in the future, but they don't always update due to (looks like) a typo this PR fixes.

image

### DIFF
--- a/client/webserver/site/src/js/markets.ts
+++ b/client/webserver/site/src/js/markets.ts
@@ -2587,8 +2587,6 @@ class MarketList {
   updateSpots (note: SpotPriceNote) {
     for (const row of this.markets) {
       if (row.mkt.xc.host !== note.host) continue
-      const spot = note.spots[row.mkt.name]
-      if (!spot) continue
       const xc = app().exchanges[row.mkt.xc.host]
       const mkt = xc.markets[row.mkt.name]
       setPriceAndChange(row.tmpl, xc, mkt)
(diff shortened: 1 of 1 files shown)


norwnd:
I think this will also fix the 2nd aspect of this issue which is the discrepancy sometimes seen in Price/Change values in header and left-doc-market-row.

image

chappjc (approved):

client/webserver/site/src/js/markets.ts
```diff
@@ -2588,7 +2588,7 @@ class MarketList {
     for (const row of this.markets) {
       if (row.mkt.xc.host !== note.host) continue
       const spot = note.spots[row.mkt.name]
-      if (!spot) return
+      if (!spot) continue
```
buck54321, line 2591: Can we just delete these two lines? In **app.ts**, we're handling the spot note like
```Typescript
for (const [mktName, spot] of Object.entries(n.spots)) xc.markets[mktName].spot = spot
```
So the `mkt.spot` below is identical to `spot`, and in `setPriceAndChange`, the first line is
```Typescript
if (!mkt.spot) return
```

norwnd: I guess we can, seems to be working same without, applied.

buck54321 (approved):

Issue #2254, opened 2023-03-25
ui: display rate for the collapsed view of orders in the 'your orders' section


artikozel:
In v0.6-beta as it is currently the rate at which orders are placed are not displayed in the collapsed view of the order menu. I believe there is enough real estate there to display it alongside the order size either as a simple "@ rate" or in any way the devs see fit, because the only visual confirmation available at first glance are the little arrow pointers in the depth chart and this information is just as critical to trading as order size.
However, if said real estate was left empty for a reason, then perhaps indicating to the user where their orders sit in the orderbook with a simple dot icon next to the order in the orderbook view would suffice.
![image](https://user-images.githubusercontent.com/44892068/227707598-5a22f4e5-3f30-447b-9793-b42eeb86f314.png)

chappjc:
I agree with this too.  Would you object to rate here @buck54321?

buck54321:
Will it fit?

chappjc:
There seems to be amble blank space there.

![image](https://user-images.githubusercontent.com/9373513/228897924-c632d780-fc91-4ba6-855c-df588ef3f896.png)

buck54321:
Looks good to me

norwnd:
I can do it.

chappjc:
Go for it.  Tightly scoped diff please, since we might backport.

PR #2281, opened 2023-04-01
ui: display your orders rate in header

norwnd:
Resolves https://github.com/decred/dcrdex/issues/2254.

Rate fits reasonably well with the longest `canceled/settling` status and max rate step we currently have = 6

image

### DIFF
--- a/client/webserver/site/src/html/markets.tmpl
+++ b/client/webserver/site/src/html/markets.tmpl
@@ -390,6 +390,8 @@
                     <span data-tmpl="side"></span>
                     <span data-tmpl="qty" class="ms-1"></span>
                     <span data-tmpl="baseSymbol" class="ms-1 grey"></span>
+                    <span class="ms-1">@</span>
+                    <span data-tmpl="rate" class="ms-1"></span>
                     <span class="flex-grow-1 d-flex align-items-center justify-content-end">
                       <span data-tmpl="status"></span>
                       <span data-tmpl="activeLight" class="active-indicator active"></span>
--- a/client/webserver/site/src/js/markets.ts
+++ b/client/webserver/site/src/js/markets.ts
@@ -1284,7 +1284,7 @@ export default class MarketsPage extends BasePage {
       details.side.classList.add(ord.sell ? 'sellcolor' : 'buycolor')
       header.side.classList.add(ord.sell ? 'sellcolor' : 'buycolor')
       details.qty.textContent = header.qty.textContent = fourSigFigs(ord.qty / this.market.baseUnitInfo.conventional.conversionFactor)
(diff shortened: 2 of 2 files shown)


norwnd:
Note, `fourSigFigs` results in non-precise rate shown in "Your Orders" section (should be `1.000001`, not `1`):

image

Is it something we want to fix (assuming it's possible to do without using external decimal-arithmetic libs) ? Reading through [this discussion](https://github.com/decred/dcrdex/pull/2208#discussion_r1129324708) I've got a feeling that we don't care about this when displaying spot price (or similar). For order  rate however user might be somewhat uneasy to see such rounding imo.

chappjc (approved):

buck54321:
I draw a distinction between numerical "display" values and "data" values. Display values, such as the market statistics shown at the top of the markets view, can employ rounding or truncation and should use significant figures to make values "pretty". Data values, such as order data, should not employ rounding, and should use e.g. `UnitInfo.Conventional.ConversionFactor` to decide how many digits/decimals to display.

norwnd:
To resolve the issue mentioned in https://github.com/decred/dcrdex/pull/2281#issuecomment-1492852434 I'll need to apply similar fix implemented in https://github.com/decred/dcrdex/pull/2278#issuecomment-1500847677, for rate included in "header" in this PR.

chappjc:
Can we just get this merged with the current reasonable solution (perhaps just forcing at least one decimal)?  https://github.com/decred/dcrdex/pull/2278#issuecomment-1500847677 may contain the resolution, but is is a can of worms and will preculde this small change from landing in 0.6.1

chappjc:
1.000001 is a rate that isn't likely to be practical (too much precision), so I'm inclined to just merge this as-is if there's not a quick resolution tomorrow.
Or for the purposes of this PR (without https://github.com/decred/dcrdex/pull/2278#issuecomment-1500847677 and strict rate step observation), we could use `Doc.formatFiveSigFigs(rate)` instead of  of `fourSigFigs(rate)`.

norwnd:
> 1.000001 is a rate that isn't likely to be practical (too much precision), so I'm inclined to just merge this as-is if there's not a quick resolution tomorrow.

I'd rather not try to come up with yet another way to handle this specific case (and I don't see anything simple we can do here), so I'd wait until https://github.com/decred/dcrdex/pull/2278#issuecomment-1500847677 is ready,

but merging as is is also fine, I believe, because we show this trimmed rate in "details" anyway, so showing in header as well doesn't really break anything.

chappjc:
Right it's no deviation from existing convention.  I'm asking for @buck54321 call though regarding asap merge resolution.

buck54321 (approved):

PR #2283, opened 2023-04-03
Use zero IPv4 loopback address instead of IPv6 unspecified address for CSP connect-src directive

### LINKED ISSUE #1652: ui: some UI elements suddenly changed into weird symbols
Playing around in UI (before registering with a server, trying to add testnet server) suddenly I see this (many, but not all, elements start displaying weird symbol combinations instead of English-names):

image

image

image

image

screenshots above are taken in Brave browser (navigating UI without refreshing the page),  opening the same page in a different tab, opening in Firefox (and logging in) or refreshing the page turns everything back to normal,

haven't been able to reproduce 2nd time,
and didn't find anything related in `dexc` logs.

ukane-philemon:
- This fixes an issue where using the IPv6Unspecified address `[::]` for csp connect-src directive could cause issues resulting in some icons not loading properly.
Screenshot 2023-04-03 at 8 06 02 AM
- Add missing log arg.

Closes #1652 

@norwnd, please can you confirm that this PR fixes the issue on your end?

### DIFF
--- a/client/asset/eth/multirpc.go
+++ b/client/asset/eth/multirpc.go
@@ -261,7 +261,7 @@ func (p *provider) subscribeHeaders(ctx context.Context, sub ethereum.Subscripti
 				return sub, nil
 			}
 			if time.Since(lastWarning) > 5*time.Minute {
-				log.Warnf("can't resubscribe to %q headers: %v", err)
+				log.Warnf("can't resubscribe to %q headers: %v", p.host, err)
 			}
 			select {
 			case <-time.After(time.Second * 30):
--- a/client/webserver/webserver.go
+++ b/client/webserver/webserver.go
@@ -516,12 +516,19 @@ func (s *WebServer) Connect(ctx context.Context) (*sync.WaitGroup, error) {
 	if err != nil {
 		return nil, fmt.Errorf("Can't listen on %s. web server quitting: %w", s.addr, err)
 	}
-	// Update the listening address in case a :0 was provided.
-	s.addr = listener.Addr().String()
-	// Work around a webkit (safari) bug with the handling of the connect-src
-	// directive of content security policy. See:
-	// https://bugs.webkit.org/show_bug.cgi?id=201591
-	s.csp = fmt.Sprintf("%s ws://%s", baseCSP, s.addr)
+
+	addr, allowInCSP := prepareAddr(listener.Addr())
+	if allowInCSP {
+		// Work around a webkit (safari) bug with the handling of the
+		// connect-src directive of content security policy. See:
+		// https://bugs.webkit.org/show_bug.cgi?id=201591. TODO: Remove this
+		// workaround since the issue has been fixed in newer versions of
+		// Safari. When this is removed, the allowInCSP variable can be removed
+		// but prepareAddr should still return 127.0.0.1 for unspecified
+		// addresses.
+		s.csp = fmt.Sprintf("%s ws://%s", baseCSP, s.addr)
+	}
+	s.addr = addr
 
 	// Shutdown the server on context cancellation.
 	var wg sync.WaitGroup
@@ -562,6 +569,21 @@ func (s *WebServer) Connect(ctx context.Context) (*sync.WaitGroup, error) {
 	log.Infof("Web server listening on %s", s.addr)
 	fmt.Printf("\n\t****  OPEN IN YOUR BROWSER TO LOGIN AND TRADE  --->  http://%s  ****\n\n", s.addr)
 	return &wg, nil
+}
+
+// prepareAddr prepares the listening address in case a :0 was provided.
+func prepareAddr(addr net.Addr) (string, bool) {
+	// If the IP is unspecified, default to `127.0.0.1`. This is a workaround
+	// for an issue where all ip addresses other than exactly 127.0.0.1 will
+	// always fail to match when used in CSP directives. See:
+	// https://w3c.github.io/webappsec-csp/#match-hosts.
+	defaultIP := net.IP{127, 0, 0, 1}
+	tcpAddr, ok := addr.(*net.TCPAddr)
+	if ok && (tcpAddr.IP.IsUnspecified() || tcpAddr.IP.Equal(defaultIP)) {
+		return net.JoinHostPort(defaultIP.String(), strconv.Itoa(tcpAddr.Port)), true
+	}
+
+	return addr.String(), false
 }
 
 // authorize creates, stores, and returns a new auth token to identify the user.
--- a/client/webserver/webserver_test.go
+++ b/client/webserver/webserver_test.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"io"
 	"math/rand"
+	"net"
 	"net/http"
 	"os"
 	"strings"
@@ -1018,4 +1019,61 @@ func testTrade(t *testing.T, async bool) {
 
 	tCore.tradeErr = tErr
 	ensure(`{"ok":false,"msg":"expected dummy error"}`)
+}
+
+func Test_prepareAddr(t *testing.T) {
+	tests := []struct {
+		name       string
+		addr       net.Addr
+		allowInCSP bool
+		want       string
+	}{{
+		name: "OK: IPv4",
+		addr: &net.TCPAddr{
+			IP:   net.IPv4(127, 0, 0, 1),
+			Port: 7232,
+		},
+		want:       "127.0.0.1:7232",
+		allowInCSP: true,
+	}, {
+		name: "IPv6 loopback",
+		addr: &net.TCPAddr{
+			IP:   net.IPv6loopback,
+			Port: 7232,
+		},
+		want: "[::1]:7232",
+	}, {
+		name: "OK: IPv6 unspecified",
+		addr: &net.TCPAddr{
+			IP:   net.IPv6unspecified,
+			Port: 7232,
+		},
+		want:       "127.0.0.1:7232",
+		allowInCSP: true,
+	}, {
+		name: "OK: zero IPv4",
+		addr: &net.TCPAddr{
+			IP:   net.IPv4zero,
+			Port: 7232,
+		},
+		want:       "127.0.0.1:7232",
+		allowInCSP: true,
+	}, {
+		name: "others",
+		addr: &net.UDPAddr{
+			IP:   []byte{},
+			Port: 7232,
+		},
+		want: ":7232",
+	}}
+
+	for _, test := range tests {
+		gotAddr, allowInCSP := prepareAddr(test.addr)
+		if gotAddr != test.want {
+			t.Fatalf("%s: address: got %s, want %s", test.name, gotAddr, test.want)
+		}
+		if allowInCSP != test.allowInCSP {
+			t.Fatalf("%s: allow in CSP: got %v, want %v", test.name, allowInCSP, test.allowInCSP)
+		}
+	}
 }


client/webserver/webserver.go
```diff
@@ -516,8 +516,23 @@ func (s *WebServer) Connect(ctx context.Context) (*sync.WaitGroup, error) {
...
+	tcpAddr, ok := listener.Addr().(*net.TCPAddr)
+	if !ok {
+		return nil, fmt.Errorf("listener.Addr() is not a *net.TCPAddr")
+	}
+
+	// If the IP is the IPv6 address [::], use the zero IPv4 address instead.
+	// This is a workaround for an issue where browers see [::] as an invalid IP
+	// when used in the connect-src directive of a CSP like we do below.
+	if tcpAddr.IP.Equal(net.IPv6unspecified) {
+		tcpAddr.IP = net.IPv4zero
```
chappjc, line 530: `tcpAddr` is a pointer.  Don't modify the listener's `net.TCPAddr` that it owns.
We have no reason to change the listener, we would just need to change the CSP line given to the browser.  However, the premise of the fix is misinformed as my other comment describes.

chappjc (changes requested):
The issue goes beyond IPv6 vs. IPv4.  Setting a wildcard ([::] **or** 0.0.0.0) in the CSP source list is not fundamentally appropriate as it would completely defeat CSP.  Further, CSP would fail to even match the IPv4 address, even if the csp line would be considered valid.

The CSP spec: (a) fails to match *any* IPv4 address other than "127.0.0.1", and (b) considers **all** IPv6 addresses invalid.

https://w3c.github.io/webappsec-csp/#match-hosts

> NOTE: A future version of this specification may allow literal IPv6 and IPv4 addresses, depending on usage and demand. Given the weak security properties of IP addresses in relation to named hosts, however, authors are encouraged to prefer the latter whenever possible.

https://w3c.github.io/webappsec-csp/#framework-directive-source-list

> NOTE: Though IP address do match the grammar above, only 127.0.0.1 will actually match a URL when used in a source expression (see [§ 6.7.2.6 Does url match source list in origin with redirect count?](https://w3c.github.io/webappsec-csp/#match-url-to-source-list) for details). The security properties of IP addresses are suspect, and authors ought to prefer hostnames whenever possible.

The whole reason we were even trying to manually add `ws://127.0.0.1:5758` to the csp line was because of a [bug in the Safari browser](https://bugs.webkit.org/show_bug.cgi?id=201591) where `self` did not work as required, rejecting `ws://` urls, which appears to be long since fixed.

The **only** hosts that are permissible to add to the line is 127.0.0.1 or a resolvable hostname (not an IP).  If the listen address is something else, and their browser is affected by the bug that prevented `'self'` from working as expected, they are out of luck.

Our resolution should be just to limit what values of `s.addr` we allow to be appended to `s.csp`.

ukane-philemon:
> The CSP spec: (a) fails to match any IPv4 address other than "127.0.0.1", and (b) considers all IPv6 addresses invalid.

Interesting, thanks for looking up this information. It seems we only need to manually add `127.0.0.1` to the CSP directive `IF` the user does not specify a host or the host is not `127.0.0.1` because of the Safari  browser bug, correct?

> Ah, do the docker environments have IPv4 only, no IPv6? That could explain why they might work.

Since we are using `tcp` which will listen on all available IPv4 and IPv6 addresses if a host is not specified, using `0.0.0.0` will not cause an invalid csp error but without providing a host, `listener.Addr().String()` will return `[::]` which will cause the invalid csp error.

chappjc:
> > The CSP spec: (a) fails to match any IPv4 address other than "127.0.0.1", and (b) considers all IPv6 addresses invalid.
> 
> Interesting, thanks for looking up this information. It seems we only need to manually add `127.0.0.1` to the CSP directive `IF` the user does not specify a host or the host is not `127.0.0.1` because of the Safari browser bug, correct?

We should only add something to the csp line if it is 127.0.0.1 or a hostname.  At a bare minimum, we just need to reject IPv6.  IPv4 addresses won't throw the invalid line error in the browser, but they also will **never match** except in the case of 127.0.0.1.

> > Ah, do the docker environments have IPv4 only, no IPv6? That could explain why they might work.
> 
> Since we are using `tcp` which will listen on all available IPv4 and IPv6 addresses if a host is not specified, using `0.0.0.0` will not cause an invalid csp error but without providing a host, `listener.Addr().String()` will return `[::]` which will cause the invalid csp error.

I'm asking about what network stack the Docker environment has access to.   A host with no IPv6 stack will not have the socket translate the IPv4 wildcard into the IPv6 wildcard when creating the listener.

chappjc:
To clarify, this csp line twiddling was in fact required even when a host name of 127.0.0.1 was used with webaddr because `self` wasn't working in Safari for ws:// urls when the origin was http://

chappjc:
> > I'm asking about what network stack the Docker environment has access to.
> 
> In a Docker environment, the container has access to the network stack of the host machine it is running on. So yes, It does have access to IPv4 and IPv6.

It would seem to depend on the network driver you choose in the `docker run` command: https://docs.docker.com/network/

ukane-philemon:
Will revert these changes according to our discussion and also test on Safari too.

ukane-philemon:
It seems a better solution to the issue at hand is removing the safari bug workaround which must have been resolved by https://bugs.webkit.org/show_bug.cgi?id=235873. 

I investigated this issue on Safari browser version 15.6.1 (15613.3.9.1.16, 15613), and the extra csp directive only caused more issues(the browser was unable to connect to websocket when the host:port used in the url did not match the csp directive  ws://host:port).

I removed the workaround and everything was alright. However, using `[::]:port` on safari is invalid and `0.0.0.0:port` is considered insecure, so maybe we can log `0.0.0.0:port` instead of `[::]:port` for users that decides to use the URL logged when the web server is started.

chappjc:
Using a url in the browser with 0.0.0.0 makes no sense in general; only using that as a listen address makes sense to indicate listening on all interfaces.  So don't draw any conclusion as about that url. I'm shocked the browser even loaded anything.

However, all ip addresses other than exactly 127.0.0.1 will always fail to match, and all IPv6 are invalid.

Yes, I think this workaround is obsolete, but I don't know how many mac users are stuck with older browsers.

> the URL logged when the web server is started

Yeah, this needs to change. If the addr is an IP and is "unspecified", we've should substitute in either 127.0.0.1 or somehow get the actual addresses that were bound to.  The latter may not be possible.

chappjc:
For just release-v0.6, which I wanted to tag today: https://github.com/decred/dcrdex/pull/2287/files  
I've tested it with anything sensible that I could think of, but please look and test so we can merge that asap.

client/asset/eth/multirpc.go
```diff
@@ -259,7 +259,7 @@ func (p *provider) subscribeHeaders(ctx context.Context, sub ethereum.Subscripti
 				return sub, nil
 			}
 			if time.Since(lastWarning) > 5*time.Minute {
-				log.Warnf("can't resubscribe to %q headers: %v", err)
+				log.Warnf("can't resubscribe to %q headers: %v", p.host, err)
```
chappjc, line 262: Separate commit please.

client/webserver/webserver.go
```diff
@@ -564,6 +560,43 @@ func (s *WebServer) Connect(ctx context.Context) (*sync.WaitGroup, error) {
...
+	addr := l.Addr().String()
+	host, port, err := net.SplitHostPort(addr)
+	if err == nil && host != "" {
+		addr = host
+	} else {
+		// If SplitHostPort failed, IPv6 addresses may still have brackets.
+		addr = strings.Trim(addr, "[]")
+	}
+
+	ip := net.ParseIP(addr)
```
chappjc, line 578: Also, since this method is not so trivial, we'd want to unit test it, and that would be easiest if the input were not a `net.Listener`.  Only `l.Addr()` is used, so it's a bit pointless to make it accept a `net.Listener`.  I'd make it a vanilla function, not a method, and consider using the typed [`net.TCPAddress`](https://pkg.go.dev/net#TCPAddr) somehow to avoid redundant processing.  Creating the listener did all this work already, including splitting host:port, and parsing the IP.

Also, I cannot tell at a glance looking at this method how it deals with IPv6 in general (not just [::] or [::1]).  It should not allow _any_ IPv6 to enter the csp string.

client/webserver/webserver_test.go
```diff
@@ -1019,3 +1020,61 @@ func testTrade(t *testing.T, async bool) {
...
+		want:       "127.0.0.1:7232",
+		allowInCSP: true,
+	}, {
+		name: "OK: IPv6",
+		addr: &net.TCPAddr{
+			IP:   net.IPv6loopback,
+			Port: 7232,
+		},
+		want:       "127.0.0.1:7232",
+		allowInCSP: true,
```
chappjc, line 1045: IPv6 loopback shouldn't be translated into something else.  Don't modify the addr and don't allow in csp.

ukane-philemon: > IPv6 loopback shouldn't be translated into something else. 

I was thinking it's better we log `127.0.0.1` instead of `[::1]`, no? The average non-techy user would find `127.0.0.1 ` more appealing in the logs.

chappjc: They are not interchangeable.  Binding to `[::1]` does not mean you are also bound to 127.0.0.1

client/webserver/webserver.go
```diff
@@ -564,6 +571,21 @@ func (s *WebServer) Connect(ctx context.Context) (*sync.WaitGroup, error) {
...
 
+// prepareAddr prepares the listening address in case a :0 was provided.
+func prepareAddr(addr net.Addr) (string, bool) {
+	// If the IP is unspecified, defualt to `127.0.0.1`. This is a workaround
+	// for an issue where all ip addresses other than exactly 127.0.0.1 will
+	// always fail to match when used in CSP directives. See:
+	// https://w3c.github.io/webappsec-csp/#match-hosts.
+	defaultIP := net.IP{127, 0, 0, 1}
+	tcpAddr, ok := addr.(*net.TCPAddr)
+	if ok && (tcpAddr.IP.IsUnspecified() || tcpAddr.IP.Equal(net.IPv6loopback) || tcpAddr.IP.Equal(defaultIP)) {
```
chappjc, line 582: What's this `tcpAddr.IP.Equal(defaultIP)` condition for?

ukane-philemon: Was aiming at a one-liner, but maybe I should add a few more lines.

chappjc: OK, I suppose it's fine.  Was just confusing.  Looked like a bug.

client/asset/eth/multirpc.go
chappjc: This is not resolved.

client/webserver/webserver.go
```diff
@@ -564,6 +571,21 @@ func (s *WebServer) Connect(ctx context.Context) (*sync.WaitGroup, error) {
 	return &wg, nil
 }
 
+// prepareAddr prepares the listening address in case a :0 was provided.
+func prepareAddr(addr net.Addr) (string, bool) {
+	// If the IP is unspecified, defualt to `127.0.0.1`. This is a workaround
```
chappjc, line 576: defualt

client/asset/eth/multirpc.go
ukane-philemon: Do you mean separate PR?

chappjc: Which separate commit?  It's still part of the webserver commit

ukane-philemon: This: https://github.com/decred/dcrdex/pull/2283/commits/4be67857b718fabc6a57a4fa4c6a12671a259a11

chappjc: Oh you touch this line in three separate commits now.
Please squash and rebase things, and start that commit with `client/asset/eth: `

ukane-philemon: Okay.

ukane-philemon:
Rebased and squashed.

chappjc:
I want to rebase this PR into master directly to keep the eth commit separate, so I need you to reorder the commit so that you can squash it down to two commits.  Right now you have: webserver, eth, webserver fixup

chappjc (approved):
Ok.  Release already has a slightly different workaround, so no backport.

PR #2284, opened 2023-04-03
client: Remember last candle duration

### LINKED ISSUE #2198: candle chart doesn't remember last bin size selection or show current selection
When navigating away from and back to the markets view, the candles are always the 5-minute candles.

There is a reason that the default is 5 minutes. The `High24` and `Low24` fields of the `msgjson.Spot` were only added in #1825, so we get the values from the 5 minute candles until servers are up-to-date. I don't know if it makes sense to fix this before then or not.

Also, there is no indication of the currently selected bin size. It can be inferred by looking at the x-axis, but we also be more explicit, highlighting a button or even just displaying it somewhere.

ukane-philemon:
This adds the ability to remember a user's last candle duration selection and fixes an issue where the candle selection is not known when navigating away from and back to the market's view. Closes #2198


norwnd (commented):
Seems to be working!

buck54321:
I think this causes an issue with our 24-hour high/low values. See description from issue #2198, but we are expecting to have the 5-minute candles loaded. See also `setHighLow`.

ukane-philemon:
> I think this causes an issue with our 24-hour high/low values

@buck54321, we've been using the fiveMinBin in `setHighLow` as default but now we can just use the `spot.high24` and `spot.low24`  if the current candle dur is 24h, right?

Also, for the `1h` dur,  do we intend to do what Cache.Delta(...) does?  Or leave the existing behaviour? It seems all this data would eventually come from the backend in the future.

buck54321:
> > I think this causes an issue with our 24-hour high/low values
> 
> @buck54321, we've been using the fiveMinBin in `setHighLow` as default but now we can just use the `spot.high24` and `spot.low24` if the current candle dur is 24h, right?

The most recent release, v0.5.9, does not include the `high24` and `low24` fields.

> Also, for the `1h` dur, do we intend to do what Cache.Delta(...) does? Or leave the existing behaviour? It seems all this data would eventually come from the backend in the future.

It wouldn't be accurate enough to calculate the 24-hour change/high/low from the 1-hour bin. This is why I was forcing loading of the 5-min bins first.

ukane-philemon:
@buck54321, I'm forcing a request of the fiveMinBin if we don't have cache in https://github.com/decred/dcrdex/pull/2284/commits/1a82aaf8961afc3a7fb60842c58ecdc78613f74a, hope that resolves the potential issue with the 24high/low rates?

chappjc:
On account of the redefined candles in 0.6, if no candle bin size was selected, I believe the default should now be 1hr instead of 5min.  I think this change is needed now that 0.6 is faithfully producing candles from match data only, not just mid-book for epochs.  For the 5 min candles to be useful, matches need to be far more frequent.

ukane-philemon:
@chappjc what do you think?

> @buck54321- we've been using the fiveMinBin in setHighLow as default but now we can just use the spot.high24 and spot.low24 if the current candle dur is 24h, right?

client/webserver/site/src/js/markets.ts
```diff
@@ -802,6 +799,21 @@ export default class MarketsPage extends BasePage {
...
       page.durBttnBox.appendChild(bttn)
     }
+
+    // Check if we already have the fiveMinBin candles cache, if not, request
+    // for it first since we use it to determine the 24hour high/low in
+    // this.setHighLow().
+    const lastCandleDur = State.fetchLocal(State.lastCandleDurationLK)
+    const cache = this.market?.candleCaches[fiveMinBinKey]
+    if (!cache && lastCandleDur !== fiveMinBinKey) {
+      this.candleDur = fiveMinBinKey
```
chappjc, line 809: OK we super need this on 0.6, and to be changed to the 1hr bin key.
Can we get this updated today?

ukane-philemon: Are we going to use `spot.high24` and `spot.low24` now?

chappjc: It's available from all dex hosts I'm aware of.
What happens if it's not set?  Can we detect if it's zero and do something differently without big changes?

ukane-philemon: > Can we detect if it's zero and do something differently without big changes?

Yeah. Okay.

chappjc: Or to be safe we can keep the `lowRate`/`highRate` loop and just make the default candles 1hr.  I don't follow the problem exactly.  Maybe @buck54321 has a simple way to change the default while keeping the current high/low logic the way it is, not using the spot note, which could cause a regression.

chappjc: Like maybe after the initial `loadCandles` using `fiveMinBinKey` like it currently does, we just do a `this.candleDurationSelected('1hr')` right after it.  It's kinda dumb, but it might be fine.

chappjc:
```diff
diff --git a/client/webserver/site/src/js/markets.ts b/client/webserver/site/src/js/markets.ts
index 2e7451bd4..44664421e 100644
--- a/client/webserver/site/src/js/markets.ts
+++ b/client/webserver/site/src/js/markets.ts
@@ -85,6 +85,7 @@ const buyBtnClass = 'buygreen-bg'
 const sellBtnClass = 'sellred-bg'
 
 const fiveMinBinKey = '5m'
+const oneHrBinKey = '1h'
 
 const percentFormatter = new Intl.NumberFormat(document.documentElement.lang, {
   minimumFractionDigits: 1,
@@ -874,7 +875,8 @@ export default class MarketsPage extends BasePage {
 
     // candlesticks
     this.candleDur = fiveMinBinKey
-    this.loadCandles()
+    this.loadCandles() // gotta do this with fiveMinBinKey first to get high/low
+    this.candleDurationSelected(oneHrBinKey)
 
     this.setLoaderMsgVisibility()
     this.setRegistrationStatusVisibility()
```

That way we don't have to worry about an old server lacking high/low in the spot note.

client/webserver/site/src/js/markets.ts
```diff
@@ -598,24 +596,32 @@ export default class MarketsPage extends BasePage {
...
-      for (const s of this.stats) {
-        s.tmpl.high.textContent = '-'
-        s.tmpl.low.textContent = '-'
+    let [high, low] = [0, 0]
+    const spot = this.market.cfg.spot
+    // Use spot values for 24 hours high and low rates if it is available. We
+    // will default to setting it from candles if it's not.
+    const lowRateSet = spot && spot.low24 !== undefined && spot.low24 !== 0
+    const highRateSet = spot && spot.high24 !== undefined && spot.high24 !== 0
+    if (lowRateSet && highRateSet) {
```
buck54321, line 605:
```Typescript
if (spot && spot.low24 && spot.high24)
```

client/webserver/site/src/js/markets.ts
```diff
@@ -598,24 +596,32 @@ export default class MarketsPage extends BasePage {
...
+      high = spot.high24
+      low = spot.low24
+    } else {
+      const cache = this.market?.candleCaches[fiveMinBinKey]
+      if (!cache) {
+        for (const s of this.stats) {
+          s.tmpl.high.textContent = '-'
+          s.tmpl.low.textContent = '-'
```
buck54321, line 613: This will always be the case for old servers on first load, since we're defaulting to one-hour bins now. Maybe something like below ?

```Typescript
if (!cache) {
  if (this.candleDur !== fiveMinBinKey) {
    this.candleDur = fiveMinBinKey
    this.requestCandles()
    return
  }
  for (const s of this.stats) {
    s.tmpl.high.textContent = '-'
    s.tmpl.low.textContent = '-'
  }
  return
}
```
Effectively forcing old servers to default to five-minute bins and ignoring the last user's choice.

client/webserver/site/src/js/markets.ts
```diff
@@ -802,6 +808,22 @@ export default class MarketsPage extends BasePage {
...
+    // Check if we already have the fiveMinBin candles cache, if not, request
+    // for it first since we use it to determine the 24hour high/low in
+    // this.setHighLow() as a fallback if there is no data for 24hour high/low
+    // in spot.
+    const lastCandleDur = State.fetchLocal(State.lastCandleDurationLK)
+    const cache = this.market?.candleCaches[fiveMinBinKey]
+    if (!cache && lastCandleDur !== fiveMinBinKey) {
+      this.candleDur = fiveMinBinKey
+      this.requestCandles()
+    }
```
buck54321, line 821: This seems like the wrong place to do this. `setCandleDurBttns` is called in `setMarket` before `candleDurationSelected -> loadCandles` below, so `candleCaches` will always be empty.

client/webserver/site/src/js/markets.ts
```diff
@@ -802,6 +808,22 @@ export default class MarketsPage extends BasePage {
...
+
+    // The fiveMinBin candles have been requested, so we can set the user's last
+    // known configuration now.
+    this.candleDur = lastCandleDur
```
buck54321, line 825: Should handle initialization in the constructor.

buck54321 (changes requested):

client/webserver/site/src/js/markets.ts
```diff
@@ -598,22 +599,34 @@ export default class MarketsPage extends BasePage {
...
-    for (let i = cache.candles.length - 1; i >= 0; i--) {
-      const c = cache.candles[i]
-      if (low === 0 || (c.lowRate > 0 && c.lowRate < low)) low = c.lowRate
-      if (c.highRate > high) high = c.highRate
+      // Set high and low rates from candles.
+      for (let i = cache.candles.length - 1; i >= 0; i--) {
+        const c = cache.candles[i]
+        if (low === 0 || (c.lowRate > 0 && c.lowRate < low)) low = c.lowRate
+        if (c.highRate > high) high = c.highRate
+      }
```
buck54321, line 629: Not from this work, but there seems to be something missing here. Don't we need something like

```Typescript
const aDayAgo = new Date().getTime() - 86400000
for (let i = cache.candles.length - 1; i >= 0; i--) {
  const c = cache.candles[i]
  if (c.endStamp < aDayAgo) break
  ...
}
```

client/webserver/site/src/js/markets.ts
```diff
@@ -875,6 +888,7 @@ export default class MarketsPage extends BasePage {
     // candlesticks
     this.candleDur = fiveMinBinKey
     this.loadCandles()
+    this.candleDurationSelected(oneHrBinKey)
```
buck54321, line 891: This sort of defeats the purpose of remembering the user's selection if you're going to force it to one-hour bins every time they load a new market. We typically would do something like
```Typescript
 this.candleDur = State.fetchLocal(State.lastCandleDurationLK) || oneHrBinKey
```
in the constructor.



Also, you're calling `loadCandles` and `candleDurationSelected` also calls `loadCandles`

client/webserver/site/src/js/markets.ts
ukane-philemon: Good catch. I think we do.

client/webserver/site/src/js/markets.ts
ukane-philemon: > This sort of defeats the purpose of remembering the user's selection if you're going to force it to one-hour bins every time they load a new market

Uhmm, yeah. I thought of this but maybe I didn't fully understand why we were supposed to set the 1h bin as default.

chappjc: Yeah, the snipped I posted was not considering the local storage bit.

But the idea was a simple way to make the default be 1hr, but still always fetch and cache the 5min candles for the high/low reasons you mentioned.  The simplest way to do that is just to do loadCandles twice in a row.  I'm open to any solution, but we just need the 1hr candles as the default in 0.6.1 one way or another.

chappjc: > Uhmm, yeah. I thought of this but maybe I didn't fully understand why we were supposed to set the 1h bin as default.

Default means when they haven't chosen a bin already.

The reason 5m is no longer good for the default is because the market looks flat with practical trading volumes.

buck54321: The `if (this.candleDur !== fiveMinBinKey) {` snippet that's been added in `setHighLow` should take care of extra loading for older servers, but will not incur an extraneous request for new servers.

chappjc: I think I'm miscommunicating "default".  There's no need to store or persist it because it's just in the code.  If the user has never selected a bin size themselves, then we use the default bin size of 1h.

ukane-philemon: >If the user has never selected a bin size themselves, then we use the default bin size of 1h.

Yeah. I believe my recent commits resolve our concerns correctly and https://github.com/decred/dcrdex/pull/2284#discussion_r1171530003.

client/webserver/site/src/js/markets.ts
```diff
@@ -598,22 +599,36 @@ export default class MarketsPage extends BasePage {
...
+    // Use spot values for 24 hours high and low rates if it is available. We
+    // will default to setting it from candles if it's not.
+    if (spot && spot.low24 && spot.high24) {
+      high = spot.high24
+      low = spot.low24
+    } else {
+      const cache = this.market?.candleCaches[fiveMinBinKey]
+      if (!cache) {
+        if (this.candleDur !== fiveMinBinKey) {
+          this.candleDur = fiveMinBinKey
```
buck54321, line 613: I realized that doing this without doing `loadCandles` will cause some UI problems. We don't want to change this anyway. We want to show the user what they last selected. The changes [here](https://github.com/buck54321/dcrdex/compare/1fc4e44...buck54321:dcrdex:candles-dur-arg) allow the five-min candles to be loaded in the background and the UI to still function properly with the user's selection.

client/webserver/site/src/css/market.scss
```diff
@@ -417,7 +417,8 @@ div[data-handler=markets] {
 
       &:hover,
       &.selected {
-        border-color: #6cac78;
+        border-color: #dd7900;
+        color: #dd7900;
```
chappjc, line 421: I feel like we need to apply the selected attribute on page load too.  When I select a bin size, it gets these colors, but when I refresh the page, each bin size button looks the same.

client/webserver/site/src/js/markets.ts
```diff
@@ -872,10 +889,6 @@ export default class MarketsPage extends BasePage {
     // depth chart
     ws.request('loadmarket', makeMarket(host, base, quote))
 
-    // candlesticks
-    this.candleDur = fiveMinBinKey
-    this.loadCandles()
```
ukane-philemon, line 877: Moved this to `setCandleDurBttns` since we were always undoing our work there so I think it's better it stays there. Resolves https://github.com/decred/dcrdex/pull/2284/files#r1174081327

client/webserver/site/src/js/markets.ts
```diff
@@ -802,6 +816,9 @@ export default class MarketsPage extends BasePage {
       Doc.bind(bttn, 'click', () => this.candleDurationSelected(dur))
       page.durBttnBox.appendChild(bttn)
     }
+
+    // load candlesticks here since we are resetting page.durBttnBox above.
+    this.loadCandles()
```
chappjc, line 821: Oh yeah, `page.durBttnBox` needs to be populated in order for `loadCandles` to apply the `selected` class.  Seems fine but have not re-tested.

chappjc (approved):

client/webserver/site/src/js/markets.ts
buck54321, line 821: I feel like `loadMarket` should still be called in `setMarket`, just moved to right after `this.setCandleDurBttns()`. Not a big deal though.

buck54321 (approved):

PR #2285, opened 2023-04-03
client/asset/eth: avoid confirmations overflow

norwnd:
I think I observed this overflow recently on match card, consider additional protection when `uints` are used.


chappjc (approved):

chappjc:
Potentially security issue if swapping too soon.  Tested.  Merging and including with 0.6

PR #2286, opened 2023-04-03
site: Check for uninitialized market in certain note handlers

ukane-philemon:
This diff uses the JS optional chaining feature to prevent errors caused by attempting to access a null field. If a user losses internet connection and regains it, some js errors will be logged when notes try to access the `dex` or `market` field on `MarketPage` before those fields are initialized. But the changes here are not limited to only note handlers since it is cleaner and less code to use optional chaining to prevent runtime errors.
Screenshot 2023-04-03 at 3 20 20 PM

### DIFF
--- a/client/webserver/site/src/js/markets.ts
+++ b/client/webserver/site/src/js/markets.ts
@@ -690,7 +690,7 @@ export default class MarketsPage extends BasePage {
 
     if (!this.assetsAreSupported().isSupported) return // assets not supported
 
-    if (this.market.dex.tier < 1) return // acct suspended or not registered
+    if (!this.market || this.market.dex.tier < 1) return // acct suspended or not registered
 
     const { base, quote } = this.market
     const hasWallets = base && app().assets[base.id].wallet && quote && app().assets[quote.id].wallet
@@ -909,8 +909,8 @@ export default class MarketsPage extends BasePage {
   }
 
   /*
-   * reportDepthMouse accepts informations about the mouse position on the
-   * chart area.
+   * reportDepthMouse accepts information about the mouse position on the chart
+   * area.
    */
   reportDepthMouse (r: MouseReport) {
     while (this.hovers.length) (this.hovers.shift() as HTMLElement).classList.remove('hover')
@@ -937,9 +937,9 @@ export default class MarketsPage extends BasePage {
   }
 
   /*
-   * reportDepthZoom accepts informations about the current depth chart zoom level.
-   * This information is saved to disk so that the zoom level can be maintained
-   * across reloads.
+   * reportDepthZoom accepts information about the current depth chart zoom
+   * level. This information is saved to disk so that the zoom level can be
+   * maintained across reloads.
    */
   reportDepthZoom (zoom: number) {
     State.storeLocal(State.depthZoomLK, zoom)
@@ -2023,6 +2023,7 @@ export default class MarketsPage extends BasePage {
    * handlePriceUpdate is the handler for the 'spots' notification.
    */
   handlePriceUpdate (note: SpotPriceNote) {
+    if (!this.market) return // This note can arrive before the market is set.
     if (note.host === this.market.dex.host && note.spots[this.market.cfg.name]) {
       this.setCurrMarketPrice()
     }
@@ -2035,6 +2036,7 @@ export default class MarketsPage extends BasePage {
    */
   async handleBondUpdate (note: BondNote) {
     const dexAddr = note.dex
+    if (!this.market) return // This note can arrive before the market is set.
     if (dexAddr !== this.market.dex.host) return
     // If we just finished legacy registration, we need to update the Exchange.
     // TODO: Use tier change notification once available.
@@ -2085,6 +2087,7 @@ export default class MarketsPage extends BasePage {
    */
   handleEpochNote (note: EpochNote) {
     app().log('book', 'handleEpochNote:', note)
+    if (!this.market) return // This note can arrive before the market is set.
     if (note.host !== this.market.dex.host || note.marketID !== this.market.sid) return
     if (this.book) {
       this.book.setEpoch(note.epoch)
--- a/client/webserver/site/src/html/bodybuilder.tmpl
+++ b/client/webserver/site/src/html/bodybuilder.tmpl
@@ -104,7 +104,7 @@
 {{end}}
 
 {{define "bottom"}}
-<script src="/js/entry.js?v=LcglrVTq"></script>
+<script src="/js/entry.js?v=WECSAQK"></script>
 </body>
 </html>
 {{end}}


client/webserver/site/src/js/markets.ts
```diff
@@ -975,13 +975,13 @@ export default class MarketsPage extends BasePage {
...
     }
     return {
-      host: market.dex.host,
+      host: market?.dex?.host,
       isLimit: limit,
       sell: sell,
-      base: market.base.id,
-      quote: market.quote.id,
+      base: market?.base.id,
+      quote: market?.quote.id,
```
chappjc, line 982: The result is now instead of a javascript error, we return an incorrect result to the caller of `parseOrder`.

client/webserver/site/src/js/markets.ts
```diff
@@ -1097,7 +1097,7 @@ export default class MarketsPage extends BasePage {
       this.maxEstimateTimer = null
       if (counter !== this.maxOrderUpdateCounter) return
       const res = await postJSON(path, {
-        host: this.market.dex.host,
+        host: this.market?.dex?.host,
```
chappjc, line 1100: Now we will make the request but with no valid `host`.

chappjc (commented):
Seems that falling back to return `undefined` instead of hitting an error is not always better.  We'll create incorrect objects, make invalid requests, etc.

client/webserver/site/src/js/markets.ts
ukane-philemon: I think the backend would return an error for an invalid host/data, no?

ukane-philemon: > Maybe but please consider that using the optional chaining parameter is in general not always a good idea. Instead of console errors we now have incorrect behavior.

Noted. Most of the places it is used in this PR are where we do equality comparisons and I think we won't risk incorrect behaviour there.

client/webserver/site/src/js/markets.ts
```diff
@@ -226,7 +226,7 @@ export default class MarketsPage extends BasePage {
     // TODO: Use dexsettings page?
     const registerBttn = Doc.tmplElement(page.notRegistered, 'registerBttn')
     bind(registerBttn, 'click', () => {
-      app().loadPage('register', { host: this.market.dex.host })
+      app().loadPage('register', { host: this.market?.dex?.host })
```
chappjc, line 229: Here as well we're now punting an invalid value to another piece of code that will have to deal with undefined.

client/webserver/site/src/js/markets.ts
```diff
@@ -616,7 +616,7 @@ export default class MarketsPage extends BasePage {
       if (c.highRate > high) high = c.highRate
     }
 
-    const qconv = app().unitInfo(this.market.cfg.quoteid, this.market.dex).conventional.conversionFactor
+    const qconv = app().unitInfo(this.market?.cfg.quoteid, this.market?.dex).conventional.conversionFactor
```
chappjc, line 619: What will `unitInfo` do with undefined?

client/webserver/site/src/js/markets.ts
ukane-philemon: @chappjc, I believe your concerns have been addressed in https://github.com/decred/dcrdex/commit/d0c56ae4e8fd1a3d143a0d73c9206499a0c4d261, right?

client/webserver/site/src/js/markets.ts
```diff
@@ -226,7 +226,7 @@ export default class MarketsPage extends BasePage {
     // TODO: Use dexsettings page?
     const registerBttn = Doc.tmplElement(page.notRegistered, 'registerBttn')
     bind(registerBttn, 'click', () => {
-      app().loadPage('register', { host: this.market.dex.host })
+      if (this.market?.dex) app().loadPage('register', { host: this.market.dex.host })
```
buck54321, line 229: Unnecessary. The only time `page.notRegistered` is shown is in `setRegistrationStatusVisibility` which already checks for `this.market` and `this.market.dex`.

client/webserver/site/src/js/markets.ts
```diff
@@ -598,7 +598,8 @@ export default class MarketsPage extends BasePage {
 
   /* setHighLow calculates the high and low rates over the last 24 hours. */
   setHighLow () {
-    const cache = this.market?.candleCaches[fiveMinBinKey]
+    if (!this.market) return
+    const cache = this.market.candleCaches[fiveMinBinKey]
```
buck54321, line 602: Unnecessary. The only place `setHighLow` is called is in `handleCandlesRoute`, where we already check for `this.market`.

client/webserver/site/src/js/markets.ts
```diff
@@ -690,7 +691,7 @@ export default class MarketsPage extends BasePage {
 
     if (!this.assetsAreSupported().isSupported) return // assets not supported
 
-    if (this.market.dex.tier < 1) return // acct suspended or not registered
+    if (!this.market?.dex || this.market.dex.tier < 1) return // acct suspended or not registered
```
buck54321, line 694: This also seems like an impossible condition to hit based on how `resolveOrderFormVisibility` is used.

client/webserver/site/src/js/markets.ts
```diff
@@ -795,6 +796,7 @@ export default class MarketsPage extends BasePage {
 
   setCandleDurBttns () {
     const { page, market } = this
+    if (!market?.dex) return
```
buck54321, line 799: The only place `setCandleDurBttns` is used is in `setMarket` right after we've populated `this.market`. This is an impossible condition.

client/webserver/site/src/js/markets.ts
```diff
@@ -1084,7 +1086,7 @@ export default class MarketsPage extends BasePage {
     if (!this.maxLoaded) this.maxLoaded = app().loading(page.maxOrd)
     const [bid, qid] = [this.market.base.id, this.market.quote.id]
     const [bWallet, qWallet] = [app().assets[bid].wallet, app().assets[qid].wallet]
-    if (!bWallet || !bWallet.running || !qWallet || !qWallet.running) return
+    if (!bWallet || !bWallet.running || !qWallet || !qWallet.running || !this.market.dex) return
```
buck54321, line 1089: Is it possible to have `this.market` but not `this.market.dex`? I don't think it is.

client/webserver/site/src/js/markets.ts
```diff
@@ -1231,6 +1233,7 @@ export default class MarketsPage extends BasePage {
    */
   setMarketBuyOrderEstimate () {
     const market = this.market
+    if (!market?.dex) return
```
buck54321, line 1236: Not possible? We wouldn't show an order form if `this.market` wasn't ready.

buck54321 (changes requested):
It's not exactly clear to me under what conditions we might have no `this.market`, but be receiving spot and order updates. You mention connection issues. Can you explain what sequence would land us on the markets view, with spot and order updates coming in, but with no connected exchanges? Is it after a login with no available connections?
Regardless, most of these changes are unnecessary, because they require `this.market` to be populated in order for the requisite page elements to be shown that would lead to the error. Please audit a little deeper to see which are useful.

client/webserver/site/src/js/markets.ts
chappjc: > Is it possible to have `this.market` but not `this.market.dex`? I don't think it is.

I had the _exact_ same question.  This is really foggy.

ukane-philemon:
> Please audit a little deeper to see which are useful.

Will do. I agree some use of optional chaining in some methods in this might be redundant since callers might have already checked for such cases.

client/webserver/site/src/js/markets.ts
ukane-philemon: Not really, `createWallet` also calls this method and it could be that there is no dex connection by the time we use `resolveOrderFormVisibility`.

ukane-philemon:
> Can you explain what sequence would land us on the markets view, with spot and order updates coming in, but with no connected exchanges?

While on the markets page, no connection and the page has been refreshed. Turning the internet connection on could result in some notes being sent to the markets page before it is reloaded(i.e after connection to the server has been successful).

It occurs randomly since the markets page might be refreshed before the notes start hitting the FE. I will limit this safety check to only notes handlers.

client/webserver/site/src/js/markets.ts
```diff
@@ -1398,6 +1398,7 @@ export default class MarketsPage extends BasePage {
     app().log('book', 'handleBookRoute:', note)
     const mktBook = note.payload
     const market = this.market
+    if (!market?.dex) return
```
buck54321, line 1401: This is part of an order book subscription that is initiated with `'loadmarket'`, which isn't called until after we've set `this.market`. I'm pretty sure will have `this.market` here. Same with `handleBookOrderRoute`, `handleUnbookOrderRoute`, `handleUpdateRemainingRoute`, `handleEpochOrderRoute`, `handleCandlesRoute`, and `handleCandleUpdateRoute`. 

As far as I can tell, the only places we could potentially get an update before `this.market` is set is `handleEpochNote`, `handlePriceUpdate`, and `handleBondUpdate`.

buck54321 (changes requested):

client/webserver/site/src/js/markets.ts
buck54321: @ukane-philemon Unless you see a way that I don't, we can assume that if `this.market` is not `undefined`, then `this.market.dex` is also populated. That means you can remove the chaining operator after `dex` throughout.

ukane-philemon: I think I missed this, will remove.

client/webserver/site/src/js/markets.ts
ukane-philemon: Okay, I didn't trace all note handlers that are initiated **after** `this.market` is ready. Thanks.

ukane-philemon:
In my previous efforts, I exercised excessive caution by proposing the adoption of JavaScript's optional chaining feature. However, following a recent discourse with @buck54321, and after considering our utilization of specific note handlers, I have come to the realization that such precaution is unnecessary, thereby rendering the implementation of js optional chaining superfluous.

chappjc (approved):

PR #2287, opened 2023-04-03
[0.6] client/webserver: keep IPv6 out of csp header

chappjc:
Minimal fix for release-v0.6.

Pertaining to https://github.com/decred/dcrdex/pull/2283, this just filters out IPv6 addresses from injection into the csp part of the http response header.

Even though all IPv4 except 127.0.0.1 _does not match_ (we must rely on `'self'` working correctly for these), they do not make the CSP entry invalid, so we are continuing to insert those so we do not create a change of behavior.

On master we may actually remove this `content-src` injection since the browser bug was fixed about a year ago.  On master, we should also update the backend log line that prints out the URL so that it doesn't include the wildcard address since that's no good for a browser address.

All this PR does is to avoid adding IPv6 addresses, since those actually make the CSP **invalid**. Examples:
- "[::1]:5759"
- "[::]:5759"
- "[2607:f8b0:4000:81a::200e]:5759"
- "0.0.0.0:5759" -- this is the culprit in the Docker issue because on a machine with an IPv6 stack, the socket resolution generally converts the wildcard in to the equivalent IPv6 wildcard of [::]

### DIFF
--- a/client/webserver/webserver.go
+++ b/client/webserver/webserver.go
@@ -516,7 +516,13 @@ func (s *WebServer) Connect(ctx context.Context) (*sync.WaitGroup, error) {
 	// Work around a webkit (safari) bug with the handling of the connect-src
 	// directive of content security policy. See:
 	// https://bugs.webkit.org/show_bug.cgi?id=201591
-	s.csp = fmt.Sprintf("%s ws://%s", baseCSP, s.addr)
+	if tcpAddr, ok := listener.Addr().(*net.TCPAddr); !ok || tcpAddr.IP.To4() != nil {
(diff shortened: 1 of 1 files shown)


ukane-philemon (approved):
This works, at least for newer versions of the Safari browser.

ukane-philemon:
> On master we may actually remove this content-src injection since the browser bug was fixed about a year ago.

We could add a notice somewhere, maybe in the release notes that this workaround would be removed in subsequent dcrdex releases.

chappjc:
> We could add a notice somewhere, maybe in the release notes that this workaround would be removed in subsequent dcrdex releases.

Will do.

Issue #2241, opened 2023-03-22
ui: display error message when unable to connect wallet


norwnd:
Observed with ETH wallet but could affect others too, 
- I had single Ethereum Provider configured 
- upon `dexc` start it wasn't able to connect (I mocked up this failure to test something else), hence wallet is shown in "Off" state which is all good so far
- but then clicking on **connect** button results into something weird, some frames change but no errors are displayed:

[connect-wallet-no-error-shown.webm](https://user-images.githubusercontent.com/112318969/226813312-52d6cd02-4598-491d-97c2-0379f69bc2df.webm)

<br>

while in the logs the error is:
```
2023-03-22 05:41:08.044 [ERR] WEB: error connecting to eth wallet: connectWallet: failed to connect eth wallet: connect failure: failed to connect
```
so the user should see something similar ^ in UI.

As a sidenote, I'm not even sure why this `connect` button is there in the first place , wouldn't it be **better to try re-connect wallet automatically every 30s or something** ?

norwnd:
Somewhat related issue: clicking on `connect` button doesn't change anything until I refresh page even though according to logs wallet does get reconnected (so from user perspective, he'll continue clicking on `connect` button when what he actually is to refresh browser page).

norwnd:
> Somewhat related issue: clicking on connect button doesn't change anything until I refresh page even though according to logs wallet does get reconnected (so from user perspective, he'll continue clicking on connect button when what he actually is to refresh browser page).

This behavior is likely the result of me running 2 different **dexc** instances, accessing both from same browser, actually.

ukane-philemon:
@chappjc, is this still an issue?

chappjc:
I think you solved this one.

Issue #1712, opened 2022-07-15
password error shouldn't go to notification bell


buck54321:
![image](https://user-images.githubusercontent.com/6109680/179294113-d908a88a-1c6c-471f-9f99-a505b754d449.png)

vctt94:
I can work on that

buck54321:
It's yours. I think I saw that particular notification when I was changing wallet types, but you should look around. This is probably part of a bigger issue with forms. Anywhere an error is returned on a form submission, it should be displayed on the form only. Anywhere you see
```Javascript
app().checkResponse(res)
```
instead of 
```Javascript
app().checkResponse(res, true)
```
is probably wrong. I'm tempted to just get rid of the notification forwarding behavior of `Application.checkResponse` altogether, but I haven't looked closely at the implications of that.

vctt94:
So I was looking into the code, and right now all check response which fails (and !skipnote), goes to the notification belt. From comments: 

```
  /*
   * checkResponse checks the response object as returned from the functions in
   * the http module. If the response indicates that the request failed, a
   * message will be displayed in the drop-down notifications and false will be
   * returned.
   */
```

If removing the forwarding behavior of checkresponse, should any of those notifications still be kept in the belt and moved to notification.go?

```

buck54321:
> If removing the forwarding behavior of checkresponse, should any of those notifications still be kept in the belt and moved to notification.go?

You'd need to check it on a case-by-case basis, but my gut says that no API errors should go to the bell. Hopefully you can just remove the error forwarding of `checkResponse`. That'd be good.

PR #2288, opened 2023-04-04
site: Notify users when connecting wallets fail

ukane-philemon:
Closes # 2241

UI changes:
Screenshot 2023-04-04 at 11 30 19 PM

> @norwnd (#2241): As a sidenote, I'm not even sure why this connect button is there in the first place , wouldn't it be better to try re-connect wallet automatically every 30s or something?

This part is not handled in this PR. @chappjc, what do you think?

IMO, using a reconnect loop for wallet connection when are not sure that we will ever connect seems to be undesirable. Wallet connection will be attempted in any part of the client code that requires a connected wallet, plus, like the note above, If a user is missing wallet files, we are going to try forever until they take action.

### DIFF
--- a/client/webserver/locales/en-us.go
+++ b/client/webserver/locales/en-us.go
@@ -335,4 +335,5 @@ var EnUS = map[string]string{
 	"connected":              "Connected",
 	"Remove":                 "Remove",
 	"unready_wallets_msg":    "Your wallets must be connected and unlocked before trades can be processed. Resolve this ASAP!",
+	"Error":                  "Error",
 }
--- a/client/webserver/site/src/html/wallets.tmpl
+++ b/client/webserver/site/src/html/wallets.tmpl
@@ -535,6 +535,13 @@
       </div>
     </form>
 
+    {{- /* Error Modal */ -}}
+    <form id="errorModal" class="d-hide mw-425">
+      <div class="form-closer hoverbg"><span class="ico-cross"></span></div>
+      <div class="fs22 text-center">[[[Error]]]</div>
+      <div id="errorModalMsg" class="fs16 errcolor my-2 text-center"></div>
+    </form>
+
   </div>
 </div>
 {{template "bottom"}}
--- a/client/webserver/site/src/js/locales.ts
+++ b/client/webserver/site/src/js/locales.ts
@@ -108,6 +108,7 @@ export const ID_EMPTY_DEX_ADDRESS_MSG = 'EMPTY_DEX_ADDRESS_MSG'
 export const ID_SELECT_WALLET_FOR_FEE_PAYMENT = 'SELECT_WALLET_FOR_FEE_PAYMENT'
 export const ID_UNAVAILABLE = 'UNAVAILABLE'
 export const ID_WALLET_SYNC_FINISHING_UP = 'WALLET_SYNC_FINISHING_UP'
+export const ID_CONNECT_WALLET_ERR_MSG = 'CONNECTING_WALLET_ERR_MSG'
 
 export const enUS: Locale = {
   [ID_NO_PASS_ERROR_MSG]: 'password cannot be empty',
@@ -217,7 +218,8 @@ export const enUS: Locale = {
   [ID_UNAVAILABLE]: 'unavailable',
   [ID_EMPTY_DEX_ADDRESS_MSG]: 'DEX address cannot be empty',
   [ID_SELECT_WALLET_FOR_FEE_PAYMENT]: 'Select a valid wallet to post a bond',
-  [ID_WALLET_SYNC_FINISHING_UP]: 'finishing up'
+  [ID_WALLET_SYNC_FINISHING_UP]: 'finishing up',
+  [ID_CONNECT_WALLET_ERR_MSG]: 'Failed to connect {{ assetName }} wallet: {{ errMsg }}'
 }
 
 export const ptBR: Locale = {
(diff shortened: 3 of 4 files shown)


chappjc:
Yeah, I don't want to replace the connect button with a background reconnect loop.

Regarding any error from connect, I'm pretty sure we're trying to show errors on the page or form near the UI element interaction that generated the error, not the notification menu.  https://github.com/decred/dcrdex/issues/1712#issuecomment-1198568703

buck54321:
I'd prefer to show the message on the page rather than as a tooltip.

chappjc:
How about an error message "form" that opens up with the messgae, so we don't have to pre-allocate real estate for an possible error?

buck54321:
> How about an error message "form" that opens up with the messgae, so we don't have to pre-allocate real estate for a possible error?

Pre-allocating page space is the simpler option. The "form" you mention, as I understand, is an interactive (probably modal) dialog that would explain the error and possibly provide guidance or options for resolution. I do agree that such an experience is desirable. I'm not sure that we should try that in this PR.

chappjc:
Yeah, just something you can close when done reading.  But didn't mean to suggest we try to provide resolutions or anything smarter that just showing the error.
It's fine too drop the error in div or something on the page.  It can be hidden so it doesn't eat space unless needed. I just didn't want to sacrifice layout quality

ukane-philemon:
> I just didn't want to sacrifice layout quality

Yup, these errors can be ugly and I don't quite get where in the page @buck54321 wants this kind of error message since the table where the connect btn is located already has a lot.

buck54321:
> > I just didn't want to sacrifice layout quality
> 
> Yup, these errors can be ugly and I don't quite get where in the page @buck54321 wants this kind of error message since the table where the connect btn is located already has a lot.

Just to echo @chappjc above

> I'm pretty sure we're trying to show errors on the page or form near the UI element interaction that generated the error

I think it looks alright on the page.

https://github.com/buck54321/dcrdex/compare/d165f1f...buck54321:dcrdex:ukaine-connect-err-on-page

image

ukane-philemon:
> I think it looks alright on the page.

For long errors like the eth error in my PR comment, it will easily look very ugly IMO. But if you want to, we can dump everything below the button.

chappjc:
> If using a tooltip is not a good resolution to this, we can use a general error modal that can be closed by the user when they are done reading and digesting the error, showing the whole error on the page beside the connect btn does not look desirable IMO.

A tooltip is not, I think we agree.  Sure, give the modal a try quick.  We just don't want a major production over this.

ukane-philemon:
https://github.com/decred/dcrdex/pull/2288/commits/757c9bc16f63c10c9dc69784d3804b2edc7f1327 UI changes.

Screenshot 2023-04-06 at 10 58 57 PM
Screenshot 2023-04-06 at 10 57 43 PM

chappjc:
Seems alright I guess.  The work "Message" in "Error Message" is not needed though.  It's more like "Connect Error".  I also figured we'd have an "x" at the top right like all the other forms instead of an "Ok" button.  BTW, it's "OK" not "Ok"

chappjc:
Anyway, you and @buck54321 work out where to put the error, but IMO it gets messy putting it right next to the button.  I think the dialog form that opens is a good compromise because you can't ignore it like tooltips and notifications, and it doesn't touch the page layout.

ukane-philemon:
> I also figured we'd have an "x" at the top right like all the other forms instead of an "Ok" button.

I think it's better they acknowledge that they have seen the error, the x at the top is usually for form dismissal but we can have that instead of the btn if that's better or keep both?

> It's more like "Connect Error"

I wanted to be more flexible instead of tying the modal to only `Connect Error`.

chappjc:
>  I think it's better they acknowledge that they have seen the error, the x at the top is usually for form dismissal but we can have that instead of the btn if that's better or keep both?

Whatever really.  The button just looks funny.  It's chunky and big, and requires a translation (!).

> I wanted to be more flexible instead of tying the modal to only `Connect Error`.

Of course, but the "Connect" part can be a variable inserted by the JS caller.  If that's not trivial for whaterver reason, just "Error" is preferable to "Error Message", which looks silly.

ukane-philemon:
Update the images above.

buck54321 (changes requested):
Please consider https://github.com/buck54321/dcrdex/compare/1e50d26...buck54321:dcrdex:error-modal-in-forms

ukane-philemon:
> Please consider [buck54321/dcrdex@1e50d26...buck54321:dcrdex:error-modal-in-forms](https://github.com/buck54321/dcrdex/compare/1e50d26...buck54321:dcrdex:error-modal-in-forms)

I could have used the existing form structure we have(less code) but the modal is not a "form". Still, we can run with your suggestion, I understand you prefer less code in this case.

buck54321 (approved):

buck54321:
> I could use the existing form structure we have(less code) but the modal is not a "form". Still, we can run with your suggestion, I understand you prefer less code in this case.

It's not about less code. The existing forms **are** modal, since they prevent page interaction until the form is submitted or dismissed. No reason to reinvent the wheel.

chappjc:
Also, this isn't the first form used this way.  The "Export Wallet" form is also purely informational:

eth-export-key
(don't bother with the key, it's simnet)

ukane-philemon:
Noted.

PR #2293, opened 2023-04-10
multi: allow no dcr reg fee config on server, remove dcr-specific fields

chappjc:
The first commit removes the server's startup requirement of having DCR reg fee info specified.  This was a hold-over from before we even had multi-asset reg fee support.  There is still the RegFees map (which may also be empty in a bonds-only config).

The second commit removes the warning about a server that does not support DCR for registration. This also removes the handling of ancient servers that even lacked the RegFees map (only had the dcr-specific fields).

### DIFF
--- a/client/core/bookie.go
+++ b/client/core/bookie.go
@@ -883,32 +883,6 @@ func (dc *dexConnection) refreshServerConfig() (*msgjson.ConfigResult, error) {
 		}
 	}
 
-	// Patch ConfigResponse.RegFees if no entry for DCR is there, meaning it is
-	// likely an older server using cfg.Fee and cfg.RegFeeConfirms.
-	if dcrAsset := cfg.RegFees["dcr"]; dcrAsset == nil {
-		dc.log.Warnf("Legacy server %v does not provide a regFees map.", dc.acct.host)
-		if cfg.RegFees == nil {
-			cfg.RegFees = make(map[string]*msgjson.FeeAsset)
-		}
-		if cfg.Fee > 0 {
-			cfg.RegFees["dcr"] = &msgjson.FeeAsset{ // v0 is only DCR
-				ID:    42,
-				Confs: uint32(cfg.RegFeeConfirms),
-				Amt:   cfg.Fee,
-			}
-		} else {
-			dc.log.Warnf("Server %v does not support DCR for registration", dc.acct.host)
-		}
-	} else {
-		if cfg.Fee > 0 && dcrAsset.Amt != cfg.Fee {
-			dc.log.Warnf("Inconsistent DCR fee amount: %d != %d", dcrAsset.Amt, cfg.Fee)
-		}
-		if dcrAsset.Confs != uint32(cfg.RegFeeConfirms) {
-			dc.log.Warnf("Inconsistent DCR fee confirmation requirement: %d != %d",
-				dcrAsset.Confs, cfg.RegFeeConfirms)
-		}
-	}
-
 	// Update the dex connection with the new config details, including
 	// StartEpoch and FinalEpoch, and rebuild the market data maps.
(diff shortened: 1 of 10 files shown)


dex/msgjson/types.go
```diff
@@ -1295,9 +1295,7 @@ type ConfigResult struct {
 	// by bond version.
 	BondExpiry uint64 `json:"DEV_bondExpiry"`
 
-	RegFees        map[string]*FeeAsset `json:"regFees"`
-	Fee            uint64               `json:"fee"`            // DEPRECATED
-	RegFeeConfirms uint16               `json:"regfeeconfirms"` // DEPRECATED
```
chappjc, line 1300: Legacy clients will see a `Fee` of 0, and block registration (See `(*Core).Register`)

client/core/bookie.go
```diff
@@ -883,32 +883,6 @@ func (dc *dexConnection) refreshServerConfig() (*msgjson.ConfigResult, error) {
...
-			cfg.RegFees = make(map[string]*msgjson.FeeAsset)
-		}
-		if cfg.Fee > 0 {
-			cfg.RegFees["dcr"] = &msgjson.FeeAsset{ // v0 is only DCR
-				ID:    42,
-				Confs: uint32(cfg.RegFeeConfirms),
-				Amt:   cfg.Fee,
-			}
-		} else {
-			dc.log.Warnf("Server %v does not support DCR for registration", dc.acct.host)
```
chappjc, line 900: Released clients will warn about this, but it's no longer an issue.

client/core/bookie.go
```diff
@@ -883,32 +883,6 @@ func (dc *dexConnection) refreshServerConfig() (*msgjson.ConfigResult, error) {
...
 	}
 
-	// Patch ConfigResponse.RegFees if no entry for DCR is there, meaning it is
-	// likely an older server using cfg.Fee and cfg.RegFeeConfirms.
-	if dcrAsset := cfg.RegFees["dcr"]; dcrAsset == nil {
-		dc.log.Warnf("Legacy server %v does not provide a regFees map.", dc.acct.host)
-		if cfg.RegFees == nil {
-			cfg.RegFees = make(map[string]*msgjson.FeeAsset)
-		}
-		if cfg.Fee > 0 {
```
chappjc, line 893: Release clients will incorrectly refer to this as a "Legacy server" if the server lacks a DCR entry in the map, but the behavior is otherwise fine (block registering with DCR).

client/core/bookie.go
```diff
@@ -883,32 +883,6 @@ func (dc *dexConnection) refreshServerConfig() (*msgjson.ConfigResult, error) {
...
-			dc.log.Warnf("Server %v does not support DCR for registration", dc.acct.host)
-		}
-	} else {
-		if cfg.Fee > 0 && dcrAsset.Amt != cfg.Fee {
-			dc.log.Warnf("Inconsistent DCR fee amount: %d != %d", dcrAsset.Amt, cfg.Fee)
-		}
-		if dcrAsset.Confs != uint32(cfg.RegFeeConfirms) {
-			dc.log.Warnf("Inconsistent DCR fee confirmation requirement: %d != %d",
-				dcrAsset.Confs, cfg.RegFeeConfirms)
-		}
```
chappjc, line 909: Released clients will also warn about this because these dcr-specific fields will be zero with a server running the code changes in this PR, but the clients will still use any set DCR values in the `RegFees` map.

JoeGruffins (approved):

martonp (commented):
When running an older client, the UI still shows DCR as a registration asset, even though the
server does not support it:
Screenshot 2023-04-11 at 10 45 24 AM

martonp (approved):
They should just upgrade to 0.6 anyways.

PR #2299, opened 2023-04-14
dex/networks/ltc: witness limits fixes and decoder simplifications

chappjc:
ltcsuite is adopting our MW tx deserialization code, and it [revealed](https://github.com/ltcsuite/ltcd/issues/17#issuecomment-1507856230) a small usability issue with the `decoder` API, which this fixes.

This also updates the `maxWitnessItemsPerInput` and `maxWitnessItemSize` to match the current values following the btcsuite backports in https://github.com/ltcsuite/ltcd/pull/25

### DIFF
--- a/dex/networks/ltc/tx.go
+++ b/dex/networks/ltc/tx.go
@@ -26,18 +26,23 @@ const (
 
 type decoder struct {
 	buf [8]byte
-	io.Reader
-	tee *bytes.Buffer
+	rd  io.Reader
+	tee *bytes.Buffer // anything read from rd is Written to tee
 }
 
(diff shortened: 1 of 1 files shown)


buck54321 (approved):
> ltcsuite is adopting our MW tx deserialization code

Nice! This is testing well on loadbot, simnet trade tests, and `TestOnlineDeserializeBlock`, as well as manual self trading.

chappjc:
> Nice!

They're still just throwing a way the MW tx haha, but yup, very cool

> This is testing well on loadbot, simnet trade tests, and `TestOnlineDeserializeBlock`, as well as manual self trading.

Thanks for testing it so thoroughly.

PR #2307, opened 2023-04-15
multi: Send recent matches in orderbook sync

### LINKED ISSUE #2296: Recent Matches vanish
image

It was there briefly but now the list is empty.

martonp:
Server orderbook is updated to cache the 100 most recent matches in memory, and they are returned whenever a client syncs the orderbook.

Closes #2296

### DIFF
--- a/dex/msgjson/types.go
+++ b/dex/msgjson/types.go
@@ -869,6 +869,9 @@ type OrderBook struct {
 	Orders       []*BookOrderNote `json:"orders"`
 	BaseFeeRate  uint64           `json:"baseFeeRate"`
 	QuoteFeeRate uint64           `json:"quoteFeeRate"`
+	// RecentMatches is [rate, qty, timestamp]. Quantity is signed.
+	// Negative means that the maker was a sell order.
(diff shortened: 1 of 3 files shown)


dex/msgjson/types.go
```diff
@@ -869,6 +869,9 @@ type OrderBook struct {
 	Orders       []*BookOrderNote `json:"orders"`
 	BaseFeeRate  uint64           `json:"baseFeeRate"`
 	QuoteFeeRate uint64           `json:"quoteFeeRate"`
+	// RecentMatches is [rate, qty, timestamp]. Quantity is signed.
+	// Negative means that the maker was a sell order.
```
chappjc, line 873: Should this be **taker** was a sell order? Or does `Reset` have it switched?

chappjc (changes requested):
utACK.  But is the qty sign flipped?

dex/msgjson/types.go
chappjc: Tested and taker sells loaded this way show as red (correct), so looks like the comment is just a typo.

PR #2312, opened 2023-04-18
server/market: startup balance checks must convert base to quote

chappjc:
When an account-based asset is the quote asset in a market, and there is a booked buy order, the quantity to check must be in units of the quote asset (the account-based asset).

Test/repro: On the simnet harness fund your wallet with something like `0.39` ETH.  On the DCR-ETH market, book a **buy** order at 0.01 ETH/DCR for 1 lot, which should say "Total: 0.10 ETH".
Stop the server process and start again with `./run`.  It would unbook the order because it was comparing the base asset (10 DCR or 1000000000 atoms) with the eth balance of 0.039 ETH (390000000 gwei).

With this change, it will compare quote units (plus fees) against balance.

### DIFF
--- a/server/market/market.go
+++ b/server/market/market.go
@@ -358,18 +358,18 @@ ordersLoop:
 			var addr string
 			var qty, lots uint64
 			var redeems int
-			if lo.Sell {
+			if lo.Sell { // sell base => redeem acct-based quote
 				addr = lo.Address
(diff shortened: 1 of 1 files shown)


buck54321 (approved):

buck54321:
Was able to repro and confirm that this is resolves the issue.

PR #2313, opened 2023-04-19
site: correct order buy/sell colors

chappjc:
This corrects the `side-indicator`'s sell color.  It was correct on the light theme.

This also prevents both `buy` and `sell` classes on the `sideLight` div.  Both issues can be seen:

image

`.side-indicator.sell` used the buy color.  And the `sideLight` div had both `buy sell` classes.

The second issue addressed in this PR is the "active" indicator light not being shown for orders with settling matches.

Both issues visible here:

image


chappjc:
Test:

image

image

image

PR #2317, opened 2023-04-20
client/core: add core methods and webserver endpoints for shielded wallets

buck54321:
Just copied the `ShieldedWallet` interface from #2250.

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -10042,7 +10042,7 @@ func (c *Core) NewShieldedAddress(assetID uint32) (string, error) {
 }
 
 // ShieldFunds moves funds from the transparent account to the shielded account.
-func (c *Core) ShieldFunds(assetID uint32, amt uint64) (dex.Bytes, error) {
+func (c *Core) ShieldFunds(assetID uint32, amt uint64) ([]byte, error) {
 	sw, err := c.shieldedWallet(assetID)
 	if err != nil {
 		return nil, err
@@ -10052,20 +10052,49 @@ func (c *Core) ShieldFunds(assetID uint32, amt uint64) (dex.Bytes, error) {
 
 // UnshieldFunds moves funds from the shielded account to the transparent
 // account.
-func (c *Core) UnshieldFunds(assetID uint32, amt uint64) (dex.Bytes, error) {
+func (c *Core) UnshieldFunds(assetID uint32, amt uint64) ([]byte, error) {
 	sw, err := c.shieldedWallet(assetID)
 	if err != nil {
 		return nil, err
 	}
+
 	return sw.UnshieldFunds(c.ctx, amt)
 }
 
 // SendShielded sends funds from the shielded account to the provided shielded
 // or transparent address.
-func (c *Core) SendShielded(assetID uint32, toAddr string, amt uint64) (dex.Bytes, error) {
+func (c *Core) SendShielded(appPW []byte, assetID uint32, toAddr string, amt uint64) ([]byte, error) {
+	_, err := c.encryptionKey(appPW)
+	if err != nil {
+		return nil, fmt.Errorf("password error: %w", err)
+	}
+
 	sw, err := c.shieldedWallet(assetID)
 	if err != nil {
 		return nil, err
 	}
-	return sw.SendShielded(c.ctx, toAddr, amt)
+
+	coinID, err := sw.SendShielded(c.ctx, toAddr, amt)
+	if err != nil {
+		// Let errors go to logs and UI forms.
+		c.log.Errorf("Error sending shielded funds: %v", err)
+		return nil, err
+	}
+
+	// Send success notification.
+	ui, err := asset.UnitInfo(assetID)
+	if err != nil {
+		c.log.Errorf("Error getting unit info for shielded asset %d", assetID)
+	} else {
+		coinIDStr, err := asset.DecodeCoinID(assetID, coinID)
+		if err != nil {
+			c.log.Errorf("Error decoding coin ID %x from shielded send: %v", coinID, err)
+		} else {
+			val, unit := ui.ConventionalString(amt), ui.Conventional.Unit
+			subject, details := c.formatDetails(TopicShieldedSendSuccess, val, unit, toAddr, coinIDStr)
+			c.notify(newSendNote(TopicShieldedSendSuccess, subject, details, db.Success))
+		}
+	}
+
+	return coinID, nil
 }
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -10007,3 +10007,65 @@ func (c *Core) saveDisabledRateSources() {
 		c.log.Errorf("Unable to save disabled fiat rate source to database: %v", err)
 	}
 }
+
+func (c *Core) shieldedWallet(assetID uint32) (asset.ShieldedWallet, error) {
+	w, found := c.wallet(assetID)
+	if !found {
+		return nil, fmt.Errorf("no %s wallet", unbip(assetID))
+	}
+	sw, is := w.Wallet.(asset.ShieldedWallet)
+	if !is {
+		return nil, fmt.Errorf("%s wallet is not a shielded wallet", unbip(assetID))
+	}
+	return sw, nil
+}
+
+// ShieldedStatus is the shielded balance and last address associated with the
+// shielded wallet.
+func (c *Core) ShieldedStatus(assetID uint32) (*asset.ShieldedStatus, error) {
+	sw, err := c.shieldedWallet(assetID)
+	if err != nil {
+		return nil, err
+	}
+	return sw.ShieldedStatus()
+}
+
+// NewShieldedAddress creates a new shielded address. Visit the link below for
+// additional notes on shielded address reuse.
+// https://electriccoin.co/blog/shielded-address-contexts/
+func (c *Core) NewShieldedAddress(assetID uint32) (string, error) {
+	sw, err := c.shieldedWallet(assetID)
+	if err != nil {
+		return "", err
+	}
+	return sw.NewShieldedAddress()
+}
+
+// ShieldFunds moves funds from the transparent account to the shielded account.
+func (c *Core) ShieldFunds(assetID uint32, amt uint64) (dex.Bytes, error) {
+	sw, err := c.shieldedWallet(assetID)
+	if err != nil {
+		return nil, err
+	}
+	return sw.ShieldFunds(c.ctx, amt)
+}
+
+// UnshieldFunds moves funds from the shielded account to the transparent
+// account.
+func (c *Core) UnshieldFunds(assetID uint32, amt uint64) (dex.Bytes, error) {
+	sw, err := c.shieldedWallet(assetID)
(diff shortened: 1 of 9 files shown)


chappjc (approved):
Seems fine.  Clean interface.  Just the same comment about `dex.Bytes` as in the prerequisite zec wallet PR.  BTW, those outputs are eaten presently.  Would you send those tx IDs to the frontend in string format presumably?

client/core/core.go
```diff
@@ -10002,3 +10002,65 @@ func (c *Core) saveDisabledRateSources() {
...
+	sw, err := c.shieldedWallet(assetID)
+	if err != nil {
+		return "", err
+	}
+	return sw.NewShieldedAddress()
+}
+
+// ShieldFunds moves funds from the transparent account to the shielded account.
+func (c *Core) ShieldFunds(assetID uint32, amt uint64) (dex.Bytes, error) {
+	sw, err := c.shieldedWallet(assetID)
```
martonp, line 10041: These functions should take a password.

martonp (commented):
LGTM, I just think shield, unsheield, and send should take a password.

client/core/core.go
chappjc: Why?  webserver is authenticated

martonp: True, I just thought because `Send` also requires a password. `Send` uses it to unlock the wallet, I guess that could be done over here as well. Also, the RPCServer doesn't check the password anywhere.

buck54321: It probably makes sense for `SendShielded`, since that can send to any address. We require a password for `(*Core).Send`.

chappjc: The rpc server is authenticated like the webserver.  But unlocking the wallet is a good reason

chappjc: > The rpc server is authenticated like the webserver. But unlocking the wallet is a good reason

Actually, why is unlocking that wallet a good reason?  There are methods to unlock the wallet

buck54321: Zcash wallet isn't an `Authenticator` anyway, but wallets that are authenticated required unlocking for sends. But our main goal is to add an authentication step even if the user has a cached password. That's why we're not using `getCachedPasswordUsingRequest` in `apiSend`.

chappjc: Alright, makes sense.

But I've been a little confused about the Core API and why certain things require a password.  Take bitcoind for example.  You do `walletpassphrase` and then you can do `sendtoaddress`/`signrawtransaction`/etc. without entering the wallet password again.  The node's RPC server's are user/password protected.
Similarly, there is authentication on dexc's web and RPC servers.  Although in the case of webserver the endpoints requiring auth really pertain to successfully having logged in once and stored an auth cookie, while with the RPC server it just means you passed the correct RPC credentials.

Anyway, the reason why it confuses me a bit is because `Core` methods are a Go API, not an interface to the user (there's always another layer between the user and Core where permissioning requests happens).  It's not a big deal to me, I'm just not clear about it.

Put another way, it seems to me that a bunch of the `Core` methods that require a password could really just check internally an `isLoggedIn` flag.  Combined with the interface's auth (pass+cookie for webserver, or user/pass for rpcserver), that seems to have the right effect.

buck54321: We typically require a password when we need a `Crypter`, and we don't ever store a `Crypter` in memory, nor do we want to. I think what you're getting at is that for many methods, the `Crypter` is only needed to unlock wallets, so why not just have the caller unlock wallets before using those methods. That's fair. I think the reason is more because our UX goals are bleeding over into our back end design. We want the user to be able to perform all actions with a single form. If they want to trade on a locked wallet, we just add an app password field to the trade form, for instance, rather than showing them two different dialogs. Perhaps we should leave that UX intact, but have the front end send the `/openwallet` request separately. That's reasonable, I think. I don't know if it's worth an API-breaking refactor at this point, and the "password any time encryption might be needed" approach doesn't bother me so much.

chappjc: > the `Crypter` is only needed to unlock wallets, so why not just have the caller unlock wallets before using those methods. That's fair

Exactly.  And if it does not require an unlock, then we probably didn't need the password on the core method I think.  In general, Send can require an unlock, even if ZEC doesn't, so I get it.

>  I think the reason is more because our UX goals are bleeding over into our back end design

Yes, but not just backend.  It's transcending backend *layers* webserver -> Core.

> I don't know if it's worth an API-breaking refactor at this point, and the "password any time encryption might be needed" approach doesn't bother me so much.

No not really worth it.  I'm just trying to get to the heart of our policy regarding *Core*'s password requirements when there's no obvious need for a crypter.  Sometimes it seems like we're just double authorizing at two different backend layers.

Since some assets can require unlock to send, I think you've made a good case to keep it as it is, and have this shielding require a pass too, but there are other was to approach this as you've layed out.

buck54321:
> BTW, those outputs are eaten presently. Would you send those tx IDs to the frontend in string format presumably?

I've added a notification instead. I don't see a reason to send the coin ID presently. Notably, even though we send the coin ID from `apiSend` to the front-end, it is ignored. We rely on a notification to persist the info in the notification tray. 

There's definitely an argument for dropping those coin ID returns. To me, it seems like a fundamental piece of information when a tx is generated, and I don't want to presume what consumers will need or want.

JoeGruffins (approved):
These are probably existing. loadbot seems to be broken. Also on testnet if I send funds out the coins do not show as locked or gone for a long while. I am however unable to trade with those funds.

JoeGruffins:
Probably also unrelated but will see this error log on start up: `[ERR] WEB: wallet does not support options` but don't know which wallet.

JoeGruffins:
Trying to inspect a tx with the cli, I think this worked before but any idea what this wants?
```
$ zcash-cli getrawtransaction 09b2a0ad15b8d8e0337973bb630afc451987719af995e60147d3a0030a7f29e
error code: -8
error message:
parameter 1 must be hexadecimal string (not '09b2a0ad15b8d8e0337973bb630afc451987719af995e60147d3a0030a7f29e')
...
[joe@hyrule ~]$ zcash-cli help getrawtransaction
getrawtransaction "txid" ( verbose "blockhash" )
...
Arguments:
1. "txid"      (string, required) The transaction id


```

client/core/core.go
```diff
@@ -10052,20 +10052,49 @@ func (c *Core) ShieldFunds(assetID uint32, amt uint64) (dex.Bytes, error) {
...
 	}
+
 	return sw.UnshieldFunds(c.ctx, amt)
 }
 
 // SendShielded sends funds from the shielded account to the provided shielded
 // or transparent address.
-func (c *Core) SendShielded(assetID uint32, toAddr string, amt uint64) (dex.Bytes, error) {
+func (c *Core) SendShielded(appPW []byte, assetID uint32, toAddr string, amt uint64) ([]byte, error) {
+	_, err := c.encryptionKey(appPW)
```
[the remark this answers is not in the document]
chappjc: Ohhhh...  is the use case more like: you walk away from your computer with dex opened and logged in, someone walks up to it and tries to send funds, so a password prompt is there to prevent that?
If so maybe at some point we build a purely frontend mechanism for checking the password based on a hash of the correct pass used on login.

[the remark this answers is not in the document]
buck54321: > Ohhhh... is the use case more like: you walk away from your computer with dex opened and logged in, someone walks up to it and tries to send funds, so a password prompt is there to prevent that?

Yes.

> If so maybe at some point we build a purely frontend mechanism for checking the password based on a hash of the correct pass used on login.

I guess. Front-end would need to add additional unlock wallet calls in some places, and webserver would need to track password changes obviously. The RPC server would need to mirror all of the changes. It's not a clear win for me, though I understand the desire.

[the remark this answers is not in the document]
chappjc: I think guarding against a desktop that is not physically secure is fairly unnecessary, but alright.

> Front-end would need to add additional unlock wallet calls in some places, and webserver would need to track password changes obviously.

I'm talking about places where we do this only for the extra auth, _not_ when it's used to unlock the wallet.

This doesn't pertain to the RPC server.  Each RPC is authorized with the rpcuser/pass, and as I was describing above, the only thing Core needs to check is if the user had logged in prior.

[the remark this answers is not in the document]
martonp: What if the password was cached in core on login, perhaps optionally with a timeout? This is what dcrwallet does with the `walletpassphrase` command.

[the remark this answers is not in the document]
chappjc: I'm really not concerned with places where the crypter from the appPW is actually used, but the cases where we're using the backend to add a second auth.  Note the line of code this thread is commenting on.  It's simply a redundant authorization.

This is getting off on a tangent though because we've established the Send methods really might actually want to unlock on the fly, and I'm really fine with having them take a password for that use.  I'm just trying to figure out our policy of when we have a `Core` method check a password if it doesn't go anywhere else after the check.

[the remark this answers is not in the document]
martonp: I agree that there's no point to do redundant checks in core.

I don't mean to get too of topic, but the reason I was thinking about the caching was that by moving the bots out of `core`, the bot package needs to cache the password, because it needs to be sent to `Trade`, which unlocks wallets if they have not yet been unlocked. A password is also needed for `Cancel`, but that just does the redundant auth which can be removed.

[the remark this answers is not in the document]
martonp: If `Core` would become a go interface that other 3rd party applications could potentially program against, I think it would be a nicer interface without the additional password arguments after login.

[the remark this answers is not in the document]
buck54321: > What if the password was cached in core on login, perhaps optionally with a timeout? This is what dcrwallet does with the `walletpassphrase` command.

Please no. The RPC auth model is not appropriate for a universal-consumer interface like **core**.

> If `Core` would become a go interface that other 3rd party applications could potentially program against, I think it would be a nicer interface without the additional password arguments after login.

**core** already is a Go interface that other "3rd party apps" can program against. In fact, that's its entire purpose.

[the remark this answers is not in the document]
chappjc: > **core** already is a Go interface that other "3rd party apps" can program against. In fact, that's its entire purpose

That we all agree on.

The concerns of authorizing certain requests is all that's a bit misplaced sometimes (not talking about combining wallet unlock for convenience, which is generally ok).  Consider: Go consumer X, a middleware like webserver, provides the correct app pass to (*Core).Login.  Later, same Go consumer must provide it again for other things.  These are compile time relations, in the same application and memory space, unlike external connections like rpc/http.  So we just have some minor separation of concerns issues.  I think that's all martop is referring to in terms of the usability of the Core API.

Not to worry though -- I do NOT think caching the app password in Core is particularly relevant or appropriate.  This is either the job of middleware (webserver/rpcserver) or frontend itself.  An external bot can handle it imo without Core assistance.

The only thing I keep coming back to is getting clear about what core methods should even require an app password.  The answer in my options is those that use it, not just check it.

chappjc (approved):

PR #2321, opened 2023-04-24
site: update deps for yaml fix

chappjc:
`npm audit` is reporting a number of vulnerable packages due to a yaml bug.  Resolving this requires a major (semver) update to stylelint and the css minimizer.
These aren't runtime vulnerabilities, so we won't backport this, but we might as well do the major updates now that we're on a new dev cycle.


client/webserver/site/src/css/forms.scss
```diff
@@ -238,10 +238,7 @@ button.form-button {
   @extend .flex-center;
 
   position: fixed;
-  left: 0;
-  top: 0;
-  bottom: 0;
-  right: 0;
```
chappjc, line 244: https://stylelint.io/user-guide/rules/declaration-block-no-redundant-longhand-properties/

JoeGruffins (approved):

martonp (approved):

PR #2324, opened 2023-04-26
client/ui: simplify enable/disable wallet for token/parent wallets 

### LINKED ISSUE #2242: ui: simplify enable/disable wallet for token/parent wallets
Potential UX improvement to consider is to permit the following (if there aren't any fundamental limitations to doing this):
- disable all token wallets when parent wallet gets disabled
- enable parent wallet when token wallet gets enabled

or maybe **even simpler way to handle it** would be to remove this enable/disable functionality from token wallets and just let parent wallet fully manage it,

currently user needs to handle this manually:

imageimage

ukane-philemon:
Closes #2242.


chappjc:
I don't I agree that token wallets should not be able to be disabled.
You can create a token wallet some time after creating the ETH wallet, and disabling a token wallet was working just fine.  I don't think we should prohibit that.

The error here seemed really really clear to me:

https://user-images.githubusercontent.com/112318969/226815886-a178b280-5e11-42ff-9018-c74de45c8a84.png

client/rpcserver/handlers.go
```diff
@@ -1080,7 +1080,8 @@ var helpMsgs = map[string]helpMsg{
 	toggleWalletStatusRoute: {
 		pwArgsShort: "appPass",
 		argsShort:   `assetID disable`,
-		cmdSummary:  `Disable or enable an existing wallet.`,
+		cmdSummary: `Disable or enable an existing wallet. When disabling a parent wallet, all
+		token wallets that depend on the wallet will be disabled.`,
```
buck54321, line 1084: "parent wallet" is what we use in code, but may not me customary. How about something like "When disabling a chain's primary asset wallet, all token wallets for that chain will be disabled too."

client/core/core.go
```diff
@@ -2173,52 +2173,68 @@ func (c *Core) ToggleWalletStatus(assetID uint32, disable bool) error {
...
 
-		if wallet.connected() {
-			wallet.Disconnect() // before disable or it refuses
+		// Disconnect and disable all affected wallets.
+		for _, wallet := range affectedWallets {
+			if wallet.connected() {
+				wallet.Disconnect() // before disable or it refuses
+			}
+			wallet.setDisabled(true)
 		}
```
buck54321, line 2208: Might not matter, but we should probably disconnect and disable the parent wallet last.

client/webserver/locales/en-us.go
```diff
@@ -285,8 +285,8 @@ var EnUS = map[string]string{
 	"Pending":                     "Pending",
 	"disable_wallet":              "Disable Wallet",
 	"enable_wallet":               "Enable Wallet",
-	"disable_wallet_warning":      "Note: This wallet will not be connected to when you start the DEX client software and cannot be used until it is enabled.",
-	"enable_wallet_message":       "This wallet will resume operation and might take some time to sync.",
+	"disable_wallet_warning":      "Note: This wallet will not be connected to when you start the DEX client software and cannot be used until it is enabled. This will also disable all token wallets that depend on this wallet.",
+	"enable_wallet_message":       "This wallet will resume operation and might take some time to sync. This will also enable the parent wallet if this is a token wallet.",
```
buck54321, line 289: "If this is a token wallet, the wallet for the chain's primary asset will also be enabled."

buck54321 (changes requested):
Looking good. A couple of nits.

buck54321 (approved):

Issue #1960, opened 2022-11-16
shutdown of core should guarantee all subsystems are done


chappjc:
Got this in Decrediton testing.  This happened because after Core is shutdown (according to waitgroup), Decrediton's adapter closed the logger and nil'd the log rotator, however, it seems that one of the `asset.Wallet`s still hit the `TipChange` callback.  Need to see if there is a simple way for Core shutdown to do a bit better cleanup.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x7f8e719c7f07]

goroutine 240533 [running]:
github.com/jrick/logrotate/rotator.(*Rotator).Write(0x0, {0xc009a68480, 0x76, 0x2?})
	/home/jon/go/pkg/mod/github.com/jrick/logrotate@v1.0.0/rotator/rotator.go:133 +0x27
main.logWriter.Write({}, {0xc009a68480, 0x76, 0x78})
	/home/jon/github/decred/decrediton/modules/dex/libdexc/log.go:25 +0x56
github.com/decred/slog.(*Backend).printf(0xc00023d880, {0x7f8e721f330c, 0x3}, {0x7f8e721f3b88, 0x4}, {0x7f8e7222ddad, 0x29}, {0xc003e1e060, 0x2, 0x2})
	/home/jon/go/pkg/mod/github.com/decred/slog@v1.2.0/log.go:340 +0x342
github.com/decred/slog.(*slog).Errorf(0x0?, {0x7f8e7222ddad?, 0x7f8e7184e9a7?}, {0xc003e1e060?, 0x7f8e72b5db20?, 0x7f8e00000001?})
	/home/jon/go/pkg/mod/github.com/decred/slog@v1.2.0/log.go:466 +0x5f
decred.org/dcrdex/client/core.(*Core).tipChange(0xc0001f4000, 0x58acfd0?, {0x7f8e72cfa5a0, 0xc0037dfd60})
	/home/jon/go/pkg/mod/decred.org/dcrdex@v0.5.6/client/core/core.go:7147 +0x274
created by decred.org/dcrdex/client/core.(*Core).loadWallet.func1
	/home/jon/go/pkg/mod/decred.org/dcrdex@v0.5.6/client/core/core.go:2027 +0xaa
/home/jon/github/decred/decrediton/node_modules/electron/dist/electron exited with signal SIGABRT
error Command failed with exit code 1.
[...]

PR #2325, opened 2023-04-26
assets: ensure context cancellation take precedence over ticker during new block update

ukane-philemon:
Related fix for: #1960


chappjc (commented):
Looks good.  Thanks!

PR #2326, opened 2023-04-27
client/asset/btc: RPC connect and reconfig fixes

chappjc:
A few issues related to BTC/clone RPC connection and reconfiguration popped out:

- Running dexc on simnet, you can create a wallet that connects to bitciond/clone on mainnet (or reverse)
- You cannot switch to the correct one because if errors trying to validate the previous deposit address from the other network
- "Clone" asset always return `restartRequired = true` when attempting to reconfigure because a condition for live-reconfigurability compares the new wallet type against `"bitcoindRPC"` when clones have strings like `"dogecoindRPC"`.
- Reconfiguring a disconnected wallet could fail in certain circumstances because the context could already be canceled, blocking some of the reconfigure checks that involve an RPC with the new client.

### DIFF
--- a/client/asset/btc/rpcclient.go
+++ b/client/asset/btc/rpcclient.go
@@ -208,6 +208,9 @@ func (wc *rpcClient) reconfigure(cfg *asset.WalletConfig, currentAddress string)
 		restartRequired = true
 		return
 	}
+	if wc.ctx == nil || wc.ctx.Err() != nil {
+		return true, nil // not connected, ok to reconfigure, but restart required
+	}
 
 	parsedCfg := new(RPCWalletConfig)
 	if err = config.Unmapify(cfg.Settings, parsedCfg); err != nil {
--- a/client/asset/btc/rpcclient.go
+++ b/client/asset/btc/rpcclient.go
@@ -226,27 +226,34 @@ func (wc *rpcClient) reconfigure(cfg *asset.WalletConfig, currentAddress string)
 	// If the RPC configuration has changed, try to update the client.
 	oldCfg := wc.rpcConfig
 	if *newCfg != *oldCfg {
+		if wc.connectFunc != nil {
+			return true, nil // this asset needs a new rpcClient, then connect
+		}
+
 		cl, err := newRPCConnection(parsedCfg, wc.cloneParams.SingularWallet)
 		if err != nil {
 			return false, fmt.Errorf("error creating RPC client with new credentials: %v", err)
 		}
 
-		// If the wallet is in active use, check the supplied address.
-		if parsedCfg.ActivelyUsed {
-			// We can't use wc.ownsAddress because the rpcClient still has the
-			// old requester stored, so we'll call directly.
-			method := methodGetAddressInfo
-			if wc.legacyValidateAddressRPC {
-				method = methodValidateAddress
-			}
-
-			ai := new(GetAddressInfoResult)
-			if err := call(wc.ctx, cl, method, anylist{currentAddress}, ai); err != nil {
-				return false, fmt.Errorf("error getting address info with new RPC credentials: %w", err)
-			} else if !ai.IsMine {
+		// Require restart if the wallet does not own or understand our current
+		// address. We can't use wc.ownsAddress because the rpcClient still has
+		// the old requester stored, so we'll call directly.
+		method := methodGetAddressInfo
+		if wc.legacyValidateAddressRPC {
+			method = methodValidateAddress
+		}
+		ai := new(GetAddressInfoResult)
+		if err := call(wc.ctx, cl, method, anylist{currentAddress}, ai); err != nil {
+			return false, fmt.Errorf("error getting address info with new RPC client: %w", err)
+		} else if !ai.IsMine {
+			// If the wallet is in active use, check the supplied address.
+			if parsedCfg.ActivelyUsed { // deny reconfigure
 				return false, errors.New("cannot reconfigure to a new RPC wallet during active use")
 			}
-		}
+			// Allow reconfigure, but restart to trigger dep address refresh and
+			// full connect checks, which include the getblockchaininfo check.
+			return true, nil
+		} // else same wallet, skip full reconnect
 
 		chainInfo := new(getBlockchainInfoResult)
 		if err := call(wc.ctx, cl, methodGetBlockchainInfo, nil, chainInfo); err != nil {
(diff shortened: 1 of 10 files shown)


JoeGruffins (approved):

client/asset/btc/rpcclient.go
```diff
@@ -157,6 +171,17 @@ func (wc *rpcClient) connect(ctx context.Context, _ *sync.WaitGroup) error {
...
+	if err != nil {
+		return fmt.Errorf("getblockchaininfo error: %w", err)
+	}
+	if !chainOK(wc.cloneParams.Network, chainInfo.Chain) {
+		return errors.New("wrong net")
+	}
+	// Ensure we can get as well as decode an address from the wallet.
+	if _, err = wc.externalAddress(); err != nil {
+		return fmt.Errorf("address error: %w", err)
+	}
```
JoeGruffins, line 184: I'm not sure at all, but does this create a danger of going over the gap limit for these wallets from multiple connections with no use?

chappjc: Fair point.  It would take doing that 1000x I think, but still.

Actually, I'm not seeing why `(*xcWallet).Connect` would not error when getting or validating deposit address.

chappjc: Ah, `btcutil.DecodeAddress` does not check that segwit addresses' hrp of the prefix is for the intended network.  We have to do that: https://github.com/decred/dcrdex/pull/2326/commits/2302618501d840de6b7066c8c51c845915c26d2f
Removed the `externalAddress` request from `connect` since it is redundant with `(*xcWallet).Connect`'s call to get a deposit address if there isn't one.  Now it's only `reconfigure` that tries to get and check a new address, and even that is not fatal, instead just requiring a full wallet restart.

dex/networks/dgb/params.go
```diff
@@ -54,6 +55,7 @@ var (
 	})
 	// TestNetParams are the clone parameters for testnet.
 	TestNetParams = btc.ReadCloneParams(&btc.CloneParams{
+		Name:             "testnet4",
```
chappjc, line 58: None of these `Name`s are necessary, they are just for completion, and debugging.  Initially I wanted to have the `"chain"` string from `getblockchaininfo` compared against these values that we define as expected, but that's not where the nodes get the string to return in that field, so we just have the dumb switch with "main"/"test"/reg" instead.

client/asset/btc/rpcclient.go
```diff
@@ -256,6 +256,20 @@ func (wc *rpcClient) reconfigure(cfg *asset.WalletConfig, currentAddress string)
 			return false, errors.New("wrong net")
 		}
 
+		// Ensure we can get a new address and decode it.
+		if wc.addrFunc != nil { // with an addrFunc, cannot check since it likely uses the prior rpcClient
```
chappjc, line 260: The trailing comment here is actually a minor pitfall with the new approach of having the `addrFunc` implementation as a closure that captures the parent `asset.Wallet` concrete type, although I still like it in general.

client/asset/btc/rpcclient.go
```diff
@@ -225,6 +248,28 @@ func (wc *rpcClient) reconfigure(cfg *asset.WalletConfig, currentAddress string)
...
+		var addrStr string
+		if err := call(wc.ctx, cl, methodNewAddress, nil, &addrStr); err != nil {
+			wc.log.Warnf("Failed to retrieve address (%v). Full wallet reconnect may be required.", err)
+			return true, nil // force a restart so connect uses wc.externalAddress
+		}
+		if _, err = wc.decodeAddr(addrStr, wc.chainParams); err != nil {
+			wc.log.Warnf("Failed to decode address (%v). Full wallet reconnect may be required.", err)
+			return true, nil // force a restart so connect uses wc.externalAddress
+		}
+
```
buck54321, line 272: [This change](https://github.com/buck54321/dcrdex/compare/2302618...buck54321:dcrdex:reconfig-connect) should resolve #2330 too, if you can live with that funky atomic pattern.

chappjc: > if you can live with that funky atomic pattern.

Yeah, I think so.  I think there's a missing `return` in the `if passedChecks {` branch, but otherwise makes sense.
Although it might be a problem if there's stuff going on that uses the replaced client that is briefly pointing to a bad node.

chappjc: Yeah, I'm not so sure about that now.  There's also some unsafe assignments some fields of `rpcClient` that we consider safe in `connect` by convention (always `connect`/`Connect` before using it).

Possibly the following tweak to the current PR could resolve the need to remake the accounts for forcing full reconnect (new `rpcClient` instance):

```go
		if wc.connectFunc != nil || wc.addrFunc != nil {
			return true, nil // back to full reconnect
		}
```

buck54321: It seemed like we should be checking versions too.

buck54321: > There's also some unsafe assignments some fields of `rpcClient` that we consider safe in `connect` by convention (always `connect`/`Connect` before using it).

Like `rpcConfig` and `ctx`? Yeah. we definitely rely on our assumption about the user synchronizing connect/reconfig, but that's not new.

chappjc: Right.  Not new, but I think we have to stick with that convention.

chappjc: > It seemed like we should be checking versions too.

I see your point.  I'll factor out the meat of the checks from the `connect` method.

martonp (approved):

chappjc:
The commits from this PR that a desirable for 0.6.1 are:
- [reconfig disconnected wallets](https://github.com/decred/dcrdex/pull/2326/commits/b400b60bdbf4e46ac727cd6cae8e48ca73e55a17)
- [live reconfig of btc-clones](https://github.com/decred/dcrdex/pull/2326/commits/7c0973be1515795e816a10efcd2a48d7f05d5f98)

The others with added network checks would be good, but I don't want to risk breakage.  Plus the average user is less likely to be working with testnet and simnet.  They may be backported in the future.

client/asset/btc/rpcclient.go
```diff
@@ -1012,12 +1050,16 @@ func (wc *rpcClient) getBlockchainInfo() (*getBlockchainInfoResult, error) {
...
+		SubVersion      string `json:"subversion"`
 		ProtocolVersion uint64 `json:"protocolversion"`
 	}{}
 	err := wc.call(methodGetNetworkInfo, nil, r)
 	if err != nil {
 		return 0, 0, err
 	}
+	// TODO: We might consider checking getnetworkinfo's "subversion" field,
+	// which is something like "/Satoshi:24.0.1/".
+	wc.log.Debugf("Node at %v reports subversion \"%v\"", wc.rpcConfig.RPCBind, r.SubVersion)
```
chappjc, line 1062: For future checks:

```
"/Bitcoin Cash Node:26.0.1(EB32.0)/"
"/MagicBean:5.4.2/" (zcashd)
"/DigiByte:7.17.3/"
"/Satoshi:24.0.1/"
"/LitecoinCore:0.21.2.2/"
```

client/asset/btc/rpcclient.go
chappjc: I think I now have this as clear and foolproof as it can get.  If things get tricky when trying to handle changes to the RPC connection settings, it's best to fallback to full reconnect instead of "live" reconfigure.  I incorporated one aspect of your suggestion, which is to validate the dep address regardless of "active use" of the wallet and force a full wallet restart if it's not owned.  Also the presence of a `wc.connectFunc` will trigger that.

buck54321 (approved):

PR #2329, opened 2023-04-28
webserver/site: fix handling of NoAuth wallets

### LINKED ISSUE #2328: zec: create wallet form looks like native-RPC hybrid
image

same after F5 on the wallets page

ukane-philemon:
Closes #2328.

Zec wallet does not require a password and uses the `NoAuth` boolean field in its wallet def. But we seem to add that as criteria to display the single `Create` btn that is used only by seeded/token wallets that do not have required config opts.

The `walletDef.noauth` field is already properly handled by the last else statement in the block of code near code change.

### DIFF
--- a/client/webserver/site/src/js/forms.ts
+++ b/client/webserver/site/src/js/forms.ts
@@ -305,7 +305,7 @@ export class NewWalletForm {
     const page = this.page
     this.current.selectedDef = walletDef
     const appPwCached = State.passwordIsCached() || (this.pwCache && this.pwCache.pw)
-    Doc.hide(page.auth, page.oneBttnBox, page.newWalletPassBox)
+    Doc.hide(page.walletPassAndSubmitBttn, page.oneBttnBox, page.newWalletPassBox)
     const guideLink = walletDef.guidelink
     const configOpts = walletDef.configopts || []
     // If a config represents a wallet's birthday, we update the default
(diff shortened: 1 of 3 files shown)


client/webserver/site/src/js/forms.ts
```diff
@@ -326,7 +326,7 @@ export class NewWalletForm {
         break
       }
     }
-    const noWalletPWNeeded = walletDef.noauth || walletDef.seeded || Boolean(this.current.asset.token)
+    const noWalletPWNeeded = walletDef.seeded || Boolean(this.current.asset.token)
```
chappjc, line 329: I think the reason this was done wrong is because the variable `noWalletPWNeeded` was named incorrectly at conception.  It should really be `isCreated` afaict.

The other thing is that `page.auth` is badly named too.  It is a `div` that includes both the `newWalletPassBox` **and** the `submitAdd` button.

image

image

Almost every time I come to this code, I have to go figure out what `page.auth` is and why that should be shown when `noWalletPWNeeded` (because it actually still includes the Add/Create!) button.

So `page.auth` -> `page.walletPassAndSubmitBttn`?

Both suggested changes are code churn, but this code is hard to reason about presently.

chappjc (commented):
Looks better. :+1: 

image

I just have a couple thoughts in the review about issues that led to this error.

client/webserver/site/src/js/forms.ts
ukane-philemon: I renamed `noWalletPWNeeded` to `displayCreateBtn` cuz that's what it does basically. IMO `isCreated` raises more questions(what is created?).

chappjc (approved):

PR #2332, opened 2023-04-30
client/{mm,core}: Market maker balance segregation

martonp:
New configurations are added to the market maker config specifying what part of the wallet's balance each market maker will have control over. Using this config, the `MarketMaker` sets initial balances for each of the bots, and this is increased and decreased as the bot makes trades.

A `wrappedCore` is also added. This wraps the core methods with additional logic that checks and updates each bot's balances of each asset. A `wrappedCore` is passed to each bot, and the `core` methods behave as if the entire balances of the wallets are only what has been allocated to that bot.


client/mm/balance_handler.go
```diff
@@ -0,0 +1,949 @@
...
+	return c.core.SupportedAssets()
+}
+
+// SingleLotFees just forwards the call to the underlying core.
+func (c *coreWithSegregatedBalance) SingleLotFees(form *core.SingleLotFeesForm) (uint64, uint64, error) {
+	return c.core.SingleLotFees(form)
+}
+
+// Cancel just forwards the call to the underlying core.
+func (c *coreWithSegregatedBalance) Cancel(pw []byte, oidB dex.Bytes) error {
```
chappjc, line 548: can nuke pw now

client/mm/balance_handler.go
```diff
@@ -0,0 +1,949 @@
...
+}
+
+// coreWithSegregatedBalance implements the clientCore interface. A separate
+// instance should be created for each bot, and the core functions will behave
+// as if the entire balance of the wallet is the amount that has been reserved
+// for the bot.
+type coreWithSegregatedBalance struct {
+	balanceHandler *balanceHandler
+	botID          string
+	core           clientCore
```
chappjc, line 516: Named field is fine since it makes infinite recursion harder to do accidentally when shadowing a method like `Trade`, but it is possible to embed and remove all the dummy passthrough methods.  Did you have other considerations?

martonp: Nope, embedding sounds good.

client/asset/btc/btc.go
```diff
@@ -6149,6 +6149,33 @@ func (btc *baseWallet) FundMultiOrder(mo *asset.MultiOrder, maxLock uint64) ([]a
...
+	var inputSize, outputSize uint64
+	if btc.segwit {
+		inputSize = dexbtc.RedeemP2WPKHInputTotalSize
+		outputSize = dexbtc.P2WPKHOutputSize
+	} else {
+		inputSize = dexbtc.RedeemP2PKHInputSize
+		outputSize = dexbtc.P2PKHOutputSize
+	}
+
+	const numInputs = 12 // // plan for lots of inputs to get a safe estimate
```
JoeGruffins, line 6173: nit: double //

JoeGruffins (approved):

martonp:
I've moved the balance tracking to be part of `MarketMaker` instead of a separate `balanceHandler` struct. This is more convenient because the `orderInfo`s are also useful for creating the history of all actions a bot took and tracking the P/L.

client/mm/mm.go
```diff
@@ -88,34 +723,54 @@ func (m *MarketMaker) Run(ctx context.Context, cfgs []*BotConfig, pw []byte) err
...
 		switch {
 		case cfg.MMCfg != nil:
 			wg.Add(1)
 			go func(cfg *BotConfig) {
 				logger := m.log.SubLogger(fmt.Sprintf("MarketMaker-%s-%d-%d", cfg.Host, cfg.BaseAsset, cfg.QuoteAsset))
-				RunBasicMarketMaker(m.ctx, cfg, m.core, oracle, pw, logger)
+				mktID := dexMarketID(cfg.Host, cfg.BaseAsset, cfg.QuoteAsset)
+				RunBasicMarketMaker(m.ctx, cfg, m.wrappedCoreForBot(mktID), oracle, pw, logger)
 				wg.Done()
 			}(cfg)
```
buck54321, line 775: Is there any way to update the bot config, or do you have to shut down and restart?

client/mm/mm.go
```diff
@@ -73,6 +201,513 @@ func duplicateBotConfig(cfgs []*BotConfig) error {
...
+
+		baseBalance := balanceTracker[cfg.BaseAsset]
+		quoteBalance := balanceTracker[cfg.QuoteAsset]
+
+		var baseRequired, quoteRequired uint64
+		if cfg.BaseBalanceType == Percentage {
+			baseRequired = baseBalance.balanceAvailable * cfg.BaseBalance / 100
+		} else {
+			baseRequired = cfg.BaseBalance
+		}
```
buck54321, line 301: Do we need to validate the balance types across `[]*BotConfig` somehow? What if the user is trading on a couple markets with the same base asset, and they give `Percentage`  values that sum to > 100? What happens if they provide mixed `Percentage` and `Amount` balance types for the same asset?

buck54321 (changes requested):

client/mm/mm.go
martonp: This is being validated in `setupBalances`.

client/mm/mm.go
martonp: Currently you have to shut down and restart.

client/mm/wrapped_core.go
```diff
@@ -0,0 +1,442 @@
...
+// wrappedCore implements the clientCore interface. A separate
+// instance should be created for each bot, and the core functions will behave
+// as if the entire balance of the wallet is the amount that has been reserved
+// for the bot.
+type wrappedCore struct {
+	clientCore
+
+	mm    *MarketMaker
+	botID string
+	core  clientCore
```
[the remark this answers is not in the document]
martonp: No, because sometimes the `wrapped_core` needs to call a function with the same name in the actual core. In this case the field needs to be used, otherwise there will be an infinite loop.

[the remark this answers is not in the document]
buck54321: You can call methods on an embedded field by using the type name.

```Go
mkt, err := c.clientCore.ExchangeMarket(form.Host, form.Base, form.Quote)
```

client/mm/mm.go
buck54321: Ah. I guess you mean because it would trigger an error below at 
```Go
if baseRequired > baseBalance.balanceAvailable-baseBalance.balanceReserved { ...
```
I still wonder if it wouldn't be easy to screw up. As an example, say the balance of xyz is 2. We run one bot that reserves 1 xyz, and one bot that reserves 50% of xyz. If the balance of xyz changes to 1.5 for whatever reason, then those allocations cannot both be satisfied. 

OK for now, but just wanted to point it out.

client/mm/wrapped_core.go
```diff
@@ -0,0 +1,442 @@
...
+		Host:          host,
+		Base:          base,
+		Quote:         quote,
+		UseMaxFeeRate: true,
+	})
+	if err != nil {
+		return false, err
+	}
+
+	fundingFees, err := c.core.MaxFundingFees(base, uint32(len(placements)), options)
```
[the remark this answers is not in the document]
martonp: There is an `if sell` block above that returns early.

buck54321 (approved):
I still have some questions about how some things are going to work practically, but I'm comfortable with these changes and I want to get this in so we can get #2416 rebased and out of draft.

PR #1957, opened 2022-11-12
dexc-desktop: prepare installable desktop app

buck54321:
```
Introduce a new cmd, dexc-desktop. This coordinates a WebView
window with our normal dexc stack.

Move some common utilities to a new package client/app.
```

~~Prepare Debian archive for installation, including control files, Desktop Entry specificiation and icons. User can install with their preferred package manager.~~ Debian packaging moved to separate branch.

PR #2333, opened 2023-05-01
packaging: dexc-desktop packaging for Darwin

ukane-philemon:
This builds on top of #1957.

EDIT: `Decred DEX Client` in the images have been shortened to `Decred DEX`.
Display:
View in the Applications folder.
Screenshot 2023-05-01 at 9 54 12 AM

DMG installer view.
Screenshot 2023-05-01 at 1 42 01 PM

Dock view.
Screenshot 2023-05-01 at 9 52 20 AM

Screenshot 2023-05-01 at 9 52 00 AM

TODO:
- We might have to create icons of various sizes:
16x16 pixels (1x and 2x), 32x32 pixels (1x and 2x), 64x64 pixels (1x and 2x), 128x128 pixels (1x and 2x), 256x256 pixels (1x and 2x), 512x512 pixels (1x and 2x), 1024x1024 pixels (1x only) which will be converted to a single .icns file for MacOs to display the most appropriate icon anywhere.
- This PR includes a background image to instruct users on how to install the .app folder but we can improve on that too.

@martonp, please I'd appreciate your feedback on this since you also use MacOS.


client/cmd/dexc-desktop/pkg/pkg-darwin.sh
```diff
@@ -0,0 +1,334 @@
...
+
+# Build the webpack bundle prior to building the webserver package, which embeds
+# the files.
+if [[ -n "$1" ]]; then
+	pushd ../../../webserver/site
+	go generate # just check, no write
+	npm ci
+	npm run build
+	popd
+fi
```
ukane-philemon, line 301: This isn't required while testing but can be triggered.

client/cmd/dexc-desktop/pkg/pkg-darwin.sh
```diff
@@ -0,0 +1,334 @@
...
+		staple="$(xcrun stapler staple "$1")"
+		if [ $? -eq 0 ]; then
+			echo "The disk image is now notarized"
+		else
+			echo "$staple"
+			echo "The notarization failed with error $?"
+			exit 1
+		fi
+	fi
+}
```
ukane-philemon, line 129: These aren't been used atm since a $99 dev account is required to get an Apple dev signature for code signing. Notarizing the dexc-desktop dmg is debatable but IMO we might not need it.

chappjc:
Can rebase with dexc-desktop merged. No rush though.

ukane-philemon:
Uhmm, I'm on the markets page and everything looks fine except I now noticed the order form is a mess, no js error.

Screenshot 2023-05-10 at 4 00 43 PM

client/cmd/dexc-desktop/pkg/pkg-darwin.sh
```diff
@@ -0,0 +1,334 @@
...
+
+TARGETS="darwin/amd64" # unable to build for darwin/arm64 on a darwin/amd64 machine.
+FLAVOR="-tray"
+TAGS_DEXC="systray"
+prepare
+build_targets
+cleanup
+
+pushd ./installers
+sha256sum *.dmg > dexc-v${VER}-manifest.txt
```
martonp, line 333: I think this `sha256sum` is just on linux. I have `shasum -a 256` on mac.

ukane-philemon: Surprisingly, I have both.

martonp (commented):
The open logs folder selector is not working. 

Why don't we just include the create-dmg script in the repo and have pkg-darwin.sh call the create-dmg script? Then we can just grab any updates they make.

chappjc:
> The open logs folder selector is not working.

Did you guys try converting spaces as per my code comment: https://github.com/decred/dcrdex/blob/c07ac7163339deb35def662ff44cca34c238f13f/client/app/fileurl_darwin.go#L21-L23

ukane-philemon:
>Did you guys try converting spaces as per my code comment:

There was no space in the resulting folder url.

ukane-philemon:
Why don't we just include the create-dmg script in the repo and have pkg-darwin.sh call the create-dmg script? Then we can just grab any updates they make.

IMO, we can do with less code independently. Updates can be added if need be.

martonp:
> IMO, we can do with less code independently. Updates can be added if need be.

Why would it be more code to use their script? It seems like most of this is copy-pasted from there.

client/cmd/dexc-desktop/pkg/pkg-darwin.sh
```diff
@@ -0,0 +1,332 @@
...
+		  <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+		  <plist version="1.0"><dict>
+		  <key>CFBundleDevelopmentRegion</key><string>en</string>
+		  <key>CFBundleDisplayName</key><string>${APP_NAME}</string>
+		  <key>CFBundleExecutable</key><string>${APP_NAME}</string>
+		  <key>CFBundleIconFile</key><string>${ICON_FILE_NAME}</string>
+		  <key>CFBundleIdentifier</key><string>com.decred.dcrdex</string>
+		  <key>CFBundleInfoDictionaryVersion</key><string>1.0</string>
+		  <key>CFBundleName</key><string>${APP_NAME}</string>
+		  <key>CFBundleShortVersionString</key><string>1.0.0</string>
```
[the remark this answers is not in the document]
ukane-philemon: I don't think so. It's for the bundle which is different from the app version. You can check the Info.plist of other applications installed on your device to confirm.

ukane-philemon:
> Why would it be more code to use their script? It seems like most of this is copy-pasted from there.

Yes, I had to use a chunk of their code since it solves the issue at hand. I thought we'd prefer a custom build flow but @chappjc can decide which is better.

ukane-philemon:
Encountered these display issues and added commits to resolve them in this PR. @martonp please confirm if you can reproduce the weird display without these changes and if these changes fix them on your device. 

Screenshot 2023-05-13 at 10 56 39 AM
Screenshot 2023-05-13 at 10 56 15 AM
Screenshot 2023-05-13 at 10 58 24 AM
Screenshot 2023-05-13 at 6 49 05 AM
Screenshot 2023-05-13 at 10 58 01 AM

ukane-philemon:
When I resize the web view window on my Mac with the dev console open, it chops off the top of the page by an amount equal to the console's minimum height. This issue persists even after refreshing the page, but clicking on the resize icon at the top right corner when the dev console is closed resolves it. I haven't observed this problem on Safari or other browsers.

However, this behaviour doesn't occur when resizing the window with the dev console detached. It may be considered a non-issue since we don't anticipate users resizing the web view window while the dev console is open.

https://github.com/decred/dcrdex/assets/57448127/b6c6e529-2679-48ed-8b41-3df23b7a1ab8

chappjc:
> > Why would it be more code to use their script? It seems like most of this is copy-pasted from there.
> 
> Yes, I had to use a chunk of their code since it solves the issue at hand. I thought we'd prefer a custom build flow but @chappjc can decide which is better.

I have no opinion on the issue.  It's not clear to me what it would look like doing it the way @martonp suggests.  Is it this? https://github.com/create-dmg/create-dmg
It's up to you guys.

ukane-philemon:
> Is it this? https://github.com/create-dmg/create-dmg

Yes, it is.

client/cmd/dexc-desktop/pkg/pkg-darwin.sh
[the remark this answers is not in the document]
ukane-philemon: Thanks for the review and testing, I pushed your review changes in https://github.com/decred/dcrdex/pull/2333/commits/26be2ab406cdea63a64ebff8efbfbf0782900d2a.

ukane-philemon:
> I'm fixing up the "Open logs folder" functionality in https://github.com/decred/dcrdex/pull/2351.

Nice, tested well on MacOS.

ukane-philemon:
> Why don't we just include the create-dmg script in the repo and have pkg-darwin.sh call the create-dmg script? Then we can just grab any updates they make.

I agree this can eliminate additional complexity and maintenance overhead, regardless of the amount of code involved. We can have it if the others agree on this @martonp.

chappjc:
I think you should use grab the  create-dmg in full, and make the adjustments in our separate script, add @martonp says.  Does that work out?

martonp (commented):
The UI is looking good now, however I see some other issues.

On the tray icon, when I click "Open" it is always opening a new window even if one is already open. Also if another App is in the foreground and I click the "Open", it will just create a new window behind the app that is in the foreground.

I actually think on Mac the tray icon is not the way to go. When there is an app running on mac, even if there is not a window open, it will still be displayed on the dock at the bottom of the screen. When you right click the icon at the bottom of the screen, there are usually some options. The "Open logs folder" option can go over there. It feels strange that the app is running but I see no icon on the bottom dock. Every other app is there.

chappjc:
> On the tray icon, when I click "Open" it is always opening a new window even if one is already open. Also if another App is in the foreground and I click the "Open", it will just create a new window behind the app that is in the foreground.

I was surprised by this too, but @buck54321 makes the point that this is like making multiple tabs, which webview does not support.

> I actually think on Mac the tray icon is not the way to go. When there is an app running on mac, even if there is not a window open, it will still be displayed on the dock at the bottom of the screen. When you right click the icon at the bottom of the screen, there are usually some options. 

On all systems, I made my case in chat that the tray icon is pointless except when all windows are closed and it's just waiting for orders to settle or something.  On this note, I still don't like that closing the "last window" should terminate the singular dex process running `Core`.  I think that the sole utility of a tray icon is so that you can close all the windows and let it continue running, where the tray provides a way to resurrect a window.

On the Mac specific points, even if we adjust things as per my previous paragraph, it's debatable if the tray icon is desirable or not.  Only the main dexc process that has the running Core serves this tray.  Only that process would be capable of providing the same menu items but on the doc as you suggest.
Also jrick made the point in chat that it's still nice to have some visual other than the subtle black dots in the Dock.

I don't use Mac enough to have an informed opinion, but I do think the functionality of the tray icon on all systems is flawed (the "last window" kills the Core process design).

client/cmd/dexc-desktop/pkg/pkg-darwin.sh
```diff
@@ -0,0 +1,132 @@
...
+		  <key>CFBundleDevelopmentRegion</key><string>en</string>
+		  <key>CFBundleDisplayName</key><string>${APP_NAME}</string>
+		  <key>CFBundleExecutable</key><string>${APP_NAME}</string>
+		  <key>CFBundleIconFile</key><string>${ICON_FILE_NAME}</string>
+		  <key>CFBundleIdentifier</key><string>com.decred.dcrdex</string>
+		  <key>CFBundleInfoDictionaryVersion</key><string>1.0</string>
+		  <key>CFBundleName</key><string>${APP_NAME}</string>
+		  <key>CFBundleShortVersionString</key><string>${VER%%-*}</string>
+		  <key>CFBundleVersion</key><string>${BUILD_VER}</string>
+		  <key>CFBundleSignature</key><string>????</string>
```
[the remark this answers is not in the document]
ukane-philemon: dexc sounds good.

ukane-philemon:
> On the tray icon, when I click "Open" it is always opening a new window even if one is already open. Also if another App is in the foreground and I click the "Open", it will just create a new window behind the app that is in the foreground

I understand your concerns. We had some discussion around this on matrix, and as @chappjc has said, creating sub webview window is flawed. Also, creating a new webview window will make another icon appear in the dock and it was unpleasant to look at, so I had to add a .plist item to make it behave like a background app so users can create multiple windows without displaying unnecessary icons in the dock. Our use of a separate dexc process to create a new window allow for multiple webview window on Mac and I don't know if there's a workaround.

martonp:
Why does anyone need to have multiple dexc windows open. 

I don't see any subtle black dots. I see this:

Screenshot 2023-05-22 at 4 22 49 AM

On Mac, all the applications that are running, whether or not there is a window open, are displayed on the bottom dock like this, with a dot below them.

Screenshot 2023-05-22 at 4 23 40 AM

Even if I close all the windows, the app will still stay running. To actually terminate the program, you have to right click the icon and select "Quit." I think the ideal behavior would be that if there are no active trades it just quits. If there are active trades it should give a confirmation alert first.

I definitely don't think this should be a background app. I think there should just be one window allowed, get rid of the dock on top, and just have the bottom dock icon behavior as I wrote in the previous paragraph.

chappjc:
> Why does anyone need to have multiple dexc windows open.

I tend to agree.  I was confused.

> I don't see any subtle black dots. I see this:
> 
> Screenshot 2023-05-22 at 4 22 49 AM
> On Mac, all the applications that are running, whether or not there is a window open, are displayed on the bottom dock like this, with a dot below them.

Those are the two options, yes.  The black dots option I'm talking about is actually those gray dots there.  If that's how mac users work, ok.

> Screenshot 2023-05-22 at 4 23 40 AM
> Even if I close all the windows, the app will still stay running. To actually terminate the program, you have to right click the icon and select "Quit." I think the ideal behavior would be that if there are no active trades it just quits. If there are active trades it should give a confirmation alert first.

I haven't tested this PR, but dexc-desktop right now kills the app when all the windows are closed, unless there are trades running.

> I definitely don't think this should be a background app. I think there should just be one window allowed, get rid of the dock on top, and just have the bottom dock icon behavior as I wrote in the previous paragraph.

What is wrong with having it as a background app?  You don't need the window when you have placed all your orders and just want to let it work.  We don't need the systray bits to make that work on Mac because of how the dock works, I acknowledge you points, but I still think it's desirable.  Are you saying that on mac it's just as good to *minimize* the window to achieve this?

ukane-philemon:
I have been trying all I suggestions I found all day if we can somehow do it the `Mac way` with webview and these are some of  the issues.

1. Webview.Run() panics on macOS when it is not  executed on the main thread. So we can't create new windows from a goroutine without using the current setup (executing dexc-desktop with the current webserver url). This will create new icon in the dock for each window (this didn't look right so i added `<key>LSUIElement</key><true/>` to make all the dock icons disappear even when there are multiple windows open)

2. If we decide to maintain a single window for MacOS, we can but after testing and trying to achieve this on my end I noticed MacOS exits dexc-desktop when the single window is closed (and the app does not get the signal), whether there is an active order or not. This means we can't send notes and users with pending orders will have to use the `Hide` button or minimize the window.

3. We can keep the current setup for now, dexc-desktop will manage all processes, notify users if the are pending orders, run in the background (with the systray visible to show sign of life). Just that there will be no little dot below the icon in the dock. Users will have to use the systray to manage dexc-desktop (shutdown/create new windows).

> Even if I close all the windows, the app will still stay running

For some apps, yes. But these apps usually have control over the windows but it's not the same way with webview lib.

martonp (approved):
DMG stuff LGTM. I think it may be possible to have the application be hidden instead of closed when the window closes by calling some objective-c code in the cgo section, but I haven't figured out exactly how yet.

I remove the `<key>LSUIElement</key><true/>` element and also created the webview in the original process instead of using the subprocess. Then if I right click the icon and select "hide", that is the exact behavior we want when the red button is clicked.

buck54321:
> 1. (this didn't look right so i added `<key>LSUIElement</key><true/>` to make all the dock icons disappear even when there are multiple windows open)

Hmm. Do you mean that you can't pin an icon to the dock anymore? 

Sorry for the questions. I'm trying to get a mac right now. Hopefully I'll be more helpful soon.

ukane-philemon:
> I remove the <key>LSUIElement</key><true/> element and also created the webview in the original process instead of using the subprocess. Then if I right click the icon and select "hide", that is the exact behavior we want when the red button is clicked.

But sadly we don't have control over the red btn yet when running webview in the original process.

ukane-philemon:
> Are you saying when running in the background, they can't open a window by clicking on the app's icon in the dock?

Yep.

> Hmm. Do you mean that you can't pin an icon to the dock anymore?

You can, for easy access.

> Sorry for the questions. I'm trying to get a mac right now. Hopefully I'll be more helpful soon.

np.

martonp:
If this returned 0 instead of 1, I think we'd be good...

https://github.com/webview/webview/blob/8387ff8945fc010e7c4203c021943ce4ca12a276/webview.h#LL832C27-L832C74

ukane-philemon:
> If this returned 0 instead of 1, I think we'd be good...
> 
> https://github.com/webview/webview/blob/8387ff8945fc010e7c4203c021943ce4ca12a276/webview.h#LL832C27-L832C74

This will cause dexc-desktop to continue running (without a window), but we won't be able to "resurrect" the window if the user clicks on the dock icon again.

ukane-philemon:
> Would there be a way to add an option on the icon menu to create a new window?

Other apps have this. But I don't see how we can achieve this with webview.

chappjc:
@ukane-philemon Would you mind rebasing and squashing down with a new commit message?  Thanks.

buck54321:
> > Are you saying when running in the background, they can't open a window by clicking on the app's icon in the dock?
> 
> Yep. But the systray will be visible and users can open new windows from there.

This is going to cause a lot of confusion. There's got to be some signal we can work with to catch them clicking the launcher icon, no? Or does Mac just completely ignore it because of the systray entry?

chappjc:
I think @ukane-philemon is properly addressing this from https://github.com/decred/dcrdex/pull/2372.  This PR is primarily dealing with the dmg packaging.

> > > Are you saying when running in the background, they can't open a window by clicking on the app's icon in the dock?
> > 
> > 
> > Yep. But the systray will be visible and users can open new windows from there.
> 
> This is going to cause a lot of confusion. There's got to be some signal we can work with to catch them clicking the launcher icon, no? Or does Mac just completely ignore it because of the systray entry?

I think @ukane-philemon is properly addressing this from https://github.com/decred/dcrdex/pull/2372.  This PR is primarily dealing with the dmg packaging.

client/webserver/site/.stylelintrc
```diff
@@ -12,6 +12,8 @@
         "scss/double-slash-comment-empty-line-before": null,
         "scss/at-import-no-partial-leading-underscore": null,
         "scss/at-import-partial-extension": null,
-        "scss/at-extend-no-missing-placeholder": null
+        "scss/at-extend-no-missing-placeholder": null,
+        "declaration-block-no-redundant-longhand-properties": null,
+        "value-no-vendor-prefix": null
```
chappjc, line 17: Why these changes?

client/webserver/site/src/css/forms.scss
```diff
@@ -238,7 +238,10 @@ button.form-button {
   @extend .flex-center;
 
   position: fixed;
-  inset: 0;
```
chappjc, line 241: Why undoing this?  A linter update recently made us do this, and I don't think we should disable that linter.

client/webserver/site/src/html/orders.tmpl
```diff
@@ -35,12 +35,12 @@
         {{end}}
         <div class="apply-bttn d-hide"><button class="bg2 demi">[[[apply]]]</button></div>
       </div>
-      <div id="exportOrders" class="export-archived-records-bttn mt-3">
+      <div id="exportOrders" class="export-archived-records-bttn mt-3 pb-2">
```
chappjc, line 38: Will you please comment on all these frontend changes?  I didn't notice them in prior review, and I don't think they are relevant to the DMG package additions.

client/cmd/dexc-desktop/pkg/pkg-darwin.sh
```diff
@@ -0,0 +1,132 @@
...
+SCRIPTPATH="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+INSTALLERS_DIR="$SCRIPTPATH/installers"
+APP_DIR="${SCRIPTPATH}/${APP_NAME}.app"
+CONTENTS_DIR="${APP_DIR}/Contents"
+RESOURCES_DIR="${CONTENTS_DIR}/Resources"
+APP_EXCE_DIR="${CONTENTS_DIR}/MacOS"
+
+# For fancy disk image container icon and background.
+ICON_FILE_NAME=dexc-icon.icns
+ICON_FILE="${SRC_DIR}/dexc-icon.icns"
```
buck54321, line 37: Our icon doesn't seem to be standardized.
image

client/webserver/site/src/css/forms.scss
ukane-philemon: Not supported on webkit. https://github.com/decred/dcrdex/pull/2333#issuecomment-1546698329

client/webserver/site/src/html/orders.tmpl
ukane-philemon: The display didn't look right on webkit. See https://github.com/decred/dcrdex/pull/2333#issuecomment-1546698329.

client/cmd/dexc-desktop/pkg/pkg-darwin.sh
ukane-philemon: Kinda.

> We might have to create icons of various sizes:
16x16 pixels (1x and 2x), 32x32 pixels (1x and 2x), 64x64 pixels (1x and 2x), 128x128 pixels (1x and 2x), 256x256 pixels (1x and 2x), 512x512 pixels (1x and 2x), 1024x1024 pixels (1x only) which will be converted to a single .icns file for MacOs to display the most appropriate icon anywhere.

We can improve on that, but maybe in another work?

client/cmd/dexc-desktop/pkg/pkg-darwin.sh
```diff
@@ -0,0 +1,132 @@
...
+# Bail out on any unhandled errors
+set -e;
+
+# Any command that exits with non-zero code will cause the pipeline to fail
+set -o pipefail;
+
+export GOWORK=off
+
+# For release set metadata to "release".
+VER=`cat ../../../app/VERSION`
```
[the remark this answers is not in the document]
ukane-philemon: It's in this PR.

client/webserver/site/.stylelintrc
ukane-philemon: To allow for changes fixing  https://github.com/decred/dcrdex/pull/2333#issuecomment-1546698329

ukane-philemon:
Installer view updated in https://github.com/decred/dcrdex/pull/2333/commits/aab94469f907d0cde24eb4d3c4d2132e07f3a597: 
Screenshot 2023-05-27 at 7 42 11 PM

client/cmd/dexc-desktop/pkg/pkg-darwin.sh
chappjc: > While it's not necessary, we can create an icon with rounded borders like the others in the dock image. I used the softened-icon-dark.png logo in the client/webserver/site/img folder.

I assume that's what @buck54321 meant by "standardized".  Probably should although this PR is fine to merge now IMO.

client/webserver/site/src/css/forms.scss
chappjc: Oh, right.  That's kinda messed up that webkit doesn't support `inset.  It is supported in every major browser, including Safari since v14.1: https://caniuse.com/mdn-css_properties_inset
What do we know about the webkit version provided by macdrive?

ukane-philemon: I running Safari ver 15.6.1 but know nothing about the WebKit version provided by Macdrive. I think it will use whatever webkit version it finds. 

To get Webkit Version on macOS run: ` defaults read /System/Library/Frameworks/WebKit.framework/Resources/Info.plist CFBundleShortVersionString
` I got `15609` which corresponds with Safari ver 15.6.1 according to [ the comment here ](https://stackoverflow.com/questions/4590102/how-to-check-which-safari-version-is-installed-on-mac-programmatically#:~:text=Show%20activity%20on%20this%20post,%2FWebkit%2FResource%2Finfo.)

client/cmd/dexc-desktop/pkg/pkg-darwin.sh
ukane-philemon: Found time to design one on Figma yesterday: https://www.figma.com/file/ao9XWuHAKDUJWfq3w2kjgD/dexc-macOS-icon?type=design&node-id=0-1&t=204FpylWMx9s29kT-0
@buck54321, does it look good?

I don't know where design superpower came from cuz I'm really not good with design tools 😄 .

buck54321 (approved):

buck54321:
Notably the window is stuck on top like a modal. But I'm guessing #2372 will fix that.

client/cmd/dexc-desktop/pkg/pkg-darwin.sh
```diff
@@ -0,0 +1,132 @@
...
+pushd ../../../webserver/site
+go generate # just check, no write
+npm ci
+npm run build
+popd
+
+function build_targets() {
+  for TARGET in ${TARGETS}; do
+    OS=${TARGET%%/*}
+    ARCH=${TARGET##*/}
```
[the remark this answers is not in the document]
ukane-philemon: Just getting the first and last part of `TARGET` separated by `/`.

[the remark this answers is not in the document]
chappjc: This was all added in https://github.com/decred/dcrdex/pull/1859. Voodoo to me too.

ukane-philemon:
> Notably the window is stuck on top like a modal.

Stuck? Can I see a video of what you mean?

client/cmd/dexc-desktop/pkg/pkg-darwin.sh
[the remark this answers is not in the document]
ukane-philemon: Here is a proper explanation: https://chat.openai.com/share/9dfbd2c1-56a5-42bf-a0eb-970392f11e75

ukane-philemon:
> [Video](https://drive.google.com/file/d/1eeG9gnha5DkPozn5YKQYAMOatN7Xxd2f/view?usp=sharing)

Seen and was able to reproduce too with our current use of webview but this is resolved in https://github.com/decred/dcrdex/pull/2372 since we are using native macOS APIs.

PR #2336, opened 2023-05-03
zec: implement fee-rate to short-circuit fee checks

buck54321:
Zcash uses a standard fee of 1000 zats per tx, regardless of tx size. We aren't plumbed for that, and with ZIP-0317 not activated, we probably never will be. Instead, use a standard rate of 10 zats / byte which should guarantee that any tx will have > 1000 zats in fees.


client/asset/zec/zec.go
```diff
@@ -564,26 +566,9 @@ func zecTx(tx *wire.MsgTx) *dexzec.Tx {
...
-	if err != nil {
-		return 0, err
-	}
-	if feeRate <= 0 {
-		return 0, fmt.Errorf("fee could not be estimated")
-	}
-	return uint64(math.Round(feeRate * 1e5)), nil
+// estimateFee returns the asset standard legacy fee rate.
+func estimateFee(context.Context, btc.RawRequester, uint64) (uint64, error) {
+	return dexzec.LegacyFeeRate, nil
```
chappjc, line 571: To clarify, this is for coherence, even though we don't have a code path that executes this anymore?

chappjc (approved):

client/asset/zec/zec.go
```diff
@@ -259,6 +256,11 @@ type zecWallet struct {
 	lastAddress atomic.Value // "string"
 }
 
+// FeeRate returns the asset standard fee rate for Zcash.
+func (w *zecWallet) FeeRate(context.Context) (uint64, error) {
```
chappjc, line 260: I think this is incorrect.  If the goal is to satisfy a `client/asset.FeeRater`, this does not do it.

I started writing a test like this:

```go
func TestFeeRate(t *testing.T) {
	wi, err := NewWallet(&asset.WalletConfig{
		Type: walletTypeRPC,
		Settings: map[string]string{
			"rpcuser":     "user",
			"rpcpassword": "pass",
			"rpcport":     "33770",
		},
		PeersChange: func(uint32, error) {},
		TipChange:   func(err error) {},
	}, dex.StdOutLogger("tst", dex.LevelDebug), dex.Simnet)
	if err != nil {
		t.Fatal(err)
	}

	ctx, cancel := context.WithCancel(context.Background())
	cancel() // context must be ignored
	rate, err := wi.(*zecWallet).FeeRate(ctx)
	if err != nil {
		t.Fatal(err)
	}
	if rate != 11 {
		t.Error("wrong")
	}
}
```

But then went to do `wi.(asset.FeeRater).FeeRate` and realized it didn't apply.

chappjc: Noticed something related:

```go
var _ asset.FeeRater = (*ExchangeWalletNoAuth)(nil) // FAIL, does not implement FeeRater
```

because

```go
func (btc *ExchangeWalletFullNode) FeeRate() uint64 {
```

I'm pretty certain that's a mistake.  It should be a `intermediaryWallet` method.

chappjc: Well, I really don't want to make another layer of intermediary wallet types for the sake of making `ExchangeWalletNoAuth` a `FeeRater`, so Iet's just dup the method and add the assertion:

```go
var _ asset.FeeRater = (*ExchangeWalletFullNode)(nil)
var _ asset.FeeRater = (*ExchangeWalletNoAuth)(nil)

func (btc *intermediaryWallet) feeRaterHelper() uint64 {
	// NOTE: With baseWallet having an optional external fee rate source, we may
	// consider making baseWallet a FeeRater by allowing a nil local func.
	rate, err := btc.feeRate(1)
	if err != nil {
		btc.log.Tracef("Failed to get fee rate: %v", err)
		return 0
	}
	return rate
}

func (btc *ExchangeWalletFullNode) FeeRate() uint64 {
	return btc.feeRaterHelper()
}

func (btc *ExchangeWalletNoAuth) FeeRate() uint64 {
	return btc.feeRaterHelper()
}
```

martonp: Nothing wrong with it, but I'm just wondering why does `ExchangeWalletNoAuth` need to be a `FeeRater`? In the ZCash `NewWallet` function we return the `zecWallet` type.

chappjc: Just a related bug.

dex/networks/zec/params.go
```diff
@@ -18,6 +18,14 @@ const (
...
 	InitTxSizeBase = MinimumTxOverhead + btc.P2PKHOutputSize + btc.P2SHOutputSize // 29 + 34 + 32 = 95
 	InitTxSize     = InitTxSizeBase + btc.RedeemP2PKHInputSize                    // 95 + 149 = 244
+
+	// LegacyFeeRate returns a standard 10 zats / byte. Prior to ZIP-0317, Zcash
+	// used a standard tx fee of 1000 zats, regardless of tx size. We don't
+	// handle fees correctly yet though, so just use a fee / byte that
+	// guarantees a tx fee of > 1000 zats. A single input spending a p2pkh
+	// output is 149 bytes, so 10 zats / byte should be sufficient for any
+	// tx.
+	LegacyFeeRate = 10
```
chappjc, line 28:
```suggestion
	// However, zcashd v5.5 begins making stricter fee requirements for both
	// relay and block inclusion. The release notes state that relay by default
	// still works with 1000 zat fee txns, but it may be adjusted by operators,
	// and it may need to be set to the higher ZIP 317 rate. For mining, there
	// is a small allowance on the number of "unpaid actions" allowed in a
	// block, so we should take care to pay the ZIP 317 "conventional" rate,
	// which is multiples of 5000 zats, and a minimum of 10000. To ensure we
	// have no unpaid actions in our (transparent) transactions, we need a
	// higher rate. For a 242 byte transaction, like a swap init, we can emulate
	// this with about 42 zats/byte. Even with 100 zats/byte, our typical redeem
	// of ~342 bytes would pay 34200 zats, which is only about a penny, so to
	// ensure our transactions are relayed and mined, we go with a high rate.
	LegacyFeeRate = 84
```

I think it would be prudent to put that rate on v0.6.1 as well.  We've already merged the `LegacyFeeRate` on `release-v0.6`, but we'd update it.

JoeGruffins (approved):

chappjc (approved):

PR #2337, opened 2023-05-03
zec: implement FeeRate to short-circuit fee checks - release-v0.6

buck54321:
Counterpart of #2336 for release-v0.6 branch. We don't have a `zecWallet` type yet, so I'm not adding the `FeeRate` to the zec wallet yet, but setting the `FeeEstimator` is effectively the same thing.

### DIFF
--- a/client/asset/zec/zec.go
+++ b/client/asset/zec/zec.go
@@ -5,10 +5,7 @@ package zec
 
 import (
 	"context"
-	"encoding/json"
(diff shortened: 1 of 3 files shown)


chappjc:
> but setting the `FeeEstimator` is effectively the same thing

Since there's no external estimator either, yeah, same thing.

chappjc (approved):

PR #2338, opened 2023-05-03
zec: implement zip 317

buck54321:
Adds support for ZIP-0317 fee calculations, though we've always done fees for Zcash wrong. Previously, fee should have been 1000 zats per tx, regardless of size. Now, fees are 5000 zats per "logical action".

Making these calculations possible required a pretty deep integration. I've tried to do so in a way that makes this easy to review. This has been roughly tested manually.


server/asset/zec/zec.go
```diff
@@ -148,6 +151,39 @@ func (be *ZECBackend) Contract(coinID []byte, redeemScript []byte) (*asset.Contr
...
+	}
+
+	return fees >= zecTx.TxFeesZIP317()
+}
+
+// CalcOrderFunds is the ZIP-0317 compliant version of calc.RequiredOrderFunds.
+// Satisfies the asset.OrderEstimator interface.
+func (be *ZECBackend) CalcOrderFunds(swapVal, inputCount, inputsSize, maxSwaps uint64) uint64 {
+	return dexzec.RequiredOrderFunds(swapVal, inputCount, inputsSize, maxSwaps)
+}
```
martonp, line 185:
``` golang
var _ asset.OrderEstimator = (*ZECBackend)(nil)
```

martonp (commented):
It's looking good to me. Still need to test though.

docs/images/zip-0317.tex
```diff
@@ -0,0 +1,20 @@
+\documentclass[]{article}
```
chappjc, line 1: Why don't we just link to the ZIP 317 page?  I don't think we need to check in a 120 KB png file either.

buck54321: Ooooh. I was reading at https://github.com/zcash/zips/blob/main/zip-0317.rst, where the tex doesn't render

client/asset/btc/btc.go
```diff
@@ -4951,11 +5031,10 @@ var dummyP2PKHScript = []byte{0x76, 0xa9, 0x14, 0xe4, 0x28, 0x61, 0xa,
 
 // EstimateSendTxFee returns a tx fee estimate for sending or withdrawing the
 // provided amount using the provided feeRate.
-func (btc *intermediaryWallet) EstimateSendTxFee(address string, sendAmount, feeRate uint64, subtract bool) (fee uint64, isValidAddress bool, err error) {
+func (btc *baseWallet) EstimateSendTxFee(address string, sendAmount, feeRate uint64, subtract bool) (fee uint64, isValidAddress bool, err error) {
```
buck54321, line 5034: This makes the electrum wallets `TxFeeEstimator`s. Anything wrong with that @chappjc?

chappjc: Not in principle.  With your move of `estimateSendTxFee` to be a `baseWallet` method, it looks like the existing `(*electrumWallet).listUnspent` is adequate.

buck54321:
Replaced by #2553

PR #2339, opened 2023-05-03
server/asset/btc: compile time configurable poll interval

chappjc:
https://github.com/decred/dcrdex/commit/12147ba908e9c17dcf3cce7e23231796afb0b428 but for BTC
1 second polling is too fast for remote nodes.
This variable affects all btc clone assets.


chappjc:
> Working as intended, but why a compile time setting rather than a config setting? I think the locktimes are compile time because they should never be touched by node operators, but this seems like it would be useful for everyone. I guess I missed commenting on this for eth.

Niche I guess. Just solving a practical issue with minimal code bloat

PR #2343, opened 2023-05-05
server/asset/zec: fixes for bond and 0-conf tx fee checks

chappjc:


chappjc:
Tested bond with every single asset that is a Bonder.

PR #2347, opened 2023-05-09
docs: Arch Linux package and OS packaging README update

chappjc:
I created an Arch User Repository (AUR) package for the dcrdex client here: https://aur.archlinux.org/packages/dcrdex ([PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=dcrdex))

If you run Arch or a derivative, you can install that with one of the AUR helpers like `yay` or `pamac`.

This PR calls out our current push for packaging as described in https://proposals.decred.org/record/ae7c4fe, and it starts a list of the OS packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,12 +45,13 @@ few simple options for obtaining the software:
 
 1. Download the the standalone DEX client for your operating system for the
    [latest release on GitHub](https://github.com/decred/dcrdex/releases).
-2. [Use Decrediton](https://docs.decred.org/wallets/decrediton/decrediton-setup/),
+2. Use your operating system's package manager. See [OS Packages](#os-packages)
+   for more info.
+3. [Use Decrediton](https://docs.decred.org/wallets/decrediton/decrediton-setup/),
    the official graphical Decred wallet, which integrates the DEX client, and go
    to the DEX tab.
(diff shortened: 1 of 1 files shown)


JoeGruffins (approved):
Does that arch build not need to build the stuff in `client/webserver/site`?

For mac maybe we want to add here? https://docs.brew.sh/Adding-Software-to-Homebrew

JoeGruffins:
I tried it out and I guess it doesn't need to build the site. The files downloaded at that tag have the `dist` folder checked in already is why?

chappjc:
> I tried it out and I guess it doesn't need to build the site. The files downloaded at that tag have the `dist` folder checked in already is why?

Right.  It's pretty slick what you can do with go embed, so we check in the `dist` folder on the release branches only.  On any system with go you can do `go run decred.org/dcrdex/client/cmd/dexc@v0.6.1` and it'll work.

If we were to make a `-git` kind of AUR package, we'd have to do the webpack build too.

Homebrew sounds like a good idea for Mac.
However, chat on matrix indicates that Chocolatey is probably pointless and that winget would be best.

buck54321 (approved):

PR #2355, opened 2023-05-15
client/btc: Live test to provide Wallet Type

dev-warrior777:
Like `ltc` client `firo` client wallet `NewWallet` constructor can create different Types of wallet based on the passed in `WalletConfig.Type`. When calling this constructor from `livetest.go` the `Type` field is an empty string.
This small change adds the desired wallet type to the WalletConfig sent to the wallet constructor by `livetest.go` instead of an empty string. This should be a non-breaking change.
This is nice to have but if rejected firo `NewWallet` constructor will default to creating a full node RPC wallet if it receives the null string, which is not so clean.

### DIFF
--- a/client/asset/btc/livetest/livetest.go
+++ b/client/asset/btc/livetest/livetest.go
@@ -62,6 +62,7 @@ func tBackend(ctx context.Context, t *testing.T, cfg *Config, walletName *Wallet
 	reportName := fmt.Sprintf("%s:%s-%s", cfg.Asset.Symbol, walletName.Node, walletName.Name)
 
 	walletCfg := &asset.WalletConfig{
+		Type:     walletName.WalletType,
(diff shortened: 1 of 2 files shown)


dev-warrior777:
On the other hand everyone else is using `""` as the `Type` selector.
https://github.com/decred/dcrdex/blob/c07ac7163339deb35def662ff44cca34c238f13f/client/asset/dcr/dcr.go#LL69C1-L72C1
Which may or may not be just using a default
https://github.com/decred/dcrdex/blob/c07ac7163339deb35def662ff44cca34c238f13f/client/asset/ltc/ltc.go#LL198C1-L201C1

dev-warrior777:
I reverted my own code to what the majority is doing. ;-)

Issue #2354, opened 2023-05-13
asset: Make spv wallet a FeeRater and use `NoLocalFeeRate` for spv wallets.


ukane-philemon:
> @chappjc from matrix: I have oracles enabled and instead of going with [mempool.space ](http://mempool.space/)'s fastest fee, it uses the stupid high fee from server here's a bug somewhere. Burned over a $100 on a basic send because of this. 

> We need to take the approach described in the comments(See [comment](https://github.com/decred/dcrdex/blob/master/client/asset/btc/btc.go#L1297-L1300) and [NoLocalFeeRate](https://github.com/decred/dcrdex/blob/master/client/asset/btc/btc.go#L1308-L1313)) (for SPV wallet use it as the local estimator function, and make the spv wallet a FeeRater). The other approach would be to make btc.(*baseWallet).Send work like Redeem, using targetFeeRateWithFallback instead of feeRateWithFallback but with Send, the feeRate input is not merely a suggestion.
Basically, the external fee rates only get used for SPV (non-FeeRater) wallets in a handful of internal paths, but it's not usable to the consumer.

PR #2356, opened 2023-05-15
asset: Implement `asset.FeeRater` for SPV wallets and minor refactor

ukane-philemon:
Closes: #2354

### DIFF
--- a/client/asset/bch/bch.go
+++ b/client/asset/bch/bch.go
@@ -212,15 +212,15 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network) 
 		// Bitcoin Cash don't take a change_type argument in their options
 		// unlike Bitcoin Core.
 		OmitAddressType: true,
-		// Bitcoin Cash uses estimatefee instead of estimatesmartfee, and even
-		// then, they modified it from the old Bitcoin Core estimatefee by
-		// removing the confirmation target argument.
-		FeeEstimator: estimateFee,
-		AssetID:      BipID,
+		AssetID:         BipID,
 	}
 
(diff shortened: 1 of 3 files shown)


chappjc (commented):
Looks good.  Will test it out, but looks right.

chappjc (approved):

client/asset/bch/bch.go
```diff
@@ -212,15 +212,15 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network)
...
-		AssetID:      BipID,
+		AssetID:         BipID,
 	}
 
 	switch cfg.Type {
 	case walletTypeRPC, walletTypeLegacy:
+		// Bitcoin Cash uses estimatefee instead of estimatesmartfee, and even
+		// then, they modified it from the old Bitcoin Core estimatefee by
+		// removing the confirmation target argument.
+		cloneCFG.FeeEstimator = estimateFee
```
chappjc, line 223: Nice catch moving this under here.  Looks like `LegacyBalance:      cfg.Type != walletTypeSPV,` above would be best done in this `case` as well, but it's not wrong presently, so nbd.

client/asset/btc/btc.go
```diff
@@ -983,9 +983,8 @@ func (btc *ExchangeWalletSPV) RemovePeer(addr string) error {
 var _ asset.FeeRater = (*ExchangeWalletFullNode)(nil)
 var _ asset.FeeRater = (*ExchangeWalletNoAuth)(nil)
```
chappjc, line 984: These two assertions are a bit pointless now, but they don't hurt either.

PR #2357, opened 2023-05-16
multi: Add wallet balance tooltip and all known wallet balances to translation map

### LINKED ISSUE #2177: ui: wallet balance breakdown tooltips
It would be nice to supply balance section in UI with tooltips:

image

example of tooltips text I have in mind (based on my current understanding):
- for **Available**: _this one is probably self explanatory, no tooltip needed_
- for **Locked**: funds temporarily locked to service your outstanding orders (pay for, and cover necessary fees)
- for **Immature**: funds coming your way shortly, waiting for confirmation in blockchain
- for **Bond Reserves (locked)**: _TODO (I haven't looked too deep in how exactly bonds work yet, suggestions ?)_

This feature is probably especially relevant after **Bond Reserves (locked)** section was introduced.

**UX consideration:** tooltip icon should be small enough, and/or pop-up on click (not on immediate hover), so that info-text doesn't jump in user's face when he doesn't need to see it.

ukane-philemon:
Closes #2177

UI changes:

Screenshot 2023-05-16 at 5 48 34 PM

Edit: IMO, reducing the tooltip width and adding border-radius makes the tooltip look slick.
Screenshot 2023-05-16 at 7 28 16 PM

### DIFF
--- a/client/asset/interface.go
+++ b/client/asset/interface.go
@@ -900,7 +900,7 @@ type Balance struct {
 	// Other is a place to list custom balance categories. It is recommended for
 	// custom balance added here to have a translation and tooltip info in
 	// client/webserver/site/src/js/wallet.js#customWalletBalanceCategory
-	Other map[string]CustomBalance `json:"other"`
+	Other map[BalanceCategory]CustomBalance `json:"other"`
 }
 
 // CustomBalance is a balance category used to track funds for a particular
@@ -915,16 +915,20 @@ type CustomBalance struct {
 	Locked bool `json:"locked"`
 }
 
-// These values are used as a map key for custom balances and may be recognized
-// in the frontend to support translation. It is recommended for custom balance
-// categories listed here to have a translation and tooltip info in
+// BalanceCategory is a string identifier for a custom balance category.
+type BalanceCategory string
+
+// Balance categories for custom balances These values are used as a map key for
+// custom balances and may be recognized in the frontend to support translation.
+// It is recommended for custom balance categories listed here to have a
+// translation and tooltip info in
 // client/webserver/site/src/js/wallet.js#customWalletBalanceCategory. If any of
 // these balance categories should change, the customWalletBalanceCategory
 // function in the wallet.js file above should be updated with the new value.
 const (
-	BondReserves    = "Bond Reserves"
-	ReservesDeficit = "Reserves Deficit"
-	Shielded        = "Shielded"
+	BalanceCategoryBondReserves    = "Bond Reserves"
+	BalanceCategoryReservesDeficit = "Reserves Deficit"
+	BalanceCategoryShielded        = "Shielded"
 )
 
 // Coin is some amount of spendable asset. Coin provides the information needed
--- a/client/asset/interface.go
+++ b/client/asset/interface.go
@@ -900,11 +900,18 @@ type Balance struct {
 	// Other is a place to list custom balance categories. It is recommended for
 	// custom balance added here to have a translation and tooltip info in
 	// client/webserver/site/src/js/wallet.js#customWalletBalanceCategory
-	Other map[string]*CustomBalance `json:"other"`
+	Other map[string]CustomBalance `json:"other"`
 }
 
+// CustomBalance is a balance category used to track funds for a particular
+// purpose or for a special kind of balance (e.g Zcash Shielded wallet
+// balances).
 type CustomBalance struct {
+	// Amount is the balance in the wallet for this custom category. It is is
+	// subtracted from Balance.Available above.
 	Amount uint64 `json:"amt"`
+	// Locked is a flag to indicate that this Amount is not included in
+	// Balance.Available and is included in Balance.Locked.
 	Locked bool   `json:"locked"`
 }
 
--- a/client/asset/interface.go
+++ b/client/asset/interface.go
@@ -897,8 +897,15 @@ type Balance struct {
 	// Locked is the total amount locked in the wallet which includes but
 	// is not limited to funds locked for swap but not actually swapped yet.
 	Locked uint64 `json:"locked"`
-	// Other is a place to list custom balance categories.
-	Other map[string]uint64 `json:"other"`
+	// Other is a place to list custom balance categories. It is recommended for
+	// custom balance added here to have a translation and tooltip info in
+	// client/webserver/site/src/js/wallet.js#customWalletBalanceCategory
+	Other map[string]*CustomBalance `json:"other"`
+}
+
+type CustomBalance struct {
+	Amount uint64 `json:"amt"`
+	Locked bool   `json:"locked"`
 }
 
 // Coin is some amount of spendable asset. Coin provides the information needed
--- a/client/asset/interface.go
+++ b/client/asset/interface.go
@@ -912,7 +912,7 @@ type CustomBalance struct {
 	Amount uint64 `json:"amt"`
 	// Locked is a flag to indicate that this Amount is not included in
 	// Balance.Available and is included in Balance.Locked.
-	Locked bool   `json:"locked"`
+	Locked bool `json:"locked"`
 }
 
 // Coin is some amount of spendable asset. Coin provides the information needed
--- a/client/asset/interface.go
+++ b/client/asset/interface.go
@@ -907,7 +907,7 @@ type Balance struct {
 // purpose or for a special kind of balance (e.g Zcash Shielded wallet
 // balances).
 type CustomBalance struct {
-	// Amount is the balance in the wallet for this custom category. It is is
+	// Amount is the balance in the wallet for this custom category. It is
 	// subtracted from Balance.Available above.
 	Amount uint64 `json:"amt"`
 	// Locked is a flag to indicate that this Amount is not included in
--- a/client/asset/interface.go
+++ b/client/asset/interface.go
@@ -897,6 +897,12 @@ type Balance struct {
 	// Locked is the total amount locked in the wallet which includes but
 	// is not limited to funds locked for swap but not actually swapped yet.
 	Locked uint64 `json:"locked"`
+	// BondReserves is the amount of funds locked in the wallet for expenses
+	// associated with bond maintenance.
+	BondReserves uint64 `json:"bondReserves"`
+	// ReservesDeficit is the difference between the available balance and the
+	// amount reserved for specific purposes.
+	ReservesDeficit uint64 `json:"reservesDeficit"`
 	// Other is a place to list custom balance categories. It is recommended for
 	// custom balance added here to have a translation and tooltip info in
 	// client/webserver/site/src/js/wallet.js#customWalletBalanceCategory
@@ -926,9 +932,7 @@ type BalanceCategory string
 // these balance categories should change, the customWalletBalanceCategory
 // function in the wallet.js file above should be updated with the new value.
 const (
-	BalanceCategoryBondReserves    = "Bond Reserves"
-	BalanceCategoryReservesDeficit = "Reserves Deficit"
-	BalanceCategoryShielded        = "Shielded"
+	BalanceCategoryShielded = "Shielded"
 )
 
 // Coin is some amount of spendable asset. Coin provides the information needed
--- a/client/asset/interface.go
+++ b/client/asset/interface.go
@@ -915,6 +915,18 @@ type CustomBalance struct {
 	Locked bool `json:"locked"`
 }
 
+// These values are used as a map key for custom balances and may be recognized
+// in the frontend to support translation. It is recommended for custom balance
+// categories listed here to have a translation and tooltip info in
+// client/webserver/site/src/js/wallet.js#customWalletBalanceCategory. If any of
+// these balance categories should change, the customWalletBalanceCategory
+// function in the wallet.js file above should be updated with the new value.
+const (
+	BondReserves    = "Bond Reserves"
+	ReservesDeficit = "Reserves Deficit"
+	Shielded        = "Shielded"
+)
+
 // Coin is some amount of spendable asset. Coin provides the information needed
 // to locate the unspent value on the blockchain.
 type Coin interface {
--- a/client/asset/zec/shielded_test.go
+++ b/client/asset/zec/shielded_test.go
@@ -90,7 +90,7 @@ func TestShieldUnshield(t *testing.T) {
 		t.Fatalf("Balance error: %v", err)
 	}
 
-	shieldedBal := bal.Other["Shielded"]
+	shieldedBal := bal.Other[asset.BalanceCategoryShielded]
 	if shieldedBal.Amount != status.Balance {
 		t.Fatalf("Account balance does not match sum of Orchard address balances. %d != %d", shieldedBal.Amount, status.Balance)
 	}
--- a/client/asset/zec/shielded_test.go
+++ b/client/asset/zec/shielded_test.go
@@ -90,8 +90,9 @@ func TestShieldUnshield(t *testing.T) {
 		t.Fatalf("Balance error: %v", err)
 	}
 
-	if bal.Other["shielded"] != status.Balance {
-		t.Fatalf("Account balance does not match sum of Orchard address balances. %d != %d", bal.Other["shielded"], status.Balance)
+	shieldedBal := bal.Other["Shielded"]
+	if shieldedBal.Amount != status.Balance {
+		t.Fatalf("Account balance does not match sum of Orchard address balances. %d != %d", shieldedBal.Amount, status.Balance)
 	}
 
 	_, err = w.UnshieldFunds(ctx, 5e7)
--- a/client/mm/price_oracle_test.go
+++ b/client/mm/price_oracle_test.go
@@ -53,7 +53,7 @@ func TestFetchBinanceSpread(t *testing.T) {
 
 func TestFetchCoinbaseSpread(t *testing.T) {
 	testSpreader(t, fetchCoinbaseSpread, "btc", "usd")
-}x
+}
 
 func TestFetchBittrexSpread(t *testing.T) {
 	testSpreader(t, fetchBittrexSpread, "dcr", "btc")
--- a/client/webserver/site/src/css/main.scss
+++ b/client/webserver/site/src/css/main.scss
@@ -481,7 +481,7 @@ div.form-closer {
   color: white;
   max-width: 300px;
   font-size: 15px;
-  border-radius: 7.5px;
+  border-radius: 4px;
 }
 
 #loader {
--- a/client/webserver/site/src/css/main.scss
+++ b/client/webserver/site/src/css/main.scss
@@ -479,8 +479,9 @@ div.form-closer {
   background-color: #040012;
   border: 1px solid #333;
   color: white;
-  max-width: 500px;
+  max-width: 300px;
   font-size: 15px;
+  border-radius: 7.5px;
 }
 
 #loader {
--- a/client/webserver/site/src/html/bodybuilder.tmpl
+++ b/client/webserver/site/src/html/bodybuilder.tmpl
@@ -18,7 +18,7 @@
       <span data-tmpl="text"></span>
     </span>
   </div>
-  <div id="tooltip" class="flex-center "></div>
+  <div id="tooltip" class="flex-center"></div>
   {{template "header" .}}
 {{end}}
 
--- a/client/webserver/site/src/html/bodybuilder.tmpl
+++ b/client/webserver/site/src/html/bodybuilder.tmpl
@@ -18,7 +18,7 @@
       <span data-tmpl="text"></span>
     </span>
   </div>
-  <div id="tooltip" class="flex-center"></div>
+  <div id="tooltip" class="flex-center "></div>
   {{template "header" .}}
 {{end}}
 
--- a/client/webserver/site/src/html/wallets.tmpl
+++ b/client/webserver/site/src/html/wallets.tmpl
@@ -46,7 +46,7 @@
                     <tbody id="balanceDetailBox">
                       <tr id="balanceDetailRow">
                         <td data-tmpl="category" class="grey">
-                         <span></span>
+                         <span data-tmpl="name"></span>
                          <span data-tmpl="tooltipMsg" data-tooltip="" class="ico-info d-hide"></span>
                         </td>
                         <td data-tmpl="subBalance"></td>
--- a/client/webserver/site/src/html/wallets.tmpl
+++ b/client/webserver/site/src/html/wallets.tmpl
@@ -45,7 +45,10 @@
                   <table id="balanceTable" class="fs15 mt-2 w-100">
                     <tbody id="balanceDetailBox">
                       <tr id="balanceDetailRow">
-                        <td data-tmpl="category" class="grey"></td>
+                        <td data-tmpl="category" class="grey">
+                         <span></span>
+                         <span data-tmpl="tooltipMsg" data-tooltip="" class="ico-info d-hide"></span>
+                        </td>
                         <td data-tmpl="subBalance"></td>
                       </tr>
                     </tbody>
--- a/client/webserver/site/src/js/locales.ts
+++ b/client/webserver/site/src/js/locales.ts
@@ -245,16 +245,16 @@ export const enUS: Locale = {
   [ID_IMMATURE_C]: 'Immature',
   [ID_SWAPPING]: 'Swapping',
   [ID_BONDED]: 'Bonded',
-  [ID_LOCKED_BAL_MSG]: 'Total funds temporarily locked to cover the costs of your bond maintenance, outstanding orders, and necessary fees',
-  [ID_IMMATURE_BAL_MSG]: 'Funds coming your way shortly, waiting for confirmation on the blockchain',
-  [ID_LOCKED_SWAPPING_BAL_MSG]: 'Funds currently locked in ongoing/active matches being executed at present',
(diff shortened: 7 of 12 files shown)


chappjc:
I wonder if everything with a tooltip really needs the "i" symbol, or we can just give the test a tooltip on hover.

ukane-philemon:
> I wonder if everything with a tooltip really needs the "i" symbol

I'd say yes. The i symbol help users notice there is a hidden info easily. I've notice in other applications, the tooltip is displayed on the i icon itself and not on surrounding text (this is debated since it's a matter of making it easy to hover on a small icon vs not pushing it in their face).

client/webserver/site/src/js/locales.ts
```diff
@@ -224,7 +239,22 @@ export const enUS: Locale = {
...
   [ID_REFUND_IMMINENT]: 'Will happen in the next few blocks',
-  [ID_REFUND_WILL_HAPPEN_AFTER]: 'Refund will happen after {{ refundAfterTime }}'
+  [ID_REFUND_WILL_HAPPEN_AFTER]: 'Refund will happen after {{ refundAfterTime }}',
+  [ID_AVAILABLE_C]: 'Available',
+  [ID_LOCKED_C]: 'Locked',
+  [ID_IMMATURE_C]: 'Immature',
+  [ID_SWAPPING]: 'Swapping',
+  [ID_BONDED]: 'Bonded',
+  [ID_LOCKED_BAL_MSG]: 'Total funds temporarily locked to cover the costs of your bond maintenance, outstanding orders, and necessary fees',
+  [ID_IMMATURE_BAL_MSG]: 'Funds coming your way shortly, waiting for confirmation on the blockchain',
```
chappjc, line 249: How about "Incoming funds awaiting confirmation".
"coming your way" is coloquial, "shortly" is subjective and mostly meaningless, and "on the blockchain" is superflous.

client/webserver/site/src/js/locales.ts
```diff
@@ -224,7 +239,22 @@ export const enUS: Locale = {
...
+  [ID_LOCKED_SWAPPING_BAL_MSG]: 'Funds currently locked in ongoing/active matches being executed at present',
```
chappjc, line 250: "currently" and "at present" and even "ongoing/active"  are repetitive.  I'd remove "at present" and possibly "ongoing/active".
"Funds currently locked in settling matches" would even convey the same info.

client/webserver/site/src/js/locales.ts
```diff
@@ -224,7 +239,22 @@ export const enUS: Locale = {
...
+  [ID_LOCKED_BOND_BAL_MSG]: 'Funds locked in active bonds',
+  [ID_RESERVES_DEFICIT]: 'Reserves Deficit',
+  [ID_RESERVES_DEFICIT_MSG]: 'This is the difference between the available balance in your wallet and the amount that has been reserved for specific purposes',
```
chappjc, line 253: This msg has sentence structure with "This is", unlike the others.

client/webserver/site/src/js/locales.ts
```diff
@@ -224,7 +239,22 @@ export const enUS: Locale = {
...
+  [ID_BOND_RESERVES]: 'Bond Reserves',
+  [ID_BOND_RESERVES_MSG]: 'Funds reserved to cover the expenses associated with bond maintenance',
+  [ID_SHIELDED]: 'Shielded',
+  [ID_SHIELDED_MSG]: 'Funds kept secure and private in a protected address'
```
chappjc, line 257: "in a protected address" might not be accurate.  Also I think it's an aggregation of shielded addresses in cases.

client/asset/interface.go
```diff
@@ -858,8 +858,15 @@ type Balance struct {
...
-	Other map[string]uint64 `json:"other"`
+	// Other is a place to list custom balance categories. It is recommended for
+	// custom balance added here to have a translation and tooltip info in
+	// client/webserver/site/src/js/wallet.js#customWalletBalanceCategory
+	Other map[string]*CustomBalance `json:"other"`
+}
+
+type CustomBalance struct {
+	Amount uint64 `json:"amt"`
+	Locked bool   `json:"locked"`
```
chappjc, line 869: Semantically, `Locked` is a flag to indicate that this `Amount` is not included in (subtracted from) the wallet's `Available` and included in `Locked`.  Please doc on this struct and it's fields so devs know  how to use this type and the parent `Balance` type.

client/asset/interface.go
```diff
@@ -858,8 +858,15 @@ type Balance struct {
 	// Locked is the total amount locked in the wallet which includes but
 	// is not limited to funds locked for swap but not actually swapped yet.
 	Locked uint64 `json:"locked"`
-	// Other is a place to list custom balance categories.
-	Other map[string]uint64 `json:"other"`
+	// Other is a place to list custom balance categories. It is recommended for
+	// custom balance added here to have a translation and tooltip info in
+	// client/webserver/site/src/js/wallet.js#customWalletBalanceCategory
+	Other map[string]*CustomBalance `json:"other"`
```
chappjc, line 864: This is a very small struct, best to store it in the map by value IMO.

client/webserver/site/src/html/bodybuilder.tmpl
```diff
@@ -18,7 +18,7 @@
       <span data-tmpl="text"></span>
     </span>
   </div>
-  <div id="tooltip" class="flex-center"></div>
+  <div id="tooltip" class="flex-center "></div>
```
chappjc, line 21: What's this space for?

ukane-philemon: Thanks for the review @chappjc, the space is intended. I remember wanting to add a class there but had to use the existing `tooltip` id.

client/webserver/site/src/js/wallets.ts
```diff
@@ -1316,3 +1330,20 @@ function assetIsConfigurable (assetID: number) {
...
+ * customWalletBalance returns the translated string and a message for the
+ * provided balance category.
+ */
+function customWalletBalanceCategory (category: string): [string, string] {
+  switch (category) {
+    case 'Reserves Deficit':
+      return [intl.prep(intl.ID_RESERVES_DEFICIT), intl.prep(intl.ID_RESERVES_DEFICIT_MSG)]
+    case 'Bond Reserves':
+      return [intl.prep(intl.ID_BOND_RESERVES), intl.prep(intl.ID_BOND_RESERVES_MSG)]
+    case 'Shielded':
```
chappjc, line 1344: If we are going to match on these strings, they should be made canonical in client/asset/interface.go with docs there so the coder knows that they are enumerated in one place since they may be recognized on the frontend and if they are ever changed, the consumer code needs to change too.

client/webserver/site/src/js/locales.ts
```diff
@@ -224,7 +239,22 @@ export const enUS: Locale = {
...
-  [ID_REFUND_WILL_HAPPEN_AFTER]: 'Refund will happen after {{ refundAfterTime }}'
+  [ID_REFUND_WILL_HAPPEN_AFTER]: 'Refund will happen after {{ refundAfterTime }}',
+  [ID_AVAILABLE_C]: 'Available',
+  [ID_LOCKED_C]: 'Locked',
+  [ID_IMMATURE_C]: 'Immature',
+  [ID_SWAPPING]: 'Swapping',
+  [ID_BONDED]: 'Bonded',
+  [ID_LOCKED_BAL_MSG]: 'Total funds temporarily locked to cover the costs of your bond maintenance, settling matches, and necessary fees',
+  [ID_IMMATURE_BAL_MSG]: 'Incoming funds awaiting confirmation',
+  [ID_LOCKED_SWAPPING_BAL_MSG]: 'Funds currently locked settling matches',
```
chappjc, line 250: "Funds currently locked **in** settling matches"

client/asset/interface.go
```diff
@@ -897,10 +897,36 @@ type Balance struct {
...
+// in the frontend to support translation. It is recommended for custom balance
+// categories listed here to have a translation and tooltip info in
+// client/webserver/site/src/js/wallet.js#customWalletBalanceCategory. If any of
+// these balance categories should change, the customWalletBalanceCategory
+// function in the wallet.js file above should be updated with the new value.
+const (
+	BondReserves    = "Bond Reserves"
+	ReservesDeficit = "Reserves Deficit"
+	Shielded        = "Shielded"
+)
```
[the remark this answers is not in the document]
ukane-philemon: > I'm curious what you found that made you revert? It generally made sense to me.

Oh, nothing really. Maybe I was just avoiding more changes  😆 , but can add it.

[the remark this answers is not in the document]
chappjc: Thanks.  I find that kind of thing very helpful so you can quickly enumerate all acceptable/known values.  The IDE and language server (gopls) are great at completion when the known values by type are well-defined.

client/asset/interface.go
```diff
@@ -897,10 +897,40 @@ type Balance struct {
...
+// Balance categories for custom balances These values are used as a map key for
+// custom balances and may be recognized in the frontend to support translation.
+// It is recommended for custom balance categories listed here to have a
+// translation and tooltip info in
+// client/webserver/site/src/js/wallet.js#customWalletBalanceCategory. If any of
+// these balance categories should change, the customWalletBalanceCategory
+// function in the wallet.js file above should be updated with the new value.
+const (
+	BalanceCategoryBondReserves    = "Bond Reserves"
+	BalanceCategoryReservesDeficit = "Reserves Deficit"
```
buck54321, line 930: IMO these two should be fields of `asset.Balance` because these can be non-zero for (eventually) any asset, unlike shielded which is very specific to Zcash.

client/asset/zec/shielded_test.go
```diff
@@ -90,8 +90,9 @@ func TestShieldUnshield(t *testing.T) {
 		t.Fatalf("Balance error: %v", err)
 	}
 
-	if bal.Other["shielded"] != status.Balance {
-		t.Fatalf("Account balance does not match sum of Orchard address balances. %d != %d", bal.Other["shielded"], status.Balance)
+	shieldedBal := bal.Other["Shielded"]
```
buck54321, line 93: Use the `asset.BalanceCategoryShielded` const?

client/webserver/site/src/css/main.scss
```diff
@@ -479,8 +479,9 @@ div.form-closer {
   background-color: #040012;
   border: 1px solid #333;
   color: white;
-  max-width: 500px;
+  max-width: 300px;
   font-size: 15px;
+  border-radius: 7.5px;
```
buck54321, line 484: This is larger than any border radius we're using anywhere else (except where we're using it to make a circle). How about `3px`?

client/webserver/site/src/js/locales.ts
```diff
@@ -111,6 +111,21 @@ export const ID_WALLET_SYNC_FINISHING_UP = 'WALLET_SYNC_FINISHING_UP'
...
+export const ID_LOCKED_BAL_MSG = 'ID_LOCKED_BAL_MSG'
+export const ID_IMMATURE_BAL_MSG = 'ID_IMMATURE_BAL_MSG'
+export const ID_LOCKED_SWAPPING_BAL_MSG = 'ID_LOCKED_SWAPPING_BAL_MSG'
+export const ID_LOCKED_BOND_BAL_MSG = 'ID_LOCKED_BOND_BAL_MSG'
+export const ID_RESERVES_DEFICIT = 'ID_RESERVES_DEFICIT'
+export const ID_RESERVES_DEFICIT_MSG = 'ID_RESERVES_DEFICIT_MSG'
+export const ID_BOND_RESERVES = 'ID_BOND_RESERVES'
+export const ID_BOND_RESERVES_MSG = 'ID_BOND_RESERVES_MSG'
+export const ID_SHIELDED = 'ID_SHIELDED'
+export const ID_SHIELDED_MSG = 'ID_SHIELDED_MSG'
```
buck54321, line 128: Remove the `ID_` prefix from the values.

client/webserver/site/src/js/locales.ts
```diff
@@ -111,6 +111,21 @@ export const ID_WALLET_SYNC_FINISHING_UP = 'WALLET_SYNC_FINISHING_UP'
 export const ID_CONNECT_WALLET_ERR_MSG = 'CONNECTING_WALLET_ERR_MSG'
 export const ID_REFUND_IMMINENT = 'REFUND_IMMINENT'
 export const ID_REFUND_WILL_HAPPEN_AFTER = 'REFUND_WILL_HAPPEN_AFTER'
+export const ID_AVAILABLE_C = 'ID_AVAILABLE_C'
+export const ID_LOCKED_C = 'ID_LOCKED_C'
+export const ID_IMMATURE_C = 'ID_IMMATURE_C'
```
buck54321, line 116: What is the `_C` suffix? Can you choose something more informative?

client/webserver/site/src/html/wallets.tmpl
```diff
@@ -45,7 +45,10 @@
                   <table id="balanceTable" class="fs15 mt-2 w-100">
                     <tbody id="balanceDetailBox">
                       <tr id="balanceDetailRow">
-                        <td data-tmpl="category" class="grey"></td>
+                        <td data-tmpl="category" class="grey">
+                         <span></span>
```
buck54321, line 49: Add a `data-tmpl` attribute so we don't have to do the `firstElementChild` thing.

client/webserver/site/src/js/wallets.ts
```diff
@@ -815,26 +816,39 @@ export default class WalletsPage extends BasePage {
...
       const row = page.balanceDetailRow.cloneNode(true) as PageElement
       if (firstOther) {
         row.classList.add('first-other')
         firstOther = false
       }
       page.balanceDetailBox.appendChild(row)
       const tmpl = Doc.parseTemplate(row)
-      tmpl.category.textContent = category
+      if (tmpl.category.firstElementChild) tmpl.category.firstElementChild.textContent = category
+      else tmpl.category.textContent = category
```
buck54321, line 828: In what case would `tmpl.category` not have a child element?

client/webserver/site/src/js/wallets.ts
```diff
@@ -815,26 +816,39 @@ export default class WalletsPage extends BasePage {
...
       tmpl.subBalance.textContent = Doc.formatCoinValue(subBalance, ui)
     }
-    addSubBalance('Available', bal.available)
-    addSubBalance('Locked', totalLocked)
-    addSubBalance('Immature', bal.immature)
-    const sortedCats = Object.entries(bal.other || {})
-    sortedCats.sort((a: [string, number], b: [string, number]): number => a[0].localeCompare(b[0]))
+    addSubBalance(intl.prep(intl.ID_AVAILABLE_C), bal.available, '')
+    addSubBalance(intl.prep(intl.ID_LOCKED_C), totalLocked, intl.prep(intl.ID_LOCKED_BAL_MSG))
+    addSubBalance(intl.prep(intl.ID_IMMATURE_C), bal.immature, intl.prep(intl.ID_IMMATURE_BAL_MSG))
```
buck54321, line 837: Good work getting rid of these hard-coded strings.

client/webserver/site/src/js/wallets.ts
```diff
@@ -815,26 +816,39 @@ export default class WalletsPage extends BasePage {
...
+      return category + ' (' + intl.prep(intl.ID_LOCKED) + ') '
+    }
+    if (bal.contractlocked > 0) addSubBalance(lockedBal(intl.prep(intl.ID_SWAPPING)), bal.contractlocked, intl.prep(intl.ID_LOCKED_SWAPPING_BAL_MSG))
+    if (bal.bondlocked > 0) addSubBalance(lockedBal(intl.prep(intl.ID_BONDED)), bal.bondlocked, intl.prep(intl.ID_LOCKED_BOND_BAL_MSG))
+    for (const [cat, bal] of sortedBalCats) {
+      let [balCategory, tooltipMsg] = customWalletBalanceCategory(cat)
+      if (bal.locked) balCategory = lockedBal(balCategory)
+      addSubBalance(balCategory, bal.amt, tooltipMsg)
+    }
+    app().bindTooltips(page.balanceDetailBox)
```
buck54321, line 851: This is where it seems obvious to me that we have an inconsistency in the API. If `bondLocked` is a field of the `Balance`, why would bond reserves and bond deficit be part of the `other` map?

client/webserver/site/src/js/wallets.ts
```diff
@@ -815,26 +816,39 @@ export default class WalletsPage extends BasePage {
...
     firstOther = true
-    if (bal.contractlocked > 0) addSubBalance('Swapping (locked)', bal.contractlocked)
-    if (bal.bondlocked > 0) addSubBalance('Bonded (locked)', bal.bondlocked)
-    for (const [cat, sub] of sortedCats) addSubBalance(cat, sub)
+    const lockedBal = (category: string): string => {
+      return category + ' (' + intl.prep(intl.ID_LOCKED) + ') '
+    }
+    if (bal.contractlocked > 0) addSubBalance(lockedBal(intl.prep(intl.ID_SWAPPING)), bal.contractlocked, intl.prep(intl.ID_LOCKED_SWAPPING_BAL_MSG))
+    if (bal.bondlocked > 0) addSubBalance(lockedBal(intl.prep(intl.ID_BONDED)), bal.bondlocked, intl.prep(intl.ID_LOCKED_BOND_BAL_MSG))
+    for (const [cat, bal] of sortedBalCats) {
```
buck54321, line 846: We need to display `orderLocked` as well. Not sure why we weren't already doing that. 

I also think we should reorganize the display so that the locked sub-balances show as subcategories of the locked total, e.g. roughly
image
But this can be done in a followup.

buck54321 (changes requested):

client/webserver/site/src/css/main.scss
ukane-philemon: > This is larger than any border radius we're using anywhere else

Yup, it is. I added an image in my first comment to show what it'd look like.

> How about 3px?

Ermm, if that's what we want then np.

client/webserver/site/src/js/locales.ts
ukane-philemon: > What is the _C suffix? Can you choose something more informative?

`Capitalized`. Maybe I should use the full term?

client/asset/interface.go
chappjc: As per my other comment, the idea is we don't add these to the map when they are zero because they are very often zero and it's going to be a waste of space on frontend.  So if we go this route and put them in asset.Balance, frontend just has to know to omit these from the table display if they are zero (IMO), because we don't want them displayed for every asset.

buck54321: > frontend just has to know to omit these from the table display if they are zero

This is what I figured we would do. Makes more sense as fields of the struct, since the front end doesn't need to guess whether they need to be displayed or not. The `Other` field is for custom balances categories that are unique to a particular asset, e.g. shielded, staked. I would not expect the front end to try to reason about whether to show zero balances from `Other`. If the asset chooses to include the entry, it should be displayed.

client/webserver/site/src/js/wallets.ts
ukane-philemon: > I also think we should reorganize the display so that the locked sub-balances show as subcategories of the locked total, e.g. roughly

Okay.

client/asset/interface.go
chappjc: Makes sense to me.

client/webserver/site/src/css/main.scss
buck54321: We use 3 or 4 px for all of our border radii. The few places you see larger values are when we are creating circles.

client/webserver/site/src/js/locales.ts
buck54321: It just wasn't obvious. If you don't want to write out `_CAPITALIZED`, you can use `_TITLE`

client/webserver/site/src/css/main.scss
ukane-philemon: I'd use 4 px.

client/asset/interface.go
ukane-philemon: Updated.

client/webserver/site/src/css/main.scss
ukane-philemon: Updated.

client/mm/price_oracle_test.go
```diff
@@ -53,7 +53,7 @@ func TestFetchBinanceSpread(t *testing.T) {
 
 func TestFetchCoinbaseSpread(t *testing.T) {
 	testSpreader(t, fetchCoinbaseSpread, "btc", "usd")
-}x
```
buck54321, line 56: weird

client/webserver/site/src/js/locales.ts
```diff
@@ -224,7 +241,24 @@ export const enUS: Locale = {
...
   [ID_TAKER_FOUND_MAKER_REDEMPTION]: 'Redeemed by {{ makerAddr }}',
   [ID_REFUND_IMMINENT]: 'Will happen in the next few blocks',
-  [ID_REFUND_WILL_HAPPEN_AFTER]: 'Refund will happen after {{ refundAfterTime }}'
+  [ID_REFUND_WILL_HAPPEN_AFTER]: 'Refund will happen after {{ refundAfterTime }}',
+  [ID_AVAILABLE_TITLE]: 'Available',
+  [ID_LOCKED_TITLE]: 'Locked',
+  [ID_IMMATURE_TITLE]: 'Immature',
+  [ID_SWAPPING]: 'Swapping',
+  [ID_BONDED]: 'Bonded',
+  [ID_LOCKED_BAL_MSG]: 'Total funds temporarily locked to cover the costs of your bond maintenance, settling matches, and necessary fees',
```
buck54321, line 250: ~~settling matches, and necessary fees~~ -> live orders and matches

Also can be locked for other reasons though. For Decred, we add value locked in tickets too.

client/webserver/site/src/js/locales.ts
```diff
@@ -224,7 +241,24 @@ export const enUS: Locale = {
...
+  [ID_IMMATURE_BAL_MSG]: 'Incoming funds awaiting confirmation',
```
buck54321, line 251: Still a little off. This is really only coinbase and stakebase outputs that require a large number of confirmations by consensus. Redeemed funds never end up here.

client/webserver/site/src/js/locales.ts
```diff
@@ -224,7 +241,24 @@ export const enUS: Locale = {
...
+  [ID_LOCKED_SWAPPING_BAL_MSG]: 'Funds currently locked in settling matches',
+  [ID_LOCKED_BOND_BAL_MSG]: 'Funds locked in active bonds',
+  [ID_RESERVES_DEFICIT]: 'Reserves Deficit',
+  [ID_RESERVES_DEFICIT_MSG]: 'Difference between the available balance in your wallet and the amount that has been reserved for specific purposes',
```
buck54321, line 255: The apparent wallet balance shortcoming to maintain bonding level. If this persists, you may need to add funds to stay fully bonded.

client/webserver/site/src/js/locales.ts
```diff
@@ -224,7 +241,24 @@ export const enUS: Locale = {
...
+  [ID_BOND_RESERVES]: 'Bond Reserves',
+  [ID_BOND_RESERVES_MSG]: 'Funds reserved to cover the expenses associated with bond maintenance',
+  [ID_SHIELDED]: 'Shielded',
+  [ID_SHIELDED_MSG]: 'Total funds kept private',
```
buck54321, line 259: ~~private~~ -> shielded

client/webserver/site/src/js/locales.ts
```diff
@@ -224,7 +241,24 @@ export const enUS: Locale = {
...
+  [ID_ORDER]: 'Order',
+  [ID_LOCKED_ORDER_BAL_MSG]: 'Total amount of funds currently locked for swap, but not actually swapped yet.'
```
buck54321, line 261: Funds locked in unmatched orders.

buck54321 (changes requested):
Looks and tests well. Just some suggestions for tooltip language.

client/webserver/site/src/js/locales.ts
ukane-philemon: Any other specific message in mind?

client/webserver/site/src/js/locales.ts
ukane-philemon: I have updated the message with your suggestion in commit https://github.com/decred/dcrdex/pull/2357/commits/27d2601a3a40358fbc1aee915bd9c83a979cd8e5. But is `Reserves Deficit` only for `Fidelity Bonds`?

client/webserver/site/src/js/locales.ts
ukane-philemon: Updated.

client/webserver/site/src/js/locales.ts
buck54321: Good enough for now. I may follow up with a change later.

JoeGruffins (approved):
Working well on simnet.

client/webserver/site/src/js/wallets.ts
ukane-philemon: > I also think we should reorganize the display so that the locked sub-balances show as subcategories of the locked total, e.g. roughly

@buck54321, From the comment for these balance fields, It seems they are not included in `asset.Balance.Locked` and might be misleading to include them as a subcategory of the locked total, or I'm I missing something?


https://github.com/decred/dcrdex/blob/4a2ba0ff4606c81b7c4918c84ef8927756eeea10/client/core/types.go#L111-L118

PR #2358, opened 2023-05-17
dcr: update harness for new pow

chappjc:


dex/testing/dcr/create-wallet.sh
```diff
@@ -60,8 +60,8 @@ if [ "${ENABLE_VOTING}" = "2" ]; then
 cat >> "${WALLET_DIR}/${NAME}.conf" <<EOF
 enablevoting=1
 enableticketbuyer=1
-ticketbuyer.limit=6
-ticketbuyer.balancetomaintainabsolute=1000
+ticketbuyer.limit=7
```
davecgh, line 63: Did you ultimately need to bump this?  I expect the real problem was the combination of the `ticketbuyer.balancetomaintainabsolute` line as well as sending all the coins to `beta` too early.

chappjc: 6 was still fine. The real problem was a misplaced sendtoaddress that was spending 1000 DCR after block 100, but this `ticketbuyer.balancetomaintainabsolute` didn't help either.

JoeGruffins (approved):

buck54321 (approved):

PR #2362, opened 2023-05-20
client/{core, btc}: MultiTrade

martonp:
This adds a `core.MultiTrade` function. It allows callers to fund multiple trades in one go. It is only implemented for BTC currently, but will be implemented for other assets later, after this is reviewed. In order to fund multiple orders in one go, the BTC wallet first attempts to fund each of the orders with the existing UTXOs. If this was not possible and splitting is not allowed, the orders that are able to be funded are returned. If splitting is allowed, a split transaction with one output per order will be created.

This also updates `FundOrder` to return the fees spent on any split transactions, and this is stored in the database.

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -761,6 +761,11 @@ func (s *swapOptions) feeBump() (float64, error) {
 	return bump, nil
 }
 
+type fundMultiOptions struct {
+	Split       *bool   `ini:"swapsplit"`
+	SplitBuffer *uint64 `ini:"splitbuffer"`
+}
+
 // redeemOptions are order options that apply to redemptions.
 type redeemOptions struct {
 	FeeBump *float64 `ini:"redeemfeebump"`
@@ -915,6 +920,7 @@ var _ asset.Bonder = (*baseWallet)(nil)
 var _ asset.Authenticator = (*ExchangeWalletSPV)(nil)
 var _ asset.Authenticator = (*ExchangeWalletFullNode)(nil)
 var _ asset.Authenticator = (*ExchangeWalletAccelerator)(nil)
+var _ asset.MultiOrderFunder = (*baseWallet)(nil)
 
 // RecoveryCfg is the information that is transferred from the old wallet
 // to the new one when the wallet is recovered.
@@ -2087,7 +2093,10 @@ func (btc *baseWallet) estimateSwap(lots, lotSize, feeSuggestion, maxFeeRate uin
 			return nil, false, 0, errors.New("balance too low to both fund order and maintain bond reserves")
 		}
 
-		kept := leastOverFund(reserves, utxos)
+		enough := func(_, sum uint64) (bool, uint64) {
+			return sum >= reserves, 0
+		}
+		kept := leastOverFund(enough, utxos)
 		utxos := utxoSetDiff(utxos, kept)
 		sum, _, inputsSize, _, _, _, _, err = tryFund(utxos, orderEnough(val, lots, bumpedMaxRate, btc.initTxSizeBase, btc.initTxSize, btc.segwit, false))
 		if err != nil {
@@ -2200,26 +2209,26 @@ func (btc *baseWallet) SingleLotRedeemFees(_ uint32, feeSuggestion uint64, optio
 // Equal number of coins and redeemed scripts must be returned. A nil or empty
 // dex.Bytes should be appended to the redeem scripts collection for coins with
 // no redeem script.
-func (btc *baseWallet) FundOrder(ord *asset.Order) (asset.Coins, []dex.Bytes, error) {
+func (btc *baseWallet) FundOrder(ord *asset.Order) (asset.Coins, []dex.Bytes, uint64, error) {
 	ordValStr := amount(ord.Value).String()
 	btc.log.Debugf("Attempting to fund order for %s %s, maxFeeRate = %d, max swaps = %d",
 		ordValStr, btc.symbol, ord.MaxFeeRate, ord.MaxSwapCount)
 
 	if ord.Value == 0 {
-		return nil, nil, fmt.Errorf("cannot fund value = 0")
+		return nil, nil, 0, fmt.Errorf("cannot fund value = 0")
 	}
 	if ord.MaxSwapCount == 0 {
-		return nil, nil, fmt.Errorf("cannot fund a zero-lot order")
+		return nil, nil, 0, fmt.Errorf("cannot fund a zero-lot order")
 	}
 	if ord.FeeSuggestion > ord.MaxFeeRate {
-		return nil, nil, fmt.Errorf("fee suggestion %d > max fee rate %d", ord.FeeSuggestion, ord.MaxFeeRate)
+		return nil, nil, 0, fmt.Errorf("fee suggestion %d > max fee rate %d", ord.FeeSuggestion, ord.MaxFeeRate)
 	}
 	if ord.FeeSuggestion > btc.feeRateLimit() {
-		return nil, nil, fmt.Errorf("suggested fee > configured limit. %d > %d", ord.FeeSuggestion, btc.feeRateLimit())
+		return nil, nil, 0, fmt.Errorf("suggested fee > configured limit. %d > %d", ord.FeeSuggestion, btc.feeRateLimit())
 	}
 	// Check wallets fee rate limit against server's max fee rate
 	if btc.feeRateLimit() < ord.MaxFeeRate {
-		return nil, nil, fmt.Errorf(
+		return nil, nil, 0, fmt.Errorf(
 			"%v: server's max fee rate %v higher than configued fee rate limit %v",
 			dex.BipIDSymbol(BipID), ord.MaxFeeRate, btc.feeRateLimit())
 	}
@@ -2227,7 +2236,7 @@ func (btc *baseWallet) FundOrder(ord *asset.Order) (asset.Coins, []dex.Bytes, er
 	customCfg := new(swapOptions)
 	err := config.Unmapify(ord.Options, customCfg)
 	if err != nil {
-		return nil, nil, fmt.Errorf("error parsing swap options: %w", err)
+		return nil, nil, 0, fmt.Errorf("error parsing swap options: %w", err)
 	}
 
 	bumpedMaxRate, err := calcBumpedRate(ord.MaxFeeRate, customCfg.FeeBump)
@@ -2257,7 +2266,7 @@ func (btc *baseWallet) FundOrder(ord *asset.Order) (asset.Coins, []dex.Bytes, er
 			extraSplitOutput = reserves + bondsFeeBuffer(btc.segwit, btc.feeRateLimit())
 		}
 		if err != nil {
-			return nil, nil, fmt.Errorf("error funding swap value of %s: %w", amount(ord.Value), err)
+			return nil, nil, 0, fmt.Errorf("error funding swap value of %s: %w", amount(ord.Value), err)
 		}
 	}
 
@@ -2282,24 +2291,24 @@ func (btc *baseWallet) FundOrder(ord *asset.Order) (asset.Coins, []dex.Bytes, er
 			btc.log.Errorf("calcBumpRate error: %v", err)
 		}
 
-		splitCoins, split, err := btc.split(ord.Value, ord.MaxSwapCount, spents,
+		splitCoins, split, splitFees, err := btc.split(ord.Value, ord.MaxSwapCount, spents,
 			inputsSize, fundingCoins, splitFeeRate, bumpedMaxRate, extraSplitOutput)
 		if err != nil {
 			if err := btc.ReturnCoins(coins); err != nil {
 				btc.log.Errorf("Error returning coins: %v", err)
 			}
-			return nil, nil, err
+			return nil, nil, 0, err
 		} else if split {
 			fmt.Printf("original coins: %s, split coins %s\n", coins, splitCoins)
-			return splitCoins, []dex.Bytes{nil}, nil // no redeem script required for split tx output
+			return splitCoins, []dex.Bytes{nil}, splitFees, nil // no redeem script required for split tx output
 		}
-		return coins, redeemScripts, nil // splitCoins == coins
+		return coins, redeemScripts, 0, nil // splitCoins == coins
 	}
 
 	btc.log.Infof("Funding %s %s order with coins %v worth %s",
 		ordValStr, btc.symbol, coins, amount(sum))
 
-	return coins, redeemScripts, nil
+	return coins, redeemScripts, 0, nil
 }
 
 func (btc *baseWallet) fundInternal(keep uint64, minConfs uint32, lockUnspents bool,
@@ -2311,7 +2320,10 @@ func (btc *baseWallet) fundInternal(keep uint64, minConfs uint32, lockUnspents b
 	}
 
 	if keep > 0 {
-		kept := leastOverFund(keep, utxos)
+		enough := func(_, sum uint64) (bool, uint64) {
+			return sum >= keep, 0
+		}
+		kept := leastOverFund(enough, utxos)
 		btc.log.Debugf("Setting aside %v BTC in %d UTXOs to respect the %v BTC reserved amount",
 			toBTC(sumUTXOs(kept)), len(kept), toBTC(keep))
 		utxosPruned := utxoSetDiff(utxos, kept)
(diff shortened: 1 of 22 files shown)


client/core/core.go
```diff
@@ -5959,13 +5892,245 @@ func (c *Core) prepareTradeRequest(pw []byte, form *TradeForm) (*tradeRequest, e
...
+
+	redeemAddresses := make([]string, 0, len(form.Placements))
+	for range form.Placements {
+		redeemAddr, err := toWallet.RedemptionAddress()
+		if err != nil {
+			return nil, codedError(walletErr, fmt.Errorf("%s RedemptionAddress error: %w",
+				assetConfigs.toAsset.Symbol, err))
+		}
+		redeemAddresses = append(redeemAddresses, redeemAddr)
+	}
```
buck54321, line 6048: If we're replacing orders en masse when the market moves, we could potentially be generating a lot of unused redemption addresses. I wonder if we need to consider the gap limit.

buck54321 (changes requested):

client/asset/btc/btc.go
```diff
@@ -2365,6 +2377,209 @@ func (btc *baseWallet) fund(keep uint64, minConfs uint32, lockUnspents bool,
...
+	}
+	if len(coins) == len(values) || !allowSplit {
+		for pt, fc := range fundingCoins {
+			btc.fundingCoins[pt] = fc
+		}
+		btc.node.lockUnspent(false, spents)
+		return coins, redeemScripts, 0, nil
+	}
+
+	return btc.fundMultiWithSplit(keep+reserves, values, splitTxFeeRate, maxFeeRate, splitBuffer)
```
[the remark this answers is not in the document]
chappjc: > Something feels backwards. If I want a split, and you don't split, that would be wrong.

I believe this has been the behavior since the beginning, at least when it is decided the split tx adds more "baggage" than the split would save by avoiding overlock.  Perhaps the context is different here?  Is the bool even "more optional" in this case?

client/core/core.go
chappjc: I also think gap limit is a real concern, particularly with BTC.  Since the gap limit was already becoming an issue before this, we recenly made it so DCR now always [uses the wrap gap policy](https://github.com/decred/dcrdex/commit/32607367cef05aef22cb57d3fc524c7894b046ed) so we don't make big gaps.  With BTC I'm not sure what we did or can do.  I think there's a 1000 address limit with the full node wallet and 250 with native, but we should research that.

Do you think we need to make some kind of address pool for BTC to alleviate this?

client/asset/btc/btc.go
```diff
@@ -2365,6 +2377,209 @@ func (btc *baseWallet) fund(keep uint64, minConfs uint32, lockUnspents bool,
...
+	}
+
+	var totalOutputRequired uint64
+	requiredForOrders := make([]uint64, len(values))
+	for i, value := range values {
+		req := calc.RequiredOrderFundsAlt(value.Value, swapInputSize, value.MaxSwapCount, btc.initTxSizeBase, btc.initTxSize, maxFeeRate)
+		req = req * (100 + splitBuffer) / 100
+		requiredForOrders[i] = req
+		totalOutputRequired += req
+	}
```
buck54321, line 2483: Would it be useful to check existing utxos to see if we already have one that satisfies our needs here? Bots will likely cancel and replace a lot of fully unmatched orders, If these orders change by a rate that keeps the old funding coins withing the split buffer, we can reuse the utxos. If we can maximize our reuse of those outputs and minimize the number of inputs that go into the split tx (by reducing the total funding needed), we could potentially see significant fee savings.

client/asset/interface.go
```diff
@@ -1019,3 +1026,48 @@ type Order struct {
...
+	// funding for the quote asset's wallet, the number of lots will not be
+	// Value / DEXConfig.LotSize, because an order is quantified in the base
+	// asset, so lots is always (order quantity) / (base asset lot size).
+	MaxSwapCount uint64 // uint64 for compatibility with quantity and lot size.
+}
+
+// MultiOrder is order details needed for FundMultiOrder.
+type MultiOrder struct {
+	// Version is the asset version of the "from" asset with the init
+	// transaction (this wallet). Most backends only support one version.
```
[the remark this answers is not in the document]
martonp: I agree the comment is a bit strange... I just copied it from the `Order` type. Should we just change `Version` to `SwapVersion`? The `Order` type also has a `RedeemVersion`. We have `fromAsset` and `toAsset` on `assetSet` and `fromWallet` and `toWallet` on `walletSet` btw.

client/asset/btc/btc.go
```diff
@@ -761,6 +761,11 @@ func (s *swapOptions) feeBump() (float64, error) {
 	return bump, nil
 }
 
+type fundMultiOptions struct {
+	Split       *bool   `ini:"swapsplit"`
+	SplitBuffer *uint64 `ini:"splitbuffer"`
```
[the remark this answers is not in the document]
martonp: I don't have this implemented for the regular split transactions, only for these MultiTrade splits. I wasn't planning on exposing the MultiTrade functionality through the UI. Why would you use the buffer with a regular trade?

client/asset/btc/btc.go
```diff
@@ -2365,6 +2377,209 @@ func (btc *baseWallet) fund(keep uint64, minConfs uint32, lockUnspents bool,
 	return btc.fundInternal(keep, minConfs, lockUnspents, enough)
 }
 
+// fundMultiBestEffors makes a best effort to fund every order. If it is not
+// possible, it returns coins for the orders that could be funded. The coins
+// that fund each order are returned in the same order as the values that were
+// passed in.
```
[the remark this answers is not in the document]
martonp: They're not really requesting more than is available, it's just that without a split the wallet would not be able to fund all of the orders without going over the `keep` amount.

client/asset/btc/btc.go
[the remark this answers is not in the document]
martonp: The `MultiTrade` will never lock more the balance.Available - keep` funds. The bot will pass `keep` as the amount that has been reserved by bots trading on other markets. If splitting is not allowed, then `MultiTrade` will only fund   the orders that it is able to without dipping into the other bots' allocation.

client/asset/btc/btc.go
martonp: `fundMultiBestEffort` handles this. It tries to fund all of the orders with existing UTXOs. If it is not able to, only then is the split done.

client/asset/btc/btc.go
```diff
@@ -2311,7 +2317,10 @@ func (btc *baseWallet) fundInternal(keep uint64, minConfs uint32, lockUnspents b
 	}
 
 	if keep > 0 {
-		kept := leastOverFund(keep, utxos)
+		enough := func(_, sum uint64) (bool, uint64) {
+			return sum >= keep, 0
+		}
+		kept := leastOverFund(enough, utxos)
```
chappjc, line 2323: Looks like you've shaddowed the `enough` provided as an input to `fundInternal`, so that `tryFund` a few lines down now gets this simplified enough func.

How about a constructor in coin_selection.go like:

```go
func reserveEnough(amt uint64) func(_, sum uint64) (bool, uint64) {
	return func(_, sum uint64) (bool, uint64) {
		return sum >= amt, 0
	}
}
```

so that it can be used inline here and in `estimateSwap` like

```go
kept := leastOverFund(reserveEnough(keep), utxos)`
```

client/asset/btc/btc.go
```diff
@@ -5474,6 +5711,75 @@ func (btc *baseWallet) BondsFeeBuffer() uint64 {
...
 }
 
+// FundMultiOrder funds multiple orders in one shot. MaxLock is the maximum
+// amount that the wallet can lock for these orders. If maxLock == 0, then
+// there is no limit. An error is returned if the wallet does not have enough
+// available balance to fund each of the orders, however, if splitting is
+// not enabled and all of the orders cannot be funded due to mismatches in
+// UTXO sizes, the orders that can be funded are funded. The values in the
+// MultiOrder parameter are funded in the order they are passed in.
+func (btc *baseWallet) FundMultiOrder(mo *asset.MultiOrder, maxLock uint64) ([]asset.Coins, [][]dex.Bytes, uint64, error) {
```
chappjc, line 5721: In `The values in the MultiOrder parameter are funded in the order they are passed in.`, does that guarantee that it stops on the first failed order?  I ask because I'm pretty sure the caller of `FundMultiOrder` will want to know the values of the orders that it was able to fund, and if it skips to the next value if one fails, you won't know what you got funded.  Can the doc be clarified?

client/asset/btc/btc.go
```diff
@@ -5474,6 +5711,75 @@ func (btc *baseWallet) BondsFeeBuffer() uint64 {
...
+		return nil, nil, 0, fmt.Errorf("error getting wallet balance: %w", err)
+	}
+	if bal.Available < totalRequiredForOrders {
+		return nil, nil, 0, fmt.Errorf("insufficient funds. %d < %d",
+			bal.Available, totalRequiredForOrders)
+	}
+
+	if mo.FeeSuggestion > mo.MaxFeeRate {
+		return nil, nil, 0, fmt.Errorf("fee suggestion %d > max fee rate %d", mo.FeeSuggestion, mo.MaxFeeRate)
+	}
```
chappjc, line 5758: I suggest doing these cheap fee rate checks before the balance checks.

client/asset/btc/btc.go
```diff
@@ -5474,6 +5711,75 @@ func (btc *baseWallet) BondsFeeBuffer() uint64 {
...
+	// Check wallets fee rate limit against server's max fee rate
+	if btc.feeRateLimit() < mo.MaxFeeRate {
+		return nil, nil, 0, fmt.Errorf(
+			"%v: server's max fee rate %v higher than configued fee rate limit %v",
+			dex.BipIDSymbol(BipID), mo.MaxFeeRate, btc.feeRateLimit())
+	}
+
+	customCfg := new(fundMultiOptions)
+	err = config.Unmapify(mo.Options, customCfg)
```
chappjc, line 5767: I'd like if we can avoid marshalling into an in-memory ini file or using the go-ini package for this.  It very inefficient, and a large dep we should avoid getting dug in on further.

We've defined `splitKey` for instance, so I think we can do that for this too, as long as the docs for `fundMultiOptions` refer to the const keys so it is less likely to break with future changes.

client/asset/btc/btc.go
```diff
@@ -2361,6 +2370,228 @@ func (btc *baseWallet) fund(keep uint64, minConfs uint32, lockUnspents bool,
...
+	// If could not fully fund, fund as much as possible in the priority
+	// order.
+	allFundingUTXOs = fundInOrder(values)
+	return returnValues(allFundingUTXOs)
+}
+
+// fundMultiWithSplit creates a split transaction with one output for each
+// order in order to fund the orders. An error is returned if there is not
+// enough funds in the wallet to cover the split transaction for each order.
+func (btc *baseWallet) fundMultiWithSplit(keep, maxLock uint64, values []*asset.MultiOrderValue, splitTxFeeRate, maxFeeRate, splitBuffer uint64) ([]asset.Coins, [][]dex.Bytes, uint64, error) {
```
chappjc, line 2472: I think we should wrap these func signatures after the `MultiOrderValue`.

chappjc, line 2472: Will you doc `splitBuffer` somewhere, either here or on the tagged options struct.
I don't understand it's purpose or how to set it.

chappjc (changes requested):

client/asset/btc/btc.go
martonp: Yes it does guarantee that. I'll update docs.

client/asset/btc/btc.go
martonp: I've created a special `decodeFundMultiOptions` function. Is this what you were going for? Would it be better if we just passed json strings for all of these options instead of `map[string]string` and using ini?

chappjc: Yeah, that's good for now I think.  No silly ini intermediate.  Maybe about json strings, dunno.  It all comes back to the pain point of having all the json keys be strings too instead of typed, which would allow a direct unmarshal into a struct from json, but I recall the challenge is typing values on the frontend.  Good for now imo.

client/asset/btc/btc.go
```diff
@@ -761,9 +762,47 @@ func (s *swapOptions) feeBump() (float64, error) {
...
+	// Use the splitKey const defined above in the options map to set this option.
+	Split *bool
+	// SplitBuffer, if set, will instruct the wallet to add a buffer onto each
+	// output of the multi-order split transaction (if the split is needed).
+	// SplitBuffer is defined as a percentage of the output. If a .1 BTC output
+	// is required for an order and SplitBuffer is set to 5, a .105 BTC output
+	// will be created.
+	//
+	// Use the splitBufferKey const defined above in the options map to set this.
+	SplitBuffer *uint64
```
chappjc, line 780: This still doesn't quite explain the motivation for the split buffer.  As far as I am aware, it makes sense in the following situations:
- this is the quote asset in a market, not the base asset
- you might cancel the order later, and put up a new one with a slightly different rate, thus requiring a different quote amount for the same base assets amount e.g. still 1 lot

The above makes the most sense if the price is going down, thus requiring more quote asset for the same amount of base asset.  If price goes up, this would worsen the overlock.

client/asset/interface.go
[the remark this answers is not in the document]
chappjc: As soon as we started including information about the redeem asset, this wording became less redundant.  It's unfortunate, but there is ambiguity to resolve with these comments.

client/asset/btc/btc.go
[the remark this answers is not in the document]
buck54321: > I wasn't planning on exposing the MultiTrade functionality through the UI.

Where are users going to set these values?

> Why would you use the buffer with a regular trade?

What do you mean? We're not dealing with "regular trades" here. Just bot stuff, right? Isn't the point of the buffer to enable utxo reuse without additional splits or excessive overlock.

client/asset/btc/btc.go
```diff
@@ -761,6 +761,11 @@ func (s *swapOptions) feeBump() (float64, error) {
 	return bump, nil
 }
 
+type fundMultiOptions struct {
+	Split       *bool   `ini:"swapsplit"`
+	SplitBuffer *uint64 `ini:"splitbuffer"`
+}
```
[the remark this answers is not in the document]
buck54321: So only people who are writing their own software can benefit from these settings?

client/asset/btc/btc.go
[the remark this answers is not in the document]
buck54321: If the caller asks to fund 5 orders, but we return partial results for 3, is the caller going to place those 3 orders? Do they have any say in which orders get funded?

It's kinda confusing because for normal order funding, if we can't do it without a split, we definitely can't do it with a split. I don't think that's the case with `FundMulti`, because overlocks stack.

client/asset/btc/btc.go
chappjc: > How is the end user going to set the values? Are they just expected to know what options are available for all assets?

Valid question.  `PreOrder` provides these options for regular trades, so perhaps it needs to work with multi-trades too (or a similar method) so that a bot frontend can show what options are available.

> The point of the INI stuff is so that we have some flexibility for assets to define their own options

IMO, INI only makes sense with reading a wallet setting file, not handling order options.   `OrderOptions` provided _to_ user/frontend  from preorder are a well-defined struct, while options returned _from_ frontend are `{"string": "string"}` objects that unmarshal naturally to `map[string]string`.  No ini use required there.

When we do `config.Unmapify` all over the place, we're just taking advantage of the `ini` tags on the struct fields.  This is what I find unfortunate.  There's no real reason we need to write to an in-memory ini-formatted buffer to convert `map[string]string` -> `struct` except because we decided to put the keys in the struct tags.  We could have instead written an unmarshal helper for each of these "options structs" and we'd not require the ini package or the intermediate data format.  I think we'll get bit by this.

> and the `map[string]string` is easy to store in the DB

True, it's easy, however `config.Data(options)` to prep bytes for a db bucket could just as easily have been JSON with little extra data overhead.

client/core/core.go
```diff
@@ -5630,6 +5630,29 @@ func (c *Core) PreOrder(form *TradeForm) (*OrderEstimate, error) {
...
+		return nil, err
+	}
+
+	orders := make([]*Order, 0, len(reqs))
+
+	var numSuccessful uint64
+	for _, req := range reqs {
+		corder, err := c.sendTradeRequest(req)
+		if err != nil {
+			return nil, fmt.Errorf("error sending trade request after %d successful sends: %v", numSuccessful, err)
```
chappjc, line 5647: This is dicey because they don't know what the orders are they just submitted if some succeeded.
One kinda wonky pattern I've seen is to return a custom error type that includes any data, which could be the partial list of submitted orders.

```go
type errorMultiTrade struct {
    error
    submitted []*Order
}
```

Any better ways to deal with this?

client/asset/btc/btc.go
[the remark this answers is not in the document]
chappjc: I had roughly the same question: https://github.com/decred/dcrdex/pull/2362#discussion_r1210727916
It stops on the first one that fails.  I agree that's a difficult method to use effectively.  I think it would need a flag indicate if you want to allow that, or return the coins and fail entirely if you can't get them all.

Related, in caller, it returns and error even if some orders were submitted: https://github.com/decred/dcrdex/pull/2362#discussion_r1212192557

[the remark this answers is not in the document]
martonp: The idea was that the caller passes the orders in order of priority. I can definitely add a flag. I was considering adding an `allOrNothing` flag for this, but I didn't because I didn't see any of the bots passing true for this.

client/asset/btc/btc.go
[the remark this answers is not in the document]
martonp: > @martonp please check my understanding. Ignoring fees, if we want to fund 3 1-lot orders with a lot size of 1, and we have three utxos of 1, 1, and 1000, we would end up locking up all of it because fundMultiBestEffort would succeed without error, resulting in an overlock of 999 and precluding any additional orders until the overlocked order matches and a swap is generated. And that is regardless of what the user has set for fundMultiOptions.Split.

Yes, this is correct, as long as `maxLock` is less than or equal to 1002. I'm thinking maybe the confusion is coming from the fact that the goal if splitting in `FundOrder` is avoiding overlock, but the goal of splitting in `FundMultiOrder` is to be able to fund multiple orders, and `maxLock` is what deals with the overlocking. Maybe splitting in `FundMultiOrder` should be called something else?

client/asset/btc/btc.go
martonp: > The point of the INI stuff is so that we have some flexibility for assets to define their own options and the map[string]string is easy to store in the DB, as opposed to say a custom struct. Are we going to store the user's bot programs to disk?

I just meant replacing the `ini` with json strings in the DB, just because @chappjc said `ini` was expensive. Not sure marshalling and unmarshalling json is much cheaper though.

> How is the end user going to set the values? Are they just expected to know what options are available for all assets?

The bots themselves have a configuration. The end-user only has to know the bot configuration, not the configuration for the wallets. On the UI we can hide the splitting related settings if neither of the wallets support it, but if someone is running market making using the RPC and a json config file, the settings would just be ignored if they are not relevant. For the UI, something like `PreOrder` would definitely be needed to know whether or not the settings should be hidden.

client/asset/btc/btc.go
martonp: Yes that motivation is right. I'll add the motivation to the docs as well.

client/asset/btc/btc.go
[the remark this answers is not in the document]
martonp: Yes, or people running market makers. There will be settings in the `BotConfig` which tell the market maker whether or not to set these.

client/rpcserver/types.go
```diff
@@ -578,6 +585,73 @@ func parseTradeArgs(params *RawParams) (*tradeForm, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	numPlacements, err := checkIntArg(params.Args[5], "numPlacements", 32)
+	if err != nil {
+		return nil, err
+	}
+
+	placements := make([]*core.QtyRate, 0, 4)
```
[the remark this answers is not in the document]
martonp: Good call.

client/asset/btc/btc.go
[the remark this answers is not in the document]
buck54321: > I'm thinking maybe the confusion is coming from the fact that the goal if splitting in `FundOrder` is avoiding overlock, but the goal of splitting in `FundMultiOrder` is to be able to fund multiple orders, and `maxLock` is what deals with the overlocking.

Adding to the confusion, the default for `allowSplit` is `btc.useSplitTx()` which is set in configuration and has a known definition. 

How would a caller calculate `maxLock`? They won't know how to calculate fees.

Splitting should really be a default behavior for a bot, we just need to be smarter about how we do it. On what reasoning do you think a bot operator may want to not enable splitting?

client/asset/btc/btc.go
buck54321: Example: Ignoring fees, orders of (10, 10, 10), utxos of (10, 10, 15), `maxLock = 30`, `allowSplits = true`, `fundMultiBestEffort` would fail because of `maxLock`. `fundMultiWithSplit` would consume all three inputs to create outputs of (10, 10, 10,  5), even though we already had two outputs of 10. The ideal split would just consume the utxo of value 15 and create outputs at (10, 5). Inputs are much more expensive than outputs. So the current algo almost triples the fees associated with the split transaction compared to the ideal.

martonp: Ah good point. I'll combine the best effort logic with splitting.

client/asset/btc/btc.go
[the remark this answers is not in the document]
martonp: > Adding to the confusion, the default for allowSplit is btc.useSplitTx() which is set in configuration and has a known definition.

True, I'll remove that default.

> How would a caller calculate maxLock? They won't know how to calculate fees.

`maxLock` is set to the balance that is allocated to the bot that is placing the orders. You can see this logic in https://github.com/decred/dcrdex/pull/2332

> Splitting should really be a default behavior for a bot, we just need to be smarter about how we do it. On what reasoning do you think a bot operator may want to not enable splitting?

I think splitting should be the default, but potentially someone would want to turn on the bot and play around with it without automatically having to spend money.

client/asset/btc/btc.go
martonp: Why would you want that? With the base asset, once you have an output exactly the size of the order you are placing, the best effort funding should match that output with the order every time it is cancelled and replaced.

client/asset/btc/btc.go
```diff
@@ -2361,6 +2411,234 @@ func (btc *baseWallet) fund(keep uint64, minConfs uint32, lockUnspents bool,
...
+		return nil, nil, 0, err
+	}
+	txHash := tx.TxHash()
+
+	coins := make([]asset.Coins, len(values))
+	redeemScripts := make([][]dex.Bytes, len(values))
+	ops := make([]*output, len(values))
+	for i := range coins {
+		coins[i] = asset.Coins{newOutput(&txHash, uint32(i), uint64(tx.TxOut[i].Value))}
+		redeemScripts[i] = []dex.Bytes{tx.TxOut[i].PkScript}
```
buck54321, line 2598: A redeem script is the script for a p2sh address. p2pkh addresses have no redeem script.

client/asset/btc/btc.go
buck54321: I'm not following. I'm saying that if market ABC-XYZ existed on server 1 with lot size 1 and fee rate 10, and market ABC-XYZ existed on server 2 with lot size 1 and fee rate 11, the sized ABC utxos from orders on server 1 would not be reusable for orders on server 2, and could result in split txs being generated. I don't consider this situation important, and mesh may make it impossible.

A more likely example might be two markets on the same server with the same base asset, one with lot size 10 and another with lot size 20. If we have a utxo sized at 20 with fees for one swap, it couldn't be used on the market with lot size 10 for a 2-lot order. Buffer fixes that.

client/asset/btc/btc.go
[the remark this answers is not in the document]
buck54321: > `maxLock` is set to the balance that is allocated to the bot that is placing the orders. You can see this logic in #2332

Ah. I see `form.MaxLock = c.balanceHandler.botBalance(c.botID, fromAsset)` now.

> .. `maxLock` is what deals with the overlocking. 

If `maxLock` is being used to indicate the balance available to the bot, can it simultaneously be used to limit overlock?

client/asset/btc/btc.go
martonp: That's true, but I was imagining that generally all the orders for each market maker would pretty much always be on the books. Then, occasionally, there would be a cancel order and a new order in the next epoch in which likely the same output that was previously used would be used again.

client/asset/btc/btc.go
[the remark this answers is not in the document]
martonp: The bot code could potentially calculate the amount required for the orders, add a buffer on top, and pass that for max lock, if that's desired. The main thing I was concerned with here is not using the other bots' funds.

client/asset/btc/btc.go
buck54321: I took a swing at a hybrid approach [here](https://github.com/buck54321/dcrdex/compare/e2d6f5b...buck54321:dcrdex:fundmulti), but the diff is kinda messy, so might be easier to look at [this](https://github.com/buck54321/dcrdex/blob/4b01c6af4593fe4ee1b235d28746e2b1fffa4fde/client/asset/btc/btc.go#L2436) first. (Not tested and tests not updated)

The primary differences
- Separates available funding (`maxLock`) from allowable overfunding (`overfundBuffer`)
- Only creates splits when existing outputs don't work

A notable consequence
- Orders can be funded out of order in partial results. For instance, if we're funding values (1, 3, 1), and we have utxos (1, 1, 50), low overfund allowance, and `allowSplit = false`, the first and last orders will be funded, but not the middle, since that would require an high overlock

client/asset/btc/spv_test.go
```diff
@@ -98,6 +98,22 @@ func (c *tBtcWallet) FetchInputInfo(prevOut *wire.OutPoint) (*wire.MsgTx, *wire.
 func (c *tBtcWallet) ResetLockedOutpoints() {}
 
 func (c *tBtcWallet) LockOutpoint(op wire.OutPoint) {
+	if c.lockedCoins != nil {
+		// check if already locked
+		for _, l := range c.lockedCoins {
+			// print the
```
JoeGruffins, line 104: print the ....

JoeGruffins (approved):

martonp:
I've been testing using https://github.com/decred/dcrdex/pull/2409 and this has been working well.

chappjc:
> I've been testing using #2409 and this has been working well.

Great.  I'll to a few manual mult-trades with the RPC client, but I have no more comments or requests on the code.

client/asset/btc/btc.go
```diff
@@ -2361,6 +2419,429 @@ func (btc *baseWallet) fund(keep uint64, minConfs uint32, lockUnspents bool,
...
+	// and continuing until all orders are funded.
+	allFundingUTXOs := fundAllOrders()
+	if allFundingUTXOs != nil {
+		return returnValues(allFundingUTXOs)
+	}
+
+	// If could not fully fund, fund as much as possible in the priority
+	// order.
+	allFundingUTXOs = fundInOrder(values)
+	return returnValues(allFundingUTXOs)
```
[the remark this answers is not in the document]
chappjc: Good point @JoeGruffins.  @martonp has resolved this as suggested, so don't change anything now, but my only comment would have been regarding the semantics of the parameter being called `splitAllowed` instead of `allowPartialFallback` or similar.  As it stands, `fundMultiBestEffort` is not realistically going to be used from anything other than `fundMulti`, so I don't have a big issue with `splitAllowed`.

chappjc (approved):
All working as expected via the RPC server's `multitrade` command.  I'm excited to see how this comes together in the final bot project.

client/asset/btc/btc.go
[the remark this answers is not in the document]
chappjc: Another simple approach would be to set maxLock to the smaller of the bot balance or some multiple of the total-value of the multi-trade.  e.g. in the (1,1,1000) UTXO case where the total order value is just 3, if you set maxLock to 9 and avoid insane levels of overlock.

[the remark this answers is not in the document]
chappjc: Regarding default to split, always doing a split is not great for several reasons, the top being the fees, but also the greatly slowed order submission because server is then waiting on network tx propagation to verify the fresh 0-conf funding UTXO.

client/core/core.go
chappjc: We're really close on the gap limit issues thanks to you, @buck54321.  Just a few minor comments there needing response/revision https://github.com/decred/dcrdex/pull/2368

PR #2363, opened 2023-05-21
site: add cache buster utilities and test

buck54321:
Take it or leave it. Add script to update cache busters. Set cache busters to a hash of the relevant directory. Ensure cache busters are current in run_tests.sh.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,8 @@ jobs:
       run: npm run lint
     - name: npm run build
       working-directory: ./client/webserver/site
-      run: npm run build
+      run: ./ci_tests.sh
+
   lint-docs:
     name: Lint Markdown
     runs-on: ubuntu-latest
--- a/client/webserver/site/bust_caches.sh
+++ b/client/webserver/site/bust_caches.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
-SITE_DIR=`cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P`
+SITE_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P)
+cd "${SITE_DIR}"; npm run build; cd -
 source ${SITE_DIR}/cache_utilities.bash
 setcssbuster
 setjsbuster
--- a/client/webserver/site/bust_caches.sh
+++ b/client/webserver/site/bust_caches.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 SITE_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P)
-cd "${SITE_DIR}"; npm run build; cd -
+cd "${SITE_DIR}"; npm ci && npm run build; cd -
 source ${SITE_DIR}/cache_utilities.bash
 setcssbuster
 setjsbuster
--- a/client/webserver/site/bust_caches.sh
+++ b/client/webserver/site/bust_caches.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+SITE_DIR=`cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P`
+source ${SITE_DIR}/cache_utilities.bash
+setcssbuster
+setjsbuster
+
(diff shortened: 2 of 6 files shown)


client/webserver/site/cache_utilities.bash
```diff
@@ -0,0 +1,15 @@
+SITE_DIR=`cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P`
+BB_FILEPATH=${SITE_DIR}/src/html/bodybuilder.tmpl
+CSS_DIR=${SITE_DIR}/src/css
+JS_DIR=${SITE_DIR}/src/js
+
+bbhash () { sha1sum ${SITE_DIR}/src/html/bodybuilder.tmpl | head -n1 | cut -d " " -f1 ; }
+hashdir () { sha1sum $1/* | md5sum | head -n1 | cut -d " " -f1 ; }
```
buck54321, line 7: I guess this hashing scheme isn't going to work, because tests are passing locally but not in CI. Probably some small difference in the output of `sha1sum` when reading a directory. I initially had something like
```bash
git ls-files -s ${CSS_DIR} | git hash-object --stdin
```
but `ls-files -s` prints the commited hash, not the current hash, so you would be have to commit, generate cache busters, and then amend, which is crap. Also no guarantee that git versions are consistent with that `ls-files` directory either. 

Maybe this won't work after all.

chappjc: I think it's because we have https://github.com/decred/dcrdex/blob/5daece26338af5c056e4ed00ee282973c00c300b/client/webserver/site/webpack/common.js#L41-L43
so it turns out to be the annoying process like you said.

We could remove that from the webpack build and then the content hash wouldn't change with the git commit advancing.

client/webserver/site/cache_utilities.bash
```diff
@@ -0,0 +1,15 @@
+SITE_DIR=`cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P`
+BB_FILEPATH=${SITE_DIR}/src/html/bodybuilder.tmpl
+CSS_DIR=${SITE_DIR}/src/css
+JS_DIR=${SITE_DIR}/src/js
```
chappjc, line 4: The relevant hashes are really the outputs in dist I think.  There are other things that could change the build output without the sources changing (dependencies updated).

buck54321: Unfortunately, the output files can be either development or production builds, and our current CI flow doesn't guarantee that those files are built by the time `run_tests.sh` runs.

I've now made the cache buster a concatenation of the source dir and output hashes, but only checking the source dir hash for validation.

buck54321:
I've tried to remove the commit hash from the js build before hashing, but there must be other inconsistencies in the output files. Gonna have to give up on CI checking the output hash I think.

chappjc:
> I've tried to remove the commit hash from the js build before hashing, but there must be other inconsistencies in the output files. Gonna have to give up on CI checking the output hash I think.

Are you sure you are using `clean-install` not `install` locally?

run_tests.sh
```diff
@@ -2,15 +2,31 @@
...
 
+# Ensure cache busters are updated
+set +x
+source ./client/webserver/site/cache_utilities.bash
+CSS_HASH=`hashcsssrc`
+CSS_BUSTER=`csssrcbuster`
+JS_HASH=`hashjssrc`
+JS_BUSTER=`jssrcbuster`
+if [ "${CSS_HASH}" != "${CSS_BUSTER}" ] || [ "${JS_HASH}" != "${JS_BUSTER}" ]; then
+	setcssbuster
```
chappjc, line 20: If it gets in here it will have to first do `npm ci && npm run build` before running the `set` helpers

client/webserver/site/bust_caches.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+SITE_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P)
+cd "${SITE_DIR}"; npm run build; cd -
```
chappjc, line 3: if you include `npm ci` here, I think CI will be consistent with local.

.github/workflows/build.yml
```diff
@@ -49,7 +49,8 @@ jobs:
       run: npm run lint
     - name: npm run build
       working-directory: ./client/webserver/site
-      run: npm run build
+      run: ./ci_tests.sh
```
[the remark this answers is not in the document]
chappjc: Well, hmm it could get long that way.  Whatever I guess, but maybe the script should be `check_cache_busters.sh` and `npm run build` can remain outside of it (or include both `npm ci` and `npm run build` too in the script)

chappjc (approved):

Issue #2366, opened 2023-05-22
dexc-desktop: systray option for conditional shutdown


buck54321:
per @chappjc on Element

> I think if the systray on Win/Lin had a checkbox "Shutdown when all windows are closed" or the opposite "Keep running on windows closed", then I think everyone would be happy

chappjc:
Does https://github.com/decred/dcrdex/pull/2367 resolve this and would you like to reword or reverse that check box logic?

PR #2367, opened 2023-05-22
dexc-desktop: checkbox to persist after window close

chappjc:
Resolves https://github.com/decred/dcrdex/issues/2366

image


buck54321 (approved):

ukane-philemon:
@martonp, do you think this optional setting will be a good addition for macOS?

PR #2368, opened 2023-05-22
client/asset/btc: recycle unused redemption and refund addresses

buck54321:
Concerns about overshooting the gap policy have come up. This enables a file-backed recycled address cache for unused redemption and refund addresses.

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -509,11 +509,10 @@ func (ci *auditInfo) SecretHash() dex.Bytes {
 // swapReceipt is information about a swap contract that was broadcast by this
 // wallet. Satisfies the asset.Receipt interface.
 type swapReceipt struct {
-	output        *output
-	contract      []byte
-	signedRefund  []byte
-	expiration    time.Time
-	refundAddress string
+	output       *output
+	contract     []byte
+	signedRefund []byte
+	expiration   time.Time
 }
 
 // Expiration is the time that the contract will expire, allowing the user to
@@ -542,14 +541,6 @@ func (r *swapReceipt) String() string {
 // funds to the user in the case a contract expires.
 func (r *swapReceipt) SignedRefund() dex.Bytes {
 	return r.signedRefund
-}
-
-var _ asset.RefundReceipt = (*swapReceipt)(nil)
-
-// RefundAddress is an add-on method that implements asset.RefundReceipt so that
-// unused refund addresses will be returned.
-func (r *swapReceipt) RefundAddress() string {
-	return r.refundAddress
 }
 
 // RPCConfig adds a wallet name to the basic configuration.
@@ -4049,11 +4040,10 @@ func (btc *baseWallet) Swap(swaps *asset.Swaps) ([]asset.Receipt, asset.Coin, ui
 			return nil, nil, 0, fmt.Errorf("error serializing refund tx: %w", err)
 		}
 		receipts = append(receipts, &swapReceipt{
-			output:        output,
-			contract:      contracts[i],
-			expiration:    time.Unix(int64(contract.LockTime), 0).UTC(),
-			signedRefund:  refundBuff.Bytes(),
-			refundAddress: refundAddr.String(),
+			output:       output,
+			contract:     contracts[i],
+			expiration:   time.Unix(int64(contract.LockTime), 0).UTC(),
+			signedRefund: refundBuff.Bytes(),
 		})
 	}
 
@@ -4829,7 +4819,7 @@ func (btc *baseWallet) recyclableAddress() (string, error) {
 			recycledAddr = addr
 			break
 		} else if err != nil {
-			btc.log.Errorf("Error checking ownership of recycled address: %v", err)
+			btc.log.Errorf("Error checking ownership of recycled address %q: %v", addr, err)
 			// Don't delete the address in case it's just a network error for
 			// an rpc wallet or something.
 		} else { // we don't own it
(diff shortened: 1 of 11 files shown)


chappjc:
> Concerns about overshooting the gap policy have come up. I wanted to see if there was any appetite for a solution like this. 

To a large extent, yeah, this is probably very helpful.

Two things:
- the refund address that is requested internally in the Swap method is the other culprit
- ideally this would all be internal to the wallet.  Refund addresses could be keyed to swap funding coins, such that the usual ReturnCoins on order cancel could find and move the unused funding address to the reuse map.  Redeem addresses... yeah I guess you have the only solution there, Core involvement, since there's no interaction with the "to" wallet when the order is canceled (right?).   I'm going to consider that more though.

client/core/trade.go
```diff
@@ -3524,6 +3524,9 @@ func (t *trackedTrade) returnCoins() {
 		} else {
 			t.coinsLocked = false
 		}
+		if returner, is := t.wallets.fromWallet.Wallet.(asset.AddressReturner); is {
+			returner.ReturnAddress(t.Trade().Address)
```
chappjc, line 3528: The redeem address here is for the `toWallet` I believe.

buck54321:
> Refund addresses could be keyed to swap funding coins, such that the usual ReturnCoins on order cancel could find and move the unused refund address associated with the particular planned swap to the reuse map.

I see the issue, but I'm having trouble seeing the solution. Swap funding coins are deleted from our tracking map in `Swap`, right after the refund address is created. I think only core knows if our swap ends up being redeemed by the counterparty, i.e.
that we **will not refund**.

chappjc:
> Swap funding coins are deleted from our tracking map in `Swap`, right after the refund address is created.

Yes, you are correct.  When `ReturnCoins` is used after the swap is made, it's being used to release any locked change.

But more to the point, the refund addresses are only created when a swap is created, and with a bot that's probably less frequent than cancelling unfilled orders.  Is that right and we can generally just be concerned with the unused redeem addresses that are requested when an order is first placed?

buck54321:
> Is that right and we can generally just be concerned with the unused redeem addresses that are requested when an order is first placed?

Not at all. I understand what we need, I just didn't see how to do it without core.

chappjc:
I do see that now.  But I'm trying to roughly characterize where the address waste comes from, proportionally.  If a client is cancelling a lot of unfilled orders, then it's mostly the redeem address.  If a client is settling a lot of trades, then it's the refund address that is goes unused.  So I supposed it's really dependent on use case.   A relatively inactive bot that spends most of the time tracking an oracle to move it's mid-market price around without actually trading much will burn the redeem addresses.

martonp:
I think this is good. We just can just return the refund addresses from Swap, and store them in the db. On cancels and refunds, core returns the redeem addresses, and on redemptions core returns the refund address. Overall it seems pretty clean, core requests addresses from the wallets, and returns them if they are not needed. Definitely the redeem addresses will be wasted more often, but why not do both?

JoeGruffins (approved):

chappjc (changes requested):

client/asset/btc/btc.go
```diff
@@ -847,6 +857,12 @@ type baseWallet struct {
 	// bond amounts. This amount only changes via ReserveBondFunds, and it is
 	// used to recognize when all reserves have been released.
 	bondReservesNominal int64 // only set by ReserveBondFunds
+
+	recycledAddrMtx sync.Mutex
+	// recycledAddrs are returned, unused redemption addresses. We track these
+	// to avoid issues with the gap policy.
+	recycledAddrs map[string]struct{}
+	recyclePath   string
```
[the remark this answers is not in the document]
buck54321: `writeRecycledAddrsToFile`

martonp (approved):
LGTM, working well

client/asset/btc/btc.go
```diff
@@ -5907,3 +6015,15 @@ func (btc *baseWallet) ConfirmRedemption(coinID dex.Bytes, redemption *asset.Red
...
 }
+
+// writeRecycledAddrsToFile should be called with the recycledAddrMtx locked.
+func (btc *baseWallet) writeRecycledAddrsToFile() {
+	addrs := make([]string, 0, len(btc.recycledAddrs))
+	for addr := range btc.recycledAddrs {
+		addrs = append(addrs, addr)
+	}
+	contents := []byte(strings.Join(addrs, "\n"))
+	if err := os.WriteFile(btc.recyclePath, contents, 0644); err != nil {
```
chappjc, line 6026: Missed this in last review, sorry, but I think the file should be 0600, not 0644 since it contains information that could be considered a privacy leak.

client/asset/btc/btc.go
```diff
@@ -4276,9 +4310,83 @@ func (btc *baseWallet) DepositAddress() (string, error) {
...
+		} else if err != nil {
+			btc.log.Errorf("Error checking ownership of recycled address %q: %v", addr, err)
+			// Don't delete the address in case it's just a network error for
+			// an rpc wallet or something.
+		} else { // we don't own it
+			delete(btc.recycledAddrs, addr)
+		}
+	}
+	if len(btc.recycledAddrs) != nRecycles {
+		btc.writeRecycledAddrsToFile()
```
chappjc, line 4336: Performing file system operations while a mutex is locked is not my favorite thing in the world.  Alright if workarounds are ugly or a lot of work, but one option is `func (btc *baseWallet) writeRecycledAddrsToFile(addrs []string)`, called after unlock.

A second idea is to defer writing of the file until the wallet is shutdown.  i.e. `(btc *baseWallet) connect(ctx context.Context)` could end with:

```go
	wg.Add(1)
	go func() {
		<-ctx.Done()
		btc.writeRecycledAddrsToFile()
	}()

	return &wg, nil
}
```

We don't seem to read from the file except on wallet start, so we only really need to write to it on shutdown.  The consequences of a stale file left from a crashed dexc are minimal.

client/asset/interface.go
```diff
@@ -618,6 +618,18 @@ type NewAddresser interface {
...
 }
 
+// AddressReturner is a wallet that allows recycling of unused redemption or refund
+// addresses. Asset implementations should log any errors internally. The caller
+// is responsible for only returning unused addresses.
+type AddressReturner interface {
+	// ReturnRefundContracts should be called with the Receipt.Contract() data
+	// for any swaps that will not be refunded.
+	ReturnRefundContracts(contracts [][]byte)
+	// ReturnRedemptionAddress acceptsa  Wallet.RedemptionAddress() if the
```
chappjc, line 628: acceptsa :)

chappjc (changes requested):

chappjc (approved):
Great.  Confirmed on-chain that unused addresses from canceled orders are getting used in subsequent swaps.

I think we're fine without this for dcrwallet by virtue of the "wrap" gap policy.

PR #2375, opened 2023-06-02
build: golangci-lint v1.53.2

dajohi:


ukane-philemon (approved):

.golangci.yml
buck54321, line 1: Can add

```YAML
output:
  format: github-actions,colored-line-number
```

and remove the cli flag.

.golangci.yml
```diff
@@ -0,0 +1,25 @@
...
+    - ineffassign
+    - makezero
+    - nosprintfhostport
+    - reassign
+    - rowserrcheck
+    - sqlclosecheck
+    - tparallel
+    - typecheck
+    - unconvert
+    - vetshadow
```
buck54321, line 25: From what I can tell, **vetshadow** is enabled by default with **govet**. They don't even document it as a separate linter. Can remove?

---

#### For documentation, the new linters added in the pull request are...
bidichk: Checks for dangerous unicode character sequences
durationcheck: check for two durations multiplied together
exportloopref: checks for pointers to enclosing loop variables
grouper: An analyzer to analyze expression groups, e.g. const, var
nosprintfhostport: Checks for misuse of Sprintf to construct a host with port in a URL
reassign: Checks that package variables are not reassigned in another package
tparallel: detects inappropriate usage of t.Parallel() method in your Go test codes
typecheck: Like the front-end of a Go compiler, parses and type-checks Go code

buck54321 (changes requested):

buck54321 (approved):

PR #2378, opened 2023-06-05
dex/networks: dgb & doge have duplicate netid

dev-warrior777:
Regtest net ids are the same for Digibyte & Doge. They use the old btc one unchanged


buck54321 (approved):
They weren't the same, but at least that'll be obvious now.

PR #2379, opened 2023-06-07
client/asset: Firo should manually unlock unspents

dev-warrior777:
After checking `listlockunspent` after `lockunspent true` & `sendtoaddress` Firo should use the code path to check and 'manually' check and unlock outputs.

Discovered while compare/contrast while testing Dash so dash may be similar.

### DIFF
--- a/client/asset/firo/firo.go
+++ b/client/asset/firo/firo.go
@@ -154,7 +154,7 @@ func NewWallet(cfg *asset.WalletConfig, logger dex.Logger, network dex.Network) 
 		NumericGetRawRPC:         false, // getrawtransaction uses either 0/1 Or true/false
 		LegacyValidateAddressRPC: true,  // use validateaddress to read 'ismine' bool
 		SingularWallet:           true,  // one wallet/node
-		UnlockSpends:             false, // checked after sendtoaddress
+		UnlockSpends:             true,  // checked after sendtoaddress and listlockunspent
 		AssetID:                  BipID,
 		FeeEstimator:             estimateFee,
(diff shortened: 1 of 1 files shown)


dev-warrior777:
Let me meditate in this one .. could be a function of `livetest` code I am running. 
`client/asset/firo/regnet_test.go`

Firo otherwise runs just fine.

Today I tested dash with the same livetest code: both with and without `UnlockSpends` in an endless loop and I eventually get `livetest.go:274: unlocked output not returned` .. restarted the harness and got the same result
Tried big coins, many small coins for funding, mixed

dev-warrior777:
Ok you cannot run this livetest in an endless loop:
1. you will eventually run out of coins
2. you have to clean up - mine the mempool, unlock some/all locked outputs left from the test

What started me to suspect that Firo needs help unlocking with `BTCCloneCFG.UnlockSpends` was seeing locked coins at the end of a successful test run. But setting `BTCCloneCFG.UnlockSpends`  to true did not change anything.

There does seem to be a case to answer for `wallet.ReturnCoins(nil)` i.e. Unlock all locked unspents. For me it failed on the two coins I tested because what was sent as marshalled json over rpc was:
`lockunspent true null`

I hacked a fix for this which special cases nil ops.
image
We just send `lockunspent true`


Long story short Firo does not need Manual Unlock and this was tested mostly by verifying that coins funding a raw transaction disappeared from `listunspent` after `sendrawtransaction`

dev-warrior777:
Do you want a PR for the _Unlock all locked unspents_  case?

Anyway I am closing this PR as unnecessary for Firo.

PR #2382, opened 2023-06-13
btc: more sensible bond fee buffer

chappjc:
```
When used for computing the bond fee buffer, the "fee rate limit" makes
an extremely high fee rate buffer, which makes BTC bonds impractical.
This change:
- switches to using a dynamic fee rate with the configured fallback rate
  for both ReserveBondFunds and BondsFeeBuffer, where the former is
  used as the requirement for the frontend to proceed with registration
- uses the exported BondsFeeBuffer method when automatically creating
  the "extra" split output in certain order funding edge cases instead
  of a buffer based on the high limit.
- tweaks the bond tx input count to be slightly less conservative, going
  from 12 -> 8 possible inputs

The motivation for these changes is that using the wallet's configured
fee rate limit creates extremely high fee buffers, whereas we have
defined the "fallback" fee rate to be a reasonable fallback where a
current estimate is not available, which also happens to be applicable
to a *future* fee rate.

Further, this is a buffer and precision is not critical, but turning
users away from on-boarding is a major issue.  The right approach
is to base reserves on a padded estimate based on current rates,
not the swap fee rate limit.

Finally, given this softening of the reserves amount, no adjustment to
the enforced bond reserves in Reconfigure is done since we are now
using current rates as a hit of future rates.

Padding of this fee buffer already comes from:
- planning for a lot of inputs
- 4 parallel bond tracks
- the exported BondsFeeBuffer method using double the provided fee rate

Even if fee rates skyrocket, and the reserves become insufficient for
bond renewal, the user will still post bond if their balance allows, and
they may deposit some nominal additional amount to cover any such
deficit.
```

**Before**

image

(that's about 1500 USD, not sensible)

**After**

image

(that's <100 USD, or 2x the bond amount plus realistic fees)

### DIFF
--- a/client/asset/btc/btc.go
+++ b/client/asset/btc/btc.go
@@ -1444,17 +1444,7 @@ func (btc *baseWallet) Reconfigure(ctx context.Context, cfg *asset.WalletConfig,
 	if err != nil {
 		return false, err
 	}
-
-	oldCfg := btc.cfgV.Swap(newCfg).(*baseWalletConfig)
-	if oldCfg.feeRateLimit != newCfg.feeRateLimit {
-		// Adjust the bond reserves fee buffer, if enforcing.
-		btc.reservesMtx.Lock()
-		if btc.bondReservesNominal != 0 {
-			btc.bondReservesEnforced += int64(bondsFeeBuffer(btc.segwit, newCfg.feeRateLimit)) -
-				int64(bondsFeeBuffer(btc.segwit, oldCfg.feeRateLimit))
-		}
-		btc.reservesMtx.Unlock()
-	}
+	btc.cfgV.Store(newCfg) // probably won't matter if restart/reinit required
 
 	return restart, nil
 }
@@ -1576,7 +1566,7 @@ func (btc *baseWallet) Balance() (*asset.Balance, error) {
 }
 
 func bondsFeeBuffer(segwit bool, highFeeRate uint64) uint64 {
-	const inputCount uint64 = 12 // plan for lots of inputs
+	const inputCount uint64 = 8 // plan for lots of inputs
 	var largeBondTxSize uint64
 	if segwit {
 		largeBondTxSize = dexbtc.MinimumTxOverhead + dexbtc.P2WSHOutputSize + 1 + dexbtc.BondPushDataSize +
@@ -2255,7 +2245,7 @@ func (btc *baseWallet) FundOrder(ord *asset.Order) (asset.Coins, []dex.Bytes, er
 			useSplit = true
 			coins, fundingCoins, spents, redeemScripts, inputsSize, sum, err = btc.fund(reserves, minConfs, true,
 				orderEnough(ord.Value, ord.MaxSwapCount, bumpedMaxRate, btc.initTxSizeBase, btc.initTxSize, btc.segwit, useSplit))
-			extraSplitOutput = reserves + bondsFeeBuffer(btc.segwit, btc.feeRateLimit())
+			extraSplitOutput = reserves + btc.BondsFeeBuffer(ord.FeeSuggestion)
 		}
 		if err != nil {
 			return nil, nil, fmt.Errorf("error funding swap value of %s: %w", amount(ord.Value), err)
@@ -5094,7 +5084,9 @@ func (btc *baseWallet) RegisterUnspent(inBonds uint64) {
 // available balance, otherwise reserves will be adjusted regardless and the
 // funds are pre-reserved. This returns false if the available balance was
 // insufficient iff the caller requested it be respected, otherwise it always
-// returns true (success).
+// returns true (success). The fee buffer is used only when enabling the
+// reserves (when starting from zero) to compute the fee buffer. It may also be
+// zero, in which case the wallet will attempt to obtain it's own estimate.
 //
 // The reserves enabled with this method are enforced when funding transactions
 // (e.g. regular withdraws/sends or funding orders), and deducted from available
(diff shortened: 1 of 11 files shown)


JoeGruffins (approved):

client/asset/btc/btc.go
```diff
@@ -1444,17 +1444,7 @@ func (btc *baseWallet) Reconfigure(ctx context.Context, cfg *asset.WalletConfig,
...
-	if oldCfg.feeRateLimit != newCfg.feeRateLimit {
-		// Adjust the bond reserves fee buffer, if enforcing.
-		btc.reservesMtx.Lock()
-		if btc.bondReservesNominal != 0 {
-			btc.bondReservesEnforced += int64(bondsFeeBuffer(btc.segwit, newCfg.feeRateLimit)) -
-				int64(bondsFeeBuffer(btc.segwit, oldCfg.feeRateLimit))
-		}
-		btc.reservesMtx.Unlock()
-	}
+	btc.cfgV.Store(newCfg) // probably won't matter if restart/reinit required
```
chappjc, line 1447: Back to not touching the bond reserves on wallet config change.

client/asset/btc/btc.go
```diff
@@ -5471,8 +5462,9 @@ func (btc *baseWallet) RefundBond(ctx context.Context, ver uint16, coinID, scrip
 // BondsFeeBuffer suggests how much extra may be required for the transaction
 // fees part of required bond reserves when bond rotation is enabled.
 func (btc *baseWallet) BondsFeeBuffer() uint64 {
-	// 150% of the fee buffer portion of the reserves.
-	return 15 * bondsFeeBuffer(btc.segwit, btc.feeRateLimit()) / 10
+	feeRate := btc.targetFeeRateWithFallback(1, btc.fallbackFeeRate())
+	// 110% of the fee buffer portion of the reserves.
+	return 11 * bondsFeeBuffer(btc.segwit, feeRate) / 10
```
chappjc, line 5467: As a reminder, `BondsFeeBuffer` informs frontend, and gates the progress through the registration forms.  Then `ReserveBondFunds` is used just prior to posting the first bond.  As such, by making the fee rate come from a live estimate, it's possible that the live estimate will go up before they actually hit submit and it calls `ReserveBondFunds`.
For this reason, I might put a multiplier on `feeRate`, while using 110% for the entire buffer.

Thoughts?

buck54321 (commented):
How about a `feeSuggestion uint64` argument to `BondsFeeBuffer` and `ReserveBondFunds` so that wallets without external API enabled will still get realistic values?

chappjc:
> How about a `feeSuggestion uint64` argument to `BondsFeeBuffer` and `ReserveBondFunds` so that wallets without external API enabled will still get realistic values, even if the user has misconfigured their fallback rate?

Oh another pain is that on 0.6, we don't have the recent changes that made all wallets FeeRaters even if they don't have internal sources, only external.  https://github.com/decred/dcrdex/pull/2356

It may still be feasible to do a `dc.fetchFeeRate(bondAsset.ID)`, but `Core` can't get to SPV wallet's external rate sources directly in 0.6.

Lemme see if the feeSuggestion is attainable and can be used as a fallback if internal wallet fetching fails.

chappjc:
Refactored so that the `BondsFeeBuffer` and `ReserveBondFunds` methods now take a fee rate.  If zero, those methods will try to get an internal rate or use the configured fallback.

client/asset/btc/btc.go
```diff
@@ -5469,10 +5466,15 @@ func (btc *baseWallet) RefundBond(ctx context.Context, ver uint16, coinID, scrip
...
-	return 15 * bondsFeeBuffer(btc.segwit, btc.feeRateLimit()) / 10
+// fees part of required bond reserves when bond rotation is enabled. The
+// provided fee rate may be zero, in which case the wallet will use it's own
+// estimate or fallback value.
+func (btc *baseWallet) BondsFeeBuffer(feeRate uint64) uint64 {
+	if feeRate == 0 {
+		feeRate = btc.targetFeeRateWithFallback(1, 0)
+	}
+	feeRate *= 2 // double the current fee rate estimate so this fee buffer does not get stale too quickly
+	return bondsFeeBuffer(btc.segwit, feeRate)
```
chappjc, line 5477: With `BondsFeeBuffer` based on a live rate, there's a real possibility that it could be hard to get past the wallet funding form if they refresh the page after sending the exact recommended amount, as this would force a new estimate that could be a bit higher than what they just sent.

How can we have this more robust, but also using a sensible rate for the asset?  Possibly we round the feeRate up to the nearest 10 so single digit variation won't change the result?

chappjc: Yeah, this definitely needs a solution.  I just produced the behavior I described:

`2023-06-15 17:43:58.974 [DBG] CORE[btc]: Retrieved fee rate from external API: 29`
image

Then I sent that exact amount, but did **not* wait for it to confirm and proceed automatically.

Then I clicked back and then again into BTC:
`[DBG] CORE[btc]: Retrieved fee rate from external API: 39`
image

When my tx confirms, it won't satisfy the frontend's checks, although ReserveBondFunds probably would because of the 150% vs 200% scaling described in my previous comment.

chappjc: I'm going to have the backend (either webserver handler or core) cache the fee buffer for maybe an hour.  This value is heavily padded and meant to be good for weeks anyway.

chappjc:
Will push a refactor shortly.  Sorry if there are reviews in progress.

chappjc:
OK, despite now using a live fee estimate as the basis for the bonds fee buffer, the wallet wait form should not be fickle with respect to the required amount to proceed.  Also, the (*Core).PostBond method now has an optional fee buffer field in it's parameter struct so that the fee buffer that was planned for will be used when setting the reserves.

Last commit fixes an existing js bug where the wallet wait form would be skipped entirely if the wallet existed and had at least one bond amount in available balance.

buck54321 (approved):

PR #2385, opened 2023-06-14
ui: rework app initialization

buck54321:
This reworks app initialization to achieve a number of ux goals.

1) Adds chance for user to quickly configure native wallets and add
   known servers (view-only), https://github.com/decred/dcrdex/pull/1986#discussion_r1061747565
2) Adds seed backup dialog
3) Makes the Wallets view the landing page

image

### DIFF
--- a/client/webserver/site/src/html/init.tmpl
+++ b/client/webserver/site/src/html/init.tmpl
@@ -1,0 +1,111 @@
+{{define "init"}}
+{{template "top" .}}
+<div id="main" data-handler="init" class="main align-items-center justify-content-center flex-column stylish-overflow">
+  <div class="position-absolute" id="forms">
+
+    {{- /* App Initialization */ -}}
+    <form id="appPWForm">
+      <div class="px-2 py-1 text-center fs28 sans-light">[[[Set App Password]]]</div>
+      <div class="fs18 mt-2 sans-light">[[[reg_set_app_pw_msg]]]</div>
+      <hr class="dashed my-4">
+      <div class="pb-3">
+        <label for="appPW" class="form-label ps-1 mb-1">[[[Password]]]</label>
+        <input type="password" class="form-control select" id="appPW" autocomplete="new-password">
+      </div>
+      <div class="pb-3">
+        <label for="appPWAgain" class="form-label ps-1 mb-1">[[[Password Again]]]</label>
+        <input type="password" class="form-control select" id="appPWAgain" autocomplete="off">
+      </div>
+      <div>
+        <input class="form-check-input" type="checkbox" id="rememberPass">
+        <label for="rememberPass" class="form-check-label ps-1">[[[Remember my password]]]</label>
+      </div>
+      <div class="pb-3 d-hide mt-3" id="seedRestore">
+        <label for="seedInput" class="form-label ps-1 mb-1">[[[Restoration Seed]]]</label>
+        <input type="text" class="form-control select" id="seedInput" autocomplete="off">
+      </div>
+      <div class="d-flex justify-content-between mt-4">
+        <div class="px-1 fs13 pointer d-flex justify-content-start align-items-center" id="showSeedRestore"><span class="ico-plus fs11"></span> <div class="ps-2">[[[Restore from seed]]]</div></div>
+        <span></span>
+        <button id="appPWSubmit" type="submit" class="col-8 justify-content-center fs15 bg2 selected">[[[Submit]]]</button>
+      </div>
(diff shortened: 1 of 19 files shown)


ukane-philemon (commented):
Good work @buck54321. I will test try this on my end.

chappjc:
Very nice.  Wallets above servers would be better for the multi-coin wallet pivot, but this is great either way.

JoeGruffins:
Getting the same error as cli when trying to use:
```
$ ./dexc
failed creating web server: error loading localized html templates: error adding template init: template: init:35: comment ends before closing delimiter
```

client/webserver/site/src/html/init.tmpl
```diff
@@ -0,0 +1,110 @@
...
+      </div>
+      <div class="d-flex justify-content-between mt-4">
+        <div class="px-1 fs13 pointer d-flex justify-content-start align-items-center" id="showSeedRestore"><span class="ico-plus fs11"></span> <div class="ps-2">[[[Restore from seed]]]</div></div>
+        <span></span>
+        <button id="appPWSubmit" type="submit" class="col-8 justify-content-center fs15 bg2 selected">[[[Submit]]]</button>
+      </div>
+      <div class="fs15 pt-3 text-center d-hide errcolor text-break" id="appPWErrMsg"></div>
+    </form>
+
+    {{- /* Quick Config Form */-}}
```
ukane-philemon, line 35: nit: Missing space before `-`

JoeGruffins:
Looks good. A few nits:

Cannot click the actual check boxes here. Can press anywhere else but not on the check boxes themselves.
image

"skip this for now" looks like it should be clickable but clicking it does nothing. It seems you cannot skip looking at the seed at all.
image

After just making the client and navigating to markets you will see this. I think there is an extra loading screen when you go the login route that waits for markets to load and so we don't ever get here in the ui normally. This page mus be reloaded at some point to show the markets.
image

Probably in master, but cannot view token markets without making a token wallet. I can see the eth market already with no eth wallet yet so not sure why it is different. On loading candles indefinitely.
image

client/webserver/site/src/js/init.ts
```diff
@@ -0,0 +1,293 @@
...
+    for (const walletRow of this.wallets) {
+      const { asset: a, type, checkbox } = walletRow
+      if (!checkbox.checked) continue
+      const createForm = {
+        assetID: a.id,
+        appPass: this.pw,
+        walletType: type
+      }
+      ani.opts.message = intl.prep(intl.ID_CREATING_WALLET, { name: a.name })
+      const res = await postJSON('/api/newwallet', createForm)
```
martonp, line 234: Wallets could be created in parallel:
``` javascript
    const createWalletPromise = (walletRow: WalletConfigRow) => {
      const { asset: a, type, checkbox } = walletRow
      if (!checkbox.checked) return Promise.resolve()
      const createForm = {
        assetID: a.id,
        appPass: this.pw,
        walletType: type
      }
      return postJSON('/api/newwallet', createForm).then(res => {
        if (!app().checkResponse(res)) failedWallets.push(a.name)
      })
    }
    await Promise.all(this.wallets.map(createWalletPromise))
```

martonp (commented):
Looks good other than the issues which @JoeGruffins also noticed.

chappjc (approved):
This is working very well for me with the last commit. Just some thoughts--

When you choose only to add wallets, no DEX host, we could add a link like this on the wallets page:

image

But that's not a new issue and not necessary.  It could just be a plain `<a href="/register">`

The other thing, which is think is just a design decision, is when you add a DEX using this new interface, it's always "view-only" even if you had already had an account.  We assume no account discovery by default is desirable.  Should we?  I restored from seed where I already had an account, and I saw:

image

And clicking discovered it so I got this simple form:

image

A slightly different approach that would avoid the need to add a "view-only" checkbox on this simplified init flow would be to first try assuming you want to discover an account, and then fall back to view-only if there was no known account.  I don't think backend changes are needed for this.

Anyway, no need to make any changes.  It works as-is!  Thoughts?  We can merge if there's no need to update in your mind.

JoeGruffins (approved):

martonp (approved):

chappjc:
Pls squash/rebase and we can merge.

chappjc:
Rebased and squashed for you so we can merge this since it has been good with approvals for a while.
I'd still like to know what you think about the ideas [here](https://github.com/decred/dcrdex/pull/2385#pullrequestreview-1499566153) but any more changes can wait.

PR #2388, opened 2023-06-14
kvdb: recover from badger db truncation error

buck54321:
See also https://github.com/dgraph-io/badger/issues/744. Probably a Windows thing. Adopting the fix from https://github.com/decred/dcrdata/pull/955.


chappjc (approved):
LGTM.

Worth noting that the truncation option is removed in latest badger versions as it happens automatically now.  We might consider migrating at some point.

PR #2390, opened 2023-06-14
CI: use cache github action for Go mod and test

chappjc:
Copying https://github.com/decred/dcrd/pull/3145, this caches the go module build output folder and build cache.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,6 +10,11 @@ jobs:
       matrix:
         go: ['1.19', '1.20']
     steps:
+      - uses: awalsh128/cache-apt-pkgs-action@1850ee53f6e706525805321a3f2f863dcf73c962 #v1.3.0
+        with:
+          packages: git-restore-mtime libgtk-3-dev libwebkit2gtk-4.0-dev
+          version: 1.0
+
       - name: Set up Go
         uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 #v3.5.0
         with:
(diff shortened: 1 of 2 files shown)


chappjc:
Hmm, it didn't actually use the cache in run 2 although it did load the cache.  https://github.com/decred/dcrdex/actions/runs/5271639637/jobs/9533204074

davecgh:
> Hmm, it didn't actually use the cache in run 2 although it did load the cache.

The first thing that jumps out to me is `env GORACE=halt_on_error=1 go test -race -short -count 1 ./...`.  That `-count 1` is the canonical way to expressly bust the cache.

```
$ go help test | grep -C3 count
If a run of go test has any test or non-test flags outside this set,
the result is not cached. To disable test caching, use any test flag
or argument other than the cacheable flags. The idiomatic way to disable
test caching explicitly is to use -count=1. Tests that open files within
the package's source root (usually $GOPATH) or that consult environment
variables only match future runs in which the files and environment
variables are unchanged. A cached test result is treated as executing
```

davecgh:
You'll also want to see the updates I made to the original PR around timestamp stabilization and the explanation given since you'll likely hit a few of those too.

chappjc:
The `awalsh128/cache-apt-pkgs-action` for apt packages seems to be working well.
`git restore-mtime` also seems to work as expected for the test data files that are in the repo.

Only test that doesn't seem to be cached correctly is bolt, and I don't see what's causing go to miss there.

davecgh:
> `git restore-mtime` also seems to work as expected for the test data files that are in the repo.

Probably not a big deal, since I see it's less than a second in the grand scheme of things, but it is pretty slow comparatively speaking because git doesn't actually store timestamps, so, I'm pretty sure what it's actually doing is finding the most recent commit that touched every single file and then updating the timestamp of that file to match the commit timestamp.

Of course, the benefit is that it works without needing any kind of special naming scheme, so seems like a reasonable solution.

For dcrd, I just wrote a small script that only specifically considers files in a `testdata` directory (because directories with that name are special to Go in that it won't treat it as a package and thus all test data should be in a directory named with that special name)  and creates a timestamp based on the hash of the file contents to ensure the cache is busted when the file contents change.

chappjc:
> Probably not a big deal, since I see it's less than a second in the grand scheme of things, but it is pretty slow comparatively speaking because git doesn't actually store timestamps, so, I'm pretty sure what it's actually doing is finding the most recent commit that touched every single file and then updating the timestamp of that file to match the commit timestamp.

That's correct, and it supposedly requires `fetch-depth: 0` with `actions/checkout` to do that.  This was suggested by bcmills on some golang issue.  It's not perfect, but it works.

I would have stolen your script, but we were already installing other apt packages so it didn't add much to ensure this one was also installed.

If it turns out that `git restore-mtime` is a significant burden, we'll definitely switch.

davecgh:
> That's correct, and it supposedly requires `fetch-depth: 0` with `actions/checkout` to do that. This was suggested by bcmills on some golang issue. It's not perfect, but it works.

Ah, yeah, that makes sense.  That `fetch-depth: 0` tells it to fetch all commits on all branches instead of only the most recent one requested.  Indeed, I see it increases the checkout time from about 1s to 5-6s.  So, it adds an additional 4-5s, and then about half a second for the actual timestamp restoration.

Thus, all considered, it essentially adds about 5 seconds to the overall process to be able to restore the timestamps that way.

chappjc:
Will change to the hash-mod-based time stamp reset method in the future if needed

PR #2392, opened 2023-06-14
[0.6.2] btc: more sensible bond fee buffer

chappjc:
https://github.com/decred/dcrdex/pull/2382 for v0.6.2
Rebased on https://github.com/decred/dcrdex/pull/2391 so SPV wallets are `FeeRater`s, which gives `(*Core).feeSuggestionAny` access to the wallet's fee rate first.


client/webserver/site/src/js/register.ts
```diff
@@ -93,11 +93,11 @@ export default class RegistrationPage extends BasePage {
       const wallet = asset.wallet
       if (wallet) {
         const bondAsset = this.currentDEX.bondAssets[asset.symbol]
-        if (wallet.synced && wallet.balance.available > bondAsset.amount) {
+        const bondsFeeBuffer = await this.getBondsFeeBuffer(assetID, page.regAssetForm)
+        if (wallet.synced && wallet.balance.available > 2 * bondAsset.amount + bondsFeeBuffer) {
```
buck54321, line 97: Nice! I had just noticed this while doing some other work.

client/webserver/api.go
```diff
@@ -263,13 +263,42 @@ func (s *WebServer) apiRegister(w http.ResponseWriter, r *http.Request) {
...
+// given asset are cached for 45 minutes. These values are meant to provide a
+// sensible but well-padded fee buffer for bond transactions now and well into
+// the future, so a long expiry is appropriate.
+func (s *WebServer) bondsFeeBuffer(assetID uint32) (feeBuffer uint64, err error) {
+	// (*Core).BondsFeeBuffer returns a fresh fee buffer based on a current (but
+	// padded) fee rate estimate. We assist the frontend by stabilizing this
+	// value for up to 45 minutes from the last request for a given asset. A web
+	// app could conceivably do the same, but we'll do this here between the
+	// backend (Core) and UI so that a webapp does not need to employ local
+	// storage/cookies and associated caching logic.
```
buck54321, line 276: My thinking was initially to use front end logic too, but the webserver cache is probably better. We are leaving rpcserver hanging for now.

buck54321 (approved):
Looks good.

client/webserver/api.go
chappjc: Correct re: the rpcserver.  However, changing responses from BondsFeeBuffer doesn't really matter for rpc since an rpc user would just record the value they obtained on their first call to BondsFeeBuffer and pass that on to PostBond, whereas on the frontend there's a the possibility to make the UI fetch a new value and you've lost the original.

client/webserver/site/src/js/register.ts
chappjc: Yeah, I totally missed this condition when I was rushing through https://github.com/decred/dcrdex/pull/2200

chappjc: Oh, the same thing exists on settings.ts.  Fixing that in another PR.  At least the first DEX uses the right info.

buck54321: The UI for bonds in settings is pretty rough. I was thinking about opening an issue. Would be good to see the bond sizes, fee buffer, durations, USD conversions, etc.

chappjc: 100%

PR #2395, opened 2023-06-15
site: bond asset choice wording tweaks

chappjc:
Based on feedback from the community and other users new to Decred, this tweaks some of the bond messaging during registration:

- "How will you post bond?" ==> "How will you create your bond?" on the form where you choose the bond asset
- Includes the `whatsabond` info text on the same form, just below the previous question/title

image

### DIFF
--- a/client/webserver/site/src/css/forms.scss
+++ b/client/webserver/site/src/css/forms.scss
@@ -1,5 +1,5 @@
 #regAssetForm {
-  position: relative;
+  max-width: 425px;
 
   div.reg-asset-allmkts {
     min-width: 210px;
--- a/client/webserver/site/src/css/forms.scss
+++ b/client/webserver/site/src/css/forms.scss
@@ -1,5 +1,7 @@
 #regAssetForm {
-  max-width: 425px;
+  #whatsabond {
+    max-width: 425px;
+  }
 
   div.reg-asset-allmkts {
     min-width: 210px;
--- a/client/webserver/site/src/js/settings.ts
+++ b/client/webserver/site/src/js/settings.ts
@@ -80,12 +80,12 @@ export default class SettingsPage extends BasePage {
       const wallet = asset.wallet
       if (wallet) {
         const bondAsset = this.currentDEX.bondAssets[asset.symbol]
-        if (wallet.synced && wallet.balance.available > bondAsset.amount) {
+        const bondsFeeBuffer = await this.getBondsFeeBuffer(assetID, page.regAssetForm)
+        if (wallet.synced && wallet.balance.available >= 2 * bondAsset.amount + bondsFeeBuffer) {
           this.animateConfirmForm(page.regAssetForm)
           return
         }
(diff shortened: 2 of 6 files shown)


client/webserver/site/src/css/forms.scss
```diff
@@ -1,5 +1,5 @@
 #regAssetForm {
-  position: relative;
+  max-width: 425px;
```
buck54321, line 2: This is causing new wrapping in the lot size column of the markets list, the net effect being only one market row is shown without scrolling instead of two.

before:
image

after:
image


Looks like it wants to grow to about 430px. I do think that 425 should be a reasonable max width for form elements, so we probably need some layout work here. For now, can you just assign the max width to the new whatsabond element? Putting `text-nowrap` on some elements seems to work too.

buck54321 (changes requested):

client/webserver/site/src/css/forms.scss
```diff
@@ -1,5 +1,7 @@
 #regAssetForm {
-  position: relative;
+  #whatsabond {
+    max-width: 425px;
```
chappjc, line 3: Looking good with latest specificity:

image

client/webserver/site/src/js/settings.ts
```diff
@@ -80,12 +80,13 @@ export default class SettingsPage extends BasePage {
...
-        if (wallet.synced && wallet.balance.available > bondAsset.amount) {
+        const bondsFeeBuffer = await this.getBondsFeeBuffer(assetID, page.regAssetForm)
+        if (wallet.synced && wallet.balance.available >= 2 * bondAsset.amount + bondsFeeBuffer) {
           this.animateConfirmForm(page.regAssetForm)
           return
         }
-        const txFee = await this.getRegistrationTxFeeEstimate(assetID, page.regAssetForm)
-        this.walletWaitForm.setWallet(wallet, txFee)
+        this.walletWaitForm.setWallet(wallet, bondsFeeBuffer)
+        this.currentForm = page.walletWait
```
chappjc, line 89: Not really the solution for master (`this.currentForm`), but this PR is going to cherry-pick to release-v0.6

client/webserver/site/src/js/settings.ts
```diff
@@ -80,12 +80,13 @@ export default class SettingsPage extends BasePage {
       const wallet = asset.wallet
       if (wallet) {
         const bondAsset = this.currentDEX.bondAssets[asset.symbol]
-        if (wallet.synced && wallet.balance.available > bondAsset.amount) {
+        const bondsFeeBuffer = await this.getBondsFeeBuffer(assetID, page.regAssetForm)
+        if (wallet.synced && wallet.balance.available >= 2 * bondAsset.amount + bondsFeeBuffer) {
           this.animateConfirmForm(page.regAssetForm)
           return
         }
-        const txFee = await this.getRegistrationTxFeeEstimate(assetID, page.regAssetForm)
```
chappjc, line 87: We can probably remove the webserver's handler for this now, but will get to that later just in case.

buck54321 (approved):

PR #2397, opened 2023-06-16
client: Improve 'Return all coins' processing

dev-warrior777:
`ReturnCoins(nil)` implements a sequence that includes `lockunspent true null`

There may not be 2 params for this. Bitcoin accepts `lockunspent true` so that is a valid input to this request. And although bitcoin accepts `lockunspent true null` other btc clones may not.

Bitcoin parameter checking changes by 0.16 to ignore this null. So any clones < bitcoin 0.16 *may* not allow the null parameter and error.

Known clones that would error: Zcash, Doge, Firo
image


chappjc:
> Bitcoin accepts `lockunspent true` so that is a valid input to this request

Has it always, back to 0.21?

What about others like likecoind bitcoin-cash-node etc?

chappjc:
Excellent, thanks for verifying!

client/asset/btc/rpcclient.go
```diff
@@ -495,10 +495,31 @@ func (wc *rpcClient) listUnspent() ([]*ListUnspentResult, error) {
...
+	if !unlock {
+		wc.log.Warn("attempting to lock all unspent outputs 'lockunspent false'")
+	}
+	var rpcReturn bool
+	err := wc.call(methodLockUnspent, anylist{unlock}, &rpcReturn)
+	if err != nil {
+		return fmt.Errorf("'lockunspent %v': %v: ", unlock, err)
+	}
+	if !rpcReturn {
+		return fmt.Errorf("'lockunspent %v' unsuccessful: %v: RPC returns %v", unlock, err, rpcReturn)
```
chappjc, line 511: `err` is always `nil` here so I don't see a reason to include it's value in the message

dev-warrior777:
image
 The exposure is not great. And DCR could be ignored.

The changes I did in `btc_test.go` were because when I ran the `run_tests.sh` the test failed .. then I tried to improve the bitcoin virtual rpc there.

client/asset/btc/rpcclient.go
```diff
@@ -495,10 +495,31 @@ func (wc *rpcClient) listUnspent() ([]*ListUnspentResult, error) {
...
+
 // lockUnspent locks and unlocks outputs for spending. An output that is part of
 // an order, but not yet spent, should be locked until spent or until the order
 // is canceled or fails.
 func (wc *rpcClient) lockUnspent(unlock bool, ops []*output) error {
+	if ops == nil {
+		return wc.lockAllUnspent(unlock)
+	}
 	var rpcops []*RPCOutpoint // To clear all, this must be nil->null, not empty slice.
 	for _, op := range ops {
```
ukane-philemon, line 524: @chappjc, please correct me if I'm wrong, but since we are now performing a nil check earlier, I believe this comment is no longer applicable.

dev-warrior777: I agree but thought it not my place to delete it .. although now I could if you like

ukane-philemon: > I agree but thought it not my place to delete it .. although now I could if you like

I think the `ops` parameter should own this comment instead of `rpcops`.

dev-warrior777: Then since there is no comment at function level, and the comment is never now going to be true would you agree that I should remove it. I removed on my local and all looks sensible to me.

dev-warrior777: Would referencing this doc (from the `btc_test.go` test case) in `rpcClient.lockUnspent`or the function level doc of `rpcClient.lockUnspent` be useful? Keeping the main path clutter free

image

JoeGruffins (approved):

client/asset/btc/rpcclient.go
dev-warrior777: Fixed in https://github.com/decred/dcrdex/pull/2397/commits/b7354fc99d3ada4141e9176697ec6a0b928b0f4d
Regularized comments as per approval

dev-warrior777:
I think all requested changes were done. I am only revisiting this because there was a merge conflict in `client/asset/btc_test.go` when I fetch to my local

dev-warrior777:
This https://github.com/decred/dcrdex/pull/2397/commits/ac9747f81f2c831ffd5cbdeb0790a0e3bebf0926 fixes Issue #2419

PR #2401, opened 2023-06-16
ui: add helper function on SettingsPage to ensure currentForm is updated

buck54321:
Fixes the bug from #2400 too.

### DIFF
--- a/client/webserver/site/src/js/settings.ts
+++ b/client/webserver/site/src/js/settings.ts
@@ -86,14 +86,12 @@ export default class SettingsPage extends BasePage {
           return
         }
         this.walletWaitForm.setWallet(wallet, bondsFeeBuffer)
-        this.currentForm = page.walletWait
-        forms.slideSwap(page.regAssetForm, page.walletWait)
+        this.slideSwap(page.walletWait)
         return
       }
 
       this.newWalletForm.setAsset(assetID)
-      this.currentForm = page.newWalletForm
-      forms.slideSwap(page.regAssetForm, page.newWalletForm)
+      this.slideSwap(page.newWalletForm)
     })
 
     // Approve fee payment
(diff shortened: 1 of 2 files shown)


client/webserver/site/src/js/settings.ts
```diff
@@ -158,6 +157,15 @@ export default class SettingsPage extends BasePage {
     })
   }
 
+  /*
+   * slideSwap animates the replacement of the currently shown form with the
+   * newForm and sets this.currentForm.
+   */
+  slideSwap (newForm: PageElement) {
+    forms.slideSwap(this.currentForm, newForm)
```
ukane-philemon, line 165: We should `await` here, no?

buck54321: I didn't think so. What's your reasoning?

ukane-philemon: > I didn't think so. What's your reasoning?

`forms.slideSwap(..., ...)` returns a `Promise<void>`.

ukane-philemon: After a close look at our use of `forms.slideSwap`, we definitely should `await`. `slideSwap` will have become an `async` function so we can `await` here. All current callers are `async` functions too but we don't need to `await` when we use `this.slideSwap` because we are not expecting a return value. Other use of `forms.slideSwap` should  `await` if they are used in `async` functions.

martonp: You need `await` if you want to ensure that `this.currentForm = newForm` is only set after the animation is completed.

chappjc (approved):
OK, needs resolve.

chappjc:
Still needs resolve, but also good to merge.

chappjc:
squashed in busted caches

PR #2405, opened 2023-06-19
client/core: ensure redeem is accepted before confirm/complete match

chappjc:
In certain cases, the redeem confirmation status was skipping the retries of the `'redeem'` request.  This was happening because `confirmRedemption` would advance the status to confirmed regardless of whether there was a stored `RedeemSig`, which indicates that the redeem request was accepted by the server.
This fixes the issue by <s>adding the `RedeemSig` check to `shouldConfirmRedemption`, which allows `resendPendingRequests` when they are needed.</s> only advancing to `MatchConfirmed` if both (a) match is confirmed, and (b) server has received and ack'd our `redeem` request, which is necessary to complete our role in the swap dance.

### DIFF
--- a/client/core/trade.go
+++ b/client/core/trade.go
@@ -92,8 +92,9 @@ type matchTracker struct {
 
 	// confirmRedemptionNumTries is just used for logging.
 	confirmRedemptionNumTries int
-	// redemptionConfs and redemptionConfsReq are set while the redemption
-	// confirmation process is running.
+	// redemptionConfs and redemptionConfsReq are updated while the redemption
+	// confirmation process is running. Their values are not updated after the
+	// match reaches MatchConfirmed status.
 	redemptionConfs    uint64
 	redemptionConfsReq uint64
 
@@ -2094,7 +2095,7 @@ func (c *Core) resendPendingRequests(t *trackedTrade) {
 			swapCoinID = proof.MakerSwap
 		case side == order.Taker && status == order.TakerSwapCast:
 			swapCoinID = proof.TakerSwap
-		case side == order.Maker && status == order.MakerRedeemed:
+		case side == order.Maker && status >= order.MakerRedeemed:
 			redeemCoinID = proof.MakerRedeem
 		case side == order.Taker && status >= order.MatchComplete:
 			redeemCoinID = proof.TakerRedeem
@@ -2819,11 +2820,20 @@ func (c *Core) sendRedeemAsync(t *trackedTrade, match *matchTracker, coinID, sec
 		if match.Side == order.Maker && match.Status < order.MatchComplete {
 			// As maker, this is the end. However, this diverges from server,
 			// which still needs taker's redeem.
-			match.Status = order.MatchComplete
+			if conf := match.redemptionConfs; conf > 0 && conf >= match.redemptionConfsReq {
+				match.Status = order.MatchConfirmed // redeem tx already confirmed before redeem request accepted by server
+			} else {
+				match.Status = order.MatchComplete
+			}
 		}
 		err = t.db.UpdateMatch(&match.MetaMatch)
 		if err != nil {
 			err = fmt.Errorf("error storing redeem ack sig in database: %v", err)
+		}
+		if match.Status == order.MatchConfirmed {
+			subject, details := t.formatDetails(TopicRedemptionConfirmed, match.token(), t.token())
+			note := newMatchNote(TopicRedemptionConfirmed, subject, details, db.Success, t, match)
+			t.notify(note)
 		}
 		t.mtx.Unlock()
 	}()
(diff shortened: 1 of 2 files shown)


client/core/trade.go
```diff
@@ -2853,6 +2856,11 @@ func (t *trackedTrade) redeemFee() uint64 {
 // This method accesses match fields and MUST be called with the trackedTrade
 // mutex lock held for writes.
 func (t *trackedTrade) confirmRedemption(match *matchTracker) {
+	if len(match.MetaData.Proof.Auth.RedeemSig) == 0 { // shouldConfirmRedemption should prevent this
+		t.dc.log.Warnf("Not confirming match %v before it is reported to server", match)
+		return // let resendPendingRequests have it
+	}
+
```
JoeGruffins, line 2863: This prevents `ConfirmRedemption` from happening until the server can be contacted? Maybe the wallet needs to send a new transaction and that is time sensitive. Is it possible to allow that to continue anyway?

Also in the unlikely event the server disappears, would this trade stay unconfirmed forever?

chappjc: Good point.  Will have to approach this differently, probably using the self governed state we added previously to identify when to confirm regardless

martonp: Even if the trade is not self governed, the server may not be able to find the redemption due to an issue for which the redemption needs to be resubmitted. I don't think the redemption confirmation check should ever have to wait on the server.

chappjc: On a side note, our `ConfirmRedemption` implementation for BTC is very unlikely to work if it gets to "resubmitting" the transaction.  In most case, you probably just need to rebroadcast the original. Submitting a new one (different txid) is very likely to be rejected by the network as a double spend.  Even if it is not rejected, the original one could still be mined (or maybe already was and we just didn't see it for whatever reasons) and we'd never be able to confirm anything.

The pathological case I was addressing with this PR was actually needing to just rebroadcast the original (or wait for it to propagate / rebroadcast automatically in time).  The redeem request to the server was failing because server couldn't see it.  Ultimately it was mined on a different best block that the server had, and it took a reorg for it to land on the same chain the server was following.  But since the match was considered confirmed, the client never tried to send the request again.
It's worse on 0.6 because BTC and DCR confirm automatically (no checks).

martonp (commented):
LGTM. `shouldConfirmRedemption` could return false if `match.redemptionConfs > match.redemptionConfsReq`, but nbd.

chappjc:
> LGTM. `shouldConfirmRedemption` could return false if `match.redemptionConfs > match.redemptionConfsReq`, but nbd.

Will you walk me through this?  I might be missing a detail.

Summary of the changes in current PR:

- `sendRedeemAsync` sets status to `MatchComplete`, or if confs already checked and `confs >= req`, `MatchConfirmed`
- `confirmRedemption` sets `match.redemptionConfs, match.redemptionConfsReq`.  If there's also a `RedeemSig` (or the trade is self-governed), it sets to `MatchConfirmed`

`shouldConfirmRedemption` is unchanged and will say false if already `MatchConfirmed`, which is correct since that state is only when confs >= req, set from a previous `confirmRedemption` call.  So it will keep updating confs until either of the above sets it to  `MatchConfirmed`

martonp:
>  Is this last case of extra updates past req what you meant?

Yes. I guess it's better this way though.. the UI will be kept up to date.

JoeGruffins:
I'm not sure but I think inside of `confirmRedemption` if `match.redemptionConfs >= match.redemptionConfsReq` but we are waiting on `(len(match.MetaData.Proof.Auth.RedeemSig) > 0 || t.isSelfGoverned())` a lot of the stuff in `confirmRedemption` can be skipped. There is no need to continue doing `toWallet.Wallet.ConfirmRedemption` especially.

Also, I guess setting `match.Status = order.MatchConfirmed` inside of `sendRedeemAsync` it will skip the notification on `redemptionConfirmed` in `confirmRedemption`. Does `sendRedeemAsync` need to set the status to confirmed or can that not just be left to `confirmRedemption`?

chappjc:
> I'm not sure but I think inside of `confirmRedemption` if `match.redemptionConfs >= match.redemptionConfsReq` but we are waiting on `(len(match.MetaData.Proof.Auth.RedeemSig) > 0 || t.isSelfGoverned())` a lot of the stuff in `confirmRedemption` can be skipped. There is no need to continue doing `toWallet.Wallet.ConfirmRedemption` especially.

The idea was to just keep updating the confirmation count, but you and @martonp both found this notable, so will change it.  I had removed a short-circuit condition from a previous iteration: https://github.com/decred/dcrdex/compare/d0530fddea72201fe2cbe1ed01155823b0de356d..c17079563ceeaf401bd6c89399dfe88cff402d83

> Also, I guess setting `match.Status = order.MatchConfirmed` inside of `sendRedeemAsync` it will skip the notification on `redemptionConfirmed` in `confirmRedemption`. Does `sendRedeemAsync` need to set the status to confirmed or can that not just be left to `confirmRedemption`?

I was trying to skip having to wait for another tick, but I think it'll do it in the same tick anyway.  Updating.

JoeGruffins (approved):
Thanks for answering our nits.

chappjc:
> Thanks for answering our nits.

Not nits.  I repeatedly missed some important details on this one.  Thanks for reviewing.

buck54321 (approved):
Just a rebase needed.

martonp (approved):

PR #2410, opened 2023-06-22
update cache_utilities for mac

buck54321:


client/webserver/site/cache_utilities.bash
```diff
@@ -53,7 +59,11 @@ hashjssrc () { hashdir "${JS_DIR}" ; }
 # hashjssrc hashes the compiled js.
 hashjsdist () {
     cp "${JS_FILE}" js.tmp
-    sed -i 's/commitHash="[^"]*"//' js.tmp
+    if [[ "$OSTYPE" == "darwin"* ]]; then
+        sed -i '' 's/commitHash="[^"]*"//' js.tmp
+    else
+        sed -i 's/commitHash="[^"]*"//' js.tmp
```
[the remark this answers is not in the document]
buck54321: Oh yeah.

client/webserver/site/cache_utilities.bash
```diff
@@ -5,14 +5,20 @@ CSS_FILE=${SITE_DIR}/dist/style.css
 JS_DIR=${SITE_DIR}/src/js
 JS_FILE=${SITE_DIR}/dist/entry.js
 
-hashfile () { sha1sum $1 | cut -d " " -f1 ; }
+if [[ "$OSTYPE" == "darwin"* ]]; then
+   hashfunc="shasum"
+else
+   hashfunc="sha1sum"
+fi
```
chappjc, line 12: Some of the decred folks seem to prefer using openssl as a portable approach.  `openssl sha1 -r blah.txt`
The `sha*sum` functions seem be be part of typical "coreutils" packags in NIX systems.  I've always felt like that's more common than having openssl installed personally, but it does make the app consistent so there's no question about syntax or output format.

client/webserver/site/cache_utilities.bash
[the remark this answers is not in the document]
buck54321: I'm seeing some weird behavior with this recommendation implemented. `jsdistbuster` works correctly when I run it directly, but when run in `bust_caches.sh` it sets the wrong value.

client/webserver/site/cache_utilities.bash
ukane-philemon: > I wonder is openssl available out of the box on Mac?

yh, but the latest macOS, `openssl version` shows Apple uses another flavour `LibreSSL`.

client/webserver/site/cache_utilities.bash
```diff
@@ -52,10 +52,9 @@ hashjssrc () { hashdir "${JS_DIR}" ; }
 
 # hashjssrc hashes the compiled js.
 hashjsdist () {
-    cp "${JS_FILE}" js.tmp
-    sed -i 's/commitHash="[^"]*"//' js.tmp
-    HASH=$(hashfile js.tmp | cut -c1-8)
-    rm js.tmp
+    sed -i.tmp 's/commitHash="[^"]*"//' "${JS_FILE}"
+    HASH=$(hashfile "${JS_FILE}" | cut -c1-8)
+    rm "${JS_FILE}.tmp"
```
buck54321, line 57: Oh wait. This won't work. We **must** create a copy and discard it, and this edits the original in place. Doing `sed -i.tmp` doesn't create a new file to edit in place, it uses that path for a backup of the original.

chappjc: > Oh wait. This won't work. We **must** create a copy and discard it, and this edits the original in place. Doing `sed -i.tmp` doesn't create a new file to edit in place, it uses that path for a backup of the original.

Oh hah, got that backwards!

chappjc:
LGTM, but I don't have access to a Mac currently.

ukane-philemon:
> LGTM, but I don't have access to a Mac currently.

I can branch off this and test gh CI run.

chappjc:
> > LGTM, but I don't have access to a Mac currently.
> 
> I can branch off this and test gh CI run.

I mean I think this PR is g2g if someone else with a Mac can just check that `./bust_caches.sh` runs on their machine.

ukane-philemon:
> I mean I think this PR is g2g if someone else with a Mac can just check that ./bust_caches.sh runs on their machine.

Okay, testing.

martonp (approved):

chappjc:
> But [fails CI ](https://github.com/ukane-philemon/dcrdex/actions/runs/5360380751/jobs/9725174655)run even after busting for this change [b0cbc74](https://github.com/decred/dcrdex/commit/b0cbc74d48da04b758cc6205ecbbdb1019abaf24)

So the buster you get don't work for CI?

buck54321:
@ukaine tested same on a Mac and is passing. If you pull this last commit [aac3a2b](https://github.com/decred/dcrdex/pull/2410/commits/aac3a2b4aca57d63158d2cf9220184452c5665a5), go to the site directory, and run

```Bash
npm run build
source cache_utilities.bash
jsdistbuster
```

what do you get?

ukane-philemon:
> So the buster values you get don't work for CI?

I guess so. I don't know if I was supposed to build before busting. I simply made changes to the js file, ran the bust script, and committed, then I pushed the changes to test CI but it failed.

chappjc:
> > So the buster values you get don't work for CI?
> 
> I guess so. I don't know if I was supposed to build before busting. I simply made changes to the js file, ran the bust script, and committed, then I pushed the changes to test CI but it failed.

If you use bust_cashes.sh, it should build for you.

https://github.com/decred/dcrdex/blob/aac3a2b4aca57d63158d2cf9220184452c5665a5/client/webserver/site/bust_caches.sh

Which piece of the buster was wrong, first, second or both?

chappjc:
Never forget to do `npm ci` before `npm run build` BTW.  That can change your output easily.  Never do `npm install`.

ukane-philemon:
> If you use bust_cashes.sh, it should build for you.

Yeah, that was what I relied on.

> Never do npm install

Yh, I never do.

chappjc:
> > If you use bust_cashes.sh, it should build for you.
> 
> Yeah, that was what I relied on.

Dang.  Try removing your node_modules folder manually I guess.

Also, I assume you're running either Node 18 or 20?

chappjc:
> Which piece of the buster was wrong, first, second or both?

Also this @ukane-philemon.  If you run on the current commit here in buck's PR, which piece of the buster does it tweak?

client/webserver/site/src/html/bodybuilder.tmpl
```diff
@@ -104,7 +104,7 @@
 {{end}}
 
 {{define "bottom"}}
-<script src="/js/entry.js?v=88986a77|0cdffb94"></script>
+<script src="/js/entry.js?v=ebcfcdea|a99362bc"></script>
```
chappjc, line 107: Getting the same values locally.

@ukane-philemon Do you have `core.autocrlf` set to something with git?

ukane-philemon:
> Also, I assume you're running either Node 18 or 20?

I'm using node 20.3.1.

ukane-philemon:
> > Which piece of the buster was wrong, first, second or both?
> 
> Also this @ukane-philemon. If you run on the current commit here in buck's PR, which piece of the buster does it tweak?

It seems the second part is the black sheep (`ebcfcdea|5e62b35d`) from https://github.com/decred/dcrdex/commit/b0cbc74d48da04b758cc6205ecbbdb1019abaf24

client/webserver/site/src/html/bodybuilder.tmpl
ukane-philemon: > @ukane-philemon Do you have core.autocrlf set to something with git?

No, I don't.

chappjc:
> It seems the second part is the black sheep (`ebcfcdea|5e62b35d`) from [b0cbc74](https://github.com/decred/dcrdex/commit/b0cbc74d48da04b758cc6205ecbbdb1019abaf24)

So at least we know the input part is fine, which means the `hashfile` and `hashdir` functions are OK for you.  So it could be an issue with `hashjsdist`.  Perhaps hack into `cache_utilities.bash` and add some temporary variables and echo statements, and comment the `rm` line so you can inspect and hash it manually after.  Do a diff of the js file to see if the sed worked as intended.

Anyway, you don't need to test CI with your own branch.  Just run bust_caches.sh on this branch until the values don't change from what's already committed.

chappjc:
Another thing to try is to change "sed" to "perl -pe" to see if your sed is the culprit.

ukane-philemon:
> ust run bust_caches.sh on this branch until the values don't change from what's already committed.

Alright.

ukane-philemon:
> 

`perl -pe` worked.

ukane-philemon (approved):
G2G

chappjc (approved):

PR #2411, opened 2023-06-23
server/auth: base order limit scales with bond strength

chappjc:
Relates to https://github.com/orgs/decred/projects/2?pane=issue&itemId=2420755

This is a step toward considering bond strength for user limits.

This also addresses a practical issue for v0.6 users where they can have a non-zero tier, but also have a low or even negative `UserSettlingLimit` because bond strength/tier is was not considered in computation of the limit.  This happens when a user fails swaps causing their tier to drop, and then the "target tier" maintenance code posts additional bond, pushing their account back into the green.  In this case, they have a positive tier and are allowed to submit orders, but their orders get blocked because their total bonding tier/strength was not considered in the order size checks.

The essence of the fix is to multiply the "base" order limit size, which is the limit for users with no swap history, by the user's bond strength (plus one for legacy fee users).  This is still balanced against swap history, with the same alternate method that considers swap amounts.  This solution does not yet resolve the referenced project item because we might consider dropping the alternate swap amount weighting scheme or directly using the user's tier (bond strength offset by other violations and successes).  Doing that requires more careful analysis to ensure highly active users do _not_ have their limits unexpectedly reduced.

A second change is clamping the computed lot limit on both ends.  Previously it was only clamped at the high end at `auth.absTakerLotLimit`.  Now it ensures at least `currentLotSize` for an account with non-zero tier.  Possibly this should be omitted, but it feels right.

### DIFF
--- a/server/auth/auth.go
+++ b/server/auth/auth.go
@@ -790,14 +790,50 @@ const (
 
 // UserSettlingLimit returns a user's settling amount limit for the given market
 // in units of the base asset. The limit may be negative for accounts with poor
-// swap history.
+// swap history. The user must be connected.
 func (auth *AuthManager) UserSettlingLimit(user account.AccountID, mkt *dex.MarketInfo) int64 {
+	// WIP: https://github.com/orgs/decred/projects/2?pane=issue&itemId=2420755
+	// tier is a balance of bonding strength (bondTier) and user score, where
+	// the latter is computed from all violations and match outcomes. See the
+	// integrateOutcomes and tier methods. However, for order size limits, we
+	// consider swap outcome amounts, and consider only swap outcomes (ignore
+	// preimage and cancel rate violations). Thus, we scale the base limit by
+	// bond strength, and balance that with weighted swap history.
+
 	currentLotSize := int64(mkt.LotSize)
-	base := currentLotSize * auth.initTakerLotLimit
+	baseLimit := currentLotSize * auth.initTakerLotLimit
+
+	// The base limit scales with bond strength (plus one if legacy fee paid).
+	// This is particularly important for an account that has offset their
+	// violations with additional bond to gain a positive tier.
+	client := auth.user(user)
+	if client == nil {
+		return 0 // offline
+	}
+	client.mtx.Lock()
+	tier := client.tier
+	baseScale := client.bondTier()
+	if client.legacyFeePaid {
+		baseScale++
+	}
+	client.mtx.Unlock()
+
+	if tier < 1 { // but order router should have checked this
+		return 0
+	}
+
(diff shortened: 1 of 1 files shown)


server/auth/auth.go
```diff
@@ -790,14 +790,50 @@ const (
...
+	if client == nil {
+		return 0 // offline
+	}
+	client.mtx.Lock()
+	tier := client.tier
+	baseScale := client.bondTier()
+	if client.legacyFeePaid {
+		baseScale++
+	}
+	client.mtx.Unlock()
```
ukane-philemon, line 819: Can we do `RLock` instead?

server/auth/auth.go
```diff
@@ -790,14 +790,50 @@ const (
...
+
+	if tier < 1 { // but order router should have checked this
+		return 0
+	}
```
ukane-philemon, line 823: Can we not return a limit for an online user that has paid a legacy fee but currently have zero tiers?

server/auth/auth.go
chappjc: Possible.  Would have to convert the mutex type.  Will see if it's beneficial

server/auth/auth.go
chappjc: Zero tier means not allowed to trade, by definition.  The incoming order processing pipeline should already have prevented this, but it's here for good measure.

Legacy fee users get a tier increment already, so if such a user is not positive tier, that means they have to many violations.

ukane-philemon: > Legacy fee users get a tier increment already, so if such a user is not positive tier, that means they have too many violations.

Okay, thanks for the clarification.

server/auth/auth.go
chappjc: Have researched use patterns, and with the current design of `(*Market).Run`, there is no concurrent use of `UserSettlingLimit` that would make an `RWMutex` useful here (all the other locks require a full write lock).

chappjc:
Tested on simnet with bond tier 1 and 2 accounts, confirmed higher *base* limit for higher bonding.  Did not test with a legacy fee account, but the logic that applies `baseScale++` there is trivial.

This will fix krill's issue of a negative limit encountered despite being in a positive trading tier. (multiple active bonds automatically posted to overcome some penalization).

server/auth/auth.go
ukane-philemon: okay, np.

martonp (approved):

JoeGruffins (approved):

chappjc:
Going to port this to `release-v0.6` since it's a problematic bug, but we aren't going to fast-track a patch release since it's server-side only, where things are generally run from source.

PR #2416, opened 2023-06-27
client/mm: Update Basic Market Maker 

martonp:
This updates the basic market maker strategy to place multiple orders at various distances from the basis price instead of just one. It uses the new MultiTrade function to do this.

The basis price calculation is also updated to use fiat rates as a fallback if there are no trades on the dex or an oracle rate.

Something still missing from this is adjusting the rates based on how many previous buys/sells have been matched. There
needs to be a skew setting that places orders from the side with more previous matches farther from the basis price.

### DIFF
--- a/client/mm/mm_basic.go
+++ b/client/mm/mm_basic.go
@@ -8,9 +8,9 @@ import (
 	"errors"
 	"fmt"
 	"math"
-	"sort"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"decred.org/dcrdex/client/core"
 	"decred.org/dcrdex/client/orderbook"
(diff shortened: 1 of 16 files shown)


client/mm/mm_basic.go
```diff
@@ -638,36 +839,74 @@ func (m *basicMarketMaker) run() {
...
-	wg.Wait()
+	// Periodically update asset fee rates
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		refreshTime := time.Minute * 10
+		for {
+			select {
+			case <-time.NewTimer(refreshTime).C:
+				err := m.updateFeeRates()
```
buck54321, line 850: The only arguments to `MaxFundingFees` are `numTrades` and `options`. I don't see a reason to run this on a timer. We need to update these only if one of those values changes.

client/mm/mm_basic.go
```diff
@@ -83,7 +98,32 @@ type MarketMakingConfig struct {
...
+	// operation of this market maker, orders will be frequently placed and
+	// canceled as the price moves. If the price moves upward and there is
+	// no split buffer, the wallet may need to create a new split transaction
+	// to fund the new orders.
+	//
+	// The split buffer is a percentage of the total amount required for an
+	// order. For example if the order requires 0.1 BTC and the split buffer
+	// is 5, the wallet will create a split transaction with an output of
+	// 0.105 BTC.
+	SplitBuffer *uint64 `json:"splitBuffer"`
```
buck54321, line 122: This is a little off for me. The option to use splits should be a per-asset option. And EVM assets won't use this at all. These really are better represented using the `settings map[string]string` system that we had in place for asset-specific options.

What do you think of https://github.com/decred/dcrdex/compare/master...buck54321:multifundopts ?

client/mm/mm_basic.go
martonp: I've updated `MaxFundingFees` to use the server's `MaxFeeRate` instead of the wallet's fee rate limit.

client/mm/mm_basic.go
martonp: It is better like this, but it would be nice to block people from setting a buffer on the quote asset.

buck54321: Add a `QuoteOnly bool` field to `ConfigOption`?

martonp: I thought of that too.. kind of out of place over there since it's just specific to a market making strategy. Since assets share a lot of these configs, maybe the keys could be defined in `client/asset/interface.go`, and then based on the key, the market maker could make sure it's always 0 for the base asset.

buck54321: > the keys could be defined in client/asset/interface.go, and then based on the key, the market maker could make sure it's always 0 for the base asset.

That a step backwards, imo. Let the assets decide. I'm all for adding fields to `ConfigOption`. It's a free-for-all there. Use `omitempty` so we're not bloating the results.

client/mm/mm_basic.go
```diff
@@ -133,124 +191,164 @@ func steppedRate(r, step uint64) uint64 {
...
+		return steppedRate(uint64(basisPrice), mkt.RateStep)
+	}
+
+	// TODO: add a configuration to turn off use of fiat rate?
+	if fiatRate > 0 {
+		return steppedRate(fiatRate, mkt.RateStep)
+	}
+
+	if cfg.EmptyMarketRate > 0 {
+		return mkt.ConventionalRateToMsg(cfg.EmptyMarketRate)
```
buck54321, line 336: `steppedRate(cfg.EmptyMarketRate, mkt.RateStep)`? I don't see us validating `EmptyMarketRate` anywhere else.

buck54321 (changes requested):

buck54321 (approved):
Not necessarily from this PR, but when running `startmarketmaking` using `dexcctl` the request hangs and an error is logged 

`unable to send request: Post "https://127.0.0.3:5757": local error: tls: bad record MAC`

but the market maker **is** started.

Please look into it and pick up any fixes in the arb PR.

Issue #2303, opened 2023-04-15
Market maker testing framework


martonp:
Loadbot can be expanded with new program types that simulate various market conditions. Then, various bot configurations could be tested to see how they fare in those conditions.

JoeGruffins:
I will work on this.

Some possible new programs for the bots: 
- sideways - This would oscillate randomly between a range. 
- trending or increasing/decreasing price - I guess going up or down forever? Can go back the other way once it hits a point?
- volatile - All over the place?

Does this need to test arbitrage in some way? I guess it would be possible to set up mock exchanges although that may be involved.

Should the market maker clients also be automated?

JoeGruffins:
So yes to also testing arbitrage? Using actual exchange rates does make sense. The testing framework is only for simnet though right?

martonp:
Yes for arbitrage, and yes only simnet. I think it can be two separate PRs though.

PR #2420, opened 2023-07-01
loadbot: Add live price oracles.

JoeGruffins:
depends on #2409
part of #2303

Add a boolean livemidgap that allows loadbot start with a live rate
based on mainnet coin USD values. This also alters ts to keep the midgap
close to live rates rather than random values.


JoeGruffins:
> Trying to run on dcr_btc

Ok, will look into it. I have been testing on btc_eth because the dcr bots seem to freeze the dcr harness regularly for me. It was incorrect not to test dcr_btc as well, regardless.

JoeGruffins:
I think this fixes it https://github.com/decred/dcrdex/compare/84e6b4958cf05a5cc99b8763e2139f33f6fa2180..ff67113063d00e6759bf2ba1e28c56fca691f4f8

dex/testing/loadbot/loadbot.go
```diff
@@ -747,3 +757,55 @@ func symmetricWalletConfig(numCoins int, midGap uint64) (
...
+	as := map[uint32]*core.SupportedAsset{baseID: b, quoteID: q}
+	liveRates := core.FetchCoinpaprikaRates(ctx, log, as)
+	if len(liveRates) != 2 {
+		liveRates = core.FetchMessariRates(ctx, log, as)
+	}
+	if len(liveRates) != 2 {
+		liveRates = core.FetchDcrdataRates(ctx, log, as)
+	}
+	if len(liveRates) != 2 {
+		return 0.0, errors.New("unabel to get live rates")
```
buck54321, line 807: unabel

dex/testing/loadbot/loadbot.go
```diff
@@ -747,3 +757,55 @@ func symmetricWalletConfig(numCoins int, midGap uint64) (
...
+		return 0.0, fmt.Errorf("%d not a supported asset", quoteID)
+	}
+	as := map[uint32]*core.SupportedAsset{baseID: b, quoteID: q}
+	liveRates := core.FetchCoinpaprikaRates(ctx, log, as)
+	if len(liveRates) != 2 {
+		liveRates = core.FetchMessariRates(ctx, log, as)
+	}
+	if len(liveRates) != 2 {
+		liveRates = core.FetchDcrdataRates(ctx, log, as)
+	}
```
buck54321, line 805: Only on dcr_btc market, really.

dex/testing/eth/harness.sh
```diff
@@ -114,13 +114,28 @@ cat > "${NODES_ROOT}/genesis.json" <<EOF
...
+        "balance": "110000000000000000000000"
+    },
+    "946dfaB1AD7caCFeF77dE70ea68819a30acD4577": {
+        "balance": "110000000000000000000000"
+    },
+    "3e983c7e08d22CaB842Bd54111087E09C578b3c3": {
+        "balance": "110000000000000000000000"
+    },
+    "653445e4C2f48B370fE10e59940E5e9Cb853D03F": {
+        "balance": "110000000000000000000000"
```
buck54321, line 138: What's the deal with these again? They derive from a seed used in testing somewhere? Why not just give it all to alpha and request as needed?

buck54321 (changes requested):

dex/testing/eth/harness.sh
JoeGruffins: At least two are from the simnet testing program and I believe the other are the harness nodes. Sure can give all to alpha and send if that's less confusing. I should have commented better here either way.

dex/testing/loadbot/loadbot.go
JoeGruffins: fixed

dex/testing/loadbot/loadbot.go
JoeGruffins: checking if dcr is involved now

JoeGruffins:
I can't figure out what they all are. The simnet tests look ok as far as funding goes https://github.com/decred/dcrdex/compare/b63f7b29d2b20119b7c199c34e969e4781640da3..3242a5c95a8098fc6691805770daf794be591b97

However, it seems the the `nodeclient_harness_test.go` tests in client/assets/eth are not passing due to unexpected balance differences.

buck54321:
> Also setting the harness to mine every 2 seconds rather than 15.

Can we not? What problem are we solving?

buck54321:
> > Also setting the harness to mine every 2 seconds rather than 15.
> 
> Can we not? What problem are we solving?

Actually, I don't even see where you've done this.

JoeGruffins:
> Actually, I don't even see where you've done this.

Sorry, I meant while running those tests. I did not change for this pr.

buck54321 (approved):

PR #2425, opened 2023-07-04
client/btc: Improve ReturnCoins processing

dev-warrior777:
Improve processing for ReturnCoins.

[original pr](https://github.com/decred/dcrdex/pull/2397)

### DIFF
--- a/client/asset/btc/btc_test.go
+++ b/client/asset/btc/btc_test.go
@@ -413,15 +413,75 @@ func (c *tRawRequester) RawRequest(_ context.Context, method string, params []js
 			return json.Marshal(false)
 		}
 		coins := make([]*RPCOutpoint, 0)
-		_ = json.Unmarshal(params[1], &coins)
-		if string(params[0]) == "false" {
-			if c.lockedCoins != nil {
-				c.lockedCoins = append(c.lockedCoins, coins...)
-			} else {
-				c.lockedCoins = coins
+
+		// There may not be 2 params for this. Bitcoin accepts 'lockunspent true'
+		// so that is a valid input to this request. And although bitcoin accepts
+		// 'lockunspent true null' other btc clones may not.
+		// Bitcoin parameter checking changes by 0.16 to ignore this null. So any
+		// clones < bitcoin 0.16 *may* not allow the null parameter and error.
+		// Known clones that would error: Zcash, Doge, Firo
+
+		oneParam := func(param0 json.RawMessage) ([]byte, error) {
+			if string(param0) == "true" {
+				c.lockedCoins = nil
+				return json.Marshal(true)
 			}
+			return json.Marshal(true)
 		}
-		return json.Marshal(true)
+
+		switch len(params) {
+		case 1:
+			return oneParam(params[0])
+		case 2:
+			if string(params[1]) == "null" {
+				// Then is the 1 param case plus an unused 'null'
+				return oneParam(params[0])
+			}
+
+			err := json.Unmarshal(params[1], &coins)
+			if err != nil {
+				return json.Marshal(false)
+			}
+
+			if string(params[0]) == "false" {
+				var dup bool
+				var newLockedList []*RPCOutpoint
+				for _, coin := range coins {
+					dup = false
+					for _, lockedCoin := range c.lockedCoins {
+						if lockedCoin.TxID == coin.TxID && lockedCoin.Vout == coin.Vout {
+							dup = true
+							break
+						}
+					}
+					if !dup {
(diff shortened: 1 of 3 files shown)


JoeGruffins (approved):
Assuming nothing has changed from the other pr.

client/asset/btc/btc_test.go
```diff
@@ -413,15 +413,75 @@ func (c *tRawRequester) RawRequest(_ context.Context, method string, params []js
...
+					if lockedCoin.TxID == coin.TxID && lockedCoin.Vout == coin.Vout {
+						remove = true
+						break
+					}
+				}
+				if !remove {
+					newLockedList = append(newLockedList, lockedCoin)
+				}
 			}
+			c.lockedCoins = newLockedList
```
JoeGruffins, line 480: Small nit, you could avoid some code duplication by using a function. For example https://go.dev/play/p/TJ8noAdRQ8V

dev-warrior777:
> Assuming nothing has changed from the other pr.

The same exactly

client/asset/btc/btc_test.go
dev-warrior777: Thanks .. appreciated .. actually I _did_ use one function above: `oneParam` but did not max the potential code reuse in the looping there

client/asset/btc/rpcclient.go
```diff
@@ -495,11 +495,35 @@ func (wc *rpcClient) listUnspent() ([]*ListUnspentResult, error) {
...
+	}
+	return nil
+}
+
 // lockUnspent locks and unlocks outputs for spending. An output that is part of
 // an order, but not yet spent, should be locked until spent or until the order
 // is canceled or fails.
 func (wc *rpcClient) lockUnspent(unlock bool, ops []*output) error {
-	var rpcops []*RPCOutpoint // To clear all, this must be nil->null, not empty slice.
+	if ops == nil {
```
chappjc, line 523: comparing a slice to nil is often error prone.  Reason not to check `len(ops) == 0`?

dev-warrior777: Perhaps inexperience with golang.

dev-warrior777: There is probably more wrong with `ReturnCoins(nil)` anyway. It was first written against btc. But some older clones do not seem to work quite the same way. 

I see for at least one coin that even the more correct  `lockunspent true` does not work as I would expect. I sometimes see, for example, 3 coins locked -- then `lockunspent true` but instead of individual coins returned to unspent they are amalgamated into one - sometimes - hard to catch.

My current thinking is to Not use `lockunspent true` at all for all coins. Just take the less efficient path of iterating each coin and unlocking individually in a loop. It works for all clones as far as I can see from my testing some weeks ago; and is probably the more tested C++ code path.

We currently do not use the `ReturnCoins(nil)` version of the function much. I did a search before and the results are [above](https://github.com/decred/dcrdex/pull/2397#issuecomment-1594887959)

Switching to (finding then) iterating over locked coins seems non-breaking, then. I would leave the `btc_test.go` as is.

What do you think?

client/asset/btc/rpcclient.go
```diff
@@ -495,11 +495,35 @@ func (wc *rpcClient) listUnspent() ([]*ListUnspentResult, error) {
 	return unspents, wc.call(methodListUnspent, anylist{uint8(0)}, &unspents)
 }
 
+// lockAllUnspent unlocks all locked outputs for spending again. If unlock is
+// true btc will return 'true' if successful. If unlock is false btc will also
+// return 'true' but do nothing so a warning is issued below.
+//
+// See Also: btc_test.go: RawRequest case: methodLockUnspent
+func (wc *rpcClient) lockAllUnspent(unlock bool) error {
```
buck54321, line 503: Why accept an `unlock` argument at all? Also, why name the method `lock` when it **un**locks?

client/asset/btc/rpcclient.go
```diff
@@ -495,11 +495,35 @@ func (wc *rpcClient) listUnspent() ([]*ListUnspentResult, error) {
...
+	if !unlock {
+		wc.log.Warn("attempting to lock all unspent outputs 'lockunspent false'")
+	}
+	var rpcReturn bool
+	err := wc.call(methodLockUnspent, anylist{unlock}, &rpcReturn)
```
buck54321, line 508: Verified that this works for Dogecoin and Firo. Doesn't work for Zcash, but doesn't matter because #2553.

buck54321 (changes requested):

buck54321:
Reopen if it's important.

PR #2428, opened 2023-07-06
client/UI: Add pending bond message.

### LINKED ISSUE #2408: ui: misleading bond message
When starting up after being shut down for a while.

image

(Note also the typo `You` instead of `Your`)

The logic in **markets.ts** goes like...

```Typescript
if (market.dex.tier >= 1) {
  ...
} else if (this.hasPendingBonds()) {
  ...
} else {
  ...
  Doc.show(page.bondRequired)
}
```

But there must be a non-insignificant delay between startup and broadcast of new bonds, because the message persists for enough time to cause confusion. I think we need to check target tier instead of pending bond count or something.

JoeGruffins:
If target bond tier is more than zero but bonds have not been sent yet, inform the user of the state rather than incorrect message that they must change their bond settings.

Also updating the simnet setup script to use bonds and attempting to fix a couple bugs.

image

closes #2408

### DIFF
--- a/client/cmd/dexcctl/simnet-setup.sh
+++ b/client/cmd/dexcctl/simnet-setup.sh
@@ -89,7 +89,8 @@ if [ $RESTORING == "true" ]; then
 else
 
   echo registering with DEX
-  ./dexcctl -p abc --simnet register 127.0.0.1:17273 100000000 42 ~/dextest/dcrdex/rpc.cert
+  ./dexcctl -p abc --simnet login
+  ./dexcctl -p abc --simnet postbond 127.0.0.1:17273 1000000000 42 true ~/dextest/dcrdex/rpc.cert
 
   echo mining fee confirmation blocks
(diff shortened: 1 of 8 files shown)


JoeGruffins:
Is "sent shortly" good or should I have stuck with "posted shortly"

chappjc:
> Is "sent shortly" good or should I have stuck with "posted shortly"

Sent or better, created.  Agreement now is that post is confusing

JoeGruffins:
Changed sent and posted to created.

chappjc (approved):
LGTM.  Thanks for fixing the rpcserver docs and refundtx formatting.
login is needed even after init before postbond?

JoeGruffins:
> login is needed even after init before postbond?

It appears so. Without login...
```
$ ./dexcctl -p abc --simnet postbond 127.0.0.1:17273 1000000000 42 true ~/dextest/dcrdex/rpc.cert
bond key derivation failed: bondKeyIdx: not logged in
```

PR #2429, opened 2023-07-06
git: ignore dirs created when code run

dev-warrior777:
Sometimes a dir `regtest` is created when dex-tests/dex-code is run inside the repo tree. This pr adds
```
mainnet/
testnet/
regtest/
```
 to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,6 @@ client/asset/btc/electrum/example/wallet/wallet
 **/testdata/fuzz
 client/asset/eth/cmd/getgas/getgas
 client/cmd/dexc-desktop/pkg/installers
+mainnet/
+testnet/
+regtest/


chappjc:
> Sometimes a dir `regtest` is created when dex-tests/dex-code is run inside the repo tree.

Where specifically?

dev-warrior777:
> Where specificaly

If I run a `go test -v -count=1 ./...` in client/asset/firo` against the firo regtest harness a directory called `regtest` will appear in the same directory and an empty file inside at the end of the run
image

chappjc:
Thanks for clarifying.  Let's be more specific to avoid future surprises.  This glob should work but please check:

`client/asset/**/regtest` etc.

chappjc:
and/or have the test clean up the recycled-addrs.txt file

dev-warrior777:
> and/or have the test clean up the recycled-addrs.txt file

but it also happens in other places ... perhaps anywhere in the tree I did not check .. for now I will just close this

chappjc:
> > and/or have the test clean up the recycled-addrs.txt file
> 
> but it also happens in other places ... perhaps anywhere in the tree I did not check .. for now I will just close this

Where?  I don't want to blindly block three very generic folder names from git.

dev-warrior777:
Does it come from pr 2368

chappjc:
> Does it come from pr 2368

Yes

dev-warrior777:
>> Does it come from pr 2368

> Yes

Then what can I do to help ;-)

PR #2431, opened 2023-07-10
finish polygon

buck54321:
Generalize the Ethereum wallet and server backend. Prepare getgas utility and RPC compatibility testing. Add tokens to Polygon harness and get gas estimates. Update some front end token-related code. Deploy USDC swap contract for Polygon mainnet and get gas estimates. Update simnet trade tests.


chappjc (commented):
How do you use deploy?  Where do you send the fee gwei?  The deploy docs refer to "the wallet", but what wallet?  Does this require a dexc assetdb wallet for the asset?  I've used remix with web3 injection to deploy things as needed previously, so it's entirely independent.

Why is deploy.go in cleint/asset/eth instead of client/asset/eth/cmd/deploy?  That seems wrong.  If you want it reusable from tools other than cmd/deploy, maybe it should be in client/asset/eth/deploy.  Seems like it's just poluting client/asset/eth like it is currently.

The deploy binary is checked in.

buck54321:
> How do you use deploy?

Instructions for using **deploy** are at the top of *client/asset/eth/cmd/deploy/main.go*.

> Where do you send the fee gwei? The deploy docs refer to "the wallet", but what wallet? Does this require a dexc assetdb wallet for the asset?

See `IMPORTANT:` at the top of *deploy/main.go*. **deploy** uses the same wallet credentials as **getgas**. It offers similar auxiliary functions to **getgas** for estimating funding requirements and returning balance.

> I've used remix with web3 injection to deploy things as needed previously, so it's entirely independent.

What's the benefit of it being entirely independent? Is "web3 injection" just a webpage using the provider API by bouncing through a wallet extension? That's what is seems to say [here](https://www.doubloin.com/learn/what-is-injected-web3). 

Does Remix offer any advantages? For instance, I'm not sure how etherscan gets the metadata associated with the contract. For instance, [here](https://etherscan.io/address/0x8C17e4968B6903E1601be82Ca989c5B5E2c7b400#code) is the ethereum mainnet swap contract. How do they get the source code and ABI info? Does Remix do that?

> Why is deploy.go in cleint/asset/eth instead of client/asset/eth/cmd/deploy?

I can move it if you'd rather I export some stuff.

chappjc:
> > Why is deploy.go in cleint/asset/eth instead of client/asset/eth/cmd/deploy?
> 
> I can move it if you'd rather I export some stuff.

Don't bother.  I'd rather have some new tangential stuff in client/asset/eth (ContractDeployer) than unnecessary exports.

Everything is fine, no need to change anything.  For the sake of discussion though:

> > How do you use deploy?
> 
> Instructions for using **deploy** are at the top of _client/asset/eth/cmd/deploy/main.go_.

I had read that entire block comment, but didn't read into get gas.  Sorry.

> > I've used remix with web3 injection to deploy things as needed previously, so it's entirely independent.
> 
> What's the benefit of it being entirely independent?

I mainly mentioned that approach to explain why the proposed tool gave me pause. However, reasons why I personally prefer an approach for contract deployment that is separate:

(1) It's already available.  We don't have to make the package modifications and trade-offs above.  (2) It's interactive from compilation to deployment.  At a glance, I can't tell what compiled runtime bytecode deploy/main.go is going to deploy.  (It's the `Bin` in the pre-generated contract.go).  When I deploy a contract interactively, I be sure to set the solc version and optimization iterations, make sure it didn't have any warnings etc., enter the target erc20 address, set and sanity-check the fees. (3) I use the wallet of my choice to deploy (something with no other purpose than being a "deployer", not a dex wallet)

> Does Remix offer any advantages? I'm not sure how etherscan gets the metadata associated with the contract. For instance, [here](https://etherscan.io/address/0x8C17e4968B6903E1601be82Ca989c5B5E2c7b400#code) is the ethereum mainnet swap contract. How do they get the source code and ABI info? Does Remix do that?

You upload the source to etherscan manually.  Remix isn't related to that from what I recall.  When you upload, you specify the same solc build config (version and opt iterations) used when building the contract's bin.

dex/networks/polygon/params.go
```diff
@@ -13,4 +21,131 @@ var (
...
+			// dex.Mainnet: common.HexToAddress(""),
+			// dex.Testnet: common.HexToAddress(""),
+			dex.Simnet: common.HexToAddress(""), // Filled in by MaybeReadSimnetAddrs
+		},
+	}
+
+	testTokenID, _ = dex.BipSymbolID("dextt.polygon")
+	usdcTokenID, _ = dex.BipSymbolID("usdc.polygon")
+
+	Tokens = map[uint32]*dexeth.Token{
```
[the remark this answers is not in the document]
chappjc: Oh, USDC on Polygon testnet is [here] (from https://mumbai.polygonscan.com/token/0x0fa8781a83e46826621b3bc094ea2a0212e71b23): https://developers.circle.com/developer/docs/usdc-on-testnet#bridged-usdc-on-polygon-testnet
Is the issue that that does not have availability on say UniSwap?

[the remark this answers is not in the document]
chappjc: I also realized after posting my last comment is that this is described as _Bridged_ USDC, and I don't know what that implies.  

> Mumbai Testnet is the Polygon testing network. Polygon USDC is a bridged version of USDC that has been bridged from Ethereum to the Polygon PoS network via the Polygon PoS Bridge.

Does that mean you automatically receive Polygon USDC if you have Goerli USDC?  Or does that mean you somehow have to *use* this bridge?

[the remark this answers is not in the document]
chappjc: Ohhh  https://wallet.polygon.technology/polygon/bridge/deposit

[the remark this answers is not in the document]
chappjc: Fun.  It's got me waiting now.  Let's see if it works:

image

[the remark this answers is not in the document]
chappjc: The bridge worked!  Mumbai USDC should be fine.

dex/testing/polygon/harness.sh
```diff
@@ -323,11 +346,78 @@ cat > "${NODES_ROOT}/test_block10_hash.txt" <<EOF
 ${TEST_BLOCK10_HASH}
```
martonp, line 346: This already happened in the previous PR, but for me when the 10th block hash is being retrieved, the 10th block has not yet been mined. Works fine on ETH though.

client/asset/eth/multirpc.go
```diff
@@ -1763,15 +1713,15 @@ func domain(addr string) (string, error) {
...
 	for _, p := range providers {
 		// Need to run API tests on this endpoint.
-		for _, t := range newCompatibilityTests(p.ec, p.chainID, p.net, log) {
+		for _, t := range newCompatibilityTests(p.ec, compat, log) {
 			ctx, cancel := context.WithTimeout(ctx, defaultRequestTimeout)
 			err := t.f(ctx, p)
 			cancel()
 			if err != nil {
-				return fmt.Errorf("RPC Provider @ %q has a non-compliant API: %v", p.host, err)
+				return fmt.Errorf("%s: RPC Provider @ %q has a non-compliant API: %v", t.name, p.host, err)
```
martonp, line 1724: I see this error log in the node after the compatibility test.. I guess the connection is not being closed properly after the tests. Probably nbd though.

Screenshot 2023-07-11 at 5 36 42 AM

martonp (commented):
Regarding deployment, why don't we just use hardhat?
I've used it in https://github.com/decred/dcrdex/pull/2223 to do testing. It's nice because it allows you to directly test the smart contract without worrying about anything else. You can also deploy with it.
https://hardhat.org/

client/asset/eth/cmd/deploy/deploy
ukane-philemon, line 1: Do we need this binary file?

dex/networks/polygon/params.go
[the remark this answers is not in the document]
buck54321: > I also realized after posting my last comment is that this is described as _Bridged_ USDC,

In my ongoing quest to figure out how Polygon is anchored to Ethereum, this appears to be the most pertinent factor. Polygon bridging is tied to POS somehow, so somewhat decentralized. But I believe I read that there's still a 5-of-8 (or something) multi-sig on the Ethereum contracts.

MATIC itself has no representation on Ethereum. That's purely Polygon.

dex/testing/polygon/harness.sh
buck54321: Mmkay. The mine-alpha script for polygon doesn't seem to always work right, and mining on polygon is touchy anyway. It's difficult to produce a small number of blocks and sometimes no blocks are produced, but the mine-alpha script still exits. Will look for a solution.

dex/networks/polygon/params.go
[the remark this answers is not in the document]
chappjc: Yes.  On the flip side Arbitrum and Optimism use ETH as their value.

client/asset/eth/multirpc.go
buck54321: What were you running when you saw this message? dexc or a test?

martonp: dexc

client/asset/eth/eth.go
```diff
@@ -4816,13 +4817,21 @@ func (getGas) EstimateFunding(ctx context.Context, net dex.Network, assetID, con
 	return nil
 }
 
-// ReturnETH returns the estimation wallet's Ethereum balance to a specified
+// Return returns the estimation wallet's Ethereum balance to a specified
```
chappjc, line 4820: Still says "Ethereum balance" but it seems you're now returning whatever, ETH or MATIC, or tokens now.

dex/networks/polygon/params.go
```diff
@@ -36,9 +36,9 @@ var (
 
 	ContractAddresses = map[uint32]map[dex.Network]common.Address{
 		0: {
-			// dex.Mainnet: common.HexToAddress(""),
-			// dex.Testnet: common.HexToAddress(""),
-			dex.Simnet: common.HexToAddress(""), // Filled in by MaybeReadSimnetAddrs
+			dex.Mainnet: common.HexToAddress("0xd45e648D97Beb2ee0045E5e91d1C2C751Cd0Bc00"), // txid: 0xbb7d09fb3832b35fbbed641453a90f217a2736cf1419848887dfee2dbb14187e
```
chappjc, line 39: Confirmed.  https://polygonscan.com/address/0xd45e648D97Beb2ee0045E5e91d1C2C751Cd0Bc00
Will you please attempt to "Verify and Publish" the Solidity source with Polyscan?  If we can't get it verified, it might be a discrepancy with the optimization settings we used to generate the bytecode in the repo, whereas we'd want 200 iterations of optimization.

buck54321: Done

chappjc:
Any suggested Polygon RPC providers?  Every single one at https://github.com/decred/dcrdex/wiki/Ethereum#rpc-provider-list-partial have Polygon endpoints, but I wondered if you has success with any.

chappjc:
https://rpc.ankr.com/polygon_mumbai is probably fine

buck54321:
Alchemy is definitely working. I think I was having problems with Infura and Rivet. I need to do some testing with those.

client/asset/polygon/polygon.go
```diff
@@ -5,24 +5,43 @@ package polygon
...
+		Type:        walletTypeToken,
+		Tab:         "Polygon token",
+		Description: desc,
+	}, nets...)
+}
+
 func init() {
 	asset.Register(BipID, &Driver{})
+	registerToken(simnetTokenID, "A token wallet for the DEX test token. Used for testing DEX software.", dex.Simnet)
+	registerToken(usdcTokenID, "The USDC Ethereum ERC20 token.", dex.Mainnet)
```
chappjc, line 32:
```suggestion
	registerToken(usdcTokenID, "The USDC Ethereum ERC20 token.", dex.Mainnet, dex.Testnet)
```

server/asset/eth/eth.go
```diff
@@ -242,34 +239,40 @@ var _ asset.AccountBalancer = (*ETHBackend)(nil)
...
-	}
 	return &ETHBackend{&AssetBackend{
 		baseBackend: &baseBackend{
-			net:        net,
-			baseLogger: logger,
-			tokens:     make(map[uint32]*TokenBackend),
+			net:             net,
+			baseLogger:      logger,
+			tokens:          make(map[uint32]*TokenBackend),
+			baseChainID:     bipID,
```
chappjc, line 252: Many of the error messages in this file and rpcclient.go confused me with references to "eth" or "ethereum".  Maybe like this: https://github.com/chappjc/dcrdex/commit/69a03e11a270456ef9bc3f26b2ea6a62ca41fa37

server/asset/polygon/polygon.go
```diff
@@ -0,0 +1,78 @@
...
+func init() {
+	asset.Register(BipID, &Driver{eth.Driver{
+		DriverBase: eth.DriverBase{
+			Ver: version,
+			UI:  dexpolygon.UnitInfo,
+		},
+	}})
+
+	registerToken(testTokenID, 0)
+	// registerToken(usdcID, 0)
```
chappjc, line 45:
```suggestion
	registerToken(usdcID, 0)
```

server/asset/polygon/polygon.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+const (
+	BipID              = 966
+	ethContractVersion = 0
+	version            = 0
+)
+
+var (
+	testTokenID, _ = dex.BipSymbolID("dextt.polygon")
```
chappjc, line 62:
```suggestion
	testTokenID, _ = dex.BipSymbolID("dextt.polygon")
	usdcID, _ = dex.BipSymbolID("usdc.polygon")
```

These suggestions are also in https://github.com/chappjc/dcrdex/commit/69a03e11a270456ef9bc3f26b2ea6a62ca41fa3

chappjc (changes requested):
Basic wallet functionality for MATIC and USDC worked well after enabling usdc.polygon (see comments and diffs).

Tested Mumbai (**testnet**) wallet with blast and alchemy endpoints w/ api keys, but here are some public endpoints that could work:

https://polygon-mumbai-bor.publicnode.com
https://polygon-mumbai.blockpi.network/v1/rpc/public
https://rpc-mumbai.polygon.technology
https://endpoints.omniatech.io/v1/matic/mumbai/public
wss://polygon-testnet.public.blastapi.io
https://rpc.ankr.com/polygon_mumbai


image

image
(fine, although above it doesn't mention polygon or overlay the logo; an overlay might be too small anyway)

image

image

I configured MATIC on dex-test.ssgen.io (testnet), but did not test any trades with MATIC, only setup the endpoints and confirmed it would start.  Next, I will test a MATIC trade, and then enable USDC.POLYGON on server.

I used these on the testnet server (only the first needs an apikey):

```
wss://polygon-mumbai.blockpi.network/v1/ws/<apikey>
wss://polygon-testnet.public.blastapi.io
https://polygon-mumbai.blockpi.network/v1/rpc/public
https://polygon-mumbai-bor.publicnode.com
https://rpc.ankr.com/polygon_mumbai
```

chappjc:
I suppose the following should say "MATIC" instead of "POLYGON":

image

Settled fine!

image

chappjc:
Another MATIC vs. POLYGON that I noticed:

image
(right)

image
(wrong)

Probably also this:
image

JoeGruffins (approved):
Working well for me.

One thing I noticed on order history, symbols not there, does it look like this for others?

image

chappjc:
> One thing I noticed on order history, symbols not there, does it look like this for others?

Same here.  All the symbols are missing.

In the browser console I see 404s to the generic coin letter images, also upper case, which breaks their loading although we don't want the generic ones anyway:

image

client/webserver/site/src/js/orders.ts
```diff
@@ -187,19 +187,19 @@ export default class OrdersPage extends BasePage {
...
 
-      set('fromQty', fromQty)
-      Doc.tmplElement(tr, 'fromLogo').src = Doc.logoPath(from)
-      set('fromSymbol', from)
-      set('toQty', toQty)
+      tmpl.fromQty.textContent = fromQty
+      tmpl.fromLogo.src = Doc.logoPath(from)
+      tmpl.fromSymbol.textContent = from
+      tmpl.toQty.textContent = toQty
       Doc.tmplElement(tr, 'toLogo').src = Doc.logoPath(to)
```
chappjc, line 194: We have polygon.png, not MATIC.png, so this ends up trying to load M.png.

dex/bip-id.go
```diff
@@ -479,7 +479,7 @@ var bipIDs = map[uint32]string{
 	890:   "xsel",
 	900:   "lmo",
 	916:   "meta",
-	966:   "matic",
+	966:   "polygon", // changed from "matic"
```
chappjc, line 482: I wonder if we should keep it here as "matic", since our `bipIDs` maps ID to symbol, not coin name.

This npm package addressed it by mapping to both symbol and coin name.  https://github.com/bitcoinjs/bip44-constants#readme
Although I suppose it doesn't really matter if we just use the correct fields from `UnitInfo`...

buck54321: The problems we're seeing are all places where we were using symbol and we should have been using conventional unit. I like this change particularly because it brings those mistakes to the surface.

dex/testing/eth/harness.sh
```diff
@@ -416,6 +404,8 @@ ${ERC20_SWAP_CONTRACT_ADDR}
 EOF
 
 cd "${NODES_ROOT}/harness-ctl"
+
+# DRAFT NOTE: Why are we doing this token send to nowhere?
 TOKEN_SEND_AMT=5000
 echo "Sending 5000 dextt.eth to testing."
 "./sendTokens" "${SIMNET_TOKEN_ADDRESS}" "${TOKEN_SEND_AMT}"
```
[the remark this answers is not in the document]
buck54321: I can't find a reference to the `SIMNET_TOKEN_ADDRESS` anywhere.

dex/testing/polygon/harness.sh
```diff
@@ -323,11 +345,78 @@ cat > "${NODES_ROOT}/test_block10_hash.txt" <<EOF
...
+echo "ERC20 SWAP contract address is ${ERC20_SWAP_CONTRACT_ADDR}. Saving to ${NODES_ROOT}/erc20_swap_contract_address.txt"
+cat > "${NODES_ROOT}/erc20_swap_contract_address.txt" <<EOF
+${ERC20_SWAP_CONTRACT_ADDR}
+EOF
+
+ERC20_SWAP_CONTRACT_ADDR=$("${HARNESS_DIR}/alpha" "--preload ${HARNESS_DIR}/contractAddress.js --exec contractAddress(\"${ERC20_SWAP_CONTRACT_HASH}\")" | sed 's/"//g')
+echo "ERC20 SWAP contract address is ${ERC20_SWAP_CONTRACT_ADDR}. Saving to ${NODES_ROOT}/erc20_swap_contract_address.txt"
+cat > "${NODES_ROOT}/erc20_swap_contract_address.txt" <<EOF
+${ERC20_SWAP_CONTRACT_ADDR}
+EOF
```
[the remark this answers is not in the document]
buck54321: Trying it out. Reduced min `NUM` for `mine-alpha` script to `1` too, So hopefully that helps.

client/asset/eth/cmd/getgas/README.md
```diff
@@ -8,7 +8,7 @@ If this is a new asset, you must populate either `dexeth.VersionedGases` or
 `dexeth.Tokens` with generous estimates before running `getgas`.
 
 ### Use
-- Create a credentials file. `getgas` will look in `~/ethtest/getgas-credentials.json`. You can override that location with the `--creds` CLI argument. The credentials file should have the JSON format in the example below. The seed can be anything.
+- Create a credentials file. `getgas` will look in `~/ethtest/getgas-credentials.json` for Ethereum, and `~/ethtest/getgas-credentials_polygon.json` for Polygon. You can override that location with the `--creds` CLI argument. The credentials file should have the JSON format in the example below. The seed can be anything.
```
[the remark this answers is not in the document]
buck54321: OK if I do this now and screw up all of your credentials files?

norwnd:
Perhaps for **token** deposit/send views it would be nice to display which **Blockchain** was chosen (when user clicked deposit/send button), so that user doesn't accidentally send from Ethereum blockchain to Polygon (or vice versa):

image
image

dex/testing/eth/harness.sh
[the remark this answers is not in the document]
JoeGruffins: Its the address made by the seed in the `./simnet-setup.sh` script so we have funds for trading immediately.

image

buck54321:
### NOTICES

#### 1) For anyone who has used **getgas**, **deploy**, or anyone that has run **nodeclient_harness_test.go** on testnet

I have consolidated all of these credientials-file-based wallets into a single file at **~/dextest/credentials.json**.  See the **getgas** README for the format. The **~/ethtest** directory is no longer used for anything, nor are the **~/ethtest/testnet-credentials.json**, **~/ethtest/getgas-credentials**, **~/ethtest/getgas-credentials_polygon.json**, **~/ethtest/providers.json**, or **~/ethtest/providers_polygon.json** files.

Sorry if this causes some work for you, but the old scheme was getting out of hand.

#### 2) For anyone who has sent funds to a testnet or mainnet Polygon wallet

As per [@norwnd's suggestion](https://github.com/decred/dcrdex/pull/2431#issuecomment-1646791817), I have made changes such that all account-based assets will use the same seed, so will derive the same key and address.

client/asset/polygon/polygon.go
```diff
@@ -5,24 +5,44 @@ package polygon
...
 }
 
 const (
 	// BipID is the BIP-0044 asset ID for Polygon.
-	BipID         = 966
-	walletTypeRPC = "rpc"
+	BipID           = 966
+	walletTypeRPC   = "rpc"
+	walletTypeToken = "token"
+	minerGasCeil    = 8_000_000 // config.Defaults.Miner.GasCeil
```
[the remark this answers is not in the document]
buck54321: Ah. Yes. That was confusing.

For reference, the value ultimately comes from `DefaultConfig()` in **bor/internal/cli/server/config.go**.

We don't need the `MinerBlockGasCeil` field after all.

buck54321:
> Also, I see "polygon" written in some places and "matic" in others

I'm hunting them down. Please point them out where you see them.

buck54321:
> It would be better if this was sorted so that the base chain comes first, and then all of its tokens listed below. I think ideally only the token would be listed on the sidebar and then within the token section you would see each of the networks it is on and your balance on each network.

I agree.. Mind if I save it for a separate, dedicated effort.

client/core/core.go
```diff
@@ -2671,10 +2671,14 @@ func (c *Core) assetSeedAndPass(assetID uint32, crypter encrypt.Crypter) (seed,
 // on external wallet software and their key derivation paths, this seed may be
 // usable for accessing funds outside of DEX applications, e.g. btcwallet.
 func AssetSeedAndPass(assetID uint32, appSeed []byte) ([]byte, []byte) {
+	seedAssetID := assetID
+	if ai, _ := asset.Info(assetID); ai != nil && ai.IsAccountBased {
```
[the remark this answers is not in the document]
buck54321: Changes made

- Make `*ETHWallet` the `WalletRestorer`, so now the export button doesn't show for tokens.
- Reject token asset IDs in **core** with an error pointing to the parent
- If an asset ID for a token somehow makes it into `AssetSeedAndPass`, use the `const accountBasedSeedAssetID = 60 // ETH` to derive the seed

martonp:
I still see polygon instead of matic in the following places.

Screenshot 2023-08-22 at 8 00 48 AM
Screenshot 2023-08-22 at 8 00 24 AM
Screenshot 2023-08-22 at 8 00 17 AM
Screenshot 2023-08-22 at 8 00 03 AM
Screenshot 2023-08-22 at 7 59 36 AM

Also, token icons are messed up on the orders page:

Screenshot 2023-08-22 at 8 09 11 AM

martonp:
Overall LGTM.. I'll test a bit more still on testnet.

martonp:
I'm able to do both matic and polygon usdc swaps on testnet. On the orders page, the block explorer links for polygon token transactions do not work.

buck54321:
Updated the free server list and upped the `defaultGasFeeLimit`. Tested on mainnet.

PR #2432, opened 2023-07-10
[0.6] backport two fixes for release branch

chappjc:
This backports two important bug fixes for the 0.6 release branch:
- https://github.com/decred/dcrdex/pull/2405 To prevent users being penalized for not reporting their redeems.
- https://github.com/decred/dcrdex/pull/2411 So positive tier users always have a positive order limit, big enough for at least a single lot.

### DIFF
--- a/client/core/trade.go
+++ b/client/core/trade.go
@@ -92,8 +92,9 @@ type matchTracker struct {
 
 	// confirmRedemptionNumTries is just used for logging.
 	confirmRedemptionNumTries int
-	// redemptionConfs and redemptionConfsReq are set while the redemption
-	// confirmation process is running.
+	// redemptionConfs and redemptionConfsReq are updated while the redemption
+	// confirmation process is running. Their values are not updated after the
+	// match reaches MatchConfirmed status.
 	redemptionConfs    uint64
 	redemptionConfsReq uint64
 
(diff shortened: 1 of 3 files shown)


client/core/trade.go
```diff
@@ -2851,14 +2861,25 @@ func (c *Core) sendRedeemAsync(t *trackedTrade, match *matchTracker, coinID, sec
 // mutex lock held for writes.
 func (t *trackedTrade) confirmRedemption(match *matchTracker) {
 	confirmer, isConfirmer := t.wallets.toWallet.Wallet.(asset.RedemptionConfirmer)
-	if !isConfirmer {
+	if !isConfirmer || (match.redemptionConfs > 0 && match.redemptionConfs >= match.redemptionConfsReq) { // already there, stop checking
```
chappjc, line 2864: This is the only difference from the master version of this fix.  On master, all wallets are confirmers, but not in 0.6, so we merge these cases where handling is the same (mark it confirmed if we have submitted the redeem or it's self-governed).

client/core/trade.go
```diff
@@ -2834,12 +2835,21 @@ func (c *Core) sendRedeemAsync(t *trackedTrade, match *matchTracker, coinID, sec
 		if match.Side == order.Maker && match.Status < order.MatchComplete {
 			// As maker, this is the end. However, this diverges from server,
 			// which still needs taker's redeem.
-			match.Status = order.MatchComplete
+			if conf := match.redemptionConfs; conf > 0 && conf >= match.redemptionConfsReq {
+				match.Status = order.MatchConfirmed // redeem tx already confirmed before redeem request accepted by server
```
chappjc, line 2839: This condition isn't updated with a `!isConfirmer || ` so there's a delay until the next tick gets to `confirmRedemption` with the newy set `RedeemSig`.  Should we add the confirmer condition here too or is the delay fine?

chappjc: Made it a fast path in https://github.com/decred/dcrdex/pull/2432/commits/f35a0c77866849bdbcf065da22a949e876e55f38.
Still tests fine.

JoeGruffins (approved):

buck54321 (approved):

PR #2433, opened 2023-07-11
client/dcr: MultiTrade

martonp:
Implements FundMultiOrder in the DCR wallet. Also removes the MultiOrderFunder interface because all of the wallets now implement MultiTrade.

### DIFF
--- a/client/asset/dcr/coin_selection.go
+++ b/client/asset/dcr/coin_selection.go
@@ -4,6 +4,7 @@
 package dcr
 
 import (
+	"math"
 	"math/rand"
 	"sort"
 	"time"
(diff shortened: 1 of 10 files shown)


client/asset/dcr/dcr_test.go
```diff
@@ -1152,6 +1159,1209 @@ func checkSwapEstimate(t *testing.T, est *asset.SwapEstimate, lots, swapVal, max
...
+				{newOutput(&txHashes[2], 0, 22e5, wire.TxTreeRegular)},
+			},
+			expectedRedeemScripts: [][]dex.Bytes{
+				{nil, nil},
+				{nil},
+			},
+			expectedLockedCoins: []*wire.OutPoint{
+				{txHashes[0], 0, wire.TxTreeRegular},
+				{txHashes[1], 0, wire.TxTreeRegular},
+				{txHashes[2], 0, wire.TxTreeRegular},
```
chappjc, line 1342: maybe use wire.NewOutPoint or make a helper to rid of this linter error

JoeGruffins (approved):

client/asset/dcr/coin_selection.go
```diff
@@ -165,37 +185,78 @@ searchLoop:
...
+		return false
+	}
+
+	if len(utxos) == 1 {
+		e, _ := enough(0, 0, utxos[0])
+		return e
+	}
+
+	valueWithoutLast := sumUTXOs(utxos[:len(utxos)-1])
+	sizeWithoutLast := sumUTXOSize(utxos[:len(utxos)-1])
```
JoeGruffins, line 206: Why does this exclude the last utxo?

chappjc (approved):

client/asset/dcr/coin_selection.go
martonp: Because the `enough` functions in DCR check if adding a new UTXO to the existing set will make it enough, but here I'm just trying to check if the entire set is enough or not. So I just take the entire set without the last element, and use the last element as the new UTXO in the enough function.

JoeGruffins: Oh, I was over thinking this. I get it now.

PR #2434, opened 2023-07-11
doge: enable background mining in harness

dev-warrior777:
## Harnesses:

- switch on mining for DGB
- add mining for DOGE

set/export NOMINER="1" to disable

DGB documented to have issues after mining ~100 blocks. Selectively disable as needed.

### DIFF
--- a/dex/testing/doge/harness.sh
+++ b/dex/testing/doge/harness.sh
@@ -276,6 +276,7 @@ tmux send-keys -t $SESSION:0 C-c
 tmux send-keys -t $SESSION:1 C-c
 tmux send-keys -t $SESSION:2 C-c
 tmux send-keys -t $SESSION:3 C-c
+tmux send-keys -t $SESSION:5 C-c
 tmux wait-for alpha${SYMBOL}
 tmux wait-for beta${SYMBOL}
 # seppuku
--- a/dex/testing/doge/harness.sh
+++ b/dex/testing/doge/harness.sh
@@ -15,6 +15,9 @@ GAMMA_RPC_PORT="23771"
 ALPHA_MINING_ADDR="mjzB71SzfAi8BwNvobPZ983d8vCdxuQD2i"
 BETA_MINING_ADDR="mwL7ypWCMEhBhGcsqWNwkvAasuXP3duXQk"
 
+# Uncomment to disable background mining
+#NOMINER="1"
+
 set -ex
 NODES_ROOT=~/dextest/${SYMBOL}
 rm -rf "${NODES_ROOT}"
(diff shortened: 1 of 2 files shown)


chappjc:
I think when we validate 8.22 (once they release final, not rc) we can enable DGB miner.  For now it's broken.

DOGE is good.

dev-warrior777:
What I can do is re-submit a pr with just Doge .. if that works. I have a couple more lines in DGB `harness.sh` that would need adding for 8.x anyway.

I will do that as a new Doge PR if that is ok.

chappjc:
Don't make a new PR pls.  No need for the churn.

dex/testing/doge/harness.sh
```diff
@@ -315,6 +318,18 @@ done
 
 tmux send-keys -t $SESSION:4 "./mine-alpha 2${DONE}" C-m\; ${WAIT}
 
-# Reenable history and attach to the control session.
+################################################################################
+# Setup watch background miner -- if required
+################################################################################
+if [ -z "$NOMINER" ] ; then
+  tmux new-window -t $SESSION:5 -n "miner" $SHELL
```
chappjc, line 325: Please add `$SESSION:5` to the `quit` script up above.

dex/testing/doge/harness.sh
```diff
@@ -273,6 +276,7 @@ tmux send-keys -t $SESSION:0 C-c
 tmux send-keys -t $SESSION:1 C-c
 tmux send-keys -t $SESSION:2 C-c
 tmux send-keys -t $SESSION:3 C-c
+tmux send-keys -t $SESSION:5 C-c
```
JoeGruffins, line 279: Also 4? If I understand correctly, this is just sending ctrl-c to the harnesses to quit whatever they are doing?

dev-warrior777: I tried both ways .. does not seem to make much difference.

`Window #4` (not 2) is Harness Control for doge. It's the active window as we call `quit` and you could see `^C` on the command line as the only difference. 

Then we kill the session.

JoeGruffins: Oh, 4 is the harness, then it cannot be doing something other than `./quit` I see

dev-warrior777: Your question did make me think more .. what if you started a long running process in the harness window/tty programmatically.
In this case you would have to take care of terminating it yourself if you wanted to do that early.
You would just send it a signal!

chappjc:
draft still?

dev-warrior777:
> draft still?

I think it is good to go

buck54321 (approved):

PR #2437, opened 2023-07-17
client/core: check if wallet is synchronized for spend actions

### LINKED ISSUE #2436: disallow sending funds until wallet is 100% synced
Sending funds from an unsynchronized wallet leads to making txns spending already spend outputs, and transactions that just won't ever get mined.  This is confusing to users and hard to diagnose in support.

ukane-philemon:
Closes #2436

### DIFF
--- a/client/core/core.go
+++ b/client/core/core.go
@@ -5324,8 +5324,8 @@ func (c *Core) Send(pw []byte, assetID uint32, value uint64, address string, sub
 		return nil, err
 	}
 
-	if !wallet.synchronized() {
-		return nil, fmt.Errorf("%s is still syncing", unbip(assetID))
+	if err = wallet.checkPeersAndSyncStatus(); err != nil {
+		return nil, err
 	}
 
 	var coin asset.Coin
@@ -5383,8 +5383,9 @@ func (c *Core) ApproveToken(appPW []byte, assetID uint32, dexAddr string, onConf
 		return "", err
 	}
 
-	if !wallet.synchronized() {
-		return "", fmt.Errorf("%s is still syncing", unbip(assetID))
+	err = wallet.checkPeersAndSyncStatus()
+	if err != nil {
+		return "", err
 	}
 
 	dex, connected, err := c.dex(dexAddr)
@@ -5432,8 +5433,9 @@ func (c *Core) UnapproveToken(appPW []byte, assetID uint32, version uint32) (str
 		return "", err
 	}
 
-	if !wallet.synchronized() {
-		return "", fmt.Errorf("%s is still syncing", unbip(assetID))
+	err = wallet.checkPeersAndSyncStatus()
+	if err != nil {
+		return "", err
 	}
 
 	onConfirm := func() {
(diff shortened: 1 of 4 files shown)


client/core/core.go
```diff
@@ -10419,9 +10431,18 @@ func (c *Core) SendShielded(appPW []byte, assetID uint32, toAddr string, amt uin
...
+	if !found {
+		return nil, fmt.Errorf("no %s wallet", unbip(assetID))
+	}
+
+	sw, is := w.Wallet.(asset.ShieldedWallet)
+	if !is {
+		return nil, fmt.Errorf("%s wallet is not a shielded wallet", unbip(assetID))
+	}
+
+	if !w.synchronized() {
```
chappjc, line 10444: Would make sense for `ShieldFunds` and `UnshieldFunds`, possibly even `NewShieldedAddress` to check also, so maybe the `shieldedWallet` can return a bool too.

ukane-philemon: `ShieldFunds` and `UnshieldFunds` are like balance checks, but I don't think they should be affected?

IMO, we could make it more obvious that wallets are still syncing and a tooltip about what they should not do until it's done. What do you think about this?

chappjc: > `ShieldFunds` and `UnshieldFunds` are like balance checks, but I don't think they should be affected?

No, these make transactions, sending value in transactions between accounts.  They should be blocked if the wallet is not synchronized.

ukane-philemon: > No, these make transactions, sending value in transactions between accounts

Oh, let me check.

ukane-philemon: Updated, but not checking in `NewShieldedAddress` yet unless you really think we should.

client/core/core.go
```diff
@@ -5324,6 +5324,10 @@ func (c *Core) Send(pw []byte, assetID uint32, value uint64, address string, sub
 		return nil, err
 	}
 
+	if !wallet.synchronized() {
+		return nil, fmt.Errorf("%s is still syncing", unbip(assetID))
+	}
```
chappjc, line 5329: Only other thought is that we could add peers>0 check to this as well.  See other spots where we directly access the xcWallet fields under mutex lock, or use the state method, which is slightly more work.

ukane-philemon: Updated. `postRequiredBonds` does not check for peer count atm but checks for `wallet.synchronized()`, do you want us to also check for peer count there?

chappjc: Let's do it.
I don't like transactions going nowhere

ukane-philemon: Updated my last commit.

chappjc (approved):
LGTM, and you appear to have fixed a data race in `confirmRedemption`.

ukane-philemon:
> you appear to have fixed a data race in confirmRedemption.

Good thing. I didn't notice.

client/core/trade.go
```diff
@@ -2903,12 +2903,8 @@ func (t *trackedTrade) confirmRedemption(match *matchTracker) (bool, error) {
 	// In some cases the wallet will need to send a new redeem transaction.
 	toWallet := t.wallets.toWallet
 
-	if toWallet.peerCount < 1 {
-		return false, fmt.Errorf("%s wallet has no peers", unbip(toWallet.AssetID))
-	}
```
ukane-philemon, line 2908: Oh, this. 🌚

PR #2440, opened 2023-07-18
dexc-desktop: fix darwin installer script version

ukane-philemon:


client/app/VERSION
ukane-philemon, line 1: No other consumer, except for `dexc-desktop/pkg/pkg-darwin.sh`

chappjc (approved):
Thanks.  In a future revision where we need to allow the caller of pkg-darwin.sh (e.g. a decred release builder) to set the version like `0.7.0+release` we can update the script so it is uses shell parameter expansion to override the set value. http://www.gnu.org/software/bash/manual/bashref.html#Shell-Parameter-Expansion

ukane-philemon:
> a decred release builder

I don't know when we'd want to have this but maybe this might be helpful if we plan to use gh actions:  https://github.com/arduino/arduino-ide/blob/f6a43254f5c416a2e4fa888875358336b42dd4d5/.github/workflows/build.yml#L79

Issue #2442, opened 2023-07-19
client/dcr: restarting dexc mid-trade results in stuck trade


norwnd:
I'm observing this issue pretty consistently when restarting **dexc** mid-trade (right after Taker swap): https://github.com/decred/dcrdex/pull/2042#discussion_r1168188798

I've looked a bit into it and think it more likely has to do with the following being out of sync than with reorgs, something like:
- when **dexc** starts, trades are pulled from DB and redemption search starts (through `FindRedemption` method `findRedemptionQueue` gets filled)
- **dcr spv** wallet starts in parallel to this ^ and inspects blocks that keep coming in `monitorBlocks`, it seems to be doing some time-based delays (at least I don't see obvious way it's synched with 1st process from above)

so when **dexc** restarts (perhaps with several minutes delay), it misses the block we are interested in (or a whole range of blocks maybe) would be my guess ... but I give up this investigation, leaving this in case somebody wants to pick it up.

chappjc:
Is it revoked?  Shouldn't be searching for a redemption unless the match is revoked.
Can you provide a specific repro?

norwnd:
Now that match is revoked on both sides, it looks like this (more restarts don't change anything):<br>

<img width="400" alt="image" src="https://github.com/decred/dcrdex/assets/112318969/e6e60f28-0f79-4627-bfd9-0a1c845ad203">
<img width="400" alt="image" src="https://github.com/decred/dcrdex/assets/112318969/a8ca2e07-cc11-4606-a422-f1efc86511b3">

my understanding: Maker should be trying (and succeeding) to redeem, but it doesn't and I see this in logs:
```
2023-07-20 08:39:42.927 [INF] CORE: Match 898cdf9d705b9febb59dfbc82db8b56619417d4a200ed73ef42e6459e3355875 not yet redeemable: current confs = 0, required confs = 1
```

chappjc:
It's not consistent, but I did reproduce it twice.  PR just for debugging, with a possible workaround: https://github.com/decred/dcrdex/pull/2444
However, there seems to be a bug in dcrwallet/wallet/udb that keeps the transaction unconfirmed.

PR #2444, opened 2023-07-20
client/asset/dcr: workaround TxDetails unconfirmed bug

chappjc:
Pertains to https://github.com/decred/dcrdex/issues/2442.

There is an issue given:
- taker's swap tx is DCR
- you shutdown dexc after the (taker) DCR swap transaction is broadcast, but before it is mined
- start dexc, login
- wait for DCR block(s), transaction always appears unconfirmed, swap never proceeds (except refund)

I believe you must be both maker and taker (self trade) for it to halt the swap because if you are not also the taker who authored the DCR swap transaction, then for maker, the `SwapConfirmations` method will fallback to cfilters scan when `lookupTxOutput` fails to find the transaction.  Or so I would expect since the counterparty contract transaction would not be a wallet transaction.

The above is not always reproducible, but I managed to get it to happen twice with the native SPV wallet.

I tracked the culprit down to `dcrwallet/wallet/udb.(*Store).TxDetails`.  It will find and return an **unmined** transaction that it finds with `existsRawUnmined` and `unminedTxDetails` (but it's really mined, several confs even).

https://github.com/decred/dcrwallet/blob/master/wallet/udb/txquery.go#L180-L194

I do not know if it would find a **mined** transaction *also* if it could get down to `latestTxRecord` and `minedTxDetails`, but it seems the transaction is left in `bucketUnmined`.  i.e. `deleteRawUnmined` is not called as expected.  There are only two call sites for `deleteRawUnmined`:
- `wallet/udb.(*Store).moveMinedTx`
- `wallet/udb.(*Store).RemoveUnconfirmed`

I don't know which isn't getting called.

Perhaps one detail worth noting is that I generally use a split tx in the order, meaning the resulting DCR swap contract transaction will have no change, so it's just the input/debit that belongs to the authoring wallet, and the one swap output technically belongs to neither wallet.

### DIFF
--- a/client/asset/dcr/dcr.go
+++ b/client/asset/dcr/dcr.go
@@ -30,6 +30,7 @@ import (
 	dexdcr "decred.org/dcrdex/dex/networks/dcr"
 	walletjson "decred.org/dcrwallet/v3/rpc/jsonrpc/types"
 	_ "decred.org/dcrwallet/v3/wallet/drivers/bdb"
+	"github.com/davecgh/go-spew/spew"
 	"github.com/decred/dcrd/blockchain/stake/v5"
 	"github.com/decred/dcrd/blockchain/standalone/v2"
 	"github.com/decred/dcrd/chaincfg/chainhash"
@@ -3513,6 +3514,7 @@ func (dcr *ExchangeWallet) lookupTxOutput(ctx context.Context, txHash *chainhash
 	// Check for an unspent output.
 	output, err := dcr.wallet.UnspentOutput(ctx, txHash, vout, wire.TxTreeUnknown)
 	if err == nil {
+		dcr.log.Infof("lookupTxOutput: UnspentOutput found it! %v:%d", txHash, vout)
 		return output.TxOut, output.Confirmations, 0, nil
 	} else if !errors.Is(err, asset.CoinNotFoundError) {
 		return nil, 0, 0, err
@@ -3530,6 +3532,7 @@ func (dcr *ExchangeWallet) lookupTxOutput(ctx context.Context, txHash *chainhash
 	if int(vout) >= len(msgTx.TxOut) {
 		return nil, 0, 0, fmt.Errorf("tx %s has no output at %d", txHash, vout)
 	}
+	dcr.log.Debug(spew.Sdump(tx))
 
 	txOut = msgTx.TxOut[vout]
 	confs = uint32(tx.Confirmations)
@@ -3538,6 +3541,9 @@ func (dcr *ExchangeWallet) lookupTxOutput(ctx context.Context, txHash *chainhash
 	if confs == 0 {
 		// Only counts as spent if spent in a mined transaction,
 		// unconfirmed tx outputs can't be spent in a mined tx.
+		if dcr.wallet.SpvMode() {
+			return txOut, confs, -1, nil // make it do a cfilters scan because GetTransaction lies
+		}
 		return txOut, confs, 0, nil
 	}
 
@@ -4541,10 +4547,14 @@ func (dcr *ExchangeWallet) SwapConfirmations(ctx context.Context, coinID, contra
 	_, confs, spendFlag, err = dcr.lookupTxOutput(ctx, txHash, vout)
 	if err == nil {
 		if spendFlag != -1 {
+			dcr.log.Infof("SwapConfirmations: lookupTxOutput DID find %v (confs = %v, spent = %v)", txHash, confs, spendFlag)
 			return confs, spendFlag > 0, nil
 		} // else go on to block filters scan
+		dcr.log.Infof("SwapConfirmations: lookupTxOutput DID find %v, but going to cfilters scan for spend status", txHash)
 	} else if !errors.Is(err, asset.CoinNotFoundError) {
 		return 0, false, err
+	} else {
+		dcr.log.Infof("SwapConfirmations: lookupTxOutput did NOT find %v, going to cfilters scan for confs", txHash)
 	}
 
 	// Prepare the pkScript to find the contract output using block filters.
--- a/client/asset/dcr/spv.go
+++ b/client/asset/dcr/spv.go
@@ -81,6 +81,10 @@ type dcrWallet interface {
 	UnlockAccount(ctx context.Context, account uint32, passphrase []byte) error
 	LoadPrivateKey(ctx context.Context, addr stdaddr.Address) (key *secp256k1.PrivateKey, zero func(), err error)
 	TxDetails(ctx context.Context, txHash *chainhash.Hash) (*udb.TxDetails, error)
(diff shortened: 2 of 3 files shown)


client/asset/dcr/dcr.go
```diff
@@ -3538,6 +3541,9 @@ func (dcr *ExchangeWallet) lookupTxOutput(ctx context.Context, txHash *chainhash
 	if confs == 0 {
 		// Only counts as spent if spent in a mined transaction,
 		// unconfirmed tx outputs can't be spent in a mined tx.
+		if dcr.wallet.SpvMode() {
+			return txOut, confs, -1, nil // make it do a cfilters scan because GetTransaction lies
+		}
```
chappjc, line 3546: This is a possible workaround, which unfortunately forces the cfilters scan path even when `GetTransaction` did return a result (it's a wallet transaction i.e. checking your own swap confirms or it's a self-trade)

client/asset/dcr/spv.go
```diff
@@ -690,7 +695,34 @@ func (w *spvWallet) GetTransaction(ctx context.Context, txHash *chainhash.Hash)
...
+			goto gotit
+		}
+		_, confs2, blockHashP, err := w.dcrWallet.TransactionSummary(ctx, txHash)
+		if err != nil {
+			w.log.Warnf("TransactionSummary: %v", err)
+		} else if confs2 > 0 {
+			w.log.Warnf("TransactionSummary says yes! block %v, confs %d", blockHashP, confs2)
+			ret.BlockHash = blockHashP.String()
+			ret.Confirmations = int64(confs2)
+			goto gotit
```
chappjc, line 722: All of this is because I wanted to see if any of the other `wallet` methods can give a conf or block inspite of `TxDetails`.  It does not appear to return any different results.  This seems to be because they all stop on `existsRawUnmined`, just like `TxDetails`.

buck54321:
Appears that under normal operation without shutting down, dcrwallet always reaches `deleteRawUnmined` through `moveMinedTx`, ultimately from a headers announcement.

```Bash
decred.org/dcrwallet/v3/wallet/udb.deleteRawUnmined({0x2811878, 0xc000b69ac0}, {0xc000b4d018, 0x20, 0x20})
	/path/to/dcrwallet/wallet/udb/txdb.go:1377 +0x36
decred.org/dcrwallet/v3/wallet/udb.(*Store).moveMinedTx(0x20?, {0x2811878?, 0xc000b69ac0}, {0x7fbf4c7a8c50, 0xc000b69bc0}, 0xc000b4cfd0, {0xc0013b42d0, 0x44, 0x44}, {0xc0000efd30, ...}, ...)
	/path/to/dcrwallet/wallet/udb/txmined.go:1125 +0xc09
decred.org/dcrwallet/v3/wallet/udb.(*Store).InsertMinedTx(0xbe5466cf452821e6?, {0x280acf8, 0xc0003cb648}, 0xc000b4cfd0, 0xc000b39770)
	/path/to/dcrwallet/wallet/udb/txmined.go:1269 +0xbf1
decred.org/dcrwallet/v3/wallet.(*Wallet).processTransactionRec thoughord(0xc0001a4200, {0x2804a98, 0xc0004c17c0}, {0x280acf8?, 0xc0003cb648}, 0xc000b4cfd0, 0xc000b44b60, 0xc000b39770)
	/path/to/dcrwallet/wallet/chainntfns.go:448 +0x17e
decred.org/dcrwallet/v3/wallet.(*Wallet).extendMainChain(0xc0001a4200, {0x2804a98, 0xc0004c17c0}, {0x1b93765, 0x12}, {0x280acf8, 0xc0003cb648}, 0xc000b44b60, 0x0?, {0xc0003cb638, ...})
	/path/to/dcrwallet/wallet/chainntfns.go:76 +0x80e
decred.org/dcrwallet/v3/wallet.(*Wallet).ChainSwitch.func1({0x280acf8, 0xc0003cb648})
	/path/to/dcrwallet/wallet/chainntfns.go:170 +0x890
decred.org/dcrwallet/v3/wallet/walletdb.Update({0x2804a98, 0xc0004c17c0}, {0x2804de0, 0xc00012a480}, 0xc0005f9bd0)
	/path/to/dcrwallet/wallet/walletdb/interface.go:252 +0x195
decred.org/dcrwallet/v3/wallet.(*Wallet).ChainSwitch(0xc0001a4200, {0x2804a98, 0xc0004c17c0}, 0xc00011e3f8, {0xc0003cb590, 0x1, 0x1}, 0xc00139e930)
	/path/to/dcrwallet/wallet/chainntfns.go:114 +0x2da
decred.org/dcrwallet/v3/spv.(*Syncer).handleBlockAnnouncements.func2(0xc00011e360, {0x2804a98, 0xc0004c17c0}, 0xc000b44b60, 0xc000b359e0?, {0xc0003cb558, 0x1, 0x4?}, 0xc0005f9ed8, 0xc0005f9ef0, ...)
	/path/to/dcrwallet/spv/sync.go:1223 +0xa8b
decred.org/dcrwallet/v3/spv.(*Syncer).handleBlockAnnouncements(0xc00011e360, {0x2804a98, 0xc0004c17c0}, 0xc000268960, {0xc0003cb558, 0x1, 0x1}, 0xc001381860?)
	/path/to/dcrwallet/spv/sync.go:1239 +0x2ca
decred.org/dcrwallet/v3/spv.(*Syncer).receiveHeadersAnnouncements.func1()
	/path/to/dcrwallet/spv/sync.go:940 +0x4b
created by decred.org/dcrwallet/v3/spv.(*Syncer).receiveHeadersAnnouncements
	/path/to/dcrwallet/spv/sync.go:939 +0x33
```

Way back at `handleBlockAnnouncements`, the new blocks are scanned for relevant transactions.

```Go
matchingTxs, err = s.scanChain(ctx, rp, bestChain, bmap)
```
which are passed in to `ChainSwitch`.

On startup though, they add headers via `(*Syncer).Run -> connectToCandidates,connectToPersistent -> startupSync -> getHeaders` where `ChainSwitch` is called with a `nil` `relevantTxs` argument, and I don't see a place where it attempts to catch up.

Still poking around.

chappjc:
> On startup though, they add headers via `(*Syncer).Run -> connectToCandidates,connectToPersistent -> startupSync -> getHeaders` where `ChainSwitch` is called with a `nil` `relevantTxs` argument, and I don't see a place where it attempts to catch wallet transactions in those blocks.

Interesting.  Just one thing I want to clarify is that when I've reproduced the bug, the wallet started up at the same best block as when it shut down, and only a bit later (minutes) the tx would be really mined.  This would correspond to the `rescanPoint == nil` case I believe.

chappjc:
As per convo with @jrick, there's a very long standing known bug with wallet startup when there are unconfirmed transactions, whereby they are often never confirmed without a rescan.  That appears to be what we are seeing.

Assuming the underlying wallet bug isn't fixed, I propose the workaround I called out in `lookupTxOutput` where we force a cfilter scan if the number of confirmations is 0 and it's an SPV wallet.

buck54321:
Critical fix pulled out in #2555. Closing this in favor of issue #2556

PR #2448, opened 2023-07-26
site: Fix order marker bug

### LINKED ISSUE #2373: ui: depth chart marker for own orders reluctantly appears
The triangle marker on the depth chart that identifies the location of your own orders does not appear until you hover your mouse over the order in the "Your Orders" list below the order form.

image

Note that hovering does make it get big, and then it gets smaller, as expected, but neither size marker is there until you do that.

ukane-philemon:
Closes #2373


buck54321 (approved):
Looks like the `status` fix resolved the issue, but the `AsyncOrderSubmitted` change seems right.

PR #2451, opened 2023-07-28
server/auth: only send penalize note on tier change

buck54321:


martonp (approved):

JoeGruffins (approved):
This looks to stop cancels from triggering the bug.

PR #2452, opened 2023-07-31
dex/testing: DCR watch mining can be disabled with NOMINER

dev-warrior777:
DCR watch mining is still on by default but now can be disabled by setting NOMINER to a non-empty string value.


dev-warrior777:
It also switches to SESSION:0 Harness control

buck54321 (approved):

PR #2460, opened 2023-08-04
client/bond.go: Avoid dropping tier.

### LINKED ISSUE #2459: bonds: Users with a good server score do not stay bonded.
Client bonds use the tier from the server to decide how many bonds to queue up. If client is targeting tier 1 and has no good behavior it will create a new bond before the one runs out to stay bonded. However if the server tells the client they are tier 2 because of good behavior, the client will assume it does not need to create a new bond when the one bond is about to expire becoming unbonded and having tier set to 0 suddenly, canceling all orders.

JoeGruffins:
If client has good standing with the server, it will not preemptively create a bond before its only bond expires. Always restore weak bonds if they are the last bonds to avoid going to zero. Reserve enough coin to restore live bonds rather than tracking the delta.

closes #2459

### DIFF
--- a/client/core/bond.go
+++ b/client/core/bond.go
@@ -227,6 +227,8 @@ func (c *Core) bondStateOfDEX(dc *dexConnection, bondCfg *dexBondCfg) *dexAcctBo
 			} else {
 				if int64(bond.LockTime) <= bondCfg.replaceThresh {
 					state.weak = append(state.weak, bond) // but not yet expired (still live or pending)
+					c.log.Infof("Soon to expire bond found: %v (%s)",
+						coinIDString(bond.AssetID, bond.CoinID), unbip(bond.AssetID))
 				}
 				liveBonds = append(liveBonds, bond)
 			}
@@ -259,6 +261,11 @@ func (c *Core) bondStateOfDEX(dc *dexConnection, bondCfg *dexBondCfg) *dexAcctBo
 
 	tierDeficit := int64(state.targetTier) - state.tier
 	state.mustPost = tierDeficit + state.weakStrength - state.pendingStrength
+	// If weak bonds expiring would put the number of live and pending bonds
+	// to zero, replace the weak bonds.
+	if state.mustPost <= 0 && state.weakStrength >= state.liveStrength+state.pendingStrength {
+		state.mustPost = state.weakStrength - state.pendingStrength
+	}
 
 	return state
 }
@@ -269,9 +276,9 @@ type bondID struct {
 }
 
 type acctUpdate struct {
-	acct         *dexAccount
-	tierChange   int64
-	reserveDelta int64
+	acct       *dexAccount
+	tierChange int64
+	reserve    uint64
 }
 
 // refundExpiredBonds refunds expired bonds and returns the list of bonds that
@@ -498,7 +505,6 @@ func (c *Core) rotateBonds(ctx context.Context) {
 
 	now := time.Now().Unix()
 
-	netReserveDeltas := make(map[uint32]int64) // for all DEX conns, per asset, compare to lastReserves
 	actuatedTierChanges := make(map[uint32][]acctUpdate)
 
 	for _, dc := range c.dexConnections() {
@@ -541,19 +547,28 @@ func (c *Core) rotateBonds(ctx context.Context) {
 		}
 
 		if acctBondState.tierChange != 0 {
-			reserveDelta := bondOverlap * -acctBondState.tierChange * int64(bondAsset.Amt)
-			netReserveDeltas[acctBondState.bondAssetID] += reserveDelta // negative means free reserves
-			// After applying the net reserveDelta, we would normally *zero* the
+			// Reserve enough funds to recreate current bonds.
+			reserve := bondOverlap * uint64(acctBondState.liveStrength) * bondAsset.Amt
+			// After applying the reserve, we would normally *zero* the
 			// tierChange field, but it could change, so we remember it.
 			actuatedTierChanges[acctBondState.bondAssetID] = append(actuatedTierChanges[acctBondState.bondAssetID],
-				acctUpdate{dc.acct, acctBondState.tierChange, reserveDelta},
+				acctUpdate{dc.acct, acctBondState.tierChange, reserve},
 			)
 		}
 
 		c.postRequiredBonds(dc, bondCfg, acctBondState, bondAsset, wallet, expiredStrength, unlocked)
 	}
 
-	for assetID, reserveDelta := range netReserveDeltas {
+	for assetID, actuatedTierChange := range actuatedTierChanges {
+		var reserveDelta int64
+		// Update the unactuated tier change counter.
+		for _, s := range actuatedTierChange {
+			s.acct.authMtx.Lock()
+			reserveDelta += s.acct.totalReserved - int64(s.reserve)
+			s.acct.totalReserved = int64(s.reserve)
+			s.acct.tierChange -= s.tierChange
+			s.acct.authMtx.Unlock()
+		}
 		if reserveDelta != 0 { // net adjustment across all dexConnections
 			wallet, err := c.connectedWallet(assetID)
 			if err != nil { // we grabbed it above so shouldn't happen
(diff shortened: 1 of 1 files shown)


JoeGruffins:
I'm not super confident about these changes, but they seem to work correctly. We want to reserve the same amount as bonded so we can re bond later I think?

client/core/bond.go
```diff
@@ -259,6 +261,11 @@ func (c *Core) bondStateOfDEX(dc *dexConnection, bondCfg *dexBondCfg) *dexAcctBo
 
 	tierDeficit := int64(state.targetTier) - state.tier
 	state.mustPost = tierDeficit + state.weakStrength - state.pendingStrength
+	// If weak bonds expiring would put the number of live and pending bonds
+	// to zero, replace the weak bonds.
+	if state.mustPost <= 0 && state.weakStrength >= state.liveStrength+state.pendingStrength {
```
martonp, line 266: The liveStrength does not include the bonds in the weakStrength category. Would it work if we just post 1 bond if `state.liveStrength+state.pendingStrength == 0`?

JoeGruffins: Yeah. That may be more correct.

buck54321: The `liveStrength` **does** include weak bonds, so `weakStrength` is always `<= liveStrength`.

buck54321:
I was looking at something like this, @JoeGruffins.

https://github.com/buck54321/dcrdex/compare/95e0e58...buck54321:dcrdex:mustpost-fix

JoeGruffins:
@buck54321 Great! Thanks! Will apply that to this pr if it's alright. How about the reserves changes here, do they look ok?

client/core/bond.go
martonp: Ah my mistake.

client/core/bond.go
```diff
@@ -257,8 +259,41 @@ func (c *Core) bondStateOfDEX(dc *dexConnection, bondCfg *dexBondCfg) *dexAcctBo
...
+		// that case, just make sure that our local bond data is sufficient to
+		// get our target tier. For clients with low targetTier and good trading
+		// history, this will be the primary route through which weak bonds are
+		// renewed.
+		// Note: For users with targetTier > 1, unknown bonus tiers can also
+		// cause us to broadcast renewal bonds that are too small, resulting
+		// in split bonds that cost the user unnecessary tx fees. This solution
+		// is only a stop-gap. A proper solution will require more data from the
+		// API so we can know our bonus tiers.
+		bondTierDeficit := int64(state.targetTier) - state.liveStrength - state.pendingStrength + state.weakStrength
```
martonp, line 292: This basically just cancels out the benefit the user gets from good trading history. Their good trading history will not allow them to have less locked in bonds. If we do
``` golang
if state.liveStrength - state.weakStrength + state.pendingStrength == 0 {
    state.mustPost = 1
}
```
then the user posts the minimum they need to maintain the tier they desire.

buck54321: I don't follow. Can you give an example with numbers?

martonp: targetTier = 2
currentTier = 4
liveStrength = 2
weakStrength = 2
pendingStrength = 0

In this situation, the original logic would post 0, and the updated logic would post 2. However we only need to post one to make sure that we have at least one live bond.

JoeGruffins: If currentTier was 2 there, then would it drop below target for a moment?

JoeGruffins: Yeah. Because if weakStrength would ever bring tier below target tier, the original code will fire. I think one is always enough. It's difficult to reason about though, for some reason.

JoeGruffins: @buck54321 is this what this is referring to?
```
		// Note: For users with targetTier > 1, unknown bonus tiers can also
		// cause us to broadcast renewal bonds that are too small, resulting
		// in split bonds that cost the user unnecessary tx fees.
```
I'm not sure I understand.

JoeGruffins: So in that scenerio, the extra bonds would be needed if our bonus tiers went down? Is this the same extra fees we would incur if we increased our target tier? You mean that they would expire at different times?

buck54321: Yes, they would expire at different times. Where normally we could send a single bond tx with bond weight 4 once every expiry, we would now be sending a weight 1 bond and the user would see their tier drop from 7 to 4 for some unknown reason. To fix it by broadcasting additional bonds would create "parallel tracks", adding more txs per expiry period.

JoeGruffins: I see! Thank you.

@martonp ok with re-posting all weak bonds?

martonp: > Posting 2 is correct, imo. We should aim to maintain `bonded_tier`, not `effective_tier = bonded_tier + bonus_tiers`. In fact, any attempt to maintain effective tier for low-tier targets would ultimately fail to maintain bonded status, as @JoeGruffins was seeing in #2459.

I think it's not that bad to repost all the bonds and ignore the bonuses.. it's just a few extra bonds being posted, but my understanding was that the reason all orders were being canceled was that there were no bonds at all. If we just ensure that there is at least one bond at all times, wouldn't that be prevented?

buck54321: Oh. We're basing trading limits on bonded tier anyway, so that's the value we have to maintain.
https://github.com/decred/dcrdex/blob/6a7d7132045502faa617a9309cecf0e0fb24c536/server/auth/auth.go#L815

buck54321 (commented):
I think we may have an issue with reserves handling in `authDEX`. We're basing `dc.acct.tierChange` on the effective tier.
https://github.com/decred/dcrdex/blob/6a7d7132045502faa617a9309cecf0e0fb24c536/client/core/core.go#L6723
and then below, **in the case of a re-auth**, we're basing our reserves updates on this value
https://github.com/decred/dcrdex/blob/6a7d7132045502faa617a9309cecf0e0fb24c536/client/core/core.go#L6768
but does that make sense with consideration for bonus tiers?

Similar concern for `handleTierChangeMsg`?

client/core/bond.go
```diff
@@ -541,19 +575,28 @@ func (c *Core) rotateBonds(ctx context.Context) {
 		}
 
 		if acctBondState.tierChange != 0 {
-			reserveDelta := bondOverlap * -acctBondState.tierChange * int64(bondAsset.Amt)
-			netReserveDeltas[acctBondState.bondAssetID] += reserveDelta // negative means free reserves
-			// After applying the net reserveDelta, we would normally *zero* the
+			// Reserve enough funds to recreate current bonds.
+			reserve := bondOverlap * uint64(acctBondState.liveStrength) * bondAsset.Amt
```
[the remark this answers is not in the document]
JoeGruffins: The reason for the change is that now the `acctBondState.tierChange` may not be reliable to decide how much to reserve. If bonus tiers go up then we will start reserving 0 even though we need to reserve enough for at least one bond.

I think the original logic was created this way to allow for less reserves as we needed less because of bonus bond tiers, and I attempted to save that logic, but maybe we should just simply reserve target tiers worth?

> The use of a bare liveStrength here is curious.

It was my thinking that we want enough to re-post all live bonds. That is strong and weak. I think strong and pending could also work, but pending, weak, and strong would be overkill, as we are making pending bonds in response to weak bonds so it would count those twice... Would the second return from `bondTotalInternal` be the same as `liveStrength` here?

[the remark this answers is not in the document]
JoeGruffins: Hmm, should this bit subtracted then from live bonds? bondOverlap indicates 2x current? If it was three then 3x current. Also I guess this does not include fees we should reserve? Can clean this up better after v2 ConnectResult? I'm unclear on exactly how much we want to reserve. Do we want to count bonus tiers? We must count any negative tiers... If we are -1 because of bad behavior, we will always need to reserve target + 1.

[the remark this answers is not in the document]
buck54321: > I think the original logic was created this way to allow for less reserves as we needed less because of bonus bond tiers, and I attempted to save that logic, but maybe we should just simply reserve target tiers worth?

Maybe, but trading limits scale with **bonded** tier. Bonus tiers are an illusion. I'm not certain why they exist at all, tbh. The server doesn't need to add them. They do nothing for the client as "tiers". Client should be more interested in score as a form of reputation and as a penalty buffer. I may propose removing these bonus tiers if I can get the upgrade logic worked out.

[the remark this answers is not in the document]
JoeGruffins: Ok. Since I am unsure what the ultimate plan for the bonus tiers is will leave this for now. If you want to absorb usable changes here into #2471 or have a different plan please do. I can remove the reserves changes from this as well.

I don't have a strong opinion on whether the user should have expanded trading limits for good behavior.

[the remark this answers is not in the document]
buck54321: I guess I was thinking that it should look something like this. https://github.com/buck54321/dcrdex/compare/04116f8...buck54321:dcrdex:reserves-adjustment

JoeGruffins:
Looks ok.

buck54321 (approved):
I still think this is a little bit off, but this should fix the bug and we can work on finding better solutions. Thanks for digging in, @JoeGruffins

PR #2463, opened 2023-08-04
multi: use average rate for market order rate display

### LINKED ISSUE #1969: UI: Market order rate can display average value
Currently **Market** order rate contains placeholder `market`:

image

it would probably be better to display **average** rate number there instead, so that the user will at least get a rough idea on how expensive his order was (instead of doing the calculation himself).

- probably better to prefix it with `~` sign
- maybe add a **tooltip** explanation too, clarifying that it's just an average value and that exact matches are listed below.

ukane-philemon:
The rate display for market orders on the `orders`, `order` and 'market'(on the "user-order card") page, including the `Actual Rate` field when exporting a market order, will use the average order rate instead of just zero or `market`. Closes #1969.

### DIFF
--- a/client/core/helpers.go
+++ b/client/core/helpers.go
@@ -344,7 +344,8 @@ func (ord *OrderReader) RateString() string {
 // AverageRateString returns a formatting string containing the average rate of
 // the matches that have been filled in an order.
 func (ord *OrderReader) AverageRateString() string {
-	if len(ord.Matches) == 0 {
+	nMatch := len(ord.Matches)
+	if nMatch == 0 {
 		return "0"
 	}
 	var baseQty, rateProduct uint64
@@ -352,7 +353,11 @@ func (ord *OrderReader) AverageRateString() string {
 		baseQty += match.Qty
 		rateProduct += match.Rate * match.Qty // order ~ 1e16
 	}
-	return "~ " + ord.formatRate(rateProduct/baseQty)
+	rateStr := ord.formatRate(rateProduct / baseQty)
+	if nMatch > 1 {
+		rateStr = "~ " + rateStr // "~" only makes sense if the order has more than one match.
+	}
+	return rateStr
 }
 
 // SwapFeesString is a formatted string of the paid swap fees.
--- a/client/core/helpers.go
+++ b/client/core/helpers.go
@@ -337,6 +337,11 @@ func (ord *OrderReader) RateString() string {
 	rateStr := ord.formatRate(ord.Rate)
 	if ord.Type == order.MarketOrderType {
 		rateStr = ord.AverageRateString()
+		if len(ord.Matches) > 1 {
+			rateStr = "~ " + rateStr // "~" only makes sense if the order has more than one match.
+		} else {
+			return "market" // "market" is better than 0 BTC/ETH ?
+		}
 	}
 	return fmt.Sprintf("%s %s/%s", rateStr, ord.QuoteUnitInfo.Conventional.Unit, ord.BaseUnitInfo.Conventional.Unit)
 }
@@ -354,9 +359,6 @@ func (ord *OrderReader) AverageRateString() string {
 		rateProduct += match.Rate * match.Qty // order ~ 1e16
 	}
 	rateStr := ord.formatRate(rateProduct / baseQty)
-	if nMatch > 1 {
-		rateStr = "~ " + rateStr // "~" only makes sense if the order has more than one match.
-	}
 	return rateStr
 }
 
--- a/client/core/helpers.go
+++ b/client/core/helpers.go
@@ -336,11 +336,13 @@ func (ord *OrderReader) SimpleRateString() string {
 func (ord *OrderReader) RateString() string {
 	rateStr := ord.formatRate(ord.Rate)
 	if ord.Type == order.MarketOrderType {
+		nMatches := len(ord.Matches) 
+		if nMatches == 0 {
+			return "market" // "market" is better than 0 BTC/ETH ?
+		}
 		rateStr = ord.AverageRateString()
 		if len(ord.Matches) > 1 {
 			rateStr = "~ " + rateStr // "~" only makes sense if the order has more than one match.
-		} else {
-			return "market" // "market" is better than 0 BTC/ETH ?
 		}
 	}
 	return fmt.Sprintf("%s %s/%s", rateStr, ord.QuoteUnitInfo.Conventional.Unit, ord.BaseUnitInfo.Conventional.Unit)
@@ -349,8 +351,7 @@ func (ord *OrderReader) RateString() string {
 // AverageRateString returns a formatting string containing the average rate of
 // the matches that have been filled in an order.
 func (ord *OrderReader) AverageRateString() string {
-	nMatch := len(ord.Matches)
-	if nMatch == 0 {
+	if len(ord.Matches) == 0 {
 		return "0"
 	}
 	var baseQty, rateProduct uint64
@@ -358,8 +359,7 @@ func (ord *OrderReader) AverageRateString() string {
 		baseQty += match.Qty
 		rateProduct += match.Rate * match.Qty // order ~ 1e16
 	}
-	rateStr := ord.formatRate(rateProduct / baseQty)
-	return rateStr
+	return ord.formatRate(rateProduct / baseQty)
 }
 
 // SwapFeesString is a formatted string of the paid swap fees.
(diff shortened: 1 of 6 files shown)


JoeGruffins (approved):

martonp (commented):
Rate is showing 0 on the markets page.

Screenshot 2023-08-06 at 3 19 51 AM

ukane-philemon:
> Rate is showing 0 on the markets page.
> 
> Screenshot 2023-08-06 at 3 19 51 AM

Thanks for testing. This might be possible if the match rate is zero(which should not be possible for testnet/mainnet/simnet). Please how can I reproduce this exact market order(I'm assuming this is simnet, no)?

JoeGruffins:
I see it too. The match card does not update until I refresh the page, and then the rate shows.

martonp:
> Please how can I reproduce this exact market order(I'm assuming this is simnet, no)?

Just open a buy limit order then open a market sell.

client/webserver/site/src/js/markets.ts
```diff
@@ -2257,7 +2258,7 @@ export default class MarketsPage extends BasePage {
 
   handleMatchNote (note: MatchNote) {
     const mord = this.metaOrders[note.orderID]
-    if (!mord) return this.refreshActiveOrders()
+    if (!mord || note.match.status === OrderUtil.NewlyMatched) return this.refreshActiveOrders()
```
ukane-philemon, line 2261: We didn't care about the new match before now, but now we do.

PS: @buck54321, is there a reason we don't consider the match `status` in `core/OrderReader.AverageRateString()`?

buck54321: Refreshing every active order in order to update two fields is a bit of overkill. A little more targeted approach here please.

client/webserver/site/src/js/orderutil.ts
```diff
@@ -119,6 +120,29 @@ export function settled (order: Order) {
   }, 0)
 }
 
+/* averageRateString returns a formatting string containing the average rate of
+the matches that have been filled in an order. */
+export function averageRateString (ord: Order): string {
+  const nMatch = ord.matches?.length
+  if (nMatch === 0) return '0'
+  let rateStr = Doc.formatCoinValue(app().conventionalRate(ord.baseID, ord.quoteID, averageRate(ord)))
+  if (nMatch > 1) rateStr = '~ ' + rateStr // "~" only makes sense if the order has more than one match.
```
buck54321, line 129: I guess this whole function only makes sense for market orders then? If so, the function name should reflect that fact.

client/webserver/site/src/js/markets.ts
```diff
@@ -1482,7 +1482,14 @@ export default class MarketsPage extends BasePage {
...
       details.qty.textContent = header.qty.textContent = Doc.formatCoinValue(ord.qty, market.baseUnitInfo)
-      details.rate.textContent = header.rate.textContent = Doc.formatRateFullPrecision(ord.rate, market.baseUnitInfo, market.quoteUnitInfo, cfg.ratestep)
+      let rate = ord.rate
+      let ratePrefix = ''
+      const nMatch = ord.matches?.length
+      if (ord.type === OrderUtil.Market && nMatch > 0) {
+        rate = OrderUtil.averageRate(ord)
+        if (nMatch > 1) ratePrefix = '~ ' // ~ only makes sense if the order has more than one match
+      }
+      details.rate.textContent = header.rate.textContent = ratePrefix + Doc.formatRateFullPrecision(rate, market.baseUnitInfo, market.quoteUnitInfo, cfg.ratestep)
```
buck54321, line 1492: This will show `0.000000` if a market order is in epoch or was a no-match. That's pre-existing, but not ideal. Maybe this is where we should show the (translated) string `'market'` if we have no matches.

client/core/helpers.go
```diff
@@ -310,29 +310,38 @@ func (ord *OrderReader) StatusString() string {
 
 // SimpleRateString is the formatted match rate.
 func (ord *OrderReader) SimpleRateString() string {
+	if ord.Type == order.MarketOrderType {
+		return ord.AverageRateString()
+	}
```
buck54321, line 315: `SimpleRateString` is used exclusively in CSV-formatted files where there is an expectation of machine-readability. `AverageRateString` can add tildes and whitespace `"~ "`. :Also, backwards compatibility issues. Better leave this as `"0"` for market orders here.

client/core/helpers.go
```diff
@@ -310,29 +310,38 @@ func (ord *OrderReader) StatusString() string {
...
 	return ord.formatRate(ord.Rate)
 }
 
 // RateString is a formatted rate with units.
 func (ord *OrderReader) RateString() string {
+	rateStr := ord.formatRate(ord.Rate)
 	if ord.Type == order.MarketOrderType {
-		return "market"
+		rateStr = ord.AverageRateString()
 	}
```
buck54321, line 324: Is `"0"` better than `"market"` when there are no matches?

client/core/helpers.go
```diff
@@ -310,29 +310,38 @@ func (ord *OrderReader) StatusString() string {
...
 	var baseQty, rateProduct uint64
 	for _, match := range ord.Matches {
 		baseQty += match.Qty
 		rateProduct += match.Rate * match.Qty // order ~ 1e16
 	}
-	return ord.formatRate(rateProduct / baseQty)
+	rateStr := ord.formatRate(rateProduct / baseQty)
+	if nMatch > 1 {
+		rateStr = "~ " + rateStr // "~" only makes sense if the order has more than one match.
```
buck54321, line 342: Actually, `AverageRateString` is used in CSV files too, so can't actually do this here at all. I think you need to confine all changes to `(*OrderReader).RateString`.

buck54321 (changes requested):

client/core/helpers.go
ukane-philemon: Okay.

client/core/helpers.go
ukane-philemon: Uhmm, I'm not too sure. But maybe we should say "market" if there are no matches.

client/core/helpers.go
martonp: I don't see the "~" in the exported CSV.

martonp:
Before I place a market order I see the following in the screen:
Screenshot 2023-08-21 at 5 22 48 AM

Then I place a market sell order, and I see just this row, and it is not clickable. Once there is a match, the other rows come back and it becomes a normal clickable row.
Screenshot 2023-08-21 at 5 22 57 AM

client/core/helpers.go
martonp: Or maybe just a "~".

buck54321: `"`market"` is better.

client/core/helpers.go
ukane-philemon: I think already fixed @buck54321 concern before you tested.

client/core/helpers.go
ukane-philemon: > `"`market"` is better.

*IF* there are no orders, yh?

ukane-philemon:
> Before I place a market order I see the following in the screen: Screenshot 2023-08-21 at 5 22 48 AM
> 
> Then I place a market sell order, and I see just this row, and it is not clickable. Once there is a match, the other rows come back and it becomes a normal clickable row. Screenshot 2023-08-21 at 5 22 57 AM

I'm unable to reproduce this @martonp, Is anyone else seeing this?

martonp:
https://github.com/decred/dcrdex/assets/6186350/441d1ea6-1138-447a-89c0-1cc3ae4b87e2

ukane-philemon:
@martonp, I was able to reproduce the mishap you noticed and it has been fixed in https://github.com/decred/dcrdex/pull/2463/commits/d265f2b918ce17d32bc4243f21c3d3a1f0ec6f53.

martonp (approved):
LGTM

buck54321 (changes requested):

client/webserver/site/src/js/markets.ts
```diff
@@ -1533,6 +1536,17 @@ export default class MarketsPage extends BasePage {
     this.setDepthMarkers()
   }
 
+  /*
+   marketOrderRateString uses the market config rate step to format the average
+   market order rate.
+  */
+  marketOrderRateString (ord: Order, mkt: CurrentMarket) :string {
+    if (!ord.matches) return intl.prep(intl.ID_MARKET_ORDER)
```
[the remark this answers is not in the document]
ukane-philemon: Okay, thanks.

buck54321:
Ready but needs a rebase.

ukane-philemon:
@buck54321, do you want me to squash?

buck54321:
> @buck54321, do you want me to squash?

Nah. I'll squash when I merge.

buck54321 (approved):

PR #2464, opened 2023-08-06
webserver: fix live-test

ukane-philemon:
These are just some changes I think the webserver live test can use. I think it needs more improvement and could work on them little by little.

- Set "initErrors" to false to allow for error-free account creation.
- Initiate and use math/rand.Rand instead of deprecated math/rand.Seed.
- Set the Rate and MatchID feild for user order match.
- Use caser.Title instead of deprecated string.Title.

### DIFF
--- a/client/webserver/live_test.go
+++ b/client/webserver/live_test.go
@@ -12,7 +12,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"math"
-	"math/rand"
+	mrand "math/rand"
 	"sort"
 	"strconv"
 	"strings"
@@ -39,6 +39,8 @@ import (
 	dexbtc "decred.org/dcrdex/dex/networks/btc"
 	"decred.org/dcrdex/dex/order"
 	ordertest "decred.org/dcrdex/dex/order/test"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
 )
 
 const (
@@ -64,6 +66,9 @@ var (
 	doubleCreateAsyncErr  = false
 	randomizeOrdersCount  = false
 	initErrors            = false
+
+	rand   *mrand.Rand
+	titler = cases.Title(language.AmericanEnglish)
 )
 
 func dummySettings() map[string]string {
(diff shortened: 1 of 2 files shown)


JoeGruffins (commented):
I actually cannot run the tests on master because I have ipv6 turned off on my machine.
```
--- FAIL: TestServer (0.01s)
    live_test.go:2036: Connect error: connect failure: Can't listen on [::1]:54321. web server quitting: listen tcp [::1]:54321: socket: address family not supported by protocol
```

This diff seems to give me a panic:
```
$ go test --tags=live
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x531a58]

goroutine 1 [running]:
math/rand.(*Rand).Int63(...)
        /usr/lib/go/src/math/rand/rand.go:95
math/rand.(*Rand).Int31(...)
        /usr/lib/go/src/math/rand/rand.go:109
math/rand.(*Rand).Int31n(0x0, 0xa)
        /usr/lib/go/src/math/rand/rand.go:144 +0x38
math/rand.(*Rand).Intn(0x1736f60?, 0xc000041e60?)
        /usr/lib/go/src/math/rand/rand.go:182 +0x2e
decred.org/dcrdex/client/webserver.mkDexAsset({0x199f064, 0x3})
        /home/joe/git/dcrdex/client/webserver/live_test.go:236 +0xe5
decred.org/dcrdex/client/webserver.init()
        /home/joe/git/dcrdex/client/webserver/live_test.go:300 +0x594
exit status 2
FAIL    decred.org/dcrdex/client/webserver      0.023s
```

ukane-philemon:
> This diff seems to give me a panic:

Whoops! Fixed in https://github.com/decred/dcrdex/pull/2464/commits/fd6cb851f44352bbe08fa7f3b8e138f7b21b4b72. I also addressed this `I actually cannot run the tests on master because I have ipv6 turned off on my machine.`.

JoeGruffins (approved):
Thanks for changing that!

Issue #2264, opened 2023-03-28
client/dcr: Add ticket purchasing.


JoeGruffins:
Part of proposal https://proposals.decred.org/record/ca6b749

Proposed interface:
```
type TicketBuyer interface {
	TicketPrice() (uint64, error)
	SetVSP(string) error // For SPV only, RPC wallets must set on wallet start same as mixing
	PurchaseTickets(n int) error
	Tickets() ([]*Ticket, error)
        VotingPreferences()([]*Agenda, []*TspendPolicy, []*TreasuryPolicy, error)
	SetVotingPreferences(choices map[string]uint8, tspendPolicy map[string]string, treasuryPolicy map[string]string) error
}
```

As I understand it we are not currently interested in the dcrwallet's automatic ticket buyer functionality. Only manual ticket purchasing.

<details closed>
<summary>change log</summary>
23/03/29 - Condense Policy getters
</details>

JoeGruffins:
@buck54321 added methods from Treasury related policies. I guess we want those from the beginning.

Should we allow purchasing tickets without a vsp? I would think no? Related but I can't find an rpc method to check on VSP set in config.

I believe the wallet will pay the vsp fee for us. If for some reason the fee payment tx falls through, I'm not sure if it tracks that or tries to pay again automatically.

I guess we also need `processUnmanagedTicket(txhash) error` in case of wallet being restored?

I doesn't look like we need grpc support.

Also please assign me.

JoeGruffins:
Should `TicketPrice() (uint64, error)` include VSP fee or that be a separate return?

JoeGruffins:
@jholdstock tells me on matrix we don't need to worry about the VSP fee, wallet will continue to retry.

buck54321:
> @buck54321 added methods for Treasury related policies. I guess we want those from the beginning.

Probably? We didn't commit to anything tspend stuff in the prop, but it is core functionality. I'm not well versed on the details though. What's the difference between "treasury policy" and "tspend policy".

Can `Agendas`, `TspendPolicy`, and `TreasuryPolicy` be combined into a single state getter?

 > Should we allow purchasing tickets without a vsp?

Definitely not now. Maybe never. 

> Related but I can't find an rpc method to check on VSP set in config.

That's unfortunate. We may need to open some PRs for dcrwallet. Also want the `ticketbuyer.balancetomaintainabsolute` setting for RPC wallets. For native wallets, we can work around this stuff if needed.

[...]

PR #2466, opened 2023-08-07
asset: add ListVSPs method to TicketBuyer

buck54321:
... so that users can select a voting service provider through the dexc GUI.

### DIFF
--- a/client/asset/dcr/coin_selection_test.go
+++ b/client/asset/dcr/coin_selection_test.go
@@ -1,3 +1,5 @@
+//go:build !harness && !vspd
+
 package dcr
 
 import (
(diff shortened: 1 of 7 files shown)


JoeGruffins (approved):
Looks great.

Do you want to work on the UI? I'd like to do it but idk if I'm the guy you want. Or if someone else wants to take it on that's better with web than me.

If noone has started on it by the time rpc is cleaned up I will start on it. I'll also post a note on the issue.
 
Tagging issue #2264

buck54321:
I'm playing with some UI stuff already. I'd like to get the basic structure in place and then we can all take shots at improvements.

JoeGruffins:
I was testing out on simnet, and noticed the returned fees from `StakeStatus` don't look right for the rpc wallet `Fees: (uint64) 18446744073709548636,`. They look sane for the internal wallet though. Not related to these changes, just wanted to mention somewhere.

PR #2471, opened 2023-08-11
auth: v2 ConnectResult

### LINKED ISSUE #2461: bonds: Client lowering tier may become unbonded for extended period.
If the client starts off bonded with a large amount of bonds and then lowers their target tier, they may be locked from bonding by the max bonded amount until the bond expiry has elapsed.

To replicate on simnet, you can use the `simnet-setup.sh` but set the bonded amount from `1000000000` to `8000000000` (anything over 40 dcr ). Log in through web and set tier to 1 after a minute. When the first bonds are about to expire, you will get a warning `2023-08-04 09:52:13.873 [WRN] CORE: Unable to post new bond with equivalent of 80.00000000 DCR currently bonded (limit of 40.00000000 DCR)` And be locked from creating new bonds and at tier 0 until the old ones are refunded.

buck54321:
Fully resolves #2461, though we're in a precarious state if the server is < v2 connect result (see notes here and in #2460). 

Add a `TierReportV2` field to `ConnectResult`, `TierChangedNotification`, and `PostBondResult`. The `TierReportV2 account.TierReport` field deprecates the `Tier int64` and `LegacyFeePaid bool` fields in these responses, though the fields are still populated (`V0PURGE`).

The `account.TierReport` includes information about **bonus tiers** and **revoked tiers**, allowing the client to maintain a their **bonded tier** accurately.

### DIFF
--- a/client/core/bond.go
+++ b/client/core/bond.go
@@ -14,6 +14,7 @@ import (
 	"decred.org/dcrdex/dex"
 	"decred.org/dcrdex/dex/keygen"
 	"decred.org/dcrdex/dex/msgjson"
+	"decred.org/dcrdex/server/account"
 	"github.com/decred/dcrd/dcrec/secp256k1/v4"
 	"github.com/decred/dcrd/hdkeychain/v3"
 )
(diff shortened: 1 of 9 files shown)


buck54321:
Replaced by #2501

PR #2473, opened 2023-08-14
wiki: Update Native-BTC-Wallet-Recovery-and-Rescanning.md and Home.md

ukane-philemon:
Part of #1641.

Now that ltc and bitcoin cash also support rescanning, I think they should be included somehow in the guide. However, that will come in a separate work.

PR #2474, opened 2023-08-14
multi: fix bug in `spvWallet.reconfigure(...)`

ukane-philemon:
While testing and updating the wiki for PR #2473, I encountered a panic caused by attempting to reconfigure an unloaded btc wallet.

This diff addresses the issue mentioned above.

To reproduce using BTC: 
1. Delete the wallet dir or `wallet.db` file and start DEX
2. Attempt to reconfigure the wallet from the BTC settings page.


buck54321:
Can you provide instructions for reproducing?

ukane-philemon:
> A better solution would be to recognize the problem on the front end and to 1) hide unavailable UI elements, and 2) show messaging directing users to appropriate documentation.

I agree, we can also do some work on the FE. Although the code here: https://github.com/decred/dcrdex/blob/master/client/core/core.go#L3406-L3435 provides us with the ability to fix the "wallet not found" issue during wallet reconfiguration, disabling the `Submit` button on the settings page and asking the user to do full recovery will suffice as well.

So instead of "ReconfigureWallet" doing the job, it should return an error if the wallet is not found? Seems the block of code above will become useless and we can remove some of it(the parts that create the missing wallet).

ukane-philemon:
The `Submit` btn is disabled and a support msg is displayed if an SPV wallet is not running.

Screenshot 2023-08-15 at 5 55 19 PM

client/webserver/site/src/js/wallets.ts
```diff
@@ -1084,11 +1087,28 @@ export default class WalletsPage extends BasePage {
     this.updateDisplayedReconfigFields(currentDef)
   }
 
+  showOrHideRecoverySupportMsg (wallet:WalletState, forceHide?: boolean) {
+    const page = this.page
+    if ((!wallet.running && !wallet.disabled) && wallet.traits & traitRecoverer && !forceHide) {
+      page.reconfigSupportMsg.textContent = intl.prep(intl.ID_WALLET_RECOVERY_SUPPORT_MSG, { walletType: wallet.type })
+      Doc.show(page.reconfigSupportMsg)
+      page.submitReconfig.setAttribute('disabled', '')
```
buck54321, line 1095: Could add this rule to `wallets.scss`

```SCSS
#submitReconfig[disabled] {
  cursor: not-allowed;
}

```

buck54321 (changes requested):

client/webserver/site/src/js/locales.ts
```diff
@@ -262,7 +263,8 @@ export const enUS: Locale = {
   [ID_ORDER]: 'Order',
   [ID_LOCKED_ORDER_BAL_MSG]: 'Funds locked in unmatched orders',
   [ID_CREATING_WALLETS]: 'Creating wallets',
-  [ID_ADDING_SERVERS]: 'Connecting to servers'
+  [ID_ADDING_SERVERS]: 'Connecting to servers',
+  [ID_WALLET_RECOVERY_SUPPORT_MSG]: 'Note: Your {{ walletType }} wallet is not running! Click on the "Recover" button below to fix it.'
```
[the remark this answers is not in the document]
ukane-philemon: Used the wallet symbol instead.

JoeGruffins (approved):
Looks great, working well for me. I recently had a lot of problems syncing the native btc wallet on mainnet and I think this will help. I had to recover the wallet twice before it finally synced. I don't think it's a problem on our end, and it seems deleting and starting again is the easiest solution.

ukane-philemon:
>  recently had a lot of problems syncing the native btc wallet on mainnet

I've seen a lot of panic recently while syncing on mainnet but I don't have much information on the cause yet.

ukane-philemon:
Rebased.

client/asset/btc/spv_wrapper.go
```diff
@@ -268,10 +268,21 @@ var _ tipNotifier = (*spvWallet)(nil)
...
 // flag is set.
 func (w *spvWallet) reconfigure(cfg *asset.WalletConfig, currentAddress string) (restartRequired bool, err error) {
+	// If the wallet type is not SPV, then we can't reconfigure the wallet.
 	if cfg.Type != walletTypeSPV {
 		restartRequired = true
 		return
 	}
+
+	// Check if the SPV wallet exists. If it doesn't, then we can't reconfigure it.
+	exists, err := walletExists(w.dir, w.chainParams)
```
buck54321, line 278: When the file is corrupted, will we see an error here?

client/webserver/site/src/js/wallets.ts
```diff
@@ -1086,11 +1089,28 @@ export default class WalletsPage extends BasePage {
...
+    Doc.hide(page.reconfigSupportMsg)
+  }
+
   changeWalletType () {
     const page = this.page
     const walletType = page.changeWalletTypeSelect.value || ''
     const walletDef = app().walletDefinition(this.selectedAssetID, walletType)
     this.reconfigForm.update(this.selectedAssetID, walletDef.configopts || [], false)
+    const wallet = app().walletMap[this.selectedAssetID]
+    this.showOrHideRecoverySupportMsg(wallet, walletDef.seeded)
```
buck54321, line 1113: When I have a borked native wallet and I change the wallet type to RPC, the recovery support message is correctly hidden, but that's only because the rpc wallet isn't seeded. If we, for instance, had two seeded wallet types, and the user tried to change to the other type, the message would still show and the user would not be able to submit. The logic should be more like this https://github.com/buck54321/dcrdex/compare/ccc0c7e...buck54321:dcrdex:recovery-viz

buck54321 (changes requested):

client/asset/btc/spv_wrapper.go
ukane-philemon: If any of the required wallet files are missing.

buck54321 (approved):

